PR #2638, opened 2018-11-20
Update mpas-source for correction to prevent thermo energy conservation error

jonbob:
This PR lowers the permitted surface flux error in the mushy layer picard iteration to 0.9*ferrmax. It also adds warning field to conservation check and corrects warning diagnostics which allocate errors to conduction/melt/growth. See [MPAS PR #116](https://github.com/MPAS-Dev/MPAS-Model/pull/116)

* Tested in two-year run of BGCEXP_BCRC_CNPRDCTC_1850.ne30_oECv3 on anvil

[non-BFB]

PR #2966, opened 2019-06-03
Update mpas-seaice and supporting scripts

jonbob:
This PR brings in a new mpas-source submodule that updates the mpas-seaice codebase, as well as supporting scripts. The seaice codebase brings in several small updates as well as introducing a snicar feature, which is turned off by default. It also makes a small modification to a convergence criteria to prevent thermo energy conservation errors that has been tested in the BGC v1 runs (see PR #2638) that can cause small answer changes in certain situations.

[NML]
[non-BFB]


jonbob:
merged to next

jonbob:
merged to master

PR #2968, opened 2019-06-04
Make cice build using CMake!

jgfouca:
[BFB]

### DIFF
--- a/components/cice/bld/configure
+++ b/components/cice/bld/configure
@@ -518,7 +518,7 @@ if ($print>=2) { print "CPP definitions set by configure: \'$cfg_cppdefs\'$eol";
 #-----------------------------------------------------------------------------------------------
 
 my $fp_filename      = 'Filepath';             # name of output filepath file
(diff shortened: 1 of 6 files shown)


components/cice/bld/configure
```diff
@@ -518,7 +518,7 @@ if ($print>=2) { print "CPP definitions set by configure: \'$cfg_cppdefs\'$eol";
 #-----------------------------------------------------------------------------------------------
 
 my $fp_filename      = 'Filepath';             # name of output filepath file
-my $cpp_filename     = 'CICE_cppdefs';         # name of output file for cice's cppdefs
```
jonbob: Is there a reason we use "CCSM_cppdefs" instead of something with E3SM or CIME?

jgfouca: It's a legacy name. I just changed it to match the other components.

jonbob: OK, just curious

jgfouca:
Merged to next.

PR #2969, opened 2019-06-04
Updates SBeTR module to simplify cmake for ELM

bishtgautam:
In SBeTR submodule, all `use bshr_assert_mod, only <>` statements are removed from
`bshr_assert.h` and explicitly added them in all files that have `#include "bshr_assert.h"`.
This simplifies E3SM Cmake for ELM.

[BFB]


jgfouca:
Merged to next.

PR #2972, opened 2019-06-05
Add UA and COARE ocean surface flux schemes

JackReevesEyre:
This branch adds two algorithms to calculate ocean surface fluxes (wind stress, latent and sensible heat fluxes) in CIME. The new algorithms are selected by a new CIME namelist variable: 

ocn_surface_flux_scheme = 0  [default] -- original scheme as used in E3SMv1;
                                           = 1               -- COARE v3.0 scheme (from Fairall et al. 2003);
                                           = 2               -- University of Arizona (UA) scheme (from Zeng et al., 1998).

The desired option should be added to user_nl_cpl.

[BFB]
[FCC]
[NML]

EW-108, EW-127

### DIFF
--- a/cime/src/drivers/mct/cime_config/namelist_definition_drv.xml
+++ b/cime/src/drivers/mct/cime_config/namelist_definition_drv.xml
@@ -698,7 +698,7 @@
      </values>
   </entry>
 
-  <entry id="flux_scheme">
+  <entry id="ocn_surface_flux_scheme">
      <type>integer</type>
      <category>control</category>
      <group>seq_infodata_inparm</group>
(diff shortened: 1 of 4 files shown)


kaizhangpnl:
@singhbalwinder Jack and I tested the branch independently for BFB. It is also reproducible from a restart. @JackReevesEyre What does the tag FCC mean?

JackReevesEyre:
@kaizhangpnl FCC means "flag climate changing": using the COARE or UA scheme changes the climate of the model compared to the default scheme. The differences are greater than just round off. Is that correct usage of this label?

rljacob:
Yes it is.

kaizhangpnl:
Now I remember seeing this tag before. :) Thanks!

singhbalwinder:
Merged to next

kaizhangpnl:
Thank you, Balwinder!

JackReevesEyre:
Thank you all!

PR #2981, opened 2019-06-07
Workaround bug in cmake f90 dependency scanner

jgfouca:
Cmake continues to struggle to find the correct .mod dependencies for some cice files.
These seems to happen when the file has certain preprocessor directives and the
use statements are within a subroutine.

[BFB]

### DIFF
--- a/components/cice/src/source/ice_shortwave.F90
+++ b/components/cice/src/source/ice_shortwave.F90
@@ -56,6 +56,15 @@
       use ice_constants
       use ice_blocks
       use ice_diagnostics
+      use ice_meltpond
+      use ice_calendar
+      use ice_domain
+      use ice_flux
+      use ice_grid
+      use ice_itd
+      use ice_orbital
+      use ice_state
+      use ice_therm_vertical, only: heat_capacity
       use ice_communicate, only: my_task
       use perf_mod, only: t_startf, t_stopf, t_barrierf
 !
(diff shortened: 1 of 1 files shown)


rljacob:
Does the file have to have both cpp directives AND use statements in a subroutine?   The latter is completely legal Fortran and we can't remove all instances of it.

jgfouca:
@rljacob , yes. @ambrad and I tracked the problem down to:

```
#ifdef CCSM3FRC
         dalvdrn_ccsm3, & ! visible direct albedo (diag)    (fraction)
         dalidrn_ccsm3, & ! near-ir direct albedo (diag)    (fraction)
         dalvdfn_ccsm3, & ! visible diffuse albedo (diag)   (fraction)
         dalidfn_ccsm3, & ! near-ir diffuse albedo (diag)   (fraction)
#endif
```

Any 'use' statements that occur within subroutines after this preprocessor segment will not get detected by CMake's dependency scanner.

ambrad:
@jgfouca, have you been able to make a minimal breaking example modeling this issue? So far I have failed. The following, plus a number of variations on it, all yield the correct `CMakeFiles/mbe.dir/depend.make` file. I'm wondering if there's an extreme subtlety here, or if maybe the particular environment you're building in is part of the issue.
```
$ for i in CMakeLists.txt *.F90; do echo ">> $i"; cat $i; done
>> CMakeLists.txt
cmake_minimum_required (VERSION 3.5)

project (mbe Fortran)

set (SOURCES a.F90 b.F90 main.F90)
add_executable (mbe ${SOURCES})
>> a.F90
module a_mod

#ifdef CCSM3FRC
  integer :: hi
#endif

contains

  subroutine foo()
    use b_mod
  end subroutine foo
end module a_mod
>> b.F90
module b_mod
contains

  subroutine bar()
  end subroutine bar
end module b_mod
>> main.F90
program main
  use a_mod, only: foo
end program main
```

jgfouca:
@ambrad , no, but I plan on trying to do so this week.

PR #2982, opened 2019-06-07
Make mpas-ocean build using CMake!

jgfouca:
[BFB]


jgfouca:
Is dependent on https://github.com/MPAS-Dev/MPAS-Model/pull/277

rljacob:
Replaced by #3043

PR #2994, opened 2019-06-12
Update mpas-seaice to bring in improvements to the snow model

jonbob:
This PR brings in a new mpas-source submodule with updates only to the seaice core. These changes add improvements to the seaice snow model, though they are turned off by default and therefore should be bfb. The snow model does add some new variables to the seaice Registry and corresponding namelists.

[BFB]
[NML]

### DIFF
--- a/components/mpas-seaice/bld/build-namelist
+++ b/components/mpas-seaice/bld/build-namelist
@@ -485,16 +485,22 @@ if ($CONTINUE_RUN eq 'TRUE') {
         add_default($nl, 'config_do_restart_hbrine', 'val'=>".true.");
         add_default($nl, 'config_do_restart_zsalinity', 'val'=>".true.");
         add_default($nl, 'config_do_restart_bgc', 'val'=>".true.");
+        add_default($nl, 'config_do_restart_snow_density', 'val'=>".true.");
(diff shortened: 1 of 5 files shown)


jonbob:
passes an ERS_P480_Ld5.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.anvil_intel test, results are BFB with a run from master

njeffery:
@jonbob :  Completed a successful 10 year GCASE with 5 snow layers and all the new snow mods active. The code works as expected with options on, though the PR has options off by default. 

Some additional details to use the new snow tracers:
with config_use_snow_grain_radius = true
add to streams "SnowAgingPropertiesInput" which reads the file snicar_drdt_bst_fit_60_c04262019.nc

When using the zaerosols:
config_use_shortwave_bioabsorption = true
add to streams "StandardAerosolsInput" and "ModalAerosolsInput"
which reads the file "standard_optics_mpas_cice.nc" and "snicar_optics_5bnd_mam_c140303_mpas_cice.nc". respectively

jonbob:
merged to next

jonbob:
Merged to master

PR #2997, opened 2019-06-13
Building homme without netcdf and pio

oksanaguba:
Add an option to build homme without pio and netcdf libraries, not active by default. This is needed for performance runs when installing netcdfF is not trivial. Affects only homme, homme suite on anvil passes.

[bfb]


oksanaguba:
This commit ran on 1 rank. I will try to make it a proper PR now.

jayeshkrishna:
@oksanaguba : Just curious, is this PR to optionally build HOMME without PIO ?

oksanaguba:
... except that there is no motivation to do sweat without pio, too. so this feature will be ok only when building preqx* and theta* execs.

oksanaguba:
@jayeshkrishna yes, we have this issue once in a while that netcdf F is not built on a system. Then we usually end up building it locally, but I had some issues doing this, so, I decided to comment pio calls and cmake build stuff in standalone homme. This is when movies are not needed, for example, for performance runs.

mt5555:
@jayeshkrishna : any chance PIO2 will support being built without any external libraries, and just write binary data?  or just ignore all write requests?

oksanaguba:
Homme tests on anvil (with the usual build) pass. This is ready.

components/homme/CMakeLists.txt
```diff
@@ -315,22 +331,25 @@ IF (CPRNC_DIR)
...
-  set(NETCDF_INCLUDE_DIR ${NetCDF_Fortran_INCLUDE_DIRS})
+    # cprnc's cmake wont search for netcdf, so we have to find it first:
+    FIND_PACKAGE(NetCDF "4.0" COMPONENTS C Fortran)
+    set(NETCDF_LIBRARIES ${NetCDF_Fortran_LIBRARIES})
+    set(NETCDF_INCLUDE_DIR ${NetCDF_Fortran_INCLUDE_DIRS})
 
-  # needed for CPRNC build system
-  file(WRITE "${CMAKE_CURRENT_SOURCE_DIR}/utils/cime/tools/cprnc/Macros.cmake" "#dummy Macros file for non-CIME machines")
+    # needed for CPRNC build system
+    file(WRITE "${CMAKE_CURRENT_SOURCE_DIR}/utils/cime/tools/cprnc/Macros.cmake" "#dummy Macros file for non-CIME machines")
```
ambrad: If you want to solve another problem in this PR, you could bring in this commit:
https://github.com/ambrad/E3SM/commit/fa48fd298678be89bf36357db731f4b10215d548
It runs the file line above only if the file is missing. This solves the auto-reconfig issue when doing multiple builds from the same repo.

components/homme/cmake/HommeMacros.cmake
```diff
@@ -123,7 +123,12 @@ macro(createTestExec execName execType macroNP macroNC
   # Add this executable to a list
   SET(EXEC_LIST ${EXEC_LIST} ${execName} CACHE INTERNAL "List of configured executables")
 
-  TARGET_LINK_LIBRARIES(${execName} pio timing ${COMPOSE_LIBRARY} ${BLAS_LIBRARIES} ${LAPACK_LIBRARIES})
+  if(BUILD_HOMME_WITHOUT_PIOLIBRARY)
+    TARGET_LINK_LIBRARIES(${execName} timing ${COMPOSE_LIBRARY} ${BLAS_LIBRARIES} ${LAPACK_LIBRARIES})
+  ELSE()
+    TARGET_LINK_LIBRARIES(${execName} pio timing ${COMPOSE_LIBRARY} ${BLAS_LIBRARIES} ${LAPACK_LIBRARIES})
```
ambrad: So that there aren't two lines with lists have N-1 redundant entries, you could make a list with the N-1 entries, then use `list(APPEND ...)` for pio.

components/homme/cmake/HommeMacros.cmake
```diff
@@ -110,7 +110,7 @@ macro(createTestExec execName execType macroNP macroNC
   ADD_EXECUTABLE(${execName} ${EXEC_SOURCES})
   SET_TARGET_PROPERTIES(${execName} PROPERTIES LINKER_LANGUAGE Fortran)
 
-  IF (${CXXLIB_SUPPORTED})
+  IF (CXXLIB_SUPPORTED_CACHE)
```
ambrad: Is this fixing a separate issue? That's fine, but I'm just checking this bit is orthogonal to the PIO part to be sure I understand what's going on.

components/homme/src/share/prim_driver_base.F90
```diff
@@ -283,13 +291,17 @@ subroutine prim_init1_geometry(elem, par, dom_mt)
     ! ===============================================================
     ! Allocate and initialize the graph (array of GridVertex_t types)
     ! ===============================================================
+#ifndef HOMME_WITHOUT_PIOLIBRARY
```
ambrad: This seems to say that if the user requested an unstructured mesh (e.g., RRM), and PIO isn't available, then it will silently build a structured mesh. Is that true? If so, might it be better to throw an informative error the first time that issue is hit? (Ex: "Mesh files are not supported when BUILD_HOMME_WITHOUT_PIOLIBRARY=TRUE.")

components/homme/src/share/prim_driver_base.F90
```diff
@@ -474,13 +492,17 @@ subroutine prim_init1_geometry(elem, par, dom_mt)
 
     if (topology=="cube") then
        if(par%masterproc) write(iulog,*) "initializing cube elements..."
+#ifndef HOMME_WITHOUT_PIOLIBRARY
```
ambrad: Some of this code is getting cluttered with cpp stuff. Does it really not compile w/o these? If it _does_ compile and is just a logical/usage error, I recommend erroring with an informative message but removing all the ifdefs. Ideally, the ifdefs would surround just lines of code that won't compile, e.g., the lines in the unstructured mesh module.

components/homme/src/common_movie_mod.F90
```diff
@@ -3,6 +3,7 @@
 #endif
 
 module common_movie_mod
+#ifndef HOMME_WITHOUT_PIOLIBRARY
```
ambrad: In my opinion, it's better to put all the ifdefs here at the subroutine level and let the callers call these functions as usual. That way, caller code doesn't get cluttered with ifdefs. Ifdef at the tightest point possible, thus minimizing the number (ideally to 0) in non-I/O-implementation code.

components/homme/CMakeLists.txt
oksanaguba: I could, but by some reason I thought Luca would get it into his refactor branch?

components/homme/cmake/HommeMacros.cmake
oksanaguba: It is separate, this check did not work (we prob. never ran on machine without this flag supported?).

components/homme/src/share/prim_driver_base.F90
oksanaguba: So, should I throw this message at config time? Homme does not build with/without meshes, the same exec usually should handle both. This 'without pio' option is not done very nicely, but it serves its purpose and we won't use it regularly.

components/homme/CMakeLists.txt
ambrad: That branch won't go into master for a long time, so this PR might be a better place. Just a suggestion.

components/homme/src/share/prim_driver_base.F90
ambrad: IMO, throw the error at runtime when the netcdf-based mesh logic first starts.

oksanaguba:
I am trying to get only function bodies ifdef-ed, but etting issues with common_io_mod and common_movie_mod for start. my understanding is that cleanest is to disable subroutines bodies only. but some types are based on pio types. for example, common_io_mod uses nf_double  and var_desc_tfrom pio, if i ifdef that block of code but leave the rest of definitions in the mod, it will not work because some of derived types depend on these two. also, function signatures depend on these types. so, i do not see how to do it cleanly. instead, waht if i work backward and free only prim_main and prim_driver mod from new ifdefs? it is also not very clean.

oksanaguba:
Not sure the last version is acceptable. If yes, then there are a few things to do: fix list append, bring AMB's fix for finding cprnc, make a warning message about mesh runs without pio, and maybe set pio_interp and not vertical_interpolation in cmake if without_pio is on.

components/homme/src/prim_main.F90
```diff
@@ -188,7 +188,9 @@ program prim_main
      call abortmp("Please get sure the directory exist or specify one via output_dir in the namelist file.")
   end if
 #endif
-  
+
```
ambrad: Might be good to remove all the whitespace changes in this file.

ambrad:
Btw, re: netcdf-based types: in this case you can simply define types as, e.g., integer, when building w/o netcdf. Then all subroutines and variable decls will work automatically. But that's optional. The main thing was not to pollute app code with ifdefs, which is now done.

oksanaguba:
Builds look ok but build without pio does not run: ` writing restart header, restnum=000000000`, seems like it is trying to write restart because i prob. disabled code that reads restart frequency.

oksanaguba:
restart issue seems to be fixed (need to rerun both builds on anvil), if restart freq < nmax, code will abort.

oksanaguba:
Two builds, with and without pio library, on anvil get the same stdout output with this namelist:
```
&ctl_nl
NThreads=1
partmethod    = 4
topology      = "cube"
test_case     = "jw_baroclinic"
u_perturb = 1
rotate_grid = 0
ne=20
qsize = 4
nmax=100
statefreq=1
!restartfreq   = 999999999
!restartfile   = "./R0001"
runtype       = 0
mesh_file='/dev/null'
tstep=10      ! ne30: 300  ne120: 75
rsplit=2       ! ne30: 3   ne120:  2
qsplit = 1
tstep_type = 5
integration   = "explicit"
nu=1e13
nu_div=2.5e13
nu_p=1e13
nu_q=1e13
nu_s=1e13
nu_top = 2.5e5
se_ftype     = 0
limiter_option = 9
vert_remap_q_alg = 1
hypervis_scaling=0
hypervis_order = 2
hypervis_subcycle=4    ! ne30: 3  ne120: 4
theta_hydrostatic_mode=.true.
/

&vert_nl
vform         = "ccm"
vfile_mid = './sabm-20.ascii'
vfile_int = './sabi-20.ascii'
/

&prof_inparm
profile_outpe_num = 100
profile_single_file             = .true.
/

&analysis_nl
output_prefix     = "camBaroMoist-"
interp_gridtype   = 2
output_timeunits  = 1,1
output_frequency  = 3,3
output_start_time = 0,0
output_end_time   = 30,30
output_varnames1  = 'zeta', 'u', 'v', 'ps', 'dp3d'
output_varnames2  = 'Q', 'Q2', 'Q3', 'Q4','phys_lat','phys_lon'
io_stride         = 8
output_type       = 'netcdf'
/
```

oksanaguba:
homme tests pass on anvil.

components/homme/CMakeLists.txt
```diff
@@ -315,22 +327,30 @@ IF (CPRNC_DIR)
...
+    # cprnc's cmake wont search for netcdf, so we have to find it first:
+    FIND_PACKAGE(NetCDF "4.0" COMPONENTS C Fortran)
+    set(NETCDF_LIBRARIES ${NetCDF_Fortran_LIBRARIES})
+    set(NETCDF_INCLUDE_DIR ${NetCDF_Fortran_INCLUDE_DIRS})
+
+    # needed for CPRNC build system
+    file(WRITE "${CMAKE_CURRENT_SOURCE_DIR}/utils/cime/tools/cprnc/Macros.cmake" "#dummy Macros file for non-CIME machines")
+
+    set (cprnc_dummy_file "${CMAKE_CURRENT_SOURCE_DIR}/utils/cime/tools/cprnc/Macros.cmake")
+    if (NOT EXISTS "${cprnc_dummy_file}")
```
ambrad: This is the fix, but the equivalent line above, line 343 in this SHA1, should be removed. The point is to write the file only if it doesn't exist.

oksanaguba: thanks, done.

oksanaguba:
Merged to next.

PR #2998, opened 2019-06-13
Update history script infrastructure to allow writing .h0. files on other time frequencies than monthly

### LINKED ISSUE #3034: ELM cannot output multiple monthly values in a single history file
From @ckoven, There is an old bug that ELM inherited from CLM, which is that if the user specifies the first argument of hist_mfilt to be anything other than 1, the fortran code resets that value to 1 anyway.  I.e. it isn't possible to write the first set of history fields with a frequency that isn't equal to one file per each time unit (specified by hist_nhtfrq).  In general this is ok, but when doing large and long NINST ensembles this behavior leads to a very cluttered run directory.

ckoven:
Enables land model to write multiple monthly values in a single history file.

Fixes #3034
[BFB]

### DIFF
--- a/components/clm/src/main/controlMod.F90
+++ b/components/clm/src/main/controlMod.F90
@@ -334,9 +334,7 @@ contains
        ! History and restart files
 
        do i = 1, max_tapes
-          if (hist_nhtfrq(i) == 0) then
-             hist_mfilt(i) = 1
-          else if (hist_nhtfrq(i) < 0) then
+          if (hist_nhtfrq(i) < 0) then
              hist_nhtfrq(i) = nint(-hist_nhtfrq(i)*SHR_CONST_CDAY/(24._r8*dtime))
           endif
        end do
--- a/components/clm/src/main/histFileMod.F90
+++ b/components/clm/src/main/histFileMod.F90
@@ -66,7 +66,7 @@ module histFileMod
   integer, public :: &
        hist_ndens(max_tapes) = 2         ! namelist: output density of netcdf history files
   integer, public :: &
-       hist_mfilt(max_tapes) = 30        ! namelist: number of time samples per tape
+       hist_mfilt(max_tapes) = (/ 1, (30, ni=2, max_tapes)/)        ! namelist: number of time samples per tape
   logical, public :: &
        hist_dov2xy(max_tapes) = (/.true.,(.true.,ni=2,max_tapes)/) ! namelist: true=> do grid averaging
   integer, public :: &


bishtgautam:
@ckoven My interpretation of the code changes is that they are allowing multiple monthly time slices to be written in a single history file (i.e. not setting `hist_mfilt(i) = 1` when `hist_nhtfrq(i) == 0`). The history file doesn't necessarily need to be a h0 file.  What do you think?

ckoven:
@bishtgautam yeah, I guess that's correct.  It isn't necessarily that the current code modifies only the h0 file write frequency, but rather any monthly history file gets its mfilt value reset to 1 in the current code.  because the h0 file defaults to monthly averaging, that one gets reset to mfilt=1 by default.

note, I should also point out for provenance tracking purposes that this is a thing that's been updated on CTSM too: https://github.com/ESCOMP/ctsm/pull/398

bishtgautam:
@ckoven Thanks for the contribution. If you can run the regression test suite to verify everything is okay, I would integrate this PR.

ckoven:
@bishtgautam ok great, will do. thanks.

rljacob:
@ckoven any update on the testing?

ckoven:
running these tests now, apologies for the delay.

ckoven:
Hi All,
I ran a pair of e3sm_land_developer tests on cori-haswell. Results are:

Baseline (b932309) test resuilts:
```
./cs.status.Eb93230902-F82498d5c-fates-api-7-v9 | grep FAIL
  ERS_Ld3.f45_f45.ICLM45ED.cori-haswell_intel.clm-fates (Overall: FAIL) details:
    FAIL ERS_Ld3.f45_f45.ICLM45ED.cori-haswell_intel.clm-fates RUN time=40
```

Modified code (544db3b) test results:
```
./cs.status.E544db3bc6-F82498d5c-fates-api-7-v9 | grep FAIL
    FAIL ERS.f19_f19.ICLM45.cori-haswell_intel TPUTCOMP Error: Computation time increase > 10 pct from baseline
    FAIL ERS.f19_g16.I1850CLM45.cori-haswell_intel.clm-betr TPUTCOMP Error: Computation time increase > 10 pct from baseline
  ERS_Ld3.f45_f45.ICLM45ED.cori-haswell_intel.clm-fates (Overall: FAIL) details:
    FAIL ERS_Ld3.f45_f45.ICLM45ED.cori-haswell_intel.clm-fates RUN time=50
```

The fail in both baseline and test for FATES is because I had an interim fates version loaded that allowed seed rain but didn't account for that in the carbon balance.  I say that based on loking at the fail log, which shows the same value in the seed flux as in the carbon balance error:
```
7:  carbon balance error detected
7:  error fraction relative to biomass stock:  5.022841113191215E-002
7:  call index:            1
7:  flux in (npp):    0.000000000000000E+000
7:  flux out (fragmentation/harvest):   0.000000000000000E+000
7:  net:   0.000000000000000E+000
7:  dstock:    15.3424658187721     
7:  error=net_flux-dstock:   15.3424658187721     
7:  biomass   289.926829255682     
7:  litter  0.184639721141678     
7:  seeds   15.3424658187448     
7:  previous total   290.111468976796     
7:  lat lon  -90.0000000000000        325.000000000000     
```

For the increase in CPU times between the baseline and the test, I have no idea.  I'd guess its just noise in the runtimes because I don't see how the code would change the runtimes, but not really sure.

rljacob:
cori has a lot of variability.   You're saying we can ignore the FATES fail?

ckoven:
Yes, the code changes are orthogonal to FATES/big leaf, so we can ignore the baseline FATES fail here.

bishtgautam:
@ckoven Thanks for running the tests. I will update the first comment to be in the E3SM style(*), add BFB, and merge this PR to next.

*The entire first comment of the PR gets copied/pasted during merge commits.

ckoven:
Thanks @bishtgautam!  Do I need to do anything else?

bishtgautam:
@ckoven, Nope, you are all set.

PR #2883, opened 2019-04-25
Fixes conservation problem for non-water species

beharrop:
All non-water tracers are kept as dry mixing ratios, but CLUBB and the
gravity wave parameterizations assume wet mixing ratios, and break
conservation when they are called. This fix converts dry mixing ratios
to wet mixing ratios prior to calling the routines that need wet mixing
ratios. The code fixes follow similar fixes from CESM2, with additional
modifications for E3SM-specific components.

Fixes #2704

[Non-BFB] - Non Bit-For-Bit
[CC] - Climate Changing

PR #3003, opened 2019-06-18
Modifications to preserve the vertical shape of CO2 tracers

beharrop:
Follows suggestions by Keith Lindsay for modifying the fix
to conserve non-water tracer species (#2883 ) so that the 
nearly constant profile of CO2 is preserved.

Non-BFB when CO2 tracers are active.

[BFB]

### DIFF
--- a/components/cam/src/physics/cam/clubb_intr.F90
+++ b/components/cam/src/physics/cam/clubb_intr.F90
@@ -932,6 +932,7 @@ end subroutine clubb_init_cnst
 
    use ppgrid,         only: pver, pverp, pcols
    use constituents,   only: cnst_get_ind, cnst_type
+   use co2_cycle,      only: co2_cycle_set_cnst_type
    use camsrfexch,     only: cam_in_t
    use ref_pres,       only: top_lev => trop_cloud_top_lev  
    use time_manager,   only: is_first_step   
@@ -1181,6 +1182,7 @@ end subroutine clubb_init_cnst
    type(pdf_parameter), dimension(pverp) :: pdf_params                  ! PDF parameters                    [units vary]
    character(len=200)                    :: temp1, sub                  ! Strings needed for CLUBB output
    logical                               :: l_Lscale_plume_centered, l_use_ice_latent
+   character(len=3), dimension(pcnst)    :: cnst_type_loc               ! local override option for constituents cnst_type
 
 
    ! --------------- !
@@ -1302,7 +1304,10 @@ end subroutine clubb_init_cnst
    call physics_state_copy(state,state1)
 
    ! constituents are all treated as wet mmr by clubb
-   call set_dry_to_wet(state1)
+   ! don't convert co2 tracers to wet mixing ratios
+   cnst_type_loc(:) = cnst_type(:)
+   call co2_cycle_set_cnst_type(cnst_type_loc, 'wet')
+   call set_dry_to_wet(state1, cnst_type_loc)
 
    if (micro_do_icesupersat) then
      naai_idx      = pbuf_get_index('NAAI')
@@ -2665,6 +2670,7 @@ end subroutine clubb_init_cnst
     use physconst,              only: gravit, zvir, latvap
     use ppgrid,                 only: pver, pcols
     use constituents,           only: pcnst, cnst_get_ind, cnst_type
+    use co2_cycle,              only: co2_cycle_set_cnst_type
     use camsrfexch,             only: cam_in_t
     
     implicit none
@@ -2708,6 +2714,8 @@ end subroutine clubb_init_cnst
     
     logical  :: lq(pcnst)
 
+    character(len=3), dimension(pcnst) :: cnst_type_loc         ! local override option for constituents cnst_type
+
 #endif
     obklen(pcols) = 0.0_r8
     ustar(pcols)  = 0.0_r8
@@ -2718,7 +2726,10 @@ end subroutine clubb_init_cnst
     ! ----------------------- !
 
     ! Assume 'wet' mixing ratios in surface diffusion code.
-    call set_dry_to_wet(state)
+    ! don't convert co2 tracers to wet mixing ratios
+    cnst_type_loc(:) = cnst_type(:)
+    call co2_cycle_set_cnst_type(cnst_type_loc, 'wet')
+    call set_dry_to_wet(state, cnst_type_loc)
 
     call cnst_get_ind('Q',ixq)
     if (use_sgv) then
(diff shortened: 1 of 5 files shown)


beharrop:
As noted in the discussion for #2883 (https://github.com/E3SM-Project/E3SM/pull/2883#issuecomment-492329352) it was pointed out to us by Keith Lindsay at NCAR that the bug fix that conserves mass of non-water tracer species no longer preserves the vertical shape of the CO2 profile.  K. Lindsay suggested an additional fix that is applied only to the CO2 tracers.  As a result, this PR will only be non-BFB when the CO2 tracers are active.

singhbalwinder:
Thanks @beharrop . Would you please mention if you have done any kind of testing? I think CO2 tracer is turned on in BGC compsets by default. If you can check if the code works with non-BGC (e.g. FC5AV1C-L) compsets and one of the BGC compsets, it will be helpful. A 5 day smoke test would be enough.

beharrop:
Hi @singhbalwinder , I tested this by running two 5 timestep simulations, one with master (hash a7eb1ee9c507ed5a03ef68f4d95c2ce5bc9fc766) and then the same code base with these new code changes.  Both used the FC5AV1C-04P2 compset and the results were BFB identical.  I also copied these codes to main-1.1 to test with the 1850_CAM5%CMIP6_CLM45%CNPECACNTBC_MPASCICE%SPUNUP_MPASO%SPUNUP_MOSART_SGLC_SWAV_BGC%BCRC compset.  CO2 tracers were active, but since they weren't in prognostic mode, they didn't feedback to the simulation and the energy readouts in the atm.log files were BFB identical.

Are these tests sufficient or would you like more testing done?

singhbalwinder:
That's great! Thanks @beharrop ! I think it is enough. Does that mean this PR is BFB even for BGC cases? If CO2 is not interacting with the model's state, I think it won't change the answers for any model configuration, right?

beharrop:
@singhbalwinder , I haven't tested this, but it should be non-BFB for prognostic CO2 (BPRP) compsets where the radiative heating will see the CO2 tracers.  In any situation where the CO2 tracers are either inactive or passive, then this PR should not be answer changing (consistent with the tests I have done so far).

Since prognostic CO2 will be an important component of the CBGCv2 campaign, I figured the non-BFB flag was appropriate, but if you feel otherwise I have no objections to changing it.

rljacob:
Don't we have a compset with prognostic CO2 that would notice this change?

beharrop:
Yes, the BPRP compset uses prognostic CO2 and should have non-BFB answers.

singhbalwinder:
@beharrop : Our criteria for labeling a PR non-BFB is that it has to break at least one test in our testing suite. I see the following BGC tests in out tests:
```
SMS.ne30_oECv3.BGCEXP_BCRC_CNPECACNT_1850.clm-bgcexp
SMS.ne30_oECv3.BGCEXP_BCRC_CNPRDCTC_1850.clm-bgcexp
```
They both use BCRC so won't be non-BFB. Unless I am missing something, I didn't see any BPRP test in our testing suite (https://github.com/E3SM-Project/E3SM/blob/master/cime/config/e3sm/tests.py).

beharrop:
Thanks for the clarification, @singhbalwinder . I don't see any prognostic CO2 tests in the suite either.  I don't know if there are plans to add prognostic CO2 tests in the future (@susburrows or @kvcalvin might know), but at this point I don't have any reason to expect this PR to break any of the tests in the current suite.

rljacob:
Do you mean "BDRD"?  I can't find any compset with "BPRP" in its short or long name.

beharrop:
@susburrows has a branch with prognostic CO2, which is given the BPRP moniker.  I didn't realize that it was never merged into any other branches.  The prognostic CO2 compset will be coming, but since it isn't available in master or maint-1.1 yet, this PR is currently only fixing the diagnostic output CO2 fields for the BCRC, BCRD, BDRC, and BDRD experiments (which don't influence the simulation).

I've switched the tag to BFB for this PR.

kvcalvin:
I think we will want a BPRP test for v2, but we can wait to add it until everything is working

susburrows:
I implemented a quick version of BPRP on a branch so that Bryce could get started with testing, but @maltrud wanted to review and/or make a fresh implementation of the way the coupling is done on the ocean side, so it has not been merged to master yet.

@maltrud , would you be able to take a look at the prognostic CO2 coupling (BPRP compset) as well, while you are working on the 1% case? (I think it touches the same parts of the code)

rljacob:
@maltrud what exactly is this PR waiting for?

rljacob:
@susburrows can you make a PR to add your BPRP test?  Does this PR need something from @maltrud ?

maltrud:
@rljacob and @susburrows i'm not sure why i'm on this--all of the changes appear to be cam files.

rljacob:
@maltrud, this needs BPRP for testing and you are apparently involved in that from this comment https://github.com/E3SM-Project/E3SM/pull/3003#issuecomment-503686249

maltrud:
there is no official BPRP compset yet and this PR doesn't need it.  BPRP will be added in the future so any modifications necessary to make it compatible with MPAS will be done at that time.

wlin7:
The added logics for preserving dry mmr for CO2 tracers are all robust. @beharrop , there is also a call to set_dry_to_wet in physpkg.F90 near the end of tphysac, Should the same logics be applied?

beharrop:
@wlin7  I don't believe the same logic is needed in tphysac for the set_dry_to_wet call because that is only being done in preparation to hand off information to the dycore, while this PR and #2883 are only aimed at fixing issues in the physics parameterizations.  I will add that in the tests I did, there did not seem to be any conservation or shape-preserving issues associated with the dycore, so I think we are safe leaving tphysac as it is.  Thanks for checking on this.

wlin7:
Sounds good, @beharrop . No need to do that given the fix is solely concerning physics parameterization.

rljacob:
@singhbalwinder please start merging this.

singhbalwinder:
Pushed to next

singhbalwinder:
@beharrop : This PR produced DIFFs for the following two tests on next:
```
SMS.ne30_oECv3.BGCEXP_BCRC_CNPECACNT_1850.sandiatoss3_intel.clm-bgcexp
SMS.ne30_oECv3.BGCEXP_BCRC_CNPRDCTC_1850.sandiatoss3_intel.clm-bgcexp
```
I confirmed it on Comly and found the following variables which differed:
`a2x_Sa_co2prog, x2l_Sa_co2prog and x2oacc_Sa_co2prog`

Do you expect these differences? If yes, I will label this PR NBFB and ask for blessing these tests.

beharrop:
Hi @singhbalwinder ,
Those fields are the coupler fields related to prognostic CO2.  Since the CO2 tracers are changing from this commit, then these field changes are expected.

Apparently, these fields are being set even when prognostic CO2 is not being used.  There is logic in the import/export routines for the land and ocean that only read these variables if prognostic CO2 is being used, so the rest of the simulation should be unaffected by these changes (in agreement with the global energy sums remaining unchanged in the atmosphere).

PR #3011, opened 2019-06-25
Change atm input files to netcdf3 format.

rljacob:
Change atm input files from netcdf4 to netcdf3 format.

This affects CMIP6 cases and a couple of SCAM cases.

[BFB]


rljacob:
See discussion about NetCDF4 issues here:
https://acme-climate.atlassian.net/wiki/spaces/EIDMG/pages/769130507/Picking+a+netcdf+type+for+all+input+files

List of files converted:  https://gist.github.com/rljacob/adacea651891d2532c78dcd36a17443b.   Looks like the files we actually use are only in the atmosphere.

rljacob:
Closing this.  It should be done in maint-1.0.

PR #3017, opened 2019-06-25
DSS omega in SL transport; fix a unit test; fix Intel 18 build.

ambrad:
1. Omega is needed for diagnostics. DSS it in SL transport, just like it is DSSed in Eulerian transport.
2. An SL unit test had an uninitialized variable. Fix it.
3. An unused feature in SL transport had a bug. Fix it and the test.
4. Intel 18 with -O3 -fp-model-fast has a problem with a vectorization pragma. Comment it out. Comment out a similar one for safety.

[BFB] except on Anvil with Intel 18, where it solves a BFB problem.

### DIFF
--- a/components/cam/bld/namelist_files/namelist_definition.xml
+++ b/components/cam/bld/namelist_files/namelist_definition.xml
@@ -5891,7 +5891,7 @@ Default in homme: 0 (does not use time split wrt HV).
 </entry>
 
 <entry id="hypervis_subcycle_q" type="integer" category="se"
-       group="ctl_nl" valid_values="1" >
+       group="ctl_nl" valid_values="" >
 Number of hyperviscosity subcycles done in tracer advection code.
 The only supported value in CAM is 1.
 Default: Set by build-namelist.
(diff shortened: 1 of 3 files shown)


oksanaguba:
It is prob. ok on anvil, too. I run on anvil, default for e3sm is intel 17.

mt5555:
merged to next

rljacob:
There is a diff on sandiatoss3 that we think is from this PR:
SMS.ne4_ne4.FC5AV1C-L.sandiatoss3_intel.cam-thetahy_sl 
https://my.cdash.org/testDetails.php?test=48201170&build=1670233

Can you confirm?

mt5555:
confirmed.  This failure of this test and other tests now also confirm that OMEGA (vertical velocity of pressure surfaces) is now used by the physics for more than diagnostics and changes to OMEGA to impact the solution.

rljacob:
It was unclear if OMEGA was used in the prognostic equations?

mt5555:
no, it was unclear if OMEGA is used by the physics, other than for diagnostics.

PR #3018, opened 2019-06-26
Update mpas-source: ocean advection CPU optimization

mark-petersen:
ocean tracer advection CPU optimizations, including:
  - moving options to module private variable to reduce config calls
  - added contiguous keyword to pointers
  - passed pointers to automatic arrays
  - replaced diagnostic string argument with logical flag
  - replaced scratch fields with local thread-shared allocatable arrays
  - changed loop order in provisional thickness calc to reduce flops
  - introduced temps in horz flux calculation to reduce flops
  - moved min, max calculations to separate loops (both vert, horz)
  - modified high-order vert flux to move conditionals outside loops
  - in-lined high-order vert flux functions
  - reordered a few other calculations for efficiency
  - general cleanup to remove stale variables, reduce long lines, etc

Changes are not b4b, but are at round-off.

Tested mostly with a QU240 test case both after a single step (to confirm roundoff) and in 2-day runs. Most tests on Cori. Changes resulted in 2-3x speedup for tracer adv timers, depending on configuration.

[non-BFB]


mark-petersen:
Passes:
```
PASS PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-knl_intel COMPARE_base_single_thread
PASS PEM_Ln9.ne30_oECv3wLI.A_WCYCL1850-DIB-ISMF_CMIP6.cori-knl_intel COMPARE_base_modpes
```
see https://github.com/MPAS-Dev/MPAS-Model/pull/268 for original discussion. 

On MPAS-Ocean nightly regression suite, errors after ~5 steps are all 1e-13 or less in average difference, 1e-10 or less in max difference. I don't see any cause for concern here.

jonbob:
@mark-petersen - can you move this to e3sm-develop?

mark-petersen:
@jonbob I copied it to E3SM-Project, I think that's what you meant. Sorry, I should have done that originally. This PR is tied to my fork, so if it needs any extra commits, I'll add them to both.

jonbob:
Thanks @mark-petersen - can you make the PR from E3SM-Project? Or do I need to use your fork? With the submodules it always seems a little odd...

jonbob:
@mark-petersen - never mind, I'll just add a remote

jonbob:
merged to next

jonbob:
merged to master and all related testing DIFFs blessed

jonbob:
@mark-petersen - you can delete this branch on your fork now

PR #3019, opened 2019-06-26
Replaces mkatmsrffile tool with a simpler version

singhbalwinder:
mkatmsrffile tool depends upon  model compiled libraries such as PIO,
MCT, CSM_SHARE along with MPI. The code has been modified such that
these dependecies are removed. Now the code depends on just NETCDF
library. This code produces BFB results as the old code.

[BFB] - Bit-For-Bit


components/cam/tools/mkatmsrffile/Makefile
```diff
@@ -1,11 +1,33 @@
...
-MODDIR=$(LIBDIR)/include
+#-------------------------------------
+#Instructions:
+# Load a compiler and the associated 
+# netcdf library. Update variables "FC"
+# and "NETCDF_PATH" to add the compiler
+# and path to netcdf home directory
+#-------------------------------------
 
+FC := ifort
```
brhillman: You might add a check here to see if FC is already an environment variable, which would allow compiling with something other than ifort (i.e., gnu) without changing anything in the Makefile. You can accomplish this with:
```
ifeq ($(FC),$(null))
    FC = ifort
endif
```

components/cam/tools/mkatmsrffile/Makefile
```diff
@@ -1,11 +1,33 @@
...
+NETCDF_PATH := $(NETCDF_HOME)
 
-mkatmsrffile: mkatmsrffile.F90 
-	$(FC) $(FFLAGS) -I$(MODDIR) $< -o $@  $(LIBS) $(LDFLAGS) 
+EXE := exe_mkatmsrffile
```
brhillman: Kind of picky, but can we call this just "mkatmsrffile" or something similar? The rest of the tools seem to follow this convention (`gen_domain`, `cube_to_target`, etc).

components/cam/tools/mkatmsrffile/Makefile
```diff
@@ -1,11 +1,33 @@
...
 
+#---------------------------
+# USER INPUT ENDS HERE
+#---------------------------
+
+#include string
+INC = -I$(INC_NETCDF)
+LIB = -L$(LIB_NETCDF) -lnetcdf -lnetcdff
+
+FFLAGS = -O2
```
brhillman: Similar to above, maybe make this `FFLAGS += -02` to allow passing some additional flags via environment variable? Like if we wanted to add bounds checking, without editing the Makefile?

components/cam/tools/mkatmsrffile/README
```diff
@@ -1,25 +1,21 @@
...
-atmosphere through the coupler.  This is the default namelist setting
-drydep_method='xactive_lnd'.  However with modal areosols this method
-is not adequate and we must recalculate these fields in the atmosphere
-(see subroutine interp_map in mo_drydep.F90).  For unstructured grids
-it was determined to create this offline interpolation tool rather
-than generalize the subroutine interp_map.  
-
-1. Locate or create scrip mapping files for the 1x1 degree input grids and
-   the desired model output grid.
+Atmospheric drydeposition at the surface depends on certain surface properties including soil and land use properties.  In most cases, these calculations can be handled in the land model and passed to the atmosphere through the coupler.  This method is the default namelist setting drydep_method='xactive_lnd'.
```
brhillman: I don't particularly like the change to make the text all one line here. Is there a reason for this? It makes it hard to read unless you have your editor configured a certain way.

components/cam/tools/mkatmsrffile/README
```diff
@@ -1,25 +1,21 @@
...
+Atmospheric drydeposition at the surface depends on certain surface properties including soil and land use properties.  In most cases, these calculations can be handled in the land model and passed to the atmosphere through the coupler.  This method is the default namelist setting drydep_method='xactive_lnd'. 
+
+However with modal areosols, this method is not adequate, and we must recalculate these fields in the atmosphere (see subroutine interp_map in mo_drydep.F90).  
+
+For unstructured grids, it was determined to create this offline interpolation tool rather than generalize the subroutine interp_map.  
+
+
+1. Locate or create scrip mapping files for the 1x1 degree input grids and the desired model output grid.
 
 2. Run the script scrip_makemaps.csh
```
brhillman: We should remove this mention of scrip_makemaps.csh, and replace with your additional text below about ESMF_RegridWeightGen.

components/cam/tools/mkatmsrffile/README
```diff
@@ -1,25 +1,21 @@
...
+1. Locate or create scrip mapping files for the 1x1 degree input grids and the desired model output grid.
 
 2. Run the script scrip_makemaps.csh 
 https://svn-ccsm-models.cgd.ucar.edu/tools/mapping/makemaps/trunk/scrip_makemaps.csh
-   The script will create 4 mapping files, you need only the 1x1 -> model grid 
-   area average file.
+   The script will create 4 mapping files, you need only the 1x1 -> model grid area average file.
+
+   Alternatively, if you have the ESMF binaries, then you can run:
+   ESMF_RegridWeightGen --source <1x1 scrip file> --destination <original scrip file> --weight <map_1x1_to_original_method_date.nc> --method conserve
```
brhillman: The text "original scrip file" is a little misleading. And in fact I'd probably re-word this whole bullet point to something like the following:

2. Generate mapping weights from 1x1 grid to target grid. This can be done using ESMF with the following command (assuming an ESMF build is in your PATH):
```
ESMF_RegridWeightGen --source <path to 1x1 scrip file> \
                     --destination <path to target scrip file> \
                     --weight <name of output mapping file> \
                     --method conserve
```

singhbalwinder:
Thanks @brhillman . These are all good suggestions. I will fix all these in my next commit.

singhbalwinder:
@brhillman : I think I have addressed all your review comments. Please let me know if you notice anything else which needs to be modified.

brhillman:
@singhbalwinder sorry for the delay. Looks good!

wlin7:
Verified and merged to next.

wlin7:
Merged to master.

Issue #2963, opened 2019-06-03
SMS_R_Ld5.ne4_ne4.FSCM5A97 test fails with PIO2


dqwu:
CAM uses an invalid PIO file id while inquiring info about a NetCDF file.

The error message is:
`[0]  ERROR: (shr_scam_getCloseLatLonNC) ERROR from nf90_inquire`

The bug is at components/cam/src/control/ncdio_atm.F90:205
`call shr_scam_getCloseLatLon(ncid%fh,scmlat,scmlon,closelat,closelon,latidx,lonidx)`

For PIO2, the ncid returned to the user is not the actual ncid returned by the underlying netcdf/pnetcdf library. This opaque pio_ncid is unusable from outside PIO2, such as nf90_inquire calls (direct calls to the NetCDF library, bypassing PIO) in shr_scam_getCloseLatLonNC.

PR #2771 has fixed similar issues for other shr_scam_getCloseLatLon calls. However, the fixes merged in the PR were created before line 205 (with the bug) was added to ncdio_atm.F90, so the PR did not end up fixing this bug.

rljacob:
This test fails with the default PIO1 on anvil as well:
SMS_R_Ld5.ne4_ne4.FSCM5A97.anvil_intel 
https://my.cdash.org/testDetails.php?test=48150310&build=1668884

This test has either never passed on anvil or hasn't passed in a long time.

bogensch:
I found the issue.  ROOTPE needs to be explicitly set to 0 for all components in cime/scripts/lib/CIME/test_scheduler.py for PTS_MODE.  The test worked on anvil once I did this.  I will issue a cime PR to address this.

jayeshkrishna:
@bogensch : Can you also include the PIO fix mentioned in this issue in your PR (you can cherry-pick https://github.com/E3SM-Project/E3SM/commit/28ffe751dd5dfb38133864fdb3f10397d79a1276 since @dqwu has fixed this issue in our testing branch)?

bogensch:
will do

bogensch:
@jayeshkrishna are you talking about including the mod to file components/cam/src/control/ncdio_atm.F90 ?  If so, I can't include that in my cime PR since it's not in the cime/ directory

jayeshkrishna:
Ok, in that case please don't close this issue (Refer to the issue "See Issue #...") when you merge your PR. I will create a separate PR with the PIO fix and merge it (closing this issue)

[...]

PR #3024, opened 2019-06-27
Fix for failing SCM developer test on Anvil

bogensch:
The SMS_R_Ld5.ne4_ne4.FSCM5A97 test was failing on Anvil.  The reason is that ROOTPE needs to be explicitly set to zero for all component models for PTS_MODE (single column model mode).  All developer tests pass with this change. 

Addresses #2963


jgfouca:
Nice catch!

PR #3009, opened 2019-06-24
Adding restart tests for theta

oksanaguba:
Adding ERS tests for hydro/nonhydro theta dycore. 

[bfb] but will need baselines.

PR #3025, opened 2019-06-27
CICE_CPPDEFS was not being set correctly in some cases.

jgfouca:
Also, some minor fixes to CIME to catch this error sooner.

[BFB]

### DIFF
--- a/components/cice/cime_config/buildnml
+++ b/components/cice/cime_config/buildnml
@@ -70,9 +70,10 @@ def buildnml(case, caseroot, compname):
                                  from_dir=os.path.join(srcroot, "components/cice/bld")).split()
 
         if int(config[0]) >= 0:
-            case.set_value("CICE_BLCKX",         config[2])
-            case.set_value("CICE_BLCKY",         config[3])
-            case.set_value("CICE_MXBLCKS",       config[4])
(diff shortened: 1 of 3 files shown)


components/cice/cime_config/buildnml
```diff
@@ -70,9 +70,10 @@ def buildnml(case, caseroot, compname):
                                  from_dir=os.path.join(srcroot, "components/cice/bld")).split()
 
         if int(config[0]) >= 0:
-            case.set_value("CICE_BLCKX",         config[2])
-            case.set_value("CICE_BLCKY",         config[3])
-            case.set_value("CICE_MXBLCKS",       config[4])
+            cice_blckx, cice_blcky, cice_mxblcks = config[2:5]
```
jgfouca: The bug is here. The original code has the potential for case's CICE_BLCKX, CICE_BLCKY, and CICE_MXBLCKS to become out-of-sync with the local variables cice_blckx, etc which are used to define CICE_CPPDEFS below.

jgfouca:
@oksanaguba , once this PR is merged, if you rebase your #3009 PR, it should work with those ERS tests in the same shared test suite.

oksanaguba:
@jgfouca thanks -- I will make a note about this.

rljacob:
You can merge this.

PR #3006, opened 2019-06-18
Update mpas-source to current ocean-BGC submodule.

mark-petersen:
This updates the ocean-BGC submodule to python 3 compliance (https://github.com/MPAS-Dev/MPAS-Model/pull/274).

[BFB]

PR #3026, opened 2019-06-27
Update mpas-source to use short BGC tag

### LINKED ISSUE #3022: get_BGC.sh no longer working
The script get_BGC.sh (in components/mpas-source/src/core_ocean) is used to download the BGC code so that git passwords are not requested in the middle of a compile (when you are not paying attention and forget that this may be needed). There is also a script get_cvmix.sh, for cvmix source, which still works.

When working correctly, the first time you execute get_BGC.sh after a new E3SM clone or after a new version of MPAS-O requires an update to the BGC code, I get (on Anvil)

     $ ./get_BGC.sh
      ** Using git to acquire BGC source. ** 
     Enter passphrase for key '/home/worleyph/.ssh/id_rsa': 

After entering my passphrase, if I execute this again, I get

     $ ./get_BGC.sh
      BGC version is current. Skip update

With current master, every time I execute ./get_BGC.sh it requests a passphrase, and when trying to build E3SM, there it (also) requests a passphrase, e.g. on Compy

     Enter passphrase for key '/qfs/people/worl312/.ssh/id_rsa':

(but compilation then completes successfully).

I have observed this on both Compy and Anvil with latest master. It broke very recently.

mark-petersen:
Change BGC repository tag from long to short version.

[BFB]
Fixes #3022


mark-petersen:
This is a correction to https://github.com/E3SM-Project/E3SM/pull/3006. That one passed the nightly regression tests, so this one should too. So I added the BFB flag.

mark-petersen:
Passes
```
```

jonbob:
@mark-petersen - can you test BGC compsets instead?

mark-petersen:
Can you give me an example create_test line? I haven't run those before.

jonbob:
Sure: SMS.ne30_oECv3.BGCEXP_BCRC_CNPRDCTC_1850.sandiatoss3_intel.clm-bgcexp

jonbob:
Minus the sandiatoss part....

jonbob:
@mark-petersen - I went ahead and merged to next. It's broken as-is, so this can do no harm -- but please do try the test that uses BGC code

mark-petersen:
Well, I can see that the run downloads and compiles the BGC code, which is the test for this PR. The BGC submodule is here in the run directory:
```
bld/ocn/source/core_ocean/BGC
```
That BGC test compiled but failed to run, but I think it is an unrelated cori-knl problem, because it took 10 minutes to just fetch the repo earlier today.

jonbob:
merged to master

PR #3028, opened 2019-06-27
Adds support for co2 vertical interpolation

singhbalwinder:
The code changes introduced via  7fbec4f
facilitated horizontal interpolation but was missing the vertical
interpolation.

I have created a dummy state variable so that the interpolation
routines can read interface (or mid point) heights from the state
to enable vertical interpolation.

co2 tracer can be enabled only if we add -co2_cycle to
CAM_CONFIG_OPTS in env_build.xml file. To enable bi-linear
interpolation, we also need to add -cppdefs -DCO2_BILIN_REGRID
to CAM_CONFIG_OPTS in env_build.xml file.

[BFB] - Bit-For-Bit

### DIFF
--- a/components/cam/src/chemistry/utils/tracer_data.F90
+++ b/components/cam/src/chemistry/utils/tracer_data.F90
@@ -611,7 +611,7 @@ contains
 
     type(trfile),        intent(inout) :: file
     type(trfld),         intent(inout) :: flds(:)
-    type(physics_state), optional, intent(in)    :: state(begchunk:endchunk)
+    type(physics_state), intent(in)    :: state(begchunk:endchunk)
     
     type(physics_buffer_desc), optional, pointer :: pbuf2d(:,:)
     integer :: ncol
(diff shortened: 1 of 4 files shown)


beharrop:
Thanks @singhbalwinder for working on this.  I have a question about your implementation.  Does using state_at_init in the advance_trcdata call mean that we are using the pressure levels at initialization to do the interpolation?  Maybe I am missing something, but I would have thought at each time step, we would want to use the current pressure levels to read in and interpolate the next emission entry.

singhbalwinder:
You are right. We need the current state vector to do the interpolation. I will look into the code to see if we can send in the current state vector and modify the code accordingly.

beharrop:
The new commits to handle the grids and error checking look OK to me, @singhbalwinder

rljacob:
@singhbalwinder is this ready?

singhbalwinder:
We are still working on testing an alternative algorithm for interpolation. it may take some time to fully implement it and test it in various scenarios. This PR will have no impact on the WCYCL experiments. It only affects the BGC experiments.

rljacob:
Any further progress on this PR?  I know its not v2 but it is almost 2 months old.

rljacob:
@singhbalwinder what is up with this old PR?

singhbalwinder:
Closing this PR util it is ready.

PR #2933, opened 2019-05-17
Infrastructure upgrade to support finite volume physics grid (rebased)

whannah1:
This PR updates the atmosphere grid infrastructure to allow for a finite volume physics grid (hereafter referred to as “physgrid”) as an alternative to the traditional approach of having physics columns coincide with the GLL nodes of the dynamics grid. Although this requires many large code changes, (mostly in dyn_grid.F90) the model is still BFB when run with a typical configuration.

To make sure ongoing development does not break the coupling of the FV physgrid, I also added a new test "SMS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L" to the e3sm_atm_developer suite. 

[BFB]

PR #3030, opened 2019-06-28
EAM: enable output of vorticity and divergence 

mt5555:
enable outfld() calls for vorticity and divergence

add code to dp coupling layer to compute vorticity and divergence from u,v, and then DSS the results using the make_c0() routine.  

update all DSS calls in dp coupling layer and make_c0() routine to use
nlyr pack/unpack interface and have all routines use single edge_g buffer.
This reduces memory footprint and removes many edge buffer allocate/deallocate statements

[BFB]

### DIFF
--- a/components/cam/src/dynamics/se/dp_coupling.F90
+++ b/components/cam/src/dynamics/se/dp_coupling.F90
@@ -35,7 +35,6 @@ CONTAINS
     use physics_buffer,          only: physics_buffer_desc, pbuf_get_chunk, pbuf_get_field
     use shr_vmath_mod,           only: shr_vmath_exp
     use time_manager,            only: is_first_step
-    use viscosity_mod,           only: compute_zeta_C0
     use cam_abortutils,          only: endrun
     use gravity_waves_sources,   only: gws_src_fnct
     use dyn_comp,                only: frontgf_idx, frontga_idx, hvcoord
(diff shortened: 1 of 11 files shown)


mt5555:
will most likely create conflicts with PR #2933 
this should be merged after PR #2933

mt5555:
the ability to output vorticity and divergence is useful for tropical cyclone / storm detection and for dycore evalation.

mt5555:
@bogensch : do you know why calls to outfld() of some dycore variables in the dp coupling layer are disabled when running single column?

bogensch:
@mt5555 .  When constructing the SE SCM in E3SM we found we could not simply initialize the dycore with just one gauss lobatto point.  Thus, we tricked the model and told it that it only has one physics column, but initialize the full dynamics grid, even though only one column is active.  Since all outfld calls are done on the physics grid, this created an error/conflict when passing the full array.  I suppose it would work by just passing it the value for the single active column, but I imagine that output would be pretty much garbage anyway for the SCM for vorticity and divergence.

mt5555:
thanks -makes sense.

oksanaguba:
How to use this output? thanks!

mt5555:
it adds to new variables for possible output, VOR and DIV, which you would add to your fincl= line in user_nl_cam.

mt5555:
rebaselined after physgrid code went in. @whannah1 : I mucked around with a little of your code, can you take a quick look at the diff's for this PR?

mt5555:
Also, as physics only ever returns 2 tendencies for momentrum, any objections if I remove the num_FM_vars() variable?

whannah1:
@mt5555 I think I added that for the physgrid to simplify the code where the Oksana had added ifdefs for the theta dycor option.

whannah1:
I'm a little confused, why did Oksana need 3 FM vars when MODEL_THETA_L is defined?

PR #3032, opened 2019-06-28
Update MALI to use new Albany basal friction law

matthewhoffman:
This merge updates MPAS source code to use the updated basal friction
law from Albany.  This is a Schoof-type basal friction law that accounts
for subglacial water pressure.

[BFB] for compsets without active ice sheet
[Non-BFB] for compsets with active MALI component
[NML]

### DIFF
--- a/cime/config/e3sm/machines/config_compilers.xml
+++ b/cime/config/e3sm/machines/config_compilers.xml
@@ -1097,6 +1097,7 @@ for mct, etc.
 </compiler>
 
 <compiler MACH="cori-haswell" COMPILER="intel">
+  <ALBANY_PATH>/global/project/projectdirs/acme/software/AlbanyTrilinos20190823/albany-build/install</ALBANY_PATH>
   <CONFIG_ARGS>
     <base> --host=Linux </base>
   </CONFIG_ARGS>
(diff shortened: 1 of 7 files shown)


rljacob:
Is this really BFB?  even with an ice-sheet test?

matthewhoffman:
@rljacob , no, this is only BFB for "standard" compsets that do not include active ice sheets.  I've updated the description to be more explicit - it wasn't clear to me how to handle this sort of situation.

rljacob:
If its non-BFB for any test or case, I like to label the PR as non-BFB but you can add a comment with details as you did above.  When we later search for non-BFB PRs, its good to have ones with _any_ change included.

rljacob:
Is this a v2 feature?  If so, it needs a code review document on confluence.

jonbob:
@matthewhoffman - I think this PR will require a code review document

jonbob:
A MALI test is included in the e3sm_integration suite, which is tested for baselines on sandiatoss3 and for machine coverage on anvil, compy, bebop and theta. Note that the MALI test currently passes only on sandiatoss3 and bebop. This PR updates the version of MALI on sandiatoss3 and cori, so for purposes of testing I've focused on cori. A test merge shows these results:
* SMS.f09_g16_a.MALI.cori-knl_intel    PASS
* ERS.f09_g16_g.MALISIA.cori-knl_intel    PASS
* SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.cori-knl_intel    PASS
* SMS.f09_g16_a.MALI.cori-knl_gnu   FAIL (no library for gnu)

@rljacob - we would still like to merge this PR today. MALI is not part of any upcoming simulation plans, so its absence on some platforms will not hold up progress.

rljacob:
Go ahead.

jonbob:
merged to next

jonbob:
merged to master and expected NML DIFF's blessed

PR #2789, opened 2019-03-08
Implements remaining changes for gustiness feature for EAMv2

singhbalwinder:
This PR completes the gustiness feature in EAM by implementing all the right namelist settings for gustiness. A new use case file `2000_cam5_av1c-04p2_gust.xml` has all the namelist values for invoking this feature using `FC5AV1C-L` compset. Several new namelist variables required for this feature will now appear in the namelist (`atm_in`).

A test mod directory is also created to add a test for this feature in the nightly testing.

[BFB]
[NML]

PR #3035, opened 2019-06-28
discontinue support for Eulerian dycore for the SCM (atm)

bogensch:
This PR will disable the Eulerian dycore for running the single column model (SCM).  The SCM will now only work (and be supported) with the Spectral Element (SE) dycore, which has now been well tested/validated, and is consistent with EAM global simulations.  

With this PR an error and abort message will be provided to the user should they try to configure the SCM with any dycore other than SE.  In addition, hooks for the Eulerian core are removed from the scam_arm97.xml use_case, as this is the compset used for SCM developer tests.  

Note that the SCM developer test for Eulerian SCM is removed in a companion cime PR.  

All developer tests pass, except SMS_R_Ld5.T42_T42.FSCM5A97, which is an expected fail.

### DIFF
--- a/components/cam/bld/configure
+++ b/components/cam/bld/configure
@@ -1211,10 +1211,10 @@ if (defined $opts{'scam'}) {
 }
 my $scam = $cfg_ref->get('scam') ? "ON" : "OFF";
 
-# Currently only Eulerian and Spectral Element dycores support SCM
-if ($scam eq 'ON' and $dyn_pkg ne 'eul' and $dyn_pkg ne 'se') {
+# Currently only Spectral Element dycore supports SCM
+if ($scam eq 'ON' and $dyn_pkg ne 'se') {
     die <<"EOF";
-**  ERROR: SCAM mode only works with Eulerian dycore or Spectral Element dycore.
+**  ERROR: The Single Column Model only works with Spectral Element dycore.
 **         Requested dycore is: $dyn_pkg
 EOF
 }
--- a/components/cam/bld/namelist_files/use_cases/scam_arm97.xml
+++ b/components/cam/bld/namelist_files/use_cases/scam_arm97.xml
@@ -3,8 +3,7 @@
 
 <namelist_defaults>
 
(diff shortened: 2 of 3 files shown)


wlin7:
Hi @bogensch , nice to see we can finally say goodbye to the old EUL-based SCM framework. Thanks for the hard work. All the changes to the two files look good. I think the user_case file scam_generic.xml also needs to be updated because all the cases in the SCM case library uses compset F_SCAM5 while uses this generic file.

Also this just remind me of another issue on parameter settings. After PR #2789, two  pairs of clubb parameters become fully decoupled. clubb_C1b used to take the value clubb_C1 if clubb_C1b is not set in namelist. Similarly for clubb_gamma_coef and clubb_gamma_coefb. 

To conform to EAMv1 LR configuration, in SCM run scripts, clubb_C1b and clubb_gamma_coefb now need to be explicitly set to the same values as clubb_C1 and clubb_gamma_coef, respectively. These changes are done for existing use_case files for full model in PR #2789, but not for the SCM scripts.

bogensch:
Thanks @wlin7 for your careful review.  I updated scam_generic.xml to reflect only SE SCM.  

Also, thanks for reminding me about the two CLUBB parameters.  These have been updated on the SCM wiki in the scripts.

wlin7:
Merged to next.

wlin7:
Merged to master.

components/cam/bld/configure
```diff
@@ -1211,10 +1211,10 @@ if (defined $opts{'scam'}) {
 }
 my $scam = $cfg_ref->get('scam') ? "ON" : "OFF";
 
-# Currently only Eulerian and Spectral Element dycores support SCM
-if ($scam eq 'ON' and $dyn_pkg ne 'eul' and $dyn_pkg ne 'se') {
+# Currently only Spectral Element dycore supports SCM
+if ($scam eq 'ON' and $dyn_pkg ne 'se') {
     die <<"EOF";
-**  ERROR: SCAM mode only works with Eulerian dycore or Spectral Element dycore.
+**  ERROR: The Single Column Model only works with Spectral Element dycore.
```
rljacob: I'm late but should add "EAM" before "Single Column Model" in this error message.  I believe the CAM one still supports eulerian?

PR #3037, opened 2019-07-01
Enable archiving of performance data for new NERSC allocations

worleyph:
m3411 and m3412 are new ALCC allocations at NERSC associated with the E3SM
project. Performance archiving is enabled, with performance data
sent to the usual central /project/projectdirs/acme/performance_archive location.

The now terminated ALCC allocations m2830 and m2833 also have
performance archiving disabled.

BFB


worleyph:
Note that I have not tried testing this yet, and won't be able to until I receive access to the m3411 allocation. (@kvcalvin .)

worleyph:
@noel, my tests all worked fine - specifying acme, m3411, and m3412 all archived performance data in the desired location and the correct project was used for each test job. Please integrate soon, so that performance data for these new allocations are captured. Thanks.

ndkeen:
merged to next

PR #3038, opened 2019-07-01
Add diagnostic output for hub height winds

beharrop:
Added an option for outputting U and V at a fixed height above the
surface (90m for wind turbine hub height). New variables are called
"U90M" and "V90M". Since the vertinterp scheme only works with pressure,
a copy was made (vertinterpz) for interpolating with height coordinates.
Offline tests with a snapshot of wind data suggested linear
interpolation was slightly better than logarithmic interpolation for
interpolating low-level winds, so linear interpolation is used in
"vertinterpz".

[BFB]

### DIFF
--- a/components/cam/src/control/interpolate_data.F90
+++ b/components/cam/src/control/interpolate_data.F90
@@ -1030,7 +1030,7 @@ contains
     !
 
     k = nlev-1
-    do while ( (any(found(:) == .false.) == .true.) .and. k>=1 )
+    do while ( (any(found(:) .eqv. .false.) .eqv. .true.) .and. k>=1 )
        do i=1,ncol
           if ((.not. found(i)) .and. zmid(i,k)>zout .and. zout>=zmid(i,k+1)) then
              found(i) = .true.
--- a/components/cam/src/control/interpolate_data.F90
+++ b/components/cam/src/control/interpolate_data.F90
@@ -1029,14 +1029,27 @@ contains
     ! If all indices for this level have been found,
     ! do the interpolation
     !
-    do k=1,nlev-1
+    !do k=1,nlev-1
+    !   do i=1,ncol
+    !      if ((.not. found(i)) .and. zmid(i,k)>zout .and. zout>=zmid(i,k+1)) then
+    !         found(i) = .true.
+    !         kupper(i) = k
+    !      end if
+    !   end do
+    !end do
+
+    k = nlev-1
+    do while ( (any(found(:) == .false.) == .true.) .and. k>=1 )
        do i=1,ncol
           if ((.not. found(i)) .and. zmid(i,k)>zout .and. zout>=zmid(i,k+1)) then
              found(i) = .true.
              kupper(i) = k
           end if
        end do
+       k = k-1
     end do
+    
+    
     !
     ! If we've fallen through the k=1,nlev-1 loop, we cannot interpolate and
     ! must extrapolate from the bottom or top data level for at least some
--- a/components/cam/src/control/interpolate_data.F90
+++ b/components/cam/src/control/interpolate_data.F90
@@ -1026,17 +1026,8 @@ contains
     error = .false.
     !
     ! Store level indices for interpolation.
-    ! If all indices for this level have been found,
-    ! do the interpolation
+    ! Once all indices have been found, do the interpolation
     !
-    !do k=1,nlev-1
-    !   do i=1,ncol
-    !      if ((.not. found(i)) .and. zmid(i,k)>zout .and. zout>=zmid(i,k+1)) then
-    !         found(i) = .true.
-    !         kupper(i) = k
-    !      end if
-    !   end do
-    !end do
 
     k = nlev-1
     do while ( (any(found(:) == .false.) == .true.) .and. k>=1 )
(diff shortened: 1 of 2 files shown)


beharrop:
@kvcalvin I assumed 90m would be a good hub height based on your range of 80-100m suggested previously.  If you want a different value to be used, please let me know.

components/cam/src/control/interpolate_data.F90
```diff
@@ -977,6 +977,96 @@ subroutine vertinterp(ncol, ncold, nlev, pmid, pout, arrin, arrout)
...
+    !
+    ! Method:  Copies vertinterp and shifts its logic so height decreases 
+    !          with increasing level index.
+    !
+    ! Author:  Bryce Harrop
+    ! Note:    A more elegant method would be to introduce controlling logic
+    !          to vertinterp to get rid of the need for two separate 
+    !          subroutines.
+    !          Simple tests were done for linear and logarithmic 
+    !          interpolation, and the results suggested linear is better.
```
susburrows: Suggest updating this comment to be more precise: "linear is better for interpolation of low-level winds"

wlin7:
Hi @beharrop , the codes (specifically the new function vertinterpz) look good and can certainly good. I just wonder if we really want a generic version to interpolate to any specified altitude. Most likely we may only want such interpolated fields at certain low altitude level. For this consideration, the codes would be more efficient if the order of k-loop is reversed. Another consideration to make the code even more efficient is to change to using do-while loop, and let the loop immediately stop once the kupper is located.

kvcalvin:
@beharrop 90m seems like a good value.

rljacob:
@beharrop please make the changes Susannah requested.

beharrop:
@rljacob , I have made the changes @susburrows requested.

I have not yet made the changes suggested by @wlin7 since those would require additional testing and validation.

beharrop:
@wlin7 , I have made your suggested modifications.  Please let me know if there are any remaining issues you would like me to address.

rljacob:
@wlin7 please answer the question above.
@susburrows please re-review.

susburrows:
Looks good.  Thanks @beharrop .

rljacob:
@wlin7 is this ready?

wlin7:
@beharrop , like the way you coded the changes. All looks good to me. And sorry I missed the notification email.
@singhbalwinder , it is ready to go. Thanks.

singhbalwinder:
Why does this PR shows differences in config_machines.xml and config_compilers.xml?

beharrop:
Hi @singhbalwinder , the short answer is because I am a git novice.

The long answer is that the Cori maintenance happened after the initial PR was made but before the additional modifications suggested by @susburrows and @wlin7 .  In order to test the modifications I made, I needed to update the machine files to get the code to run again, but I didn't know how to remove them from getting added to the pull request.  Those machine file updates do not need to be here.  Sorry for the confusion.

singhbalwinder:
Thanks @beharrop. No worries. I just wanted to make sure that I am not overlooking something. I will rebase it to the latest master to remove these conflicts and then merge it.

beharrop:
Sounds good.  Thanks @singhbalwinder

singhbalwinder:
@beharrop : I have rebased this branch on top of the current master. Would you please check to see if everything looks good before I merge it? Thanks!

beharrop:
@singhbalwinder, I didn't try running the code, but it looks fine.

singhbalwinder:
Thanks @beharrop. I would feel more comfortable merging it if we at least do a smoke test. If you can list minimal configuration or namelist changes that I need to make to invoke this code, I can run a short test to ensure that it won't break anything after the merge.

beharrop:
@singhbalwinder.  The test would be to include U90M and V90M in the history output.  I went ahead and ran a test with these outputs included.  The model ran successfully, but I noticed a problem with the output: U90M and V90M are identical.  In cam_diagnostics.F90, state%u was passed to vertinterpz to compute V90M.  I have corrected this and tested it again.  The model still runs and now those two variables are not identical.  I have added a commit with this correction.

singhbalwinder:
Great! Thanks @beharrop for running the test and fixing the bug. I will merge it to next.

singhbalwinder:
Merged to next

dqwu:
This PR has caused build errors for some e3sm_developer tests:
https://my.cdash.org/viewTest.php?onlyfailed&buildid=1713720

Detailed information:
```
components/cam/src/control/interpolate_data.F90:1033:20:

     do while ( (any(found(:) == .false.) == .true.) .and. k>=1 )
                    1
Error: Logicals at (1) must be compared with .eqv. instead of ==
```

beharrop:
@dqwu , @singhbalwinder , I have changed the "==" to ".eqv." for the logicals.  I'm not sure how to run the test suite, but the model still runs fine and produces the same answers for U90M and V90M.

rljacob:
@beharrop what machine are you running on?

beharrop:
@rljacob , I did most of the initial work on Cori, but since NERSC has been down the past couple days, the recent runs I did on Compy.

rljacob:
@beharrop the machine that complained uses gnu and you were probably using intel.  gnu is available on anvil if you have an account there.  You should be able to reproduce the problem so you can verify the fix.

rljacob:
@singhbalwinder if you can't verify it works on gnu before tonight, revert the PR from next.

beharrop:
@rljacob , I technically have an account with LCRC, but I have never logged in to Anvil before, so it may take a little while for me to test this.

singhbalwinder:
I will test it and push it to next soon.

PS. I took a day off today. My responses may be slow.

beharrop:
Thanks, @singhbalwinder!  Let me know if there are any remaining issues from the test.

singhbalwinder:
pushed to next again.

rljacob:
This can be merged to master.

PR #3039, opened 2019-07-01
code refactor to remove most uses of ps_v

mt5555:
HOMME maintains two variables, ps_v (surface pressure) and dp3d ( layer pressure thickness)
previously, ps_v had to be used outside of the timestepping loop, while dp3d had to be used inside the timestepping loop and dp3d was not valid outside the timestepping loop.  This created some confusion and was error prone.   

With this change, dp3d is always valid and should be used in most places.  ps_v is still used in the dp_coupling layer (after the veritical remap) and in preqx diagnostics.  

After the EAM physics, there is a pressure adjustment to account for changes in water vapor (in order to conserve dry mass) due to our use of a total mass pressure coordiante.  this adjustment is currently done in the surface pressure, meaning that sources/sinks of water vapor are distributed uniformly throughout the coordiante.  with this change, these sources/sinks can be applied locally (but this has not been done yet since it may be climate changing).  

[BFB] except for roundoff level changes in the "baro2c" HOMME subtest.


components/homme/src/share/prim_driver_base.F90
```diff
@@ -1468,113 +1404,114 @@ subroutine applyCAMforcing_tracers(elem,hvcoord,np1,np1_qdp,dt,adjustment)
...
+                     if (elem%state%Qdp(i,j,k,q,np1_qdp) < 0 ) then
+                        v1=0  ! Q already negative, dont make it more so
+                     else
+                        v1 = -elem%state%Qdp(i,j,k,q,np1_qdp)
+                     endif
+                  endif
+                  elem%state%Qdp(i,j,k,q,np1_qdp) = elem%state%Qdp(i,j,k,q,np1_qdp)+v1
+                  if (q==1) then
+                     elem%derived%FQps(i,j)=elem%derived%FQps(i,j)+v1/dt
+                     dp_adj(i,j,k)=dp_adj(i,j,k) + v1/dt
```
AaronDonahue: Why do you normalize by dt here.  For FQps above it makes sense, since you will multiply by dt again when you update ps.  But, if "adjust_ps" is false below than dp_adj will be different but won't capture the timestep in the adjustment.

mt5555: good catch!   the adjust_ps=.false. capability will be new with this PR, and hasn't been used.  this would have been a very difficult bug to track down.

components/homme/src/preqx/share/prim_advance_mod.F90
```diff
@@ -825,9 +756,13 @@ subroutine advance_hypervis_dp(elem,hvcoord,hybrid,deriv,nt,nets,nete,dt2,eta_av
                  ! normalize so as to conserve IE
                  ! scale by 1/rho (normalized to be O(1))
                  ! dp/dn = O(ps0)*O(delta_eta) = O(ps0)/O(nlev)
+#if 0
                  dpdn(:,:) = ( hvcoord%hyai(k+1) - hvcoord%hyai(k) )*hvcoord%ps0 + &
                       ( hvcoord%hybi(k+1) - hvcoord%hybi(k) )*elem(ie)%state%ps_v(:,:,nt)
                  ttens(:,:,k,ie) = ttens(:,:,k,ie) * hvcoord%dp0(k)/dpdn(:,:)
+#else
+                 ttens(:,:,k,ie) = ttens(:,:,k,ie) * hvcoord%dp0(k)/elem(ie)%state%dp3d(:,:,k,nt)
+#endif
```
mt5555: @bartgol :  this code change I believe should also be made in preqx_kokkos, but I cant locate this code.  Can you point me to it?

bartgol: Ah! I believe we do not have that adjustment for `nu_p=0` in preqx_kokkos. I can add either add it within this PR or do it in a subsequent one.

oksanaguba:
@bartgol we would also need a test, but maybe there is no rush about it, cause nu_p is used only for 1 homme test (and templates).

bartgol:
@oksanaguba Ok, so should the kokkos code be fixed in another PR? The kokkos code is "safe" btw, in the sense that if nu_p<=0, then the setup of the `SimulationParams` object will throw an error. So there's no risk of having tests failing for unknown reasons.

mt5555:
This is good - no reason for the kokkos support to support this case (nu_p=0), and it means we dont have to update the kokkos code in this PR - thanks.

components/homme/src/share/prim_driver_base.F90
```diff
@@ -1435,7 +1457,7 @@ subroutine applyCAMforcing_tracers(elem,hvcoord,np1,np1_qdp,dt,adjustment)
                   elem%state%Qdp(i,j,k,q,np1_qdp) = elem%state%Qdp(i,j,k,q,np1_qdp)+v1
                   if (q==1) then
                      elem%derived%FQps(i,j)=elem%derived%FQps(i,j)+v1/dt
-                     dp_adj(i,j,k)=dp_adj(i,j,k) + v1/dt
+                     dp_adj(i,j,k)=dp_adj(i,j,k) + v1
```
oksanaguba: is it true that if !use_moisture and !adjust_ps, then dp3d is still adjusted for vapor?

components/homme/src/preqx/share/element_ops.F90
```diff
@@ -231,18 +222,25 @@ end subroutine copy_state
...
   implicit none
   
   type (element_t), intent(inout)   :: elem
-  real (kind=real_kind), intent(in) :: temperature(np,np,nlev)
+  real (kind=real_kind), intent(in) :: temperature(np,np,nlev),ps(np,np)
   type (hvcoord_t),     intent(in)  :: hvcoord                      ! hybrid vertical coordinate struct
-  integer :: tl
+  integer :: tl,k
 
+  real (kind=real_kind)  :: temp(np,np)
```
AaronDonahue: What is `temp` used for?

components/homme/src/preqx/share/prim_advance_mod.F90
```diff
@@ -825,9 +756,13 @@ subroutine advance_hypervis_dp(elem,hvcoord,hybrid,deriv,nt,nets,nete,dt2,eta_av
                  ! normalize so as to conserve IE
                  ! scale by 1/rho (normalized to be O(1))
                  ! dp/dn = O(ps0)*O(delta_eta) = O(ps0)/O(nlev)
+#if 0
```
AaronDonahue: Does this mean this chunk of code is slated to be removed in the future?  If so, why are we keeping it?

components/homme/src/preqx/share/prim_state_mod.F90
```diff
@@ -319,7 +317,7 @@ subroutine prim_printstate(elem, tl,hybrid,hvcoord,nets,nete)
 
     !   mass = integral( ps-p(top) )
     do ie=nets,nete
-       tmp(:,:,ie)=elem(ie)%state%ps_v(:,:,n0) 
+       tmp(:,:,ie)=hvcoord%hyai(1)*hvcoord%ps0 + sum(elem(ie)%state%dp3d(:,:,:,n0),3)
```
AaronDonahue: A general comment about all cases where surface pressure is defined in this way.  Isn't `hyai(1)` the top of the model?  Shouldn't it be `hyai(nlev)`?

components/homme/src/share/prim_driver_base.F90
```diff
@@ -1406,12 +1336,24 @@ subroutine applyCAMforcing_tracers(elem,hvcoord,np1,np1_qdp,dt,adjustment)
...
-  use control_mod,        only : use_moisture
+  ! NOTE about ps_v/dp3d
+  ! init:
+  !   (both ps_v and dp3d are valid)
+  ! do: 
+  !    physics  (uses ps_v to compute pressure levels. doesn't change ps_v)
+  !    applyCAMforcing_tracers  use ps_v for initial pressure.  
+  !                             may adjust dp3d for mass conservation (if adjust_ps=.false.)
+  !                             ps_v no longer valid
+  !    dynamaics                should only use dp3d
```
AaronDonahue: minor comment, "dynamics" is misspelled here.

components/homme/src/share/prim_driver_base.F90
```diff
@@ -1436,30 +1380,44 @@ subroutine applyCAMforcing_tracers(elem,hvcoord,np1,np1_qdp,dt,adjustment)
...
 #endif
 
+#ifdef MODEL_THETA_L
+  if (rsplit==0) then
+     adjust_ps=.true.   ! we are always on ref levels
+  else
+     adjust_ps=.true.   ! eventually change this to false
+  endif
+#else
+  adjust_ps=.true.      ! preqx doesn't use
```
AaronDonahue: if preqx "doesn't use" does that mean this should be `.false.`?

components/homme/src/share/prim_driver_base.F90
```diff
@@ -1468,113 +1426,114 @@ subroutine applyCAMforcing_tracers(elem,hvcoord,np1,np1_qdp,dt,adjustment)
                   ! dyn_in%elem(ie)%state%Qdp(i,j,k,ic,tl_fQdp) = &
                   !        dyn_in%elem(ie)%state%Qdp(i,j,k,ic,tl_fQdp) + fq 
                   elem%state%Qdp(i,j,k,ic,np1_qdp) = &
-                  dp(i,j,k)*elem%derived%FQ(i,j,k,ic)
-
+                       dp(i,j,k)*elem%derived%FQ(i,j,k,ic)
+                  
                   if (ic==1) then
```
AaronDonahue: just nitpicky, so feel free to ignore.  But isn't the usage of `ic` effectively the same as `q` in this routine?  Same for `fq` and `v1`?  Could we get away with making these the same to make the loops more readable?

components/homme/src/share/prim_driver_base.F90
```diff
@@ -1468,113 +1426,114 @@ subroutine applyCAMforcing_tracers(elem,hvcoord,np1,np1_qdp,dt,adjustment)
...
-
-     endif ! if adjustment
+                  v1 = dt*elem%derived%FQ(i,j,k,q)
+                  if (elem%state%Qdp(i,j,k,q,np1_qdp) + v1 < 0 .and. v1<0) then
+                     if (elem%state%Qdp(i,j,k,q,np1_qdp) < 0 ) then
+                        v1=0  ! Q already negative, dont make it more so
+                     else
+                        v1 = -elem%state%Qdp(i,j,k,q,np1_qdp)
+                     endif
+                  endif
```
AaronDonahue: Just a comment on the above code that always confused me.  If we are not happy with negative Qdp then why do we even allow v1=0 to be a valid case thus allowing negative Qdp to propagate through the code?  Is it so that an error checker will catch it later?

Follow up, how do we feel about a case where `Qdp+v1 < 0` but `v1>0`?  My guess is that when we consider even the possibility of negative Qdp we're thinking of very small neg. numbers so its kind of an edge case and we just need to deal with it in some way.  But if `v1<0` inspires us to not make Qdp even more negative, but if `v1>0` then we trigger the "else" case and just bump v1 up so that Qdp+v1=0 then there is an inconsistentcy in how we see the relationship between the tendency and the value.

components/homme/src/theta-l/element_ops.F90
```diff
@@ -150,17 +145,37 @@ subroutine get_pottemp(elem,pottemp,hvcoord,nt,ntQ)
...
+  implicit none
+    
+  real (kind=real_kind), intent(out)  :: pi(np,np,nlev)
+  real (kind=real_kind), intent(in)   :: dp(np,np,nlev)
+  type (hvcoord_t),     intent(in)    :: hvcoord                      ! hybrid vertical coordinate struct
+  
+  integer :: k
+  real(kind=real_kind), dimension(np,np,nlevp) :: pi_i
+
+  pi_i(:,:,1)=hvcoord%hyai(1)*hvcoord%ps0
```
AaronDonahue: see my comment about using `hyai(nlev)` vs. `hyai(1)` to get the surface pressure.

AaronDonahue: Is this building `pi` from the "ground up"?  If so, then are the `k` indices correct in this usage recalling that `k=1` is the top of model?

components/homme/src/theta-l/prim_state_mod.F90
```diff
@@ -229,7 +228,8 @@ subroutine prim_printstate(elem, tl,hybrid,hvcoord,nets,nete)
        enddo
 
        ! surface pressure
-       tmp(:,:,ie)=elem(ie)%state%ps_v(:,:,n0)
+       !tmp(:,:,ie)=elem(ie)%state%ps_v(:,:,n0)
+       tmp(:,:,ie)=hvcoord%hyai(1)*hvcoord%ps0 + sum(elem(ie)%state%dp3d(:,:,:,n0),3)
```
AaronDonahue: see my comment about using `hyai(1)` vs. `hyai(nlev)` for the surface pressure.

components/homme/src/theta-l/prim_state_mod.F90
```diff
@@ -857,10 +862,13 @@ subroutine prim_energy_halftimes(elem,hvcoord,tl,n,t_before_advance,nets,nete)
        call TimeLevel_Qdp(tl, qsplit, tmp, t1_qdp) !get np1 into t2_qdp
     endif
     do ie=nets,nete
+#if 0
        do k=1,nlev
           dpt1(:,:,k) = ( hvcoord%hyai(k+1) - hvcoord%hyai(k) )*hvcoord%ps0 + &
                ( hvcoord%hybi(k+1) - hvcoord%hybi(k) )*elem(ie)%state%ps_v(:,:,t1)
        enddo
+#endif
```
AaronDonahue: Are we planning on deleting this code at some point, hence the `#if 0`?

components/homme/src/preqx/share/prim_state_mod.F90
```diff
@@ -168,8 +168,7 @@ subroutine prim_printstate(elem, tl,hybrid,hvcoord,nets,nete)
     time1 = time - dt
 
 
-    ! npts = np
-    npts=SIZE(elem(1)%state%ps_v(:,:,n0),1)
+    npts=np
```
AaronDonahue: **Not directly related to this PR, but general comment:**
Out of curiosity I dug into `global_integral` to see if this change has any impact there and I noticed that in the `global_integral` code `npts` is only passed to define the bounds on the input array for which we are integrating the global mass.  Within the body of the code the loop goes over `np`.  If we want the generality in `global_integral` to pass arbitrary sized arrays of `npts`X`npts` then why not have the loop indices in `global_integral` use `npts`?

components/homme/src/share/prim_driver_base.F90
mt5555: pushed fix.

components/homme/src/share/prim_driver_base.F90
mt5555: thanks - fixed.

components/homme/src/preqx/share/element_ops.F90
mt5555: removed

components/homme/src/preqx/share/prim_advance_mod.F90
mt5555: removed

components/homme/src/preqx/share/prim_state_mod.F90
mt5555: the full formula would be:
sum(dp) = [ hyai(nlevp)-hyai(1) ] p0  + [ hybi(nlevp)-hyba(1)] ps 

But at the surface, we must be pure terrain following and hyai(nlevp)=0, hybi(nlevp)=1
At the TOM, we must be pure pressure, and hybi(1)=0,  Thus we are left with:

sum(dp) = -hyai(1)*p0 + ps

hyai(1)*p0 is the pressure at the TOM  (prescribed as a boundary condition)

AaronDonahue: ahhh yes, thank you for the explanation.

components/homme/src/share/prim_driver_base.F90
mt5555: forgot to finish that comment - no improved.

components/homme/src/share/prim_driver_base.F90
mt5555: agreed.

components/homme/src/share/prim_driver_base.F90
mt5555: fixed

components/homme/src/theta-l/element_ops.F90
mt5555: see notes above.  We build pi from the TOM (where hydrostatic pressure is prescribed as a boundary condition) down to the surface.

components/homme/src/preqx/share/prim_state_mod.F90
mt5555: agreed - global_integral() used to support arbitrary size arrays, but now the arrays must be of size npts=np, and this argument should be removed.  Lets fix this in a separate PR.

components/homme/src/share/prim_driver_base.F90
mt5555: That's true, if phrased that way it seems inconsistent.  But it can also be phrased this way:

If Q<0, then we neglect all the forcing.
if Q>0 but Q+fq<0, then we only apply a % of the forcing  
if Q+fq>0    we apply the full forcing

So in this sense I believe it is consistent.  We never change the sign of the forcing, all we do is reduce it if it would drive Q negative (but we never reduce it below zero).

mt5555: No objections to changing this in the future, but for this PR we need to keep it this way to maintain BFB.

components/homme/src/theta-l/element_ops.F90
rljacob: You have a variable called "pi" that isn't 3.14 ?

mt5555: p = nonhydrostatic pressure, pi=hydrostatic pressure.  But I can see some might not like that convention.  I'll rename it phydro

components/homme/src/share/prim_driver_base.F90
AaronDonahue: Oh absolutely.  I think the PR is ready to go.  I just brought it up as something I've thought everytime I've looked at this chunk of code.

AaronDonahue:
@mt5555 , thank you for taking the time to reply to all my comments.  Sorry if a few of them were a bit in the weeds.  The PR is ready to go.

components/homme/src/share/prim_driver_base.F90
```diff
@@ -1229,7 +1212,18 @@ subroutine prim_step(elem, hybrid,nets,nete, dt, tl, hvcoord, compute_diagnostic
     ! for ftype==4, also apply dynamics tendencies from forcing
     ! for ftype==4, energy diagnostics will be incorrect
     ! ===============
-    if (ftype==4) call ApplyCAMforcing_dynamics(elem,hvcoord,tl%n0,dt,nets,nete)
+    if (ftype==4) then
```
oksanaguba: Not sure how relevant it is but energy_halftimes() uses Q, while valid Q is only computed after remap? So, Q is not valid for ftype4 diagnostics calls and below, for HV? However, energy itself does not use Rstar, so, for energy it must be ok?

PR #3044, opened 2019-07-06
Modify 'make baseline' to continue running when a baseline fails.

### LINKED ISSUE #3049: Improve HOMME's cmake logic to check compiler in a CUDA build
Currently, when HOMME has to check if it's a GPU build, cmake inspects the name of the C++ compiler for the substring `nvcc`. This breaks when building with the MPI wrappers, using the standard way of modifying MPI's undelying C++ compiler (i.e., changing the OMPI_CXX env variable), since the C++ compiler name is still `mpicxx` (with a different back-end).

A better way to check that the C++ compiler is indeed `nvcc_wrapper` from Kokkos would be to try to run the compiler with the flag `--nvcc-wrapper-show` (which is peculiar to `nvcc_wrapper`) and check the output. `nvcc_wrapper` would output a string a long the line of `nvcc -ccbin g++ -arch=sm_70` while any other compiler would generate an error/warning about the unrecognized option, plus an error because no file was specified.

pbosler:
Fixed an issue where a failed test would cause 'make baseline' to crash without running other tests.

Fixes #3049 

[BFB]


ambrad:
I like this change. It can be useful on testbed machines to make it through the baseline process even if a test fails.

What do you think of cherry-picking this related commit into the PR? It disables the SL test when SL isn't built. This fixes the problem of running this test on a GPU machine. SL doesn't yet run end-to-end on GPU: https://github.com/ambrad/E3SM/commit/ef9529572804cfb563e9eea1b15c424f7f3c74b1

ambrad:
I suppose a counterargument is that someone might think the fact that `make baseline` passes means the build is good on a new machine. Currently that is true; with this PR, it no longer would be.

pbosler:
I think it's good that each baseline test gets a chance to run -- it will cover more code, at a minimum.  Plus, if a test is known to fail, like the SL test on GPU, that shouldn't stop other dev from moving forward.  I'll add "Test failed" to the output to make sure it's clear.  

I agree with cherry-picking the above SL commit.

oksanaguba:
What if we are running making baselines on some machine thru ./create_test ... ? Does this new change mean case status will not show that some baseline was not generated? If there is nothing alarming in the case status, there is no reason to read log files closely. I guess i can test this by messing with some namelist to cause it crash...

pbosler:
Sample output:

```Running test baroCamMoist-run ... /ascldap/users/pabosle/hommexx/cuda/tests/baseline/baroCamMoist/baroCamMoist-run.sh > baroCamMoist-run.out 2> baroCamMoist-run.err
test baroCamMoist-run was run successfully
Running test baroCamMoistSL-run ... /ascldap/users/pabosle/hommexx/cuda/tests/baseline/baroCamMoistSL/baroCamMoistSL-run.sh > baroCamMoistSL-run.out 2> baroCamMoistSL-run.err
Test failed.
baroCamMoistSL-run.err file not found. Check baroCamMoistSL-run directory for details.```

**Comment by ambrad** [comment]
Pete, re: Oksana's comment. You could move the `exit -7` line to the end of the routine. Accumulate a fail count. If the count is 0 at the end, do nothing; if it is >0, then exit -7. That way, the baselines all get a chance to get created, but the exit status is still bad if at least one fails.

**Comment by pbosler** [comment]
With the new change, the error return is back:

```test preqx-nlev72-qsize10-r3-lim9-dry-kokkos-ne12-ndays9-run was run successfully
1 of 79 tests failed.
make[3]: *** [test_execs/CMakeFiles/baseline] Error 249
make[2]: *** [test_execs/CMakeFiles/baseline.dir/all] Error 2
make[1]: *** [test_execs/CMakeFiles/baseline.dir/rule] Error 2
make: *** [baseline] Error 2```

**Comment by pbosler** [comment]
Add a fix for #3049

**Comment by jgfouca** [comment]
@oksanaguba , are you OK with this PR?

**Comment by oksanaguba** [comment]
@jgfouca not yet. I want to run this from cime and confirm it behaves as it should. this is a low priority feature, not a bug (I will change labels).

File: components/homme/test/reg_test/run_tests/test-list.cmake [hunk 1]
```diff
@@ -9,7 +9,6 @@ SET(HOMME_TESTS
   baro2c.cmake
   baro2d.cmake
   baroCamMoist.cmake
-  baroCamMoist-SL.cmake
```
mt5555: Is this PR removing the baroCamMoist-SL test?

rljacob:
It *is* fixing a bug so adding back the bug-fix lable.

components/homme/test/reg_test/run_tests/test-list.cmake
oksanaguba: Only without compose build. So, should run on our machines, we don't run routinely on gpus.

oksanaguba:
@rljacob this PR fixes infrastructure issues; one is about nvcc-wrapper, maybe, it can be considered a bug; I only meant it does not fix bugs in homme as in dycore.

components/homme/test/reg_test/run_tests/test-list.cmake
ambrad: Right, line 27 of this file appends baroCamMoist-SL.cmake to the list if HOMME_ENABLE_COMPOSE=TRUE.

pbosler:
@rljacob @oksanaguba: I originally added "bug fix" label because I thought it was a bug that one baseline failure would prevent all of the others from running.   Depending on the order of the tests relative to the failure, this could be a large number.  In my case, for example, a test we expected to fail, did fail; it was number 8 of 91 and its failure prevented the other 83 baselines from being run -- to me that seemed like a bug since we depend on those baselines to verify subsequent PRs.

rljacob:
@mt5555 please update your review if your changes have been made.

oksanaguba:
I still need to test it,so, it is not ready.

rljacob:
This PR is a month old already.  Make time to test it.

oksanaguba:
I am trying this now: I merged it and broke 1st ru for baroCamMoist by meddling with nl. The 1st run for this set did not run, the rest ran (because they are omp runs with different namelists). Total output is 
```
[onguba@skybridge-login7 acmebld]$ ./cs.status.20190815_162407_bsqt7c 
20190815_162407_bsqt7c: 1 test
  HOMME_P24.f19_g16_rx1.A.sandiatoss3_intel (Overall: PASS) details:
    PASS HOMME_P24.f19_g16_rx1.A.sandiatoss3_intel CREATE_NEWCASE
    PASS HOMME_P24.f19_g16_rx1.A.sandiatoss3_intel XML
    PASS HOMME_P24.f19_g16_rx1.A.sandiatoss3_intel SETUP
    PASS HOMME_P24.f19_g16_rx1.A.sandiatoss3_intel SHAREDLIB_BUILD time=0
    PASS HOMME_P24.f19_g16_rx1.A.sandiatoss3_intel NLCOMP
    PASS HOMME_P24.f19_g16_rx1.A.sandiatoss3_intel MODEL_BUILD time=274
    PASS HOMME_P24.f19_g16_rx1.A.sandiatoss3_intel SUBMIT
    PASS HOMME_P24.f19_g16_rx1.A.sandiatoss3_intel RUN time=456
    PASS HOMME_P24.f19_g16_rx1.A.sandiatoss3_intel GENERATE
    PASS HOMME_P24.f19_g16_rx1.A.sandiatoss3_intel SHORT_TERM_ARCHIVER
```
I moved to break a simpler test, baro2b. It did not run, movies folder is empty, there is an error, but the output at the end contains
```
Running test baro2b-run ... /gpfs1/onguba/acme_scratch/sandiatoss3/HOMME_P24.f19_g16_rx1.A.sandiatoss3_intel.G.20190815_163733_6dnjlo/bld/tests/baseline/baro2b/baro2b-run.sh > baro2b-run.out 2> baro2b-run.err
test baro2b-run was run successfully
```
So, for now the code completes baselines (the next test, baro2c, has movies), but it also shows PASS 
```
PASS HOMME_P24.f19_g16_rx1.A.sandiatoss3_intel GENERATE
PASS HOMME_P24.f19_g16_rx1.A.sandiatoss3_intel SHORT_TERM_ARCHIVER
```
I think it should show FAIL.

oksanaguba:
@pbosler ^^

oksanaguba:
Since in this version case status returns pass even when some baselines failed (which can lead to broken baselines), could I suggest that the change about SL test is moved to a new PR and this PR is not being merged for now?

rljacob:
@pbosler, @oksanaguba what is the status of this PR?

pbosler:
@oksanaguba The reason the test passes is because pass/fail is determined by the executable's return value to the shell, e.g., the `RUN_STAT` variable in `testing_utils.sh`:

```
cmd="homme_executable > $THIS_STDOUT 2> &THIS_STDERR"
$cmd
RUN_STAT=$?

```
Hence, in order to show failure, the test must crash, throw an exception, or otherwise not `return 0` to the shell.  Is it possible that the break you introduced succeeded in writing an error message but still returned 0?

oksanaguba:
Thanks, Pete @pbosler . @rljacob I will get back to this PR in 1-2 weeks.

rljacob:
@oksanaguba what is the status of this PR?

oksanaguba:
I did not have time to work on this.

rljacob:
But what's the status going forward?

rljacob:
Closing until this can be worked on.

PR #3045, opened 2019-07-06
Chemistry preprocessor compiler fixes for Cori

cameronsmith1:
The chemistry pre-processor would not compile and run correctly on Cori.  It was previously configured to work on Titan.

This PR fixes the compiler and CPP options so that it will work on Cori.

cpp is used by the chemistry pre-processor to manipulate the fortran output. It seems that modern versions of cpp behave differently than older versions, so the chemistry pre-processor has problems and produces different fortran output. There are two main issues:

1) cpp adds its copyright information to the top of the output using
   C-type comments, which confuses the chemistry pre-processor (and
   probably the E3SM Fortran compiler).

2) cpp deletes parts of the fortran output because it misidentifies it
   as a comment.

Issue 1 occurs with the -C option to cpp, and issue 2 occurs without
the -C option to cpp.  This commit solves the problem by replacing the
-C option with -traditional-cpp.

For future convenience, I copied pp_linoz_mam4_resus_mom_soag.inp from the pre-processed directory in E3SM to the chemistry pre-processor input directory (and added the header so that it will work with the standalone pre-processor).  Note: linoz_mam4_resus_mom_soag is the standard chemistry configuration for E3SMv1.

[E3SM is BFB.  Output of chemistry pre-processor on Cori is now the same as it was on Titan.]

### DIFF
--- a/components/cam/chem_proc/src/Makefile
+++ b/components/cam/chem_proc/src/Makefile
@@ -140,7 +140,6 @@ ifeq ($(COMPILER),xlf95)
   FFLAGS      := -g -c -qarch=auto -qnosave -qfree=f90 -qmoddir=$(OBJ_DIR) -I $(OBJ_DIR) -qstrict -O3
 endif
 ifeq ($(COMPILER),gfortran)
-# FFLAGS      := -g -c -ffree-form
   FFLAGS      := -g -c -ffree-form -J $(OBJ_DIR)
 endif
 ifeq ($(COMPILER),g95)
--- a/components/cam/chem_proc/src/Makefile
+++ b/components/cam/chem_proc/src/Makefile
@@ -140,10 +140,14 @@ ifeq ($(COMPILER),xlf95)
   FFLAGS      := -g -c -qarch=auto -qnosave -qfree=f90 -qmoddir=$(OBJ_DIR) -I $(OBJ_DIR) -qstrict -O3
 endif
 ifeq ($(COMPILER),gfortran)
-  FFLAGS      := -g -c -ffree-form
+# FFLAGS      := -g -c -ffree-form
+  FFLAGS      := -g -c -ffree-form -J $(OBJ_DIR)
 endif
 ifeq ($(COMPILER),g95)
   FFLAGS      := -g -c -ffree-form
+endif
+ifeq ($(COMPILER),ftn)
+  FFLAGS      := -f free
 endif
 
 #------------------------------------------------------------------------
--- a/components/cam/chem_proc/src/cam_chempp/mozpp.main.f
+++ b/components/cam/chem_proc/src/cam_chempp/mozpp.main.f
@@ -173,7 +173,8 @@
       sim_dat_filename = 'sim.dat'
       sim_dat_filespec = trim(sim_dat_path) // 'sim.dat'
       cpp_dir  = ' '
-      cpp_opts = '-P -C -I.'
+!     cpp_opts = '-P -C -I.'  ! -C causes problems on Cori
+      cpp_opts = '-P -I.'
 
 !-----------------------------------------------------------------------
 !        ... Assign default input, output units
--- a/components/cam/chem_proc/src/cam_chempp/mozpp.main.f
+++ b/components/cam/chem_proc/src/cam_chempp/mozpp.main.f
@@ -174,7 +174,7 @@
       sim_dat_filespec = trim(sim_dat_path) // 'sim.dat'
       cpp_dir  = ' '
 !     cpp_opts = '-P -C -I.'  ! -C causes problems on Cori
-      cpp_opts = '-P -I.'
+      cpp_opts = '-P -I. -traditional-cpp'  ! Works on Cori (2019-07-15)
 
 !-----------------------------------------------------------------------
 !        ... Assign default input, output units
--- a/components/cam/chem_proc/inputs/pp_linoz_mam4_resus_mom_soag.inp
+++ b/components/cam/chem_proc/inputs/pp_linoz_mam4_resus_mom_soag.inp
@@ -1,0 +1,109 @@
+BEGSIM
+output_unit_number = 7
+output_file        = linoz_mam4_resus_mom_soag.doc
+procout_path       = ../output/
+src_path           = ../bkend/
+procfiles_path     = ../procfiles/cam/
+sim_dat_path       = ../output/
+sim_dat_filename   = linoz_mam4_resus_mom_soag.dat
+
+Comments
+     E3SMv1 DECK chemistry.
+End Comments
+
+   SPECIES
+
+      Solution
+ O3
+ H2O2, H2SO4, SO2, DMS -> CH3SCH3, SOAG -> C
+ so4_a1 -> NH4HSO4
+ pom_a1 -> C, soa_a1 -> C, bc_a1 -> C
+ dst_a1 -> AlSiO5, ncl_a1 -> NaCl
+ mom_a1 -> C8520H11360O8520
+ num_a1 -> H
+ so4_a2 -> NH4HSO4
+ soa_a2 -> C, ncl_a2 -> NaCl
+ mom_a2 -> C8520H11360O8520
+ num_a2 -> H
+ dst_a3 -> AlSiO5, ncl_a3 -> NaCl
+ so4_a3 -> NH4HSO4
+ bc_a3 -> C, pom_a3 -> C
+ soa_a3 -> C
+ mom_a3 -> C8520H11360O8520
+ num_a3 -> H
+ pom_a4 -> C, bc_a4 -> C
+ mom_a4 -> C8520H11360O8520
+ num_a4 -> H
+      End Solution
+
+      Fixed
+ M, N2, O2, H2O, OH, NO3, HO2, cnst_O3 -> O3
+      End Fixed
+
+      Col-int
+ O3 = 0.
+ O2 = 0.
+      End Col-int
+
+   End SPECIES
+
+   Solution Classes
+      Explicit
+        O3
+      End Explicit
+      Implicit
+        H2O2, H2SO4, SO2, DMS, SOAG
+        so4_a1, pom_a1
+        soa_a1, bc_a1, dst_a1, ncl_a1
+	mom_a1
+        num_a1
+        so4_a2, soa_a2, ncl_a2, mom_a2, num_a2
+        dst_a3, ncl_a3, so4_a3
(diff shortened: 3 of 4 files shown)


cameronsmith1:
The one issue I know about so far, is that recreating the mechanism linoz_mam4_resus_mom_soag creates different code:

```
diff ./mo_imp_sol.F90 ../../../src/chemistry/pp_linoz_mam4_resus_mom_soag/mo_imp_sol.F90
297,298c297,298
<        call addfld( trim(solsym(j))
<        call addfld( trim(solsym(j))
---
>        call addfld( trim(solsym(j))//'_CHMP', (/ 'lev' /), 'I', '/cm3/s', 'chemical production rate' )
>        call addfld( trim(solsym(j))//'_CHML', (/ 'lev' /), 'I', '/cm3/s', 'chemical loss rate' )
658,659c658,659
<        call outfld( trim(solsym(j))
<        call outfld( trim(solsym(j))
---
>        call outfld( trim(solsym(j))//'_CHMP', prod_out(:,:,i), ncol, lchnk )
>        call outfld( trim(solsym(j))//'_CHML', loss_out(:,:,i), ncol, lchnk )
```

susburrows:
Hi @cameronsmith1 , are there specific aspects of this that we need to review, e.g., whether one of us can reproduce the build on cori?

cameronsmith1:
I think the big question is: why is there a discrepancy in the fortran produced now versus previously?

rljacob:
Is there any config that tests the chemistry?

susburrows:
@rljacob The linoz_mam4_resus_mom_soag mechanism is the default atmospheric chemistry mechanism, so it is exercised by all of the tests that have an active EAM atmosphere component.

singhbalwinder:
@cameronsmith1 : Would you please outline the steps you are using to create these files? There are two ways:
1. Use the make file to compile the model and then run the executable with the input file
2. Build a case (using create_newcase) and run just the ./case.setup with the following additional arguments to CAM_CONFIG_OPTS:
`-usr_mech_infile /path/to/user_supplied_mech.in`

I used the 2nd way and I got the files which were exactly the same as in `pp_linoz_mam4_resus_mom_soag` folder except that all new files have a disclaimer on the top and few extra lines.  Specifically, I didn't see differences you are seeing in `mo_imp_sol.F90`.

I haven't tried the 1st method yet.

cameronsmith1:
I was using method 1 to build the files.  This makes it easy to generate the files that need to be put into pp_* directory.  

I thought that method 2 didn't work anymore (I am glad it does).  The disclaimer caused all sorts of errors when creating the output code using method 1, and the header is a C header, so I expect it will fail to compile as part of E3SM. 

As for tests, the previously generated output from the pre-processor that is archived in the PP_* directory is tested, but there is no test that makes sure the pre-processor will compile and still build the existing output code (AFAIK).

cameronsmith1:
I did a new commit to change the cpp flags, and that seems to have fixed the problem.  Specifically, I replaced ```-C``` with ```-traditional-cpp```

Can someone test that this works on other machines?   Specifically, I have been checking that running the chemistry pre-compiler on components/cam/chem_proc/inputs/pp_linoz_mam4_resus_mom_soag.inp will produce the fortran output stored in components/cam/src/chemistry/pp_linoz_mam4_resus_mom_soag/

cameronsmith1:
What do we need to do to move forward on merging this to master?

singhbalwinder:
Thanks @cameronsmith1 for figuring this out. I will test it on my end on a different machine. I am good with changes in mozzpp.main.f file. The other files seem machine/compiler specific. Do you think we should integrate those? Also, I don't think we should integrate the chemistry input file. I am assuming that you used the input file just for testing purposes.

cameronsmith1:
Thanks for testing on other systems.  In a perfect universe we should have something that works on every machine.  However, at a minimum we have should have something that works on at least one machine -- that was Titan, and now I think it should be Cori.  Hence, I think it is critical we put in my Cori specific changes, unless there is something that is more universal.

The pre-processor input directory is supposed to contain all the input files, so I think it is a major oversight not to have the input file for the main E3SMv1 chemistry there.  This PR just corrects the oversight.  It is also critical when developing new mechanisms to have an input file to work from, and this provides that.

singhbalwinder:
I have tested this on Compy with the Intel compiler. I had to set `setenv COMPILER ifort` variable in my environment to compile it with Intel (otherwise it was picking up the defaults). The use of `traditional-cpp` removed the GNU disclaimer from the files and the files I got were exactly the same as in `pp_linoz_mam4_resus_mom_soag` (except few additional white spaces and lines).

Therefore the code seems to be working as expected on Compy if we replace the `-C` flag with `traditional-cpp` (to get rid of the disclaimer). I think we should add this change and also add the remaining files as you mentioned.

cameronsmith1:
That is good news.  Thanks, @singhbalwinder.  

There are other compiler options configurations commented out in make_chempp.cori.  These are options that have worked in the past, and I think they are worth keeping as options for the next time we need to get the pre-processor working.   Does uncommenting the intel option work on compy?   If so, I suggest we add a comment in the file to note that it works on Compy.  If not, is there a tweak that will implement what you did?

singhbalwinder:
> Does uncommenting the intel option work on compy?

If you are referring to the changing `make_chempp.cori` file above, it won't work on Compy as Compy doesn't have `PrgEnv-*` modules. I think if we want to compile this code on a new (or different) machine we will have to load a compiler and run make on that machine. This tool needs just a compiler, so it is fairly easy to build.

I tested it further on Cori and was not able to reproduce your original error (`call outfld( trim(solsym(j) .....`). The code was behaving as expected if I just remove `-C' flag`.  We don't even need to add `-traditional-cpp` or `-traditional` flag.

components/cam/chem_proc/src/Makefile
```diff
@@ -140,11 +140,15 @@ ifeq ($(COMPILER),xlf95)
   FFLAGS      := -g -c -qarch=auto -qnosave -qfree=f90 -qmoddir=$(OBJ_DIR) -I $(OBJ_DIR) -qstrict -O3
 endif
 ifeq ($(COMPILER),gfortran)
-  FFLAGS      := -g -c -ffree-form
+# FFLAGS      := -g -c -ffree-form
+  FFLAGS      := -g -c -ffree-form -J $(OBJ_DIR)
```
singhbalwinder: I do not know about this `-J` flag. The code works without this new addition. Is there a reason this was added?

components/cam/chem_proc/src/cam_chempp/mozpp.main.f
```diff
@@ -173,7 +173,8 @@ program mozart_pp
       sim_dat_filename = 'sim.dat'
       sim_dat_filespec = trim(sim_dat_path) // 'sim.dat'
       cpp_dir  = ' '
-      cpp_opts = '-P -C -I.'
+!     cpp_opts = '-P -C -I.'  ! -C causes problems on Cori
+      cpp_opts = '-P -I. -traditional-cpp'  ! Works on Cori (2019-07-15)
```
singhbalwinder: I think we can just remove the `-C` flag. Also, we should just get rid of the commented-out line.

components/cam/chem_proc/src/Makefile
cameronsmith1: The -J flag controls where the intermediate compiled files go.   Without the -J flag they clogged up the src directory.  The -J flag puts them in a separate directory $(OBJ_DIR), which I think has been the historical place for them.  Hence, it is not strictly necessary, but it is valuable and increases consistency across different compilers without any downside.  Hence, I recommend we keep it in the PR.

components/cam/chem_proc/src/cam_chempp/mozpp.main.f
cameronsmith1: Did you diff the fortran files to those archived in src/chemistry ?

cameronsmith1: My personal preference is to leave in (as comments) things that may be relevant to getting the pre-processor to work an another machine.  It is a cheap alternative to IF statements for different machines.

components/cam/chem_proc/src/Makefile
singhbalwinder: Thanks! It works for gfortran but it doesn't do anything if we use Intel. Line 143 above (`# FFLAGS      := -g -c -ffree-form`) is not doing anything. Is it okay if I remove it?

components/cam/chem_proc/src/cam_chempp/mozpp.main.f
singhbalwinder: Agreed. Lets keep it the way it is.

singhbalwinder:
I am fine with the code in its current form. It is being tested on Compy and Cori and it works as expected with these new changes (thanks @cameronsmith1 !). 

We should add a test for this code. It might be a good idea to add that test when we add tests for other CAM tools.

components/cam/chem_proc/src/Makefile
cameronsmith1: Yes, you can remove the commented line.  I was keeping it while I was testing new options.  Since I ended up just adding an extra option (-J) there seems little value in keeping the original line around.  Can you remove it?  Or do you need me to do it?

cameronsmith1:
I just tested this PR after the recent Cori upgrade, and it still compiles and produces the expected code output (BFB).

What do we need to do to finalize this PR?

singhbalwinder:
Thanks Philip! I will make that change and merge this to next and master.

cameronsmith1:
I am back from jury duty.   How is this going?

components/cam/chem_proc/src/Makefile
cameronsmith1: I just removed the commented out line, for clarity.

cameronsmith1: I just pushed a commit to remove the commented-out line.

cameronsmith1:
I just pushed a commit to remove the commented-out in the Makefile, that we discussed.

I think this is now ready for merging to next, then master.

singhbalwinder:
Thanks @cameronsmith1 ! There is another big atm PR being integrated right now. I will wait for that to merge to master before merging this one.

singhbalwinder:
Merged to next

cameronsmith1:
Thanks!

cameronsmith1:
Note: a problem arose with the NERSC modules soon after this PR was pushed to master. It has now been fixed by NERSC.  See issue https://github.com/E3SM-Project/E3SM/issues/3137 for details.

PR #3050, opened 2019-07-09
Update gravity wave drag mods to ingest improved tuning for QBO via nml

wlin7:
Convective heating rate conversion factor (CF) and gravity wave tendency efficiency
for Beres scheme (effgw_beres) are retuned to improve QBO for adoption in EAMv2.

In the new tuning with much improved QBO, the conversion factor is
changed to 12.5 from 20.0, and the tendency efficiency is reduced to 0.35 from 0.4.

This implementation adds namelist control for the conversion factor to incorporate the
new tuning for V2 while preserving reproducibility of all existing compsets.
The new values for CF and effgw_beres will be ingested via new compsets and corresponding use case files.

[BFB]

### DIFF
--- a/components/cam/bld/namelist_files/namelist_definition.xml
+++ b/components/cam/bld/namelist_files/namelist_definition.xml
@@ -1168,6 +1168,12 @@ In the low-top model, this helps to conserve momentum and produce a QBO.
 Default: set by build-namelist.
 </entry>
 
+<entry id="gw_convect_hcf" type="real" category="gw_drag"
+       group="gw_drag_nl" valid_values="" >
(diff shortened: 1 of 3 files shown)


rljacob:
Code review document:
https://acme-climate.atlassian.net/wiki/spaces/ED/pages/1000374369/W8+-+Retuning+of+GWD+to+improve+QBO

singhbalwinder:
Merged to next

singhbalwinder:
This PR is not NML, so Ihave removed NML tag.  @wlin7 and I tested it independently.

PR #3051, opened 2019-07-09
Improve accuracy of the SCM "Replay" mode

bogensch:
This PR improves the accuracy of the SCM Replay mode.  Previous implementation allowed the user to “replicate the behavior” of an E3SM run in SCM, but with temperature errors on the order of ~0.1 K after a day.  This PR will provide the user with capability to produce “quasi-bit-for-bit” results with Replay mode, with temperature errors on the order of ~1.e-5 K.  

The reason why Replay mode is unable to provide fully B4B results is because the dynamics tendency computed for Replay is not identical to the way that the full model computes its dynamics tendencies (which involves sub-cycling in the SE dynamical core).  Thus, there is issue with roundoff.  Note that in the past, CAM Replay mode with Eulerian dy-core could produce “B4B” results but only because if the flag to generate Replay mode output was turned on, they updated their T and Q based on the tendency computed for Replay purposes.  Thus the run to generate Replay forcing was NOT b4b with a normal production run.  We are working on a truly B4B Replay mode for E3SM and a future PR will support this which will involve higher precision computation of the dynamical tendencies.    

This PR also renames the old “-camiop” flag to “-e3smreplay” flag to be more current and consistent with E3SM.  Other flags are renamed accordingly throughout the code.

Note that in this PR the e3sm_developer test “SMS_R_Ld5.ne4_ne4.FSCM5A97” is an expected fail when comparing results to baseline.  This is because where the large-scale vertical velocity is read in and updated had to be moved so that CLUBB could see it at the consistent timestep.  It was validated that results for selected tested cases (ARM97, GOAMAZON, DYCOMS-RF01, BOMEX) have very minimal impacts with this fix included.  

All other e3sm_developer tests pass.

### DIFF
--- a/cime/src/drivers/mct/main/cime_comp_mod.F90
+++ b/cime/src/drivers/mct/main/cime_comp_mod.F90
@@ -1887,7 +1887,7 @@ contains
     !----------------------------------------------------------
 
     areafact_samegrid = .false.
-#if (defined BFB_E3SM_SCM_IOP )
+#if (defined E3SM_SCM_REPLAY )
     if (.not.samegrid_alo) then
        call shr_sys_abort(subname//' ERROR: samegrid_alo is false - Must run with same atm/ocn/lnd grids when configured for scam iop')
     else
--- a/cime/src/drivers/mct/main/cime_comp_mod.F90
+++ b/cime/src/drivers/mct/main/cime_comp_mod.F90
@@ -1887,7 +1887,7 @@ contains
     !----------------------------------------------------------
 
     areafact_samegrid = .false.
-#if (defined BFB_CAM_SCAM_IOP )
+#if (defined BFB_E3SM_SCM_IOP )
     if (.not.samegrid_alo) then
        call shr_sys_abort(subname//' ERROR: samegrid_alo is false - Must run with same atm/ocn/lnd grids when configured for scam iop')
     else
--- a/components/cam/bld/build-namelist
+++ b/components/cam/bld/build-namelist
@@ -3979,7 +3979,7 @@ if ($cfg->get('scam')) {
 }
 
 # CAM generates IOP file for SCAM
-if ($cfg->get('e3smreplay')) {
+if ($cfg->get('e3smreplay') or $cfg->get('e3smreplay_b4b')) {
     add_default($nl, 'inithist', 'val'=>'CAMIOP');
     add_default($nl, 'ndens', 'val'=>'1,1');
     add_default($nl, 'mfilt', 'val'=>'1,10');
--- a/components/cam/bld/build-namelist
+++ b/components/cam/bld/build-namelist
@@ -3979,7 +3979,7 @@ if ($cfg->get('scam')) {
 }
 
 # CAM generates IOP file for SCAM
-if ($cfg->get('camiop')) {
+if ($cfg->get('e3smreplay')) {
     add_default($nl, 'inithist', 'val'=>'CAMIOP');
     add_default($nl, 'ndens', 'val'=>'1,1');
     add_default($nl, 'mfilt', 'val'=>'1,10');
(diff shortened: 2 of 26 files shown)


cime/src/drivers/mct/main/cime_comp_mod.F90
```diff
@@ -1887,7 +1887,7 @@ subroutine cime_init()
     !----------------------------------------------------------
 
     areafact_samegrid = .false.
-#if (defined BFB_CAM_SCAM_IOP )
+#if (defined E3SM_SCM_REPLAY )
     if (.not.samegrid_alo) then
        call shr_sys_abort(subname//' ERROR: samegrid_alo is false - Must run with same atm/ocn/lnd grids when configured for scam iop')
```
wlin7: Not essential but just to have a more consistent message. change "when confihgured for scam iop" to " ... for scm replay". Similarly for line 1702.

components/cam/bld/build-namelist
```diff
@@ -3979,7 +3979,7 @@ if ($cfg->get('scam')) {
 }
 
 # CAM generates IOP file for SCAM
-if ($cfg->get('camiop')) {
+if ($cfg->get('e3smreplay')) {
     add_default($nl, 'inithist', 'val'=>'CAMIOP');
```
wlin7: About the 'CAMIOP' type of inithist: do we want to rename this type to reflect the current naming convention for this feature? If changed, it would trigger changes in several other files (including configure, build-namelist, namelist_definition.xml, and cam_history.F90). It can certainly be considered in a future PR.

components/cam/src/control/scamMod.F90
```diff
@@ -678,13 +678,13 @@ subroutine readiopdata(iop_update_surface,hyam,hybm)
        'readiopdata.F90', __LINE__)
 
 !
-!     if the dataset is a CAM generated dataset set use_camiop to true
-!       CAM IOP datasets have a global attribute called CAM_GENERATED_IOP      
+!     if the dataset is a CAM generated dataset set use_replay to true
+!       E3SM IOP datasets have a global attribute called E3SM_GENERATED_IOP
```
wlin7: line 682 global attribute "E3SM_GENERATED_FORCING" instead of "E3SM_GENERATED_IOP"

components/cam/src/physics/cam/cam_diagnostics.F90
```diff
@@ -1574,11 +1574,12 @@ subroutine diag_surf (cam_in, cam_out, ps, trefmxav, trefmnav )
     call outfld('RHREFHT',   ftem,      pcols, lchnk)
 
 
-#if (defined BFB_CAM_SCAM_IOP )
+#if (defined E3SM_SCM_REPLAY )
     call outfld('shflx   ',cam_in%shf,   pcols,   lchnk)
     call outfld('lhflx   ',cam_in%lhf,   pcols,   lchnk)
     call outfld('trefht  ',cam_in%tref,  pcols,   lchnk)
     call outfld('Tg', cam_in%ts, pcols, lchnk)
+    call outfld('Tsair',cam_in%ts, pcols, lchnk)
```
wlin7: I think cam_in%tref is more suitable as Tsair here. If PS is not exactly reproduced, tsair could be used for vertical interpolation of T.

wlin7:
Hi @bogensch , the PR looks good. The suggested changes are really not essential. Can you take a look if you would like to get this PR into master as is for now, and probably consider the suggested changes in a future PR?

bogensch:
Hi @wlin7 thanks for the careful review, as always.  Sounds good to me, yes let's get the PR in for now and I'll consider your comments in another SCM branch I'm working on.

wlin7:
Thanks @bogensch . your plan sounds good to me. Not sure when regression tests are going to start tonight. Look to merge it tomorrow.

wlin7:
Merged to next.

wlin7:
Merged. Request to bless SCM test on sandiatoss3 and melvin.

PR #3052, opened 2019-07-09
Remove P96 from pg2 test

whannah1:
Removes the processor specification for pg2 test and also remove settings in config_pesall.xml that forced 96 processors on any platform.

[BFB]


rljacob:
merged to next.

rljacob:
@whannah1 its still picking up 96 procs.  Did you specify that in cime/config/e3sm/allactive/config_pesall.xml ?

whannah1:
ah, yes I did. I forget why I felt the need to do that. 
Can I just remove the pg2 entries in that file entirely? Or does it need to have a value there?

rljacob:
You can remove them.

whannah1:
Done.

Issue #1970, opened 2017-12-07
Netcdf file types at NERSC


rljacob:
Recording this observation from @ndkeen 

> When I walk the entire inputdata directory at NERSC, I counted: `Number of files=     44566  Total GB=  8173.320 `   Of those files, 36837 are netcdf files (identified by .nc extension). I then (after loading default pnetcdf module on Cori) used `ncdump -k` on each file, and here are results:

```
classic               33103
64-bit offset          3619
netCDF-4 classic model   55
netCDF-4                 25
Permission denied        27
Unknown file format       4
cdf5                      4
total                 36837
```

rljacob:
Closing this.  See #2292 for more.

PR #3053, opened 2019-07-09
Update use cases to use netcdf3 input files.

rljacob:
Update all use cases that were using a netcdf4 file to instead use the netcdf3 version.

This will allow pnetcdf3 to be used for all these cases.

Only atmosphere files were changed.  Most for cmip6.  One for scam

This update to maint-1.0 will be brought forward to master later.

[BFB]

### DIFF
--- a/cime/config/e3sm/testmods_dirs/allactive/v1cmip6/shell_commands
+++ b/cime/config/e3sm/testmods_dirs/allactive/v1cmip6/shell_commands
@@ -1,7 +1,6 @@
 #!/bin/bash
 ./xmlchange --append CAM_CONFIG_OPTS='-cosp'
 ./xmlchange --id BUDGETS --val TRUE
-./xmlchange PIO_TYPENAME=netcdf
 if [ `./xmlquery --value MACH` == cetus ]||[ `./xmlquery --value MACH` == mira ]; then sed s/64M/128M/ env_mach_specific.xml >tmp && mv tmp env_mach_specific.xml; fi
 
 if [ `./xmlquery --value MACH` == theta ]; then
--- a/components/cam/bld/namelist_files/use_cases/1850_cam5_CMIP6.xml
+++ b/components/cam/bld/namelist_files/use_cases/1850_cam5_CMIP6.xml
@@ -5,7 +5,7 @@
 <cosp_lite>.true.</cosp_lite>
 
 <!-- Solar constant from CMIP6 input4MIPS -->
-<solar_data_file>atm/cam/solar/Solar_1850control_input4MIPS_c20171101.nc</solar_data_file>
+<solar_data_file>atm/cam/solar/Solar_1850control_input4MIPS_c20181106.nc</solar_data_file>
 <solar_data_ymd>18500101</solar_data_ymd>
 <solar_data_type>FIXED</solar_data_type>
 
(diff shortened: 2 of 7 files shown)


cameronsmith1:
Am I correct that the data hasn't changed (just converted from netcdf 4 to 3)?

I recall that last time we did this it appeared that it was BFB.   Do we know if this is still true?

When we did something similar before, and went back later, it was easy to get confused.  Specifically, the different file dates makes it easy to jump to the wrong conclusion, ie to mistakenly assume that the data was different.  This is particularly likely to be an issue in this case as people will want to know how their configuration differs from the published CMIP6 simulation. Is there a way we can make it easy for users to figure out what is really going on?

rljacob:
Yes no data changes.  This was confirmed by running cprnc for the 2 files.

Have we done this before?   The new files were made a few months ago but AFAIK this is the first time use cases have been updated to use them.

cameronsmith1:
> Have we done this before? The new files were made a few months ago but AFAIK this is the first time use cases have been updated to use them.

Not for these specific files, but for other files.  See PR https://github.com/E3SM-Project/E3SM/pull/2174, issue https://github.com/E3SM-Project/E3SM/issues/1970, and several other related issues.

One of the most logical things to do (IMHO) when trying to do work based on an old configuration with a new code base is to build the same compset and check that the namelists are the same.  This change will produce many differences, and the user will need to figure out what has changed for every single file, or ignore the differences and potentially miss some other important change.

last time, I recall, a comment was added to the metadata in the new files, which is a help.   Did that happen here too?

Is there anything else we can do to avoid future confusion?

rljacob:
Yes there is metadata in each file clearly indicating that it was only a format change.

So PR #2174 changed these same files for 1950 on master.  This PR changes them for 1850 and 20TR on maint-1.0 and eventually master.   But the 1950 change uses files that end in c180205.  If that was just a format change, we should use those instead of the c181106 ones to further reduce confusion.

cameronsmith1:
> So PR #2174 changed these same files for 1950 on master. This PR changes them for 1850 and 20TR on maint-1.0 and eventually master. But the 1950 change uses files that end in c180205. If that was just a format change, we should use those instead of the c181106 ones to further reduce confusion.

That is my recollection, and if we can use those previous files that would certainly be an improvement.

I am pinging @PeterCaldwell and @golaz to make sure I am remembering correctly.

rljacob:
Now I'm having second thoughts.  There is more and better metadata in the c181106 files.   Compare the global attributes for each
c180205:
```
	:history = "Created on 05/25/2017 at Pacific Northwest National Lab. Contact: hailong.wang@pnnl.gov\n",
			"Converted from netcdf-4 to netcdf-3 by P. Caldwell using netcdf4_to_netcdf3.py on 2018-02-05\n",
			"" ;
```

c181106:'
```
		:history = "Tue Nov  6 10:42:34 2018: ncks -O -3 --glb E3SM_input_conversion=20181106: Format converted from netcdf4-classic to classic (netCDF3) for compatibility with PnetCDF reader. No data were changed /lcrc/group/acme/public_html/inputdata/atm/cam/chem/trop_mozart_aero/emis/DECK_ne30/cmip6_mam4_bc_a4_elev_1850-2014_c170525.nc inputdata/atm/cam/chem/trop_mozart_aero/emis/DECK_ne30/cmip6_mam4_bc_a4_elev_1850-2014_c181106.nc\n",
			"Created on 05/25/2017 at Pacific Northwest National Lab. Contact: hailong.wang@pnnl.gov" ;
		:E3SM_input_conversion = "20181106: Format converted from netcdf4-classic to classic (netCDF3) for compatibility with PnetCDF reader. No data were changed" ;
		:NCO = "4.6.7" ;

```
The c181106 files include the name of the file that was converted and adds 2 new global attributes.

cameronsmith1:
I agree that the new metadata is better.  I am not sure, though, whether this outweighs the confusion from an extra set of files.  If you do choose to go with the new files, can you also update the other compsets that use the 180205 files?

rljacob:
The 180205 files have been used in code that was checked in.  So we can't delete them.   Might as well just continue to use them instead of the never-used 181106 ones.

Is there a list somewhere of all the files @PeterCaldwell converted?

cameronsmith1:
We shouldn't delete the 180205 files.  But we could modify master so that the compsets switch to using the newer netcdf files.

I recall that @PeterCaldwell had a script for automatically finding and converting the netcdf4 files, and it would spit out the files it converted.  Would it be easier just to search the data archives for '180205'?

rljacob:
Closing this.  A new PR will replace it.

PR #3058, opened 2019-07-11
Update mpas-seaice for snicar and snow grain model

jonbob:
This PR brings in a new mpas-source submodule that only updates the mpas-seaice code, as well as the supporting scripts to match. The seaice has changes to:
* the snicar coding (MPAS-Dev/MPAS-Model PR #278);
* snow grain model  (bug fixes)

Because these features are off by default, this PR is BFB but does have a namelist change.

[NML]
[BFB]

### DIFF
--- a/components/mpas-seaice/bld/build-namelist
+++ b/components/mpas-seaice/bld/build-namelist
@@ -785,7 +785,7 @@ add_default($nl, 'config_ratio_C_to_N_proteins');
 
 add_default($nl, 'config_shortwave_type');
 add_default($nl, 'config_albedo_type');
-add_default($nl, 'config_use_snicar');
+add_default($nl, 'config_use_snicar_ad');
 add_default($nl, 'config_visible_ice_albedo');
 add_default($nl, 'config_infrared_ice_albedo');
 add_default($nl, 'config_visible_snow_albedo');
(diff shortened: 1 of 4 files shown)


rljacob:
The SNICAR part his covered by this code review document:
https://acme-climate.atlassian.net/wiki/spaces/ED/pages/976617535/C1+SNICAR-AD+Shortwave+Radiation+for+Sea+Ice+and+Land

Where is the code review document for the snow grain model?

rljacob:
Why is this v2 feature turned off?   If we want in v2, it should be on. Otherwise, @golaz has to remember to turn it on when tuning.

jonbob:
@rljacob - my understanding is that it's turned off until it goes through the post-v2 tuning stage. After that, a decision will be made about having it on by default and a separate PR would be made.

jonbob:
@rljacob - the snow grain model changes are minor bug fixes found by @njeffery during testing. We had made a single PR because they both had been pushed to the head of the seaice repo, but I can redo if necessary.

rljacob:
Thanks for clarifying the "snow grain model".

rljacob:
Turns out there are 2 features in this PR.  They both have to have their code review documents approved.

rljacob:
Both code review documents for this PR have been approved.

jonbob:

PR #3059, opened 2019-07-11
Fix bug in setting gases above model top for RRTMGP

brhillman:
Fix bug in setting the gases above model top. An extra level is added above the model top in the radiation code to handle heating between the top of the model and the top of the atmosphere. The code in question here was supposed to copy the gases from the top-most model level into the extra level above the model top, but instead we accidentally copied the *second* model level.

[non-BFB] only for the RRTMGP tests.


brhillman:
Differences in solar and longwave heating from a 1-month simulation with ne4, with and without the fix (zonal mean profiles):
qrs_fix_gas
qrl_fix_gas

wlin7:
The change looks good. @singhbalwinder , can you take a quick look and approve if you see no problem? Thanks.

wlin7:
Merged to next.

wlin7:
Merged to master. Request to bless RRTMGP test for sandiatoss3.

PR #3060, opened 2019-07-11
Turn eddyProductVariables AM on by default

jonbob:
This PR turns the mpas-ocean eddyProductVariables analysis member on by default for all grids. It does not impact any results, just simply adds an output stream. For some ultra-low resolutions, this has previously been turned off, so there may be namelist changes for those tests.

[NML]
[BFB]

### DIFF
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -857,10 +857,6 @@
 
 <!-- AM_eddyProductVariables -->
 <config_AM_eddyProductVariables_enable>.true.</config_AM_eddyProductVariables_enable>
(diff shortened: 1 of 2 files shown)


mark-petersen:
Thank you @jonbob and @vanroekel. We should have done this originally, but at least we are changing it now. Since low-res does not matter, let's just make it simple and only have a single line:
```
<config_AM_eddyProductVariables_enable>.true.</config_AM_eddyProductVariables_enable>
```
and no exceptions.

Are the product variables written in the monthly time average by default? They probably are.

jonbob:
Modified to have on by default for all grids, though that make cause namelist diffs for tests with low-res. Also fixed the output interval so that turning this on actually creates output. @mark-petersen, I did check  and the eddy product variables are written in the monthly time average by default.

jonbob:
merged to next

jonbob:
merged to master and NML DIFFs blessed

PR #3063, opened 2019-07-12
Change dust emission size distribution and shortwave absorption props

yfenganl:
Change dust emission size distribution to emit more coarse dust particles. This is consistent
with the recent emission measurement data. Change shortwave absorption properties to
reduce the absorption of dust particles. This results in better agreement with the AERONET
surface measurements of aerosol absorption optical depth.

Change dust_emis_fact for default and all typically used cases including CMIP6 to tune
dust emissions to match the aerosol optical depth constraint with the updated emission
size distribution above.

[CC]
[NML]


yfenganl:
the design document is at:
https://acme-climate.atlassian.net/wiki/spaces/ED/pages/1003520636/NA2+Improved+Dust+Aerosol+Physics+Design+Document

singhbalwinder:
Pushed to next

Issue #2983, opened 2019-06-07
Model crashes when online interpolation is turned on


singhbalwinder:
The current master code crashes when online interpolation is turned on via namelist option:
`interpolate_output = .true.`

The model crashes with error:
```
105: forrtl: severe (408): fort: (3): Subscript #1 of the array CAM_GRIDS has value -1 which is less than the lower bound of 1
105: e3sm.exe           0000000000A7BC76  cam_grid_support_        2207  cam_grid_support.F90
105: e3sm.exe           00000000009189CC  cam_history_mp_h_        3711  cam_history.F90
105: e3sm.exe           000000000098E161  cam_history_mp_ws        4629  cam_history.F90
```
I tracked it down and found out that `CAM_GRIDS`'s index `gridind` is not initialized for procs >96 for ne4 grid. It has a value of -1 (default value), when it it not initialized. This explains why the model works when run with 1 or 2 nodes on Anvil (36 cores per node). As soon as we ask for 3 or more nodes, it fails. Anvil asks for 4 nodes by default for ne4 grid therefore it fails on Anvil.

The reason `gridind` is not initialized is due to `par%dynproc` which is FALSE for all procs >96. `par%dynproc` is set in `components/homme/src/share/parallel_mod.F90`. I do not know this code well so I would appreciate if anybody can suggest me how should we resolves this issue. Tagging @AaronDonahue as he has most recently added new features to this part of the code.

I ran into another issue when I ran the model with interpolation turned on for PE layout using very few processors. The code fails with the following error:
```
4: forrtl: severe (408): fort: (3): Subscript #1 of the array BBUFFER has value -1 which is less than the lower bound of 1
4: e3sm.exe           000000000411FA4C  interp_mod_mp_wri         303  interp_mod.F90
```
I was able to get past it by commenting out the following OpenMP directive in  `components/cam/src/dynamics/se/interp_mod.F90`:
`!$omp parallel do private (ie, bpter, icol)`

The code was failing with only some PE layouts while working with others so I thought it might be an OpenMP or MPI issue. I didn't look into it further. Tagging @amametjanov and @worleyph if they have any insights.

worleyph:
For the first error, @mt5555 will have to give the definitive answer, and @amametjanov is also familiar with this aspect of the code, but the dynamics does not support using more processes than elements. The interpolate_output logic probably should not be trying to use processes that do not contain an element (or am I not understanding the error condition)?

rljacob:
[...]

Issue #2959, opened 2019-05-30
Shape mismatch error in mesh_mod.F90


singhbalwinder:
While running `SMS_D_Ln5.conusx4v1_conusx4v1.FC5AV1C-L.compy_intel` test on Compy, which uses newest Intel compiler (`intel/19.0.3`), the model crashes with the following error:
```forrtl: severe (408): fort: (33): Shape mismatch: The extent of dimension 1 of array A_CORNER_ELEMS is 4 and the corresponding extent of array ELEM_NEIGHBOR is 2

Image              PC                Routine            Line        Source
e3sm.exe           0000000009202BA6  Unknown               Unknown  Unknown
e3sm.exe           0000000000B6F338  mesh_mod_mp_find_        1044  mesh_mod.F90
e3sm.exe           0000000000B7F982  mesh_mod_mp_meshc        1566  mesh_mod.F90
e3sm.exe           00000000033C94EB  prim_driver_base_         326  prim_driver_base.F90
e3sm.exe           00000000033C83C6  prim_driver_base_         106  prim_driver_base.F90
e3sm.exe           0000000002590FEA  dyn_comp_mp_dyn_i         153  dyn_comp.F90
e3sm.exe           0000000000AAD11F  inital_mp_cam_ini          31  inital.F90
e3sm.exe           0000000000864D73  cam_comp_mp_cam_i         159  cam_comp.F90
e3sm.exe           00000000008304DC  atm_comp_mct_mp_a         312  atm_comp_mct.F90
e3sm.exe           0000000000469CC7  component_mod_mp_         257  component_mod.F90
e3sm.exe           00000000004295BB  cime_comp_mod_mp_        1332  cime_comp_mod.F90
e3sm.exe           0000000000460797  MAIN__                    122  cime_driver.F90
e3sm.exe           00000000004176A2  Unknown               Unknown  Unknown
libc-2.17.so       00002B67E4C753D5  __libc_start_main     Unknown  Unknown
e3sm.exe           00000000004175A9  Unknown               Unknown  Unknown
```

I googled it and this webpage (http://diamondized16.rssing.com/chan-9066188/all_p13.html) suggests that this error can be avoided either via a flag (`-warn shape`) which turns it into a warning or may be via revising the code. I would like to get your suggestions on how should we proceed.

singhbalwinder:
@mt5555 : I have assigned this to you temporarily but please feel free to reassign.

oksanaguba:
[...]

PR #3064, opened 2019-07-12
fix EAM compiler errors and better abort message

mt5555:
avoid compiler error with assumed shape array
abort if ntasks_atm>dyn_pes and interpolated output turned on
remove unused variables from elem struct

Fixes E3SM-Project/E3SM#2983
Fixes E3SM-Project/E3SM#2959

[BFB]

### DIFF
--- a/components/cam/src/dynamics/se/interp_mod.F90
+++ b/components/cam/src/dynamics/se/interp_mod.F90
@@ -1,12 +1,12 @@
 module interp_mod
   use cam_logfile,         only: iulog
   use shr_kind_mod,        only: r8 => shr_kind_r8
-  use dimensions_mod,      only: nelemd, np, ne
+  use dimensions_mod,      only: nelem,nelemd, np, ne
   use interpolate_mod,     only: interpdata_t
   use interpolate_mod,     only: interp_lat => lat, interp_lon => lon
   use interpolate_mod,     only: interp_gweight => gweight
   use dyn_grid,            only: elem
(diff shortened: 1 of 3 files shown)


mt5555:
sorry, I dont know why I closed this - must have clicked something by mistake.

components/cam/src/dynamics/se/interp_mod.F90
```diff
@@ -69,6 +69,11 @@ subroutine setup_history_interpolation(interp_ok, mtapes, interp_output,    &
     type(horiz_coord_t), pointer       :: lon_coord
     character(len=max_hcoordname_len)  :: gridname
 
+    if(any(interp_output)) then
+       if (.not. par%dynproc) &
+       call endrun('setup_history_interpolation: interpolated output not supported if atm_ntasks>dyn_npes')
```
singhbalwinder: I think it will be useful if we print out atm_ntasks and dyn_npes values in the error message so that the user knows what needs to be changed.

mt5555: good idea - will fix.

mt5555:
documenting test results:   

Ran SMS_P108x2.ne4_ne4.FC5AV1C-L, and then added "interpolate_output=.true." to user_nl_cam.  code then aborted with endrun, with descriptive error message in log file, as expected.

singhbalwinder:
Merged to next

PR #3065, opened 2019-07-12
bugfix in NH heating term

mt5555:
Fixes wrong index to average w at the top. Found by Luca Bertagna.

[BFB] except for EAM theta NH tests and HOMME standalone NH tests

### DIFF
--- a/components/homme/src/theta-l/prim_advance_mod.F90
+++ b/components/homme/src/theta-l/prim_advance_mod.F90
@@ -1284,7 +1284,7 @@ contains
              pnh,exner,temp_i,caller='advance_hypervis')
         
         do k=1,nlev
-           k2=max(k,nlev)
+           k2=min(k+1,nlev)
            if (theta_hydrostatic_mode) then
               heating(:,:,k)= (elem(ie)%state%v(:,:,1,k,nt)*vtens(:,:,1,k,ie) + &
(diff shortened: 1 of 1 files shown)


components/homme/src/theta-l/prim_advance_mod.F90
```diff
@@ -1284,7 +1284,7 @@ subroutine advance_hypervis(elem,hvcoord,hybrid,deriv,nt,nets,nete,dt2,eta_ave_w
              pnh,exner,temp_i,caller='advance_hypervis')
         
         do k=1,nlev
-           k2=max(k,nlev)
+           k2=min(k+1,nlev)
```
oksanaguba: Is average supposed to be
lev1+lev2, lev2+lev3,...,level nlev + level nlevp ?
I read this as 
k=1 => lev1+lev2, k=2 => lev2 + lev 3,...., k=nlev => level nlev + level min(nlev+1, nlev) = level nlev + level nlev?

Why not state k2 := k+1 ?

mt5555: the viscosity tendencies are not computed at k=nlevp. so at the surface, we cant average, and just use the nlev value.

oksanaguba:
Pushed to next, tested in next 2 commits before that, all as expected.

PR #3066, opened 2019-07-12
Fixed syntax error in dp_coupling

whannah1:
Fixes erroneous line continuation characters caught by PGI compiler.

[BFB]


whannah1:
Interestingly, this was not caught by intel during development, but then subsequently caught by PGI when testing on Summit.

sarats:
We should get some PGI tests in our periodic testing workflow. Such tests can't be run on Summit for a variety of reasons (queue waits, lack of CI support, lack of single submit etc.)

I asked Compy support to install PGI **community edition** along with requisite dependencies. I think that would be a good platform for such testing. The intention is just checking for obvious PGI problems (syntax errors etc.), not for GPU or POWER architecture tests.

rljacob:
Yes we will be doing regular PGI testing on compy.

PR #3075, opened 2019-07-17
Make create_dirs more robust when multiple threads are running

jgfouca:
[BFB]


rljacob:
What did this fix?

jgfouca:
@rljacob :
                # In a multithreaded situation, we may have lost a race to create this dir.
                # We do not want to crash if that's the case.

PR #3076, opened 2019-07-18
add gnu-5.3.0 to anvil-intel environment

sarich:
Allows homme to link with compatible glibc/libstdc++ libraries on anvil


rljacob:
Looks like it worked!

PR #3077, opened 2019-07-18
Add transient test to production for maint1.0

rljacob:
Add test of transient water cycle fully coupled case to production suite for maint1.0.  This will read in a few different input files compared to the PI case.

[BFB]


cameronsmith1:
Adding the CMIP6 transient is a good idea.   Thanks!

PR #3079, opened 2019-07-19
add +mkl to the anvil config

### LINKED ISSUE #2979: 18x4 ne4 F case on Anvil producing NaNs
@singhbalwinder noticed that using 4 threads per process on Anvil was leading to some failures. I determined that using an 18x4 decomposition (thus using 2 nodes) for 

     --res ne4_ne4 --compset FC5AV1C-L --compiler intel

fails approximately 70% of the time. When it fails, it either reports NaNs in reprosum, e.g.

     SHR_REPROSUM_CALC: Input contains  0.10000E+03 NaNs and  0.00000E+00 INFs on process       0
      ERROR: shr_reprosum_calc ERROR: NaNs or INFs in input

 right after the output

      nstep, te        1   0.32979826924921732E+10   0.32981382047614088E+10   0.21501496260284258E-02   0.98515947862105517E+05

or, more rarely, 

      ERROR: lapack_wrap.F90: Singular matrix detected in band_solve

right after 

      nstep, te        2   0.26361939827774453E+10   0.32982129229854422E+10   0.91980619458198216E+01   0.98035837897079109E+05

If the simulation gets beyond these two failure points (e.g., through timestep 3), it always completes successfully. A decomposition of 9x4 (i.e. one node) always succeeds. I do not know if this means that the MPI library is involved or not.

Adding debug statements identifies these NaNs at various points in the simulation, but the process of printing them out can also "move them around", eliminating them from one field or elements of that field, but showing up then in other locations, indicating a memory issue? Increasing the OMP stack size had no apparent effect.

I was able to eliminate the problem by adding a critical section around the second large loop in the routine parcel_dilute in zm_conv.F90, starting at line 3667:

     !$omp critical
     do k = pver, msg+1, -1
        do i=1,ncol
     ...
        end do ! Loop for columns
     
     end do  ! Loop for vertical levels.
     !$omp end critical

Starting the critical section just inside of the k-loop did not eliminate the problem however. I see nothing about the code in this loop that appears suspect, though it is a little complicated. Perhaps this is a compiler error? The sensitivity to the thread count (which means that threads for one process per node are on both sockets) might indicate an architecture issue, but this should still work. The NaNs occur on most/all processes, and this critical section should be a local to a process, so the weird thread layout for a couple of processes should be irrelevant here?

So, I do not have a real diagnosis, and the 'fix' is not acceptable, but perhaps someone else has some insight, or suggestions as to what to do.

mt5555:
add +mkl to the anvil config
fixes #2979

[BFB]

### DIFF
--- a/cime/config/e3sm/machines/config_machines.xml
+++ b/cime/config/e3sm/machines/config_machines.xml
@@ -1169,7 +1169,7 @@
       </modules>
       <modules compiler="intel">
         <command name="add">+intel-17.0.0</command>
-        <command name="add">+mkl</command>
+        <command name="add">+mkl-11.0.5.192</command>
         <command name="add">+netcdf-c-4.4.1-f77-4.4.4-intel-17.0.0-serial</command>
       </modules>
       <modules compiler="intel" mpilib="mvapich">
--- a/cime/config/e3sm/machines/config_machines.xml
+++ b/cime/config/e3sm/machines/config_machines.xml
@@ -1168,6 +1168,7 @@
       </modules>
       <modules compiler="intel">
         <command name="add">+intel-17.0.0</command>
+        <command name="add">+mkl</command>
         <command name="add">+netcdf-c-4.4.1-f77-4.4.4-intel-17.0.0-serial</command>
       </modules>
       <modules compiler="intel" mpilib="mvapich">
--- a/cime/config/e3sm/machines/config_machines.xml
+++ b/cime/config/e3sm/machines/config_machines.xml
@@ -1148,6 +1148,7 @@
     <mpirun mpilib="default">
       <executable>srun</executable>
       <arguments>
+        <arg name="kill-on-bad-exit">-K</arg> 
         <arg name="num_tasks"> -l -n {{ total_tasks }} </arg>
         <arg name="binding">--cpu_bind=cores</arg>
         <arg name="thread_count">-c $SHELL{if [ FALSE = `./xmlquery --value SMP_PRESENT` ];then echo 1;else echo $OMP_NUM_THREADS;fi}</arg>


mt5555:
@amametjanov : I was going to ask if #2979 was fixed yet, but thought it would be almost as easy to submit this PR.  is this what is needed for #2979 ?

oksanaguba:
Could we also change line <executable>srun</executable> to <executable>srun -K</executable> in  cime/config/e3sm/machines/config_machines.xml for anvil?

mt5555:
great idea.  Following the sooty machine config, I think it should be: 
 <arguments>
        <arg name="kill-on-bad-exit"> -K </arg>
...

singhbalwinder:
Thanks @mt5555 for working on it. Do we need to add a version number for mkl? Sometimes the defaults change underneath which changes model's answers....and it can sometimes be hard to figure out. This is one of the machines where we maintain baselines, so it might help down the road.

sarich:
+mkl on anvil currently loads +mkl-11.0.5.192
I agree we should make it explicit.

sarich:
There's some build failures in e3sm_developer, they build ok on master:

```
FAIL ERS.f09_g16_g.MALISIA.anvil_intel MODEL_BUILD time=91
FAIL ERS_Ld5.T62_oQU120.CMPASO-NYF.anvil_intel MODEL_BUILD time=86
FAIL SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.anvil_intel MODEL_BUILD time=95
FAIL SMS.f09_g16_a.IGCLM45_MLI.anvil_intel MODEL_BUILD time=341
FAIL SMS_P12x2.ne4_oQU240.A_WCYCL1850.anvil_intel.allactive-mach_mods MODEL_BUILD time=364
```


```
gmake[5]: Entering directory `/lcrc/group/acme/sarich/acme_scratch/anvil/SMS.f09_g16_a.IGCLM45_MLI.anvil_intel.20190719_130201_9swxko/bld/glc/source/tools/registry'
...
../../external/ezxml/ezxml.o: In function `ezxml_decode':
ezxml.c:(.text+0x37ce): undefined reference to `_intel_fast_memmove'
ezxml.c:(.text+0x3a7f): undefined reference to `_intel_fast_memmove'
ezxml.c:(.text+0x3cd6): undefined reference to `_intel_fast_memmove'
ezxml.c:(.text+0x3d8c): undefined reference to `_intel_fast_memmove'
../../external/ezxml/ezxml.o: In function `ezxml_set_attr':
ezxml.c:(.text+0x6170): undefined reference to `_intel_fast_memmove'
../../external/ezxml/ezxml.o:ezxml.c:(.text+0x61b9): more undefined references to `_intel_fast_memmove' follow
parse.o: In function `main':
parse.c:(.text+0x1a): undefined reference to `__intel_new_feature_proc_init'
gmake[5]: *** [parse] Error 1
gmake[5]: Leaving directory `/lcrc/group/acme/sarich/acme_scratch/anvil/SMS.f09_g16_a.IGCLM45_MLI.anvil_intel.20190719_130201_9swxko/bld/glc/source/tools/registry'
```

singhbalwinder:
Thanks @sarich for running the tests. Do all other tests run fine with a PASS?

mt5555:
I added the version number to mkl.  but I dont know how to fix those compiler errors.

amametjanov:
@sarich, the build error is probably due to missing `-mkl` flag while linking `parse` executable:
```
parse: $(OBJS)
        $(CC) $(CPPFLAGS) $(CFLAGS) $(EZXML_PATH)/ezxml.o  -I$(EZXML_PATH) -o $@ $(OBJS)
```

A workaround is to use pre-built `parse` binary (e.g. built from master branch) by defining MPAS_TOOL_DIR env-var in config_machines: e.g. from Theta
```
<env name="MPAS_TOOL_DIR">/projects/ccsm/acme/tools/mpas</env>
```
See https://github.com/MPAS-Dev/MPAS-Model/blob/6a557fb6d714dcf03bd35ec065117542d97f8c5a/src/tools/Makefile

sarich:
@singhbalwinder Yes, all the other e3sm_developer tests run ok

sarich:
@amametjanov Thanks, I'll take a look at the parse build.

mt5555:
@sarich : based on Az's comments in #2979,  I think we need to delay this PR, since we dont want to use such an old (circa 2013) version of mkl.  But we haven't confirmed the newer versions of mkl fix the original issue.

mt5555:
@oksanaguba : regarding sbatch -K:  
on Anvil, with -K,  I get "sbatch: invalid option -- 'K'"

There is a -k option, but this doesn't seem what we want:  

-k, --no-kill
              Do not automatically terminate a job if one of the nodes it has been
              allocated fails.  The user  will  assume  the  responsibilities  for
              fault-tolerance  should  a node fail.  When there is a node failure,
              any active job steps (usually MPI jobs) on  that  node  will  almost
              certainly  suffer a fatal error, but with --no-kill, the job alloca-
              tion will not be revoked so the user may launch new job steps on the
              remaining nodes in their allocation.

              By  default  Slurm  terminates the entire job allocation if any node
              fails in its range of allocated nodes.

oksanaguba:
I use it in srun: srun -K ... , otherwise i had runs hang occupying nodes till i noticed they actually crashed hours ago.

mt5555:
thanks - sorry I confused srun and sbatch.

rljacob:
What is the status of this PR?  If the changes are harmless, go ahead and merge it.

mt5555:
based on issue #2979,  I think this is not the right solution and I will close this PR

PR #3083, opened 2019-07-22
switch mpi_allreduce to use MPI_IN_PLACE

### LINKED ISSUE #3082: maint-1.0 crashes with A_WCYCL "L" configuration on Compy
On compy, with the maint-1.0 branch, this case

./create_newcase -case /qfs/people/wolf966/E3SM-cases/20190717.test3.compy -compset A_WCYCL1850S_CMIP6 -mach compy -res ne30_oECv3_ICG -compiler intel --project e3sm -pecount L

will crash - with errors indicating memory corruption in HOMME's dof_mod.F90:

*** Error in `/compyfs/tayl519/e3sm_scratch/w1/bld/e3sm.exe': corrupted size vs. prev_size: 0x0000000038dc59d0 ***

With the following workarounds, (and changing I/O type to netcdf because the current maint-1.0 will try and fail to read a few NETCDF4 files), the code will run and get to the point where it reads the restart files and then stops (because I dont have the correct restart files)

Any one of these will let the code run:

1.  Change the two arrays used in mpi_allreduce from heap arrays to stack arrays.  E3SM master happens to have done this (for completely unrelated reasons) and hence E3SM master avoids this issue.  

2.  Break the mpi_allreduce (which is of length 5400) into smaller arrays.  doing in mpi_allreduce on 1:5000 followed by 5001:5400 works.  Also doing 5400 mpi_allreduces, each of lenght 1 works.   

3.  use the MPI_IN_PLACE feature of mpi_allreduce.

4.  Running the "M" configuration instead of "L" (as reported by @jonbob )

So the main thing that doesn't work is calling mpi_allreduce() when the atmosphere is using 2700 tasks, and with large arrays (> 5000) that happens to have been allocated instead of
placed on the stack.

Behavior is very strange, but since there are so many configurations that work, it does suggest the bug might not be in the E3SM code but could be in Intel MPI

mt5555:
workaround for an array corruption issue possibly caused by mpi_allreduce

fixes #3082

[BFB]

### DIFF
--- a/components/homme/src/share/dof_mod.F90
+++ b/components/homme/src/share/dof_mod.F90
@@ -278,7 +278,7 @@ contains
 
   subroutine SetElemOffset(par,elem,GlobalUniqueColsP)
 #ifdef _MPI
-     use parallel_mod, only : mpi_sum
+     use parallel_mod, only : mpi_sum, MPI_IN_PLACE
 #endif
      type (parallel_t) :: par
      type (element_t) :: elem(:)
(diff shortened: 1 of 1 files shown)


jonbob:
Merged to maint-1.0

PR #3084, opened 2019-07-22
Update default compy layouts to reflect slower MPAS-O on maint-1.0

jonbob:
This PR updates the default "L" and "XL" pecounts for B-cases on compy from those optimized for master. The maint-1.0 branch still has shorter timesteps for mpas-o so the ocean requires more processors than master does for efficient layouts.

[BFB]

### DIFF
--- a/cime/config/e3sm/allactive/config_pesall.xml
+++ b/cime/config/e3sm/allactive/config_pesall.xml
@@ -7781,13 +7781,13 @@
         </rootpe>
       </pes>
       <pes compset=".*CAM5.+CLM45.+MPASCICE.+MPASO.+MOSART.+" pesize="L">
-        <comment> -compset A_WCYCL* -res ne30_oEC* on 80 nodes pure-MPI </comment>
+        <comment> -compset A_WCYCL* -res ne30_oEC* on 92 nodes pure-MPI </comment>
         <ntasks>
           <ntasks_atm>2700</ntasks_atm>
           <ntasks_lnd>2700</ntasks_lnd>
           <ntasks_rof>2700</ntasks_rof>
           <ntasks_ice>2700</ntasks_ice>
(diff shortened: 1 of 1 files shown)


amametjanov:
These pecounts on maint-1.0 show:
* on  92 nodes -- 11.8 sypd
* on 185 nodes -- 19.9 sypd

Just for comparison, pecounts on master:
* 80 nodes -- 10.9 sypd
* 160 nodes -- 16.9 sypd

jonbob:
Thanks @amametjanov !

jonbob:
merged to maint-1.0

golaz:
@amametjanov : any reason why the pe layouts for master should be slower than maint-1.0? I understand that the layout for maint-1.0 and master must be different because of the slower time step on maint-1.0, but I don't understand my the master layouts would be slower.

amametjanov:
Diffs in configuration: I ran with CPL history turned on (`HIST_OPTION=$STOP_OPTION` with $case.cpl.hi.0001-01-06-00000.nc at the end of a 5-day bench run), Jon ran with `HIST_OPTION=never`. After configuring production IO for multi-year runs, pecount may need to be adjusted: user_nl_* files were empty in both runs.

After turning off HIST_OPTION, master shows:
* 80 nodes -- 11.6 sypd
* 160 nodes -- 21.4 sypd

And just to log the time-step diffs:
* master has `config_dt = '00:30:00'`, OCN runs 17 sypd @12nodes, 34 sypd @25nodes, 54 sypd @50nodes
* maint-1.0 has `config_dt = '00:10:00'`, OCN runs 12 sypd @24nodes, 20 sypd @50nodes

jonbob:
Thanks for tracking down and explaining the differences, @amametjanov . That all makes plenty of sense, and good to know master is as fast as maint-1.0 under more matched tests

PR #3092, opened 2019-07-29
Update CIME to ESMCI cime5.8.7

jgfouca:
Update CIME to ESMCI cime5.8.7

Squash merge of jgfouca/branch-for-to-acme-2019-07-29

Features:

* A new tool generate_cylc_workflow.py is added. This tool will create a cylc 7.8.3 compatible suite.rc file in the case directory. The user sets the parameters in the case for the span of a submitted run and the number of resubmits, sets up and builds the case and then generates the workflow file. The user can either set the RESUBMIT variable using xmlchange or use --cycles argument to generate_cylc_workflow.py to set the number of times the workflow is repeated.

* New workflow XML files, replaces some of the stuff that was in env_batch
* Several updates and fixes for MOM6 in cime xml files, drivers, and dice.
* Adds a --ensemble option to create_clone allowing the user to generate an ensemble of cases cloned from --clone

Bug Fixes:
* ftp protocol won't make 0-size files anymore

Other:
Driver namelist change:   from eps_ogrid = 0.01 to eps_ogrid = 0.1

[BFB]
[NML]

### DIFF
--- a/cime/ChangeLog
+++ b/cime/ChangeLog
@@ -1,6 +1,73 @@
 ======================================================================
 
 Originator: Chris Fischer 
+Date: 7-19-209
+Tag: cime5.8.7
+Answer Changes: Climate changing for CAM ne120, ne120pg3, ne0CONUS grids
+Tests: scripts_regression_tests
+Dependencies:
+
+Brief Summary:
+    - Fix SE to gx1v7 domain and mapping files.
+    - Updates and fixes for MOM6 in cime.
+    - Maint 5.6 merge.
+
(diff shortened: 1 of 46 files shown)


jgfouca:
testing passed, squash-merging to next.

rljacob:
Jim, I added some things to the description but they didn't make it in to the next commit message.

jgfouca:
@rljacob , I must have forgotten to refresh this page when I copy/pasted. I'll make sure master is correct.

PR #3094, opened 2019-07-30
Add modifications for HOMME in "library mode"

bartgol:
Adds modifications for infrastructure in homme/hommexx, for 'library mode', cmake and build fixes. Adds bugfix in hommexx jw_baro initialization.

[BFB] because there are no nightlies with hommexx.


bartgol:
I added a bug fix to the kokkos targets related to edge buf not being inited. Despite doing MPI in C++, we still need edgebufs for tests initialization routines; if this turns out being a memory problem, we can try to clean up these buffers after the test initialization.

components/homme/src/preqx_kokkos/prim_driver_mod.F90
```diff
@@ -98,32 +101,60 @@ end subroutine initialize_hommexx_session
     ! ==================================
     call prim_init1_elem_arrays(elem,par)
 
+    ! ==================================
+    ! Initialize the buffers for exchanges
+    ! ==================================
+    call prim_init1_buffers(elem,par)
```
oksanaguba: This will call 
```
    call initEdgeBuffer(par,edge_g,elem,max((qsize+3)*nlev+2,6*nlev+1))
```
which I think can be replaced with just nlev size buffer. Otherwise, this big edge_g buffer sits in the memory for the whole run (unless you freed it and I did not notice), but it is only needed for 1 test init. Maybe, freeing it after tests init is also an option? unless it introduces even more infrastructure diffs.

bartgol: Good call. If the init only needs a small buffer, I'll go ahead and change this to an ad-hoc call to initEdgeBuffer. Free-ing it after initialization is also an option, but it may be more code edits than simply allocate an nlev-wide buffer.

bartgol:
I added a fix as suggested by @oksanaguba , to minimize the memory footprint of the edge buffer. I also did a quick fix to generate valid namelists in the kokkos target folders inside test_exec (not for testing, just used for quick checks during development).

bartgol:
Standalone homme tests pass on my machine, so I think this is ready to be merged into next.

bartgol:
@jgfouca I ended up having to do another couple of mods to homme/cmake/Kokkos.cmake (things were working because I don't know why). Basically I now handle three cases:

1) an installation path is provided: great, use that
2) no installation path is provided:
2.1) if E3SM_INTERNAL_KOKKOS_ALREADY_BUILT evaluates to true, use internal kokkos, but _don't_ build it
2.2) otherwise, use internal kokkos _and_ build it.

The issue is that if you end up adding the kokkos subdir twice, you get a cmake error.

jgfouca:
@bartgol , looks good.

bartgol:
Testing details: the tests are BFB (both branch against master, as well as cxx vs f90) with a Kokkos::Serial build. 

Config file:
```
BASELINE_DIR=${WORK_DIR}/e3sm/e3sm-build/homme/gcc/opt/master
# BASELINE_DIR=${WORK_DIR}/hommexx/hommexx-build/gcc/opt/master

rm -rf CMakeFiles
rm -f  CMakeCache.txt

cmake -Wno-dev \
  -D CMAKE_VERBOSE_MAKEFILE:BOOL=OFF                            \
  -D CMAKE_CXX_COMPILER:STRING="mpicxx"                         \
  -D CMAKE_C_COMPILER:STRING="mpicc"                            \
  -D CMAKE_Fortran_COMPILER:STRING="mpifort"                    \
  -D CMAKE_C_FLAGS:STRING="-w"                                  \
  -D CMAKE_CXX_FLAGS:STRING="-Wno-unknown-pragmas"              \
  -D CMAKE_EXE_LINKER_FLAGS:STRIN=-ldl                          \
  -D ADD_C_FLAGS:STRING="-ffp-contract=off"                     \
  -D ADD_CXX_FLAGS:STRING="-ffp-contract=off"                   \
  -D ADD_Fortran_FLAGS:STRING="-ffp-contract=off -w"            \
  -D NETCDF_DIR:PATH=${NETCDF_C_ROOT}                           \
  -D WITH_PNETCDF:BOOL=FALSE                                    \
  -D BUILD_HOMME_SWEQX:BOOL=OFF                                 \
  -D BUILD_HOMME_THETA:BOOL=OFF                                 \
  -D BUILD_HOMME_PREQX:BOOL=OFF                                 \
  -D BUILD_HOMME_PREQX_KOKKOS:BOOL=ON                           \
  -D BUILD_HOMME_THETA_KOKKOS:BOOL=ON                           \
  -D AVX_VERSION:STRING=0                                       \
  -D HOMMEXX_VECTOR_SIZE:STRING=4                               \
  -D HOMME_FIND_BLASLAPACK:BOOL=ON                              \
  -D BLAS_LIBRARIES:STRING=${BLAS_ROOT}/lib64/libblas.so        \
  -D LAPACK_LIBRARIES:STRING=${LAPACK_ROOT}/lib64/liblapack.so  \
  -D HOMME_BASELINE_DIR:PATH=${BASELINE_DIR}                    \
  -D USE_NUM_PROCS:STRING=4                                     \
  -D USE_MPI_OPTIONS:STRING="--bind-to core"                    \
  ../../../../../e3sm-src/branch/components/homme
```
Libs/modules:
- gcc 7.2.0 (sems)
- openmpi 1.8.7 (sems)
- netcdf 4.4.1 (sems)
- hdf5 1.8.12 (sems)
- lapack 3.8

Note: in order to compile and run master, the same mod as in commits 2e6c0ea and 9e30fa0 was added in master, to allow kokkos targets to run.

Next I will post results for the GPU build on V100.

oksanaguba:
How cxx vs F ran if build_preqx=off?

bartgol:
> How cxx vs F ran if build_preqx=off?

If preqx_kokkos is on, Homme automatically turns preqx on, in order to be able to do comparison. Sorry, the config file is a bit misleading on that...

oksanaguba:
Detailed comment on this PR:

Modifications include:

introduction of a Session struct (with static members). This struct
allow to set/detect how Hommexx should behave. For now, the only
member is m_throw_instead_of_abort, which can be set to tell homme
to throw instead of aborting. This helps if you use Homme from another
app, since aborting causes to call Kokkos::finalize, which can be
catastrophic if the host app still has allocated views.
Follow-up to 1: the runtime_abort function is now throwing rather than
aborting if Session::m_throw_instead_of_abort is true.
addition of some capabilities to BoundaryExchange (more register method,
for runtime dimensions vector/tensor fields) and to Connectivity (skip
some sanity checks, useful for unit tests on 'fake' artificial meshes
that do not meet all sanity checks).
split the implementation of prim_init2 in preqx_kokkos into multiple
subroutines. This can help a host app to tamper with Hommexx's internal
data structures to achieved desired behaviors. One important example:
replace the views in Homme::Elements with unmanaged views built from
pointers managed in a host app before the Elements structure is
passed to all the functors in Homme (since at that time, functors
copy the Elements structure (shallow copy), so further mods to the
version in Homme::Context will not be propagated to the functors).
Setting the CUDA_BUILD macro in the Hommexx_config.h file, which is
visible even if Homme is used as library, while the CMake-style solution
of adding -DCUDA_BUILD to CMAKE_CXX_FLAGS seems to be flaky.
Including Homme's SetCompilerFlag script by full path, to avoid name
clashes.
Exporting HOMME_[C|CXX|Fortran]_FLAGS in the parent scope (compared to
Homme's root directory). The host app can decide what to do with
Homme's flags. E.g., whether to prepend or append them to its own flags.

components/homme/src/share/parallel_mod.F90
```diff
@@ -176,20 +170,33 @@ function initmp(npes_in,npes_stride) result(par)
...
+#ifdef _MPI
+#include <mpif.h>
     if (MPI_DOUBLE_PRECISION==20 .and. MPI_REAL8==18) then
        ! LAM MPI defined MPI_REAL8 differently from MPI_DOUBLE_PRECISION
        ! and LAM MPI's allreduce does not accept on MPI_REAL8
        MPIreal_t    = MPI_DOUBLE_PRECISION
     else
        MPIreal_t    = MPI_REAL8
     endif
-
```
oksanaguba: Why were these lines deleted? Without them MPI2_real is not defined. It is used now in theta-l runs, so, I am not sure how tests got passed.

bartgol: My bad. When I split the method in two I copied and pasted, and I must have forgotten this line. I will bring it back now.

oksanaguba:
Homme and eam_theta tests pass on skybridge against today's next . On anvil, there are a lot of messages
```
[cli_11]: aborting job:
Fatal error in MPI_Allreduce:
Invalid datatype, error stack:
MPI_Allreduce(937): MPI_Allreduce(sbuf=0x2b7a5dd0f7e0, rbuf=0x7ffde89f5bd0, count=1, INVALID DATATYPE, MPI_MAXLOC, MPI_COMM_WORLD) failed
MPI_Allreduce(873): Invalid datatype
```
which are related to MPI2real_t type. I cannot see what in this branch would cause problems (consider the same type is in master). Since there are other gcc issues on anvil, i am not sure what to do. Homme has been failing on anvil in nightlies for a while and we may want to wait for centos7, if it is coming soon. @mt5555  ?

oksanaguba:
I just realized our development on anvil will be affected, too, so, i 'd better trace this. So, this  PR cannot be merged yet.

oksanaguba:
All is fine. At first I forgot to pull the latest fix into anvil integration repo. Now on anvil eam_theta suite passes (except for thetahy_sl test that is broken in master, too) and homme tests pass if barocammoistSL is disabled in both 'master' branch and next. For 'master' I used next with this commit:
```
commit a28bc157390e25d7245fde5778bd29d316e62567
Merge: aa74f65 d6de015
Author: Robert Jacob <jacob@anl.gov>
Date:   Thu Aug 29 22:35:27 2019 -0500

    Merge branch 'rljacob/compset/addcmip6' into next (PR #3118)
```
I will merge it without another round of testing next week.

oksanaguba:
merged to next without testing -- testing was done a few days ago.

dqwu:
SMS_B.ne4_ne4.F-EAMv1-AQP1.anlworkstation_gnu.cam-hommexx has build errors:
https://my.cdash.org/testDetails.php?test=49726168&build=1698805

SMS_B.ne4_ne4.F-EAMv1-AQP1.melvin_gnu.cam-hommexx has the same issue:
https://my.cdash.org/testDetails.php?test=49725881&build=1698793

dqwu:
Error messages in e3sm.log:
```
components/cam/src/dynamics/se/gravity_waves_sources.F90:119:35:

     use prim_driver_mod, only     : deriv1
                                   1
Error: Symbol ‘deriv1’ referenced at (1) not found in module ‘prim_driver_mod’
components/cam/src/dynamics/se/gravity_waves_sources.F90:149:61:

         gradth_gll(:,:,:,k,ie) = gradient_sphere(theta,deriv1,elem(ie)%Dinv)
                                                             1
Error: Symbol ‘deriv1’ at (1) has no IMPLICIT type
```

oksanaguba:
@dqwu thanks, I will try to fix it today.

oksanaguba:
I pushed fix that I tested on skybridge for cam_hommexx test, homme suite, and eam_theta suite. I also tested it on skybridge for cxx_vs_f tests in standalone homme, with profile short. Will re-merge to next now.

oksanaguba:
Re-merged to next.

mt5555:
adding a little background on deriv1 for future reference:

HOMME used to use  deriv(nthr), with one identical copy for each thread, and it was passed around to all subroutines.  a few months ago we replaced this with deriv1, dropped the thread dependence, and made it a global module variable.

dqwu:
SMS_B.ne4_ne4.F-EAMv1-AQP1.anlworkstation_gnu.cam-hommexx has been fixed:
https://my.cdash.org/viewTest.php?onlydelta&buildid=1698995

Issue #3095, opened 2019-07-30
Error with module files during case.setup in cori after down time  


amrapallig:
Many modules have been upgraded for cori after the maintenance and 
E3SM/cime/config/cesm/machines/config_machines.xml has the older versions of those modules that are no longer available for the machine.

For example, the error I got during ./case.setup is
ERROR: module command /opt/modules/default/bin/modulecmd python swap cray-mpich cray-mpich/7.7.0 failed with message:
cray-mpich(3):ERROR:105: Unable to locate a modulefile for 'cray-mpich/7.7.0'

I tried to fix few locally with available default versions of the moduefiles in the following file as mentioned. E3SM/cime/config/cesm/machines/config_machines.xml.
   cray-mpich/7.7.0   to  cray-mpich/7.7.6
   pmi/5.0.13   to   pmi/5.0.14
   craype/2.5.14  to  craype/2.5.18
   PrgEnv-intel/6.0.4  to  PrgEnv-intel/6.0.5
There are still more to fix.

I am getting module error one at a time, during case set up, and I am not sure how many more module files are needed to be fixed.

rljacob:
Work on cori started right after it came back up.  Watch for a PR with fixes.

ndkeen:
Indeed several of the modules we use on Cori were "removed" after the SW maintenance.  In the past, we've been able to use a module "pe_archive" to regain access to the older module versions, however, this one has been removed as well. I asked NERSC about this and I copy/pasted the reply below.

I've started down the path of updating several of our modules to current defaults for Cori.
Most of the tests in e3sm_developer pass, however, any that use MPAS will fail because of an issue with perl modules. The error will indicate a missing `Switch.pm`.  We've run into this before (when first porting to Cori) and I'm not sure exactly what happened to fix the issue for all of us. NERSC refused to install the module in the official perl install and I had resorted to installing my own Perl modules in `/project/projectdirs/acme/perl5`. I suggested other E3SM folks use this solution and it seemed to work, however, because this never came up again, I'm wondering if NERSC _did_ eventually add Switch. The perl modules I installed are still there, but something isn't working and I get a segfault when trying to use XML::LibXML. I'm trying to get around this, but it could be that we need to ask NERSC if they can allow Switch.  I did email them about this in a separate ticket and they suggested we be patient as many software packages need to be rebuilt.

[...]

PR #3097, opened 2019-07-31
Updated module versions for Cori (KNL and haswell) after Cori maint of July 2019

ndkeen:
After the Cori software upgrade in July 2019, there were several modules removed (or at least not easily used again) and this change attempts to use more recent versions of the modules. 

cray-mpich/7.7.0 --> cray-mpich/7.7.6
PrgEnv-intel/6.0.4 --> PrgEnv-intel/6.0.5
craype/2.5.14 --> craype/2.5.18
pmi/5.0.13 --> pmi/5.0.14
cray-netcdf-hdf5parallel/4.4.1.1.6 --> cray-netcdf-hdf5parallel/4.6.1.3
cray-hdf5-parallel/1.10.1.1 --> cray-hdf5-parallel/1.10.2.0 
cray-parallel-netcdf/1.8.1.3 --> cray-parallel-netcdf/1.8.1.4
cmake/3.11.4 --> cmake/3.14.4

PrgEnv-gnu/6.0.4 --> PrgEnv-gnu/6.0.5
gcc/7.3.0 --> gcc/8.2.0
cray-libsci/18.03.1 --> cray-libsci/19.02.1

For cori-knl, the default Intel compiler version is still intel/18.0.1.163, and the `--compiler=intel19` version is still intel/19.0.3.199.

However, for cori-haswell, as we only have one intel compiler, it will use intel/19.0.3.199 which is a change from before.  Note that intel/19.0.3.199 is the current NERSC default.

Also removed the edison entry in config_machines.xml as well as any edison references in config_compilers.xml simply to clean up. There are still several references to edison in other files that need cleaning.

Fixes https://github.com/E3SM-Project/E3SM/issues/3095

[BFB] because compiler version did not change.


ndkeen:
merged to next

ndkeen:
Looking at differences in results between master of July 22 before the Cori maint and now using this branch.

This test produces identical results:
`SMS_D_Ln5.ne4_ne4.FC5AV1C-L.cori-knl_intel`

ndkeen:
Now adding an environment variable to bring in the Perl Switch as part of the repo.
Hopefully can remove after NERSC makes more updates.

`setenv PERL5LIB "/project/projectdirs/acme/perl5/lib/perl5/x86_64-linux-thread-multi"`

tangq:
I just tried a case on cori-haswell with the maint-1.0 code. The module issue is still there. Perhaps, this PR has not been merged to maint-1.0 yet.

ndkeen:
That is correct. These changes are on next now. And a branch of maint-1.0 that I posted on slack.

rljacob:
Try using the branch ndk/maint-1.0/cori-july2019-upgradeb

tangq:
The branch works fine...building the new case now. Thank you both @ndkeen and @rljacob .

The queue time of cori-haswell seems getting worse - even my 1-hour test runs have to wait 3 days before it runs.

ndkeen:
Thanks for posting the branch @rljacob (my prior post was from phone and I didn't have easy access). Glad to hear that branch works for you @tangq. FYI, I'm seeing some potential issues with certain tests: hanging or running very slow. Not yet clear if it's because of these changes or if there's something else I can do yet.

rljacob:
We still don't have a test result from all the cori runs but that is because of other problems.  e3sm_prod_next_intel was fine and other's are reporting this branch works.  Its needed so merging.

PR #3101, opened 2019-07-31
Add number of nodes to srun on Anvil

amametjanov:
Add number of nodes to srun on Anvil

[BFB]

Addresses E3SM-Project/E3SM#3062, ESMCI/cime#3104


amametjanov:
Fixes [PET_Ln9_PS.ne30_oECv3_ICG.A_WCYCL1850S.anvil_intel](https://my.cdash.org/testDetails.php?test=48870699&build=1683644).

PR #3111, opened 2019-08-05
Fixes bug in crop Planting Date

### LINKED ISSUE #3110: Incorrect index used in planting date code
An incorrect index is used in the crop planting date calculation in the PhenologyMod. This results in incorrect values to be used for the precipitation seasonal calculations. This is also the likely cause of the crop test failing on some machines.

bbye:
This corrects the incorrect index value used in the plant date calculator
in the phenology module. Fixes #3110

[Non-BFB] - Non Bit-For-Bit (for global crop model only)

### DIFF
--- a/components/clm/src/biogeochem/CNPhenologyBeTRMod.F90
+++ b/components/clm/src/biogeochem/CNPhenologyBeTRMod.F90
@@ -1349,6 +1349,7 @@ contains
     integer fp,p      ! patch indices
     integer c         ! column indices
     integer g         ! gridcell indices
+    integer t         ! topographic indices
     integer h         ! hemisphere indices
     integer idpp      ! number of days past planting
     real(r8) dayspyr  ! days per year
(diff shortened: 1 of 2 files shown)


bbye:
I tested this on compy. Previously the test didn't run, but now it works. I don't expect any changes to the other tests. This will only effect the crop model, but I don't anticipate any baseline diffs with the crop test either (i.e., SMS_Ly2_P1x1.1x1_smallvilleIA.ICLM45CNCROP).

jqyin:
@bbye  SMS_Ly2_P1x1.1x1_smallvilleIA.ICLM45CNCROP.melvin_gnu.clm-force_netcdf_pio is BFB after PR merged, is that expected?

bbye:
Yes, that is expected. I don't expect this test to have any changes. The non-BFB tag is for the global crop model, which there are no tests for.

jqyin:
Merged #3111

PR #3113, opened 2019-08-06
Fix ne30pg2 domain files

whannah1:
Previous domain files were erroneously created using an ocean mask from the coarse oQU240 grid. This updates the ne30pg2 configuration to use new domain files generated using the gx1v6 grid file for the ocean mask. This clears up issues along coastlines and inland bodies of water, such as the Caspian and Black seas. domain files for the oECv3 grid were also generated and put on the inputdata server, so I've included them as well.

[non-BFB] for ne30pg2, BFB otherwise.

### DIFF
--- a/cime/config/e3sm/config_grids.xml
+++ b/cime/config/e3sm/config_grids.xml
@@ -911,7 +911,7 @@
       <grid name="rof">r05</grid>
       <grid name="glc">null</grid>
       <grid name="wav">null</grid>
-      <mask>gx1v6</mask>
+      <mask>oEC60to30v3</mask>
     </model_grid>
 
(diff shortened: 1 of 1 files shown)


ambrad:
I have a physgrid run queued on cori-knl using these. I'll report back when it's done. Ping me if a few days pass and I've forgotten.

rljacob:
@ambrad how did your testing go?

ambrad:
@rljacob, thanks for pinging me; I forgot to update this page. The test shows excellent results and is documented in the section "FC5AV1C-L ne30pg2 with good domain files" on the Confluence "Physgrid results" page.

rljacob:
merged to next.

PR #3128, opened 2019-08-14
Fix compiler error when fv_nphys=0

whannah1:
When building with GNU in debug mode on certain machines it will complain
if fv_nphys=0 even though those blocks of code are never used. This
fixes the issue by using a dummy variable to hold the product of
fv_nphys*fv_nphys.

[BFB]


rljacob:
The comment in the code you replaced says "Note: the max() is to fool the compiler debug mode when fv_nphys=0".  Did the max no longer work?

Rob

whannah1:
@AaronDonahue,  that is the correct behavior. We're only interested in the block/element, which always has nphys_sq FV cells.

@rljacob, the max would still have worked, but I figure the new way was more sensible and made the code a bit cleaner.

PR #3129, opened 2019-08-15
Remove the default-loaded module `craype-hugepages2M` which interferes with GNU on Cori

### LINKED ISSUE #3127: Problems with GNU-built E3SM executables on Cori after SW upgrade
I had found that any test using GNU would build OK, but fail at runtime immediately with errors that look like:

```
160: seg 0, mem 0: remap(0x43afe000,0x4277000,0x4277000,0x3,0x2aaab2ab0000) failed, error 14
192: seg 0, mem 0: remap(0x43afe000,0x4277000,0x4277000,0x3,0x2aaab2ab0000) failed, error 14
 85: seg 0, mem 0: remap(0x43afe000,0x4277000,0x4277000,0x3,0x2aaab2ab0000) failed, error 14
```

Assuming this was related to something on the system, I reported to NERSC and waited a while.  It was suggested I try to unload the hugepages modules. This is a module that is now default loaded for all Cori users. It can sometimes improve performance of apps, but we did not see any performance impact with these hugepages modules (there are a few different sizes). Simply removing this module and rebuilding, the GNU runs seem to be running. However, several are failing for others reasons.

I'm testing a branch now.

ndkeen:
After Cori software upgrade, one change was to load the `craype-hugepages2M` and this PR would simply add a command to explicitly remove that module for cori-knl and cori-haswell.

These are a few hugepages modules (at different sizes 2M,4M8M, etc) that can be loaded and require a rebuild to use. The use of hugepages for some apps can impact performance variability and reduce the cost of accessing memory in several MPI_Alltoall operations. However, initial tests when porting to Cori showed that there was no performance difference using these. I would just leave the module loaded (in favor of default machine settings), but for whatever reason, our GNU-built executables are failing unless the module is unloaded.

[bfb]
Fixes #3127

### DIFF
--- a/cime/config/e3sm/machines/config_machines.xml
+++ b/cime/config/e3sm/machines/config_machines.xml
@@ -122,7 +122,8 @@
         <command name="rm">cray-petsc</command>
         <command name="rm">esmf</command>
         <command name="rm">zlib</command>
-
+        <command name="rm">craype-hugepages2M</command>
+        
         <!-- first load basic defaults, then remove/swap/load as necessary -->
         <command name="load">craype</command>
         <command name="load">PrgEnv-intel</command>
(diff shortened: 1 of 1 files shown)


worleyph:
hugepages is also causing memory failures (where they did not occur before) in F cases with very large meshes, e.g. ne1024 .

ndkeen:
Are we sure we can point to the hugepages module as the reason for failure? I think Cray/NERSC would certainly not expect there to be any faults from this module -- doesn't mean it's not there.

worleyph:
This solved my build failure, and it is the only change I made.

ndkeen:
merged to next

PR #3131, opened 2019-08-15
Eliminate dependence of CICE message tags on global number of blocks

### LINKED ISSUE #3130: CICE gather_scatter routines can use message tags that are too large
The routines in ice_gather_scatter.F90 use message tags that
are offsets of the global block id. For high resolution runs
using SE meshes, e.g. ne1024, the number of blocks can be very
large, exceeding the maximum MPI message tag on some of our
targer systems, e.g Cori-KNL.

worleyph:
The routines in ice_gather_scatter.F90 use message tags that
are offsets of the global block id. For high resolution runs
using SE meshes, e.g. ne1024, the number of blocks can be very
large, exceeding the maximum MPI message tag on some of our
targer systems, e.g Cori-KNL. The use of global block ids is
not necessary, and here the algorithm for the one routine we
use in an F case (scatter_global_dbl) is restructured to
eliminate this dependency, depending instead on the number
of blocks on each process after the decomposition. Even this
may not be sufficient unless a large number of processes
is used, but it does provide a mechanism to decrease the
message tag size, and also saves some memory. And this change
is sufficient for the motivating case (ne1024 SE mesh and using
2048 processes for CICE on Cori-KNL).

Fixes #3130 

BFB

### DIFF
--- a/components/cice/src/mpi/ice_gather_scatter.F90
+++ b/components/cice/src/mpi/ice_gather_scatter.F90
@@ -193,6 +193,7 @@
 
    integer (int_kind) :: &
      i,j,n          ,&! dummy loop counters
+     nrecvs         ,&! current recv count for a particular process
      nsends         ,&! number of actual sends
      src_block      ,&! block locator for send
      src_task       ,&! source of message
@@ -211,6 +212,9 @@
    real (dbl_kind), dimension(:,:,:), allocatable :: &
      msg_buffer
 #else
+   integer (int_kind), dimension(:), allocatable :: &
+     nrecvs_p        ! number of recvs from each process
+
    real (dbl_kind), dimension(:,:), allocatable :: &
      msg_buffer
 #endif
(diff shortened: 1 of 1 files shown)


worleyph:
Note 1: I modified only the routine scatter_global_dbl, as this is the only routine that we use in an E3SM F case. I developed similar modifications for the other routines in ice_gather_scatter.F90, but backed them out of this PR since I can not easily test them, and they are irrelevant to E3SM at the moment.

worleyph:
Note 2: I labelled this as Atmosphere as well as Sea Ice, since this really only impacts F cases. While modifying CICE is generally a bad idea, as it is not really part of E3SM, I have heard of no recent plans to add the capability to use MPAS-SeaIce in F cases, and something needed to be done to allow this particular F case to run.

worleyph:
@rljacob , who should I ask to review and who should I ask to integrate this PR?

worleyph:
Sanity checks on Compy indicate that this is BFB.

philipwjones:
@brhillman This is a very old version of the CICE code that was extracted by folks to get a thermodynamic-only sea-ice model for use by the atmosphere in F cases. It was never supported by CICE or the ice community. It really needs to be replaced by the community (and E3SM)-supported icepack code, which is a column package that derived from the CICE column physics. Not sure whether this belongs to infrastructure (@rljacob ?) or someone else. Since icepack derived from CICE, I suspect the calls are similar, but would require some mods and then inserting an approved hash from the icepack repo (presumably already done for MPAS-seaice? @proteanplanet ?)

akturner:
We have developed a prescribed ice mode for MPAS-Seaice. When I find the time this will be integrated into E3SM, and validated by the atmosphere folks. It's my understanding that CICE is only currently needed in E3sm for the prescribed ice mode for F compsets

akturner:
Opps didn't mean to close this. Sorry.

worleyph:
@akturner , great! I won't look to make any more changes to CICE after this PR goes in. Do you know whether the prescribed ice mode will support any mesh or only MPAS meshes? Reason I ask is that this particular problem showed up when trying to use the spectral element ne1024 mesh, which is of some interest to the atmosphere modelers. (@brhillman or @mt5555 should elaborate here.)

PR #3135, opened 2019-08-15
Update climate reproducibility tests (MVK, PGN, TSC) 

jhkennedy:
Updates the MVK, PGN, and TSC climate reproducibility tests to all be production ready. They will all generate an appropriate baseline, run the comparison and produce a pass-fail result, and through [EVV](https://github.com/LIVVkit/evv4esm) provide a portable website with in-depth test details to analyze why these tests pass/fail. An example of the output websites produced for both a failing case and a bit-for-bit case can be seen here:

https://livvkit.github.io/evv4esm/index.html

This will allow the `e3sm_atm_nbfb` test suite, under `E3SM_Custom_Tests` on the [test dashboard](https://my.cdash.org/index.php?project=ACME_Climate), to pass and be operationally effective for climate reproducibility testing (tests' motivation discussed [here](https://github.com/E3SM-Project/E3SM/blob/4c05816d7ebfce2b2974f4ae71179d915c83fa83/cime/scripts/climate_reproducibility/README.md) ).

 These tests add external python dependencies to CIME (but only when one of these tests are run) and as such require a python environment with some of the scientific python stack. An `e3sm_simple` environment has been deployed to `cori` in the same manner as the `e3sm_unified` environment and will provide the necessary dependencies; see E3SM-Project/e3sm-unified#53. Additionally, the jenkins job which runs these test on `cori-knl` will be updated to use the `e3sm_simple` environment through  E3SM-Project/E3SM_test_scripts#6. 

**New baselines will need to be generated for these tests.**

[B4B]


jhkennedy:
In 20a0896 I've added the MVK test to the `e3sm_atm_nbfb` suite, however it's a little expensive to run -- currently on Cori it has 6 resubmits and uses 256 nodes for ~0.35 hours for each submit. @rljacob can we afford to add to the testing? Alternatively, we could deploy `e3sm_simple` to a machine that we aren't allocation limited and create an `e3sm_atm_nbfb_extended` test suite which includes MVK.

jhkennedy:
Also, @singhbalwinder , @huangmy , @salilmahajan , and @PeterCaldwell will likely want to follow along with this PR/discussion.

jhkennedy:
Thanks @jgfouca !

Issue #2680, opened 2019-01-09
Non-Standard Exner Definition in dp_coupling.F90?


PeterCaldwell:
@AaronDonahue noticed that the Exner function defined in the SE dycore's dp_coupling routine and passed into physics uses a non-standard (incorrect?) definition for the Exner function. The Exner function is the conversion factor (p/p_0)^{Rd/cp} needed to convert potential temperature into temperature. In this equation, p_0 is a reference pressure typically taken to be 1000mb (see e.g. http://glossary.ametsoc.org/wiki/Potential_temperature). In E3SM, p_0 is instead taken to be pint(i,pver+1), which is (at least approximately) the time-and-space-varying surface pressure. Here are the lines of code:

https://github.com/E3SM-Project/E3SM/blob/d0241119c45f6864d90fdee50dabccec3bdfa865/components/cam/src/dynamics/se/dp_coupling.F90#L511-L512

This weird definition caused conservation errors when Aaron implemented code which used the normal definition of theta in some places and this weird Exner in others... so at the very least, this current definition is dangerous.

Aaron used grep to identify locations using Exner and found it was used for the following:

> To calculate the bottom potential temperature for export to the surface models.
> In CLUBB*
> In vertical diffusion
> Unicon
> UW-Shallow Convection

and points out: 
> CLUBB is an interesting case.  CLUBB carries two versions of exner, there is the standard CAM value in the state structure and there is a local “exner_clubb” that is used for CLUBB specific calculations.  I talked with Peter B about it and this was because the definition of TH in CLUBB theory was inconsistent with CAM and they were experiencing massive issues over topography.  Defining exner_clubb fixed the problem.

I suspect the weird definition of Exner was intended for export to the surface models but is being improperly used in the other cases (except CLUBB, which noticed the problem and created exner-clubb as a workaround)? 

Does anyone know the history of this "feature"?

PeterCaldwell:
Yes, I was wondering about that as well while I wrote this issue, but thought cappa was probably defined as -Rd/cp. I just checked in control/physconst.F90 and you're right - cappa=Rd/cp so E3SM's definition is also the reciprocal of the normal definition.

AaronDonahue:
To clarify the line for exporting values to the surface is
```components/cam/src/control/camsrfexch.F90:462:      cam_out%thbot(i) = state%t(i,pver) * state%exner(i,pver)```

PeterCaldwell:
[...]

PR #3141, opened 2019-08-19
EAM physgrid: high-order, property preserving remap

ambrad:
Adds high-order, property-preserving remap between physics and GLL grids.

Provide a module, gllfvremap_mod, whose use in dynamics/se is optionally enabled
with atmosphere namelist option se_fv_phys_remap_alg=1 (default 0 uses the
original remap algorithm), to remap between physics and GLL grids. The remap

-- is mass conserving;
-- produces tracers that do not violate extrema established before the remap;
-- remaps tendencies only, so that the background state is not affected by remap;
-- has order of accuracy N, where N is the parameter in the physgrid specification pgN.

[BFB]


ambrad:
[v2 design document](https://acme-climate.atlassian.net/wiki/spaces/ED/pages/981172667/W16%2BPhysics%2BGrid)
[compact description and results](https://acme-climate.atlassian.net/wiki/spaces/COM/pages/1010171942/Physgrid+results)

ambrad:
This branch will not run with the updated, better domain files until PR #3113 is merged.

rljacob:
PR #3113 has been merged.

components/cam/src/dynamics/se/dyn_comp.F90
```diff
@@ -209,6 +211,7 @@ subroutine dyn_init1(fh, NLFileName, dyn_in, dyn_out)
     ! Initialize FV physics grid variables
     if (fv_nphys > 0) then
       call fv_physgrid_init()
+      if (se_fv_phys_remap_alg == 1) call gfr_init(par, elem, fv_nphys)
```
whannah1: would it make more sense to have this inside fv_physgrid_init()?

components/cam/src/dynamics/se/stepon.F90
```diff
@@ -598,6 +600,7 @@ subroutine stepon_final(dyn_in, dyn_out)
    ! Deallocate variables needed for the FV physics grid
    if (fv_nphys > 0) then
       call fv_physgrid_final()
+      if (se_fv_phys_remap_alg == 1) call gfr_finish()
```
whannah1: Similar to the comment above, would it make sense to move this into fv_physgrid_final()?

ambrad:
I'm running this branch on cori and seeing substantial climo differences relative to before I merged master; I'm looking into the cause.

components/cam/src/dynamics/se/stepon.F90
ambrad: Thanks; I'll do both.

components/cam/src/dynamics/se/dp_coupling.F90
```diff
@@ -113,9 +116,14 @@ subroutine d_p_coupling(phys_state, phys_tend,  pbuf2d, dyn_out)
...
         call t_startf('dyn_to_fv_phys')
-        call dyn_to_fv_phys(elem,ps_tmp(1:nphys_sq,:),zs_tmp(1:nphys_sq,:),     &
-                                 T_tmp(1:nphys_sq,:,:),uv_tmp(1:nphys_sq,:,:,:),&
-                                 om_tmp(1:nphys_sq,:,:),q_tmp(1:nphys_sq,:,:,:))
+        if (se_fv_phys_remap_alg == 0) then
+          call dyn_to_fv_phys(elem,ps_tmp(1:nphys_sq,:),zs_tmp(1:nphys_sq,:),     &
+               T_tmp(1:nphys_sq,:,:),uv_tmp(1:nphys_sq,:,:,:),&
+               om_tmp(1:nphys_sq,:,:),q_tmp(1:nphys_sq,:,:,:))
+        else
+          call gfr_dyn_to_fv_phys(par, dom_mt, tl_f, hvcoord, elem, ps_tmp, zs_tmp, &
```
whannah1: Should we try to think of a naming scheme that makes it clearer what the two options are? Maybe something like dyn_to_fv_phys_low-order() and dyn_to_fv_phys_high-order()?
Maybe it's moot since there might not be a good reason to keep both in the long run.

components/homme/src/share/interpolate_mod.F90
```diff
@@ -1685,21 +1685,6 @@ subroutine interpolate_vector3d(interpdata,elem,fld_cube,nlev,fld,input_coords,f
...
-       do k=1,nlev
-          ! convert fld from contra->latlon
-          v1 = fld(i,k,1)
-          v2 = fld(i,k,2)
-
-          fld(i,k,1)=D(1,1)*v1 + D(1,2)*v2
-          fld(i,k,2)=D(2,1)*v1 + D(2,2)*v2
-       end do
-    end do
-
```
whannah1: What's going on here? Does this affect anything outside the physgrid mapping routines?

components/homme/src/share/gllfvremap_mod.F90
```diff
@@ -0,0 +1,1707 @@
...
+    call gfr_init_geometry(elem, gfr)
+  end subroutine gfr_init
+
+  subroutine gfr_finish()
+    if (.not. allocated(gfr%fv_metdet)) return
+    deallocate(gfr%fv_metdet, gfr%D_f, gfr%Dinv_f, gfr%qmin, gfr%qmax, gfr%phis, &
+         gfr%spherep_f)
+  end subroutine gfr_finish
+
+  subroutine gfr_dyn_to_fv_phys_hybrid(hybrid, nt, hvcoord, elem, nets, nete, &
```
whannah1: Personally I'm not a fan of the "self-documenting code" idea, so I would appreciate some descriptions for each of these sub-routines. I also think comment lines of 60-100 dashes between subroutines make it a lot easier to scan big modules like this. So you can leave it this way if you like, but I think it's always good practice to think about the poor postdoc who's gonna have to try to trace through the code in 5-10 years.

components/homme/src/share/interpolate_mod.F90
ambrad: Good eye. This is inert code (i.e., it runs, but it has no effect) that I noticed while working on physgrid. I proposed to Mark and Oksana that I should remove this code chunk in this PR.

components/homme/src/share/gllfvremap_mod.F90
ambrad: I'll add more documentation and delimit groups of related functions.

components/cam/src/dynamics/se/dp_coupling.F90
ambrad: There are 5 public API functions that dynamics/se uses. Please feel free to suggest names for each:
```
               gfr_init, &
￼	       gfr_finish, &
￼	       gfr_dyn_to_fv_phys, &
￼	       gfr_fv_phys_to_dyn, &
￼	       gfr_fv_phys_to_dyn_topo,
```
If you would like me to rename the original functions, too, let me know what names you would like.

whannah1: Initially, I thought it makes sense to use some indication of "high order" in the names, whether it's "high_order" or "HO", but that also seems a bit too general if a higher order scheme gets implemented later.  
Would it be accurate to just call this "2nd order" and use "O2" in the front (or back) of the names instead of "gfr"?

ambrad: I'll be sure to mention in the added documentation (but see also [the latex doc](https://acme-climate.atlassian.net/wiki/download/attachments/1010171942/physgrid-ho-remap-doc.pdf?version=4&modificationDate=1565894780686&cacheVersion=1&api=v2)) that the OOA is N for pgN for u,v,T and max(2,N) for Q because of the limiter. So this current method has arbitrarily high order in N. But it's true: other methods might come in later. For example, I'm working on OOA > 1 for pg1 right now, and I might propose a boosted OOA for N < 3 later that would require a new halo exchange pattern to be implemented. On the other hand, I think we can expect any new method to be called through this interface, so I think it's safe to use the names you'd like without concern another interface will be introduced.

whannah1: oh ok, I misunderstood the OOA then. I guess the current name convention is fine then. Maybe it would be better to rename my scheme with something indicating "low order". In any case, let's just keep the current name convention.

components/homme/src/share/gllfvremap_mod.F90
ambrad: I pushed a commit that acts on all PR review comments up to this point.

components/cam/src/dynamics/se/dp_coupling.F90
oksanaguba: what is gfr_ for here?

components/cam/src/dynamics/se/dp_coupling.F90
```diff
@@ -466,11 +478,18 @@ subroutine p_d_coupling(phys_state, phys_tend,  dyn_in)
...
         ! Map FV physics state to dynamics grid
-        call fv_phys_to_dyn(elem,T_tmp(1:nphys_sq,:,:),   &
-                                 uv_tmp(1:nphys_sq,:,:,:),&
-                                 q_tmp(1:nphys_sq,:,:,:))
+        if (se_fv_phys_remap_alg == 0) then
+          call fv_phys_to_dyn(elem,T_tmp(1:nphys_sq,:,:),   &
+               uv_tmp(1:nphys_sq,:,:,:),&
+               q_tmp(1:nphys_sq,:,:,:))
+        else
+          dtime = get_step_size()
```
oksanaguba: How is dtime used in remap?

components/cam/src/dynamics/se/dyn_comp.F90
```diff
@@ -112,6 +112,7 @@ subroutine dyn_init1(fh, NLFileName, dyn_in, dyn_out)
     integer :: neltmp(3)
     integer :: npes_se
     integer :: npes_se_stride
+    integer :: ie
```
oksanaguba: Is this a debug leftover?

components/homme/src/share/gllfvremap_mod.F90
```diff
@@ -0,0 +1,2253 @@
...
+       gfr_fv_phys_to_dyn_topo, &
+       ! If nphys == 1, reconstruct the field to boost the OOA. If
+       ! nphys > 1, returns immediately.
+       gfr_pg1_reconstruct_topo, & ! call after the gfr_fv_phys_to_dyn_topo and the DSS
+       gfr_pg1_reconstruct         ! call after gfr_fv_phys_to_dyn and the DSS
+
+  integer, parameter :: nphys_max = np
+
+  real(kind=real_kind), parameter :: &
+       zero = 0.0_real_kind, half = 0.5_real_kind, &
```
oksanaguba: awesome

components/homme/src/share/gllfvremap_mod.F90
```diff
@@ -0,0 +1,2253 @@
...
+  ! For testing.
+  public :: &
+       gfr_test, &
+       gfr_g2f_scalar, gfr_f2g_scalar, gfr_f_get_latlon, gfr_f_get_cartesian3d, &
+       gfr_g_make_nonnegative, gfr_dyn_to_fv_phys_topo_elem, gfr_f2g_dss
+
+  ! Interfaces to support calling inside or outside a horizontally
+  ! threaded region.
+  interface gfr_dyn_to_fv_phys
+     module procedure gfr_dyn_to_fv_phys_hybrid
```
oksanaguba: is _hybrid for threaded? what is _dom_mt for? _mt probably stands multithreaded?

components/homme/src/share/gllfvremap_mod.F90
```diff
@@ -0,0 +1,2253 @@
...
+
+    if (nphys > np) then
+       ! The FV -> GLL map is defined only if nphys <= np. If we ever are
+       ! interested in the case of nphys > np, we will need to write a different
+       ! map. See "!assume" annotations for mathematical assumptions in
+       ! particular routines.
+       call abortmp('gllfvremap_mod: nphys must be <= np')
+    end if
+    if (qsize == 0) then
+       call abortmp('gllfvremap_mod: qsize must be >= 1')
```
oksanaguba: Why is qsize>0 needed?

components/homme/src/share/gllfvremap_mod.F90
```diff
@@ -0,0 +1,2253 @@
...
+          call gfr_dyn_to_fv_phys_topo_elem(elem, ie, phis(:,ie))
+       end if
+
+       call calc_dp(hvcoord, elem(ie)%state%ps_v(:,:,nt), dp)
+       call gfr_g2f_scalar(ie, elem(ie)%metdet, dp, dp_fv)
+
+       call get_temperature(elem(ie), wr2, hvcoord, nt)
+       call calc_p(hvcoord, elem(ie)%state%ps_v(:,:,nt), p)
+       call gfr_g2f_scalar(ie, elem(ie)%metdet, p, p_fv)
+       wr2 = wr2*(p/p0)**kappa
```
oksanaguba: Why is here Exner pressure? Is this code only for theta-l?

components/homme/src/share/gllfvremap_mod.F90
```diff
@@ -0,0 +1,2253 @@
...
+
+    if (hybrid%par%dynproc) then
+       call gfr_f2g_mixing_ratios_he(hybrid, nets, nete, gfr%qmin(:,:,nets:nete), &
+            gfr%qmax(:,:,nets:nete))
+    end if
+
+    if (nf > 1 .or. .not. gfr%boost_pg1) then
+       do ie = nets,nete
+          if (gfr%check) wr(:,:,1) = elem(ie)%state%phis
+          call limiter_clip_and_sum(np, elem(ie)%spheremp, gfr%qmin(1,1,ie), &
```
oksanaguba: you limit topo?

components/homme/src/share/gllfvremap_test_mod.F90
```diff
@@ -0,0 +1,552 @@
...
+             if (abs(b - a) > 5*eps*abs(a)) then
+                write(iulog, '(a,i3,es12.4,es12.4,es12.4,a)') 'gfrt> test3 q mass', &
+                     q, a, b, abs(b - a)/abs(a), ' ERROR'
+             end if
+          end if
+       end if
+    end do
+  end subroutine run
+
+  subroutine gfr_check_api(hybrid, nets, nete, hvcoord, elem)
```
oksanaguba: What does this routine do?

oksanaguba:
1. I assume this is bfb cause it does not touch original Walter's test in nightlies.

2. It is a suggestion to add description to each of routines in components/homme/src/share/gllfvremap_mod.F90 . 

3. As a follow up, would be good to have a PR with at least 1 ne4 run with pg2 (or 3?).

4. Minor question: if Qdp is remapped instead of Q, why u, v, T, etc. are not remapped as u*dp, ... values? Similarly, why remap omega_p instead of omega? Also, I did not really follow up what's going on with Exner pressure in remap code.

AaronDonahue:
I've been following the conversation a bit and in response to Okasana's #4 above, I'm not sure if this would be a good time to mention issue #2680 about how Exner pressure is calculated in dp_coupling and how it is used elsewhere.

components/cam/src/dynamics/se/dp_coupling.F90
ambrad: If Q is a tendency, dt is needed to reconstruct full state for use in the limiter.

components/cam/src/dynamics/se/dyn_comp.F90
ambrad: Yes; will fix.

components/homme/src/share/gllfvremap_mod.F90
ambrad: dom_mt is the name of the structure that contains threading information. _hybrid is the impl that is called when in a threaded region. _dom_mt is the impl that is called when not, and it then sets up a threaded region (using dom_mt) and calls _hybrid.

components/homme/src/share/gllfvremap_mod.F90
ambrad: Well, I suppose it isn't. But a use case for physgrid without any tracers seems contrived. Since I don't have a test for qsize=0, I prefer to abort on this b/c I think it's more useful to clearly say that an anomalous situation has arisen.

components/homme/src/share/gllfvremap_mod.F90
ambrad: No, it works for all dycores. The idea is to remap theta so that remap is adiabatic. theta_v would probably be even better, but I don't want to bring in qv.

components/homme/src/share/gllfvremap_mod.F90
ambrad: Yes. This is up for discussion, along with everything else we're thinking about re: topo. But for now I feel it's safest not to introduce new topo extrema. So the topo remap problem is: remap such that the average height is conserved and so that no new extrema are created.

components/homme/src/share/gllfvremap_test_mod.F90
ambrad: It runs a sequence of tests that are permitted to use gllfvremap_mod
only through its public API, i.e., the same routines called in the
full atmosphere. This is in contrast to the unit tests in
gllfvremap_mod.F90, which access internal functions. The API tests are
in three parts, as follows (copy-pasted from comments):

```
    !! Test 1.                                                                                                                                                                                                                                                   
    ! Test physgrid API.                                                                                                                                                                                                                                         
    !   Test that if tendency is 0, then the original field is                                                                                                                                                                                                   
    ! recovered with error eps ~ machine precision.                                                                                                                                                                                                              
    !   OOA for pgN should min(N, 2). The OOA is limited to 2 because                                                                                                                                                                                            
    ! of the way the tendencies are created. Instead of setting the FV                                                                                                                                                                                           
    ! value to the average over the subcell, the sampled value at the                                                                                                                                                                                            
    ! cell midpoint (ref midpoint mapped to sphere) is used. This                                                                                                                                                                                                
    ! creates a 2nd-order error. Tests 2 and 3 test whether the remaps                                                                                                                                                                                           
    ! achieve design OOA, in particular OOA 3 for pg3 u, v, T, phis                                                                                                                                                                                              
    ! fields.                                                                                                                                                                                                                                                    

    !! Test 2.                                                                                                                                                                                                                                                   
    ! Test topo routines. For pgN, phis should have OOA min(N,2)                                                                                                                                                                                                 
    ! because the field is limited.                                                                                                                                                                                                                              

    !! Test 3.                                                                                                                                                                                                                                                   
    ! Test FV fields that were not covered in the previous tests. This                                                                                                                                                                                           
    ! is done by copying them to look like tendencies.                                                                                                                                                                                                           
    !   For pg4, u,v,T should have l2 errors that are eps because pg4                                                                                                                                                                                            
    ! reconstructions the fields exactly, even with DSS.                                                                                                                                                                                                         
    !   For pg4, q should have l2 errors that converge at OOA >= 2. 2                                                                                                                                                                                            
    ! is the formal OOA b/c of the limiter. The limiter acts on the                                                                                                                                                                                              
    ! tendency, which in this case is exactly what is being examined.                                                                                                                                                                                            
    !   For pgN, N=1,2,3, u,v,T should have OOA N.                                                                                                                                                                                                               
    !   For pgN, q should have OOA min(N,2).                                                                                                                                                                                                                     
```

ambrad:
1. Yes.
2. OK, I'll add more.
3. Or I could do it in this PR with ne4pg2. Can I meet with you to learn how to add a newcase test?
4. u,v,omega(_p) is a topic of discussion. One argument is: remap to conserve average velocity; another is, remap to conserve momentum. Re: T, `theta*dp` is in fact remapped as part of making it approximately (up to moisture) adiabatic. Re: omega_p, I was again thinking of conserving its average value, but maybe omega is better; not sure. @mt555, any thoughts on converting omega_p to omega before remapping?

oksanaguba:
For a test, we can meet but it is rather simple:

See this file cime/config/e3sm/tests.py for a list of tests .
See this folder /home/onguba/acme-master/components/cam/cime_config/testdefs/testmods_dirs/cam/preqx_ftype0 for shell commands and namelist . Suffix thetahy_ftype0 will be picked up from running a test SMS.ne4_ne4.FC5AV1C-L.cam-thetahy_ftype0 and so, shell commands and nl for this test will be used. From what I see, Walter's test does not have a separate folder in testmods_dirs , but it is needed in case you want to run theta-l instead of preqx .

ambrad:
@AaronDonahue, I use p0 as defined in CAM (or in standalone Homme) in the calculation of Exner pressure. Does this answer the question?

components/cam/src/dynamics/se/dp_coupling.F90
ambrad: (G)LL (F)V (r)emap. I tend to give public functions in a module a common prefix like one does in C; see, e.g., [scalable space-filling curve, sfcmap](https://github.com/E3SM-Project/E3SM/blob/master/components/homme/src/share/spacecurve_mod.F90) and [scalable grid initialization, sgi](https://github.com/E3SM-Project/E3SM/blob/master/components/homme/src/share/scalable_grid_init_mod.F90)

AaronDonahue:
@ambrad, I'm guessing that there is an internal "Exner" calculation used in Homme and that as you stated it is defined correctly based on p0.  The issue I referred to is related to how "Exner" is calculated for use by the physics, which is (we believe) incorrect and different.  The issue points out that this can lead to confusion when someone attempts to use `state%exner` from physics, when in fact it isn't the classic Exner formula.  
In short, based on your response the issue is probably unrelated, but I didn't want to miss a chance to bring up what could be an inconsistency in the model and a potentially difficult to find bug.

ambrad:
@oksanaguba, I believe I've addressed all your comments:
1. omega is now remapped, with omega_p computations still remaining until an orthogonal PR that cleans up omega_p in general is made.
2. We discussed `u,v` vs `dp*u,dp*v` in person.
3. I added more comments. I think every routine now has a header doc.
4. I discussed and worked out approved p and dp calculations with the group.
5. With your help, I added test SMS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L.cam-thetahy_pg2.
Thanks for the initial review round.

whannah1:
@ambrad, can you summarize what you guys discussed about mass weighting the momentum fields? I'm curious about what your conclusions were.

ambrad:
@whannah1

The choices are to remap `dp*(u,v,omega)` or `(u,v,omega)`. An additional
constraint is that we want and have implemented remap of `(u,v)` as follows:
```
  GLL (u,v) -> GLL contravariant velocity -> FV contravariant velocity -> FV (u,v).
```
A nonmandatory constraint is that I think we should do the same thing for each
of `(u,v)` and `omega`. At this point, it's a question of conserving linear
momentum in the reference element or average linear velocity in the reference
element. Or perhaps one could introduce a nonlinear remap operator that
conserves energy in the element. There's no sense of a global linear momentum to
conserve. The DSS will break any of these. In any case, one is not obviously
better than the others. I decided it's better not to give the impression of
conserving some well-defined momentum, so I went with remapping velocity
alone. This choice also nicely coincides with how we remap phis (no `dp`): we
conserve average height in the remap, not some sort of mass-weighted height.

ambrad:
I ran e3sm_developer on sandiatoss3.

oksanaguba:
running merged next with developer suite on skybridge now.

oksanaguba:
Tests are ok ./cs.status.20190909_171401_5e8tyq  .

oksanaguba:
merged to next, will need bless for a new test tomorrow.

components/homme/src/share/gllfvremap_mod.F90
```diff
@@ -0,0 +1,2297 @@
...
+    type (domain1d_t), intent(in) :: dom_mt(:)
+    type (hybrid_t), intent(out) :: hybrid
+    integer, intent(out) :: nets, nete
+
+    integer :: ithr
+
+    ithr = omp_get_thread_num()
+    nets = dom_mt(ithr+1)%start
+    nete = dom_mt(ithr+1)%end
+    hybrid = hybrid_create(par, ithr, hthreads)
```
amametjanov: Should the access to `dom_mt` be with `ithr` and not `ithr+1` for zero-length arrays? A seg-fault shows up with `SMS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L.anvil_intel.cam-thetahy_pg2` (MPI-only run with 108 tasks for ATM) for tasks 96 and up: e.g.
```
106: forrtl: severe (174): SIGSEGV, segmentation fault occurred
106: Image              PC                Routine            Line        Source
106: e3sm.exe           0000000002981B51  Unknown               Unknown  Unknown
106: e3sm.exe           000000000297FC8B  Unknown               Unknown  Unknown
106: e3sm.exe           000000000290F104  Unknown               Unknown  Unknown
106: e3sm.exe           000000000290EF16  Unknown               Unknown  Unknown
106: e3sm.exe           000000000288B147  Unknown               Unknown  Unknown
106: e3sm.exe           0000000002896D40  Unknown               Unknown  Unknown
106: libpthread-2.12.s  00002B87996807E0  Unknown               Unknown  Unknown
106: e3sm.exe           000000000127A829  gllfvremap_mod_mp        1769  gllfvremap_mod.F90
106: e3sm.exe           00000000016897FA  inidat_mp_read_in         466  inidat.F90
106: e3sm.exe           00000000013CF77D  startup_initialco          54  startup_initialconds.F90
106: e3sm.exe           00000000011EB75F  inital_mp_cam_ini          42  inital.F90
106: e3sm.exe           0000000000500F40  cam_comp_mp_cam_i         159  cam_comp.F90
106: e3sm.exe           00000000004F7E39  atm_comp_mct_mp_a         312  atm_comp_mct.F90
106: e3sm.exe           000000000043B808  component_mod_mp_         257  component_mod.F90
106: e3sm.exe           000000000042A861  cime_comp_mod_mp_        1337  cime_comp_mod.F90
106: e3sm.exe           00000000004388B9  MAIN__                    122  cime_driver.F90
```

whannah1: It should definitely be (ithr+1). I actually ran into a very similar issue, but I'm having trouble remembering what it was. I'm gonna look back through some slack conversations with Andrew to try and recall the details.

ambrad: No. There are two separate issues here. I will need to fix the second.

The first is whether `ithr` or `ithr+1` should be used for indexing in general, given that `ithr` starts at 0. This depends on how `dom_mt` is declared. If it's declared as a `pointer`, then it inherits the index space from the allocation of the pointer. If it's declared as an assumed-shape array, then it does not inherit the index space. This issue comes up enough that I've created a single-file reproducer of the various cases:
```F90
module ptr
contains
  subroutine ptr1(a)
    real, pointer, intent(in) :: a(:) ! pointer attribute means a has index space of allocation
    print *,'ptr1',a(0)
  end subroutine ptr1

  subroutine ptr2(a)
    real, intent(in) :: a(:) ! non-pointer attribute means a has default index space
    print *,'ptr2',a(1)
  end subroutine ptr2

  subroutine ptr3(a)
    real, intent(in) :: a(4:6) ! non-pointer attribute also allows reindexing
    print *,'ptr3',a(4)
  end subroutine ptr3

#if 0
  subroutine ptr4(a)
    real, pointer, intent(in) :: a(4:6) ! not allowed
    print *,'ptr4',a(4)
  end subroutine ptr4
#endif
end module ptr

program main
  use ptr

  real, pointer :: a(:)

  allocate(a(0:3))
  a(0) = 42
  a(1) = 2
  call ptr1(a)
  call ptr2(a)
  call ptr3(a)
  deallocate(a)
end program main
```
Thus, `ithr+1` is correct in this case.

However, the second issue is a genuine bug. Here, for `ne=4`, there are 96 elements. Thus, if more than 96 procs are used, then this code is indexing a dom_mt that has 0 length. No index value is correct in this case. I need to fix this code so if an MPI rank has 0 elements, it doesn't do anything.

ambrad: I've opened https://github.com/E3SM-Project/E3SM/issues/3241 and will create a PR to fix this in the next few days.

amametjanov: Just curious, wouldn't this declaration work with `dom_mt(ithr)` access?
```
type (domain1d_t), pointer, intent(in) :: dom_mt(:)
```

ambrad: To be clear, the issue you found relates to having more processes than elements. Declaration of dom_mt is orthogonal to that problem.

Re: declaring dom_mt: As a matter of personal taste, I don't like to declare something as a pointer unless it needs to be, as doing so narrows the scope of application of the subroutine. In this case, taking an assumed-shape array as input provides the greatest generality.

amametjanov: Yes, that works, but another seg-fault occurs at `hybrid_create`: need the `par%dynproc` check.

ambrad: Right, exactly. Unless I'm mistaken, PR #3242 takes care of exactly the issue of including a dynproc check at all necessary spots. As I report in that PR, the 108-proc run for the 96-element case now completes. Thanks again for reporting the issue.

PR #3146, opened 2019-08-21
Add project for compy to config_machines.xml

wadeburgess:
Add <PROJECT>e3sm</PROJECT>


rljacob:
I think we decided not to define the PROJECT because it will be different for the 3 different groups that use compy.  Users have to specify it or set the  an environment variable called PROJECT.

PR #3147, opened 2019-08-21
Remove the module craype-hugepages2M for Cori.

ndkeen:
Remove the module craype-hugepages2M for Cori.
This module was added as a default module after the Cori SW upgrade.
It was not expected to have an impact, and perhaps does not.
However, it's best to keep environment as close as possible to what it was before upgrade.
This module *might* be causing some non-reproducable runtime errors.

[BFB]


rljacob:
@ndkeen go ahead and merge this.

ndkeen:
merged to maint-1.0

PR #3149, opened 2019-08-21
Adds AODALL to compute total aerosol (incl. nighttime) optical depth

singhbalwinder:
Aerosol optical depth (AOD) and column burden are currently calculated
in the model during daytime only, which makes the burden output
little useful. This module is modified to calculate burden and total
 optical depth during nighttime as well. The burden diagnostic
variables already exist in the model but a new variable AODALL is
added in order to keep the original one (AODVIS). The 2-D burden
variables for each size mode (BURDEN1, BURDEN2, …), four of them for
the default configuration, are moved out of the "history_verbose" if
loop, so that they can be written out when history_aero_optics is
set to "true".

[BFB] - Bit-For-Bit


singhbalwinder:
This PR is BFB except that it adds a new variable to the `cam.h0` history file.

singhbalwinder:
Thanks @hwangacme ! I will merge it to maint-1.0.

kaizhangpnl:
Sorry, I just noticed that it was assigned to me as well. I have a comment to post.

PR #3154, opened 2019-08-23
bug fix for SCM time index

bogensch:
This bug fix effects single column model (SCM) cases when the temporal resolution of the forcing is different than that of E3SM.  Here the index "i" in the iteration loop to find the next available forcing is redundant and can result in forcing being read in with a time step lag.  

Also cleaned up the print statements to the E3SM logs to make them more verbose and to print out the correct values.

All E3SM developer tests pass, except SMS_R_Ld5.ne4_ne4.FSCM5A97, which is an expected fail


wlin7:
Merged to next.

wlin7:
Merged to master. Requests submitted to bless diffs for the SCM tests on sandiatoss3 and melvin.

PR #3156, opened 2019-08-26
FATES api 8, nutrients

rgknox:
This changeset updates the interface with FATES, which accommodates nutrient fluxes through FATES and into the ELM land model.  Changes are backwards incompatible, FATES had a major api update.   This included changes to calls in the clmfates_interface, as well as code handling history writing. Notably, a change was added that allows history variables to exist, with a second dimension of size 1. Previously, the code would not allow this condition.

[non-BFB] for cases with FATES active
[BFB] for all other cases

### DIFF
--- a/components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml
+++ b/components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml
@@ -102,7 +102,7 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <!-- ================================================================== -->
 <!-- FATES default parameter file                                       -->
 <!-- ================================================================== -->
-<fates_paramfile >lnd/clm2/paramdata/fates_params_default_2trop_c190213.nc</fates_paramfile>
+<fates_paramfile >lnd/clm2/paramdata/fates_params_default_12pft_c190905.nc</fates_paramfile>
 
 
 <!-- soil order related parameters (relative to {csmdata}) -->


rgknox:
rebased changes off master

dqwu:
This PR (merged to next branch) has caused the following e3sm_developer test to fail on anlworkstation:
ERS_Ld3.f45_f45.ICLM45ED.anlworkstation_gnu.clm-fates

Error messages in e3sm.log:
```
...
clmfates_interfaceMod.F90:: reading fates_seed_suppl
ENDRUN:clmfates_interface::ReadParametereadNcdio:: - Error reading.  Var:fates_seed_suppl
ERROR: Unknown error submitted to shr_abort_abort.
```

rgknox:
@dqwu , thanks for the feedback.  I'm now realizing that with this update I will have to update the default FATES parameter file.

jqyin:
re-merged to next to fix fates test

wlin7:
@jqyin , do you think it ready to re-merge the PR?

rgknox:
@wlin7 , I will let @jqyin respond as well.  But I forgot to mention that the land_developer test suite does pass on cori now.

components/clm/src/main/histFileMod.F90
```diff
@@ -983,17 +989,19 @@ subroutine hist_update_hbuf(bounds)
...
-          num2d = tape(t)%hlist(f)%field%num2d
-          if ( num2d == 1) then
+          numdims = tape(t)%hlist(f)%field%numdims
+          if ( numdims == 1) then   
              call hist_update_hbuf_field_1d (t, f, bounds)
           else
+             num2d = tape(t)%hlist(f)%field%num2d
              call hist_update_hbuf_field_2d (t, f, bounds, num2d)
           end if
        end do
```
amametjanov: `numdims` is a local variable that can be overwritten by multiple threads (race hazard). I'll push an update:
```
-!$OMP PARALLEL DO PRIVATE (f, num2d)
+!$OMP PARALLEL DO PRIVATE (f, numdims, num2d)
        do f = 1,tape(t)%nflds
           numdims = tape(t)%hlist(f)%field%numdims
           if ( numdims == 1) then
```

PR #3158, opened 2019-08-27
Add configuration and bug fixes to enable running ne1024

brhillman:
Add configuration for new ne1024 and ne512 grids, as well as a few bug fixes to prevent integer overflows when using grids of this size. This allows using create_newcase to configure a case with resolution `ne1024np4_360x720cru_oRRS15to5` or `ne512np4_360x720cru_oRRS15to5`.

[bfb]

### DIFF
--- a/cime/config/e3sm/config_grids.xml
+++ b/cime/config/e3sm/config_grids.xml
@@ -2270,8 +2270,7 @@
     <domain name="oRRS15to5">
       <nx>5778136</nx>
       <ny>1</ny>
-      <!--file grid="ice|ocn">$DIN_LOC_ROOT/share/domains/domain.ocn.oRRS15to5.160207.nc</file-->
-      <file grid="ice|ocn">$DIN_LOC_ROOT/share/domains/domain.ocn.oRRS15to5.190514.nc</file>
+      <file grid="ice|ocn">$DIN_LOC_ROOT/share/domains/domain.ocn.oRRS15to5.160207.nc</file>
       <desc>oRRS15to5 is an MPAS ocean grid with a mesh density function that is roughly proportional to the Rossby radius of deformation, with 15 km gridcells at low and 5 km gridcells at high latitudes:</desc>
     </domain>
 
--- a/cime/config/e3sm/config_grids.xml
+++ b/cime/config/e3sm/config_grids.xml
@@ -1147,6 +1147,16 @@
       <mask>oRRS15to5</mask>
     </model_grid>
 
+    <model_grid alias="ne1024np4_360x720cru_oRRS15to5" compset="(DOCN|XOCN|SOCN|AQP1)">
+      <grid name="atm">ne1024np4</grid>
+      <grid name="lnd">360x720cru</grid>
+      <grid name="ocnice">oRRS15to5</grid>
+      <grid name="rof">null</grid>
+      <grid name="glc">null</grid>
+      <grid name="wav">null</grid>
+      <mask>oRRS15to5</mask>
+    </model_grid>
+
     <!--- mali grids -->
 
     <model_grid alias="ne30_oECv3_aisgis" compset="_MALI">
@@ -2232,7 +2242,8 @@
     <domain name="oRRS15to5">
       <nx>5778136</nx>
       <ny>1</ny>
-      <file grid="ice|ocn">$DIN_LOC_ROOT/share/domains/domain.ocn.oRRS15to5.160207.nc</file>
+      <!--file grid="ice|ocn">$DIN_LOC_ROOT/share/domains/domain.ocn.oRRS15to5.160207.nc</file-->
+      <file grid="ice|ocn">$DIN_LOC_ROOT/share/domains/domain.ocn.oRRS15to5.190514.nc</file>
       <desc>oRRS15to5 is an MPAS ocean grid with a mesh density function that is roughly proportional to the Rossby radius of deformation, with 15 km gridcells at low and 5 km gridcells at high latitudes:</desc>
     </domain>
 
@@ -2519,9 +2530,9 @@
     </gridmap>
 
     <gridmap atm_grid="ne1024np4" ocn_grid="oRRS15to5">
-      <map name="ATM2OCN_FMAPNAME">cpl/gridmaps/ne1024np4/map_ne1024np4_to_oRRS15to5_mono.190508.nc</map>
-      <map name="ATM2OCN_SMAPNAME">cpl/gridmaps/ne1024np4/map_ne1024np4_to_oRRS15to5_mono.190508.nc</map>
-      <map name="ATM2OCN_VMAPNAME">cpl/gridmaps/ne1024np4/map_ne1024np4_to_oRRS15to5_mono.190508.nc</map>
+      <map name="ATM2OCN_FMAPNAME">cpl/gridmaps/ne1024np4/map_ne1024np4_to_oRRS15to5_mono.190509.nc</map>
+      <map name="ATM2OCN_SMAPNAME">cpl/gridmaps/ne1024np4/map_ne1024np4_to_oRRS15to5_mono.190509.nc</map>
+      <map name="ATM2OCN_VMAPNAME">cpl/gridmaps/ne1024np4/map_ne1024np4_to_oRRS15to5_mono.190509.nc</map>
       <map name="OCN2ATM_FMAPNAME">cpl/gridmaps/oRRS15to5/map_oRRS15to5_to_ne1024np4_monotr.190509.nc</map>
       <map name="OCN2ATM_SMAPNAME">cpl/gridmaps/oRRS15to5/map_oRRS15to5_to_ne1024np4_monotr.190509.nc</map>
     </gridmap>
--- a/cime/config/e3sm/config_grids.xml
+++ b/cime/config/e3sm/config_grids.xml
@@ -1137,7 +1137,7 @@
       <mask>tx0.1v2</mask>
     </model_grid>
 
-    <model_grid alias="ne1024_360x720cru_ne1024" compset="(DOCN|XOCN|SOCN|AQP1)">
+    <model_grid alias="ne1024np4_360x720cru_ne1024np4" compset="(DOCN|XOCN|SOCN|AQP1)">
       <grid name="atm">ne1024np4</grid>
       <grid name="lnd">360x720cru</grid>
       <grid name="ocnice">ne1024np4</grid>
--- a/cime/config/e3sm/config_grids.xml
+++ b/cime/config/e3sm/config_grids.xml
@@ -1137,6 +1137,26 @@
       <mask>tx0.1v2</mask>
     </model_grid>
 
+    <model_grid alias="ne512np4_360x720cru_ne512np4">
+      <grid name="atm">ne512np4</grid>
+      <grid name="lnd">360x720cru</grid>
+      <grid name="ocnice">ne512np4</grid>
+      <grid name="rof">r05</grid>
+      <grid name="glc">null</grid>
+      <grid name="wav">null</grid>
+      <mask>oRRS15to5</mask>
+    </model_grid>
+
+    <model_grid alias="ne512np4_360x720cru_oRRS15to5">
+      <grid name="atm">ne512np4</grid>
+      <grid name="lnd">360x720cru</grid>
+      <grid name="ocnice">oRRS15to5</grid>
+      <grid name="rof">r05</grid>
+      <grid name="glc">null</grid>
+      <grid name="wav">null</grid>
+      <mask>oRRS15to5</mask>
+    </model_grid>
+
     <model_grid alias="ne1024np4_360x720cru_ne1024np4" compset="(DOCN|XOCN|SOCN|AQP1)">
       <grid name="atm">ne1024np4</grid>
       <grid name="lnd">360x720cru</grid>
@@ -1946,6 +1966,14 @@
       <desc>ne240np4 is Spectral Elem 1/8-deg grid:</desc>
     </domain>
 
+    <domain name="ne512np4">
+      <nx>14155778</nx>
+      <ny>1</ny>
+      <file grid="atm|lnd" mask="oRRS15to5">$DIN_LOC_ROOT/share/domains/domain.lnd.ne512np4_oRRS15to5.190417.nc</file>
+      <file grid="ice|ocn" mask="oRRS15to5">$DIN_LOC_ROOT/share/domains/domain.ocn.ne512np4_oRRS15to5.190417.nc</file>
+      <desc>ne512np4 is Spectral Elem 6km grid:</desc>
+    </domain>
+
     <domain name="ne1024np4">
       <nx>56623106</nx>
       <ny>1</ny>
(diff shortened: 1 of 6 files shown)


singhbalwinder:
Thanks @brhillman ! This looks good to me. I will merge as soon as @mt5555 approves it.

brhillman:
@mt5555 can you take a look at this and approve?

cime/config/e3sm/config_grids.xml
```diff
@@ -1137,6 +1137,46 @@
       <mask>tx0.1v2</mask>
     </model_grid>
 
+    <model_grid alias="ne512np4_360x720cru_ne512np4">
```
oksanaguba: Do you know what cru is for?

cime/config/e3sm/config_grids.xml
```diff
@@ -2213,7 +2270,8 @@
     <domain name="oRRS15to5">
       <nx>5778136</nx>
       <ny>1</ny>
-      <file grid="ice|ocn">$DIN_LOC_ROOT/share/domains/domain.ocn.oRRS15to5.160207.nc</file>
+      <!--file grid="ice|ocn">$DIN_LOC_ROOT/share/domains/domain.ocn.oRRS15to5.160207.nc</file-->
```
oksanaguba: What was the reason to replace this file? Related: did you upload new file(s) to input repo?

cime/config/e3sm/config_grids.xml
```diff
@@ -2499,6 +2557,40 @@
...
+      <map name="ATM2LND_SMAPNAME">cpl/gridmaps/ne512np4/map_ne512np4_to_360x720cru_mono.190409.nc</map>
+      <map name="LND2ATM_FMAPNAME">cpl/gridmaps/360x720cru/map_360x720cru_to_ne512np4_mono.190409.nc</map>
+      <map name="LND2ATM_SMAPNAME">cpl/gridmaps/360x720cru/map_360x720cru_to_ne512np4_mono.190409.nc</map>
+    </gridmap>
+
+    <gridmap atm_grid="ne1024np4" ocn_grid="oRRS15to5">
+      <map name="ATM2OCN_FMAPNAME">cpl/gridmaps/ne1024np4/map_ne1024np4_to_oRRS15to5_mono.190509.nc</map>
+      <map name="ATM2OCN_SMAPNAME">cpl/gridmaps/ne1024np4/map_ne1024np4_to_oRRS15to5_mono.190509.nc</map>
+      <map name="ATM2OCN_VMAPNAME">cpl/gridmaps/ne1024np4/map_ne1024np4_to_oRRS15to5_mono.190509.nc</map>
+      <!--map name="OCN2ATM_FMAPNAME">cpl/gridmaps/oRRS15to5/map_oRRS15to5_to_ne1024np4_monotr.190509.nc</map-->
```
oksanaguba: What's the difference between mono and monotr files? If you want to keep these comments with monotr files, would you want to add a comment on why?

oksanaguba:
Not sure if it is applicable here but for ne1024, maybe, not in this PR, but did you consider phys grid setup for pg2?

components/cam/bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -1436,6 +1442,8 @@
 <nu_p hgrid="ne60np4" >  1.0e14 </nu_p>
 <nu_p hgrid="ne120np4">  1.0e13 </nu_p>
 <nu_p hgrid="ne240np4">  1.1e12 </nu_p>
+<nu_p hgrid="ne512np4">  2.0e11 </nu_p>
+<nu_p hgrid="ne1024np4"> 2.5e10 </nu_p>
```
oksanaguba: So, that's 3.0 scaling.

components/cam/src/dynamics/se/restart_dynamics.F90
```diff
@@ -115,7 +115,7 @@ subroutine write_restart_dynamics(File, dyn_out)
     type(element_t), pointer :: elem(:)
     real(kind=r8) :: time
     integer :: ndcur, nscur
-    integer, pointer :: ldof(:)
+    integer(kind=pio_offset_kind), pointer :: ldof(:)
```
oksanaguba: How does this help? Is this for bigger numbers?

components/cam/src/dynamics/se/restart_dynamics.F90
```diff
@@ -359,13 +360,13 @@ function get_restart_decomp(elem, lev) result(ldof)
     do k=1,lev
        do ie=1,nelemd
           do i=1,np*np
-             ldof(j) = (elem(ie)%GlobalID-1)*np*np+(k-1)*nelem*np*np+i
+             ldof(j) = int(elem(ie)%GlobalID-1, pio_offset_kind)*np*np &
```
oksanaguba: Why is conversion done only on 1st multipliers, not, say, on np or nelem or i?

oksanaguba:
Is PR expected to be bfb for nightly tests? You replaced 1 domain file, is it used for anything besides new resolutions? If so, does this mean this PR will not be bfb for whatever case that file is used? I added a few questions, mainly, for my own benefit. For new files, it is just a suggestion to document which files you uploaded in the PR.

components/cam/src/dynamics/se/restart_dynamics.F90
brhillman: Yeah, we ran into integer overflows with ne1024. Specifying the larger integer type prevents that.

components/cam/src/dynamics/se/restart_dynamics.F90
brhillman: Converting any of them will result in the rest being promoted.

sarats: Some compilers like IBM XL are more picky. So, if you wish data type conversion on all, it's better to be explicit.

cime/config/e3sm/config_grids.xml
brhillman: Actually, that's a good question. I was having some trouble during domain checks with the new grids and the new way we generate mapping files (using TempestRemap instead of ESMF), and this was one of the changes I made when trying to debug that, thinking that the domains needed to be consistent with the maps. But I think this one shouldn't matter now that I look at it, I think the ocean domain is just copied straight through without any mapping. I have a test building now to see if I can get rid of this change. Will be good to narrow that down. Thanks for catching!

cime/config/e3sm/config_grids.xml
brhillman: Yeah, again this was related to problems with domain checks. "monotr" from ocean to atmos would be the transpose map computed based on the atmos to ocean mono map, while "mono" from ocean to atmos would be a separate independently calculated mono map. Not sure of the details, but again this was a change I implemented when trying to figure out the domain check errors I was running into. I will submit another test to confirm that this was a problem. The monotr map I think is the "preferred" algorithm if it works, but I recall having trouble with it.

components/cam/bld/namelist_files/namelist_defaults_cam.xml
brhillman: Is that a problem?

oksanaguba: No -- this was just for me to remember.

cime/config/e3sm/config_grids.xml
brhillman: Climate Research Unit, which is apparently the source for some of the land forcing data. The 360x720cru grid is described as the "Exact half-degree CRUNCEP datm forcing grid with CRUNCEP land-mask". According to the NCAR Research Data Archive (https://rda.ucar.edu/datasets/ds314.3/#!description), "The CRUNCEP is an atmospheric forcing dataset used forcing the land surface models. Specifically this dataset is designed to drive the Community Land Model over a long period. The CRUNCEP is a combination of two existing datasets; the CRU TS3.2 0.5 X 0.5 monthly data covering the period 1901 to 2002 and the NCEP reanalysis 2.5 X 2.5 degree 6-hourly data covering the period 1948 to 2016."

cime/config/e3sm/config_grids.xml
oksanaguba: Mark has a page explaining mapping suffixes https://acme-climate.atlassian.net/wiki/spaces/ED/pages/178848194/Transition+to+TempestRemap+for+Atmosphere+grids .

brhillman: Yeah, I'm familiar with that document. The reason for using mono vs monotr here was that I was getting domain check failures with monotr, but I can revisit this.

cime/config/e3sm/config_grids.xml
brhillman: I just reverted to the old file. My test confirmed that this change was not needed, it was just done for consistency with the other new files that were needed. Reverting to minimize changes in this PR.

cime/config/e3sm/config_grids.xml
brhillman: I just removed the commented out lines. Since the domain files use the mono maps, I think I'd have to replace both domain files and mapping files to test this. Since it's working now, I elected to just remove the commented out ones and move on.

brhillman:
@oksanaguba yes this should be BFB. I ran `e3sm_developer`, and for the tests that ran (master *before* this branch had some fails, I checked out the first commit on master before my branch to setup baselines) I got BFB answers. The only other failures I got were a few `MEMCOMP exception` and `TPUTCOMP error` notices.

oksanaguba:
Since description goes to commit message, i am moving info on new input repo files here:

New files added to repo to run configurations:
atm/cam/topo/USGS-gtopo30_ne1024np4_16xconsistentSGH_20190528.nc
atm/cam/topo/USGS-gtopo30_ne512np4_16xconsistentSGH_20190212.nc
atm/chem/trop_mam/atmsrf_ne512np4_interp_20190409.nc
atm/chem/trop_mam/atmsrf_ne1024np4_20190621.nc (transfer in progress)

Mapping files:
cpl/gridmaps/ne512np4/map_ne512np4_to_oRRS15to5_mono.190402.nc
cpl/gridmaps/oRRS15to5/map_oRRS15to5_to_ne512np4_mono.190402.nc
cpl/gridmaps/ne512np4/map_ne512np4_to_360x720cru_mono.190409.nc
cpl/gridmaps/360x720cru/map_360x720cru_to_ne512np4_mono.190409.nc
cpl/gridmaps/ne1024np4/map_ne1024np4_to_oRRS15to5_mono.190509.nc
cpl/gridmaps/oRRS15to5/map_oRRS15to5_to_ne1024np4_mono.190509.nc
cpl/gridmaps/ne1024np4/map_ne1024np4_to_360x720cru_mono.190508.nc
cpl/gridmaps/360x720cru/map_360x720cru_to_ne1024np4_mono.190508.nc

Domains (transfers in progress):
share/domains/domain.lnd.ne512np4_oRRS15to5.190417.nc
share/domains/domain.ocn.ne512np4_oRRS15to5.190417.nc
share/domains/domain.lnd.360x720cru_oRRS15to5.190417.nc
share/domains/domain.lnd.ne1024np4_oRRS15to5.190514.nc
share/domains/domain.ocn.ne1024np4_oRRS15to5.190514.nc

brhillman:
Thanks @oksanaguba !

oksanaguba:
I think it is ready. I am not sure how much of concern is what Sarat mentioned, about XL compiler, but if it is not covered by tests, any PR is a subject of such concern.

brhillman:
@oksanaguba it's possible the fix on that line was not necessary, I can do a test without it to see...Mark and I were just going through searching for possible overflow situations since we couldn't narrow down exactly what was triggering it easily.

oksanaguba:
It is up to you to see if you need those changes (if you don't, it is reasonable to revert), i am just saying if those lines are compiler-dependent, so are many other lines in other PRs. I am not aware of requirements to run PRs on summit (yet), if it is needed, we need to come up with a procedure to run with XL each PR with instructions.

components/cam/src/dynamics/se/restart_dynamics.F90
mt5555: we do this throughout out the code, and it's a long established part of the fortran standard.  IBM XL should be ok with it.

brhillman:
@oksanaguba I vote we leave it then, especially with Mark's comment above.

sarats:
Let's go ahead. We will revisit if there is an issue. 

IBM XL is adhering very strictly to the standard and triggered errors (e.g, https://github.com/E3SM-Project/E3SM/pull/3186) which wasn't an issue with other compilers. Hence my initial comment.

brhillman:
@singhbalwinder this should be ready to go.

rljacob:
Cool!  What compset do you run with these resolutions?

brhillman:
@rljacob I'm using FC5AV1C-H01A (although FC5AV1C-L should work as well). For these, since for the time being we are primarily interested in running hindcasts, I am specifying the atmos initial condition to one @wlin7 has prepared from some high resolution reanalysis data (and because spinning up an initial condition for a F-compset would be expensive at these resolutions). That is done by adding that information to user_nl_cam though, since he has not uploaded that to the repo as far as I know. Also, so far we've been able to run both the old and new dycores (preqx and theta-l), in both hydrostatic and non-hydrostatic modes, and with both 72 and 120 vertical level configurations.

PeterCaldwell:
It would be nice if the initial condition file was bundled rather than something the user needs to know. @wlin7 - is the initial condition for ne1024 part of inputdata yet? If not, could you add it? @brhillman - could you make another PR to make this the default init file for ne1024 runs when you get a chance? No hurry, but worth doing.

brhillman:
I could move it over to the repo, and add it here if @wlin is okay with that

brhillman:
I guess one argument against providing it is that since it's not spun up, it probably shouldn't be interpreted as a "F-compset" initial condition.

PeterCaldwell:
> I guess one argument against providing it is that since it's not spun up, it probably shouldn't be interpreted as a "F-compset" initial condition.

I'd hope that anyone running ne1024 would appreciate that it isn't a normal F climatological simulation, so I'm still in favor of making this change if Wuyin is ok with it...

singhbalwinder:
I think it is okay to merge this branch but I just wanted to double check given the above discussion about init cond file. I will merge this branch as soon as I get a go ahead.

brhillman:
Hold off, I am testing the configuration with the default initial condition file (moved the data to the inputdata repository on Blues this morning). I'll ping you when it's ready.

singhbalwinder:
@rljacob asked me to hold off on merging anything to next today. So I will merge it after you give me a go ahead and everything clears up on the testing side.

brhillman:
@singhbalwinder it's good to go now, I pushed the updates to the namelist to grab the initial condition files I uploaded.

PeterCaldwell:
@singhbalwinder and @brhillman  - what's the status on this now? I consider it to be very high priority to get this merged...

brhillman:
@PeterCaldwell it's ready on my end.

singhbalwinder:
Thanks @brhillman. I will merge it right away to next. 

@PeterCaldwell : Sorry about the delay. I was waiting for testing to clear and Rob asked me not merge anything on Friday as there is nobody to address if anything breaks during the weekend.

brhillman:
Sounds great, thanks @singhbalwinder !

singhbalwinder:
Pushed to next.

PR #3163, opened 2019-08-28
Update ELM routines to speed up restart write

jonbob:
This PR adds a couple of new routines to help speed up the restart write process. It brings in:
* a function GetGlobalIndexArray that adds capability to get global indices on an array
   instead of a single point;
* a new routine for only defining the subGrid restart, subgridRest_define_only, which is
  identical to subgridRest_write_only but does not gather data. 
Previously the restart write process called the same subroutine twice, once to define the variables and once to output them, but was unnecessarily gathering all the data for both steps. This PR also replaces calls to GetGlobalIndex on single points inside loops with calls to GetGlobalIndexArray.

The time for high-res (r0125) restart file output dropped from ~2000s to ~55s using these changes.

[BFB]

### DIFF
--- a/components/clm/src/main/subgridRestMod.F90
+++ b/components/clm/src/main/subgridRestMod.F90
@@ -19,6 +19,7 @@ module subgridRestMod
   use LandunitType       , only : lun_pp
   use ColumnType         , only : col_pp
   use VegetationType     , only : veg_pp
+  use perf_mod           , only : t_startf, t_stopf
   use restUtilMod
 
   !
@@ -61,13 +62,19 @@ contains
 
     if (flag /= 'read') then
        if (flag == 'define') then
+          call t_startf('subgridRest_define')
           call subgridRest_define_only(bounds, ncid, flag)
+          call t_stopf('subgridRest_define')
        else
+          call t_startf('subgridRest_write')
           call subgridRest_write_only(bounds, ncid, flag)
+          call t_stopf('subgridRest_write')
        end if
     end if
 
+    call t_startf('subgridRest_write-read')
     call subgridRest_write_and_read(bounds, ncid, flag)
+    call t_stopf('subgridRest_write-read')
 
   end subroutine subgridRest
 
--- a/components/clm/src/main/subgridRestMod.F90
+++ b/components/clm/src/main/subgridRestMod.F90
@@ -60,15 +60,9 @@ contains
     !------------------------------------------------------------------------
 
     if (flag /= 'read') then
-       if (flag == 'define') then
-          call t_startf('subgridRest_define')
-          call subgridRest_write_only(bounds, ncid, flag)
-          call t_stopf('subgridRest_define')
-       else if (flag == 'write') then
-          call t_startf('subgridRest_write')
-          call subgridRest_write_only(bounds, ncid, flag)
-          call t_stopf('subgridRest_write')
-       end if
+       call t_startf('subgridRest_write')
+       call subgridRest_write_only(bounds, ncid, flag)
+       call t_stopf('subgridRest_write')
     end if
 
     call t_startf('subgridRest_write-read')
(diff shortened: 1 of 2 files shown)


minxu74:
I have tested the branch on Summit. It still failed with the pnetcdf io type but do passed with the netcdf type. For the netcdf io type, the land restart file was still under writing after 2 hours walltime using the branch w/o the fixes, while the time was reduced to 173 second with the fixes.

bishtgautam:
@billsacks, These modifications could be useful for speeding up the writing of CLM restart files too.

components/clm/src/main/subgridRestMod.F90
```diff
@@ -58,13 +60,245 @@ subroutine subgridRest( bounds, ncid, flag )
...
+       if (flag == 'define') then
+          call subgridRest_define_only(bounds, ncid, flag)
+       else
+          call subgridRest_write_only(bounds, ncid, flag)
+       end if
     end if
 
     call subgridRest_write_and_read(bounds, ncid, flag)
 
   end subroutine subgridRest
```
amametjanov: Can you also add calls to start and stop timers around the three calls to `subgridRest_*`? E.g.:
```
use perf_mod, only : t_startf, t_stopf
...
call t_startf('subgridRest_define')
call subgridRest_define_only(bounds, ncid, flag)
call t_stopf('subgridRest_define')
```
It'd help isolate any imbalance among read, define and write in future. TIA.

jonbob: @amametjanov - thanks for the suggestion. I committed changes to do so, if you want to review them

bishtgautam:
@jonbob It just struck me instead of introducing a new `subgridRest_define_only()`, could you have modified the `subgridRest_write_only()` [though the name is not correct as it is called for both write and define] in the following way?

```
subroutine subgridRest_write_only(bounds, ncid, flag)

....
...

   ! Only set values 
    if (flag == 'write') then
       do g=bounds%begg,bounds%endg
          igarr(g)= mod(ldecomp%gdc2glo(g)-1,ldomain%ni) + 1
       enddo
    endif 
    call restartvar(ncid=ncid, flag=flag, varname='grid1d_ixy', xtype=ncd_int,    &
         dim1name='gridcell',                                          &
         long_name='2d longitude index of corresponding gridcell',     &
         interpinic_flag='skip', readvar=readvar, data=igarr)
```

jonbob:
@bishtgautam - I had thought about that as well and just wasn't sure which made more sense, adding a bunch of if/endif statements or having a separate subroutine. It's one of the reasons I very much wanted input from the land team, so thank you. I'm happy to go back to a single subroutine, if that is your preference. It would avoid issues with code duplication and potentially make it easier to add more fields to the output

bishtgautam:
>It would avoid issues with code duplication and potentially make it easier to add more fields to the output

While `if (flag == 'write')` isn't an elegant solution, you captured the motivation for my proposed solution in the above statement. 

Assuming memory isn't an issue, another solution could be:
- define multiple new variables (e.g. `igarr_mod` for `mod(ldecomp%gdc2glo(g)-1,ldomain%ni) + 1` and `igarr_div` for `(ldecomp%gdc2glo(g) - 1)/ldomain%ni + 1`) instead of existing variable (e.g. `igarr`), 
- have a single `if (flag == 'write')` at the very beginning of the subroutine that populates these variables, and 
- use these new variables instead of old variables.

jonbob:
@bishtgautam - let's try it with the if (flag == 'write') around every place that actually loads data and see what it looks like. While that is some extra logic, I think it's cleaner and simpler -- if anyone ever wants to add more output, it will be obvious what to do. I'll code it up and commit it -- if we don't like it, it's easy enough to revert

jonbob:
@bishtgautam - I committed what we had discussed. Please see what you think and let me know

bishtgautam:
@jonbob I thought the if-loop should be `if (flag == 'write')` instead of `if (flag == 'read')`. What do you think?

jonbob:
cut-and-paste error! Good catch, @bishtgautam

jonbob:
@bishtgautam - OK, try again? And I'll run a quick test just to make sure we still are getting the same results

bishtgautam:
@jonbob The changes look good to me. Thanks.

jonbob:
I tested the current status of this PR and verified the lnd restart files are still BFB. The timers show the overall restart write is 56-s, and show the difference between the three subgridRest calls are relatively small:
```
"l:subgridRest_define"                           -       1360     1360 1.360000e+03   2.860266e+00     0.002 
"l:subgridRest_write-read"                       -       1360     1360 2.720000e+03   8.909315e+03     6.559
"l:subgridRest_write"                            -       1360     1360 1.360000e+03   1.728696e+03     1.278
```
So it may not have been entirely necessary to add the logic  if (flag == 'write') around all the data gathering in the subgridRest routines, but it did save ~1.2-s.

I think this PR is good to go, and I'll remove the "DO NOT MERGE" label

bishtgautam:
> So it may not have been entirely necessary to add the logic if (flag == 'write') around all the data gathering in the subgridRest routines, but it did save ~1.2-s.

@jonbob, Since the time saving so small, should we just remove the `if (flag == 'write')` logic for keeping the code cleaner? I can make the necessary changes if you are tired of dealing with the ELM code :)

jonbob:
@bishtgautam - ah, the ELM code is yours to decide about and I'm happy either way. I can make the changes if you want, just let me know... And thanks for your help

bishtgautam:
@jonbob, In that case, please remove the `if (flag == 'write')`. Thanks.

jonbob:
@bishtgautam - OK, all that logic is removed and the code does appear cleaner. The results are still BFB and the timing difference is in the noise, 53s in my latest test. I hope this one is good to go?

bishtgautam:
@jonbob Thank you for the changes and this PR is ready to go.

jqyin:
Merged #3163

jonbob:
Thanks @jqyin

PR #3165, opened 2019-08-29
Fix DYN_OMEGA output

whannah1:
Fixes DYN_OMEGA output for phys grid runs. Previously only omega_p=omega*p was output, omega_p needs to be multiplied by pressure to output omega.

[BFB] because DYN_OMEGA is a new quantity introduced for phys grid runs.

### DIFF
--- a/components/cam/src/dynamics/se/stepon.F90
+++ b/components/cam/src/dynamics/se/stepon.F90
@@ -238,6 +238,8 @@ subroutine stepon_run2(phys_state, phys_tend, dyn_in, dyn_out )
    real(r8) :: dp(np,np,nlev),fq,fq0,qn0, ftmp(npsq,nlev,2)
    real(r8) :: tmp_dyn(np,np,nlev,nelemd)
    real(r8) :: fmtmp(np,np,nlev)
+   real(r8) :: p_tmp(np,np,nlev)      ! temporary pressure for DYN_OMEGA output
+   real(r8) :: om_tmp(np,np,nlev)     ! temporary omega for DYN_OMEGA output
    real(r8) :: dtime
    integer :: nlev_tot
    nlev_tot=(3+pcnst)*nlev
@@ -457,8 +459,15 @@ subroutine stepon_run2(phys_state, phys_tend, dyn_in, dyn_out )
       call outfld('DYN_Q'     ,dyn_in%elem(ie)%state%Q(:,:,:,1)       ,npsq,ie)
       call outfld('DYN_U'     ,dyn_in%elem(ie)%state%V(:,:,1,:,tl_f)  ,npsq,ie)
       call outfld('DYN_V'     ,dyn_in%elem(ie)%state%V(:,:,2,:,tl_f)  ,npsq,ie)
-      call outfld('DYN_OMEGA' ,dyn_in%elem(ie)%derived%omega_p(:,:,:) ,npsq,ie)
       call outfld('DYN_PS'    ,dyn_in%elem(ie)%state%ps_v(:,:,tl_f)   ,npsq,ie)
+      ! Multiply omega_p by pressure t get omega for output
+      p_tmp(:,:,1) = hvcoord%hyai(1)*hvcoord%ps0 + dyn_in%elem(ie)%state%dp3d(:,:,1,tl_f)/2
+      do k = 2,nlev
+         p_tmp(:,:,k) = p_tmp(:,:,k-1) + ( dyn_in%elem(ie)%state%dp3d(:,:,k-1,tl_f) &
+                                          +dyn_in%elem(ie)%state%dp3d(:,:,k  ,tl_f) )/2
+         om_tmp(:,:,k) = dyn_in%elem(ie)%derived%omega_p(:,:,k) * p_tmp(:,:,k)
+      enddo
+      call outfld('DYN_OMEGA',om_tmp(:,:,:),npsq,ie)
    end do
    
    
--- a/components/cam/src/dynamics/se/stepon.F90
+++ b/components/cam/src/dynamics/se/stepon.F90
@@ -238,8 +238,9 @@ subroutine stepon_run2(phys_state, phys_tend, dyn_in, dyn_out )
    real(r8) :: dp(np,np,nlev),fq,fq0,qn0, ftmp(npsq,nlev,2)
    real(r8) :: tmp_dyn(np,np,nlev,nelemd)
    real(r8) :: fmtmp(np,np,nlev)
-   real(r8) :: p_tmp(np,np,nlev)      ! temporary pressure for DYN_OMEGA output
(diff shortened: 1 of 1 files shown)


components/cam/src/dynamics/se/stepon.F90
```diff
@@ -457,8 +459,15 @@ subroutine stepon_run2(phys_state, phys_tend, dyn_in, dyn_out )
       call outfld('DYN_Q'     ,dyn_in%elem(ie)%state%Q(:,:,:,1)       ,npsq,ie)
       call outfld('DYN_U'     ,dyn_in%elem(ie)%state%V(:,:,1,:,tl_f)  ,npsq,ie)
       call outfld('DYN_V'     ,dyn_in%elem(ie)%state%V(:,:,2,:,tl_f)  ,npsq,ie)
-      call outfld('DYN_OMEGA' ,dyn_in%elem(ie)%derived%omega_p(:,:,:) ,npsq,ie)
       call outfld('DYN_PS'    ,dyn_in%elem(ie)%state%ps_v(:,:,tl_f)   ,npsq,ie)
+      ! Multiply omega_p by pressure t get omega for output
+      p_tmp(:,:,1) = hvcoord%hyai(1)*hvcoord%ps0 + dyn_in%elem(ie)%state%dp3d(:,:,1,tl_f)/2
+      do k = 2,nlev
```
oksanaguba: I am not sure this is correct. From how I read it, p at 1st mid level will depend on dp(1) and dp(2), which is not how it is usually computed. Here is the code from another place:
```
  ! compute pressure on interfaces                                                                                   
  p_i(:,:,1)=hvcoord%hyai(1)*hvcoord%ps0
  do k=1,nlev
     p_i(:,:,k+1)=p_i(:,:,k) + dp(:,:,k)
  enddo
  do k=1,nlev
     p(:,:,k) = (p_i(:,:,k+1)+p_i(:,:,k))/2
  enddo
```
Could you please use it? 

Also, I need to check whether theta-l model computes omega/p or omega, so, this correction is not needed for theta-l.

oksanaguba:
There is this code in dp_coupling:
```
!for theta there is no need to multiply omega_p by p
#ifndef MODEL_THETA_L
    !$omp parallel do private (lchnk, ncols, ilyr, icol)
    do lchnk = begchunk,endchunk
      ncols = get_ncols_p(lchnk)
      do ilyr = 1,pver
        do icol = 1,ncols
          if (.not.single_column) then
            phys_state(lchnk)%omega(icol,ilyr) = phys_state(lchnk)%omega(icol,ilyr) &
                                                *phys_state(lchnk)%pmid(icol,ilyr)
          end if
        end do ! icol
      end do ! ilyr
    end do ! lchnk
#endif
```
Is your code needed?

mt5555:
Yes, that could could also be removed if the original phys_state(lchunk)%omega() was set via the element_ops.F90 get_field interface. 

That would probably introduce a roundoff level change, which is why we put in the #ifdef instead?

oksanaguba:
I don't remember why it is ifdef instead of get_field . We probably can use get_field without nonbfb in preqx and in theta-l.

oksanaguba:
This is just a comment on variables' names: variables related to omega should be called omega_something... We sometimes end up grepping for some quantity through the whole code, and if someone is looking for omega, they will miss usage of 'om_' .

whannah1:
@oksanaguba, the code from the dp_coupling is irrelevant, since this only applies to the physics grid. The DYN_OMEGA output is for looking at the omega field on the dynamics grid when the FV physics grid is used. 

I'll change "om" to "omega"

I'll also change the pressure calculation to average the interface pressures. I grabbed the current method from somewhere else in the code, but now I forget where... I wish I could remember so we could change that too.

oksanaguba:
Thanks @whannah1 . So, does it mean now with phys_grid we can output quantities on phys grid as well as on dyn grid (those are probably only quantities that can be derived from homme state)? Unrelated, does it mean output in EAM now supports both grids, probably, in one file? What happens if someone wants to have output on interpolated grid? Asking just for future. 

Last question -- did you try to run usual, not phys grid, run, and compare DYN_OMEGA with OMEGA? Though the changes are really small. I will add BFB label cause DYN_OMEGA is new.

oksanaguba:
I set BFB label and changed description, but I recall you already have a test with phys_grid in nightlies. Does the test output DYN_OMEGA? Then this won't be BFB.

oksanaguba:
Test is in e3sm_developer, SMS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L .

whannah1:
@oksanaguba, we don't output DYN_OMEGA in the tests, so it should be BFB. 

To answer your other question, we can output both multiple grids per file, so phygrid OMEGA and DYN_OMEGA can coexist in 1 file. 

Not sure about output on an interpolated grid, I didn't realized that was an option. I haven't compared OMEGA to DYN_OMEGA in an np4 run, but I have a recent run that outputs both, so I can check. I suspect they might be a tiny bit different due to where the outfld calls are. I think in practice we should only ever compare DYN_OMEGA to itself between cases.

oksanaguba:
Merged to next without testing, should not affect tests as per conversation above.

dqwu:
SMS_D_Ln5.ne4_ne4.FC5AV1C-L failed on both anlworkstation and melvin:
https://my.cdash.org/testDetails.php?test=49726169&build=1698805
https://my.cdash.org/testDetails.php?test=49725882&build=1698793

e3sm.log shows:
```
At line 467 of file components/cam/src/dynamics/se/stepon.F90
Fortran runtime error: Index '73' of dimension 3 of array 'p_i' outside of expected range (1:72)
```

whannah1:
@dqwu looks like I used "nlev" for p_i. Stupid mistake. I'll push a fix.

oksanaguba:
This is my mistake that I did not explicitly asked about it but @whannah1 it is expected that developer runs at least 1 test with the code that was modified.

whannah1:
@oksanaguba, I did run several tests and I even looked at the output, but I never ran in debug mode, which seems to be how this was caught. I'll make sure to run a debug test next time.

oksanaguba:
Thanks @whannah1 , could I please ask you to run one of these failing tests and see if the problem is resolved? I usually run against baselines that I make out of corresponding base clone (in this case it would be next before this branch went it). If you do not run, I will before merging the fix.

whannah1:
@oksana, I'll do that now

oksanaguba:
@whannah1 were your tests ok? it would be better to sort this out today (if you have no time, I will run tests) . thanks!

whannah1:
baseline is done, waiting on the test now.

whannah1:
ok, test passed.

oksanaguba:
Thanks @whannah1 . Re-merged to next.

oksanaguba:
(just for future, please, add in description which tests you run and on what machine(s) )

PR #3167, opened 2019-08-30
Host climate reproducibility tests at machine specific locations

jhkennedy:
This adds the ability to specify an `CIME_HTML_ROOT` and `CIME_URL_ROOT` in `config_machines.xml` so that output websites produced by the climate reproducibility tests (MVK, PGN, TSC) can be copied into a location that is hosted, and the URL reported in `TestStatus.log` so it appears on CDASH

Both `CIME_HTML_ROOT` and `CIME_URL_ROOT` can be following this hierarchy:

1. Environment variable
2. File `$HOME/.cime/config`
3. `config_machines.xml`

`CIME_HTML_ROOT` and `CIME_URL_ROOT` are currently only specified for `cori-knl` and `cori-haswell`

[B4B]


jhkennedy:
@jgfouca You may want me to do this at ESMCI/cime since I'm now touching more than just the MVK, TSC, and PGN tests.

rljacob:
@jhkennedy open an issue at ESMCI about the need for machine specific web serving directories.   Explain your plan to add what you already added.

jgfouca:
@jhkennedy , it's fine as-is. You mostly just touched files they (NCAR) would never miss with.

PR #3023, opened 2019-06-27
Add compy and fix some bugs for maint-1.0

rljacob:
Add compy to the maint-1.0 machine files.

Allow compiling with PGI or Intel using special wrapper scripts for IntelMPI.
Add pelayouts from master.
Use the settings that were on master as of 88f572f69 except:
Change ne4 pelayouts to be 80 tasks.

Also fix a couple of general maint-1.0 bugs:
Fix a badly defined compset in EAM.
Fix a bug in cime preventing the use of conditionally defined MPI wrappers.

Fixes #3020 
Fixes #3027

PR #3168, opened 2019-08-30
Adds intel mpi specific options for pgi compiler on Compy

singhbalwinder:
Intel MPI for PGI compiler on Compy was missing specification of
right mpi wrapper for mpi compilers. This PR adds the missing
specifications in config_compilers.xml.

[BFB] - Bit-For-Bit


rljacob:
I thought I added PGI correctly but I guess I only did it for maint-1.0 in https://github.com/E3SM-Project/E3SM/pull/3023

singhbalwinder:
There was another error due to some missing libraries for the mpipgcc compiler. RC supoort fixed that and mct compiled fine after that.

PR #3171, opened 2019-09-03
Updates land initial condition for 20th Century CMIP6 compset

bishtgautam:
The new restart file is generated by remapping the following file:
20180316.DECKv1b_A1.ne30_oEC.edison.1980-01-01/20180316.DECKv1b_A1.ne30_oEC.edison.clm2.r.1980-01-01-00000.nc

[BFB]

### DIFF
--- a/components/clm/bld/namelist_files/use_cases/20thC_CMIP6_transient.xml
+++ b/components/clm/bld/namelist_files/use_cases/20thC_CMIP6_transient.xml
@@ -47,7 +47,7 @@
 
 <fsurdat res="ne30np4">lnd/clm2/surfdata_map/surfdata_ne30np4_simyr1850_2015_c171018.nc </fsurdat>
 <flanduse_timeseries res="ne30np4">lnd/clm2/surfdata_map/landuse.timeseries_ne30np4_hist_simyr1850_2015_c20171018.nc</flanduse_timeseries>
-<finidat res="ne30np4" mask="oEC60to30v3">lnd/clm2/initdata_map/I1850CLM45.ne30_oECv3.edison.intel.36b43c9.clm2.r.0001-01-06-00000_c20171023.nc</finidat>
+<finidat res="ne30np4" mask="oEC60to30v3">lnd/clm2/initdata_map/20180316.DECKv1b_A1.ne30_oEC.edison.clm2.r.1980-01-01-00000.8575c3f_c20190904.nc</finidat>
 
 <fsurdat res="r0125">lnd/clm2/surfdata_map/surfdata_0.125x0.125_simyr1850_c190730.nc</fsurdat>
 


bishtgautam:
The new land initial condition is available on Compy at: 
```
/compyfs/inputdata/lnd/clm2/initdata_map/20180316.DECKv1b_A1.ne30_oEC.edison.clm2.r.1980-01-01-00000.8575c3f_c20190903.nc
```
The new file is currently being transferred to E3SM public inputdata server.

@kaizhangpnl, Please give this branch or directly test out the `20180316.DECKv1b_A1.ne30_oEC.edison.clm2.r.1980-01-01-00000.8575c3f_c20190903.nc`.

rljacob:
@kaizhangpnl please approve this PR.

kaizhangpnl:
@rljacob thanks for the reminder. It seems this PR change will conflict with the compset - it's a compset for CMIP6 20th century simulations, but we now assign the 1980-01 state as initial condition. @golaz what do you think?

rljacob:
@golaz your input is needed on this PR.  See above.

rljacob:
@golaz still need your input on this PR.

golaz:
I'm not sure what is the best option for this. This file is for 1980; if we make it the new default, shouldn't we also change the default stat time? Or should we use IC files that were actually used for the DECK v1 simulations?

Currently, when we run AMIP simulations for development purposes, we manually override initial conditions to match what we want, see for example https://acme-climate.atlassian.net/wiki/spaces/EWCG/pages/1098940429/20200108.alpha20.amip.ne30+r05+oECv3.compy

rljacob:
@bishtgautam what do you think should be done here?

bishtgautam:
Defining land IC within each new compset definition has been problematic. e.g. if a fully coupled compset and F compset are using the same land configuration (i.e. SP-mode) for a given resolution, then both compsets should use the same land IC. Thus, the land IC should be defined within `components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml` and not in `components/clm/bld/use_cases/<NEW_COMPSET>.xml`. In V1, we did not pay a close attention on how the land IC files were being defined, but I believe we should consider sorting this out for V2 so we avoid as much of manual overriding. If V2 simulations are already underway, we should sort this out for V3.

PRs #3333 and #3344 are similarly fixing issues with land IC.

rljacob:
V2 tuning simulations are underway but no production sims.   I agree this should be fixed for V2.   @bishtgautam what about @golaz question on start time?

bishtgautam:
@golaz,
>This file is for 1980; if we make it the new default, shouldn't we also change the default stat time?

[run.20200108.alpha20.amip.ne30_r05_oECv3.compy.csh#L72](https://github.com/E3SM-Project/SimulationScripts/blob/master/archive/v2/alpha/amip/run.20200108.alpha20.amip.ne30_r05_oECv3.compy.csh#L72) is modifying the start time for F20TRC5-CMIP6. My inclination would be to fix the compset definition so that it automatically starts at 1980.

rljacob:
@golaz So yes we should make 1980 the default and it won't change anything because people do that in their runscripts.

golaz:
I'm fine with making 1980 the default start year consistently with this new file.

rljacob:
@bishtgautam should this still be WIP ?

bishtgautam:
Updated the title

jqyin:
@bishtgautam can you rebase over latest master to resolve the conflict? it's not very clear to me.

bishtgautam:
@jqyin, I have rebased my branch.

jqyin:
merged #3171

PR #3177, opened 2019-09-09
Change high-frequency output to daily, remove duplicate field from stream

jonbob:
This PR makes a small change to the mpas-ocean streams file, making the output interval daily for the high-frequency stream as requested for high-res simulations. It also removes a duplicate field from that stream definition.


jonbob:
merged to maint-1.0

PR #3181, opened 2019-09-10
Enable more variables to outpit at standard levels

wlin7:
Add more variables to be able to output at following pressure levels
1000, 975, 950, 925, 900, 850, 800, 700, 600, 500, 400, 300, 200, and 100
hPa, as well as lowest model level. With these, U, V, T, Q, RH, OMEGA,
Z3 can all output at these levels.
    
[BFB]


wlin7:
Merged to maint-1.0.

PR #3182, opened 2019-09-13
Update MKL linking on Anvil

amametjanov:
Also disable certain "bad" thread counts that span across sockets.

Addresses E3SM-Project/E3SM#2979

[BFB]


worleyph:
@amametjanov , unassigned myself as a reviewer as I will be unable to review the code in a timely fashion.

PR #3184, opened 2019-09-16
Fixes issue #3183

### LINKED ISSUE #3183: cam.buildlib failed on Summit with default IBM XL compiler
On Summit, after merging pending PR #3153 locally to fix build issue #3105, there is a new build error from cam (reproducible with F case):
`ERROR: BUILD FAIL: cam.buildlib failed`

The error message in atm.bldlog shows:
```
"/autofs/nccs-svm1_sw/summit/.swci/1-compute/opt/spack/20180914/linux-rhel7-ppc64le/xl-16.1.1-3/spectrum-mpi-10.3.0.1-20190611-aqjt3jo53mogrrhcrd2iufr435azcaha/include/mpif-config.h", line 63.18: 1514-004 (S) Name given for constant with PARAMETER attribute was defined elsewhere with conflicting attributes. Name is ignored.
** parallel_mod   === End of Compilation 1 ===
1501-511  Compilation failed for file parallel_mod.F90.
```
Line 63 of mpif-config.h is:
`parameter (MPI_MAX_PROCESSOR_NAME=256-1)`

With above information we can find out that the failure is caused by PR #3094.
More specifically, by the following code block added in commit https://github.com/E3SM-Project/E3SM/commit/1af9d61a6e3c070dcd62eafad98489674ff66dd1
```
  subroutine initmp_from_par(par)
...
    character(len=MPI_MAX_PROCESSOR_NAME)               :: my_name
    character(len=MPI_MAX_PROCESSOR_NAME), allocatable  :: the_names(:)
...
#ifdef _MPI
#include <mpif.h>
...
```
This code block has two build issues:
1) Causes cam.buildlib to fail when mpif.h is included (mpif-config.h is included indirectly), as described above. Note, this build error is only reproducible with default IBM XL compiler on Summit (not reproducible with GNU compiler on Summit, though)
2) When _MPI macro is not defined, definition error occurs as MPI_MAX_PROCESSOR_NAME parameter is undefined to the compiler.

A proposed fix is like below:
```
  subroutine initmp_from_par(par)
...
#ifdef _MPI
#include <mpif.h>
    character(len=MPI_MAX_PROCESSOR_NAME)               :: my_name
    character(len=MPI_MAX_PROCESSOR_NAME), allocatable  :: the_names(:)
...
```
PS, for module parallel_mod, mpif.h is already included at the beginning:
```
#ifdef _MPI
#include <mpif.h>
#endif
```
Is there a reason to include it again inside each of the 4 subroutines listed below?
subroutine init_par
subroutine initmp_from_par(par)
subroutine syncmp(par)
subroutine syncmp_comm(comm)

bartgol:
The constant MPI_MAX_PROCESSOR_NAME is only defined if mpi headers are
included, so it *must* be used only inside blocks guarded by `#ifdef _MPI`

Also, removing redundant inclusion of mpif.h.

Fixes #3183

### DIFF
--- a/components/homme/src/share/parallel_mod.F90
+++ b/components/homme/src/share/parallel_mod.F90
@@ -121,8 +121,6 @@ contains
     type(parallel_t), intent(out) ::  par
 
 #ifdef _MPI
-#include <mpif.h>
-
     integer(kind=int_kind)                              :: ierr
     logical :: running   ! state of MPI at beginning of initmp call
 #ifdef CAM
@@ -182,6 +180,7 @@ contains
 
   subroutine initmp_from_par(par)
     type (parallel_t),intent(in):: par
+#ifdef _MPI
     character(len=MPI_MAX_PROCESSOR_NAME)               :: my_name
     character(len=MPI_MAX_PROCESSOR_NAME), allocatable  :: the_names(:)
     integer(kind=int_kind),allocatable                  :: tarray(:)
@@ -189,8 +188,6 @@ contains
     integer(kind=int_kind)                              :: ierr,tmp_min,tmp_max
     integer :: node_color
            
-#ifdef _MPI
-#include <mpif.h>
     if (MPI_DOUBLE_PRECISION==20 .and. MPI_REAL8==18) then
        ! LAM MPI defined MPI_REAL8 differently from MPI_DOUBLE_PRECISION
        ! and LAM MPI's allreduce does not accept on MPI_REAL8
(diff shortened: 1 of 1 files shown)


dqwu:
PS, please add "Fixes #3183" to this PR's text (though it is in the title), such that the issue can be automatically closed after merge. Thanks.

bartgol:
Ah, I accidentally put "issue" in the commit text, sorry. I also removed the extra mpif include I missed. Should be good now.

bartgol:
@dqwu I forgot to add `-f` to my last push, so there was still an extra mpif.h include left. I just pushed the updated version.

PR #3153, opened 2019-08-23
Add explicit types to literals in COARE ocean surface flux scheme

JackReevesEyre:
The COARE ocean surface flux scheme had un-typed literals (including some in functions that caused a build issue on Summit). This change adds explicit types - in most cases converting to "_R8" reals. This shouldn't affect the climate of "out of the box" cases (which use a different ocean surface flux scheme). But, when using the COARE scheme (ocn_surface_flux_scheme = 1 in user_nl_cpl), it is non-BFB with the earlier COARE implementation. 

Fixes #3105

[BFB] [FCC]

PR #3186, opened 2019-09-17
Fix some build errors for IBM compiler

dqwu:
With PR #3153 and #3184 locally merged for testing, there are still build errors reported by IBM XL compiler on Summit.

This PR fixes the following build errors for IBM compiler:
1511-077 (E) ELSEWHERE statement is in error.
1513-041 (S) Arguments of the wrong type were specified for the INTRINSIC procedure "max".
1515-019 (S) Syntax is incorrect.
1515-021 (E) Syntax error: Token " & " is expected.
1515-022 (S) Syntax Error: Extra token " , " was found.

[BFB]

### DIFF
--- a/components/cam/src/control/scamMod.F90
+++ b/components/cam/src/control/scamMod.F90
@@ -580,8 +580,8 @@ subroutine setiopupdate
           write(iulog,*) 'iopTimeIdx (IOP index) =', iopTimeIdx
           write(iulog,*) 'nstep = ',get_nstep()
           write(iulog,*) 'ncdate (E3SM date) =',ncdate,' ncsec=',ncsec
-          write(iulog,*) 'next_date (IOP file date) =',next_date_print,' &
-	                  next_sec=',next_sec_print
+          write(iulog,*) 'next_date (IOP file date) =',next_date_print,&
+                         'next_sec=',next_sec_print
           write(iulog,*)'******* do iop update'
       endif
 
--- a/components/homme/src/share/gllfvremap_mod.F90
+++ b/components/homme/src/share/gllfvremap_mod.F90
@@ -184,10 +184,10 @@ contains
           ! Document state of pg1. dcmip2016_test1 shows it is too coarse. For
           ! boost_pg1 = true, stepon's DSS loop needs to be separated from its
           ! tendency application loop.
-          write(iulog,*), 'gfr> Warning: pg1 is too coarse; see comments at top of gllfvremap_mod.F90'
+          write(iulog,*) 'gfr> Warning: pg1 is too coarse; see comments at top of gllfvremap_mod.F90'
           if (.not. gfr%boost_pg1) then
-             write(iulog,*), 'gfr> Warning: If you want to try pg1, use the boosted-accuracy &
-                  boost_pg1 option and call gfr_pg1_reconstruct(_topo).'
+             write(iulog,*) 'gfr> Warning: If you want to try pg1, use the boosted-accuracy &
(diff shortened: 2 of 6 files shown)


components/cam/src/control/scamMod.F90
```diff
@@ -581,7 +581,7 @@ subroutine setiopupdate
           write(iulog,*) 'nstep = ',get_nstep()
           write(iulog,*) 'ncdate (E3SM date) =',ncdate,' ncsec=',ncsec
           write(iulog,*) 'next_date (IOP file date) =',next_date_print,' &
-	                  next_sec=',next_sec_print
+	                  &next_sec=',next_sec_print
```
jayeshkrishna: Just curious, does the following work (entire string in the next line)?
```
write(iulog,*) 'next_date (IOP file date) =',next_date_print,&
	                  'next_sec=',next_sec_print
```

dqwu: Yes, that also works.

sarats: This looks cleaner.

components/homme/src/share/gllfvremap_mod.F90
```diff
@@ -1956,7 +1956,7 @@ subroutine check_g2f_mixing_ratio(gfr, hybrid, ie, qi, elem, dp, dp_fv, q_g, q_f
        qmax_f = maxval(q_f(:nf,:nf,k))
        qmin_g = minval(elem(ie)%state%Q(:,:,k,qi))
        qmax_g = maxval(elem(ie)%state%Q(:,:,k,qi))
-       den = gfr%tolfac*max(1e-10, maxval(abs(elem(ie)%state%Q(:,:,k,qi))))
+       den = gfr%tolfac*max(1e-10_real_kind, maxval(abs(elem(ie)%state%Q(:,:,k,qi))))
```
sarats: Just checking if this should be _r8 or _real_kind?

components/homme/src/share/gllfvremap_mod.F90
```diff
@@ -184,10 +184,10 @@ subroutine gfr_init(par, elem, nphys, check, boost_pg1)
           ! Document state of pg1. dcmip2016_test1 shows it is too coarse. For
           ! boost_pg1 = true, stepon's DSS loop needs to be separated from its
           ! tendency application loop.
-          write(iulog,*), 'gfr> Warning: pg1 is too coarse; see comments at top of gllfvremap_mod.F90'
+          write(iulog,*) 'gfr> Warning: pg1 is too coarse; see comments at top of gllfvremap_mod.F90'
           if (.not. gfr%boost_pg1) then
-             write(iulog,*), 'gfr> Warning: If you want to try pg1, use the boosted-accuracy &
-                  boost_pg1 option and call gfr_pg1_reconstruct(_topo).'
+             write(iulog,*) 'gfr> Warning: If you want to try pg1, use the boosted-accuracy &
+                  &boost_pg1 option and call gfr_pg1_reconstruct(_topo).'
```
sarats: Would omitting & on second line work?

dqwu: It works with GNU compiler, but not XL.
According to the link below, "Lines can be broken within a string constant, in which case the initiating ampersand is mandatory"
http://fortranwiki.org/fortran/show/Continuation+lines

components/cam/src/control/scamMod.F90
dqwu: OK, I will update it with Jayesh's suggestion.

Issue #3169, opened 2019-08-31
GCC 8 fails to build 2 e3sm_developer tests on anlworkstation


dqwu:
This build issue is caused by PR #2989 on seaice update.

SMS.T62_oQU120_ais20.MPAS_LISIO_TEST has build errors like below:
```
mpas_seaice_advection_incremental_remap_tracers.f90:1035:34:

           thisTracer % array2D => tracerField2DReal % array
                                  1
Error: Assignment to contiguous pointer from non-contiguous target at (1)
```
https://github.com/MPAS-Dev/MPAS-Model/issues/342 has already reported similar build errors (Intel v19 in debug mode on PNNL Compy cluster), which will be fixed by https://github.com/MPAS-Dev/MPAS-Model/pull/343

SMS_P12x2.ne4_oQU240.A_WCYCL1850.anlworkstation_gnu.allactive-mach_mods has build errors like below:
```
mpas_seaice_velocity_solver_constitutive_relation.f90:174:51:

 !$omp declare simd(seaice_evp_constitutive_relation)
                                                   1
Error: Symbol ‘seaice_evp_constitutive_relation’ at (1) has already been host associated
```
@amametjanov 
You can run following commands to reproduce the build errors on anlworkstation.
```
git clone --branch dqwu/anlworkstation_gcc8 https://github.com/E3SM-Project/E3SM.git
cd E3SM
git submodule update --init
cd cime/scripts
./create_test SMS.T62_oQU120_ais20.MPAS_LISIO_TEST --no-run --machine=anlworkstation
./create_test SMS_P12x2.ne4_oQU240.A_WCYCL1850.anlworkstation_gnu.allactive-mach_mods --no-run --machine=anlworkstation
```

amametjanov:
Yes, gnu apparently does not support `omp declare simd` and the build issue can be avoided by MPAS-Dev/MPAS-Model#348.

dqwu:
@amametjanov 
With latest next branch (mpas-source submodule has been updated), these 2 tests passed.
Therefore, this issue is ready to be closed, thanks.

amametjanov:
Fixed by E3SM-Project/E3SM#3230

PR #3192, opened 2019-09-23
Enable newer PIO2 interfaces for MPAS

jayeshkrishna:
Turning on newer PIO2 interfaces in MPAS for both
scorpio and scorpio classic

[BFB]

### DIFF
--- a/cime/config/e3sm/machines/config_compilers.xml
+++ b/cime/config/e3sm/machines/config_compilers.xml
@@ -78,6 +78,12 @@ for mct, etc.
      compilers -->
 <compiler>
   <SUPPORTS_CXX>FALSE</SUPPORTS_CXX>
+  <CPPDEFS>
+    <!-- MPAS Components rely on this flag to use the newer PIO/Scorpio interfaces. These newer interfaces are available on both Scorpio and Scorpio classic -->
+    <append MODEL="ice"> -DUSE_PIO2 </append>
+    <append MODEL="ocn"> -DUSE_PIO2 </append>
+    <append MODEL="glc"> -DUSE_PIO2 </append>
+  </CPPDEFS>
 </compiler>
 
 <compiler COMPILER="cray">


jayeshkrishna:
Also see MPAS-Dev/MPAS-Model#373 on the effort to cleanup PIO usage in MPAS (The short version : MPAS developers require backwards compatibility with older versions of PIO and hence cannot accept a cleanup that removes older PIO1 interfaces)

jayeshkrishna:
This change looks good on anlworkstation with the GNU compiler, currently testing it out with the XL compiler.

dqwu:
This PR has been tested with both scorpio_classic and scorpio on anlworkstation.
The test result looks good.

scorpio_classic (pio1) e3sm_developer tests:
https://my.cdash.org/buildSummary.php?buildid=1706021

scorpio (pio2) e3sm_developer tests:
https://my.cdash.org/buildSummary.php?buildid=1706018

The 2 failed tests have known build errors (see issue #3169).

cime/config/e3sm/machines/config_compilers.xml
```diff
@@ -78,6 +78,12 @@ for mct, etc.
      compilers -->
 <compiler>
   <SUPPORTS_CXX>FALSE</SUPPORTS_CXX>
+  <CPPDEFS>
+    <!-- MPAS Components rely on this flag to use the newer PIO/Scorpio interfaces. These newer interfaces are available on both Scorpio and Scorpio classic -->
+    <append MODEL="ice"> -DUSE_PIO2 </append>
+    <append MODEL="ocn"> -DUSE_PIO2 </append>
+    <append MODEL="glc"> -DUSE_PIO2 </append>
+  </CPPDEFS>
```
jonbob: In this file MODEL is used to refer to components by name, like "cism" or "cam". But this must work to pass all the tests? I'm more curious than anything....

jayeshkrishna: Yes, this change sets the flags for only ice, ocn and glc components.

jonbob: I was just curious because if I grep "MODEL" in that file, all the results are model names like "cam" or "clm" instead of component types like atm or lnd:
```[wolf966@compy-e machines]$ grep MODEL config_compilers.xml 
    <append MODEL="moby"> -DDIR=NOOP </append>
    <append MODEL="csm_share"> -std=c99 </append>
    <append MODEL="cism"> -D CISM_GNU=ON </append>
    <append MODEL="cism"> -D CISM_GNU=ON </append>
    <append MODEL="datm"> -Mnovect </append>
    <append MODEL="dlnd"> -Mnovect </append>
    <append MODEL="drof"> -Mnovect </append>
    <append MODEL="dwav"> -Mnovect </append>
    <append MODEL="dice"> -Mnovect </append>
    <append MODEL="docn"> -Mnovect </append>
    <append MODEL="cam"> </append>
    <append MODEL="datm"> -Mnovect </append>
    <append MODEL="dlnd"> -Mnovect </append>
    <append MODEL="drof"> -Mnovect </append>
    <append MODEL="dwav"> -Mnovect </append>
    <append MODEL="dice"> -Mnovect </append>
    <append MODEL="docn"> -Mnovect </append>
```

rljacob:
Can you change the cpp to "USE_PIO2API" ?

jayeshkrishna:
The cpp cannot be changed to USE_PIO2API without changing it in MPAS (MPAS logic distinguishes between PIO1 and PIO2, there are some new constants used with USE_PIO2 that was not available in PIO1).

cime/config/e3sm/machines/config_compilers.xml
jayeshkrishna: @jgfouca can you confirm that the above usage of MODEL attribute is supported by CIME (MODEL=ice/ocn/glc)?

(It works as expected for me but @jonbob wants to confirm that the usage is correct)

jayeshkrishna:
@rljacob : I will start a conversation with MPAS developers on whether they are ok with such a change (USE_PIO2API : (1) MPAS could still distinguish between PIO1 and PIO2, but instead of using USE_PIO2 when PIO2 is detected the code could use USE_PIO2API (2) Allow users to define USE_PIO2 or USE_PIO2API to use PIO2 interfaces)

cime/config/e3sm/machines/config_compilers.xml
jgfouca: @jayeshkrishna , @jonbob , looking at build.py:

```
    with open(file_build, "w") as fd:
        stat = run_cmd("MODEL={} SMP={} {} {} {} {} "
                       .format(compclass, stringify_bool(smp), cmd, caseroot, libroot, bldroot),
                       from_dir=bldroot,  arg_stdout=fd,
                       arg_stderr=subprocess.STDOUT)[0]
```

It looks like MODEL is set to the compclass, so your usage is correct. I'm concerned that all the other uses may be wrong.

jayeshkrishna: Thanks @jgfouca

jgfouca: I'm doing some tests now to confirm

jonbob: Thanks for following up on this, @jayeshkrishna and @jgfouca . @rljacob , do we need to fix this file so the MODEL usage is consistent?

jgfouca: Looking through Tools/Makefile, it appears as though MODEL actually is supposed to be a component name, not component class; this may be a case of a mislabelled variable in build.py. I vaguely remember sorting this stuff out in the CMake work.

jgfouca: Unfortunately, it appears there is no guarantee on the contents of MODEL. The default is the comp class, but many of the buildlib scripts for individual components appear to override it with their name. This is one of the things I cleaned-up in the Cmake work.

jonbob: @jgfouca - I think it might be more appropriate for MODEL to be the specific model itself, especially in this context. For example, we might need to add compiler options for mpas-o that would not be correct for docn -- and yet they both would be referred to as component "OCN"

jgfouca: @jonbob , I agree. I think we want to support selectors for both model name and type.

rljacob: Looking at how MODEL is used in config_compilers for both cesm and e3sm, its using component names.  It even uses "gptl", "driver" and "csm_share".  So we should continue that here.  @jayeshkrishna please make that change.

rljacob:
I'm trying to figure out how to avoid the confusion of "I configured to use pio1 but "USE_PIO2" is in the logs".    Are these the cases SCORPIO and MPAS want to support?
use pio1 with pio1 APIs
use pio1 with pio2 APIs
use pio2 with pio2 APIs

jayeshkrishna:
I have to read the code again, but I think they only care about

use pio1 with pio1 APIs
use pio2 with pio2 APIs

cime/config/e3sm/machines/config_compilers.xml
rljacob: Has this been resolved?

jayeshkrishna: No, I haven't fixed it in this PR (its already merged to next). I can create a new PR with the changes (In this case however the flags work for all relevant model types - e.g. works for docn too).

PR #3193, opened 2019-09-23
Add Global RCE Compset

whannah1:
Adds various changes to support a new compset for global radiative convective equilibrium (RCE) 'F-EAMv1-RCEMIP', which is configured according to the RCEMIP protocol (more detail can be found at [http://myweb.fsu.edu/awing/rcemip.html](http://myweb.fsu.edu/awing/rcemip.html)). The compset is similar to an aquaplanet, but requires the use of RRTMGP. The additional features added to support RCE are listed below:
- constant SST mode - new namelist variable **sst_constant_value** in docn_in which is set from xml variable **DOCN_AQPCONST_VALUE** in env_run.xml
- Homogenous shortwave radiation (no diurnal cycle) with globally constant solar zenith angle specified in the atm_in namelist with **constant_zenith_deg** (disabled by default with -1)
- constant solar total irradiance at TOA specified with in the atm_in namelist with **fixed_total_solar_irradiance** (disabled by default with -1)
- constant surface albedo (both direct and diffuse) specified in the atm_in namelist with **constant_albedo** (disabled by default with -1)
- A namelist variable **do_aerosol_rad** was added to allow aerosol radiative calculations to be disabled in RRTMGP
- A minor change was made to allow **sday=0** to turn off rotation

New prescribed ozone and initial condition files were created following the RCEMIP protocol and added to the input data repo to allow the new compset to be run for ne4 and ne30 resolutions. 

Temperature clipping was also enabled for RRTMGP when using an aquaplanet because we learned that the code does not support very low temperatures that tend to occur in aquaplanet configurations. Robert Pincus is working to extend support to these lower tempertures for a future RRTMGP update.

[BFB] except for namelist changes


rljacob:
Compare this with similar work in CIME:  https://github.com/ESMCI/cime/pull/3231

whannah1:
Closing this for now. I'm planning to meet with Brian Medeiros at NCAR soon to work out a plan to avoid CIME conflicts. Afterwards I'll remake this branch to utilize NCARs changes to CIME and submit separate PRs to CIME and E3SM. 

It will probably require at least 3 PRs to reproduce the functionality of this branch. I can add the new radiation features to a PR soon, but without the CIME changes the RCE compset will not work, so a PR can be added later that implements the actual compset.

PR #3195, opened 2019-09-23
Fix aquaplanet restart bug with physgrid

whannah1:
This fixes a restart bug when running an aquaplanet case with the FV physgrid. Topography needs to be reset to zero or else mountains of water will appear on restart.


rljacob:
@whannah1 did you mean to close this?

whannah1:
yes, I realized there was a problem with this branch.

PR #3198, opened 2019-09-24
Revert Compy's Intel compiler from 19.0.3(intelmpi) to 18.0.0(mvapich2)

singhbalwinder:
Intel 19.0.3 produces wrong answers when the model is run with CLUBBv2.
We saw this behavior only on Compy. The same compiler works fine on
Cori. All other compilers, libraries and machines work fine.

Intel 18.0.0 doesn't work with intel mpi due to some bug in the mpi
implementation. Therefore, mvapich2 is used instead with intel
18.0.0

[BFB] - Bit-For-Bit

### DIFF
--- a/cime/config/e3sm/machines/config_machines.xml
+++ b/cime/config/e3sm/machines/config_machines.xml
@@ -1869,7 +1869,7 @@
     <NODENAME_REGEX>compy</NODENAME_REGEX>
     <OS>LINUX</OS>
     <COMPILERS>intel,pgi</COMPILERS>
-    <MPILIBS>impi,mvapich2</MPILIBS>
+    <MPILIBS>mvapich2,impi</MPILIBS>
     <SAVE_TIMING_DIR>/compyfs</SAVE_TIMING_DIR>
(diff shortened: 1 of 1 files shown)


rljacob:
All of the pgi tests are failing with: ERROR: Undefined env var 'MKLROOT'
Somehow related to this PR?

singhbalwinder:
Yes, for some reason MKLROOT is not set if we load pgi compiler.
```
sing201@compy01:~$ module purge && module load pgi/18.10 mkl/2018
sing201@compy01:~$ setenv | grep -i mklroot
```
we get no output but if we use intel, we get:
```
sing201@compy01:~$ module purge && module load intel/18.0.0 mkl/2018
sing201@compy01:~$ setenv | grep -i mklroot
MKLROOT=/share/apps/intel/2018/compilers_and_libraries_2018.0.128/linux/mkl
```
I will check ask support to see if they can fix it.

singhbalwinder:
It has been fixed now and I have verified that MKLROOT does show up with the PGI compiler as well.

PR #3074, opened 2019-07-17
Two way irrigation coupling between ELM and MOSART

hydrotian:
Add two-way coupled irrigation scheme to land and river models. 
This function can be turn on/off by a flag in user_nl_clm.

ELM computes the surface water irrigation demand and sends to MOSART and
then MOSART determines the irrigation supply based on the demand and the
water availability and sends back to ELM to constrain the surface water irrigation.

[BFB]
[NML]

PR #3201, opened 2019-09-24
Update CIME to ESMCI cime5.8.9

jgfouca:
Update CIME to ESMCI cime5.8.9

Squash merge of jgfouca/branch-for-to-acme-2019-09-24

Features:
* Add support for running Radiative Convective Equilibrium including new aquaplanet mode 11.
* Add support to archive metadata for lens experiment type.
* Make the streams file legitimate xml files so that we can parse them.

Bug fixes:
* Fix script_regression_tests directory cleanup
* fix issue with workflow prior to run on master
* dont call bcast on mpi_comm_null in seq_flux_mct

[BFB]

### DIFF
--- a/cime/src/components/data_comps/docn/cime_config/config_component.xml
+++ b/cime/src/components/data_comps/docn/cime_config/config_component.xml
@@ -13,7 +13,7 @@
        This file may have ocn desc entries.
   -->
   <description modifier_mode="1">
-    <desc ocn="DOCN[%NULL][%DOM][%SOM][%SOMAQP][%IAF][%SST_AQUAP][%AQP1][%AQP2][%AQP3][%AQP4][%AQP5][%AQP6][%AQP7][%AQP8][%AQP9][%AQP10][%AQPFILE]">DOCN </desc>
+    <desc ocn="DOCN[%NULL][%DOM][%SOM][%SOMAQP][%IAF][%SST_AQUAP][%AQP1][%AQP2][%AQP3][%AQP4][%AQP5][%AQP6][%AQP7][%AQP8][%AQP9][%AQP10][%AQP11][%AQPFILE]">DOCN </desc>
     <desc option="NULL">  null mode</desc>
     <desc option="DOM">  prescribed ocean mode</desc>
     <desc option="SOM">  slab ocean mode</desc>
@@ -30,6 +30,7 @@
     <desc option="AQP8">  analytic aquaplanet sst - option 8</desc>
     <desc option="AQP9">  analytic aquaplanet sst - option 9</desc>
     <desc option="AQP10">  analytic aquaplanet sst - option 10</desc>
+    <desc option="AQP11">  CONSTANT, UNIFORM aquaplanet sst - option 11</desc>
     <desc option="AQPFILE">  file input aquaplanet sst </desc>
   </description>
 
@@ -44,7 +45,7 @@
 
   <entry id="DOCN_MODE">
     <type>char</type>
-    <valid_values>prescribed,sst_aquap1,sst_aquap2,sst_aquap3,sst_aquap4,sst_aquap5,sst_aquap6,sst_aquap7,sst_aquap8,sst_aquap9,sst_aquap10,sst_aquapfile,som,som_aquap,interannual,null</valid_values>
+    <valid_values>prescribed,sst_aquap1,sst_aquap2,sst_aquap3,sst_aquap4,sst_aquap5,sst_aquap6,sst_aquap7,sst_aquap8,sst_aquap9,sst_aquap10,sst_aquap11,sst_aquapfile,som,som_aquap,interannual,null</valid_values>
     <default_value>prescribed</default_value>
     <values match="last">
       <value compset="_DOCN%NULL_">null</value>
@@ -62,6 +63,7 @@
       <value compset="_DOCN%AQP8_">sst_aquap8</value>
       <value compset="_DOCN%AQP9_">sst_aquap9</value>
       <value compset="_DOCN%AQP10_">sst_aquap10</value>
+      <value compset="_DOCN%AQP11_">sst_aquap11</value>
       <value compset="_DOCN%AQPFILE_">sst_aquapfile</value>
     </values>
     <group>run_component_docn</group>
--- a/cime/src/components/data_comps/docn/cime_config/namelist_definition_docn.xml
+++ b/cime/src/components/data_comps/docn/cime_config/namelist_definition_docn.xml
@@ -56,6 +56,7 @@
       <value docn_mode="sst_aquap8">''</value>
       <value docn_mode="sst_aquap9">''</value>
       <value docn_mode="sst_aquap10">''</value>
+      <value docn_mode="sst_aquap11">''</value>
       <value docn_mode="sst_aquapfile">aquapfile</value>
       <value docn_mode="som">som</value>
       <value docn_mode="som_aquap">som</value>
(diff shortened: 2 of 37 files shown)


whannah1:
oh, I didnt know @jedwards4b was also working on RCE stuff! I created a PR the other day that adds a new RCE compset and new features to support it, so we should figure out if any of those things conflict with each other.

cime/src/components/data_comps/docn/cime_config/config_component.xml
```diff
@@ -30,6 +30,7 @@
     <desc option="AQP8">  analytic aquaplanet sst - option 8</desc>
     <desc option="AQP9">  analytic aquaplanet sst - option 9</desc>
     <desc option="AQP10">  analytic aquaplanet sst - option 10</desc>
+    <desc option="AQP11">  CONSTANT, UNIFORM aquaplanet sst - option 11</desc>
```
whannah1: So here's one place where my PR duplicates this functionality. I don't mean to toot my own horn, but I like my implementation better. I added an "AQPCONST" mode. I think more descriptive names are better. But since the names aren't different I don't think there would be a conflict in this file.

cime/src/components/data_comps/docn/cime_config/namelist_definition_docn.xml
```diff
@@ -642,4 +644,16 @@
...
+    <type>real</type>
+    <category>docn</category>
+    <group>docn_nml</group>
+    <desc>
+      Sea Surface Temperature for RCE runs
+    </desc>
+    <values>
+      <value>26.85</value>
+    </values>
+  </entry>
```
whannah1: FYI, in my RCE PR this new variable is called DOCN_AQPCONST_VALUE and is specified through env_run.xml.

cime/src/components/data_comps/docn/mct/docn_comp_mod.F90
```diff
@@ -760,7 +760,7 @@ subroutine prescribed_sst(xc, yc, lsize, sst_option, sst)
 
     ! Control
 
-    if (sst_option < 1 .or. sst_option > 10) then
+    if (sst_option < 1 .or. sst_option > 11) then
        call shr_sys_abort ('prescribed_sst: ERROR: sst_option must be between 1 and 10')
```
whannah1: If we do end up using "11" to be for constant SST we should change the error message here, although I don't think we should do that.

cime/src/drivers/mct/cime_config/namelist_definition_drv.xml
```diff
@@ -4724,4 +4724,29 @@
...
+    <type>real</type>
+    <category>seq_flux_mct</category>
+    <group>seq_flux_mct_inparm</group>
+    <desc>
+       Surface albedo for diffuse radiation
+    </desc>
+    <values>
+      <value>0.07</value>
+    </values>
+  </entry>
```
whannah1: This is kinda odd. Comparing with my PR, I added an ATM namelist variable for the constant albedo. What's the argument for adding it to the drv_in file?

cime/src/share/util/shr_orb_mod.F90
```diff
@@ -57,24 +57,27 @@ real(SHR_KIND_R8) pure FUNCTION shr_orb_cosz(jday,lat,lon,declin,dt_avg)
...
+          if (dt_avg /= 0.0_shr_kind_r8) use_dt_avg = .true.
+       end if
+       ! If dt for the average cosz is specified, then call the shr_orb_avg_cosz
+       if (use_dt_avg) then
+          shr_orb_cosz =  shr_orb_avg_cosz(jday, lat, lon, declin, dt_avg)
+       else
+          shr_orb_cosz = sin(lat)*sin(declin) - &
+               cos(lat)*cos(declin) * &
+               cos((jday-floor(jday))*2.0_SHR_KIND_R8*pi + lon)
+       end if
```
whannah1: Does E3SM use this file?
I implemented my constant zenith mode for E3SM in components/cam/src/utils/orbit.F90

rljacob:
@whannah1 if you have specific questions about how RCE was implemented by @cacraigucar  in CIME you should follow up on the original PR:  https://github.com/ESMCI/cime/pull/3231.

cime/src/components/data_comps/docn/cime_config/config_component.xml
rljacob: Well the convention in docn is to just number the options so you should stick with that or change all the option names to be descriptive.

cime/src/components/data_comps/docn/cime_config/namelist_definition_docn.xml
rljacob: Its a docn thing so should be in the docn namelist.

cime/src/components/data_comps/docn/mct/docn_comp_mod.F90
rljacob: Yes that error message should be modified.

cime/src/drivers/mct/cime_config/namelist_definition_drv.xml
rljacob: Probably because the albedo is calculated in the coupler/driver.

cime/src/share/util/shr_orb_mod.F90
rljacob: Yes E3SM uses this file.  Multiple models need to know the zenith angle so that code is in "share".

cime/src/components/data_comps/docn/cime_config/config_component.xml
whannah1: The numbers come from the original aqua planet MIP protocol, so the original numbers make sense, therefore making 11 = constant SST is not consistent with the previous options and a more description option makes way more sense, like AQPFILE

cime/src/components/data_comps/docn/cime_config/namelist_definition_docn.xml
whannah1: Well, actually I should clarify that my option also goes into the docn namelist variable sst_aquap_constant, but I made it an xml variable so that it could be changed through the compset. The RCEMIP protocol calls for 3 specific SST values, so I thought it would be good to be able to have 3 compsets that correspond to these.

rljacob: That's a good point.  Please open an issue in ESMCI/CIME.

cime/src/share/util/shr_orb_mod.F90
whannah1: Since my RCE compset works by only changing the other file, then I think that means that the atm component is NOT using this file, even though other components might be.

rljacob:
@whannah1 actually since that PR has already been merged to the CIME master, just open an issue with suggested changes.

cime/src/components/data_comps/docn/cime_config/config_component.xml
rljacob: that's a good point.  Please open an issue in ESMCI/cime.

cacraigucar:
When you open a new issue, please include in addition to @cacraigucar, @brianpm who is one of the scientists who proposed the changes.  The other scientist is Kevin Reed (and I'm not sure what his github ID is)  Perhaps Brian can add him to the issue once its posted.

cacraigucar:
adding @kevinareed to this discussion

rljacob:
Issue regarding RCE changes in cime:  https://github.com/ESMCI/cime/issues/3249

jgfouca:
@jedwards4b , this appears to cause a DIFF in SEQ.f19_g16.X, is that expected?

jedwards4b:
If the number of fields in the cpl file changes in an X compset this will cause answer changes in the other fields.   I'm not sure where the change in the number of fields came from but given that the answer change makes sense.

jgfouca:
Thanks, @jedwards4b

jgfouca:
Squash merged to next.

rljacob:
@jgfouca were you worried about the SEQ.f19_g16.X.melvin_gnu diff?  We think that's from PR #3074 and @bishtgautam just forgot to bless it.

jgfouca:
@rljacob , oh, I didn't realize that test was already diffing.

PR #3205, opened 2019-09-25
Fix no_batch determination in case_submit

### LINKED ISSUE #3166: case.submit error on anlworkstation
This issue seems to be caused by PR #3155.
Not sure if other machines with a non-batch system are also affected.

It can be reproduced by running a simple F case (--compset FC5AV1C-H01B --res ne4_ne4)
on anlworkstation.

Before that PR is merged, the output information is:
```
...
mach anlworkstation 
resubmit_num 0
2019-08-29 11:28:43 CASE.RUN HAS FINISHED
Submitted job case.run with id None
```
After that PR is merged, the output information is:
```
...
mach anlworkstation 
resubmit_num 0
2019-08-29 11:38:43 CASE.RUN HAS FINISHED
ERROR: No result from jobs [('case.run', None)]
```
CaseStatus file also shows:
```
...
2019-08-29 11:38:43: case.run success 
 ---------------------------------------------------
2019-08-29 11:38:43: case.submit error 
ERROR: No result from jobs [('case.run', None)]
```

jgfouca:
Fixes #3166 

[BFB]

### DIFF
--- a/cime/scripts/lib/CIME/case/case_submit.py
+++ b/cime/scripts/lib/CIME/case/case_submit.py
@@ -56,8 +56,9 @@ def _submit(case, job=None, no_batch=False, prereq=None, allow_fail=False, resub
     # flag will stay in effect for the duration of the RESUBMITs
     env_batch = case.get_env("batch")
     external_workflow = case.get_value("EXTERNAL_WORKFLOW")
-    if resubmit and env_batch.get_batch_system_type() == "none" or external_workflow:
+    if env_batch.get_batch_system_type() == "none" or resubmit and external_workflow:
         no_batch = True
+
     if no_batch:
         batch_system = "none"
     else:


jgfouca:
@jedwards4b , will you take a quick look at this? This bug was introduced into E3SM in a recent CIME update. My hunch is that no_batch should always be true if env_batch.get_batch_system_type() == "none".

jedwards4b:
There is also a bug a couple of lines down:
replace
```
if batch_system != "none" and env_batch_has_changed or external_workflow:
```
with
```
if batch_system != "none" and (env_batch_has_changed or external_workflow):
```

jgfouca:
@jedwards4b , I think you are talking about this line:

```
if batch_system != "none" and env_batch_has_changed and not external_workflow:
```

I believe this line is correct.

jedwards4b:
hmm - you have an and where I have an or - is that logically the same as adding ()?  I'll need to think about it.

jgfouca:
@jedwards4b , we can resolve in the next split.

PR #3217, opened 2019-09-30
Update coupler bypass/pointclm for offline land

dmricciuto:
Capability for ICB* offline ELM compsets to work with topographic units and  other v2 architecture changes.  New namelist options for adding temperature and co2 at specified times, and for specifying if constant climate is desired in transient simulations (coupler bypass mode only).  Better handling of memory in coupler bypass mode for high resolution simulations.
Moves old "pointclm" directory to "OLMT" - offline land model testbed.  Python scripts updated to work with anvil and compy, supports site, site ensemble site regional/global offline land model simulations.
[BFB]


roger1450:
@dmricciuto  is this version capable of running a continued/resubmitted job?

qzhu-lbl:
@dmricciuto  Is this version capable of running a continued/resubmitted job?

jqyin:
Merge #3217

PR #3222, opened 2019-10-02
Fixed the name for mpaso stream configuration file in the comment

herrwang0:
In the comment section: 
stream.ocean_forward --> stream.ocean.forward

[BFB]

### DIFF
--- a/components/mpas-ocean/cime_config/user_nl_mpaso
+++ b/components/mpas-ocean/cime_config/user_nl_mpaso
@@ -12,6 +12,6 @@
 !
 ! Users cannot use this file to configure streams. In order to configure streams, 
 ! write a stream configuration file, and place it in 
-! SourceMods/src.mpaso/streams.ocean_forward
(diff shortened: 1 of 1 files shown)


jonbob:
It's correct that "stream.ocean_forward" is specified in the comments of this file, but the correct name is "stream.ocean" and not "stream.ocean.forward". I'll fix this to be consistent

herrwang0:
Got it. 

A related question: is there a way similar to actually add streams to stream.ocean (rather than providing a full stream.ocean)?

jonbob:
@herrwang0 - no, we've never figured out a good way to do that. Hopefully it's not too hard to copy the streams.ocean file that gets created in the run directory into the SourceMods/src.mpaso location and add streams to that file?

herrwang0:
Thanks for the response. 
Yes, copy and edit streams.ocean works for me

jonbob:
merged to next

jonbob:
merged to master.
@herrwang0 - you can delete the branch now if you want

PR #3057, opened 2019-07-10
Update MPAS-O: 3d varying GM bolus and 2d varying phase speed

mark-petersen:
In the GM bolus calculation, there is a specified diffusivity and phase
speed (see Ferrari et al. 2010). Currently MPAS-O assumes these values
are fixed in time and space. This PR allows separate flags to enable
2D varying (+time) phase speed (based on the first baroclinic mode phase
speed) and a separate flag for 3D (+time) varying diffusivity.

See https://github.com/MPAS-Dev/MPAS-Model/pull/288

[NML]
[non-BFB]

PR #3226, opened 2019-10-07
Fix threading issue in MPAS-O GM routine

mark-petersen:
This PR fixes a threading issue that was introduced in [E3SM PR #3057](https://github.com/E3SM-Project/E3SM/pull/3057) that brought in the initial implementation of 3D varying GM bolus. After that commit, some E3SM threading tests were observed to fail with different results.

Tested with:
* SMS_P12x2.T62_oQU240.CMPASO-NYF.compy_intel - BFB results 10 times
* SMS_P12x2.ne4_oQU240.A_WCYCL1850.compy_intel.allactive-mach_mods - BFB results 10 times

This last test is one that was failing on sandiatoss3.


mark-petersen:
Corresponding MPAS PR at https://github.com/MPAS-Dev/MPAS-Model/pull/376

jonbob:
merged to next

PR #3228, opened 2019-10-08
Update seaice version in mpas-source submodule to get index allocation fix

jonbob:
This PR brings in a new mpas-source submodule that updates the seaice code to fix an index allocation error (see [MPAS-Model PR #355](https://github.com/MPAS-Dev/MPAS-Model/pull/355)). The MPAS PR fixed one bug:
- [MPAS-Model Issue #344](https://github.com/MPAS-Dev/MPAS-Model/issues/344)

correcting a compile time error that appears with intel19 compiler versions.

[BFB]


jonbob:
merged to master

PR #3230, opened 2019-10-10
Update mpas-seaice version in mpas-source submodule

jonbob:
This PR brings in a new mpas-source submodule that updates the version of mpas-seaice as well as supporting scripts changes. The mpas-seaice update brings in four separate merges, only one of which impacts E3SM. Those merges are:
* seaice standalone testing forcing README and files (ee79424);
* zbgc modifications that should only impact certain BGC compsets (340e782);
* ifdef'd threading support for GNU-openmp (93bbc4e); and
* strain/stress standalone testing addition (818b069).

The only impact on E3SM should be for BGC compsets, though there will be NML changes.

[NML]
[BFB]


jonbob:
merged to next

jonbob:
Merged to master but still need to bless NML DIFFs on cori and sandiatoss3, whenever they report

jonbob:
NML and real BGC DIFF's blessed

Issue #3150, opened 2019-08-23
Reduce debugging output from MOSART


amametjanov:
At https://github.com/E3SM-Project/E3SM/blob/master/components/mosart/src/riverroute/RtmMod.F90#L2267:

> Please comment out or surround this line with #ifndef NDEBUG ... #endif to activate it for debug runs. A 5-day run produced 122721 writes to e3sm.log.

And at https://github.com/E3SM-Project/E3SM/blob/master/components/mosart/src/riverroute/RtmTimeManager.F90#L734:

> Please comment out or surround this write with #ifndef NDEBUG ... #endif.

rljacob:
@hydrotian please fix this issue.

rljacob:
@liho745 please fix this issue.

PR #3234, opened 2019-10-14
Improve throughput on Compy

amametjanov:
Improve throughput on Compy by reducing stdout writes:  Turn off MOSART debug writes and increase tolerance in an ELM BalanceCheck.  Also add a throughput tolerance check to compy.

Fixes E3SM-Project/E3SM#3150

[BFB]

### DIFF
--- a/cime/scripts/lib/CIME/SystemTests/system_tests_common.py
+++ b/cime/scripts/lib/CIME/SystemTests/system_tests_common.py
@@ -474,7 +474,7 @@ class SystemTestsCommon(object):
                         diff = (baseline - current)/baseline
                         tolerance = self._case.get_value("TEST_TPUT_TOLERANCE")
                         if tolerance is None:
-                            tolerance = 0.25
+                            tolerance = 0.1
                         expect(tolerance > 0.0, "Bad value for throughput tolerance in test")
                         if diff < tolerance and self._test_status.get_status(THROUGHPUT_PHASE) is None:
                             self._test_status.set_status(THROUGHPUT_PHASE, TEST_PASS_STATUS)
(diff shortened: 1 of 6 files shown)


amametjanov:
Before these updates:
```
    TOT Run Time:     821.140 seconds       26.488 seconds/mday         8.94 myears/wday
```
After:
```
    TOT Run Time:     668.517 seconds       21.565 seconds/mday        10.98 myears/wday
```
The new throughput (with today's master) is now within 2% (as opposed to 20% slower) compared to throughput of the same case built with master from July 30 (hash `a089690`):
```
    TOT Run Time:     656.696 seconds       21.184 seconds/mday        11.17 myears/wday
```

rljacob:
Your description was not detailed enough.  Enter the new one when this merged to master.

singhbalwinder:
So, what are the reasons for this performance improvement? Is it simply the removal of log file write statements or are there some other flags involved as well?

rljacob:
Writing lots of output to stdout every timestep is expensive.

cime/scripts/lib/CIME/SystemTests/system_tests_common.py
```diff
@@ -474,7 +474,7 @@ def _compare_throughput(self):
                         diff = (baseline - current)/baseline
                         tolerance = self._case.get_value("TEST_TPUT_TOLERANCE")
                         if tolerance is None:
-                            tolerance = 0.25
+                            tolerance = 0.1
```
jgfouca: @amametjanov , I just noticed this change during my latest CIME split. Can you explain why this change was needed? It's an impactful change since it changes the default fallback tput tolerance (when no TEST_TPUT_TOLERANCE is defined in config_machines.xml) on every machine for both E3SM and CESM.

amametjanov: Throughput tolerance was both too wide (25%) and ignored (not a real regression test fail). For tests with baselines, it should be useful to know about greater than 10% variance (even if TEST_TPUT_TOLERANCE is not defined for the machine). This is what happened on Compy without any notice.

jgfouca: @amametjanov , thanks

PR #3235, opened 2019-10-15
Adjust max node number for sandiatoss3 short q

oksanaguba:
Adjust max node number for sandiatoss3 short q.

According to help for short q,
'Walltime limit is 4 hours, node limit is 24 nodes per job.'

[BFB]

### DIFF
--- a/cime/config/e3sm/machines/config_batch.xml
+++ b/cime/config/e3sm/machines/config_batch.xml
@@ -409,7 +409,7 @@
 
   <batch_system MACH="sandiatoss3" type="slurm" >
     <queues>
-      <queue nodemax="24" walltimemax="04:00:00" strict="true" default="true">short,batch</queue>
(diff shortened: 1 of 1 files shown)


jgfouca:
Will this work for both skybridge and chama?

oksanaguba:
'Walltime limit is 4 hours, node limit is16 for chama.' , so, I will change it to 16. 24 is for skybridge.

oksanaguba:
How is it decided which tests run on champ or on skybridge? I login to skybridge only and always see those 2 big tests there.

rljacob:
Its always the integration suite but skybridge runs on "next" and chama on "master".

jgfouca:
@oksanaguba , the Jenkins test drives decide, but we want sandiatoss3 to be able to accommodate both. It sounds like 16 is the right value.

PR #3238, opened 2019-10-16
Flow of Agricultural Nitrogen (FAN) 

juliusvira:
This PR contains a prototype implementation of the FAN model (https://www.geosci-model-dev-discuss.net/gmd-2019-233/) in the E3SM Land Model. The model evaluates volatilization losses of nitrogen from fertilizers and livestock manure. The N fertilization rate and timing is determined by the crop model; the manure N is read from an additional stream file.

At present, the FAN code is only "one-way" coupled to the ELM N cycle: FAN uses the fertilization applied in crop model, but the losses evaluated by FAN are fed back to ELM. Therefore, enabling FAN should not change answers for any existing variables in ELM. The NH3 losses evaluated by FAN/ELM are generally similar to those in the CLM implementation except where the fertilizer applications differ.

### DIFF
--- a/cime/scripts/Tools/jenkins_generic_job
+++ b/cime/scripts/Tools/jenkins_generic_job
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 """
-Jenkins runs this script to perform a test of an e3sm 
+Jenkins runs this script to perform a test of an e3sm
 test suite. Essentially, a wrapper around create_test and
 wait_for_tests that handles cleanup of old test results and
 ensures that the batch system is left in a clean state.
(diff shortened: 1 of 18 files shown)


bishtgautam:
@juliusvira, Could you rebase your branch of the latest e3sm master to resolve the conflicts? thanks.

rljacob:
@juliusvira let us know if you need help rebasing.

juliusvira:
The problem is that I'm no longer at Cornell and have very little time to work on this. There seem be some conflicts but I have not yet had a chance to see how difficult they will be to resolve.

rljacob:
Closing until a new developer can be found.

PR #3249, opened 2019-10-23
Initial support for tri-grid configurations

jonbob:
This PR brings in scripts changes to support the first tri-grid configurations. It's intended to be able to run land-only, AMIP, and fully-coupled compsets at specific resolutions on which land and runoff are both on either the r05 or r0125 grids. Specifically it supports:
* r05_r05.I1850CLM45CN
* r0125_r0125.I1850CLM45CN
* ne30_r05_ne30.F20TRC5-CMIP6
* ne30_r0125_ne30.F20TRC5-CMIP6
* ne30_r05_oECv3.A_WCYCL1850
* ne30_r0125_oECv3.A_WCYCL1850

### DIFF
--- a/components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml
+++ b/components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml
@@ -221,6 +221,8 @@ lnd/clm2/surfdata_map/surfdata_48x96_simyr2000_c130927.nc</fsurdat>
 <fsurdat hgrid="64x128"    sim_year="2000">
 lnd/clm2/surfdata_map/surfdata_64x128_simyr2000_c170111.nc</fsurdat>
 
+<fsurdat hgrid="0.125x0.125"  sim_year="2000" use_crop=".false." >
+lnd/clm2/surfdata_map/surfdata_0.125x0.125_simyr2000_c190730.nc</fsurdat>
 <fsurdat hgrid="0.9x1.25"  sim_year="2000" use_crop=".false." >
 lnd/clm2/surfdata_map/surfdata_0.9x1.25_simyr2000_c180404.nc</fsurdat>
 <fsurdat hgrid="1.9x2.5"   sim_year="2000" use_crop=".false." >
@@ -284,6 +286,8 @@ lnd/clm2/surfdata_map/surfdata_360x720cru_simyr1850_c180216.nc</fsurdat>
 <fsurdat hgrid="48x96"        sim_year="1850" use_crop=".false." >
 lnd/clm2/surfdata_map/surfdata_48x96_simyr1850_c130927.nc</fsurdat>
 
+<fsurdat hgrid="0.125x0.125"  sim_year="1850" use_crop=".false." >
+lnd/clm2/surfdata_map/surfdata_0.125x0.125_simyr1850_c190730.nc</fsurdat>
 <fsurdat hgrid="0.9x1.25"     sim_year="1850" use_crop=".false." >
 lnd/clm2/surfdata_map/surfdata_0.9x1.25_simyr1850_c180306.nc</fsurdat>
 <fsurdat hgrid="1.9x2.5"      sim_year="1850" use_crop=".false." >
@@ -1405,7 +1409,7 @@ this mask will have smb calculated over the entire global land surface
 <map frm_hgrid="0.5x0.5"    frm_lmask="AVHRR"  to_hgrid="0.5x0.5"   to_lmask="nomask" 
 >lnd/clm2/mappingdata/maps/0.5x0.5/map_0.5x0.5_AVHRR_to_0.5x0.5_nomask_aave_da_c111021.nc</map>
 <map frm_hgrid="0.5x0.5"    frm_lmask="MODIS"  to_hgrid="0.5x0.5"   to_lmask="nomask" 
->lnd/clm2/mappingdata/maps/0.5x0.5/map_0.5x0.5_MODIS_to_0.5x0.5_nomask_aave_da_c111021.nc</map>
+>lnd/clm2/mappingdata/maps/0.5x0.5/map_0.5x0.5_MODIS_to_0.5x0.5_nomask_aave_da_c190417.nc</map>
 <map frm_hgrid="0.5x0.5"    frm_lmask="nomask"  to_hgrid="0.5x0.5"   to_lmask="nomask" 
 >lnd/clm2/mappingdata/maps/0.5x0.5/map_0.5x0.5_nomask_to_0.5x0.5_nomask_aave_da_c111021.nc</map>
 <map frm_hgrid="10x10min"    frm_lmask="IGBPmergeICESatGIS"  to_hgrid="0.5x0.5"   to_lmask="nomask" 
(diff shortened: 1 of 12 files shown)


jonbob:
Testing with test-merge to master:
* SMS.r05_r05.I1850CLM45CN.compy_intel - PASS
* SMS.r0125_r0125.I1850CLM45CN.compy_intel - FAIL
* SMS.ne30_r05_ne30.F20TRC5-CMIP6.compy_intel - FAIL (PASS with pio type netcdf for ATM)
* SMS.ne30_r0125_ne30.F20TRC5-CMIP6.compy_intel - FAIL (PASS with pio type netcdf for ATM)
* SMS.ne30_r05_oECv3.A_WCYCL1850.compy_intel - FAIL (missing finidat file for ELM, PASS with finidat = ' ' in user_nl_clm)
* SMS.ne30_r0125_oECv3.A_WCYCL1850.compy_intel - IN QUEUE

TO FIX: 
* Solar_1850-2299_input4MIPS_c20171207.nc replace or change to netcdf3 
  @singhbalwinder - there are two files for this with different date stamps, one of which
  is netcdf3 and the other is HDF5. Comparing with cprnc says the files are identical -- is it OK
  to change CAM scripts to point at the netcdf3 version?

* missing finidat file for SMS.ne30_r05_oECv3.A_WCYCL1850.compy_intel -- @bishtgautam ?

* wrong params file for SMS.r0125_r0125.I1850CLM45CN.compy_intel? incorrect definition? @bishtgautam ?

singhbalwinder:
Yes, if the files are identical, I don't see a reason not to switch to netcdf3. Thanks @jonbob .

jonbob:
Thanks @singhbalwinder - I think the files were simply converted by @sarich but not in the scripts. I'll make that change in this PR and add you to the reviewers, if you'd like

rljacob:
Yes all the configs should point to the netcdf3 files.

rljacob:
All *new* configs should point to and use only netcdf3 files.   The old ones will be updated soon.

jonbob:
@rljacob - these were not new use-case configs, just old ones that point to a netcdf3 issue in testing. I've switched the files they point at and re-testing

rljacob:
Ok that's fine.

bishtgautam:
@jonbob The `SMS.r0125_r0125.I1850CLM45CN.compy_intel` was crashing in the land component because the model was running out of memory. After I increased `NTASKS_LND` from 160 to 320, the land model initialized successfully. But, the model crashed during MOSART initialization, which I suspect is also related to model running out of memory. I have increased NTASKS for ATM and OCN to 320 and submitted a new job, which is in queue now. I will give an update after the job finishes.

jonbob:
Thanks for tracking that down, @bishtgautam . And it makes all kinds of sense -- I hadn't realized how small the layout was

components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml
```diff
@@ -236,6 +248,10 @@ lnd/clm2/surfdata_map/surfdata_ne4np4_simyr2000_c160614.nc</fsurdat>
 lnd/clm2/surfdata_map/surfdata_ne4pg2_simyr2000_c190620.nc</fsurdat>
 <fsurdat hgrid="ne240np4"      sim_year="2000" use_crop=".false." >
 lnd/clm2/surfdata_map/surfdata_ne240np4_simyr2000_c170821.nc</fsurdat>
+<fsurdat hgrid="r05"         sim_year="2000" >
+lnd/clm2/surfdata_map/surfdata_0.5x0.5_simyr2000_c190418.nc</fsurdat>
+<fsurdat hgrid="r0125"       sim_year="2000" use_crop=".false." >
+lnd/clm2/surfdata_map/surfdata_0.125x0.125_simyr2000_c190730.nc</fsurdat>
```
tangq: We've decided to switch to year 2010 for the current climate condition. Year 2010 files will be needed.

components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml
```diff
@@ -355,6 +377,13 @@ lnd/clm2/surfdata_map/surfdata_twpx4v1_simyr2000_c170706.nc</fsurdat>
 <flanduse_timeseries hgrid="ne11np4" sim_year_range="1850-2000"
  use_crop=".false."  >lnd/clm2/surfdata_map/surfdata.pftdyn_ne11np4_hist_simyr1850-2005_c160803.nc</flanduse_timeseries>
 
+<flanduse_timeseries hgrid="r0125" sim_year_range="1850-2000"
+ use_crop=".false." >lnd/clm2/surfdata_map/landuse.timeseries_0.125x0.125_hist_simyr1850-2015_c191004.nc</flanduse_timeseries>
+
+<flanduse_timeseries hgrid="r05" sim_year_range="1850-2000"
+  use_crop=".false." >lnd/clm2/surfdata_map/landuse.timeseries_0.5x0.5_hist_simyr1850-2015_c191004.nc</flanduse_timeseries>
```
tangq: Should the sim_year_range be 1850-2015, if it is for the historical transient and AMIP runs?

jonbob:
@bishtgautam - can you help me with @tangq's comments? Do we need different surface and landuse files?

tangq:
@jonbob , I was checking the r0125 changes that might need to be added to the North America RRM PR and came across the files and made the suggestions above.

When I was creating the F2010 compsets, at least because of the land use and land cover data is different for years 2000 and 2010, I had to update the land surface data files.

bishtgautam:
@jonbob, @tangq, I believe we don't need to update LULC files but I might have to generate new land surface file. Let me look into this and get back to you.

tangq:
To clarify...For the LULC files, I only suggested to update the "sim_year_range" settings to 1850-2015, not the data files. I am not sure if "sim_year_range" will be pick up and used by the model to determine the period of the files. If so, it may cause errors.

bishtgautam:
> To clarify...For the LULC files, I only suggested to update the "sim_year_range" settings to 1850-2015, not the data files

That is exactly right and for this reason I think there won't be a need to generate new files. I will add support for `sim_year_range = 1850-2015` if it doesn't exist already, and make other necessary changes.

jonbob:
With thanks to @bishtgautam and @singhbalwinder, here's the current testing status:
- SMS.r05_r05.I1850CLM45CN.compy_intel - PASS
- SMS.r0125_r0125.I1850CLM45CN.compy_intel - PASS (on 640 pes)
- SMS.ne30_r05_ne30.F20TRC5-CMIP6.compy_intel - PASS
- SMS.ne30_r0125_ne30.F20TRC5-CMIP6.compy_intel - PASS
- SMS.ne30_r05_oECv3.A_WCYCL1850.compy_intel - FAIL (missing finidat file for ELM, PASS with finidat = ' ' in user_nl_clm)
- SMS.ne30_r0125_oECv3.A_WCYCL1850.compy_intel - PASS

jonbob:
Thanks @bishtgautam - that fixed the last failing test. Once we address @tangq's comments, this should be about ready

bishtgautam:
@tangq, The LULC files are applicable to `1850-2015`, but the current surface dataset is for 2000. I have created new surface datasets for 2010. But, I'm not very inclined to set surface dataset with 2010 LULC for `sim_year="2000"` because the compset name is then misleading. 

Here are few questions to you and broader community on changes to the compsets.

- If we are to switch the present day to be 2010 instead of 2000, which of the following option are we going to choose (or there is an another option)? 
  - Redefine existing compsets (i.e. `FC5AV1C-L` will be become `2010_CAM5%AV1C-L_CLM45%SPBC_CICE%PRES_DOCN%DOM_SROF_SGLC_SWAV`  instead of `2000_CAM5%AV1C-L_CLM45%SPBC_CICE%PRES_DOCN%DOM_SROF_SGLC_SWAV`), 
  - Add new compests (i.e. `FC5AV1C-2010-L` as `2010_CAM5%AV1C-L_CLM45%SPBC_CICE%PRES_DOCN%DOM_SROF_SGLC_SWAV`)
  - Keep existing compsets as is but modify certain files under the hood (i.e.  for `sim_year="2000"`, the land surface dataset actually points to year 2010)

- Regarding transient runs:
  - Similarly, are we going to redefine all `20TR` compsets (i.e. `./query_config --compset all | grep 20TR`) to be from 1850-2015? Or, we going to create new compsets? Or, are we going to keep the compsets as is and modify files under the hood?

These comments are not specific to this PR

tangq:
@bishtgautam , we are not updating the old `sim_year="2000"` compsets. Instead, we created new F2010 compsets to avoid confusion.

The new F2010 compsets (F2010C5-CMIP6-LR, F2010C5-CMIP6-HR) were first created on maint-1.0. On the current master, the low-res version (F2010SC5-CMIP6) is already available and the high-res version will be there soon. We need the 2010 land surface data to support these 2010 compsets. So, the new `sim_year=2010` compsets use the land surface data of 2010. The old `sim_year=2000` compsets keep using 2000 surface data -- No under-the-hood tweaks.

For the transient cases, only those of CMIP6 will need to be 1850-2015. The old CMIP5 ones will stay as 1850-2005.

bishtgautam:
>The old `sim_year=2000` compsets keep using 2000 surface data -- No under-the-hood tweaks.

This is great. 

Following land surface dataset for `sim_year=2010` are available on E3SM inputdata server:
- `surfdata_0.5x0.5_simyr2010_c191025.nc`
- `surfdata_0.125x0.125_simyr2010_c191025.nc`

We can use these new surface datasets when new F2010 compsets are merged from maint-1.0 to master or directly created on master.

tangq:
> Following land surface dataset for `sim_year=2010` are available on E3SM inputdata server:
> 
>     * `surfdata_0.5x0.5_simyr2010_c191025.nc`
> 
>     * `surfdata_0.125x0.125_simyr2010_c191025.nc`

@bishtgautam , we need to add the default name list settings to pick up these files for r05 and r0125 grids when `sim_year=2010`, right?

bishtgautam:
>@bishtgautam , we need to add the default name list settings to pick up these files for r05 and r0125 grids when sim_year=2010, right?

But, this PR is not adding any compsets for `sim_year=2010`.

tangq:
@bishtgautam , I meant adding changes to components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml. For example, for r0125 the changes should be something similar to

`<fsurdat hgrid="r0125"       sim_year="2010" use_crop=".false." >
lnd/clm2/surfdata_map/surfdata_0.125x0.125_simyr2010_c191025.nc</fsurdat>`

bishtgautam:
One could make the changes you have suggested in `namelist_defaults_clm4_5.xml`, but since there is no compset (at least not yet in this PR) that can use the 2010 surface dataset, why make that change? Moreover, one needs to test if that surface dataset correctly works.

tangq:
I agree that we can make these changes later when adding the tri-grid support for 2010 compsets.

jonbob:
@tangq - are your issues with this PR all resolved? It sounds like we'll add the 2010 compsets later and include support for these grids?

jonbob:
@rljacob - can you please review this PR soon? I think it's about ready, but I would very much appreciate another set of eyes on the additions to the automated testing suites.

tangq:
@jonbob , yes, they are resolved. We can add them in future PRs. This PR is good to go now.

components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml
```diff
@@ -1666,35 +1685,37 @@ this mask will have smb calculated over the entire global land surface
 <!-- Created by lnd/clm/bld/namelist_files/createMapEntry.pl--> 
 
 <map frm_hgrid="0.5x0.5"    frm_lmask="AVHRR"  to_hgrid="0.125x0.125"   to_lmask="nomask" 
->lnd/clm2/mappingdata/maps/0.125x0.125/map_0.5x0.5_AVHRR_to_0.125x0.125_nomask_aave_da_c140702.nc</map>
+>lnd/clm2/mappingdata/maps/0.125x0.125/map_0.5x0.5_AVHRR_to_0.125x0.125_nomask_aave_da_c190725.nc</map>
```
rljacob: Is this switching to a netcdf3 version of this file?

bishtgautam: What ncdump option can I use to verify if the file is netcdf3?

tangq: I use `ncdump -k`

bishtgautam: Found it. `ncdump -k`

bishtgautam: The file is in `classic` format, but the file is not used by the model. It is used during preprocessing to create land surface dataset. Do you want me to regenerate all files?

rljacob: No only if its read in by the model.   But this raises a new question:  why are files not needed by the model in the namelist_defaults_clm4_5.xml?

bishtgautam: The `mksurfdata_map` (pre-processing tools) needs information about mapping files for remapping raw datasets onto new land surface resolution. The information about the mapping files in stored in the `namelist_defaults_clm4_5.xml`.

tangq: To answer @rljacob 's question, my understanding is that putting these mapping file info in the namelist_defaults_clm4_5.xml ensures that the pre-processing tools and the model are consistent by using the same *.xml files to provide settings.

jonbob:
merged to next

dqwu:
@jonbob 
The new test SMS.r05_r05.I1850CLM45CN failed on melvin, see:
https://my.cdash.org/testDetails.php?test=50894161&build=1722007

There are no run time errors, but some input files are missing:
```
  Model clm missing file fsurdat = '/sems-data-store/ACME/inputdata/lnd/clm2/surfdata_map/surfdata_0.5x0.5_simyr1850_c190418.nc'
Failed to find or download file /ascldap/users/jenkins/acme/scratch/SMS.r05_r05.I1850CLM45CN.melvin_gnu.C.JNextDeveloper20191029_221231/run/inputdata_checksum.dat
  Model clm missing file fatmlndfrc = '/sems-data-store/ACME/inputdata/share/domains/domain.lnd.r05_oEC60to30v3.190418.nc'
Failed to find or download file /ascldap/users/jenkins/acme/scratch/SMS.r05_r05.I1850CLM45CN.melvin_gnu.C.JNextDeveloper20191029_221231/run/inputdata_checksum.dat
```

jonbob:
Merged to master
Blessed new baselines on melvin and anvil but will have to modify the pe-layouts  for new tests on sandiatoss3

PR #3204, opened 2019-09-25
Fix physgrid issues with PE layout and aqua restarts

whannah1:
This fixes an issues where aqua planet cases were crashing on restart due to not correctly setting the topography to zero. Instead the realistic topography was added on restart, which caused the run to crash. This PR resolves this issue by explicitly resetting the topography to zero when restarting a physgrid run.
 
This also fixes an issue running certain physgrid configurations, such as ne4pg2, because the default NTASK was set higher than the number of physics columns. To get the physgrid specification to match correctly in config_pesall.xml an underscore was to be added to other grid query strings. This should not affect non-physgrid cases. The underscore is used because it always separates the atmosphere grid from the other grids. For example, A typical ne4 grid string is
a%ne4np4_l%ne4np4_oi%ne4np4_r%null_g%null_w%null_z%null_m%oQU240
But the string for ne4pg2 is
a%ne4np4.pg2_l%ne4np4_oi%ne4np4_r%null_g%null_w%null_z%null_m%oQU240
so in order to differentiate the grids we use "ne4np4_" and "ne4np4.pg2" to get different default PE layouts. Using "ne4np4" without the underscore leads to an error when running physgrid cases.

After some discussion we decided to also include a change to the default physgrid mapping algorithm to use the high-order option. This will make all pg2 tests non-BFB.

[BFB] except for SMS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L

PR #3251, opened 2019-10-23
Change se_fv_phys_remap_alg default to 1

whannah1:
This changes the default physgrid mapping algorithm to the higher order version, which has proven to produce more realistic results.

[BFB] except for pg2 tests that do not use theta dycor option

### DIFF
--- a/components/cam/bld/namelist_files/namelist_defaults_cam.xml
+++ b/components/cam/bld/namelist_files/namelist_defaults_cam.xml
@@ -1572,7 +1572,7 @@
 <semi_lagrange_cdr_check> .false. </semi_lagrange_cdr_check>
 <semi_lagrange_nearest_point_lev> 0 </semi_lagrange_nearest_point_lev>
 
-<se_fv_phys_remap_alg> 0 </se_fv_phys_remap_alg>
+<se_fv_phys_remap_alg> 1 </se_fv_phys_remap_alg>
 
 <!-- ================================================================== -->
 <!-- Defaults for driver namelist seq_infodata_inparm                   -->


wlin7:
@whannah1 , this change  will not impact the common tests that are built with NGP=0. Which nightly test is expected to diff and to watch for? Tests like SMS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L.sandiatoss3_intel.cam-thetahy_pg2?

whannah1:
@wlin7 I'm pretty sure the theta tests already use the high-order mapping option, but any other pg2 test will diff. The tests with np4 should be unaffected.

wlin7:
Thanks @whannah1 . Can you add these to the PR message? Not going to merge this today as there is another nonbfb going in.

wlin7:
Merged to next.  SMS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L test on compy, sandiatoss3 and melvin are expected to diff, both in NML and simulation results

wlin7:
SMS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L test on compy, sandiatoss3 and melvin are not affected. The parameter se_fv_phys_remap_alg is not involved. 

@whannah1 , is this actually expected? I thought these are the pg2 tests that would see DIFF.

ambrad:
In components/homme/src/share/control_mod.F90:201, set
```
    integer, public :: se_fv_phys_remap_alg = 1
```

whannah1:
Closing this and moving the changes to the branch for PR #3204 because this is needed to get passing pg2 tests.

PR #3259, opened 2019-10-28
Updates MPP to fix error on Melvin

bishtgautam:
Corrects a `#endif` statement to avoid compiler error on Melvin.

[BFB]


oksanaguba:
Why did the test pass today?

jqyin:
Merged #3259

PR #3263, opened 2019-10-29
Update mpas-source, COMPASS python scripts

mark-petersen:
This PR brings in a new mpas-source submodule that only contains an update to the MPAS-Ocean internal COMPASS testing suite. It is not used by E3SM and should have no impact.

[BFB]


jonbob:
merged to next

jonbob:
merged to master

PR #3264, opened 2019-10-29
Add (or re-add) support for NLDAS configurations

jonbob:
This PR adds support for NLDAS configurations, for use mostly by MOSART developers. There are new compsets, as well as support for an NLDAS_NLDAS grid. The new compsets are:
* INLDASCNPECACNTBC
  (2000_DATM%NLDAS_CLM45%CNPECACNTBC_SICE_SOCN_MOSART_SGLC_SWAV)
* IM20TRNLDASCNPECACNTBC
  (20TR_DATM%NLDAS_CLM45%CNPECACNTBC_SICE_SOCN_MOSART_SGLC_SWAV)
* RM
  (2000_SATM_DLND%LCLMR_SICE_SOCN_MOSART_SGLC_SWAV)
There are new modes added to both datm and dlnd in support of these configurations.

Not clear if this should be added to automated testing suite 

[BFB] for all current tests

### DIFF
--- a/components/mosart/cime_config/config_compsets.xml
+++ b/components/mosart/cime_config/config_compsets.xml
@@ -10,7 +10,7 @@
   <!---MOSART RM compsets -->
 
   <compset>
-    <alias>RM</alias>
+    <alias>RMOSARTNLDAS</alias>
     <lname>2000_SATM_DLND%LCLMR_SICE_SOCN_MOSART_SGLC_SWAV</lname>
   </compset>
 
--- a/components/mosart/cime_config/config_compsets.xml
+++ b/components/mosart/cime_config/config_compsets.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0"?>
+
+<compsets>
+
+  <help>
+  Compset format is explained in cime/config/e3sm/allactive/config_compsets.xml
+  or in the CIME documenation
+  </help>
+
+  <!---MOSART RM compsets -->
+
+  <compset>
+    <alias>RM</alias>
+    <lname>2000_SATM_DLND%LCLMR_SICE_SOCN_MOSART_SGLC_SWAV</lname>
+  </compset>
+
+</compsets>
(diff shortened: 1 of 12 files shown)


tanzeli1982:
@jonbob, I have only one minor comment. I noticed that there already exist NLDAS2 and CLMNLDAS2 in the code. Have they had any use? Must we keep both?

jonbob:
@tanzeli1982 - CESM must have added the NLDAS2 support to CIME, and since we share CIME I think we have to keep both. I looked at NLDAS2 and it seems more complicated than the version you want -- do you know anything about it?

components/mosart/cime_config/config_compsets.xml
```diff
@@ -0,0 +1,17 @@
...
+
+  <help>
+  Compset format is explained in cime/config/e3sm/allactive/config_compsets.xml
+  or in the CIME documenation
+  </help>
+
+  <!---MOSART RM compsets -->
+
+  <compset>
+    <alias>RM</alias>
```
hydrotian: How about we change this alias to something more conventional? For example "RMOSARTNLDAS" - "R" for River case, "MOSART" for the model name, and "NLDAS" for the grid?

jonbob:
@tanzeli1982 - I don't think I have an fsurdat file for the IM20TRNLDASCNPECACNTBC compset, just the year 2000 one.

tanzeli1982:
> @tanzeli1982 - I don't think I have an fsurdat file for the IM20TRNLDASCNPECACNTBC compset, just the year 2000 one.

That file should work.

components/mosart/cime_config/config_compsets.xml
tanzeli1982: > I'm fine with the current ones but I will let Zeli decide.

I am fine with the current one too.

jonbob:
merged to next

jonbob:
merged to master

hydrotian:
@jonbob Thanks!

tanzeli1982:
@jonbob Thanks a lot.

hydrotian:
@jonbob I was trying to use RMOSARTNLDAS compset on current master but got error message like
```
 4:  (shr_strdata_init)  ERROR: dmodel_domain_fracname_from_stream
 4:  can only be provided when taking the data model domain
 4:  from the domain of the first stream
 4:  (i.e., when the domain file is null).
```
Any idea? Thanks!

jonbob:
@hydrotian - can you please tell me exactly what case you were trying to run? Especially the grid resolution?

hydrotian:
@jonbob I was creating a new case using
```
--res NLDAS_NLDAS --compset RMOSARTNLDAS
```

jonbob:
@hydrotian - thanks. I'm getting the same error. Let me see what I can figure out....

rljacob:
Guess these should have been added to the test suite?

PR #3267, opened 2019-10-29
Physgrid: Introduce new topography file format.

ambrad:
Improve treatment of topography for physgrid configurations.  Better GLL/PG consistency and error checking when reading netcdf files.

[BFB]


ambrad:
e3sm_developer testing in progress.

See Topography section on the [physgrid results page](https://acme-climate.atlassian.net/wiki/spaces/COM/pages/1010171942/Physgrid+results) for details.

components/cam/src/dynamics/se/inidat.F90
```diff
@@ -451,30 +452,45 @@ subroutine read_inidat( ncid_ini, ncid_topo, dyn_in)
...
+      read_pg_grid = fv_nphys > 0 .and. se_fv_phys_remap_alg == 0
+      if (read_pg_grid) then
          ! Copy phis field to GLL grid
          call infld(fieldname, ncid_topo, 'ncol', 1, nphys_sq, &
               1, nelemd, phys_tmp, found, gridname='physgrid_d')
-         if (se_fv_phys_remap_alg == 0) then
-            call fv_phys_to_dyn_topo(elem,phys_tmp)
+         call fv_phys_to_dyn_topo(elem,phys_tmp)
+      else
+         ! Attempt to read a mixed GLL-FV topo file, which contains PHIS_d in
```
whannah1: I think we need a better description of what happens when the topo is on the GLL grid. Maybe I'm behind on this discussion, but last time we talked about it I thought we were sure that the topo used by the dynamics needs to be defined on the FV grid for the purpose of the the parameterizations that handle the "sub-grid" topography that live in physics world. 
Is this change just a convenience so we don't have to create the topo on the FV grid? In other words, it allows us to use the conventional smoothing of topo on the GLL grid rather than defining a new smoothing operator on the FV grid?

components/homme/src/tool/tool_main.F90
```diff
@@ -0,0 +1,109 @@
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+! This program is an offline tool. See the documentation in the following
+! subroutines, implemented below; each is a tool:
+!   * topo_convert
```
whannah1: Maybe I'm just not understanding this, but the name "tool_main" seems very generic. Can we be more descriptive of what this is actually used for? It seems to be an offline file processing tool, would that be accurate?

components/cam/src/dynamics/se/inidat.F90
```diff
@@ -451,30 +452,45 @@ subroutine read_inidat( ncid_ini, ncid_topo, dyn_in)
           end do
     end do
 
+    read_pg_grid = .false.
```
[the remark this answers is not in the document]
ambrad: This PR will go in after that one, so I'll merge master at the end when this PR is ready.

components/homme/src/tool/tool_main.F90
ambrad: Yes. Homme previously did not have a clear offline tool system, just workarounds. This new executable will be the place for any new offline tool to live. How about the name offline_tool_main? (I'd like to keep the directory name as 'tool' because Homme directory names tend to be short, which is kind of nice.)

components/cam/src/dynamics/se/inidat.F90
ambrad: First, let me quote a commit message so we have this information available in the conversation:

I. Background. Topography data include phi_s, SGH, SGH30, and landm_coslat. SGH
and SGH30
    area(Omega)^-1 sqrt( integral_Omega (h1(x) - h2(x))^2 dx ),
where h1 and h2 are two topography references.

SGH captures the smoothing between an ~5-km length scale and the actual
grid. SGH30 captures the smoothing between the ~5-km length scale and the true
(30") topography.

1. We want phi_s on the dynamics grid and the FV grid to be consistent in the
sense that an integral over an FV subcell of either representation gives the
same value.

2. We also what topography to be as rough as the simulation can tolerate.

1 is currently true because we read FV PHIS, remap to GLL, then remap back to FV
to make 1 true. But this procedure does not satisfy 2 because it introduces new
smoothing.

II. New topo file format, the mixed GLL-FV format. Thus, a new topography file
format will have the fields
   PHIS_d, phi_s on the GLL grid
   PHIS, phis_s on the FV grid
   SGH, SGH30, LANDM_COSLAT, LANDFRAC on the FV grid
The old file format, which does not contain PHIS_d, is still supported.

III. Final tool chain. PHIS_d will have exactly the smoothness the dycore
requires. Then PHIS_d is remapped to PHIS. The other fields can be computed from
PHIS. This requires cube_to_target and other tools.

IV. Converter. This PR provides a tool in standalone Homme to convert an old GLL
file to a new mixed GLL-FV file. The remapped SGH is augmented with additional
variance due to the remap. The results are good, but the final tool chain should
be used for new grid development work.

components/homme/src/tool/tool_main.F90
whannah1: oh ok, since it really is a general offline tool then a general name makes sense. I was just worried it was more specific than that. so "tool_main" is fine.

components/cam/src/dynamics/se/inidat.F90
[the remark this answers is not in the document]
whannah1: ok, that works. The conflicts shouldn't be too hard to fix since the block isn't that big.

components/homme/CMakeLists.txt
```diff
@@ -101,6 +101,7 @@ OPTION(BUILD_HOMME_PREQX_ACC "Primitive equations FEM with OpenACC" ON)
 OPTION(BUILD_HOMME_PREQX_KOKKOS "Primitive equations FEM with Kokkos" OFF)
 OPTION(BUILD_HOMME_SWIM  "Shallow water equations implicit" OFF)
 OPTION(BUILD_HOMME_PRIM  "Primitive equations implicit" OFF)
+OPTION(BUILD_HOMME_TOOL  "Offline tool" OFF)
```
mt5555: any reason not to make this default ON?

mt5555:
nice PR.  I'm looking forward to moving the interpolate, template and toposmoothing tools into the new tool framework.

components/cam/src/dynamics/se/inidat.F90
rljacob: Does this mean the dual grid is no longer needed?

components/homme/CMakeLists.txt
ambrad: No reason. Will do.

components/cam/src/dynamics/se/inidat.F90
ambrad: I believe adoption of (1) physgrid and (2) tri-grid will make the GLL dual grid unnecessary in v2. This PR does not change the status of the dual grid; it was already unnecessary for physgrid topo. This PR focuses on maximizing the quality of topo on each grid while keeping the two topo representations consistent, which does not require the GLL dual grid.

SCREAM may not use tri-grid, in which case the GLL dual grid will still be needed for that project. But I believe the status of that point is uncertain.

ambrad:
This PR should go in only after #3204 and #3251. Once those go in, I'll merge master and resolve the expected conflicts in inidat.F90.

mt5555:
@whannah1 : what's the status of #3204 - i seem to remember it was delayed because of some CIME issue.   should we really delay this PR until afer #3204?

components/homme/src/tool/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
+# macro for building the tools executable
+
+macro(tool_setup)
+
+SET(TARGET_DIR        ${HOMME_SOURCE_DIR}/src/tool)
+SET(UTILS_SHARE_DIR   ${HOMME_SOURCE_DIR}/utils/csm_share)
+SET(SRC_DIR           ${HOMME_SOURCE_DIR}/src)
+SET(SRC_SHARE_DIR     ${HOMME_SOURCE_DIR}/src/share)
+SET(TEST_SRC_DIR      ${HOMME_SOURCE_DIR}/src/test_src)
```
oksanaguba: Probably tools won't need test_src?

components/homme/src/tool/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+SET(EXEC_SOURCES ${TOOL_SRCS_F90})
+
+# Set up defaults
+IF (NOT PREQX_NP)
+  SET (PREQX_NP 4) 
+ENDIF ()
+IF (NOT PREQX_PLEV)
+  SET (PREQX_PLEV 20) 
+ENDIF ()
+IF (NOT PREQX_USE_PIO)
```
oksanaguba: I would assume most of tools would need PIO=TRUE here (native input/output)...

components/homme/src/tool/CMakeLists.txt
ambrad: Test modules get used in various spots in the code, so it's needed to build.

components/homme/src/tool/CMakeLists.txt
ambrad: Good point.

components/homme/src/tool/CMakeLists.txt
oksanaguba: this is very minor: I am only asking because there is a code in tests_src that needs commenting if one wants to run intel with checks, for example. I thought tools won't need tests, so, it would make debugging easier. or maybe build tools with CAM=true directive.

ambrad: You might open an issue about "code in tests_src that needs commenting if one wants to run intel with checks," as that implies something is broken.

What I meant above is that there is core code that uses `test_mod`, which in turn uses modules in `test_src`. So new cpp code would be needed to isolate `test_src` sufficiently. I consider that orthogonal to this PR. I believe setting CAM to true would create unresolved `use`s of modules from CAM.

ambrad:
I'm running e3sm_developer again b/c I had not sync'ed my repo's master and my branch after the recent IMEX PR. In my e3sm_dev run that just finished, all tests except HOMME's standalone theta-l tests pass.

ambrad:
Now that PR #3204 in in master, I'll merge master into this branch, resolve the inidat conflict, and rerun tests.

ambrad:
All e3sm_developer tests pass.

PR #3269, opened 2019-10-30
Add gnu modules on Bebop

amametjanov:
Add gnu modules on Bebop

[BFB]


amametjanov:
Does not affect nightlies, skipping integration tests.

PR #3271, opened 2019-10-31
Fixing bug in theta NH restart

oksanaguba:
Found by Az. Should not affect tests because the bug is only triggered by gnu.

[bfb] for nightlies, triggered only by gnu.


oksanaguba:
This may also fix compy ERS tests for theta.

AaronDonahue:
If there are no objections I'm going to start integrating this PR into next.

PR #3274, opened 2019-10-31
Update clubb to print tunable parameters and make TTEND consistent

wlin7:
clubb_intr.F90 is updated to record the CLUBB tunable parameter values
in the atm.log file.

Also change is to replace STEND_CLUBB with TTEND_CLUBB. Previously
the quantity in STEND_CLUBB is inconsistent with the name and the unit
attributes. This is now corrected using  new variable TTEND_CLUBB.

[BFB]

### DIFF
--- a/components/cam/src/physics/cam/clubb_intr.F90
+++ b/components/cam/src/physics/cam/clubb_intr.F90
@@ -784,7 +784,7 @@ end subroutine clubb_init_cnst
     call addfld ('CLOUDCOVER_CLUBB', (/ 'ilev' /), 'A', 'fraction', 'Cloud Cover') 
     call addfld ('WPTHVP_CLUBB',     (/ 'lev' /),  'A',     'W/m2', 'Buoyancy Flux')
     call addfld ('RVMTEND_CLUBB',  (/ 'lev' /),  'A',    'g/kg /s', 'Water vapor tendency')
-    call addfld ('STEND_CLUBB',      (/ 'lev' /),  'A',      'k/s', 'Temperature tendency')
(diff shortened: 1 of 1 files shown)


rljacob:
@wlin7 please get this merged.

AaronDonahue:
@singhbalwinder , you are still listed as a reviewer for this PR.  Do you approve?

AaronDonahue:
Is it fair to make a note somewhere that technically this is BFB, but given the change to the output file it can trigger a cprnc test as non-BFB?

rljacob:
No log files aren't tested so it won't trigger a non-BFB.

AaronDonahue:
doesn't this PR also get rid of the `STEND` output field and replace it with the `TTEND`?

oksanaguba:
@AaronDonahue is stend static energy tendency? I gripped and there are some stend in physics. what do you mean by getting rid of it? thanks!

AaronDonahue:
It looks to me like this PR will replace the output field for the history file that had the name `STEND` with a `TTEND` output field which will be the temperature.  I.e, taking `stend/cpair`.  Unless I'm not reading the changes correctly.

rljacob:
Oh I see its also changing the history file output.   Since the old variable goes away, I don't think that triggers a BFB diff.  Also won't do anything if its not in the standard output.

AaronDonahue:
sorry, correction.  I mean `STEND_CLUBB` and the new `TTEND_CLUBB`

AaronDonahue:
Ahhh yes, you are right.  Looking further up in the code it isn't actually a default out unless `history_clubb` is true, which it isn't by default.  So no worries.

PR #3277, opened 2019-11-01
Convert the format of the following two files from dos to unix: Veget…

xyk-ornl:
This commit converts the format of two files in ELM from dos to unix.
VegetationDataType.F90 and VegetationType.F90


jqyin:
Merged #3277

PR #3278, opened 2019-11-01
fix bugs that cause two land output variables to be NaN

xyk-ornl:
Two diagnostic variables SEN_NLOSS_LITTER and SEN_PLOSS_LITTER are NaN from current ELM runs due to bugs in the calculation. This commit fixes those bugs.


jqyin:
Merged #3278

PR #3279, opened 2019-11-01
fix output bugs in maint-1.1

xyk-ornl:
Two diagnostic variables SEN_NLOSS_LITTER and SEN_PLOSS_LITTER are NaN from historical CBGC runs due to bugs in the calculation. This commit fixes those bugs in the maint-1.1.


xyk-ornl:
@jonbob Can you please help merge this PR into maint-1.1? We need to have this for the future CBGC simulations. Thanks

jonbob:
of course! Once it's reviewed I'll merge right away

jonbob:
Is there a reported bug this is fixing? If not, we should remove the "bug fix" label...

acme-y9s:
No, there is not a reported bug.

jonbob:
merged to maint-1.1

PR #3280, opened 2019-11-01
Correct the marine BGC settings for future BGC compsets

jonbob:
This PR corrects the marine BGC settings for future BGC compsets. For maint-1.1 only

[BFB] for all tested compsets


jonbob:
merged to maint-1.1

PR #3281, opened 2019-11-01
Fixing energy computations in EAM

oksanaguba:
Implementing fixes according to 

https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/2015MS000448 :
Energy considerations in the Community Atmosphere Model (CAM)
David L. Williamson  Jerry G. Olson  Cecile Hannay  Thomas Toniazzo  Mark Taylor  Valery Yudin

1)Removed geopotential_dse call in physics update, the call was updating T using BAB's formula with geopotential. Instead, update temperature as dT = dS / c_p . Compute zi, zm, and s with new T after that.
2)Remove update of T from tend%dtdt right before dycore call. This allows latest state%t updated by physics_update to be used in the dycore.
3)Change definition of total energy in check_energy.F90 .
4)Change CLUBB's definition of total energy to remove geopotential term.

Climo and TE statistics are on https://acme-climate.atlassian.net/wiki/spaces/COM/pages/1056112812/Energy+fixer+heat+glob+TE+6-year+run+climo+AMB+diagn

[NonBFB] for atmosphere because state%s and state%t are computed differently.


oksanaguba:
I should compute zm, zi in temperature_from_dse .

kaizhangpnl:
Is there a related github issue or confluence page available? It's better to have a description about why this PR is needed. Thanks.

oksanaguba:
@kaizhangpnl there are a couple of issues in the code as described here https://agupubs.onlinelibrary.wiley.com/doi/10.1002/2015MS000448: 1) temperature adjustment from physics includes geopotential while energy formula for the model in pressure coordinated does not need geopotential 2) tendency vs state calculations for temperature need to be cleaned. 

These issues need to be fixed. This is a work-in-progress pr, I will do a better description when it is ready.

kaizhangpnl:
I see. Thanks for your reply. Could you tag me when it is ready for integration?

oksanaguba:
There are a few changes that are left for future bfb clean up PR. Keeping changes in this PR minimal.

oksanaguba:
Not sure why some changes were causing a crash, the changes were removed in https://github.com/E3SM-Project/E3SM/pull/3281/commits/cd83c905b00f4326a01acfc8828384f521e15d3f . @mt5555 , would you know? thnaks!

oksanaguba:
Commit ff31512 and commit cd83c90 are BFB for 1 month ne30 run. More details are on https://acme-climate.atlassian.net/wiki/spaces/COM/pages/1056112812/Energy+fixer+heat+glob+TE+6-year+run+climo+AMB+diagn . This is ready.

mt5555:
Regarding tmp_t:   The subroutine is expecting an array of leading dimension pcol, not ncol.

Assuming state%t's leading dimension is ncol:  this would probably work:   state%t(:,:pver) . (and using state%t(:ncol,:pver) is a bad idea since it will force an temporary copy).

mt5555:
As this is an answer changing (but probably not climate changing) bugfix, I think it can not go in to master without approval of the watercycle group.  

Since it also puts in the correct energy in CLUBB, @vlarson should also comment. 

We could also consider only enabling this code for the "theta" NH dycore, so that it wont change the existing watercycle simulations with the v1 dycore, but will be evaluated as part of bringing in the v2 NH dycore.

components/cam/src/physics/cam/clubb_intr.F90
```diff
@@ -2138,15 +2140,19 @@ real(r8) :: ugust  ! function: gustiness as a function of convective rainfall
       wv_a = 0._r8
       wl_a = 0._r8
       do k=1,pver
-         clubb_s(k) = cpair*((thlm(i,k)+(latvap/cpair)*rcm(i,k))/exner_clubb(i,k))+ &
-                      gravit*state1%zm(i,k)+state1%phis(i)
-         se_a(i) = se_a(i) + clubb_s(k)*state1%pdel(i,k)*invrs_gravit
+         enthalpy = cpair*((thlm(i,k)+(latvap/cpair)*rcm(i,k))/exner_clubb(i,k))
+         clubb_s(k) = enthalpy + gravit*state1%zm(i,k)+state1%phis(i)
+!         se_a(i) = se_a(i) + clubb_s(k)*state1%pdel(i,k)*invrs_gravit
+         se_a(i) = se_a(i) + enthalpy * state1%pdel(i,k)*invrs_gravit
```
mt5555: is it correct to include the latent heat (vapor) in the enthalpy?
it also may be confusing to use "se_a", "se_b", which suggests static energy, but here in this code it is no longer static energy?

oksanaguba: In the same file
thlm(i,k)    = state1%t(i,k)*exner_clubb(i,k)-(latvap/cpair)*state1%q(i,k,ixcldliq)
and
rcm is cloud water mixing ratio (which i think what q(,,ixcldliq) is)
and so
enthalpy=cpair*((thlm(i,k)+(latvap/cpair)*rcm(i,k))/exner_clubb(i,k))
=cpair*state1%t ?
For se_a and se_b, i was trying to keep number of changes minimal.

rljacob:

oksanaguba:
Not that I know of.

components/cam/src/dynamics/se/dp_coupling.F90
```diff
@@ -587,22 +587,12 @@ subroutine derived_phys(phys_state, phys_tend, pbuf2d)
...
-             ! hydrostatic case: integrate zi term by parts, use CP=CV+R to get:
-             ! E = CP_air T + phis   (Holton Section 8.3)
-             ! to use this, update geopotential.F90, and other not-yet-found physics routines:
-             ! (check boundary layer code, others which have gravit and zi() or zm()
-             phys_state(lchnk)%s(i,k) = cpair*phys_state(lchnk)%t(i,k)    &
-                                      + phys_state(lchnk)%phis(i)
-#else
              phys_state(lchnk)%s(i,k) = cpair*phys_state(lchnk)%t(i,k)    &
                                       + gravit*phys_state(lchnk)%zm(i,k)  &
                                       + phys_state(lchnk)%phis(i)
```
oksanaguba: I probably should have removed phis term from here.

oksanaguba:
in next

PR #3291, opened 2019-11-06
Makes short queue on Compy the default for small jobs

singhbalwinder:
Short queue is now default with a nodemax count of 40. Jobs over 40
nodes will go to slurm queue. This is done so that most tests can run 
in the short queue.

[BFB] - Bit-For-Bit

### DIFF
--- a/cime/config/e3sm/machines/config_batch.xml
+++ b/cime/config/e3sm/machines/config_batch.xml
@@ -398,7 +398,8 @@
 
    <batch_system MACH="compy" type="slurm">
     <queues>
-      <queue walltimemax="00:59:00" default="true">slurm</queue>
+      <queue walltimemax="00:30:00" nodemax="40" default="true">short</queue>
+      <queue walltimemax="01:00:00" >slurm</queue>
     </queues>
    </batch_system>
 
--- a/cime/config/e3sm/machines/config_batch.xml
+++ b/cime/config/e3sm/machines/config_batch.xml
@@ -398,8 +398,8 @@
 
    <batch_system MACH="compy" type="slurm">
     <queues>
-      <queue walltimemax="02:00:00" nodemax="40" default="true">short</queue>
-      <queue walltimemax="06:00:00">slurm</queue>
+      <queue walltimemax="02:00:00" strict="true" nodemin="1" nodemax="40" default="true">short</queue>
+      <queue walltimemax="06:00:00" nodemin="1" nodemax="460">slurm</queue>
     </queues>
    </batch_system>
 
--- a/cime/config/e3sm/machines/config_batch.xml
+++ b/cime/config/e3sm/machines/config_batch.xml
@@ -398,8 +398,8 @@
 
    <batch_system MACH="compy" type="slurm">
     <queues>
-      <queue walltimemax="00:30:00" nodemax="40" default="true">short</queue>
-      <queue walltimemax="01:00:00" >slurm</queue>
+      <queue walltimemax="02:00:00" nodemax="40" default="true">short</queue>
+      <queue walltimemax="06:00:00">slurm</queue>
     </queues>
    </batch_system>
 


cime/config/e3sm/machines/config_batch.xml
```diff
@@ -398,7 +398,8 @@
 
    <batch_system MACH="compy" type="slurm">
     <queues>
-      <queue walltimemax="00:59:00" default="true">slurm</queue>
+      <queue walltimemax="00:30:00" nodemax="40" default="true">short</queue>
```
jgfouca: 30 minutes is quite a small walltimemax. How was this number determined?

singhbalwinder: This is the default time the queue should be using, the max is 2hrs. Is there a way to mention the default time? I just saw walltimemax and walltimemin options there.

jgfouca: @singhbalwinder , CIME is good at determining a reasonable walltime selection for jobs. This field is for MAX walltime and should be treated as such.

singhbalwinder: Thanks. I will change it to 2hrs.

singhbalwinder: I have modified walltimemax values for both queues

rljacob:
@singhbalwinder go ahead on put this on next to see if it help with testing.  But I think we need to work on the queue definitions some more.

singhbalwinder:
Merged to next

rljacob:
2 tests asked for more than 2 hours but were still submitted to the short queue.  @jgfouca do you know why?
```
e3smtest e3sm       40     short 44719   PD       0:00  2:08:00 PartitionTimeLimit test.SMS_D_Ld1.ne30_r05_oECv3.A_WCYCL1850.compy_pgi.JNe
e3smtest e3sm       40     short 44721   PD       0:00  3:00:00 PartitionTimeLimit test.SMS_D_Ld1.ne30_oECv3_ICG.A_WCYCL1850S_CMIP6.compy_
```

cime/config/e3sm/machines/config_batch.xml
```diff
@@ -398,7 +398,8 @@
 
    <batch_system MACH="compy" type="slurm">
     <queues>
-      <queue walltimemax="00:59:00" default="true">slurm</queue>
+      <queue walltimemax="02:00:00" nodemax="40" default="true">short</queue>
```
jgfouca: One more comment, I believe you want to add
```
strict="true"
```
to the short queue definition.

singhbalwinder: Ok, I will do that. Just for my understanding, what does this "strict" keyword do? From its name, it sounds like it will enforce everything strictly. I am not sure how the "non-strict" version decides what to allow and what not to allow. May be I am completely misinterpreting this variable.

rljacob:
I just had a talk with Jim about this.  "strict" should really be called "walltimemax_strict".  It makes the walltimemax a hard limit because it by default is not.   So we need to add that to make sure anything requested more then 2 hours doesn't get the short queue in its case.run.

Now that I understand this better, we could use the queue settings to enforce our queue policies at least at the cime level.  Let me make a change to this branch.

singhbalwinder:
Sounds good. Thanks @rljacob .

rljacob:
@jgfouca I'm trying to add more queue options and checking a few cases and don't understand what is happening.

I have config_batch edited to be this for compy:
```
      <queue walltimemin="00:30:00" walltimemax="02:00:00" nodemax="40" default="true" strict="true">short</queue>
      <queue walltimemin="01:30:00" walltimemax="36:00:00" nodemin="1" nodemax="100"  strict="true">slurm</queue>
      <queue walltimemin="01:30:00" walltimemax="24:00:00" nodemin="101" nodemax="180"  strict="true">slurm</queue>
      <queue walltimemin="01:30:00" walltimemax="06:00:00" nodemin="181" nodemax="460"  strict="true">slurm</queue>
```
I try to make a high-res case without specifying  a walltime.  It will take 310 nodes.
./create_newcase --case HRtry --res ne120_oRRS18v3_ICG --compset A_WCYCL2000_H01AS

And I get this message

`WARNING: No queue on this system met the requirements for this job. Falling back to defaults`
And it winds up being in the short queue for 2 hours.  I was hoping it would pick the "slrum" queue for 6 hours given the above.  What is wrong?

jgfouca:
@rljacob you may need to put a print statement in env_batch.py:select_best_queue

rljacob:
I added a couple of print statements and getting more confused.  In the routine you recommended I have:
```
    def select_best_queue(self, num_nodes, num_tasks, walltime=None, job=None):
        # Make sure to check default queue first.
        qnames = self._get_all_queue_names()
        for qname in qnames:
            print( "queue check            ", qname)
            if self.queue_meets_spec(qname, num_nodes, num_tasks, walltime=walltime, job=job):
```

I then went in to queue_meets_spec and printed out some of the specs from the queue.
```
        nodemin, nodemax, jobname, walltimemax, jobmin, jobmax, strict = specs
        print( "  queue meet            ", num_nodes,nodemin,nodemax,walltimemax)
```

And I'm getting this:
```
job is case.run USER_REQUESTED_WALLTIME None USER_REQUESTED_QUEUE None
('queue check            ', 'short')
('  queue meet            ', 310, None, 40, '02:00:00')
('queue check            ', 'short')
('  queue meet            ', 310, None, 40, '02:00:00')
('queue check            ', 'slurm')
('  queue meet            ', 310, None, 100, '36:00:00')
('queue check            ', 'slurm')
('  queue meet            ', 310, None, 100, '36:00:00')
('queue check            ', 'slurm')
('  queue meet            ', 310, None, 100, '36:00:00')
WARNING: No queue on this system met the requirements for this job. Falling back to defaults
```

Lots of weirdness:  there are 4 queue entries but the loop is done 5 times.  It seem to only be reading the first 2 lines based on nodemax and maxwalltime output.  Nodemin is "none" is all lines but should be "1" in the second queue entry.

jgfouca:
@rljacob , yikes. Let me login and look.

rljacob:
Thanks.  To reproduce, modify the compy config_batch as in https://github.com/E3SM-Project/E3SM/pull/3291#issuecomment-551235983  and then try the create_newcase command in that same comment.

jgfouca:
Yeah, i'm just going to checkout this branch.

jgfouca:
Oh, you have additional customized edits? I'll grab those too.

jgfouca:
@rljacob , it looks like the code in env_batch.py is not set up to handle multiple queues with the same name.

jgfouca:
It makes sense that this is not really a safe thing to do. If we have multiple entries with the same name, how would the user pick one from the CLI?

rljacob:
Ok I guess so but how did the entries for theta ever work?  tagging @amametjanov

amametjanov:
I think it's working on Theta. Queue policies are here: https://www.alcf.anl.gov/user-guides/job-scheduling-policy-xc40-systems

rljacob:
@amametjanov I believe you have edited config_batch to match Theta's policies but have you confirmed that CIME puts a job with a certain node count in the right queue in its case.run ?

rljacob:
wait the queue name is always the same. Does it always set the walltime correctly based on the nodecount?

rljacob:
@jgfouca I reduced the number of queue entries for compy to 2 (short and slurm) and it still does the "for qname in qnames:" loop 3 times.

jgfouca:
@rljacob , that is expected since a copy of the default queue is prepended to the search list.

rljacob:
Just added the strict option and updates next.   Hopefully the testing will finish by itself tonight.

jgfouca:
@rljacob , can I merge this?

rljacob:
Sure go ahead.

rljacob:
For those following along, we continued to have problems with testing finishing overnight but it was a a slow test archiving issue.

PR #3294, opened 2019-11-07
Correct ocn_co2_type setting for bcrd case

### LINKED ISSUE #3289: incorrect ocn_co2_type setting for bcrd case
in the file components/mpas-ocean/bld/build-namelist, there is an incorrect setting in the bcrd section:

} elsif ($ocn_co2_type eq 'bcrd' ) {
       add_default($nl, 'config_ecosys_atm_co2_option', 'val'=>"constant");
       add_default($nl, 'config_ecosys_atm_alt_co2_option', 'val'=>"constant");
       add_default($nl, 'config_ecosys_atm_alt_co2_use_eco', 'val'=>".false.");
       add_default($nl, 'config_ecosys_atm_co2_constant_value', 'val'=>$atm_co2_const_val);

the default config_ecosys_atm_alt_co2_option should be "diagnostic" instead of "constant".

jonbob:
This PR brings in a bug fix to correct the ocn_co2_type setting for bcrd cases in the maint-1.1 branch

Fixes #3289 

[BFB]

### DIFF
--- a/components/mpas-ocean/bld/build-namelist
+++ b/components/mpas-ocean/bld/build-namelist
@@ -960,7 +960,7 @@ if ($ocn_co2_type eq 'constant' ) {
         add_default($nl, 'config_ecosys_atm_co2_constant_value', 'val'=>$atm_co2_const_val);
 } elsif ($ocn_co2_type eq 'bcrd' ) {
         add_default($nl, 'config_ecosys_atm_co2_option', 'val'=>"constant");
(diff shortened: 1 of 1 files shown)


jonbob:
merged to maint-1.1

rljacob:
@jonbob @kevans32 does this bug fix change answers for the BCRD cases?

jonbob:
@rljacob - it does change answers for tests run out-of-the-box. My understanding is that this change was implemented in the run_e3sm script but also committed to the repo to make sure it didn't get lost.

PR #3298, opened 2019-11-07
update anvil queues for time limits depending on size

sarich:
acme-large Jobs with > 60 nodes have a max time of 12 hours
acme-medium Jobs with 6-60 nodes have a max time of 24 hours
acme-small (default) Jobs with 1-5 nodes have a max time of 48 hours


rljacob:
I think we should set the machine name back to "anvil" and get rid of any "anvil-centos7" or "anvil-centos6" machine entries.

Issue #3299, opened 2019-11-07
output of omega in cam


oksanaguba:
https://github.com/E3SM-Project/E3SM/blob/c59c811106329786577ce1a5e05349ec4271f2d0/components/cam/src/dynamics/se/stepon.F90#L469

here omega is is multiplied by pressure, but in theta dycore it does not need to, because %derived%omega_p is just omega. should this be fixed?

whannah1:
@oksanaguba , good point. So in the theta dycor the "omega_p" variable is still used, but the data put into it is not "omega/p" but rather just "omega"?

oksanaguba:
yes. that is why in dp_coupling it is not multiplied either:
```
!for theta there is no need to multiply omega_p by p
#ifndef MODEL_THETA_L
    !$omp parallel do private (lchnk, ncols, ilyr, icol)
    do lchnk = begchunk,endchunk
      ncols = get_ncols_p(lchnk)
      do ilyr = 1,pver
        do icol = 1,ncols
          if (.not.single_column) then
            phys_state(lchnk)%omega(icol,ilyr) = phys_state(lchnk)%omega(icol,ilyr) &
                                                *phys_state(lchnk)%pmid(icol,ilyr)
          end if
        end do ! icol
      end do ! ilyr
    end do ! lchnk
#endif
```

whannah1:
ok, I can throw an ifdef into that part of stepon.F90 and make a PR.

oksanaguba:
no rush, you can stick it into some other pr (i assume omega is not in output so all would be bfb?)

whannah1:
yea, it's not standard output so it shouldn't make any tests fail. I'm not planning any PRs for E3SM right now so I might as well do a tiny PR.

PR #3302, opened 2019-11-07
Make DYN_OMEGA consistent with theta dycore

whannah1:
For the theta dycore the "omega_p" variable is just omega, 
rather than omega/p, so this makes sure the DYN_OMEGA output accounts for this difference.

Fixes issue #3299 .

[BFB]


oksanaguba:
e3sm_atm_developer passes got this PR merged into today's next against the same next.

oksanaguba:
merged to next

PR #3308, opened 2019-11-12
Update mpas-source for python 3

mark-petersen:
Update BGC and cvmix submodules within MPAS-Ocean to adhere to python 3.


xylar:
One advantage of this merge is that any future changes in OceanBGC or CVMix on their respective `maint-1.0` branches won't require changes outside of those repos, since we're pointing to the branches rather than specific commit hashes.  Hopefully, no one is worried about the potential hazards that also implies.

jonbob:
@xylar - I think that's very dangerous then. If so, answers could become non-BFB without our realizing it. Is there a way to get a hash instead?

rljacob:
@xylar go ahead and tag those CVMIX and OceanBGC maintenance branches and use the tag in the build script.   We will almost never need to update those tags.

vanroekel:
@xylar  and @mark-petersen can I ask why we are pointing to an E3SM fork of CVMix and not simply updating the tag to CVMix in get_cvmix.sh?  Also, what do you mean by changes to CVMix/OceanBGC on their maint-1.0 branches?  It looks like you are pointing to a tag of a fork, is this right?  So if I had a bug fix for CVMix I'd have to change it in CVMix and then the E3SM for and then update the tag here too?  This sounds overly complicated and confusing to me.  I would suggest we simply update the CVMix and BGC tags instead.  I am not understanding the advantage of this change and see more potential pitfalls.

rljacob:
The background is that any external repo that the maint-1.0 version of E3SM depends on will possibly needs its own corresponding maint branch to fix problems.  We don't control branches in CVMix (?) so made a fork here.

xylar:
@vanroekel, I think we don't want to bring all the other recent change besides the python 3 fix in from CVMix, that would defeat the whole purpose of having the `E3SM/maint-1.0` branch.  We just want the bug fixes that are preventing us from building CVMix.  Same with OceanBGC.  So we have `maint-1.0` branches that will be updated only with bug fixes that are approved for `maint-1.0` and presumably are bit-for-bit unless there's some reason they can't be.

xylar:
> @xylar - I think that's very dangerous then. If so, answers could become non-BFB without our realizing it. Is there a way to get a hash instead?

Okay, we can stick with a hash or tag.

> @xylar go ahead and tag those CVMIX and OceanBGC maintenance branches and use the tag in the build script. We will almost never need to update those tags.

It's hard for me to know what that tag should be, @rljacob.  It shouldn't be `maint-1.0` because that's a dynamic thing.  I think the most reasonable thing is to go back to a hash.

rljacob:
hash is even better.  I'm not sure what's allowed in the "*_TAG" variable in those scripts.

xylar:
https://github.com/MPAS-Dev/MPAS-Model/pull/395 should address all the remaining concerns besides @vanroekel's.  

@vanroekel, if you have suggestions of how to fix python 3 builds on `maint-1.0` in a way that you find more friendly or intuitive, please let us know quickly.  The lack of python 3 support is holding up our transition to a new python 3 testing environment that @jhkennedy set up before he moved to Alaska and we're hoping to deploy as the new default.

xylar:
@mark-petersen, please update the source following the merge of MPAS-Dev/MPAS-Model#395

mark-petersen:
Updated and ready to merge.

rljacob:
@jonbob go ahead and merge this.

jonbob:
merged to maint-1.0

PR #3313, opened 2019-11-14
Update mpas-source to fix pool memory leak

mark-petersen:
This update is required when using RK4 time-stepping in the ocean model, as we hit an out-of-memory error. It is only used in idealized cases, not in E3SM, so our standard simulations are not affected.

[BFB]


jonbob:
@mark-petersen - if this is a bug report, it should be an issue first and then a "bug fix" PR. We can just leave those labels off for this

mark-petersen:
It was not an E3SM bug, it only was in MPAS for RK-4, so I took the bug label off here.

mark-petersen:
Ran a few tests to be sure:
```
PASS PEM_Ln9.ne30_oECv3wLI.A_WCYCL1850-DIB-ISMF_CMIP6.cori-knl_intel COMPARE_base_modpes
PASS PET_Ln9.T62_oEC60to30v3.GMPAS-IAF.cori-knl_intel COMPARE_base_single_thread
```

jonbob:
merged to next

jonbob:
merged to master

PR #3319, opened 2019-11-15
Fix create_test bug when determining BASELINE_ROOT

jgfouca:
[BFB]


rljacob:
What kind of testing did you do?   And how was it different from the testing for https://github.com/E3SM-Project/E3SM/pull/3314?

jgfouca:
@rljacob , tested it by hand and through jenkins. I'm a bit baffled by #3314 ; I remember testing it pretty thoroughly by hand and via scripts_regression_tests, but it clearly was totally broken. I tried testing it again by hand this morning and it immediately broke. I searched up through my shell history and it appeared to be the same code I tested by hand yesterday (which appeared to work), so  *shrug*?

PR #3320, opened 2019-11-15
Homme DIRK arithmetic

ambrad:
These are the changes to Homme arithmetic that I would like to make prior to BFB-converting the code to KNL/GPU-performant C++/Kokkos.

[non-BFB] for theta-l nonhydrostatic standalone and EAM tests, [BFB] for all others


ambrad:
I have arranged this PR so that each commit shows exactly one change, and each non-BFB commit is small. (Only the move to imex_mod, which is a BFB change, is a big commit) Thus, I recommend you read through the changes of each commit rather than look at the overall code changes.

components/homme/src/theta-l/imex_mod.F90
```diff
@@ -246,15 +246,15 @@ subroutine compute_stage_value_dirk(n0,np1,alphadt,qn0,dt2,elem,hvcoord,hybrid,&
                 do k=1,nlev-1
                    dphi(i,j,k)=dphi(i,j,k) + x(k+1,i,j)-x(k,i,j)
                 enddo
-                dphi(i,j,nlev)=dphi(i,j,nlev) + (0 - x(k,i,j) )
```
oksanaguba: why did this work before? was k updated to nlev anyway?

components/homme/src/theta-l/imex_mod.F90
```diff
@@ -130,12 +130,12 @@ subroutine compute_stage_value_dirk(n0,np1,alphadt,qn0,dt2,elem,hvcoord,hybrid,&
 
     ! dirk settings
     maxiter=20
-    deltatol=1e-13  ! exit if newton increment < deltatol
-    !restol=1e-13    ! exit if residual < restol  
+    deltatol=1.0e-13_real_kind  ! exit if newton increment < deltatol
```
oksanaguba: could you elaborate here? I think code has a lot of consts defined without _real_kind. if precision changes, won't consts be converted properly?

oksanaguba:
Is nonbfb here from rearranging arithmetics or from undefined index too?

components/homme/src/theta-l/imex_mod.F90
ambrad: Yes, k was updated to nlev. I imagine that happens, in practice, with all compilers, but I don't think it's in the standard. (I might be wrong.)

components/homme/src/theta-l/imex_mod.F90
ambrad: No. 1e-13, e.g., will be viewed as single rather than double unless a flag is given.

ambrad:
Non-bfb is from rearranging arithmetic, in particular distribution of 1/dp3d_i.

AaronDonahue:
@ambrad and @oksanaguba I've just been added as the integrator for this PR.  As a final check before I begin the integration process I wanted to ask, is this PR ready to go?

oksanaguba:
It is ready. Before you push to next, would you consider running tests with next+this PR against next? If there is something in the PR that breaks a build, say (however unlikely), it is harder to fix when it is already pushed. Either I or Andrew can give you a script to do this. I use anvil.

AaronDonahue:
@oksanaguba , Absolutely.  That was my plan to begin with.  But before I set up the tests I wanted to make sure I was testing a finished PR.  I'll let you know how the tests go and when I'm merging to next.

oksanaguba:
Thanks! here is my script that I tweak each time (depending on what needs to be tested), Andrew I think has a more concise one here https://github.com/E3SM-Project/E3SM/pull/3204 
```
#!/bin/bash -e

brr=~/acme-integrator
master=~/acme-next ##### NEXT!!!
baselinefold=~/as/acme-tests

#test=e3sm_developer
test=HOMME_P24.f19_g16_rx1.A.anvil_intel

#generate baseline from master
#${master}/cime/scripts/create_test -g -o --baseline-root ${baselinefold} -b master ${test}
#run test from branch
#${brr}/cime/scripts/create_test -c --baseline-root ${baselinefold} -b master ${test}

test=ERP_Ln9.ne4_ne4.FC5AV1C-L.anvil_intel

#generate baseline from master
#${master}/cime/scripts/create_test -g -o --baseline-root ${baselinefold} -b master ${test}
#run test from branch
#${brr}/cime/scripts/create_test -c --baseline-root ${baselinefold} -b master ${test}

test=e3sm_atm_developer
suff=' --machine anvil --compiler intel '

#generate baseline from master
#${master}/cime/scripts/create_test $suff -g -o --baseline-root ${baselinefold} -b master ${test}
#run test from branch
#${brr}/cime/scripts/create_test    $suff -c --baseline-root ${baselinefold} -b master ${test}

test=e3sm_atm_integration
suff=' --machine anvil --compiler intel '

#generate baseline from master
${master}/cime/scripts/create_test $suff -g -o --baseline-root ${baselinefold} -b master ${test}
#run test from branch
${brr}/cime/scripts/create_test    $suff -c --baseline-root ${baselinefold} -b master ${test}
```

AaronDonahue:
awesome thanks!

ambrad:
Looks like @mt5555 merged to next a few hours ago. Are we sure the integrator was intended to change for this PR?

AaronDonahue:
ahhh good catch @ambrad .  So I think this one is on its way, no worries.  @rljacob , should I reassign @mt5555 so that the github record is accurate?

rljacob:
done.

PR #3321, opened 2019-11-18
Change dtime calculation for python3

### LINKED ISSUE #3315: python 3 changed dtime to float
While running our testing under python3 for the first time, the type of "dtime"  in atm_in was changed to a float.
```
  BASE: dtime = 1800
  COMP: dtime = 1800.0
```

Note from @jhkennedy , python3 changed how division works.
image

Solution is to use `//` instead of ` /` in the calculation of dtime in buildnml.

rljacob:
Change EAM dtime calculation so it will still be an
integer in python 3.

Fixes #3315 

[BFB]


rljacob:
Merged to next.

PR #3323, opened 2019-11-20
Homme theta-l C++/Kokkos DIRK conversion, part 1

ambrad:
Bring in scream::tridiag, build wrappers for it, impl some low-level DIRK-related routines, set up unit tests and test these, add some GPU BFB utils.


components/homme/src/share/cxx/utilities/MathUtils.hpp
```diff
@@ -36,6 +36,54 @@ KOKKOS_INLINE_FUNCTION constexpr FPType max(const FPType val, FPPack... pack) {
...
+}
+
+template <typename Pack>
+KOKKOS_INLINE_FUNCTION
+Pack bfb_pow (const Pack& a, const Real& e) {
+  Pack b;
+  for (int s = 0; s < VECTOR_SIZE; ++s)
+    b[s] = std::pow(a[s], e);
+  if (OnGpu<ExecSpace>::value)
+    zero_ulp_n(b, 20);
```
bartgol: The Cuda programming guide states that the max ulp error for pow is 2. Is there a reason why we should use 20 instead of 2? It may mask some other upstream non bfb errors.

components/homme/src/share/cxx/utilities/scream_tridiag.hpp
```diff
@@ -0,0 +1,656 @@
...
+    auto* const xi = X + ios;
+    for (int j = 0; j < nrhs; ++j)
+      xim1[j] = (xim1[j] - duim1[j] * xi[j]) / dim1[j];
+  }
+}
+
+template <typename TridiagDiag>
+KOKKOS_INLINE_FUNCTION
+void bfb_thomas_factorize (TridiagDiag dl, TridiagDiag d, TridiagDiag du,
+                           typename std::enable_if<TridiagDiag::rank == 1>::type* = 0) {
```
bartgol: Up for debate: I dislike dummy enable_if arguments, aimed at triggering sfinae. I prefer to see them in the template argument, or, better, to call an "impl" function, which takes extra template arguments, which is then specialized on those.

The reason is mostly readability. Extra args need to be parsed before realizing they are just there for sfinae reasons, especially for relatively new developers.

If you don't mind the change from arg to template arg, I can do the mod myself.

components/homme/test_execs/thetal_kokkos_ut/dirk_ut.cpp
```diff
@@ -0,0 +1,415 @@
...
+        const auto tmp = p[k];
+        p[k] = p[k-1] + 0.5*(delta + p[k]);
+        delta = tmp;
+      }
+    }
+  deep_copy(a, am);
+}
+
+template <typename V>
+void require_all_equal (const int nlev, const V& a, const V& b,
```
bartgol: This could be migrated to `share/cxx/utilities/TestUtils.hpp`, and make all our test cpp files a bit shorter. I'll make a not to myself.

components/homme/src/share/cxx/utilities/MathUtils.hpp
ambrad: Keep in mind this function is meant to apply to the single pow call. So all it's doing is making pow less accurate. That won't mask other errors, since all other operations are done at full precision and are not modified.

As for 20 base-2 digts vs 2 base-10 digits ~ 7 base-2 digits, the issue is carrying. I quote the doc for zeroulpn:
```C++
// Set the least significant n bits of the fraction to 0. Use this for
// BFB-testing between GPU and host. Table 8 of Appendix E of the Cuda
// programming guide gives the accuracy of various functions in base
// 10. However, we also have to account for carries in the calculations. Thus,
// even if an answer is accurate to 2 base-10 UL or ~7 bits, carries can make
// more bits differ. Thus, zeroing (or any sort of rounding) is not assured to
// work. However, you can reduce the probability of a diff to a small number by
// choosing n to be large. Full double precision range is 53 bits; choose
// something like n = 20 to get a carry-based diff every 1 in a million calls,
// still leaving 33 bits.
```

components/homme/src/share/cxx/utilities/scream_tridiag.hpp
ambrad: This is in scream_tridiag.hpp, which is copied in from scream. We can debate this there, but in the Hxx effort, we can't make changes to this file.

components/homme/test_execs/thetal_kokkos_ut/dirk_ut.cpp
ambrad: Until I'm done with dirk_ut, I prefer you not make modifications to it.

components/homme/src/share/cxx/utilities/MathUtils.hpp
bartgol: Ah, I see. I thought CUDA's API was in bits, since gamma has a max error of 11ulps, which seemed "large" if given in decimal digits.

When I said "mask other ops", I meant that if the base of the pow had 10 wrong bits (due to other non bfb stuff), zeroing out 20 bits would not let you see that diff.

Anyhow, my understanding of CUDA programming guide was wrong, so nvm.

ambrad: No, I'm wrong: I think you're right about the base-2 interpretation. They write, e.g., 2 [ULP] = 2*ulp, with ulp something like `ldexp(2,-53)`. So a base-2 interpretation is correct. I'll modify the zeroulpn doc in my next PR to reflect that.

As for masking other ops through a reduced-precision pow (for example), that's a really interesting point. I have an idea to fix that. Let x, n be the arguments to zeroulpn, and let y be the result. zeroulpn will zero the last n bits of the fraction of y. Then replace those with the last n bits of x. That's an n-bit perturbation that will require x to be BFB in those least significant digits.

bartgol: Nice idea. No rush on that, we can put it in with dirk-phase-2.

oksanaguba:
@rljacob this is pr to Luca's branch, not to master. I was confused, too. if you don't mind, I will reassign back to Luca.

rljacob:
@bartgol don't post PRs from branch to branch.

bartgol:
@rljacob what's the rationale behind prohibiting branch-to-branch PRs? I know I could simply merge the two branches locally, and then push, but this way seems more transparent to me.

rljacob:
We want the PR count on this repo to only track branches waiting to be merged to master.

ambrad:
@rljacob whoops, sorry, my fault. @bartgol, in the future, we can PR into our own forks, carry on discussion there, then merge the branch to homme_theta_kokkos.

PR #3327, opened 2019-11-25
Add snicar and snow files to mpas-seaice streams file

jonbob:
This PR adds the autogeneration of two new streams, for snicar and snow data files, to the script that makes the mpas-seaice streams file. The files it points to have been uploaded to the inputdata repository, and the code that actually uses the new files was introduced earlier but is off by default.

[BFB]


jonbob:
merged to next

jonbob:
merged to master

PR #3329, opened 2019-11-27
Add performance benchmarking tests to maint-1.0

amametjanov:
Add performance benchmarking tests to maint-1.0 for scaling and historical checkpointing:
* high- and low-res F-compset without IO
* high- and low-res G-compset without IO
* high- and low-res W-compset with production IO

Also, update machine configuration for Anvil and Compy. Load-balance PE-layouts on Compy.

[BFB]


cime/config/e3sm/testmods_dirs/bench/fcomp_noio/shell_commands
```diff
@@ -0,0 +1,11 @@
...
+./xmlchange HIST_OPTION=none
+./xmlchange SAVE_TIMING=TRUE
+# on KNLs, run hyper-threaded with 64x2
+if [ `./xmlquery --value MACH` == theta ]||[ `./xmlquery --value MACH` == cori-knl ]; then
+  ./xmlchange MAX_MPITASKS_PER_NODE=64
+  ./xmlchange MAX_TASKS_PER_NODE=128
+  ./xmlchange NTHRDS=2
+else
+  ./xmlchange NTHRDS=1
+fi
```
rljacob: Does "fcomp_noio" mean this is for modifying an F-case?  Then it should be in cam/cime_config/testmods

cime/config/e3sm/testmods_dirs/bench/getBenchTimings
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
```
rljacob: I'm not sure if testmods_dirs is where we want to put these scripts.

cime/config/e3sm/testmods_dirs/bench/gmpas_noio/user_nl_mpaso
```diff
@@ -0,0 +1,2 @@
+config_am_highfrequencyoutput_enable=false
+config_am_globalstats_enable=false
```
jonbob: @amametjanov - there are other output streams you could turn off, if you want. But I think these are the two most expensive ones...

cime/config/e3sm/tests.py
```diff
@@ -120,5 +120,76 @@
...
+                 ("SMS_PS.ne120_oRRS18v3_ICG.A_WCYCL1950S_CMIP6_HR","bench-wcycl"),
+                 ("SMS_PM.ne120_oRRS18v3_ICG.A_WCYCL1950S_CMIP6_HR","bench-wcycl"),
+                 ("SMS_PL.ne120_oRRS18v3_ICG.A_WCYCL1950S_CMIP6_HR","bench-wcycl"))
+    ),
+    "e3sm_bench_lores_g" : (None, "01:00:00", (
+                 ("SMS_P320.T62_oEC60to30v3.GMPAS-IAF","bench-gmpas_noio"),
+                 ("SMS_P480.T62_oEC60to30v3.GMPAS-IAF","bench-gmpas_noio"),
+                 ("SMS_P640.T62_oEC60to30v3.GMPAS-IAF","bench-fcomp_noio"),
+                 ("SMS_P960.T62_oEC60to30v3.GMPAS-IAF","bench-fcomp_noio"),
+                 ("SMS_P1280.T62_oEC60to30v3.GMPAS-IAF","bench-fcomp_noio"))
```
jonbob: @amametjanov - I think the three lines above should be "bench-gmpas_noio"?

PR #3330, opened 2019-11-27
Fixes incorrect conflict resolution on next

bishtgautam:
A conflict on next was incorrectly resolved and this PR fixes it.
There was no error during code compilation because the incorrect code
was within an `#ifdef` block.


bishtgautam:
@jqyin This PR is for next.

rljacob:
@bishtgautam you don't have to make a PR to fix a problem on next.  Just fix it directly.

rljacob:
Is that ifdef block supposed to be active for some compset?  If its dead code, remove it.  If not, add a test for it.

PR #3287, opened 2019-11-03
NA RRM tri-grid support for F2010SC5-CMIP6 compset

tangq:
The F2010SC5-CMIP6 compset needs new input files to be able to run on the North
America (NA) RRM tri-grid

	* Add the atm initial condition (ncdata) file for the northamericax4v1 grids
	* Add the land surface data in year 2010 for r0125 grids

[NML] - Namelist Changing

PR #3331, opened 2019-11-27
Update northamericax4v1 settings

tangq:
Some of the North America RRM settings need to be updated. In particular, the old land domain file is created with the bad overlap mesh file. This PR updates it to the correct file. 
Also preset the EPS_AGRID threshold for this RRM grid.
    
[BFB] except for northamericax4v1
[NML] no change except for northamericax4v1

### DIFF
--- a/cime/config/e3sm/config_grids.xml
+++ b/cime/config/e3sm/config_grids.xml
@@ -2257,10 +2257,8 @@
       <ny>1440</ny>
       <file grid="atm" mask="oEC60to30v3">$DIN_LOC_ROOT/share/domains/domain.lnd.r0125_oEC60to30v3.190812.nc</file>
       <file grid="lnd" mask="oEC60to30v3">$DIN_LOC_ROOT/share/domains/domain.lnd.r0125_oEC60to30v3.190812.nc</file>
-      <file grid="atm" mask="oRRS15to5">$DIN_LOC_ROOT/share/domains/domain.lnd.r0125_oRRS15to5.191022.nc</file>
-      <file grid="lnd" mask="oRRS15to5">$DIN_LOC_ROOT/share/domains/domain.lnd.r0125_oRRS15to5.191022.nc</file>
-      <!--file grid="lnd" mask="gx1v6">$DIN_LOC_ROOT/share/domains/domain.lnd.r0125_gx1v6.191017.nc</file-->
-      <file grid="lnd" mask="gx1v6">$DIN_LOC_ROOT/share/domains/domain.lnd.r0125_gx1v6.191113.nc</file>
+      <file grid="lnd" mask="oRRS15to5">$DIN_LOC_ROOT/share/domains/domain.lnd.r0125_oRRS15to5.191122.nc</file>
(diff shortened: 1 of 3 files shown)


tangq:
The atmosphere IC and land surface data settings are updated in #3287

tangq:
Can this PR be merged now? @ndkeen is working on the PE-layout of this NA RRM configuration and will need this fix to run his tests. Thanks.

tangq:
Wait a second...We also need to increase EPS_AGRID from 1.e-12 to 3.e10 as the default for the northamericax4v1_r0125_northamericax4v1 grid. This PR looks like the right one to make this change. But I am not sure how to do it...Any suggestions?

tangq:
@rljacob , I tried to find where the default EPS_AGRID setting can be changed and found it can be done in cime/scripts/Tools/config_grid.xml (https://github.com/E3SM-Project/E3SM/pull/1022/commits/0656e77177514e46957ee680892630f0ffcbee89). But it seems that this commit used an old cime version. That configuration file does exist any more.

Do you know where it can be done in the current cime? Similar tolerance adjustment is also needed for the Antarctica RRM. Thanks.

rljacob:
Its in the driver namelist.  See cime/src/drivers/mct/cime_config/namelist_definition_drv.xml.

tangq:
I tried the following, but it didn't work. Is the syntax wrong?

```
--- a/cime/src/drivers/mct/cime_config/namelist_definition_drv.xml
+++ b/cime/src/drivers/mct/cime_config/namelist_definition_drv.xml
@@ -1400,6 +1400,7 @@
     </desc>
     <values>
       <value>$EPS_AGRID</value>
+      <value atm_grid="ne0np4_northamericax4v1">3.e-10</value>
     </values>
   </entry>
```

wlin7:
@tangq , that does take effect, as you can see in drv_in. True the EPS_AGRID value in env_run.xml does not change, still setting to default 1e-12, which is controlled by cime/src/drivers/mct/cime_config/config_component.xml. This can indeed be misleading in some way.

tangq:
That's right. It is really confusing. Is there a way to change the default EPS_AGRID value in env_run.xml depending on which grid it is on?

tangq:
@wlin7 , I made a commit to modify default EPS_AGRID via drv_in. Unless there is an easy way to make the same change in env_run.xml (which is less confusing), this PR seems good to go.

wlin7:
@tangq  just to double check with you that it is now intended to use domain.lnd.r0125_gx1v6.191007.nc instead of dom, ain.lnd.r0125_gx1v6.191113.nc for mask gx1v6. It was previously updated with the one with newer time stamp. I lost track why the decision is reversed. Thanks.

tangq:
@wlin7 , the "191113" file was used for testing purposes. The "191007" one is the right one to use. Thanks for doubling check!

wlin7:
Thanks for the clarification, @tangq . Merged to next. Please review if the PR message is accurate which can be updated when merging to master.

tangq:
@wlin7 , I updated the PR message. It looks good now. Thanks.

wlin7:
Thanks for the update @tangq. Merged to master.

PR #3333, opened 2019-12-03
Fixes F compsets to support ne4np4

bishtgautam:
Fixes F1850SC5-CMIP6, F2010C5-CMIP6-HR and F2010C5-CMIP6-LR.
Moves the definition of fsurdat and finidat from use_cases namelist to
land model default namelist

[BFB]

### DIFF
--- a/components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml
+++ b/components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml
@@ -140,6 +140,12 @@ ic_tod="0" sim_year="1850" glc_nec="0" use_crop=".false." >lnd/clm2/initdata_map
 ic_tod="0" sim_year="2000" glc_nec="0" use_crop=".false." >lnd/clm2/initdata_map/clmi.ICRUCLM45.ne16_oQU240.1155ba0.clm2.r.nc
 </finidat>
 
+<!-- HOMME grid ne30 resolution -->
+
+<finidat hgrid="ne30np4"   maxpft="17"  mask="gx1v6" use_cn=".false." ic_ymd="20150101" more_vertlayers=".false."
+ic_tod="0" sim_year="2015" glc_nec="0" use_crop=".false." >lnd/clm2/initdata_map/20180716.DECKv1b_A3.ne30_oEC.edison.clm2.r.2010-01-01-00000.nc
+</finidat>
+
 <!-- HOMME grid ne120 resolution -->
 
 <finidat hgrid="ne120np4"   maxpft="17"  mask="gx1v6" use_cn=".false." ic_ymd="18500101" more_vertlayers=".false."
@@ -250,6 +256,11 @@ lnd/clm2/surfdata_map/surfdata_1x1_vancouverCAN_simyr2000_c130927.nc</fsurdat>
 lnd/clm2/surfdata_map/surfdata_1x1_mexicocityMEX_simyr2000_c130927.nc</fsurdat>
 <fsurdat hgrid="1x1_urbanc_alpha"    sim_year="2000" use_crop=".false." >
 lnd/clm2/surfdata_map/surfdata_1x1_urbanc_alpha_simyr2000_c130927.nc</fsurdat>
+
+<fsurdat hgrid="ne30np4" sim_year="2015" use_crop=".false.">
+lnd/clm2/surfdata_map/surfdata_ne30np4_simyr2010_c20181025.nc</fsurdat>
+<fsurdat hgrid="ne4np4" sim_year="2015" use_crop=".false.">
+lnd/clm2/surfdata_map/surfdata_ne4np4_simyr2000_c160614.nc</fsurdat>
 
 <!-- for pre-industrial simulations - year 1850 -->
 <fsurdat hgrid="360x720cru"   sim_year="1850" use_crop=".false."                           >
--- a/components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml
+++ b/components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml
@@ -172,6 +172,10 @@ ic_tod="0" sim_year="1850" glc_nec="0" use_crop=".false." >lnd/clm2/initdata_map
 ic_tod="0" sim_year="2000" glc_nec="0" use_crop=".false." >lnd/clm2/initdata_map/clmi.ICRUCLM45.ne120_oRRS18v3.be55494.clm2.r.nc
 </finidat>
 
+<finidat hgrid="ne120np4"   maxpft="17"  mask="gx1v6" use_cn=".false." ic_ymd="20150101" more_vertlayers=".false."
+ic_tod="0" sim_year="2015" glc_nec="0" use_crop=".false." >lnd/clm2/initdata_map/cori-knl.20190214_maint-1.0.F2010-CMIP6-HR.noCNT.ARE.nudgeUV.ne120_oRRS18v3.clm2.r.2011-01-01-00000.nc
+</finidat>
+
 <!-- HOMME grid ne240 resolution -->
 <finidat hgrid="ne240np4"   maxpft="17"  mask="tx0.1v2" use_cn=".false." ic_ymd="18500101" more_vertlayers=".false."
 ic_tod="0" sim_year="1850" glc_nec="0" use_crop=".false." >lnd/clm2/initdata_map/clmi.I1850CLM45.0521-01-01.ne240np4_tx0.1v2.162cd32_simyr1850_c170910.nc
@@ -199,6 +203,14 @@ ic_tod="0" sim_year="2000" glc_nec="0" use_crop=".true." irrigate=".true." nu_co
 <finidat hgrid="ne0np4_conus_x4v1_lowcon" ic_ymd="101" sim_year="2000">lnd/clm2/initdata_map/clmi.ICRUCLM45.0021-01-01.conusx4v1.0ac6464_simyr2000_c161130.nc</finidat>
 <finidat hgrid="ne0np4_enax4v1" sim_year="2000">lnd/clm2/initdata_map/clmi.ICRUCLM45SP.2000-01-01.enax4v1_oRRS18to6_simyr2000_c170621.nc</finidat>
 <finidat hgrid="ne0np4_twpx4v1" ic_ymd="101" sim_year="2000">lnd/clm2/initdata_map/clmi.ICRUCLM45SP.2000-01-01.twpx4v1_oRRS18to6v3_simyr2000_c170712.nc</finidat>
+
+<!-- for present day simulations - year 1950 -->
+<fsurdat hgrid="ne120np4"     sim_year="1950" >
+lnd/clm2/surfdata_map/surfdata_ne120np4_simyr1950_c20180108.nc</fsurdat>
+
+<!-- for present day simulations - year 2015 -->
+<fsurdat hgrid="ne120np4"     sim_year="2015" >
+lnd/clm2/surfdata_map/surfdata_ne120np4_simyr2010_c20181023.nc</fsurdat>
 
 <!-- for present day simulations - year 2000 -->
 <fsurdat hgrid="360x720cru"   sim_year="2000" use_crop=".false." >
@@ -282,7 +294,7 @@ lnd/clm2/surfdata_map/surfdata_1x1_brazil_simyr1850_c140610.nc</fsurdat>
 
 
 <fsurdat hgrid="ne30np4"      sim_year="1850" use_crop=".false." >
-lnd/clm2/surfdata_map/surfdata_ne30np4_simyr1850_c180306.nc</fsurdat>
+lnd/clm2/surfdata_map/surfdata_ne30np4_simyr1850_2015_c171018.nc</fsurdat>
 <fsurdat hgrid="ne16np4"      sim_year="1850" use_crop=".false." >
 lnd/clm2/surfdata_map/surfdata_ne16np4_simyr1850_c160108.nc</fsurdat>
 <fsurdat hgrid="ne120np4"     sim_year="1850" >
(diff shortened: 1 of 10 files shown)


tangq:
Have you tested whether the F2010SC5-CMIP6 compset picks up the same land files before and after this change?

bishtgautam:
In E3SM-v1.0, we made the mistake when defining new compsets using xml files under `components/clm/bld/namelist_files/use_cases` by including grid specific values (i.e. fsurdat, finidat). All grid specific entries should go in `components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml`. 

Due to this mistake, as noted on [this confluence page](https://acme-climate.atlassian.net/wiki/spaces/ED/pages/961250902/F2010C5-CMIP6-LR), `./create_test SMS_D.ne4_ne4.F2010C5-CMIP6-LR` has finidat and fsurdat that corresponds to `ne30np4` even though the `--res` of the test is `ne4np4`.

I have verified that this PR now corrects the finidat and fsurdat for `./create_test SMS_D.ne4_ne4.F2010C5-CMIP6-LR`. The test ran successfully after changing `./xmlchange PIO_TYPENAME=netcdf`, which is also noted on [the confluence page](https://acme-climate.atlassian.net/wiki/spaces/ED/pages/961250902/F2010C5-CMIP6-LR).

@tangq, I can fix all other incorrectly defined compsets for maint-1.0 in this PR instead of just fixing F2010C5-CMIP6-LR. Would you prefer that? What are the various `--res` values that you would want to make sure aren't broken by this PR.

A separate PR is needed to fix master.

(cc: @cameronsmith1)

cameronsmith1:
The hgrid selection flags can also be done in the use-case file.  Is there a reason for transferring it to the defaults file?   It seems there is a risk of changing previous compsets (some of which may be very old, and which we may not care about).  

I do acknowledge that the way I/we have been creating specific compsets to avoid the possibility of altering earlier compsets also causes reduction in flexibility for new versions of the model. I still think there should be an effort to figure out a better policy for managing compsets for the long term.

tangq:
@bishtgautam , I have 2 meetings this afternoon. I will get back to you later today or tomorrow.

tangq:
@bishtgautam, on maint-1.0, the affected compsets are F1850SC5-CMIP6, F2010C5-CMIP6-LR, and F2010C5-CMIP6-HR. These compsets were created based on the existing CMIP6 compsets at that time, so the land files are set in the use_case files. The F1850SC5-CMIP6 and F2010C5-CMIP6-LR compsets were created for the ne30np4 resolution, while F2010C5-CMIP6-HR for the ne120np4 resolution.

On master, it is the F2010SC5-CMIP6 compset on the ne30np4 grid.

It would be great if you can fix all the affected compsets on maint-1.0 on this PR and a separate PR for the master. I am fine with either way where the land settings are set. After your changes, just make sure the files are picked up as they are used to be. Thanks.

cameronsmith1:
@bishtgautam : will this affect the B compsets too?

When you put this on master,  there will also be the BGC compsets to fix, although they may not have been merged from maint-1.1 to master yet.

bishtgautam:
@cameronsmith1 Which B compsets + res should I test to make sure these changes don't adversely impact them?

It would be easier to fix master directly than trying to merge maint-1.0 or maint-1.1 to master. Also, I'm wasn't planning on fixing maint-1.1, unless someone has a need to fix it.

bishtgautam:
@tangq This PR now fixes F1850SC5-CMIP6, F2010C5-CMIP6-LR, and F2010C5-CMIP6-HR compsets.

components/clm/bld/namelist_files/use_cases/1950_CMIP6HR_control.xml
```diff
@@ -18,7 +18,6 @@
 <stream_year_last_popdens  phys="clm4_5" use_cn=".true." ndepsrc="stream" >1955</stream_year_last_popdens>
 
 <!-- CMIP6 HighResMIP compsets -->
-<fsurdat hgrid="ne120np4" >lnd/clm2/surfdata_map/surfdata_ne120np4_simyr1950_c20180108.nc </fsurdat>
```
tangq: @bishtgautam , why do some compsets still have finidat defined in the use_case xml file? I thought your plan was to move them all to the defaults xml file.

bishtgautam: I only made changes for fsurdat + finidat for F1850SC5-CMIP6, F2010C5-CMIP6-LR, and F2010C5-CMIP6-HR. While change for fsurdat is straightforward, the finidat is a little complicated. If you provide me with res+compset for all CMIP6 compsets that are supported in maint-1.0, I will make relevant changes.

cameronsmith1:
> @cameronsmith1 Which B compsets + res should I test to make sure these changes don't adversely impact them?

I would say: all of the compsets in ```cime/config/e3sm/allactive/config_compsets.xml```  for all of the different resolutions that e3sm has ever used.     This will be a huge set.  Most combinations have never been used in practice, but I see no easy way to know which have been used and which haven't.   Although, there are certainly some that are particularly critical, most notably the CMIP6 configurations (including the hi-resMIP and BGC configurations), IMHO.

The root of the problem is a conflict between two obvious goals:
A) Don't change existing compsets, so they can be rerun with newer code,
B) Allow configurations to run with the latest version of datasets, and allow users to run with new resolution+compset combinations.

So far with E3SM, (A) has been the greater priority.

How about a different strategy to achieve your aims?:
1) Add 'hgrid=xx' to the fsurdat and finidat lines in the CLM use_case files (it is already there for some of them).
2) Add the ne4np4 defaults to the CLM defaults.xml file.

I think this should preserve existing configurations because #1 should over-ride the defaults file for those resolutions (correct?), and the ne4np4 resolutions will default as you specify them (there is a risk that any ne4np4 test cases will change, but that will be easy to check by running the test cases).

The other strategy I see is to add the ne4np4 defaults to the relevant CLM use_case files, and add hgrid=xx as needed. This seems slightly lower risk, but the previous option is better for ne4np4 in the future (IMHO), so I would go with the other strategy.

> It would be easier to fix master directly than trying to merge maint-1.0 or maint-1.1 to master. Also, I'm wasn't planning on fixing maint-1.1, unless someone has a need to fix it.

At some point maint-1.1 will need to be merged to master and all the compsets should work.  Hence, if you put changes into maint-1.0 that will stop the BGC compsets from getting the correct defaults on master after the maint-1.1 merge, it could be a big mess.

bishtgautam:
>The root of the problem is a conflict between two obvious goals:
A) Don't change existing compsets, so they can be rerun with newer code,

If you try to achieve the above goal, compsets for v2.0 would need to be given a different alias and we can create these new compsets correctly this time.

>Add 'hgrid=xx' to the fsurdat and finidat lines in the CLM use_case files (it is already there for some of them).

Unfortunately, this doesn't work. My suggestion is to **only** fix compsets on maint-1.0 that we want to run on a resolution that is different from the resolution those compsets were initially designed to run on.


>At some point maint-1.1 will need to be merged to master and all the compsets should work. Hence, if you put changes into maint-1.0 that will stop the BGC compsets from getting the correct defaults on master after the maint-1.1 merge, it could be a big mess.

I'm assuming that current watercycle and BGC compsets on maint-1.0 and maint-1.1 won't be exactly BFB in v2.0. If that is true, you can either:
- Merge maint-1.0 and maint-1.1 to master; then fix the compsets to be consistent, or
- Directly fix the compsets on master for v2.0.

components/clm/bld/namelist_files/use_cases/1950_CMIP6HR_control.xml
tangq: I am not sure about other CMIP6 F-compsets, but there are what I used on maint-1.0:
F1850SC5-CMIP6.ne30
F2010C5-CMIP6-LR.ne30
F2010C5-CMIP6-HR.ne120

I made the comment above on the 1950 file, but it is also true for 1850_CMIP6_control.xml and 1850_SCMIP6_control.xml (one of which is used by F1850SC5-CMIP6) that contain the finidat setting. Is there a reason why it's not moved to the default file? As long as there is a reason for that (not overlooked), I am fine with the changes.

components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml
```diff
@@ -271,7 +294,7 @@ lnd/clm2/surfdata_map/surfdata_1x1_brazil_simyr1850_c140610.nc</fsurdat>
 
 
 <fsurdat hgrid="ne30np4"      sim_year="1850" use_crop=".false." >
-lnd/clm2/surfdata_map/surfdata_ne30np4_simyr1850_c180306.nc</fsurdat>
+lnd/clm2/surfdata_map/surfdata_ne30np4_simyr1850_2015_c171018.nc</fsurdat>
```
rljacob: Why is the timestamp going backwards?  From 180306 to 171018 ?

cameronsmith1:
> > The root of the problem is a conflict between two obvious goals:
> > A) Don't change existing compsets, so they can be rerun with newer code,
> 
> If you try to achieve the above goal, compsets for v2.0 would need to be given a different alias and we can create these new compsets correctly this time.

I expect the v2 simulations will use the same CMIP6 input files but change the tuning.  If that is true, then the v2 compsets will need different aliases.

What do you consider to be doing them 'correctly'?  

> > Add 'hgrid=xx' to the fsurdat and finidat lines in the CLM use_case files (it is already there for some of them).
> 
> Unfortunately, this doesn't work. My suggestion is to **only** fix compsets on maint-1.0 that we want to run on a resolution that is different from the resolution those compsets were initially designed to run on.

Why won't my idea work?

I think your strategy will ultimately work for the CMIP6 compsets.  However, there are others, such as the AV1C compsets (used in the development, initial tuning, and publication of the atm for v1) and the v0 compsets (the baseline from which E3SM improvements have been measured), which don't specify the land datasets, and just use whatever is the land default.  I think some of the AV1C compsets were run with ne30np4 (can you remember @tangq ?), so their land files will now be different.

> > At some point maint-1.1 will need to be merged to master and all the compsets should work. Hence, if you put changes into maint-1.0 that will stop the BGC compsets from getting the correct defaults on master after the maint-1.1 merge, it could be a big mess.
> 
> I'm assuming that current watercycle and BGC compsets on maint-1.0 and maint-1.1 won't be exactly BFB in v2.0. If that is true, you can either:
> 
>     * Merge maint-1.0 and maint-1.1 to master; then fix the compsets to be consistent, or
> 
>     * Directly fix the compsets on master for v2.0.

The compsets are already not BFB with master because the code has changed.  But probably the better question is whether the same namelist files get generated.  The answer to this is probably already 'no', but I think that is mostly because new namelist parameters have been added, which shouldn't really count.  Hence the even better question is whether any of the _compset-relevant_ namelist options are different.  This is less well defined but I think it is what we really mean.

components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml
bishtgautam: Because all supported 1850 compsets were using the older file. (e.g. https://github.com/E3SM-Project/E3SM/pull/3333/files#diff-7ece8b6f482388c76e3a4cab97c792eeL34)

cameronsmith1: The c171018 file is the surfdata file used in the CMIP6 compsets.   I don't know where the c180306 file came from, or why it is a default.

cameronsmith1: Where did the c180306 file come from?
What was it for?

tangq:
@cameronsmith1 , do you mean these **AV1C** compsets on ne30?
Screen Shot 2019-12-04 at 5 09 45 PM

cameronsmith1:
Yes.  Those AV1C compsets just pickup the default land options (AFAIK).  The older AV1C compsets probably don't matter much (IMHO).  But I recall that the Rasch E3SM-atm model description paper was based on the AV1C-04P2 compsets.  My recollection is those simulations used ne30np4, although I could be wrong.

bishtgautam:
@cameronsmith1 

>What do you consider to be doing them 'correctly'?

Don't add grid specific values in the xml file when defining a compset

> Why won't my idea work?

I tried adding `hgrid = ne30np4`, but the case for`--res ne4_ne4` still picks up the file corresponding to `ne30np4`. I have no idea if this is a bug related to how the xml file is being correctly parsed or the xml-related code wasn't intended to what we are trying to do.

cameronsmith1:
> > What do you consider to be doing them 'correctly'?
> 
> Don't add grid specific values in the xml file when defining a compset

As I understand it, this is essentially the policy B from my earlier comment.  Unfortunately it will inevitably break policy A (ability to recreate existing compsets in a modern codebase).   Ultimately, I think it would be best to setup a hybrid system that follows either policy A or B depending on the type of simulation (A for CMIP simulations, and B for general simulations).
 
> > Why won't my idea work?
> 
> I tried adding `hgrid = ne30np4`, but the case for`--res ne4_ne4` still picks up the file corresponding to `ne30np4`. I have no idea if this is a bug related to how the xml file is being correctly parsed or the xml-related code wasn't intended to what we are trying to do.

This has worked when I have tried it, so I am not sure why it isn't working for you.  I just talked to @tangq and he confirmed that it worked for him recently for the atm xml file.

bishtgautam:
maint-1.0 has `<fsurdat hgrid="ne120np4" > ... </fsurdat>` on [2010_CMIP6HR_control.xml#L21](https://github.com/E3SM-Project/E3SM/blob/maint-1.0/components/clm/bld/namelist_files/use_cases/2010_CMIP6HR_control.xml#L21). The fsurdat is incorrectly set for the following test using maint-1.0:
```
>./create_test SMS.ne4_ne4.F2010C5-CMIP6-HR --no-build

>cat <case-dir>/CaseDocs/lnd_in | grep fsur
 fsurdat = '/pic/projects/climate/csmdata//lnd/clm2/surfdata_map/surfdata_ne120np4_simyr2010_c20181023.nc'
```

Maybe it is only an issue with land xml or the way fsurdat is defined in the xml file. Do you guys have any idea what is going wrong?

cameronsmith1:
Unfortunately I haven't delved deeply into the xml processing.  All I can say is that when I have used the hgrid condition it has worked, although I confess that I haven't tested that it did the right thing for the right reason.  

@rljacob do you know what is going on?  Or who else might know about the xml processing?

rljacob:
@jgfouca should be able to help.

jgfouca:
I'm a bit unclear on the problem. I tried this:
```
%  ./create_test SMS.ne4_ne4.F2010C5-CMIP6-HR --no-build
% cd $CASE
% ./xmlquery GRID
GRID: a%ne4np4_l%ne4np4_oi%ne4np4_r%r05_m%oQU240_g%null_w%null
```

At first glance, this looks correct. Is there something else I should be testing?

rljacob:
This PR needs a better description.  Open an issue for what exactly was broken about the ne4 F compsets.  In general, they have been working fine.  Clarify title.

rljacob:
@bishtgautam what is the status of this PR?

rljacob:
Go ahead and merge this.

PR #3334, opened 2019-12-03
LDFLAGS should not be added via target_link_libraries

jgfouca:
target_link_libraries adds -l and is clearly intended for adding
libraries, not general link flags. target_link_options is fairly
new to CMake, so we use CMAKE_EXE_LINKER_FLAGS instead.

[BFB]


jgfouca:
Looks good on melvin, merging to next.

jgfouca:
@bishtgautam

PR #3338, opened 2019-12-05
Fix F2010SC5-CMIP6.northamericax4v1_r0125_northamericax4v1

tangq:
The fix includes two parts: 1) adding ATM2ROF maps between the northamericax4v1 atm grid and the r0125 land grid; 2) turn off atm_c2_rof coupling when atm_prognostic is true.

### DIFF
--- a/cime/config/e3sm/config_grids.xml
+++ b/cime/config/e3sm/config_grids.xml
@@ -2763,6 +2763,11 @@
       <map name="LND2ATM_SMAPNAME">cpl/gridmaps/r0125/map_r0125_to_northamericax4v1_mono.20191022.nc</map>
     </gridmap>
 
+    <gridmap atm_grid="ne0np4_northamericax4v1" rof_grid="r0125">
+      <map name="ATM2ROF_FMAPNAME">cpl/gridmaps/northamericax4v1np4/map_northamericax4v1_to_r0125_mono.20191022.nc</map>
+      <map name="ATM2ROF_SMAPNAME">cpl/gridmaps/northamericax4v1np4/map_northamericax4v1_to_r0125_mono.20191022.nc</map>
+    </gridmap>
+
     <gridmap atm_grid="T62" ocn_grid="gx3v7">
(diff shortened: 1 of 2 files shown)


rljacob:
If you're doing 1, do you still have to do 2?

tangq:
@rljacob , to answer your question I will need to know more about the behavior of MOZART-heat. If atm_c2_rof coupling does not change the results of other components than MOZART, removing 2) should be fine. Otherwise, it feels saver to keep 2) till we have plans to analyze MOZART-heat results from simulations with prognostic atmosphere.

jonbob:
@tangq - @rljacob makes a good point. MOSART-heat does not feed back to the coupled system at all and the heatflag is off by default anyway. But I know the MOSART developers would like to be able to test without an active atmosphere, so removing 2) is preferable for them.

tangq:
@jonbob , I am a bit confused. I thought what 2) does is to ensure the MOSART tests are done without an active atmosphere, which is desired by the MOSART developers.

I might be wrong about what the developers wanted. But either way if their changes do not feedback to the coupled system as you mentioned above, I don't see a reason to keep 2).

tangq:
I removed the atm_prognostic flag as suggested from the discussions above. It should be ready to go.

jonbob:
@rljacob - is it OK to merge this now? Or do we need to wait for @eroesler to review?

jonbob:
Sorry, @eroesler was a suggestion and not a requested reviewer. I'll merge now

jonbob:
@tangq - is there an issue that this fixes? The "bug fix PR" label means this should have an issue number associated with it

tangq:
@jonbob , this PR provides the atm2rof maps for the F2010SC5-CMIP6.northamericax4v1_r0125_northamericax4v1 configuration. I am not aware of an issue associated with it. I removed the "bug fix PR".

jonbob:
Perfect -- I'll get it merged right away. Thanks

jonbob:
merged to next

tangq:
Thank you, @jonbob

jonbob:
You're welcome, @tangq -- thanks for getting this together and everything to make it work

jonbob:
merged to master

PR #3344, opened 2019-12-06
Fixes IC for BGC test for CMIP6 compset

### LINKED ISSUE #3272: Test SMS.ne30_oECv3.BGCEXP_BCRC_CNPRDCTC_1850 fail on cori
Test SMS.ne30_oECv3.BGCEXP_BCRC_CNPRDCTC_1850 fail with ERROR:  ERROR mpas-o buildnml: ocn bgc cannot be run with more than 1 thread. 
When run step by step using --no-setup, after the thread thing is fixed, the simulation fail with error in reading restart file.

bishtgautam:
The SMS.ne30_oECv3.BGCEXP_BCRC_CNPRDCTC_1850 test was using an incorrect
land initial condition, which corresponded to SP mode. Now the test
does a cold start in the land model.

Fixes #3272
[BFB]

### DIFF
--- a/components/clm/bld/namelist_files/use_cases/1850_CMIP6_control.xml
+++ b/components/clm/bld/namelist_files/use_cases/1850_CMIP6_control.xml
@@ -33,6 +33,6 @@
       
 <fsurdat>lnd/clm2/surfdata_map/surfdata_ne30np4_simyr1850_2015_c171018.nc </fsurdat>
 <flanduse_timeseries>lnd/clm2/surfdata_map/landuse.timeseries_ne30np4_hist_simyr1850_c20171102.nc </flanduse_timeseries>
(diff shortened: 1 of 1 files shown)


jonbob:
@bishtgautam - was the problem in maint-1.0 or master? I don't think maint-1.0 is used for any BGC runs....

bishtgautam:
@jonbob You are correct that maint-1.0 wasn't used for BGC and I will redo this PR for master. Also, I now also realize that the failing test work with the current master when one uses testmod (e.g. SMS.ne30_oECv3.BGCEXP_BCRC_CNPRDCTC_1850.sandiatoss3_intel.clm-bgcexp)

rljacob:
@bishtgautam should this PR be closed?

Issue #3342, opened 2019-12-05
rounding problem on theta


sarich:
PR #3309 broke the builds on theta. They were broken the weeks before for various other reasons, those may be fixed now.
cime/scripts/lib/CIME/XML/env_batch.py:
```
￼                     # We don't want floating-point data
￼                     try:
￼                         rval = round(float(rval))
￼                     except ValueError:
￼                         pass

```
round() returns an integer for python3, but returns a float for python2. Now submit is trying to use 128.0 procs.

```
---------------------------------------------------
2019-11-23 15:10:34: ERROR: Command: '/projects/ccsm/acme/tools/cobalt/dsub -A ClimateEnergy_3 -t 00:31:40 -n 128.0 -q default --mode script --env ARGS_FOR_SCRIPT='--skip-preview-namelist' .case.test' failed with error 'Usage: qsub.py --help
Usage: qsub.py [options] <executable> [<excutable options>]

Refer to man pages for JOBID EXPANSION and SCRIPT JOB DIRECTIVES.


qsub.py: error: option -n: invalid integer value: '128.0'' from dir '/lus/theta-fs0/projects/ClimateEnergy_3/climenrg/J/PET.f19_g16.X.theta_intel.allactive-mach-pet.JNextIntegration20191123_034231'
 ---------------------------------------------------
```

rljacob:
@jgfouca what to do here?  Does this mean we have to make the test user have python3 in their path before starting to run?

jgfouca:
@sarich , @rljacob , boy is that irritating. I hope we can move CIME to python3-only soon so we can stop hitting these issues.

Please change the code to this:

```
rval = int(round(float(rval)))
```

Which should make it an int in all cases.

ndkeen:
I was hitting this same issue this morning. Nice work finding the problem. The change proposed did allow a test job to submit.

Would be nice if the job names were used (currently the job name is blank).

amametjanov:
Fixed by E3SM-Project/E3SM#3345

PR #3345, opened 2019-12-06
fix python rounding issue

sarich:
Fixes issue #3342  
round() returns an integer for python3 and a float
for python2, this forces an integer for either case


jgfouca:
Skipped integration testing.

PR #3350, opened 2019-12-09
Fix Depends-file usage for legacy build system

jgfouca:
The legacy system only used the Depends.mach.compiler file if
it existed. The Cmake system does not work that way and changes
were implemented that caused all matching Depends files to be
used in both build systems.

This commit restores the legacy behavior to the legacy system.

[BFB]


jgfouca:
@worleyph

PR #3352, opened 2019-12-10
Update mpas-source for compass initialization

mark-petersen:
Add COMPASS initialization scripts for CUSP8 and CUPS12. Update jigsaw scripts for planar meshes.

[BFB]


jonbob:
merged to next

jonbob:
merged to master

PR #3353, opened 2019-12-10
make supercell test case threadsafe

mt5555:
minor bugfixes for the standlone HOMME supercell test case

thread safe initialization
bug fix for restart
turn on lustre hints on Cori-KNL
[BFB]


components/homme/src/test_mod.F90
```diff
@@ -77,7 +77,8 @@ subroutine set_test_initial_conditions(elem, deriv, hybrid, hvcoord, tl, nets, n
   endselect
 
   !initial conditions for initial run, runtype=0
-  if (runtype == 0) then
+  ! also does other test case setup.
```
oksanaguba: Does this init code run for restart runs? What does the comment above, 'other test case', mean?

components/homme/src/test_src/dcmip12_wrapper.F90
```diff
@@ -634,7 +634,7 @@ subroutine set_hybrid_coefficients(hv, hybrid, eta_t, c)
     tmp        = max( (hv%etai(k)-eta_c)/(1.0-eta_c), 0.0_rl)
     hv%hybi(k) = tmp**c
     hv%hyai(k) = hv%etai(k) - hv%hybi(k)
-    if(hybrid%par%masterproc) write(*,'(i4,a,f18.15,a,f18.15,a,f18.15)') &
+    if(hybrid%masterthread) write(*,'(i4,a,f18.15,a,f18.15,a,f18.15)') &
```
oksanaguba: Why change masterproc to masterthread here? Without looking, i'd think this is not intended?

components/homme/src/test_mod.F90
mt5555: i meant the other setup in addition to the initial condition.  some test cases (like supercell) initialize arrays that need to be done both on new runs and restart runs.

components/homme/src/test_src/dcmip12_wrapper.F90
mt5555: it's in a threaded region, but we only want one thread to print this message.

PR #3354, opened 2019-12-11
Fix erroneous memset calls

### LINKED ISSUE #3351: reversed parameters in 2 calls to memset in gptl.c in the timing library
Two calls to memset in gptl.c in the timing library have their arguments in the wrong order.

line 3672

     memset( timerlist[t], length  * count[t] * sizeof (char), 0 );

should be

     memset( timerlist[t], 0, length  * count[t] * sizeof (char) );

and line 3694

     memset( newtimers, max_count * sizeof (char *), 0 );

should be

     memset( newtimers, 0, max_count * sizeof (char *) );

The other calls to memset in gptl.c are correct. This error has not caused any problems (and has been in there for a long time) because the third argument specifies the number of locations to be "filled" by the value of the second argument. So putting the '0' fill value as the third parameter means that no memory is being initialized (and initialization to '0' apparently is not really necessary).

This still needs to be fixed. This issue is being created so that a subsequent PR can point to the issue as a bug fix.

worleyph:
Two calls to memset in gptl.c in the timing library have their
arguments in the wrong order. The other calls to memset in gptl.c are
correct. In practice this has not caused any problems (due to the
values being swapped), and the result is BFB. To avoid future issues,
it is still important to fix this.

Fixes  #3351 

[BFB]

### DIFF
--- a/cime/src/share/timing/gptl.c
+++ b/cime/src/share/timing/gptl.c
@@ -3669,7 +3669,7 @@ static int merge_thread_data()
     /* allocate memory to hold list of timer names */
     if( !( timerlist[t] = (char *)malloc( length * count[t] * sizeof (char)) ) && count[t])
       return GPTLerror ("%s: memory allocation failed\n", thisfunc);
-    memset( timerlist[t], length  * count[t] * sizeof (char), 0 );
+    memset( timerlist[t], 0, length  * count[t] * sizeof (char) );
 
     x = 0;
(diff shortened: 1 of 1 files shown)


worleyph:
ran a simple sanity test on Compy (F case, ne4_ne4 on 96 processes without threading), and it compiled, ran, and was BFB (and timing data also looked fine).

jgfouca:
Skipping integration.

PR #3359, opened 2019-12-11
Clean up use of LDFLAGS in config_compilers.xml

### LINKED ISSUE #3356: PnetCDF link errors on Summit related to LDFLAGS
To reproduce this issue, try to build ne4 F case on Summit with GNU compiler (the default XL compiler has other build errors so far, see issue #3355).

In config_compilers.xml, Summit relies on LDFLAGS to set PnetCDF library for the linker:
```
<compiler MACH="summit" COMPILER="gnu">
...
  <LDFLAGS>
    <append>-L$ENV{NETCDF_C_PATH}/lib -lnetcdf -L$ENV{NETCDF_FORTRAN_PATH}/lib -lnetcdff -L$ENV{ESSL_PATH}/lib64 -lessl -L$ENV{NETLIB_LAPACK_PATH}/lib -llapack</append>
    <append MPILIB="!mpi-serial"> -L$ENV{PNETCDF_PATH}/lib -lpnetcdf -L$ENV{HDF5_PATH}/lib -lhdf5_hl -lhdf5 </append>
  </LDFLAGS>
...
</compiler>
```

After PR #3334 is merged to master, LDFLAGS will only be used for general link flags, not for adding libraries. Then we will encounter link errors for PnetCDF:
```
bld/ibm/spectrum-mpi/nodebug/nothreads/mct/lib/libpio.a(nf_mod.F90.o): In function `__nf_mod_NMOD_inquire_dimlen':
externals/scorpio_classic/pio/nf_mod.F90:1420: undefined reference to `nfmpi_inq_dimlen'
```
@jgfouca 
According to other machine settings, there are 2 possible fixes:
1) Move all libraries (including PnetCDF) from LDFLAGS to SLIBS
2) Use <PNETCDF_PATH> ... </PNETCDF_PATH> to set PnetCDF path

PS, besides Summit, we need to check if some other machines are also affected by PR #3334 especially when they contain libraries in LDFLAGS.

jgfouca:
LDFLAGS should be used for adding general linker flags.

SLIBS should be used for adding libs to be linked.

Fixes #3356 

[BFB]

### DIFF
--- a/cime/config/e3sm/machines/config_compilers.xml
+++ b/cime/config/e3sm/machines/config_compilers.xml
@@ -6,15 +6,13 @@ This file defines compiler flags for building CIME.  General flags are listed fi
 followed by flags specific to particular operating systems, followed by particular machines.
 
 More general flags are replaced by more specific flags.
-Flags of the sort ADD_FLAG indicate that the field should be appended to an already existing FLAG definition.
+Flags of the sort <append> indicate that the field should be appended to an already existing FLAG definition.
 
 Attributes indicate that an if clause should be added to the Macros so that these flags are added
 only under the conditions described by the attribute(s).
 
 The env_mach_specific file may set environment variables or load modules which set environment variables
 which are then  used in the Makefile.   For example the NETCDF_PATH on many machines is set by a module.
-
-Do not use variables CPPDEFS and SLIBS here, instead use ADD_CPPDEFS and ADD_SLIBS
 
 ========================================================================
  Serial/MPI compiler specification
@@ -67,8 +65,8 @@ a C++ linker. If CXX_LINKER=FORTRAN, then the above CXX_LDFLAGS line
 should specify extra LDFLAGS needed when linking C++ and fortran code
 using a fortran linker.
 
-These should NOT be specified via <ADD_SLIBS USE_CXX="true"> or
-<ADD_LDFLAGS USE_CXX="true">, because those mess up the configure step
+These should NOT be specified via <SLIBS> <append USE_CXX="true"> or
+<LDFLAGS> <append USE_CXX="true">, because those mess up the configure step
 for mct, etc.
 
 ========================================================================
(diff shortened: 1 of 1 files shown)


cime/config/e3sm/machines/config_compilers.xml
```diff
@@ -1869,11 +1848,11 @@ ntel/x86_64/2013/composer_xe_2013/composer_xe_2013_sp1.3.174/mkl/include </base>
     <append> -qzerosize -qfree=f90 -qxlf2003=polymorphic</append>
     <append> -qspillsize=2500 -qextname=flush </append>
   </FFLAGS>
-  <LDFLAGS>
+  <SLIBS>
     <append>-lxlopt -lxl -lxlsmp -L$ENV{NETCDF_C_PATH}/lib -lnetcdf -L$ENV{NETCDF_FORTRAN_PATH}/lib -lnetcdff -L$ENV{ESSL_PATH}/lib64 -lessl -L$ENV{NETLIB_LAPACK_PATH}/lib -llapack</append>
     <append MPILIB="!mpi-serial"> -L$ENV{PNETCDF_PATH}/lib -lpnetcdf -L$ENV{HDF5_PATH}/lib -lhdf5_hl -lhdf5 </append>
     <append>  -Wl,--relax -Wl,--allow-multiple-definition </append>
```
dqwu: Not sure if this line should still be kept as LDFLAGS (is it kind of general linker flags?)

cime/config/e3sm/machines/config_compilers.xml
```diff
@@ -984,9 +976,11 @@ flags should be captured within MPAS CMake files.
   <LAPACK_LIBDIR>/software/tools/compilers/intel_2017/mkl/lib/intel64</LAPACK_LIBDIR>
   <LDFLAGS>
     <append compile_threaded="TRUE"> -fopenmp </append>
-    <append COMP_CLASS="cpl"> -L$NETCDF_PATH/lib -Wl,-rpath=$NETCDF_PATH/lib -lnetcdff -lnetcdf </append>
     <!-- <append COMP_CLASS="cpl"> -L$ENV{CLM_PFLOTRAN_SOURCE_DIR} -lpflotran $ENV{PETSC_LIB} </append> -->
   </LDFLAGS>
+  <SLIBS>
+    <append> -L$NETCDF_PATH/lib -Wl,-rpath=$NETCDF_PATH/lib -lnetcdff -lnetcdf </append>
```
dqwu: Is there a reason to remove `COMP_CLASS="cpl"` in the original append line here?

cime/config/e3sm/machines/config_compilers.xml
jgfouca: Good catch, this was a mistake.

cime/config/e3sm/machines/config_compilers.xml
```diff
@@ -1574,7 +1556,7 @@ flags should be captured within MPAS CMake files.
   <LDFLAGS>
     <base> -mkl -lstdc++ </base>
     <append compile_threaded="TRUE"> -qopenmp </append>
-    <append COMP_CLASS="cpl">-L$(NETCDF_FORTRAN_PATH)/lib64</append>
+    <append>-L$(NETCDF_FORTRAN_PATH)/lib64</append>
```
dqwu: Is there a reason to remove `COMP_CLASS="cpl"` here?

cime/config/e3sm/machines/config_compilers.xml
jgfouca: Whoever did that just didn't understand our build system. We only do linking once for e3sm.exe, so it's pointless/redundant to limit LDFLAGS to the coupler.

dqwu: OK, thanks for the information.

jgfouca:
Skipping integration.

PR #3360, opened 2019-12-11
Change all ocean output files to single precision

mark-petersen:
Currently, all MPAS-Ocean output files are double precision. This is simply not needed, and doubles file space and write time. I added the single line to specify single precision to all output streams with 2D and 3D fields. For standard simulations, this only alters:
- monthly time averages, used for MPAS-Analysis
- high frequency stats, used for animations

Restart files remain double precision, of course. The time average is computed in double precision, and the final value is then written in single precision.

This will be bit-for-bit for comparison of restart and coupler files, but not for ocean output files.

[BFB]

### DIFF
--- a/components/mpas-ocean/cime_config/buildnml
+++ b/components/mpas-ocean/cime_config/buildnml
@@ -641,6 +641,7 @@ def buildnml(case, caseroot, compname):
         lines.append('')
         lines.append('<stream name="eddyProductVariablesOutput"')
         lines.append('        type="output"')
+        lines.append('        precision="single"')
         if ocn_grid.startswith("oRRS1"):
             lines.append('        io_type="pnetcdf,cdf5"')
         else:
@@ -663,6 +664,7 @@ def buildnml(case, caseroot, compname):
         lines.append('')
         lines.append('<stream name="highFrequencyOutput"')
         lines.append('        type="output"')
+        lines.append('        precision="single"')
         lines.append('        io_type="{}"'.format(ocn_pio_typename))
         lines.append('        filename_template="mpaso.hist.am.highFrequencyOutput.$Y-$M-$D_$h.$m.$s.nc"')
         lines.append('        filename_interval="00-01-00_00:00:00"')
@@ -710,6 +712,7 @@ def buildnml(case, caseroot, compname):
         lines.append('')
         lines.append('<stream name="mixedLayerDepthsOutput"')
         lines.append('        type="output"')
+        lines.append('        precision="single"')
         lines.append('        io_type="{}"'.format(ocn_pio_typename))
         lines.append('        filename_template="mpaso.hist.am.mixedLayerDepths.$Y-$M-$D.nc"')
         lines.append('        filename_interval="00-01-00_00:00:00"')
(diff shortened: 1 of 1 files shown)


mark-petersen:
@vanroekel and I had a conversation about this last month, and we couldn't think of any reason to save any output in double precision. Can anyone else?

One possible accommodation is to save all density fields as (density-1000) to gain a few digits. This would require adding a few variables.

@xylar does MPAS-Analysis make any assumption about double precision fields in the monthly time averages?

xylar:
@mark-petersen, all I can say is that I ran into some pretty serious plotting artifacts when I worked with 4-byte-float salinity output from POP back in the day, though I'm not sure why.  Density does seem like a case where we would be more likely to run into trouble, and writing out density - 1000 seems like a pain (plus a pain to parse again).  But we could try just writing out densities with 4-byte floats and run the analysis to see how it goes.

Can we do a test G-case run of a year or so and run MPAS-Analysis on it to make sure we're not going to have problems?

mark-petersen:
Ran one year of EC60to30 G case on cori-knl, available here:
```
/global/cscratch1/sd/mpeterse/acme_scratch/cori-knl/191218_oEC60to30v3_single_precision_output_1
```
and confirmed that these streams are all writing 4-byte reals:
```
grep 'precision="single"' streams.ocean -B 2|grep stream
<stream name="eddyProductVariablesOutput"
<stream name="highFrequencyOutput"
<stream name="mixedLayerDepthsOutput"
<stream name="timeSeriesStatsMonthlyOutput"
<stream name="timeSeriesStatsMonthlyMaxOutput"
<stream name="timeSeriesStatsMonthlyMinOutput"
```
e.g.
```
ncdump mpaso.hist.am.timeSeriesStatsMonthly.0001-12-01.nc|grep float|head
	float timeMonthly_avg_ssh(Time, nCells) ;
...
```
where `float` is 4-byte and `double` is 8-byte. Next, test with MPAS-Analysis...

xylar:
@mark-petersen, I've got a job in to run the full analysis on the output.  I'll post the results here as soon as they're up.

xylar:
The analysis is here:
https://portal.nersc.gov/project/m3412/mpas_analysis_output/test/191218_oEC60to30v3_single_precision_output_1/yrs001-001/

I don't see any troubling artifacts of the single precision, even in potential density, so I think we're good to go.

mark-petersen:
@xylar thanks for the quick analysis! That is very reassuring!

rljacob:
Waiting on further analysis from @mark-petersen

mark-petersen:
@milenaveneziani @vanroekel @maltrud Does anyone want to take a look at the analysis Xylar posted and comment or approve? I really think we need to change our output from double to single precision.

vanroekel:
@mark-petersen I'm not sure what to look for in the analysis.  With only a one year run, it's unclear what a problem would look like.  I'm fine approving this based on @xylar's pass through the plots

milenaveneziani:
What about tendency fields? Could we keep outputting those in DB?

milenaveneziani:
I just spoke with Tarun, who's been calculating freshwater budgets for the CESM LENS model outputs: he says that all terms were saved as single precision in that case.
So, I think (hope) we should be fine making the switch, even for T and S budget terms.

mark-petersen:
@jonbob, we are still testing the Redi PR, so it makes sense to merge this now. Please proceed when you have the chance.

rljacob:
NOTE:  If any of our test cases output ocean history files, I think cprnc will complain that the single precision are not BFB with the double precision ones.

jonbob:
merged to next

jonbob:
merged to next

jonbob:
merged to next

jonbob:
merged to master

PR #3363, opened 2019-12-12
Fix preqx_kokkos/HOMMEBFB

oksanaguba:
Minor fix to use gllfvremap_util_mod in preqx_kokkos. Issue happened because a PR that affects this file was merged to master very recently, so, clone with hommebfb test did not have that change.


oksanaguba:
I merged today's master into this branch and ran hommebfb test on skybridge. test passed.

jgfouca:
Skipping integration.

Issue #1832, opened 2017-10-11
SMS_D_Ly2_P1x1.1x1_smallvilleIA.ICLM45CNCROP fails with GNU on several machines


ndkeen:
The DEBUG build for the smallville test in acme_developer fails with GNU.  It worked with Intel DEBUG.  This was on edison and I think it also failed with GNU/DEBUG on cori.

```
0: Backtrace for this error:
0: #0  0x13c418f in ???
0:      at /home/abuild/rpmbuild/BUILD/glibc-2.22/nptl/../sysdeps/unix/sysv/linux/x86_64/sigaction.c:0
0: #1  0xf35a75 in __shr_infnan_mod_MOD_shr_infnan_isnan_double
0:      at /global/cscratch1/sd/ndk/acme_scratch/edison/mf-intel18/SMS_D_Ly2_P1x1.1x1_smallvilleIA.ICLM45CNCROP.edison_gnu.force_netcdf_pio.20171011_090751_ngxc90/bld/gnu/mpt/debug/nothreads/MCT/noesmf/c
1a1l1i1o1r1g1w1e1/csm_share/shr_infnan_mod.F90:227
0: #2  0x835303 in __cnfiremod_MOD_cnfirearea
0:      at /global/cscratch1/sd/ndk/wacmy/ndk_machinefiles_nersc-intel18-release/components/clm/src/biogeochem/CNFireMod.F90:379
0: #3  0x8130c7 in __cnecosystemdynmod_MOD_cnecosystemdynnoleaching2
0:      at /global/cscratch1/sd/ndk/wacmy/ndk_machinefiles_nersc-intel18-release/components/clm/src/biogeochem/CNEcosystemDynMod.F90:799
0: #4  0x4f72cc in __clm_driver_MOD_clm_drv
0:      at /global/cscratch1/sd/ndk/wacmy/ndk_machinefiles_nersc-intel18-release/components/clm/src/main/clm_driver.F90:772
0: #5  0x4e532f in __lnd_comp_mct_MOD_lnd_run_mct
0:      at /global/cscratch1/sd/ndk/wacmy/ndk_machinefiles_nersc-intel18-release/components/clm/src/cpl/lnd_comp_mct.F90:464
0: #6  0x424bfd in __component_mod_MOD_component_run
0:      at /global/cscratch1/sd/ndk/wacmy/ndk_machinefiles_nersc-intel18-release/cime/src/drivers/mct/main/component_mod.F90:658
[...]

PR #3364, opened 2019-12-17
Update compiler flags on Theta

amametjanov:
Remove machine-specific flags and avoid FP exceptions on `invalid` FP values with GNU.

Addresses E3SM-Project/E3SM#1832

[BFB]


amametjanov:
Checked with `e3sm_integration` and `e3sm_hi_res` on Theta -- no errors.

rljacob:
@sarich please get this merged.

PR #3365, opened 2019-12-17
NH IMEX improvements

mt5555:
1.  Switch newton iteration to iterate on w instead of PHI.   This formulation removes all scans (for improved GPU performance), improves precision (allowing residuals down to 1e-11) and makes for better normalization of the Newton increment and residual norms.  

2.  Add "tstep_type=9", an IMEX method based on the KGU53 explicit table used by the hydrostatic code, and adding a 2nd order implicit table.

3.  adjust reference profiles to those found best for HS+topo simulations (but resulting in a degradation in the DCMIP Shar mountain test case)

4. remove obsolete IMEX timestepping algorithms.  

[non-BFB] for theta-l H and NH tests

testing and documentation of KGU53-IMEX method:
https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/1041401881/IMEX+methods

### DIFF
--- a/components/homme/src/theta-l/imex_mod.F90
+++ b/components/homme/src/theta-l/imex_mod.F90
@@ -102,7 +102,6 @@ contains
     real (kind=real_kind) :: dphi(np,np,nlev)    
     real (kind=real_kind) :: dphi_n0(np,np,nlev)    
     real (kind=real_kind) :: phi_n0(np,np,nlevp)    
-    real (kind=real_kind) :: delta_phi(np,np,nlevp)    ! phi_np1-phi_n0
     real (kind=real_kind) :: Ipiv(nlev,np,np)
     real (kind=real_kind) :: Fn(np,np,nlev),x(nlev,np,np)
     real (kind=real_kind) :: gwh_i(np,np,nlevp)  ! w hydrostatic
@@ -130,10 +129,10 @@ contains
     ! dirk settings
     maxiter=20
     deltatol=1.0e-13_real_kind  ! exit if newton increment < deltatol
+
     !restol=1.0e-13_real_kind    ! exit if residual < restol  
     ! condition number and thus residual depends strongly on dt and min(dz)
     ! more work needed to exit iteration early based on residual error
-    delta_phi(:,:,nlevp)=0
     min_rcond=1.0e20_real_kind
 
     do ie=nets,nete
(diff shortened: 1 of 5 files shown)


components/homme/src/theta-l/imex_mod.F90
```diff
@@ -69,31 +73,29 @@ subroutine compute_stage_value_dirk(n0,np1,alphadt,qn0,dt2,elem,hvcoord,hybrid,&
     ! compute_gwphis()            used to compute g*a(k) u(nt) dot grad_phis  in SPHI(nt)   
     !
     ! We then precompute:
-    !  w_rhs = w_0 + alphadt1 * SW(n0)  + alphadt1*SW(nm1) 
-    !  phi_rhs = w_0 + alphadt1 * SPHI(n0)  + alphadt1*SPHI(nm1) -  dt2*compute_gwphi(np1)
+    !  w_rhs = w_0 + alphadt_n0 * SW(n0)  + alphadt_nm1*SW(nm1) 
+    !  phi_rhs = w_0 + alphad_n0 * SPHI(n0)  + alphadt1_nm1*SPHI(nm1) -  dt2*compute_gwphi(np1)
```
ambrad: spelling: iteration

components/homme/src/theta-l/imex_mod.F90
```diff
@@ -113,12 +114,13 @@ subroutine compute_stage_value_dirk(n0,np1,alphadt,qn0,dt2,elem,hvcoord,hybrid,&
     real (kind=real_kind) :: Jac2U(nlev-1,np,np)
 
 
-    real (kind=real_kind) :: wgdtmax
+    real (kind=real_kind) :: wmax
```
ambrad: Switch `real*8` below to `real (kind=real_kind)`.

components/homme/src/theta-l/model_init_mod.F90
```diff
@@ -84,6 +84,19 @@ subroutine model_init2(elem,hybrid,deriv,hvcoord,tl,nets,nete )
       call phi_from_eos(hvcoord,elem(ie)%state%phis,&
            temp,elem(ie)%derived%dp_ref,elem(ie)%derived%phi_ref)
       elem(ie)%derived%theta_ref=0
+#endif
+#ifdef HV_REFSTATES_V2
```
ambrad: Does this change affect hydrostatic runs? Is this something for Luca to incorporate in his next merge?

components/homme/src/theta-l/prim_advance_mod.F90
```diff
@@ -118,10 +118,9 @@ subroutine prim_advance_exp(elem, deriv, hvcoord, hybrid,dt, tl,  nets, nete, co
 !                 optimal: for windspeeds ~120m/s,gravity: 340m/2
 !                 run with qsplit=1
 !                 (K&G 2nd order method has CFL=4. tiny CFL improvement not worth 2nd order)
-!   tstep_type=6  KG4+BE/CN  2nd order
 !   tstep_type=7  IMKG254a
 !   tstep_type=8  KG3+BE/CN  KG3 2nd order explicit, 1st order off-centering implicit
-!   tstep_type=9  KG5+BE/CN  KG5(3rd order linear, 3.87CFL) + BE/CN,  2nd order IMEX
+!   tstep_type=9  KGU53+BE/CN  KGU53 3rd order explicit, 1st order off-centering implicit
```
ambrad: Is this still supposed to say "1st order off-centering implicit"?

mt5555: fixed

components/homme/src/theta-l/model_init_mod.F90
mt5555: yes - but I think this code Luca will still be using the Fortran.

AaronDonahue:
@rljacob and @mt5555 , I've merged this to next and I think it will need a bless on the theta tests (if I'm not mistaken).  But I'm guessing that means it would also need a bless when merged to master.  Given the holiday week should I hold off on the merge to master until the New Year?

rljacob:
A bless is only done when merging to master.   If you can merge to master tomorrow and issue the bless request, that should be ok.

AaronDonahue:
Ok cool.

rljacob:
@mt5555 this also caused the HOMME and HOMMEBFB tests to fail.  Is that expected?

AaronDonahue:
Those tests appear to also be "theta" related:
```
The following tests FAILED:
	 11 - thetah-test22 (Failed)
	 12 - thetanh-test22 (Failed)
	 14 - thetah-TC-ftype4 (Failed)
	 15 - thetah-TC (Failed)
	 16 - thetanh-TC (Failed)
	 17 - thetanh-TC-nudiv (Failed)
	 18 - thetanh-TC-nutop (Failed)
	 19 - thetanhwet-TC (Failed)
```

mt5555:
yes, HOMMEBFB failures are also expected.

AaronDonahue:
@mt5555 I also see a fail on Anvil for test HOMMEBFB_P24.f19_g16_rx1.A.anvil_intel.  It looks like the test fails because there are now 11 test cases instead of 5.  I'm guessing this is expected as well?

oksanaguba:
on anvil hommebfb won't work because there is no cmake file for it. it won't build.

AaronDonahue:
ahhh ok. Just wanted to make sure.

PR #3367, opened 2019-12-23
Update mpas source framework for inc files

mark-petersen:
This is a minor MPAS framework update. It does not alter E3SM execution. It creates two new autogenerated config files on compile so that, in the future, each core may create a `core_config` module and include all the config flags at once, rather than declare each variable individually.

[BFB]


mark-petersen:
Passed:
```
./create_test -p acme -q debug --walltime 00:30:00 \
```
Also test compiled stand-alone MPAS with all cores, with gnu and intel on LANL IC.

mark-petersen:
@jonbob this is ready to merge when the repo is open.

jonbob:
Merged to  next

jonbob:
Merged to master

PR #3372, opened 2020-01-02
Disable Gravity Wave Latitude Taper for SE dycore

whannah1:
This disables the latitude taper used by the frontal gravity wave scheme. According to a conversation with Julio Bacmeister, this was added for stability in CAM when using a lat-lon grid with a polar filter. This is not required for an unstructured grid, so this disables the taper when dycore_is("UNSTRUCTURED") returns true.

[non-BFB]

### DIFF
--- a/components/cam/src/physics/cam/gw_drag.F90
+++ b/components/cam/src/physics/cam/gw_drag.F90
@@ -574,6 +574,7 @@ subroutine gw_tend(state, sgh, pbuf, dt, ptend, cam_in)
   use gw_oro,     only: gw_oro_src
   use gw_front,   only: gw_cm_src
   use gw_convect, only: gw_beres_src
+  use dycore,     only: dycore_is
   !------------------------------Arguments--------------------------------
   type(physics_state), intent(in) :: state      ! physics state structure
   ! Standard deviation of orography.
@@ -809,7 +810,11 @@ subroutine gw_tend(state, sgh, pbuf, dt, ptend, cam_in)
         call gw_cm_src(ncol, pgwv, kbotbg, u, v, frontgf, &
              src_level, tend_level, tau, ubm, ubi, xv, yv, c)
 
-        do_latitude_taper = .true.
+        if (dycore_is('UNSTRUCTURED')) then
+          do_latitude_taper = .false.
+        else
+          do_latitude_taper = .true.
+        end if
 
         ! Solve for the drag profile with C&M source spectrum.
         call gw_drag_prof(ncol, pgwv, src_level, tend_level, do_latitude_taper, dt, &
--- a/components/cam/src/physics/cam/gw_drag.F90
+++ b/components/cam/src/physics/cam/gw_drag.F90
@@ -669,6 +669,8 @@ subroutine gw_tend(state, sgh, pbuf, dt, ptend, cam_in)
   ! local override option for constituents cnst_type
   character(len=3), dimension(pcnst) :: cnst_type_loc
 
+  logical :: do_latitude_taper
+
   !------------------------------------------------------------------------
 
   ! Make local copy of input state.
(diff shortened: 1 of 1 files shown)


rljacob:
Do you know what this does to the climate?

whannah1:
@rljacob I think the effect on the climate will be minimal. I can run some multi-year tests to quantify the effect if you think it might be important to know, but personally I think it's just a good idea to do this either way since the latitude taper has no physical meaning and offers no practical benefit.

rljacob:
I agree it should be fixed.  But master is getting tuned for v2 so any non-BFB changes have to have their effects understood.

whannah1:
@rljacob do you want me to quantify the effect with some multi-year runs? Or would you rather just put this PR off until after the v2 release?

rljacob:
Lets see what @golaz thinks.

oksanaguba:
Is purpose of this PR to remove dependencies on lat/lon in physics for RCEMIP? Out of curiosity, I grepped for trig. functions and %lat in cam/src/physics/cam folder. I am copying below instances that may be relevant. Are these not an issue for RCEMIP?
```
(base) [onguba@blueslogin3 cam]$ pwd
/home/onguba/acme-b4/components/cam/src/physics/cam
(base) [onguba@blueslogin3 cam]$ grep 'cos(' *
ghg_data.F90:      coslat(i) = cos(clat(i))
phys_grid.F90:         dist2 = acos(sin(latr) * sin(clat) + cos(latr) * cos(clat) * cos(clon - lonr)) * rearth       
tphysidl.F90:   real(r8) acoslat                            ! abs(acos(coslat))
tphysidl.F90:      coslat (i) = cos(clat(i))
tphysidl.F90:               acoslat = abs(acos(coslat(i)))
tphysidl.F90:               acoslat = abs(acos(coslat(i)))
tropopause.F90:        tP = (25000.0_r8 - 15000.0_r8 * (cos(pstate%lat(i)))**2)

aoa_tracers.F90:                phys_state(c)%q(i,k,ixht) = 2._r8 + sin(phys_state(c)%lat(i))
aoa_tracers.F90:          qrel              = 2._r8 + sin(state%lat(i))
aoa_tracers.F90:          q(j,:) = 2._r8 + sin(lat(gcid(j)))
phys_grid.F90:         dist2 = acos(sin(latr) * sin(clat) + cos(latr) * cos(clat) * cos(clon - lonr)) * rearth       
tphysidl.F90:      sinsq  (i) = sin(clat(i))*sin(clat(i))

subcol_tstcp.F90:            if (state%lat(i) > 0.7854_r8) then
subcol_tstcp.F90:            else if (state%lat(i) < -0.7854_r8) then
subcol_tstcp.F90:            else if (state%lat(i) > 0.7854_r8) then
subcol_tstcp.F90:            else if (state%lat(i) < -0.7854_r8) then
```

whannah1:
@oksanaguba This isn't just for the RCEMIP compset, although that's definitely how I found out about it. The latitude taper just has no purpose when using the cubed sphere, so it seems sensible to make sure it's not used. 

I don't think any of those uses of latitude are important for the RCE compset. We won't use idealized physics or ghg data. the subcolumn one lookslike a potential issue, but my understanding is that no one plans on using the subcolumn feature in E3SM.

oksanaguba:
Climo evaluations for this PR are on https://acme-climate.atlassian.net/wiki/spaces/COM/pages/1098907683/Taper+PR+3372 .

whannah1:
@oksanaguba thanks for doing the test! How long were these runs?
The temperature and wind differences in DJF over the south pole are much larger than I expected!

whannah1:
Looking a the Golaz (2019) paper it looks like the southern polar warming shown by removing the taper here might be moving in the right direction because that paper showed a cold bias around 200-300 hPa over Antarctica.

oksanaguba:
6-year run takes maybe 18 hours on 60 anvil nodes for the default 1 degree setup, F compset.

oksanaguba:
About cold bias (which also is mentioned in Woollings&Thuburn "Entropy sources..." ): The new run has cold bias at SP reduced, but NP got warmer, too. In NP the default run is already 'too warm', in the new run temperatures got even higher: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/onguba/preqx/preqx-ne30-taperPR.115605.vs-default/viewer/lat_lon/model_vs_model/t-200mb-global/ann.html and https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/onguba/preqx/preqx-ne30-taperPR.115605.vs-obs/viewer/lat_lon/era-interim/t-200mb-global-era-interim/ann.html .

oksanaguba:
I looked at some plots for zonal T vs observations. Differences that come from this PR are very small compared to diffs against observations.

oksanaguba:
in next

PR #3373, opened 2020-01-02
Explicitly remove the darshan module for Cori

ndkeen:
Explicitly remove the darshan module for Cori.
This module does some IO statistics for NERSC, but is currently causing issues.
We do not need this module, so removing.

Also removing the setting of a PERL environment variable that was only
there temporarily while NERSC completed the perl install.

[bfb]


ndkeen:
I'm running e3sm_developer on cori-knl (and 1 simple test on cori-haswell).

PR #3375, opened 2020-01-04
Update RRTMGP to point to latest version on remote repo

brhillman:
Update RRTMGP submodule to point to latest version on external repository. This brings in a number of fixes for OpenACC, which we will need downstream on ACME-ECP to get the MMF configuration on the GPU, and will need eventually for SCREAM as well. This will cause BFB differences for the `cam-rrtmgp` test. These differences are due to only two changes:

1) Commit de24173cc97ec2e2d7eb30810ef205789c27e2bb in the external repo refactors the `get_col_dry` routine to put the calculations on the GPU via OpenACC. This requires splitting a computation into two parts, and this causes floating point differences that will produce a DIFF.

2) The absorption coefficient data is updated to the most recent version available from the remote repository. This also causes small differences.

Manually reverting both of these brings the test back to BFB, so we can be sure that the diff is resulting from these trivial changes and is acceptable. All other tests should be BFB.

[NBFB]

### DIFF
--- a/components/cam/src/physics/rrtmgp/radiation.F90
+++ b/components/cam/src/physics/rrtmgp/radiation.F90
@@ -475,7 +475,7 @@ contains
       ! impossible from this initialization routine because I do not thing the
       ! rad_cnst objects are setup yet.
       ! the other tasks!
-      ! TODO: This needs to be fixed to ONLY read in the data if masterproc, and then broadcast to
+      ! TODO: This needs to be fixed to ONLY read in the data if masterproc, and then broadcast
       call set_available_gases(active_gases, available_gases)
       call rrtmgp_load_coefficients(k_dist_sw, coefficients_file_sw, available_gases)
       call rrtmgp_load_coefficients(k_dist_lw, coefficients_file_lw, available_gases)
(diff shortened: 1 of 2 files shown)


singhbalwinder:
@brhillman : Thanks for this PR. Would you please select at least one reviewer for this PR? I will review this as well.

brhillman:
Thanks @singhbalwinder sorry I missed this the other day. I've asked @whannah1 to review as well, he already reviewed the corresponding PR that's going into ACME-ECP.

components/cam/src/physics/rrtmgp/radiation.F90
```diff
@@ -1618,10 +1620,6 @@ subroutine radiation_driver_lw(state, pbuf, cam_in, is_cmip6_volc, &
             ))
             call t_stopf('rad_calculations_lw')
 
-            ! Check stuff
```
whannah1: Are these checks somewhere else? Or are they not valid?

brhillman: These were debug checks I had inserted at one point when trying to track down some NaN's in a history write. I removed them here because I didn't think they were really necessary for production runs.

brhillman:
@singhbalwinder this should be ready to go when you can get to it.

singhbalwinder:
pushed to next

PR #3378, opened 2020-01-07
WIP Add anvil, melvin hommebfb cmake files

oksanaguba:


rljacob:
Why are machine-specific cmake pieces needed for this?

oksanaguba:
The way homme nightly (cime) tests are set is that cime calls `cmake -C ${MACHINE}.cmake ...` command to config homme. So, there is a cmake file for each machine. Now we need another cmake file for BFB runs because those will use different compiler flags. These homme builds do not use what's in cime in config_machines.xml, etc.

rljacob:
Ok I see HOMME already has those.

rljacob:
You're going to have add them for every machine we run e3sm_developer or e3sm_integration.

oksanaguba:
i don't have access to all machines. there may be way to disable hommebfb test on many machines.

rljacob:
We'll talk about this on the Thursday IG telecon.

oksanaguba:
Anvil file is ok, melvin testing is an issue (cprnc is not working on a rhel6 machine). Since now the plan is to separate homme testing from e3sm's, closing this pr for now.

PR #3385, opened 2020-01-10
update to fates api 8.1

rgknox:
This set of changes points E3SM to an updated fates tag, sci1.31.1_api8.1.0, which
has an updated parameter file format, now loaded on blues.

[non-BFB]

### DIFF
--- a/components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml
+++ b/components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml
@@ -105,7 +105,7 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <!-- ================================================================== -->
 <!-- FATES default parameter file                                       -->
 <!-- ================================================================== -->
-<fates_paramfile >lnd/clm2/paramdata/fates_params_default_12pft_c190905.nc</fates_paramfile>
+<fates_paramfile >lnd/clm2/paramdata/fates_params_default_12pft_api8.1.c200106.nc</fates_paramfile>
 
 
 <!-- soil order related parameters (relative to {csmdata}) -->


rgknox:
Tested land developer suite on cori.  Baseline comparison with master is b4b, except the fates run (expected).  NLFAIL is also expected on fates run.  Here are the fails, for reference:

```
  FAIL ERS.f19_g16.I1850CNRDCTCBC.cori-haswell_intel.clm-rd TPUTCOMP Error: Computation time increase > 10 pct from baseline
    FAIL ERS.f19_g16.I1850GSWCNPRDCTCBC.cori-haswell_intel.clm-ctc_f19_g16_I1850GSWCNPRDCTCBC TPUTCOMP Error: Computation time increase > 10 pct from baseline
  ERS_Ld3.f45_f45.ICLM45ED.cori-haswell_intel.clm-fates (Overall: NLFAIL) details:
    FAIL ERS_Ld3.f45_f45.ICLM45ED.cori-haswell_intel.clm-fates NLCOMP
    FAIL ERS_Ld3.f45_f45.ICLM45ED.cori-haswell_intel.clm-fates TPUTCOMP Error: Computation time increase > 10 pct from baseline
  SMS.r05_r05.I1850CLM45CN.cori-haswell_intel (Overall: FAIL) details:
    FAIL SMS.r05_r05.I1850CLM45CN.cori-haswell_intel RUN time=44
```

bishtgautam:
@rgknox Is this PR BFB for FATES test?

rgknox:
@bishtgautam , the FATES test should not be b4b.  
My tests compared our FATES api 8.1 tag to the tag before it.  But I'm looking at the differences between the hash E3SM used to point to, and the new hash, and there is plenty of answer changing code changes in there.

bishtgautam:
Could it be that the 11-days is too short for FATES vegetation to show up?

rgknox:
It is initialized with vegetation, they are just saplings, but they are there.  It is true though, that a non-trivial number of processes do not get exercised when we don't have any matured ecosystems represented.  The logistics of enabling and maintaining tests that have fully spun-up ecosystems are difficult though; we would constantly be changing and updating finidat files or some equivalent that contains fates demographic data, with every model updated.
Are you getting b4b results in your tests?

bishtgautam:
>Are you getting b4b results in your tests?

Yup, the E3SM nightly testing [reported](https://my.cdash.org/viewTest.php?onlypassed&buildid=1766469) ERS_Ld3.f45_f45.ICLM45ED.sandiatoss3_intel.clm-fates is BFB

bishtgautam:
Should I mark this as BFB and integrate it into the master? Or, you want to figure out why the FATES test is passing?

rgknox:
I would just move it through.  I'm confident things are fine.  Thanks @bishtgautam

Issue #2899, opened 2019-05-02
Potential issue with SMS_D.ne4_ne4.FC5AV1C-L using intel19 'Attempt to use pointer ELEM when it is not associated with a target'


ndkeen:
On Cori, I'm testing using newer versions of Intel MPI (impi) and the v19 compiler.

With `SMS_D_Ln5.ne4_ne4.FC5AV1C-L`, I see:

```
129: forrtl: severe (408): fort: (7): Attempt to use pointer ELEM when it is not associated with a target
129:
129: Image              PC                Routine            Line        Source
129: e3sm.exe           000000000BC8F186  Unknown               Unknown  Unknown
129: e3sm.exe           00000000024EC92E  dyn_comp_mp_dyn_i         189  dyn_comp.F90
129: e3sm.exe           0000000000A8C459  inital_mp_cam_ini          31  inital.F90
129: e3sm.exe           000000000083458A  cam_comp_mp_cam_i         159  cam_comp.F90
129: e3sm.exe           000000000080069F  atm_comp_mct_mp_a         312  atm_comp_mct.F90
129: e3sm.exe           000000000045B652  component_mod_mp_         257  component_mod.F90
129: e3sm.exe           000000000041C570  cime_comp_mod_mp_        1328  cime_comp_mod.F90
129: e3sm.exe           000000000045222D  MAIN__                    122  cime_driver.F90
129: e3sm.exe           000000000040A732  Unknown               Unknown  Unknown
129: e3sm.exe           000000000BD73619  Unknown               Unknown  Unknown
```

ndkeen:
Note that with  https://github.com/E3SM-Project/E3SM/pull/2939, I updated the version of Intel 19 to be 19.0.3.199. I wanted to note that this happens to be a case where the number of MPI's is larger than the number of elements. It could very well be that elem is used before init. Should we change the code to account for this?

Here is the code in dyn_comp.F90

```
    if (par%nprocs .lt. npes_cam) then
! Broadcast quantities to auxiliary processes
#ifdef SPMD
       call mpibcast(neltmp, 3, mpi_integer, 0, mpicom)
#endif
       if (.not.par%dynproc) then
          nelemdmax = neltmp(1)
          nelem     = neltmp(2)
          call set_horiz_grid_cnt_d(neltmp(3))
       endif
    endif


    !
    ! This subroutine creates mapping files using SE basis functions if requested
    !
    call native_mapping_readnl(NLFileName)
[...]

PR #3390, opened 2020-01-14
Update module version on cori in preparation for the new default CDT 19.11

ndkeen:
Update module version on cori in preparation for the new default CDT 19.11

This means updating to the following versions:

craype/2.6.2
cray-mpich/7.7.10
cray-netcdf-hdf5parallel/4.6.3.2
cray-hdf5-parallel/1.10.5.2
cray-parallel-netcdf/1.11.1.1
gcc/8.3.0
cray-libsci/19.06.1

Additionally, for CDT 19.11, the default is to use dynamic linking.
The tests I ran are passing without any additional modifications regarding linking.

Affects both cori-knl and cori-haswell

### DIFF
--- a/cime/config/e3sm/machines/config_machines.xml
+++ b/cime/config/e3sm/machines/config_machines.xml
@@ -137,7 +137,7 @@
       </modules>
 
       <modules mpilib="mpt">
-        <command name="swap">cray-mpich cray-mpich/7.7.6</command>
+        <command name="swap">cray-mpich cray-mpich/7.7.10</command>
       </modules>
 
       <modules compiler="intel">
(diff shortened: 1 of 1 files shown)


ndkeen:
As it's been a while, I have created another branch and re-implemented this change. Ran the tests again. There is a known failure with intel19 and our default ne4 tests. On cori-knl, the PE layout uses more MPI's than elements and that causes some issues that intel19 catches. So it's a combination of these 2 things. I can alter PE layout for ne4 cases on cori-knl to kick the can down the road, or try to fix the issue.  https://github.com/E3SM-Project/E3SM/issues/2899

ndkeen:
Closing this PR and will create another similar one.

PR #3392, opened 2020-01-14
UMich improved LW radiative treatments

wlin7:
Porting of the first implementation of University of Michigan's
improved longwave radiative treatments.

The new features include the use of surface spectral emissivity,
a new ice-cloud optics scheme (MC6), and a hybrid LW scattering solver.

Special note: This PR is created to aid the browsing of the updated codes. Not to be merged for now. No reviewer needs to be assigned either.


rljacob:
Is this supposed to eventually be in v2?

golaz:
This is a set of change that is under consideration for v2. By default, it needs to be off and BFB, with namelist flags to be turned on later.

components/cam/src/cpl/atm_comp_mct.F90
```diff
@@ -956,4 +1055,272 @@ subroutine atm_write_srfrest_mct( x2a_a, a2x_a, &
...
+      integer :: lat_varID,emis_varID, lon_varID
+      real, allocatable:: lat(:), lon(:)
+      integer ::start(4),count(4)
+      integer :: ncols,j
+      integer ::ilats, ilons
+      real(r8) :: lats2(ncols)           ! array of chunk latitudes
+      real(r8) :: lons2(ncols)           ! array of chunk longitude
+      real :: surface_emis(ncols, nbndlw), minvalue
+
+      filename = "/global/cscratch1/sd/xianwen/data/surface_emissivity_1x1_RRTMG_53deg.nc"
```
golaz: We cannot had stuff hard-coded like that.

wlin7: Totally agreed, @golaz . All these will be taken care of soon. Note that the input of this climatology surface emissivity data also need to be reworked.

brhillman:
Why are these changes being developed for RRTMG instead of RRTMGP?

wlin7:
Good point, @brhillman . Need to keep that in mind. The developer has not worked with RRTMGP. Also, the current E3SM (v1 and v2alpha) is using RRTMG by default. We have to go with RRTMG for now.

brhillman:
For what it's worth, I *believe* RRTMGP already has the flexibility to do scattering in the LW (would have to check with Robert), and surface emissivity input already has a spectral band dimension (although the current EAM treats every band the same so this is passed as a uniform array, but RRTMGP will handle spectral emissivity). It should also be straightforward to bring the new ice cloud optics scheme implemented here into the RRTMGP interface. What concerns me about what is going on here is that the active external/community development has shifted to RRTMGP, but here we are continuing to add new features to the older RRTMG code. Just something to think about.

rljacob:
Should switching to RRTMGP be a consideration for v2?

brhillman:
@rljacob I don't understand why it hasn't been considered, but I haven't been involved in any of those discussions. The code is a lot more flexible and probably easier to add these types of things to, it uses the same assumptions as RRTMG under the hood, and the code itself is more efficient. I think the only downside is the current configuration will run slightly slower I think due to the fact that the absorption coefficient data has twice the spectral resolution of the RRTMG version we have, but this is something that Robert Pincus and I are planning to work on once I finish the GPU implementation of RRTMGP in E3SM-MMF. If this is a dealbreaker for v2 we can take that on sooner, but it's only a small penalty at the moment. Also, RRTMGP is the only configuration that  works with the aquaplanent and RCE compsets that @whannah1 recently implemented.

whannah1:
@brhillman aqua planet does not require any specific rad package.

Adding my 2 cents, I'm also very surprised that RRTMG is being used for v2. Makes me think there's a disconnect between the developers and the decision makers...

golaz:
Water Cycle has had many discussions about what should be included in v2, and RRTMG has not been mentioned. I remember it being discussed as a possibility during a E3SM meeting maybe 18 or 24 months ago, but I don't recall of any follow up to that.

brhillman:
@golaz maybe that's my fault for not being a strong enough advocate for that. Regarding including RRTMGP in v2, the main advantage is just that it's a more maintainable codebase and it's rigorously tested, but I don't think it changes answers substantially so I don't think anyone has really pushed for it's inclusion on science merits. But, the code is in master, and ready to go, and I suppose we should talk more about this offline and not dominate the PR thread. My concern here (related to this PR) is just the divergent development efforts.

cauchyhk:
Hi everyone, as the PI who led the team to develop these schemes, I just want to make a couple of points here
1. Not all scattering treatments are equally efficient and accurate. Our implementation of a hybrid 2S/4S scheme is based on extensive comparison summarized in a paper published last year by our team (led by co-I Ping Yang's group) https://www.sciencedirect.com/science/article/pii/S0022407319305606
2. The surface emissivity feature was included in the original RRTM code, not a new feature in the RRTMGP. All we did is to provide the surface emissivity file (which itself is not a trivial task, Huang et al. JAS 2016) and, what's more important, to ensure no addition or reduction of LW energy when upward LW flux is passed from surface modules to the atmospheric module. In other words, to recalculate skin-temperature in a way consistent with the spectral emissivity data and with the broadband LW flux passed from the surface module (appendix in Huang et al., J Climate, 2018). This is not something readily available in the RRTMGP.  
With these being said, if there is a V2 version available with RRTMGP, we will be happy to play with that code and put our modifications in. My estimate is about 1-2 weeks that we can get this done.

rljacob:
@cauchyhk , we will be moving to RRTMGP.  In E3SM, it is located in components/cam/src/physics/rrtmgp/external.  Note that this is a git submodule pulled directly from the develop branch of https://github.com/RobertPincus/rte-rrtmgp so ideally your changes would go there first and we would then update the version we are using.

wlin7:
The implementation currently has restart non-BFB issue when all the new features are on.  It is under investigation.

rljacob:
Notes from IG telecon:  restart still being worked on.

cauchyhk:
> Notes from IG telecon: restart still being worked on.

Hi Robert, 

      I believe my group has solved the non BFB issue for the restart. My postdoc Xianwen Jing has communicated it to Wuyin. Please let me know if we need to forward that email to you.

rljacob:
update from IG telecon:  this will still be integrated for RRTMG support.

cauchyhk:
@rljacob Thanks for the update. Is there anything that we can do at Michigan's side to help this integration? Just let us know.

rljacob:
@wlin7 is this ready?

wlin7:
@rljacob , The codes can go in with no effect on existing tests. But the way reading the prescribed surface spectral emissivity is not suitable for production run. I am working on that. I would prefer waiting for that update. Meanwhile, we look to include  the relevant implementation for RRTMGP in this PR.

rljacob:
I think the RRTMG and RRTMGP changes should be separate PRs since they touch different parts of the code.  Is there a reason to combine in to one PR?

rljacob:
Also please update the PR description.  You don't need a file list because that is provided by github and can be found by git commands.

wlin7:
No, Rob, the changes for RRTMG and RRTMGP do not have to combine. The  changes in this PR outside of RRTMG will work for both. BTW, I will finalize the PR description when it is about ready to be merged.

rljacob:
note from telecon:  this is being updated for I/O and code structure.

rljacob:
@wlin7 what is the status of this PR?

wlin7:
Not finished yet, @rljacob .

rljacob:
Notes from telecon:  a few more updates are coming.  I/O needs to be optimized.

rljacob:
Update from call:  testing new I/O strategy is in progress.

rljacob:

rljacob:
This won't be ready for v2.

rljacob:
update from IG telecon:  waiting on integrator

rljacob:
notes from telecon:  waiting on code changes.

wlin7:
To be updated. Closed for now.

PR #3397, opened 2020-01-15
Update tests.py to separate HOMME and HOMMEBFB into a new suite

wadeburgess:


cime/config/e3sm/tests.py
```diff
@@ -126,6 +124,14 @@
             )
         },
 
+    "e3sm_homme" : {
```
rljacob: We want to take "e3sm" out of the name.  Call it "homme_integration"

PR #3402, opened 2020-01-16
Fix chunk cost calc. to be thread safe and add decomposition diagnostics

### LINKED ISSUE #3395: chunk cost calculation incorrect when threading in ATM
The variable irtc_rate was mistakenly left out of the private variable declarations in the OMP directives implementing threading over chunks in three locations in physpkg.F90. This leads to errors, confirmed on both Cori-KNL and Compy. This error is only obvious when the user_nl_cam option

     phys_chnk_cost_write = .true.

is set, but it is occurring in every threaded run (just not causing any other problems, that I am aware of).

### LINKED ISSUE #3401: shr_sys_irtc is not thread safe
As identified in issue #3395, shr_sys_irtc is not thread safe. I do not know if it is intended to be. As we currently call this in physpkg.F90 in cam/src/physics/cam within threaded loops, we need to at least wrap the calls in critical sections.

Based on experiments on Compy, shr_sys_irtc can be made thread safe if we wrap the whole body of the routine in a critical section, including the call to system_clock. However, even if system_clock is thread safe (why woulnd't it be?) there are local save variables that are updated within the routine, and so are subject to race conditions when called within a threaded region, so most of the body would need to be in a critical section in any case.

worleyph:
a) Modify chunk cost calculation logic to make it thread safe.

Errors occur in the logic calculating the cost per chunk when threading
over the chunk loops in physpkg.F90, confirmed on both Cori-KNL and
Compy. This error is only obvious when the user_nl_cam option

     phys_chnk_cost_write = .true.

is set, but it is occurring in every threaded run (just not causing
any other problems, that we are aware of).

There are two sources of the error (both of which must be corrected to
eliminate the problem). First, the CIME utility routine shr_sys_irtc,
used to capture the wallclock time, is not thread safe. Second, a
parameter in the shr_sys_irtc call is set within the function and
so needs to be added to the list of PRIVATE variables for each of the
3 relevant OpenMP parallelized loops.

To make shr_sys_irtc thread safe for this use case requires only
adding a threadprivate declaration for two local save variables. As
this does not make shr_sys_irtc thread safe in all possible scenarios,
a comment is added describing use cases for which the routine is
thread safe, and use cases for which it is not. These modifications
are part of this PR, so as to eliminate what was heretofore
unrecognized failure case.

shr_sys_irtc uses the fortran command system_clock in its
implementation. Upon examining the context of the use of shr_sys_irtc
to calculate the chunk cost, it was decided that it was just as
convenient to use system_clock directly, declaring a few additional
variables and including them in the list of PRIVATE variables for the
associated OpenMP parallelized loops. This also eliminates any
ambiguity about thread safety.

b) Add atmospheric physics decomposition diagnostics

In recent investigations into the performance impact on atmospheric
physics of changing the PCOLS setting, it was determined that output
of some physics decomposition specifics was very useful, i.e.,

     - the total number of mesh columns used in the atmospheric physics (physics columns),
     - the total number of chunks,
     - the number of processes active during the physics (same as ATM processes currently)
     - the minimum and maximum number of threads available per physics processes,
     - the minimum and maximum number of physics columns per chunk,
     - the minimum and maximum number of chunks per physics process, and
     - the minimum and maximum number of physics columns per process.

As optimizing over PCOLS will likely become more common in the future,
this output is added at the end of phys_grid_init. The format of the
immediately preceding phys_grid_init output is also modified, to be
consistent with the new output (and to be easier to read).

Fixes #3395
Fixes #3401

[BFB]

### DIFF
--- a/cime/src/share/util/shr_sys_mod.F90
+++ b/cime/src/share/util/shr_sys_mod.F90
@@ -213,8 +213,10 @@ integer(SHR_KIND_I8) FUNCTION shr_sys_irtc( rate )
    integer(SHR_KIND_IN)      :: count
    integer(SHR_KIND_IN)      :: count_rate
    integer(SHR_KIND_IN)      :: count_max
+
    integer(SHR_KIND_IN),save :: last_count   = -1
    integer(SHR_KIND_I8),save :: count_offset =  0
+!$OMP THREADPRIVATE (last_count, count_offset)
 
    !----- formats -----
    character(*),parameter :: subName =   '(shr_sys_irtc) '
--- a/cime/src/share/util/shr_sys_mod.F90
+++ b/cime/src/share/util/shr_sys_mod.F90
@@ -224,6 +224,14 @@ integer(SHR_KIND_I8) FUNCTION shr_sys_irtc( rate )
 
 !-------------------------------------------------------------------------------
 ! emulates Cray/SGI irtc function (returns clock tick since last reboot)
+!
+! This function is not intended to measure elapsed time between
+! multi-threaded regions with different numbers of threads. However,
+! use of the threadprivate declaration does guarantee accurate
+! measurement per thread within a single multi-threaded region as
+! long as the number of threads is not changed dynamically during
+! execution within the multi-threaded region.
+!
 !-------------------------------------------------------------------------------
 
    call system_clock(count=count,count_rate=count_rate, count_max=count_max)
--- a/components/cam/src/physics/cam/phys_grid.F90
+++ b/components/cam/src/physics/cam/phys_grid.F90
@@ -271,6 +271,12 @@ module phys_grid
    integer, private :: nproc_busy_d    ! number of processes active during the dynamics
                                        !  (assigned a dynamics block)
 
+! Physics grid decomposition environment
+   integer, private :: nlthreads       ! number of OpenMP threads available to this process
+   integer, dimension(:), allocatable, private :: npthreads
+                                       ! number of OpenMP threads available to each process
+                                       !  (deallocated at end of phys_grid_init)
+
 ! Physics grid decomposition options:  
 ! -1: each chunk is a dynamics block
 !  0: chunk definitions and assignments do not require interprocess comm.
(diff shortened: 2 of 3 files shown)


worleyph:
@ndkeen , the information generated by the chunk cost calcuation and the output of the decompostion specifics might be of interest to you, which is why I asked you to review this.

worleyph:
@wlin7 , I asked you to be the integrator since @singhbalwinder is on vacation. As this is a real bug (though seemingly innocuous) I'd like to get it in soon.

worleyph:
I've done my testing on Compy - was able to reproduce the issue there (first saw it on Cori-KNL) and also work out the fix. I'll try Cori-KNL later today or tomorrow.

worleyph:
@wlin7 , it case it is not obvious, please do not proceed with this PR yet :-) .

worleyph:
Tested the latest versions of the fixes (adding threadprivate declaration in shr_sys_irtc in shr_sys_mod.F90; adding irtc_rate to private variables declarations three places in physpkg.F90) on Cori-KNL using master. Both were required to eliminate the errors in computing the chunk costs.

worleyph:
(put this comment in the wrong place originally - repeating here)

@rljacob - what should I do? Should I submit a CIME-only PR with the addition of the threadprivate declaration to shr_sys_irtc, and then back out the critical section additions in this PR? Both PRs would then need to go in in order eliminate the bug.

rljacob:
Go ahead and do one PR here.  It will go back to CIME on the next subtree split.

worleyph:
@ndkeen , I changed the format and added some more information to the physics decomposition diagnostics. Please take a look at the latest.

worleyph:
@amametjanov , I added you as a reviewer to make sure that I implemented your suggestion correctly (adding threadprivate declaration to shr_sys_irtc function).

worleyph:
@amametjanov , I'm rethinking this again ... adding threadprivate to shr_sys_irtc is a good thing, and I think that we should still make that change. However, looking at how I am using shr_sys_irtc in physpkg.F90, it is almost as simple to just call system_clock directly. And then there is no ambiguity about what is going on under the covers (and no need for save variables as everything is local to each iteration of the loop and thread-sensitive variables can be added to the PRIVATE declaration). I'll make this change, and would appreciate it if you would then have a look. I can always back this out if you find it less clear than using shr_sys_irtc.

worleyph:
@wlin7 , this PR is ready for integration. Thanks.

wlin7:
Thanks  for the reminder, @worleyph . Merged to next.

wlin7:
Merged to master.

PR #3404, opened 2020-01-16
Start using new NERSC project directory named CFS

ndkeen:
NERSC now has a new CFS filesystem and has copied the data for us, but we now need to point to it.

Changing references from /project/projectdirs/acme to /global/cfs/cdirs/acme

Data in /project is now read-only, so tests should still work.

[bfb]


ndkeen:
tests are completing, merged to next

ndkeen:
Fixes #3407

PR #3406, opened 2020-01-18
Homme V2 integration updates

ambrad:
* Call prim_finalize from EAM, and call Kokkos::finalize when using SL transport (indirectly) from prim_finalize.
* Final tool for physgrid topography file.
* Fix reading of GLL PHIS_d when reading the new physgrid topo file format.

[BFB]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@
 *.pyc
 buildnmlc
 buildlibc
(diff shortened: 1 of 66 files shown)


ambrad:
e3sm_developer passes on sandiatoss3.

PR #3411, opened 2020-01-24
Fix vertical remap algorithm 10 in homme, PPM with linextrap boundaries.

ambrad:
Previously, minval and maxval were over all entries of ao, but they should be
over `ao(1:nlev)`.

[non-BFB] for 2 standalone Homme tests and SMS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L.sandiatoss3_intel.cam-thetahy_sl_pg2

### DIFF
--- a/components/homme/src/share/vertremap_base.F90
+++ b/components/homme/src/share/vertremap_base.F90
@@ -622,8 +622,8 @@ subroutine remap_Q_ppm(Qdp,nx,qsize,dp1,dp2)
               endif
             enddo
          elseif (vert_remap_q_alg == 10) then
-            ext(1) = minval(ao)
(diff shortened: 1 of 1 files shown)


ambrad:
Standalone on SKX with AVX512. Expected fails:
```
   21 - thetah-sl-test11conv-r1t2-cdr20 (Failed)
   22 - thetah-sl-test11conv-r0t1-cdr30-rrm (Failed)
```
I checked the output and the failures are due to the expected and reasonable numerical diffs.

ambrad:
e3sm_developer passed on sandiatoss3.

oksanaguba:
in next

PR #3412, opened 2020-01-24
Replace current cosp-1.4 with cosp2

wlin7:
This PR implements cosp2 and then replace existing cosps with cosp2.

The cosp directory components/cam/src/physics/cosp is removed. cosp2 modules are placed under components/cam/src/physics/cosp2. The cosp interface module, cam/src/physics/cam/cospsimulator_intr.F90 is also replaced. The data structure of the new interface module differs from the previous version. The subroutine names and argument list remain unchanged. No change in upstream caller routines is involved.

[BFB] for tests not using cosp.

Note: MODIS simulator is cosp2 has an issue and is waiting to be fixed. One option is to turn MODIS simulator off until it is fixed.


rljacob:
A bunch of .svn files got checked in.  The branch will have to be redone.

wlin7:
@rljacob , thanks for the quick check. Reopen it just to post this message. I updated the branch after  removing .svn/ and pushed to remote. However, all the .svn files still remain in the files changed for this PR. I didn't know this behavior. Does that mean a new PR needed after a change like this in a branch? Thanks.

rljacob:
Right.  Remove a file doesn't remove its history of having once been in the branch.  The branch has to be completely redone.

wlin7:
Ok, will do a new PR. The number of "files changed" does change from 72 to 43 after some delay, and the .svn files no more shown up in the list. Previously refreshing several times, and all were there.

PR #3413, opened 2020-01-24
Replace current cosp-1.4 with cosp2

wlin7:
This PR implements cosp2 and then replaces existing cosp (v1.4) with cosp2.

The cosp directory components/cam/src/physics/cosp is removed. cosp2 modules are placed under components/cam/src/physics/cosp2. The cosp interface module, cam/src/physics/cam/cospsimulator_intr.F90 is also replaced. The data structure of the new interface module differs from the previous version. The subroutine names and argument list remain unchanged. No change in upstream caller routines is involved.

This is to replace PR #3412 which was previously created but contained unwanted .svn files.

[BFB] for tests not using cosp.

Note: MODIS simulator in cosp2 has an issue and is waiting to be fixed. One option is to turn MODIS simulator off until it is fixed.


rljacob:
The problem is the branch still has commit 13a499d9e which has the .svn files committed.   That's why I said the branch needs to be redone.  You need to remove that commit.

rljacob:
Really,  Yuying Zhang needs to redo the branch without commiting .svn files.

wlin7:
Done, @rljacob , with a new branch YuyingZhang/atm/cosp2_new based off current master, using the final state of YuyingZhang/atm/cosp2. The PR is now #3415. Thanks. I previously only re-created the PR after git rm .svn files. That makes no difference on the branch history. The previous branch will be deleted after the new PR #3415 is reviewed by @zyuying .

PR #3414, opened 2020-01-24
ne30pg2_r05_oECv3 resolution configuration for v2 AMIP simulations.

ambrad:
config_grids.xml and namelist_defaults_cam.xml changes to support the
ne30pg2_r05_oECv3 resolution for v2 AMIP runs with physgrid.

A future PR may modify some map files for coupled simulations.

[BFB]


ambrad:
e3sm_developer passed on sandiatoss3. No diffs, as expected.

PR #3415, opened 2020-01-26
Implement cosp2 and replace cosp-1.

wlin7:
This PR implements cosp2 and then replaces existing cosp (v1.4) with cosp2.

The cosp directory components/cam/src/physics/cosp is removed. cosp2 modules
are placed under components/cam/src/physics/cosp2. The cosp interface module,
cam/src/physics/cam/cospsimulator_intr.F90 is also replaced. The data structure
of the new interface module differs from the previous version. The subroutine names
and argument list remain unchanged. No change in upstream caller routines is involved.

[BFB] for tests not using cosp


golaz:
@wlin7, thanks for working on this.

Could we use the occasion of upgrading COSP from 1.4 to 2 to also make COSP work with RRTMGP? 

@brhillman: are you still planning to work on enabling COSP with RRTMGP? If so, it should be done with COSP2.

wlin7:
Note: this is to replace PR #3413. MODIS simulator in cosp2 has an issue and is waiting to be fixed. One option is to turn MODIS simulator off until it is fixed.

wlin7:
Yes, @golaz, I will make sure of it. I heard from @zyuying (and @brhillman indirectly) that using RRTMGP with cosp1, Ben always had to do some adhoc changes.  Ben, can you please chime in?

brhillman:
I was working on enabling COSP1 with RRTMGP, which requires some changes to the RRTMGP interface to pass the right variables in and out. I will wait for this PR and then base that work on COSP2. To be clear, COSP1 with RRTMGP does not require adhoc changes, but COSP1 with MMF does require adhoc changes. Moving to COSP2 will be quite welcome, as it will make that effort a lot easier as well. So, my vote is to merge this PR, and then I will work on the RRTMGP interface for COSP2 once that’s in. Does that sound reasonable?

brhillman:
Can I also suggest that rather than copy all of the COSP2 code directly into our repo, that we instead make COSP2 a submodule? We already do this for a number of components (RRTMGP is an example) and since COSP2 exists on github, it would allow us to keep pace with COSP2 development as things change there. Otherwise we are responsible for manually making bugfixes as we find them ourselves.

rljacob:
Where is the COSP2 github master?

Yes we should make it a submodule unless there is some reason we need a highly customized version.

brhillman:
https://github.com/CFMIP/COSPv2.0

rljacob:
So following the pattern from rrtmpg, one would create cam/src/physics/cosp2/external and put the entire cosp2 repo there as a submodule.  Then add any additional needed code to cam/src/physics/cosp2/   What do you think @wlin7 ?

wlin7:
Summarized a discussion with @zyuying on adopting the submodule approach for cosp2 below.

1. Both this PR and the COSP2 on github include codes that are subject to modifications when used in individual models. These files will be locally managed as part of E3SM repository.
2. Only those under the src/ directory of https://github.com/CFMIP/COSPv2.0 will be imported as external submodule (using the file structure outlined above by @rljacob)
3. Supporting cosp Fortran codes currently under cam/src/physics/cosp2 will continue to be managed as local files, including all the files under optics/ and subcol/  subdirectories.
4. Specifically for files under optics/ and subcol, they allow to be modified depending on what processes are considered in the simulators. For example, E3SM implementation includes snow particle s in MODIS simulator while the standard release of COSP2 does not.
5. Going forward, E3SM implementation may include diagnostics not available in the standard COSP2 release (master branch of COSP2) -- COSP2 will manage its own release schedule. In that case, subset of files in cosp2 submodule may be converted to locally managed to suit the need of E3SM diagnostics. This will be managed via modification to Cosp.cmake.

rljacob:
git submodule doesn't let you take pieces of a repo.  So you can't include just the src/ directory of https://github.com/CFMIP/COSPv2.0 .  It has to be the whole repo.  The EAM build can ignore everything but the src/ directory.

wlin7:
This PR is replaced by PR #3458 to use cosp2 as a submodule.

PR #3418, opened 2020-01-28
Sarich/add gen domain test

sarich:
Add tool testing framework (at this point only gen_domain test) to CIME scripts_regression_tests.py

### DIFF
--- a/cime/scripts/tests/scripts_regression_tests.py
+++ b/cime/scripts/tests/scripts_regression_tests.py
@@ -152,6 +152,17 @@ class A_RunUnitTests(unittest.TestCase):
                         run_cmd_assert_result(self, 'PYTHONPATH={}:$PYTHONPATH python -m doctest {} 2>&1'.format(LIB_DIR, filepath), from_dir=LIB_DIR)
                     else:
                         print("{} has no doctests".format(filepath))
+###############################################################################
+class C_TestGridGeneration(unittest.TestCase):
+###############################################################################
+    def test_gen_domain(self):
+        inputdata = MACHINE.get_value("DIN_LOC_ROOT")
+        cime_root = get_cime_root()
+        testname = os.path.join(cime_root, "tools", "mapping", 
+                                "gen_domain_files", "test_gen_domain.sh")
+        temp_dir = tempfile.mkdtemp()
+        run_cmd_assert_result(self, "{} --inputdata_root={} --cime_root={}".format(testname, inputdata, cime_root), from_dir=temp_dir)
+        shutil.rmtree(temp_dir)
 
 ###############################################################################
 def make_fake_teststatus(path, testname, status, phase):
--- a/cime/scripts/tests/scripts_regression_tests.py
+++ b/cime/scripts/tests/scripts_regression_tests.py
@@ -152,17 +152,47 @@ class A_RunUnitTests(unittest.TestCase):
                         run_cmd_assert_result(self, 'PYTHONPATH={}:$PYTHONPATH python -m doctest {} 2>&1'.format(LIB_DIR, filepath), from_dir=LIB_DIR)
                     else:
                         print("{} has no doctests".format(filepath))
+
 ###############################################################################
 class C_TestGridGeneration(unittest.TestCase):
 ###############################################################################
+    @classmethod
+    def setUpClass(cls):
+        cls._do_teardown = []
+        cls._testroot = os.path.join(TEST_ROOT, "TestGridGeneration")
+        cls._testdirs = []
+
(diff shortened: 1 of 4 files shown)


sarich:
It works on compy and anlworkstation. You should be able to run it with

./cime/scripts/tests/scripts_regression_tests.py C_TestGridGeneration

 (which calls up the bash script cime/tools/mapping/gen_domain_files/test_gen_domain.sh)
Any new tool tests can then be added in scripts_regression_tests.py as a new member function in C_TestGridGeneration class.

The tools look like they have some machine/compiler/arch issues that won't run on every system right now.

cime/scripts/tests/scripts_regression_tests.py
```diff
@@ -152,6 +152,17 @@ def test_lib_doctests(self):
...
                         print("{} has no doctests".format(filepath))
+###############################################################################
+class C_TestGridGeneration(unittest.TestCase):
+###############################################################################
+    def test_gen_domain(self):
+        inputdata = MACHINE.get_value("DIN_LOC_ROOT")
+        cime_root = get_cime_root()
+        testname = os.path.join(cime_root, "tools", "mapping", 
+                                "gen_domain_files", "test_gen_domain.sh")
+        temp_dir = tempfile.mkdtemp()
```
jgfouca: Minor point: generally unittests should not do their cleanup directly in the test. It should be done in a teardown routine.

cime/scripts/tests/scripts_regression_tests.py
```diff
@@ -152,6 +152,17 @@ def test_lib_doctests(self):
                         run_cmd_assert_result(self, 'PYTHONPATH={}:$PYTHONPATH python -m doctest {} 2>&1'.format(LIB_DIR, filepath), from_dir=LIB_DIR)
                     else:
                         print("{} has no doctests".format(filepath))
+###############################################################################
```
jgfouca: Nit: please put a newline between the end of a function and new class declaration.

cime/scripts/tests/scripts_regression_tests.py
```diff
@@ -152,6 +152,17 @@ def test_lib_doctests(self):
...
+class C_TestGridGeneration(unittest.TestCase):
+###############################################################################
+    def test_gen_domain(self):
+        inputdata = MACHINE.get_value("DIN_LOC_ROOT")
```
jgfouca: Oh, almost forgot; You should skip this test if the model != e3sm. We don't want to make CESM/UCF folks wait for a test of a tool they don't care about.

PeterCaldwell:
I'm surprised test_gen_domain.sh always builds cprnc. Shouldn't it check if cprnc already exists? Similarly (but probably a negligible time saver), should it check whether it can use an existing env_mach_specific.sh rather than always creating a new one?

PeterCaldwell:
Beyond this comment, I rely on @brhillman and @jgfouca 's determination of whether this is what we want. Thanks for your work on this, @sarich !

cime/tools/mapping/gen_domain_files/test_gen_domain.sh
```diff
@@ -0,0 +1,134 @@
...
+fi
+
+# Setup the test case. gen_domain takes as inputs a mapping file and the names
+# of the ocean and land grids. Note that we could probably get these by parsing
+# the specified baseline filenames or global netcdf attributes, so that we
+# could maybe simplify this to specifying just ocn_baseline and lnd_baseline,
+# and then figuring out which mapping file and grid names we need.
+ocn_name=oQU240
+lnd_name=ne4np4
+mapping_file=${inputdata_root}/cpl/gridmaps/oQU240/map_oQU240_to_ne4np4_aave.160614.nc
```
brhillman: So, I hard-coded these into the original script for demonstration purposes. Curious what @sarich and @jgfouca have to say about this, for this tool and others that we want to add. I am happy having this just as a minimal test of functionality (i.e., the code builds and runs and gives right answers for at least one case), but I think ideally we would want to test a few different grids and make sure we are getting the right answers (different combinations of grids test the code in different ways). Any input on this? I don't have a good answer.

rljacob: Someone will have to figure out what combinations provide the best coverage.  That can be added in a future PR.

PeterCaldwell: Totally agreed. Let's keep this PR simple and choose other grids in a future PR.

jgfouca:
Totally forgot about this.

PR #3424, opened 2020-01-31
Added modex machine config files

serbinsh:
PR to add the machine-specific configuration options to the e3sm config files for building and running on the modex compute cluster at BNL.

### DIFF
--- a/cime/config/e3sm/machines/config_batch.xml
+++ b/cime/config/e3sm/machines/config_batch.xml
@@ -576,4 +576,15 @@
     </queues>
    </batch_system>
 
+  <!-- modex is PBS -->
+  <batch_system MACH="modex" type="pbs">
+    <directives>
+      <directive>-l nodes={{ num_nodes }}:ppn={{ tasks_per_node }}</directive>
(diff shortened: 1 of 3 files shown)


serbinsh:
The equivalent files are CESM/CIME.  Is there where I would place a PR like this or should I be doing this somewhere else?

Summary: modex is a 100 core, multi-node cluster that I manage and use in the BNL TEST group (https://www.bnl.gov/envsci/testgroup/) so this doesn't impact any other researcher at any other institution.

rljacob:
@serbinsh this is the right place for adding E3SM specific machines.   Just edit the description above to remove the last 2 sentences.  The description of the PR should be exactly the text you want in the git log.

serbinsh:
OK @rljacob I updated the description above....let me know if you would like any changes or more information.

rljacob:
Is this machine for just your own use?

serbinsh:
@rljacob Its for our groups use so not just mean but me, my postdocs and students, as well as external collaborators whom use the modex cluster.  This is a cluster located at BNL

wlin7:
Merged to next.

wlin7:
Merged to master.

PR #3429, opened 2020-02-04
Upgrade the software stack for anlworkstation

dqwu:
This PR upgrades some modules used by anlworkstation.

gcc is upgraded to 8.2.0 (from 6.2.0) via softenv.

mpich/szip/hdf5/netcdf/pnetcdf are built with gcc 8.2.0 (parallel
libs are built with mpich 3.3.2). Currently, these new modules are
not supported yet in softenv.

mpich is upgraded to 3.3.2 (from 3.2).

pnetcdf is upgraded to 1.12.0 (from 1.8.1).

[BFB]

### DIFF
--- a/cime/config/e3sm/machines/config_machines.xml
+++ b/cime/config/e3sm/machines/config_machines.xml
@@ -711,31 +711,28 @@
         <command name="add">+cmake-3.12.4</command>
       </modules>
       <modules compiler="gnu">
-        <command name="add">+gcc-6.2.0</command>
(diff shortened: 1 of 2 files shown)


dqwu:
@jayeshkrishna 
I am not authorized to push to next branch, could you please do that? Thanks.

dqwu:
@jayeshkrishna 
Nightly build on e3sm_developer tests looks good. It seems that PR #3262 is causing DIFFs for 2 ECA tests, which is expected, since the PR updates dynamic allocation and stoichiometry features.

PR #3433, opened 2020-02-06
Define hybrid BGC cases and create bgc production test suite

rljacob:
Define 4 existing BGC compsets as hybrid so they can run out-of-the-box with the correct restart files.
Compsets changed are:
BGCEXP_BCRC_CNPECACNT_20TRS and BGCEXP_CNTL_CNPECACNT_1850S
BGCEXP_CNTL_CNPRDCTC_1850 and BGCEXP_BCRD_CNPRDCTC_20TR

Also add testmod directories for each of the above so the namelist settings will be the same as production runs.

Also add an e3sm_bgcprod test suite which can be used to do 2-day smoke tests of all 4 cases.

[non-BFB] only for the compsets mentioned above and only because the initial condition changes.


rljacob:
Currently this only works on compy because I haven't moved the data files to the server yet.

rljacob:
The restart files needed are now on the data server.

rljacob:
To run the new test suite, check out this branch and do:   ./create_test e3sm_bgcprod

rljacob:
@qzhu-lbl @susburrows @kvcalvin if there are no objections, I'll merge this on Friday.

susburrows:
Thanks Rob, I just quickly looked it over but these changes look like no-brainers.  We definitely want these compsets working out-of-the-box with the settings and restart files from the production runs, and under testing.

PR #3437, opened 2020-02-07
cam: fix bug in bld/configure

jgfouca:
Get caseroot via cli argument instead of from environment.

Also a minor cleanup of cmake handling of cosp lib.

[BFB]


singhbalwinder:
Thanks Jim for fixing it. It looks good to me.

jgfouca:
@singhbalwinder good, feel free to integrate.

singhbalwinder:
pushed to next

PR #3439, opened 2020-02-10
Disable calls to shr_taskmap_write when INFO_TASKMAP_COMP is zero

### LINKED ISSUE #3438: shr_taskmap_write called unnecessarily when INFO_TASKMAP_COMP is zero
If INFO_TASKMAP_COMP in env_run.xml is zero (the default), then the mapping
of MPI tasks to cores is not output, but it is still calculated for
CAM, CLM, Mosart, MPAS-ALBANY-LANDICE, MPAS-OCEAN, and MPAS-SEAICE. For
large task counts, this can be expensive, and it is not necessary for
any of these components except for CAM. For example, these unnecessary calls to
shr_taskmap add approximately one minute to the cost of initialization for a high 
resolution A_WCYCL case using 64000 processes for each component (except CAM)
on Cori-KNL (approximately 15 seconds for each unnecessary call: CLM, Mosart, MPAS-OCEAN, MPAS-SEAICE).

worleyph:
If INFO_TASKMAP_COMP in env_run.xml is zero (the default), then the mapping
of MPI tasks to cores is not output, but it is still calculated, for
CAM, CLM, Mosart, MPAS-ALBANY-LANDICE, MPAS-OCEAN, and MPAS-SEAICE. For
large task counts, this can be expensive, and it is not necessary for
any of these components except for CAM.  For example, these unnecessary calls to
shr_taskmap_write add approximately one minute to the cost of initialization for a high
resolution A_WCYCL case using 64000 processes for each component (except CAM)
on Cori-KNL (approximately 15 seconds for each unnecessary call: CLM, Mosart, 
MPAS-OCEAN, MPAS-SEAICE). Here this is addressed, and
shr_taskmap_write is no longer called when INFO_TASKMAP_COMP is zero
for any of the above mentioned component models except for CAM. 

Fixes #3438 

[BFB]

### DIFF
--- a/components/clm/src/cpl/lnd_comp_mct.F90
+++ b/components/clm/src/cpl/lnd_comp_mct.F90
@@ -85,7 +85,6 @@ contains
     integer  :: dtime_sync                           ! coupling time-step from the input synchronization clock
     integer  :: dtime_clm                            ! clm time-step
     logical  :: exists                               ! true if file exists
-    logical  :: no_taskmap_output                    ! true then do not write out task-to-node mapping
     logical  :: verbose_taskmap_output               ! true then use verbose task-to-node mapping format
     logical  :: atm_aero                             ! Flag if aerosol data sent from atm model
     logical  :: atm_present                          ! Flag if atmosphere model present
@@ -164,8 +163,6 @@ contains
 
     if (info_taskmap_comp > 0) then
 
-       no_taskmap_output = .false.
-
        if (info_taskmap_comp == 1) then
           verbose_taskmap_output = .false.
        else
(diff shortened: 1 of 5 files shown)


worleyph:
Tested with 

      -compset A_WCYCL1950S_CMIP6_LR -res ne30_oECv3_ICG

on Cori-KNL with a 6 node (384 MPI tasks) PE layout, both with

         <entry id="INFO_TASKMAP_COMP" value="0">

and

         <entry id="INFO_TASKMAP_COMP" value="1">

and it behaved as expected, with no shr_taskmap_write timing events for the components for which shr_taskmap_write is supposed to be disabled.

worleyph:
@rljacob , I assigned this to you becuase you handled the other taskmap PRs. Please reassign if there is a better choice. Changes for each component (clm, mosart, mpas-ocean, mpas-seaice, mpas-albany-landice) are identical.

rljacob:
merged to next

PR #3440, opened 2020-02-11
Remove darshan module and switch to CFS for Cori

ndkeen:
Make changes for maint-1.0 already on master.
Only affects cori-knl and cori-haswell.
Remove the darshan module which is not needed.
Switch to using the CFS filesystem instead of /project.
Remove temporary PERL environment settings.

[bfb]


ndkeen:
The maint-1.0 tests ran OK.

ndkeen:
merged

PR #3448, opened 2020-02-14
Adds output of topography in land history file

### LINKED ISSUE #3447: Different values of topo in clm2.h0 with different version of PIO
Reported by @jayeshkrishna:
>Hi, I am trying to do some baseline comparison tests with SMS_Ln9.ne4_ne4.FC5AV1C-L.anvil_intel.cam-outfrq9s and I see that while generating the baselines with scorpio classic the variable topo(lndgrid) in clm.h0.* has 0 values (all values in the variable are 0s). However with scorpio (PIO_VERSION=2) when comparing the result with the baselines I see that the model generates non-zero values (0s, 7s and 6.*s) . Have you seen this behavior before? What is the expected value for this variable?

bishtgautam:
The missing call to output topography in clm2.h0 files is added

Fixes #3447
[non-BFB]

### DIFF
--- a/components/clm/src/main/histFileMod.F90
+++ b/components/clm/src/main/histFileMod.F90
@@ -2667,6 +2667,7 @@ contains
           call ncd_io(varname='lon', data=ldomain%lonc, dim1name=grlnd, ncid=nfid(t), flag='write')
           call ncd_io(varname='lat', data=ldomain%latc, dim1name=grlnd, ncid=nfid(t), flag='write')
        end if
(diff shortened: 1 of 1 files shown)


jayeshkrishna:
Is this PR BFB for all tests (the PR writes out the contents of the variable)?

bishtgautam:
I guess different compilers initialize a variable in the netcdf files differently when the variable is not written out. I will mark it as non-bfb.

rljacob:
But why only 3 tests?  Not every I-case test?

jqyin:
Merged #3448

PR #3454, opened 2020-02-18
Fixes initialization bug in CLUBBv2 interface

### LINKED ISSUE #3457: Some variables are not initialized in CLUBB v2
Some new fields are added to CLUBBv2 which were not present in the older version of CLUBB. These variables( wpthvp, wp2thvp, rtpthvp, thlpthvp, rcm and cloud_frac) are used by advance_clubb_core() subroutine but were not initialized in  CLUBBv2 interface.

zhangshixuan1987:
The newly added fields wpthvp, wp2thvp, rtpthvp, thlpthvp, rcm, cloud_frac used by advance_clubb_core() subroutine were not initialized. This PR fixes this bug.

These uninitialized variables will not cause problems for the default setup of clubbv2 (results stays BFB), while it will produce issues if alternative options (namelists) added into CLUBBv2 are activated.

Fixes #3457
[BFB]

### DIFF
--- a/components/cam/src/physics/cam/clubb_intr.F90
+++ b/components/cam/src/physics/cam/clubb_intr.F90
@@ -178,6 +178,14 @@ module clubb_intr
     qrl_idx, &          ! longwave cooling rate
     radf_idx 
 
+ integer :: &          ! newly added pbuf fields for CLUBB
+    wpthvp_idx, &       ! < w'th_v' >
+    wp2thvp_idx, &      ! < w'^2 th_v' >
+    rtpthvp_idx, &      ! < r_t'th_v' >
+    thlpthvp_idx, &     ! < th_l'th_v' >
+    rcm_idx, &          ! Cloud water mixing ratio
+    cloud_frac_idx !, & ! Cloud fraction
+
   integer :: &          !PMA adds pbuf fields for ZM gustiness
     prec_dp_idx, &
     snow_dp_idx, &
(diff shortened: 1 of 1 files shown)


rljacob:

Please fix the title of this PR.  It should start with a verb like "Fix".

singhbalwinder:
I have fixed the PR title and description. I will add a GitHub issue for this fix.

wlin7:
@zhangshixuan1987 , can you explain with what namelist options, these changes are going to matter?

All the code changes are good by themselves. I am just want to make sure if it is necessary to add the several fields to the pbuf. Take the example of the variable wp2thvp. The initialization and passing via pbuf seems to only be needed when ipdf_call_placement  == ipdf_post_advance_fields.

In advance_clubb_core, ipdf_call_placement is hardwired to ipdf_pre_call_placement, which means wp2thvp will always be computed before being used in call to advance_wp2_wp3 (and other ensuing calls). In that case, neither initialization nor passing via pbuf is needed. 

Thanks.

zhangshixuan1987:
Hi Wuyin, 

Thank you very much. We have a conversation with @vlarson before we made those changes, Vince pointed out that _"CLUBB now has two call orderings: i) In E3SMv1's default order, PDF closure is called before the moments are advanced.  ii) In a second option, PDF closure is called after the moments are advanced.  This second option is one that DOE will probably want to try, because it should mitigate the pockets of spurious supersaturation that are left behind with ordering i)._  

There is probably not a problem when ordering i) is used, because then those variables are assigned values before they're used in the PDF closure. I believe that this is also the scenario as you described above.  However, the changes are going to matter when ordering ii) is used.  The corresponding variables (wpthvp, rcm, etc.) should be initialized with a pbuf_set_field call in clubb_ini_cam as the variables (e.g. wpthvp) would bee needed by CLUBB to advance its moment equations. 

Thanks again,
 
Shixuan

wlin7:
Thanks Shixuan for the explanation. I heard from Balwinder there is a follow-up PR  soon. It make more senses to me now.

singhbalwinder:
pushed to next

PR #3456, opened 2020-02-19
Rework NLDAS to better match existing implementation

jonbob:
This PR changes the scripts to get rid of the NLDAS configurations, which offered potential conflicts and confusion with NLDAS2 configurations already available in CIME. The NLDAS2 dataset also covers a longer time period and is more complete. The compsets change from:

- INLDASCNPECACNTBC
  (2000_DATM%NLDAS_CLM45%CNPECACNTBC_SICE_SOCN_MOSART_SGLC_SWAV)
- IM20TRNLDASCNPECACNTBC
  (20TR_DATM%NLDAS_CLM45%CNPECACNTBC_SICE_SOCN_MOSART_SGLC_SWAV)

to:

- INLDAS2CNPECACNTBC
  (2000_DATM%NLDAS2_CLM45%CNPECACNTBC_SICE_SOCN_MOSART_SGLC_SWAV)
- I2003NLDAS2CNPECACNTBC
  (2003_DATM%NLDAS2_CLM45%CNPECACNTBC_SICE_SOCN_MOSART_SGLC_SWAV)
- I2010NLDAS2CNPECACNTBC
  (2010_DATM%NLDAS2_CLM45%CNPECACNTBC_SICE_SOCN_MOSART_SGLC_SWAV)

The MOSART-only compset RMOSARTNLDAS keeps its same name and forcing file. NOTE: this compset is currently broken and is under analysis.

### DIFF
--- a/cime/src/components/data_comps/datm/cime_config/config_component.xml
+++ b/cime/src/components/data_comps/datm/cime_config/config_component.xml
@@ -10,13 +10,12 @@
        This file may have atm desc entries.
   -->
   <description modifier_mode="1">
-    <desc atm="DATM[%QIA][%WISOQIA][%CRU][%CRUv7][%GSWP3v1][%NLDAS][%NLDAS2][%CPLHIST][%1PT][%NYF][%IAF][%JRA]"> Data driven ATM </desc>
+    <desc atm="DATM[%QIA][%WISOQIA][%CRU][%CRUv7][%GSWP3v1][%NLDAS2][%CPLHIST][%1PT][%NYF][%IAF][%JRA]"> Data driven ATM </desc>
     <desc option="QIA"> QIAN data set </desc>
     <desc option="WISOQIA">QIAN with water isotopes</desc>
     <desc option="CRU"> CRUNCEP data set </desc>
     <desc option="CRUv7"> CLM CRU NCEP v7 data set </desc>
     <desc option="GSWP3v1"> GSWP3v1 data set </desc>
-    <desc option="NLDAS"> NLDAS data set </desc>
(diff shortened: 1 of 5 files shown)


jonbob:
@hydrotian - do you mean the error in the RMOSARTNLDAS configuration? Yes, I'm still working on that. This PR just gets rid of having the two different versions of NLDAS.

hydrotian:
@jonbob I see. Thanks for clarifying it.

jonbob:
I added the "do not merge" tag. It's failing testing of SMS.NLDAS_NLDAS.INLDAS2CNPECACNTBC.anvil_intel for reasons I don't understand yet

jonbob:
@hydrotian - I have an update on the RMOSART compset that is failing. There are a couple of issues -- one with the dlnd data stream and another with the compset itself. I've figured out the dlnd problem, but the compset change will take more time. We merged this PR before the mosart heat PR -- which now requires cpl fields from the atm. This compset is defined to use the stub atm, which does not pass all those required fields to the cpl. So we'll have to modify the compset definition to use the data atm, and then work through whatever issues arise from that.

jonbob:
After testing, none of the new NLDAS2 compsets actually work within E3SM. Two of the I-cases fail in CLM, with errors in lnd2atm where the code checks for temperatures greater than 400 K as well as in BalanceCheckMod. The other I-case fails in datm. The new RMOSARTNLDAS is no longer valid after the mosart-heat integration because MOSART now requires fields from the atm that are not available in the stub atm model. I will close this PR and replace it with one that renames the NLDAS compsets to avoid confusion with NLDAS2.

PR #3465, opened 2020-02-24
Add MMF configuration

whannah1:
This adds a Multi-scale Model Framework (i.e. MMF or "super-parameterization") configuration to the atmosphere model, with which the effects of convection and turbulence are treated by coupling each physics column of the global atmosphere model with an instance of a cloud resolving model (CRM). This implementation was manually ported over from a separate fork originally created for the Exascale Computing Project (ECP). 

This implementation differs from the ECP fork in a few ways, but the most notable difference is that the MMF is implemented as a separate physics package. This isolation of the two physics packages is intentional to prevent conflicts and confusion. Most files in the **src/physics/cam** folder that required changes are duplicated in the **src/physics/crm** folder, and the build system makes sure these files are used when the "-use_MMF" configure flag is present. 

The current branch was directly compared to the ECP fork and found to be BFB on a range of tests with and without the MMF.

Note that we agreed to break the "MAML" option in this initial implementation, but this will be repaired in a future PR. 

[BFB]


rljacob:
What kind of testing did you do?   This should ideally have no impact on current atmosphere cases right?

rljacob:
@singhbalwinder and @wlin7 your review doesn't have to determine if MMF is implemented correctly.  Just that the rest of EAM is not broken.

whannah1:
@rljacob that's correct. The non-MMF cases will be unaffected.
I was mostly running SMS tests and comparing against baselines from both the ECP and E3SM repos. I also ran our ECP test suite numerous times:
```
ERP_Ld11.ne4_oQU240.A_WCYCL2000
ERP_Ld5.ne4_ne4.FC5AV1C-L
ERP_Ld5.ne30_ne30.FC5AV1C-L
ERS_Ld5.ne30_ne30.FC5AV1C-L

ERP_Ld3_P96.ne4_ne4.FSP1V1-TEST
ERS_Ld3_P96.ne4_ne4.FSP1V1-TEST.cam-crmout
ERP_Ld3_P96.ne4_ne4.FSP2V1-TEST
ERP_Ld3_P96.ne4_ne4.FSP2V1-ECPP-TEST
SMS_D_Ln5_P96.ne4_ne4.FSP1V1-TEST
SMS_D_Ln5_P96.ne4_ne4.FSP2V1-TEST
```

whannah1:
After Ben and address some minor things I can run the entire developer/integration suite.

whannah1:
FYI - the current to do list we have is:
- verify that MMF w/ RRTMGP is still BFB with ECP/master
- Make RRTMGP default rad option for MMF cases
- Update names of MMF compsets (mostly to replace "SP" with "MMF")
- update MMF test suite to use shorter tests
- change "crm" folder to "mmf"? (not sure about this)

Those are all pretty small/simple changes, which shouldn't affect the review

whannah1:
We struggled a bit to get the test suite passing on Cori, and we still don't fully understand what the problem was, but all tests in the MMF test suite pass now. I'll run a developer suite next, but I think the branch is ready to get reviewed.

whannah1:
e3sm_atm_developer tests all pass

whannah1:
@singhbalwinder and @wlin7 , when you review this you might notice some changes in the normal RRTMGP directory, like the file src/physics/rrtmgp/slingo.F90. These changes had to be made to get RRTMGP workin for the MMF. They are also coming in through another PR, which we assume will get merged before this one. So you should be able to ignore those, but I'll make sure to revisit this after you are done with the review.

Also, note that you don't need to look at anything in the "src/physics/crm" folder (unless you want to) as nothing in there will affect non-MMF configurations.

components/cam/bld/config_files/definition.xml
```diff
@@ -315,5 +315,47 @@ that setting to allow for cross-compilation, and for instances where the
 $OSNAME value is too generic.  For example, currently on both cray-xt and
 bluegene systems $OSNAME has the value "linux".
 </entry>
+<entry id="crm" valid_values="0,1" value="0">
```
brhillman: `crm` and `use_MMF` flags appear to be duplicates...do we need both?

components/cam/bld/configure
```diff
@@ -266,6 +266,21 @@ OPTIONS
                         perl \$OSNAME variable.
      -usr_src <dir1>[,<dir2>[,<dir3>[...]]]
                         Directories containing user source code.
+
+  Options for MMF CAM:
```
brhillman: I'd lose the "CAM" here. Replace with "EAM".

components/cam/bld/configure
```diff
@@ -1278,6 +1305,38 @@ EOF
...
 if ($print>=2) { print "Produce IOP file for SCAM: $e3smreplay$eol"; }
 
+#-----------------------------------------------------------------------------------------------
+# Superparameterization mode (MMF)
+
+if (defined $opts{'use_MMF'}) {
+    if (defined $opts{'use_ECPP'})            { $cfg_ref->set('use_ECPP', 1); }
+    if (defined $opts{'use_MAML'})            { $cfg_ref->set('use_MAML', 1); }
+    $cfg_ref->set('use_MMF', 1);
+    $cfg_ref->set('crm', 1);
```
brhillman: Is the `crm` flag necessary?

components/cam/bld/config_files/definition.xml
whannah1: I think I meant to delete this. I'm pretty sure I removed the crm variable from the configure code.

components/cam/bld/configure
whannah1: will do

components/cam/bld/config_files/definition.xml
whannah1: Looks like I just forgot to delete that after I deleted the spots where it was used, so that's an easy fix.

components/cam/bld/namelist_files/use_cases/1850_cam5_av1c-04p2-MMF-1mom.xml
```diff
@@ -0,0 +1,102 @@
+<?xml version="1.0"?>
```
brhillman: We aren't actually using the 1850 use cases, are we?

whannah1: not at the moment, but it was an easy thing to add, should I delete it?

components/cam/bld/namelist_files/use_cases/20TR_cam5_av1c-04p2-MMF-1mom.xml
```diff
@@ -0,0 +1,107 @@
+<?xml version="1.0"?>
```
brhillman: The 20TR use cases are also unused and untested I believe.

whannah1: Aren't they used for AMIP runs?

components/cam/cime_config/config_component.xml
```diff
@@ -96,6 +96,38 @@
       <value compset="_CAM[45]%WCBC"  >-carma bc_strat</value>
       <value compset="_CAM4%WCSF"     >-chem waccm_mozart_sulfur -carma sulfate</value>
       <value compset="_CAM[45]%WCMX"  >-waccmx</value>
+      <!-- old MMF / super-parameterization compsets -->
```
brhillman: NOTE: these are in here right now only to maintain testing against the ECP fork. These will go away in a future PR.

components/cam/bld/namelist_files/use_cases/20TR_cam5_av1c-04p2-MMF-1mom.xml
whannah1: I mean, I know they aren't used by us currently

components/cam/cime_config/config_component.xml
whannah1: I'll add a note in the file

components/cam/cime_config/config_component.xml
```diff
@@ -250,11 +282,24 @@
...
+      <value compset="2000_CAM5%MMF1"    >2000_cam5_av1c-04p2-MMF-1mom</value>
+      <value compset="2000_CAM5%MMF2"    >2000_cam5_av1c-04p2-MMF-2mom</value>
+      <value compset="20TR_CAM5%MMF1"    >20TR_cam5_av1c-04p2-MMF-1mom</value>
+      <value compset="20TR_CAM5%MMF2"    >20TR_cam5_av1c-04p2-MMF-2mom</value>
     </values>
     <group>run_component_cam</group>
     <file>env_run.xml</file>
     <desc>CAM namelist use_case.  Provides the use_case option for the
-      CAM build-namelist utility (which is called from
+      CAM build-namelist utility (which its called from
```
brhillman: Typo

components/cam/cime_config/config_component.xml
```diff
@@ -250,11 +282,24 @@
       <value compset="_CAM5%AQUA"     >aquaplanet_EAMv1</value>
       <!-- Radiative-Convective Equilibrium (RCE) -->
       <value compset="_CAM5%RCE"      >RCEMIP_EAMv1</value>
+      <!-- MMF / Super-Parameterization -->
+      <value compset="1850_CAM5%SP1"     >1850_cam5_av1c-04p2-MMF-1mom</value>
```
brhillman: I think the 1850 and 20TR lines should go away. They do not have corresponding compset definitions in config_compsets.xml, and we are not using or testing them.

components/cam/cime_config/config_component.xml
whannah1: I don't remember changing that, but I'll fix it

components/cam/cime_config/config_component.xml
whannah1: ok

components/cam/bld/namelist_files/use_cases/1850_cam5_av1c-04p2-MMF-1mom.xml
whannah1: removed

components/cam/bld/namelist_files/use_cases/20TR_cam5_av1c-04p2-MMF-1mom.xml
whannah1: removed

components/cam/src/physics/cam/flux_avg.F90
```diff
@@ -206,12 +210,12 @@ subroutine smooth(new, old, res, nstep, deltat, ncol)
    ! but if it is large we will add it at the rate required to put
    ! the residual back into the flux over a 2 hour period
```
brhillman: Note that the default is a 4-hour period now, not a 2-hour period.

components/cam/src/physics/cam/flux_avg.F90
```diff
@@ -180,8 +181,11 @@ subroutine smooth(new, old, res, nstep, deltat, ncol)
    integer,  intent(in)    :: ncol
 
    real(r8) :: temp(pcols)
+   real(r8) :: maxres
    integer i
 
+   real(r8), parameter :: timescale = 14400._r8    ! 4 hours
```
brhillman: Should this be a namelist parameter?

whannah1: yea, but we don't need this for pg2, so it's not a priority. I would argue we should do this in a small future PR.

components/cam/src/physics/cam/flux_avg.F90
whannah1: correct, but this does not affect non-MMF E3SM because it is only used by the MMF.

components/cam/src/physics/cam/phys_control.F90
```diff
@@ -343,6 +360,14 @@ subroutine phys_ctl_readnl(nlfile)
       end if
    end if
 
+   ! Check settings for MMF_microphysics_scheme
+   if ( .not.(MMF_microphysics_scheme .eq. 'm2005'   .or. &
+              MMF_microphysics_scheme .eq. 'sam1mom' .or. &
+              MMF_microphysics_scheme .eq. unset_str )) then
```
brhillman: Should we encapsulate this in a `if (use_MMF)` and then pull out the check against `unset_str` here? If it's still `unset_str`, then I think we failed to set the microphysics scheme, and we should abort right here.

whannah1: sure.

whannah1: this check actually says unset_str is a valid value, which I guess is why we didn't need the "use_MMF". So I think I can just remove the check for unset_str.

components/cam/src/utils/error_function.F90
```diff
@@ -0,0 +1,646 @@
+module error_function
```
brhillman: What's up with this module? This is a brand new file? Where did this come from?

components/cam/src/physics/cam/flux_avg.F90
brhillman: Right I mean the comment here is incorrect now.

components/cam/src/utils/error_function.F90
whannah1: It's used by the CRM 2-mom micro scheme. Since we aren't going to use the 2-mom code we don't really need it... but until we have P3 in the CRM we need the 2-mom code to work on refactoring the ECPP code, so I think we should keep it and try to remember to remove this module later.

components/cam/src/physics/cam/flux_avg.F90
brhillman: Right, just noting.

components/cam/src/physics/cam/flux_avg.F90
whannah1: oh, sorry, misunderstood, will fix.

rljacob:
@whannah1 is this ready?

whannah1:
@rljacob, still waiting on reviews from @wlin7 and @singhbalwinder.
Ben needed to fix the conflicting rad files (to bring in changes from another PR) and found a minor bug in the configure file, but that shouldn't hold up their review. Ben already has a fix in place and is working on making sure the tests still pass. So we mostly just need a review from someone on the E3SM, specifically to look at the files I touched outside of the cam/src/physics/crm directory.

components/cam/bld/configure
```diff
@@ -1828,6 +1886,36 @@ $cfg_cppdefs .= " -DMAXPATCH_PFT=numpft+1 -DLSMLAT=1 -DLSMLON=1";
 # Building for perturbation growth tests
 if ($pergro eq "ON") { $cfg_cppdefs .= " -DPERGRO"; }
 
+# Building for MMF/super-parameterization
```
wlin7: I think this block had better be conditioned on if (use_MMF eq "ON"), at least for the addition to cfg_cppdefs.

whannah1: yea, Ben caught that last week. He has a fix being tested right now. The MMF code was erroneously being built for non-MMF cases, but amazingly the non-MMF tests were still BFB.

components/cam/bld/configure
```diff
@@ -1278,6 +1305,37 @@ EOF
...
+**  ERROR: MMF mode may only use single thread
+**         nthreads set to $nthreads
+**         
+EOF
+    }
+
+   # print "---> $opts{'crm_nx'} $opts{'crm_ny'} $opts{'crm_nz'} $opts{'crm_dx'} $opts{'crm_dt'} $opts{'MMF_microphysics_scheme'} $eol";
+}
+
+my $use_MMF = $cfg_ref->get('use_MMF') ? "ON" : "OFF";
```
wlin7: Why using "ON" or "OFF" for  use_MMF? 0 or 1 would be better. Plus, when $use_MMF="OFF", if ($use_MMF), which is used in the block start line 2660,  would still be true, causing unintended writing to src filepath.

wlin7: I missed something when not looking at the full file. use_MMF defined here would not be visible inside sub write_Filepath. In that case, use_MMF seems to be undefined in sub write_fileppath.

components/cam/bld/configure
wlin7: Huh, I am late to the game. Interesting behavior,

components/cam/bld/configure
whannah1: Ben is fixing this too.

brhillman:
@wlin7 I think all your concerns should be address in the rebase I just pushed.

brhillman:
Rebase on current (as of a week ago) master and fixes for a major refactor of RRTMGP cloud optics that went into master since this PR was opened, so this should be easier to merge now. All of these changes were BFB with the state of this branch before the rebase was done. Note that `configure` has been cleaned up quite a bit, fixing some erroneous logic that would *always* build the MMF version of the code, even if the model was not going to use MMF.

rljacob:
IG telecon notes:  still getting tested by watercycle.

whannah1:
@rljacob I don't understand your comment. 
What is still being tested?

rljacob:
The testing  @wlin7 and @singhbalwinder are doing for their review.

components/cam/src/physics/crm/sam/crm_output_module.F90
```diff
@@ -0,0 +1,581 @@
...
+      real(crm_rknd), allocatable :: prectend     (:)    ! column integrated tend in precip water+ice [kg/m2/s]
+      real(crm_rknd), allocatable :: precstend    (:)    ! column integrated tend in precip ice       [kg/m2/s]
+      real(crm_rknd), allocatable :: taux     (:)    ! zonal CRM surface stress perturbation      [N/m2]
+      real(crm_rknd), allocatable :: tauy     (:)    ! merid CRM surface stress perturbation      [N/m2]
+      real(crm_rknd), allocatable :: z0m          (:)    ! surface stress                             [N/m2]
+      real(crm_rknd), allocatable :: timing_factor(:)    ! crm cpu efficiency
+
+#ifdef MAML
+      ! MAML variables
+      real(crm_rknd), allocatable :: crm_pcp(ncrms,crm_nx,crm_ny) ! CRM precip rate for MAML (m/s)
```
wlin7, line 136: Didn't intend to review this module.  Created a test case anyway with compset F-MAM2-MAML, and picked up these coding mistake: crm_pcp and crm_snw, declaraed as allocatable, should give just shape, no actual sizes for the dimensions.

wlin7:
While there appears to be an issue with MMF2-MAML, it is certain all things for default E3SM build and run ok.

whannah1:
@wlin7 what issue did you notice for MMf2-MAML? 
That shouldn't in the test suite and none of us are planning to use that combination right now, so whatever the problem is, it should be ok.

wlin7:
I agreed, @whannah1 . While the bug is obvious to fail the compilation, if not using MAML, it is fine for now. My test case uses F-MMF2-MAML compset.

wlin7:
Sorry, @whannah1 . I should have made it  more explicit. The declarations at lines 136 and 137 are not appropriate.

```
cam/src/physics/crm/sam/crm_output_module.F90(136): error #6646: ALLOCATABLE or POINTER attribute dictates a deferred-shape-array.   [CRM_PCP]
      real(crm_rknd), allocatable :: crm_pcp(ncrms,crm_nx,crm_ny) ! CRM precip rate for 1 (m/s)

```

wlin7:
There are some more compilation errors if I want to push ahead with  the build with MAML enabled. If the development and testing for MAML is still on going, I think the PR is ready to go as is for now.

whannah1:
MAML is definitely still under development. I wanted to include some of the previous work here, because it will make the porting of MAML changes from the ECP fork much easier. We had previously agreed that it would be ok to temporarily break MAML for this PR because we want to re-engineer how it's implemented in E3SM to make it more flexible and robust.

wlin7:
Sounds good to me. I happened to pick a compset that I thought including everything for a test. I have approved the PR at the current stage.

components/cam/src/chemistry/modal_aero/aero_model.F90
```diff
@@ -2919,6 +2952,12 @@ subroutine modal_aero_depvel_part( ncol, t, pmid, ram1, fv, vlc_dry, vlc_trb, vl
                   gravit*slp_crc(i,k) / vsc_dyn_atm(i,k) ![m s-1] Stokes' settling velocity SeP97 p. 466
           vlc_grv(i,k) = vlc_grv(i,k) * dispersion
 
+          ! in the MMF NaN's were occurring here but the root cause was not
+          ! identified, so this check was added to work around the issue
+          if (use_MMF) then
+            if ( ieee_is_nan(vlc_grv(i,k)) ) vlc_grv(i,k) = 0.0_r8
```
singhbalwinder, line 2958: If there isn't a GitHub issue regarding this, we should probably open an issue to look at it. I think it is important to know where these NaNs are coming from.

components/cam/src/physics/crm/RNG_MT.F90
```diff
@@ -0,0 +1,142 @@
+! Mersenne Twister Uniform Random Number Generator ((0,1) Real Number)
```
singhbalwinder, line 1: There are some Random number generator algorithms under `cime/src/share/RandNum/`. I haven't carefully read this code but an RNG similar to this is at `cime/src/share/RandNum/src/mt19937/mersennetwister_mod.F90`. Unless there is a strong need to add a new one, I think we should use the existing RNGs.

components/cam/src/utils/error_function.F90
singhbalwinder: This code looks very similar to `cime/src/share/util/shr_spfn_mod.F90`. Is this a newer version of the existing code?

components/cam/src/physics/crm/RNG_MT.F90
whannah1: I initially planned to use an exiting RNG, but for this case I needed to get a single number at a time and ensure that the sequence was reproducible for each column. I remember concluding that the existing RNG routines were well equipped for this, so I just made a simpler implementation of the MT algorithm to handle this. 

In either case, this is only used for a special feature where the CRM orientation slowly rotates using a biased random walk in angular space. The idea was to get away from the fixed CRM orientation, which gives different systematic biases when comparing N/S to E/W. 

However, after testing this option I realized that the real problem is how the CRMs line up with the wind field, which causes cloud systems to "cycle" through the domain. There's also changes in model behavior when the CRM lines up with the shear. This is why the original Grabowski implementation constantly updated the CRM orientation based on the low-level wind. The N/S orientation minimizes the occurrence of having the CRM line up with the prevailing flow compared to the E/W orientation. The randomly changing orientation ends up representing the worst of both worlds by making the CRM line up with the prevailing flow everywhere at least once a day, which gives a mix of the worst systematic biases in both E/W and N/S options, rather than reducing them like I had hoped.

I'm still actively thinking about this problem, so I'd like to keep this code in place to revisit it, but it's not regularly used for the MMF. I'd like to revisit this in the next year or so and decide whether to keep it or not. If we decide to make this on by default then we can switch to using an existing RNG library.

components/cam/src/chemistry/modal_aero/aero_model.F90
whannah1: I spent weeks tracing this down a couple years ago and got nowhere. I'll need to try taking this out and seeing if I can reproduce the problem. I forget the details, but I assume this was from the 2-mom case, which none of us are running now. And since we're planning to rip out the current 2-mom scheme in favor of the new P3 from SCREAM, maybe we should just take this out? 

The only potential problem I see with taking it out is that it might start breaking our 2-mom tests, and that could hinder our efforts to refactor ECPP. 

I'm not sure what to do here, curious to hear what others think.  
@brhillman @pressel @crjones-amath @lee1046

brhillman: @singhbalwinder I agree that there should be a Github issue for this if it's still a problem. However, my preference would be to leave this code in for the purpose of this PR in case it's still needed, and since the issue is only relevant to certain configurations of the MMF, then it doesn't become an issue in E3SM until this PR is merged anyways. So I think it makes the most sense to leave this in for now, merge the PR, and then confirm that removing this code still results in NaNs for the configurations that @whannah1 found before and then open an issue.

components/cam/src/physics/crm/RNG_MT.F90
brhillman: @singhbalwinder I agree, but I think we should defer these changes until after this PR is merged.

components/cam/src/chemistry/modal_aero/aero_model.F90
singhbalwinder: Thanks. This check is within `use_MMF` if condition so it doesn't affect the default E3SM model run. If this block is needed for the time being, it can be left in the code.

components/cam/src/physics/crm/RNG_MT.F90
singhbalwinder: Sounds good!

rljacob:
Merged to next.

PR #3466, opened 2020-02-24
Upgrade KernelVariables team-idx calculation

jgfouca:
And have CaarFunction use this upgraded capability.

This seems to pass caar_ut_test on waterman, although the state of Hommexx on waterman is a bit flaky. It took many hours of fiddling (with help from Luca) to get a waterman.cmake that worked. Even now, some of the ut_tests seem to fail randomly.

I wanted to put this up for a review before I convert the other functors.


components/homme/cmake/machineFiles/waterman.cmake
```diff
@@ -1,44 +1,47 @@
...
-set (ENABLE_HORIZ_OPENMP CACHE BOOL OFF)
-
-set (USE_NUM_PROCS 8 CACHE STRING "")
-
-
-
-
+# TPL settings
+set(NETCDF_DIR $ENV{NETCDF_ROOT} CACHE FILEPATH "")
+set(NetCDF_Fortran_PATH /ascldap/users/lbertag/workdir/libs/netcdf/netcdf-f/netcdf-f-install/waterman/gcc CACHE FILEPATH "")
```
bartgol: I'm surprised that oksana's netcdf installation did not work. At some point, however, we should request the waterman support to have a netcdf-fortran module (even if all in the same folder, like sems does).

components/homme/src/share/cxx/scream_kokkos_utils.hpp
```diff
@@ -0,0 +1,189 @@
+#ifndef SCREAM_KOKKOS_UTILS_HPP
+#define SCREAM_KOKKOS_UTILS_HPP
+
+#include "Types.hpp"
+#include "ExecSpaceDefs.hpp"
+
+#include "../../algorithms/src/Kokkos_Random.hpp"
```
bartgol: This is misleading at best. I think it works cause it switches to a system include, and uses the search path `${KOKKOS_SRC}/core/src`. I think the best way to handle this is to modify the include directory in `cmake/Kokkos.cmake`, and add also `${KOKKOS_SRC}/algorithms/src`.

components/homme/src/share/cxx/KernelVariables.hpp
```diff
@@ -8,6 +8,7 @@
 #define KERNEL_VARIABLES_HPP
 
 #include "Types.hpp"
+#include "scream_kokkos_utils.hpp"
```
bartgol: Can the file be renamed simply `kokkos_utils.hpp`? There is no 'scream' in homme, and it helps keeping things separated. If/when scream gets into e3sm, we can think of a place where to put this utils that can be shared across different components (perhaps `components/share`).

components/homme/src/share/cxx/scream_kokkos_utils.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+#include <chrono>
+#include <cassert>
+#include <type_traits>
+
+// This file should not be merged with share/scream_kokkos_meta.hpp.
```
bartgol: This comment should probably be deleted, cause it refers to a possible misunderstanding of files in the scream repo. Here, we don't have those files, so this is confusing at best.

components/homme/src/share/cxx/scream_kokkos_utils.hpp
```diff
@@ -0,0 +1,189 @@
...
+  { return 0; }
+
+  template <typename MemberType>
+  KOKKOS_INLINE_FUNCTION
+  void release_workspace_idx(const MemberType& /*team_member*/, int /*ws_idx*/) const
+  { }
+};
+
+template <typename ExeSpace = Kokkos::DefaultExecutionSpace>
+class TeamUtils : public _TeamUtilsCommonBase<ExeSpace>
```
bartgol: Design choice, feel free to push back: do we need the inheritance pattern here? There's no extension of the base class, nor virtual functions (because Cuda). It looks to me as this class should simply be a template class specialized later on, without a base class (there's close to 0 code reuse anyways).

Second thought: can we leave only the template class in this hpp file, and relegate the specializations to a cpp file, with ETI? That way we would keep the interface (that one needs to be aware of) in a small tidy file, rather than a long one. Or would that require relocatable-device-code? (I admit that I'm a noob when it comes to understand rdc mechanism).

components/homme/src/theta-l_kokkos/cxx/CaarFunctorImpl.hpp
```diff
@@ -105,6 +106,8 @@ struct CaarFunctorImpl {
   Kokkos::RangePolicy<ExecSpace, TagPostExchange> m_policy_post;
   Kokkos::TeamPolicy<ExecSpace, TagDp3dLimiter>   m_policy_dp3d_lim;
 
+  mutable TeamUtils<ExecSpace> m_tu;
```
bartgol: Why mutable?

components/homme/src/share/cxx/KernelVariables.hpp
```diff
@@ -107,6 +128,7 @@ struct KernelVariables {
 
   int ie, iq;
   const int team_idx;
+  TeamUtils<ExecSpace>* team_utils;
```
bartgol: Do you really need to store team_utils inside the kv object? It seems to me that it is superfluous.

components/homme/src/share/cxx/scream_kokkos_utils.hpp
bartgol: I think the "second thought" would require rdc, so you can disregard that comment.

The thought about simple specialization vs inheritance is the only one to focus here.

ambrad:
Looking good. I won't approve yet, of course, since there's a ways to go, including performance testing. Exciting!

rljacob:
What does "Caar" stand for here?

bartgol:
@rljacob CAAR = Compute And Apply Rhs

The hommexx code opted for a short name, which, admittedly, makes little sense unless you know the f90 code routine that it came from.

Perhaps at some point we should rename it to something more verbose, like ExRKStageFunctor or something along that line.

components/homme/cmake/machineFiles/waterman.cmake
jgfouca: I think that may have been a permissions issue on her side.

components/homme/src/share/cxx/KernelVariables.hpp
jgfouca: Yes.

components/homme/src/share/cxx/scream_kokkos_utils.hpp
jgfouca: Done.

components/homme/src/share/cxx/scream_kokkos_utils.hpp
jgfouca: I'm tempted to keep it since we are getting some code re-use in the shared constructor and in some of the queries.

components/homme/src/theta-l_kokkos/cxx/CaarFunctorImpl.hpp
jgfouca: Because the functor operator() are const and getting a team-idx is non-const (has to change TeamUtils internals top keep track of state).

components/homme/src/share/cxx/KernelVariables.hpp
jgfouca: I think we do if we want to keep the current approach. I wanted to integrate TeamUtils with KernelVariables since there's overlap in functionality and to minimize changes to functors. In particular, leveraging the KernelVariables destructor to release workspace indices would not be possible without this member.

components/homme/src/share/cxx/scream_kokkos_utils.hpp
jgfouca: I don't like this either, but I'm not eager to dive into Homme/Kokkos cmake integration.

components/homme/src/theta-l_kokkos/cxx/CaarFunctorImpl.hpp
bartgol: I don't see where you would violate const. All methods in TeamUtils are const.

jgfouca: Good catch. It was just that the KernelVariables member was non-const and needed to be changed to const.

components/homme/src/share/cxx/scream_kokkos_utils.hpp
bartgol: I can take care of that, if you want.

jgfouca:
@bartgol , @ambrad , I think I've converted everything that needs to be converted (any maybe a few things that's didn't really need to be).

I'll begin more-serious waterman testing tomorrow.

jgfouca:
I think this will be the final implementation. Starting performance testing now.

jgfouca:
prim_main_loop

jgfouca:
caar_compute

jgfouca:
caar_benchV

jgfouca:
caar_dp3d

jgfouca:
compute_stage_value_dirk

jgfouca:
hvf-bhwk

jgfouca:
hvf-bexch

jgfouca:
tl-at-prim_advec

jgfouca:
tl-sc-vert-remap

jgfouca:
@ambrad , @bartgol , results posted. I made a plot for each timer that Andrew recommended. All of the vertical axes are mislabelled, they are in seconds, not columns per sec. Results look great, timers are either near an exact match or we get some speed ups.

ambrad:
Looks great!
* Cols/sec would actually be extremely useful here. I'd like to see a slightly negative slope (ie, increased efficiency with increasing ne), possibly plateauing.
* Is this a performance build, ie, with optimizations?
* Would you post the output of the grep line I gave you for the ne18 case?
Thanks!

jgfouca:
These were all made with the following cmake (Release build) settings:
```
SRC_DIR=..
MY_CXX_FLAGS="-Xcudafe --diag_suppress=esa_on_defaulted_function_ignored -Wno-unknown-pragmas"
#NETCDF_F_ROOT=~lbertag/workdir/libs/netcdf/netcdf-f/netcdf-f-install/waterman/gcc                                                                                                                                                                
NETCDF_F_ROOT=~jgfouca/packages/netcdf-fortran-4.5.2-gcc-7.2.0-openmpi-4.0.1-install
rm -rf CMakeFiles
rm -f  CMakeCache.txt
cmake -Wno-dev \
  -D CMAKE_BUILD_TYPE:STRING=Release                            \
  -D CMAKE_VERBOSE_MAKEFILE:BOOL=OFF                            \
  -D CMAKE_CXX_COMPILER:STRING=mpicxx                           \
  -D CMAKE_C_COMPILER:STRING=mpicc                              \
  -D CMAKE_Fortran_COMPILER:STRING=mpif90                       \
  -D CMAKE_C_FLAGS:STRING="-w"                                  \
  -D CMAKE_CXX_FLAGS:STRING="${MY_CXX_FLAGS}"                   \
  -D CMAKE_Fortran_FLAGS:STRING="-w"                            \
  -D CMAKE_EXE_LINKER_FLAGS:STRING=-ldl                         \
  \
  -D OPT_FLAGS:STRING="-O3"                                     \
  -D DEBUG_FLAGS:STRING="-ffp-contract=off"                     \
  \
  -D BUILD_HOMME_SWEQX:BOOL=OFF                                 \
  -D BUILD_HOMME_THETA:BOOL=OFF                                 \
  -D BUILD_HOMME_PREQX:BOOL=OFF                                 \
  -D BUILD_HOMME_THETA_KOKKOS:BOOL=ON                           \
  -D BUILD_HOMME_PREQX_KOKKOS:BOOL=OFF                          \
  -D HOMMEXX_GB_CONFIG:BOOL=ON                                  \
  -D HOMME_BASELINE_DIR:PATH=${BASELINE_DIR}                    \
  -D USE_QUEUING:BOOL=FALSE                                     \
  -D USE_NUM_PROCS:STRING=1                                     \
  -D USE_MPI_OPTIONS:STRING="--bind-to core"                    \
  -D ENABLE_OPENMP:BOOL=FALSE                                   \
  -D HOMME_ENABLE_COMPOSE:BOOL=OFF                              \
  \
  -D HOMMEXX_EXEC_SPACE:STRING=CUDA                             \
  -D HOMMEXX_BFB_TESTING:BOOL=OFF                               \
  -D AVX_VERSION:STRING=0                                       \
  -D HOMMEXX_VECTOR_SIZE:STRING=1                               \
  \
  -D NETCDF_DIR:PATH=${NETCDF_F_ROOT}                           \
  -D WITH_PNETCDF:BOOL=FALSE                                    \
  -D HOMME_FIND_BLASLAPACK:BOOL=ON                              \
  \
  -D Kokkos_ENABLE_OPENMP:BOOL=OFF                              \
  -D Kokkos_ENABLE_CUDA:BOOL=ON                                 \
  -D Kokkos_ENABLE_CUDA_LAMBDA:BOOL=ON                          \
  -D Kokkos_ENABLE_EXPLICIT_INSTANTIATION:BOOL=OFF              \
  -D Kokkos_ARCH_VOLTA70=ON                                     \
  \
  ${SRC_DIR}
```

jgfouca:
```
bash-4.2$ grep "PAT_remap\|caar\|hvf\|vertical\|RK2\|dirk\|prim_main" HommeTime_stats_with_chg_ne18
"prim_main_loop"                      -          1        1 1.000000e+00   1.112486e+00     1.112 (     0      0)     1.112 (     0      0)
"caar compute"                        -          1        1 4.000000e+01   1.713911e-01     0.171 (     0      0)     0.171 (     0      0)
"caar_bexchV"                         -          1        1 2.000000e+01   4.227252e-02     0.042 (     0      0)     0.042 (     0      0)
"caar dp3d"                           -          1        1 2.000000e+01   4.800139e-03     0.005 (     0      0)     0.005 (     0      0)
"compute_stage_value_dirk"            -          1        1 2.000000e+01   1.383421e-01     0.138 (     0      0)     0.138 (     0      0)
"hvf-bhwk"                            -          1        1 4.000000e+00   3.362980e-02     0.034 (     0      0)     0.034 (     0      0)
"hvf-bexch"                           -          1        1 8.000000e+00   1.326480e-02     0.013 (     0      0)     0.013 (     0      0)
"tl-at prim_advec_tracers_remap_RK2"  -          1        1 4.000000e+00   2.476308e-01     0.248 (     0      0)     0.248 (     0      0)
"tl-sc vertical_remap"                -          1        1 2.000000e+00   4.995980e-02     0.050 (     0      0)     0.050 (     0      0)
bash-4.2$ 
bash-4.2$ 
bash-4.2$ grep "PAT_remap\|caar\|hvf\|vertical\|RK2\|dirk\|prim_main" HommeTime_stats_without_chg_ne18
"prim_main_loop"                      -          1        1 1.000000e+00   1.051974e+00     1.052 (     0      0)     1.052 (     0      0)
"caar compute"                        -          1        1 4.000000e+01   1.712993e-01     0.171 (     0      0)     0.171 (     0      0)
"caar_bexchV"                         -          1        1 2.000000e+01   4.238065e-02     0.042 (     0      0)     0.042 (     0      0)
"caar dp3d"                           -          1        1 2.000000e+01   4.802662e-03     0.005 (     0      0)     0.005 (     0      0)
"compute_stage_value_dirk"            -          1        1 2.000000e+01   1.396975e-01     0.140 (     0      0)     0.140 (     0      0)
"hvf-bhwk"                            -          1        1 4.000000e+00   3.381073e-02     0.034 (     0      0)     0.034 (     0      0)
"hvf-bexch"                           -          1        1 8.000000e+00   1.328915e-02     0.013 (     0      0)     0.013 (     0      0)
"tl-at prim_advec_tracers_remap_RK2"  -          1        1 4.000000e+00   2.065989e-01     0.207 (     0      0)     0.207 (     0      0)
"tl-sc vertical_remap"                -          1        1 2.000000e+00   3.399083e-02     0.034 (     0      0)     0.034 (     0      0)
```

jgfouca:
@ambrad , @bartgol , I'm too accustomed to looking at cols/s, where higher is better. In this case, higher=took longer, worse. So we did get some slowdowns on a couple of the timers.

jgfouca:
prim_main_loop

jgfouca:
caar_compute

jgfouca:
caar_bexchV

jgfouca:
caar_dp3d

jgfouca:
dirk

jgfouca:
hvf-bhwk

jgfouca:
hvf-bexch

jgfouca:
tl-at

jgfouca:
tl-sc

jgfouca:
rk2

jgfouca:
@ambrad , @bartgol , I think this PR is in its final form unless there is a big performance surprise. Please do a final review.

jgfouca:
I further trimmed TeamUtils<GPU> when sharing is off as per request by Andrew.

jgfouca:
prim_main_loop

jgfouca:
prim_main

jgfouca:
caar_compute

jgfouca:
caar_bench

jgfouca:
caar_dp3d

jgfouca:
drik

jgfouca:
hvf_bhwk

jgfouca:
hvf_bench

jgfouca:
rk2

jgfouca:
remap

jgfouca:
Measurements were taken at ne6, ne18, and ne30 (for ws-sharing only). There is still significant damage to the RK2 timer when sharing is on. I had to turn down over-provisioning to 1.25 in order for ne30 to work, but this probably damaged performance at ne18.

bartgol:
To be clear, for rk2 and vert-remap, the ws should turn automatically off, so we should only have the cost of creating the TU, right?

jgfouca:
Yes, buffer-sharing has to be explicitly turned-on via CMake.

ambrad:
This is really nice.

PR #3471, opened 2020-02-26
fates initialization sequence bug fix

### LINKED ISSUE #3461: ERS_Ld3.f45_f45.ICLM45ED.anvil_intel.clm-fates failed with scorpio
This failure can also be reproduced on anlworkstation with Address Sanitizer.

First, subroutine initialize1 (CLM initialization first phase, in components/clm/src/main/clm_initializeMod.F90) calls set_fates_global_elements, which calls fates_history_maps.
At this time, num_elements is still 0, element_list is still null.
fates_hdim_levelem is allocated with size 0 (actually occupies 1 byte for some compilers)
```
allocate( fates_hdim_levelem(num_elements))
...
! Make an element array, each index is the PARTEH global identifier index

do iel = 1, num_elements
    fates_hdim_levelem(iel) = element_list(iel)
end do
```
Later, InitPARTEHGlobals is called by subroutine init (initializes the hlm_fates_interface_type, in components/clm/src/main/clmfates_interfaceMod.F90)
It will set num_elements and allocate for element_list
```
select case(hlm_parteh_mode)
case(prt_carbon_allom_hyp)

    num_elements = 1
    allocate(element_list(num_elements))
    element_list(1) = carbon12_element
```
The issue is that fates_history_maps is called before PARTEH globals are initialized.
After this, num_elements is set to 1, but fates_hdim_levelem buffer remains empty.

Next, in htape_timeconst, scorpio is used to put variable fates_levelem
`call ncd_io(varname='fates_levelem',data=fates_hdim_levelem, ncid=nfid(t),flag='write')`

For this ncd_io call, num_elements is 1, but size(fates_hdim_levelem) is 0
scorpio classic is OK with this, but for scorpio, PnetCDF will try to copy 4 bytes of integer data from an empty buffer (actually occupies 1 byte), and cause a segmentation fault.

Besides fates_hdim_levelem, a few other buffers that also rely on num_elements have the same issue.
```
allocate( fates_hdim_levelem(num_elements))
allocate( fates_hdim_elmap_levelpft(num_elements*numpft))
...
allocate( fates_hdim_agemap_levelage(num_elements*nlevage))
```

One workaround would be adding the following check before calling ncd_io:
```
if (num_elements > 0 .and. size(fates_hdim_levelem) > 0) then
    call ncd_io(varname='fates_levelem',data=fates_hdim_levelem, ncid=nfid(t),flag='write')
    call ncd_io(varname='fates_elmap_levelpft',data=fates_hdim_elmap_levelpft, ncid=nfid(t),flag='write')
...
    call ncd_io(varname='fates_agemap_levelage',data=fates_hdim_agemap_levelage, ncid=nfid(t),flag='write')
end if
```

However, this might avoid outputting data that is required by the model developers, so a better fix might be something else in fates.

rgknox:
This is a fix to the ordering of FATES and FATES-PARTEH intiialization. Specifically, this 
change in order ensures that the number of chemical species to be simulated in FATES is 
determined earlier in the intiialization sequence, which will allow E3SM to correctly allocate 
array spaces in its history machinery.  

Fixes #3461
[non-BFB]

### DIFF
--- a/components/clm/src/main/clmfates_interfaceMod.F90
+++ b/components/clm/src/main/clmfates_interfaceMod.F90
@@ -110,6 +110,7 @@ module CLMFatesInterfaceMod
    use FatesInterfaceMod     , only : SetFatesTime
    use FatesInterfaceMod     , only : set_fates_ctrlparms
    use FatesInterfaceMod     , only : InitPARTEHGlobals
+   use FatesInterfaceMod     , only : set_fates_global_elements
 
    use FatesHistoryInterfaceMod, only : fates_history_interface_type
    use FatesRestartInterfaceMod, only : fates_restart_interface_type
@@ -196,7 +197,6 @@ module CLMFatesInterfaceMod
       procedure, private :: init_soil_depths
       procedure, public  :: ComputeRootSoilFlux
       procedure, public  :: wrap_hydraulics_drive
-      procedure, public  :: ELMFatesGlobalElements
 
    end type hlm_fates_interface_type
 
@@ -209,6 +209,8 @@ module CLMFatesInterfaceMod
 
    character(len=*), parameter, private :: sourcefile = &
         __FILE__
+
+   public  :: ELMFatesGlobalElements
    
 contains
 
@@ -216,15 +218,35 @@ contains
 
    subroutine ELMFatesGlobalElements()
 
-       ! --------------------------------------------------------------------------------
-       ! This is one of the first calls to fates
-       ! Used for setting dimensions.  This MUST
-       ! be called after NL variables are specified and
-       ! after the FATES parameter file has been read in
-       ! Aside from setting global dimension info, which
-       ! is used in the history file, we also transfer
-       ! over the NL variables to FATES global settings.
-       ! --------------------------------------------------------------------------------
+     ! --------------------------------------------------------------------------------
+     ! This is one of the first calls to fates
+     ! Used for setting dimensions.  This MUST
+     ! be called after NL variables are specified and
+     ! after the FATES parameter file has been read in
+     ! Aside from setting global dimension info, which
+     ! is used in the history file, we also transfer
+     ! over the NL variables to FATES global settings.
+     ! --------------------------------------------------------------------------------  
+     
+     integer                                        :: pass_masterproc
+     integer                                        :: pass_vertsoilc
+     integer                                        :: pass_spitfire     
+     integer                                        :: pass_ed_st3
+     integer                                        :: pass_logging
+     integer                                        :: pass_ed_prescribed_phys
+     integer                                        :: pass_planthydro
+     integer                                        :: pass_inventory_init
+     integer                                        :: pass_is_restart
+     
+     ! --------------------------------------------------------------------------------
+     ! This is one of the first calls to fates
+     ! Used for setting dimensions.  This MUST
+     ! be called after NL variables are specified and
+     ! after the FATES parameter file has been read in
+     ! Aside from setting global dimension info, which
+     ! is used in the history file, we also transfer
+     ! over the NL variables to FATES global settings.
+     ! --------------------------------------------------------------------------------
        
        if (use_fates) then
 
@@ -379,15 +401,6 @@ contains
       ! local variables
       integer                                        :: nclumps   ! Number of threads
       logical                                        :: verbose_output
-      integer                                        :: pass_masterproc
-      integer                                        :: pass_vertsoilc
-      integer                                        :: pass_spitfire     
-      integer                                        :: pass_ed_st3
-      integer                                        :: pass_logging
-      integer                                        :: pass_ed_prescribed_phys
-      integer                                        :: pass_planthydro
-      integer                                        :: pass_inventory_init
-      integer                                        :: pass_is_restart
       integer                                        :: nc        ! thread index
       integer                                        :: s         ! FATES site index
       integer                                        :: c         ! HLM column index
--- a/components/clm/src/main/clmfates_interfaceMod.F90
+++ b/components/clm/src/main/clmfates_interfaceMod.F90
@@ -43,6 +43,7 @@ module CLMFatesInterfaceMod
 
    use SoilStateType     , only : soilstate_type 
    use clm_varctl        , only : iulog
+   use clm_varctl        , only : use_fates
    use clm_varctl        , only : use_vertsoilc 
    use clm_varctl        , only : use_fates_spitfire
    use clm_varctl        , only : fates_parteh_mode
@@ -136,7 +137,7 @@ module CLMFatesInterfaceMod
    use FatesPlantHydraulicsMod, only : UpdateH2OVeg
    use FatesPlantHydraulicsMod, only : RestartHydrStates
    use FatesInterfaceMod      , only : bc_in_type, bc_out_type
-
+   
    implicit none
    
    type, public :: f2hmap_type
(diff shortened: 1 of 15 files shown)


rgknox:
thanks to @dqwu and @bishtgautam for their help and for identifying this.

rgknox:
question: This PR wraps in some cime machine definition updates for the LBNL local cluster. Should I remove those and submit in a different PR?

bishtgautam:
@rgknox, The CIME changes shouldn't be a problem when those are contributed back to the main CIME repo (cc: @jgfouca).

rgknox:
Should I remove them from this PR and submit them to ESMCI/CIME?

bishtgautam:
>Should I remove them from this PR and submit them to ESMCI/CIME?

No, leave those changes in. 

FYI, e3sm is considering adding CIME as a submodule.

rgknox:
I ran some tests on lawrencium.  It looks like cprnc is having some trouble with a history variables using the dimension: "fates_levelem". This dimension does seem to be related to the changes, but I'm not sure what is going on.  There are two history variables that use this dimension: 
```
	fates_levelem = 1 ;
	int fates_levelem(fates_levelem) ;
		fates_levelem:long_name = "FATES element (C,N,P,...) identifier" ;
		fates_levelem:units = "-" ;
	float LITTER_IN_ELEM(time, fates_levelem, lat, lon) ;
	float LITTER_OUT_ELEM(time, fates_levelem, lat, lon) ;
```
In both the base and the test history file, the dimension reports as 1:
```
data:

 fates_levelem = 1 ;

```
However, for both runs master and test, have no-data flags for the all LITTER_IN_ELEM and LITTER_OUT_ELEM.

The cprnc diff log on the histofy file ends with the following error:
```
 fates_agmap_levagepft   (fates_levagepft)
                  84  (     7) (     1)
                  84        7        1
                  84        7        1
                  84  (     7) (     1)
          avg abs field values:         4
                                        4
 fates_levelem   (fates_levelem)
forrtl: severe (71): integer divide by zero
cprnc              00000000004B67BD  Unknown               Unknown  Unknown
cprnc              00000000004B4657  Unknown               Unknown  Unknown
cprnc              000000000045E4F4  Unknown               Unknown  Unknown
cprnc              000000000045E306  Unknown               Unknown  Unknown
cprnc              000000000042C6A6  Unknown               Unknown  Unknown
cprnc              0000000000430856  Unknown               Unknown  Unknown
Unknown            00002AF72A663630  Unknown               Unknown  Unknown
cprnc              000000000040D5CE  compare_vars_mod_         371  compare_vars_mod.F90.in
cprnc              000000000040AA15  compare_vars_mod_         199  compare_vars_mod.F90.in
cprnc              00000000004196CA  MAIN__                     92  cprnc.F90
cprnc              000000000040941E  Unknown               Unknown  Unknown
libc.so.6          00002AF72A892545  Unknown               Unknown  Unknown
cprnc              0000000000409329  Unknown               Unknown  Unknown
```

rgknox:
I'm adding a fix on the fates side that should fix these erroneous values for LITTER_IN_ELEM and LITTER_OUT_ELEM.  Will update fates, update E3SM pointer to FATES, and re-test.

rgknox:
see https://github.com/NGEET/fates/pull/620

rgknox:
I re-built cprnc on our lawrencium machine, updated some machine files and I'm now pointing to an updated version of fates that adresses some issues regarding unitialized output variables.  I merged in master on my test branch, and compare it to master as of this morning.  I noticed that 1 test fails unexpectedly and one test did not pass baseline for some reason...

```
ash-4.2$ ./cs.status.E25205dc-Fab1647a-initbug-pt2 | grep FAIL
    FAIL ERS.f09_g16.ICLM45BC.lawrencium-lr3_intel TPUTCOMP Error: Computation time increase > 25 pct from baseline
    FAIL ERS.f19_f19.I1850CLM45CN.lawrencium-lr3_intel TPUTCOMP Error: Computation time increase > 25 pct from baseline
    FAIL ERS.f19_f19.I20TRCLM45CN.lawrencium-lr3_intel BASELINE E7dba747-Fab1647a-master: DIFF
    FAIL ERS.f19_f19.ICLM45.lawrencium-lr3_intel TPUTCOMP Error: Computation time increase > 25 pct from baseline
    FAIL ERS.f19_g16.I1850GSWCNPRDCTCBC.lawrencium-lr3_intel.clm-ctc_f19_g16_I1850GSWCNPRDCTCBC TPUTCOMP Error: Computation time increase > 25 pct from baseline
    FAIL ERS.f19_g16.I20TRGSWCNPRDCTCBC.lawrencium-lr3_intel.clm-ctc_f19_g16_I20TRGSWCNPRDCTCBC TPUTCOMP Error: Computation time increase > 25 pct from baseline
    FAIL ERS_Ld3.f45_f45.ICLM45ED.lawrencium-lr3_intel.clm-fates BASELINE E7dba747-Fab1647a-master: DIFF
  SMS.r05_r05.I1850CLM45CN.lawrencium-lr3_intel (Overall: FAIL) details:
    FAIL SMS.r05_r05.I1850CLM45CN.lawrencium-lr3_intel RUN time=87
```
These two struck me as odd: 
FAIL ERS.f19_f19.I20TRCLM45CN.lawrencium-lr3_intel BASELINE E7dba747-Fab1647a-master: DIFF
    FAIL SMS.r05_r05.I1850CLM45CN.lawrencium-lr3_intel RUN time=87

I will look into it, but was wondering if these are expected fails..

rgknox:
I ran a baseline comparison of this branch with itself to see if the cprnc generates errors, and they are now gone. My guess is that the fixes in this branch may hopefully fix future ICLMED baseline comparisons.
Pending any comments about the two tests I note in my last comment, I think this PR is ready for integration testing.

rljacob:
@bishtgautam see above.

bishtgautam:
@rgknox I missed your earlier comment. I will review this PR.

components/clm/src/main/clmfates_interfaceMod.F90
```diff
@@ -206,9 +209,148 @@ module CLMFatesInterfaceMod
...
+     integer                                        :: pass_vertsoilc
+     integer                                        :: pass_spitfire     
+     integer                                        :: pass_ed_st3
+     integer                                        :: pass_logging
+     integer                                        :: pass_ed_prescribed_phys
+     integer                                        :: pass_planthydro
+     integer                                        :: pass_inventory_init
+     integer                                        :: pass_is_restart
+     
+     if (use_fates) then
```
bishtgautam, line 241: Why not have the following?
   `if (.not. use_fates) return`

rgknox: I can look through this, but there is a small possibility that we need the last call in the routine even if we don't have fates turned on... but I could double check that.

rgknox: well, according to Ryan from the past: https://github.com/E3SM-Project/E3SM/pull/3471/files/55752fa50e3312cdb39d703828772af056cfbb39#diff-26b8ffc0ec3b66f6b87a776e9ee64416R345
But I don't trust that guy

rgknox: @bishtgautam , github code linking has been weird for me with low bandwidth lately, so to expand upon my cheeky comment: as far as I can tell, I think we do actually need some of the code that comes after the "if" clause, so we shouldn't return early.

bishtgautam: Ok, sounds good.

rljacob:
Is this ready to merge?

bishtgautam:
Yes, this PR is ready. I will merge it when next is opened for new merges.

rljacob:
Testing shows a diff in ERS_Ld3.f45_f45.ICLM45ED.melvin_gnu.clm-fates.  Expected?

bishtgautam:
I will change the tag to non-BFB and open a ticket.

PR #3473, opened 2020-02-27
For NERSC, change the PROJECT (repo name) from "acme" to "e3sm".

ndkeen:
For NERSC, change the PROJECT (repo name) from "acme" to "e3sm".

This can be manually set for existing cases by:
case.submit -a="--account e3sm"

bfb

### DIFF
--- a/cime/config/e3sm/machines/config_machines.xml
+++ b/cime/config/e3sm/machines/config_machines.xml
@@ -61,7 +61,7 @@
     <OS>CNL</OS>
     <COMPILERS>intel,gnu</COMPILERS>
     <MPILIBS>mpt</MPILIBS>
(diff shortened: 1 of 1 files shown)


worleyph:
Noel, as discussed, please add e3sm to SAVE_TIMING_DIR_PROJECTS, e.g.

        <SAVE_TIMING_DIR_PROJECTS>acme,e3sm,m3411,m3412</SAVE_TIMING_DIR_PROJECTS>

I still do not see a /global/cfs/cdirs/e3sm/ directory. It is this transition that will be tricky, or do we need to do this at all?

ndkeen:
I already merged to next. I could leave this PR and create another one. Or revert this change and make another PR.

PR #3480, opened 2020-03-04
Disable database integrity check

jgfouca:
It's clear this check is doing more harm than good, so disabling.

[BFB]


rljacob:
Is the long-term plan still to allow a "cime wizard" mode which deactivates this (it would still be on by default)?

jgfouca:
That's one of several options that I will need to explore. I think there may be more elegant solutions possible.

PR #3483, opened 2020-03-05
Update mpas-source, COMPASS only

mark-petersen:
This update only changes files within mpas-source that pertain to testing and setup. It does not alter any E3SM routines.

[BFB]


jonbob:
merged to next

jonbob:
merged to master

PR #3484, opened 2020-03-05
Adds support for F20TRC5-CMIP6 compset at ne120 resolution

bishtgautam:
The definition of ne120np4 is modified to use 0.125deg river grid instead
of 0.5deg river grid. The missing landuse timeseries for ne120 resolution
is added for F20TRC5-CMIP6

[non-BFB]


bishtgautam:
This PR is non-BFB because it changes the definition of ne120np4 grid, but we don't have any tests for that grid.

rljacob:
@jqyin please start merging this.

rljacob:
@jqyin this needs your attention.

jqyin:
Merged #3484

PR #3486, opened 2020-03-06
Set spectral bounds based on RRTMGP inputdata

### LINKED ISSUE #3479: Possible RRTMGP bug: bandwidth
Possible bug reported by Xianglei Huang from UMich:

> 	We are moving along to work with RRTMGP. One thing that we noticed is that the RRTMGP code that we obtained following your previous instruction has some inconsistencies with the RRTMGP reported in Pincus 2019 RRTMGP paper. The details are as follows:
> 
> 	The boundaries for two spectral bands in RRTMGP are not the same as what described in Pincus paper, and they apparently are still the old values used in the RRTMG_LW. For example, the first band should be 0-250cm-1, not 0-350cm-1. We have tested and validated that 0-250cm-1 (as in Pincus paper) should be the right bandwidth. 
> 
> 	In components/cam/src/physics/rrtmgp/radconstants.F90
> Line 37, 2600 should be 2680
> Line 40, 2600 should be 2680
> Line 84, 350 should be 250
> Line 85, 2600 should be 2680
> Line 88, 350 should be 250
> Line 89, 2600 should be 2680
> 
> 	These mistakes might not directly affect the simulations you have done, assuming they have updated the data files correctly but just copied & pasted wrong band widths. It might warrant a double check with developers who put RRTMGP to this code to make sure about this. For us, because we need to recompute surface skin temperature, a wrong bandwidth can mess up the entire simulation.
> 
> 	Xianglei

brhillman:
Set spectral band limits based on values from RRTMGP inputdata files, rather than hard-coding these values. This makes the interface more flexible and helps avoid confusion as to what bands are in use. Also add a routine to check that these band limits are set correctly. This routine is to be used for testing purposes, and is not called during runtime by default, but a call can be inserted after setting the band limits to effectively unit test to confirm that bands are set consistent with what we expect. Fixes #3479 but noting that we should probably revisit the cloud and aerosol optical depth coefficients that are tabulated in the input files prepared for RRTMG, which has very slightly different bands.

[BFB]

### DIFF
--- a/components/cam/src/physics/rrtmgp/radconstants.F90
+++ b/components/cam/src/physics/rrtmgp/radconstants.F90
@@ -16,7 +16,7 @@ module radconstants
 ! radiation interface.
 
 use shr_kind_mod,   only: r8 => shr_kind_r8
-use cam_abortutils,     only: endrun
+use cam_abortutils, only: endrun
 
 implicit none
 private
@@ -32,16 +32,11 @@ integer, parameter, public :: nswbands = 14
 ! Note: Currently rad_solar_var extends the lowest band down to
 ! 100 cm^-1 if it is too high to cover the far-IR. Any changes meant
 ! to affect IR solar variability should take note of this.
-
-real(r8),parameter :: wavenum_low(nswbands) = & ! in cm^-1
-  (/ 820._r8, 2600._r8, 3250._r8, 4000._r8, 4650._r8, 5150._r8, 6150._r8, &
-    7700._r8, 8050._r8,12850._r8,16000._r8,22650._r8,29000._r8,38000._r8/)
-real(r8),parameter :: wavenum_high(nswbands) = & ! in cm^-1
-  (/2600._r8, 3250._r8, 4000._r8, 4650._r8, 5150._r8, 6150._r8, 7700._r8, &
-    8050._r8,12850._r8,16000._r8,22650._r8,29000._r8,38000._r8,50000._r8/)
+real(r8), dimension(nswbands) :: wavenum_sw_lower, wavenum_sw_upper
 
 ! Solar irradiance at 1 A.U. in W/m^2 assumed by radiation code
 ! Rescaled so that sum is precisely 1368.22 and fractional amounts sum to 1.0
+! NOTE: This is copied from RRTMG; should be revisited for RRTMGP
 real(r8), parameter :: solar_ref_band_irradiance(nswbands) = & 
    (/ &
     12.89_r8,  12.11_r8,  20.3600000000001_r8, 23.73_r8, &
@@ -54,7 +49,6 @@ real(r8), parameter :: solar_ref_band_irradiance(nswbands) = &
 integer, parameter, public :: idx_sw_diag = 11 ! index to sw visible band
 integer, parameter, public :: idx_nir_diag = 9 ! index to sw near infrared (778-1240 nm) band
 integer, parameter, public :: idx_uv_diag = 12 ! index to sw uv (345-441 nm) band
-
 integer, parameter, public :: rrtmg_sw_cloudsim_band = 10  ! rrtmg band for .67 micron
 
 ! Number of evenly spaced intervals in rh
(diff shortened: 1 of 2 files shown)


singhbalwinder:
@brhillman : Is this PR BFB for the RRTMP test as well?

brhillman:
@singhbalwinder yes

singhbalwinder:
Thanks, I will merge it today if repo is open.

singhbalwinder:
Pushed to next

rljacob:
This PR seemed to cause some new fails on compy-pgi related to restart tests:
```
ERP_Ln9.ne4_ne4.F-EAMv1-AQP1.compy_pgi
ERP_Ln9.ne4_ne4.F-EAMv1-RCEMIP.compy_pgi
ERP_Ln9.ne4_ne4.FC5AV1C-L.compy_pgi
ERS.ne11_oQU240.A_WCYCL1850.compy_pgi
ERS_Ld5.ne4_ne4.FC5AV1C-L.compy_pgi.cam-rrtmgp
ERS_Ln9.ne4_ne4.FC5AV1C-L.compy_pgi
```

singhbalwinder:
Thanks @rljacob . I wasn't paying attention to Compy. @brhillman: would you be able to take a look at it?

brhillman:
@rljacob I'm not sure how that's possible, half of those tests will not even build this code. The ones that might are `ERP_Ln9.ne4_ne4.F-EAMv1-AQP1.compy_pgi`, `ERP_Ln9.ne4_ne4.F-EAMv1-RCEMIP.compy_pgi`, and `ERS_Ld5.ne4_ne4.FC5AV1C-L.compy_pgi.cam-rrtmgp`.

brhillman:
@singhbalwinder I do not have an account on compy. Did these same tests fail on cori?

singhbalwinder:
Cori doesn't test the integration suite, so we don't know if these fail on Cori. Other machines which run integration suite are all passing, probably because they all use Intel compiler. This issue could be specific to PGI/Compy. I will take a look to see what might be going on.

brhillman:
I do see an issue now if carma is used with RRTMGP, carma is trying to read the spectral bounds before the RRTMGP interface initializes them. That might be causing the RRTMGP-related tests to fail, since carma seems to be initialized whether it is active or not. So if that stuff is getting written to restart files, then that could be a problem. My fault for not catching this. May have to revert to the old behavior but with the bands fixed to address #3479.

singhbalwinder:
Great! Would you be able to fix it soon? 
It is surprising that PGI caught it but not Intel.

brhillman:
Yeah oddly ERS tests with RRTMGP are passing for me on cori with intel. There's a couple ways of fixing this: 1) move initialization of the bands to before carma_register, which I think means moving the radiation initialization to the radiation_register routine and then moving it up above the call to carma_register 2) remove carma, which I don't think is being used at all (and is the only other code that actually uses this at runtime). 3) revert to the old behavior of hard-coding the bands.

brhillman:
@rljacob is there any reason most parameterizations seem to do their file reading in `_init` routines? To keep the functionality this PR added, I would need to move the reading of the RRTMGP inputdata to the `radiation_register` routine, and then move the `carma_register` call after the `radiation_register` call (or vice versa). The alternative seems to be to revert to the old behavior of hard-coding the RRTMGP bands into the `radconstants` module, but hard-coding values kind of makes me cringe. Thoughts?

rljacob:
I'm pretty sure the idea is that time-invariant data should be read once at initialization of the model. All those reads go in the "init" method and then the values are saved in some structure that can be queried during the run.   This is done to minimize I/O operations during the run.

brhillman:
Oh, maybe I had it backwards then...I was thinking that "register" was called before "init". If init is called first, then disregard my concern about this PR, and I'm back to being stumped.

brhillman:
Nope, I was right, `phys_init` is called after `phys_register`, and CARMA is trying to use things in `phys_register` that are not setup until `phys_init` is called. So I would have to move that file reading up from init into register.

brhillman:
Note that both `phys_init` and `phys_register` are called at model initialization, in `cam_init`.

rljacob:
It looks like ERIO.ne30_g16_rx1.A.sandiatoss3_intel was broken by this commit and its been failing ever since.
https://my.cdash.org/viewTest.php?onlyfailed&buildid=1797946

PR #3495, opened 2020-03-12
Changing modified CORE runoff files time interpolation to be consistent with standard CORE runoff.

darincomeau:
This adds entries for time interpolation for the modified CORE runoff files used by cryosphere compsets to be 'upper' (from the default 'linear'), to be consistent with the default CORE runoff file.

[non-BFB] For Cryosphere compsets only (G-cases using the 'ISMF' or 'DIB' tags).

### DIFF
--- a/cime/src/components/data_comps/drof/cime_config/namelist_definition_drof.xml
+++ b/cime/src/components/data_comps/drof/cime_config/namelist_definition_drof.xml
@@ -533,7 +533,13 @@
     <values>
       <value>linear</value>
       <value stream="rof.diatren_ann_rx1">upper</value>
+      <value stream="rof.diatren_ann_ais00_rx1">upper</value>
+      <value stream="rof.diatren_ann_ais45_rx1">upper</value>
+      <value stream="rof.diatren_ann_ais55_rx1">upper</value>
       <value stream="rof.diatren_iaf_rx1">upper</value>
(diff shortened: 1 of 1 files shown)


darincomeau:
To clarify, this will only affect G-cases using the ISMF or DIB tags.

billsacks:
@rljacob I appreciate your pointing out this change that might affect CESM. I'm not an appropriate reviewer of this change, though. @klindsay28 are you a good person to look at this? (If not, then do you know who is?)

billsacks:
(I'll keep my name on here so that I give it an official stamp of approval once I get an okay from someone on the cesm side.)

klindsay28:
To me, this looks like a reasonable set of mods that are ensuring consistent settings for some particular streams.

It doesn't look like it will affect defined CESM compsets, as POP, MOM, and CICE don't use DROF_MODE settings that trigger these streams.

rljacob:
@jonbob don't forget about this one too.

jonbob:
Also on my list

rljacob:
@jonbob can this be made a priority?

jonbob:
@rljacob - I'll merge this one as soon as next is open

jonbob:
merged to next

jonbob:
merged to master

PR #3497, opened 2020-03-13
Provide clear messages about dycore and physgrid.

### LINKED ISSUE #3493: Can one tell which dynamical core is being used based on atm.log?
As far as I can tell, it is not possible to determine which dynamical core (preqx, theta-l, ...) is actually being used by the model by inspecting atm.log output file. But maybe I'm not looking at the right place. 

This would be very useful to have (especially since I accidentally mis-configured one simulation earlier this week).

ambrad:
In v2, a fair bit is changing in EAM. Help keep track of these changes by providing clear output of each. Example output:
```
 Running dycore: theta-l
 Running tracer transport method: semi-Lagrangian
... other messages ...
 gfr> Running with dynamics and physics on separate grids (physgrid).
```
In an E3SM run, these messages will appear in atm.log.

Also provide an XS PE layout for the pg2 coupled compset on Compy.

fixes #3493

[BFB]

### DIFF
--- a/cime/config/e3sm/allactive/config_pesall.xml
+++ b/cime/config/e3sm/allactive/config_pesall.xml
@@ -7825,7 +7825,7 @@
   <grid name="a%ne30np4.pg2_l%.+_oi%oEC60to30">
     <mach name="compy">
       <pes compset=".*CAM5.+CLM45.+MPASSI.+MPASO.+MOSART.+" pesize="XS">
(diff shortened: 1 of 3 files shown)


ambrad:
I'm running e3sm_dev right now.

AaronDonahue:
Pushing to next now.

PR #3498, opened 2020-03-16
Replace WHERE with DO loops in landice for PGI v19.10

### LINKED ISSUE #2862: Build error on MPAS-LI with PGI compiler on Summit
There is a build error when comping MPAS-LI on Summit, The error message is  as follows:

```
PGF90-S-0000-Internal compiler error. items were added to sem.p_dealloc but not freed       0  (mpas_li_subglacial_hydro.f90: 1288)
```
The test name is SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.summit_pgi.  I have tried PGI/18.7, 18.10 and 19.1 on Summit and got the same error.

jonbob:
This PR brings in a new mpas-source submodule that has changes only in the landice core that replaces WHEREs with DO loops for PGI v19.10.

Fixes #2862 

[BFB]


jonbob:
merged to next

jonbob:
merged to master

PR #3503, opened 2020-03-18
Update mpas-source: init mode, COMPASS, minor ice shelf diagnostics

mark-petersen:
These changes are all related to the creation of initial conditions for ice shelves. It does not alter forward simulations without ice shelves.

[BFB]


jonbob:
merged to next

mark-petersen:
@jonbob I updated the mpas-source to include: 
Add mpas_ocn_init_smoothing.F to ocean.cmake

Sorry, I'll check that from now on. Stand-alone does not compile with cmake yet.

jonbob:
Really strange to me that my tests passed on anvil? But re-merging now...

jonbob:
re-merged to pick up filename added to ocean.cmake

jonbob:
merged to master

Issue #3377, opened 2020-01-06
get_cvmix.sh and get_BGC.sh not working


worleyph:
executing the scripts 

     components/mpas-source/src/core_ocean/get_cvmix.sh
     components/mpas-source/src/core_ocean/get_BGC.sh

no longer eliminates the need to download updated copies of CVMIX and BGC during case builds on Cori-KNL. From the e3sm.bldlog ...

      ** Using git to acquire cvmix source. **
      ** Using git to acquire BGC source. **

jgfouca:
@worleyph , this is likely related to the CMake-ification of the MPAS build in E3SM. I didn't quite understand your comment. Are you saying running these scripts once, either by-hand or as part of a build, should prevent them from being run again during subsequent builds of the same case?

worleyph:
By-hand certainly. That was the original motivation. @philipwjones can give you the background to this as Phil wrote the original scripts, I believe.

jgfouca:
@worleyph , just looking at the core_ocean/Makefile vs. ocean.cmake, I can clearly see that I changed this behavior (inadvertently). Sorry about that! Fix incoming.

philipwjones:
@jgfouca - I didn't write the originals, but I did modify them to avoid downloading files for every build. Basically, this is a case where we need to get the libs if they haven't yet been downloaded, but if the correct version is there and the files have been downloaded, the script should detect that and not do another download. I agree that it was something probably missed in the CMake transition.

jgfouca:
@philipwjones,

I took another look at the ocean Makefile, this is the target for getting cvmix:
```
cvmix_source: get_cvmix.sh
        (/bin/bash ./get_cvmix.sh)
        (cd cvmix)
```
Since `cvmix_source` is a "fake" target (not associated with a real file/directory) won't it get run every time the user runs Make?

philipwjones:
@jgfouca - the script gets run every time, but the first part of the script checks for the presence of the proper git clone and avoids a new clone/download. So as long as the script gets run by the CMake version, it should work properly?

jgfouca:
Oh, it looks like the code to avoid repeating downloads is in the script itself, not the build system. In that case, I'm not sure why things aren't working as they were before.

philipwjones:
Hmmm...agreed.

jgfouca:
[...]

PR #3507, opened 2020-03-20
Update mpas-source: cvmix and BGC submodules

mark-petersen:
This is a significant change to the structure of the MPAS repository.
Instead of external code dependencies being downloaded by home-grown
scripts, we instead express the dependency as a git submodule. This will
impact the cvmix and BGC dependencies for core_ocean.

For additional discussion, see #3377 and https://github.com/MPAS-Dev/MPAS-Model/pull/420

[bfb]


mark-petersen:
@jgfouca this only has the MPAS updates. Please add whatever is needed within the E3SM repo.

mark-petersen:
@jonbob, this can go in after the sea ice change. This pointing to the `e3sm/testing` branch, but that is easy to point back at the last minute.

jgfouca:
I don't think any changes are needed in E3SM.

rljacob:
If a submodule has its own submodules, does "git submodule update --init" at the top level retrieve all of them?   I think this is our first example of that in E3SM.

jgfouca:
@rljacob , all submodule commands will need `--recursive`

mark-petersen:
@jonbob and @jgfouca if you could each test this PR on a different machine, that would ensure everything is working. 

@jgfouca are there any e3sm scripts that need `--recursive` added to a `git submodule update` command, or is that always done before launching e3sm scripts?

mark-petersen:
@jonbob I updated `e3sm/develop` on the mpas repo, so this PR can be merged after testing. I was able to obtain the cvmix and BGC submodules on cori with 
`git submodule update --recursive --init`
so everything appears to work.

jonbob:
Passed SMS.ne4_oQU240.A_WCYCL1850.anvil_intel

jonbob:
merged to next

jgfouca:
> @jgfouca are there any e3sm scripts that need --recursive added to a git submodule update command, or is that always done before launching e3sm scripts?

Jenkins should already be doing that. It's possible there are cron jobs that need updating. I updated the ones in our test script repo.

jonbob:
merged to master

PR #3505, opened 2020-03-19
Bug fix for COSP2 MODIS simulator

wlin7:
This is to incorporate bug fix by Dustin Swales for MODIS simulator
in COSP2 as described at https://github.com/ESCOMP/CAM/pull/95.

Issue 1:
From COSP1.4 to COSP2 there was a change in the LUTs used by the particle-size
retrieval within the MODIS simulator, but this change was not accounted for on
the EAM side. This mismatch in LUTs resulted in the retrieval failing more often,
hence the reduction in MODIS total cloud fraction.
Solution 1:
Use the new LUTs in the COSP/EAM interface (cosp2/optics/cosp_optics.F90).

Additionally there was a bug-fix to the MODIS optics in COSP2 that hadn't
yet been propagated into EAM. The implementation of COSP2 in EAM occurred
around the same time when this fix was implemented in COSP2
(and as a patch to COSP1.4)

Issue 2: The combining of the MODIS optical fields, single-scattering
albedo and asymmetry parameter, were incorrect.
Solution 2: Propagate changes in the computation of optical properties into CAM.

[BFB] except for COSP-MODIS diagnostics

Issue #3482, opened 2020-03-05
cosp2 fails cam-cosplite_nhtfrq5 in an odd way: bloated e3sm.exe?


wlin7:
cosp2 in next via PR #3458 fails cosplite test in likely all machines. Take anvil as an example,
The test is SMS_D_Ln5.ne4_ne4.FC5AV1C-L.anvil_intel.cam-cosplite_nhtfrq5

Backtrace of the error messages are 
```
62: forrtl: severe (408): fort: (8): Attempt to fetch from allocatable variable SD_CS when it is not allocated
 62: 
 62: Image              PC                Routine            Line        Source             
 62: e3sm.exe           000000000B6F8E36  Unknown               Unknown  Unknown
 62: e3sm.exe           000000000225FFC9  cospsimulator_int        2099  cospsimulator_intr.F90
 62: e3sm.exe           000000000134F8B9  radiation_mp_radi        1572  radiation.F90
 62: e3sm.exe           0000000002742CE6  physpkg_mp_tphysb        2767  physpkg.F90
 62: e3sm.exe           000000000270CFA1  physpkg_mp_phys_r        1047  physpkg.F90
 62: e3sm.exe           0000000000856CE6  cam_comp_mp_cam_r         250  cam_comp.F90
 62: e3sm.exe           000000000082997A  atm_comp_mct_mp_a         574  atm_comp_mct.F90
```
The allocatable variable SD_CS that the above error message refers to is not supposed to be invoked, because it is used only when lradar_sim = .true. (radar simulator), and with cosplite, lradar_sim = .false.

A manually creased case using the same compset FC5AV1C-L and grid ne4_ne4, using the following command to enable cosp
  
./xmlchange -file env_build.xml CAM_CONFIG_OPTS="-cosp" -append

and add cosp_lite = .true. in user_nl_cam, would run well. 

The configurations for the manual case and SMS_D_Ln5.ne4_ne4.FC5AV1C-L.anvil_intel.cam-cosplite_nhtfrq5 are supposed to be the same.

But e3sm.exe generated by SMS_D_Ln5.ne4_ne4.FC5AV1C-L.anvil_intel.cam-cosplite_nhtfrq5 is more than twice as large as the one generated from the manual case.

[...]

PR #3508, opened 2020-03-23
Change member in cosp_optical_inputs type to be allocatable

### LINKED ISSUE #3506: cosp2 appears to be much more expensive than cosp1, really so?
Since cosp2 (PR #3458 was merged, there are several reports that cosp2 appears to be much more expensive than cosp1 (see the comment section of #3458). @amametjanov and @golaz noticed that the cost of full coupled run increases by 20% on multiple machines. Az also showed that based on cosp_run timer, the slow-down from cosp1 to cosp2 can be 11x. The allactive test 
SMS_Ld2.ne30_oECv3_ICG.A_WCYCL1850S_CMIP6.anvil_intel.allactive-v1cmip6 also reported increase in computational time

`Computation time increase > 10 pct from baseline`

@wlin7 tested in 10-dam atm simulations on anvil using the hashes before and after the merge , and found cosp2 itself cost 2x as much as cosp1, based on cosp_run timer statitics. The ratio of cosp to total simulation time increases from 1.1% to 2.1%. The system throughput was stable during the test, so the pre-COSP2 and post-COSP2 performance are fairly comparable in that settings. The cost of cosp as a ratio of the total simulation cost is much less than the degradation in performance observed by Az and Chris on fully coupled runs.

Moreover, cosp2 developer @dustinswales indicated that cosp2 is less expensive than cosp1 in CESM (cosp_amwg mode). This is apparently inconsistent with what observed with E3SM.

During allactive coupled run, it was also observed that the execution of cosp2 become slower over time during the course of the simulation. Longer atm-only simulations are being carried out on anvil to see if this behavior reproducible. Further tests will be performed to see if the 20% jump in cost unique to the coupled run (which by itself does not make sense, but it could imply some memory issue. The full coupled simulation with cosp2 did show increase in memory usable over time)

worleyph:
The COSP2 cosp_optical_inputs type includes a member that is of type
radar_cfg. radar_cfg itself contains members that are large, which causes
performance issues when allocating a variable of type
cosp_optical_inputs (implicitly as a local variable or explicitly with
an allocate). This can be resolved by instead defining the member
of type radar_cfg in type cosp_optical_inputs to be allocatable and
explicitly allocating/deallocating this member when using variables of
type cosp_optical_inputs.

[BFB]

Fixes #3506

### DIFF
--- a/components/cam/src/physics/cam/cospsimulator_intr.F90
+++ b/components/cam/src/physics/cam/cospsimulator_intr.F90
@@ -3312,6 +3312,8 @@ CONTAINS
              y%tautot_S_ice(       npoints, ncolumns         ),&
              y%tautot_S_liq(       npoints, ncolumns)         ,&
              y%fracPrecipIce(npoints,   ncolumns))
+    allocate(y%rcfg_cloudsat)
+
   end subroutine construct_cospIN
   
   !%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
@@ -3470,6 +3472,8 @@ CONTAINS
     if (allocated(y%ss_alb))              deallocate(y%ss_alb)
     if (allocated(y%fracLiq))             deallocate(y%fracLiq)
     if (allocated(y%fracPrecipIce))       deallocate(y%fracPrecipIce)
+    if (allocated(y%rcfg_cloudsat))       deallocate(y%rcfg_cloudsat)
+
   end subroutine destroy_cospIN
   !%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
   ! SUBROUTINE destroy_cospstateIN     
--- a/components/cam/src/physics/cosp2/Cosp.cmake
+++ b/components/cam/src/physics/cosp2/Cosp.cmake
@@ -6,7 +6,7 @@ set(COSP_SOURCES
    cam/src/physics/cosp2/external/src/cosp_constants.F90
    cam/src/physics/cosp2/external/src/simulator/cosp_cloudsat_interface.F90
    cam/src/physics/cosp2/external/src/cosp_config.F90
-   cam/src/physics/cosp2/external/src/cosp.F90
+   cam/src/physics/cosp2/local/cosp.F90
    cam/src/physics/cosp2/external/src/cosp_stats.F90
    cam/src/physics/cosp2/external/src/simulator/quickbeam/quickbeam.F90
    cam/src/physics/cosp2/external/src/simulator/parasol/parasol.F90
--- a/components/cam/src/physics/cosp2/local/cosp.F90
+++ b/components/cam/src/physics/cosp2/local/cosp.F90
@@ -1,0 +1,3662 @@
+! %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+! Copyright (c) 2015, Regents of the University of Colorado
+! All rights reserved.
+!
+! Redistribution and use in source and binary forms, with or without modification, are
+! permitted provided that the following conditions are met:
+!
+! 1. Redistributions of source code must retain the above copyright notice, this list of
+!    conditions and the following disclaimer.
+!
+! 2. Redistributions in binary form must reproduce the above copyright notice, this list
+!    of conditions and the following disclaimer in the documentation and/or other
+!    materials provided with the distribution.
+!
+! 3. Neither the name of the copyright holder nor the names of its contributors may be
+!    used to endorse or promote products derived from this software without specific prior
+!    written permission.
+!
+! THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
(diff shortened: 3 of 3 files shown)


worleyph:
Ran 

       --compset FC5AV1C-L --res ne30_ne30

on 900 MPI tasks (no OpenMP) on Compy to verify that this compiles and eliminates the performance degradation observed with the unmodified code (as compared to COSP1). This still needs to be exercised more widely, on other systems and with other PE layouts, including with threading, to verify that this really does resolve all of the COSP2-related performance issues.

worleyph:
Note that the modification in cosp.F90 is a one line change (< before; > after)

     163c163
     <      type(radar_cfg) :: &
     ---
     >      type(radar_cfg),allocatable :: &

rljacob:
Should we try to make a PR to the cosp2 master for the changes in cosp.F90?

wlin7:
Yes, @rljacob .I have emailed cosp2 developer @dustinswales of this fix and PR. He probably will incorporate the changes to cosp.F90 in COSPv2 repo. Once he does, I will switch back to use the submodule as used to be.

wlin7:
Separate test on anvil also sees the performance degradation now gone.  And indeed it appears cosp2 is faster than cosp1 in cosp_lite mode, in terms of cosp_run timer.

Merged to next.

wlin7:
For record how much this fix helps with the performance:, cosp2 in this branch is 3x faster than cosp1. Before the change, not even considering the effect of load imbalance, cosp2 is 2x slower than cosp1. But note that cosp2 is currently with MODIS simulator disabled. A fix for modis is ready in #3505.

PeterCaldwell:
Wow, thanks @ambrad and @worleyph ! Heroic work over the weekend!

ambrad:
Uh, @worleyph did everything. I was just following his work on github and took a peak at the COSPv2 repo at the end, based on his final measurement, to see if there was anything obvious in cosp_optical_inputs that could be the problem, which there was. But don't worry: I was still working this weekend. :)

worleyph:
My strategy was to post working results as often as possible to allow others to jump in and accelerate the process - which worked to perfection in this case with the contributions by @ambrad, @amametjanov , @wlin7 , and @zyuying . And credit should definitely go to everyone for paying attention over the weekend. My being partially retired and hourly, the weekend is no different than any other day of the week to me (which can be annoying to my collaborators :-) ).

wlin7:
Notes on test return for SMS_Ld2.ne30_oECv3_ICG.A_WCYCL1850S_CMIP6.anvil_intel.allactive-v1cmip6

1.  This remains: "Error: Computation time increase > 10 pct from baseline"
2. The test took longer after this PR than before. (March 24 return vs March 23). However, manually running this test does show it took less time after this PR. 
3. The warning of "Error: Computation time increase > 10 pct from baseline" actually existed before COSP2 PR. For example, on the [March 5th return.](https://my.cdash.org/testDetails.php?test=55013612&build=1784393) and persists since. Some other PR may needs to be looked at concerning the increase in time.

PeterCaldwell:
Sorry for not giving credit where credit was due. The point of my comment was supposed to be "wow, nice job" - I should have been more inclusive in my call outs. Thanks also @wlin7 , @amametjanov , and @zyuying !

wlin7:
Not a problem at all, @PeterCaldwell . Several of us working on and monitoring the cosp2 PR, including @amametjanov and @zyuying , were puzzled by the behavior, and appreciated finally got  relieved. You are absolutely right it is @worleyph and @ambrad's amazing effort and intuition that nailed it within a short period time!

amametjanov:
Re-ran 1-month comparisons on compy, anvil and cori-knl: throughput is faster with cosp2 on compy and anvil; 6 sec/day slower on cori-knl due to "KNL variability" (157 vs 163 sec/day cosp1-vs-cosp2). I think it's okay to go ahead and "bless" baselines (note that baseline DIFF and throughput warnings started as soon as cosp2 PR was first merged to `next`).

I think the length of the regression test should be increased from 2 to 5 days (or better 1 month, which runs under 2 hours of wall clock) to get a more accurate throughput metrics despite IO variability.

wlin7:
Hi @amametjanov , just to double check with you on the DIFF. COSP2 was merged to next on March 7th. From what I can see in CDash, the SMS_Ld2.ne30_oECv3_ICG.A_WCYCL1850S_CMIP6.anvil_intel.allactive-v1cmip6 returned DIFF since March 5th, along with the warning of > 10% increase in time. The most recent bless was Jan. 27th. COSP2 definitely added to the DIFF. But I want to make sure there was no other PR before COSP2 that could cause diff.

Also, COSP2 is expected to only affect cam.h0 and cam.h5 of the test. But the test said cam.h1, h2, h3, h4 and cpl.hi all have diffs.

rljacob:
The atm integrators should track down the source(s) of the diff.

wlin7:
Sorry, I was wrong. COSP2 actually was first merged on March 3rd, but an unallocated array used in a call caused the cosplite-nhtfrq6 test to fail. The failure occurs only when the model is built in DEBUG mode (see #3482). 

Since cosp2 was in next on March 3d, it must have caused the allactive-v1cmp6 test to diff. Still, it should only affect cam.h0 and cam.h5, which contain COSP diagnostic fields. I will further look into it.

wlin7:
Quick update: It is clear all the diffs since March 4 are due to COSP2.

Even the most recent test return, only cam.h0 and cam.h5 diff from baselines in physical quantities. cpl.hi are identical. Though cam.h1, h2 and h3 do not contain cosp diagnostic fields, they include the  cosp coordinate variables (all atm histfiles do when cosp is on). The diff are on those coordinate variables.

I am going to submit request to bless the test.

wlin7:
Merged to master. 

Request to bless SMS_Ld2.ne30_oECv3_ICG.A_WCYCL1850S_CMIP6.anvil_intel.allactive-v1cmip6 submitted.

PR #3510, opened 2020-03-23
Submodule operations must now be recursive

jgfouca:
I grepped our repo and updated the ones I could find.

[BFB]


jgfouca:
Skipping integration.

Issue #3270, opened 2019-10-30
Many tests with LND fail with GNU on cori-knl and cori-haswell


ndkeen:
After the Cori upgrade, I had trouble with GNU. In fact, all GNU tests failed immediately with a runtime error that was eventually attributed to the hugepages module. NERSC added this module by default, and we now remove it.  After the PR to remove hugepages, I ran e3sm_developer with GNU and some tests passed, but most failed. I think most or all fail with 
` Program received signal SIGILL: Illegal instruction.` 

Example of how to reproduce (I just tried with master as of Oct23rd) is: 
`ERS.f09_g16.I1850CLM45CN --compiler=gnu`

Looking back at e3sm_developer test after the PR to remove hugepages, the only tests that passed were:

```
ERS.f09_g16_g.MALISIA.cori-knl_gnu
ERS.f19_g16_rx1.A.cori-knl_gnu
ERS.ne30_g16_rx1.A.cori-knl_gnu
ERS_Ld5.T62_oQU120.CMPASO-NYF.cori-knl_gnu
SEQ.f19_g16.X.cori-knl_gnu
SMS.ne30_f19_g16_rx1.A.cori-knl_gnu
```

Trying with DEBUG=TRUE has same failure and I *do* get a stack trace (below)
I also tried using only 1 thread -- same fail.
Same error using gnu 8.3.0 (currently we use 8.2.0) and higher version of cray-libsci.
Same error on cori-haswell with gnu 8.2.0.

However, with gnu 8.3.0 (as well as higher version of cray-libsci), the test works on cori-haswell.

Stack from DEBUG runs:
```
29: Program received signal SIGILL: Illegal instruction.
29: 
29: Backtrace for this error:
29: 
29: Program received signal SIGILL: Illegal instruction.
29: 
29: Backtrace for this error:
29: #0  0x2aaab21ea1df in ???
29: #0  0x2aaab21ea1df in ???
29: #1  0xa2934a in __soiltemperaturemod_MOD_setmatrix_snownonurban
29: #1  0xa2934a in __soiltemperaturemod_MOD_setmatrix_snownonurban
29:     at /global/cscratch1/sd/ndk/wacmy/m17-oct23/components/clm/src/biogeophys/SoilTemperatureMod.F90:3729
29:     at /global/cscratch1/sd/ndk/wacmy/m17-oct23/components/clm/src/biogeophys/SoilTemperatureMod.F90:3729
[...]

PR #2948, opened 2019-05-28
Fix the setting of ocn_c2_glcshelf based on value of land_ice_flux_mode

jonbob:
This PR fixes the value of ocn_c2_glcshelf that is passed from mpas-o to the coupler and used to determine whether or not the coupler computes ice sheet melt fluxes. It is now based on all possible values of land_ice_flux_mode and fixes issues with ocean ISMF compsets.

Fixes #2947
[BFB]

PR #2726, opened 2019-02-06
Add first stage of ocean/ice shelf coupling

matthewhoffman:
This merge introduces coupler changes to allow coupling between the ocean model and ice shelves within the GLC model.

As part of model coupling necessary for Antarctic simulations, ice shelf melt fluxes due to (primarily) ocean temperatures must be calculated.  These melt fluxes correspond to oceanic heat and moisture fluxes at the upper ocean surface, and ice sheet heat and ice column thickness changes.  Coupling is characterized by 1) calculations carried out by ‘subroutine compute_melt_fluxes’ on the GLC grid, but 2) on the OCN coupling timestep (<GLC coupling timestep).  This is why these fluxes cannot be calculated in either component and are done here in the coupler itself.  These factors result in 1) frequent OCN2GLC remapping that is not associated with a GLC run step and 2) accumulation of GLC-bound ‘subroutine compute_melt_fluxes’ output fields, but no equivalent accumulation of OCN-bound outputs.

Inputs to the flux calculations (all mapped to GLC grid):
From OCN:
*   oceanTemperature(:) -> o2x_gx%rAttr(index_o2x_So_blt,:)
*  oceanSalinity(:) -> o2x_gx(1)%rAttr(index_o2x_So_bls,:)
*   oceanHeatTransferVelocity(:) -> o2x_gx(1)%rAttr(index_o2x_So_htv,:) 
*   oceanSaltTransferVelocity(:) -> o2x_gx(1)%rAttr(index_o2x_So_hsv,:)

From GLC:
*   interfacePressure(:) -> o2x_gx(1)%rAttr(index_o2x_So_phieff,:)
*   iceTemperature(:) -> g2x_gx%rAttr(index_g2x_Sg_tbot,:)
*   iceTemperatureDistance(:) -> g2x_gx%rAttr(index_g2x_Sg_dztbot,:)
*   iceFloatingMask(:) -> g2x_gx%rAttr(index_g2x_Sg_icemask_floating,:)

Outputs from flux calculations (all calculated natively on GLC grid):
To OCN:
*   g2x_gx%rAttr(index_g2x_Sg_blis,:) -> outInterfaceSalinity(:)
*   g2x_gx%rAttr(index_g2x_Sg_blit,:) -> outInterfaceTemperature(:) 
*   g2x_gx%rAttr(index_g2x_Fogx_qiceho,:) -> outOceanHeatFlux(:) 
[...]

PR #3513, opened 2020-03-24
Update maint1.2: add 3D varying GM 

mark-petersen:
This update adds the following PRs from MPAS ocean/develop:

-    Adds options for 3d varying GM bolus and 2d varying phase speed https://github.com/MPAS-Dev/MPAS-Model/pull/288
-    Add GM bolus eddy stats https://github.com/MPAS-Dev/MPAS-Model/pull/339
-    Fix threading issue in MPAS-O GM routine https://github.com/MPAS-Dev/MPAS-Model/pull/376
-    compute landIceMask using geometric_features https://github.com/MPAS-Dev/MPAS-Model/pull/447
-    bug fixes for the nonlocal source term in KPP https://github.com/MPAS-Dev/MPAS-Model/pull/305

and from E3SM master:
- Change all ocean output files to single precision #3360

[non-BFB]

### DIFF
--- a/components/mpas-ocean/cime_config/buildnml
+++ b/components/mpas-ocean/cime_config/buildnml
@@ -358,6 +358,7 @@ def buildnml(case, caseroot, compname):
         lines.append('')
         lines.append('<stream name="output"')
         lines.append('        type="output"')
+        lines.append('        precision="single"')
         if ocn_grid.startswith("oRRS1"):
             lines.append('        io_type="pnetcdf,cdf5"')
         else:
@@ -637,6 +638,7 @@ def buildnml(case, caseroot, compname):
         lines.append('')
         lines.append('<stream name="eddyProductVariablesOutput"')
         lines.append('        type="output"')
+        lines.append('        precision="single"')
         if ocn_grid.startswith("oRRS1"):
             lines.append('        io_type="pnetcdf,cdf5"')
         else:
@@ -659,6 +661,7 @@ def buildnml(case, caseroot, compname):
         lines.append('')
         lines.append('<stream name="highFrequencyOutput"')
         lines.append('        type="output"')
+        lines.append('        precision="single"')
         lines.append('        io_type="{}"'.format(ocn_pio_typename))
         lines.append('        filename_template="mpaso.hist.am.highFrequencyOutput.$Y-$M-$D_$h.$m.$s.nc"')
         lines.append('        filename_interval="00-01-00_00:00:00"')
@@ -706,6 +709,7 @@ def buildnml(case, caseroot, compname):
         lines.append('')
         lines.append('<stream name="mixedLayerDepthsOutput"')
         lines.append('        type="output"')
+        lines.append('        precision="single"')
         lines.append('        io_type="{}"'.format(ocn_pio_typename))
         lines.append('        filename_template="mpaso.hist.am.mixedLayerDepths.$Y-$M-$D.nc"')
         lines.append('        filename_interval="00-01-00_00:00:00"')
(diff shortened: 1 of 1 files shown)


mark-petersen:
@jonbob This needs the additional 5 GM flags noted here:
https://github.com/MPAS-Dev/MPAS-Model/pull/487#pullrequestreview-380414560
I already added the single precision output and the `land_ice_flux_mode`. I think that is everything, but I only tested in stand-alone, so we shall see...

jonbob:
Thanks @mark-petersen - I'll run the scripts to auto-generate the corresponding bid changes and commit them to your branch

jonbob:
@mark-petersen - I did a test merge and some testing. I don't think we want that last E3SM PR:
* Fix the setting of ocn_c2_glcshelf based on value of land_ice_flux_mode #2948

The corresponding changes in the cpl aren't there, and it requires more changes to the ocean driver

mark-petersen:
@jonbob I took off that last commit that added #2948. Let me know now that goes.

xylar:
@darincomeau, and @matthewhoffman, will melt rates work as expected without #2948?

matthewhoffman:
@xylar , I don't recall.  @jonbob , I don't remember the details of #2948 , but the concerning thing is that without it, if `config_land_ice_flux_mode=='standalone'` (which is what we currently use to get ISMF calculated, as far as I recall), then `ocn_c2_glcshelf` will be set to `.true.`.  Presumably that will trigger the parts of the coupler code you are referring to as not being in this branch.

jonbob:
I think the larger PR that [#2948](https://github.com/E3SM-Project/E3SM/pull/2948) was fixing was one of Jeremy's ([#2726](https://github.com/E3SM-Project/E3SM/pull/2726)) that @matthewhoffman and I reworked. We would have to integrate both of them into maint-1.2 to get that functionality

jonbob:
@mark-petersen - my initial tests pass now, with that last commit removed. My inclination is that we should merge this to maint-1.2 and decide separately about the land_ice_flux_mode, which could come in as a different PR. @matthewhoffman , @xylar - is that OK with you?

xylar:
@jonbob, yep, that sounds fine to me.  I just want us to make sure that melt fluxes work as expected before we get too far running with this branch.

mark-petersen:
Since it is just us using maint-1.2, I think it is OK to merge this and then make more corrections or additions next week. 

@jonbob is worthwhile requesting that the software engineering team run the E3SM regression suite on maint-1.2 after this (what they normally do for `next`)? So we get the full suite of tests?

jonbob:
I think it's easier if we run whatever tests we think are important. I can run a suite of test if you think we need to

darincomeau:
Is it too late to add https://github.com/E3SM-Project/E3SM/pull/3516, which was just opened?

jonbob:
No, never too late!

jonbob:
@darincomeau - though that should probably be a separate PR into maint-1.2, since it's not at all related to this mpas one

jonbob:
@mark-petersen - I ran at least some small fully-coupled tests. Since this PR will change answers, we're only really going to be able to pass smoke tests... We could try some ERS tests, just to make sure restarts are still BFB

darincomeau:
> @darincomeau - though that should probably be a separate PR into maint-1.2, since it's not at all related to this mpas one

That occurred to me, but I saw E3SM master commits in as well. But a separate PR works.

jonbob:
Passed:
* ERS.T62_oEC60to30v3.GMPAS-IAF.anvil_intel
* SMS.ne4_oQU240.A_WCYCL1850.anvil_intel

@mark-petersen - it's your PR, so please weigh in on the amount of testing required

jonbob:
@ndkeen has an open issue about GNU being broken on cori ([#3270](https://github.com/E3SM-Project/E3SM/issues/3270))

jonbob:
We'll run a longer test after merging

jonbob:
merged to maint-1.2

mark-petersen:
@jonbob thanks for merging, I didn't check email over the weekend. It sounds like we need to add #3516 on a separate PR to maint-1.2?

jonbob:
@mark-petersen - no, I'm ahead of you on that one! Please see [#3517](https://github.com/E3SM-Project/E3SM/pull/3517)

jonbob:
@mark-petersen - however, we do need to decide whether or not to include Jeremy's work from  [PR #2726](https://github.com/E3SM-Project/E3SM/pull/2726) and [PR #2948](https://github.com/E3SM-Project/E3SM/pull/2948)

PR #3516, opened 2020-03-26
For NERSC machines, change directory names that reference "acme" to "e3sm"

ndkeen:
For NERSC machines, change directory names that reference "acme" to "e3sm".

This is after the repo name change has been completed.

[bfb]


ndkeen:
merged to next

PR #3522, opened 2020-03-27
Fix ARKODE-HOMME interface bugs

asteyer:
This fixes a bug in the Jacobian calculation in the ARKODE-HOMME interface and adds ifndef statements to avoid calling the limiter and applying warnings when ARKODE is used.


oksanaguba:
If this functionality is needed, should there be a test? not in nightlies, but something that arcode team can use when homme is changed again.

rljacob:
@asteyer see above question about testing.

asteyer:
@oksanaguba The issue that arose was that a change in the eos code in HOMME broke arkode -  with arkode on, HOMME would not even compile.  What kind of test could ensure that this would not happen?

oksanaguba:
I was thinking of a test that would use ARCODE, at least to bisect some possible changes when homme is modified again. However none (?) of us user ARCODE, so, it may be overkill.

asteyer:
OK, I will work on a test that will fail if the arkode Jacobian hasn't been updated properly.

rljacob:
@asteyer how is the test coming?

rljacob:
@asteyer this PR is held up waiting for you to add a test.

rljacob:
@asteyer still need a test.

asteyer:
@oksanaguba Any suggestions on where to put the test and how it should be run?  Running it with the other tests would require installation of sundials on every machine where homme is tested which doesn't make sense since only 1 or 2 people will every use the interface.  On the other hand, putting a test in the source code with the Jacobian unit test would not help prevent the issue this PR was meant to address.

oksanaguba:
A test can be in homme but turned off by default (based on a directive from cmake that controls sundials build). The test should not be used in nightlies. However if you never build with arkode, then it is too much effort to make a test for it. As to how to do it exactly, I could commit some template for the test. Maybe we should ask @mt5555 , should we do a test in this PR?

rljacob:
Can the test go in the homme_integration suite?  That is only run on sandiatoss3.

oksanaguba:
This is about running homme with some external library for experimental features for external collaborators -- we do not really want to support it in nightlies.

rljacob:
In that case I withdraw my request for a test.  Go ahead and start merging this.

oksanaguba:
copy of this PR was created in PR #3770  . I wasn't able to merge this PR or to merge master into it, I assume previous master merge into this branch did not handle cime properly.

PR #3525, opened 2020-03-30
Set io_type to ocn_pio_typename for all streams that were not set

jonbob:
This PR modifies the script that builds the streams.ocean file. It changes the io_type to be ocn_pio_typename for all streams that were previously unset. This fixes problems on platforms that do not have pnetcdf built.

[BFB]

### DIFF
--- a/components/mpas-ocean/cime_config/buildnml
+++ b/components/mpas-ocean/cime_config/buildnml
@@ -460,6 +460,7 @@ def buildnml(case, caseroot, compname):
         if ocn_forcing == 'datm_forced_restoring':
             lines.append('<stream name="surface_salinity_monthly_data"')
             lines.append('       type="input"')
+            lines.append('       io_type="{}"'.format(ocn_pio_typename))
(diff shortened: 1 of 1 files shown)


jonbob:
@mark-petersen - it is set from an entry in env_run.xml, but it's difficult to find with grep because it has a compclass prefix. That section of the xml file looks like:
```
    <entry id="PIO_TYPENAME">
      <type>char</type>
      <valid_values>netcdf,pnetcdf,nothing</valid_values>
      <desc>pio io type</desc>
      <values>
        <value compclass="ATM">pnetcdf</value>
        <value compclass="CPL">pnetcdf</value>
        <value compclass="OCN">pnetcdf</value>
        <value compclass="WAV">pnetcdf</value>
        <value compclass="GLC">pnetcdf</value>
        <value compclass="ICE">pnetcdf</value>
        <value compclass="ROF">pnetcdf</value>
        <value compclass="LND">pnetcdf</value>
        <value compclass="ESP">pnetcdf</value>
        <value compclass="IAC">pnetcdf</value>
      </values>
    </entry>```

jonbob:
merged to next

jonbob:
merged to master

PR #3526, opened 2020-03-31
Adds wsubmin namelist parameter to gustiness use case

singhbalwinder:
Vince Larson made us aware that gustiness parameterization is missing
wsubmin in the use case file. This PR fixes that. It also changes
the test mod directory to add a shell command to use the use case file
for the gustiness test rather than mentioning namelist parameters
explicitly.

[Non-BFB] - Non Bit-For-Bit
[NML] - Namelist Changing

See confluence for a more detailed description about these tags.


wlin7:
@singhbalwinder , is it true that only the gustiness parameterization test currently uses this use_case file? I don't see an existing compset using it.

singhbalwinder:
Yes, it is only the gustiness test which uses this compset. Only that test is expected to fail.

rljacob:
@wlin7 what is the status of this PR?

wlin7:
Merged to next. Expect cam-gust_param tests to diff.

wlin7:
Merged to master. 

Request to bless ERS_Ld5.ne4_ne4.FC5AV1C-L.sandiatoss3_intel.cam-gust_param.
compy's cam-gust_param test has been failing before this PR. Still need to be bless after the pre-existing issue is resolved.

PR #3527, opened 2020-03-31
For NERSC machines, edit directory names to reflect the repo name change from "acme" to "e3sm"

ndkeen:
For NERSC machines, edit directory names to reflect the repo name change from "acme" to "e3sm".
Also remove the machine entry for edison as it is retired.

[bfb]


ndkeen:
merged

Issue #3536, opened 2020-04-03
usr_mech_infile option doesn't build chemistry preprocessor on cori


tangq:
The CAM_CONFIG_OPTS option "-usr_mech_infile" is really useful for testing new chemistry mechanisms and it is much needed for the NGD AtmPhys tasks of new chemistry mechanism development.

It worked in July 2019 as documented [here](https://github.com/E3SM-Project/E3SM/pull/3045#issuecomment-511054628) to recreate the E3SMv1 chemistry code, but it doesn't work on cori with the current master.

It complains about missing the tar output from the pre-processor when building the namelists. I tracked down the error and found that the chemistry pre-processor failed to build with components/cam/chem_proc/src/Makefile (COMPILE=pgf95). However, my offline build with components/cam/chem_proc/src/make_chempp.cori (COMPILER=gfortran) is successful.

The namelist seems created by a perl script (components/cam/bld/configure). I don't know perl enough to fix it.

@singhbalwinder , you used it successfully last year as mentioned above. Do you have any ideas why it fails now? Do you have a fix for it? It seems to me that we need to use make_chempp.cori instead of Makefile on cori, but I am not sure if that will cause problems on other machines. 

@wlin7 , I know you use perl. Do you have any suggestions?

Thanks.

singhbalwinder:
It was working fine on Cori at some point (like you mentioned), I am not sure what has changed between then and now. There is no test for this preprocessor, therefore there is no way to know when this functionality breaks (due to machine changes or the code changes).

I will have to look at the code to see what went wrong. The Makefile defaults to choosing pgf95 as the default compiler. As a temporary fix, I sometimes replace "pgf95" with "ifort", when I am using Intel compiler to compile the code. It might work on Cori. On Compy, I recently just made the following change to make it work:

```
ifeq ($(UNAMES),Linux)
  COMPILER := pgf95
endif
```
to:
``` 
ifeq ($(UNAMES),Linux)
  COMPILER := ifort
#pgf95
endif
```

We should find a permanent solution to fix this in the code.

wlin7:
@tangq , in cam/bld/configure file, indeed the only thing it does is to specify a compiler (then call chem_preprocess routine, which is in another perl module).

If you think gfortran will do it, you can temporarily change line 1408 the configure file (current master) to see if it  works

 `($chem_nadv) = chem_preprocess($cfg_ref,$print,"gfortran");`

I can update the configure file later in a more appropriate way.

[...]

PR #3537, opened 2020-04-04
Use the same compiler for chemistry pre-processor and the model

tangq:
Following @wlin7 's suggestion, this PR solves the issue https://github.com/E3SM-Project/E3SM/issues/3536 to use the same compiler for the chemistry pre-processor (PP) and the model.

This PR also makes it easier to implement a routine test for the PP.

The tests are successful on both cori and compy. But the log file (MAKE.out) is only saved on compy. Need to add some changes to this PR to save MAKE.out on all machines.

### DIFF
--- a/cime/config/e3sm/tests.py
+++ b/cime/config/e3sm/tests.py
@@ -59,6 +59,7 @@ _TESTS = {
             "PET_Ln5.ne4_ne4.FC5AV1C-L.allactive-mach-pet",
             "PEM_Ln5.ne4_ne4.FC5AV1C-L",
             "SMS_D_Ln5.ne4_ne4.FC5AV1C-L.cam-cosplite_nhtfrq5",
+            "SMS_Ln1.ne4_ne4.FC5AV1C-L.cam-chem_pp",
             "ERS_Ld5.ne4_ne4.FC5AV1C-L.cam-rrtmgp",
             "ERS_Ld5.ne4_ne4.FC5AV1C-L.cam-gust_param",
             "REP_Ln5.ne4_ne4.FC5AV1C-L",
--- a/components/cam/bld/perl5lib/Build/ChemPreprocess.pm
+++ b/components/cam/bld/perl5lib/Build/ChemPreprocess.pm
@@ -125,6 +125,11 @@ sub chem_preprocess
 		chmod 0555, "$chem_preprocessor/$chem_proc_exe";
 		if ($print) { print "creating $chem_preprocessor/$chem_proc_exe \n"; }
 	    }
+	} else {
+          my $log_file = "$chem_proc_bld/MAKE.out";
+          open OUT,">> $log_file"  || die "Failed to open $log_file\n";
+          print OUT  "Use pre-built chem-preprocessor: $proc_exe_path\n";
+          close OUT;
 	}
 
 	run_chem_preproc($chem_proc_bld,$proc_exe_path,$chem_mech_infile,$chem_proc_src,$cam_bld);
--- a/components/cam/chem_proc/inputs/pp_linoz_mam4_resus_mom_soag.in
+++ b/components/cam/chem_proc/inputs/pp_linoz_mam4_resus_mom_soag.in
@@ -1,12 +1,3 @@
-BEGSIM
-output_unit_number = 7
-output_file        = linoz_mam4_resus_mom_soag.doc
-procout_path       = ../output/
-src_path           = ../bkend/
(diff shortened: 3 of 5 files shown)


tangq:
Trying to understand components/cam/bld/perl5lib/Build/ChemPreprocess.pm. It seems that lines 345-352 (pasted below) should create the pre-processor building log file MAKE.out and leave it there no matter the building is successful or not.

I am confused why MAKE.out is not there in my tests...
```
    my $log_file = "$chem_proc_bld/MAKE.out";
    my $makefile = "$chem_proc_src/src/Makefile";
    my $cmd = "$gmake -f $makefile > $log_file 2>&1";
    run_shell_command($cmd) or die "Failed: $cmd";

    my $cmd = "rm -f $chem_proc_bld/obj/*.o $chem_proc_bld/obj/*.mod $chem_proc_bld/../*.mod";
    run_shell_command($cmd);

```

tangq:
I ran components/cam/bld/configure with `--verbose`. The log file is at compy:/compyfs/tang338/E3SM_simulations/tst/case_scripts/test/log

I hope it's helpful to you, @wlin7 and @singhbalwinder .

wlin7:
Hi @tangq , as you indicated, there is no mechanism in ChemPreprocess.pm that does the removal of MAKE.out. I did a test on cori using your branch, after case setup, build and submit, MAKE.out remained under ~wlin/ACME/Cases/testing/cori-knl/testUChem2/Buildconf/camconf/chem_proc. The full setup script I used is~wlin//ACME/Cases/testing/scripts/setup.testUChem2.csh.  I wonder what causes the difference between your test and mine. Can you share your full test procedure?

tangq:
@wlin7 , I used this script ~tang30/E3SM_scripts/tst_usr_mech_infile.csh and stopped the script when it starts building the model. Our steps should be essentially the same. I am trying to use your script now.

tangq:
Hi @wlin7 , when I use your script, the MAKE.out file is stored. I am trying to understand the differences...

tangq:
I am lost... @wlin7 , I manually ran the commands in your script and completed `./case.setup --verbose`. In this manual test. The MAKE.out file is not stored.

Compared to the case which stores MAKE.out, I noticed that the executable file "campp" is stored with MAKE.out, while campp is not stored when MAKE.out is not. Hopefully, this can give us some clue.

When I only change the case name in your script and expect a MAKE.out file, the test doesn't store it. This kind of random behavior is weird. Tests are:

1. Make.out exist: script (~tang30/E3SM_scripts/setup.testUChem2.csh), $casedir=/global/cscratch1/sd/tang30/E3SM_simulations/testUChem2/

2. Make.out does NOT exist: script (~tang30/E3SM_scripts/setup.testUChem3.csh), $casedir=/global/cscratch1/sd/tang30/E3SM_simulations/testUChem3/

wlin7:
Nice job, @tangq . Very true build_chem_preproc is only called when campp does not exist. This would be the case when a new repo (or branch) is used to create the case. I will update the perl module to save such info on MAKE.out.

wlin7:
It is done and tested, @tangq . When using the same repo, the 1st test built campp and saves the full compilation log to MAKE.out. All follow-up tests, with pre-built campp present, a simple line is recorded to MAKE.out indicating using pre-built campp with the fullpath to the executable.

singhbalwinder:
I think @wlin7 and I were working at the same time. I have merged @wlin7 changes into mine and push my changes which adds a new test to e3sm_atm_integration test suite.

tangq:
Thank you, @wlin7 and @singhbalwinder for the prompt work!

@singhbalwinder , I added an inline version of DECKv1 chemistry input file (pp_linoz_mam4_resus_mom_soag.in). It seems that the convention is that the offline input files are *.inp, while inline ones are *.in.

Can you change the test to use this inline input file directly? I see 2 benefits: 1) more direct steps, so less confusing; 2) the *.in file will be used for new inline mechanism development. When finalized, we can update the test to use the same *.in file.

With this last change, I think the PR is good to go.

singhbalwinder:
Thanks @tangq ! It indeed made it simpler and less confusing.

wlin7:
Merged to next.

wlin7:
Merged to master. Requests submitted to create baseline for
 
```
SMS_Ln1.ne4_ne4.FC5AV1C-L.compy_pgi.cam-chem_pp
SMS_Ln1.ne4_ne4.FC5AV1C-L.sandiatoss3_intel.cam-chem_pp
```

PR #3540, opened 2020-04-07
template.case.run.sh: improve robustness

jgfouca:
Setup environment before doing python stuff in case env is needed
for python.

Removing the timing tree should be recursive.

[BFB]


jgfouca:
@sbrus89

Issue #3538, opened 2020-04-06
setting PCOLS at runtime (discussion)


worleyph:
Recent experiments indicate that the performance of the physics in the
current version of EAM can be improved in certain circumstances by
optimizing the choice of the declared size of the first dimension
(pcols) in the basic atmospheric physics data structure (chunk). (See
Confluence page https://acme-climate.atlassian.net/wiki/x/wAZ_QQ .)
While this was the case when the design was first implemented in the
model in the early 2000's, the analysis then was that this had more to
do with processor architecture (cache size or vector vs. RISC) than
problem granularity (mesh resolution and degree of parallelism), and
that a default value of pcols=16 was pretty good for most non-vector
processors. The recent experiments instead indicate that setting pcols
to minimize wasted space and, perhaps, the number of chunks,
especially for the PG2 meshes and at scale, can improve upon the
performance of the default sufficiently to make it worthwhile to start
optimizing over pcols again, and to do so on the basis of problem
granularity (as well as for processor architecture).

The other result from the original design experiments was that setting
pcols at compile-time allowed the compilers (at the time) to generate
more efficient code, leading to better performance than
otherwise. This is the current design, and changing pcols requires
a clean build of EAM. However, the code base has evolved significantly
since the original experiments were conducted, with some data 
structures (still chunk-based) now using runtime-calculated first
dimension size declaration, one example being psetcols. psetcols (and
the other examples) is still determined by compile-time parameters
(pcols and psubcols) but this may not be exploitable by the
compiler. Compilers and the target processor architectures have also
evolved over the last twenty years, so it is unclear as to whether the
original empirical results are even relevant to current compilers and
target processors.

With pcols being a compile-time parameter, optimization studies must
include a rebuild in every experiment to determine the optimal setting
for a given compset, mesh resolution, PE layout, and target
system. This increases the effort (both turnaround time and amount of
manual interaction) required for these studies compared to what would
be required if pcols could be specified at runtime. There is also a
desire to develop a better default value than 16, taking into account
the mesh resolution and PE layout, for example. See Github issue
#3485  ("Automated selection of pcols in EAM"). With the current
compile-time dependency, any "smart" setting of pcols must occur in
the CIME layer, which will be more restricted than what can take place
during runtime, when more is knowable about the particular case. With
a runtime setting of pcols, we could also allow pcols to vary between
the processes. I am not sure that this would be all that useful, but
when load balancing is restricted, e.g. phys_loadbalance=0, or when
[...]

PR #3545, opened 2020-04-13
Implement pcols as a runtime option in EAM

worleyph:
In the earlier 2000's, empirical experiments indicated that setting
the first dimension of the chunk data structure (pcols) at
compile-time allowed the compiler to generate more
performance-efficient code, and this motivated this design choice in
CAM. However, CAM/EAM has evolved significantly since then, and, for
example, physics_types now uses psetcols to define this first
dimension, and the compiler never sees a compile-time parameter. So
the compile-time setting of pcols may no longer serve a useful
purpose. Extensive experiments indicate that there is still a
small performance advantage from setting pcols at compile-time, but
also that being able to set pcols at runtime greatly simplifies the
process of optimizing over the pcols parameter.

Here pcols is implemented as a runtime option, currently specified
via the user_nl_cam namelist variable phys_chnk_fdim. phys_chnk_fdim
must be at least as large as 1, but there is no specified upper
bound. The current compile-time value of the CPP variable PCOLS is used
to define the default value for phys_chunk_fdim (and pcols). PCOLS is
now represented in EAM as the parameter ppcols, but is only referenced	
in defining the default for phys_chunk_fdim.

To restore the compile-time setting of pcols, and also to specify
a pcols value of PCOLS, add -pcols <PCOLS> to CAM_CONFIG_OPTS in
env_build.xml . For example, to specify 32 as the compile-time value
of pcols, add '-pcols 32' . This allows the recovery of any lost
performance from the runtime-specification of pcols.

Note that this is the first part of a two-part introduction of runtime
pcols. This PR sets the runtime pcols during the call to
read_namelist. To calculate a better default than the compile-time
PCOLS CPP variable requires that it be calculated in phys_grid_init,
and this requires changing the order of certain steps in the
initialization of EAM. The current PR, however, stands on its own, and
is acceptable even if the second part PR is never submitted.

BFB

This addresses some aspects of Issue #3538 , but not all. The issue
will be resolved by a second runtime pcols PR. If we decide
not to submit this follow-on PR, the issue will also be closed.

### DIFF
--- a/components/cam/bld/config_files/definition.xml
+++ b/components/cam/bld/config_files/definition.xml
@@ -211,7 +211,7 @@ Maximum number of sub-columns in a column (physics data structure).
 </entry>
 <entry id="ppcols" value="FALSE">
 Whether pcols is a compile time parameter. Otherwise it is the default value
-for the runtime parameter phys_chunk_fdim.
+for the runtime parameter phys_chnk_fdim.
 </entry>
 <entry id="cam_exe" value="cam">
 Name of CAM executable.
--- a/components/cam/bld/config_files/definition.xml
+++ b/components/cam/bld/config_files/definition.xml
@@ -209,6 +209,10 @@ Maximum number of columns in a chunk (physics data structure).
 <entry id="psubcols" value="1">
 Maximum number of sub-columns in a column (physics data structure).
 </entry>
+<entry id="ppcols" value="FALSE">
+Whether pcols is a compile time parameter. Otherwise it is the default value
+for the runtime parameter phys_chunk_fdim.
+</entry>
 <entry id="cam_exe" value="cam">
 Name of CAM executable.
 </entry>
--- a/components/cam/bld/configure
+++ b/components/cam/bld/configure
@@ -1294,16 +1294,21 @@ set_horiz_grid("$cfgdir/$horiz_grid_file", $cfg_ref);
 if ($print>=2) { print "Horizontal grid specifier: $hgrid$eol"; }
 
 #-----------------------------------------------------------------------------------------------
-# Maximum number of columns in a chunk.
+# Maximum number of columns in a chunk. If set as an option, then it is a compile-time parameter
+# (ppcols == 'TRUE'). Otherwise, pcols is just the default value for a runtime parameter.
 if (defined $opts{'pcols'}) {
     $cfg_ref->set('pcols', $opts{'pcols'});
+    $cfg_ref->set('ppcols', 'TRUE');
+} else {
+    $cfg_ref->set('ppcols', 'FALSE');
 }
 my $pcols = $cfg_ref->get('pcols');
 
-# Override PCOLS setting if configuring for SCAM
+# Override PCOLS setting if configuring for SCAM. Also make it a compile-time parameter.
 if ($scam eq 'ON') {
     $pcols = 1;
     $cfg_ref->set('pcols', $pcols);
+    $cfg_ref->set('ppcols', 'TRUE');
 }
 
 # Check valid value of pcols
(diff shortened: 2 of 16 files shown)


worleyph:
This has been tested on Compy and Cori-KNL using FC5AV1C-L and ne30_ne30 and ne30pg2_ne30pg2 for 675x1, 2700x1, and 5400x1 (Compy) and 2700x2 (Cori-KNL).
As the chemistry packages included in this compset required some modification to allow for the runtime setting of pcols, I am concerned that other packags may require similar modification. @singhbalwinder indicated that there are no E3SM-supported packages to be concerned about at the moment (though some are being worked on as part of NGD . 

I also tried aquaplanet - it worked, but I have not yet tried his other suggestion: -chem none - I'll look at this today.

worleyph:
Note that the current production runscripts used by @ambrad (and @golaz ?) add -pcols to CAM_CONFIG_OPTS in env_build.xml. This will continue to work, and will retain the full compile-time setting of pcols performance achieved in the current version of EAM.

worleyph:
@singhbalwinder , found a bug when using -chem none. I will push a fix as soon as I can (should be a simple fix).

worleyph:
Implemented, tested, and pushed fix - now working.

worleyph:
@ambrad , @amametjanov , @singhbalwinder , any chance that you could look at this? Thanks.

ambrad:
Sorry @worleyph; didn't know what to write about this. Looks fine if you've decided to go with runtime pcols. Tbc, just specifying -pcols at compile time satisfies the performance goal. I think others who want runtime pcols specification should weigh in and make the final decision.

worleyph:
Thanks @ambrad. If you don't feel that you can review this, I'll remove you as a reviewer.

worleyph:
@mt5555 , @sarats , @philipwjones , it's time to decide whether to move forward with this. I'm trying to guess what will be useful for future development, but can't make this decision myself. This current PR "does no harm" for production jobs in that it can still be run with an explicit compile-time setting of pcols, but most users won't bother with this and will get a runtime pcols, still pcols=16, so a slight performance degradation compared with not accepting this PR. Note that a user can always set pcols at runtime (in user_nl_cam) to try to improve performance, but it will still not be compile-time unless the user builds this with the compile-time option set.

The more future-looking change would be in the follow-on PR, that reorganizes initialization to allow pcols to set in phys_grid_init, and then calculates pcols based on case specifics when not directly specified (compile-time or runtime) and also allows it to vary between virtual SMPs. With this change, I can encode the "rules" for calculating pcols based on system and phys_grid option, for example. 

Note that pcols could also be set per process - which is what will happen for phys_loadbalance=0 - but could be generalized to do this in other situations as well. This does not make any sense with the current load balancing schemes though.

I'd like this to go in soon, or decide to drop this development soon, as the longer we wait the more complicated the merge will be and I have other work that will change depending on whether we proceed with this.

 I also don't know whether this is essentially dead code for future versions of the model or not. Is there any reason to continue to develop this part of EAM?

whannah1:
Not sure how much my opinion matters here, but I'm in favor of merging this. There's a chance that this feature will be overlooked by people who might otherwise benefit from it, but I think the enhanced flexibility is worth the 1-2% performance loss if people don't use the compile time option. That should be a reduction of ~20-40 sim-days-per-day for a 5 sypd throughput. I think that's acceptable.

singhbalwinder:
Hi @worleyph : I am also in favor going ahead with this as it will make it more flexible. Having this capability will allow folks who want to use this capability to experiment with this. The only downside I see is the impact on performance, which is also very minimal.

ndkeen:
I'm certainly happy with it being a compile-time option.

sarats:
Concur with approving this PR, the performance impact is minimal with the additional fallback of explicit compile-time pcols. It does pave the way for flexibility in exploring load balancing design and certain problems of interest in the ECP (E3SM-MMF) effort.

worleyph:
@amametjanov , if you don't have time to review this, or don't feel that you have a relevant opinion, I'll remove you as a reviewer. Please tell me what you want me to do. Thanks.

worleyph:
@amametjanov seems to be off-line. @sarats , adding you as a reviewer (so there is at least one reviewer other than the integrator). @singhbalwinder , please start the integration process as soon as Sarat approves the PR (assuming that he does :-) ). There has been a lot going on recently. I don't know if conflicts will pop up now or not, or whether there is code that I have not tested against that will break with these changes. Fortunately, any failures will be at build time.

worleyph:
Thanks @amametjanov . @sarats , since you already spoke up with an informal approval, I'll not ask you to put any more time into this, and am removing you from reviewer list. @singhbalwinder , this is ready to go. Thanks.

singhbalwinder:
Thanks @worleyph ! I will merge it as soon as next opens again.

worleyph:
Thanks, @sarats. There was a pattern to the substantive changes, so not particularly intellectually challenging. Just don't know whether I identified everything - I let the compiler tell me if there was a problem, so a different compset / CAM_CONFIG_FLAGS may introduce issues.

worleyph:
@singhbalwinder , is next open for this to be merged yet? Thanks.

singhbalwinder:
Merged to next.

PR #3546, opened 2020-04-13
Update mpas-source: Ocean initialization & testing

mark-petersen:
Large number of files, but all are within COMPASS (python initialization and testing) except a few files for MPAS-Ocean init mode. These are compiled by E3SM but not used for the forward model.

[BFB]


jonbob:
merged to next

jonbob:
passed  SMS.ne4_oQU240.A_WCYCL1850.anvil_intel sanity test

jonbob:
merged to master

PR #3551, opened 2020-04-20
Fix PIO xml case setup

### LINKED ISSUE #3550: PIO_TYPENAME setting only followed during testing.
Our config_pio.xml file has this block:
```
  <entry id="ATM_PIO_TYPENAME">
    <values>
      <value compset="DATM">netcdf</value>
    </values>
  </entry>
```
Any case that has datm for the atmosphere, e.g. any I or G case, should have PIO_TYPENAME for the atmosphere set to "netcdf". 

But this seems to only happen in the Jenkins-run testing.  Any attempt to build an I case or I test by hand results in a value of pnetcdf.   No idea why.

First noticed this when a namelist diff on compy showed a change from pnetcdf to netcdf.  Could not reproduce the setting of netcdf.   This was next  e932d0797.

We don't do a lot of testing on master.  Melvin has everything set to netcdf because its a workstation.

jgfouca:
Component-specific settings need to be set after generic
settings so that they take precedence.

Fixes #3550 

[BFB]

### DIFF
--- a/cime/scripts/lib/CIME/XML/pio.py
+++ b/cime/scripts/lib/CIME/XML/pio.py
@@ -5,6 +5,9 @@ from CIME.XML.standard_module_setup import *
 
 from CIME.XML.entry_id import EntryID
 from CIME.XML.files import Files
+
+from collections import OrderedDict
+
 logger = logging.getLogger(__name__)
 
 class PIO(EntryID):
@@ -19,7 +22,8 @@ class PIO(EntryID):
 
     def get_defaults(self, grid=None, compset=None, mach=None, compiler=None, mpilib=None): # pylint: disable=unused-argument
         # should we have a env_pio file
-        defaults = {}
+        defaults = OrderedDict()
+        save_for_last = []
 
         # Load args into attribute dict
         attributes = {}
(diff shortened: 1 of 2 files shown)


cime/scripts/lib/CIME/XML/pio.py
```diff
@@ -31,6 +35,15 @@ def get_defaults(self, grid=None, compset=None, mach=None, compiler=None, mpilib
         for node in self.get_children("entry"):
             value = self.get_default_value(node, attributes)
             if value:
-                defaults[self.get(node, "id")] = value
+                myid = self.get(node, "id")
+                iscompvar = myid.split("_")[0] != "PIO"
```
jgfouca, line 39: @jedwards4b , this seemed like a pretty weak way to check if the current node is component specific, but I could not see any other way to achieve this. Please comment if you know of a better way.

jedwards4b: For compvars you can have PIO_STRIDE_LND=4 or LND_PIO_STRIDE=4 and both will do the same thing.  This seems to break that capability and only allow PIO_ at the beginning of the name - use env_base.py check_if_compvar instead.  I think that it will also tell you whether this is specific to a component or a general setting.

rljacob:
This changed the PIO_TYPENAME for the homme tests.
```
  BASE: pio_typename = 'pnetcdf'
  COMP: pio_typename = 'netcdf'
```

Needs a bless request.

PR #3552, opened 2020-04-21
Fix RRTMGP for MMF

whannah1:
This fixes a bug in RRTMGP with the MMF configuration. The radiative fluxes were not multiplied by the pressure thickness. This caused the radiative heating to have the wrong units. The fix simply mirrors how radiative tendencies are treated for the non-MMF configuration with RRTMGP.

This PR also updates the MMF tests to ensure threading is disabled, and adds a test to the e3sm_integration suite. 

[BFB except for MMF tests]

### DIFF
--- a/cime/config/e3sm/tests.py
+++ b/cime/config/e3sm/tests.py
@@ -154,6 +154,7 @@ _TESTS = {
             "SMS.ne30_oECv3.BGCEXP_BCRC_CNPRDCTC_1850.clm-bgcexp",
             "SMS_D_Ld1.T62_oEC60to30v3.DTESTM",
             "SMS_D_Ld1.ne30_r05_oECv3.A_WCYCL1850",
+            "ERP_Ln9_P96x1.ne4pg2_ne4pg2.F-MMF1-TEST.cam-crmout",
             )
         },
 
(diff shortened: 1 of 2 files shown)


whannah1:
In case anyone is curious about the effect of this, I created a plot of the tests I was working with. The radiative tendencies were effectively scaled down to smaller values since we were not multiplying by the pressure thickness. Since both SW and LW were both messed up, the TOA net values still looked ok. The real clue was noticing that the precipitation was dramatically reduced due to the much weaker destabilization by radiative cooling (I think). The fix puts the results in line with what we get from RRTMG.  

image

whannah1:
@crjones-amath we found another problem, which is a bit confusing to explain. The plot I posted in a previous comment actually used a different branch based onthe first MMF branch, not rebased on E3SM master. There were a bunch of changes that got added in the mean time, and one of them messed up the cloud radiation stuff. Ben helped me figure out the fix and I just added a commit for it. It just bring the MMF version of radiation.F90 up to date with the non-MMF version. So you might want to take a quick look just to be thorough. I'm also running another couple tests.

crjones-amath:
@whannah1 Thanks for pinging me on this. This update looks good to me.

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #3557, opened 2020-04-28
Copy changes from standalone CIME to shr_mem_mod

jgfouca:
These changes somehow got missed in the CIME update.

[BFB]

### DIFF
--- a/cime/src/share/util/shr_mem_mod.F90
+++ b/cime/src/share/util/shr_mem_mod.F90
@@ -48,25 +48,25 @@ CONTAINS
     endif
 
     ierr = GPTLget_memusage (msize, mrss0, mshare, mtext, mdatastack)
-    allocate(mem_tmp(1024*1024*128))    ! 1 MWord, 1024 MB (128*8)
+    allocate(mem_tmp(1024*1024))    ! 1 MWord, 8 MB
     mem_tmp = -1.0
     ierr = GPTLget_memusage (msize, mrss1, mshare, mtext, mdatastack)
     deallocate(mem_tmp)
     ierr = GPTLget_memusage (msize, mrss2, mshare, mtext, mdatastack)
     mb_blk = 0.0_shr_kind_r8
     if (mrss1 - mrss0 > 0) then
-       mb_blk = (1024.0_shr_kind_r8)/((mrss1-mrss0)*1.0_shr_kind_r8)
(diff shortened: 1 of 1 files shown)


cime/src/share/util/shr_mem_mod.F90
```diff
@@ -48,25 +48,25 @@ subroutine shr_mem_init(prt, strbuf)
     endif
 
     ierr = GPTLget_memusage (msize, mrss0, mshare, mtext, mdatastack)
-    allocate(mem_tmp(1024*1024*128))    ! 1 MWord, 1024 MB (128*8)
+    allocate(mem_tmp(1024*1024))    ! 1 MWord, 8 MB
```
rljacob, line 51: The error code for the allocate statement should be checked.

cime/src/share/util/shr_mem_mod.F90
```diff
@@ -48,25 +48,25 @@ subroutine shr_mem_init(prt, strbuf)
...
+       mb_blk = (8.0_shr_kind_r8)/((mrss1-mrss0)*1.0_shr_kind_r8)
     endif
 
     if (lprt) then
-       write(s_logunit,'(A,f16.2)') '1 GB memory   alloc in MB is ',(mrss1-mrss0)*mb_blk
-       write(s_logunit,'(A,f16.2)') '1 GB memory dealloc in MB is ',(mrss1-mrss2)*mb_blk
-       write(s_logunit,'(A,f16.2)') 'Memory block size conversion in Kbytes is ',mb_blk*1024_shr_kind_r8
+       write(s_logunit,'(A,f16.2)') '8 MB memory   alloc in MB is ',(mrss1-mrss0)*mb_blk
+       write(s_logunit,'(A,f16.2)') '8 MB memory dealloc in MB is ',(mrss1-mrss2)*mb_blk
+       write(s_logunit,'(A,f16.2)') 'Memory block size conversion in bytes is ',mb_blk*1024_shr_kind_r8*1024.0_shr_kind_r8
```
rljacob, line 64: I think this is wrong.  The first 1024 should be modified to be 1024.0.   Not sure this should be 1024 squared.  @amametjanov ?

amametjanov: Updated. That seems ok: a value in MB is being divided by 1024 squared to get a value in bytes.

rljacob:
@jgfouca you need to re-merge this to next.

jgfouca:
@rljacob , done

PR #3559, opened 2020-04-29
Add script support for ne30_oARRM60to10

jonbob:
This PR brings in script support for fully-coupled cases using the ne30_oARRM60to10 grid resolution. It adds appropriate defaults for mpas-o and mpas-seaice, including spun-up initial conditions from a multi-year G-case run. The coupling frequency is set to a 15-minute default for all components except the ocean, which couples every 30-minutes.

All new mapping and domain files have been staged on the inputdata server.

Since this is a new resolution, all tested configurations are BFB.

[BFB]

### DIFF
--- a/cime/config/e3sm/config_grids.xml
+++ b/cime/config/e3sm/config_grids.xml
@@ -1711,6 +1711,13 @@
       <desc>oARRM60to10 is an Arctic-Region-Refined MPAS ocean grid with 30 km gridcells at -90 deg latitude, 60 km gridcells at -40 deg latitude, 30 km gridcells at the equator, and 10 km gridcells at 90 deg latitude; North Atlantic and North Pacific have different resolution between 0 and 60 deg latitudes:</desc>
     </domain>
 
+    <domain name="oARRM60to10_ICG">
+      <nx>619264</nx>
+      <ny>1</ny>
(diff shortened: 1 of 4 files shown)


stephenprice:
Thanks Jon. I will admit that I wouldn't necessarily know if there was something wrong here ... but with that caveat in mind, it all looks great to me. Thanks for the help!

jonbob:
Tested with one-year simulation on anvil and passed a smoke test to make sure it found all necessary data out-of-the-box:
SMS.ne30_oARRM60to10_ICG.A_WCYCL1850S.anvil_intel

jonbob:
@stephenprice - we won't know if there are any real issues until we run several years with it. But I think all the pieces are in place to begin that exercise. I'll commit it to maint-1.2 once you review it.

stephenprice:
@jonbob -- consider it 'reviewed', unless by that you mean having us run it out for a bit first. If so, happy to do that from this branch.

jonbob:
@stephenprice - if you have time, could you please officially approve this? There should be a button to add your review... But if not, just let me know and I'll go ahead and merge it without that

jonbob:
Thanks! I'll get it in today so we can crank up a real run whenever we decide it's time

jonbob:
Merged to maint-1.2

PR #3560, opened 2020-04-29
update mpas-source: fix incorrect path for ocean cvmix

mark-petersen:
When changing to submodules for the ocean cvmix package, the path to cvmix source had changed, but one instance of this path was not changed, leading to build failures with PGI on Summit. This PR fixes that bad path. See https://github.com/MPAS-Dev/MPAS-Model/pull/494

[bfb]

PR #3561, opened 2020-04-29
update mpas-source: COMPASS only

mark-petersen:
Update COMPASS initialization scripts. Does not affect E3SM simulations.

[BFB]


mark-petersen:
The merge order needs to be: #3391, #3561, #3560.

mark-petersen:
Changed this one to be after #3560.

mark-petersen:
Rebased this branch, and updated mpas-source. It is still only COMPASS files, plus the one line change in #3560 so it can go in after that.

jonbob:
merged to next

jonbob:
merged to master -- but curious why it's not showing up here already

PR #3562, opened 2020-05-05
Enable COSP with RRTMGP

brhillman:
Enable running the CFMIP Observation Simulator Package (COSP) with RRTMGP. This provides a number of additional diagnostics that are directly comparable to satellite retrievals of cloud properties from a number of different instruments. To build with COSP, simply add `-cosp` to `CAM_CONFIG_OPTS` in `env_build.xml` (i.e., `./xmlchange --append CAM_CONFIG_OPTS='-cosp'`). No other action should need to be taken, and h0 output files should include a selection of COSP outputs. 

[BFB]

### DIFF
--- a/components/cam/src/physics/rrtmgp/cam_optics.F90
+++ b/components/cam/src/physics/rrtmgp/cam_optics.F90
@@ -66,30 +66,30 @@ contains
       integer :: iband, ilev, icol
 
       ! Initialize outputs
-      tau_out = 0
-      ssa_out = 0
-      asm_out = 0
-      liq_tau_out = 0
-      ice_tau_out = 0
-      snw_tau_out = 0
+      tau_out = 0._r8
+      ssa_out = 0._r8
+      asm_out = 0._r8
+      liq_tau_out = 0._r8
+      ice_tau_out = 0._r8
+      snw_tau_out = 0._r8
 
       ! Initialize local variables
-      ice_tau = 0
-      ice_tau_ssa = 0
-      ice_tau_ssa_g = 0
-      ice_tau_ssa_f = 0
-      liq_tau = 0
-      liq_tau_ssa = 0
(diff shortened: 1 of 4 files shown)


brhillman:
Comparison between a short test with RRTMGP and RRTMG can be found on confluence [here](https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/1407746072/COSP+in+RRTMGP).

components/cam/src/physics/rrtmgp/cam_optics.F90
```diff
@@ -63,7 +65,15 @@ subroutine get_cloud_optics_sw( &
 
       integer :: iband, ilev, icol
 
-      ! Initialize
+      ! Initialize outputs
```
wlin7, line 68: These initializations to 0s should following the practice of using 0.0_r8. Don't know whenotherwise it would become a problem, though. Similarly for the block starting from line 261.

brhillman:
@wlin7 confirmed this is BFB on cori-knl for the `e3sm_developer` suite.

components/cam/src/physics/rrtmgp/cam_optics.F90
brhillman: I think that using `0` is not a problem, but using `0.0` could be a problem because `0.0` would be interpreted as single precision. I can change them regardless, but using the integer literals shouldn't cause a problem here.

brhillman: Fixed on the branch, I'll push the commit after testing.

brhillman: Fixed.

wlin7:
Merged to next.

wlin7:
Merged to master.

PR #3568, opened 2020-05-12
Additional physgrid support files

whannah1:
Add mapping/domain/topo/drydep files for running with certain tri-grid+physgrid configurations, and removes support for "pg1" grids. The new grids are:
ne30pg3_r05_oECv3
ne30pg4_r05_oECv3
ne120pg2_r05_oECv3
conusx4v1_r05_oECv3
conusx4v1pg2_r05_oECv3

[BFB]

### DIFF
--- a/cime/config/e3sm/config_grids.xml
+++ b/cime/config/e3sm/config_grids.xml
@@ -1665,6 +1665,16 @@
       <mask>oEC60to30v3</mask>
     </model_grid>
 
+    <model_grid alias="ne30pg1_r05_oECv3">
+      <grid name="atm">ne30np4.pg1</grid>
+      <grid name="lnd">r05</grid>
+      <grid name="ocnice">oEC60to30v3</grid>
+      <grid name="rof">r05</grid>
+      <grid name="glc">null</grid>
+      <grid name="wav">null</grid>
+      <mask>oEC60to30v3</mask>
+    </model_grid>
+
     <model_grid alias="ne30pg2_r05_oECv3">
       <grid name="atm">ne30np4.pg2</grid>
       <grid name="lnd">r05</grid>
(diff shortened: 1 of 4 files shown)


cime/config/e3sm/config_grids.xml
```diff
@@ -1237,6 +1187,16 @@
       <mask>gx1v6</mask>
     </model_grid>
 
+    <model_grid alias="ne120pg2_r05_oECv3">
```
rljacob, line 1190: Normally the ne120 atmosphere is run with the oRRS18v3 ocean grid.  Is there a reason you are using oECv3?

whannah1: I don't know what people use. I mainly just wanted to do some ne120 hindcasts with fixed SST and not have to spin up a new r05 land initial condition. 
Should I just remove this grid?

rljacob: Yes go ahead.

rljacob: There's a ne120pg2_ne120pg2 grid you could use for fixed-SST runs.  Not sure about land initial conditions for that.

whannah1: I wanted a tri-grid case to not have to change the land.

rljacob: If this grid works for you, go ahead and keep it.

rljacob:
@whannah1 have all the new data files been uploaded to the server?

whannah1:
@rljacob I haven't been able to do this yet. There seems to be a problem with my Argonne collaborator account and I'm waiting to hear back from the help desk. Also, the simulations with these grids started crashing as soon as I start the runs, so I'm wondering if the mapping files became corrupted somehow. I need to dig back into this to make sure everything is working.

rljacob:
@whannah1 what is the status of this PR?

whannah1:
Still struggling with it, but the main grids I wanted to add are working fine for the most part. 

We also recently decided that we need to add the ne45pg2 to get the work per node that we need to use our INCITE time. ne30np4 and ne30pg3 are similar in terms of ncol, but using "pg2" just makes more sense in general, so I wanted to add that in to this PR. Unfortunately, I've been running tests all last week and working closely with @mt5555 and I still can't get a single run to get through the first time step. 

Because of the ne45 complications I put off adding the other files to blues, but if you prefer I can separate these two things, make a separate PR for ne45pg2 and try to get the files copied over for the grids in this PR.

rljacob:
Yes make a separate PR for ne45pg2 when its ready.  I'd like to get this one finished.

whannah1:
@rljacob I rebased to fix the conflict and uploaded all the files to blues, so I think this PR is ready to go. I added some tri-grid versions oof the CONUS grid, so I'll note that in the description.

rljacob:
merged to next.

Issue #3565, opened 2020-05-08
Improve the nudging implementation in EAM/E3SM 


kaizhangpnl:
The nudging implementation in E3SM (atmosphere model) has some problems:

- The original code requires the nudging data to have only one time slice per file. If the nudging file has multiple time slices, the model will read the first time slice only and give wrong results. 
- The original code can only use the step-function nudging. No interpolation is applied between two available time slices (6h). 
- There is a bug related to the intermittent nudging configuration.

Sun et al. (2019) improved the nudging implementation in E3SM and these problems have been fixed. We will port these changes to the master (as recommended by Chris for v2).

PR #3569, opened 2020-05-12
implement revised nudging code

jiansunpnnl:
We have implemented a revised version of nudging code in E3SM to fix the issue #3565. 

Changes include: 
1. add linear interpolation for nudging input data
2. flexible handling of nudging input data (e.g., multiple time slices per file; but do not support multiple-day data per file)
3. bugfixes
- reset the nudging tendency to zero for intermittent nudging (in nudging.F90)
- location of calculation of nudging tendency (in nudging.F90 and physpkg.F90)
- location of update of nudging tendency (in physpkg.F90, affect FV dycore only)
- duplicate namelist definitions (in namelist_definition.xml)
4. suggest to use the following default settings for the newly added namelist variables
- Nudge_Method = 'Linear'
- Nudge_Loc_Same = .TRUE.
- Nudge_Tau             = -999.
- Nudge_Curr           = .FALSE.
- Nudge_File_Ntime = 0
5. More details about these changes could be found in the comments of namelist_definition.xml and nudging.F90

    modified:   components/cam/bld/namelist_files/namelist_definition.xml
    modified:   components/cam/src/physics/cam/nudging.F90
    modified:   components/cam/src/physics/cam/physpkg.F90

Original development branch can be found here: 
https://github.com/E3SM-Project/E3SM/tree/jiansunpnnl/ndg_loc

The revised nudging implementation has been evaluated in Sun et al. (2019). 

[BFB] when nudging is not switched on 
[BFB] when nudging is switched on, but use nudge_method = step, nudge_loc_same = .false., nudge_tau = -999, nudge_file_ntime = 1, nudge_curr = .false. and one time slice per nudging input data file

### DIFF
--- a/components/cam/bld/namelist_files/namelist_definition.xml
+++ b/components/cam/bld/namelist_files/namelist_definition.xml
@@ -283,11 +283,11 @@
        group="nudging_nl" valid_values="" >
        Use user-defined relaxation time scale (unit: hour).
        If Nudge_Tau > 0, the relaxation time scale is determined by Nudge_Tau;
-       If not, the relaxation time scale is determined by Nudge_Times_Per_Day;
+       If not, the relaxation time scale is determined by Nudge_Times_Per_Day and nudging strength specified in the namelist (e.g. Nudge_Ucoef).
        Default: -999
        </entry>
 
-<entry id="Nudge_Loc" type="logical" category="nudging"
+<entry id="Nudge_Loc_Same" type="logical" category="nudging"
        group="nudging_nl" valid_values="" >
        If TRUE, change the location of calculation of nudging tendency to 
        the same location where the nudging data is written out.
@@ -303,13 +303,13 @@
        Default: FALSE
        </entry>
 
-<entry id="Num_Slice" type="integer" category="nudging"
+<entry id="Nudge_File_Ntime" type="integer" category="nudging"
        group="nudging_nl" valid_values="" >
        Number of time slices per nudging data file
        The current nudging code only works for the nudging data file with one-day data. 
        It does not work correctly when the nudging data file contains multiple-day data.
-       Thus, Num_Slice has to equal to 1 or Nudge_Times_Per_Day.
-       Set default Num_Slice to 0 to force the user to set its value correctly.
+       Thus, Num_File_Ntime has to equal to 1 or Nudge_Times_Per_Day.
+       Set default Num_File_Ntime to 0 to force the user to set its value correctly.
        Default: 0
        </entry>
 
--- a/components/cam/bld/namelist_files/namelist_definition.xml
+++ b/components/cam/bld/namelist_files/namelist_definition.xml
@@ -269,13 +269,13 @@
        group="nudging_nl" valid_values="" >
        Method to perform Nudging analyses.
        1. Step: the model meteorology is nudged toward the same (future) time 
-          slice of the constraining data within each nudging window (e.g., 6-hr)
+          slice of the constraining data within each nudging window (e.g., 6-hr).
        2. Linear: the model meteorology is nudged toward the state at the next 
           model time step, which is linearly interpolated between two neighboring 
-          time slices of the constraining data 
+          time slices of the constraining data. 
        3. IMT: the model meteorology is nudged toward the constraining data at 
           the same model time step and nudging is only applied when the constrainging 
-          data is available
+          data is available.
        Default: Linear
        </entry>
 
@@ -289,22 +289,28 @@
 
 <entry id="Nudge_Loc" type="logical" category="nudging"
        group="nudging_nl" valid_values="" >
-       If TRUE, change the location of calculation of nudging tendency to the same location
-       where the nudging data is written out
-       If FALSE, calculate the nudging tendency at the beginning of tphysbc (the same as the default model)
+       If TRUE, change the location of calculation of nudging tendency to 
+       the same location where the nudging data is written out.
+       If FALSE, calculate the nudging tendency at the beginning of 
+       tphysbc (the same as the default model).
        Default: TRUE
        </entry>
 
 <entry id="Nudge_Curr" type="logical" category="nudging"
        group="nudging_nl" valid_values="" >
-       Linearly interpolate nudging data to current (set to TRUE) or future (set to FALSE, used by the default model) model time step
+       Linearly interpolate nudging data to current (set to TRUE) or future 
+       (set to FALSE, used by the default model) model time step.
        Default: FALSE
        </entry>
 
 <entry id="Num_Slice" type="integer" category="nudging"
        group="nudging_nl" valid_values="" >
        Number of time slices per nudging data file
-       Default: 1 
+       The current nudging code only works for the nudging data file with one-day data. 
+       It does not work correctly when the nudging data file contains multiple-day data.
+       Thus, Num_Slice has to equal to 1 or Nudge_Times_Per_Day.
+       Set default Num_Slice to 0 to force the user to set its value correctly.
+       Default: 0
        </entry>
 
 <!-- Aerosols: Data (CAM version) -->
--- a/components/cam/bld/namelist_files/namelist_definition.xml
+++ b/components/cam/bld/namelist_files/namelist_definition.xml
@@ -265,217 +265,35 @@
        Default: none
        </entry>
 
-
-<!-- Nudging Parameters -->
-
-<entry id="Nudge_Model" type="logical" category="nudging"
+<entry id="Nudge_Method" type="char*80" category="nudging"
        group="nudging_nl" valid_values="" >
-       Toggle Model Nudging ON/OFF.
+       Method to perform Nudging analyses.
+       Default: Step
+       </entry>
+
+<entry id="Nudge_Tau" type="real" category="nudging"
+       group="nudging_nl" valid_values="" >
+       Use user-defined relaxation time scale. 
+       Default: -999
+       </entry>
+
+<entry id="Nudge_Loc" type="logical" category="nudging"
+       group="nudging_nl" valid_values="" >
+       Change the location of calculation of nudging tendency to the same location 
+       where the nudging data is written out
        Default: FALSE
        </entry>
 
-<entry id="Nudge_Path" type="char*256" input_pathname="abs" category="nudging"
+<entry id="Nudge_Curr" type="logical" category="nudging"
        group="nudging_nl" valid_values="" >
-       Full pathname of analyses data to use for nudging.
-       Default: none
+       Linearly interpolate nudging data to current or future model time step
+       Default: FALSE
        </entry>
 
-<entry id="Nudge_File_Template" type="char*80" category="nudging"
+<entry id="Num_Slice" type="integer" category="nudging"
        group="nudging_nl" valid_values="" >
-       Template for Nudging analyses file names.
-       Default: none
-       </entry>
-
-<entry id="Nudge_Times_Per_Day" type="integer" category="nudging"
-       group="nudging_nl" valid_values="" >
-       Number of analyses files per day.
-       Default: none
-       </entry>
-
-<entry id="Model_Times_Per_Day" type="integer" category="nudging"
-       group="nudging_nl" valid_values="" >
-       Number of time to update model data per day.
-       Default: none
-       </entry>
-
-<entry id="Nudge_Uprof" type="integer" category="nudging"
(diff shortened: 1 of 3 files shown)


components/cam/bld/namelist_files/namelist_definition.xml
```diff
@@ -265,217 +265,52 @@
...
+       Linearly interpolate nudging data to current (set to TRUE) or future 
+       (set to FALSE, used by the default model) model time step.
+       Default: FALSE
        </entry>
 
-<entry id="Nudge_Vwin_Ldelta" type="real" category="nudging"
+<entry id="Nudge_File_Ntime" type="integer" category="nudging"
        group="nudging_nl" valid_values="" >
-       Steepness of LOW end of Vertical Window.
-       Default: none
```
tangq, line 478: @jiansunpnnl , it is not clear to me how to specify the nudging time period and nudging spatial window without these options. Are they handled in a different way? Can you clarify, like give an example? Thanks.

components/cam/src/physics/cam/nudging.F90
```diff
@@ -151,6 +223,12 @@ module nudging
 !      Nudge_Vwin_Hindex   - REAL HI model index of transition
 !      Nudge_Vwin_Ldelta   - REAL LO transition length
 !      Nudge_Vwin_Hdelta   - REAL HI transition length
+!      Nudge_Method        - CHAR method to perform Nudging analyses. [Step,Linear,IMT]
+!      Nudge_Tau           - REAL relaxation time scale if Nudge_Tau > 0
+!      Nudge_Loc_Same      - LOGICAL change the location of calculation of nudging tendency 
+!      Nudge_Curr          - LOGICAL linearly interpolate nudging data to current
+!                                    or future model time step
+!      Nudge_File_Ntime    - INT  Number of time slices per nudging data file
 !    /
```
tangq, line 232: There are so many options in this nudging_nl configuration. It is more efficient to demonstrate the usage by examples. Here seems the right place to include 1 or 2 examples with advanced settings (similar to Table 3 of https://www.geosci-model-dev.net/12/2679/2019/, but with modifications accordingly).

components/cam/bld/namelist_files/namelist_definition.xml
jiansunpnnl: Hi @tangq , thank for the comments. The nudging time period and nudging spatial window are specified by **the same way** as the default model. The new options **DO NOT** affect these settings. A good example could be found in the Table 3 of Tang et al. 2019.

Reference
Tang, Q., Klein, S. A., Xie, S., Lin, W., Golaz, J.-C., Roesler, E. L., Taylor, M. A., Rasch, P. J., Bader, D. C., Berg, L. K., Caldwell, P., Giangrande, S. E., Neale, R. B., Qian, Y., Riihimaki, L. D., Zender, C. S., Zhang, Y., and Zheng, X.: Regionally refined test bed in E3SM atmosphere model version 1 (EAMv1) and applications for high-resolution modeling, Geosci. Model Dev., 12, 2679–2706, https://doi.org/10.5194/gmd-12-2679-2019, 2019.

components/cam/src/physics/cam/nudging.F90
jiansunpnnl: Hi Qi, thanks for this suggestion. Please see the example settings for the revised nudging codes in your first comment.

tangq: The Table in my paper doesn't have the new settings added by this PR. Can you make a new one and add it as comments in the code here?

jiansunpnnl: Thanks, @tangq . I have added two example namelist settings for the typical global nudged simulation and CONUS RRM simulation (based on the Table 3 of your paper) as comments in the nudging.F90. It could be viewed at the commit **f78268c**.

components/cam/src/physics/cam/nudging.F90
```diff
@@ -231,6 +240,97 @@ module nudging
...
+!      Nudge_File_Ntime    = 1
+!
+! More advanced example settings for the CONUS RRM simulation (i.e., Table 3 of Tang et al. (2019), 
+! using UV-only, tau=6h, four time slices per nudging input data file, nudging spatial window) 
+! with this revised nudging code are:
+!
+!    &nudging_nl
+!      Nudge_Model         = .true.
+!      Nudge_Path          = 'PATH_TO_DATA'
+!      Nudge_File_Template = ‘interim_se_%y%m%d00_%y%m%d18_TQUV-%s.nc’
```
tangq, line 286: This file template is for 1 time slice per file because of "-%s". It should be deleted to be consistent with Nudge_File_Ntime    = 4.

components/cam/bld/namelist_files/namelist_definition.xml
wlin7: I think @tangq 's observation is due to a strange behavior of git. Normally, the text shaded in light red would mean to be remove after the PR, which is not true in the current case.

tangq: @wlin7 , thanks for the clarification. Perhaps, we can report it to github.com. I was confused why these "deleted" options can still be used.

kaizhangpnl: Hi @tangq @wlin7 This portion of namelist was duplicated in the original xml file. Jian removed the duplication.

jiansunpnnl: Hi @wlin7 and @tangq . Thanks for the clarification. As we have mentioned in the point#4 of this PR (i.e., the bug fixes section), the nudging-related namelist variables are declared twice in the namelist_definition.xml file, which does not seem necessary to me. I deleted the duplicate declarations and the nudged simulation worked fine. This is the reason why you see some variables are deleted in this file. Hope this is clear.

components/cam/src/physics/cam/nudging.F90
jiansunpnnl: Thanks @tangq. I have just updated the comments in the nudging.F90 about this part.

jiansunpnnl: Thanks, @tangq . I have just deleted the "-%s" and updated the comments in the nudging.F90.

components/cam/bld/namelist_files/namelist_definition.xml
tangq: Thanks for the clarifications, @kaizhangpnl @jiansunpnnl . It is good to know that it's not a github.com issue.

components/cam/src/physics/cam/nudging.F90
tangq: Thank you, @jiansunpnnl, for making the changes.

jiansunpnnl: No problem. Thanks for pointing out this issue.

wlin7:
First thank you @jiansunpnnl for updating this nudging module. It is much more flexible now. A few comments below.

The variable 'Nudge_Loc_Same' does not seem to convey the meaning well by itself. The comment says to calculate the nudging tendency at the same place as where the nudging data are written out. Do you mean just following the call to diag_phys_writeout? We don't actually write out the nudging data. Change to "where model state are written out" would seem more accurate to me. And how about changing the variable name to something like "Nudge_Loc_PhysOut".

Also, can you explain the purpose of calculating the nudging tendency there? Easier to do offline evaluation?

To be more informative, nml variable Nudge_Curr can be changed to Nudge_CurrentStep.

On the code part, I think you intend to completely remove PS nudging. There is inconsistency in that aspect, however. PS-related arrays are still declared and allocated, but reading of PS data has been commented out. Nudge_PSstep is still computed, though Target_PS would always be 0s. If anyone trying to apply PS nudging by turning on Nudge_PSprof and Nudge_PScoef, the results would be very undesirable. I would think to still read in PS, include maintaining two time levels for interpolation with Linear method.

In subroutine nudge_init, with Linear method, INTP_T and INTP_Q are always allocated. However, the interpolations to Target_T and Target_Q are only done when Nudge_Tprof and Nudge_Qprof are non-zeros. Can we also apply the if-condition to the allocation -- and only allocate INTP_T and INTP_Q when they are needed to reduce use of RAM. For completeness, the same can be applied to INTP_U and INTP_V. That would require corresponding changes in  subroutine nudging_update_analyses_se, though.

I like that the 'NUDGING: Current time slice' is printed in subroutine open_netcdf for verification. However, when nudging is applied to multiple variables, the identical message are duplicated. I think it would be useful to add vname and strt3(3) to the output. With that, people can immediately tell which time slice is read without comparing the tinfo value with that in the nudging data file.

That said, all suggested changes are just optional. I did a test run using a recently derived analysis nudging files. It works very well. Thanks again.

jiansunpnnl:
Hi @wlin7 . Thanks for your suggestions. Below are my answers to your questions:
1. Regarding the name issue of "Nudge_Loc_Same" and "Nudge_Curr", I have changed them to your suggested names.
2. Regarding the concern of RAM, it is a great suggestion to use the if-condition to only allocate arrays for the nudged variables. I have followed your suggestion to change the codes correspondingly.
3. Regarding the concern of messages in the subroutine "open_netcdf", I have added the information you suggested.
4. Regarding the issue of PS nudging, I forgot to mention that the default nudging code on the master branch does not support PS nudging, either. As you have noticed, the default nudging code allocates the array for PS but does not read in the data from the netCDF file. The nudging tendency of PS is also not calculated. To me this is equivalent to removing PS nudging completely. However, since our nudging paper does not evaluate the performance of PS nudging, I therefore do not delete all the parts related to PS nudging in the revised code and keep the treatment consistent with the default nudging code. If I add linear interpolation or step function for PS nudging, the revised nudging code will not be BFB identical with the default nudging code even if I use one time slice per input data file, which I think is not desired. I discussed with @kaizhangpnl about this issue and he suggested to issue an error message with "Nudge_PSprof /= 0", explicitly telling the users that the nudging codes do not support PS nudging yet.
5. Regarding the concern of "Nudge_Loc_PhysOut", I agree with you that "where model state are written out" is more accurate and I have changed the comments in the codes. The reason why we want to change the location of calculation of nudging tendency is that we realized the location would affect the model results when we perform nudged simulation towards model's own meteorology. But this effect is smaller when we perform nudged simulation towards reanalysis data. Therefore, we suggest to set "Nudge_Loc_PhysOut = True".

I am testing the latest changes in the nudging code and would let you know whether my implementation is correct. Thank you again for your detailed and great suggestions!

wlin7:
Thanks for updating and detailed explanation, @jiansunpnnl . Specifically for PS nudging, I like the idea of issuing warning and aborting the run if Nudge_Psprof != 0. 

Added a comment on the duplicated nml definitions for nudging parameters. We didn't notice that.  Otherwise we would be less confident to use this interface for reviewing code changes.

For record, yesterday I also confirmed another BFB: nudging file with multiple time slices vs one file per time, with Nudge_File_Ntime set accordingly. Confident to use from now on. Thanks again.

components/cam/bld/namelist_files/namelist_definition.xml
```diff
@@ -265,217 +265,52 @@
...
+       Default: -999
        </entry>
 
-<entry id="Nudge_PScoef" type="real" category="nudging"
+<entry id="Nudge_Loc_PhysOut" type="logical" category="nudging"
        group="nudging_nl" valid_values="" >
-       Coeffcient for PS nudging.
-       Default: none
+       If TRUE, change the location of calculation of nudging tendency to 
+       the same location where the nudging data is written out.
```
wlin7, line 293: Please also update this description.

jiansunpnnl:
Hi @wlin7 , thanks for your reply and confirmation. It is glad to know that the revised codes work well on your machine, too.

The latest commit "5910536" should contain all the changes based on your and Qi's suggestions. Let me know if they look good to you. Thank you for your time and help.

kaizhangpnl:
@singhbalwinder Would you please help on the integration? Thanks!

singhbalwinder:
Thanks Jian for fixing the commit message. I will start integrating this PR.

jiansunpnnl:
Thanks for your help, @singhbalwinder . Let me know the commit message needs further revision or you have any question about the code implementation.

singhbalwinder:
Merged to next

jiansunpnnl:
Hi, @singhbalwinder , thank for pointing out the bug of nudging code in a free-running simulation. It is fixed in the commit "**eb2017e**".

PR #3570, opened 2020-05-13
Fix ftype1 in NH theta

oksanaguba:
Fixing ftype1 for NH runs, previously, NH variables were not updated in dp_coupling layer.

### DIFF
--- a/components/cam/src/dynamics/se/stepon.F90
+++ b/components/cam/src/dynamics/se/stepon.F90
@@ -1,5 +1,4 @@
 !BOP
-!
 ! !MODULE: stepon -- FV Dynamics specific time-stepping
 !
 ! !INTERFACE:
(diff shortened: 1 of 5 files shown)


oksanaguba:
eam_theta passed against today's master except for ftype1_thetanh test, as expected. will run whole suite tomorrow.

mt5555:
Looks good and more modular.  Can you remove applyCAMforcing_dynamics()?  This entire subroutine is now just an elem loop, so it looks like it no longer makes sense to exist with this newer design.

mt5555:
@bartgol this new interface will have to be back ported to hommexx?   we could instead keep the old interface an only change code in EAM.

bartgol:
So, other than the added modularity, the main difference is that phi and w_i were not forced with ftype 1? I think we ever handled ftype 1 in hommexx. I only see support for ftype 0 and 2. I guess we can add ftype 1 as well.

oksanaguba:
told to do it in 1-file PR.

PR #3574, opened 2020-05-13
Update modules on Cori (cori-knl and cori-haswell) to reflect the currents defaults

ndkeen:
Update modules on Cori (cori-knl and cori-haswell) to reflect the currents defaults.

Except for the Intel compiler -- leaving this as-is at v18.
We will update the compiler version in a different PR.

This PR updates
 cray-mpich/7.7.6 -> cray-mpich/7.7.10
 craype/2.5.18 -> craype/2.6.2
 cray-netcdf-hdf5parallel/4.6.1.3 -> cray-netcdf-hdf5parallel/4.6.3.2
 cray-hdf5-parallel/1.10.2.0 -> cray-hdf5-parallel/1.10.5.2
 cray-parallel-netcdf/1.8.1.4 -> cray-parallel-netcdf/1.11.1.1

These changes should be BFB and have been with testing so far (e3sm_developer plus maint-1.0 tests)

The performance and memory use are still under investigation. 

[bfb]

### DIFF
--- a/cime/config/e3sm/machines/config_machines.xml
+++ b/cime/config/e3sm/machines/config_machines.xml
@@ -119,7 +119,6 @@
         <command name="rm">craype-sandybridge</command>
         <command name="rm">craype-ivybridge</command>
         <command name="rm">craype</command>
-        <command name="rm">craype-hugepages2M</command>
         <command name="rm">papi</command>
         <command name="rm">cmake</command>
         <command name="rm">cray-petsc</command>
@@ -199,6 +198,7 @@
       <env name="HDF5_USE_FILE_LOCKING">FALSE</env>
     </environment_variables>
     <environment_variables compiler="intel">
+      <env name="CRAYPE_LINK_TYPE">static</env>
       <env name="FORT_BUFFERED">yes</env>
     </environment_variables>
   </machine>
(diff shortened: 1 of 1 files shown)


jayeshkrishna:
Will the Cori nightlies fail with the change (does change to TPUT - throughput and MEMCOMP - memory compare affect the nightlies?) ?

How did you narrow down that the I/O libraries caused this issue (Were you able to narrow it down to a single library update)?

ndkeen:
I think there will be multiple MEMCOMP and TPUT fails, but I submit they are OK.  

I ran several different e3sm_developer tests. The only real diffs in this PR are craype, cray-mpich, and IO-related modules (changed as a group)
1) with all new modules 
2) with *only* updating the craype module (all others same as master)
3) with only updating craype and cray-mpich modules (so only IO modules are new)
4) with updating these modules to versions even higher than the default

And I found that with (3), the tests would not show a difference in memory use.

rljacob:
notes from telecon:  Noel is going to do another set of tests.

ndkeen:
By adding `CRAYPE_LINK_TYPE=static` to env variables, this seems to remedy the performance degradation I was using. The craype module update is what is triggering the use of dynamic linking (which by itself should not matter, but this must be changing what exactly is linked). However, there were still some tests showing slight decrease (or at least variability) in performance with everything in place. Therefore, I'd like to close this PR and create a new one that only updates mpich & netcdf modules (which seem to have no impact). And then make a another PR to update craype so that we can look at it more closely.

PR #3575, opened 2020-05-14
Fix ftype1 in NH theta

oksanaguba:
Fixing ftype1 for NH runs, previously, NH variables were not updated in dp_coupling layer.

[nonBFB] for cam-thetanh_ftype1 test.

### DIFF
--- a/components/cam/src/dynamics/se/stepon.F90
+++ b/components/cam/src/dynamics/se/stepon.F90
@@ -226,6 +226,7 @@ subroutine stepon_run2(phys_state, phys_tend, dyn_in, dyn_out )
    use hycoef,          only: hyai, hybi
    use cam_history,     only: outfld, hist_fld_active
    use prim_driver_base,only: applyCAMforcing_tracers
+   use prim_advance_mod,only: applyCAMforcing_dynamics
    use element_ops,     only: get_temperature
 
    type(physics_state), intent(inout) :: phys_state(begchunk:endchunk)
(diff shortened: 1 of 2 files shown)


oksanaguba:
eam_theta passed against yesterday's master (the same one that was used for the branch) except for ftype1_thetanh (diff) test, as expected. will run whole suite tomorrow.

mt5555:
should we add an "#ifndef CAM" around the line of code which applies the w tendency?  Currently I think we just assume that is zero and thus harmless to apply to w?

Or, since we dont have any test cases that actually generate a w tendency,  maybe we should remove it altogether?

oksanaguba:
I ignored w, phi issue for HY runs just because in CAAR we also have w, phi local computations. Or were those removed?

oksanaguba:
i misunderstood your question but now looking at the code i see that FM is used only with two components in some of the code in stepon while it is assumed that FM has 3 components. i will look more.

mt5555:
just to expand more: I was thinking this line of code in applyCAMforcing_dynamics:

 elem(ie)%state%w_i(:,:,1:nlev,np1) = elem(ie)%state%w_i(:,:,1:nlev,np1) + dt*elem(ie)%derived%FM(:,:,3,:)

Not sure the right solution, but we can either assume FM(:,:,3,:)=0, or disable it in an ifndef CAM.

oksanaguba:
e3sm_atm_integration all pass except for SMS.ne4_ne4.FC5AV1C-L.anvil_intel.cam-thetanh_ftype1 as expected.

oksanaguba:
in next

PR #3577, opened 2020-05-15
Updates the SMS.r05_r05.I1850CLM45CN test

bishtgautam:
Updates the `SMS.r05_r05.I1850CLM45CN` test to
`SMS.r05_r05.I1850CLM45CN.clm-qian_1948` to avoid downloading
multiple DATAM files by setting the end year to be the same as
the start year.

[BFB]


rljacob:
@bishtgautam just noticed that this test was changed in e3sm_land_developer only.  I'd like it to also be changed in e3sm_integration.

bishtgautam:
This test gets automatically updated in e3sm_integration (because e3sm_integration includes e3sm_developer, which in turn includes e3sm_land_developer)

bishtgautam:
@rljacob Btw, should I redo this PR because of the recent change of CIME to a submodule?

rljacob:
Sure that would help but not because of the submodule.  Because tests.py moved out of cime to e3sm/cime_config.

rljacob:
I see what happened.  There was a PR that moved this test out of developer because it was taking to much time/memory on melvin.  So on latest master, its placed in e3sm_integration.   So yes just redo this PR.

PR #3578, opened 2020-05-15
Modifies chem pp optimization level from O2 to O1 for PGI compilers

singhbalwinder:
Chem pp was hanging on Compy when compiled with the PGI compiler.
An easy fix is to decrease the optimization level from 2 to 1.

It works fine on all other testing machines and it also works fine
on Compy with the Intel compiler with O2 optimizations.

[BFB] - Bit-For-Bit

### DIFF
--- a/components/cam/chem_proc/src/Makefile
+++ b/components/cam/chem_proc/src/Makefile
@@ -133,13 +133,13 @@ ifeq ($(FC),lf95)
   endif
 endif
 ifeq ($(FC),pgf90)
-  FFLAGS      := -O2 -c -g -Mfree -module $(OBJ_DIR)
+  FFLAGS      := -O1 -c -g -Mfree -module $(OBJ_DIR)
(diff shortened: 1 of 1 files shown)


tangq:
@singhbalwinder , do the results change with the decreased optimization level from 2 to 1 for the PGI compiler on compy? If so, how large are the changes? It would be better if it's BFB.

rljacob:
No one is using pgi for production on compy so non-BFB changes are fine.

singhbalwinder:
This code generates fortran files. I ran two tests one with Intel O2 (default) and other with PGI O1 and compared the fortran files. They are exactly the same. I think that is a pretty good indication that the model results will stay BFB.

rljacob:
Merged to next.

wlin7:
Merged to master. Thanks @rljacob for taking care of the next.

PR #3579, opened 2020-05-18
Fix MMF explicit scalar momentum transport (ESMT)

whannah1:
A few fixes to enable the scalar momentum transport scheme (ESMT) in the MMF. The main change is to just update the dimension ordering which were switched during the refactoring to support running the MMF on GPUs. 

[BFB] since we currently don't test this feature.


rljacob:
@brhillman is this ready to merge?

whannah1:
@rljacob this PR is ready, the main reason we held off is because of the ERP test issues we discovered awhile back, and perhaps another issue that I'm forgetting. But now that we switched to the ERS test for MMF we can merge this whenever.

rljacob:
Lets merge #3639 first since that will fix the ERS test for the MMF compset.

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #3580, opened 2020-05-18
Add MMF-RCE compset

whannah1:
This includes the new MMF-RCE compset as well as a couple minor fixes in the version of radiation.F90 that is used for MMF, which did not include the ability to reduce the solar constant and disable aerosol radiative effects, which were implemented in the non-MMF version of RRTMGP.

### DIFF
--- a/components/cam/cime_config/config_component.xml
+++ b/components/cam/cime_config/config_component.xml
@@ -131,6 +131,8 @@
       <value compset="_CAM5%MMF[12]-TEST_"     >-crm_nx  8 -crm_nx_rad  2 -crm_ny 1 -crm_ny_rad 1 </value>
       <value compset="_CAM5%MMF[12]-MAML-TEST" >-crm_nx  8 -crm_nx_rad  8 -crm_ny 1 -crm_ny_rad 1 </value>
       <value compset="_CAM5%MMF2-ECPP-TEST"    >-crm_nx  8 -crm_nx_rad  1 -crm_ny 1 -crm_ny_rad 1 </value>
+      <value compset="_CAM5%MMF[12]-RCE_"      >-crm_nx 64 -crm_nx_rad 16 -crm_ny 1 -crm_ny_rad 1 </value>
+      <value compset="_CAM5%MMF[12]-RCE_"      >-aquaplanet -rce </value>
       <!--  -->
       <value compset="GEOS_CAM"       >-offline_dyn</value>
       <value compset="GEOS_CAM4"      >-nlev 56</value>
@@ -294,6 +296,7 @@
       <value compset="2000_CAM5%SP2"     >2000_cam5_av1c-04p2-MMF-2mom</value>
       <value compset="2000_CAM5%MMF1"    >2000_cam5_av1c-04p2-MMF-1mom</value>
       <value compset="2000_CAM5%MMF2"    >2000_cam5_av1c-04p2-MMF-2mom</value>
+      <value compset="_CAM5%MMF1-RCE" >RCEMIP_EAMv1</value>
     </values>
     <group>run_component_cam</group>
     <file>env_run.xml</file>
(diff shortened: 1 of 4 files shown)


components/cam/cime_config/config_component.xml
```diff
@@ -131,6 +131,8 @@
       <value compset="_CAM5%MMF[12]-TEST_"     >-crm_nx  8 -crm_nx_rad  2 -crm_ny 1 -crm_ny_rad 1 </value>
       <value compset="_CAM5%MMF[12]-MAML-TEST" >-crm_nx  8 -crm_nx_rad  8 -crm_ny 1 -crm_ny_rad 1 </value>
       <value compset="_CAM5%MMF2-ECPP-TEST"    >-crm_nx  8 -crm_nx_rad  1 -crm_ny 1 -crm_ny_rad 1 </value>
+      <value compset="_CAM5%MMF[12]-RCE_"      >-crm_nx 64 -crm_nx_rad 16 -crm_ny 1 -crm_ny_rad 1 </value>
+      <value compset="_CAM5%MMF[12]-RCE_"      >-aquaplanet -rce </value>
```
brhillman, line 135: Shouldn't these be combined into a single line?

components/cam/cime_config/config_component.xml
```diff
@@ -294,6 +296,7 @@
       <value compset="2000_CAM5%SP2"     >2000_cam5_av1c-04p2-MMF-2mom</value>
       <value compset="2000_CAM5%MMF1"    >2000_cam5_av1c-04p2-MMF-1mom</value>
       <value compset="2000_CAM5%MMF2"    >2000_cam5_av1c-04p2-MMF-2mom</value>
+      <value compset="_CAM5%MMF1-RCE" >RCEMIP_EAMv1</value>
```
brhillman, line 299: Is there a reason you don't use MMF[12] here? Assuming it's only valid for 1-moment, but is there harm in leaving the config open to 2-moment?

components/cam/cime_config/config_component.xml
```diff
@@ -427,6 +430,7 @@
     <value compset="DOCN%AQP">$SRCROOT/components/cam/cime_config/usermods_dirs/aquap</value>
     <value compset="DOCN%SOMAQP">$SRCROOT/components/cam/cime_config/usermods_dirs/aquap</value>
     <value compset="CAM5%RCE">$SRCROOT/components/cam/cime_config/usermods_dirs/rcemip</value>
+    <value compset="CAM5%MMF1-RCE">$SRCROOT/components/cam/cime_config/usermods_dirs/rcemip</value>
```
brhillman, line 433: See comment above about globbing for both MMF1 and MMF2.

components/cam/cime_config/config_component.xml
whannah1: They could be. I figured I should separate the part that deals with the CRM domain size since it's separated for the other options above. I'd prefer to keep it this way unless there's a non-cosmetic reason to merge the lines.

components/cam/cime_config/config_component.xml
whannah1: this should be [12], I'll correct it

components/cam/cime_config/config_component.xml
whannah1: yea, I'll fix this too

whannah1:
ok, I finally addressed @brhillman 's comments, so the PR should be ready to go.

brhillman:
@rljacob should we add a regression test for this to e3sm_developer or e3sm_integration before merging this? This is not covered by our current MMF test, but will probably only be used by a few folks.

rljacob:
If the feature needs to be kept working, the answer is "yes".

rljacob:
Use e3sm_integration.

brhillman:
Great thanks!

brhillman:
Pushed to next.

brhillman:
Merged to master.

PR #3581, opened 2020-05-18
Update MMF radiation_tend for cam_optics changes

brhillman:
Fixes the MMF-specific `radiation_tend` to use the new optics interfaces, which commit a4eea9c222f26dd2d98664259c07aaf141b3c862 modified to provide additional outputs to use with COSP. Commit 417e7cdd495d90df975d7f4b4df0e753c5732fcc then broke the MMF build because it relied on the old interface, so this fixes the broken build. Also switches the MMF test in the integration suite from an ERP test to an ERS test so that we have a test that works in master for MMF until we can track down the ERP fail.

[BFB]


rljacob:
Can you update the description so refer to hashes instead of PR numbers?   This is because that message gets copied in to the merge commit message and it should be free of any URL's or github refs.

brhillman:
@rljacob apologies, is the edited description what you are looking for?

rljacob:
Looks great.  Thanks!

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #3583, opened 2020-05-19
Homme: Fix a restart bug.

### LINKED ISSUE #3582: Homme: When theta_hydrostatic_mode is true, need to 0-init w_i and phinh_i on restart.
An ERS test with theta-l in hydrostatic mode non-deterministically fails due to a detected NaN in repro_sum. This was bisected to homme/src/theta-l/prim_state_mod.F90:291, the integral over w_i. The integral over phinh_i two lines later in principle should have this failure mode, too, but so far in repeated tests it happens not to have triggered.

This bug does not change answers because w_i and phinh_i is not used in the dynamics, but it will stop the run if prim_printstate's global_integral call leads to a detected NaN. Thus, the fix will be BFB.

ambrad:
When restarting with theta_hydrostatic_mode = true, init w_i to phinh_i to
0. This prevents a possible NaN dection in prim_printstate. This PR does not
change answers because w_i and phinh_i are not actually used in hydrostatic
mode.

I detected a second source of potential NaNs in prim_printstate: When Qdp is
written at restart, the unused time slot is uninitialized. This can then lead to
NaNs in repro_sum. Fix this by 0'ing out all of Qdp before writing restart data
to it.

Fixes #3582

[BFB]

### DIFF
--- a/components/cam/src/dynamics/se/restart_dynamics.F90
+++ b/components/cam/src/dynamics/se/restart_dynamics.F90
@@ -642,6 +642,9 @@ CONTAINS
     endif
 #endif
 
+    do ie = 1,nelemd
+       elem(ie)%state%Qdp = 0
+    end do
     do q=1,qsize_d
        call pio_setframe(File,qdesc(q), t)
        call pio_read_darray(File, qdesc(q), iodesc3d, var3d, ierr)
--- a/components/cam/src/dynamics/se/restart_dynamics.F90
+++ b/components/cam/src/dynamics/se/restart_dynamics.F90
@@ -605,7 +605,12 @@ CONTAINS
     end do
 
 #ifdef MODEL_THETA_L
-    if ( .not. theta_hydrostatic_mode )then
+    if ( theta_hydrostatic_mode ) then
+       do ie=1,nelemd
+          elem(ie)%state%w_i = 0
+          elem(ie)%state%phinh_i = 0
+       end do       
+    else
        call pio_setframe(File,Wdesc, t)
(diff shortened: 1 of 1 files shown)


ambrad:
eam_theta_pg2 passes. I'm now running e3sm_developer against baselines and will report back when this is done.

ambrad:
Repeated runs with melvin_gnu now pass always, and before then I was able to demonstrate each fix, when removed, could trigger a NaN failure.

eam_theta_pg2 passes on compy_intel.

I'll update this comment with sandiatoss3 state when that's done.

components/cam/src/dynamics/se/restart_dynamics.F90
```diff
@@ -637,6 +642,9 @@ subroutine read_restart_dynamics (File, dyn_in, dyn_out, NLFileName)
     endif
 #endif
 
+    do ie = 1,nelemd
+       elem(ie)%state%Qdp = 0
```
oksanaguba, line 646: If without this there are nans in printstate, does it mean Qdp should have been init-ed with Q*dp (Q and dp i assume are read from restart)?

ambrad: It's a separate issue as to whether Qdp should be set to Q*dp.

To be clear, restart only has values for one of Qdp's two time slots.

printstate is seeing NaNs because it is sometimes looking at the wrong time level. Or maybe it's looking at both Qdp time levels; printstate is a bit of a mess.

oksanaguba:
in next

PR #3589, opened 2020-05-21
New ocean mesh structure with GPU replication

mark-petersen:
See https://github.com/MPAS-Dev/MPAS-Model/pull/496 by @philipwjones 

This change adds a mesh structure with all current mesh pool data and replicates that data on a GPU (or other accelerator) for the duration of the simulation to reduce the need for data transfers of static mesh data. It is a critical piece of infrastructure needed for future GPU modifications being staged. In a later change, this structure was replaced by public module variables instead since a large user-defined type did not perform well on GPUs. Mesh variables can now be accessed just by "use"-ing this module.

This module also has a revised index naming in which nCells, nCellsSolve and nCellsArray are replaced by nCellsAll, nCellsOwned, and nCellsHalo(n) where the latter differs from nCellsArray by only including the index of each halo depth and no longer includes nCellsOwned as the first entry. Similar changes were introduced for Edges, Vertices.

NOTE: this change requires only one block per MPI task and will abort if more than one block per MPI task is attempted. Any multi-block tests will fail.

This has been tested on summit in a QU240 test case and is bit-for-bit since the structure itself is not yet being used anywhere. However, I have verified the data is correctly transferred to the GPU accelerator on Summit.

[b4b]


mark-petersen:
Passes
```
PET_Ln9.T62_oQU240.GMPAS-IAF.cori-knl_gnu
PEM_Ln9.T62_oQU240.GMPAS-IAF.cori-knl_intel
```

mark-petersen:
Rebased and ready to merge. The four COMPASS files are related to this PR because testing of multiple blocks was removed.

jonbob:
merged to next

jonbob:
merged to master

PR #3593, opened 2020-05-24
move homme standalone utilities into homme_tools

mt5555:
4 tools moved into the homme_tools framework recently created by @ambrad .  Related tool code removed from model source code.  Namelists for the individual tools are now much simpler and dont require running the full model.  Documentation for all tools in homme/test/tool/README

new tools added:
1. SCRIP and latlon metadata generation
2. GLL topo smoothing
3. HOMME's built in netcdf interpolation to lat/lon grids
4. generate mapping files based on HOMME's built in bilinear option
5. SEMapping ( latlon meta data output ) code moved from EAM to this tool

HOMME's "template" test was replaced with a homme_tool test, that tests options 1-4 above with an ne4 grid (and in the case of metadata, a lowres RRM grid).

[BFB]


oksanaguba:
Ran these with debug flags:

src/tool/native_mapping.F90
line 198 count_total used before defined?

src/share/coordinate_systems_mod.F90 line 627 -- ran with output and cart3D%x, cart3D%y can be both zero
       x =  cart3D%y/cart3D%x

src/test_src/dcmip12_wrapper.F90 
routine get_coordinates() is called for midlevels and interfaces, but it operates only with midlevels in there. k can exceed bounds.

subroutine get_coordinates(lat,lon,hyam,hybm, i,j,k,elem,hvcoord)
...
  ! get hybrid coeffiecients at midpoint of vertical level k
  hyam = hvcoord%hyam(k)
  hybm = hvcoord%hybm(k)

oksanaguba:
For my own benefit I went thru these tools:

tool_mappingfiles.nl -> calls gll_mapping_file, dumps file coord*nc and mapping file from ne4 to 9x16 using bilin

tool-template1,2 -> calls grid_tempalte_tool, for ne4 and ne0, scrip data generation, area, cv_lat,...

tool-topooutput -> calls topo_gll_to_smoothed, empty infilename, so, dumps PHIS

tool-toposmooth-gll -> calls topo_gll_to_smoothed, uses PHIS from above to smooth it

topo-zinterp -> calls interpolate_tool and interpolates phis_smoothed to some default latlon grid.

oksanaguba:
homme suit did not pass with multiple tests failing. For example, swtc1 fails to complete, crashes somewhere in the output. I rebuilt with debug flags, had to set PLEV to 6 in CMakeLists.txt for this test (default was 4, but the test uses 6), and got crash for running with 2 ranks here:
```
 ii =          295
 ii =          295
 ii =          295
 ii =           -1
forrtl: severe (408): fort: (3): Subscript #1 of the array INTERPDATA has value -1 which is less than the lower bound of 1

swtcA              0000000000BBB5F6  Unknown               Unknown  Unknown
swtcA              00000000008C2FE4  interpolate_mod_m        1357  interpolate_mod.F90
swtcA              000000000057CDFE  interp_movie_mod_         264  interp_movie_mod.F90
```
which corresponds to this code 
``` 
          if (global_elem_gid(j,i) == -1 ) then
             missing_pts = missing_pts + 1
             if (par%masterproc) &
                  print *,'Error: point not claimed by any element j,i,lat(j),lon(i)=',j,i,lat(j),lon(i)
          else if (local_elem_gid(j,i) == global_elem_gid(j,i)  ) then
             ii = local_elem_num(j,i)
print *, 'ii = ', ii
             interpdata(ii)%n_interp = interpdata(ii)%n_interp + 1
```
With 1 rank tests gets thru.

oksanaguba:
Seems that at least a few more tests failed due to the same issue.

mt5555:
Fixed the interpolation error.  Removed compiler warnings when qsize<6.  
Also removed the "SEMapping.nc" tool from EAM and moved this into homme_tool

components/homme/CMakeLists.txt
```diff
@@ -354,6 +354,12 @@ if(NOT BUILD_HOMME_WITHOUT_PIOLIBRARY)
       file(WRITE "${cprnc_dummy_file}" "#dummy Macros file for non-CIME machines")
     endif ()
 
+    # cprnc's CMake system needs various nonstandard variables:
```
oksanaguba, line 357: this is great! -- I was fighting it just now.

mt5555:
This PR is finally ready.  eam_theta and eam_preqx. test suites PASS. 
HOMME test: all pass except for the new test due to missing baselines.

rljacob:
Is it ready now?

mt5555:
yes.

components/homme/src/interpolate_driver_mod.F90
```diff
@@ -540,9 +540,9 @@ subroutine interpolate_vars(infile, outfile, elem, par, interpdata)
     real(kind=real_kind), pointer :: zeta(:,:,:,:),div(:,:,:,:)
 
     integer(kind=nfsizekind) :: start(3), count(3)
-    integer :: starti(3), counti(3)
+    integer :: starti(3), counti(3), index(1)
```
oksanaguba, line 543: How did this bug come up?

mt5555: here's some more info:  https://github.com/E3SM-Project/scorpio/issues/327

it's probably been in there for a long time, since most people stopped using HOMME's built in interpolation tool several years ago

mt5555:
running the new hommetool test on Anvil and Darwin (macos) shows a couple of issues:

with PIO1:   all other tests are fine and PASS. the hommetool test will run on Anvil, but subtest #6 produces bad output (zero data in file).   On Darwin, subtest #6 often hangs, but sometimes runs.  when it runs, output is bad.

with SCORIP:  with dqwu/fix_box_compute_counts, all tests, including hommetool test look good.   

But we cant switch to SCORPIO until dqwu's branch is merged into E3SM master.  Thus for now, I'm going to update this branch to disable subtest #6.  

We should re-enable it in a future PR that also switches HOMME to SCORPIO

oksanaguba:
i would vote to not to disable the test and to keep it, just because it can be easily forgotten.

mt5555:
my only hesitation with keeping the test enabled is that it hangs on some platforms, and then the rest of the tests never get run.  On the perf call today, it looks like the SCORPIO bug fixes will go in E3SM master next week, so we could delay this PR until then.

mt5555:
per recommendation of integrator,  @oksanaguba ,  this PR will be merged as soon as SCORPIO is updated in E3SM master, hopefully sometime next week.

rljacob:
@oksanaguba SORPIO update is done so please merge this.

oksanaguba:
in next

PR #3595, opened 2020-05-25
Add anvil bfb file

oksanaguba:
Adding a homme bfb cmake file for anvil to enable hommexx testing.


oksanaguba:
This is a very simple PR but I had to modify ADD_TEST line for unit tests to call them like mpiexec -np 12 ./remap_ut 
instead of
 mpiexec -np 12 remap_ut
Still, unit tests on an allocated node randomly fail with
2/6 Test #2: random_init_ut_test ..............SIGPIPE***Exception:   0.14 sec

oksanaguba:
To add -- when run manually, tests are always ok.

rljacob:
@oksanaguba please add a description for this PR.

components/homme/test/unit_tests/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ macro(cxx_unit_test target_name target_f90_srcs target_cxx_srcs include_dirs con
   IF (${NUM_CPUS} EQUAL 1)
     ADD_TEST(${target_name}_test ${target_name})
   ELSE()
-    ADD_TEST(${target_name}_test mpiexec -np ${NUM_CPUS} ${target_name})
```
bartgol, line 12: I'm surprised this ever worked without your fix.

components/homme/cmake/machineFiles/anvil-bfb.cmake
```diff
@@ -0,0 +1,66 @@
...
+  SET (HOMME_USE_MKL "TRUE" CACHE FILEPATH "") # for Intel
+  #turn on preqxx target and thus strict fpmodel for F vs CXX comparison
+  SET (ADD_Fortran_FLAGS "-traceback -fp-model strict -qopenmp -O1" CACHE STRING "")
+  SET (ADD_C_FLAGS "-traceback -fp-model strict -qopenmp -O3" CACHE STRING "")
+  SET (ADD_CXX_FLAGS "-traceback -fp-model strict -qopenmp -O3" CACHE STRING "")
+  SET (BUILD_HOMME_PREQX_KOKKOS TRUE CACHE BOOL "")
+  SET (ENABLE_PREQX_KOKKOS_BFB_TESTS TRUE CACHE BOOL "")
+  SET (HOMME_TESTING_PROFILE "short" CACHE STRING "")
+
+#or just these? was used before
```
bartgol, line 48: ENABLE_OPENMP is on by default, so SetCompilerFlags _should_ be finding openmp (which should add the `-qopenmp` flag). So having the flag here might be not necessary (not sure if one may even "not want it").

That said, I don't see any specification of Kokkos, so unless one passes Kokkos directives to cmake or a path to a pre-installed kokkos, the `-qopenmp` won't matter for the xx target (Kokkos will be built in serial only).

As for the traceback, it only helps in case of crash, to figure out the offending line, if I understand correctly. The cost is a small executable size increase, I think. I guess we can probably keep it.

oksanaguba:
For cxx targets I don't think we focus on OMP tests yet, so, kokkos settings are not in these files. 

This PR is not quite ready, I still see random fails in unit tests. Texting a fix now.

oksanaguba:
This version was run on anvil and sandiatoss3 and I saw no issues. Before for anvil, all tests were using srun but unit tests were using mpiexec, also, some unit tests were not using launcher and executables were not found. I made all of unit tests run with srun on anvil and added launcher definition to sandiatoss3-bfb file.

This is ready.

PR #3606, opened 2020-05-27
Add low-res tri-grid configurations for testing

jonbob:
This PR adds support for ne4pg2-r05-oQU480 and ne16pg2-r05-oQU240 configurations, including all mapping and domain files. All new files have been placed on the inputdata server and should now be available. Though these resolutions are designed for testing, no tests have been added yet -- though that should happen before merging.

### DIFF
--- a/components/cam/bld/namelist_files/namelist_defaults_cam.xml
+++ b/components/cam/bld/namelist_files/namelist_defaults_cam.xml
@@ -229,9 +229,10 @@
 <bnd_topo hgrid="10x15"     >atm/cam/topo/USGS-gtopo30_10x15_remap_c050520.nc</bnd_topo>
 
 <bnd_topo hgrid="ne4np4"   npg="0">atm/cam/topo/USGS-gtopo30_ne4np4_16x.c20160612.nc</bnd_topo>
-<bnd_topo hgrid="ne4np4"   npg="2">atm/cam/topo/USGS-gtopo30_ne4pg2_16xdel2-PFC-consistentSGH.c20190618.nc</bnd_topo>
+<bnd_topo hgrid="ne4np4"   npg="2">atm/cam/topo/USGS-gtopo30_ne4np4pg2_16x.c20160612_converted.nc</bnd_topo>
 <bnd_topo hgrid="ne11np4"  npg="0">atm/cam/topo/USGS-gtopo30_ne11np4_16xconsistentSGH.c20160612.nc</bnd_topo>
 <bnd_topo hgrid="ne16np4"  npg="0">atm/cam/topo/USGS-gtopo30_ne16np4_16xconsistentSGH.c20160612.nc</bnd_topo>
+<bnd_topo hgrid="ne16np4"  npg="2">atm/cam/topo/USGS-gtopo30_ne16np4pg2_16xdel2_20200527.nc</bnd_topo>
 <bnd_topo hgrid="ne30np4"  npg="0">atm/cam/topo/USGS-gtopo30_ne30np4_16xdel2-PFC-consistentSGH.nc</bnd_topo>
 <bnd_topo hgrid="ne30np4"  npg="2">atm/cam/topo/USGS-gtopo30_ne30np4pg2_16xdel2.c20200108.nc</bnd_topo>
 <bnd_topo hgrid="ne60np4"  npg="0">atm/cam/topo/USGS-gtopo30_ne60np4_16xconsistentSGH.c20140517.nc</bnd_topo>
@@ -823,8 +824,10 @@
 
 <!-- Dry Dep surface data file needed by prognostic MAM on unstructured grid only. -->
 <drydep_srf_file hgrid="ne4np4"  >atm/cam/chem/trop_mam/atmsrf_ne4np4_c151204.nc</drydep_srf_file>
+<drydep_srf_file hgrid="ne4np4" npg="2">atm/cam/chem/trop_mam/atmsrf_ne4pg2_200527.nc</drydep_srf_file>
 <drydep_srf_file hgrid="ne11np4" >atm/cam/chem/trop_mam/atmsrf_ne11np4_c151204.nc</drydep_srf_file>
 <drydep_srf_file hgrid="ne16np4" >atm/cam/chem/trop_mam/atmsrf_ne16np4_110920.nc</drydep_srf_file>
+<drydep_srf_file hgrid="ne16np4" npg="2">atm/cam/chem/trop_mam/atmsrf_ne16pg2_200527.nc</drydep_srf_file>
 <drydep_srf_file hgrid="ne30np4" npg="0">atm/cam/chem/trop_mam/atmsrf_ne30np4_110920.nc</drydep_srf_file>
 <drydep_srf_file hgrid="ne30np4" npg="2">atm/cam/chem/trop_mam/atmsrf_ne30pg2_200129.nc</drydep_srf_file>
 <drydep_srf_file hgrid="ne60np4" >atm/cam/chem/trop_mam/atmsrf_ne60np4_110920.nc</drydep_srf_file>
--- a/components/cam/bld/namelist_files/namelist_defaults_cam.xml
+++ b/components/cam/bld/namelist_files/namelist_defaults_cam.xml
@@ -229,7 +229,7 @@
 <bnd_topo hgrid="10x15"     >atm/cam/topo/USGS-gtopo30_10x15_remap_c050520.nc</bnd_topo>
 
 <bnd_topo hgrid="ne4np4"   npg="0">atm/cam/topo/USGS-gtopo30_ne4np4_16x.c20160612.nc</bnd_topo>
-<bnd_topo hgrid="ne4np4"   npg="2">atm/cam/topo/USGS-gtopo30_ne4np4pg2_16x.c20160612_converted.nc</bnd_topo>
+<bnd_topo hgrid="ne4np4"   npg="2">atm/cam/topo/USGS-gtopo30_ne4np4pg2_16x_converted.c20200527.nc</bnd_topo>
 <bnd_topo hgrid="ne11np4"  npg="0">atm/cam/topo/USGS-gtopo30_ne11np4_16xconsistentSGH.c20160612.nc</bnd_topo>
 <bnd_topo hgrid="ne16np4"  npg="0">atm/cam/topo/USGS-gtopo30_ne16np4_16xconsistentSGH.c20160612.nc</bnd_topo>
 <bnd_topo hgrid="ne16np4"  npg="2">atm/cam/topo/USGS-gtopo30_ne16np4pg2_16xdel2_20200527.nc</bnd_topo>
--- a/cime/config/e3sm/config_grids.xml
+++ b/cime/config/e3sm/config_grids.xml
@@ -3377,6 +3377,11 @@
       <map name="ROF2LND_FMAPNAME">lnd/clm2/mappingdata/maps/ne4np4/map_r05_TO_ne4np4_aave.160621.nc</map>
     </gridmap>
 
+    <gridmap lnd_grid="ne4np4.pg2" rof_grid="r05">
+      <map name="LND2ROF_FMAPNAME">cpl/gridmaps/ne4pg2/map_ne4pg2_to_r05_mono.200527.nc</map>
+      <map name="ROF2LND_FMAPNAME">cpl/gridmaps/ne4pg2/map_r05_to_ne4pg2_mono.200527.nc</map>
+    </gridmap>
+
     <gridmap lnd_grid="ne11np4" rof_grid="r05">
       <map name="LND2ROF_FMAPNAME">lnd/clm2/mappingdata/maps/ne11np4/map_ne11np4_TO_r05_aave.160621.nc</map>
       <map name="ROF2LND_FMAPNAME">lnd/clm2/mappingdata/maps/ne11np4/map_r05_TO_ne11np4_aave.160621.nc</map>
--- a/cime/config/e3sm/config_grids.xml
+++ b/cime/config/e3sm/config_grids.xml
@@ -794,6 +794,16 @@
       <mask>oQU240</mask>
     </model_grid>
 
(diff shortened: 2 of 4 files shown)


jonbob:
Passes SMS.ne4pg2_r05_oQU480.A_WCYCL1850.anvil_intel but fails SMS.ne16pg2_r05_oQU240.A_WCYCL1850.anvil_intel with a CAM error:
````
 CAM_PIO_CHECK_VAR INFO: variable PHIS_d is not on file
 reading FV PHIS on gridname physgrid_d
 CAM_PIO_CHECK_VAR INFO: variable PHIS is not on file
 ERROR: Could not find PHIS field on input datafile
````

**Comment by rljacob** [comment]
@ambrad does ne16pg2 not have a PHIS field in its input?

**Comment by ambrad** [comment]
@rljacob, there's no ne16pg2 topo file. I can make one and then have @jonbob add the XML for it in this PR. Good?

**Comment by ambrad** [comment]
@rljacob @mt has there been a decision on whether ne16pg2.g, ne16.g, ne16pg2_scrip.nc (in this case) should go into the atm grids section of the inputdata server? I believe Mark's argument is a tool traditionally has produced these without a problem; supporting this argument, the grid directory contains essentially all RRM-related grid data. I believe Rob's argument is different versions of the tool -- indeed, even different builds, depending on flags -- can change the bits. Resolution?

**Comment by rljacob** [comment]
Yes that works.

**Comment by ambrad** [comment]
@jonbob, I've created the two files you are missing and put them on the input data server:
```
inputdata/atm/cam/chem/trop_mam/atmsrf_ne16pg2_200527.nc
inputdata/atm/cam/topo/USGS-gtopo30_ne16np4pg2_16xdel2_20200527.nc
```
To use them, insert these XML entries in namelist_defaults_cam.xml:
```
<drydep_srf_file hgrid="ne16np4" npg="2">atm/cam/chem/trop_mam/atmsrf_ne16pg2_200527.nc</drydep_srf_file>
<bnd_topo hgrid="ne16np4" npg="2">atm/cam/topo/USGS-gtopo30_ne16np4pg2_16xdel2_20200527.nc</bnd_topo>
```

**Comment by rljacob** [comment]
My thinking is that we need to save any files that fed in to a production run.  So that's mostly about ne30pg2.  You don't have to save them for these test resolutions.

**Comment by mt5555** [comment]
with this PR, I think if we add 👍 
`
    <gridmap lnd_grid="ne4np4.pg2" rof_grid="r05">
      <map name="LND2ROF_FMAPNAME">cpl/gridmaps/ne4pg2/map_ne4pg2_to_r05_mono.200527.nc</map>                           
      <map name="ROF2LND_FMAPNAME">cpl/gridmaps/ne4pg2/map_r05_to_ne4pg2_mono.200527.nc</map>                           
    </gridmap>
`

then we will also have  ne4pg2_ne4pg2 and ne4pg2_oQU480 "dual grid" compsets?

**Comment by jonbob** [comment]
Thanks @ambrad - I'll get those new file designations added to my branch and retest. By the way, I did create the mapping files using your executables on compy and following the steps you had documented. Thanks for all the help

**Comment by jonbob** [comment]
@mt5555 - thanks for pointing that out. I'll add that code snippet and see if those two dual-grid configurations also work now.

**Comment by jonbob** [comment]
@ambrad - just a quick question. I added the new files to namelist_defaults_cam.xml and was just curious about this section of it:
````
<drydep_srf_file hgrid="ne4np4"  >atm/cam/chem/trop_mam/atmsrf_ne4np4_c151204.nc</drydep_srf_file>
<drydep_srf_file hgrid="ne4np4" npg="2">atm/cam/chem/trop_mam/atmsrf_ne4pg2_200527.nc</drydep_srf_file>
<drydep_srf_file hgrid="ne11np4" >atm/cam/chem/trop_mam/atmsrf_ne11np4_c151204.nc</drydep_srf_file>
<drydep_srf_file hgrid="ne16np4" >atm/cam/chem/trop_mam/atmsrf_ne16np4_110920.nc</drydep_srf_file>
<drydep_srf_file hgrid="ne16np4" npg="2">atm/cam/chem/trop_mam/atmsrf_ne16pg2_200527.nc</drydep_srf_file>
<drydep_srf_file hgrid="ne30np4" npg="0">atm/cam/chem/trop_mam/atmsrf_ne30np4_110920.nc</drydep_srf_file>
<drydep_srf_file hgrid="ne30np4" npg="2">atm/cam/chem/trop_mam/atmsrf_ne30pg2_200129.nc</drydep_srf_file>
````
Should I add 'npg="0"' where that is not specified? Or is it OK as is?

**Comment by jonbob** [comment]
With updated EAM default, this branch still passes SMS.ne4pg2_r05_oQU480.A_WCYCL1850.anvil_intel and now passes SMS.ne16pg2_r05_oQU240.A_WCYCL1850.anvil_intel as well (with a tweak to the mpassi timestep). Thanks @ambrad and @rljacob

**Comment by jonbob** [comment]
@rljacob - I do have a question for you, concerning mpassi time steps. We currently use oQU240 as the ice/ocn grid for tests using atm resolutions including ne4, ne4pg2, ne11, and ne16pg2 -- and the EAM default timestep is different for each of those. But the mpassi timestep has to match the coupling frequency, and I don't know how to make the mpassi default_namelist settings aware of the EAM resolution or corresponding coupling frequency. Any ideas?

**Comment by jonbob** [comment]
@mt5555 - I tested with the code snippet and got these results:
* I had to also add a grid alias for the ne4pg2_oQU480 configuration;
* the A_WCYCL1850 smoke test for the ne4pg2_oQU480 dual-grid fails because ne4pg2 is not supported in CLM/ELM. I can see that ne4np4 has a different number of horizontal grid points than ne4pg2, so I'm assuming we can't tell ELM to run on the ne4np4 grid; and
* it looks like ne4pg2_ne4pg2 is already defined, but probably breaks in ELM for the same reason.
So we would have to ask ELM to support another resolution, which I think is one of the reasons we moved to the trip-grid. Do you want me to push on this?

**Comment by ambrad** [comment]
@jonbob re: npg=0: It is OK as is.

**Comment by jonbob** [comment]
Thanks @ambrad

**Comment by mt5555** [comment]
thanks @jonbob, and sorry for the distraction.  

I was able to run ne4pg2_ne4pg2  - i'll try ne4pg2_oQU480 out of curiosity...

**Comment by jonbob** [comment]
@mt5555 - ne4pg2_ne4pg2 ran with active lnd? or dlnd? Please let me know if you get something you want to include in this PR

**Comment by mt5555** [comment]
yes - it runs with active land.  So it looks like CLM can run on the ne4pg2 grid (although with finidat='').   Not sure why it doesn't run with either oQU240 or oQU480.

**Comment by whannah1** [comment]
@jonbob why wouldn't ELM be able to run on the ne4pg2 grid?

**Comment by jonbob** [comment]
@whannah1 - it would run fine, it just doesn't have all of the forcing files defined for that resolution

**Comment by whannah1** [comment]
@jonbob oh ok, I thought maybe you were saying it couldn't be configured to run on that grid for some reason.

**Comment by jonbob** [comment]
@whannah1 - I'm sorry if that wasn't clear. ELM has lots of files required to run on any given resolution and it's a hardship for them to support too many grids

**Comment by mt5555** [comment]
this PR enables ne4pg2_ne4pg2 compsets, at least with some F cases, which is a nice bonus!

no need to further delay for other dual grid setups - they can be added later if needed.

**Comment by jonbob** [comment]
Excellent @mt5555 -- do we need the 
````
<gridmap lnd_grid="ne4np4.pg2" rof_grid="r05"> 
  <map name="LND2ROF_FMAPNAME">cpl/gridmaps/ne4pg2/map_ne4pg2_to_r05_mono.200527.nc</map> 
  <map name="ROF2LND_FMAPNAME">cpl/gridmaps/ne4pg2/map_r05_to_ne4pg2_mono.200527.nc</map> 
</gridmap>
````
to enable the ne4pg2_ne4pg2 compsets?

**Comment by mt5555** [comment]
yes, please leave that in.   The "dual-grid" setups need that, since ROF is always on the r05 grid.

**Comment by rljacob** [comment]
@jonbob I can't recall how mpassi generates its "namelist" but the buildnml script should be able to query CIME for the coupling frequency and ATM_GRID is also in the case xml.

**Comment by jonbob** [comment]
OK @rljacob -- let me see what I can come up with...

**Comment by jonbob** [comment]
@rljacob - I have something working for the mpassi timestep. It still passes both tests, despite having the default timestep for the ocn/ice grid set to something else.

**Comment by jonbob** [comment]
@rljacob - any sense for what tests to add to the suites with this PR? And if we should replace any of the old ones that are not tri-grid?

**Comment by rljacob** [comment]
Lets see if we can replace this test in developer:
SMS_P12x2.ne4_oQU240.A_WCYCL1850.melvin_gnu.allactive-mach_mods

With the ne4 tri-grid.

**Comment by jonbob** [comment]
@rljacob - the following tests all fail, with problems in the lnd:
* SMS_P12x2.ne4pg2_r05_oQU480.A_WCYCL1850.anvil_intel.allactive-mach_mods
* SMS_P16x2.ne4pg2_r05_oQU480.A_WCYCL1850.anvil_intel.allactive-mach_mods
* SMS_P24x2.ne4pg2_r05_oQU480.A_WCYCL1850.anvil_intel.allactive-mach_mods
However, this one passes:
* SMS_P36x2.ne4pg2_r05_oQU480.A_WCYCL1850.anvil_intel.allactive-mach_mods
So I'm not sure we can replace the old dual-grid test with the tri-grid, unless we're OK throwing more processors at it.

I'll try to understand the problem in ELM. I don't think it's simply a memory issue, since it runs unthreaded on 12 pes. But nothing obvious right away...

**Comment by rljacob** [comment]
@jonbob when you say "since it runs unthreaded on 12 pes." you mean SMS_P12x1.ne4pg2_r05_oQU480.A_WCYCL1850.anvil_intel.allactive-mach_mods runs fine?

I would like to merge this since the grids/maps are useful and we can figure out later how to add tests.

**Comment by rljacob** [comment]
Why is this labeled non-bfb?

**Comment by jonbob** [comment]
Sorry -- the non-BFB label was a mis-click. And yes, SMS_P12x1.ne4pg2_r05_oQU480.A_WCYCL1850.anvil_intel.allactive-mach_mods ran fine, which made the P12x2 (and P24x2) failures worrisome. In any case, I couldn't use this to replace SMS_P12x2.ne4_oQU240.A_WCYCL1850.melvin_gnu.allactive-mach_mods as requested, though I know there are r1 and r2 lnd/rof meshes in the works. I'll get this committed then

**Comment by jonbob** [comment]
merged to next

File: components/cam/bld/namelist_files/namelist_defaults_cam.xml [hunk 1]
```diff
@@ -229,9 +229,10 @@
 <bnd_topo hgrid="10x15"     >atm/cam/topo/USGS-gtopo30_10x15_remap_c050520.nc</bnd_topo>
 
 <bnd_topo hgrid="ne4np4"   npg="0">atm/cam/topo/USGS-gtopo30_ne4np4_16x.c20160612.nc</bnd_topo>
-<bnd_topo hgrid="ne4np4"   npg="2">atm/cam/topo/USGS-gtopo30_ne4pg2_16xdel2-PFC-consistentSGH.c20190618.nc</bnd_topo>
+<bnd_topo hgrid="ne4np4"   npg="2">atm/cam/topo/USGS-gtopo30_ne4np4pg2_16x.c20160612_converted.nc</bnd_topo>
```
rljacob, line 232: This leads to BFB diffs.   Was this intended?

whannah1: We shouldn't be adding redundant topo files. Was the one that was already causing a problem?

jonbob: It was not intended -- I can remove that one.

jonbob:
@rljacob, @whannah1 - please see the conversation with @ambrad above to see where the bnd_topo file came from

rljacob:
inputdata/atm/cam/topo/USGS-gtopo30_ne4np4pg2_16x.c20160612_converted.nc should have had its date string updated to reflect that it is a new/altered file.

components/cam/bld/namelist_files/namelist_defaults_cam.xml
rljacob: Don't remove it.  But I think we should rename it if this is the only place it appears.

rljacob:
@ambrad please change the names of the new files you created so that the date string is the date you made/changed them.

ambrad:
@rljacob do you mean that I should simply rename the file on the blues inputdata server, or something more?

rljacob:
@ambrad just rename them on the blues server.  Then this PR needs to be updated to use the new names.

ambrad:
@rljacob OK, thanks. Optionally, if you prefer, we can use the old file. Walter above is correct: this brings the ne4pg2 topo file in line with the new topo file format, providing optimal data on pg2 and ne4 grids. But the original format is still supported, and for ne4, topo quality doesn't really matter.

whannah1:
I think it's worth using the new file so that our test configurations are consistent with our production configurations that use the new topo format. I realize there isn't a practical benefit to using the new format here, but that's my vote.

ambrad:
@jonbob I've moved the new ne4pg2 topo file to USGS-gtopo30_ne4np4pg2_16x_converted.c20200527.nc, where c20200527 reflects that I made this file on 27 May, so the line of XML that refers to this file should be changed to use this new name. Thanks.

jonbob:
Thanks @ambrad

rljacob:
Thanks this all looks good.  @jonbob you can re-merge to next.

jonbob:
re-merged to next

rljacob:
@whannah1 do you know why ERS_Ln9_P96x1.ne4pg2_ne4pg2.F-MMF1-RCEMIP.sandiatoss3_intel didn't show any diffs from this PR?

whannah1:
@rljacob That RCE test doesn't use topography, mapping, or domain files, so the lack of diffs is expected.

jonbob:
merged to master and expected ne4pg2 DIFFs blessed

PR #3607, opened 2020-05-28
Ensure sync is called before freedecomp

### LINKED ISSUE #3605: pio_freedecomp aborts when trying to sync a closed file
A 5 month run (which triggers CICE h0 file writing) of F2010SC5-CMIP6 + ne4pg2_ne4pg2 crashes in Scorpio (but not in Scorpio classic), with error message:

59: PIO: FATAL ERROR: Aborting... An error occured, Syncing file (ncid=80) failed. Invalid file id. Unable to find internal structure associated with the file id. err=-33.

Problem seems to be that pio_freedecomp() when called with this interface, will call pio_syncfile(), but as the file has already been closed, this aborts.  

@jayeshkrishna : what's the correct fix?  should we call pio_freedecomp() before closing the file?  Or maybe pio_freedecomp() should not be calling pio_syncfile?  
Another option woudl be to use "call pio_freedecomp(File%iosystem,iodesc2d)" which avoids the pio_syncfile() call, but then the code segfaults later.  

code is in cice, ice_history_write.F90

`
      !-----------------------------------------------------------------
      ! close output dataset
      !-----------------------------------------------------------------

      call  pio_closefile(File)
      if (my_task == master_task) then
         write(nu_diag,*) ' '
         write(nu_diag,*) 'Finished writing ',trim(ncfile(ns))
      endif
      
      ! -------------------------
      ! clean-up PIO descriptors
      ! -------------------------
      call pio_freedecomp(File,iodesc2d)
      call pio_freedecomp(File,iodesc3d)
`

jayeshkrishna:
Freeing decompositions in CICE history write routine
before the file close.

Fixes #3605 

[BFB]

### DIFF
--- a/components/cice/src/drivers/cpl/ice_history_write.F90
+++ b/components/cice/src/drivers/cpl/ice_history_write.F90
@@ -621,6 +621,12 @@
          endif
       enddo ! num_avail_hist_fields
 
+      ! -------------------------
+      ! clean-up PIO descriptors
+      ! -------------------------
+      call pio_freedecomp(File,iodesc2d)
+      call pio_freedecomp(File,iodesc3d)
+
       !-----------------------------------------------------------------
       ! close output dataset
       !-----------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


jonbob:
merged to next

jonbob:
merged to master

PR #3608, opened 2020-05-28
theta-l dycore fixes: POTTEMP initialization, default ne4 nu_div value

mt5555:
Dont neglict moisture when initializing theta-l dycore's virtural potential temperature
variable.

Also added a T/theta conversion consistency check of the initial condition.  

theta-l ne4 was accidently picking up the preqx value of nu_div=2.5*nu.  theta-l should
be using nu_div=nu.

tested by running for 1 year using the F2010SC5-CMIP6 compset on ne4np4_ne4np4 and ne4pg2_ne4pg2 grids with theta-l dycore in hydrostatic mode and nonhydrostatic mode (all 4 combinations run for 1 year).  

Running ne4pg2_ne4pg2 required PR #3606 and uses CLM cold start.

[non-BFB] but only for theta-l tests.

### DIFF
--- a/components/homme/src/theta-l/element_ops.F90
+++ b/components/homme/src/theta-l/element_ops.F90
@@ -398,7 +398,7 @@ recursive subroutine get_field(elem,name,field,hvcoord,nt,ntQ)
 
 
   !_____________________________________________________________________
-  subroutine set_thermostate(elem,ps,temperature,hvcoord)
+  subroutine set_thermostate(elem,ps,temperature,hvcoord,Q1)
   !
   ! Assuming a hydrostatic intital state and given surface pressure,
   ! and no moisture, compute theta and phi 
@@ -412,6 +412,7 @@ recursive subroutine get_field(elem,name,field,hvcoord,nt,ntQ)
   real (kind=real_kind), intent(in) :: temperature(np,np,nlev)
   type (hvcoord_t),     intent(in)  :: hvcoord                      ! hybrid vertical coordinate struct
   real (kind=real_kind), intent(in) :: ps(np,np)
+  real (kind=real_kind), intent(in), optional :: Q1(np,np,nlev)
   
   !   local
   real (kind=real_kind) :: p(np,np,nlev)
@@ -433,11 +434,16 @@ recursive subroutine get_field(elem,name,field,hvcoord,nt,ntQ)
   enddo
   elem%state%ps_v(:,:,nt)=ps
 
+
+  if (present(Q1)) then
+     call get_R_star(dp,elem%state%Q(:,:,:,1))  ! compute Rstar, store in dp
+     elem%state%vtheta_dp(:,:,:,nt)=elem%state%vtheta_dp(:,:,:,nt)*dp(:,:,:)/Rgas
+  endif
+
 !set phi, copy from 1st timelevel to all
   call tests_finalize(elem,hvcoord)
 
   end subroutine set_thermostate
-
 
 
   !_____________________________________________________________________
--- a/components/homme/src/theta-l/element_ops.F90
+++ b/components/homme/src/theta-l/element_ops.F90
@@ -436,7 +436,8 @@ recursive subroutine get_field(elem,name,field,hvcoord,nt,ntQ)
 
 
   if (present(qv)) then
-     call get_R_star(dp,qv(:,:,:))! get ideal gas constant for moist air
+     ! get ideal gas constant for moist air, store in variable "dp"
+     call get_R_star(dp,qv(:,:,:))
      elem%state%vtheta_dp(:,:,:,nt)=elem%state%vtheta_dp(:,:,:,nt)*dp(:,:,:)/Rgas
   endif
 
--- a/components/homme/src/theta-l/element_ops.F90
+++ b/components/homme/src/theta-l/element_ops.F90
@@ -417,6 +417,7 @@ recursive subroutine get_field(elem,name,field,hvcoord,nt,ntQ)
   !   local
   real (kind=real_kind) :: p(np,np,nlev)
   real (kind=real_kind) :: dp(np,np,nlev)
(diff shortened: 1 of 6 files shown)


mt5555:
@whannah1 : will you test this to confirm @ambrad conjecture that this will improve the forecast skill when using theta-l dycore?

components/homme/src/theta-l/element_ops.F90
```diff
@@ -412,6 +412,7 @@ subroutine set_thermostate(elem,ps,temperature,hvcoord)
   real (kind=real_kind), intent(in) :: temperature(np,np,nlev)
   type (hvcoord_t),     intent(in)  :: hvcoord                      ! hybrid vertical coordinate struct
   real (kind=real_kind), intent(in) :: ps(np,np)
+  real (kind=real_kind), intent(in), optional :: Q1(np,np,nlev)
```
whannah1, line 415: Minor comment, I think naming this variable something like "qv" would be clearer just because I associate "Q1" with a heating tendency. At the very least, maybe we can just add a descriptive comment here saying that this should be water vapor.

components/homme/src/theta-l/element_ops.F90
```diff
@@ -433,13 +434,18 @@ subroutine set_thermostate(elem,ps,temperature,hvcoord)
   enddo
   elem%state%ps_v(:,:,nt)=ps
 
+
+  if (present(Q1)) then
+     call get_R_star(dp,Q1(:,:,:))  ! compute Rstar, store in dp
```
whannah1, line 439: "R" refers to a gas constant, correct? Might be good to clarify that with a comment describing what's happening here.

AaronDonahue:
@whannah1 , should I wait until those hindcasts are finished before integrating?

whannah1:
@AaronDonahue yea let's wait. I'm having trouble getting an SL case to run, I must be configuring the dycor time step or hyperviscosity incorrectly.

components/homme/src/theta-l/element_ops.F90
mt5555: fixed.

components/homme/src/theta-l/element_ops.F90
mt5555: fixed.

whannah1:
I ran two hindcasts and calculated the skill scores. The skill of the moisture field is very much improved over what I saw previously, so that's good. I'm a bit surprised about the T850 field though. 

Could it be that this difference is expected? 
Or is the addition of moisture into the initial state messing up the temperature via some energy conservation check?

image

mt5555:
The time=0 data for T850 suggests there is still an issue with this PR.  With the bug fix, I thought the initial temperature should be identical between the two dycores (and thus have the same skill score?.  Let me verify that before proceeding with this PR.

whannah1:
Looking deeper into the temperature difference, we thought the lower T850 skill score could be due to the gravity wave response in conjunction with the fact that the SL case is less dissipitive. So here's some maps of the first 3-hourly output of T850 and OMEGA850 (on the physgrid):
image

And here's the difference (SL - PREQX):
image

It worries me a bit that the temperature difference looks qualitatively similar to the moisture field. If the skill score was due to larger initial gravity wave amplitude then I would expect the difference to be more concentrated on small scales.

ambrad:
You might want to separate theta-l from SL effects by running with
```
transport_alg = 0
dt_tracer_factor = 1
hypervis_subcycle_q = 1
```

whannah1:
@ambrad I keep forgetting that SL and Theta-L are not synonymous. Using the options you suggested doesn't change the result though.

image

components/homme/src/theta-l/element_ops.F90
```diff
@@ -433,13 +434,19 @@ subroutine set_thermostate(elem,ps,temperature,hvcoord)
   enddo
   elem%state%ps_v(:,:,nt)=ps
 
+
+  if (present(qv)) then
+     ! get ideal gas constant for moist air, store in variable "dp"
+     call get_R_star(dp,qv(:,:,:))
```
whannah1, line 440: Can we change the "dp" variable to "Rstar"?
As I was looking into this I keep getting confused about what's happening on these lines.

AaronDonahue:
@whannah1 , is this PR ready to be merged?

mt5555:
No, not ready.  Need to fix the discrepancy in the initial condition differences.

whannah1:
@mt5555 I thought we decided to go ahead and merge this because the initial condition issue appeared to be a separate bug?

mt5555:
Your latest np4 results convinced me there is still something that needs to be fixed in the theta-l initial condition code.  This PR is probably ok, but lets wait until we fix the initial condition problem just to minimize the number of non-BFB PRs.

mt5555:
Potential fix:  HOMME's "use_moisture" flag was set in prim_init2(), after EAM read the initial condition.  Thus caused the T->virtual potential temperature conversion to be initially use the dry formula (since use_moisture was not yet set to .true.).

whannah1:
This latest change fixes the difference in the initial condition. Here's a map of the first time step output of T850 and Q850, which the lower panels showing the difference (theta-l - preqx):

image

I only ran a single day, but the 3-hourly RMS error relative to ERA5 data:

image

I wouldn't put much weight on the difference in the T850 error, but it is nice to see that it's a tiny bit lower for theta-l. Either way, I think it's safe to conclude that the initial states are now effectively identical between preqx and theta-l.

AaronDonahue:
per correspondence with @mt5555 , this PR is ready to merge.

AaronDonahue:
@mt5555, the DIFF's I'm seeing for all of the thetahy tests are due to this PR correct?

mt5555:
yes, all the new DIFFs are in the theta dycore cases, as expected.

PR #3613, opened 2020-06-01
Update mpas-source: analysis member update, remove unused modules

mark-petersen:
This PR includes several updates to analysis members, and removes modules that are not used in E3SM simulations:
1. Adds normalVelocity arrays to eddy products https://github.com/MPAS-Dev/MPAS-Model/pull/544
2. Adds Ocean Heat Content AM https://github.com/MPAS-Dev/MPAS-Model/pull/455
3. Ocean/remove unused modules https://github.com/MPAS-Dev/MPAS-Model/pull/569

It includes changes to E3SM namelists and streams files for MPAS-Ocean. COMPASS changes associated with these PRs are also included.

[NML]
[BFB]


mark-petersen:
@jonbob when you run the namelist script, you should see these namelist groups go away:
```
vmix
vmix_const
vmix_rich
vmix_tanh
hmix_del2_tensor
hmix_del4_tensor
```
and this one is new:
```
&AM_oceanHeatContent
    config_AM_oceanHeatContent_enable = .true.
    config_AM_oceanHeatContent_compute_interval = '0000-00-00_01:00:00'
    config_AM_oceanHeatContent_output_stream = 'oceanHeatContentOutput'
    config_AM_oceanHeatContent_compute_on_startup = .true.
    config_AM_oceanHeatContent_write_on_startup = .false.
/
```

The streams file will include these new variables in the monthly time average:
```
normalGMBolusVelocityTimesTemperature
normalGMBolusVelocitySquared
normalVelocityTimesTemperature
normalVelocitySquared
oceanHeatContentSFCtoBot
oceanHeatContentSfcto700m
oceanHeatContent700mto2000m
oceanHeatContent2000mtoBot
```
and will need a new stream, similar to the other AMs:
```
<stream name="oceanHeatContentOutput"
        type="output"
        precision="single"
        io_type="pnetcdf"
        filename_template="mpaso.hist.am.oceanHeatContent.$Y-$M-$D.nc"
        filename_interval="00-01-00_00:00:00"
        reference_time="01-01-01_00:00:00"
        output_interval="00-01-00_00:00:00"
        clobber_mode="truncate"
        packages="oceanHeatContentAMPKG">

    <var name="xtime"/>
    <var name="oceanHeatContentSfcToBot"/>
    <var name="oceanHeatContentSfcTo700m"/>
    <var name="oceanHeatContent700mto2000m"/>
    <var name="oceanHeatContent2000mtoBot"/>
</stream>
```
@vanroekel do you want `AM_oceanHeatContent` default on for all resolutions, and at what for computation and output? I copied everything from `AM_eddyProductVariables` in the notes above, you can just double-check that.

mark-petersen:
The GM variables
```
normalGMBolusVelocityTimesTemperature
normalGMBolusVelocitySquared
``` 
can go with the other GM variables for monthly averaging under
```
components/mpas-ocean/cime_config/buildnml
828       if not ocn_grid.startswith("oRRS1"):
```

jonbob:
@mark-petersen - did you add the new AM file to the cmake list?

jonbob:
@mark-petersen - do you expect the group "partial_bottom_cells" to go away?

xylar:
>  do you expect the group "partial_bottom_cells" to go away?

Yes, that's expected, because I got moved to "init" mode.

jonbob:
Thanks @xylar

jonbob:
@mark-petersen - the settings you listed for the new AM do not correspond with the default from Registry. What do you want them to be?

vanroekel:
@jonbob I would say there is no need to compute or write on startup.  The other options seem fine as is in @mark-petersen comment

jonbob:
Thanks @vanroekel

jonbob:
@mark-petersen - you will have to add the new analysis member file to ocean.cmake. My tests do not build without that change. Will that fix, I did just get a smoke test SMS.ne4_oQU240.A_WCYCL1850.anvil_intel to pass

jonbob:
A longer test shows that mpaso.hist.am.oceanHeatContent.00XX-XX-XX.nc files are being successfully created

mark-petersen:
OK, cmake line is corrected. Yes, partial_bottom_cells namelist was only for initial conditions, so was removed here.

mark-petersen:
@jonbob could you recheck with the last commit? I added the last six to the monthly time average:
```
normalGMBolusVelocityTimesTemperature
normalGMBolusVelocitySquared
normalVelocityTimesTemperature
normalVelocitySquared
oceanHeatContentSFCtoBot
oceanHeatContentSfcto700m
oceanHeatContent700mto2000m
oceanHeatContent2000mtoBot
```
@vanroekel can you confirm that you want all these variables in the monthly time average as the default behavior? The first four are 3D fields, so large. It is fine, but I wanted to be sure. First two only appear if GM is on.

jonbob:
@mark-petersen - no, that last commit caused the tests to fail:
````
ERROR: Requested field normalVelocitySquared not available
````
I grepped the mpas-source and could not find that array. We'll pick this up tomorrow...

**Comment by vanroekel** [comment]
@mark-petersen yes it would be good to have all normal velocity fields.  For the OHC fields, I'd like to know what @milenaveneziani or @xylar think.  These fields are also computed within MPAS-Analysis so I don't know how much we save by computing these as the model runs.  My sense is this is not a heavy calculation for mpas-analysis.

**Comment by xylar** [comment]
@milenaveneziani would know better than I do which product terms would be useful.  But any products we compute in MPAS-Analysis are after time averaging, so not very accurate even if they're fast.  My guess is that we would use the products if they were available.

**Comment by milenaveneziani** [comment]
I agree with what @xylar says: not a heavy calculation for MPAS-A, but averaging the products is more accurate than doing it post-processing. And besides, they are 4 2-d fields, so I think we are fine with adding them to the monthly stream.

**Comment by mark-petersen** [comment]
This passes 
```
```
@jonbob this is ready if your tests also pass.

jonbob:
Passed SMS.ne4_oQU240.A_WCYCL1850.anvil_intel.
Merged to next

mark-petersen:
rebased, added https://github.com/MPAS-Dev/MPAS-Model/pull/589. Ready to merge.

jonbob:
Re-merged to next. 

Failed: SMS_D.ne4_oQU240.A_WCYCL1850.anvil_intel due to EAM problems
Passed: SMS_D.T62_oQU240.GMPAS-IAF.anvil_intel

jonbob:

Issue #3588, opened 2020-05-21
Mass conservation with pg2 grids


beharrop:
While testing CO2 conservation with the v2alpha-clubbv2 branch, I noticed that the mass conservation is not maintained with the ne4pg2 grid, but is with ne4np4.  I have been using CO2_FFF (fossil fuel flux CO2 tracer) to test for mass conservation because when there is no emissions file prescribed, the fluxes are exactly zero and conservation should be maintained (to machine level round off).  To demonstrate the conservation issue, I have attached a figure showing the relative error in the global mean CO2_FFF from its starting value using ne4pg2 and ne4np4 grids.  I have also copied the printouts of the global dry mass of that tracer from the gmean_mass calls for the two grids.

I am tagging a few people that I think might be interested @wlin7 , @susburrows , @kvcalvin , @kaizhangpnl , @whannah1 , @jonbob .  These runs were done on Compy, intel compiler, and FC5AV1C-04P2 compset (with CAM_CONFIG_OPTS="-co2_cycle" added to enable CO2).

![image](https://user-images.githubusercontent.com/17884885/82584425-54739400-9b49-11ea-95a5-979514e3f391.png)

				ne4np4				ne4pg2
m=42name=CO2_FFF		5.6285784223748E+00		5.6282525870713E+00
m=42name=CO2_FFF		5.6285784223748E+00      	5.6282525870669E+00
m=42name=CO2_FFF 		5.6285784223748E+00      	5.6282525870615E+00
m=42name=CO2_FFF 		5.6285784223748E+00      	5.6282525870628E+00
m=42name=CO2_FFF 		5.6285784223748E+00      	5.6282525870583E+00
m=42name=CO2_FFF 		5.6285784223748E+00      	5.6282525870570E+00
m=42name=CO2_FFF 		5.6285784223748E+00      	5.6282525870590E+00
m=42name=CO2_FFF 		5.6285784223748E+00      	5.6282525870572E+00
m=42name=CO2_FFF 		5.6285784223748E+00      	5.6282525870523E+00
m=42name=CO2_FFF 		5.6285784223748E+00      	5.6282525870430E+00

whannah1:
@ambrad will also be interested in this.

whannah1:
[...]

PR #3594, opened 2020-05-25
v2 SL and physgrid remap performance, mass conservation

ambrad:
Improve vectorization and threading in SL transport, physgrid remap, and theta-l.

Fix a mass conservation error due to finite precision effects.

These code mods would cause no answer change in exact arithmetic but are non-BFB in finite precision.

Summary:
- Physgrid remap: Switch indexing from (:nf,:nf) to (:nf2) to improve vectorization.
- SL transport:
   * Speed up global CAAS by locally reducing what can be reduced decomp-BFB.
   * Improve own_q, rmt_q thread load balance.
   * Flip level, q indexing to improve memory locality.
   * After a performance regression to calc_q a few months ago to work around a compiler bug, address the performance issue in a different way. I avoid using pragma ivdep but get the performance that annotation provides by explicitly blocking. The result is effective vectorization of a key computation in SL transport.
- Fix dp_fv so it agrees numerically with the pdel calculation in dp_coupling.F90: derived_physics.
- Add a C++ block to the cori-knl/intel19 block in config_compilers.xml.
- Work around an Intel 17 -O3 bug to fix a standalone-Homme test on Anvil.

Fixes #3588.

[non-BFB] 4 EAM tests and 3 standalone Homme tests

PR #3614, opened 2020-06-01
EAM mass conservation test

ambrad:
Set up a theta-sl-pg2 ne4pg2 test with -co2_cycle. Use POSTRUN_SCRIPT to run a
python-2/3-compatible script to read "qv( 42)" and "dry M" output from Homme's
prim_state_mod. Tracer 42 is CO2_FFF in this test. Check mass conservation for
tracer 42 and dry air in the script. If they are fine, exit with 0; if bad, 1. A
non-0 exit in a POSTRUN_SCRIPT makes the RUN phase fail.

The new test is SMS_Ld9.ne4pg2_ne4pg2.FC5AV1C-04P2.cam-thetahy_sl_pg2_mass

Addresses verification part of #3588

[non-BFB] BFB for all existing tests, but adds one new test

### DIFF
--- a/components/homme/src/preqx/share/prim_state_mod.F90
+++ b/components/homme/src/preqx/share/prim_state_mod.F90
@@ -148,7 +148,7 @@ contains
     IEner_wet = 0
     ! dynamics timelevels
     n0=tl%n0
-    call TimeLevel_Qdp( tl, qsplit, n0q) !get n0 level into t2_qdp 
+    call TimeLevel_Qdp(tl, qsplit, n0q) ! get n0 level into n0q
 
     dt=tstep*qsplit
     if (rsplit>0) dt = tstep*qsplit*rsplit  ! vertical REMAP timestep 
--- a/cime/config/e3sm/tests.py
+++ b/cime/config/e3sm/tests.py
@@ -271,7 +271,8 @@ _TESTS = {
         "tests"    : (
                  "SMS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L.cam-thetahy_pg2",
                  "SMS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L.cam-thetahy_sl_pg2",
-                 "ERS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L.cam-thetahy_sl_pg2"
+                 "ERS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L.cam-thetahy_sl_pg2",
+                 "SMS_Ld9.ne4pg2_ne4pg2.FC5AV1C-04P2.cam-thetahy_sl_pg2_mass",
                  )
     },
     "e3sm_bench_hires_g" : {
--- a/cime/config/e3sm/tests.py
+++ b/cime/config/e3sm/tests.py
@@ -62,6 +62,7 @@ _TESTS = {
             "ERS_Ld5.ne4_ne4.FC5AV1C-L.cam-rrtmgp",
             "ERS_Ld5.ne4_ne4.FC5AV1C-L.cam-gust_param",
             "REP_Ln5.ne4_ne4.FC5AV1C-L",
+            "SMS_Ld9.ne4pg2_ne4pg2.FC5AV1C-04P2.cam-thetahy_sl_pg2_mass",
             )
         },
 
@@ -271,8 +272,7 @@ _TESTS = {
         "tests"    : (
                  "SMS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L.cam-thetahy_pg2",
                  "SMS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L.cam-thetahy_sl_pg2",
-                 "ERS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L.cam-thetahy_sl_pg2",
-                 "SMS_Ld9.ne4pg2_ne4pg2.FC5AV1C-04P2.cam-thetahy_sl_pg2_mass",
+                 "ERS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L.cam-thetahy_sl_pg2"
                  )
     },
     "e3sm_bench_hires_g" : {
--- a/components/cam/cime_config/testdefs/testmods_dirs/cam/thetahy_sl_pg2_mass/shell_commands
+++ b/components/cam/cime_config/testdefs/testmods_dirs/cam/thetahy_sl_pg2_mass/shell_commands
@@ -1,0 +1,5 @@
+#!/bin/bash
+./xmlchange CAM_TARGET=theta-l
+./xmlchange --append CAM_CONFIG_OPTS="-co2_cycle"
+./xmlchange RUN_STARTDATE=1950-02-14
+./xmlchange POSTRUN_SCRIPT="$CIMEROOT/../components/homme/utils/e3sm_test/check_mass_conservation.py"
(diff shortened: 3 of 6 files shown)


ambrad:
Only the last commit is new; the other commits will disappear once #3594 is merged to master.

ambrad:
@mt5555 I'm asking for your review to check if you're OK with the new homme/utils/e3sm_test directory to contain POSTRUN_SCRIPTs.

@susburrows @beharrop what do you think of this? Note that only commit https://github.com/E3SM-Project/E3SM/pull/3614/commits/07a55c8d60c104a79675d61eb5540ea54738f832 is part of this PR.

The script reads the atm.log file, parsing some of Homme's prim_printstate output, whose frequency is controlled by statefreq. The script produces output like this:
```
dry M               : mass rel err 1.996e-14 tol: 2.454e-13
tracer 42           : mass rel err 3.278e-16 tol: 2.454e-15
PASS
```
This is just to indicate what went wrong if something does. Otherwise, the RUN phase simply PASSes if the script detects mass was conserved. If the script detects a failure, it returns with a non-0 exit code, making the test's RUN phase FAIL. So it's fully automated for use in nightlies.

beharrop:
@ambrad, this seems reasonable to me.  I am not familiar with Homme's prim_printstate output, is dry M the dry air mass or the mass of a generic dry mixing ratio tracer?

ambrad:
@beharrop, dry M is the dry air, so `sum(spheremp*dp3d*(1 - qv))`, where qv is the specific humidity, tracer 1. In CAM, the equivalent variable is pdeldry.

ambrad:
I should add that `dry M` is conserved according to entirely different computations than CO2_FFF. The dynamics solver is responsible for the conservation of dry air. The tracer transport and physgrid remap solvers are responsible for the conservation of tracers like CO2_FFF.

beharrop:
@ambrad thanks for clarifying.

components/homme/src/preqx/share/prim_state_mod.F90
```diff
@@ -148,6 +148,7 @@ subroutine prim_printstate(elem, tl,hybrid,hvcoord,nets,nete)
     IEner_wet = 0
     ! dynamics timelevels
     n0=tl%n0
+    call TimeLevel_Qdp( tl, qsplit, n0q) !get n0 level into t2_qdp
```
mt5555, line 151: typo in comment "t2_qdp" -> n0q

ambrad: Thanks; fixed.

ambrad:
e3sm_developer shows the one extra test (43 now instead of 42 tests), and it DIFFs as expected. The RUN phase PASSes, showing the mass-conservation-checking script succeeds. I checked this new test on Cori KNL as well as sandiatoss3.

susburrows:
I thought I had posted a comment on this, but it looks like I didn't hit the button to post -- sorry for the delay!  Thanks for putting this test together so quickly -- it will be very useful to have this.  I did have a couple of questions about it though:

1. First, I am wondering if FC5AV1C-04P2 is the unambiguously right compset to use (and will this impose a maintenance burden if it needs to get updated to new composers in the future)?

2. Second, I am wondering if it is possible to reference the 'CO2_FFF' tracer by name instead of by index number.  The indexing could change in future model versions, which could mess up the test and require it to be revised in the future.

PeterCaldwell:
I thought that FC5AV1C-04P2 was deprecated in favor of F2010C5-CMIP6?

susburrows:
Yeah, thanks for adding that Peter - that's why I was thinking it might not be the right one to use, but I wasn't sure if we had an "official" stance on the canonical compset for atm-only.  I suspect the test could be done equally well with the 2010C5-CMIP6 compset, but am not certain - @ambrad and @beharrop , would it work to just swap them?

beharrop:
@susburrows I have never tried it with the F2010C5-CMIP6 compset.  In theory, so long as the `CO2_FFF` tracer is enabled, it is the 42nd tracer in `q`, and there are no fossil fuel emissions prescribed, then the test _should_ work.  So long as the compset chosen uses all of the code pieces we hope to test, I don't see it making much difference.

ambrad:
Great, thanks.
1. I'll switch to F2010C5-CMIP6.
2. Yes: I'll add some code to the python script to extract the index given the tracer name. But how can 42 not always be the right answer? :)

tangq:
A minor thing to note:

The name F2010C5-CMIP6 is the compset name on the maint-1.0 branch. The equivalent one on the master is F2010SC5-CMIP6. This is because of some naming issue with existing compset names when the compset was added.

susburrows:
@ambrad Thanks!  Hopefully fossil-fuel CO2 is not always the answer to life the universe and everything? :)

oksanaguba:
in next

oksanaguba:
SMS_Ld9.ne4pg2_ne4pg2.FC5AV1C-04P2.compy_pgi.cam-thetahy_sl_pg2_mass fails with 

1:  CNST_ADD: advected tracer index greater than pcnst =            40

I think you use at least 42 tracers in this test?

ambrad:
Yes, 44. I'll look at the logs. Perhaps the test is in a shared-build block but needs to have its own build.

oksanaguba:
Seems that it is the same issue for sandiatoss fails -- except that now build with 44 tracers was used but other tests do not like 44 tracers expecting 40. also, I do not have permissions (yet) to new Jenkins folders on sandiatoss, so, cannot see what happened to the python script. does its output need to go to cdash somehow?

ambrad:
Thanks @onguba. I looked, and the problem is indeed that I put the test in a block where multiple tests use the same 44-constituent build, causing errors in the 40-constituent runs. I can fix this today, or please feel free simply to revert this PR from next, and I'll switch this PR to WIP and you can remerge some time in the future. Let me know which you prefer.

oksanaguba:
Today would be great -- thanks! 
About cdash -- I do not see output from python in cdash logs that it points (it prob just dumps TestCase.log), but I do not know what to look for.

ambrad:
postrun.external.log.10410758.200623-202747.gz contains the python output. It looks like
```Using log file /nscratch/e3sm-jenkins/acme_scratch/sandiatoss3/SMS_Ld9.ne4pg2_ne4pg2.FC5AV1C-04P2.sandiatoss3_intel.cam-thetahy_sl_pg2_mass.C.JNextIntegration20200623_194336/run/atm.log.10410758.200623-202747
Tracer CO2_FFF has index 42
dry M               : mass rel err 1.996e-14 tol: 2.454e-13
tracer 42           : mass rel err 4.918e-16 tol: 2.454e-15
PASS
```
The RUN part of the run FAILs if the above FAILs.

ambrad:
For completeness: The script parses atm.log lines of the form
```
 qv(  8)=   0.0000000000000000E+00  0.8648745031418346E+07  0.5430258724952878E+10
 qv(  9)=   0.0000000000000000E+00  0.3956685110790033E-06  0.2299348841248413E-03
 qv( 10)=   0.0000000000000000E+00  0.7216257444267939E+00  0.3296552775124487E+04
```
where the integer index refers to the index in the "Advected constituent list" near the start of the atm.log file; the first real-number column is min mixing ratio; the second, max; the third, global constituent mass. We also have access to simulation time before the qv block:
```
 nstep=         960  time=   3.33333333333333       [day]
```

ambrad:
Using a similar script, we show global mass of a constituent as a time series:
mass-cons

PR #3617, opened 2020-06-03
Add oARRM60to10 tri-grid support

jonbob:
This PR adds support for ne30pg2_r05_oARRM60to10 and ne30pg2_r05_oARRM60to10_ICG configurations. The oARRM60to10 mesh was previously supported only for C- and G-compsets, so this adds fully-coupled capability.

[BFB]

### DIFF
--- a/cime/config/e3sm/config_grids.xml
+++ b/cime/config/e3sm/config_grids.xml
@@ -1054,6 +1054,16 @@
       <mask>oEC60to30v3</mask>
     </model_grid>
 
+    <model_grid alias="ne30pg2_r05_oARRM60to10_ICG" compset="(CAM5.+CLM.+MPASO%SPUNUP)">
+      <grid name="atm">ne30np4.pg2</grid>
+      <grid name="lnd">r05</grid>
+      <grid name="ocnice">oARRM60to10_ICG</grid>
+      <grid name="rof">r05</grid>
+      <grid name="glc">null</grid>
+      <grid name="wav">null</grid>
+      <mask>oARRM60to10</mask>
+    </model_grid>
+
     <model_grid alias="ne30_r05_oECv3_ICG" compset="(CAM5.+CLM.+MPASO%SPUNUP)">
       <grid name="atm">ne30np4</grid>
       <grid name="lnd">r05</grid>
@@ -1655,6 +1665,16 @@
       <mask>oEC60to30v3</mask>
     </model_grid>
 
+    <model_grid alias="ne30pg2_r05_oARRM60to10">
+      <grid name="atm">ne30np4.pg2</grid>
+      <grid name="lnd">r05</grid>
+      <grid name="ocnice">oARRM60to10</grid>
+      <grid name="rof">r05</grid>
+      <grid name="glc">null</grid>
+      <grid name="wav">null</grid>
+      <mask>oARRM60to10</mask>
+    </model_grid>
+
     <model_grid alias="ne30pg2_r05_ne30pg2" compset="(DOCN|XOCN|SOCN|AQP1)">
       <grid name="atm">ne30np4.pg2</grid>
       <grid name="lnd">r05</grid>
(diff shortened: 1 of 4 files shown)


jonbob:
passed test-ARRM.A_WCYCL1850S.ne30pg2_r05_oARRM60to10.anvil

stephenprice:
@jonbob, the 'cmip6' version of this is in the queue now on Grizzly. But the queues are so bad there right now that I won't likely know until next week (even a 5 day tests takes forever). I'll keep you posted.

stephenprice:
Sorry for the long delay on this. My initial 5-day test case failed on Griz., in the atm. w/ this 

error: set_time_float_from_date: error return from ESMF_TimeSet for set_time_float_from_date
ERROR: CHKRC

Let me know if that means anything to you (I suspect not). I will go back and check over the case setup, build, etc. again this morning.

jonbob:
@stephenprice - no worries. And what a shock, an ESMF problem! Please let me know if I should take a look at it as well...

stephenprice:
You did not see similar on Anvil I'm assuming. Does that suggest an issue w/ the version of ESMF on IC? I don't really have a good feeling for where to go from here. Is there somewhere it makes sense to look into this deeper within the build logs?

jonbob:
Or send me the path to your test and I can see what's different?

stephenprice:
The case directory is here: 
/lustre/scratch3/turquoise/sprice/e3sm_cases/20200604.A_WCYCL1850S_CMIP6.ne30pg2_r05_oARRM60to10_ICG.grizzly.intel.openmpi.master.lrgLayout

The build / run dirs are here:

/lustre/scratch4/turquoise/sprice/E3SM/scratch/20200604.A_WCYCL1850S_CMIP6.ne30pg2_r05_oARRM60to10_ICG.grizzly.intel.openmpi.master.lrgLayout

jonbob:
OK, so it doesn't really use ESMF -- just gives me nightmares seeing those letters! There's a fake esmf time_manager that I think was a brainchild of John Michalakes, just a way of claiming he was using esmf because he was using their api for time. Anyway, that's besides the point, except to say it's not a grizzly issue with any software version. But it does point at some issue with a time value somewhere, and there are plenty of calls in CAM to that routine. Could we try one thing quickly? And that's to turn off the cosp simulator -- you would just have to add this to user_nl_cam:
``` cosp_lite              = .false.```
Otherwise, I can try the cmip6 compset on anvil and see if I have the same issue? Or you could try the vanilla water-cycle compset on grizzly so we could more directly compare? Whichever one makes more sense to you...

stephenprice:
Thanks. I'll try that. Not easy to debug this on Griz. due to the queue, but I'll let you know ASAP.

stephenprice:
I can also build / submit the vanilla WC 1850S. I'm not sure cmip6 nuance is important to us ... as long as we can run a hist and rcp 8.5 from those init. conds.

jonbob:
Well, try the vanilla version on grizzly instead then and I'll see what the cmip6 one does on anvil? At least then we'll have a more complete set of information...

stephenprice:
If that is not too much of a pain for you, that would be great. Hopefully maybe it's useful for fleshing out any bugs in that compset that might be used with other grids.

stephenprice:
The CMIP6 compset run w/ the altered cam nl value is in the queue. When trying to build the the 'vanilla' 1850S on Griz. I get an error I have never seen before:

Out of memory allocating the message block.compilation aborted for /lustre/scratch3/turquoise/sprice/e3sm_repo/E3SM/components/clm/src/data_types/ColumnDataType.F90 (code 1)

No idea what to make of that one but will revisit it tomorrow.

stephenprice:
@jonbob -- unfortunately, adding that  cam nl option doesn't seem to change anything for me. I still see that same error at run time.

stephenprice:
To summarize: OOM errors seem to be node specific and something to do w/ the state Griz is in right now. Other runtime errors appear to have been due to input data (cam/chem files) being truncated during data download, leading to some time management errors. When the correct input data are used, the configuration runs as expected (cc'ing @wlin7 on the last bit of info).

jonbob:
merged to next

jonbob:
merged to master

PR #3618, opened 2020-06-03
make PSC T threshold a namelist variable

tangq:
This PR makes the hard-coded T threshold of PSC ozone loss a name list variable to facilitate model tuning. The default T threshold is set to 193 K, the same as the hard-coded value, so it won't change the model results by default.

(cherry picked from commit 17aa44dd9e5e05bd8491d3468dc45e32e9a2c0b9)


golaz:
@wlin7 and @rljacob  : can we go ahead and merge this?

wlin7:
Merged to next.

tangq:
Thank you, @wlin7

wlin7:
Merged to master.

PR #3620, opened 2020-06-04
Implement pcols as a runtime option in EAM

worleyph:
In the earlier 2000's, empirical experiments indicated that setting
the first dimension of the chunk data structure (pcols) at
compile-time allowed the compiler to generate more
performance-efficient code, and this motivated this design choice in
CAM. However, CAM/EAM has evolved significantly since then, and, for
example, physics_types now uses psetcols to define this first
dimension, and the compiler never sees a compile-time parameter. So
the compile-time setting of pcols may no longer serve a useful
purpose. Extensive experiments indicate that there is still a
small performance advantage from setting pcols at compile-time, but
also that being able to set pcols at runtime greatly simplifies the
process of optimizing over the pcols parameter.

Here pcols is implemented as a runtime option, currently specified
via the user_nl_cam namelist variable phys_chnk_fdim. phys_chnk_fdim
must be at least as large as 1, but there is no specified upper
bound. The current compile-time value of the CPP variable PCOLS is used
to define the default value for phys_chunk_fdim (and pcols). PCOLS is
now represented in EAM as the parameter ppcols, but is only referenced
in defining the default for phys_chunk_fdim.

To restore the compile-time setting of pcols, and also to specify
a pcols value of PCOLS, add -pcols to CAM_CONFIG_OPTS in
env_build.xml . For example, to specify 32 as the compile-time value
of pcols, add '-pcols 32' . This allows the recovery of any lost
performance from the runtime-specification of pcols.

Note that this is the first part of a two-part introduction of runtime
pcols. This PR sets the runtime pcols during the call to
read_namelist. To calculate a better default than the compile-time
PCOLS CPP variable requires that it be calculated in phys_grid_init,
and this requires changing the order of certain steps in the
initialization of EAM. The current PR, however, stands on its own, and
is acceptable even if the second part PR is never submitted.

[BFB]

This addresses some aspects of Issue #3538 , but not all. The issue
will be resolved by a second runtime pcols PR. If we decide
not to submit this follow-on PR, the issue will also be closed.

### DIFF
--- a/components/cam/src/physics/crm/crm_physics.F90
+++ b/components/cam/src/physics/crm/crm_physics.F90
@@ -44,12 +44,12 @@ module crm_physics
 
    type(physics_state)                     :: state_save
    type(physics_tend)                      :: tend_save
-   real(r8), dimension(pcols, pver)        :: cldo_save     ! saved cloud fraction
-   real(r8), dimension(pcols, pver, pcnst) :: q_aero        ! used to keep aerosol changes from offline physics
+   real(r8), dimension(:,:), allocatable   :: cldo_save     ! saved cloud fraction
+   real(r8), dimension(:,:,:), allocatable :: q_aero        ! used to keep aerosol changes from offline physics
 #ifdef MODAL_AERO
-   real(r8)          :: qqcw_save(pcols,pver,pcnst)
-   real(r8)          :: qqcw_all(pcols,pver,pcnst)
-   real(r8)          :: dgnumwet_save(pcols, pver, ntot_amode)
+   real(r8), dimension(:,:,:), allocatable :: qqcw_save
+   real(r8), dimension(:,:,:), allocatable :: qqcw_all
+   real(r8), dimension(:,:,:), allocatable :: dgnumwet_save
    real(r8),pointer  :: dgnumwet(:,:,:)
 #endif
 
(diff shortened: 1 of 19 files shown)


worleyph:
Hi @whannah1 , this is the second attempt at a PR that I submitted in April 13. It has been extensively reviewed, but by the time @singhbalwinder tried to merge it into next, your MMF PR has already gone in, and broke mine :-(. Could you please look quickly at the changes to the code in physics/crm and see whether it is acceptable to you?  I don't think that there is anything to worry about, but you are the expert. I did run some F-MMF1-TEST cases on Compy (ne4pg2_ne4pg2) and it worked, as far as I can tell. Not reproducible with respect to change in PE layout, but it diverges during nstep 8 (in atm.log), so same as master.

I'd really like to fast track this and get it in before I run into more conflicts, as I don't want to have to repeat this exercise - spent 7 hours today reimplementing the changes in master, and adding the needed code to crm/ . Thanks.

worleyph:
@whannah1  - sorry - of course you know all about this - I'm a little slap happy.

components/cam/src/physics/crm/crm_physics.F90
```diff
@@ -227,6 +237,23 @@ subroutine crm_physics_init(species_class)
    end if
 #endif
 
+   allocate (cldo_save(pcols,pver), stat=ierror)
```
whannah1, line 240: Is it ok that there's no deallocate statement for these module variables?

whannah1:
@worleyph sorry for the extra effort on your part. We're hoping we can reduce the amount of duplicate code needed to support the MMF after we solve some of our bigger issues, so this kind of thing will hopefully not happen again.

components/cam/src/physics/crm/crm_physics.F90
worleyph: These are "global" variables for the module, and the original declarations never "went away", so I assume that having these variables hang around until the end of the model run is the intent.

worleyph:
@singhbalwinder , ready to try again, when you have the time. Thanks.

singhbalwinder:
merged to next

worleyph:
Thanks @singhbalwinder .

PR #3623, opened 2020-06-05
Update mpas-source, compass only. Add WC14

mark-petersen:
COMPASS initial conditions only. Add Water Cycle 14 km case.

[bfb]


jonbob:
merged to next

jonbob:
merged to master

PR #3625, opened 2020-06-06
Add total column ozone output for diagnosing ozone hole

tangq:
Ozone hole diagnostics require daily output. This commit adds the inline calculation for total column ozone and the corresponding output field "TOZ" and thus reduces the cost of saving daily 3D "O3" to 2D "TOZ".

Additionally, the inline calculation is more accurate, because it uses instantaneous values instead of the averages, which is potentially important for ozone hole areas where large variances may occur.

[BFB] - Bit-For-Bit


wlin7:
Merged to next.

PR #3627, opened 2020-06-08
Update mpas-source: cvmix driver area weighted average bug

mark-petersen:
Fixes a bug in the area weighted averaging for 3 CVMix quantities. See https://github.com/MPAS-Dev/MPAS-Model/pull/556

[non-BFB]


mark-petersen:
Diff also shows COMPASS changes in previous PR #3623

jonbob:
merged to next

PR #3630, opened 2020-06-10
Add limiters on RRTMGP cloud and aerosol optics

brhillman:
Add limiters on RRTMGP cloud and aerosol optical properties. This prevents the internal RRTMGP sanity checks from aborting the run with these quantities go *slightly* outside the expected range. The behavior is now to always clip these values, but to throw an error and abort when they fall outside the expected range plus/minus an accepted tolerance. This fixes an error that arose in some coupled runs where single scattering albedo would go slightly above 1.0, possibly due to floating point precision issues.

[BFB]

### DIFF
--- a/components/cam/src/physics/rrtmgp/radiation.F90
+++ b/components/cam/src/physics/rrtmgp/radiation.F90
@@ -1353,12 +1353,12 @@ contains
 
                ! Check (and possibly clip) values before passing to RRTMGP driver
                if (rrtmgp_clip_optics) then
-                  call clip_values(cld_tau_gpt_sw,  0._r8, huge(cld_tau_gpt_sw), trim(subname) // ' cld_tau_gpt_sw', warn=.true.)
-                  call clip_values(cld_ssa_gpt_sw,  0._r8,                1._r8, trim(subname) // ' cld_ssa_gpt_sw', warn=.true.)
-                  call clip_values(cld_asm_gpt_sw, -1._r8,                1._r8, trim(subname) // ' cld_asm_gpt_sw', warn=.true.)
-                  call clip_values(aer_tau_bnd_sw,  0._r8, huge(aer_tau_bnd_sw), trim(subname) // ' aer_tau_bnd_sw', warn=.true.)
-                  call clip_values(aer_ssa_bnd_sw,  0._r8,                1._r8, trim(subname) // ' aer_ssa_bnd_sw', warn=.true.)
-                  call clip_values(aer_asm_bnd_sw, -1._r8,                1._r8, trim(subname) // ' aer_asm_bnd_sw', warn=.true.)
+                  call clip_values(cld_tau_gpt_sw,  0._r8, huge(cld_tau_gpt_sw), trim(subname) // ' cld_tau_gpt_sw', warn=.true., tolerance=1e-10_r8)
+                  call clip_values(cld_ssa_gpt_sw,  0._r8,                1._r8, trim(subname) // ' cld_ssa_gpt_sw', warn=.true., tolerance=1e-10_r8)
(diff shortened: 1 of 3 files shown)


components/cam/src/physics/rrtmgp/radiation.F90
```diff
@@ -1334,6 +1348,14 @@ subroutine radiation_tend(state,   ptend,    pbuf,          cam_out, cam_in,  &
                   aer_asm_bnd_sw = 0
                end if
 
+               ! Check (and possibly clip) values before passing to RRTMGP driver
+               call handle_error(clip_values(cld_tau_gpt_sw,  0._r8, huge(cld_tau_gpt_sw), trim(subname) // ' cld_tau_gpt_sw', tolerance=1e-10_r8))
```
wlin7, line 1352: Any evidence that cld_tau_gpt_sw and aer_tau_gpt_sw could get an infinite number? If certainly unlikely, the clipping calls for them can be safely removed.

golaz:
@brhillman and @wlin7 : is this now ready to be merged on master? I tried running FAMIP and F2010 CMIP6-based compsets with master and they both failed with ssa errors. This PR will hopefully fix these errors. Thanks.

wlin7:
Yes,  Chris. Just merged to next.

components/cam/src/physics/rrtmgp/radiation.F90
brhillman: Not physically, this would be an indication that things went haywire somewhere. I wanted to keep these here for consistency, and to make sure we were handling the sanity checking ourselves before RRTMGP had to deal with it. This will catch cases where the tau values go to nan/inf before RRTMGP does.

wlin7:
Merged to master.

amametjanov:
This converted `clip_values` subroutines to functions and a couple of calls were missed leading to build errors with Intel and PGI for `ERS_Ln9_P96x1.ne4pg2_ne4pg2.F-MMF1-TEST.cam-crmout`: e.g. with intel
```
/lcrc/group/acme/azamat/repos/E3SM-tests/components/cam/src/physics/crm/rrtmgp/radiation.F90(2498): error #6552: The CALL statement is invoking a function subprogram as a subroutine.   [RADIATION_UTILS^CLIP_VALUES_2D]
      call clip_values(albedo_direct, 0._r8, 1._r8, varname='albedo_direct')
-----------^
```

This patch fixes the build issue:
```
diff --git a/components/cam/src/physics/crm/rrtmgp/radiation.F90 b/components/cam/src/physics/crm/rrtmgp/radiation.F90
index 566c8c0..01390ba 100644
--- a/components/cam/src/physics/crm/rrtmgp/radiation.F90
+++ b/components/cam/src/physics/crm/rrtmgp/radiation.F90
@@ -2495,8 +2495,8 @@ contains
       ! NOTE: this does actually issue warnings for albedos larger than 1, but this
       ! was never checked for RRTMG, so albedos will probably be slight different
       ! than the implementation in RRTMG!
-      call clip_values(albedo_direct, 0._r8, 1._r8, varname='albedo_direct')
-      call clip_values(albedo_diffuse, 0._r8, 1._r8, varname='albedo_diffuse')
+      call handle_error(clip_values(albedo_direct, 0._r8, 1._r8, varname='albedo_direct'), fatal=.false.)
+      call handle_error(clip_values(albedo_diffuse, 0._r8, 1._r8, varname='albedo_diffuse'), fatal=.false.)

    end subroutine set_albedo

```
Can you include it in one of the PRs: e.g. E3SM-Project/E3SM#3639 ? Thanks.

brhillman:
@amametjanov good catch yes I will.

PR #3631, opened 2020-06-10
mpas-source: Move daysSinceStartOfSim calculation to master when threaded

mark-petersen:
See https://github.com/MPAS-Dev/MPAS-Model/pull/601

This PR moves the calculation of daysSinceStartOfSim, which is done in three places, to inside a loop with "omp master" pragmas around it. Previously, one was inside an "omp single" pragma, but that did not guarantee that the master thread would get the correct result. While this does not impact the results at all, it was causing random problems in testing using threaded configurations where cprnc output from globalStats files would find an error, just in the daysSinceStartOfSim field. With thanks to @amametjanov

[BFB]


jonbob:
merged to master

PR #3632, opened 2020-06-11
Adding E3SM support for 3DGM namelist options to maint-1.2

darincomeau:
This adds changes from #3057 for E3SM support of 3DGM namelist options to the maint-1.2 branch. As part of this, 3DGM is turned on by default (as it was in #3057), so it is not BFB. The changes were copied directly from the E3SM (non mpas-source) file changes in #3057 (hash 
9fc9d6f).

[non-BFB]
[NML]

### DIFF
--- a/components/mpas-ocean/bld/build-namelist
+++ b/components/mpas-ocean/bld/build-namelist
@@ -600,6 +600,10 @@ add_default($nl, 'config_Leith_visc2_max');
 ###################################################
 
 add_default($nl, 'config_use_standardGM');
+add_default($nl, 'config_gm_lat_variable_c2');
+add_default($nl, 'config_gm_kappa_lat_depth_variable');
+add_default($nl, 'config_gm_min_stratification_ratio');
+add_default($nl, 'config_gm_min_phase_speed');
 add_default($nl, 'config_use_Redi_surface_layer_tapering');
 add_default($nl, 'config_Redi_surface_layer_tapering_extent');
 add_default($nl, 'config_use_Redi_bottom_layer_tapering');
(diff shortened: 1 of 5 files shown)


darincomeau:
We realized this was missing when @stephenprice tried to run with 3DGM turned on, and errors.

darincomeau:
For testing I just ran a A_WCYCL1850-DIB-ISMF_CMIP6.ne30_oECv3wLI smoke test, and @stephenprice is using this commit on his new A_WCYCL1850-DIB-ISMF_CMIP6.ne30_ECwISC30to60E1r2.cori-knl.maint1p2-3DGM run, but I think this was just an oversight omission from #3513 .

rljacob:
Please replace the PR numbers with hashes in the description above.

darincomeau:
@rljacob I've added the hashes to the description.

rljacob:
Is the MPAS code to implement 3DGM in maint-1.2?  I thought that was added in what we called 1.3?

darincomeau:
The MPAS source code to implement 3DGM is in maint-1.2 (via 8f8fefc), but the E3SM files that generate the new namelist options did not make it to maint-1.2, so we can't actually use 3DGM on maint-1.2 without this.

darincomeau:
But I think I see what you're getting at, re: the simulations being referenced as being 1.2 or 1.3. Perhaps it's better to continue offline.

darincomeau:
> Is the MPAS code to implement 3DGM in maint-1.2? I thought that was added in what we called 1.3?

1.3 was when 3DGM went into master, not maint-1.2

jonbob:
merged to maint-1.2

PR #3633, opened 2020-06-11
For Cori, update module version of the craype module and force static linking

ndkeen:
For Cori, update module version of the craype module and force static linking.

With this newer version of craype (2.6.2), the new default is to use dynamic linking.
E3SM still works with dynamic linking, but the performance was degraded.
Forcing static linking seems to resolve this (by setting an environment variable
CRAYPE_LINK_TYPE to static).

[bfb]

### DIFF
--- a/cime/config/e3sm/machines/config_machines.xml
+++ b/cime/config/e3sm/machines/config_machines.xml
@@ -154,7 +154,7 @@
       </modules>
 
       <modules>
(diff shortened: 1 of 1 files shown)


ndkeen:
Previously I encountered some runs that were slower with this change, but most recent tests do not show the timing is outside of normal variations.

ndkeen:
merged to next

ndkeen:
merged to master

PR #3634, opened 2020-06-11
Update mpas-source: three ocean bug fixes

mark-petersen:
fixed bad array sizes in local ocean arrays
  https://github.com/MPAS-Dev/MPAS-Model/pull/598
Remove RediKappa from gm package
  https://github.com/MPAS-Dev/MPAS-Model/pull/602
Fixes bugs in tracer_advection_std
  https://github.com/MPAS-Dev/MPAS-Model/pull/587

[BFB]

PR #3636, opened 2020-06-12
Update mpas-source: Add mixed layer AM, add del2 ref. cell width

mark-petersen:
This PR includes the following updates:
1. Adds Mixed layer heat and salinity budgets https://github.com/MPAS-Dev/MPAS-Model/pull/568
2. Add reference cell width to del2 and del4 scaling https://github.com/MPAS-Dev/MPAS-Model/pull/599
3. Moves constrain_Haney_number namelist (init mode only) to init mode Registry file.

[BFB]
[NML]


mark-petersen:
The new AM will have a new namelist and streams section. @jonbob could you copy the format of the last one for ocean heat content in #3613.

From @vanroekel: I would like these added to the monthly average AM AND the monthly min / max AMs as well. 
```
 <var_array name="activeTracerForcingMLTend"/>
 <var_array name="activeTracerHorMixMLTend"/>
 <var_array name="activeTracerNonLocalMLTend"/>
 <var_array name="activeTracerVertAdvectionMLTend"/>
 <var_array name="activeTracerHorAdvectionMLTend"/>
 <var_array name="activeTracerVertMixMLTend"/>
 <var_array name="activeTracersML"/>
 <var_array name="bruntVaisalaFreqML"/>
 <var_array name="activeTracersTendML"/>
```
These should be the current E3SM defaults for these flags:
```
config_hmix_use_ref_cell_width = .false.
config_hmix_ref_cell_width = 30.0e3
```

mark-petersen:
This can go after #3634, and currently includes those diffs.

jonbob:
@mark-petersen, @vanroekel - do we want the new AM on by default? Using the defaults from Registry, we get this:
````
<!-- AM_mixedLayerHeatBudget -->
<config_AM_mixedLayerHeatBudget_enable>.false.</config_AM_mixedLayerHeatBudget_enable>
<config_AM_mixedLayerHeatBudget_compute_interval>'output_interval'</config_AM_mixedLayerHeatBudget_compute_interval>
<config_AM_mixedLayerHeatBudget_output_stream>'mixedLayerHeatBudgetOutput'</config_AM_mixedLayerHeatBudget_output_stream>
<config_AM_mixedLayerHeatBudget_compute_on_startup>.true.</config_AM_mixedLayerHeatBudget_compute_on_startup>
<config_AM_mixedLayerHeatBudget_write_on_startup>.true.</config_AM_mixedLayerHeatBudget_write_on_startup>
````

**Comment by vanroekel** [comment]
I would say yes, we should flip falses to true and true to false.  I don't think we need compute/write on startup as the default.

**Comment by jonbob** [comment]
Thanks @vanroekel - I'll make sure that gets changed for E3SM

**Comment by jonbob** [comment]
@vanroekel - OK, all my changes to the scripts are pushed back, if you can take a look at the results streams file and make sure the defaults are correct. I ran a one-month test and can verify that the mpaso.hist.am.mixedLayerHeatBudget is created

File: components/mpas-ocean/cime_config/buildnml [hunk 1]
```diff
@@ -751,6 +751,29 @@ def buildnml(case, caseroot, compname):
...
         lines.append('</stream>')
         lines.append('')
+        lines.append('<stream name="mixedLayerHeatBudgetOutput"')
+        lines.append('        type="output"')
+        lines.append('        precision="single"')
+        lines.append('        io_type="{}"'.format(ocn_pio_typename))
+        lines.append('        filename_template="mpaso.hist.am.mixedLayerHeatBudget.$Y-$M-$D.nc"')
+        lines.append('        filename_interval="00-01-00_00:00:00"')
+        lines.append('        reference_time="01-01-01_00:00:00"')
+        lines.append('        output_interval="00-01-00_00:00:00"')
```
vanroekel, line 761: @jonbob since these variables are in the monthly avg/max/min should we set output_interval='none' here?

jonbob: sure -- or else have that AM off by default?

vanroekel: I think the default now is to have all the output streams off for their individual files but have them in the monthly averages.  I'd definitely like to keep this on by default.

jonbob: On but with output_interval='none'?

vanroekel: yes I think that's right.

jonbob: Do you want the heat content one treated the same way? While I'm at it?

jonbob: OK, I changed the output_interval and pushed a new commit

jonbob:
merged to next

mark-petersen:
@jonbob and @vanroekel thanks for the help. The last commit includes the updated `maxMeshDensity` variable. I don't know how that passed the tests for that PR - I'll have to look into that.

jonbob:
I think we mostly test with oQU240 and other low-res meshes -- the E3SM testing only threw an error with debugging on. But it also caused DIFFs for oEC60to30v3 tests, though nothing at less resolution because of the different default settings for those meshes

mark-petersen:
I threw in a few tests:
```
PASS PEM_Ln9.T62_oQU240.GMPAS-IAF.cori-knl_gnu
PASS PET_Ln9.T62_oQU240.GMPAS-IAF.cori-knl_intel
PASS PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-knl_gnu
PASS SMS_Ln9.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_intel
```

jonbob:
reverted the revert and merged to next with fixes to the problem exposed by testing

jonbob:
merged to master and executed NML DIFF bless requests issued

PR #3637, opened 2020-06-15
Update LANL IC machine files to fix small issues

jonbob:
This PR fixes two small issues on the LANL IC machines badger and grizzly. Specifically, it:
* removes -lmpi_cxx from c++ libraries for the gnu compiler on both machines
* removes -lmpi_cxx from c++ libraries for the intel compiler on badger
* reorders the options for compiler on badger, so intel is default
* reorders the options for mpi on both machines, so openmpi is default
No tests are run on either of these platforms, so there should be no impact on E3SM tested configurations.

[BFB]


stephenprice:
The machine file changes look sensible to me. FYI, I think that we found intel + mvapich was similar to intel + impi.

stephenprice:
I can't speak to the compiler file changes since I didn't hit any problems there. But ... did you mean to remove the "-lmpi_cxx" line for "gnu" under Grizzly whereas you removed the same  line under "intel" for Badger?

jonbob:
Sorry if that wasn't clear -- I removed "-lmpi_cxx" from the gnu compiler on both machines. I tested it on grizzly and with that library specified, a smoke test fails to build because it does not find that library

stephenprice:
Thanks. If you are confident those subtractions are correct, I'm going to defer to @maltrud on signing off on those. Let us know if you are looking for additional testing here.

jonbob:
Ah, missed that one -- thanks @maltrud

jonbob:
Sorry @stephenprice -- I think you saw that as well and I didn't understand what you were referring to

jonbob:
merged to next

jonbob:
merged to master

PR #3639, opened 2020-06-15
Fix setting of swband/lwband for history files

brhillman:
Fix setting of swband and lwband coordinate variables for history files.
Since add_hist_coord simply sets up pointers to the arrays that are
passed in, the data needs to persist. Thus, we make these arrays
module-scope variables in radiation.F90. Also a little bit of clean-up
in radiation.F90 and radconstants.F90. In particular, we now use
nswbands and nlwbands from radconstants.F90, rather than defining new
versions of these in radiation.F90. Since these values are set at
compile time and should never change, we should be using these in
radiation.F90 anyways, and it does not make sense to make new versions
of these that could inadvertently get changed from what we expect in
radconstants.F90. So instead of setting new values for these based on
the absorption coefficient data, we simply verify that the number of
bands in the input files matches what we expect in the code (14 for,
shortwave, 16 for longwave). These fixes are BFB in regards to the
simulation, but EAM history files will be non-BFB due to the fix in the
swband and lwband coordinate variables. This will only affect
simulations that are long enough to generate history files, i.e., either
long (month-plus) runs, or high temporal output frequency runs (like the
crmout test).

[non-BFB]

### DIFF
--- a/components/cam/src/physics/rrtmgp/radconstants.F90
+++ b/components/cam/src/physics/rrtmgp/radconstants.F90
@@ -24,15 +24,28 @@ save
 
 ! SHORTWAVE DATA
 
-! number of shorwave spectral intervals
+! Number of spectral intervals
 integer, parameter, public :: nswbands = 14
+integer, parameter, public :: nlwbands = 16
 
 ! Wavenumbers of band boundaries
 !
 ! Note: Currently rad_solar_var extends the lowest band down to
 ! 100 cm^-1 if it is too high to cover the far-IR. Any changes meant
 ! to affect IR solar variability should take note of this.
-real(r8), dimension(nswbands) :: wavenum_sw_lower, wavenum_sw_upper
+!real(r8), dimension(nswbands) :: wavenum_sw_lower, wavenum_sw_upper
+real(r8) :: wavenum_sw_lower(nswbands) = & ! in cm^-1
+  (/ 820._r8, 2680._r8, 3250._r8, 4000._r8, 4650._r8, 5150._r8, 6150._r8, &
+    7700._r8, 8050._r8,12850._r8,16000._r8,22650._r8,29000._r8,38000._r8/)
+real(r8) :: wavenum_sw_upper(nswbands) = & ! in cm^-1
+  (/2680._r8, 3250._r8, 4000._r8, 4650._r8, 5150._r8, 6150._r8, 7700._r8, &
+    8050._r8,12850._r8,16000._r8,22650._r8,29000._r8,38000._r8,50000._r8/)
+real(r8) :: wavenum_lw_lower(nlwbands) = &! Longwave spectral band limits (cm-1)
+    (/   10._r8,  250._r8, 500._r8,   630._r8,  700._r8,  820._r8,  980._r8, 1080._r8, &
+       1180._r8, 1390._r8, 1480._r8, 1800._r8, 2080._r8, 2250._r8, 2390._r8, 2680._r8 /)
+real(r8) :: wavenum_lw_upper(nlwbands) = &! Longwave spectral band limits (cm-1)
+    (/  250._r8,  500._r8,  630._r8,  700._r8,  820._r8,  980._r8, 1080._r8, 1180._r8, &
+       1390._r8, 1480._r8, 1800._r8, 2080._r8, 2250._r8, 2390._r8, 2680._r8, 3250._r8 /)
 
 ! Solar irradiance at 1 A.U. in W/m^2 assumed by radiation code
 ! Rescaled so that sum is precisely 1368.22 and fractional amounts sum to 1.0
@@ -70,12 +83,6 @@ integer, parameter, public :: nrh = 1000
 integer, parameter, public :: idx_lw_diag = 7 ! index to (H20 window) LW band
 
 integer, parameter, public :: rrtmg_lw_cloudsim_band = 6  ! rrtmg band for 10.5 micron
-
-! number of lw bands
-integer, parameter, public :: nlwbands = 16
-
-real(r8), dimension(nlwbands) :: wavenum_lw_lower, wavenum_lw_upper
-
 
 !These can go away when old camrt disappears
 ! Index of volc. abs., H2O non-window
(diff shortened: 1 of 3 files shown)


components/cam/src/physics/rrtmgp/radconstants.F90
```diff
@@ -256,6 +265,42 @@ end subroutine get_sw_spectral_boundaries
...
+   real(r8), dimension(nswbands) :: lower_bounds, upper_bounds
+   integer :: i
+
+   ! Get band limits
+   call get_sw_spectral_boundaries(lower_bounds, upper_bounds, units)
+
+   ! Compute band midpoints
+   band_midpoints = 0._r8
+   do i = 1,nswbands
+      band_midpoints(i) = (lower_bounds(i) + upper_bounds(i)) / 2._r8
```
whannah1, line 280: isn't is better to avoid using "._r8" here and just use "2"?

brhillman: I've recently been advised to use the explicit kind on the floating point literals, even though I think they end up being identical in these cases.

brhillman:
@whannah1 @wlin7  I am going to rebase this branch and add the patch that @amametjanov mentioned to fix the bug that #3630 introduced.

wlin7:
All is good. Also verify the spectral band boundaries specified in radconstants.F90 against those in the rrtmgp coefficients files. Starting to merge.

wlin7:
Merged to next.

wlin7:
@amametjanov , after this PR in next, the MMF test now builds ok but still fails to run. Do we know the cause of runtime failure? If the new issue is not directly caused by this PR, I think we can merge it to master first.

amametjanov:
atm.log for `ERS_Ln9_P96x1.ne4pg2_ne4pg2.F-MMF1-TEST.sandiatoss3_intel.cam-crmout` shows:
```
 ERROR:
  WARNING: radiation_tend cld_asm_gpt_sw:         5345  values above threshold ;
  max =   1.144905381791343E+243
```
and the backtrace from e3sm.log:
```
e3sm.exe           000000000202BF3D  Unknown               Unknown  Unknown
e3sm.exe           0000000001CF4642  shr_abort_mod_mp_         114  shr_abort_mod.F90
e3sm.exe           00000000007BDFFC  radiation_utils_m         334  radiation_utils.F90
e3sm.exe           00000000006519FE  radiation_mp_radi        1568  radiation.F90
e3sm.exe           00000000005FA979  physpkg_mp_tphysb        3067  physpkg.F90
e3sm.exe           00000000005EDD04  physpkg_mp_phys_r        1087  physpkg.F90
e3sm.exe           00000000004FF6E1  cam_comp_mp_cam_r         250  cam_comp.F90
e3sm.exe           00000000004F59F2  atm_comp_mct_mp_a         394  atm_comp_mct.F90
e3sm.exe           0000000000439838  component_mod_mp_         257  component_mod.F90
e3sm.exe           000000000042DEDA  cime_comp_mod_mp_        2202  cime_comp_mod.F90
e3sm.exe           00000000004368E9  MAIN__                    122  cime_driver.F90
e3sm.exe           0000000000417DAE  Unknown               Unknown  Unknown
```
Looking at `components/cam/src/physics/crm/rrtmgp/radiation.F90`:
```
1565                         ! Check (and possibly clip) values before passing to RRTMGP driver
1566                         call handle_error(clip_values(cld_tau_gpt_sw,  0._r8, huge(cld_tau_gpt_sw), trim(subname) // ' cld_tau_gpt_     sw', tolerance=1e-10_r8))
1567                         call handle_error(clip_values(cld_ssa_gpt_sw,  0._r8,                1._r8, trim(subname) // ' cld_ssa_gpt_     sw', tolerance=1e-10_r8))
1568                         call handle_error(clip_values(cld_asm_gpt_sw, -1._r8,                1._r8, trim(subname) // ' cld_asm_gpt_     sw', tolerance=1e-10_r8))
1569                         call handle_error(clip_values(aer_tau_bnd_sw,  0._r8, huge(aer_tau_bnd_sw), trim(subname) // ' aer_tau_bnd_     sw', tolerance=1e-10_r8))
1570                         call handle_error(clip_values(aer_ssa_bnd_sw,  0._r8,                1._r8, trim(subname) // ' aer_ssa_bnd_     sw', tolerance=1e-10_r8))
1571                         call handle_error(clip_values(aer_asm_bnd_sw, -1._r8,                1._r8, trim(subname) // ' aer_asm_bnd_     sw', tolerance=1e-10_r8))
```
Same error on anvil.

Similar error on compy+pgi:
```
87:  ERROR: 
87:   WARNING: radiation_tend: cld_tau_gpt_lw:          4032  values above threshold ; max =     577473.5274481252
87: Abort(1001) on node 87 (rank 87 in comm 0): application called MPI_Abort(MPI_COMM_WORLD, 1001) - process 87
```
I think this should be fixed before merging to master.

brhillman:
Thanks @amametjanov I pushed a commit earlier today to fix this.

amametjanov:
PGI aborts on `cld_tau_gpt_lw`.

brhillman:
@amametjanov can you be more specific? `cld_tau_gpt_lw` is a variable, so where exactly is PGI aborting? I'll try building on Summit (the only machine I have access to with PGI right now) and see if I can reproduce.

brhillman:
Oh I think I see...there's another uninitialized variable that got missed here. For some reason this didn't bail out on my test on sandiatoss3...I'll push yet another commit...

amametjanov:
Thanks.
@wlin7 please re-merge to next for another test round. `ERS_Ln9_P96x1.ne4pg2_ne4pg2.F-MMF1-TEST.compy_pgi.cam-crmout` will still fail, but that's in the restart run due to unrelated issue.

brhillman:
Yeah, I'm been confused about what's going on with PGI for a while. It's failing at line 2635 in `cam_history.F90`, which should not even be triggered. I get what the error message is saying (the string that's being attempted to be written to the `errormsg` variable exceeds the size of `errormsg`), but this line should not even be executed (the enclosing logical should evaluate false). Checking the test on sandiatoss3, this line is never written to the log files, so the line is indeed *not* executed there, but PGI builds seem to encounter it.

wlin7:
Re-merged to next after the fixes.

wlin7:
Merged to master. Request to bless diff for sandiatoss3 next MMF1 test. MMF1 test on compy continues to fail (due to other cause). MMF1 tests on anvil and theta have not returned.

brhillman:
Awesome, thanks @wlin7 !

PR #3640, opened 2020-06-16
Modified to write large netcdf file.

pnlfang:


bishtgautam:
@pnlfang I was expecting only a few lines to be modified, but this PR shows the whole file has been modified. Also, could you add some description regarding this PR in the first comment box?

mt5555:
looks like instead of editing an existing file, a new file was added to the parent e3sm directory.

bishtgautam:
Thanks, @mt5555.

@pnlfang, I'm closing this PR because a new file has been added instead of modifying the existing file. Also, please follow the E3SM branch naming convention.

PR #3645, opened 2020-06-16
For cori-knl, update the version of default intel compiler to be version 19

ndkeen:
For cori-knl, update the version of default intel compiler to be version 19.
The specific version is intel/19.0.3.199 which is the current machine default.
We also remove the "intel19" compiler option from cori-knl to simplify.

Also adjust the C++ flags to use -fp-model consistent for cori-knl and cori-haswell just to be the same as the current fortran flags, however, we intend on revisiting the compiler flag settings as a whole.

This change is NOT bit-for-bit, but all the tests are passing otherwise.


ndkeen:
merged to next

ndkeen:
merged to master. will also start bless ticket

PR #3649, opened 2020-06-17
Update mpas-source: fix E3SM Makefile

mark-petersen:
Add -DUSE_PIO2 to override flags in E3SM Makefile. 

[BFB]
fixes E3SM-Project/E3SM#3646

PR #3651, opened 2020-06-17
ocean: alter 3D GM

mark-petersen:
See https://github.com/MPAS-Dev/MPAS-Model/pull/537

Adds horizontal variability of GM bolus kappa value of Visbeck et al 1997. The Danabasoglu and Marshall 2007 modifications still function independently of this option. The scheme contains options for min and max bolus value and more work is needed to properly pair this scheme with the ramp function for variable resolution.

[non-BFB]
[NML]

### DIFF
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -114,13 +114,22 @@
 <config_Leith_visc2_max>2.5e3</config_Leith_visc2_max>
 
 <!-- Redi_isopycnal_mixing -->
-<config_use_Redi>.false.</config_use_Redi>
-<config_Redi_kappa>1800.0</config_Redi_kappa>
-<config_Redi_kappa ocn_forcing="datm_forced_restoring">600.0</config_Redi_kappa>
-<config_Redi_maximum_slope>0.3</config_Redi_maximum_slope>
+<config_use_Redi>.true.</config_use_Redi>
+<config_use_Redi ocn_grid="oRRS30to10">.false.</config_use_Redi>
+<config_use_Redi ocn_grid="oRRS30to10v3">.false.</config_use_Redi>
+<config_use_Redi ocn_grid="oRRS30to10wLI">.false.</config_use_Redi>
+<config_use_Redi ocn_grid="oRRS30to10v3wLI">.false.</config_use_Redi>
+<config_use_Redi ocn_grid="oRRS18to6">.false.</config_use_Redi>
+<config_use_Redi ocn_grid="oRRS18to6v3">.false.</config_use_Redi>
+<config_use_Redi ocn_grid="oRRS15to5">.false.</config_use_Redi>
+<config_Redi_kappa>400.0</config_Redi_kappa>
+<config_Redi_kappa ocn_forcing="datm_forced_restoring">400.0</config_Redi_kappa>
+<config_Redi_set_RediKappa_to_GMKappa>.false.</config_Redi_set_RediKappa_to_GMKappa>
+<config_Redi_maximum_slope>0.01</config_Redi_maximum_slope>
 <config_Redi_closure>'constant'</config_Redi_closure>
 <config_Redi_use_slope_taper>.true.</config_Redi_use_slope_taper>
 <config_Redi_use_surface_taper>.true.</config_Redi_use_surface_taper>
+<config_Redi_use_N2_based_taper>.true.</config_Redi_use_N2_based_taper>
 <config_Redi_N2_limit_term1>.true.</config_Redi_N2_limit_term1>
 
 <!-- GM_eddy_parameterization -->
@@ -134,7 +143,13 @@
 <config_use_GM ocn_grid="oRRS15to5">.false.</config_use_GM>
 <config_GM_kappa>1800.0</config_GM_kappa>
 <config_GM_kappa ocn_forcing="datm_forced_restoring" ocn_grid="oEC60to30v3wLI">600.0</config_GM_kappa>
-<config_GM_closure>'N2_dependent'</config_GM_closure>
+<config_GM_closure>'visbeck'</config_GM_closure>
+<config_GM_baroclinic_mode>3.0</config_GM_baroclinic_mode>
+<config_GM_visbeck_alpha>0.005</config_GM_visbeck_alpha>
+<config_GM_visbeck_min_depth>50.0</config_GM_visbeck_min_depth>
+<config_GM_visbeck_max_depth>1000.0</config_GM_visbeck_max_depth>
+<config_GM_visbeck_min_kappa>300.0</config_GM_visbeck_min_kappa>
+<config_GM_visbeck_max_kappa>1200.0</config_GM_visbeck_max_kappa>
 <config_GM_min_stratification_ratio>0.1</config_GM_min_stratification_ratio>
 <config_GM_constant_gravWaveSpeed>0.3</config_GM_constant_gravWaveSpeed>
 
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -114,7 +114,7 @@
 <config_Leith_visc2_max>2.5e3</config_Leith_visc2_max>
 
 <!-- Redi_isopycnal_mixing -->
-<config_use_Redi>.true.</config_use_Redi>
+<config_use_Redi>.false.</config_use_Redi>
 <config_use_Redi ocn_grid="oRRS30to10">.false.</config_use_Redi>
 <config_use_Redi ocn_grid="oRRS30to10v3">.false.</config_use_Redi>
 <config_use_Redi ocn_grid="oRRS30to10wLI">.false.</config_use_Redi>
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -145,7 +145,7 @@
 <config_GM_kappa ocn_forcing="datm_forced_restoring" ocn_grid="oEC60to30v3wLI">600.0</config_GM_kappa>
 <config_GM_closure>'visbeck'</config_GM_closure>
 <config_GM_baroclinic_mode>3.0</config_GM_baroclinic_mode>
-<config_GM_visbeck_alpha>0.005</config_GM_visbeck_alpha>
+<config_GM_visbeck_alpha>0.13</config_GM_visbeck_alpha>
 <config_GM_visbeck_min_depth>50.0</config_GM_visbeck_min_depth>
 <config_GM_visbeck_max_depth>1000.0</config_GM_visbeck_max_depth>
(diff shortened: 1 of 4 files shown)


vanroekel:
@jonbob for this PR I'd like the following defaults for GM/Redi parameters based on e3sm testing
```
config_use_Redi = .true. -- For low res and RRM only
config_Redi_kappa = 400.
config_Redi_set_RediKappa_to_GMKappa = .false.
config_Redi_maximum_slope = 0.01
config_GM_closure = 'visbeck'
config_GM_baroclinic_mode = 3.0
config_GM_visbeck_alpha = 0.005
config_GM_visbeck_max_kappa = 1200.
```
These are all changes from the MPAS defaults

mark-petersen:
This goes after #3636 and #3649. Those commits are showing here right now.

@vanroekel asked for these defaults:
```
config_use_Redi = .true. -- For low res and RRM only
config_Redi_kappa = 400.
config_Redi_set_RediKappa_to_GMKappa = .false.
config_Redi_maximum_slope = 0.01
config_GM_closure = 'visbeck'
config_GM_baroclinic_mode = 3.0
config_GM_visbeck_alpha = 0.005
config_GM_visbeck_max_kappa = 1200.
```

jonbob:
@mark-petersen - what's going on with this PR? We're getting all kinds of weird messages

mark-petersen:
Rebased on master. This only has the single commit to update mpas-source. That includes MPAS-Dev/MPAS-Model#537 but also the makefile change in #3649. @jonbob if you could make the namelist changes above, that would be great.

jonbob:
@vanroekel - please let me know if these new defaults look correct to you. I'm running a quick debug check on a test merge

jonbob:
Passed a SMS_D.ne4_oQU240.A_WCYCL1850.anvil_intel sanity test

mark-petersen:
Tested with current head:
```
PASS PEM_Ln9.T62_oQU240.GMPAS-IAF.cori-knl_intel
PASS PET_Ln9.T62_oQU240.GMPAS-IAF.cori-knl_gnu 
PASS PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-knl_gnu
```

components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
```diff
@@ -134,7 +143,13 @@
...
 <config_GM_kappa>1800.0</config_GM_kappa>
 <config_GM_kappa ocn_forcing="datm_forced_restoring" ocn_grid="oEC60to30v3wLI">600.0</config_GM_kappa>
-<config_GM_closure>'N2_dependent'</config_GM_closure>
+<config_GM_closure>'visbeck'</config_GM_closure>
+<config_GM_baroclinic_mode>3.0</config_GM_baroclinic_mode>
+<config_GM_visbeck_alpha>0.005</config_GM_visbeck_alpha>
+<config_GM_visbeck_min_depth>50.0</config_GM_visbeck_min_depth>
+<config_GM_visbeck_max_depth>1000.0</config_GM_visbeck_max_depth>
+<config_GM_visbeck_min_kappa>300.0</config_GM_visbeck_min_kappa>
+<config_GM_visbeck_max_kappa>1200.0</config_GM_visbeck_max_kappa>
```
mark-petersen, line 152: @vanroekel you specified `config_GM_visbeck_max_kappa = 1200.` on your post above, but please double check these other three for depth and kappa, which just came from the defaults.

jonbob:
merged to next

jonbob:
@vanroekel, @mark-petersen - we are failing some automated tests with this PR, specifically ERP_Ld3.ne30_oECv3_ICG.A_WCYCL1850S.anvil_intel (an exact restart test) fails on next but passes if I revert the merge of this branch. I'll check with a lower resolution test as well, to see if that holds any clues.

vanroekel:
@jonbob thanks.  Can you also run a restart test with config_use_redi=.false.?  This will help isolate where the issue is.

jonbob:
@vanroekel - yes, I'll do that right away and post results here

jonbob:
@vanroekel - here's an update:
* ERS.ne4_oQU240.A_WCYCL1850.anvil_intel passes both before and after the latest PR
* with config_use_redi=.false., ERP_Ld3.ne30_oECv3_ICG.A_WCYCL1850S.anvil_intel still fails

vanroekel:
@jonbob can you explain the differences between the two tests?  I'm not familiar with ERP vs ERS.

jonbob:
@vanroekel - I had to look, because the names have changed over the past few years. This is what I found:
````
ERS    exact restart from startup (default 6 days + 5 days)
       do an 11 day initial test - write a restart at day 6     (suffix: base)
       if $IOP_ON is set then suffix is base_iop
       do a  5  day restart test starting from restart at day 6 (suffix: rest)
       if $IOP_ON is set then suffix is rest_iop
       compare component history files ".base" and ".rest" at day 11

ERP    pes counts hybrid (open-MP/MPI) restart bfb test from startup, default 6 days + 5 days (previousy PER)
       initial pes set up out of the box
       do an 11 day initial test - write a restart at day 6     (suffix base)
       half the number of tasks and threads for each component
       do a  5  day restart test starting from restart at day 6 (suffix rest)
       this is just like an ERS test but the pe-counts/threading count are modified on restart````

**Comment by jonbob** [comment]
so I'm trying a low-res ERP test right now

**Comment by vanroekel** [comment]
Thanks!  So I'm guessing this is an openmp issue.  One last question, is the test using 'visbeck' GM?  The new default?  If so, could you also do a quick run with that set to 'constant'?  It will further isolate the issue.

**Comment by vanroekel** [comment]
@jonbob I took a quick look at the new GM code and think I may have found something.  If you look here

https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/src/core_ocean/shared/mpas_ocn_gm.F#L492-L494

I think the OMP private list also needs to include 'k'

Not sure this will solve the problem, but I think it is a bug for threading.  Let me know if I'm off base here, as I'm not super familiar with omp.

**Comment by jonbob** [comment]
Well, if it's jumping out at you it's worth a shot. Let me see if the low-res one fails as well -- we'll have a much easier time testing with that one if it does

**Comment by jonbob** [comment]
The low-res ERP test did not fail, so I have a ERP_Ld3.ne30_oECv3_ICG.A_WCYCL1850S.anvil_intel test in the queue with   config_gm_closure = 'constant'. And then I'll try with the omp fix...

**Comment by jonbob** [comment]
@vanroekel - still failing with the omp fix you described and
````
 config_use_redi=.false.
 config_gm_closure = 'constant'
````

**Comment by vanroekel** [comment]
thanks @jonbob can you point me at your run directory so I can check the namelist for the ocean?

**Comment by jonbob** [comment]
@vanroekel - /lcrc/group/acme/ac.jwolfe/acme_scratch/anvil/ERP_Ld3.ne30_oECv3_ICG.A_WCYCL1850S.anvil_intel.20200630_094454_up4rkw

**Comment by mark-petersen** [comment]
This loop doesn't have openMP pragmas. 
https://github.com/MPAS-Dev/MPAS-Model/pull/537/files#diff-f01cba32c377a8c9529bd4f2b608e0fbR183

It needs
```
!$omp do schedule(runtime) private(i)
```
But it is in the redi section, and you still get the error with redi off, so this must not be the cause.

**Comment by jonbob** [comment]
@mark-petersen - iEdge needs to be private as well in that loop?

**Comment by mark-petersen** [comment]
I'm able to replicate this on cori:
```
./create_test -q debug --walltime 00:30:00 --force-procs 1024 ERP_Ld3.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_intel
PASS ERP_Ld3_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_intel RUN time=1476
FAIL ERP_Ld3_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_intel COMPARE_base_rest
```
If it is a threading error, PET should catch it too.

**Comment by mark-petersen** [comment]
> @mark-petersen - iEdge needs to be private as well in that loop?
Yes. Thanks.

**Comment by amametjanov** [comment]
Just noting that PEM test is failing comparison too: both ERP and PEM tests halve number of MPI tasks in the second (compare) run. The issue is likely due to different number of MPI tasks, not threading (PET integration tests pass).
```
$ head  cime/scripts/lib/CIME/SystemTests/pem.py
"""
Implementation of the CIME PEM test: Tests bfb with different MPI
processor counts

This is just like running a smoke test twice - but the pe-counts
are modified the second time.
(1) Run with pes set up out of the box (suffix base)
(2) Run with half the number of tasks (suffix modpes)
"""
```

**Comment by jonbob** [comment]
Thanks @amametjanov - that's a great point

**Comment by jonbob** [comment]
@mark-petersen - any chance this is a halo issue? now that rediKappa is over edges, does it need to get treated differently?

**Comment by mark-petersen** [comment]
@amametjanov thanks for pointing out that it is not threading. I can confirm that. PET passes but PEM does not:
```
PASS PET_Ld3_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_intel COMPARE_base_single_thread

PASS PEM_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_intel RUN time=495
FAIL PEM_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_intel COMPARE_base_modpes
```
This does not show up at lower resolution:
```
PASS PEM_Ln9.T62_oQU240.GMPAS-IAF.cori-knl_intel COMPARE_base_modpes
PASS PEM_Ln9.T62_oQU240.GMPAS-IAF.cori-knl_gnu COMPARE_base_modpes
```
@jonbob I started looking at the halo size in this PR, and tested a making some larger, but no success yet.

**Comment by mark-petersen** [comment]
@jonbob on the current master, this test:
```
./create_test PEM_Ln9.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_intel -q debug --walltime 00:30:00 --force-procs 1024
```
passes, as we know from the nightly tests. But if I set
```
components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
<config_use_Redi>.true.</config_use_Redi>
```
on master, it passes the run but fails the compare. Also, on this PR, I fixed the openMP lines locally, and did the opposite:
```
+<config_use_Redi>.false.</config_use_Redi>
```
(but not  config_gm_closure = 'constant', as you did above). And it passes that PEM!

Could you confirm those results? It appears to contradict your fail with redi off yesterday.

This means there is a halo update or halo size problem within Redi on master, and this PR is not involved. I'll max out the Redi halos next and see if it passes.

jonbob:
I'll run some tests and confirm what you're seeing

jonbob:
@mark-petersen - I can confirm that the PEM test passes on next with redi off. My bad yesterday -- it turns out you can't just change settings in the user_nl files in the run directory, because there are two run directories and they don't get synced.

jonbob:
@mark-petersen - what happened with the increased halos?

jonbob:
@mark-petersen, @vanroekel - what should we do with this PR? We can't leave it breaking tests on next for long. We can either revert it or make the defaults turn redi off until we understand the problem.

vanroekel:
@jonbob I would suggest we put it in but flip redi to off by default until we understand the issue.

jonbob:
Re-merged to next with Redi off by default

jonbob:
merged to master and submitted bless requests for expected DIFF's

PR #3654, opened 2020-06-19
Support for JLSE iris nodes and the Intel GPU

naromero77:
Initial support for the Intel GPU compiler on the JLSE iris nodes.

### DIFF
--- a/cime/config/e3sm/machines/config_compilers.xml
+++ b/cime/config/e3sm/machines/config_compilers.xml
@@ -2205,7 +2205,7 @@ ntel/x86_64/2013/composer_xe_2013/composer_xe_2013_sp1.3.174/mkl/include </base>
   </SLIBS>
 </compiler>
 
-<compiler MACH="jlse-iris" COMPILER="intelGPU">
+<compiler MACH="jlse-iris" COMPILER="intelgpu">
   <CONFIG_ARGS>
     <base> --host=Linux </base>
   </CONFIG_ARGS>
--- a/cime/config/e3sm/machines/config_compilers.xml
+++ b/cime/config/e3sm/machines/config_compilers.xml
@@ -2205,6 +2205,29 @@ ntel/x86_64/2013/composer_xe_2013/composer_xe_2013_sp1.3.174/mkl/include </base>
   </SLIBS>
 </compiler>
 
+<compiler MACH="jlse-iris" COMPILER="intelGPU">
+  <CONFIG_ARGS>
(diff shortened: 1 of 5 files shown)


naromero77:
@xyuan Just to let you know that this is in progress.

naromero77:
The last commit is identical to the commit from the MCT repo. There was no way to cherry-pick from completely different repo (MCT and E3SM) or even create a `git patch` from MCT repo and apply it to E3SM repo via `git am <patch file>`. So, it was done the old fashioned way using a plain vanilla `diff` and `patch` commands.

rljacob:
Can you update the description at the top?  It should be the text that will go in the merge commit.

naromero77:
@rljacob By `top` do you mean the PR description? It already contains the text from the original commit on the CIME files.

Sorry, not following.

rljacob:
Yes the PR description.  I think everything starting with "But the ./case.build" should be deleted.  That should have instead been in the first comment after the PR description.

naromero77:
Its cleaned up now.

cime/config/e3sm/machines/config_compilers.xml
```diff
@@ -2205,6 +2205,29 @@ ntel/x86_64/2013/composer_xe_2013/composer_xe_2013_sp1.3.174/mkl/include </base>
   </SLIBS>
 </compiler>
 
+<compiler MACH="jlse-iris" COMPILER="intelGPU">
```
sarats, line 2208: Very minor issue: can we rename to intelgpu to be consistent with similar naming scheme for Summit etc.?

naromero77: Yes, we can do `intelgpu` instead.

cime/src/externals/mct/mpi-serial/configure
```diff
@@ -3462,6 +3466,9 @@ case $ac_fc_v_output in
   *mGLOB_options_string*)
     ac_fc_v_output=`echo $ac_fc_v_output | sed 's/"-mGLOB[^"]*"/ /g'` ;;
 
+  # With Intel ifx, remove -loopopt=0
+  *-loopopt=0*)
+    ac_fc_v_output=`echo $ac_fc_v_output | sed 's/-loopopt=0//g'` ;;
```
rljacob, line 3471: @naromero77 was this necessary because configure can't tell the difference between the old and new intel compilers?

PR #3657, opened 2020-06-19
Move E3SM config files out of cime directory

rljacob:
Move all the e3sm config files from cime/config/e3sm to E3SM/cime_config.

This is preparation for converting cime to a submodule.

[BFB]


rljacob:
I did a run on compy called WcycTR2 on this branch.  @worleyph or @sarats can you check that all the profiling happened?   The Depends and template files are all found correctly.  Not sure about syslog.

worleyph:
@rljacob , it looks good to me.

worleyph:
Or maybe not - I need to try this myself I guess. There were some checkpoints, but only for the very beginning. I'll try running a longer job, and make sure that syslog would collect more information that we can check for.

worleyph:
syslog worked fine for me (same case on Compy, but capturing checkpoints every 30 seconds).

jgfouca:
Merged to next.

Issue #3622, opened 2020-06-05
land/sea ice mask inconsistency with ne120np4_r0125_oRRS18to6v3


wlin7:
This runtime error arises when trying to run an F-case using ne120np4_r0125_oRRS18to6v3 grid. It is intended to do an F-case nudging simulation to spin up land and derive a land IC at r0125 resolution.

The error message is like
`ERROR: (seq_domain_check)   inconsistency between land mask and sea ice mask:  n=          101  fracl=  -1.434970921420591E-003  fraci=  0.984238415340975`

Below is the backtracing in the codes
  ```
          shr_abort_mod_mp_           114  shr_abort_mod.F90
           seq_domain_mct_mp         427  seq_domain_mct.F90
           cime_comp_mod_mp_        1906  cime_comp_mod.F90
           MAIN__                             122  cime_driver.F90
```
I assumed up to this point what involved are just the domain files. For this case, the following domain files are used.
```
           LND_DOMAIN_FILE = domain.lnd.r0125_oRRS18to6v3.200212.nc
           OCN_DOMAIN_FILE = domain.ocn.oRRS18to6v3.170111.nc
           ICE_DOMAIN_FILE = domain.ocn.oRRS18to6v3.170111.nc

```
Note that with the same domain files, there is no problem with at least two other tri-grids, with difference in just atm grid,, one for lower resolution and the other for hjgher: ne30np4_r0125_oRRS18to6v3 and ne256pg2_r0125_oRRS18to6v3. I believe it is also ok for ne1024np4.

I am puzzled. Could there be other files or factors involved in determining the consistency between land mask and sea ice mask? 

Pinging @mt5555 , @brhillman , @rljacob , @jonbob , @paullric , @czender for help.

wlin7:
The above domain.lnd and domain.ocn were generated from different times.  

The attributes in "domain.lnd.r0125_oRRS18to6v3.200212"  indicates this domain file is generated based on mapfile "/global/homes/t/taylorm/proj/mapping/map_oRRS18to6v3_to_r0125_nco.nc" by @mt5555.

Considering potential inconsistency because the domain.lnd and domain.ocn were not generated based on the same source, I re-did a gen_domain using map_oRRS18to6v3_to_r0125_nco.nc. The new domain.lnd.r0125_oRRS18to6v3 is identical. The new domain.ocn is used to replace the obe above. Same problem remains.

wlin7:
[...]

PR #3663, opened 2020-06-23
Add support for grids using r0125_oRRS18to6v3

wlin7:
Following grids and their pg2 counterparts are added: ne30_r0125_oRRS18to6v3 and ne120_r0125_oRRS18to6v3.
The mapfiles between the components are also populated.

[BFB]

### DIFF
--- a/cime/config/e3sm/config_grids.xml
+++ b/cime/config/e3sm/config_grids.xml
@@ -964,6 +964,16 @@
       <mask>gx1v6</mask>
     </model_grid>
 
+    <model_grid alias="ne30_r0125_oRRS18to6v3" compset="(DOCN|XOCN|SOCN|AQP1)">
+      <grid name="atm">ne30np4</grid>
+      <grid name="lnd">r0125</grid>
+      <grid name="ocnice">oRRS18to6v3</grid>
+      <grid name="rof">r0125</grid>
+      <grid name="glc">null</grid>
+      <grid name="wav">null</grid>
+      <mask>oRRS18to6v3</mask>
+    </model_grid>
+
     <model_grid alias="northamericax4v1_r0125_northamericax4v1" compset="(DOCN|XOCN|SOCN|AQP1)">
       <grid name="atm">ne0np4_northamericax4v1</grid>
       <grid name="lnd">r0125</grid>
@@ -1181,6 +1191,16 @@
       <grid name="atm">ne120np4</grid>
       <grid name="lnd">ne120np4</grid>
       <grid name="ocnice">oRRS18to6v3_ICG</grid>
+      <grid name="rof">r0125</grid>
+      <grid name="glc">null</grid>
+      <grid name="wav">null</grid>
+      <mask>oRRS18to6v3</mask>
+    </model_grid>
+
+    <model_grid alias="ne120_r0125_oRRS18to6v3" compset="(DOCN|XOCN|SOCN|AQP1)">
+      <grid name="atm">ne120np4</grid>
+      <grid name="lnd">r0125</grid>
+      <grid name="ocnice">oRRS18to6v3</grid>
       <grid name="rof">r0125</grid>
       <grid name="glc">null</grid>
       <grid name="wav">null</grid>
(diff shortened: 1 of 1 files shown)


wlin7:
These grids can be used for F-case nudging simulations with ne30 and ne120 atm grids to spin up r0125 land grid with oRRS18to6v3 ocean mask. The generated land restart file (via finidat) can be used by other tri-grid configurations that use the same land grid and ocean mask, esp. for coupling with very high resolution atm grids.

wlin7:
@mt5555 , LDN2ATM_FMAP and LND2ATM_SMAP are using highorder map_r0125_to_ne30np4_highorder.190801.nc. This exists before the current PR.  Would you suggest to change to use mono map as for ne120?

@jonbob , the netcdf header shows you created the highorder map /global/cfs/cdirs/e3sm/inputdata/cpl/gridmaps/ne30np4/map_r0125_to_ne30np4_highorder.190801.nc. Was that also you putting it in config_grids.xml for LND2ATM_FMAP and LND2ATM_SMAP? Please also check out the comments by @mt5555  in issue #3622 on the use of highorder map.

mt5555:
The FMAP  really should be "mono" - so that should be changed (although this is an unusual case so not sure if it's worth the trouble).  

The SMAP could be highorder - it's a far more accurate map, and I was recommending it a while ago for the increased accuracy.  But the weights can have 20% overshoots ( you can see this by running "ncks --chk_map  on the file).   I think we dont really know if this is a problem or not, so until someone studies this, I now err on the side of caution and do not think we should use maps with such large overshoots. 

For r0125->ne30, that's high res -> low res, meaning there will be a lot of downscalings and accuracy is not an issue, so that's another reason in this case to use "mono" as well.

wlin7:
Thanks for the very helpful comments, Mark. LND2ATM maps from r0125 to ne30 are now changed to mono type.

rljacob:
Do working compsets exist for this grid?

wlin7:
> Do working compsets exist for this grid?

Yes, @rljacob . I am using F2010SC5-CMIP6 to go with either ne30 or ne120 atm grids. Other compsets , if there is an interest, should also work.

cime/config/e3sm/config_grids.xml
```diff
@@ -2842,6 +2872,23 @@
       <map name="LND2ATM_SMAPNAME">cpl/gridmaps/fv0.23x0.31/map_fv0.23x0.31_to_ne120np4_aave_110331.nc</map>
     </gridmap>
 
+    <gridmap atm_grid="ne120np4" lnd_grid="r0125">
+      <map name="ATM2LND_FMAPNAME">lnd/clm2/mappingdata/maps/ne120np4/map_ne120np4_to_0.125_nomask_aave.160613.nc</map>
+      <map name="ATM2LND_SMAPNAME">lnd/clm2/mappingdata/maps/ne120np4/map_ne120np4_to_0.125_nomask_aave.160613.nc</map>
```
brhillman, line 2877: You are using mono maps for ne30<->r0125, but aave maps for ne120<->r0125? Should we use mono for both? Pinging @mt5555

mt5555: Yes, we should.  For np4 grids, mono is slightly better than aave.   Probably not worth changing, since this is a v1 compset (E3SM v2 atmosphere will be using "pg2" grid, and thus not this compset).  

Based on this page:

https://acme-climate.atlassian.net/wiki/spaces/ED/pages/178848194/Transition+to+TempestRemap+for+Atmosphere+grids 

ATM2LND_FMAPNAME is an SE->FV flux map, so my recommendation is "mono"
ATM2LND_SMAPNAME is an SE->FV state map, so my recommendation is "intbilin"

rljacob:
Why are we adding ne30 grids to master and not ne30pg2?

whannah1:
Echoing Rob's question, I think it's ok to add more np4 grids, but pg2 should be the default going forward, so any np4 grid support should only be added if the corresponding pg2 grid is also added.

wlin7:
Thanks for the all good comments. These two grid combinations are for a special purpose of generating initial condition for r0125 from atm nudging simulations. The focus is r0125 lnd.  Apparently wan't thinking about pg2 grid. I will add the companion settings for pg2.

wlin7:
Merged to next.

cime/config/e3sm/config_grids.xml
wlin7: Changed to use mono maps. intbilin map generated using ncremap with mwf method failed to pass check. May revisit later.

wlin7:
> This all looks good to me. I notice you don't make any changes to domain files, so I wonder if the defaults that already exist are using the same mapping files or not. 

It is so, @brhillman . I don't need to specify new domain files for these grids. The domain files used 
 are domain.lnd.r0125_oRRS18to6v3 and domain.ocn.oRRS18to6v3. They are based on OCN2LND map and pre-existing, used by grids like ne256pg2 and ne1024pg2.

wlin7:
Merged to master.

PR #3666, opened 2020-06-24
Updated ARKODE-HOMME interface to use SUNDIALS 5.*

cjvogl:
This PR significantly cleans up the ARKODE-HOMME interface by replacing the calls in HOMME to C-wrapper functions written specifically for HOMME with calls to new Fortran ARKODE modules (available in SUNDIALS >= 5.3.0).

This refactor makes the ARKODE-HOMME interface substantially easier to maintain and enables more efficient solver approaches (e.g. a column-wise nonlinear solver approaches similar to the native HOMME method for implicit stages).  All code changes are wrapped within preprocessor-directive condition statements, meaning this should produce BFB results with master when the interface is not active (it is inactive by default).

@gardner48 and @cswoodward contributed in various ways to this.


components/homme/src/theta-l/CMakeLists.txt
```diff
@@ -50,22 +52,13 @@ SET(THETAL_SRCS_F90
...
-  )
-ENDIF ()
-
 # If the user specified a file for custom compiler options use those
 IF (DEFINED THETA_CUSTOM_FLAGS_FILE)
   setCustomCompilerFlags(THETA_CUSTOM_FLAGS_FILE THETAL_SRCS_F90)
 ENDIF ()
 
 # Make SRCS global so the tests can access it
-SET(EXEC_SOURCES ${THETAL_SRCS} ${THETAL_SRCS_C} ${THETAL_SRCS_F90} )
```
mt5555, line 68: was this deletion intentional?

mt5555:
does this need PR #3522 ?  (maybe it includes it?)

components/homme/src/theta-l/CMakeLists.txt
cjvogl: Yes, the `THETA_L_SRCS_C` list only contained files that wrapped the ARKODE C functions in a way that they could be called from Fortran.  Essentially, ARKODE now provides Fortran wrappers so we don't need to maintain those "THETA_L_SRCS_C" files in HOMME.

cjvogl:
> 
> 
> does this need PR #3522 ? (maybe it includes it?)

Ah, I was unaware of that PR.  This PR does include the updates to the `use` statements but not the change from passing `phi_np1` to `get_dirk_jacobian` to `dphi`.

I can include that in this PR and then this will include all of #3522

oksanaguba:
Since these ARKODE PRs are getting in, could i ask for a small change to move ARKODE's version of prim_advance_exp to its own file?

cjvogl:
> 
> 
> Since these ARKODE PRs are getting in, could i ask for a small change to move ARKODE's version of prim_advance_exp to its own file?

@oksanaguba, I'm happy to do that.  Another option would be to eliminate the ARKODE-specific version of `prim_advance_exp` and include the ARKODE-specific code in a single `prim_advance_exp` surrounded be `ifdef` directives.  We (@mt5555 and I) tried this a while back, but we abandoned the idea b/c at the time there the ARKODE-specific code wasn't as compact.  Now it would only require 3 `ifdef` directives: 1 for the `use` statements, 1 for the ARKODE-specific variable declarations, and 1 for the ARKODE-specific calls in the body of `prim_advance_exp`.  Would that be acceptable?

mt5555:
I think a single routine with an ARKODE #ifdef is a good idea.

rljacob:
@cjvogl please make the requested changes.

cjvogl:
> 
> 
> @cjvogl please make the requested changes.

I have.

I am waiting on the following:
1) input from @mt5555 on whether to include #3522 into this PR
2) determination of why combing what should be two identical functions into one is changing the results (when it should not)

mt5555:
If you need anything from #3522, go ahead and merge it into your branch.  But we'll leave #3522 open since it's waiting on a test.

cjvogl:
@rljacob and @mt5555, this should be good to go:  applying the same changes as in #3522 (due to changes in the arguments to `get_dirk_jacobian`) addressed the differences in results.

rljacob:
@oksanaguba please start merging this.

oksanaguba:
Homme suite looks ok (only expected hommetool fail) against the same next branch and e3sm_atm_integration looks ok (all pass).

oksanaguba:
in next

PR #3671, opened 2020-06-28
Implement runtime calculation of pcols in EAM

### LINKED ISSUE #3485: Automated selection of pcols in EAM
@worleyph has run a number of very nice, comprehensive experiments showing that specifying pcols acccording to certain considerations can give meaningful speedup in the EAM physics. See [here](https://acme-climate.atlassian.net/wiki/spaces/EPG/pages/1098778304/Performance+impact+of+optimizing+PCOLS+for+ne30+and+ne30pg2+physics+mesh+resolution) for a description of the experiments. See the top of [this page](https://acme-climate.atlassian.net/wiki/spaces/COM/pages/1112047753/Physgrid+v2+integration+ne30pg2+r05+oECv3) for instructions to manually specify pcols for two compsets of interest. Now that the performance gain is established, it would be useful for the optimal pcols to be calculated automatically.

### LINKED ISSUE #3538: setting PCOLS at runtime (discussion)

worleyph:
In PR #3620, EAM was modified to allow setting pcols, the first
dimension of the chunk data structure, at runtime. See the discussion
in this previous PR and in the associated issue #3538 for a
description as to why this decision was made.

Here, support for setting pcols at runtime is generalized to allow a
default to be calculated that

a) minimizes the number of chunks subject to constructing
chunks all with approximately the same cost (based on either number of
columns or on a user-specified relative cost per column) and
assigning the same number of chunks to each computational thead,
constrained by the physics load balancing option (-1, 0, 1, 2, 3, 4,
or 5) and the target number of chunks per thread; and

b) minimizes the difference between pcols and ncols, where ncols is
the actual number of columns assigned to a chunk, thus minimizing
wasted space and increasing cache locality. This minimization is done
per process, as pcols only needs to be the same for chunks assigned to
a process.

The current options are still supported:

1) set pcols to a specific value at compile-time for all processes by
   adding -pcols <pcols value> to CAM_CONFIG_OPTS in env_build.xml

2) set pcols to a specific value at runtime for all processes by
   adding

       phys_chnk_fdim = <pcols value>

   to user_nl_cam.

However, the new default is

      phys_chnk_fdim = 0

which causes phys_grid_init to compute a pcols value satisfying the
above criteria ( (a) and (b) ), subject to constraints defined by two
new namelist variables:

     phys_chnk_fdim_max:  an upper bound on pcols
     phys_chnk_fdim_mult: a requirement that pcols be a multiple of this value

Note that if phys_chnk_fdim > 1 or if -pcols is specified in
CAM_CONFIG_OPTS, then these namelist variables are ignored.

There is now also support for specifying the defaults for
phys_chnk_fdim_max and phys_chnk_fdim_mult by system, dycore, and
physics grid type. The following rules are included in this PR
(from namelist_defaults_cam.xml):

     
     <phys_chnk_fdim_max > 16 </phys_chnk_fdim_max>
     <phys_chnk_fdim_max dyn="se" npg="2"> 16 </phys_chnk_fdim_max>
     <phys_chnk_fdim_max mach="compy" dyn="se" npg="0"> 24 </phys_chnk_fdim_max>
     <phys_chnk_fdim_max mach="cori-knl" dyn="se" npg="0"> 128 </phys_chnk_fdim_max>

     
     <phys_chnk_fdim_mult > 1 </phys_chnk_fdim_mult>
     <phys_chnk_fdim_mult mach="cori-knl" dyn="se"> 2 </phys_chnk_fdim_mult>

Based on these, the current default (pcols = 16) is still approximated
by the "generic" default for phys_chnk_fdim_max, in that the calculated
pcols will be <= 16, and this will determine the total number of
chunks, so there may necessarily be more than one chunk per
thread. This is simply meant to be a conservative choice, given that
pcols=16 has been working in CAM/EAM for a long time. However, this is
now an upper bound, and pcols could be smaller for a given process for
a given case (grid resolution and number of computational threads).

The other rules are the results of extensive empirical experiments on
Compy and Cori-KNL for the indicated set of options. Appropriate
defaults for other systems or problem settings can be evaluated and
then implemented in the same way. However, these can always be
overridden by setting these variables in user_nl_cam.

There were some subtleties in implementing this capability, requiring,
for example, reordering steps in the atmosphere intialization to allow
pcols to be set in phys_grid_init, and adding the system name to the
EAM/CAM configure command. See the commit messages for the details.

[BFB]
[NML]

Fixes #3538
Fixes #3485

### DIFF
--- a/components/cam/bld/config_files/definition.xml
+++ b/components/cam/bld/config_files/definition.xml
@@ -210,8 +210,7 @@ Maximum number of columns in a chunk (physics data structure).
 Maximum number of sub-columns in a column (physics data structure).
 </entry>
 <entry id="ppcols" value="FALSE">
-Whether pcols is a compile time parameter. Otherwise it is the default value
-for the runtime parameter phys_chnk_fdim.
+Whether pcols is a compile time parameter.
 </entry>
 <entry id="cam_exe" value="cam">
 Name of CAM executable.
@@ -318,6 +317,13 @@ in Perl's $OSNAME variable.  This parameter allows the user to override
 that setting to allow for cross-compilation, and for instances where the
 $OSNAME value is too generic.  For example, currently on both cray-xt and
 bluegene systems $OSNAME has the value "linux".
+</entry>
+<entry id="mach" value="">
+System name for which CAM is being built.  The default value is the first
+part (delimited by '.') of the name returned by Perl's Sys:Hostname hostname
+routine.  This parameter allows the user to override that setting to allow
+for system-specific namelist defaults when the hostname does not return the
+desired name.
 </entry>
 <entry id="use_MMF" valid_values="0,1" value="0">
 Switch to enable or disable MMF/super-parameterization: 0=off, 1=on.
--- a/components/cam/src/dynamics/se/inital.F90
+++ b/components/cam/src/dynamics/se/inital.F90
@@ -13,8 +13,13 @@ contains
 
 subroutine cam_initial(dyn_in, dyn_out, NLFileName)
 
-   use dyn_comp,             only: dyn_init1, dyn_init2, dyn_import_t, dyn_export_t
+   use dyn_comp,             only: dyn_init1, dyn_init2, dyn_import_t, &
+                                   dyn_export_t, frontgf_idx, frontga_idx
    use phys_grid,            only: phys_grid_init
+   use physpkg,              only: phys_register
+   use phys_control,         only: use_gw_front
+   use physics_buffer,       only: pbuf_add_field, dtype_r8
+   use ppgrid,               only: pcols, pver
    use chem_surfvals,        only: chem_surfvals_init
    use cam_initfiles,        only: initial_file_get_id
    use startup_initialconds, only: initial_conds
@@ -32,7 +37,18 @@ subroutine cam_initial(dyn_in, dyn_out, NLFileName)
 
    ! Define physics data structures
    if(par%masterproc  ) write(iulog,*) 'Running phys_grid_init()'
-   call phys_grid_init( )
+   call phys_grid_init()
+
+   ! Initialize index values for advected and non-advected tracers
+   call phys_register()
+
+   ! Frontogenesis indices
+   if (use_gw_front) then
+      call pbuf_add_field("FRONTGF", "global", dtype_r8, (/pcols,pver/), &
+           frontgf_idx)
+      call pbuf_add_field("FRONTGA", "global", dtype_r8, (/pcols,pver/), &
+           frontga_idx)
+   end if
 
    ! Initialize ghg surface values before default initial distributions
    ! are set in inidat.
--- a/components/cam/src/physics/cam/phys_grid.F90
+++ b/components/cam/src/physics/cam/phys_grid.F90
@@ -277,12 +277,26 @@ module phys_grid
                                        ! number of OpenMP threads available to each process
                                        !  (deallocated at end of phys_grid_init)
 
-! Physics fields data structure (chunk) first dimension:
-   integer, private, parameter :: min_pcols = 1
-!pw   integer, private, parameter :: def_pcols = ppcols             ! default 
-!pw++
-   integer, private, parameter :: def_pcols = -1                 ! default 
-!pw--
+! Physics fields data structure (chunk) first dimension (pcols) option:
+!  1 <= pcols_opt: pcols is set to pcols_opt
+!  0 >= pcols_opt: calculate pcols based on lbal_opt, chunks_per_thread, number of
+!       columns and threads in virtual SMP and relative costs per column (if provided), 
+!       attempting to minimize wasted space and number of chunks, subject to:
+!   0 <  pcols_max, then this is an upper bound on the calculated pcols.
+!        Otherwise, pcols_max is ignored.
+!   1 <  pcols_mult, then pcols is required to be a multiple of pcols_mult
+!        (if pcols_max > 0 and pcols_mult <= pcols_max)
+!        Otherwise, pcols_mult is ignored.
+   integer, private, parameter :: def_pcols_opt  = 0              ! default 
+   integer, private :: pcols_opt = def_pcols_opt
+
+   integer, private, parameter :: min_pcols_max  = 1
+   integer, private, parameter :: def_pcols_max  = -1             ! default 
+   integer, private :: pcols_max = def_pcols_max
+
+   integer, private, parameter :: min_pcols_mult = 1
+   integer, private, parameter :: def_pcols_mult = 1              ! default 
+   integer, private :: pcols_mult = def_pcols_mult
 
 ! Physics grid decomposition options:  
 ! -1: each chunk is a dynamics block
@@ -1465,6 +1479,8 @@ logical function phys_grid_initialized ()
 !========================================================================
 !
    subroutine phys_grid_defaultopts(phys_chnk_fdim_out, &
+                                    phys_chnk_fdim_max_out, &
+                                    phys_chnk_fdim_mult_out, &
                                     phys_loadbalance_out, &
                                     phys_twin_algorithm_out, &
                                     phys_alltoall_out, &
@@ -1478,6 +1494,10 @@ logical function phys_grid_initialized ()
 !------------------------------Arguments--------------------------------
      ! physic data structures declared first dimension
      integer, intent(out), optional :: phys_chnk_fdim_out
+     ! physic data structures declared first dimension upper bound
+     integer, intent(out), optional :: phys_chnk_fdim_max_out
+     ! physic data structures declared first dimension required factor
+     integer, intent(out), optional :: phys_chnk_fdim_mult_out
      ! physics load balancing option
      integer, intent(out), optional :: phys_loadbalance_out
      ! algorithm to use when determining column pairs to assign to chunks
@@ -1490,7 +1510,13 @@ logical function phys_grid_initialized ()
      logical, intent(out), optional :: phys_chnk_cost_write_out
 !-----------------------------------------------------------------------
      if ( present(phys_chnk_fdim_out) ) then
-       phys_chnk_fdim_out = def_pcols
+       phys_chnk_fdim_out = def_pcols_opt
+     endif
+     if ( present(phys_chnk_fdim_max_out) ) then
+       phys_chnk_fdim_max_out = def_pcols_max
+     endif
+     if ( present(phys_chnk_fdim_mult_out) ) then
+       phys_chnk_fdim_mult_out = def_pcols_mult
      endif
      if ( present(phys_loadbalance_out) ) then
        phys_loadbalance_out = def_lbal_opt
@@ -1516,6 +1542,8 @@ logical function phys_grid_initialized ()
 !========================================================================
 !
    subroutine phys_grid_setopts(phys_chnk_fdim_in, &
+                                phys_chnk_fdim_max_in, &
+                                phys_chnk_fdim_mult_in, &
                                 phys_loadbalance_in, &
                                 phys_twin_algorithm_in, &
                                 phys_alltoall_in,    &
@@ -1532,6 +1560,10 @@ logical function phys_grid_initialized ()
 !------------------------------Arguments--------------------------------
      ! physic data structures declared first dimension
      integer, intent(in), optional :: phys_chnk_fdim_in
+     ! physic data structures declared first dimension upper bound
+     integer, intent(in), optional :: phys_chnk_fdim_max_in
+     ! physic data structures declared first dimension required factor
+     integer, intent(in), optional :: phys_chnk_fdim_mult_in
      ! physics load balancing option
      integer, intent(in), optional :: phys_loadbalance_in
      ! option to use load balanced column pairs
@@ -1543,8 +1575,8 @@ logical function phys_grid_initialized ()
      ! flag whether to write out estimated and actual cost per chunk
      logical, intent(in), optional :: phys_chnk_cost_write_in
 !-----------------------------------------------------------------------
-     if ( present(phys_chnk_fdim_in) ) then
 #ifdef PPCOLS
+     if ( present(phys_chnk_fdim_in) ) then
         if (phys_chnk_fdim_in /= pcols) then
            if (masterproc) then
               write(iulog,*)                                     &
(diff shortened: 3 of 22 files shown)


worleyph:
This PR has been extensively tested using F cases (ne4, ne4pg2 on 40x1 and 96x1) on Compy and (ne30, ne30pg2, 675x1 and 2700x2) on Cori-KNL using the Intel compiler, against all of the different phys_loadbalance options and with example column cost specifications.

If approved and merged, there will be one more PR, to tweak the algorithm assigning chunks to processes. While this final PR is not related directly to adding support for runtime pcols, the development and testing of this capability led to the discovery that such tweaking would be useful.

worleyph:
@whannah1 , I've asked you to review this because it affects the MMF work (potentially) as much as anyone's. In particular, there is some support explicitly for when providing column cost estimates (though you may need to set phys_chnk_fdim_max larger than the current defaults to see much advantage to the new logic). If this is not possible, please just ask and I'll remove you from the reviewer list.

worleyph:
@singhbalwinder , I looked at the "conflicting file" and do not see why there is a conflict (and I do not know what the process is for resolving conflicts). A recent PR added

         use cam_control_mod,  only: moist_physics

and I removed

        use phys_control,     only: use_gw_front
        use physics_buffer,   only: pbuf_add_field, dtype_r8
        use ppgrid,           only: pcols, pver

and then edited

         use cam_abortutils,   only : endrun

to fix the spacing

         use cam_abortutils,   only: endrun

worleyph:
@ambrad , note that the rules that were identified for optimal pcols in the earlier experiments are encoded here (for the default runtime option).

ambrad:
This looks super nifty. If you run ne30pg2 on 30 nodes of Compy, do you get pcols=18, or at least 9 if your maximum of 16 does not permit this by default? (But I believe 18 is better than 9.)

worleyph:
@ambrad, yes - currently you'd get 9. The earlier experiments indicated that 16 with 2 chunks per thread was better than 32 with 1 chunk per thread, for 675x1 PE layout on both Cori-KNL and on Compy, so I went with 16 as the default. You could always override this in user_nl_cam, but maybe a better upper bound would be 24. I'll check this out, and push a change if this is a better default. (24 is the default for Compy for ne30, so that kind of makes sense).

worleyph:
@ambrad, yep - 18 is slightly faster than 9 (around 1% for "CPL:ATM_RUN" and 2.5% for bc_physics + ac_physics) ; I'll check Cori-KNL as well (tomorrow), and make a change to bump the max up to 24 for Compy (at least), to capture this result. Thanks for pointing this out.

singhbalwinder:
@worleyph : This looks great! Thanks for this PR, it is a nice addition to the model. The conflict seems very minor. I will resolve it while merging it to next and master.

worleyph:
@singhbalwinder , it you want to wait until tomorrow to push to next, I'll have an update to the "rules" for Compy and Cori-KNL (for pg2). No big deal if you don't want to wait, as I can add these as part of my next PR.

singhbalwinder:
@worleyph : Please go ahead and push those changes. I will have to wait for reviewers approval before merging it to next.

worleyph:
@ambrad , thanks for pointing out that 1 chunk with pcols=18 is better than 2 chunks with pcols=9 (for the F case using pg2). I ran a suite of experiments, and 1 chunks with pcols=20 is better than 2 chunks with pcols=10 as well (but 2 chunks with pcols=12 is better than 1 chunkd with pcols=24, and so on for even larger pcols). I've reset the rule to set phys_chnk_fdim_max = 20 for Compy by default when using pg2.

I also ran similar experiments on Cori-KNL, and for that system I changed to rule to set phys_chnk_fdim_max = 30 by default when using pg2 (2 chunks with pcols=16 is still better than 1 chunks with pcols=32, but 2 chunks with pcols=16 is worse than 1 chunks with pcols=30).

For ne4 (my quick validation experiment after the change) on Compy 1 chunks with pcols=24 is a little bit better than 2 chunks with pcols=12, but they are close, so I'll go with the ne30-based results. Anything in this neighbor should be a reasonable default, and Compy is not that sensitive in any case. 

You are welcome to keep poking at this (or to tell me to do so) to further improve these default. The optimum will likely be case specific, and the goal is simply to keep the defaults in the ball park.

worleyph:
Sounds like a plan. After I get my next PR through (touching code in the same region) I’ll look at refactoring (or just trying to clean up). I’ll then ask you for advice on creating some tests, if I can’t get you to do this yourself :-).

components/cam/bld/config_files/definition.xml
```diff
@@ -319,6 +318,13 @@ that setting to allow for cross-compilation, and for instances where the
 $OSNAME value is too generic.  For example, currently on both cray-xt and
 bluegene systems $OSNAME has the value "linux".
 </entry>
+<entry id="mach" value="">
```
whannah1, line 321: So does this need to be set separate from the "mach" value that goes into the create_new_case command?
BTW, Summit is another problematic machine in this respect. The value for hostname is like "login3".

worleyph: I put this is because similar logic is in place for identifying the OS (right above or below this), and something needs to be specified. In practice, the mach variable from create_newcase is used. See cam/cime_config/config_component.xml, where I added

           <value compset=""               >-mach $MACH</value>

This shows up as the first option in the CAM_CONFIG_OPTS line in env_build.xml.

components/cam/src/dynamics/se/inital.F90
```diff
@@ -32,7 +37,18 @@ subroutine cam_initial(dyn_in, dyn_out, NLFileName)
...
    ! Define physics data structures
    if(par%masterproc  ) write(iulog,*) 'Running phys_grid_init()'
-   call phys_grid_init( )
+   call phys_grid_init()
+
+   ! Initialize index values for advected and non-advected tracers
+   call phys_register()
+
+   ! Frontogenesis indices
+   if (use_gw_front) then
```
whannah1, line 46: Why not push these pbuf_add_field() calls down into phys_register() since there are already a lot of pbuf_add_field() calls in there?

worleyph: I think that I asked someone about this, and was encouraged to leave these pbuf_add_field calls in the dynamics. In any case, the relevant variables (frontgf_idx and frontga_idx) are defined in dyn_comp.F90, and are not in all of the dycores. While we are not supporting any dycore but SE, I am trying not to deliberately introduce anything  that would prevent restoring the functionality of one of these other dycores. There was less of a logicial structure change by doing it this way. If one of the EAM developers wants to make this move, I don't have an opinion (beyond what I just said).
In summary - I did not think that it was my place to propose such a change, and it was not necessary for what I needed.

components/cam/src/physics/cam/phys_grid.F90
```diff
@@ -1221,55 +1314,50 @@ subroutine phys_grid_init( )
...
-          max_process_ncols = process_ncols(p)
-      enddo
+      min_process_nthreads = minval(npthreads)
+      max_process_nthreads = maxval(npthreads)
+      min_process_nchunks  = minval(npchunks)
+      max_process_nchunks  = maxval(npchunks)
+      min_process_ncols    = minval(process_ncols)
+      max_process_ncols    = maxval(process_ncols)
+      min_pcols            = minval(pcols_proc)
+     !max_pcols            = maxval(pcols_proc)    already calculated
```
whannah1, line 1324: I would take out this commented line since it's not needed. If you want to leave a note about this was calculated above then just do it in a normal comment block.

components/cam/src/physics/cam/phys_grid.F90
```diff
@@ -4540,13 +4704,81 @@ subroutine create_chunks(opt, chunks_per_thread)
...
+         endif
+         if (cols_per_chnk_max > 0) then
+! if calculated pcols is too large, recalculate with more chunks per thread
+            tmp_chnks_per_thrd = chnks_per_thrd
+            do while (pcols_vsmp(smp) > cols_per_chnk_max)
+               tmp_chnks_per_thrd = tmp_chnks_per_thrd + 1
+               pcols_vsmp(smp) = &
+                  ceiling(real(nvsmpcolumns(smp),r8)/real(nvsmpthreads(smp)*tmp_chnks_per_thrd,r8))
+               if (opt == 5) then
+! pcols should not be (much) larger than the maximum block size
```
whannah1, line 4769: To match the style above I think these single comment lines should be indented. Personally, I find the left-justified comment make the logic/loop structure harder to parse visually, so it becomes a distraction.

components/cam/src/physics/cam/phys_grid.F90
```diff
@@ -4540,13 +4704,81 @@ subroutine create_chunks(opt, chunks_per_thread)
...
+   maxblksiz_vsmp(:) = 0
+   do j=firstblock,lastblock
+      p = get_block_owner_d(j)
+      smp = proc_vsmp_map(p)
+      blksiz = get_block_gcol_cnt_d(j)
+      if (blksiz > maxblksiz_vsmp(smp)) maxblksiz_vsmp(smp) = blksiz
+   enddo
+
+!
+! Calculate pcols for chunks in each virtual SMP
```
whannah1, line 4738: I also agree with @ambrad's sentiment that this block should ideally be it's own routine that could be covered with a  unit test. Same goes for the section below that utilizes the optional column costs.

components/cam/src/physics/cam/phys_grid.F90
worleyph: Will do. Thanks.

components/cam/src/physics/cam/phys_grid.F90
```diff
@@ -4610,15 +4834,55 @@ subroutine create_chunks(opt, chunks_per_thread)
...
+! pcols_vsmp based on the actual number of columns assigned to each chunk), 
+! but it is needed now in order to allocate the gcol array in the
+! chunks data structure. The gcol array will be reallocated after the return
+! from the create_chunks routine, after which there will be no wasted space.
+            max_pcols = maxval(pcols_vsmp)
+         endif
+      endif
+#endif
 !
 ! Allocate chunks and knuhcs data structures
```
whannah1, line 4881: I realize this is perpendicular to this PR, but what's the point of the 'knuhcs' variable? I've never noticed it, but I imagine it could probably have a more accurately descriptive name.

components/cam/src/physics/cam/phys_grid.F90
worleyph: There is a mix of styles in this routine, but, you are right, I picked the less used style. I'll rework this, at least cleaning up the comments in the code that I just added. Thanks.

components/cam/src/physics/cam/phys_grid.F90
worleyph: Okay for this to be future work?

components/cam/src/physics/cam/phys_grid.F90
worleyph: inverse of chunks. This is a very old choice (mine, trying to be clever I guess).

components/cam/src/physics/cam/phys_grid.F90
whannah1: yea sure, just wanted to vocalize my support for more modularity.

components/cam/src/dynamics/se/inital.F90
whannah1: ok, right, this is obviously outside the scope of this PR.

worleyph: FYI - I asked this identical question in github issue  #3538 , and you were the only one to respond :-). After seeing that not all dycores support the  frontal gw scheme, I gave up looking at adding it to phys_register.

whannah1: Sounds like we should just take a machete to these other dycor options...

worleyph:
@whannah1, note that we can also add rules (for default phys_chnk_fdim_max and _mult) for when use_MMF == 1, if you find that the current defaults are inappropriate for MMF runs. I expect that you might want to make use_cost_d = .true. the default for MMF, whether you read cost_d(:)  in or or provide a cost estimate "analytically", in which case increasing phys_chnk_fdim_max larger than the current defaults may be useful. This larger phys_chnk_fdim_max might be best calculated internally based on the range of values in cost_d(:), but that is future work, and perhaps should wait until we have a real use case. I've been trying to anticipate needs here, and worry that I will misunderstand your actual needs.

whannah1:
@worleyph, The cost feature is going to be a great resource for the MMF work, but it won't be the default unless we decide that the regional load balancing feature helps with testing. So I agree that we shouldn't anticipate any modifications until we have more experience using the column cost and have a real use case.

components/cam/src/physics/cam/phys_grid.F90
worleyph: @whannah1 , I made a sweep through the routine create_chunks (which the above code is from) to use a consistent style for all comments. I adopted what I think that you prefer. This has been pushed to the branch.

components/cam/src/physics/cam/phys_grid.F90
worleyph: @whannah1: Done - just deleted the line. Has been pushed to the branch.

components/cam/src/physics/cam/phys_grid.F90
whannah1: I would actually prefer that the "bang" character is also indented. The non-indented style seems to harken back to the F77 days.

worleyph:
@singhbalwinder , just remembered about the need to add NML since this adds namelist variables (that show up in atm_in with default settings).

components/cam/src/physics/cam/phys_grid.F90
worleyph: Done. Hopefully close enough that further tweaking can wait until this part of the code is refactored by introducing more modularity.

whannah1: @worleyph Thanks for doing that!

worleyph:
@singhbalwinder , I think that this one is ready to be merged to next. Thanks.

singhbalwinder:
Pushed to next after resolving a minor conflct.

PR #3672, opened 2020-06-29
CIME fix for setting maximum I/O requests automatically

### LINKED ISSUE #3658: PIO MAX_PEND_REQ parameter in env_run.xml not being interpreted correctly
in env_run.xml is the parameter

    <entry id="PIO_REARR_COMM_MAX_PEND_REQ_COMP2IO" value="0">

and the description

     <desc>pio rearranger communication max pending requests (comp2io) :
       0 implies that CIME internally calculates the value ( = max(64, 2 * PIO_NUMTASKS) ), 
      -1 implies no bound on max pending requests </desc>

The default value is 0. With master using scorpio, the internal variable max_pend_req inherits this value (0, and not what is described above). Internally, a value of max_pend_req == 0 signals to use MPI_AlltoAllw for the communication pattern, instead of the point-to-point algorithm. Therefore, the defaults

    <entry id="PIO_REARR_COMM_TYPE" value="p2p">
    <entry id="PIO_REARR_COMM_FCD" value="2denable">
    <entry id="PIO_REARR_COMM_MAX_PEND_REQ_COMP2IO" value="0">
    <entry id="PIO_REARR_COMM_ENABLE_HS_COMP2IO" value="TRUE">
    <entry id="PIO_REARR_COMM_ENABLE_ISEND_COMP2IO" value="FALSE">

which should specify a point-to-point algorithm using handshaking and not using MPI_Isend will instead use MPI_AlltoAllw. Changing to only 

    <entry id="PIO_REARR_COMM_MAX_PEND_REQ_COMP2IO" value="64">

implements the expected communication protocol. (This was verified by adding instrumention to see what protocol was actually being used in the two cases.)

The reason that this is important (beyond being misdocumented otherwise) is that MPI_AlltoAllw performs terribly for the comp2io scorpio communication pattern with typical number of I/O reader/writers. In the example that led to this discovery (Issue #3650), the performance difference is a factor of 30 on Cori-KNL with 16384 processes.

jayeshkrishna:
The maximum I/O pending requests setting in env_run.xml,
PIO_REARR_COMM_MAX_PEND_REQ_COMP2IO, is by
default set to 0. CIME needs to set the variable to an
appropriate value in this case (when its set to 0).
Re-introducing this behavior that was disabled by
ESMCI/cime#3499.

Also updating the logging of the I/O rearranger options
to make it more intuitive to the user

Also see ESMCI/cime#3597 that fixes the disabling of
automatic setting of I/O requests in CIME

Fixes #3658 

[BFB]

### DIFF
--- a/cime/src/drivers/mct/cime_config/config_component.xml
+++ b/cime/src/drivers/mct/cime_config/config_component.xml
@@ -2430,7 +2430,7 @@
     <default_value>64</default_value>
     <group>run_pio</group>
     <file>env_run.xml</file>
-    <desc>pio rearranger communication max pending requests (io2comp) : -1 implies no bound on max pending requests </desc>
+    <desc>pio rearranger communication max pending requests (io2comp) : 0 implies that CIME internally calculates the value (= 64), -1 implies no bound on max pending requests </desc>
   </entry>
 
   <!-- Handshaking is a very bad idea for large process counts and for -->
--- a/cime/src/drivers/nuopc/cime_config/config_component.xml
+++ b/cime/src/drivers/nuopc/cime_config/config_component.xml
@@ -1989,7 +1989,7 @@
     <default_value>64</default_value>
     <group>run_pio</group>
     <file>env_run.xml</file>
-    <desc>pio rearranger communication max pending requests (io2comp) : -1 implies no bound on max pending requests </desc>
+    <desc>pio rearranger communication max pending requests (io2comp) : 0 implies that CIME internally calculates the value (= 64), -1 implies no bound on max pending requests </desc>
   </entry>
 
   <!-- Handshaking is a very bad idea for large process counts and for -->
--- a/cime/src/share/util/shr_pio_mod.F90
+++ b/cime/src/share/util/shr_pio_mod.F90
@@ -812,7 +812,7 @@ contains
       end select
 
       ! buf(3) = max_pend_req_comp2io
-      if((pio_rearr_comm_max_pend_req_comp2io < 0) .and. &
+      if((pio_rearr_comm_max_pend_req_comp2io <= 0) .and. &
           (pio_rearr_comm_max_pend_req_comp2io /= PIO_REARR_COMM_UNLIMITED_PEND_REQ)) then
 
         ! Small multiple of pio_numiotasks has proven to perform
@@ -844,7 +844,7 @@ contains
       end if
 
       ! buf(6) = max_pend_req_io2comp
-      if((pio_rearr_comm_max_pend_req_io2comp < 0) .and. &
+      if((pio_rearr_comm_max_pend_req_io2comp <= 0) .and. &
           (pio_rearr_comm_max_pend_req_io2comp /= PIO_REARR_COMM_UNLIMITED_PEND_REQ)) then
         write(shr_log_unit, *) "Invalid PIO rearranger comm max pend req (io2comp), ", pio_rearr_comm_max_pend_req_io2comp
         write(shr_log_unit, *) "Resetting PIO rearranger comm max pend req (io2comp) to ", PIO_REARR_COMM_DEF_MAX_PEND_REQ
--- a/cime/src/share/util/shr_pio_mod.F90
+++ b/cime/src/share/util/shr_pio_mod.F90
@@ -867,16 +867,6 @@ contains
         buf(8) = 0
       end if
 
-      ! Log the rearranger options
-      write(shr_log_unit, *) "PIO rearranger options:"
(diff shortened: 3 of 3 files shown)


worleyph:
@jayeshkrishna , I also ran some quick experiments, after adding print statements to swapm to see whether MPI_Alltoallw or the point-to-point algorithm was used, and the default:
 
     PIO_REARR_COMM_TYPE = 'p2p'
     PIO_REARR_COMM_FCD = '2denable'
     PIO_REARR_COMM_MAX_PEND_REQ_COMP2IO = 0

 did use point-to-point, while setting 

     PIO_REARR_COMM_TYPE = 'coll' 

did use MPI_Alltoallw (ignoring all of the rest of the option values).

I knew what these meant in PIO1, but what is being selected for when

     PIO_REARR_COMM_TYPE == 'default'

and what happens when

     PIO_REARR_COMM_FCD = 'io2comp', 'comp2io', 'disable', or 'default'

that is, what is selected for comp2io when PIO_REARR_COMM_FCD == 'io2comp', etc.

jayeshkrishna:
The default for
```
PIO_REARR_COMM_TYPE == 'default'
```
the communication type is "p2p"
For the flow control direction, if the user chooses "io2comp", flow control is only enabled from I/O processes to compute processes (p2p, no handshake and blocking sends for communication from compute processes to I/O processes). The default is "2denable" (flow control in both directions).

worleyph:
@jayeshkrishna , thanks for the explanation.

dqwu:
@jayeshkrishna 
The typical use case that users encounter:
* 0 (implies that CIME internally calculates the value)
* -1 (PIO_REARR_COMM_UNLIMITED_PEND_REQ, implies no bound on max pending requests)
* Other negative values (invalid values, but might be handled the same as 0)
* Some positive values (valid values)

For pio_rearr_comm_max_pend_req_comp2io <= 0, "User-specified PIO rearranger ..." message is used so far.

We might need more verbose user messages for each case (an if statement for user messages).

jayeshkrishna:
I can modify the error/user messages to distinguish between invalid values (< 0) vs case where CIME calculates the value (== 0) . The unlimited case and valid values case should be evident in the existing (in this PR) messages

amametjanov:
Related to cleanup of logged messages, could you add fixes to the following warnings?
IBM compiler:
```
/autofs/nccs-svm1_home1/azamat/repos/E3SM-tests/externals/scorpio/src/clib/pio_darray_int.c:840:47: warning: address of array 'file->varlist' will always evaluate to 'true' [-Wpointer-bool-conversion]
    pioassert(file && file->iosystem && file->varlist && varids && varids[0] >= 0 &&
                                     ~~ ~~~~~~^~~~~~~

/autofs/nccs-svm1_home1/azamat/repos/E3SM/externals/scorpio/src/clib/pio_darray.c:1551:281: warning: format specifies type 'unsigned int' but the argument has type 'MPI_Datatype' (aka 'struct ompi_datatype_t *') [-Wformat]
                                "Writing variable (%s, varid=%d) to file (%s, ncid=%d) failed. Unable to find a default fillvalue for variable, unsupported variable type (MPI type = %x)", pio_get_vname_from_file(file, varid), varid, pio_get_fname_from_file(file), file->pio_ncid, vtype);
                                                                                                                                                                                      ~~                                                                                                ^~~~~
```

GNU, PGI and IBM:
```
/autofs/nccs-svm1_home1/azamat/repos/E3SM-tests/externals/scorpio/src/clib/pio_sdecomps_regex.cpp:251:7: warning: control may reach end of non-void function [-Wreturn-type]
      }
      ^
```

jayeshkrishna:
@amametjanov I have created a Scorpio issue regarding the warnings - https://github.com/E3SM-Project/scorpio/issues/322

jayeshkrishna:
@dqwu : I have modified the messages by updating the branch. Does the latest messages look good to you (I have also added the required XML doc for io2comp case when the user specifies the value as 0)?

dqwu:
@jayeshkrishna 
For PIO_REARR_COMM_MAX_PEND_REQ_COMP2IO, the existing desc has:
0 implies that CIME internally calculates the value ( = max(64, 2 * PIO_NUMTASKS) )
The value is calculated based on PIO_NUMTASKS.

For PIO_REARR_COMM_MAX_PEND_REQ_IO2COMP, the new desc has:
0 implies that CIME internally calculates the value (= 64)
Since the value is set to a constant, should we use "sets" instead of "calculates" here?

jayeshkrishna:
The description is more generic (we could change the behavior later, right now its just set to 64)

PR #3673, opened 2020-06-29
[WIP] HOMME: added possibility of building libraries instead of execs

bartgol:
This PR allows one to configure Homme via cmake, without building executables, and building 'libraries' instead.

This can be handy if Homme is built as part of a larger cmake project (namely, E3SM, or SCREAM), so that we can simply link against homme's pre-built library, without injecting much of its targets setting into the host project (i.e., simply doing `target_link_library(my_tgt $my_homme_lib)` should be all you need).

### DIFF
--- a/components/homme/CMakeLists.txt
+++ b/components/homme/CMakeLists.txt
@@ -12,15 +12,17 @@ IF (NOT DEFINED PROJECT_NAME)
   SET (LAST_GIT_COMMIT_SHA ${LAST_GIT_COMMIT_SHA} CACHE STRING "The sha of the last git commit.")
   MESSAGE (STATUS "The sha of the last commit is ${LAST_GIT_COMMIT_SHA}")
 
-  # Build executables/tests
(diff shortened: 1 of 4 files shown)


rljacob:

bartgol:
I paused after some discussion on the hommexx slack channel. There is some disagreement regarding how to treat non-standalone vs standalone build. My goal was to discuss it at the next dycore meeting, but we haven't had one in a while.

I am planning on resuming work on this soon, though. I do think we need this PR (possibly amended, in some form), to go in, in order for SCREAM to properly work.

rljacob:
What is the update on this PR?

bartgol:
@rljacob I'm waiting to finish the PR that brings in homme-nh-xx first (coming today or tomorrow), so that I do cmake changes just once.

bartgol:
I'm closing this PR, since I need to fully test the capability in the SCREAM repo before I can claim the mods are working. When I'm confident the upgrades are completed, I will issue a new PR.

PR #3675, opened 2020-06-30
Enable CRM friction 

whannah1:
The MMF is almost always run in a 2D configuration in which momentum fields are un-coupled, and so surface friction is felt very indirectly and often does not match the direction of near-surface wind in the CRM, which can lead to unrealistic boundary layer structures. This PR enables a pre-existing CRM friction calculation to be on by default, which does not affect the MMF coupling or double count the actual friction because the momentum fields remain uncoupled. Eventually the ESMT scheme will handle all convective momentum transport in all MMF configurations after we port it to SAM++ and validate that it runs well on GPUs. 

[non-BFB]

### DIFF
--- a/components/cam/bld/namelist_files/namelist_defaults_cam.xml
+++ b/components/cam/bld/namelist_files/namelist_defaults_cam.xml
@@ -1174,7 +1174,7 @@
 <srf_flux_avg eddy_scheme="HBR"      >0</srf_flux_avg>
 <srf_flux_avg eddy_scheme="diag_TKE" >0</srf_flux_avg>
 <srf_flux_avg eddy_scheme="CLUBB_SGS">0</srf_flux_avg>
-<srf_flux_avg use_MMF="1"          >1</srf_flux_avg>
+<srf_flux_avg use_MMF="1"          >0</srf_flux_avg>
 
 <!-- MMF CRM mean state acceleration -->
 <use_crm_accel    use_MMF="0">.false.</use_crm_accel>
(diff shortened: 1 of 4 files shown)


whannah1:
To provide some validation of this change, I have an older pair of 1-month ne30pg2 runs and I also ran a pair of 5 member ensembles with ne4pg2 for 20 days. 

Here's a list of some global mean statistics comparing the two ensemble means. This is meant to quantitatively show that enabling the friction does not change the general behavior of the model. 
```
                  Avg 1          Avg 2           RMSE
   PRECT          2.227          2.215         0.3576
   LHFLX          67.17          66.81          2.411
   SHFLX          21.27          21.63         0.9736
     TMQ          24.55          24.56         0.2655
TGCLDLWP        0.06713        0.06734       0.006135
TGCLDIWP        0.05082        0.05074       0.005002
    FLNT          221.5          221.3           2.47
    FLNS          54.94          54.52           1.43
```

Here's the same table of stats from the 1-month mean data of the ne30pg2 runs. The RMSE differences might seem large, like for PRECT, but that's just the result of precip happening in one case where there was none in the other. I'm sure these would look more reasonable if these cases were run out for a year or more. 

```
                  Avg 1          Avg 2           RMSE
   PRECT          2.728          2.716          2.156
   LHFLX          78.76          78.64          15.53
   SHFLX          18.16          18.53           6.47
     TMQ          24.95          24.92          1.772
TGCLDLWP         0.0743        0.07413         0.0247
TGCLDIWP        0.04995        0.04975        0.02184
    FLNT          231.6          231.3          9.077
    FLNS          51.34          50.74           5.73
    FSNT          243.9          243.7          12.08
    FSNS          163.0          162.8          13.62
```

Also from the older 1-month ne30pg2 runs, here's 1-month mean maps of zonal wind at 850 and 200 mb to get a sense that the wind patterns are similar. 

image

And here's a few more time mean plots to show that convection related fields are also not strongly affected.

image

I intentionally opted not to show the differences here, but I can plot them if someone wants to see them. The spatial differences in these runs is the result of different weather, which is substantial for a 1-month run. If these were 12-month runs then I think the difference maps would be useful.

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #3677, opened 2020-06-30
Update MMF tests

whannah1:
The MMF "test" compsets were originally created to shorten the run time of tests, but without any special development needs these compsets are unnecessary and just clutter up the compset configuration. If we decide we need smaller tests again it will be better to add a specific test mod rather than a special compset. 

[non-BFB] only for MMF tests since they will change


brhillman:
Not sure if this needs the non-BFB label, this will change the MMF test we use in the integration tests, but the code itself is BFB. New baselines will need to be added.

whannah1:
I updated the text to mention that it's non-BFB

whannah1:
Updated the branch to address a couple minor conflicts created by recent PRs.

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #3680, opened 2020-07-02
Update provenance capture logic in maint-1.1

### LINKED ISSUE #3659: full performance archiving support for Compy and Anvil missing in Maint versions
In tags V1.2.0-beta, V1.1.1, V1.1.0, and maint-1.0-pf-v3.8, provenance.py is missing any mention of the system Compy, and it still has Anvil listed as using the PBS batch system (and not SLURM). Because of this, certain provenance and system information is not being archived when jobs based on these tags are run. Based on a recent examination of the performance archive on Compy, this is true for over half of the Compy E3SM workload.

worleyph:
When Compy support was added for maint-1.1, the provenance.py file
was not updated to also enable capture of provenance data for Compy.
The job monitoring script syslog.compy was also not added.

Also, when Anvil changed from a PBS batch scheduler to SLURM, the
provenance.py and syslog.anvil files in maint-1.1 were not updated
accordingly.

Finally, the projects for which performance archiving is supported
by default has not been updated in maint-1.1 for cori-haswell,
cori-knl, theta, and summit and the path to the performance archive
has not been updated in maint-1.1 for cori-haswell, cori-knl, and
summit.

Because of these missing changes, expected provenance data is not
being captured with the other performance data in the performance
archive. Here these deficiencies are remedied.

The job process monitoring scripts (syslog.XXX) for cori-haswell,
cori-knl, summit, and Tteta are also updated, as they have slightly
improved behavior over the originals.

The decommissioned systems edison, mira, and titan were also removed
from provenance.py, but not from config_machines.xml, and the associated
job progress monitoring scripts were deleted.

Note that maint-1.1 does not work on anvil, cori-haswell,
cori-knl, and theta currently (experiments were not run on
Summit, and it does work on Compy), so these updates to the
performance archiving logic on these systems are simply for
documentation purposes.

[BFB]

Fixes #3659 for maint-1.1


worleyph:
Tested correctness of performance archiving on Compy (F case on ne4_ne4 grid). I also tried on Anvil, but maint-1.1 is missing some support there? ('ERROR: No queues found"). Will try again, if someone can explain what I should do to work around this.

Will also look at Cori-Haswell, Cori-KNL, and Theta, (to the extent that maint-1.1 works there).

worleyph:
FYI - I can't get maint-1.1 to work on Cori-KNL either. First, have to replace acme by es3m, and then it is dying in generate_cice_decomp.pl becuase something is trying to pass 64.0 (instead of 64) to the -nproc option.

Giving up. If someone wants me to test this some place other than compy, please advise where (and how).

worleyph:
a) Tested on Compy, which worked.

b) Tested on Anvil, which failed during create_newcase because 

     ERROR: No queues found

adding '-q acme-small' only generated

     WARNING: queue 'acme-small' is unknown to this system

after which is failed with the same error message.

c) Tested on Cori-Haswell, which failed during case.setup with:

     cray-hdf5-parallel(3):ERROR:105: Unable to locate a modulefile for 'cray-hdf5-parallel/1.10.2.1'

d) Tested on Cori-KNL, which failed during case.setup with:

      ERROR: Command: './generate_cice_decomp.pl -ccsmroot XXX -res ne30np4 -nx 48602 \
      -ny 1 -nproc 1200.0 -thrds 1 -output all' failed with error 'b'Value "1200.0" invalid for option \
      nproc (number expected)

e) Tested on Theta, which failed during case.setup

     cray-mpich(3):ERROR:105: Unable to locate a modulefile for 'cray-mpich/7.6.2'
     cray-netcdf(3):ERROR:105: Unable to locate a modulefile for 'cray-netcdf/4.4.1.1.3'

So it appears the maint-1.1 has only been kept running on Compy, and I can't test my modifications on the other systems until they have been successfully reported.

worleyph:
Since only Compy is a supported system for maint-1.1, assigning this to the Compy POC.

PR #3689, opened 2020-07-10
Improve load balancing when assigning chunks

worleyph:
The current EAM physics load balancing scheme attempts to create
chunks that are all have approximately the same computational cost,
and then assigns the chunks to processes in such a way as to minimize
the number of columns that need to be sent to other processes during
d_p_coupling and p_d_coupling (hopefully decreasing communication
overhead). However, if not all chunks have the identical number of
columns or if the columns are not all of equal cost, then the
more expensive chunks are not necessarily spread equally between the
processes, and unnecessary load imbalance can result. Here,
modifications are made to improve the load balance in these
situations, while still preserving some of the communication avoiding
capability of the original algorithm.

To enable this optimization, the threading schedule for chunks in
physpkg is specifed as SCHEDULE(static,1), so that chunks are assigned
in a wrap map fashion among the threads on a process. Given that there
are typically relatively few chunks assigned to each thread, and that
the cost of the chunks can be estimated pretty well and assigned
appropriately among the threads (if we know the schedule), this is
unlikely to hurt performance. Experiments indicate that performance
improves with this change (in conjunction with the rest of the PR).

As part of the optimization to the assignment of chunks to threads,
the algorithm to calculate the number of chunks to processes is
modified to calculate the number of chunks to assign per thread, and
then use this to determine the number of chunks to assign to each
process. This supports future heterogenous systems where not all
processes have the same number of threads, though it still assumes
that all threads are equally capable.

To document the load balance and associated performance, performance
per process over the chunk loops in physpkg.F90 is collected and
compared to the cost per chunk (which was already collected). The
output of print_cost_p is modified to include these new data. The
output of print_cost_p is also now archived with the other performance
data at the end of the run.

Finally, there is currently a check whether a runtime pcols
specification (phys_chnk_fdim) disagrees with a compile-time
specification. If so, a warning message is output and the model
continues running using the compile-time specification. This is even
triggered by the default setting of phys_chnk_fdim, which means that
the warning message is always output when pcols is specified at
compile-time. This is modified so that a warning message is only
generated when the runtime pcols namelist variable (phys_chnk_fdim)
specifies a specific pcols that is not the same as the compile-time
set value. A value phys_chnk_fdim < 1 indicates that the model should
decide what value of pcols to use, and setting pcols at compile-time
implicitly defines what this default value is, so the two settings are
not in conflict.

BFB

### DIFF
--- a/components/cam/src/physics/cam/phys_grid.F90
+++ b/components/cam/src/physics/cam/phys_grid.F90
@@ -354,7 +354,7 @@ module phys_grid
    integer, private :: phys_alltoall = def_alltoall
 
 ! Physics computational cost
-   real(r8), public :: phys_pcost ! measured physics cost on this process
+   real(r8), public :: phys_proc_cost ! measured physics cost on this process
 
 contains
 !========================================================================
@@ -2443,32 +2443,39 @@ logical function phys_grid_initialized ()
    integer  :: ncols                  ! number of columns in chunk
    integer  :: ierr                   ! error return
 
-   real(r8) :: lsums(4)               ! working arrays for local sums
-   real(r8) :: gsums(4)               ! working arrays for global sums
-   real(r8) :: ccest_lsum, ccest_gsum ! local and global sums of 
-                                      !  estimated normalized chunk costs
-   real(r8) :: ccost_lsum, ccost_gsum ! local and global sums of 
-                                      !  measured chunk costs
-   real(r8) :: tcest(0:nlthreads-1)   ! estimated normalized thread cost
-                                      !  for chunk loops in bc_physics 
+   real(r8) :: local_sums(4)          ! working vector for local sums
+   real(r8) :: global_sums(4)         ! working vector for global sums
+   real(r8) :: chnk_estcost_lsum      ! local sum of estimated
+                                      !  chunk costs
+   real(r8) :: chnk_estcost_gsum      ! global sum of estimated
+                                      !  chunk costs
+   real(r8) :: chnk_cost_lsum         ! local sum of measured
+                                      !  chunk costs
+   real(r8) :: chnk_cost_gsum         ! global sum of measured
+                                      !  chunk costs
+   real(r8) :: thrd_estcost(0:nlthreads-1)
+                                      ! estimated thread cost for
+                                      !  chunk loops in bc_physics 
                                       !  and ac_physics
-   real(r8) :: pcest, pcest_sum       ! estimated normalized process cost
-                                      !  for chunk loops in bc_physics 
-                                      !  and ac_physics, and sum over all
+   real(r8) :: proc_estcost           ! estimated process cost for
+                                      !  chunk loops in bc_physics 
+                                      !  and ac_physics
+   real(r8) :: proc_estcost_sum       ! sum of proc_estcost over all
                                       !  processes
-   real(r8) :: pcost_sum              ! sum of over processes of
-                                      !  measured process cost for chunk 
-                                      !  loops in bc_physics and 
-                                      !  ac_physics (phys_pcost)
-   real(r8) :: avg_ccest              ! average estimated chunk cost
-   real(r8) :: avg_ccost              ! average measured chunk cost
-   real(r8) :: avg_pcest              ! average estimated process cost
-   real(r8) :: avg_pcost              ! average measured process cost
-   real(r8) :: ccost                  ! chunk cost
-   real(r8) :: norm_ccest             ! normalized estimated chunk cost
-   real(r8) :: norm_ccost             ! normalized chunk cost
-   real(r8) :: norm_pcest             ! normalized estimated process cost
-   real(r8) :: norm_pcost             ! normalized process cost
+   real(r8) :: proc_cost_sum          ! sum over processes of measured
+                                      !  process cost for chunk loops
+                                      !  in bc_physics and ac_physics
+                                      !  (phys_proc_cost)
+   real(r8) :: avg_chnk_estcost       ! average estimated chunk cost
+   real(r8) :: avg_chnk_cost          ! average measured chunk cost
+   real(r8) :: avg_proc_estcost       ! average estimated process cost
+   real(r8) :: avg_proc_cost          ! average measured process cost
+   real(r8) :: chnk_estcost           ! chunk estimated cost
+   real(r8) :: chnk_cost              ! chunk cost
+   real(r8) :: norm_chnk_estcost      ! normalized estimated chunk cost
+   real(r8) :: norm_chnk_cost         ! normalized chunk cost
+   real(r8) :: norm_proc_estcost      ! normalized estimated process cost
+   real(r8) :: norm_proc_cost         ! normalized process cost
    real(r8) :: pspeedup               ! local speed-up relative to sum of
                                       !  local chunk cost, so a measure of
                                       !  load balance, but not traditional
@@ -2486,59 +2493,47 @@ logical function phys_grid_initialized ()
       ! the chunks assigned to each thread, assuming a static round-robin
       ! distribution. Estimated process cost is then calculated as the
       ! maximum over the estimated thread costs.
-      ccest_lsum = 0.0_r8
-      ccost_lsum = 0.0_r8
-      tcest(:)   = 0.0_r8
+      chnk_estcost_lsum = 0.0_r8
+      chnk_cost_lsum    = 0.0_r8
+      thrd_estcost(:)   = 0.0_r8
       do lcid = begchunk, endchunk
-         ccest_lsum = ccest_lsum + chunks(lchunks(lcid)%cid)%estcost
-         ccost_lsum = ccost_lsum + lchunks(lcid)%cost
-         tid = mod(lcid-begchunk,nlthreads)
-         tcest(tid) = tcest(tid) + chunks(lchunks(lcid)%cid)%estcost
+         chnk_estcost_lsum = chnk_estcost_lsum + chunks(lchunks(lcid)%cid)%estcost
+         chnk_cost_lsum    = chnk_cost_lsum + lchunks(lcid)%cost
+         tid               = mod(lcid-begchunk,nlthreads)
+         thrd_estcost(tid) = thrd_estcost(tid) + chunks(lchunks(lcid)%cid)%estcost
       enddo
-      pcest = maxval(tcest)
+      proc_estcost = maxval(thrd_estcost)
 
 #if ( defined SPMD )
-      gsums(:) = 0.0_r8
-      lsums(1) = ccest_lsum
-      lsums(2) = ccost_lsum
-      lsums(3) = pcest
-      lsums(4) = phys_pcost
-      call MPI_ALLREDUCE(lsums, gsums, 4, MPI_REAL8, MPI_SUM, &
+      global_sums(:) = 0.0_r8
+      local_sums(1)  = chnk_estcost_lsum
+      local_sums(2)  = chnk_cost_lsum
+      local_sums(3)  = proc_estcost
+      local_sums(4)  = phys_proc_cost
+      call MPI_ALLREDUCE(local_sums, global_sums, 4, MPI_REAL8, MPI_SUM, &
                          mpicom, ierr)
-      ccest_gsum = gsums(1)
-      ccost_gsum = gsums(2)
-      pcest_sum  = gsums(3)
-      pcost_sum  = gsums(4)
+      chnk_estcost_gsum = global_sums(1)
+      chnk_cost_gsum    = global_sums(2)
+      proc_estcost_sum  = global_sums(3)
+      proc_cost_sum     = global_sums(4)
 #else
-      ccost_gsum = ccost_lsum
-      ccest_gsum = ccest_lsum
-      pcost_sum  = phys_pcost
-      pcest_sum  = pcest
+      chnk_cost_gsum    = chnk_cost_lsum
+      chnk_estcost_gsum = chnk_estcost_lsum
+      proc_cost_sum     = phys_proc_cost
+      proc_estcost_sum  = proc_estcost
 #endif
 
       ! Calculate average estimated chunk cost
-      avg_ccest = ccest_gsum/real(nchunks,r8)
-      if (avg_ccest < 0.000001_r8) then
-         avg_ccest = -1.0_r8
-      endif
+      avg_chnk_estcost = chnk_estcost_gsum/real(nchunks,r8)
 
       ! Calculate average measured chunk cost
-      avg_ccost = ccost_gsum/real(nchunks,r8)
-      if (avg_ccost < 0.000001_r8) then
-         avg_ccost = -1.0_r8
-      endif
+      avg_chnk_cost = chnk_cost_gsum/real(nchunks,r8)
 
       ! Calculate average estimated process cost
-      avg_pcest = pcest_sum/real(npes,r8)
-      if (avg_pcest < 0.000001_r8) then
-         avg_pcest = -1.0_r8
-      endif
+      avg_proc_estcost = proc_estcost_sum/real(npes,r8)
 
       ! Calculate average measured process cost
-      avg_pcost = pcost_sum/real(npes,r8)
-      if (avg_pcost < 0.000001_r8) then
-         avg_pcost = -1.0_r8
-      endif
+      avg_proc_cost = proc_cost_sum/real(npes,r8)
 
       ! Take turns writing to fname.
       if (iam == 0) then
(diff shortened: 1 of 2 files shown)


worleyph:
The earlier version of this development was tested extensively on both Compy and Cori-KNL. This was re-implemented on current master a few days ago and tested only on Compy. There are no differences in the primary algorithms between the two implementations.

worleyph:
@jgfouca , adding you as a reviewer as I added a few lines to provenance.py. I mimicked other code to get the effect that I wanted, and it worked, but please verify that what I did is acceptable.

worleyph:
@sarats , adding you as a reviewer as primarily an advertisement of the new data being archived. Don't know if this will be directly useful to you, but wanted you to know about it.

worleyph:
@whannah1 and @ambrad , I have very much appreciated your reviews of the previous two PRs. This is the last one in the "trilogy" looking at improving load balancing, especially when column weights are provided (though this one is useful even when all weights == 1 but the number of columns is not an integer multiple of the number of chunks). I am not going to assign this to either of you a priori, but if you want I can add you (or you can add yourself).

Note there there will likely be a 4th PR in the series, looking to implement the refactoring to increase modularity requested by the two of you in the previous PR, but I have not even begun to think about that, so it is unlikely to show up for awhile.

worleyph:
@ambrad and @whannah1 , you might also want to look at the new chunk/process load balancing data being output, to see if it would be useful in your work.

worleyph:
@amametjanov , adding you as a reviewer to get your opinion on specifying the threading schedule in physpkg.F90 . Didn't really feel that I had an option, and in a few performance studies it has been an advantage.

components/cam/src/physics/cam/physpkg.F90
```diff
@@ -969,8 +969,10 @@ subroutine phys_run1(phys_state, ztodt, phys_tend, pbuf2d,  cam_in, cam_out)
 #if (! defined SPMD)
     integer :: mpicom = 0
 #endif
-    integer(i8) :: beg_count                     ! start time for a chunk
-    integer(i8) :: end_count                     ! stop time for a chunk
+    integer(i8) :: beg_pcount                    ! start time for the process
+    integer(i8) :: end_pcount                    ! stop time for the process
+    integer(i8) :: beg_ccount                    ! start time for a chunk
+    integer(i8) :: end_ccount                    ! stop time for a chunk
```
whannah1, line 975: I think more verbose names would be good here, like "beg_chnk_cnt" and "beg_proc_cnt". When I was looking at the diffs below my eyes were initially missing where the "c" or "p" was added.

components/cam/src/physics/cam/phys_grid.F90
```diff
@@ -353,6 +353,9 @@ module phys_grid
    integer, private, parameter :: def_alltoall = 1                ! default
    integer, private :: phys_alltoall = def_alltoall
 
+! Physics computational cost
+   real(r8), public :: phys_pcost ! measured physics cost on this process
```
whannah1, line 357: This is the "pre-measured" cost, right? 
Might be good to clarify that it's not currently being measured.

components/cam/src/physics/cam/phys_grid.F90
```diff
@@ -2433,52 +2437,107 @@ subroutine print_cost_p()
...
+                                      !  for chunk loops in bc_physics 
+                                      !  and ac_physics
+   real(r8) :: pcest, pcest_sum       ! estimated normalized process cost
+                                      !  for chunk loops in bc_physics 
+                                      !  and ac_physics, and sum over all
+                                      !  processes
+   real(r8) :: pcost_sum              ! sum of over processes of
+                                      !  measured process cost for chunk 
+                                      !  loops in bc_physics and 
+                                      !  ac_physics (phys_pcost)
```
whannah1, line 2462: Thanks for the nicely formatted descriptions! The wording for this one seems a bit weird though, is it right?

components/cam/src/physics/cam/phys_grid.F90
```diff
@@ -2433,52 +2437,107 @@ subroutine print_cost_p()
...
+   real(r8) :: lsums(4)               ! working arrays for local sums
+   real(r8) :: gsums(4)               ! working arrays for global sums
+   real(r8) :: ccest_lsum, ccest_gsum ! local and global sums of 
+                                      !  estimated normalized chunk costs
+   real(r8) :: ccost_lsum, ccost_gsum ! local and global sums of 
+                                      !  measured chunk costs
+   real(r8) :: tcest(0:nlthreads-1)   ! estimated normalized thread cost
+                                      !  for chunk loops in bc_physics 
+                                      !  and ac_physics
+   real(r8) :: pcest, pcest_sum       ! estimated normalized process cost
```
whannah1, line 2455: Similar to another suggestion, I would personally prefer more verbose variable names like "proc_cost_est" instead of "pcest", but given the nice descriptions in the variable declarations it's not hard to figure things out, so it's not critical.

ambrad:
It looks like this PR changes default behavior in, e.g., a 16x8 MPI x thread run on KNL. Is that true? If so, do you have performance data associated with these changes? You could use the best case Noel finds for RRM on Cori-KNL and then run that case with these changes.

components/cam/src/physics/cam/phys_grid.F90
```diff
@@ -2433,52 +2437,107 @@ subroutine print_cost_p()
...
-      if (abs(cost_gsum) > 0.000001_r8) then
-         avg_cost = cost_gsum/nchunks
-      else
-         avg_cost = 1.0_r8
+
+      ! Calculate average estimated chunk cost
+      avg_ccest = ccest_gsum/real(nchunks,r8)
+      if (avg_ccest < 0.000001_r8) then
+         avg_ccest = -1.0_r8
       endif
```
whannah1, line 2523: What situation is this checking for? 
All local chunk costs were zero?

components/cam/src/physics/cam/phys_grid.F90
```diff
@@ -2433,52 +2437,107 @@ subroutine print_cost_p()
...
+
+      ! Calculate average estimated process cost
+      avg_pcest = pcest_sum/real(npes,r8)
+      if (avg_pcest < 0.000001_r8) then
+         avg_pcest = -1.0_r8
+      endif
+
+      ! Calculate average measured process cost
+      avg_pcost = pcost_sum/real(npes,r8)
+      if (avg_pcost < 0.000001_r8) then
```
whannah1, line 2539: In the spirit of not having "magic numbers" in the code, maybe this threshold value should be assigned to a parameter.

components/cam/src/physics/cam/phys_grid.F90
worleyph: No - this is the experimentally measured data. See use in physpkg.F90

worleyph:
@ambrad, the threading specification is a change of behavior (or could be - going with the default for a system doesn't tell you much). The assignment algorithm uses the fact that the assignment of chunks to threads is known in order to decide how to load balance the assignment between threads. Without fixing this to some schedule, this optimization is not possible, and the wrap map is more convenient than a block map when implementing this approach. Based on the granularity of the "chunk" work and the relatively small number of chunks per thread (compared to particle codes, for example), I can't see there being any locality-preservation-based advantage to using a block map. f we have a good estimate of cost (columns essentially equal cost or user-provided estimates), then the static schedule with explicit control of assignment should be better than dynamic, for lower overhead if for no other reason. We could try fixing the schedule to dynamic and compare against this as well.

I did run a fewer experiments looking at performance, and the wrap map with the other changes was superior compared to the current default (on Cori-KNL and Compy), but this was not exhaustive. If you have a favorite test, I can look at that as well. Note that it pcols generates one chunk per thread will make this irrelevant.

The new output specifically looks at load balance within a process (among other metrics), and looks pretty good in all of my experiments.

components/cam/src/physics/cam/physpkg.F90
worleyph: Will make the change soon. Thanks for the suggestion.

components/cam/src/physics/cam/phys_grid.F90
worleyph: Yes - something got mangled. I'll add a fix when I make the requested change to variable names.

components/cam/src/physics/cam/phys_grid.F90
worleyph: I'll revisit these names as well, and hopefully come up with something more self-documenting. Thanks.

ambrad:
In case you want to do a comparison performance study on configurations getting a lot of attention at present, they are these:
- ne30pg2 F-case on 30 Compy nodes,
- ne30pg2 coupled on 90 Compy nodes (can also study this with F-case on 68 Compy nodes),
- and the RRM case Noel is studying.

components/cam/src/physics/cam/phys_grid.F90
worleyph: I hate for code to abort for performance documentation reasons (I prefer erroneous performance data to killing an otherwise working simulation) so put checks like these in to prevent divide-by-zero even if there is not reason for this to ever occur. In this case, perhaps a user specificed that all columns had no cost, in which case the sum was zero? 

The use of the magic number was sloppiness/laziness on my part. There is nothing special about the number, and it should be relative to the numerator. I could just replace it by /= 0.0_r8, to eliminate the one known source of this error, and assume that others simply will never occur? Advice? If we stick to a bound away from a small number, I'll rework this to be more "precise", i.e. < max_ccest*epsilon, or something like this.

worleyph:
@ambrad , I can document the Compy performance this weekend (are there existing cases that I can peek at, to make sure that I get them right?).

 Cori-KNL is down until next Wednesday, I am travelling starting Wednesday evening through the following Tuesday and may not be able to fire up a test until I get back. When I can, does the RRM case require a specific version of the model? Or can I just use this PR branch (based off of master as of two days ago)? Where would I look at the case specification? (Can ask Noel as well, once I can get back on Cori-KNL.)

ambrad:
@worleyph I'll send you an email with info.

worleyph:
@ambrad - thanks. I'll take a look. FYI - realized that my above comment was a little strange. This PR was motivated by performance issues, both real world and contrived (assuming some user-provided estimated cost), and it does address those quite adequately. I am happy to run additional validation performance experiments using current production scenarios just to be sure that there are not any unintended consequences.

worleyph:
@ambrad, the Compy examples do not use threading, so the change in thread scheduling will not be evaluated. The PE layout you pointed me at and the pcols setting (==18) generates one chunk per process, so the PR branch will behave identical to the unmodified code (though chunk assignment to processes may differ, but this will only affect communication overhead in d_p_coupling and p_d_coupling in some random way depending on what nodes are allocated, and no more than the differences between two runs with different node allocations).

Should I modify this to introduce threading or change the PE layout or pcols to force a scenario where the new code will have different behavior from the unchnaged code? (Even with threading, I'd need to have more than one chunk assigned to a thread to see some substantive difference.) 

Thanks.

ambrad:
Great. So the Compy cases by inspection will not be affected. That leaves the Cori-KNL cases to consider. If pcols=4 with 16x8 on 113 ends up being the final config, it may be that this PR will not affect that case either. If that's true, then I agree that we can conclude by simple inspection that this PR won't harm performance on production cases of interest.

worleyph:
@jgfouca , just checking, the file that is beingn renamed (adding $lid as a suffix) and then copied to the performance archive is not guaranteed to exist. The user can disable its creation by setting a namelist variable in user_nl_cam. Will its nonexistence cause any problems to this code in CIME?

components/cam/src/physics/cam/physpkg.F90
worleyph: @whannah1 , done. Please take a look.

components/cam/src/physics/cam/phys_grid.F90
worleyph: Done,

components/cam/src/physics/cam/phys_grid.F90
worleyph: @whannah1 , done. please take a look. I also made similar changes in create_chunks, pheap_adjust (now proc_heap_adjust) and cheap_adjust (now chnk_heap_adjust).

components/cam/src/physics/cam/phys_grid.F90
worleyph: Reworked  this. Please take a look - google searched for best practice logic to avoid INF and NaN and implemented this - no more magic numbers (and also more robust).

jgfouca:
@worleyph , I assume you mean atm_chunk_costs.txt? If it might not exist, you should protect the code that tries to move it with an existence check:

```
    atm_chunk_costs_src_path = os.path.join(rundir, "atm_chunk_costs.txt")
    if os.path.exists(atm_chunk_costs_src_path):
        atm_chunk_costs_dst_path = os.path.join(rundir, "atm_chunk_costs.{}".format(lid))
        shutil.move(os.path.join(rundir, "atm_chunk_costs.txt"), atm_chunk_costs_path)
        gzip_existing_file(atm_chunk_costs_path)
```

worleyph:
@jgfouca , what about the globs_to_copy_append request?

         globs_to_copy.append(os.path.join(rundir, "atm_chunk_costs.{}.gz".format(lid)))

Should this instead be:

         atm_chunk_costs_dst_gz_path = os.path.join(rundir, "atm_chunk_costs.{}.gz".format(lid)))
         if os.path.exists(atm_chunk_costs_dst_gz_path):
             globs_to_copy.append(atm_chunk_costs_dst_gz_path)

?

Thanks.

jgfouca:
@worleyph , an empty glob match is not an error, so it should be fine.

worleyph:
@jgfouca , modified provenance.py as per your instructions. Worked as expected. If you could just glance at it one more time, I'd appreciate it. Thanks.

jgfouca:
@worleyph , looks good.

components/cam/src/physics/cam/phys_grid.F90
whannah1: I think that's what I meant, that this is an "input" cost rather than an "output" cost

worleyph: This is an 'output" cost, if I am using your terminology correctly. It is measured during the course of the model run, and the experimentally determined value is output at the end in atm_chunk_costs.txt . It is not set by the user or the model before the run or via input.

components/cam/src/physics/cam/phys_grid.F90
whannah1: Lokks good.

components/cam/src/physics/cam/phys_grid.F90
whannah1: ah ok, I assumed it was the opposite.

worleyph:
@amametjanov , ran some sanity checks on both Compy and Cori-KNL with single and multiple chunks per thread when chunks all have the same number of columns, and do not see any performance degradation from specifying a (static,1) schedule. (This situation is where the optimizations in this PR should have no effect, so all it is really measuring is whether specifying this schedule is faster or slower than the default). Earlier experiments indicated that these modifications demonstrated a performance advantage when not all chunks had the same number of columns.

Please review and indicate whether you approve or not. Thanks.

worleyph:
I identified another tweak that may be able to further improve performance, by improving the communication overhead optimization while still preserving the improved load balancing when assigning chunks to processes. Please do not integrate the PR until I have examined this (4 line) modification. I will push an update if it proves useful.

worleyph:
@whannah1 , @amametjanov , @ambrad , and @sarats , the "tweak" is in. It was more involved than I anticipated, and I added another diagnostic (see the commit message) to document what is going on. All algorithm changes are in the section where chunks are assigned to processes. As the number of chunks assigned to each process is calculated beforehand, nothing serious can go wrong. I did test that it is BFB and that the resulting assignment is at least as good with respect to load balancing and is better with respect to minimizing the number of columns that need to be communicated during dynamics/physics coupling. Have not tested exhaustively (across all phys_loadbalance options), but did test phys_loadbalance=2 (the default), =-1, and =0 . Testing was performed on both Compy and Cori-KNL, with threading and with both 1 chunk per thread, with 2 chunks per thread, and with a large number of chunks per thread.

worleyph:
Upon review this morning, I'm not comfortable with one of the changes in the last commit, though it seems to be working fine. I'll continue working on this for a few more days. I am offline for most of the next 7 days, so may not finish this until the end of next week, but will update here if there are any changes in this timeline.

worleyph:
I resolved the issue that I was concerned with (had the right idea yesterday - just implemented in a deficient way). This should be the last commit, but I'll exercise the PR carefully over the next week to verify that it has the desired performance characteristics.

worleyph:
@singhbalwinder , I believe that this is ready to go. I see that there is a conflict in provenance.py, and the "resolve conflicts" button is greyed out with a message that the conflicts are too complex to be displayed. If you look at my changes, they are just the addition of some logic (to rename/add a suffix to the file atm_chunk_costs.txt, then gzip it, and then to include this file with the other provenance data in the performance archive), for a total of 6 new lines in one location (one of which is blank) and one new line in a different location.

singhbalwinder:
Thanks @worleyph ! Since this is a BFB PR, I will merge it today after resolving the conflict.

rljacob:
provenance.py is inside cime/scripts which is now  a submodule.  So that piece will have be put in as a separate PR to ESMCI/cime.   Might be best to rebase this.  Then this PR can include the update to the cime submodule that includes the changes to provenance.py.

singhbalwinder:
Thanks @rljacob. I didn't realize that this file is now under CIME. @worleyph : Please let me know if you need any help with rebasing this PR.

worleyph:
So we can no longer make changes in E3SM that touch CIME, have it go into effect immediately, and then be merged into the CIME preperty then next time there is a sync? Or are changes in CIME proper available immediately now?

worleyph:
@singhbalwinder , I have no idea what I should do. What I can do is just start over with a new PR using the latest master, and hope that I can get it in in time to avoid having to repeat the work. I also prefer that someone else submit a PR to ESMI/cime.

singhbalwinder:
Rebasing this PR should be easy. I will do that today. If you can test it after the rebase to ensure that everything is okay, that will be great. I can submit a PR to ESMCI as well.

rljacob:
Right you can no longer make changes in E3SM that touch CIME.  All the config files were moved out of cime but you are modifying the python.  Its now easier to bring in latest CIME since Jim doesn't have to do a split/merge cycle with subtree but it still requires a PR to update the cime submodule hash.

worleyph:
@singhbalwinder , thank you. I'll wait for you to tell me when to start testing, and please do submit a CIME PR for me.

worleyph:
@rljacob , would it be possible in the future to split out the e3sm-specific python routines from CIME and put them into E3SM? provenance.py is the primary source (and the one I care about), but I see e3sm-specific code in case_run.py and in e3sm_cime_mgmt.py and a few other places as well.

rljacob:
@jgfouca would have to answer if its possible.

jgfouca:
@worleyph , yes, it should be possible and would be a good longterm goal for CIME to support that. I'll make an issue.

worleyph:
@jgfouca , Thanks! I was not hopeful, but if you don't ask ... so I am glad that I asked.

singhbalwinder:
@worleyph : I have pushed the rebased branch. This should all the changes except for the changes in provenance.py.

singhbalwinder:
Pushed to next. I will create a PR for changes in provenance.py under ESMCI/CIME.

PR #3691, opened 2020-07-10
Adds capability to read aircraft and surface emissions for prognostic CO2

beharrop:
Adds capability to run emissions-forced simulations by reading in CO2 emissions, both surface and aircraft.  Conservation of emissions is handled by offline regridding with a conservative remapping algorithm.  Both surface and aircraft emissions are added to the total CO2 tracer as well as the CO2 from fossil fuel flux tracer (CO2_FFF).  These capabilities are intended for the v2 BGC campaign.  Code changes have no impact on simulations when emissions datasets are not specified in the atmosphere namelist.

[BFB]


rljacob:
Note from IG telecon:  @singhbalwinder and @wlin7 think this is ready.

rljacob:
Note from IG telecon:  this needs a rebase.

beharrop:
Thanks for the update @rljacob .  @susburrows has informed that it we will need this merged in before the v2 code freeze as it is necessary for our BGC simulations.  @singhbalwinder is there anything we can do to get this moving forward?

components/cam/src/chemistry/utils/aircraft_emit.F90
```diff
@@ -1,95 +1,175 @@
...
+  logical, parameter  :: horz_native(N_AERO) = (/.false., .false., .false., .false., .false., &
+       .false., .true., .false., .false.,.false./)
+
+  !Logical to control whether to convert units to mmr (mass mixing ratios) or not
+  !sometimes unit coversion is handled at other places in the code
+  ! (e.g. unit conversion of ac_CO2 is handled in co2_cycle.F90)
+  logical, parameter  :: convert_to_mmr(N_AERO) = (/.true., .true., .true., .true., .true., &
+       .true., .false., .true., .true.,.true./)
+
+  !mixtyoe is only used for advected tracers
```
singhbalwinder, line 131: typo

singhbalwinder (approved):
This looks good to me. I will fix a typo and remove a line I added as a reminder to myself.

singhbalwinder:
I have rebased this branch for cam->eam change and this branch is now up-to-date with the current master. 

@beharrop : If you have time, would you please run your tests to ensure that the new branch is producing the expected results. I will run some tests with different compiler/machines to ensure that it is not breaking any tests.

Since Bryce and I were involved in writing code in this branch, I think we should select another reviewer to review the changes.

beharrop:
Thanks for doing the rebase, @singhbalwinder.  I'll run some tests with the rebase and get back to you with my assessment.

beharrop:
@singhbalwinder I was able to successfully run tests reading in both surface and aircraft emissions for a historical and scenario simulation (both done with an ne4pg2 atmosphere grid).  Everything ran smoothly and the results are consistent with previous tests I have run, so I think from a science standpoint the rebase went smoothly.

rljacob:
@susburrows can you approve this?

susburrows (approved):
I didn't do a line-by-line review of detailed testing, but I've run these code changes merged in a different code base, and have visually inspected the code changes here on Github.  I can verify that they appear to be reasonable and appear to achieve the desired result.  Furthermore, they will only affect simulations with active atmospheric CO2.  Given Bryce's and Balwinder's more detailed testing, as long as these changes are passing integrating testing, I think we can safely merge them.

beharrop:
I was looking over these again, I am not sure why the changes to vertical_diffusion.F90 are here.  I think they may be a problem from rebasing.  I don't think we want to get rid of the shape preserving changes we made in PR #3003.  These changes have nothing to do with the emissions anyway.  @singhbalwinder, what is the easiest way to remove the changes from that particular file?

singhbalwinder:
I think it might have gotten mixed up during my rebase as this branch was older than #3003's branch before rebase. Thanks for taking another look at it!

I will fix it but is this file used anymore? Please correct me if I am wrong but I thought all new F compsets now use clubb_surface rather than vertical_diffusion.F90.

beharrop:
My understanding is that vertical_diffusion.F90 is not being used, but I am not entirely sure.  I also don't know what long-term plans, if any, there might be for those routines.

whannah1:
@beharrop you're right that the vertical diffusion is bypassed when using CLUBB, but I've always been a little concerned with that approach in regards to how friction is applied to the atmosphere. The basic problem is that by skipping the vertical diffusion you also skip applying certain fluxes until tphysbc, and for friction this means that the dycor always sees friction at a delay. I talked with Peter Bogenshutz about this and he never really considered it to be a problem when he originally did this in CAM. I've tested some things in the MMF and it changes the solution considerably and makes it less stable. So I think this something that should be looked at in more detail.

beharrop:
Thanks @whannah1 for clarifying.  I was thinking about this just from a tracer transport issue, but that is an interesting point on the friction.  If this coupling gets modified in the future, I'd appreciate getting pinged just to make sure all the CO2 coupling still works as expected.

wlin7:
Merged to next.

wlin7:
Fixed the call to aircraft_emit_init in crm/physpkg.F90 and remerged to next.

wlin7:
There are other changes made to cam/physpkg.F90 need to also be added to crm/physpkg.F90. Done (
32c3734) and re-merged to next.

wlin7:
Merged to master.

PR #3693, opened 2020-07-11
standalone HOMME cmake updates for gcc10 and latest CIME

mt5555:
the gfortran with gcc10 has much stricture type checking and no longer allows  logicals to be mixed with integers.  
 
Add option to standalone HOMME builds to allow relax this type checking.  

Added cmake necessary to build cprnc from latest version of CIME

Add HOMME machine file for "mappy" new machine replacing melvin


oksanaguba:
in next and master

PR #3694, opened 2020-07-12
[WIP] Add a new data type to transfer state variables and fluxes between ru…

changliao1025:
Add a new data type to transfer state variables and fluxes between runoff and land components.

Previously, there were no land-to-river and river-2-land data types. In the recent development, we decided to add them explicitly for land and river coupling.

[BFB]  Currently, this data type is not actually used directly by any algorithms.

NL-6  This is part of the coupling effort between land and river model development.

### DIFF
--- a/components/clm/src/main/rof2lndType.F90
+++ b/components/clm/src/main/rof2lndType.F90
@@ -1,0 +1,253 @@
+
+!50==================================================
+ ! Author: Chang Liao( changliao at pnnl.gov )
+ ! Module: H2SC (hillslope based soil column drainage function)
+ ! rof->lnd exchange
+ ! First edit: 20180530
+ ! Revison: 20200626 Chang Liao
+ !50==================================================
+module rof2lndType
+
+    !-----------------------------------------------------------------------
+    
+    !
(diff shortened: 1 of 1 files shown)


rljacob:
State and flux variables go through the coupler for land and runoff models.  What does this do?

changliao1025:
Before the coupler can transfer state and flux variables, a data type must be used to store all the information.
Earlier, when land and river components exchange state and flux, they were using the atmosphere to land data type (atm2lnd_type) to get around, which was not the ideal solution at all and it does cause lots of confusion.
In the new development, we plan to add both rof2lnd and lnd2rof as two independent data types to hold this information.

bishtgautam:
@changliao1025 You also need to modify ELM code to use the new `rof2lndType`. Also, I believe the description of this PR is not correct. This new data type stores values about the river component that is sent by the coupler.

rljacob:
@changliao1025 please make the requested changes for this PR to advance.

bishtgautam:
After a discussion with @changliao1025, this PR is being closed for now.

PR #3695, opened 2020-07-13
update mpas-source: Fixes CVMix surface layer averaging

mark-petersen:
See https://github.com/MPAS-Dev/MPAS-Model/pull/608

Currently when cvmix computes averages for the surface layer (needed for
finding boundary layer depth) it simply does an arithmetic mean. This
is reasonable for the 60 layer mesh (with near constant 10m resolution
through 200m), but is not correct for stretched grids. Here the averaging is 
changed to a layerThickness weighted average.

[non-BFB]


vanroekel:
code changes look right and the 5 year G-case test looked good.  Approving by testing and visual inspection.

mark-petersen:
Running standard tests. Passes:
```
PET_Ln9.T62_oQU240.GMPAS-IAF.cori-knl_intel
PET_Ln9.T62_oQU240.GMPAS-IAF.cori-knl_gnu
PEM_Ln9.T62_oQU240.GMPAS-IAF.cori-haswell_gnu
PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.cori-haswell_intel
PEM_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_intel
```
The run failed on:
```
PET_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_gnu
```
but that looks like an error in the atmosphere initialization. The ocean and sea ice did not even start yet. It also only occurs with gnu, not intel. It does not appear to be related to this PR.

mark-petersen:
I confirmed that 
```
PET_Ln9_P1024.ne30_oECv3_ICG.A_WCYCL1850S.cori-knl_gnu
```
fails on today's `master` the same way as this PR, during atmospheric init. 

pinging @ndkeen, FYI. That same test does pass on cori-knl with intel.

jonbob:
merged to next

jonbob:

PR #3697, opened 2020-07-14
Remove CARMA

whannah1:
This removes all code and support for the Community Aerosol and Radiation Model for Atmospheres (CARMA). While CARMA has some advantages, there does not seem to be any interest in using or supporting it, so it was decided that it should be removed. This decision was discussed via confluence decision log.

[BFB]


whannah1:
I'm currently running the e3sm_atm_developer suite on Cori.

whannah1:
All e3sm_atm_developer tests passed, except for the expected NLFAIL from removing the CARMA namelist options. 

I tried to run e3sm_integration, but I got several failures just from generating the baselines, which I now understand is because many of those tests are not regularly run on Cori. 

So let me know if there's any other tests that I should run for this.

whannah1:
Reran the e3sm_atm_developer test suite on Cori after rebasing to include the recent CIME submodule change. Everything passes as before with the expected NLCOMP fails.

rljacob:
Were there CARMA options in the namelist even if it wasn't being used?

whannah1:
@rljacob that's correct. Variables like "carma_model" would be set to "none" by default. There was a whole "carma_nl" section .

whannah1:
Link to the decision log on Confluence:
This decision was discussed via confluence decision log:
https://acme-climate.atlassian.net/wiki/spaces/EIDMG/pages/1565884547/Should+we+remove+CARMA

whannah1:
@AaronDonahue I know this is a big PR, but most of it can be ignored since we're just deleting unused files like the stuff in the physics/carma folder.

AaronDonahue:
@whannah1 , thanks for the heads up.  I noticed that it was mostly just deleting stuff.  I will try to get to this today or tomorrow.

rljacob:
@AaronDonahue I mostly just wanted you to check it was removed correctly from the physics driver.

components/cam/test/system/input_tests_master
```diff
@@ -111,7 +111,6 @@ sm340 TSM.sh f4c5dm fv2d_8tsk 9s
 sm351 TSM.sh f0.9c4m outfrq24h+1850-2005_cam4 2d
 sm352 TSM.sh f0.9c4m outfrq24h+1850_cam4 2d
 
-sm360 TSM.sh f1.9c5carmdusdm outfrq3s_carma 9s
```
AaronDonahue, line 114: git review won't let me tag a line not impacted (At least I don't know how to), but if you go up to line 54 in the new version of this file you have `cb460 TCB.sh f1.9c4carmsuldh`, didn't you delete this test?  It might be good to grep for the string "carm" in this file, I think that will catch all the straggler lines that reference the above deleted tests but don't mention carma outright.

whannah1: Good catch! I still don't understand what anything in this part of the code is for, so I wasn't quite sure what to look for.

whannah1:
Latest commit addresses Aaron's comment. I grepped around for "carm" and it seems that I didn't miss anything else.

rljacob:
merged to next.

PR #3698, opened 2020-07-14
Replace 'sqs -f' call with 'scontrol show jobid' on NERSC systems

### LINKED ISSUE #3699: '-f' is no longer a valid option for 'sqs' command on NERSC, causing case.submit to fail
A recent update to NERSC system software removed the '-f' option
from the sqs command. 'sqs -f' is called from provenance.py for
both cori-knl and cori-haswell, causing job submission to fail
on those systems. 'sqs -f' is also called within the job progress
monitoring scripts syslog.cori-knl and syslog.cori-haswell,
which will cause these scripts to abort (though this will not affect
a running model).

worleyph:
A recent update to NERSC system software removed the '-f' option
from the sqs command. 'sqs -f' is called from provenance.py for
both cori-knl and cori-haswell, causing job submission to fail
on those systems. 'sqs -f' is also called within the job progress
monitoring scripts syslog.cori-knl and syslog.cori-haswell,
which will cause these scripts to abort (though this will not affect
a running model). 'sqs -f' is equivalent to 'scontrol show jobid',
so here 'sqs -f' is replaced by 'scontrol show jobid' in provenance.py
and in the job progress monitoring scripts. Note that the output
from 'scontrol show jobid' that is captured in provenance.py is
still saved in a file with the original name (sqsf_jobid.$lid),
because this name is looked for in the performance archiving
postprocessing scripts.

[BFB]

Fixes #3699

### DIFF
--- a/cime/scripts/lib/CIME/provenance.py
+++ b/cime/scripts/lib/CIME/provenance.py
@@ -170,7 +170,7 @@ def _save_prerun_timing_e3sm(case, lid):
                 run_cmd_no_fail(cmd, arg_stdout=filename, from_dir=full_timing_dir)
                 gzip_existing_file(os.path.join(full_timing_dir, filename))
         elif mach in ["cori-haswell", "cori-knl"]:
-            for cmd, filename in [("sinfo -a -l", "sinfol"), ("sqs -f %s" % job_id, "sqsf_jobid"),
+            for cmd, filename in [("sinfo -a -l", "sinfol"), ("scontrol show jobid %s" % job_id, "sqsf_jobid"),
                                   # ("sqs -f", "sqsf"),
                                   ("squeue -o '%.10i %.15P %.20j %.10u %.7a %.2t %.6D %.8C %.10M %.10l %.20S %.20V'", "squeuef"),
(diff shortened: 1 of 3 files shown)


worleyph:
@ndk, successfully ran a sanity check on Cori-KNL. The Cori-Haswell job won't run for a couple more hours (based on the sqs output).

worleyph:
@rljacob , this is a bug fix PR. Should I create a corresponding bug? I imagine that Noel will intregrate this almost immediately, as master won't run without these modifications (and the test suite has performance archiving disabled, so won't see these changes).

ndkeen:
Merged to next. Should I go ahead and merge to master?

worleyph:
cori-haswell test reported back - worked fine for me.

worleyph:
Oh, and a big big thanks @wlin7 and @ndk for identifying the workarounds (sqs.old and scontrol show jobid) so quickly.

worleyph:
@rljacob , should @ndkeen merge this to master?

rljacob:
Yes

rljacob:
@worleyph yes make a bug report.

PR #3703, opened 2020-07-15
Fix issue with sqs on Cori in maint-1.0

ndkeen:
After Cori maintenance, the version of slurm was updated which no longer supports the `-f` option to `sqs`
This command is used for job provenance.

This PR simply replaces "sqs -f" with "scontrol show jobid" in a few files.

[bfb]


ndkeen:
closing in favor of another

rljacob:
replaced by #3704

PR #3704, opened 2020-07-15
Fix the sqs command on Cori in maint-1.0 which affects job provenance

ndkeen:
After Cori maintenance, the version of slurm was updated.
This new version no longer supports "sqs -f" which we use for job provenance.
Changing the places where we use "sqs -f" to be "scontrol show jobid" should do the same thing.

Only affects Cori.

[bfb]


ndkeen:
Tested with `SMS_D_P64x2_Ln5.ne4_ne4.FC5AV1C-L`

ndkeen:
merged into maint-1.0

PR #3705, opened 2020-07-15
cmake: add much-improved support for removing flags and deoptimizing files

### LINKED ISSUE #3648: e3sm_remove_flags does not work in some practical applications
On Compy, adding the following to the Depends file (Depends.intel.pgi):

     set(TEST_FILE
       cam/src/physics/cam/phys_grid.F90)
     ...
       foreach(ITEM IN LISTS TEST_FILE)
         e3sm_remove_flags("${ITEM}" "-O2")
       endforeach()

does not remove -O2 from the compile line for phys_grid.F90, as reported in the e3sm.bldlog file (-O2 actually shows up twice, and neither is removed):

     ... -O2 -debug minimal  -O2    -free -DUSE_CONTIGUOUS=contiguous,  ...

If I instead also add -O1, i.e.

       foreach(ITEM IN LISTS TEST_FILE)
         e3sm_remove_flags("${ITEM}" "-O2")
         e3sm_add_flags("${ITEM}" "-O1")
       endforeach()

then a -O1 is added (so this works) but -O2 is still not removed:

     ... -O2 -debug minimal  -O2    -free -DUSE_CONTIGUOUS=contiguous,  -O1 ...

If I then try to remove the previously added -O1, i.e.

       foreach(ITEM IN LISTS TEST_FILE)
         e3sm_remove_flags("${ITEM}" "-O2")
         e3sm_add_flags("${ITEM}" "-O1")
         e3sm_remove_flags("${ITEM}" "-O1")
       endforeach()

this does remove the -O1:

     ... -O2 -debug minimal  -O2    -free -DUSE_CONTIGUOUS=contiguous,  ...

so remove does work in this context.

Being able to remove the -O2 specified earlier in the FFLAGS is important for resolving the issue raised in #3644 , and to isolate the reason that -O1 is required for reproducibility for PR #3620 when using the PGI compiler on Compy.

Note that the same result holds for the PGI compiler (where, for this same compset, -O2 shows up only once), but this issue should be independent of the compiler.

jgfouca:
Flags added via the global CMAKE_<LANG>_FLAGS cannot be easily removed from
individual files. This was impacting PGI users especially because -O0 does not
override previous -O2 flags, so the -O2 flag must be removed from the compile
command in order to deoptimize the compilation.

This PR switches from managing f90 flags via the global CMAKE_Fortran_FLAGS
to managing them through individual source file properties. This will allow
e3sm_remove_flags to work properly for all flags.

Also, remove some old blocks in config_compilers.xml.

Fixes #3648 

[BFB]

### DIFF
--- a/components/cmake/cmake_util.cmake
+++ b/components/cmake/cmake_util.cmake
@@ -80,6 +80,7 @@ function(e3sm_deoptimize_file FILE_ARG FFLAGS_NOOPT)
     # we have to remove the optimization flags first
 
     # Until we know which particular flags are related to optimization, we have to guess.
+    e3sm_remove_flags(${FILE_ARG} "-O1")
     e3sm_remove_flags(${FILE_ARG} "-O2")
     e3sm_remove_flags(${FILE_ARG} "-O3")
   endif()
--- a/components/cmake/cmake_util.cmake
+++ b/components/cmake/cmake_util.cmake
@@ -66,10 +66,22 @@ function(e3sm_remove_flags FILE_ARG FLAGS_ARG)
     if (NOT EXISTS ${PROJECT_SOURCE_DIR}/${FILE_ARG})
       message(FATAL_ERROR "Trying to set flags on non-existent source: ${FILE_ARG}")
     endif()
-    set(REAL_FILE "../../${FILE_ARG}")
+    set(REAL_FILE "${SOURCE_PATH}/${FILE_ARG}")
   endif()
 
   get_property(ITEM_FLAGS SOURCE ${REAL_FILE} PROPERTY COMPILE_FLAGS)
   string(REPLACE "${FLAGS_ARG}" "" ITEM_FLAGS "${ITEM_FLAGS}")
   set_property(SOURCE ${REAL_FILE} PROPERTY COMPILE_FLAGS "${ITEM_FLAGS}")
+endfunction()
+
+function(e3sm_deoptimize_file FILE_ARG FFLAGS_NOOPT)
+  if (CMAKE_Fortran_COMPILER_ID STREQUAL "PGI")
+    # PGI does not support bulk-disabling of optimization by appending -O0,
+    # we have to remove the optimization flags first
+
+    # Until we know which particular flags are related to optimization, we have to guess.
+    e3sm_remove_flags(${FILE_ARG} "-O2")
+    e3sm_remove_flags(${FILE_ARG} "-O3")
+  endif()
+  e3sm_add_flags(${FILE_ARG} "${FFLAGS_NOOPT}")
 endfunction()
--- a/cime_config/machines/config_compilers.xml
+++ b/cime_config/machines/config_compilers.xml
@@ -779,7 +779,7 @@ flags should be captured within MPAS CMake files.
     <base>-lstdc++</base>
   </CXX_LIBS>
   <SLIBS>
-    <append>$SHELL{$ENV{NETCDF_FORTRAN_PATH}/bin/nf-config --flibs} $SHELL{$ENV{NETCDF_C_PATH}/bin/nc-config --libs}</append> 
+    <append>$SHELL{$ENV{NETCDF_FORTRAN_PATH}/bin/nf-config --flibs} $SHELL{$ENV{NETCDF_C_PATH}/bin/nc-config --libs}</append>
     <append>-L$ENV{MKLROOT}/lib/intel64 -Wl,--no-as-needed -lmkl_gf_lp64 -lmkl_sequential -lmkl_core -lpthread -lm -ldl</append>
   </SLIBS>
   <NETCDF_C_PATH>$ENV{NETCDF_C_PATH}</NETCDF_C_PATH>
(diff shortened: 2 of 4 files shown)


components/cmake/cmake_util.cmake
```diff
@@ -66,10 +66,22 @@ function(e3sm_remove_flags FILE_ARG FLAGS_ARG)
...
+    set(REAL_FILE "${SOURCE_PATH}/${FILE_ARG}")
   endif()
 
   get_property(ITEM_FLAGS SOURCE ${REAL_FILE} PROPERTY COMPILE_FLAGS)
   string(REPLACE "${FLAGS_ARG}" "" ITEM_FLAGS "${ITEM_FLAGS}")
   set_property(SOURCE ${REAL_FILE} PROPERTY COMPILE_FLAGS "${ITEM_FLAGS}")
 endfunction()
+
+function(e3sm_deoptimize_file FILE_ARG FFLAGS_NOOPT)
+  if (CMAKE_Fortran_COMPILER_ID STREQUAL "PGI")
```
sarats, line 78: So, you are removing existing flags applies to PGI only as that's the troublemaker, correct? Rest of the compiler families would keep appending flags. Just something we should record for future reference.

jgfouca: Yes, see comment below.

jgfouca:
I should mention, one consequence of this PR is that the component.flags file will be much larger than before, since it will be listing flags per-file instead of per-component. One the one hand, this makes it a bit more tedious to do a temporary flag tweak for the entire component (you have to do a search/replace instead of a single edit). On the other hand, it makes it easy to do temporary flag tweaks to individual files.

jgfouca:
@worleyph , if you're back, can you verify this fix?

worleyph:
@jgfouca , just started looking at it now. Just verifying - this only affects PGI?

jgfouca:
@worleyph , no this effects everything.

sarats:
@jgfouca Hmm...when we talked earlier and above code too, it only impacts PGI.

jgfouca:
@sarats , @worleyph , sorry for the lack of clarity. The alternate behavior of e3sm_deoptimize_file (try to actually remove flags instead of just appending deoptimization flags) is specific to PGI.

The bigger change of managing flags on a per-file basis (which allows e3sm_remove_flags to work in all contexts) affects everything.

worleyph:
@jgfouca , I am testing e3sm_remove_flags. What is  e3sm_deoptimize_file, and why/when/where would it be used? Thanks.

components/cmake/cmake_util.cmake
```diff
@@ -66,10 +66,22 @@ function(e3sm_remove_flags FILE_ARG FLAGS_ARG)
...
+    # PGI does not support bulk-disabling of optimization by appending -O0,
+    # we have to remove the optimization flags first
+
+    # Until we know which particular flags are related to optimization, we have to guess.
+    e3sm_remove_flags(${FILE_ARG} "-O2")
+    e3sm_remove_flags(${FILE_ARG} "-O3")
```
amametjanov, line 84: Should remove `-O1` too?

jgfouca: Maybe, since CIME does not currently distinguish between optimization flags and other flags, I'm kind of guessing here.

jgfouca:
@worleyph , the intent of that function is to make it easy to remove compilation optimizations for specific files, since that seems to be a commonly needed feature.

worleyph:
I see - in case you don't know what flags are set (and are too lazy to look :-) ).

worleyph:
@jgfouca , sorry to be slow with this. I experimentally verified that e3sm_remove_flags was behaving correctly for PGI on Compy. I'll test Intel as well. I am also trying to apply this to the original use case - trying to determine why applying -O1 (or -O0) globally or by component in Macros.cmake was not behaving the same as applying it to all files individually in a Depends file. My first experiments on this did not produce the expected results, and I want to dig into this a little before approving this PR.

worleyph:
Tested removing both '-O2' and '-fp-model source' from the PERFOBJS list of files (which have '-O3' and '-fp-model fast' added) for an Intel/Compy build and it worked fine. Even when not necessary, this might be useful in cleaning up the compile lines to make them easier to read (something to think about for future work on Depends files).

worleyph:
So, I have no idea what is going on, but it appears to be unrelated to this PR. For you enjoyment ...

     a) In Macros.cmake and Macros.make, set non-DEBUG default optimization level to -O1 . 
        With this (the default was -O2), pgi/Compy is BFB with change in PE layout. With the -O2, it is not.

     b) In Macros.cmake and Macros.make, (still) set  non-DEBUG default optimization level to -O1 .
        Then, also in Macros.cmake and Macros.make added '-O2' to the rule for the cam component, i.e.

          if("${COMP_NAME}" STREQUAL "cam")
            set(FFLAGS "${FFLAGS} -O2 -Mnovect")
          endif()

        I was expecting this to be nonBFB, and would then use a Depends file to back out '-O2' from files
        that could not handle it. However, this experiment (-O1 for eveything except -O2 for CAM) was BFB.
        The problem files are then elsewhere. (This was an F case.) 

Again, this irrelevant to this PR (even though it was this experiment that led to the realization, from @amametjanov, that -O1 or -O0 does not override -O2 ).

Az, I would suggest modifying the Macros.cmake and Macros.make to to the above (-O1 for everything, and -O2 for cam), and possibly add similar modifications for the other components (ocn, lnd, etc.) after testing. It currently is not BFB, so we don't want to leave it this way. I'll create a github issue (if there isn't one already - I can't remember.)

components/cmake/cmake_util.cmake
amametjanov: My vote is to remove `-O1` too, because pgi's `FFLAGS_NOOPT` is set to `-O0`.

jgfouca: @amametjanov , will do

jgfouca:
Merged to next.

jgfouca:
Testing on next revealed that languages other than fortran (C and CXX) were not being handled correctly due to this PR. Fix coming.

PR #3708, opened 2020-07-16
Cime becomes a submodule

jgfouca:
This is a big day and a big change for E3SM but impacts should be very minimal since the cime code should be identical to what's currently in the E3SM cime subdirectory.

From hereon, E3SM/CIME integration should be as simple as updating this submodule.

[BFB]


jgfouca:
developer testing looks great on mappy, merging to next.

PR #3679, opened 2020-07-02
Update provenance capture logic in maint-1.0

worleyph:
When compy support was added for maint-1.0, the provenance.py file
was not updated to also enable capture of provenance data for Compy.
The job monitoring script syslog.compy was also not added.

Also, when anvil changed from a PBS batch scheduler to SLURM, the
provenance.py and syslog.anvil files in maint-1.0 were not updated
accordingly.

In addition, summit support does not appear in provenance.py
for maint-1.0, and there is also no syslog.summit file.

Finally, the projects for which performance archiving is supported
by default has not been updated in maint-1.0 for cori-haswell,
cori-knl, theta, and summit, nor has the location of the performance
archive on Summit.

Because of these missing changes, expected provenance data is not
being captured with the other performance data in the performance
archive. Here these deficiencies are remedied.

The job process monitoring scripts (syslog.XXX) for cori-haswell,
cori-knl, and theta are also updated, as they have slightly improved
behavior over the originals.

The decommissioned systems edison, mira, and titan were also removed
for provenance.py, but not from config_machines.xml, and the associated
job progress monitoring scripts were deleted.

Note that maint-1.0 does not work on summit currently, so these
updates to the performance archiving logic are simply for
documentation purposes.

[BFB]

Fixes #3659 for maint-1.0

PR #3710, opened 2020-07-17
removing sqs command for cori which was introduced by PR #3679

bibiraju:
Removing sqs command for cori which was introduced by PR #3679. This is due the merge conflicts with PR #3704 

Only affects Cori.

[BFB]

### DIFF
--- a/cime/config/e3sm/machines/syslog.cori-haswell
+++ b/cime/config/e3sm/machines/syslog.cori-haswell
@@ -13,7 +13,7 @@ set dir = $6
 # Target length was determined empirically (maximum number of lines before job mapping 
 #  information starts + number of nodes), and it may need to be adjusted in the future.
 # (Note that calling script 'touch'es the e3sm log file before spawning this script, so that 'wc' does not fail.)
-set nnodes = `sqs -f $jid | grep -F NumNodes | sed 's/ *NumNodes=\([0-9]*\) .*/\1/' `
+set nnodes = `scontrol show jobid $jid | grep -F NumNodes | sed 's/ *NumNodes=\([0-9]*\) .*/\1/' `
 @ target_lines = 150 + $nnodes
 sleep 10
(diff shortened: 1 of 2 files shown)


ndkeen:
Hmm, I thought I already merged this change into maint-1.0 here:
https://github.com/E3SM-Project/E3SM/pull/3704

bibiraju:
@ndkeen But the sqs command was re-introduced with this PR #3679. I noticed it only after merging it. So created this one.

ndkeen:
Re-introduced?  By mistake?

bibiraju:
@ndkeen PR #3679 introduced one more occurrence of sqs command in two files. This below line was introduced by PR #3679 which was created before #3704 but merged only after 3704 was merged.
`set nnodes = `sqs -f $jid | grep -F NumNodes | sed 's/ *NumNodes=\([0-9]*\) .*/\1/' ` `

The above line is new code from #3679 and uses sqs command in it. PR #3704 removed other occurrences of sqs but this.

ndkeen:
@bibiraju I would think the right thing to do is revert the change where this happened.

bibiraju:
ok. I will wait for @worleyph as he is the author of the original PR.

rljacob:
@ndkeen no just merge this PR.

worleyph:
@ndkeen , please go ahead and merge this PR.

rljacob:
The maint-1.0 test suite is running fine even though this PR hasn't been merged.  Did NERSC put the sqs command back?

ndkeen:
I just haven't had time to do this -- I wanted to at least run a test.
But, I also noticed that cdash was reporting OK so assumed it wasn't high priority.
I think the sqs command that was in provenance.py *was* changed and allows jobs to run.
However, currently, jobs without this PR would not save info about other jobs on the system which gets saved to perf data.

rljacob:
@ndkeen please finish this PR.

PR #3712, opened 2020-07-18
JLSE library updates

amametjanov:
Update libraries on JLSE machine:
- fix gnu compiler
- add openmpi library
- update pe and affinity settings

[NML]

### DIFF
--- a/components/cam/src/chemistry/mozart/mo_drydep.F90
+++ b/components/cam/src/chemistry/mozart/mo_drydep.F90
@@ -2007,7 +2007,6 @@ contains
     
     if (aqua_planet) then
       fraction_landuse = 0.
-      soilw_3d = 0.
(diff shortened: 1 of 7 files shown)


components/cam/src/chemistry/mozart/mo_drydep.F90
```diff
@@ -2007,7 +2007,6 @@ subroutine get_landuse_and_soilw_from_file(do_soilw)
     
     if (aqua_planet) then
       fraction_landuse = 0.
-      soilw_3d = 0.
```
amametjanov, line 2010: With aqua-planet ON, `soilw_3d` is not allocated, removing: it's allocated with `do_soilw=.true.`. `e3sm_developer` tests still pass after this change.

PR #3716, opened 2020-07-20
[WIP] openmp4.5 version of CRM code

xyuan:
This is working version of CRM code with OpenMP4.5, changes include:
1): change crm_input and crm_output from type to module because OpenMP4.5 doesn't support class map
2): add OpenMP clause to crm/sam code.

non-BFB


cime_config/machines/config_machines.xml
```diff
@@ -2801,7 +2801,7 @@
       <executable>/gpfs/alpine/world-shared/cli115/mpirun.summit</executable>
       <!-- <executable>jsrun</executable> -->
       <arguments>
-        <arg name="num_tasks"> -n {{ total_tasks }} -N {{ tasks_per_node }}</arg>
+        <arg name="num_tasks"> -n {{ total_tasks }} -N {{ tasks_per_node }} -gpu </arg>
```
mrnorman, line 2804: This will use the GPU setup for all IBM cases. I think it would be wise to create an ibm compiler for CPU runs and an ibmgpu compiler for GPU runs like we do with other comiplers.

components/cam/src/physics/crm/crm_history.F90
```diff
@@ -488,73 +487,73 @@ subroutine crm_history_out(state, ptend, crm_state, crm_rad, crm_output, crm_ecp
    call outfld('CRM_T   ',crm_state%temperature, pcols, lchnk )
 
    if (MMF_microphysics_scheme .eq. 'sam1mom') then
-      call outfld('CRM_QV  ',(crm_state%qt-crm_output%qcl-crm_output%qci), pcols, lchnk )
+      call outfld('CRM_QV  ',(crm_state%qt-crm_output_qcl-crm_output_qci), pcols, lchnk )
```
mrnorman, line 490: @whannah1 , @brhillman How do yall feel about these changes: removing the `crm_output` data structure and putting flat arrays in their place? I think I would rather see the data structure remain, and instead we just change the call to `crm(...)` to pass these arrays individually.

components/cam/src/physics/crm/crm_physics.F90
```diff
@@ -310,9 +310,9 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
 
    use crm_state_module,       only: crm_state_type
    use crm_rad_module,         only: crm_rad_type, crm_rad_initialize, crm_rad_finalize
-   use crm_input_module,       only: crm_input_type
-   use crm_output_module,      only: crm_output_type, crm_output_initialize, crm_output_finalize
    use crm_ecpp_output_module, only: crm_ecpp_output_type
+   use crm_input_module
```
mrnorman, line 314: Why remove the `only:` clause?

components/cam/src/physics/crm/crm_physics.F90
```diff
@@ -411,6 +406,14 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
    real(crm_rknd) :: crm_rotation_std     ! scaling factor for rotation (std dev of rotation angle)
    real(crm_rknd) :: crm_rotation_offset  ! offset to specify preferred rotation direction 
 #endif
+   integer, save :: icall = 0
+
+   ! CRM types
+   type(crm_state_type)  :: crm_state
```
mrnorman, line 412: @xyuan , it looks like you're getting rid of passing this to the subroutine as an argument and instead storing it as a local data object upon initialization. Is that what you're doing? And if so, what is the  reason for doing this?

components/cam/src/physics/crm/rrtmg/radiation.F90
```diff
@@ -755,14 +755,14 @@ subroutine radiation_init(phys_state)
...
+       call addfld ('CRM_QRAD   ', (/'crm_nx_rad','crm_ny_rad','crm_nz    '/), 'A', 'K/s', 'Radiative heating tendency')
+       call addfld ('CRM_QRS    ', (/'crm_nx_rad','crm_ny_rad','crm_nz    '/), 'I', 'K/s', 'CRM Shortwave radiative heating rate')
+       call addfld ('CRM_QRSC   ', (/'crm_nx_rad','crm_ny_rad','crm_nz    '/), 'I', 'K/s', 'CRM Clearsky shortwave radiative heating rate')
+       call addfld ('CRM_QRL    ', (/'crm_nx_rad','crm_ny_rad','crm_nz    '/), 'I', 'K/s', 'CRM Longwave radiative heating rate' )
+       call addfld ('CRM_QRLC   ', (/'crm_nx_rad','crm_ny_rad','crm_nz    '/), 'I', 'K/s', 'CRM Longwave radiative heating rate' )
        call addfld ('CRM_CLD_RAD', (/'crm_nx_rad','crm_ny_rad','crm_nz    '/), 'I', 'fraction', 'CRM cloud fraction' )
-       call addfld ('CRM_TAU  ', (/'crm_nx_rad','crm_ny_rad','crm_nz'/), 'A', '1', 'CRM cloud optical depth'  )
-       call addfld ('CRM_EMS  ', (/'crm_nx_rad','crm_ny_rad','crm_nz'/), 'A', '1', 'CRM cloud longwave emissivity'  )
+       call addfld ('CRM_TAU    ', (/'crm_nx_rad','crm_ny_rad','crm_nz    '/), 'A', '1', 'CRM cloud optical depth'  )
```
mrnorman, line 764: Thanks for fixing this. This kills GNU too. I think we must not be testing MMF + GNU in any of our E3SM regressions.

components/cam/src/physics/crm/sam/ADV_MPDATA/advect_scalar.F90
```diff
@@ -1,7 +1,9 @@
 module advect_scalar_mod
   use advect_scalar2D_mod
   use advect_scalar3D_mod
+#if defined(_OPENACC)
```
mrnorman, line 4: Why do this here? I though `#if defined(_OPENACC)` was done inside openacc_utils. This requires `#ifdef` everywhere we use it whereas having it inside `openacc_utils` looks a little cleaner.

components/cam/src/physics/crm/sam/ADV_MPDATA/advect_scalar.F90
```diff
@@ -41,26 +50,39 @@ subroutine advect_scalar (ncrms,f,fadv,flux)
     else
       call advect_scalar2D(ncrms, f, u, w, rho, rhow, flux)
     endif
-
+#if defined(_OPENACC)
     !$acc parallel loop collapse(2) async(asyncid)
+#elif defined(_OPENMP)
```
mrnorman, line 55: @xyuan , I think this is potentially a problem. `_OPENMP` is defined for any OpenMP-enabled compiler, but we only want this to be run if we're using OpenMP **4.5**. So I would recommend changing the define you pass to be `-D_OPENMP45` and to test for `#if defined(_OPENMP45)` instead of testing for `_OPENMP`.

components/cam/src/physics/crm/sam/ADV_MPDATA/advect_scalar2D.F90
```diff
@@ -105,8 +130,11 @@ subroutine advect_scalar2D (ncrms, f, u, w, rho, rhow, flux)
         enddo
       enddo
     endif  ! nonos
-
+#if defined(_OPENACC)
     !$acc parallel loop collapse(3) async(asyncid)
```
mrnorman, line 134: OpenACC directives are by default ignored if the compiler isn't OpenACC-capable and OpenACC isn't enabled. So there's no need to wrap these in `#ifdef`s I don't think.

components/cam/src/physics/crm/sam/ADV_MPDATA/advect_scalar3D.F90
```diff
@@ -109,9 +136,13 @@ subroutine advect_scalar3D (ncrms, f, u, v, w, rho, rhow, flux)
         enddo
       endif
       if (rank.gt.nsubdomains-nsubdomains_x-1) then
+#if defined(_OPENACC)
         !$acc parallel loop collapse(4) async(asyncid)
+#elif defined(_OPENMP)
+        !$omp target teams distribute parallel do collapse(4)
+#endif
         do k=1,nzm
-          do j=ny+1,dimy2_v
```
mrnorman, line 114: Why change `nyp1` to `ny+1`? Unless `nyp1` is causing a bug, I think we should keep it the same.

components/cam/src/physics/crm/sam/MICRO_SAM1MOM/cloud.F90
```diff
@@ -34,81 +35,83 @@ subroutine cloud(ncrms,q,qp,qn)
...
-            else
-              qn(icrm,i,j,k) = 0.
-            endif
-            tabs(icrm,i,j,k) = tabs1
-            qp(icrm,i,j,k) = max(real(0.,crm_rknd),qp(icrm,i,j,k)) ! just in case
-          end do
+#elif defined(_OPENMP)
+    !$omp target teams distribute parallel do private(icrm,i,j,k,qsatt,dqsat,dtabs,tabs1,fff,dfff) 
+#endif
+    do iloop = 0, loop_size-1
```
mrnorman, line 43: This is a pretty sweeping change to the kernels. If there's a bug we're working around in OpenMP, then please create one version of the new manually collapsed kernel for OpenMP4.5, and leave the original kernel there with OpenACC.

components/cam/src/physics/crm/sam/MICRO_SAM1MOM/micro_params.F90
```diff
@@ -35,10 +35,17 @@ module micro_params
   real(crm_rknd), parameter :: b_grau = 0.5  ! Fall speed exponent for graupel
 
   ! Autoconversion
+#ifdef CLUBB_CRM   /*microphysical tuning for CLUBB*/
```
mrnorman, line 38: What is this addition here for?

components/cam/src/physics/crm/sam/MICRO_SAM1MOM/microphysics.F90
```diff
@@ -588,15 +729,25 @@ subroutine precip_fall(ncrms,hydro_type, omega, ind)
...
+#elif defined(_OPENMP)
+        !$omp target teams distribute parallel do collapse(4)
+#endif
         do k=1,nzm
           do j=1,ny
             do i=1,nx
               do icrm = 1 , ncrms
                 kb=max(1,k-1)
                 ! Add limited flux correction to fz(k).
-                fz(icrm,i,j,k) = fz(icrm,i,j,k) + pp(www(icrm,i,j,k))*min(real(1.,crm_rknd),mx(icrm,i,j,k), mn(icrm,i,j,kb)) - &
```
mrnorman, line 598: If this is a bug workaround, can you keep the OpenACC code as is and place an `#ifdef` for the workaround for IBM OpenMP4.5?

components/cam/src/physics/crm/sam/SGS_TKE/diffuse_mom2D.F90
```diff
@@ -62,22 +74,47 @@ subroutine diffuse_mom2D(ncrms,grdf_x, grdf_z, dimx1_d, dimx2_d, dimy1_d, dimy2_
...
     do k=1,nzm
       do i=1,nx
         do icrm = 1 , ncrms
           kc=k+1
           ib=i-1
-          dudt(icrm,i,j,k,na)=dudt(icrm,i,j,k,na)-(fu(icrm,i,j,k)-fu(icrm,ib,j,k))
-          dvdt(icrm,i,j,k,na)=dvdt(icrm,i,j,k,na)-(fv(icrm,i,j,k)-fv(icrm,ib,j,k))
-          dwdt(icrm,i,j,kc,na)=dwdt(icrm,i,j,kc,na)-(fw(icrm,i,j,k)-fw(icrm,ib,j,k))
+          dfu = (fu(icrm,i,j,k)-fu(icrm,ib,j,k))
+          dfv = (fv(icrm,i,j,k)-fv(icrm,ib,j,k))
```
mrnorman, line 88: This doesn't appear to be a case that needs atomics. You're allowed to read from i and i-1 in a loop so long as you aren't writing to the same variable. fu, fv, and fw are read-only, so it's OK to read from i and i-1.

components/cam/src/physics/crm/sam/SGS_TKE/diffuse_mom2D.F90
```diff
@@ -123,35 +175,71 @@ subroutine diffuse_mom2D(ncrms,grdf_x, grdf_z, dimx1_d, dimx2_d, dimy1_d, dimy2_
...
     !$acc parallel loop collapse(3) async(asyncid)
+#elif defined(_OPENMP)
+    !$omp target teams distribute parallel do collapse(3)
+#endif
     do k=1,nzm
       do i=1,nx
         do icrm = 1 , ncrms
           kc=k+1
           rhoi = 1./(rho(icrm,k)*adz(icrm,k))
-          dudt(icrm,i,j,k,na)=dudt(icrm,i,j,k,na)-(fu(icrm,i,j,kc)-fu(icrm,i,j,k))*rhoi
```
mrnorman, line 139: Again, I don't believe this needs atomics.

components/cam/src/physics/crm/sam/SGS_TKE/diffuse_mom2D.F90
```diff
@@ -123,35 +175,71 @@ subroutine diffuse_mom2D(ncrms,grdf_x, grdf_z, dimx1_d, dimx2_d, dimy1_d, dimy2_
...
+#if defined(_OPENACC)
     !$acc parallel loop collapse(3) async(asyncid)
+#elif defined(_OPENMP)
+    !$omp target teams distribute parallel do collapse(3)
+#endif
     do k=2,nzm
       do i=1,nx
         do icrm = 1 , ncrms
           rhoi = 1./(rhow(icrm,k)*adzw(icrm,k))
-          dwdt(icrm,i,j,k,na)=dwdt(icrm,i,j,k,na)-(fw(icrm,i,j,k+1)-fw(icrm,i,j,k))*rhoi
```
mrnorman, line 150: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/SGS_TKE/diffuse_mom3D.F90
```diff
@@ -63,22 +73,47 @@ subroutine diffuse_mom3D(ncrms,grdf_x, grdf_y, grdf_z, dimx1_d, dimx2_d, dimy1_d
...
+#elif defined(_OPENMP)
+    !$omp target teams distribute parallel do collapse(4)
+#endif
     do k=1,nzm
       do j=1,ny
         do i=1,nx
           do icrm = 1 , ncrms
             kc=k+1
             ib=i-1
-            dudt(icrm,i,j,k,na)=dudt(icrm,i,j,k,na)-(fu(icrm,i,j,k)-fu(icrm,ib,j,k))
```
mrnorman, line 73: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/SGS_TKE/diffuse_mom3D.F90
```diff
@@ -100,22 +135,47 @@ subroutine diffuse_mom3D(ncrms,grdf_x, grdf_y, grdf_z, dimx1_d, dimx2_d, dimy1_d
...
+#elif defined(_OPENMP)
+    !$omp target teams distribute parallel do collapse(4)
+#endif
     do k=1,nzm
       do j=1,ny
         do i=1,nx
           do icrm = 1 , ncrms
             jb=j-1
             kc=k+1
-            dudt(icrm,i,j,k,na)=dudt(icrm,i,j,k,na)-(fu(icrm,i,j,k)-fu(icrm,i,jb,k))
```
mrnorman, line 110: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/SGS_TKE/diffuse_mom3D.F90
```diff
@@ -166,40 +241,76 @@ subroutine diffuse_mom3D(ncrms,grdf_x, grdf_y, grdf_z, dimx1_d, dimx2_d, dimy1_d
...
+#endif
     do k=1,nzm
       do j=1,ny
         do i=1,nx
           do icrm = 1 , ncrms
             kc=k+1
             rhoi = 1./(rho(icrm,k)*adz(icrm,k))
-            dudt(icrm,i,j,k,na)=dudt(icrm,i,j,k,na)-(fu(icrm,i,j,kc)-fu(icrm,i,j,k))*rhoi
-            dvdt(icrm,i,j,k,na)=dvdt(icrm,i,j,k,na)-(fv(icrm,i,j,kc)-fv(icrm,i,j,k))*rhoi
+            dfu = (fu(icrm,i,j,kc)-fu(icrm,i,j,k))*rhoi
```
mrnorman, line 270: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/SGS_TKE/diffuse_mom3D.F90
```diff
@@ -166,40 +241,76 @@ subroutine diffuse_mom3D(ncrms,grdf_x, grdf_y, grdf_z, dimx1_d, dimx2_d, dimy1_d
...
     !$acc parallel loop collapse(4) async(asyncid)
+#elif defined(_OPENMP)
+    !$omp target teams distribute parallel do collapse(4)
+#endif
     do k=2,nzm
       do j=1,ny
         do i=1,nx
           do icrm = 1 , ncrms
             rhoi = 1./(rhow(icrm,k)*adzw(icrm,k))
-            dwdt(icrm,i,j,k,na)=dwdt(icrm,i,j,k,na)-(fw(icrm,i,j,k+1)-fw(icrm,i,j,k))*rhoi
```
mrnorman, line 197: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/SGS_TKE/diffuse_scalar2D.F90
```diff
@@ -79,24 +101,39 @@ subroutine diffuse_scalar2D (ncrms,dimx1_d,dimx2_d,dimy1_d,dimy2_d,grdf_x,grdf_z
...
     do k=1,nzm
       do i=1,nx
         do icrm = 1 , ncrms
           ib=i-1
+#if defined(_OPENACC)
+          !$acc atomic update
+#elif defined(_OPENMP)
+          !$omp atomic update
+#endif
           dfdt(icrm,i,j,k)=dfdt(icrm,i,j,k)-(flx(icrm,i,j,k)-flx(icrm,ib,j,k))
```
mrnorman, line 118: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/SGS_TKE/diffuse_scalar2D.F90
```diff
@@ -107,32 +144,51 @@ subroutine diffuse_scalar2D (ncrms,dimx1_d,dimx2_d,dimy1_d,dimy2_d,grdf_x,grdf_z
...
       do i=1,nx
         do icrm = 1 , ncrms
           kb=k-1
           rhoi = 1./(adz(icrm,k)*rho(icrm,k))
           dfdt(icrm,i,j,k)=dtn*(dfdt(icrm,i,j,k)-(flx(icrm,i,j,k)-flx(icrm,i,j,kb))*rhoi)
+#if defined(_OPENACC)
+          !$acc atomic update
+#elif defined(_OPENMP)
+          !$omp atomic update
+#endif
```
mrnorman, line 183: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/SGS_TKE/diffuse_scalar3D.F90
```diff
@@ -147,29 +186,50 @@ subroutine diffuse_scalar3D (ncrms,dimx1_d,dimx2_d,dimy1_d,dimy2_d,grdf_x,grdf_y
...
     do k=1,nzm
       do j=1,ny
         do i=1,nx
           do icrm = 1 , ncrms
             ib=i-1
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
```
mrnorman, line 203: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/SGS_TKE/diffuse_scalar3D.F90
```diff
@@ -147,29 +186,50 @@ subroutine diffuse_scalar3D (ncrms,dimx1_d,dimx2_d,dimy1_d,dimy2_d,grdf_x,grdf_y
...
             dfdt(icrm,i,j,k)=dfdt(icrm,i,j,k)-(flx_x(icrm,i,j,k)-flx_x(icrm,ib,j,k))
             jb=j-1
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
```
mrnorman, line 210: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/SGS_TKE/diffuse_scalar3D.F90
```diff
@@ -181,35 +241,56 @@ subroutine diffuse_scalar3D (ncrms,dimx1_d,dimx2_d,dimy1_d,dimy2_d,grdf_x,grdf_y
...
         do i=1,nx
           do icrm = 1 , ncrms
             kb=k-1
             rhoi = 1./(adz(icrm,k)*rho(icrm,k))
             dfdt(icrm,i,j,k)=dtn*(dfdt(icrm,i,j,k)-(flx_z(icrm,i,j,k)-flx_z(icrm,i,j,kb))*rhoi)
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
```
mrnorman, line 282: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/accelerate_crm.F90
```diff
@@ -266,18 +309,36 @@ subroutine accelerate_crm(ncrms, nstep, nstop, ceaseflag)
...
+                !$acc atomic update
+#elif defined(_OPENMP)
+                !$omp atomic update
+#endif
                 u(icrm,i,j,k) = u(icrm,i,j,k) + crm_accel_factor * utend_acc(icrm,k) 
+#if defined(_OPENACC)
+                !$acc atomic update
+#elif defined(_OPENMP)
+                !$omp atomic update
+#endif
```
mrnorman, line 334: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/accelerate_crm.F90
```diff
@@ -266,18 +309,36 @@ subroutine accelerate_crm(ncrms, nstep, nstop, ceaseflag)
...
                 v(icrm,i,j,k) = v(icrm,i,j,k) + crm_accel_factor * vtend_acc(icrm,k) 
               endif
+#if defined(_OPENACC)
+              !$acc atomic update
+#elif defined(_OPENMP)
+              !$omp atomic update
+#endif
```
mrnorman, line 341: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/accelerate_crm.F90
```diff
@@ -342,7 +422,17 @@ subroutine accelerate_crm(ncrms, nstep, nstop, ceaseflag)
                   if (qt_res < 0._r8) then
                     ! qv was clipped; need to reduce qcl and qci accordingly
                     factor = 1._r8 + qt_res / (qcl(icrm,i,j,k) + qci(icrm,i,j,k))
+#if defined(_OPENACC)
+                    !$acc atomic update
+#elif defined(_OPENMP)
+                    !$omp atomic update
+#endif
```
mrnorman, line 429: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/accelerate_crm.F90
```diff
@@ -342,7 +422,17 @@ subroutine accelerate_crm(ncrms, nstep, nstop, ceaseflag)
...
                     qcl(icrm,i,j,k) = qcl(icrm,i,j,k) * factor
+#if defined(_OPENACC)
+                   !$acc atomic update
+#elif defined(_OPENMP)
+                   !$omp atomic update
+#endif
```
mrnorman, line 435: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/advect2_mom_xy.F90
```diff
@@ -33,24 +36,54 @@ subroutine advect2_mom_xy(ncrms)
               ic = i+1
               fu1 = dx25*(u(icrm,ic-1,j,k)+u(icrm,i-1,j,k))*(u(icrm,i-1,j,k)+u(icrm,ic-1,j,k))
               fu2 = dx25*(u(icrm,ic  ,j,k)+u(icrm,i  ,j,k))*(u(icrm,i  ,j,k)+u(icrm,ic  ,j,k))
+#if defined(_OPENACC)
+              !$acc atomic update
+#elif defined(_OPENMP)
+              !$omp atomic update
+#endif
```
mrnorman, line 43: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/advect2_mom_xy.F90
```diff
@@ -33,24 +36,54 @@ subroutine advect2_mom_xy(ncrms)
...
               dudt(icrm,i,j,k,na)  = dudt(icrm,i,j,k,na)  - (fu2-fu1)
               fv1 = dx25*(u(icrm,ic-1,j,k)+u(icrm,ic-1,jb,k))*(v(icrm,i-1,j,k)+v(icrm,ic-1,j,k))
               fv2 = dx25*(u(icrm,ic  ,j,k)+u(icrm,ic  ,jb,k))*(v(icrm,i  ,j,k)+v(icrm,ic  ,j,k))
+#if defined(_OPENACC)
+              !$acc atomic update
+#elif defined(_OPENMP)
+              !$omp atomic update
+#endif
```
mrnorman, line 51: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/advect2_mom_xy.F90
```diff
@@ -33,24 +36,54 @@ subroutine advect2_mom_xy(ncrms)
...
               dvdt(icrm,i,j,k,na)  = dvdt(icrm,i,j,k,na)  - (fv2-fv1)
               fw1 = dx25*(u(icrm,ic-1,j,k)*rho(icrm,k)*adz(icrm,k)+u(icrm,ic-1,j,kcu)*rho(icrm,kcu)*adz(icrm,kcu))*(w(icrm,i-1,j,kc)+w(icrm,ic-1,j,kc))
               fw2 = dx25*(u(icrm,ic  ,j,k)*rho(icrm,k)*adz(icrm,k)+u(icrm,ic  ,j,kcu)*rho(icrm,kcu)*adz(icrm,kcu))*(w(icrm,i  ,j,kc)+w(icrm,ic  ,j,kc))
+#if defined(_OPENACC)
+              !$acc atomic update
+#elif defined(_OPENMP)
+              !$omp atomic update
+#endif
```
mrnorman, line 59: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/advect2_mom_xy.F90
```diff
@@ -33,24 +36,54 @@ subroutine advect2_mom_xy(ncrms)
...
 
               jc = j+1
               ib = i-1
               fu1 = dy25*(v(icrm,i,jc-1,k)+v(icrm,ib,jc-1,k))*(u(icrm,i,j-1,k)+u(icrm,i,jc-1,k))
               fu2 = dy25*(v(icrm,i,jc  ,k)+v(icrm,ib,jc  ,k))*(u(icrm,i,j  ,k)+u(icrm,i,jc  ,k))
+#if defined(_OPENACC)
+              !$acc atomic update
+#elif defined(_OPENMP)
+              !$omp atomic update
+#endif
```
mrnorman, line 70: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/advect2_mom_xy.F90
```diff
@@ -33,24 +36,54 @@ subroutine advect2_mom_xy(ncrms)
...
               dvdt(icrm,i,j,k,na) = dvdt(icrm,i,j,k,na) - (fv2-fv1)
               fw1 = dy25*(v(icrm,i,jc-1,k)*rho(icrm,k)*adz(icrm,k)+v(icrm,i,jc-1,kcu)*rho(icrm,kcu)*adz(icrm,kcu))*(w(icrm,i,j-1,kc)+w(icrm,i,jc-1,kc))
               fw2 = dy25*(v(icrm,i,jc  ,k)*rho(icrm,k)*adz(icrm,k)+v(icrm,i,jc  ,kcu)*rho(icrm,kcu)*adz(icrm,kcu))*(w(icrm,i,j  ,kc)+w(icrm,i,jc  ,kc))
+#if defined(_OPENACC)
+              !$acc atomic update
+#elif defined(_OPENMP)
+              !$omp atomic update
+#endif
```
mrnorman, line 86: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/advect2_mom_xy.F90
```diff
@@ -70,12 +107,27 @@ subroutine advect2_mom_xy(ncrms)
             ic = i+1
             fu1 = dx25*(u(icrm,ic-1,j,k)+u(icrm,i-1,j,k))*(u(icrm,i-1,j,k)+u(icrm,ic-1,j,k))
             fu2 = dx25*(u(icrm,ic  ,j,k)+u(icrm,i  ,j,k))*(u(icrm,i  ,j,k)+u(icrm,ic  ,j,k))
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
```
mrnorman, line 114: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/advect2_mom_xy.F90
```diff
@@ -70,12 +107,27 @@ subroutine advect2_mom_xy(ncrms)
...
             dudt(icrm,i,j,k,na)  = dudt(icrm,i,j,k,na)  - (fu2-fu1)
             fv1 = dx25*(u(icrm,ic-1,j,k)+u(icrm,i-1,j,k))*(v(icrm,i-1,j,k)+v(icrm,ic-1,j,k))
             fv2 = dx25*(u(icrm,ic  ,j,k)+u(icrm,i  ,j,k))*(v(icrm,i  ,j,k)+v(icrm,ic  ,j,k))
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
```
mrnorman, line 122: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/advect2_mom_xy.F90
```diff
@@ -70,12 +107,27 @@ subroutine advect2_mom_xy(ncrms)
...
             dvdt(icrm,i,j,k,na)  = dvdt(icrm,i,j,k,na)  - (fv2-fv1)
             fw1 = dx25*(u(icrm,ic-1,j,k)*rho(icrm,k)*adz(icrm,k)+u(icrm,ic-1,j,kcu)*rho(icrm,kcu)*adz(icrm,kcu))*(w(icrm,i-1,j,kc)+w(icrm,ic-1,j,kc))
             fw2 = dx25*(u(icrm,ic  ,j,k)*rho(icrm,k)*adz(icrm,k)+u(icrm,ic  ,j,kcu)*rho(icrm,kcu)*adz(icrm,kcu))*(w(icrm,i  ,j,kc)+w(icrm,ic  ,j,kc))
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
```
mrnorman, line 130: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/advect2_mom_z.F90
```diff
@@ -81,46 +107,85 @@ subroutine advect2_mom_z(ncrms)
...
-            dudt(icrm,i,j,k,na)=dudt(icrm,i,j,k,na)-(fuz(icrm,i,j,kc)-fuz(icrm,i,j,k))*rhoi
-            dvdt(icrm,i,j,k,na)=dvdt(icrm,i,j,k,na)-(fvz(icrm,i,j,kc)-fvz(icrm,i,j,k))*rhoi
+            dfuz = (fuz(icrm,i,j,kc)-fuz(icrm,i,j,k))*rhoi
+            dfvz = (fvz(icrm,i,j,kc)-fvz(icrm,i,j,k))*rhoi
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
+            dudt(icrm,i,j,k,na)=dudt(icrm,i,j,k,na)-dfuz
```
mrnorman, line 147: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/advect2_mom_z.F90
```diff
@@ -81,46 +107,85 @@ subroutine advect2_mom_z(ncrms)
...
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
+            dvdt(icrm,i,j,k,na)=dvdt(icrm,i,j,k,na)-dfvz
+          !  dudt(icrm,i,j,k,na)=dudt(icrm,i,j,k,na)-(fuz(icrm,i,j,kc)-fuz(icrm,i,j,k))*rhoi
+          !  dvdt(icrm,i,j,k,na)=dvdt(icrm,i,j,k,na)-(fvz(icrm,i,j,kc)-fvz(icrm,i,j,k))*rhoi
```
mrnorman, line 155: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/coriolis.F90
```diff
@@ -23,17 +28,35 @@ subroutine coriolis(ncrms)
...
-              u_av=0.25*(u(icrm,i,j,k)+u(icrm,ic,j,k)+u(icrm,i,jb,k)+u(icrm,ic,jb,k))
-              dvdt(icrm,i,j,k,na)=dvdt(icrm,i,j,k,na)-0.5*(fcory(icrm,j)+fcory(icrm,jb))*(u_av-ug0(icrm,k))
+              delta_dudt = fcory(icrm,j)*(v_av-vg0(icrm,k))-fcorzy(icrm,j)*w_av
+              delta_dvdt = 0.5*(fcory(icrm,j)+fcory(icrm,jb))*(u_av-ug0(icrm,k))
+
+#if defined(_OPENACC)
+              !$acc atomic update
+#elif defined(_OPENMP)
+              !$omp atomic update
+#endif
```
mrnorman, line 42: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/coriolis.F90
```diff
@@ -23,17 +28,35 @@ subroutine coriolis(ncrms)
...
+#elif defined(_OPENMP)
+              !$omp atomic update
+#endif           
+              dudt(icrm,i,j,k,na)=dudt(icrm,i,j,k,na)+delta_dudt
+#if defined(_OPENACC)
+              !$acc atomic update
+#elif defined(_OPENMP)
+              !$omp atomic update
+#endif  
+              dvdt(icrm,i,j,k,na)=dvdt(icrm,i,j,k,na)-delta_dvdt
```
mrnorman, line 49: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/coriolis.F90
```diff
@@ -42,8 +65,20 @@ subroutine coriolis(ncrms)
...
               w_av=0.25*(w(icrm,i,j,kc)+w(icrm,ib,j,kc)+w(icrm,i,j,k)+w(icrm,ib,j,k))
-              dudt(icrm,i,j,k,na)=dudt(icrm,i,j,k,na)+fcory(icrm,j)*(v(icrm,i,j,k)-vg0(icrm,k))-fcorzy(icrm,j)*w_av
-              dvdt(icrm,i,j,k,na)=dvdt(icrm,i,j,k,na)-fcory(icrm,j)*(u(icrm,i,j,k)-ug0(icrm,k))
+              delta_dudt = fcory(icrm,j)*(v(icrm,i,j,k)-vg0(icrm,k))-fcorzy(icrm,j)*w_av
+              delta_dvdt = fcory(icrm,j)*(u(icrm,i,j,k)-ug0(icrm,k))
+#if defined(_OPENACC)
+              !$acc atomic update
+#elif defined(_OPENMP)
+              !$omp atomic update
+#endif
```
mrnorman, line 74: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/coriolis.F90
```diff
@@ -42,8 +65,20 @@ subroutine coriolis(ncrms)
...
+              dudt(icrm,i,j,k,na)=dudt(icrm,i,j,k,na)+delta_dudt
+#if defined(_OPENACC)
+              !$acc atomic update
+#elif defined(_OPENMP)
+              !$omp atomic update
+#endif       
+              dvdt(icrm,i,j,k,na)=dvdt(icrm,i,j,k,na)-delta_dvdt
```
mrnorman, line 81: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/crm_module.F90
```diff
@@ -230,21 +301,21 @@ subroutine crm(lchnk, ncrms, dt_gl, plev,       &
...
-
-  crm_state_u_wind      => crm_state%u_wind     (1:ncrms,:,:,:)
-  crm_state_v_wind      => crm_state%v_wind     (1:ncrms,:,:,:)
-  crm_state_w_wind      => crm_state%w_wind     (1:ncrms,:,:,:)
-  crm_state_temperature => crm_state%temperature(1:ncrms,:,:,:)
-  crm_state_qt          => crm_state%qt         (1:ncrms,:,:,:)
-  crm_state_qp          => crm_state%qp         (1:ncrms,:,:,:)
-  crm_state_qn          => crm_state%qn         (1:ncrms,:,:,:)
-  
+  crm_rad_temperature = crm_rad%temperature(1:ncrms,1:crm_nx_rad,1:crm_ny_rad,1:crm_nz)
```
mrnorman, line 304: I think I like this approach. I chose to go this route with the C++ code. This will require extra copies, but it shields the OpenACC and OpenMP runtimes from the horrors of pbuf, which is a good thing.

Since you're doing this, however, I think it woudl be best to leave all of the data structures alone, meaning crm_output, crm_input, crm _rad, crm_state, etc. Since you're already creating a contiguous data structure to copy the data into, you don't need to get rid of the classes and pointers being passed into the `crm(...)` call. @whannah1  and @brhillman , what do y'all think?

components/cam/src/physics/crm/sam/crm_module.F90
```diff
@@ -725,13 +903,22 @@ subroutine crm(lchnk, ncrms, dt_gl, plev,       &
...
         do j=1,ny
           do i=1,nx
             do icrm = 1 , ncrms
               i_rad = (i-1) / (nx/crm_nx_rad) + 1
               j_rad = (j-1) / (ny/crm_ny_rad) + 1
+#if defined(_OPENACC)
+              !$acc atomic update
+#elif defined(_OPENMP)
+              !$omp atomic update
+#endif
```
mrnorman, line 921: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/crm_module.F90
```diff
@@ -861,74 +1053,139 @@ subroutine crm(lchnk, ncrms, dt_gl, plev,       &
...
       do i=1,nx
         do icrm = 1 , ncrms
           do k=1,nzm
             l = plev-k+1
             tmp1 = rho(icrm,nz-k)*adz(icrm,nz-k)*dz(icrm)*(qcl(icrm,i,j,nz-k)+qci(icrm,i,j,nz-k))
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
```
mrnorman, line 1071: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/crm_module.F90
```diff
@@ -861,74 +1053,139 @@ subroutine crm(lchnk, ncrms, dt_gl, plev,       &
...
+                crm_output_cldtop(icrm,l) = crm_output_cldtop(icrm,l) + 1
                 flag_top(icrm,i,j) = .false.
             endif
             if(pres(icrm,nz-k).ge.700.) then
+#if defined(_OPENACC)
+                !$acc atomic update
+#elif defined(_OPENMP)
+                !$omp atomic update
+#endif
```
mrnorman, line 1088: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/crm_module.F90
```diff
@@ -861,74 +1053,139 @@ subroutine crm(lchnk, ncrms, dt_gl, plev,       &
...
             endif
-
-            !$acc atomic update
-            crm_output%gliqwp(icrm,l) = crm_output%gliqwp(icrm,l) + qcl(icrm,i,j,k)
+#if defined(_OPENACC)
+           !$acc atomic update
+#elif defined(_OPENMP)
+           !$omp atomic update
+#endif
+            crm_output_gliqwp(icrm,l) = crm_output_gliqwp(icrm,l) + qcl(icrm,i,j,k)
```
mrnorman, line 1173: I must've missed these two. Thanks for catching this.

components/cam/src/physics/crm/rrtmg/radiation.F90
whannah1: This is for RRTMG which we are abandoning anyway, so while this is good to fix, I don't think it's important.

components/cam/src/physics/crm/sam/ADV_MPDATA/advect_scalar2D.F90
whannah1: Totally agree, we shouldn't be controlling this with ifdefs.

components/cam/src/physics/crm/sam/MICRO_SAM1MOM/micro_params.F90
whannah1: I think this is probably because Xingqiu was copying over work he did on the ECP repo which still has the CLUBB code that we took out of the E3SM repo. So this should not be added.

components/cam/src/physics/crm/sam/fft.F90
```diff
@@ -3,8 +3,12 @@ module fft_mod
 contains
 
       subroutine fft991_crm(a,work,trigs,ifax,inc,jump,n,lot,isign)
-      !$acc routine seq
       use params, only: crm_rknd
+#if defined(_OPENACC)
```
mrnorman, line 7: I believe the placement of `!$acc routine seq` matters for the PGI compiler. Please move `!$acc routine seq` back to the original location for OpenACC for all routines that use it.

components/cam/src/physics/crm/crm_physics.F90
whannah1: We should definitely keep the only clause in.

components/cam/src/physics/crm/sam/forcing.F90
```diff
@@ -20,45 +22,92 @@ subroutine forcing(ncrms)
...
+#endif
     do k=1,nzm
       do j=1,ny
         do i=1,nx
           do icrm = 1 , ncrms
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
```
mrnorman, line 61: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/forcing.F90
```diff
@@ -20,45 +22,92 @@ subroutine forcing(ncrms)
...
             t(icrm,i,j,k)=t(icrm,i,j,k) + ttend(icrm,k) * dtn
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
```
mrnorman, line 67: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/forcing.F90
```diff
@@ -20,45 +22,92 @@ subroutine forcing(ncrms)
...
             dudt(icrm,i,j,k,na)=dudt(icrm,i,j,k,na) + utend(icrm,k)
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
```
mrnorman, line 100: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/forcing.F90
```diff
@@ -20,45 +22,92 @@ subroutine forcing(ncrms)
...
+#elif defined(_OPENMP)
+              !$omp atomic update
+#endif
               qpoz(icrm,k) = qpoz(icrm,k) + micro_field(icrm,i,j,k,index_water_vapor)
             end if
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
```
mrnorman, line 94: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/ice_fall.F90
```diff
@@ -134,38 +175,73 @@ subroutine ice_fall(ncrms)
             ! The cloud ice increment is the difference of the fluxes.
             dqi=coef*(fz(icrm,i,j,k)-fz(icrm,i,j,k+1))
             ! Add this increment to both non-precipitating and total water.
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
```
mrnorman, line 182: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/ice_fall.F90
```diff
@@ -134,38 +175,73 @@ subroutine ice_fall(ncrms)
...
             ! The latent heat flux induced by the falling cloud ice enters
             ! the liquid-ice static energy budget in the same way as the
             ! precipitation.  Note: use latent heat of sublimation.
             lat_heat  = (fac_cond+fac_fus)*dqi
             ! Add divergence of latent heat flux to liquid-ice static energy.
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
```
mrnorman, line 201: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/ice_fall.F90
```diff
@@ -134,38 +175,73 @@ subroutine ice_fall(ncrms)
...
     do j=1,ny
       do i=1,nx
         do icrm = 1 , ncrms
           coef=dtn/dz(icrm)
           dqi=-coef*fz(icrm,i,j,1)
+#if defined(_OPENACC)
+          !$acc atomic update
+#elif defined(_OPENMP)
+          !$omp atomic update
+#endif
```
mrnorman, line 229: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/ice_fall.F90
```diff
@@ -134,38 +175,73 @@ subroutine ice_fall(ncrms)
...
+          !$acc atomic update
+#elif defined(_OPENMP)
+          !$omp atomic update
+#endif      
           precsfc(icrm,i,j) = precsfc(icrm,i,j)+dqi
+#if defined(_OPENACC)
+          !$acc atomic update
+#elif defined(_OPENMP)
+          !$omp atomic update
+#endif
```
mrnorman, line 235: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/press_grad.F90
```diff
@@ -26,28 +29,57 @@ subroutine press_grad(ncrms)
...
-            dvdt(icrm,i,j,k,na)=dvdt(icrm,i,j,k,na)-(p(icrm,i,j,k)-p(icrm,i,jb,k))*rdy
-            dwdt(icrm,i,j,k,na)=dwdt(icrm,i,j,k,na)-(p(icrm,i,j,k)-p(icrm,i,j,kb))*rdz
+            dpdx = (p(icrm,i,j,k)-p(icrm,ib,j,k))*rdx
+            dpdy = (p(icrm,i,j,k)-p(icrm,i,jb,k))*rdy
+            dpdz = (p(icrm,i,j,k)-p(icrm,i,j,kb))*rdz
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
```
mrnorman, line 39: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/press_grad.F90
```diff
@@ -26,28 +29,57 @@ subroutine press_grad(ncrms)
...
+            dudt(icrm,i,j,k,na)=dudt(icrm,i,j,k,na)-dpdx
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
```
mrnorman, line 45: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/press_grad.F90
```diff
@@ -26,28 +29,57 @@ subroutine press_grad(ncrms)
...
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif      
+            dvdt(icrm,i,j,k,na)=dvdt(icrm,i,j,k,na)-dpdy
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
```
mrnorman, line 51: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/press_grad.F90
```diff
@@ -26,28 +29,57 @@ subroutine press_grad(ncrms)
...
     do k=1,nzm
       do j=1-YES3D,ny !bloss: 0,n* fixes computation of dp/d* in stats.
         do i=0,nx
           do icrm = 1 , ncrms
+#if defined(_OPENACC)
+            !$acc atomic update
+#elif defined(_OPENMP)
+            !$omp atomic update
+#endif
```
mrnorman, line 70: I don't believe this needs atomics.

components/cam/src/physics/crm/sam/sat.F90
```diff
@@ -11,7 +11,11 @@ module sat_mod
 
   real(crm_rknd) function esatw_crm(t)
     use params, only: crm_rknd
+#if defined(_OPENACC)
```
mrnorman, line 14: Please return the OpenACC's `!$acc routine seq` back to its original location

components/cam/src/physics/crm/crm_physics.F90
```diff
@@ -729,25 +732,20 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
...
+      crm_input_qccl(1:ncol,1:pver) = state%q(1:ncol,1:pver,ixcldliq)
+      crm_input_qiil(1:ncol,1:pver) = state%q(1:ncol,1:pver,ixcldice)
+      crm_input_ps(1:ncol) = state%ps(1:ncol)
+      crm_input_pmid(1:ncol,1:pver) = state%pmid(1:ncol,1:pver)
+      crm_input_pint(1:ncol,1:pver+1) = state%pint(1:ncol,1:pver+1)
+      crm_input_pdel(1:ncol,1:pver) = state%pdel(1:ncol,1:pver)
+      crm_input_phis(1:ncol) = state%phis(1:ncol)
+      crm_input_ul(1:ncol,1:pver) = state%u(1:ncol,1:pver)
+      crm_input_vl(1:ncol,1:pver) = state%v(1:ncol,1:pver)
+      crm_input_ocnfrac(1:ncol) = cam_in%ocnfrac(1:ncol)
```
whannah1, line 748: I'm not sure we want to do this. We should have a discussion about how best to undo this input/output class suff.

components/cam/src/physics/crm/crm_physics.F90
whannah1: ok, I see why he did this now, but I'm not sure we want to make this change. We should really have a larger discussion about changes this big.

components/cam/src/physics/crm/crm_physics.F90
xyuan: OpenMP4.5 doesn't support to map types/class at the moment, another way to do this is use similar implementation as crm_state by creating new variables inside crm_module code. This way will keep the crm_physics and crm_history code nice. @whannah1 

I will try to remove some of the atomic update, but I have found that OpenMP is different from OpenACC in the way of atomic update, it turns out requiring more atomic update than OpenACC. @mrnorman

whannah1:
@xyuan I think we need to close this PR and have some discussion about these changes, and likely break this into several PRs. I'm uncomfortable with many of the changes, and even if we decide to keep them, we need to redo this in a way that doesn't inject ifdefs through the whole CRM code because that's unacceptable.

mrnorman:
I had a quick convo with Walter. Many of the `#ifdef`s are necessary, though I believe we can reduce this somewhat. Regarding whether you close this PR and create a new one or keep working through this PR, Walter feels more comfortable with this going through a new PR. I'm fine either way.

I think if the crm_output and crm_input changes were reverted in lieu of copying to local contiguous allocatables in crm_module, a lot of the concerns would be alleviated.

mrnorman:
Also, I realize my comments probably come across as overly negative. This was a lot of work, and I think it's great. With some restructuring, I think it will really improve the viability of the Fortran port. Particularly, copying to local contiguous allocatables might even help us work around the PGI pool allocator issue.

mrnorman:
Macros might be difficult. I'm fine if all we do is remove the `#ifdef _OPENACC` pragmas from the parallel loops

whannah1:
If we're really going to keep all these ifdefs then I think the variable name should be more clear that they are only for the MMF on the GPU, something like "MMF_OMP_GPU".

whannah1:
Oh, I guess so.... idk

xyuan:
parameterized macro could help us to do this if we use cpp preprocessor. I will try it.

mrnorman:
@xyuan Some thoughts about macros-based directives:
* We have to come up with a name we all like. Maybe `!$par` ?
* We should only use them for fully collapsed loops, which I believe is all of the loops.
* They should take a single parameter: an integer for the number of loops to collapse, so they'll be macro functions
* If we have private variables, that should be added after the macro function rather than handled within it. I think both OpenACC and OpenMP just use `private(vars...)`.
* For atomics, we can just have `!$atomic` maybe
* All rogue OpenACC and OpenMP pragmas should be left outside the macros (thing like `!$acc routine seq` or `!$acc cache(...)`)
* We can't use variadic macro functions (macro functions with an arbitrary number of parameters) because not all Fortran compilers support this.

What do you think?

xyuan:
Cool! that is also what I thought. Thanks very much@mrnorman

components/cam/src/physics/crm/sam/crm_module.F90
```diff
@@ -861,74 +1053,139 @@ subroutine crm(lchnk, ncrms, dt_gl, plev,       &
...
+                !$omp atomic update
+#endif
                 cwpl(icrm,i,j) = cwpl(icrm,i,j)+tmp1
                 cltemp(icrm,i,j) = max(cf3d(icrm,i,j,nz-k), cltemp(icrm,i,j))
             else if(pres(icrm,nz-k).lt.400.) then
+#if defined(_OPENACC)
+                !$acc atomic update
+#elif defined(_OPENMP)
+                !$omp atomic update
+#endif
```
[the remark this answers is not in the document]
rljacob: So both $acc and $omp "atomic update" pragmas should be removed?  Everywhere you said its not needed?

rljacob:
Yes we don't want macros the built-in Fortran pre-processor can't handle.

Why is it even necessary to use pre-processor calls to isolate !acc and !omp directives?  Won't a compiler for one just ignore the other?

xyuan:
> Yes we don't want macros the built-in Fortran pre-processor can't handle.
> 
> Why is it even necessary to use pre-processor calls to isolate !acc and !omp directives? Won't a compiler for one just ignore the other?

!$omp and !$acc can be handled by the compiler, we want to reduce the #ifdef(_OPENACC) type of code by replacing them with macros.

rljacob:
If the compiler can handle it, why did you do this in multiple places?
```
#if defined(_OPENACC)
          !$acc atomic update
#elif defined(_OPENMP)
          !$omp atomic update
#endif
```

mrnorman:
OpenACC directives don't need any `#ifdef` wrappers. The only reason OpenMP needs them is when there are GPU-specific directives that would cause failed compilation with CPU-only OpenMP. I think atomic update should not need an `#ifdef _OPENMP45` because it is valid CPU OpenMP

xyuan:
This is double protection for the OpenMP code, if the compiler option only has -qsmp, instead of -qsmp=omp, the compiler is confusing with these options, while env. "_OPENMP" and "_OPENACC" can protect this situation.

naromero77:
Its a pretty large PR and it would take me several hours to look through all the code. Here are some high level comments:

1. Yes, let's try to minimize the ifdefs where its feasible. I don't think they can all be eliminated as has already been said.

2. Are we confident that the OpenACC and OpenMP versions are spitting out the same numerical results? Doesn't need to be done as part of this PR, but it would be desirable to have some CI in place for both the OpenACC and OpenMP versions shortly thereafter. 

3. Related to #2. I am not sure what the plan is for OpenACC in E3SM. If we are going kill off the OpenACC versions in the 1+ year, perhaps its not worth developing macros that can switch back and forth between OpenACC and OpenMP. For OpenACC, we really just mean PGI compiler. For OpenMP, we mean several compilers (IBM XL, Cray, Intel, and NVidia (rebranded PGI)). There may be inconsistency between OpenMP implementation/behaviour -- you might discover that the macros are not robust enough to handle all these compilers.

I don't think its necessary to split up this PR into several smaller ones as long as everyone's issue are addressed and there is some level of confidence that the numerical results have not changed.

components/cam/src/physics/crm/crm_history.F90
brhillman: Yeah, I agree Matt.

components/cam/src/physics/crm/rrtmg/radiation.F90
brhillman: @whannah1 is your comment here maybe misplaced? I think these addflds are relevant to both RRTMG and RRTMGP.

whannah1: @brhillman If I grep for "CRM_TAU" I only see results in crm/rrtmg/radiation.F90

xyuan:
I am closing this PR, and redo it using the latest master branch

Issue #612, opened 2016-01-14
Arithmetic exception: Floating invalid operation error in components/homme/src/share/edge_mod.F90


singhbalwinder:
SMS_D.ne16_ne16.FC5.blues_nag test crashes at run time when compiled with the NAG compiler. The hash for the code is 7a3eefd5d5d (master branch). The model runs into this error when "-nan" debug flag is used. This debug flag forces all uninitialized variables/arrays to take a value of NaN. The code crashes when such variable/array is used without being initialized to a valid value.
Following is the error:

```
Runtime Error: *** Arithmetic exception: Floating invalid operation - aborting
/blues/gpfs/home/bsingh/acme/srccodes/master_test_del/ACME/components/homme/src/share/edge_mod.F90, line 1013: Error occurred in EDGE_MOD:EDGEVUNPACK
/blues/gpfs/home/bsingh/acme/srccodes/master_test_del/ACME/components/cam/src/dynamics/se/inidat.F90, line 396: Called by INIDAT:READ_INIDAT
/blues/gpfs/home/bsingh/acme/srccodes/master_test_del/ACME/components/cam/src/control/startup_initialconds.F90, line 54: Called by STARTUP_INITIALCONDS:INITIAL_CONDS
/blues/gpfs/home/bsingh/acme/srccodes/master_test_del/ACME/components/cam/src/dynamics/se/inital.F90, line 42: Called by INITAL:CAM_INITIAL
/blues/gpfs/home/bsingh/acme/srccodes/master_test_del/ACME/components/cam/src/control/cam_comp.F90, line 159: Called by CAM_COMP:CAM_INIT
/blues/gpfs/home/bsingh/acme/srccodes/master_test_del/ACME/components/cam/src/cpl/atm_comp_mct.F90, line 262: Called by ATM_COMP_MCT:ATM_INIT_MCT
/blues/gpfs/home/bsingh/acme/srccodes/master_test_del/ACME/cime/driver_cpl/driver/component_mod.F90, line 229: Called by COMPONENT_MOD:COMPONENT_INIT_CC
/blues/gpfs/home/bsingh/acme/srccodes/master_test_del/ACME/cime/driver_cpl/driver/cesm_comp_mod.F90, line 1145: Called by CESM_COMP_MOD:CESM_INIT
/blues/gpfs/home/bsingh/acme/srccodes/master_test_del/ACME/cime/driver_cpl/driver/cesm_driver.F90, line 102: Called by CESM_DRIVER
[...]

PR #3720, opened 2020-07-20
Update mpas-source: Add warning messages for snow and ice layers

### LINKED ISSUE #3669: warning needed for changing the number of snow/ice layers in MPAS-SI
Config options for the number of snow and ice layers in sea ice can be inconsistent with initial conditions. 
The current code will eventually crash without properly identifying the cause.
Critical error messages are needed.

akturner:
This PR brings in a new mpas-source submodule with changes only to the seaice core. It compares config options with initial condition values and adds a critical error if they differ.

see [MPAS-Dev PR #612](https://github.com/MPAS-Dev/MPAS-Model/pull/612)

Fixes #3669

[BFB]


jonbob:
merged to next

jonbob:
merged to master

PR #3723, opened 2020-07-21
Add switch for expanded snowpack model (use_extrasnowlayers)

amschne:
Several developments necessary to add a switch that will toggle on an expanded snowpack model valid for use on ice sheets.  This switch, called "use_extrasnowlayers," will also toggle on a new snowcapping routine (adapted from CLMv5) that removes snow from the bottom of the column, rather than diverting fresh new snow to an export term, and brings in improvements to the snowpack (and firn) densification scheme.  The switch is set to false by default, as this PR is designed to maintain BFB reproducibility unless a user sets the switch to true.  The switch, even when true, does not currently change the maximum number of snow layers (equal to 5) nor the maximum snowpack mass (equal to 1 m SWE).  This PR provides the control logic and software necessary to use up to 16 total snow layers, but the user will still have to increase these values (nlevsno and h2osnomax) in the clm_varpar.F90 and clm_varcon.F90 files (before, there was no way to increase the number of snow layers above 5).  Results from these developments, including simulated firn density profiles, are being submitted for publication in Geoscientific Model Development.

[BFB]

### DIFF
--- a/components/clm/src/biogeophys/BalanceCheckMod.F90
+++ b/components/clm/src/biogeophys/BalanceCheckMod.F90
@@ -444,21 +444,23 @@ contains
                      + qflx_snwcp_ice(c) + qflx_snwcp_liq(c) + qflx_sl_top_soil(c)
 
                 if (lun_pp%itype(l) == istdlak) then 
-                   if ( do_capsnow(c) .and. .not. use_extrasnowlayers) then
-                      snow_sources(c) = qflx_snow_grnd_col(c) &
-                           + frac_sno_eff(c) * (qflx_dew_snow(c) + qflx_dew_grnd(c) ) 
+                   if (.not. use_extrasnowlayers) then
+                      if ( do_capsnow(c)) then
+                         snow_sources(c) = qflx_snow_grnd_col(c) &
+                              + frac_sno_eff(c) * (qflx_dew_snow(c) + qflx_dew_grnd(c) ) 
 
-                      snow_sinks(c)   = frac_sno_eff(c) * (qflx_sub_snow(c) + qflx_evap_grnd(c) ) &
-                           + (qflx_snwcp_ice(c) + qflx_snwcp_liq(c) - qflx_prec_grnd(c))  &
-                           + qflx_snow_melt(c)  + qflx_sl_top_soil(c)
-                   else if (.not. use_extrasnowlayers) then
-                      snow_sources(c) = qflx_snow_grnd_col(c) &
-                           + frac_sno_eff(c) * (qflx_rain_grnd_col(c) &
-                           +  qflx_dew_snow(c) + qflx_dew_grnd(c) ) 
+                         snow_sinks(c)   = frac_sno_eff(c) * (qflx_sub_snow(c) + qflx_evap_grnd(c) ) &
+                              + (qflx_snwcp_ice(c) + qflx_snwcp_liq(c) - qflx_prec_grnd(c))  &
+                              + qflx_snow_melt(c)  + qflx_sl_top_soil(c)
+                      else
+                         snow_sources(c) = qflx_snow_grnd_col(c) &
+                              + frac_sno_eff(c) * (qflx_rain_grnd_col(c) &
+                              +  qflx_dew_snow(c) + qflx_dew_grnd(c) ) 
 
-                      snow_sinks(c)  = frac_sno_eff(c) * (qflx_sub_snow(c) + qflx_evap_grnd(c) ) &
-                           + qflx_snow_melt(c)  + qflx_sl_top_soil(c)
-                   else if (use_extrasnowlayers) then !++ams i.e., firn model
+                         snow_sinks(c)  = frac_sno_eff(c) * (qflx_sub_snow(c) + qflx_evap_grnd(c) ) &
+                              + qflx_snow_melt(c)  + qflx_sl_top_soil(c)
+                      endif
+                   else ! firn model
                       snow_sources(c) = qflx_snow_grnd_col(c) &
                            + frac_sno_eff(c) * (qflx_rain_grnd_col(c) &
                            +  qflx_dew_snow(c) + qflx_dew_grnd(c) ) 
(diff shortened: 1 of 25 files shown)


rljacob:
Is this for v2 or v3?

amschne:
I'm not sure.  However, these mods are turned off by default, and when off, should reproduce BFB output compared with master (@stephenprice, @jonbob).

stephenprice:
@rljacob since it is BFB when off (and off by default) we'd like to go ahead and include in v2 if possible. That gives us the option of having it on for some v2 Cryo configs, assuming we can vet it and have confidence in the results by then. Do you see problems with that, assuming we can get it reviewed and merged in time?

rljacob:
That's fine.  Please add a test which turns the feature on for some case.

stephenprice:
@amschne -- we can help you add a test where this code is active. We can follow up on some of the initial discussion for that on slack (and move back here once the discussion is more relevant to the PR).

components/clm/src/biogeophys/SnowHydrologyMod.F90
```diff
@@ -1167,475 +1287,612 @@ subroutine DivideSnowLayers(bounds, num_snowc, filter_snowc, &
...
-
-                mbc_phi(c,2) = mbc_phi(c,2)+zmbc_phi  ! (combo)
-                mbc_pho(c,2) = mbc_pho(c,2)+zmbc_pho  ! (combo)
-                moc_phi(c,2) = moc_phi(c,2)+zmoc_phi  ! (combo)
-                moc_pho(c,2) = moc_pho(c,2)+zmoc_pho  ! (combo)
-                mdst1(c,2) = mdst1(c,2)+zmdst1  ! (combo)
-                mdst2(c,2) = mdst2(c,2)+zmdst2  ! (combo)
-                mdst3(c,2) = mdst3(c,2)+zmdst3  ! (combo)
-                mdst4(c,2) = mdst4(c,2)+zmdst4  ! (combo)
+       if (.not. use_extrasnowlayers) then
```
bishtgautam, line 1290: Here the code is:

```
if (.not. use_extrasnowlayers) then
   ! 469 lines of code
else
  ! 133 lines of code
end if
```

Given the extremely large number of lines in the `if` condition, it is very hard to figure out where the `else` statement starts. Please introduce new subroutines that split the wor as

```
if (.not. use_extrasnowlayers) then
   call <approach-1>
else
   call <approach-2>
end if
```

components/clm/src/biogeophys/SnowHydrologyMod.F90
```diff
@@ -1700,7 +1957,374 @@ subroutine DivideSnowLayers(bounds, num_snowc, filter_snowc, &
...
+      do fc = 1, num_c
+         c = filter_c(fc)
+         !g = col_pp%gridcell(c)
+         t = col_pp%topounit(c)
+
+         if (forc_t(t) > tfrz + 2._r8) then
+            bifall(c) = 50._r8 + 1.7_r8*(17.0_r8)**1.5_r8
+         else if (forc_t(t) > tfrz - 15._r8) then
+            bifall(c) = 50._r8 + 1.7_r8*(forc_t(t) - tfrz + 15._r8)**1.5_r8
+         !else if ( new_snow_density == LoTmpDnsTruncatedAnderson1976 ) then
```
bishtgautam, line 2225: Please remove commentted out lines

components/clm/src/biogeophys/SnowHydrologyMod.F90
```diff
@@ -1700,7 +1957,374 @@ subroutine DivideSnowLayers(bounds, num_snowc, filter_snowc, &
...
+               t_for_bifall_degC = (forc_t(t)-tfrz)
+            else ! below comment from CESM2
+               ! Below -57.55 deg C, the following function starts to decrease with
+               ! decreasing temperatures. Limit the function to avoid this turning over.
+               t_for_bifall_degC = -57.55_r8
+            end if
+            bifall(c) = -(50._r8/15._r8 + 0.0333_r8*15_r8)*t_for_bifall_degC - 0.0333_r8*t_for_bifall_degC**2
+         end if
+
+         !if (wind_dependent_snow_density .and. forc_wind(t) > 0.1_r8 ) then
```
bishtgautam, line 2243: Please remove commentted out line

components/clm/src/biogeophys/SnowHydrologyMod.F90
```diff
@@ -1750,8 +2374,37 @@ subroutine Combo(dz,  wliq,  wice, t, dz2, wliq2, wice2, t2)
...
+     use SnowSnicarMod, only : snw_rds_max
+     implicit none
+     ! !ARGUMENTS:
+     real(r8), intent(IN) :: rds1         ! Layer 1 radius
+     real(r8), intent(IN) :: rds2         ! Layer 2 radius
+     real(r8), intent(IN) :: swtot        ! snow water total layer 2
+     real(r8), intent(IN) :: zwtot        ! snow water total layer 1
+     real(r8) :: mass_weighted_snowradius ! resulting bounded mass weighted snow radius
+
+     !SHR_ASSERT( (swtot+zwtot > 0.0_r8), errMsg(sourcefile, __LINE__))
```
bishtgautam, line 2396: Please remove commentted out line

components/clm/src/biogeophys/SnowHydrologyMod.F90
```diff
@@ -1700,7 +1957,374 @@ subroutine DivideSnowLayers(bounds, num_snowc, filter_snowc, &
...
+
+      !SHR_ASSERT_ALL((ubound(bifall) == (/bounds%endc/)), errMsg(sourcefile, __LINE__))   
+
+      associate(forc_t    => top_as_inst%tbot   , & ! Input:  [real(r8) (:) ]  atmospheric temperature (Kelvin)
+                forc_wind => top_as_inst%windbot  & ! Input:  [real(r8) (:) ]  atmospheric wind speed (m/s)
+                )
+
+      do fc = 1, num_c
+         c = filter_c(fc)
+         !g = col_pp%gridcell(c)
```
bishtgautam, line 2218: Please remove commentted out line

components/clm/src/biogeophys/SnowHydrologyMod.F90
```diff
@@ -587,24 +623,46 @@ subroutine SnowCompaction(bounds, num_snowc, filter_snowc, &
...
              c = filter_snowc(fc)
+             if (use_extrasnowlayers) then
+                t = col_pp%topounit(c)
+             end if
              if (j >= snl(c)+1) then
 
-                wx = h2osoi_ice(c,j) + h2osoi_liq(c,j)
-                void = 1._r8 - (h2osoi_ice(c,j)/denice + h2osoi_liq(c,j)/denh2o) / dz(c,j)
+                if (.not. use_extrasnowlayers) then
+                   ! ++ams this is redundant and should be removed
```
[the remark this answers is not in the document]
amschne: Yes.  I will remove these lines for both cases in my next commit.

components/clm/src/biogeophys/SnowHydrologyMod.F90
```diff
@@ -587,24 +623,46 @@ subroutine SnowCompaction(bounds, num_snowc, filter_snowc, &
...
                      /(frac_sno(c) * dz(c,j))
                 ! If void is negative, then increase dz such that void = 0.
                 ! This should be done for any landunit, but for now is done only for glacier_mec 1andunits.
-                l = col_pp%landunit(c)
-                if (ltype(l)==istice_mec .and. void < 0._r8) then
-                   dz(c,j) = h2osoi_ice(c,j)/denice + h2osoi_liq(c,j)/denh2o
-                   void = 0._r8
+                
+                if (.not. use_extrasnowlayers) then
+                   ! ++ams I don't think this is necessary (removed in CLMv5)
```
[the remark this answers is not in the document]
amschne: I'm not sure about this block.  I don't believe it (lines 660--663) is necessary, however, I'm not sure if removing it will maintain the model's BFB reproducibility.  Therefore I have "removed" these lines only for the new (use_extrasnowlayers) feature.

components/clm/src/biogeophys/BalanceCheckMod.F90
```diff
@@ -444,38 +444,54 @@ subroutine ColWaterBalanceCheck( bounds, num_do_smb_c, filter_do_smb_c, &
                      + qflx_snwcp_ice(c) + qflx_snwcp_liq(c) + qflx_sl_top_soil(c)
 
                 if (lun_pp%itype(l) == istdlak) then 
-                   if ( do_capsnow(c) ) then
+                   if ( do_capsnow(c) .and. .not. use_extrasnowlayers) then
```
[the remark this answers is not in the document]
amschne: Done

jonbob:
@bishtgautam - I was working on fixing this PR so that nlevsno and h2osnomax would be set based on the new use_extrasnowlayers namelist option, and think I found a problem. It looks like this setting is not broadcast in main/controlMod.F90 -- doesn't it need to be for all processors to have it?

bishtgautam:
You are correct. Those values need to be broadcasted.

jonbob:
Thanks @bishtgautam. Should I just make those changes and push them to this commit? I'm not sure it's fully functional at this point, so we could add the "do not merge" label?

bishtgautam:
I recommend making a new commit that fixes the issue. I'm fine with you making the new commit. If you rather prefer I make the commit that is fine by me.

jonbob:
A commit to this PR?

bishtgautam:
@jonbob Yes, make a new commit to this PR that broadcasts `use_extrasnowlayers`.

jonbob:
With the last couple of commits I am able to successfully run:
* SMS.ne30_oECv3_gis.IGCLM45_MLI.anvil_intel.clm-extrasnowlayers
which can be added to automated testing to track this capability

rljacob:
@golaz @stephenprice @kvcalvin is this needed for any v2 sims?

stephenprice:
@rljacob , sorry for the delay in responding. This is not strictly needed for v2. However, I think we were hoping to get it in if possible. The default is off so that the old snowpack model is used. @amschne and @czender should comment but I think the code is BFB in this (defaults on) case.

We've got a test branch right now to look at how difficult it would be to rebase these additions onto master (not terribly) and we are currently testing with this branch. Regardless of if they make it into v2 or not, we will certainly be working with this code during the v2 simulation campaign time period.

However, it's likely @jonbob who would be responsible for the final integration so I will let him have the final say. He has a much better feeling for how many other things we're competing with here.

stephenprice:
@rljacob -- I followed up with @jonbob. We will NOT push to have this formally approved / in for the WC v2 code freeze. Rather, it will be something we'll push for as part of any v2 Cryo tag after things w/ WC settle down a bit (CC'ing @czender and @amschne).

ndkeen:
Is this issue causing the error here:  https://github.com/E3SM-Project/E3SM/issues/3894
?

jonbob:
@ndkeen - not this one. It's not even merged into he repo. The PR that has created the need to redo the spinups is [PR #3846](https://github.com/E3SM-Project/E3SM/pull/3846)

ndkeen:
OK thanks -- feel free to delete my comments here then!

rljacob:
Ok.

rljacob:
Notes from IG telecon:  @jonbob and @stephenprice will get back to this.

rljacob:
Notes from IG:  Has been rebased on a different branch.  @stephenprice is testing.

jonbob:
Moved to #4097 after rebase/update

PR #3724, opened 2020-07-21
Make coupling_period calculation dependent on NCPL_BASE_PERIOD

jonbob:
This PR modifies the MOSART build-namelist so that the coupling frequency it computes, in the form of variable coupling_period, is dependent on the value of NCPL_BASE_PERIOD. Previously it was hard-wired to assume the base period was one day and ROF_NCPL was the number of couplings per day. This update was necessary for getting IG cases working, where the NCPL_BASE_PERIOD is often a year so that the landice code can run on order of days.

[BFB]


jonbob:
Tested by @stephenprice during IG-case development

stephenprice:
Confirming this fix works for an IG test case.

jonbob:
merged to next

jonbob:
merged to master

PR #3725, opened 2020-07-21
Make dtime calculation dependent on value of NCPL_BASE_PERIOD

jonbob:
This PR modifies the ELM namelist build system so that the coupling frequency it computes, in the form of variable dtime, is dependent on the value of NCPL_BASE_PERIOD. Previously it was hard-wired to assume the base period was one day and LND_NCPL was the number of couplings per day. This update was necessary for getting IG cases working, where the NCPL_BASE_PERIOD is often a year so that the landice code can run on order of days.

Tested by @stephenprice during IG-case development

[BFB]

### DIFF
--- a/components/clm/bld/CLMBuildNamelist.pm
+++ b/components/clm/bld/CLMBuildNamelist.pm
@@ -2188,27 +2188,22 @@ sub setup_logic_delta_time {
     if ( $l_ncpl <= 0 ) {
       fatal_error("bad value for -l_ncpl option.\n");
     }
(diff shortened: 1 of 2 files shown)


stephenprice:
Confirming this fix works in an IG test case.

jonbob:
@bishtgautam - I think added that line because there is similar logic for l_ncpl (LND_NCPL) which also comes from env_run.xml and should be defined as well. But I don't need to make it less clear!

jonbob:
merged to next

jonbob:
merged to master

PR #3726, opened 2020-07-22
Adds a new land test

bishtgautam:
Adds tests mod to SMS.r05_r05.I1850CLM45CN, clm-qian_1948, that avoids
downloading multiple DATAM files by setting the end year to be
the same as the start year. Also move test back to land_developer

[BFB]


rljacob:
@bishtgautam you still did this wrong.  Don't touch e3sm_land_developer.  Change the test in e3sm_integration.

bishtgautam:
@rljacob I have updated the PR

rljacob:
Sorry I changed my mind. We took that test out of developer because melvin couldn't handle it.  Maybe mappy can.  I force pushed back to your original version and will merge this.

bishtgautam:
In the first only commit, I forgot to remove the same test that is separately added in e3sm_integration_test suite https://github.com/E3SM-Project/E3SM/blob/f7fbb1ea2e62ab9d4c23610bea7b4b101f7cf319/cime_config/tests.py#L161

rljacob:
ok.  fixed. re-merged.

Issue #3628, opened 2020-06-08
Fatal error in PMPI_Waitall in DEBUG (with threads) using ne11 F-case with Eulerian transport and theta-l in H mode (cori-knl)


ndkeen:
This issue started with a problem I encountered in scream, but we eventually ruled out scream code and can reproduce the problem in E3SM. It is only a problem in DEBUG and only a problem with threading. And, there were some PE layouts that *did* work, but most fail in the same way.
It uses theta-l in hydrostatic mode, Eulerian transport. SO far, I've only tried on cori-knl.

I'm using a master of June 1st.

```
res=ne11_ne11
compset=FC5AV1C-L

./xmlchange CAM_TARGET=theta-l
./xmlchange DEBUG=true

cat <<EOF >> user_nl_cam                                                                                             
  semi_lagrange_cdr_alg = 3                                                                                          
  theta_hydrostatic_mode = .true.                                                                                    
  tstep_type = 5                                                                                                     
  state_debug_checks = .true.                                                                                        
  dtime = 300                                                                                                        
  se_tstep = 50                                                                                                      
  transport_alg = 0                                                                                                  
  dt_tracer_factor = 1                                                                                               
  hypervis_subcycle_q = 1                                                                                            
EOF                                                                                                                  
```

I'm running with stacked PE layouts and 64 MPI's per node. It has always run OK with 1 thread.

```
 56: Rank 56 [Mon Jun  8 11:14:22 2020] [c1-1c1s15n1] Fatal error in PMPI_Waitall: See the MPI_ERROR field in MPI_Status for the error code
 56: forrtl: error (76): Abort trap signal
 56: Image              PC                Routine            Line        Source             
 56: e3sm.exe           000000000CE8EB0E  Unknown               Unknown  Unknown
 56: e3sm.exe           000000000C6B4C20  Unknown               Unknown  Unknown
 56: e3sm.exe           000000000CE61F50  Unknown               Unknown  Unknown
 56: e3sm.exe           000000000CF8CB41  Unknown               Unknown  Unknown
 56: e3sm.exe           000000000C85B552  Unknown               Unknown  Unknown
 56: e3sm.exe           000000000C83FE00  Unknown               Unknown  Unknown
 56: e3sm.exe           000000000C83FF46  Unknown               Unknown  Unknown
 56: e3sm.exe           000000000C82E3A5  Unknown               Unknown  Unknown
[...]

PR #3727, opened 2020-07-22
add ne45pg2 grid support

whannah1:
Add support for the ne45pg2 grid, and specifically the ne45pg2_r05_oECv3 tri-grid. 

ne45pg2 has a similar number of columns to ne30np4 and almost exactly 1 degree uniform resolution. (Do not add support for a ne45np4 grid as this is  not planned to be used.) The namelist settings for ne45 follow the new protocol for setting the dynamics time step directly in seconds. Dycor hyperviscosity settings have been specified for both preqx and theta-l options and verified to work with the initial condition file generated through an iterative spin-up process. The physics time step for ne45 is reduced to 20 min.

[BFB]

### DIFF
--- a/components/cam/bld/namelist_files/namelist_defaults_cam.xml
+++ b/components/cam/bld/namelist_files/namelist_defaults_cam.xml
@@ -1540,8 +1540,6 @@
 <!-- viscosity  -->
 
 <nu_top> 2.5e5 </nu_top>
-<nu_top dyn_target="preqx"   hgrid="ne45np4"> 2.5e5 </nu_top>
-<nu_top dyn_target="theta-l" hgrid="ne45np4"> 2.5e5 </nu_top>
 <nu_top dyn_target="preqx" hgrid="ne240np4"> 1.0e5 </nu_top>
 <nu_top dyn_target="theta-l" hgrid="ne120np4"> 1e5 </nu_top>
 <nu_top dyn_target="theta-l" hgrid="ne240np4"> 4.3e4 </nu_top>
@@ -1557,7 +1555,7 @@
 <nu hgrid="ne16np4" >  7.0e15 </nu>
 <nu hgrid="ne30np4" >  1.0e15 </nu>
 <nu dyn_target="preqx"   hgrid="ne45np4" > 3.0E+14 </nu>
-<nu dyn_target="theta-l" hgrid="ne45np4" > 3.5e-8 </nu>
+<nu dyn_target="theta-l" hgrid="ne45np4" > 3.4e-8 </nu>
 <nu hgrid="ne60np4" >  1.25e14 </nu>
 <nu hgrid="ne120np4">  1.0e13 </nu>
 <nu dyn_target="theta-l" hgrid="ne120np4">  1.5e13 </nu>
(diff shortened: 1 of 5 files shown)


components/cam/bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -1534,6 +1540,8 @@
 <!-- viscosity  -->
 
 <nu_top> 2.5e5 </nu_top>
+<nu_top dyn_target="preqx"   hgrid="ne45np4"> 2.5e5 </nu_top>
+<nu_top dyn_target="theta-l" hgrid="ne45np4"> 2.5e5 </nu_top>
```
mt5555, line 1544: these two lines are not needed, since they are using the default value?

components/cam/bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -1548,6 +1556,8 @@
 <nu hgrid="ne11np4" >  2.0e16 </nu>
 <nu hgrid="ne16np4" >  7.0e15 </nu>
 <nu hgrid="ne30np4" >  1.0e15 </nu>
+<nu dyn_target="preqx"   hgrid="ne45np4" > 3.0E+14 </nu>
+<nu dyn_target="theta-l" hgrid="ne45np4" > 3.5e-8 </nu>
```
mt5555, line 1560: This is a little confusing, just want to make sure it is intentional:  this will be the first cubed-sphere configuration that is using the tensor HV?   

With tensor HV, the coefficient is grid independent, so we should need to set it for each resolution.  That is, the value 3.5e-8 should work with all grids.

components/cam/bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -1668,10 +1681,20 @@
 <se_nsplit dyn_target="preqx" hgrid="ne0np4_twpx4v1" > 4 </se_nsplit>
 <se_nsplit dyn_target="preqx"> 2 </se_nsplit>
 
+<!-- ne45 defaults to new time step interface -->
+<se_tstep hgrid="ne45np4">  200 </se_tstep>
```
mt5555, line 1685: se_tstep was already set above

components/cam/bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -1668,10 +1681,20 @@
...
+<dt_remap_factor  dyn_target="preqx" hgrid="ne45np4"> 2 </dt_remap_factor>
+<dt_tracer_factor dyn_target="preqx" hgrid="ne45np4"> 1 </dt_tracer_factor>
+<qsplit    dyn_target="preqx" hgrid="ne45np4"> -1 </qsplit>
+<rsplit    dyn_target="preqx" hgrid="ne45np4"> -1 </rsplit>
+<se_nsplit dyn_target="preqx" hgrid="ne45np4"> -1 </se_nsplit>
```
mt5555, line 1690: this block of XML breaks our convention of preqx using the old interface and theta-l using the new interface.   just wanted to make sure it was done intentionally.

components/cam/bld/namelist_files/namelist_defaults_cam.xml
whannah1: I originally had different values, I'll remove these.

components/cam/bld/namelist_files/namelist_defaults_cam.xml
whannah1: Good point. 
So Maybe I should just add a line like this?
  <nu dyn_target="theta-l"> 3.4e-8</nu>
That matches the value used for RRM, so it wouldn't affect that configuration.

components/cam/bld/namelist_files/namelist_defaults_cam.xml
whannah1: good, I'll move these lines around t be consistent with how everything is grouped.

components/cam/bld/namelist_files/namelist_defaults_cam.xml
whannah1: Yea, this was done intentionally to make it clear that the "old" style defaults are being overridden for ne45, but I can move them to be more consistent.

components/cam/bld/namelist_files/namelist_defaults_cam.xml
whannah1: After some offline discussion I decided to leave it in for now, and maybe later fix up this section to have a standard scaling anytime the tensor hyperviscosity is used.

whannah1:
I uploaded the ne45 files to blues and adjusted the permissions. I ran a SMS test on Summit to verify that the files get downloaded and it passed.

whannah1:
PR #3628 has been merged into the CIME master, so this PR should be ready to go.

rljacob:
@whannah1 what is a case I can run to test this PR?

whannah1:
@rljacob Just run a simple F-compset with the new grid:
SMS_Ln9.ne45pg2_r05_oECv3.FC5AV1C-L

You could also add an aqua test of the ne45pg2_ne45pg2 grid:
SMS_L9.ne45pg2_ne45pg2.F-EAMv1-AQP1

The ne45pg2_ne45pg2 grid was added awhile back. I think it was from a resolution convergence tests I did with the original physgrid PR. I never fully supported it, but it should work out of the box now for aqua planets since I've added the domain files and initial condition. It won't work for realistic cases since I never added topography or land model files

PR #3728, opened 2020-07-22
Update CIME submodule

jgfouca:
[NML] adds IOP setting


rljacob:
which namelist gets the iop setting?  docn?

jgfouca:
@rljacob:

```
2020-07-22 16:10:27: NLCOMP
Comparison failed between '/ascldap/users/jgfouca/acme/scratch/ERS.ne30_g16_rx1.A.mappy_gnu.C.20200722_160953_lvjho5/CaseDocs/drv_in' with '/sems-data-store/ACME/baselines/mappy/gnu/master/ERS.ne30_g16_rx1.A.mappy_gnu/CaseDocs/drv_in'
  found extra variable: 'iop_mode'
```

rljacob:
@bogensch what exactly does Iop_mode do?  It doesn't seem to be SCREAM-specific since its implemented by modifying docn and scam.

bogensch:
Based on the case used (in this instance, by “case” we mean a case from the Single Column Model case library), we create a domain that has a homogenous surface based on the latitude/longitude from the case forcing file.  We also apply the large-scale forcing of the case to the entire domain.

For a more longwinded description please see PR#210 in the SCREAM repo.  (and just FYI that I’m about to unplug for vacation for a week, if you’re wondering why I go silent after this for a while).

rljacob:
Ok that PR helped.  So its still a global domain but with a uniform forcing at every column.

PeterCaldwell:
yes

rljacob:
I feel like the flag for this option should have a more intuitive name but can't think of anything.

PR #3733, opened 2020-07-23
Make dtime calculation dependent on value of NCPL_BASE_PERIOD

jonbob:
This PR modifies the EAM namelist build system so that the coupling frequency it computes, in the form of variable dtime, is dependent on the value of NCPL_BASE_PERIOD that is set in env_run.xml. Previously it was hard-wired to assume the base period was one day and ATM_NCPL was the number of couplings per day. This update was necessary for getting IG cases working, where the NCPL_BASE_PERIOD is often a year so that the landice code can run on order of days.

Tested by @stephenprice during BG-case development

[BFB]


stephenprice:
@jonbob - I've confirmed this works as expected (specifically, for a test where the coupling base period is one year).

jonbob:
Thanks @singhbalwinder

jonbob:
merged to next

jonbob:
merged to master

Issue #3731, opened 2020-07-23
Build fail on theta with master (and likely maint-1.0)


ndkeen:
After theta maintenance, I'm seeing a build fail in pio.

```
In file included from /usr/include/math.h(312),
                 from /opt/intel/compilers_and_libraries_2018.0.128/linux/compiler/include/math.h(95),
                 from /projects/ClimateEnergy_4/ndkeen/wacmy/m07-jul23/externals/scorpio/src/clib/pio_internal.h(45),
                 from /projects/ClimateEnergy_4/ndkeen/wacmy/m07-jul23/externals/scorpio/src/clib/pio_timer.h(7),
                 from /projects/ClimateEnergy_4/ndkeen/wacmy/m07-jul23/externals/scorpio/src/clib/pio_timer.c(2):
/usr/include/bits/mathcalls-helper-functions.h(21): error: identifier "_Float128" is undefined
  __MATHDECL_1 (int, __fpclassify,, (_Mdouble_ __value))
  ^
```

`create_test SMS.ne4_ne4.FC5AV1C-L`

ndkeen:
Updating the compiler to intel v19 has allowed others past this issue.

However, with maint-1.0, we are still using intel18.

And, with maint-1.0, I'm also seeing he following build error (which is really cmake error). Not yet sure why I'm seeing this now and not the above error, but it could be that the above error is with master.

```
CMake Error at /projects/ClimateEnergy_4/ndkeen/E3SM_simulations/20200517-maint-1.0-tro.A_WCYCL20TRS_CMIP6_HR.ne120_oRRS18v3_ICG.unc11/build/intel/mpt/nodebug/threads/pio/pio2/\
CMakeFiles/CMakeTmp/cmTC_cc3b3Targets.cmake:16 (add_library):
  Target "MPI::MPI_C" links to item " -Xlinker --enable-new-dtags -Xlinker
  -rpath -Xlinker
  /opt/intel/compilers_and_libraries_2018.0.128/linux/mpi/intel64/lib/release_mt
  -Xlinker -rpath -Xlinker
  /opt/intel/compilers_and_libraries_2018.0.128/linux/mpi/intel64/lib
  -Xlinker -rpath -Xlinker /opt/intel/mpi-rt/2107.0.0/intel64/lib/release_mt
  -Xlinker -rpath -Xlinker /opt/intel/mpi-rt/2017.0.0/intel64/lib" which has
[...]

PR #3736, opened 2020-07-23
Update modules on Theta

amametjanov:
Update intel compiler on Theta to latest available default module after PM.

Fixes E3SM-Project/E3SM#3731

[non-BFB]

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -1568,7 +1568,7 @@
         <command name="load">cmake/3.14.5</command>
       </modules>
       <modules compiler="intel">
-        <command name="load">intel/18.0.0.128</command>
+        <command name="load">intel/19.0.5.281</command>
         <command name="load">PrgEnv-intel/6.0.5</command>
       </modules>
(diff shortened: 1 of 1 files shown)


dqwu:
@amametjanov 
It seems that netcdf4p type is not supported on Theta, since only serial NetCDF and HDF5 modules are loaded.
```
-- Checking NetCDF version - 4.7.3./*!<
-- Found NetCDF: /opt/cray/pe/netcdf/4.7.3.3/INTEL/19.1/lib/libnetcdf.a (found suitable version "4.7.3./*!<", minimum required is "4.3.3")
-- Checking whether NetCDF has parallel support
-- Checking whether NetCDF has parallel support - no
```
Should we use parallel version instead?
```
-        <command name="load">cray-hdf5/1.10.6.1</command>
-        <command name="load">cray-netcdf/4.7.3.3</command>
+        <command name="load">cray-hdf5-parallel/1.10.6.1</command>
+        <command name="load">cray-netcdf-hdf5parallel/4.7.3.3</command>
```

PR #3738, opened 2020-07-23
Simplify MMF aerosol optics

whannah1:
The initial implementation of the MMF in E3SM included the developments from PNNL to enhance the cloud-aerosol interactions, which including modifying the code that does the aerosol water uptake and wet radius that has a large impact on the aerosol optical properties. The modifications were done so that each CRM column would be considered and take advantage of the explicit cloud field information from the CRM. In our quest for a faster MMF that can compete with conventional climate models we found that the calculation of aerosol optical properties was a significant burden that cannot be offloaded to GPUs given the current framework. Therefore is was decided to simplify the aerosol optics by only calculating it once for the physics column and not for each CRM radiative column. Additionally, since the water uptake calculation is mainly done for the optics, we could revert those modifications by deleting the modified version of modal_aero_wateruptake.F90. This is not expected to affect the behavior of MMF-ECPP configurations.

[non-BFB] only for MMF


whannah1:
I've got a pair of 90 day ne4pg2 simulations using this branch with the 1-mom MMF. I'm also running some 2-mom cases, but they take much longer. I'll focus on zonal mean plots since they are just easy to look at.

blue = baseline
red = test with PR branch

First some basic water/cloud variables:
image

Next, some net radiative flux for TOM and sfc:
image

Looking at the aerosol optical depth we can see an overall increase. Since the MMF1 compset uses prescribed aerosol we can interpret this as a change in the amount of water uptake and corresponding change to the aerosol optics, rather than any change in the aerosol concentration. 
image

I think the aerosol optical depth increase might be expected because the estimated relative humidity is probably higher in cloud-free parts of CRM because the CAM version of modal_aero_wateruptake.F90 assumes 100% RH in the cloudy parts of the physics column, and if the CRM produces supersaturated cloudy regions, then this would cause the clear-air portions to have a lower RH given a fixed CRM mean RH. So if the RH is estimated to be higher I think it makes sense that the optical depth would go up. 

Ultimately, I think this change doesn't matter much for us, especially for 1-mom simulations. Larger AOD should probably be considered erroneous which will mean we'll have to tune the 2-mom microphysics to compensate when we eventually go to do climate simulations. On the other hand, this modification gives us a 6% boost in overall throughput, which is helpful for doing long climate experiments. 

Given that the MMF has other issues that are probably larger in magnitude relative to these aerosol radiative errors (i.e the precipitation issues from throttling and the checkerboard noise/variance trapping), I still think this is a good change.

whannah1:
I ran a pair of 1-month 2-mom tests (no ECPP) with this branch, and the results are mostly identical, similar to the plots above, but the AODVIS variable shows a larger difference in the 2-mom tests (dashed) compared to the 1-mom (solid):

image

It's also interesting that the AOD is much higher with 2-mom, but I don't have an explanation. The higher AOD with the simplified optics seems consistent with my argument above about having erroneously higher RH in the estimate used for the water uptake. 

To put the difference in a quantitative perspective, if I take the RMS difference of the time-mean AODVIS data, before doing the zonal mean, the value is roughly 20-30% of the global mean.

Despite the magnitude of these differences, I am still in favor of this change.

whannah1:
So after much discussion and experiments, the branch was refactored to calculate an average clear air relative humidity within the CRM. Previous attempts to get the clear air RH in each radiative column required a subjective cloud fraction threshold, but the AOD was strongly sensitive to how this was chosen. Aggregating the clear air RH only checks for the existence of condensate, consistent with how cloud fraction is determined. 

An important aspect of this change was that I refactored the aerosol section of tphysbc() so that the calls t0 modal_aero_calcsize_* and modal_aero_wateruptake_dr happen outside of aero_model_wetdep. Both of these routines are called at the top of aero_model_wetdep, and it seems the reason for this is that these calls are prerequisites for aero_model_wetdep, but it seems there's no reason the calls can't be moved outside of aero_model_wetdep. I feel that this is better than changing the aero_model_wetdep interface to pass the clear air RH down to modal_aero_wateruptake_dr, and it seems to simplify the code, but that's somewhat of a subjective conclusion.

While this new method is objectively better for the MMF, it actually still leads to systematically higher AOD (see figure below). When I investigated the clear air RH values coming out of the CRM there were several cases of super-saturation, which does not seem to be erroneous. Given that the modal_aero_wateruptake was already enforcing an upper limit of 98% RH I made sure the new option to provide the clear air RH also imposes this same limit for consistency.

The figure below shows the results of this method for 1-month runs:

image
 solid red  = MMF 1-mom control
 dashed red = MMF 1-mom PR branch
 solid blue = MMF 2-mom control
 dashed blue = MMF 2-mom PR branch

AOD seems to be the only thing that shows large differences. Other radiative quantities show no significant changes
image

I'm currently testing the changes to make sure the non-MMF cases are BFB.

whannah1:
The latest commits are an additional refactor of the aerosol physics section to put all calls to modal_aero_calcsize* and modal_aero_wateruptake_dr down into aero_model_wetdep(). This applies for cases with prognostic, bulk, or prescribed aerosols. I reran the MMF cases and the results are identical for 1-mom and show very small changes in the zonal mean AOD for 2-mom. 

On Cori I ran the e3sm_atm_developer suite and an additional aqua planet test:
SMS_Ln5.ne4pg2_ne4pg2.F-EAMv1-AQP1.cori-knl_intel
in order to test a case with prescribed aerosols, and all of these tests passed with no diffs. 

I think with these changes this branch is ready to be reviewed.

whannah1:
A little note about performance. From the limited timing tests I've done, it seems this change gives anywhere from 0-5% MMF speedup on the CPU depending on the configuration. The number of rad cols seems to be a big factor, as one would expect.

components/cam/src/physics/cam/physpkg.F90
```diff
@@ -39,8 +39,13 @@ module physpkg
   use cam_logfile,     only: iulog
   use camsrfexch,      only: cam_export
 
-  use modal_aero_calcsize,    only: modal_aero_calcsize_init, modal_aero_calcsize_diag, modal_aero_calcsize_reg
-  use modal_aero_wateruptake, only: modal_aero_wateruptake_init, modal_aero_wateruptake_dr, modal_aero_wateruptake_reg
+  use modal_aero_calcsize,    only: modal_aero_calcsize_init, &
+                                    modal_aero_calcsize_diag, &
```
singhbalwinder, line 43: I don't think we need `calcsize_diag` and `calcsize_sub` here as `calcsize_diag` has been moved and calcsize_sub is never called from physpkg.F90

components/cam/src/physics/cam/physpkg.F90
```diff
@@ -39,8 +39,13 @@ module physpkg
...
+                                    modal_aero_calcsize_reg, &
+                                    modal_aero_calcsize_sub
+  use modal_aero_wateruptake, only: modal_aero_wateruptake_init, &
+                                    modal_aero_wateruptake_dr, &
```
singhbalwinder, line 47: Since wateruptake_dr has been moved, we do not need it here in this use statement.

components/cam/src/physics/cam/physpkg.F90
```diff
@@ -1878,7 +1883,7 @@ subroutine tphysbc (ztodt,               &
                                check_tracers_data, check_tracers_init, &
                                check_tracers_chng, check_tracers_fini
     use dycore,          only: dycore_is
-    use aero_model,      only: aero_model_wetdep
+    use aero_model,      only: aero_model_wetdep, wetdep_lq
```
singhbalwinder, line 1886: `wetdep_lq` is not used here.

components/cam/src/physics/cam/physpkg.F90
```diff
@@ -1922,6 +1927,8 @@ subroutine tphysbc (ztodt,               &
     !
 
     type(physics_ptend)   :: ptend            ! indivdual parameterization tendencies
+    type(physics_ptend)   :: ptend_tmp1
+    type(physics_ptend)   :: ptend_tmp2
```
singhbalwinder, line 1931: Both ptend_tmp* are not used here (perhaps they were used during debugging....but not used anymore)

components/cam/src/chemistry/modal_aero/aero_model.F90
```diff
@@ -105,7 +105,7 @@ module aero_model
   integer :: nwetdep = 0
   integer,allocatable :: wetdep_indices(:)
   logical :: drydep_lq(pcnst)
-  logical :: wetdep_lq(pcnst)
+  logical, public :: wetdep_lq(pcnst)
```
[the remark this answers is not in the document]
whannah1: good catch, I forgot to delete this. It was from my original refactor which moved the calcsize routine outside of aero_model_wetdep, which required a ptend init.

components/cam/src/physics/cam/physpkg.F90
whannah1: good catch, I'll fix this.

whannah1: FYI, we actually still need this for the MMF when running with ECPP, but I should probably move it down into the ECPP driver. I'll leave that for when I get deeper into the ECPP refactor.

components/cam/src/physics/cam/physpkg.F90
whannah1: thanks, fixed.

components/cam/src/physics/cam/physpkg.F90
whannah1: Thanks, fixed.

components/cam/src/physics/cam/physpkg.F90
whannah1: You are correct, my bad, will fix.

components/cam/src/chemistry/modal_aero/aero_model.F90
```diff
@@ -1515,20 +1525,32 @@ subroutine aero_model_wetdep(dt, dlf, dlf2, cmfmc2, state,                    &!
...
-    call t_startf('calcsize')
-    ! for prognostic modal aerosols the transfer of mass between aitken and accumulation
-    ! modes is done in conjunction with the dry radius calculation
-    call modal_aero_calcsize_sub(state, ptend, dt, pbuf)
-    call t_stopf('calcsize')
-
-    call t_startf('wateruptake')
-    call modal_aero_wateruptake_dr(state, pbuf)
-    call t_stopf('wateruptake')
+    ! If not using prognostic aerosol call the diagnostic version
```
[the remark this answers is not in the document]
whannah1: It actually switches to the BAM version of aero_model_wetdep, but in either case I think you're right that's it's safe to remove this. I originally put it there because I wasn't 100% sure how things were behaving and thought it was safer this way, but I'm more confident that removing this is safe. So I'll remove it.

whannah1:
I rebased and force pushed to resolve a conflict with aerosol optics in the MMF radiation driver. Shouldn't have any affect on the non-MMF cases.

components/cam/src/chemistry/bulk_aero/aero_model.F90
```diff
@@ -600,12 +608,34 @@ subroutine aero_model_wetdep( dt, dlf, dlf2, cmfmc2, state, sh_e_ed_ratio,    &
...
     type(wetdep_inputs_t) :: dep_inputs  ! obj that contains inputs to wetdepa routine
 
+    integer           :: nmodes
+    logical           :: clim_modal_aero ! true when radiatively constituents present (nmodes>0)
+    logical           :: prog_modal_aero ! Prognostic modal aerosols present
+
+    ! set flags to check aerosol settings
+    call rad_cnst_get_info(0, nmodes=nmodes)
+    clim_modal_aero = (nmodes > 0)
+    call phys_getopts(prog_modal_aero_out=prog_modal_aero)
```
singhbalwinder, line 618: Sorry, I didn't notice it earlier but I think the above three lines belongs to the init routine. These logicals will not change during the simulation.

I can imagine how this works in case we have chem set to `none` (when we are using prescribed aerosols). I do not know how this plays out if somebody tries to revive BAM (I don't think anybody is working on it though). Unless there is a special case, I think this code is not even compiled when `prog_model_aero` or `clim_modal_aero` is true, so I wonder if we need these checks. May be we can try removing these checks and see if we have any unexpected failures. This is one way to just make sure that everything we care about is still working as expected.

components/cam/src/chemistry/modal_aero/aero_model.F90
```diff
@@ -1515,20 +1519,24 @@ subroutine aero_model_wetdep(dt, dlf, dlf2, cmfmc2, state,                    &!
...
-    ! for prognostic modal aerosols the transfer of mass between aitken and accumulation
-    ! modes is done in conjunction with the dry radius calculation
-    call modal_aero_calcsize_sub(state, ptend, dt, pbuf)
-    call t_stopf('calcsize')
+    ! If not using prognostic aerosol call the diagnostic version
 
-    call t_startf('wateruptake')
-    call modal_aero_wateruptake_dr(state, pbuf)
-    call t_stopf('wateruptake')
+    ! Calculate aerosol size distribution parameters
```
singhbalwinder, line 1528: May be we can add the timers back.... (`call t_startf('calcsize')` and `t_startf('wateruptake')`)

singhbalwinder:
Thanks Walter for addressing my comments! I have two more minor comment.

components/cam/src/chemistry/bulk_aero/aero_model.F90
whannah1: ok I can do that. The way I've implemented things here should preserve the previous behavior, even for BAM, so I don't think there's anything to worry about there. I'm still a little confused why these flags are used this way though. I would think that this all would have been built to specifically check whether the aerosol option was "bam" or "prescribed" or whatever, but these flags make it hard to understand what's going on.

components/cam/src/chemistry/modal_aero/aero_model.F90
whannah1: Sure, I figure they weren't that important, but I'll add them back.

components/cam/src/chemistry/bulk_aero/aero_model.F90
singhbalwinder: I agree. I wish we had clear logicals for each configurable option. Model is configurable in multiple ways so it is hard play out all the cases. We mostly care about cases for which we have the nightly tests. If all the tests are passing, we generally consider it safe to integrate. 

As far as I know, I do not think we need to checks for `prog_model_aero` and `clim_modal_aero` here but it doesn't hurt to add these.

components/cam/src/chemistry/modal_aero/aero_model.F90
singhbalwinder: Thanks! I was not so concerned about the "calsize_diag" calls but it is good to have timers for calcsize_sub and wateruptake.

singhbalwinder:
I think it is ready for integration.

brhillman:
Merged to master.

PR #3743, opened 2020-07-26
fix DYN_PS output

mt5555:
The call to OUTFLD() for DYN_PS (surface pressure on the dynamics grid) is made during the lagrangian floating vertical levels phase, where surface pressure  is not a prognostic variable (state%ps_v=0) and PS should be computed from state%dp3d

Also remove an unnecessary temp array in VOR/DIV output

[BFB] since we dont have any test cases which output DYN_PS

### DIFF
--- a/components/cam/src/dynamics/se/stepon.F90
+++ b/components/cam/src/dynamics/se/stepon.F90
@@ -403,24 +403,13 @@ subroutine stepon_run2(phys_state, phys_tend, dyn_in, dyn_out )
    if (hist_fld_active('VOR')) then
       call compute_zeta_C0(tmp_dyn,dyn_in%elem,par,tl_f)
       do ie=1,nelemd
-         !do j=1,np
-         !   do i=1,np
-         !      ftmp(i+(j-1)*np,:,1) = tmp_dyn(i,j,:,ie)
-         !   end do
-         !end do
-         !call outfld('VOR',ftmp(:,:,1),npsq,ie)
          call outfld('VOR',tmp_dyn(1,1,1,ie),npsq,ie)
       enddo
    endif
    if (hist_fld_active('DIV')) then
       call compute_div_C0(tmp_dyn,dyn_in%elem,par,tl_f)
       do ie=1,nelemd
-         do j=1,np
-            do i=1,np
-               ftmp(i+(j-1)*np,:,1) = tmp_dyn(i,j,:,ie)
-            end do
-         end do
-         call outfld('DIV',ftmp(:,:,1),npsq,ie)
+         call outfld('DIV',tmp_dyn(1,1,1,ie),npsq,ie)
       enddo
    endif
    if (hist_fld_active('FU') .or. hist_fld_active('FV') ) then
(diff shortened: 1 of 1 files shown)


AaronDonahue:
@mt5555 , who do you think would be a good reviewer for this PR?  Also, since you change the way ps_v is calculated for the DYN_PS output is it fair to say this will be a Non-BFB change?

whannah1:
@AaronDonahue I can review this. I would argue the PR is BFB since it only changes the history output, but maybe we could note that the history output is non-BFB?

components/cam/src/dynamics/se/stepon.F90
```diff
@@ -443,7 +432,12 @@ subroutine stepon_run2(phys_state, phys_tend, dyn_in, dyn_out )
       call outfld('DYN_Q'     ,dyn_in%elem(ie)%state%Q(:,:,:,1)       ,npsq,ie)
       call outfld('DYN_U'     ,dyn_in%elem(ie)%state%V(:,:,1,:,tl_f)  ,npsq,ie)
       call outfld('DYN_V'     ,dyn_in%elem(ie)%state%V(:,:,2,:,tl_f)  ,npsq,ie)
-      call outfld('DYN_PS'    ,dyn_in%elem(ie)%state%ps_v(:,:,tl_f)   ,npsq,ie)
+
+      ! at this point in the code we are on floating lagrangian levels
+      ! need to compute ps for output
+      p_i(:,:,nlevp) = hvcoord%hyai(1)*hvcoord%ps0 + &
```
whannah1, line 438: Seems like the calculation of p_i below is redundant with this here, so it's probably a good idea to remove the one lower down

AaronDonahue: Isn't the p_i calculation below inside of a compiler flag, so not always guaranteed?

whannah1: Right, but this part is not within an ifdef, so whether or not that section gets triggered by running with a non-theta-l configuration the values of p_i will already be calculated, so no need to calculate them twice.

mt5555:
Agree with the redundant p_i calculation but not sure which is best.   The one lower down is computing p_i at all levels, while the first one is only computing the surface pressure (p_i(:,:,nlevp)).   how about is the first one instead computed into an array names "ps"?

I wouldn't want to change the second calculation to only go to nlev and assume the nlevp calculation was done above, since that seems error prone w.r.t. future code cut & paste operations.

whannah1:
@mt5555 sorry, I missed that it's only the surface pressure that gets calculated. I guess we should just leave it as is.

AaronDonahue:
> @mt5555 , who do you think would be a good reviewer for this PR? Also, since you change the way ps_v is calculated for the DYN_PS output is it fair to say this will be a Non-BFB change?

@whannah1 , thanks for reviewing.  I'm thinking that we will want to label Non-BFB because it will require a new baseline.  Am I correct that DYN_PS is a default output?

AaronDonahue:
@mt5555 and @rljacob .  I'm going on vacation tomorrow for a week, would it be alright if I integrated this when I get back?

mt5555:
no rush - low priority.  DYN_PS is rarely used -mostly for debugging.

AaronDonahue:
@mt5555 , this has been merged to master now.  I think since it is in your fork you are responsible for deleting the branch, if you want to.

PR #3747, opened 2020-07-27
Update CIME submodule again

jgfouca:
Change list:
* Remove E3SM CIME subtree managent tools and libraries
* Add default ATM_NCPL=72 for ne45
* Modifications in MCT-driver to support MOSART stratification (adds  coszen field)
* Archive atm_chunk_costs with other performance data

Bug fixes:
* Use SRCROOT instead of CIMEROOT in key places

[BFB] - but adds a new field, so tests will DIFF


rljacob:
Adds a new field to what?   We're going to need more info on these submodule update PRs.

jgfouca:
@rljacob , done

rljacob:
I'm not sure all tests will diff.  Just ones with MOSART.

jgfouca:
@rljacob, my "A" compset tests were diffing

jonbob:
@jgfouca - that's expected if you're now passing a new field through the cpl. The A-cases dummy up data to pass based on the number of fields

jgfouca:
Thx @jonbob

rljacob:
That's the X-case.  The A-case is all data models but I think one of those was modified to send the new field even if there is no full MOSART model.

jonbob:
Sorry -- @rljacob is completely correct. My bad...

jgfouca:
> The A-case is all data models but I think one of those was modified to send the new field even if there is no full MOSART model.

@rljacob , @jonbob , is that something we need to change before merging this PR?

jonbob:
@jgfouca - what did the A-case DIFFs look like? Do you have the output from cprnc?

rljacob:
@jgfouca no you don't need to change anything.

jgfouca:
@jonbob, they just reported that the hist file had an extra field when compared to the baseline. There were no "real" diffs.

@rljacob , thanks, I will merge to master.

PR #3748, opened 2020-07-28
Modify PGI compiler flags to eliminate non-reproducibility on Compy

### LINKED ISSUE #3739: F case on Compy using PGI compiler is not BFB wrt change in PE layout
An F case 

     -res ne4pg2_ne4pg2 -compset FC5AV1C-L -mach compy -compiler pgi

is not not BFB with respect to changes in PE layout (64x1 vs. 96x1) with the current optimization flags. 

Experiments indicate that this can be resolved (for this case and experimental setup, at least) by setting the non-DEBUG flags for the whole model to -O1, in both Macros.cmake:

       if("${DEBUG}" STREQUAL "FALSE")
         set(FFLAGS "${FFLAGS}  -O1")
         set(CFLAGS "${CFLAGS}  -O1")
       endif()

and Macros.make:

       ifeq ($(DEBUG),FALSE)
         FFLAGS := $(FFLAGS)  -O1
         CFLAGS := $(CFLAGS)  -O1
      endif

Moreover, the '-O2' performance can be recovered without losing BFB by adding '-O2' back in for the 'cam' component only, i.e.. for Macros.cmake:

      if("${COMP_NAME}" STREQUAL "cam")
        set(FFLAGS "${FFLAGS} -O2 -Mnovect")
      endif()

and for Macros.make:

       ifeq ($(COMP_NAME),cam)
         FFLAGS := $(FFLAGS) -O2 -Mnovect
       endif

Further investigations may be able to identify which non-CAM files require -O1 for BFB behavior (in LND? CPL? driver?), but the above may be sufficient in the meantime, as it is performant while recovering BFB. Similar epxeriments may be required to identify what compiler flags work for the other active components (OCN, ICE, ...)  in a fully coupled model run.

(@jgfouca , if the problem is found in the driver or share code, is this a CMAKE buld or non-CMAKE? If non-CMAKE, is there something equivalent to e3sm_remove_flags that works in a non-CMAKE Depends file?)

### LINKED ISSUE #3746: F case on Compy using PGI is not BFB wrt compile-time vs. runtime PCOLS
F cases on Compy when using the PGI compiler are not reproducible with respect to setting pcols at runtime (the default) as compared to setting pcols at compile-time, via the -pcols otpion in CAM_CONFIG_OPTS in env_builds.xml.

Earlier experiments indicated that thie issue is resolved by building the whole model using -O1. This has not been adopted, and the default, specified in Macros.cmake, is still -O2 .

Recent experiment indicate that this issue can be resolved by removing -O2 and replacing it with -O1 for three EAM files:

       cam/src/chemistry/aerosol/dust_sediment_mod.F90
       cam/src/chemistry/modal_aero/modal_aero_convproc.F90
       cam/src/physics/cam/zm_conv.F90

With the fix to e3sm_remove_files in PR #3705 , this logic can now be implemented in a Depends.compy.pgi.

worleyph:
a) An F case

      -res ne4pg2_ne4pg2 -compset FC5AV1C-L -mach compy -compiler pgi

is not BFB with respect to changes in PE layout, for example,
64x1 compared to 96x1, with the current optimization flags.
Reproducibility can be restored by compiling the CICE component
with the flag -Mnovect. This is implemented by adding the rule

         <append COMP_NAME="cice">-Mnovect</append>

to config_compilers.xml, which adds

       if("${COMP_NAME}" STREQUAL "cam")
         set(FFLAGS "${FFLAGS} -Mnovect")
       endif()

to Macros.cmake and

       ifeq ($(COMP_NAME),cice)
         FFLAGS := $(FFLAGS) -Mnovect
       endif

Macros.make in a case directory.

b) F cases on Compy when using the PGI compiler are not reproducible with
respect to setting pcols at runtime (the default) as compared to
setting pcols at compile-time, via the -pcols option in
CAM_CONFIG_OPTS in env_builds.xml.

This issue is resolved by lowering the optimization from -O2 to -O1
for three files

   cam/src/chemistry/aerosol/dust_sediment_mod.F90
   cam/src/chemistry/modal_aero/modal_aero_convproc.F90
   cam/src/physics/cam/zm_conv.F90

This is implemented in a Depends file for compy
when using the PGI compiler (Depends.compy.pgi and
Depends.compy.pgi.cmake),

Fixes #3739 
Fixes #3746 

[Non-BFB] (for Compy jobs using PGI)

### DIFF
--- a/cime_config/machines/Depends.compy.pgi
+++ b/cime_config/machines/Depends.compy.pgi
@@ -1,0 +1,10 @@
+# Files that cannot be compiled with -O2 without losing reproducibility
+O1MODELSRC=\
+dust_sediment_mod.o \
+modal_aero_convproc.o \
+zm_conv.o
+
+ifeq ($(DEBUG),FALSE)
+  $(O1MODELSRC): %.o: %.F90
+	  $(FC) -c $(INCLDIR) $(INCS) $(FFLAGS) $(FREEFLAGS) -O1 -Mnovect $<
+endif
--- a/cime_config/machines/Depends.compy.pgi.cmake
+++ b/cime_config/machines/Depends.compy.pgi.cmake
@@ -1,0 +1,13 @@
+# Files that cannot be compiled with -O2 without losing reproducibility
+set(O1MODELSRC
+  cam/src/chemistry/aerosol/dust_sediment_mod.F90
+  cam/src/chemistry/modal_aero/modal_aero_convproc.F90
+  cam/src/physics/cam/zm_conv.F90)
+
+if (NOT DEBUG)
+  foreach(ITEM IN LISTS O1MODELSRC)
+    e3sm_remove_flags("${ITEM}" "-O2")
+    e3sm_add_flags("${ITEM}" "-O1 -Mnovect")
+  endforeach()
+
+endif()
--- a/cime_config/machines/config_compilers.xml
+++ b/cime_config/machines/config_compilers.xml
@@ -1069,6 +1069,7 @@ flags should be captured within MPAS CMake files.
     <append DEBUG="FALSE"> -O2 </append>
     <append DEBUG="TRUE">-C -Mbounds -traceback -Mchkfpstk -Mchkstk -Mdalign  -Mdepchk  -Mextend -Miomutex -Mrecursive  -Ktrap=fp -O0 -g -byteswapio -Meh_frame</append>
     <append COMP_NAME="cam">-Mnovect</append>
+    <append COMP_NAME="cice">-Mnovect</append>
   </FFLAGS>
   <NETCDF_PATH> $ENV{NETCDF_HOME}</NETCDF_PATH>
   <PIO_FILESYSTEM_HINTS>lustre</PIO_FILESYSTEM_HINTS>


worleyph:
@amametjanov , this resolves the issues for the nonreproducible reproducers that I am aware of. We could try to identify what is triggering the problem in the 3 identified files, but I am not sure it is that important since we are not using PGI for production runs on Compy. I did not even try to identify which files are at fault (requiring -Mnovect) in CICE, given that we added -Mnovect for all of CAM (and most of the data components) without trying to identify specific files, so this is consistent with earlier decisions.

worleyph:
If/when we change the version of the PGI compiler on Compy, we will need to revisit these reproducers, to make sure that the problems do not reoccur (given that the existing test suites do not see these issues).

amametjanov:
Checked with the same tests as in this https://github.com/E3SM-Project/E3SM/issues/3739#issuecomment-663672151. Runs, which were not BFB before, are BFB now.

Incidentally, threaded debug jobs (`SMS_P96x2_D.ne4_ne4.FC5AV1C-L.compy_pgi` and `SMS_P96x2_D.ne4pg2_ne4pg2.FC5AV1C-L.compy_pgi`) are running into out-of-bounds error:
```
42: 0: Subscript out of range for array uvm (/qfs/people/mame117/repos/E3SM-tests/components/cice/src/source/ice_grid.F90: 1635)
42:     subscript=2, lower bound=4, upper bound=4, dimension=1
```
Not seeing any obvious errors there. Just noting.

PR #3751, opened 2020-07-29
Fix RRTMGP aerosol optical depth

brhillman:
Fix missing delta-scaling calculation in aerosol optical depth that is input to RRTMGP, and fix band indices used for diagnostic aerosol optical property outputs. The delta-scaling calculation accounts for forward scattering from aerosol in the shortwave flux computations, and *does* affect the climate (to the tune of ~0.12 W/m^2 net TOA flux in a 10 year AMIP-style run at ne30). Omission of this calculation was simply an oversight. The diagnostic band bug arose because RRTMG and RRTMGP have different band ordering, and in order to continue using the existing RRTMG optical property look-up tables, we simply re-order the spectral dimension after computing the optics using the old look-up tables. However, the output of the aerosol optical properties happens in some routines buried a few levels down in the aerosol optics code, which happened before we had a chance to re-order the spectral dimension, so the diagnostic band indices we were using (which were setup to be consistent with the new RRTMGP band ordering) were inconsistent with the order of the optical properties at this point in the code, so we were outputting the wrong bands. In order to fix this, we now assume the RRTMG band ordering throughout the E3SM side of the code, and only re-order at the last moment (after all diagnostic outputs have finished). The band index fix is actually BFB, other than for the purely diagnostic aerosol optical property outputs (such as AODVIS, the 500 nm wavelength aerosol optical depth). The first fix brings the net TOA flux into closer agreement with RRTMG, although differences are still expected due to differences in the solar spectral flux and differences in atmospheric absorption. The second fix brings the visible wavelength optical depth into much better agreement with RRTMG as well. @yfenganl has documented the effects of these changes on Confluence [here](https://acme-climate.atlassian.net/wiki/spaces/EWCG/pages/1659240982/20200701.v1like-fixer2.F2010.ne30pg2+r05+oECv3.compy). Will affect any atmosphere configuration using RRTMGP, including those using MMF. Expected diffs:

ERS_Ld5.ne4_ne4.FC5AV1C-L.cam-rrtmgp
ERS_Ln9_P96x1.ne4pg2_ne4pg2.F-MMF1.cam-crmout

[non-BFB]

### DIFF
--- a/components/cam/src/physics/crm/rrtmgp/radiation.F90
+++ b/components/cam/src/physics/crm/rrtmgp/radiation.F90
@@ -18,8 +18,9 @@ module radiation
    use rad_constituents, only: N_DIAG
    use radconstants,     only: &
       nswbands, nlwbands, &
-      set_sw_spectral_boundaries, set_lw_spectral_boundaries, &
-      get_sw_spectral_midpoints, get_lw_spectral_midpoints
+      get_band_index_sw, get_band_index_lw, test_get_band_index, &
+      get_sw_spectral_midpoints, get_lw_spectral_midpoints, &
+      rrtmg_to_rrtmgp_swbands
    use cam_history_support, only: add_hist_coord
 
    ! RRTMGP gas optics object to store coefficient information. This is imported
@@ -505,10 +506,6 @@ contains
       ! by the absorption coefficient data.
       nswgpts = k_dist_sw%get_ngpt()
       nlwgpts = k_dist_lw%get_ngpt()
-
-      ! Set band intervals based on inputdata
-      call set_sw_spectral_boundaries(k_dist_sw%get_band_lims_wavenumber())
-      call set_lw_spectral_boundaries(k_dist_lw%get_band_lims_wavenumber())
 
       ! Set number of levels used in radiation calculations
 #ifdef NO_EXTRA_RAD_LEVEL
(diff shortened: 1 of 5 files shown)


components/cam/src/physics/crm/rrtmgp/radiation.F90
```diff
@@ -2011,9 +2037,6 @@ subroutine calculate_fluxes_sw(gas_names, gas_vmr, &
       end do
       deallocate(gas_names_lower)
 
-      call handle_error(cld_optics_day%validate())
-      call handle_error(aer_optics_day%validate())
```
whannah1, line 2015: Were these just redundant?

components/cam/src/physics/crm/rrtmgp/radiation.F90
```diff
@@ -2656,6 +2682,31 @@ end subroutine output_fluxes_lw
 
    !----------------------------------------------------------------------------
 
+   ! Utility function to reorder an array given a new indexing
+   function reordered(array_in, new_indexing) result(array_out)
```
whannah1, line 2686: I'm wondering if calling this something like "reorder_1D" would be a better name in case you need a 2D version later on?

whannah1: Also, shouldn't something this generic be in a separate "radiation_utils.F90" module?

brhillman: In the event I need a 2d version I'd overload it. Prefer to keep the name as-is for now until we need that, because the interface would stay the same.

components/cam/src/physics/crm/rrtmgp/radiation.F90
brhillman: RRTMGP does this internally anyways. This was just an extra check added at some point in the debugging process to catch where things were going awry.

brhillman:
@whannah1 I don't think it belongs in another module since it's just a one-off routine that's only going to be used in the module it's defined it. Prefer to keep that all self-contained.

whannah1:
@brhillman sounds good.

singhbalwinder:
merged to next

PR #3754, opened 2020-07-31
Allow MOSART to read/write in unstructured grid format

donghuix:
Following changes are made:

1. If there are lat/lon dimensions in the input file, MOSART will read/write in a structured grid format. If the dimension is `gridcell`, MOSART will read/write in an unstructured grid format.

2. Put elevation profiles into one input variable.

3. Adds a test of MOSART for an unstructured global domain. 

[BFB]

### DIFF
--- a/components/mosart/src/riverroute/RtmMod.F90
+++ b/components/mosart/src/riverroute/RtmMod.F90
@@ -4084,6 +4084,7 @@ contains
     enddo
 
     deallocate(ele)
+    call pio_freedecomp(ncid, iodesc)
 
   end subroutine read_elevation_profile
 
--- a/components/mosart/src/riverroute/RtmMod.F90
+++ b/components/mosart/src/riverroute/RtmMod.F90
@@ -3546,8 +3546,7 @@ contains
           allocate (TUnit%wr_bf(begr:endr))
           TUnit%wr_bf = 0.0_r8   
 
-          allocate( TUnit%e_eprof_in2( 11, begr:endr ) )    
-          !allocate( e_eprof_in2( 11, begr:endr ) )  
+          allocate( TUnit%e_eprof_in2( Tctl%npt_elevProf, begr:endr ) )    
 
           ! --------------------------------- 
           ! (assign elevation values to TUnit%e_eprof_in2( :, : ) ).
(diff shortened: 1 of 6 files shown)


components/mosart/src/riverroute/RtmMod.F90
```diff
@@ -3515,63 +3547,12 @@ subroutine MOSART_init
           TUnit%wr_bf = 0.0_r8   
 
           allocate( TUnit%e_eprof_in2( 11, begr:endr ) )    
+          !allocate( e_eprof_in2( 11, begr:endr ) )
```
bishtgautam, line 3550: Remove this commented out line.

components/mosart/src/riverroute/RtmMod.F90
```diff
@@ -3515,63 +3547,12 @@ subroutine MOSART_init
           TUnit%wr_bf = 0.0_r8   
 
           allocate( TUnit%e_eprof_in2( 11, begr:endr ) )
```
bishtgautam, line 3549: Memory allocation here is assuming that there will be at most 11 layers. But, in [`read_elevation_profile`](https://github.com/E3SM-Project/E3SM/pull/3754/files#diff-659dea48e8aa2bc7dd7a44a748a8e357R4083) the number of layers could greater than 11. Thus, move the memory allocation of `TUnit%e_eprof_in2` within `read_elevation_profile`.

bishtgautam:
@rljacob, This PR is based on the work done under the ICoM project. Are you are okay with bringing in model developments from the ICoM project in E3SM master?

cc: @eclare108213

components/mosart/src/riverroute/RtmMod.F90
donghuix: Ok.

components/mosart/src/riverroute/RtmMod.F90
donghuix: There is pre-defined number of elevation profile: Tctl%npt_elevProf = 11, which is read from mosart_in. It is also used to allocate other elevation profile related variable memory. 

How about we add a check in the subroutine read_elevation_profile that if the number of points in the elevation profile is not consistent with Tctl%npt_elevProf then the program stop.

rljacob:
Yes as long as it doesn't change answers for v2 configs.

donghuix:
@bishtgautam Please the new commit for the modification:

1.  Removed the unnecessary comment out line.
2. Use Tctl%npt_elevProf to allocate memory, and read the dim of elevation profile and stop the program if the size is not consistent.
3. If 'ele' is not given in the input file, search 'ele0', 'ele1' etc. to read the elevation profile. If both 'ele' and 'ele0' are not in the input file, report an error.

PR #3756, opened 2020-08-03
FATES API 13

rgknox:
This set of changes updates E3SM compatibility with FATES API 13. 
Noteable changes are interface hooks for fixed biogreography, external 
lightning datasets and land-use datasets. The first two interface changes
 are just hooks, and will forgo functionality until work is allocated to do so. 
The land-use dataset read functionality is forthcoming.

[non-BFB]

### DIFF
--- a/components/clm/src/main/clm_varctl.F90
+++ b/components/clm/src/main/clm_varctl.F90
@@ -213,7 +213,9 @@ module clm_varctl
   !----------------------------------------------------------
 
   logical, public            :: use_fates = .false.              ! true => use  ED
-  logical, public            :: use_fates_spitfire = .false.  ! true => use spitfire model
+  integer, public            :: fates_spitfire_mode = 0                ! 0 for no fire; 1 for constant ignitions
+  logical, public            :: use_fates_fixed_biogeog = .false.           ! true => use fixed biogeography mode
+  logical, public            :: use_fates_spitfire = .false.           ! true => use spitfire model
   logical, public            :: use_fates_logging = .false.            ! true => turn on logging module
   logical, public            :: use_fates_planthydro = .false.         ! true => turn on fates hydro
   logical, public            :: use_fates_cohort_age_tracking = .false. ! true => turn on cohort age tracking
--- a/components/clm/src/main/clm_varctl.F90
+++ b/components/clm/src/main/clm_varctl.F90
@@ -215,7 +215,6 @@ module clm_varctl
   logical, public            :: use_fates = .false.              ! true => use  ED
   integer, public            :: fates_spitfire_mode = 0                ! 0 for no fire; 1 for constant ignitions
   logical, public            :: use_fates_fixed_biogeog = .false.           ! true => use fixed biogeography mode
-  logical, public            :: use_fates_spitfire = .false.           ! true => use spitfire model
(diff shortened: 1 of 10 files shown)


rgknox:
Ran land developer suite, looked good on lawrencium. All b4b except a mozart test that failed, which I assume is expected.
Following those tests, I added in changes to how we send time info to fates, that were orininally authored by @glemieux, but somehow got dropped (or just forgot to add them), back in.  New-test are running and looking good so far.  It is possible this change might affect results very subtly (fates tests only).

rgknox:
second test is PASS.  No further updates, let me know if I can provide anything else for integration.

*Note: these changes also now point us to a version of FATES that includes changes by @amametjanov , which get E3SM FATES running on cray compilers

components/clm/src/main/clmfates_interfaceMod.F90
```diff
@@ -211,6 +211,12 @@ module CLMFatesInterfaceMod
    ! developer will at least question its usage (RGK)
    private :: hlm_bounds_to_fates_bounds
 
+   ! The GetAndSetTime function is used to get the current time from the CLM 
+   ! time procedures and then set to the fates global time variables during restart, 
+   ! init_coldstart, and dynamics_driv function calls
```
glemieux, line 216: The `init_coldstart` subroutine should have `GetAndSetTime` added to it's begging as well.

glemieux: It should get called between these two lines: https://github.com/E3SM-Project/E3SM/blob/e277f97cef0043babcb4aa87ee01da976c2c7637/components/clm/src/main/clmfates_interfaceMod.F90#L1320-L1322

rgknox: thanks, will add

rgknox:
updated and re-tested, tests are nominal

components/clm/src/main/clm_varctl.F90
```diff
@@ -213,7 +213,9 @@ module clm_varctl
   !----------------------------------------------------------
 
   logical, public            :: use_fates = .false.              ! true => use  ED
-  logical, public            :: use_fates_spitfire = .false.  ! true => use spitfire model
+  integer, public            :: fates_spitfire_mode = 0                ! 0 for no fire; 1 for constant ignitions
+  logical, public            :: use_fates_fixed_biogeog = .false.           ! true => use fixed biogeography mode
+  logical, public            :: use_fates_spitfire = .false.           ! true => use spitfire model
```
bishtgautam, line 218: It seems that `use_fates_spitfire` has been replaced by `fates_spitfire_mode`. If true, can `use_fates_spitfire` be deleted?

rgknox: yes, good catch @bishtgautam

rgknox: updated code

bishtgautam:
@rgknox Is this PR BFB for the FATES test?

rgknox:
no, too much has changed on the fates side, I would not expect the same answers (although I suppose its possible)

bishtgautam:
In that case, I will mark this as a non-BFB

PR #3762, opened 2020-08-05
Make sure land surface data is set for r0125 grid

brhillman:
Fix setting of fsurdat for `sim_year=2000` and `hgrid=r0125`. This was previously unset, so resolutions like 
`ne30_r0125_ne30` with compset `FC5AV1C-L` would fail at setup. Partially fixes an issue with high resolution 
SCREAM compsets downstream.

[BFB]

### DIFF
--- a/components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml
+++ b/components/clm/bld/namelist_files/namelist_defaults_clm4_5.xml
@@ -265,6 +265,8 @@ lnd/clm2/surfdata_map/surfdata_ne4pg2_simyr2000_c190620.nc</fsurdat>
 lnd/clm2/surfdata_map/surfdata_ne240np4_simyr2000_c170821.nc</fsurdat>
 <fsurdat hgrid="r05"         sim_year="2000" >
 lnd/clm2/surfdata_map/surfdata_0.5x0.5_simyr2000_c190418.nc</fsurdat>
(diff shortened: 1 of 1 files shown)


brhillman:
Note that there is a setting for `hgrid=0.125x0.125`, which as far as I can tell is the same grid but incorrectly specified. In `config_grids.xml`, this grid does not exist, but instead seems to be referred to as `r0125`. So I suspect that maybe the `0.125x0.125` entry (and the others like it) should be removed. Instead though, I just added an entry for `r0125`. This can easily be changed.

bishtgautam:
@brhillman My guess is this PR BFB but can you confirm that it is BFB?

brhillman:
@bishtgautam I don't see how it wouldn't be but I can run e3sm_developer, although I don't see how that would actually catch anything here since I don't think any tests are using r0125.

bishtgautam:
I agree with you it BFB as we don't have test for it.

brhillman:
@bishtgautam any thoughts on removing the `hgrid=0.125x0.125` configurations? Are these used anywhere? Based on my reading of the configuration files it appears to me that this was meant to be r0125 instead of 0.125x0.125.

bishtgautam:
@brhillman Apologies as this PR fell off my radar. I will merge this PR at the next available opening. I agree with you that `hgrid=0.125x0.125` should be replaced by `hgrid=r0125xr0125`.

PR #3764, opened 2020-08-06
add support for hydrostatic mode even when using IMEX timestepping

mt5555:
Allow one to run both NH and H cases with tstep_type=9 (IMEX) method

For H model, slightly less efficient than tstep_type=5 and only agrees to roundoff

[BFB]

### DIFF
--- a/components/homme/src/theta-l/imex_mod.F90
+++ b/components/homme/src/theta-l/imex_mod.F90
@@ -137,6 +137,18 @@ contains
 
     call t_startf('compute_stage_value_dirk')
 
+    if (theta_hydrostatic_mode) then
+       itercount=0
+       itererr=0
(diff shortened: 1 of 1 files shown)


oksanaguba:
why is it needed?

mt5555:
it lets us test the explicit table of all the IMEX methods without having to add another tstep_type.
Also, SCREAM could make tstep_type=9 the default for all simulations, H or NH.

PeterCaldwell:
If I understand correctly, @paullric is quite keen on this capability because he wants to test the impact of H vs NH with nothing else varying.

rljacob:
@oksanaguba is this ready to merge?

oksanaguba:
in next

PR #3765, opened 2020-08-07
Addition of daily sea ice output option consistent with E3SM HR and switch to single precision

proteanplanet:
This pull request adjusts sea ice output in preparation for E3SM V2.  It includes:

1) Addition of a daily sea ice output stream that is the same as being used in E3SM HR simulations (previously there was no preset daily output stream available within E3SM for MPAS-SeaIce).  Please note that daily sea ice output is switched off by default even with this option available, and the daily stream can be turned on by setting config_am_timeseriesstatsdaily_enable = true in MPAS-SeaIce (it is false by default).

2) Switch the monthly sea ice model output to single precision. 

This branch has been tested on Cori-KNL in B-Case tests and works as prescribed. 

[BFB]


jonbob:
@vanroekel, @maltrud, @akturner - could you all please review this whenever you get a chance?

rljacob:
@maltrud and @akturner please review.

akturner:
Approved

akturner (approved):

maltrud (approved):

jonbob:
@xylar - can you please confirm that this will be OK with MPAS-Analysis? Thanks

jonbob:
merged to next

jonbob:
merged to master

PR #3767, opened 2020-08-09
Improve performance of print_cost_p routine

### LINKED ISSUE #3766: output of chunk costs (print_cost_p) is expensive for large process counts
The runtime of the routine print_cost_p, which writes out the cost of each chunk to the file atm_chunk_costs.txt, is large for large process counts on Cori-KNL. For example, for an F case using the ne256pg2 mesh and 32768 processes, this output took 18 minutes to complete. As print_cost_p is now called by default, this issue needs to be addressed quickly.

Given the severity of the performance problem, I have applied a "bug" label to this issue, even though it is functionally correct.

An alternative implementation has been developed, decreasing the runtime of print_cost_p for the above mentioned case from 18 minutes to less than 6 seconds. A PR with this change will be submitted soon.

worleyph:
The current algorithm used in print_cost_p has each process, in turn,
open the file atm_chunk_cost.txt, write out the measured cost
assocated with each chunk to the end of the file, and then close the
file. This approach is efficient with respect to memory usage, but is
costly in terms of runtime for large process counts on systems such as
Cori-KNL at NERSC. For example, for an F case using the ne256pg2 
mesh and 32768 processes, this algorithm took 18 minutes to complete.

Here an alternative algorithm is implemented. All data
to be written out is first gathered to the ATM masterproc. The
file atm_chunk_cost.txt is then opened/created once, the data
is written out by masterproc, and the file is then closed. The memory
requirements are higher with this approach, but, at scale, this
alternative approach has been observed to be almost 300 times faster
on Cori-KNL, decreasing the runtime of print_cost_p for the previously 
mentioned case to less than 6 seconds.

Some timers are also added, to better attribute where time is
being spent during the finalization of the ATM component.

Fixes #3766 

[BFB]

### DIFF
--- a/components/cam/src/control/cam_comp.F90
+++ b/components/cam/src/control/cam_comp.F90
@@ -455,11 +455,18 @@ subroutine cam_final( cam_out, cam_in )
    real(r8) :: mpi_wtime
 #endif
 
+   call t_startf ('phys_final')
    call phys_final( phys_state, phys_tend , pbuf2d)
+   call t_stopf ('phys_final')
+
+   call t_startf ('stepon_final')
    call stepon_final(dyn_in, dyn_out)
+   call t_stopf ('stepon_final')
 
    if(nsrest==0) then
+      call t_startf ('cam_initfiles_close')
       call cam_initfiles_close()
+      call t_stopf ('cam_initfiles_close')
    end if
 
    call hub2atm_deallocate(cam_in)
--- a/components/cam/src/cpl/atm_comp_mct.F90
+++ b/components/cam/src/cpl/atm_comp_mct.F90
@@ -644,7 +644,9 @@ CONTAINS
     type(mct_aVect)             ,intent(inout) :: x2a_a
     type(mct_aVect)             ,intent(inout) :: a2x_a
 
+    call t_startf("cam_final")
     call cam_final( cam_out, cam_in )
+    call t_stopf("cam_final")
 
   end subroutine atm_final_mct
 
--- a/components/cam/src/physics/cam/phys_grid.F90
+++ b/components/cam/src/physics/cam/phys_grid.F90
@@ -97,7 +97,8 @@ module phys_grid
                                local_dp_map
    use mpishorthand
 #endif
-   use spmd_utils,       only: iam, masterproc, npes, proc_smp_map, nsmps
+   use spmd_utils,       only: iam, masterproc, masterprocid, &
+                                    npes, proc_smp_map, nsmps
    use m_MergeSorts,     only: IndexSet, IndexSort
    use cam_abortutils,   only: endrun
    use perf_mod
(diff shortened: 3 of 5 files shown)


worleyph:
I ran a number of tests on Cori-KNL during development, and ran one sanity check using this branch.

worleyph:
Since print_cost_p is executed by default, getting this in soon is important for people running high resolution (large process count) cases on Cori-KNL using master. However, execution of print_cost_p can also be disabled by setting

     phys_chnk_cost_write = .false.

in user_nl_cam in the meantime. Production type jobs are unlikely to notice much impact.

worleyph:
A subsequent algorithmic change may still be coming, in a separate PR, combining lower memory requirements with still only masterproc writing out the data. However, it is not a given that this alternative approach will perform well, and I wanted to get a reasonable fix in as soon as possible. The algorithm in the current PR is a strong candidate to be the final solution.

sarats:
Looks good, thanks for adding all the detailed timers.

worleyph:
@sarats , I finished the "low memory version". It does not collect all data on process 0 before writing it out, but rather receives/writes out data for each process, allocating just enough space to hold two processes worth of remote data (double buffering so that can overlap writing with receive of data from next process). It is a little bit slower (single experiments on Cori-KNL, so could be subject to perturbation, but still likely qualitatively correct):

     ne4pg2_ne4pg2 using 96 processes on 32 nodes: 0.031 seconds vs. 0.015 seconds
     ne30pg2_ne30pgw using 1350 processes on 41 nodes: 0.341 seconds vs. 0.208 seconds
     ne120_ne120 using 5400 processes on 169 nodes: 1.187 seconds vs. 0.777 seconds
     ne256pg2_ne256pg2 using 65536 processes on 1024 nodes: 6.011 seconds vs. 5.121 seconds

so around twice as expensive for small process counts, but still less than 2 seconds, and only 20% slower for the largest experiment, where larger memory requirements and scaling of MPI_Allgather are likely to start becoming an issue. (And this is still around 6 seconds, as compared to 1200 seconds for the original algorithm).

My inclination is to have this PR go in as is, to capture the "fast" version in the repository, and then submit a subsequent PR with the low memory algorithm, as an example of "best practice" (to not introduce even more nonscalable memory usage), even though the nonscalable memory is in the noise compared to the memory usage in the rest of the model with current resolutions and process counts.

Do you agree? Or should only the low memory version go in (and I should update this PR) or only the fast version should go in (and I should forget about the low memory version)?

sarats:
I agree with putting in the "fast" version first. 

Skimming over the code, looks like we are aggregating 5 integers (i4) and 4 double precision (r8) values (total 52 bytes)  per chunk and aggregating over all chunks on the root process. So memory usage seems like O(nchunks*52*k) where k=10 accounts for MPI buffers and statuses etc.?
Even for high processor count like 100k, it's probably <50MB which shouldn't cause problems as we are not pushing memory capacity in our problem configurations.

Thanks to the detailed timers, we can disable via `phys_chnk_cost_write = .false.` if it ever becomes a bottleneck (memory-wise or expensive MPI collectives).

As you already finished the low-memory version, it would be good to add that as a best practice. I will leave it to your discretion based on any further effort that's needed. There is inherent serialization? (writing one process data as received) in the low-memory version, so thinking if it makes sense to have both options so that we can take advantage of a better collective implementation/interconnect.

worleyph:
Thanks @sarats . @singhbalwinder , guess this one is ready to be integrated.

singhbalwinder:
Merged to next

PR #3770, opened 2020-08-13
Fix ARKODE-HOMME interface bugs v2

oksanaguba:
This fixes a bug in the Jacobian calculation in the ARKODE-HOMME interface and adds ifndef statements to avoid calling the limiter and applying warnings when ARKODE is used.

[bfb]


oksanaguba:
this is a copy of PR #3522 where I assume cime was not merged in properly.

oksanaguba:
in next

PR #3771, opened 2020-08-14
Land Use Translator update for E3SM land surface files

aldivi:
This PR updates the land use translator (LUT) to be consistent between iESM and the ELM tool, fixes some bugs, and updates the LUT processing. The new LUT generates identical results regardless of machine (mac, cori, compy), compiler (gcc or intel), or compiler optimization (0 to 3). The LUT file is identical for iESM and for ELM tool, with a STANDALONE pre-processing definition enabled by default for the ELM tool. Initial data file for the LUT have been put into the .../rawdata folder on NERSC and COMPY to facilitate using the LUT. I anticipate generating LUT input files for all 8 cmip6 future scenarios. Additional updates to the mksurfdata makefile eliminate the current line edit, and additions to the namelist_defaults_clm4_5_tools.xml file allow for all 8 cmip6 future scenarios to be potentially be created.

The baseline historical surface data and two future scenario (SSP5RCP85, SSP2RCP45) land surface data files have been generated for ne30np4 resolution using the new LUT, including fsurdat files for 1850 and 2015, and these data look good. A 4-year ELM test run using the historical data has successfully been conducted on Cori.

A new documentation page is also published at https://acme-climate.atlassian.net/wiki/spaces/ED/pages/1708589675/Creating+land+surface+datasets+for+E3SM
Should there be additional directions for putting newly created surface files in a more central data location?


bishtgautam:
@aldivi The changes in FATES need to first go into FATES repo and then they will be brought in E3SM.

cc: @rgknox @glemieux

rgknox:
Also, @aldivi (sorry if this was already done/discussed), but do these new LUT files need to be loaded to the server (blues?)
e.g.:
https://github.com/E3SM-Project/E3SM/pull/3771/files#diff-2363f74581ed9707b701ce58dd0f8f15R324

bishtgautam:
@rgknox It is fine to make a new commit that updates the FATES submodule to the appropriate hash.

aldivi:
This is independent of FATES. The FATES submodule in this branch is an older one because it is the one that was working when I started. I may need to figure out a better local repo system because whenever I change or update local branches the FATES connection breaks.

I changed the PR title.

aldivi:
Also, if Blues is the main server for the data files, then yes, the data files need to be put there. I can do this, although I am not sure whether I have access.

rljacob:
@aldivi if you put the files on NERSC or compy, someone with access can put them on blues.

aldivi:
The files are on NERSC and Compy, in their respective locations. I will have to compile a list for whomever would make the transfer. I am not sure whether the IAC component input file path for iESM exists on Blues.

bishtgautam:
@aldivi Are the updated LULC files expected to be used in the v2 watercycle campaign? If yes, is the new transient LULC file for ELM already being used by the v2 watercycle campaign?

cc: @golaz

aldivi:
The only one of the new files that is currently being used is the SSP2RCP45 surface file for the COVID sims, and this is probably with v1.

Since the v2 code freeze hasn't happened yet, I didn't think any v2 sim campaigns had started.

I would expect all v2 sims to use the new files, as a new v2 1850 spinup (and historical transient) will have to be generated, which can use the new files.

bishtgautam:
@aldivi You are correct the v2 watercycle campaign hasn't started. I should have asked whether the new LULC file being used in the current testing of compset(s)/resolution(s) for v2 watercycle and it appears that the new LULC file hasn't been created. I just wanted to check if you or someone else is working with @golaz and the watercycle team to make sure the new LULC file will be tested by the watercycle campaign.

aldivi:
I am willing to work with folks to incorporate the updated land data into workflows. It sounds like @golaz would be a good person to contact. One challenge is that there will be PFT jumps at the starting dates (1850, 2015) until new spinups and historical runs are completed. But the surface files are the same format as before, and work.
I wanted to make sure that the LUT code and results were legitimized before asking people to start working with the new files. The baseline 1850 state and baseline historical transient land surface files are available, and two future scenarios. I can create more future scenarios over time.

aldivi:
I also tried to change the FATES hash as @rgknox suggested, before I pushed the minor change above, but it doesn't appear to have worked.

aldivi:
Here is the list of files and their locations on NERSC/Compy, separated by LUT inputs vs outputs vs land surface files
The files are in the same directories under .../inputdata for both machines.
I have converted all the CMIP6 scenarios to LUH1 format and added these to the LUT input directory

LUT inputs

The following LUT input files are in .../inputdata/lnd/clm2/rawdata/LUT_input_files_current
LUH2_HIST_LUH1_format.nc
LUH2_HIST_LUH1_format_harvest.nc
LUH2_HIST_lo_LUH1_format.nc	    
LUH2_HIST_lo_LUH1_format_harvest.nc 
LUH2_HIST_hi_LUH1_format.nc
LUH2_HIST_hi_LUH1_format_harvest.nc 
LUH2_SSP1_RCP19_LUH1_format.nc
LUH2_SSP1_RCP19_LUH1_format_harvest.nc 
LUH2_SSP1_RCP26_LUH1_format.nc	
LUH2_SSP1_RCP26_LUH1_format_harvest.nc
LUH2_SSP2_RCP45_LUH1_format.nc	   
LUH2_SSP2_RCP45_LUH1_format_harvest.nc  
LUH2_SSP3_RCP70_LUH1_format.nc	
LUH2_SSP3_RCP70_LUH1_format_harvest.nc  
LUH2_SSP4_RCP34_LUH1_format.nc	
LUH2_SSP4_RCP34_LUH1_format_harvest.nc 
LUH2_SSP4_RCP60_LUH1_format.nc
LUH2_SSP4_RCP60_LUH1_format_harvest.nc
LUH2_SSP5_RCP34_LUH1_format.nc
LUH2_SSP5_RCP34_LUH1_format_harvest.nc 
LUH2_SSP5_RCP85_LUH1_format.nc
LUH2_SSP5_RCP85_LUH1_format_harvest.nc
iESM_Ref_CropPast1850_c10142019.nc  
iESM_Ref_CropPast2000_c10142019.nc   
iESM_Ref_CropPast2015_c10142019.nc
surfdata_360x720_mcrop1850_c07082020.nc
surfdata_360x720_mcrop2000_c03062014.nc
surfdata_360x720_mcrop2015_c07082020.nc
mksrf_landuse_template.nc
surfdata_360x720_potveg.nc


LUT outputs

LUT outputs are stored in a single directory for each scenario.
There is one file per year for a given scenario.
There is an additional text file that lists each file needed to generate a given scenario.
Historical baseline:
.../inputdata/lnd/clm2/rawdata/LUT_LUH2_HIST_LUH1f_07082020
SSP2RCP45:
.../inputdata/lnd/clm2/rawdata/LUT_LUH2_SSP2_RCP45_LUH1f_07222020
SSP5RCP85:
.../inputdata/lnd/clm2/rawdata/LUT_LUH2_SSP5_RCP85_LUH1f_07082020

Land surface files

The following land surface files are in .../inputdata/lnd/clm2/surfdata_map
landuse.timeseries_ne30np4_HIST_simyr1850-2015_200728.nc
landuse.timeseries_ne30np4_SSP2_RCP45_simyr2015-2100_200728.nc
landuse.timeseries_ne30np4_SSP5_RCP85_simyr2015-2100_200728.nc
surfdata_ne30np4_simyr1850_c200729.nc
surfdata_ne30np4_simyr2015_c200728.nc

bishtgautam:
@aldivi, Few comments regarding the netcdf files:

- Can we rename `/compyfs/inputdata/lnd/clm2/rawdata/LUT_input_files_current` because today's 'current' will no longer be current in the future? How about adding the `ddmmyy` to the directory name? 
- Few netcdf files in the directory have a suffix `cddmmyy.nc`. How about renaming the files so all of the files have the suffix and updating the code?
- Can we remove `format` from the filenames?
- What is the difference between `*_format_harvest.nc` and `_format.nc`?
- The data in `*_format.nc` does not have any attributes such as `long_name` and `units`.
- Shouldn't `surfdata_*.nc` be added in `lnd/clm2/surfdata_map` instead of `lnd/clm2/rawdata/LUT_input_files_current`?

aldivi:
Note that the LUT and surface files and ELM will be restructured for v3 to incorporate the LUH2 data without conversion to LUH1 format. This update ensures valid surface data in the existing format for use by v2.

Are the files in lnd/clm2/rawdata/LUT_input_files_current/LUT_input_files_current being used to generate files in LUT_LUH2_HIST_LUH1f_07082020, LUT_LUH2_SSP2_RCP45_LUH1f_07222020, and LUT_LUH2_SSP5_RCP85_LUH1f_07082020?
Yes, although the output files (e.g., those in LUT_LUH2_HIST_LUH1f_07082020) have been moved to those locations after being created in my scratch directory. The default output location is in the working land_use_translator code directory in an "outputs" folder.

Can we rename lnd/clm2/rawdata/LUT_input_files_current because today's 'current' will no longer be current in the future? How about adding the ddmmyy to the directory name?
Well, this is named so that current is always current, to make it easier to find the current files, rather than sifting through dated directories. Also, a new directory does not have to be created each time files are added (but new files shouldn't be needed at this point, at least once the scenario files are rebuilt/renamed). Dating all of the scenario files as you suggested should allow this "current" naming scheme to work. So I prefer to keep it as "current," but can change it if you think it is necessary.

Few netcdf files in the directory have a suffix cddmmyy.nc. How about renaming the files so all of the files have the suffix and updating the code?
Can we remove format from the filenames?
I propose to replace "format" with the date, but in the same date format as the other files in this directory (cmmddyyyy). "_harvest.nc" should still be the suffix because these files are paired and with identical names (except for "harvest" at the end) they are listed next to each other in the directory and only the state file is needed as an input argument to the LUT.

What is the difference between *_format_harvest.nc and _format.nc?
The file with "harvest" is the harvest data, while the other file is the land cover state data. This was the previous file naming convention.

The data in *_format.nc does not have any attributes such as long_name and units.
Hmmm. I am using previously made files as templates. It appears that the fillvalue is not correct either in the state file. I should be able to rebuild the state files with the appropriate info. I didn't originally do this because I already had a template that was being used.

Shouldn't surfdata_*.nc be added in lnd/clm2/surfdata_map instead of lnd/clm2/rawdata/LUT_input_files_current?
No, these are files specific to this system, with a special format. So I don't think they should be in the surfdata_map directory. They are generated/extracted from LUT outputs, except for the year 2000 data which represent satellite derived PFTs (this should probably be updated when the new LUT is developed for v3). For iESM, these file are also in a specific directory within the iac input data directory.

I haven't solved the problem of generating the text files automatically. These files contain the full path, which is different on each machine. And the location of the LUT output files is also user-dependent. They could be in a user's scratch directory, or in the machine's inputdata directory. Maybe the files were created in one directory, and then moved to another, which is how I created the current ones. Any automatically generated file would likely have to be modified. I guess it wouldn't hurt to have a template other than the ones that already exist for other scenarios, but those already have mostly correct paths if the LUT output files are put into the rawdata directory for use by others. Any suggestions?

bishtgautam:
@aldivi, FYI. I updated your comment to include `>`

bishtgautam:
@aldivi, You answered all my questions, but I have a followup on the two questions. 

- [x] 1. Thanks for the explanation.
- [ ] 2. If `LUT_input_files_current` will always be current, then how about renaming it to be `LUT_input_files`?
- [x] 3. Your proposed rename of files sounds good.
- [x] 4. Okay, let's keep `format` in the filenames.
- [x] 5. Thanks for the explanation.
- [x] 6. Adding the attributes in the files sounds good to me.
- [x] 7. Okay, let's not move the surface datasets.
- [ ] 8. To automatically generate those text files, the [namelist_defaults_clm4_5_tools.xml#L323]( https://github.com/E3SM-Project/E3SM/blob/master/components/clm/bld/namelist_files/namelist_defaults_clm4_5_tools.xml#L323) needs to be updated.

aldivi:
2. Sure, I will rename LUT_input_files_current to LUT_input_files. Since there is an older LUT_input_files on NERSC, I will rename that one LUT_input_files_08012019, as that is the date of the most recent file in that directory.

4. I will replace "format" with the date as per #3, so it doesn't have to be in the name.

8. Yeah, namelist_defaults_clm4_5_tools.xml also needs to be modified with respect to the the final location of the input files for mksurfdata. Another complication. At least the paths in this file are not machine dependent. This file should be updated along with the text file so that a default filename can be put into the mksurfdata namelist for the single-year surface file (otherwise creating the namelist fails). I modified this default namelist file manually also, as the final location of these files should also be known. But this entry is more easily updated in the namelist if it is incorrect. I don't explicitly mention the default namelist file in the instructions, but I should (I just note the importance of making sure that the mksrf_fvegtype namelist option is correct). The text filename with all of the LUT output files is also in the mksurfdata namelist and determines the files to open for the transient PFTs/harvest, so this text file has to be correct.

aldivi:
I have renamed and rebuilt the LUT scenario input files with updated attributes as described above, and put them on NERSC and Compy. The baseline historical and ssp5 rcp85 have older dates because they were already made, and are still the same data. The ssp2 rcp45 file has the newer date compared to the existing land surface dataset because I had to update the urban data, which are not used at all by this processing, but are present.

I do not have permissions to change the LUT_input_files directory on NERSC, so for now the files are still in LUT_input_files_current on both machines.

I have updated the instructions page on confluence also.

bishtgautam (approved):

bishtgautam (changes requested):
The FATES submodule still needs to be fixed.

bishtgautam:
In your branch, FATES is at `bbc171011b043ad1daebbba2866c61134d6f13e9` instead of `30072be99edffa73a27465682b5d20fdbd09da1e`

aldivi:
That is strange, because when I look online at https://github.com/aldivi/E3SM/tree/aldivi/elm/update-lut/components/clm/src/external_models
it shows  30072be99edffa73a27465682b5d20fdbd09da1e

bishtgautam:
@aldivi You are correct, but I'm not sure why github is showing changes to FATES under https://github.com/E3SM-Project/E3SM/pull/3771/files

rgknox:
That set of changes it is pointing to is odd. (I wonder if this is just a github issue and not a git issue)

@aldivi , did you remember to add and commit the corrected submodule pointer from FATES?

> cd external_models/fates
> git checkout  30072be99edffa73a27465682b5d20fdbd09da1e
> cd ..
> git add fates
> git commit -m "Setting fates submodule pointer to 30072be"

bishtgautam:
@rljacob This development branch includes two merges of the master and I believe those two merges can be avoided. Should this PR be replaced by another one that does not include the merges of the master?

rljacob:
I'd prefer that but its up to you.

aldivi:
I didn't cd up one directory, and I used commit -a, rather than an explicit add.
I don't know what I am doing wrong with respect to submodules or their pointers. They don't seem to behave as described.
Can just the non-fates changes be merged? This PR is not for changes to FATES.

aldivi:
actually, i did cd up to the top E3SM git directory before doing the commit

bishtgautam:
@aldivi I cherry-picked your commits and created a new branch `bishtgautam/lnd/aldivi-update-lut`. This new branch does not have any merges of the master and does not include any changes to FATES submodule. Could you please run the LUT code in my branch and make sure the results are the same w.r.t. your branch? Thanks.

aldivi:
Thank you, Gautam, for cleaning up my mess. Sorry about that. I am trying a different system to reduce this submodule problem in the future. I am guessing that the master merges were related to trying to get the submodule pointer reset.

I have tested your branch all the way through land surface file creation on Cori, and compared the code files and the outputs, and it is all good with the minor mods described below.

In your branch there are two minor code mods, and two files to delete, for this to be final. These two files in .../tools/clm4_5/mksurfdata_map/ are old and not useful and can be confusing:
landuse_timeseries_hist_simyr1850-2005.txt
landuse_timeseries_hist_simyr1850-2015_LUH2.txt

The first code mod is in .../mksurfdata_map/src/Makefile.common on line 251. Not sure how this got through, but a second dash is needed before flibs. The correct line is:
251>   LDFLAGS := $(shell \$(LIB_NETCDF)/../../../bin/nf-config --flibs)

The second mod is in .../land_use_translator/updateannuallanduse_v2.c and keeps the output files from having two creation dates. Three new lines are inserted before line 5919, and line 5919 has a minor mod and becomes line 5922. The correct code is:
5919>   numkeep = numkeep - 10;
5920>   char outbase[1000];
5921>   strncpy(outbase, usesfname, numkeep);
5922>   sprintf(out_land_filebase, "LUT_%s", outbase);

I am going to request an account at argonne for blues and anvil so I can access the main data server.

Note that the defaults in namelist_defaults_clm4_5.xml that are used for compsets have not been updated with the three new land surface files, nor have new land surface files been created to replace the current ones, as this will cause inconsistencies and confusion with current runs.
As new spinups are generated with new land surface files at the desired resolutions for v2, these defaults will need to be updated so that the compsets are consistent with each other.

I have also found two issues that are not included here, and so I guess I need to report them as issues in the appropriate places:

I found a bug in the mosart.buildnml file that checks for a restart file by appending the file name twice, causing case_setup to fail.
I found that the scorpio code fails to read or write netcdfs for at least one netcdf library version. Scorpio assumes that the netcdf read and write functions accept NULL pointers, which isn't necessarily the case. I had to allocate and assign values and deallocate pointers to get a different ELM test to run on lobata. I don't yet know what happened to my job on Cori so I don't know if it is an issue there.

bishtgautam:
@aldivi Thanks for testing out my branch.

>In your branch there are two minor code mods, and two files to delete, for this to be final.

Could you make those changes and commit them to my branch? Once the changes are committed, I will close this PR and open a new one.

> I have also found two issues 

Feel free to open new issues under https://github.com/E3SM-Project/E3SM/issues.

aldivi:
I pushed the mods stated above to your branch.
Thanks.

bishtgautam:
Replaced by #3822

PR #3773, opened 2020-08-16
Make one of homme targets noio

oksanaguba:
To keep no-io-builds functionality in tests.
This also fixes no-io build in the current master.

[bfb]


rljacob:
@oksanaguba please merge this.

oksanaguba:
in next

oksanaguba:
This did not resolve all no-io build issues, 'native output' targets would fail, but we don't need native executables for no-io runs anyway.

PR #3660, opened 2020-06-21
Reinstate clubbv2

wlin7:
This is  to reinstate clubbv2 by reverting the previous revert of clubbv2 to clubbv1 (PR #3518).
The PR also adds clubb_c_k10h as a namelist variable to control c_K10h value at runtime.

[non-BFB]

PR #3774, opened 2020-08-17
Update CIME to bring in support for MALI gis1to10km configuration

jonbob:
This PR brings in a new CIME submodule with two small changes that are necessary prior to an E3SM PR to bring in the MALI gis1to10km configuration. The two changes are:
* add mpas.gis1to10km as option for GLC_GRID; and
* increase i90_MXDEP from 4 to 5 in mpeu.

Neither change should have any impact on current configurations. See [ESMCI PR #3660](https://github.com/ESMCI/cime/pull/3660).

Update the CIME hash from

9239542 (HEAD, origin/master, origin/HEAD, master) Merge pull request 3644 from ESMCI/darincomeau/reorder_drof_options

to the current master

896bdf5 (HEAD, origin/master, origin/HEAD, master) Merge pull request #3660 from ESMCI/jonbob/add-gis1to10km

Other changes
- cprnc: Fix integer single-variable analysi
- Improve check_input_data.py to use os.path.isfile() rather than os.path
-  Nag build fix
- fix py3 issue
- WIP Query config components improvements

No impact on E3SM
-  changes needed to have cdeps 3d stream input working
-  port to hafs
- remove hard-wired nuopc mapping files except for required custom ones
- remove FoX from buildlib and allow CDEPS to handle

[BFB]


jonbob:
In MPAS we have a separate branch for staging changes to go into E3SM, and nothing can get merged there without a PR into E3SM. That avoids this issue of bringing in a bunch of unexpected changes. And then we have periodic framework updates that merge changes from shared parts of MPAS. I'm not sure that would work for CIME, but I don't know that it's reasonable for anyone who needs a simple CIME change to be responsible for a bunch of updates they don't have any knowledge of.

jonbob:
@rljacob - if you want to use the pattern you listed, it may be that you or @jgfouca need to do all the CIME merges. Most of us could probably get something like that from the git history, but may not be able to do much other than list it. And it's still seems like a dangerous approach, because if there's ever a problem from a CIME commit that brings in several different updates, it's painful to go through and figure out which one is responsible. That's exactly why we have the workflow I described for MPAS...

jonbob:
@rljacob - "git log --first-parent" shows nine or ten commits between our previous CIME hash and the head of master?

jgfouca:
I think we should continue to observe how E3SM developers like @jonbob are needing to modify ESMCI/CIME and try to come up with solutions that allow them to work within the E3SM repo instead if at all possible.

jonbob:
Thanks @jgfouca. I'm not sure what the answer is. The file we're most likely to change is
```
src/drivers/mct/cime_config/config_component.xml
```
which has become more-or-less E3SM-specific anyway. But in this case, I also had to modify a file in mpeu that I wouldn't normally expect to touch. So what does make sense if CIME is having this much integration? Should it be that a developer would request a CIME update to master in E3SM just prior to a specific CIME PR, so that what the developer brings in is limited to their own changes? Or that we just accept that anything on CIME master has been validated? Or is it even feasible to have an e3sm/develop branch of CIME, which is what we do for MPAS, that just gets our specific updates and then periodically CIME master gets merged onto it? Or does that just get us back into the world of Jim having to do those awful merges that sucked up so much of his time?

jonbob:
@rljacob - is this PR description more what you had in mind?

jgfouca:
@jonbob , I think the vision is that E3SM developers should rarely need to make ESMCI/CIME changes for their work and the CIME submodule will normally be updated by a CIME developer, like me.

In the case where you have an E3SM PR that needs a CIME submodule update, I think the best thing is to just ask for a CIME update in the slack channel, making sure to @ me and rob. Once that PR is in, the other PR should be rebased and submitted.

I think we will continue to track the ESMCI/master branch. If something unacceptable has been put in CIME, hopefully it will either get caught in ESMCI PR review or E3SM integration testing of the E3SM PR that updates CIME. 

The awful merges from before are gone forever, thankfully. Now that E3SM's CIME has the same commit history as the main CIME, there's no commit re-writing shenanigans. Even if we did feel that we needed a special E3SM branch in ESMCI/CIME, git is very good at merging commits with common ancestry.

rljacob:
@jonbob instead of the title of the merge commit, I used the first line in the description and separated them in to E3SM and non-E3SM.

jonbob:
Thanks @rljacob and @jgfouca. I guess what we should hope for is that we don't have to make many changes to CIME, especially since you split out so much of what is E3SM-specific and most likely needs to be modified by developers.

jonbob:
merged to next

jonbob:
merged to master

PR #3779, opened 2020-08-19
Integrate SAM++ codebase into E3SM-MMF

mrnorman:
This PR integrates a C++ version of the CRM used for the MMF (SAM++). The SAM++ code lives in physics/crm/samxx and includes standalone tests. Switching between the old and new CRM is done with the "-crm [sam|samxx]" switch, which adds -DMMF_SAM or -DMMF_SAMXX in the CPPDEFS. A new compset `F-MMFXX` has been added for the samxx code. 

The code is only single-precision different than the Fortran code in standalone, which should cover all of the code except for the additions to crm_module.F90 and setperturb.cpp, which uses PRNG and cannot be used in standalone precision tests.

This PR brings in two new submodules: https://github.com/mrnorman/YAKL  and  https://github.com/NVlabs/cub. YAKL and samxx directories are each added as static libraries with CUDA device symbols resolved immediately in CMake and are then linked into `libatm.a`.

We integrated the code directly in `components/cmake/build_model.cmake` with some hooks in `components/cmake/common_setup.cmake` and `components/CMakeLists.txt`. The only reason `components/CMakeLists.txt` needed to be edited to enable CUDA as a CMake language if it's requested. Also, the `gnugpu` compiler was added for Summit. `gcc/81.1` must be used with `cuda/10.1.105` in the Summit modules for `gnugpu`, or some incompatible CUDA stuff will show up at build time.

[BFB]


components/cam/bld/config_files/definition.xml
```diff
@@ -340,6 +340,9 @@ Switch to enable or disable fractional cloudiness in m2005 microphysics in CRM:
 <entry id="MMF_microphysics_scheme" valid_values="sam1mom,m2005" value="sam1mom">
 SPCAM Cloud microphysics type.
 </entry>
+<entry id="crm" valid_values="sam,samxx" value="sam">
+SPCAM model
```
whannah1, line 344: gah! I thought I removed all mentions of SPCAM! Anyway, just want to register my disgust. No need to change this.

components/cam/cime_config/config_component.xml
```diff
@@ -114,21 +114,21 @@
...
       <value compset="_CAM5%MMF2"              >-MMF_microphysics_scheme m2005 </value>
       <value compset="_CAM5%MMF2"              >-chem linoz_mam4_resus_mom_soag -rain_evap_to_coarse_aero </value>
       <value compset="_CAM5%MMF2-ECPP"         >-use_ECPP </value>
       <value compset="_CAM5%MMF[12]-MAML"      >-use_MAML </value>
-      <value compset="_CAM5%MMF[12]"           >-crm_nx 64 -crm_nx_rad 16 -crm_ny 1 -crm_ny_rad 1 </value>
+      <value compset="_CAM5%MMF[12X]"          >-crm_nx 64 -crm_nx_rad 16 -crm_ny 1 -crm_ny_rad 1 </value>
       <value compset="_CAM5%MMF[12]-MAML"      >-crm_nx 64 -crm_nx_rad 64 -crm_ny 1 -crm_ny_rad 1 </value>
-      <value compset="_CAM5%MMF[12]-RCE"       >-aquaplanet -rce </value>
-      <!--  -->
+      <value compset="_CAM5%MMF[12X]-RCE"      >-aquaplanet -rce </value>
```
whannah1, line 131: There's likely a conflict here from the recent PR that added a MMF aqua planet compset. I can fix this after you rebase.

whannah1:
@mrnorman , I'm gonna need you to undo the changes to the ECPP files. We can't use ECPP with samxx right now anyway, so we don't need those files to be in the fortran directory. Also, I'm currently in the middle of a huge refactor of ECPP which will touch almost every line of all those files, so I'd rather not deal with unnecessary conflicts from that. 

Looking at what you did, I think I just need you to undo the move of **crm_ecpp_output_module.F90** from **physics/crm/sam/** to **physics/crm/**. It looks like the other files were just copied into the crm/samxx/fortran/ directory, so those can just be deleted, which means you might have to remove some call blocks and use statements.

mrnorman:
I don't want to use any code in samxx/fortran. That's for standalone tests only. I moved crm_ecpp_output_module.F90 to the crm/ folder because it has to exist in both sam and samxx, so there's no reason to copy it. samxx doesn't use ECPP, true, but crm_physics does, so it still has to exist. Can you help me understand how moving crm_ecpp_output_module.F90 inhibits us from integrating the optics changes?

mrnorman:
Oh, sorry, you're doing an ECPP refactor. I can do a rote copy of ECPP for now, sure, but after your refactor, we need to move it to the crm/ folder to keep from having needlessly duplicated code.

whannah1:
ok, sounds good. Having a temporary copy of crm_ecpp_output_module.F90 in the samxx folder should be fine. 
I guess that means you should also leave those other copies of ecpp files in the samxx/fortran directory.

I'll plan to rebase my ECPP work after samxx gets integrated and deal with deleting the copied files as part of my ECPP branch.

mrnorman:
IIRC, only output_module is needed because the other ECPP files are functions, not data referenced from crm_physics

whannah1:
The ecppvars file is also needed, which I already changed a bunch, and renamed. I didn't think you would need to touch any of this stuff, but if I had known I would have waited a bit.

jgfouca:
@mrnorman , you'll need to make an ESMCI/CIME PR for your CIME change.

cime_config/machines/config_compilers.xml
```diff
@@ -1525,6 +1585,39 @@ flags should be captured within MPAS CMake files.
...
+  <SCXX> g++ </SCXX>
+  <SFC> gfortran </SFC>
+  <PIO_FILESYSTEM_HINTS>gpfs</PIO_FILESYSTEM_HINTS>
+  <NETCDF_C_PATH>$ENV{NETCDF_C_PATH}</NETCDF_C_PATH>
+  <NETCDF_FORTRAN_PATH>$ENV{NETCDF_FORTRAN_PATH}</NETCDF_FORTRAN_PATH>
+  <SUPPORTS_CXX>TRUE</SUPPORTS_CXX>
+  <CUDA_FLAGS>
+    <append> -O3 -arch sm_70 --use_fast_math </append>
+  </CUDA_FLAGS>
+  <USE_CUDA>TRUE</USE_CUDA>
```
jgfouca, line 1618: It would probably be better for USE_CUDA to be set in the parent "gnugpu" block. That way, all child blocks that inherit from it will automatically have it.

cime_config/machines/config_compilers.xml
```diff
@@ -1525,6 +1585,39 @@ flags should be captured within MPAS CMake files.
...
+    <append DEBUG="FALSE"> -O2 </append>
+  </FFLAGS>
+  <CPPDEFS>
+    <append COMP_NAME="gptl"> -DHAVE_SLASHPROC </append>
+  </CPPDEFS>
+  <SLIBS>
+    <append MPILIB="mpi-serial">-L$ENV{NETCDF_C_PATH}/lib -lnetcdf -L$ENV{NETCDF_FORTRAN_PATH}/lib -lnetcdff -L$ENV{ESSL_PATH}/lib64 -lessl -L$ENV{NETLIB_LAPACK_PATH}/lib -llapack</append>
+    <append MPILIB="!mpi-serial"> -L$ENV{PNETCDF_PATH}/lib -lpnetcdf -L$ENV{HDF5_PATH}/lib -lhdf5_hl -lhdf5 -L$ENV{NETCDF_C_PATH}/lib -lnetcdf -L$ENV{NETCDF_FORTRAN_PATH}/lib -lnetcdff -L$ENV{ESSL_PATH}/lib64 -lessl -L$ENV{NETLIB_LAPACK_PATH}/lib -llapack</append>
+  </SLIBS>
+  <MPICC> mpicc </MPICC>
```
jgfouca, line 1605: Child settings do not need to be repeated if they are identical to the parent.

mrnorman:
@jgfouca , I created a PR and assigned you to it just now.

cime_config/machines/config_compilers.xml
mrnorman: I can do that. I imagine a scenario coming soon, though, where machine="frontier" and compiler="gnugpu" will want HIP instead of CUDA.

jgfouca:
@mrnorman , thanks for the CIME PR. I just merged it to CIME master. You'll need to remove any cime submodule changes on this PR branch. Once the CIME submodule has been updated on E3SM master, you can rebase or upstream merge this branch to make it complete.

cime_config/machines/config_compilers.xml
mrnorman: There is no parent entry for Summit. I will push a commit that gets rid of the repeated compilers, but that's only about 5 lines (which doesn't include mpiCC, which is different than the parent gnugpu entry)

cime_config/machines/config_compilers.xml
mrnorman: Actually, can I push back on this because of that reasoning? Otherwise, I'd have to set USE_CUDA to false explicitly for non-CUDA gnugpu entries later.

mrnorman:
@whannah1 @jgfouca , The previous commit breaks the build, where I move `crm_ecpp_output_module.F90` out of `crm/` and into both `crm/sam/` and `crm/samxx/`, but I cannot figure out why the build is breakin. You'll see I dump out the total `${SOURCES}` in the CMake configure phase. It shows both `params_kind.F90` and `crm_ecpp_output_module.F90` in the `${SOURCES}`. However, during compilation, there's an error:

```bash
 896 /autofs/nccs-svm1_home2/imn/E3SM/components/cam/src/physics/crm/samxx/crm_ecpp_output_module.F90:3:7:
 897 
 898     use params_kind, only: crm_rknd
 899        1
 900 Fatal Error: Can't open module file 'params_kind.mod' for reading at (1): No such file or directory
```

This shouldn't be happening because `params_kind.F90` is in `${SOURCES}`, and therefore, this dependency should be accounted for. Any ideas why CMake is failing this dependency?

mrnorman:
Reproduce this with the script below:

<details>
#/bin/bash

CASE=samxx_ne4_gnu_gpu
COMPSET=F-MMFXX
PROJ=cli115
RES=ne4pg2_ne4pg2
PECOUNT=84x1
MACHINE=summit
COMPILER=gnugpu
N_UNITS=5
UNIT=ndays

./create_newcase --compset ${COMPSET}   \
                 --case ${CASE}         \
                 --res ${RES}           \
                 --project ${PROJ}      \
                 --machine ${MACHINE}   \
                 --pecount ${PECOUNT}   \
                 --compiler ${COMPILER} \
                 --handle-preexisting-dirs r

cd $CASE

./xmlchange HIST_OPTION=$UNIT
./xmlchange HIST_N=$N_UNITS
./xmlchange STOP_N=$N_UNITS
./xmlchange REST_OPTION=never
./xmlchange --append -id CAM_CONFIG_OPTS -val " -crm_nx 64 -crm_nx_rad 4 "

cat > user_nl_cam << EOF
nhtfrq    = 0,-1
mfilt     = 1,24
fincl2    = 'PS','PSL','TS','PRECT','TMQ','LHFLX','SHFLX','FSNT','FLNT','FLNS','FSNS','FSNTC','FLNTC','LWCF','SWCF','TGCLDLWP','TGCLDIWP','CLDLOW','CLDMED','CLDHGH','CLDTOT' ,'TAUX','TAUY','UBOT','VBOT','TBOT','QBOT','OMEGA850','OMEGA500','T500','T850','Q850','U200','U850','V200','V850','T','Q','Z3' ,'U','V','OMEGA','QRL','QRS','CLOUD','CLDLIQ','CLDICE'
EOF
</details>

cime_config/machines/config_compilers.xml
jgfouca: Yes, if you don't think all children would want this, then feel free to do keep it as is.

jgfouca:
@mrnorman , is it a `-I` problem?

jgfouca:
I'll see if I can reproduce the problem on summit.

mrnorman:
I'm pretty sure I found the error just now. samxx's CMakeLists.txt was GLOBing for Fortran files, so it was trying to compile that file rather than leaving it to libatm.a. I'm working on a fix now.

mrnorman:
I'll switch back to E3SM master CIME once people have done the needed tests on this branch and are happy with it.

jgfouca:
@mrnorman , it's building for me now.

mrnorman:
I'm done trying to use clang-format. It doesn't work

components/cam/src/physics/crm/crm_physics.F90
```diff
@@ -813,12 +837,80 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
...
+      ! Load latitude, longitude, and unique column ID for all CRMs
+      allocate(longitude0(ncol))
+      allocate(latitude0 (ncol))
+      allocate(gcolp     (ncol))
+      do icrm = 1 , ncol
+        latitude0 (icrm) = get_rlat_p(lchnk,icrm) * 57.296_r8
+        longitude0(icrm) = get_rlon_p(lchnk,icrm) * 57.296_r8
+        gcolp     (icrm) = get_gcol_p(lchnk,icrm)
+      enddo
+      ! Load the MSA parameters
```
whannah1, line 849: I think we need a more descriptive comment here, without the abbreviation. I would suggest something like this:
"set CRM mean state acceleration (MSA) parameters from namelist"

components/cam/src/physics/crm/crm_physics.F90
```diff
@@ -813,12 +837,80 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
...
+               crm_output%qg_mean, crm_output%qr_mean, crm_output%mu_crm, crm_output%md_crm, crm_output%eu_crm, &
+               crm_output%du_crm, crm_output%ed_crm, crm_output%flux_qt, crm_output%flux_u, crm_output%flux_v, &
+               crm_output%fluxsgs_qt, crm_output%tkez, crm_output%tkesgsz, crm_output%tkz, crm_output%flux_qp, &
+               crm_output%precflux, crm_output%qt_trans, crm_output%qp_trans, crm_output%qp_fall, crm_output%qp_evp, &
+               crm_output%qp_src, crm_output%qt_ls, crm_output%t_ls, crm_output%jt_crm, crm_output%mx_crm, crm_output%cltot, &
+               crm_output%clhgh, crm_output%clmed, crm_output%cllow, crm_output%sltend, crm_output%qltend, crm_output%qcltend, &
+               crm_output%qiltend, crm_output%tk, crm_output%tkh, crm_output%qcl, crm_output%qci, crm_output%qpl, crm_output%qpi, &
+               crm_output%z0m, crm_output%taux, crm_output%tauy, crm_output%precc, crm_output%precl, crm_output%precsc, &
+               crm_output%precsl, crm_output%prec_crm, latitude0, longitude0, gcolp, igstep, &
+               use_crm_accel, crm_accel_factor, crm_accel_uv)
```
whannah1, line 904: Should we just make these interfaces identical now instead of dealing with it later?
I'll have to change the interface when I address the aerosol optics stuff, so I can take care of it then.

mrnorman: Changing it now will mess with @xyuan 's upcoming PR. I'd rather wait until after his PR to do this.

components/cam/src/physics/crm/samxx/boundaries.cpp
```diff
@@ -0,0 +1,5 @@
+#include "boundaries.h"
+
+void boundaries(int flag) {
+  periodic(flag);
+}
```
whannah1, line 5: Why not just put the contents of "periodic" right here in boundaries?

components/cam/src/physics/crm/crm_physics.F90
mrnorman: Done

components/cam/src/physics/crm/crm_physics.F90
whannah1: ok, well the changes I need to make will probably mess with @xyuan 's PR either way then.

components/cam/src/physics/crm/samxx/boundaries.cpp
mrnorman: Similarity with the original Fortran code organization.

xyuan:
You could go first, I will change the OpenMP to meet these changes. No worries for me

components/cam/src/physics/crm/crm_physics.F90
mrnorman: He said he's OK with us changing this. I can make them the same. I'll work on it today.

mrnorman: Actually, @whannah1 , this is going to get complicated. `crm_output_module.F90` has ifdefs for M2005, ESMT, and MODAL_AERO. Those were pretty ugly in the interface before. They also aren't currently passed to samxx because they aren't used. This might get pretty messy.

whannah1: ok, then let's just leave it and figure it out later. My hope is that we can implement P3 in a way that doesn't require any ifdef nonsense, but maybe after that's implemented we can just delete the fortran SAM code and solve this problem.

components/cam/src/physics/crm/samxx/test/build/download_data.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+wget https://code.ornl.gov/imn/data/raw/master/crmdata_nx8_ny8_nz28_nxrad2_nyrad2.nc
+wget https://code.ornl.gov/imn/data/raw/master/crmdata_nx32_ny1_nz28_nxrad2_nyrad1.nc
```
whannah1, line 4: What's the point of this file?

mrnorman: To grab data needed for standalone tests of samxx

components/cam/src/physics/crm/samxx/pressure.cpp
```diff
@@ -0,0 +1,294 @@
+
```
whannah1, line 1: Maybe we should add a comment here explaining the 2 FFT options?
Unless it's already explained somewhere else?

mrnorman: I can add a comment for it. Once we've fully exercised the new FFTs, I can take this out. I had this in there to allow me to get double precision similarity with the standalone fortran code during development.

components/cam/src/physics/crm/samxx/vars.cpp
```diff
@@ -0,0 +1,1759 @@
...
+  dogzip2D =false                       ;
+  nstatmom =1000                        ;
+  nstatmomstart =99999999               ;
+  nstatmomend =99999999                 ;
+  savemomsep =false                     ;
+  savemombin =false                     ;
+  nmovie =1000                          ;
+  nmoviestart =99999999                 ;
+  nmovieend =99999999                   ;
+  isInitialized_scamiopdata = false     ;
```
whannah1, line 429: A lot of these variables aren't used here. They were just carried over from the standalone SAM and are menat to control SAM's output and restart files. So we could delete some or all. We'd just need to be careful to check that they aren't used, but I'm 90% sure they are just set and no used.

components/cam/src/physics/crm/samxx/samxx_const.h
```diff
@@ -0,0 +1,224 @@
...
+real constexpr crm_dx     = CRM_DX;
+real constexpr crm_dy     = crm_dx;
+real constexpr crm_dt     = CRM_DT;
+
+int  constexpr YES3D = YES3DVAL  ; // Domain dimensionality: 1 - 3D, 0 - 2D
+int  constexpr nx_gl = crm_nx    ; // Number of grid points in X
+int  constexpr ny_gl = crm_ny    ; // Number of grid points in Y
+int  constexpr nz_gl = crm_nz    ; // Number of pressure (scalar) levels
+int  constexpr nsubdomains_x  = 1; // No of subdomains in x
+int  constexpr nsubdomains_y  = 1; // No of subdomains in y
```
whannah1, line 32: The nsubdomain variables is another thing we could remove since we're not going to use this for parallelize the CRM.

whannah1:
The recent force push is from rebasing on master and fixing the samxx code to output the new variable needed for simplifying the aerosol optics. 
I heard some talk on slack about renaming the compsets so they all use "ELM" instead of "CLM", so if that get's merged before this it'll create another conflict, but we're in the clear if we get this merged soon.

whannah1:
Merged to next.

whannah1:
re-merged to next

whannah1:
merged to master

components/cam/src/physics/crm/crm_input_module.F90
```diff
@@ -1,5 +1,5 @@
 module crm_input_module
-   use params, only: crm_rknd
+   use params_kind, only: crm_rknd
 #ifdef MODAL_AERO
    use modal_aero_data, only: ntot_amode
 #endif
```
amametjanov, line 5: File moved from `.../crm/sam/...` to `.../crm/...`. This also needs a path update to `crm_input_module.F90` in `Depends.pgigpu.cmake`.

Same for the other four: i.e.
```
diff --git a/cime_config/machines/Depends.pgigpu.cmake b/cime_config/machines/Depends.pgigpu.cmake
index 6496cb012e..f84c231247 100644
--- a/cime_config/machines/Depends.pgigpu.cmake
+++ b/cime_config/machines/Depends.pgigpu.cmake
@@ -41,11 +41,11 @@ set(FILES_NEED_OPENACC_FLAGS
   cam/src/physics/crm/sam/coriolis.F90
   cam/src/physics/crm/sam/crmtracers.F90
   cam/src/physics/crm/sam/crm_ecpp_output_module.F90
-  cam/src/physics/crm/sam/crm_input_module.F90
+  cam/src/physics/crm/crm_input_module.F90
   cam/src/physics/crm/sam/crm_surface.F90
-  cam/src/physics/crm/sam/crm_output_module.F90
-  cam/src/physics/crm/sam/crm_rad_module.F90
-  cam/src/physics/crm/sam/crm_state_module.F90
+  cam/src/physics/crm/crm_output_module.F90
+  cam/src/physics/crm/crm_rad_module.F90
+  cam/src/physics/crm/crm_state_module.F90
   cam/src/physics/crm/sam/damping.F90
   cam/src/physics/crm/sam/grid.F90
   cam/src/physics/crm/sam/diagnose.F90
@@ -75,7 +75,7 @@ set(FILES_NEED_OPENACC_FLAGS
   cam/src/physics/crm/sam/uvw.F90
   cam/src/physics/crm/sam/vars.F90
   cam/src/physics/crm/sam/zero.F90
-  cam/src/physics/crm/sam/openacc_utils.F90
+  cam/src/physics/crm/openacc_utils.F90
   cam/src/physics/crm/sam/sat.F90 )

 # add accelerator/gpu flags for MPAS files
```

components/cam/bld/configure
```diff
@@ -277,6 +277,7 @@ OPTIONS
      -MMF_microphysics_scheme <string>   CRM microphysics package name [sam1mom | m2005 ].
      -use_ECPP          use CRM clouds for vertical transport, aqueous chemistry and wet removable of aerosols 
      -crm_adv           CRM advection scheme [MPDATA | UM5]
+     -crm <model>       CRM model [sam | samxx]
```
amametjanov, line 280: Can you take a look at `components/cam/cime_config/testdefs/testmods_dirs/cam/crm_gpu/shell_commands`:
```
./xmlchange CAM_CONFIG_OPTS="-phys cam5 -use_MMF  -rad rrtmg -nlev 72 -microphys mg2 -crm_nz 58 
-crm_adv MPDATA -crm_dt 5 -crm_nx 64 -crm_ny 1 -crm_dx 1000 -crm_nx_rad 1 -crm_ny_rad 1 
-bc_dep_to_snow_updates -MMF_microphysics_scheme sam1mom -chem none -cppdefs ' -DSP_DIR_NS 
-DSP_MCICA_RAD ' "

./xmlchange ATM_NCPL=288
```

With the new `-crm $sam_model` addition, does that just need a `-crm sam` addition or something more?
Current configure error in `SMS_P36x1_Ld1.ne4pg2_ne4pg2.FC5AV1C-H01A.blues_pgigpu.cam-crm_gpu` is
```
Command: '/gpfs/fs1/home/e3smtest/jenkins/workspace/ACME_blues_next/E3SM/components/cam/bld/configure 
-s -ccsm_seq -ice none -ocn docn -caseroot /lcrc/group/acme/e3smtest/acme_scratch/blues/J/SMS_P36x1_Ld1.ne4pg2_ne4pg2.FC5AV1C-H01A.blues_pgigpu.cam-crm_gpu.JNextGpu20200913_235859 
-comp_intf mct  -spmd -spmd -smp -nosmp -dyn se -dyn_target preqx -res ne4np4.pg2   -phys cam5 
-use_MMF  -rad rrtmg -nlev 72 -microphys mg2 -crm_nz 58 -crm_adv MPDATA -crm_dt 5 -crm_nx 64 -crm_ny 1 
-crm_dx 1000 -crm_nx_rad 1 -crm_ny_rad 1 -bc_dep_to_snow_updates -MMF_microphysics_scheme sam1mom 
-chem none -cppdefs ' -DSP_DIR_NS -DSP_MCICA_RAD '  ' 
failed with error 'ERROR:  is not a valid value for parameter crm: valid values are sam,samxx' from dir 
'/lcrc/group/acme/e3smtest/acme_scratch/blues/J/SMS_P36x1_Ld1.ne4pg2_ne4pg2.FC5AV1C-H01A.blues_pgigpu.cam-crm_gpu.JNextGpu20200913_235859/Buildconf/camconf'
```

amametjanov (commented):
A couple of questions re: `e3sm_gpu + pgigpu` runs . Can you take a look?

components/cam/bld/configure
whannah1: @amametjanov that's an odd configuration you have there. You're using the hi-res compset with a low-res grid. Also, we changed the names of many variables, so like "SP_DIR_NS" doesn't exist anymore. I'm pretty sure the error is related to trying to manually combine the MMF options with the E3SM compset. Can you just use the "F-MMF1" compset that's currently in E3SM?

components/cam/src/physics/crm/crm_input_module.F90
whannah1: I think we're considering the pgigpu configuration "dead" at this point because it's often slower than the CPU due to the pool allocator issue. Especially since SAM++ is working so well, I doubt we'll ever use the openACC code again, but we should wait a bit to make that decision, or allow @xyuan to remove it as part of his work with openmp.

PR #3780, opened 2020-08-19
Reorder Blues config

amametjanov:
Reorder Blues config

[BFB]


rljacob:
How does this fix the problem?  If you don't have a  <NODENAME_REGEX> entry in blues, then create_newcase with no --mach argument should not pick up the machine as blues.

PR #3781, opened 2020-08-19
Integrate with Travis CI and the E3SM singularity container

lukaszlacinski:
The PR integrates the E3SM repository with Travis CI. The .travis.yml file describes a Travis CI deployment environment and how to clone the E3SM repository with all submodules. It also tells Travis CI, that after the environment is set up, to run
1) .travis/setup.sh script to install Go and Singularity,
2) .travis/run.sh script to create, set up, and build A_WCYCL1850.ne4_oQU240 case
Step 2) takes from 5 to 15 minutes on the Travis CI platform. Such a wide range of execution time can be caused by different total compute load and memory consumption on a machine where a VM for the deployment environment is launched.

[BFB]

### DIFF
--- a/.travis/run.sh
+++ b/.travis/run.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+
+function command_yellow
+{
+    echo -e "\e[33m\$ $1\e[0m"
+    eval $1
+    return $?
+}
+
+function command_yellow_rc
+{
+    command_yellow "$1"
+    rc=$?
+    if [ $rc -ne 0 ]; then
+        exit $rc
+    fi
+}
+
+command_yellow_rc "mkdir -p $HOME/projects/e3sm/cesm-inputdata"
+command_yellow_rc 'cd cime/scripts'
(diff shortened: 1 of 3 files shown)


.travis/run.sh
```diff
@@ -0,0 +1,31 @@
...
+        exit $rc
+    fi
+}
+
+command_yellow_rc "mkdir -p $HOME/projects/e3sm/cesm-inputdata"
+command_yellow_rc 'cd cime/scripts'
+command_yellow_rc './create_newcase --case master.A_WCYCL1850.ne4_oQU240.baseline --compset A_WCYCL1850 --res ne4_oQU240'
+command_yellow_rc 'cd master.A_WCYCL1850.ne4_oQU240.baseline'
+command_yellow_rc './case.setup'
+command_yellow './case.build'
```
rljacob, line 25: So you are only trying to build and not run?  If the case, I think we can avoid downloading data.

lukaszlacinski: If I am not mistaken any missing input data are downloaded in the "run" step.

rljacob: Oh yeah that's right.  Good.

rljacob:
What is the status of this PR?  It is still "draft".

rljacob (approved):

PR #3784, opened 2020-08-20
Fixes bug in crop model

### LINKED ISSUE #2995: Phosphorus balance error in crop model when nitrif_denitrif=.true.
The crop model exits with a phosphorus balance error when the nitrif_denitrif flag is on. This is not detected in the crop test (i.e., 1x1_smallvilleIA) because the flag for nitrif_denitrif is set to false, but in other compsets with crops this flag is set to true (and it should be set to true).

bbye:
Updates allocation mod to avoid negative sminp_to_ppool

Fixes #2995

[non-BFB] (crop model only)


bbye:
I ran the developer checks and compared baseline. All the tests pass, including the crop test. The change might be non-BFB for some crop simulations, but the current tests do not indicate that. The reason is the current crop test doesn't have nitrification-denitrification on.

PR #3790, opened 2020-08-21
refactor the crm code to create a local copy type data for OpenMP 4.5 implementation

xyuan:
since OpenMP 4.5 doesn't support fortran types, this PR refactors the CRM code to create a local copy of types data in crm_data_module, it keeps the same CRM/E3SM interface.

BFB


xyuan:
close this PR and redo the OpenMP with latest master

PR #3792, opened 2020-08-24
add rrtmgp_enable_limiter_warnings namelist flag

whannah1:
This adds a flag to enable limiter warnings in RRTMGP. The flag is true by default, maintaining the previous behavior, but is automatically set to false for aqua planet cases, because these cases often produce very low temperatures aloft that can lead to very large log files full of unhelpful warnings. 

[BFB]

### DIFF
--- a/components/cam/bld/namelist_files/namelist_definition.xml
+++ b/components/cam/bld/namelist_files/namelist_definition.xml
@@ -3760,7 +3760,7 @@ File path for RRTMGP longwave gaseous absorption coefficients file.
 File path for RRTMGP shortwave gaseous absorption coefficients file.
 </entry>
 
-<entry id="rrtmgp_enable_limiter_warnings" type="logical"
+<entry id="rrtmgp_enable_temperature_warnings" type="logical"
        category="radiation" group="radiation_nl" valid_values="">
 Enables low temperature warnings in RRTMGP.
 Disabling this is useful to avoid filling up logs for aqua planet cases.
--- a/components/cam/bld/namelist_files/namelist_definition.xml
+++ b/components/cam/bld/namelist_files/namelist_definition.xml
@@ -3760,6 +3760,13 @@ File path for RRTMGP longwave gaseous absorption coefficients file.
 File path for RRTMGP shortwave gaseous absorption coefficients file.
 </entry>
 
+<entry id="rrtmgp_enable_limiter_warnings" type="logical"
+       category="radiation" group="radiation_nl" valid_values="">
+Enables low temperature warnings in RRTMGP.
+Disabling this is useful to avoid filling up logs for aqua planet cases.
+Default: TRUE
+</entry>
+
 <!-- Rayleigh Friction Parameterization -->
 
 <entry id="rayk0" type="integer" category="rayleigh_friction"
--- a/components/cam/src/physics/crm/rrtmgp/radiation.F90
+++ b/components/cam/src/physics/crm/rrtmgp/radiation.F90
@@ -114,8 +114,8 @@ module radiation
    ! The RRTMGP warnings are printed when the state variables need to be limted,
    ! such as when the temperature drops too low. This is not normally an issue,
    ! but in aquaplanet and RCE configurations these situations occur much more 
-   ! frequently, so this flag was added to be able to disable the warning messages.
-   logical :: rrtmgp_enable_limiter_warnings = .true.
+   ! frequently, so this flag was added to be able to disable those messages.
+   logical :: rrtmgp_enable_temperature_warnings = .true.
 
    ! Model data that is not controlled by namelist fields specifically follows
    ! below.
@@ -223,7 +223,7 @@ contains
                               use_rad_dt_cosz, spectralflux,   &
                               do_aerosol_rad,                  &
                               fixed_total_solar_irradiance,    &
-                              rrtmgp_enable_limiter_warnings
+                              rrtmgp_enable_temperature_warnings
 
       ! Read the namelist, only if called from master process
       ! TODO: better documentation and cleaner logic here?
@@ -252,7 +252,7 @@ contains
       call mpibcast(spectralflux, 1, mpi_logical, mstrid, mpicom, ierr)
       call mpibcast(do_aerosol_rad, 1, mpi_logical, mstrid, mpicom, ierr)
       call mpibcast(fixed_total_solar_irradiance, 1, mpi_real8, mstrid, mpicom, ierr)
-      call mpibcast(rrtmgp_enable_limiter_warnings, 1, mpi_logical, mstrid, mpicom, ierr)
+      call mpibcast(rrtmgp_enable_temperature_warnings, 1, mpi_logical, mstrid, mpicom, ierr)
 #endif
 
       ! Set module data
@@ -276,7 +276,7 @@ contains
                          iradsw, iradlw, irad_always, &
                          use_rad_dt_cosz, spectralflux, &
                          do_aerosol_rad, fixed_total_solar_irradiance, &
-                         rrtmgp_enable_limiter_warnings
+                         rrtmgp_enable_temperature_warnings
       end if
    10 format('  LW coefficents file: ',                                a/, &
              '  SW coefficents file: ',                                a/, &
@@ -1530,11 +1530,11 @@ contains
                      call handle_error(clip_values( &
                         tmid_col(1:ncol,1:nlev_rad), k_dist_lw%get_temp_min(), k_dist_lw%get_temp_max(), &
                         trim(subname) // ' tmid' &
-                     ), fatal=.false.)
+                     ), fatal=.false., warn=rrtmgp_enable_temperature_warnings)
                      call handle_error(clip_values( &
                         tint_col(1:ncol,1:nlev_rad+1), k_dist_lw%get_temp_min(), k_dist_lw%get_temp_max(), &
                         trim(subname) // ' tint' &
-                     ), fatal=.false.)
+                     ), fatal=.false., warn=rrtmgp_enable_temperature_warnings)
                      call t_stopf('rad_check_temperatures')
 
                      ! Set gas concentrations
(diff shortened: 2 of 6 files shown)


whannah1:
All tests in the e3sm_atm_developer suite pass.

brhillman:
Did you do a baseline comparison? I assume we can add the BFB flag?

whannah1:
Yea, I did baseline comparison and meant to add the BFB flag, but I spaced it.

components/cam/src/physics/rrtmgp/radiation_utils.F90
```diff
@@ -333,7 +334,9 @@ subroutine handle_error(error_message, fatal)
          if (fatal_local) then
             call endrun(trim(error_message))
          else
-            print *, trim(error_message)
+            if (present(warn)) then
```
brhillman, line 337: Did you test this? The way you have this setup, it appears as though you will always throw a warning message so long as ANY value of warn is passed to `handle_error`. I think you should create a `warn_local` variable, set it to `.true.`, and then if `warn` is present then set `warn_local = warn` and check the logical value of `warn_local`. The way it is now you never check the value of `warn`, only if it's present. I.e., you want something like the following:
```
warn_local = .true.
if (present(warn)) warn_local = warn
```
and then within the error checking block:
```
if (fatal_local) then
   call endrun(trim(error_message))
else if (warn_local) then
   print *, trim(error_message)
end if

File: components/cam/src/physics/rrtmgp/radiation.F90 [hunk 2]
```diff
@@ -1219,11 +1230,11 @@ subroutine radiation_tend(state,   ptend,    pbuf,          cam_out, cam_in,  &
          call handle_error(clip_values( &
             tmid(1:ncol,1:nlev_rad), k_dist_lw%get_temp_min(), k_dist_lw%get_temp_max(), &
             trim(subname) // ' tmid' &
-         ), fatal=.false.)
+         ), fatal=.false., warn=rrtmgp_enable_limiter_warnings)
```
brhillman, line 1233: Did you mean to only apply this to temperature limiters? If so, I'd suggest renaming the namelist variable to something like "rrtmgp_enable_temperature_warnings", since we now have limiters for optical properties as well. If you meant to apply it to all limiters, then you should add this clause to the other calls to `clip_values`. I'd prefer the former, and then if we want to disable warnings about optical properties we can control that separately (at a later time maybe).

brhillman, line 1233: It doesn't look like there's a corresponding change for this in the MMF version of this file, so these changes have only been applied to the non-MMF radiation interface. Since you've added the namelist machinery there, please make the corresponding change to the call to `clip_values` in components/cam/src/physics/crm/rrtmgp/radiation.F90 as well.

components/cam/src/physics/rrtmgp/radiation_utils.F90
brhillman, line 337: Have you tested this? The way you have this setup now, it doesn't look like you ever actually test the value of `warn`, only if it's present. Thus, I expect that you will ALWAYS be throwing a warning, whether you set `warn` to `.true.` or `.false.` in the call. Instead, I suggest something like the following:
```
logical :: warn_local
warn_local = .true.
if (present(warn)) warn_local = warn
```
and then in the error checking block:
```
if (fatal_local) then
   call endrun(trim(error_message))
else if (warn_local) then
   print *, trim(error_message)
end if
```

whannah1: shoot, I actually didn't test it. I'll fix this.

components/cam/src/physics/rrtmgp/radiation.F90
whannah1: I was initially going to apply to other limiters, but I realized later that the temperature ones were the only problematic ones. I'll change the name.

components/cam/src/physics/rrtmgp/radiation.F90
```diff
@@ -108,6 +108,12 @@ module radiation
    ! Disabled when value is less than 0.
    real(r8) :: fixed_total_solar_irradiance = -1
 
+   ! The RRTMGP warnings are printed when the state variables need to be limted,
```
brhillman, line 111: In your current implementation, you are only controlling warning messages for *temperature*, but we have limiters for optical properties as well. I'd suggest either renaming the namelist variable to `rrtmgp_enable_temperature_warnings`, or making this flag control the optical property limiters as well. I think I'd prefer the former, and then if we later want to disable warnings for optical properties too we can add a flag to do that.

components/cam/src/physics/rrtmgp/radiation.F90
whannah1: oh, I thought I already did this. Maybe I just forgot to commit it. I'll add it.

components/cam/src/physics/rrtmgp/radiation.F90
whannah1: will do.

components/cam/src/physics/rrtmgp/radiation_utils.F90
whannah1: This appears to be a duplicate comment? I'll resolve for now.

components/cam/bld/namelist_files/namelist_definition.xml
```diff
@@ -3760,6 +3760,13 @@ File path for RRTMGP longwave gaseous absorption coefficients file.
 File path for RRTMGP shortwave gaseous absorption coefficients file.
 </entry>
 
+<entry id="rrtmgp_enable_limiter_warnings" type="logical"
+       category="radiation" group="radiation_nl" valid_values="">
+Enables low temperature warnings in RRTMGP.
```
brhillman, line 3765: See comment below, but right now this only applies to temperature warnings, but has a generic name of "rrtmgp_enable_limiter_warnings". Since we have limiters for other variables as well, I suggest either renaming or applying to the other limiters as well so that the description and name of this namelist variable are consistent and representative of what the code is doing.

components/cam/src/physics/crm/rrtmgp/radiation.F90
```diff
@@ -276,7 +286,8 @@ subroutine radiation_readnl(nlfile, dtime_in)
              '  Use average zenith angle:                           ',l5/, &
              '  Output spectrally resolved fluxes:                  ',l5/, &
              '  Do aerosol radiative calculations:                  ',l5/, &
-             '  Fixed solar consant (disabled with -1):             ',f10.4/ )
+             '  Fixed solar consant (disabled with -1):             ',f10.4/, &
+             '  Enable limiter warnings:                            ',l5/ )
```
brhillman, line 290: This should say "Enable temperature warnings" since you changed the name of the variable.

components/cam/src/physics/rrtmgp/radiation.F90
```diff
@@ -271,7 +281,8 @@ subroutine radiation_readnl(nlfile, dtime_in)
              '  Use average zenith angle:                           ',l5/, &
              '  Output spectrally resolved fluxes:                  ',l5/, &
              '  Do aerosol radiative calculations:                  ',l5/, &
-             '  Fixed solar consant (disabled with -1):             ',f10.4/ )
+             '  Fixed solar consant (disabled with -1):             ',f10.4/, &
+             '  Enable limiter warnings:                            ',l5/ )
```
brhillman, line 285: Likewise, this should read "Enable temperature warnings" for consistency with the renamed namelist variable.

components/cam/src/physics/rrtmgp/radiation_utils.F90
```diff
@@ -312,11 +312,15 @@ function clip_values_3d(x, min_x, max_x, varname, tolerance) result(error_messag
...
    end function clip_values_3d
    !-------------------------------------------------------------------------------
-   subroutine handle_error(error_message, fatal)
+   subroutine handle_error(error_message, fatal, warn)
       use cam_abortutils, only: endrun
       character(len=*), intent(in) :: error_message
       logical, intent(in), optional :: fatal
+      logical, intent(in), optional :: warn
       logical :: fatal_local = .true.
+      logical :: warn_local = .true.
```
brhillman, line 321: I think this could cause problems down the road with threaded cases. Variables that get default values set in declaration implicitly get the "save" attribute in F90, so if you were to run with multiple threads sharing this module then it's possible that the value of warn_local from a different thread gets set to `.false.`, but you assume it has a default value of `.true.`. This is why below in my previous treatment of `fatal_local`, I explicitly set a value regardless of the presence of the `fatal` argument (i.e., whether it's true or false). I probably should have deleted the extra code in the declaration of fatal_local that assign a default value to avoid confusion. Regardless, I think what you have here presents a threading bug vulnerability, so I'd suggest tweaking slightly.

brhillman: Actually, it's probably not just threaded cases. You probably change the default value to `.false.` on your first invoking of this routine with a `.false.` argument. I.e., if you set it to `.false.` in timestep 0 in the call to the temperature clipper, then ALL subsequent calls to `handle_error` probably have `warn_local = .false.`, which is not the desired behavior.

components/cam/src/physics/crm/rrtmgp/radiation.F90
whannah1: Good call. Done.

components/cam/src/physics/rrtmgp/radiation.F90
whannah1: Done.

components/cam/src/physics/rrtmgp/radiation_utils.F90
whannah1: Agreed. Fixed in recent commit.

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #3794, opened 2020-08-25
Minimum droplet number conc. option and cleanup

kaizhangpnl:
Implemented an option to impose a minimum droplet number conc. (in-cloud value) in the simulation. Also cleaned up some unnecessary comments added during v1 development. 

Example: To impose a minimum droplet number concentration of 10 cm-3, set 

micro_mincdnc = 10.D6

[Non-BFB] only when micro_mincdnc (by default -999.) is specified to a
       positive number or CDNC is prescribed.

### DIFF
--- a/components/cam/src/physics/cam/micro_mg2_0.F90
+++ b/components/cam/src/physics/cam/micro_mg2_0.F90
@@ -135,9 +135,7 @@ real(r8), parameter :: minrefl = 1.26e-10_r8    ! minrefl = 10._r8**(mindbz/10._
 ! autoconversion size threshold for cloud ice to snow (m)
 real(r8) :: dcs
 
-!!== KZ_DCS 
 logical :: dcs_tdep
-!!== KZ_DCS 
 
 ! minimum mass of new crystal due to freezing of cloud droplets done
 ! externally (kg)
@@ -175,6 +173,8 @@ logical :: use_hetfrz_classnuc
 
 real(r8) :: ncnst ! constant droplet concentration
 real(r8) :: ninst ! constant ice concentration
+
+real(r8) :: mincdnc     ! if mincdnc > 0, then impose a minimum droplet number conc.  
 
 real(r8) :: rhosu       ! typical 850mn air density
 
@@ -223,11 +223,9 @@ contains
 subroutine micro_mg_init( &
      kind, gravit, rair, rh2o, cpair,    &
      tmelt_in, latvap, latice,           &
-!!== KZ_DCS 
      rhmini_in, micro_mg_dcs, micro_mg_dcs_tdep, &
-!!== KZ_DCS 
      microp_uniform_in, do_cldice_in, use_hetfrz_classnuc_in, &
-     do_nccons_in, do_nicons_in, ncnst_in, ninst_in, &
+     do_nccons_in, do_nicons_in, ncnst_in, ninst_in, mincdnc_in, &
      micro_mg_precip_frac_method_in, micro_mg_berg_eff_factor_in, &
      allow_sed_supersat_in, ice_sed_ai, prc_coef1_in,prc_exp_in,  &
      prc_exp1_in, cld_sed_in, mg_prc_coeff_fix_in, alpha_grad_in, &
@@ -255,9 +253,7 @@ subroutine micro_mg_init( &
   real(r8), intent(in)  :: rhmini_in    ! Minimum rh for ice cloud fraction > 0.
   real(r8), intent(in)  :: micro_mg_dcs
   real(r8), intent(in)  :: ice_sed_ai   !Fall speed parameter for cloud ice
-!!== KZ_DCS 
   logical,  intent(in)  :: micro_mg_dcs_tdep
-!!== KZ_DCS 
 
   logical,  intent(in)  :: microp_uniform_in    ! .true. = configure uniform for sub-columns
                                             ! .false. = use w/o sub-columns (standard)
@@ -281,6 +277,7 @@ subroutine micro_mg_init( &
 
   real(r8), intent(in)  :: ncnst_in
   real(r8), intent(in)  :: ninst_in        
+  real(r8), intent(in)  :: mincdnc_in
 
   character(128), intent(out) :: errstring    ! Output status (non-blank for error return)
 
@@ -288,9 +285,7 @@ subroutine micro_mg_init( &
 
   dcs = micro_mg_dcs
 
-!!== KZ_DCS 
   dcs_tdep = micro_mg_dcs_tdep 
-!!== KZ_DCS 
 
  prc_coef1 = prc_coef1_in
  prc_exp   = prc_exp_in
@@ -332,6 +327,7 @@ subroutine micro_mg_init( &
   nicons = do_nicons_in
   ncnst = ncnst_in
   ninst = ninst_in
+  mincdnc = mincdnc_in
   use_hetfrz_classnuc = use_hetfrz_classnuc_in
 
   ! typical air density at 850 mb
@@ -427,9 +423,7 @@ subroutine micro_mg_tend ( &
   ! Size calculation functions.
   use micro_mg_utils, only: &
        size_dist_param_liq, &
-!!== KZ_DCS 
        size_dist_param_ice, &
-!!== KZ_DCS 
        size_dist_param_basic, &
        avg_diameter
 
@@ -760,9 +754,7 @@ subroutine micro_mg_tend ( &
   ! relative humidity
   real(r8) :: relhum(mgncol,nlev)
 
-!!== KZ_DCS 
   real(r8) :: dcst(mgncol,nlev)        ! t-dependent dcs
-!!== KZ_DCS 
 
   ! parameters for cloud water and cloud ice sedimentation calculations
   real(r8) :: fc(nlev)
@@ -865,9 +857,7 @@ subroutine micro_mg_tend ( &
   qs = qsn
   ns = nsn
 
-!!== KZ_DCS
   call get_dcst(mgncol,nlev,t,dcst)
-!!== KZ_DCS
 
 
   ! cldn: used to set cldm, unused for subcolumns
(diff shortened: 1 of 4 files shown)


components/cam/src/physics/cam/micro_mg2_0.F90
```diff
@@ -1217,6 +1207,11 @@ subroutine micro_mg_tend ( &
            qcic(i,k)=min(qc(i,k)/lcldm(i,k),5.e-3_r8)
            ncic(i,k)=max(nc(i,k)/lcldm(i,k),0._r8)
 
+           ! impose minimum droplet number conc (in-cloud) 
+           if (mincdnc.gt.0.) then
```
xuezhengllnl, line 1211: To be consistent with other lines and the variable type,  how about 
if (mincdnc.gt.0._r8) then

kaizhangpnl: Thanks for the suggestion. It's fixed.

components/cam/src/physics/cam/micro_mg2_0.F90
```diff
@@ -2754,6 +2757,11 @@ subroutine micro_mg_tend ( &
         if (dumc(i,k).ge.qsmall) then
 
 
+           ! impose minimum droplet number conc (in-cloud) 
+           if (mincdnc.gt.0._r8) then
+              nctend(i,k)=( max(nc(i,k)+nctend(i,k)*deltat, mincdnc/rho(i,k)) - nc(i,k))/deltat
```
wlin7, line 2762: This nctend seems to be for whole grid (instead of in-cloud). should the limitation term be multiplied by lcldm?

wlin7:
Hi @kaizhangpnl , the places you imposed a minimum in-cloud conc look good, except one noted above whether the limit should be converted to grid-mean value.

I don't have a full picture when the limit should be imposed. There are several places where conc is updated but not exercising the same constraint. They are listed below. The limiter could be unnecessary or not applicable. Can you please take another look? The line numbers are from the original codes before your changes

- L1655 where unconstrained nc is used.
- L2191 where  nctend is used to compute dumnc
- L2775 where nctend is updated.  In if-block starting L2754, nctend calculated from earlier codes were constrained. Should it be moved after L2775, or should a new constrain be imposed?
- L2845, the reset of nctend would ensure nc in next iteration be 0. Does not matter because updated qc would be < qsmall?
- 

Can I say if we impose minimum on resulting nc (or ncic), we don't really have to also constrain nctend?

wlin7:
The comments above are for file micro_mg2_0.F90. The changes in the other files are all clear.

golaz:
@kaizhangpnl, can you work on finalizing this soon? We are waiting for this to go on master before starting new simulations. Thanks.

golaz:
Thanks @kaizhangpnl. Yes, it would be better to do this in parallel. Commit the code to the branch so that @wlin7 and @xuezhengllnl can look at that while at the same time you run your test simulation.

kaizhangpnl:
@golaz Thanks for your feedback. I will push two commits. The first one is a bug fix (see Wuyin's first comment above) and it has been evaluated in short simulations. The second one includes more changes and it is to address Wuyin's comments on the overall consistency between in-cloud & grid mean cdnc and state & tendency update. For both commits, I've submitted jobs to looked at the impact on aerosol ERF (based on alpha3 configuration provided by Xue). Hopefully I can analyse the results by tomorrow.

kaizhangpnl:
@singhbalwinder It seems that I messed up the git commits in this branch. I wanted to squash three commits into one, but when doing this, github complained: 

`(base) bash-4.2$ git push origin kaizhangpnl/atm/PR-mincdnc 
To https://github.com/E3SM-Project/E3SM.git
 ! [rejected]        kaizhangpnl/atm/PR-mincdnc -> kaizhangpnl/atm/PR-mincdnc (non-fast-forward)
error: failed to push some refs to 'https://github.com/E3SM-Project/E3SM.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Merge the remote changes (e.g. 'git pull')
hint: before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.`

Then I "git pull" and got the request to

`    Merge branch 'kaizhangpnl/atm/PR-mincdnc' of https://github.com/E3SM-Project/E3SM into kaizhangpnl/atm/PR-mincdnc`

I did this twice and in the end there are many commits. Do you have any suggestion on how to resolve this? It would be great to correct this branch rather than create a new one (I guess then I will need to issue another PR?). Thanks!

wlin7:
Hi @kaizhangpnl , no worry about the appearance of more commits than intended. All the intended changes are there. The squash and git pull command caused some back and forth on some of the changes. To make it look better, you can do another git squash, then force pushing it. 
```
git rebase -i 5e84714
git push --force origin kaizhangpnl/atm/PR-mincdnc
```
5e84712 is the first commit of 09/03 shown in the log. In the interactive rebasing, retaining the first 'pick', and replace all the rest of 'pick' with squash. On the next screen, you can put in a new commit message to reflect all the changes, and remove or comment out all the rest.

rebasing caused the commit hash to be re-calculated. This could lead to the resulting HEAD  to appear behind than in the origin or remote branch. We know you are the only person updating this branch, and the squashing is done properly. So it is safe to force push your local to remote.

You can save a copy of micro_mg2_0.F90 before doing git squash, and compare to the one after all this is done, just to make sure.

singhbalwinder:
Thanks Wuyin. Like Wuyin mentioned, after a rebase we have to force push the changes after making sure everything looks okay. Once you force push, it is permanent, so there is no going back.

kaizhangpnl:
@wlin7 and @singhbalwinder: Many thanks for the suggestion. It worked! 

BTW, I've analyzed the new simulations with the alpha3 settings provided by Xue. The impact of imposing an in-cloud minimum CDNC of 10 per cc is similar to what we got before (~0.4W/m2), though alpha3 has a much larger aerosol ERF (-1.90W/m2) compared to alpha22 (-1.59W/m2). The results can be found here: 

https://compy-dtn.pnl.gov/zhan524/E3SM/v2/alpha/erf/nudging_20200821_alpha3_CTRL/
https://compy-dtn.pnl.gov/zhan524/E3SM/v2/alpha/erf/nudging_20200821_alpha3_mincdnc10_v2/ 

We can schedule a meeting to discuss the details about the new implementation.

wlin7 (approved):
All changes look good.

xuezhengllnl (approved):

singhbalwinder:
Since this is BFB, I will start merging it.

singhbalwinder:
Merged to next

rljacob:
These are the tests we think diffed from this PR.
ERP_Ln9.ne4_ne4.F-EAMv1-AQP1.sandiatoss3_intel
ERP_Ln9.ne4_ne4.F-EAMv1-RCEMIP.sandiatoss3_intel

rljacob:
Those cases are aquaplanet while ERS_Ld5.ne4_ne4.FC5AV1C-L.sandiatoss3_intel.cam-rrtmgp passed so it must be aquaplanet config and this PR that is causing the changes.  Not rrtmgp.

whannah1:
Repeating what I said on slack, I think the diff is due to the block of code in micro_mg2_0.F90 that was moved that contains the call to size_dist_param_liq(). Since the nccons flag is true for AQP and RCE compsets this branch changes the order in which the nctend variable is updated. I don't really think this is a bad change though, so it's probably ok to just allow the PR to be non-BFB.

kaizhangpnl:
I just looked at the code and found it should be NBFB for aqua-planet. There is one change in this PR which changes the order of size_dist_param_liq and the nctend update when prescribed CNDC is applied. 

https://github.com/E3SM-Project/E3SM/blob/64dc9f14bc17095dadc896c840d7ee577ff2e012/components/cam/src/physics/cam/micro_mg2_0.F90#L2782-L2804

The change is to resolve the inconsistency pointed by Wuyin. 

So, it is an expected NBFB.

kaizhangpnl:
Just saw @whannah1 's comment. Thanks!

singhbalwinder:
Thank you all! I think we all now agree that it should be tagged as NBFB (thanks to @whannah1 for verifying it). I will wait for @wlin7 response before merging to it master and issuing bless requests for the failing tests.

wlin7:
Great @kaizhangpnl , you and @whannah1 already isolate the issue. This suggests it is safe to go ahead as long as not using Aquaplanet config. But do we want to ensure BFB as well for Aquaplanet?

xuezhengllnl:
> Great @kaizhangpnl , you and @whannah1 already isolate the issue. This suggests it is safe to go ahead as long as not using Aquaplanet config. But do we want to ensure BFB as well for Aquaplanet?

I think the updated position of nctend for the constant CDNC configuration is more reasonable, which can make sure the updated CDNC is still equal to the prescribed ncnst.

xuezhengllnl:
> I just looked at the code and found it should be NBFB for aqua-planet. There is one change in this PR which changes the order of size_dist_param_liq and the nctend update when prescribed CNDC is applied.
> 
> https://github.com/E3SM-Project/E3SM/blob/64dc9f14bc17095dadc896c840d7ee577ff2e012/components/cam/src/physics/cam/micro_mg2_0.F90#L2782-L2804
> 
> The change is to resolve the inconsistency pointed by Wuyin.
> 
> So, I think it is an expected NBFB. @wlin7 @singhbalwinder @xuezhengllnl : Please help double check. Thanks!

@kaizhangpnl I think the code should be BFB for configurations with interactive CNDC mode.

kaizhangpnl:
Yes, it's NBFB only for aqua-planet or any other simulations prescribing CDNC.

wlin7:
> I think the updated position of nctend for the constant CDNC configuration is more reasonable, which can make sure the updated CDNC is still equal to the prescribed ncnst.

But in that  if (ncons) block, it is noted that nctend may be further adjusted if mean droplet size goes out of bound. Do we know the implication of that? We will lose that protection if retaining the current order. Restoring cdnc to prescribed ncnst seems to be always in place via those if-(nccons) blocks early in each of the k-loop.

To maintain BFB for prescribed CDNC as well, we can restore the calculation sequence, and in the follow-up mincdnc constraint, condition it on " .not. nc .and. mincdnc .gt. 0"

xuezhengllnl:
@wlin7 it does make sense. We hate unrealistically giant cloud droplets more than output CDNC somewhat off the prescribed value.

wlin7:
> it does make sense. We hate unrealistically giant cloud droplets more than output CDNC somewhat off the prescribed value.

Thanks @xuezhengllnl . Considering this possibility, we'd better restore the order of calling size_dist_param_liq, and impose a more strict condition for the limiter. Since this only impacts case with prescribed cdnc, it is also ok to merge this PR as is to allow new simulations start right away; and have another PR to restore the extra BFB.

rljacob:
I thought there were going to be some changes to this PR?

singhbalwinder:
We decided to merge it as it is and I will ask for blessing the failing tests. Later, we will issue another PR to fix the BFB issue. @xuezhengllnl needs to run some urgent simulations and these changes are needed for those runs.

Issue #3795, opened 2020-08-26
Sea ice snow melt diagnostic bug


akturner:
When the sea ice completely melts away in the thermodynamics, the final little bit of snow melt is not added to the melts array.

See https://github.com/CICE-Consortium/Icepack/pull/328.

PR #3796, opened 2020-08-26
Update mpas-source: fixes sea ice melt bug

akturner:
This corrects a bug in the snow melt diagnostic part of the column package in MPAS-Seaice. When the sea ice completely melts away in the thermodynamics, the final little bit of snow melt is not added to the melts array.

see MPAS PR https://github.com/MPAS-Dev/MPAS-Model/pull/647

Fixes https://github.com/E3SM-Project/E3SM/issues/3795

[non-BFB] potentially for tests with active seaice


jonbob:
merged to next

jonbob:
Merged to master and DIFFs blessed on compy

PR #3803, opened 2020-08-28
Update CIME submodule again

jgfouca:
Changes:
* Add much better support for allowing users to xmlchange case settings while the model is running
* Add source file feature to env machine settings
* Avoid artificial limit on string lengths in shr_string_listMerge
* topography-based subgrid structure in E3SM
* Sets coupling frequencies for new MPAS-Ocean grid in E3SM.
* Include submodule status in provenance.

Fixes:
* do not try to set var from env_test unless defined in the case

No impact on E3SM:
* new cdeps stream schema and stream definition file
* Cmeps driver env var
* fix logic in cdeps build
* Fix two issues. The FV3 grid problem and by default end the GSWP3 forcing in 2013, as the 2014 data is bad

Details:
Updates cime submodule from bea8a7fc to e0dcc590

[BFB]


rljacob:
Jim, can you including the commit message the old and new hashes of CIME being used?

jgfouca:
Merged to next.

rljacob:
@jgfouca can you update this PR to bring in the latest CIME?  We need some science changes integrated since this started.

jgfouca:
@rljacob , i updated this PR on Friday. Is that recent enough to pick up those changes?

rljacob:
yes it is

rljacob:
@jgfouca when you updated this last Friday, you needed to update the description as well.  The new fields aren't even mentioned.

jgfouca:
@golaz , now that this is in, please try to do some of the concurrent xml changes we talked about https://github.com/ESMCI/cime/issues/3338

PR #3805, opened 2020-09-01
Fixes default compute interval for heat budget

vanroekel:
Currently the mixed layer heat budget is never calculated since the outputStream interval is set to 'none'. This is a simple one line change to enable computation of the mixed layer heat budget once per hour.

[NML]
[BFB]


vanroekel:
@jonbob poking through recent runs I noticed there were no values for the mixed layer heat budget.  Turns out it was off by default.  This PR will turn on the computation once per hour.

jonbob:
merged to next

jonbob:
Re-merged to next and now passes SMS.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.anvil_intel sanity test

vanroekel:
thanks for the fix @jonbob!  Sorry about that formatting error.

jonbob:
Ah, I just feel bad for not seeing that (or testing it)...

jonbob:

PR #3808, opened 2020-09-02
[WIP] Rename land compsets

bishtgautam:
The land compsets are renamed to `ELM` from `CLM45`. The compsets that are
named `CLM40` or `CLM50` are deleted.

[BFB]


rljacob:
Replaced by #3810

PR #3813, opened 2020-09-08
Adding dam construction year to MOSART Water Management scheme

hydrotian:
Add dam construction feature to MOSART-WM to allow dams taking effects after construction. 
This feature is particularly useful for historical transient runs with water management on.
Dam built year was from GRaND database.

[BFB] (namelist will change as new flag is added)

### DIFF
--- a/components/mosart/src/wrm/WRM_type_mod.F90
+++ b/components/mosart/src/wrm/WRM_type_mod.F90
@@ -79,7 +79,7 @@ MODULE WRM_type_mod
      real(r8), pointer :: MaxStorTarget(:)  ! (nd)
 
      integer , pointer :: YEAR(:)           ! (nd) year dam was constructed and operationnal
-     integer , pointer :: active(:)         ! (nd) whether dam is not functional (before construction) or during filling (<10 yrs after built and <80% of capacity) or fully functional (10 yrs after built or >80% of capacity) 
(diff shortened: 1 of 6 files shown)


bishtgautam:
@hydrotian Could you fix the title and update the first comment to indicate if this PR is a BFB or not?

components/mosart/src/wrm/WRM_type_mod.F90
```diff
@@ -78,6 +79,7 @@ MODULE WRM_type_mod
      real(r8), pointer :: MaxStorTarget(:)  ! (nd)
 
      integer , pointer :: YEAR(:)           ! (nd) year dam was constructed and operationnal
+     integer , pointer :: active(:)         ! (nd) whether dam is not functional (before construction) or during filling (<10 yrs after built and <80% of capacity) or fully functional (10 yrs after built or >80% of capacity)
```
bishtgautam, line 82: I would have guessed `active` will be either `0` or `1`, but it could have the following values:
- `0`: Not functional
- `1`: Filling
- `2`: Fully functional

How about renaming `active` to indicate stages (e.g. `activity_stage`)?

hydrotian: Thanks for the suggestion. Sure I can do it.

liho745 (approved):

bishtgautam (approved):

PR #3814, opened 2020-09-08
Updated CLUBBv2 paramters and flags

xuezhengllnl:
Converted several local CLUBB parameters and flags to namelist variables.
 Local variables include altitude_threshold, ipdf_call_placement, and l_vert_avg_closure. 
The corresponding new namelist variables are clubb_vert_avg_closure, 
clubb_ipdf_call_placement, and clubb_altitude_thresh

[NML] New namelist variables are not added to atm_in by default
[BFB] when the revised variables are the default values

### DIFF
--- a/components/cam/src/physics/cam/clubb_intr.F90
+++ b/components/cam/src/physics/cam/clubb_intr.F90
@@ -391,7 +391,11 @@ end subroutine clubb_init_cnst
     use cam_abortutils,  only: endrun
     use stats_variables, only: l_stats, l_output_rad_files
     use mpishorthand
-    use model_flags,     only: l_diffuse_rtm_and_thlm, l_stability_correct_Kh_N2_zm
+    use model_flags,     only: l_diffuse_rtm_and_thlm, l_stability_correct_Kh_N2_zm, &
+                               l_vert_avg_closure, l_trapezoidal_rule_zt, &
+                               l_trapezoidal_rule_zm, l_call_pdf_closure_twice,&
+                               ipdf_call_placement
+
     use parameters_tunable, only: clubb_param_readnl
 #endif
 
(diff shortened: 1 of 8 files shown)


components/cam/src/physics/cam/clubb_intr.F90
```diff
@@ -469,9 +480,12 @@ subroutine clubb_readnl(nlfile)
       call mpibcast(clubb_tk2,                1,   mpir8,   0, mpicom)
       call mpibcast(relvar_fix,               1,   mpilog,  0, mpicom)
       call mpibcast(clubb_use_sgv,            1,   mpilog,   0, mpicom)
+      call mpibcast(clubb_vert_avg_closure,   1,   mpilog,   0, mpicom)
+      call mpibcast(clubb_ipdf_call_placement,   1,   mpilog,   0, mpicom)
```
wlin7, line 484: change mpilog to mpiint in 
`call mpibcast(clubb_ipdf_call_placement,   1,   mpilog,   0, mpicom)`

xuezhengllnl: Thanks for identifying the data type issue. I already changed and committed it to the branch

wlin7 (approved):

wlin7:
The changes look good and are consistent with CLUBB's style of defining and setting the parameters.

xuezhengllnl:
> The changes look good and are consistent with CLUBB's style of defining and setting the parameters.

Thank you, Wuyin, for your carefully review.

components/cam/src/physics/clubb/advance_clubb_core_module.F90
```diff
@@ -105,8 +107,8 @@ module advance_clubb_core_module
                                        ! predictive fields
 
   ! Select the placement of the call to CLUBB's PDF.
-  integer, parameter :: &
-    ipdf_call_placement = ipdf_pre_advance_fields
+  !integer, parameter :: &
```
singhbalwinder, line 110: This commented out code can be completely removed.

components/cam/src/physics/clubb/advance_xm_wpxp_module.F90
```diff
@@ -4327,9 +4330,9 @@ pure function damp_coefficient( coefficient, Cx_Skw_fnc, max_coeff_value, &
       Cx_Skw_fnc            ! Initial skewness function before damping
 
     ! Local variables
-    real( kind = core_rknd ), parameter :: &
+!    real( kind = core_rknd ), parameter :: &
```
singhbalwinder, line 4333: Same here. We can probably remove all commented out lines including comments, since they are not relevant anymore.

singhbalwinder (approved):
Looks great! I have just two very minor comments.

components/cam/src/physics/clubb/advance_xm_wpxp_module.F90
xuezhengllnl: Thank you @singhbalwinder for the suggestion. I removed the unnecessary lines and comments. I will clean up the unnecessary lines in the future commits as well.

components/cam/src/physics/clubb/advance_clubb_core_module.F90
xuezhengllnl: Removed. Thanks!

wlin7:
Thanks @singhbalwinder for also giving a review. Note that the default values set namelist_default_cam.xml for the new variables do not take effect. Haven't figured out why.  The good side of this is that since these new namelist variables are not added to atm_in in default configuration, there would be no NML diff for the tests. The hardwired  values in the code take over as default.

@xuezhengllnl , can you also check in your test run that atm_in does not change when not setting the new namelist variables in user_nl_cam?

wlin7:
Note that to add default settings for the new namelist variables to atm_in, need to add them one by one in cam/bld/build-namelist. Many namelist variables introduced didn't go that path, relying on use_case files instead. We can choose not to add the new variables in this PR to build-namelist either, to make it simple.

wlin7:
Merged to next.

xuezhengllnl:
@wlin7 Thank you so much for figuring this out. The default values might be changed for v2. We can decide if we want to record these variables in atm_in then.

wlin7:
Merged to master.

PR #3817, opened 2020-09-08
Update MPAS-Source: fix Rayleigh damping and line continuation

mark-petersen:
Two bug fixes:
- fix Rayleigh drag logic on implicit vertical mix MPAS-Dev/MPAS-Model/pull/676, only impacts MPAS ocean-only spin-up simulations, not E3SM simulations.
- fixed bad continuation character in ocn init tidal boundary MPAS-Dev/MPAS-Model/pull/675

[BFB]


jonbob (commented):
@mark-petersen - in general I would prefer to see changes to the ocean core separate from changes to compass. Are they related in this PR?

mark-petersen:
Yes, I thought about the same thing. We couldn't test the EC60to30 meshes without the Rayleigh damping fix, and it had to go in first to have a reproducible commit on the MPAS repo to make the grids. Sorry about that. I do agree with the policy to keep these things separate.

mark-petersen:
Ran some tests to be sure. Passes:
```
SMS_D_Ln9.T62_oQU120_ais20.MPAS_LISIO_TEST.cori-haswell_intel
SMS_D_Ln9.T62_oQU120_ais20.MPAS_LISIO_TEST.cori-haswell_gnu
SMS_D_Ln9.T62_oQU120_ais20.MPAS_LISIO_TEST.cori-knl_gnu
```

jonbob:
I'm guessing you just threw in the bug fix? This is still not how we want these things to go...

xylar:
@jonbob, in @mark-petersen's defense, we really did need the bug fix for the spin-up of the EC30to60km meshes we were testing on COMPASS.  It might have been cleaner to include the bug fix first, then the meshes.  But we are really coming down to the wire in terms of testing these meshes in time for v2, so I put a lot of pressure on him to get them in to `ocean/develop` ASAP.

mark-petersen:
I'll change the merge to e3sm/develop to just before the bug fix.

jonbob:
@xylar - I understand that there's a lot that needs to get in, but this happens over and over again. There is a process and we've been bitten before by combining changes that do not need to be mixed together. In the end, I believe those shortcuts cost us more time than they save

xylar:
@mark-petersen, wait, now I'm concerned about `ocean/develop`.  Does this mean that there is not a commit of `ocean/develop` with the Rayleigh damping bug fix and each of the EC30to60km meshes?  Because that would also be a big problem.  How is someone supposed to figure out what commit to got back to to create those meshes and initial conditions if there is no working commit?

@jonbob, I agree on principle.  But if that's how it's going to go, the bug fix should be going in firs, then then meshes.   The other way around and we have the spin-up problem I'm referring to above.

mark-petersen:
I can add just the Fortran code changes in this one and have the next one be COMPASS-only. I'll change the description in a minute.

xylar:
I think I seen now.  Because the bug fix is in the middle of a bunch of COMPASS changes, we have to do COMPASS, bug fix, more COMPASS.

If that's okay for the E3SM timeline, I do think that's the way to go.

jonbob:
@xylar, @mark-petersen - I'm sorry to be hard about this, but it is the process we've all agreed to follow. We had such a bad time historically untangling problems that occurred with PR's that mixed updates, with bisecting commits and more. I understand that the bug fix is fairly trivial and I can look at it and be confident it couldn't do any damage, but it's a fine line...

xylar:
@jonbob, I totally agree.

I also really hope we can agree to just take COMPASS out of MPAS-Model as I have proposed.  That would cut out a lot of pointless traffic like this.

mark-petersen:
OK. This is ready to go. Running a few tests to double check...

jonbob:
@xylar - I agree with you totally about that. When people have asked what would help get PRs through quicker, I bring that up and tell them what good idea you had

mark-petersen:
Yes, I agree too. We will work on the details of separating COMPASS in the coming week.

jonbob (approved):
Looks good by visual inspection -- and I'll run some sanity tests after merging. Thanks for splitting up the PR

jonbob:
test merge passed SMS_D.ne4_oQU240.A_WCYCL1850.anvil_intel sanity test

jonbob:
merged to next

jonbob:
merged to master

PR #3701, opened 2020-07-15
Updating Scorpio to version 1.1.2

jayeshkrishna:
Updating the version of Scorpio to 1.1.2 . This new release
includes,

* Fix for a memory corruption issue with the BOX rearranger
  This fix is required for ne30 pg2 production runs (see #3684),
  ne1024 pg2 runs (see E3SM-Project/scorpio#323) and standalone
  HOMME (see E3SM-Project/scorpio#324)
* Increases the range of I/O decomposition ids allowed. This fix
  is required for some long running simulations
  (see E3SM-Project/scorpio#315)

Also changing the default I/O library in HOMME to Scorpio.

Fixes #3684 

[BFB]

PR #3819, opened 2020-09-09
Script support for new EC30to60E2r2 ocean mesh

jonbob:
Bring in script support for a new ocn/ice EC30to60E2r2. This includes all necessary mapping, runoff, and domain files for the following configurations:
* EC30to60E2r2_r05_ne30pg2 trigrid;
* EC30to60E2r2_T62 CORE forced;
* EC30to60E2r2_TL319 JRA forced.

All mapping/domain/grid files have been placed in the E3SM public inputdata directory and are world-readable.

### DIFF
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -148,6 +148,7 @@
 <config_use_GM ocn_grid="oRRS15to5">.false.</config_use_GM>
 <config_GM_kappa>1800.0</config_GM_kappa>
 <config_GM_kappa ocn_forcing="datm_forced_restoring" ocn_grid="oEC60to30v3wLI">600.0</config_GM_kappa>
+<config_GM_kappa ocn_forcing="datm_forced_restoring" ocn_grid="EC30to60E2r2">600.0</config_GM_kappa>
 <config_GM_closure>'visbeck'</config_GM_closure>
 <config_GM_baroclinic_mode>3.0</config_GM_baroclinic_mode>
 <config_GM_visbeck_alpha>0.13</config_GM_visbeck_alpha>
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -45,6 +45,7 @@
 <config_dt ocn_grid="oRRS15to5">'00:03:45'</config_dt>
 <config_dt ocn_grid="oARRM60to10">'00:10:00'</config_dt>
 <config_dt ocn_grid="oARRM60to6">'00:05:00'</config_dt>
+<config_dt ocn_grid="EC30to60E2r2">'00:30:00'</config_dt>
 <config_time_integrator>'split_explicit'</config_time_integrator>
 
 <!-- hmix -->
@@ -65,6 +66,7 @@
 <config_hmix_scaleWithMesh ocn_grid="oRRS15to5">.true.</config_hmix_scaleWithMesh>
 <config_hmix_scaleWithMesh ocn_grid="oARRM60to10">.true.</config_hmix_scaleWithMesh>
 <config_hmix_scaleWithMesh ocn_grid="oARRM60to6">.true.</config_hmix_scaleWithMesh>
+<config_hmix_scaleWithMesh ocn_grid="EC30to60E2r2">.true.</config_hmix_scaleWithMesh>
 <config_maxMeshDensity>-1.0</config_maxMeshDensity>
 <config_hmix_use_ref_cell_width>.false.</config_hmix_use_ref_cell_width>
 <config_hmix_ref_cell_width>30.0e3</config_hmix_ref_cell_width>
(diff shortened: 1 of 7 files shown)


darincomeau:
@jonbob I think you'll need a CIME PR akin to https://github.com/ESMCI/cime/pull/3694 to set the coupling frequencies?

darincomeau (commented):
Do you want to set GM_kappa = 600 as default here?
Edit: I was trying to comment in a specific location, but its in `components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml`

jonbob:
Yes, I will have to make the CIME change. I thought I'd ask you about that and think about setting a better default for G-cases -- it's something crazy right now like ATM_NCPL = 1 right now. What would you suggest? 48? By the way, they're going to redo CIME soon, so I'm hoping this change can go into the new setup instead of the current one

jonbob:
@darincomeau - I'm not sure about GM_kappa. It's only been 600 for the meshes with ice cavities? But I'll check and see what Luke has been running with. Thanks for the careful review!

darincomeau:
> @darincomeau - I'm not sure about GM_kappa. It's only been 600 for the meshes with ice cavities? But I'll check and see what Luke has been running with. Thanks for the careful review!

Yes I noticed that too, at least by default. I'm also curious because I've currently set it to 600 for the ECwISC30to60E1r2 mesh, following the lead of oEC60to30v3wLI.

darincomeau:
> Yes, I will have to make the CIME change. I thought I'd ask you about that and think about setting a better default for G-cases -- it's something crazy right now like ATM_NCPL = 1 right now. What would you suggest? 48? By the way, they're going to redo CIME soon, so I'm hoping this change can go into the new setup instead of the current one

Ok that's good to hear on the new setup. I'd defer to you on a better default value. I think it's good to have to set it though so it's not overlooked, but it will be nice to not have to do a separate PR for that!

vanroekel:
@jonbob and @darincomeau I think we eventually will want to move to using visbeck for all cases and the bolus_kappa won't matter.  But for now, I'd set it to 600 for G-cases.  I'm not sure we want to go to that value for B-cases though

vanroekel:
@jonbob have you added the r2 files to the html server yet?  I tried running with this branch but had no luck due to many missing files.

darincomeau:
> But for now, I'd set it to 600 for G-cases. I'm not sure we want to go to that value for B-cases though.

Yeah we only set it to 600 by default for 'datm-forced' cases. Thanks for the confirmation.

xylar:
@jonbob, could you change "EC30to60E2v2" to "EC30to60E2r2" in the title and description of the PR?  I don't want there to be anymore confusion about the naming conventions for the new meshes than there already will be, given the very short, short and long names.  Thanks!

jonbob:
Thanks @xylar

jonbob:
@vanroekel - just adding them this morning. I wanted to test the configurations first but I can get the files staged right away. One question for you is about the ICG spin-ups -- how long and if I should get them started?

jonbob:
Corresponding CIME PR is made: [#3701](https://github.com/ESMCI/cime/pull/3701)

jonbob:
Testing for necessary files is going OK. I've been able to run on anvil with the following configurations:
* EC30to60E2r2_T62 CORE forced
* EC30to60E2r2_TL319 JRA forced
The trigrid one can't write necessary new mapping files in one directory on /home/ccsm-data/inputdata due to permissions problems -- I've sent email to the owner of it. So I'll try the trigrid on compy instead

vanroekel:
@jonbob I pushed changes to add the ICG spin. ups.  It seems to work okay.  but please take a look.

components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
```diff
@@ -846,6 +853,7 @@
 <config_AM_mocStreamfunction_enable ocn_grid="oRRS15to5">.false.</config_AM_mocStreamfunction_enable>
 <config_AM_mocStreamfunction_enable ocn_grid="oARRM60to10">.false.</config_AM_mocStreamfunction_enable>
 <config_AM_mocStreamfunction_enable ocn_grid="oARRM60to6">.false.</config_AM_mocStreamfunction_enable>
+<config_AM_mocStreamfunction_enable ocn_grid="EC30to60E2r2">.true.</config_AM_mocStreamfunction_enable>
```
darincomeau, line 856: @jonbob are you sure you want this set to true? In the ocn log file I'm getting the following written at every hour:
```
0001-01-01_01:00:00  WC time:    1599851079.141
 ocn_moc_streamfunction AM: config_AM_mocStreamfunction_region_group not found in regionGroupNames
 ocn_moc_streamfunction AM: setting regionGroupNumber=1
 ocn_moc_streamfunction AM: config_AM_mocStreamfunction_transect_group not found in transectGroupNames
 ocn_moc_streamfunction AM: setting transectGroupNumber=1
```

darincomeau (commented):
@jonbob I was able to run 5 day tests for the following:
`A_WCYCL1850S_CMIP6.ne30pg2_r05_EC30to60E2r2`
`GMPAS-IAF.T62_EC30to60E2r2`
`GMPAS-JRA1p4.TL319_EC30to60E2r2`

I cherry-picked your CIME branch so the G-cases had `ATM_NCPL` set correctly.

components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
jonbob: maybe not! I just saw that we have the file it would use, but that seems excessive. I'll go turn it off by default and anyone who wants that output can turn it on

vanroekel: I agree this should be off by default.  It will mess with mpas-analysis which won't work well (will miss the GM bolus contribution.

milenaveneziani: the MOC online AM does not add the GM bolus velocity contribution to the streamfunction calculation, so this should be off whenever GM is on.

milenaveneziani: It should be off for the EC60to30v3 too btw. And instead it should be 'true' for the RRS30to10.

jonbob: Thanks @milenaveneziani for the clarification. We'll have to make this changes in another PR. Or one of us can make in issue for that so it doesn't get lost over the weekend -- just let me know if you have time to do it or else I will.

milenaveneziani: @jonbob: just created an issue.

vanroekel (approved):
approved based on testing and inspection

darincomeau (approved):
Approved based on testing and visual inspection

components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
jonbob: Thanks @milenaveneziani . If you're happy with this PR, could you please add your review?

milenaveneziani (approved):

jonbob:
@milenaveneziani and @darincomeau - thank you both so much for the thorough reviews

jonbob:
merged to next

jonbob:
merged to master

PR #3823, opened 2020-09-10
Update CIME to master to support new ocn-ice grids

jonbob:
This PR updates the CIME hash from e0dcc59 to current master (84df628). This is needed to set coupling frequencies for new MPAS-Ocean/Seaice grids EC30to60E2r2 and WC14to60E2r3, support for which is being brought in with [PR #3819](https://github.com/E3SM-Project/E3SM/pull/3819). This update corresponds to [ESMCI PR #3701](https://github.com/ESMCI/cime/pull/3701) and does not bring in any other ESMCI updates.

[BFB]


jgfouca (approved):

jgfouca:
Thanks @jonbob ! The changes are small enough that you could do an expedited review and integration if you want.

jonbob:
Thanks @jgfouca - I'm used to just going through the next/master regular process, so I may stick with it. Plus the PR that will use it won't go onto next until maybe later today? I do appreciate your quick turnaround on all this

jonbob:
@rljacob - if you're around today, we would like to get this in along with the mesh PR it goes with

rljacob (approved):

jonbob:
merged to next

jonbob:
merged to master

PR #3824, opened 2020-09-10
Fix MMF History Variables

whannah1:
Clean up and rename MMF history variables from "SP*" convention to "MMF_*". Also fix the "timing factor" variable, now called the "subcycle factor" which has always given an incorrect (and useless) value. 

[BFB] except for an expected NLCOMP fail for ERS_Ln9_P96x1.ne4pg2_ne4pg2.F-MMF1.cam-crmout


mrnorman:
It looks fine to me.

mrnorman (approved):

PR #3832, opened 2020-09-16
Various MMF bug fixes

whannah1:
Fix a few minor issues with the MMF:
- revive standalone test framework
- aerosol optics syntax bug
- YAKL updates

[BFB]


components/cam/src/physics/crm/RNG_MT.F90
```diff
@@ -50,9 +50,8 @@
 !-----------------------------------------------------------------------
 !-----------------------------------------------------------------------
 module RNG_MT
-   use shr_kind_mod,  only: r8 => shr_kind_r8
-
    implicit none
+   integer, parameter :: r8 = selected_real_kind(13)
```
brhillman, line 54: Hard-coding this is to remove dependency on shr_kind_mod?

components/cam/src/physics/crm/openacc_utils.F90
```diff
@@ -8,7 +8,7 @@ module openacc_utils
   use cudafor
 #endif
   implicit none
-  integer :: ierr
+  integer, private :: ierr
```
brhillman, line 11: Usually prefer to see "private" by default for everything in the module, and then explicitly make certain variables/routines public. Is there a reason to instead assume everything is public here?

components/cam/src/physics/crm/sam/accelerate_crm.F90
```diff
@@ -20,11 +20,10 @@
 ! -----------------------------------------------------------------------------
 module accelerate_crm_mod
     use grid, only: nx, ny
-    use shr_kind_mod, only: r8=>shr_kind_r8
     use params, only: asyncid, rc=>crm_rknd
-
     implicit none
     public
+    integer, parameter :: r8 = selected_real_kind(13)
```
brhillman, line 26: I still think these type defs should be in a centralized place, maybe in `params` if not adopted from `shr_kind_mod`?

components/cam/src/physics/crm/openacc_utils.F90
rljacob: Yes all modules should have "private" as default.

components/cam/src/physics/crm/sam/accelerate_crm.F90
rljacob: Yes use the kind modifiers provided by shr_kind_mod.

whannah1: Ben, this should be resolved in the latest changes.

components/cam/src/physics/crm/openacc_utils.F90
whannah1: addressed in recent commit

components/cam/src/physics/crm/sam/MICRO_M2005/drop_activation.F90
```diff
@@ -13,9 +13,8 @@ module drop_activation
 ! July, 2009:  adopted from the module of ndrop used in CAM. 
 !
 !---------------------------------------------------------------------------------------------------- 
-   use shr_kind_mod, only: r8 => shr_kind_r8
    use modal_aero_data, only: ntot_amode
-   use params, only: crm_rknd
+   use params_kind
```
brhillman, line 17: Can we state explicitly which types from params_kind we are using here?

components/cam/src/physics/crm/sam/accelerate_crm.F90
```diff
@@ -20,19 +20,18 @@
 ! -----------------------------------------------------------------------------
 module accelerate_crm_mod
     use grid, only: nx, ny
-    use shr_kind_mod, only: r8=>shr_kind_r8
-    use params, only: asyncid, rc=>crm_rknd
-
+    use params, only: asyncid
+    use params_kind
```
brhillman, line 24: Can we say `use params_kind, only: crm_rknd` here for clarity?

components/cam/src/physics/crm/sam/crm_module.F90
```diff
@@ -24,6 +24,7 @@ module crm_module
   use damping_mod
   use ice_fall_mod
   use coriolis_mod
+  use params_kind
```
brhillman, line 27: As above, can we say `use params_kind, only: crm_rknd` here?

components/cam/src/physics/crm/RNG_MT.F90
brhillman, line 54: I think here we should use `use params_kind, only: r8`

components/cam/src/physics/crm/crm_rad_module.F90
```diff
@@ -8,9 +8,9 @@ module crm_rad_module
    ! update the radiation in the future, should we choose to put the radiation
    ! calculations on the CRM.
    use params_kind,       only: crm_rknd
-   use shr_kind_mod, only: r8 => shr_kind_r8
 
    implicit none
+   integer, parameter :: r8 = selected_real_kind(13)
```
brhillman, line 13: I think now that we have the single definition of `r8`, we should use `use params_kind, only: crm_rknd, r8` in the above use statement and eliminate this definition.

brhillman (approved):

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #3834, opened 2020-09-17
Change remove "_p" from area/lat/lon output variables

whannah1:
This changes the area/lat/lon variables in the history output for a physgrid run to not have the "_p" suffix. 

[BFB]

### DIFF
--- a/components/cam/src/dynamics/se/dyn_grid.F90
+++ b/components/cam/src/dynamics/se/dyn_grid.F90
@@ -700,10 +700,10 @@ contains
     if (fv_nphys>0) then
 
       gridname = 'physgrid_d'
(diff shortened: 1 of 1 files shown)


oksanaguba:
@whannah1 did you run pg2 tests or should I? will there be cprnc errors due to renaming?

whannah1:
I ran an ne4pg2 test and verified that the variable names were changed in the output.

whannah1:
I think the only test that compares history files is an MMF test (crm-out).I'm not sure if this will cause a diff in that file comparison or not.

oksanaguba:
Just in case I ran atm_integration and one of D cases on anvil. All is ok, in next.

PR #3839, opened 2020-09-20
Increase field widths in atm_chunk_cost.txt

### LINKED ISSUE #3838: fields  in atm_chunk_cost.txt not wide enough for some cases
The field width for the chunk id in the file atm_chunk_cost.txt is too small when the number of chunks is 500,000 or larger, resulting in the output of "*****". The field width for the process id in the file is also too small when the number of processes is 1,000,000 or larger.

Recent runs of high resolution SCREAM on Cori-KNL have exceeded this chunk count limit. 

A fix will be simple to implement.

worleyph:
The field width for the chunk id in the file atm_chunk_cost.txt is too
small when the number of chunks is 500,000 or larger, resulting in the
output of "*****". The field width for the process id in the file
is also too small when the number of processes is 1,000,000 or larger.

Here both of these conditions are checked for. If either is true, then
the field widths for chunk id and process id are increased enough
to support chunk and process counts up to 999,999,999 .

Fixes #3838 

[BFB]

### DIFF
--- a/components/cam/src/physics/cam/phys_grid.F90
+++ b/components/cam/src/physics/cam/phys_grid.F90
@@ -2539,6 +2539,10 @@ logical function phys_grid_initialized ()
                                       !  load balance, but not traditional
                                       !  speed-up (relative to serial cost)
 
+   logical :: wide_format             ! flag indicating that should use wide
+                                      !  format, supporting larger cid, 
+                                      !  lcid, and procid
+
    character(len=*), parameter :: fname = 'atm_chunk_costs.txt'
 
 !-----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


worleyph:
Sanity checks have been conducted on Cori-KNL using FC5AV1C-L with resolution ne4pg2_ne4pg2 and with a 96x1 PE layout, forcing the "wide_format" output by modifying the code appropriately (and then changing it back before committing).

worleyph:
@sarats and @ndkeen , please review this.

worleyph:
@rljacob , should I close this PR and create a new one based on current master (with cam renamed as eam)? Only one routine is modified, so will be straightforward.

rljacob:
Yes go ahead.

worleyph:
Replaced with PR #3865 .

PR #3812, opened 2020-09-04
HOMME: Kokkos implementation of theta-l target

bartgol:
This PR brings in the kokkos implementation of the theta-l target in HOMME. It is the implementation that we used for the 2020 Supercomputing/Gordon-Bell submission (paper accepted, not GB), with very few modifications following the merge of master into this branch.

After the merge, I manually verified that our unit tests (CXX vs F90 bfb comparison) pass on SKX and V100 architectures (KNL still to do). I am currently checking that also end-to-end tests are bfb (cxx vs f90), when run in debug mode. I also manually verified that the performance on V100 of the current branch (after master merge) has not changed compared to the version we used for the SC/GB submission.

Note: this PR also upgrades the version of kokkos in our external submodule. As a consequence, some machine files have to be updated (I haven't gotten to that yet).

[nonbfb] for homme suites (nonbfb suite -- sl executables; bfb suite - f theta executables, 1 sl test, new baselines needed)
[bfb] for EAM

PR #3840, opened 2020-09-20
Change options for homme anvil and sandiatoss3 bfb runs

oksanaguba:
This change is needed for PR #3812 .
Affects only homme anvil-bfb and skybridge-bfb runs.


rljacob:
This does not effect WC and can be merged to master during v2 freeze.

PR #3841, opened 2020-09-21
EAM: Remove Q from cam/dynamics/se restart.

ambrad:
Previously, the restart file contained both Q and Qdp. These fields are redundant and numerous. This PR removes Q. This will reduce restart file size and write and read time. This should benefit both SCREAM and RRM in v2.

Also remove some debug-build output from SL transport that I had meant to remove a while ago.

The model with this PR can restart from old restart files since this PR simply makes the read ignore Q. But of course old code won't be able to restart from the trimmed restart file.

[BFB]

### DIFF
--- a/components/cam/src/dynamics/se/restart_dynamics.F90
+++ b/components/cam/src/dynamics/se/restart_dynamics.F90
@@ -14,7 +14,7 @@ module restart_dynamics
 
   type(var_desc_t) :: FQpsdesc, omegadesc
 
-  type(var_desc_t), pointer :: qdesc(:), qdesc_dp(:)
+  type(var_desc_t), pointer :: qdesc_dp(:)
 
   integer :: ncol_dimid,  nlev_dimid, nlevp_dimid
   public :: init_restart_dynamics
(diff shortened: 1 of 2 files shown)


ambrad:
e3sm_atm_integration and e3sm_developer pass w.r.t. baselines created from today's trunk.

ambrad:
@ndkeen if interested, you should be able to bring this branch cleanly into your current SCREAM perf-analysis branch.

mt5555 (approved):

ambrad:
FC5AV1C-L cam.r file size is reduced by 37%.

PeterCaldwell:
37% is a much bigger improvement than I was expecting! This may have a noticeable effect on run times! Awesome work!

ambrad:
@PeterCaldwell that's for FC5AV1C-L cam.r, i.e., with 40 tracers. SCREAMS's 10 tracers will make the reduction more like 14%.

PeterCaldwell:
Ok, thanks for the extra info. I meant that this was a big win for v2, but 14% for SCREAM isn't bad either.

mt5555:
This suggests there are about 35 3D fields added to the restart file by the physics. 
@PeterCaldwell do you know why so many extra fields are needed?

ambrad:
@ndkeen reports ne256pg2 goes from 248.9 GB to 188.8 GB, so the new file size is 75.8% of the old one.

My ne30pg2 E3SMv1 number above should be interpreted as that the new file size is 64% of the old.

So those are the results for the two key configurations.

AaronDonahue:
@ambrad , this has now been merged to master.

ambrad:
Thanks!

PR #3845, opened 2020-09-22
Rename CAM to EAM in E3SM

rljacob:
Rename the source code directory "cam" to "eam" and rename all compsets to use "EAM" instead of "CAM" or "CAM5".
Remove all "CAM4" compsets.

Any submodule that was under "components/cam" is now under "components/eam".

Make all changes to CIME to allow this name change including config files, testmods, and build-namelist scripts.   Update cime module to commit that has the necessary config_files changes.

Also change restart and history files to have "eam" instead of "cam" in their names.

[BFB]


rljacob:
e3sm_atm_developer --no-build passes.   Only one test in e3sm_integration fails to setup.

SMS_Ln1.ne4_ne4.FC5AV1C-L.anvil_intel.eam-chem_pp fails during case.setup with:
```
ERROR: Command: '/gpfs/fs1/home/jacob/E3SM/components/eam/bld/configure -s -ccsm_seq -ice none -ocn docn -caseroo
t /lcrc/group/acme/jacob/acme_scratch/anvil/SMS_Ln1.ne4_ne4.FC5AV1C-L.anvil_intel.eam-chem_pp.20200922_160257_592
05b -comp_intf mct  -spmd -spmd -smp -nosmp -dyn se -dyn_target preqx -res ne4np4   -mach anvil -phys cam5 -clubb
_sgs -microphys mg2 -chem linoz_mam4_resus_mom -rain_evap_to_coarse_aero -nlev 72 -clubb_sgs -microphys mg2 -chem
 linoz_mam4_resus_mom_soag -rain_evap_to_coarse_aero -nlev 72 -bc_dep_to_snow_updates -usr_mech_infile /gpfs/fs1/
home/jacob/E3SM/cime/../components/eam/chem_proc/inputs/pp_linoz_mam4_resus_mom_soag.in ' failed with error 'tar:
 /lcrc/group/acme/jacob/acme_scratch/anvil/SMS_Ln1.ne4_ne4.FC5AV1C-L.anvil_intel.eam-chem_pp.20200922_160257_5920
5b/Buildconf/eamconf/chem_proc/cam.subs.tar: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now

```
Anyone know what the fix might be?  Something the chem_proc must still be expecting "cam" or "camconf".

rljacob:
@wlin7 or @tangq can you help with the above?

rljacob:
Still todo:  change output filenames, pass e3sm_integration.

wlin7:
> @wlin7 or @tangq can you help with the above?

Looking into it now.

wlin7:
@rljacob , a fix has been pushed to the branch. The perl module controlling the build of chem preprocessor (now in components/eam/bld/perl5lib/Build/ChemPreprocess.pm) still point the path to components/cam/. As a result, the build of chem preprocessor failed. The tar file shown in the error message would only be created after the campp is built and executed.

Please note that in ChemPreprocess.pm, 'cam' is till used in several occasions, including a path to components/eam/chem_proc/procfiles/cam. The preprocessor is still called campp. Should we rename them as well now?

components/eam/chem_proc/src/Makefile would also need to be updated if to rename all now.

wlin7:
The test SMS_Ln1.ne4_ne4.FC5AV1C-L.anvil_intel.eam-chem_pp would still fail during the build.
For one, cmake does not try to compile the codes prepared by the chem_pp, for example Buildconf/eamconf/chem_proc/source/chem_mods.F90. I stop trying now.

rljacob:
Thanks @wlin7 .   I forgot about the other "cam" instances in subdirectories.  Lets leave those alone for now unless, like CIME, they need to match the name "eam".

wlin7:
Sounds good, @rljacob . Those 'cam' instances are secondary and the top-level configure/build structure do not have to  pay attention to it. 

The build failure with the chem_proc test likely would also happen to the build for other tests. Specifically for this test, the compilation of some chemistry modules requires module chem_mods.  With master, cmake knows to compile it using Buildconf/camconf/chem_proc/source/chem_mods.F90 before compiling other chem modules. I was expecting after renaming to eam, cmake would look for Buildconf/eamconf/chem_proc/source/chem_mods.F90 to compile. But I don't see it in the bldlog.

tangq:
@wlin7 and @rljacob , sorry for the delay. I am working on 3 abstracts for the PI meeting next month.

But I don't know much about Perl and cmake. Is there anything else I can try?

rljacob:
@wlin7 I'm still getting an error at setup time but its a different error.  Did you push all your changes? 

```
ERROR: Command: '/gpfs/fs1/home/jacob/E3SM/components/eam/bld/configure -s -ccsm_seq -ice none -ocn docn -caseroot /lcrc
/group/acme/jacob/acme_scratch/anvil/SMS_Ln1.ne4_ne4.FC5AV1C-L.anvil_intel.eam-chem_pp.20200923_142426_4vftst -comp_intf
 mct  -spmd -spmd -smp -nosmp -dyn se -dyn_target preqx -res ne4np4   -mach anvil -phys cam5 -clubb_sgs -microphys mg2 -
chem linoz_mam4_resus_mom -rain_evap_to_coarse_aero -nlev 72 -clubb_sgs -microphys mg2 -chem linoz_mam4_resus_mom_soag -
rain_evap_to_coarse_aero -nlev 72 -bc_dep_to_snow_updates -usr_mech_infile /gpfs/fs1/home/jacob/E3SM/cime/../components/
eam/chem_proc/inputs/pp_linoz_mam4_resus_mom_soag.in ' failed with error '*** Chem preprocessor FAILED.
 See: /lcrc/group/acme/jacob/acme_scratch/anvil/SMS_Ln1.ne4_ne4.FC5AV1C-L.anvil_intel.eam-chem_pp.20200923_142426_4vftst
/Buildconf/eamconf/chem_proc/chem_mech.doc' from dir '/lcrc/group/acme/jacob/acme_scratch/anvil/SMS_Ln1.ne4_ne4.FC5AV1C-
L.anvil_intel.eam-chem_pp.20200923_142426_4vftst/Buildconf/eamconf'
```

rljacob:
@tangq you can just look for other places in the chemistry build/configure that assume the source code path has "cam" at the top.

rljacob:
@wlin7 nevermind.  I had some junk in my working copy from a previous attempt to fix this.  I can get past the setup.

rljacob:
@wlin7 I can also get a successful build with just your changes.  Did you see a build fail for that case?

rljacob:
For those of you following along, the other problem in this branch is that flds_co2a is false in drv_in.  It should be true for almost all our cases.  You can see it right after case.setup by looking at drv_in in CaseDocs but the error doesn't occur until runtime when index_x2l_Sa_co2diag is 0.

wlin7:
> @wlin7 I can also get a successful build with just your changes. Did you see a build fail for that case?

@rljacob , nice to know you are able to build successfully. Possible your local branch actually include some extra changes? Up to the changes I made yesterday, my test still failed during build, on both anvil and cori-knl. bldlog from anvil below
`/lcrc/group/acme/ac.wlin/acme_scratch/anvil/SMS_Ln1.ne4_ne4.FC5AV1C-L.anvil_intel.eam-chem_pp.20200923_225554_f56mpl/bld/e3sm.bldlog.200923-225944`

Could there be some racing? Does not matter though, now that you can build and continue updating the branch.

rljacob:
Still stuck on the flds_co2a problem so here's how to reproduce for anyone who wants to help.
Check out this branch and make a simple F-case:
```
 ./create_newcase --case Ftest --compset FC5AV1C-L --res ne4_ne4 
cd Ftest
./case.setup
```
now do "grep flds_co2a CaseDocs/drv_in" and you'll get:
`  flds_co2a = .false.`
It should be true.

What's supposed to happen is when cime/src/drivers/mct/cime_config/buildnml is called to build drv_in it processes the config_components_e3sm.xml for the driver (in that same mct/cime_config dir) and this block of code:
```
  <entry id="CCSM_BGC">
    <type>char</type>
    <valid_values>none,CO2A,CO2A_OI,CO2B,CO2C,CO2C_OI,CO2_DMSA</valid_values>
    <default_value>none</default_value>
    <values match="last">
      <value compset="_EAM\d+"     >CO2A</value>
```
Should result in CCSM_BGC being set to CO2A because the FC5AV1C-L compset has "EAM" in it.  But it stays "none" which makes flds_co2a false.

klindsay28:
@rljacob, I think the `\d+` in `"_EAM\d+"` will only match if `EAM` has trailing digits.
Did you drop version number suffixes with the switch to `EAM`?
Does this work if you drop the `\d+`?

rljacob:
That did it @klindsay28!  Thanks!   I may have blindly tried that earlier but at the time didn't realize one had to recreate the case for the change to take effect.

rljacob:
@wlin7 now I am seeing the build error with chem_pp case.

@jgfouca I'm seeing build problems with cases that have chem_pp or cosp.   IIRC, those have their own little builds.
The chem_pp case has errors like:
```
/qfs/people/jaco316/E3SM/components/eam/src/chemistry/mozart/mo_tracname.F90(8): error #7002: Error in opening the compiled module file.  Check INCLUDE paths.   [CHEM_MODS]
      use chem_mods, only : grpcnt, gas_pcnst
```
cosp cases have:
```
/qfs/people/jaco316/E3SM/components/eam/src/physics/cam/cospsimulator_intr.F90(21): error #7002: Error in opening the compiled module file.  Check INCLUDE paths.   [QUICKBEAM]
  use quickbeam,            only: radar_cfg
```
You can reproduce with
SMS.ne4_ne4.FC5AV1C-L.compy_intel.eam-cosplite
SMS_Ln1.ne4_ne4.FC5AV1C-L.compy_intel.eam-chem_pp

rljacob:
Also having a build error with the MMF SAM++ case:
```
/qfs/people/jaco316/E3SM/components/eam/src/physics/crm/crm_history.F90(378): error #7002: Error in opening the compiled module file.  Check INCLUDE paths.   [CRM_ECPP_OUTPUT_MODULE]
   use crm_ecpp_output_module, only: crm_ecpp_output_type
```
ERS_Ln9_P96x1.ne4pg2_ne4pg2.F-MMFXX.compy_intel

rljacob:
Found some cam/src references in a couple of cmake files and changing those fixes the MMFXX and cosp2 cases.  Still haven't build errors with the chem_pp case.

tangq:
@rljacob , the chem_pp case can have two different building ways: inline and offline. Which one did you use?

The inline method uses the -usr_mech_infile option to point to a chemistry input mechanism file (*.in), which will be used by the chemistry pre-processor as input to generate the fortran code at $casedir/Buildconf/camconf/source.

The offline method uses the existing fortran code at chemistry/pp_* folders.

The chemistry pre-processor has its own make file at cam/chem_proc/src. But it's already below the cam (eam) level, so it shouldn't matter.

You are likely using the offline method. I suspect that cam/bld/configure has some hard-coded references to the cam folder. The configure perl script needs to identify the correct pp_* folder for different chemistry schemes. The fortran code in the pp_* folders defines `use chem_mods, only : grpcnt, gas_pcnst`. But I don't know perl to narrow down the cause.

jgfouca:
@rljacob , looking at this now. I didn't even make it as far when I tried this on mappy:
```
CMake Error at cmake/cmake_util.cmake:53 (message):
  Trying to set flags on non-existent source: cam/src/dynamics/fv/geopk.F90,
  looked for /home/jgfouca/E3SM/components/cam/src/dynamics/fv/geopk.F90
```

jgfouca:
OK, some the Depends* files still needed to be updated. I pushed.

jgfouca:
@rljacob , SMS.ne4_ne4.FC5AV1C-L.mappy_gnu.eam-cosplite now works for me.

rljacob:
@jgfouca were you able to get SMS_Ln1.ne4_ne4.FC5AV1C-L.{mach}_{compiler}.eam-chem_pp to build anywhere?

@tangq its the inline method.  This is added to CAM_CONFIG_OPTS:
`./xmlchange --append CAM_CONFIG_OPTS='-usr_mech_infile $CIMEROOT/../components/eam/chem_proc/inputs/pp_linoz_mam4_resus_mom_soag.in'`

rljacob:
I've now been able to get SMS_Ln1.ne4_ne4.FC5AV1C-L.anvil_intel.eam-chem_pp to build twice so it must be fixed.  @wlin7 can you try that test again?

tangq:
@rljacob , for the inline method the CAM_CONFIG_OPTS was updated correctly, it should work.

Glad to know it works for you on anvil now. Is it the commit updating Depends files from @jgfouca the fix for it?

wlin7:
> I've now been able to get SMS_Ln1.ne4_ne4.FC5AV1C-L.anvil_intel.eam-chem_pp to build twice so it must be fixed. @wlin7 can you try that test again?

I don't have the luck, @rljacob . With the branch updated thru @jgfouca 's push this afternoon, 08e19d9, I still got the same compilation error -- chem_mods still not exist when it is used during compiling mo_tracname.F90

The build log is
`/lcrc/group/acme/ac.wlin/acme_scratch/anvil/SMS_Ln1.ne4_ne4.FC5AV1C-L.anvil_intel.eam-chem_pp.20200925_175448_zg9nmf/bld/e3sm.bldlog.200925-175917`

And the command to run the test is 
`./create_test SMS_Ln1.ne4_ne4.FC5AV1C-L.anvil_intel.eam-chem_pp   --machine=anvil --walltime 01:30:00`

jgfouca:
@rljacob , @wlin7, I'm having trouble logging into blues at the moment, so I can't see the build log. That said, looking at `physics/crm/crm_history.F90`, this file is ripe for failed cmake dependency scans. Try moving the use statements to the top of the module and see if that helps it build.

rljacob:
Ok but its a chemistry case, not a crm case, that is causing issues.  Please try SMS_Ln1.ne4_ne4.FC5AV1C-L.*.*.eam-chem_pp on any platform so I can get a third data point.

rljacob:
Actually try compy.  I got a build error from SMS_Ln1.ne4_ne4.FC5AV1C-L.compy_intel.eam-chem_pp

rljacob:
On compy, I'm comparing content of bld/cmake-bld/cmake/atm/CMakeFiles/atm.dir between a chem_pp test from master and from the head of this branch.  "chem_mods" is missing from "fortran.internal" in the failing test.

wlin7:
> On compy, I'm comparing content of bld/cmake-bld/cmake/atm/CMakeFiles/atm.dir between a chem_pp test from master and from the head of this branch. "chem_mods" is missing from "fortran.internal" in the failing test.

Good finds, @rljacob . With this clue, and given that Filepath has eamconf/chem_proc/source in proper order, I started to think the problem has nothing to do with the  build system system (including cmake). Since all other files under the same directory eamconf/chem_proc/source are listed in fortran.internal; the problem may be with the file chem_mods.F90. Further check shows in the eam-chem_pp test, chem_mods.F90 is empty. It must be that campp execution was not successful. We were misled by the passing of the campp step. I am checking on that,

wlin7:
A quick test: copying chem_mods.F90 from cam-chem_pp test is copied into the source directory for eam-chem_pp test, the build is ok. Checking why eam-chem_pp test failed to generate content for chem_mods.F90.

rljacob:
I'm also trying to understand why the test works for me on anvil but not on compy and it doesn't work for @wlin7  on compy.  The only difference so far: the shell.   I use tcsh on anvil and @wlin7 has bash.  I have bash on compy.

singhbalwinder:
Are you guys using PGI compiler or intel? I am asking this because I had issues with the PGI compiler in the past. May be try SMS_D test instead of SMS so that compiler uses no optimizations.

wlin7:
@singhbalwinder , my tests use intel compiler, both on anvil and cori-knl.

@rljacob , for your successful builds, can you please check if Buildconf/eamconf/chem_proc/source/chem_mods.F90 not  empty? It must not be empty since your tests were ok.

I tracked down the problem somewhat further. campp is built and executed during eam's configure command. Build is ok. But running campp looks for file eam/chem_proc/procfiles/cam/mo_chem.mod.

This file does not exist in the branch rljacob/eam/change-name I checked out. So the run failed. Rob, can you check if the file exist in your branch? It exists in the master branch. There are two other plain files under cam/chem_proc/procfiles  that do not exist in the eam counterpart. Why these files disappear? 

Balwinder, can you please also check out the branch and see if the following files exist
 
```
chem_proc/procfiles/cam/mo_chem.mod
chem_proc/procfiles/mo_chem.mod
chem_proc/procfiles/mo_grid.mod
```
I was concerned I could have messed up something during testings. But a fresh clone and checking out the branch, these files still do not exist. Rob has an interesting comment on SHELL used. The execution of campp does assume a shell type. Not sure if it matters or not.

A wild guess, Rob, could that be possible that these files were not added to git in your development branch, thus not pushed to remote? I guess in your compy repo, the files do not exist either. Even if it is so, how it happens seems mysterious.

wlin7:
@rljacob , after copying the three files from master to my local repo for rljacob/eam/change-name, the test went well. Should I update the branch or leave it to you?  Curious how this happened.

rljacob:
You can see my successful build (and run) in /lcrc/group/acme/jacob/acme_scratch/anvil/SMS_Ln1.ne4_ne4.FC5AV1C-L.anvil_intel.eam-chem_pp.20200926_224726_ljx625.

I checked and  Buildconf/eamconf/chem_proc/source/chem_mods.F90 is not empty.

These files that end in ".mod" should not have that extension.  They are regular Fortran files and should end in .F90.  The ".mod" extension is reserved for the output of compiling a module and is something that should ever be added to a repo.  Our top-level .gitignore file say to ignore anything with .mod at the end.  Those files are present on master on my laptop (cam, bash) and in my clone on anvil (eam, tcsh) but not my clone on compy (eam, bash).  Rename them to .F90 and see if your test goes ok.

singhbalwinder:
Thanks @wlin7  for figuring this out. This is a great find! I would also think that `*.mod` files would be binary files and should not reside in the source code due to reasons @rljacob  mentioned. I checked the master code and they do exist in the repo(see [here](https://github.com/E3SM-Project/E3SM/tree/master/components/cam/chem_proc/procfiles)).

Perhaps we should change the files names to `*_mod.info` instead of `.mod` (or something similar). These files contain cam specific module info, so they are named as `*.mod` (which is not a good practice and can cause confusion). We would also need to change the following files to update them with the new file names:
```
chem_proc/bkend/cam.mod.files.PP
chem_proc/bkend/mozart.mod.files.PP
chem_proc/bkend/wrf.mod.files.PP
```
I didn't run a test, I found the above files by a simple grep.

rljacob:
An experiment for @wlin to try:  change your shell to tcsh (just type "tcsh") and do a fresh clone of the branch and see if those 3 .mod files show up.

rljacob:
I think I know why it works just for me in my clone on anvil. When I started this work on anvil, I cloned master (which had the .mod files) and did the following for the directory rename:
>  mv cam eam
> git add eam
> git rm -R cam
> git commit 

This was recommended somewhere for renaming an entire directory in git.   The unix "mv" preserved the .mod files but the git commit did NOT commit them because .mod is in the .gitignore file.   So they are not in the branch and anyone who checks out the branch won't have them and the test won't work.  I have them only in that particular clone (~jacob/E3SM on anvil) because its the one where I did the cam to eam move.  "git status" never alerted me to them because they are ignored.

The files must have been in the repo (in cam) before the .gitignore was added.   You can keep checking out files that were added before they were made ignorable.

So yes adding them back is the fix but you won't be able to commit them if they end in ".mod".

singhbalwinder:
Yes, that explains it. We probably should consider renaming these files and these files can end with *.in extension as most of the files for F90 source generation in E3SM uses *.in extension.

wlin7:
Thanks for the explanation, Rob. I understand now. Agreed with Balwinder we should just rename those files. They are actually just text files. Maybe extension .inc is good to use, or _mod.

wlin7:
Working on the fix as @singhbalwinder suggested: change the names to .in, and update the several .PP files accordingly. To push to remote after a test.

rljacob:
How are those .mod files currently being used in the code on master?   If its with Fortran90 "use" I don't know how that ever worked since they are F90 source files with a .mod extension, not a compiled module file.

singhbalwinder:
Chemistry preprocessor generates F90 files which are then compiled by the model. We typically run chemistry preprocessor only once and save the F90 files for subsequent model runs. The *.mod files are used as an "input" for generating F90 files (just like *.in files are used to generate F90 files using genf90 perl script). These files are not directly used for compilation (or used in "use" statement), they rather serve as an input for generating F90 files.

wlin7:
After changing *.mod to *.in and the companion changes in the .PP files, the test built is ok.

But not fully solved yet. campp runs ok, generating a file ".F90", yes, just .F90 instead of chem_mods.F90. Though the build system compiled it into chem_mods.mod, it shouldn't go that way. Haven't figured out where the name is defined from campp.  The test can proceed any way. (hope for other systems, too). Will fix it later.

singhbalwinder:
I did another grep and found out that these files are also being used in `cam/chem_proc/src/cam_chempp/mozpp.main.f` file. The use of ".mod" should be changed to ".in" in this file. I hope it will work after this change. My naive greps did not find any other place where these files are being used.

wlin7:
Nice catch, @singhbalwinder . These are definitely relevant. However, I hesitate what need to be changed. 

After changing '.mod' to '.in' in string pattern matches at two spots with context of setting temp filename,  including one for "chem_mods", campp does generate eamconf/chem_proc/source/chem_mods.F90, but still also generates a file ".F90" containing  module m_rxt_id. 

If changing just one pattern matching that affect setting temp name chem_mods.F90,  no ".F90" file generated. 

The test would build ok with changes in one or two spots. Morever, there are also other pattern matching for '.mod' in the code. I can't tell if they need to be changed.

Pushed a commit with change on just one spot now. @tang please take a look what to do with the other pattern matching of ".mod" in mozpp.main.f.

rljacob:
The chem_pp now passes for me on compy.  This is ready to merge and will go in today.

singhbalwinder (approved):
I didn't go through each and every file but if the tests are passing, I think everything should be good.

wlin7 (approved):
I went thru all the changes and don't see any anomaly.  Since all tests have passed, this should be ready. 'randomly' select allactive  and eam-cosplite for testing and both went well.

rljacob:
e3sm_integration passes on compy,intel with this branch.

rljacob:
merged to next

worleyph:
@rljacob , I just now tried next to see if the CAM => EAM broke anything with the performance archiving. Looks fine. The only think that I noticed was that the performance archive subdirectory for a job using current master saves files with the following names:

     chem_mech.doc
     chem_mech.in

(for an FC5AV1C-L case), but the job using next did not save files with these names (and no new files were saved when using next, so two fewer files were saved). These were in CaseDocs .

rljacob:
Thanks Pat.  We'll have to add that back with a future PR.  @wlin7 do you know what copied those to CaseDocs?

wlin7:
Need to look into the scripts, @rljacob. I can see multiple instances tin the script that chem_mech.* files are copied to $cam_bld, but haven't found one with CaseDocs as the destination. Will work on this fix.

wlin7:
Amazing, Pat. Nothing can escape your eyes. This turns out to be something in cime

cime/scripts/lib/CIME/case/preview_namelists.py (at the end)
 ```
# Copy over chemistry mechanism docs if they exist
    if (os.path.isdir(os.path.join(casebuild, "camconf"))):
        for file_to_copy in glob.glob(os.path.join(casebuild, "camconf", "*chem_mech*")):
            safe_copy(file_to_copy, docdir)
```
We certainly can't just change camconf to eamconf in this file. Need to have an equivalent action in a script outside cime.

**Comment by rljacob** [comment]
Thanks @wlin7.  I can make a PR with that change.

**Comment by rljacob** [comment]
@wlin7 actually that is beyond my python skills.  Can you open an issue in ESMCI/cime ?

**Comment by wlin7** [comment]
Thank you @rljacob . I actually don't have a quick clue how to extract the atm name within this script. From COMP_CLASSES with some  conditional check?

**Comment by wlin7** [comment]
Same here. I would be more confident if it is a perl script. I am going to open a case.

**Comment by wlin7** [comment]
A feasible way is to parse the COMPSET name to extract the atm comp name. Something like the following can do it

```
  # Copy over chemistry mechanism docs if they exist
   compset = self.get_value("COMPSET")
   atmcompname = (compset.split('%')[0]).split('_')[1].lower()
   atmfamilyname = re.sub('\d+','',atmcompname)
   if (os.path.isdir(os.path.join(casebuild, atmfamilyname+"conf"))):
       for file_to_copy in glob.glob(os.path.join(casebuild, atmfamilyname+"conf", "*chem_mech*")):
           safe_copy(file_to_copy, docdir)

```

wlin7:
This is fixed by @jedwards4b via https://github.com/ESMCI/cime/pull/3725. It is much cleaner than I posted above. I grepped for just the upper case comp name among the xml files, and only COMPSET carries that. Thanks @jedwards4b .

PR #3846, opened 2020-09-22
Switch on sea ice defaults and snicar on lnd to match snow on sea ice

proteanplanet:
This pull request switches on several new features in the sea ice and land models for E3SM V2 that are already on master and have been tested in multiple B-case simulations, individually and together:

1) Switch on snicar on sea ice as the new default in mpassi_in:  
config_use_snicar_ad = true

2) Switch on snicar in elm so that radiative transfer over snow on land and sea ice are identical:
use_snicar_ad = .true.

3) Increase the number of snow layers on sea ice from 1 to 5 layers:
config_nsnowlayers = 5

4) Switch on the snow morphology and adjust defaults:
config_fallen_snow_radius = 54.526
config_snow_redistribution_scheme = 'ITDrdg'
config_use_column_snow_tracers = true
config_use_effective_snow_density = true
config_use_snow_grain_radius = true
config_use_snow_liquid_ponds = true

The changes have been tested in a 30 year B-case simulation on Compy, as indicated here:
https://acme-climate.atlassian.net/wiki/spaces/EWCG/pages/1638957185/20200720.snow5Snicar2.A+WCYCL1850S.ne30+r05+oEC60to30v3.compy

They have also passed a five-day B-case smoke test on this branch in the following configuration on Cori:
-compset A_WCYCL1850 -mach cori-knl -res ne30pg2_r05_EC30to60E2r2

[NML]
[non-BFB]

### DIFF
--- a/components/elm/bld/CLMBuildNamelist.pm
+++ b/components/elm/bld/CLMBuildNamelist.pm
@@ -99,15 +99,15 @@ OPTIONS
                                 bgc   = Carbon Nitrogen with methane, nitrification, vertical soil C,
                                         CENTURY decomposition
                                         (or CLM45BGC if phys=clm4_5/clm5_0, use_cn=true, use_vertsoilc=true,
-                                         use_century_decomp=true, use_nitrif_denitrif=true, and use_lch4=true,
-                                         use_dynroot)
+                                         use_century_decomp=true, use_nitrif_denitrif=true, use_lch4=true,
+                                         and use_snicar_ad=true, use_dynroot)
                                         This toggles on the namelist variables:
-                                         use_cn, use_lch4, use_nitrif_denitrif, use_vertsoilc, use_century_decomp,
+                                         use_cn, use_lch4, use_nitrif_denitrif, use_vertsoilc, use_century_decomp, use_snicar_ad
                                          use_dynroot
                                 fates    = functionaly assembled terrestrial ecosystem simulator
                                           with native below ground bgc
                                           This toggles on the namelist variables:
-                                          use_fates, use_vertsoilc, use_century_decomp
+                                          use_fates, use_vertsoilc, use_snicar_ad, use_century_decomp
 
      -bgc_spinup "on|off"     CLM 4.5 Only. For CLM 4.0, spinup is controlled from configure.
                               Turn on given spinup mode for BGC setting of CN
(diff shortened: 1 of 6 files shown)


proteanplanet:
> @proteanplanet : This looks good. One question. Do the default changes need to be made in mpassi Registry.xml?

@njeffery: Good question.  Not for the coupled model, which uses the XML files changed here to set the namelist, but I anticipate we would also want to make this the default in stand-alone MPAS-SI for consistency.  Perhaps @akturner has a view on this?  I've just made him a reviewer.  I can submit a separate PR to MPAS develop with this change. It won't affect the coupled model, but will make things consistent for the curious explorer.

njeffery (approved):
I've reviewed the namelist changes for sea ice and approve.  These are the best parameters for a suite of  BCASE runs completed.  Recommend that the namelist changes be added the mpassi Registry.

akturner (approved):
This looks great. I think it would be good to make this the standalone MPAS-Seaice default in registry and add it to the standard configurations, while preserving the old default.

vanroekel (approved):
approved by visual inspection

vanroekel:
@proteanplanet  and @jonbob not sure if this matters but clm is being changed to elm, you may need to rebase here?

proteanplanet:
> @proteanplanet and @jonbob not sure if this matters but clm is being changed to elm, you may need to rebase here?

@vanroekel: This was deliberately created after the switch to ELM.  It's covered.

jonbob:
merged to next

jonbob:
@proteanplanet and @njeffery - I wasn't thinking and this PR probably needs to include updates to the _ICG restart files we use for spunup initial conditions. In the meantime, _ICG tests and runs will fail. @rljacob, we can revert this from next or expect that all tests with _ICG will fail until we add new spunup files that include the new number of snow layers in seaice

rljacob:
How long will new spinups take?  They don't have to be perfect for testing.   Its 4 cases so you can leave  them in a fail for now.

jonbob:
@rljacob - we've only been spinning up G-cases for a month lately. I propose I do something like that for the oEC60to30v3 grid so the _ICG tests will pass and include it in this PR. I've submitted a case this morning and will modify the restart files, get them into the inputdata repo, and point at them in a re-merge of this PR today. That way tonight's testing can clear the FAILs and make them into DIFFs. And once all the ocean and seaice changes are solidified and on master, which probably will go up to the last minute of the deadline, I think we should spinup all the resolutions that have ICG configurations using the new code and settings. Does that sounds reasonable?

rljacob:
Sounds good to me.

jonbob:
re-merged to next to include new oEC60to30v3 ICG files, which are staged on the inputdata server

jonbob:
The ERP_Ld3.ne30_oECv3_ICG.A_WCYCL1850S tests now DIFF instead of FAIL

jonbob:

ndkeen:
I think this will fix https://github.com/E3SM-Project/E3SM/issues/3894

rljacob:
No this was merged 10 days ago and caused #3894.

PR #3847, opened 2020-09-23
Add script support for new ocn/ice WC14to60E2r3 grid

jonbob:
Bring in the initial support for the new ocn/ice WC14to60E2r3 grid, which has enhanced resolution of 14km along the coast of North America, the Northern Atlantic subpolar gyre, and the Arctic Ocean. The high resolution regions smoothly transition to the background resolution of the standard low resolution 60to30km grid. All new mapping and other required files have been staged on the inputdata server, with the exception of seaice and ocean initial condition files for the spunup _ICG configuration.

### DIFF
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -2210,7 +2210,7 @@
       <file grid="atm|lnd" mask="oARRM60to10">$DIN_LOC_ROOT/share/domains/domain.lnd.T62_oARRM60to10.180716.nc</file>
       <file grid="atm|lnd" mask="oARRM60to6">$DIN_LOC_ROOT/share/domains/domain.lnd.T62_oARRM60to6.180803.nc</file>
       <file grid="atm|lnd" mask="EC30to60E2r2">$DIN_LOC_ROOT/share/domains/domain.lnd.T62_EC30to60E2r2.200908.nc</file>
-      <file grid="atm|lnd" mask="WC14to60E2r3">$DIN_LOC_ROOT/share/domains/domain.lnd.T62_WC14to60E2r3.200714.nc</file>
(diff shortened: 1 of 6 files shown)


vanroekel (approved):
approved based on visual inspection

jonbob:
Updated with new mapping/domain files after problem with SCRIP file generation found. All new files have been staged on the inputdata repo and the old (potentially bad) ones removed

golaz:

jonbob:
@maltrud - could you please give the default settings an especially good look, and then review once we resolve them? Thanks

maltrud:
@jonbob i found one that needs to be added:
<config_hmix_use_ref_cell_width>.false.</config_hmix_use_ref_cell_width> should be true.
also, i haven't messed with the time steps very much.  it may be possible for baroclinic to be 15 min, and barotropic to be larger, too, but that can wait.

jonbob:
@maltrud -- just for this mesh? Or default for all resolutions? I just checked master and it also has <config_hmix_use_ref_cell_width>.false.</config_hmix_use_ref_cell_width>

maltrud:
@jonbob sorry for not being clear.  i mean just for this mesh.  changing for others would not be BFB and is untested.

jonbob:
@maltrud - thanks for the clarification. OK, that is reset if you can take another peek

maltrud:
thanks, @jonbob that looks good.

maltrud (approved):

jonbob:
rebased
passes SMS.ne30pg2_r05_WC14to60E2r3-1900_ICG.A_WCYCL1850S.anvil_intel

jonbob:
merged to next

jonbob:
merge passes SMS.ne30pg2_r05_WC14to60E2r3_ICG.A_WCYCL1850S.anvil_intel

jonbob:
merged to master

PR #3850, opened 2020-09-25
Added prescribed ice mode to MPAS-Seaice component

akturner:
Added prescribed ice mode to MPAS-Seaice component. This involves modifications to MPAS-Seaice as well as modifications in E3SM.

### DIFF
--- a/components/mpas-seaice/bld/build-namelist
+++ b/components/mpas-seaice/bld/build-namelist
@@ -58,6 +58,8 @@ OPTIONS
                            Options are: free, non-free
      -iceberg_mode         variable for iceberg mode
                            Options are: none, data, prognostic
+     -prognostic_mode      variable for prognostic mode
+                           Options are: full, prescribed
      -ntasks_ice           NTASKS_ICE for this case
      -ninst_ice            NINST_ICE for this case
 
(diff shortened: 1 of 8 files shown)


jonbob:
@akturner - just a couple of things about this PR. First, let me use the automated scripts to regenerate the bld-namelist scripts -- just because we'll get rid of some unexpected diffs. But the file changes in mpas-source also show some ocean files? That doesn't seem right...

jonbob:
OK, the ocean changes in the mpas-source diff was just the wrong base for comparison, so that's not an issue. But many of the changes you've added by hand in the mpas-seaice bld-namelist files don't get automatically generated using the scripts, so either a reference to them is missing from Registry or something is broken. I'm not sure how to proceed

proteanplanet (approved):
This appears to be correct on inspection.  Importantly, if there are problems, they don't appear to affect any other configuration.  

I would like to make it clear that this _will not_ produce the same results as using CICE in E3SM F-case simulations, and nor should it be expected to.  MPAS-SI includes a number of improvements and fixes that do not exist in the current implementation of CICE in E3SM.   The results that MPAS-SI produces would most likely be more trustworthy than our current prescribed ice mode.

(changes to namelist_defaults_mpassi.xml  and buildnml will be merged with changes from https://github.com/E3SM-Project/E3SM/pull/3846 and coming from branch https://github.com/E3SM-Project/E3SM/tree/proteanplanet/polar/v2output)

jonbob:
merged to next. @akturner - we can talk on Monday if you're happy with the changes. We can revert this PR and rework if necessary

jonbob:
reverted from next -- the build-namelist changes made the new prescribed ice mode non-functional so it will have to be reworked so the automated scripts can generate the build-namelist files correctly. It also caused some build failures using gnu

akturner:
Will redo another branch with changes requested

PR #3857, opened 2020-09-28
Sea ice model output for E3SM V2

proteanplanet:
This pull request is to align MPAS-SeaIce output with data requirements for E3SM Version 2.   Both the Daily and Monthly streams have been switched to single precision, and the total output amounts to 87/Month 899M/month, a total reduction of

regionalStatistics ~26K/month
timeSeriesStatsDaily ~200M/month
timeSeriesStatsMonthly ~100M/month

Roughly 3.6GB/Year on the EC30to60E2r2 mesh, similar to the E3SM V1 land output.

[NML]
[BFB]


vanroekel:
@proteanplanet just so I understand, this PR is adding daily output, which is off now (6 total fields) and you are adding a handful of fields to the monthly output?  It is not clear from your description.  If I'm on target, could you modify the description to highlight what you are changing from v1 output?  Most of the monthly stream seems unchanged.

rljacob:
Please edit the PR description.  That is supposed to be copied in to the git commit message but its far to long.  Just describe the code changes.  You can move all the details to comment.  Also it looks like some words or numbers are missing.

proteanplanet:
> @proteanplanet just so I understand, this PR is adding daily output, which is off now (6 total fields) and you are adding a handful of fields to the monthly output? It is not clear from your description. If I'm on target, could you modify the description to highlight what you are changing from v1 output? Most of the monthly stream seems unchanged.

@vanroekel: We are adding: 1) Six new daily 2D Fields that provide basic state information (ice and snowmass and movement); 2) Four new monthly 2D fields associated with the new snow morphology; 3) Three new monthly 3D fields providing the full thickness and snow distribution PDF; 4) Switching on eight basic 2D monthly fields that were in the V1 stream files but were curiously turned off in the namelist, including sea ice velocity. 

Note that this is toned-down from my email request in recent days, as I missed an important factor in my back-of-envelope calcualtions for file size (apologies).  I believe this set is now at the minimum required level.

beharrop (approved):
Sounds good to me

njeffery (approved):
Looks good.  Verified that the additional snow fields are correct.

proteanplanet:
This has been updated to include monthly albedo per surface type, and melt pond fraction to accompany melt pond albedo, and is ready for merging.

vanroekel (approved):
approved by visual inspection

rljacob:
@vanroekel Still ok to merge now after the freeze?

vanroekel:
yes please proceed.

jonbob:
merged to next

golaz:

jonbob:

PR #3860, opened 2020-09-28
Allow up to 10 atm history tapes

beharrop:
This PR changes the maximum number of history files the atmosphere can output from six to ten.  Most of the code was already setup, but namelist variables needed to be created or updated to make use of the higher number of tapes.

[BFB]

### DIFF
--- a/components/cam/bld/namelist_files/namelist_definition.xml
+++ b/components/cam/bld/namelist_files/namelist_definition.xml
@@ -1284,7 +1284,7 @@ Default: none
 
 <!-- History and Initial Conditions Output -->
 
-<entry id="avgflag_pertape" type="char*1(6)"  category="history"
+<entry id="avgflag_pertape" type="char*1(10)"  category="history"
        group="cam_history_nl" valid_values="A,B,I,X,M" >
 Sets the averaging flag for all variables on a particular history file
 series. Valid values are:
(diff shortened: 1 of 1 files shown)


tangq:
@beharrop , since we will have the RRM runs, we may output more regional tapes in the future. I would suggest increasing the number limit to a larger number (20?).

beharrop:
The ten limit is the max I can do without going into cam_history.F90 and making more modifications.  Since this PR is needed to accommodate the output requests for the v2 production runs, I wanted to do only what was needed to move forward with that issue.  Would it be OK if we waited on increasing the number of tapes more until later?

tangq:
I thought changing to 20 was a minor modification and thus my suggestion. We can definitely leave it for a future task if it requires more changes.

wlin7 (approved):
The changes are good. In the current codes (cam_history_support.f90), the max # of tapes is 12. But that doesn't mean we need to define up to that number.

wlin7:
> @beharrop , since we will have the RRM runs, we may output more regional tapes in the future. I would suggest increasing the number limit to a larger number (20?).

@tangq , for regional output, each history file can output for multiple regions, as long as saving the same variables at the same frequency. So when limited by max of tape but still need to save multiple regions, this is an option to use.

tangq:
That's good to know. Thanks, @wlin7

golaz:
@rljacob, we also need this PR for v2 Water Cycle.

golaz (approved):

PR #3861, opened 2020-09-29
update mpas-source: Modify ocean freezing temperature to match sea ice

mark-petersen:
Previously, ocean freezing formulation did not match the sea ice.

See https://github.com/MPAS-Dev/MPAS-Model/pull/528 for details.

[NML]
[non-BFB]


xylar:
@mark-petersen, how do we make the associated namelist changes for E3SM?

mark-petersen:
@jonbob helps out with that, as part of the testing and review process.

proteanplanet:
> @mark-petersen, how do we make the associated namelist changes for E3SM?

@xylar: There is a script for doing that.  I included the necessary changes manually in my test B-case, but @jonbob tells me they aren't needed anymore.

proteanplanet (approved):
Approved upon inspection.

xylar (approved):
Okay, I'm approving based on the assumption that the E3SM namelist options for `eos` get updated to match the MPAS-Ocean defaults.

jonbob:
@xylar and @proteanplanet - if you have a chance, could you please look at the defaults from the autogenerating scripts? Thanks

xylar:
@jonbob, thanks for checking.  I can't comment on the line directly because it's out of range of the diff but on:
https://github.com/E3SM-Project/E3SM/blob/7823ee25390b369c4e85e0596080f78da9ba3643/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml#L316
the new default should be `0`:
```
<config_open_ocean_freezing_temperature_coeff_0>0.0</config_open_ocean_freezing_temperature_coeff_0>
```

xylar:
All the other changes look good.

jonbob:
@xylar - thanks for catching that. I don't think it's correct right now and I'll fix it. The automated scripts mostly point out differences between what's in Registry as the default and the old default in the coupled system. Unfortunately there are dozens of differences and unless understand that some old default has changed, just keep what has been in the system. I'll fix it now and get you to check again? And tomorrow is fine -- this one won't go in today

components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
```diff
@@ -313,16 +313,15 @@
 
 <!-- eos -->
 <config_eos_type>'jm'</config_eos_type>
-<config_open_ocean_freezing_temperature_coeff_0>-1.8</config_open_ocean_freezing_temperature_coeff_0>
+<config_open_ocean_freezing_temperature_coeff_0>0.0</config_open_ocean_freezing_temperature_coeff_0>
```
xylar, line 316: Yep, that's it.  Thanks!

xylar (approved):
Re-approving based on E3SM namelist additions

proteanplanet:
I have checked each of bld/build-namelist, bld/build-namelist-section, bld/namelist_files/namelist_defaults_mpaso.xml, and bld/namelist_files/namelist_definition_mpaso.xml with vimdiff against the scripts I created for the B-case test on Anvil, and all area in order.

mark-petersen:
@jonbob due to directory renaming, there were odd directory conflicts when I tried to cherry-pick. So I made two fresh commits, but with an old-fashioned unix copy of just those files from what we had before. Using `git diff --stat` it shows the same number of lines changed in the namelist scripts as before, and visual inspection also looks correct. I'm running a sanity test now, but this is ready to merge.

jonbob (approved):
Looks good from a visual inspection. I'll also run some sanity tests during the merge

jonbob:
merged to next

golaz:
@rljacob : this is needed to finalize v2 WaterCycle and should go on master

jonbob:

PR #3863, opened 2020-09-30
Deallocate local vector that is not deallocated automatically

### LINKED ISSUE #3854: allocated array returned to pointer variable not deallocated
In  subroutine atm_domain_mct( lsize, gsMap_a, dom_a ) in atm_comp_mct.F90:

         integer , pointer  :: idata(:)    ! temporary
         ...
         call mct_gsMap_orderedPoints(gsMap_a, iam, idata)

where mct_gsMap_orderedPoints (subroutine OrderedPoints_(GSMap, PEno, Points) in m_GlobalSegMap.F90)

     ! !DESCRIPTION:                                                                                                                                                         
     ! The function {\tt OrderedPoints\_()} takes the input {\tt GlobalSegMap}                                                                                               
     ! arguement {\tt GSMap} and returns a vector of the points owned by                                                                                                     
     ! {\tt PEno}.  {\tt Points} is allocated here.  The calling process                                                                                                     
     ! is responsible for deallocating the space.                       

And idata is not explicitly deallocated in  atm_domain_mct. Based on a google searches, it appears that

     "If you have pointers, you need to take care of deallocating the memory yourself"

that is, there is no automatic deallocation for allocated arrays pointed to by a local pointer variable.

Is this the correct interpretation? Should this be explicitly deallocated?

If so, it is a simple fix, and I can take care of it, but I am not confidant in my understanding.

worleyph:
A vector is allocated inside of a call to the routine mct_gsMap_orderedPoints
and pointed to by a local pointer variable in the routine atm_domain_mct.
This allocated storage is not automatically deallocated when returning
from atm_domain_mct. This is corrected here with an explicit deallocate.

Fixes #3854 

[BFB]

### DIFF
--- a/components/eam/src/cpl/atm_comp_mct.F90
+++ b/components/eam/src/cpl/atm_comp_mct.F90
@@ -733,10 +733,12 @@ CONTAINS
     !
     allocate(data(lsize))
     !
-    ! Initialize attribute vector with special value
+    ! Initialize attribute vector with special value,
+    ! then deallocate storage pointed to by idata
     !
     call mct_gsMap_orderedPoints(gsMap_a, iam, idata)
     call mct_gGrid_importIAttr(dom_a,'GlobGridNum',idata,lsize)
+    if (associated(idata)) deallocate(idata)
     !
     ! Determine domain (numbering scheme is: West to East and South to North to South pole)
     ! Initialize attribute vector with special value


worleyph:
Passed a sanity check on compy. In an alternative implementation, I put a timer around the deallocate (inside of the if-test), verifying that the deallocate did take place.

Note that other "motifs" used in EAM add a nullify after the deallocate of a pointer. Since this is a local pointer variable, the nullify is not required, and it allowed this to be a one-executable-line PR.

rljacob (approved):

rljacob:
This can proceed to master during v2 freeze.  No impact on WC.

wlin7:
Merged to next.

wlin7:
Merged to master.

PR #3865, opened 2020-09-30
Increase field widths in atm_chunk_cost.txt

### LINKED ISSUE #3838: fields  in atm_chunk_cost.txt not wide enough for some cases

worleyph:
The field width for the chunk id in the file atm_chunk_cost.txt is too
small when the number of chunks is 500,000 or larger, resulting in the
output of "*****". The field width for the process id in the file
is also too small when the number of processes is 1,000,001 or larger.

Here both of these conditions are checked for. If either is true, then
the field widths for chunk id and process id are increased enough
to support chunk and process ids as large as 999,999,999 .

Fixes #3838 

[BFB]

### DIFF
--- a/components/eam/src/physics/cam/phys_grid.F90
+++ b/components/eam/src/physics/cam/phys_grid.F90
@@ -2539,6 +2539,10 @@ logical function phys_grid_initialized ()
                                       !  load balance, but not traditional
                                       !  speed-up (relative to serial cost)
 
+   logical :: wide_format             ! flag indicating that should use wide
+                                      !  format, supporting larger cid, 
+                                      !  lcid, and procid
+
    character(len=*), parameter :: fname = 'atm_chunk_costs.txt'
 
 !-----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


worleyph:
This PR is identical to PR #3839 except based on current master, witn CAM => EAM. Closing #3839 . Ran a sanity check on Compy, building and running a case and comparing atm_chunk_costs.txt with a run using unmodified master. This did not exercise the "wide format" capability, but that was tested in the original PR and the changes are identical here.

rljacob:
This can proceed to master post-freeze.  No impact on WC.

sarats (approved):

PR #3867, opened 2020-09-30
GNU compiler change after Tulip OS upgrade

amametjanov:
GNU compiler change after Tulip OS upgrade

[BFB]


rljacob:
This does not effect WC and can be merged during v2 freeze

rljacob (approved):

PR #3868, opened 2020-09-30
Allows crops of to be put on separate land units

bishtgautam:
The default model configuration when the prognostic crop model is off is for the
crops to share the landunit with natural vegetation. In this default configuration, the
irrigated water for the crops is also used by natural vegetation. These code modifications
allow for crops and natural vegetation to be on separate landunits when the prognostic
crop model is off.

[BFB]

### DIFF
--- a/components/elm/bld/CLMBuildNamelist.pm
+++ b/components/elm/bld/CLMBuildNamelist.pm
@@ -1546,9 +1546,6 @@ sub setup_cmdl_irrigation {
                   "both irrigation and crop can NOT be on.\n");
     }
   } else {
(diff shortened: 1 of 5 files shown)


hydrotian:
@golaz @vanroekel This PR is to fix the irrigation landunit flag issue for the V2 water cycle simulations.

vanroekel:
@hydrotian can you verify you have tested this PR and it fixes the issue you have been having?

hydrotian:
Yes I verify that this PR fixes the issue we had.

vanroekel:
great, then I approve integrating this for WC v2

golaz:

bbye (approved):

hydrotian (approved):

PR #3870, opened 2020-10-01
Add the 'case.submit' step to the Travis test

lukaszlacinski:
This PR is to check if it is possible to not only build a case, but actually run it on a Travis CI VM which has limited resources (7.5 GB RAM), https://docs.travis-ci.com/user/reference/overview/.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,4 +47,4 @@ before_install:
 
 script:
     # Build a case
-    - travis_wait 30 singularity exec --hostname singularity e3sm.sif .travis/run.sh
+    - travis_wait 60 singularity exec --hostname singularity e3sm.sif .travis/run.sh
--- a/.travis/run.sh
+++ b/.travis/run.sh
@@ -62,6 +62,7 @@ if [ $rc -ne 0 ]; then
     exit $rc
 fi
 
+command_yellow "./xmlchange STOP_N=1"
 # Unfortunately, case.submit always return 0, even if it fails.
 # To find out if it succeeded, we check if the run log has been gzipped.
 command_yellow "./case.submit"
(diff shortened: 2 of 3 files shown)


rljacob:
This does not effect WC and can be merged during v2 freeze.

lukaszlacinski:
Adding the 'case.submit' step to the Travis test revealed a bug in https://github.com/ESMCI/cime/blob/master/scripts/lib/CIME/case/check_input_data.py. https://github.com/ESMCI/cime/pull/3730 that was merged today fixes the bug, but as long as the E3SM repo does not points to the cime repo with with the fix merged, the Travis test will fail.

rljacob:
You mean it will fail on this branch?

lukaszlacinski:
Yes. This E3SM master branch points to cime@54ef8b which is two days old, and the bug fix was merged to the cime master branch today, so the case.submit step this PR adds to the Travis test will fail.

rljacob:
@lukaszlacinski E3SM cime submodule now has the changes this PR needs.  I tried to rerun but looks like there is still a fail.

rljacob:
Looks like its running out of time.  We can try reducing the run length.  Before the ./case.submit command, issue this command:  ./xmlchange STOP_N=1
That will change it to run for 1 day instead of 5.

rljacob:
Still out of time.  Can you capture the output of ./case.submit and show that?  We'll be able to tell if it was still downloading data or actually started a run.

rljacob:
It looks like the default optimization level for gnu is -O which will make it slow.  Try adding this to the singularity entry in config_compilers
```
  <FFLAGS>
    <append DEBUG="FALSE"> -O2  </append>
  </FFLAGS>
```

rljacob:
Note from IG telecon:  this case takes to much memory and also file download time causes it to run out of time.  So can't run it.

PR #3872, opened 2020-10-02
Finish moving cam to eam

rljacob:
Finish moving cam to eam by moving 3 files from cam/src/physics/crm/samxx
that were missed.

[BFB]


rljacob:
@whannah1 The ERS_Ln9_P96x1.ne4pg2_ne4pg2.F-MMFXX test passed after the cam2eam rename even though I had missed 2 source files from samxx/test.  Are they not needed in that system test?

whannah1:
@rljacob these files are only for SAM++ standalone tests, so they aren't used for the full model tests and it's not surprising that they wouldn't affect the dashboard tests.

whannah1 (approved):

golaz:
@rljacob : you should merge this for v2 when it's ready.

PR #3876, opened 2020-10-05
Update mapping and domain files for EC30to60E2r2 grid

jonbob:
This PR updates the mapping and domain files for the ocn/ice EC30to60E2r2 grid. There was a bug found in the code that makes SCRIP files from an mpas mesh and this PR corrects any possible errors for this mesh by replacing the old mapping and domain files with those generated from a fixed SCRIP mesh file. A new date stamp has been added and all files are staged on the inputdata server.

There is one test in e3sm_prod that uses this grid and it should change answers, but otherwise this PR should not impact any other currently tested configurations.

[non-BFB] only for tests  using this resolution


golaz:
@rljacob : this should also go into v2

jonbob:
passes SMS.ne30pg2_r05_EC30to60E2r2-1900_ICG.A_WCYCL1850S.anvil_intel

rljacob (approved):

rljacob:
I prefer we use "non-BFB only for" instead of "BFB except for".   When we are looking for non-BFB commits in the log, we won't find the latter.

jonbob:
Thanks @rljacob - I changed the label to match

vanroekel (approved):
approved by visual inspection and B-case testing

jonbob:
merged to next

jonbob:
merged to master. There is an expected DIFF on anvil that so far shows up as a FAIL, though I believe that is due to machine issues and not this PR. The same test passed on compy and passes sometimes on anvil, but the error code from FAILs does not look code-related

PR #3880, opened 2020-10-07
V1 planar standalone homme

celdredsandia:
This adds doubly periodic planar capabilities to standalone HOMME. More details are found at https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/1877639335/Planar+HOMME.

should be [bfb], but not yet


components/homme/src/common_movie_mod.F90
```diff
@@ -133,30 +133,32 @@ module common_movie_mod
...
                                                                1,2,4,0,  &  !u
                                                                1,2,4,0,  &  !v
-                                                               1,0,0,0,  &  !lon
-                                                               1,0,0,0,  &  !lat
+                                                               1,0,0,0,  &  !lon (x for planar)
+                                                               1,0,0,0,  &  !lat (y for planar)
                                                                4,0,0,0,  &  ! time
                                                                1,2,4,0,  &  ! zeta
                                                                1,2,4,0,  &  ! div
+                                                               1,2,4,0,  &  ! eta
```
oksanaguba, line 150: what is eta in here?

components/homme/src/prim_movie_mod.F90
```diff
@@ -219,10 +224,18 @@ subroutine prim_movie_init(elem, par, hvcoord,tl)
     call nf_global_attribute(ncdf, 'np', np)
     call nf_global_attribute(ncdf, 'ne', ne)
 
+  if (geometry=="sphere") then
```
oksanaguba, line 227: not sure how slow string comparison can be, could this be replaced with boolean quantities, like sphgeometry = (geometry == "sphere")?

components/homme/src/common_movie_mod.F90
celdredsandia: eta is the absolute vorticity (sum of "planetary" and relative vorticity). I will add a comment to this effect. I am also open to changing the name, eta is just the one I am familiar with

components/homme/src/prim_movie_mod.F90
celdredsandia: it definitely could be! I am not sure if the performance matters much here though, since these comparisons happen only in initialization where I don't think the difference between boolean and string comparison should matter.

components/homme/test_execs/theta-l-nlev128-native/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+#_______________________________________________________________________
+# create executable needed by planar density current (native output)
```
oksanaguba, line 2: there is already theta exec with 128 levels, theta-nlev128 .

aslo, it may be too much work here but would it be good to get rid of -l- suffix in theta-l?

oksanaguba:
This PR adds grid files to homme. At least some of them are already in INPUTSERVER/atm/cam/inic/homme . To carry two copies of the same grid is prob not the best solution? I am not sure what should be done here since I can see someone wanting to run these grids from standalone homme. Should all atm. RRM grids be moved to homme?

also, at least this grid /test/mesh_files/mountain_10_x2.g is already in homme, there should be only 1 copy of it.

components/homme/src/share/derivative_mod_base.F90
```diff
@@ -11,7 +11,7 @@ module derivative_mod_base
   use parallel_mod,   only : abortmp
   use element_mod,    only : element_t
   use control_mod,    only : hypervis_scaling, hypervis_power
-  use physical_constants, only : rrearth
+  use physical_constants, only : scale_factor_inv, laplacian_rigid_factor
```
oksanaguba, line 14: Just curious -- why does laplacian have its rigid factor but not other derivatives? what is rigid factor?

oksanaguba: nm, see it below

components/homme/test/reg_test/run_tests/test-list.cmake
```diff
@@ -31,6 +31,13 @@ SET(HOMME_TESTS
   thetanh-TC-nutop.cmake
   thetanhwet-TC.cmake
   hommetool.cmake
+  thetah-nhgw.cmake
```
oksanaguba, line 34: yay to tests

components/homme/src/test_mod.F90
```diff
@@ -75,15 +78,27 @@ subroutine set_test_initial_conditions(elem, deriv, hybrid, hvcoord, tl, nets, n
     case('mtest2');
     case('mtest3');
     case('held_suarez0');
-    case('jw_baroclinic');
+    case('hgw');
```
oksanaguba, line 81: just being picky -- these tests names are very short.

oksanaguba:
Sorry for getting to it so late. 

I would suggest removing string comparison from init code and rename tests to something longer (if it does not involve much effort).

I found it hard to read because changes include a lot of formatting changes (imo formatting changes should not be here). 

Something needs to be done about RRM grids.

Meanwhile I can start testing it.

oksanaguba:
@celdredsandia do you use all these *.g rrm files in your tests? would it be enough to use just one, mountain, that is already in homme?

components/homme/src/prim_movie_mod.F90
mt5555: string comparisons should be avoided inside performance critical loops.  But at this level in the code, I think they are easier to read and  I vote to keep it.

mt5555:
I second the comment about the *.g files.  In general we try to avoid large binary files in the git respository, but we make an exception for those needed by the HOMME standalone tests so they can ran.   Could we only add .g files needed for the standalone tests?   And if possible, could those be smaller versions (not ne30 type grids.).   

I think they should also go in test/mesh_files.  (so we dont have to duplicate mountain_10_x2.g)

components/homme/test_execs/theta-l-nlev128-native/CMakeLists.txt
celdredsandia: I introduced this because the planar code doesn't yet support interpolated (non-native) grid output, so I need theta execs that use native output.

I agree that we should change theta-l to theta at some point, but the name theta-l is prevalent throughout the code, so this should probably be a separate pull request.

components/homme/src/test_mod.F90
celdredsandia: I will make these names longer and more descriptive.

celdredsandia:
I have made the test names more descriptive, removed the unneeded .g files and added a comment about what eta (absolute vorticity) is. I think this pull request should be ready to go now in terms of testing and integration!

components/homme/test_execs/CMakeLists.txt
```diff
@@ -161,20 +161,25 @@ ENDIF ()
...
   ADD_SUBDIRECTORY(swtcB)
+  ADD_SUBDIRECTORY(swtcC)
 ENDIF()
 
 
 # Add the test exec subdirs for the preqx executable
 IF(${BUILD_HOMME_PREQX})
   ADD_SUBDIRECTORY(baroC)
   ADD_SUBDIRECTORY(baroCam)
+  ADD_SUBDIRECTORY(preqx-nlev10-native)
```
oksanaguba, line 172: This is a minor comment but with these new execs time building homme is doubled. should these execs be wrapped in some IF_PLANAR_TESTS variable?

celdredsandia: So right now there are the planar regression tests, which only need a few of the execs; and the full planar test suite (similar to a DCMIP-type suite of tests) that needs all of them. Only the regression tests should be run to check BFB pull requests; the full suite is useful for analyzing non-BFB changes and generating results for comparison with other models.

I can rework how the executables are built for the planar test suite so we can comment out most of them, if the build time is an issue.

oksanaguba: thanks -- probably, let's just leave it as is for now.

celdredsandia: Sounds good! I will make sure and put these changes in V2

oksanaguba:
```
The following tests FAILED:
          1 - verifyBaselineResults (Failed)
          4 - swtc5 (Failed)
          6 - baro2b (Failed)
          7 - baro2c (Failed)
          8 - baro2d (Failed)
          9 - baroCamMoist (Failed)
         10 - baroCamMoist-acc (Failed)
         13 - preqx-TC-ftype4 (Failed)
         14 - thetah-TC-ftype4 (Failed)
         15 - thetah-TC (Failed)
         16 - thetanh-TC (Failed)
         17 - thetanh-TC-nudiv (Failed)
         18 - thetanh-TC-nutop (Failed)
         19 - thetanhwet-TC (Failed)
         20 - hommetool (Failed)
         21 - thetah-nhgw (Failed)
         22 - preqx-nhgw (Failed)
         23 - thetanh-nhgw (Failed)
         24 - thetah-nhgw-slice (Failed)
         25 - preqx-nhgw-slice (Failed)
         26 - thetanh-nhgw-slice (Failed)
         27 - sweqx-dbvor (Failed)
         30 - thetah-sl-dcmip16_test1pg2 (Failed)
```

oksanaguba:
At least swtc5 and baro2b ran but failed diffs. maybe, these lines
```
old 
                  vartmp(:,:,k)=cvlist(ie)%vert_latlon(k,:,:)%lon*180/dd_pi
new 
                   vartmp(:,:,k)=cvlist(ie)%vert_latlon(k,:,:)%lon*latlon_adj_factor
```

oksanaguba:
In src/theta-l_kokkos/CMakeLists.txt 
there is a block
```
  SET(TEST_SRC_F90
    ${TEST_SRC_DIR}/asp_tests.F90
    ${TEST_SRC_DIR}/baroclinic_inst_mod.F90
    ${TEST_SRC_DIR}/dcmip12_wrapper.F90
    ${TEST_SRC_DIR}/dcmip16_wrapper.F90
    ${TEST_SRC_DIR}/dcmip2012_test1_conv.F90
    ${TEST_SRC_DIR}/dcmip2012_test1_2_3.F90
    ${TEST_SRC_DIR}/dcmip2012_test4.F90
    ${TEST_SRC_DIR}/dcmip2016-baroclinic.F90
    ${TEST_SRC_DIR}/dcmip2016-kessler.F90
    ${TEST_SRC_DIR}/dcmip2016-physics-z.F90
    ${TEST_SRC_DIR}/dcmip2016-supercell.F90
    ${TEST_SRC_DIR}/dcmip2016-terminator.F90
    ${TEST_SRC_DIR}/dcmip2016-tropical-cyclone.F90
    ${TEST_SRC_DIR}/held_suarez_mod.F90
    ${TEST_SRC_DIR}/mtests.F90
  )
```
you need to add your test files (moist_planar.F90, etc.) there. I don't know why it is different for the F executable, which looks for all files in test_src.

oksanaguba:
Also, that missing line 'jw_baroclinic', while it does nothing, probably stilll needed cause otherwise 
```
    case default;               call abortmp('unrecognized test case')
```

oksanaguba:
Seems the comment about src/theta-l_kokkos/CMakeLists.txt hasn't been addressed yet. Thanks!

celdredsandia:
The SHA this branch is based on is before theta-l_kokkos was merged, so that file doesn't exist. What is the best way to proceed in that case?

rljacob:
Replaced by #3987

PR #3884, opened 2020-10-09
Fixes support for user-defined ELM grid

bishtgautam:
This code modification restores the support for user-defined land grid that
was broken when the name of the land model was changed. A new test is
added for `--res ELM_USRDAT`.

[BFB]

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -37,6 +37,7 @@ _TESTS = {
             "ERS.r05_r05.RMOSGPCC.mosart-gpcc_1972",
             "ERS.MOS_USRDAT.RMOSGPCC.mosart-mos_usrdat",
             "SMS.MOS_USRDAT.RMOSGPCC.mosart-unstructure",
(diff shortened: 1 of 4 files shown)


rljacob (changes requested):
Replace CLM_USRDAT with ELM_USRDAT everywhere.

bishtgautam:
@rljacob You wanted me to rename local variables as in https://github.com/E3SM-Project/E3SM/pull/3884/commits/615ba48a48cd30029b0aa6e95ed8483aa386e56f? Or, do you want me to rename namelist variables such that `case.get_value("CLM_CO2_TYPE")` changes to `case.get_value("ELM_CO2_TYPE")`?

rljacob (approved):

rljacob:
Nevermind.  I thought I saw some more instances of CLM_USRDAT in elm but I think you got them al.

rgknox:
Thanks @bishtgautam , these grids are very useful for FATES runs

PR #3885, opened 2020-10-09
Fixes a bug in dry size calculations and these calculations now use current state

singhbalwinder:
Dry aerosol size calculations were using an incorrect log of 
sigma g (geometric std. dev. of aerosol mode) during
accumulation<->aitken transfer. This Pr fixes that bug.

This PR is NBFB as dry aerosol sizes and water uptake are 
now using the current "state" (mass mixing ratios of the 
species). Earlier, an older "state" was used to compute
these quantities. I have modified RRTMGP and CRM codes
to accommodate this change.

An extensive refactor of modal_aero_calcsize is not included
in this PR so that it is easier to understand where the NBFB
differences are coming from. Code in this branch is now
BFB with the code where modal_aero_calcsize is refactored
extensively. I will port those changes as a separate PR which
will be BFB.

[NBFB]

### DIFF
--- a/components/eam/src/chemistry/utils/modal_aero_calcsize.F90
+++ b/components/eam/src/chemistry/utils/modal_aero_calcsize.F90
@@ -510,7 +510,6 @@ subroutine modal_aero_calcsize_sub(state, deltat, pbuf, ptend, do_adjust_in, &
 
    integer  :: lchnk                ! chunk identifier
    integer  :: ncol                 ! number of columns
-   integer  :: list_idx_local
 
    real(r8), pointer :: t(:,:)      ! Temperature in Kelvin
    real(r8), pointer :: pmid(:,:)   ! pressure at model levels (Pa)
@@ -609,10 +608,9 @@ subroutine modal_aero_calcsize_sub(state, deltat, pbuf, ptend, do_adjust_in, &
       do_aitacc_transfer = do_aitacc_transfer_default
    end if
 
-   list_idx_local = 0
    if(present(list_idx_in)) then
-      if(.not. present(dgnumdry_m)) call endrun('list_idx_in is present but dgnumdry_m pointer is missing'//errmsg(__FILE__,__LINE__))
-      list_idx_local = list_idx_in
+      if(.not. present(dgnumdry_m)) call endrun('list_idx_in is present but dgnumdry_m' // &
+      ' pointer is missing'//errmsg(__FILE__,__LINE__))
    endif
 
    lchnk = state%lchnk
@@ -1481,7 +1479,7 @@ subroutine modal_aero_calcsize_diag(state, pbuf, list_idx_in, dgnum_m)
 
    call rad_cnst_get_info(list_idx, nmodes=nmodes)
 
-   if (list_idx /= 0) then
+   if (present(list_idx_in)) then
       if (.not. present(dgnum_m)) then
          call endrun('modal_aero_calcsize_diag called for'// &
                      'diagnostic list but dgnum_m pointer not present')
@@ -1494,7 +1492,7 @@ subroutine modal_aero_calcsize_diag(state, pbuf, list_idx_in, dgnum_m)
 
    do n = 1, nmodes
 
-      if (list_idx == 0) then
+      if (.not.present(dgnum_m)) then
          call pbuf_get_field(pbuf, dgnum_idx, dgncur_a, start=(/1,1,n/), kount=(/pcols,pver,1/))
       else
          dgncur_a => dgnum_m(:,:,n)
--- a/components/eam/src/chemistry/utils/modal_aero_calcsize.F90
+++ b/components/eam/src/chemistry/utils/modal_aero_calcsize.F90
@@ -470,8 +470,10 @@ end subroutine modal_aero_calcsize_init
 
 !===============================================================================
 
-subroutine modal_aero_calcsize_sub(state, ptend, deltat, pbuf, do_adjust_in, &
-   do_aitacc_transfer_in)
+subroutine modal_aero_calcsize_sub(state, deltat, pbuf, ptend, do_adjust_in, &
+   do_aitacc_transfer_in, list_idx_in, dgnumdry_m)
+
+  use shr_log_mod ,  only: errMsg => shr_log_errMsg
 
    !-----------------------------------------------------------------------
    !
@@ -490,12 +492,14 @@ subroutine modal_aero_calcsize_sub(state, ptend, deltat, pbuf, do_adjust_in, &
 
    ! arguments
    type(physics_state), target, intent(in)    :: state       ! Physics state variables
-   type(physics_ptend), target, intent(inout) :: ptend       ! indivdual parameterization tendencies
    real(r8),                    intent(in)    :: deltat      ! model time-step size (s)
    type(physics_buffer_desc),   pointer       :: pbuf(:)     ! physics buffer
 
-   logical, optional :: do_adjust_in
-   logical, optional :: do_aitacc_transfer_in
+   type(physics_ptend), optional, target, intent(inout) :: ptend       ! indivdual parameterization tendencies
+   logical,  optional :: do_adjust_in
+   logical,  optional :: do_aitacc_transfer_in
+   integer,  optional :: list_idx_in
+   real(r8), optional, pointer :: dgnumdry_m(:,:,:) ! interstital aerosol dry number mode radius (m)
 
 #ifdef MODAL_AERO
 
@@ -506,6 +510,7 @@ subroutine modal_aero_calcsize_sub(state, ptend, deltat, pbuf, do_adjust_in, &
 
    integer  :: lchnk                ! chunk identifier
    integer  :: ncol                 ! number of columns
+   integer  :: list_idx_local
 
    real(r8), pointer :: t(:,:)      ! Temperature in Kelvin
    real(r8), pointer :: pmid(:,:)   ! pressure at model levels (Pa)
@@ -528,6 +533,8 @@ subroutine modal_aero_calcsize_sub(state, ptend, deltat, pbuf, do_adjust_in, &
 
    logical  :: dotendqqcw(pcnst)
    logical  :: noxf_acc2ait(ntot_aspectype)
+   logical  :: ptend_present
+
 
    character(len=fieldname_len)   :: tmpnamea, tmpnameb
    character(len=fieldname_len+3) :: fieldname
@@ -602,6 +609,12 @@ subroutine modal_aero_calcsize_sub(state, ptend, deltat, pbuf, do_adjust_in, &
       do_aitacc_transfer = do_aitacc_transfer_default
    end if
 
+   list_idx_local = 0
+   if(present(list_idx_in)) then
+      if(.not. present(dgnumdry_m)) call endrun('list_idx_in is present but dgnumdry_m pointer is missing'//errmsg(__FILE__,__LINE__))
+      list_idx_local = list_idx_in
+   endif
+
    lchnk = state%lchnk
    ncol  = state%ncol
 
@@ -609,15 +622,21 @@ subroutine modal_aero_calcsize_sub(state, ptend, deltat, pbuf, do_adjust_in, &
    pmid => state%pmid
    pdel => state%pdel
    q    => state%q
-      
-   dotend => ptend%lq
-   dqdt   => ptend%q
 
-   call pbuf_get_field(pbuf, dgnum_idx, dgncur_a)
-
-   dotendqqcw(:) = .false.
-   dqqcwdt(:,:,:) = 0.0_r8
-   qsrflx(:,:,:,:) = 0.0_r8
+   if(present(ptend)) then
+      ptend_present = .true.
+      dotend => ptend%lq
+      dqdt   => ptend%q
+      call pbuf_get_field(pbuf, dgnum_idx, dgncur_a)
+      dotendqqcw(:) = .false.
+      dqqcwdt(:,:,:) = 0.0_r8
+      qsrflx(:,:,:,:) = 0.0_r8
+   else
+      dgncur_a => dgnumdry_m
+      ptend_present = .false.
+      dqqcwdt(:,:,:)  = huge(dqqcwdt)
+      qsrflx(:,:,:,:) = huge(qsrflx)
+   endif
 
    nait = modeptr_aitken
    nacc = modeptr_accum
@@ -626,7 +645,7 @@ subroutine modal_aero_calcsize_sub(state, ptend, deltat, pbuf, do_adjust_in, &
    ! tadj = adjustment time scale for number, surface when they are prognosed
    !           currently set to deltat
    tadj = deltat
-   tadj = 86400
+   tadj = 86400.0_r8
    tadj = max( tadj, deltat )
    tadjinv = 1.0_r8/(tadj*(1.0_r8 + 1.0e-15_r8))
    fracadj = deltat*tadjinv
@@ -693,7 +712,7 @@ subroutine modal_aero_calcsize_sub(state, ptend, deltat, pbuf, do_adjust_in, &
          !    compute number tendencies that will bring numbers to their
          !    current diagnosed values
          !
-         if (lna > 0) then
+         if (lna > 0 .and. ptend_present) then
             dotend(lna) = .true.
             do k=top_lev,pver
                do i=1,ncol
@@ -702,7 +721,7 @@ subroutine modal_aero_calcsize_sub(state, ptend, deltat, pbuf, do_adjust_in, &
                end do
             end do
          end if
-         if (lnc > 0) then
+         if (lnc > 0 .and. ptend_present) then
             dotendqqcw(lnc) = .true.
             do k=top_lev,pver
                do i=1,ncol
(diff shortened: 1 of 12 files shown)


singhbalwinder:
I have run RRTMGP and CRM tests from the integration test suite and they work fine with these changes.

I will revisit these codes in a separate PR to clean up some of the logic. I hope to remove calsize_diag and have just calcsize_sub routine to work as calcsize_diag for prognostic aerosol and diagnostic calculations.

singhbalwinder:
Here are the diagnostics showing impact of this bug fix and having the calcsize use the current state:
https://compy-dtn.pnl.gov/sing201/diags/BugfixedCalcsize_vs_Master/viewer/

components/eam/src/physics/cam/modal_aer_opt.F90
```diff
@@ -1265,17 +1301,30 @@ subroutine modal_aero_lw(list_idx, state, pbuf, tauxar)
...
+            call modal_aero_wateruptake_dr(state, pbuf, list_idx, dgnumdry_m, dgnumwet_m, &
+              qaerwat_m, wetdens_m, clear_rh_in=clear_rh)
+         else
+            call modal_aero_wateruptake_dr(state, pbuf, list_idx, dgnumdry_m, dgnumwet_m, &
+              qaerwat_m, wetdens_m)
+         endif
+      else
+         !For prognostic aerosols
+         call modal_aero_calcsize_sub(state, dt, pbuf,list_idx_in=list_idx, dgnumdry_m=dgnumdry_m)
+         call modal_aero_wateruptake_dr(state, pbuf, list_idx, dgnumdry_m, dgnumwet_m, &
```
whannah1, line 1320: Don't we need logic for when clear_rh is present here?

whannah1 (approved):
This looks good to me as it is, but It's a shame that we have to duplicate so many lines for the clear_rh to be passed into the radiation. I vaguely recall being able to just pass an optional argument down and the state of being "present" or not travels with that argument, so we might not need all these new "if present" blocks.

components/eam/src/chemistry/utils/modal_aero_calcsize.F90
```diff
@@ -626,7 +643,7 @@ subroutine modal_aero_calcsize_sub(state, ptend, deltat, pbuf, do_adjust_in, &
    ! tadj = adjustment time scale for number, surface when they are prognosed
    !           currently set to deltat
    tadj = deltat
-   tadj = 86400
+   tadj = 86400.0_r8
```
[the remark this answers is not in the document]
singhbalwinder: It probably doesn't matter here but sometime, it can make a difference. It also makes it more explicit that `tadj` is a double precision real number.

components/eam/src/physics/cam/modal_aer_opt.F90
singhbalwinder: I will clean this up in my next BFB PR. Since MMF was not using prognostic aerosols, I didn't include this logic for `clear_rh`

kaizhangpnl:
@singhbalwinder Thanks again for working on this. Does this PR also resolve the memory problem we had before?

components/eam/src/physics/cam/modal_aer_opt.F90
```diff
@@ -371,20 +401,23 @@ end subroutine modal_aer_opt_init
 
 !===============================================================================
 
-subroutine modal_aero_sw(list_idx, state, pbuf, nnite, idxnite, is_cmip6_volc, ext_cmip6_sw, trop_level,  &
-                         tauxar, wa, ga, fa)
+subroutine modal_aero_sw(list_idx, dt, state, pbuf, nnite, idxnite, is_cmip6_volc, ext_cmip6_sw, trop_level,  &
+                         tauxar, wa, ga, fa, clear_rh)
```
kaizhangpnl, line 405: Would it be better to use "clear_rh_mmf" here, so that we know this is for MMF?

components/eam/src/chemistry/utils/modal_aero_calcsize.F90
```diff
@@ -1149,6 +1170,7 @@ subroutine modal_aero_calcsize_sub(state, ptend, deltat, pbuf, do_adjust_in, &
                      drv_c = drv_c_acc
                   end if
 
+                  dumfac = exp(4.5_r8*alnsg_amode(n)**2)*pi/6.0_r8
```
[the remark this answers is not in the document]
kaizhangpnl: Will your comment here be permanent? If not, it would be better to have the bugfix information (date and reason) in the code.

components/eam/src/physics/cam/modal_aer_opt.F90
singhbalwinder: I am revisiting this logic in a follow-up PR. I will change the name of this variable and add a comment about it in that PR.

singhbalwinder:
@kaizhangpnl : My comment about bug fix is permanent but only on GitHub. I will add a comment about this in the code in a follow up PR. Thanks for your review.

kaizhangpnl (approved):

singhbalwinder:
> @singhbalwinder Thanks again for working on this. Does this PR also resolve the memory problem we had before?

Sorry, I missed answering this question. The memory issue is more apparent when we invoke diagnostics (which are disabled here). This PR partly addresses that issue by allocating arrays in the init routine rather than at each time step.

wlin7:
Hi @singhbalwinder , should we wait for approval from all the suggested reviewers for this PR?

hwangacme (approved):

components/eam/src/physics/cam/modal_aer_opt.F90
brhillman: Regardless the answer is "no". The "presence" of an optional parameter like `clear_rh` would be passed to a subroutine. So actually I think this whole block can be cleaned up by just passing `clear_rh_in=clear_h` in both cases. Nothing wrong with being explicit though probably, so long as the clear_rh present case gets fixed.

components/eam/src/physics/rrtmgp/cam_optics.F90
```diff
@@ -503,9 +507,15 @@ subroutine set_aerosol_optics_sw(icall, state, pbuf, &
       tau_w = 0._r8
       tau_w_g = 0._r8
       tau_w_f = 0._r8
-      call aer_rad_props_sw(icall, state, pbuf, &
-                            count(night_indices > 0), night_indices, is_cmip6_volc, &
-                            tau, tau_w, tau_w_g, tau_w_f)
+      if(present(clear_rh)) then
```
brhillman, line 510: I believe these logical blocks are unnecessary, you should just be able to pass clear_rh and avoid the duplicated code.

components/eam/src/physics/cam/aer_rad_props.F90
```diff
@@ -262,8 +265,13 @@ subroutine aer_rad_props_sw(list_idx, state, pbuf,  nnite, idxnite, is_cmip6_vol
 
    ! Contributions from modal aerosols.
    if (nmodes > 0) then
-      call modal_aero_sw(list_idx, state, pbuf, nnite, idxnite, is_cmip6_volc, ext_cmip6_sw(:,:,idx_sw_diag), trop_level, &
-                         tau, tau_w, tau_w_g, tau_w_f)
+      if(present(clear_rh)) then
```
brhillman, line 268: I think this extra logic is unnecessary; you can just pass clear_rh as an optional and `model_aero_sw` will determine if it is a valid argument.

components/eam/src/physics/cam/aer_rad_props.F90
```diff
@@ -418,7 +429,11 @@ subroutine aer_rad_props_lw(is_cmip6_volc, list_idx, state, pbuf,  odap_aer)
 
    ! Contributions from modal aerosols.
    if (nmodes > 0) then
-      call modal_aero_lw(list_idx, state, pbuf, odap_aer)
+      if (present(clear_rh)) then
```
brhillman, line 432: Again should just be able to pass clear_rh.

components/eam/src/physics/cam/modal_aer_opt.F90
```diff
@@ -585,17 +613,32 @@ subroutine modal_aero_sw(list_idx, state, pbuf, nnite, idxnite, is_cmip6_volc, e
...
-      ! water uptake and wet radius for the climate list has already been calculated
-      call pbuf_get_field(pbuf, dgnumwet_idx, dgnumwet_m)
-      call pbuf_get_field(pbuf, qaerwat_idx,  qaerwat_m)
+
+      ! Calculate aerosol size distribution parameters and aerosol water uptake
+      ! For prescribed aerosol codes
+      if (clim_modal_aero .and. .not. prog_modal_aero) then
+         ! diagnostic aerosol size calculations
+         call modal_aero_calcsize_diag(state, pbuf, list_idx, dgnumdry_m)
+         if (present(clear_rh)) then
```
brhillman, line 622: Can just pass clear_rh to both.

components/eam/src/physics/cam/modal_aer_opt.F90
```diff
@@ -1265,17 +1301,30 @@ subroutine modal_aero_lw(list_idx, state, pbuf, tauxar)
...
    if (list_idx == 0) then
-      ! water uptake and wet radius for the climate list has already been calculated
-      call pbuf_get_field(pbuf, dgnumwet_idx, dgnumwet_m)
-      call pbuf_get_field(pbuf, qaerwat_idx,  qaerwat_m)
+      ! Calculate aerosol size distribution parameters and aerosol water uptake
+      ! For prescribed aerosol codes
+      if (clim_modal_aero .and. .not. prog_modal_aero) then
+         ! diagnostic aerosol size calculations
+         call modal_aero_calcsize_diag(state, pbuf, list_idx, dgnumdry_m)
+         if (present(clear_rh)) then
```
brhillman, line 1309: Again could just pass clear_rh in both calls.

brhillman (approved):
This looks fine to me. As @whannah1 said, the additional logic around calls that pass `clear_rh` are unnecessary; you should be able to pass that optional argument in all cases, and the called routine will check whether the optional argument is a valid parameter. That is, the behavior of "present" should follow from the top down. The advantage of the extra logic you've added is that it's explicit, but the disadvantage is that I think it hurts the readability and maintainability a little, because of the duplicated code blocks. That said, I'm fine if you want to leave it as is in this PR.

singhbalwinder:
Thanks @hwangacme and @brhillman for your reviews.

I agree that the "present" logic is a little messy. This part of the code is very tricky as there are multiple versions of radiation.F90 in the code (for RRTMG, RRTMGP, CAMRT, CAM etc.). This makes it really hard to predict which version will be called in a particular configuration (normal, RRTMGP, MMF etc.). I placed these "present" logic so that the code works in all scenarios. I will see if I can remove some of these in my followup PR. If the "present" logic is missing and the optional variable is accessed, the code generally crashes with a segmentation fault (without much useful info even in debug mode). Since `clear_rh` is only provided by MMF codes, I will see if I can remove the "present" logic in routines never called by MMF codes.

@wlin7 : Please go ahead and start merging this PR if repo is open. I will rebase my followup PR once this is merged and address all these issues in that PR.

brhillman:
@singhbalwinder I didn't mean to remove the present logic in the low level routines, I mean that you can remove the present logic around the subroutine calls, because you can pass an optional parameter clear_rh whether it evaluates present or not. Passing an optional that may or may not be there is different than accessing the variable directly, which I agree would create a segfault.

BTW, we have worked pretty hard to keep the call signatures as consistent as possible between RRTMGP, RRTMG, and the MMF version of RRTMGP. I'm not sure how I feel about them diverging. Walter's clear_rh addition may even be useful outside of the MMF context, I don't know. I haven't looked closely enough at it.

wlin7:
Merged to next.

wlin7:
Merged to master.

Request to bless 44 DIFF sandiatoss3 next tests, leave out pre-existing 2 DIFFs
Request to bless 40 DIFF compy integration_next tests. leave out 1 pre-existing DIFF; compy has 3 new FAIL, one of which was DIFF in the previous return.

 SMS_D_Ld1.ne30_oECv3_ICG.A_WCYCL1850S_CMIP6.compy_pgi.allactive-v1cmip6
 SMS_Ld9.ne4pg2_ne4pg2.FC5AV1C-04P2.compy_pgi.eam-thetahy_sl_pg2_mass
 SMS_D_Ld1.ne30_r05_oECv3.A_WCYCL1850.compy_pgi  (DIFF to FAIL)

To watch for test returns for other machines. behop all tests failed, there must be other systematic issue.

wlin7:
Bless 11 tests for mappy. Note that while tested using the same commit, 6 DIFFs reported on 10/15/2020, but 11 on 10/16/2020. DIFFs on 10/15/2020 were not blessed.

wlin7:
Request to bless 3 tests each for anvil and compy e3sm_prod_next_intel.

Also bless compy e3sm_integration_next_pgi on 2 tests that were  FAILs since the merge of this PR until 10/21/20. Both passed before this PR. Presumably the current DIFFs (as of 10/22/20) are due to this PR.

```
SMS_D_Ld1.ne30_oECv3_ICG.A_WCYCL1850S_CMIP6.compy_pgi.allactive-v1cmip6
SMS_Ld9.ne4pg2_ne4pg2.FC5AV1C-04P2.compy_pgi.eam-thetahy_sl_pg2_mass
```

Two other DIFFs in compy e3sm_integration_next_pgi existed well before this PR and remain so. Although the current PR contributes to DIFF, they are yet to be blessed. Listed below for reference

```
MS_D_Ld1.ne30_r05_oECv3.A_WCYCL1850.compy_pgi 
SMS_Ld1.ne4_ne4.F-EAMv1-AQP1.compy_pgi.eam-clubb_only


```

wlin7:
@rljacob , what do we need to do about the remaining two DIFFs (see comment immediately above) in compy e3sm_integration_next_pgi?

rljacob:
Go ahead and request a bless for them.  They were probably missed earlier.

wlin7:
Request submitted to bless the two remaining compy DIFFs as noted above.

components/eam/src/physics/cam/modal_aer_opt.F90
```diff
@@ -65,6 +66,17 @@ module modal_aer_opt
 character(len=4) :: diag(0:n_diag) = (/'    ','_d1 ','_d2 ','_d3 ','_d4 ','_d5 ', &
                                        '_d6 ','_d7 ','_d8 ','_d9 ','_d10'/)
 
+!Declare the following threadprivate variables to be used for calcsize and water uptake
+!These are defined as module level variables to aviod allocation-deallocation in a loop
+real(r8), pointer :: dgnumdry_m(:,:,:) ! number mode dry diameter for all modes
+real(r8), pointer :: dgnumwet_m(:,:,:) ! number mode wet diameter for all modes
+real(r8), pointer :: qaerwat_m(:,:,:)  ! aerosol water (g/g) for all modes
+real(r8), pointer :: wetdens_m(:,:,:)  ! aerosol water (g/g) for all modes
+!$OMP THREADPRIVATE(dgnumdry_m, dgnumwet_m, qaerwat_m)
```
ambrad, line 75: @singhbalwinder should wetdens_m appear in the threadprivate list?

singhbalwinder: `wetdens` is not being used in the code (or impact any calculations) and I have removed it entirely in my revised code (PR pending). 

I created this PR (with minimal changes)so that we can get all the NBFB changes in the master. Sorry about the confusion.

components/eam/src/physics/cam/modal_aer_opt.F90
```diff
@@ -585,17 +613,32 @@ subroutine modal_aero_sw(list_idx, state, pbuf, nnite, idxnite, is_cmip6_volc, e
...
    if (list_idx == 0) then
-      ! water uptake and wet radius for the climate list has already been calculated
-      call pbuf_get_field(pbuf, dgnumwet_idx, dgnumwet_m)
-      call pbuf_get_field(pbuf, qaerwat_idx,  qaerwat_m)
+
+      ! Calculate aerosol size distribution parameters and aerosol water uptake
+      ! For prescribed aerosol codes
+      if (clim_modal_aero .and. .not. prog_modal_aero) then
+         ! diagnostic aerosol size calculations
+         call modal_aero_calcsize_diag(state, pbuf, list_idx, dgnumdry_m)
```
ambrad, line 621: @singhbalwinder there might be another potential issue, if I've traced the code correctly.

1. Here, dgnumdry_m is passed to modal_aero_calcsize_diag.
2. dgnumdry_m was allocated during initialization as a module-level array that persists throughout the run.
3. But modal_aero_calcsize_diag calls `allocate(dgnum_m(pcols,pver,nmodes), stat=stat)` for the given set of arguments to modal_aero_calcsize_diag that we see at this line.
4. Thus, it appears that dgnumdry_m is allocated at each call to modal_aero_calcsize_diag. But because this module assumes dgnumdry_m is allocated just once at initialization, these subsequent allocations are not deallocated. And since dgnum_m has type `pointer` rather than `allocatable`, it won't be auto-deallocated.

Sorry if I got that wrong.

singhbalwinder: Thanks Andrew. This indeed seems like a bug. My goal was to remove “calcsize_diag” entirely and replace it with “calcsize_sub” but there isn’t enough time now to remove it and test it throughly. I was testing “calcsize_sub” routine all the time because it is one relevant for prognostic aerosols. “calcsize_diag” is never called for prognostic aerosols now.

This routine is called from different places in the code with different configurations. My intention here was to get dry size (dgnum) using the "fresh state” even with list_dix = 0. The extra allocate should be removed and we should expect dgnum to be already allocated in this routine. Thanks again for catching this!

PR #3890, opened 2020-10-14
Update mksurfdata defaults for 0.5x0.5 res and fix a small bug

aldivi:
The default filenames for inputs to mksurfdata at 0.5x0.5 res are updated.
A PFT tolerance in the normalizencheck_landuse has been updated to match its neighbor and to
match the pft tolerance in the mkpft function, which is called prior
to the call to normalizencheck_landuse.

[BFB]


aldivi:
This has been tested on Compy and compared with previous outputs from NERSC.
I have updated the land surface file documentation page to include the details for running mksurfdata on Compy (https://acme-climate.atlassian.net/wiki/spaces/ED/pages/1708589675/Creating+land+surface+datasets+for+E3SM)

bishtgautam (approved):

rljacob:
This can go in.

PR #3892, opened 2020-10-14
Machine file changes for PNNL's Sooty

singhbalwinder:
This machine is an old machine but it is still useful.

[BFB] - Bit-For-Bit

### DIFF
--- a/cime_config/machines/config_pio.xml
+++ b/cime_config/machines/config_pio.xml
@@ -5,6 +5,7 @@
   <entry id="PIO_VERSION">
     <values>
       <value>2</value>
+      <value mach="sooty">1</value>
     </values>
   </entry>
 
--- a/cime_config/machines/config_compilers.xml
+++ b/cime_config/machines/config_compilers.xml
@@ -969,6 +969,28 @@ flags should be captured within MPAS CMake files.
   </SLIBS>
 </compiler>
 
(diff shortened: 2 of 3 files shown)


singhbalwinder:
@jayeshkrishna : This machine has an old compiler so PIO 2 (SCORPIO) doesn't work on this machine. I have set PIO version for this machine as 1 in the machine files. Is that an okay place to set machine defaults?

cime_config/machines/config_pio.xml
```diff
@@ -5,6 +5,7 @@
   <entry id="PIO_VERSION">
     <values>
       <value>2</value>
+      <value mach="sooty">1</value>
```
singhbalwinder, line 8: @jayeshkrishna : This is where I have set PIO to be 1 for this machine.

jayeshkrishna (approved):
Looks good to me. Did you test the changes on another machine (compy)? Just to make sure that the change (PIO version change for ex) does not affect other machines.

singhbalwinder:
Thanks @jayeshkrishna . I did a quick test on Compy and the selected PIO version was 2 (which is what we expect). I think this PR is ready to go and I will merge it when the repo is open.

rljacob:
This can be merged.  No impact on WC

singhbalwinder:
Thanks @rljacob . I have merged it to next.

PR #3896, opened 2020-10-19
Mass test: Double tol on CO2_FF global mass based on compy_pgi change.

ambrad:
Fixes the postrun-script failure in the mass conservation test. A 2x increase in the tolerance is reasonable.

[BFB]

### DIFF
--- a/components/homme/utils/e3sm_test/check_mass_conservation.py
+++ b/components/homme/utils/e3sm_test/check_mass_conservation.py
@@ -91,7 +91,7 @@ tracer_idx = parse_tracer_index(atm_fn, tracer_name)
 print('Tracer {} has index {}'.format(tracer_name, tracer_idx))
 d = gather_mass_data(atm_fn, tracer_idx)
 good = (conservative('dry M', d['day'], d['dryM'], 1e-11, True) and
-        conservative('tracer {}'.format(tracer_idx), d['day'], d['qmass'], 1e-13, True))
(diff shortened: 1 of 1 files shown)


ambrad:
This is in response to PR #3885.

singhbalwinder (approved):
We should probably also note here that @ambrad tested with different compilers and with and without compiler debug flags. The test PASSed with the Intel compiler and with the PGI when run with the debug flags. The test fails when it is run with the PGI compiler without the debug flags. Debug flags includes changing the optimization to 0, therefore this failure might be related to the PGI compiler optimizations.

singhbalwinder:
@wlin7 : I assigned you this PR but please feel free to reassign.

ambrad:
Good point @singhbalwinder. Output was as follows:

compy_pgi, no debug
```
Tracer CO2_FFF has index 42
dry M               : mass rel err 2.995e-14 tol: 2.454e-13
tracer 42           : mass rel err 3.934e-15 tol: 2.454e-15  <- so 2x bump fixes this
FAIL
```

compy_pgi, debug
```
Tracer CO2_FFF has index 42
dry M               : mass rel err 3.993e-14 tol: 2.454e-13
tracer 42           : mass rel err 1.475e-15 tol: 2.454e-15
PASS
```

compy_intel, no debug
```
Tracer CO2_FFF has index 42
dry M               : mass rel err 2.995e-14 tol: 2.454e-13
tracer 42           : mass rel err 6.557e-16 tol: 2.454e-15
PASS
```

wlin7:
Merged to next and master.

PR #3900, opened 2020-10-23
Defense against use of user_nl_cam/user_nl_clm after name changes

wlin7:
When recycling a run script for a new case, one may not realize
the case specific namelist settings may still go to user_nl_cam.
This change is to force users to rename it to user_nl_eam before
proceeding, if user_nl_cam is still in use, or present in the
case directory. Similar is done for user_nl_clm.

[BFB]


components/eam/bld/build-namelist
```diff
@@ -138,6 +138,23 @@ my $commandline = "$cfgdir/build-namelist @ARGV";
 
 # Process command-line options.
 
+if ( -e "../../user_nl_cam") {
```
singhbalwinder, line 141: Do we store path to the case directory somewhere in this script? It will be better to use that instead of a relative path. If not, this should also be okay.

singhbalwinder (approved):
Thanks Wuyin for working on this. It looks good!

wlin7:
This check is on when running any of the follow steps: case.setup, case.build, case.submit, preview_namelist. So it is effective even after the model is built but then wants to make changes by adding settings to user_nl_cam. 

When user_nl_cam is present in the case directory, running any of the above commands would produce a notice as follows,
                   ********************************************************************
                   NOTICE: 
                   atm name has been changed to EAM. Please use user_nl_eam instead of 
                   user_nl_cam for non-default namelist settings. You may change it in 
                   your run script or rename the file in your case directory! 
                   ********************************************************************
This message is embedded in the build failure error. The special format of the quoted text is intended to give a clean message on the screen, separating from other text in the error message.

components/eam/bld/build-namelist
wlin7: Thanks for the quick check, @singhbalwinder . build-namelist always runs from Buildconf/eamconf (until it is ever changed), so I feel safe just to use the relative path. Plus, we expect to remove it in probably not too distant future after we all get use to it.

PeterCaldwell:
Thanks for implementing this, Wuyin!

components/eam/bld/build-namelist
wlin7: To answer your question, Balwinder.  build-namelist ran with several command-line options, including the following which can be parse to get the case directory
`-infile /global/cfs/cdirs/e3sm/wlin/E3SM_testings/test_user_nl_cam/case_scripts/Buildconf/eamconf/cesm_namelist`

After a check at the point of build-namelist is invoked, I am surprised to see no case specific ENV variables available to it, other than 'PWD', which is the fullpath to Buildconf/eamconf.

wlin7:
> Thanks for implementing this, Wuyin!

Sure, Peter. The initial thought was spawned from our discussion during the DYAMOND-2 meeting. It was then refined from IG  with suggestions from Balwinder et al.

components/eam/bld/build-namelist
singhbalwinder: Thanks for looking into it. I think it is fine as it is as (like you said), it is going to be temporary.

rljacob:
Thanks @wlin7 .  But I was hoping you would modify buildnml right before build-namelist is called.  Then you can get things like caseroot.

wlin7:
> Thanks @wlin7 . But I was hoping you would modify buildnml right before build-namelist is called. Then you can get things like caseroot.

Agreed @rljacob checking it in buildnml would catch it one step earlier in the process and can see all the case info. Since the relative path is always valid, we don't need to involve caseroot here. Of course, one big reason I went with build-namelist is because perl is still my default scripting language option.  Can we go with it for now?

wlin7:
@rljacob , moved the check to buildnml. With this, the error message returned to user is cleaner  because buildnml does not have to capture the  message via run_cmd_no_fatal and embed it the failure error. When user_nl_cam is detected, it would now simply print the following and stop. Previously, it would print a lengthy message including where in build-namelist the execution is aborted, along with the full command line options.

```
   Calling /global/u2/w/wlin/ACME/integration/postCIME5/E3SM.activeMerge/components/eam/cime_config/buildnml

 *** STOP: It appears you are still using user_nl_cam, please change it to user_nl_eam. ***

```
Thanks.

wlin7:
Similar check is added for user_nl_clm

rljacob (approved):
Looks good!

wlin7:
Merged to next. Thank you @singhbalwinder and @rljacob for the tips leading to a cleaner version.

bishtgautam:
The name of this PR is a little misleading as it also catches the error for `user_nl_clm`.

@wlin7 Thanks for fixing the land model build scripts too.

wlin7:
> The name of this PR is a little misleading as it also catches the error for `user_nl_clm`.
Thanks for pointing this out, @bishtgautam . It has been updated.

wlin7:
Nightly tests catches the test active-v1cmip6 is still using user_nl_cam. No problem with user_nl_elm as the correct filename has been in use.

Update the branch with user_nl_cam renamed, and remerged to next.

The settings in v1cmip6/user_nl_eam are for output only. It may still cause 'non-BFB' if multiple history files are checked.

NML diff will occur after remerging, compared to the state before this PR.

rljacob:
I missed that one.   That will create a real diff since the extra history files are currently missing (or old) in the baseline.

wlin7:
Merged to master. 

Request submitted to bless allactive-v1cmip6 for sandiatoss3. The compy test failed today, to be blessed when FAIL is cleared.

PR #3902, opened 2020-10-26
Adds saving of GM induced MOC streamfunction

mark-petersen:
This PR adds the ability to compute and separately save the MOC stream function due to GM bolus velocity. This is saved separately from the MOC from the resolved velocity. See https://github.com/MPAS-Dev/MPAS-Model/pull/726

[BFB]
[NML]

### DIFF
--- a/components/mpas-ocean/cime_config/buildnml
+++ b/components/mpas-ocean/cime_config/buildnml
@@ -1029,6 +1029,8 @@ def buildnml(case, caseroot, compname):
             lines.append('    <var name="normalGMBolusVelocitySquared"/>')
             lines.append('    <var name="gmKappaScaling"/>')
             lines.append('    <var name="gmBolusKappa"/>')
+            lines.append('    <var name="mocStreamvalLatAndDepthGM"/>')
+            lines.append('    <var name="mocStreamvalLatAndDepthRegionGM"/>')
 
         lines.append('    <var_struct name="tracersSurfaceFlux"/>')
         lines.append('    <var name="penetrativeTemperatureFlux"/>')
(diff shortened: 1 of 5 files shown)


components/mpas-ocean/cime_config/buildnml
```diff
@@ -1029,6 +1029,8 @@ def buildnml(case, caseroot, compname):
             lines.append('    <var name="normalGMBolusVelocitySquared"/>')
             lines.append('    <var name="gmKappaScaling"/>')
             lines.append('    <var name="gmBolusKappa"/>')
+            lines.append('    <var name="mocStreamvalLatAndDepthGM"/>')
+            lines.append('    <var name="mocStreamvalLatAndDepthRegionGM"/>')
```
mark-petersen, line 1033: @vanroekel I put these in the monthly time average, with the other GM variables.

mark-petersen:
@jonbob it was simple for me to remove the two namelist flags from E3SM, so I did. I think this is completely ready.

components/mpas-ocean/cime_config/buildnml
vanroekel: great, thanks!

vanroekel (approved):
looks good to me. I don't know if it is okay to just delete flags from the namelist areas or if there has to be something done with the automated scripts.  So will have to wait for final check from @jonbob

jonbob:
@rljacob - I believe @vanroekel wants this for v2, if we can add the milestone

vanroekel:
yes, this would be good for v2 and has no impact on the simulation it is purely analysis.

jonbob:
I can confirm that the changes to the bld files are exactly what I get from running the automated scripts

jonbob:
merged to next

jonbob:

PR #3903, opened 2020-10-26
bug fix for nitrogen fixation calculation for ECA

### LINKED ISSUE #3895: negative nitrogen fixation due to negative moisture content
Because the current solver of the Richards' equation may occasionally produce negative water content, the updated soil water content when used for nitrogen fixation calculation will result in unphysical negative nitrogen fixation in the subroutine NitrogenFixation_balance. A short-term fix is applied in NitrogenFixation_balance. However, a long-term solution is to make sure the Rcihards' equation solver more robust.

jinyuntang:
This PR does a quick fix to avoid negative nitrogen fixation
when the soil water content computed from the Richards'
equation solver is negative. It also corrects some typos
in the flux limiter solver for phenology update. 

Fixes #3895
[non-BFB] for ECA


bishtgautam (approved):
Looks good

bishtgautam:
@jinyuntang Is this BFB?

bishtgautam:
Nevermind, I read the full description of the PR.

PR #3917, opened 2020-10-28
Add user_nl_mpassi for elm bgcexp testmod in order to pass automated tests

### LINKED ISSUE #3883: Two BGC configurations fail automated testing
Two BGC configurations that are part of automated testing now FAIL because the files they use for initial conditions have the incorrect number of snow layers. I can think of two quick solutions:
1. spinup up some short results using a G-case and replace the files these tests point at; or
2. set the number of snow layers to 1 in user_nl_mpassi for the purpose of testing.
Mat, Nicole -- any preference on this?

jonbob:
This PR adds user_nl_mpassi to the elm bgcexp tested, with a single setting to make nSnowLayers = 1. This will allow two BGC tests in the automated suites to pass, that have otherwise been failing since [PR #3846](https://github.com/E3SM-Project/E3SM/pull/3846) was merged. 

Note: this PR is for testing use only and not to be used for science runs.

Fixes #3883 

[NML]
[non-BFB] for the two automated tests, because #3846 will have changed answers once these tests run again

### DIFF
--- a/components/elm/cime_config/testdefs/testmods_dirs/elm/bgcexp/user_nl_mpassi
+++ b/components/elm/cime_config/testdefs/testmods_dirs/elm/bgcexp/user_nl_mpassi
@@ -1,0 +1,1 @@
+config_nsnowlayers = 1


jonbob:
Passes SMS.ne30_oECv3.BGCEXP_BCRC_CNPRDCTC_1850.anvil_intel.elm-bgcexp

rljacob (approved):

jonbob:
merged to next

jonbob:

Issue #3758, opened 2020-08-05
Need more output from cmake


rljacob:
When running ./case.build interactively, this is some of what the user sees output to the screen:
```
Building pio with output to file /compyfs/jaco316/e3sm_scratch/Ftest/bld/pio.bldlog.200804-185823
   Calling /qfs/people/jaco316/E3SM3/cime/src/build_scripts/buildlib.pio
Building csm_share with output to file /compyfs/jaco316/e3sm_scratch/Ftest/bld/csm_share.bldlog.200804-185823
   Calling /qfs/people/jaco316/E3SM3/cime/src/build_scripts/buildlib.csm_share
Component csm_share build complete with 20 warnings

Time spent not building: 2.689481 sec
Time spent building: 581.197731 sec
MODEL BUILD HAS FINISHED SUCCESSFULL
```
There is nothing output to the screen for several minutes after the "Component csm_share..." message because at that point cmake takes over and builds the rest of the model including the time-consuming full models.

It would be good if the build system output a message listing the components cmake was building and where to find the log file.

worleyph:
@rljacob , I asked this question of @jgfouca when CMake was first introduced - there are no separate build logs for components any more (anything built with CMake). Everything is dumped into e3sm.log, and there is also no separation between (i.e., no timing associated with) the builds for each component. Everything is mushed together.

rljacob:
I don't necessarily want them separated out.  I just want case.build to report what is happening and point to the location of e3sm.log and not go quiet when cmake kicks in.

sarats:
I do think there is value in getting separate logs to debug issues etc. Just wondering if there would be a way to separate out logs as components will be different CMake targets?
If possible, compilation timing info is also useful to figure out if certain optimizations for files should be turned off etc. 
Just my $0.02.

rljacob:
I'd like to see a message after csm_share like:
Building rest of model with output to file /compyfs/jaco316/e3sm_scratch/Ftest/bld/e3sm.bldlog.200804-185823

jonbob:
Maybe there could be an option to separate output by component. It certainly is one of the things I find most irritating about the current cmake system -- it can be incredibly painful to figure out where errors are coming from.

jgfouca:
Fixed by #3918

PR #3918, opened 2020-10-28
Remove support for Makefile-based builds for E3SM components

### LINKED ISSUE #3769: case.build --use-old
Case created with `./cime/scripts/create_test SMS.ne4_ne4.FC5AV1C-L` with latest master is running into error with `case.build --use-old`:
```
azamat@blueslogin4:SMS.ne4_ne4.FC5AV1C-L.anvil_intel.20200813_105247_2q1j5j$ tail -n 2 bld/e3sm.bldlog.200813-105654 
gmake: *** No rule to make target `/lcrc/group/acme/azamat/acme_scratch/anvil/SMS.ne4_ne4.FC5AV1C-L.anvil_intel.20200813_105247_2q1j5j/bld/lib/libocn.a', 
needed by `/lcrc/group/acme/azamat/acme_scratch/anvil/SMS.ne4_ne4.FC5AV1C-L.anvil_intel.20200813_105247_2q1j5j/bld/e3sm.exe'.  Stop.
gmake: *** Waiting for unfinished jobs....
```

jgfouca:
Build system details:
1. Add new --separate-builds flag for case.build. Component builds will be done one at a time and produce separate log files for each component.
2. Remove --use-old flag. The old system is not being maintained or tested and it's no longer safe to use.
3. Produce more logging output for case.build.
4. Re-establish support for building only specific components with case.build. This capability had been lost in the initial cmake conversion. Building components this way will cause them to get their own log file.
5. Remove buildlib and Depends* files that are no longer needed
6.Adds new compiler/link wrapper: E3SM/cime/scripts/Tools/e3sm_compile_wrap.py. It will get copied to $case/Tools and this copy is the one that will be used by the build system. This will allow the script to be a new extension point for our build system that can be customized on a case-by-case basis or customized for everyone via PRs to CIME.

Fixes #3769 #3758 

Updates from CIME:
* Fix E3SM nightly TSC test failure
* Some build system updates for e3sm
* Add support for github actions

[BFB]


worleyph:
On Compy:

     $ ./create_newcase --case SCALE_A_WCYCL1850_ne4pg2_r05_oQU480b_sepbld -res ne4pg2_r05_oQU480 -compset A_WCYCL1850 -mach compy -compiler intel -project e3sm

then modified env_mach_pes.xml to change PE count for the components from 96 to 64 (graph partitions for OCN and ICE do not exist for 96 for oQU480). 

./case.setup, ./case.build, and ./case.run worked successfully. However,

     $ ./case.build --separate-builds
     ...
     Building ice model with output to file /compyfs/worl312/e3sm_scratch /SCALE_A_WCYCL1850_ne4pg2_r05_oQU480b_sepbld/bld/ice.bldlog.201029-072913
        Calling make, see top of log file for specific call
     ERROR: BUILD FAIL: build ice failed, cat /compyfs/worl312/e3sm_scratch/SCALE_A_WCYCL1850_ne4pg2_r05_oQU480b_sepbld/bld/ice.bldlog.201029-072913

The error message in ice.bldlog is

     /bin/sh: /compyfs/worl312/e3sm_scratch/SCALE_A_WCYCL1850_ne4pg2_r05_oQU480b_sepbld/bld/cmake-bld/mpas-source/src/tools/parse: No such file or directory

jgfouca:
@worleyph , it looks like, at some point in the past, some of the tool dependencies got broken in MPAS. This PR:
https://github.com/MPAS-Dev/MPAS-Model/pull/737

will fix it.

worleyph:
@jgfouca , will you update this branch with that fix when it becomes available? It is important to demonstrate "--separate-builds" with a A_WCYCL case. Thanks.

jgfouca:
@worleyph , yes. I hope @mark-petersen can help me expedite that MPAS PR and get it into E3SM ASAP.

jgfouca:
@worleyph , the MPAS updates are in E3SM master and I've upstream-merged this branch. It should be ready to go.

jgfouca:
I'm going to see if I can make the timing feature a bit nicer, hold off on reviews.

jgfouca:
OK, this is ready now.

sarats:
Is the timing info on by default?

jgfouca:
@sarats , no. Due to the interleaved output you get when `make -jN` and N>1, the timing output can be very difficult to interpret. In order to avoid this situation, asking for timing (`./case.build --timing`) sets the build to serial (`-j1`) to this feature is off by default.

worleyph:
@jgfouca,

      "Due to the interleaved output you get when make -jN and N>1, 
      the timing output can be very difficult to interpret. In order to avoid 
      this situation, asking for timing (./case.build --timing) sets the build
      to serial (-j1)"

Wouldn't it be better to leave this for the user to decide? They can always set -j1 in the env_build.xml .

worleyph:
@sarats , the total time to build each component is listed at the end of the component bldlog file (when using --separate-builds), even when --timing is not specified.

worleyph:
@jgfouca , how do I parse this?

     Elapsed time: 2 s. (time), 0 s. (clock)

(time vs. clock? clock is always zero? resolution of timer is seconds?  Seems pretty coarse for compiling individual files.)

worleyph:
"Due to the interleaved output you get when make -jN and N>1, 
       the timing output can be very difficult to interpret."

You could annotate each timing output line to indicate which file the timing output refers to?

worleyph:
> They can always set -j1 in the env_build.xml .

in env_run.xml I mena . (Keep forgetting where this is.)

jgfouca:
Apologies for the churn, I think I can improve on this such that the output interleaving is not an issue.

jgfouca:
@worleyph , I've arrived at a solution that solves the interleaving problem with timing output. I'm a little concerned that it's not as portable as the last solution though. Do any of you use tcsh as your login shell?

worleyph:
@jgfouca , I use tcsh on some systems (but not compy where I am currently doing testing). You might ask @sarat exactly what is wanted from timing. In the "old days" there was time spent building each component model (reported as part of the output from case.build to the screen? to CaseStatus? a little fuzzy). The per file compilation cost seemed like it was generated by the compilers? is there a compiler flag for this, and can this output be redirected to the the Cmake bldlogs, like it was for the non-Cmake builds?

jgfouca:
@worleyph , timings are now available per compilation, link, and component (depending on case.build flags), so I don't know where I could add any more timing info. If wanted to rely on the compilers themselves to provide the timings, there'd be two major issues: first, we'd have to figure out what flags each compiler vendor needed and add them to config_compilers. Second, due to the way Cmake/Make prints commands as they are launched, I think we'd be back to dealing with confusing interleaved output.

worleyph:
Prefer that @sarat describe what he is most interested in. I'm not the primary customer, and am uncomfortable driving the requirements (just reporting what I vaguely remember being available in the old build system).

My shell on Cori is /bin/csh . I can try out your new approach there as well, if you wish.

jgfouca:
@worleyph , yes, plz check `./case.build --timing` on any case on cori.

worleyph:
@jgfouca , testing today (both on Compy, with latest update, and on Cori-KNL). Please note that

     ./case.build --help

needs to be updated, as it still has

        --timing              Enable timing of all compilations and links, will force GMAKE_J to be 1 to avoid  interleaved output

and GMAKE_J is not set to 1 in your latest modification, correct?

worleyph:
@jgfouca , on Compy I am now seeing (what I assume) is the compile time appended to each compile line, e.g.

     ... -o CMakeFiles/atm.dir/__/__/eam/src/control/cam_instance.F90.o 0:00.18

This doesn't always work with the interleaved output, e.g.

     ... -I/qfs/people/worl312/E3SM/jgfouca/cmake_misc/E3SM/components/eam/src/chemistry/pp_linoz_mam4_resus_mom_soag0:00.18

but seems to be pretty good most of the time. I defer to others whether this format is useful. I imagine that "grep"ing for or sorting the timings to look for large costs might be a problem, but a simple perl/sed/??? script could probably extract this information for easier postprocessing. Having some sort of string to grep for might still help, but don't know what that would involve to implement.

My experiments on Cori-KNL (where my shell is csh) worked as well, with the same timing output.

Once people sign off on the format, is there any reason that --timing could not be the default? (Adding instead a --notiming flag, or just make it on all the time?)

worleyph:
@jgfouca , what is the format of the timing data:

     0:00.18

minute:second? hour:minute?

worleyph:
Others - please speak up - not sure what aspect of timing the builds is considered useful. @jgfouca , was testing maint-1.2, which has the old Make logic. The ./case.build spits out ...

     sesp built in 8.606230 seconds
     sglc built in 8.640947 seconds
     swav built in 8.885189 seconds
     srof built in 9.671653 seconds
     docn built in 12.425522 seconds
     Component ice build complete with 1 warnings
     cice built in 52.073996 seconds
     Component atm build complete with 1 warnings
     cam built in 142.792368 seconds
     Component lnd build complete with 1 warnings
     clm built in 230.511388 seconds
     Building e3sm with output to /compyfs/worl312/e3sm_scratch/FC5AV1C-L_ne4_ne4_m1.2_bibi/bld/e3sm.bldlog.201105-093804 
     Time spent not building: 3.698927 sec
     Time spent building: 574.816288 sec
     MODEL BUILD HAS FINISHED SUCCESSFULLY

Perhaps this is what people want (readily available). This is available via the CMAKE logic, but is buried within the bldlg files when building the components separately. Only timing data visible from the call to ./case.build with the CMAKE logic is

     Time spent not building: 2.874031 sec
     Time spent building: 3017.987297 sec
     MODEL BUILD HAS FINISHED SUCCESSFULLY

(Large time spent building in the aboive is due to GMAKE_J == 1 and building with all active components. Previous output was for an F case with default GMAKE_J .)

amametjanov:
+1 on the maint-1.2 style of component/library build timing output. Per-file build time (new feature) is also useful. Compilers' internal build-time flags produce different types of output (depends on a compiler). Having cmake output build times (independent of compiler) would be useful.

jgfouca:
@worleyph , thanks for looking so closely at this.

> and GMAKE_J is not set to 1 in your latest modification, correct?

Correct, I will update.

> This doesn't always work with the interleaved output, e.g.

I wonder if there's a cap on the line length that `/usr/bin/time --format "%C" ` is able to handle? If so, there won't be much that I can do about that. Was that build command especially long?

> I imagine that "grep"ing for or sorting the timings to look for large costs might be a problem

I agree. The format is trivial to change... how would you feel about this: `--format '%C\\nElapsed time: %E'` ? That way, you could do a trivial grep on the log for `Elapsed time` and visually scan for large values. If then opened the log and searched for the large time, it should take you straight to the expensive compilation.

> minute:second? hour:minute?

M:SS.hundredths_of_seconds. Please advise if you think there's a way to make this more clear in the output.

> Once people sign off on the format, is there any reason that --timing could not be the default?

I was wondering that too since it now incurs no overhead. My concerns are:
1) Portability of the /usr/bin/time injection. I think it's pretty portable on Linux, but I know it won't work on BSD/Mac.
2) Output slicing problems like you saw 
3) Slight differences in output behavior could surprise users. The standard behavior prints commands as they are launched, the --timing behavior prints commands as they are completed, meaning that commands will be printed after compilation errors and warnings.

> Perhaps this is what people want (readily available).

I can understand that, but remember, with cmake the default behavior is for the build to be done in one big step, so I will not be able to produce the per-component built times unless `--separate-builds` is provided. That said, I can add some additional timings to the top-level output (not the log file) when `--separate-builds` is on.

worleyph:
> cmake the default behavior is for the build to be done in one big step

Is this necessary? If we "voted", would there be any downside to making "--separate-builds" the default?

> Was that build command especially long?

SInce it was "interleaved" I didn't look very hard. The file in question is 

     /compyfs/worl312/e3sm_scratch/SCALE_A_WCYCL1850_ne4pg2_r05_oQU480xc_sepbldfix_timing/bld/atm.bldlog.201105-061207.gz

sarats:
Late to the party, too many firefights this week. I haven't had a chance to test-drive this branch myself, so these are my initial (not fully informed) thoughts.

As Pat alluded to, component level build times are the initial motivation. I always looked at individual file-level times using compiler flags myself but a CMake level option is desirable/preferred. 

The timing output that is amenable for parsing would be quite useful. If possible, could we add the current CMake Object along with Elapsed time for easier correlation?

sarats:
At the risk of feature creep, is it feasible to print file name and build time to a separate file (e.g., build-times.out)? That would even be helpful to track build time externally (like PACE) over time.

jgfouca:
@worleyph , @sarats 

> Is this necessary? If we "voted", would there be any downside to making "--separate-builds" the default?

Yes, splitting builds into seperate chunks serializes the build at a high level and will reduce build performance. For example, on mappy, the difference between a A_WCYCL1850 build normally and with --separate-builds is 95s vs. 256s.

> The timing output that is amenable for parsing would be quite useful. If possible, could we add the current CMake Object along with Elapsed time for easier correlation?

Unfortunately, with the current approach of using `/usr/bin/time`, this is not possible. If we want to get really fancy here, instead of using `/usr/bin/time` as our wrapper, we could use a custom script that we bundle with E3SM or CIME. With a custom script, there's no limit on the functionality we can add. This could also help with portability.

jgfouca:
@rljacob , can you weigh-in here? Do we want to write our own compiler wrapper? I don't think it would require a ton of work or much maintenance and it would give us the ability to add a variety of features.

rljacob:
If its not a ton of work and allows us to do more of what we want, sounds ok to me.

jgfouca:
@rljacob , @worleyph , @sarats , OK, the wrapper is complete and will now always be on.

The default wrapper is in E3SM/cime/scripts/Tools/e3sm_compile_wrap.py and it will get copied to $case/Tools and this copy is the one that will be used by the build system. This will allow the script to be a new extension point for our build system that can be customized on a case-by-case basis or customized for everyone via PRs to CIME.

 I measured some build times on mappy and the injection of this wrapper seems to add negligible slowdowns for builds (about 1%).

jgfouca:
@amametjanov , ^^

amametjanov:
`case.build --timing` appears to have been removed. Need to update PR description at the top.

`case.build --separate-builds` is working:
* output of `time` at the bottom of component build logs
```
$ zcat bld/atm.bldlog.201119-185150.gz | tail -3
real    2m59.913s
user    11m0.759s
sys     4m26.189s
```
* stdout of build is printing total component build time
```
$ grep "built in" case-build2.log 
   atm built in 179.940130 seconds
   lnd built in 315.923113 seconds
   ice built in 181.151319 seconds
   ocn built in 151.314544 seconds
   rof built in 40.765838 seconds
   glc built in 1.257437 seconds
   wav built in 1.223882 seconds
   iac built in 1.365059 seconds
   esp built in 1.234944 seconds
   cpl built in 26.443022 seconds
```
* answers to built-time questions like "what's taking so long?"
```
$ zgrep "built in" bld/*.bldlog.* | awk '$5 > 30' | sort -n -k5,5
bld/atm.bldlog.201119-185150.gz:Target CMakeFiles/atm.dir/__/__/eam/src/physics/rrtmg/ext/rrtmg_lw/rrtmg_lw_k_g.f90.o built in 32.999894 seconds
bld/atm.bldlog.201119-185150.gz:Target CMakeFiles/atm.dir/__/__/eam/src/physics/cam/unicon.F90.o built in 37.143444 seconds
bld/atm.bldlog.201119-185150.gz:Target CMakeFiles/atm.dir/__/__/eam/src/physics/clubb/mt95.f90.o built in 57.079373 seconds
bld/lnd.bldlog.201119-185150.gz:Target CMakeFiles/lnd.dir/__/__/elm/src/data_types/ColumnDataType.F90.o built in 57.262427 seconds
bld/lnd.bldlog.201119-185150.gz:Target CMakeFiles/lnd.dir/__/__/elm/src/biogeochem/CNCarbonFluxType.F90.o built in 89.443586 seconds
bld/ice.bldlog.201119-185150.gz:Target CMakeFiles/ice.dir/__/__/core_seaice/model_forward/mpas_seaice_core_interface.f90.o built in 90.404366 seconds
bld/ocn.bldlog.201119-185150.gz:Target CMakeFiles/ocn.dir/__/__/core_ocean/driver/mpas_ocn_core_interface.f90.o built in 102.752803 seconds
bld/lnd.bldlog.201119-185150.gz:Target CMakeFiles/lnd.dir/__/__/elm/src/data_types/VegetationDataType.F90.o built in 151.107892 seconds
```
This was with case created by `create_newcase` (`ne4pg2_r05_oQU480.A_WCYCL1850.anvil_intel`).

Thanks for enabling individual file build timing.

amametjanov (approved):

jgfouca:
Great stuff, @amametjanov! I will add those tricks to https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/1011942171/E3SM+Build+System+and+CMake

jgfouca:
@lukaszlacinski , I added you as a reviewer since I needed to make a change to our Travis stuff in order to diagnose CI fails on this branch. `.travis/run.sh` was not printing the e3sm bldlog when the build fails which is, by far, the most useful of all the bldlogs.

sarats:
@jgfouca Is it too much to ask for the build times to be written to a different log file for easier tracking?

jgfouca:
@sarats , it should be easy to do that. I might just use @amametjanov 's shell command as a post-processing step. Are these timings something that should be saved in our provenance?

sarats:
@jgfouca Yes, adding these to PACE via provenance files is the plan. So, this would happen automatically if they end up in the performance_archive locally.

sarats:
If I might suggest an output format (easier for tools like Gnuplot):

```
# Target Build_time
/path/sample.f90 10.2 sec
...
```

sarats:
Thinking if component build times should be included in the same file or another one for simplicity.

jgfouca:
@sarats, @worleyph  See https://github.com/ESMCI/cime/pull/3782

> Thinking if component build times should be included in the same file or another one for simplicity.

This probably won't be possible since `--separate-builds` will be off by default.

jgfouca:
Merged to next.

PR #3923, opened 2020-10-30
Allow dynamic crop fraction taking effect when crop model turned off 

hydrotian:
Allow `do_transient_crops = T` to take effect when prognostic crop model is turned off.
In addition to the dynamic crop fraction, this PR also corrects the inconsistency between
land domain file and MOSART input parameter, in terms of land fraction along the coastal line. 
Without this fix, the irrigation demand at coastal grid will be incorrectly assigned in MOSART.

[BFB]

### DIFF
--- a/components/mosart/src/cpl/rof_comp_esmf.F90
+++ b/components/mosart/src/cpl/rof_comp_esmf.F90
@@ -25,7 +25,7 @@ module rof_comp_esmf
   use seq_flds_mod
   use esmf
   use esmfshr_mod
-  use RunoffMod        , only : rtmCTL, TRunoff, THeat
+  use RunoffMod        , only : rtmCTL, TRunoff, THeat, TUnit
   use RtmVar           , only : rtmlon, rtmlat, ice_runoff, iulog, &
                                 nsrStartup, nsrContinue, nsrBranch, & 
                                 inst_index, inst_suffix, inst_name, RtmVarSet, heatflag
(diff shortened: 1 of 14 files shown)


golaz:
@rljacob : FYI, this is part of new land/river features that we are hoping to use for v2.

components/mosart/src/riverroute/RtmHistFlds.F90
```diff
@@ -164,15 +164,15 @@ subroutine RtmHistFldsInit()
          avgflag='A', long_name='WRM supply provided ', &
          ptr_rof=StorWater%Supply, default='active')
```
tanzeli1982, line 165: I worry that the units change would confuse users because time step is not easily found in outputs.

components/mosart/src/cpl/rof_comp_esmf.F90
```diff
@@ -705,7 +705,7 @@ subroutine rof_import_esmf( x2r_array, rc)
        else
           rtmCTL%qdto(n,nliq) = 0.0_r8
        endif
-       rtmCTL%qdem(n,nliq) = fptr(index_x2r_Flrl_demand,n2) * (rtmCTL%area(n)*0.001_r8)
+       rtmCTL%qdem(n,nliq) = fptr(index_x2r_Flrl_demand,n2) / TUnit%domainfrac(n) * (rtmCTL%area(n)*0.001_r8)
```
tanzeli1982, line 708: It is better to make sure TUnit%domainfrac > 0.

components/mosart/src/cpl/rof_comp_mct.F90
```diff
@@ -596,7 +596,7 @@ subroutine rof_import_mct( x2r_r)
        else
           rtmCTL%qdto(n,nliq) = 0.0_r8
        endif
-       rtmCTL%qdem(n,nliq) = x2r_r%rAttr(index_x2r_Flrl_demand,n2) * (rtmCTL%area(n)*0.001_r8)
+       rtmCTL%qdem(n,nliq) = x2r_r%rAttr(index_x2r_Flrl_demand,n2) / TUnit%domainfrac(n) * (rtmCTL%area(n)*0.001_r8)
```
tanzeli1982, line 599: Same above.

tanzeli1982 (commented):
@hydrotian I have several minor comments. It should easy to handle.

tanzeli1982 (approved):

components/mosart/src/cpl/rof_comp_esmf.F90
hydrotian: Added check when read in the data

hydrotian:
@bbye could you please take a look at the PR when you get time? Thanks!

bbye (approved):

PR #3926, opened 2020-11-01
Initialize radiation outputs to fillvalue

### LINKED ISSUE #3906: Radiation outputs are zero when nhtfrq < irad[s,l]w
We recently discovered while running simulations with very high frequency output that there's an interesting "feature" (read: "bug"?) in our radiation fields output that requires some careful user attention if trying to get meaningful high frequency outputs. Since radiation is often called somewhat infrequently to save cost, it's possible to select an output frequency that's shorter than the radiation update frequency. When this happens, the radiation outputs will apparently contain zeros for the times between radiation update calls. For example, if radiation is only called every hour (the default if not specified explicitly in a namelist), but you try to write history files every half hour, then every other time index in the output files will contain all zeros. This is true for both RRTMG and RRTMGP, but only shows up when trying to write at intervals shorter than the radiation update interval (i.e., when `nhtfrq < irad[s,l]w`). I think this is actually because the output fields are not defined with the flag `flag_xyfill=.true.` set. At least with that flag, intermediate times are filled with `fillvalue`, so that when taking means over a time period you aren't inadvertently averaging a bunch of zeros (something we found out the hard way). I think we should implement a PR to fix the `addfld` calls to add this flag to prevent this from happening in the future, but for now users should take care to make sure that `nhtfrq >= irad[s,l]w`.

brhillman:
Initialize radiation outputs to fillvalue. This addresses the case where radiation outputs can contain zero values when the frequency of updating the radiative calculations does not evenly divide the output frequency. Instead, these should be filled with fillvalues so that averages are not biased low by including zero values which should have just been set to missing values. Also add calculation for "HR" heating rate variables, which was defined as an output for RRTMGP but was never actually populated with values. This fix will be BFB for our test suites, but longer RRTMGP runs that include HR as an output will be NON-BFB since we have added the HR calculation, which would have been zero before (this should NOT require new baselines though, since tests in our suite are not long enough to generate h0 files with HR and HR does not appear to be a default output variable).

Fixes #3906.

[BFB]

### DIFF
--- a/components/eam/src/physics/crm/rrtmgp/radiation.F90
+++ b/components/eam/src/physics/crm/rrtmgp/radiation.F90
@@ -1178,6 +1178,9 @@ contains
       real(r8) :: qrsc(pcols,pver)
       real(r8) :: qrlc(pcols,pver)
 
+      ! Temporary variable for heating rate output
+      real(r8) :: hr(pcols,pver)
+
       ! Options for MMF/SP
       logical :: use_MMF
       character(len=16) :: MMF_microphysics_scheme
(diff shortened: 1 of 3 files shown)


whannah1 (approved):
Looks good to me.

singhbalwinder:
Merged with next.

PR #3927, opened 2020-11-02
Update mpas-source: Fix cmake for tool exes

mark-petersen:
see https://github.com/E3SM-Project/E3SM/pull/3918 and https://github.com/MPAS-Dev/MPAS-Model/pull/737. 
Regardless of MPAS_TOOL_DIR, the tool exes need to be targets
and targets that depend of these exes need to always depend on the
needed exe. This will ensure the exes get built regardless of which
targets the user is building.

@jgfouca tested this change with the E3SM case `ne4pg2_r05_oQU480 A_WCYCL1850`. Tested both case.build and case.build --separate-builds with both MPAS_TOOL_DIR enabled and disabled.

[BFB]


mark-petersen:
@jgfouca please approve. @jonbob can merge today if the schedule allows.

mark-petersen:
Passes:
```
```
as sanity tests. But I think that does not test the specific changes here.

jgfouca (approved):
I approve the CMake changes. It looks like there are also changes to src/core_init_atmosphere/mpas_init_atm_static.F

mark-petersen:
Yes, that is expected because we merge from the MPAS `develop` branch. It does not affect E3SM.

jonbob:
can the atm changes be handled separately? is there some reason that has to be included?

mark-petersen:
Because the workflow is from `develop` to `e3sm/develop`, and MPAS-Atmosphere changes were previously merged into `develop`. I could split them into two E3SM PRs if you prefer.

jonbob:
Yes, in general we're trying to avoid having unrelated changes tag along on these PR's

mark-petersen:
Done. The cmake changes are now merged separately into MPAS develop and MPAS e3sm/develop. We will need to merge `develop` into `e3sm/develop` with atmospheric changes later.

jonbob:
merged to next

jonbob:
passes SMS_D.ne4pg2_r05_oQU480.A_WCYCL1850.anvil_intel sanity test

jonbob:
merged to master

PR #3928, opened 2020-11-02
CMake and code reorg changes in Homme, to accommodate use from inside another cmake project.

bartgol:
These mods are targeted for use within SCREAM. There are two main needs:

1) build homme with cmake but not in standalone mode
2) break up some setup/init functions in order to be able to inject host app code into homme (e.g., swap hommexx kokkos views with some views allocated by the host app).

Note: this PR is replacing (and improving) the abandoned pr #3673 

[bfb]


bartgol:
Standalone unit tests (labels unit and dev) pass.

oksanaguba:
you assigned it to yourself -- are you going to integrate it?

bartgol:
> you assigned it to yourself -- are you going to integrate it?

Ah, I thought the assignee was the one in charge of making requested fixes. I guess I never really understood what the assignee has to do then. :)

rljacob:
We use "assignee" for the integrator since github doesn't provide an "integrator" field.

oksanaguba:
@bartgol in description, does 2) mean instead of homme allocating views they are taken from outside homme?

I assume this PR is supposed to be bfb?

thanks

bartgol:
> @bartgol in description, does 2) mean instead of homme allocating views they are taken from outside homme?

The init functions have been broken up to allow that, yes. So in homme standalone we have something like
```
call prim_create_c_data_structures(...)
call prim_init_elements_views (...)
call prim_init_kokkos_functors (...)
```
while scream (which is the main reason for these mods) will do
```
call prim_create_c_data_structures(...)
call init_homme_views_with_app_ptrs (...)
call prim_init_kokkos_functors (...)
```
This allows the app to be in charge of memory, and homme to simply use it.

> 
> I assume this PR is supposed to be bfb?

Yes.

Thanks!

oksanaguba:
Is this ready? saw another commit yesterday, should I start merging?

bartgol:
I think so. Sorry about the last minute commit. I noticed a bug while running this in scream. Everything seems good now.

components/homme/cmake/HommeMacros.cmake
```diff
@@ -172,6 +175,94 @@ macro(createTestExec execName execType macroNP macroNC
...
+  # First configure the file (which formats the file as C)
+  HommeConfigFile (${THIS_CONFIG_IN} ${THIS_CONFIG_HC} ${THIS_CONFIG_H} )
+
+  ADD_DEFINITIONS(-DHAVE_CONFIG_H)
+
+  ADD_LIBRARY(${libName} ${libSrcs})
+  TARGET_INCLUDE_DIRECTORIES (${libName} PUBLIC ${inclDirs} ${modulesDir} ${CMAKE_CURRENT_BINARY_DIR})
+  SET_TARGET_PROPERTIES(${libName} PROPERTIES Fortran_MODULE_DIRECTORY ${modulesDir})
+  SET_TARGET_PROPERTIES(${libName} PROPERTIES LINKER_LANGUAGE Fortran)
+  IF(BUILD_HOMME_WITHOUT_PIOLIBRARY)
```
oksanaguba, line 222: I thought from reading changes above library build will never have io/netcdf support?
In here
```
IF (HOMME_BUILD_EXECS OR BUILD_HOMME_TOOL)
   SET(BUILD_HOMME_WITHOUT_PIOLIBRARY OFF CACHE BOOL "")
 ELSE ()
   SET(BUILD_HOMME_WITHOUT_PIOLIBRARY ON CACHE BOOL "")
 ENDIF ()
```
?

components/homme/cmake/HommeMacros.cmake
```diff
@@ -172,6 +175,94 @@ macro(createTestExec execName execType macroNP macroNC
...
+
+  STRING(TOUPPER "${PERFORMANCE_PROFILE}" PERF_PROF_UPPER)
+  IF ("${PERF_PROF_UPPER}" STREQUAL "VTUNE")
+    TARGET_LINK_LIBRARIES(${libName} ittnotify)
+  ENDIF ()
+
+  TARGET_LINK_LIBRARIES(${libName} timing ${COMPOSE_LIBRARY} ${BLAS_LIBRARIES} ${LAPACK_LIBRARIES})
+
+  IF (HOMME_USE_KOKKOS)
+    TARGET_LINK_LIBRARIES(${libName} kokkos)
```
oksanaguba, line 240: I asked you this before but I don't remember: 
```
components/cmake/build_model.cmake:      target_link_libraries (${TARGET_NAME} PRIVATE Kokkos::kokkos)
```
Is this  Kokkos::kokkos to not to allow a system-installed library to be used? Does it need to be here as well?

components/homme/cmake/HommeMacros.cmake
```diff
@@ -172,6 +175,94 @@ macro(createTestExec execName execType macroNP macroNC
...
+  ENDIF ()
+
+  IF (NOT HOMME_USE_MKL)
+    IF (NOT HOMME_FIND_BLASLAPACK)
+      TARGET_LINK_LIBRARIES(${libName} lapack blas)
```
oksanaguba, line 245: It seems homme does not use blas/lapack anymore. Maybe we should try to remove this dependency (not in this PR).

bartgol: So it uses mkl, but not lapack? Interesting.

That said, if something is not used, I'm all for removing deps. Especially for libs like blas/lapack that tend to be available in system directories (with potential compiler version clash at link time).

components/homme/cmake/HommeMacros.cmake
bartgol: CMake allows for namespaces for target names. When using namespaces, cmake is assuming it is a cmake target (rather than a library name). I guess it can make things a bit easier.

Some libraries alias their FOO::BAR libraries, so that BAR is a valid target name. I _think_ kokkos doesn't, but I have to check. If they do offer the alias, then `kokkos` is enough, otherwise the namespace is needed.

As for safety, I don't think it's likely to find a machine with a system installation of kokkos, so I'm ok removing `Kokkos::` (agian, if possible).

components/homme/cmake/HommeMacros.cmake
bartgol: You mean in the `else` branch we should hard code it to OFF (without offering the chance to set it on)? I guess that makes sense.

components/homme/cmake/HommeMacros.cmake
ambrad: I'm not positive, but I think MKL and BLAS/LAPACK usage has always been 1-1 in Homme; there's no need for MKL if there's no need for BLAS/LAPACK. The fast DIRK solver is used in both C++ and F90 in all builds now, and I believe the original DIRK solver was the only user of MKL or BLAS/LAPACK before the fast solver was implemented for all builds. So that's why these dependencies aren't needed anymore. The code has ifdefs -- with no config-level settings -- that toggle the old solver, though. So it's conceivable someone might want to play around with that at some point. But in that case, it's reasonable to expect them to insert the dependencies manually using Extrae_LIBRARY.

oksanaguba:
On anvil, atm_integration is ok
```
./cs.status.20201114_224430_sz6bo1 | grep FAIL
```

oksanaguba:
On anvil, homme test is ok
```
./cs.status.20201114_235115_85dlzl 
```

components/homme/cmake/HommeMacros.cmake
bartgol: Grepping through the folder, I found two calls to DGEMM in `test_src/dcmip2016-supercell.F90`. I'm not sure if that's still used in practical tests, but it seems to be still compiled.

ambrad: Good find. DGESVD is also used in that file, so BLAS/LAPACK or the MKL interface to it are still needed. MKL isn't always available, so on at least some platforms we can expect to have to provide BLAS/LAPACK to get Homme to link.

oksanaguba:
HOMMEBFB is ok too `HOMMEBFB_P24.f19_g16_rx1.A.anvil_intel.C.20201115_001749_ufi95f` .

rljacob:
Can be merged.  No impact on WC

components/homme/cmake/HommeMacros.cmake
ambrad: Just in case this comes up again and we look back at these comments: It just occurred to me that gllfvremap_mod also uses LAPACK, some QR factorization routines in particular.

oksanaguba:
yes, but there is already a fix. Affected calls/functions/modules were wrapped in ifndef with HOMME_BUILD_WITHOUT_PIO, the change will be brought to master eventually.

components/homme/cmake/HommeMacros.cmake
oksanaguba: yes, but there is already a fix. Affected calls/functions/modules were wrapped in ifndef with HOMME_BUILD_WITHOUT_PIO, the change will be brought to master eventually.

PR #3932, opened 2020-11-05
Fixes and enables radiation diagnostics 

### LINKED ISSUE #2575: Radiation diagnostics out of memory crash
I have spent more than 2 weeks trying to figure out a model crash. Finally I was able to identify that the crash can be reproduced with current master without any code modification. I just checked out the latest master, made no change to the code, and enabled 10 radiation diagnostics for a run. The model ran for 1 month, wrote out h0 files, and ran a few more days and crashed with an error message stating “out of memory”. Could somebody please help?

### LINKED ISSUE #3468: Radiation diagnostics are not consistent with climate calculation
The calculation of diagnostic radiation package is not consistent with the climate list. When setting radiation diagnostic 1 identical to the climate diagnostic (rad_diag_1 = rad_climate), the calculated FSNT, FLNT, and AOD all change between the two packages. I attach figures for the difference in FSNT and aerosol AOD. 
FSNT
AOD

singhbalwinder:
Radiation diagnostic calls are enabled where aerosol species mentioned in radiation diagnostics list (`rad_diag_*` in atm_in file) can participate in all the same processes as prognostic radiation calls (mentioned in `rad_climate` list in atm_in). The missing processes for the diagnostic calls were:

Aerosol size adjustment
Aitken<->Accumulation  aerosol transfer

Enabling these calculations ensure that radiation diagnostic call with exactly the same species as radiation climate call, produces BFB diagnostic fields (issue #3468).

Since, radiation diagnostic lists (rad_diag_*) can exclude species or even an entire mode (or modes), I have relied on `rad_cnst_*` calls to get info about mode numbers, number of species in a mode and mode (or species) properties. `rad_cnst_*` calls guarantee that only species/modes present in the `rad_diag_*` lists are accessed. I have tested the following cases:

Excluding all aerosols
Excluding BC from all modes
Excluding SOA from all modes
Excluding Aitken mode
Excluding Accumulation modes
Radiation diagnostic list exactly the same as radiation climate list

For diagnostic lists, “Aerosol size adjustment” process is always ON but “Aiken<->Accumulation transfer” is turned off for diagnostic calls where Aitken or Accumulation mode is absent. I have reworked the mapping so that missing species in aitken and accumulation modes are accounted for.

The modal_aero_calcsize_sub subroutine is heavily refactored where different processes are refactored into their own routines (for readability) and similar calculations are combined together.

This code also fixes the memory leak issue mentioned in #2575. It also fixes another memory leak recently introduced by PR #3885 (thanks for Andrew Bradley for finding this!). This PR also cleans up logic for `clear_rh` variables following Ben Hillman’s suggestions.

Fixes #2575
Fixes #3468
[BFB] (for prognostic radiation calls, the answers will change for the diagnostic calls as this PR fixes a bug identified in issue #3468)

### DIFF
--- a/components/eam/src/chemistry/utils/modal_aero_calcsize.F90
+++ b/components/eam/src/chemistry/utils/modal_aero_calcsize.F90
@@ -542,6 +542,7 @@ subroutine modal_aero_calcsize_sub(state, deltat, pbuf, ptend, do_adjust_in, &
    integer  :: ilist
 
    real(r8), parameter :: close_to_one = 1.0_r8 + 1.0e-15_r8
+   real(r8), parameter :: seconds_in_a_day = 86400.0_r8
 
    real(r8) :: state_q(pcols,pver,pcnst), dqdt(pcols,pver,pcnst)
    real(r8) :: pdel(pcols,pver)   ! pressure thickness of levels
@@ -675,7 +676,7 @@ subroutine modal_aero_calcsize_sub(state, deltat, pbuf, ptend, do_adjust_in, &
    !----------------------------------------------------------------------------
    ! tadj = adjustment time scale for number, surface when they are prognosed
    !----------------------------------------------------------------------------
-   tadj    = max( 86400.0_r8, deltat )
+   tadj    = max( seconds_in_a_day, deltat )
    tadjinv = 1.0_r8/(tadj*close_to_one)
    fracadj = max( 0.0_r8, min( 1.0_r8, deltat*tadjinv ) )
 
@@ -842,7 +843,7 @@ end subroutine modal_aero_calcsize_sub
 !---------------------------------------------------------------------------------------------
 #ifdef MODAL_AERO
 subroutine set_initial_sz_and_volumes(list_idx, top_lev, pver, ncol, imode, & !input
-     dgncur, v2ncur, dryvol )                                       !output
+     dgncur, v2ncur, dryvol )                                                 !output
 
   !-----------------------------------------------------------------------------
   !Purpose: Set initial defaults for the dry diameter, volume to num
@@ -1144,6 +1145,7 @@ real(r8), intent(out) :: dgnyy
 !local
 real(r8), parameter :: relax_factor = 27.0_r8 !relax_factor=3**3=27,
                                               !i.e. dgnumlo_relaxed = dgnumlo/3 and dgnumhi_relaxed = dgnumhi*3
+real(r8), parameter :: szadj_block_fac = 1.0e6_r8
 real(r8) :: dgnumlo, dgnumhi, sigmag, cmn_factor
 
 
@@ -1166,11 +1168,11 @@ dgnyy   = dgnumlo
 if ( do_aitacc_transfer ) then
    !for n=nait, divide v2nxx by 1.0e6 to effectively turn off the
    !         adjustment when number is too small (size is too big)
-   if (imode == nait) v2nxx = v2nxx/1.0e6_r8
+   if (imode == nait) v2nxx = v2nxx/szadj_block_fac
 
    !for n=nacc, multiply v2nyy by 1.0e6 to effectively turn off the
    !         adjustment when number is too big (size is too small)
-   if (imode == nacc) v2nyy = v2nyy*1.0e6_r8
+   if (imode == nacc) v2nyy = v2nyy*szadj_block_fac
 
    !Also change the v2nyyrl/v2nxxrl so that
    !the interstitial<-->activated adjustment is turned off
@@ -1419,7 +1421,7 @@ subroutine  aitken_accum_exchange(ncol, lchnk, list_idx, update_mmr, tadjinv, &
   real(r8) :: num_a, drv_a, num_c, drv_c
   real(r8) :: num_a_acc, num_c_acc
   real(r8) :: drv_a_acc, drv_c_acc
-  real(r8) :: v2nzz, pdel_fac, num_t, drv_t
+  real(r8) :: v2n_geomean, pdel_fac, num_t, drv_t
   real(r8) :: drv_a_noxf, drv_c_noxf, drv_t_noxf, dummwdens
   real(r8) :: num_t0, num_t_noxf
   real(r8) :: duma, cmn_factor, dgnum, sigmag, voltonumb_ait, voltonumb_acc
@@ -1461,21 +1463,23 @@ subroutine  aitken_accum_exchange(ncol, lchnk, list_idx, update_mmr, tadjinv, &
 
   ! set dotend() for species that will be transferred
   ! for both mass and number
-  do iq = 1, nspecfrm_csizxf(list_idx)
-     lsfrm = lspecfrma_csizxf(iq,list_idx)
-     lstoo = lspectooa_csizxf(iq,list_idx)
-     if ((lsfrm > 0) .and. (lstoo > 0)) then
-        if(update_mmr)dotend(lsfrm) = .true.
-        if(update_mmr)dotend(lstoo) = .true.
-     end if
-     !for cloud borne aerosols
-     lsfrm = lspecfrmc_csizxf(iq,list_idx)
-     lstoo = lspectooc_csizxf(iq,list_idx)
-     if ((lsfrm > 0) .and. (lstoo > 0)) then
-        if(update_mmr)dotendqqcw(lsfrm) = .true.
-        if(update_mmr)dotendqqcw(lstoo) = .true.
-     end if
-  end do
+  if(update_mmr) then
+     do iq = 1, nspecfrm_csizxf(list_idx)
+        lsfrm = lspecfrma_csizxf(iq,list_idx)
+        lstoo = lspectooa_csizxf(iq,list_idx)
+        if ((lsfrm > 0) .and. (lstoo > 0)) then
+           dotend(lsfrm) = .true.
+           dotend(lstoo) = .true.
+        end if
+        !for cloud borne aerosols
+        lsfrm = lspecfrmc_csizxf(iq,list_idx)
+        lstoo = lspectooc_csizxf(iq,list_idx)
+        if ((lsfrm > 0) .and. (lstoo > 0)) then
+           dotendqqcw(lsfrm) = .true.
+           dotendqqcw(lstoo) = .true.
+        end if
+     end do
+  endif
 
   !------------------------------------------------------------------------
   ! Identify accum species cannot be transferred to aitken mode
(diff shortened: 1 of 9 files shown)


components/eam/src/chemistry/utils/modal_aero_calcsize.F90
```diff
@@ -1484,9 +2139,9 @@ subroutine modal_aero_calcsize_diag(state, pbuf, list_idx_in, dgnum_m)
          call endrun('modal_aero_calcsize_diag called for'// &
                      'diagnostic list but dgnum_m pointer not present')
       end if
-      allocate(dgnum_m(pcols,pver,nmodes), stat=stat)
-      if (stat > 0) then
-         call endrun('modal_aero_calcsize_diag: allocation FAILURE: dgnum_m')
+      if(.not. associated(dgnum_m)) then
```
singhbalwinder, line 2142: I am reworking this logic where an error is issued if this variable is not allocated already.

ambrad: Great. You might also check whether you can make dgnum_m and other arrays `allocatable` rather than `pointer`. That would provide code-level protection against future memory leaks.

singhbalwinder: That's a great point. I have reworked the logic to use allocatable arrays in calcsize and wateruptake. I am no longer doing any allocation within the calsize_diag routine.

singhbalwinder:
Tagging @golaz and @rljacob as this PR is for V2.

This PR is now ready for reviews. I have cleaned up some logics and got rid of some pointer arrays in modal_aero_calcsize.F90.

rljacob:
The Travis CI build is failing and that should be looked at.

singhbalwinder:
I can build/run it just fine on Compy. Is there a way to know the exact error message from Travis CI? I tried clicking the "Details" but it just mentions the log file:
```
ERROR: BUILD FAIL: build e3sm failed, cat /home/travis/projects/e3sm/scratch/master.A_WCYCL1850.ne4_oQU240.baseline/bld/e3sm.bldlog.201106-225355
```

I will try the GCC compiler as well....may be it is a compiler issue.

hwangacme (approved):

kaizhangpnl:
@singhbalwinder The code is BFB only for the simulations without the radiative forcing diagnostics, right?

components/eam/src/chemistry/utils/modal_aero_calcsize.F90
```diff
@@ -2,75 +2,118 @@ module modal_aero_calcsize
...
+!      and rad_diag_3 (e.g. skipping rad_diag_2). Therefore the arrays should
+!      have a length of N_DIAG (unless we define another array which maps info
+!      such as to include info about the missing diagnostics)
+!---------------------------------------------------------------------------------
+integer, parameter :: npair_csizxf = N_DIAG           !total number of possible diagnostic calls
+
+logical :: do_adjust_allowed                          !flag to turn on/off  aerosol size adjustment process
+
+!flag to turn on/off  aerosol aitken<->accumulation transfer process
+logical :: do_aitacc_transfer_allowed(0:npair_csizxf) ! This is an array as it can be different for each radiatio diagnostic call
```
kaizhangpnl, line 77: typo "radiatio"

singhbalwinder:
Yes, that is correct. Code is BFB only for the prognostic calculations. Answers will change for the radiation diagnostic calculations as there was a bug in the current implementation (see issue #3468).

kaizhangpnl:
Thanks. So maybe state it explicitly in the commit message?

kaizhangpnl:
@singhbalwinder How did you verify/evaluate the refactored code? Do you have a pair of PD&PI simulations with the radiative forcing diagnostics switched on? If so, I can analyse the aerosol forcing fields and see if the results are consistent with our previous estimate.

singhbalwinder:
The only test I did was the one mentioned in issue #3468. This test verifies that if rad_diag list is exactly the same as rad_climte, we get BFB answers between diagnostics (e.g. fsnt_d1) and the prognostic variables. This test does exercise all processes of the calcsize_sub. I haven't run PD/PI simulations but I think its a great idea to run these simulations. I will talk to you offline about the namelist options (rad_diags) to use for these simulations.

kaizhangpnl (approved):
Visual inspection only. Will work with Balwinder on the validation.

kaizhangpnl:
Effective aerosol forcing diagnostics (with decomposition) look reasonable: 

https://compy-dtn.pnl.gov/zhan524/E3SM/v2/alpha/erf/nudging_calcsz_refac/

brhillman (approved):

wlin7:
Merged to next.

wlin7:
Re-merged to next after commit 6668bde.

dqwu:
@singhbalwinder 
On anlworkstation, two e3sm_developer tests failed with latest next branch, see
https://my.cdash.org/viewTest.php?onlyfailed&buildid=1927425

They might also fail on mappy's e3sm_developer_next_gnu build later.

I have confirmed that the failures are caused by this PR.
One of the tests uses debug mode (SMS_D_Ln5.ne4_ne4.FC5AV1C-L) and the stack trace is:
```
[0] At line 456 of file components/eam/src/chemistry/utils/modal_aero_wateruptake.F90
[0] Fortran runtime error: Index '1' of dimension 3 of array 'wetdens' below lower bound of 465579216
[0] 
[0] Error termination. Backtrace:
[0] #0  0x1a0a51f in __modal_aero_wateruptake_MOD_modal_aero_wateruptake_dr
[0] 	at components/eam/src/chemistry/utils/modal_aero_wateruptake.F90:456
[0] #1  0x1160a1d in __modal_aer_opt_MOD_modal_aero_sw
[0] 	at components/eam/src/physics/cam/modal_aer_opt.F90:633
[0] #2  0xf6d8b9 in __aer_rad_props_MOD_aer_rad_props_sw
[0] 	at components/eam/src/physics/cam/aer_rad_props.F90:269
[0] #3  0xb786cf in __radiation_MOD_radiation_tend
[0] 	at components/eam/src/physics/rrtmg/radiation.F90:1281
[0] #4  0x12d75aa in tphysbc
[0] 	at components/eam/src/physics/cam/physpkg.F90:2710
[0] #5  0x12e83f0 in __physpkg_MOD_phys_run1
[0] 	at components/eam/src/physics/cam/physpkg.F90:1052
[0] #6  0x5ee849 in __cam_comp_MOD_cam_run1
[0] 	at components/eam/src/control/cam_comp.F90:245
[0] #7  0x5de609 in __atm_comp_mct_MOD_atm_init_mct
[0] 	at components/eam/src/cpl/atm_comp_mct.F90:396
[0] #8  0x434756 in __component_mod_MOD_component_init_cc
[0] 	at cime/src/drivers/mct/main/component_mod.F90:248
[0] #9  0x425e64 in __cime_comp_mod_MOD_cime_init
[0] 	at cime/src/drivers/mct/main/cime_comp_mod.F90:2261
[0] #10  0x42d592 in cime_driver
[0] 	at cime/src/drivers/mct/main/cime_driver.F90:122
[0] #11  0x42d6d2 in main
[0] 	at cime/src/drivers/mct/main/cime_driver.F90:23
```

singhbalwinder:
Thanks @dqwu ! I was able to reproduce it using GNU on Cori. This error doesn't show up using Intel compiler. I have pushed the fix.

wlin7:
Re-merged to next after commit a3a71e2 to fix gnu runtime issue (referencing null pointer). Thanks @dqwu and @singhbalwinder for the test and the quick fix.

rljacob:
How do we know which cases have "radiation diagnostics" and so will show diffs?    Does that refer to cosp?

singhbalwinder:
No, radiation diagnostics are extra calculations which are invoked only when we add extra fields (e.g. `rad_diag_1` etc.) in the atmosphere namelist file (atm_in). These fields look like the following:
```
rad_diag_1            = 'A:Q:H2O', 'N:O2:O2', 'N:CO2:CO2','A:O3:O3', 'N:N2O:N2O', 'N:CH4:CH4','N:CFC11:CFC11', 'N:CFC12:CFC12'
rad_diag_2            = 'A:Q:H2O', 'N:O2:O2', 'N:CO2:CO2','A:O3:O3', 'N:N2O:N2O', 'N:CH4:CH4','N:CFC11:CFC11', 'N:CFC12:CFC12', 'M:mam4_mode1:/compyfs/inputdata/atm/cam/physprops/mam4_mode1_rrtmg_aeronetdust_c141106.nc','M:mam4_mode2:/compyfs/inputdata/atm/cam/physprops/mam4_mode2_rrtmg_c130628.nc', 'M:mam4_mode3:/compyfs/inputdata/atm/cam/physprops/mam4_mode3_rrtmg_aeronetdust_c141106.nc', 'M:mam4_mode4:/compyfs/inputdata/atm/cam/physprops/mam4_mode4_rrtmg_c130628.nc'  

rad_diag_3            = 'A:Q:H2O', 'N:O2:O2', 'N:CO2:CO2','A:O3:O3', 'N:N2O:N2O', 'N:CH4:CH4','N:CFC11:CFC11', 'N:CFC12:CFC12', 'M:mam4_mode1_no_bc:/compyfs/inputdata/atm/cam/physprops/mam4_mode1_rrtmg_aeronetdust_c141106.nc','M:mam4_mode2:/compyfs/inputdata/atm/cam/physprops/mam4_mode2_rrtmg_c130628.nc', 'M:mam4_mode3_no_bc:/compyfs/inputdata/atm/cam/physprops/mam4_mode3_rrtmg_aeronetdust_c141106.nc', 'M:mam4_mode4_no_bc:/compyfs/inputdata/atm/cam/physprops/mam4_mode4_rrtmg_c130628.nc'
```
We can specify up to 10 diagnostic (from `rad_diag_1 to rad_diag_10`) in atm_in.  For each diagnostic list (`rad_diag_*`), the model repeats the radiation calculations based on species present in that diagnostic list.

Currently, we do not have any tests for these in our testing suite but it will be a good idea to add one. This PR resolves issue #3468, which shows that this PR changes answers for the diagnostic calls.

wlin7:
Re-merged to next after commit b870ebc that fixed non-BFB and runtime issue with GNU build.

Previous gnu fix in commit a3a71e2 broke BFB. next was reset to state prior to this PR. This re-merge is now a clean commit to next. Much thanks to @singhbalwinder for quickly resolving the issues.

dqwu:
@singhbalwinder I also ran some e3sm_integration tests on anlworkstation with GNU compiler.
ERS.ne11_oQU240.A_WCYCL1850 failed, and TestStatus shows:
`FAIL ERS.ne11_oQU240.A_WCYCL1850.anlworkstation_gnu COMPARE_base_rest`

wlin7:
This PR is still causing non-BFB in next. Next is reset to the state before this PR and forced push to remote.

rljacob:
Notes from IG telecon:  working on GNU.  Should be done soon.

rljacob:
Notes from IG telecon:  this is now working but needs to wait for another PR #3998  to change optimization.

wlin7:
Re-merged to next after #3998 and updates in this PR.

wlin7:
Merged to master.

PR #3943, opened 2020-11-14
Make SCM compatible with theta-l dycore (+ bug fixes)

bogensch:
This PR makes the SCM functional with the theta-l dycore.  The theta-l dycore will now compute the large scale vertical advection for the SCM.  SCM computes column only vertical advection (done through remap routine via updated floating levels given the prescribed large scale vertical velocity) and NOT horizontal advection.

Several SCM cases have been run (ARM97, BOMEX, DYCOMS, GATE) to verify that the SCM with theta-l dycore produces scientifically credible answers that are similar to the preqx version of the SCM.   

These modifications have been verified to work for BOTH preqx and theta-l dycores.

This PR also includes two SCM bug fixes:
1. Make sure that the updated surface fluxes are read to cam_in arrays (in cam_comp.F90).  This will prevent surface fluxes from being “seen” by CLUBB at one time step offset (reported by Mikael Witte).
2. Fix in scm_setfield to be sure that the model is updating surface pressure when replay is not used (else could cause crashes when SCM timestep is smaller than IOP forcing).

This PR is B4B EXCEPT for SCM (i.e. test SMS_R_Ld5.ne4_ne4.FSCM5A97).  All other e3sm_developer tests pass.


components/homme/src/share/prim_driver_base.F90
```diff
@@ -1331,8 +1344,15 @@ subroutine prim_step_flexible(hybrid, elem, nets, nete, dt, tl, hvcoord, compute
           end if
        end if
 
-       call prim_advance_exp(elem, deriv1, hvcoord, hybrid, dt, tl, nets, nete, &
-            compute_diagnostics_it)
+       if (.not. single_column) then
```
oksanaguba, line 1347: my personal preference would be to have ifdefs here, since these loops will be evaluated every timestep (though maybe for negligible cost), but i think other people prefer if-statements to ifdefs.

oksanaguba (approved):

components/homme/src/share/prim_driver_base.F90
```diff
@@ -1273,11 +1275,22 @@ subroutine prim_step_flexible(hybrid, elem, nets, nete, dt, tl, hvcoord, compute
...
 
     real(kind=real_kind) :: dt_q, dt_remap, dp(np,np,nlev)
-    integer :: ie, q, k, n, n0_qdp, np1_qdp
+    integer :: ie, q, k, n, n0_qdp, np1_qdp, nets_in, nete_in
     logical :: compute_diagnostics_it, apply_forcing
 
+    logical :: scm_update_T
+
+    if (single_column) then
+      nets_in=1
```
ambrad, line 1287: As a style point, you might consider using nets_in, nete_in as the arguments to the function and make nets, nete the local variables. That way nets, nete get used throughout the definition rather than nets_in, nete_in.

components/homme/src/share/prim_driver_base.F90
```diff
@@ -1356,16 +1376,22 @@ subroutine prim_step_flexible(hybrid, elem, nets, nete, dt, tl, hvcoord, compute
...
           end if
        end if
        ! defer final timelevel update until after Q update.
     enddo
     call t_stopf("prim_step_dyn")
 
-    if (qsize > 0) then
+    ! Update tracers based on LS vertical velocity
+    scm_update_T = .false.
+    call prim_step_scm(elem, nets_in, nete_in, dt, tl, hvcoord, scm_update_T)
```
ambrad, line 1392: This line confuses me. It appears like prim_step_scm will always get here, i.e., even when single_column is false.

components/homme/src/share/prim_driver_base.F90
```diff
@@ -1331,8 +1344,15 @@ subroutine prim_step_flexible(hybrid, elem, nets, nete, dt, tl, hvcoord, compute
...
-       call prim_advance_exp(elem, deriv1, hvcoord, hybrid, dt, tl, nets, nete, &
-            compute_diagnostics_it)
+       if (.not. single_column) then
+         call prim_advance_exp(elem, deriv1, hvcoord, hybrid, dt, tl, nets, nete, &
+              compute_diagnostics_it)
+       else
+         ! Update floating temperature based prescribed LS vertical velocity,
+         !  but don't update tracers at this point
+         scm_update_T = .true.
+         call prim_step_scm(elem, nets_in, nete_in, dt, tl, hvcoord, scm_update_T)
```
ambrad, line 1354: Here you're in a `single_column` block, so you could just write 1, 1 if you wanted to instead of nets_in, nete_in.

components/homme/src/share/prim_driver_base.F90
```diff
@@ -1356,16 +1376,22 @@ subroutine prim_step_flexible(hybrid, elem, nets, nete, dt, tl, hvcoord, compute
                 end do
              else
                 ! Set np1_qdp to -1 to remap dynamics variables but
-                ! not tracers.
-                call vertical_remap(hybrid, elem, hvcoord, dt_remap, tl%np1, -1, nets, nete)
+                ! not tracers
+                call vertical_remap(hybrid, elem, hvcoord, dt_remap, tl%np1, &
```
ambrad, line 1380: Here you're _not_ in a `single_column` block but you use nets_in, nete_in. This combined with the condition on the Prim_Advec_Tracers_remap call below implies the intention is something like: dynamics are active, but tracer transport is fully prescribed. In that case, you don't need to even modify subroutine prim_step_flexible _at all_. Its whole purpose is to support long SL tracer transport time steps. But if you're not using large-scale advection, then you might as well set dt_tracer_factor=1 <= dt_remap_factor and never even call this function. If you do that, I recommend you check that if single_column is true, then dt_tracer_factor <= dt_remap_factor or else abort.

ambrad (changes requested):
I'd like us to resolve the issue of whether prim_step_flexible should even be modified before this is merged. This includes discussing why you'd ever want dt_tracer_factor > dt_remap_factor or even dt_tracer_factor > 1 in a SCM run. Thanks.

bogensch:
Thanks @ambrad for the comments and feedback, much appreciated.  Just to clarify a few things before I modify:

There are definitely some SCM cases where the LS vertical advection is prescribed.  If that is the case, then there are conditionals outside of the HOMME directory (in the SE interface) to make sure that the SCM never goes into the dycore at all.

However, some SCM cases (probably half in our library) only prescribes forcing for the horizontal LS advection.  Thus, the vertical advection needs to be computed by the dycore.  In these cases, we do go into the dycore with the intention of computing LS vertical advection by updating the floating levels using the prescribed LS vertical velocity, then remapping.  

The E3SM SCM philosophy is that we want all needed column computations to be done as consistently as possible as the full model would.  The SCM by default uses the same time step settings as an E3SM ne30 run to achieve consistency.  Thus, in this case the tracer remap timestep is larger than the dynamics timestep, consistent with the full model.  Therefore in order for the SCM to be consistent with the full model for the computation of LS vertical advection, it seems to me we need to go into the prim_step_flexible routine.  Does that make sense?

ambrad:
> However, some SCM cases (probably half in our library) only prescribes forcing for the horizontal LS advection. Thus, the vertical advection needs to be computed by the dycore.

How are the tracers moved in the horizontal? Or are they uniform everywhere, so they effectively have no horizontal motion?

> Thus, in this case the tracer remap timestep is larger than the dynamics timestep, consistent with the full model. Therefore in order for the SCM to be consistent with the full model for the computation of LS vertical advection, it seems to me we need to go into the prim_step_flexible routine. Does that make sense?

I must confess that it does not make sense to me. I agree roughly on the intent. But there are enough details different in the SCM prescribed dynamics w.r.t. the full dynamics that it seems wiser to choose a simple approach to advancing the SCM.

For example, in set_prescribed_scm, there is:
```F90
    do k=1,nlev
      eta_dot_dpdn(:,:,k)=elem(1)%derived%omega_p(1,1,k)   
    enddo  
    eta_dot_dpdn(:,:,nlev+1) = eta_dot_dpdn(:,:,nlev) 
    
    do k=1,nlev
      elem(1)%state%dp3d(:,:,k,np1) = elem(1)%state%dp3d(:,:,k,n0) &
        + dt*(eta_dot_dpdn(:,:,k+1) - eta_dot_dpdn(:,:,k)) 
#ifdef MODEL_THETA_L
      ! If updating Q and using the theta-l dycore the dp_star (floating
      !   levels) are stored in divdp.  Thus feed in the latest predicted
      !   floating level here.
      if (update_Q) then
        elem(1)%derived%divdp(:,:,k) =  elem(1)%state%dp3d(:,:,k,np1)
      endif
#endif
    enddo
```
But if the dynamics has moved back to the reference levels already (as happens in the _flexible variant in the call to vertical_remap with the -1 argument), then the fluxes are being applied at the wrong elevations. sl_advection's sl_vertically_remap_tracers takes care of this using some calculations that require a divergence and so fully 3D data. Restriction of some calculations to the range nets_in:nete_in means such data are not available, e.g.:
```F90
                ! Set np1_qdp to -1 to remap dynamics variables but
                ! not tracers
                call vertical_remap(hybrid, elem, hvcoord, dt_remap, tl%np1, &
		                    -1, nets_in, nete_in)
```
You could probably fix that and then call sl_advection's calc_vertically_lagrangian_levels, probably with some preparatory work. But then you need to make vstar reflect v at the tracer time start. Right now, in prim_step_scm, this seems to run at every dynamics time step, invalidating vstar:
```F90
      if (transport_alg > 0) then
        elem(ie)%derived%vstar=elem(ie)%state%v(:,:,:,:,tl%n0)
      end if
```
Likely there are other details I'm not seeing.

If you just set dt_tracer_factor=1 whenever you run SCM, you can use code path that you've been using for years. To me, that seems the best approach given the differences in details in the SCM and full dynamics.

components/homme/src/share/prim_driver_base.F90
ambrad: Did the test `SMS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L.cam-thetahy_sl_pg2` run when you ran `e3sm_developer`? It would alarm me if that test passes; it seems this unguarded call to prim_step_scm must be invalidating some data. Maybe the test doesn't run for enough time steps.

bogensch:
Thanks @ambrad, that is very clear.  I see the point you are getting at and how it could conflict with my naive interpretation of how this section of the code works.  However, I do have a few specific questions for you to shoot down before I concede. 

I guess the main point of consistency I’m worried about between the full model and SCM for theta-l is that in theta-l the tracer remap timestep dt_q is now 1800 s for ne30.  I have a branch where I revert the tracer transport to being called with the same time step as the dynamics (300 s) and this difference does not have a non-negligible impact for cases where vertical advection is large.  Thus, I am worried if reverting the tracer time step back to 300 s for theta-l, that this would create a numerical inconsistency between full model and SCM.  

Though, as you state, I understand that sl_vertically_remap_tracers uses fully 3D data.  However, it looks to my naive eye that in a case where there would be no horizontal advection that this subroutine call would reduce to “vertical_remap” since they both use the remap1 routine (assuming that divdp contains the floating levels and dp3d the reference levels, as I have done)?  But it sounds like what you are saying is that for the scenario i’ve outlined above, it is NOT appropriate to be remapping tracers with a larger time step than dynamics (presumably due to missing some key processes/computations happening in Prim_Advec_Tracers_remap)?  

I guess, another way to phrase my question… If in the SCM I were to replace the “sl_vertically_remap_tracers” call with a call to a modified vertical_remap (where only tracers were remapped and not dynamics) in an effort to retain the “dt_q” remap time step…. this would NOT be appropriate?

ambrad:
> in the SCM I were to replace the “sl_vertically_remap_tracers” call with a call to a modified vertical_remap (where only tracers were remapped and not dynamics) in an effort to retain the “dt_q” remap time step

You could remap both dynamics and tracers at dt_q. That would use the non-_flexible prim_step variant that you've always used, now just with dt_dyn = dt_q 6x longer than previously.

bogensch:
Thanks @ambrad .  I'll be pushing commits, but I'll let you know when it's ready for re-review.

components/eam/bld/namelist_files/use_cases/scam_generic.xml
```diff
@@ -6,4 +6,10 @@
 <ncdata             >atm/cam/inic/homme/cami_mam3_Linoz_ne4np4_L72_c160909.nc</ncdata>
 <soag_ext_file      >atm/cam/chem/trop_mozart_aero/emis/aces4bgc_nvsoa_soag_elev_2000_c160427.nc</soag_ext_file>
 
+<!-- SCM is configured with ne4 grid, but we want settings for an ne30 simulation -->
+<nu dyn_target="theta-l" hgrid="ne4np4"  >  1.0e15    </nu>
```
ambrad, line 10: To help my understanding: Is nu actually used anywhere in a SCM simulation?

ambrad (approved):
I'm approving based on the criterion that this PR no longer touches code relevant to usual simulations. Thanks for the helpful discussion. I left one new question just to help me understand better which of the dynamics settings are actually meaningful in a SCM simulation.

components/eam/bld/namelist_files/use_cases/scam_generic.xml
bogensch: good find, yeah this is not needed and will be removed

AaronDonahue:
@wlin7 and @oksanaguba , are you comfortable with me merging this branch given @ambrad 's approval?

oksanaguba:
@AaronDonahue yes

AaronDonahue:
Ok great, I'm going to go ahead and begin integrating.

AaronDonahue:
This has been merged to master, deleting the branch now.

PR #3948, opened 2020-11-18
Update mpas-source: updated sea ice with alignment of atmos coupling

akturner:
Updated sea ice component to implement alignment of atmospheric coupling with cime

Non-BFB


proteanplanet (approved):
This has been tested in fully coupled E3SM simulations and is congruent with the CESM flux method currently in CIME.   It is not BFB, but produces climatologically similar results to the original scheme in MPAS-SeaIce, and is a preferred algorithm.

vanroekel (approved):
approving based on @proteanplanet's extensive testing.  I am approving integration of this PR for water cycle v2.  It is nonBFB, but it is needed for v2 simulations.

jonbob:
Merged to next

jonbob:

Issue #3934, opened 2020-11-07
memleak in ERP_Ld3.ne30_oECv3_ICG.A_WCYCL1850S.sandiatoss3_intel 


rljacob:
After https://github.com/ESMCI/cime/pull/3736 was merged to master in https://github.com/E3SM-Project/E3SM/pull/3891, the ERP_Ld3.ne30_oECv3_ICG.A_WCYCL1850S.sandiatoss3_intel  test started to fail with a memleak:

`2020-11-05 21:29:05: memleak detected, memory went from 3525.930000 to 4290.060000 in 1 days`

If this is real what caused it?   Also if its real, why just that case and not other coupled runs with that resolution?

rljacob:
We thought the memleak might be from PR #3885  which is fixed in PR #3932 but the code causing the memleak there isn't used in a regular coupled case.

singhbalwinder:
Yeah, I confirmed running a case that memleak fixed in #3932 is not used in FC5AV1C-L compset (which is used in WCYCL compset). So #3932 won't fix this issue.

Here are two more cases where I saw ERP failing with a memory leak in the past and we can still reproduce it:

@guangxinglin saw ERP failing with a memleak on Cori with maint-1.0 branch. I was able to reproduce it using an F case:
```
PASS ERP.ne4_ne4.FC5AV1C-L.cori-knl_intel CREATE_NEWCASE
PASS ERP.ne4_ne4.FC5AV1C-L.cori-knl_intel XML
PASS ERP.ne4_ne4.FC5AV1C-L.cori-knl_intel SETUP
PASS ERP.ne4_ne4.FC5AV1C-L.cori-knl_intel SHAREDLIB_BUILD time=645
PASS ERP.ne4_ne4.FC5AV1C-L.cori-knl_intel MODEL_BUILD time=1385
PASS ERP.ne4_ne4.FC5AV1C-L.cori-knl_intel SUBMIT
PASS ERP.ne4_ne4.FC5AV1C-L.cori-knl_intel RUN time=531
PASS ERP.ne4_ne4.FC5AV1C-L.cori-knl_intel COMPARE_base_rest
FAIL ERP.ne4_ne4.FC5AV1C-L.cori-knl_intel MEMLEAK memleak detected, memory went from 3258.000000 to 3740.000000 in 9 days
PASS ERP.ne4_ne4.FC5AV1C-L.cori-knl_intel SHORT_TERM_ARCHIVER
```
[...]

PR #3951, opened 2020-11-19
Set PIO_ROOT=1 for ne30-coupled cases on sandiatoss3

amametjanov:
Set PIO_ROOT=1 for ne30-coupled cases on sandiatoss3 to avoid MEMLEAK test issues.

Fixes E3SM-Project/E3SM#3934

[NML] - due to pio_root changes

### DIFF
--- a/cime_config/machines/config_pio.xml
+++ b/cime_config/machines/config_pio.xml
@@ -28,15 +28,13 @@
   <entry id="PIO_STRIDE">
     <values>
       <value>$MAX_MPITASKS_PER_NODE</value>
-      <value mach="yellowstone" grid="a%ne120.+oi%gx1">60</value>
-      <value mach="mira|cetus">128</value>
-      <value mach="mira" grid="a%ne120" compset=".*CAM5.+MPASO.*">-99</value>
     </values>
   </entry>
 
   <entry id="PIO_ROOT">
     <values>
       <value>0</value>
+      <value mach="sandiatoss3" grid="a%ne30" compset=".*EAM.+MPAS.*">1</value>
     </values>
   </entry>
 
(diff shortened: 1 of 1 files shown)


amametjanov:
@jgfouca do you recall why this was set?
```
  <entry id="PIO_BUFFER_SIZE_LIMIT">
    <values>
      <value MACH="sandiatoss3">1</value>
      <value MACH="ghost">1</value>
    </values>
  </entry>
```
Can you check if removing just this will fix the memleak issue (which happens only on sandiatoss3)?

jgfouca:
Nice job, @amametjanov . I apologize, I have no idea why that buffer size was set. Since ghost was included, it was likely some other Sandia person. @oksanaguba ?

@rljacob , can i go ahead and integrate this since it's specific to sandiatoss3 only?

oksanaguba:
@jgfouca I have no idea about this buffer size.

rljacob:
Yes go ahead.

amametjanov:
If nobody knows about `PIO_BUFFER_SIZE_LIMIT=1`, might as well remove it for consistency. All other machines default it to `-1`.

jgfouca:
Merged to next.

jayeshkrishna:
I second what @amametjanov mentioned above, PIO_BUFFER_SIZE_LIMIT is typically set to -1 to indicate that CIME/Scorpio can decide the appropriate value for the buffer size limit (this should be the default case). Its possible that its a bug to set it to 1 (the intention might have been to set it to -1), I cannot think of a reason why the limit needs to be set to such a low value.

rljacob:
ERP_Ld3.ne30_oECv3_ICG.A_WCYCL1850S.sandiatoss3_intel is now passing!

rljacob:
@jgfouca did you update the PIO_BUFFER_SIZE_LIMIT as mentioned above?

jgfouca:
@rljacob , no. Do you want me to open a new PR?

amametjanov:
@jgfouca , E3SM-Project/E3SM#3955 should do it.

jgfouca:
Thanks, @amametjanov !

PR #3955, opened 2020-11-20
Move PIO_ROOT=1 mod to ERP testmod

amametjanov:
Move PIO_ROOT=1 mod to ERP testmod. Also, remove
PIO_BUFFER_SIZE_LIMIT=1 for sandiatoss3 and ghost machines: default
to -1 so that the amount of buffering/caching is determined within Scorpio.
Also, increase MOSART rh0 restart file name length.

Fixes E3SM-Project/E3SM#3934, Fixes E3SM-Project/E3SM#3959

[BFB] - new test baseline


rljacob:
Description should also mention the change to PIO_BUFFER_SIZE

cime_config/testmods_dirs/allactive/pioroot1/shell_commands
```diff
@@ -0,0 +1 @@
+./xmlchange PIO_ROOT=1
```
rljacob, line 1: You should wrap this with an "if mach=sandiatoss3" or it will be set to 1 on all machines that run this test.

rljacob (changes requested):

cime_config/testmods_dirs/allactive/pioroot1/shell_commands
amametjanov: ERP had a FAIL on anvil too, on Nov 20 with hash 15bf0dae8 on next (that had PIO_ROOT=1 just for sandiatoss3) -- https://my.cdash.org/viewTest.php?onlyfailed&buildid=1921288:
```
2020-11-20 03:36:36: memleak detected, memory went from 3470.250000 to 4019.950000 in 1 days
```
a 15.84% mem-usage increase. So the ERP testmod appears to be needed for all machines.

rljacob: oh ok.

rljacob (approved):

rljacob:
I'm not sure separating issues by commas fill close each of them in the "Fixes" line.

PR #3957, opened 2020-11-24
Fix energy leak in GWD

oksanaguba:
Add an energy brute force fix to the GWD. Off by default.

The fix works only if orographic GW are on.

[bfb]


golaz:
@oksanaguba : thanks for starting to work on this.

@rljacob : this is something we would like to consider for v2 when it is ready.

oksanaguba:
Here are 10 years of the run with this PR:
Screen Shot 2020-11-24 at 8 40 12 PM

oksanaguba:
This now is only waiting for more years to complete, along with a baseline run.

I will also put in a small test for this fix.

oksanaguba:
Results for 20 years with the PR and 18 years with the baseline are at https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/2064941070/GWD+PR+1 .

singhbalwinder:
Thanks @oksanaguba.  The changes look good to me. I have assigned reviewers and I will merge it after it is reviewed.

singhbalwinder (approved):

components/eam/src/physics/cam/gw_drag.F90
```diff
@@ -886,14 +895,28 @@ subroutine gw_tend(state, sgh, pbuf, dt, ptend, cam_in)
...
+           ptend%s(:ncol,k) = ptend%s(:ncol,k) + ttgw(:,k) &
+                -(ptend%u(:ncol,k) * (u(:,k) + ptend%u(:ncol,k)*0.5_r8*dt) &
+                +ptend%v(:ncol,k) * (v(:,k) + ptend%v(:ncol,k)*0.5_r8*dt))
+           ttgw(:,k) = ttgw(:,k) &
+                -(ptend%u(:ncol,k) * (u(:,k) + ptend%u(:ncol,k)*0.5_r8*dt) &
+                +ptend%v(:ncol,k) * (v(:,k) + ptend%v(:ncol,k)*0.5_r8*dt))
+           ttgw(:,k) = ttgw(:,k) / cpairv(:ncol, k, lchnk)
+        end do
+     else
+        !new code, brute force to conserve energy in each level locally
```
wlin7, line 912: Hi @oksanaguba , compared to original, with fix1, it seems to amount to throwing away temperature tendency ttgw (which is of diffusion of dry static energy and conversion of tke)  from all gw sources. Is this interpretation correct?

components/eam/src/physics/cam/check_energy.F90
```diff
@@ -21,6 +21,9 @@ module check_energy
 !                  MG2 (now includes prognostic rain and snow)    
 !               2. Better printout information for energy and water conservation check 
 !               3. Additional water conservation check utilities 
+!
```
wlin7, line 24: @oksanaguba ,These added comments and similar ones in clubbb_intr.F90 and physics_types.F90 seem to be intended when incorporating your other proposed energy fixes as well. Are we planning to add those at a later time? If not, I would think not to include the comments in these three files.

components/eam/src/physics/cam/gw_drag.F90
oksanaguba: i am not sure what you mean. the change in this pr overwrites ttgw at the end of gw_tend call but at that point ttgw is not actually used anywhere below (as it is not used in the original). even for diagnostics, ptend%s is used instead:
  call outfld ('TTGW', ptend%s/cpairv(:,:,lchnk),  pcols, lchnk)
also, from what i saw ttgw is always ptend%s/c_p .

assuming ptend%u and ptend%v are given, to conserve energy one needs to modify ptend%s (which modifies temperature later, in the interface). there are a few ways to conserve energy, but if ttgw=ptend%s carries some other meaning/quantity that is needed, please, let me know. thanks.

components/eam/src/physics/cam/check_energy.F90
oksanaguba: those comments are for something that was merged in January. so the changes are already in. i was told to add comments like this, but i forgot to do it in January.

oksanaguba:
@singhbalwinder and @wlin7 just now i pushed a minor fix about namelists logic. before the PR was inserting the new parameter into each namelist, so, the PR would have NLFAILs for each test. Now there is no isse when i run atm_integration suite:
```
[ac.onguba@blueslogin4 as]$ ./cs.status.20201202_223931_9g7sxu | grep FAIL
    FAIL SMS.ne4_ne4.FC5AV1C-L.anvil_intel.eam-preqx_ftype4 TPUTCOMP Error: Computation time increase > 10 pct from baseline
```

oksanaguba:
Another detail is that this fix works only if orographic GW are on. I assume this is the default, but I am not sure about all cases.

oksanaguba:
Seems that use_gw_oro is always true:
```
[ac.onguba@blueslogin4 acme-b4]$ grep -r use_gw_oro *
components/eam/doc/ChangeLog: - Three new namelist variables, use_gw_oro, use_gw_front, and
components/eam/bld/namelist_files/namelist_definition.xml:<entry id="use_gw_oro" type="logical" category="gw_drag"
components/eam/bld/build-namelist:    add_default($nl, 'use_gw_oro'    , 'val'=>'.true.');
components/eam/bld/build-namelist:    add_default($nl, 'use_gw_oro'    , 'val'=>'.true.');
components/eam/bld/build-namelist:if ($nl->get_value('use_gw_oro') =~ /$TRUE/io) {
components/eam/src/physics/cam/phys_control.F90:logical, public, protected :: use_gw_oro = .true.
components/eam/src/physics/cam/phys_control.F90:      use_hetfrz_classnuc, use_gw_oro, use_gw_front, use_gw_convect, &
components/eam/src/physics/cam/phys_control.F90:   call mpibcast(use_gw_oro,                      1 , mpilog,  0, mpicom)
components/eam/src/physics/cam/gw_drag.F90:  use phys_control,  only: use_gw_oro, use_gw_front, use_gw_convect, use_gw_energy_fix1
components/eam/src/physics/cam/gw_drag.F90:  orographic_only = (use_gw_oro .and. .not. do_spectral_waves)
components/eam/src/physics/cam/gw_drag.F90:  if (use_gw_oro) then
components/eam/src/physics/cam/gw_drag.F90:  if (use_gw_oro) then
```
This fix should be revisited in the future at least because of this issue.

components/eam/src/physics/cam/check_energy.F90
wlin7: I see now and that makes sense. Thanks for the explanation, @oksanaguba.

components/eam/src/physics/cam/gw_drag.F90
wlin7: Sorry for not making it clearer. I was focusing on the resulting ptend%s that will be returned to the caller. 

When I mentioned ttgw, I didn't mean the final one. I meant the ttgw returned from each call to gw_drag_pro before this conditional nlock.  In the original codes, ptend%s accumulates these ttgw,   plus -(udu+vdv). With fix1, the accumulation in ptend%s up to this block is not retained.  

ttgw calculated inside gw_drag_pro is the sum of diffusion of dry static energy and conversion of kinetic energy, as the comments and code flow in gw_drag_pro suggest. I have to admit I am not an expert on GWD; just following the codes

oksanaguba: yes, the original values in ttgw and ptend%s are lost.

oksanaguba:
@singhbalwinder and @wlin7 -- are you ok to merge this? WC wanted it soon.

wlin7:
> @singhbalwinder and @wlin7 -- are you ok to merge this? WC wanted it soon.

Sorry, @oksanaguba , I still have some reservation because the fix appears like discounting part of the gwd processes (in a systematic way, I would say). It would not only remove contribution to ptend%s  from ttgw due to gw_oro, but also gw_front and gw_convect. All three sources are active in our typical compsets. Your results no doubt shows that part of calculation is responsible for the energy conservation issue. 

Have you checked with Yaga if she is comfortable with the fix1 approach?

oksanaguba:
We are talking to Yaga and her co-workers (i do not see her on github, but i will ask her about this). It will take me 1-2 weeks to understand what would be a good energy fix here. Please, also note that the fix is off by default and that this PR is here because WC group requested it. 

Regarding overwriting ptend%s, in convective and frontal GW code there is a call to momentum_energy_conservation(). In there, the second part of the routine does essentially the same thing as the fix introduced here (up to minor details). I do not understand why the routine wasn't called in the oro GW part (aside from tend_levels assigned probably in a wrong way for oro part of the code). That is, in convect and front GW ptend%s was computed for energy conservation similarly to this fix. 

Essentially, there is no way to conserve energy without recomputing ptend%s the way the fix here is implemented. There are minor details whether to peanutbutter energy and whether to do it below or above some regions. These details are what I am going to ask Yaga.

Thanks for your thoughts, tagging @golaz in case there is any interest in this discussion.

wlin7:
Hi @oksanaguba , thanks for pointing out the call to momentum_energy_conservation for gw_convect and gw_front; and it is true doing the conservation constraint in similar spirit. There is still a difference there, though. By calling momentum_energy_conservation, it just subtracts a net energy change below source level, in both ttgw (for output) and ptend%s (which will be carried forward).

That said, I agree it is ok to merge this PR for now, because it is not on by default, and allow for tests to see the impact on other aspects of the simulation in addition to energy conservation. The impact on simulation results is of my major concern because the fix seems to drop some part of the gwd processes. Maybe the net impact on simulation results are small and all will be good.

wlin7:
@oksanaguba , as for your question above why there wasn't a call to  momentum_energy_conservation for gw_oro, I think it is because the source for gw_oro is at bottom level. 

There is probably something we can learn from the treatment in that subroutine:  it first computes the net energy gain or loss due to the gwd (convection or frontal), then subtracting that amount of energy over the levels below the source level to maintain energy conservation. That means if we  add ptend%s accumulated from gw_convect and gw_front to your fix1 assignment for ptend%s, it would still preserve the conservation property as with the current fix1.

And if that is true, it would mean the source of energy leak is solely in the ttgw from gw_oro in the original update of ptend%s. And if we want to account for gw_oro's ttgw in ptend%s, while applying conservation constraint, we could still call  momentum_energy_conservation but change to subtract mass weighted net column energy gain/loss from all levels. 

Just some more speculation here, I should say.

oksanaguba:
My understanding is that the fix in momentum_.. routine is peanutbuttered where the GW are active (at least, convect GW are active in region 0<z<h according to their papers), not below. For oro waves, i am not sure what's going on, iirc tend_level was set to 72 everywhere in the run i tried. The way i see it, there is a confusion about state%s and ptend%s. The former is DSE, the latter is the change in enthalpy (roughly, h=c_pT, not s=c_pT+gz). When in EAM ptend%s is computed and applied, it is the same as computing and applying dT, not dDSE. Thus i am not sure about comments for DSE in GW code and whether it is important to have DSE tendency (or, maybe, it should be converted to temperature tendency).

That aside, the difference between fix1 and the original code is that the original code peanutbutters the leak, and fix1 does it locally. I agree that the leak is probably in the oro part only (though i did not check numerically), but momentum_... routine also applies taucd to the momentum flux. There was probably a reason not to apply it in the oro part. That is another question that i will try to figure out.

wlin7:
> My understanding is that the fix in momentum_.. routine is peanutbuttered where the GW are active (at least, convect GW are active in region 0<z<h according to their papers), not below. 

You are right the GWD effect more active above the source level. So my understanding of the implementation is that the artificial conservation fix, which is expected to be small,  is applied to the levels below (from tend_level to plev). I think this is purposely done to preserve the parameterized effect in the active region. For oro_gw, it make senses tend_level=72 because. that is the source level, thanks for the confirmation.

As for enthalpy h vs DSE, as ptend is for local time changes, dh=dDSE. Therefore, dT=ds/cpair.

oksanaguba:
Do not merge yet.

oksanaguba:
Closing this, will do another PR soon.

PR #3963, opened 2020-11-28
Eliminate call-by-reference argument error in call to proc_heap_adjust

### LINKED ISSUE #3961: LCHUNKS error with ne11 F-case using DEBUG and threads on Anvil
On Anvil, with DEBUG, we see the following error with `res=ne11_ne11 compset=FC5AV1C-L`

```
 24: forrtl: severe (408): fort: (2): Subscript #1 of the array LCHUNKS has value 827 which is greater than the upper bound of 826
 24:
 24: Image              PC                Routine            Line        Source
 24: e3sm.exe           000000000CFDF250  Unknown               Unknown  Unknown
 24: e3sm.exe           00000000039397D5  phys_grid_mp_phys        1015  phys_grid.F90
 24: e3sm.exe           000000000426ABC5  inital_mp_cam_ini          40  inital.F90
 24: e3sm.exe           000000000087D34F  cam_comp_mp_cam_i         155  cam_comp.F90
 24: e3sm.exe           0000000000838445  atm_comp_mct_mp_a         315  atm_comp_mct.F90
 24: e3sm.exe           00000000004652ED  component_mod_mp_         257  component_mod.F90
 24: e3sm.exe           0000000000429C61  cime_comp_mod_mp_        1399  cime_comp_mod.F90
 24: e3sm.exe           000000000045BDE1  MAIN__                    122  cime_driver.F90
 24: e3sm.exe           0000000000416E1E  Unknown               Unknown  Unknown
```

I see the error with intel18 compiler and with 8 nodes using threads (144 MPI's with 18x2).

Same test on Cori does not fail in this way (but does require a fix as explained here https://github.com/E3SM-Project/E3SM/issues/3628 to run in DEBUG -- we don't think that issue is otherwise related to this one)

@worleyph found that changing -O0 to -O1 in the DEBUG definition in Macros.cmake eliminates the error

worleyph:
a) Replace array reference argument with a scalar

The call to

  call proc_heap_adjust(inv_proc_heap(ntmp2), proc_heap_remove,   &
                        max_nvsmptasks, proc_heap_len, proc_heap, &
                        proc_cost, inv_proc_heap)

in phys_grid.F90 where

  subroutine proc_heap_adjust(updated, remove, heap_dlen, &
                              heap_len, heap, cost, inv_heap)
  ...
  integer, intent(in)     :: updated
  ...
  integer, intent(inout)  :: inv_heap(0:npes-1)

modifies 'inv_heap' in the routine proc_heap_adjust, and changes the
value at the location 'inv_proc_heap(ntmp2)'. Compilation with
optimization greater than -O0 apparently caches this value (since it
is declared 'intent(in)'), so 'updated' is not changed within the
scope of the 'proc_heap_adjust' routine. -O0 compilation does not,
changing the value on 'updated' internally when 'inv_proc_heap(ntmp2)'
is changed, which causes a failure. This bug is fixed by changing
the call as follows:

  integer :: updated
  ...
  updated = inv_proc_heap(ntmp2)
  call proc_heap_adjust(updated, proc_heap_remove,   &
                        max_nvsmptasks, proc_heap_len, proc_heap, &
                        proc_cost, inv_proc_heap)

b) Fix mis-declared variables

The local variables 

   real(r8):: proc_ip      ! process id of parent of current element
   real(r8):: proc_il      ! process id of left child of current element
   real(r8):: proc_ir      ! process id of right child of current element

in the routine proc_heap_adjust in phys_grid.F90 are mis-declared. These
should be integer (though the error is innocuous except for the Cray
compiler).

[BFB]

Fixes #3961

### DIFF
--- a/components/eam/src/physics/cam/phys_grid.F90
+++ b/components/eam/src/physics/cam/phys_grid.F90
@@ -6384,6 +6384,9 @@ logical function phys_grid_initialized ()
             proc_heap_remove = .true.
          endif
          if ((proc_heap_update) .or. (proc_heap_remove)) then
+            ! Pass scalar 'updated' to 'proc_heap_adjust' instead of
+            ! 'inv_proc_heap(ntmp2)' to prevent this argument from changing
+            ! value when 'inv_proc_heap' is modified inside of 'proc_heap_adjust'
             updated = inv_proc_heap(ntmp2)
             call proc_heap_adjust(updated, proc_heap_remove,   &
                                   max_nvsmptasks, proc_heap_len, proc_heap, &
--- a/components/eam/src/physics/cam/phys_grid.F90
+++ b/components/eam/src/physics/cam/phys_grid.F90
@@ -6488,9 +6488,9 @@ logical function phys_grid_initialized ()
    integer :: proc_last    ! process id of last element
    integer :: proc_updated ! process id of updated element
    integer :: proc_i       ! process id of current element
-   real(r8):: proc_ip      ! process id of parent of current element
-   real(r8):: proc_il      ! process id of left child of current element
-   real(r8):: proc_ir      ! process id of right child of current
+   integer :: proc_ip      ! process id of parent of current element
+   integer :: proc_il      ! process id of left child of current element
+   integer :: proc_ir      ! process id of right child of current
                            !  element
 
    integer :: last_nonleaf ! index of last non-leaf in heap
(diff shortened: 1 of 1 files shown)


worleyph:
The error was observed only on Anvil and only when compiled with DEBUG==.true. (actually, when compiled with -O0) for a particular grid resolution/decomposition (with threads), but likely could occur in other situations. Tested that this eliminates the issue on Anvil with the reproducer identified by @ndkeen , and is BFB when DEBUG==.false. .

worleyph:
The mis-declared variables issue was identified by @amametjanov .

ambrad (approved):

amametjanov (approved):

worleyph:
@singhbalwinder , could you please integrate this bug fix PR? Thanks.

singhbalwinder:
Sure. I will start merging it today.

singhbalwinder:
Merged to next

PR #3964, opened 2020-11-30
Update production test for v2

rljacob:
Update the 2 atm-only and 1 fully coupled cases in the production suite to
use v2 settings.  Use tri-grid resolutions and CMIP6 compsets for all.
Also use a test mod that will turn on theta-l and other settings being used
in v2 testing.

[BFB]


rljacob:
Merged to next so the  new test names can get in cdash.

jonbob (approved):
From a visual inspection this looks like it's replicating one of the current candidates for the fully-coupled v2 configuration

PR #3965, opened 2020-11-30
Changes NTASKS for ELM-USRDAT test

bishtgautam:
ELM-USRDAT is a single-point test, so NTASKS are set to 1.

[BFB]


dqwu (approved):
Looks good, thanks.

PR #3967, opened 2020-11-30
account name on bebop cluster has been renamed 'e3sm'

sarich:
This change puts the new name in the cime config file

[BFB]


rljacob:
You can merge this to master.

PR #3971, opened 2020-12-03
Remove references to "cam" in submodule paths comment strings

singhbalwinder:
Replace references to "cam" and shorten submodule names in .gitmodules file

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -15,7 +15,7 @@
 	path = externals/kokkos
 	url = git@github.com:E3SM-Project/kokkos.git
 	branch = e3sm/kokkos
-[submodule "components/cam/src/physics/rrtmgp/external"]
+[submodule "RRMGP"]
 	path = components/eam/src/physics/rrtmgp/external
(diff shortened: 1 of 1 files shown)


jgfouca (approved):

.gitmodules
```diff
@@ -15,7 +15,7 @@
 	path = externals/kokkos
 	url = git@github.com:E3SM-Project/kokkos.git
 	branch = e3sm/kokkos
-[submodule "components/cam/src/physics/rrtmgp/external"]
+[submodule "RRMGP"]
```
sarats, line 18: Just eyeballing, isn't the 'T' missing in RRTMGP here?

sarats: Or does this not matter in terms of finding submodule....

rljacob: Not for finding since what is in quotes is just a label to use for submodule commands.  But the T should be there.

singhbalwinder: Thanks for noticing the typo. I will fix it.

PR #3798, opened 2020-08-27
Three small BFB fixes.

ambrad:
This PR fixes three innocuous bugs, in three separate commits:

1. SL transport was exchanging one more level than necessary in its trajectory calculation. The data were ignored, so fixing this reduces comm by a little but does not change the answer.
1. The dynamics-physics remap module can do a total-area calculation on each grid independent of phys_grid.F90, but previously this was possible only in unit testing and not full simulations. Fix this. 
1. Threading was not working properly from a performance perspective in HOMME's boundary exchange. But it was still giving the correct answer.

Fixes #3587.

[BFB]

PR #3979, opened 2020-12-08
Fixes a performance bug that is already on E3SM master.

ndkeen:
A logical variable is set at declaration, when it should be set in the routine.
Fixing this will avoid a thread sync and the performance improvement
will be larger with more threads (especially useful for KNL)

[bfb]

### DIFF
--- a/components/homme/src/share/bndry_mod_base.F90
+++ b/components/homme/src/share/bndry_mod_base.F90
@@ -66,10 +66,12 @@ contains
     character*(80) errorstring
 
     logical(kind=log_kind),parameter              :: Debug=.FALSE.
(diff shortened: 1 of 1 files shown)


ambrad:
Context: This PR brings in the changes to the file https://github.com/E3SM-Project/E3SM/pull/3798/files#diff-44006873f967109dd68cde2ccecfc984a1c5ad031773cc6004924088c1b34d0c in PR #3798.

ambrad (approved):
I did a direct visual comparison of these changes. Other than a shift in line numbers, they are identical to what we did in master. Thanks!

AaronDonahue:
Merging this pull request to maint-1.0

Issue #3908, opened 2020-10-27
maint-1.1 not supported on Cori-KNL or on Cori-Haswell


worleyph:
I) For Cori-KNL using FC5AVIC-L and ne4_ne4:

 a) maint-1.1 still refers to /project/projectdirs/acme/inputdata

% ./case.setup
ERROR: inputdata root is not a directory: /project/projectdirs/acme/inputdata

 b) after replacing acme with e3sm in the case directory (three files), it then dies with 

% ./case.setup
...
   Calling /global/u2/w/worleyph/E3SM/MAINT/maint-1.1/E3SM/components/cice/cime_config/buildnml
ERROR: Command: './generate_cice_decomp.pl -ccsmroot /global/u2/w/worleyph/E3SM/MAINT/maint-1.1/E3SM -res ne4np4 -nx 866 -ny 1 -nproc 96.0 -thrds 1 -output all' failed with error 'b'Value "96.0" invalid for option nproc (number expected)\nSYNOPSIS\n     generate_cice_decomp.pl [options]\nOPTIONS\n     -ccsmroot <path>               Full pathname for ccsmroot\n                                    (required)\n     -nproc <number>      (or -n)   Number of mpi tasks used.\t\n                                    (required)\n     -nx <number>                   number of lons \n                                    (optional, default is 320)\n     -ny <number>                   number of lats\n                                    (optional, default is 384)\n     -res <resolution>    (or -r)   Horizontal resolution \n                                    (optional, default gx1v6)\n     -thrds <number>      (or -t)   Number of threads per mpi task\n                                    (optional, default 1)\n     -output <type>\t  (or -o)   Either output: all, maxblocks, bsize-x, bsize-y, or decomptype\n\t\t\t            (optional, default all)\n\nEXAMPLES\n\n   generate_cice_decomp.pl -res gx1v6 -nx 320 -ny 384  -nproc 80 -output maxblocks\n\n   will return a single value -- the optimum max number of blocks to utilize'' from dir '/global/u2/w/worleyph/E3SM/MAINT/maint-1.1/E3SM/components/cice/bld'

-------------

II) For Cori-Haswell using FC5AVIC-L and ne4_ne4:

 a) maint-1.1 still refers to /project/projectdirs/acme/inputdata

% ./case.setup
ERROR: inputdata root is not a directory: /project/projectdirs/acme/inputdata

 b) after replacing acme with e3sm in the case directory (three files), it then dies with 

% ./case.setup
[...]

PR #3980, opened 2020-12-09
Updates for Cori to use maint-1.1

ndkeen:
Several accumulated changes for Cori that are part of maint-1.0.
These changes should have no impact on results or performance.

Use scontrol instead of sqs for provenance.
Use CFS filesystem instead of projectdirs
Use e3sm_scratch instead of acme_scratch
Use project name of e3sm instead of acme
Change the way number of processors specified on srun line which can impact tests like PEM.
For cori-knl include fancier way of computing the -c flag which only affects certain cases.
Fix typo on module name.
Don't load darshan or hugepages module
Comment MPICH_CPUMASK_DISPLAY env variable as it's too verbose.
Remove PERL5LIB env var setting as that was temporary solution

[bfb]

I verified that the following test builds/runs on cori-knl:
SMS_D_P64x2_Ln5.ne4_ne4.FC5AV1C-L

### DIFF
--- a/cime/config/e3sm/machines/config_machines.xml
+++ b/cime/config/e3sm/machines/config_machines.xml
@@ -61,15 +61,15 @@
   <TESTS>e3sm_developer</TESTS>
   <COMPILERS>intel,gnu</COMPILERS>
   <MPILIBS>mpt</MPILIBS>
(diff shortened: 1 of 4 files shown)


ndkeen:
Fixes #3908

Issue #3909, opened 2020-10-27
'-f' is no longer a valid option for 'sqs' command on NERSC, causing case.submit to fail for maint-1.2


worleyph:
As documented in Issue #3699 , '-f' is no longer a valid option for 'sqs' command on NERSC. A workaround was implemented for master and for maint-1.0. However it was not implemented for maint-1.2 (for either Cori-KNL or Cori-Haswell). It was also not implemented for maint-1.1, but maint-1.1 fails during setup (see Issue #3908 ), so it has other problems.

worleyph:
Files that need to be changed are provenance.py, syslog.cori-haswell, and syslog.cori-knl . After making these changes, an F case ran successfully with full performance data archiving.

rljacob:
Can you make the PR with those changes?

worleyph:
@ndk submitted the PR to fix this for maint-1.0 (PR #3704 ). Changes should be identical. Could Noel take on this task? He'll need to make the identical changes (and more) in maint-1.1 .

rljacob:
sure.

ndkeen:
Thought PR 3981 would have closed this issue.

PR #3809, opened 2020-09-02
Update mpas-source: bugfix 3rd order advection coeff

mattdturner:
Fix bug in MPAS third-order advection coefficients, where the coefficients incorrectly have a plus sign instead of a minus.

See https://github.com/MPAS-Dev/MPAS-Model/pull/656

[non-BFB]

PR #3981, opened 2020-12-09
Several accumulated changes for Cori that are part of maint-1.0 and master

ndkeen:
Several accumulated changes for Cori that are part of maint-1.0 and master.

These changes should have no impact on results or performance.

Change using sqs to scontrol for provenance
For cori-knl include fancier way of computing the -c flag which only affects certain cases.
Fix typo on module name to remove
Don't load darshan or hugepages module

[bfb]


ndkeen:
Tested with `SMS_D_P64x2_Ln5.ne4_ne4.FC5AV1C-L` on cori-knl

ndkeen:
Fixes #3909

ndkeen:
Fixes #3809

PR #3983, opened 2020-12-10
Bug fix in sea ice coupling to freezing temperature

proteanplanet:
Bug fix in sea ice coupling file to use the same freezing temperature as in the main MPAS-SeaIce code according to the flag 'mushy', 'linear_salt' or the default -1.8 Celsius.

[non-BFB]

### DIFF
--- a/components/mpas-seaice/driver/ice_comp_mct.F
+++ b/components/mpas-seaice/driver/ice_comp_mct.F
@@ -57,8 +57,9 @@ module ice_comp_mct
    use seaice_time_integration
    use seaice_error, only: seaice_check_critical_error
 
-   use ice_constants_colpkg, only: &
-       Tffresh
+   use ice_constants_colpkg, only: Tffresh
+
+   use ice_colpkg, only: colpkg_sea_freezing_temperature
 
 !
 ! !PUBLIC MEMBER FUNCTIONS:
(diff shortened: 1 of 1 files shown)


akturner:
@proteanplanet: Does the ocean use the same freezing temperature as the sea ice? If not is this an issue?

xylar:
@akturner, @proteanplanet and I made sure they're the same: https://github.com/E3SM-Project/E3SM/pull/3861

proteanplanet:
> @proteanplanet: Does the ocean use the same freezing temperature as the sea ice? If not is this an issue?

Yes, the ocean is using the same freezing temperature as the 'mushy' option now.  This was fixed in MPAS-Ocean PRs as per Xylar's comment.

akturner (approved):

components/mpas-seaice/driver/ice_comp_mct.F
```diff
@@ -2050,9 +2054,6 @@ subroutine ice_import_mct(x2i_i, errorCode)!{{{
 !-----------------------------------------------------------------------
       do i = 1, nCellsSolve
         seaSurfaceTemperature(i)  = seaSurfaceTemperature(i) - seaiceFreshWaterFreezingPoint
-        seaFreezingTemperature(i) = -1.8_RKIND                 ! hardwired for NCOM
```
rljacob, line 2053: NCOM?  Wow that's old.

njeffery:
Verified changes. GCASE build and submission was successful.

rljacob (approved):

jonbob:
no corresponding bug to fix (issue to close) so I removed that label

jonbob:
merged to next

vanroekel:
@jonbob @proteanplanet sorry for the delay here, this will be a nonBFB PR but I'm approving this for integration for WC v2. It looks like what we discussed.

components/mpas-seaice/driver/ice_comp_mct.F
proteanplanet: I know. I was surprised to see it here.

jonbob:
merged to master and expected DIFFs blessed -- except that sandiatoss3 hasn't reported yet and there should be DIFFs there as well

jonbob:
blesses for sandiatoss3 DIFFs are also requested

PR #3985, opened 2020-12-11
Support for ne30np4pg2 in ELM

bishtgautam:
- Adds mapping files for ne30np4pg2
- Adds surface dataset for ne30np4pg2 for 1850 and 2000

[BFB]

### DIFF
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -1174,6 +1174,56 @@
       <mask>oEC60to30v3</mask>
     </model_grid>
 
+    <model_grid alias="ne30pg2_oECv3_ICG" compset="(EAM.+ELM.+MPASO%SPUNUP)">
+      <grid name="atm">ne30np4.pg2</grid>
+      <grid name="lnd">ne30np4.pg2</grid>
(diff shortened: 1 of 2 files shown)


bishtgautam:
The newly generated surface datasets have been uploaded to the E3SM inputdata server.

rljacob:
What were the pg2 runs using before this?

jonbob:
@rljacob - before this we were using the trigrid approach when atm was on ne30pg2, so lnd was running on r05

jonbob:
Thanks @bishtgautam - we'll make a longer test run over the weekend

jonbob:
I added grid configuration support for this new ELM file, so it can run with a variety of ocn/ice v2 meshes

jonbob (approved):
Test case ran one month smoke test out-of-the box

jonbob:
@bishtgautam - can you please merge this soon? I think it will be necessary for the v2 configurations. Thanks!

bishtgautam:
Sorry, for dropping the ball on this. I will merge this PR in a few hours.

PR #3998, opened 2020-12-20
Change optimizations of modal_aero_calcsize.F90 to -O1 for Compy's PGI

singhbalwinder:
In prepration for the PR which refactors calcsize,
modal_aero_calcsize.F90 needs to be compiled with level 1 optimizations
(-O1). This applies only to the PGI compiler. GNU and Intel work fine
 with the default optimizations. Further investigations show that if we
use just the "-nofma" flag (no "fused multiply and add" optimization),
the results stay BFB. "-O1" optimizations implies "-nofma", so compiling
with -O1 is also sufficient for maintaing BFB answers with the PGI
compiler.

[Non-BFB] for compy's PGI compiler only


jgfouca (approved):

wlin7:
Merged to next.

wlin7:
Compy diffs blessed;merged to master.

PR #4000, opened 2020-12-21
Update ocean/seaice defaults for wc

vanroekel:
This PR makes the ocean and seaice configurations identical to water cycle v2 simulations out of the box.

[NML]
[non-BFB]


jonbob (approved):
Looks exactly like the settings in 20201217.beta1_01.piControlSI.compy

proteanplanet (approved):
This review is to approve only the sea ice changes, and I have requested @jonbob to review this with the added changes to MPAS-SI defaults.

vanroekel:
@jonbob I've modified the PR to affect the EC30to60E2r2 only on the ocean side.  Please double check my changes

proteanplanet:
@jonbob @vanroekel: I, too, can just make sure that EC30to60E2r2  is affected here, if you would like?

jonbob (approved):
Looks perfect. I'll get it merged right away. Thanks to you both

jonbob:

Issue #3989, opened 2020-12-15
Turning config_use_high_frequency_coupling on in seaice causes PEM failures


jonbob:
[PR #3984](https://github.com/E3SM-Project/E3SM/pull/3984) was merged to master and turned on config_use_high_frequency_coupling in MPAS-seaice, but is causing PEM failures. If we can't figure it out quickly, we should either revert that PR or submit a new one turning it back off

proteanplanet:
@jonbob Something is not right here.  I have been running with this switched on using E3SM master for months.

jonbob:
@proteanplanet - a PEM test failure doesn't mean the code won't run, just that it's getting different answers on different pe counts

proteanplanet:
@jonbob Talk to you about it in half an hour...

jonbob:
Sounds good. In the meantime, I tested that part of the code a bit and it's not due to the changes to ice_atmos you made in [PR #3948](https://github.com/E3SM-Project/E3SM/pull/3948) -- so it must be something already there in the highfreq coding

akturner:
@jonbob @proteanplanet : The MPAS-Seaice standalone parallel tests fail with the high frequency coupling. Will investigate.

PR #4001, opened 2020-12-21
Update mpas-source: fixed high frequency coupling bug in MPAS-Seaice

akturner:
Corrects a bug in MPAS_Seaice where the wrong velocity indexing is used for high frequency coupling.

Fixes https://github.com/E3SM-Project/E3SM/issues/3989

[BFB except with high freq coupling]


jonbob:
This PR has extraneous updates to components/elm/src/external_models/fates that need to be removed. Testing with only the components/mpas-source submodule hash change gives the following test results:
PASS PET_Ln9.ne30_oECv3_ICG.A_WCYCL1850S.anvil_intel
PASS PEM_Ln9.ne30_oECv3_ICG.A_WCYCL1850S.anvil_intel

The second one is the test that has been failing in automated testing -- so this does fix the bug. Thanks @akturner

PR #4003, opened 2020-12-22
Update mpas-source: Fix bug with high freq coupling in MPAS-Seaice

akturner:
This PR fixes a bug with high frequency coupling in MPAS_Seaice where the cell indices was used for the velocity fields instead of the vertex indices in step_therm1. CICE does the same but with a structured grid the consequences are less.

Fixes https://github.com/E3SM-Project/E3SM/issues/3989

[non-BFB] because high-frequency coupling in on by default currently


jonbob:
the PEM test that had been failing is now passing:

PASS PEM_Ln9.ne30_oECv3_ICG.A_WCYCL1850S.anvil_intel

proteanplanet (approved):
Approved by visual inspection.

jonbob:
Merged to next

jonbob:

PR #4004, opened 2020-12-22
Update mpas-source: Updated sea ice emissivity

akturner:
Updated sea ice emissivity for consistent coupling

Merges https://github.com/MPAS-Dev/MPAS-Model/pull/771

[non-BFB]


proteanplanet (approved):
Strictly speaking, the only change required in this PR is to src/core_seaice/column/constants/cesm/ice_constants_colpkg.F90.  All other changes have originated from working on another PR at the same time as this.   I have checked with @jonbob, and this point ends up not mattering when this is merged into E3SM Master owing to the previous PR: https://github.com/E3SM-Project/E3SM/pull/4003.  However, to be clear, this PR is for a single change to emissivity from 0.95 to 1.0 to ensure instantaneous radiative transfer using our current coupling methods, as physics requires.  Previous tests to sensitivity to this parameter in coupled systems have shown to be minimal.

jonbob:
merged to next

jonbob:

PR #4005, opened 2020-12-23
Add intelmpi modules on Chrysalis

amametjanov:
Add intelmpi modules on Chrysalis

[BFB]


amametjanov:
All expected `e3sm_integration` tests pass: 87 out of 88, except for `SMS_B.ne4_ne4.F-EAMv1-AQP1.chrysalis_intel.eam-hommexx` that fails on other machines too.

PR #4007, opened 2020-12-27
Add new compsets and atm use_case files for v2

wlin7:
Add new compsets along with default atmos namelist parameters for
most used configurations for v2 finalization. The default atmos
parameter values are based on v2alpha5_59 simulation.  The new
compsets with simplified names are
    
F2010, F20TR and WCYCL1850

The parameter settings from alpha5_59 are set via use_case files in this PR
  
 This PR also 
- cleans up obsolete av1c and av1f series of compsets, the 
   associated settings in config_components and use_case files.
- changes to use namelist_defaults_eam.xml 
- ieflx_opt to get value automatically based on config option 
                     
[BFB] except for allactive-v1cmip6 tests


wlin7:
This is an alternative to #4008. With this PR, the existing compsets are not affected. Previous v2 finalization simulations can be re-produced using their run scripts without changes. 

However for the two new F compsets, still need to reset SSTICE data files, because the changes would need to made in cime for the config file for docn.

Until the changes to cime (and committed), need the following to reproduce the results using the new compsets

For F2010SC5v2-CMIP6, add the following to run script
`$xmlchange_exe SSTICE_DATA_FILENAME="/compyfs/inputdata/ocn/docn7/SSTDATA/sst_ice_CMIP6_DECK_E3SM_1x1_2010_clim_c20190821.nc"`

For F20TRC5v2-CMMIP6, add. the following
 ```
$xmlchange_exe -file env_run.xml -id SSTICE_DATA_FILENAME -val "/compyfs/inputdata/ocn/docn7/SSTDATA/sst_ice_CMIP6_DECK_E3SM_1x1_c20180213.nc"
 $xmlchange_exe -file env_run.xml -id SSTICE_YEAR_ALIGN -val 1869
 $xmlchange_exe -file env_run.xml -id SSTICE_YEAR_START -val 1869
 $xmlchange_exe -file env_run.xml -id SSTICE_YEAR_END -val 2016

```

**Comment by xuezhengllnl** [comment]
Hi @wlin7,

Thanks for working on updating the atm default parameters for v2. I have a quick comment before going through the detailed changes. Instead of adding v2 at the first part of the case name, would it be better if we replace cam5 by eam2?

**Comment by wlin7** [comment]
Hi @xuezhengllnl , thanks for bringing this up. Eventually we will need to replace the reference to 'cam5'.  @golaz and @rljacob  can help decide if now is the time.

Another note for the review: comparison of atm_in files would still show diff. This is because all nml variables that are set in use_case files contain an extra space after equal signs, compared to when they are set via user_nl_eam. Simulations have also been performed for all three compsets on compy. They are BFB with alpha5_59 but w/o irrigation nor any settings in user_nl_elm. The script for control coupled case is modified based on [alpha5_59 run.](https://acme-climate.atlassian.net/wiki/spaces/EWCG/pages/2053767288/20201124.alpha5+59+fallback.piControl.ne30pg2+r05+EC30to60E2r2-1900+ICG.compy). The script is further modified accordingly for AMIP and F2010 runs.

**Review by xuezhengllnl — APPROVED** [review]

**Review by rljacob — CHANGES_REQUESTED** [review]
"v2" has the same problem as "c5":  it  eventually becomes outdated and can not be guaranteed to give "v2" answers.

**Comment by xuezhengllnl** [comment]
@wlin7 I checked the parameters in user_case xml files and the new compset related files. These new files look good to me.

**Comment by golaz** [comment]
@wlin7 : this is a massive clean-up. Thank you so much for being willing to undertake this (we should call this "where no one has dared to go before").

Looking at the files, I do have a question for @rljacob in terms of what can be done with CIME. I suspect the answer to my question is "No, can't be done with CIME". But I'll ask anyway.

We now have three new use_cases file:

```
components/eam/bld/namelist_files/use_cases/1850_eam_CMIP6.xml
components/eam/bld/namelist_files/use_cases/2010_eam_CMIP6.xml
components/eam/bld/namelist_files/use_cases/20TR_eam_CMIP6.xml
```

These files contain both **physics settings** as well as **forcing information**, which are essentially orthogonal to each other. As a result, there is significant overlap in the contents of these three files. In an ideal world, it would be nice to further split the content along physics and forcing to avoid unnecessary repetition. We would have files such as:

```
1850_CMIP6.xml
2010_CMIP6.xml
20TR_CMIP6.xml
eam.xml
```

and then CIME would combine their content appropriately for the requested compset. If we later decided to change some physics setting, we would only need to change a single file (`eam.xml`) and all compsets relying on it (in this case 1850_eam_CMIP6, 2010_eam_CMIP6, 20TR_eam_CMIP6) would automatically inherit the change without us having to worry about updating multiple files (and missing one).

If CIME does not support this directly, I wonder whether we could use some xml include mechanism to accomplish this.

**Comment by rljacob** [comment]
I don't think CIME can do that but @jgfouca would know for sure.

Your assumption is incorrect.  They are not supposed to be orthogonal.  The only physics settings that should be in those files are the ones necessary for tuning the model to that particular forcing.  The physics settings common to all three should be moved to namelist_defaults which this PR hasn't done yet.

**Comment by jgfouca** [comment]
@golaz , @rljacob , I think it would be easier to pursue an XML include option. If that doesn't pan out, we could look into a different approach. Based on some grepping around, I think the only think that uses those xml files is `eam/bld/build-namelist` which not owned by CIME, it's owned by the atm team. It's that huge pile of perl that no one understands.

**Comment by golaz** [comment]
@rljacob : if the common settings can later be moved somewhere else, that would take care of removing unnecessary duplication and leave the use_cases files with only forcing dependent settings. I get lost by the number of configuration layers that we have.

**Comment by wlin7** [comment]
>  The only physics settings that should be in those files are the ones necessary for tuning the model to that particular forcing. The physics settings common to all three should be moved to namelist_defaults which this PR hasn't done yet.

That is very true, @rljacob . With #4033, only following 'physics' parameters remain in the use_case file
cosp_lite, history_amwg, history_aerosol, history_aero_optics.  

If we  can determine suitable default for them (esp. the history tape options), they can be moved to  namelist_defaults_eam.xml, too. cosp_lite could set default to true in configure when cosp is enabled.

**Review by xuezhengllnl — APPROVED** [review]

**Review by xuezhengllnl — APPROVED** [review]

**Comment by wlin7** [comment]
Did a test merge to next. There is no actual conflict in namelist_defaults_cam.xml as indicated above.

**Comment by golaz** [comment]
@wlin7 : I verified that I can BFB reproduce the reference simulation on compy (20201217.beta1_01.piControlSI.compy) when using the 'L' layout. I think the PR is ready to be merged.

As already discussed elsewhere (https://github.com/E3SM-Project/E3SM/issues/4038), the results are NBFB when changing layouts: S, M, and L layouts all give different answers. But that's a separate issue that has been present for a while and has nothing to do with this PR.

**Comment by golaz** [comment]
@rljacob : can we merge this on master as a first step?

I'd like to run some tests on Chrysalis, but that branch fails there (hangs in IO). Presumably, that's because this branch is missing some recent updates for Chrysalis. Merging to master would take care of that.

**Comment by rljacob** [comment]
@wlin7 you can start merging this.

**Comment by rljacob** [comment]
I'll start merging this today if @wlin7 can't get to it.

**Comment by rljacob** [comment]
@wlin7 and @golaz Sorry I got confused and wound up merging #4008 to next and master.

**Comment by wlin7** [comment]
No problem with #4008, @rljacob .

Merged to next. Note that there was no conflict regarding components/eam/bld/namelist_files/namelist_defaults_cam.xml.

**Comment by wlin7** [comment]
This PR adds ieflx_opt to atm_in. For all F tests, the value is 0 which is also the default when not set. It will cause nml only diff for F tests (FC5AV1C-L, SCM, cosp, etc.).

**Comment by wlin7** [comment]
Confirmed, F20TR from next with this PR and #4042, but with #3481 reverted BFB when changing PE layout. Also BFB F20TRC5-CMIP6 from master with #4042 changes on tpert  used via src.eam/. @rljacob , your merging of #4008 makes the F20TRC5-CMIP6 test with master easier, as it now gives the same atm_in as F20TR currently only in next.

**Comment by wlin7** [comment]
Following tests need to be blessed (taking sandiatoss3.next as example). gust_param and A_WCYCL185S_CMIP6 tests are expected. The other A_WCYCL1850 tests, though the associated use_case file was not touched, ieflx_opt =2 is now set as default for B-cases. It is not set in the use_case file for A_WCYCL1850, which is 1850_cam5_av1c-04p2.xml, hence ieflx_opt=0 used as default before this PR.
```
ERP_Ld3.ne30_oECv3_ICG.A_WCYCL1850S.sandiatoss3_intel.allactive-pioroot1
ERS.ne11_oQU240.A_WCYCL1850.sandiatoss3_intel   
ERS_Ld5.ne4_ne4.FC5AV1C-L.sandiatoss3_intel.eam-gust_param      
NCK.ne11_oQU240.A_WCYCL1850.sandiatoss3_intel   
PEM_Ln9.ne30_oECv3_ICG.A_WCYCL1850S.sandiatoss3_intel   
PET_Ln9_PS.ne30_oECv3_ICG.A_WCYCL1850S.sandiatoss3_intel.allactive-mach-pet     
SMS_D_Ld1.ne30_r05_oECv3.A_WCYCL1850.sandiatoss3_intel  
SMS_P12x2.ne4_oQU240.A_WCYCL1850.sandiatoss3_intel.allactive-mach_mods

```

**Comment by wlin7** [comment]
Requests submitted to bless diffs on
sandiatoss3, mappy, compy_developer, compy_prod, anvil.

Merged to master.

**Comment by wlin7** [comment]
Request submitted to bless two other sandiatoss3 diffs that require fixes via #4047.
```
SMS_D_Ld1.ne30_oECv3_ICG.A_WCYCL1850S_CMIP6.sandiatoss3_intel.allactive-v1cmip6
SMS_Ld9.ne4pg2_ne4pg2.FC5AV1C-L.sandiatoss3_intel.eam-thetahy_sl_pg2_mass
```
And to bless the nml only diffs, too, for the tests using FC5AV1C-L on compy, mappy and sandiatoss3.

PR #4008, opened 2020-12-27
Set default atmos parameters for v2 in existing compsets

wlin7:
Set default atmos parameter values for the following three compsets that
are used  for v2 finalization simulations based on alpha5_59

   A_WCYCL1850S_CMIP6, F20TRC5-CMIP6 and  F2010SC5-CMIP6

[NBFB] non-BFB for simulations using these  compsets

### DIFF
--- a/components/eam/bld/namelist_files/use_cases/1850_cam5_CMIP6.xml
+++ b/components/eam/bld/namelist_files/use_cases/1850_cam5_CMIP6.xml
@@ -58,44 +58,70 @@
 <!-- Tunable parameters for 72 layer model -->
 
 <ice_sed_ai              > 500.0     </ice_sed_ai>
-<cldfrc_dp1              > 0.045D0   </cldfrc_dp1>
-<clubb_ice_deep          > 16.e-6    </clubb_ice_deep>
 <clubb_ice_sh            > 50.e-6    </clubb_ice_sh>
 <clubb_liq_deep          > 8.e-6     </clubb_liq_deep>  
 <clubb_liq_sh            > 10.e-6    </clubb_liq_sh>
 <clubb_C2rt              > 1.75D0    </clubb_C2rt>
-<zmconv_c0_lnd           > 0.007     </zmconv_c0_lnd>
-<zmconv_c0_ocn           > 0.007     </zmconv_c0_ocn>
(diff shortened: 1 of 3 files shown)


wlin7:
This is an alternative to #4007. Here the existing compsets are used for v2 ,but with default atmos parameters updated following alpha5_59. However, after this PR, the settings for previous v2 finalization simulations need educated adjustments in order to re-produce.

xuezhengllnl:
Hi Wuyin,

As we discussed offline, it is more reasonable to have new use_case files (like #4008) for v2 simulations.

rljacob:
Merged to next

rljacob:
The production suite for those compsets uses a testmod to set most of the v2 params.  But it was missing one:

clubb_ipdf_call_placement              =  2

So there are diffs in the production tests.

wlin7:
@rljacob , with the proposed path,  we want to go with PR #4007, then #4033. not to use this PR. Can I revert it?

rljacob:
To late.

rljacob:
I thought I was workjng on #4007.  Can you still merge #4007 with this one being merged?

wlin7:
It is on me. No impact on #4007, nor #4033. Ok to leave the PR there. We may just let @xuezhengllnl be aware that the use case files have been updated.

wlin7:
Just to double check. The diff due to this PR must have been blessed, right? I am going to merge #4007, it will cause diff again tomorrow.

rljacob:
Yes I just submitted the bless requests.   How will #4007 create diffs?

rljacob:
If you want to undo this, just add more commits to #4007

wlin7:
> Yes I just submitted the bless requests. How will #4007 create diffs?

A_WCYCL1850_CMIP6 and WCYCL1850 will use the same atm use_case file after #4007, which will then be 1850_eam_CMIP6.xml.  

The three use case files here will no more be used after #4007. We can do it anytime if ever needed. The settings in there are essentially the same as for v2. So not bad just leave them there.

rljacob:
Shouldn't we remove A_WCYCL1850_CMIP6?

wlin7:
We may. Probably in a future round of cleanup.

PR #4011, opened 2020-12-30
MMF Variance Transport

whannah1:
This adds a new feature to the MMF to fix the checkerboard noise problem associated with "variance trapping". The new option diagnoses the variance associated with liquid-ice static energy (hl) and total non-precipitating water (qt) fluctuations within the embedded CRMs. The variance is then transported as a scalars on the global GCM grid. The coupling between the  CRM variance and GCM variance tracers is done in a similar manner to the traditional MMF coupling. The new mode is currently optional, but will likely become the default after further verification. This PR also includes a fix for the C++ version of setperturb(), because the initial random values were found to not be sufficiently uncorrelated.

The new model is enabled with the "-use_MMF_VT" configure flag. Two versions of this feature are included, which either use the total variance (i.e. "bulk") or the variance of low-pass filtered fields. The filtered version is enabled by the namelist variable "MMF_VT_wn_max", whose value sets the largest wavenumber that is retained from the FFT-based low-pass filter. The default value of MMF_VT_wn_max=0 disables the filter. 

[non-BFB] (only for C++ MMF)


components/eam/src/physics/crm/samxx/crm_variance_transport.cpp
```diff
@@ -0,0 +1,286 @@
...
+      t_vt(k,icrm) = 0.0;
+      q_vt(k,icrm) = 0.0;
+  });
+
+  // do k = 1,nzm
+  //  do j = 1,ny
+  //    do i = 1,nx
+  //      do icrm = 1,ncrms
+  parallel_for( SimpleBounds<4>(nzm,ny,nx,ncrms) , YAKL_LAMBDA (int k, int j, int i, int icrm) {
+    real t_tmp = t(k,j+offy_s,i+offx_s,icrm);
```
mrnorman, line 145: Rename these because of confusion with tmp_t and tmp_q

mrnorman: If you want, you can inline these temps. `yakl::atomicAdd` doesn't need the 2nd argument to be stored in a temp like Fortran's `!$acc atomic` does

mrnorman (approved):
Looks pretty good to me. I recommend:

* Running on the CPU with `-DYAKL_DEBUG`, which will throw an exception if arrays are out of bounds and dump core files to trace.
* Running samxx GPU to compare against sam CPU to make sure the results are still correct.

I also recommend a small ne4pg2 run on the GPU to make sure you've included all of the `auto &var = ::var` instances. If you haven't, you'll get an invalid memory address segfault, which should dump useful core files. Be sure to specify `-g` to get lines / symbols beyond function entry points. Then, look for any global variables you forgot to "auto" into local scope.

If your core files aren't useful or you don't get any, let me know, and I'll try to run a GPU case on my laptop.

components/eam/src/physics/crm/crm_output_module.F90
```diff
@@ -88,6 +88,11 @@ module crm_output_module
       real(crm_rknd), allocatable :: qcltend (:,:)          ! CRM output tendency of cloud liquid water
       real(crm_rknd), allocatable :: qiltend (:,:)          ! CRM output tendency of cloud ice
 
+      real(crm_rknd), allocatable :: t_vt_tend (:,:)       ! CRM output tendency for variance transport
+      real(crm_rknd), allocatable :: q_vt_tend (:,:)       ! CRM output tendency for variance transport
+      real(crm_rknd), allocatable :: t_vt_ls   (:,:)       ! variance transport tendency
+      real(crm_rknd), allocatable :: q_vt_ls   (:,:)       ! variance transport tendency
+
```
pressel, line 95: Comments could be more descriptive here.  Perhaps stating if they are tendencies from the crm -> gcm or gcm -> CRM. Also the comments should say what species these are the variance tendencies of.

components/eam/src/physics/crm/sam/crm_variance_transport.F90
```diff
@@ -0,0 +1,368 @@
...
+      do icrm = 1,ncrms
+         t_mean(icrm,k) = t_mean(icrm,k) * factor_xy
+         q_mean(icrm,k) = q_mean(icrm,k) * factor_xy
+      end do
+   end do
+
+   !----------------------------------------------------------------------------
+   ! calculate anomalies
+   !----------------------------------------------------------------------------
+   if (filter_wn_max>0) then ! use filtered state for anomalies
```
pressel, line 215: This is probably nit-picky, but I think it is better to call these fluctuations rather than anomalies.

components/eam/src/physics/crm/sam/crm_variance_transport.F90
```diff
@@ -0,0 +1,368 @@
...
+   use vars,         only: t
+   use microphysics, only: micro_field, index_water_vapor
+   implicit none
+
+   ! interface arguments
+   integer, intent(in) :: ncrms
+
+   ! local variables
+   real(crm_rknd), parameter   :: pert_scale_min = 1.0 - 0.1
+   real(crm_rknd), parameter   :: pert_scale_max = 1.0 + 0.1
```
pressel, line 301: Comments should be used to explain what these parameters are and how they are set.  Perhaps it would be better to not hardwire these here.

components/eam/src/physics/crm/samxx/crm_variance_transport.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#include "samxx_const.h"
+#include "vars.h"
+#include "YAKL_fft.h"
+
+void VT_filter(real4d &f_in, real4d &f_out);
+void VT_diagnose();
+void VT_forcing();
```
pressel, line 9: Add end of file line.

components/eam/src/physics/crm/samxx/post_timeloop.cpp
```diff
@@ -236,6 +247,11 @@ void post_timeloop() {
     vln  (k,icrm) = vln  (k,icrm) * factor_xy;
   });
 
+#if defined(MMF_VT)
+  // extra diagnostic step here for output tendencies
+  VT_diagnose();
+#endif
+
```
pressel, line 254: @mrnorman Is the C++ code base adopting as a standard the use of preprocessor directives to implement optional behavior? If so, I worry this is a decision we may come to regret.

components/eam/src/physics/crm/samxx/setperturb.cpp
```diff
@@ -1,47 +1,39 @@
...
+  // which avoids a problematic sensitivity to pcols.
   int  constexpr perturb_num_layers  = 5;    // Number of levels to perturb
   real constexpr perturb_t_magnitude = 1.0;  // perturbation LSE amplitube [K]
-
-  real2d t02 = real2d("t02",perturb_num_layers,ncrms);
-  yakl::memset(t02,0.);
-
+  real factor_xy = 1. / (nx*ny);
+  auto t_host     = t    .createHostCopy();
+  auto t0_host    = t0   .createHostCopy();
```
pressel, line 10: Formatting question; do mean for there to be extra spaces before the "."?

components/eam/src/physics/crm/samxx/setperturb.cpp
```diff
@@ -1,47 +1,39 @@
...
+      }
+      for (int j = 0; j < ny; j++) {
+        for (int i = 0; i < nx; i++) {
+          t_host(k,j+offy_s,i+offx_s,icrm) = t_host(k,j+offy_s,i+offx_s,icrm) * t0_host(k,icrm) / t02;
+        }
+      }
+    }
+  }
+  t_host.deep_copy_to(t);
+}
```
pressel, line 39: Missing end of file line.

pressel: Did you mean for this change to the initial perturbations to be part of the PR?

components/eam/src/physics/crm/samxx/post_timeloop.cpp
mrnorman: I think we should shield the C++ code from these in `const.h`, though, by setting `constexpr bool` variables. @whannah1 , can you go into `samxx/samxx_const.h`, and put the following:
```
#ifdef MMF_VT
  bool constexpr use_mmf_vt = true;
#else
  bool constexpr use_mmf_vt = false;
#endif
```

Then use `if (use_mmf_vt) { ... }` instead of `ifdef`

components/eam/src/physics/crm/samxx/setperturb.cpp
mrnorman: Only if it allows `.createHostCopy()` to line up with other variables above and below this code snippet.

components/eam/src/physics/crm/samxx/post_timeloop.cpp
mrnorman: @pressel , note that there are circumstances when we cannot shield the code from `ifdef`s. The most notable of these situations is when the paramter list to a routine depends on the `ifdef`.

mrnorman: Also, this might get hairy. Sometimes, you have to shield code with these, and the only ways to do that are `if constexpr`, which is not necessarily supported by all GPU compilers (notably nvcc the last time I checked); or to use `std::enable_if<>`, which is really ugly to have in user code. I'd honestly rather have `ifdef` than `enable_if` templates if it's asthetic we're talking about.

pressel (approved):
Hi Walter. This looks good to me. It is nice to see these ideas we have been talking about for so long implemented in the code and in action!

components/eam/src/physics/crm/samxx/post_timeloop.cpp
pressel: @mrnorman I agree that ifdefs probably can't be avoided entirely, but I think should be avoid if possible without adding excess code complexity. We are on the same page.

whannah1: I've gone through and removed all ifdefs. I added 2 new namelist variables which are passed into the CRM that control whether VT is used and whether the bulk or filtered method is used.

components/eam/src/physics/crm/samxx/setperturb.cpp
whannah1: Done. Yes, this fix was important for Vt. I'd like to keep them together for this instance rather than separating this out into a separate PR.

components/eam/src/physics/crm/samxx/setperturb.cpp
whannah1: I'll lose the spacing.

components/eam/src/physics/crm/samxx/crm_variance_transport.h
whannah1: done.

components/eam/src/physics/crm/sam/crm_variance_transport.F90
whannah1: I don't care either way, so I'll use "fluctuations".

components/eam/src/physics/crm/crm_output_module.F90
whannah1: done.

components/eam/src/physics/crm/samxx/crm_variance_transport.cpp
whannah1: done.

components/eam/src/physics/crm/sam/crm_variance_transport.F90
whannah1: I agree in general about the hardwiring... but in this case the value doesn't seem to make much difference, so I really doubt we'd want to be adjusting this via the namelist. I suppose it might come in handy in some odd situation where we are having trouble getting past a crash for a production run. I'm just gonna leave them in place for now (with comments) and we can turn them into namelist variables pretty quickly if there's a need.

whannah1:
I've tried @mrnorman 's suggestions regarding building in debug mode and running CPU/GPU cases on Summit and verified that everything is consistent with my initial tests. I also revisited the sensitivity of the filtered VT with some GPU cases and qualitatively reproduced the results I found with fortran cases, so I'm confident the filter is working properly in both Fortran and C++. The latest commits should address all the comments so far.

components/eam/src/physics/crm/sam/crm_variance_transport.F90
pressel: That sounds good to me, adding comments is sufficient.

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #4013, opened 2021-01-03
cmake file for standalone HOMMME with e3sm default modules

mt5555:
cmake machine file for standalone HOMME, suitable when the 
intel/20.0.4-kodw73g and  intel-mpi/2019.9.304-tkzvizk  modules are loaded.  
These are the default E3SM modules as of 2021/1/1

Switch standalone HOMME builds to use Intel fortran and C compilers
Switch regression tests to use mpirun instead of srun


rljacob:
Notes from IG:  was delayed because we might switch to different MPI.  Can now go in.

amametjanov:
Requested and replaced by E3SM-Project/E3SM#4044.

PR #4010, opened 2020-12-30
Add performance data collection support for chrysalis

worleyph:
a) Add support for chrysalis to provenance.py

Created a new CIME branch (worleyph/chrysalis_support)
with a modified provenance.py that adds the chrysalis
system to the list of supported systems.

The CIME branch also (a) removes support for mira and
titan systems, and (b) fixes archiving of batch job
output files for slurm systems anvil, compy, cori-haswell,
and cori-knl.

After merging, CIME will point to this branch, moving the hash from
    5cef5558 to 5ffbaf480e

b) Add job monitoring script for chrysalis

Added syslog.chrysalis, to provide job progress monitoring
and checkpointing for the chrysalis system.

Fixes #4009 

[BFB]

PR #4016, opened 2021-01-04
Change default location for SAVE_TIMING_DIR on Chrysalis

worleyph:
Currently, jobs on Anvil and Chrysalis archive performance data
in the same location. This may cause errors if Jenkins jobs,
one on Anvil and one on Chrysalis, attempt to access these
performance data simultaneously. To avoid this possibility,
the default location for archiving performance data, specified
in SAVE_TIMING_DIR, is changed for Chrysalis.

[BFB]


worleyph:
Tested with an F case on Chrysalis. Even without the performance archiving support changes in #4010 , enough data was archived in the new location (/lcrc/group/e3sm/PERF_chrysalis/performance_archive) to verify that this is working correctly.

sarats (approved):
Looks good

PR #4023, opened 2021-01-06
Update mpas-source: Updated sea ice for algal grazing bug

### LINKED ISSUE #3490: Generalize grazing for multiple sea ice algal groups
Grazing as a linear function of biomass is not sufficient to differentiate multiple algal groups.  A simple power law formulation can distinguish heavily and lightly grazed species.

akturner:
This pull request brings in a new version of MPAS-Seaice that corrects a bug with the algal grazing closure.

Adds https://github.com/MPAS-Dev/MPAS-Model/pull/742

Fixes #3490

[non-BFB] for BGC configurations only


njeffery (approved):
I confirmed that changes are consistent with new grazing implementation.   I also tested these changes in a 6 month GCASE with ocean-ice BGC coupled and changes were similar to a  previous test case.  The new namelist values in Registry will still need to be added as defaults in a later PR.

jonbob:
merged to next

jonbob:

PR #4024, opened 2021-01-06
Enable FATES harvest using fdyndat data passed from ELM to FATES

aldivi:
ELM code (for harvest control and FATES harvest control) has been updated so that ELM 
input harvest can be passed to FATES to drive FATES harvest code. The ELM namelist 
variables do_harvest and fdyndat can now be used to drive FATES harvest. The FATES 
code has not been modified, as it has already been updated for this purpose. Modifications
to the harvest read and get year functions were required to ensure that the correct harvest
year data are read in. ELM harvest results are unchanged, and FATES harvest matches 
the input based on secondary forest fraction. Branch run control has been updated to enable
branch runs.

[non-BFB] only for FATES and BGC cases

### DIFF
--- a/components/elm/src/dyn_subgrid/dynSubgridControlMod.F90
+++ b/components/elm/src/dyn_subgrid/dynSubgridControlMod.F90
@@ -231,12 +231,8 @@ contains
     end if
 
     if (dyn_subgrid_control_inst%do_harvest) then
-       if (.not. use_cn) then
-          write(iulog,*) 'ERROR: do_harvest can only be true if use_cn is true'
-          call endrun(msg=errMsg(sourcefile, __LINE__))
-       end if
-       if (use_fates) then
-          write(iulog,*) 'ERROR: do_harvest currently does not work with use_fates'
+       if (.not. (use_cn .or. use_fates)) then
+          write(iulog,*) 'ERROR: do_harvest can be true only if use_cn is true or use_fates is true'
           call endrun(msg=errMsg(sourcefile, __LINE__))
        end if
     end if
--- a/components/elm/bld/ELMBuildNamelist.pm
+++ b/components/elm/bld/ELMBuildNamelist.pm
@@ -2636,11 +2636,8 @@ sub setup_logic_do_harvest {
   if (string_is_undef_or_empty($nl->get_value('flanduse_timeseries'))) {
     $cannot_be_true = "$var can only be set to true when running a transient case (flanduse_timeseries non-blank)";
   }
-  elsif (!value_is_true($nl->get_value('use_cn'))) {
-    $cannot_be_true = "$var can only be set to true when running with CN (use_cn = true)";
-  }
-  elsif (value_is_true($nl->get_value('use_fates'))) {
-    $cannot_be_true = "$var currently doesn't work with FATES";
+  elsif (!( value_is_true($nl->get_value('use_cn')) || value_is_true($nl->get_value('use_fates')) )) {
+    $cannot_be_true = "$var can only be set to true when running with CN (use_cn == true) or when using FATES (use_fates == true)";
   }
   
   if ($cannot_be_true) {
--- a/components/elm/src/dyn_subgrid/dynHarvestMod.F90
+++ b/components/elm/src/dyn_subgrid/dynHarvestMod.F90
@@ -25,17 +25,22 @@ module dynHarvestMod
   use ColumnDataType        , only : col_cf, col_nf, col_pf  
   use VegetationType        , only : veg_pp                
   use VegetationDataType    , only : veg_cs, veg_cf, veg_ns, veg_nf  
-  use VegetationDataType    , only : veg_ps, veg_pf  
   use topounit_varcon      , only : max_topounits
-  
+  use VegetationDataType    , only : veg_ps, veg_pf  
+  use VegetationDataType    , only : veg_ps, veg_pf
+  use elm_varctl            , only : use_cn, use_fates
+  use FatesConstantsMod      , only : hlm_harvest_area_fraction
+  use FatesConstantsMod      , only : hlm_harvest_carbon
+
+  !
   ! !PUBLIC MEMBER FUNCTIONS:
   implicit none
   private
   save
   public :: dynHarvest_init    ! initialize data structures for harvest information
-  public :: dynHarvest_interp  ! get harvest data for current time step, if needed
-  public :: CNHarvest          ! harvest mortality routine for CN code
-  
+  public :: dynHarvest_interp_harvest_types  ! get harvest data for each harvest type, if needed
+  public :: CNHarvest          ! harvest mortality routine for CN code (non-FATES)
+  !
   ! !PRIVATE MEMBER FUNCTIONS:
   private :: CNHarvestPftToColumn   ! gather pft-level harvest fluxes to the column level
   
@@ -46,15 +51,37 @@ module dynHarvestMod
   ! the pftdyn data
   type(dyn_file_type), target :: dynHarvest_file ! information for the file containing harvest data
 
-  ! Define the underlying harvest variables
+  ! Define the underlying input harvest variables
+  ! these are hardcoded to match the LUH input data
+  ! these are fraction of vegetated area harvested, split into five harvest type variables
+  ! in next iteration (v3) change these to fraction of grid cell
+  ! HARVEST_VH1 = harvest from primary forest
+  ! HARVEST_VH2 = harvest from primary non-forest
+  ! HARVEST_SH1 = harvest from secondary mature forest
+  ! HARVEST_SH2 = harvest from secondary young forest
+  ! HARVEST_SH3 = harvest from secondary non-forest
+  ! note that these are summed for CNHarvest in harvest(:) and applied to total forest area only
+  ! the individual category rates are passed to fates in harvest_rates(:,:)
+
+  ! for FATES: capacity for passing harvest data in units of carbon harvested per year (per grid cell) has been added
+  ! but these data are not yet included in the input file
+  ! the code here can be changed to wood_harvest_units = harvest_carbon to pass carbon data to FATES if:
+  !  the carbon data are in the same input variables as listed below
+  !  and the carbon units in the input file match that expected by FATES
+
   integer, public, parameter :: num_harvest_vars = 5
   character(len=64), public, parameter :: harvest_varnames(num_harvest_vars) = &
-       [character(len=64) :: 'HARVEST_VH1', 'HARVEST_VH2', 'HARVEST_SH1', 'HARVEST_SH2', 'HARVEST_SH3']
+       [character(len=64)  :: 'HARVEST_VH1', 'HARVEST_VH2', 'HARVEST_SH1', 'HARVEST_SH2', 'HARVEST_SH3']
+  ! the units flag must match the units of harvest_varnames
+  ! set this here because dynHarvest_init is called after alm_fates%init
+  ! this flag is accessed only if namelist do_harvest is TRUE
+  integer, public          :: wood_harvest_units = hlm_harvest_area_fraction
   
   type(dyn_var_time_uninterp_type) :: harvest_vars(num_harvest_vars)   ! value of each harvest variable
 
-  real(r8) , allocatable   :: harvest(:,:) ! harvest rates
-  logical                  :: do_harvest ! whether we're in a period when we should do harvest
+  real(r8) , allocatable, public   :: harvest_rates(:,:) ! category harvest rates (d1) in each gridcell (d2)
+
+  logical, private         :: do_harvest ! whether we're in a period when we should do harvest
   !---------------------------------------------------------------------------
 
 contains
@@ -65,10 +92,7 @@ contains
     ! !DESCRIPTION:
     ! Initialize data structures for harvest information.
     ! This should be called once, during model initialization.
-     
-    ! This also calls dynHarvest_interp for the initial time
-    
-    ! !USES:
+    !
     use elm_varctl            , only : use_cn
     use dynVarTimeUninterpMod , only : dyn_var_time_uninterp_type
     use dynTimeInfoMod        , only : YEAR_POSITION_START_OF_TIMESTEP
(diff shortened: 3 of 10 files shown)


components/clm/src/dyn_subgrid/dynHarvestMod.F90
```diff
@@ -94,64 +119,64 @@ subroutine dynHarvest_init(bounds, harvest_filename)
 
     SHR_ASSERT_ALL(bounds%level == BOUNDS_LEVEL_PROC, subname // ': argument must be PROC-level bounds')
```
rgknox, line 120: was going to check if this code is thread-safe seeing how a global (harvest_rates) is being allocated, but both the initialization and the interp code asserts this is the proc level. 
check

components/clm/src/dyn_subgrid/dynTimeInfoMod.F90
```diff
@@ -144,7 +144,7 @@ end subroutine set_current_year
   ! ======================================================================
 
   !-----------------------------------------------------------------------
-  subroutine set_current_year_get_year(this)
+  subroutine set_current_year_get_year(this, offset)
```
rgknox, line 147: this looks fine to me, seeing how the original functionality of this routine is preserved, and should not impact existing calls to it, pinging @bishtgautam

components/clm/src/main/clmfates_interfaceMod.F90
```diff
@@ -294,7 +301,7 @@ subroutine ELMFatesGlobals()
 
         call set_fates_ctrlparms('parteh_mode',ival=fates_parteh_mode)
 
-        if(is_restart()) then
+        if(is_restart() .or. nsrest .eq. nsrBranch) then
```
rgknox, line 304: thanks for these updates @aldivi

rgknox (approved):
changes look good

bishtgautam:
@aldivi The first comment of each PR is usually a concise description that is copied/pasted when merging the PR to next/master. Could you edit your first comment to be concise? You can always add additional comments in the PR that describe the changes in more detail. Also, please indicate in your first comment if this PR BFB or non-BFB. An example of the first comment of a PR is https://github.com/E3SM-Project/E3SM/pull/3691. Thanks.

aldivi:
ELM code (for harvest control and FATES harvest control) has been updated so that ELM input harvest can be passed to FATES to drive FATES harvest code. The ELM namelist variables do_harvest and fdyndat can now be used to drive FATES harvest. The FATES code has not been modified, as it has already been updated for this purpose.

This has been tested for all harvest happening in single day in FATES with a global 4x5 data set (landuse.timeseries_4x5_hist_simyr1850-2015_200311.nc) on cori and also a single point dataset (landuse.timeseries_1x1_brazil_hist_simyr1850-2015_c200310.nc) on lobata. Running this new code with ELM harvest gives the same results as running ELM harvest with the corresponding master. Running with FATES harvest applies the correct harvest fraction within FATES, as represented by the amount of land moving from primary to secondary land, and does change biomass/harvest results compared to ELM harvest because FATES is a different vegetation model. Furthermore, FATES has a single-source daily mortality routine that selects the largest daily mortality source, and in some grid cells general mortality is greater than logging mortality, even with a single-day harvest, such that in some cells the prescribed harvest is not applied. FATES fire was not enabled for these tests, as it also competes with the selection of the dominant daily mortality source.

It is important to note that the landuse.timeseries files (for fdyndat) contain model year harvest data in the year+1 data element. This is a legacy format that puts the desired pft and harvest data in the same labelled year because they had been created and read in at the same time. For example, pft data and harvest data are read in at the beginning of year y, and since the pft data represented the state at the beginning of y+1, while the harvest data represented the rate during year y, both were stored in as year y+1 to simplify the storage and reading process. The pft y+1 and harvest y data are also processed at the same time when the land files are created. This storage/labelling of these data was not updated when elm was updated to separate the pft and harvest reading and processing.

So additional modifications to the harvest read and get year functions were required to ensure that the correct harvest year data are read in.

I have also updated some controls to allow branch runs to work.

bishtgautam:
Thanks @aldivi

jenniferholm:
@aldivi 
Thanks for that last description, it was very helpful!

bishtgautam:
@aldivi This PR results in unexpected non-BFB DIFFs in the three non-FATES tests (see https://my.cdash.org/viewTest.php?onlyfailed&buildid=1965321). I will revert this PR from next for now. Please fix the DIFFs by generating baselines and 
comparing your code modifications against those baselines. 

FYI, here are some [notes](https://gist.github.com/bishtgautam/31a73b933e2492698bccfc7b55acf01e) to generate/compare baselines.

aldivi:
@bishtgautam Yes, you will get non-BFB diffs with those tests, which are the transient ones. I couldn't figure out exactly which landuse.timeseries files were used, but if they are like the others they include changing pfts and harvest.
I mischaracterized this as a BFB with ELM-only. This is because I first tested it without the year offset of the input harvest data to ensure that the calculation I moved did not change anything (the sum of the 5 harvest categories). But with the year offset the harvest amounts are different, so there is no BFB behavior overall.

bishtgautam:
@aldivi Few questions for you:
1. Does this PR result in non-BFB for a transient run with SP mode?
2. Does this PR result in non-BFB for a transient run with BGC mode?
3. In this PR, does FATES work with and without BGC? 
4. Does this PR result in non-BFB for transient FATES run with/without BGC?

(cc: @rgknox)

aldivi:
@bishtgautam 

FATES does not run with SP mode, and in this version it does not run with BGC either (FATES and CN cannot both be active). I think that recently FATES has been enabled with the ELM BGC modes, but this PR would have to be merged with the current code to find out how well these play together. It is likely that some conditions will have to be set so that both CN harvest and FATES harvest are not happening at the same time.
So in answer to your questions:
Does this PR result in non-BFB for a transient run with SP mode?
Not applicable
Does this PR result in non-BFB for a transient run with BGC mode?
Not applicable
In this PR, does FATES work with and without BGC?
Only without BGC.
Does this PR result in non-BFB for transient FATES run with/without BGC?
non-BFB for transient FATES run without BGC

bishtgautam:
@aldivi But your PR is causing non-BFB results for (1) transient SP mode; (2)transient BGC mode. see https://my.cdash.org/viewTest.php?onlyfailed&buildid=1965321

aldivi:
Ah, the first two questions were for non-FATES runs.
Yes, the results are non-BFB in BGC mode.
They should also be non-BFB in SP mode.

rljacob:
This can't be merged until after we make the v2.0 tag for watercycle.  attn: @golaz

bishtgautam:
Apart from @rljacob's [comment above](https://github.com/E3SM-Project/E3SM/pull/4024#issuecomment-804383538) regarding the water cycle experiment, I would also like @kvcalvin to approve this PR in regards to non-BFB BGC changes.

aldivi:
@rljacob Does the v2 watercycle use transient runs with harvest? Will there be an additional v2 tag, or is that it?
The issue is that the input harvest data are not currently being applied to the correct year. Fixing this year offset is what causes the non-BFB behavior. So the question is whether we are fine with continuing to apply the input harvest data to the following year, or if we want it to be applied to the correct year. Over the course of a simulation, this may not make a huge difference.

rljacob:
There will be another tag for BGC.  Its up to @golaz if he wants this for WC or not.

kvcalvin:
@aldivi : Just to clarify - all of the non-BFB differences are because wood harvest (and the associated LULC) is shifted by a year (to the correct year). Is that correct?

aldivi:
@kvcalvin Before the time shift of harvest prescription I checked the monthly values over a year for conversion flux, wood C, tot veg c, tot som C, tot ecosys C, and atmospheric co2, and the differences were zero (on the same machine, cori). The only code calculation that was changed in ELM was that the harvest type sum was moved from read-in to the harvest routine, which didn't change the answers.
There is also the possibility of separating the time shift from the fates harvest update.

kvcalvin (approved):

thorntonpe (approved):

rljacob:
Notes from telecon:  this is ready to go after v2.0 tag.

rljacob:
@bishtgautam do you think this should be rebased ?

bishtgautam:
Okay

rljacob:
@bishtgautam what is up with this branch?

aldivi:
@rljacob @bishtgautam I have been trying to get the rebased version to run, but continue to encounter bugs, although they do not appear to be related to the new harvest code. Some I have fixed to continue (like a bad pft memory access when the crop model is off), but I have recently just hit a memory failure during land model initialization that needs some effort to track down.
The three automated test failures were with transient land tests that will fail because of the shifted year for input harvest data. Were you running these tests on the non-rebased version? If not, how did you get the rebased version to run?

rljacob:
The tests were run after this branch was merged to next which, after conflicts are resolved, is as good as a rebase.  However that was using next from 27 days ago.

aldivi:
@rljacob ok. This should be the same then because gautam's rebase was 28 days ago. I am going to try with debug mode turned off, although it seems weird that it would run in regular mode if debug mode has a bunch of memory errors associated with the regular code.

aldivi:
@rljacob @bishtgautam Is there still a land-developer test suite? I tried several different land suite names and none of them worked on cori, although last year i did run a land test suite. I want to create some baselines of the pre-rebased commit and compare. I had to make a few code edits to get it running, but I think it is ok. I have to do another run to check the year 2000 harvest because the data was not set correctly.

bishtgautam:
The test suite is `e3sm_land_developer`

aldivi:
@rljacob @bishtgautam 
Ok. I have completed testing on cori. The fates harvest code works correctly, and I did two different baseline comparisons against the oct 7 2021 commit in this branch (just prior to the merge of the fates harvest code). I set the harvest year offset to zero and all of the e3sm_land_developer tests passed. I then set the harvest year offset to 1 in order to read the correct year of input data, and all of the tests passed except for the three transient tests (which is expected):
ERS.ne11_oQU240.I20TRELM.cori-knl_intel
ERS.f19_g16.I20TRGSWCNPRDCTCBC.cori-knl_intel.elm-ctc_f19_g16_I20TRGSWCNPRDCTCBC
ERS.f19_g16.I20TRGSWCNPECACNTBC.cori-knl_intel.elm-eca_f19_g16_I20TRGSWCNPECACNTBC

I did have to make a few mods to get it running, including updating a new namelist check in the configuration, allowing for a non-crop-model pft check if fates is active, and updating the array shape parameter for the harvest data. I have pushed these changes to this rebased branch.
I also ran this with and without debug mode.

bishtgautam:
@aldivi what is the default value for harvest year offset?

rljacob:
@aldivi do you have access to compy?  You should check the exact machine/compiler the ERS test originally failed on.

aldivi:
@rljacob @bishtgautam I have tried to generate the same baselines on compy using both the intel and pgi compilers and they fail to build.
The default offset for the harvest data is 1 year.

aldivi:
@bishtgautam @rljacob It turns out that git does not save/restore submodule tags in the main repo associated with particular branches/commits, and the FATES tag changed in the past month. I realigned the oct 7 2021 master commit (07ff719) with its original FATES tag (36e60bb) and was able to generate baselines on compy. I then made the same comparisons as on cori, with the same results: without the input data year shift all land developer tests pass, and with the shift the three transient tests differ (as expected). I also ran the FATES harvest code on compy and the max difference in secondary forest fraction per grid cell between compy and cori is less than 3x10^-12.

aldivi:
@rljacob @bishtgautam Just checking in on the status of this merge, before master gets too far ahead again. Thanks.

rljacob:
@bishtgautam you can merge this when you're ready.

jenniferholm (approved):
Approve these changes. Thanks

jenniferholm:
@bishtgautam - is there time to resolve these conflicts and get this PR approved, updated, and merged into master?  This PR has been waiting for a year now, and we need to get these harvesting data code changes integrated.  Other people are waiting on these updates so that they can perform their work.

bishtgautam:
Later today, I will work on merging this to the next while resolving conflicts.

bishtgautam:
@rljacob This PR is non-BFB for transient WC runs. The maint-2.0 hasn't been created. Is it okay to merge this PR?

rljacob:
Since the BGC leads want this, yes.

bishtgautam:
@aldivi @jenniferholm, This is now merged to next.

cc: @rgknox @ckoven

aldivi:
Thanks, Gautam!
Alan

--
Alan Di Vittorio, Ph.D.
Lawrence Berkeley National Laboratory
Climate and Ecosystem Sciences Division
***@***.***

rljacob:
Will this PR cause this test to diff?  SMS.ne30pg2_r05_oECv3.F20TR.compy_intel.eam-wcprod

aldivi:
@rljacob that specific test was not run (only land developer suite was run), but it looks like it is a transient case (F20TR), in which case it will diff

aldivi:
@bishtgautam Any transient run with harvest enabled will diff because the driving harvest data are shifted. This test diffs:
ERS.ne11_oQU240.I20TRELM.cori-knl_intel
The regular I20TRELM compset uses ELM%SP. Does this test use the same compset?

rljacob:
That test, ERS.ne11_oQU240.I20TRELM, is **not** diffing on chrysalis, compy or mappy.

components/elm/src/dyn_subgrid/dynSubgridControlMod.F90
```diff
@@ -231,12 +231,8 @@ subroutine check_namelist_consistency
     end if
 
     if (dyn_subgrid_control_inst%do_harvest) then
-       if (.not. use_cn) then
-          write(iulog,*) 'ERROR: do_harvest can only be true if use_cn is true'
-          call endrun(msg=errMsg(sourcefile, __LINE__))
-       end if
-       if (use_fates) then
-          write(iulog,*) 'ERROR: do_harvest currently does not work with use_fates'
+       if (.not. (use_cn .or. use_fates)) then
```
bishtgautam, line 234: @aldivi Here is the code that ensures harvest is turned on only when BGC is also turned on. Thus, your modifications to the harvest code should not change SP mode. Do you agree?

aldivi:
@bishtgautam yes, this is correct that harvest is shut off if not using bgc or fates. I thought this was the case, but that particular compset made me pause. Thank you for checking the control logic.

PR #4032, opened 2021-01-13
EAM: Improve definitions and checks related to time stepping

ambrad:
This PR has no runtime ramifications other than output in the case of a time step misspecification.

* Expand the definitions of dt_remap_factor, dt_tracer_factor in the XML.
* Add full details to an error message in a time step consistency check routine.
* Separately, add a typedef that permits the SL transport code to build and run properly on CPU when Kokkos provides the Cuda execution space. If Cuda is not enabled, the type is equivalent to the existing type, so there is no runtime change from the code change.

[BFB]

### DIFF
--- a/components/eam/bld/namelist_files/namelist_definition.xml
+++ b/components/eam/bld/namelist_files/namelist_definition.xml
@@ -5618,16 +5618,20 @@ Dynamics timestep.  Replaces se_nsplit.
 
 <entry id="dt_tracer_factor" type="integer" category="se"
        group="ctl_nl" valid_values="" >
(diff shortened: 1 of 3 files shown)


AaronDonahue (approved):
@ambrad , thanks for putting this in.  I can't speak to the CUDA ifdef change, but I approve the changes for the XML file and the revised error message.

mt5555 (approved):

AaronDonahue:
This has been pushed to next

AaronDonahue:
@ambrad , this has been merged to master.  I'll let you take care of removing the branch on your fork.

PR #4033, opened 2021-01-13
Consolidate v2 atm parameter values into a common default file

wlin7:
Consolidate v2 atm parameter values into a common default file
    
Optimized atm parameter values for v2 were initially set in run script,
then migrated to use case files for corresponding compsets. This PR
further consolidates the common settings among the v2 compsets into
namelist_defaults_eam.xml from use case files.
    
This PR also renames physics package name used for v2 during
configuration and runtime to 'default'.
    
[NBFB] Tests using active atm component except for wcprod and v1cmip6
will be NBFB as many of the new default parameter values incorporated
from use case files have not yet been used in the nightly test cases.

[NML] nml diff for all tests involving eam except for the wcprod tests


wlin7:
Many changes in this PR, for the cleanup of v1ac and v1af series of compsets and the associated use_case files are already included in #4007 that has been merged.

rebased with master 3f00614 and forced pushed to remote.

xuezhengllnl (approved):
I checked the changes in these 12 files.  Everything looks good to me.

wlin7:
Thank you @xuezhengllnl  for the review.
@rljacob , I just unassigned you and assigned it to myself. 
F-case tests will diff, v1cmip6 tests will pass. Look to merge tomorrow.

wlin7:
Merged to next. Prior to merging, WCYC1850 verified BFB between master and this PR.

wlin7:
Pending merging to master. Tests using F-EAMv1* and F-MMF* compsets failed due to namelist settings. Fixes are on the way.  Other tests returned are mostly as expected: wcprod tests PASS, v1cmip6 pass except on chrysalis which exists before this PR. A few other tests using FC5AV1C-L and A_WYCL1850(S) reported FAIL but are actually diff. To get further feedback from. IG (@rljacob )

Not to bless the diffs and not to revert the PR from next for now.

wlin7:
Reverted from next. @rljacob determined the restart tests (ERS and ERP) have failed. No clue yet on the cause. Could be some block of codes triggered by the new namelist defaults (like #4038). Looking into it.

wlin7:
It has been determined that the clubb option to suppress supersaturation, clubb_ipdf_call_placement  =  2, is causing restart run to be NBFB. This option is being used in latest v2 finalization settings.

xuezhengllnl:
> It has been determined that the clubb option to suppress supersaturation, clubb_ipdf_call_placement = 2, is causing restart run to be NBFB. This option is being used in latest v2 finalization settings.

Hi @wlin7, when clubb_ipdf_call_placement = 2, the place of stat variables output in clubb/advance_clubb_core_module.F90 is different from clubb_ipdf_call_placement = 1 or 3. Not sure if this difference has anything with the NBFB issue for now.

singhbalwinder:
@wlin7 : Is it just the restart which is NBFB or the model with different PE layout or model in general is also NBFB?

worleyph:
I just tried a simple ne4_ne4 F case with this flag set, and it was BFB with respect to load balancing (in one 96x1 PE layout on Chrysalis). It did change answers compared to not using the flag, so something was happening by setting it.

wlin7:
Nice to hear that, @worleyph . My test with current master (previously with the branch for this PR) on compy, ERS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L.compy_pgi.eam-thetahy_sl_pg2, is still giving NBFB when clubb_ipdf_call_placement=2. But BFB when the value is 1 (current default) or 3.

worleyph:
I also ran with @ambrad 's suggestion, adding '-init=snan,arrays' for all of EAM. This did not fail, so this does not appear to be an issue with uninitialized space. (The run was BFB with a run without '-init=snan,arrays' as well.) This was ne4_ne4 F case on Chrysalis using 96x1, and a initial run compiled with DEBUG.

worleyph:
I'll try on Compy as well, and also look at '-init=snan,arrays' for a restart run.

wlin7:
I think it is encouraging the test passes on chrysalis. WIth the value '2', it is expected to passed some variables in and out the subroutine between call. I was concerned those could have been lost in a restart run. Now that chrysalis passes, the logic looks like is there.

golaz:
@forsyth2 and I verified that results were BFB for 'S' and 'L' layouts on compy and chrysalis last week; so it's probably only non-BFB for restarts. I should have verified for restart as well.

worleyph:
@wlin7 , I have not tested whether restarts work on chrysalis or not. I have just been looking at reproducibility and uninitialized space. (I did not know where you observed the error - should have tried to reproduce your failure on Chrysalis before I started. I'll try this soon.)

worleyph:
restarts appear to fail on Chrysalis as well (just comparing atm.log using a 1 day initial run, a 1 day continuation run, and a 2 day initial run) The continuation run atm.log differed from the day 2 timesteps in the 2 day initial run.

singhbalwinder:
There could be many reasons but one could be that one or more variables need to go into the restart file (like @wlin mentioned) or it could be that a field is written but it is not read during the restart run appropriately.

wlin7:
> There could be many reasons but one could be that one or more variables need to go into the restart file (like @wlin mentioned) or it could be that a field is written but it is not read during the restart run appropriately.

It is more like the case now. @golaz , yes, this is only a restart NBFB. It was only tested when this PR was merged to next and making the option '2' the new default. @xuezhengllnl is also examining the codes, starting from #3814 when this parameter is introduced. But I am afraid the restart NBFB very likely has existed before that PR.

The restart NBFB issue is independent of this PR. Please continue the investigation with issue #4064.  Thanks.

wlin7:
Revert the revert in next, then remerged the PR to next.
F-MMF and RCECMIP tests will fail due to runtime segmentation faults.

wlin7:
Merged to master. 

Blessing diffs on chrysalis (43), sandiatos3 (43), and mappy (10). wcprod tests not affected. 
5 fails due to this PR: F-EAMv1-AQP1, F-EAMv1-RCEMIP, and 3 F-MMF tests. They will be resolved separately.

wlin7:
compy has 10 diffs due to this PR. Being blessed.

Fix (#4086) is in for the fails of AQP1, RCEMIP and MMF tests. They will be diffs and need to be blessed.

wlin7:
RCEMIP and MMF tests ran ok and the diffs are blessed.

wlin7:
Bless a final diff due to 4033, EAMv1-AQP1.clubb-only. The test exposed a bug that is fixed in #4090 -- only having an impact when not using ZM.

After #4090, all issues  originated from 4033 have been resolved.

PR #4042, opened 2021-01-20
Fixes a NBFB issue when clubb_use_sgv and zmconv_tp_fac are used

### LINKED ISSUE #4038: BFBFLAG=True still gives NBFB results with different PElayout on compy
BFBFLAG=True is expected to give BFB reproducibility with different PElayout. It is not the case with a recent master running on compy and intel compiler.

The problem can be produced with following code and configuration
 ```
          master hash c0b0c779bbf6728153765e02a37d6057f6a73cd8
           compset: A_WCYCL1850S_CMIP6
           grid:         ne30pg2_r05_EC30to60E2r2-1900_ICG        
```
The simulations were done using the following two run scripts (the 1ts one uses 90 nodes -- PE=L, the 2nd 46 nodes -- PE=M, all MPI)
```
 /qfs/people/linw288/E3SM/Cases/prod/scripts/run.20201225.alpha5_59_fallback.piControl.EC30to60E2r2.compy.csh
/qfs/people/linw288/E3SM/Cases/prod/scripts/run.20210114M.alpha5_59_fallback.piControl.EC30to60E2r2.compy.csh
```

singhbalwinder:
The model produces NBFB answers when run with phys_loadbalance =0
and phys_loadbalance=2 with the following settings in atm_in:

clubb_use_sgv          = .true.
zmconv_tp_fac          = 2.0D0

The reason is that the "tpert" variable is used in zm_convr subroutine
directly without collecting the "gathered" columns of "tpert". To
fix this, a new variable "tpertg" is declared which has the "gathered"
columns of "tpert". "tpertg" is being used instead of "tpert" for
cloud properties calculations.

Fixes #4038

[Non-BFB] for only `ERS_Ld5.ne4_ne4.FC5AV1C-L.<machine-compiler>.eam-gust_param` test

### DIFF
--- a/components/eam/src/physics/cam/zm_conv.F90
+++ b/components/eam/src/physics/cam/zm_conv.F90
@@ -499,10 +499,11 @@ subroutine zm_convr(lchnk   ,ncol    , &
 
    real(r8) hu_nm1g(pcols,pver)        !songxl 2014-05-20
 
-   real(r8) rprdg(pcols,pver)           ! wg gathered rain production rate
+   real(r8) rprdg(pcols,pver)          ! wg gathered rain production rate
    real(r8) capeg(pcols)               ! wg gathered convective available potential energy.
    real(r8) tlg(pcols)                 ! wg grid slice of gathered values of tl.
-   real(r8) landfracg(pcols)            ! wg grid slice of landfrac  
+   real(r8) landfracg(pcols)           ! wg grid slice of landfrac  
+   real(r8) tpertg(pcols)              ! wg grid slice of gathered values of tpert (temperature perturbation from PBL)
 
    integer lclg(pcols)       ! wg gathered values of lcl.
    integer lelg(pcols)
@@ -796,6 +797,7 @@ subroutine zm_convr(lchnk   ,ncol    , &
       maxg(i) = maxi(ideep(i))
       tlg(i) = tl(ideep(i))
       landfracg(i) = landfrac(ideep(i))
+      tpertg(i) = tpert(ideep(i))
    end do
 !
 ! calculate sub-cloud layer pressure "thickness" for use in
@@ -848,7 +850,7 @@ subroutine zm_convr(lchnk   ,ncol    , &
                maxg    ,j0      ,jd      ,rl      ,lengath , &
                rgas    ,grav    ,cpres   ,msg     , &
                pflxg   ,evpg    ,cug     ,rprdg   ,limcnv  , &
-               landfracg, hu_nm1g, tpert)   !songxl 2014-05-20
+               landfracg, hu_nm1g, tpertg)   !songxl 2014-05-20
                                             !PMA adds tpert to the calculation
 !
 ! convert detrainment from units of "1/m" to "1/mb".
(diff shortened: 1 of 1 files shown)


singhbalwinder:
I ran an SMS.ne4_ne4.FC5AV1C-L baseline test and this PR is BFB w.r.t. master.

worleyph:
What is the relationship between il2g and lengath? I see tpertg set for (i=1,lengath) but declared pcols, and then in cldprp, tpertg is accessed using (i = 1,il2g).

singhbalwinder:
@worleyph : Looks like this code has evolved over time so variables are not used consistently. `lengath` is being sent as an argument in `cldprp` and received as `il2g`. So `lengath` and `il2g` are same for `cldprp` subroutine.

I think there may be times when `lengath` can be equal to `pcols`, so all variables are dimensioned `pcols` (which is the max).

worleyph:
Thanks. Was just concerned about potentially uninitialized variables.

whannah1:
It really seems that the whole gathering step is unnecessary. I'd love to rewrite the ZM code and include removing this step. I doubt there would be any performance difference by just passing "ideep" into cldprp() and using it as a conditional within a ncol loop.

rljacob:
Can someone explain what "gather" means within zm?

whannah1:
The "gather" thing has to do with only doing the deep convection calculations over the columns where deep clouds are actually "firing". This must have made a difference on older machines, but it seems pointless now. The data for the columns with deep convection are copied to these "gathered" arrays, which are then sent to the cldprp() routine. I guess removing the copying might actual make it a tad faster.

worleyph:
so this is really the "pack/unpack" of other parts of the physics, done to improve vectorization. Depending on how computationally intensive it is, there may still be an advantage on some processor architecture. 

And this was not a BFB issue, but a real bug, correct?

singhbalwinder:
Yes, this was a real bug. `tpert` was being used directly without being "packed" or "gathered".

worleyph:
> Yes, this was a real bug. tpert was being used directly without being "packed" or "gathered".

but being accessed as if it had been gathered.

rljacob:
And why was this bug not noticed until we set clubb_use_sgv = .true.,
zmconv_tp_fac = 2.0D0 ?

whannah1:
@rljacob is zmconv_tp_fac normally 0?

polunma (approved):

singhbalwinder:
Like @whannah1 mentioned, without these new namelist options, tpert is always zero.

wlin7:
Merged to next. The changes are very clear, additional review not necessary.

worleyph:
>Awesome detective work here! Thanks @singhbalwinder, @worleyph, @wlin7.

and @whannah1

whannah1:
@worleyph sorry, you're right, but since it's a somewhat random mixup of a small temperature perturbation I still don't think it's worth investigating for systematic climate impacts.

worleyph:
science question then, which I am not qualified to answer :-). Would note that tpert for columns that should be ignored are being used. Don't know if tpert for these columns is special for any reason (and should not be used). Sounds like this is not the case.

singhbalwinder:
I just realized that there is a test (`ERS_Ld5.ne4_ne4.FC5AV1C-L.sandiatoss3_intel.eam-gust_param`) which uses these namelist parameters. I am assuming that test will fail, so I am labeling this PR `NBFB`.

wlin7:
Thanks for the tip, @singhbalwinder . This PR affects gust_param test as well as tests using the updated A_WCYCL1850S_CMIP6, F20TRC5-CMIP6, and F2010SC5-CMIP6 (following the merged of #4008). SInce this PR went in together with #4007,  requests for blessing are submitted together. See more at the  section for #4007.

Merged to master.

PR #4043, opened 2021-01-21
Add gnu modules for intel-mpi on Chrysalis

amametjanov:
Add gnu modules for intel-mpi on Chrysalis

[BFB]


dqwu (approved):
Tested on Chrysalis and it looks good.

PR #4044, opened 2021-01-23
Update HOMME cmake files on Chrysalis

amametjanov:
Update HOMME cmake files on Chrysalis

[BFB]


amametjanov:
`homme_integration` tests build for [intel, gnu] x [intel_mpi, open_mpi]. Default runs with intel + intel_mpi PASS.
Runs with gnu (gcc/9.2.0, also gcc/8.2.0 on anvil) seg-fault in `thetah-sl-test11conv-r1t2-cdr20`.

PR #4048, opened 2021-01-24
homme_tool updates to support interpolating NE1024 data

mt5555:
-  fix integer overflow in inputfile PIO decomp
-  skip requested variables that are not in the input file
-  skip 2D string variables (not working in SCORPIO's pio_copy_var )
-  add ability to thread the interpolating setup loop.   Run 16 MPI x 8 Threads on Cori, to reduce I/O time.  

[BFB]

### DIFF
--- a/components/homme/src/share/interpolate_mod.F90
+++ b/components/homme/src/share/interpolate_mod.F90
@@ -1166,7 +1166,7 @@ contains
     err=0
     do j=1,nlat
        
-!$OMP PARALLEL DO NUM_THREADS(omp_get_max_threads()) SHARED(err) PRIVATE(i,sphere,ii,cart,sphere2_xyz,sphere_xyz) reduction (max: err)
+!$OMP PARALLEL DO NUM_THREADS(omp_get_max_threads()) PRIVATE(i,sphere,ii,cart,sphere2_xyz,sphere_xyz) reduction (max: err)
        do i=1,nlon
           sphere%r=1
           sphere%lat=lat(j)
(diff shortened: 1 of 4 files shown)


rljacob:
@mt5555 is this ready to merge?

mt5555:
pushed to next.  self integrating since this should only impact standalone homme tool code.

components/homme/src/share/interpolate_mod.F90
```diff
@@ -1177,9 +1181,12 @@ subroutine setup_latlon_interp(elem,interpdata,par)
              sphere2_xyz = spherical_to_cart( ref2sphere(cart%x,cart%y,     &
                   elem(ii)%corners3D,cubed_sphere_map,elem(ii)%corners,elem(ii)%facenum ))
              sphere_xyz = spherical_to_cart(sphere)
+!$OMP CRITICAL
              err=max(err,distance(sphere2_xyz,sphere_xyz))
```
ambrad, line 1185: Could use a reduction instead of critical region. In any case, I recommend you compute `distance` outside of the critical region, then do `err=max(err,d))` inside, thus minimizing the computations in the critical region.

ambrad (approved):
I made a suggest for increased threading efficiency, but that's optional.

dqwu:
@mt5555 
On anlworkstation, building test SMS_P12x2.ne4_oQU240.A_WCYCL1850.anlworkstation_gnu.allactive-mach_mods
with OpenMPI has the following error:
```
components/homme/src/share/interpolate_mod.F90:1169:129:

 !$OMP PARALLEL DO NUM_THREADS(omp_get_max_threads()) SHARED(err) PRIVATE(i,sphere,ii,cart,sphere2_xyz,sphere_xyz) reduction (max: err)
                                                                                                                                 1
Error: Symbol ‘err’ present on multiple clauses at (1)
```
Do you know how to fix it?

mt5555:
sorry about that. looks like we just remove the shared(err).  I'm testing right now...

PR #4052, opened 2021-01-26
Update mappy to gcc-8.1.0 to match summit

jgfouca:
Causes roundoff diffs (only on mappy) for:

```
DIFF SMS_D_Ln5.ne4_ne4.FC5AV1C-L.mappy_gnu (phase RUN)
DIFF SMS_R_Ld5.ne4_ne4.FSCM5A97.mappy_gnu.eam-scm (phase RUN)
```

[non-BFB]


jgfouca:
Skipping integration.

PR #4060, opened 2021-01-27
Update mappy serial env

jgfouca:
[BFB]


jgfouca:
Skipping integration testing.

PR #4074, opened 2021-02-04
Add support for ne30pg2_oRRS18to6v3 fully-coupled configurations

jonbob:
Adds a new grid alias for ne30pg2_oRRS18to6v3, to allow testing of the v1 high-res ocean grid with the v2 low-res atm. The mapping files were already in place for a trigrid setup with lnd/rof on r0125.

### DIFF
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -1945,6 +1945,16 @@
       <mask>WC14to60E2r3</mask>
     </model_grid>
 
+    <model_grid alias="ne30pg2_oRRS18to6v3">
+      <grid name="atm">ne30np4.pg2</grid>
+      <grid name="lnd">ne30np4.pg2</grid>
+      <grid name="ocnice">oRRS18to6v3</grid>
+      <grid name="rof">r05</grid>
+      <grid name="glc">null</grid>
+      <grid name="wav">null</grid>
+      <mask>oRRS18to6v3</mask>
+    </model_grid>
+
     <model_grid alias="ne30_r05_oECv3">
       <grid name="atm">ne30np4</grid>
       <grid name="lnd">r05</grid>
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -1945,6 +1945,7 @@
       <mask>WC14to60E2r3</mask>
     </model_grid>
 
+<!-- WARNING: THIS GRID COMBINATION IS NOT SCIENTIFICALLY SUPPORTED -->
     <model_grid alias="ne30pg2_oRRS18to6v3">
       <grid name="atm">ne30np4.pg2</grid>
       <grid name="lnd">ne30np4.pg2</grid>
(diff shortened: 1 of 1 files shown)


jonbob:
@vanroekel - the only issue I saw when I setup a test case was the mapping files for ne30pg2<=>oRRS18to6v3 that were already there were "mono" both ways and for all map types (states, fluxes, etc). I'm not sure that's correct, especially for states going from a low-res atm to the high-res ocn.

rljacob:
How about adding this to the e3sm_hi_res suite?

jonbob:
Thanks @rljacob - if they find this configuration useful, I'm happy to add it to the test suite

rljacob:
@vanroekel is this still needed?

maltrud:
@vanroekel i think it shouldn't be supported.  there are lots of issues with running low res atm with high res ocean.

vanroekel:
@maltrud I'm surprised by that.  Justin Small does this type of thing all the time.  I think this is something I'd like to pursue eventually, but it is very low priority.  I'm not sure I'd say it's needed. I'm happy to close this @rljacob if you prefer and work with @jonbob and @maltrud to make this work eventually.

maltrud:
@vanroekel i should have been clearer.  i think it shouldn't be supported, not that it shouldn't be possible.  Justin knows what he's doing--i worry a bit that a non-expert might use such a setup without knowing the possible issues.

vanroekel:
I see now, thanks @maltrud for clarifying.  I would like to run this configuration eventually so think it should be added as I'd prefer not run experiments with this configuration from a personal branch.  Two thoughts
1. @jonbob or @rljacob is there some way to flag this configuration as not supported and experimental only?
2.  we close the PR and try work on this to make a configuration we are happy with and then do a PR

I'm open to either, but do think we need to add this if we are going to run it in the future.

maltrud:
@vanroekel flagging as 'unsupported' sounds good to me.

vanroekel:
that option is my preference as well, if it is possible.

jonbob:
We don't currently have any mechanism for that. We can put comments like that around the entries in the config_grids file, but it's not clear if that would so what we want. If someone queried the available grids, it would come out in the list like any other configuration

rljacob:
How about you add a comment, and we can add more of a disclaimer to query_config that not everything listed is scientifically supported.

jonbob:
Sounds good. I'll do that and ask you for a review

jonbob:
@rljacob - by the way, the mapping files were already specified from an old commit back in July 2020 but are not on the inputdata repo server. I did find them on NERSC so I'll have to bring them over as well

jonbob:
@rljacob - could you please take a look at what I added? I put a comment in config_grids.xml but also found there is a "support" attribute for the model_grid XML definition, though we haven't used it. It does look like it's been used in the ups config_grids?

rljacob (approved):

vanroekel (approved):
This looks good to me.  Thanks for adding the comments.

jonbob:
merged to master

PR #4079, opened 2021-02-05
Add ESMT to SAM++

xyuan:
Add ESMT functionality to the C++ SAMXX code. Comparisons were made with the standalone model and the full E3SM-MMF run and show good agreement. 

[BFB]


whannah1 (approved):

whannah1:
A note about the ifdefs (since I know @brhillman  will yell at me for that) - While I'm 99.9% sure this will be a feature that we want on all the time for 2D CRMs, I think we need to do a little more investigating (particularly in regard to surface friction issues and the new variance transport). Once we decide that we do want ESMT to be on by default I'll be happy to create another PR that removes the ifdefs in favor of control via a namelist variable in both the fortran and C++ CRM code. I assume that future PR will also be able to remove the "USE_ORIG_FFT" blocks.

mrnorman:
Has this been tested on the GPU?

xyuan:
> Has this been tested on the GPU?

Yes, I compared with CPU and GPU runs, and the Fortran CPU runs as well.

mrnorman:
@whannah1 , I agree, now that we've vetted the new FFT routines, we can get rid of USE_ORIG_FFT. Thanks @xyuan for doing this. Please let me know offline how painful it was to code in C++ YAKL and what I can improve to make it more user-friendly.

brhillman:
Is this BFB for the default config?

whannah1:
@brhillman yes, ESMT is still disabled by default and this shouldn't affect non-ESMT configurations.
And since we're still controlling this with ifdefs there won't be any namelist diff.

brhillman:
Adding the BFB flag then.

mrnorman (approved):

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #4085, opened 2021-02-09
Remove _ICG ocn/ice grids, instead key off compset string SOI

jonbob:
Makes changes to the mpas seacice and ocean components to key off the "SOI" compset string (for spunup ocean-ice) instead of using the grid _ICG suffix to determine which initial files to use. It also removes the _ICG grids from config_grids.xml and renames tests using _ICG in the automated suites.

[BFB] - though it will change some test names


jonbob:
One issue that needs to be resolved has to do with ne120np4-oRRS18to6v3 configurations. There were two different sets of mapping files, depending on whether or not it used the _ICG version of the ocean grid:
```
    <gridmap atm_grid="ne120np4" ocn_grid="oRRS18to6v3">
      <map name="ATM2OCN_FMAPNAME">cpl/gridmaps/ne120np4/map_ne120np4_to_oRRS18to6v3_mono.20200702.nc</map>
      <map name="ATM2OCN_VMAPNAME">cpl/gridmaps/ne120np4/map_ne120np4_to_oRRS18to6v3_mono.20200702.nc</map>
      <map name="ATM2OCN_SMAPNAME">cpl/gridmaps/ne120np4/map_ne120np4_to_oRRS18to6v3_mono.20200702.nc</map>
      <map name="OCN2ATM_FMAPNAME">cpl/gridmaps/ne120np4/map_oRRS18to6v3_to_ne120np4_mono.20200702.nc</map>
      <map name="OCN2ATM_SMAPNAME">cpl/gridmaps/ne120np4/map_oRRS18to6v3_to_ne120np4_mono.20200702.nc</map>
    </gridmap>

    <gridmap atm_grid="ne120np4" ocn_grid="oRRS18to6v3">
      <map name="ATM2OCN_FMAPNAME">cpl/gridmaps/ne120np4/map_ne120np4_to_oRRS18to6v3_aave.170111.nc</map>
      <map name="ATM2OCN_VMAPNAME">cpl/gridmaps/ne120np4/map_ne120np4_to_oRRS18to6v3_conserve.170111.nc</map>
      <map name="ATM2OCN_SMAPNAME">cpl/gridmaps/ne120np4/map_ne120np4_to_oRRS18to6v3_conserve.170111.nc</map>
      <map name="OCN2ATM_FMAPNAME">cpl/gridmaps/oRRS18to6v3/map_oRRS18to6v3_to_ne120np4_aave.170111.nc</map>
      <map name="OCN2ATM_SMAPNAME">cpl/gridmaps/oRRS18to6v3/map_oRRS18to6v3_to_ne120np4_aave.170111.nc</map>
    </gridmap>
```
 It seems like one compset assumes it uses docn:

    <model_grid alias="ne120_r0125_oRRS18to6v3" compset="(DOCN|XOCN|SOCN|AQP1)">
      <grid name="atm">ne120np4</grid>
      <grid name="lnd">r0125</grid>
      <grid name="ocnice">oRRS18to6v3</grid>
      <grid name="rof">r0125</grid>
      <grid name="glc">null</grid>
      <grid name="wav">null</grid>
      <mask>oRRS18to6v3</mask>
    </model_grid>
```
though in general I'm not sure we want to use all "mono" mapping files

**Comment by rljacob** [comment]
I'm glad to see "ICG" is being removed from the grid alias.  But I have a different suggestion for how to handle specifying a spunup ocean for the initial condition..

Initial conditions are supposed to be in the first part of the longname:  1800, 20TR, 2015, etc.   So if the initial condition includes a spunup ocean, we could have 1800SO, 20TRSO,  2015SO.    If you need both ocean and sea ice, that would be 1800SOI,  20TRSOI1, 2015SOI, etc.

The stuff after the "%" in  "MPASO%" is supposed to be for compile-time configuration options.

**Comment by jonbob** [comment]
Thanks @rljacob - let me look at that and see how hard that would be to implement

**Comment by jonbob** [comment]
@rljacob - as far as I can tell, right now the difference between compsets like A_WCYCL1850 and A_WCYCL1850S is that ocn/ice set %SPUNUP. Do any of the other components key off the "S" suffix? I think it's pretty straightforward  for the ocn and ice components to change IC's and the corresponding namelist/streams settings from the compset name instead of %SPUNUP, but it does mean reworking a bunch of compsets. But I also think when the "S" suffix compsets were added, it was to indicate the %SPUNUP and maybe nothing else? If that's true, do you want to change them to "SO" and "SOI" or just let "S" indicate spunup ocn/ice?

**Comment by rljacob** [comment]
I'm not sure what you're asking.  The compset alias can be left as  A_WCYCL1850 and A_WCYCL1850S to distinguish between cold-start and spunup ICs.   But the model buildnml's should key of the longname, not the alias.  We can leave "S" in the alias as meaning ocean AND ice are spun-up (if that's the current meaning).  I don't think anything else uses the "S".

The longname for A_WYCL1805S should change from:
1850_EAM%CMIP6_ELM%SPBC_MPASSI%SPUNUP_MPASO%SPUNUP_MOSART_SGLC_SWAV
to
1850SOI_EAM%CMIP6_ELM%SPBC_MPASSI_MPASO_MOSART_SGLC_SWAV

**Comment by rljacob** [comment]
Also we're going to remove "A_" from those aliases so they will all have to change anyway if you want to change "S" to "SOI"

**Comment by jonbob** [comment]
Thanks @rljacob - I think I understand then. Let me work on implementing it that way

**Comment by jonbob** [comment]
@rljacob - do you prefer "SOI" to "S", for clarity? Or do our users understand the naming convention as is?

**Comment by rljacob** [comment]
Lets make it "SOI".

**Comment by jonbob** [comment]
Aye captain

**Comment by jonbob** [comment]
@rljacob - OK, I've changed to add "SOI" to the compset longname and removed %SPUNUP, which was not used for compile-time settings. Please take a look whenever you have a chance

**Comment by jonbob** [comment]
@vanroekel - I wanted to check with you about how we allow for different spunup initial conditions going forward. This came up doing the WC/RRM call yesterday when we discussed this PR. Right now we end up with separate grids for this situation, so that the mpas scripts can differentiate between EC30to60E2r2 and EC30to60E2r2-1900, for example. I can think of a couple of options and open to any suggestions:

1. make use of the new "SOI" compset longnames, so that  ```1850SOI_EAM%CMIP6_ELM%SPBC_MPASSI_MPASO_MOSART_SGLC_SWAV```
 and
 ```1850SOI-1900_EAM%CMIP6_ELM%SPBC_MPASSI_MPASO_MOSART_SGLC_SWAV```
are used to differentiate;
2. decide on one best spunup IC for each grid and have that be the default for E3SM, and then assume that anyone who wants to test other ICs is skilled enough to understand how to swap them in; or
3. keep it as-is

Any thoughts?

**Comment by golaz** [comment]
@jonbob and @vanroekel : my suggestion would to entirely stop using grid resolution as a differentiator for initial condition, i.e. we should have a single grid named "EC30to60E2r2" and get rid of the one named "EC30to60E2r2-1900".

I like the suggestion to differentiate IC in the SOI portion of the long compset name:

 `1850SOI-1900_EAM%CMIP6_ELM%SPBC_MPASSI_MPASO_MOSART_SGLC_SWAV` 

'WCYCL1850' can be an alias to that long compset. Users who need alternate IC can refer to the long compset name directly.

**Comment by golaz** [comment]
@vanroekel : should we include the new ICs as part of this PR?

**Review by rljacob — APPROVED** [review]

**Comment by jonbob** [comment]
@golaz - I agree that we should get rid of the duplicate grid names. I'm not sure we can make the compset longname have the "-1900" because that would mean all grids would have to have that spinup accessible, especially for the WCYCL1850 that is likely to be the v2 default. Instead, my preference is to have a single default spinup for every grid and let any mpas developers who want to test other ICs do so using streams files changes, This also gets rid of possible confusion for users who may not understand all the workings that go along with compset longnames?

**Comment by rljacob** [comment]
Don't forget to update the description.

**Comment by rljacob** [comment]
What is the difference between "1900" and "ICG" ?

**Comment by jonbob** [comment]
It's not really a difference between 1900 and ICG, because there is a 1900_ICG. It's a difference in what forcing conditions were used to spinup the models.

**Comment by golaz** [comment]
@rljacob : initially, we only had "ICG" to refer to the spun-up initial conditions. Eventually, we started using alternate initial conditions, which became "ICG-1900".

**Comment by golaz** [comment]
@jonbob and @rljacob : so we cannot have a single grid and different initial conditions?

The grid would be `EC30to60E2r2` and then the long compset names using different initial conditions:

- `1850SOI-1900_EAM%CMIP6_ELM%SPBC_MPASSI_MPASO_MOSART_SGLC_SWAV`
- `1850SOI2_EAM%CMIP6_ELM%SPBC_MPASSI_MPASO_MOSART_SGLC_SWAV`
- `1850SOI3_EAM%CMIP6_ELM%SPBC_MPASSI_MPASO_MOSART_SGLC_SWAV`
- ...

**Comment by vanroekel** [comment]
@jonbob and @golaz I would prefer Jon's option (2) I'd like to explore ways to make the IC specification a part of the namelist instead of the stream in the future.  I think this is possible, but beyond this PR.  This would allow an IC exploration that is much more robust and reproducible.

In my quest to spin up the ocean again to be unified between WC14 and EC30to60, I'm going to advocate we drop the 1900 EN4 for now.  I've noticed it is far fresher in the Arctic and the temperature is not significantly different.  Given our problems with surface fresh biases, I think we should move back to the PHC, which is what I'm trying now.  This gets rid of the -1900 issue in the long name.  I need to finish the spin up to confirm though.

**Comment by rljacob** [comment]
So what does the "1900" mean?   Something about the year 1900 ?

**Comment by vanroekel** [comment]
The 1900 signifies that the initial condition is from a state estimate of the ocean state for the year 1900.  From the Met office

**Comment by jonbob** [comment]
Thanks @vanroekel and @golaz for the clarification

**Comment by jonbob** [comment]
@golaz - sorry for the confusion. I think we could work it so we can have a single grid with several different ICs, and have that specified by a compset longname. It's the short compset alias that would only point at a single one of these.

**Comment by jonbob** [comment]
Passed e3sm_developer on anvil

**Comment by jonbob** [comment]
@wlin7, @vanroekel, @darincomeau - Chris would like us to get this merged soon, if you have time to review. Thanks!

File: components/mpas-ocean/cime_config/buildnml [hunk 1]
```diff
@@ -74,184 +82,210 @@ def buildnml(case, caseroot, compname):
...
-        ic_date += '171031'
-        ic_prefix += 'oEC60to30v3wLI60lev'
         decomp_date += '170328'
         decomp_prefix += 'mpas-o.graph.info.'
         restoring_file += 'sss.monthlyClimatology.PHC2_salx_040803.oEC60to30v3wLI.nc'
         analysis_mask_file += 'masks_SingleRegionAtlanticWTransportTransects_EC60to30v3wLI_171116.nc'
+        ic_date += '171031'
+        ic_prefix += 'oEC60to30v3wLI60lev'
+        if ocn_ic_mode == 'spunup':
+            ic_date += '171116'
```
darincomeau, line 126: @jonbob I just tried testing this (`oEC60to30v3wLI`) grid with a spunup compset and the `ic_date` from the if statement is concatenating the previous defined `ic_date`, so the resulting file it's looking for is `oEC60to30v3wLI60levoEC60to30v3wLI60lev.restart_theta_year26.171031171116.nc`. This logic probably affects other grids as well.

darincomeau: the `ic_prefix` is getting double-concatenated as well

darincomeau (changes requested):

jonbob:
Thanks @darincomeau for the testing. I had run some of the automated suites but needed someone to test something more like this. I'll make changes tomorrow and we can test again

jonbob:
@darincomeau - could you please try again? I'm not sure why we originally had "+=" for setting things like ic_date, but I've replaced those with plain ol' "="

vanroekel (approved):
approved by visual inspection

jonbob:
@darincomeau - there's still a problem. It gets the ocn/ice files fine but I'm afraid it's making EAM unhappy. The config_component.xml for EAM has this in it:
```
  <entry id="CAM_NML_USE_CASE">
    <type>char</type>
    <valid_values></valid_values>
    <default_value>UNSET</default_value>
    <values>
      <value compset="1850_EAM"        >1850_cam5</value>
      <value compset="20TR_EAM"        >1850-2005_cam5</value>
      <value compset="AMIP_EAM"        >1850-2005_cam5</value>
      <value compset="RCP2_EAM"        >2006-2100_cam5_rcp26</value>
....
```
We can't even use 1850.*_EAM because they have something special for 1850S:
```
      <value compset="1850S_EAM.*CMIP6"  >1850S_cam5_CMIP6</value>
```
sigh...

jonbob:
@rljacob - I'm not sure how to proceed here with the compset string SOI. That isn't consistent with EAM's config_component file, which assumes compsets are named something like 1850_EAM. I could add an optional pattern to match like 
```
      <value compset="1850(?:SOI)?_EAM.*CMIP6"  >1850_eam_CMIP6</value>
```
or just duplicate the entries? Otherwise we can move the SOI to a different place in the compset longname? Any ideas?

darincomeau:
I came here to say the `ne30_oECv3wLI` grid issue is fixed, pointing to expected ocn/ice conditions for the `A_WCYCL1850S_CMIP6` (spun-up) and `A_WCYCL1850-DIB-ISMF_CMIP6` (cold-start) ICs, but seems like there are larger issues now...

Also, a cryo-specific issue that predates this PR is that we don't have 'S' compsets for the cryo configurations (e.g. `A_WCYCL1850S-DIB-ISMF_CMIP6`). I can put together a separate PR to add these once this PR is resolved.

jonbob:
Thanks @darincomeau - yes, fix one thing and then on to the next

rljacob:
Adding an optional pattern sounds like a good quick fix.   I think the right thing to do is also rename those namelists.  What do you think @wlin7 ?

golaz:
Need input from @wlin7 here. The eventual goal is to be able to completely remove those 1850**S** EAM compsets (wtih the S standing for different meanings: Special, Spin-up, etc..).

jonbob:
@wlin7 - any thoughts? I'll start adding the optional pattern-matching, but I would like to get this merged this week if possible.

wlin7:
Sorry for the late response from me on eam use_case. We intend to only use 1850_eam_CMIP6.xml  for all 1850 cases. So the one @jonbob initially considered (but had to rethink because of the concern for 1850S_EAM*) can be used, which is
`<value compset="1850.*_EAM.*CMIP6"  >1850_eam_CMIP6</value>`

The other entry, as follows, should be removed.
`<value compset="1850S_EAM.*CMIP6"  >1850S_cam5_CMIP6</value>`

wlin7:
Hi @jonbob , nice to know the xml parser can recognize optional pattern. I am not familiar with this (non-capturing group re), though,  does the following

`<value compset="1850(?:SOI)?_EAM.*CMIP6"  >1850_eam_CMIP6</value>`

captures all of 1850_EAM*, 1850S_EAM*, 1850SOI_EAM*, 1850SOI1_EAM*?  Or just 1850_EAM* and 1850SOI_EAM*? I am confused with what the question mark can do in this regular expression. I expect it be the latter.

On the file modifications, can you please change  to only apply to the lines of 1850_eam_CMIP6, 20TR_eam_CMIP6 and 2010_eam_CMIP6? The others will be removed or need to be updated anyway. Thanks.

wlin7 (approved):
@jonbob , the changes to eam and elm settings with the optional pattern are good. Ok if opt to not scale back the changes to the config_component.xml files. I can resolve the conflicts when merging the PR for additional cleanup.

components/mpas-ocean/cime_config/buildnml
jonbob: @darincomeau - can you take a look again at this and review? I think we need to get it merged as soon as possible

darincomeau (approved):
Approved based on testing cryo configurations.

jonbob:
merged to next, starting sanity tests and integration/prod tests that have new names

jonbob:
merged to master, and requested new baselines for renamed tests. Also opened an issue for BGCEXP tests that now fail

PR #4086, opened 2021-02-09
Fix for aero_model when chem is none after phys name change

wlin7:
A conditional block in build-namelist for prescribed_aero_model was
not updated accordingly after the change for phys name (#4033). This
causes prescribed_aero_model to be none and several tests that use
modal model as default to fail during runtime. The tests affected
are  AQP1, RCEMIP and MMF tests

[BFB] no impact on other tests. AQP1, RCEMIP, MMF tests will still be
      BFB but they are the results of #4033.


wlin7:
Hi @whannah1 , @brhillman , not that this simple fix needs a review. Just to share with you where the change. was missed. It only affects tests then use -chem none.

wlin7:
Merged to master.

PR #4087, opened 2021-02-09
HOMMEXX: fix some valgrind errors when unit-testing the remap functor

bartgol:
There was an error when setting the scratch mem buffer inside the `RemapStateProvider` struct, which caused a view to be a bit further down in the buffer than originally planned. This was completely harmless in actual runs, since the mem buffer was much bigger than `RemapStateProvider` needed, so the view still ended inside the allocated memory. However, when unit testing the remap functor, we allocated only enough memory to fit the remap needs, so the error in the view creation was exposed. This PR fixes this error.

Additionally, our remap unit testing was not comparing the tracers after the remap. I added those checks as well.

[bfb]

### DIFF
--- a/components/homme/src/theta-l_kokkos/cxx/RemapStateProvider.hpp
+++ b/components/homme/src/theta-l_kokkos/cxx/RemapStateProvider.hpp
@@ -85,13 +85,10 @@ struct RemapStateProvider {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


ambrad (approved):

oksanaguba:
in next

PR #4090, opened 2021-02-10
Fix an issue that uses dcape related codes when ZM is not used

wlin7:
A block of codes in tphysac access pbuf for T_STAR and Q_STAR
that are used for dcape calculation. The block should only
be invoked when deep_scheme = 'ZM' but proper condition was not
applied, causing the test F-EAMv1-AQP1.eam-clubb_only to fail as
it sets deep_scheme = 'off'. The fix now adds a condition on
deep_scheme. The above test only became an issue after #4033 which
set default zmconv_trigdcape_ull = .true.

[BFB] for all tests, while F-EAMv1-AQP1.eam-clubb_only would report
      diff due to #4033.

### DIFF
--- a/components/eam/src/physics/cam/physpkg.F90
+++ b/components/eam/src/physics/cam/physpkg.F90
@@ -1477,6 +1477,8 @@ subroutine tphysac (ztodt,   cam_in,  &
     real(r8), pointer, dimension(:,:) :: t_star   ! temperature
     real(r8), pointer, dimension(:,:) :: q_star   ! moisture
 
+    character(len=16)  :: deep_scheme             ! Default set in phys_control.F90
+
     ! Debug physics_state.
     logical :: state_debug_checks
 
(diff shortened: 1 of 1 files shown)


wlin7:
This is a straightforward fix for the failure of SMS_Ld1.ne4_ne4.F-EAMv1-AQP1.chrysalis_intel.eam-clubb_only.  The test exposes an issue of the dcape implementation, which will cause runs to fail when not using ZM as deep_scheme.

The branch has been tested BFB for tests.

components/eam/src/physics/cam/physpkg.F90
```diff
@@ -1776,7 +1779,7 @@ subroutine tphysac (ztodt,   cam_in,  &
 
     ! DCAPE-ULL: record current state of T and q for computing dynamical tendencies
     !            the calculation follows the same format as in diag_phys_tend_writeout
-    if (do_zmconv_dcape_ull .or. do_zmconv_dcape_only) then
+    if (deep_scheme .eq. 'ZM' .and. (do_zmconv_dcape_ull .or. do_zmconv_dcape_only)) then
```
singhbalwinder, line 1782: Since this is a string comparison, do we need `trim(adjustl(deep_scheme)) .eq. 'ZM'`?

singhbalwinder (approved):
You have run the tests, so I am not sure if my suggestion will make a difference or not but I think it is better to use `trim` function for string comparisons.

components/eam/src/physics/cam/physpkg.F90
wlin7: Good point. But I think it is ok. It will force users to specify just 'ZM' as the scheme name. (it is currently set so as default). Existing codes, say in convect_deep.F90, also use it as is without trimming. Thanks for the quick review, @singhbalwinder .

wlin7:
Merged to next.

wlin7:
Merged to master.

PR #4092, opened 2021-02-11
Update support for EC30to60 and WC14, including consistent ocn/ice settings

vanroekel:
Updates support for the EC30to60 and WC14to60 ocn/ice meshes. A few parameters were inconsistent between these resolutions that should not have been, so they are unified here. Also fixes an incorrect mapping filename and adds default layouts on E3SM platforms.

[NML] only for configurations that use these meshes
[non-BFB] only for configurations that use these meshes

### DIFF
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -4488,7 +4488,7 @@
 
     <gridmap ocn_grid="WC14to60E2r3" rof_grid="JRA025">
       <map name="ROF2OCN_ICE_RMAPNAME">cpl/cpl6/map_JRA025_to_WC14to60E2r3_smoothed.r150e300.200929.nc</map>
-      <map name="ROF2OCN_LIQ_RMAPNAME">cpl/cpl6/map_JRA025_to_WC14to60E2r2_smoothed.r150e300.200929.nc</map>
+      <map name="ROF2OCN_LIQ_RMAPNAME">cpl/cpl6/map_JRA025_to_WC14to60E2r3_smoothed.r150e300.200929.nc</map>
     </gridmap>
 
     <gridmap ocn_grid="oQU480" rof_grid="r05">
--- a/cime_config/allactive/config_pesall.xml
+++ b/cime_config/allactive/config_pesall.xml
@@ -9937,4 +9937,70 @@
       </pes>      
     </mach>
   </grid>
+  <grid name="oi%EC30to60E2r2|oi%oEC60to30v3">
+    <mach name="anvil|chrysalis">
+      <pes compset="DATM.+MPASO" pesize="any">
+        <comment>none</comment>
+        <ntasks>
+          <ntasks_atm>324</ntasks_atm>
+          <ntasks_lnd>324</ntasks_lnd>
+          <ntasks_rof>324</ntasks_rof>
+          <ntasks_ice>320</ntasks_ice>
+          <ntasks_ocn>640</ntasks_ocn>
+          <ntasks_glc>1</ntasks_glc>
+          <ntasks_cpl>324</ntasks_cpl>
+        </ntasks>
+        <nthrds>
+          <nthrds_atm>1</nthrds_atm>
+          <nthrds_lnd>1</nthrds_lnd>
+          <nthrds_rof>1</nthrds_rof>
+          <nthrds_ice>1</nthrds_ice>
+          <nthrds_ocn>1</nthrds_ocn>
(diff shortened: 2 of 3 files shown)


cime_config/config_grids.xml
```diff
@@ -4488,7 +4488,7 @@
 
     <gridmap ocn_grid="WC14to60E2r3" rof_grid="JRA025">
       <map name="ROF2OCN_ICE_RMAPNAME">cpl/cpl6/map_JRA025_to_WC14to60E2r3_smoothed.r150e300.200929.nc</map>
-      <map name="ROF2OCN_LIQ_RMAPNAME">cpl/cpl6/map_JRA025_to_WC14to60E2r2_smoothed.r150e300.200929.nc</map>
```
vanroekel, line 4491: @jonbob while not exactly a parameter setting, I fixed this here as it was a setting that was causing WC14 failure for JRA.  If you'd prefer me to separate this to a different PR, I'm happy to do so.  It wasn't clear to me if this was a clear cut different PR or not.

jonbob: Oh, maybe we should make it separate? Or else we can make this PR more for generalized EC30t60 and WC14 support, and include this and the pe-layout changes as well?

vanroekel: Why don't we go with the latter and put in the pe layouts here instead of a ton of extremely small PRs.

jonbob: OK, we can change the title of this PR to reflect general support. I'll work on adding the layout update here today, and maybe we can get 4091 in first?

vanroekel: Sounds like a good plan to me

jonbob:
layout changes tested on anvil and compy

jonbob:
This should wait for 4091 to clear, since they will both impact the WC14to60 test

vanroekel:
@jonbob  this one is now ready to go. I've updated the WC14 defaults on del2 and del4 to be consistent with the EC30to60E2r2

jonbob:
merged to next, passed sanity testing

jonbob:
merged to master and expected NML/DIFFs blessed

xylar:
This merge appears to have broken the `PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.anvil_intel` test case:
```
    Case dir: /lcrc/group/e3sm/ac.xylar/scratch/anvil/PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.anvil_intel.20210219_041702_07d6p4
    Errors were:
        env_batch.xml appears to have changed, regenerating batch scripts
        manual edits to these file will be lost!
        
        wget failed with output:  and errput --2021-02-19 04:26:31--  https://web.lcrc.anl.gov/public/e3sm/inputdata/ice/mpas-cice/oEC60to30v3wLI/mpas-cice.graph.info.170905.part.320
        Resolving web.lcrc.anl.gov (web.lcrc.anl.gov)... 140.221.70.30
        Connecting to web.lcrc.anl.gov (web.lcrc.anl.gov)|140.221.70.30|:443... connected.
        HTTP request sent, awaiting response... 404 Not Found
        2021-02-19 04:26:31 ERROR 404: Not Found.
        
        svn export failed with output:  and errput svn: E170000: URL 'https://svn-ccsm-inputdata.cgd.ucar.edu/trunk/inputdata/ice/mpas-cice/oEC60to30v3wLI/mpas-cice.graph.info.170905.part.320' doesn't exist
        
        ERROR: Could not find all inputdata on any server

```
Presumably, we just need to generate and upload the required graph partition?

xylar:
I'm happy to make a proper bug report if that's helpful.

jonbob:
@xylar - please do make an issue about that. Or I can, if you're busy. I am a bit confused why this happened, but @akturner had reported the same thing last night. I think the quick fix is to add a 320-pe graph file for that mesh -- and then make sure I understand what happened...

jonbob:
@xylar - OK, I made and staged the 320 graph file for oEC60to30v3wLI. I'll track down the issue and make sure nothing else is broken.

xylar:
Okay, so it sounds like I don't need to make an issue?  Thanks for staging the missing file so quickly.

jonbob:
I think it's OK. I understand where the change came from, and I think we have all the other necessary graph files in place.

xylar:
Great, I reran the test (with a different branch) and everything worked as expected so I agree that things look good.

PR #4097, opened 2021-02-16
Add switch for expanded snowpack model (use_extrasnowlayers)

jonbob:
Several developments necessary to add a switch that will toggle on an expanded snowpack model valid for use on ice sheets. This switch, called "use_extrasnowlayers," will also toggle on a new snowcapping routine (adapted from CLMv5) that removes snow from the bottom of the column, rather than diverting fresh new snow to an export term, and brings in improvements to the snowpack (and firn) densification scheme. The switch is set to false by default, as this PR is designed to maintain BFB reproducibility unless a user sets the switch to true. The switch, even when true, does not currently change the maximum number of snow layers (equal to 5) nor the maximum snowpack mass (equal to 1 m SWE). This PR provides the control logic and software necessary to use up to 16 total snow layers, but the user will still have to increase these values (nlevsno and h2osnomax) in the clm_varpar.F90 and clm_varcon.F90 files (before, there was no way to increase the number of snow layers above 5). Results from these developments, including simulated firn density profiles, are being submitted for publication in Geoscientific Model Development.

Moved from #3723 (with all credit to @amschne) because of the number of changes in rebasing/updating

[BFB]


stephenprice:
@bishtgautam, it is worth noting here that (pretty sure) you already reviewed all of the snowpack mods (from @amschne) previously.

stephenprice:
@jonbob -- noting here that we had also discussed whether or not this line:

https://github.com/E3SM-Project/E3SM/blob/efc2385dbcdb019872c59cd60d3707bcad8365eb/cime_config/config_grids.xml#L1613
... should be changed to:
`<grid name="atm">ne30pg2</grid>`

(Note that the current PR does not include any changes to /cime_config/config_grids.xml)

bishtgautam (approved):

cime_config/tests.py
```diff
@@ -110,6 +110,7 @@
         "tests" : (
             "ERS_P480_Ld5.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF",
             "PEM_P480_Ld5.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF",
+            "SMS.ne30_oECv3_gis.IGCLM45_MLI.clm-extrasnowlayers",
```
bishtgautam, line 113: Should the compset use `ELM` instead of `CLM45`?

jonbob: Thanks @bishtgautam - yes, that's absolutely correct. I had another set of commits for this PR but decided it was suffering from PR-creep, so I hadn't checked all the changes in that last commit. But this was one of them... I'll fix that and commit right now

components/elm/cime_config/config_compsets.xml
```diff
@@ -758,7 +758,8 @@
 
   <compset>
     <alias>IGELM_MLI</alias>
-    <lname>2000_DATM%QIA_ELM%SP_SICE_SOCN_MOSART_MALI%SIA_SWAV</lname>
+    <!--lname>2000_DATM%QIA_ELM%SP_SICE_SOCN_MOSART_MALI%SIA_SWAV</lname-->
+    <lname>2000_DATM%QIA_ELM%SP_SICE_SOCN_MOSART_MALI_SWAV</lname>
```
stephenprice, line 762: This means the MALI component will be active in the build, correct? I can't remember what we decided about that. For now, I'm happy either way. If it's not a pain for any reason, let's leave it like this so the build gets exercised automatically when we run this compset.

components/elm/cime_config/testdefs/testmods_dirs/elm/extrasnowlayers/user_nl_elm
```diff
@@ -0,0 +1 @@
+use_extrasnowlayers = .true.
```
stephenprice, line 1: Is this setting a default value in the default namelist? If so, it should probably be .false. 
We want the old snowpack model to be used by default for now or there will likely be non BFB changes wherever there is snow on the sfc. We can easily toggle this on as we test / improve on the SMB.

components/elm/cime_config/config_compsets.xml
jonbob: Yes, the MALI build will be active for this compset, but not the one with %SIA. Having the two compsets allows us to control the build that way

components/elm/cime_config/testdefs/testmods_dirs/elm/extrasnowlayers/user_nl_elm
jonbob: This is in a testmod so we can turn it on by using a configuration like:
```
SMS.ne30_oECv3_gis.IGELM_SIA.elm-extrasnowlayers
```
but otherwise have it off by default. We could also turn it on using user_nl_elm by hand

stephenprice: thanks for the clarification

jonbob:
@stephenprice - as for the atm grid naming, I didn't add that configuration to this PR so as you said, it doesn't matter. I'd like to get this PR merged and then add a new all-active grid for the gis1to10km MALI setup in a separate PR. Then we can go through all of that, plus decide if we want a bigrid or trigrid setup (or both). The next CIME update will bring in the trigrid fixes and may make that an option

stephenprice:
@jonbob -- This all sounds good to me. Let's get this one done. FYI, I started a new branch last night w/ (I think) most of the necessary changes for supporting a few of the other all-active bigrid configs. we want to try out (1/2 deg and 1/4 deg). We can talk about that further this afternoon.

stephenprice:
@jonbob -- just in from Mauro: we should be pointing to this Albany lib path in the config_compilers file:

`/global/cfs/cdirs/e3sm/software/albany-trilinos/albany-install-2021-01-05/`

Apparently the lib path that's in there now is the one w/ kokkos issues.

jonbob:
merged to next

jonbob:
reverted from next due to testing issues in debug mode

jonbob:
test merge into next passes e3sm_developer on anvil, except ERS_Ld20.f45_f45.IELMFATES.anvil_intel.elm-fates which has known issues

jonbob:
after reverting/reverting-the-revert/re-merging, a test merge with next passes e3sm developer testing and is BFB with master except:
* DIFF ERS_Ld20.f45_f45.IELMFATES.chrysalis_intel.elm-fates (phase BASELINE)
* DIFF SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.chrysalis_intel (phase BASELINE)
* DIFF SMS.f09_g16_a.IGELM_MLI.chrysalis_intel (phase BASELINE)
Note that all three of those tests have current DIFFs on the dashboard. Comparing the files that actually DIFF with the most recent NextIntegration results, the three failing tests have not changed answers -- just those tests have DIFFs that have not been blessed

jonbob:
reverted the original revert and re-merged to next

jonbob:
The BFB check with master was comparing chrysalis_intel. Overnight testing shows it's not BFB for compy_pgi but the cprnc output shows round-off level differences. I ran a couple of the tests by hand and they are BFB in debug mode, so it's something in optimization.

jonbob:
merged to master and request issue to bless DIFFs

PR #4098, opened 2021-02-16
Fix an issue in nudging code 

zhangshixuan1987:
-- In current nudging.F90 the character length of the nudging file name is defined as 80,
   when the nudging file name is longer than 80 characters, the nudging code can not
   get the full name for the nudging file. In this way, nudging will not be applied.
   To avoid this problem, code changes are made to extend the character length of
   the nudging file name to 256.

-- Note that: when the filename can not be identified by the nudging.F90, nudging will
   complain that the nudging data does not exist, and nudging will be automatically
   switched off. This can be only identified by checking the atm.log file. 

[BFB]

### DIFF
--- a/components/cam/src/physics/cam/nudging.F90
+++ b/components/cam/src/physics/cam/nudging.F90
@@ -413,7 +413,7 @@ module nudging
   logical::         Nudge_File_Present=.false.
   logical::         Nudge_Initialized =.false.
   character(len=cl) Nudge_Path
-  character(len=cs) Nudge_File,Nudge_File_Template
+  character(len=cl) Nudge_File,Nudge_File_Template
   integer           Nudge_Times_Per_Day
   integer           Model_Times_Per_Day
   real(r8)          Nudge_Ucoef,Nudge_Vcoef
@@ -1675,7 +1675,7 @@ contains
 
    integer :: cnt3(3)               ! array of counts for each dimension
    integer :: strt3(3)              ! array of starting indices
-   character(len=cs) :: nudge_file1
+   character(len=cl) :: nudge_file1
    integer :: n, n_cnt, ncid1, ind
    integer :: timesiz               ! size of time dimension on dataset
    integer :: Year, Month, Day, Sec
@@ -2388,7 +2388,7 @@ contains
 
    integer :: cnt4(4)               ! array of counts for each dimension
    integer :: strt4(4)              ! array of starting indices
-   character(len=cs) :: nudge_file1
+   character(len=cl) :: nudge_file1
    integer :: n, n_cnt, ncid1, ind
    integer :: Year, Month, Day, Sec
 
@@ -3118,7 +3118,7 @@ contains
   ! local variable  
   integer :: YMD3, YMD4, Nudge_Next1_Sec, Nudge_Next1_Year, &
              Nudge_Next1_Month, Nudge_Next1_Day, istat
-  character(len=cs)       :: nudge_file
+  character(len=cl)       :: nudge_file
 
   YMD3 = (Nudge_Next_Year*10000) + &
          (Nudge_NEXT_Month*100) + Nudge_Next_Day
--- a/components/cam/src/physics/cam/nudging.F90
+++ b/components/cam/src/physics/cam/nudging.F90
@@ -414,7 +414,7 @@ module nudging
   logical::         Nudge_ON          =.false.
   logical::         Nudge_File_Present=.false.
   logical::         Nudge_Initialized =.false.
-  logical::         Nudge_Allow_Missing_File = .true.  
+  logical::         Nudge_Allow_Missing_File = .false.  
   character(len=cl) Nudge_Path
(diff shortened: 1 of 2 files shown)


rljacob:
Should it be fatal error if the nudging file is not found?

zhangshixuan1987:
@rljacob : Actually it is not,  here is the information from my atm.log file:

 _**NUDGING: Reading analyses:
 /global/cscratch1/sd/zhan391/202010_SciDAC_simulation/run/E3SMv1_FC5_NDGDAT_CPL
 _DeepCu_DT0300.ne30_ne30.cori-knl/E3SMv1_FC5_NDGDAT_CPL_DeepCu_DT0300.ne30_ne30
 .cori-knl.cam.h0.0000-01-01-10800.n
 NUDGING: WARNING - analyses file NOT FOUND. Switching
 NUDGING:           nudging OFF to coast thru the gap.**_
 

The nudging will be switched off, and the model will continue to run as a "free-running" simulation.

singhbalwinder:
I think @rljacob is asking if the model should exit with an error message if the nudging file is not found. I agree with @rljacob. I also think the model should exit rather than silently (or with a warning) continuing with the nudging turned OFF. Generally a user would like to run with nudging turned on if a nudging file is mentioned in the namelist.

@zhangshixuan1987 and @kaizhangpnl : If you agree, would you please add an endrun call so that the model exits with an error message if nudging file is not read correctly?

zhangshixuan1987:
@singhbalwinder : I agree that the model should exit rather than silently (or with a warning) continuing with the nudging turned OFF.  I would like to ask @kaizhangpnl  if he has different viewpoints on this. If not, I can add an endrun call and push this change.

tangq:
I believe this fix should also be applied to the master branch besides the maint-1.0 branch.

zhangshixuan1987:
@tangq : yes, I will submit a pull request to the master branch after this.

kaizhangpnl:
Thanks @zhangshixuan1987. The PR looks good to me.

kaizhangpnl (approved):

components/cam/src/physics/cam/nudging.F90
```diff
@@ -1391,10 +1391,7 @@ subroutine nudging_timestep_init(phys_state)
        Nudge_ON=.true.
      else
        Nudge_ON=.false.
-       if(masterproc) then
-         write(iulog,*) 'NUDGING: WARNING - analyses file NOT FOUND. Switching '
-         write(iulog,*) 'NUDGING:           nudging OFF to coast thru the gap. '
-       endif
+       call endrun('NUDGING: Nudging data file NOT FOUND')
```
singhbalwinder, line 1394: I have a minor suggestion just for improving the error message. Can we do something like:
```
use shr_log_mod ,     only: errMsg => shr_log_errMsg
...

character(len=cxx)  :: err_str
....
....

write(err_str,*) 'NUDGING: Nudging data file(',trim(adjustl(Nudge_File)),')  NOT FOUND', errmsg(__FILE__, __LINE__)
call endrun(err_str)
```

You can modify the error message the way you like but I wanted to include the filename so that user knows which file is missing. The error message will also print the F90 file name and the line number where the error occurred.

singhbalwinder (changes requested):

components/cam/src/physics/cam/nudging.F90
zhangshixuan1987: @singhbalwinder:  I will make adjustments following your suggestion and push the changes. Thank you!

kaizhangpnl: Shixuan, I remember the filename will be printed out in nudging_update_analyses_se.

zhangshixuan1987: Hi @kaizhangpnl: I checked the nudging.f90 file again. I think that in Line 1342-1344: 
     if(masterproc) then
      write(iulog,*) 'NUDGING: Reading analyses:',trim(Nudge_Path)//trim(Nudge_File)
     endif
     
These codes are executed before the call of nudging_update_analyses_se(), but I think that you are right that the filename is actually printed out. 

Hi @singhbalwinder: I think that @kaizhangpnl raised a good point. The nudging file name has already been printed out before the endrun.

singhbalwinder: Sounds good! Thanks @zhangshixuan1987 !

rljacob (approved):

zhangshixuan1987:
@singhbalwinder: Hi Balwinder, I have enhanced the code chages following yours and @tangq 's suggestion:

      1. For the endrun call when nudging data is not found, the error message
        is improved to print the F90 file name and line number where
        the error occurred.
    
     2. Add a logical namelist variable "Nudge_Allow_Missing_File" to allow the
        user to decide whether the model should exist or continue to run when
        the nudged data files are not found

singhbalwinder (approved):

PR #4099, opened 2021-02-17
For chrsyalis: initial set of PE layouts for RRMv2

ndkeen:
Adding initial default PE layouts for chrys. For completeness, here are some estimated SYPD achieved, with XS, S, M, L noted ( each of those use 32x2)
 
```
nodes   SYPD
 14      1.8  XS  
 19      2.4
 23      2.6
 37      4.4
 39      4.6  S  
 48      4.9
 60      5.8
 73      7.1
 77      7.8  M 
147     10.2  L
```

[bfb]

### DIFF
--- a/cime_config/allactive/config_pesall.xml
+++ b/cime_config/allactive/config_pesall.xml
@@ -9936,5 +9936,123 @@
         </rootpe>
       </pes>      
     </mach>
+    <mach name="chrysalis">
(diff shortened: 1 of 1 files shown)


ndkeen:
Note here is an example case dir where I have run the S (39-node) case from a run_e3sm script.

`/lcrc/group/e3sm/ac.ndkeen/E3SM_simulations/beta1_01.piControlSI.mfrrmpe.cmmod039b32x2.S.5d.wh0.wr`

PR #4100, opened 2021-02-17
Fix an issue in nudging code

zhangshixuan1987:
This commits addresses two issues in nudging code (PR#4098)

  1. Generally a user would like to run with nudging turned on if a nudging data file
     is mentioned in the namelist. However, current nudging code would allow the model
     to continue to run even if the nudging data file is not found. The code changes are
     made to add an endrun call so that the model exists with an error message if
     nudging file is not found rather than silently (or with a warning) continuing with
     the nudging turned OFF.
  2. The character length of the nudging file name is defined as 80 in current nudging.f90.
     If the nudging file name is longer than 80 characters, the nudging code can not
     get the full name for the nudging file. In this way, nudging will not be applied.
     To avoid this problem, code changes are made to extend the character length of
     the nudging file name to 256.

[BFB]: no impact on other tests.

### DIFF
--- a/components/eam/bld/namelist_files/namelist_definition.xml
+++ b/components/eam/bld/namelist_files/namelist_definition.xml
@@ -62,6 +62,19 @@
        Toggle Model Nudging ON/OFF.
        Default: FALSE
        </entry>
+       
+<entry id="Nudge_Allow_Missing_File" type="logical" category="nudging"
+       group="nudging_nl" valid_values="" >
+       User-defined flag to inform Nudging if all nudged data files exist.
+       1. If .false.: all nudged data files must show up during the
+           whole nudging period. Otherwise, the model will exit if a
+           nudged data file is not found.
+       2. If .true.: user allows the model to continue to run even if a
+          nudged data file is not found. The model will continue to run with the
+          nudging turned off, and only a warning message will be added to the
+          simulation log file.
+       Default: FALSE
+       </entry>
 
 <entry id="Nudge_Path" type="char*256" input_pathname="abs" category="nudging"
        group="nudging_nl" valid_values="" >
--- a/components/eam/src/physics/cam/nudging.F90
+++ b/components/eam/src/physics/cam/nudging.F90
@@ -379,6 +379,7 @@ module nudging
   use cam_abortutils  ,only:endrun
   use spmd_utils  ,only:masterproc
   use cam_logfile ,only:iulog
+  use shr_log_mod, only:errMsg => shr_log_errMsg
   use perf_mod
 #ifdef SPMD
   use mpishorthand
@@ -391,6 +392,7 @@ module nudging
   private
 
   public:: Nudge_Model,Nudge_ON
+  public:: Nudge_Allow_Missing_File
   public:: nudging_readnl
   public:: nudging_init
   public:: nudging_timestep_init
@@ -412,6 +414,7 @@ module nudging
   logical::         Nudge_ON          =.false.
   logical::         Nudge_File_Present=.false.
   logical::         Nudge_Initialized =.false.
+  logical::         Nudge_Allow_Missing_File = .false.  
   character(len=cl) Nudge_Path
   character(len=cl) Nudge_File,Nudge_File_Template
   integer           Nudge_Times_Per_Day
@@ -520,6 +523,7 @@ contains
    integer ierr,unitn
 
    namelist /nudging_nl/ Nudge_Model,Nudge_Path,                       &
+                         Nudge_Allow_Missing_File,                     &
                          Nudge_File_Template,Nudge_Times_Per_Day,      &
                          Model_Times_Per_Day,                          &
                          Nudge_Ucoef ,Nudge_Uprof,                     &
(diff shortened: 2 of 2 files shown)


kaizhangpnl (approved):

golaz:
@rljacob : while we don't need this specifically for the v2 Water Cycle simulation campaign, this PR should still be on the future maint-2.0 branch since some users will want to use the nudging capability with v2.

rljacob:
Ok.
We should think about adding a test with nudging on.

rljacob (approved):

tangq (approved):

rljacob:
The old warning message has me curious:  "analyses file NOT FOUND. Switching nudging OFF to coast thru the gap."  What was the use case where they wanted this behavior?  Was it just a run length that was longer than the nudging period?

tangq:
@rljacob , there are user name list variables to control the start and end dates of nudging, so the code can handle the simulation length longer than the nudging period (a typical application of nudging runs). I guess it is more likely for the cases when the nudging files have missing dates.

Since the nudging files are normally created from analysis/reanalysis files, it's out of our control if there are missing dates. Thinking this a bit more, we may want to add a name list variable for this. For example, nudging_file_all_exist is 1) true, then end run if some nudging files are not found within the nudging period; 2) false, then continue with a warning message (old behavior).

@rljacob, what do you think?

zhangshixuan1987:
> The old warning message has me curious: "analyses file NOT FOUND. Switching nudging OFF to coast thru the gap." What was the use case where they wanted this behavior? Was it just a run-length that was longer than the nudging period?

@rljacob:  I do not know the use case for the developer who added this warning information. But one scenario in my mind is that nudging with incomplete reanalysis data, say, if your nudging window is 6-h, but we only have analysis at 00z, 06z, and 18Z, while the analysis data at 12z is missing. If you still want to run the model with nudging using only the available analysis, then the old way is one option. However, I do not think this is a clean way for nudging simulation.

zhangshixuan1987:
> Since the nudging files are normally created from analysis/reanalysis files, it's out of our control if there are missing dates. Thinking this a bit more, we may want to add a name list variable for this. For example, nudging_file_all_exist is 1) true, then end run if some nudging files are not found within the nudging period; 2) false, then continue with a warning message (old behavior).

zhangshixuan1987:
> @rljacob , there are user name list variables to control the start and end dates of nudging, so the code can handle the simulation length longer than the nudging period (a typical application of nudging runs). I guess it is more likely for the cases when the nudging files have missing dates.
> 
> Since the nudging files are normally created from analysis/reanalysis files, it's out of our control if there are missing dates. Thinking this a bit more, we may want to add a name list variable for this. For example, nudging_file_all_exist is 1) true, then end run if some nudging files are not found within the nudging period; 2) false, then continue with a warning message (old behavior).
> 
> @rljacob, what do you think?

Hi @tangq: I think that @tangq's idea is one option, but I think that this may also

rljacob:
Take a look at this statement around line 1691:
 if(.not.Nudge_File_Present) return

Maybe that is the place for a fatal error if the nudging file isn't found.

tangq:
> > @rljacob , there are user name list variables to control the start and end dates of nudging, so the code can handle the simulation length longer than the nudging period (a typical application of nudging runs). I guess it is more likely for the cases when the nudging files have missing dates.
> > Since the nudging files are normally created from analysis/reanalysis files, it's out of our control if there are missing dates. Thinking this a bit more, we may want to add a name list variable for this. For example, nudging_file_all_exist is 1) true, then end run if some nudging files are not found within the nudging period; 2) false, then continue with a warning message (old behavior).
> > @rljacob, what do you think?
> 
> Hi @tangq: I think that your idea is one option, but would this also add a potential danger that: if the users set nudging_file_all_exist = false., and turn on nudging, the model is still allowed to continue to run even if the nudging data file is not found.

@zhangshixuan1987 , the difference is that if the user sets the additional flag nudging_file_all_exist = false (default value is true), the user is expected to know that turning nudging on alone doesn't guarantee nudging is effective and the user has checked that not all the nudging files exist and decide to set nudging_file_all_exist = false.

I agree that there is no perfect way to handle all the nudging scenarios, but some flags like this one can avoid misuse of nudging to an extent.

zhangshixuan1987:
> nudging_file_all_exist

Hi @tangq : I agree with you. I think that the users at least know what will happen by setting  "nudging_file_all_exist = .false." (i.e. allow nudging to continue even if the nudging file does not exist). I have a question regarding the name of this flag, would it be better to name it "nudging_allow_mising_file" ?

rljacob:
Is this supposed to be merged after #4098?   Why not just add commits to #4098?

rljacob:
Oh I see, #4098 is for maint-1.0

tangq:
@zhangshixuan1987 , I like your suggestion about the flag name.

golaz:
What's the status of this?

zhangshixuan1987:
@golaz: Hi Chris, I think the PR is ready to merge.

singhbalwinder (approved):

singhbalwinder:
I will start merging this once the repo opens.

singhbalwinder:
Merged to next

PR #4102, opened 2021-02-17
Shorten BGC coupled tests

rljacob:
Shorten the BGC coupled tests in e3sm_integration to 2 days.
These were taking the most time.

[BFB] but test name will change.


rljacob:
These are easily some of the most expensive tests in integration.

jonbob:
@rljacob - I have no problem with this, but the results won't be BFB

rljacob:
But the 2 days will be identical in both runs right?  There's no way for our test system to recognize that because the test name changes with the addition of "Ld2".   Did we previously use "non-BFB" for a new test or test name change?

jonbob:
You're right -- it will be a new test name and have new results. I'm not sure how we have documented those kinds of changes before

rljacob:
Lets leave this case as BFB.  The non-BFB tag should be added if any existing test will change answers.

jonbob:
that makes sense -- sorry to make a mess of it

PR #4103, opened 2021-02-18
Update EAM metadata for CF and CMIP

rljacob:
Update EAM output metadata for CF and CMIP

Add CF standard_name to variables that were part of CMIP6 and require no modifications.
Fix errors in units flagged by the CF checker.
Add a few CMIP global attributes
Set FIllvalue to CMIP standard.

[BFB]


chengzhuzhang:
for eam h0 units fix based on CF checker results: https://gist.github.com/rljacob/c792a6bce67207258545b3c593f2594e
Errors haven't fixed: lev ilev CONCLD SO2_CLXF SOAG_CLXF bc_a4_CLXF, hstobie_linoz, mlip, mpoly mprot num_a1_CLXF num_a2_CLXF num_a4_CLXF pom_a4_CLXF so4_a1_CLXF so4_a2_CLXF
Also noticed two variables that have potential bad "long name"
```
RAM1
        float RAM1(time, ncol) ;
                RAM1:units = "frac" ;
                RAM1:long_name = "RAM1" ;
                RAM1:cell_methods = "time: mean" ;
airFV
        float airFV(time, ncol) ;
                airFV:units = "frac" ;
                airFV:long_name = "FV" ;
                airFV:cell_methods = "time: mean" ;
```

rljacob:
Progress on standard_name:
```
        float TS(time, ncol) ;
                TS:units = "K" ;
                TS:long_name = "Surface temperature (radiative)" ;
                TS:standard_name = "surface_temperature" ;
                TS:cell_methods = "time: mean" ;
```

rljacob:
This is ready.  I'll merge to next today if there's no objections.

components/eam/src/control/cam_history.F90
```diff
@@ -2093,7 +2093,7 @@ subroutine AvgflagToString(avgflag, time_op)
     case ('B')
       time_op(:) = 'mean00z'
     case ('I')
-      time_op(:) = ' '
+      time_op(:) = 'point'
```
wlin7, line 2096: More commonly such fields would be known as "instantaneous". I am not aware if there is a discussion that we should use "point" instead.

rljacob: For the cell_method:time, CF convention says its "point".

klindsay28: @rljacob's answer is supported by the list of `cell_methods` values in CF's [Appendix E: Cell Methods](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#appendix-cell-methods).

components/eam/src/physics/cam/cam_diagnostics.F90
```diff
@@ -602,27 +606,34 @@ subroutine diag_init()
...
+   call addfld ('LHFLX',horiz_only,    'A','W/m2','Surface latent heat flux', &
+   standard_name = 'surface_upward_latent_heat_flux')
+   call addfld ('QFLX',horiz_only,    'A','kg/m2/s','Surface water flux', &
+   standard_name = 'water_evapotranspiration_flux')
 
    call addfld ('TAUX',horiz_only,    'A','N/m2','Zonal surface stress')
    call addfld ('TAUY',horiz_only,    'A','N/m2','Meridional surface stress')
-   call addfld ('TREFHT',horiz_only,    'A','K','Reference height temperature')
+   call addfld ('TREFHT',horiz_only,    'A','K','Reference height temperature', &
+      standard_name='air_temperature')
```
wlin7, line 619: Surprised the standard name is not "surface_air_temperature". Probably we just have to take "air_temperature" because that is the only one suitable.

components/eam/src/physics/cam/cam_diagnostics.F90
```diff
@@ -602,27 +606,34 @@ subroutine diag_init()
...
    call addfld ('TREFHTMN',horiz_only,    'M','K','Minimum reference height temperature over output period')
    call addfld ('TREFHTMX',horiz_only,    'X','K','Maximum reference height temperature over output period')
-   call addfld ('QREFHT',horiz_only,    'A','kg/kg','Reference height humidity')
-   call addfld ('U10',horiz_only,    'A','m/s','10m wind speed')
-   call addfld ('RHREFHT',horiz_only,    'A','fraction','Reference height relative humidity')
+   call addfld ('QREFHT',horiz_only,    'A','kg/kg','Reference height humidity', &
+    standard_name = 'specific_humidity')
```
wlin7, line 623: This one does have a more suitable standard name, "surface_specific_humidity".
See also https://cfconventions.org/Data/cf-standard-names/28/build/cf-standard-name-table.html

components/eam/src/physics/cam/cam_diagnostics.F90
```diff
@@ -602,27 +606,34 @@ subroutine diag_init()
...
+   call addfld ('U10',horiz_only,    'A','m/s','10m wind speed', &
+     standard_name='wind_speed')
```
wlin7, line 625: Similarly "surfce_wind_speed" would suit better, but looks like we have such no choice.

components/eam/src/physics/cam/cam_diagnostics.F90
rljacob: I used what is in the published CMIP6 data.  I think reference height is not the surface  so we can't call this surface_specific_humidity.

components/eam/src/physics/cam/cloud_diagnostics.F90
```diff
@@ -155,9 +155,9 @@ subroutine cloud_diagnostics_init()
     call addfld ('TGCLDCWP',horiz_only,    'A',wpunits,'Total grid-box cloud water path (liquid and ice)', &
          sampling_seq=sampling_seq)
     call addfld ('TGCLDLWP',horiz_only,    'A',wpunits,'Total grid-box cloud liquid water path', &
-         sampling_seq=sampling_seq)
+         sampling_seq=sampling_seq, standard_name='atmosphere_mass_content_of_cloud_condensed_water')
```
wlin7, line 158: This standard name should be "atmosphere_mass_content_of_cloud_liquid_water", in contrast with cloud_ice  below.

components/eam/src/physics/cam/cloud_diagnostics.F90
```diff
@@ -155,9 +155,9 @@ subroutine cloud_diagnostics_init()
     call addfld ('TGCLDCWP',horiz_only,    'A',wpunits,'Total grid-box cloud water path (liquid and ice)', &
```
wlin7, line 155: This one can use the name 'atmosphere_mass_content_of_cloud_condensed_water', which includes liquid and ice.

components/eam/src/physics/rrtmg/radiation.F90
```diff
@@ -708,7 +714,8 @@ subroutine radiation_init(phys_state)
           call addfld('FLUT'//diag(icall), horiz_only,    'A',    'W/m2', 'Upwelling longwave flux at top of model', &
```
wlin7, line 714: FLUT is more commonly used than FLUTC, so it is useful to include standard name, 'toa_outgoing_longwave_flux"

Similarly quite a lot attention are paid to SWCF and LWCF for assess model performance involving cloud processes.  Including their standard_names should also be useful
SWCF: toa_shortwave_cloud_radiative_effect
LWCF: toa_longwave_cloud_radiative_effect

components/eam/src/control/cam_history.F90
wlin7: Thanks @rljacob , @klindsay28. I was wondering there might be a reference for that use, but I  didn't find it.

components/eam/src/physics/rrtmg/radiation.F90
rljacob: As I said in the description, this only adds standard names for variables we published already AND require no modifications.  Adding more standard_names can be part of a future PR or you can add them to this one today if you want.

components/eam/src/physics/cam/cam_diagnostics.F90
wlin7: Ok, then, Rob. This field is the closest we can have for 'surface' specific humidity, though.

rljacob:
To check the standard_names, for match the EAM variable name with CMIP6 name in this table (columns 1 and 3)
https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/925304036/Amon+variable+conversion+table
Then look up the standard_name using the CMIP6 name in this table
https://pcmdi.llnl.gov/mips/cmip3/variableList.html#Table_A1a

components/eam/src/physics/cam/cloud_diagnostics.F90
rljacob: Yep. I meant to put that on TGCLDCWP.

components/eam/src/physics/rrtmg/radiation.F90
wlin7: Sounds good, Rob. I was about to ask if we want to add for more variables in this PR. There will be more than just the few I mentioned, I think a future PR is better, Thanks for the explanation.

golaz: @rljacob : there probably won't be time for another PR.

rljacob: Right.  I meant another PR sometime in the future.   It will take some time to figure out other standard_names.

wlin7: @golaz , @rljacob : I am updating this PR with additional standard names.

rljacob: Only do the ones that are really obvious and it has to be done before next testing starts tonight.

components/eam/src/physics/cam/cam_diagnostics.F90
czender: I looked into how to better handle the issue of reference height and related variables that are defined, e.g., at a single height (or time, or location). The `standard_name` should be, as it is in this PR, the same as the full 3D field. However, there is a CF-ism called a _scalar coordinate variable_ that can be used to define the (in this case) reference height. That scalar coordinate would then be added to the coordinates attribute of the data variable (`RHREFHT` or `TREFHT` etc). It seems too intrusive to add new scalar coordinates this late in the game for v2, though maybe this could be done for v3. See examples here: http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#scalar-coordinate-variables

wlin7 (approved):
Approved all the changed except for commit  833bf54. Please take a look of that one, Rob.

wlin7:
> does this include snow and rain?

Yes, both PRECC and PRECT include snow and rain. I don't see a standard_name for so-called large-scale precipitation flux. Don't know it is it appropriate to also use 'precipitation_flux'; so just leave it alone.

wlin7:
> We can't call this a "flux" because the units are wrong.
You are right. A conversion would be needed to have the unit of kg/m2/s  as in Table_v1a. This can be done by multiplying 1000 kg/m3. If following the other [CF standard_name table](https://cfconventions.org/Data/cf-standard-names/28/build/cf-standard-name-table.html), "precipitation rate" would fit the output variable (name and unit) better. There are names of "lwe_stratiform_precipitation_rate", "lwe_convective_precipitation_rate", and "lwe_precipitationn_rate", for PRECL, PRECC, PRECT, respectively. Do we want that? (lwe for liquid water equivalent).

wlin7:
> How is FLUT calculated? For CMIP, @golaz said toa_outgoing_longwave_flux is FSNTOA - FSNT + FLNT

Good question. Energetically in equilibrium, yes. During calculations in RRTMG, FLUT is LW flux at rrtmg model top, purely based on LW radiative transfer. This model top differs from EAM model top, 0.0001 hPa vs. 0.1 hPa. @brhillman should be able to explain in more details.

rljacob:
@wlin7 If we have to convert or calculate anything, its to complicated for this PR.  Remove the PRECC, PRECT and FLUT standard_names.

wlin7:
Ok, Rob. BTW, I didn't mean to do the conversion for PRECC and PRECT in this PR. They should refer to the same quantity without conversion. The conversion factor is the liquid water density. That said, given the standard_name seems to go with the standard unit,  agreed not to attach a standard_name here.

Also, if we don't want the given standard_name for FLUT, should I also remove it for FLUTC?

rljacob:
No leave FLUTC because that is exactly what we used in CMIP6.

wlin7:
Done, Rob. We can have some discussion in tomorrow's meeting if we want to do this for more (and which) output fields.

PR #4106, opened 2021-02-18
Implement OpenMP CRM for MMF

xyuan:
This PR implements the OpenMP version of the CRM code in E3SM. It keeps the same E3SM/CRM interface, and uses a separate directory (samomp) to hold the OpenMP code. This PR also adds an additional compset (`F-MMFOMP`) and a test to `e3sm_integration` to test the new code.

[BFB]


components/eam/bld/namelist_files/namelist_defaults_eam.xml
```diff
@@ -1555,7 +1555,7 @@
 <rsplit > -1 </rsplit>
 <se_nsplit > -1 </se_nsplit>
 
-<se_tstep >                  600 </se_tstep>
+<se_tstep >                  300 </se_tstep>
```
whannah1, line 1558: This change shouldn't be here.

components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -75,6 +75,8 @@ subroutine crm_physics_register()
    use gator_mod, only: gator_init
 #elif defined(MMF_SAM)
    use setparm_mod      ,   only: setparm
+#elif defined(MMF_SAMOMP)
+   use setparm_mod     ,    only: setparm
```
whannah1, line 79: Would it be better to just use:
`#elif defined(MMF_SAM) || defined(MMF_SAMOMP)`
to avoid the duplicate lines?

components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -298,6 +300,8 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
    use cpp_interface_mod, only: crm
 #elif defined(MMF_SAM)
    use crm_module       , only: crm
+#elif defined(MMF_SAMOMP)
```
whannah1, line 303: same comment as above.

whannah1 (approved):
I didn't look at all of the CRM code, but I skimmed over a lot of it and it looks good.

whannah1:
For anybody in the future wondering why we have two different fortran CRM code bases, I figured I'd leave an explanatory comment. For various infrastructure related reasons we knew that we wanted to have an OpenMP port of the fortran code. The OpenACC code does not perform well on the GPU, so we are not planning to use it much, but we also don't want to delete it. The first attempt to do the OpenMP port in a single code base made us realize that there's no good way to have openACC and OpenMP directives live together, because it requires a mountain of #ifdefs that makes the code very difficult to read. Having two separate code bases is not a great alternative, but we decided this was the lesser of the evils. Of course this means that any bug fix or new feature needs to be added to 3, and eventually 4, different CRM codes, but that's just the price of needing to accommodate many different architectures and their associated quirks.

whannah1:
Something I forgot in my review is that this PR doesn't add a compset or tests for the openMP CRM. We should probably add them. @brhillman do you agree?

components/eam/bld/namelist_files/namelist_defaults_eam.xml
xyuan: Yes, change it back to 600

components/eam/src/physics/crm/crm_physics.F90
xyuan: that is great, thanks.

components/eam/src/physics/crm/crm_physics.F90
xyuan: yes, thanks

xyuan:
> Something I forgot in my review is that this PR doesn't add a compset or tests for the openMP CRM. We should probably add them. @brhillman do you agree?

We will add the compset or tests in new PR @whannah1 @brhillman @mrnorman

components/eam/src/physics/crm/samomp/SGS_TKE/diffuse_scalar.F90
```diff
@@ -0,0 +1,74 @@
...
+      call diffuse_scalar2D (ncrms,dimx1_d,dimx2_d,dimy1_d,dimy2_d,grdf_x,       grdf_z,f,fluxb,fluxt,tkh,rho,rhow,flux)
+    endif
+
+    !$omp target teams distribute parallel do collapse(2)
+    do k=1,nzm
+      do icrm = 1 , ncrms
+        fdiff(icrm,k)=0.
+      enddo
+    enddo
+    !$omp target teams distribute parallel do collapse(2)
```
mrnorman, line 56: Should this be `collapse(4)`?

components/eam/src/physics/crm/samomp/accelerate_crm.F90
```diff
@@ -0,0 +1,341 @@
...
+        do j = 1, ny
+          do i = 1, nx
+            do icrm = 1, ncrms
+              ! don't let T go negative!
+              t(icrm,i,j,k) = max(tmin, t(icrm,i,j,k) + crm_accel_factor * ttend_acc(icrm,k))
+              if (crm_accel_uv) then
+                u(icrm,i,j,k) = u(icrm,i,j,k) + crm_accel_factor * utend_acc(icrm,k) 
+                v(icrm,i,j,k) = v(icrm,i,j,k) + crm_accel_factor * vtend_acc(icrm,k) 
+              endif
+              !$omp atomic update
```
mrnorman, line 242: Why does this need to be atomic?

components/eam/src/physics/crm/samomp/pressure.F90
```diff
@@ -0,0 +1,337 @@
...
+      enddo
+    enddo
+
+    !-------------------------------------------------
+    !   Perform inverse Fourier transformation:
+    if(RUN3D) then
+      !$omp target teams distribute collapse(2)
+      do k=1,nzslab
+        do i = 1 , nx_gl+1
+          !$omp parallel do private(ftmp_y,work)
```
mrnorman, line 258: I'm curious why this is being split. If this is to mirror the OpenACC, I may have been working around an OpenACC bug here, where the compiler handled the private variables wrongly. It's maybe worth seeing is `!$omp target teams distribute parallel do private(...)` is worthy trying at line 255.

components/eam/src/physics/crm/samomp/pressure.F90
```diff
@@ -0,0 +1,337 @@
...
+            call fft991_crm(ftmp_y,work,trigxj,ifaxj,1,nx2,ny_gl,1,+1)
+            f(icrm,i,:,k) = ftmp_y
+          enddo
+        enddo
+      enddo
+    endif
+    !$omp target teams distribute collapse(2)
+    do k=1,nzslab
+      do j = 1 , ny_gl
+        !$omp parallel do private(ftmp_x, work)
```
mrnorman, line 271: same as above

components/eam/src/physics/crm/samomp/pressure.F90
```diff
@@ -0,0 +1,337 @@
...
+        do icrm = 1 , ncrms
+          !$omp flush(ftmp_x,work)
```
mrnorman, line 273: What does the flush directive do?

mrnorman (commented):
This looks good; only a few comments inline with code. I just want to confirm. Does this run with correct answers in standalone mode?

mrnorman (approved):

components/eam/src/physics/crm/samomp/pressure.F90
xyuan: yes, this is pretty annoying part of the implementation of OpenMP,  no-spliting of OpenMP clause doesn't work in OpenMP only if we sequence execute the FFT, otherwise it gives unmapped memory error. Separating of the clause gives the best performance without not too much modifications of the code, however, it is still a bottleneck in performance. That is also the reason why I work on the alternative algorithm which is not included in this PR.

components/eam/src/physics/crm/samomp/pressure.F90
xyuan: It flushes the cache, and makes a thread’s temporary view of memory consistent with memory and enforces an order on the memory operations of the variables explicitly specified or implied

components/eam/src/physics/crm/samomp/SGS_TKE/diffuse_scalar.F90
xyuan: yes, I fixed it.

components/eam/src/physics/crm/samomp/accelerate_crm.F90
xyuan: I was a bit of confusing the atomic update in this one, it is used for identifying bugs. I will test to check whether we can remove it or not.

components/eam/src/physics/crm/samomp/pressure.F90
mrnorman: Must be a common bug then...

components/eam/src/physics/crm/samomp/pressure.F90
mrnorman: I'd be curious to know what this maps to in CUDA. Any ideas?

components/eam/src/physics/crm/samomp/pressure.F90
xyuan: I worked with Eric on this kernel, and he analysis the ptx code. He suggests to use large stack memory size, it indeed works for the small grid size cases, but fails when we run large grid size cases. I would also think this is memory problem that handled in the OpenMP.

components/eam/src/physics/crm/samomp/pressure.F90
xyuan: We may get some ideas if we study the ptx code, actually Eric did this study.

rljacob:

whannah1:
@rljacob Waiting for MMF refactor PR

rljacob:
New plan is to put this PR in now and not wait for the refactor.  Other work is waiting on on this PR.

brhillman:
I'm getting a build failure on my local merge to next for the new F-MMFOMP test:
```
/home/bhillma/codes/e3sm/branches/next/components/eam/src/physics/crm/crm_physics.F90:936:69:

                use_crm_accel_tmp, crm_accel_factor, crm_accel_uv_tmp)
                                                                     1
Error: Type mismatch in argument ‘crm_accel_factor_in’ at (1); passed INTEGER(4) to REAL(8)
```

components/eam/src/physics/crm/samomp/crm_module.F90
```diff
@@ -0,0 +1,2499 @@
...
+!---------------------------------------------------------------
+use setparm_mod, only : setparm
+
+contains
+
+subroutine crm(lchnk, ncrms, dt_gl, plev,       &
+                crm_input, crm_state, crm_rad,  &
+                crm_ecpp_output, crm_output, crm_clear_rh, &
+                latitude0, longitude0, gcolp, igstep, &
+                use_crm_accel_in, crm_accel_factor_in, crm_accel_uv_in)
```
brhillman, line 48: Arguments for variance transport have since been added. In order to support the same function call, this interface needs to include these arguments, and then should assert that variance transport is not requested since it is not implemented in this version of the code.

brhillman (changes requested):

mt5555:
this PR changes 589 files?   is github confused? or something might be wrong with the branch.

xyuan:
> this PR changes 589 files? is github confused? or something might be wrong with the branch.

yes, accidently merger master to this branch. Ben fixed it.

brhillman (approved):

components/eam/cime_config/config_compsets.xml
```diff
@@ -291,64 +295,64 @@
...
-	<value  compset="GEOS_EAM\d+%WTSM">1990-01-01</value>
-	<value  compset="GEOS_EAM\d+%TMOZ">2008-01-01</value>
-	<value  compset="GEOS_EAM\d+%TBAM">2008-01-01</value>
-	<value  compset="GEOS_EAM\d+%S">2005-01-01</value>
-	<value  compset="RCP[2468]_EAM\d+">2005-01-01</value>
-	<value  compset="C2R[68]_EAM">1950-01-01</value>
-	<value  compset="C2R4_EAM">2004-01-01</value>
-	<value  compset="AR95_EAM">1995-07-18</value>
-	<value  compset="AR97_EAM">1997-06-19</value>
+  <value  compset="20TR_EAM">1850-01-01</value>
```
rljacob, line 298: Looks like a bunch of indentation was removed that should not have been.

brhillman: Ah strange, I'll fix before merging.

brhillman: Oh, these are hard tabs in master @rljacob  it looks like someone's editor maybe removed them.

rljacob: As far as I can tell, tabs are ok in XML.  xmllint doesn't complain.  So leave them as tabs.

whannah1: Looks like my editor is the one that did this. No idea how though... I never touched this block

whannah1: Fixed.

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #4110, opened 2021-02-21
better control of vertical remap options

mt5555:
reorg vertical remap code and add new options

vert_remap_q_alg=0          unchanged
vert_remap_q_alg=1           unchanged
vert_remap_q_alg=2          change default region of 1st order to 1 layer from 2 layers
vert_remap_q_alg=3          removed (it was identical to 1)
vert_remap_q_alg=10        unchanged
vert_remap_q_alg=11         new: q_alg=10 with out limiters, for research

[BFB]

### DIFF
--- a/components/homme/src/share/vertremap_base.F90
+++ b/components/homme/src/share/vertremap_base.F90
@@ -33,7 +33,6 @@ module vertremap_base
   use hybvcoord_mod, only          : hvcoord_t
   use perf_mod, only               : t_startf, t_stopf  ! _EXTERNAL
   use parallel_mod, only           : abortmp, parallel_t
-  use control_mod, only : vert_remap_q_alg
 
 ! options:
 ! remap1                  ! remap any field, splines, monotone
@@ -75,7 +74,7 @@ end subroutine remap_calc_grids
 
 
 
-subroutine remap1(Qdp,nx,qsize,dp1,dp2)
+subroutine remap1(Qdp,nx,qsize,dp1,dp2,remap_alg)
   ! remap 1 field
   ! input:  Qdp   field to be remapped (NOTE: MASS, not MIXING RATIO)
   !         dp1   layer thickness (source)
@@ -84,7 +83,7 @@ subroutine remap1(Qdp,nx,qsize,dp1,dp2)
   ! output: remaped Qdp, conserving mass, monotone on Q=Qdp/dp
   !
   implicit none
-  integer, intent(in) :: nx,qsize
+  integer, intent(in) :: nx,qsize,remap_alg
   real (kind=real_kind), intent(inout) :: Qdp(nx,nx,nlev,qsize)
   real (kind=real_kind), intent(in) :: dp1(nx,nx,nlev),dp2(nx,nx,nlev)
   ! ========================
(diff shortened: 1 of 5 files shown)


components/homme/src/share/vertremap_base.F90
```diff
@@ -610,26 +611,25 @@ subroutine remap_Q_ppm(Qdp,nx,qsize,dp1,dp2)
...
+              ao(nlev+k) = ao(nlev)
+           enddo
+        else if (remap_alg==10) then
+           ext(1) = minval(ao(1:nlev))
+           ext(2) = maxval(ao(1:nlev))
+           call linextrap(dpo(2), dpo(1), dpo(0), dpo(-1), ao(2), ao(1), ao(0), ao(-1), ext(1), ext(2))
+           call linextrap(dpo(nlev-1), dpo(nlev), dpo(nlev+1), dpo(nlev+2),&
+                ao(nlev-1), ao(nlev), ao(nlev+1), ao(nlev+2), ext(1), ext(2))
+        else if (remap_alg==11) then
+           call linextrap(dpo(2), dpo(1), dpo(0), dpo(-1), ao(2), ao(1), ao(0), ao(-1), ext(1), ext(2))
```
ambrad, line 627: I think this is calling linextrap with uninit'ed ext(1:2) values. I think you should add a boolean argument `limit` to linextrap saying whether to limit using `ext`.

mt5555: fixed.

components/homme/src/share/vertremap_base.F90
```diff
@@ -806,8 +811,10 @@ subroutine linextrap(dx1,dx2,dx3,dx4,y1,y2,y3,y4,lo,hi)
     a  = num/den
     y4 = (1-a)*y1 + a*y2
 
-    y3 = max(lo, min(hi, y3))
-    y4 = max(lo, min(hi, y4))
+    if (limit==0) then
```
ambrad, line 814: I think `limit==1` is meant here.

mt5555: attempting to fix a 2nd time...

ambrad (approved):

mt5555:
initial results with extended 1st order reconstruction were misleading. force pushed a much simpler / smaller version of this branch.

oksanaguba:
so, option 2 won't cause nonbfb? i am not sure it is in the tests. 
should there be a test for the new option 11? i can change one of namelists.

oksanaguba:
Tests use only 1 and 10, should i add 0, 2 and 11 then?

mt5555:
EAM has migrated to 10.  some of the old tests still use 1.  so I say no reason to test the others.

oksanaguba:
closed?

mt5555:
E3SM uses qalg=10.  New qalg=11 added, to provide option for a not-quite-montone truly 2nd order algorithm.
Older algorithms (qalg=0,1,2) should be abandoned and do not need to be tested.

For testing, will update the following tests:

thetanh-test22.nl  use  qalg=11
thetah-test22.nl   use  qalg=10

PR #4112, opened 2021-02-22
fix ESMT bugs in MMF 

xyuan:
Fix some bugs in scalar_momentum fortran code.
1): In fortran code, the multidimensional array argument must be in the same shape.
2): C++ array need to be different indexed as the Fortran array.
3): Testing of the code requires non-zero vertical velocity (w /= 0), simply check the in/out ul_esmt, vl_esmt differences are not enough because the vertical velocity w=0, it leads to very tiny output differences.

[BFB]


whannah1 (approved):
Thanks for finding these bugs!

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #4113, opened 2021-02-23
Update ELM and MOSART variables and global metadata

rljacob:
Update ELM and MOSART history files for CF and CMIP

Replace non-compliant units like "none" and "unitless" with "1".
Add ability to add standard_name to ELM history and define a couple
Add some CMIP global attributes.
In MOSART, change spval and fillvalue to CMIP standard 1.e+20

[BFB]

### DIFF
--- a/components/elm/src/biogeochem/CH4Mod.F90
+++ b/components/elm/src/biogeochem/CH4Mod.F90
@@ -353,12 +353,12 @@ contains
     end if
 
     this%finundated_col(begc:endc) = spval
-    call hist_addfld1d (fname='FINUNDATED', units='unitless', &
+    call hist_addfld1d (fname='FINUNDATED', units='1', &
          avgflag='A', long_name='fractional inundated area of vegetated columns', &
          ptr_col=this%finundated_col)
 
     this%finundated_lag_col(begc:endc) = spval
-    call hist_addfld1d (fname='FINUNDATED_LAG', units='unitless',  &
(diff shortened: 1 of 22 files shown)


rljacob:
After testing, cprnc noticed some fill value differences from baselines from this test:
SMS_Ln9.ne4_ne4.FC5AV1C-L.chrysalis_intel.eam-outfrq9s
In the following variables
LAKEICEFRAC, TG_R, TG_U, TKE1, TLAKE, TSA_R, TSA_U

In addition, the coupler reported diffs in l2x_Flrl_Tqsub which look like fillvalue diffs.

These diffs don't actually change answers (no other fields diffed).

But the finidat files have the old FIllvalue so there are implications to changing the value in ELM that require more thought.  Will back out the fillvalue change.

rljacob:
What is l2x_Flrl_Tqsub?
```
l2x(index_l2x_Flrl_Tqsub,i)  = lnd2atm_vars%Tqsub_grc(g)
real(r8), pointer :: Tqsub_grc(:) => null()             ! grc Temperature of subsurface runoff
lnd2atm_vars%Tqsub_grc(g) = avg_tsoil(lnd2atm_vars%zwt_grc(g),lnd2atm_vars%t_soisno_grc(g,-nlevsno+1:nlevgrnd))
avgtsoil() ! Function for estimating average soil temperature within the saturated soil zone (which produces subsurface runoff)
```

PR #3853, opened 2020-09-26
ELM-Erosion bug fixing

tanzeli1982:
Fix a units bug in ELM-Erosion

[B4B]

PR #4116, opened 2021-02-23
Update CIME submodule

jgfouca:
To 2b49cf54f1fb0a53832762dc29f19fc31c491d31

Changes:
1) Adds support of DATM_MODE=CLM1PT for ELM
2) Update driver-mct support for trigrid configuration where lnd is not on the atm grid

Fixes:
1) Fix compare test results undefined
2) SCORPIO needs _PNETCDF
3) scripts_regression_tests: Fixes cleaning up only on success
4) bless/compare_test_results should not process build-only tests
5) wait_for_tests: Make update success more robust

This PR is mostly BFB but will cause DIFFs to a significant number of tests due to:
```
Could not find match for file1 variable fracr_lfrin in file2
...
diff_test: the two files DIFFER only in their field lists
```
It will cause diffs in river model variables in tri-grid cases because of the use of the new fracr_lfrin

[non-BFB] in tri-grid cases with MOSART.


jgfouca:
It looks like the new field is from one of our CIME pushes (Merge pull request #3853 from ESMCI/jonbob/update-trigrid-support).

jonbob:
@jgfouca - yes, that makes sense. And sorry if it makes this PR more painful

jgfouca:
@jonbob it's no big deal, I'll just need to remember to ask for a bless when it's merged.

jgfouca:
Merged to next

rljacob:
@jonbob please confirm that this will be non-bfb for tri-grid cases with MOSART because of using lfrin.

jonbob:
This PR is non-BFB for anything using a trigrid configuration with active lnd and rof. It changes the fractions used in lnd2rof coupling but only when they are on the same grid.

PR #4122, opened 2021-02-25
update mpas-source: Removes shortwave energy conservation fix

mark-petersen:
See https://github.com/MPAS-Dev/MPAS-Model/pull/820

Coupled testing showed unexpected extremely strong sensitivity to the
previous shortwave radiation energy fixer. In that fix the excess
energy was used to heat the bottom layer. This yielded a strong
response in tropical precipitation, dramatically increasing the double
ITCZ bias. This removes that energy fixer and only tracks the solar
radiation that is discarded for extremely shallow cells

[NML]
[non-BFB]

### DIFF
--- a/components/mpas-ocean/bld/build-namelist
+++ b/components/mpas-ocean/bld/build-namelist
@@ -669,6 +669,7 @@ if (($OCN_ICEBERG eq 'true') && ($OCN_FORCING eq 'active_atm')) {
 add_default($nl, 'config_sw_absorption_type');
 add_default($nl, 'config_jerlov_water_type');
 add_default($nl, 'config_surface_buoyancy_depth');
+add_default($nl, 'config_enable_shortwave_energy_fixer');
 
 ###########################################
 # Namelist group: tidal_potential_forcing #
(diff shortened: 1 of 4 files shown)


mark-petersen:
@vanroekel you said in the PR:
> This removes that energy fixer and only tracks the solar radiation that is discarded for extremely shallow cells

It sounds like we should add the new variable `bottomLayerShortwaveTemperatureFlux` to the E3SM monthly averages? A sum in time makes more sense conceptually, but the monthly average times the duration is the same thing.

vanroekel:
@mark-petersen I wasn't sure what the best place for that variable was.  You are right that the most correct is to put in the sum in time, but I was indeed hoping the average times the length of the month would get us close enough to closing the budget.

vanroekel:
@jonbob I just pushed a commit to the MPAS-Dev corresponding PR that adds the option to add in the missing heat, so I wanted to let you know that there will also be a namelist change coming with this PR too, `config_enable_shortwave_energy_fixer` which should default to false.

mark-petersen:
This now contains the additional flag 'config_enable_shortwave_energy_fixer', default should be false.

jonbob:
merged to next, sanity testing launched

jonbob:
merged to master and expected DIFFs/NML DIFFs blessed -- except for compy_developer, which either ran before this PR was merged or has yet to report

PR #4124, opened 2021-02-25
Update CircleCI timeout

lukaszlacinski:
Update Circle CI timeout value to 60 minutes.
add "-v" to ./case.build

[BFB]

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,9 +3,6 @@ jobs:
   build:
     machine:
       image: ubuntu-2004:202010-01
-    branches:
-      only:
-        - master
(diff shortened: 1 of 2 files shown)


rljacob:
Will this do what the old Travis CI did?  When a PR was opened, it did a test merge to master and ran the build test on that.

rljacob:
I was able to do that in the settings on CircleCI.

lukaszlacinski:
> Since PR-only builds can be set in the settings, we may not want this. For example, we want to build on PRs to the maint-1.0 branch.

I changed it back to run on all branches.

PR #4126, opened 2021-02-26
update mpas-source: Fixes missing edgeSignOnCell in redi_term2

mark-petersen:
See https://github.com/MPAS-Dev/MPAS-Model/pull/819

Currently redi_term2 is missing a factor of edgeSignOnCell in its calculation. It is being correctly applied in the limiter, but not the original calculation.

[non-BFB]


mattdturner:
This passes:
```
PASS:
SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_pgi
PEM_Ln9.T62_oQU240.GMPAS-IAF.compy_pgi
PET_Ln9.T62_oQU240.GMPAS-IAF.compy_pgi
SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_pgi.C
  - PASS SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_pgi BASELINE /compyfs/e3sm_baselines/pgi
/master:
```

I thought that this PR was supposed to be non-BFB?  Does it make sense that it passes the `SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_pgi` test?

jonbob:
Passes:
PEM_Ln9.ne30_oECv3.A_WCYCL1850S.anvil_intel
ERS_P640_Ld5.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel
SMS_D_Ld1.ne30_oECv3.A_WCYCL1850S_CMIP6.chrysalis_intel.allactive-v1cmip6

jonbob:
@mattdturner - I am surprised it's BFB. I checked and redi is on for oQU120, so it's not that. @vanroekel is running a long test over the weekend, and between his test and the automated testing we should get a better sense of its impact.

jonbob (approved):
Looks good from a visual inspection and testing inside E3SM

maltrud:
i suspect that the clipping code doesn't get invoked very often in the QU120?  especially in a short run.

jonbob:
Thanks @maltrud -- that's a much better explanation. I didn't ever look at the details of this change...

jonbob:
merged to next. Thanks @mattdturner for help testing, and thanks @maltrud for the explanation of why it might not change answers in some short tests

vanroekel:
Hmm, I'm not sure that makes sense to me @maltrud.  The change affects redi_term2 independent of clipping.  let me take a look at the code again.  I would have expected non BFB here.

vanroekel:
okay I see why this passes BFB.  This change will only affect how often the limiter is applied.  If the limiter is not applied then the redi terms are applied correctly (independent of this bug fix).  This PR will only be nonBFB when the tracer tries to go out of bounds.  So I would expect as @maltrud suggests this will be BFB for short tests and will likely have a minor impact on long integrations too.

jonbob:

PR #4127, opened 2021-02-26
Updates WC14 and EC30to60 initial conditions

vanroekel:
This PR updates the WC14 and EC30to60 initial conditions so they are consistent for WCv2

[non-BFB] for tests with either of these grids

### DIFF
--- a/components/mpas-ocean/cime_config/buildnml
+++ b/components/mpas-ocean/cime_config/buildnml
@@ -272,7 +272,7 @@ def buildnml(case, caseroot, compname):
         ic_date = '210210'
         ic_prefix = 'ocean.EC30to60E2r2'
         if ocn_ic_mode == 'spunup':
-            ic_date = '210210'
+            ic_date = '201001'
             ic_prefix = 'mpaso.EC30to60E2r2.rstFromG-anvil'
 
     elif ocn_grid == 'WC14to60E2r3':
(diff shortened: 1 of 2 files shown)


vanroekel:
@jonbob when I was testing the new WC14 files were staged it doesn't seem there is a default WC14 layout for chrysalis (I found a graph file was missing for ice because it was a weird selection).  Should that be added here?  I could copy what @tangq is using for his run and put it here or make a separate PR.  Having that in place is not as high priority as getting the EC30to60 IC corrected

jonbob:
@vanroekel - let's just get this in as-is today and make a separate PR for chrysalis layouts

jonbob (approved):
I checked and it looks like the new WC14 restart files are available on the inputdata server. I'll double-check that they are correctly identified after a test merge.

jonbob:
Test merge passes:
SMS_PS.northamericax4v1pg2_WC14to60E2r3.A_WCYCL1850S_CMIP6.anvil_intel
ERS_P640_Ld5.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel
I checked and the tests are getting the new correct ICs.

jonbob:
merged to master but will need to bless expected DIFFs after tonight's testing

jonbob:
expected DIFFs blessed

PR #4132, opened 2021-03-01
Fix compy layouts for G-cases with oEC60to30v3wLI

jonbob:
Removes a redundant part of config_pesall.xml that was introduced mainly for the new EC30to60E2r2 grid and instead reuses older layouts for oEC60to30v3*. It removes an issue where an oEC60to30v3wLI case was failing with  an error for "More than one PE layout matches given PE specs".

[BFB]

### DIFF
--- a/cime_config/allactive/config_pesall.xml
+++ b/cime_config/allactive/config_pesall.xml
@@ -9538,7 +9538,7 @@
       </pes>
     </mach>
   </grid>
-  <grid name="a%T62.+_oi%oEC60to30.*">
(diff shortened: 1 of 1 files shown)


rljacob (approved):

jonbob:
Passed sanity testing on anvil and ERS_P480_Ld5.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.compy_intel was successful. Merged to next

jonbob:
merged to master

PR #4136, opened 2021-03-03
Remove old CAMRT package and support

brhillman:
Remove the unsupported, untested, and very likely broken old CAMRT radiative transfer package. This removes the package itself, as well as support for CAMRT-specific restart variables, namelist variables and support, and build options. This actually simplifies our radiative transfer interface a little bit by allowing removal of some additional variables that were needed for CAMRT backwards compatibility, but were unused for RRTMG and RRTMGP.

[BFB]


rljacob:
Do we need camrt to run simple configs?  I'm looking at this code in configure
```
# Set default
my $rad_pkg = 'rrtmg';
if ($phys_pkg =~ m/^ideal$|^adiabatic$|^cam[34]$/) {
    $rad_pkg = 'camrt';
}
```

whannah1:
@rljacob Just speculating here, but I do not think CAMRT is critical for the idealized physics. CAMRT is much simpler when you're using single moment microphysics, but we were able to get around that with the MMF and I assume we can use a similar workaround when we revive the idealized physics.

components/eam/bld/config_files/definition.xml
```diff
@@ -63,9 +63,9 @@ Switch to turn on/off advecting CLUBB moments: 0 => no, 1 => yes
 PBL package: uw (University of Washington), hb (Holtslag and Boville), hbr
  (Holtslag, Boville, and Rasch), clubb_sgs.
 </entry>
-<entry id="rad" valid_values="rrtmg,rrtmgp,camrt" value="">
+<entry id="rad" valid_values="rrtmg,rrtmgp" value="">
 Radiative transfer calculation: 
-camrt (CAM3 and CAM4 RT package), rrtmg (RRTMG package from AER), rrtmgp (RRTMGP package from AER).
+rrtmg (RRTMG package from AER), rrtmgp (RRTMGP package from AER).
```
whannah1, line 68: Should we bother defining what "AER" stands for here?

components/eam/bld/configure
```diff
@@ -1062,25 +1062,14 @@ if ($print>=2) { print "Using $pbl_pkg PBL scheme.$eol"; }
...
-
-    # The camrt radiation doesn't work with the modal aerosols
-    if ($chem_pkg =~ /_mam/) {
-        die "configure ERROR: radiation package: $rad_pkg is not compatible\n".
-            "                 with aerosol package $chem_pkg\n";
-    }
-} elsif ($rad_pkg eq 'rrtmg' or $rad_pkg eq 'rrtmgp') {
+if ($rad_pkg eq 'rrtmg' or $rad_pkg eq 'rrtmgp') {
 
     # The rrtmg package doesn't work with the CAM3 prescribed aerosols
```
whannah1, line 1074: You could probably delete this block if you want to up your "deleted lines" score

components/eam/bld/namelist_files/namelist_defaults_eam.xml
```diff
@@ -145,42 +145,6 @@
 
 <!-- Bulk aerosol physical properties (includes optics) -->
 
-<!-- CAM3 Bulk aerosols.  Optics for CAM-RT -->
```
whannah1, line 148: Do you know if we would need special optics files if we were to attempt to revive the BAM option and run it with RRTMG?

whannah1 (approved):
Looks good to me!

components/eam/bld/config_files/definition.xml
brhillman: I can link to the source websites if you want. AER is Atmospheric and Environmental Research, the group behind both projects. I guess this isn't relevant now that we don't have CAMRT to distinguish between different sources for these codes. I could also refer to the relevant papers here, but this is far more detail than anything else in this file.

components/eam/bld/configure
brhillman: I'm not sure whether or not to remove this here or in a separate PR that removes other mentions of cam3 and cam4 values for `-phys`. Looking a little deeper, the only values of `-phys` that are used by EAM compsets in `components/eam/cime_config/config_component` are `default`, `ideal`, and `adiabatic`. I.e., these appear to be the only ways that EAM is used. So it appears as though ALL of this logic surrounding cam3/cam4/cam5 can go away, and greatly simplify the configure script. Thoughts? Pinging @rljacob for opinions?

components/eam/bld/namelist_files/namelist_defaults_eam.xml
brhillman: Most likely, but they may exist already. Using CAM-RT optics files would not be an option I don't think, I believe the bands are different.

components/eam/bld/configure
rljacob: You should ask @wlin7 who has been looking at simplifying the configure script.

components/eam/bld/config_files/definition.xml
whannah1: I just meant to define the acronym to make it easier for someone to look it up themselves.

components/eam/bld/configure
```diff
@@ -1062,25 +1062,14 @@ if ($print>=2) { print "Using $pbl_pkg PBL scheme.$eol"; }
 
 # Set default
 my $rad_pkg = 'rrtmg';
-if ($phys_pkg =~ m/^ideal$|^adiabatic$|^cam[34]$/) {
-    $rad_pkg = 'camrt';
-}
```
whannah1, line 1067: Can you put this line back and change it to RRTMG? 
I'll need this when I revive the idealized physics compsets.

whannah1 (approved):
new changes look good

wlin7:
Did a final review. A very thorough clean up of the codes that are related to CAMRT and cam3/cam4. Great work, @brhillman .
Also tested BFB with master (using a test merge of this PR) for F2010. Restart test also good.
Merged to next.

wlin7:
Merged to master.

PR #4142, opened 2021-03-04
dycore topography treatment improvements

mt5555:
Three new options (disabled by default) which improve ability of the dycore to work with rougher topography.

1) introduce balanced geopotential state to pressure gradient term to reduce discretization error
2) turn on HV reference profile on dp3d (dont dissipate background state)
3) for HV on potential temperature, use an approximation to Laplacian on pressure surfaces 

Testing here:
https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/2003339427/Hyperviscosity+Reference+States

[BFB]


mt5555:
one more change coming before this sholdl be merged:  need to make the potential temperature threshold a namelist variable.

components/eam/bld/namelist_files/namelist_definition.xml
```diff
@@ -5598,12 +5598,42 @@ Indicates form of eqn. for vtheta in theta model.
...
+<entry id="hv_theta_thresh" type="real" category="se"
+       group="ctl_nl" valid_values="" >
+Max threshold for Laplacian on pressure surface d(theta)/dp correction term. 
+Default (set by dycore)
+</entry>
+
 <entry id="cubed_sphere_map" type="integer" category="se"
        group="ctl_nl" valid_values="" >
-Defines which map is used for referene element in homme.
+Defines which map is used for referece element in HOMME.
```
ambrad, line 5634: "reference"

components/eam/bld/namelist_files/namelist_definition.xml
```diff
@@ -5766,7 +5796,7 @@ Default: Set by build-namelist.
 Tracer transport algorithm:
 0: Eulerian flux-form.
 12: COMPOSE semi-Lagrangian.
-Default: 0 (preqx), 12 (theta)
```
ambrad, line 5769: Why is this being changed? preqx still will get the value 0, right?

components/homme/src/theta-l/share/prim_advance_mod.F90
```diff
@@ -1444,7 +1445,14 @@ subroutine compute_andor_apply_rhs(np1,nm1,n0,qn0,dt2,elem,hvcoord,hybrid,&
         mgrad(:,:,1,k) = (dpnh_dp_i(:,:,k)*gradphinh_i(:,:,1,k)+ &
               dpnh_dp_i(:,:,k+1)*gradphinh_i(:,:,1,k+1))/2
         mgrad(:,:,2,k) = (dpnh_dp_i(:,:,k)*gradphinh_i(:,:,2,k)+ &
-              dpnh_dp_i(:,:,k+1)*gradphinh_i(:,:,2,k+1))/2
+             dpnh_dp_i(:,:,k+1)*gradphinh_i(:,:,2,k+1))/2
+
+        if (pgrad_correction==1) then
+           T0 = TREF-.0065*TREF*Cp/g     ! = 97
```
ambrad, line 1451: The constant `0.0065` should have a `_real_kind` annotation.

ambrad (changes requested):

components/homme/src/theta-l/share/prim_advance_mod.F90
ambrad: This constant occurs twice in the code. Perhaps it should be made a true constant rather than hardcoded.

mt5555:
splitting vtheta namelist change into it's on PR.

ambrad:
Since you're changing a pre-existing `0.0065` to `0.0065_real_kind`, I think to keep the standalone C++-F90 BFB tests passing you should remove the `f` in the literal `0.0065f` in this line:
https://github.com/E3SM-Project/E3SM/blob/dbdd4b32c4b77686cb816ab2fedca3011a46638c/components/homme/src/theta-l_kokkos/cxx/ElementOps.hpp#L86

components/homme/src/theta-l/share/model_init_mod.F90
```diff
@@ -87,21 +81,30 @@ subroutine model_init2(elem,hybrid,deriv,hvcoord,tl,nets,nete )
...
-      call phi_from_eos(hvcoord,elem(ie)%state%phis,&
-           temp,elem(ie)%derived%dp_ref,elem(ie)%derived%phi_ref)
-      elem(ie)%derived%theta_ref=0
-      elem(ie)%derived%dp_ref=0
-#endif
+      if (hv_ref_profiles==0) then
+         ! keep PHI profile, but dont use theta and dp:
+         elem(ie)%derived%theta_ref=0
+         elem(ie)%derived%dp_ref=0
+      endif
```
ambrad, line 88: You might change this to a `select case` block.

mt5555:
replaced two occurrences of 0.0065 with "tref_lapse_rate"

two definitions of tref_lapse_rate via #define HOMMEXX_BFB_TESTING.  old version is to preserve BFB because of V2 code freeze.  new version should be adopted after V2 branch.

mt5555:
@ambrad , can you finalize your review, at your leisure.

ambrad:
@mt5555, I think this will still lead to failures in the standalone C++-vs-F90 tests. Right now, as noted above, C++ hard-codes `0.0065f` to match the old F90. Now you've updated the F90 in the HOMMEXX BFB build to use `.0065_real_kind` but you haven't updated the C++ yet to match that. I linked the C++ line above that should be changed.

I don't know whether the C++-vs-F90 standalone tests have baselines, but if so, those will diff of course.

mt5555:
3rd attempt pushed.  lets just go with singe precision definition to preserve BFBness.

ambrad (approved):

oksanaguba:
homme tests won't build, error is
```
/gpfs/fs1/home/ac.onguba/acme-integrator/components/homme/src/theta-l/share/model_init_mod.F90(102): error #6460: This is not a field name that is defined in the encompassing structure.   [LAP_P_WK]
            elem(ie)%derived%lap_p_wk(:,:,k)=laplace_sphere_wk(temp(:,:,k),deriv,elem(ie),&
-----------------------------^
/gpfs/fs1/home/ac.onguba/acme-integrator/components/homme/src/theta-l/share/model_init_mod.F90(102): error #6366: The shapes of the array expressions do not conform.
            elem(ie)%derived%lap_p_wk(:,:,k)=laplace_sphere_wk(temp(:,:,k),deriv,elem(ie),&
------------^
```
The fix should go into homme/src/theta-l_kokkos/element_state.F90 -- should I fix it?

mt5555:
that's strange - suggests i messed something up at some point.  let me take a look.

oksanaguba:
This is for theta_kokkos exec, so, not surprising if you did not run it.

oksanaguba:
there is a fail in caar_ut, too, will address next.

oksanaguba:
Ran hommebfb suite, all was ok.
Ran e3sm_integration on anvil, all was ok, too
```
./cs.status.20210329_213332_cipfx3 | grep FAIL
```

components/homme/src/theta-l/share/viscosity_theta.F90
```diff
@@ -77,8 +79,7 @@ subroutine biharmonic_wk_theta(elem,stens,vtens,deriv,edgebuf,hybrid,nt,nets,net
 
    !if tensor hyperviscosity with tensor V is used, then biharmonic operator is (\grad\cdot V\grad) (\grad \cdot \grad) 
    !so tensor is only used on second call to laplace_sphere_wk
-   var_coef1 = .true.
-   if(hypervis_scaling > 0)    var_coef1 = .false.
+   var_coef1 = .false.
```
oksanaguba, line 82: Would this change, setting var_coef1 to false, cause issues with ne30 runs? This is not a compy issue.

oksanaguba:
I ran ERS_Ln5.ne4pg2_ne4pg2.FC5AV1C-L.compy_pgi.eam-thetahy_sl_pg2 with next+this PR against next minus this PR. There is a fail in baseline comp. I then replaced some of files in the failing clone with some files from the baseline clone. This caused the test to pass. I then started reverting some of affected files one-by-one to see which files cause nonbfb. It turns out that only prim_advance_mod is an issue. When it is replaced with its baseline version while other files are from the PR, the code passes. When it is replaced by the PR version, the code fails. I assume this can be resolved by us visually confirming that changes in prim_advance_mod are ok. However, there is a separate issue in HV mod. 

```
...
#	modified:   model_init_mod.F90
#	modified:   prim_advance_mod.F90
#	modified:   viscosity_theta.F90
identical

#	modified:   model_init_mod.F90
#	modified:   viscosity_theta.F90
diff

#	modified:   model_init_mod.F90
#	modified:   prim_advance_mod.F90
identical

#	modified:   prim_advance_mod.F90
identical
```

components/homme/src/theta-l/share/viscosity_theta.F90
mt5555: this is correct. for constant coefficient HV, var_coef is always .false.  for the tensor, it's false on the first call and true on the second cal.

oksanaguba:
Though it would not be necessary I wanted to run with const HV but there are tests with const HV, for example, SMS.ne4_ne4.FC5AV1C-L.eam-preqx_ftype4. Will merge now.

PR #4143, opened 2021-03-04
make vtheta_thresh a namelist variable

mt5555:
change default value of vtheta_thresh from 10K to 100K
Also add the ability to specify a different value in user_nl_eam

As this is very rarely triggered, labeling:
[BFB]

### DIFF
--- a/components/eam/bld/namelist_files/namelist_definition.xml
+++ b/components/eam/bld/namelist_files/namelist_definition.xml
@@ -5614,7 +5614,7 @@ Default: 1.
 
 <entry id="vtheta_thresh" type="real" category="se"
        group="ctl_nl" valid_values="" >
-Minim threshold for virtual potential temperature in the theta dycore.
+Minimum threshold for virtual potential temperature in the theta dycore.
 Values below this will trigger a warning and be truncated.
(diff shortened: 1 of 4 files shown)


mt5555:
on next

golaz:
@mt5555 : can this be merged to master?

rljacob:
I think it can be.  @mt5555 why wasn't there a namelist diff with this?

golaz:
@rljacob : do you want to merge it?

mt5555:
I can merge late this afternoon, if Rob doesn't do it first.

Took the morning off

rljacob:
merged.

mt5555:
FYI, there was no namelist diff because the option was only added to namelist_definition.xml (but not namelist_defaults and buildnml).    So a user can set it in user_nl_eam, but otherwise it wont appear in atm_in.   This is following existing practice for many of the esoteric atmosphere parameters that are only of interested to developers.

oksanaguba:
This PR changed vtheta_thresh value from 10 to 100. The same change should have been done in 

src/theta-l_kokkos/cxx/CaarFunctorImpl.hpp

PR #4147, opened 2021-03-08
Fix unit test

oksanaguba:
The cxx part of the test needed updated value for theta threshold.

[bfb] except for hommebfb suite.


oksanaguba:
On anvil this fix works.

oksanaguba:
in next

PR #4157, opened 2021-03-12
fates bgc bugfixes

rgknox:
These are bug fixes for bgc coupling of fates. FATES is now instantiated only on active columns,
which is consistent with the soil column filter upon which CNP biogeochemistry is active. 
Also, a litter flux was zero'd upon allocation which is critical for mass balance checks when FATES
is active.  Finally, the FATES submodule pointer is updated to its most recent tag: sci.1.43.3_api.14.2.0

Fixes: #4186 

[non-BFB] for only FATES

### DIFF
--- a/components/elm/src/data_types/ColumnDataType.F90
+++ b/components/elm/src/data_types/ColumnDataType.F90
@@ -7785,7 +7785,7 @@ contains
     allocate(this%actual_immob_nh4                (begc:endc))                    ; this%actual_immob_nh4              (:)   = nan
     allocate(this%smin_no3_to_plant               (begc:endc))                    ; this%smin_no3_to_plant             (:)   = nan
     allocate(this%smin_nh4_to_plant               (begc:endc))                    ; this%smin_nh4_to_plant             (:)   = nan 
-    allocate(this%plant_to_litter_nflux           (begc:endc))                    ; this%plant_to_litter_nflux         (:)   = nan
+    allocate(this%plant_to_litter_nflux           (begc:endc))                    ; this%plant_to_litter_nflux         (:)   = 0._r8
     allocate(this%plant_to_cwd_nflux              (begc:endc))                    ; this%plant_to_cwd_nflux            (:)   = nan
     ! C4MIP output variable
     allocate(this%plant_n_to_cwdn                  (begc:endc))                   ; this%plant_n_to_cwdn               (:)  =nan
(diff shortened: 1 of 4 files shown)


rgknox:
worked with @pnlfang on these fixes

bishtgautam:
@rgknox Is this BFB for FATES?

rgknox:
@bishtgautam this is not b4b with fates active.  It should be b4b with all other tests.

bishtgautam (approved):

rgknox:
I'm noticing that with FATES on, we also may need to zero plant to cwd fluxes, see here:

https://github.com/E3SM-Project/E3SM/blob/master/components/elm/src/data_types/ColumnDataType.F90#L10924

I recommend holding on this PR, as I may need to modify and update some code for that variable when fates is active

rgknox:
I checked through 'plant_to_cwd_nfluxes' and 'plant_to_cwd_pfluxes'.  These variables are not used for anything, so they are benign.  No modifications necessary.

bishtgautam (approved):

rgknox:
@rljacob , also added modification to fix #4186

rljacob:
@bishtgautam make a note on why it was reverted.

bishtgautam:
@rgknox, This PR shows a conflict with the FATES submodule because a more recent hash of FATES is already in the e3sm master (i.e. @amametjanov's fix for IBM). Could you update the hash of the FATES submodule in this PR, so it doesn't show a conflict?

rljacob:
Devops telecon:  needs to be rebased and then remerged.

rgknox:
I've been investigating why #4216 does not pass COMPARE_base_rest, and found that even without an API change with FATES, this test starts to fail when we point the submodule from fates tag sci.1.43.2_api.14.2.0 to sci.1.44.1_api.14.2.0.  None of the FATES side variables are triggering the difference, only the heterotrophic respiration "HR" is showing differences.  My suspicion is that this is something latent in the FATES-ELM coupling, and just hasn't expressed itself until the API change in #4216.  Either way, I'm going to try and work through it on this PR.

rgknox:
Updates were made to the FATES side of the code to enable compare_base_rest to pass.  I released a patch branch for FATES to achieve this.  This patch (in FATES) will also need to be applied to our API 15 tags, to enable resolution of #4216.  These fixes will obviously be merged into FATES master as well (now API 16).

I have no further planned changes to this branch.

rgknox:
with PR #4216 merged, I will update FATES pointer in this PR to API 15. Please hold
@ckoven

PR #4141, opened 2021-03-04
Add required C++ libs for Summit compilers to build E3SM

dqwu:
For Summit, CXX_LINKER is set to FORTRAN by default. With more and
more C++ code added to SCORPIO recently, we need add required C++
libs to build e3sm.exe.

PR #4159, opened 2021-03-16
Add required C++ libs for GNU compiler on Cori to build E3SM

dqwu:
This fix is required for GNU compiler on Cori to build e3sm.exe with latest
release of SCORPIO (not required for Intel compiler).

Similar CXX_LIBS settings have been applied to GNU compilers on Theta
(see PR #3293) and Summit (see PR #4141).


jayeshkrishna (approved):

ndkeen (approved):

ndkeen:
I see that `SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.cori-knl_gnu` will build with this change (and not without it)

PR #4165, opened 2021-03-17
Fix issue in config_pes.ml for oECv3 tests on compy

jonbob:
Fix issue in config_pes.ml that causes problems with oECv3 tests on compy, due to more than one PE layout matching given specs

[BFB]

### DIFF
--- a/cime_config/allactive/config_pesall.xml
+++ b/cime_config/allactive/config_pesall.xml
@@ -9538,9 +9538,9 @@
       </pes>
     </mach>
   </grid>
-  <grid name="a%T62.+_oi%EC30to60E2r2|oi%oEC60to30v3.*">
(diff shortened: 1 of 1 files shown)


mattdturner (approved):
This PR allows `PEM_Ln9_P1024.ne30_oECv3.A_WCYCL1850S.compy_pgi` to PASS now, which previously failed.

It also doesn't cause any of my usual Compy tests to FAIL.

Test `PEM_Ln3.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_intel` FAILs due to a missing input file (`mpas-cice.graph.info.151209.part.80`), but it FAILs on `master` as well.

Approved.

jonbob:
merged to next

jonbob:
merged to master

PR #4167, opened 2021-03-19
Defines core kind for configurations not using CLUBB_SGS

singhbalwinder:
MMF tests were breaking as CLUBB_SGS is not defined for MMF like configurations.
[BFB]

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -129,7 +129,7 @@ _TESTS = {
             "SMS.T62_oQU120_ais20.MPAS_LISIO_TEST",
             "SMS.f09_g16_a.IGELM_MLI",
             "SMS_P12x2.ne4_oQU240.A_WCYCL1850.allactive-mach_mods",
-            "ERS_Ln9_P96x1.ne4pg2_ne4pg2.F-MMF1.eam-crmout",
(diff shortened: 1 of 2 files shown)


brhillman (changes requested):
This looks great. One request, can we add `ERS_Ln9_P96x1.ne4pg2_ne4pg2.F-MMF1.eam-crmout` to `e3sm_developer`, or else move it from `e3sm_integration` to `e3sm_developer` so that this is easier to catch before merging PRs in the future?

singhbalwinder:
I have moved the MMF test from the integration to the developer suite.

brhillman (approved):
Looks great, thanks Balwinder!

rljacob:
Going back to when those tests last ran, they only take about 2.5 minutes each so 1 won't add much to developer.

brhillman:
Hey @singhbalwinder, there's a problem with moving that MMF test to the developer suite. Since it explicitly requests 96 MPI tasks, this will fail on our test platforms that do not have this many processors. I think we should remove the task specification from the test, so the test should instead be just: `ERS_Ln9.ne4pg2_ne4pg2.F-MMF1.eam-crmout`. What do you think?

brhillman (approved):

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #4183, opened 2021-03-29
Correct melt potential in ocean coupling file for MPAS-Ocean

proteanplanet:
This PR corrects the sea ice melt potential being passed from the ocean model to the sea ice model in E3SM so that it references the MPAS-Ocean's salinity-dependent freezing temperature in ocn_comp_mct.F and makes the melt potential coupling consistent with MPAS-SeaIce.  The code change has been tested in a 170-year run on Chrysalis and produces a comparable climate to a simulation without this bug fix. The change is nevertheless required for consistency with mushy-layer sea ice thermodynamics.   Melt potential is calculated as the energy in the first layer of the ocean above freezing temperature available to melt sea ice.  Refer to simulation: https://acme-climate.atlassian.net/wiki/spaces/EWCG/pages/2477457498/20210316.v2beta3GM900.OcnMctIce.ne30pg2+oECv3.chrysalis for more information.  No changes are required in MPAS-Ocean to match this change in E3SM.

[non-BFB]

### DIFF
--- a/components/mpas-ocean/driver/ocn_comp_mct.F
+++ b/components/mpas-ocean/driver/ocn_comp_mct.F
@@ -2306,7 +2306,7 @@ contains
 
    integer, dimension(:), pointer :: landIceMask
 
-   real (kind=RKIND), dimension(:), pointer :: seaIceEnergy, freezingTemp, accumulatedFrazilIceMass, frazilSurfacePressure, &
+   real (kind=RKIND), dimension(:), pointer :: seaIceEnergy, accumulatedFrazilIceMass, frazilSurfacePressure, &
                                                filteredSSHGradientZonal, filteredSSHGradientMeridional, &
                                                avgCO2_gas_flux, DMSFlux, surfaceUpwardCO2Flux, &
                                                avgOceanSurfaceDIC, &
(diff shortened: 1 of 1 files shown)


akturner (approved):

maltrud:
@proteanplanet i have a couple of questions.  1) we form frazil not just in the top layer, but that is being ignored here, right?  2) there are nonlinearities (which are probably small) that are also being ignored.  for example, freezingTemp(<SST>), where <> is temporal average, is not the same as <freezingTemp(SST)>, right?  also, when multiplying by thickness, this uses the most current value (not an average), and again there is a nonlinearity here that is ignored.

proteanplanet:
@maltrud  I understand your questions - I initially thought along the same lines as you until working through it again on the weekend.  Frazil is currently formed in the top 100m of the ocean in MPAS.  However, this change is to the melting potential, not the freezing potential.  That is why seaIceEnergy, which is the melting potential, is zero if there is an accumulated frazil mass.  All this field is doing is providing the amount of energy potentially available for melting at the coupler timestep.  It does not need to accumulate energy available for melting over the coupling period, because that naturally accumulates in the ocean heat content as part of the ocean integration, including melting frazil if it is available to be melted within mpas_ocn_frazil_forcing.F.  All the meltp field is doing is saying "at this coupling timestep, this is the temperature of the upper ocean layer, and as a result, this is the amount of energy that could potentially be used for melting in the forthcoming coupler period in the ocean layer immediately in contact with sea ice".   I attach here an example of the seaIceEnergy (i.e. meltp) field from the coupler:

InteRFACE1alphaC cpl hi 0036-01-09-28800 o2x_Fioo_meltp nh

As a consequence, all nonlinearities are being taken into account here without need for further calculations.

maltrud (approved):

maltrud:
@proteanplanet thanks for the detailed explanation.  i'll approve.

jonbob:
merged to next

jonbob:
merged to master and expected DIFFs blessed -- except two mappy tests that failed due to an issue with PR #4185 and will need to be blessed once that PR is fixed

PR #4185, opened 2021-03-30
Update mpas-source: added sea ice idealized testcases

akturner:
Added series of idealized standalone sea ice test cases
Minor modification of namelist options to accommodate test cases
Removed unused test case code and files from MPAS_Seaice and build system.

[NML]
[BFB]

### DIFF
--- a/components/mpas-seaice/bld/build-namelist
+++ b/components/mpas-seaice/bld/build-namelist
@@ -534,6 +534,7 @@ add_default($nl, 'config_nSnowLayers');
 # Namelist group: initialize #
 ##############################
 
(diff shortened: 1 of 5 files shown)


jonbob (approved):
Approved based on sanity testing and visual inspection

jonbob:
@akturner - one quick question. I see that the earth radius is now a namelist variable. Is the default the same as it has been in E3SM? I'm just trying to make sure answers will be BFB

jonbob:
sanity tests passed, as did BFB check -- merged to next

jonbob:
re-merged to next

jonbob:

PR #4188, opened 2021-03-30
Add default PE settings for mappy and sandiatoss3

jgfouca:
Recent CIME changes mean that CIME is picking up config_pesall.xml for
additional cases as it should. This caused peylayouts to change on a couple test machines for
cases that were mistakenly looking at the driver-only config_pes.xml.
But those were better layouts.  This PR will restore them.

[NML]


rljacob (approved):

jgfouca:
Skipping integration.

PR #4145, opened 2021-03-05
Refactor MMF physics package

whannah1:
This encapsulates a major refactor of the MMF physics package in order to make the MMF less sensitive to E3SM development. Several major changes were made to accomplish this and reduce the amount of duplicate code:
  - Switch the PBL diffusion scheme to the older Holtslag-Boville (HB) option (MMF has it's own version now)
  - Disable all "offline" physics calculations in tphysbc for MMF physpkg.F90
  - Allow the PBL calculations in tphysac to diffuse DSE and qv
  - re-enabled surface flux smoothing by default (needed for stability with HB)
  - 2-mom MMF microphysics option only allowed with ECPP

Some other major changes and clean-up were convenient to include here:
  - 2-mom+ECPP compset is fixed (although still not included in integration suites)
  - MMF time step setting fixed (no need for genmmf test modifier)
  - MAML option removed (Jungmin is working on new version)
  - turbulent mountain stress enabled by default for MMF (for both stability and realism)

Other noteworthy things:
  - ESMT and variance transport are still disabled by default (validation work is ongoing)
  - changes to srf_flux_avg routines are mostly cosmetic - originally I had larger, functional changes to these routines, but decided to roll them back and keep the cosmetic clean up. I also changed the smoothing timescale back to the original 2 hours.
  
  [BFB] for non-MMF
  [NBFB] for all MMF compsets

PR #4196, opened 2021-04-01
Enable ieflx_opt=3 for water temperature heat flux from ocean

wlin7:
Internal energy carried by precipitation is accounted for in the ocean,
but not in the atmosphere. This implementation provides the option to
distribute globally uniformly the apparent heat flux from the ocean due
to the disparity in treating precipitation water temperature.

[BFB] for existing F and B configurations, i.e., ieflx=0 or 2.

### DIFF
--- a/components/eam/src/control/camsrfexch.F90
+++ b/components/eam/src/control/camsrfexch.F90
@@ -94,6 +94,7 @@ module camsrfexch
      real(r8), allocatable :: lwup(:)       ! longwave up radiative flux
      real(r8), allocatable :: lhf(:)        ! latent heat flux
      real(r8), allocatable :: shf(:)        ! sensible heat flux
+     real(r8), allocatable :: h2otemp(:)    ! water temperature heat flux from ocean
      real(r8), allocatable :: wsx(:)        ! surface u-stress (N)
      real(r8), allocatable :: wsy(:)        ! surface v-stress (N)
      real(r8), allocatable :: tref(:)       ! ref height surface air temp
@@ -196,6 +197,9 @@ CONTAINS
        allocate (cam_in(c)%shf(pcols), stat=ierror)
        if ( ierror /= 0 ) call endrun('HUB2ATM_ALLOC error: allocation error shf')
 
+       allocate (cam_in(c)%h2otemp(pcols), stat=ierror)
+       if ( ierror /= 0 ) call endrun('HUB2ATM_ALLOC error: allocation error h2otemp')
+
        allocate (cam_in(c)%wsx(pcols), stat=ierror)
        if ( ierror /= 0 ) call endrun('HUB2ATM_ALLOC error: allocation error wsx')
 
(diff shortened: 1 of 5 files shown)


wlin7:
WIP: waiting until companion PRs for MPAS-O and cime are merged.

jonbob (approved):
Looks good from a visual inspection, and a five-year test with these changes looks good

oksanaguba:
@wlin7  --my understanding is that instead of Tsurf or TBOT some other temperature from the ocean will be used in option (3) of influx. What is the difference between Tsurf and that new temperature? thanks!

golaz (approved):
Based on visual inspection.

kaizhangpnl (approved):
It seems that the heat flux calculation is done in the ocean model. Everything else looks good to me.

wlin7:
> From @kaizhangpnl : It seems that the heat flux calculation is done in the ocean model. Everything else looks good to me.

Yes, @kaizhangpnl . It is computed in mpas-o and imported to atm via coupler.

> From @oksanaguba : my understanding is that instead of Tsurf or TBOT some other temperature from the ocean will be used in option (3) of influx. What is the difference between Tsurf and that new temperature?

Deferring to @jonbob and @mattdturner . Jon, Mat, can you explain how it is computed? Thanks.

jonbob:
At LANL we are rich in Mat/Matts -- for this one we need @maltrud

singhbalwinder:
Since this PR is approved by all and the changes look good to me too, is it okay if I merge this to next? Do we want to add a test for this new option?

wlin7:
> At LANL we are rich in Mat/Matts -- for this one we need @maltrud

Yes, we always know. Thanks, @jonbob . I didn't realize I pick a Matt when github popped up the choices.

whannah1:
Per our slack conversation, you can remove the changes to the MMF version of camsrfexch.F90 since that duplicate file will be deleted by PR #4145

golaz:
@oksanaguba : replying to your comment above. The temperature is the same (Tsurf = SST over ocean), but the key difference is that we are only doing the calculation for water leaving/entering the ocean. Previously, we were computing it for all the water leaving/entering the atmosphere. As a result, we were overestimating the magnitude of the correction (and possibly compensating for the missing SW term in the ocean).

oksanaguba:
@golaz thanks! do you know what's the difference between the old influx (which was 0.4?) and the new one or not yet? what will happen to the land fluxes in the long run?

oksanaguba:
by the long run I meant future, not long simulations.

golaz:
@oksanaguba : the land model doesn't know about the water temperature, therefore there should not be a correction for water falling on and evaporating from land. But there should be a correction for water falling on land and running off to the ocean. MPAS-O does include that term.

golaz:
@rljacob : another PR we need for b2beta4. Can this go in?

rljacob:
So this will be BFB?  Sure we can add it tomorrow.  Too late for testing tonight.

singhbalwinder:
Merged to next

rljacob:
@singhbalwinder did you push your copy of next?  I don't see your merge.

singhbalwinder:
Thanks @rljacob for noticing that. I think next was updated before I pushed my changes, so they got rejected. I will try again in a few minutes.

singhbalwinder:
I have now merged this branch with next.

golaz:
@singhbalwinder : is this ready to be merged to master?

PR #4202, opened 2021-04-03
Adds a print message if a fatal error occurs inside CLUBB core codes

singhbalwinder:
If CLUBB core codes experience a fatal error, the added print messages
print out location of the error (lat, lon and column) and the time step
at which the error occured.

[BFB] - Bit-For-Bit

### DIFF
--- a/components/eam/src/physics/cam/clubb_intr.F90
+++ b/components/eam/src/physics/cam/clubb_intr.F90
@@ -22,7 +22,7 @@ module clubb_intr
   use ppgrid,        only: pver, pverp
   use phys_control,  only: phys_getopts
   use physconst,     only: rair, cpair, gravit, latvap, latice, zvir, rh2o, karman, &
-                           tms_orocnst, tms_z0fac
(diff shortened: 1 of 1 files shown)


singhbalwinder:
@golaz and @rljacob : @vlarson would like to get this change into v2, if possible. This is a pretty minor BFB modification but can help the CLUBB developers debug the code in case the code experiences a fatal crash inside CLUBB core codes.

singhbalwinder:
Please turn on "Hide whitespace changes" options as I removed a lot of white spaces from clubb_intr.F90 file.

vlarson (approved):
Thank you, Balwinder.  Looks good.  

This new error message will be helpful if E3SM crashes inside CLUBB.  It will avoid the need to re-run E3SM in order to determine the grid column in which the crash occurred.

rljacob (approved):

wlin7:
Merged to next.

wlin7:
Merged to master

PR #4203, opened 2021-04-06
Fixes a bug in rad diags calculations due to prescribed volcanic emissions

singhbalwinder:
Radiation diagnostic calculations were producing different
answers even for the exact same list of aerosols mentioned in
the namelist in rad_diag_1 and rad_diag_2 fields. This bug
appears only for the compsets which turn on the prescribed
volcanic emissions (e.g. F2010). The bug doesn't affect other
compsets (e.g. FC5AV1C-L).
The reason was the use of unit conversion factor (km to m)
which got applied repeatedly for each diagnostic call. This bug
*does not* affect any simulation which doesn't use repeated
radiation diagnostic calls via rad_diag_1 etc. in the namelist

[BFB] - Bit-For-Bit

### DIFF
--- a/components/eam/src/physics/cam/aer_rad_props.F90
+++ b/components/eam/src/physics/cam/aer_rad_props.F90
@@ -2,7 +2,7 @@ module aer_rad_props
 
 !------------------------------------------------------------------------------------------------
 ! Converts aerosol masses to bulk optical properties for sw and lw radiation
-! computations.  
+! computations.
 !------------------------------------------------------------------------------------------------
 
 use shr_kind_mod,     only: r8 => shr_kind_r8
@@ -38,7 +38,7 @@ public :: &
    aer_rad_props_lw             ! return LW optical props of aerosols
 
 ! Private data
-real(r8), parameter :: km_inv_to_m_inv = 0.001_r8      !1/km to 1/m 
+real(r8), parameter :: km_inv_to_m_inv = 0.001_r8      !1/km to 1/m
 character(len=fieldname_len), pointer :: odv_names(:)  ! outfld names for visible OD
 integer  :: idx_ext_sw, idx_ssa_sw, idx_af_sw, idx_ext_lw !pbuf indices for volcanic cmip6 file
 integer  :: ihuge ! a huge integer
@@ -49,7 +49,7 @@ contains
 
 subroutine aer_rad_props_init()
    use phys_control, only: phys_getopts
-   
+
    !Local variables
    integer                    :: i, ierr
    integer                    :: numaerosols  ! number of aerosols
@@ -62,7 +62,7 @@ subroutine aer_rad_props_init()
 
    rhuge = huge(1.0_r8)
    ihuge = huge(1)
-   
+
    call phys_getopts( history_aero_optics_out    = history_aero_optics, &
                       history_amwg_out           = history_amwg,    &
                       prog_modal_aero_out        = prog_modal_aero )
@@ -72,7 +72,7 @@ subroutine aer_rad_props_init()
 
    call addfld ('AEROD_v', horiz_only, 'A', '1', &
       'Total Aerosol Optical Depth in visible band', flag_xyfill=.true.)
-   
+
    !For testing puposes only, the following addfld call should be removed before merging to master
    call addfld ('extinct_lw_bnd7',(/ 'lev' /),    'A','1/m','EXTINCT LW H2O window band 7 output', flag_xyfill=.true.)
    call addfld ('extinct_lw_inp',(/ 'lev' /),    'A','1/km',&
@@ -99,11 +99,11 @@ subroutine aer_rad_props_init()
    end do
 
    ! Determine default fields
-   if (history_amwg ) then 
+   if (history_amwg ) then
       call add_default ('AEROD_v', 1, ' ')
-   endif   
-   
-   if ( history_aero_optics ) then 
+   endif
+
+   if ( history_aero_optics ) then
       call add_default ('AEROD_v', 1, ' ')
       do i = 1, numaerosols
          odv_names(i) = 'ODV_'//trim(aernames(i))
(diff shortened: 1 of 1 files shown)


singhbalwinder:
Thanks to Mingxuan for finding this issue.

singhbalwinder:
Please select "hide white spaces" when reviewing the changes as I have removed the extra white spaces.

@golaz and @rljacob : This PR fixes a bug in radiation diagnostic calls, so should be included in v2, if possible.

tangq:
It seems to me that this bug will affect the PD&PI estimation of the radiative effects for the v2 test runs.

components/eam/src/physics/cam/aer_rad_props.F90
```diff
@@ -231,22 +231,21 @@ subroutine aer_rad_props_sw(list_idx, dt, state, pbuf,  nnite, idxnite, is_cmip6
...
+   !Special treatment for CMIP6 volcanic aerosols, where extinction, ssa
    !and af are directly read from the prescribed volcanic aerosol file
 
-   !Point ext_cmip6_sw to unphysically huges values so that the code breaks if they are used for is_cmip6_volc=.false. case
+   !Point ext_cmip6_sw to null so that the code breaks if they are used for is_cmip6_volc=.false. case
    !This is done to avoid having optional arguments in modal_aero_sw call
-   rhuge_arr = rhuge
-   ext_cmip6_sw => rhuge_arr
-   trop_level = ihuge
+   ext_cmip6_sw => null()
```
kaizhangpnl, line 239: Remove rhuge as well?

kaizhangpnl:
Hi @singhbalwinder this bug seems that all use cases with "CMIP6" (https://github.com/E3SM-Project/E3SM/tree/master/components/eam/bld/namelist_files/use_cases) are affected, right? Also, maint-1.0 needs to be fixed too.

singhbalwinder:
Thanks Kai and Qi for your review. I have removed the `rhuge` and redefined `ihuge`. 

You are right, this fix will affect all CMIP6 "use cases" (since they all use prescribed volcanic aerosols) if we have used `rad_diag` (radiation diagnostics) namelist variable in the namelist. Simulations which didn't use `rad_diag` should remain BFB.

singhbalwinder:
I will fix it in the maint-1.0 branch as well.

Please let me know if you would like me to run PI-PD simulations to see impact of this change.

mingxuanwupnnl:
Thank you, Balwinder. I tested the fix, and found that the simulation would not be BFB when rad_diag_* are in the namelist. As you pointed out, we might also need to change radiation.F90 in rrtmgp/rrtmg, right?

mingxuanwupnnl (commented):
The fix in aer_rad_props.F90 looks good to me. However, the model would not be BFB when rad_diag_* are in the namelist. We may also fix that.

kaizhangpnl:
> I will fix it in the maint-1.0 branch as well.
> 
> Please let me know if you would like me to run PI-PD simulations to see impact of this change.

Yes, it would be nice to evaluate the impact. Maybe we can use your previous simulations as a reference (with the bug)? I can help with the post-processing and analysis.

golaz:
@singhbalwinder and @rljacob : I agree, the should be merged to be in the v2 code base.

wlin7:
> The fix in aer_rad_props.F90 looks good to me. However, the model would not be BFB when rad_diag_* are in the namelist. We may also fix that.

Hi @mingxuanwupnnl , do you mean we need additional code changes? The bug is only with the conversion factor being repeatedly multiplied to the pbuf variables, which is now fixed with new work variables. I feel like no other changes needed, but I could be wrong as I haven't run such rad_diags_* tests.

singhbalwinder:
I also believe we do not need any additional changes.

@mingxuanwupnnl : I think even if we switch the ordering of the do-loop in radiation.F90:
`do icall = N_DIAG, 0, -1`
the results will still not be BFB with simulations which used `rad_diags_*` in the namelist in the simulations which has the bug. 

In the buggy simulations, the unit conversion multiplier will be applied at least twice (depending on how many `rad_diag_*` we have in the namelist) to the prognostic (`icall=0`) calculations. If we revert the order of the do-loop, this multiplier will be applied exactly once. Therefore the results will be non-BFB even if we change the order. Does that make sense?

I am running few simulations to access the impact of this bug fix.

mingxuanwupnnl:
@singhbalwinder @wlin7 I'm a little confused here. Just try to check my understanding. 

When this fix is applied, it is BFB (compared with the code without fix) when rad_diag_* are not used.  

For two simulations which both apply this fix, the one that uses rad_diag_* has different values for prognostic vars (e.g., T, U, V) compared with the one that does not use rad_diag_*. Do we need to fix that? Or there are always differences for prognostic vars between with and without rad_diag_*?

singhbalwinder:
All prognostic variables (such as T,P, U, Q, MMRs etc.) should be BFB for simulations whether we use `rad_diag_*` in the namelist or not.

Unfortunately, this bug was causing the prognostic variables to differ between the simulations run with and without rad_diag_*. It is only triggered if we turn on prescribed volcanic emission. With the bug fix in place, now the prognostic variables will be BFB in simulations run with or without rad_diag_*.

This is how I understand this code, please correct me if I got something wrong.

The reason is the way do-loop executes in radiation.F90:
`do icall = N_DIAG, 0, -1`

In a simulation with the bug, the unit conversion is applied for each iteration of this loop (which is wrong as it should be applied only once). Therefore the prognostic call (`icall=0`), which is the last one, will have unit conversion factor multiplied `N_DIAG` times. That is the reason why the prognostic variables will be different in simulation where `N_DIAG=0` (i.e. no `rad_diag_*` in the namelist) vs. `N_DIAG>0` (i.e. one or more `rad_diag_*` in the namelist).

In bug fixed simulations, unit conversion is applied only once (using a work variable instead of the `pbuf` variable) therefore the prognostic variables will not depend on the value of `N_DIAG`. 

Since using a work variable fixes this issue, we do not need to change the order of the do-loop. I haven't looked into the code but there may be other parts of the code which may be  dependent on this do-loop order. If we change the order, we might risk breaking that unless we rigorously evaluate the model after changing the order.

wlin7:
Hi @kaizhangpnl , @mingxuanwupnnl , can you please confirm the PR is ready to go by approving the changes? Thanks.

kaizhangpnl (approved):

wlin7:
Merged to next.

mingxuanwupnnl (approved):

wlin7:
Merged to master.

PR #4205, opened 2021-04-06
Adding 'CRYO1850' compset

darincomeau:
Adds a new Cryosphere compset that is similar to `WCYCL1850`, except uses Cryosphere settings for Antarctic runoff (disables Antarctic runoff from the coupler to the ocn, includes ice-shelf melt fluxes and data icebergs). This compset points to spun-up ocn/ice initial conditions (restarts from a one month G-case).

### DIFF
--- a/cime_config/allactive/config_compsets.xml
+++ b/cime_config/allactive/config_compsets.xml
@@ -316,14 +316,12 @@
   <lname>20TR_EAM%CMIP6_ELM%CNPECACNTBC_MPASSI%BGC_MPASO%OIECOOIDMS_MOSART_SGLC_SWAV_BGC%BDRD</lname>
 </compset>
 
-<!-- E3SM v2 cryosphere compsets -->
+<!-- E3SM cryosphere compsets -->
 
 <compset>
   <alias>CRYO1850</alias>
   <lname>1850SOI_EAM%CMIP6_ELM%SPBC_MPASSI%DIB_MPASO%IBISMF_MOSART_SGLC_SWAV</lname>
 </compset>
-
-<!-- E3SM v1 cryosphere compsets -->
 
 <compset>
   <alias>A_WCYCL1850-DIB</alias>
--- a/cime_config/allactive/config_compsets.xml
+++ b/cime_config/allactive/config_compsets.xml
@@ -316,12 +316,14 @@
   <lname>20TR_EAM%CMIP6_ELM%CNPECACNTBC_MPASSI%BGC_MPASO%OIECOOIDMS_MOSART_SGLC_SWAV_BGC%BDRD</lname>
 </compset>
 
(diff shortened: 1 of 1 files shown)


darincomeau:
Tested a `CRYO1850.ne30pg2_SOwISC12to60E2r4` configuration on Cori-KNL and confirmed the intended namelist/streams defaults mentioned in the above comment are correct.

darincomeau:
This compset is intended to replace the following compset for v2 Cryosphere simulations:
`<compset>
  <alias>A_WCYCL1850-DIB-ISMF_CMIP6</alias>
  <lname>1850_EAM%CMIP6_ELM%SPBC_MPASSI%DIB_MPASO%IBISMF_MOSART_SGLC_SWAV</lname>
</compset>`

stephenprice:
This looks fine to me @darincomeau. Note that this is filed under "". Do we want to put it in a v2 section instead (or should we be adding a v2 section if there isn't one already?).

darincomeau:
Good point @stephenprice - I placed it under the `<!-- E3SM v1 cryosphere compsets -->` heading to keep it alongside the other cryo compsets. I was hesitant to add a new `<!-- E3SM v2 cryosphere compsets -->` heading since the other simulation campaigns have not done so yet, but rather have them under `<!-- E3SM CMIP6 COMPSETS science compsets -->`.

stephenprice:
Maybe the v1 vs. v2 distinction is not so important? Right now we don't have that many cryo specific compsets so I agree it may be better to organize by "cryo" rather than by model version. I'm fine w/ whatever you decide. @jonbob may have opinions on this as well, which I'll obviously defer to.

stephenprice (approved):

jonbob:
@darincomeau and @stephenprice - I don't think it matters much but it might be more accurate to have this compset under a new v2 heading. Or include it in something like the "E3SM CMIP6 COMPSETS science compsets" part near the top, but include cryo in that new heading? Mostly I worry that putting it in a v1 compset list is not quite right. But it's really your call

darincomeau:
@jonbob sure - I just pushed with a new compset group heading.

jonbob:
Thanks @darincomeau - I do think that's more clear

jonbob (approved):
Approved based on visual inspection and developer testing

stephenprice:
looks good to me too

rljacob:
Just remove "v1" or "v2" and call them "E3SM cryosphere compsets".

@jonbob can you add a test with this compset to the e3sm_prod test suite?

darincomeau:
Thanks for the suggestion @rljacob - I've made that change.

jonbob:
@rljacob - yes, I'll make a PR to enhance prod testing to include this compset as well as the JRA one that had failed due to the CIME issue

jonbob:
merged to master

PR #4206, opened 2021-04-07
Set GMAKE_J to 4 for 'singularity' machine

lukaszlacinski:
By default, GMAKE assumes 16 threads/CPUs available on the 'singularity' machine. It sometimes causes a failure when building kokkos on hardware that has less CPUs like small workstations and laptops. Then, a user is required to set explicitly GMAKE_J to a value that matches the user's hardware. To avoid that additional step, this PR changes GMAKE_J default value, 16, to 4 which should work for most small workstations and laptops.


rljacob (approved):

rljacob:
@lukaszlacinski the feature branch for this PR doesn't seem to exist anymore in your fork.

lukaszlacinski:
It looks like a deleted the fork repo about a month ago. I just re-created the repo with the PR, https://github.com/E3SM-Project/E3SM/pull/4206.

PR #4207, opened 2021-04-07
Update CIME, add new field h2otemp from mpas-o

jonbob:
Change CIME submodule to point to branch based off of master 81d4f2033b5cbe7f (CIME 5.8.41)   Pass a new field, h2otemp, from the ocn to the atm. The CIME changes also include adding this new field to the cpl heat budget. In addition, bring in the ocn driver code that exports the field to the cpl. It should not change answers for active-model configurations until the correspond atm PR that uses the new field is merged.

Other CIME changes since last update:
Save SCORPIO I/O stats
CESM-only changes:  ESMF threading, izumi config

[non-BFB] for X-cases, since the number of cpl fields will increase
All cases with DATM or EAM will say "DIFF" because number of fields changes.

### DIFF
--- a/components/mpas-ocean/driver/mpaso_cpl_indices.F
+++ b/components/mpas-ocean/driver/mpaso_cpl_indices.F
@@ -18,6 +18,7 @@ module mpaso_cpl_indices
   integer :: index_o2x_So_dhdy
   integer :: index_o2x_Fioo_meltp
   integer :: index_o2x_Fioo_frazil
+  integer :: index_o2x_Faoo_h2otemp
   integer :: index_o2x_Faoo_fco2_ocn
   integer :: index_o2x_Faoo_fdms_ocn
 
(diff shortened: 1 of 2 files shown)


golaz:
@rljacob and @jgfouca : it would be useful to get this in quickly. We need this for v2beta4 that I'd like to start before the weekend and I would prefer to use master for it. Thanks.

rljacob:
Is "the atm PR that uses the new field" PR #4196?   The description says that will still be BFB

rljacob:
What happens with F-cases for this new field?

rljacob (approved):

jonbob:
Yes, sorry -- the atm PR will still be BFB because this new option is not the default. And I'll test an F-case and post the results

jonbob:
@rljacob - ERS.ne4_ne4.FC5AV1C-L.anvil_intel.eam-thetanh_ftype2 passed just fine

jgfouca (approved):

jonbob:
merged to next

golaz:
@rljacob : is this ready to be merged to master?

rljacob:
Merged to master.

golaz:
Thanks @rljacob. By any chance, can you also merge https://github.com/E3SM-Project/E3SM/pull/4196?

PR #4210, opened 2021-04-08
Add script support for ECwISC30to60E2r1 ocn/ice grid

darincomeau:
Brings in support for the new ECwISC30to60E2r1 ocn/ice grid, which is the Cryosphere v2 low-resolution configuration. This includes all necessary mapping, runoff, and domain files for the following configurations:

ne30pg2_ECwISC30to60E2r1
T62_ECwISC30to60E2r1 (CORE-forced)
TL319_ECwISC30to60E2r1 (JRA-forced)

All mapping/domain/grid files have been placed in the E3SM public inputdata directory and are world-readable.

I've run 5 day smoke tests on the following compset/resolution configurations on Anvil:

`CRYO1850.ne30pg2_ECwISC30to60E2r1`
`GMPAS-JRA1p4-DIB-ISMF.TL319_ECwISC30to60E2r1`
`GMPAS-DIB-IAF-ISMF.T62_ECwISC30to60E2r1`

`WCYCL1850.ne30pg2_ECwISC30to60E2r1`
`GMPAS-IAF.T62_ECwISC30to60E2r1`

This mesh is similar to the `EC30to60E2r2` mesh, except this mesh includes Antarctic ice-shelf cavities, and uses GEBCO bathymetry.

### DIFF
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -2722,7 +2722,7 @@
       <nx>237984</nx>
       <ny>1</ny>
       <file grid="ice|ocn">$DIN_LOC_ROOT/share/domains/domain.ocn.ECwISC30to60E2r1.201007.nc</file>
-      <desc>ECwISC30to60E2r1 is a MPAS ocean grid generated with the jigsaw/compass process using the eddy closure density function that is roughly comparable to the pop 1 degree resolution:</desc>
+      <desc>ECwISC30to60E2r1 is a MPAS ocean grid generated with the jigsaw/compass process using the eddy closure density function that has 30 km gridcells at the equator, 60 km at mid-latitudes, and 35 km at high latitudes. Additionally, it has ocean under landice cavities:</desc>
     </domain>
 
     <!-- ROF (river) grids-->
@@ -3180,7 +3180,7 @@
 
     <gridmap atm_grid="ne30np4.pg2" ocn_grid="ECwISC30to60E2r1">
       <map name="ATM2OCN_FMAPNAME">cpl/gridmaps/ne30pg2/map_ne30pg2_to_ECwISC30to60E2r1_mono.201006.nc</map>
-      <map name="ATM2OCN_VMAPNAME">cpl/gridmaps/ne30pg2/map_ne30pg2_to_ECwISC30to60E2r1_mono.201006.nc</map>
+      <map name="ATM2OCN_VMAPNAME">cpl/gridmaps/ne30pg2/map_ne30pg2_to_ECwISC30to60E2r1_bilin.210409.nc</map>
       <map name="ATM2OCN_SMAPNAME">cpl/gridmaps/ne30pg2/map_ne30pg2_to_ECwISC30to60E2r1-nomask_bilin.201006.nc</map>
       <map name="OCN2ATM_FMAPNAME">cpl/gridmaps/ECwISC30to60E2r1/map_ECwISC30to60E2r1-nomask_to_ne30pg2_mono.201006.nc</map>
       <map name="OCN2ATM_SMAPNAME">cpl/gridmaps/ECwISC30to60E2r1/map_ECwISC30to60E2r1-nomask_to_ne30pg2_mono.201006.nc</map>
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -450,6 +450,16 @@
       <mask>SOwISC12to60E2r4</mask>
     </model_grid>
 
+    <model_grid alias="T62_ECwISC30to60E2r1" compset="(DATM|XATM|SATM)">
+      <grid name="atm">T62</grid>
+      <grid name="lnd">T62</grid>
+      <grid name="ocnice">ECwISC30to60E2r1</grid>
+      <grid name="rof">rx1</grid>
+      <grid name="glc">null</grid>
+      <grid name="wav">null</grid>
+      <mask>ECwISC30to60E2r1</mask>
+    </model_grid>
+
     <model_grid alias="TL319_oEC60to30v3" compset="(DATM|XATM|SATM)">
       <grid name="atm">TL319</grid>
       <grid name="lnd">TL319</grid>
@@ -538,6 +548,16 @@
       <grid name="glc">null</grid>
       <grid name="wav">null</grid>
       <mask>SOwISC12to60E2r4</mask>
+    </model_grid>
+
+    <model_grid alias="TL319_ECwISC30to60E2r1" compset="(DATM|XATM|SATM)">
+      <grid name="atm">TL319</grid>
+      <grid name="lnd">TL319</grid>
+      <grid name="ocnice">ECwISC30to60E2r1</grid>
+      <grid name="rof">JRA025</grid>
+      <grid name="glc">null</grid>
+      <grid name="wav">null</grid>
+      <mask>ECwISC30to60E2r1</mask>
     </model_grid>
 
     <!-- finite volume grids -->
@@ -1202,6 +1222,16 @@
       <grid name="glc">null</grid>
       <grid name="wav">null</grid>
       <mask>SOwISC12to60E2r4</mask>
+    </model_grid>
+
+    <model_grid alias="ne30pg2_ECwISC30to60E2r1">
+      <grid name="atm">ne30np4.pg2</grid>
+      <grid name="lnd">ne30np4.pg2</grid>
+      <grid name="ocnice">ECwISC30to60E2r1</grid>
+      <grid name="rof">r05</grid>
+      <grid name="glc">null</grid>
+      <grid name="wav">null</grid>
+      <mask>ECwISC30to60E2r1</mask>
     </model_grid>
 
     <model_grid alias="northamericax4v1_r0125_northamericax4v1" compset="(DOCN|XOCN|SOCN|AQP1)">
@@ -2263,6 +2293,7 @@
       <file grid="atm|lnd" mask="WC14to60E2r3">$DIN_LOC_ROOT/share/domains/domain.lnd.T62_WC14to60E2r3.200929.nc</file>
       <file grid="atm|lnd" mask="WCAtl12to45E2r4">$DIN_LOC_ROOT/share/domains/domain.lnd.T62_WCAtl12to45E2r4.210318.nc</file>
       <file grid="atm|lnd" mask="SOwISC12to60E2r4">$DIN_LOC_ROOT/share/domains/domain.lnd.T62_SOwISC12to60E2r4.210119.nc</file>
+      <file grid="atm|lnd" mask="ECwISC30to60E2r1">$DIN_LOC_ROOT/share/domains/domain.lnd.T62_ECwISC30to60E2r1.201007.nc</file>
       <desc>T62 is Gaussian grid:</desc>
     </domain>
 
@@ -2301,6 +2332,8 @@
       <file grid="ice|ocn" mask="WCAtl12to45E2r4">$DIN_LOC_ROOT/share/domains/domain.ocn.TL319_WCAtl12to45E2r4.210318.nc</file>
       <file grid="atm|lnd" mask="SOwISC12to60E2r4">$DIN_LOC_ROOT/share/domains/domain.lnd.TL319_SOwISC12to60E2r4.210119.nc</file>
       <file grid="ice|ocn" mask="SOwISC12to60E2r4">$DIN_LOC_ROOT/share/domains/domain.ocn.TL319_SOwISC12to60E2r4.210119.nc</file>
+      <file grid="atm|lnd" mask="ECwISC30to60E2r1">$DIN_LOC_ROOT/share/domains/domain.lnd.TL319_ECwISC30to60E2r1.201007.nc</file>
+      <file grid="ice|ocn" mask="ECwISC30to60E2r1">$DIN_LOC_ROOT/share/domains/domain.ocn.TL319_ECwISC30to60E2r1.201007.nc</file>
       <desc>TL319 is JRA lat/lon grid:</desc>
     </domain>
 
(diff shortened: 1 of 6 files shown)


darincomeau:
Here are  more details on the generation of the initial conditions:

The original mesh file `ocean.ECwISC30to60E2r1.200916.nc` was not fully dynamically adjusted, and would crash after 12 hours of simulation time.

A five day GMPAS-IAF case was run using `ocean.ECwISC30to60E2r1.210407.nc` as input with `config_rayleigh_friction = .true.`. The ocn/ice restarts at the end of this run were used to create `ocean.ECwISC30to60E2r1.200916.nc`, `seaice.ECwISC30to60E2r1.200916.nc`, which are dynamically adjusted (as determined by their ability to stably run with default settings). These are set as the non-spunup, cold-start ICs. To avoid confusion, I renamed `ocean.ECwISC30to60E2r1.210407.nc` > `ocean.ECwISC30to60E2r1.noDynamicAdjustment.210407.nc` in the inputdata directory.

`ocean.ECwISC30to60E2r1.210407.nc` and `seaice.ECwISC30to60E2r1.210407.nc` were then used as ICs for a one month `GMPAS-DIB-IAF-ISMF` case. The restarts from that were used to create the spun-up ICs `mpaso.ECwISC30to60E2r1.rstFromG-anvil.210408.nc`, `mpassi.ECwISC30to60E2r1.rstFromG-anvil.210408.nc`. The `GMPAS-DIB-IAF-ISMF` compset was used for this one month spinup to ensure the restarts would have all required fields to be used as ICs for configurations with active ISMF & DIB.

cime_config/config_grids.xml
```diff
@@ -2683,6 +2718,13 @@
       <desc>SOwISC12to60E2r4 is a MPAS ice/ocean grid with enhanced resolution of 12km in the Southern Ocean around Antarctica. The high resolution regions smoothly transition to the background resolution of the standard low resolution 60to30km grid:</desc>
     </domain>
 
+    <domain name="ECwISC30to60E2r1">
+      <nx>237984</nx>
+      <ny>1</ny>
+      <file grid="ice|ocn">$DIN_LOC_ROOT/share/domains/domain.ocn.ECwISC30to60E2r1.201007.nc</file>
+      <desc>ECwISC30to60E2r1 is a MPAS ocean grid generated with the jigsaw/compass process using the eddy closure density function that is roughly comparable to the pop 1 degree resolution:</desc>
```
xylar, line 2725:
```suggestion
      <desc>ECwISC30to60E2r1 is a MPAS ocean grid generated with the jigsaw/compass process using the eddy closure density function that is roughly comparable to a uniform 0.5 degree resolution:</desc>
```

For one thing, the comparison to POP seems increasingly irrelevant.  For another, the resolution is about 2x a POP 1 degree grid.  I also wonder if we should be giving some description of where the higher and lower res regions are as the previous EC grids did, but don't have strong feelings about it.

cime_config/config_grids.xml
```diff
@@ -3136,6 +3178,14 @@
       <map name="OCN2ATM_SMAPNAME">cpl/gridmaps/SOwISC12to60E2r4/map_SOwISC12to60E2r4-nomask_to_ne30pg2_mono.210119.nc</map>
     </gridmap>
 
+    <gridmap atm_grid="ne30np4.pg2" ocn_grid="ECwISC30to60E2r1">
+      <map name="ATM2OCN_FMAPNAME">cpl/gridmaps/ne30pg2/map_ne30pg2_to_ECwISC30to60E2r1_mono.201006.nc</map>
+      <map name="ATM2OCN_VMAPNAME">cpl/gridmaps/ne30pg2/map_ne30pg2_to_ECwISC30to60E2r1_mono.201006.nc</map>
```
xylar, line 3183: Should this be:
```suggestion
      <map name="ATM2OCN_VMAPNAME">cpl/gridmaps/ne30pg2/map_ne30pg2_to_ECwISC30to60E2r1_bilin.201006.nc</map>
```
This seems to be the case for both `EC30to60E2r2` and `SOwISC12to60E2r4`.

xylar: See this discussion on this meeting page:
https://acme-climate.atlassian.net/wiki/spaces/ECG/pages/2350776353/2-17-2021+Cryosphere+Simulation+Campaign

I think the same applies to all meshes.  Atmospheric velocities should not be interpolated with `mono`/`aave`/`conserve`, but rather with `bilin` or even better something higher order once that's a standard option (e.g. with `mbtempest`).

xylar: The other `ATM2OCN_VMAPNAME` mapping files are `patch`.  I'm not sure what that is, but it's also used for other meshes so I guess it must be non-conservative?

xylar:
@darincomeau, by inspection, everything looks good except possibly the 2 things I commented on.  Are there some tests you suggest I run with this branch?  If so I'm happy to.

cime_config/config_grids.xml
darincomeau: Yes this was clearly a "copy and lightly modify from a similar grid" job on my part. I removed the POP reference and added the description of where the higher and lower res regions were. I don't want to refer to this as comparable to a 0.5 degree grid since previous similar grids refer to a comparable 1 degree grid (while neither are really accurate), so I think resolution description is explicit enough for here. Let me know if you agree (or not!).

cime_config/config_grids.xml
darincomeau: Thanks for catching this @xylar - you're right this certainly should be changed. I've made this new `bilin` mapping file and made the change.

darincomeau:
@xylar if you wouldn't mind running a case with the compset `CRYO1850` and grid  `ne30pg2_ECwISC30to60E2r1`, that would be great.

cime_config/config_grids.xml
```diff
@@ -2683,6 +2718,13 @@
       <desc>SOwISC12to60E2r4 is a MPAS ice/ocean grid with enhanced resolution of 12km in the Southern Ocean around Antarctica. The high resolution regions smoothly transition to the background resolution of the standard low resolution 60to30km grid:</desc>
     </domain>
 
+    <domain name="ECwISC30to60E2r1">
+      <nx>237984</nx>
+      <ny>1</ny>
+      <file grid="ice|ocn">$DIN_LOC_ROOT/share/domains/domain.ocn.ECwISC30to60E2r1.201007.nc</file>
+      <desc>ECwISC30to60E2r1 is a MPAS ocean grid generated with the jigsaw/compass process using the eddy closure density function that has 30 km gridcells at the equator, 60 km at mid-latitudes, and 35 km at high latitudes. Additionally, it has ocean under landice cavities:</desc>
```
xylar, line 2725:
```suggestion
      <desc>ECwISC30to60E2r1 is a MPAS ocean grid generated with the jigsaw/compass process using the eddy closure density function that has 30 km gridcells at the equator, 60 km at mid-latitudes, and 35 km at high latitudes. Additionally, it has ocean in ice-shelf cavities:</desc>
```
I don't think "landice cavities" are really a thing.  And The ocean is in the cavities, under ice shelves so it either needs to be "in ice-shelf cavities" or "under ice shelves".

darincomeau: Another 'copy from elsewhere' job - thanks for noticing, I've made the suggested change.

xylar: Ah, I see!  I'll resist the urge to use the "blame" GitHub feature...

xylar (approved):
With some coaching from @darincomeau, I was able to run a 5-day test with this compset on Chrysalis.  Seems to work just fine!

cime_config/config_grids.xml
```diff
@@ -3136,6 +3178,14 @@
       <map name="OCN2ATM_SMAPNAME">cpl/gridmaps/SOwISC12to60E2r4/map_SOwISC12to60E2r4-nomask_to_ne30pg2_mono.210119.nc</map>
     </gridmap>
 
+    <gridmap atm_grid="ne30np4.pg2" ocn_grid="ECwISC30to60E2r1">
+      <map name="ATM2OCN_FMAPNAME">cpl/gridmaps/ne30pg2/map_ne30pg2_to_ECwISC30to60E2r1_mono.201006.nc</map>
+      <map name="ATM2OCN_VMAPNAME">cpl/gridmaps/ne30pg2/map_ne30pg2_to_ECwISC30to60E2r1_bilin.210409.nc</map>
+      <map name="ATM2OCN_SMAPNAME">cpl/gridmaps/ne30pg2/map_ne30pg2_to_ECwISC30to60E2r1-nomask_bilin.201006.nc</map>
+      <map name="OCN2ATM_FMAPNAME">cpl/gridmaps/ECwISC30to60E2r1/map_ECwISC30to60E2r1-nomask_to_ne30pg2_mono.201006.nc</map>
+      <map name="OCN2ATM_SMAPNAME">cpl/gridmaps/ECwISC30to60E2r1/map_ECwISC30to60E2r1-nomask_to_ne30pg2_mono.201006.nc</map>
```
stephenprice, line 3186: Does it matter that the ocn -> atm state mapping file is conservative while the atm -> ocn state mapping file is not (the former is mono while the latter is just bilinear)? I.e., should those two directions be using a mapping file w/ the same properties for any reason?

cime_config/config_grids.xml
```diff
@@ -3814,6 +3864,14 @@
       <map name="OCN2ATM_SMAPNAME">cpl/gridmaps/SOwISC12to60E2r4/map_SOwISC12to60E2r4_to_T62_aave.210119.nc</map>
     </gridmap>
 
+    <gridmap atm_grid="T62" ocn_grid="ECwISC30to60E2r1">
+      <map name="ATM2OCN_FMAPNAME">cpl/gridmaps/T62/map_T62_to_ECwISC30to60E2r1_aave.201006.nc</map>
+      <map name="ATM2OCN_SMAPNAME">cpl/gridmaps/T62/map_T62_to_ECwISC30to60E2r1-nomask_bilin.201006.nc</map>
+      <map name="ATM2OCN_VMAPNAME">cpl/gridmaps/T62/map_T62_to_ECwISC30to60E2r1_patch.201006.nc</map>
+      <map name="OCN2ATM_FMAPNAME">cpl/gridmaps/ECwISC30to60E2r1/map_ECwISC30to60E2r1_to_T62_aave.201006.nc</map>
+      <map name="OCN2ATM_SMAPNAME">cpl/gridmaps/ECwISC30to60E2r1/map_ECwISC30to60E2r1_to_T62_aave.201006.nc</map>
```
stephenprice, line 3872: same comment as above w.r.t. differing types of mapping file use for to / from state

cime_config/config_grids.xml
```diff
@@ -3886,6 +3944,14 @@
       <map name="OCN2ATM_SMAPNAME">cpl/gridmaps/SOwISC12to60E2r4/map_SOwISC12to60E2r4_to_TL319_aave.210119.nc</map>
     </gridmap>
 
+    <gridmap atm_grid="TL319" ocn_grid="ECwISC30to60E2r1">
+      <map name="ATM2OCN_FMAPNAME">cpl/gridmaps/TL319/map_TL319_to_ECwISC30to60E2r1_aave.201006.nc</map>
+      <map name="ATM2OCN_SMAPNAME">cpl/gridmaps/TL319/map_TL319_to_ECwISC30to60E2r1-nomask_bilin.201006.nc</map>
+      <map name="ATM2OCN_VMAPNAME">cpl/gridmaps/TL319/map_TL319_to_ECwISC30to60E2r1_patch.201006.nc</map>
+      <map name="OCN2ATM_FMAPNAME">cpl/gridmaps/ECwISC30to60E2r1/map_ECwISC30to60E2r1_to_TL319_aave.201006.nc</map>
+      <map name="OCN2ATM_SMAPNAME">cpl/gridmaps/ECwISC30to60E2r1/map_ECwISC30to60E2r1_to_TL319_aave.201006.nc</map>
```
stephenprice, line 3952: same comment as above w.r.t. differing types of mapping file use for to / from state

components/mpas-seaice/bld/namelist_files/namelist_defaults_mpassi.xml
```diff
@@ -76,11 +77,13 @@
 <config_initial_latitude_north ice_grid="oEC60to30v3wLI">75.0</config_initial_latitude_north>
 <config_initial_latitude_north ice_grid="ECwISC30to60E1r2">75.0</config_initial_latitude_north>
 <config_initial_latitude_north ice_grid="SOwISC12to60E2r4">85.0</config_initial_latitude_north>
+<config_initial_latitude_north ice_grid="ECwISC30to60E2r1">85.0</config_initial_latitude_north>
```
stephenprice, line 80: Should this be 75.0 to be consistent w/ ECwISC30to60E1r2?

components/mpas-seaice/bld/namelist_files/namelist_defaults_mpassi.xml
```diff
@@ -76,11 +77,13 @@
...
 <config_initial_latitude_north ice_grid="oRRS30to10v3wLI">85.0</config_initial_latitude_north>
 <config_initial_latitude_south>-60.0</config_initial_latitude_south>
 <config_initial_latitude_south ice_grid="oEC60to30v3wLI">-75.0</config_initial_latitude_south>
 <config_initial_latitude_south ice_grid="ECwISC30to60E1r2">-75.0</config_initial_latitude_south>
 <config_initial_latitude_south ice_grid="SOwISC12to60E2r4">-85.0</config_initial_latitude_south>
+<config_initial_latitude_south ice_grid="ECwISC30to60E2r1">-85.0</config_initial_latitude_south>
```
stephenprice, line 86: as above, should this be -75.0 to be consistent w/ ECwISC30to60E1r2?

stephenprice (approved):
The only thing I didn't look at too carefully here was the buildnml additions (probably redundant if Xylar already tested, but I can also check to make sure all the necessary files are there).

components/mpas-seaice/bld/namelist_files/namelist_defaults_mpassi.xml
darincomeau: @stephenprice I suppose this should be something we decide on, so thanks for pointing this out. I put this at +/- 85 to be consistent with the SORRM grid for v2 runs, which I was basing on discussions from v1 that the smaller disk was better in some way. I'm open to reverting this back to +/- 75, but I should redo the spun up ICs used for B-cases to be consistent with that.

stephenprice: I'm not sure if it's important enough to redo initial conditions. Maybe the consensus is that the initial lats. for sea ice should be narrower going forward?

darincomeau: @stephenprice thinking more on this, I think we should just stick to +/- 75. I can't recall a consensus reach on moving to +/- 85. Note these default sea ice ICs are only used for G-cases, as the B-cases point to 1 month restarts from a G-case. I just checked the `SOwISC12to60E2r4` restartFromG ICs I generated and those also used +/- 75, so I should make that correction for that grid in a separate PR.

darincomeau:
Per comment from @stephenprice , I'm going to change the default sea ice 'cold start' initial condition bounds from +/- 85 > +/- 75, to be consistent with other grids. I've added the Do Not Merge label until I've generated these files.

darincomeau:
I've updated the default sea ice initial condition latitude bounds to be +/- 75, consistent with other grids. I've regenerated the cold start (but dynamically adjusted) initial conditions and spun up initial conditions from one month G-case restarts using the corrected sea ice initial condition defaults, and removed the previous versions from the inputdata directory.

Removed the Do not merge tag; this PR is ready to resume review.

darincomeau:
@stephenprice , @xylar , I see you've already approved, but since I've changed the initial condition, could I ask someone else to run a test case with this grid and the `CRYO1850` compset?

xylar:
@darincomeau, I'll be happy to rurun my testing tomorrow.  If you need it done today, maybe @stephenprice can give it a go?

darincomeau:
@xylar tomorrow would be fine - thanks!

jonbob:
@xylar - did you have a chance to retest? Or should I?

xylar:
@jonbob, sorry for the delay.  My test just finished and was successful!

xylar:
Results are in:
```
/lcrc/group/e3sm/ac.xylar/scratch/chrys/20210415.CRYO1850.ne30pg2_ECwISC30to60E2r1.chrysalis/
```

jonbob (approved):
Approved based on visual inspection and developer testing

jonbob:
merged to next, running quick sanity tests

jonbob:
test merge passed SMS_P480.T62_ECwISC30to60E2r1.GMPAS-IAF.anvil_intel

jonbob:
merged to master

PR #4213, opened 2021-04-11
add code to compute NH pressure at interfaces

mt5555:
add an interface to the dycore's nonhydrostatic pressure on layer interfaces

based on the analysis here:
https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/2579530246/NH+surface+pressure

we do not change the current approach at the surface (using the mu=1 approximation), but this does switch from a cell average type algorithm to more accurate interpolation for the interior interface values.

[BFB] since this is a diagnostic not currently used

### DIFF
--- a/components/homme/src/theta-l/share/eos.F90
+++ b/components/homme/src/theta-l/share/eos.F90
@@ -209,11 +209,17 @@ implicit none
    
 
    if (present(pnh_i_out)) then
-      ! boundary values already computed.  interior only:
+      ! boundary values already computed. interpolate interior
       do k=2,nlev
-         pnh_i(:,:,k)=(hvcoord%d_etam(k)*pnh(:,:,k)+hvcoord%d_etam(k-1)*pnh(:,:,k-1))/&
-              (hvcoord%d_etai(k)*2)
+         pnh_i(:,:,k)=(hvcoord%d_etam(k-1)*pnh(:,:,k)+hvcoord%d_etam(k)*pnh(:,:,k-1))/&
+              (hvcoord%d_etam(k-1)+hvcoord%d_etam(k))
       enddo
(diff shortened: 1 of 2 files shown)


mt5555:
@bartgol : does this require corresponding changes in HOMMEXX?

bartgol:
This is just to couple to eam, right? pnh_i is never _used_ inside Homme, so there should be no need for changes in Hommexx.

rljacob:
Is the title right?  "compute to compute" ?

mt5555:
replaced algorithm.  Removed use of weights based on eta coordinate (per  @oksanaguba recommendation) and replaced with linear interpolation in the hydrostatic pressure coordinate.  

This has the advantage that if pnh=pi (on midpionts), then the interpolation pnh_i = pi_i.

This is a desirable property, since in the NH model, pnh is often very close to pi, and so pnh_i should be close to pi_i.

components/homme/src/theta-l/share/eos.F90
```diff
@@ -210,16 +210,12 @@ subroutine pnh_and_exner_from_eos2(hvcoord,vtheta_dp,dp3d,dphi,pnh,exner,&
 
    if (present(pnh_i_out)) then
       ! boundary values already computed. interpolate interior
+      ! use linear interpolation in hydrostatic pressure coordinate
+      ! if pnh=pi, then pnh_i will recover pi_i
```
bartgol, line 214: I don't think this statement is true, since linear interpolation is not the inverse of midpoint averaging.

mt5555: it's true because dp3d=delta pi.

bartgol (approved):

PR #4216, opened 2021-04-12
Allow snow to occlude LAI in FATES

ckoven:
These changes allow FATES to consider snow occlusion of LAI. This accompanies https://github.com/NGEET/fates/pull/734 and fixes https://github.com/NGEET/fates/issues/359.

Should be non-bit-for-bit only in FATES simulations.

### DIFF
--- a/components/elm/src/main/elmfates_interfaceMod.F90
+++ b/components/elm/src/main/elmfates_interfaceMod.F90
@@ -135,6 +135,7 @@ module ELMFatesInterfaceMod
    use EDBtranMod            , only : btran_ed, &
                                       get_active_suction_layers
    use EDCanopyStructureMod  , only : canopy_summarization, update_hlm_dynamics
+   use EDCanopyStructureMod  , only : UpdateFatesAvgSnowDepth
    use FatesPlantRespPhotosynthMod, only : FatesPlantRespPhotosynthDrive
    use EDAccumulateFluxesMod , only : AccumulateFluxes_ED
    use FatesSoilBGCFluxMod   , only : UnPackNutrientAquisitionBCs
@@ -821,7 +822,7 @@ contains
       call this%wrap_update_hlmfates_dyn(nc,               &
                                          bounds_clump,     &
                                          canopystate_inst, &
-                                         frictionvel_inst)
+                                         frictionvel_inst, .false.)
       
       ! ---------------------------------------------------------------------------------
       ! Part IV: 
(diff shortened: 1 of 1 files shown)


peterdschwartz:
@ckoven I just wanted to clarify how the submodule update will be handled.  Is the plan to update this branch once  NGEET/fates#734  is resolved or will there be other intermediate E3SM PRs ? Thanks

ckoven:
@peterdschwartz I should defer to @rgknox on how he thinks this should best be staged.

rgknox:
The fates side PR has been integrated into fates.  I recommend we update the fates pointer to that tag and proceed. We don't have any fates-e3sm PRs that are ahead of this one  in precedence.

rgknox:
@ckoven could you update the fates pointer to: https://github.com/NGEET/fates/releases/tag/sci.1.45.1_api.15.0.0
I realize we discussed pointing to the tag you've used, but tag 1.45.1 is only one tag forward, and contains a useful bug-fix for fates-hydro users.

ckoven:
@rgknox ok I updated to sci.1.45.1_api.15.0.0.

rgknox:
Thanks for pointing to the more recent tag @ckoven 

@amametjanov : this new tag has the IBM compiler fix you submitted

rgknox (approved):

peterdschwartz:
It has been merged to next.

peterdschwartz:
Looks like there's FAILs for the compare_base_rest phases for just the FATES test.  Do you think this will be a quick fix or should I revert? 
Here's a description of the testing I did on cori-knl for this in case it helps debugging. 
After merging to next (no conflicts), the compare_base_rest did fail on the first run.
Then I ran the test for only this branch and next(prior to merge) which both passed.
I then re-merged this and the test ended up passing as well so I thought it was safe to push.

I had thought the initial failure may have been due to not syncing the submodules correctly after merge.  I'm not sure why it would PASS, but from my testing you may have to try debugging with the merged branch to reproduce the failure.

ckoven:
Hi @peterdschwartz I'll try to figure this out once Cori is back online.  Could you let me know which exact test is failing? Also, just to confirm, you have it passing at commit 3ba2c07 but failing at commit 33031c6? Thanks-

peterdschwartz:
@ckoven The test is ERS_Ld20.f45_f45.IELMFATES (example from CDash: https://my.cdash.org/viewTest.php?onlyfailed&buildid=2014082 ).   And yes, only the merge commit gave that failure for me, even though there were no conflicts at the time.

peterdschwartz:
Has there been any luck reproducing/investigating the error?  If not, I may need to revert the merge commit to next today so no more nonBFB merges don't make it through to master.

ckoven:
@peterdschwartz I haven't been able to reproduce and sort this out yet due to computer issues.  So feel free to revert for now.

rljacob:
status:  waiting on @ckoven

ckoven:
updated this PR to include @rgknox's fix to re-enable b4b restarts.

peterdschwartz:
Looks like the merge just causes an expected diff in the fates test, so I will merge to master and request a bless for these tests.

PR #4220, opened 2021-04-14
Adjust PE layouts for RRMv2 configuration on chrysalis (use 64x1)

ndkeen:
Adjust PE layouts for RRMv2 configuration on chrysalis (use 64x1)
```
nodes sypd label
 16   2.4   T
 40   5.7   XS
 60   8.0   S
 80  10.5   M
100  12.3   L
```
[non-BFB] - previous PE layouts used threads, these are pureMPI

### DIFF
--- a/cime_config/allactive/config_pesall.xml
+++ b/cime_config/allactive/config_pesall.xml
@@ -10098,90 +10098,119 @@
           <rootpe_cpl>0</rootpe_cpl>
         </rootpe>
       </pes>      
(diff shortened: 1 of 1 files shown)


ndkeen:
As we will very soon have openmpi on chrysalis which is faster, I'm rerunning some tests. Also, Az suggested I try to prefer layouts that are multiples of 20 to better take advantage of the network configuration of the machine (20 nodes per switch).

PR #4224, opened 2021-04-15
Reduce default output when run in prescribed/prognostic mode

jonbob:
Reduces the number of analysis members mpas-seaice calculates and outputs when run in prognostic mode, for performance reasons. The only output necessary in this situation are the mpassi restart files, since atm developers are unlikely to look at seaice output.

[BFB]

### DIFF
--- a/components/mpas-seaice/bld/build-namelist
+++ b/components/mpas-seaice/bld/build-namelist
@@ -190,6 +190,7 @@ print $fh  <<"EOF";
 <?xml version="1.0"?>
 <config_definition>
 <entry id="ice_grid" value="$ICE_GRID">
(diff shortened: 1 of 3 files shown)


worleyph (approved):
Tested changes using 

     -compset FC5AV1C-L-MPASSI -res ne4_oQU480 

and it worked as expected (eliminating all mpassi .nc files except for a single mpassi restart file). After disabling restart writes in env_run.xml, this case then generated no .nc or .bin files.

Scan of code changes look reasonable to me as well, but I am not familiar with this code.

akturner (approved):

akturner:
These changes look fine to me

jonbob:
merged to next

jonbob:
merged to master

PR #4225, opened 2021-04-16
Introduce our own copies of coupler and data,stub,x models

rljacob:
Add our own copies of the coupler and the data, stub x models from CIME.

Include all relevant git commit history from those directories (constructed using filter-repo).

Modify cime/config/e3sm/config_files to use E3SM versions in all cases.  Ignore versions in cime/src.

Modify buildlib_cmake pieces in each cime_config dir as needed.

Code taken from cime master 599e0bdd5e78a9c0

[BFB]


rljacob:
@jonbob can you add your changes from add-h2otemp to the coupler in this branch?  In its new location of E3SM/driver-mct

jonbob:
@rljacob -- doing that right now

rljacob:
To move the mct driver and the data, stub, x components, I used filter-repo and put the filtered content in separate repos which were then "git pulled" in with --allow-separate-histories

See https://github.com/E3SM-Project/cimemctcomps  and https://github.com/E3SM-Project/cimemctdriver

rljacob:
Verified to be BFB by-hand with an F-case.

jgfouca:
@rljacob , it looks like the CIME submodule update is bringing in more than just the changes needed to support our mct move. Did this PR also do a full CIME update?

rljacob:
Yes it does.   Let me add that to the description.

rljacob:
Updated.  Its a lot because it includes @fischer-ncar 's mods in https://github.com/ESMCI/cime/pull/3917

rljacob:
I'll wait for https://github.com/ESMCI/cime/pull/3930 to be merged and then the submodule here can point back to CIME master.

jgfouca:
@rljacob , ok, I went ahead and merged that CIME PR since everyone had approved it.

jgfouca (approved):

rljacob:
cime submodule now pointing to cime master.

jonbob (approved):
Successfully ran A, G, and X-case tests with this branch

jonbob:
@rljacob - any idea when this will get merged? I think WC has some cpl changes they will need to make soon, and it would be much easier if we did that after this change

rljacob:
I'll start merging it tomorrow.

rljacob:
merged to next

PR #4228, opened 2021-04-18
Fixes a bug in rad diags calculations due to prescribed volcanic emissions (maint-1.0)

singhbalwinder:
Radiation diagnostic calculations were producing different
answers even for the exact same list of aerosols mentioned in
the namelist in rad_diag_1 and rad_diag_2 fields. This bug
appears only for the compsets which turn on the prescribed
volcanic emissions (e.g. F2010). The bug doesn't affect other
compsets (e.g. FC5AV1C-L).
The reason was the use of unit conversion factor (km to m)
which got applied repeatedly for each diagnostic call. This bug
does not affect any simulation which doesn't use repeated
radiation diagnostic calls via rad_diag_1 etc. in the namelist

[BFB] - Bit-For-Bit


kaizhangpnl (approved):

wlin7:
Merged to maint-1.0.

Issue #4073, opened 2021-02-04
Oscillations in lowest-level wind speeds over some land regions


quantheory:
Over certain areas (e.g. the Amazon), E3SMv1 produces wild oscillations in the surface winds, causing the winds to largely "turn off" or "blink" every other time step. This appears to be a numerical instability caused by only doing land-atmosphere coupling every half hour in the default model configuration (at least at ne30np4 resolution). The instability seems to result from a tug-of-war between the surface stresses (which can rapidly reduce the surface wind speed) and dynamics (which are pushing to increase or maintain the wind speed in the boundary layer).

This looks quite bad when observing the winds over short time spans, because the boundary layer winds are dominated by high frequency noise. However, it's unclear how much this issue affects the model climate. The rapid pulsing of the winds does affect other surface fluxes, e.g. by increasing evaporation over the affected areas.

This issue is a regression from CAM5, which addressed this kind of instability using the implicit surface stress ("ISS") option, described on page 106/107 of [the CAM5 model description](https://www.cesm.ucar.edu/models/cesm1.0/cam/docs/description/cam5_desc_save.pdf). This option was only implemented for the older CAM turbulence schemes, however, and not for CLUBB (which I believe uses its own implicit solver for propagating the effects of the surface stresses upwards).

There is another way to get rid of these oscillations, which is to set `sfc_flux_avg = 1` in the atmosphere model namelist. This smooths out surface fluxes (latent and sensible heat, as well as the surface stresses) by smearing them out over a ~2 hour timescale, which is similar to how CAM5's ISS scheme handled the surface stresses.

While this gets rid of the instability, smearing the effect of incoming fluxes over a few hours may not be an ideal solution. There may be some other way to improve the land-atmosphere coupling that avoids adding this kind of explicit smoothing.

whannah1:
@quantheory I've actually been thinking about this issue a lot in the last few days. In the case of E3SM with CLUBB there's a "delay" in when the surface friction is applied, because it's not until AFTER the dynamics calculations that CLUBB applies the friction and diffuses the momentum. So in that approach there's no guarantee that the friction reflects the current state wind when it's applied in tphysbc. I'm not sure how much of these oscillations we can attribute to that, but I suspect it plays a role.

quantheory:
[...]

PR #4229, opened 2021-04-19
Add options to affect surface stress and gustiness calculations.

quantheory:
This PR adds two options to env_run.xml that affect the calculation of surface stresses. Both options are off by default, and this PR should be a BFB change unless these options are activated.

The first of these options is ATM_SUPPLIES_GUSTINESS, which causes the atmosphere to export a new field `ugust` to the coupler (and relevant surface components). This extra gustiness is then used in the calculation of the stress by the surface components.

The second option added in this PR is `ATM_FLUX_INTEGRATION_METHOD`. The default is `explicit`, which uses the existing stress calculations. If this option is set to `implicit_stress`, an implicit method is used to calculate the surface stresses, which reduces the instability due to a coarse time step. This requires two additional fields to be exported from the coupler (`wsresp` and `tau_est`), which represent a linearization of the response of the boundary layer scheme to the surface stresses. This commit provides a partial solution to issue #4073.

### DIFF
--- a/components/eam/bld/namelist_files/namelist_definition.xml
+++ b/components/eam/bld/namelist_files/namelist_definition.xml
@@ -3508,7 +3508,15 @@ is sent to the coupler every time step, using the fields 'wsresp' and 'tau_est'.
 
 <entry id="export_gustiness" type="logical" category="pbl"
        group="phys_ctl_nl" valid_values="" >
-If .true., a gustiness value will be exported to the coupler.
+If .true., a gustiness value will be exported to the coupler, which allows
+surface components to distinguish between the effect of mean wind and gustiness
+when calculating fluxes (and in particular surface drag).
+
+If .false., no extra value is exported, and gustiness is instead applied by the
+atmosphere by scaling up the exported wind fields (ubot and vbot). This is
+simpler and requires slightly less coupling overhead, but produces numerical
+artifacts (oscillations) in surface stresses, particularly when the gustiness is
+large compared with the mean wind.
 </entry>
 
 
--- a/components/eam/bld/namelist_files/namelist_definition.xml
+++ b/components/eam/bld/namelist_files/namelist_definition.xml
@@ -3500,6 +3500,12 @@ Bretherton; 'HBR' for Rasch modified version of 'HB'.
 Default: 'HB'
 </entry>
 
+<entry id="linearize_pbl_winds" type="logical" category="pbl"
+       group="phys_ctl_nl" valid_values="" >
+If .true., a linearization of the response of the pbl scheme to surface stresses
+is sent to the coupler every time step, using the fields 'wsresp' and 'tau_est'.
+</entry>
+
 
 <!-- Flags for implementing code modifications under polar project -->
 
--- a/components/eam/bld/namelist_files/namelist_definition.xml
+++ b/components/eam/bld/namelist_files/namelist_definition.xml
@@ -3506,6 +3506,11 @@ If .true., a linearization of the response of the pbl scheme to surface stresses
 is sent to the coupler every time step, using the fields 'wsresp' and 'tau_est'.
 </entry>
 
+<entry id="export_gustiness" type="logical" category="pbl"
+       group="phys_ctl_nl" valid_values="" >
+If .true., a gustiness value will be exported to the coupler.
+</entry>
+
 
 <!-- Flags for implementing code modifications under polar project -->
 
--- a/components/eam/src/physics/clubb/advance_windm_edsclrm_module.F90
+++ b/components/eam/src/physics/clubb/advance_windm_edsclrm_module.F90
@@ -37,7 +37,8 @@ module advance_windm_edsclrm_module
                rho_ds_zm, invrs_rho_ds_zt, &
                fcor, l_implemented, &
                um, vm, edsclrm, &
-               upwp, vpwp, wpedsclrp )
+               upwp, vpwp, wpedsclrp, &
+               um_pert, vm_pert, upwp_pert, vpwp_pert)
 
     ! Description:
     ! Solves for both mean horizontal wind components, um and vm, and for the
@@ -160,6 +161,13 @@ module advance_windm_edsclrm_module
     real( kind = core_rknd ), dimension(gr%nz,edsclr_dim), intent(inout) ::  &
       wpedsclrp      ! w'edsclr' (momentum levels)           [m/s {units vary}]
 
+    ! Variables used to track perturbed version of winds.
+    real( kind = core_rknd ), intent(inout), dimension(:), pointer ::  & 
+      um_pert,   & ! perturbed <u>    [m/s]
+      vm_pert,   & ! perturbed <v>    [m/s]
+      upwp_pert, & ! perturbed <u'w'> [m^2/s^2]
+      vpwp_pert    ! perturbed <v'w'> [m^2/s^2]
+
     ! Local Variables
     real( kind = core_rknd ), dimension(gr%nz) ::  &
       um_tndcy,    & ! u wind component tendency                    [m/s^2]
@@ -177,10 +185,12 @@ module advance_windm_edsclrm_module
       solution  ! The solution to the tridiagonal matrix            [units vary]
 
     real( kind = core_rknd ), dimension(gr%nz) :: &
-      wind_speed  ! wind speed; sqrt(u^2 + v^2)                     [m/s]
+      wind_speed, &   ! wind speed; sqrt(u^2 + v^2)                 [m/s]
+      wind_speed_pert ! pertubed wind speed; sqrt(u^2 + v^2)        [m/s]
 
     real( kind = core_rknd ) :: &
-      u_star_sqd  ! Surface friction velocity, u_star, squared      [m/s]
+      u_star_sqd, &   ! Surface friction velocity, u_star, squared  [m/s]
+      u_star_sqd_pert ! perturbed u_star, squared                   [m/s]
 
     logical :: &
       l_imp_sfc_momentum_flux  ! Flag for implicit momentum surface fluxes.
@@ -191,7 +201,13 @@ module advance_windm_edsclrm_module
 
     logical :: l_first_clip_ts, l_last_clip_ts ! flags for clip_covar
 
+    ! Whether perturbed winds are being solved.
+    logical :: l_perturbed_wind
+
     !--------------------------- Begin Code ------------------------------------
+    l_perturbed_wind = (.not. l_predict_upwp_vpwp) .and. associated(um_pert) &
+         .and. associated(vm_pert) .and. associated(upwp_pert) .and. &
+         associated(vpwp_pert)
 
     dummy_nu = 0._core_rknd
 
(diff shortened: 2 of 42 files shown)


whannah1:
@quantheory as a formality, the PR description should be something short and descriptive that can be copied and pasted into the commit message created by the integrator. So you should move all the analysis and discussion to a comment instead.

whannah1:
I also noticed that you've included changes to the cime submodule. My understanding is that we should not be commiting submodule changes to an E3SM PR, so you'll need to remove these from the branch.

components/eam/bld/namelist_files/namelist_definition.xml
```diff
@@ -3481,6 +3481,17 @@ Bretherton; 'HBR' for Rasch modified version of 'HB'.
...
+<entry id="linearize_pbl_winds" type="logical" category="pbl"
+       group="phys_ctl_nl" valid_values="" >
+If .true., a linearization of the response of the pbl scheme to surface stresses
+is sent to the coupler every time step, using the fields 'wsresp' and 'tau_est'.
+</entry>
+
+<entry id="export_gustiness" type="logical" category="pbl"
+       group="phys_ctl_nl" valid_values="" >
+If .true., a gustiness value will be exported to the coupler.
+</entry>
```
whannah1, line 3493: Seems like the description here could use more details about what this is for. I'm thinking just something like "helps stabilize near-surface wind oscillations" or something similar will help clarify why the option exists. It's a good idea to be verbose here for the sake of people who look at this in the future and have no idea what the various options are used for.

components/eam/src/physics/clubb/advance_windm_edsclrm_module.F90
```diff
@@ -440,6 +456,182 @@ subroutine advance_windm_edsclrm &
...
+                               rho_ds_zm, invrs_rho_ds_zt,              & ! In
+                               l_imp_sfc_momentum_flux, vpwp_pert(1),   & ! In
+                               rhs(:,windm_edsclrm_vm)                  ) ! Out
+
+
+       ! Store momentum flux (explicit component)
+
+       ! The surface flux, x'w'(1) = x'w'|_sfc, is set elsewhere in the model.
+!      upwp(1) = upwp_sfc
+!      vpwp(1) = vpwp_sfc
```
whannah1, line 496: I hate to see abandoned code be commited in a PR, do we need to keep these lines around?

quantheory:
@whannah1 I'm copying the original message below, and then I'll edit the top of the thread.

In combination with ESMCI/CIME#3929, this PR adds two options to env_run.xml that affect the calculation of surface stresses. Both options are off by default, and this PR should be a BFB change unless these options are activated.

The first of these options is ATM_SUPPLIES_GUSTINESS, which causes the atmosphere to export a new field `ugust` to the coupler (and relevant surface components). This extra gustiness is then used in the calculation of the stress by the surface components.

This change is motivated by a problem with the current implementation that results in (mild) instability in the surface wind calculations. Right now, the gustiness calculated in EAM is applied to the mean wind fields (`u`/`v`) before exporting them to the coupler, by multiplying them in `camsrfexch` by a factor `(vmag_gust + vmag)/vmag`, i.e. by scaling the wind from its original magnitude `vmag` to `vmag + vmag_gust`.

The problem with this multiplier comes when the mean wind is small but the gustiness is high. In that case, the magnitude of the surface stress may be high, but the mean momentum change applied to a grid cell should *not* be high, because the gustiness is producing winds acting from different directions that should largely cancel. Put differently, in the limit where (`u`, `v`) -> 0, the stresses (`wsx`, `wsy`) should go smoothly to 0, even if the gustiness (and hence u*) is larger. The surface stress calculations can only satisfy this condition if they have access to both the mean wind and the gustiness, so the gustiness needs to be exported as a separate field.

To show a concrete example of how the current code can cause problems, consider the following time series from an E3SM F2010 case, which is taken from a column in the Andes:

U_time_fluxdiag
V_time_fluxdiag
U10_time_fluxdiag

The black line in this case shows a simple out-of-the-box run with the default 1800 second (half hour) time step. There is a mild numerical instability in this case, which causes the winds to tend to oscillate up and down with a period of every 2-3 time steps. The grey line shows a run with a 300 second (5 minute) time step. This should result in less instability if this is a simple case of numerical instability due to a coarse time step, but in practice you can see that it does not really help. The fact that U10 is typically much larger than U^2 + V^2 here is evidence of the impact of the gustiness parameterization in this case. We can see that there are large changes in the surface stresses as well:

TAUX_time_fluxdiag
TAUY_time_fluxdiag

With `ATM_SUPPLIES_GUSTINESS=TRUE`, the picture changes a bit:

U_time_fluxdiag
V_time_fluxdiag

Here, the pink line is a run with `ATM_SUPPLIES_GUSTINESS` enabled at 300 second time step, which is much more stable than the default configuration at the same time step (grey line), at least for the instability near the beginning of the run. However, the red line, which has `ATM_SUPPLIES_GUSTINESS` at the default time step, is even more oscillatory than the default code (black line). This is, I believe, due to the numerical instabilities described in #4073, which can occur at coarse time steps regardless of the gustiness parameterization. Because the `ATM_SUPPLIES_GUSTINESS` option reduces the amount of stress when the gustiness is high, the boundary layer wind is larger and produces more instability in this case.

The second option added in this PR is `ATM_FLUX_INTEGRATION_METHOD`. If this option is set to `implicit_stress`, an implicit method is used to calculate the surface stresses, which reduces the instability due to a coarse time step. This has to be a coupler option, because it requires two additional fields to be exported from the coupler (`wsresp` and `tau_est`), which represent a linearization of the response of the boundary layer scheme to the surface stresses. For the green line in the following plots, both `ATM_SUPPLIES_GUSTINESS` and `ATM_FLUX_INTEGRATION_METHOD` have been changed:

U_time_fluxdiag
V_time_fluxdiag

You can see that this combination is more stable than either the default run or the run with `ATM_SUPPLIES_GUSTINESS` alone, though some weird oscillations persist. I've started some convergence tests, which preliminarily suggest that there's less numerical error associated with this combination over land. However, this has not really been evaluated in long runs, and the impact on fluxes over ocean is uncertain.

So I would not advocate for this as a default configuration at this point, but I do think that the option is useful as a benchmark for future solutions to issue #4073. (There have been several other suggested solutions, but most of them involve exporting extra fields from the atmosphere, so from a software engineering perspective a lot of the necessary changes will be similar.)

quantheory:
@whannah1 The CIME changes are required for this PR to function (the namelist scripts will fail otherwise), so this PR actually can't be merged unless/until ESMCI/CIME#3929 is accepted *and* the submodule in E3SM is updated with those changes.

Once that happens, I can remove the cime update at that point, but there's no point in doing it now, since it will just render this branch non-functional.

whannah1:
@quantheory I think merging the submodule changes here is still bad practice that will create conflicts. I know I've had some bad experiences with this myself. Maybe we can just make sure the submodule PR is merged before we merge this PR?
Maybe @jgfouca or @rljacob should chime in here.

quantheory:
@whannah1 That's exactly what I meant. This PR should *not* be merged until the CIME PR is complete.

whannah1:
Right, but either way, you'll need to remove the submodule changes from this branch, right?

quantheory:
Not really? I don't think that submodule updates like this are different from any other potential merge conflict. As long as you are careful to make sure that the right CIME commit is being included when you do the final merge, it doesn't really matter what version was being used on the branch before.

rljacob:
When E3SM changes require simultaneous CIME changes, its ok to update the submodule in the E3SM PR. 

@quantheory in this case, this PR comes at an awkward time because I'm in the middle of splitting off the coupler src code so E3SM has its own copy.   Once #4225 is merged, I'll ask you to rebase this and make the coupler changes in our version.

whannah1:
@rljacob thanks for the clarification.

rljacob:
@quantheory when you say "because the gustiness is producing winds acting from different directions" is that what "gustiness" always means in this paramaterization?   You can get wind gusts in the same direction as the mean wind.

quantheory:
@rljacob Part of the gustiness calculation is derived from CLUBB, and the variables used from CLUBB are just the variance of perturbations in `u` and `v` from the mean (`up2 + vp2`). So in that case, the gustiness parameterization is rather agnostic about the direction of the gusts. When I talk about gusts acting from different directions, I really just mean that the effect of the gusts are by definition not in any particular direction *on average*, because they represent the effects of deviation from the mean wind. (So even if we simplify the model by only having one horizontal dimension, there will be gusts in the direction of the mean wind, and also "lulls" where the wind is less active than the mean wind.)

More to the point for the particular situation I'm talking about, when the mean wind is small, the "direction" of the mean wind is not very well-defined anyway. For instance if `u = 1.e-5 m/s` and `v = 0 m/s`, the wind is blowing a tiny amount to the east, but if `u = -1.e-5 m/s`, the wind is blowing a tiny amount to the west. In both of these two cases, the velocity of the mean wind is basically zero, and the momentum flux from the land, which acts to oppose the mean wind, should also be more or less zero in both cases. But in the current code, if the gustiness is large, then the calculated stress can also be quite significant in these cases, i.e. there's a strong western momentum flux in one case, and a strong eastern one for the other case. I would argue that this discontinuous behavior should be considered a bug, or at least something quite unrealistic in the model.

(In practice this problem is not as bad as I make it sound here because there are various limiters in place for small winds, and because with our default settings CLUBB completely ignores the *direction* of the momentum flux from the surface, instead calculating the *magnitude* of the drag and just assuming that this drag is always acting to slow down the lowest layer winds. But you can still get some weird behaviors from the current gustiness code, as shown above.)

rljacob:
@quantheory you can now rebase this to the head of master and remove the dependence on a cime PR.  Make those change directly in the E3SM/driver-mct and E3SM/share directories and add them to this PR.

quantheory:
@rljacob OK, sounds good. I have a bit of a busy schedule for the next few days, but I should be able to get to it next week.

quantheory:
@rljacob I'm finally getting around to this today. I just pushed some updates so that I could tag the version that I've been using for convergence experiments (for future reference), but I should have a rebase of this branch soon.

components/eam/bld/namelist_files/namelist_definition.xml
quantheory: Got it. I've added something a bit more verbose.

components/eam/src/physics/clubb/advance_windm_edsclrm_module.F90
quantheory: Took them out!

quantheory:
I accidentally squashed some extra commits while rebasing this branch, so I force-pushed fix up the git history a bit, but I think this should be ready to merge once checks pass again? Unless there's anyone else I need to get to review this.

rljacob:
Can you update the description?  I assume this no longer needs a CIME PR.

Also please add tests for each option to extra_coverage.

quantheory:
@rljacob Done and done (at least, I think that the tests I added are what we want). I'm running `e3sm_atm_extra_coverage` on cori right now to make sure I've done this right, will post again when tests pass.

quantheory:
OK, this version passes the new tests.

quantheory:
@rljacob Just poking you about this again, since I will be out for most of next week.

rljacob (approved):

rljacob:
Looks good!  Did you want some other action  from me on this?

rljacob:
Adding @bishtgautam to approve elm changes.

quantheory:
@rljacob Mostly wanted to confirm that things look good to you. Also was not sure which integrator(s) to ask about it (thanks for adding someone for the ELM changes).

rljacob:
@singhbalwinder is the integrator (the "assignee").

singhbalwinder:
I will start integrating as soon as @bishtgautam approves it.

oksanaguba:
Should we wait for climo? I have 2 out of 4 runs so far. Clubb-control run crashed.

components/elm/src/biogeophys/UrbanFluxesMod.F90
```diff
@@ -851,6 +935,19 @@ subroutine UrbanFluxes (bounds, num_nourbanl, filter_nourbanl,
          end if
       end if
 
+      ! Check for convergence of stress.
```
bishtgautam, line 938: Move the `if (implicit_stress` outside the do-loop

components/elm/src/biogeophys/UrbanFluxesMod.F90
```diff
@@ -664,7 +715,36 @@ subroutine UrbanFluxes (bounds, num_nourbanl, filter_nourbanl,
...
+               c = filterc_copy(fc)
+               l = col_pp%landunit(c)
+               if (.not. (abs(tau_diff(l)) < dtaumin)) then
+                  fnc_iter = fnc_iter + 1
+                  filterc_copy(fnc_iter) = c
+               end if
+            end do
+         end if
+
+      end do ITERATION ! end iteration
```
bishtgautam, line 747: Is this code active even when `implicit_stress=.false.`?

components/elm/src/biogeophys/BareGroundFluxesMod.F90
```diff
@@ -266,7 +314,27 @@ subroutine BareGroundFluxes(bounds, num_nolakeurbanp, filter_nolakeurbanp, &
...
+               p = filterp(f)
+               if (.not. (abs(tau_diff(p)) < dtaumin)) then
+                  fn = fn + 1
+                  filterp(fn) = p
+               end if
+            end do
+            if (fn == 0) then
+               exit ITERATION
+            end if
+         end if
```
bishtgautam, line 332: Is this convergence test active when `implicit_stress=.false.`?

bishtgautam (commented):
I have made few minor comments.

Should the values of `implicit_stress` and `atm_gustiness` be the same in EAM and ELM. If yes, is there a check to ensure values that are indeed the same in the two models?

quantheory:
@oksanaguba The changes for the CTRL-CLUBBAC run are to move CLUBB to tphysac, which is completely separate from the changes in this branch. None of the code modifications in this branch should be active in that run.

This tag has bit-for-bit changes only (with default settings), so I don't think a new climo should be necessary, although the comparsion between the CTRL and EQSFC runs will be relevant in the future if we want to make the implicit stress code the default.

components/elm/src/biogeophys/UrbanFluxesMod.F90
quantheory: Yes, but only on the last iteration, and it has no effect. I don't think there's any meaningful cost to doing it, but if you think it would be clearer, I could put an `if (implicit_stress)` block around it.

components/elm/src/biogeophys/BareGroundFluxesMod.F90
quantheory: Same answer as for the `UrbanFluxesMod` equivalent of this block.

quantheory:
@bishtgautam There are no options named `implicit_stress` or `atm_gustiness` in EAM, although there are a pair of options that control exports of the relevant fields (`linearize_pbl_winds` and `export_gustiness`). In theory, I suppose it's OK to change the EAM options without changing the ELM options (e.g. if you wanted to modify stress calculations over ocean/sea ice but not land), but not vice versa. In practice, I'm only allowing for/testing `implicit_stress == linearize_pbl_winds` and `atm_gustiness == export_gustiness`.

Anyway, right now this is enforced in the scripting level by the EAM/ELM buildnml scripts, which set the namelist variables for both models by copying them from the relevant env_run.xml options. If you want I can add runtime checks in the Fortran code as well, but I didn't bother in the first go-around because it seemed redundant (and I was being a bit lazy about this).

bishtgautam (approved):

components/elm/src/biogeophys/UrbanFluxesMod.F90
bishtgautam: Thanks for the explanation. There isn't a need to change things.

bishtgautam:
> Anyway, right now this is enforced in the scripting level by the EAM/ELM buildnml scripts

This is good.

singhbalwinder:
Merged to next

singhbalwinder:
@quantheory : There are some fails on the [dash board](https://my.cdash.org/index.php?project=E3SM). There is a build failure for `SMS_D_Ln5.ne4_ne4.FC5AV1C-L.chrysalis_intel.eam-clubb_sp` test on Chrysalis which needs to be fixed. There are many failures on Compy but we think they are due to PGI optimizations as:
1. We have seen this at least three times in the past where tests (which are failing with PGI) are passing with Intel and changing optimizations to `-O1` or `-O0` on PGI would also make the tests to PASS on PGI.
2. In the current tests, test run with debug options (`_D` in the test name) is passing on Compy PGI. This indicates that test run with `-O0` is passing but tests run with the default optimizations are failing.

To confirm this, you can pick a failing test and add `_D` to the test to see if it passes with `_D`. If it does, it is most likely an optimization issue. Do you think you can take a look at the failing test (clubb_sp test)and fix it today sometime? Let me know if you need any help with that.

quantheory:
@singhbalwinder Yes, I can look at the clubb_sp test in a few minutes.

Is there anything you need me to do regarding the PGI issues? If this is arising from compiler bug or some particular issue with optimizations on a particular version of that compiler, I'm not sure that there's anything I can do about it without access to the particular machine where this is crashing.

singhbalwinder:
Thanks! I can run a simple test to confirm the PGI behavior on Compy (Sorry, I didn't know that you don't have access to Compy)

rljacob:
It likely any pgi compiler will have the same result.

Also the "fails" on compy in this case are mostly diffs with baselines. 
Since the option is off by default, it must be changes that were made to existing code to implement the new stress options.  If you want to take the time to find the culprit, you could.  But we're willing to just bless the baseline diffs for compy/pgi.

quantheory:
@singhbalwinder I just tested the commit I just pushed, and it seems to have fixed the clubb_sp test.

singhbalwinder:
Re-merged to next

oksanaguba:
It seems there is no way to easily tell from logs which options related to stresses/gustiness are on. If there is ever a follow up, it would be good to have options explicitly stated in atm.log .

PR #4231, opened 2021-04-19
Remove check for threading when marine BGC is active

### LINKED ISSUE #3619: Ocean BGC not supporting threading?
Use latest master with any machine that has a config with threads>1.   Create a case with the following.  (I did this on cori).
--compset BGCEXP_BCRC_CNPRDCTC_1850 --res ne30_oECv3

During ./case.setup, you get this error:
```
   Calling /global/u2/r/rjacob/E3SM/components/mpas-ocean/cime_config/buildnml
ERROR:  ERROR mpas-o buildnml: ocn bgc cannot be run with more than 1 thread
```

I thought ocean BGC threading was fixed on master?

jonbob:
Removes a check in the mpas-ocean buildnml script that threw an error if the number of threads was greater than 1 and bgc was being used. Recent PRs have made the marine BGC in mpas-ocean and mpas-seaice thread-safe, so this is no longer necessary.

Fixes #3619 

[BFB]

### DIFF
--- a/components/mpas-ocean/cime_config/buildnml
+++ b/components/mpas-ocean/cime_config/buildnml
@@ -56,13 +56,6 @@ def buildnml(case, caseroot, compname):
     if not os.path.isdir(mpasoconf_dir): os.mkdir(mpasoconf_dir)
 
     #--------------------------------------------------------------------
-    # MPAS-O with active BGC fails bfb restart with more than 1 thread
-    # DO NOT run with ocean bgc turned on with more than 1 thread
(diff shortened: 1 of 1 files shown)


jonbob:
Test merge passes PET_Ld5_P480.ne30_oECv3.BGCEXP_BCRC_CNPRDCTC_1850.chrysalis_intel.elm-bgcexp

rljacob (approved):

jonbob:
merged to master

PR #4233, opened 2021-04-20
update NGGPS benchmark namelists

mt5555:
add NGGPS 3km benchmark namelist used for SC20 paper from Oksana Guba
add NGGPS 3km benchmark namelist used most recently on Summit from Andrew Bradley


mt5555:
adding namelists for reference. no impact on tests. skipping integration testing.

PR #4237, opened 2021-04-21
[WIP] Updates land heat budget output in the coupler output

bishtgautam:
The missing ground heat flux is sent to the coupler to be included
in the heat budget outputted by the coupler.

[BFB]


bishtgautam:
@golaz I have opened this PR for heat budget fixes. Since this branch requires a CIME update, the branch will be rebased ones e3sm updates CIME.

golaz:
@rljacob : we'd like to have this for the next v2beta.

rljacob:
@golaz see discussion in https://github.com/ESMCI/cime/pull/3932.  There is a philosophical problem here in that it has not traditionally been the job of the coupler budget calculation to include terms that aren't flows between components.

rljacob:
In any case, this modification to the coupler, if we do it, will have to happen to our version after #4225 is merged to master.

rljacob:
Discussion from v2 finalization:  budgets are fine and this is not needed.

PR #4239, opened 2021-04-21
Write and read back standard name for all variables in the model

### LINKED ISSUE #4230: Restart run with GNU fails to write restart files due to uninitialized value in "standard_name"
I recently stumbled onto an error that only applies to GNU and is missed by our current ERS test. The error only occurs when a run starts as a restart and tries to write a new set of restart files. This is specific to the restart history file (*eam.rh0*). Currently, our ERS test misses this because it disables writing restart files for the second run. 

To reproduce the problem in an ERS test simply comment out the line that sets REST_OPTION=never in cime/scripts/lib/CIME/SystemTests/ers.py. Alternatively, you can run an SMS test and manually set RESUBMIT=1, so that a second run tries to write restart files. 

@dqwu and @jayeshkrishna were able to trace the problem down to uninitialized values in the "standard_name" variable that was added in PR #4103. So the problem can easily be fixed by initializing this variable at line 1785 in components/eam/src/control/cam_history.F90 like so:
    tape(t)%hlist(f)%field%standard_name = ''

I was further able to verify that PR #4103 is the source of the problem by running a modified ERS test (as described above) for  the commits before and after this PR was merged:
    2f03689881f1de211675bd3ee53f3afa68f2d18e    (Update mpas-source: pgi syntax fix)
    2fbae49c31288775b4f8e69528b390f8cb0d9006    (Merge branch 'rljacob/eam/fix-metadata' (PR #4103))
And the test with the second commit will fail.

dqwu:
Initialize standard name for fields in the history list while
reading data from restart files.

Furthermore, use a special variable (array of strings) in the
restart file to store standard name for all variables in the
model, which can be read back later to create the attributes.

Fixes #4230

### DIFF
--- a/components/eam/src/control/cam_history.F90
+++ b/components/eam/src/control/cam_history.F90
@@ -1780,6 +1780,7 @@ CONTAINS
         tape(t)%hlist(f)%field%decomp_type = decomp(f,t)
         tape(t)%hlist(f)%field%numlev = tmpnumlev(f,t)
         tape(t)%hlist(f)%field%mixing_ratio = ''
+        tape(t)%hlist(f)%field%standard_name = ''
         tape(t)%hlist(f)%hwrt_prec = tmpprec(f,t)
 
         mdimcnt = count(allmdims(:,f,t) > 0)
--- a/components/eam/src/control/cam_history.F90
+++ b/components/eam/src/control/cam_history.F90
@@ -102,7 +102,7 @@ module cam_history
   !
   !   The size of these parameters should match the assignments in restart_vars_setnames and restart_dims_setnames below
   !   
-  integer, parameter :: restartvarcnt              = 37
+  integer, parameter :: restartvarcnt              = 38
   integer, parameter :: restartdimcnt              = 10
   type(rvar_id)      :: restartvars(restartvarcnt)
   type(rdim_id)      :: restartdims(restartdimcnt)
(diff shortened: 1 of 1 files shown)


components/eam/src/control/cam_history.F90
```diff
@@ -1761,6 +1774,8 @@ subroutine read_restart_history (File)
         ierr = pio_get_var(File,zonal_complement_desc, (/f,t/), tape(t)%hlist(f)%field%zonal_complement)
         ierr = pio_get_var(File,avgflag_desc, (/f,t/), tape(t)%hlist(f)%avgflag)
         ierr = pio_get_var(File,longname_desc, (/1,f,t/), tape(t)%hlist(f)%field%long_name)
+        tape(t)%hlist(f)%field%standard_name = ''
```
whannah1, line 1777: seems weird to mix this initialization line in with a block of pio_get_var lines. I think it'd be cleaner to move it down below.

whannah1 (approved):
Looks good to me!

components/eam/src/control/cam_history.F90
jayeshkrishna: I am not sure if it can be moved down, the default inits later on in the code is for attributes that are not saved (so not read back)
@dqwu: Can you remove the above initialization? Currently long_name and other attributes don't have any initialization (if you want to add these inits maybe add them in a block above all the get_vars, but add it for all the field attributes)

jayeshkrishna: Now looking at the code below, sampling_seq field has an initialization just before its read back. So I am ok with the init here as is. I will approve the PR soon.

jayeshkrishna (approved):

rljacob:
merged to next

PR #4242, opened 2021-04-22
Change passing of heat associated with frazil between ocean and ice

jonbob:
Brings in changes in both the mpas-ocean and mpas-seaice drivers to change the field passed between them that represents the heat exchange due to frazil formation. There are related changes in the coupler diagnostics, and the field Fioo_meltp is removed since it is no longer used. Because the coupler fields are changed, tests will show DIFFs even though this is not expected to really change answers.

[non-BFB]


vanroekel (approved):
approved by visual inspection

jonbob:
passed ERS.T62_EC30to60E2r2.GMPAS-IAF.anvil_intel sanity test

jonbob:
merged to next

jonbob:

PR #4250, opened 2021-04-23
Add a bubble test in planar homme

oksanaguba:
Added a planar bubble test, dry or Kessler or RJ physics.
Reorganized init constants of planar homme.
Added 2 bubble tests, dry and Kessler, to nightlies.

[bfb] but new baselines for homme.


oksanaguba:
@celdredsandia would you please look at this? in particular, init of the planar domain/variables that I moved? it is not clean/ready yet. also, the code in namelist mod
```
    if (geometry == "plane") then
      scale_factor = 1.0D0
      scale_factor_inv = 1.0D0
      laplacian_rigid_factor = 0.0D0 !this eliminates the correction to ensure the Laplacian doesn't damp rigid motion

! makes the y-direction cells identical in size to the x-dir cells
! this is important for hyperviscosity, etc.
! Also adjusts Sy so y-dir domain is centered at 0
      if (planar_slice .eqv. .true.) then
        Ly = (Lx / ne_x) * ne_y
        Sy = -Ly/2.0D0
      end if
....
```
probably needs to be moved to a subroutine.

mt5555:
what was wrong with the dcmip subroutine "set_hybrid_coefficients(hv, hybrid, eta_t, c)"?   several of the dcmip test cases use equally spaced levels in z, so it's supposed to work for that case.

mt5555:
documenting based on recent conversation:

Several of the dcmip16 cases, we had a lot of trouble with noise that was fixed by going to a discrete hydrostatic balance instead of the analytical balance usually given in the test case.    Not sure if this is an issue in this case.

In this case, that might be harder because there is a background state in hydrostatic balance + perturbation.  Is the perturbation really NH?  can we use the original phi_from_eos() in test_finalize?  Should we try calling that on the background state and then adding the perturbation?

oksanaguba:
are we talking about this code?
```
  ! set analytic vertical coordinates at t=0
  if(.not. initialized) then
    if (hybrid%masterthread) write(iulog,*) 'initializing dcmip2012 test 1-1: 3d deformational flow'
    call get_evenly_spaced_z(zi,zm, 0.0_rl,ztop)                        ! get evenly spaced z levels
    hvcoord%etai  = exp(-zi/H)                                          ! set eta levels from z
    call set_hybrid_coefficients(hvcoord,hybrid, hvcoord%etai(1),1.0_rl)! set hybrid A and B from eta levels
    call set_layer_locations(hvcoord, .true., hybrid%masterthread)
    initialized = .true.
  endif
```
This is my understanding that this code sets z, then sets eta, A, and B. Considering that usually ps=p0, this code has already set z and p. If we want hydrostatic balance, then T is already defined from above. Is this related to the issue you mentioned?

In the bubble code, I do not believe we want the hydrostatic balance. I decided to not to use the code above not because of that but because it seemed better to init z, then T, then p, and only then A and B from pressure levels. There is an ambiguity in how one would init A and B from pressure, but some monotone B obeying BC constraints seem to work.

components/homme/src/share/control_mod.F90
```diff
@@ -253,6 +253,21 @@ module control_mod
   real (kind=real_kind), public :: dcmip16_mu_q    = -1       ! additional uniform viscosity (scalar tracers); -1 implies it defaults to dcmip16_mu_s value
   real (kind=real_kind), public :: interp_lon0     = 0.0d0
 
+!PLANAR
+  real (kind=real_kind), public :: Lx, Ly !domain size
+  real (kind=real_kind), public :: Sx, Sy !shift
+  real (kind=real_kind), private, parameter :: tol_zero=1e-10 !tolerance to determine if lx,ly,sx,sy are set
```
celdredsandia, line 259: Should these live in control_mod or in physical_constants (where they were before)? I put them in physical constants since that is where rearth and rrearth live, and Lx/Ly are like that for planar. I'm okay with either choice though.

components/homme/src/share/control_mod.F90
```diff
@@ -253,6 +253,21 @@ module control_mod
...
+  real (kind=real_kind), public :: bubble_dT = 0.5         !bubble dTheta
+  real (kind=real_kind), public :: bubble_xycenter = 0.0   !bubble xy position
+  real (kind=real_kind), public :: bubble_zcenter = 2000.0  !bubble z position
+  real (kind=real_kind), public :: bubble_ztop = 10000.0   !bubble z top
+  real (kind=real_kind), public :: bubble_radius = 1500.0  !bubble z top
+  logical,               public :: bubble_cosine  = .TRUE. !bubble uniform or cosine
+
+  public :: set_planar_defaults
+
 contains
```
celdredsandia, line 271: I like making these flexible like this. FYI, these constants currently don't agree with either the small bubble from the literature or the large bubble that me & Matt Norman have been running in the CRM. But with this flexibility it doesn't matter! We might want to change the defaults to match one of those, though.

components/homme/src/share/control_mod.F90
```diff
@@ -617,4 +632,73 @@ subroutine test_timestep_make_parameters_consistent(par, nerr)
...
+       Lx = 51.2D0 * 1000.0D0
+       Ly = 51.2D0 * 1000.0D0
+       Sx = -25.6D0 * 1000.0D0
+       Sy = -25.6D0 * 1000.0D0
+    else if (test_case == "planar_rising_bubble" .OR. test_case == "planar_moist_rising_bubble") then
+       Lx = 1.0D0 * 10000.0D0
+       Ly = 1.0D0 * 10000.0D0
+       Sx = -5000.0D0
+       Sy = -5000.0D0
+! THESE ARE WRONG AND NEED TO BE FIXED WHEN THESE CASES ARE ACTUALLY IMPLEMENTED....
```
celdredsandia, line 682: Same here- maybe adjust these defaults to match one of the existing bubble cases from the CRM?

components/homme/src/share/control_mod.F90
```diff
@@ -617,4 +632,73 @@ subroutine test_timestep_make_parameters_consistent(par, nerr)
          write(iulog,'(a,i2)') 'test_timestep_make_parameters_consistent nerr', nerr
   end subroutine test_timestep_make_parameters_consistent
 
+
+subroutine set_planar_defaults()
+
```
celdredsandia, line 637: I am fine with making this a subroutine, it cleans the namelist code up a tiny bit.

components/homme/src/test_src/dry_planar.F90
```diff
@@ -170,11 +172,141 @@ subroutine planar_rising_bubble_init(elem,hybrid,hvcoord,nets,nete)
...
+      ! get horizontal coordinates at column i,j
+      x  = elem(ie)%spherep(i,j)%lon
+      y  = elem(ie)%spherep(i,j)%lat
+
+      do k=1,nlevp
+
+        rr = sqrt( (x-bubble_xycenter)   *(x-bubble_xycenter)+&
+                   (y-bubble_xycenter)   *(y-bubble_xycenter)+&
+                   (zi(k)-bubble_zcenter)*(zi(k)-bubble_zcenter) )
+
```
celdredsandia, line 228: This needs some special case code for the slice configuration, I think?

components/homme/src/test_src/dry_planar.F90
```diff
@@ -170,11 +172,141 @@ subroutine planar_rising_bubble_init(elem,hybrid,hvcoord,nets,nete)
...
+        !set pot. temperature on interfaces
+        if ( rr < bubble_radius ) then 
+          if (bubble_cosine) then
+            th0(k) = bubble_T0 + bubble_dT * cos(rr*dd_pi / bubble_radius / 2.d0)
+          else
```
celdredsandia, line 233: This cos dependence is different than the one used in the CRM/small bubble tests from the literature. My vote is to change it to be the same.

components/homme/src/test_src/dry_planar.F90
```diff
@@ -170,11 +172,141 @@ subroutine planar_rising_bubble_init(elem,hybrid,hvcoord,nets,nete)
...
+            th0(k) = bubble_T0 + bubble_dT
+          endif
+
+        else
+          th0(k) = bubble_T0
+        endif
+
+        !set interface pressure from const lapse rate, does not depend on x,y, or _dT
+        pi(k) =  p0*( (270.0 - zi(k)*g/cp)/270.0  )**(1.0/kappa)
+      enddo
```
celdredsandia, line 243: Should use bubble_T0 here instead of hard wiring 270. Also, maybe rename bubble_T0 and bubble_dT to bubble_theta0 and bubble_dtheta, since they are really potential temperature?

components/homme/src/test_src/dry_planar.F90
```diff
@@ -170,11 +172,141 @@ subroutine planar_rising_bubble_init(elem,hybrid,hvcoord,nets,nete)
...
+
+      elem(ie)%state%ps_v(i,j,:) = pi(nlevp)
+      elem(ie)%state%v = 0.0
+      elem(ie)%state%w_i= 0.0
+
+      do k=1,nlev
+        !set pottemp, dp, other state vars on midlevels
+        th0m(k)=(th0(k)+th0(k+1))/2.0
+        dpm(k)=(pi(k+1)-pi(k))
```
celdredsandia, line 252: Why not directly set theta and pseudo-density, since there are formulas that exist for them? My concern is this linear interpolation to going to lead to an discrete imbalance between the background theta0 and pseudo-density fields, when really they should be in discrete hydrostatic balance.

mt5555:
two comments above are seperate, I think unrelated issues.

regarding set_hybrid_coefficients:  for test cases that use equally spaced in z levels, there is a very common operation to then set the hybrid coefficiients for this grid. It's such a common operation that HOMME has a subroutine to do this, that is used by a variety of test cases.   This test case is not using that routine - it has its own code duplicating that functionality.  does this test case need something special?

components/homme/test/bubble/nh.nl
```diff
@@ -0,0 +1,68 @@
...
+  statefreq         = 200                       ! number of steps between screen dumps
+  restartfreq       = -1                        ! don't write restart files if < 0
+  runtype           = 0                         ! 0 = new run
+  qsplit            = -1                        ! timesteps set via se_tstep, dt_remap_factor, dt_tracer_factor
+  rsplit            = -1
+  integration       = 'explicit'                ! explicit time integration
+  tstep_type        = 9                         ! 1 => default method
+!  tstep_type        = 5                         ! 1 => default method
+  tstep             = 0.1 !! 20x20x60lev 0.1                       ! dynamics timestep
+  dt_remap_factor   = 1				! remap every 1 time steps
```
celdredsandia, line 29: Can it run at a higher time step?

components/homme/test/bubble/nh.nl
```diff
@@ -0,0 +1,68 @@
...
+  dt_tracer_factor  = 1                         ! tracers run at dynamics time step
+  hypervis_scaling  = 3.0
+  nu                = 0.02 !!!!0.216784
+  nu_top            = 0.0
```
celdredsandia, line 33: It would be nice to find a scaling and nu value that worked generically for all planar tests. I can work on that!

components/homme/src/test_src/dry_planar.F90
```diff
@@ -170,11 +172,141 @@ subroutine planar_rising_bubble_init(elem,hybrid,hvcoord,nets,nete)
...
+  hvcoord%hyai = ai
+  hvcoord%hybi = bi
+
+  !set : hyam hybm 
+  hvcoord%hyam = 0.5_rl *(ai(2:nlev+1) + ai(1:nlev))
+  hvcoord%hybm = 0.5_rl *(bi(2:nlev+1) + bi(1:nlev))
+
+  !  call set_layer_locations: sets  etam, etai, dp0, checks that Am=ai/2+ai/2
+  call set_layer_locations(hvcoord, .true., hybrid%masterthread)
+
```
celdredsandia, line 288: This should probably be subroutine, since setting equally spaced z levels for a given p/T profile is going to show up in a lot of the planar test cases.

components/homme/src/share/control_mod.F90
oksanaguba: will do, thanks

components/homme/src/test_src/dry_planar.F90
oksanaguba: I do not understand how the slice works, will ask you later, thanks.

components/homme/test/bubble/nh.nl
oksanaguba: I just now tried the default EAM value of 3.4e-8  and the bubble is rather noisy. if you have a suggestion I will appreciate that.

celdredsandia: How does the value used for the gravity wave tests (0.216784) work?

components/homme/src/test_src/dry_planar.F90
oksanaguba: For the hydrostatic balance, the code was updated to compute new zi from given pressure/dp and const theta T0 for the reference profile though I do not see any differences in the dry bubble solutions. The code you point to is a perturbation which should not be in the hydro balance. Moreover, since it is a perturbation, we can average and it will only affect the shape, nothing else. To do it right, I think I would have to follow discrete averaging rules for homme. For example, for phi on midlevels it is a simple average, but for theta on interfaces it is a weighted average as in Taylor2020 eqn 48. I would have to invert eqn 48, which seems not needed here.

mt5555:
This test case will need to be updated to use HOMME's initialization code.  when we switched to supporting multiple models (preqx, theta-l), we changed the test cases so that they do not access the elem() struct, but instead call various interface functions provided in each model's element_ops.F90.

There are two interfaces, set_state() which works pointwise, and set_elem_state() which works element wise.

oksanaguba:
Thanks for pointing to set_state* functions, will do everything you guys suggested after I make a functional moist bubble setup.

oksanaguba:
Going back to the question why we cannot use sigma coordinate here:
This is the code I used to init Ai, Bi.
```
  ai(:) = 0.0; bi(:) = 0.0
  ai(1) = pi(1)/p0; 

  do k=2,nlevp
    bi(k) = pi(k)/pi(nlevp)
  enddo
```
With it for 60 levels B varies from 1 to 0.84184687 with it at the top [ 0.000000000000000E+000  0.841846871692207       0.844367418433873  ...]. I plotted PS field from one of successful dry bubble runs (nh-nice-bubble.nl):
Screen Shot 2021-06-20 at 1 22 08 PM
It seems PS can be 99996 Pa, which makes the one before last top interface pressure value to be
0.841846871*99996 = 84181.3197
while Ai(1)=0.83933170393 and the top pressure is always 83933 < 84181 . 

Maybe I am supposed to push the top pressure higher or even to zero? I do not see the need though, we have a few nice bubble runs with the current setup.

oksanaguba:
About set_state and set_elem_state -- they are using temperature, not theta, to init, which in case of bubble test would require exponent and in general does not seem to be neat. Is it necessary to use them here?

celdredsandia:
> About set_state and set_elem_state -- they are using temperature, not theta, to init, which in case of bubble test would require exponent and in general does not seem to be neat. Is it necessary to use them here?

I think they work this way to allow both preqx and theta to use the same initialization routines. This probably doesn't matter for the moist bubble, since it doesn't make sense to run in hydrostatic mode. However, I am still a fan of using some sort of generalized routine like set_state or set_elem_state, to divorce the specification of the test case from the dynamical core's choice of thermodynamic/predicted variables.

oksanaguba:
@celdredsandia -- two of your comments above are:
1) use cos function for the bubble as in literature. I used cos(R*pi/2) as in this paper https://agupubs.onlinelibrary.wiley.com/doi/epdf/10.1002/2017MS001078 and also Klemp2015.

2) set some defaults, like Lx, Ly, in the bubble test to match your other code. Could I please ask you to give me defaults (because I was confused about small/large bubbles in your code)?

oksanaguba:
@celdredsandia about set_state: the bubble is run in hydro code, but it is only set for theta-l, there is abort otherwise. I can insert the code using set_elem_state, but none of that is a clean approach if I need to convert to and from temperature. besides, set_elem_state does not set tracers, so, it does not hide tracers setup. what would you recommend?

celdredsandia:
> @celdredsandia -- two of your comments above are:
> 
> 1. use cos function for the bubble as in literature. I used cos(R*pi/2) as in this paper https://agupubs.onlinelibrary.wiley.com/doi/epdf/10.1002/2017MS001078 and also Klemp2015.
> 2. set some defaults, like Lx, Ly, in the bubble test to match your other code. Could I please ask you to give me defaults (because I was confused about small/large bubbles in your code)?

The large bubble uses Lx=Ly=Lz = 20km, and a bubble radius of 2km, with the middle at x=10km, z=3km. Maybe match that (which should also support running in hydrostatic mode)?

oksanaguba:
I prob should also include bubble tests in this PR...

oksanaguba:
Verification runs for the previous states of this branch are here https://acme-climate.atlassian.net/wiki/spaces/COM/pages/2898591748/State+of+bubble+branch+June+2021

celdredsandia:
> @celdredsandia about set_state: the bubble is run in hydro code, but it is only set for theta-l, there is abort otherwise. I can insert the code using set_elem_state, but none of that is a clean approach if I need to convert to and from temperature. besides, set_elem_state does not set tracers, so, it does not hide tracers setup. what would you recommend?

I would still use set_elem_state, to match what is done with the other moist tests (ie DCMIP2016), and then set tracers the same way as in those tests. Then if our equation of state or choice of prognostic/thermodynamic variables ever changes, the tests should "just work".

mt5555:
HOMME has three interfaces for test cases.  Chris is correct in that these were designed so test cases can be used with any dycore in HOMME, and it hides the choice of thermodynamic variable, and avoids #ifdef THETA_MODEL type code in the initialization routines.   The best and newest interface is set_elem_state().  It should be pretty general - I would expect it could work for this bubble test case without any modifications.  

We dont have an interface for setting tracers, since the expectation is that all dycores in HOMME will support the same tracer variables and mixing ratio conventions.  Test cases should only initialize the mxing ratio, and then later on HOMME will initialize Qdp.  Some test cases also initialize Qdp by mistake, but this will be overwritten.

oksanaguba:
When inverting theta for T to call set_elem_state, I have to use theta-l EOS, which is not consistent with the design that the test will work out of box?

celdredsandia:
> When inverting theta to compute T to call set_elem_state, I have to use theta-l EOS, which is not consistent with the design that the test will work out of box?

I think it will depend on exactly how we specify the moist state, right? ie is the constant "theta" used in specifying the initial condition dry potential temperature, moist potential temperature, virtual potential temperature? This is usually not very clear in the few examples of moist bubbles in the literature (to my knowledge), so I think as long as we make a (well-documented) choice and stick with it things should be fine.

oksanaguba:
@celdredsandia @mt5555 would you please look and see whether bubble init routine (now using set_elem_state) is ok?

oksanaguba:
The last commit ran on anvil against corresponding pre-branch commit for homme and atm_integration suites. all is ok.

oksanaguba:
@mt5555 this is ready.

oksanaguba:
not ready -- found part of code that access q=5 tracer but the bubble runs only wiht 3.

oksanaguba:
Running now with debug flags, something inside dcmip16 physics causes FPE. I switched to master and it was the same line:
```
forrtl: error (75): floating point exception
theta-l-nlev30     00000000012CF4DB  Unknown               Unknown  Unknown
libpthread-2.28.s  0000155545B6FB20  Unknown               Unknown  Unknown
theta-l-nlev30     000000000138A227  Unknown               Unknown  Unknown
theta-l-nlev30     0000000000E23BDA  dcmip2016_physics         221  dcmip2016-physics-z.F90
theta-l-nlev30     0000000000DBDA55  dcmip16_wrapper_m         506  dcmip16_wrapper.F90
theta-l-nlev30     0000000001055E7B  test_mod_mp_compu         232  test_mod.F90
theta-l-nlev30     0000000000B33D6E  prim_driver_base_        1102  prim_driver_base.F90
theta-l-nlev30     0000000000F9B3AC  MAIN__                    240  prim_main.F90
```

It seems that lat in that routine is not init-ed. Changing to
```
lat = 0.0
      ! get forced versions of u,v,p,qv,qc,qr. rho is constant
      call DCMIP2016_PHYSICS(test, u_c, v_c, p_c, th_c, qv_c, qc_c, qr_c, rho_c, dt, z_c, zi_c, lat, nlev, &
                             precl(i,j,ie), pbl_type, prec_type)
```
in dcmip16_wrapper let dcmip16_test1 run in master and moist bubble run in this branch with debug flags.

oksanaguba:
Same is true for dcmip16 test2 (TC), but not 3 (supercell), lat is not init-ed.

oksanaguba:
Another round of tests is needed, but the last commit is bfb with the previous one for both moist bubbles and the dry bubble.

oksanaguba:
Last commit ran for homme and hommebdb on anvil, all is bfb except that there are 2 new tests.

this is ready.

components/homme/src/test_src/dcmip16_wrapper.F90
```diff
@@ -544,8 +554,12 @@ subroutine dcmip2016_test1_forcing(elem,hybrid,hvcoord,nets,nete,nt,ntQ,dt,tl)
     elem(ie)%derived%FQ(:,:,:,2) = (rho_dry/rho)*dp*(qc-qc0)/dt
     elem(ie)%derived%FQ(:,:,:,3) = (rho_dry/rho)*dp*(qr-qr0)/dt
 
+
+    if (.not. case_planar_bubble ) then 
     qi=4; elem(ie)%derived%FQ(:,:,:,qi) = dp*ddt_cl
     qi=5; elem(ie)%derived%FQ(:,:,:,qi) = dp*ddt_cl2
+    endif
```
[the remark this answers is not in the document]
mt5555: can you add a variable, something like "use_toy_chemistry"?   that will make it clear what the new code changes are doing.   it could just be local, with dcmip2016_test1 calling this routine with .true., while the bubble calls this routine with .false.

mt5555 (approved):

oksanaguba:
added an internal variable as requested, homme suites are all ok on anvil.

oksanaguba:
in next

oksanaguba:
The new tests were configured for 16 elements, so 24 ranks run scripts fail. I did not notice that because in the log files I saw 
```
test thetanh-moist-bubble-run was run successfully
```
which is not correct, the runs failed due to not enough parallelism.

oksanaguba:
fixed namelists, they ran on anvil, re-merged

rljacob:
homme integration tests are failing.

PR #4260, opened 2021-04-27
Update defaults in MPAS-Ocean streams and namelist for v2

jonbob:
Adds frazilTemperatureTendency to the mpas-ocean timeSeriesStatsMonthlyOutput stream, updates ocean namelist defaults to match current 20210422.v2rc1a.piControl.ne30pg2_EC30to60E2r2.chrysalis run.

[NML]
[non-BFB] for tests with fully-active compsets and a few specific grids in G-cases


jonbob:
At this point, the only namelist default that is changed is config_am_timeseriesstatsmonthly_compute_interval = 'dt'. config_ssh_grad_relax_timescale is handled in a different PR, I think, and we're still undecided about config_GM_constant_kappa? If that's not correct, they can be easily added

rljacob:
Is this really BFB with current master?

jonbob:
@rljacob - at this point it is, since it only adds a new field to an output stream and changes how often an analysis member is calculated. However, the other possible changes from @golaz's run would not be

maltrud (approved):

vanroekel:
@jonbob even if we are unsure on E&G vs constant GM should we set the default B-case value for constant GM to 900 in this PR?

jonbob:
Thanks @vanroekel - I'm happy to add that. Should it be for all grids? Or just the one we're testing?

vanroekel:
I think SORRM is using GM=900 as well and I think all the EC60toXX should use 900 as the base value.  So I'd say make it default for all meshes where GM is enabled.

golaz (approved):
I also agree with @vanroekel's suggestion of changing the default constant GM value to 900.

jonbob:
OK, I've changed the default for config_GM_constant_kappa to 900, and modified the PR to show it is no longer BFB for fully-active configurations. Please note that this is still set to 600 for most grids running in G-cases.

vanroekel (approved):
Thanks @jonbob for the changes

jonbob:
merged to next

jonbob:
merged to master and expected DIFFs/NML DIFFs blessed

PR #4264, opened 2021-04-28
Adding switch to change nlevsno

amschne:
Adds back in the switch to change the number of snow layers
from 5 to 16 when use_extrasnowlayers (firn model) is enabled.


rljacob:
Please don't use PR's to merge from one development branch to another.

amschne:
Sorry about that!

PR #4266, opened 2021-04-30
Fix to allow continued support for microp_scheme != MG2

AaronDonahue:
This fix is a response to an issue we found in SCREAM when doing an
upstream merge.  Because SCREAM uses the P3 microphysics scheme and
not MG2 the case build would fail during the namelist setup.

The reason was that the variable micro_mincdnc had no default defined
when the microphysics scheme was not mg2.  Since this and one other
new namelist variable are only defined in mg2 the solution was to have
build-namelist ignore these variables for non-MG2 cases.

[BFB]


jgfouca (approved):

singhbalwinder:
Merged to next.

PR #4271, opened 2021-05-04
Change TFREEZE_SALTWATER_OPTION default to mushy

### LINKED ISSUE #4162: TFREEZE_SALTWATER_OPTION needs to be “mushy” by default for E3SM
This was reported to me by @proteanplanet. cime/src/drivers/mct/cime_config/config_component_e3sm.xml sets the seawater freezing temperature option to “minus1p8” but that is not consistent with the rest of our model and needs to be switched. We may also want to wrap a line in shr_frz_mod.F90.in, which does not allow a freezing temperature below -2.0:
```
shr_frz_freezetemp = max(shr_frz_freezetemp,-2.0_R8)
```
with logic for cime_model so it does not get applied for E3SM.

jonbob:
Changes the default value of TFREEZE_SALTWATER_OPTION from "minus1p8" to "mushy". This variable is used to determine how shr_frz_freezetemp is calculated, and is referenced by the docn and dice data models. So it should change answers only for configurations using those models, including some A-, C-, D-, and E-cases.

Fixes #4162 

[NML]
[non-BFB] for some A-, C-, and D-cases

### DIFF
--- a/driver-mct/cime_config/config_component_e3sm.xml
+++ b/driver-mct/cime_config/config_component_e3sm.xml
@@ -174,7 +174,7 @@
   <entry id="TFREEZE_SALTWATER_OPTION">
     <type>char</type>
     <valid_values>minus1p8,linear_salt,mushy</valid_values>
-    <default_value>minus1p8</default_value>
+    <default_value>mushy</default_value>
     <group>run_physics</group>
     <file>env_run.xml</file>
     <desc>Freezing point calculation for salt water.</desc>


jonbob:
@proteanplanet - it turns out the offending code in shr_frz_mod.F90.in, which does not allow a freezing temperature below -2.0, is still in CIME and must be worked out with NCAR. I'll start that PR as well....

proteanplanet:
@jonbob Thanks. It's unlikely this will affect our solution, since we need a very salty ocean for it to make a difference.  So we can proceed with E3SM matters without further consideration to that for the moment for the F-Case.  However, as a rule, lines of code the simply overide the physics are just poor modeling, especially where a hard-wired constant is used.   If the code was delivering freezing temperatures lower that -2˚C, then that's something you actually need to know about, rather than just overiding the physics and proceeding without a crash. So let's pursue that in parallel with just the scripting change to make sure for E3SM we are using 'mushy' as the default freezing temp.

proteanplanet (approved):
This is the correct change to make to match the MPAS-SI configuration.

jonbob:
@wlin7 - could you please look at the impact this has on an F-case? That's what we're most concerned with before making this change

wlin7:
Thanks, @jonbob . I am doing the tests now.

wlin7:
@jonbob  I made a mistake in previous comments. Actually, no impact for F-case with MPASSI either.

For mushy config, I used your branch for this PR. For minus1p8, my existing runs used master of April. 16.

FYI: my runs on chrysalis are
       F2010-MPASSI-mushy:  /lcrc/group/e3sm/ac.wlin/E3SM_simulations/20210506.F2010-MPASSI.ne30pg2_r05.chrysalis/run
       F2010-MPASSI-minus1p8: /lcrc/group/e3sm/ac.wlin/E3SM_simulations/20210416.F2010-MPASSI.ne30pg2_r05.chrysalis/tests/M_1x20_nyears/run

You can see the option for tfreeze_option is different, set as default as intended.  It is all good if the change turns out having no impact. But I also wonder if there is a possibility the change somehow does not take effect.

wlin7:
Thanks, @jonbob . For the change to take effect, does it require companion update in mpas-si source codes? I remember, when I checked out the branch jonbob/driver/update-tfreeze-option and did 'git submodule update --init --recursive',  I didn't see any modules need to be updated.

jonbob:
@wlin7 - no, there aren't any corresponding MPAS changes required. This is really just a coupler/data-models setting and I don't believe it's used by any active components.

wlin7:
No problem at all, @jonbob . Thanks for checking the codes and nice to know how this parameter may affect various config. We may say the F-cases tests are also an indirect test. This PR is ready to go then.

jonbob:
merged to next

jonbob:
merged to master and expected DIFFs and NML DIFFs bless requests filed

PR #4194, opened 2021-04-01
threading/zeroing fix in  PhosphorusBiochemMin_balance

rgknox:
These changes are a bug fix.  An array was being zerod outside the scope of the thread, which causes problems.

Fixes #4082

PR #4279, opened 2021-05-07
Updates FATES test to use one thread

bishtgautam:
Explicitly specify the FATES test to use one thread because
on Chrysalis the FATES test is using two threads and reporting
an error in carbon balance.

[BFB]

### DIFF
--- a/components/elm/cime_config/testdefs/testmods_dirs/elm/fates/shell_commands
+++ b/components/elm/cime_config/testdefs/testmods_dirs/elm/fates/shell_commands
@@ -1,2 +1,3 @@
 #!/bin/bash
 if [ `./xmlquery --value MACH` == bebop ]; then ./xmlchange --id LND_PIO_TYPENAME --val netcdf; fi
+./xmlchange NTHRDS=1


bishtgautam:
@rgknox The FATES test is failing in C-Balance. The issue could be similar to #4194. For the time being, I have modified the test so land only uses one thread.

@ndkeen You routinely run tests with a single thread. Would forcing the FATES test to use a single thread be an issue for you?

rljacob:
E3SM code should be able to run threaded.

ndkeen:
Sure that might allow the test to pass, but wouldn't we want it to run with threads?  I would think any internal testing that happens before a PR would also involve threads, right?

rljacob:
How is the FATES test failing on the other platforms?

rljacob:
I see now.  Other fails were caused by #4157

bishtgautam:
@rljacob @ndkeen, I agree that threading in the land model should work and be BFB. I'm not sure if fixing the threading in ELM+FATES is a high priority item for FATES team (cc: @rgknox, @glemieux).


@ndkeen 
> Sure that might allow the test to pass, but wouldn't we want it to run with threads? I would think any internal testing that happens before a PR would also involve threads, right?

Instead of "internal testing", let's make all land tests to use more than one thread.

bishtgautam:
@rgknox Does this PR look good to you?

rgknox (approved):

peterdschwartz:
Merged to master and requested bless for ERS_Ld20.f45_f45.IELMFATES.chrysalis_intel.elm-fates

rgknox:
Thanks all

PR #4284, opened 2021-05-10
Support for SCM functionality for multiple columns

bogensch:
PR changes the name of the ambiguously named flag “iop_mode” with “scm_multcols”, which turns on single column model (SCM) functionality but for multiple columns. Specifically, its purpose is to take the column on the globe intended to be run in SCM mode, but propagates that point (i.e. meaning surface type and location etc.) to multiple columns, as defined by the user. All columns are then forced identically according to the Intensive Observation Period (IOP) file, which is typically required to run in SCM mode. This PR also adds the flexibility for the user to determine the number of columns in the x & y direction (PTS_NX, PTX_NY and scm_nx, scm_ny).  

Background/Motivation for changes: This PR is intended to support running a doubly periodic cloud resolving model with the planar version of HOMME (specifically, the doubly periodic version of SCREAM) but could be potentially useful for other applications.

All e3sm_developer tests pass as expected on cori-knl

[BFB]


jgfouca (approved):

singhbalwinder:
Merged to next

Issue #4286, opened 2021-05-12
MPAS-Seaice model fails with small negative mass error 


akturner:
MPAS_Seaice periodically fails with very small (<1e-22) negative mass errors.

PR #4287, opened 2021-05-12
Update mpas-source: correct small negative mass IR error

akturner:
Update MPAS-Seaice so that small negative masses generated by incremental remapping are zeroed out rather than killing the model. Incremental remapping can legitimately generate very small negative masses (< -epsilon^2) so if this should happen the model should not fail.

Fixes https://github.com/E3SM-Project/E3SM/issues/4286

[BFB] except for very rare cases when small negative masses before would kill the model


proteanplanet (approved):
Approved:  This allows for a slightly negative mass to occurr during advection due to computational limits, which is later "zapped" when sea ice mass is removed that is less than the puny amount in the sea ice model.

vanroekel:
A minor question - should we track how often this is invoked?  Tracking how much negative mass is removed, but maybe it is so small as to not be important.

akturner:
@vanroekel : I think the amount will be miniscule. Before, this error would result in termination of the code, which was a very rare event. Even when negative mass is generated it is < 1e-22 kg so I can't imagine it would ever be a large amount. Most simulations never encounter this issue. I can, however, add a counter if required.

akturner:
@vanroekel: Actually I dont think kg is the units, maybe ks/m2. I can check that.

vanroekel:
@akturner if you think this will stay miniscule I'm fine with the PR as is, your point about being less than 1e-22 is also important, it will take a long time to adding up to mattering!  Thanks for the reply!

jonbob:
passed sanity testing of the merge

jonbob:
merged to master

PR #4294, opened 2021-05-17
Change crm_rad type to allocatable

whannah1:
This change is needed looking forward to changes to accommodate running the MMF with OpenMP CPU threads. The crm_rad type is changed from a collection of pointers to a collection of allocatable variables. 

[BFB]

### DIFF
--- a/components/eam/src/physics/crm/crm_physics.F90
+++ b/components/eam/src/physics/crm/crm_physics.F90
@@ -43,12 +43,12 @@ module crm_physics
    ! Physics buffer indices  
    integer :: ttend_dp_idx    = -1
 
-   integer :: crm_rad_idx_t    = -1
-   integer :: crm_rad_idx_qv   = -1
-   integer :: crm_rad_idx_qc   = -1
-   integer :: crm_rad_idx_qi   = -1
-   integer :: crm_rad_idx_cld  = -1
-   integer :: crm_rad_idx_qrad = -1
+   integer :: crm_t_rad_idx    = -1
+   integer :: crm_qv_rad_idx   = -1
+   integer :: crm_qc_rad_idx   = -1
+   integer :: crm_qi_rad_idx   = -1
+   integer :: crm_cld_rad_idx  = -1
+   integer :: crm_qrad_idx     = -1
 
    integer :: crm_rad_idx_nc   = -1
    integer :: crm_rad_idx_ni   = -1
@@ -176,12 +176,12 @@ subroutine crm_physics_register()
    call pbuf_add_field('CRM_T',        'global',dtype_r8,dims_crm_3D,idx)
 
    ! Radiation
-   call pbuf_add_field('CRM_T_RAD',    'physpkg',dtype_r8,dims_crm_rad,crm_rad_idx_t)
-   call pbuf_add_field('CRM_QV_RAD',   'physpkg',dtype_r8,dims_crm_rad,crm_rad_idx_qv)
-   call pbuf_add_field('CRM_QC_RAD',   'physpkg',dtype_r8,dims_crm_rad,crm_rad_idx_qc)
-   call pbuf_add_field('CRM_QI_RAD',   'physpkg',dtype_r8,dims_crm_rad,crm_rad_idx_qi)
-   call pbuf_add_field('CRM_CLD_RAD',  'physpkg',dtype_r8,dims_crm_rad,crm_rad_idx_cld)
-   call pbuf_add_field('CRM_QRAD',     'global', dtype_r8,dims_crm_rad,crm_rad_idx_qrad)
+   call pbuf_add_field('CRM_T_RAD',    'physpkg',dtype_r8,dims_crm_rad,crm_t_rad_idx)
+   call pbuf_add_field('CRM_QV_RAD',   'physpkg',dtype_r8,dims_crm_rad,crm_qv_rad_idx)
+   call pbuf_add_field('CRM_QC_RAD',   'physpkg',dtype_r8,dims_crm_rad,crm_qc_rad_idx)
+   call pbuf_add_field('CRM_QI_RAD',   'physpkg',dtype_r8,dims_crm_rad,crm_qi_rad_idx)
+   call pbuf_add_field('CRM_CLD_RAD',  'physpkg',dtype_r8,dims_crm_rad,crm_cld_rad_idx)
+   call pbuf_add_field('CRM_QRAD',     'global', dtype_r8,dims_crm_rad,crm_qrad_idx)
 
    call pbuf_add_field('REI',          'physpkg',dtype_r8,dims_gcm_2D,idx)  ! Effective radius (ice)
    call pbuf_add_field('REL',          'physpkg',dtype_r8,dims_gcm_2D,idx)  ! Effective radius (liquid)
@@ -374,12 +374,12 @@ subroutine crm_physics_init(state, pbuf2d, species_class)
    ! Initialize pbuf variables
    if (is_first_step()) then
 
-      call pbuf_set_field(pbuf2d, crm_rad_idx_t,   0._r8)
-      call pbuf_set_field(pbuf2d, crm_rad_idx_qv,  0._r8)
-      call pbuf_set_field(pbuf2d, crm_rad_idx_qc,  0._r8)
-      call pbuf_set_field(pbuf2d, crm_rad_idx_qi,  0._r8)
-      call pbuf_set_field(pbuf2d, crm_rad_idx_cld, 0._r8)
-      call pbuf_set_field(pbuf2d, crm_rad_idx_qrad,0._r8)
+      call pbuf_set_field(pbuf2d, crm_t_rad_idx,  0._r8)
+      call pbuf_set_field(pbuf2d, crm_qv_rad_idx, 0._r8)
+      call pbuf_set_field(pbuf2d, crm_qc_rad_idx, 0._r8)
+      call pbuf_set_field(pbuf2d, crm_qi_rad_idx, 0._r8)
+      call pbuf_set_field(pbuf2d, crm_cld_rad_idx,0._r8)
+      call pbuf_set_field(pbuf2d, crm_qrad_idx,   0._r8)
       if (MMF_microphysics_scheme .eq. 'm2005') then
          call pbuf_set_field(pbuf2d, crm_rad_idx_nc,0._r8)
          call pbuf_set_field(pbuf2d, crm_rad_idx_ni,0._r8)
@@ -591,16 +591,16 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
    integer                     :: idx_vt_t, idx_vt_q  ! variance transport constituent indices
 
    ! pointers for crm_rad data on pbuf
-   real(crm_rknd), pointer :: crm_rad_tmp_qrad(:,:,:,:) ! rad heating
-   real(crm_rknd), pointer :: crm_rad_tmp_temperature(:,:,:,:) ! rad temperature
-   real(crm_rknd), pointer :: crm_rad_tmp_qv (:,:,:,:) ! rad vapor
-   real(crm_rknd), pointer :: crm_rad_tmp_qc (:,:,:,:) ! rad cloud water
-   real(crm_rknd), pointer :: crm_rad_tmp_qi (:,:,:,:) ! rad cloud ice
-   real(crm_rknd), pointer :: crm_rad_tmp_cld(:,:,:,:) ! rad cloud fraction
-   real(crm_rknd), pointer :: crm_rad_tmp_nc(:,:,:,:) ! rad cloud droplet number (#/kg)
-   real(crm_rknd), pointer :: crm_rad_tmp_ni(:,:,:,:) ! rad cloud ice crystal number (#/kg)
-   real(crm_rknd), pointer :: crm_rad_tmp_qs(:,:,:,:) ! rad cloud snow (kg/kg)
-   real(crm_rknd), pointer :: crm_rad_tmp_ns(:,:,:,:) ! rad cloud snow crystal number (#/kg)
+   real(crm_rknd), pointer :: crm_qrad   (:,:,:,:) ! rad heating
+   real(crm_rknd), pointer :: crm_t_rad  (:,:,:,:) ! rad temperature
+   real(crm_rknd), pointer :: crm_qv_rad (:,:,:,:) ! rad vapor
+   real(crm_rknd), pointer :: crm_qc_rad (:,:,:,:) ! rad cloud water
+   real(crm_rknd), pointer :: crm_qi_rad (:,:,:,:) ! rad cloud ice
+   real(crm_rknd), pointer :: crm_cld_rad(:,:,:,:) ! rad cloud fraction
+   real(crm_rknd), pointer :: crm_nc_rad (:,:,:,:) ! rad cloud droplet number (#/kg)
+   real(crm_rknd), pointer :: crm_ni_rad (:,:,:,:) ! rad cloud ice crystal number (#/kg)
+   real(crm_rknd), pointer :: crm_qs_rad (:,:,:,:) ! rad cloud snow (kg/kg)
+   real(crm_rknd), pointer :: crm_ns_rad (:,:,:,:) ! rad cloud snow crystal number (#/kg)
 
    !------------------------------------------------------------------------------------------------
    !------------------------------------------------------------------------------------------------
@@ -688,8 +688,8 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
    !------------------------------------------------------------------------------------------------
    ! Retrieve pbuf fields and constituent indices
    !------------------------------------------------------------------------------------------------
-   call pbuf_get_field(pbuf, pbuf_get_index('CRM_QRAD'),    crm_rad_tmp_qrad)
-   crm_rad%qrad(:,:,:,:) = crm_rad_tmp_qrad(:,:,:,:)
+   call pbuf_get_field(pbuf, pbuf_get_index('CRM_QRAD'),    crm_qrad)
+   crm_rad%qrad(:,:,:,:) = crm_qrad(:,:,:,:)
 
    call pbuf_get_field(pbuf, pbuf_get_index('PREC_DP'),  prec_dp  )
    call pbuf_get_field(pbuf, pbuf_get_index('SNOW_DP'),  snow_dp  )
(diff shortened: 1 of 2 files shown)


components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -164,12 +176,13 @@ subroutine crm_physics_register()
    call pbuf_add_field('CRM_T',        'global',dtype_r8,dims_crm_3D,idx)
 
    ! Radiation
-   call pbuf_add_field('CRM_T_RAD',    'physpkg',dtype_r8,dims_crm_rad,idx)
-   call pbuf_add_field('CRM_QV_RAD',   'physpkg',dtype_r8,dims_crm_rad,idx)
-   call pbuf_add_field('CRM_QC_RAD',   'physpkg',dtype_r8,dims_crm_rad,idx)
-   call pbuf_add_field('CRM_QI_RAD',   'physpkg',dtype_r8,dims_crm_rad,idx)
-   call pbuf_add_field('CRM_CLD_RAD',  'physpkg',dtype_r8,dims_crm_rad,idx)
-   call pbuf_add_field('CRM_QRAD',     'global', dtype_r8,dims_crm_rad,idx)
+   call pbuf_add_field('CRM_T_RAD',    'physpkg',dtype_r8,dims_crm_rad,crm_rad_idx_t)
```
brhillman, line 179: I think it'd be better to name the indices like the variables they reference, in keeping with the precedent set in other places in the code. I.e., `crm_t_rad_idx` instead of `crm_rad_idx_t` to refer to `CRM_T_RAD`.

components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -562,6 +589,19 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
    logical(c_bool)             :: crm_accel_uv
    integer                     :: igstep
    integer                     :: idx_vt_t, idx_vt_q  ! variance transport constituent indices
+
+   ! pointers for crm_rad data on pbuf
+   real(crm_rknd), pointer :: crm_rad_tmp_qrad(:,:,:,:) ! rad heating
```
brhillman, line 594: I'd probably rather see these named like the variables in pbuf that they are going to refer to. I.e., `crm_qrad_tmp` instead of `crm_rad_tmp_qrad`. Or maybe better yet, `crm_qrad` and drop the `_tmp` altogether.

components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -648,19 +688,8 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
...
-   end if
-
-   call pbuf_get_field(pbuf, pbuf_get_index('CRM_QRAD'),    crm_rad%qrad)
-   call pbuf_get_field(pbuf, pbuf_get_index('CRM_T_RAD'),   crm_rad%temperature)
-   call pbuf_get_field(pbuf, pbuf_get_index('CRM_QV_RAD'),  crm_rad%qv)
-   call pbuf_get_field(pbuf, pbuf_get_index('CRM_QC_RAD'),  crm_rad%qc)
-   call pbuf_get_field(pbuf, pbuf_get_index('CRM_QI_RAD'),  crm_rad%qi)
-   call pbuf_get_field(pbuf, pbuf_get_index('CRM_CLD_RAD'), crm_rad%cld)
+   call pbuf_get_field(pbuf, pbuf_get_index('CRM_QRAD'),    crm_rad_tmp_qrad)
+   crm_rad%qrad(:,:,:,:) = crm_rad_tmp_qrad(:,:,:,:)
```
brhillman, line 692: I think you have to be careful about dimension size here. Is `crm_rad%qrad` allocated with size `pcols` or `ncol`? If `ncol`, then this won't work in general.

components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -1184,14 +1224,48 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
...
+      end if
+
+      do i = 1,ncol
+         crm_rad_tmp_qrad       (i,:,:,:) = crm_rad%qrad       (i,:,:,:)
+         crm_rad_tmp_temperature(i,:,:,:) = crm_rad%temperature(i,:,:,:)
+         crm_rad_tmp_qv         (i,:,:,:) = crm_rad%qv         (i,:,:,:)
+         crm_rad_tmp_qc         (i,:,:,:) = crm_rad%qc         (i,:,:,:)
+         crm_rad_tmp_qi         (i,:,:,:) = crm_rad%qi         (i,:,:,:)
+         crm_rad_tmp_cld        (i,:,:,:) = crm_rad%cld        (i,:,:,:)
+         if (MMF_microphysics_scheme .eq. 'm2005') then
```
brhillman, line 1261: I think it's probably better to move the logical out of the loop here and just have two loops. Maybe combine with the other logical block above that gets the fields?

components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -1184,14 +1224,48 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
...
          k = pver-m+1
          do i = 1,ncol
             crm_rad%qrad(i,:,:,m) = crm_rad%qrad(i,:,:,m) * state%pdel(i,k) ! for energy conservation
          end do
       end do
 
+      !---------------------------------------------------------------------------------------------
+      ! put rad data back in pbuf
+      !---------------------------------------------------------------------------------------------
+      call pbuf_get_field(pbuf, crm_rad_idx_t,    crm_rad_tmp_temperature)
```
brhillman, line 1240: Weren't these fields already read from pbuf above? Or was that only if first step?

components/eam/src/physics/crm/crm_rad_module.F90
```diff
@@ -8,65 +8,91 @@ module crm_rad_module
...
-      rad%temperature => null()
-      rad%qv => null()
-      rad%qi => null()
-      rad%cld => null()
-
-      rad%nc => null()
-      rad%ni => null()
-      rad%qs => null()
-      rad%ns => null()
+      deallocate(rad%qrad)
```
brhillman, line 85: It might be wise to add `if (allocated(rad%qrad)) deallocate(rad%qrad)` patterns here for safety, but probably doesn't matter.

components/eam/src/physics/crm/crm_physics.F90
whannah1: yea, I figured you would say that. I wish we had used the "crm_rad" prefix to begin with. I'll change it to match.

components/eam/src/physics/crm/crm_physics.F90
whannah1: I wanted to clearly distinguish between the "temporary" pointers and the actual rad variables. I guess I can drop the "tmp" though.

components/eam/src/physics/crm/crm_physics.F90
whannah1: Eventually the pbuf variables will need to be pcol and the crm_rad variables will need "ncrms" which is summed over all ncol of all chunks for a given MPI task. So we have to be very explicit there when we make that change.
The size used for allocation of the crm_rad type is pcols, so there shouldn't be any problem, but I can also change this to be more explicit.

components/eam/src/physics/crm/crm_physics.F90
whannah1: Why would that be better?

components/eam/src/physics/crm/crm_physics.F90
whannah1: This is another spot that will need to be changed to accommodate pushing the chunk loop down into crm_physics, so I just structured this how it will look when it's inside one of the chunk loops. I'd like to keep it that way even though all these lines will be touched when I change "pbuf" to "pbuf_chunk" and add an indent for the chunk loop.

components/eam/src/physics/crm/crm_rad_module.F90
whannah1: We didn't do that in the other crm type modules, but you're right, I'll add it.

components/eam/src/physics/crm/crm_physics.F90
whannah1: done.

components/eam/src/physics/crm/crm_physics.F90
whannah1: done.

components/eam/src/physics/crm/crm_physics.F90
whannah1: added 1:ncol

components/eam/src/physics/crm/crm_rad_module.F90
whannah1: done.

components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -43,6 +43,18 @@ module crm_physics
...
    integer :: ttend_dp_idx    = -1
 
+   integer :: crm_t_rad_idx    = -1
+   integer :: crm_qv_rad_idx   = -1
+   integer :: crm_qc_rad_idx   = -1
+   integer :: crm_qi_rad_idx   = -1
+   integer :: crm_cld_rad_idx  = -1
+   integer :: crm_qrad_idx     = -1
+
+   integer :: crm_rad_idx_nc   = -1
```
brhillman, line 53: These should be named consistent with the other indices (e.g., `crm_nc_rad_idx`).

components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -185,10 +198,10 @@ subroutine crm_physics_register()
    call pbuf_add_field('CONCLD',       'global', dtype_r8,dims_gcm_3D,idx)  ! convective cloud fraction
 
    if (MMF_microphysics_scheme .eq. 'm2005') then
-      call pbuf_add_field('CRM_NC_RAD','physpkg',dtype_r8,dims_crm_rad,idx)
-      call pbuf_add_field('CRM_NI_RAD','physpkg',dtype_r8,dims_crm_rad,idx)
-      call pbuf_add_field('CRM_QS_RAD','physpkg',dtype_r8,dims_crm_rad,idx)
-      call pbuf_add_field('CRM_NS_RAD','physpkg',dtype_r8,dims_crm_rad,idx)
+      call pbuf_add_field('CRM_NC_RAD','physpkg',dtype_r8,dims_crm_rad,crm_rad_idx_nc)
```
brhillman, line 201: I think these still need to be changed to, e.g., `crm_nc_rad_idx`.

components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -360,6 +373,20 @@ subroutine crm_physics_init(state, pbuf2d, species_class)
...
    if (is_first_step()) then
+
+      call pbuf_set_field(pbuf2d, crm_t_rad_idx,  0._r8)
+      call pbuf_set_field(pbuf2d, crm_qv_rad_idx, 0._r8)
+      call pbuf_set_field(pbuf2d, crm_qc_rad_idx, 0._r8)
+      call pbuf_set_field(pbuf2d, crm_qi_rad_idx, 0._r8)
+      call pbuf_set_field(pbuf2d, crm_cld_rad_idx,0._r8)
+      call pbuf_set_field(pbuf2d, crm_qrad_idx,   0._r8)
+      if (MMF_microphysics_scheme .eq. 'm2005') then
+         call pbuf_set_field(pbuf2d, crm_rad_idx_nc,0._r8)
```
brhillman, line 384: Should be `crm_nc_rad_idx` consistent with above.

components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -758,23 +787,34 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
...
 
+      ! Initialize CRM radiation variables
+      call pbuf_get_field(pbuf, crm_t_rad_idx,  crm_t_rad)
+      call pbuf_get_field(pbuf, crm_qv_rad_idx, crm_qv_rad)
+      call pbuf_get_field(pbuf, crm_qc_rad_idx, crm_qc_rad)
+      call pbuf_get_field(pbuf, crm_qi_rad_idx, crm_qi_rad)
+      call pbuf_get_field(pbuf, crm_cld_rad_idx,crm_cld_rad)
+      call pbuf_get_field(pbuf, crm_qrad_idx,   crm_qrad)
+      if (MMF_microphysics_scheme .eq. 'm2005') then
+         call pbuf_get_field(pbuf, crm_rad_idx_nc,crm_nc_rad)
```
brhillman, line 798: Should be `crm_nc_rad_idx` for consistency.

components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -1184,14 +1224,48 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
...
+      !---------------------------------------------------------------------------------------------
+      call pbuf_get_field(pbuf, crm_t_rad_idx,  crm_t_rad)
+      call pbuf_get_field(pbuf, crm_qv_rad_idx, crm_qv_rad)
+      call pbuf_get_field(pbuf, crm_qc_rad_idx, crm_qc_rad)
+      call pbuf_get_field(pbuf, crm_qi_rad_idx, crm_qi_rad)
+      call pbuf_get_field(pbuf, crm_cld_rad_idx,crm_cld_rad)
+      call pbuf_get_field(pbuf, crm_qrad_idx,   crm_qrad)
+
+      if (MMF_microphysics_scheme .eq. 'm2005') then
+         call pbuf_get_field(pbuf, crm_rad_idx_nc, crm_nc_rad, start=(/1,1,1,1/), kount=(/pcols,crm_nx_rad, crm_ny_rad, crm_nz/))
```
brhillman, line 1248: Should be `crm_nc_rad_idx` for consistency.

components/eam/src/physics/crm/crm_physics.F90
brhillman: Performance, but probably negligible here.

components/eam/src/physics/crm/crm_physics.F90
whannah1: shoot, totally missed those

components/eam/src/physics/crm/crm_rad_module.F90
```diff
@@ -8,65 +8,105 @@ module crm_rad_module
...
+      call phys_getopts(MMF_microphysics_scheme_out = MMF_microphysics_scheme)
+
+      if (allocated(rad%qrad))        deallocate(rad%qrad)
+      if (allocated(rad%temperature)) deallocate(rad%temperature)
+      if (allocated(rad%qv))          deallocate(rad%qv)
+      if (allocated(rad%qc))          deallocate(rad%qc)
+      if (allocated(rad%qi))          deallocate(rad%qi)
+      if (allocated(rad%cld))         deallocate(rad%cld)
+
+      if (MMF_microphysics_scheme .eq. 'm2005') then
```
brhillman, line 104: Just a note, I don't think this needs to be changed, but with the addition of the `if (allocated())` patterns, we don't actually need the query and logical about the microphysics scheme.

components/eam/src/physics/crm/crm_physics.F90
whannah1: fixed.

components/eam/src/physics/crm/crm_physics.F90
whannah1: fixed

components/eam/src/physics/crm/crm_physics.F90
whannah1: fixed

components/eam/src/physics/crm/crm_physics.F90
whannah1: fixed

components/eam/src/physics/crm/crm_physics.F90
whannah1: fixed

components/eam/src/physics/crm/crm_rad_module.F90
whannah1: yea, I figured we might as well skip the check of whether of not stuff is allocated, but I have no idea if there's any performance implication.

components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -648,19 +688,8 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
...
-      call pbuf_get_field(pbuf, pbuf_get_index('CRM_NS_RAD'), crm_rad%ns, start=(/1,1,1,1/), kount=(/pcols,crm_nx_rad, crm_ny_rad, crm_nz/))
-   end if
-
-   call pbuf_get_field(pbuf, pbuf_get_index('CRM_QRAD'),    crm_rad%qrad)
-   call pbuf_get_field(pbuf, pbuf_get_index('CRM_T_RAD'),   crm_rad%temperature)
-   call pbuf_get_field(pbuf, pbuf_get_index('CRM_QV_RAD'),  crm_rad%qv)
-   call pbuf_get_field(pbuf, pbuf_get_index('CRM_QC_RAD'),  crm_rad%qc)
-   call pbuf_get_field(pbuf, pbuf_get_index('CRM_QI_RAD'),  crm_rad%qi)
-   call pbuf_get_field(pbuf, pbuf_get_index('CRM_CLD_RAD'), crm_rad%cld)
+   call pbuf_get_field(pbuf, pbuf_get_index('CRM_QRAD'),    crm_qrad)
```
brhillman, line 691: You should use `crm_qrad_idx` here instead of querying again, but not a big deal.

components/eam/src/physics/crm/crm_rad_module.F90
brhillman: No problem.

brhillman (approved):
Looks good to me. One tiny comment remaining, you are still querying for the pbuf index for `crm_qrad` in one spot, but now this is a saved index so we can omit this extra query and use the index directly.

components/eam/src/physics/crm/crm_physics.F90
whannah1: good catch, I'll fix this.

whannah1: done.

xyuan (approved):

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #4214, opened 2021-04-12
Transplant MPAS code base to E3SM

mark-petersen:
Move MPAS from a submodule to code within E3SM. All functionality in E3SM is identical to before. Subtree commands were used to keep the commit history for each MPAS component.

MPAS stand-alone may still be compiled in the `components/mpas-framework` directory.

Diagram of revised directory structure
image

See line-by-line instructions for how this PR was created [in the comment below](https://github.com/E3SM-Project/E3SM/pull/4214#issuecomment-818208881)

[Instructions on how to convert PRs from MPAS to E3SM](https://github.com/E3SM-Project/E3SM/pull/4214#issuecomment-832007454)

[BFB]

PR #4295, opened 2021-05-18
Mass and heat conserving frazil coupling

proteanplanet:
This PR ensures the conservation of heat and mass in frazil coupling between MPAS-Ocean and MPAS-SeaIce. The code changes has been tested on Chrysalis in an E3SM B-case simulation, and full documentation of the test out to year 40 is provided at: https://acme-climate.atlassian.net/wiki/spaces/EWCG/pages/2781544502/20210518.v2rc1a.Frazil3.ne30pg2+EC30to60E2r2.chrysalis
A comparison is also provided in a companion test that still includes the limit on frazil liquidus temperature of -0.1˚C: https://acme-climate.atlassian.net/wiki/spaces/EWCG/pages/2773353093/20210516.v2rc1a.Frazil2.ne30pg2+EC30to60E2r2.chrysalis
Relatively accurate conservation is achieved with this PR in the coupled model, indicated at the former link, and there is significantly less heat uptake in the ocean relative to previous E3SM simulations in both test simulations.  The sea ice climate is generally thicker, as indicated in the above links, too.  The -0.1˚C limit was applied to CICE in 2013 as part of efforts to stabilize POP in low salinity environments, but the main change there was included in the associated code, and no justification for an additional  -0.1˚C limit could be found in the documentation.  Removal of the 0.1˚C limit significantly influences year-to-year variations in AMOC for the years provided in the above simualtions.

[non-BFB]

### DIFF
--- a/components/mpas-seaice/driver/ice_comp_mct.F
+++ b/components/mpas-seaice/driver/ice_comp_mct.F
@@ -1684,8 +1684,8 @@ contains
 !    o  swndf    -- sw: nir diffuse downward
 !    o  swvdf    -- sw: vis diffuse downward
 !    o  swnet    -- sw: net
-!    o  q        -- ocn freeze/melt heat flux
-!    o  frazil   -- ocn frazil production
+!    o  q        -- ocn frazil heat flux(+) / melt potential(-)
+!    o  frazil   -- ocn frazil mass flux 
 !    o  bcphidry -- Black Carbon hydrophilic dry deposition flux
 !    o  bcphodry -- Black Carbon hydrophobic dry deposition flux
 !    o  bcphiwet -- Black Carbon hydrophilic wet deposition flux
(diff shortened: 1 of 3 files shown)


components/mpas-seaice/driver/ice_comp_mct.F
```diff
@@ -1759,9 +1759,8 @@ subroutine ice_import_mct(x2i_i, errorCode)!{{{
       i,n
 
    real (kind=RKIND) :: &
-      frazilProduction, &
-      meltingPotential, &
-      freezingPotential
+      frazilMassFlux, &
+      frazilMassFluxRev
```
akturner, line 1763: What does Rev stand for? I would suggest a name describing the variables function

components/mpas-seaice/driver/ice_comp_mct.F
```diff
@@ -1970,13 +1972,25 @@ subroutine ice_import_mct(x2i_i, errorCode)!{{{
...
+           ! the frazilProduction, which is positive when frazil forms, otherwise zero.
+           ! frazilProduction is used to balance the mass budget according to what 
+           ! the sea ice model needs to complete the conversion of frazil to sea ice for
+           ! the given quantity of energy the ocean model has passed to the sea ice model.
+
+           call frazil_mass(freezingMeltingPotential(i), frazilMassFluxRev, seaSurfaceSalinity(i), &
                                    config_thermodynamics_type)
-           freezingMeltingPotential(i) = freezingPotential + min(meltingPotential,0.0_RKIND)
+
+           frazilMassAdjust(i) = frazilMassFlux-frazilMassFluxRev
```
akturner, line 1993: Could you explain this in more detail?

mark-petersen:
@proteanplanet, this will be the first PR after the "MPAS transplant". Since it is the first one, I'll run the commands myself after #4214 is merged to master tomorrow morning (hopefully). That will bring over the (small) changes from MPAS-Dev/MPAS-Model#874, so the code will show here altogether.

mark-petersen:
@proteanplanet, this is now ready for testing. It was simple to bring the changes previously in MPAS-Dev/MPAS-Model#874, now in commit e85c39c65c2e0b36ed762dfe6d01de38d703e034.

components/mpas-seaice/driver/ice_comp_mct.F
proteanplanet: Please see the revised documentation.  I set this up as a draft, and it was not ready for review at the time you looked at this.  However, I believe the in-code documentation answers your questions.

components/mpas-seaice/driver/ice_comp_mct.F
proteanplanet: Please see the revised documentation.  I set this up as a draft, and it was not ready for review at the time you looked at this.  However, I believe the in-code documentation answers your questions.

eclare108213 (commented):
How is update_ocn_f set in this configuration?  It adjusts the water and salt fluxes passed back to the ocean depending on the thermodynamic scheme.  At the moment the choices are limited, just on/off -- it probably ought to distinguish between constant- and non-constant-volume ocean models.  It would be better to treat this coupling issue consistently (i.e. using a namelist flag) rather than having some cases coded in Icepack and other cases hard-coded in mct.  That said, I know you want to get this fixed and move on.  We can look into making it more consistent when we update Icepack in E3SM.

proteanplanet:
> How is update_ocn_f set in this configuration? It adjusts the water and salt fluxes passed back to the ocean depending on the thermodynamic scheme. At the moment the choices are limited, just on/off -- it probably ought to distinguish between constant- and non-constant-volume ocean models. It would be better to treat this coupling issue consistently (i.e. using a namelist flag) rather than having some cases coded in Icepack and other cases hard-coded in mct. That said, I know you want to get this fixed and move on. We can look into making it more consistent when we update Icepack in E3SM.

update_ocn_f  is true in this configuration, and therefore ktherm ==2 ocean updates do not occur within the first conditional at line 1280 in ice_therm_itd.F90, which I think is what you are referring to.

maltrud (approved):
Approved based on developer testing

golaz:
@rljacob or @jonbob : can this still go on next tonight, or is it too late?

rljacob:
Its not to late.  I think 8pm MT is the cut off.

eclare108213 (approved):

akturner (approved):

jonbob:
merged to next

jonbob:
passes sanity testing

jonbob:
merged to master and expected DIFFs blessed -- except for mappy_developer which did not yet report

jonbob:
expected DIFFs on mappy also blessed

PR #4301, opened 2021-05-20
Update MMF derived types

whannah1:
This updates the derived MMF types. Some were classes with type-bound procedures, and there was a general lack of consistency across types. This goes along with PR #4294 in preparation for making the MMF safe for OpenMP threading on CPUs. 

[BFB]


xyuan (approved):
I am ok with these changes, actually, it is going to be better for OpenMP offload code.

components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -562,6 +562,27 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
    logical(c_bool)             :: crm_accel_uv
    integer                     :: igstep
    integer                     :: idx_vt_t, idx_vt_q  ! variance transport constituent indices
+
+   ! pointers for crm state data on pbuf
+   real(crm_rknd), pointer :: crm_state_u_wind     (:,:,:,:) ! CRM u-wind component
```
brhillman, line 567: Since these are used only to access pbuf fields, I think the naming should follow the pbuf names like we do elsewhere. I.e., `crm_u` instead of `crm_state_u_wind`.

whannah1: will do.

whannah1: done.

brhillman:
Merged to next.

brhillman:
Re-merged to next with OMP fix.

brhillman:
Merged to master.

PR #4302, opened 2021-05-20
Move MPAS standalone make from mpas-framework to mpas-CORE directories

mark-petersen:
Change make process for stand-alone MPAS components from:
```
cd components/mpas-framework
make clean CORE=ocean 
make gfortran CORE=ocean
````
to (for the ocean core)
```
cd components/mpas-ocean
make clean 
make gfortran 
```
where other cores are similar.

This does not touch any E3SM cmake files.

[BFB]

### DIFF
--- a/components/mpas-albany-landice/Makefile
+++ b/components/mpas-albany-landice/Makefile
@@ -1,0 +1,1 @@
+../mpas-framework/Makefile
--- a/components/mpas-albany-landice/src/build_options.mk
+++ b/components/mpas-albany-landice/src/build_options.mk
@@ -3,7 +3,7 @@ ifeq "$(ROOT_DIR)" ""
 endif
 EXE_NAME=landice_model
 NAMELIST_SUFFIX=landice
-FCINCLUDES += -I$(ROOT_DIR)/core_landice/mode_forward -I$(ROOT_DIR)/core_landice/shared -I$(ROOT_DIR)/core_landice/analysis_members
+FCINCLUDES += -I$(ROOT_DIR)/mode_forward -I$(ROOT_DIR)/shared -I$(ROOT_DIR)/analysis_members
 override CPPFLAGS += -DCORE_LANDICE
 
 # ===================================
--- a/components/mpas-framework/Makefile
+++ b/components/mpas-framework/Makefile
@@ -1,5 +1,19 @@
 MODEL_FORMULATION =
 
+#
+# Autodetect CORE from path
+#
+override CORE =
+ifneq ($(findstring components/mpas-ocean,$(PWD)),)
+	override CORE = ocean
+endif
+ifneq ($(findstring components/mpas-seaice,$(PWD)),)
+	override CORE = seaice
+endif
+ifneq ($(findstring components/mpas-albany-landice,$(PWD)),)
+	override CORE = landice
+endif
+
 ifneq "${MPAS_SHELL}" ""
         SHELL = ${MPAS_SHELL}
 endif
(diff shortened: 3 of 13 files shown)


mark-petersen:
Compiles all three cores successfully, without `CORE=ocean` etc. Ran ocean nightly test suite with head of `compass/master`

xylar (approved):
These changes look great to me.  I find it very convenient not to need to set the `CORE` environment or make variable.

I tested both the `nightly` and `sia_integration` test suites from `compass/master` with `E3SM/master` as a baseline.  The only test that didn't pass were the ocean RK4 restart test and the land-ice Enthalpy Benchmark A, both of which are known issues unrelated to this PR.

akturner:
I like these changes. It would be helpful though if the testing_and_setup/seaice directory could be moved to mpas-seaice/testing_and_setup/seaice from mpas-framework/testing_and_setup/seaice.

akturner (changes requested):
Move mpas-framework/testing_and_setup/seaice to mpas-seaice/testing_and_setup/seaice

mark-petersen:
@akturner that is a good point. I moved testing_and_setup/seaice from mpas-framework to mpas-seaice.

@matthewhoffman what do you want to do with `components/mpas-framework/testing_and_setup/compass/landice/`? @xylar did we mean to remove `components/mpas-framework/testing_and_setup/compass` altogether? I think so.

xylar:
>  @xylar did we mean to remove components/mpas-framework/testing_and_setup/compass altogether? I think so.

@mark-petersen, that's up to @matthewhoffman.  My impression is that `landice` hasn't moved their workflow to `compass/legacy` yet.  I intentionally only removed the `ocean` directory.  But I do think it's pretty confusing to keep this in E3SM and now is hopefully the time for `landice` to consider the move to the `compass` repo.

matthewhoffman:
Yes, our plan is to transition to the new version of COMPASS at the same time as the repository transition.  A complicating factor is that we have not ported over the higher order test cases yet, so the new COMPASS does not give us complete test coverage.  That said, if there is a desire to eliminate the old COMPASS files now in this PR, we could use the COMPASS setup in the old MPAS-Dev/MPAS-Model repo (but with an exe from the new E3SM repo) in the interim while we finish the transition (which is what we were planning to do anyway - I hadn't realized that anything from testing_and_setup made the transition to the new repo).

xylar:
@matthewhoffman, why not use `compass/legacy` for that purpose instead of `MPAS-Dev/MPAS-Model`?  The purpose of `compass/legacy` in my view is as the "right" place for our old COMPASS test cases to live until they are ported to `compass/master`.

matthewhoffman:
@xylar , you are right about that - the only reason is the automated testing that Mike Kelleher at ORNL set up is currently using the old COMPASS in MPAS-Dev/MPAS-Model.  I was hoping to limit the number of temporary steps for him to retrofit it for before we are fully operational with the new COMPASS and new MPAS.

xylar:
Okay, in that case I would indeed keep using MPAS-Dev/MPAS-Model because moving E3SM doesn't seem any easier than moving to `compass/legacy`.

matthewhoffman (approved):
Thanks @mark-petersen for working out the kinks in these changes.  I tested with and without Albany and with and without DEBUG mode and everything works as expected.  I also confirmed `make clean` is functional and `make -j 4`.  I didn't test every possible combination, but no issues came up.

mark-petersen:
Rather than overload this PR, I created a new PR with the move and delete of the testing directories, https://github.com/E3SM-Project/E3SM/pull/4310

xylar:
I thinks that's wise.

akturner (approved):

jonbob:
merged to next

jonbob:
merged to master

PR #4304, opened 2021-05-20
Add 1e-15 to fEdge to avoid divide by zero at equator

### LINKED ISSUE #4308: Potential divide by zero in ocean parameterization
This bug creates a divide by zero error, but only if latEdge is very close to zero, so it is dependent on the mesh. The exact jigsaw output varies by compiler and machine, so we did not see this error until now. This bug makes QU240 stand-alone nightly tests fail, but only with debug on.

mark-petersen:
This fixes a divide by zero error in the ocean parameterizations, but it only occurs on an edge if latEdge is very close to zero, so it is dependent on the mesh. This bug makes QU240 stand-alone nightly tests fail, but only with debug on. With this fix it passes.

Fixes #4308
[non-BFB]

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_gm.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_gm.F
@@ -636,7 +636,7 @@ contains
            do iEdge=1,nEdges
               cell1 = cellsOnEdge(1, iEdge)
               cell2 = cellsOnEdge(2, iEdge)
-              Lr = min(cGMphaseSpeed(iEdge) / abs(fEdge(iEdge)),          &
+              Lr = min(cGMphaseSpeed(iEdge) / (1.0E-15_RKIND + abs(fEdge(iEdge))), &
                        sqrt(cGMphaseSpeed(iEdge) / (2.0_RKIND*betaEdge(iEdge))))
 
               do k=minLevelEdgeBot(iEdge)+1,maxLevelEdgeTop(iEdge)


mark-petersen:
This was already approved by @vanroekel at https://github.com/MPAS-Dev/MPAS-Model/pull/873 before the transplant. Adding 1e-15 makes it potentially non-bfb.

mark-petersen:
This turns out to be BFB with previous for standalone nightly suite for both gnu and intel. It is possibly nonBFB on some compilers and machines

mark-petersen:
Passes 
```
```

jonbob (commented):
@mark-petersen - it's not clear from your description if this is a bug fix? If it is, it needs a bug/issue to close. It's tagged as a "bug" which I don't understand

jonbob:
@mark-petersen - can you also look at the PR description? It refers to this as "This bug creates a divide by zero error"...

mark-petersen:
Sorry, I was indecisive. It really is a bug fix, so I made a bug report and altered the description.

jonbob (approved):
Thanks -- that makes it clear

jonbob:
merged to next

jonbob:
@rljacob - I think just because it's adding a small amount to the denominator for all meshes. I'm more curious why there were ocean configurations it didn't impact. @mark-petersen ?

xylar:
@jonbob, my guess would be that this particular parameterization is only on for some configurations and not others.

@rljacob, the divide by zero error we saw was with a different variant of the QU240 mesh, not the one in E3SM (one generated from scratch with newer tools) so there is no reason to think that the E3SM QU240 would be affected by this change any differently than any other mesh that uses the GM parameterization.  All meshes using the parameterization would be affected, even those that didn't happen to have cells near the equator, because the fix is to add a very small positive number everywhere (i.e. it affects all cells, not just those that are in danger of causing a divide by zero).  Does that make sense?

rljacob:
Yes its clear from the code that this will change answers in many cases.  The PR description implies it only changes QU240 cases with debugging on.

jonbob:
Thanks @xylar -- I hadn't paid much attention to where in the code it was. It makes sense that we wouldn't see any impact on the prod test suite which mostly uses the EC30to60E2r2 mesh

xylar:
@rljacob, I agree.  @mark-petersen, would you consider revising the PR description for posterity to specify the effects of these changes on all meshes, not just the bug it's meant to fix?  Could you also add which parameterization it affects and which E3SM configurations that's relevant to?  It probably doesn't matter much in this particular PR since it's merged but it's worthwhile to note for the future that these were missing pieces of information that we should have included.

mark-petersen:
Thanks for the feedback. I altered the description accordingly.

PR #4305, opened 2021-05-21
Improve Chrysalis WC low-res coupled L (Large) PE layout

ambrad:
L PE layout for Chrysalis.                                                                                                                                                                
                                                                                                                                                                                              
The 101-node layout was too slow for the ocean. Increase to 105 nodes.

[BFB]

### DIFF
--- a/cime_config/allactive/config_pesall.xml
+++ b/cime_config/allactive/config_pesall.xml
@@ -8398,7 +8398,7 @@
         </rootpe>
       </pes>
       <pes compset=".*EAM.+ELM.+MPASSI.+MPASO.+MOSART.+" pesize="M">
-        <comment> -compset A_WCYCL* -res ne30pg*EC30to60* on 60 nodes pure-MPI, ~19.4 sypd </comment>
+        <comment> -compset A_WCYCL* -res ne30pg*EC30to60* on 60 nodes pure-MPI, ~24 sypd </comment>
         <ntasks>
           <ntasks_atm>3200</ntasks_atm>
           <ntasks_lnd>640</ntasks_lnd>
@@ -8425,7 +8425,7 @@
         </rootpe>
       </pes>
       <pes compset=".*EAM.+ELM.+MPASSI.+MPASO.+MOSART.+" pesize="M80">
-        <comment> -compset A_WCYCL* -res ne30pg*EC30to60* on 80 nodes pure-MPI, ~21 sypd </comment>
+        <comment> -compset A_WCYCL* -res ne30pg*EC30to60* on 80 nodes pure-MPI, ~30 sypd </comment>
         <ntasks>
           <ntasks_atm>4160</ntasks_atm>
           <ntasks_lnd>320</ntasks_lnd>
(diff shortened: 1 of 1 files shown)


ndkeen:
I'm testing again with new checkout. Note your "commit summary" looks like it included older changes. This is just the last one "L PE layout for Chrysalis"

ambrad:
Yes, there is just one commit. I force-pushed because of submodule issues in my original PR branch when updating to the current main branch. Now the branch is clean and has one commit with +4/-4 line changes.

ndkeen:
Using a script based on the one below and changing it to:
a) use this branch
b) start with initial conditions
c) run for 1 month (with restart)
d) use openmpi (which is not quite default)

I see 38.4 SYPD 
Compare to same script, but using the previous 101-node layout and it's 35.4

Merging to next.


```
https://github.com/E3SM-Project/SimulationScripts/blob/4ecf7b318d40de9eb1d5a90b819f4efb13997fbe/archive/v2/rc/coupled/run.20210423.v2rc1e.piControl.ne30pg2_EC30to60E2r2.chrysalis.sh
```

golaz:
@ambrad and @ndkeen : thanks for adjusting the ocean node count to get that extra performance. I never would have thought we would get to a model close to 40 SYPD.

I'm confused about the compset naming convention in the xml, but that's probably just my CIME ignorance. The compset that we now use is "WCYCL1850". Is it still going to pickup the changes which are marked for "A_WCYCL*" ? Thanks.

ambrad:
@golaz, the A_WCYCL string you point out is in a `<comment>` block. So it's an incorrect comment, but the code itself is fine. The match is on
```
<grid name="a%ne30np4.pg.+_oi%EC30to60E2r2">
<pes compset=".*EAM.+ELM.+MPASSI.+MPASO.+MOSART.+" pesize="L">
```
and of course the machine "chrysalis". When running the WC run script, one sees, e.g.,
```
Pes     specification file is /gpfs/fs1/home/ac.ambradl/e3sm/cime_config/allactive/config_pesall.xml
Machine is chrysalis
Pes setting: grid match    is a%ne30np4.pg.+_oi%EC30to60E2r2
Pes setting: machine match is chrysalis
Pes setting: compset_match is .*EAM.+ELM.+MPASSI.+MPASO.+MOSART.+
Pes setting: pesize match  is L
Pes setting: grid          is a%ne30np4.pg2_l%ne30np4.pg2_oi%EC30to60E2r2_r%r05_g%null_w%null_z%null_m%EC30to60E2r2
Pes setting: compset       is 1850SOI_EAM%CMIP6_ELM%SPBC_MPASSI_MPASO_MOSART_SGLC_SWAV_SIAC_SESP
Pes setting: tasks       is {'NTASKS_ATM': 5400, 'NTASKS_LND': 320, 'NTASKS_ROF': 320, 'NTASKS_ICE': 5120, 'NTASKS_OCN': 1280, 'NTASKS_CPL': 5440}
Pes setting: threads     is {'NTHRDS_ATM': 1, 'NTHRDS_LND': 1, 'NTHRDS_ROF': 1, 'NTHRDS_ICE': 1, 'NTHRDS_OCN': 1, 'NTHRDS_CPL': 1}
Pes setting: rootpe      is {'ROOTPE_ATM': 0, 'ROOTPE_LND': 5120, 'ROOTPE_ROF': 5120, 'ROOTPE_ICE': 0, 'ROOTPE_OCN': 5440, 'ROOTPE_CPL': 0}
Pes setting: pstrid      is {}
Pes other settings: {}
Pes comments:  -compset A_WCYCL* -res ne30pg*EC30to60* on 105 nodes pure-MPI, ~38 sypd
```
This PR is in next already, so I'm tempted to leave the comments as the are, since they are inert w.r.t. runtime behavior. But I can open another PR to fix the comments. What do you suggest as the right comment for the matching strings above?

golaz:
@ambrad : thanks for the explanation. I knew I was probably missing something. I should have looked more clsely. There is no need to do change anything to this PR.

ndkeen:
OK to merge with master? I don't think any tests will even use this layout.

PR #4309, opened 2021-05-25
Make OpenMPI default MPI library on Chrysalis

amametjanov:
Make OpenMPI default MPI library on Chrysalis

[non-BFB] - for MPAS and WC e3sm_integration tests

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -1126,7 +1126,7 @@
     <NODENAME_REGEX>chr.*</NODENAME_REGEX>
     <OS>LINUX</OS>
     <COMPILERS>intel,gnu</COMPILERS>
-    <MPILIBS>impi,openmpi</MPILIBS>
+    <MPILIBS>openmpi,impi</MPILIBS>
     <PROJECT>e3sm</PROJECT>
     <SAVE_TIMING_DIR>/lcrc/group/e3sm/PERF_Chrysalis</SAVE_TIMING_DIR>
     <SAVE_TIMING_DIR_PROJECTS>.*</SAVE_TIMING_DIR_PROJECTS>


amametjanov:
Results of baseline comparisons, IntelMPI (existing baselines) vs. OpenMPI (`./cime/scripts/create_test -c -b master --mpilib openmpi e3sm_integration`), are at https://my.cdash.org/viewTest.php?buildid=2016326 .

worleyph:
Do we know why it is not BFB with IntelMPI?

amametjanov:
My guess is reprosum.

rljacob:
No if it was reprosum a lot more F-cases would diff.

The only diffs are in cases with mpas-ocean AND the oEC30to60 grid   and  cases with mpas-landice.  So either OpenMPI has uncovered a bug in those codes or OpenMPI still has a bug that only those configs hit.

ambrad:
We might ask someone on the ocean side to look at the diffs. There are essentially bare MPI_Allreduce calls (via mpas_dmpar_sum_real*) in some of the analysis members, so these would cause diffs in diagnostic values.

There's also one use of mpas_dmpar_sum_real in a non-analysis routine, but that's for sea surface salinity restoring, which I don't think is used in simulations of interest (or else they wouldn't be PE-invariant).

rljacob:
@jonbob or @xylar see above.

xylar:
I don't really have enough experience with the MPAS framework to be any help here.  @mark-petersen or @matthewhoffman, is this something either of you would be able to help look into?  It does seem problematic the MPAS framework isn't behaving.

philipwjones:
As @ambrad states, the MPAS framework is just a minimal wrapper around the MPI_AllReduce. You wouldn't expect them to be b4b for different PE layouts.  But none of the sums impact the simulation, so it should only be diagnostic sums and the diffs should not grow.

xylar:
@philipwjones, thanks, that's reassuring.

worleyph:
If these allreduces in the analysis routines are the source of the issue, then this should not be OpenMPI-specific. The comparison is between Intel-MPI and OpenMPI, but are there reproducibility tests that are failing as well? Or is the comparison between Intel-MPI and OpenMPI looking at different output than are checked for reproducibility?

philipwjones:
Is this OpenMPI library using the new optimizations that perform part of the reductions in the switch hardware? You could imagine that not being b4b with Intel. You could probably write a quick allreduce unit test and see.

rljacob:
It's not failing a reproducibility test.  There are diffs between OpenMPI and a baseline done with IntelMPI.

worleyph:
> It's not failing a reproducibility test. There are diffs between OpenMPI and a baseline done with IntelMPI.

Was just wondering whether the fields being compared in a reproducibility test are different than the fields being compared in the OpenMPI vs. IntelMPI baseline.

rljacob:
Pretty sure its the same.

Digging in to one of the tests, PEM_Ln9.ne30_oECv3.A_WCYCL1850S, diffs were found in mpassi.hist.am.regionalStatistics in these fields:
```
 RMS totalIceArea                     1.0754E-09            NORMALIZED  5.7007E-17
 RMS totalIceVolume                   2.1004E-12            NORMALIZED  5.1152E-17
 RMS totalSnowVolume                  6.5637E-14            NORMALIZED  3.2799E-17
 RMS totalKineticEnergy               8.0687E-02            NORMALIZED  1.4031E-16
 RMS rmsIceSpeed                      1.4164E-18            NORMALIZED  9.4399E-17
 RMS averageAlbedo                    2.4037E-17            NORMALIZED  1.2615E-16
```
And the coupler history was identical which does suggest the 2 runs are BFB in prognostic variables.

rljacob:
The MALI test did have some diffs in the cpl hist:
```
 RMS g2x_Sg_lithop                    3.0525E-16            NORMALIZED  2.0454E-23
 RMS g2x_Sg_dztbot                    1.7560E-21            NORMALIZED  2.1162E-23
 RMS x2g_Flgl_qice                    5.1456E-21            NORMALIZED  1.4488E-15
```
and in .mali.hist.am
```
 RMS totalIceArea                     1.9531E-03            NORMALIZED  1.4201E-16
 RMS iceThicknessMean                 2.2737E-13            NORMALIZED  1.1909E-16
```

ambrad:
MALI includes a global linear solve with undoubtedly a reduction in the Krylov subspace iteration. For the non-MALI cases, the diffs look right: purely in diagnostics, none in prognostics.

worleyph:
Yes. MALI will be the first component that we will have to give up reproducibility for (based on our current mitigation strategies). OCN will be next, if/when it moves to the implicit solver, just like in the old days :-).

rljacob:
SMS_D_Ld1.ne30_r05_oECv3.A_WCYCL1850.chrysalis_intel again had no diffs in the coupler history but did have a few in mpaso.hist.am.globalStats
```
 RMS totalVolumeChange                2.5600E+02            NORMALIZED  5.1179E-09
 RMS absoluteFreshWaterConservation   2.5600E+02            NORMALIZED  6.8607E-01
 RMS tangentialVelocitySum            7.8125E-03            NORMALIZED  6.2487E-15
 RMS relativeVorticitySum             1.8626E-09            NORMALIZED  4.5455E-02
 RMS vertAleTransportTopSum           1.5259E-05            NORMALIZED  1.5995E-16
 RMS layerThicknessPreviousTimestepSu 2.5600E+02            NORMALIZED  1.9309E-16
 RMS relativeVorticityRms             2.1176E-22            NORMALIZED  2.1997E-16
 RMS kineticEnergyCellRms             8.6736E-19            NORMALIZED  2.0043E-16
 RMS pressureRms                      3.7253E-09            NORMALIZED  1.4177E-16
 RMS vertVelocityTopRms               1.3553E-20            NORMALIZED  1.2903E-16
 RMS riverRunoffFluxRms               8.4703E-22            NORMALIZED  1.9164E-16
 RMS tangentialVelocityAvg            2.9646E-21            NORMALIZED  6.0992E-15
 RMS relativeVorticityAvg             8.6189E-26            NORMALIZED  4.5455E-02
 RMS enstrophyAvg                     2.0195E-28            NORMALIZED  2.1792E-16
 RMS layerThicknessPreviousTimestepAv 1.4211E-14            NORMALIZED  2.3165E-16
```

ambrad:
@worleyph for clarity: For a fixed PE layout, MALI could in principle not diff when swapping MPIs by using your reprosum module. Here it's diff'ing because of a bare MPI_Allreduce.

The separate issue of PE-layout BFB-variance even with reprosum is due to the preconditioners being mathematically different as a function of number of ranks.

philipwjones:
Reproducible sums in MPAS have been a to-do list item for a while. The new implicit solver does have a preconditioner that will change with PE layout in addition to the sums for inner products, but we do have a non-performant option that is b4b we can turn on for testing.

worleyph:
> For a fixed PE layout, MALI could in principle not diff when swapping MPIs by using your reprosum module. 

Do you mean for a fixed preconditioner? Not sure what is being "reproduced" if the PE layout remains fixed. Though some allreduce implementations may sum things differently from call to call (but generally this nondeterminism can be disabled?).

In general, reprosum is too expensive for (practical) use in iterative solvers, at least it was back in the POP days. There are cheaper almost-always-reproducible options that can be used if this is to be pursued, but the decomposition-specific preconditioner does sabotage this. Not sure if @philipwjones 's nonperformant option requires a reproducible allreduce or not.

ambrad:
@worleyph this is tangential to this issue, but what I mean is that the block preconditioners in MALI and the SI solver in MPAS form one block in the block-Jacobi matrix per rank, then apply incomplete factorizations or something like that to each block. This is true also when multigrid sits on top of the block preconditioners. When the number of ranks change, the blocks change, so the answer is mathematically different for a finite number of iterations. For this type of BFB-variance, you could leave #ranks fixed and swap MPIs and get BFB same answers. For a fixed #ranks, it's only the MPI_Allreduce calls in the Krylov subspace solver that lead to non-BFBness when swapping MPIs, and this is fixable if desired.

worleyph:
@ambrad , thank you for the clarification. We don't normally look at reordering the MPI ranks in our experiments, so I was not thinking of this. The only time reordering has come up in the past is when trying to minimize communication overhead by picking an ordering other than the default. We haven't worried about that for awhile (or even examined it), though @sarat did some work on that not too long ago.

worleyph:
@ambrad . Sorry - I'm being slow here. You mean two otherwise identical runs are likely to be assigned different compute nodes, and MPI_Allreduce may produce different results because of the differing hardware "topology"?

ambrad:
@worleyph, no. Sorry, I'm doing a poor job of explaining things.

This issue concerns non-BFBness when the MPI implementation is swapped from Intel to OpenMPI. That is, OpenMPI runs diff against IntelMPI-produced baselines in some cases. These cases are limited to bare MPI_Allreduce calls in (1) MPAS analysis members (purely diagnostic) and (2) MALI's Krylov subspace solver (prognostic). If a reproducible global sum were swapped in in these two cases, the diffs would go away: IntelMPI and OpenMPI runs would give identical output.

A separate source of non-BFBness, and the sort that leads to diffs among different PE layouts, is the result of preconditioners -- i.e., approximations to the matrix -- whose matrix values are a function of number of ranks. If the number of ranks changes, then the preconditioner matrix changes. But if the number of ranks is the same from run to run, then so are the answers. So this separate source of non-BFBness is not what this issue discusses.

rljacob:
@mark-petersen or @matthewhoffman or @jonbob can you confirm the fields listed in above comments are diagnostics?

vanroekel:
@rljacob all the fields you list above are diagnostic, from the global stats analysis member

worleyph:
@ambrad - I agree with your analysis. (But, again, reprosum is likely too costly for use in an iterative solver. But maybe MALI is executed at such a low frequency that we can tolerate this?) Sorry about misinterpreting the phrase "swapping MPIs". It is perfectly clear in retrospect, but the idea of swapping MPI tasks got stuck in my head.

As far as the preconditioner goes, if we used point instead of block (for Krylov or multigrid methods), the domain decomposition issue goes away, correct? (or even block that is independent of the domain decomposition, say the cells in the HOMME since cells are not decomposed during the dynamics) Back in the POP days, no one ever came up with a block preconditioner that actually accelerated things much. This definitely is not true for MALI, and also for MPAS-Ocean?

matthewhoffman:
@rljacob , for the MALI diffs you list, the two in mali.hist.am are diagnostic, but the three in cpl hist are not.  Those are coupling fields used by other components, so changes to those fields will impact the coupled system.  That test uses the shallow-ice velocity solver which does not include an implicit global velocity solve, so I would say the changes in those fields is unexpected (based on my current understanding of this thread).

philipwjones:
There are at least two reproducible sum algorithms out there that claim to be performant that I've wanted to try:
Arteaga et al (incl Oli Fuhrer) https://ieeexplore.ieee.org/document/6877351  and
Neuman et al (IJHPCA 2020) https://journals.sagepub.com/doi/10.1177/1094342020938425 

And yes, Pat, the preconditioner developed for the ocean barotropic solver is very effective and is a block preconditioner. I think the bit-for-bit option replaces the optimal block preconditioner with a less effective and simpler preconditioner.

worleyph:
@philipwjones , do you have copies of these articles that you can share? Thanks.

philipwjones:
@worleyph Sure, I'll email you. The second actually uses existing algorithms you're familiar with, but has just done an effective job at vectorizing/optimizing so they can make up some of the performance difference. The first does have non-negligible additional cost, though appears to be small for large applications.

rljacob (approved):

rljacob:
@matthewhoffman Since watercycle doesn't use MALI, we'll go ahead with this and open an issue about these MALI diffs.

rljacob:
Looking at the other test case with MALI that had diffs: SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.   This one has the same 2 fields diff with baselines in mali.hist.am but NO diffs in the cpl.hist or the mpassi and mpaso histories.

The main difference between the two is the one with coupler diffs has ELM and the other one uses the stub land model:
```
2000_DATM%NYF_SLND_MPASSI_MPASO_DROF%NYF_MALI%SIA_SWAV - no diffs in cpl hist
2000_DATM%QIA_ELM%SP_SICE_SOCN_MOSART_MALI%SIA_SWAV - diffs in cpl hist
```

matthewhoffman:
@rljacob , thanks for pointing out that difference.  It's not immediately obvious to me where such a difference would come from.  I saw the issue you opened and will follow up there once we've looked into it.

rljacob:
Saw expected diffs on next from diagnostics.  Blessing.

PR #4312, opened 2021-05-25
Update OpenMPI modules for Chrysalis on maint-1.0

amametjanov:
Update OpenMPI modules for Chrysalis on maint-1.0 and also
* make OpenMPI the default MPI library
* make SMT the default when multi-threading: MAX_TASKS_PER_NODE=128
* adjust MPI task affinity with SMT: --cpu_bind=threads when MAX_MPITASKS_PER_NODE =128
* calculate Slurm's -c/--cpus-per-task depending on tasks_per_node
* adjust Slurm's -m plane option for tasks_per_node
* set granularity of OMP threading depending on MAX_TASKS_PER_NODE

[non-BFB] for some ocean diagnostics and only on Chrysalis.

### DIFF
--- a/cime/config/e3sm/machines/config_machines.xml
+++ b/cime/config/e3sm/machines/config_machines.xml
@@ -996,7 +996,7 @@
     <NODENAME_REGEX>chr.*</NODENAME_REGEX>
     <OS>LINUX</OS>
     <COMPILERS>intel,gnu</COMPILERS>
-    <MPILIBS>impi,openmpi</MPILIBS>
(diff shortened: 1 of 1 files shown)


ndkeen (approved):

rljacob (approved):

rljacob:
Is this non-BFB because of the diagnostics as discussed in https://github.com/E3SM-Project/E3SM/pull/4309?  Or something else?

amametjanov:
Yes, I marked it non-BFB because the two MPI libraries weren't BFB on master.

PR #4314, opened 2021-05-26
Fix diagnostic outputs for ozone columns

tangq:
This PR adds the online calculations of the ozone column data needed by this e3sm diags plot (https://github.com/E3SM-Project/e3sm_diags/issues/395). These online diagnostics of ozone columns are more accurate than the offline calculations.

- Fix the existing total column ozone (TOZ) diagnostic by replacing the wet air mass with the dry air mass
- Add the online integrations for the stratospheric column ozone (SCO) and tropospheric column ozone (TCO) diagnostics

[BFB] except for tests that compare eam.h0 which will new diagnostic variables in SCO and TCO and changes in TOZ.

### DIFF
--- a/components/eam/src/chemistry/mozart/mo_chm_diags.F90
+++ b/components/eam/src/chemistry/mozart/mo_chm_diags.F90
@@ -398,6 +398,7 @@ contains
     real(r8), dimension(ncol)      :: df_noy, df_sox, df_nhx
 
     real(r8) :: area(ncol), mass(ncol,pver), drymass(ncol,pver)
+    real(r8) :: wrk1d(ncol)
     real(r8) :: wgt
     character(len=16) :: spc_name
     real(r8), pointer :: fldcw(:,:)  !working pointer to extract data from pbuf for sum of mass for aerosol classes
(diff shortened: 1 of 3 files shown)


rljacob:
Is this really fixing a bug?  was the previous ozone diags incorrect?

golaz:
@rljacob and @wlin7 : this is something we will need for finalizing v2.

components/eam/src/chemistry/mozart/mo_chm_diags.F90
```diff
@@ -448,19 +455,46 @@ subroutine chm_diags( lchnk, ncol, vmr, mmr, rxt_rates, invariants, depvel, depf
...
 
     call outfld( 'AREA', area(:ncol),   ncol, lchnk )
     call outfld( 'MASS', mass(:ncol,:), ncol, lchnk )
+    call outfld( 'DRYMASS', drymass(:ncol,:), ncol, lchnk )
 
-    ! convert ozone from mol/mol to DU
-    wrk(:ncol,:) = pdel(:ncol,:)*vmr(:ncol,:,id_o3)*avogadro*rgrav/air_molwgt/DUfac*1.e3_r8
-    ! total column ozone, vertical integration
+    ! convert ozone from mol/mol (w.r.t. dry air mass) to DU
+    wrk(:ncol,:) = pdeldry(:ncol,:)*vmr(:ncol,:,id_o3)*avogadro*rgrav/mwdry/DUfac*1.e3_r8
```
wlin7, line 466: Hi @tangq , Though no impact on the results, this conversion is repeated three times. I think it is preferable doing the conversion only once and putting them on wrk, then using a different horizontal-only work array for the calculations of TOZ, SCO and TCO.

tangq: @wlin7 , thanks for the suggestions. I implemented the changes per your suggestion in the new commit.

tangq: I also confirmed that the new commit gives the same results for ozone columns.

tangq:
> Is this really fixing a bug? was the previous ozone diags incorrect?

@rljacob , the previous diags uses the wet air mass, which is incorrect. This PR fixes it with the right dry air mass in the calculation.

wlin7 (approved):
Thanks for the update and verification test. All the changes are straightforward to follow.

tangq:
Good morning, @wlin7 .

It's ready to be merged now. Since v2 needs this, can it be merged today?

wlin7:
Yes, Qi. This PR will be merged to next today.

wlin7:
Note: This PR will be perceived as non-BFB for tests that compare eam.h0 files as existing variable TOZ will differ and new variables (SCO and TCO) are added.

wlin7:
Total column ozone, TOZ, whose results would change due to this PR, is not saved in eam.h0; tests camparing eam.h0 did not report DIFF, but did report having 2 new fields. The DIFFs in today's return are not due to this PR. Let other PRs to claim today's diff first. The tests involving eam.h0 should still be blessed, eventually.

Note: though total ozone is not saved to h0, it can be calculated as TOZ=SCO+TCO.  @tangq , is this the plan?

wlin7:
Merged to master. 

Bless wcprod tests on anvil and compy (rrm), and on mappy for the following two tests that report diff on eam.h0
```
SMS_Ln9.ne4_ne4.FC5AV1C-L.mappy_gnu.eam-outfrq9s
SMS_R_Ld5.ne4_ne4.FSCM5A97.mappy_gnu.eam-scm
```
To review similar tests like the above two (involving eam.h0) and bless them though they don't report a diff.

Compy's other wcprod test failed to build; to be blessed later.

tangq:
> Note 1: though total ozone is not saved to h0, it can be calculated as TOZ=SCO+TCO. @tangq , is this the plan?

@wlin7 , this is correct. Total column can be calculated by adding the tropospheric and stratospheric components together. For v2 production runs, we need the daily total column ozone for the ozone hole diagnostics, so will save it separately on the h2 tape.

PR #4317, opened 2021-05-27
Fixes MOSART unstructured grid test on cori

bishtgautam:
Increases number of MPI tasks for the test

[BFB]

### DIFF
--- a/components/mosart/cime_config/testdefs/testmods_dirs/mosart/mos_usrdat/shell_commands
+++ b/components/mosart/cime_config/testdefs/testmods_dirs/mosart/mos_usrdat/shell_commands
@@ -3,4 +3,5 @@
 ./xmlchange LND_DOMAIN_PATH="\$DIN_LOC_ROOT/share/domains"
 ./xmlchange ATM_DOMAIN_PATH="\$DIN_LOC_ROOT/share/domains"
 ./xmlchange DATM_CLMNCEP_YR_END=1972
+./xmlchange NTASKS=64
 


ndkeen (commented):
Should this sort of change be made to default PE layouts?  I don't have strong opinion either way. I know that using 64 MPI's looks good on cori-knl, but that's just the natural choice for that machine as it fits on one node. Should the number of MPI's be based on the machine? Or pick value that would ensure it only runs on 1 node for all machines? Say 16 or 32?

components/mosart/cime_config/testdefs/testmods_dirs/mosart/mos_usrdat/shell_commands
```diff
@@ -3,4 +3,5 @@
 ./xmlchange LND_DOMAIN_PATH="\$DIN_LOC_ROOT/share/domains"
 ./xmlchange ATM_DOMAIN_PATH="\$DIN_LOC_ROOT/share/domains"
 ./xmlchange DATM_CLMNCEP_YR_END=1972
+./xmlchange NTASKS=64
```
rljacob, line 6: wrap this in "if machine=cori-knl".

ndkeen:
I'm testing a change to add a default layout for 1 node on cori (ie use 64 MPI's).
`ndk/machinefiles/cori-1node-default-layout`

ndkeen:
Adding a default for cori did work and allowed this test to pass. It just uses 64 MPI's (no threads) as a default unless overridden.  However, I see that several other tests are now picking this up. About 10 tests will use this.  All of them pass. Should I make a new issue/PR ?

rljacob:
Yes go ahead.

ndkeen:
I think https://github.com/E3SM-Project/E3SM/pull/4408 may be a better way to increase the number of MPI's used by this test on cori (and for any 1-node job). So we could probably close this?

ndkeen:
I think you can close this as 4408 has been merged

PR #4318, opened 2021-05-27
Set GMAKE_J to 4 for 'singularity' machine

lukaszlacinski:
By default, GMAKE assumes 16 threads/CPUs available on the 'singularity' machine. It sometimes causes a failure when building kokkos on hardware that has less CPUs like small workstations and laptops. Then, a user is required to set explicitly GMAKE_J to a value that matches the user's hardware. To avoid that additional step, this PR changes GMAKE_J default value, 16, to 4 which should work for most small workstations and laptops.


rljacob (approved):

rljacob:
@lukaszlacinski I tried merging this but the branch is based off a very old version of master.

Please update master on your fork, make a new branch which just this change, and make a new PR.

rljacob:
@lukaszlacinski nevermind.  I'm going to add this to PR #4378

Issue #4319, opened 2021-05-27
Incorrect results at start of simulation with seaice conservation analysis member


akturner:
When a simulation starts or restarts the sea ice conservation analysis member generates incorrect results

PR #4320, opened 2021-05-27
Set compute on startup to false for sea ice conservation analysis member

akturner:
Set compute on startup to false for sea ice conservation analysis member. This fixes an issue with the sea ice conservation analysis member with incorrect results as the start or restart of a simulation.

Fixes https://github.com/E3SM-Project/E3SM/issues/4319

[BFB]


proteanplanet:
Unfortunately, this doesn't fix the problem.  I have run a test on Chrysalis, restarting every year, and the config_am_conservationCheck_compute_on_startup set to false, and the output file on the restart year (in this case every year), only include 11 months, and not the January values.   You may like to look at the simulation to see the issue: /lcrc/group/e3sm/ac.afroberts/E3SM_simulations/20210527.v2rc2c.ConservationAM.ne30pg2_EC30to60E2r2.chrysalis/run

proteanplanet (changes requested):
See comment above.

akturner:
@proteanplanet: Try conservationCheckOutput:clobber_mode="append"

proteanplanet:
@akturner I am currently serving on a review panel, and request that you please run a quick B-case to test this, and make sure it works, whereupon I can check it if you point me to the location.

proteanplanet:
conservationCheckOutput:clobber_mode="append" worked without the need to set config_AM_conservationCheck_compute_on_startup = false.  I have created a separate branch at https://github.com/E3SM-Project/E3SM/tree/proteanplanet/ice/ConsAMFix, and will creat the PR once I have several years integrated in E3SM for checking.  This PR can be closed and need not be integrated into E3SM.

akturner:
@proteanplanet: Thanks for testing this. I don't think you want to compute on startup either, so would prefer to merge this PR than the other. Is that ok?

proteanplanet:
@akturner This PR does not work. Please look at the output from the other PR, and the results for this one. I am also fixing the area problem now in the other branch.  The accumulated area is never divided by the number of timesteps, hence the problem Nicole mentions with the area statistics.  Also, I can't make changes to this PR consistent with recent updates to E3SM, which is very fluid, because you've done it on your fork - the normal E3SM workflow is to branch from master, and not create a fork.   If you would like to proceed with this PR, please present a fully tested B-case for us to review.  However, I would prefer to proceed with my PR.

akturner:
@proteanplanet  I added the append change to this branch. So it should work. We can use the other branch but can you make the change to the compute on startup. Thanks

proteanplanet:
@akturner Ok. I am making the startup change on the other branch, and testing it in a B-case, whereupon I'll ask you to review it, also with the area change.

PR #4322, opened 2021-05-27
Add support for land on bi-grid ne1024pg2_oRRS18to6v3

### LINKED ISSUE #4311: Long integer needed for writing elm restart file at 1024pg2
The file directly involved is elm/src/main/ncdio_pio.F90.

The size of the dimension 'column' is  116501645.  In subroutine ncd_getiodesc,  for an instance, to declare an array of (column=116501645, levtot=20), an expression amounts to 116501645*20 would be assigned to 4-byte integer 'fullsize', which ends up having an invalid value of ' -1964934396'.  Similar problem can also occur to the assignments to compDOF. A minval(compDOF)=-2092163035 is also detected when turning on debugging. 

8 byte integer must be used for at least the following
 ```
 integer*8 :: fullsize                      ! size of entire array on cdf
 integer*8, pointer  :: compDOF(:)
```
What is less clear is that compDOF is passed into
 `call pio_initdecomp(pio_subsystem, baseTYPE, dims(1:ndims), compDOF, iodesc_list(iodnum)%iodesc)`

Should changes also be made to the subroutine pio_initdecomp in piolib_mod.F90. piolib_mod.F90 appears to have interfaces to recognize both _i4 and _i8. for DOF.

With integer*8 for fullsize, the elm restart writing is successful, but continuing run with the restart file would fail with large errors on various balance checks.  This is before using 8-byte for compDOF. 

Further testing underway, to check if the types for other variables also need to be updated. A PR will be created after the issue is resolved.

### PR DESCRIPTION
**Pull request by wlin7**
This is to support land on ne1024pg2 grid as the atmosphere. River model
for this grid configuration continues to use r0125.

Land restart file at ne1024pg2 has to use 64bit_data format. To enable,
Changes are made to have ELM use the netcdf format as specified via
PIO_LND_NETCDF_FORMAT. Existing codes are hardwired to use 64BIT_OFFSET.

To support creating large land files at ne1024pg2, mainly restart files,
changes are also made to use 8-byte long integer for resolving the fullsize
of the data array and indexing the associated segments.

Fixes #4311.
 [BFB]

### DIFF
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -2579,8 +2579,8 @@
     <domain name="ne1024np4.pg2">
       <nx>25165824</nx>
       <ny>1</ny>
-      <file grid="atm|lnd" mask="oRRS18to6v3">$DIN_LOC_ROOT/share/domains/domain.lnd.ne1024pg2_oRRS18to6v3.210523.nc</file>
-      <file grid="ice|ocn" mask="oRRS18to6v3">$DIN_LOC_ROOT/share/domains/domain.ocn.ne1024pg2_oRRS18to6v3.210523.nc</file>
+      <file grid="atm|lnd" mask="oRRS18to6v3">$DIN_LOC_ROOT/share/domains/domain.lnd.ne1024pg2_oRRS18to6v3.200212.nc</file>
+      <file grid="ice|ocn" mask="oRRS18to6v3">$DIN_LOC_ROOT/share/domains/domain.ocn.ne1024pg2_oRRS18to6v3.200212.nc</file>
       <desc>ne1024np4.pg2 is Spectral Elem 3km grid w/ 2x2 FV physics grid per element:</desc>
     </domain>
 
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -1537,6 +1537,16 @@
       <mask>oRRS18to6v3</mask>
     </model_grid>
 
+    <model_grid alias="ne1024pg2_oRRS18to6v3">
+      <grid name="atm">ne1024np4.pg2</grid>
+      <grid name="lnd">ne1024np4.pg2</grid>
+      <grid name="ocnice">oRRS18to6v3</grid>
+      <grid name="rof">r0125</grid>
+      <grid name="glc">null</grid>
+      <grid name="wav">null</grid>
+      <mask>oRRS18to6v3</mask>
+    </model_grid>
+
     <model_grid alias="ne1024np4_360x720cru_oRRS15to5" compset="(DOCN|XOCN|SOCN|AQP1)">
       <grid name="atm">ne1024np4</grid>
       <grid name="lnd">360x720cru</grid>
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by wlin7** [comment]
The approach to have ELM respect the NETCDF_FORMAT set in env_run.xml (and support cdf5) mirrors the works done by @brhillman and @jayeshkrishna for atm (PR [#2911](https://github.com/E3SM-Project/E3SM/pull/2911)). The difference is that ELM didn't have a similar module like eam_instance.F90. A new module elm_instance.F90 is therefore introduced. It seems possible the function could be added to the existing module elm_instMod.F90. Opt not to unless recommended.

**Comment by wlin7** [comment]
For the grid to work, it still needs a fsurdat, and eventually a finidat. fsurdat for ne1024pg2 is not yet finalized.

For testing, the following files can be used

```
fsurdat='/global/cfs/cdirs/e3sm/wlin/inputdata/ne1024pg2/surfdata_ne1024pg2_simyr2010_c210521.nc'
finidat='/global/cfs/cdirs/e3sm/wlin/inputdata/ne1024pg2/20210524.ICRUELM.ne1024pg2_oRRS18to6v3.elm.r.2010-01-01-00000.nc'
```

**Comment by wlin7** [comment]
With the file changes in this PR, it has been verified restart BFB for this grid. With 8192x2 land tasks and staggered 7680x2 tasks for other components on 256 cori-knl nodes,  the throughput is 10.9 SYPD.

**Comment by rljacob** [comment]
what compset is getting 10 SYPD with land at ne1024?

**Comment by wlin7** [comment]
> what compset is getting 10 SYPD with land at ne1024?

Oh, forgot to mention. Land-only, ICRUELM. First test of the grid, as a means to prepapre land initial condition. Thanks for bringing this up, Rob.

File: cime_config/config_grids.xml [hunk 1]
```diff
@@ -2569,8 +2579,8 @@
     <domain name="ne1024np4.pg2">
       <nx>25165824</nx>
       <ny>1</ny>
-      <file grid="atm|lnd" mask="oRRS18to6v3">$DIN_LOC_ROOT/share/domains/domain.lnd.ne1024pg2_oRRS18to6v3.200212.nc</file>
-      <file grid="ice|ocn" mask="oRRS18to6v3">$DIN_LOC_ROOT/share/domains/domain.ocn.ne1024pg2_oRRS18to6v3.200212.nc</file>
+      <file grid="atm|lnd" mask="oRRS18to6v3">$DIN_LOC_ROOT/share/domains/domain.lnd.ne1024pg2_oRRS18to6v3.210523.nc</file>
```
brhillman, line 2582: Why do these need to be updated? Did the domain files change for some reason? I think these should be the same.

wlin7: Good point, Ben.  I was curious . There were a  number of domain files that have the same time stamp (200212). But the files for ne1024pg2_oRRS18to6v3 do not exist in the inputdata (not on cori where I did the test).  Just check on the inputdata server. The 200212 domain files are available there. I will revert this change.

brhillman (approved):

rljacob:
Status:  waiting on generating surface map data for this grid.

rljacob:
note:  will get an update and then this can be merged.

rljacob:
Note:  this resolution is being tested in SCREAM.  Might need more input file updates.

rljacob:
Test run was successful.

bishtgautam (approved):

wlin7:
Thanks for merging this PR, @rljacob . 

I noticed in the commit message that you had to resolve the conflict for fsurdat setting for simyr2010 at ne30np4.pg2 grid.
The intended one is put in via #4543. The surfdata for ne1024pg2 is also replaced with an improved one.

PR #4327, opened 2021-06-02
Remove limit on shr_frz_freezetemp and replace it with an abort

jonbob:
The share code in the coupler currently calculates shr_frz_freezetemp based on a seawater freezing temperature option, but then limits it to not allow a freezing temperature below -2.0. This PR removes that limit and replaces it with an abort if the shr_frz_freezetemp is calculated to be below -2.6C.

[non-BFB] for some A-, C-, and D-cases

### DIFF
--- a/share/util/shr_frz_mod.F90.in
+++ b/share/util/shr_frz_mod.F90.in
@@ -105,11 +105,11 @@ contains
     endif
 
 #if ({DIMS}==0)
-    if (shr_frz_freezetemp < -2.5_R8) &
-       call shr_sys_abort(subname//' ERROR: shr_frz_freezetemp is less than -2.5')
+    if (shr_frz_freezetemp < -2.6_R8) &
+       call shr_sys_abort(subname//' ERROR: shr_frz_freezetemp is less than -2.6')
 #else
-    if (minval(shr_frz_freezetemp) < -2.5_R8) &
-       call shr_sys_abort(subname//' ERROR: shr_frz_freezetemp is less than -2.5')
+    if (minval(shr_frz_freezetemp) < -2.6_R8) &
+       call shr_sys_abort(subname//' ERROR: shr_frz_freezetemp is less than -2.6')
 #endif
 
   end function shr_frz_freezetemp_{DIMS}d
--- a/share/util/shr_frz_mod.F90.in
+++ b/share/util/shr_frz_mod.F90.in
@@ -104,7 +104,8 @@ contains
             &call shr_frz_freezetemp_init first with a valid tfreeze_option')
     endif
 
-    shr_frz_freezetemp = max(shr_frz_freezetemp,-2.0_R8)
+    if (shr_frz_freezetemp<-2.0_R8 .and. s_loglev > 0) &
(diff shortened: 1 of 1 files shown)


jonbob:
@proteanplanet - do we want to output a warning to the logs whenever it calculates a temperature below -2? Or something more extreme?

proteanplanet:
> @proteanplanet - do we want to output a warning to the logs whenever it calculates a temperature below -2? Or something more extreme?

I would be happy if it was set to crash at -2.5˚C, but otherwise proceeded gracefully, with no other limit applied.  Before answering, I checked the WOCE climatology, and it's not uncommon for ocean surface temperatures to dip below -2˚C in parts of the Southern Ocean, so I'm at a loss as to why this number was chosen as a limit, much less an overriding limit to the physics applied.

xylar:
> I would be happy if it was set to crash at -2.5˚C, but otherwise proceeded gracefully, with no other limit applied. 

I would appreciate having a run with ice-shelf cavities performed before such a limit goes in.  We certainly do end up with water that cold at the ocean surface, but not in locations that the sea-ice coupler should see.  Still, it would be an unpleasant surprise if Cryo runs stop working because of this limit.

jonbob:
@xylar - please note that this code isn't used in fully-active configurations, only A- (all data), C- (ocean only), and D- (seaice only) cases. So not even G-cases. However, I'm happy to test it in a C-case on a mesh with ice-shelf cavities, if that even is a configuration we expect to be reasonable.

xylar:
@jonbob, ah, I missed that detail.  Ignore my comment.

jonbob:
Testing runs of CMPASO-IAF and DTESTM on a T62_EC30to60E2r2 resolution show this lower freezing temperature limit may not be low enough. The C-case has gone almost 2 years with no issues but the D-case failed right away in the abort line added by this PR. I added more output and minimum freezing temperatures are lower than -2.5. I looked at the docn file used for this test:
```
/lcrc/group/e3sm/data/inputdata/ocn/docn7/SOM/pop_frc.1x1d.090130.nc
```
and the maximum salinity value it has is 45.38462, which translates to a freezing temperature of -2.57264. So I'm currently testing with the abort test checking for shr_frz_freezetemp less than -2.6.

jonbob:
A DTESTM run using a T62_EC30to60E2r2 resolution went five years successfully on anvil. The corresponding CMPASO-IAF run crashed in the fourth year with an unrelated node failure. @proteanplanet - I believe this is now ready for review

proteanplanet (approved):
This is approved as a consequence of testing, but also because a hard stop occurs if freezing temperatures become seemingly unreasonable in A, D and C cases.  The alternative was that the code proceeded when a limit was applied, unknown to the user.   This approach is much improved.  The limit could be lowered at a later time if it proves too restrictive.

jonbob:
results from running e3sm_developer on chrysalis with a test merge yields all PASS except:
* DIFF ERS_Ld5.T62_oQU120.CMPASO-NYF.chrysalis_intel (phase BASELINE)
which is expected

jonbob:
merged to next

jonbob:

PR #4334, opened 2021-06-08
Update pgigpu tests on Blues

amametjanov:
Update pgigpu tests on Blues:
- update path in cime_config/machines/Depends.ibmgpu.cmake and Depends.pgigpu.cmake;
- reduce num_nodes in `pgigpu` tests from 3 to 1 for better queue throughput.

[BFB]

### DIFF
--- a/cime_config/machines/Depends.ibmgpu.cmake
+++ b/cime_config/machines/Depends.ibmgpu.cmake
@@ -53,7 +53,7 @@ set(FILES_NEED_OPENMP_FLAGS
   eam/src/physics/crm/samomp/params.F90
   eam/src/physics/crm/samomp/dmdf.F90
   eam/src/physics/crm/samomp/domain.F90
-  eam/src/physics/crm/samomp/ecppvars.F90
+  eam/src/physics/crm/ecppvars.F90
   eam/src/physics/crm/samomp/fft.F90
   eam/src/physics/crm/samomp/fftpack5.F90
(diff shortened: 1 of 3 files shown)


cime_config/machines/Depends.ibmgpu.cmake
```diff
@@ -53,7 +53,7 @@ set(FILES_NEED_OPENMP_FLAGS
   eam/src/physics/crm/samomp/params.F90
   eam/src/physics/crm/samomp/dmdf.F90
   eam/src/physics/crm/samomp/domain.F90
-  eam/src/physics/crm/samomp/ecppvars.F90
+  eam/src/physics/crm/ecppvars.F90
```
amametjanov, line 56: @whannah1 this path change appears to be done in E3SM-Project/E3SM#4301. Need to propagate it to other compilers too.

rljacob: Did you add this change here as a courtesy or does Blues have the IBM compiler?

amametjanov: Just wanted to update all related Depends.* files in one PR. This will be needed for `ibmgpu` on Summit.

whannah1 (approved):
Looks good. 
Note to self - we can probably delete everything in the depends files that has to do with OpenACC since we don't plan to use that version of the CRM for the GPU.

PR #4337, opened 2021-06-10
Fix extrasnowlayers test, using corrected compset and updated name

jonbob:
Fixes a mis-named test in the e3sm_ocnice_extra_coverage suite, which had an incorrect compset and was using "clm" instead of "elm" for the testmod.

[BFB]


rljacob (approved):

jonbob:
merged to master

PR #4343, opened 2021-06-14
Add output of some nonhydro fields on dyn grid

oksanaguba:
Added w_i, z3_i, mu_i, pnh.

[bfb]

### DIFF
--- a/components/homme/src/theta-l/share/element_ops.F90
+++ b/components/homme/src/theta-l/share/element_ops.F90
@@ -154,8 +154,11 @@ recursive subroutine get_field(elem,name,field,hvcoord,nt,ntQ)
          field = elem%state%w_i(:,:,1:nlevp,nt)
       endif
     case('geo_i');
-      !is phinh_i set for HY runs or should here be an abort?
-      field = elem%state%phinh_i(:,:,1:nlevp,nt)
+      if(theta_hydrostatic_mode) then
+          call abortmp('ERROR: get_field_i is not supported for phi_i in theta HY')
+      else
+          field = elem%state%phinh_i(:,:,1:nlevp,nt)
+      endif
     case('mu_i');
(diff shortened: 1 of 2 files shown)


oksanaguba:
I tested this in scream PR with identical changes and here for just one of those runs, nonhydrostatic.

components/homme/src/theta-l/share/element_ops.F90
```diff
@@ -154,8 +154,11 @@ subroutine get_field_i(elem,name,field,hvcoord,nt)
          field = elem%state%w_i(:,:,1:nlevp,nt)
       endif
     case('geo_i');
-      !is phinh_i set for HY runs or should here be an abort?
-      field = elem%state%phinh_i(:,:,1:nlevp,nt)
+      if(theta_hydrostatic_mode) then
+          call abortmp('ERROR: get_field_i is not supported for phi_i in theta HY')
+      else
+          field = elem%state%phinh_i(:,:,1:nlevp,nt)
+      endif
```
mt5555, line 161: since we are modifying this file, might as well fix it instead of aborting.  Something like:

    if(theta_hydrostatic_mode) then
       call phi_from_eos(hvcoord,elem%state%phis,elem%state%vtheta_dp(:,:,:,nt),elem%state%dp3d(:,:,:,nt),field)
    else
       field = elem%state%phinh_i(:,:,:,nt)
    endif

mt5555 (changes requested):

components/homme/src/theta-l/share/element_ops.F90
oksanaguba: done, thanks. ran HY output and all looks ok.

mt5555 (approved):

rljacob:
Is this ready to merge?

oksanaguba:
in next

Issue #3477, opened 2020-03-02
MMF ERP test w/ physgrid is not BFB with prescribed aerosols


whannah1:
While finalizing the development on the branch used for PR #3465 I realized that the new MMF configuration is not BFB when used with 1-moment microphysics on the pg2 grid. This is evident by a failure during the "COMPARE_base_rest" phase of the test that checks if the restart run matches the result of the initial run. I confirmed the failure by checking that the global energy values printed in the atm log files diverge at the point where the restart run begins (timestep 6 for a Ln9 test).

The 1-moment microphysics requires using prescribed aerosols, so I'm speculating that there might be a problem generally with using prescribed aerosols on the physgrid. This issue also appears to be present on the master branch of the [ECP fork](https://github.com/E3SM-Project/ACME-ECP) that was used for the MMF development up to this point. If this is correct then the non-MMF tests with prescribed aerosols on the physgrid should also fail, but I've had difficulty confirming this either way. Although, I can get a non-MMF prescribed aerosol test with pg2 to FAIL. I think to really confirm my suspicion I need to get the same test to pass again by zeroing out the aerosols fields each time step, or something similar. All my attempts to do this lead to crashes that I can't explain.  

Some other important pieces of evidence
- ERS tests with 1-moment microphysics pass on both np4and pg2 grids
- MMF tests with 2-moment microphysics pass on both np4 and pg2 grids

Here's a list of test results from Cori for reference:
- **FAIL** - ERP_Ln9_P96.**ne4pg2_ne4pg2**.F-MMF1-TEST
- PASS - ERP_Ln9_P96.ne4_ne4.F-MMF1-TEST
- PASS - ERS_Ln9_P96.ne4pg2_ne4pg2.F-MMF1-TEST
- PASS - ERS_Ln9_P96.ne4_ne4.F-MMF1-TEST
- PASS - ERP_Ln9_P96.ne4_ne4.F-MMF2-TEST
- PASS - ERP_Ln9_P96.**ne4pg2_ne4pg2**.F-MMF2-TEST

There is another (perhaps related) issue (also evident on ECP/master) where MMF ERP tests will fail on restart comparison when the number of tasks/processors is not specified, as evidenced by these results:
- **FAIL** - ERP_Ln9.ne4_ne4.F-MMF1-TEST
- PASS - ERP_Ln9_P96.ne4_ne4.F-MMF1-TEST

[...]

PR #4333, opened 2021-06-08
Split tphysbc for MMF to allow OMP threading

whannah1:
This splits the tphysbc() routine into 2 separate routines that can be called in separate threaded chunk loops while calling the CRM in between without threading. The mmf_clear_rh variable is converted to a pbuf variable to avoiding passing it around in phys_run1().  I also included some clean up of physpkg.F90 since there were several unnecessary variables and use statements.

This change allows more flexibility for tuning performance on GPU machines, but will require another PR to refactor crm_physics.F90 and push the chunk loop down into crm_physics_tend() so that all CRMs in a given MPI task can be solved in parallel.

[BFB]

PR #4344, opened 2021-06-14
Add test mod to get passing MMF ERP tests

whannah1:
This adds a new test mod "eam-mmf_max_subcycle" that allows ERP tests to pass and updates MMF tests in integration suite.
Fixes issue #3477

[BFB] except for modified test names


cime_config/tests.py
```diff
@@ -133,7 +133,7 @@
             "SMS.T62_oQU120_ais20.MPAS_LISIO_TEST",
             "SMS.f09_g16_a.IGELM_MLI",
             "SMS_P12x2.ne4_oQU240.A_WCYCL1850.allactive-mach_mods",
-            "ERS_Ln9.ne4pg2_ne4pg2.F-MMF1.eam-mmf_crmout",
+            "ERS_Ln9_P96x1.ne4pg2_ne4pg2.F-MMF1.eam-mmf_crmout",
```
brhillman, line 136: You cannot specify 96 procs here because we run developer on mappy, which does not have 96 procs.

rljacob: If you want to run on that many, you could move the test to integration.

components/eam/cime_config/testdefs/testmods_dirs/eam/mmf_max_subcycle/shell_commands
```diff
@@ -0,0 +1 @@
+./xmlchange --append -id CAM_CONFIG_OPTS -val " -cppdefs ' -DMMF_DIR_NS -DMMF_MAX_SUBCYCLE ' "
```
brhillman, line 1: I wish we had a better way of specifying the NS orientation. If we were to change from that default at some point in the compset definitions, this testmod would override it.

cime_config/tests.py
whannah1: How many tasks can mappy run with? We could just set it to a lower value. We mainly just need to ensure that threading is disabled (until the thread fix gets merged).

components/eam/cime_config/testdefs/testmods_dirs/eam/mmf_max_subcycle/shell_commands
whannah1: Yea, I agree. We could bake it into the code or make it a namelist variable instead.
making it a namelist variable would make a lot of sense. I can make a new PR for that.

cime_config/tests.py
```diff
@@ -165,10 +165,10 @@
             "SMS_Ld2.ne30_oECv3.BGCEXP_BCRC_CNPRDCTC_1850.elm-bgcexp",
             "SMS_D_Ld1.T62_oEC60to30v3.DTESTM",
             "SMS_D_Ld1.ne30_r05_oECv3.A_WCYCL1850",
-            "ERS_Ln9_P96x1.ne4pg2_ne4pg2.F-MMFXX",
+            "ERP_Ln9_P96x1.ne4pg2_ne4pg2.F-MMFXX.eam-mmf_max_subcycle",
```
brhillman, line 168: I kind of think "fixed_subcycle" might be more appropriate, since it's the fact that the subcycle is normally dynamic that breaks ERP tests, and using a fixed ncycle is what is needed to pass ERP. But I'm fine keeping "max". Just something to consider.

whannah1: good point, I'll change it.

whannah1: Done.

cime_config/tests.py
whannah1: Actually, the more I think about this I should just cancel this PR, wait till #4333 gets merged and then make sure that the MMF tests don't specify the task count at all.

brhillman: I think the main point was to force the number of threads to be 1, not to specify the task count. Could temporarily add an `xmlchange NTHRDS=1` to the test mod but not a bad idea to just wait until threading is fixed. Alternatively I think you could add machine default PR layouts for F-MMF cases that specify 1 thread.

brhillman:
Merged to next.

brhillman:
Re-merged to next.

cime_config/tests.py
```diff
@@ -166,11 +166,11 @@
             "SMS_Ld2.ne30_oECv3.BGCEXP_CNTL_CNPRDCTC_1850.elm-bgcexp",
             "SMS_D_Ld1.T62_oEC60to30v3.DTESTM",
             "SMS_D_Ld1.ne30pg2_r05_EC30to60E2r2.A_WCYCL1850S_CMIP6",
-            "ERS_Ln9_P96x1.ne4pg2_ne4pg2.F-MMFXX",
-            "ERS_Ln9_P96x1.ne4pg2_ne4pg2.F-MMFXX.eam-mmf_use_VT",
+            "ERP_Ln9.ne4pg2_ne4pg2.F-MMFXX.eam-mmf_fixed_subcycle",
+            "ERS_Ln9.ne4pg2_ne4pg2.F-MMFXX.eam-mmf_use_VT",
             "ERS_Ln9_P96x1.ne4pg2_ne4pg2.F-MMFXX.eam-mmf_use_ESMT",
```
amametjanov, line 171: These two are erroring out on Theta (Cori-KNL too) with `forrtl: severe (154): array index out of bounds`:
```
ERS_Ln9.ne4pg2_ne4pg2.F-MMFXX.eam-mmf_use_VT
ERS_Ln9_P96x1.ne4pg2_ne4pg2.F-MMFXX.eam-mmf_use_ESMT
```

cime_config/tests.py
```diff
@@ -166,11 +166,11 @@
...
-            "ERS_Ln9_P96x1.ne4pg2_ne4pg2.F-MMFOMP",
-            "ERS_Ln9_P96x1.ne4pg2_ne4pg2.F-MMF1-RCEMIP",
+            "ERP_Ln9_P96x1.ne4pg2_ne4pg2.F-MMFOMP.eam-mmf_fixed_subcycle",
```
amametjanov, line 172: This has build errors on Theta (and Chrysalis) due to threading flags (ERP forces a threaded build). Probably best to keep it as `ERS` until OpenMP offload is fixed for CPU-only runs with Intel.

whannah1: Ah, I didn't realize ERP would force threading. Why is that the case? 
Either way, I'll revert this test to ERS in a new PR.

cime_config/tests.py
whannah1: That's odd. The "use_ESMT" test was having issues on Cori, but I thought I fixed it by putting the "96x1" back in. I need to do some digging to figure this out.

whannah1: I just ran some tests on Cori and the "use_VT" test passes with 2 threads, but fails with 1... Not sure what to make of that.

whannah1: Ok, I've identified the issue. It was a tricky to track down, but both tests were actually broken by a previous PR that updated the YAKL FFT routines. I'm going to create an issue, but in the mean time those tests will need to stay broken until we can create a new PR to update the FFT calls.

brhillman:
Re-merged to next to revert OMP test to ERS without threads.

brhillman:
Merged to master.

PR #4338, opened 2021-06-11
Fix restart and area bug in MPAS-SI Conservation Analysis Member

proteanplanet:
This corrects a problem in the MPAS-SeaIce Conservation Analysis member configuration whereby the first month of history after a restart of E3SM was dropped from the analysis history.  Now, with this PR, results from the a restarted simulation are appended to the existing analysis member file for the restart year from the simulation previous to the restart, if such a file exists.   This PR does not change answers in E3SM, but ensures that the new diagnostics on sea ice conservation are able to be calculated seamlessly over hundreds of years, if need be. The PR branch code has been tested in a four-year B-case with annual restarts, for which results are given in comments below subsequent to this pull request. [BFB]

PR #4345, opened 2021-06-15
Fix restart and area bug in MPAS-SI Conservation Analysis Member (renamed)

proteanplanet:
This corrects a problem in the MPAS-SeaIce Conservation Analysis member configuration whereby the first month of history after a restart of E3SM was dropped from the analysis history.  It also corrects another problem in that the mean monthly area of sea ice on the mesh was not being reset or averaged correctly.  This PR is identical to https://github.com/E3SM-Project/E3SM/pull/4338, but corrects the naming of the branch to meet the E3SM convention.  Results follow. Location for test simulation with yearly restart is at:

/lcrc/group/e3sm/ac.afroberts/E3SM_simulations/20210609.v2rc2c.ConservationAM2.ne30pg2_EC30to60E2r2.chrysalis/run

image

`
Conservation checks: 0007-01-01_00:00:00

Area analysis Global NH SH

Earth radius (m) = 6.371229E+06
Earth area (m2) = 5.101011E+14
Domain area (m2) = 3.603951E+14 1.537943E+14 2.066007E+14
Sea-ice area (m2) = 2.295497E+13 1.246381E+13 1.049116E+13

[NML]
[BFB]


proteanplanet:
This work has already passed review under  #4338.

jonbob:
passes sanity testing
merged to next

jonbob:
merged to master and NML DIFFs blessed

Issue #4347, opened 2021-06-17
Iceberg latent heat flux has wrong sign in data iceberg forcing (DIB)


darincomeau:
MPAS-Seaice calculates the latent heat flux associated with the data iceberg forcing (DIB) to be sent to the ocean. The sign is currently positive, meaning it gives heat to the ocean, when it should be negative., analogous to the sea ice latent heat flux to the ocean.

From a `mpaso.hist.am.timeSeriesMonthly.` from a recent `CRYO1850` run:

```        float timeMonthly_avg_icebergHeatFlux(Time, nCells) ;
                timeMonthly_avg_icebergHeatFlux:units = "W m^{-2}" ;
                timeMonthly_avg_icebergHeatFlux:long_name = "Iceberg heat flux at cell centers from coupler. Positive into the ocean." ;
```

and

```        float timeMonthly_avg_seaIceHeatFlux(Time, nCells) ;
                timeMonthly_avg_seaIceHeatFlux:units = "W m^{-2}" ;
                timeMonthly_avg_seaIceHeatFlux:long_name = "Sea ice heat flux at cell centers from coupler. Positive into the ocean." ;
```
So the sign convention is the same.

The two variables however are different signs:

```data:

 timeMonthly_avg_icebergHeatFlux =
  0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0.04608132,
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0.1605959, 0, 0, 0,
    8.760676e-06, 0.3183545, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0.0179162,
```

whereas the corresponding sea ice latent heat flux shows:

```data:

 timeMonthly_avg_seaIceHeatFlux =
  -1.841673, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, -9.961944, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0, 0, 0, -2.313874, 0, -9.972008, 0, 0, 0, 0, 0, 0, 0,
```

[...]

PR #4348, opened 2021-06-17
Changes sign of iceberg latent heat flux.

darincomeau:
Changes the sign of the iceberg latent heat flux in MPAS-Seaice, which is then sent to MPAS-Ocean.

Fixes https://github.com/E3SM-Project/E3SM/issues/4347

[non-BFB] for all Cryosphere configurations, currently the `CRYO1850` compset, and v1 compsets with `DIB` in the compset name.

### DIFF
--- a/components/mpas-seaice/src/shared/mpas_seaice_forcing.F
+++ b/components/mpas-seaice/src/shared/mpas_seaice_forcing.F
@@ -1797,7 +1797,7 @@ contains
        do iCell = 1, nCellsSolve
 
           bergFreshwaterFlux(iCell) = bergFreshwaterFluxData(iCell)
-          bergLatentHeatFlux(iCell) = bergFreshwaterFluxData(iCell) * &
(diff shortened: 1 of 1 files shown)


darincomeau:
While we're confident this bug fix is correct, putting the Do Not Merge tag on while we conduct testing simulations to determine the impact (on both v1 and v2 simulations).

stephenprice (approved):

darincomeau:
We've run several 50+ year cases in both v1 and v2 configurations and see small but noticeable climate impacts as expected, so I'm removing the Do Not Merge tag.

darincomeau:
@proteanplanet I added you as a reviewer since you found the bug - could you review when you get a chance?

proteanplanet (approved):
So really it's  bergFreshwaterFluxData(iCell) *  (specificHeatFreshIce*bergTemperature - seaiceLatentHeatMelting), which makes sense.    Approved.

jonbob:
merged to next

jonbob:
merged to master

PR #4349, opened 2021-06-18
Add -axCORE-AVX2 and -O3 flags for Intel compiler on Chrysalis

amametjanov:
Add core-avx2 and -O3 flags for Intel compiler on Chrysalis

[BFB] - in actual runs on next

### DIFF
--- a/cime_config/machines/config_compilers.xml
+++ b/cime_config/machines/config_compilers.xml
@@ -835,8 +835,20 @@ flags should be captured within MPAS CMake files.
   <CPPDEFS>
     <append COMP_NAME="gptl">-DHAVE_SLASHPROC</append>
   </CPPDEFS>
+  <CFLAGS>
+    <append>-static-intel</append>
+    <append> -march=core-avx2 </append>
+    <append DEBUG="FALSE"> -O3 </append>
+  </CFLAGS>
+  <CXXFLAGS>
+    <append>-static-intel</append>
+    <append> -axCORE-AVX2 </append>
+    <append DEBUG="FALSE"> -O3 </append>
+  </CXXFLAGS>
   <FFLAGS>
-    <append DEBUG="FALSE">-O2 -debug minimal -qno-opt-dynamic-align</append>
+    <append>-static-intel</append>
+    <append> -march=core-avx2 </append>
+    <append DEBUG="FALSE">-O3 -qno-opt-dynamic-align</append>
   </FFLAGS>
   <PIO_FILESYSTEM_HINTS>gpfs</PIO_FILESYSTEM_HINTS>
   <SLIBS>
@@ -845,12 +857,6 @@ flags should be captured within MPAS CMake files.
   <NETCDF_C_PATH>$ENV{NETCDF_C_PATH}</NETCDF_C_PATH>
   <NETCDF_FORTRAN_PATH>$ENV{NETCDF_FORTRAN_PATH}</NETCDF_FORTRAN_PATH>
   <PNETCDF_PATH>$ENV{PNETCDF_PATH}</PNETCDF_PATH>
-  <CFLAGS>
-    <append>-static-intel</append>
-  </CFLAGS>
-  <FFLAGS>
-    <append>-static-intel</append>
-  </FFLAGS>
   <LDFLAGS>
     <append>-static-intel </append>
   </LDFLAGS>
--- a/cime_config/machines/config_compilers.xml
+++ b/cime_config/machines/config_compilers.xml
@@ -847,7 +847,7 @@ flags should be captured within MPAS CMake files.
   </CXXFLAGS>
   <FFLAGS>
     <append>-static-intel</append>
-    <append> -march=core-avx2 </append>
+    <append> -axCORE-AVX2 </append>
     <append DEBUG="FALSE">-O3 -qno-opt-dynamic-align</append>
   </FFLAGS>
   <PIO_FILESYSTEM_HINTS>gpfs</PIO_FILESYSTEM_HINTS>


amametjanov:
Results of e3sm_integration tests with comparison to master baselines are 2 PASS, 90 DIFF (92 total): https://my.cdash.org/viewTest.php?buildid=2027183 .

Compiler flag changes are non-BFB, but the run-to-run reproducibility still holds: e.g. `REP.ne11_oQU240.A_WCYCL1850.chrysalis_intel`.

Performance contribution based on 1 node x 64 MPI tasks x 1 thread (1x64x1), no-io runs of `PFS.ne11_oQU240.A_WCYCL1850.chrysalis_intel`:
- master: 9.539 sypd
- with `-march=core-avx2`: 10.136 sypd, or 6.25%
- with `-march=core-avx2` and `-O3`: 10.224 sypd or 7.18%

Chrysalis nodes have AMD EPYC 7532 2 socket 32-Core chips with support for avx2.

From Intel's doc (module `intel/20.0.4-kodw73g` with `ifort (IFORT) 19.1.3.304 20200925`):
```
-march=<cpu>
          generate code exclusively for a given <cpu>
            core-avx2  - Generates code for processors that support Intel(R)
                         Advanced Vector Extensions 2 (Intel(R) AVX2), Intel(R)
                         AVX, SSE4.2 SSE4.1, SSE3, SSE2, SSE, and SSSE3
                         instructions
-O3       optimize for maximum speed and enable more aggressive optimizations
          that may not improve performance on some programs
```

golaz:
@amametjanov : these results sound very encouraging. Another 6 or 7% on top of what you have already achieved on chrysalis would be awesome.

To err on the safe side, maybe we should rerun a portion of the Water Cycle spin-up simulation (100 or 200 years) to make sure that there is no discernible climate impact before merging it for v2.

golaz:
I've started a new simulation with the new flags. The simulation should be long enough by next week to verify that the climate is unchanged.

golaz:
Here is a brief summary of what I have found so far in my testing of the new compiler settings:

- Performance increases nicely: v2 prototype on Chrysalis goes from **41** to **43 SYPD** using the L (105) layout.
- Climate appears to be the same (based on 150 years).
- A few PE layouts produce non-BFB results: M (60) and M80 (80). The other layouts I tested (XS, S, L) are BFB.

Tests results are documented on https://acme-climate.atlassian.net/wiki/spaces/EWCG/pages/2939944995/20210625.v2rc3c-compiler.piControl.ne30pg2+EC30to60E2r2.chrysalis

I would hate to leave this performance gain on the table, but the non-BFB behavior is also unsettling.

Tagging @amametjanov, @rljacob, @ambrad

singhbalwinder:
These are interesting experiments. I am assuming these are pure MPI runs, i.e., no threading, is that right?

rljacob:
Were both M and M80 different from each other and different from  XS, S, L ?

golaz:
Yes, three different answers: {XS,S,L}, M and M80:

```
6b7d9cb17ea22ba968913937caba55db  atm_M_1x10_ndays.txt
56f869f0942c86a9398e6ff25755ba88  atm_M80_1x10_ndays.txt
8dc1a514f1c43f71b60399b620f63536  atm_S_1x10_ndays.txt
8dc1a514f1c43f71b60399b620f63536  atm_S_2x5_ndays.txt
8dc1a514f1c43f71b60399b620f63536  atm_XS_1x10_ndays.txt
```

golaz:
M and M80 appear to be identical with each other for the first three time steps and identical to the other ones for the first two time steps. For details:

```/lcrc/group/e3sm/ac.golaz/E3SM_simulations/20210625.v2rc3c-compiler.piControl.ne30pg2_EC30to60E2r2.chrysalis/tests```

**Comment by golaz** [comment]
I also ran a test with the M2 layout overnight. It is BFB with XS, S, and L. So the problematic layouts are M and M80.

**Comment by golaz** [comment]
@rljacob : we need to make a decision now as we are basically ready to start the v2 Water Cycle simulation campaign:

1. Accept the new compiler flags, and remove the M and M80 layouts.
2. Reject the new compiler flags and forget about the additional performance gain.
3. Try an intermediate approach with only `-march=core-avx2` which may provide most of the gains, without the problems from `-O3`?

I think it would be a bad idea to change the compiler flags after the v2 campaign has started.

**Comment by amametjanov** [comment]
Re-posting my notes from Slack:
>Re. the non-BFB of M and M80: these are non-traditional PE-layouts with 3200 and 4160 ATM tasks. XS has 675 ATM tasks, S-1350, M2-2700, L-5400. Nothing else is jumping out as a culprit.
I guess it's a trade-off of having an option of packing onto 60 nodes with M and 80 nodes with M80 and remaining BFB without -O3; or keeping -O3 and not using M and M80 (and other "uneven divide" layouts).

It'd be good to keep BFB (no matter what the PE-layout is). Since most of the speedup was from `-march=core-avx2`, I vote for option 3.

**Comment by golaz** [comment]
Has anyone verified the hypothesis that `-O3` causes the non-BFB behavior? Or do I need to do that myself?

**Comment by amametjanov** [comment]
I'll check

**Comment by rljacob** [comment]
Option 3 is fine once verified.  Else we can delay this PR until after the maint-2.0 branch, assuming the BFB problems are fixed.

**Comment by amametjanov** [comment]
Turns out that non-BFB is caused by `-march=core-avx2` in FFLAGS. Replacing just that with `-axCORE-AVX2` restores BFB across all available PE-layouts for the WC-case *and* we can keep `-O3`.

```
-ax<code1>[,<code2>,...]
          generate code specialized for processors specified by <codes>
          while also generating generic IA-32 instructions.
          <codes> includes one or more of the following:
            ...
            CORE-AVX2
                    May generate Intel(R) Advanced Vector Extensions 2
                    (Intel(R) AVX2), Intel(R) AVX, SSE4.2, SSE4.1, SSSE3, SSE3,
                    SSE2, and SSE instructions for Intel(R) processors.
```

golaz:
Thanks, @amametjanov. I'll try this out today.

golaz:
All good. I get BFB results for all layouts: XS, S, M, M2, M80 and L. This can be merged to master. I'm already using it in the simulation I just started (using cherry pick to get these changes).

Thanks @amametjanov for all your hard work getting the performance up on Chrysalis. I've never imagined we could get up to 43 SYPD. Really awesome work.

amametjanov:
Ok, I'm glad we're back to BFB. But I'm worried that the speedup in long runs won't be as much as before. In my short 20-day, 1-node, no-IO tests with `PFS.ne11_oQU240.A_WCYCL1850.chrysalis_intel`, I'm only getting 2.5% speedup:
- master: 9.500 sypd
- this updated branch: 9.734 sypd

golaz:
I'll let you know what kind of speed-up I observe in my long simulations.

golaz:
@amametjanov and @rljacob : the performance is definitely not as good as it was before. I'm seeing around 41-42 SYPD (see https://acme-climate.atlassian.net/wiki/spaces/EWCG/pages/2967928872/v2.LR.piControl).

I've started the piControl simulation with these settings, so we should merge them and use them for the rest of the v2 Water Cycle campaign.

rljacob:
merged to next.

rljacob:
Testing shows the new flags are BFB.

I looked at a build log to check the flags and found this:
-convert big_endian -assume byterecl -ftz -traceback -assume realloc_lhs -fp-model source -static-intel  -axCORE-     AVX2  -qopenmp  -O2 -debug minimal -O3 -qno-opt-dynamic-align -free"

@amametjanov is that right?  -O2 and -O3 together?

amametjanov:
Yes, -O3 is expected to override -O2.

PR #4350, opened 2021-06-22
Update CIME submodule

jgfouca:
... to 19c689b3a8a554f9892862a323ee140e41a794f0

Changes:
1) Adds --chksum argument to case.submit and create_test allowing
   users to verify checksums without explicitly calling
   check_input_data
2) support the GLC component having multiple grids (e.g., Greenland & Antarctica),
   but the implementation handles any component having multiple grids
3) Record preview_run output in provenance
4) Compells all CIME CCS scripts to use python3!

Fixes:
1) updates the three E3SM climate reproducibility tests (MVK, PGN,
   and TSC) to work with updated component names (eam and elm) and
   time step namelist variables.
2) Fixes unittest raising error on py37, os.path.basename needed to be mocked.
3) Fixes error in parsing and comparing namelist values in the form of "(scale * data path)"
4) Fixes only calling custom_success_msg function when there is a
   return value from run_and_log_case_status. Adds warnings when
   walltime is outside of min/max for the selected queue.
5) Change cdash upload settings:
   Use https instead of http but dont do verification. This seems
   to be the most robust approach.

Previous CIME changes removed official support for python2, but a fair number of
scripts continued to have '#!/usr/bin/env python ' at their top which potentially
allowed python2 to be used depending on the machine's environment. All scripts have
been changed to force python3 to be used. I've also changed all the scripts living
in E3SM to do this as well. This should help us avoid mixing pythons when doing our
work.

[BFB]


jgfouca:
@bishtgautam , @bartgol , I'm not 100% sure I should be changing python stuff in components/elm/tools and components/homme. I can revert these changes if you want.

oksanaguba:
in homme, you changed parselogs, parseE3, check_mass_conservation -- so @mt5555 should know.

bishtgautam:
I don't know if anyone is using `components/elm/tools/shared/PTCLM/` (@dmricciuto, Are you using it?)

I will let @dmricciuto chime in for changes in `components/elm/tools/shared/OLMT/`

jgfouca:
We won't get reports from Sandia machines until this PR is in, so I'm going to start integrating.

PR #4351, opened 2021-06-22
Update mpas-source: Sea ice index allocation bug

akturner:
Fixes bug in sea ice model with index allocation

Updates MPAS with https://github.com/MPAS-Dev/MPAS-Model/pull/878

[BFB]


jonbob:
closed via [PR #4352](https://github.com/E3SM-Project/E3SM/pull/4352), which contained an additional fix

PR #4353, opened 2021-06-23
Updates basin definitions for MOC calculation

vanroekel:
Updates the Atlantic basin to exclude the Mediterranean sea and adds a
new AtlanticMed region that includes it.  The default MOC calculation will
now not include the Mediterranean Sea.  This affects current v2
production meshes only

* EC30to60E2r2
* ECwISC30to60E2r1
* SOwISC12to60E2r4
* WC14to60E2r3

This PR only influences analysis calculations in MPAS-Ocean and should thus be BFB

### DIFF
--- a/components/mpas-ocean/cime_config/buildnml
+++ b/components/mpas-ocean/cime_config/buildnml
@@ -261,7 +261,7 @@ def buildnml(case, caseroot, compname):
         decomp_date = '200904'
         decomp_prefix = 'mpas-o.graph.info.'
         restoring_file = 'sss.PHC2_monthlyClimatology.EC30to60E2r2.200910.nc'
-        analysis_mask_file = 'EC30to60E2r2_moc_masks_and_transects.nc'
+        analysis_mask_file = 'EC30to60E2r2_mocBasinsAndTransects20210623.nc'
         ic_date = '210210'
         ic_prefix = 'ocean.EC30to60E2r2'
         if ocn_ic_mode == 'spunup':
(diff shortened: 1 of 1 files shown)


vanroekel:
@golaz here is the AMOC PR we discussed over email yesterday.

jonbob:
@vanroekel - I'll get it in as soon as it's ready, but will start at least a quick test now

xylar:
@jonbob and @vanroekel, I'm working on the remaining 2 files right now.

xylar:
Okay, all 4 files should be in place now.  Let me know if you have any trouble.

xylar (approved):
I'm reviewing by inspecting.  These are, indeed, the 4 meshes we agree to update the MOC masks for and the new file names are correct.

jonbob:
@vanroekel - I hate to say this, but the branch name does not meet the naming convention. It states:
````
Convention:  E3SM branch names must have the form:
<Github username>/<source code area or component>/<feature-description>
Use lower-case for everything.  Even your username.
Use hyphens instead of underscores.
````

vanroekel:
@jonbob no problem.  Do I need to close the PR and reissue?  Or can I edit without doing that?

jonbob:
I think you can change it in GitHub, but I've never tried. Let me look at one of mine and see...

jonbob:
@vanroekel - it looks easy. Here's a webpage: [renaming-a-branch](https://docs.github.com/en/github/administering-a-repository/managing-branches-in-your-repository/renaming-a-branch)

xylar:
@jonbob, that's super handy!  If we don't have to close PRs and open new ones to rename branches, that would alleviate my concern about this whole naming convention issue.

vanroekel:
when you rename with the trick from @jonbob, the PR does get closed. I had no option to not close it.  But could do it via github

xylar:
Grrrr!  Darn you, GitHub!

PR #4357, opened 2021-06-25
Add mixed layer eddy test cases to MPAS-Ocean init mode

mark-petersen:
Add two mixed layer eddy test cases to MPAS-Ocean init mode. See https://github.com/MPAS-Dev/MPAS-Model/pull/855
- Single temperature front in a periodic channel, with temperature restoring at the boundaries. This is similar to the mixed layer eddy cases in [Fox-Kemper et al., 2008](https://doi.org/10.1175/2007JPO3792.1).
- Double temperature fronts in a doubly periodic domain without temperature restoring. This is similar to the mixed layer eddy case in [Hamlington et al., 2014](https://doi.org/10.1175/JPO-D-13-0139.1) and was used in [Li & Van Roekel, 2021](https://doi.org/10.5194/gmd-14-2011-2021).

This PR only adds to the subdirectory `components/mpas-ocean/src/mode_init`, which is not compiled by E3SM.

[BFB]


mark-petersen:
@qingli411 please test this with your test case at https://github.com/MPAS-Dev/compass/pull/77. In this PR, you can, for example
```
cd components/mpas-ocean
make clean # if needed
make gnu-nersc USE_PIO2=true OPENMP=false GEN_F90=true DEBUG=true 
```
so stand-alone compiles the same way as before. I was able to successfully compile with gnu and intel on grizzly. If you can confirm that your test case works, we can merge this PR, as init mode code is not actually compiled with E3SM.

jonbob (approved):
Approved - none of this should impact E3SM simulations

jonbob:
merged to next

qingli411:
@mark-petersen, I just tested this PR with https://github.com/MPAS-Dev/compass/pull/77 on Cori with GNU. It compiles. The init step runs fine for both "single_front" and "double_front" configurations. But in the forward step I saw the following error for both configurations, although the model does not crash and seem to be running fine and producing the output file. 

```
darshan library warning: unable to create darshan log file //tmp/qingli_ocean_model_id43683094_mmap-log-14471554222139438997-0.darshan
PIO: ERROR: Creating file (output/output.0001-01-01.nc) failed. Internal error. Specified netCDF file does not exist. (error num=-220), (/global/cscratch1/sd/hgkang/my_programs/scorpio/src/clib/pioc_support.c:2588)
```

I'm using the environment for Cori, GNU described in https://mpas-dev.github.io/compass/legacy/machine_specific_instructions/nersc.html.

jonbob:
merged to master

PR #4358, opened 2021-06-25
Adding libstdc++ to libs for linking with PIO2

brobey:
Adding a link to the stdc++ library when using PIO2. Scorpio has some C++ code that requires the addition of the C++ standard library. This fixes the build on Darwin with the PIO2 library for the PGI, GCC and Intel builds (pgi, fortran and ifort) as well as the PGI OpenACC build (pgi-summit), all with OpenMPI. The change has been tested for builds on Darwin and Grizzly.

### DIFF
--- a/components/mpas-framework/Makefile
+++ b/components/mpas-framework/Makefile
@@ -929,6 +929,9 @@ ifdef MPAS_EXTERNAL_CPPFLAGS
 	override CPPFLAGS += $(MPAS_EXTERNAL_CPPFLAGS)
 endif
 ####################################################
+ifeq "$(USE_PIO2)" "true"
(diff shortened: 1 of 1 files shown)


xylar:
Thanks @brobey.  This is a change I'd also like to see because this is what is keeping compass from being able to use the latest SCORPIO.

@jonbob, before this PR gets too far, could you help @brobey get his branch name to match E3SM requirements?

brobey:
Resubmitting with branch name BobRobey/PIO2BuildFix

jonbob:
@brobey - I'll close this with the comment that it's been moved to [PR #4360](https://github.com/E3SM-Project/E3SM/pull/4360)

PR #4361, opened 2021-06-28
Homme: Work around or fix an issue re: threading in standalone Homme.

ambrad:
It was reported that some standalone Homme SL transport tests were crashing with
multiple threads. The failure is an inconsistent shared-pointer count in a
Kokkos array.

This error has not appeared on, e.g., Cori-KNL, despite always running
threaded there. In addition, I'm not seeing a bug in the SL C++ code. But I
do know how to avoid the problem, and this commit does that.

[BFB]

### DIFF
--- a/components/homme/src/share/compose/compose_cedr.cpp
+++ b/components/homme/src/share/compose/compose_cedr.cpp
@@ -24,6 +24,13 @@
 # define ConstExceptGnu const
 #endif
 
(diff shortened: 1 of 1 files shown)


oksanaguba (approved):

ambrad:
e3sm_developer passes against 28 June baselines on chrysalis.

oksanaguba:
in next

PR #4362, opened 2021-06-28
Adding libstdc++ to libs

brobey:
Adding a link to the stdc++ library when using PIO2. Scorpio has some C++ code that requires the addition of the C++ standard library. This fixes the build on Darwin with the PIO2 library for the PGI, GCC and Intel builds (pgi, fortran and ifort) as well as the PGI OpenACC build (pgi-summit), all with OpenMPI. The change has been tested for builds on Darwin and Grizzly.

Currently the builds with the three compilers fail with different, non-informative errors. This fix can also be done by setting the MPAS_EXTERNAL_LIBS variable to -lstdc++. This latter approach can be used to fix older builds. It is unlikely that the mpif90 compiler wrapper would include the C++ standard library, so this fix is likely to be necessary on all platforms. In any case, a duplicate library inclusion should not cause a problem.

[BFB]

### DIFF
--- a/components/mpas-framework/Makefile
+++ b/components/mpas-framework/Makefile
@@ -929,6 +929,9 @@ ifdef MPAS_EXTERNAL_CPPFLAGS
 	override CPPFLAGS += $(MPAS_EXTERNAL_CPPFLAGS)
 endif
 ####################################################
+ifeq "$(USE_PIO2)" "true"
+	override LIBS += -lstdc++
+endif
 
 ifeq "$(CONTINUE)" "true"
 all: mpas_main


xylar (approved):
As documented here https://github.com/MPAS-Dev/compass/pull/171#issuecomment-871387875, I have tested this branch extensively with the compass ocean nightly test suite.  With this change, I have been able to use SCORPIO 1.2.1 (the latest version), whereas without it I needed to use version 1.1.6.

xylar:
@matthewhoffman and @akturner, it would be important for you to make sure this doesn't do any harm for your standalone cores, just in case.

@mark-petersen, maybe you could run the nightly suite or some other tests on LANL IC (with the existing SCORPIO 1.1.6) to make sure you also don't see any problems.

matthewhoffman (approved):
This works for MALI.  Thanks, @brobey

mark-petersen (approved):
This works for me. Passes expected tests in nightly regression suite on grizzly with intel and gnu, both optimized and debug, using conda environments
```
load_dev_compass_1.0.0_grizzly_gnu_mvapich.sh
load_dev_compass_1.0.0_grizzly_intel_impi.sh
```

xylar:
I tested this today on OSX and it didn't work for me.  I needed `-lc++` instead of `-lstdc++` to get things to compile.

Given that this branch has already been approved and is ready to merge, and OSX is not one of our most critical platforms for standalone MPAS work, I will follow up with a fix in a separate PR.

brobey:
Thanks for finding this and working on a fix. Of course OSX has to be different.

From: Xylar Asay-Davis ***@***.***>
Reply-To: E3SM-Project/E3SM ***@***.***>
Date: Monday, July 12, 2021 at 3:51 AM
Cc: Robert Robey ***@***.***>, Mention ***@***.***>
Subject: [EXTERNAL] Re: [E3SM-Project/E3SM] Adding libstdc++ to libs (#4362)


I tested this today on OSX and it didn't work for me. I needed -lc++ instead of -lstdc++ to get things to compile.

Given that this branch has already been approved and is ready to merge, and OSX is not one of our most critical platforms for standalone MPAS work, I will follow up with a fix in a separate PR.

jonbob (approved):
Approved by visual inspection -- it should not impact E3SM at all

jonbob:
merged to next

jonbob:
merged to master. Thanks @brobey -- and please feel free to delete the branch on your fork now, if you wish

xylar:
Thanks @brobey for making this happen!

PR #4369, opened 2021-07-01
Fixes MPAS-Ocean RK4 restart error

### LINKED ISSUE #4376: Ocean RK4 time step fails bfb restart test
In stand-alone mode, MPAS-Ocean fails the restart test when using RK4 time stepping. This does not effect E3SM, which does not use RK4. See original issue at https://github.com/MPAS-Dev/MPAS-Model/issues/777.

vanroekel:
The RK4 time integrator has been failing restart tests.  This fixes that failure which was due to a difference in order of operations in computing the transport velocity when GM was enabled.

E3SM does not use RK4 time stepping on for any tests, so this is BFB for E3SM testing.

[BFB]
fixes #4376

### DIFF
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_rk4.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_rk4.F
@@ -692,12 +692,6 @@ module ocn_time_integration_rk4
          !$omp end do
          !$omp end parallel
 
-         ! Compute normalGMBolusVelocity and the tracer transport velocity
-         if (config_use_GM) then
-             call ocn_gm_compute_Bolus_velocity(statePool, &
-                meshPool, scratchPool, timeLevelIn=2)
-         end if
-
          if (config_use_GM) then
             !$omp parallel
             !$omp do schedule(runtime) 
@@ -836,6 +830,8 @@ module ocn_time_integration_rk4
       real (kind=RKIND), dimension(:, :), pointer :: layerThicknessCur, normalVelocityCur
       real (kind=RKIND), dimension(:, :), pointer ::  normalVelocityProvis, highFreqThicknessProvis
 
+      integer :: iEdge
+      integer, pointer :: nEdges
       logical, pointer :: config_filter_btr_mode
 
       err = 0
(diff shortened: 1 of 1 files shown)


vanroekel:
with this fix the mpas-ocean nightly restart test for RK4 passes on anvil

vanroekel:
@xylar I believe this fixes the RK4 restart issue.  Could you please try when you have a chance?

vanroekel:
@jonbob should we move the bug report issue from the MPAS-Model repo to E3SM?

xylar (approved):
@vanroekel, this is really great!  I verified in compass that the RK4 restart test on Linux with gnu now works as expected: variables from the full run are identical to those from the restart run.

I will also test on Chrysalis with intel and will update here once I get those results.

jonbob:
@vanroekel - it's fine as an MPAS bug report, no need to move it. It may not close the MPAS issue automatically, but I can do that by hand.

xylar:
> It may not close the MPAS issue automatically, but I can do that by hand.

That's right, it won't close the issue automatically because it's in a different repo.

xylar:
> It may not close the MPAS issue automatically, but I can do that by hand.

That's right, it won't close the issue automatically because it's in a different repo.

mark-petersen (approved):
@vanroekel thanks for the correction. Now passes
```
ocean_global_ocean_QU240_PHC_RK4_restart_test
```
on grizzly, IC for both gnu and intel, debug and optimized, using compass and
```
load_dev_compass_1.0.0_grizzly_gnu_mvapich.sh
load_dev_compass_1.0.0_grizzly_intel_impi.sh
```
Optimized gnu and optimized intel pass ALL tests in nightly regression. 

As a side note, debug gnu and debug intel both fail: 
```
  test execution:      ERROR
  see: case_outputs/ocean_ice_shelf_2d_5km_z-star_restart_test.log
  test execution:      ERROR
  see: case_outputs/ocean_ice_shelf_2d_5km_z-level_restart_test.log
```
@xylar should I make a new issue for that?

xylar:
@mark-petersen, yes, please make an issue for that.  Either here or in the compass repo.  Please assign it to both @cbegeman and me.

jonbob:
merged to next

jonbob:
merged to master

PR #4370, opened 2021-07-01
Various YAKL issues improved

mrnorman:
* Update YAKL CMake with new conveniences (YAKL will handle source file flags and other things through a macro now)
* Improve documentation of YAKL and SAMXX integration in CMake files
* Add GPU YAKL_DEBUG
* Improve YAKL header filenames to avoid future name clashes
* Improve YAKL internal macro define names to be more readable
* Change YAKL CMake variables to avoid future clashes with other CMake packages

New wiki documentation on the process of integrating SAM++ and YAKL into E3SM, CIME, and CMake is below:

https://github.com/E3SM-Project/E3SM/wiki/YAKL-and-SAMXX-CIME-and-CMake-Integration-Notes

### DIFF
--- a/components/eam/src/physics/crm/samxx/CMakeLists.txt
+++ b/components/eam/src/physics/crm/samxx/CMakeLists.txt
@@ -1,6 +1,6 @@
 ###############################################################################
 ## This assumes you have already run add_subdirectory(${YAKL_HOME} ${YAKL_BIN})
-## It also assumes you've enabled CXX and C as languages
+## It also assumes you've enabled CXX and C as languages (and CUDA if used)
 ###############################################################################
 
 file(GLOB F90_SRC   cpp_interface_mod.F90 fft.F90 params.F90 fftpack5.F90 fftpack5_1d.F90)
@@ -9,7 +9,7 @@ file(GLOB SAMXX_SRC ${CXX_SRC} ${F90_SRC})
 
 # samxx will be a static library
 add_library(samxx STATIC ${SAMXX_SRC})
-if ("${ARCH}" STREQUAL "CUDA")
+if ("${YAKL_ARCH}" STREQUAL "CUDA")
   # samxx will be CUDA-linked with device symbols resolved at library creation
   set_target_properties(samxx PROPERTIES LINKER_LANGUAGE CUDA CUDA_SEPARABLE_COMPILATION OFF CUDA_RESOLVE_DEVICE_SYMBOLS ON)
 endif()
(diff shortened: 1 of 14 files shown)


mrnorman:
@bartgol , I haven't forgotten about your request for YAKL taking in targets and using generators to set flags on C++ files, but I wanted to get this out fairly quick since I'm sure the ssh submodule issues are annoying people.

mrnorman:
If there are other relatively easy requests to improve annoying things about YAKL and / or its CMake behavior, though, now is a good time to make them.

jgfouca:
@mrnorman , can you explain this change:

> Fix YAKL submodules from ssh to https

If you look at E3SM/.gitmodules, all of E3SM's submodules are set to use ssh. I don't want to get into a situation where we have multiple protocols used to manage submodules since this exposes us to the problems of both.

mrnorman:
image

mrnorman:
I'm happy to change it back. Either way is fine with me.

jgfouca:
@mrnorman , I don't understand Danqing's logic since we already have so many ssh submodules. @rljacob care to comment?

mrnorman:
To be honest, I don't understand his logic fully either. Using SSH, once you have keys on github, they should be valid for any repo you clone IIRC. So if they can clone, say, kokkos with SSH, they should be able to clone a YAKL submodule with SSH too, so long as it's also on github.com.

mrnorman:
When we bring in eigen from gitlab for PAM, however, that might need to be https because it's from a different website where the E3SM user might not have keys installed. But that's a convo for another time.

components/eam/src/physics/crm/samxx/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ file(GLOB SAMXX_SRC ${CXX_SRC} ${F90_SRC})
 
 # samxx will be a static library
 add_library(samxx STATIC ${SAMXX_SRC})
-if ("${ARCH}" STREQUAL "CUDA")
+if ("${YAKL_ARCH}" STREQUAL "CUDA")
```
bartgol, line 12: I'm not sure why you need the YAKL_ARCH var, and not just do 
```
if (USE_CUDA)
```

bartgol (approved):
Looks good. I only had a comment, but I don't think it is really important.

components/eam/src/physics/crm/samxx/CMakeLists.txt
mrnorman: I'll change this to the simpler logic as you suggested. In other projects, I'm used to using YAKL_ARCH because I don't have a USE_CUDA variable defined like we do in E3SM.

bartgol:
Re: ssh vs https. I think https can actually be more of a hassle for certain systems. E.g., at SNL we had issues with https, due to proxy and certificate configurations. Every fix we tried to use seemed to be like a blanket that is too small, and always leaves some part uncovered.

ssh requires ssh keys, but that's fairly standard, and way easier to fix if not working...imho.

components/eam/src/physics/crm/samxx/CMakeLists.txt
bartgol: Makes sense. As I said, not a big deal. Though less cmake vars to  track is always better. :)

mrnorman:
OK, YAKL submodules are back to SSH, and the SAM++ standalone tests are updated to work with the new YAKL, they run and pass on summit on CPU and GPU.

jgfouca (approved):

components/eam/src/physics/crm/samxx/CMakeLists.txt
mrnorman: Should it be `if (USE_CUDA)` or if `(${USE_CUDA})` ? Or does it matter?

rljacob:
Yes as mentioned by @bartgol and also in the IG call today, https presents some problems for proxy servers so we have to use ssh.  note @dqwu

components/eam/src/physics/crm/samxx/CMakeLists.txt
bartgol: If you use recent enough cmake (I think >3.8 maybe? which should be the case everywhere by now...) the former is preferable. The latter would cause a warning, telling you to do `if (VAR_NAME)` rather than `if( ${VAR_NAME} )`.

whannah1:
@mrnorman @brhillman does this PR still need work or is it ready to go?

mrnorman:
To the best of my knowledge, it's good to go.

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #4374, opened 2021-07-02
Reduce MPAS output

amametjanov:
Turn off netcdf output for
- mpassi.hist.am.regionalStatistics
- mpassi.hist.am.timeSeriesStatsDaily
- mpaso.hist.am.globalStats
- mpaso.hist.am.highFrequencyOutput

[BFB]

### DIFF
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -566,7 +566,11 @@
 <config_use_MacroMoleculesTracers_sea_ice_coupling>.false.</config_use_MacroMoleculesTracers_sea_ice_coupling>
 
 <!-- AM_globalStats -->
-<config_AM_globalStats_enable>.false.</config_AM_globalStats_enable>
+<config_AM_globalStats_enable>.true.</config_AM_globalStats_enable>
+<config_AM_globalStats_enable ocn_grid="oQU480">.false.</config_AM_globalStats_enable>
+<config_AM_globalStats_enable ocn_grid="oQU240">.false.</config_AM_globalStats_enable>
+<config_AM_globalStats_enable ocn_grid="oQU240wLI">.false.</config_AM_globalStats_enable>
+<config_AM_globalStats_enable ocn_grid="oQU120">.false.</config_AM_globalStats_enable>
 <config_AM_globalStats_compute_interval>'output_interval'</config_AM_globalStats_compute_interval>
 <config_AM_globalStats_compute_on_startup>.true.</config_AM_globalStats_compute_on_startup>
 <config_AM_globalStats_write_on_startup>.false.</config_AM_globalStats_write_on_startup>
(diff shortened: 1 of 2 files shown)


amametjanov:
When running benchmarks, it's useful to not have any IO: PFS tests help with that. However, by default, these netcdf files are still written out: e.g.
```
azamat@chrlogin1:run$ lr *.nc
-rw-r--r-- 1 azamat E3SM 6166764 Jul  1 20:20 PFS_Ld1_P64x1.ne11_oQU240.A_WCYCL1850.chrysalis_intel.20210701_200412_c7qcg0.mpassi.hist.0001-01-01_00000.nc
-rw-r--r-- 1 azamat E3SM   33696 Jul  1 20:29 PFS_Ld1_P64x1.ne11_oQU240.A_WCYCL1850.chrysalis_intel.20210701_200412_c7qcg0.mpassi.hist.am.regionalStatistics.0001.01.nc
-rw-r--r-- 1 azamat E3SM 4751984 Jul  1 20:29 PFS_Ld1_P64x1.ne11_oQU240.A_WCYCL1850.chrysalis_intel.20210701_200412_c7qcg0.mpassi.hist.am.timeSeriesStatsDaily.0001-01-01.nc
-rw-r--r-- 1 azamat E3SM  122680 Jul  1 20:29 PFS_Ld1_P64x1.ne11_oQU240.A_WCYCL1850.chrysalis_intel.20210701_200412_c7qcg0.mpaso.hist.am.globalStats.0001-01-01.nc
-rw-r--r-- 1 azamat E3SM 1992992 Jul  1 20:29 PFS_Ld1_P64x1.ne11_oQU240.A_WCYCL1850.chrysalis_intel.20210701_200412_c7qcg0.mpaso.hist.am.highFrequencyOutput.0001-01-01_00.00.00.nc
```

Would it be okay to turn these off by default?

If so, this PR turns off some IO, but it'd be good to turn off `mpassi.hist` output too:
```
azamat@chrlogin1:run$ lr *.nc
-rw-r--r-- 1 azamat E3SM 26212 Jul  1 20:46 PFS_P64x1.ne11_oQU240.A_WCYCL1850.chrysalis_intel.20210701_203652_tvmf7f.mpassi.hist.0001-01-01_00000.nc
```
Please feel free to push and modify this branch as necessary.

xylar:
@amametjanov, are these analysis members turned on somewhere for production runs?  If so, I couldn't find where.  If not, it would be extremely important that this PR include the necessary changes to make sure they are turned on during production runs.

To me, it seems safer to turn them off with user namelist options for testing and do leave them on by default, but I am eager to hear what other MPAS developers think.

milenaveneziani:
These am's seem to be on in production runs (I have checked two recent production runs and they are enabled).

We could *probably* have `mpaso.hist.am.highFrequencyOutput` and `mpassi.hist.am.timeSeriesStatsDaily` off by default, but my preference is to turn them off intentionally during testing as @xylar said.

xylar:
> These am's seem to be on in production runs (I have checked two recent production runs and they are enabled).

Thanks @milenaveneziani.  My concern would be if this PR would turn them off in production runs.

vanroekel:
This pr appears to change the ams to default off for all runs, which i don't think is. a good idea.  I'd much prefer something like
```
<config_AM_highFrequencyOutput_enable>.false.</config_AM_highFrequencyOutput_enable>
<config_AM_highFrequencyOutput_enable ocn_grid="oQU240">.false.</config_AM_highFrequencyOutput_enable>
```
This would only disable for QU240.  It would be even better if we could do this for a certain config, but I'm not sure how to do that

jonbob:
I agree with @vanroekel -- this is not a good change to make by default. I think we could make a testmod that does this instead, and just use that when we want minimal I/O for testing. @amametjanov - would that work for you? Maybe it could be extended to minimize I/O for other parts of the system as well, like turning off restarts?

jonbob:
@amametjanov - how would you like to proceed with this? I think changing the defaults is probably unlikely, but it would be pretty straightforward to make a testmod that would implement reduced I/O

amametjanov:
Pushed an update to turn off AM output by default for QU grids. One remaining item is to turn off `*.mpassi.hist.0001-01-01_00000.nc` by default too. Could you recommend something?

xylar:
@amametjanov, do you also want to turn off `timeSeriesStatsMonthly` in MPAS-Ocean and MPAS-Seaice?  Or are those written out infrequently enough to not cause you trouble?

xylar:
Hmm, I've given this a bit more thought and I'm concerned that we use `QU*` tests to determine if the code is bit-for-bit.  It seems like a very big problem for those tests if a lot of output that is standard in our production runs is turned off.  I don't know a huge amount about what output is compared in E3SM tests but it seems like output from analysis members *should* at least be included in bit-for-bit comparisons because it would be a big problem in production runs if the analysis is broken.

@vanroekel, @milenaveneziani, @jonbob and others, what is your feeling about this?  Rather than turning output off by default, could we (and should we) have two sets of QU tests, one that is stripped down and explicitly turn off this output and another that make sure the output is part of the the bit-for-bit comparison?  

In general, it seems kind of problematic to me to have the `QU*` configurations drift too far from the production configurations.

vanroekel:
I agree with @xylar that we shouldn't let all tests drift too far from production.  I can think of more than one case when a QU test didn't pick up problems that were in production configs.  However, I also think we do have standard testing of production configurations (at least for water cycle).  But if it isn't too hard it may be quite useful to have a set of QU_* tests that are light weight for performance and such and another that mirror production runs as @xylar suggests.

amametjanov:
Thanks for the comments. Closing.

PR #4377, opened 2021-07-02
Add namelist variable for MMF CRM orientation

whannah1:
This replaces the MMF_DIR_NS and MMF_ORIENT_RAND preprocessor flags with a namelist variable "MMF_orientation_angle", specified in degrees. The angle defaults to 90 for 2D MMF cases and 0 for 3D cases. The random rotation mode can be enabled with MMF_orientation_angle=-1. Changes are non-BFB because the value of pi was changed to use the physconst value instead of a locally defined value with arbitrary precision.

[BFB] except for MMF

### DIFF
--- a/components/eam/cime_config/config_component.xml
+++ b/components/eam/cime_config/config_component.xml
@@ -61,7 +61,6 @@
       <value compset="_EAM%MMF[12]"           >-crm sam</value>
       <value compset="_EAM%MMFXX"             >-crm samxx</value>
       <value compset="_EAM%MMFOMP"            >-crm samomp</value>
-      <value compset="_EAM%MMF"               >-cppdefs ' -DMMF_DIR_NS ' </value>
       <value compset="_EAM%MMF"               >-rad rrtmgp </value>
       <value compset="_EAM%MMF[1XO]"          >-MMF_microphysics_scheme sam1mom -chem none</value>
       <value compset="_EAM%MMF2"              >-MMF_microphysics_scheme m2005 </value>
--- a/components/eam/src/physics/crm/crm_physics.F90
+++ b/components/eam/src/physics/crm/crm_physics.F90
@@ -560,8 +560,7 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
    real(crm_rknd) :: norm_rand                     ! normally distributed random number - Box-Muller (1958)
    real(crm_rknd) :: crm_rotation_std              ! scaling factor for rotation (std dev of rotation angle)
    real(crm_rknd) :: crm_rotation_offset           ! offset to specify preferred rotation direction 
-   real(crm_rknd), dimension(pcols) :: crm_angle   ! local copy of CRM orientation angle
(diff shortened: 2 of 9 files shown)


components/eam/cime_config/config_component.xml
```diff
@@ -60,7 +60,6 @@
       <!-- they will be removed in a future PR -->
       <value compset="_EAM%SP[12]V1"        >-use_MMF -crm_adv MPDATA -nlev 72 -crm_nz 58 -crm_dx 1000 -crm_dt 5 </value>
       <value compset="_EAM%SP[12]V1"        >-microphys mg2  </value>
-      <value compset="_EAM%SP[12]V1"        >-cppdefs ' -DMMF_DIR_NS ' </value>
```
lee1046, line 63: Can you remind me why we are still keeping  EAM%SP compsets? No one seems to be using them.

whannah1: It was for some sense of backwards compatibility, but I agree with you sentiment that they aren't useful anymore. I'll make a note to make another PR to remove them.

components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -673,50 +661,41 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
...
 
-      if (.not. is_first_step()) then
-         ! get current crm angle from pbuf, except on first step
-         call pbuf_get_field(pbuf, pbuf_get_index('CRM_ANGLE'), crm_angle)
+      ! Rotate the CRM using a random walk
+      if ( (crm_ny.eq.1) .or. (crm_nx.eq.1) ) then
+         if (.not. is_first_step()) then
+            ! get current crm angle from pbuf, except on first step
+            call pbuf_get_field(pbuf, crm_angle_idx, crm_angle_ptr)
+            crm_angle(:pcols) = crm_angle_ptr(:pcols)
```
lee1046, line 674: I don't think it is wrong but I am just curious why you decided to use a pointer, crm_angle_ptr, instead of just using crm_angle inside pbuf_get_field.

oksanaguba:
I saw this PR by accident, I have a general question about new options that we put into ATM:
would it be good to have the options printed in atm.log*? we cannot really do this for the existing options, but we could try to do this for new PRs.

whannah1:
@oksanaguba I don't have a strong opinion either way since I never check the log files for namelist parameters. However, I could see it being useful for debugging situations where you think you're setting a namelist variable and some bug is causing a different value to be used.

components/eam/src/physics/crm/crm_physics.F90
whannah1: Yea, I guess I can do it like that. I had done it this way for another branch because passing pointers to the CRM is problematic given how it's been written for the GPU, but in this case the variable is only used locally. I'll switch to just using a pointer to make the code a bit simpler.

lee1046 (approved):
Changes are all straightforward and easy to understand. It is good to go.

brhillman:
Merged to next. Note that this PR adds a new namelist variable, so we see NLFAILS, but all cases are BFB except MMF.

brhillman:
Merged to master.

PR #4378, opened 2021-07-03
Update Intel compiler and MPI modules on Anvil

### LINKED ISSUE #4391: siac buildnml and python 3.7.6
While doing some hand testing on anvil, I loaded python 3.7.6 with:
`module load anaconda3/2020.02`

then tried to create a simple F-case:
`--compset FC5AV1C-L --res ne4_ne4`
and got an error curing case.setup:

```
   Running /gpfs/fs1/home/jacob/E3SM2/components/stub_comps/siac/cime_config/buildnml 
/usr/bin/env: python3: No such file or directory
ERROR: /gpfs/fs1/home/jacob/E3SM2/components/stub_comps/siac/cime_config/buildnml /gpfs/fs1/home/jacob/E3SM2/cime/scripts/Fcasekok FAILED, see above
Exit 1
```

Turns out siac buildnml only had this code:
```
#!/usr/bin/env python3

"""
build stub model namelist
"""
# DO NOTHING
```

while the other stub model's buildnml's had:
```
#!/usr/bin/env python3

"""
build stub model namelist
"""
# DO NOTHING
#pylint: disable=unused-argument
def buildnml(case, caseroot, compname):
    pass
```

BUT:  python 3.8.3, also available on anvil, did NOT have any error with the incomplete buildnml.

Why doesn't anyone else see this on another platfrom?   All cases run the siac buildnml.

amametjanov:
Update Intel compiler and MPI modules on Anvil:
- intel 19.1.3.304 20200925 (same as Chrysalis)
- intel-mpi 2019.9.304

Also update GMAKE_J value for singularity machine.
Fix the siac buildnml to look like other stub buildnml.

Fixes #4391 
[non-BFB] - for e3sm_prod baselines on anvil

### DIFF
--- a/cime_config/machines/config_compilers.xml
+++ b/cime_config/machines/config_compilers.xml
@@ -751,6 +751,9 @@ flags should be captured within MPAS CMake files.
   <LDFLAGS>
     <append compile_threaded="TRUE"> -static-intel</append>
   </LDFLAGS>
+  <MPICC  MPILIB="impi"> mpiicc  </MPICC>
+  <MPICXX MPILIB="impi"> mpiicpc </MPICXX>
+  <MPIFC  MPILIB="impi"> mpiifort </MPIFC>
   <PIO_FILESYSTEM_HINTS>gpfs </PIO_FILESYSTEM_HINTS>
   <SLIBS>
     <append> $SHELL{$ENV{NETCDF_FORTRAN_PATH}/bin/nf-config --flibs} </append>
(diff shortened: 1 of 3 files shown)


rljacob:
Does this fix the kokkos build?

amametjanov:
Yes. But it appears there is an issue with the latest mvapich version 2.3.6. Runs with intel-mpi pass at 100%, runs with mvapich are at 78 out of 92 test pass rate. I asked to back off to a previous mvapich version.

rljacob:
Is there a performance difference between intel mpi and mvapich on anvil?

amametjanov:
Yes, mvapich has always been fastest in the past on Anvil (Infiniband network). Modules were rebuilt on one of the compute nodes, which have proper Mellanox driver versions. Testing...

amametjanov:
Backing off to mvapich 2.3.4 did not help. Trying out OpenMPI 4.1.1 (same as Chrysalis), but it will take a while. Making Intel-MPI the default now, until we get a working and faster alternative.

Intel-MPI is running faster now.
20210707-mvapich-vs-impi

rljacob:
merged to next

PR #4380, opened 2021-07-05
Update sandia HPC PROJECT, AKA WCID

jgfouca:
[BFB]


jgfouca:
I've confirmed the old WCID is now being rejected. The new one is not working either although the HPC admins approved it, so I think it's just a matter of time until it starts working. I'm going to go ahead and merge this since the old one no longer works anyway.

PR #4384, opened 2021-07-07
Remove LANDM_COSLAT and replace uses with LANDFRAC

### LINKED ISSUE #4276: remove landm_coslat array
during modernization of the topo data set tool chain, I was looking at the landm_coslat() array.  

This array is in all our EAM topo input files, but isn't currently used.

documented places where it appears in the code:

It is read during initialization (in phys_inidat()
It is stored as a global array in the comsrf module
It is passed as an argument in tphysbc()

It is used in stratiform_tend(), called when the 'RK' microphysics scheme is turned on (not used in E3SM)
(actually use in in cldwat.F90::pcond() called by stratiform)_tend)

It appears in the RRTMG interface.  RRTMG will write it to history files, but otherwise doesn't appear to use it.  RRTMGP doesn't use it at all.

restart_physics.F90: written to restart file

Suggestion: Lets remove this array, so we dont have to compute it in the topo tool chain, dont have to read it during init, and dont have to read/write it to restart files.

brhillman:
Remove the LANDM variable from the COMSRF module and replace all uses
with cam_in%landfrac. LANDM was previously read in from the LANDM_COSLAT
variable in the topography dataset. This was a smoothed version of the
input topography, but was actually inconsistent with the land fraction
data specified by the surface models. Instead of carrying around two
versions of land fraction, we opt to remove the LANDM variable and
instead use a land fraction consistent with the surface models. This
will be BFB for all currently supported configurations *except* for MMF,
where we were still using the old cloud effective radius calculations
with LANDM instead of LANDFRAC. Since these are switched over to use
LANDFRAC now, the results will be non-BFB but should still be
reasonable. Also, any cases using RK microphysics will DIFF because
`stratiform_tend` was also using LANDM, buried down in the findmcnew
routine, and also in the effective radii computation. The code was
changed here but *not tested* for this configuration because we do not
have a supported configuration for that microphysics package at the
moment. 

Fixes #4276

[non-BFB] only for MMF cases.

### DIFF
--- a/components/eam/src/dynamics/se/dyn_comp.F90
+++ b/components/eam/src/dynamics/se/dyn_comp.F90
@@ -225,7 +225,7 @@ CONTAINS
     use time_mod,         only: time_at
     use control_mod,      only: runtype
     use cam_control_mod,  only: aqua_planet, ideal_phys, adiabatic
-    use comsrf,           only: landm, sgh, sgh30
+    use comsrf,           only: sgh, sgh30
     use cam_instance,     only: inst_index
     use element_ops,      only: set_thermostate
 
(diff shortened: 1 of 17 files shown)


whannah1 (approved):
Looks good.

mt5555 (approved):
Looks great. thanks for removing all the related code from the topo tools as well.

PR #4388, opened 2021-07-08
Fix desp buildnml

jasonb5:
`buildnml` for the `desp` component was using a path that was
locally constructed rather than using CIME config files. This fixes
`buildnml` to use the `NAMELIST_DEFINITION_FILE` variable
to get the path, this is inline with other components.


jasonb5:
This requires CIME PR https://github.com/ESMCI/cime/pull/4028.

jgfouca:
@jasonb5 , I'll do a CIME update soon.

PR #4389, opened 2021-07-08
Update CIME submodule

jgfouca:
to 0067c2dc3629dab249ef0069b0fce58004443c50

Changes:
1) Remove tons of CIME files for items that got moved into different repos
   A lot of this is stuff we used to share with CESM but no longer do.
2) Adds default walltime for queues
3) Add support for multi GPUs per node
4) Changes the inputdata structure to add a field "inventory"
   currently this field is only used for NEON tower data
5) Give an error message when no default namelist value can be found
6) Updates the use of imp to importlib.
7) Additional support for ww3 as a component in the e3sm config file
8) allow a test to have multiple testmods, separated by two hyphens in a row
  – for example, foo-testmod1--bar-testmod2--baz-testmod3
9) Update python version check to 3.6 or later

Fixes:
1) Fixes to allow a variable to be specified in user_nl but not in defaults
2) Set default python to 'python3' for scripts tests CMake file
3) allow multiple copies of max_tasks_per_node
4) Adds missing value for desp component
5) Fix external mct location
6) Fix bugs in climate reproducibility tests (MVK and PGN)

[BFB]


jasonb5:
Theres another CIME update that will need to be included to fix mct shared, I'll be pushing to CIME shortly. This fix requires that a new submodule is added to E3SM to externals/mct.

jgfouca:
@jasonb5 , thanks. I noticed every case is failing with 
```
can't copy 'E3SM/cime/src/externals/mct/mct/Makefile': doesn't exist or not a regular file
```

jasonb5:
Here's the issue https://github.com/ESMCI/cime/issues/4018 should have the PR soon.

rljacob:
I'll get MCT in to master today and then you can rebase this.

rljacob:
Master now has externals/mct.

jgfouca:
This branch looks good on mappy. Merging to next.

dqwu:
@jgfouca 
It seems that 3 cime_developer tests failed on anlworkstation with this cime submodule update in E3SM next branch. Could you please take a look?
https://my.cdash.org/viewTest.php?onlyfailed&buildid=2037963

PR #4390, opened 2021-07-08
Add MCT as a submodule

rljacob:
Add MCT as a submodules in E3SM/externals and use latest version 2.11.0
This will be used instead of the source in cime/src/externals which is
being removed.  Not active until a future PR makes changes to the CIME build.

[BFB]

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -49,3 +49,6 @@
 [submodule "components/mpas-ocean/src/gotm"]
 	path = components/mpas-ocean/src/gotm
 	url = git@github.com:E3SM-Project/GOTMcode.git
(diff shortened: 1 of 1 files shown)


rljacob:
This external isn't being used yet so this PR has no impact on the code.

jgfouca (approved):

jgfouca:
@rljacob , what's the difference between this submodule and the driver-mct subdirectory we already have?

rljacob:
driver-mct is the source for the top level driver that uses MCT datatypes/method.  externals/mct is the source for those datatypes/methods.

PR #4397, opened 2021-07-12
Ocean SI code - Fix bugs, add OpenMP directives, SI flag change

hyungyukang:
This PR fixes bugs on the updated SI codes, added OpenMP directives for the SI solver, and changed the flag of the semi-implicit barotropic mode solver (`config_time_intesgrator`) from `semi_implicit` to `split_implicit`.


hyungyukang:
Opened PR in wrong way. Close this PR.

PR #4399, opened 2021-07-13
Add CIME HTML dir on Anvil for atm non-bfb tests

mkstratos:
This adds the CIME_URL_ROOT and CIME_HTML_ROOT for Anvil, which is used by CIME in the non-bit for bit tests to make evv4esm results available on a public facing webpage.

This change places the results in the test user's sub-directory as a top-level sub-directory in the E3SM project (e.g. https://web.lcrc.anl.gov/public/e3sm/e3smtest for the "e3smtest" user).


rljacob:
@sarich start merging this.

PR #4401, opened 2021-07-13
Fix incorrectly named iceberg term in totalFreshWaterTemperatureFlux

jonbob:
Replaces an incorrectly named iceberg term in the calculation of totalFreshWaterTemperatureFlux in mpas-ocean. This term was causing unrealistic heating in the southern hemisphere for cryo configurations with data icebergs on.

[non-BFB] for cryo configurations with DIB on

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_surface_bulk_forcing.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_surface_bulk_forcing.F
@@ -562,7 +562,7 @@ contains
                                          inLandIceCavity=.false.) / rho_sw
 
            totalFreshWaterTemperatureFlux(iCell) = rainTemperatureFlux(iCell) + evapTemperatureFlux(iCell) + &
-                                                   seaIceFreshWaterTemperatureFlux(iCell) + icebergFreshWaterFlux(iCell)
+                                                   seaIceFreshWaterTemperatureFlux(iCell) + icebergFreshWaterTemperatureFlux(iCell)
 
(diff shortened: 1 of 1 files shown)


darincomeau:
Thanks @jonbob - I've got a test running now with this and https://github.com/E3SM-Project/E3SM/pull/4348. Should have 10 years by tomorrow.

jonbob:
Thanks so much @darincomeau - how long does it take to see the problematic heating?

darincomeau:
10 years should be enough to see that problematic Southern Ocean heating.

darincomeau:
Test run (this fix and #4348).
image

Previous run with only #4348
image

WCYCL1850 run
image

Note atmospheric tunings are slightly different in last two runs (rc3e) than current test run (v2.LR.piControl). But this certainly looks like the smoking gun we've been looking for that's been causing the problematic Southern Ocean SST heating in Cryo configurations.

darincomeau (approved):
Approving based on inspection and 10 year test that shows significant improvement in Southern Ocean SST summer warm bias in Cryosphere configurations (with DIB).

xylar:
That's great!

jonbob:
Whew -- big relief! I'll get this merged as quickly as possible

xylar (approved):
I'm approving this by visual inspection -- the bug and fix is very clear -- and based on Darin's successful testing and very promising results

jonbob:
merged to next

maltrud (approved):
sorry for tardiness.  all looks good fixing my bug.

jonbob:
merged to master

PR #4404, opened 2021-07-14
Cori needs the MCT host setting

jgfouca:
Setting it to cray seems to work.


jgfouca:
Skipping integration.

PR #4405, opened 2021-07-14
Update Sandiatoss3 gcc compiler

jgfouca:
We don't use GCC directly, but we do use its stdlibc++. Kokkos
now requires a newer stdlibc++, so we have to load a newer gcc
module.

[BFB]


jgfouca:
Did a bit of testing on chama and it looks OK. Skipping integration.

PR #4408, opened 2021-07-15
For Cori, change the default layout for any case using only 1 node to use full node

### LINKED ISSUE #4406: Set a default number of MPI's to use for 1-node jobs on Cori
For cori-knl, use 64 MPI's as the default for any 1-node job and use 32 MPI's for cori-haswell.

The `ERS.MOS_USRDAT.RMOSGPCC.cori-knl_intel.mosart-mos_usrdat` test keeps failing on cori-knl as it runs out of time.
However, if I use 64 MPI's on 1 node, it completes. It seems like the default for these machines should be number of typical cores used per node (and 1 thread).

Testing e3sm_developer on cori-knl, all tests passed, however, I see that this change means there are several (about 8) other tests that will now use 64 MPI's. I will also test on cori-haswell.

branch: `ndk/machinefiles/cori-1node-default-layout2`

ndkeen:
For Cori, change the default layout for any case using only 1 node use a full node --

for cori-knl, use 64 MPI's
for cori-haswell, use 32 MPI's

Fixes #4406
bfb

### DIFF
--- a/cime_config/allactive/config_pesall.xml
+++ b/cime_config/allactive/config_pesall.xml
@@ -73,6 +73,76 @@
         </rootpe>
       </pes>
     </mach>
+    <mach name="cori-knl">
+      <pes compset="any" pesize="any">
+        <comment>1 node default</comment>
+        <ntasks>
+          <ntasks_atm>64</ntasks_atm>
+          <ntasks_lnd>64</ntasks_lnd>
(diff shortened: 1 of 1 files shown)


ndkeen:
Note this also involved adding a partition file for a test on cori-haswell that uses 32 MPI's.

ndkeen:
Merged to next

PR #4409, opened 2021-07-15
Only send ocean CO2 flux to cpl if atm is expecting it

### LINKED ISSUE #4394: MPAS-Ocean tries to send the atmosphere the CO2 flux even if atm doesn't expect it
in ocn_comp_mct.F, the ocean tries to send the co2 flux through the coupler to the atmosphere even if atm doesn't expect it, such as in a G-case with marine BGC turned on.  1 line fix simply checks to see if index_o2x_Faoo_fco2_ocn is not zero before loading the flux into the o2x_o % rAttr array.  Otherwise, memory corruption occurs.

maltrud:
Make sure that index_o2x_Faoo_fco2_ocn has a non-zero value before loading the co2 surface flux for export.  Memory corruption can occur without this fix. 

Fixes #4394

[BFB]

### DIFF
--- a/components/mpas-ocean/driver/ocn_comp_mct.F
+++ b/components/mpas-ocean/driver/ocn_comp_mct.F
@@ -2479,7 +2479,7 @@ contains
        end if
 
        ! BGC fields
-       if (config_use_ecosysTracers) then
+       if (config_use_ecosysTracers .and. index_o2x_Faoo_fco2_ocn /= 0) then
           ! convert from mmolC/m2/s to kg CO2/m2/s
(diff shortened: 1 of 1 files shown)


jonbob (approved):
Approved, based on visual inspection and previous testing

jonbob:
@maltrud - just fyi, the branch naming convention is to use dashes and not underscores, plus all lowercase. No big deal this time, but I do need to point that out

jonbob:
merged to next

jonbob:
merged to master

PR #4412, opened 2021-07-16
Adding lwrad_off treatment for RRTMGP

lee1046:
lwrad_off option was missing in RRTMGP. This capability is required for some SCM cases.

### DIFF
--- a/components/eam/src/physics/crm/rrtmgp/radiation.F90
+++ b/components/eam/src/physics/crm/rrtmgp/radiation.F90
@@ -14,7 +14,7 @@ module radiation
    use shr_kind_mod,     only: r8=>shr_kind_r8, cl=>shr_kind_cl
    use ppgrid,           only: pcols, pver, pverp, begchunk, endchunk
    use cam_abortutils,   only: endrun
-   use scamMod,          only: scm_crm_mode, single_column, swrad_off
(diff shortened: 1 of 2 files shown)


components/eam/src/physics/crm/rrtmgp/radiation.F90
```diff
@@ -2172,6 +2172,13 @@ subroutine calculate_fluxes_lw(gas_names, gas_vmr, emis_sfc, &
       ncol = size(pmid,1)
       nlev = size(pmid,2)
 
+      ! When lwrad_off is set to true, zero fluxes and return
+      if (lwrad_off) then
+         call reset_fluxes(fluxes_allsky)
+         call reset_fluxes(fluxes_clrsky)
+         return
+      end if
```
whannah1, line 2180: Don't we need to set qrl and qrlc to zero here like in the other file?

lee1046: qrl and qrlc are computed from fluxes inside calculate_heating_rate(), which are called right after calling the subroutine calculate_fluxes_lw. Since all fluxes are reset to zero inside calculate_fluxes_lw(), qrl and qrlc become zero as well in the case of lwrad_off = True

PR #4416, opened 2021-07-17
Fix Homme standalone build after Kokkos update.

ambrad:
We need to request C++14 support instead of C++11.

This should fix the sandiatoss3 standalone Homme nightlies.

[BFB] for CIME tests; probably BFB for Homme standalone tests


oksanaguba:
I am going to merge this and let nightlies run tests.

jgfouca (approved):

oksanaguba:
@rljacob those will be blessed.

rljacob:
Its much better then before when it was a build fail for all tests!

PR #4417, opened 2021-07-20
Ensure sync is called before freedecomp

### LINKED ISSUE #4415: maint-1.0 fatal error with Scorpio
Cases using maint-1.0 got fatal error when using Scorpio, PIO_VERSION=2 (default since May 10 after PR #4193):

`PIO: FATAL ERROR: Aborting... An error occured, Syncing file (ncid=131) failed. Invalid file id. Unable to find internal structure associated with the file id. err=-33. Aborting since the error handler was set to PIO_INTERNAL_ERROR... (/gpfs/fs1/home/ac.wlin/E3SM/integration/E3SM.maint-1.0/externals/scorpio/src/clib/pio_file.c: 233)`

This problem exists at least on compy and chrysalis.
The runs would be ok if using PIO_VERSION=1. But the problem seems not cause by PR #4193 which updated scorpio. Using the commit immediately before  #4193 also has the same behavior: Scorpio classic (PIO_VERSION=1) ok, but Scorpio (PIO_VERSION=2) not. 

There is also a seemingly unexplained behavior. The problem does not occur to short runs ( a few steps or a few days), although they also set to write history and restart files. The tests so far show the fatal run occur just before finishing the 1st month. That is probably why the issue was not captured by the nightly tests for maint-1.0.  Further examination suggests the problem likely occur when trying to write cice.h, which by default only at monthly boundary.

The problem can be reproduced using --compset FC5AV1C-L and --res ne30_oECv3, with all MPI on 4 or 16 nodes.

jayeshkrishna:
Freeing decompositions in CICE history write routine
before the file close.

The corresponding change in E3SM master was added in
PR #3607

Fixes #4415

[BFB]

### DIFF
--- a/components/cice/src/drivers/cpl/ice_history_write.F90
+++ b/components/cice/src/drivers/cpl/ice_history_write.F90
@@ -621,6 +621,12 @@
          endif
       enddo ! num_avail_hist_fields
 
+      ! -------------------------
+      ! clean-up PIO descriptors
+      ! -------------------------
+      call pio_freedecomp(File,iodesc2d)
+      call pio_freedecomp(File,iodesc3d)
+
       !-----------------------------------------------------------------
       ! close output dataset
       !-----------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


wlin7 (approved):
The code changes are good. A test was also done to confirm this fixed the issue (#4415) while maintaining BFB.

dqwu (approved):
Looks good to me.

jayeshkrishna:
Freeing decompositions in CICE history write routine
before the file close.

The corresponding change in E3SM master was added in
PR #3607

Fixes #4415

[BFB]

PR #4421, opened 2021-07-21
Update the LC machine config files to newer cmake version

AaronDonahue:
This commit updates the cmake version used by the two Livermore
Computing machines quartz and syrah.

The recent update to Kokkos requires a cmake version of 3.16 or
higher.

[BFB]


rljacob:
@AaronDonahue please merge this old PR.

PR #4425, opened 2021-07-22
Another CIME submodule update

jgfouca:
to 231c8e783d9df1896f7ac0bdd1f31e74f496bdac

Changes:
1) Fails if xmllint not found in path.
2) jenkins_generic_job: no reason for this magic config setting
 Just have baselines off by default. It's not hard to turn them on

Fixes:
1) Ensure Cmake is available for wait_for_tests

[BFB]


jgfouca:
This is needed to restore cdash reports from blues/anvil.

jgfouca:
Nothing controversial in this, so I'm going to begin integrating.

jgfouca:
Merged to next.

PR #4426, opened 2021-07-22
Fmyuan/bugfix budg flux g

fmyuan:
The ELM 'CNPBudgetMod' has bug relevant to 'all_time' flux integral ('p_inf' in the code). It will crash model when it runs into end of first year, if continuing from a pre-stage restart file. A simple test is to run ELM in about 25 years of accelerated_spinup (ad_spinup) first, then run normal spinup from it, model would crash when first year ends. 

A suggested change to this same modules is to modify its error checking tolerance. It should be consistent with what criteria used in EcosystemBalanceCheckMod. For unknown reason, that grid-level budget errors appear accumulated with time, until some points of over 0.01 error-tolerance which crashing model. A single point tests show this crashing might occur around 200 years of spinup run.

### DIFF
--- a/components/elm/src/biogeochem/CNPBudgetMod.F90
+++ b/components/elm/src/biogeochem/CNPBudgetMod.F90
@@ -857,7 +857,7 @@ contains
     ! !LOCAL VARIABLES:
     integer :: f, s, s_beg, s_end ! data array indicies
     real(r8) :: time_integrated_flux, state_net_change
-    real(r8), parameter :: error_tol = 0.01_r8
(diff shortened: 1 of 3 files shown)


rljacob:
@bishtgautam please look at this.

bishtgautam:
@hydrotian also found a similar issue recently and I'm looking into it. I have opened https://github.com/E3SM-Project/E3SM/issues/4510 to track this issue.

cc: @susburrows @jenniferholm

rljacob:
Looking for a better fix.

PR #4430, opened 2021-07-27
Remove swrad_off and lwrad_off flags

whannah1:
This removes the swrad_off and lwrad_off flags for disabling radiation in SCM simualtions in favor of using iradsw=0 and iradlw=0 for disabling radiation. 

[BFB]

### DIFF
--- a/components/eam/src/physics/crm/rrtmg/radiation.F90
+++ b/components/eam/src/physics/crm/rrtmg/radiation.F90
@@ -34,8 +34,7 @@ use cam_abortutils,  only: endrun
 use error_messages,  only: handle_err
 use cam_control_mod, only: lambm0, obliqr, mvelpp, eccen
 use scamMod,         only: scm_crm_mode, single_column,have_cld,cldobs,&
-                           have_clwp,clwpobs,have_tg,tground,swrad_off,&
-                           lwrad_off
+                           have_clwp,clwpobs,have_tg,tground
 use perf_mod,        only: t_startf, t_stopf
 use cam_logfile,     only: iulog
 
@@ -1330,14 +1329,6 @@ end function radiation_nextsw_cday
     calday = get_curr_calday()
     call zenith (calday, clat, clon, coszrs, ncol, dt_avg)
     
-    ! We can bypass the shortwave calculation by setting the cosine of the solar
-    ! zenith angle to zero for all columns, because the shortwave code collapses
-    ! the inputs to daytime-only arrays. In case the swrad_off flag is set then,
-    ! we force shortwave to not be calculated by setting coszrs = 0.
-    if (swrad_off) then
-      coszrs(:)=0._r8
-    endif    
-
     ! Output cosine solar zenith angle
     call outfld('COSZRS', coszrs(1:ncol), ncol, lchnk)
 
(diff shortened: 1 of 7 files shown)


components/eam/src/physics/crm/rrtmg/radiation.F90
```diff
@@ -364,18 +367,17 @@ real(r8) function radiation_nextsw_cday()
...
    offset = 0
-   do while (.not. dosw)
-      nstep = nstep + 1
-      offset = offset + dtime
-      if (radiation_do('sw', nstep)) then
-         radiation_nextsw_cday = get_curr_calday(offset=offset) 
-         dosw = .true.
-      end if
-   end do
-   if(radiation_nextsw_cday == -1._r8) then
```
lee1046, line 375: Don't you want to keep this warning print statement?

components/eam/src/physics/crm/rrtmg/radiation.F90
```diff
@@ -2106,21 +2100,6 @@ subroutine radiation_tend( state, ptend,pbuf, &
                          clm_seed,     lu,           ld                                            )
                 call t_stopf ('rad_rrtmg_lw')
 
-                if (lwrad_off) then
```
lee1046, line 2109: just make sure variables enclosed in if(lwrad_off) block is initialized to zero elsewhere.

bogensch (approved):
Thanks for doing this, Walter, and FYI, I have just committed a change to the SCM library repo so that SCM scripts will be up to date with this modification.

components/eam/src/physics/crm/rrtmg/radiation.F90
whannah1: No, Ben and I both studied the logic here and couldn't see any good reason for keeping it.

components/eam/src/physics/crm/rrtmg/radiation.F90
whannah1: I'm pretty sure we don't actually need to worry about these when we are not running radiation. The only one I'm not sure about is the cam_out variable, but I'm not sure there's any situation where we would be running an active surface component without any radiation calculations. I'm gonna omit zeroing these out for now and we can come back and fix it if it breaks one of the SCM cases.

lee1046 (approved):
All looks good to me.

whannah1:
e3sm_atm_developer tests passed baseline comparison.
Running a coupled test now.

components/eam/src/physics/crm/rrtmg/radiation.F90
oksanaguba: If these vars are not needed, it seems , then would it be better to remove them from declarations? Same for cam_out%flwds . if you remove their init, I assume running with debug (which probably has -init=snan ) would help to see if these are used.

whannah1: The variables are definitely used in typical use cases for diagnostic output, but the only cases that require disabling radiation are the handful of SCM cases that have the radiative tendencies included in the forcing. Ben and I had a long discussion about this and we agreed that if the variables are not already properly initialized then that is outside the scope of this PR. 

Additionally, we're not sure whether the step to zero out all the variables is actually necessary or not, and I'm not exactly sure how to test it because I'm not sure any of the SCM cases require the combination of an active surface component and disabled radiation, which seems to be the only way it would be problematic. 

So I feel that it's better to remove this now and then implement a proper initialization block later if a problem surfaces.

whannah1:
Coupled test passed.
ERS_Ld3.ne30pg2_EC30to60E2r2.WCYCL1850.cori-knl_intel

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #4432, opened 2021-07-29
change default MMF time step

whannah1:
Change the default atmosphere time step of 20 min (ATM_NCPL=72) to apply to all MMF cases on any grid. 

[BFB]


brhillman:
Merged to next.

brhillman:
Merged to master.

PR #4433, opened 2021-07-29
Add NERSC HTTPS server with E3SM container for CircleCI test

lukaszlacinski:
Add NERSC HTTPS server with E3SM container for CircleCI test and use Petrel HTTPS as a backup

Based on the latest experience with the power outage at Argonne National Lab, this PR adds the NERSC HTTPS server as the main server that keeps the E3SM Singularity container. The CircleCI test script will fall back to the ANL ALCF Petrel HTTPS server if wget fails to download the container from the NERSC HTTPS server. The download time from the NERSC server seems to be 3 times smaller than from the ANL Petrel HTTPS server.

Also remove old travis CI config files.


rljacob (approved):

rljacob:
What do you mean by "The NERSC server seems to be 3 times smaller than from the ANL Petrel HTTPS server." ?

lukaszlacinski:
> What do you mean by "The NERSC server seems to be 3 times smaller than from the ANL Petrel HTTPS server." ?

The download time from the NERSC server is about 0.5 minutes. From the Petrel, it is about 1.5 minutes.

PR #4439, opened 2021-08-03
read data river nutrients in MOSART, pass to MPASO

maltrud:
This PR allows river outflow nutrient loads calculated by GNEWS to be read into MOSART, then passed through the coupler to MPAS-Ocean to be used as a surface nutrient flux when ocean BGC is turned on.  The nutrient inputs to MOSART are concentrations (kg-Nutrient/kg-water) that vary with location, but are constant in time. These concentrations are multiplied by the calculated model runoff to create a nutrient flux to the ocean.  Note that this currently only works with MOSART, not with data runoff.  Also note that this is not enabled by any currently defined compset so one needs to be an expert to use this capability for now.

[NML] due to new cpl flag
[BFB]


components/mosart/src/cpl/rof_comp_mct.F90
```diff
@@ -708,6 +732,21 @@ subroutine rof_export_mct( r2x_r )
...
+               r2x_r%rAttr(index_r2x_Forr_rofDOP,ni) =  tmp1*rtmCTL%nutrDOP(n)
+               r2x_r%rAttr(index_r2x_Forr_rofDOC,ni) =  tmp1*rtmCTL%nutrDOC(n)
+               r2x_r%rAttr(index_r2x_Forr_rofPP ,ni) =  tmp1*rtmCTL%nutrPP(n)
+               r2x_r%rAttr(index_r2x_Forr_rofDSi,ni) =  tmp1*rtmCTL%nutrDSi(n)
+               r2x_r%rAttr(index_r2x_Forr_rofPOC,ni) =  tmp1*rtmCTL%nutrPOC(n)
+               r2x_r%rAttr(index_r2x_Forr_rofPN ,ni) =  tmp1*rtmCTL%nutrPN(n)
+               r2x_r%rAttr(index_r2x_Forr_rofDIC,ni) =  tmp1*rtmCTL%nutrDIC(n)
+               r2x_r%rAttr(index_r2x_Forr_rofFe,ni)  =  tmp1*rtmCTL%nutrFe(n)
+             end if
           endif
```
liho745, line 750: Same comments on the naming of nutrDIN, nutrDIP etc.

components/mosart/src/riverroute/RunoffMod.F90
```diff
@@ -77,6 +77,17 @@ module RunoffMod
...
+     real(r8), pointer :: nutrDIP(:)       ! Nutrient flux for DIP
+     real(r8), pointer :: nutrDON(:)       ! Nutrient flux for DON
+     real(r8), pointer :: nutrDOP(:)       ! Nutrient flux for DOP
+     real(r8), pointer :: nutrDOC(:)       ! Nutrient flux for DOC
+     real(r8), pointer :: nutrPP(:)        ! Nutrient flux for PP
+     real(r8), pointer :: nutrDSi(:)       ! Nutrient flux for DSi
+     real(r8), pointer :: nutrPOC(:)       ! Nutrient flux for POC
+     real(r8), pointer :: nutrPN(:)        ! Nutrient flux for PN
+     real(r8), pointer :: nutrDIC(:)       ! Nutrient flux for DIC
+     real(r8), pointer :: nutrFe(:)        ! Nutrient flux for Fe
```
liho745, line 91: The units of these variables should be added.

liho745:
If I understand correctly, what from GNEWS are the nutrient concentration values that are spatially distributed at different river outlets but temporarily constant. These concentration values are multiplied with the MOSART simulated streamflow values that are spatiotemporally varying, hence providing the spatiotemporally varying nutrient fluxes to the ocean grids via the flux coupler. If this is correct, perhaps the constant concentration part can be reflected in the PR description.

components/mosart/src/cpl/rof_comp_mct.F90
```diff
@@ -693,6 +701,22 @@ subroutine rof_export_mct( r2x_r )
...
+               r2x_r%rAttr(index_r2x_Forr_rofDOC,ni) =  tmp1*rtmCTL%nutrDOC(n)
+               r2x_r%rAttr(index_r2x_Forr_rofPP ,ni) =  tmp1*rtmCTL%nutrPP(n)
+               r2x_r%rAttr(index_r2x_Forr_rofDSi,ni) =  tmp1*rtmCTL%nutrDSi(n)
+               r2x_r%rAttr(index_r2x_Forr_rofPOC,ni) =  tmp1*rtmCTL%nutrPOC(n)
+               r2x_r%rAttr(index_r2x_Forr_rofPN ,ni) =  tmp1*rtmCTL%nutrPN(n)
+               r2x_r%rAttr(index_r2x_Forr_rofDIC,ni) =  tmp1*rtmCTL%nutrDIC(n)
+               r2x_r%rAttr(index_r2x_Forr_rofFe,ni)  =  tmp1*rtmCTL%nutrFe(n)
+             end if
+
           endif
```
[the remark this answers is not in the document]
maltrud: @liho745 , that's a good idea.  i'll change to conc*

components/mosart/src/riverroute/RunoffMod.F90
maltrud: @liho745 , i'll add units.

maltrud:
@liho745 , i added more detail to the PR description as you suggested.

liho745 (approved):

jonbob:
@maltrud - I have just one quick comment. Does it make sense to add a cpl namelist flag so that these fields only get defined in seq_flds_mod when we're using this capability, sort of like the ocn/ice BGC fields don't get defined unless flds_bgc_oi is true? Otherwise all these new fields will be passed around for all configurations. If it doesn't make sense to be part of this PR, should we make an issue and take it on soon?

hydrotian (approved):

jonbob:
@maltrud - with my last commit, I think we avoid creating and passing the extra cpl fields when this capability is not being used. I'm happy to sign off on this PR and get it merged if it works for you when you do throw this cpl flag

maltrud:
@jonbob , yes it worked fine when i manually turned on the cpl flag.  so i think we're ready.

jonbob:
OK, running e3sm_developer on a test merge on chrysalis

jonbob:
Passed e3sm_developer on chrysalis, except all tests show the NML DIFF.

jonbob (approved):
Passed e3sm_developer on chrysalis

jonbob:
merged to next

jonbob:

Issue #4420, opened 2021-07-21
Mods needed for XLF on Summit


amametjanov:
When building with IBM compiler on Summit, some errors come up that can be fixed in pending/existing PRs.
```
$ git checkout 6d75c33b0d9b0900f8312f01cc81c83cb6470e86
$ ./cime/scripts/create_test SMS.ne4_ne4.FC5AV1C-L.summit_ibm.eam-cosplite
```

Error 1:
```
"/autofs/nccs-svm1_home1/azamat/repos/E3SM/components/elm/src/main/timeinfoMod.F90", line 18.15: 1516-044 (S) A conversion from type INTEGER is not permitted.
** timeinfomod   === End of Compilation 1 ===
1501-511  Compilation failed for file timeinfoMod.F90.
Target CMakeFiles/lnd.dir/__/__/elm/src/main/timeinfoMod.F90.o built in 0.103351 seconds
gmake[2]: *** [cmake/lnd/CMakeFiles/lnd.dir/__/__/elm/src/main/timeinfoMod.F90.o] Error 1
```
Possible fix:
```
diff --git a/components/elm/src/main/timeinfoMod.F90 b/components/elm/src/main/timeinfoMod.F90
index 96d48631e9..9eab687a3c 100644
--- a/components/elm/src/main/timeinfoMod.F90
+++ b/components/elm/src/main/timeinfoMod.F90
@@ -15,7 +15,7 @@ module timeinfoMod
   integer  :: jday_mod  = 1            ! day into year?
   real(r8) :: thiscalday_mod  = 1.0_r8  ! day number including hours
   real(r8) :: nextsw_cday_mod = 1.0_r8 !nextsw_cday = mod((nstep/(86400._r8/dtime))*1.0_r8,365._r8)+1._r8
-  logical  :: end_cd_mod = 0           ! end of current day
+  logical  :: end_cd_mod = .false.     ! end of current day
   logical  :: doalb = .false.
   !$acc declare copyin(dtime_mod,dayspyr_mod, year_curr, year_prev, &
   !$acc mon_curr, mon_prev, day_curr, day_prev, secs_curr, secs_prev, nstep_mod, &
```

Error 2:
```
"/autofs/nccs-svm1_home1/azamat/repos/E3SM/components/elm/src/biogeochem/CNPBudgetMod.F90", line 145.35: 1516-036 (S) Entity p_f_name has undefined type.
[...]

PR #4441, opened 2021-08-05
Update Summit and Ascent machine config

amametjanov:
Update Summit and Ascent machine config:
- add custom Perl install on Ascent
- update CIME syntax for SMT batching on Ascent and Summit
- update Summit's python to 3.7.0
- enable C++14 kokkos builds with IBM compilers on Summit
- add CXXFLAGS for threading on summit+ibm
- add -qnosmp to NOOPT_FILES for summit+ibm
- add a 1-node default pelayout
- replace a goto with a loop exit in COSP2 submodule

Fixes E3SM-Project/E3SM#4420

[NML] - due default pelayout changes

### DIFF
--- a/cime_config/allactive/config_pesall.xml
+++ b/cime_config/allactive/config_pesall.xml
@@ -5,16 +5,16 @@
   <grid name="any">
     <mach name="any">
       <pes compset="any" pesize="any">
-        <comment>none</comment>
+        <comment>any grid, any compset, any machine, 1 node</comment>
         <ntasks>
-          <ntasks_atm>16</ntasks_atm>
-          <ntasks_lnd>16</ntasks_lnd>
-          <ntasks_rof>16</ntasks_rof>
-          <ntasks_ice>16</ntasks_ice>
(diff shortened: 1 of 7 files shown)


rljacob:
Can we just do something like this for Theta @sarich ?

dqwu:
@amametjanov 
If the gcc module for summit is upgraded to a higher version in the future, the path /sw/summit/gcc/8.1.1/lib64 also needs to be updated, right?

amametjanov:
Not necessarily (unless they delete `/sw/summit/gcc/8.1.1`).
The latest `xl/16.1.1-10` is still configured with gcc-4.8.5.
Loading of `gcc/8.1.1` unloads any `xl` module.
So we have to be more explicit in linking with newer gcc.

dqwu:
@amametjanov 
parallel-netcdf/1.12.1 and parallel-netcdf/1.12.2 are available on Summit now.
Could you please consider upgrading parallel-netcdf/1.8.1 (out of date) to one of them in this PR? Thanks.

dqwu (approved):

PR #4452, opened 2021-08-09
Add surface pressure to grad ssh

mark-petersen:
This PR adds a the contribution of surface pressure to the (grad ssh) term. 

This does not affect the default E3SM pressure gradient configuration (`config_pressure_gradient_type = 'Jacobian_from_TS'`), which is for layered models, and already starts with the surface pressure when computing pressure down the column. 

The grad ssh option (`config_pressure_gradient_type = 'ssh_gradient'`) is used for single-layer tide simulations for ICOM. It was missing the surface pressure, which was required to run with land ice cavities, but will also be needed to include atmospheric and sea ice pressure contributions. No new name-list options are added.

[BFB]

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_vel_tidal_potential.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_vel_tidal_potential.F
@@ -65,7 +65,8 @@ module ocn_vel_tidal_potential
    !*** physical constants
    real (kind=RKIND) ::  &
       betaSelfAttrLoad,  &! self-attraction and loading beta
-      tidalPotRamp        ! scale for ramping tidal potential
+      tidalPotRamp, &     ! scale for ramping tidal potential
+      rho0gInv            ! 1 / rho0 / gravity
 
    !*** eventually these will be private module arrays and 
    !*** not in the forcing pool so retain pointers here as placeholders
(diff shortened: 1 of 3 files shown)


mark-petersen:
Passes sanity tests
```
```

components/mpas-ocean/src/shared/mpas_ocn_vel_tidal_potential.F
```diff
@@ -169,9 +170,10 @@ subroutine ocn_vel_tidal_potential_tend(ssh, tend, err)!{{{
          kMax = maxLevelEdgeTop(iEdge)
 
          potentialGrad = - gravity*invdcEdge* &
-                         (tidalPotEta(cell2) - tidalPotEta(cell1) + &
-                          betaSelfAttrLoad* &
-                          (ssh(cell2) - ssh(cell1)))
+            (  tidalPotEta(cell2) - tidalPotEta(cell1) &
+             + betaSelfAttrLoad &
+              *((ssh(cell2) - ssh(cell1)) &
+                + rho0gInv*(surfacePressure(cell2) - surfacePressure(cell1))))
```
mark-petersen, line 176: @nairita87 I altered the spacing here and added `rho0gInv`. Please double check this.

nairita87 (approved):
Below is the plot of the global kinetic energy (rms) with and without ice shelf cavities on an EC60to30 mesh. The plot shows that there is no appreciable change in the global kinetic energy due to the inclusion of ice shelf cavities
with_and_without_comparison

The tidal analysis shows that there are reduction of complex errors near the Antarctic ice shelf cavities
standard_m2amp_ec60to30_original_params
standard_m2amp_ec60to30wisc_original_params

nairita87:
The terms in this PR are needed for single layer tidal simulations, in order for ice shelves to be included, otherwise there is a large kinetic energy and the simulation dies

jonbob (approved):
Looks good from a visual inspection

jonbob:
merged to next and passed sanity testing

jonbob:
merged to master

PR #4456, opened 2021-08-11
Fix MMF damping and variance transport test mod

whannah1:
This fixes two loosely related issues that cropped up in MMF tests using intel with variance transport enabled. One issue has to do with a bug in the C++ damping routine that only affects tests with intel. The other change has to do with updating the VT test mod to use the namelist variable for the filtered variance transport mode. 

[BFB]

### DIFF
--- a/components/eam/cime_config/testdefs/testmods_dirs/eam/mmf_use_VT/shell_commands
+++ b/components/eam/cime_config/testdefs/testmods_dirs/eam/mmf_use_VT/shell_commands
@@ -1,1 +1,1 @@
-./xmlchange --append -id CAM_CONFIG_OPTS -val " -use_MMF_VT -cppdefs ' -DMMF_DIR_NS -DMMF_VT_KMAX=4 ' "
+./xmlchange --append -id CAM_CONFIG_OPTS -val " -use_MMF_VT "
--- a/components/eam/cime_config/testdefs/testmods_dirs/eam/mmf_use_VT/user_nl_eam
+++ b/components/eam/cime_config/testdefs/testmods_dirs/eam/mmf_use_VT/user_nl_eam
@@ -1,1 +1,1 @@
-MMF_VT_wn_max = 4
+mmf_vt_wn_max = 4
--- a/components/eam/cime_config/testdefs/testmods_dirs/eam/mmf_use_VT/user_nl_eam
+++ b/components/eam/cime_config/testdefs/testmods_dirs/eam/mmf_use_VT/user_nl_eam
@@ -1,0 +1,1 @@
+MMF_VT_wn_max = 4
--- a/components/eam/src/physics/crm/samxx/damping.cpp
+++ b/components/eam/src/physics/crm/samxx/damping.cpp
@@ -19,13 +19,14 @@ void damping() {
 
   real constexpr tau_min    = 60.0;
   real constexpr tau_max    = 450.0;
-  real constexpr damp_depth = 0.4;
+  real constexpr fractional_damp_depth = 0.4;
 
-  int1d  n_damp("n_damp",ncrms);
-  real2d t0loc ("t0loc" ,nzm,ncrms);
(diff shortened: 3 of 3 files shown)


whannah1:
Just verified that this branch is actually BFB running this test on Chrysalis with baseline comparison:
ERS_Ln9_P96x2.ne4pg2_ne4pg2.F-MMFXX.chrysalis_intel

whannah1:
This PR is another example where the rules about what qualifies as "BFB" are hazy. The changes will fix the failing test we've been seeing on Cori and Theta, so there will be a change in the dashboard. So we are just fixing a problem that only occurs on certain machines with certain a compiler version. Seems like that should qualify as BFB.

mrnorman:
Walter originally uncovered what appears to be an Intel compiler bug. To reproduce on OLCF's Andes cluster:

```C++
// walter_bug.cpp
#include "YAKL.h"
int main() {
  yakl::init();
  { 
    typedef double real; 
    int  constexpr nzm = 28;
    int  constexpr nz  = nzm+1; 
    real constexpr damp_depth = 0.4;
    int ncrms = 200;
    typedef yakl::Array<real,2,yakl::memDevice,yakl::styleC> real2d;
    typedef yakl::Array<int ,1,yakl::memDevice,yakl::styleC> int1d;
    real2d z     ( "z" , nz , ncrms );
    int1d  n_damp( "n_damp" , ncrms );
    yakl::c::parallel_for( ncrms , YAKL_LAMBDA (int icrm) {
      for (int k=nzm-1; k>=0; k--) { 
        if(z(nzm-1,icrm)-z(k,icrm) < damp_depth*z(nzm-1,icrm)) {
          n_damp(icrm)=nzm-1-k+1;
        }
      } 
    });
  }
  yakl::finalize();
}
```

```bash
# Log into andes.olcf.ornl.gov
module load intel
# icc (ICC) 19.0.3.199 20190206
git clone git@github.com:mrnorman/YAKL.git
cd YAKL
git submodule update --init
cd ..
icc -O2 -g -I YAKL -I YAKL/gptl YAKL/YAKL.cpp walter_bug.cpp && ./a.out
```

The code works with `-O1` but fails with `-O2` and higher with a segmentation fault on Intel. It works with `g++` and `clang++` with `-Ofast`.

mrnorman:
@whannah1 , you can work around this bug by replacing `damp_depth` with the literal `0.4`. So, can you comment out the if statement and then below it put the following?
```C++
// Working around an intel bug that cannot handle damp_depth below
if(z(nzm-1,icrm)-z(k,icrm) < 0.4*z(nzm-1,icrm)) {
```

mrnorman (approved):

mrnorman:
@xyuan , can you find the appropriate avenue for reporting this bug to Intel? I figure it should probably come from Argonne since Argonne has closer ties to Intel. Please let me know if you cannot reproduce the error or the workaround at ALCF or if you have any questions about it. Thanks!

whannah1:
I tested Matt's proposed alternative fix on Cori and I still got the error. Interestingly, I only got the error when using 1 OMP thread, whereas a case with 2 threads ran fine. The current fix in this PR is working on Cori for 1 and 2 threads, so I assume it will also work on theta. Matt agreed we should go ahead with the current fix.

mrnorman:
If we plan to use ALCF GPUs in the future, we need to get Intel on this seeming bug as soon as we can.

brhillman:
Merged to next.

brhillman:
Merged to master.

whannah1:
For future reference, the non-BFB result for any MMF test with the use_VT test mod is expected because when I initially created the test mod I used a CPP flag to enabled the "filtered" mode for the CRM variance transport, but I forgot that I had switched control of this mode to a namelist variable. In other words, the use_VT mod was turning on the "bulk" variance transport when we actually wanted the "filtered" variance transport to exercise the FFT portion of the VT code. This PR includes a simple fix for that, so the use_VT tests will diff.

PR #4460, opened 2021-08-12
[WIP] 2D spherical transport tests for mpas ocean

pbosler:
# Summary

OceanNGD task ft2.

This PR accompanies [MPAS-Dev/compass](https://github.com/MPAS-Dev/compass) PR 210 to implement the standard spherical 2D transport test cases from [Lauritzen, Skamarock, Prather, & Taylor, GMD 2012](https://gmd.copernicus.org/articles/5/887/2012) in MPAS ocean.

## New module

The new module `src/shared/mpas_ocn_transport_tests.F` contains the main contributions.  It's in `shared` because its methods are required by both the `init` and `forward` modes.  This module contains the initial condition functions for 3 tracers, which are written into the MPAS debugTracers arrays, and the definitions of the time-varying prescribed velocity fields for each of 4 new test cases.  It provides the basic interface for tracer transport tests that will (in future work) be used for 3D tracer transport tests.

## Hooks to time integration

Using prescribed velocity requires that we overwrite the MPAS computed velocity with the velocity field associated with each test.    This functionality is now in the forward mode RK4 and split_explicit integrators.   At each time step a subroutine checks if the configuration namelist defines a spherical transport test and, if so, it writes the appropriate velocity data.   If not, the normal time stepping procedure continues.

### DIFF
--- a/components/mpas-ocean/src/driver/Makefile
+++ b/components/mpas-ocean/src/driver/Makefile
@@ -10,11 +10,11 @@ OBJS = mpas_ocn_core.o \
 
 all: core_ocean
 
-core_ocean: $(OBJS) 
+core_ocean: $(OBJS)
 
 mpas_ocn_core.o:
 
-mpas_ocn_core_interface.o: mpas_ocn_core.o
+mpas_ocn_core_interface.o: mpas_ocn_core.o ../inc/structs_and_variables.inc
 
 clean:
 	$(RM) *.o *.mod *.f90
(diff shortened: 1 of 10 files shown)


vanroekel:
@pbosler I wanted to mention before the work on this PR progresses too far that the branch name for your PR does not follow e3sm naming conventions.  Since this requires closing a PR, I wanted to mention this before too long.  The convention is 
```
githubUserName/component/short-name-describing-branch
```
so something like
```
pbosler/ocn/2d-transport-tests
```
sorry for the inconvenience.

PR #4464, opened 2021-08-13
Update RRTMGP and YAKL

brhillman:
Update RRTMGP to point to our (E3SM) fork, which includes the C++/YAKL port we need for SCREAM and MMF. This will be non-BFB for the RRTMGP test in `e3sm_atm_integration` and for the MMF tests (that all use RRTMGP) only because of a floating point literal change in RRTMGP that was made to make the Fortran and C++ codes BFB with one another. The code is otherwise BFB (i.e., a test in which this single change was implemented in master is BFB with this branch). Also brings YAKL up to latest version.

[non-BFB] (only for RRTMGP and MMF tests)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -14,8 +14,8 @@
 	branch = e3sm/kokkos
 [submodule "rrtmgp"]
 	path = components/eam/src/physics/rrtmgp/external
-	url = git@github.com:RobertPincus/rte-rrtmgp.git
-	branch = develop
+	url = git@github.com:E3SM-Project/rte-rrtmgp.git
+	branch = master
 [submodule "externals/scorpio"]
 	path = externals/scorpio
 	url = git@github.com:E3SM-Project/scorpio.git


rljacob:
Who is going to be in charge of maintaining the fork of RRTMGP (keeping it up to date with what we would have used in the original repo.) ?

brhillman:
@rljacob I have mostly been responsible for maintaining the RRTMGP fork, with some help from @mrnorman. However, this has mostly been maintaining the RRTMGP C++ code (since this is the only version we are planning on using for MMF and for the SCREAM codebase going forward), and then pushing some fixes upstream to the original RRTMGP repo that Robert Pincus maintains (primarily to maintain BFB between the Fortran and C++ codes). We have not been pulling bleeding edge development from the original RRTMGP repo into E3SM, as nobody has requested integration of new features.

whannah1 (approved):

brhillman:
Merged to next.

PR #4472, opened 2021-08-18
Add in-line self attraction and loading for global tidal simulation

sbrus89:
Self attraction and loading (SAL) forcing involves a spherical harmonic transform (SHT) of the SSH [(Ray, 1998)]( https://doi.org/10.1080/01490419809388134). The spherical harmonic coefficients are scaled by the load Love numbers, then an inverse transformation is performed to compute the SAL field. The SAL forcing is incorporated in the pressure gradient term along with the tidal potential.

Two different approaches are implemented for performing the SHT:
1. SSH is gathered onto a single MPI rank and interpolated to a Gaussian mesh. A very fast library, SHTns, is used to compute the forward/inverse SHT. The SAL field is then interpolated back to the unstructured mesh and scattered back to the other MPI ranks. This work was done by @knbarton.
2. The SHT integration is performed on each subdomain and the SH coefficients are globally summed with an all-reduce. 

Approach 1 ) performs better at low core-counts, but 2) scales better for larger runs.

This PR includes an init mode option for verification testing of the forward/inverse SHT.

This work was done as a part of the tides task of the ICoM project.

[NML]
[BFB] (for all standard E3SM tests, and for tests with SAL on when `config_parallel_self_attraction_loading_bfb = .true.`)


sbrus89:
Here are some verification results for both SHT algorithms:

Approximation errors with different SH orders are shown for two different analytical functions. The first is on a quasi uniform resolution mesh with resolutions: QU60 (165k cells), QU30 (659k cells), and QU15 (2.6M cells).

spiral_with_background

The second analytical function is on a  "equatorial refined" mesh with resolutions: ER120to60 (60k cells), ER12to30 (118k cells), and ER12to15 (286k cells). The oscillations vary according to mesh resolution such that there are approximately 30 grid cells per period.

EC120to15_with_background

Convergence results for the QU meshes are:

QU_convergence
The convergence of the serial approach is slower due to interpolation errors. Convergence for the parallel approach is limited by discretization error due to mid-point rule integration. 


And convergence results for the ER meshes are:

EC_convergence

Looking across equivalent accuracy levels, the performance of the two SHT algorithms for the QU meshes are:

QU60
QU60_scaling

QU30
QU30_scaling

QU15
QU15_scaling


The performance comparisons for the ER  meshes are:

ER120to60
EC120to60_scaling

ER120to30
EC120to30_scaling

ER120to15
EC120to15_scaling

All runs were done on Compy with Intel compilers.

knbarton:
I performed harmonic analysis comparison to benchmark TPXO8 tidal data from a single-layer barotropic run on Icosahedron 9 (16km) uniform mesh. This shows the serial version of SAL in this pull request gives the same tidal results as the serial version built previously. I have also attached a comparison with the scalar SAL approximation to show the improvement due to full SAL calculation. 

This was build on Cori with these settings : `make gnu-nersc USE_PIO2=true OPENMP=false DEBUG=false GEN_F90=true`

------
**M2 Constituent**

M2_plot
M2_plot
M2_plot

M2 Complex RMSE Values in Deep water (<1000m) between 66N and 66S :
Serial before : 0.06399980070877255m
Serial currently: 0.06399973949098071m
Scalar approximation: 0.072925093765029m

**K1 Constituent**

K1 Complex RMSE Values in Deep water (<1000m) between 66N and 66S :
Serial before : 0.02360560164062904m
Serial currently: 0.023605630922693552m 
Scalar approximation: 0.02605653002810445m

**N2 Constituent**

N2 Complex RMSE Values in Deep water (<1000m) between 66N and 66S :
Serial before : 0.01316337401656041m
Serial currently: 0.013163374004691027m
Scalar approximation: 0.014364836226592457m

**O1 Constituent**

O1 Complex RMSE Values in Deep water (<1000m) between 66N and 66S :
Serial before : 0.014989959013047288m
Serial currently: 0.014989994509038505m
Scalar approximation: 0.021686930435006916m

**S2 Constituent**

S2 Complex RMSE Values in Deep water (<1000m) between 66N and 66S :
Serial before : 0.02929243376361691m
Serial currently: 0.02929240084208756m
Scalar approximation: 0.034033753955931854m

------

We do expect a slight difference in results between the previous serial SAL calculation and the one in this PR due to changing from N = Nlat - 1 to N = Nlat/2 - 1.

Currently, there is an error when using the parallel SAL. When that is resolved, I will share those harmonic analysis results as well.

knbarton:
Here are the results for the parallel SAL implementation on Icosahedron 9 mesh using the same settings as before.

RMSE Values in Deep water (<1000m) between 66N and 66S in meters :
M2 Constituent: 0.06416788614061514
K1 Constituent: 0.023685223640391814
N2 Constituent: 0.013199779825957164
O1 Constituent: 0.015113880782811768
S2 Constituent: 0.02942663706041666

M2 Constituent plots:
M2_plot

components/mpas-ocean/src/shared/mpas_ocn_vel_self_attraction_loading.F
```diff
@@ -0,0 +1,2396 @@
...
+        LoveDat(1:4,1436) = (/ 1435.0_RKIND, -6.1353070000_RKIND, 1.2783782000e-3_RKIND, -1.8381416000e-3_RKIND /)
+        LoveDat(1:4,1437) = (/ 1436.0_RKIND, -6.1355685000_RKIND, 1.2776168000e-3_RKIND, -1.8369666000e-3_RKIND /)
+        LoveDat(1:4,1438) = (/ 1437.0_RKIND, -6.1358290000_RKIND, 1.2768562000e-3_RKIND, -1.8357928000e-3_RKIND /)
+        LoveDat(1:4,1439) = (/ 1438.0_RKIND, -6.1360887000_RKIND, 1.2760962000e-3_RKIND, -1.8346203000e-3_RKIND /)
+        LoveDat(1:4,1440) = (/ 1439.0_RKIND, -6.1363475000_RKIND, 1.2753368000e-3_RKIND, -1.8334492000e-3_RKIND /) !}}}
+        LoveDat(1:4,1441) = (/ 1440.0_RKIND, -6.1366054000_RKIND, 1.2745781000e-3_RKIND, -1.8322792000e-3_RKIND /)
+
+
+        if ((trim(config_ocean_run_mode) .eq. 'init') .and. & 
+            (trim(config_init_configuration) .eq. 'test_sht')) then
```
mark-petersen, line 2287: @knbarton, when we compile this in E3SM, we do not include the init code or the init flags (beginning in 'config_init').  It looks like you use this for testing during init mode? Could we just delete the condition
```
trim(config_init_configuration) .eq. 'test_sht'
```
and always set the scaling variable to 1 for init mode?

components/mpas-ocean/src/shared/mpas_ocn_vel_self_attraction_loading.F
```diff
@@ -0,0 +1,2396 @@
...
+
+            scaling(:) = 1.0_RKIND
```
mark-petersen, line 2289: Also, `scaling` is a general term. Could we use `LoveScaling`?

components/mpas-ocean/src/shared/mpas_ocn_vel_self_attraction_loading.F
sbrus89: @mark-petersen, I added this. I'll go ahead and take the condition out, as suggested.

components/mpas-ocean/src/shared/mpas_ocn_vel_self_attraction_loading.F
sbrus89: I agree, I'll change this too.

mark-petersen:
@sbrus89 thanks. Please check my last commit d6daa07. Stand-alone was not compiling with `SHTNS=false` because it referenced some shtns variables. I just put those lines within the cpp constructs, so should be fine for the parallel version.

components/mpas-ocean/src/shared/mpas_ocn_vmix.F
```diff
@@ -829,7 +829,7 @@ subroutine ocn_vel_vmix_tend_implicit_spatially_variable_mannings(meshPool, forc
            ! averaging the two bottomDepth cells.
            implicitCd = max(0.0025_RKIND, &
                         min(0.1_RKIND, &
-                          0.16_RKIND/log(250.0_RKIND*(bottomDepth(cell1)+bottomDepth(cell2)))**2 ))
+                          0.16_RKIND/(log(250.0_RKIND*(bottomDepth(cell1)+bottomDepth(cell2))))**2 ))
```
mark-petersen, line 832: Note, this parenthesis correction was added to this PR because `config_use_implicit_bottom_roughness` is only used in the tidal simulations used in this PR.

components/mpas-ocean/src/shared/mpas_ocn_vel_self_attraction_loading.F
```diff
@@ -0,0 +1,2399 @@
...
+                    !------------
+                    ! n = m
+                    !------------
+                    n = m
+
+                    ! Calculate associated Legendre polynomial for n=m (output pmnm2)
+                    call associatedLegendrePolynomials(n, m, startIdx, endIdx, l, pmnm2, pmnm1, pmn)
+
+                    ! Compute local integral contribution
+                    do iCell = endIdx, startIdx, -1
```
mark-petersen, line 293: @sbrus89 we typically place OMP pragmas on the outer loop, e.g.
```
               !$omp parallel
               !$omp do schedule(runtime)
               do iCell = 1, nCellsAll
                   ...
               end do
               !$omp end do
               !$omp end parallel
``` 
with private clauses for internal indices as needed. That said, I think the `$OMP` would go on the outermost loop m here, if `m` iterations are independent, which they appear to be. 

OpenMP is particularly important for the parallel implementation, but could also be used for serial now that we found that bash error.

mark-petersen: BTW, OMP is not needed for init mode. It is much less important for init subroutines in forward mode.

mark-petersen:
@sbrus89 and/or @knbarton could you, with parallel and serial with full SAL on, conduct these three tests after the OpenMP commit, in optimized mode:
1. restart test: run (say) a day, restart an extra day, and separately two days, compare restarts at end
2. partition test: run identical short test with two partitions (say 16 vs 32 cores for a day)
3. thread test: run with OPENMP=false, run for a day, recompile with OPENMP=true, set `OMP_NUM_THREADS=2`, run for day.

for all three, use a bfb comparison of last restart file, with python or ncdiff. Simple ncdiff commands: 
```
ncdiff file1.nc file2.nc diff.nc
ncwa -O -y mabs -v normalVelocity,temperature,salinity diff.nc diffmabs.n
```
max absolute difference should be exactly zero. You can use compass for this, but if you already have SAL runs set up, might be easier to just copy your settings from there.

components/mpas-ocean/src/shared/mpas_ocn_vel_self_attraction_loading.F
```diff
@@ -0,0 +1,2444 @@
...
+                    l = SHOrderDegreeToIndex(n,m)
+                    SnmRe_local(l) = 0.0_RKIND
+                    SnmIm_local(l) = 0.0_RKIND
+                    SnmRe(l) = 0.0_RKIND
+                    SnmIm(l) = 0.0_RKIND
+                enddo
+            enddo
+            !$omp end do
+
+            !$omp do schedule(runtime) private(blk,startIdx,endIdx,n,l,iCell)
```
mark-petersen, line 281: @sbrus89 I updated this. Please double check. I added all indices interior to the m loop.

sbrus89: @mark-petersen, I saw this. Those should be private by default, but it is better to be explicit.

mark-petersen (commented):
Merged into master locally and retested. Passes
```
```
Also compiled with gnu and intel on grizzly with SHTNS off (default). Passes `pr` test suite for both, with bfb match to compass reference. Compiled gnu with `SHTNS=true` but off using config flags. That also passes `pr` suite with both debug and optimized gnu.

sbrus89:
@jonbob and @mark-petersen: Now that #4700 has been submitted, I'll work on testing this with the reproducible sum implemented by @philipwjones .

philipwjones:
@sbrus89 you will notice that the reproducible sums require both the local and global sum to be computed with the same algorithm, so you'll have to refactor the transforms a little to expose an array (or array product) to be summed over. Let me know if you need anything

sbrus89:
@philipwjones, thanks for pointing that out. So that means the array `Snm_local` that gets passed to the `globalSum` needs to be of dimension (nCells,nOrder) so that the summation over all global cells is handled by the call to `globalSum`. Is there a way to tell `globalSum` to sum over the rows for each column, or is a separate call required for each column?

philipwjones:
@sbrus89 if the number of rows varies with each column, you'll probably want to use the sum-product interface and send a mask for the second array. The index space interface expects a single start-end along each dimension.

components/mpas-ocean/src/shared/mpas_ocn_vel_self_attraction_loading.F
```diff
@@ -0,0 +1,2457 @@
...
+                    nCellsPerProc, nCellsDisplacement, MPI_INTEGER, 0, domain % dminfo % comm, err)
+
+            ! SAL calculations only need to be initialized on process 1
+            if (curProc.eq.0) then
+                ! Set variables for file locations
+                mpasToGridFile = trim(config_mpas_to_grid_weights_file)
+                gridToMpasFile = trim(config_grid_to_mpas_weights_file)
+
+                ! Open netcdf weights files
+                call check( nf90_open(path = mpasToGridFile, mode = nf90_nowrite, ncid = toNcId) ,err)
```
matthewhoffman: Throughout this section, and elsewhere in this module, you should use a temporary error variable for the check, and then ior it with the existing err variable.  Otherwise an error early in the routine could be overwritten by a success later in the routine.  The typical way we do this is like:
```
call check( nf90_open(path = mpasToGridFile, mode = nf90_nowrite, ncid = toNcId), err_tmp)
err = ior(err, err_tmp)
```

You could also modify the `check` routine to do that for you.  If so, you'll still want to make similar adjustments in the various MPI_* routines (like MPI_SCATTERV).

(I noticed this after reviewing code Holly copied for her sea level model PR into MALI: https://github.com/MALI-Dev/E3SM/pull/21/files)

rljacob:
Waiting on reprosum in MPAS.

components/mpas-ocean/src/shared/mpas_ocn_vel_self_attraction_loading.F
mark-petersen: @sbrus89 could you add this `ior` line after each time an error status is returned from a subroutine? Thanks.

sbrus89: @mark-petersen, yes I'll do that. @knbarton has a PR into my fork adding these lines.

mark-petersen:
Passes 
```
SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.anvil_intel
```
with SAL off. Also passes stand-alone nightly suite with SAL off with gnu debug. Testing with SAL on now...

hollyhan:
Hi @sbrus89, @knbarton and @mark-petersen,

Hope all is well! 

I've adopted the subroutines `interpolate` and `ocn_vel_self_attraction_loading_init`in the code `mpas_ocn_vel_self_attraction_loading.F` in my work in coupling a sea-level model to MALI where I interpolate between the MALI mesh and the global Gaussian grid [.](https://github.com/MALI-Dev/E3SM/pull/21) While this PR is going through some testing, we (@matthewhoffman) have recently learnt that compiling out code with `DEBUG=true` yields an upper bound error due to`fbouds-check` flag this section of the code:`https://github.com/MALI-Dev/E3SM/blob/8b40fd42c4a5cbdf620883911b46140b8bfbcf8c/components/mpas-albany-landice/src/mode_forward/mpas_li_bedtopo.F#L726-L732`.  We found that the reason for this bound error is that because the index `n` that goes through within the inner while loop (`do while (rowValues(n). EQ. nRow`) is incremented (`n = n + 1`) is in the inner loop, then it goes out of bound at the last iteration of the inner loop and becomes bigger than `n_S`, skipping the outer loop condition `do while (n .LE. n_S)`. We are curious if you have run into this issue and have a solution for it already. I think @mark-petersen's comment directly above makes this comment very timely as well. 

Looking forward to hearing your thoughts!
Holly

mark-petersen:
Turning on SAL in this PR using 
```
config_use_tidal_potential_forcing = .true.
config_use_self_attraction_loading = .true.
config_use_parallel_self_attraction_loading = .true.
```
it passes all tests in the nightly suite with gnu debug on grizzly, except bfb decomp and bfb threading. When additionally  switching to
```
config_parallel_self_attraction_loading_bfb = .true.
```
It passes bfb decomp, as expected.  It still does not pass the bfb threading test. Sometimes it hangs when run with multiple threads, suggesting a race condition.

@philipwjones and/or @brian-oneill could you review the `$omp` lines in file `components/mpas-ocean/src/shared/mpas_ocn_vel_self_attraction_loading.F`? In particular, that starts with a 
```
!$omp parallel
```
but that is not needed, right? Parallel threads run throughout now? 

@philipwjones, when the bfb flag above is on, it runs the code you added in #4700. Can that run using multiple threads? If so, was it tested for bfb comparisons between different thread counts?

philipwjones:
@mark-petersen The omp parallel lines are needed - your recollection was actually opposite. We removed the global parallel region because it was causing problems with developers not realizing they were in a threaded region, so now each routine must manage its own threaded regions.  I looked through the rest of that routine and while not an exhaustive review, all the omp directives look fine.  The reprosums are not threaded (they could be made so with appropriate atomics there too), so the omp single around them is appropriate.

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F
```diff
@@ -826,7 +828,10 @@ subroutine ocn_time_integrator_split(domain, dt)!{{{
                         sshSubcycleCurWithTides(iCell) = &
                                  sshSubcycleCur(iCell) - &
                               tidalPotentialEta(iCell) - &
-                              config_self_attraction_and_loading_beta* &
+                              !!!!!!!!!!!!!!!! note config flag here is not set
+                              !correctly in init for tidal !!!!!!!!!
+                           ssh_sal_on * ssh_sal(iCell) - &
+                           (1 - ssh_sal_on) * config_self_attraction_and_loading_beta* &
```
mark-petersen, line 834: @philipwjones is it OK to use a real-valued config flag within a cell loop, or is it better to copy that to a local module variable on init?

philipwjones: It's ok, but I prefer a local module variable for several reasons. (1) you can check on init for valid values, (2) you can make a shorter variable name and (3) the config values are currently pointers which on occasion can cause problems. We will be fixing (3) soon.

mark-petersen (approved):
This now passes nightly regression suite with both intel and gnu, with the following flags:
```
config_use_tidal_potential_forcing = .true.
config_use_self_attraction_loading = .true.
config_use_parallel_self_attraction_loading = .true.
config_parallel_self_attraction_loading_bfb = .true.
```

@jon, this is ready to be merged now.

sbrus89:
@mark-petersen, it looks like the CI tests are failing after the last commit. It looks like some submodules may have been updated as a part of that commit. Does that have anything to do with it?

mark-petersen:
@sbrus89 thanks for pointing that out. I didn't mean to include that last commit in this PR anyway, so I removed it. I committed submodule changes by mistake, but the rest of the PR removes the forcing var_struct from the input stream. I will put that in a separate PR.

Could you test the head of the PR now with your standard case, for one last check?

mark-petersen (approved):
I retested with today's commits using gnu debug on badger with SAL and tidal potential on, and bfb flag first on. The full nightly suite passes.

@jonbob this is ready to merge when the repo is open.

jonbob:
@philipwjones and @matthewhoffman - can you please review after the changes have been made? Thanks

hollyhan:
This comment is to confirm that the issue I experienced in my previous comment (https://github.com/E3SM-Project/E3SM/pull/4472#issuecomment-1069750359) has been addressed by @knbarton's commit (badff01341450b7a792d446b3752e0f2cc86c785).

Previously, I experienced a bound-check error during coupled MALI - Sea Level Model simulations in which the subroutine `interpolation` adopted from the code `[components/mpas-ocean/src/shared/mpas_ocn_vel_self_attraction_loading.F` is utilized. The error only appeared in the DEBUG mode. The coupled MALI-SLM simulations were terminated with a runtime error with the following message: "Fortran runtime error: Index '1627' of dimension 1 of array 'rowvalues' above upper bound of 1626" (The array size of 'rowValues', 'colValues' and 'sValues (or n_S)' in the mapfile I used is 1626.) 

After incorporating this fix, coupled MALI - Sea Level Model simulations compiled in DEBUG mode do not experience any errors, and the simulation results (i.e., MALI thickness data interpolated between the MALI mesh and the global Gaussian grid) are correct. 

Thanks again, @knbarton, @mark-petersen, @sbrus89!

matthewhoffman (approved):

sbrus89:
Thanks for pointing this out and helping to test the fix @hollyhan!

components/mpas-ocean/src/shared/mpas_ocn_tendency.F
```diff
@@ -417,6 +422,15 @@ subroutine ocn_tend_vel(tendPool, statePool, forcingPool, &
...
                                tendVel, err)
 
+      ! Compute ssh additions due to self-attraction and loading
+      if(mpas_is_alarm_ringing(domain % clock, 'salComputeAlarm', ierr=err)) then 
+#ifdef MPAS_DEBUG
+          call mpas_log_write('       Computing SAL')
+#endif
+          call ocn_compute_self_attraction_loading(domain, forcingPool, dminfo, ssh, err)
+          call mpas_reset_clock_alarm(domain % clock, 'salComputeAlarm', ierr=err)
+      endif
```
philipwjones, line 432: after you've computed ssh_sal, you need to update it on the device for later use, so need to add at the end of this if block
!$acc update device (ssh_sal)

components/mpas-ocean/src/shared/mpas_ocn_vel_tidal_potential.F
```diff
@@ -171,9 +174,9 @@ subroutine ocn_vel_tidal_potential_tend(ssh, surfacePressure, tend, err)!{{{
 
          potentialGrad = - gravity*invdcEdge* &
             (  tidalPotEta(cell2) - tidalPotEta(cell1) &
-             + betaSelfAttrLoad &
-              *((ssh(cell2) - ssh(cell1)) &
-                + rho0gInv*(surfacePressure(cell2) - surfacePressure(cell1))))
+             + (1.0_RKIND - ssh_sal_on) * betaSelfAttrLoad *( (ssh(cell2) - ssh(cell1)) &
```
philipwjones, line 177: In the acc directives at the beginning of this loop, please add ssh_sal to the list of present() variables

philipwjones (changes requested):

mark-petersen:
@philipwjones can you check the acc changes I just made in [03a5551](https://github.com/E3SM-Project/E3SM/pull/4472/commits/03a5551d1f99edd6b8cbdf1955b20928091b28be)?

components/mpas-ocean/src/shared/mpas_ocn_vel_tidal_potential.F
philipwjones: Thanks for adding.

philipwjones (approved):
Changes for acc look fine. Approving based on all the other discussion/testing.

jonbob:
@sbrus89 - the scripts that update the bld files to be consistent with Registry changes only came up with a missing blank line in namelist_definition_mpaso.xml. Is it OK to push to your branch? Or I can tell you where the line is if you prefer to handle it

sbrus89:
@jonbob, thanks for catching that. Sure, it's totally fine to push to my branch.

rljacob:
Is this ready?

sbrus89:
As far as I know, it is.

jonbob:
Passes:
* SMS_D_Ld1.ne30pg2_r05_EC30to60E2r2.WCYCL1850.chrysalis_intel

with expected NML DIFFs.

jonbob:
merged to next

jonbob:
Missed changes that needed to go in the mpas-ocean build-namelist file. Retesting passes:
* SMS_D_Ld1.ne30pg2_r05_EC30to60E2r2.WCYCL1850.chrysalis_intel

Still with expected NML DIFFs. Re-merged to next

jonbob:

PR #4476, opened 2021-08-20
Push chunk loop into crm_physics()

whannah1:
This refactors the way the CRM is called so that all CRM instances in a single task can be run at once on the GPU, rather than all CRMs in a given "chunk". On properly configured GPU cases the speedup from this change is significant, and there is no notable change to CPU configurations tested on Summit and Cori-KNL. The changes required to push the chunk loop down are quite substantial, but there were additional complications from the crm_history() subroutine and the crm_rad and  crm_ecpp_output data structures that required more extensive changes. A new "ncrms" variable is defined in crm_physics to define the dimension of variables that hold all CRMs in a task (i.e. combined across all chunks in a task).

Note about testing - the changes are non-BFB for MMF tests only because the grouping of CRM columns has changed, which changes the the time step and sedimentation subcycling. Otherwise the changes are BFB, which can be demonstrated using the mmf_fixed_subcycle test mod. 

[BFB] - except for MMF tests without mmf_fixed_subcycle test mod

### DIFF
--- a/components/eam/src/physics/crm/crm_history.F90
+++ b/components/eam/src/physics/crm/crm_history.F90
@@ -368,7 +368,8 @@ subroutine crm_history_init(species_class)
 end subroutine crm_history_init
 !---------------------------------------------------------------------------------------------------
 !---------------------------------------------------------------------------------------------------
-subroutine crm_history_out(state, ptend, crm_state, crm_rad, crm_output, crm_ecpp_output, qrs, qrl)
+subroutine crm_history_out(state, ptend, crm_state, crm_rad, crm_output, & 
+                           crm_ecpp_output, qrs, qrl, icrm_beg, icrm_end)
    use physics_types,          only: physics_state, physics_tend, physics_ptend
    use phys_control,           only: phys_getopts
    use crm_state_module,       only: crm_state_type
@@ -391,6 +392,9 @@ subroutine crm_history_out(state, ptend, crm_state, crm_rad, crm_output, crm_ecp
 
    real(r8), dimension(:,:), intent(in) :: qrs        ! shortwave radiative heating rate
    real(r8), dimension(:,:), intent(in) :: qrl        ! longwave radiative heating rate
+   
+   integer, intent(in) :: icrm_beg ! CRM dimension index range
+   integer, intent(in) :: icrm_end ! CRM dimension index range
 
    !----------------------------------------------------------------------------
    ! local variables
@@ -405,7 +409,7 @@ subroutine crm_history_out(state, ptend, crm_state, crm_rad, crm_output, crm_ecp
    integer :: lchnk                    ! chunk identifier
    integer :: ncol                     ! number of atmospheric columns
    integer :: ixcldliq, ixcldice       ! liquid and ice constituent indices
-   integer :: i, k                     ! loop iterators
+   integer :: i, k, icrm               ! loop iterators
    logical :: use_ECPP
    logical :: use_MMF_VT
    character(len=16) :: MMF_microphysics_scheme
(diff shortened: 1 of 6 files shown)


xyuan (approved):

components/eam/src/physics/crm/crm_history.F90
```diff
@@ -443,126 +447,129 @@ subroutine crm_history_out(state, ptend, crm_state, crm_rad, crm_output, crm_ecp
    ! CRM_QRS + CRM_QRL output in radiation_tend will show a time lag.
 
    ! GCM level rad heating tendencies
-   call outfld('MMF_QRL   ',qrl/cpair, pcols, lchnk )
-   call outfld('MMF_QRS   ',qrs/cpair, pcols, lchnk )
+   call outfld('MMF_QRL   ',qrl/cpair, ncol, lchnk )
```
brhillman, line 450: I think we’ve looked at this probably half a dozen times, and I still get confused about outfld dimensions, but I don’t think you want to do this here, unless you explicitly make qrl and qrs have size ncol as passed to outfld. I think the size of the array or pointer you pass to outfld needs to match the value of the dimension you pass, or weird behavior can result. I.e., I think you could do:
```
call outfld(‘MMF_QRL’, qrl(1:ncol,:)/cpair, ncol, lchnk)
```
or
```
call outfld(‘MMF_QRL’, qrl/cpair, pcols, lchnk)
```
but not what you have changed this to.

brhillman: Also side note, why are we outputting these here? This same field is also output in radiation.F90, with the name QRS, QRL. This appears to be the same field, so we now have three versions of the heating rate: CRM_QRL (the crm-resolved heating), QRL (the gridmean heating rate), and MMF_QRL (also the gridmean heating rate). It looks like this has been in here, but why?

components/eam/src/physics/crm/crm_history.F90
```diff
@@ -443,126 +447,129 @@ subroutine crm_history_out(state, ptend, crm_state, crm_rad, crm_output, crm_ecp
...
    ! Why do we output this here?
    call outfld('PRES    ',state%pmid, pcols, lchnk )
    call outfld('DPRES   ',state%pdel, pcols, lchnk )
 
    ! CRM state variables on CRM grid
-   call outfld('CRM_U   ',crm_state%u_wind,      pcols, lchnk )
-   call outfld('CRM_V   ',crm_state%v_wind,      pcols, lchnk )
-   call outfld('CRM_W   ',crm_state%w_wind,      pcols, lchnk )
-   call outfld('CRM_T   ',crm_state%temperature, pcols, lchnk )
+   call outfld('CRM_U   ',crm_state%u_wind     (icrm_beg:icrm_end,:,:,:), ncol, lchnk )
```
brhillman, line 458: I think this might be a little dangerous, we might want to add an assertion to make sure that ncol = icrm_end - icrm_beg + 1.

components/eam/src/physics/crm/crm_history.F90
```diff
@@ -572,55 +579,56 @@ subroutine crm_history_out(state, ptend, crm_state, crm_rad, crm_output, crm_ecp
...
+   gwp(:ncol,:pver) = crm_output%gicewp(icrm_beg:icrm_end,:pver) + crm_output%gliqwp(icrm_beg:icrm_end,:pver)
    cwp(:ncol,:pver) = cicewp(:ncol,:pver) + cliqwp(:ncol,:pver)
 
-   call outfld('GCLDLWP' ,gwp,     pcols, lchnk)
-   call outfld('TGCLDCWP',tgwp,    pcols, lchnk)
-   call outfld('TGCLDLWP',tgliqwp, pcols, lchnk)
-   call outfld('TGCLDIWP',tgicewp, pcols, lchnk)
-   call outfld('ICLDTWP' ,cwp,     pcols, lchnk)
-   call outfld('ICLDIWP' ,cicewp,  pcols, lchnk)
+   call outfld('GCLDLWP' ,gwp,     ncol, lchnk)
```
brhillman, line 593: Same comment here as above. Don’t these arrays have size pcols, but you’re passing ncol to outfld? I am pretty sure we need to be consistent.

components/eam/src/physics/crm/crm_history.F90
whannah1: I'm not sure why we are outputting these here, I'm ok to get rid of them. I didn't realize they were a duplicate. 

But to your first point this is actually a very important change because the data is ONLY ncol values because we are working with data where each set of "ncol" values has been stapled together across all chunks on a given task. So I think it's only correct to use ncol rather than pcols here. I can add the explicit dimensions though, that's not a bad idea.

components/eam/src/physics/crm/crm_history.F90
whannah1: This is guaranteed by how the icrm values are set in the above routine, but it wouldn't hurt to have an extra check here.

components/eam/src/physics/crm/crm_history.F90
whannah1: I guess in this case we could use pcols... but I'd like to be consistent in our use of "ncol" in this whole routine rather than pcols.

components/eam/src/physics/crm/physpkg.F90
```diff
@@ -800,24 +803,30 @@ subroutine phys_run1(phys_state, ztodt, phys_tend, pbuf2d,  cam_in, cam_out)
   !-----------------------------------------------------------------------------
   ! CRM physics
   !-----------------------------------------------------------------------------  
+
+  ! Determine column start and end indices for crm_ecpp_output
+  ncol_sum = 0
   do c=begchunk, endchunk
+    ncol_beg(c) = ncol_sum + 1
```
brhillman, line 810: Probably makes more sense to call these icol_beg and icol_end since they are indices to the start and end of the chunk? I would also maybe argue that this block belongs closer to where it’s used, since it only appears to be used in one place right before the call to tphysbc2.

whannah1: ok, I'll do that.

components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -399,8 +404,8 @@ subroutine crm_physics_init(state, pbuf2d, species_class)
       call pbuf_set_field(pbuf2d, pbuf_get_index('RPRDSH')     , 0._r8)
       call pbuf_set_field(pbuf2d, pbuf_get_index('RPRDTOT')    , 0._r8)
       call pbuf_set_field(pbuf2d, pbuf_get_index('NEVAPR_SHCU'), 0._r8)
-      call pbuf_set_field(pbuf2d, pbuf_get_index('PREC_SH')    , 0._r8)
-      call pbuf_set_field(pbuf2d, pbuf_get_index('SNOW_SH')    , 0._r8)
+      call pbuf_set_field(pbuf2d, prec_dp_idx                  , 0._r8)
```
brhillman, line 407: Why did you only update a couple of these?

whannah1: Well I already added the indices in this routine, so I figured I'd use them. I thought about trying to remove all the extra pbuf_get_index calls, but I figured it's better to leave that to a dedicated clean up PR.

components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -632,448 +631,470 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf, cam_in, cam_out, &
...
+   crm_accel_factor = 0.
+   crm_accel_uv = .false.
+   call phys_getopts(use_crm_accel_out    = use_crm_accel_tmp)
+   call phys_getopts(crm_accel_factor_out = crm_accel_factor)
+   call phys_getopts(crm_accel_uv_out     = crm_accel_uv_tmp)
+   use_crm_accel = use_crm_accel_tmp
+   crm_accel_uv = crm_accel_uv_tmp
+
+   if (masterproc) then
+     if (use_crm_accel .and. MMF_microphysics_scheme/='sam1mom') then
```
brhillman, line 651: You should wrap MMF_microphysics_scheme in a `trim` for safety here.

components/eam/src/physics/crm/crm_history.F90
whannah1: Added check in latest commits

components/eam/src/physics/crm/sam/crm_module.F90
```diff
@@ -81,14 +81,21 @@ subroutine crm(lchnk, ncrms, dt_gl, plev,       &
...
     !-----------------------------------------------------------------------------------------------
-
-    integer , intent(in   ) :: lchnk                            ! chunk identifier (only for lat/lon and random seed)
     integer , intent(in   ) :: ncrms                            ! Number of "vector" GCM columns to push down into CRM for SIMD vectorization / more threading
     integer , intent(in   ) :: plev                             ! number of levels in parent model
     real(r8), intent(in   ) :: dt_gl                            ! global model's time step
     type(crm_input_type), target,intent(in   ) :: crm_input
     type(crm_state_type), target,intent(inout) :: crm_state
-    type(crm_rad_type),   target,intent(inout) :: crm_rad
+    real(crm_rknd), dimension(ncrms,crm_nx_rad,crm_ny_rad,crm_nz), intent(inout) :: crm_rad_qrad
```
brhillman, line 89: Why did we need to undo the encapsulation of crm_rad?

components/eam/src/physics/crm/crm_history.F90
brhillman: To be clear, I don’t want to use pcols…I want ncol, but we also need to make sure the arrays being passed to the routine are only sized ncol and not pcols but with the ncol argument.

components/eam/src/physics/crm/samomp/crm_module.F90
```diff
@@ -369,10 +369,10 @@ subroutine crm(lchnk, ncrms, dt_gl, plev,       &
   crm_rad_qi  = 0.
   crm_rad_cld = 0.
 #ifdef m2005
-  crm_rad%nc = 0.0
-  crm_rad%ni = 0.0
-  crm_rad%qs = 0.0
-  crm_rad%ns = 0.0
+  crm_rad_nc = 0.0
```
brhillman, line 372: These don’t look like they ever get associated now? I’m not seeing where these are associated to the crm_rad fields.

components/eam/src/physics/crm/sam/crm_module.F90
whannah1: as discussed on slack - I've reverted the interface changes but kept a few changes in crm_module to make the 2-mom sections consistent with the 1-mom sections.

components/eam/src/physics/crm/samomp/crm_module.F90
whannah1: It happens in the allocate routine.

brhillman:
Merged to next.

brhillman:
Merged to master.

Issue #4474, opened 2021-08-19
MMF aquaplanet and RCE compsets are NOT aquaplanets


whannah1:
Due to an XML issue with how we added wildcards in `components/eam/cime_config/config_component.xml` the MMF compsets that should be enabling the aquaplanet mode, including RCE, are not actually adding the appropriate flag. 
To verify this issue I pulled up the recent MMF RCE test on Chrysalis from the integration suite on the master branch and found that the `-aquaplanet` flag is missing from `CAM_CONFIG_OPTS` and the ncdata file entry was set to a file with realistic topography:
` ncdata		= '/lcrc/group/e3sm/data/inputdata/atm/cam/inic/homme/cami_mam3_Linoz_ne4np4_L72_c160909.nc'`

So, unfortunately, we haven't been testing what we thought we were testing. Luckily, this is all very simple to fix. For the MMF AQP compsets the problematic line is here:
`<value compset="_EAM%MMF*-AQUA"         >-aquaplanet </value>`
And we've recently fixed this same issue for the MMF SCM compset as follows
`<value compset="EAM%MMF[12OX]*-SCM">... </value>`
But there's a better way to use the wildcard and make sure all MMF configurations can be covered by simple adding a period before the asterisk:
`<value compset="_EAM%MMF.*-AQUA"        >-aquaplanet </value>`

I just verified that this change works well on Cori. 

I have a branch that I'm currently working on that will be non-BFB for all MMF compsets. This branch will also modify the default MMF configuration in that file, so I'm tempted to just include this simple fix into that PR rather than make a separate non-BFB PR.

PR #4477, opened 2021-08-20
Fix MMF AQP/RCE compsets

whannah1:
This fixes the wildcards used to specify compset options for certain MMF compsets. There are also some additional changes to the MMF physpkg.F90 initialization code to avoid using the microp_aero module. This ensures MMF cases will avoid an error that gets throw if liqcf_fix=true. The microp_aero code and the liqcf_fix option are irrelevant for all MMF configurations. 

fixes issue #4474 

[BFB] except for MMF AQP and RCE compsets

### DIFF
--- a/components/eam/bld/namelist_files/use_cases/2000_cam5_av1c-04p2-MMF-1mom.xml
+++ b/components/eam/bld/namelist_files/use_cases/2000_cam5_av1c-04p2-MMF-1mom.xml
@@ -19,7 +19,6 @@
 <convproc_do_gas>.false.</convproc_do_gas>
 <convproc_method_activate>2</convproc_method_activate>
 <demott_ice_nuc>.true.</demott_ice_nuc>
-<liqcf_fix>.false.</liqcf_fix>
 <regen_fix>.true.</regen_fix>
 <resus_fix>.true.</resus_fix>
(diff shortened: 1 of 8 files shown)


lee1046:
Do you mind applying the wildcard change in config_compset.xml as well? 
in config_compset.xml, we have `<value  compset="EAM%MMF[12OX]*-SCM">GREGORIAN</value>` in line 315 and `<value compset="EAM%MMF[12XO]*-SCM">TRUE</value>` in line 359.

lee1046:
If liqcf_fix parameter is not going to be used when running MMF, can the parameter be removed from the use_case nameslist for MMF?

whannah1:
> If liqcf_fix parameter is not going to be used when running MMF, can the parameter be removed from the use_case nameslist for MMF?

I think so? I'll try it.

whannah1:
So removing liqcf_fix from the MMF use case files worked. It doesn't really matter whether we leave it or not, but I like simplifying those files so I'll push the change. There's probably more of the variables we could remove now, but that's a task for another PR.

lee1046 (approved):
Everything looks good to me.

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #4481, opened 2021-08-22
Enable F2010 default SSTICE_DATA for all grids

wlin7:
The intended default F2010 SSTICE_DATA does not cover all grids due to
the grid attribute. This is to enable it for F2010 on all grids.

[BFB] non-BFB only for F2010[SC5-CMIP6] tests for some grids


tangq (approved):
The changes are straightforward with visual check. It looks good to me. Please go ahead and merge it.

wlin7:
Thanks for the review @tangq .
Merged to next.

wlin7:
Merged to master. 
No diff report. The prod test of F2010SC5-CMIP6 uses ne30pg2_r05_oECv3 that is previously accounted for on the grid attribute concerning setting the SSTICE_DATA_FILENAME.

PR #4482, opened 2021-08-23
Change bubble test init in standalone homme

oksanaguba:
This fixes an oversaturated state in the moist bubble initialization. Also, qsat in initialization now depends on RJ or Kessler physics. 

[nonbfb] for 1 homme test, moist bubble.

### DIFF
--- a/components/homme/src/share/control_mod.F90
+++ b/components/homme/src/share/control_mod.F90
@@ -269,9 +269,10 @@ module control_mod
   real (kind=real_kind), public :: bubble_xyradius = 2000.0!bubble radius along x or y axis
   real (kind=real_kind), public :: bubble_zradius = 1500.0 !bubble radius along z axis
   logical,               public :: bubble_cosine  = .TRUE. !bubble uniform or cosine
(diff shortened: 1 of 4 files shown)


oksanaguba:
Upn a conversation with Mark, removed a call to tests_finalize at the end of the init. The call is used in Weiran's paper to make NH and HY bubbles init identical, but in general, we agree that the init in NH bubble should remain non hydrostatic.

mt5555 (approved):

oksanaguba:
A note on PBL: moist bubble tests use dcmip2016_test1_forcing, which then uses pbl_type=dcmip16_pbl_type (0 or 1) to turn on PBL (0 is RJ PBL, 1 is something else). By default, dcmip16_pbl_type is set to 0, so, it is active in current moist bubble tests. The only dcmip2016 test that disables PBL is test3. 

If we want to disable PBL, we need to add dcmip16_pbl_type=-1 to the namelists. From the memory, disabling the PBL in the current namelists did not change much. We can consider it a part of the bubble test, too.

oksanaguba:
in next

PR #4483, opened 2021-08-23
Update F compsets with mpas-seaice for prescribed ice mode

wlin7:
F20TR-MPASSI is added for 20th century transient AMIP with
prescribed ice mode from mpas-seaice. Existing F2010SC5-CMIP6-MPASSI
is renamed as F2010-MPASSI. Note that F2010SC5-CMIP6-MPASSI
was using use_case file 2010S_cam5_CMIP6.xml, which would override the
final v2 default values for clubb_tk1 and gw_convect_hcf.

[BFB] for existing compsets

### DIFF
--- a/components/eam/cime_config/config_compsets.xml
+++ b/components/eam/cime_config/config_compsets.xml
@@ -57,6 +57,11 @@
   </compset>
 
   <compset>
+    <alias>F20TR-MPASSI</alias>
+    <lname>20TR_EAM%CMIP6_ELM%SPBC_MPASSI%PRES_DOCN%DOM_MOSART_SGLC_SWAV</lname>
+  </compset>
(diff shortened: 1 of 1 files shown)


jonbob (approved):
Approved based on visual inspection -- looks like the right long-names for using mpassi in these F-cases

rljacob:
Is anything still using 2010S_cam5_CMIP6.xml after this PR?

wlin7:
> Is anything still using 2010S_cam5_CMIP6.xml after this PR?

@rljacob  This file and the associated compset, along with FC5AV1C-L and tests using it will be removed in the next PR coming in today.

wlin7:
Merged to next.

PR #4484, opened 2021-08-23
add ideal age tracer

maltrud:
Add classic ideal age as a new passive tracer to the ocean, which is used to help trace ocean ventilation, for example.  The tracer source is effectively just an increment of dt, and the surface value is reset to zero in the top layer every timestep (except for under ice shelves).  This feature is turned off by default.

[NML]
[BFB]


xylar:
@maltrud, before reviews get too far, I think folks are being stricter than they used to about the branch naming convention.  This branch would need to be named: `maltrud/ocean/add-ideal-age`.  All lower case and with dashes rather than underscores.  GitHub doesn't make it possible to change the branch name for an open PR so we try to nip this in the bud before there are any reviews or substantive comments on a PR.

maltrud:
closed due to violation of branch naming convention.  new one is #4487.

PR #4486, opened 2021-08-24
Expanding OpenACC to encompass all of diagnostics routine

brobey:
Expanding the OpenACC region in diagnostics to encompass the whole routine. The OpenACC region has also been expanded for the equation-of-state routine since it is called from diagnostics. Two versions of the equation-of-state interface now exist -- one to enable the OpenACC calls from diagnostics and one for backward compatibility for other calls.

Should not impact CPU code - results should be bit-for-bit identical on the CPU.
Tested on Darwin with a V100 GPU with the QU240 test problem

[BFB]


jonbob:
@brobey - ah, thanks for getting the branch naming convention right. Could you please just modify the PR title to be descriptive instead of the branch name? Thanks again

jonbob:
Running e3sm_developer on chrysalis on a test merge

jonbob:
passed e3sm_developer, merged to next

jonbob:
merged to master

PR #4487, opened 2021-08-24
add ideal age tracer

maltrud:
Add classic ideal age as a new passive tracer to the ocean, which is used to help trace ocean ventilation, for example. The tracer source is effectively just an increment of dt, and the surface value is reset to zero in the top layer every timestep (except for under ice shelves). This feature is turned off by default.  Note, this is the second PR attempt since i named the first one incorrectly (#4484).

### DIFF
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F
@@ -140,7 +140,8 @@ module ocn_time_integration_split
          tracersTendPool,   &! structure holding tracer tendencies
          forcingPool,       &! structure holding forcing variables
          scratchPool,       &! structure holding temporary variables
-         swForcingPool       ! structure holding short-wave forcing vars
+         swForcingPool,     &! structure holding short-wave forcing vars
+         tracersIdealAgeFieldsPool     ! structure holding idealAge-related fields
 
       logical :: &
          activeTracersOnly   ! only compute tendencies for active tracers
@@ -207,7 +208,8 @@ module ocn_time_integration_split
          highFreqThicknessCur,        &! high frq thick at current time
          highFreqThicknessNew,        &! high frq thick at new     time
          lowFreqDivergenceCur,        &! low frq div    at current time
-         lowFreqDivergenceNew          ! low frq div    at new     time
+         lowFreqDivergenceNew,        &! low frq div    at new     time
+         tracerGroupIdealAgeMask       ! mask for resetting surface ideal age
 
       type (field1DReal), pointer :: &
          effectiveDensityField         ! field pointer for halo update
@@ -1917,6 +1919,28 @@ module ocn_time_integration_split
       call mpas_pool_begin_iteration(tracersPool)
       do while ( mpas_pool_get_next_member(tracersPool, groupItr) )
          if (groupItr%memberType == MPAS_POOL_FIELD) then
+
+            ! Reset iAge to zero where mask == 0
+            if (trim(groupItr%memberName) == 'idealAgeTracers') then
+                call mpas_pool_get_array(tracersPool, &
+                                         groupItr%memberName, &
+                                         tracersGroupNew, 2)
+                call mpas_pool_get_subpool(forcingPool, &
+                                           'tracersIdealAgeFields', &
+                                            tracersIdealAgeFieldsPool)
+                call mpas_pool_get_array(tracersIdealAgeFieldsPool, &
+                                         'idealAgeTracersIdealAgeMask', &
+                                          tracerGroupIdealAgeMask)
+
+                !$omp parallel
+                !$omp do schedule(runtime)
+                do iCell = 1, nCellsAll
+                   tracersGroupNew(1,1,iCell) = tracerGroupIdealAgeMask(1,iCell)*tracersGroupNew(1,1,iCell)
+                end do ! cells
+                !$omp end do
+                !$omp end parallel
+             endif
+
             call mpas_dmpar_field_halo_exch(domain, &
                          groupItr%memberName, timeLevel=2)
          end if
--- a/components/mpas-ocean/src/shared/mpas_ocn_tracer_ideal_age.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_tracer_ideal_age.F
@@ -9,9 +9,9 @@
 !
 !  ocn_tracer_ideal_age
 !
-!> \brief MPAS ocean restoring
-!> \author Todd Ringler
-!> \date   06/08/2015
+!> \brief MPAS ideal age tracer routines
+!> \author Mathew Maltrud
+!> \date   08/19/2021
 !> \details
 !>  This module contains routines for computing the tracer tendency due to restoring
 !
@@ -58,8 +58,8 @@ contains
 !  routine ocn_tracer_ideal_age_compute
 !
 !> \brief   computes a tracer tendency to approximate ideal age
-!> \author  Todd Ringler
-!> \date    06/09/2015
+!> \author  Mathew Maltrud
+!> \date    08/19/2021
 !> \details
 !>  This routine computes a tracer tendency to approximate ideal age
 !
@@ -112,26 +112,31 @@ contains
 
       integer :: iCell, iLevel, iTracer
 
-      !move to ocean constants
-      real (kind=RKIND), parameter :: c0 = 0.0_RKIND
-      real (kind=RKIND), parameter :: c1 = 1.0_RKIND
-
       err = 0
 
+      ! add a tendency increment equivalent to "dt" to entire domain below top level
       !$omp parallel
       !$omp do schedule(runtime) private(iLevel, iTracer)
       do iCell=1,nCellsSolve
-        do iLevel=minLevelCell(iCell),maxLevelCell(iCell)
+        do iLevel=minLevelCell(iCell) + 1, maxLevelCell(iCell)
           do iTracer=1,nTracers
-             ! zero tracers at surface to zero where idealAgeMask == zero
-             ! idealAgeMask should be equal to 1.0 elsewhere
-             tracers(iTracer, iLevel, iCell) = idealAgeMask(iTracer, iCell) * tracers(iTracer, iLevel, iCell)
-
-             ! add a tendency increment equivalent to "dt" to entire domain
-             tracer_tend(iTracer, iLevel, iCell) =   tracer_tend(iTracer, iLevel, iCell) +  &
-                                                     layerThickness(iLevel,iCell) * c1
+             tracer_tend(iTracer, iLevel, iCell) = tracer_tend(iTracer, iLevel, iCell) +  &
+                layerThickness(iLevel,iCell)
           enddo
         enddo
+      enddo
+      !$omp end do
+      !$omp end parallel
+
+      ! zero surface tendency where mask == 0
+      !$omp parallel
+      !$omp do schedule(runtime) private(iLevel, iTracer)
+      do iCell=1,nCellsSolve
+        iLevel = 1
+          do iTracer=1,nTracers
+             tracer_tend(iTracer, iLevel, iCell) = tracer_tend(iTracer, iLevel, iCell) +  &
+                idealAgeMask(iTracer, iCell)*layerThickness(iLevel,iCell)
+          enddo
       enddo
       !$omp end do
       !$omp end parallel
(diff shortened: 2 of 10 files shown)


xylar:
@maltrud, I'm not having any luck with this branch when I run standalone MPAS-Ocean tests on Chrysalis with intel and impi.  When I compile in debug mode, I'm seeing:
```
forrtl: severe (408): fort: (7): Attempt to use pointer IDEALAGEMASK when it is not associated with a target

libpnetcdf.so.3.0  0000155551067332  for_emit_diagnost     Unknown  Unknown
ocean_model        000000000179EA81  ocn_tracer_ideal_         189  mpas_ocn_tracer_ideal_age.F
ocean_model        00000000010DB1E8  ocn_forward_mode_         443  mpas_ocn_forward_mode.F
ocean_model        00000000010D7BC7  ocn_core_mp_ocn_c          78  mpas_ocn_core.F
ocean_model        0000000000415409  mpas_subdriver_mp         336  mpas_subdriver.F
ocean_model        00000000004105FD  MAIN__                     18  mpas.F
ocean_model        0000000000410562  Unknown               Unknown  Unknown
libc-2.28.so       000015554DABD7B3  __libc_start_main     Unknown  Unknown
ocean_model        000000000041046E  Unknown               Unknown  Unknown
```
This is for a test where I'm *not* using ideal age so the fact that `idealAgeMask` isn't available is to be expected.  I tried to figure out where the package was defined that should determine if this variable is allocated but couldn't figure it out.

Is there any chance I could convince you to run the `nightly` test suite from `compass` yourself?  At least in the longer term, that might get us to a working code in standalone MPAS-Ocean runs than if I run standalone tests and discover issues, you come up with fixes, and 24 hours later I run tests again.  If you're willing to give this a shot, please take a look at:
https://mpas-dev.github.io/compass/latest/developers_guide/quick_start.html
and other parts of the developer's guide.  If not, we can keep iterating but this could take awhile...

components/mpas-ocean/src/shared/mpas_ocn_tracer_ideal_age.F
```diff
@@ -112,30 +112,35 @@ subroutine ocn_tracer_ideal_age_compute(nTracers, nCellsSolve, minLevelCell, max
...
       enddo
       !$omp end do
       !$omp end parallel
 
+      ! zero surface tendency where mask == 0
+      !$omp parallel
+      !$omp do schedule(runtime) private(iLevel, iTracer)
+      do iCell=1,nCellsSolve
+        iLevel = 1
+          do iTracer=1,nTracers
```
xylar, line 136:
```suggestion
          iLevel = 1
          do iTracer=1,nTracers
```
Please fix the indentation here.

components/mpas-ocean/src/shared/mpas_ocn_tracer_ideal_age.F
```diff
@@ -145,20 +150,44 @@ end subroutine ocn_tracer_ideal_age_compute!}}}
...
+      call mpas_pool_get_array(tracersIdealAgeFieldsPool, &
+                               'idealAgeTracersIdealAgeMask', &
+                                idealAgeMask)
+
+      call mpas_pool_get_array(forcingPool, 'landIceMask', landIceMask)
+
+      ! set mask = 0 for open ocean
+      ! set mask = 1 under ice shelves
+      idealAgeMask = 0.0_RKIND
+      where (landIceMask /= 0) idealAgeMask(1,:) = 1.0_RKIND
```
xylar, line 190:
```suggestion
      if (associated(idealAgeMask)) then
          ! set mask = 0 for open ocean
          ! set mask = 1 under ice shelves
          idealAgeMask = 0.0_RKIND
          if (associated(landIceMask)) then
              where (landIceMask /= 0) idealAgeMask(1,:) = 1.0_RKIND
          end if
      end if
```

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F
```diff
@@ -1917,6 +1919,28 @@ subroutine ocn_time_integrator_split(domain, dt)!{{{
...
+                                         'idealAgeTracersIdealAgeMask', &
+                                          tracerGroupIdealAgeMask)
+
+                !$omp parallel
+                !$omp do schedule(runtime)
+                do iCell = 1, nCellsAll
+                   tracersGroupNew(1,1,iCell) = tracerGroupIdealAgeMask(1,iCell)*tracersGroupNew(1,1,iCell)
+                end do ! cells
+                !$omp end do
+                !$omp end parallel
```
xylar, line 1941:
```suggestion
                if (associated(idealAgeTracersIdealAgeMask)) then
                   !$omp parallel
                   !$omp do schedule(runtime)
                   do iCell = 1, nCellsAll
                      tracersGroupNew(1,1,iCell) = tracerGroupIdealAgeMask(1,iCell)*tracersGroupNew(1,1,iCell)
                   end do ! cells
                   !$omp end do
                   !$omp end parallel
                endif
```
I'm not sure if this is the right way but somehow this needs to not happen when not using ideal age.

Similar changes are needed in `si` module.  It seems like you're not doing this in RK4, so presumably equivalent code needs to be added to the RK4 module.

xylar:
With the fixes above, I was able to run successfully both with and without `idealAgeTracers` on Chrysalis with intel/impi.

To turn on ideal age, I needed both:
```
&tracer_forcing_idealAgeTracers
    config_use_idealAgeTracers = .true.
    config_use_idealAgeTracers_idealAge_forcing = .true.
```
With `config_use_idealAgeTracers_idealAge_forcing = .false.`, the tracer is present but always remains zero.

Here is what I see on level 4 (Fortran indexing) of the QU240 and QUwISC240 performance test cases after 10 days:
QU240
QUwISC240

It looks like it's doing exactly what is expected.  The ideal age in the surface layer is always zero.  The age in deeper layers increases with time, but at different rates that presumably relate to advection and mixing.  Under ice shelves (and nearby), the age is much higher than in other regions, as expected.

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F
philipwjones: If there's a logical flag for ideal age, I would prefer the checking of a flag rather than checking for association. A lot faster and less reliant on the implementation.

xylar: @philipwjones, that does sound better.  I'm a bit confused since there are something like 6 different packages associated with `idealAgeTracers` bue presumably there are 1 or more logical flags that could be used instead of the various `associated` checks.

xylar:
I ran the `nightly` regression suite on Chrysalis with intel/impi. All tests passed and most were bit-for-bit with a baseline that I made on 8/20.  The one test that is not bit-for-bit is: `ocean/ice_shelf_2d/5km/z-level/restart_test`.  This test case is the only one to use inactive top cells, so that would presumably be the source of the problem.

I am running with a test merge with the current `E3SM/master`, so I need to check if the change might have been introduced by `master` rather than this PR.

components/mpas-ocean/src/shared/mpas_ocn_tracer_ideal_age.F
```diff
@@ -112,30 +112,35 @@ subroutine ocn_tracer_ideal_age_compute(nTracers, nCellsSolve, minLevelCell, max
...
-      real (kind=RKIND), parameter :: c1 = 1.0_RKIND
-
       err = 0
 
+      ! add a tendency increment equivalent to "dt" to entire domain below top level
       !$omp parallel
       !$omp do schedule(runtime) private(iLevel, iTracer)
       do iCell=1,nCellsSolve
-        do iLevel=minLevelCell(iCell),maxLevelCell(iCell)
+        do iLevel=minLevelCell(iCell) + 1, maxLevelCell(iCell)
```
xylar, line 121: @maltrud, are you sure that this is correct?  Don't you want to increment even in the top layer because you are only resetting to zero in the open ocean, not under ice shelves?

maltrud:
@xylar i only set config_use_idealAgeTracers_idealAge_forcing = .true. as a default for e3sm (in namelist_defaults) but forgot to set that as the default in Registry_idealAge.  i have corrected that.

xylar:
Thanks, @maltrud. I don't think that addresses my comments directly because I'm running standalone, not E3SM. But still very good to have fixed.

maltrud:
@xylar i think it should?  if Registry sets the default to true, then you should have to do it manually, even in stand-alone, right?

xylar:
I see. Yes, that would make it on in standalone, too. Thanks!

I'm on my phone so code browsing is limited right now.

xylar:
@maltrud, I'm back on my Linux laptop and looking at the code but I don't see the change you mentioned.  Did you perhaps push your change to your old branch name and not this one?

maltrud:
@xylar , no, i haven't pushed it yet so you won't see changes.  i'll do that now.....

components/mpas-ocean/src/shared/mpas_ocn_tracer_ideal_age.F
```diff
@@ -145,20 +150,48 @@ end subroutine ocn_tracer_ideal_age_compute!}}}
...
+      call mpas_pool_get_array(forcingPool, 'landIceMask', landIceMask)
+
+      if (associated(idealAgeMask)) then
+
+      ! set mask = 0 for open ocean
+      ! set mask = 1 under ice shelves
+      idealAgeMask = 0.0_RKIND
+      where (landIceMask /= 0) idealAgeMask(1,:) = 1.0_RKIND
+
+      endif !  associated(idealAgeMask)
```
xylar, line 194: @maltrud, thanks for updating this. The indentation would need to be fixed.

But more importantly, @philipwjones had asked for a different solution that uses one of the logical variables to determine if `idealAgeMask` would be available rather than `associated()`.  Could you switch to that approach here and in the other similar code blocks instead?

mark-petersen:
Rebased on master. Compiled stand-alone with debug using both gnu and intel. Fails all tests in nightly test suite using split explicit. Will look into it further tomorrow.

xylar:
@mark-petersen, the tests are presumably failing because most of the issues I raised in my review have not yet been addressed.  We need to wrap several places that ideal age variables get accessed in `if` statements that check if the `idealAgeTracersPKG` and/or `idealAgeTracersIdealAgePKG` (that's a nasty name, by the way!) are active.  If you have taken over the PR, you could feel free to add these statements, at which point I'm happy to re-test.

maltrud:
@xylar and @mark-petersen :  something is very messed up with my branch.  somehow it is missing everything in the cime directory and who knows what else.  is there any chance of going back to my original add_idealAge branch where everything is there?

xylar:
@maltrud, we really need you to be working off of the latest `master`.  Otherwise, we have to do a test merge ourselves to test.

You need to run:
```
git submodule update --init --recursive
```
This is always required when building E3SM.  This will give you cime.  If it's not working for you on a specific directory, `rm -rf` the `cime` directory, then hard reset to the remote branch, then try again.

maltrud:
@xylar yes i know i need to do 'git submodule update --init --recursive', and i did for this case.  apparently only cime was missing and i have no idea why.  what i didn't know was that i could invoke it again without over-writing everything i've done, so i do have cime now and can actually build the code.

mark-petersen:
Passes gnu debug nightly test suite with ideal age tracers off and again on, except:
```
ocean/global_ocean/QU240/PHC/restart_test
  test execution:      SUCCESS
  test validation:     FAIL
ocean/global_ocean/QU240/PHC/decomp_test
  test execution:      SUCCESS
  test validation:     FAIL
  test execution:      SUCCESS
  test validation:     FAIL
  test execution:      SUCCESS
  test validation:     FAIL
```
All pass on master.

mark-petersen:
Now passes stand-alone nightly regression suite with all combinations of: ideal age tracers off and on; gnu and intel; debug and optimized. 

@maltrud since I've made some changes to the code, please retest what you had done before, and make sure it works as expected.

mark-petersen:
Passes, with ideal age off
```
ERS_Ld3.T62_oQU240.GMPAS-IAF.chrysalis_intel
```

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F
```diff
@@ -1917,10 +1919,32 @@ subroutine ocn_time_integrator_split(domain, dt)!{{{
...
+             endif
+
+            call mpas_dmpar_field_halo_exch(domain, groupItr%memberName, timeLevel=2)
```
mark-petersen, line 1944: > @maltrud said: looks to me like now only ideal age gets a halo update, instead of all tracers...

This halo update is outside of the ideal age if block. You might have been looking at an old version. But it is hidden, I'll put it at the top with a comment.

components/mpas-ocean/src/shared/mpas_ocn_tracer_ideal_age.F
```diff
@@ -145,20 +151,50 @@ end subroutine ocn_tracer_ideal_age_compute!}}}
...
+      call mpas_pool_get_subpool(forcingPool, &
+                                 'tracersIdealAgeFields', &
+                                  tracersIdealAgeFieldsPool)
+      call mpas_pool_get_array(tracersIdealAgeFieldsPool, &
+                               'idealAgeTracersIdealAgeMask', &
+                                idealAgeMask)
+
+      call mpas_pool_get_array(forcingPool, 'landIceMask', landIceMask)
+
+      if (config_use_idealAgeTracers.and.config_use_idealAgeTracers_idealAge_forcing) then
```
mark-petersen, line 190: > @maltrud said:  the routine exits if config_use_idealAgeTracers is .false.

Good point. I'll take that out.

mark-petersen:
Rebased on revised #4518, made corrections after comments above.

@maltrud can you retest using whatever your previous testing was? Thanks.

maltrud:
@mark-petersen , technically, shouldn't the halo update be _after_ tracerGroupsNew is updated?  i guess it doesn't matter here since we're resetting ideal age to zero for 1:nCellsAll?

mark-petersen:
Good point. Since we do a halo update anyway, let's loop through to `nCellsOwned`. Just as sec...

mark-petersen:
@maltrud try that. See if you have other comments. After you test, I'll run it through the whole battery of tests a final time.

maltrud:
@mark-petersen i ran for 1 month of the EC30 and it looks good.  please run the battery of tests at your convenience.

mark-petersen:
Passes 
```
```

mark-petersen:
With gnu debug on grizzly, passes all in new 'pr' regression suite.

xylar:
> With gnu debug on grizzly, passes all in new 'pr' regression suite.

@mark-petersen, are you using the `E3SM-Project` submodule in compass as your baseline for these tests?  If so, it would be surprising to me that the tests are passing on Grizzly since comparison with `E3SM/master` is failing on Chrysalis and therefore would also fail for this PR.  If you are using current `E3SM/master` as your baseline, there is a pretty high risk of unexpected changes creeping in as I am seeing in #4518.

xylar (approved):
I tested this with the `pr` test suite from compass on Chrysalis.  I tested optimized runs with both intel (impi) and gnu (openmpi).  All tests were bit-for-bit.

I don't have time to re-test the ideal age tracer itself, so I'm just approving based on my previous testing and my most recent do-no-harm tests. 

This PR is based off of #4518 so should be rebased onto E3SM/master once that PR gets merged so the file diffs in this PR remain meaningful.

mark-petersen (approved):
This can be merged after #4518.

mark-petersen:
Testing head again after rebasing. Passes
```
```
On grizzly, standalone gnu optimized passes and matches COMPASS baseline for `pr` suite.

jonbob:
Passed e3sm_developer on chrysalis, merged to next

jonbob:

PR #4489, opened 2021-08-25
Adding CRYO1950 compset

darincomeau:
Adds a new Cryosphere compset that is similar to WCYCL1950, except uses Cryosphere settings for Antarctic runoff (disables Antarctic runoff from the coupler to the ocn, includes ice-shelf melt fluxes and data icebergs). This compset points to spun-up ocn/ice initial conditions (restarts from a one month G-case).


jonbob (approved):
Approved based on a visual inspection and developer testing

Issue #4449, opened 2021-08-06
CLUBB does not conserve energy


oksanaguba:
There are a few issues:
1) clubb uses a brute force approach to conserve energy that got broken by me in the PR #3281 . 
2) even when 1) is fixed, from debugging a first few time steps in an ne30 F case, there is a limiter in physics_update that fixes negative vapor produced by clubb. The limiter changes energy, but there is no energy fix after the limiter. This leads to energy leaks from clubb.

I tried to fix (1), but due to (2) I could not get rid of energy diagnostics error messages. I assume that (2) also contributes to very infrequent MG2 energy errors and maybe ZM errors.

More info is on https://acme-climate.atlassian.net/wiki/spaces/COM/pages/3063218348/CLUBB+energy .

One could try to put a local, active only in a column, energy fixer in physics_update. Overall, I would think limiters should be a part of the physics package, not an interface, but that is an unlikely change. Maybe, placing a mass-conserving limiter in clubb_intr before brute-force energy conservation is computed would fix it. However, at least 2 more packages, ZM and MG2, might need the same change before the limiter can be removed from physics_update.

I am not sure how to proceed. I can test change (1) for global energy trends to see if it reduces the energy balance error further.

In case there is interest @wlin @golaz @mt5555 .

vlarson:
Hi @oksanaguba, 

- Do the negative vapor values occur only in the upper atmosphere, or at the top grid level?

- Have you looked at Brian Griffin's fixer in clubb_intr?:

https://github.com/larson-group/e3sm_release/tree/bmg929/atm/clubb_E3SM_master_1eaa248

It is designed to conserve both energy and water.  Brian has tested it thoroughly in our branch.

- Does your code call the PDF closure after the higher-order moments are updated (clubb_config_flags%ipdf_call_placement == ipdf_post_advance_fields)?  That should help yield a consistent water field when advance_clubb_core is exited.

- Within CLUBB, there is an attempt to clip liquid (rcm) in order to prevent vapor (rvm) from going negative (see subroutine clip_rcm):

https://github.com/larson-group/clubb_release/blob/67dc7892f77f7b81f62460a0747540ea5b2afff6/src/CLUBB_core/advance_clubb_core_module.F90#L3050

[...]

PR #4490, opened 2021-08-25
[wip] modify clubb energy fixer

oksanaguba:
The fix is needed due to the bug introduced in the PR #3281. however, the fix itself cannot remove energy leaks from clubb due to the limiter in physics_update. the fix was tested wrt climb in an F case but there may be no urge to take it into master.

see also issue https://github.com/E3SM-Project/E3SM/issues/4449 .

[non-BFB] for everything with EAM and CLUBB.

### DIFF
--- a/components/eam/src/physics/cam/clubb_intr.F90
+++ b/components/eam/src/physics/cam/clubb_intr.F90
@@ -38,8 +38,6 @@ module clubb_intr
   use shr_kind_mod,     only: core_rknd=>shr_kind_r8
 #endif
 
-use phys_debug_util, only: phys_debug_col
-
   implicit none
 
   private
@@ -48,7 +46,6 @@ use phys_debug_util, only: phys_debug_col
   ! ----------------- !
   ! Public interfaces !
   ! ----------------- !
-
 
   public :: clubb_ini_cam, clubb_register_cam, clubb_tend_cam, &
 #ifdef CLUBB_SGS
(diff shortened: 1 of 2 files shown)


oksanaguba:
Climo results for the commit 0cf2457 are here https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/onguba/theta/theta-clubb-leak.471017.0002-0006/def/ .

oksanaguba:
energy balance for the current commit:

Screen Shot 2021-08-31 at 2 29 12 PM

oksanaguba:
@wlin7 or @golaz  -- would you comment on this? either I can do more runs, or the issue (a very small leak of energy from clubb) remains and I will close this PR.

@vlarson -- after running with the change in this branch and with energy diagnostics, I do not see energy leak messages from clubb anymore. letting you know in case you want to look at this code.

rljacob:
Its too late to add this for v2 WC but @kvcalvin might want it for 2.1

components/eam/src/physics/cam/clubb_intr.F90
```diff
@@ -2489,7 +2504,9 @@ subroutine clubb_tend_cam( &
          ptend_loc%v(i,k)   = (vm(i,k)-state1%v(i,k))*invrs_hdtime                  ! north-south wind
          ptend_loc%q(i,k,ixq) = (rtm(i,k)-rcm(i,k)-state1%q(i,k,ixq))*invrs_hdtime  ! water vapor
          ptend_loc%q(i,k,ixcldliq) = (rcm(i,k)-state1%q(i,k,ixcldliq))*invrs_hdtime ! Tendency of liquid water
-         ptend_loc%s(i,k)   = (clubb_s(k)-state1%s(i,k))*invrs_hdtime               ! Tendency of static energy
+
+         ptend_loc%s(i,k)   = cpair*( temp_a(i,k) - state1%T(i,k) )*invrs_hdtime    ! Tendency from clubb
+         ptend_loc%s(i,k)   = ptend_loc%s(i,k)    - se_dis_k(k)*invrs_hdtime        ! Tendency to conserve energy
```
oksanaguba, line 2509: I need to replace += op here with just =

rljacob:
This would have to be approved by BGC.  @kvcalvin

rljacob:
@oksanaguba can you create an issue describing the bug?

rljacob:
I'm looking for an issue that specifically mentions what was broken by #3281 and this PR will fix.

PR #4501, opened 2021-08-29
Eliminate recursve call to the CPL:RUN timer

### LINKED ISSUE #4496: recursive timer call in water cycle compsets
While not necessarily a bug, it is often a sign of a timer logic bug when a timer is called recursively. In runs with water cycle compsets with budgets enabled we currently see for a five day run:

                                                                On  Called Recurse
        "CPL:RUN"                                               -    4878    240 

so one recursive call per ATM physics timestep. (Total calls is complicated, since CPL:RUN timers show up in multiple locations, not just where the recursion is taking place.)

This is not in fact a bug, but I do consider it an unfortunate feature, and one that could be eliminated, at the cost of some inconsistency in design or a larger, but consistent, change.

CPL:RUN is a peculiar timer (along with CPL:COMM and CPL:BUDGET). In CIME there are special versions of the start/stop timers called t_drvstartf and t_drvstopf, which call t_startf and t_stopf internally. There are also flags associated with the t_drvstartf/t_drvstopf calls that direct the routine to surround the particular start/stop timers with CPL:RUN start/stop timers (cplrun=.true.), or CPL:COMM start/stop timers (cplcom=.true.), or CPL:BUDGET start/stop timers (budget=.true.).

The problem is occurring because of the following type of logic:

     call t_drvstartf('timerA', cplrun=.true.)
        call t_drvstartf('timerB',cplrun=.true.)
        call t_drv_stopf('timerB',cprun=.true.)
     call t_drvstartf('timerA',cplrun=.true.)

This really does not make much sense, since we have already marked that anything between the 'timerA' calipers is a 'CPL:RUN' event, so including everything inside of 'timerB'. However, here the actual logic is

            call t_drvstartf ('CPL:ATMOCNP',cplrun=.true.,barrier=mpicom_CPLID,hashint=hashint(7))
            ...  
            if (do_budgets) then
               call cime_run_calc_budgets3()
            endif
            ...
            call t_drvstopf  ('CPL:ATMOCNP',cplrun=.true.,hashint=hashint(7))

and inside of cime_run_calc_budgets3 is

            call t_drvstartf ('CPL:BUDGET0',cplrun=.true.,budget=.true.,barrier=mpicom_CPLID)
            ...
            call t_drvstopf ('CPL:BUDGET0',cplrun=.true.,budget=.true.)

The problem "goes away" if I set cplrun=.false. in the t_drvstartf/t_drvstopf calls in cime_run_calc_budgets3. Also, cime_run_calc_budgets3 is ONLY called between the 'CPL:ATMOCN' calipers, so this does not do damage elsewhere. The consistency issue arises because there are also two similar routines

     cime_run_calc_budgets1
     cime_run_calc_budgets2

with similar logic and which are not called within an existing CPL:RUN timer. Here cplrun=.true. in the t_drvstartf/t_drvstopf calls is the correct thing to do.

So, if we want to fix this anomaly, we can either

a) just change cime_run_calc_budgets3 to remove the cplrun=.true. specifications, and perhaps add comments that this was done in case the use of cime_run_calc_budgets3 changes, or

b) add an optional (or required) parameter to cime_run_calc_budgets1, cime_run_calc_budgets2, and cime_run_calc_budgets3 that controls whether cplrun is .true. or .false. internally, and treat this explicitly where and when these routines are called.

This is all in 

     E3SM/driver-mct/main/cime_comp_mod.F90

worleyph:
The following logic in cime_run_atmocn_setup (in cime_comp_mod.F90):

   call t_drvstartf ('CPL:ATMOCNP',cplrun=.true.,barrier=mpicom_CPLID,hashint=hashint(7))
   ...
   if (do_budgets) then
      call cime_run_calc_budgets3()
   endif
   ...
   call t_drvstopf  ('CPL:ATMOCNP',cplrun=.true.,hashint=hashint(7))

and in cime_run_calc_budgets (also in cime_comp_mod.F90):

   call t_drvstartf ('CPL:BUDGET0',cplrun=.true.,budget=.true.,barrier=mpicom_CPLID)
   ...
   call t_drvstopf ('CPL:BUDGET0',cplrun=.true.,budget=.true.)

nests calls to the CPL:RUN timer, leading to recursion being recorded
in the timing library. This is not a bug, but it is also not necessary.
Everything in the CPL:ATMOCNP timer is marked as being in the CPL:RUN
timer. It adds nothing to redundantly mark the CPL:BUDGET0 timer as
being in the CPL:RUN timer.

Here an optional logical parameter 'in_cplrun' is added to
cime_run_calc_budgets3, and to cime_run_calc_budgets1 and
cime_run_calc_budgets2 for consistency, such that if set to 'true' it
indicates that the call to the routine is already inside of the
CPL:RUN timer and that this should not be set again within the
routine.

Fixes #4496

BFB

### DIFF
--- a/driver-mct/main/cime_comp_mod.F90
+++ b/driver-mct/main/cime_comp_mod.F90
@@ -4243,7 +4243,7 @@ contains
        ! ocn budget
        !----------------------------------------------------------
        if (do_budgets) then
-          call cime_run_calc_budgets3()
+          call cime_run_calc_budgets3(in_cplrun=.true.)
        endif
 
        if (drv_threading) call seq_comm_setnthreads(nthreads_GLOID)
@@ -4753,7 +4753,7 @@ contains
 
 !----------------------------------------------------------------------------------
 
-  subroutine cime_run_calc_budgets1()
+  subroutine cime_run_calc_budgets1(in_cplrun)
 
     !----------------------------------------------------------
     ! Budget with old fractions
@@ -4766,9 +4766,21 @@ contains
     ! it will also use the current r2x_ox here which is the value from the last timestep
     ! consistent with the ocean coupling
 
+    logical,intent(in),optional :: in_cplrun  ! flag indicating whether routine
+                                              ! called within the scope of the
+                                              ! CPL:RUN timer
+
+    logical :: lcplrun
+    !-------------------------------------------------------------------------------
+
+    lcplrun  = .true.
+    if (present(in_cplrun)) then
+       lcplrun = .not. in_cplrun
+    endif
+    
     if (iamin_CPLID) then
        call cime_comp_barriers(mpicom=mpicom_CPLID, timer='CPL:BUDGET1_BARRIER')
-       call t_drvstartf ('CPL:BUDGET1',cplrun=.true.,budget=.true.,barrier=mpicom_CPLID)
+       call t_drvstartf ('CPL:BUDGET1',cplrun=lcplrun,budget=.true.,barrier=mpicom_CPLID)
        if (lnd_present) then
           call seq_diag_lnd_mct(lnd(ens1), fractions_lx(ens1), infodata, do_l2x=.true., do_x2l=.true.)
        endif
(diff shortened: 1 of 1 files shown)


worleyph:
Tested on Chrysalis using 

     --compset WCYCL1850-4xCO2 --res ne30pg2_EC30to60E2r2 --machine chrysalis --pecount XS

comparing 1 month initial and 1 month continue between PR and unmodified. BFB with only change the expected one in the timing files.

sarats (approved):
Clear fix and solves the recursive timer issue.

rljacob:
Remind me why timers need to be concerned about what scope they are called in?

driver-mct/main/cime_comp_mod.F90
```diff
@@ -4766,9 +4766,21 @@ subroutine cime_run_calc_budgets1()
...
+    logical,intent(in),optional :: in_cplrun  ! flag indicating whether routine
+                                              ! called within the scope of the
+                                              ! CPL:RUN timer
+
+    logical :: lcplrun
+    !-------------------------------------------------------------------------------
+
+    lcplrun  = .true.
+    if (present(in_cplrun)) then
+       lcplrun = .not. in_cplrun
```
rljacob, line 4778: Why set lcplrun to the not of in_cplrun?   Don't you want them to have the same logical value?

worleyph: If in_cplrun is .true., then already within the scope of a CPL:RUN timer, and don't want to start/stop another CPL:RUN timer. Would be nice to be able to query this directly, but "recursive" calls to a timer can occur in valid situations (e.g., when a function containing timers  is called recursively). CPL:RUN is a different sort of timer, and these recursive calls are a distraction.

driver-mct/main/cime_comp_mod.F90
```diff
@@ -4243,7 +4243,7 @@ subroutine cime_run_atmocn_setup(hashint)
        ! ocn budget
        !----------------------------------------------------------
        if (do_budgets) then
-          call cime_run_calc_budgets3()
+          call cime_run_calc_budgets3(in_cplrun=.true.)
```
rljacob, line 4246: these cime_run_calc_budgetsX routines are each called once.  They don't need arguments.  Just change the value in the timer call in the routine.
What about cime_run_calc_budgets2?

worleyph:
> Remind me why timers need to be concerned about what scope they are called in?

@rljacob , not sure how to answer this. The GPTL library keeps track of whether a particular timer is active (has been started), and whether it is then started again before the original start call has been stopped, marking this second call as a recursive call and not double counting the time attributed to the timer. (Not sure what it does about the timer call count - I'll have to look at the code to see what it does.)

driver-mct/main/cime_comp_mod.F90
worleyph: Seems dangerous to me. These are subroutines that could be called from multiple locations, or the context to the call could be changed. The comments for some of these routines also seem out of date (RASM2 only for cime_run_calc_budgets3?). Seemed more consistent to take this approach. These are optional parameters, so don't have to be used.

> What about cime_run_calc_budgets2?

I do not understand the question.

rljacob: does the call to cime_run_calc_budgets2 need similar modifications?

worleyph: No. It is not called within the scope of an existing CPL:RUN timer. 

cime_run_calc_budgets3 is called from cime_run_atmocn_setup, and it is within cime_run_atmocn_setup that the CPL:RUN timer is started before the call to cime_run_calc_budgets3.

cime_run_calc_budgets1 and cime_run_calc_budgets2 are both called from cime_run and are not already with an active CPL:RUN timer.

rljacob: These cime_run_calc_budget subroutines were created just to make the main integration loop have fewer LOC.  They won't be called twice but might be moved.

rljacob (approved):

jgfouca:
Merged to next.

PR #4457, opened 2021-08-11
GPU port and reorganization of ocean tracer advection

philipwjones:
Initial GPU port for the ocean tracer advection routines that also includes some significant reorganization of these routines. In particular, this includes:
   - OpenACC port for all advection options, including loop directives and data transfers
   - Move of several advection-specific variables out of Registry and into local module allocatables to reduce size of mesh struct
   - move of some advection coefficient calculations out of framework and into local coefficient module
   - extraction of some common vertical flux calculation into a vertical module shared by both advection options
   - removal of some leftover pool retrievals and other general cleanup

Should be bit-for-bit when running on CPU (tested only with PGI on summit so far). 
Fixes #4386

PR #4518, opened 2021-09-05
add land ice logical to avoid allocation errors

### LINKED ISSUE #4517: ocean standalone LandIceFraction not associated error
When running master, 596b667b02 , ocean stand-alone with intel 19 debug, split explicit, no ice sheet pressure, we get the error
```
forrtl: severe (408): fort: (7): Attempt to use pointer LANDICEFRACTION when it is not associated with a target

ocean_model        0000000002EAC1E6  Unknown               Unknown  Unknown
ocean_model        0000000001635CA7  ocn_diagnostics_m         321  mpas_ocn_diagnostics.f90
ocean_model        00000000016D7246  ocn_init_routines         152  mpas_ocn_init_routines.f90
ocean_model        000000000154906E  ocn_forward_mode_         311  mpas_ocn_forward_mode.f90
```

This is also a potential problem for E3SM, it just was not caught by standard tests. The same configuration passes with gnu, debug and optimized.

mark-petersen:
Use of land ice variables, which are in packages and may not be allocated, were not protected by if statements.

Fixes #4517
[BFB]


mark-petersen:
@xylar and @philipwjones, see if you both agree with the logic here. It was only the intel 19 compiler in stand-alone that caught this, but it is good practice to protect unallocated variables. @philipwjones you had said previously that your preference is to use logicals in `if` statements, as they are faster than checking for associated variables or config strings.

xylar:
@mark-petersen, I believe this logical is not defined in the most intuitive place.  I would expect it to be added here:
https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-ocean/src/shared/mpas_ocn_surface_land_ice_fluxes.F#L62-L68

I'm also hesitant to call the flag `landIceOn` because this might suggest full land-ice coupling.  What this flag really indicates is that ice shelves depress the ocean surface and a `landIceMask` indicates where atmosphere and sea-ice fluxes cannot go.  Maybe `landIcePressureOn` would be an appropriate shorthand for this.

I'm pretty sure Matt Turner introduced the `landIceFluxesOff` and `standaloneOff` logicals. I feel like we should add `landIcePressureOn`, and then we should should probably rename `landIceFluxesOff` to `landIceFluxesOn` (switch the logic) and change `standaloneOff` to `landIceCoupledOn` (which doesn't require any changes in the logic, just a rename).

If this is needed sometime this week, please make these changes and I'll try to test.  If not, I can make the changes myself sometime next week.

components/mpas-ocean/src/shared/mpas_ocn_surface_land_ice_fluxes.F
```diff
@@ -677,14 +678,24 @@ subroutine ocn_surface_land_ice_fluxes_init(err)!{{{
...
+         landIceFluxesOn = .true.
+         standaloneOn    = .false.
+      case ('pressure_only','Pressure_Only','PRESSURE_ONLY')
+         landIcePressureOn = .true.
+         landIceFluxesOn = .false.
+         standaloneOn    = .false.
+      case ('off','Off','OFF')
+         landIcePressureOn = .false.
+         landIceFluxesOn = .false.
+         standaloneOn    = .false.
```
mark-petersen, line 695: @xylar thanks for the suggestion. I think this is much improved, we should never not use positive logic ;).  I also made the logic flags public if they are used elsewhere (specifically diagnostics). 

I thought about substituting `couplingOn` for `standAloneOff` but that doesn't make sense for these last two cases, i.e. `standAloneOff` ( = `.not.standAloneOn`) is really what is needed to run the subroutine `ocn_surface_land_ice_fluxes_build_arrays`.

components/mpas-ocean/src/shared/mpas_ocn_surface_land_ice_fluxes.F
```diff
@@ -677,14 +678,24 @@ subroutine ocn_surface_land_ice_fluxes_init(err)!{{{
 
       select case (trim(config_land_ice_flux_mode))
       case ('standalone','Standalone','STANDALONE')
-         landIceFluxesOff = .false.
-         standaloneOff    = .false.
+         landIcePressureOn = .true.
+         landIceFluxesOn = .true.
+         landIceStandaloneOn = .true.
```
mark-petersen, line 683: @xylar thanks for the suggestion. I think this is much improved, we should never not use positive logic ;).  I also made the logic flags public if they are used elsewhere (specifically diagnostics). 

I thought about substituting `landIceCoupledOn` for `standAloneOff` but that doesn't make sense for these last two cases, i.e. the new `LandIceStandAloneOn` is really what is needed to run the subroutine `ocn_surface_land_ice_fluxes_build_arrays`.

See also lines just below this for the different cases.

philipwjones: I actually introduced a lot of the internal "off" flags because we were always only checking whether something was off and returning, so the only logical check in the code was for an "off" case and we kept negating an "on" flag for that conditional. So if there are more "on" checks now for the allocates, it makes sense to switch to an "on" flag.

philipwjones (approved):
These changes look ok to me from a code standpoint. I don't know enough about the land ice use cases to verify all the different possible flag combinations.

mark-petersen:
Passes stand-alone nightly regression suite with debug using gnu and intel (intel was failing before this PR). With gnu optimized, gnu is bfb with master on stand-alone nightly regression.

mark-petersen:
Passes:
```
ERS_Ld3.T62_oQU240.GMPAS-IAF.chrysalis_intel
```

components/mpas-ocean/src/shared/mpas_ocn_surface_land_ice_fluxes.F
xylar: Thanks @mark-petersen.  @philipwjones, I think even if the flags happened to fit the places in the code where logicals were needed, the combination of `*Off` and `*On` logicals were confusing.  But it's probably my mistake not reviewing the PR(s) that introduced the changes more thoroughly in the first place.

xylar:
@mark-petersen, I think we are now using the more exhaustive `pr` suite for testing standalone, not `nightly`.  I will run it through the combination you ran (Intel and gnu, debug and optimized) on Chrysalis and approve as long as that suite is bit-for-bit where expected.

xylar:
@mark-petersen, your branch name does not conform to E3SM naming conventions.  The branch name needs to be all lowercase.  @jonbob, does this PR need to be re-opened with a new branch?

rljacob:
Since the review is far along, go ahead with the branch name as-is.

xylar:
Thanks @rljacob!  This is something that seems to happen on a regular basis, though, so the linter I suggested would sure help.

xylar:
@mark-petersen, I'm trying to run the `pr` test suite in debug mode but the cosine bell convergence test is just too heavy for debug mode as far as I can tell.

We should add a protocol for testing compass and standalone MPAS-Ocean PRs to the compass documentation.  My recommendation would be that we make sure both Intel and GNU compilers get tested, and that `pr` is tested in optimized and `nightly` in debug mode.  Does that sound right to you?  I would suggest that we not require any specific machine.  I can do this once I'm not on vacation.

xylar:
Hmm, the slow debug runtime seems only to be true for Intel.  GNU in debug mode is running just fine on the cosine bell.

jonbob:
@mark-petersen - you've assigned yourself for this PR. Do you not need me to merge it?

mark-petersen:
Thanks. I made the bug report at the same time and did that by mistake.

xylar:
I am testing the `pr` and `nightly` test suites on this branch, using E3SM/master from 2021/08/06 as a baseline.  I had no trouble running the baseline with Gnu and Intel in debug mode on Chrysalis (meaning I am not able to reproduce the bug reported in #4517 on that machine).

I will check these off as they pass:
- [ ] `pr`: intel impi optimized
- [ ] `nightly`: intel impi debug
- [ ] `pr`: gnu openmpi optimized
- [ ] `pr`: gnu openmpi debug

xylar:
@mark-petersen, I'm seeing non-bit-for-bit results for a lot of tests in `pr`.  This may be due to changes in `E3SM/master` since the last submodule update on 2021/08/06.  If so, the cause of these differences needs to be tracked down and either approved (by updating the submodule) or fixed before I can approve this PR.

This is a bigger project than I'm likely to be able to take on until I get back from vacation, meaning I don't think I can approve this PR until either you do that work or I get back.

xylar:
With intel impi optimized, the following `pr` tests are non-bit-for-bit:
```
00:21 FAIL ocean_global_ocean_QU240_PHC_RK4_performance_test
00:37 FAIL ocean_global_ocean_QU240_PHC_RK4_restart_test
00:36 FAIL ocean_global_ocean_QU240_PHC_RK4_decomp_test
00:36 FAIL ocean_global_ocean_QU240_PHC_RK4_threads_test
00:16 FAIL ocean_ice_shelf_2d_5km_z-level_restart_test
```
With gnu openmpi optimized, the following:
```
00:42 FAIL ocean_global_ocean_QU240_PHC_dynamic_adjustment
00:20 FAIL ocean_global_ocean_QU240_PHC_RK4_performance_test
00:39 FAIL ocean_global_ocean_QU240_PHC_RK4_restart_test
00:37 FAIL ocean_global_ocean_QU240_PHC_RK4_decomp_test
00:37 FAIL ocean_global_ocean_QU240_PHC_RK4_threads_test
00:48 FAIL ocean_global_ocean_ECwISC30to60_PHC_performance_test
00:19 FAIL ocean_ice_shelf_2d_5km_z-level_restart_test
```

xylar:
It does, indeed, seem like these issues are already on `E3SM/master` so we need to figure out where they were introduced.  To me, the most likely place seems to be https://github.com/E3SM-Project/E3SM/pull/4486, both because that was one of only a few MPAS-Ocean PRs and because compass testing was not included in its review.

xylar:
I tested E3SM from e6375e22c0 (before https://github.com/E3SM-Project/E3SM/pull/4486) and many of the bit-for-bit differences were gone, but not all.  The following still are non-bit-for-bit.
intel impi optimized:
```
00:41 FAIL ocean_global_ocean_QU240_PHC_dynamic_adjustment
00:54 FAIL ocean_global_ocean_ECwISC30to60_PHC_performance_test
00:19 FAIL ocean_ice_shelf_2d_5km_z-level_restart_test
```
gnu openmpi optimized:
```
00:16 FAIL ocean_ice_shelf_2d_5km_z-level_restart_test
```

That's as far as I'm likely to get this evening, but I'll pick up with further testing when I have time.  Naturally, anyone else is welcome to track down these changes but I'm not comfortable merging this PR until they are accounted for.

philipwjones:
@xylar #4486 only added directives. There were no other code changes for CPU runs. Given that and @mark-petersen 's successful tests on the other PR, are you sure your testing is working correctly?

xylar:
@philipwjones, yes, I believe my testing is working correctly.  I ran my baseline today using the `E3SM-Project` submodule in compass (which is different from the current E3SM/master).  I suspect that @mark-petersen is testing using `E3SM/master` as a baseline, but that means that changes seem to have crept in untested and unaccounted for since the last time the submodule in compass was updated.

xylar:
I certainly haven't traced any bit-for-bit changes to #4486.  I have simply found that some of them were introduced in or after #4486 and others were introduced before that.  It will likely take me several days (without help) to figure out which PRs and which commits introduced the non-bit-for-bit changes.  It will likely take longer to determine if these changes are acceptable or not.

mark-petersen:
I was able to reproduce Xylar's results above with on grizzly with compass master 958898c7 and
```
./conda/configure_compass_env.py --conda ~/miconda3 --compiler gnu

Currently Loaded Modules:
  1) cmake/3.16.2   3) mvapich2/2.3       5) hdf5-parallel/1.8.16   7) netcdf-h5parallel/4.7.3
  2) gcc/6.4.0      4) friendly-testing   6) pnetcdf/1.11.2         8) mkl/2019.0.4

ocean/global_ocean/QU240/PHC/RK4/performance_test
  test execution:      SUCCESS
  baseline comparison: FAIL
ocean/global_ocean/QU240/PHC/RK4/restart_test
  test execution:      SUCCESS
  test validation:     PASS
  baseline comparison: FAIL
ocean/global_ocean/QU240/PHC/RK4/decomp_test
  test execution:      SUCCESS
  test validation:     PASS
  baseline comparison: FAIL
ocean/global_ocean/QU240/PHC/RK4/threads_test
  test execution:      SUCCESS
  test validation:     PASS
  baseline comparison: FAIL
  test execution:      SUCCESS
  test validation:     PASS
  baseline comparison: FAIL
```
on this PR, compared to the baseline in compass/E3SM-Project 5fc4a7ea. 

I also tested just before and after #4452: 3674e3bd (before) and 5fc4a7ea (after). Both pass all bfb comparisons against 5fc4a7ea.

mark-petersen:
Here are the commits where bfb comparison tests against 5fc4a7e fail for the first time on grizzly gnu optimized:

* 7794aa0adb Merge branch 'philipwjones/ocn/gputadvect' (PR #4457)
`ocean_ice_shelf_2d_5km_z-level_restart_test`
* aa70c750c2 Merge branch 'brobey/mpas-ocean/openacc-diagnostics-new' (PR #4486)
```
00:27 FAIL ocean_global_ocean_QU240_PHC_RK4_performance_test
00:54 FAIL ocean_global_ocean_QU240_PHC_RK4_restart_test
00:53 FAIL ocean_global_ocean_QU240_PHC_RK4_decomp_test
00:53 FAIL ocean_global_ocean_QU240_PHC_RK4_threads_test
```

xylar:
Thanks @mark-petersen.  I have a compass PR to accept the changes from #4457 here: https://github.com/MPAS-Dev/compass/pull/233

Once that is addressed, I can look at the RK4 changes introduced in #4486, but if that PR was expected to be bit-for-bit it seems like maybe something isn't quite right with RK4 changes there?  What do you think @philipwjones?

mark-petersen:
Tested between #4486 merge (aa70c750c2, the new baseline in https://github.com/MPAS-Dev/compass/pull/236) and this PR using stand-alone `pr` test suite, gnu optimized on grizzly. All tests pass and compare bfb. Altered compass to double precision output (see https://github.com/MPAS-Dev/compass/issues/235).

xylar:
I might be able to re-test tomorrow (Saturday).

xylar (approved):
@mark-petersen, thanks for this work and sorry for the drama surrounding it.  I think the results has been good, though.  We have accounted for the differences in testing strategy and the non-bit-for-bit changes that emerged before this PR.

I tested the compass `pr` suite on Chrysalis using the current `E3SM-Project` submodule as a baseline.  Results were bit-for-bit with intel impi optimized, gnu openmpi optimized and gnu openmpi debug.  The baseline fails for the reasons pointed out in the PR for intel impi debug.

I ran the `nightly` suites with intel impi debug but without a baseline and all tests passed.  I tried to run the `pr` test suite with this configuration but the 60-km cosine bell test case seemed to hang as I had seen in my earlier testing.  The `pr` test suite may just be too big for debug mode...

**Update**: the `pr` test suite did eventually run with intel impi debug.

jonbob:
passed e3sm_developer on chrysalis, merged to next

jonbob:
merged to master

PR #4524, opened 2021-09-08
Align sea ice settings of SOwISC12to60E2 and WCAtl12to45E2 with WC14to60

proteanplanet:
This aligns the E3SM V2 Cryospheric campaign regionally-refined Southern Ocean mesh with the V2 Water Cycle WC14 mesh settings for sea ice. Thus the Antarctic regionally-refined settings in E3SM would become consistent with the Arctic settings, and vice versa.  The two changes made are that the number of iterations in the atmospheric flux calculations use 10 not 5 iterations, and high-frequency coupling is switched on, which means that a transient Ekman solution is permitted in the sea ice model. Note that there are no changes to the Water Cycle configurations, but two non-BFB changes to the regionally-refined Cryospheric configuration.   This change should be included in the V2 Tag.

### DIFF
--- a/components/mpas-seaice/bld/namelist_files/namelist_defaults_mpassi.xml
+++ b/components/mpas-seaice/bld/namelist_files/namelist_defaults_mpassi.xml
@@ -384,11 +384,9 @@
 <config_calc_surface_temperature>true</config_calc_surface_temperature>
 <config_use_form_drag>false</config_use_form_drag>
 <config_use_high_frequency_coupling>true</config_use_high_frequency_coupling>
-<config_use_high_frequency_coupling ice_grid="WCAtl12to45E2r4">false</config_use_high_frequency_coupling>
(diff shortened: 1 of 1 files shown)


proteanplanet:
After consultation with @vanroekel, added WCAtl12to45E2r4 to the meshes to be aligned with WC14, and added him as a reviewer.

stephenprice (approved):
Thanks for keeping us consistent @proteanplanet. Looks good to me.

darincomeau (approved):
Approved based on visual inspection and a successful one-month smoke test with the `SOwISC12to60E2r4` grid.

vanroekel (approved):
looks good to me.  Thanks @proteanplanet

jonbob:
Passes sanity testing and merged to next

jonbob:
merged to master

PR #4526, opened 2021-09-08
Fix CIMEROOTs for buildnml scripts

jgfouca:
It looks like these did not get updated when the scripts moved to
E3SM. Also, remove old python module load from mappy env.

[BFB]


rljacob (approved):
I can't tell what those mods to that python command are doing but trust you.

jgfouca:
@rljacob , the issue was that `CIMEROOT` was not actually pointing to cime. This problem was not noticed since these buildnmls are mostly called as a library and therefore do not need additional imports. I only noticed this because CIME's scripts_regression_tests does sometimes call these as shell commands and they weren't working.

PR #4528, opened 2021-09-09
Fix the CircleCI script that installs Singularity

lukaszlacinski:
This PR removes a bug that was introduced with the last PR that added a backup HTTP server with the Singularity source code tarball.


rljacob:
Bug was in #4508

PR #4530, opened 2021-09-11
Updates Intel's compiler version of PNNL cluster Constance

singhbalwinder:
Constance now uses Intel/19.0.5


rljacob:
@singhbalwinder can you finish merging this?

PR #4533, opened 2021-09-13
Hommexx: fixes to ForcingFunctor, and improved its unit tests

bartgol:
The ForcingFunctor had some bugs in the case where we do hard adjustment of tracers (ftype=2). The reason it was not showing is simply that we were not testing that case.

I fixed the functor, and added testing for both adjustment=true and adjustment=false. I did _not_ add testing for `adjust_ps=false`, since it is hard coded to true in both f90 and cxx.

This PR will need a follow-up to sort out forcing calls for SCREAM and EAM cases.

[bfb] but needs new homme baselines.


ambrad (approved):

ambrad:
@bartgol @oksanaguba is this still WIP, or is it ready to go in now that the v2 tag was made?

bartgol:
I think we can remove WIP.

oksanaguba:
Could I please ask you to run also other tests on weaver, at least cxx vs F? I will run homme suites on anvil against baselines.

bartgol:
> Could I please ask you to run also other tests on weaver, at least cxx vs F? I will run homme suites on anvil against baselines.

Yes, I'm going to run the dev test suite, which includes baselines and F90-vs-CXX.

bartgol:
> Could I please ask you to run also other tests on weaver, at least cxx vs F? I will run homme suites on anvil against baselines.

I ran `ctest -L dev` on weaver, and all tests passed. There were only 18 tests (6 tests, times f90, cxx, f90-vs-cxx) since preqx was turned off when I generated baselines. Would you like me to test preqx too?

oksanaguba:
homme tests and atm_integration pass on anvil up to rare fails for forcing_ut. 
new bfb tests fail on weaver, so this is not done.

relevant, forcing logic for CAM and SCREAM still needs to be sorted, but it will be a separate PR.

oksanaguba:
New tests pass on weaver for bfb part.

oksanaguba:
Regarding these not new lines in ForcingFunctor:
```
      // If NUM_LEV=NUM_LEV_P, the code above already took care of the last level
      if (NUM_LEV!=NUM_LEV_P) {
        // Last level
        Kokkos::single(Kokkos::PerThread(kv.team),[&](){
          fphi(LAST_INT_PACK)[LAST_INT_PACK_END] = (fphi(LAST_INT_PACK)[LAST_INT_PACK_END] - phi_i(LAST_INT_PACK)[LAST_INT_PACK_END]) / m_dt;
        });
      }
```
They may not be needed, they deal with the last level for phi, which is always phi_s. F code does the same thing, but it was much easier to write in F. Forcing for phi can be of num_lev size, too.

oksanaguba:
One more comment -- there is no easy way to verify that diagnostics calls match between F and cxx, since we do not have tests (I could have run a few runs and compare text but I did not).

oksanaguba:
All theta-f* tests on weaver look as expected, this is ready.

oksanaguba:
in next

oksanaguba:
a follow-up Pr needs to address missing applycam_tracers for scream (cause that is not done in dp_coupling layer, I assume). it would need information on which nstep iteration is calling prim_run_subcycle. this can be done by introducing iteration var in control mod to be visible in prim_run_subcycle (solution 1) or by rewriting the loop in EAM interface that calls prim_run_subcycle (solution 2). I will first implement solution 1 since it is much smaller code change and add tests, then we can switch to solution 2 (more invasive, but guarded by tests).

PR #4539, opened 2021-09-14
Change compset name for build test

rljacob:
Remove the "A_" from the compset used in the build test to match
new v2 standards.

[BFB]


rljacob:
the circleci check passed to merging to master.

PR #4545, opened 2021-09-17
remove se_fv_nphys namelist variable

whannah1:
se_fv_nphys is an old variable from the original physgrid implementation that is no longer needed.

[BFB]


ambrad (approved):
This is almost certainly correct, but I recommend running e3sm_developer against baselines, e.g. on Chrysalis, to be sure of what we can expect when this is merged: for example, NLDIFF.

whannah1:
All e3sm_atm_developer tests pass on Chrysalis.

PR #4550, opened 2021-09-21
For Cori, update cmake module version

### LINKED ISSUE #4548: Error at Setup on Cori after NERSC installed new default cmake version
On Sep 21, NERSC updated default cmake version. The old version (that we were using) is still there, however, a module warning now appears of that it used. If CIME detects a warning/error from a module command, it will fail. This can be turned off (using allow_error=true) and I tested that. However, in this case, it seems easy enough to simply update cmake version just to avoid this.

`ndk/machinefiles/cori-update-cmake`

ndkeen:
On Sep 21, NERSC updated default cmake version. The old version (that we were using) is still there, however, a module warning now appears if that it used. If CIME detects a warning/error from a module command, it will fail. This behavior can be turned off (using allow_error=true) and I tested that. However, in this case, it seems easy enough to simply update cmake version just to avoid this. (Thanks @whannah1)

Fixes #4548

[bfb]

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -179,7 +179,7 @@
         <command name="rm">git</command>
         <command name="load">git</command>
         <command name="rm">cmake</command>
-        <command name="load">cmake/3.20.2</command>
+        <command name="load">cmake/3.21.3</command>
         <command name="load">perl5-extras</command>
       </modules>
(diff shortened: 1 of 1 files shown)


ndkeen:
I tested with e3sm_developer on cori-knl and a single test on cori-haswell.

PR #4557, opened 2021-09-23
Adds lake water storage to prevent negative runoff

mabrunke:
This branch incorporates the UArizona's new lake water storage option.  @hydrotian found that this option eliminates negative runoffs from lakes, as we intended it to do.

[NML]
[non-BFB] for X-cases due to new cpl field

### DIFF
--- a/driver-mct/shr/seq_flds_mod.F90
+++ b/driver-mct/shr/seq_flds_mod.F90
@@ -1387,6 +1387,15 @@ contains
     attname  = 'Faxx_evap'
     call metadata_set(attname, longname, stdname, units)
 
+    if (trim(cime_model) == 'e3sm') then
+       call seq_flds_add(l2x_states,"Sl_wslake")
+       longname = 'Lake water storage'
+       stdname  = 'lake_water_storage'
+       units    = 'kg m-2'
+       attname  = 'Sl_wslake'
+       call metadata_set(attname, longname, stdname, units)
+    end if
+
     ! Dust flux (particle bin number 1)
     call seq_flds_add(l2x_fluxes,"Fall_flxdst1")
     call seq_flds_add(x2a_fluxes,"Fall_flxdst1")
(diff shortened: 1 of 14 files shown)


rljacob:
Is this BFB with current watercycle cases?  Is there a test that uses it?

rljacob:
Also please fix the title of the PR.

driver-mct/shr/seq_flds_mod.F90
```diff
@@ -1387,6 +1387,15 @@ subroutine seq_flds_set(nmlfile, ID, infodata)
...
+    if (trim(cime_model) == 'e3sm') then
+       call seq_flds_add(l2x_states,"Flrl_wslake")
+       longname = 'Lake water storage flux'
+       stdname  = 'lake_water_storage_flux'
+       units    = 'kg m-2 s-1'
+       attname  = 'Flrl_wslake'
+       call metadata_set(attname, longname, stdname, units)
+    end if
+
     ! Dust flux (particle bin number 1)
```
jonbob, line 1399: @mabrunke - we no longer share this code with cesm so the check for cime_model is unnecessary and probably should be removed

hydrotian (approved):

mabrunke:
@rljacob I'll have to re-do my tests.  Right now, I'm getting diffs for all F cases, but I think it's because I need to compare to an older version of master.

jonbob (approved):
Approved based on visual inspection and developer testing. But it should be BFB with current codebase for all configurations except X-cases and should not be merged until that is tested

jonbob:
Thanks @mabrunke - and that failed test is expected, since we have added fields to the cpl

mabrunke:
@peterdschwartz Sounds good to me.

rljacob:
@peterdschwartz did you make the code mods you mentioned above?

peterdschwartz:
@rljacob @mabrunke I pushed the changes.  Other PRs may need a warning  for when this goes in as it will cause 
`FIELDLIST field lists differ (otherwise bit-for-bit)` 
Only one of the tests SEQ.X.  is actually nonBFB, but (im assuming) all tests will need to be blessed.

peterdschwartz:
Merged to next

PR #4560, opened 2021-09-26
[wip] add nu_top code to thetaxx

oksanaguba:
duplicated code -- to make it run and bfb with F.
one could potentially try to re-use 2 functors, firstLaplace and updateStates if they depend on multiplier vector (nu, or dt*rspheremp, or dt*nu_scale*nu), but if the plan is to use different BE packing with smaller number of levels for nu_tom, like in F code, then these 2 functors will remain different.

add tests

run on gpu

remove buffer vars


oksanaguba:
trying to shorted a view of touched files

Issue #4495, opened 2021-08-26
Fail with SMS.T62_oQU120_ais20.MPAS_LISIO_TEST when trying newer version of gcc9


ndkeen:
ON cori-knl, we currently use gcc/8.3.0. Both `SMS.T62_oQU120_ais20.MPAS_LISIO_TEST` and `SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST` complete with gnu using gcc/8.3.0 (though both show a 10% memory growth in the first 5 days which may be unrelated -- even more memory growth with Intel compiler).

I tried updating to gcc/9.3.0 and I found an error with
`SMS.T62_oQU120_ais20.MPAS_LISIO_TEST`, however, `SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST` still completes. 
 Note the odd filenames:

```
-rw-rw-r--  1 ndk ndk     9961 Aug 26 13:13  atm.log.46099248.210826-131221
-rw-rw-r--  1 ndk ndk   229513 Aug 26 13:13 'log.seaice.0125.d****.err'
-rw-rw-r--  1 ndk ndk   238495 Aug 26 13:13 'log.seaice.0120.d****.err'
-rw-rw-r--  1 ndk ndk   516380 Aug 26 13:13 'log.seaice.0119.d****.err'
-rw-rw-r--  1 ndk ndk   572879 Aug 26 13:13 'log.seaice.0079.d****.err'
-rw-rw-r--  1 ndk ndk    65859 Aug 26 13:13 'log.seaice.0078.d****.err'
-rw-rw-r--  1 ndk ndk   488992 Aug 26 13:13 'log.seaice.0076.d****.err'
-rw-rw-r--  1 ndk ndk   197716 Aug 26 13:13 'log.seaice.0063.d****.err'
-rw-rw-r--  1 ndk ndk   371334 Aug 26 13:13  log.seaice.0062.d0480.err
-rw-rw-r--  1 ndk ndk    16877 Aug 26 13:13 'log.seaice.0054.d****.err'
-rw-rw-r--  1 ndk ndk   235309 Aug 26 13:13 'log.seaice.0049.d****.err'
-rw-rw-r--  1 ndk ndk   460980 Aug 26 13:13  abort_seaice_0001-01-01_00.00.00_block_55.nc
-rw-rw-r--  1 ndk ndk   498144 Aug 26 13:13  abort_seaice_0001-01-01_00.00.00_block_51.nc
[...]

Issue #4151, opened 2021-03-09
Explicit interface error message in ODEmod.F90 when using gfortran-10


mpaiao:
I am a new user of E3SM-FATES, and currently I am trying to install the model on my computer (MacBook) to familiarize with the code and run initial simulation.  I am currently using version 10.2.0_4 of gfortran, gcc, and g++, installed with HomeBrew. When I try to compile the model using IELMFATES, it fails due to multiple messages similar to the one below:
```
Error: Explicit interface required for polymorphic argument at (1)
/path/to/E3SM/components/elm/src/external_models/sbetr/src/betr/betr_math/ODEMod.F90:57:16:

   57 |     call odefun(extra, y0, dt, t, nprimeq, neq, f)
      |                1
Error: Explicit interface required for polymorphic argument at (1)
```
Is there any fortran compilation flag that I should turn on to ignore these errors without modifying the code? 

For reference, I managed to compile this file, but I had to edit it. I added two abstract interfaces at the beginning of the file (right before ```interface get_rerr```).
```
  abstract interface
     subroutine odebbks(extra,y0,dt,t,nprimeq,neq,f)
        use bshr_kind_mod , only : r8 => shr_kind_r8
        use gbetrType     , only : gbetr_type
        implicit none
        class(gbetr_type), intent(inout) :: extra
        real(r8)         , intent(in)    :: y0(neq)
        real(r8)         , intent(in)    :: t
        real(r8)         , intent(in)    :: dt
        integer          , intent(in)    :: nprimeq
        integer          , intent(in)    :: neq
        real(r8)         , intent(out)   :: f(neq)
     end subroutine odebbks
  end interface


  abstract interface
     subroutine oderkx(extra,y0,dt,t,neq,f)
        use bshr_kind_mod , only : r8 => shr_kind_r8
        use gbetrType     , only : gbetr_type
        implicit none
        class(gbetr_type), target        :: extra
        real(r8)         , intent(in)    :: y0(neq)
        real(r8)         , intent(in)    :: t
        real(r8)         , intent(in)    :: dt
        integer          , intent(in)    :: neq
        real(r8)         , intent(out)   :: f(neq)
     end subroutine oderkx
  end interface

```
and replaced the declaration ```external :: odefun``` with either ```procedure(odebbks) :: odefun``` or ```procedure(oderkx) :: odefun``` in every subroutine in ODEmod.F90. I'm still working on the compilation flags and options to build the executables, so I am not sure these changes work fine.

[...]

PR #4563, opened 2021-09-27
Adding support for ANL GCE nodes

jayeshkrishna:
Adding support for the ANL CELS GCE compute nodes


jayeshkrishna:
Note: To build E3SM on the ANL CELS compute nodes the following minor CIME patch is required for now (I will be creating a CIME PR soon for this change)
```
diff --git a/scripts/lib/CIME/XML/env_mach_specific.py b/scripts/lib/CIME/XML/env_mach_specific.py
index 86b0ff869..3d477efc5 100644
--- a/scripts/lib/CIME/XML/env_mach_specific.py
+++ b/scripts/lib/CIME/XML/env_mach_specific.py
@@ -149,7 +149,7 @@ class EnvMachSpecific(EnvBase):
         # setup script if it exists.
         init_path = self.get_module_system_init_path("sh")
         if init_path:
-            source_cmd = "source {} && ".format(init_path)
+            source_cmd = ". {} && ".format(init_path)
         else:
             source_cmd = ""
 

```

dqwu:
With GCC 11, there are some build errors from components/elm/src/external_models/sbetr/src/betr/betr_math/ODEMod.F90:
```
  672 |           call odefun(extra, y, dt2, tt, nprimeq, neq, f)
      |                      1
Error: Explicit interface required for polymorphic argument at (1)
```
This is a know issue, see #4151

dqwu:
After all build errors are fixed, a simple ne4 F case still fails and log.seaice.0001.err shows
```
----------------------------------------------------------------------
Beginning MPAS-seaice Error Log File for task       1 of      16
...
ERROR:  picard convergence failed!
ERROR:  ==========================
```
This is also a known issue for GCC 9 or higher, see #4495

jayeshkrishna:
@dqwu: I have rebased the branch with the new compiler flags required to build the F case. With the latest version of the branch and adding in the CIME fix (mentioned above) and the sbetr fix (PR #4151) I can build the F case (--compset F2010 --res ne4_oQU240) on the ANL GCE nodes (Fixes: https://gist.github.com/jayeshkrishna/9ed36d03dfa72b1a6504d9fd1ad29001).

However as you noted above the F case still fails with the "picard convergence error" in the MPAS sea ice component.

Since we can build the case and the required fixes are actively being worked on, this PR should be good to go. We can continue using anlworkstation (old ANL compute nodes) for E3SM tests with Scorpio until these issues are fixed on master.

dqwu (approved):

PR #4564, opened 2021-09-27
Update CIME to 80220601778

rljacob:
Update cime to master version 80220601778
post cime6.0.9 tag

E3SM-relevant changes
Fix PGN and TSC tests
Allow E3SM to use cmake macros instead of config_compilers (not yet active)
Fix scripts_regression_tests on mappy for e3sm

Other changes:
Adds a container machine definition for CESM
update modules and build flags on cori for CESM
if nuopc and any datamodel add a cdeps directory

[BFB]


jgfouca (approved):

rljacob:
merged to next

PR #4571, opened 2021-09-30
Fix some Hommexx SL and physgrid remap issues.

ambrad:
* Fix some implicit `this` captures. These don't actually cause a problem unless HOMMEXX_CUDA_SHARE_BUFFER is on.
* Fix an C++17-style static_assert to make it C++11 compliant. This fixes the sandiatoss3 homme_integration_next_intel build.

[BFB]


ambrad:
* Reproduced limiters_ut failure on Anvil; with this PR it now passes.
* Built HOMMEBFB_P24.f19_g16_rx1.A.sandiatoss3_intel with this PR.
* Ran cuda-memcheck on compose_ut and gllfvremap_ut with HOMMEXX_CUDA_SHARE_BUFFER on and off; now clean for all four.

ambrad:
Closing as we'll add these commits to the original PR.

PR #4573, opened 2021-10-01
add nu_top to theta xx

oksanaguba:
Adding nu_top functionality to thetaxx code.

The code now contains 2 new functors that are similar to the old ones, firstLaplace and updateStates. One reason is that nu_top code in xx does not have functionality to use fewer vertical levels in mpi. Another is that F code has some logic around multipliers that we want to follow to preserve bfb (thus the old functors cannot be reused).

Removing code for (nu_top>0 && hv_sub_tom==0) in the hv_subcycle loop in F code. It should not be used.

[non-BFB] for homme suites because of 1 removed test and 2 new tests (one for hypervis_subcycle>1 and the other one is for nu_top) in the bfb mode and because thetanhwet-TC used the block of code (nu_top>0 && hv_sub_tom==0), which was removed.
[NL DIFFS] for theta SMS tests that had outdated options, which this PR fixed (last commit).


oksanaguba:
This commit on weaver:
```
      Start 74: theta-f0-tt5-hvs1-hvst0-r3-qz1-nutopoff-ne2-nu3.4e18-ndays1_cxx_vs_f90
24/30 Test #74: theta-f0-tt5-hvs1-hvst0-r3-qz1-nutopoff-ne2-nu3.4e18-ndays1_cxx_vs_f90 ...........   Passed    0.63 sec
      Start 75: theta-f1-tt5-hvs1-hvst0-r3-qz1-nutopoff-ne2-nu3.4e18-ndays1_cxx_vs_f90
25/30 Test #75: theta-f1-tt5-hvs1-hvst0-r3-qz1-nutopoff-ne2-nu3.4e18-ndays1_cxx_vs_f90 ...........   Passed    0.62 sec
      Start 76: theta-f1-tt5-hvs1-hvst0-r0-qz1-nutopoff-ne2-nu3.4e18-ndays1_cxx_vs_f90
26/30 Test #76: theta-f1-tt5-hvs1-hvst0-r0-qz1-nutopoff-ne2-nu3.4e18-ndays1_cxx_vs_f90 ...........   Passed    0.62 sec
      Start 77: theta-f1-tt10-hvs3-hvst0-r3-qz1-nutopoff-ne2-nu3.4e18-ndays1_cxx_vs_f90
27/30 Test #77: theta-f1-tt10-hvs3-hvst0-r3-qz1-nutopoff-ne2-nu3.4e18-ndays1_cxx_vs_f90 ..........   Passed    0.62 sec
      Start 78: theta-f1-tt10-hvs3-hvst5-r3-qz1-nutopon-ne2-nu3.4e18-ndays1_cxx_vs_f90
28/30 Test #78: theta-f1-tt10-hvs3-hvst5-r3-qz1-nutopon-ne2-nu3.4e18-ndays1_cxx_vs_f90 ...........   Passed    0.62 sec
      Start 79: theta-f1-tt10-hvs1-hvst0-r2-qz10-nutopoff-GB-ne2-nu3.4e18-ndays1_cxx_vs_f90
29/30 Test #79: theta-f1-tt10-hvs1-hvst0-r2-qz10-nutopoff-GB-ne2-nu3.4e18-ndays1_cxx_vs_f90 ......   Passed    0.59 sec
      Start 80: theta-f1-tt10-hvs1-hvst0-r2-qz10-nutopoff-GB-sl-ne2-nu3.4e18-ndays1_cxx_vs_f90
30/30 Test #80: theta-f1-tt10-hvs1-hvst0-r2-qz10-nutopoff-GB-sl-ne2-nu3.4e18-ndays1_cxx_vs_f90 ...   Passed    0.59 sec
```

oksanaguba:
Not sure this is important but there was situation when I init-ed a HV multiplier and it had padding which was init-ed to nans by default. I think it got propagated to w, phi fields then in lines like this:
```
          w(ilev)      += wtens(ilev);
          phi_i(ilev)  += phitens(ilev);
```
where w is an interface vector and tens is a midlevels vector. I am not sure what prevented tens before with padding to not to have nans.

oksanaguba:
bfb for the same tests as above

components/homme/src/theta-l_kokkos/cxx/HyperviscosityFunctorImpl.hpp
```diff
@@ -33,14 +35,21 @@ class HyperviscosityFunctorImpl
   // TODO: don't pass nu_ratio1/2. Instead, do like in F90: compute them from
   //       nu, nu_div, and hv_scaling
   struct HyperviscosityData {
-    HyperviscosityData(const int hypervis_subcycle_in, const Real nu_ratio1_in, const Real nu_ratio2_in, const Real nu_top_in,
+    HyperviscosityData(const int hypervis_subcycle_in, 
+                       const int hypervis_subcycle_tom_in, 
+		       const Real nu_ratio1_in, const Real nu_ratio2_in, const Real nu_top_in,
```
ambrad, line 40: Please make indentation reasonably consistent with the surrounding text.

components/homme/src/theta-l_kokkos/cxx/HyperviscosityFunctorImpl.hpp
```diff
@@ -57,9 +66,12 @@ class HyperviscosityFunctorImpl
 
     bool consthv;
   };
+  //hyperviscosityData
 
+  //at compuile time we do not really know how many levels we will use
```
ambrad, line 71: If needed we can change this dimension to runtime-specified.

oksanaguba:
@ambrad -- thanks for the comments. I did not mean that I'd submit a PR with so many formatting issues, I just wanted to know whether there is something really wrong. 

I just ran with cudamemcheck and there is probably something:
```
 finished I/O sync
 Entering main timestepping loop
========= Invalid __global__ read of size 1
=========     at 0x00001a80 in _ZN6Kokkos4Impl36cuda_parallel_launch_constant_memoryINS0_11ParallelForIZN5Homme25HyperviscosityFunctorImpl3runEiddEUlRKNS0_14CudaTeamMemberEE0_NS_10TeamPolicyIJNS_4CudaEEEESA_EEEEvv
=========     by thread (31,13,0) in block (79,0,0)
=========     Address 0x47406f60 is out of bounds
=========     Device Frame:_ZN6Kokkos4Impl36cuda_parallel_launch_constant_memoryINS0_11ParallelForIZN5Homme25HyperviscosityFunctorImpl3runEiddEUlRKNS0_14CudaTeamMemberEE0_NS_10TeamPolicyIJNS_4CudaEEEESA_EEEEvv (_ZN6Kokkos4Impl36cuda_parallel_launch_constant_memoryINS0_11ParallelForIZN5Homme25HyperviscosityFunctorImpl3runEiddEUlRKNS0_14CudaTeamMemberEE0_NS_10TeamPolicyIJNS_4CudaEEEESA_EEEEvv : 0x1a80)
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame:/lib64/libcuda.so.1 (cuLaunchKernel + 0x28c) [0x3091bc]
=========     Host Frame:/home/projects/ppc64le-pwr9-nvidia/cuda/10.1.105/lib64/libcudart.so.10.1 [0x102f4]
=========     Host Frame:/home/projects/ppc64le-pwr9-nvidia/cuda/10.1.105/lib64/libcudart.so.10.1 (cudaLaunchKernel + 0x230) [0x6bf80]
```

oksanaguba:
Does 'invalid read of size 1' mean another this-> issue?

oksanaguba:
this commit is still bfb on weaver

components/homme/src/theta-l_kokkos/cxx/HyperviscosityFunctorImpl.cpp
```diff
@@ -276,6 +285,7 @@ void HyperviscosityFunctorImpl::run (const int np1, const Real dt, const Real et
       // Fix w at surface:
       // Adjust w_i at the surface, since velocity has changed
       if (m_process_nh_vars) {
```
ambrad, line 287: I think this line is causing the implicit capture of `this`.

ambrad:
I left a note at the line that I"m guessing is causing the cuda-memcheck message. But I learned last week that Summit's Cuda version now warns on implicit captures, so if that line isn't it, you could build on Summit and look for the compiler warning.

components/homme/src/theta-l_kokkos/cxx/HyperviscosityFunctorImpl.cpp
oksanaguba: Thanks -- that was it.

oksanaguba:
HV unit test probably needs to be modified to have nu_top, in the future. I won't put this into this PR.

Another test missing is tensor HV + thetaxx. Tensor is tested in preqxx, but should be added to thetaxx (another PR).

oksanaguba:
HV unit test is broken on gnu in this branch.

components/homme/cmake/machineFiles/weaver.cmake
```diff
@@ -22,25 +49,32 @@ set(HOMMEXX_VECTOR_SIZE 1 CACHE STRING "")
...
+set(Kokkos_ENABLE_CUDA_LAMBDA TRUE CACHE BOOL "")
+set(Kokkos_ARCH_VOLTA70 TRUE CACHE BOOL "")
+set(Kokkos_ENABLE_DEPRECATED_CODE FALSE CACHE BOOL "")
+set(Kokkos_ENABLE_EXPLICIT_INSTANTIATION FALSE CACHE BOOL "")
 
 # Compilers
 set(CMAKE_C_COMPILER mpicc CACHE STRING "")
 set(CMAKE_Fortran_COMPILER mpif90 CACHE STRING "")
-set(CMAKE_CXX_COMPILER mpicxx CACHE STRING "")
+######## CHANGE THAT
```
bartgol, line 79: Are you going to revert this back to mpicxx before merging?

components/homme/src/theta-l/share/prim_advance_mod.F90
```diff
@@ -789,15 +772,13 @@ subroutine advance_hypervis(elem,hvcoord,hybrid,deriv,nt,nets,nete,dt2,eta_ave_w
...
-
-
-
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
   !  sponge layer
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
   if (nu_top>0 .and. hypervis_subcycle_tom>0 ) then
   dt=dt2/hypervis_subcycle_tom
   do ic=1,hypervis_subcycle_tom
+
```
bartgol, line 781: While you're at it, would you mind fixing the indentation of the if block? It should have one extra indentation, but right now it is flushed.

components/homme/src/theta-l_kokkos/cxx/HyperviscosityFunctorImpl.cpp
```diff
@@ -65,13 +73,41 @@ void HyperviscosityFunctorImpl::init_params(const SimulationParams& params)
...
     h_nu_scale_top = Kokkos::create_mirror_view(m_nu_scale_top);
 
-    Kokkos::Array<Real,NUM_BIHARMONIC_PHYSICAL_LEVELS> lev_nu_scale_top = { 4.0, 2.0, 1.0 };
-    for (int phys_lev=0; phys_lev<NUM_BIHARMONIC_PHYSICAL_LEVELS; ++phys_lev) {
+    const auto etai_h = Kokkos::create_mirror_view(m_hvcoord.etai);
+    const auto etam_h = Kokkos::create_mirror_view(m_hvcoord.etam);
+    Kokkos::deep_copy(etai_h, m_hvcoord.etai);
+    Kokkos::deep_copy(etam_h, m_hvcoord.etam);
+
+    for (int phys_lev=0; phys_lev < NUM_LEV*VECTOR_SIZE; ++phys_lev) {
```
bartgol, line 81: Since you don't care about the padding at the end, why not looping from 0 to `NUM_PHYSICAL_LEV` at this point?

components/homme/test/reg_test/run_tests/create-theta-temporary.sh
```diff
@@ -0,0 +1,14 @@
+
```
bartgol, line 1: Curious: where is this file used?

components/homme/src/theta-l/share/prim_advance_mod.F90
```diff
@@ -659,23 +659,6 @@ subroutine advance_hypervis(elem,hvcoord,hybrid,deriv,nt,nets,nete,dt2,eta_ave_w
            stens(:,:,k,3,ie)=-nu  *stens(:,:,k,3,ie) ! w
            stens(:,:,k,4,ie)=-nu_s*stens(:,:,k,4,ie) ! phi
         enddo
-        if (nu_top>0 .and. hypervis_subcycle_tom==0) then
```
bartgol, line 662: Was nu_top kept in F90 just for allowing comparison with C++? Meaning, actual runs in F90 were already no longer using nu_top (and relying only on hv_subcycle_tom)?

components/homme/src/theta-l_kokkos/cxx/HyperviscosityFunctorImpl.cpp
```diff
@@ -65,13 +73,41 @@ void HyperviscosityFunctorImpl::init_params(const SimulationParams& params)
...
+        if ( etai_h(0) == 0.0) {
+          ptop_over_press = etam_h(0)[0] / etam_h(ilev)[ivec];
+        }else{
+          ptop_over_press = etai_h(0) / etam_h(ilev)[ivec];
+        }
+      }else{
+          ptop_over_press = 0.0;
+      }
+
+      auto val = 16.0*ptop_over_press*ptop_over_press / (ptop_over_press*ptop_over_press + 1);
```
bartgol, line 101: Maybe it could help to add a comment that explains this line? Like, where does 16.0 come from (deduced from some params, or set after trial/error), or where is the x^2/ (x^2+1) (x=p0/p) coming from (e.g., deduced from some principle/formula, or an arbitrary fcn that has certain props).

components/homme/src/theta-l_kokkos/cxx/HyperviscosityFunctorImpl.hpp
```diff
@@ -149,7 +150,42 @@ class HyperviscosityFunctorImpl
         });
       }
 #endif
-    });
+    }); //team thread range
+
+    // Laplacian of layer thickness
+    m_sphere_ops.laplace_simple(kv,
```
bartgol, line 156: I _think_ you might need a team barrier before this point, since you must ensure the subtraction of the ref states is complete before you attempt to take its grad inside the laplace simple routine. Or, better, you could start with the laplace of something not involved in the ref states above, like the lapl of V, so we save ourselves a barrier.

components/homme/src/theta-l_kokkos/cxx/HyperviscosityFunctorImpl.hpp
```diff
@@ -175,7 +211,45 @@ class HyperviscosityFunctorImpl
     m_sphere_ops.vlaplace_sphere_wk_contra(kv, m_data.nu_ratio1,
                               Homme::subview(m_state.m_v,kv.ie,m_data.np1),
                               Homme::subview(m_buffers.vtens,kv.ie));
-  }
+
+    Kokkos::parallel_for(Kokkos::TeamThreadRange(kv.team,NP*NP),
```
bartgol, line 215: I was gonna say that we need a team barrier before this TTR for too, but turns out that all our sph ops have a team barrier at the end. I would argue that we should remove the team barrier at the end of our sph ops (the barrier may not be necessary until later on), but it should not be part of this PR.

components/homme/src/theta-l_kokkos/cxx/HyperviscosityFunctorImpl.hpp
```diff
@@ -278,25 +351,89 @@ class HyperviscosityFunctorImpl
...
+      if (m_process_nh_vars) {
+        wtens   = Homme::subview(m_buffers.wtens,kv.ie,igp,jgp);
+        phitens = Homme::subview(m_buffers.phitens,kv.ie,igp,jgp);
+        w       = Homme::subview(m_state.m_w_i,kv.ie,m_data.np1,igp,jgp);
+        phi_i   = Homme::subview(m_state.m_phinh_i,kv.ie,m_data.np1,igp,jgp);
+      }
+
+      Kokkos::parallel_for(Kokkos::ThreadVectorRange(kv.team,NUM_LEV),
+                           [&](const int ilev) {
+        utens(ilev)   *= rspheremp;
```
bartgol, line 418: @ambrad Do you think it might be beneficial to add fmad-like operators/functions updates for our Pack/Vector classes, to allow doing
```
u(ilev) = u(ilev) + utens(ilev)*rspheremp
```
as a single fmad op, rather than have one loop for the `*= rspheremp` part, and one loop for the `+=` part? I doubt the compiler can fuse those two loops into one, and fuse + and * (either on CPU or GPU).

bartgol (commented):
I think this looks good. Does this option get tested in the hv_ut unit test too? If not, are we planning on doing it?

Edit: nvm, I just read your last comment.

components/homme/src/theta-l_kokkos/cxx/HyperviscosityFunctorImpl.cpp
bartgol: Ah, I see, you want to avoid NaN's in the padding. You could still loop to NUM_PHYSICAL_LEV (avoiding the if in the block), and then _after_ do a for loop on `[NUM_PHYSICAL_LEV,NUM_LEV*VECTOR_SIZE)`, to zero out the padding. Tbc, this is not for perf, it's just to make the body of the for loop a tad less involved (by removing the if/else)

components/homme/cmake/machineFiles/weaver.cmake
oksanaguba: no, why? this is more to draw attn. so that someone else points to their mvcc_wrapper.

components/homme/src/theta-l/share/prim_advance_mod.F90
oksanaguba: I tried to not to touch F code in this PR, unless you really insist.

components/homme/test/reg_test/run_tests/create-theta-temporary.sh
oksanaguba: This was committed by a mistake, I was going to remove it in the last round of cleaning. This file is created temporarily by make when re-making a list of theta-l xx tests. However, that set of scripts is not flexible enough and needs updating (maybe in another PR).

components/homme/src/theta-l/share/prim_advance_mod.F90
oksanaguba: I am not sure what you mean, but this particular code was in F version is to save some mpi rounds by tying them with mpi rounds in general HV. This code is not used, so, I removed it to not to confuse us to test it. I removed similar code from xx version too. 

Overall, we need nu_top  and hv_subcycle_tom in both codes.

components/homme/src/theta-l_kokkos/cxx/HyperviscosityFunctorImpl.hpp
oksanaguba: Great, thanks, will do.

components/homme/src/theta-l_kokkos/cxx/HyperviscosityFunctorImpl.hpp
oksanaguba: I am confused, I do not see 2 loops there. I probably broke off + and * just to make it bfb, without knowing that I cannot do it in 1 line on vectors.

components/homme/cmake/machineFiles/weaver.cmake
bartgol: Because the compiler should always be mpicxx. It's the env var `OMPI_CXX` (or MPICH_CXX for mpich distros) that should point to nvcc_wrapper. Setting the compiler to be nvcc_wrapper can be problematic for MPI.

components/homme/src/theta-l/share/prim_advance_mod.F90
bartgol: I just thought "while you're at it...".  But I don't really insist.

components/homme/src/theta-l_kokkos/cxx/HyperviscosityFunctorImpl.hpp
bartgol: Sorry, I wasn't clear. Your code is fine. Separate the * and the + is the way to do it, _given our Vector class impl_. The loops I referred to are _hidden_ inside the Vector class. When you do `a *= b`, with `a` being a Vector, the overload of `*=` includes a for loop over the Vector length. So, when we do
```
utens(ilev) *= rspheremp;
u(ilev) += utens(ilev);
```
we are doing 2 for loops. And I doubt that the compiler can fuse those loops into one, and merge the *= with the +=, into a single fmad op.

bartgol: I'm actually quite sure the compiler cannot do FMA here, since it cannot know that we are not really interesting in storing `utens*rspheremp` inside utens, so it must keep the `*=` separate from the `+=`.

TBC: this impl is what we are currently doing throughout Hommexx. I am just wondering if there is room for a micro opt here.

ambrad: I haven't looked whether `u(ilev) += rspheremp*utens(ilev)` gets compiled to an FMA. I think it's possible with the ekat pack impl but unlikely with the Hxx one. Not reason not to add an explicit FMA routine.

components/homme/src/theta-l_kokkos/cxx/HyperviscosityFunctorImpl.hpp
oksanaguba: the code that shows as new on GitHub, girstLaplaceHV, is old, but I put a team barrier there, too (as well as in nutopLaplace), thanks.

oksanaguba:
@ambrad and @bartgol  -- unit test of HV needs many changes: nu needs to be different for tensor and const HV, init should  use init of the HV functor, nu_top option needs to be on (off right now cause it would require major changes), F init needs to be changes due to the all above, other variables, like dpdiss, probably need to be added to the test. Would it be ok if I leave these major changes for another PR?

ambrad:
Does hv_ut as it is right now pass? If so, then waiting to add more to it is fine with me.

oksanaguba:
Yes, it does on weaver (waiting for maintainability to be over for anvil/chrysalis to check there).

oksanaguba:
in next

PR #4574, opened 2021-10-04
Fix parameter dimension size error for plant hydraulics

pnlfang:
Due to the size change of mxpft, the 2D parameters related to plant hydraulics are not read in correctly. A fix is made in this branch.

[BFB]


hydrotian:
This is an important bug fix PR needed for the upcoming BGC spin-up runs (starting Nov. 15th). Could you please try to merge it to the master? Thanks! @peterdschwartz

PR #4580, opened 2021-10-05
Fix quotes to agree with XML format

### LINKED ISSUE #4579: XML Issue in mpas-ocean Registry_CFC.xml
There is a minor XML issue in the Registry_CFC.xml file in mpas-ocean. It does not create any run problems, but causes the XML parser we use to autogenerate bld files to fail

jonbob:
Fixes a minor XML issue in the Registry_CFC.xml file in mpas-ocean. This will allow the XML parser we use to autogenerate bld files to work.

Fixes #4579 

[BFB]


jonbob:
passed sanity testing, merged to next

PR #4583, opened 2021-10-05
fixed allocate error in ocean del4 tracer mixing

### LINKED ISSUE #4581: ocean fails with memory corruption error when tracer biharmonic mixing enabled
With tracer del4 mixing turned on, a QU240 standalone simulation fails with a memory corruption error. This occurs on both Summit with nvhpc and on Cori-KNL with Intel.

philipwjones:
fixes a bug in which a temp was being allocated before one of the dimensions was defined, causing a malloc corruption error

Because this was a pretty egregious error, it's unlikely any of our current tests actually exercise this code and I could only verify it runs to completion and gives bit-for-bit when del4 is off. 

Fixes #4581


jonbob:
Passes all e3sm_developer tests with active ocean on chrysalis:
* SMS_P12x2.ne4_oQU240.WCYCL1850NS.allactive-mach_mods

jonbob:
merged to next

PR #4585, opened 2021-10-06
Enable C++ version of RRTMGP in E3SM and E3SM-MMF

brhillman:
Enable using the C++ back end of RRTMGP for both E3SM and E3SM-MMF. The C++ code uses YAKL to enable running on CPU or GPU with the same codebase. Maintains the ability to run the F90 back end as well (remains the default). Building with the C++ version is enabled by adding `-rrtmgpxx` to `CAM_CONFIG_OPTS` in `env_build.xml`. Note that this is in addition to using the `-rad rrtmgp` option. The two flags are required because `-rad rrtmgp` sets much of the logic that determines input data and settings in `configure` and `build-namelist`, which we do not want to duplicate just because we are running the C++ back end instead of the F90 back end. A testmod (`eam-rrtmgpxx`) is also added so that RRTMGPXX can be built via `create_test` without any user interaction. Also updates both the YAKL and RRTMGP submodules. Believe it or not, this is BFB.

[BFB]


rljacob:
Does this also update the submodule versions of YAKL and RRTMGP?  mention that in the PR description.

brhillman:
@mrnorman RRTMGP repo PR merged, and updated this branch to use current RRTMGP master.

brhillman:
Since the GPU code is not BFB with the CPU code, I did a poor-man's perturbation growth test by running the CPU code with `-O0` and `-O3` optimizations and comparing the RMS difference in mass-weighted temperature with that between a GPU test and the `-O3` baseline on the CPU:
perturbation_growth

After a day, the GPU test does not appear to be diverging from the CPU test any faster than would be expected from differences in optimization levels, so it would appear that the solution with the GPU code is consistent with the CPU code to within machine precision.

PR #4586, opened 2021-10-07
Modify harmonic analysis AM namelist options and output stream

sbrus89:
The start and end dates for harmonic analysis period have been replaced by a `config_AM_harmonicAnalysis_start_delay` and a `config_AM_harmonicAnalysis_duration` option (with units of days). This makes it easier to change the start/end date of the simulation without having to change harmonic analysis AM options.

Stream output is now controlled by the AM compute routine. The HA output stream is written following the least squares solve. The `output_interval` in the streams file should be set for a long time interval (i.e. 1000 years) so it is not written by the AM driver routine. This means the `output_interval` no longer needs to be changed when the simulation length or harmonic analysis period changes.

[NML]
[BFB]


knbarton:
I think it looks good! Would it make sense to add in a warning if the analysis is set to go beyond the run duration? It looks like this is the only mistake that could still pop up.

jonbob (approved):
Looks good from a visual inspection and I'll test during the merge process

sbrus89:
@jonbob, would you like me to squash this?

jonbob:
@sbrus89 - only if you think it would be clearer. Did my commit of the bld files make it messy?

sbrus89:
@jonbob, not at all. I was just wondering. I'll leave it as is.

jonbob:
passed all e3sm_developer tests with active ocean on chrysalis, merged to next

jonbob:
merged to master, expected NML DIFFs blessed

PR #4587, opened 2021-10-07
Add MARBL ocean BGC library

maltrud:
This PR adds the MARBL ocean Biogeochemistry library to MPAS-Ocean.  this replaces the current BGC routines which were a direct port of the BEC routines from CESM1, so the E3SMv1 implementation will no longer be supported.  this will allow E3SM to stay more up to date with BGC improvements made by the community.  in E3SM, MARBL is a submodule (E3SM-Project/MARBL) which was forked from the public MARBL repository (marbl-ecosys/MARBL) and set to the CESM2.1 tag.  there are several important notes about this PR:
1) ocean BGC will not run “out of the box” since a new MARBL specific initial condition file is required.  this will be added to the config system once full validation has been finished.
2) MARBL is not yet compatible with sea ice BGC due to the addition of refractory DOC.  updates to the interface with sea ice BGC will happen soon.
3) MARBL does not simulate Phaeocyctis as a phytoplankton functional type.
4) there is an unsolved issue with the ALT_CO2 tracer that can cause MARBL to post error warnings.  this tracer is essentially a diagnostic so does not affect the simulation.  this will be sleuthed out soon.
5) testing has been done using the EC30to60E2r2 grid.  comparisons with the v1 BGC after 10 simulated years (using identical forcing) show no obvious problems.  exact restart and independence of solution with respect to processor layout have been verified on anvil with this setup.
6) it isn't clear yet how to interface with MARBL within an ice shelf cavity, so this PR does not support land ice grids.

[BFB]


philipwjones:
@mark-petersen I looked into the MARBL build recently (Mat had asked me some MARBL questions and I tried to build).  They have a custom build where we can pass the compiler and opts from the E3SM build and use those. Unfortunately, that custom build is broken in their current makefile and won't work, but that's the appropriate approach. I think I can kludge it by creating a version in our space and over-writing theirs before building. That's probably a better option until we can get them to fix their Makefile.

philipwjones:
@mark-petersen Yes, the cmake build ignores the makefile and just builds the sources, so this only impacts the standalone build with our own makefile. And as you noted, their make builds if you specify their support compilers but it dumps things in compiler-named subdirectories.  The custom build allows you to specify not only the compiler and compiler options, but where you want to put stuff. Unfortunately, they have a bug in there that assumes a bad location for the .mod files so stuff doesn't build or clean correctly.  So I was thinking I could fix that bug and put a Makefile.marbl in the top level mpas-ocean space and have our Makefile copy that version into the marbl space just before building.  Meanwhile, I probably need to submit an issue to their repo to fix too...

maltrud:
thanks for the explanation @philipwjones .  since we have our own MARBL (E3SM-Project/MARBL) we can make whatever changes we want to the Makefile without affecting the marbl-ecosys repo controlled by ncar.  not that we shouldn't let them know, but we don't need them to be involved in our modifications.

philipwjones:
@mark-petersen Ahh. ok. I though we were pulling from theirs. I'll go ahead and work out the fixes and get it into our own version.  Unfortunately, I have to quit for now, but I'll try to get something in ASAP.

philipwjones:
@maltrud @mark-petersen  Well...I actually mis-diagnosed the problem.  Combination of user error and a bad "make clean". But I now have a working standalone build. Will push changes/PRs in a sec...

mark-petersen:
> @mark-petersen, could you tell me if you modified any namelist options when you ran these?

@xylar, you can see my run here on IC
```
/lustre/scratch3/turquoise/mpeterse/runs/n/ocean_model_211011_62ffe688_bgc_on/ocean/global_ocean/QU240/PHC_BGC/performance_test/forward
```
Looking at the log file, I see the same thing as you:
```
1043   could not find MARBL interior diagnostic index for photoC_TOT_zint_100m
1044   could not find MARBL interior diagnostic index for photoC_NO3_TOT_zint_100m
1045   could not find MARBL interior diagnostic index for DOC_prod_zint
...
1212   could not find MARBL surface diagnostic index for NHy_FLUX
1213   could not find MARBL surface diagnostic index for NHx_SURFACE_EMIS
```
so 170 lines. We'll have to look at that. It's not a stream file problem.

maltrud:
@mark-petersen , @xylar please don't worry about those messages--they don't matter.

jonbob:
merged to next

jonbob:
merged to master, but expected DIFFs have not yet been blessed because those tests did not run overnight

PR #4597, opened 2021-10-12
CIME Update

jgfouca:
to a26c8e393d4e062459b6dfb9605700fe24665e50

Changes
1) Remove MODEL as a setting for the build system (Use COMP_NAME)
2) remove compare vars dependency in cprnc
3) extend the recent cmake macro changes to allow the inclusion of CMAKE_LANG_COMPILER_VERSION in cmake_macros files
 - (Should no longer need intel17, intel18 compiler names in macros!)
4) cime_bisect: Improve robustness when using custom script

Fixes:
1) Using custom cmake macros in `~/.cime`
2) Fix typo in scripts/Tools/Makefile

[BFB]


rljacob:
How do we use CMAKE_LANG_COMPILER_VERSION to allow building cases with two different versions of the same compiler?

jgfouca:
@rljacob , in the `$compiler_$mach.cmake` macro file for you machine/compiler, you can do conditional statements based on the compiler version, for example, in gnu_$mach.cmake, you could have:

```
   string(APPEND FFLAGS " -fallow-argument-mismatch  -fallow-invalid-boz ")
 endif()
```

PR #4599, opened 2021-10-14
[wip] add energy diagnostics in atm

oksanaguba:
The first step to better diagnose energy:

Compare dTE versus fluxes each physics time step using globally integrated quantities. 

Use post processing python the same way it is done for mass conservation.

Adding 1 test for energy that will probably fail for now, till we fix minor leaks from clubb, zm, mg2. 
Or I can put tolerance big enough that the test will pass.

[bfb] but a new SMS test (no new build)


rljacob:
From telecon:  still need to verify.  May not go in.

oksanaguba:
Things to do:
Init new variables with nans and init them then properly with zeros at init stage.
Possibly check that new RR diagnostics matches RR diagnostics from python scripts.
Implement a new limiter to show that RR in this code is really zero then.
Sort out restart runs.

Mainly, I do not know how to do the last item yet. Closing it for now, would be easy to re-open and finalize it.

oksanaguba:
Another item to think about -- surface stresses are not energy conserving. I do not remember where they are applied (not in clubb, I think), thus they need to be 'fixed' as well.

oksanaguba:
reopening

rljacob:
Re: ifdef or if statements.  Which piece of code would require this?

oksanaguba:
@rljacob all or almost all added lines in this PR need to be wrapped in something. Otherwise this diagnostics (with a lot of new arrays) runs each timestep.

rljacob:
Is it expensive in terms of memory or performance?

rljacob:
Wrap all new code with ifdefs?   See how much slower it is and if there's no difference, just leave it as is.

components/eam/src/physics/cam/physpkg.F90
```diff
@@ -1691,28 +1693,52 @@ subroutine tphysac (ztodt,   cam_in,  &
     nstep = get_nstep()
     call check_tracers_init(state, tracerint)
 
+!!!!!!! OG
+#if 1
```
rljacob, line 1697: @oksanaguba is this part you want to put in a conditional?

oksanaguba: yes, this is one of few places with new code.

i will get back to this PR, i need to fix an ERS test with new diagnostics.

oksanaguba:
Following the meeting discussion: this PR does global water and energy diagnostics, we can add carbon here as well.

rljacob:
update:  working on restart of diagnostics.

oksanaguba:
i still need to tune tolerances in the PY script and wrap allocates for new diagnostics in if-statements, otherwise, it is ready for a review.

oksanaguba:
Tagging @wlin7 @rljacob @chengzhuzhang for possible preliminary review.

oksanaguba:
Just a note, I have been running an ERS_D test with this branch and diagnostics on, and all looks as expected. I will run other suites, too.

oksanaguba:
To run python script manualy, one should copy it into the config folder and do
```
python check_energy_conservation.py ./
```

rljacob:
@wlin7 can you approve?

wlin7 (approved):
Changes look good and are useful. Tests (in next) confirmed all are fine.

oksanaguba:
i ran atm_developer and adjusted tolerances in the new test, because of a fail in it. the tolerances work for chrysalis.

in next.

PR #4613, opened 2021-10-19
Mosart dam bug fix

hydrotian:
Fixing a bug related to MOSART water management scheme. Move the
dam looping outside of the active stage checking.


tanzeli1982 (approved):

rljacob:
@hydrotian which cases will this affect?

hydrotian:
This only affects the cases that turn on MOSART water management feature. Should be BFB as none of the land testing case turns water management on.

rljacob:
Do we have any test cases that turn on the MOSART water management feature?

hydrotian:
So far we don't. I plan to create a test case that turns on all the water related features developed for V2 BGC, after this PR is merged.

PR #4617, opened 2021-10-22
Add Fugaku machine config

amametjanov:
Add Fugaku machine configuration with gnu/11.2.0 and fujitsu/4.7.0 compilers.

[BFB]


amametjanov:
I had to add to `externals/scorpio`
```
diff --git a/src/clib/pio.h b/src/clib/pio.h
index 85b55075..b7f0df77 100644
--- a/src/clib/pio.h
+++ b/src/clib/pio.h
@@ -65,7 +65,7 @@ unsigned long get_adios2_io_cnt();
 /** PIO_OFFSET is an integer type of size sufficient to represent the
  * size (in bytes) of the largest file supported by MPI. */
 #define PIO_OFFSET MPI_OFFSET
-#define PIO_Offset MPI_Offset
+#define PIO_Offset long long

 /** The start ID and maximum number of IDs for IO decompositions. */
 #define PIO_IODESC_START_ID 512
diff --git a/tools/CMakeLists.txt b/tools/CMakeLists.txt
index 41eb658e..ad0ed621 100644
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -8,4 +8,4 @@ if (WITH_ADIOS2)
     ADD_SUBDIRECTORY(adios2pio-nm)
   endif (ADIOS2_FOUND)
 endif(WITH_ADIOS2)
-ADD_SUBDIRECTORY(spio_finfo)
+#ADD_SUBDIRECTORY(spio_finfo)
```
for mpich-tofu, because it doesn't appear to provide MPI-IO (no pnetcdf) and some MPI data types.

PR #4622, opened 2021-10-26
Work around a Cuda issue.

ambrad:
Cuda 10.1.105 with gcc 8.5.0 erroneously reports an error for a certain line. Details are in SCREAM issue 1277. This PR works around the issue. I studied the issue in part using `-Wall -pedantic` with various compiler versions and thus have also fixed some miscellaneous warnings in SL and physgrid remap code.

[BFB]


oksanaguba:
in next

PR #4633, opened 2021-11-02
Hsiangheleellnl/atm/o3 in linozv3

hsiangheleellnl:
Pass real O3 into Linoz v3. In the Linoz code, variables named "_e3sm" are associated with real O3.


hsiangheleellnl:
will be handled by https://github.com/E3SM-Project/E3SM/pull/4635. Close it now.

PR #4636, opened 2021-11-03
Homme: Fix GLL <-> FV remap input/output for ftype=0

### LINKED ISSUE #4632: gllfvremap_mod takes wrong input for q when ftype is 0
ftype 0 is not used in practice in EAM, but it is supported. @bartgol pointed out to me that dp_coupling and gfr_fv_phys_to_dyn do not agree on the input for q when ftype is 0. This needs to be fixed. I will also add an EAM test for pg2 and ftype 0 so that this code branch gets exercised.

ambrad:
ftype 0 is not used in practice in EAM, but it is supported. @bartgol pointed out to me that dp_coupling and gfr_fv_phys_to_dyn do not agree on the inputs and outputs for q and FQ when ftype is 0. This PR fixes that. It also adds two CIME tests for ftype 0 and pg2.

Fixes #4632.

[BFB] except for **two new CIME tests** (no baseline) and Homme standalone test **thetah-sl-dcmip16_test1pg2** (diff)


ambrad:
I ran F2010-CICE.ne30pg2_ne30pg2 for 6 years with default ftype and ftype 0. Results are quite similar:
* E3SM diagnostics [viewer](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.ambradl/pg-ftype-0-vs-2/viewer/)
* Summary [figure](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.ambradl/pg-ftype-0-vs-2/summary-0002-0006.png)

We don't actually care that much about ftype 0, as I understand it, but my purpose here was to check that the new CIME tests make sense. They do.

ambrad:
Marked WIP until I finish the rest of the testing.

ambrad:
Matching the expected results, e3sm_developer passes on Chrysalis against today's baselines except for the two new tests; these diff against baselines, and the new ERS test passes COMPARE_base_rest.

ambrad:
On weaver gllfvremap_ut passes in debug/BFB and release builds.

ambrad:
The Homme standalone test thetah-sl-dcmip16_test1pg2 looks good.

ambrad:
@bartgol I asked you to review this PR, but you don't need to do a thorough review of the code. The main points are:
* gllfvremap_ut passes on weaver, showing Kokkos and F90 remap versions are synced;
* there is now a set of CIME tests using pre-existing builds (so testing time increases only a little) that test pg2 with ftype 0;
* ne30pg2 F-case climo for ftype 0 vs 2 looks good.

Thanks again for pointing out the ftype 0 discrepancy in inputs/outputs.

components/homme/src/share/gllfvremap_mod.F90
```diff
@@ -448,54 +442,28 @@ subroutine gfr_fv_phys_to_dyn_hybrid(hybrid, nt, dt, hvcoord, elem, nets, nete,
...
-                do k = 1,nlev
-                   elem(ie)%derived%FQ(:,:,k,qi) = wf1(1,k)
-                end do
-             end if
-             ! Get limiter bounds.
-             do k = 1,nlev
-                gfr%qmin(k,qi,ie) = minval(q(:nf2,k,qi,ie))
-                gfr%qmax(k,qi,ie) = maxval(q(:nf2,k,qi,ie))
-             end do
+          ! FV Q_ten
```
bartgol, line 445: So the fv->gll remap routine now does the same operation, regardless of ftype? Namely, it seems that now you do `dQ_fv = Q_new_fv - g2f(Q_old_gll)`, and then set `FQ=Q_old_gll + f2g(dQ_fv)` (at least for pg2 case). So FQ always contains a tracers value. But looking at applyCAMforcing_tracers, for ftype=0, FQ should contain a tendency. What am I missing?

bartgol (commented):
Thanks, very big PR. Sorry you had to to so much work to fix a use case that will likely never be used for production runs. I only have one question (see inline comment).

components/homme/src/share/gllfvremap_mod.F90
ambrad: Right, it turns out gllfvremap_mod doesn't need to know ftype. stepon.F90 does the conversion from state to tendency, starting at this line: https://github.com/E3SM-Project/E3SM/blob/3a87fa1738e1bfb91fb84acbf2cadbf88dddcef1/components/eam/src/dynamics/se/stepon.F90#L383

bartgol: Ah, gotcha, thanks. This makes sense.

bartgol (approved):

oksanaguba:
in next

PR #4643, opened 2021-11-04
add -ffpe-summary=none to stand-alone MPAS Makefile for gnu

mark-petersen:
Currently, gnu-compiled executables produce a summary table of exceptions at the end, which is some version of:
```
Note: The following floating-point exceptions are signalling: IEEE_DENORMAL
Note: The following floating-point exceptions are signalling: IEEE_INVALID_FLAG IEEE_UNDERFLOW_FLAG IEEE_DENORMAL
Note: The following floating-point exceptions are signalling: IEEE_INVALID_FLAG IEEE_DENORMAL
```
and prints one line per processor on finalize for both debug and optimized runs. When running on hundreds of cores, this is a headache.

Looking at on-line material [1](https://wiki.usask.ca/display/MESH/IEEE+errors+using+GNU+compiler) [2](https://fluka-forum.web.cern.ch/t/the-following-floating-point-exceptions-are-signalling/593) [3](https://stackoverflow.com/questions/39350816/floating-point-exceptions-are-signalling-in-new-gfortran-version) other modeling groups see this with gnu 5+ as well, and say to ignore it or add `-ffpe-summary=none`. See [gnu compile flags](https://gcc.gnu.org/onlinedocs/gfortran/Debugging-Options.html).

I have had numerous students and post-docs tell me they had errors when they see this. The error messages are confusing and not useful, because there is no back-trace. Note that actual IEEE errors are caught during debug with `-ffpe-trap=invalid,zero,overflow` and include a back-trace. The `-ffpe-trap` is the normal way we catch the errors.

This affects MPAS-seaice and MALI as well. This Makefile is not used for E3SM compiling.

[BFB]

### DIFF
--- a/components/mpas-framework/Makefile
+++ b/components/mpas-framework/Makefile
@@ -458,7 +458,7 @@ gnu-nersc:
 	"CFLAGS_OPT = -O3 -m64" \
 	"CXXFLAGS_OPT = -O3 -m64" \
 	"LDFLAGS_OPT = -O3 -m64" \
-	"FFLAGS_DEBUG = -g -m64 -ffree-line-length-none -fconvert=big-endian -ffree-form -ffpe-summary=none" \
+	"FFLAGS_DEBUG = -g -m64 -ffree-line-length-none -fconvert=big-endian -ffree-form -ffpe-trap=invalid,zero,overflow -ffpe-summary=none" \
 	"CFLAGS_DEBUG = -g -m64" \
 	"CXXFLAGS_DEBUG = -g -m64" \
 	"LDFLAGS_DEBUG = -g -m64" \
--- a/components/mpas-framework/Makefile
+++ b/components/mpas-framework/Makefile
@@ -335,11 +335,11 @@ gfortran:
 	"CC_SERIAL = gcc" \
 	"CXX_SERIAL = g++" \
 	"FFLAGS_PROMOTION = -fdefault-real-8 -fdefault-double-8" \
-	"FFLAGS_OPT = -O3 -m64 -ffree-line-length-none -fconvert=big-endian -ffree-form" \
+	"FFLAGS_OPT = -O3 -m64 -ffree-line-length-none -fconvert=big-endian -ffree-form -ffpe-summary=none" \
 	"CFLAGS_OPT = -O3 -m64" \
 	"CXXFLAGS_OPT = -O3 -m64" \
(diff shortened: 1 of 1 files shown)


mark-petersen:
Tested `pr` regression suite on grizzly with `make gfortran` and cori with `make gnu-nersc`. Passes `pr` suite for both debug and optimized on both. Ran some QU240 and baroclinic channel forward runs with before/after executables, and could see that summary table of IEEE exceptions was removed.

ndkeen:
Ah. We might want to use this flag in other places as well. Thanks Mark.

mark-petersen:
In this process, I found that `gnu-nersc` is not using `-ffpe-trap=invalid,zero,overflow` like the other gnu compilers in the Makefile. I have to think that was an oversight. I added an intentional divide by zero, and indeed the current settings don't produce an error, but with this last commit, we get a floating-point exception and trace-back, as expected.

philipwjones (approved):
Can you also please add the flag to the gfortran-clang target?

I built/ran a version and confirm that it ran successfully and did not generate the usual underflow messages.  They were kinda annoying so thanks @mark-petersen

mark-petersen:
Added `-fbounds-check -fbacktrace` to gnu-nersc debug flags, so it matches gfortran. I tested on cori, including with and without an intentional divide by zero, and it all works as expected. Also added `-ffpe-summary=none` to gfortran-clang per @philipwjones's request.

@matthewhoffman and @akturner if you could run a standard test on MALI and MPAS-seaice with this Makefile, I think that is enough to test this PR.

mark-petersen:
@xylar I think additional testing from you is not required. You can read my testing above. You've been involved with Makefile changes in the past, so wanted you to see this PR.

xylar (approved):
This looks good to me and passed the `nightly` regression suite on my laptop with gnu compilers.  I see:
```
Note: The following floating-point exceptions are signalling: IEEE_DENORMAL
Note: The following floating-point exceptions are signalling: IEEE_UNDERFLOW_FLAG IEEE_DENORMAL
```
without these changes but not with them.  Thanks for this work!

matthewhoffman (approved):
I tested on Cori with gnu using MALI, and ran into no problems.  It will be nice to eliminate those confusing output messages!

akturner (approved):
Passed MPAS-Seaice testing

jonbob:

jonbob:
merged to master

PR #4651, opened 2021-11-09
Progress on make macro linting tool

jgfouca:
Develop a tool to clean up CMake macro files and apply it to current files.  Clean up a lot of problems.

[BFB]


jgfouca:
@rljacob , wondering if you have thoughts on this topic. Is it worth it to try to clean/curate our Macros?

jgfouca:
Looks like I was able to remove about 350 lines of CMake macros.

rljacob:
What does "Implement MIS warning" mean?

jgfouca:
@rljacob , this new tool, `evaluator`, is designed a bit like pylint. You can enable/disable certain warnings by using abbreviated codes to refer to specific warning types. In this case `MIS` refers to warning "Missing: Having a machine/comp in config_machines.xml with no specialization in Macros".

rljacob (approved):

jgfouca:
I should note, this work will be the first major divergence between config_compilers.xml and the cmake macros. It would be incredibly tedious to try port all these macro changes back to config_compilers.xml.

rljacob:
Then you should probably delete config_compilers so people don't try to edit it?

jgfouca:
@rljacob , if you think it's already ready to be removed, I'd be happy to remove it.

rljacob:
Can people still go forward?  If they add a machine to config_compilers, can you or someone easily generate the Macros ?

jgfouca:
@rljacob , any changes to config_compilers.xml would have to be manually ported to the macros since the XML and macros are no longer in sync. I don't think it would be too hard to add a reverse-convert capability (macro->XML) so I can look into that as a way to get back in sync.

rljacob:
No we don't need a reverse and manual entry would be fine.

Issue #4645, opened 2021-11-08
MVK test with F compsets using mpas sea ice


mt5555:
MVK test uses the cime ensemble capability and runs 30 ensemble members.  If one configures a task with each ensemble using 24 cores, we have NTASKS=720, and NTASKS_PER_INST=24.   This causes MPAS-I to look for a decomposition file for 720 tasks instead of 24 tasks.   If the 720 file doesn't exist, ./create_test fails.   

Example:

./create_test MVK_P24x1.ne4_oQU240.F2010 -q short -o -g -b master --baseline-root ~/scratch1/baselines/intel

Error:
```
inished RUN for test MVK_P24x1.ne4_oQU240.F2010.compy_intel in 3.962719 seconds (FAIL). [COMPLETED 1 of 1]
    Case dir: /compyfs/tayl519/e3sm_scratch/MVK_P24x1.ne4_oQU240.F2010.compy_intel.G.20211108_112453_ru4oyl
    Errors were:
        env_batch.xml appears to have changed, regenerating batch scripts
        manual edits to these file will be lost!
        
        Model elm no file specified for finidat
        Model elm no file specified for finidat
        wget failed with output:  and errput --2021-11-08 11:36:25--  https://web.lcrc.anl.gov/public/e3sm/inputdata/ice/mpas-cice/oQU240/mpas-cice.graph.info.151209.part.720
        Resolving web.lcrc.anl.gov (web.lcrc.anl.gov)... 140.221.70.30
        Connecting to web.lcrc.anl.gov (web.lcrc.anl.gov)|140.221.70.30|:443... connected.
        HTTP request sent, awaiting response... 404 Not Found
        2021-11-08 11:36:25 ERROR 404: Not Found.
        
        Model elm no file specified for finidat
        svn export failed with output:  and errput svn: E170000: URL 'https://svn-ccsm-inputdata.cgd.ucar.edu/trunk/inputdata/ice/mpas-cice/oQU240/mpas-cice.graph.info.151209.part.720' doesn't exist
        
        ERROR: Could not find all inputdata on any server

```

relevant section from env_mach_pes.xml file: (from copy)

```
    <entry id="NTASKS">
      <type>integer</type>
      <values>
        <value compclass="ATM">720</value>
        <value compclass="CPL">720</value>
        <value compclass="OCN">720</value>
        <value compclass="WAV">720</value>
        <value compclass="GLC">720</value>
        <value compclass="ICE">720</value>
[...]

PR #4654, opened 2021-11-10
fix component namelist creation bugs when NINST>1

### LINKED ISSUE #4648: PGN test not configured for V2
Trying to run the PGN test with V2 results in configure errors with missing data sets.  This test case appears to rely on data that needs to exist in the clm2/initdata/"gridname"_v2_init/ directory.   This data only exists for v1, and the ne4_ne4 grid, running with the FC5AV1C-04P2 compset.   It looks like new data needs to be generated for v2, and for a v2 supported compset.  Instead of ne4_ne4, it might be better to use ne4_oQU240, to allow for the eventual use of F compsets using MPAS-I.  

As this test only runs for 1 timestep, why not use ne30?

Example:
```
./create_test PGN_P24x1.ne4_oQU240.F2010-CICE -q small -g -b master --baseline-root ~/scratch1/baselines/intel
...

        wget failed with output:  and errput --2021-11-08 17:18:30--  https://web.lcrc.anl.gov/public/e3sm/inputdata/lnd/clm2/initdata/ne4_oQU240_v2_init/20210915.v2.ne4_oQU240.F2010.elm.r.0002-06-01-00000.nc
```

mt5555:
A few bug fixes when running with NINST>1, as is done in the E3SM nbfb test suite

MPAS sea ice component should use NTASKS_PER_INST_ICE instead of NTASKS_ICE when determining which partition file to use

Ensure ATM and LND put the initial conditions from all ensemble members into the inputdata file list 

Similar fixes should probably be applied to the MPAS ocean (NTASKS_PER_INST_OCN) and ensemble member inputdata parsing for both MPAS ocean and ice.  Not done in this PR since we currently dont use this capability.  

fixes #4648, #4645

[BFB]

### DIFF
--- a/components/eam/bld/build-namelist
+++ b/components/eam/bld/build-namelist
@@ -4772,7 +4772,7 @@ sub check_input_files {
 
     my $fh;
     if (defined $inputdata_rootdir) {
-	open $fh, '>', $outfile or die "check_input_files: error opening $outfile: $!";
+	open $fh, '>>', $outfile or die "check_input_files: error opening $outfile: $!";
     }
 
     # Look through all namelist groups
--- a/components/eam/cime_config/buildnml
+++ b/components/eam/cime_config/buildnml
@@ -202,8 +202,9 @@ def buildnml(case, caseroot, compname):
         # call build-namelist
         # -----------------------------------------------------
 
-        if os.path.exists(os.path.join(casebuild, "eam.input_data_list")):
-            os.remove(os.path.join(casebuild, "eam.input_data_list"))
+        if inst_counter ==1:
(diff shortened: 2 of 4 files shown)


jonbob (approved):
Approved based on visual inspection -- and I will make a similar PR for mpaso and MALI

singhbalwinder (approved):
Looks good to me!

rljacob:
@singhbalwinder please merge this.

mt5555:
@singhbalwinder please merge this.

singhbalwinder:
Merged to next

PR #4657, opened 2021-11-10
Time step size definition in si_init

hyungyukang:
This PR changes time step size definition (parsed from config_dt -> passing in 'dt' itself) in si_init routine and revises some comments.

This PR passed `PEM_Ln9_D.T62_oQU240.GMPAS-IAF.cori-knl_gnu` with `config_btr_si_partition_match_mode=.true.` except `COMPARE_base_modpes`. Although the PR modifies only the semi-implicit code (`config_time_integrator='split_implicit'`), the split-explicit code (`config_time_integrator='split_explicit'`) also does not pass `COMPARE_base_modpes`. It seems there is a problem in other portions of E3SM.

[BFB]

### DIFF
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_si.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_si.F
@@ -743,8 +743,9 @@ module ocn_time_integration_si
          !
          ! The split-implicit barotropic mode solver
          !    - uses the preconditioned communication-avoiding
-         !      (single-reduction) BiCGStab method 
-         !      (Cool and Vanroose, 2017) as a linear iterative solver
+         !      (single-global synchronization) BiCGStab method 
+         !      as a linear iterative solver
+         !      (Kang et al., in preparation)
          !                    
          !    - solves the nonlinear barotropic system but deals with
          !      it as the linear system by introducing the outer and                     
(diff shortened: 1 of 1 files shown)


mark-petersen (approved):
The only change here is reading the time step within a subroutine from the `dt` variable, an 8-byte real in seconds, rather than converting `config_dt` from a string into a real. This is the proper way to do it. Thanks.

hyungyukang:
> The only change here is reading the time step within a subroutine from the `dt` variable, an 8-byte real in seconds, rather than converting `config_dt` from a string into a real. This is the proper way to do it. Thanks.

@mark-petersen , Thanks for reviewing the code!

jonbob:
Passes all active ocean tests in e3sm_developer on chrysalis:

merged to next

jonbob:
merged to master

hyungyukang:
> Passes all active ocean tests in e3sm_developer on chrysalis:
> 
> * ERS_Ld5.T62_oQU120.CMPASO-NYF
> * SMS.T62_oQU120_ais20.MPAS_LISIO_TEST
> * SMS_P12x2.ne4_oQU240.WCYCL1850NS.chrysalis_intel.allactive-mach_mods
> 
> merged to next

@jonbob , Thanks for testing and merging this PR.

Issue #4659, opened 2021-11-11
Bug in stress divergence test case in MPAS-Seaice


akturner:
There is a bug in the stress divergence test case (components/mpas-seaice/testing_and_setup/testcases/square/operators_stress_divergence/stress_divergence_scaling.py) where there is an extra use of the weights variable outside the weights do loop in the L2_norm_integral_triangle function

PR #4660, opened 2021-11-11
Fixed bug in the stress divergence test case

akturner:
Fixed a bug in the stress divergence test case where extra weight variable was erroneously used in L2 norm calculation

Fixes https://github.com/E3SM-Project/E3SM/issues/4659

[BFB]


gcapodag (changes requested):
Hi Adrian, per our discussion, I think just that extra weight multiplication has to be removed and no new computation (i.e. the new code) has to be added.
Let's talk on Slack if you'd like. Thanks.

rljacob:
@gcapodag and @akturner is this ready?

rljacob:
@gcapodag and @akturner ping.  Is this ready?

gcapodag:
@rljacob hi! I think I went over the PR last December and left my comments. I do not know if @akturner has implemented them yet.

PR #4661, opened 2021-11-12
Fixes the reading of lat from an unstructured input file in MOSART

donghuix:
Fixes a bug in MOSART for reading the latitude from an unstructured input file.

[non-BFB]

### DIFF
--- a/components/mosart/src/riverroute/RtmMod.F90
+++ b/components/mosart/src/riverroute/RtmMod.F90
@@ -1409,7 +1409,11 @@ contains
           call shr_sys_abort(subname//' ERROR gdc2glo values')
        endif
        rtmCTL%lonc(nr) = rtmCTL%rlon(i)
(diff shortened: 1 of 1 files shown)


rljacob:
What cases will this affect?

donghuix:
> What cases will this affect?

The latitude is not read correctly only when the domain is unstructured in MOSART. This bug doesn't affect MOSART simulation in current master, as the coordinate information is not used in MOSART. But one of our team member is using the coordinate to read data into MOSART in his development.

bishtgautam (approved):

rljacob:
It looks like there is a diff in one field in one test compared to baselines:
SMS.MOS_USRDAT.RMOSGPCC.chrysalis_intel.mosart-unstructure
"domr_lat" is diferent in the coupler history.  Is that expected?

PR #4663, opened 2021-11-15
Initial port to Perlmutter at NERSC

ndkeen:
Add machine file settings for Perlmutter.
Includes changes to run on PM with either gnu or gnugpu.
And introduces nvidia/nvidiagpu compilers.


ambrad (approved):

whannah1 (approved):
Everything looks good, but I'd like to point out that it would beneficial to have a separate default PE layout for running on the GPU. I'm not sure if we can currently change the PE layout based on the compiler, but I think that would be needed to make that work.

ndkeen:
merged to next

ndkeen:
Will merge this to master as several tests look same as before.

We don't like the machine name being "login" in config_machines.xml, but currently, Perlmutter is not returning anything else with the mechanisms CIME is using:

```
login10% python
Python 3.8.8 (default, Apr 13 2021, 19:58:26) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import socket
>>> socket.gethostname()
'login10'
>>> socket.getfqdn()
'login10'
>>> socket.gethostbyaddr(socket.gethostname())
('login10', [], ['10.252.1.55'])
>>>
```

There is a flag to hostname that will return "perlmutter":
```
login10% hostname
login10
login10% hostname -f
login10
login10% hostname -A
login10 x3113c0s21b0n0h0 login10.perlmutter.nersc.gov perlmutter.nersc.gov x3113c0s21b0n0h1 
```

I noted this to NERSC and it may be that once others also complain they might change it and I can update NODENAME_REGEX.

PR #4664, opened 2021-11-16
refactor ESMT calculation of PGF

whannah1:
Refactor the scalar momentum transport parameterization (ESMT) for the C++ CRM code. Previously, the pressure gradient force calculation was done on the CPU for GPU runs, which greatly reduced the throughput. This PR fixes that and maximizes the parallelism for better performance.

[BFB] except for MMF+ESMT test


xyuan (approved):
this is nice, naturally using the 4d data is going to increase the parallelism, and the speedup.

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #4666, opened 2021-11-16
Expanding the OpenACC region in the split time integration routine

brobey:
Expanding the OpenACC region in the split time integration routine and
partially in the other time integration routines. As part of this,
added routines to get scalars instead of pointers to make it
easier to port to OpenACC. These include:
   mpas_get_dimension_scalar
   mpas_get_config_scalar
   included duplicate function varinp_xxx as well as config_xxx for
     config variables to get scalar versions and added to gen_inc.c
All of the framework changes were supplemental features and the existing
functionality was not touched to avoid breakage
Also added gpu version of mpas_vector_reconstruct function to be used
by split time integration until all the uses can be done on the gpu.

Tested with Nvidia compiler 21.3 for OpenACC. Tested on Darwin and Summit. Also tested regular version with GCC and Intel.


philipwjones:
While testing on Summit, the GPU version appears to have larger diffs than expected with the GPU version on master - trying to isolate the issue. CPU version was bfb as expected.

xylar:
@brobey, @philipwjones and @jonbob, just ping me when this is ready to test.  I'm happy to run the compass `pr` test suite on several machines and compilers.

brobey:
@xylar @jonbob @philipwjones I've added fixes to the branch for the pull request and it is passing for me now on Summit with OpenACC. It is ready for additional testing and review. I've merged in the changes from master up to about a week ago.

components/mpas-framework/src/framework/mpas_pool_routines.F
```diff
@@ -4916,6 +4923,175 @@ subroutine mpas_pool_get_config_logical(inPool, key, value, record)!{{{
    end subroutine mpas_pool_get_config_logical!}}}
 
 
+!-----------------------------------------------------------------------
+!  subroutine mpas_pool_get_config_real_scalar
+!
+!> \brief MPAS Pool Real Config Access Routine
+!> \author Michael Duda, Doug Jacobsen
+!> \date   03/27/2014
```
xylar, line 4931: I don't know how picky we want to be about these headers.  But maybe these need correct dates and author(s)?

xylar (approved):
I tested this PR with the `pr` test suite in Compass, using 89d472679b (which appeared to be the version of `master` that had been merged into this branch most recently) as a baseline.  I tested with:
* `intel` and `impi` on Chrysalis
* `gnu` and `openmpi` on Chrysalis
* `intel` and `impi` on Compy

I also tested the `nightly` test suite on Chrysalis in debug mode with `gnu` and `openmpi`.

All tests were successful (bit-for-bit).

I'm approving but with the note above about author and date.  It might be better if each new subroutine had a correct author and date rather than one copied from another subroutine.

components/mpas-framework/src/framework/mpas_pool_routines.F
brobey: This file was already there before this commit. Do you want me to mark it for when it was created? As best I can tell, it was copied from a file Jacobsen and Duda created in 2013 and Mark Petersen make the copy on May 17, 2021.

xylar: @brobey, I can definitely understand the confusion.

@mark-petersen and @philipwjones can correct me if I'm wrong but I believe these dates and authors should refer to each subroutine separately, not the whole file.  Further, believe the date is intended to be that of the most recent modification, not the creation of the subroutine.

If my understanding is correct, `\author` should be `Bob Robey` and the `\date` entry should be the date you created each subroutine.

Regarding the May 17, 2021 date on the files, this is merely the date the @mark-petersen moved MPAS-Model over to the E3SM repository, and isn't a significant date in terms of code modifications.

@mark-petersen, we really need an MPAS developer's guide for folks to follow if we want these kind of conventions to be followed as we want.

philipwjones: @xylar - I don't know what to do about these. When doing some of the refactoring/rewrites, I started out by adding my name and adding "revised" with a date while preserving the original date. But then I didn't feel like I was actually modifying the algorithms in a lot of cases, so stopped adding/updating. I don't know what we really want to do for this. Acknowledge original author/algorithm, acknowledge all changes? It seems to me we'd like to give credit for the algorithm at a minimum. For minor changes, we have the git record. For significant refactoring that doesn't change the algorithm, I don't know what we want.

xylar: @philipwjones, I see.  That makes a lot of sense.  I would then suggest that whoever does the refactoring add their name to the list of authors at a minimum.

I think we also need to decide what the date is for.  I don't personally think it's useful for anything at all because git history and git blame can tell you what you need to know better than that (typically inaccurate) date can.  But if we want to keep a "creation date" and a "revised date", that would make sense.

brobey: How about if I put the following in the heading?
4926 !-----------------------------------------------------------------------
4927 !  subroutine mpas_pool_get_config_real_scalar
4928 !
4929 !> \brief MPAS Pool Scalar Real Config Access Routine
4930 !> \author Bob Robey
4931 !> \creation date   11/16/2021
4932 !> \details
4933 !> This subroutine returns the scalar value associated with a config option with the
4934 !> name key in inPool. This subroutine is based on mpas_pool_get_config_real, but 
4935 !> sets value to be a scalar real instead of a pointer to a real. 
4936 !
4937 !-----------------------------------------------------------------------

brobey: @xylar @philipwjones I went ahead and fixed the documentation headers for the new routines in mpas_pool_routines.F and mpas_vector_reconstruction.F. I set myself as the author and referenced the original subroutines they were based on.

xylar: @brobey, thanks very much!  What you've done seems perfect to me.

rljacob:
@philipwjones waiting on your review.

philipwjones:
yes @rljacob we're still trying to track down one last bfb issue on Cori-knl.

Otherwise, this PR is bfb on both CPU and GPU against master on Summit with nvhpc. And passes compass PR tests on Chrysalis, Compy with Intel when compared to the hash mentioned above. Once we track down the Cori thang, we should be good to go.

rljacob:
@philipwjones any luck on that bfb issue?

brobey:
Replacing with partial commit on branch brobey/mpas-ocean/openacc-split

Issue #4584, opened 2021-10-05
Several tests in e3sm_developer now fail in MPAS-SEAICE when using GNU v9 or higher


ndkeen:
Already noted here https://github.com/E3SM-Project/E3SM/issues/4495

I just wanted to show that all 9 of these tests in e3sm_developer are failing in the same way using GNU compiler v9 or higher. This is default version on chrysalis and many new machines becoming available use gnu9 or higher.

There is little info in the e3sm.log file for the tests. Need to look in run/ dir to spot the `log.seaice.*` files.

```
ERP_Ld3.ne4_oQU240.F2010
ERS_Ld3.ne4pg2_oQU480.F2010
SMS.T62_oQU120_ais20.MPAS_LISIO_TEST
SMS.ne4_oQU240.F2010.machine_gnu.eam-cosplite
SMS_Ln5.ne4pg2_oQU480.F2010.machine_gnu.eam-thetahy_pg2
SMS_Ln5.ne4pg2_oQU480.F2010.machine_gnu.eam-thetahy_sl_pg2
SMS_Ln5.ne4pg2_oQU480.F2010
SMS_Ln9.ne4_oQU240.F2010.machine_gnu.eam-outfrq9s
SMS_P12x2.ne4_oQU240.WCYCL1850NS.machine_gnu.allactive-mach_mods
```

With error message found in a file such as `run/log.seaice.0039.d0001.err`:

```
ERROR:  -------------------------------------
ERROR: 
ERROR:  picard convergence failed!
ERROR:  ==========================
ERROR: 
ERROR:  Surface: Tsf0, Tsf
ERROR:            0  -4.8157786256928752       -4.9276324350521730
ERROR: 
ERROR:  Snow: zTsn0(k), zTsn(k), zqsn0(k), ks(k), Sswabs(k)
ERROR:            1   0.0000000000000000        0.0000000000000000       -110121000.00000000        2.2950296274355850E-315   5.9330250652307008E-003
ERROR:            2   0.0000000000000000        0.0000000000000000       -110121000.00000000        0.0000000000000000        1.1977949299187558E-002
ERROR:            3   0.0000000000000000        0.0000000000000000       -110121000.00000000        0.0000000000000000        1.1963449907972840E-002
ERROR:            4   0.0000000000000000        0.0000000000000000       -110121000.00000000        0.0000000000000000        1.1948969626128606E-002
[...]

PR #4672, opened 2021-11-18
For gnu compiler version 9 and above, adjust compiler flag for one mpas source

ndkeen:
As a work-around for a runtime error with certain tests fail (with `ERROR:  picard convergence failed!`), slightly adjusting the optimization flags for one source file (`mpas-seaice/src/column/ice_shortwave.F90`) seems to work.
This only happens with GNU compiler version 9 and above, and only with non-DEBUG builds -- this PR should only affect such.

We simply leave the `-O2` or `-O1` setting (depending on the machine defaults) and then add: `-fno-tree-pta`, which turns off "points-to" analysis on tress (from man page: "Perform function-local points-to analysis on trees. This flag is enabled by default at -O and higher.")

The test I mostly used to debug this issue was some form of `SMS.T62_oQU120_ais20.MPAS_LISIO_TEST`

Using `SMS_P128x2.ne30pg2_ne30pg2.F2010-CICE`, I did not see any performance differences on Perlmutter.

Fixes https://github.com/E3SM-Project/E3SM/issues/4584

### DIFF
--- a/cime_config/machines/Depends.gnu.cmake
+++ b/cime_config/machines/Depends.gnu.cmake
@@ -4,8 +4,11 @@ list(APPEND MPAS_ICE_SHORTWAVE
   ${CMAKE_BINARY_DIR}/core_seaice/column/ice_shortwave.f90
 )
 
+# For optimized GNU builds that use v9 or higher, remove an optimization on one file
 if (NOT DEBUG)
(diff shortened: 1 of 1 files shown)


jgfouca (approved):

jonbob (approved):
Tested on cori-knl and works great -- thanks!

ndkeen:
merged to next

PR #4676, opened 2021-11-19
Add test for new ELM and MOSART features

hydrotian:
A new test is added to e3sm_land_developer testing suite with the following land features turned on: 
two-way irrigation, lake storage, bgc mode, and transient mode. It also turns on two river features:
inundation and water management. These features were developed for V2 BGC simulations.

[BFB]

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -42,7 +42,8 @@ _TESTS = {
             "ERS.MOS_USRDAT.RMOSGPCC.mosart-mos_usrdat",
             "SMS.MOS_USRDAT.RMOSGPCC.mosart-unstructure",
             "SMS.r05_r05.IELM.elm-topounit",
-            "ERS.ELM_USRDAT.I1850ELM.elm-usrdat"
+            "ERS.ELM_USRDAT.I1850ELM.elm-usrdat",
+            "ERS.r05_05.IELM.elm-V2_ELM_MOSART_features"
             )
         },
 
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -43,7 +43,7 @@ _TESTS = {
             "SMS.MOS_USRDAT.RMOSGPCC.mosart-unstructure",
             "SMS.r05_r05.IELM.elm-topounit",
             "ERS.ELM_USRDAT.I1850ELM.elm-usrdat",
-            "ERS.r05_05.IELM.elm-V2_ELM_MOSART_features"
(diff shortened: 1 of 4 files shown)


hydrotian:
@bishtgautam I have tested the case added here on Compy without any issue. But don't know if there could be issues on other machines. Also could you please help uploading the input files used in this case to the server if they haven't been uploaded?

- fsurdat = '$DIN_LOC_ROOT/lnd/clm2/surfdata_map/surfdata_0.5x0.5_simyr1850_c211019.nc'
- flanduse_timeseries = '$DIN_LOC_ROOT/lnd/clm2/surfdata_map/landuse.timeseries_0.5x0.5_HIST_simyr1850-2015_c211019.nc'
- paraFile = '$DIN_LOC_ROOT/rof/mosart/global_reservoir_halfdegree_c20200917.nc'
- frivinp_rtm = '$DIN_LOC_ROOT/rof/mosart/MOSART_Global_half_20210422.nc'

bishtgautam (approved):

bishtgautam:
@hydrotian I uploaded the files and merged this PR to next (d8a5f0f831046d5dd74262ac512974dd79588898). The info regarding the merge of this PR does not show up here because I forgot to add `#`  in the commit message (i.e. `PR 4676`)

rljacob:
Add modifications so that mosart will output history every day.  The test will use those as well.

hydrotian:
@rljacob Does that mean the MOSART results are not tested in this new test? I can see that some "r2x_*" fields been compared in the test report. I thought they are calculated by MOSART, but maybe I'm wrong or they are not the whole list of the variables from MOSART?

rljacob:
Only coupler history files are compared because those are always written.  Currently it will check the fields MOSART sends to the coupler.  I also want mosart history files in this to make it a more complete test.

hydrotian:
Thanks for the explanation! I've added changes to ask MOSART output daily h0 files but the comparison didn't happen between the baseline and restart h0 files. I do noticed that the ERS test generates a baseline MOSART h0 file with a .base suffix, but unlike the cpl file, it didn't copy the restart h0 file and generate a .rest file. I wonder if there's a switch need to be turned on to activate the additional comparison on the h0 files. I did however manually compare the baseline and restart files using cprnc and they are identical.

rljacob:
This needs to be remerged to next with the additions for mosart output.

PR #4678, opened 2021-11-24
Fixes a bug in O3 dry dep stomatal resistances

### LINKED ISSUE #4677: Bug in components/elm/src/biogeochem/DryDepVelocity.F90
A bug exists in the ozone dry deposition code in the calculation of stomatal resistances, carried over from CESM. 
Following Emmons et al ([2020, JAMES](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2019MS001882)), the stomatal resistances for shaded and sunlit leaves should be added in parallel as opposed to in series.

This should be a relatively straightforward bug fix, instead of
Rs = fsun * Rsun + (1‐fsun) * Rsha, the correct formula is:

1/Rs = fsun * (1/Rssun) + (1‐fsun) * (1/Rssha)

The bug is on line 459 in components/elm/src/biogeochem/DryDepVelocity.F90:
https://github.com/E3SM-Project/E3SM/blob/4f29ad3f15044b68210ab846ccb167f89f40853a/components/elm/src/biogeochem/DryDepVelocity.F90#L459

singhbalwinder:
Currently, stomatal resistances for shaded and sunlit leaves
are added in series but they should be added in parallel

Fixes #4677

[BFB]

### DIFF
--- a/components/elm/src/biogeochem/DryDepVelocity.F90
+++ b/components/elm/src/biogeochem/DryDepVelocity.F90
@@ -456,7 +456,9 @@ CONTAINS
                   rlux(ispec)=1.e36_r8
                else
 
-                  rs=(fsun(pi)*rssun(pi))+(rssha(pi)*(1._r8-fsun(pi)))
+                  !Following Emmons et al (2020, JAMES), the stomatal resistances for shaded and sunlit
+                  !leaves should be added in parallel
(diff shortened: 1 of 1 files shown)


susburrows (approved):
I've double-checked that this matches the correct equation as reported by Sam.  Thank you for taking care of this so quickly, @singhbalwinder !

PR #4680, opened 2021-11-29
GPU port of ocean vmix routines

brian-oneill:
OpenACC added to vmix velocity and tracer routines. Routines for vmix coefs untouched.
Changes include:
- replaced mesh pool with ocn_mesh module
- minimized copies of tridiag coefs in solution loops
- inlined tridiag and kpp_rhs routines
- added OpenACC directives for loop kernels and data updates

Should be bit-for-bit with CPU-only. Tested with PGI on summit.

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_vmix.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_vmix.F
@@ -25,6 +25,7 @@ module ocn_vmix
    use mpas_derived_types
    use mpas_pool_routines
    use mpas_timer
+   use ocn_mesh
 
    use mpas_constants
    use ocn_constants
@@ -132,8 +133,7 @@ contains
       integer :: err1, err2, err3
       integer :: timeLevel
 
-      integer :: iEdge, iCell, nEdges, nCells
-      integer, dimension(:), pointer :: nEdgesArray, nCellsArray
+      integer :: iEdge, iCell, k, nEdges, nCells
 
       !-----------------------------------------------------------------
       !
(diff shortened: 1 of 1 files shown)


philipwjones:
Ran 10-day QU240 tests on Summit with nvhpc (pgi). Confirmed bfb cpu. Also bfb with GPU code on master when gpu enabled.  About a 3x speedup gpu (8 core/2 gpu) over cpu (8 cores) for the vmix solve routines.

Testing on other machines and compilers now to verify bfb but don't expect any issues.

philipwjones (approved):
Approved based on visual inspection and the following tests:

Summit: QU240 10-day standalone test in both CPU-only and GPU modes, both bfb with code on master

Compy: passes all the following E3SM tests:
SMS.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.compy_intel
SMS.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.compy_pgi
------------
SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_intel
SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_pgi
------------
PET_Ln3.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_intel
PET_Ln3.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_pgi
------------
PEM_Ln3.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_pgi
PEM_Ln3.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_intel
------------
PEM_Ln9.T62_oQU240.GMPAS-IAF.compy_intel
PEM_Ln9.T62_oQU240.GMPAS-IAF.compy_pgi
------------
PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.compy_pgi
PET_Ln3.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.compy_intel
------------
PET_Ln9.T62_oQU240.GMPAS-IAF.compy_pgi
PET_Ln9.T62_oQU240.GMPAS-IAF.compy_intel
------------
SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_pgi
SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.compy_intel
------------
PEM_Ln9_D.T62_oQU240.GMPAS-IAF.compy_intel.
PEM_Ln9_D.T62_oQU240.GMPAS-IAF.compy_pgi

jonbob:
Passes all active-ocean tests in e3sm_developer on chrysalis:

merged to next

jonbob:
merged to master

xylar:
@jonbob, and @philipwjones, I'm seeing non-bit-for-bit changes as a result of this PR in `compass` test cases on Chrysalis with Intel and Intel-MPI.  I'm a little concerned about that because some of the changes are pretty far from machine precision.

See https://github.com/MPAS-Dev/compass/pull/295#issuecomment-991949465

xylar:
In retrospect, I guess it's clear that this PR should also have been flagged for testing in `compass`.   For now, I mostly want to try to understand why I'm seeing `compass` differences that the E3SM tests seem not to have picked up.

components/mpas-ocean/src/shared/mpas_ocn_vmix.F
```diff
@@ -288,54 +296,67 @@ subroutine ocn_vel_vmix_tend_implicit_rayleigh(meshPool, dt, kineticEnergyCell,
...
+                   / layerThicknessEdge(k,iEdge)
+            m        = A/bTemp(k-1)
+            C(k)     = -2.0_RKIND*dt*vertViscTopOfEdge(k+1,iEdge) &
+                   / (layerThicknessEdge(k,iEdge) + layerThicknessEdge(k+1,iEdge)) &
+                   / layerThicknessEdge(k,iEdge)
+            bTemp(k) = 1.0_RKIND - A - C(k) &
+                   ! added Rayleigh terms
+                   + dt*(rayleighDampingCoef/((1.0_RKIND - rayleighDepthVariable) + rayleighDepthVariable*edgeThicknessTotal)) &
+                   - m*C(k-1)
+            rTemp(k) = normalVelocity(k,iEdge) - m*rTemp(k-1)
```
mark-petersen, line 323: @xylar and @philipwjones, the tridiagonal solve is in-lined here. There are changes in order of operations, where BFB between before/after could be compiler and machine dependent, and particularly sensitive to optimization. It would be good to see the order of magnitude of differences with COMPASS.  The standard comparison reports that. If it only loses a few digits from machine precision, I think we can bless the COMPASS diffs. 

I think that all terms in this routine are nonzero in both COMPASS and E3SM (bottom drag, surface forcing, viscosity) so this should not be a difference in test configuration. If anything, E3SM should be more challenging because of time-varying forcing and surface fluxes.

xylar: @mark-petersen, as I stated above, the comparison is here:
https://github.com/MPAS-Dev/compass/pull/295#issuecomment-991949465

xylar: The fact that E3SM tests were all bit-for-bit and Compass tests were not suggests to me, as I said, that it just has to be something like differences in compiler flag or something like that.  I just don't see any other way to understand being bit-for-bit in E3SM but not in Compass with the same compilers.

mark-petersen:
I think the mismatch is in the routine `ocn_vel_vmix_tend_implicit_Rayleigh`, which is tested in COMPASS but not E3SM. See explanation here:
https://github.com/MPAS-Dev/compass/pull/295#issuecomment-1005160816

philipwjones:
The only compiler difference I can see is -O2 vs -O3, but I ran a before/after case with -O2 and Intel on compy and still see diffs. This was also a QU240 case with no Rayleigh.  Diffs are still at roundoff level and not surprising, but still confused about why the E3SM tests didn't pick them up. I'll look a bit more into both that issue and the diffs themselves to verify they really are roundoff.

PR #4681, opened 2021-11-30
Minor fixes to Hommexx, for building theta-l_kokkos in non-standalone mode

bartgol:
Two mods:

- fix driver mod for theta-l_kokkos, to compile test forcing stuff only in standalone mode (fixes link error).
- change Kokkos.cmake script, so that it does not add _assumed_ include/link folders/libs when Kokkos is built externally (fixes hundreds of non-existent include dir in SCREAM).

BFB.

### DIFF
--- a/components/homme/src/theta-l_kokkos/prim_driver_mod.F90
+++ b/components/homme/src/theta-l_kokkos/prim_driver_mod.F90
@@ -422,14 +422,16 @@ contains
 
     call TimeLevel_Qdp(tl, dt_tracer_factor, n0_qdp, np1_qdp)
 
+#if !defined(CAM) && !defined(SCREAM)
+    ! Test forcing is only for standalone Homme (and only for some tests/configurations)
     if (compute_forcing_and_push_to_c) then
-!    if ( .true. ) then
       call compute_test_forcing(elem,hybrid,hvcoord,tl%n0,n0_qdp,max(dt_q,dt_remap),nets,nete,tl)
       call t_startf('push_to_cxx')
       call push_forcing_to_c(elem_derived_FM,   elem_derived_FVTheta, elem_derived_FT, &
                              elem_derived_FPHI, elem_derived_FQ)
       call t_stopf('push_to_cxx')
     endif
+#endif
 
     call prim_run_subcycle_c(dt,nstep_c,nm1_c,n0_c,np1_c,nextOutputStep)
 
@@ -442,8 +444,6 @@ contains
     call init_logic_for_push_to_f(tl,statefreq,nextOutputStep,compute_diagnostics)
 
     if (push_to_f) then
-!    if (MODULO(tl%nstep,statefreq)==0 .or. tl%nstep >= nextOutputStep .or. compute_diagnostics) then
-!    if (.true.) then
       ! Set pointers to states
       elem_state_v_ptr         = c_loc(elem_state_v)
       elem_state_w_i_ptr       = c_loc(elem_state_w_i)
--- a/components/homme/cmake/Kokkos.cmake
+++ b/components/homme/cmake/Kokkos.cmake
@@ -1,38 +1,45 @@
+set (KOKKOS_LIBRARIES "kokkoscore;kokkoscontainers")
+set (KOKKOS_TPL_LIBRARIES "dl")
+
 set (KOKKOS_INSTALLATION_NEEDED FALSE CACHE BOOL "")
-if (NOT DEFINED E3SM_KOKKOS_PATH)
-  # Build kokkos submodule if user did not specify KOKKOS_PATH.
+if (DEFINED E3SM_KOKKOS_PATH)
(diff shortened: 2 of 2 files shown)


components/homme/src/theta-l_kokkos/prim_driver_mod.F90
```diff
@@ -422,14 +422,16 @@ subroutine prim_run_subcycle(elem, hybrid, nets, nete, dt, single_column, tl, hv
 
     call TimeLevel_Qdp(tl, dt_tracer_factor, n0_qdp, np1_qdp)
 
+#if !defined(CAM) && !defined(SCREAM)
```
oksanaguba, line 425: I am going to clean up this file in the next forcing PR. Why did you have to touch it? This particular case, when not SCREAM, is handled in `subroutine init_logic_for_push_to_c()` .

Can you not touch this file in this PR?

bartgol: I needd to ifdef out the next few lines, since `test_mod` is only used if CAM and SCREAM are not defined. When building this file in SCREAM, it caused a link error.

oksanaguba: I want to hide all logic about push/pull etc. in `init_logic_for_push_to_c` and `init_logic_for_push_to_f`, cause it is cleaner. We could try to have a dummy function that would call compute_test_forcing if !SCREAM? it is ugly though.

bartgol: I'm ok hiding all the logic, and reducing the number places where we use preprocessor macros. Unfortunately, all the stuff that is undefined must be ifdeffed out.

We can let SCREAM compiler all the forcing stuff, and simply rely on the runtime switches set in the init routines for skipping that stuff. Either way, we either ifdef out code in all places, or we never  ifdef it out. The former leads to arguably cleaner code (less ifdefs, all logic confined to the init routines), while the latter allows to not compile all the test stuff for SCREAM use (but we are currently compiling it anyways, so maybe it's fine).

bartgol: Actually, this is necessary:

```
    ! Test forcing is only for standalone Homme (and only for some tests/configurations)
    if (compute_forcing_and_push_to_c) then
```
since EAM does not compile all the test_mod stuff, so these lines cannot be compiled for CAM builds. SCREAM does build all the test_mod stuff, so it would be ok for SCREAM, but it definitely would not compile in EAM without the `#if !defined(CAM)..` line.

bartgol: Since no EAM compset uses theta-l_kokkos, this might be a somewhat pointless issue. But in the long term, it must be addressed.

bartgol: Long story short, we have 2 approaches:

- ifdef out all occurrences of compute/push forcing
- do not ifdef out any occurrence, but make sure test_mod and test_src/* files are always built.

oksanaguba: Do you need changes in this file for this PR? is there a test case in scream already?

I can integrate this as is, but I will change this file again later.

ambrad: @oksanaguba I hypothesize that if the test SMS_B.ne4_ne4.F-EAM-AQP1.machine_compiler.eam-hommexx is modified to build with theta-l_kokkos instead of preqx_kokkos, this linking error will be evident in the E3SM repo without needing to look at the SCREAM one.

oksanaguba:
@bartgol minot comment -- they usually want branch name to be GITUSER/COMPONENT/UNIQUE_NAME

bartgol:
> @bartgol minor comment -- they usually want branch name to be GITUSER/COMPONENT/UNIQUE_NAME

Ah, sorry. I even misspelled my username. Do you want me to resubmit with the correct name? I would have to close this PR and open a new one though.

oksanaguba:
no, thanks. I am running tests from this branch now.

oksanaguba:
on chrysalis, homme tests are all ok, but there is this:
```
[ac.onguba@chrlogin2 ess]$ ./cs.status.20211205_150009_34yndo | grep FAIL
    FAIL SMS.ne4_oQU240.F2010.chrysalis_intel.eam-thetahy_sl BASELINE master: DIFF
```
I am not sure why -- prim_driver mod is not used because we do not use thetaxx yet, and kokkos make file, I would assume, is not used by cime build, but it must be. The test ran to completion.

ambrad:
@oksanaguba, what test suite is run here? Is it possible this branch has fallen sufficiently behind master that some of the e3sm_developer, etc., tests are diffing against master? You might try to mege master into this branch (locally) and try again.

oksanaguba:
I ran e3sm_atm_integration . I do not use machine baselines, I do my own. I updated modules in both clones and re-ran, and now it passes. So, this is ready.

oksanaguba:
in next

PR #4682, opened 2021-12-01
add ttype9 to hommexx

oksanaguba:
Adding ttype9 and dcmip2012 tests 2.1, 2.2, and 3 to the bfb suite. Converting some of the tests to ttype9.

Not yet added is a limiter call between ttype9 stages.

There will be another PR to move limiter from CaarFunctorImpl and to verify it is bfb with F code.

[nonbfb] for homme bfb suite for theta-fhs3 (now with ttype9) and new baselines for theta-fdc tests.


oksanaguba:
I still need to run this on weaver

components/homme/src/theta-l_kokkos/cxx/cxx_f90_interface_theta.cpp
```diff
@@ -105,22 +106,23 @@ void init_simulation_params_c (const int& remap_alg, const int& limiter_option,
...
-    params.time_step_type = TimeStepType::IMEX_KG355;
+    //5 stage, based on 3rd order explicit KGU table
+    //2nd order implicit table
+    params.time_step_type = TimeStepType::ttype9_imex;
   } else if (time_step_type==10) {
-    params.time_step_type = TimeStepType::IMEX_KG255;
+    //5 stage, based on the 2nd order explicit KGU table
+    //2nd order implicit table
+    params.time_step_type = TimeStepType::ttype10_imex;
   }
```
ambrad, line 126: Would be good to add and `else` clause with `Errors::runtime_abort("Invalid time_step_time" + std::string(time_step_type), Errors::err_not_implemented);`

ambrad (approved):

ambrad:
Did you check that the new test run times in the BFB build will be small enough that sandiatoss3 won't have a problem?

oksanaguba:
@ambrad thanks! fixed both things.

components/homme/src/share/cxx/HommexxEnums.hpp
```diff
@@ -106,25 +106,14 @@ enum class UpdateType {
 
 enum class TimeStepType {
   // Explicit
-  LF              =  0, // LeapFrog
-  RK2             =  1,
-  IMEX_KG254_EX   =  4, // Explicit table from IMEX_KG254
-  ULLRICH_RK35    =  5,
+  ttype5          =  5,
```
bartgol, line 109: I think I might be in the minority here, but I'll say it anyways.

I think that in C++ we should take advantage of enums, and not use names like "ttype5", or "q_alg_10", or "ftype2". If we prefer ints, then let's use ints. Enum's are a way to convey meaning like a string, while keeping the underlying simplicity of an int.

That said, I understand that ttype being an int in the namelists is an argument for keeping the status quo, so I'm not going to push on this.

components/homme/src/share/cxx/SphereOperators.hpp
```diff
@@ -1031,6 +1038,9 @@ class SphereOperators
   ExecViewManaged<const Real * [NP][NP]>        m_metdet;
   ExecViewManaged<const Real * [2][2][NP][NP]>  m_d;
   ExecViewManaged<const Real * [2][2][NP][NP]>  m_dinv;
+
+  Real m_rearth;
```
bartgol, line 1042: SphereOperators doesn't really need to store rearth, only rrearth, right?

bartgol (approved):
Looks good. My only comments are more style than anything, so feel free to ignore.

oksanaguba:
@bartgol  -- thanks! I only changed how tstep types are named, but left the original structure enum class TimeStepType there. About SphOps -- you are correct, only rrearth.

oksanaguba:
commit 1b1f762 on weaver for theta-fhs* and theta-fdc* tests is bfb. unit tests look ok on weaver. I ran some preqx tests, too. I did not run baseline comparisons on weaver.

homme suites on chrysalis look as expected, too. this is ready.

ambrad:
@oksanaguba are the theta-l_kokkos tests BFB on Weaver? And to the unit tests pass on Weaver? Thanks.

oksanaguba:
Unit tests pass on weaver. I had some random fails which I fixed, but I do not see any more fails. The fails are due to rearth not being init-ed, so we will see them on CPU if it is still an issue. I think I could have done a better job init-ing rearth to 0 in all ctors and then throwing error after setups, or something like that, but I want to move to forcing now. 

I am running all theta-f* tests on weaver now. These are the ones that have bfb comparison.

oksanaguba:
All theta-f*ne2 tests on weaver for bfb comparisons pass and all ran to completion.

oksanaguba:
in next

PR #4686, opened 2021-12-03
Update BGC compsets to turn ocn bgc back on, add new ocn bgc init file

### LINKED ISSUE #4616: Get MPASO/MARBL working correctly with original BGC test compsets
Two tests configurations were modified to allow testing with MARBL in mpas-ocean to pass, but those configurations need to be reinstated once MARBL is working completely. That process includes replacing the ocean initial file with one consistent with MARBL as well as fixing the communication between ocean and seaice BGC. The compsets that were modified are:
* BGCEXP_CNTL_CNPECACNT_1850
* BGCEXP_CNTL_CNPRDCTC_1850

jonbob:
Adds new ocn BGC files that are consistent with the current MARBL-based implementation. This allows the ocn BGC to be turned back on in BGC compsets.

Fixes #4616 

[non-BFB] only for BGC compsets


jonbob:
@maltrud - there are a couple of open issues we should talk about before this gets merged. Should we delete the part of the mpaso buildnml that specifies BGC init files for the older oEC60to30v3, since it will not work with MARBL. Also, I think there are a few other compsets that need to be modified, as well the the BGC tests that use a v1 resolution.

jonbob:
Changed the test suites to use v2 versions of the compsets and resolutions. This branch passes:
* SMS_Ld2.ne30pg2_r05_EC30to60E2r2.BGCEXP_CNTL_CNPRDCTC_1850.chrysalis_intel.elm-bgcexp
* SMS_Ld2.ne30pg2_r05_EC30to60E2r2.BGCEXP_CNTL_CNPECACNT_1850.anvil_intel.elm-bgcexp
which replace the v1 tests currently in the repo.

Note that we'll remove the ECA compsets (and tests) in another PR to also reflect v2 changes.

jonbob:
@bishtgautam - I asked for your review because I removed a user_nl_mpassi file from a testmod that was defined in ELM. I don't believe that it was used anywhere else except in the automated test suite, but just wanted to make sure.

bishtgautam (approved):

maltrud:
@jonbob , do you think we should change the ocean label from "ECO" to "BGC" for consistency with ice and land?  I don't really care too much, but this might be a good time if anyone does care.  this would touch a few more files, though.

jonbob:
@maltrud - I do like that idea. We can implement it here or in the follow-on PR that would modify all BGC compsets to the v2 approach instead of v1?

maltrud:
@jonbob let's do it in the next PR for a fully v2 flavor....

maltrud (approved):

jonbob:
test merge passed SMS_Ld2.ne30pg2_r05_EC30to60E2r2.BGCEXP_CNTL_CNPRDCTC_1850.chrysalis_intel.elm-bgcexp
merged to next

PR #4687, opened 2021-12-03
Fix filename interval for 2 analysis member restarts

xylar:
The Eliassen-Palm and time-filters analysis members both had hard-coded restart file intervals of 1 year.  This causes restarts with these analysis members to fail in situations where the restart interval does not match the file interval.  This merge changes the filename interval to be the same as the output interval to match restart streams in all other analysis members.

### DIFF
--- a/components/mpas-ocean/src/analysis_members/Registry_eliassen_palm.xml
+++ b/components/mpas-ocean/src/analysis_members/Registry_eliassen_palm.xml
@@ -643,7 +643,7 @@
 			type="input;output"
 			mode="forward;analysis"
 			filename_template="restarts/eliassenPalm_restart.$Y-$M-$D.nc"
-			filename_interval="01-00-00_00:00:00"
+			filename_interval="output_interval"
 			input_interval="initial_only"
 			output_interval="stream:restart:output_interval"
 			packages="eliassenPalmAMPKG"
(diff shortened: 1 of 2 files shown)


xylar:
These analysis members are not enabled in E3SM runs by default:
https://github.com/E3SM-Project/E3SM/blob/417b03f684f47774c53ba242659cbac5ff63625d/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml#L687
https://github.com/E3SM-Project/E3SM/blob/417b03f684f47774c53ba242659cbac5ff63625d/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml#L649

So these changes will have no effect on E3SM runs.  The only effect will be in `compass` tests where the AMs are enabled.  In particular, I had trouble with these settings in https://github.com/MPAS-Dev/compass/pull/273, in which I wasn't able to restart a longer run.

xylar:
@mark-petersen and @jonbob, please let me know what E3SM testing you would like to see from me on this PR.  I'm happy to do whatever is needed but don't want to run arbitrary testing without checking with you first.

jonbob (approved):
Approved based on visual inspection -- thanks for catching this, @xylar

xylar:
Thanks, @mark-petersen.  I did the same search but much less efficiently.  Glad we found the same thing.

jonbob:
passes sanity testing -- merged to next

jonbob:
merged to master

PR #4688, opened 2021-12-04
Correction to resolve restart bfb error.

pnlfang:
This bug fix is related to the restart bfb error when plant hydraulics is activated.

### DIFF
--- a/components/elm/src/biogeophys/PhotosynthesisMod.F90
+++ b/components/elm/src/biogeophys/PhotosynthesisMod.F90
@@ -3365,22 +3365,9 @@ contains
           ! cannot invert the matrix, solve for x algebraically assuming no flux
           exit
        end if
-       if (laisun(p)>tol_lai.and.laisha(p)>tol_lai)then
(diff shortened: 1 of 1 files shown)


rljacob:
Can you explain why this fixed the restart bfb problem?  Its not obvious from looking at the code.

pnlfang:
>

pnlfang:
The error occurs when LAI is close to 0, the flux between the leaf and stem becomes 0 by setting water potential in leaf and stem equal. The original approach caused unconverged solution, which changes the flux after restart. What I did is replacing the flux equation with equal potential equation and that solved the problem.

rljacob:
@peterdschwartz please start merging this.

peterdschwartz:
@rljacob Had to take yesterday off due to the covid booster, but I'll merge it tonight when the tests come back bfb.

peterdschwartz (approved):

peterdschwartz:
merged to next

PR #4692, opened 2021-12-09
set SCREAM to run with adjust_ps=.false.

### LINKED ISSUE #4693: pressure update due to phase change
At the end of the physics timestep the pressure has to be updated to sources/sinks of water vapor.
in E3SM v1 and v2, this change is made in the surface pressure and then distributed uniformly through the column.
In a vertically lagrangian model, the correct behavior is to change the pressure locally, based on the local changs in water vapor.

This is a small change with minor climate impacts, but implementation is on hold so as to not change v2 behavior.

mt5555:
dont reproduce buggy behavior used in E3SM v1/v2

[BFB]
In the SCREAM model, will be non-BFB, with minor impacts on climate

This address #4693, but only for the scream code base, and thus this PR should not close #4693

### DIFF
--- a/components/homme/src/share/prim_driver_base.F90
+++ b/components/homme/src/share/prim_driver_base.F90
@@ -1591,7 +1591,7 @@ contains
      adjust_ps=.true.   ! stay on reference levels for Eulerian case
   else
 #ifdef SCREAM
-     adjust_ps=.false.  ! Lagrangian case should adjusting dp3d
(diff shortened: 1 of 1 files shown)


rljacob:

mt5555:
@rljacob: added an issue for this PR

rljacob:
What's the problem?  I think the buildlib code can set the variable based on the compset long name.

mt5555:
here a description of the issue:

https://github.com/E3SM-Project/scream/issues/1343

PR #4700, opened 2021-12-13
new reproducible global sum module for MPAS components

philipwjones:
Adds new module to compute global sums of distributed arrays in a reproducible way so that they are independent of summation order (eg from different partitions).  The 64-bit real algorithm is the same as the DDPDD in shr_reprosum but this interface supports some features that were added for internal MPAS use.

More details from the module comments:
This file contains reproducible sum functions for common data types (R8, R4, I8, I4) and array dimensions (scalar up to 6-d arrays).  A second interface allows for the sum of a product of two arrays that is useful for situations like dot products, volume-weighted sums or multiplicative masks. Finally, a multi-field option permits computing sums for multiple fields (of the same shape, type) at the same time to reduce the number of messages needed. For multi-field sums, only array dimensions up to 5 are supported. All sums should give identical answers independent of domain decomposition or ordering of elements.

The interfaces are overloaded, so to compute the sum of an array or scalar, the call should be as follows with mpiComm being the MPI communicator defined for the model or sub-model:

For sums of scalars across MPI tasks, use:
    sum = global_sum(scalar, mpiComm)

For sums of arrays over the full range of array indices, use:
   sum = global_sum(array, mpiComm)

For sums of arrays over a subset of array indices, use:
    sum = global_sum(array, mpiComm, indexRange)
where indexRange is an integer array of size (2*numDimensions) that contain the start, end index of each dimension. For example, an MPAS array(nVertLevels, nCells) might have an indexRange defined as
   indexRange(1) = kmin, indexRange(2) = kmax,
   indexRange(3) = 1,    indexRange(4) = nCellsOwned

To compute a sum over products (eg for a dot product, volume weighting or a multiplicative mask), the interface is the same with an extra argument for the second array/scalar, eg:
   sum = global_sum(scalar1, scalar2, mpiComm)
   sum = global_sum(array1, array2, mpiComm)
   sum = global_sum(array1, array2, mpiComm, indexRange)

To compute sums of multiple fields, the interfaces are the same as above with an added nFld in the name, so:
   sum(1:nFlds) = global_sum_nfld(scalar(:), mpiComm)
   sum(1:nFlds) = global_sum_nfld(array, mpiComm)
   sum(1:nFlds) = global_sum_nfld(array, mpiComm, indexRange)
   sum(1:nFlds) = global_sum_nfld(scalar1(:), scalar2(:), mpiComm)
   sum(1:nFlds) = global_sum_nfld(array1, array2, mpiComm)
   sum(1:nFlds) = global_sum_nfld(array1, array2, mpiComm, indexRange)

The multiple fields must be in an array with the field index as the last (right-most) index in the array. If an indexRange is provided, the same range is used for all fields. For the sum of a product - the two product arrays must have the same shape and size (including number of fields).

For reproducibility, single precision sums are computed in double precision. For double precision sums, we use the double-double algorithm of Donald Knuth, further improved by David A Bailey and presented in:
  He, Yun and Chris Ding, 2001: Using Accurate Arithmetics to Improve Numerical Reproducibility and Stability in Parallel Applications, J. of Supercomputing, 18, 259-277.
 
[BFB] for all current tests

### DIFF
--- a/components/mpas-framework/src/framework/framework.cmake
+++ b/components/mpas-framework/src/framework/framework.cmake
@@ -32,5 +32,5 @@ list(APPEND COMMON_RAW_SOURCES
   framework/regex_matching.c
   framework/mpas_field_accessor.F
   framework/mpas_log.F
-  framework/mpas_globalSumMod.F
+  framework/mpas_global_sum_mod.F
 )
--- a/components/mpas-framework/src/framework/framework.cmake
+++ b/components/mpas-framework/src/framework/framework.cmake
@@ -32,4 +32,5 @@ list(APPEND COMMON_RAW_SOURCES
   framework/regex_matching.c
   framework/mpas_field_accessor.F
   framework/mpas_log.F
+  framework/mpas_globalSumMod.F
 )
--- a/components/mpas-framework/src/framework/mpas_globalSumMod.F
+++ b/components/mpas-framework/src/framework/mpas_globalSumMod.F
@@ -1,0 +1,6491 @@
+!|||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
+!
+! mpas_globalSum
+!
+!! \file
+!! \brief This file contains reproducible sum functions for common data
+!!        data types (R8, R4, I8, I4) and array dimensions (scalar
+!!        up to 6-d arrays).  A second interface allows for the sum of 
+!!        a product of two arrays that is useful for either dot products
+!!        or multiplicative masks. Finally, a multi-field option permits
+!!        computing sums for multiple fields (of the same shape, type)
+!!        at the same time to reduce the number of messages needed. For
+!!        multi-field sums, only dimensions up to 5 are supported.
+!!
+!!        All sums should give identical answers independent of domain
+!!        decomposition or ordering of elements.
+!!
+!!        The interfaces are overloaded, so to compute the sum of an
+!!        array or scalar, the call should be as follows with mpiComm
+!!        being the MPI communicator defined for the model or sub-model:
+!!
+!!        For sums of scalars across MPI tasks, use:
+!!           sum = globalSum(scalar, mpiComm)
+!!        For sums of arrays over the full range of array indices, use:
+!!           sum = globalSum(array, mpiComm)
+!!        For sums of arrays over a subset of array indices, use:
+!!           sum = globalSum(array, mpiComm, indexRange)
+!!            where indexRange is an integer array
+!!            of size (2*numDimensions) that contain the start, end
+!!            index of each dimension. For example, an MPAS
+!!            array(nz, nCells) might have an indexRange defined as
+!!                indexRange(1) = kmin, indexRange(2) = kmax,
+!!                indexRange(3) = 1,    indexRange(4) = nCellsOwned
+!!
+!!        To compute a sum over products (eg for a dot product or for
+!!        using a multiplicative mask), the interface is the same with 
+!!        an extra argument for the second array/scalar, eg:
+!!           sum = globalSum(scalar1, scalar2, mpiComm)
+!!           sum = globalSum(array1, array2, mpiComm)
+!!           sum = globalSum(array1, array2, mpiComm, indexRange)
+!!
+!!        To compute sums of multiple fields, the interfaces are the
+!!        same as above with an added nFld in the name, so:
+!!           sum(1:nFlds) = globalSumNfld(scalar(:), mpiComm)
+!!           sum(1:nFlds) = globalSumNfld(array, mpiComm)
+!!           sum(1:nFlds) = globalSumNfld(array, mpiComm, indexRange)
+!!           sum(1:nFlds) = globalSumNfld(scalar1(:), scalar2(:), mpiComm)
+!!           sum(1:nFlds) = globalSumNfld(array1, array2, mpiComm)
+!!           sum(1:nFlds) = globalSumNfld(array1, array2, mpiComm, indexRange)
+!!        The multiple fields must be in an array with the field index
+!!        as the last (right-most) index in the array. If an indexRange
+!!        is provided, the same range is used for all fields. For the
+!!        sum of a product - the two product arrays must have the same
+!!        shape and size (including number of fields)
+!!
+!!        For reproducibility, single precision sums are computed in
+!!        double precision. For double precision sums, we use the
+!!        double-double algorithm of Donald Knuth, further improved by
+!!        David A Bailey and presented in:
+!!        He, Yun and Chris Ding, 2001: Using Accurate Arithmetics to
+!!          Improve Numerical Reproducibility and Stability in Parallel
+!!          Applications, J. of Supercomputing, 18, 259-277.
+!
+!***********************************************************************
+
+module mpas_globalSumMod
+
+   implicit none
+   private
+
+   include "mpif.h"
+
+   !--------------------------------------------------------------------
+   ! Public functions
+   !--------------------------------------------------------------------
+
+   public :: mpas_globalSum, mpas_globalSumNfld
+
+   interface mpas_globalSum
+      module procedure globalSumR8d0
+      module procedure globalSumR4d0
+      module procedure globalSumI8d0
+      module procedure globalSumI4d0
+      module procedure globalSumR8d1
+      module procedure globalSumR4d1
+      module procedure globalSumI8d1
+      module procedure globalSumI4d1
+      module procedure globalSumR8d2
+      module procedure globalSumR4d2
+      module procedure globalSumI8d2
+      module procedure globalSumI4d2
+      module procedure globalSumR8d3
+      module procedure globalSumR4d3
+      module procedure globalSumI8d3
+      module procedure globalSumI4d3
+      module procedure globalSumR8d4
+      module procedure globalSumR4d4
+      module procedure globalSumI8d4
+      module procedure globalSumI4d4
+      module procedure globalSumR8d5
+      module procedure globalSumR4d5
+      module procedure globalSumI8d5
+      module procedure globalSumI4d5
+      module procedure globalSumR8d6
+      module procedure globalSumR4d6
+      module procedure globalSumI8d6
+      module procedure globalSumI4d6
+      module procedure globalSumProdR8d1
+      module procedure globalSumProdR4d1
+      module procedure globalSumProdI8d1
+      module procedure globalSumProdI4d1
+      module procedure globalSumProdR8d2
+      module procedure globalSumProdR4d2
+      module procedure globalSumProdI8d2
+      module procedure globalSumProdI4d2
+      module procedure globalSumProdR8d3
+      module procedure globalSumProdR4d3
+      module procedure globalSumProdI8d3
+      module procedure globalSumProdI4d3
+      module procedure globalSumProdR8d4
+      module procedure globalSumProdR4d4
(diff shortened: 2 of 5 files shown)


philipwjones:
I have a standalone unit test driver for this module but didn't know where to put it. I'm happy to add if someone can point to a good spot in the MPAS spaces.

sbrus89:
@philipwjones, thanks for your work on this! I'm going to give it a test drive in #4472.

mark-petersen:
@philipwjones thanks for this addition. I put your test script here, and ran on badger with gnu using:
```
# in E3SM, this branch:
cd components/mpas-framework/src/framework
mpif90 mpas_globalSumMod.F sumTest.F90 -fno-range-check -ffree-form
mpirun -n 36 a.out
```
and it reported passing all the tests.

rljacob:
@sbrus89 and @mark-petersen please indicate if your review is done.

mark-petersen:
@rljacob: working on it now.

mark-petersen:
@philipwjones, MPAS stand-alone always promotes R4 to R8. In gnu, that is `-fdefault-real-8 -fdefault-double-8`. E3SM also promotes R4 to R8. The new module does not compile with that, in MPAS or with the test script:
```
mpif90 mpas_globalSumMod.F sumTest.F90 -fno-range-check -ffree-form -fdefault-real-8 -fdefault-double-8
mpas_globalSumMod.F:136:40:

       module procedure globalSumNfldR4d0
                                        1
Error: Ambiguous interfaces ‘globalsumnfldr4d0’ and ‘globalsumnfldr8d0’ in generic interface 'mpas_globalsumnfld' at (1)
```
because it does not recognize the difference between the R4 and R8 interfaces. It looks like all the other interfaces in MPAS framework only use a `real` and not R4 and R8. For example,
```
components/mpas-framework/src/framework/mpas_dmpar.F
  121    interface mpas_dmpar_alltoall_field
  122       module procedure mpas_dmpar_alltoall_field1d_integer
  123       module procedure mpas_dmpar_alltoall_field2d_integer
  124       module procedure mpas_dmpar_alltoall_field1d_real
  125       module procedure mpas_dmpar_alltoall_field2d_real
  126       module procedure mpas_dmpar_alltoall_field3d_real
  127       module procedure mpas_dmpar_alltoall_field4d_real
  128       module procedure mpas_dmpar_alltoall_field5d_real
  129    end interface
```

components/mpas-framework/src/framework/mpas_globalSumMod.F
```diff
@@ -0,0 +1,6491 @@
...
+!!        double precision. For double precision sums, we use the
+!!        double-double algorithm of Donald Knuth, further improved by
+!!        David A Bailey and presented in:
+!!        He, Yun and Chris Ding, 2001: Using Accurate Arithmetics to
+!!          Improve Numerical Reproducibility and Stability in Parallel
+!!          Applications, J. of Supercomputing, 18, 259-277.
+!
+!***********************************************************************
+
+module mpas_globalSumMod
```
mark-petersen: All other module names and file names are underscore, not camelcase, and the ending Mod is assumed i.e. `mpas_global_sum`

rljacob:
Are you sure the E3SM build autopromotes?  That's a mistake because its not supposed to and instead require all reals to be explicitly typed.

mark-petersen:
OK, thanks @rljacob. Then maybe his is a problem for stand-alone flags but not E3SM.

philipwjones:
The algorithms for R4 and R8 are completely different, so you can't autopromote (and that was one reason I didn't use MPAS kinds here).  It's also a bad idea anyway but that's another rant...   I think if you add a flag later to disable auto promote, the second flag supersedes?

components/mpas-framework/src/framework/mpas_globalSumMod.F
philipwjones: I can change to underscore, but the module name needs to be different from the generic interface name so I appended the mod.

sbrus89:
I ran `sumTest.F90` on Chrysalis using an Intel E3SM build and all tests passed. I also see the compile error @mark-petersen noticed building in stand-alone MPAS. We should make sure that build still works with this addition. Is the best way to do that removing the autopromote flags?

philipwjones:
@sbrus89 I'm trying to remove auto promote now. Unfortunately just removing it from the build results in different answers so I have to go through one by one to see where things are being promoted and need to be declared differently.

mark-petersen:
I can get MPAS stand-alone to compile if I comment all the R4 lines in `mpas_globalSumMod.F` like
```
module procedure globalSumR4d0
```
and I was able to call a globalSum routine successfully. The trouble is that , with the R4 procedures in, stand-alone does not compile with the flags in the Makefile. Specifically, every compiler has a line like this
https://github.com/E3SM-Project/E3SM/blob/32ce6e83ff824abf9481c23a4631595ef078c135/components/mpas-framework/Makefile#L337
If we do not use the auto-promotion flags, then MPAS uses single precision. See this block:
https://github.com/E3SM-Project/E3SM/blob/32ce6e83ff824abf9481c23a4631595ef078c135/components/mpas-framework/Makefile#L769-L777
It appears that the MPAS framework relies on these flags to choose between single and double precision throughout, which is why the framework function interfaces have real, not R4 and R8.

philipwjones:
Actually, MPAS has two things going on at once. The makefiles all have the auto promote, but there is a separate single precision ifdef flag that can change RKIND to r4. We really want the latter and not the former and the latter would work with this new module. So I'm trying to eliminate the auto-promotion in favor of the ifdef/RKIND approach.

rljacob:
I'm confused why commenting out the globalSumR4* procedure lines works.   That implies autopromotion is affecting the procedures where everything is explicitly typed which shouldn't happen.

sbrus89:
That sounds great, @philipwjones.

philipwjones:
Doh! @rljacob was right.  There's actually a bug in the SumNfldR4D0 interface where the real declaration was missing the explicit type.  Then it compiles.  I'll fix that and the naming and it should all work (I hope).

sbrus89:
That's a good catch @rljacob and @philipwjones. Are the auto-promotion flags something that should still be addressed in favor of the ifdef approach (perhaps in another PR)?

philipwjones:
Ok @sbrus89 @mark-petersen , you can try this again. I also submitted a PR to update the unit test in the other repo.

philipwjones:
@sbrus89 yeah, I'll continue to work on the auto-promote issue. Was partway through already.

sbrus89:
@philipwjones, I'm getting some `FAIL` results for `sumTest` on Chrysalis. 
These are for the `Reproducibility and nfld scalar double`  and `Multifield range/mask *d double` tests
All other tests pass.

Modules: 
```
intel/20.0.4-kodw73g
openmpi/4.1.1-qiqkjbu
```

Makefile:
```
global_sum.o: mpas_global_sum_mod.F
        mpif90 -free -c mpas_global_sum_mod.F -o mpas_global_sum_mod.o
sumTest.o: sumTest.F90
        mpif90 -c sumTest.F90 -o sumTest.o

test: global_sum.o sumTest.o
        mpif90 sumTest.o mpas_global_sum_mod.o -o test.out

clean:
        rm sumTest.o mpas_global_sum_mod.o test.out
```

Run:
`srun -n 36 ./test.out`

Output sample:
```
 Reproducibility and nfld scalar double: FAIL   7070601422.66856     
   7070601422.66856        7070601422.66856        7070601422.66856     
 Reproducibility and nfld scalar real: PASS
 Reproducibility and nfld scalar integer: PASS
 Reproducibility and nfld scalar integer(8): PASS
.
.
.
 Multi-field and reproducibility 1d double: FAIL   233099451231.900     
   233099451231.900        233099451231.900        233099451231.900     
 Multi-field and reproducibility 1d real: PASS
 Multi-field and reproducibility 1d int: PASS
 Multi-field and reproducibility 1d int(8): PASS
.
.
.
 Multifield range/mask 2d double: FAIL   20460671049156.0     
   20482290994957.1        20426790685330.5        20460671049155.8     
   20482290994956.9        20426790685330.2        20460671049155.8     
   20482290994956.9        20426790685330.2     
 Multifield range/mask 2d real: PASS
 Multifield range/mask 2d int: PASS
 Multifield range/mask 2d int(8): PASS
.
.
.
 Multi-field and reproducibility 3d double: FAIL  2.561585403678755E+015
  2.561585403680175E+015  2.561585403680171E+015  2.561585403680209E+015
 Multi-field and reproducibility 3d real: PASS
 Multi-field and reproducibility 3d int: PASS
 Multi-field and reproducibility 3d int(8): PASS
.
.
.
 Multi-field and reproducibility 4d double: FAIL  2.558734913839170E+016
  2.558734913875111E+016  2.558734913874659E+016  2.558734913874276E+016
 Multi-field and reproducibility 4d real: PASS
 Multi-field and reproducibility 4d int: PASS
 Multi-field and reproducibility 4d int(8): PASS
.
.
.
 Multi-field and reproducibility 5d double: FAIL  5.116267955530406E+016
  5.116267955584057E+016  5.116267955579368E+016  5.116267955581770E+016
 Multi-field and reproducibility 5d real: PASS
 Multi-field and reproducibility 5d int: PASS
 Multi-field and reproducibility 5d int(8): PASS


```

components/mpas-framework/src/framework/framework.cmake
```diff
@@ -32,4 +32,5 @@ list(APPEND COMMON_RAW_SOURCES
   framework/regex_matching.c
   framework/mpas_field_accessor.F
   framework/mpas_log.F
+  framework/mpas_globalSumMod.F
```
sbrus89, line 35: I think this should be updated with the new filename: `framework/mpas_global_sum_mod.F`. That probably explains the CI errors.

mark-petersen: But without the mod, because no other modules have that. Just `framework/mpas_global_sum.F`

philipwjones: As mentioned above, the _mod in this case is necessary to distinguish it from the function name. And the filename should reflect the module name for automated dependency generation. I'll change the cmake file.

sbrus89:
@philipwjones, The failed tests I mentioned do pass when I compile and link against the `.o` and `.mod` files generated with the E3SM build. Maybe I'm missing a compiler flag when I build independently?

ambrad:
@sbrus89 do you have `-fp-model=strict` in your compile line? Sometimes the default of `-fp-model=fast` can use rounding rules that break s/w extended-precision arithmetic.

sbrus89:
Thanks @ambrad; that was the issue. I figured that was the case, but hadn't checked to see what additional flags were used in the E3SM build (looks like `-fp-model=source` was used in my case). I tried building independently of E3SM using `-fp-model=` `source`, `precise`, and `consistent` and all tests passed for each of those cases.

philipwjones:
Ok, I added a flag to all the Makefile targets and a specific rule for the global sum routine that prevents optimizations that interfere with IEEE compliance and the reproducible algorithm.

I also pruned a few obsolete targets from the top-level makefile (blue gene, pathscale, titan, nag...)

So I think this should all be good now?

mark-petersen (approved):
Passed MPAS-O nightly regression suite on grizzly with gnu and intel. gnu matches bfb with previous, intel within machine precision. I wrote a little test of the globalSum routine within MPAS and it works correctly.

jonbob:
Testing on chrysalis shows these PASS:

merged to next

PR #4705, opened 2021-12-15
After recent PR, correct PE layout for "S" size of ne102pg2 F cases using CICE

### LINKED ISSUE #4704: Correct PE layout bug on cori-knl for "S" ne120pg2 F cases
After https://github.com/E3SM-Project/E3SM/pull/4696, where I updated PE layouts on cori-knl for F cases using ne120pg2, I made a mistake with the "S" size for CICE.  THe case would still run, but very slowly.  Will fix now.

[bfb]

ndkeen:
After recent PR 4696, correct PE layout for "S" size of ne102pg2 F cases using CICE

Fixes #4704
[bfb]

### DIFF
--- a/cime_config/allactive/config_pesall.xml
+++ b/cime_config/allactive/config_pesall.xml
@@ -6801,7 +6801,7 @@
           <ntasks_ocn>5440</ntasks_ocn>
           <ntasks_glc>64</ntasks_glc>
           <ntasks_wav>64</ntasks_wav>
-          <ntasks_cpl>85</ntasks_cpl>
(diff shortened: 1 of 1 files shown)


ndkeen:
merged to next

ndkeen:
How to get beyond the "Some checks were not successful" mesg here in github?

rljacob:
Its fine now.  I think you can force a rerun of the test if you log on to circleci and tell it to.  Or if you make a new commit it will rerun.  Or ignore it and merge with the command line.

PR #4706, opened 2021-12-15
Downgrade NetCDF lib from 4.8.1 to 4.8.0 for ANL GCE nodes

dqwu:
Some E3SM cases run on the ANL GCE nodes with the netcdf4p
type and NetCDF 4.8.1 might return NC_EHDFERR from nc_enddef(). 

So switching back to NetCDF 4.8.0, an older but more stable release.

For consistency on GCE nodes, we downgrade the serial version of the
NetCDF lib to 4.8.0 as well.

[BFB]


dqwu:
Related NetCDF issue: https://github.com/Unidata/netcdf-c/issues/2165

jayeshkrishna (approved):

Issue #4695, opened 2021-12-10
Update modules for Perlmutter after maint


ndkeen:
For GPU builds especially, NERSC updated preferred modules to load for cuda related modules.
Would currently fail at setup.

`ndk/machinefiles/perlmutter-env-update`

e3sm_developer tests pass except for
```
SMS_Ld20.f45_f45.IELMFATES.perlmutter_gnu.elm-fates_eca
SMS_Ld20.f45_f45.IELMFATES.perlmutter_gnu.elm-fates_rd
```
which get MEMLEAK

PR #4710, opened 2021-12-16
Update module environment for Perlmutter

ndkeen:
After maintenance, NERSC changed the way to load cuda modules.

Fixes https://github.com/E3SM-Project/E3SM/issues/4695


ndkeen:
merged to next

PR #4713, opened 2021-12-21
New algorithm for determining tropopause and additional diagnostic output

wlin7:
Add a new algorithm for determining tropopause: cold point parabolic.
Output at additional levels in the stratosphere are also added.
While the new algorithm is enabled during runtime, no additional variables
are saved by default.

[BFB]

### DIFF
--- a/components/eam/src/physics/cam/cam_diagnostics.F90
+++ b/components/eam/src/physics/cam/cam_diagnostics.F90
@@ -218,6 +218,7 @@ subroutine diag_init()
    call addfld ('Z200',horiz_only,    'A','m','Geopotential Z at 200 mbar pressure surface')
    call addfld ('Z100',horiz_only,    'A','m','Geopotential Z at 100 mbar pressure surface')
    call addfld ('Z050',horiz_only,    'A','m','Geopotential Z at 50 mbar pressure surface')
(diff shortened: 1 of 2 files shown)


tangq:
Hi @wlin7 , we made quite a bit changes in the chemistry branch (https://github.com/E3SM-Project/E3SM/pull/4707) to implement the E90 tropopause (both 2D and 3D) for chemistry. I noticed this PR also made changes to tropopause.F90. It would be good to make sure these tropopause won't conflict each other.

golaz (approved):
Approving by visual inspection; the changes look as requested.

wlin7:
Hi @tangq , thanks for calling this out. There will be some small conflicts that need to be resolved when #4707 is merged. The total number of algorithm will become 9. We can make either CPP or E90 the 9th. Currently both are the 8th in their separate PR. The good thing is they are using different designated letter (E for E90 and F for CPP). I noticed you assigned #4707 to me. I will take care of these when merging.

wlin7:
Merged to next.

rljacob:
Is this need for some v2 sims?

golaz:
@rljacob : yes, this is some additional diagnostics output we want to have for additional historical simulations part of a large ensemble with v2.

PR #4716, opened 2021-12-22
Update LANL scratch3 to scratch4

mark-petersen:
LANL /lustre/scratch3 will be decommissioned January 7, 2022. All data should now be written to /lustre/scratch4. This updates the machine files for LANL grizzly and badger.

[BFB]


jonbob (approved):

jonbob:
Thanks for catching that, @mark-petersen. I guess we should also add chicoma before long, but I don't think that one is as critical as this

mark-petersen:
Yes. I have an account on chicoma, I can try that after break.

jonbob:
passes sanity testing
merged to next

jonbob:
merged to master

PR #4719, opened 2021-12-29
add MMF AMIP compset

whannah1:
This adds a transient prescribed SST compset (i.e. AMIP) for the C++ MMF. We didn't bother with the fortran MMF because they can't run on GPUs, and therefore aren't good candidates for AMIP experiments. This also only implements the 1-moment microphysics case with prescribed aerosols as the new P3 implementation in the C++ MMF code is still a work in progress. Changes are BFB, but the `ERS_Ln9.ne4pg2_ne4pg2.F-MMFXX.eam-mmf_use_VT` test in `e3sm_integration` is replaced with `ERS_Ln9.ne4pg2_ne4pg2.F20TR-MMFXX.eam-mmf_use_VT` to add test coverage for the new compset, so a new baseline will need to be generated.

[BFB]


components/eam/bld/namelist_files/use_cases/20TR_MMF-1mom_CMIP6.xml
```diff
@@ -0,0 +1,45 @@
...
+<prescribed_aero_file>mam4_0.9x1.2_L72_2000clim_c170323.nc</prescribed_aero_file>
+<use_hetfrz_classnuc>.false.</use_hetfrz_classnuc>
+<prescribed_aero_type>CYCLICAL</prescribed_aero_type>
+<prescribed_aero_cycle_yr>01</prescribed_aero_cycle_yr>
+
+<!-- Stratospheric ozone (Linoz) updated using CMIP6 input4MIPS GHG concentrations -->
+<chlorine_loading_file      >atm/cam/chem/trop_mozart/ub/Linoz_Chlorine_Loading_CMIP6_0003-2017_c20171114.nc</chlorine_loading_file>
+<chlorine_loading_type      >SERIAL</chlorine_loading_type>
+
+<!-- Turn off ozone dry deposition -->
```
brhillman, line 39: I am pretty sure these lines are both unnecessary, and will cause a FPE with debug flags. This was the case for us with SCREAM. I believe since we are specifying `-chem none`, this will just enable some code that computes something that gets thrown away anyways.

brhillman (approved):
Thanks for doing this!

brhillman:
Merged to next.

brhillman:
Re-merged to next with fix for land use case file.

brhillman:
Remerged to next.

brhillman:
Merged to master.

PR #4720, opened 2022-01-03
Global carbon budget in EAM

beharrop:
Add the capability to check CO2 conservation in EAM over various time periods.  The three time periods (time step, monthly, total run) are all controlled by namelist flags (monthly and total are on by default).  If the relative error exceeds the threshold (default value is 5e-13), then the simulation will exit.  A summary table of the CO2 conservation is output into the atmosphere log file.

In EAM loop (surf -> tphysAC -> dycore -> tphysBC ->surf), CO2 surface fluxes and aircraft emissions are stored in tphysAC (immediately after they are applied to the CO2 tracers.  Current CO2 tracer mass is computed immediately after tphysAC (in phys_run2), then the diagnostics checks for conservation.  After conservation check, (still in phys_run2) the current CO2 is stored as the "previous timestep" value to be used in subsequent checks.

[BFB]


susburrows:
Thanks for getting this submitted, Bryce!  I will have a closer look soon.

driver-mct/cime_config/config_component_e3sm.xml
```diff
@@ -270,6 +270,7 @@
       <value compset="_MPASSI%BGC.*_MPASO%OIECO.*_BGC%BCRD" >CO2C_OI</value>
       <value compset="_MPASSI%BGC.*_MPASO%OIECO.*_BGC%BDRC" >CO2C_OI</value>
       <value compset="_MPASSI%BGC.*_MPASO%OIECO.*_BGC%BDRD" >CO2C_OI</value>
+      <value compset="_BGC%LNDATM">CO2B</value>
```
susburrows, line 273: Does this change belong in this PR?

components/eam/bld/namelist_files/namelist_defaults_eam.xml
```diff
@@ -426,7 +426,12 @@
 <flbc_file chem="waccm_mozart_sulfur">atm/waccm/lb/LBC_1765-2005_1.9x2.5_CMIP5_za_c111110.nc</flbc_file>
 
 <!-- Time-variant CO2 fossil fuel emissions -->
-<co2flux_fuel_file  sim_year="1850-2000">atm/cam/ggas/co2flux_fossil_1751-2006-monthly_0.9x1.25_c20100204.nc</co2flux_fuel_file>
+<!-- <co2flux_fuel_file  sim_year="1850-2000">atm/cam/ggas/co2flux_fossil_1751-2006-monthly_0.9x1.25_c20100204.nc</co2flux_fuel_file> -->
```
susburrows, line 429: The changes here and in the following lines update the CO2 emission files - do they belong in this mass conservation PR?

components/eam/bld/build-namelist
```diff
@@ -922,10 +927,9 @@ if ($co2_cycle) {
 	# (user specification has higher precedence than the true value set above)
 	if ($nl->get_value('co2_readflux_fuel') =~ /$TRUE/io) {
 	    add_default($nl, 'co2flux_fuel_file', 'sim_year'=>$sim_year);
-	    add_default($nl, 'co2flux_fuel_vrt_emis_file', 'sim_year'=>$sim_year);
```
susburrows, line 925: Do the changes to the handling of aircraft emissions belong in this PR?

susburrows (commented):
Bryce, thanks for putting together this PR.  I visually examined the changes and I see a few that are related to the handling of aircraft emissions, and to the specification of emission input files.  I am wondering if those belong in this particular PR or not, it's not clear to me if they are required as part of the mass conservation diagnostics.

driver-mct/cime_config/config_component_e3sm.xml
beharrop: Oops.  No, I added these to do more testing with this sort of configuration, but they aren't necessary.

components/eam/bld/namelist_files/namelist_defaults_eam.xml
beharrop: Same as before; this is not needed.

components/eam/bld/build-namelist
beharrop: Same as before; this is not needed.

components/eam/src/physics/cam/physpkg.F90
```diff
@@ -1340,6 +1343,24 @@ subroutine phys_run2(phys_state, ztodt, phys_tend, pbuf2d,  cam_out, &
...
 #endif
 
+    !
+    ! Check for carbon conservation
+    !
+    do c = begchunk, endchunk
+       call get_total_carbon(phys_state(c), 'wet')
+    end do
+    call print_global_carbon_diags(phys_state, ztodt, nstep)
+    do c = begchunk, endchunk
```
oksanaguba, line 1353: does this block need to be wrapped in if (co2_transport) like other bocks using new C diagnostics variables?

oksanaguba:
Could I suggest to add a test with this PR if it is easy? Is co2_transport set to true in AMIP-type simulations? We might add a test without a new build. 

Did you decide not to attach a python script to this diagnostics?

If it is not too much effort, could a description of the loop surf -> tphysAC -> dycore -> tphysBC ->surf with inserts on where C is read, init-ed, diagnosed, etc. be added to this PR? If we decide to revisit other diagnostics, it would help to remember the framework in this PR. Thanks!

components/eam/src/physics/cam/physics_types.F90
```diff
@@ -103,7 +103,23 @@ module physics_types
           te_ini,  &! vertically integrated total (kinetic + static) energy of initial state
           te_cur,  &! vertically integrated total (kinetic + static) energy of current state
           tw_ini,  &! vertically integrated total water of initial state
-          tw_cur    ! vertically integrated total water of new state
+          tw_cur,  &! vertically integrated total water of new state
+          tc_curr,    &! vertically integrated total carbon of current state
+          tc_init,    &! vertically integrated total carbon at start of run
```
oksanaguba, line 108: This is very minor, but the old variable te_ini, I think, is intended to be used as TE per physics time step, not per simulation (though it is probably not used at all). now there is tc_init, with a very similar name, but it is a variable for the initial state of the whole simulation.  Just. a name convention to be aware of.

components/eam/src/physics/cam/physpkg.F90
beharrop: That's a good idea.  While the calls to `get_total_carbon` and `print_global_carbon_diags` will return immediately if co2_transport is false, we would still be setting `tc_prev`, `tc_init`, `tc_mnst` needlessly.  I will add in the `co2_transport` if block.

components/eam/src/physics/cam/physpkg.F90
```diff
@@ -1340,6 +1343,24 @@ subroutine phys_run2(phys_state, ztodt, phys_tend, pbuf2d,  cam_out, &
...
+       ncol = get_ncols_p(c)
+       phys_state(c)%tc_prev(:ncol) = phys_state(c)%tc_curr(:ncol)
+       if (is_first_step()) then
+          phys_state(c)%tc_init(:ncol) = phys_state(c)%tc_curr(:ncol)
```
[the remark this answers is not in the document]
beharrop: I had not previously run any restart tests.  I performed one a few moments ago and it caused the model to crash immediately.  I will investigate what is happening there and address it.

components/eam/src/physics/cam/physics_types.F90
beharrop: Yea... that is a bit confusing.  Any suggestions on a change?  I selected the "_init" suffix because it seemed to be in line with how "init" gets used in other pieces of code, but I am open to changing it if it helps avoid confusion in the future.

beharrop:
Thanks for the suggestions @oksanaguba.  With respect to the python test script, I was under the impression someone from the infrastructure team was going to make tests for all of the carbon conservation checks.  As I recall, that was why it was requested that the output tables for CO2 in EAM match the carbon tables in ELM.  @susburrows am I remembering this correctly?  If I am mistaken, I can put together a python script that reads the table and evaluates whether conservation is satisfactory.

I will add some more notes to the PR about where the checks occur within an EAM loop.

components/eam/src/physics/cam/physpkg.F90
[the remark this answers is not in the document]
oksanaguba: I think a run with co2_transport=false would also crash, since nans are used in init of variables. just a comment to run tests with co2_transport set to true and false.

beharrop:
I have made some additional updates that are capable of handling restarts.  I write the prior carbon amounts and accumulated fluxes to the physics buffer on the last step of a run, and then on the first step of a restart, those fields are read back in.  I made the choice to have the "full run" output at the end of each job, but it always compares with the initial value of the first run.  I considered adding a flag to allow an option for resetting the initial CO2 value, but I do not currently have that set up.

beharrop:
I have run a few additional tests as requested.

- The model can restart successfully, and keeps track of accumulated CO2 fluxes and the initial CO2 as desired.
- The model can run successfully with `co2_transport = false`, but only when `_BGC` is removed from the compset and `co2_cycle` is not included in `CAM_CONFIG_OPTS` otherwise errors occur with some parts of the code expecting 44 tracers and others only expecting 40.
- In DEBUG mode, a bug was uncovered that required moving `co2_diags_init` outside of the `co2_transport` if-block to avoid errors associated with those values never being initialized.  A complete DEBUG test, however, was not possible because the land model fails on Compy with a floating invalid error.  I have been talking with @bishtgautam on how to get past this issue.

beharrop:
An update, after help from @peterdschwartz for getting past the ELM issues in DEBUG mode (see most recent commit), I have now successfully run in DEBUG mode with my branch.  I ran DEBUG cases both with and without the CO2 diagnostics turned on, and both were able to run.

Please let me know if any additional tests or checks are desired.

susburrows:
Great news, thanks, Bryce.

rljacob:
@susburrows @singhbalwinder @oksanaguba please approve.

susburrows (approved):

jonbob:
@beharrop - it would also be great if we could get the time-averaging issue resolved, either in this PR or a follow-on one, so it would be more easily comparable to the coupler carbon budget

beharrop:
To elaborate on the time-averaging issue @jonbob raised, in the atm logfile printouts, the averages do not appropriately factor in the averaging time-interval length.  The time-integrated printouts are still correct, so the conservation checks work as intended.

Like @jonbob said, it would be nice to have for comparison with the coupler budget.  I can add this in a follow-up PR where I will also plan to make the long-term error threshold dependent on the length of the simulation.  Please let me know if you would prefer to have that fix in this PR and I will move it up in my queue.

jonbob:
Thanks @beharrop for adding the description, which I forgot to do. If nothing else, maybe you could get rid of the time-averaged output, since it's not correct and actually makes comparison with the coupler budget output more confusing, at least for the time being?

beharrop:
I have gone ahead and fixed the averaging time interval that @jonbob noted above.  I have also implemented the error tolerance to grow in time as suggested by @ambrad.  I've tested these changes relative to what the previous commits: they are BFB and the average values in the printouts now match expectations.

susburrows:
@wlin7 @rljacob : Looks like Bryce has addressed all the comments -- is this ok to be merged now?  @singhbalwinder is unexpectedly out of commission for a little while, but communicated with us offline that he reviewed an earlier version and ok with approving this.

cime_config/machines/cmake_macros/intel.cmake
```diff
@@ -27,7 +27,7 @@ if (compile_threaded)
   string(APPEND FFLAGS " -qopenmp")
 endif()
 if (DEBUG)
-  string(APPEND FFLAGS " -O0 -g -check uninit -check bounds -check pointers -fpe0 -check noarg_temp_created")
+  string(APPEND FFLAGS " -O0 -g -check uninit -check bounds -check pointers -check noarg_temp_created")
```
oksanaguba, line 30: Just noticed this -- why is this PR changing flags?
Same about intel_compy.cmake .

cime_config/machines/config_compilers.xml
```diff
@@ -321,7 +321,7 @@ flags should be captured within MPAS CMake files.
   <FFLAGS>
     <base>  -convert big_endian -assume byterecl -ftz -traceback -assume realloc_lhs -fp-model source </base>
     <append compile_threaded="TRUE"> -qopenmp </append>
-    <append DEBUG="TRUE"> -O0 -g -check uninit -check bounds -check pointers -fpe0 -check noarg_temp_created </append>
+    <append DEBUG="TRUE"> -O0 -g -check uninit -check bounds -check pointers -check noarg_temp_created </append>
```
oksanaguba, line 324: Same here. Is it the problem you mentioned with land? Could you then please explain in more detail, I am rather sure such PR should not touch flags (esp. DEBUG flags below). Thanks!

cime_config/machines/cmake_macros/intel.cmake
beharrop: Thanks for looking at this closely @oksanaguba.  The flag changes for Compy were needed to run the model in DEBUG mode (courtesy of @peterdschwartz).  Without these changes, I could not test my results in DEBUG mode.  These changes aren't needed for checking carbon conservation in EAM, so if it makes sense, we could separate these issues into a different PR.  There was also a bug found in the land portion (again by @peterdschwartz) that was causing issues in DEBUG mode.

oksanaguba:
Thanks @beharrop , sorry I should be more clear -- which configuration did you run, as an example, I was asking for a test like this: PET_Ln9_PS.ne30pg2_EC30to60E2r2.WCYCL1850.allactive-mach-pet or something else?

Regarding adding a test -- I do not mean to ask to test correctness of your implementation. I just want to understand which code base is active in which runs. When you PR is merged, which value is co2_transport set to in the tests that we will have on cdash? What is the value of co2_cycle? What do these values control? Sorry if you have already explained it to me, it has been a while.

Also, since you added a value to pbuff, in case co2_transport is true by default, does it mean all tests will fail when this PR is merged? Another reason to run most of the tests manually (this is a suggestion to an integrator).

Regarding flags: Are you saying that ELM could not run with DEBUG regardless your PR? If so, is it never tested? I see at least 1 cdash test with ELM and debug. You touched only 1 ELM line in this PR, maybe, you do not need flags anymore?

Sorry I am not very clear, but it seems to me that the cases are:

1) (unlikely) ELM needs changes in DEBUG flags to run from master. Make a new small PR, someone from performance should approve it. Remove change of flags from this PR. Merge flags PR first, before this PR.

2) only your PR needs new DEBUG flags. Let's figure out why (since it is just 1 line of code) and fix the code instead of flags. Revert flags in this PR back.

3) your PR does not need these flags anymore. Remove them then.

Thanks again!

cime_config/machines/cmake_macros/intel.cmake
oksanaguba: Thanks @ndkeen I agree. See also that flags are changed in config_machines.

peterdschwartz: Well as i describe in the commit, the floating point exception debug flags hinder debugging for ELM due to the NaN logic in the code, so I don't think they should be turned on by default at least.   
To allow other components to use it, then I could put a if statement to check the component name inside the DEBUG if statment.

oksanaguba: @peterdschwartz what does 'hinder' mean in this case? Would you please tell which scenario we are dealing with here:
```
Sorry I am not very clear, but it seems to me that the cases are:

(unlikely) ELM needs changes in DEBUG flags to run from master. Make a new small PR, someone from performance should approve it. Remove change of flags from this PR. Merge flags PR first, before this PR.

only your PR needs new DEBUG flags. Let's figure out why (since it is just 1 line of code) and fix the code instead of flags. Revert flags in this PR back.

your PR does not need these flags anymore. Remove them then.
```

peterdschwartz: @oksanaguba You are mistaken about @beharrop initially modifying elm files in this PR.  I was only able to fix that bug after removing the debug flags in question and CNPBudgetMod.F90 appears to have been last modified in master in Nov.  So that's what I mean by  it hindering debugging.  Moreover, I just reran a test case on compy after re-adding the `-fpe0` flag and can confirm that it fails again. 
I agree that these floating point exceptions raise questions about the consistencies of use of NaNs with ELM as I have noticed this issue several times in the past couple years which is why I knew how to fix it.  I think that not compiling elm files with fpe0 or similar flags is a good choice in the short term.

oksanaguba: @peterdschwartz I refreshed browser and still saw 1 elm file that this PR touched, components/elm/src/biogeochem/CNPBudgetMod.F90 , and it changes DEBUG flags. It is true I do not understand what's going on. Would you please tell me which case we are dealing with here, see below. Sorry that I keep asking this, but the list below (it seems to me) covers all possible cases why the flag changes are in this PR.

1.(unlikely) ELM needs changes in DEBUG flags to run from master. Make a new small PR, someone from performance should approve it. Remove change of flags from this PR. Merge flags PR first, before this PR.

2.only your PR needs new DEBUG flags. Let's figure out why (since it is just 1 line of code) and fix the code instead of flags. Revert flags in this PR back.

3.your PR does not need these flags anymore. Remove them then.

peterdschwartz: > @peterdschwartz I refreshed browser and still saw 1 elm file that this PR touched, components/elm/src/biogeochem/CNPBudgetMod.F90 , and it changes DEBUG flags. It is true I do not understand what's going on.

In my previous comment, I was trying to clarify the timeline of commits.  With your cases listed, then I say it is Case 1 -- I don't know why you think it's unlikely as there are multiple configs and gaps in testing. 
1. @beharrop  experienced fails in DEBUG mode **without** changing any elm files.  
2. The debug mode fails pointed to lines of code not related to CNPBudgetMod.F90 
3. Me removing the related debug flags resulted in a much more informative backtrace that showed there was a bug in CNPBudgetMod.F90 which was last changed in **master** in Nov.
4. Adding back -fpe0 flag **after** fixing the typo in CNPBudgetMod.f90 still results in fails.
5. The fails are not due to this PR and maybe should not even be considered bugs since ELM relies on NaNs always evaluating to **false** in certain parts of the code.  It's more of a question of coding paradigm here i think.

If this does not clarify, I can try to talk with you next week sometime.

oksanaguba: Thanks! Then flags should go into a separate PR, and as you suggested, they should be safeguarded by an if-statement for elm. Though I do not know if that works for files like intel.cmake, etc. 

Would you be willing to make such PR?

Another option is to remove flags from here and leave debug flags in master as is. There are _D cases in tests, but they do not trigger this issue. This is another reason to document, what exactly was run for this PR.

rljacob:
Yes this can be merged without @singhbalwinder's review.

I'm still looking at the other comments on the DEBUG flag but the -fpe0 option should not be removed in this PR.  The debug tests we have with ELM all pass.

peterdschwartz:
Ok since my commit also fixes a bug in master with the budgets I will just make a new commit adding the flag to back.  I will then run some tests and debug mode and document which ones fail and continue the discussion there.

peterdschwartz:
"-fpe0" has been re-added and ERS.f09_g16.I1850ELMCN FAILed in DEBUG mode as expected on chrysalis with intel.  As I previously stated, I'll start an issue for the debug fails to discuss the best way to handle them.

oksanaguba:
@peterdschwartz -- I see other changes in flags in this PR. They need to be reverted completely, too. Changing flags, if ever needed, should be a separate PR, reviewed by performance group.

peterdschwartz:
Compy flags are added back -- the diff showing is due to a typo that is currently fixed in master.

beharrop:
e3sm_integration test suite now passes the tests highlighted in my previous post (the two bgc tests and the pg2 mass test).  There are two other tests that fail still, but these also fail with current master.

- ERP_Ld3.ne30pg2_EC30to60E2r2.WCYCL1850.compy_intel.allactive-pioroot1
- PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.compy_intel

Other failures occur when the FLEXlm library fails to checkout on Compy.  These failures occur sporadically during build on Compy and are not reproducible.

Unless there are any other tests desired from the integrators, I'd say this is good to go.  Thanks everyone for the help.

rljacob:
Yes. @wlin7 you can start merging this.

wlin7:
Merged to next. 

Basic SMS tests ok (done on compy). conflict in build-namelist is minimal, a comment line and the assignment of val to aircraft_type.

wlin7:
Merged to master.

rljacob:
@wlin7 bless requests need to filed for this test: SMS_Ld9.ne4pg2_oQU480.F2010.*_*.eam-thetahy_sl_pg2_mass for namelist diffs.

oksanaguba:
@beharrop -- regarding Rob's comment above, do you know why that test is affected? https://my.cdash.org/test/52807221 

```
Comparison failed between '/lcrc/group/e3sm/e3smtest/scratch/chrys/J/SMS_Ld9.ne4pg2_oQU480.F2010.chrysalis_intel.eam-thetahy_sl_pg2_mass.C.JNextIntegration20220320_223415/CaseDocs/atm_in' with '/lcrc/group/e3sm/baselines/chrys/intel/master/SMS_Ld9.ne4pg2_oQU480.F2010.chrysalis_intel.eam-thetahy_sl_pg2_mass/CaseDocs/atm_in'
Differences in namelist 'co2_cycle_nl':
  found extra variable: 'co2_conserv_error_tol_per_year'
  found extra variable: 'co2_print_diags_monthly'
  found extra variable: 'co2_print_diags_timestep'
  found extra variable: 'co2_print_diags_total'
```

beharrop:
@oksanaguba the "mass" test enables the co2 tracers, which should mean the co2_cycle_nl is active.  Since the co2 conservation namelist options were also added to co2_cycle_nl, those differences are expected.  Linking @ambrad in case he has any further clarifications related to that specific test.

wlin7:
> @wlin7 bless requests need to filed for this test: SMS_Ld9.ne4pg2_oQU480.F2010.___.eam-thetahy_sl_pg2_mass for namelist diffs.

Thanks for  tracking this, @rljacob . I missed it. Bless request submitted.

rljacob:
@wlin7 SMS_Ld9.ne4pg2_oQU480.F2010.chrysalis_intel.eam-thetahy_sl_pg2_mass on chrysalis next is still complaining about a missing baseline

ambrad:
Is it possible the baseline file was deleted? This test has been around for a while and so should have a baseline file already.

rljacob:
maybe.  Master didn't run last night but its last run (march 25) still showed the namelist diff.

rljacob:
I see.  This test only had a namelist diff and not a real diff.  So yeah something must have happened to the baseline.

wlin7:
The baseline directory for this test only has user_nl files remaining.

This PR was merged to next on March 17. The test return the following morning was [recording comparison of cpl.hi file with baseline](https://my.cdash.org/test/52655683) and was all good. [Comparison was still good on March 24.
](https://my.cdash.org/test/52937273).  The test was resume March 30, when it shows missing baseline.

In between, the NML diff was blessed on March 25. Could the blessing do something funny? Or something on the system side during the pause days happen?

It is safe we might just re-create a new baseline for it.

rljacob:
Yes just make another bless request.

wlin7:
> Yes just make another bless request.

Just did.

PR #4721, opened 2022-01-04
HOMME remove obsolete code

mt5555:
Remove a lot of obsolete code
three minor bug fixes:
        lap_p_wk array dimension, 
        Nthreads=-1 logic:    only enable threads if HORIZ_OPENMP or COLUMN_OPENMP enabled
        parseE3.py typo

[BFB]

### DIFF
--- a/components/homme/src/preqx/derivative_mod.F90
+++ b/components/homme/src/preqx/derivative_mod.F90
@@ -6,7 +6,7 @@ module derivative_mod
   use derivative_mod_base, only: derivative_t, subcell_integration, subcell_dss_fluxes, subcell_div_fluxes,                  &
        subcell_Laplace_fluxes, allocate_subcell_integration_matrix, derivinit, gradient, gradient_wk, vorticity, divergence, &
        gradient_sphere_wk_testcov, gradient_sphere_wk_testcontra, ugradv_sphere, vorticity_sphere, vorticity_sphere_diag,    &
-       curl_sphere,curl_sphere_wk_testcov, vlaplace_sphere_wk, element_boundary_integral, edge_flux_u_cg,                    &
+       curl_sphere,curl_sphere_wk_testcov, vlaplace_sphere_wk, element_boundary_integral, &
        limiter_optim_iter_full, limiter_clip_and_sum, laplace_sphere_wk, divergence_sphere_wk, gradient_sphere,              &
        divergence_sphere, laplace_z
   implicit none
--- a/components/homme/src/theta-l/share/prim_advance_mod.F90
+++ b/components/homme/src/theta-l/share/prim_advance_mod.F90
@@ -1751,6 +1751,7 @@ contains
         enddo
 
         ! check for layer spacing <= 1m
+        if (scale3 /= 0) then
         do k=1,nlev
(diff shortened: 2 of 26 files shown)


components/homme/src/theta-l/share/prim_advance_mod.F90
```diff
@@ -1755,6 +1752,7 @@ subroutine compute_andor_apply_rhs(np1,nm1,n0,qn0,dt2,elem,hvcoord,hybrid,&
         enddo
 
         ! check for layer spacing <= 1m
+        if (scale3 /= 0) then
```
oksanaguba, line 1755: When is scale3 zero? I looked at a couple of types, it is always 1?

components/homme/src/tool/CMakeLists.txt
```diff
@@ -72,4 +72,3 @@ tool_setup()
 ############################################################################
 createTestExec(homme_tool tool ${PREQX_NP} ${PREQX_NC} ${PREQX_PLEV} ${PREQX_USE_PIO} ${PREQX_USE_ENERGY} ${QSIZE_D})
 
-TARGET_LINK_LIBRARIES(homme_tool dl)
```
oksanaguba, line 75: From memory, I think -ldl was an issue on knl?

components/homme/src/preqx/derivative_mod.F90
```diff
@@ -6,7 +6,7 @@ module derivative_mod
   use derivative_mod_base, only: derivative_t, subcell_integration, subcell_dss_fluxes, subcell_div_fluxes,                  &
        subcell_Laplace_fluxes, allocate_subcell_integration_matrix, derivinit, gradient, gradient_wk, vorticity, divergence, &
        gradient_sphere_wk_testcov, gradient_sphere_wk_testcontra, ugradv_sphere, vorticity_sphere, vorticity_sphere_diag,    &
-       curl_sphere,curl_sphere_wk_testcov, vlaplace_sphere_wk, element_boundary_integral, edge_flux_u_cg,                    &
```
oksanaguba, line 9: Just curious, what was edge_flux_u_cg for?

oksanaguba (approved):

oksanaguba:
This is not a PR on A,B coefficients, which I assume would follow?

components/homme/src/theta-l/share/prim_advance_mod.F90
mt5555: its used by sundials/arkode - it's the interface where compute_andor_apply_rhs() returns tendencies, instead of update state.

components/homme/src/tool/CMakeLists.txt
mt5555: removed because this is the wrong place to add libraries and it will be ignored by all test_exec's

components/homme/src/preqx/derivative_mod.F90
mt5555: i think it was used in the DG code

oksanaguba:
Please, do not merge into master till homme baselines on skybridge are updated.

PR #4722, opened 2022-01-04
HOMME remap code

mt5555:
Remove obsolete/unused remap algorithms
add q_alg=11  ( from @ambrad - more accurate, doesn't impose monotonicity at boundaries)
added option to specify dynamics algorithm independently from tracer algorithm (from Jason Torchinsky summer intern project)

q_alg=11 fixes long standing issue with noise in HOMME's internal gravity wave test cases

[BFB] except for one HOMME test, which was modified to test the new code.


components/homme/src/share/vertremap_base.F90
```diff
@@ -610,26 +610,25 @@ subroutine remap_Q_ppm(Qdp,nx,qsize,dp1,dp2)
           masso(k+1) = masso(k) + ao(k) !Accumulate the old mass. This will simplify the remapping
           ao(k) = ao(k) / dpo(k)        !Divide out the old grid spacing because we want the tracer mixing ratio, not mass.
         enddo
-        !Fill in ghost values. Ignored if vert_remap_q_alg == 2
-        if (vert_remap_q_alg >= 1 .and. vert_remap_q_alg <= 3) then
+        !Fill in ghost values.
+        if (remap_alg==10) then
```
ambrad, line 614: I recommend a slightly cleaner approach here to avoid code duplication:
```F90
logical :: limit_linear_extrap
...
if (remap_alg == 10 .or. remap_alg == 11) then
  limit_linear_extrap = remap_alg == 10
  if (limit_linear_extrap) then
    ext(1) = minval(ao(1:nlev))
    ext(2) = maxval(ao(1:nlev))
  end if
  call linextrap(dpo(2), dpo(1), dpo(0), dpo(-1), ao(2), ao(1), ao(0), ao(-1), limit_linear_extrap, ext(1), ext(2))
  call linextrap(dpo(nlev-1), dpo(nlev), dpo(nlev+1), dpo(nlev+2),&
      ao(nlev-1), ao(nlev), ao(nlev+1), ao(nlev+2), limit_linear_extrap, ext(1), ext(2))
end if
```
Then make the third-to-last argument a `logical` in `linextrap`.

ambrad (changes requested):

components/homme/src/share/vertremap_base.F90
mt5555: good idea - now done.

ambrad (approved):

mt5555:
@ambrad : can you re-review this PR?  I had to delete a bunch of cxx code related to q_alg=2,3 that were being tested in the HOMMEBFB suite.

ambrad (approved):

components/homme/src/preqx_kokkos/cxx/cxx_f90_interface_preqx.cpp
```diff
@@ -63,10 +63,6 @@ void init_simulation_params_c (const int& remap_alg, const int& limiter_option,
 
   if (remap_alg==1) {
     params.remap_alg = RemapAlg::PPM_MIRRORED;
-  } else if (remap_alg == 2) {
```
oksanaguba, line 66: does this supposed to have alg=10?

oksanaguba:
to summarize: 
-- this adds alg=11 to F code with TC test using it, but not to cxx code or tests
-- removes alg2,3 from F and cxx code
-- removes alg2,3 from testing in unit tests, adds alg10 to unit tests (there was some check preventing alg10 to run)
-- adds param vert_remap_u_alg in F code with some logic, but not to cxx code, and there is no test yet for such option
-- alg1,10 are also tested in cxx-vs-f

ambrad:
@oksanaguba, re: point 3, qalg 10 is unit tested in compose_ut.cpp, but we should add test blocks for 10 and now 11 to ppm_remap_ut.cpp at some point. Re: point 4, I think there is a test in this PR that runs with qalg 11.

components/homme/src/preqx_kokkos/cxx/cxx_f90_interface_preqx.cpp
ambrad: This is for just preqx, so it doesn't matter since we're not adding new features to preqx.

PR #4723, opened 2022-01-05
add simplified sponge layer

mt5555:
Add a simplified sponge layer with namelist variable (tom_sponge_start) to set the starting layer

This is needed to run at high resolution with the E3SM 72L configuration (72L sponge layer is too thin, it starts at the 1mb layer).  72L configurations will crash at NE256 and higher resolution without making the sponge layer thicker.  

It will also be useful for the SCREAM 128L configuration, where the sponge layer is most likely too thick (it starts at the 20mb layer)

default: tom_sponge_start=0 (disabled).  

documentation:
https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/2967798203/EAM+Top+of+Model+Sponge+Layer

[BFB]

### DIFF
--- a/components/homme/src/theta-l/share/model_init_mod.F90
+++ b/components/homme/src/theta-l/share/model_init_mod.F90
@@ -23,7 +23,7 @@ module model_init_mod
   use viscosity_mod,      only: make_c0_vector
   use kinds,              only: real_kind,iulog
   use control_mod,        only: qsplit,theta_hydrostatic_mode, hv_ref_profiles, &
-       hv_theta_correction
+       hv_theta_correction, tom_sponge_start
   use time_mod,           only: timelevel_qdp, timelevel_t
   use physical_constants, only: g, TREF, Rgas, kappa
   use imex_mod,           only: test_imex_jacobian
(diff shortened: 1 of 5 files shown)


components/homme/src/theta-l/share/model_init_mod.F90
```diff
@@ -121,35 +121,28 @@ subroutine model_init2(elem,hybrid,deriv,hvcoord,tl,nets,nete )
...
+             ptop_over_press = hvcoord%etam(1) / hvcoord%etam(k)  
+          else
+             ptop_over_press = hvcoord%etai(1) / hvcoord%etam(k)  
+          endif
+          ! active for p<10*ptop (following cd_core.F90 in CAM-FV)
+          ! CAM 26L and 30L:  top 3 levels 
+          ! E3SM 72L:  top 6 levels
+          !original cam formula
+          !nu_scale_top(k) = 8*(1+tanh(log(ptop_over_press))) ! active for p<4*ptop
+          nu_scale_top(k) = 16*ptop_over_press**2 / (ptop_over_press**2 + 1)
```
oksanaguba, line 137: Would you please replace all **2 with multiplication? I do not know how compilers would interpret those but it would be good to avoid pow issues.

oksanaguba:
We will add a homme test for the new option in the bfb suite, cxx code, and a corresponding test.

wrliugit (approved):

PR #4733, opened 2022-01-12
Make surfaceTemperatureFluxWithoutRunoff private in omp pragma

### LINKED ISSUE #4732: MPAS_LISIO_TEST tests are not reproducible on chrysalis in debug mode
@xylar first found this while testing an E3SM PR. He ran an SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.chrysalis_intel test from the hash the PR branch was created from to create a baseline, and then ran a test with the PR branch to compare. The DIFF was more than expected, so at some point he ran generate and compare tests using master -- and those also showed DIFFs.

I've spent some time trying to track this down. First I did similar tests with different compsets in an attempt to pinpoint the source of the problem:
* SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST  FAIL
* SMS_D.T62_oQU120.GMPAS-IAF  FAIL
* SMS_D.T62_oQU120.CMPASO-IAF  FAIL
* SMS_D.ne30_f19_g16_rx1  PASS
* SMS_D.T62_oQU120.DTESTM  PASS

which seems to show that mpas-o is the component causing problems.

I also ran the same tests on anvil in debug and non-debug mode, as well on non-debug on chrysalis. All passed except for debug mode on chrysalis.

To make sure it wasn't somehow a problem with creating and comparing baseline tests, I did two things. First, I made three separate compare tests, all of which failed. I used cprnc to DIFF the cpl history files from all four tests (baseline and three comparisons) and all were different. Next, I setup two identical T62_oQU120.CMPASO-IAF debug cases and ran them. They also had the same kind of DIFFs. I ran them a few timesteps, dumping out history files every step as well as turning INFO_DBUG to 3. The comparison of cpl log files shows the first thing the coupler sees that is different is being received from the ocean:
```
1697,1702c1697,1702
< comm_diag xxx sorr   1 1.0389304928185756800E+17 recv ocn So_t
< comm_diag xxx sorr   2 1.2365618737287848000E+16 recv ocn So_s
< comm_diag xxx sorr   3-1.8364725259087248535E+12 recv ocn So_u
< comm_diag xxx sorr   4 1.4324605296292727661E+11 recv ocn So_v
< comm_diag xxx sorr   5-1.8738248954403150128E+05 recv ocn So_dhdx
< comm_diag xxx sorr   6 1.2781983337872917764E+06 recv ocn So_dhdy
---
> comm_diag xxx sorr   1 1.0389304929571107200E+17 recv ocn So_t
> comm_diag xxx sorr   2 1.2365618772679004000E+16 recv ocn So_s
> comm_diag xxx sorr   3-1.8364926468792946777E+12 recv ocn So_u
> comm_diag xxx sorr   4 1.4324325288994436646E+11 recv ocn So_v
> comm_diag xxx sorr   5-1.8738249293155496707E+05 recv ocn So_dhdx
> comm_diag xxx sorr   6 1.2781983335010767914E+06 recv ocn So_dhdy
1710,1711c1710,1711
< comm_diag xxx sorr  14-5.2000177344548851562E+13 recv ocn Faoo_h2otemp
< comm_diag xxx sorr  15-1.2339505129085204480E+19 recv ocn Fioo_q
---
> comm_diag xxx sorr  14-5.1747753873219843750E+13 recv ocn Faoo_h2otemp
> comm_diag xxx sorr  15-1.2339505156405143552E+19 recv ocn Fioo_q
```
At this point I've created this issue and will try to figure out which commit first introduced the problem.

jonbob:
This PR adds surfaceTemperatureFluxWithoutRunoff to an omp pragma to be private for a loop in the mpas-ocean routine ocn_surface_bulk_forcing_active_tracers. It was causing some tests to be irreproducible when run with threading on.

Fixes #4732
[non-BFB] for some threaded tests

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_surface_bulk_forcing.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_surface_bulk_forcing.F
@@ -540,7 +540,7 @@ contains
       ! indices on tracerGroup are (iTracer, iLevel, iCell)
       if (.not. bulkThicknessFluxOff) then
          !$omp parallel
-         !$omp do schedule(runtime)
+         !$omp do schedule(runtime) private(surfaceTemperatureFluxWithoutRunoff)
          do iCell = 1, nCells
 
            ! Accumulate fluxes that use the surface temperature


jonbob:
Tested with SMS_D.T62_oQU120.CMPASO-IAF.chrysalis_intel runs that previously failed DIFF tests

rljacob:
In another PR, how about adding SMS_D.T62_oQU120.CMPASO-IAF to e3sm_integration?

jonbob:
@rljacob - thanks for the suggestion and I will do that

philipwjones (approved):
Approved based on visual inspection.

xylar:
@jonbob, this is some impressive detective work!  Thanks for figuring it out.

mark-petersen (approved):
Yes, @jonbob this really was good detective work! `surfaceTemperatureFluxWithoutRunoff` should definitely be a private variable here.

jonbob:
Thanks @xylar and @mark-petersen - I'm just glad it's fixed, such a weird problem. Thanks @xylar for uncovering it

jonbob:
merged to next

jonbob:
merged to master

mark-petersen:
@jonbob and @rljacob the corrected lines in this PR are run for all E3SM cases, because `config_use_bulk_thickness_flux = .true.` is the E3SM default, and that runs this block of code. This correction needs to included in all V2 simulations. How can that be done? Is there a V2 maintenance branch?

rljacob:
This problem only shows up when debugging is on.  It was merged to master and didn't cause anything to diff.  There have been no threading reproducibility problems seen on master.

Its too late to add anything for WC v2 sims.  Those are all done.

We don't yet have a maintenance branch because master is still producing v2.0 answers.

rljacob:
Does this also need NTHRDS >1?  We have tests with debugging and fully coupled configs but they never showed any diffs.  They seem to all have NTHRDS=1 on Chrysalis.

@amametjanov Do the default pelayouts for WC cases on Chrysalis not have NTHRDS>1 ?

jonbob:
This does need to be threaded to have an impact, but the debug setting seems like a red herring. The omp issue is there regardless of optimization.

jonbob:
I checked Chris Golaz's runs on chrysalis and it looks like the WC v2 runs were not threaded. Some of the default layouts for chrysalis are threaded, but it looks like mostly with DATM

rljacob:
Does it depend on resolution?  SMS_P12x2.ne4_oQU240.WCYCL1850NS.chrysalis_intel.allactive-mach_mods never showed any random diffs and that forces everything to have 2 threads.

xylar:
@rljacob, it depends on compiling in debug mode, having threads and being on Chrysalis.  The test you listed isn't in debug mode.

rljacob:
@jonbob thought debug wasn't required.  Which is it?
Are you sure this doesn't happen on cori?  That's the only other place v2 sims with the bug were run.

rljacob:
For the record, PR #4521 was merged at v2.0.0-rc4-505, 3 commits before the v2.0.0 tag.  A lot of v2 sims were actually done with earlier tags.

jonbob:
We didn't see it in a non-debug test. But the omp issue is still there. It could be whatever race condition sets it off has yet to appear with optimization on?

jonbob:
But we're also trying to understand a G-case run on chrysalis with threading that looks way off. Testing to see if it's due to this bug

golaz:
Given the potential implications, we should investigate if any of the v2 Water Cycle simulations are impacted. We are running two configurations (LR and NARRM) on two machines (chrysalis and cori-knl) so we have four different cases.

The simulations are all listed on the big table:

https://acme-climate.atlassian.net/wiki/spaces/ED/pages/2766340117/V2+Simulation+Planning

Currently, I'm running a LE LR on cori-knl, and @singhbalwinder is running NARRM on cori-knl.

@rljacob : can you find someone to check these four configurations and report whether any of them is impacted by the bug? Thanks.

singhbalwinder:
@jonbob contacted me this morning and he thinks our simulations on Cori-KNL may be impacted by this bug as they are all using threading.

jonbob:
@golaz - I did a quick look in PACE and all of the runs you've been doing on chrysalis that I checked had threading off, but it was not an exhaustive check. I looked at the codebase and the default layouts for chrysalis do not have threading on for the v2 WC configuration, so I think you would have to specify a different layout to run into this bug on any of those runs. I looked at some of the run_e3sm scripts and didn't see it set there, but again I didn't do an exhaustive search.

It's a different story on cori-knl. The default NARRM layouts appear to be threaded. I'll search in PACE and see what I can find.

jonbob:
@golaz - the LR historical runs on cori-knl do not appear to be threaded. But the NARRM ones that @singhbalwinder has been running are. Once cori is back from maintenance, I'll see if I can run some tests to determine if the results have been impacted

jonbob:
@golaz -- the NARRM run on chrysalis is not threaded. You can look at the env_mach_pes.xml file on PACE to see the number of threads for each component, if you want to know the secrets...

rljacob:
Or run ./pelayout in the case dir.

@golaz, a lot of the v2 runs actually used a hash prior to the v2 tag so never had this bug.

jonbob:
Good point, @rljacob. I checked one of the NARRM jobs @singhbalwinder ran on cori and it was using v2.0.0-326-g417b03f684, and it's unfortunately from after the bug was introduced. So of the four cases @golaz mentioned, I think this one, NARRM on cori-knl, is the only one possibly impacted. I'll run some tests when cori is back and see if I can find any DIFFs with that configuration, but having it pass short reproducibility tests doesn't do much to show it hasn't had an impact in a long production run.

singhbalwinder:
Hi @jonbob : Just wondering if your tests revealed anything about the impact of this bug on Cori-KNL.

jonbob:
@singhbalwinder - not yet. I ran into problems on cori, but pushing on it again today

jonbob:
@singhbalwinder - I'm running a series of short (two-day) tests using the same hash, configuration, and pe-layout as your run. So far I've made three of these runs and they've all been BFB. I don't know how many it would take to feel comfortable that this bug didn't impact your production run -- that's the trouble with OpenMP issues, very difficult to prove they have not made problems.... I'll run another three tests and see what I get

rljacob:
On the cases/machine where you were able to see the bug, did it produce physically wrong answers or just roundoff diffs?

ambrad:
I have two suggestions.

1. With the production build options, run multiple times with 1 rank/node, 128 threads/rank, and compare output. This should maximally exercise this bug. If you don't see any diffs, go to step 2.
2. Still with the production build options, declare surfaceTemperatureFluxWithoutRunoff as `volatile`, like this:
```F90
real (kind=RKIND), volatile :: surfaceTemperatureFluxWithoutRunoff
```
Now you should see diffs. If you don't, then it's likely the testing isn't sensitive to the bug.

These suggestions are based on the following demonstration:
```F90
program main
  integer, parameter :: n = 1000
  integer :: i
  real :: r(n), s
  !$omp parallel do                                                                                                                                                                           
  do i = 1,n
     s = cos(real(i)/n) ! race condition
     r(i) = s
  end do
  !$omp end parallel do                                                                                                                                                                       
  do i = 2,n
     r(1) = r(1) + r(i)
  end do
  print '(es23.16)', r(1)
end program main
```
Now, I build with -O3 and -g:
```(base) [ac.ambradl@chr-0493 tmp]$ ifort -qopenmp -O3 fpp.F90; for i in $(seq 10); do OMP_NUM_THREADS=32 ./a.out; done
ifort -qopenmp -O3 fpp.F90; for i in $(seq 10); do OMP_NUM_THREADS=32 ./a.out; done
 8.4124102783203125E+02
 ...
 8.4124102783203125E+02    # all the same
(base) [ac.ambradl@chr-0493 tmp]$ ifort -qopenmp -g fpp.F90; for i in $(seq 10); do OMP_NUM_THREADS=32 ./a.out; done
ifort -qopenmp -g fpp.F90; for i in $(seq 10); do OMP_NUM_THREADS=32 ./a.out; done
 8.4124102783203125E+02
 ...
 8.4131365966796875E+02   # diffs
```
Now I declare `s` as `volatile` and rerun with -O3:
```F90
 real, volatile :: s
```
I get diffs just like in the -g case. Finally, I remove `volatile` and write correctly
```F90
!$omp parallel do private(s)
```
and get the correct answer 8.4124102783203125E+02 always.

Here's what's happening. Given OpenMP's memory model, the behavior w.r.t. surfaceTemperatureFluxWithoutRunoff is undefined. With -O3, ifort almost certainly chooses the fastest interpretation of the undefined behavior and makes surfaceTemperatureFluxWithoutRunoff effectively private in the emitted code. With -g, ifort interprets the code literally, and there is a single shared instance. `volatile` is one means of making the behavior defined regardless of compiler flag; the resulting behavior is that there is one shared variable, a bug given the intent.

jonbob:
Thanks for all the ideas, @ambrad. My guess is that you're exactly right, that ifort is interpreting the undefined behavior correctly. None of my optimized runs showed any difference in results, and I'm testing with the volatile setting right now. Assuming that behaves as expected (and you showed clearly), I think there's no reason for @singhbalwinder to question his results. I very much appreciate all your input, which does explain the behavior we've seen completely.

jonbob:
The test with surfaceTemperatureFluxWithoutRunoff declared volatile did show differences in the solution, as we would expect from @ambrad's analysis. I think at this point we can assume @singhbalwinder runs are fine, due to the optimization of this variable.

singhbalwinder:
Thanks @jonbob and @ambrad for confirming that my simulations should be fine and I don't need to re-run those.

Thanks @ambrad for the insightful suggestions!

PR #4736, opened 2022-01-13
Fix missing dependency in one case of gen_f90_targets

### LINKED ISSUE #4341: modifying files in mpas-framework is not recognized by ./case.build
I tried to add a timer to a routine in 

     mpas-framework/src/framework/mpas_stream_manager.F

./case.build did not recognize the change and the timer was not added. This was an F case using MPASSI for ICE. I then tried

     ./case.build --clean ice
     ./case.build

and it still did not add the timer. I had to

     ./case.build --clean-all
     ./case.build

to get the timer added.

jgfouca:
This bug was causing .F changes to not be reflected in rebuilds.

Fixes #4341

[BFB]


jgfouca:
@worleyph , apologies for the slow response.

mark-petersen (approved):
This looks correct. I also ran a before/after test on cori and both passed
```
./create_test SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.cori-haswell_gnu -q debug --walltime 00:30:00
```

Issue #4709, opened 2021-12-16
MEMLEAK with SMS_Ld20.f45_f45.IELMFATES.xxx.elm-fates_eca on KNL and perlmutter


ndkeen:
There are 2 tests that are failing with MEMLEAK on the 2 KNL machines (cori/theta):

```
SMS_Ld20.f45_f45.IELMFATES.cori-knl_intel.elm-fates_eca
SMS_Ld20.f45_f45.IELMFATES.cori-knl_intel.elm-fates_rd
```

I tried `SMS_Lm3.f45_f45.IELMFATES.cori-knl_intel.elm-fates_eca` and see the memory continue to grow.
I also tried GNU compiler (which is v8 on cori) and see the same issue.
Building DEBUG also shows the same type of memory growth (for both Intel/GNU).

The test passes using cori-haswell (with Intel and GNU) and presumably on other machines as I'm not seeing this test report a failure.

Although, I do see the same fail with those 2 tests on Perlmutter which is using GNU v9 on AMD's.

ndkeen:
I tried `azamat/elm/fix-datatype-memleaks` and I do see a decrease in memory use, but it's still clearly growing -- will include a plot when I get a chance to make one.

I do see that `SMS_Ld20.f45_f45.IELMFATES.cori-haswell_intel.elm-fates_rd` jsut failed with MEMLEAK on both cori-haswell and anvil -- so maybe it really is a memory leak on any machine, but the amount measured is near the threshold for causing the test to fail.

rgknox:
Looking into this too

rgknox:
Is it possible this is a red herring, false positive?  As a typical simulation progresses, the FATES model will continue to grow in memory as its representation of biodiversity increases, by design.  FATES is constantly spawning new objects (plant cohorts and area patches) and cleaning up old.  At the beginning of a cold-started simulation this is particularly unbalanced, there is more spawning of new objects  than there are cleaning up of old.  FATES will eventually hit a memory usage equilibrium of sorts, but that will not be achieved in these tests.

ndkeen:
Every time I run the problem longer, the memory continues to increase. However, it's true that while memory increase may look odd, it's not a terrible amount and the rate of increase slows.

Here is one attempt at running several years (only completed about 5 years).
![memhighwater](https://user-images.githubusercontent.com/5684727/148838362-ec56017a-d9d5-478d-916d-7248f00da96d.png)

[...]

PR #4739, opened 2022-01-15
Set memleak tolerance at 15% for FATES tests on Cori

amametjanov:
Set memleak tolerance at 15% for FATES tests on Cori with testmods.

Addresses E3SM-Project/E3SM#4709

[BFB]

Testing: https://my.cdash.org/viewTest.php?buildid=2115850


ndkeen:
I will merge this to next

rgknox (approved):
Thanks for these updates, they look good to me. I do want to note, that it is still very possible for FATES to increase memory usage, by design, more than the thresholds set here.  Hopefully, these settings will help us catch unexpected memory usage differences against a base.

ndkeen:
Fixes https://github.com/E3SM-Project/E3SM/issues/4709

Issue #4741, opened 2022-01-18
Perlmutter: Need to adjust modules after maintenance


ndkeen:
NERSC made changes again to default env for Perlmutter.
We need to remove the line loading cpe-cuda module as this will override the GNU version.
So that basically, the build will continue using GNU v10, when we actually want GNU v9.

` ndk/machinefiles/PM-module-fix`

[bfb]

PR #4742, opened 2022-01-18
For Perlmutter, remove a line to load environment module that is no longer needed

ndkeen:
NERSC made changes again to default env for Perlmutter.
We need to remove the line loading cpe-cuda module as this will override the GNU version.
So that basically, the build will continue using GNU v10, when we actually want GNU v9.
Only affects gpu builds (ie gnugpu, not gnu).

[bfb]

Fixes https://github.com/E3SM-Project/E3SM/issues/4741


ndkeen:
merged to next

PR #4747, opened 2022-01-21
Fixes a typo in compy's intel cmake file

### LINKED ISSUE #4746: cmake bug for compy in debug mode
There seems to be a bug in [intel_compy.cmake](https://github.com/E3SM-Project/E3SM/blob/master/cime_config/machines/cmake_macros/intel_compy.cmake) when `DEBUG=TRUE` that appends `-checkd all` instead of `-check all` and causes the build to fail

singhbalwinder:
This typo breaks model compilation in the debug mode on Compy

Fixes #4746

[BFB] - Bit-For-Bit


jgfouca (approved):

jgfouca:
Ty @singhbalwinder ! Skipping integration.

PR #4749, opened 2022-01-21
Some minor CMake changes in Homme

bartgol:
These changes only affect Homme standalone (since EAM does not use CMake). They allow a cmake host project to retain control of compiler flags, rather than having Homme decide.


jgfouca:
Merged to next.

bartgol:
On weaver, for Homme standalone, cmake generates the same flags in this branch as it does for master (I inspected the flags.make file for a theta target).

jgfouca:
sandiatoss3 nightlies had some problems but they were not related to this PR. I tested `homme_integration` by hand there and it worked fine. Merging.

PR #4750, opened 2022-01-24
Initialize Coriolis param in its own subroutine

xylar:
In situations where we do not expand the Earth sphere radius, the Coriolist parameter still needs to get initialized to a realistic value.  This merge creates a new subroutine for initializing the Coriolis parameter that is called regardless of whether the Earth sphere is being expanded or not.

This only affects MPAS-Ocean in init mode and is bit-for-bit in E3SM.

This bug was reported in:
https://github.com/MPAS-Dev/compass/issues/304

### DIFF
--- a/components/mpas-ocean/src/mode_init/mpas_ocn_init_mode.F
+++ b/components/mpas-ocean/src/mode_init/mpas_ocn_init_mode.F
@@ -173,6 +173,9 @@ module ocn_init_mode
       ! Expand sphere if it needs to be expanded
       call ocn_init_expand_sphere(domain, domain % streamManager, a, ierr)
 
+      ! Set the Coriolis parameter is a realist value is requested
+      call ocn_init_realistic_coriolis_parameter(domain, ierr)
+
    end function ocn_init_mode_init!}}}
 
 !***********************************************************************
--- a/components/mpas-ocean/src/mode_init/mpas_ocn_init_spherical_utils.F
+++ b/components/mpas-ocean/src/mode_init/mpas_ocn_init_spherical_utils.F
@@ -31,7 +31,8 @@ module ocn_init_spherical_utils
    implicit none
    private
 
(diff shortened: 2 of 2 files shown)


xylar:
## Testing

I will do more testing on bigger meshes but for now, I ran init mode (via compass) with the QU240 mesh and it produces a realistic Coriolis parameter:
qu240_coriolis

Without this fix, the Coriolist parameter was all zeros when `config_expand_sphere = .false.`.

ytakano3:
@xylar I used your mpas-o code and ran compass with the regional refined mesh I am working on (i.e. Kuroshio12to60) made new initial conditions. Now the file included the Coriolis parameter (which means initialization is working) and dynamical adjustment process gave reasonable files for E3SM (see below for SST, now looks good). I am now preparing and testing in E3SM mode (running G-Case) to see how simulation evolves. I will update once this is complete. Thank you for your support.

kuroshio12to60_ini_temperature_0

xylar:
@mark-petersen, I think this is a really important fix but something else is going wrong in my testing (see above) that we need to track down.

mark-petersen:
See explanation in https://github.com/MPAS-Dev/compass/issues/308.

mark-petersen (approved):
Thanks for this bit of reorganization. I included this in my testing using compass/master, and it passes the nightly regression suite, gnu optimized and debug. Also works fine for WC14.

xylar:
Thanks, @mark-petersen!

@jonbob, I believe this is ready to integrate when there is a slot for it.

jonbob:
@mark-petersen - can you please review this and sign off when you're satisfied? Thanks

mark-petersen:
@jonbob I approved above after several init mode tests. Thanks.

jonbob:
Goops -- sorry @mark-petersen. I hadn't refreshed this page and it showed as not yet approved.

jonbob:

jonbob:
merged to master

xylar:
Thanks so much, @jonbob!

PR #4760, opened 2022-01-27
remove task spec for MMF test w/ ESMT

whannah1:
The MMFXX test with ESMT enabled was still using a specific task/thread count even though we had fixed the issue that had forced us to use that approach, so this PR remedies this.

[BFB] but requires new baseline for changed test name


brhillman (approved):

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #4761, opened 2022-01-27
HOMME: remove one occurrence of the CPP ifndef SCREAM

bartgol:
The guarded function (push_forcing_to_c) is used unguarded later on, so a link error would ensue. Since the C function is compiled regardless of the SCREAM macro, the simple solution is to remove this ifndef guard.

NOTE: this should NOT affect any code/build that does not define the SCREAM macro. Such codes/builds were already preprocessing the file as if the ifndef wasn't there at all.


jgfouca (approved):

jgfouca:
Skipping integration testing.

Issue #4765, opened 2022-01-28
Update PE layouts for GCP for better thruput of tests


ndkeen:
On GCP, most tests work great with default 1 node, but there are a handful of tests that do much better with more nodes. I made a branch and tested with several suites.

`ndk/machinefiles/gcp-pelayout-adjust-for-tests`

Of course, room for improvement, but this helps with testing.  One improvement if we generate more partition files for ICE -- as GCP uses a somewhat odd number of cores (30) per node, making do with partition files we currently have.

PR #4766, opened 2022-01-28
For GCP, adjust default layouts for some common tests

ndkeen:
Add PE layouts for common configs of tests on GCP.
Change default walltime for jobs on GCP.
Always turn off HDF5 file locking.

Fixes https://github.com/E3SM-Project/E3SM/issues/4765
[bfb]


ndkeen:
merged to next

PR #4769, opened 2022-01-30
Add mode specification to conservation check streams

mark-petersen:
Currently, the conservation check analysis member does not specify its MPAS mode. This was an oversight. Without including this, the conservation check stream is included in MPAS init mode. This can cause an error, and it is never needed for init mode anyway. All the other analysis members include these lines.

[BFB]

### DIFF
--- a/components/mpas-ocean/src/analysis_members/Registry_conservation_check.xml
+++ b/components/mpas-ocean/src/analysis_members/Registry_conservation_check.xml
@@ -200,6 +200,7 @@
 	</var_struct>
 	<streams>
 		<stream name="conservationCheckOutput" type="output"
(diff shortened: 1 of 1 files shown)


xylar (approved):
Approved by inspection.  This clearly should only be in `forward` mode (and `analysis` mode if we ever use that...)

xylar:
@mark-petersen, would you like me to run the compass `pr` test suite or anything like that?  I don't think it's necessary in this case -- it shouldn't be affected by this change.

mark-petersen:
No thanks. I ran some tests with the regression suite, and it all works.

jonbob:

jonbob:
merged to master

Issue #4770, opened 2022-01-31
Default testing layouts for Perlmutter


ndkeen:
Currently, default 1-node layout is to use 64 MPI's, but that's only appropriate for CPU's.
When running on gpu's, need 4 MPI's per node (4 GPUs).
Using negative value for ntasks will use number of nodes: `<ntasks_atm>-1</ntasks_atm>` means 1 node
and the number of MPI's will be based on default MPI tasks per node.

`ndk/machinefiles/PM-default-layouts`

PR #4771, opened 2022-01-31
Change default 1-node layouts for perlmutter to be CPU/GPU aware

ndkeen:
Make default 1-node layouts (for tests) use best number of MPI's based on if build is wanting to use GPU's or not.

Fixes https://github.com/E3SM-Project/E3SM/issues/4770

[bfb]


ndkeen:
merged to next

rljacob:
How does that change make it cpu/gpu aware?

PR #4773, opened 2022-02-02
For Perlmutter, change NODENAME_REGEX to be "perlmutter"

ndkeen:
Change the NODENAME_REGEX to be perlmutter to avoid conflicts with machines using "login".

This change was possible after https://github.com/ESMCI/cime/issues/4133
It does require that users have proper setting of env var NERSC_HOST.

login34% env | grep NERSC_HOST
NERSC_HOST=perlmutter

Fixes https://github.com/E3SM-Project/E3SM/issues/4774

[bfb]

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -57,7 +57,7 @@
   -->
   <machine MACH="perlmutter">
     <DESC>Perlmutter at NERSC.  Phase1 only: Each GPU node has single AMD EPYC 7713 64-Core (Milan) and 4 nvidia A100's.</DESC>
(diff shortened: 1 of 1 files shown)


ndkeen:
merged to next

ndkeen:
I'm actually having trouble now with CIME not recognizing the machine. I'm not sure how it was working when I tested this. In  machines.py I see it's actually looking for `$ENV` in NODENAME_REGEX.  But when I try to set

`<NODENAME_REGEX>$ENV{NERSC_HOST}</NODENAME_REGEX>`

where:

```
login31% env | grep NERSC_HOST
NERSC_HOST=perlmutter
```
it still fails.

PR #4778, opened 2022-02-04
Change MMF total water used in VT_diagnose

whannah1:
To avoid a complication with the upcoming P3 implementation in the MMF we need to modify the variance transport diagnostic routine (VT_diagnose) to use the prognostic total water variable (i.e. micro_field). Unfortunately, this change is not BFB because the some of diagnostic quantities (qv+qc+qi) is not BFB with the prognostic total water that they were derived from. This isn't a problem for CRM, but it does cause this specific change to be non-BFB.

### DIFF
--- a/components/eam/src/physics/crm/samxx/crm_variance_transport.cpp
+++ b/components/eam/src/physics/crm/samxx/crm_variance_transport.cpp
@@ -99,9 +99,7 @@ void VT_filter(int filter_wn_max, real4d &f_in, real4d &f_out) {
 
 void VT_diagnose() {
   auto &t            = :: t;
-  auto &qv           = :: qv;
-  auto &qcl          = :: qcl;
-  auto &qci          = :: qci;
+  auto &micro_field  = :: micro_field;
   auto &factor_xy    = :: factor_xy;
   auto &t_vt_pert    = :: t_vt_pert;
   auto &q_vt_pert    = :: q_vt_pert;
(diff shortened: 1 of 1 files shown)


components/eam/src/physics/crm/samxx/crm_variance_transport.cpp
```diff
@@ -113,6 +111,8 @@ void VT_diagnose() {
   real2d t_mean("t_mean", nzm, ncrms);
   real2d q_mean("q_mean", nzm, ncrms);
 
+  int idx_qt = index_water_vapor;
```
brhillman, line 114: Is this correct? `idx_qt` should be index to total water, but it's being set to something called `index_water_vapor`.

whannah1: Yea, this whole thing is super confusing. index_water_vapor is used to point to the index of the prognostic variable that contains the vapor, even if it's total water. I did it this way to be consistent with other parts of the code, which are already inconsistent and confusing. What's more confusing is that for P3 we have a "idx_qt" to match the form of the other tracer indices like "idx_qc". However, even though total water contains the cloud liquid water, we still have to advect a separate "diagnostic" cloud liquid water tracer. So now I'm thinking that the code would be much more sensible if we just advected vapor and cloud water separately, but we'd still need to use total water for the 1-mom option. 

In this case I decided to mimic what is done in the mean-state acceleration routine. It's "correct" here because index_water_vapor=0, which is the micro_field index of the main prognostic total water tracer.

brhillman (approved):

brhillman:
Merged to next.

brhillman:
Merged to master.

rljacob:
Can clarify if this was non-BFB just for MMF cases?

whannah1:
@rljacob yes. This only affects the CRM code, so it's not possible for this to affect the non-MMF configuration.

PR #4779, opened 2022-02-07
Fix MPASSI BGC restarting for branch runs

jonbob:
Branch runs currently are not BFB when mpassi BGC is enabled, due to a missing setting in the seaice driver. This PR fixes that issue.


njeffery (approved):
This looks good. Tested in short GCASE simulations for restartability and ran initialized as 'branch' and 'initial' with BGC active.

jonbob:

jonbob:
merged to master

PR #4780, opened 2022-02-07
Correct NODENAME_REGEX for Perlmutter

ndkeen:
On Perlmutter, CIME is using the NERSC_HOST env variable (set to "perlmutter") to auto-detect machine.
My previous PR did not set this correctly.
Should be `<NODENAME_REGEX>$ENV{NERSC_HOST}:perlmutter</NODENAME_REGEX>`

[bfb]


ndkeen:
merged to next

PR #4783, opened 2022-02-08
Enable northamericax4v1pg2_WC14to60E2r3 for AMIP

wlin7:
DOCN added for searching component pattern in compset for
the North America RRM. This is to enable the grid for F-case
simulations which uses DOCN. A 30-node pe-layout (M) and a
60-node pe-layout (L) are also added for running on chrysalis
with this grid.

[BFB]

### DIFF
--- a/cime_config/allactive/config_pesall.xml
+++ b/cime_config/allactive/config_pesall.xml
@@ -9701,6 +9701,64 @@
           <rootpe_cpl>0</rootpe_cpl>
         </rootpe>
       </pes>      
+      <pes compset=".*EAM.+ELM.+MPASSI.+DOCN.+MOSART.+" pesize="M">
+        <comment> fmod030c64x1 s=6.2 </comment>
+        <MAX_MPITASKS_PER_NODE>64</MAX_MPITASKS_PER_NODE>
+        <MAX_TASKS_PER_NODE>128</MAX_TASKS_PER_NODE>
+        <ntasks>
+          <ntasks_atm>1920</ntasks_atm>
+          <ntasks_lnd>1920</ntasks_lnd>
+          <ntasks_rof>1920</ntasks_rof>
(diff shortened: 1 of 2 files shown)


wlin7:
The two pe-layout are more of place holder, using un-staggered PEs among the components.
10-day tests shows "M" gives 6.24 SYPD, and "L" gives 11.6 SYPD.

@amametjanov , should we do some adjustments to optimize the pe-layouts. (# of nodes can also be adjusted)
@tangq , are you planning to run the AMIP on chrysalis?

golaz:
Thanks for working on this @wlin7. Yes, @tangq is planning to run on chrysalis initially.

tangq (approved):
The changes look great. Thanks, @wlin7 !

tangq:
Yes, the AMIP run (at least the first member) will be done on chrysalis.

amametjanov (approved):
Looks like 10-day runs of `--compset F20TR --res northamericax4v1pg2_WC14to60E2r3` are getting:
- 30-nodes: 6.420 sypd (92% in EAM)
- 60-nodes: 12.208 sypd (90% in EAM)

I think stacked PE-layouts are okay. Node-count choices are good too: 3 switches for "L" runs.

wlin7:
Thanks for testing, Az. The SYPD # you got is higher that my first tests, which are nice.

wlin7:
Merged to master.

P.S. @tangq , if running on machines other than chrysalis, we would need new pe-layouts. Default ones likely too small.

tangq:
I will start the NARRM AMIP run on chrysalis with this updated master. Thanks @wlin7 !

tangq:
Hi @wlin7 , my NARRM AMIP test is also successful. Running the production run now. The AMIP run script I am using can specify customized PE layouts. We can use that for other machines if needed.

PR #4785, opened 2022-02-09
Turn off Redi diagonal terms for shallow columns 

### LINKED ISSUE #4784: Unphysical ocean temperatures due to Redi diagonal terms
An unphysical warming to 60C on the Indonesian coast occurs in the WC14 MPAS-Ocean stand-alone spin-up. This does not occur in E3SM simulations, but it may be covered up by surface fluxes, or cause unrealistic surface fluxes in shallow ocean columns.

See full description in https://github.com/MPAS-Dev/compass/issues/308

mark-petersen:
The MPAS-Ocean Redi diagonal terms are not guaranteed to produce bounded tracer fields, and in practice produce growing temperatures in a few ocean columns with fewer than 5 vertical cells in particular simulations. Redi is meant for isopycnal mixing in the deep ocean, so not applying Redi diagonal terms in very shallow regions is an acceptable solution. The problematic behavior is described in https://github.com/MPAS-Dev/compass/issues/308. 

Here we add the flag `config_Redi_min_layers_diag_terms`. Redi diagonal terms (2 and 3) are turned off from layer 1 through `config_Redi_min_layers_diag_terms-1`, and on from `config_Redi_min_layers_diag_terms` through `nVertLevels`. 

If `config_Redi_min_layers_diag_terms` is set to 1, Redi will be applied as before this PR, and simulations should compare bit-for-bit or machine precision to previous simulations.

We are adding the default `config_Redi_min_layers_diag_terms=6` (no Redi applied in columns with 5 or fewer vertical cells). I found in sensitivity tests that values of 5 and 6 produce no non-monotonic warming in the problematic case of WC14 ocean-only spin-up with no surface forcing and rough bathymetry (the most challenging case, see https://github.com/MPAS-Dev/compass/issues/308) but a value of 4 produces increasing temperatures.

[NML]
[non-BFB]
Fixes #4784

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_tracer_hmix_redi.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_tracer_hmix_redi.F
@@ -130,7 +130,7 @@ contains
 
       integer :: iCell, iEdge, cell1, cell2, iCellSelf
       integer :: i, k, iTr, nTracers, nCells, nEdges, nCellsP1
-      logical :: local_Redi_use_term2, local_Redi_use_term3
+      logical :: use_Redi_diag_terms
       integer, pointer :: nVertLevels
       integer, dimension(:), pointer :: nCellsArray, nEdgesArray
 
@@ -203,15 +203,10 @@ contains
       !$omp         kappaRediEdgeVal, iTr, tracer_turb_flux, flux, flux_term2, flux_term3, &
       !$omp         dTracerDx)
       do iCell = 1, nCells
-         if (maxLevelCell(iCell)>config_Redi_min_layers_term2) then
-            local_Redi_use_term2 = .true.
+         if (maxLevelCell(iCell)>config_Redi_min_layers_diag_terms) then
+            use_Redi_diag_terms = .true.
          else
-            local_Redi_use_term2 = .false.
-         endif
-         if (maxLevelCell(iCell)>config_Redi_min_layers_term3) then
-            local_Redi_use_term3 = .true.
-         else
-            local_Redi_use_term3 = .false.
+            use_Redi_diag_terms = .false.
          endif
 
          redi_term1(:, :, iCell) = 0.0_RKIND
@@ -270,7 +265,7 @@ contains
                                (tracers(iTr, minLevelEdgeBot(iEdge) - 1, cell2) - tracers(iTr, minLevelEdgeBot(iEdge), cell2))/(zMid(minLevelEdgeBot(iEdge) - 1, cell2) - zMid(minLevelEdgeBot(iEdge), cell2))
                endif
 
-if (config_Redi_use_term2 .and. local_Redi_use_term2) then ! mrp testing
+if (use_Redi_diag_terms) then ! mrp testing
                redi_term2(iTr, k, iCell) = redi_term2(iTr, k, iCell) - edgeSignOnCell(i, iCell)*flux_term2*invAreaCell
                redi_term2_edge(iTr, k, iEdge) = -flux_term2
 endif
@@ -303,7 +298,7 @@ endif
                                 /(zMid(k, cell1) - zMid(k + 1, cell1)) &
                                 + slopeTriadDown(k, 2, iEdge)*(tracers(iTr, k, cell2) - tracers(iTr, k + 1, cell2)) &
                                 /(zMid(k, cell2) - zMid(k + 1, cell2)))
-if (config_Redi_use_term2 .and. local_Redi_use_term2) then ! mrp testing
+if (use_Redi_diag_terms) then ! mrp testing
                   redi_term2(iTr, k, iCell) = redi_term2(iTr, k, iCell) - edgeSignOnCell(i, iCell)*flux_term2* &
                                 invAreaCell
                   redi_term2_edge(iTr, k, iEdge) = -flux_term2
@@ -345,7 +340,7 @@ endif
                                layerThicknessEdge(k, iEdge)*0.25_RKIND*kappaRediEdgeVal*slopeTriadDown(k, 2, iEdge)* &
                                (tracers(iTr, k, cell2) - tracers(iTr, k + 1, cell2))/(zMid(k, cell2) - zMid(k + 1, cell2))
                endif
-if (config_Redi_use_term2 .and. local_Redi_use_term2) then ! mrp testing
+if (use_Redi_diag_terms) then ! mrp testing
                redi_term2(iTr, k, iCell) = redi_term2(iTr, k, iCell) - edgeSignOnCell(i, iCell)*flux_term2 &
                                 *invAreaCell
                redi_term2_edge(iTr, k, iEdge) = -flux_term2
@@ -366,7 +361,7 @@ endif
          end do
          redi_term3_topOfCell(:, maxLevelCell(iCell) + 1, iCell) = 0.0_RKIND
 
-if (config_Redi_use_term3 .and. local_Redi_use_term3) then ! mrp testing
+if ( use_Redi_diag_terms) then ! mrp testing
          do k = minLevelCell(iCell), maxLevelCell(iCell)
             do iTr = 1, nTracers
                ! Add tendency for Term 3: d/dz ( h S grad phi) = ( S grad phi) fluxes
@@ -431,22 +426,17 @@ endif
       !$omp parallel
       !$omp do schedule(runtime) private(invAreaCell, k, iTr, i, iEdge)
       do iCell = 1, nCells
-         if (maxLevelCell(iCell)>config_Redi_min_layers_term2) then
-            local_Redi_use_term2 = .true.
+         if (maxLevelCell(iCell)>config_Redi_min_layers_diag_terms) then
+            use_Redi_diag_terms = .true.
          else
-            local_Redi_use_term2 = .false.
-         endif
-         if (maxLevelCell(iCell)>config_Redi_min_layers_term3) then
-            local_Redi_use_term3 = .true.
-         else
-            local_Redi_use_term3 = .false.
+            use_Redi_diag_terms = .false.
          endif
 
          invAreaCell = 1.0_RKIND/areaCell(iCell)
          do k = minLevelCell(iCell), maxLevelCell(iCell)
             do iTr = 1, ntracers
                tend(iTr, k, iCell) = tend(iTr, k, iCell) + redi_term1(iTr, k, iCell)
-if (config_Redi_use_term3 .and. local_Redi_use_term3) then ! mrp testing
+if ( use_Redi_diag_terms) then ! mrp testing
                tend(iTr, k, iCell) = tend(iTr, k, iCell) + rediKappaCell(iCell)*2.0_RKIND* &
                                      (RediKappaSfcTaper(k, iCell)*RediKappaScaling(k, iCell)* &
                                       redi_term3_topOfCell(iTr, k, iCell) - RediKappaSfcTaper(k + 1, iCell) &
@@ -455,7 +445,7 @@ endif
             end do
          end do
 
-if (config_Redi_use_term2 .and. local_Redi_use_term2) then ! mrp testing
+if (use_Redi_diag_terms) then ! mrp testing
          do i = 1, nEdgesOnCell(iCell)
             if (cellsOnCell(i, iCell) .eq. nCellsP1) cycle
             iEdge = edgesOnCell(i, iCell)
--- a/components/mpas-ocean/src/shared/mpas_ocn_tracer_hmix_redi.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_tracer_hmix_redi.F
@@ -240,6 +240,7 @@ contains
                redi_term1(iTr, k, iCell) = redi_term1(iTr, k, iCell) - (1.0_RKIND + term1TaperFactor* &
                                            (kappaRediEdgeVal - 1.0_RKIND))*edgeSignOnCell(i, iCell)*flux*invAreaCell
 
+               ! Term 2: div( h S dphi/dz)
                flux_term2 = coef*kappaRediEdgeVal*RediKappa(iEdge)*layerThicknessEdge(k, iEdge)* &
                             0.25_RKIND* &
                             (slopeTriadDown(k, 1, iEdge)*(tracers(iTr, k, cell1) - tracers(iTr, k + 1, cell1)) &
@@ -257,8 +258,10 @@ contains
                                (tracers(iTr, minLevelEdgeBot(iEdge) - 1, cell2) - tracers(iTr, minLevelEdgeBot(iEdge), cell2))/(zMid(minLevelEdgeBot(iEdge) - 1, cell2) - zMid(minLevelEdgeBot(iEdge), cell2))
                endif
 
+if (config_Redi_use_term2) then ! mrp testing
                redi_term2(iTr, k, iCell) = redi_term2(iTr, k, iCell) - edgeSignOnCell(i, iCell)*flux_term2*invAreaCell
                redi_term2_edge(iTr, k, iEdge) = -flux_term2
+endif
             end do
 
             do k = minLevelEdgeBot(iEdge) + 1, maxLevelEdgeTop(iEdge) - 1
@@ -288,10 +291,11 @@ contains
                                 /(zMid(k, cell1) - zMid(k + 1, cell1)) &
                                 + slopeTriadDown(k, 2, iEdge)*(tracers(iTr, k, cell2) - tracers(iTr, k + 1, cell2)) &
                                 /(zMid(k, cell2) - zMid(k + 1, cell2)))
+if (config_Redi_use_term2) then ! mrp testing
                   redi_term2(iTr, k, iCell) = redi_term2(iTr, k, iCell) - edgeSignOnCell(i, iCell)*flux_term2* &
                                 invAreaCell
                   redi_term2_edge(iTr, k, iEdge) = -flux_term2
-
+endif
                   dTracerDx = (tracers(iTr, k, cell2) - tracers(iTr, k, cell1))*dvEdge(iEdge)*0.25_RKIND
                   fluxRediZTop(iTr, k) = fluxRediZTop(iTr, k) + &
                                          0.5_RKIND*(slopeTriadUp(k, iCellSelf, iEdge) + slopeTriadDown(k - 1, iCellSelf, iEdge)) &
@@ -329,10 +333,11 @@ contains
                                layerThicknessEdge(k, iEdge)*0.25_RKIND*kappaRediEdgeVal*slopeTriadDown(k, 2, iEdge)* &
                                (tracers(iTr, k, cell2) - tracers(iTr, k + 1, cell2))/(zMid(k, cell2) - zMid(k + 1, cell2))
                endif
+if (config_Redi_use_term2) then ! mrp testing
                redi_term2(iTr, k, iCell) = redi_term2(iTr, k, iCell) - edgeSignOnCell(i, iCell)*flux_term2 &
                                 *invAreaCell
                redi_term2_edge(iTr, k, iEdge) = -flux_term2
-
+endif
                dTracerDx = (tracers(iTr, k, cell2) - tracers(iTr, k, cell1))*dvEdge(iEdge)*0.25_RKIND
                fluxRediZTop(iTr, k) = fluxRediZTop(iTr, k) + &
                                       0.5_RKIND*(slopeTriadUp(k, iCellSelf, iEdge) + slopeTriadDown(k - 1, iCellSelf, iEdge)) &
@@ -359,7 +364,9 @@ contains
                             * invAreaCell - RediKappaSfcTaper(k + 1, iCell)*RediKappaScaling(k + 1, iCell) &
                             *fluxRediZTop(iTr, k + 1) * invAreaCell)
 
+if (config_Redi_use_term3) then ! mrp testing
                redi_term3(iTr, k, iCell) = redi_term3(iTr, k, iCell) + flux_term3
+endif
             end do
          end do
       end do ! iCell
@@ -416,13 +423,16 @@ contains
          do k = minLevelCell(iCell), maxLevelCell(iCell)
             do iTr = 1, ntracers
                tend(iTr, k, iCell) = tend(iTr, k, iCell) + redi_term1(iTr, k, iCell)
+if (config_Redi_use_term3) then ! mrp testing
                tend(iTr, k, iCell) = tend(iTr, k, iCell) + rediKappaCell(iCell)*2.0_RKIND* &
                                      (RediKappaSfcTaper(k, iCell)*RediKappaScaling(k, iCell)* &
                                       redi_term3_topOfCell(iTr, k, iCell) - RediKappaSfcTaper(k + 1, iCell) &
                                       *RediKappaScaling(k + 1, iCell)*redi_term3_topOfCell(iTr, k + 1, iCell))
+endif
             end do
          end do
 
+if (config_Redi_use_term2) then ! mrp testing
          do i = 1, nEdgesOnCell(iCell)
             if (cellsOnCell(i, iCell) .eq. nCellsP1) cycle
             iEdge = edgesOnCell(i, iCell)
@@ -433,6 +443,7 @@ contains
             end do
          end do
       end do ! iCell
+endif
       !$omp end do
       !$omp end parallel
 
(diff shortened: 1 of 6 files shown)


mark-petersen:
This works in MPAS-Ocean stand-alone. I still need to add the E3SM namelist script changes. 

@xylar could you rerun the case described in https://github.com/MPAS-Dev/compass/issues/308 with this PR and verify that the maximum temperature and salinity do not increase? The best test is with no surface restoring and no bathymetry smoothing, which is what is on the head of compass/master right now.

components/mpas-ocean/src/shared/mpas_ocn_tracer_hmix_redi.F
```diff
@@ -240,6 +247,9 @@ subroutine ocn_tracer_hmix_Redi_tend(meshPool, layerThickness, layerThicknessEdg
                redi_term1(iTr, k, iCell) = redi_term1(iTr, k, iCell) - (1.0_RKIND + term1TaperFactor* &
                                            (kappaRediEdgeVal - 1.0_RKIND))*edgeSignOnCell(i, iCell)*flux*invAreaCell
 
+               if (.not.use_Redi_diag_terms) cycle
```
mark-petersen, line 250: @philipwjones any performance comments on my method here would be appreciated. I use a local logical variable and cycle just before computations involving terms 2 and 3. That means I have an if statement above, but that seems better than keeping an additional nCell-sized mask in memory.

philipwjones: These loops are complex enough that you're probably not getting much optimization anyway, so this doesn't do much harm.  I was actually about to re-do this routine for GPU optimization before I got distracted by the vmix routines, so I'll see what I can do with it later.

components/mpas-ocean/src/shared/mpas_ocn_tracer_hmix_redi.F
```diff
@@ -416,22 +433,29 @@ subroutine ocn_tracer_hmix_Redi_tend(meshPool, layerThickness, layerThicknessEdg
...
-                                      redi_term3_topOfCell(iTr, k, iCell) - RediKappaSfcTaper(k + 1, iCell) &
-                                      *RediKappaScaling(k + 1, iCell)*redi_term3_topOfCell(iTr, k + 1, iCell))
             end do
          end do
 
-         do i = 1, nEdgesOnCell(iCell)
-            if (cellsOnCell(i, iCell) .eq. nCellsP1) cycle
-            iEdge = edgesOnCell(i, iCell)
-            do k = minLevelEdgeBot(iEdge), maxLevelEdgeTop(iEdge)
+         if (maxLevelCell(iCell) .ge. config_Redi_min_layers_diag_terms) then
```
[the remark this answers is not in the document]
philipwjones: Yes, breaking them up here with the conditional outside is better.  And the config variables are fine to use directly. I've been copying some config variable into module variables because (1) some of the config names are just really really long and (2) if we go back to a private config with retrievals later, it would be better to retrieve once and make it a private module variable.

philipwjones (approved):
These changes look ok to me, assuming this makes sense from the parameterization standpoint and I'm not qualified to judge that.

vanroekel:
@mark-petersen the changes look good to me and I think are very reasonable for the parameterization.  Prior to approving, I'd like to wait for some testing to be posted.

xylar:
I haven't had a chance to test yet but I'll work on this tomorrow.

xylar:
I think we almost certainly need to rerun typical E3SM simulation (probably low res) for 5-10 years with no other changes than this to see what the impacts are.  I'm happy to do compass testing but I'm probably not the right person to do that E3SM test.  Is there someone who would be willing to do it?

mark-petersen:
Agreed. I can start those tomorrow. I need to work on my Ocean Sciences poster today.

vanroekel:
Yes definitely agree.  If someone could add the namelist flags to E3SM I could run a test.

xylar:
I'm working on a test now.

@mark-petersen, as a reminder from our [name conventions](https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/2523172/Branch+Tag+and+Version+name+conventions#Branch%2CTag%2CandVersionnameconventions-Development(akaTopicorFeature)BranchNames):
> Use lower-case for everything.  Even your username.

Hopefully we can let it slide.

xylar (approved):
I don't feel qualified to review the code changes themselves.  I don't understand the Redi code very well.  I'm hoping @vanroekel has provided that oversight.

I am running the `dynamic_adjustment` of the WC14 mesh on `compass/master` on Anvil with Intel and Intel-MPI.  While it hasn't quite finished yet, there are absolutely no signs of the temperature maxima we saw before so I feel very confident in approving now.

**Update**: Maximum temperature at the end of spin-up is: 30.16, so just a tick lower than the maximum in the initial condition.  Looks perfect!

Results are at:
```
/lcrc/group/e3sm/ac.xylar/compass_1.0/anvil/test_20220210/wc_with_redi_fix/ocean/global_ocean/WC14/PHC/dynamic_adjustment
```

@mark-petersen, thank you very much for addressing this important issue so quickly.

xylar:
A few visual checks showing SST after dynamic adjustment.

indonesia
north_atlantic

No sign of temperature weirdness anywhere I looked.

The Gulf Stream isn't looking great (or rather looks beautiful but not very physical).  It looks like we do still need to run in E3SM and not just compass standalone ;-)

vanroekel:
Thanks @mark-petersen I will get a longer test started today to get it running over the weekend.

vanroekel:
I ran 30 years of a B-case with this code, mpas-analysis is here
https://compy-dtn.pnl.gov/vanr302/20220211.v2.amoc.rediMinLayer.ne30pg2_EC30to60E2r2.compy_1_30/
The corresponding baseline is here
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.abarthel/E3SMv2/20220126.v2.amoc.baseline.ne30pg2_EC30to60E2r2.anvil/mpas_analysis/ts_0001-0030_climo_0001-0030/

overall the solution looks very similar with these changes, which is probably not too surprising given the run is the LR mesh.  However there are some signs of improvements, e.g., salinity bias and AMOC.  I'll keep it going, but am going to approve the PR.

vanroekel (approved):
approved by visual inspection and testing.

milenaveneziani:
it's actually quite amazing how this impacts the AMOC time series..

vanroekel:
True @milenaveneziani it is surprising that we see ~0.5 Sv increase for just changing shallow columns.  I'm going to be taking a look at bottom redi fluxes to see if we might benefit from turning off redi near bathymetry everywhere.

Also, I meant to point out, but lcrc is on maintenance.  In the rediTest run for a bunch of the global bias plots (e.g. SSS) the minimum bias is strongly reduced with this change.  I wonder if we were seeing this issue at LR just not manifested as strongly.

xylar:
You may wish to re-run the analysis you're comparing with using the latest MPAS-Analysis with the bug fix.

vanroekel:
Thanks @xylar I didn't make that connection.  I will update and rerun

jonbob:
@mark-petersen and @vanroekel - do we want to wait to merge this until after the next test? Or go ahead and get it in?

vanroekel:
@jonbob and @mark-petersen if this is high priority I'm okay merging this in as post v2 changes, but if there isn't a time crunch I wouldn't mind waiting for a few more decades of what I'm running.

I could also then run the test @mark-petersen asked about.

jonbob:
Let's wait then, @vanroekel - I think the most critical thing is to see if it helps with the AMOC issue. And it makes sense to find the right value for config_Redi_min_layers_diag_terms first

jenniferholm:
@susburrows - would it be ok if I leave this to you to okay or not, as you have been involved in this more than me? Thanks.

xylar:
@susburrows, this seems like a bug with serious enough effects that it would be irresponsible in my view not to include it.  @vanroekel, what is your feeling there?

vanroekel:
@susburrows I think the answer here is complicated.  I know in the past BGC has branched from water cycle simulations, in that case you wouldn't want to take this on for 2.1, but if you are  starting clean you likely want to include this.  I'd still like to further evaluate fully coupled E3SM simulations before saying how impactful the bug is to the broader ocean climate.  There is a massive impact in MPAS-Ocean only simulations for sure.

vanroekel:
looks like the difference in AMOC between the runs was due to a change in axis limits and not the magnitude itself.  Here is  AMOC from the redi test (this PR; blue) and the baseline test (orange).  The magnitude is quite similar.  The x-axis is just index here but the timeseries is over the 30 years.

image

mark-petersen:
@vanroekel thanks for that plot. Given that this PR solves the original problem of increasing temperature, but does not affect tuning for MOC differences, I recommend that we proceed and merge this PR as-is with `config_Redi_min_layers_diag_terms=6`, rather than testing with different values. All three reviewers have approved, so it looks like we can go ahead.

susburrows:
Hi everyone, just to make sure we're all on the same page: the reason Rob is asking BGC for approval is that any and all non-BFB PRs currently require our approval until the v2.1 tag is created, which we don't have an exact timeline for right now.  The plan has been, in part, that we are deliberately postponing the creation of the 2.1 tag until their is a non-BFB change that needs to go onto master that we don't want in v2.1. 
Postponing v2.1 as long as possible allows more time for the new land features to be tested and decisions to be made about them (all of these are land features that were originally part of the WC group activies, but when they were not ready in time for the WC deadline, the management oversight for these features was transferred to BGC).  Postponing all non-BFB changes until after the v2.1 tag is created allows us to maintain reproducibility of the physical climate with the WC configuration, which is important because when we eventually write papers about these simulations, we need to be able to reference the WC papers for the physical climate, instead of having to write a new paper documenting a different physical climate in v2.1.

Given these considerations, I think there are essentially two options here:
1) In light of the fact that we currently do not have any firm plans to perform v2 simulations that include an active ocean component, we could decide that non-BFB changes to ocean are OK, and that we only have a need to maintain BFB reproducibility with WC for land and atmosphere at this time.
2) We could postpone merging this PR to master until after the v2.1 tag is created (which might be a few months).

If we went with (2), would it have a significant negative impact on other current activities?

Also, I do think I will need to discuss this with Ben and Jennifer so we can make a joint decision, I don't think I can make a call on this unilaterally.  Therefore it will take at least a few days for us to decide.

xylar:
> If we went with (2), would it have a significant negative impact on other current activities?

In my view, this would be a disappointing outcome.  I have been waiting several months to release our standalone ocean testing software, compass, and I do not feel I can make that release until this fix is available.  It has a pretty detrimental impact on our ability to perform standalone ocean simulations with the Redi feature turned on.

However, I realize E3SM considerations must come before those of other "ecosystem" tools.  So I will defer to the rest of you with my strong preference noted.

susburrows:
Cc-ing @bpbond on this

mark-petersen:
I think @susburrows option 1 is best because it we know from stand-alone that it prevents potential strange warming. But considering the needs of BGC, another option is to merge with`config_Redi_min_layers_diag_terms=1`  so it remains bfb (or machine precision) with WC. Then we change that flag for spin-up and cryo simulations.

xylar:
@mark-petersen, could you determine if results are bit-for-bit or not with `config_Redi_min_layers_diag_terms=1`?  I could imagine that machine precision isn't good enough.

mark-petersen:
Good news! If we set `config_Redi_min_layers_diag_terms=1`, the full COMPASS nightly suite is BFB with the master branch point, 44814ae3. If I set the flag to 6, all global ocean cases, which run with Redi on, are not bfb and differ by max 1e-5 after several time steps. So that all makes sense.

So from what I've read above, it would be acceptable to merge this PR, but we need to first decide between:
1. `config_Redi_min_layers_diag_terms=6`: non BFB, but reduces spurious behavior in shallow columns, at least as observed during spin-up.
2. `config_Redi_min_layers_diag_terms=1`: BFB with previous, may be needed if BGC wants BFB match with WC sims.

Here (1) is more convenient for cryo (and, I assume, future WC sims), so is preferred by us. But I will defer to the needs for BGC to be BFB with WC simulations underway.

xylar:
@mark-petersen. thank you very much for testing this!

bpbond:
Thanks for the tag @susburrows 

I definitely see the importance of this as noted by @xylar and others. Before BGC signs off, could we see some diagnostics on the larger climate impact, from as long a run as possible? E.g. SST differences, ocean heat uptake, top-of-atmosphere radiation balance...that would be really useful to make sure we fully understand the implications of merging this non-bfb PR. 

Thanks!

mark-petersen:
@vanroekel could you post links to MPAS-Analysis to the two 30-year runs you list [above](https://github.com/E3SM-Project/E3SM/pull/4785#issuecomment-1039858486)? I think we can help @susburrows and @bpbond understand differences. In the MOC plot the two runs look quite similar.

mark-petersen:
Rebased to include the latest changes on master. No alterations were made on the code for this PR.

mark-petersen:
Reran rebased code with sanity tests, passes
```
```
and compass nightly suite with gnu and gnu-debug on grizzly.

xylar:
@vanroekel, I know you're busy but could you post the analysis for the runs you've done?  And could we discuss if they need to be run longer to satisfy @bpbond, @susburrows, @jenniferholm and the BGC team?

The decision here is holding up a compass release.

vanroekel:
oops sorry for the delay and thanks for the ping @xylar .  Fell off my radar.  Here you go

Redi with shallow columns off
https://compy-dtn.pnl.gov/vanr302/20220211.v2.amoc.rediMinLayer.ne30pg2_EC30to60E2r2.compy_1_30/

Baseline comparison
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.abarthel/E3SMv2/20220126.v2.amoc.baseline.ne30pg2_EC30to60E2r2.anvil/mpas_analysis/ts_0001-0040_climo_0001-0040/

I'm sorry to say I don't have zppy run on the new one yet.  I can try do that today.

xylar:
@vanroekel, if you are able to run zppy, that would be very helpful I think.  Once we have those results, maybe we can coordinate with the BGC folks because I don't know how familiar you all are with the ocean analysis.

Also, is it going to be a problem that we're looking at a 30-year climatology in the one case and a 40-year in the other?  Not that I want to make extra work for anyone but I would rather that we have what we need rather than iterating again and that taking another week.

xylar:
@bpbond asked me privately what a "compass release" meant and I figured that would be something others were wondering, too.

[compass](https://github.com/MPAS-Dev/compass) is a python package and associated software environment that we use for regression testing, creating E3SM initial conditions, idealized test cases, and much more for MPAS-Ocean and MALI.  One important feature of `compass` is that it keeps a git submodule for `E3SM-Project/E3SM` and `MALI-Dev/E3SM` (the fork for MALI development) that we update frequently as MPAS changes happen in E3SM.  We are nearly ready for the v1.0.0 release of `compass` but this particular bug has a strong, detrimental effect on our ability to create MPAS-Ocean initial conditions for use in E3SM from `compass`.  That is the reason I'm pushing so hard on this.  I (and many others) have put over a year of time into a complete overhaul of `compass` and we are excited for this release.  Several planned changes (incorporation of Spack and task parallelism among others) are on hold until the release happens.

I hope that provides some context.

vanroekel:
@xylar the climatologies are actually the same, I just forgot to rename the html directory.

I'm working on e3sm-diags now, it's not working though for some reason.

xylar:
> @xylar the climatologies are actually the same, I just forgot to rename the html directory.

Ah, that's a relief!

> I'm working on e3sm-diags now, it's not working though for some reason.

Let me know if I can help or we can get help from @forsyth2 or @chengzhuzhang if need be.

vanroekel:
Okay I think I have it working now.  My cfg file was so out of date it was missing a bunch of options.  I'll ping @chengzhuzhang or @forsyth2 if the analysis fails.  

I'll also replace the analysis link to what is from zppy where the directory is appropriately named to avoid confusion

vanroekel:
@xylar @bpbond @susburrows I finally got e3sm-diags going, took a while to refresh myself as to how it worked.  It's still processing, but the first 10 years are done and will show up here as they complete

https://compy-dtn.pnl.gov/vanr302/E3SMv2/20220211.v2.amoc.rediMinLayer.ne30pg2_EC30to60E2r2.compy/20220211.v2.amoc.rediMinLayer.ne30pg2_EC30to60E2r2.compy/e3sm_diags/180x360_aave/

The corresponding analysis for the baseline is here

https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.abarthel/E3SMv2/20220126.v2.amoc.baseline.ne30pg2_EC30to60E2r2.anvil/e3sm_diags/180x360_aave/

bpbond:
Thanks for the `compass` explanation @xylar and thanks for the links @vanroekel .

Re the latter, I could use some orientation to the output. ( @susburrows ?) Naively, it looks like the plots are all model vs. observations...are there some that show model vs. model, i.e. the net effect of this PR relative to `master`?

vanroekel:
@bpbond yes they are all model vs. obs  The two runs are on different machines so didn't have the data side by side to figure out the model vs. model option.  I was hoping by intercomparing the links (baseline and this PR) would be enough.

susburrows:
Quickly looking through the e3sm_diags output, it looks like this impacts ENSO and QBO, but has minimal impact on global mean TOA net radiative flux or SST at decadal scales.  Is that consistent with what you expected, @vanroekel ?

vanroekel:
@susburrows no I wouldn't expect that, especially the QBO.  I've double checked the code and namelists and they are the same expect for the redi changes.  So I'm puzzled by those differences.

susburrows:
Hmmm... I assumed the ENSO changes could be explained by changes to ocean temperature, but was also a bit surprised to see impacts on the QBO.  I don't know enough about the QBO diagnostics to know if these changes are significant enough to be cause for concern.

susburrows:
I was looking at the following plots: 
image
image

vanroekel:
I wonder if perhaps small shifts in the ocean changed the enso response, which could potentially shift QBO as well.  I'm certainly not knowledgeable enough of QBO to be definite.

susburrows:
The ENSO changes seem to be related to an increase in positive surface temperature anomalies, e.g. as shown in these plots:
image
image

susburrows:
Seems like it might be good to have someone review this who is knowledgeable about these modes of variability.

vanroekel:
thanks for pointing to those plots.  I had only looked at the climo plots from mpas-analysis which didn't show much in the way of shifting temperatures.  Given that the change to ENSO makes sense, but I'm not confident in saying anything about QBO -- @jjbenedict are you able to chime in on this?

But for ENSO, it is possible that since this is a min column change for Redi it may be changing SST right near the coast and since this is NINO3 it becomes more evident.

I do agree though it would be very good to have confirmation from someone more versed in these modes.

milenaveneziani:
Do you know on what record length the ENSO diags are computed within `e3sm_diag`?

I looked at the MPAS-Analysis results, where Enso diags are computed over the full time series (years 1-40), and there is definitely a difference in there, but it's hard to say which one is better than the other. The Redifix run has more energy at the frequencies of 2-4 years, though, so I assume that's an improvement.

susburrows:
> Do you know on what record length the ENSO diags are computed within `e3sm_diag`?

The links Luke posted above have e3sm_diags output per decade of the simulation.  The plots I copied above are for years 31-40 (this is also indicated on the plot labels).

milenaveneziani:
I wonder if those ranges indicate the periods over which climatologies are computed. 
If the Enso and other interannual variability indexes are computed over the same range, it seems way too short.

vanroekel:
@milenaveneziani I wondered the same thing about the length.  but I looked in the mpas-a output which is year 1-40 and the nino3.4 time series is definitely different, so the change does seem robust.

jjbenedict:
The ENSO and QBO diagnostics should be run on much more than 10 yr of data.  For the QBO you'd need at least 20 yr for robust statistics, but more data is better. 

Comparing 50 yr of v2.LR.piControl and 40 yr of 'rediMinLayer', the QBO results are similar.
qbo_diags
qbo_diags_rediMinLayer

@vanroekel or @alicebarthel, can we run e3sm_diags on years 1-40 of the 'baseline' 20220126.v2.amoc.baseline.ne30pg2_EC30to60E2r2.anvil?... I didn't see that output here:

[https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.abarthel/E3SMv2/20220126.v2.amoc.baseline.ne30pg2_EC30to60E2r2.anvil/e3sm_diags/180x360_aave/](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.abarthel/E3SMv2/20220126.v2.amoc.baseline.ne30pg2_EC30to60E2r2.anvil/e3sm_diags/180x360_aave/)

bpbond (approved):
Hi all, Biogeochemistry is happy to sign off on this, given that we're not depending on an active ocean for the v2 simulations; but we recommend that Chris  or Yaga Richter or someone with more expertise in QBO, etc., be looped in to assess re the graphs above. Thanks!

xylar:
@golaz, could you take a look at the QBO statistics above and let us know if anything concerning jumps out at you?

Does anyone know Yaga Richter's GitHub username?  If so, please @ her so she can take a look as well.

xylar:
@mark-petersen, @vanroekel and @jonbob, what do you feel is the status of this PR?  I personally do not feel that QBO or ENSO statistics are an important assessment of the validity (or not) of this bug fix.  Given that it is a bug, I don't fully understand why we need to have an expert look at those statistic in particular.

What is the path to getting this PR merged?

rljacob:
This is essentially a v3 change since v3 watercycle will be the first sims to see this.  Master isn't quite ready for v3 changes.

xylar:
@rljacob, I don’t agree. This is a know, potentially pretty serious bug and it should not be in the cryosphere v2 runs, which have not been tagged or started quite yet to my knowledge. It is true that the water cycle would not see this fix until v3.

xylar:
@rljacob, cryosphere definitely can’t use v2.0 It’s worse than our best v1.0 simulation. We will need a v2.2 tag for sure and have been assuming that for some time now.

jonbob:
@xylar -- I agree that this is a bug fix and should go in, especially since the BGC leadership has accepted it. However, I also think it is important for someone to look at other system diagnostics like QBO, if only to make sure the fix doesn't require any re-tuning elsewhere.

xylar:
Thanks, @jonbob, that’s a very fair point.

vanroekel:
All,  here are diagnostics on equivalent intervals averaged years 1-40  First with this PR

https://compy-dtn.pnl.gov/vanr302/E3SMv2/20220211.v2.amoc.rediMinLayer.ne30pg2_EC30to60E2r2.compy/20220211.v2.amoc.rediMinLayer.ne30pg2_EC30to60E2r2.compy/e3sm_diags/180x360_aave/model_vs_obs_0001-0040/viewer/

Baseline

https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.abarthel/E3SMv2/20220126.v2.amoc.baseline.ne30pg2_EC30to60E2r2.anvil/e3sm_diags/180x360_aave/model_vs_obs_0001-0040/viewer/

As mentioned above by @jjbenedict the QBO diagnostics become much more consistent when averaged over longer intervals.  ENSO remains different, but given ENSO is sensitive to model perturbations in ensemble members it is not surprising this small change to the ocean state changes the Nino time series.  The rest of the diagnostics seem pretty consistent between the runs.

mark-petersen:
@rljacob could you please provide guidance on this PR? It has been approved by all reviewers, and we believe that we have done our due diligence in providing comparisons in the previous comments between the baseline V2.0 and this PR.

Our preference is to merge this in as-is, as a non-BFB bug fix. My understanding is that the non-BFB merge would require a V2.0-maintenance branch at this time.

rljacob:
Go ahead and start merging it.

golaz:
@rljacob : has the maint-2.0 branch been created?

rljacob:
Yes.  It will be more broadly announced after this is merged.

jonbob:
Testing of the merge shows non-BFB and NML DIFFS, but still passes ERS, PET and PEM checks

Merged to next

jonbob:
merged to master and expected DIFFs and NML DIFFs blessed

xylar:
🎉🎉🥳🥳

Thanks @jonbob and everyone else who made this possible!

bpbond:
🥂

PR #4758, opened 2022-01-27
Add pan-arctic atmosphere RRM grid

brhillman:
Add a new pan-arctic atmosphere RRM grid (arcticx4v1pg2). Configuration and input files provided to run with four different configurations:
```
arcticx4v1pg2_oARRM60to10
arcticx4v1pg2_r05_oARRM60to10
arcticx4v1pg2_r125_oARRM60to10
```
[BFB]

PR #4795, opened 2022-02-15
Update ocean and sea ice ICs for ARRM60to10 mesh

### LINKED ISSUE #4797: B-cases with `ne30` atm and `ARRM60to10` use 15 minute coupling timestep
By default, B-cases that use `ne30pg2_oARRM60to10` have `ATM_NCPL` and `LND_NCPL` = 96, resulting in a 15 minute coupling timestep, which errors out on job submission. This should be changed to 48, for a 30 minute atm/lnd timestep and coupling frequency.

darincomeau:
Updates the ocean and sea ice initial conditions for spun-up compsets using the `ARRM60to10` grid. This is needed to replace the existing sea ice initial condition, which has one snow layer, and is not compatible with v2. These new ocean and sea ice initial conditions are generated from restarts from a one month G-case using cold start initial conditions.

This also updates the ocean analysis mask file to include additional regions in addition to just `Atlantic`, which was the only one in the previous file it is replacing.

Fixes #4797

[BFB] for all configurations that do not use the `ARRM60to10` ocean / sea-ice mesh, or use this mesh with cold-start initial conditions (i.e. G-cases).
[non-BFB] for configurations that use the `ARRM60to10` mesh with spun-up ocean / sea-ice initial conditions (i.e. B-cases).

### DIFF
--- a/components/mpas-ocean/cime_config/buildnml
+++ b/components/mpas-ocean/cime_config/buildnml
@@ -189,12 +189,12 @@ def buildnml(case, caseroot, compname):
         decomp_date = '180723'
         decomp_prefix = 'mpas-o.graph.info.'
         restoring_file = 'sss.monthlyClimatology.PHC2_salx.2004_08_03.ARRM60to10.190129.nc'
-        analysis_mask_file = 'ARRM60to10_SingleRegionAtlanticWTransportTransects_masks.nc'
+        analysis_mask_file = 'ARRM60to10_mocBasinsAndTransects20210623.nc'
         ic_date = '180715'
         ic_prefix = 'ocean.ARRM60to10'
         if ocn_ic_mode == 'spunup':
-            ic_date = '200422'
-            ic_prefix = 'ocean.ARRM60to10.restart_grizzly'
+            ic_date = '210204'
+            ic_prefix = 'mpaso.oARRM60to10.rstFromG-cori'
 
     elif ocn_grid == 'oARRM60to6':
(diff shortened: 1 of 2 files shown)


darincomeau:
I ran a 5 day `WCYCL1950.ne30pg2_oARRM60to10` smoke test to confirm the correct ocean and sea ice initial conditions were used, as well as the new analysis mask file. With the new analysis mask file, there are no longer errors in the `ocn.log` file complaining about the number of regions in the mask file.

Note: To run this smoke test, I needed to change the default behavior with the following:
`./xmlchange -file env_run.xml -id ATM_NCPL -val 48`
`./xmlchange -file env_run.xml -id LND_NCPL -val 48`

The defaults for each were 96.

milenaveneziani:
Thanks @darincomeau. Do you want me to create a case and check the streams?
Visual inspection looks good to me.

darincomeau:
@milenaveneziani approving by visual inspection and my testing is fine. I'll open an issue on the `ATM_NCPL` setting for this configuration to keep track of it, but I want to make sure whatever fix there doesn't conflict with https://github.com/E3SM-Project/E3SM/pull/4758 .

milenaveneziani (approved):

darincomeau:
Issue created: https://github.com/E3SM-Project/E3SM/issues/4797

jonbob:

jonbob:
merged to master

PR #4803, opened 2022-02-18
Initializes uninitialized variables in ndrop.F90

### LINKED ISSUE #4782: Uninitialied ndropmix and nsource in ndrop.F90 caused a PIO fatal error
The variables "ndropmix" and "nsource" are not initialized in the ndrop.F90, so they may contain "NaN" that caused a PIO fatal error when they were dumped to cam history files. 

A possible solution is to add the following two lines code to initialize the variables before the overall_main_i_loop:

```
   ndropmix = 0._r8
   nsource = 0._r8

   ! overall_main_i_loop
   do i = 1, ncol
```

The error message is as follows. The compiler is gcc/9.3.0 with cray-mpich/8.1.11 on Miller. Thank @jayeshkrishna for pointing out the cause of the error. 
```
PIO: FATAL ERROR: Aborting... An error occured, Writing variables (number of variables = 78) to file (v2.LR.piControl.eam.h3.0001-01-01-00000.nc, ncid=125) using PIO_IOTYPE_PNETCDF iotype failed. Non blocking write for variable (NDROPMIX, varid=51) failed (Number of subarray requests/regions=1, Size of data local to this process = 777600). err=-60. Aborting since the error handler was set to PIO_INTERNAL_ERROR... (/autofs/nccs-svm1_home1/mxu/MyGit/e3sm_v2/externals/scorpio/src/clib/pio_darray_int.c: 391)
```

singhbalwinder:
Variables ndropmix and nsource were uninitialized in ndrop.F90. 
Effectively calculations loop from top_lev ( 100 Pa) to bottom 
level. The levels above 100 Pa as a result were not initialized, 
but the full arrays were sent to outfld, causing problem to PIO 
when NaN are present.

They are both initialized to zero now.

Fixes #4782
[BFB]

### DIFF
--- a/components/eam/src/physics/cam/ndrop.F90
+++ b/components/eam/src/physics/cam/ndrop.F90
@@ -517,6 +517,10 @@ subroutine dropmixnuc( &
       call physics_ptend_init(ptend, state%psetcols, 'ndrop')
    end if
 
+   !initialize variables to zero
+   ndropmix(:,:) = 0._r8
+   nsource(:,:) = 0._r8
+
    ! overall_main_i_loop
    do i = 1, ncol
 


hwangacme (approved):

wlin7:
Thanks for the fix, @singhbalwinder . Just to add a note that the effective calculations in the codes involving ndropmix and nsource are never affected, thus no impact on results. The calculations (and assignments) were previously only applied to levels from trop_cloud_top_pres (100 Pa) to surface. The values  for levels 1:top_lev-1 were therefore uninitialized but the full array were sent to outfld, causing the problem to PIO.

singhbalwinder:
Thanks Wuyin for the detailed explanation. Its nice that PIO caught this uninitialized behavior.

wlin7:
Merged to next.

wlin7:
Merged to master.

PR #4589, opened 2021-10-07
Corrects the calculation of solar zenith angle, which was off by a timestep

jonbob:
The solar zenith angle calculation used by the data models was off by one timestep. This issue was first reported by CESM  in [ESMCI Issue #3380](https://github.com/ESMCI/cime/issues/3380) and subsequently fixed in [ESCOMP PR #123](https://github.com/ESCOMP/CDEPS/pull/123). This PR implements that solution in E3SM.

Fixes #4575 

[non-BFB] for configurations with data models

PR #4805, opened 2022-02-22
Add CMPASO-JRA1p4 compset for testing

jonbob:
Adds a new compset with active ocean only but using JRA1p4 forcing for datm and drof.

### DIFF
--- a/components/mpas-ocean/cime_config/config_compsets.xml
+++ b/components/mpas-ocean/cime_config/config_compsets.xml
@@ -22,6 +22,12 @@
   </compset>
 
   <compset>
+    <alias>CMPASO-JRA1p4</alias>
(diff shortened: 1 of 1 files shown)


vanroekel (approved):
approved by visual inspection -- comparison with CMPASO-IAF and GMPAS-JRA1p4 show these changes are as expected.

jonbob:
on a hunch, I cherry-picked the commit from [PR #4589](https://github.com/E3SM-Project/E3SM/pull/4589) which corrects the calculation of the solar zenith angle for some data model applications -- and the exact restart test that previously failed now passes. @proteanplanet, could you please complete your review of that PR soon, since it is impacting some configurations?

jonbob:
@rljacob - is it OK to merge this PR since it only adds a compset alias, even if a typical test using it fails? It is documented as having a support level of "Experimental, under development"

rljacob:
Yes that's fine.

jonbob:
merged to next

jonbob:
merged to master

Issue #4811, opened 2022-02-25
Update module versions on PM (including gnu v10)


ndkeen:
On Feb24th, there were some software changes on PM, and we can update the versions of several modules.
The default compiler is now GNU v11, but I thought it best to only upgrade to v10 for now.

With `ndk/machinefiles/PM-update-gnuv10` I've already tested `e3sm_developer` as well as `e3sm_mmf_integration`.
There are 2 fails in developer that are also present on cori. There are fails in mmf suite, but only 1 is a new fail (from gnu v9) but is not actually using gpu.

Issue #3954, opened 2020-11-20
Compile issue with GCC 10.1.0


ndkeen:
Wanted to try newer GNU compiler, but hit an issue here:

```
cime/src/share/timing/perf_utils.F90

  282 |    call MPI_BCAST(vec,lsize,MPI_INTEGER,0,comm,ierr)
      |                  1
......
  314 |    call MPI_BCAST(vec,lsize,MPI_LOGICAL,0,comm,ierr)
      |                  2
Error: Type mismatch between actual argument at (1) and actual argument at (2) (INTEGER(4)/LOGICAL(4)).
```

This is on cori. I created simple tester:

```
module boilerplate
  implicit none
  private
#include <mpif.h>
  save
  
  interface shr_mpi_bcast ; module procedure &
       shr_mpi_bcastl0, &
       shr_mpi_bcasti0
  end interface
  
contains
  
  SUBROUTINE shr_mpi_bcasti0(vec,comm)
    IMPLICIT none
    integer, intent(inout):: vec
    integer, intent(in)   :: comm
    integer               :: ierr,lsize
    lsize = 1
    call MPI_BCAST(vec,lsize,MPI_INTEGER,0,comm,ierr)
  END SUBROUTINE shr_mpi_bcasti0
  
  SUBROUTINE shr_mpi_bcastl0(vec,comm)
    IMPLICIT none
    logical, intent(inout):: vec
    integer, intent(in)   :: comm
    integer               :: ierr, lsize
    lsize = 1
    call MPI_BCAST(vec,lsize,MPI_LOGICAL,0,comm,ierr)
  END SUBROUTINE shr_mpi_bcastl0
end module boilerplate


program mpitest

#include <mpif.h>
  integer::rank,nprocs,ierr

  call MPI_INIT(ierr)
  call MPI_COMM_RANK(MPI_COMM_WORLD, rank, ierr)
  call MPI_COMM_SIZE(MPI_COMM_WORLD, nprocs, ierr)
  
  print*, "rank ", rank, " out of ", nprocs
  
  call MPI_FINALIZE(ierr)
end program mpitest

module swap PrgEnv-intel PrgEnv-gnu
module swap gcc gcc/10.1.0
And then build with ftn.

cori10% ftn module_interface.F90
module_interface.F90:7:6:

    7 |   save
      |      1
Warning: Legacy Extension: Blanket SAVE statement at (1) follows previous SAVE statement
module_interface.F90:22:19:

[...]

PR #4812, opened 2022-02-25
Update modules on PM (including gnu v10)

ndkeen:
On Perlmutter, update the version of modules as new defaults are available.
Default GNU version is 11, but here I'm only updating from 9.3 to 10.3

With GNU v10, we have to work-around some issue with compiler flags that I've added.
(ie adding `-fallow-argument-mismatch -fallow-invalid-boz` for all GNU builds using v10 and higher)

Fixes: https://github.com/E3SM-Project/E3SM/issues/4811
Fixes: https://github.com/E3SM-Project/E3SM/issues/3954


whannah1 (approved):

ndkeen:
I think I've decided to separate this into 2 different PRs: one to update PM modules and the other to adjust fortran flags for GNU v10 +

PR #4817, opened 2022-02-28
Add GNU compiler flags to help build with GNU v10 and above

ndkeen:
Specifically, adding -fallow-argument-mismatch -fallow-invalid-boz for any GNU build using version 10 or higher.

Fixes https://github.com/E3SM-Project/E3SM/issues/3954

[bfb]


cime_config/machines/cmake_macros/gnu.cmake
```diff
@@ -40,6 +40,9 @@ endif()
 if (NOT DEBUG)
   string(APPEND FFLAGS " -O")
 endif()
+if (CMAKE_Fortran_COMPILER_VERSION VERSION_GREATER_EQUAL 10)
+   string(APPEND FFLAGS " -fallow-argument-mismatch -fallow-invalid-boz")
```
jayeshkrishna, line 44: We also found that we needed the "-fallow-argument-mismatch" flag for building SCORPIO with gnu 11

jayeshkrishna (approved):

ndkeen:
Note it seems like these are work-around flags and there may be other ways to address.

The `-fallow-argument-mismatch` flag is used to make these warnings instead of errors:
```
   22 |     call MPI_BCAST(vec,lsize,MPI_INTEGER,0,comm,ierr)
      |                   1
......
   31 |     call MPI_BCAST(vec,lsize,MPI_LOGICAL,0,comm,ierr)
      |                   2
Error: Type mismatch between actual argument at (1) and actual argument at (2) (INTEGER(4)/LOGICAL(4)).
```

The `-fallow-invalid-boz` flag is actually only needed for one line in the code.
In `components/eam/src/control/cam_history.F90`, there is this line:
`integer, parameter :: gen_hash_key_offset = z'000053db'`

Which looks to be used to init a hash before adding values. Not familiar

cime_config/machines/cmake_macros/gnu.cmake
ndkeen: I suspect you need it for GNU v10 as well

jayeshkrishna:
Are the changes BFB (BFB flag is missing in the PR desc, and the PR labels)?

ndkeen:
I added a commit to the branch to change how we issue the `-fallow-invalid-boz` flag. Now it's only used for 2 specific files via Depends.gnu*.  Re-tested e3sm_developer.

ndkeen:
Closing this in favor of https://github.com/E3SM-Project/E3SM/pull/4822

PR #4821, opened 2022-03-01
Add `-fallow-argument-mismatch` to gfortran

xylar:
This allows the code to build with gfortran >=10.0.0

Related to #4817, but for standalone MPAS builds


xylar:
## Testing

With this fix, I was able to build with gfortran 10.3.0, whereas I got errors (see https://github.com/MPAS-Dev/compass/issues/323) without it.

With this fix, I successfully ran the ocean `nightly` test suite with gfortran 10.3.0.

xylar:
This is not working for gfortran < 10, so I need to work on that.

xylar:
I'm still testing this on Anvil but it does at least build with an older gfortran.

components/mpas-framework/Makefile
```diff
@@ -296,6 +296,12 @@ intel-mpi:
...
 
 gfortran:
+	GFORTRAN_GTE_10=$$(expr `gfortran -dumpversion | cut -f1 -d.` \>= 10) ;\
+	if [ "$${GFORTRAN_GTE_10}" = "1" ]; then \
+	    FFLAGS="-fallow-argument-mismatch"; \
+	else \
+	    FFLAGS=""; \
+	fi; \
 	( $(MAKE) all \
 	"FC_PARALLEL = mpif90" \
```
xylar, line 306: @philipwjones, do you know of a more elegant way to do this?  If so, I'm keen to learn.

philipwjones: No, not really.

philipwjones (approved):
You might consider using a more specific FFLAG name just to avoid confusion with the final FFLAGS variable that's created later for all flags.

But I approve regardless. This appears to be needed for gnu builds.

xylar:
@philipwjones, I called it `EXTRA_FFLAGS`, instead of `FFLAGS`.

xylar:
@mark-petersen, @akturner and @matthewhoffman, this is ready for your review.

components/mpas-framework/Makefile
```diff
@@ -296,6 +296,12 @@ intel-mpi:
 	"CPPFLAGS = $(MODEL_FORMULATION) -D_MPI" )
 
 gfortran:
+	GFORTRAN_GTE_10=$$(expr `gfortran -dumpversion | cut -f1 -d.` \>= 10) ;\
+	if [ "$${GFORTRAN_GTE_10}" = "1" ]; then \
```
matthewhoffman, line 300: Should this be single dollar sign?  My understanding was that $$ accesses shell variables, but single dollar sign should be used for Makefile variables (https://stackoverflow.com/questions/26564825/what-is-the-meaning-of-a-double-dollar-sign-in-bash-makefile).  If my understanding is correct, it is interesting that using $$ here still works.

The $$ is also used below on all the lines below.

mark-petersen: Huh. I tested the makefile but checking for version `>=1` so it adds the flag. Compile dies with an error with gnu V6 (as expected) because the flag `-fallow-argument-mismatch` is undefined.  I tried to change the double dollar signs to `$(GFORTRAN_GTE_10)` and `$(EXTRA_FFLAGS)` and it didn't work (the variables don't contain the correct values). So @xylar did it right - I just don't understand why these are shell variables rather than makefile variables.

mark-petersen (approved):
Tested with gnu V6 on grizzly, and it compiles fine. I also changed to check for version `>= 1` and confirmed that the flag is added, since I don't have a V10 compiler to test.

mark-petersen:
@akturner I compiled and tested above. Should be the same for sea ice, since we all use the same Makefile. Can you test or approve based on my comments? Thanks.

components/mpas-framework/Makefile
matthewhoffman: Ok, thanks for the explanation and testing.  I'm convinced. :)

matthewhoffman (approved):

akturner (approved):
Passed MPAS-Seaice build and testing

jonbob:
merged to next

jonbob:
merged to master

xylar:
Thanks @jonbob, I really appreciate it!

xylar:
Thanks @jonbob, I really appreciate it!

xylar:
Thanks @jonbob, I really appreciate it!

Issue #4809, opened 2022-02-24
GNU v9 broken on PM after software maintenance


ndkeen:
During software maintenance today on PM, the GNU builds are broken. That is for `gnu` and `gnugpu` compilers.
We are using the gcc/9.3 compiler and I verified that even a simple test fails:

```
login29% module load gcc/9.3.0
login29% cc mpi-hostname.c
cc1: error: bad value ('znver3') for '-march=' switch
cc1: note: valid arguments to '-march=' switch are: nocona core2 nehalem corei7 westmere sandybridge corei7-avx ivybridge core-avx-i haswell core-avx2 broadwell skylake skylake-avx512 cannonlake icelake-client icelake-server cascadelake bonnell atom silvermont slm goldmont goldmont-plus tremont knl knm x86-64 eden-x2 nano nano-1000 nano-2000 nano-3000 nano-x2 eden-x4 nano-x4 k8 k8-sse3 opteron opteron-sse3 athlon64 athlon64-sse3 athlon-fx amdfam10 barcelona bdver1 bdver2 bdver3 bdver4 znver1 znver2 btver1 btver2 native; did you mean 'znver1'?
```

There is a work-around that can be used which is to load `craype-x86-rome` module. I can do this in `config_machines.xml` with a PR, but it turns out we can simply `module load craype-x86-rome` and the current settings will work. We don't use `module purge`, so this change will still be present during build.

Or, we can update to gcc/10.3 or the now default gcc/11.2.
I am testing (via e3sm_developer) using 10.3 which requires both:
https://github.com/E3SM-Project/E3SM/pull/4817 and https://github.com/E3SM-Project/E3SM/pull/4822

ndkeen:
This is now fixed as we are now using updated modules (including gnu v10).
PR #4818

PR #4822, opened 2022-03-01
Add flags for GNU builds to allow using GNU v10 and higher versions

ndkeen:
Add -fallow-argument-mismatch for all GNU builds (including gnugpu).

Only add -fallow-invalid-boz for 2 specific files for gnu/gnugpu builds.

This PR adds Depends.gnugpu.cmake which is similar to Depends.gnu.cmake

Fixes https://github.com/E3SM-Project/E3SM/issues/3954
And is needed first for https://github.com/E3SM-Project/E3SM/issues/4809 and https://github.com/E3SM-Project/E3SM/pull/4818

[BFB]


cime_config/machines/Depends.gnugpu.cmake
```diff
@@ -0,0 +1,24 @@
+list(APPEND MPAS_ICE_SHORTWAVE
```
jayeshkrishna, line 1: Following the patterns of the other changes in this PR you might want to rename the list to DISABLE_TREE_PTA_LIST or something like that

jayeshkrishna: And you might want to amend the commit message to include this flag

ndkeen: The MPAS_ICE_SHORTWAVE name was already there. It's fine to rename it...
I don't know what you mean about amending commit message

jayeshkrishna: The commit message (or PR desc) does not mention the new compiler flag, -fno-tree-pta

ndkeen: Oh I see -- that was already present for gnu. So adding this flag for GNU is not new, it's just we didn't have a gnugpu Depends file yet and therefore builds with gnugpu might also need that.  I can redo PR to exclude adding those flags.

ndkeen: SO you think I should remove the flags to set `-fno-tree-pta`?

jayeshkrishna: No, just document that you are adding the flag in this PR

jayeshkrishna: ah, ok, thanks

jayeshkrishna (approved):

ndkeen:
merged to next

PR #4826, opened 2022-03-09
Updates carbon conservation analysis member

njeffery:
1. tracks individual ocean-ice carbon  fluxes
2. computes accumulated error, relative errors, error bounds over the run
3. adds flag option to end the run if carbon errors exceed bounds
4. Also added three additional diagnostics
5. Adds molecular weight of carbon as a shared constant for all carbon conservation calculations
6. Non-BFB change is a correction in merge-fluxes-bgc which uses the initial ice area to merge BGC fluxes rather than the ice area after step therm 1.

Passes all tests in the sea ice testing suite. 

[NML]
[non-BFB] only for runs with active sea ice BGC.

### DIFF
--- a/components/mpas-seaice/testing_and_setup/configurations/standard_physics/namelist.seaice
+++ b/components/mpas-seaice/testing_and_setup/configurations/standard_physics/namelist.seaice
@@ -376,12 +376,13 @@
     config_AM_ridgingDiagnostics_write_on_startup = false
 /
 &AM_conservationCheck
-    config_AM_conservationCheck_enable = false
+    config_AM_conservationCheck_enable = true
     config_AM_conservationCheck_compute_interval = 'dt'
     config_AM_conservationCheck_output_stream = 'conservationCheckOutput'
     config_AM_conservationCheck_compute_on_startup = false
     config_AM_conservationCheck_write_on_startup = false
     config_AM_conservationCheck_write_to_logfile = true
+    config_AM_conservationCheck_carbon_failure_abort = false
 /
 &AM_geographicalVectors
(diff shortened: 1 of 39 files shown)


components/mpas-seaice/testing_and_setup/configurations/standard_physics/namelist.seaice
```diff
@@ -376,12 +376,13 @@
     config_AM_ridgingDiagnostics_write_on_startup = false
 /
 &AM_conservationCheck
-    config_AM_conservationCheck_enable = false
```
akturner, line 379: Why has this been changed?

akturner (changes requested):
This stalls in my tests during the conservation check. I'm investigating this. Also, why is the conservation check now on for the standard physics configuration?

akturner:
@njeffery: MPAS-Seaice stalls in the conservation write in master so this is not an issue with this PR. Switch config_AM_conservationCheck_enable to false and this should pass testing.

njeffery:
@akturner :  done

akturner (approved):

jonbob:
Initial testing failed because the value of config_AM_conservationCheck_carbon_failure_abort was set to a 'false' string instead of a simple logical false. I ran the scripts to autogenerate the bld files from Registry changes and it fixed that issue

rljacob:
telecon notes:  ready to go.

jonbob:
test merge passes:
* SMS_D_Ld1.T62_oEC60to30v3.DTESTM
except for expected NML DIFFs

jonbob:
merged to next

jonbob:

PR #4827, opened 2022-03-10
Modifies land-atm BGC compsets names and its use case files

BunnyVon:
This PR adds use_case for v2 BGC land-atm codes and fixes typos in the alias. Smoke tests are also added for these compsets.


singhbalwinder:
The PR looks good to me. I don't see any smoke tests added though. Would you like to add those? I am also assuming that the changes in the use case files are appropriate for v2

components/eam/cime_config/config_component.xml
```diff
@@ -71,7 +71,7 @@
       <!--  -->
       <value compset="GEOS_EAM"       >-offline_dyn</value>
       <value compset="GEOS_EAM"      >-nlev 56</value>
-      <value compset="_EAM.*_BGC%B"   >-co2_cycle</value>
+      <value compset="_EAM.*_BGC%*"   >-co2_cycle</value>
```
susburrows, line 74: I think these logic changes are safe for the compsets that we currently have implemented in the code (and so won't have unintended side effects), according to my notes here: https://acme-climate.atlassian.net/wiki/spaces/EBGC/pages/151290237/BGC+Specification+of+CO2+coupling+diagnostic+prognostic+radiative+BGC

However, I didn't double-check the code base to verify this.  Can you confirm that this was looked at?

susburrows (commented):
I had one question, but otherwise also agree that this looks good by visual inspection.  However, I agree with Balwinder that it would be good to add some smoke tests, and I don't see them here.

components/eam/cime_config/config_component.xml
BunnyVon: 1. Added smoke tests just now with help from @singhbalwinder 
2. I tested this changes before PR

beharrop: I've looked through the code some and the notes @susburrows linked here, and it doesn't look like there will be any problems with this change.  

I personally like this change since I don't think prognostic atmospheric CO2 should require compset naming to begin with the letter 'B' after the _BGC tag.

beharrop (approved):

components/eam/cime_config/config_component.xml
susburrows: Thanks for cross-checking this, @beharrop , and for adding the smoke tests, @BunnyVon and @singhbalwinder .

susburrows (approved):

susburrows:
@singhbalwinder and @rljacob , I think this PR is ready to go, unless you have further comments.  (@jonbob  is also listed on the PR, but I don't think it's truly necessary for him to provide an additional review; the changes are fairly modest and have already been reviewed by both Bryce and myself, as well as tested by Sha and Balwinder).

Could this PR please be prioritized for being merged?  It is critical for the BGC v2.1 land-atmosphere experiments.

rljacob (approved):

singhbalwinder:
Merged to next

jonbob (approved):

rljacob:
Should the new smoke tests use a tri-grid?  Is BGC still planning on using those?

rljacob:
Someone will need to update the tests to use the right grids and then this has to be remerged to next.

singhbalwinder:
I can update the tests but I don't know the right grids.

jonbob:
@singhbalwinder -- I think it should be "ne30pg2_r05_EC30to60E2r2"

singhbalwinder:
Thanks @jonbob ! I will update the tests and remerge shortly.

singhbalwinder:
Asked to generate baselines for the tests (Chrysalis integration next):
```
SMS_Ln5.ne30pg2_r05_EC30to60E2r2.BGCEXP_LNDATM_CNPRDCTC_1850.chrysalis_intel

SMS_Ln5.ne30pg2_r05_EC30to60E2r2.BGCEXP_LNDATM_CNPRDCTC_20TR.chrysalis_intel
```
Merging to master

BunnyVon:
Ditto. Thank you very much for the help @singhbalwinder and @wlin7

PR #4829, opened 2022-03-11
HOMME: add CMake logic to conditionally build composef90

### LINKED ISSUE #4828: HOMME: provide CMake logic to avoid building the composef90 library
Currently, in Homme, when `HOMME_ENABLE_COMPOSE=ON`, the f90 library is always built, and the c++ one is built only if (one of) the kokkos targets is to be built. This is fine for virtually all standalone testing, where we seldom (if ever) build a kokkos target without building any f90 target. However, in SCREAM, we need to build only kokkos targets, and we have no use for composef90, so we could provide some CMake logic to allow skipping it.

bartgol:
In particular, build composef90 if and only if preqx, theta-l, or homme_tool are to be built. Otherwise, do not build the library.

Closes #4828 .

P.s: currently, in E3SM, Homme's CMake build only impacts homme standalone build. In SCREAM/EAMxx, changes to Homme's cmake build system impact the atm component, but since that's a separate repo, I only added the standalone label. I'm not really sure what the policy is in this borderline case.


ambrad (approved):

ambrad:
I recommend that you run the homme_integration tests on a supported platform and note the status here.

oksanaguba:
i will run homme suites on compy now.

bartgol:
BTW, I manually run cmake for a few configurations ("enable this, but not that" kind of tests), and verified only the needed compose targets are built.

oksanaguba:
in next

Issue #4787, opened 2022-02-10
Add initial support for AOCC (AMD optimizing compiler)


ndkeen:
Add initial support for AOCC and test on Perlmutter. Current version on PM is 3.1, but I see AMD has released 3.2.
Perlmutter currently only has GNU/nvidia compilers (not Intel).
The compiler names for AOCC are clang/flang, but I was told the flang implementation is actually older open-sourced PGI work. 
Heard rumors that AOCC might out-perform GNU on AMD hardware.
While this compiler might not be able to be used with GPUs, it will be a great option for CPU-only work when the phase2 (cpu-only) nodes of PM arrive.

Created branch `ndk/machinefiles/initial-aocc`

There are some issues with the compiler, including an internal compiler error with MPAS as well
as an issue using `-g` with optimization for flang.

However, F cases seem to be running and getting reasonable performance on the single AMD cpu of
perlmutter nodes. Using `SMS_P256x2.ne30pg2_ne30pg2.F2010-CICE` as a test (which would use 4 nodes on PM):
```
          TOT SYPD
GNU         3.0
AOCC        2.6
```

And after https://github.com/E3SM-Project/E3SM/pull/4786, the performance of CAM_run2 is much improved:

```
            TOT SYPD
AOCC            2.9    -O2
AOCC            3.0    -Ofast
AOCC            3.26   -O2 with AMD math library
```

There may be some other quick fixes, currently CAM_run3 and CAM_run2 are faster with aocc over gnu. I'm looking at a problem spot in CAM_run1 now.

Only minor testing so far with aocc. As there are cooling issues at NERSC, only a third of machine is available now. These 2 tests pass:
```
ERS_D.ne4pg2_ne4pg2.F2010-CICE.perlmutter_aocc
ERS_D.ne4pg2_oQU480.F2010-CICE.perlmutter_aocc
```

Interesting to note that the initial perf issue found in CAM_run2 (already re-worked by @ambrad !) was tracked down to the cost of fortran intrinsic `rehaspe` function. It happened to be in an inner loop, but still seems like an intrinsic should do well or at least not have drastic perf differences between compilers.  However, I found the following:
![reshape](https://user-images.githubusercontent.com/5684727/153487933-aec3c225-642b-4a65-803b-b5b78dcc69ea.png)

minxu74:
[...]

PR #4830, opened 2022-03-11
Add gnu compiler support on Crusher and AMD.

grnydawn:
Changes:
* add crusher configuration in config_machines.xml, config_batch.xml
* create gnu_crusher.cmake
* for testing amdclang and crayclang, cmake configurations for the compilers are also added
* add CPRAMD macro in shr_infnan_mod.F90.in to use ieee arithmetic

[BFB]


ndkeen:
It might be better to separate the GNU changes in this PR and the AMD compiler changes.

I was going to bring in support for the AMD compiler here https://github.com/E3SM-Project/E3SM/issues/4787

rljacob:
I already delayed this PR that people are waiting on.  It can go ahead with the AMD additions.

minxu74:
@amametjanov could you approve the PR? It shall be the same as the #4793, but has a clean history.

rljacob (approved):

minxu74:
Okay. I am merging it now

amametjanov (approved):

grnydawn:
@ndkeen Most of configurations added in this PR for amdclang are isolated to only Crusher system(using amdclang_crusher.cmake). The only configurations that affect to the amdclang compiler on other systems are generic and shown below:

if (compile_threaded)
  string(APPEND FFLAGS " -fopenmp")
  string(APPEND CFLAGS " -fopenmp")
  string(APPEND CXXFLAGS " -fopenmp")
  string(APPEND LDFLAGS " -fopenmp")
endif()
string(APPEND CPPDEFS " -DNO_R16 -DCPRAMD -DFORTRANUNDERSCORE")

ndkeen:
That's fine, I can adjust --  just think it's cleaner to keep PR topics narrowly focused. One undervalued reason is to reduce headaches when we need to go back in history to see which PR's might have caused an issue.

Issue #4824, opened 2022-03-08
Runtime error (INF values) with MMF test using GNU v10


ndkeen:
After making PRs (now in master) to update GNU compiler on PM to v10, we found that tests of the form:
`SMS_P4x1.ne4pg2_ne4pg2.F-MMFXX`

were hitting runtime fails that looked like:

```
0:  ERROR: shr_reprosum_calc ERROR: NaNs or INFs in input
1: SHR_REPROSUM_CALC: Input contains  0.10000E+01 NaNs and  0.00000E+00 INFs on process       1
1:  ERROR: shr_reprosum_calc ERROR: NaNs or INFs in input
```

We did not see error with DEBUG. Doing bisection on the files to be compiled and reducing optimization levels, found that
I could build optimized, if `components/eam/src/dynamics/se/inidat.F90` was built with reduced optimization. Further, I found that it was actually a new algorithm in GNU v10 that might be the problem:

From  https://gcc.gnu.org/gcc-10/changes.html
"At any optimization level except-Os, gfortran now uses inline packing for arguments instead of calling a library routine. If the source contains a large number of arguments that need to be repacked, code size or time for compilation can become excessive. If that is the case, -fno-inline-arg-packing can be used to disable inline argument packing. "

It seems adding `-fno-inline-arg-packing` would then go back to previous treatment. I see that I only needed this flag on the one file noted above, but perhaps we could consider using it wider.

`ndk/machinefiles/PM-add-arg-pack-flag-gnu10`

Needed this for `gnu` and `gnugpu` builds. We could make the change only for Perlmutter -- not yet sure if same issue exists on other machines with gnuv10.

PR #4836, opened 2022-03-16
For Perlmutter, with gnu v10, add compiler flag to work-around issue with MMF tests

ndkeen:
Add a flag to disable new behavior in gnu v10. For now, just do this only for Perlmutter and gnu v10+.

This PR adds to new files to be specific to Perlmutter for now.

Flag itself does not change results, but the tests were hitting errors without it.

Fixes https://github.com/E3SM-Project/E3SM/issues/4824

### DIFF
--- a/cime_config/machines/Depends.perlmutter.gnu.cmake
+++ b/cime_config/machines/Depends.perlmutter.gnu.cmake
@@ -1,0 +1,14 @@
+list(APPEND NO_INLINE_ARG_PACKING_LIST
+  eam/src/dynamics/se/inidat.F90
+)
+
+if (CMAKE_Fortran_COMPILER_VERSION VERSION_GREATER_EQUAL 10)
+
+  if (NOT DEBUG)
+    # new in gnu10, inline arg packing was causing INF values with SMS_P4x1.ne4pg2_ne4pg2.F-MMFXX
+    foreach(ITEM IN LISTS NO_INLINE_ARG_PACKING_LIST)
+      e3sm_add_flags("${ITEM}" " -fno-inline-arg-packing")
+    endforeach()
+  endif()
+
+endif()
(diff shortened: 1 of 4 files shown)


ndkeen:
merged to next

cime_config/machines/Depends.perlmutter.gnu.cmake
```diff
@@ -0,0 +1,14 @@
...
+if (CMAKE_Fortran_COMPILER_VERSION VERSION_GREATER_EQUAL 10)
+
+  if (NOT DEBUG)
+    # new in gnu10, inline arg packing was causing INF values with SMS_P4x1.ne4pg2_ne4pg2.F-MMFXX
+    foreach(ITEM IN LISTS NO_INLINE_ARG_PACKING_LIST)
+      e3sm_add_flags("${ITEM}" " -fno-inline-arg-packing")
+    endforeach()
+  endif()
+
+endif()
```
amametjanov, line 14: I think it'd be better to move this to `Depends.gnu[gpu].cmake` and remove several duplicated mach-specific files.

ndkeen: At the time, I wasn't sure if this was something specific to the machine or compiler install. Are you seeing this helps on other machines?

amametjanov: Yes, on Crusher with gcc/11.2.0: doing this in commit https://github.com/E3SM-Project/E3SM/pull/5132/commits/b25627187cc3bc6718c3ec99449af4cb071959d4 as part of #5132

ndkeen: ok i made an issue https://github.com/E3SM-Project/E3SM/issues/5136

Issue #4838, opened 2022-03-17
Module changes needed after Cori maintenance March 16th


ndkeen:
After todays maintenance, there are some module versions (and some modules) E3SM was using that are no 
longer available. I think we can keep the Intel compiler and other important modules at the same version, so this will hopefully be a BFB change.
It does include moving the gnu compiler version to 10.3.0.

Testing this branch
`ndk/machinefiles/cori-maint-cleanup-mar16`

Example of error at setup without these module changes:
```
ERROR: module command /opt/modules/default/bin/modulecmd python load PrgEnv-intel/6.0.5 failed with message:
PrgEnv-intel/6.0.5(80):ERROR:150: Module 'PrgEnv-intel/6.0.5' conflicts with the currently loaded module(s) 'PrgEnv-intel/6.0.10'
PrgEnv-intel/6.0.5(80):ERROR:102: Tcl command execution failed: conflict PrgEnv-intel
```

PR #4839, opened 2022-03-17
For Cori, update module versions after March 16th maintenance

ndkeen:
After todays maintenance, there are some module versions (and some modules) E3SM was using that are no
longer available. I think we can keep the Intel compiler and other important modules at the same version, so this will hopefully be a BFB change.
It does include moving the gnu compiler version to 10.3.0.

Fixes https://github.com/E3SM-Project/E3SM/issues/4838

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -431,15 +431,15 @@
       </modules>
 
       <modules compiler="intel">
-        <command name="load">PrgEnv-intel/6.0.5</command>
+        <command name="load">PrgEnv-intel/6.0.10</command>
         <command name="rm">intel</command>
         <command name="load">intel/19.0.3.199</command>
       </modules>
 
       <modules compiler="gnu">
-        <command name="swap">PrgEnv-intel PrgEnv-gnu/6.0.9</command>
+        <command name="swap">PrgEnv-intel PrgEnv-gnu/6.0.10</command>
         <command name="rm">gcc</command>
-        <command name="load">gcc/9.3.0</command>
+        <command name="load">gcc/10.3.0</command>
         <command name="rm">cray-libsci</command>
         <command name="load">cray-libsci/20.09.1</command>
       </modules>
@@ -467,10 +467,8 @@
       </modules>
 
       <modules>
-        <command name="rm">git</command>
-        <command name="load">git</command>
         <command name="rm">cmake</command>
-        <command name="load">cmake/3.21.3</command>
+        <command name="load">cmake</command>
         <command name="load">perl5-extras</command>
       </modules>
     </module_system>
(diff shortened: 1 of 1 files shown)


ndkeen:
Testing with e3sm_developer passes all tests on cori-knl and cori-haswell, so will go ahead and merge to next.

ndkeen:
Yes, for all maint versions -- but I wasn't sure if maint-2.0 works the same as the others? ie like maint-1.0, maint-1.1, maint-1.2 ?

ndkeen:
I also tested this branch with gnu compiler (that now uses gnuv10) and I do see some errors.  I will merge this to master as we need these changes for anything to work at all.  And then look at gnu issues.

```
ERS.ne4_oQU240.F2010.cori-haswell_gnu.eam-hommexx.rgnu                                            fail COMPARE_base_rest
ERS_Ld3.ne4pg2_oQU480.F2010.cori-haswell_gnu.eam-thetahy_sl_pg2.rgnu                              fail               RUN
ERS_Ld3.ne4pg2_oQU480.F2010.cori-haswell_gnu.eam-thetahy_sl_pg2_ftype0.rgnu                       fail               RUN
ERS_Ln9.ne4pg2_ne4pg2.F-MMF1.cori-haswell_gnu.eam-mmf_crmout.rgnu                                 fail               RUN
SMS_Ld30.f45_f45.IELMFATES.cori-haswell_gnu.elm-fates_satphen.rgnu                                fail               RUN
SMS_Ln5.ne4pg2_oQU480.F2010.cori-haswell_gnu.eam-thetahy_pg2.rgnu                                 fail               RUN
SMS_Ln5.ne4pg2_oQU480.F2010.cori-haswell_gnu.eam-thetahy_sl_pg2.rgnu                              fail               RUN
SMS_Ln5.ne4pg2_oQU480.F2010.cori-haswell_gnu.eam-thetahy_sl_pg2_ftype0.rgnu                       fail               RUN
SMS_Ln5.ne4pg2_oQU480.F2010.cori-haswell_gnu.rgnu                                                 fail               RUN
```

cime_config/machines/config_machines.xml
```diff
@@ -620,10 +618,8 @@
       </modules>
 
       <modules>
-        <command name="rm">git</command>
-        <command name="load">git</command>
```
jonbob, line 624: @ndkeen - did you mean to remove the git module?

ndkeen: Yea -- looks like the removed the module. We never needed a specific version, I just thought it was safer to do so. 
```
cori09% module load git
ModuleCmd_Load.c(244):ERROR:105: Unable to locate a modulefile for 'git'
cori09% which git
/usr/bin/git
cori09% git --version
git version 2.26.2
```

jonbob: It does cause problems not to have it -- one of our users reported this:
````
----- Starting case_setup -----
Configuring E3SM to use the COSP simulator
Successfully cleaned env_mach_specific.xml
ERROR: module command /opt/modules/default/bin/modulecmd python load git failed with message:
ModuleCmd_Load.c(244):ERROR:105: Unable to locate a modulefile for 'git'
````

jonbob: That error message was from an e3sm runscript that previously worked

ndkeen: Right, if there isn't a git module, then loading it will cause an error in CIME (unless we ignore module errors, which might be dangerous) -- therefore, only thing to do is NOT try to load the module.

jonbob: The runscript isn't trying to load the git module -- so CIME must be trying to somewhere under the covers. Is there a reason not to load the git module?

ndkeen: Yes.  In fact, there is no git module. NERSC removed it.  This change will correct the problem.

jonbob: Thanks @ndkeen -- I think I understand what's going on now. I appreciate all your help

Issue #4840, opened 2022-03-17
Update Cori modules for maint-1.0 after March 16th maintenance


ndkeen:
Following https://github.com/E3SM-Project/E3SM/pull/4839, making similar changes to maint-1.0.

One major difference is that for maint-1.0, on cori-knl, we were still using intel v18 and that is no longer available. This change brings in the same v19 version we are using on master. Therefore, would not expect BFB results.

`ndk/machinefiles/maint-1.0-cori-module-march16`

Example of error at setup without these module changes:

```
ERROR: module command /opt/modules/default/bin/modulecmd python load PrgEnv-intel/6.0.5 failed with message:
PrgEnv-intel/6.0.5(80):ERROR:150: Module 'PrgEnv-intel/6.0.5' conflicts with the currently loaded module(s) 'PrgEnv-intel/6.0.10'
PrgEnv-intel/6.0.5(80):ERROR:102: Tcl command execution failed: conflict PrgEnv-intel
```

rljacob:
Hmm, not sure what to do for users about the possible BFB change.  If they can't run with Intel18, we obviously have to update.  Our v1.0 runs were with edison so we were never going to get BFB.

ndkeen:
We actually had a similar issue during the v1hires runs on Theta where we needed to update Intel version during the simulation.

ndkeen:
Note that I updated the GNU compiler to 10.3, but it would require changes in compiler flags to make it work. I added a comment as a suggestion for what might work, but when I tested, I ran into another known (and fixed) build problem in land. So for now, was going to leave as-is as we weren't really using GNU anyway.

ndkeen:
Was fixed by https://github.com/E3SM-Project/E3SM/pull/4841, not sure why it wasn't closed.

PR #4841, opened 2022-03-17
For Cori, update module versions in maint-1.0 after March 16th maintenance.

rljacob:
This includes an unavoidable version change for the Intel compiler from 18 to 19,
so would expect results to be non BFB. The older compiler no longer
available.

Fixes https://github.com/E3SM-Project/E3SM/issues/4840

[NBFB]


ndkeen:
merged to maint-1.0

rljacob:
Note: this did cause all maint-1.0 prod tests to diff on cori-knl.

PR #4844, opened 2022-03-18
Add and update SSP370 and SSP585

jjbenedict:
Added SSP370 compset "WCYCLSSP370" and updated existing compset "WCYCLSSP585".  For SSP370, the new "WCYCLSSP370" compset was defined, new EAM and ELM use case files were created, new prescribed forcing files were created, and minor changes elsewhere were added by emulating what was done for SSP585.  For the updated SSP585, the "WCYCLSSP585" compset was defined, new land use and land cover change files were created (these files are grid-specific, and the default atm_lnd grid changed from v1 to v2), and the ELM use case file was updated to point to these new files.  The atmosphere forcing files set in the SSP585 use case were not changed because they are not grid specific and they already existed from v1.  All SSP370 EAM and ELM input files were staged in NERSC-Cori's inputdata directory only.

### DIFF
--- a/cime_config/allactive/config_compsets.xml
+++ b/cime_config/allactive/config_compsets.xml
@@ -75,6 +75,11 @@
   <lname>SSP585SOI_EAM%CMIP6_ELM%SPBC_MPASSI_MPASO_MOSART_SGLC_SWAV</lname>
 </compset>
 
+<compset>
+  <alias>WCYCLSSP370</alias>
+  <lname>SSP370SOI_EAM%CMIP6_ELM%SPBC_MPASSI_MPASO_MOSART_SGLC_SWAV</lname>
+</compset>
+
 <!-- E3SM v1 science compsets -->
 
 <compset>
@@ -351,7 +356,8 @@
   <entry id="RUN_STARTDATE">
     <values>
       <value  compset="20TR_EAM">1850-01-01</value>
-      <value  compset="SSP585_EAM">2015-01-01</value>
+      <value  compset="SSP585.*_EAM">2015-01-01</value>
+      <value  compset="SSP370.*_EAM">2015-01-01</value>
     </values>
   </entry>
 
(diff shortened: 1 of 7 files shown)


rljacob:
note from telecon:  this is being updated.

tangq (approved):
The changes seem good to me. The input data files should also be uploaded to the E3SM input data server.

golaz (approved):
Approving pending final review, merge by @wlin7.

golaz:
@wlin7 : can you also make a corresponding PR for maint-2.0? Thanks.

wlin7:
Hi @golaz , I think we don't need a separate PR for maint-2.0. The branch was essentially based on maint-2.0 (the master before maint-2.0 was created). When the next test comes back all clean, it would suggest good to go for both master and maint-2.0 and I would merge into them separately.

rljacob:
You don't need a separate *branch* but a merge of this branch to maint-2.0 should also be done through a PR.

wlin7:
Ok, Rob. Have a tech question. I did plan to use the same branch to create two PRs, and from what I found on the web, it is allowed doing so. But don't see a clickable button/link that would let me compare/create a new PR with this branch - when the branch, the link at the usual spot for creating PR is bound to this PR.

rljacob:
At the top of the pull request summary page, https://github.com/E3SM-Project/E3SM/pulls, choose "New Pull Request".  Set the base to maint-2.0 and the "compare" to the branch.

wlin7:
Thanks, Rob. That is certainly a more elegant way. Didn't know this path to find the 'compare' function, I just manually enter URL: https://github.com/E3SM-Project/E3SM/compare/maint-2.0...benedict/atm/ssp370_ssp385 to obtain the compare page then created the PR. Nice there is a much easier way.

wlin7:
merged to next.
The conflict in components/eam/cime_config/config_component.xml was due to a BGC PR that modified the use_case name (20TR_cam5_CMIP6_bgc) and the associated matching pattern for the compset.

wlin7:
Merged to master. 

All input files needed by SSP370 are available on the inputdata server.

rljacob:
Should we add tests for these cases ?

wlin7:
Good idea, Rob. I think we can add them to prod test for maint-2.0 and another pair to next/master just like other WCYCL1850. I will put the tests in a new PR for fixing a SSP585 issue.

Issue #4845, opened 2022-03-18
Update Cori modules for maint-1.2 after March 16th maintenance


ndkeen:
Following https://github.com/E3SM-Project/E3SM/pull/4839, making similar changes to maint-1.2

Some module versions need updating, some no longer avail. 
For cori-knl, the default Intel compiler is now v19, where it was v18 before.
Therefore, the results will not be identical.
This branch does contain a `intel19` compiler which is now the same as `intel`.

Branch `ndk/machinefiles/maint12-cori-module-updates`

Example of error at setup without these module changes:

```
ERROR: module command /opt/modules/default/bin/modulecmd python load PrgEnv-intel/6.0.5 failed with message:
PrgEnv-intel/6.0.5(80):ERROR:150: Module 'PrgEnv-intel/6.0.5' conflicts with the currently loaded module(s) 'PrgEnv-intel/6.0.10'
PrgEnv-intel/6.0.5(80):ERROR:102: Tcl command execution failed: conflict PrgEnv-intel 
```

ndkeen:
Was fixed by 4847, not sure why it wasn't closed.

xylar:
> not sure why it wasn't closed

The `fixes` keyword in the PR description will only do something if you merge into the default branch (`master` in this repo).

PR #4847, opened 2022-03-18
Update modules for Cori on maint-1.2

rljacob:
Updating module settings after Cori maintenance.
Similar to https://github.com/E3SM-Project/E3SM/pull/4839.
The default intel compiler for cori-knl is now v19, and will not be BFB.

Fixes https://github.com/E3SM-Project/E3SM/issues/4845


ndkeen:
merged to maint-1.2

PR #4852, opened 2022-03-23
Update pelayouts for EAM tri-grid tests

amametjanov:
Update pelayouts for EAM tri-grid tests running into OOM errors on 4 Anvil nodes.

[BFB]

### DIFF
--- a/cime_config/allactive/config_pesall.xml
+++ b/cime_config/allactive/config_pesall.xml
@@ -1460,19 +1460,19 @@
       </pes>
     </mach>
   </grid>
-  <grid name="a%ne30np4_l%r05_oi%oEC60to30v3_r%r05_g%null_w%null_z%null_m%oEC60to30v3">
+  <grid name="a%ne30np4.*_l%r05_oi%EC.+">
     <mach name="sandiatoss3|cori-knl|cori-haswell|theta|anvil|bebop">
       <pes compset="any" pesize="any">
-        <comment>none</comment>
+        <comment>pelayout for tri-grid tests with EAM</comment>
         <ntasks>
-          <ntasks_atm>256</ntasks_atm>
-          <ntasks_lnd>256</ntasks_lnd>
-          <ntasks_rof>256</ntasks_rof>
-          <ntasks_ice>256</ntasks_ice>
-          <ntasks_ocn>256</ntasks_ocn>
-          <ntasks_glc>256</ntasks_glc>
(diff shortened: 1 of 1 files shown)


rljacob:
Before: https://my.cdash.org/viewTest.php?onlyfailed&buildid=2145037
After: https://my.cdash.org/viewTest.php?buildid=2145156

sarich:
@rljacob I don't understand your comment.

rljacob:
I was just moving the URL's Az provided out of the PR description.   We don't want those in the merge commit.

sarich:
Ok I didn't like them either and edited them out

cime_config/allactive/config_pesall.xml
```diff
@@ -1460,19 +1460,19 @@
...
+        <comment>pelayout for tri-grid tests with EAM</comment>
+        <ntasks>
+          <ntasks_atm>-8</ntasks_atm>
+          <ntasks_lnd>-8</ntasks_lnd>
+          <ntasks_rof>-8</ntasks_rof>
+          <ntasks_ice>-8</ntasks_ice>
+          <ntasks_ocn>-8</ntasks_ocn>
+          <ntasks_glc>-8</ntasks_glc>
+          <ntasks_wav>-8</ntasks_wav>
+          <ntasks_cpl>-8</ntasks_cpl>
```
jonbob, line 1475: @amametjanov - what do the negative task numbers mean? I'm having a user running into trouble on cori with this resolution

amametjanov: Negative numbers note `node` count, positive numbers note MPI `task` count (divide by `MAX_MPITASKS_PER_NODE` to get the node count)

jonbob: Thanks for the explanation -- I did not know that. This PR is still causing issues on cori-knl -- a trigrid test setting up -compset WCYCL1850 -res ne30pg2_r05_EC30to60E2r2 fails with this error:
````
grid_choice a%ne30np4.pg2 grid_match a%ne30np4.*_l%r05_oi%EC.+
compset_choice any compset_match any
pesize_choice any pesize_match any
points = 10
ERROR: More than one PE layout matches given PE specs
````
But that error goes away if I get rid of this change

jonbob (commented):
Sorry -- I meant this to be a comment and not a review

PR #4835, opened 2022-03-16
Add choice of computed or constant ocean GM gravity wave speed (AMOC PR 1/6)

mark-petersen:
Currently the internal gravity wave speed (first baroclinic mode) in the ocean GM parameterization is a constant, previously named `config_GM_constant_gravWaveSpeed` but changed in this PR to `config_GM_constant_bclModeSpeed`. This remains the default, but this PR adds a new flag, `config_GM_minBclModeSpeed_method`: If 'constant' (the default) then use `config_GM_constant_bclModeSpeed`. If 'computed' then compute the gravity wave speed at every edge at every time step using the Brunt-Vaisala frequency. This functionality was tested by the AMOC focus team in 2021 on the branch [`amoc-experimental-branch`.](https://github.com/E3SM-Project/E3SM/tree/vanroekel/ocean/amoc-experimental-branch)

[NML]
[BFB]

PR #4864, opened 2022-03-30
Add horizontal tapering function to mesoscale eddy parameterizations (AMOC PR 2/6)

darincomeau:
This PR adds a new 'RossbyRadius' option to the horizontal tapering of GM and Redi, which is based on Hallberg (2013) - https://doi.org/10.1016/j.ocemod.2013.08.007, which turns the parameterization on if the Rossby Radius is resolved by at least two cells. These flags are:
```
config_Redi_horizontal_taper = 'ramp' (default), 'none', 'RossbyRadius' (new) 
config_GM_horizontal_taper = 'ramp' (default), 'none', 'RossbyRadius' (new) 
```
This PR replaces the variable `gmResolutionTaper`, which was used for both GM and Redi, to the separate variables `RediHorizontalTaper` and `gmHorizontalTaper` so that Redi and GM tapering can be chosen separately, and it is clear in the code. Logic was added to the GM and Redi schemes so that the there is a base value in the arrays for Redi Kappa and GM Kappa. The base value is then multiplied by the taper array, which has a value between 0 and 1. Arrays that are static in time are computed on init. Under the following conditions, the arrays are recomputed at every time step:
```
config_Redi_horizontal_taper = 'RossbyRadius' (recompute RediHorizontalTaper and RediKappa)
config_GM_horizontal_taper =  'RossbyRadius' (recompute gmHorizontalTaper)
```

[NML] adds MPAS-Ocean namelist options
[BFB]

### DIFF
--- a/components/mpas-ocean/bld/build-namelist
+++ b/components/mpas-ocean/bld/build-namelist
@@ -589,8 +589,8 @@ add_default($nl, 'config_GM_EG_Rhines_factor');
 add_default($nl, 'config_GM_horizontal_taper');
 add_default($nl, 'config_GM_horizontal_ramp_min');
 add_default($nl, 'config_GM_horizontal_ramp_max');
-add_default($nl, 'config_GM_Rossby_ramp_min');
-add_default($nl, 'config_GM_Rossby_ramp_max');
+add_default($nl, 'config_GMRedi_Rossby_ramp_min');
+add_default($nl, 'config_GMRedi_Rossby_ramp_max');
 
 ####################################
 # Namelist group: Rayleigh_damping #
--- a/components/mpas-ocean/bld/build-namelist
+++ b/components/mpas-ocean/bld/build-namelist
@@ -589,6 +589,8 @@ add_default($nl, 'config_GM_EG_Rhines_factor');
 add_default($nl, 'config_GM_horizontal_taper');
 add_default($nl, 'config_GM_horizontal_ramp_min');
 add_default($nl, 'config_GM_horizontal_ramp_max');
+add_default($nl, 'config_GM_Rossby_ramp_min');
+add_default($nl, 'config_GM_Rossby_ramp_max');
 
 ####################################
 # Namelist group: Rayleigh_damping #
--- a/components/mpas-ocean/bld/build-namelist
+++ b/components/mpas-ocean/bld/build-namelist
@@ -580,6 +580,7 @@ add_default($nl, 'config_Redi_min_layers_diag_terms');
 add_default($nl, 'config_use_GM');
 add_default($nl, 'config_GM_closure');
 add_default($nl, 'config_GM_constant_kappa');
+add_default($nl, 'config_gm_enable_horizontal_resolution_function');
 add_default($nl, 'config_GM_constant_bclModeSpeed');
 add_default($nl, 'config_GM_minBclModeSpeed_method');
 add_default($nl, 'config_GM_spatially_variable_min_kappa');
--- a/components/mpas-ocean/bld/build-namelist
+++ b/components/mpas-ocean/bld/build-namelist
@@ -550,14 +550,6 @@ add_default($nl, 'config_Leith_parameter');
 add_default($nl, 'config_Leith_dx');
 add_default($nl, 'config_Leith_visc2_max');
 
-####################################################
-# Namelist group: mesoscale_eddy_parameterizations #
-####################################################
-
-add_default($nl, 'config_eddying_resolution_taper');
-add_default($nl, 'config_eddying_resolution_ramp_min');
-add_default($nl, 'config_eddying_resolution_ramp_max');
-
 #########################################
 # Namelist group: Redi_isopycnal_mixing #
 #########################################
@@ -572,6 +564,9 @@ add_default($nl, 'config_Redi_N2_based_taper_enable');
 add_default($nl, 'config_Redi_N2_based_taper_min');
 add_default($nl, 'config_Redi_N2_based_taper_limit_term1');
 add_default($nl, 'config_Redi_min_layers_diag_terms');
+add_default($nl, 'config_Redi_horizontal_taper');
+add_default($nl, 'config_Redi_horizontal_ramp_min');
+add_default($nl, 'config_Redi_horizontal_ramp_max');
 
 ############################################
 # Namelist group: GM_eddy_parameterization #
@@ -580,7 +575,6 @@ add_default($nl, 'config_Redi_min_layers_diag_terms');
 add_default($nl, 'config_use_GM');
 add_default($nl, 'config_GM_closure');
 add_default($nl, 'config_GM_constant_kappa');
-add_default($nl, 'config_gm_enable_horizontal_resolution_function');
 add_default($nl, 'config_GM_constant_bclModeSpeed');
 add_default($nl, 'config_GM_minBclModeSpeed_method');
 add_default($nl, 'config_GM_spatially_variable_min_kappa');
@@ -592,6 +586,9 @@ add_default($nl, 'config_GM_EG_riMin');
 add_default($nl, 'config_GM_EG_kappa_factor');
 add_default($nl, 'config_GM_EG_Rossby_factor');
 add_default($nl, 'config_GM_EG_Rhines_factor');
+add_default($nl, 'config_GM_horizontal_taper');
+add_default($nl, 'config_GM_horizontal_ramp_min');
+add_default($nl, 'config_GM_horizontal_ramp_max');
 
 ####################################
 # Namelist group: Rayleigh_damping #
@@ -1664,7 +1661,6 @@ my @groups = qw(run_modes
                 hmix_leith
                 redi_isopycnal_mixing
                 gm_eddy_parameterization
-                mesoscale_eddy_parameterizations
                 rayleigh_damping
                 cvmix
                 gotm
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -2875,6 +2875,10 @@
 			 description="Horizontal tapering for GM. Varies between 0 and 1."
 			 packages="gm"
 		/>
+		<var name="RossbyRadius" type="real" dimensions="nEdges Time" units="m"
+			 description="Rossby Radius, computed in GM routine for some settings"
+			 packages="gm"
+		/>
 		<var name="surfaceFluxAttenuationCoefficient" type="real" dimensions="nCells Time" units="m"
 			description="The spatially-dependent length scale of exponential decay of surface fluxes. Fluxes are multiplied by $e^{z/\gamma}$, where this coefficient is $\gamma$."
 		/>
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -411,6 +411,14 @@
 					description="Maximum value in grid cell size for GM $\kappa$ ramp function.  Here cell size refers to dcEdge. Used when config_GM_horizontal_taper is set to ramp."
 					possible_values="Any positive real value."
 		/>
+		<nml_option name="config_GM_Rossby_ramp_min" type="real" default_value="0.5" units="unitless"
+					description="Minimum value of the ratio between grid-cell size (dcEdge) and Rossby radius for GM $\kappa$ ramp function.  Used when config_GM_horizontal_taper is set to RossbyRadius."
+					possible_values="Any positive real value."
+		/>
+		<nml_option name="config_GM_Rossby_ramp_max" type="real" default_value="3.0" units="unitless"
+					description="Maximum value of the ratio between grid-cell size (dcEdge) and Rossby radius for GM $\kappa$ ramp function.  Used when config_GM_horizontal_taper is set to RossbyRadius."
+					possible_values="Any positive real value."
+		/>
 	</nml_record>
 	<nml_record name="Rayleigh_damping" mode="forward">
 		<nml_option name="config_Rayleigh_friction" type="logical" default_value=".false." units="unitless"
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -411,12 +411,12 @@
 					description="Maximum value in grid cell size for GM $\kappa$ ramp function.  Here cell size refers to dcEdge. Used when config_GM_horizontal_taper is set to ramp."
 					possible_values="Any positive real value."
 		/>
-		<nml_option name="config_GM_Rossby_ramp_min" type="real" default_value="0.5" units="unitless"
-					description="Minimum value of the ratio between grid-cell size (dcEdge) and Rossby radius for GM $\kappa$ ramp function.  Used when config_GM_horizontal_taper is set to RossbyRadius."
+		<nml_option name="config_GMRedi_Rossby_ramp_min" type="real" default_value="0.5" units="unitless"
+					description="Minimum value of the ratio between grid-cell size (dcEdge) and Rossby radius for GM and Redi $\kappa$ ramp functions. Used when config_GM_horizontal_taper and/or config_Redi_horizontal_taper are set to RossbyRadius."
 					possible_values="Any positive real value."
 		/>
-		<nml_option name="config_GM_Rossby_ramp_max" type="real" default_value="3.0" units="unitless"
-					description="Maximum value of the ratio between grid-cell size (dcEdge) and Rossby radius for GM $\kappa$ ramp function.  Used when config_GM_horizontal_taper is set to RossbyRadius."
+		<nml_option name="config_GMRedi_Rossby_ramp_max" type="real" default_value="3.0" units="unitless"
+					description="Maximum value of the ratio between grid-cell size (dcEdge) and Rossby radius for GM and Redi $\kappa$ ramp functions. Used when config_GM_horizontal_taper and/or config_Redi_horizontal_taper are set to RossbyRadius."
 					possible_values="Any positive real value."
 		/>
 	</nml_record>
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -2860,11 +2860,11 @@
 			 packages="forwardMode;analysisMode"
 		/>
 		<var name="gmBolusKappa" type="real" dimensions="nEdges Time" units="m^2 s^{-1}"
-			 description="GM Bolus Kappa value.  This is constant in time, and set at init based on the constant or ramp fuction."
+			 description="GM Bolus Kappa value.  On output, it has NOT been multiplied by the horizontal taper array gmHorizontalTaper, because that is applied at the end to the normalGMBolusVelocity variable, not to the gmBolusKappa."
 			 packages="gm"
 		/>
 		<var name="RediKappa" type="real" dimensions="nEdges Time" units="m^2 s^{-1}"
-			 description="Redi Kappa value.  This is constant in time, and set at init based on the constant or ramp fuction."
+			 description="Redi Kappa value.  On output, it has already been multiplied by the horizontal taper array RediHorizontalTaper (as opposed to gmBolusKappa, which has not been multiplied by the horizontal taper)."
 			 packages="gm"
 		/>
 		<var name="RediHorizontalTaper" type="real" dimensions="nEdges Time" units="unitless"
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -295,8 +295,8 @@
 					possible_values=".true. or .false."
 		/>
 		<nml_option name="config_Redi_closure" type="character" default_value="constant" units="unitless"
-					description="Control what type of function is used for Redi $\kappa$. For 'equalGM', RediKappa is set to gmBolusKappa, so picks up the closure used by GM."
-					possible_values="'constant', 'equalGM', 'data'"
+					description="Control what type of function is used for Redi $\kappa$."
+					possible_values="'constant', 'data'"
 		/>
 		<nml_option name="config_Redi_constant_kappa" type="real" default_value="600.0" units="m^2 s^{-1}"
 					description="The Redi diffusion coefficient. Only used when config_Redi_closure = 'constant'."
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -289,20 +289,6 @@
 					possible_values="any positive real"
 		/>
 	</nml_record>
-	<nml_record name="mesoscale_eddy_parameterizations" mode="forward">
-		<nml_option name="config_eddying_resolution_taper" type="character" default_value="ramp" units="unitless"
-					description="Control how the Redi and GM Bolus $\kappa$ value varies as a function of horizontal resolution"
-					possible_values="'none' and 'ramp'"
-		/>
-		<nml_option name="config_eddying_resolution_ramp_min" type="real" default_value="20e3" units="m"
-					description="Minimum value in grid cell size for Redi and GM $\kappa$ ramp function.  Here cell size refers to dcEdge. Used when config_eddying_resolution_taper is set to ramp."
-					possible_values="Any positive real value."
-		/>
-		<nml_option name="config_eddying_resolution_ramp_max" type="real" default_value="30e3" units="m"
-					description="Maximum value in grid cell size for Redi and GM $\kappa$ ramp function.  Here cell size refers to dcEdge. Used when config_eddying_resolution_taper is set to ramp."
-					possible_values="Any positive real value."
-		/>
-	</nml_record>
 	<nml_record name="Redi_isopycnal_mixing" mode="forward">
 		<nml_option name="config_use_Redi" type="logical" default_value=".false." units="unitless"
 					description="If true, Redi isopycnal mixing is turned on"
@@ -344,6 +330,18 @@
 					description="Redi diagonal terms (2 and 3) are turned off from layer 1 through config_Redi_min_layers_diag_terms-1, and on from config_Redi_min_layers_diag_terms to nVertLevels. The Redi diagonal terms are not guaranteed to produce bounded tracer fields, and in practice produce growing temperatures in a few columns with fewer than 5 vertical cells. Redi is meant for isopycnal mixing in the deep ocean, so not applying Redi diagonal terms in very shallow regions is an acceptable solution."
 					possible_values="any integer between 0 (all layers on) and nVertLevels (all layers off)"
 		/>
+		<nml_option name="config_Redi_horizontal_taper" type="character" default_value="ramp" units="unitless"
+					description="Control how the Redi Bolus $\kappa$ value varies as a function of horizontal resolution. 'none' is constant, 'ramp' is strictly based on resolution, 'RossbyRadius' follows Hallberg (2013) - https://doi.org/10.1016/j.ocemod.2013.08.007"
+					possible_values="'none', 'ramp', 'RossbyRadius'"
+		/>
+		<nml_option name="config_Redi_horizontal_ramp_min" type="real" default_value="20e3" units="m"
+					description="Minimum value in grid cell size for Redi $\kappa$ ramp function.  Here cell size refers to dcEdge. Used when config_Redi_horizontal_taper is set to ramp."
+					possible_values="Any positive real value."
+		/>
+		<nml_option name="config_Redi_horizontal_ramp_max" type="real" default_value="30e3" units="m"
+					description="Maximum value in grid cell size for Redi $\kappa$ ramp function.  Here cell size refers to dcEdge. Used when config_Redi_horizontal_taper is set to ramp."
+					possible_values="Any positive real value."
+		/>
 	</nml_record>
 	<nml_record name="GM_eddy_parameterization" mode="forward">
 		<nml_option name="config_use_GM" type="logical" default_value=".false." units="NA"
@@ -356,10 +354,6 @@
 		/>
 		<nml_option name="config_GM_constant_kappa" type="real" default_value="600.0" units="m^2 s^{-1}"
 					description="Coefficient of standard GM parametrization of eddy transport (Bolus component), $\kappa$. Only used when config_GM_closure is set to constant."
-		/>
-		<nml_option name="config_gm_enable_horizontal_resolution_function" type="logical" default_value=".false." units="unitless"
-					description="Flag to enable the resolution tapering in the horizontal following Hallberg (2013) - https://doi.org/10.1016/j.ocemod.2013.08.007"
-					possible_values=".true. or .false."
 		/>
 		<nml_option name="config_GM_constant_bclModeSpeed" type="real" default_value="0.3" units="m/s"
 					description="The parameter setting for the first baroclinic mode speed for the vertical stream function boundary value problem. This appears as $c$ in eqn 16a of Ferrari et al. 2010 (https://doi.org/10.1016/j.ocemod.2010.01.004)."
@@ -404,6 +398,18 @@
 		<nml_option name="config_GM_EG_Rhines_factor" type="real" default_value="0.3" units="NA"
 					description="factor multiplying the Rhines length in the scheme from Eden Greatbatch (2008) Ocean Modeling -- Equation (28)"
 					possible_values="small positive values less than equal to one"
+		/>
+		<nml_option name="config_GM_horizontal_taper" type="character" default_value="ramp" units="unitless"
+					description="Control how the GM Bolus $\kappa$ value varies as a function of horizontal resolution. 'none' is constant, 'ramp' is strictly based on resolution, 'RossbyRadius' follows Hallberg (2013) - https://doi.org/10.1016/j.ocemod.2013.08.007"
+					possible_values="'none', 'ramp', 'RossbyRadius'"
+		/>
+		<nml_option name="config_GM_horizontal_ramp_min" type="real" default_value="20e3" units="m"
+					description="Minimum value in grid cell size for GM $\kappa$ ramp function.  Here cell size refers to dcEdge. Used when config_GM_horizontal_taper is set to ramp."
+					possible_values="Any positive real value."
+		/>
+		<nml_option name="config_GM_horizontal_ramp_max" type="real" default_value="30e3" units="m"
+					description="Maximum value in grid cell size for GM $\kappa$ ramp function.  Here cell size refers to dcEdge. Used when config_GM_horizontal_taper is set to ramp."
+					possible_values="Any positive real value."
 		/>
 	</nml_record>
 	<nml_record name="Rayleigh_damping" mode="forward">
@@ -2702,7 +2708,7 @@
 			packages="forwardMode;analysisMode"
 		/>
 		<var name="betaEdge" type="real" dimensions="nEdges Time" units="m^{-1} s^{-1}"
-			description="meridional gradient of the coriolis parameter"
+			description="meridional gradient of the coriolis parameter, used in GM"
 		/>
 		<var name="gmKappaScaling" type="real" dimensions="nVertLevelsP1 nEdges Time" units="NA" default_value="1.0"
 			description="spatially and depth varying GM kappa.  The scaling is based on the Brunt Vaisala Frequency relative to a maximum value below the mixed layer, follows from Danabasoglu and Marshall 2007.  If config_GM_closure is not set to N2_dependent the scaling value is set to 1 everywhere."
@@ -2861,8 +2867,12 @@
 			 description="Redi Kappa value.  This is constant in time, and set at init based on the constant or ramp fuction."
 			 packages="gm"
 		/>
-		<var name="gmResolutionTaper" type="real" dimensions="nEdges Time" units="m^2 s^{-1}"
-			 description="resolution tapering for GM and Redi"
+		<var name="RediHorizontalTaper" type="real" dimensions="nEdges Time" units="unitless"
+			 description="Horizontal tapering for Redi. Varies between 0 and 1."
+			 packages="gm"
+		/>
+		<var name="gmHorizontalTaper" type="real" dimensions="nEdges Time" units="unitless"
+			 description="Horizontal tapering for GM. Varies between 0 and 1."
 			 packages="gm"
 		/>
 		<var name="surfaceFluxAttenuationCoefficient" type="real" dimensions="nCells Time" units="m"
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -2249,13 +2249,13 @@
 		<var name="kiteAreasOnVertex" type="real" dimensions="vertexDegree nVertices" units="m^2"
 			 description="Area of the portions of each dual cell that are part of each cellsOnVertex."
 		/>
-		<var name="fEdge" type="real" dimensions="nEdges" units="s^{-1}"
+		<var name="fEdge" type="real" dimensions="nEdges" units="radians s^{-1}"
 			 description="Coriolis parameter at edges."
 		/>
-		<var name="fVertex" type="real" dimensions="nVertices" units="s^{-1}"
+		<var name="fVertex" type="real" dimensions="nVertices" units="radians s^{-1}"
 			 description="Coriolis parameter at vertices."
 		/>
-		<var name="fCell" type="real" dimensions="nCells" units="s^{-1}"
+		<var name="fCell" type="real" dimensions="nCells" units="radians s^{-1}"
 			 description="Coriolis parameter at cell centers."
 		/>
 		<var name="bottomDepth" type="real" dimensions="nCells" units="m"
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -2833,10 +2833,6 @@
 			 description="GM stream function"
 			 packages="forwardMode;analysisMode"
 		/>
-		<var name="gmStreamFuncTopOfCell" type="real" dimensions="nVertLevels nCells Time" units="m^2 s^{-1}"
-			 description="GM stream function reconstructed to the cell centers"
-			 packages="forwardMode;analysisMode"
-		/>
 		<var name="GMStreamFuncX" type="real" dimensions="nVertLevelsP1 nCells Time" units="m^2 s^{-1}"
 			 description="GM stream function"
 			 packages="forwardMode;analysisMode"
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -331,7 +331,7 @@
 					possible_values="any integer between 0 (all layers on) and nVertLevels (all layers off)"
 		/>
 		<nml_option name="config_Redi_horizontal_taper" type="character" default_value="ramp" units="unitless"
-					description="Control how the Redi Bolus $\kappa$ value varies as a function of horizontal resolution. 'none' is constant, 'ramp' is strictly based on resolution, 'RossbyRadius' follows Hallberg (2013) - https://doi.org/10.1016/j.ocemod.2013.08.007"
+					description="Control how the Redi $\kappa$ value varies as a function of horizontal resolution. 'none' is constant, 'ramp' is strictly based on resolution, 'RossbyRadius' follows Hallberg (2013) - https://doi.org/10.1016/j.ocemod.2013.08.007"
 					possible_values="'none', 'ramp', 'RossbyRadius'"
 		/>
 		<nml_option name="config_Redi_horizontal_ramp_min" type="real" default_value="20e3" units="m"
@@ -400,7 +400,7 @@
 					possible_values="small positive values less than equal to one"
 		/>
 		<nml_option name="config_GM_horizontal_taper" type="character" default_value="ramp" units="unitless"
-					description="Control how the GM Bolus $\kappa$ value varies as a function of horizontal resolution. 'none' is constant, 'ramp' is strictly based on resolution, 'RossbyRadius' follows Hallberg (2013) - https://doi.org/10.1016/j.ocemod.2013.08.007"
+					description="Control how the GM Bolus value varies as a function of horizontal resolution. 'none' is constant, 'ramp' is strictly based on resolution, 'RossbyRadius' follows Hallberg (2013) - https://doi.org/10.1016/j.ocemod.2013.08.007"
 					possible_values="'none', 'ramp', 'RossbyRadius'"
 		/>
 		<nml_option name="config_GM_horizontal_ramp_min" type="real" default_value="20e3" units="m"
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -357,6 +357,10 @@
 		<nml_option name="config_GM_constant_kappa" type="real" default_value="600.0" units="m^2 s^{-1}"
 					description="Coefficient of standard GM parametrization of eddy transport (Bolus component), $\kappa$. Only used when config_GM_closure is set to constant."
 		/>
+		<nml_option name="config_gm_enable_horizontal_resolution_function" type="logical" default_value=".false." units="unitless"
+					description="Flag to enable the resolution tapering in the horizontal following Hallberg (2013) - https://doi.org/10.1016/j.ocemod.2013.08.007"
+					possible_values=".true. or .false."
+		/>
 		<nml_option name="config_GM_constant_bclModeSpeed" type="real" default_value="0.3" units="m/s"
 					description="The parameter setting for the first baroclinic mode speed for the vertical stream function boundary value problem. This appears as $c$ in eqn 16a of Ferrari et al. 2010 (https://doi.org/10.1016/j.ocemod.2010.01.004)."
 					possible_values="Positive real numbers"
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -295,8 +295,8 @@
 					possible_values=".true. or .false."
 		/>
 		<nml_option name="config_Redi_closure" type="character" default_value="constant" units="unitless"
-					description="Control what type of function is used for Redi $\kappa$."
-					possible_values="'constant', 'data'"
+					description="Control what type of function is used for Redi $\kappa$. For 'equalGM', RediKappa is set to gmBolusKappa, so picks up the closure used by GM."
+					possible_values="'constant', 'equalGM', 'data'"
 		/>
 		<nml_option name="config_Redi_constant_kappa" type="real" default_value="600.0" units="m^2 s^{-1}"
 					description="The Redi diffusion coefficient. Only used when config_Redi_closure = 'constant'."
@@ -2708,7 +2708,7 @@
 			packages="forwardMode;analysisMode"
 		/>
 		<var name="betaEdge" type="real" dimensions="nEdges Time" units="m^{-1} s^{-1}"
-			description="meridional gradient of the coriolis parameter, used in GM"
+			description="meridional gradient of the coriolis parameter, used in the mesoscale eddy parameterization schemes"
 		/>
 		<var name="gmKappaScaling" type="real" dimensions="nVertLevelsP1 nEdges Time" units="NA" default_value="1.0"
 			description="spatially and depth varying GM kappa.  The scaling is based on the Brunt Vaisala Frequency relative to a maximum value below the mixed layer, follows from Danabasoglu and Marshall 2007.  If config_GM_closure is not set to N2_dependent the scaling value is set to 1 everywhere."
--- a/components/mpas-ocean/src/shared/mpas_ocn_diagnostics_variables.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_diagnostics_variables.F
@@ -183,7 +183,7 @@ module ocn_diagnostics_variables
 
    real (kind=RKIND), pointer :: globalVerticalStretchMax, globalVerticalStretchMin
    real (kind=RKIND), dimension(:), pointer :: betaEdge
-   real (kind=RKIND), dimension(:), pointer :: RediHorizontalTaper, gmHorizontalTaper
+   real (kind=RKIND), dimension(:), pointer :: RediHorizontalTaper, gmHorizontalTaper, RossbyRadius
 
    ! Semi-implicit Array Pointers
    real (kind=RKIND), dimension(:), pointer :: SIvec_r0,SIvec_r00,SIvec_r1 ,SIvec_rh0,SIvec_rh1,SIvec_ph0,SIvec_ph1
@@ -260,6 +260,8 @@ contains
                    RediHorizontalTaper)
          call mpas_pool_get_array(diagnosticsPool, 'gmHorizontalTaper', &
                    gmHorizontalTaper)
+         call mpas_pool_get_array(diagnosticsPool, 'RossbyRadius', &
+                   RossbyRadius)
       end if
 
       if ( config_compute_active_tracer_budgets ) then
@@ -740,8 +742,9 @@ contains
          !$acc                   RediKappaScaling,  &
          !$acc                   rediLimiterCount,  &
          !$acc                   RediKappaSfcTaper, &
-         !$acc                   RediHorizontalTaper  &
-         !$acc                   gmHorizontalTaper  &
+         !$acc                   RediHorizontalTaper,  &
+         !$acc                   gmHorizontalTaper,  &
+         !$acc                   RossbyRadius  &
          !$acc                   )
       end if
       if ( config_compute_active_tracer_budgets ) then
@@ -976,8 +979,9 @@ contains
          !$acc                   RediKappaScaling,  &
          !$acc                   rediLimiterCount,  &
          !$acc                   RediKappaSfcTaper, &
-         !$acc                   RediHorizontalTaper  &
-         !$acc                   gmHorizontalTaper  &
+         !$acc                   RediHorizontalTaper,  &
+         !$acc                   gmHorizontalTaper,  &
+         !$acc                   RossbyRadius  &
          !$acc                   )
       end if
       if ( config_compute_active_tracer_budgets ) then
@@ -1170,7 +1174,8 @@ contains
                  rediLimiterCount,  &
                  RediKappaSfcTaper, &
                  RediHorizontalTaper, &
-                 gmHorizontalTaper)
+                 gmHorizontalTaper, &
+                 RossbyRadius)
       end if
       if ( config_compute_active_tracer_budgets ) then
          nullify(activeTracerHorMixTendency,              &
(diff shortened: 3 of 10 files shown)


jonbob:
@darincomeau -- I'll run the scripts that automatically update the bld files to reflect Registry changes for this PR and make sure it's all consistent, unless that's what you've already done.

jonbob:
Thanks for getting back to me so quickly @darincomeau. I'll run the scripts and update the files if anything turns up

jonbob:
@darincomeau - this PR looks like it also includes the changes from [PR #4835](https://github.com/E3SM-Project/E3SM/pull/4835)? That PR has not been merged to the repo yet...

darincomeau:
@jonbob yes, this PR started from the head of that branch in #4835 , to avoid merge conflicts later. I'll rebase once that gets merged.

darincomeau:
5 year tests with `config_gm_enable_horizontal_resolution_function = .true.` compared against baselines with option off:

**ECwISC30to60E2r1**
[This branch with flag on](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/20220331.CRYO1850.ne30pg2_ECwISC30to60E2r1.chrysalis.CryoBranchGMHorResFun/yrs1-5/)
[Baseline](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.milena/20220301.CRYO1850.ne30pg2_ECwISC30to60E2r1.chrysalis.CryoBranchBaseline/Years1-5/)

**SOwISC12to60E2r4**
[This branch with flag on](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.sprice/20220401.CRYO1850.ne30pg2_SOwISC12to60E2r4.chrysalis.CryoBranchGMHorResFun/yrs1-5/)
[Baseline](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.sprice/20220223.CRYO1850.ne30pg2_SOwISC12to60E2r4.chrysalis.CryoBranchBaseline/yrs1-5/)

components/mpas-ocean/src/shared/mpas_ocn_gm.F
```diff
@@ -910,7 +955,37 @@ subroutine ocn_GM_init(domain, err)!{{{
...
+            config_GM_closure == 'EdenGreatbatch'.or. &
+            config_GM_closure == 'edengreatbatch'.or. &
+            config_GM_closure == 'edenGreatbatch'.or. &
+            config_gm_enable_horizontal_resolution_function) then
+            call mpas_pool_get_config(meshPool, 'sphere_radius', sphere_radius)
+            call mpas_pool_get_array(meshPool, 'latEdge', latEdge)
+            !$omp parallel
+            !$omp do schedule(runtime)
+            do iEdge=1,nEdges
+               betaEdge(iEdge) = 2.0_RKIND*omega*cos(latEdge(iEdge)) / sphere_radius
```
milenaveneziani, line 984: since `betaEdge` is already defined at this point, I think we can remove it from the two if statements later in the code, the ones for Visbeck and EdenGreatbatch? (that's something that I had noticed before, looking at Luke's maps_ocn_gm changes).

mark-petersen:
Passes MPAS-O nightly regression suite with `config_gm_enable_horizontal_resolution_function = true` with debug using gnu and intel.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
mark-petersen: Agreed. I'll remove those later two.

vanroekel (approved):
changes look consistent with what I implemented previously so approving on visual inspection and final approval from @jonbob on the supporting changes.

mark-petersen (approved):
Approved, based on my testing and E3SM simulations by others.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
```diff
@@ -663,6 +681,38 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, &
          endif
 
          nEdges = nEdgesArray(3)
+         if(config_gm_enable_horizontal_resolution_function) then
```
milenaveneziani, line 684: @mark-petersen: if we decide to remove the config option above and just add a 'rossbyRadius' option to `config_eddying_resolution_taper`, then the statement above would become:
`if (config_eddying_resolution_taper == 'rossbyRadius') then`

But before going down this line, I'd make sure the change is blessed by @vanroekel.

milenaveneziani, line 684: @mark-petersen: if we decide to remove the config option above and just add a 'rossbyRadius' option to `config_eddying_resolution_taper`, then the statement above would become:
`if (config_eddying_resolution_taper == 'rossbyRadius') then`

But before going down this line, I'd make sure the change is blessed by @vanroekel.

darincomeau:
Rebased after #4835 was merged to master. Removing 'Do not merge' flair.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
```diff
@@ -680,6 +681,38 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, &
...
+               end do
+
+               c_mode1 = max(1.0E-5_RKIND, sumN2/3.141592_RKIND)
+               cGMphaseSpeed(iEdge) = c_mode1
+               resolution = sqrt(c_mode1 / (fEdge(iEdge)**2.0 + betaEdge(iEdge)*c_mode1 + 1.0E-15_RKIND)) &
+                            / dcEdge(iEdge)
+               if (resolution .ge. 2.0 .or. c_mode1 .lt. 0.1_RKIND) then
+                  gmResolutionTaper(iEdge) = 0.0_RKIND
+               else
+                  gmResolutionTaper(iEdge) = 1.0_RKIND
```
mark-petersen, line 709: I don't understand why `gmResolutionTaper` is only 0 or 1 here, not some ramp.

xylar:
@darincomeau, it would be better to rebase and apply your last commit as a fixup on 3748f32f64860a6e1c48f46ea371bddb88a6f969 so it takes those changes out of the history entirely.  Otherwise, you can run into trouble doing future rebases.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
milenaveneziani: That's because there is no transition between GM off and on, when the resolution goes between resolving and not resolving the RR. Luke mentioned a paper by Hallberg that does something similar.

darincomeau:
Ok thanks @xylar , I've gone ahead and done that. Now I think it's all correct

xylar:
No, still seeing the changes as a commit.

xylar:
They're now part of d512e664cf617451afbed2edef5036b27d3ca2c4. Something weird with your rebasing process?

darincomeau:
I undid the rebase, and found the original commit that added the YAKL changes: https://github.com/E3SM-Project/E3SM/pull/4864/commits/cb886cc83cd74b1fd63494884628bcf40442528d . This was something like commit 3 out of 10 for the PR before the rebase. So rather than going back to that commit, amending it, and cherry-picking the rest, I had just done a hard reset of the YAKL hash (first before the rebase, and then now after rebase). But I guess this isn't ideal, and I should just go back and amend the original commit.

xylar:
Yes, an interactive rebase where you move your revert commit right under the offending commit and then choose "fixup" for the revert commit should do what you want.

jonbob:
@darincomeau - I ran the autogenerating scripts and the bld files all look fine to me, consistent with the changes to Registry

darincomeau:
ok thanks @jonbob - I meant to ping you that there may be some other changes coming through on this PR, so that step may need to be done again.

xylar:
@darincomeau, I did a rebase and the fixup you already had worked great for me.  I have:
```
[rebase]
        autosquash = true
```
in my `.gitconfig`.  But even without it as long as you have the fix-up commit directly below the commit you want to fix up (and you put `fixup` in front of it) you should get the same result in an interactive rebase if you take care of this manually rather than relying on autosquash.  In any case, it's taken care of.

The reason I was particularly concerned in this case is because the changes involve a submodule as far as I could tell and that can be more problematic than other changes that get reverted later in the history.

darincomeau:
Thanks for taking care of that @xylar !

components/mpas-ocean/src/Registry.xml
```diff
@@ -2702,7 +2708,7 @@
 			packages="forwardMode;analysisMode"
 		/>
 		<var name="betaEdge" type="real" dimensions="nEdges Time" units="m^{-1} s^{-1}"
-			description="meridional gradient of the coriolis parameter"
+			description="meridional gradient of the coriolis parameter, used in GM"
```
milenaveneziani, line 2711: Is this used only in GM, or in general in the local Rossby radius calculation?

components/mpas-ocean/src/shared/mpas_ocn_diagnostics_variables.F
```diff
@@ -244,7 +244,7 @@ subroutine ocn_diagnostics_variables_init(domain, jenkinsOn, hollandJenkinsOn, e
                    wettingVelocityField)
       end if
 
-      if ( config_use_GM ) then
+      if (config_use_GM.or.config_use_Redi) then
```
milenaveneziani, line 247: Shouldn't this be two separate if blocks, one for GM and one for Redi?

components/mpas-ocean/src/shared/mpas_ocn_gm.F
```diff
@@ -704,13 +707,54 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, &
...
+                  RediHorizontalTaper(iEdge) = RossbyRadiusTaper(iEdge)
+               end do
+               !$omp end do
+               !$omp end parallel
+            end if
+
+            deallocate (RossbyRadiusTaper)
+         end if
+
+         if (config_Redi_closure == 'equalGM') then
```
milenaveneziani, line 741: Need to update the Registry for `config_Redi_closure`. (From what I understand PR #4867 is being included in here)

components/mpas-ocean/src/shared/mpas_ocn_gm.F
```diff
@@ -704,13 +707,54 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, &
...
+            !$omp parallel
+            !$omp do schedule(runtime)
+            do iEdge=1, nEdges
+               RediKappa(iEdge) = RediHorizontalTaper(iEdge) * gmBolusKappa(iEdge)
```
milenaveneziani, line 745: Mark, this is just a comment related to our earlier confusion about Redi=equalGM not picking up the GM-kappa changes with depth. I confirmed with Luke that indeed this is the case for all 3-d GM options (not for Visbeck, because Visbeck is not truly 3-d GM kappa). So, Redi=equalGM is picking up horizontal changes of GM-kappa only. Maybe we should add a comment about this, so not to get confused again in the future.

components/mpas-ocean/src/shared/mpas_ocn_tracer_hmix_redi.F
```diff
@@ -521,49 +528,70 @@ subroutine ocn_tracer_hmix_Redi_init(domain, err)!{{{
...
+                  RediHorizontalTaper(iEdge) = 1.0_RKIND
+               else
+                  RediHorizontalTaper(iEdge) = RediHorizontalTaper(iEdge) &
+                     *(dcEdge(iEdge) - config_Redi_horizontal_ramp_min) &
+                     /(config_Redi_horizontal_ramp_max - config_Redi_horizontal_ramp_min)
+               end if
             end do
             !$omp end do
             !$omp end parallel
-          else if (config_Redi_closure == 'data') then
```
milenaveneziani, line 536: Are we sure we want to remove this option here?

mark-petersen:
PR #4867 required the same init and additional tapering variables as this PR, so I closed #4867 and adding the single flag option (config_Redi_closure = 'equalGM') to this PR.

These flags changed on this PR:
```
< &mesoscale_eddy_parameterizations
<     config_eddying_resolution_taper = 'ramp'
<     config_eddying_resolution_ramp_min = 20e3
<     config_eddying_resolution_ramp_max = 30e3
< /
75a71,73
>     config_Redi_horizontal_taper = 'ramp'
>     config_Redi_horizontal_ramp_min = 20e3
>     config_Redi_horizontal_ramp_max = 30e3
91a90,92
>     config_GM_horizontal_taper = 'ramp'
>     config_GM_horizontal_ramp_min = 20e3
>     config_GM_horizontal_ramp_max = 30e3
```
and also these options were added:
```
config_Redi_closure = 'constant' (default), 'data', 'equalGM' (new)
config_Redi_horizontal_taper = 'ramp' (default), 'contant', 'RossbyRadius' (new)
config_GM_horizontal_taper = 'ramp' (default), 'contant', 'RossbyRadius' (new)
```

components/mpas-ocean/src/shared/mpas_ocn_gm.F
```diff
@@ -680,6 +682,96 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, &
...
+               cell1 = cellsOnEdge(1, iEdge)
+               cell2 = cellsOnEdge(2, iEdge)
+               sumN2 = 0.0
+
+               do k = minLevelEdgeBot(iEdge)+1, maxLevelEdgeTop(iEdge)-1
+
+                  lt1 = 0.5_RKIND*(layerThickness(k,cell1) + layerThickness(k-1,cell1))
+                  lt2 = 0.5_RKIND*(layerThickness(k,cell2) + layerThickness(k-1,cell2))
+                  sumN2 = sumN2 + 0.5_RKIND*(lt1*sqrt(max(BruntVaisalaFreqTop(k,cell1),0.0_RKIND)) + &
+                                  lt2*sqrt(max(BruntVaisalaFreqTop(k,cell2),0.0_RKIND)))
```
mark-petersen, line 702: @vanroekel please check on the `sqrt()` here. Note that we have
```
mpas_ocn_diagnostics.F
1163       coef = -gravity / rho_sw
1179             BruntVaisalaFreqTop(k,iCell) = coef * (displacedDensity(k-1,iCell) - density(k,iCell)) &
1180               / (zMid(k-1,iCell) - zMid(k,iCell))
```
so `BruntVaisalaFreqTop` is actually `N2 = -g/rho0 * drho/dz`. I think you really want sumN2, as implied by the summation variable, not sumN.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
```diff
@@ -680,6 +681,38 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, &
...
+               end do
+
+               c_mode1 = max(1.0E-5_RKIND, sumN2/3.141592_RKIND)
+               cGMphaseSpeed(iEdge) = c_mode1
+               resolution = sqrt(c_mode1 / (fEdge(iEdge)**2.0 + betaEdge(iEdge)*c_mode1 + 1.0E-15_RKIND)) &
+                            / dcEdge(iEdge)
```
mark-petersen, line 722: @vanroekel Based on Hallberg (2013) `L = sqrt( c_g^2/ ( f^2 + 2*beta*c_g) )`, if `c_mode1 = c_g^2` then you need a 2 and a sqrt on the second c_mode1:
```
sqrt(c_mode1 / (fEdge(iEdge)**2.0 + 2*betaEdge(iEdge)*sqrt(c_mode1) + 1.0E-15_RKIND)
```

mark-petersen (approved):

components/mpas-ocean/src/shared/mpas_ocn_gm.F
mark-petersen: Considering how complex this whole calculation is, I will save the Rossby Radius here as an output variable so we can check it.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
milenaveneziani: You want sumN2 to be m/s for 'c_mode1' below to make sense. Just based on that, then the sqrt is correct and maybe the variable name should be changed to `sumN`.

milenaveneziani: something still doesn't add up though, because if we remove the sqrt, then you are left with h*N2, which is m/s^2, not m^2/s^2.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
mark-petersen: Maybe `c_mode1 = c_g` and not `c_mode1 = c_g^2`. Then line 721 would be missing a square, not a square root, and would be 
```
resolution = sqrt(c_mode1^2 / (fEdge(iEdge)**2.0 + 2*betaEdge(iEdge)*c_mode1 + 1.0E-15_RKIND)) / dcEdge(iEdge)
```

milenaveneziani: That is what I was thinking too (that c_mode1=c_g).

components/mpas-ocean/src/shared/mpas_ocn_gm.F
mark-petersen: OK, I get it. We can leave this be. The summation scalar could be called `sum_hN` and then `cGMphaseSpeed` below really is the wave speed in `m/s` (I don't understand the 1/pi - is that a radians conversion). Then it is just line  721 that needs correcting.

milenaveneziani: It looks like a radians conversion, but I don't understand where it came in. Maybe f and beta have radians in them?

mark-petersen: f is radians/sec. I added lengthy comments to the code on the source of these equations, all below, and refer to both Chelton (1998) and Hallberg (2013) with equation numbers.

darincomeau:
5 year tests:

[Baseline, constant GM](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.milena/20220301.CRYO1850.ne30pg2_ECwISC30to60E2r1.chrysalis.CryoBranchBaseline/Years1-5/)

From this PR:
[Horizontal tapering, GM/Redi constant](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/20220415.CRYO1850.ne30pg2_ECwISC30to60E2r1.chrysalis.CryoBranchHorTaperGMRediconstant/yrs1-5/)
```
config_GM_horizontal_taper = 'RossbyRadius'
config_Redi_horizontal_taper = 'RossbyRadius'
```

[GM Visbeck, Redi kappa = GM kappa](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/20220418.CRYO1850.ne30pg2_ECwISC30to60E2r1.chrysalis.CryoBranchRediEqGMVisbeckRetest/yrs1-5/)
```
config_GM_closure = 'Visbeck'
config_Redi_closure = 'equalGM'
```

[Horizontal tapering, GM Visbeck, Redi kappa = GM kappa](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/20220418.CRYO1850.ne30pg2_ECwISC30to60E2r1.chrysalis.CryoBranchHorTaperGMRediEqGM/yrs1-5/)
```
config_GM_horizontal_taper = 'RossbyRadius'
config_Redi_horizontal_taper = 'RossbyRadius'
config_GM_closure = 'Visbeck'
config_Redi_closure = 'equalGM'
```

mark-petersen:
Putting these side-by-side, the `config_Redi_closure = 'equalGM'` (right) produces much more warming in the surface layers than the baseline (left) or 'RossbyRadius' tapering (center). Note difference in vertical scales here, particularly lower right. 
image

darincomeau:
Pinging @jonbob about checking the namelists with your scripts again, since the namelist options have changed, if you wouldn't mind.

milenaveneziani:
I ran similar tests as @darincomeau, but for the SORRM mesh. Two have completed successfully. As soon as the third one is done, I will post some links to the analysis and some plots of GM and Redi kappa.

mark-petersen:
@darincomeau and @jonbob I used the scripts to update the E3SM namelist files, so I expect that everything matches. But it would still be good to double check.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
mark-petersen: Confirmed, we changed this.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
mark-petersen: This is correct. `c_mode1=c_g`

milenaveneziani:
And these are the Redi kappas computed as RediKappa*RediHorTapering in the first 2 cases and just as RediKappa in the third case:

Screen Shot 2022-04-20 at 9 13 08 AM

milenaveneziani:
So, I think this PR is doing what it is supposed to do. 
The only thing I am a bit worried about is the Redi Kappa when the ramp horizontal tapering is applied: not sure I understand those noisy 0-values that are scattered at the equator and in the subtropical North Atlantic. I would have expected simply Redi kappa=GM kappa in this case, since the ramp is applied at init only.

There is also something to say about the abrupt changes in GM kappa when using RR horizontal tapering and a 3d variable GM closure. But it seems that these abrupt changes in kappa (look at the tropics in the pictures above) do not translate in similar changes in the bolus velocity, so perhaps they are less worrisome that one would think.

milenaveneziani:
One last comment: something to consider for the SORRM: if we use HorTapering=RossbyRadius, GM will be *on* for basically all of the Southern Ocean. Only a small number of shelf-areas (perhaps where the stratification is very very low) have a high enough Rossby Radius that GM is not applied there. I have SH focused plots similar to the above that I could share on the side if anybody is interested.

jonbob:
Thanks @darincomeau and @mark-petersen -- I did run the scripts and they found it all perfectly matched the changes to Registry. Thanks!

mark-petersen:
Removed the `config_Redi_closure='equalGM'` option and put it here instead https://github.com/E3SM-Project/E3SM/pull/4904

milenaveneziani:
Here are some figures to diagnose the 'dotting' in RediKappa.
First figure is for the SORRM case with ramp horizontal tapering and GM=Visbeck/Redi=equalGM.
Top row has `gmHorizontalTaper` (note the ramp transition in the SO; also note those random dots in the tropics, probably due to noisy cell resolution changes), `gmBoulsKappa`, and a GM-kappa calculated as `gmBoulsKappa` *`gmHorizontalTaper` *`gmKappaScaling` (the `gmKappaScaling` is 1 everywhere in this case because we are using Visbeck). Bottom row has similar panels as the top ones but they are for Redi. So, note how the horizontal taper just has a binary value now, unlike the GM horizontal taper.

Screen Shot 2022-04-21 at 11 13 47 AM

Also note that `RediKappa` is the same as `RediKappa`*`RediHorizontalTaper` , which means that the `RediKappa` was multiplied times the tapering within the code, unlike the GM kappa. I think we probably want to make things consistent between Redi and GM kappa treatment for this, to avoid future confusion.

milenaveneziani:
Second figure is for the SORRM RossbyRadius-tapering and GM=Visbeck/Redi=equalGM, so similar configuration to the above but with RossbyRadius horizontal taper instead of the ramp. Everything looks good (as expected) in this case:

Screen Shot 2022-04-21 at 11 33 53 AM

milenaveneziani:
@mark-petersen: may I also suggest to add something along this line in the Registry?: 1) for `gmBolusKappa`, add 'this kappa does NOT take into account spatial/temporal changes due to horizontal tapering and vertical scaling'; 2) for `RediKappa`, add 'this kappa takes into account spatial/temporal changes due to horizontal tapering'. So that we know that RediKappa is already multiplied times the horizontal taper within the code, but gmKappa is not.

components/mpas-ocean/src/shared/mpas_ocn_tracer_hmix_redi.F
```diff
@@ -521,49 +529,71 @@ subroutine ocn_tracer_hmix_Redi_init(domain, err)!{{{
...
             call mpas_dmpar_finalize(domain%dminfo)
          end if
 
-         ! Add resolution taper
-         if (config_eddying_resolution_taper == 'none') then
-            ! Nothing to do, as we just keep the same assignment as above.
-         else if (config_eddying_resolution_taper == 'ramp') then
+         ! initialize Redi kappa array
+         if (config_Redi_closure == 'constant'.or. &
+             config_Redi_closure == 'equalGM') then
```
vanroekel, line 574: @mark-petersen I thought 'equalGM' is being moved to a different PR?  It appears here.

vanroekel:
On the last few comments from @milenaveneziani I agree strongly we should be descriptive about RediKappa getting automatically multiplied by the horizontal taper whereas GM does not. Given how the code is written we could do a similar multiplication for GM kappa as the code will zero out the bolus velocity anyway for those cells where the taper is zero.  This may be the least confusing, but would like to hear what others think.

But apart from that minor thing about equalGM and seeing that the fix @mark-petersen pushed fixes the 'dotting' issue this seems ready to go.

vanroekel:
Thinking on my suggestion more, it may be bad to multiply gmBolusKappa by the taper as it would be unnecessary computations.  Adding the Registry lines like @milenaveneziani suggests is probably the best solution in my view

milenaveneziani:
The SORRM test with the ramp where we saw the Redi kappa dotting before has completed. And, with the latest commit, the dotting is gone. See below:

Screen Shot 2022-04-22 at 9 22 19 AM

milenaveneziani:
@mark-petersen: I think, after moving the line of code to #4904 as Luke suggested and adding the comments in the Registry, this PR as well #4904 are ready to be merged.

milenaveneziani:
Also, FYI: I made similar plots for @darincomeau's latest low-res test (with the RampFix) and all looks good: GM and Redi kappas have the same patterns and the two horizontal tapering values are 0 everywhere (as it should be at low-res).

@vanroekel: like you, I thought about changing the calculation of the kappa in the code to make it consistent between GM and Redi. But I was thinking that we wanted to keep things as they are for GM, since the GM bolus velocity has to be weighted by the horizontal tapering anyway. I thought that maybe we could change the Redi kappa by *not* multiplying that times the horizontal taper. But I am not sure: maybe we can think about that in future cleanups. For now, let us just remind ourselves how we do it in the Registry.

darincomeau:
Thanks @milenaveneziani for checking that. I've posted analysis from that run on https://github.com/E3SM-Project/E3SM/pull/4904.

darincomeau:
I've reset the review requests due to the changes since original review.

components/mpas-ocean/src/Registry.xml
```diff
@@ -2860,11 +2860,11 @@
 			 packages="forwardMode;analysisMode"
 		/>
 		<var name="gmBolusKappa" type="real" dimensions="nEdges Time" units="m^2 s^{-1}"
-			 description="GM Bolus Kappa value.  This is constant in time, and set at init based on the constant or ramp fuction."
+			 description="GM Bolus Kappa value.  On output, it has NOT been multiplied by the horizontal taper array gmHorizontalTaper, because that is applied at the end to the normalGMBolusVelocity variable, not to the gmBolusKappa."
 			 packages="gm"
 		/>
 		<var name="RediKappa" type="real" dimensions="nEdges Time" units="m^2 s^{-1}"
-			 description="Redi Kappa value.  This is constant in time, and set at init based on the constant or ramp fuction."
+			 description="Redi Kappa value.  On output, it has already been multiplied by the horizontal taper array RediHorizontalTaper (as opposed to gmBolusKappa, which has not been multiplied by the horizontal taper)."
```
mark-petersen, line 2867: I added descriptive text here so that the multiplication of the taper variable is clear.

components/mpas-ocean/src/shared/mpas_ocn_tracer_hmix_redi.F
mark-petersen: Thanks, I removed that line.

components/mpas-ocean/src/Registry.xml
```diff
@@ -2860,11 +2860,11 @@
 			 packages="forwardMode;analysisMode"
 		/>
 		<var name="gmBolusKappa" type="real" dimensions="nEdges Time" units="m^2 s^{-1}"
-			 description="GM Bolus Kappa value.  This is constant in time, and set at init based on the constant or ramp fuction."
+			 description="GM Bolus Kappa value.  On output, it has NOT been multiplied by the horizontal taper array gmHorizontalTaper, because that is applied at the end to the normalGMBolusVelocity variable, not to the gmBolusKappa."
```
mark-petersen, line 2863: Also added descriptive text here.

milenaveneziani (approved):

mark-petersen (approved):
Made small changes requested above. Retested, passes nightly suite with gnu on badger using
```
+config_GM_closure = 'EdenGreatbatch'
+config_Redi_closure = 'constant'
+config_GM_horizontal_taper = 'RossbyRadius'
+config_Redi_horizontal_taper = 'RossbyRadius'
```
and passes 
```
```

vanroekel (approved):
Looks good to me, but want to note this will likely be subject to the new code review process which will require a 100 year B-case given it is likely a climate changing PR.

milenaveneziani:
@vanroekel: this PR and the next 2 are only climate changing if we change default configuration options (similarly to the first 2 that Mark did). I can definitely see how #6 (KPP changes) will have to go through the new process, but I am not sure about these 3?

vanroekel:
Okay I didn't know these would be default off, if so, that falls under 'stealth feature' which requires the following
```
Demonstrated that it’s possible to determine if feature is on/off via log file / namelists

Ensured that it’s covered by a suitable timer

Verified that it passes the “super-BFB” test suite.   

Documented performance changes via “HPC” test suite with a link to PACE results (For features with potential performance impact)

Performed a component simulation test (when relevant).  Provide E3SM diags output, comparison with baseline maintained by each component group, and determine if results in the acceptable range as documented by each component group.   This will typically be shorter (e.g. a 5 year F compset) simulations used by component developers.   

Performed a B compset, e.g. a 10 year run for a “sanity check”.   Compare with  appropriate “reference B-case” using E3SM diags, and determine if results are in the acceptable range as documented by the simulation teams.  
```

jonbob:
Testing a merge on chrysalis:
PASS ERS.ne11_oQU240.WCYCL1850NS
PASS SMS_D_Ld3.T62_oQU120.CMPASO-IAF
with expected NML DIFFs but
FAIL PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850
with BASELINE master: DIFF

jonbob:
also had a baseline DIFF for:
SMS_D_Ld1.ne30pg2_r05_EC30to60E2r2.WCYCL1850

mark-petersen:
@jonbob on anvil I did the following, and my identical PEM test passes. On 1833a0d7be
```
./create_test PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.anvil_gnu -q acme-medium --walltime 30:00

cat /lcrc/group/e3sm/ac.mpetersen/scratch/anvil/PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.anvil_gnu.20220427_174233_8ir4t2/TestStatus
...
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.anvil_gnu COMPARE_base_modpes
```
Where did you conduct your test, and with what compiler? 
`PEM_Ln9.T62_oQU240.GMPAS-IAF.anvil_gnu` also passes for me, with both default flags and
```
-<config_GM_closure>'EdenGreatbatch'</config_GM_closure>
+<config_GM_closure>'constant'</config_GM_closure>
```

mark-petersen:
`PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.anvil_intel` also passes.

jonbob:
@mark-petersen - I'm testing on chrysalis and getting this for PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel:
```
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel CREATE_NEWCASE
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel XML
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel SETUP
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel SHAREDLIB_BUILD time=312
FAIL PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel NLCOMP
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel MODEL_BUILD time=2054
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel SUBMIT
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel RUN time=143
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel COMPARE_base_modpes
FAIL PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel BASELINE master: DIFF
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel TPUTCOMP
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel MEMLEAK insuffiencient data for memleak test
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel SHORT_TERM_ARCHIVER
```

mark-petersen:
Passes
```
./create_test PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu -q compute --walltime 4:00:00
...
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu RUN time=139
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu COMPARE_base_modpes
```

jonbob:
@mark-petersen - it's not failing the PEM test itself, just not getting the same result that has been blessed on chrysalis

components/mpas-ocean/src/shared/mpas_ocn_gm.F
```diff
@@ -682,6 +684,91 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, &
...
+               ! just after eqn 4 the defines the deformation
+               ! radius as:
+               ! L = sqrt( c_g^2/ ( f^2 + 2*beta*c_g) )
+               !   = c_g * sqrt(1/( f^2 + 2*beta*c_g) )  (note c_g is guaranteed non-negative)
+               ! This comes from a combination of Chelton (1998) eqns 2.3a and
+               ! 2.3b, with a smooth transition in between.
+               RossbyRadius(iEdge) = c_mode1*sqrt( 1.0_RKIND / &
+                  (fEdge(iEdge)*fEdge(iEdge) + 2.0_RKIND*betaEdge(iEdge)*c_mode1 + 1.0E-15_RKIND))
+
+               ! If the RossbyRadius is larger than two cell widths, then this
```
milenaveneziani, line 726: @xylar : here is below where we would have to introduce a transitional/ramp tapering in the `RossbyRadius` horizontal taper.

xylar:
@vanroekel and others that weren't at last Thursday's Cryosphere special topic meeting, I figured it would be helpful to have a little background for @milenaveneziani's comment.

I feel pretty strongly that it doesn't make sense to have GM/Redi turn off abruptly when the Rossby radius is considered to be resolved.  Even when the Rossby radius is only partially resolved, some eddy mixing is occurring and having full GM/Redi is overdoing this mixing. Instead, we need to taper GM/Redi as a function of the ratio of the Rossby radius to mesh resolution.  We will need two new namelist option to set the ratio at which GM/Redi are fully on and fully off (both are is currently hard-coded to 2.0).

I don't think the current implementation makes sense.

xylar (changes requested):
I would either like to see the changes above in this PR or get assurances that this will be a follow-up PR.

vanroekel:
@xylar that is a good point.  The Hallberg recommendation of a step function is based on idealized results (e.g. baroclinic channel type runs) so it is quite possible it is not great for global cases.  Thinking on this, I'd suggest we either merge this in as is (default off) but make an issue that we need to thoroughly evaluate resolution tapering before enabling this option.  Or put this PR on hold and do the investigation sooner.  I don't think either this option (rossby radius) or what is there now are optimal.  The current ramp doesn't account for latitude variability of the rossby radius and this new option doesn't account for the 'gray zone'.  My personal feeling is to leave this PR be and not push it in until we have either good evidence this is an improvement to the simulated climate or we add the ability to blend the ramp + rossby radius as I think @xylar is suggesting.  Please correct me if I'm wrong @xylar

milenaveneziani:
Just to clarify: we are not proposing to remove the ramp option, just to transition the GM/Redi on/off in the RossbyRadius taper from a switch to a more gradual change. Right?
I agree that we could test changes in the SORRM, mostly looking a vertical sections and stratification. And it seems to me that we could do this within this PR, if we put it on hold.

xylar:
An important decision for moving forward on the ramp is which ratio the ramp should be a function of: Rossby radius/cell size or cell size/Rossby radius.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
```diff
@@ -682,6 +684,91 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, &
...
+               ! edge is sufficiently well-resolved, and turn off the
+               ! parameterization.
+               if (RossbyRadius(iEdge).ge.2.0_RKIND*dcEdge(iEdge) .or. c_mode1.lt.0.1_RKIND) then
```
xylar, line 729: @mark-petersen, could you explain where this threshold on `c_model1` came from?  There's no comment about it here.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
```diff
@@ -682,6 +684,91 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, &
...
+               cell1 = cellsOnEdge(1, iEdge)
+               cell2 = cellsOnEdge(2, iEdge)
+               sum_hN = 0.0
+
+               do k = minLevelEdgeBot(iEdge)+1, maxLevelEdgeTop(iEdge)-1
+
+                  lt1 = 0.5_RKIND*(layerThickness(k,cell1) + layerThickness(k-1,cell1))
+                  lt2 = 0.5_RKIND*(layerThickness(k,cell2) + layerThickness(k-1,cell2))
+                  sum_hN = sum_hN + 0.5_RKIND*(lt1*sqrt(max(BruntVaisalaFreqTop(k,cell1),0.0_RKIND)) + &
+                                  lt2*sqrt(max(BruntVaisalaFreqTop(k,cell2),0.0_RKIND)))
```
xylar, line 704: I have to say, it's pretty insane that we have lived for 9 years with a variable named `BruntVaisalaFreqTop` that is actually the Brunt Vaisala frequency *squared*.  That isn't good.

xylar:
Okay, I implemented what I think is a reasonable Rossby-radius ramp here:
https://github.com/xylar/E3SM/commit/37f1f48e1aee7c4720dfc10d724c4e8911dd479f
@darincomeau, are you okay with me pushing this (and the clean-up commit https://github.com/xylar/E3SM/commit/a6090eb2e7dead3256adb113284f2ae379978cd0) to this branch?

@milenaveneziani, I believe you may be best set up for testing the updated branch if @darincomeau is okay with me pushing here.  Is that right?  If so, maybe we can get something going once LCRC is back up?

darincomeau:
Sure @xylar , go ahead and push to this branch.

xylar:
Done!

Note that I have made sure this compiles but haven't done any testing beyond that yet.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
philipwjones: To be fair...in most practical instances, it's always squared since we use it for stability and not to compute an actual frequency. So while it may be important for units, I'm not sure there has been any harm due to a laxness in language.

xylar: It did lead to a confusing conversation on this PR and then quite a bit of confusion when I was looking at the code.  So I think it's just pretty bad for code readability.

xylar: If the variable were `BruntVaisalaFreqSqrTop` instead, it would only be 3 characters longer and would avoid this silly confusion.

philipwjones: Well, I'm not a fan of SelfDescribingVariableNamesThatConsumeAnEntireLineOfCode anyway, so BVSqTop is fine with me :^)

xylar: Yeah, previously I was okay with `SelfDescribingVariableNamesThatConsumeAnEntireLineOfCode` because they're easy to copy and paste.  But they sure don't play nice with short terminal lines and line-length limits.  I'd be happy with `BVSqTop`. At least it's not `N2`...

components/mpas-ocean/src/shared/mpas_ocn_gm.F
```diff
@@ -723,13 +723,18 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, &
                RossbyRadius(iEdge) = c_mode1*sqrt( 1.0_RKIND / &
                   (fEdge(iEdge)*fEdge(iEdge) + 2.0_RKIND*betaEdge(iEdge)*c_mode1 + 1.0E-15_RKIND))
 
-               ! If the RossbyRadius is larger than two cell widths, then this
-               ! edge is sufficiently well-resolved, and turn off the
-               ! parameterization.
-               if (RossbyRadius(iEdge).ge.2.0_RKIND*dcEdge(iEdge) .or. c_mode1.lt.0.1_RKIND) then
+               dc_Lr = dcEdge(iEdge)/RossbyRadius(iEdge)
```
milenaveneziani, line 726: Should we add a small number (1.0E-15_RKIND) to the denominator to avoid possible division by 0?

milenaveneziani:
@xylar: what you did makes sense to me. As a reference, this is a plot that @mark-petersen did and shared with us on slack. It shows the RR from Chelton and the estimate from one of Mark's recent simulations:

image

It seems to me that, for SORRM, the ratio r1 varies from 1.2 and above, which means a tapering factor >=0.28.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
xylar: Good thought but it looks like that shouldn't be necessary. c_mode1 >= 1e-5 so Lr > 0 always.

components/mpas-ocean/bld/build-namelist
```diff
@@ -591,6 +586,9 @@ add_default($nl, 'config_GM_EG_riMin');
 add_default($nl, 'config_GM_EG_kappa_factor');
 add_default($nl, 'config_GM_EG_Rossby_factor');
 add_default($nl, 'config_GM_EG_Rhines_factor');
+add_default($nl, 'config_GM_horizontal_taper');
+add_default($nl, 'config_GM_horizontal_ramp_min');
+add_default($nl, 'config_GM_horizontal_ramp_max');
```
milenaveneziani, line 591: this file and the next 3, I believe, should also contain the new namelist options for `GM_Rossby_ramp`

xylar: That didn't come in with the automated tool...

xylar:
@darincomeau, apologies for the force-push.  I know it's not kosher but I also didn't want a messy unnecessary commit for a step I forgot.

mark-petersen:
I tested with the head of this PR, and it still fails comparison with master with the following tests:
```
PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel
PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu
SMS_Ln9.T62_EC30to60E2r2.GMPAS-IAF.chrysalis_intel
SMS_Ln9.T62_EC30to60E2r2.GMPAS-IAF.chrysalis_gnu
```
It passes the other parts, including the PEM tests, i.e.
```
cat /lcrc/group/e3sm/ac.mpetersen/scratch/chrys/PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu.C.20220513_081121_5hlnz7/TestStatus
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu CREATE_NEWCASE
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu XML
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu SETUP
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu SHAREDLIB_BUILD time=195
FAIL PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu NLCOMP
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu MODEL_BUILD time=510
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu SUBMIT
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu RUN time=140
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu COMPARE_base_modpes
FAIL PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu BASELINE master_220514_4435bfa0e1: DIFF
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu TPUTCOMP
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu MEMLEAK insuffiencient data for memleak test
PASS PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu SHORT_TERM_ARCHIVER
```
@jonbob do you know what NLCOMP is?

This one does pass the compare with master:
```
SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.chrysalis_gnu
```

I will look into this next week. For reference, these are the commands I use for these tests. From master 996c7e7aa3
```
cd cime/scripts
./create_test PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu \
   --project e3sm \
   --baseline-root /lcrc/group/e3sm/ac.mpetersen/scratch/chrys/ \
   -b master_220514_4435bfa0e1 \
   -g
```
Then from the PR branch I run the identical command but with `-c` rather than `-g`:
```
cd cime/scripts
./create_test PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu \
   --project e3sm \
   --baseline-root /lcrc/group/e3sm/ac.mpetersen/scratch/chrys/ \
   -b master_220514_4435bfa0e1 \
   -c
```

xylar:
@milenaveneziani, I would guess that `NLCOMP` is comparing namelists with a baseline.  Any changes have to be "blessed" and the comparison will fail until then.  @jonbob, correct me if that's not correct.

It's the comparison with the baseline that is quite problematic.  This PR is not currently BFB.

jonbob:
@xylar is exactly right -- the NLCOMP failure just means there has been a namelist change associated with the PR. It is one of the aspects we track and then have to submit requests to "bless" any changes. Please note that this PR already has the NML label and also has that tag included in its description

components/mpas-ocean/src/shared/mpas_ocn_gm.F
```diff
@@ -682,6 +684,96 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, &
...
+               !   = c_g * sqrt(1/( f^2 + 2*beta*c_g) )  (note c_g is guaranteed non-negative)
+               ! This comes from a combination of Chelton (1998) eqns 2.3a and
+               ! 2.3b, with a smooth transition in between.
+               RossbyRadius(iEdge) = c_mode1*sqrt( 1.0_RKIND / &
+                  (fEdge(iEdge)*fEdge(iEdge) + 2.0_RKIND*betaEdge(iEdge)*c_mode1 + 1.0E-15_RKIND))
+
+               dc_Lr = dcEdge(iEdge)/RossbyRadius(iEdge)
+
+               ! Ramp the parameterization between the given min and max values
+               ! of the ratio of the cell width to the Rossby radius
```
milenaveneziani, line 729: Following my discussions with @xylar and @vanroekel on slack, we agreed that the check on `c_mode1` in the if statement below is not appropriate. If we really wanted to include it, then it should lead to a full-on GM (`RossbyRadiusTaper=1`). So I tried two tests: 1) I removed the check completely, and 2) I changed the first part of the if statement as follows:
```
if (dc_Lr >= config_GM_Rossby_ramp_max .or. c_mode1.lt.0.1_RKIND) then
       RossbyRadiusTaper(iEdge) = 1.0_RKIND
else if (dc_Lr <= config_GM_Rossby_ramp_min ) then
       RossbyRadiusTaper(iEdge) = 0.0_RKIND 
```

In both cases, the ocean died at the very first time step because of nans in the state variables. I don't understand why this happens, since this only involves a check modification.

vanroekel: @milenaveneziani could you let me know what your redi settings are for your tests?  Also is this E&G GM?

milenaveneziani: Thanks Luke. I am using default Redi and GM Visbeck for these.

vanroekel: These are B-cases right?  So constant 400 for Redi?  I was thinking perhaps the instability was in Redi if you were using equal GM

milenaveneziani: yes, B-cases. and yes, Redi is constant at 400. We just wanted to test the GM changes for this one.

vanroekel: Thanks for confirming that.  One other question, just so I understand, does the code run stably if it is like on this PR, where the c_mode1 check is in the wrong place?  And if you remove that incorrect check it crashes?

vanroekel: Also does the model run without the rossby radius taper option on?

milenaveneziani: exactly. I was able to run fine for 20 years with the new Rossby Radius ramp represented by the if statement with the c_mode1 check in the wrong place.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
```diff
@@ -682,6 +684,96 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, &
...
+               end do
+
+               ! See Chelton (1998)
+               ! https://journals.ametsoc.org/view/journals/phoc/28/3/1520-0485_1998_028_0433_gvotfb_2.0.co_2.xml
+               ! eqn 2.2a, also Appendix A,
+               ! for the first-mode (m=1) baroclinic gravity wave speed in m/s
+               ! c_m = 1/pi integral(N dz )
+
+               c_mode1 = max(1.0E-5_RKIND, sum_hN/3.141592_RKIND)
+               cGMphaseSpeed(iEdge) = c_mode1
```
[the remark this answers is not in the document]
milenaveneziani: To me, it makes more sense to remove it.

milenaveneziani:
Update after additional conversations with @vanroekel: I have now removed the line with the re-definition of `cGMphaseSpeed` and the check on `c_mode1` from the Rossby Radius taper code block, and the code is now stable. It has been running for a couple of simulation months now.

Note that we believe that the line with `cGMphaseSpeed` was likely also causing the BFB fails. So hopefully removing that will also solve that problem.

milenaveneziani:
If this fix is satisfactory, the only remaining question is whether we want to have an additional output variable that is `c_mode1` (first baroclinic gravity phase speed) because of course the variable `cGMphaseSpeed` is something slightly different in its original definition.

xylar:
> If this fix is satisfactory, the only remaining question is whether we want to have an additional output variable that is c_mode1 (first baroclinic gravity phase speed) because of course the variable cGMphaseSpeed is something slightly different in its original definition.

I don't think we want to add another variable for this, given that it's not yet clear if we want to use this particular capability.  It would be a good idea to add an output variable temporarily for testing but not in the PR.

mark-petersen:
Just an update on the non-BFB problem - I tried to change the order of operations back on the RediKappa and gmHorizontalTaper calculations, but that did not get rid of the non-BFB behavior in this PR. I will keep trying.

milenaveneziani:
We also need to adjust the description of the taper to add that it is not just applied to GM, but to Redi as well (I requested this change somewhere in this PR).

mark-petersen:
@milenaveneziani I think this is the commentary you asked for before
https://github.com/E3SM-Project/E3SM/pull/4864/files#diff-39f42f6faac15040cd865ccbe070b85a57b5821fe2480f4386d1bbec6214b2d8R2875
https://github.com/E3SM-Project/E3SM/pull/4864/files#diff-39f42f6faac15040cd865ccbe070b85a57b5821fe2480f4386d1bbec6214b2d8R2871

But what you said is _NOT_ true - GM and Redi tapers have separate flags and variables (RediHorizontalTaper and GMHorizontalTaper) now.

Yes, please go ahead and push whatever changes you had in mind on those, so we can have the exact code that you ran with. Thanks.

milenaveneziani:
@mark-petersen: even though GM and Redi have separate tapers, the config options `config_GM_Rossby_ramp_min` and `config_GM_Rossby_ramp_max` are used to define *both* tapers, so even though there is the 'GM' in there, they are used for both GM and Redi.

mark-petersen:
I thought that was too confusing. I have both `config_Redi_horizontal_ramp_max` and `config_GM_horizontal_ramp_max` etc. They would normally be set the same, but I thought it made more sense to separate them. Do you agree?

milenaveneziani:
So, those are separate. They apply to the ramp taper.
Instead the Rossby_ramp min/max values apply to the transition from GM fully on to GM fully off in the RossbyRadius taper. And these are *not* separate.
I know it's a bit confusing. Do you have other suggestions?

vanroekel:
for the rossby taper, what about `config_GMRedi_Rossby_ramp_min/max`?

milenaveneziani:
That works for me.

xylar:
@milenaveneziani, could you make that name change, too? (It needs to happen in various places.)

milenaveneziani:
@xylar: sure. Is that OK if I do it manually though?

xylar:
Yeah, do it manually.  @jonbob will check our work in any case.

components/mpas-ocean/src/shared/mpas_ocn_tracer_hmix_redi.F
```diff
@@ -543,27 +554,42 @@ subroutine ocn_tracer_hmix_Redi_init(domain, err)!{{{
...
+                  RediHorizontalTaper(iEdge) = 0.0_RKIND
+               else if (dcEdge(iEdge) >= config_Redi_horizontal_ramp_max) then
                   ! nothing to do, i.e. RediKappa(iCell) = RediKappa(iCell)
+                  RediHorizontalTaper(iEdge) = 1.0_RKIND
                else
                   RediKappa(iEdge) = RediKappa(iEdge) &
-                                     *(dcEdge(iEdge) - config_eddying_resolution_ramp_min) &
-                                     /(config_eddying_resolution_ramp_max - config_eddying_resolution_ramp_min)
+                                     *(dcEdge(iEdge) - config_Redi_horizontal_ramp_min) &
+                                     /(config_Redi_horizontal_ramp_max - config_Redi_horizontal_ramp_min)
```
mark-petersen, line 584: This fixed the BFB mismatch. I had to put this block of code in the same order as before, with the operations in the same order as before. Now it is BFB in E3SM tests (it was BFB in stand-alone tests before, which is partly what took me so long to find it).

milenaveneziani: yay, @mark-petersen, that's great!

mark-petersen:
Now passes compare against branch point 4435bfa0e1 for:
```
SMS_Ln9.T62_EC30to60E2r2.GMPAS-IAF.chrysalis_gnu
SMS_Ln9.T62_EC30to60E2r2.GMPAS-IAF.chrysalis_intel
PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu
```
@jonbob please proceed with your tests.

mark-petersen:
I also tested stand-alone and E3SM with the new flag settings for this PR
```
config_GM_horizontal_taper = 'RossbyRadius'
config_Redi_horizontal_taper = 'RossbyRadius'
```
in E3SM, passes
```
PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_gnu
```
in stand-alone nightly suite with all three: `config_GM_closure = 'N2_dependent', Visbeck, EdenGreatbatch` with gnu debug, and all pass.

mark-petersen:
Any other comments on this one, or is it ready for final testing before the merge? @xylar you still have a red on the approval, can you approve if you agree?

milenaveneziani:
@mark-petersen, as I mentioned (somewhere) above, my latest test is behaving as expecting. The taper looks consistent with the Rossby Radius size, and everything worked without further problems.
So, yes, this can be merged from my point of view (double green check).

xylar (approved):
@milenaveneziani did a ton of work to address my concerns so thanks for the coding and testing!

vanroekel:
I'm also happy with all the changes.  Thanks for the great discussion and thorough testing.  I'll give you my second green check!

jonbob:
The automated scripts found no discrepancies in the bld files

jonbob:
Passes:
* SMS_D_Ld1.ne30pg2_r05_EC30to60E2r2.WCYCL1850.chrysalis_intel

with expected NML DIFFs.

Merged to next

xylar:
Thanks @jonbob.  Assuming all goes well, I'm sure everyone in E3SM land will be happy to have this thread silenced (at only 155 comments).

darincomeau:
@jonbob yes, I agree. The `config_eddying_resolution_taper` namelist option is removed in this PR, so that should be replaced as you suggested.

jonbob:
re-merged to next with testmods update

jonbob:

PR #4873, opened 2022-04-06
Update modules for ANL GCE nodes

dqwu:
Most GCE nodes have been upgraded from Ubuntu 18.04 to 20.04 and
we need to update some modules and software libs accordingly to run
E3SM on these nodes.

A new machine name anlgce-ub18 is created to support two legacy
GCE nodes (compute-01.cels.anl.gov and compute-02.cels.anl.gov)
that still use Ubuntu 18.04.

Open MPI support is also added for ANL GCE nodes in this PR.

[BFB]


cime_config/machines/config_machines.xml
```diff
@@ -1217,6 +1223,102 @@
...
+      <env name="OMP_STACKSIZE">64M</env>
+    </environment_variables>
+    <environment_variables>
+      <env name="PERL5LIB">/nfs/gce/projects/climate/software/perl5/lib/perl5</env>
+    </environment_variables>
+  </machine>
+
+  <machine MACH="anlgce">
+    <DESC>ANL CELS General Computing Environment (Linux) workstation</DESC>
+    <NODENAME_REGEX>compute-(240|386)-[0-9][0-9]</NODENAME_REGEX>
```
jayeshkrishna, line 1245: Does this mean that the user needs to explicitly specify the machine name (create_newcase etc) for ubuntu 18 machines?

dqwu: No. On compute-01 and compute-02 the machine name will be identified as anlgce-ub18 and on other GCE nodes the machine name will be identified as anlgce.

dqwu: This is because anlgce-ub18 is defined before anlgce in the machine file and "compute-386-01|compute-386-02" will be matched first.

jayeshkrishna: ok, thanks. Verified that it works with compute-01 without specifying the machine name when creating the case.

jayeshkrishna (approved):

PR #4877, opened 2022-04-07
Update CIME submodule

### LINKED ISSUE #4875: testing bless tool broken?
For several recent PRs that introduced NML only DIFFs, the blessing of those tests caused the baselines to seemingly disappear and cause real DIFFs. This issue is keeping the repo closed and is critical. The error look like:
````
Comparing hists for case 'ERS_Ld5.T62_oQU120.CMPASO-NYF.chrysalis_intel.C.JNextIntegration20220406_223454' dir1='/lcrc/group/e3sm/e3smtest/scratch/chrys/ERS_Ld5.T62_oQU120.CMPASO-NYF.chrysalis_intel.C.JNextIntegration20220406_223454/run', suffix1='',  dir2='/lcrc/group/e3sm/baselines/chrys/intel/master/ERS_Ld5.T62_oQU120.CMPASO-NYF.chrysalis_intel' suffix2=''
````
and causes this for the overall test result:
````
    File 'ERS_Ld5.T62_oQU120.CMPASO-NYF.chrysalis_intel.C.JNextIntegration20220406_223454.cpl.hi.0001-01-06-00000.nc' had no compare counterpart in '/lcrc/group/e3sm/baselines/chrys/intel/master/ERS_Ld5.T62_oQU120.CMPASO-NYF.chrysalis_intel' with suffix ''
Did not compare any hist files! Missing baselines?
FAIL
````

### PR DESCRIPTION
**Pull request by jgfouca**
Update CIME submodule
    
... to 524842660578d7b8f72df78a77bd302f9d128b44
    
Changes:
1) Make sure CIME machine matches probed machine
2) Update neon data server for input files
3) New SystemTestsCompareN test type
4) Extend configure to get cmake macros files from the .cime directory
5) BIG: restructure CIME to match what would be expected in a standard python pacakge!
6) wait_for_tests: Try multiple ctest drop methods
7) Allow special XML syntax for init_path and cmd_path
8) Add Total_Build time to end of build_times.txt
    
Fixes:
1) Fix the path for PTS_DOMAIN_FILE
2) Fix config_machines xsd to allow more compilers
3) Fix xmlchange tests
4) Remove Machines warning
5) Using --retry in create_newcase was causing generated baseline to be removed
6) Fix run_tests.py when no machine is given
7) Fixes for machine check
8) Fixes issue with model detection on E3SM machines.

Fixes #4875 
    
[BFB]

### DIFF
--- a/components/CMakeLists.txt
+++ b/components/CMakeLists.txt
@@ -79,7 +79,7 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/..)
 
 # Set global targets
 add_custom_target(genf90
-  DEPENDS ${CIMEROOT}/src/externals/genf90/genf90.pl)
+  DEPENDS ${CIMEROOT}/CIME/non_py/externals/genf90/genf90.pl)
 
 # Build E3SM components
 set(IDX 0)
--- a/components/CMakeLists.txt
+++ b/components/CMakeLists.txt
@@ -2,7 +2,7 @@
 #
 # Common CMakeLists.txt: a framework for building all CIME components and more
 #
-# This is a port of cime/scripts/Tools/Makefile. As more components are ported to
+# This is a port of cime/CIME/Tools/Makefile. As more components are ported to
 # CMake, the directory level of this file will rise to the top-level directory.
 #
 # We will prefer space-separated strings over lists
--- a/.circleci/checkout.sh
+++ b/.circleci/checkout.sh
@@ -2,4 +2,5 @@
 
 # Replace all ssh URLs to submodules with HTTP URLs
 sed -i 's/git@github.com:/https:\/\/github.com\//' .gitmodules
-git submodule update --init
+sed -i 's/git@github.com:/https:\/\/github.com\//' cime/.gitmodules
+git submodule update --init --recursive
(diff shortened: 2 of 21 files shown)

### CODE REVIEW

File: components/CMakeLists.txt [hunk 1]
```diff
@@ -79,7 +79,7 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/..)
 
 # Set global targets
 add_custom_target(genf90
-  DEPENDS ${CIMEROOT}/src/externals/genf90/genf90.pl)
+  DEPENDS ${CIMEROOT}/CIME/non_py/externals/genf90/genf90.pl)
```
rljacob, line 82: Is this where the dependence on CIME genf90 comes in to the E3SM build?  Lets just our own in E3SM/externals.

jgfouca: @rljacob , yes, that's one of the places. CIME also uses it internally in the important Tools/Makefile. Since the problem is in a shared lib, it's the Makefile one that's the problem.

jgfouca:
Pushed to next.

rljacob:
Does this version have the http submodule and incompatible genf90 ?

jgfouca:
@rljacob , no, I made a special branch on CIME, `jgfouca/e3sm` or something like that, and made sure it had a version of genf90 in its genf90 submodule that works for us. I also changed .gitmodules to use ssh. Hopefully, by the time of the next CIME update, this issue will have been resolved and we can go back to pointing to CIME master.

rljacob:
The CircleCI build fail should be looked at.   There are also build fails with the homme_integration tests.

rljacob:
This is the CircleCi message:  "/home/circleci/project/cime/CIME/non_py/externals/genf90/genf90.pl: not found"

jgfouca:
@rljacob , thanks. I pushed fixes for both the homme and circle-ci problems, which were indeed both related to the CIME update.

PR #4879, opened 2022-04-09
Update OpenMP-offload tests

amametjanov:
Update OpenMP-offload tests:
- remove `SMS_Ln9.ne4pg2_ne4pg2.F-MMFOMP.eam-single_thread` from `e3sm_gpuomp`
- adjust Miller `NODENAME_REGEX` to avoid login node name conflicts

[BFB]

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -60,7 +60,6 @@
 
   <machine MACH="miller">
     <DESC> ORNL AF cluster 800-node AMD Epyc 2-sockets 64-cores per node</DESC>
-    <NODENAME_REGEX>login.*</NODENAME_REGEX>
     <OS>CNL</OS>
     <COMPILERS>gnu,cray,intel</COMPILERS>
     <MPILIBS>mpt</MPILIBS>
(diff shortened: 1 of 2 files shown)


cime_config/machines/config_machines.xml
```diff
@@ -60,7 +60,6 @@
 
   <machine MACH="miller">
     <DESC> ORNL AF cluster 800-node AMD Epyc 2-sockets 64-cores per node</DESC>
-    <NODENAME_REGEX>login.*</NODENAME_REGEX>
```
amametjanov, line 63: @minxu74 this forces everyone on a node that starts with `login` to specify `--machine ` argument to create_[newcase,test]. Maybe it can be adjusted to a miller-specific string. Removing for now.

cime_config/tests.py
```diff
@@ -436,7 +436,6 @@
 
     "e3sm_gpuomp" : {
         "tests"    : (
-                 "SMS_Ln9.ne4pg2_ne4pg2.F-MMFOMP.eam-single_thread",
```
amametjanov, line 439: @xyuan as discussed on Slack

rljacob: Why is it being removed?

amametjanov: It's been broken since Feb 7 with no plans to fix. When fixed, this can be reverted.

PR #4885, opened 2022-04-12
Initial implementation of BGC carbon budget in the coupler

jonbob:
Brings in a new coupler carbon budget which is implemented similarly to the current heat and water budgets, but in a separate source file. The output is intermingled with that from the other budgets, so that the monthly budget tables are all output before the annual ones, for example. The BGC budget is controlled overall by the BUDGETS flag found in env_run.xml but also has its own "do_bgc_budgets" flag that is included in the driver seq_infodata_inparm namelist and can be set by user_nl_cpl. By default it is off unless the compset includes BGC fluxes, in which case it is set to true.

[NML]
[BFB]

### DIFF
--- a/driver-mct/cime_config/namelist_definition_drv.xml
+++ b/driver-mct/cime_config/namelist_definition_drv.xml
@@ -1019,6 +1019,20 @@
     </values>
   </entry>
 
+  <entry id="do_bgc_budgets">
+    <type>logical</type>
+    <category>budget</category>
+    <group>seq_infodata_inparm</group>
+    <desc>
+      logical that turns on BGC diagnostic budgets, false means BGC budgets will never be written
+    </desc>
+    <values>
+      <value>.false.</value>
+      <value BGC_MODE="CO2.*">.true.</value>
+      <value BGC_MODE=".*_OI">.true.</value>
+    </values>
+  </entry>
+
   <entry id="do_histinit">
     <type>logical</type>
     <category>history</category>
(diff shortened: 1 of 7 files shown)


jonbob:
Test merges passes:
with expected NML DIFF.

Waiting on:

rljacob (approved):

susburrows (approved):
I just did a quick visual inspection of the code changes.  Thanks for putting this together, Jon.

I saw that you also added a central definition for the molecular weights of carbon, oxygen and CO2 here, which is great.  We should make sure that the rest of the BGC group is aware of this addition, so that folks can be referencing this central definition when they work on other parts of the code that use those constants.

jonbob:
Thanks for the reviews. The only issue I'm aware of is that it has restart issues but only with one term in the budget. That's confusing to me, but maybe points to a sequencing problem  with when each part of the budget is calculated. I'll keep working on it, but we could also merge this as-is and create an issue for the restart being non-bfb

susburrows:
Jon, that sounds like a good approach to me if @rljacob is ok with it.  It would be nice to have this PR merged soon so that it can be part of the v2.1 tag.

rljacob:
Yes that's fine with me.

jonbob:

susburrows:
Thanks Jon!

jonbob:
@susburrows - anytime. I'll work on the restart issue next

jonbob:

PR #4886, opened 2022-04-13
Add default BGC pelayout on Anvil

amametjanov:
Add default 30-node BGC pelayout on Anvil. Also disambiguate pelayouts for
- WCYCL and BGC compsets
- BGC compset with MPASO and DOCN components

[BFB]

### DIFF
--- a/cime_config/allactive/config_pesall.xml
+++ b/cime_config/allactive/config_pesall.xml
@@ -1528,43 +1528,6 @@
       </pes>
     </mach>
   </grid>
-  <grid name="a%ne30np4.pg2_l%r05_oi%EC30to60E2r2_r%null_g%null_w%null_z%null_m%EC30to60E2r2">
-    <mach name="anvil|bebop">
-      <pes compset="EAM." pesize="any">
-        <comment>pelayout for tri-grid tests with EAM</comment>
-        <ntasks>
-          <ntasks_atm>-8</ntasks_atm>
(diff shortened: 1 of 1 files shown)


cime_config/allactive/config_pesall.xml
```diff
@@ -7087,6 +7050,66 @@
           <rootpe_cpl>0</rootpe_cpl>
         </rootpe>
       </pes>
+      <pes compset=".*EAM.+ELM.+MPASSI.+MPASO.+SGLC_SWAV_SIAC_SESP_BGC.*" pesize="M">
+        <comment>anvil: --compset BGC* --res ne30pg2_r05_EC30to60E2r2 on 30 nodes pure-MPI, ~3 sypd </comment>
```
jonbob, line 7054: @amametjanov - the PR says it's setting up a small layout but it's specified M here?

amametjanov: It's small in the sense of smallest Anvil node-count that can run BGC (for test defaults), not in the sense of `pesize="S"` small. I'll change to "default".

jonbob: that sounds great

jonbob (approved):
Approved based on testing that shows this PR does allow the system to differentiate between layouts for WCYCL1850 and BGC compsets

PR #4880, opened 2022-04-11
Add and update SSP370 and SSP585 to maint-2.0

wlin7:
Added SSP370 compset "WCYCLSSP370" and updated existing compset "WCYCLSSP585". For SSP370, the new "WCYCLSSP370" compset was defined, new EAM and ELM use case files were created, new prescribed forcing files were created, and minor changes elsewhere were added by emulating what was done for SSP585. For the updated SSP585, the "WCYCLSSP585" compset was defined, new land use and land cover change files were created (these files are grid-specific, and the default atm_lnd grid changed from v1 to v2), and the ELM use case file was updated to point to these new files. The atmosphere forcing files set in the SSP585 use case were not changed because they are not grid specific and they already existed from v1. All SSP370 EAM and ELM input files were staged in NERSC-Cori's inputdata directory only.

PR #4887, opened 2022-04-13
Added SSP370/585 tests and updated use_case files

jjbenedict:
Tests are added for SSP370 and SSP585. 
Both are added for e3sm_prod tests, only SSP370 is added for e3sm_integration
as the two uses the same testmods settings (hybrid run from historical).

Changes are also added to ELM's SSP585 use case file to update fsurdat and 
flanduse_timeseries entries to newest file versions that use the ne30.pg2 grid.

[BFB] no impact on existing tests. New baselines will be needed after this PR.


rljacob (approved):

components/elm/bld/namelist_files/use_cases/2015-2100_SSP585_transient.xml
```diff
@@ -32,8 +32,8 @@
 
 <!-- CMIP6 DECK compsets -->
 
-<fsurdat>lnd/clm2/surfdata_map/surfdata_ne30np4_simyr1850_2015_c171018.nc </fsurdat>
-<flanduse_timeseries>lnd/clm2/surfdata_map/landuse.timeseries_ne30np4_ssp5_rcp8.5_simyr2015-2100_c190411.nc</flanduse_timeseries>
+<fsurdat>lnd/clm2/surfdata_map/surfdata_ne30np4.pg2_simyr1850_2015_c211105.nc </fsurdat>
```
wlin7, line 35: In theory, should a simyr2015 be used for this compset (and associated use_case) that to start from 2015-01-01?

@bishtgautam , can you please also chime in?
For a hybrid SSP585 simulation (or SSP370 in a similar term) starting  using v2.LR.historical run from 2015-01-01, we will need to set `CHECK_FINIDAT_FSURDAT_CONSISTENCY       = .false.` anyway; and either fsurdat should have no impact on simulation results.

bishtgautam: I believe the entries in all the `.xml` files in this directory should not include any resolution-dependent values (e.g. `fsurdat`, `finidat`, `flanduse_timeseries` etc). The resolution-dependent entries should be added in `components/elm/bld/namelist_files/namelist_defaults.xml`.

wlin7: Thank you @bishtgautam  for chiming in. I don't disagree that resolution-dependent values can be consolidated into namelist_defaults.xml, as long as they can be unambiguously determined.  Here fsurdat and flanduse_timeseries values depend on compset. I am not sure if namelist_defaults.xml can recognize compset attribute, and don't see any in existing defaults.

But that can be saved for another day. Can you comment if we should use a simyr2015 surfdata in this case? Esp. when not using a hybrid run nor setting `CHECK_FINIDAT_FSURDAT_CONSISTENCY = .false.`?

wlin7:
Note that we may still make some update to the use_case files. The tests part will have no further changes and the tests have been verified to go as intended as a hybrid run, with the specific REFCASE and REFDATE,

components/elm/bld/namelist_files/use_cases/2015-2100_SSP585_transient.xml
rljacob: Yes they can, an example from eam  1850_CMIP6_control.xml
`<fsurdat hgrid="ne30np4" mask="oEC60to30v3" >
lnd/clm2/surfdata_map/surfdata_ne30np4_simyr1850_2015_c171018.nc </fsurdat>`

bishtgautam: >But that can be saved for another day.

I'm okay with adding these res-dependent values here for now, as long as, we all recognize that this compset will only correctly work with `-res ne30pg2_EC30to60E2r2`. I will defer to @rljacob if he thinks this should be fixed in the PR or in a future PR.

>Can you comment if we should use a simyr2015 surfdata in this case? Esp. when not using a hybrid run nor setting `CHECK_FINIDAT_FSURDAT_CONSISTENCY = .false.`?

I do believe if one does not use `CHECK_FINIDAT_FSURDAT_CONSISTENCY = .false.`, the fsurdat should correspond to simyr2015.

wlin7: @bishtgautam , thanks for confirming simyr2015 fsurdat should be used in general.
 
 @rljacob , I agreed variables like fsurdat, finidat can be specified in namelist_defaults.xml, using grid, mask, amd simyr modifiers. Some compset specific variables, like flanduse_timeseries, I am afraid can not be put there. For example, we have SSP370 and SSP585, they share all ( as far as lnd is concerned) except for flanduse_timeseries.

rljacob: Are we talking about the same thing?  I just searching in elm/bld/namelist_files/use_cases and find in 20thC_CMIP6_transient.xml:
`<flanduse_timeseries hgrid="ne30np4">lnd/clm2/surfdata_map/landuse.timeseries_ne30np4_hist_simyr1850_2015_c20171018.nc</flanduse_timeseries>`

That are plenty of use case that don't have an hgrid modifier for that entry but should.

rljacob: SSP370 would be a new use_case file, 2015-2100_SSP370_transient, if we stick with the current system.  Otherwise, the perl has be modified so that "ssp" can be a modifier these xml files.

wlin7: Exactly. The use_case file for SSP370 is in the master, via #4880.

cime_config/tests.py
```diff
@@ -244,6 +245,8 @@
         "tests"   : (
             "SMS_Ld1.ne30pg2_r05_EC30to60E2r2.WCYCL1850.allactive-wcprod",
             "SMS_Ld1.ne30pg2_EC30to60E2r2.WCYCL1850.allactive-wcprod",
+            "SMS_D_Ld1.ne30pg2_EC30to60E2r2.WCYCLSSP370.allactive-wcprodssp",
```
singhbalwinder, line 248: Why do we have the same test `SMS_D_Ld1.ne30pg2_EC30to60E2r2.WCYCLSSP370.allactive-wcprodssp` in both the production and the integration suites? Is it to ensure the machine coverage for production environment? May be we can change this test to make it more like "production style":
1. Remove `_D` to make it non-debug
2. Increase the number to days or remove `_Ld1` to run it for 5 days (default)

rljacob: The production suite is run on more platforms.  It should be a subset of integration.

rljacob: Actually no its not a subset.  But it is production cases which are expensive so its a short list.

singhbalwinder: Yes, I think we can run this configuration without `_D` in the production suite, since we are running it with `_D` (debug) in the integration suite.

rljacob: You should also add this test to the production suite in maint-2.0

wlin7: Thanks for careful review, @singhbalwinder , @rljacob . I am going to remove _D in the production tests. The current PR is going to maint-2.0 (by default, as the branch was based off maint-2.0). I intend to use the same branch to create a new PR for master. Any concern on doing this? All updates in maint-2.0 since the branch was created are also in master.

rljacob: That should be fine.

singhbalwinder (approved):

components/elm/bld/namelist_files/use_cases/2015-2100_SSP585_transient.xml
wlin7: fsurdat surfdata_ne30np4.pg2_simyr1850_2015_c211105.nc is identical to the one used for v2.LR.piControl and v2.LR.historical, surfdata_ne30np4.pg2_simyr1850_c201210.nc. A new fsurdat for simyr2015 would only differ at precision level for a few land unit variables. In fact, even for simyr1850, it is using raw land unit data from around 2000. Considering these, and surfdata_ne30np4.pg2_simyr1850_2015_c211105.nc  is also in use for SSP370 simulation underway, we can keep this fsurdat for SSP585.

Attn: @golaz, @bishtgautam , @aldivi , @jjbenedict .

wlin7:
Merged to maint-2.0. To use a separate branch to create a PR for master with these changes.

wlin7:
anvil wcprodssp tests failed on maint-2.0: reference case directory for the ssp hybrid run not exists.
The reference case directory exists in the inputdata server. It was not automatically download.
Manually copy to /lcrc/group/e3sm/data/inputdata/e3sm_init
No changes to the PR needed. The tests still need to be blessed.

PR #4889, opened 2022-04-14
HOMME: prevent full rebuild due to sha change in standalone mode

bartgol:
Currently, in homme standalone mode, the GIT sha of the repo is stored in the config.h files. This means that if one adds a commit, and/or rewrites history, the sha changes. Since these files are included (directly or indirectly) by _a lot_ of other files in Homme, this cause basically a full rebuild of Homme.

This PR allows to do changes that affect the git sha (such as amending a commit) without triggering a full rebuild of all execs/libs. I verified that the only files to be rebuilt are all the copies of namelist_mod, which is the only place that uses the GIT sha (printing it to screen for debug purposes).

Side mod: when the var CPRNC_DIR is not set, Homme set it to the internal path to CIME's cprnc. However, recent changes in CIME changed the location of cprnc, so this path had to be adjusted. This only affects builds where the CPRNC_DIR var is _not_ set. If a mach file already stores CPRNC_DIR, nothing changes with this PR.


components/homme/CMakeLists.txt
```diff
@@ -409,7 +411,7 @@ if(NOT BUILD_HOMME_WITHOUT_PIOLIBRARY)
     
     SET (CPRNC_INSTALL_DIR ${HOMME_BINARY_DIR}/utils/cime/tools/cprnc)
     SET (CPRNC_BINARY ${HOMME_BINARY_DIR}/utils/cime/tools/cprnc/cprnc)
-    ADD_SUBDIRECTORY(utils/cime/tools/cprnc)
```
oksanaguba, line 412: does this change belong to this PR?

bartgol: No, but you can't build Homme standalone without specifyin CPRNC_DIR without this mod. If you'd rather split into two PRs, I can certainly do that. I just thought this would speed up things.

oksanaguba:
homme, hommebfb pass on chrysalis.

oksanaguba:
in next

PR #4891, opened 2022-04-15
Increase the maximum EAM history tapes to 15

tangq:
The "SPECIAL" v2 production runs require more than 10 EAM history output tapes (https://acme-climate.atlassian.net/wiki/spaces/EWCG/pages/1850376218/E3SMv2+history+variable+namelists+for+production+runs). But the current maint-2.0 branch only allows up to 10 EAM tapes. This PR includes changes to both the code and namelist variables and increases the limit to 15 tapes.

We should merge this to master as well.


tangq:
Not sure why so many other commits were added to this PR. I will create a new, clean one.

PR #4898, opened 2022-04-20
Change a reproducible sum algorithm in SI solver

hyungyukang:
Change a reproducible sum algorithm in the ocean split-implicit barotropic mode solver. The changed algorithm uses a global reproducible summation module in components/mpas-framework/src/framework/mpas_global_sum_mod.F

This PR passed `PEM_Ln9_D.T62_oQU240.GMPAS-IAF.cori-knl_gnu` test with `config_time_integrator = 'split_implicit'` and `config_btr_si_partition_match_mode = .true.`.

[BFB]

### DIFF
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_si.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_si.F
@@ -35,6 +35,7 @@ module ocn_time_integration_si
    use mpas_threading
    use mpas_timekeeping
    use mpas_log
+   use mpas_global_sum_mod
 
    use ocn_config
    use ocn_mesh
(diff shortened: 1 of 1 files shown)


philipwjones (approved):
Approve based on code inspection and the previous successful PEM test that shows the sums themselves are reproducible across node configurations. Note that the branch naming convention uses hyphens and not underscores.

The new algorithm is not likely to give a bfb result compared to the old, so I'm guessing the bfb flag here is because this option is off by default?

hyungyukang:
> Approve based on code inspection and the previous successful PEM test that shows the sums themselves are reproducible across node configurations. Note that the branch naming convention uses hyphens and not underscores.
> 
> The new algorithm is not likely to give a bfb result compared to the old, so I'm guessing the bfb flag here is because this option is off by default?

@philipwjones , thanks for reviewing and approving this PR. Exactly. The `config_btr_si_partition_match_mode` is turned off by default, so the SI solver with default options is exactly the same with a previous version.

jonbob:
Test merge passes:

merged to next

PR #4899, opened 2022-04-20
Added SSP370/585 tests and updated use_case files

wlin7:
Tests are added for SSP370 and SSP585.
Both are added for e3sm_prod tests, only SSP370 is added for e3sm_integration
as the two uses the same testmods settings (hybrid run from historical).

Changes are also added to ELM's SSP585 use case file to update fsurdat and
flanduse_timeseries entries to newest file versions that use the ne30.pg2 grid.

[BFB] no impact on existing tests. New baselines will be needed after this PR.


wlin7:
This PR is the same as #4887 but for master. To further check if two previous commits to maint-2.0, 5db1f41 and fe54e14, would cause issue to master. In that case, this PR would not proceed. A new branch based on current master would be created to add the SSP tests and update the use case files.

For both #4887 and this PR, @jjbenedict is double checking the proper fsurdat and flanduse_timeseries files were generated and used.

cime_config/machines/config_machines.xml
```diff
@@ -342,15 +342,15 @@
...
         <command name="rm">intel</command>
         <command name="load">intel/19.0.3.199</command>
       </modules>
 
       <modules compiler="gnu">
-        <command name="swap">PrgEnv-intel PrgEnv-gnu/6.0.9</command>
+        <command name="swap">PrgEnv-intel PrgEnv-gnu/6.0.10</command>
         <command name="rm">gcc</command>
-        <command name="load">gcc/9.3.0</command>
+        <command name="load">gcc/10.3.0</command>
```
rljacob, line 353: This cori-haswell change shouldn't be part of this PR.

PR #4902, opened 2022-04-20
Add Griffies 2005 kappa option to GM parameterization (AMOC PR 5/6)

mark-petersen:
Add new option to compute GM-Kappa, `config_GM_closure = 'Griffies'`, which uses the closure from Griffies et al (2005) eqn 15 (https://www.ocean-sci.net/1/45/2005):
```
A_{gm} = alpha |grad_z rho|  L^2 g / ( rho_0 N_0)
```
where (quoting the paper)
> alpha is a dimensionless tuning constant set to 0.07, L is a constant length scale set to 50 km, N_0 is a constant buoyancy frequency set to 0.004 s−1, g=9.8ms−1 is the acceleration of gravity,  rho_0=1035 kgm−3 is the reference density for the Boussinesq approximation, and |grad_z rho| is the average of the _horizontal_ density gradient taken over the depth range 100m to 2000m.

In this PR, we add namelist flags for the constants alpha, L, rho_0, and N_0.

Note, I have this PR into #4864 right now. I will change it to use a base of master after that is merged.

BFB
NML
do not merge


mark-petersen:
Passes MPAS-O nightly regression suite in debug mode with gnu and intel and `config_GM_closure = 'Griffies'`. Also passes 
```
```

components/mpas-ocean/src/shared/mpas_ocn_gm.F
```diff
@@ -647,6 +649,69 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, &
...
+                  sum_gradDensity = sum_gradDensity / ltsum
+
+                  ! This is Griffies et al. (2005) eqn 15:
+                  ! A_{gm} = alpha |grad_z rho| L^2 g / ( rho_0 N_0)
+                  gmBolusKappa(iEdge) = coef * sum_gradDensity
+
+                  ! These were the original lines from Lukes cryo branch:
+                  ! gmBolusKappa(iEdge) = config_GM_Griffies_alpha * config_gm_Griffies_length**2.0_RKIND * &
+                  !                      gravity * abs(sumN2) / (1.0E-15 + config_GM_Griffies_No)
+                  ! note that rho_0 is missing from the denominator.
```
mark-petersen, line 703: @vanroekel please look carefully at this. Your original lines are above, also here: https://github.com/E3SM-Project/E3SM/blob/9235712a8b87197a2b745fc55fb5ed24bf54e946/components/mpas-ocean/src/shared/mpas_ocn_gm.F#L678-L679

Compare to Griffies et al. (2005) eqn 15, you are missing the `rho_0` in the denominator. So I added it with another flag, since he uses 1035 while we use `config_density0=1026`. If your original line was missing `1/rho_0` then `gmBolusKappa` would have been too small by 1000, so results would be similar to turning off GM.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
```diff
@@ -647,6 +649,69 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, &
...
+               cell2 = cellsOnEdge(2, iEdge)
+
+               k = minLevelEdgeBot(iEdge)+1
+               zEdge = -layerThickEdgeMean(k-1,iEdge)
+               do while(zEdge > -config_GM_Griffies_min_depth .and. k < maxLevelEdgeTop(iEdge))
+                  k = k+1
+                  zEdge = zEdge - layerThickEdgeMean(k-1,iEdge)
+               end do
+
+               do while(zEdge > -config_GM_Griffies_max_depth .and. k < maxLevelEdgeTop(iEdge))
```
mark-petersen, line 682: Performance note: This is for a general vertical coordinate, but we run all standard simulations with z-star (except under ice shelves, where GM doesn't really matter). We could have two versions of this - one that computes (say) `k_GM_Griffies_start` and `k_GM_Griffies_end` on init, and another that computes it for every edge on every time step, chosen with a flag between the two.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
vanroekel: @mark-petersen yes that rho_o absolutely should be there.  I know I checked the values at one point so that change must not have gotten pushed to you.  The units don't work out to m2/s without it, so support adding rho_o to the denominator

mark-petersen: @vanroekel thanks for reviewing. Agreed, let's just use `config_density0` here. I just revised.

components/mpas-ocean/bld/namelist_files/namelist_definition_mpaso.xml
```diff
@@ -581,6 +581,54 @@ Valid values: values between zero and bottom depth
...
+	category="GM_eddy_parameterization" group="GM_eddy_parameterization">
+maximum depth for the integral of the Griffies parameter
+
+Valid values: positive reals greater than min_depth
+Default: Defined in namelist_defaults.xml
+</entry>
+
+<entry id="config_GM_Griffies_compute_levels" type="char*1024"
+	category="GM_eddy_parameterization" group="GM_eddy_parameterization">
+Compute the min and max vertical k indices based on config_GM_Griffies_min_depth, config_GM_Griffies_max_depth just one on init (for 'init')  or every time step ('every_timestep'). For z-level and z-star vertical coordinates it is better to compute on init for better performance, to avoid re-computing the min and max level for every edge and every timestep.
```
mark-petersen, line 602: @vanroekel  I added this option because on all our standard runs, we don't need to recompute the min/max level on every edge, for every time step.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
```diff
@@ -647,6 +652,104 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, &
...
+               !$omp do schedule(runtime) private(k, cell1, cell2, sum_gradDensity, ltSum, countN2, &
+               !$omp             zEdge, ltAvg)
+               do iEdge = 1, nEdges
+                  sum_gradDensity = 0.0_RKIND
+                  ltSum = 0.0_RKIND
+                  countN2 = 0
+                  cell1 = cellsOnEdge(1, iEdge)
+                  cell2 = cellsOnEdge(2, iEdge)
+
+                  do k = max(Griffies_min_level,minLevelEdgeBot(iEdge)),min(Griffies_max_level,maxLevelEdgeTop(iEdge))
```
mark-petersen, line 727: @vanroekel That flag results in a simple loop here. I have your original code above. Let me know what you think of that.

darincomeau:
5 year `CRYO1850.ne30pg2_ECwISC30to60E2r1` tests:
[Baseline, constant GM](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.milena/20220301.CRYO1850.ne30pg2_ECwISC30to60E2r1.chrysalis.CryoBranchBaseline/Years1-5/)

This PR:
`config_GM_closure = 'Griffies'`
[GM Griffies](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/20220427.CRYO1850.ne30pg2_ECwISC30to60E2r1.chrysalis.CryoBranchGMGriffies/yrs1-5/)

milenaveneziani:
Result of 5-year test with SORRM:

Baseline, constant GM: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.sprice/20220223.CRYO1850.ne30pg2_SOwISC12to60E2r4.chrysalis.CryoBranchBaseline/yrs1-5/

Baseline, Visbeck GM: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.milena/20220411.CRYO1850.ne30pg2_SOwISC12to60E2r4.chrysalis.CryoBranchBaselineGMVisbeck/Years1-5/

This PR: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.milena/20220428.CRYO1850.ne30pg2_SOwISC12to60E2r4.chrysalis.CryoBranchGMGriffies/Years1-5/

milenaveneziani:
I wanted to see how different the GM kappa and bolus velocity were with this new GM option, so I plotted them and compared them to simple Visbeck. Here is the result:
Screen Shot 2022-05-02 at 3 35 24 PM

@vanroekel: do these differences make sense to you? (these are both for a SORRM w/ramp case, so kappa's and bolus velocities are 0 in the SO)

milenaveneziani:
Besides the conflicts, I think I wanted to give @vanroekel a chance to comment on the kappa differences in the plots above, before approving this.
Also, I don't expect we'll need more testing for this, but let me know if we do.

mark-petersen:
After discussion with @vanroekel and others, we decided to close this PR, as there were not clear advantages during the testing, and completion of the PR would require further revisions and review.

PR #4867, opened 2022-04-04
Add ocean option for Redi kappa to be set equal to GM (AMOC PR 3/6)

mark-petersen:
add new option: `config_Redi_closure = 'equalGM'`, sets `RediKappa` equal to `gmBolusKappa` everywhere. Default remains `config_Redi_closure = 'constant'`.

[NML]
[BFB]

PR #4904, opened 2022-04-20
Add ocean option for Redi kappa to be set equal to GM (AMOC PR 3/6)

mark-petersen:
add new option: config_Redi_closure = 'equalGM', sets RediKappa equal to gmBolusKappa everywhere. Default remains config_Redi_closure = 'constant'.

[NML] (in namelist description only)
[BFB]


mark-petersen:
This PR is a recreation of https://github.com/E3SM-Project/E3SM/pull/4867. I was not able to open that one back up.

Note that this PR base is currently the https://github.com/E3SM-Project/E3SM/pull/4864 branch, because it builds on that, but so we just see the one commit relevant to this PR. I will change the base to master after https://github.com/E3SM-Project/E3SM/pull/4864 is merged.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
```diff
@@ -758,7 +758,19 @@ subroutine ocn_GM_compute_Bolus_velocity(statePool, &
...
-         if (config_Redi_closure == 'constant'.and. &
+         if (config_Redi_closure == 'equalGM') then
+            ! Set Redi closure to be the same as the GM closure, i.e. if
+            ! config_GM_closure is EdenGreatbatch or Visbeck, you will see those
+            ! same closure fields in RediKappa. The horizontal tapering still
+            ! comes from the config_Redi_horizontal_taper flag, not GM.
+            !$omp parallel
+            !$omp do schedule(runtime)
+            do iEdge=1, nEdges
+               RediKappa(iEdge) = RediHorizontalTaper(iEdge) * gmBolusKappa(iEdge)
```
vanroekel, line 769: if we are doing equalGM shouldn't we use the GMHorizontalTaper instead?  I also find it a little weird to use different tapers.  Are there circumstances we'd expect these to be different?  I guess the flexibility doesn't hurt, but could make configuration more difficult

vanroekel:
code looks fine here, just needs the transplant of the little bit of equalGM code from the 2/6 PR.

darincomeau:
5 year `CRYO1850.ne30pg2_ECwISC30to60E2r1` tests:

This PR:
[GM Visbeck, Redi kappa = GM kappa](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/20220421.CRYO1850.ne30pg2_ECwISC30to60E2r1.chrysalis.CryoBranchRediEqGMVisbeckRampFix/yrs1-5/)

config_GM_closure = 'Visbeck'
config_Redi_closure = 'equalGM'

milenaveneziani (approved):

mark-petersen:
Rebased after #4864 changes so this merges smoothly, plus addition of extra line mentioned by @vanroekel above. Passes nightly regression suite with gnu on badger with 
```
+config_GM_closure = 'Visbeck'
+config_Redi_closure = 'equalGM'
+config_GM_horizontal_taper = 'RossbyRadius'
+config_Redi_horizontal_taper = 'RossbyRadius'
```
and passes
```
```

darincomeau (approved):
Approved based on my 5 year tests.

components/mpas-ocean/src/shared/mpas_ocn_gm.F
vanroekel: @mark-petersen I'd still suggest we switch the `RediHorizontalTaper` here to `GMhorizontalTaper` this could give some unexpected results if users expect Redi to be exactly equal to GM.

mark-petersen: @vanroekel, good point. Sorry I missed this before. I made that change and retested.

milenaveneziani:
This should be ready to go, after @vanroekel's approval, right?

mark-petersen:
Yes. @vanroekel please review.

I tested with gnu debug on badger with nightly test suite, passes all with default flags; then again with `config_Redi_closure = 'equalGM'` as the only change, then tested again with these flags
```
config_GM_closure = 'Visbeck'
config_Redi_closure = 'equalGM'
config_GM_horizontal_taper = 'RossbyRadius'
config_Redi_horizontal_taper = 'RossbyRadius'
```
and it passes all tests.

components/mpas-ocean/src/Registry.xml
```diff
@@ -295,8 +295,8 @@
 					possible_values=".true. or .false."
 		/>
 		<nml_option name="config_Redi_closure" type="character" default_value="constant" units="unitless"
-					description="Control what type of function is used for Redi $\kappa$."
-					possible_values="'constant', 'data'"
+					description="Control what type of function is used for Redi $\kappa$. For 'equalGM', RediKappa is set to gmBolusKappa, so picks up the closure used by GM."
+					possible_values="'constant', 'equalGM', 'data'"
```
vanroekel, line 299: would it be good to say this only works for 2-d schemes (e.g. constant or visbeck)?  This won't work for edenGreatbatch

vanroekel (commented):
so one thing @milenaveneziani noted that should be made clearer I think is this capability as implemented doesn't work for all schemes.  I think only Visbeck and constant will work.  E&G won't as it uses a 3-D varying array so RediKappa can't copy that.  Putting comments in the registry would be sufficient for now

milenaveneziani:
Good point @vanroekel. I had forgotten about that. I agree that a comment in the Registry would be great at this point.

components/mpas-ocean/src/Registry.xml
mark-petersen: OK, I added that comment above.

vanroekel (approved):
This looks good to me.  Thanks for the extra comment @mark-petersen

mark-petersen:
Passes:
```
SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.chrysalis_intel
SMS_Ln9.T62_EC30to60E2r2.GMPAS-IAF.chrysalis_gnu
```

jonbob:
merged to next

jonbob:
merged to master

PR #4905, opened 2022-04-21
maint cime update

jgfouca:
Updates CIME to same version as current E3SM master (3de97a44ed61e8caa417ddb3d23791f95b53313d).

We needed some CIME fixes for baseline handling.

[BFB]


jgfouca:
Merging since baseline mgmt is broken on maint-2.

Issue #4819, opened 2022-02-28
Read failure with OpenMPI lock files generated on GPFS file system of Chrysalis


dqwu:
Recently, some tests run on Chrysalis failed when reading a specific NetCDF input file, with some lock files generated in the input directory. It seems to be a race condition as this issue is not always reproducible.
```
-rw-r--r-- 1 ac.jgfouca    E3SM 8 Feb 17 23:27 /lcrc/group/e3sm/data/inputdata/atm/cam/inic/homme/cami_mam3_Linoz_ne30np4_L72_c160214.nc-1115488256-2337493.loc
-rw-r--r-- 1 ac.ndkeen     E3SM 8 Feb 17 23:30 /lcrc/group/e3sm/data/inputdata/atm/cam/inic/homme/cami_mam3_Linoz_ne30np4_L72_c160214.nc-1117061120-2338509.loc
-rw-r--r-- 1 ac.onguba     E3SM 8 Feb 17 23:32 /lcrc/group/e3sm/data/inputdata/atm/cam/inic/homme/cami_mam3_Linoz_ne30np4_L72_c160214.nc-1117257728-2339568.loc
-rw-r--r-- 1 jayesh        E3SM 8 Feb 17 23:35 /lcrc/group/e3sm/data/inputdata/atm/cam/inic/homme/cami_mam3_Linoz_ne30np4_L72_c160214.nc-1117323264-2340199.loc
-rw-r--r-- 1 ac.brhillman  E3SM 8 Feb 17 23:37 /lcrc/group/e3sm/data/inputdata/atm/cam/inic/homme/cami_mam3_Linoz_ne30np4_L72_c160214.nc-1117454336-2340833.loc
-rw-r--r-- 1 wuda          E3SM 8 Feb 17 23:39 /lcrc/group/e3sm/data/inputdata/atm/cam/inic/homme/cami_mam3_Linoz_ne30np4_L72_c160214.nc-1118240768-2341335.loc
```
These tests were run with 28 nodes (1792 MPI tasks, 64 tasks per node). One of them is a simple SCORPIO test program which just opens cami_mam3_Linoz_ne30np4_L72_c160214.nc and tries to read a specific variable.

The failure is from MPI-IO API MPI_File_open(), which is called by PnetCDF API ncmpi_begin_indep_data():
MPI error (MPI_File_open) : MPI_ERR_OTHER: known error not in list

[...]

PR #4906, opened 2022-04-21
Prevent lock files in OpenMPI runs on Chrysalis

amametjanov:
Also reduce parallel test jobs from 6 to 4 to reduce license issues.

Addresses E3SM-Project/E3SM#4819

[BFB]

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -1730,7 +1730,7 @@
     <CCSM_CPRNC>/lcrc/group/e3sm/tools/cprnc/cprnc</CCSM_CPRNC>
     <GMAKE_J>8</GMAKE_J>
     <TESTS>e3sm_integration</TESTS>
-    <NTEST_PARALLEL_JOBS>6</NTEST_PARALLEL_JOBS>
+    <NTEST_PARALLEL_JOBS>4</NTEST_PARALLEL_JOBS>
     <BATCH_SYSTEM>slurm</BATCH_SYSTEM>
     <SUPPORTED_BY>E3SM</SUPPORTED_BY>
     <MAX_TASKS_PER_NODE>128</MAX_TASKS_PER_NODE>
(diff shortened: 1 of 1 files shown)


cime_config/machines/config_machines.xml
```diff
@@ -1810,6 +1810,9 @@
       <env name="NETCDF_FORTRAN_PATH">$SHELL{dirname $(dirname $(which nf-config))}</env>
       <env name="PNETCDF_PATH">$SHELL{dirname $(dirname $(which pnetcdf_version))}</env>
     </environment_variables>
+    <environment_variables mpilib="openmpi">
+      <env name="OMPI_MCA_sharedfp">^lockedfile</env>
```
rljacob, line 1814: Is that "^" supposed to be there?

amametjanov:
By default, OpenMPI uses lock files for shared file pointers: .loc and .lock files for both reads and writes. When jobs crash while reading/writing, lock files stay behind and that may cause hangs in subsequent jobs. This disables lock files.

Checked with `SMS_D_Ln5.ne4_oQU240.F2010` and a local inputdata dir (different DIN_LOC_ROOT/DIN_LOC_ROOT_CLMFORC). No lock files were created while the job was running (lock files are created and deleted without this change).

cime_config/machines/config_machines.xml
amametjanov: Yes, `^` is "not" to prevent lock files.
```
$ ompi_info --param sharedfp all 
            MCA sharedfp: individual (MCA v2.1.0, API v2.0.0, Component
                          v4.1.3)
            MCA sharedfp: lockedfile (MCA v2.1.0, API v2.0.0, Component
                          v4.1.3)
            MCA sharedfp: sm (MCA v2.1.0, API v2.0.0, Component v4.1.3)
```
More info at https://www.open-mpi.org/faq/?category=ompio

rljacob: ok

rljacob (approved):

tangq:
I started the 3rd NARRM v2 amip member yesterday. Shall I switch to the code base with this fix? BTW, I believe this PR should be on maint-2.0 too.

amametjanov:
Env-var was manually added to intel- and gcc-built openmpi modules. All new runs with existing modules should no longer produce lock files.
```
$ mod show openmpi/4.1.3-pin4k7o 2>&1 | grep OMPI_MCA_sharedfp
setenv("OMPI_MCA_sharedfp","^lockedfile")

$ mod show openmpi/4.1.3-sxfyy4k 2>&1 | grep OMPI_MCA_sharedfp
setenv("OMPI_MCA_sharedfp","^lockedfile")

$ date
Thu Apr 21 17:34:02 CDT 2022
```

(We can add NTEST_PARALLEL_JOBS=4 change in a different PR)

PR #4910, opened 2022-04-24
MPAS Standalone: Add OpenMP support to gnu-nersc

xylar:
For some reason this was omitted when OpenMP support was added 8 years ago: https://github.com/MPAS-Dev/MPAS/pull/445

This has become a problem in compass because we switched to building with OpenMP on by default: https://github.com/MPAS-Dev/compass/pull/364
Does not impact E3SM builds

[BFB]

### DIFF
--- a/components/mpas-framework/Makefile
+++ b/components/mpas-framework/Makefile
@@ -445,11 +445,14 @@ gnu-nersc:
 	"CFLAGS_DEBUG = -g -m64" \
 	"CXXFLAGS_DEBUG = -g -m64" \
 	"LDFLAGS_DEBUG = -g -m64" \
+	"FFLAGS_OMP = -fopenmp" \
+	"CFLAGS_OMP = -fopenmp" \
 	"BUILD_TARGET = $(@)" \
 	"CORE = $(CORE)" \
 	"DEBUG = $(DEBUG)" \
 	"SERIAL = $(SERIAL)" \
 	"USE_PAPI = $(USE_PAPI)" \
+	"OPENMP = $(OPENMP)" \
 	"USE_SHTNS = $(USE_SHTNS)" \
 	"CPPFLAGS = $(MODEL_FORMULATION) -D_MPI $(FILE_OFFSET) $(ZOLTAN_DEFINE)" )
 


xylar:
## Testing

I ran the compass ocean `pr`  and MALI `full_integration` test suites on Cori-Haswell with Gnu and OpenMP.  The ocean `pr` suite was successful.  The MALI suite showed the following validation failures that I think are unrelated to this PR (@matthewhoffman and @trhille, can you confirm?):
```
00:51 FAIL landice_greenland_fo_restart_test
00:53 FAIL landice_thwaites_restart_test
```

mark-petersen (approved):
Yes, this is obviously missing from the gnu-nersc flags. Strange that it was missing in the first place. Thanks @xylar for catching it.

xylar:
@matthewhoffman and @akturner, can you take a quick look at this and run any tests you would like to do?

trhille:
(@matthewhoffman and @trhille, can you confirm?):
> 
> ```
> 00:51 FAIL landice_greenland_fo_restart_test
> 00:53 FAIL landice_thwaites_restart_test
> ```
@xylar, I see these same tests failing on validation.

I'm creating my compass env using: `./conda/configure_compass_env.py --env_name compass_spack_albany --conda ~/miniconda3/ --compiler=gnu --spack /global/cfs/cdirs/e3sm/software/compass/cori-haswell/spack/ --mpi mpt --with_albany --without_openmp`

and building like so: `make gnu-nersc ALBANY=true DEBUG=true`

matthewhoffman (approved):
Based on the testing by @xylar and by @trhille and by inspection of the proposed changes, this looks good to me.

xylar:
Thanks @trhille and @matthewhoffman!

@trhille, I think the only difference between our tests is that I ran optimized and you ran debug.  Thanks for testing!

BTW, you don't need `--spack /global/cfs/cdirs/e3sm/software/compass/cori-haswell/spack/`.  The `--spack` flag was needed while we were testing out the branch for adding Spack but now you can just trust each machine to know where its Spack lives.

jonbob:

jonbob:
merged to master

PR #4911, opened 2022-04-25
Update CIME by one commit

jgfouca:
Needed to fix get_case_env

[BFB]


jgfouca:
Skipping integration testing since only get_case_env is impacted.

rljacob:
Is this still on that other branch of cime?

PR #4912, opened 2022-04-26
Correct NO biomass burning emissions

tangq:
Wrong molecular weight was used in the previous version for the NO
biomass burning emissions. The new NO surface emission file fixs it and
hence the new emiss_bb is higher than the old version.

[Non-BFB] - Non Bit-For-Bit for chemUCI compsets


tangq:
Tagging @wlin7 , @hsiangheleellnl : @mingxuanwupnnl found that the NO biomass burning emission is wrong in the old version in the chemUCI compsets and created a new file to fix it. This PR updates the chemUCI compsets to use the new NO surface emission file. I also uploaded the new file to the input data server.

mingxuanwupnnl (approved):

tangq:
Just to document the changes of NO biomass burning emissions (based on email exchanges with @mingxuanwupnnl)
emiss_bb (Tg NO/yr) for New (higher) vs. Old versions
image

PR #4917, opened 2022-04-27
Summit performance archive: set location to cli115 for all

sarats:
After confirming members of Walter's ALCC project, it made sense to set performance archive location to cli115 for everyone. 
That would enable automatic uploads to PACE for any project.

cc @whannah1 @xyuan 

[BFB]


amametjanov (approved):

minxu74:
@sarats The directory is under the cli115 project-shared directory. Other project members who are not in the cli115 project do not have permission to write files into the directory.  Maybe change it to the cli115 world-shared directory?

sarats:
@minxu74 I thought of that. Since the above project just has two users (Walter and Xingqiu) who are already in cli115, it would be fine.

PR #4922, opened 2022-04-29
run_e3sm script: Replace default case name and case group placeholders

sarats:
Replace default case name and case group placeholders.
This is required to avoid unwarranted usage of case_group that
interferes with aggregation of various production simulation campaigns.

[BFB]

Testing:
I was able to run ./create_newcase with --case-group "" as defined.

### DIFF
--- a/run_e3sm.template.sh
+++ b/run_e3sm.template.sh
@@ -24,8 +24,10 @@ readonly PROJECT="e3sm"
 # Simulation
 readonly COMPSET="WCYCL1850"
 readonly RESOLUTION="ne30pg2_EC30to60E2r2"
-readonly CASE_NAME="v2.LR.piControl"
-readonly CASE_GROUP="v2.LR"
+# TODO: CHANGE the following CASE_NAME to desired values
+readonly CASE_NAME="your_casename"
+# If this is part of a simulation campaign, ask your group lead about using a case_group label
+readonly CASE_GROUP=""
 
 # Code and compilation
 readonly CHECKOUT="20210806"
--- a/run_e3sm.template.sh
+++ b/run_e3sm.template.sh
@@ -24,7 +24,7 @@ readonly PROJECT="e3sm"
 # Simulation
 readonly COMPSET="WCYCL1850"
 readonly RESOLUTION="ne30pg2_EC30to60E2r2"
-# TODO: CHANGE the following CASE_NAME to desired values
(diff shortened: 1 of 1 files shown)


rljacob (approved):

sarats:
Looks like run_e3sm script would not pass the empty string to create_newcase invocation. So, I had to conditionally include the -case-group argument when it is defined.

rljacob:
We should do this on maint-2.0 as well.

run_e3sm.template.sh
```diff
@@ -24,8 +24,10 @@ readonly PROJECT="e3sm"
 # Simulation
 readonly COMPSET="WCYCL1850"
 readonly RESOLUTION="ne30pg2_EC30to60E2r2"
-readonly CASE_NAME="v2.LR.piControl"
-readonly CASE_GROUP="v2.LR"
+# TODO: CHANGE the following CASE_NAME to desired values
```
forsyth2, line 27: Is this an actual TODO for the developers/maintainers of `run_e3sm.template.sh` or is it a direction for people who copy this template? If the latter, something like ` # IMPORTANT` or `# BEFORE RUNNING` might be better.

run_e3sm.template.sh
```diff
@@ -245,18 +247,33 @@ create_newcase() {
...
+			--compset ${COMPSET} \
+			--res ${RESOLUTION} \
+			--machine ${MACHINE} \
+			--project ${PROJECT} \
+			--walltime ${WALLTIME} \
+			--pecount ${PELAYOUT}
+	else
+		${CODE_ROOT}/cime/scripts/create_newcase \
+			--case ${CASE_NAME} \
+			--case-group $(CASE_GROUP) \
```
forsyth2, line 265: Why the change from `{}` to `()`?

forsyth2 (approved):
I didn't run myself, but the code changes seem reasonable. I just made a couple minor comments.

run_e3sm.template.sh
sarats: It's for the user to do. I can change to "BEFORE RUNNING"

run_e3sm.template.sh
sarats: That was another way for handling empty string. We don't need it now as we are doing an explicit check. I will remove.

sarats:
Ok, I updated the instructions for user to set case before running.

PR #4928, opened 2022-04-30
Split config_pesall into component-specific config_pes

### LINKED ISSUE #4834: Issue with gpu-related slurm settings on Perlmutter
Early with Perlmutter, the following was one documented way to submit GPU batch jobs:
```
#SBATCH  --nodes=1
#SBATCH  --exclusive
#SBATCH  --constraint=gpu
#SBATCH  --gpus-per-task=1
#SBATCH  --gpu-bind=map_gpu:0,1,2,3
```

But now (not sure when exactly it started), with both stand-alone HOMME and screamv1 cime attempts, I get errors like this:

```
1: (GTL DEBUG: 1) cuIpcOpenMemHandle: invalid argument, CUDA_ERROR_INVALID_VALUE, line no 272
3: (GTL DEBUG: 3) cuIpcOpenMemHandle: invalid argument, CUDA_ERROR_INVALID_VALUE, line no 272
0: (GTL DEBUG: 0) cuIpcOpenMemHandle: invalid argument, CUDA_ERROR_INVALID_VALUE, line no 272
```

However, experimenting, I found this works:

```
#SBATCH  --nodes=1
#SBATCH  --exclusive 
#SBATCH  --constraint=gpu
#SBATCH  --gpus=$np
where $np is the number of MPI's
```

which doesn't make sense yet, but wanted to start a thread.  I can make this change to `config_batch.xml` but I prefer the original settings as it looks like I would need to use the `{{ total_tasks }}` variable. 

For now, can work-around by passing flags to case.submit:
```
./case.submit -a="--gpus-per-task=0 --gpu-bind=none --gpus=$np"
```

Note that we do NOT see this issue with MMF test -- seems to be OK with either of the above slurm settings, however, it's ~2.5x *slower* using the second (ie faster with what we have currently).

NERSC reports: "something did change in Slurm behavior, it now uses cgroups to enforce binding" and suggests using `--gpu-bind=none`

Ah, ok, I think this can be fixed by:
```
#SBATCH  --nodes=1
#SBATCH  --exclusive
#SBATCH  --constraint=gpu
#SBATCH  --gpus-per-task=1
#SBATCH  --gpu-bind=none
```
`ndk/machinefiles/PM-gpu-bind-none`

Well, it works, but this is till 2.5x slower for the MMF test I tried. So it's clearly not what we want.  NERSC isn't sure how long this state will exist, so might wait a bit.

amametjanov:
PE-layouts are picked based on active component of a case or based on the prior config_pesall for all-active compsets.

This PR also comes with a CIME update:
To a3c94512e105ff1f21adf500fd317ac56961635e
    
    Changes:
    1) Add RUNDIR as an accessible setting in the cmake build system
    2) First step in the direction of implementing async IO in CESM
    3) Add numeric time-stamp to jenkins archiving
    4) Update grid schema
    5) Set component-specific config_pes in E3SM
    6) Allow any case env to be used as a directives selector in config_batch.xml

Fixes #4834 

[BFB]

### DIFF
--- a/cime_config/allactive/config_pesall.xml
+++ b/cime_config/allactive/config_pesall.xml
@@ -5,7 +5,7 @@
   <grid name="any">
     <mach name="any">
       <pes compset="any" pesize="any">
-        <comment>any grid, any compset, any machine, 1 node</comment>
+        <comment>allactive: any grid, any compset, any machine, any pesize, 1 node</comment>
         <ntasks>
           <ntasks_atm>-1</ntasks_atm>
           <ntasks_lnd>-1</ntasks_lnd>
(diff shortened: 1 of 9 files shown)


amametjanov:
This PR depends on getting ESMCI/cime#4242 into E3SM.

rljacob:
Can you say what kind of testing you did?

rljacob (approved):

amametjanov:
I did baseline comparison for namelists, which would show if a different PE-layout was picked, and got no namelist diffs:
```
$ ./cime/scripts/create_test -c -b master --no-build e3sm_integration
$ ./cs.status.b | grep "Overall: " | wc -l
106
$ ./cs.status.b | grep "PASS.*NLCOMP" | wc -l
106
```

rljacob:
Those are good tests.  You should also check extra_coverage and high_res.

amametjanov:
Those also pass namelist checks on Chrysalis and Anvil:
```
azamat@chrlogin1:branch$ ./cs.status.extra_coverage | grep "Overall: " | wc -l
14
azamat@chrlogin1:branch$ ./cs.status.extra_coverage | grep "Overall: FAIL" | wc -l
2
azamat@chrlogin1:branch$ ./cs.status.extra_coverage | grep "FAIL" <-- these 2 fail on master too
  SMS_D_Ln5.enax4v1_enax4v1.F2010-CICE.chrysalis_intel (Overall: FAIL) details:
    FAIL SMS_D_Ln5.enax4v1_enax4v1.F2010-CICE.chrysalis_intel SETUP
  SMS_D_Ln5.twpx4v1_twpx4v1.F2010-CICE.chrysalis_intel (Overall: FAIL) details:
    FAIL SMS_D_Ln5.twpx4v1_twpx4v1.F2010-CICE.chrysalis_intel SETUP
azamat@chrlogin1:branch$ ./cs.status.extra_coverage | grep "PASS.*NLCOMP" | wc -l
12
azamat@chrlogin1:branch$ ./cs.status.extra_coverage | grep "FAIL.*NLCOMP" | wc -l
0

azamat@chrlogin1:branch$ ./cs.status.hi_res | grep "Overall: " | wc -l
3
azamat@chrlogin1:branch$ ./cs.status.hi_res | grep "PASS.*NLCOMP" | wc -l
3
azamat@chrlogin1:branch$ ./cs.status.hi_res | grep "FAIL.*NLCOMP" | wc -l
0
```

jgfouca:
@amametjanov , I have update CIME on your branch. Merging to next.

amametjanov:
I don't see any merges to next today as of now. Dashboard was cleared today. Go ahead.

jgfouca:
@amametjanov , I'm doing a quick check on mappy that the CIME update is not breaking e3sm_developer or causing diffs.

jgfouca:
Looks like lots of jobs were not able to handle the switch of genf90 from a submodule back to normal files. This is a known problem with the Jenkins Git plugin. Wade is blowing away workspaces so testing should work better tonight.

PR #4929, opened 2022-05-02
Differentiate input and input/output streams in MPAS stream manager

matthewhoffman:
This PR improves the MPAS Stream Manager by introducing an optional argument to differentiate between streams of direction INPUT and direction INPUT_OUTPUT.  Previously, the mpas_stream_mgr_read() routine ignored the difference
between streams of direction INPUT and direction INPUT_OUTPUT (i.e. restart streams).  However, there are situation where it would be desirable to read only truly direction INPUT streams.  One example is at the initial time of a coldstart run - one would typically want to read any input streams without reading the restart stream (which does not exist yet).  This PR adds an optional argument to force mpas_stream_mgr_read to ignore INPUT_OUTPUT streams and only read truly INPUT streams.  Its default value maintains the previous behavior.  A second commit modifies the MALI driver to make use of this new capability to improve and simplify the input stream handling at the initial time of a MALI simulation.

[BFB]

### DIFF
--- a/components/mpas-albany-landice/src/mode_forward/mpas_li_core.F
+++ b/components/mpas-albany-landice/src/mode_forward/mpas_li_core.F
@@ -130,33 +130,31 @@ module li_core
 
       if (config_do_restart) then
          call mpas_log_write("This is a restart: read stream 'restart'.")
+         call mpas_timer_start('io_read', .false.)
          call mpas_stream_mgr_read(domain % streamManager, streamID='restart', ierr=err_tmp)
+         call mpas_timer_stop('io_read')
          err = ior(err, err_tmp)
+         call mpas_timer_start('reset_io_alarms', .false.)
+         call MPAS_stream_mgr_reset_alarms(domain % streamManager, streamID='restart', ierr=err_tmp)
+         err = ior(err, err_tmp)
+         call mpas_timer_stop('reset_io_alarms')
       else
-         call mpas_log_write("This is not a restart: read stream 'input'.")
-         call mpas_stream_mgr_read(domain % streamManager, streamID='input', ierr=err_tmp)
+         call mpas_log_write("This is not a restart: read input streams with alarm ringing at initial time.")
+         call mpas_log_write("  (Streams with input_interval='initial_only', " // &
+                             "plus streams whose input_interval matches the start time.)")
+         call mpas_timer_start('io_read', .false.)
+         call mpas_stream_mgr_read(domain % streamManager, readInputOutputStreams=.false., ierr=err_tmp)
+         call mpas_timer_stop('io_read')
          err = ior(err, err_tmp)
+         call mpas_timer_start('reset_io_alarms', .false.)
+         call mpas_stream_mgr_reset_alarms(domain % streamManager, direction=MPAS_STREAM_INPUT, ierr=err_tmp)
+         err = ior(err, err_tmp)
+         call mpas_timer_stop('reset_io_alarms')
       end if
-      call MPAS_stream_mgr_reset_alarms(domain % streamManager, streamID='restart', ierr=err_tmp)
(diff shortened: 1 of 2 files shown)


matthewhoffman:
Discussed and tested for standalone MALI usage in the MALI-Dev fork of E3SM here: https://github.com/MALI-Dev/E3SM/pull/35

I've moved the PR here because it changes MPAS framework.  I could move the MALI changes into a separate PR, but I don't see a problem with including it here.  If we keep MALI changes here, I would close https://github.com/MALI-Dev/E3SM/pull/35

matthewhoffman:
@xylar , @mark-petersen , @akturner , I looked into the issues with `MPAS_stream_mgr_read` that we discussed on the MPAS call today and familiarized myself with why MALI ran into the issue the PR addresses but ocean and sea ice do not.  An important detail is that while restart streams have direction INPUT_OUTPUT, the input interval is set to `initial_only`, meaning a restart stream will only be attempted to read at the initial time.  The code that Mark brought up on our call was the stream manager read during timestepping, which is why the issue related to the restart stream does not apply.

On contrast, the relevant bit of of code in MPAS-Ocean is the read on init here:
https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-ocean/src/mode_forward/mpas_ocn_forward_mode.F#L208-L222
Note that there is specific logic ensuring that the stream read on init is only ever called 'input' or 'restart.  This is what MALI had also done in the past.  However, we wanted to be able to support having potentially our initial condition spread across multiple files.  (An example use case is a UQ study where we explore the impact of different basal friction fields, while the rest of the initial condition is otherwise identical.). To have initial conditions spread across multiple files, we wanted to have a call to `MPAS_stream_mgr_read` on init that just read any input files that are set up to be read at the initial time.  But the generic call to `MPAS_stream_mgr_read` was _also_ trying to read the 'restart' stream at the initial time.  This led to an error because the restart files don't even exist yet at the initial time of a cold start, and even if they did, we wouldn't actually want to read it.  

So the new optional argument I introduced in this PR tells `MPAS_stream_mgr_read` to ignore streams that are INPUT_OUTPUT (i.e. restart streams) when it looks for any streams to be read.  As mentioned above, without the optional argument, behavior should be identical as before.  (For how things are handled in the ocean, adding that argument should not change anything either, but there is no reason to add it.)

xylar:
@matthewhoffman, thanks for the clarification.  I'll do some standalone testing in compass.  @jonbob, do you have a suggestion for what the most relevant test in E3SM would be to make sure restarting is unaffected in both ocean and sea ice?  I'm happy to run such a test if it's not already part of your testing.

xylar:
I am running compass tests on this right now.  I tried the E3SM tests but ran into https://github.com/ESMCI/cime/issues/4251.  I will try with a clone of E3SM outside of compass.

xylar:
I did a test merge of this PR with the latest E3SM-Project submodule in compass (E3SM/master from 4 days ago).  Then, I ran the ocean PR test suite from compass in debug and optimized mode for the 5 compiler/MPI configurations we have on Anvil.  All passed and were bit-for-bit with the baseline I created when the submodule in compass got updated.

I ran `ERS.ne11_oQU240.WCYCL1850NS.chrysalis_intel` successfully with the same test merge (in this case, I did not compare it with a baseline) and am currently testing the same for Anvil.  Assuming that is successful, I'll approve.

xylar (approved):
This looks good to me!

akturner (approved):
Tested in standalone MPAS-Seaice

jonbob:
@matthewhoffman - not a big deal this time, but the branch naming convention is to use hyphens and not underscores...

jonbob:
test merge passes:

merged to next

jonbob:
merged to master

PR #4934, opened 2022-05-04
Fix coupler BGC carbon budget restarts

### LINKED ISSUE #4920: BGC carbon budget is non-BFB after restarting
The initial implementation of the carbon budget in the coupler is not BFB after restarting. The issue appears to be in the "direct" co2 fluxes from both ocn and lnd to the atm, and does not affect the overall results, which are BFB. The atm carbon budget terms are BFB, so it seems to be a bookkeeping or sequencing issue.

jonbob:
Make the BGC carbon budget output BFB after restarting by moving lnd and ocn BGC budget calls to cime_run_calc_budgets2, where the atm BGC call is. Previously the BGC budget calls were located with the physics budget calls but this resulted in the "direct" lnd2atm and ocn2atm terms being missed in the lnd and ocn budget calculations but included in the atm.

Fixes #4920 
[BFB]

### DIFF
--- a/driver-mct/main/cime_comp_mod.F90
+++ b/driver-mct/main/cime_comp_mod.F90
@@ -4655,13 +4655,15 @@ contains
        endif
        if (rof_present) then
           call seq_diag_rof_mct(rof(ens1), fractions_rx(ens1), infodata)
-          if (do_bgc_budgets) then
-             call seq_diagBGC_rof_mct(rof(ens1), fractions_rx(ens1), infodata)
-          endif
        endif
        if (ice_present) then
           call seq_diag_ice_mct(ice(ens1), fractions_ix(ens1), infodata, do_x2i=.true.)
-          if (do_bgc_budgets) then
+       endif
+       if (do_bgc_budgets) then
+          if (rof_present) then
+             call seq_diagBGC_rof_mct(rof(ens1), fractions_rx(ens1), infodata)
+          endif
+          if (ice_present) then
              call seq_diagBGC_ice_mct(ice(ens1), fractions_ix(ens1), infodata, do_x2i=.true.)
           endif
        endif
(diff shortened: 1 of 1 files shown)


jonbob:
Previously an exact restart test using a two-month baseline and one-month restart from the end of the first month in the baseline would fail, but only for the budg_dataGBGC field which contains the BGC budget information. So while this error did not impact any of the critical results, the subsequent carbon budgets themselves would be non-BFB. With this change, the code now passes this test and the restarted budgets are also BFB.

susburrows:
Thanks for following through on this fix, @jonbob !

susburrows (approved):

apcraig (approved):
Looks good to me.  Thanks.

driver-mct/main/cime_comp_mod.F90
```diff
@@ -4708,6 +4705,17 @@ subroutine cime_run_calc_budgets2(in_cplrun)
              call seq_diagBGC_ice_mct(ice(ens1), fractions_ix(ens1), infodata, do_i2x=.true.)
           endif
        endif
+       if (lnd_present) then
```
rljacob, line 4708: You could put these if blocks together with if(do_bgc_budgets) as the outer if.  Up to you.

rljacob (approved):

driver-mct/main/cime_comp_mod.F90
jonbob: Thanks for the suggestion @rljacob. It is cleaner that way

jonbob:
Re-tested with a two-month exact restart test using -res ne30pg2_r05_EC30to60E2r2  and -compset BGCEXP_CNTL_CNPECACNT_1850.

Merged to next

jonbob:
merged to master

PR #4936, opened 2022-05-05
Ensure that fates no competition mode defaults to false

### LINKED ISSUE #4869: FATES no competition defaults to true
@rgknox noted during recent work that FATES runs are defaulting to `use_fates_nocomp = .true.` in the lnd_in file unless it is specifically turned off in the user_nl_elm file.  I believe this may be due to a missing entry in the namelist_defaults.xml file. Specifically I think we just need to add:
`<use_fates_nocomp              use_fates=".true.">.false.</use_fates_nocomp>`

to this section of code: 

 https://github.com/E3SM-Project/E3SM/blob/e5406e7e668453261473f5bd32615887764989d6/components/elm/bld/namelist_files/namelist_defaults.xml#L1989-L2003

glemieux:
Adds entry for use_fates_nocomp to namelist_defaults.xml.

Fixes #4869.


peterdschwartz:
Just checking that the DIFFs for three of the FATES tests (not sat-phen) are expected with this PR?

rljacob:
@glemieux see above question.

glemieux:
Apologies for the late reply.  Yes, DIFFs are expected for the fates test mods.

PR #4939, opened 2022-05-06
Brobey/mpas ocean/openacc topofsplit

brobey:
Adding OpenACC to top of split time integration routine. This is mostly adding directives and should not impact anything else. Tested on Darwin with NVHPC (OpenACC), Intel, and GCC with bfb compared to head. Could not test on Cori-KNL - no longer supported in compass. Tried Cori-Haswell, but getting three fails with master and with changes. Still don't have a Chrysalis or Summit account to test on those platforms. Broke prior OpenACC extend split pull request in the process so will have to recreate and repost those changes.

### DIFF
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F
@@ -594,6 +594,18 @@ module ocn_time_integration_split
 
          stage1_tend_time = min(splitExplicitStep,2)
 
+         call mpas_pool_get_array(statePool, 'normalVelocity', &
+                           normalVelocityCur, stage1_tend_time)
+
+#ifdef MPAS_OPENACC
+         !$acc enter data copyin(layerThicknessCur, normalVelocityCur, sshCur)
+         !$acc update device(layerThickEdgeFlux)
+         if (config_use_freq_filtered_thickness .or. associated(highFreqThicknessNew) ) then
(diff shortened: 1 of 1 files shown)


jonbob:
@brobey - can you please change the title of this PR to something more descriptive?

philipwjones (changes requested):
See the specific bug fix needed for Summit.

Other (optional) changes. For OpenACC on Summit, we don't see any benefit from a parallel region (that is, separating parallel from the associated loop construct) so I guess I have a preference for keeping "parallel loop" as a combined directive and local present variables rather than having the separate parallel/end parallel and individual loop constructs. Were you seeing an advantage to splitting these on other machines?

Also, an optional cosmetic thing - would prefer to keep line lengths in the 75-80 char range to fit a terminal window so would prefer to break up long lines with continuation chars, even for directives.

We can reduce clutter a little by only using ifdefs for OpenACC when we need to distinguish from OpenMP for loop directives. Again, just a personal preference.

philipwjones (approved):
Approved after verifying most recent changes are bfb on Summit and continue to pass PR tests on others. And thanks for the additional cleanup.

mark-petersen (approved):
I compiled with gnu optimized, and it matches bfb with the previous commit for MPAS standalone nightly suite. Also tested with gnu debug.

jonbob:
passes sanity tests, merged to next

jonbob:
merged to master

PR #4940, opened 2022-05-06
Fixes CIME import paths in template files

jasonb5:
Fixes CIME imports for `template.case.run.sh` and `template.case.st_archive`.

Also fixes imports for `template.case.run` and `template.case.test`, not sure if these are needed as they're only called by CIME, which fixes up the import paths when executing the scripts.

[BFB]


jgfouca (approved):

jgfouca:
Skipping integration testing.

Issue #4935, opened 2022-05-04
Problem with EAM carbon budget with monthly restarts


jonbob:
I was testing the coupler carbon budget using a two-month restart test where I ran a baseline for two months with monthly restarts and then a second one-month test restarting from the first restart of the baseline test. The second run kept failing with the following error in EAM:
````
 CO2 MASS (kgCO2/m2) : period = monthly,: date =        10301           0

                       beg                          end                |        *NET CHANGE*
            0.00000000000000000E+00       0.43268619045104852E+01      |   0.43268619045104852E+01
-----------------------------------------------------------------------|-----------------------
           *SUM*             0.43268619045104852E+01                   |   0.43268619045104852E+01
    Relative Error:         -0.10020368862133358E+01
 time integrated flux =  -8.813325360304845E-003
 net change in state  =    4.32686190451049
 error                =    4.33567522987079
 ERROR: print_global_carbon_diags:Monthly CO2 conservation failure detected
````
From the initial run, the results should have looked like:
````
 CO2 MASS (kgCO2/m2) : period = monthly,: date =        10301           0

                       beg                          end                |        *NET CHANGE*
            0.43356752298707955E+01       0.43268619045104852E+01      |  -0.88133253603102091E-02
-----------------------------------------------------------------------|-----------------------
           *SUM*            -0.88133253603102091E-02                   |  -0.88133253603102091E-02
    Relative Error:          0.12316248208997671E-14
-----------------------------------------------------------------------|-----------------------
````
but for some reason the beginning mass is 0 from the restart

beharrop:
Thanks for pointing this out to me @jonbob .  I thought I had the restarts working correctly, but this must have slipped past.  I will take a look and see what's gone wrong.

jonbob:
Thanks @beharrop . I think it did work when I ran tests with restarts at year boundaries, for what that's worth

beharrop:
Can you give me more details on how you are encountering this error, @jonbob ?  I am having trouble reproducing the error you are seeing.

beharrop:
Thanks @jonbob .  I have been able to reproduce the problem and I have uncovered what is causing it too.

[...]

PR #4941, opened 2022-05-06
Fix restart issue for CO2 diagnostics in EAM

beharrop:
Prevent failure in the CO2 diagnostics checker in EAM when using restart files produced on time steps not equal to the last time step in EAM.  The fix is accomplished by removing the logical flag controlling when co2_diags_store_fields is called.

Fixes https://github.com/E3SM-Project/E3SM/issues/4935
[BFB]

### DIFF
--- a/components/eam/src/physics/cam/physpkg.F90
+++ b/components/eam/src/physics/cam/physpkg.F90
@@ -1373,9 +1373,7 @@ subroutine phys_run2(phys_state, ztodt, phys_tend, pbuf2d,  cam_out, &
              phys_state(c)%tc_mnst(:ncol) = phys_state(c)%tc_curr(:ncol)
           end if
        end do
-       if ( is_last_step() ) then
-          call co2_diags_store_fields(phys_state, pbuf2d)
-       end if
+       call co2_diags_store_fields(phys_state, pbuf2d)
     end if
 
     call t_startf ('physpkg_st2')


beharrop:
Previously, the CO2 diagnostics in EAM would fail when using restart files created during model integration on any time step not equal to the final one of the run.  For example, running two months and writing restart files every month would produce two sets of restart files.  The restart files produced at the end of the run would work fine for restarting with CO2 diagnostics, but the restart files produced after the first month would cause the model to fail owing to not having the correct starting CO2 amount.  The issue is that the starting CO2 amounts were only stored during the final time step.  The fix in this PR removes that issue so all restart files work as intended.

susburrows:
Hi Bryce, can you describe the testing you have already done on this?

beharrop:
So far, I've only tested that the model can restart BFB with the CO2 diagnostics.  Do you want me to run the developer's suite of tests for this as well?

susburrows:
Thanks, @beharrop. I think it is probably not necessary to run the developer's suite for this very minor change, but I'll defer to the integrators.

susburrows (approved):

jonbob (approved):
Thanks @beharrop - I'm confident that will fix the issue

wlin7:
Merged to next.

wlin7:
Merged to master.

PR #4944, opened 2022-05-11
Fix diagnostic bugs in ocean BGC

### LINKED ISSUE #4925: bugs in ocean BGC diagnostic fields
Segmentation faults occur for quite a few ocean BGC diagnostic arrays when E3SM is built with DEBUG = TRUE.  This happens because of differences between E3SM version 1 (ported BEC) and version 2 (MarBL), mostly with fields that are rarely assessed so hadn’t been noticed before.  Note this doesn’t affect the model results, only some output diagnostic fields.

Several ocean BGC diagnostic quantities that were 3D in v1 are now 2D in v2.  Specifically, particulate fluxes into the sediment (which were non-zero only in the bottom level of the BEC code) are now only defined at the bottom, so the diagnostic arrays need to be corrected.

With the addition of more refractory DOM pools, POC_REMIN has been renamed as POC_REMIN_DOCr.  Similarly for PON and POP.

Fully 3D nutrient and light limitation diagnostics have been removed in favor of only surface (and 100m average which are not yet included in E3SM) values.  These need to be renamed and reduced in dimension.

There are also bugs when calculating Nfix, CaCO3, and Si diagnostics.  We need to make sure the particular autotroph is actually an N-fixer, calcifier, or silicifier.

All of the Jint_100m diagnostics have been removed in MarBL, so we need to delete them from E3SM.

maltrud:
This PR fixes a number of diagnostic fields in ocean BGC that were ok for E3SMv1 but are no longer consistent with definitions require by MarBL in E3SMv2. 

Fixes #4925
[BFB]


jonbob (approved):
Approved based on testing

jonbob:
Test merge is BFB with master for:
````
SMS_Ld2.ne30pg2_r05_EC30to60E2r2.BGCEXP_CNTL_CNPRDCTC_1850.chrysalis_intel.elm-bgcexp
````
 Also gets past previous mpas-o issues for:
````
SMS_D.ne30pg2_r05_EC30to60E2r2.BGCEXP_CNTL_CNPRDCTC_1850.chrysalis_intel.elm-bgcexp
````
which now fails in ELM (see [Issue #4946](https://github.com/E3SM-Project/E3SM/issues/4946))

merged to next

PR #4945, opened 2022-05-12
modified vertical range for KEvertex calc to avoid NaNs

### LINKED ISSUE #4933: E3SM fails with RK4 and config_include_KE_vertex
Using the RK4 time integrator in E3SM with `config_include_KE_vertex = .true.` leads to the run failing early (not completing the first timestep) with a floating point exception (triggered in `mpas_ocn_vel_vadv.f90`). 
Toggling to `config_include_KE_vertex = .false.` allows the test simulations to run successfully (1-day C-case).

FYI, `config_include_KE_vertex` is FALSE by default in MPAS-O standalone (and the compass dynamic_adjustment tests). It is TRUE by default in E3SM.

alicebarthel:
Following @mark-petersen's suggestion, I modified the vertical range over which KEvertex is calculated. 
The previous version had NaNs in normalVelocity and activeTracers which led to an immediate State Validation Fail when RK4 and config_include_KE_vertex were used together. 
I successfully ran a 1-day test in E3SM (C-case).
@xylar  @mark-petersen let me know if you'd like me to run other tests. 

Fixes #4933 
[BFB]

### DIFF
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_rk4.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_rk4.F
@@ -1223,10 +1223,12 @@ module ocn_time_integration_rk4
 
       call mpas_pool_get_array(tracersPool, 'activeTracers', activeTracersCur, 1)
 
+      ! the loop below is on all vertlevels but we used edgemask to set
+      ! velocities to zero below topo (esp. edges between topo and ocean)
       !$omp parallel
       !$omp do schedule(runtime) private(k)
       do iEdge = 1, nEdges
-         do k = 1, nVertLevels !maxLevelEdgeTop(iEdge)
+         do k = 1, nVertLevels
             normalVelocityProvis(k, iEdge) = edgeMask(k, iEdge)*(normalVelocityCur(k, iEdge) + rkSubstepWeight &
                                            * normalVelocityTend(k, iEdge))
             normalVelocityProvis(k, iEdge) = normalVelocityProvis(k, iEdge) * (1.0_RKIND - wettingVelocity(k, iEdge))
@@ -1469,7 +1471,9 @@ module ocn_time_integration_rk4
          end do
       end do
       !$omp end do
-
+      
+      ! the loop below is on all vertlevels but we used edgemask to set
+      ! velocities to zero below topo (esp. edges between topo and ocean)
       !$omp do schedule(runtime)
       do iEdge = 1, nEdges
          do k = 1, nVertLevels
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_rk4.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_rk4.F
@@ -1165,6 +1165,7 @@ module ocn_time_integration_rk4
       real (kind=RKIND), dimension(:, :, :), pointer :: tracersGroupCur, tracersGroupProvis, tracersGroupTend
 
       integer, dimension(:), pointer :: minLevelCell, maxLevelCell, maxLevelEdgeTop
+      integer, dimension(:,:), pointer :: edgeMask
 
       logical, pointer :: config_use_tracerGroup
       type (mpas_pool_iterator_type) :: groupItr
@@ -1218,15 +1219,16 @@ module ocn_time_integration_rk4
       call mpas_pool_get_array(meshPool, 'minLevelCell', minLevelCell)
       call mpas_pool_get_array(meshPool, 'maxLevelCell', maxLevelCell)
       call mpas_pool_get_array(meshPool, 'maxLevelEdgeTop', maxLevelEdgeTop)
+      call mpas_pool_get_array(meshPool, 'edgeMask', edgeMask)
 
       call mpas_pool_get_array(tracersPool, 'activeTracers', activeTracersCur, 1)
 
       !$omp parallel
       !$omp do schedule(runtime) private(k)
       do iEdge = 1, nEdges
-         do k = 1, maxLevelEdgeTop(iEdge)
-            normalVelocityProvis(k, iEdge) = normalVelocityCur(k, iEdge) + rkSubstepWeight &
-                                           * normalVelocityTend(k, iEdge)
+         do k = 1, nVertLevels !maxLevelEdgeTop(iEdge)
+            normalVelocityProvis(k, iEdge) = edgeMask(k, iEdge)*(normalVelocityCur(k, iEdge) + rkSubstepWeight &
+                                           * normalVelocityTend(k, iEdge))
             normalVelocityProvis(k, iEdge) = normalVelocityProvis(k, iEdge) * (1.0_RKIND - wettingVelocity(k, iEdge))
          end do
       end do
@@ -1405,7 +1407,7 @@ module ocn_time_integration_rk4
       real (kind=RKIND), intent(in) :: rkWeight
       integer, intent(out) :: err
 
-      integer, pointer :: nCells, nEdges
+      integer, pointer :: nCells, nEdges, nVertLevels
       integer :: iCell, iEdge, k
 
       type (mpas_pool_type), pointer :: statePool, tendPool, meshPool
@@ -1419,6 +1421,7 @@ module ocn_time_integration_rk4
       real (kind=RKIND), dimension(:, :, :), pointer :: tracersGroupNew, tracersGroupTend
 
       integer, dimension(:), pointer :: minLevelCell, maxLevelCell
+      integer, dimension(:,:), pointer :: edgeMask
 
       logical, pointer :: config_use_tracerGroup
       type (mpas_pool_iterator_type) :: groupItr
@@ -1429,6 +1432,7 @@ module ocn_time_integration_rk4
 
       call mpas_pool_get_dimension(block % dimensions, 'nCells', nCells)
       call mpas_pool_get_dimension(block % dimensions, 'nEdges', nEdges)
+      call mpas_pool_get_dimension(block % dimensions, 'nVertLevels', nVertLevels)
 
       call mpas_pool_get_subpool(block % structs, 'state', statePool)
       call mpas_pool_get_subpool(block % structs, 'tend', tendPool)
@@ -1455,6 +1459,7 @@ module ocn_time_integration_rk4
 
       call mpas_pool_get_array(meshPool, 'minLevelCell', minLevelCell)
       call mpas_pool_get_array(meshPool, 'maxLevelCell', maxLevelCell)
+      call mpas_pool_get_array(meshPool, 'edgeMask', edgeMask)
 
       !$omp parallel
       !$omp do schedule(runtime) private(k)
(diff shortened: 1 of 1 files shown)


xylar:
@alicebarthel, I will try to run some tests today or tomorrow.  Thanks for submitting this!

jonbob:
@alicebarthel -- just a note for next time that the branch naming convention is to use all lower-case letters...

xylar:
# Testing

I made a test merge of this branch with `master` (currently just one merge behind) and tested the result against `master` in both E3SM and compass on Anvil.

For E3SM testing, I ran `ERS.ne11_oQU240.WCYCL1850NS.anvil_intel` and found it to be bit-for-bit with `master`.  @alicebarthel, for your reference, here is what I did:
```
cd E3SM/
git fetch --all -p
git reset --hard origin/master 
git submodule update --init --recursive 
cd cime/scripts/
./create_test --wait --walltime 01:00:00  -g -b master_20220512 --baseline-root /lcrc/group/e3sm/ac.xylar/e3sm_baselines ERS.ne11_oQU240.WCYCL1850NS.anvil_intel
cd ../..
git remote add alicebarthel/E3SM git@github.com:alicebarthel/E3SM.git
fetch alicebarthel/E3SM 
git worktree add ../ocn/fix-RK4-KEvertex-calc -b ocn/fix-RK4-KEvertex-calc
cd ../ocn/fix-RK4-KEvertex-calc/
git reset --hard alicebarthel/E3SM/ocn/fix-RK4-KEvertex-calc 
git merge --no-ff origin/master 
git submodule update --init --recursive 
cd cime/scripts/
./create_test --wait --walltime 01:00:00  -c -b master_20220512 --baseline-root /lcrc/group/e3sm/ac.xylar/e3sm_baselines ERS.ne11_oQU240.WCYCL1850NS.anvil_intel
```

I also ran the compass ocean `pr` test suite, again using E3SM `master` as a baseline.  All tests passed and were bit-for-bit.

xylar (approved):
This is a good fix to have.  In general, I think we don't want to be computing with invalid values.

xylar:
@alicebarthel, thanks for putting this in.  @jonbob, do you think my testing is sufficient or should we wait for @mark-petersen to return and do a review?

alicebarthel:
Thanks for the tests and instructions @xylar, I appreciate it. 
@jonbob Thanks for the note about naming conventions, I'll do better next time.

jonbob:
@xylar - I think your testing is complete and there's no need to wait for @mark-petersen to review. I'll run E3SM tests before I merge and between the two we should have it covered. Thanks again for your help

xylar:
Sounds great!

mark-petersen (approved):
Thanks @alicebarthel for making this PR and @xylar for testing. During my original debugging I ran numerous stand-alone nightly suites with comparisons, with rk4 and split explicit, with `config_include_KE_vertex = .true.`. Specifically, I just ran the nightly suite with split explicit and this is bfb with master using gnu optimized stand-alone on badger.

jonbob:
My tests are failing for this PR. SMS_D_Ld3.T62_oQU120.CMPASO-IAF.chrysalis_intel complains about the following:
````
 94: forrtl: severe (408): fort: (3): Subscript #1 of the array KINETICENERGYVERTEX has value 0 which is less than the lower bound of 1
````
It may be necessary for the lower do-loop extent to be max(1,minLevelEdgeBot(iEdge))?

**Comment by xylar** [comment]
Thanks @jonbob.  Sorry our other testing didn't turn this up.

@cbegeman, do you have recommendations on how to handle this based on similar loops over edges elsewhere in the code?

**Comment by xylar** [comment]
It seems most likely that we need a check for valid `edgesOnVertex`.  Presumably, we're using the dummy index for nonexistent edges on vertices and this is leading to `minLevelEdgeBot == 0`?

**Comment by alicebarthel** [comment]
Thanks @xylar for pointing out a relevant code snippet. I'll look into it and make the changes later today.

File: components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F [hunk 1]
```diff
@@ -809,14 +809,16 @@ subroutine ocn_diagnostic_solve_kineticEnergy(normalVelocity, &
...
+         kmax = maxLevelVertexTop(iVertex)
+         kineticEnergyVertex(:, iVertex) = 0.0_RKIND
         do i = 1, vertexDegree
           iEdge = edgesOnVertex(i, iVertex)
           r_tmp = dcEdge(iEdge) * dvEdge(iEdge) * 0.25_RKIND / areaTriangle(iVertex)
 #ifdef MPAS_OPENACC
           !$acc loop seq
 #endif
-          do k = 1, nVertLevels
+          do k = kmin, kmax
```
alicebarthel, line 821: Hi @philipwjones ! We changed the bounds of the loop. Are you happy with the OpenACC call to be left as it was before?

philipwjones: Yes, that's fine. Loop extent won't have an impact. Thanks for asking.

alicebarthel:
I was able to reproduce @jonbob's error with the old code. 
The updated code passed SMS_D_Ld3.T62_oQU120.CMPASO-IAF.chrysalis_intel  successfully! 
I only ran this one test. @jonbob, @xylar let me know if you are ok picking up the testing from here? Else, let me know which test/ test suites I should run. Thanks!

jonbob:
@alicebarthel - the tests no longer result in error messages, but they do indicate non-BFB behavior for both cases I ran so far:

The TestStatus for both shows:
```
FAIL ERS.ne11_oQU240.WCYCL1850NS.chrysalis_intel BASELINE master: DIFF
FAIL SMS_D_Ld3.T62_oQU120.CMPASO-IAF.chrysalis_intel BASELINE master: DIFF
```

cbegeman:
@alicebarthel I ran into non-BFB behavior on occasion when changing some loop limits when I introduced the `minLevel*` variables. Sometimes we got around this by keeping the loop limits the same and assigning, e.g., `kineticEnergyVertex(1:kmin-1, iVertex) = 0.0_RKIND` and `kineticEnergyVertex(kmax+1:nVertLevels, iVertex) = 0.0_RKIND` after the `k = 1, nVertLevels` loop. You might be able to come up with a better solution.

alicebarthel:
Thanks @cbegeman, I may try this, unless someone has a different suggestion.  I'll let you know if something else comes to mind!

jonbob:
In a similar loop in subroutine ocn_relativeVorticity_circulation, the inner loop is:
````
      do k = minLevelVertexTop(iVertex), maxLevelVertexBot(iVertex)
````
while for this fix we're using:
````
     kmin = minLevelVertexBot(iVertex)
     kmax = maxLevelVertexTop(iVertex)
````
so the "Bot" and "Top" are just reversed. If I apply this fix, the SMS_D_Ld3.T62_oQU120.CMPASO-IAF.chrysalis_intel test that showed non-BFB results now passes

alicebarthel:
Great to hear that! Thanks for looking into it @jonbob

mark-petersen (approved):
Tested head of PR, passes nightly suite with gnu optimized, and is bfb against master branchpoint

xylar:
@alicebarthel, the fix that @jonbob suggested might only work under RK4 if `normalVelocity` is defined for both valid and invalid edges (i.e. it is zero for edges between ocean and bedrock).  So I would caution that that change might fix the BFB problem but leave RK4 broken.  If that turns out to be the case, I think @ bcbegeman's solution will also either be non-BFB with split-explicit or will also break RK4, depending on the definition of `kmin`and `kmax`.

To remain BFB with split-explicit but still fix RK4, I think it really may be necessary to sum over the `kmin` and `kmax` as @jonbob suggested but to zero out the "invalid" normal velocities in RK4 (as happens automatically in split-explicit).  I'm not sure how to do this efficiently.  Maybe we can talk about it before you leave.

Sorry this is proving tricky!

jonbob:
@xylar -- you're correct as usual. I tested my fix with RK4 and it fails with a floating point exception, like Alice's first run that uncovered this issue.

alicebarthel:
Ok, I've thought about his more and now understand your comments above. 
I agree with @xylar that it won't be BFB unless we set the invalid velocities to zero. I have started to look at split.F and rk4.F for the differences, but have not had time to check the initialization values (0? weird fill_value?).
One thing of note is the calculation on L625, where uTemp is set to zero then the calculation is limited to the valid vertical range: 
               uTemp = 0.0_RKIND
               do k = minLevelEdgeBot(iEdge), maxLevelEdgeTop(iEdge)

@cbegeman I like your solution and I think it makes physical sense, but split_explicit gives non-zero values for KEvertex on vertices connecting land and ocean, so I think we would get in trouble changing them to zero (hence Jon's non BFB). 
I look forward to the rest of this conversation but I think the zero-ing of velocity should fall on someone else's to-do list if we are in a timeline for the PR.

rljacob:
status:  waiting for a bug to be fixed.

rljacob:
status:  @mark-petersen is working to fix this.

rljacob:
status: still waiting on @mark-petersen

mark-petersen:
@alicebarthel, based on the comments here, it is better to set up RK4 the same way as split explicit, i.e. to make sure that normalVelocity is always zero on the edges. Then this error does not occur in the first place. I was able to reproduce this failure and then solve it with these changes:
https://github.com/mark-petersen/E3SM/pull/new/fix-rk4-by-zeroing-normalvelocity

You can grab commit 3757046beb from my fork, branch fix-rk4-by-zeroing-normalvelocity. @alicebarthel could you test this in your set-up? Thanks.

xylar:
@mark-petersen, this is a nice solution!  @alicebarthel and I chatted about this branch today and came up with an alternative approach to the same block of code.  We were thinking of using `edgeMask` rather than 2 separate do loops but the outcome would be essentially the same.

mark-petersen:
Yes, either would work fine. @alicebarthel you can push either of those fixes to this branch, or ask for help if needed. An advantage of using `edgeMask` from 1 to n`VertLevels` is that it incorporates inactive top cells, and this my suggestion does not.

alicebarthel:
@mark-petersen Thank you so much for testing the BFB for split_explicit. I force pushed as requested. 
I suppose the next steps are out of my hand, but let me know if you need me to do anything. 
@xylar thanks for your help and review.

xylar (approved):
Re-approving based on @alicebarthel and @mark-petersen's testing and a visual inspecting.

xylar:
@alicebarthel, a rebase may be needed to resolve a conflict.  Do you want to give that a try?  Or leave it for @jonbob to sort out during merge? (@jonbob, what's your preference?)

jonbob:
@xylar and @alicebarthel - I'm happy to sort out that conflict when I merge this PR, if that's easier

alicebarthel:
I rebased it to the current master, manually resolved the conflict, and pushed it. @jonbob let me know if you need me to do anything else, and thanks for doing the merge!

jonbob:
Do we need @philipwjones to re-review? Or is this good to go?

xylar:
@jonbob, I think we can go ahead.  RK4 isn't really a target for the performance team so I think @philipwjones will be happy as long as we don't break anything (and we're not likely to do that in RK4 itself).

jonbob:
@xylar - I did resolve that conversation hoping it would add Phil's review approval. Anyway, I'll test this and get it merged as soon as possible

xylar:
> @xylar - I did resolve that conversation hoping it would add Phil's review approval. Anyway, I'll test this and get it merged as soon as possible

I don't think that's how GitHub works.  I think commenting on the code automatically makes you a reviewer whether you intended that.  So it's kind of important for people to either request changes or approve when they comment.  Otherwise, it's hard to know if they still need to approve.  But requesting changes sometimes feels rude because there's a big red flag. So I can see why people don't do it...

mark-petersen (approved):
Rebase was successful. Checked again by visual inspection and test compile. Still passes nightly suite with gnu debug.

jonbob:
test merge shows BFB results inside E3SM for:

merged to next

jonbob:
merged to master

PR #4948, opened 2022-05-13
Makefile updates to fix parallel build failures

brian-oneill:
Update object file dependency definitions in src/mode_init/Makefile and src/shared/Makefile to resolve build errors that occur when building ocean_model with multiple jobs (i.e. 'make -j').

Confirmed successful parallel build from scratch on Summit, Cori, and Chrysalis. No edits to code execution.

[BFB]

### DIFF
--- a/components/mpas-ocean/src/mode_init/Makefile
+++ b/components/mpas-ocean/src/mode_init/Makefile
@@ -61,6 +61,8 @@ mpas_ocn_init_soma.o: $(UTILS)
 
 mpas_ocn_init_lock_exchange.o: $(UTILS)
 
+mpas_ocn_init_dam_break.o: $(UTILS)
+
 mpas_ocn_init_internal_waves.o: $(UTILS)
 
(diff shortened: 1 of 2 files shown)


philipwjones (changes requested):
@brian-oneill  - Thanks for updating these. While testing, I ran into another dependency so I grep'd on "use ocn_" in the src/shared directory and updated the Makefile there with all of the use-based dependencies. With that version, everything worked fine. I pushed that onto my fork of E3SM in a branch of the same name (philipwjones:brian-oneill/ocean/makefile-updates ) so you can either just copy that version of the shared Makefile, merge my version into yours or however you want to do it. Then this should be good to go.

philipwjones (approved):
Great, thanks @brian-oneill  Approved based on quick QU240 tests and successful parallel builds across supported machines.

jonbob:

jonbob:
merged to master

PR #4951, opened 2022-05-16
Add new machine for CPU-only nodes of perlmutter (and other perlmutter changes)

ndkeen:
Add pm-cpu as a new machine -- CPU-only nodes of perlmutter
Move the machine "perlmutter" to pm-gpu.
Increase version of gnu and mpich.
Adjust default perlmutter machine/compilers.
Add in AMD compiler for pm-cpu.
Adjust pelayouts for the default 1-node cases (just for testing)
Add conditional for nersc slurm settings for GPU (only affects kokkos)

I've run e3sm_developer on pm-cpu (with gnu, amdclang, and nvidia) and with pm-gpu (with gnu and nvidia).
And the MMF tests on pm-gpu with gnugpu (though they are failing compare in the same way as current master).


PeterCaldwell:
This is great, thanks Noel. It would be helpful for SCREAM to get these merged sooner rather than later...

whannah1 (approved):
looks good

ndkeen:
merged to next

PR #4965, opened 2022-05-18
Ocean/nonhydro

scalandr:
This is a nonhydrostatic version of MPAS-O. This nonhydrostatic model enables new physics, allowing the correct representation of, for example, high-resolution internal waves dynamics, which are not well represented by the hydrostatic approximation. New vertical momentum and pressure terms are added in the formulation, and a coupled elliptic system is formed for the solution of the nonhydrostatic pressure correction. At every time step, this system is solved with the PETSc library using preconditioned iterative Krylov solvers. The flag _config_enable_nonhydrostatic_mode_ can be used to turn on and off the nonhydrostatic capabilities.

### DIFF
--- a/components/mpas-framework/Makefile
+++ b/components/mpas-framework/Makefile
@@ -335,7 +335,7 @@ gfortran:
 	"CC_SERIAL = gcc" \
 	"CXX_SERIAL = g++" \
 	"FFLAGS_PROMOTION = -fdefault-real-8 -fdefault-double-8" \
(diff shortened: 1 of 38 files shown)


xylar:
@scalandr, we need to close this for now.  Could you make the PR to https://github.com/E3SM-Ocean-Discussion/E3SM/pulls instead?  We can re-open this PR once we've discussed things over there.

PR #4966, opened 2022-05-18
add SCREAM physics copy and EKAT submodule for MMF

whannah1:
This creates a copy of physics code from the downstream SCREAM fork and adds the EKAT submodule in order to use P3 and SHOC physics in the MMF. The changes to actually support the use of P3/SHOC in the MMF will be added later. Note that merging the downstream SCREAM code into the E3SM repo is preferable, but nor currently possible due downstream changes that have broken certain integration tests. This copy of the SCREAM code will be deleted after we address the issues preventing us from doing an upstream merge.


xyuan (approved):
This is the first PR to implement P3/SHOC to MMF, in the following PR, the MMF interface will be created to use P3 and SHOC as microphysics and turbulent model for SAMXX.

PR #4972, opened 2022-05-20
MPAS Standalone: Fix check for LAPACK libraries

xylar:
Previously, there were error messages like:
```
ls: cannot access '/global/cfs/cdirs/e3sm/software/compass/cori-haswell/spack/spack_for_mache_1.4.1/var/spack/environments/dev_compass_1_1_0-alpha_4_gnu_mpt_netlib_lapack_petsc/.spack-env/view/liblapack.*': No such file or directory
```
even though this just means that the correct directory for LAPACK had not yet been found.

[BFB]

### DIFF
--- a/components/mpas-framework/Makefile
+++ b/components/mpas-framework/Makefile
@@ -609,11 +609,11 @@ ifeq "$(USE_LAPACK)" "true"
 ifndef LAPACK
 $(error LAPACK is not set.  Please set LAPACK to the LAPACK install directory when USE_LAPACK=true)
 endif
-ifneq (, $(shell ls $(LAPACK)/liblapack.*))
+ifneq ($(wildcard $(LAPACK)/liblapack.*), )
 	LIBS += -L$(LAPACK)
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I built standalone MPAS-Ocean with this fix.  Previously, I was seeing the error message shown above.  With this branch, the error message goes away.

xylar:
@mark-petersen, I believe this was introduced in https://github.com/E3SM-Project/E3SM/pull/4214, as I don't see any evidence of this `ls` style syntax in `MPAS-Dev/MPAS-Model`.  It seems like using `wildcard` instead of `shell ls` is the solution, as in the rest of the Makefile.

Could you take a look and let me know if you agree?

mark-petersen (approved):
@xylar thanks for fixing this. I don't remember putting in the `shell ls` myself, but the wildcard is better. I test compiled this with gnu and intel, each with debug and optimized, and all compiled fine.

jonbob:

jonbob:
merged to master

xylar:
Thanks as always, @jonbob!

PR #4974, opened 2022-05-22
Clean up 3 warnings/errors in ocean registry files

xylar:
These errors are harmless in E3SM but lead to the following messages during MPAS-Ocean compilation:
```
Reading registry file from standard input
WARNING: time_levs attribute on var isomip_bottomPressure in var_struct scratch is 0. It will be replaced with 1.
Warning: runtime_format attribute has no effect for immutable stream "conservationCheckRestart".
ERROR: Dimension time on variable sedimentFallVelocity doesn't exist.
```
The registry for the conservation check analysis member is the only one of these 3 that *might* be on in E3SM.  By removing the `runtime_format` attribute as I have done here, we get the same behavior as before, as the message above indicates.

The ISOMIP registry is only used in the standalone ocean model (and only in init mode for that particular test case).

The sediment transport analysis member is off by default in E3SM.

### DIFF
--- a/components/mpas-ocean/src/analysis_members/Registry_conservation_check.xml
+++ b/components/mpas-ocean/src/analysis_members/Registry_conservation_check.xml
@@ -270,7 +270,6 @@
 				packages="conservationCheckAMPKG"
 				input_interval="initial_only"
 				output_interval="stream:restart:output_interval"
-				runtime_format="single_file"
 				immutable="true">
 			<var name="performConservationPrecompute"/>
 			<var name="initialEnergy"/>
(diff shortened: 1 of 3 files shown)


xylar:
## Testing

I have made sure I could compile the code as before but that the 3 warnings/errors no longer appear.

@mark-petersen, when you have time, could you run the nightly test suite somewhere to make sure everything is fine?

@jonbob, I don't expect any E3SM changes.  I think the usual tests you run before merging to next should be sufficient in this trivial case.

mark-petersen (approved):
@xylar thanks for the cleanup! I compiled with gnu and intel, debug and optimized, and ran the nightly suite with gnu optimized and it is bfb with master.

I mentioned adding to this, but I'll do that on a different PR.

jonbob:
merged to master

PR #4980, opened 2022-05-24
Cleanup of sea ice registry

akturner:
Add CF compliant units to registry
Change non-CF compliant units in registry to CF compliant ones
Reformat Registry for clarity
Moved incremental remapping fields to separate registry file

fixes https://github.com/E3SM-Project/E3SM/issues/5023

[BFB]

### DIFF
--- a/components/mpas-seaice/src/Registry.xml
+++ b/components/mpas-seaice/src/Registry.xml
@@ -1545,7 +1545,7 @@
 	</nml_record>
 
 	<nml_record name="snow" in_defaults="true">
-	  	<nml_option name="config_snow_redistribution_scheme" type="character" default_value="none" units="none"
+		<nml_option name="config_snow_redistribution_scheme" type="character" default_value="none" units="none"
 			description="Use a snow redistribution scheme"
 			possible_values="'30percent (30% rule, precip only)','30percentsw (30% rulewith shortwave)','ITDsd (Lecomte PhD, 2014)','ITDrdg (like ITDsd but use level/ridged ice)','default or none (none)'"
 			icepack_name="snwredist"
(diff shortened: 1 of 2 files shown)


czender (commented):
Hi Adrian,
As far as I can tell, the new units that are specified are good, i.e., CF-compliant, and the units that were removed should indeed have been removed.
However, I do not know what to make of the dozens of variables that look like they might still need units.

1. For example, in Registry.xml, the all the concentrations for variables whose name begins with "skeletal" or "vertical" (are those vertically integrated?). It seems like the concentrations are examples where units = "1" are appropriate if they are stored as fractions. Example variables are "skeletalDOCConc", "verticalDOCConc", "skeletalDOCConcCell", "verticalDOCConcCell". Some related variables do have units, e.g., "verticalDOCIceCell" has units="mmol m-3". So it's unclear to me whether this PR is complete.

2. The tendency variables like "surfaceTemperatureTend" are also still unitless, though it seems likely a units attribute such as "K second-1" might be appropriate.

3. The alternative atmospheric forcing variables like "latentHeatFluxCouple" still lack units.

4. The drag variables like "oceanDragCoefficient" still lack units.

5. Some pond variables like "pondFreshWaterFlux" lack units.

6. Some biogeochemistry fluxes like "atmosBlackCarbonFlux" lack units.

7. Some SNICAR variables like "aerosolMassExtinctionCrossSection5band" lack units.

You get the idea. The units that are explicitly specified look are CF-compatible and are an improvement. There may be a few score variables that would still benefit from have explicitly specified units. Those units can presumably be added to this PR or a later one.

akturner:
@czender: Hi Charlie, I didn't update BGC or some other lesser variables. I decided they could have units added to them later.

akturner:
@czender, @njeffery: Thanks for the reviews

czender (approved):

jonbob:
passes:

merged to next

jonbob:
merged to master

PR #4988, opened 2022-05-27
FATES API 23

glemieux:
Align E3SM with fates api 22.1 and api 23.  Update the optical parameters in the fates parameter file (https://github.com/ESCOMP/CTSM/releases/tag/ctsm5.1.dev084) and the additional of logic in elmfates_interfaceMod to set the decomposition type via a new `decomp_method` parameter.

[non-BFB]

### DIFF
--- a/components/elm/src/main/elmfates_interfaceMod.F90
+++ b/components/elm/src/main/elmfates_interfaceMod.F90
@@ -323,6 +323,13 @@ verbose_output = .false.
         else
            call set_fates_ctrlparms('nu_com',cval='RD')
         end if
+        
+        ! Tell fates which decomposition method is being used by the host land model
+        if (use_century_decomp) then 
+           call set_fates_ctrlparms('decomp_method',cval='CENTURY')
+        else
+           call set_fates_ctrlparms('decomp_method',cval='NONE')
+        end if
 
         ! ELM ALWAYS has nitrogen and phosphorus "on"
         ! These may be in a non-limiting status (ie when supplements)
--- a/components/elm/src/main/elmfates_interfaceMod.F90
+++ b/components/elm/src/main/elmfates_interfaceMod.F90
@@ -59,6 +59,7 @@ module ELMFatesInterfaceMod
    use elm_varctl        , only : nsrest, nsrBranch
    use elm_varctl        , only : fates_inventory_ctrl_filename
    use elm_varctl        , only : use_lch4
+   use elm_varctl        , only : use_century_decomp
    use elm_varcon        , only : tfrz
    use elm_varcon        , only : spval
    use elm_varcon        , only : denice
--- a/components/elm/src/main/elmfates_interfaceMod.F90
+++ b/components/elm/src/main/elmfates_interfaceMod.F90
@@ -299,7 +299,7 @@ contains
 
      if (use_fates) then
 
-verbose_output = .false.
+        verbose_output = .false.
         call FatesInterfaceInit(iulog, verbose_output)
 
         ! Force FATES parameters that are recieve type, to the unset value
@@ -329,7 +329,7 @@ verbose_output = .false.
         if (use_century_decomp) then 
            call set_fates_ctrlparms('decomp_method',cval='CENTURY')
         else
-           call set_fates_ctrlparms('decomp_method',cval='NONE')
+           call set_fates_ctrlparms('decomp_method',cval='CTC')
         end if
 
         ! ELM ALWAYS has nitrogen and phosphorus "on"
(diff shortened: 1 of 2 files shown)


bishtgautam:
@glemieux Is this BFB?

glemieux:
@bishtgautam apologies I haven't tested this yet.  I'll run that today on Cori.

glemieux:
@bishtgautam apologies again; the fates tests will not be bfb.

bishtgautam:
Thanks, @glemieux for testing the branch. Could you please update the first comment to indicate the PR is [non-BFB]?

components/elm/src/main/elmfates_interfaceMod.F90
```diff
@@ -323,6 +324,13 @@ subroutine ELMFatesGlobals()
         else
            call set_fates_ctrlparms('nu_com',cval='RD')
         end if
+        
+        ! Tell fates which decomposition method is being used by the host land model
+        if (use_century_decomp) then 
+           call set_fates_ctrlparms('decomp_method',cval='CENTURY')
```
rgknox, line 330: @glemieux , there is also the CTC option in E3SM.  The choice between ctc and century has no impact on how fates operates, but it seems worth it to get it right in case this changes in the future.

rgknox:
A little background on the decomp flag we are now passing.  We added this, because when we couple with CTSM we need to know if we are running with their MIMICS decomposition scheme. If we are, we need to send some information to the soil about litter quality.  In our coupling with E3SM, the choice of decomposition scheme (century, ctc, none) has no bearing on how fates works or communicates, so this modification to the API is just for compatibility purposes.

components/elm/src/main/elmfates_interfaceMod.F90
rgknox: https://github.com/NGEET/fates/blob/master/main/FatesInterfaceMod.F90#L1439  Looks like it is expecting all caps "CTC"

glemieux: @rgknox is CTC option the default when running fates?  There doesn't appear to be a CTC equivalent to `use_century_decomp`.  Parsing the ELMBuildNamelist file, I think this is the case, but I wasn't entirely certain I'm reading it correctly:
https://github.com/E3SM-Project/E3SM/blob/3ac96425d48c438945049297c5bb1077566bb9fc/components/elm/bld/ELMBuildNamelist.pm#L915-L927

That said, it looks like its possible to avoid calling the `-soil_comp` option entirely, so maybe it's possible to have neither option selected?

glemieux: @thorntonpe would you be able to weigh in on the defaults for soil decomposition?  Is it possible not call `-soil_comp` option and thus not call `ctc` or `century`?

thorntonpe:
Hi Gregory,
The use_century_decomp flag is initialized as .false. in elm_varctl.F90. If it later gets set to .true. then the century decomp network will be used, otherwise the CTC network will be used by default. There is not an option to “not call ctc or century”. I think Ryan’s comment from a few days ago then captures the impact of this within FATES.
Is that helping?
Thanks,
Peter

From: Gregory Lemieux ***@***.***>
Sent: Monday, June 13, 2022 12:45 PM
Cc: Thornton, Peter E. ***@***.***>; Mention ***@***.***>
Subject: [EXTERNAL] Re: [E3SM-Project/E3SM] FATES API 23 (PR #4988)


@glemieux commented on this pull request.

In components/elm/src/main/elmfates_interfaceMod.F90<hxxps://github.com/E3SM-Project/E3SM/pull/4988#discussion_r895930008>:

> @@ -323,6 +324,13 @@ subroutine ELMFatesGlobals()

         else

            call set_fates_ctrlparms('nu_com',cval='RD')

         end if

+

+        ! Tell fates which decomposition method is being used by the host land model

+        if (use_century_decomp) then

+           call set_fates_ctrlparms('decomp_method',cval='CENTURY')

@thorntonpe<hxxps://github.com/thorntonpe> would you be able to weigh in on the defaults for soil decomposition? Is it possible not call -soil_comp option and thus not call ctc or century?

glemieux:
> Hi Gregory, The use_century_decomp flag is initialized as .false. in elm_varctl.F90. If it later gets set to .true. then the century decomp network will be used, otherwise the CTC network will be used by default. There is not an option to “not call ctc or century”. I think Ryan’s comment from a few days ago then captures the impact of this within FATES. Is that helping? 

That's perfect @thorntonpe.  Thanks!

glemieux:
@rljacob yes, this is ready

rgknox (approved):

rgknox:
looks good! approved

peterdschwartz:
@glemieux runs are failing due to missing of the new parameter file `/lcrc/group/e3sm/data/inputdata/lnd/clm2/paramdata/fates_params_api.22.1.0_12pft_c220307.nc`

peterdschwartz:
merged to next

peterdschwartz:
Had to add new parameter file to blues since mappy can't use `svn`.   
merged to master -- bless requests submitted.

PR #4892, opened 2022-04-15
Allow up to 15 EAM history tapes

tangq:
The "SPECIAL" v2 production runs require more than 10 EAM history output tapes (https://acme-climate.atlassian.net/wiki/spaces/EWCG/pages/1850376218/E3SMv2+history+variable+namelists+for+production+runs), exceeding the current limit (10 tapes). This PR includes changes to both the code and namelist variables and increases the limit to 15 tapes.

We should merge this to both maint-2.0 and master, but when I issued the PR for merging to maint-2.0 (https://github.com/E3SM-Project/E3SM/pull/4891), lots of unrelated commits were included (We only need the two commits in the present PR). I am not sure how to fix the PR to maint-2.0, so closed it for now.

The tests are successful and located at /lcrc/group/e3sm/ac.qtang/E3SMv2/test-output-maint-2.0.v2.NARRM.amip/tests.

PR #4991, opened 2022-05-28
Added support to allow eam history tapes up to 15 for maint-2.0

wlin7:
This is to mirror PR #4892 for master to increase eam history tapes
up to 15 to accomodate more extensive output configurations.

[BFB]

### DIFF
--- a/components/eam/bld/namelist_files/namelist_definition.xml
+++ b/components/eam/bld/namelist_files/namelist_definition.xml
@@ -1310,7 +1310,7 @@ Default: none
 
 <!-- History and Initial Conditions Output -->
 
-<entry id="avgflag_pertape" type="char*1(10)"  category="history"
+<entry id="avgflag_pertape" type="char*1(15)"  category="history"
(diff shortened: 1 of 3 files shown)


wlin7:
Hi @tangq , @beharrop , this is the same as #4892 but using a  branch that is based on maint-2.0 and for maint-2.0. Please help take a quick look.

The changes should be the same as in #4892 -- note that the differences between master and maint-2.0 in namelist_definition.xml do not showed up in the review page. The other two files, cam_history.F90 and cam_history_support.F90 remain identical between master and maint-2.0.

tangq (approved):
@wlin7 , how did you make this PR without showing numerous older commits?

beharrop (approved):
Looks good to me.  Thanks

tangq:
I noticed that the "squash and merge" button is not grey for my account. Does this mean users who are not integrators can merge changes into the maint-2.0 branch?

rljacob:
That should be fixed.  Thanks for pointing it out.

wlin7:
Merged to maint-2.0.

PR #4992, opened 2022-05-30
For cori use 4 nodes as the default for land-only tests

### LINKED ISSUE #4959: SMS_Ld1.hcru_hcru.I1850CRUELMCN.cori-knl_intel runs out of memory
The test `SMS_Ld1.hcru_hcru.I1850CRUELMCN.cori-knl_intel` is now hitting OOM on master of May 17th, and I know that it does not on May 5th. It is using the same number of nodes/MPI's, so it must be something else that changed. We can increase the number of nodes used for this compset/res on cori-knl?

### LINKED ISSUE #4981: Need to update machine name in MPAS config_pes.xml
It looks like I missed some references to `perlmutter` in the mpas config_pes files:

```
./components/mpas-seaice/cime_config/config_pes.xml:62:    <mach name="theta|perlmutter|cori-knl|cori-haswell|jlse">
./components/mpas-ocean/cime_config/config_pes.xml:75:    <mach name="theta|perlmutter|cori-knl|cori-haswell|jlse">
./components/mpas-albany-landice/cime_config/config_pes.xml:90:    <mach name="theta|perlmutter|cori-knl|cori-haswell|jlse">
./components/elm/cime_config/config_pes.xml:62:    <mach name="theta|perlmutter|cori-knl|cori-haswell|jlse">
```

I think for now we can just replace `perlmutter` with `pm-gpu|pm-cpu` (ie two machines to replace one).

ndkeen:
For cori use 4 nodes as the default for land-only tests (as it was before changes to layouts)

Change the machine name perlmutter to now be: pm-gpu and/or pm-cpu

Fixes #4959
Fixes #4981
[bfb]

### DIFF
--- a/components/elm/cime_config/config_pes.xml
+++ b/components/elm/cime_config/config_pes.xml
@@ -59,7 +59,7 @@
         </nthrds>
       </pes>
     </mach>
-    <mach name="theta|perlmutter|cori-knl|cori-haswell|jlse">
+    <mach name="theta|jlse|pm-gpu|pm-cpu">
       <pes compset="any" pesize="any">
         <comment>elm: default, 1 node x MAX_MPITASKS_PER_NODE mpi x 1 omp @ root 0</comment>
         <ntasks>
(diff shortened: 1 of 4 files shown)


ndkeen:
Tested with e3sm_developer on cori-knl

PR #4993, opened 2022-05-31
Fixed gridcell waterstate restart calls in restFileMod

### LINKED ISSUE #4649: ERS_Ld31.ne4_oQU240 failed due to mismatched TWS_MONTH_BEGIN in elm.h0
This test is routined done on compy: ERS_Ld31.ne4_oQU240.F2010.compy_intel

The restart test itself can be considered BFB in terms of simulation results (e.g., BFB for eam.h0 and cpl.hi). 

But TWS_MONTH_BEGIN in elm.h0 are different. In the base run,  it contains both valid and filled values. But in the restart run, all are filled value, causing the ERS test to fail. 

Test case below can be used to investigate (by comparing elm.h0.0001-01.nc.base against elm.h0.0001-01.nc.rest)
/compyfs/linw288/e3sm_scratch/ERS_Ld31.ne4_oQU240.F2010.compy_intel.20211109_080126_ingovy/run

peterdschwartz:
When grc_ws replaced part of waterstate_vars, restFileMod was not adjusted.
References to all unused data types were also removed from restFileMod.

Tested on chrysalis and resolved restart comparison errors.
Tested against e3sm_developer test suite.
Fixes #4649
[BFB]

### DIFF
--- a/components/elm/src/biogeophys/WaterStateType.F90
+++ b/components/elm/src/biogeophys/WaterStateType.F90
@@ -519,12 +519,6 @@ contains
 
     SHR_ASSERT_ALL((ubound(watsat_col) == (/bounds%endc,nlevgrnd/)) , errMsg(__FILE__, __LINE__))
 
-
-    call restartvar(ncid=ncid, flag=flag, varname='TWS_MONTH_BEGIN', xtype=ncd_double,  &
-         dim1name='gridcell', &
-         long_name='surface watertotal water storage at the beginning of a month', units='mm', &
-          interpinic_flag='interp', readvar=readvar, data=this%tws_month_beg_grc)
-
     call restartvar(ncid=ncid, flag=flag, varname='ENDWB_COL', xtype=ncd_double, &
          dim1name='column', long_name='col-level water mass end of the time step', &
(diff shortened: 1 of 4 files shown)


components/elm/src/biogeophys/WaterStateType.F90
```diff
@@ -519,12 +519,6 @@ subroutine Restart(this, bounds, ncid, flag, &
 
     SHR_ASSERT_ALL((ubound(watsat_col) == (/bounds%endc,nlevgrnd/)) , errMsg(__FILE__, __LINE__))
 
-
-    call restartvar(ncid=ncid, flag=flag, varname='TWS_MONTH_BEGIN', xtype=ncd_double,  &
```
bishtgautam, line 523: @peterdschwartz Why is this variable being removed from the restart file?

bishtgautam (approved):

peterdschwartz:
Merged to next

peterdschwartz:
Reverted to look into `SMS_D_Ld1.ne30pg2_EC30to60E2r2.WCYCLSSP370.chrysalis_intel.allactive-wcprodssp` failure.

Issue #5002, opened 2022-05-31
Issues with ridging island sea ice test case and special boundaries


akturner:
Special boundaries dont work correctly with ridging island sea ice test case

PR #5003, opened 2022-05-31
Fix needed for special boundaries used in ridging island test case

akturner:
Fix needed for special boundaries used in ridging island test case

Fixes https://github.com/E3SM-Project/E3SM/issues/5002

[BFB]


gcapodag (approved):

jonbob:
merged to master

PR #5007, opened 2022-06-02
Copy CIME's config_files.xml into our cime_config

jgfouca:
This is in preparation for the refactor of CIME to point
to this file.

[BFB]


jgfouca:
There is no possible impact to any testing. Skipping integration.

PR #5008, opened 2022-06-02
Changes to enforce correct selection of routing methods for MOSART

liho745:
Including a small but important bug fix to enforce correct usage of the routing methods for MOSART.

[BFB]

### DIFF
--- a/components/mosart/src/riverroute/MOSART_physics_mod.F90
+++ b/components/mosart/src/riverroute/MOSART_physics_mod.F90
@@ -30,7 +30,7 @@ MODULE MOSART_physics_mod
                              estimate_returnflow_deficit
   use WRM_subw_io_mod, only : WRM_readDemand, WRM_computeRelease
   use MOSARTinund_Core_MOD, only: ChnlFPexchg
-  use rof_cpl_indices, only : nt_rtm, rtm_tracers, nt_nliq, nt_nice
+  use rof_cpl_indices, only : nt_rtm, rtm_tracers, nt_nliq, nt_nice, KW, DW
   use perf_mod, only: t_startf, t_stopf
   use mct_mod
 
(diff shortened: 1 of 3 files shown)


components/mosart/src/riverroute/MOSART_physics_mod.F90
```diff
@@ -628,8 +628,7 @@ subroutine mainchannelRouting(iunit, nt, theDeltaT)
     else if(Tctl%RoutingMethod == 2) then
        call Routing_DW(iunit, nt, theDeltaT)
     else
-       print*, "Please check the routing method! There are only 2 methods available. 1==KW, 2==DW."
-	   call shr_sys_abort('Error in selecting routing method!')
+	   call shr_sys_abort('Wrong routing method! There are only 2 methods available. 1==KW, 2==DW.')
```
tanzeli1982, line 631: I am not sure whether the endrun() function (see the example below) is available for MOSART. But we usually use it to terminate the program. It may be just a wrapper of shr_sys_abort but I guess it will be safer and more compatible for parallel computing.

call endrun(msg=' ERROR: CH4 Conservation Error in LakeBGCDynamics during diffusion'//&
                     errMsg(__FILE__, __LINE__))

components/mosart/src/riverroute/RtmMod.F90
```diff
@@ -415,6 +415,9 @@ subroutine Rtmini(rtm_active,flood_active)
     Tctl%RoutingMethod = RoutingMethod
     Tctl%DLevelH2R     = DLevelH2R
     Tctl%DLevelR       = DLevelR
+    if(Tctl%RoutingMethod > 2) then
```
tanzeli1982, line 418: Same above

liho745: Same response above.

tanzeli1982 (approved):

liho745:
@bishtgautam and @rljacob, I have addressed your comments.

bishtgautam (approved):
@liho745  The changes look good.

rljacob:
Is this BFB ?

liho745:
I did not change the physics, so yes.

bishtgautam (approved):

peterdschwartz:
merged to next

PR #5009, opened 2022-06-04
Fix surface heat flux conversion and use more accurate theta_l formulation  in CLUBB

oksanaguba:
Two fixes in this PR:
1. change how SHFLX is used in clubb to make it consistent with clubb thermodynamic variable.
2. Redefine thetal to not to use an approximation https://glossary.ametsoc.org/wiki/Liquid_water_potential_temperature

A series of runs was performed first with fix 1 and then with both 1 and 2.
Using clubb energy fixer as a measure, fix 1 and fixes 1+2 reduce the fixer. Also, 1+2 is the best option among the two.

[non-BFB] for any active atmosphere with clubb

-----------------------------------------
A series of runs was performed first with fix 1 and then with both 1 and 2. 
https://acme-climate.atlassian.net/wiki/spaces/COM/pages/3214606402/CLUBB+SHFLX+fix
Using clubb energy fixer as a measure, fix 1 and fixes 1+2 reduce the fixer. Also, 1+2 is the best option among the two.

IIRC both are already fixed in scream.

Chris T ran longer climo simulations with the latest state of this branch 9823193800eed283cccc070fe09ecd194050df29 
https://acme-climate.atlassian.net/wiki/spaces/ATMOS/pages/3900244237

### DIFF
--- a/components/eam/src/physics/cam/clubb_intr.F90
+++ b/components/eam/src/physics/cam/clubb_intr.F90
@@ -1799,7 +1799,14 @@ end subroutine clubb_init_cnst
        rvm(i,k)     = state1%q(i,k,ixq)
        um(i,k)      = state1%u(i,k)
        vm(i,k)      = state1%v(i,k)
+
+#if 0
        thlm(i,k)    = state1%t(i,k)*exner_clubb(i,k)-(latvap/cpair)*state1%q(i,k,ixcldliq)
+#else
+       thlm(i,k) = ( state1%t(i,k) &
+                     - (latvap/cpairv(i,k,lchnk))*state1%q(i,k,ixcldliq) ) &
+                   * inv_exner_clubb(i,k)
+#endif
 
        if (clubb_do_adv) then
           if (macmic_it .eq. 1) then
@@ -2040,10 +2047,15 @@ end subroutine clubb_init_cnst
 
       !  Surface fluxes provided by host model
       wpthlp_sfc = real(cam_in%shf(i), kind = core_rknd)/(real(cpair, kind = core_rknd)*rho_ds_zm(1)) ! Sensible heat flux
-#if 1
+#if 0
       inv_exner_clubb_surf = 1._r8/((state1%pmid(i,pver)/p0_clubb)**(rair/cpair)) !phl Option 2
       wpthlp_sfc = wpthlp_sfc*inv_exner_clubb_surf
 #endif
+#if 1
+      inv_exner_clubb_surf = 1._r8/((state1%pint(i,pverp)/p0_clubb)**(rair/cpair)) !PB option
+      wpthlp_sfc = wpthlp_sfc*inv_exner_clubb_surf
+#endif
+
       wprtp_sfc  = real(cam_in%cflx(i,1), kind = core_rknd)/rho_ds_zm(1)                              ! Latent heat flux
       upwp_sfc   = real(cam_in%wsx(i), kind = core_rknd)/rho_ds_zm(1)                                 ! Surface meridional momentum flux
       vpwp_sfc   = real(cam_in%wsy(i), kind = core_rknd)/rho_ds_zm(1)                                 ! Surface zonal momentum flux
--- a/components/eam/src/physics/cam/clubb_intr.F90
+++ b/components/eam/src/physics/cam/clubb_intr.F90
@@ -1801,7 +1801,7 @@ end subroutine clubb_init_cnst
        vm(i,k)      = state1%v(i,k)
 
 #define NEWTHETAL
-#ifdef NEWTHETAL
+#ifndef NEWTHETAL
        thlm(i,k)    = state1%t(i,k)*exner_clubb(i,k)-(latvap/cpair)*state1%q(i,k,ixcldliq)
 #else
 !NCAR
--- a/components/eam/src/physics/cam/clubb_intr.F90
+++ b/components/eam/src/physics/cam/clubb_intr.F90
@@ -1484,6 +1484,10 @@ end subroutine clubb_init_cnst
    real(r8) :: sfc_v_diff_tau(pcols) ! Response to tau perturbation, m/s
    real(r8), parameter :: pert_tau = 0.1_r8 ! tau perturbation, Pa
 
+
+   real(r8) :: inv_exner_clubb_surf
+
+
 ! ZM gustiness equation below from Redelsperger et al. (2000)
 ! numbers are coefficients of the empirical equation
 
@@ -2036,6 +2040,10 @@ end subroutine clubb_init_cnst
 
       !  Surface fluxes provided by host model
       wpthlp_sfc = real(cam_in%shf(i), kind = core_rknd)/(real(cpair, kind = core_rknd)*rho_ds_zm(1)) ! Sensible heat flux
+#if 1
+      inv_exner_clubb_surf = 1._r8/((state1%pmid(i,pver)/p0_clubb)**(rair/cpair)) !phl Option 2
+      wpthlp_sfc = wpthlp_sfc*inv_exner_clubb_surf
+#endif
       wprtp_sfc  = real(cam_in%cflx(i,1), kind = core_rknd)/rho_ds_zm(1)                              ! Latent heat flux
       upwp_sfc   = real(cam_in%wsx(i), kind = core_rknd)/rho_ds_zm(1)                                 ! Surface meridional momentum flux
       vpwp_sfc   = real(cam_in%wsy(i), kind = core_rknd)/rho_ds_zm(1)                                 ! Surface zonal momentum flux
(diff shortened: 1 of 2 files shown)


components/eam/src/physics/cam/clubb_intr.F90
```diff
@@ -1795,7 +1799,20 @@ subroutine clubb_tend_cam( &
...
        thlm(i,k)    = state1%t(i,k)*exner_clubb(i,k)-(latvap/cpair)*state1%q(i,k,ixcldliq)
+#else
+!NCAR
+!       thlm(i,k) = ( state1%t(i,k) &
+!                     - (latvap/cpairv(i,k,lchnk))*state1%q(i,k,ixcldliq) ) &
+!                   * inv_exner_clubb(i,k)
+
+       thlm(i,k) = ( state1%t(i,k) &
+                     - (latvap/cpair)*state1%q(i,k,ixcldliq) ) &
+                   * exner_clubb(i,k)
```
vlarson, line 1814: Is this so-called "exner_clubb" actually the inverse of the Exner function?  If so, the variable name is confusing.

components/eam/src/physics/cam/clubb_intr.F90
```diff
@@ -2036,6 +2053,15 @@ subroutine clubb_tend_cam( &
 
       !  Surface fluxes provided by host model
       wpthlp_sfc = real(cam_in%shf(i), kind = core_rknd)/(real(cpair, kind = core_rknd)*rho_ds_zm(1)) ! Sensible heat flux
+#if 1
+      inv_exner_clubb_surf = 1._r8/((state1%pmid(i,pver)/p0_clubb)**(rair/cpair)) !phl Option 2
+      wpthlp_sfc = wpthlp_sfc*inv_exner_clubb_surf
+#endif
+#if 0
+      inv_exner_clubb_surf = 1._r8/((state1%pint(i,pverp)/p0_clubb)**(rair/cpair)) !Peter B option
+      wpthlp_sfc = wpthlp_sfc*inv_exner_clubb_surf
```
vlarson, line 2062: @oksanaguba, could you link to where this is implemented in CAM or CESM?  I know that NCAR talked about changing this, but I can't find the change in their code.

vlarson (changes requested):
The key test, in my mind, is whether the total energy imbalance is improved over mountains, as shown in Peter Lauritzen's manuscript.  Is it?

components/eam/src/physics/cam/clubb_intr.F90
oksanaguba: yes, is has been like this in EAM:
```
       exner_clubb(i,k) = (real(p0_clubb, kind = r8 )/state1%pmid(i,k))**(rair/cpair)
```
I believe it is fixed in SCREAM.

oksanaguba: To clarify, changing how Exner is defined would be out of scope of this PR.

components/eam/src/physics/cam/clubb_intr.F90
oksanaguba: Adam gave us a link (not sure, is it ok to share here?), there are many other changes in there as well.

oksanaguba:
@vlarson regarding your question about results: yes, the results are similar to Peter L.'s with CAM, the CLUBB energy redistribution is reduced if the fix is used in EAM. Results are on https://acme-climate.atlassian.net/wiki/spaces/COM/pages/3214606402/CLUBB+SHFLX+fix .

EAM also has a different definition of thetal, compared to the one used in CAM (iirc), as described in more detail on the same confluence page (with results corresponding to using a new, less approx. definition of thetal in CLUBB energy redistribution).

oksanaguba:
Here are scream PRs where these issues were fixed:
thetal https://github.com/E3SM-Project/scream/pull/1392
surface flux https://github.com/E3SM-Project/scream/pull/1332

oksanaguba:
@golaz @wlin7 @beharrop  -- would any of you have an opinion whether these 2 issues need to be addressed in EAM? One issue, how SHFLX is used in CLUBB, is a bug, another issue, definition of thetal, is more about approximations, though it seems that an unapproximated definition (not used in EAM) also makes energy distribution in clubb better.

If there is an interest, then this work can be revived and runs can repeated.

vlarson (approved):

rljacob:
I assume this changes answers which means it would need BGC leads approval.

oksanaguba:
@rljacob no, it is not ready. as you said, someone needs to decide they want this bug fixed or not.

rljacob:
@susburrows @jenniferholm @bpbond Do you want this in v2 BGC runs?

bpbond:
Sorry for the delay @rljacob . We (BGC leads) will look and discuss this on Thursday, and then respond; but if there's any urgency please let me know and we can move it up.

rljacob:
@bpbond did you reach a decision?

bpbond:
Whoops sorry to have missed this @rljacob . See email.

bpbond:
@rljacob We are going ahead with the v2.1alpha tag without this PR; it's not related to the v2 BGC runs, which anyway are already well underway.

rljacob:
@shaochengx @crterai do you want this in v3atm ?

components/eam/src/physics/cam/clubb_intr.F90
rljacob: @oksanaguba please share.  If its CESM's github, its ok.

oksanaguba: @rljacob  https://github.com/ESCOMP/CAM/pull/598

rljacob:
discussion:  this does not affect global conservation because CLUBB has brute force fixer.  But this is a bug.  CAM has fixed it.  Needs more review.

components/eam/src/physics/cam/clubb_intr.F90
wlin7: @oksanaguba , any reference that surface sensible heat flux in E3SM is formulated proportional to w'T'_sfc? To my knowledge, it has been in w'theta'_sfc already (say [over land here from CAM6 doc)](https://ncar.github.io/CAM/doc/build/html/cam6_scientific_guide/physics.html#surface-exchange-formulations)

vlarson: I think that @wlin7 asks the key question: is the coupler in E3SM delivering a surface sensible heat flux from the land or ocean model in terms of temperature (w'T') or potential temperature (w'theta')?  I am not familiar with E3SM's land model, but CLUBB conserves theta_l, not absolute temperature.  

CESM's land model appears to be delivering w'T', based on Fig. 12 of [Lauritzen et al. (2022)](https://agupubs.onlinelibrary.wiley.com/doi/pdf/10.1029/2022MS003117).  Fig. 12 shows that if the flux from the coupler is divided by the Exner function, then the energy conservation is better over mountainous regions.  If DOE sees the same pattern, then that suggests that the fix is implemented correctly.

rljacob:
@crterai @shaochengx is this still going to be in v3 atm?

crterai:
@rljacob We'd been tracking this PR in the background, but we hadn't included this in our present v3atm configuration because of potential unintended impacts. We'll likely not include it in the PR that will turn on the v3atm features, but we'll test this PR on top of master once that PR goes in, and if the Coupled Team (and others) deem it's acceptable, then we can merge this.

crterai:
@oksanaguba  - just letting you know that we have started testing this PR on top of v3atm (first in F2010 mode). Will report shortly on differences from the F-case simulation.

oksanaguba:
@crterai  thanks!

crterai:
I wasn't able to edit @oksanaguba 's original confluence page with the discussion and diagnostics, so I've reported F2010 comparisons here: https://acme-climate.atlassian.net/wiki/spaces/ATMOS/pages/3900244237/Page+to+track+impact+of+fixing+CLUBB+SHF+fix#F2010-diagnostics

crterai:
The differences in the climate look relatively small though I think they are large enough to be considered as Climate Changing. 
The summary differences from F2010 case is:
* RESTOM change is small in annual mean (-0.04 Wm-2), with no noticeable spatial coherence (mainly due to canceling effects in LW and SW)
* More outgoing longwave (FLUT) in Tropical West Pacific (Maritime continent region), far east Pacific, and over Tropical African region
* Not as large of an impact on SHF as I’d maybe expected (increase by 0.15 Wm-2 in global mean)
* Net surface fluxes have large regional signals, but small global mean difference
* More incoming shortwave in Tropical West Pacific and in global mean (0.25 Wm-2)
* Overall improvements in SLP, U850, and Z500 fields 
* No noticeable change to AODVIS
* Diurnal cycle not noticeably changed
A 30-year piControl simulation has also been run and is added to:
https://acme-climate.atlassian.net/wiki/spaces/ATMOS/pages/3900244237/Page+to+track+impact+of+fixing+CLUBB+SHF+fix#WCYCL1850-Coupled-diagnostics

wlin7:
Merged to next.

PR #5010, opened 2022-06-05
Add 5 ice-shelf BL fields to monthly output

xylar:
Collaborators have requested that these fields always be included in simulations with ice-shelf cavities so they can study the relationship between melt rates, thermal driving and friction velocity in the sub-ice-shelf boundary layer across many models.

This output would also likely be of interest to us, particularly as we explore new vertical coordinates in ice-shelf cavities.


xylar:
## Testing

I did two test QUwLI240 run on Anvil, one without ice-shelf fluxes (`pressure-only`) and one with ice-shelf fluxes (`standalone`). 
 `timeSeriesStatsMonthly` files sizes are nearly identical (121 MB without ice-shelf fluxes and 122 MB with). So all of the fields related to ice-shelf fluxes (including these 5 new ones) are in the noise.
 
Output of the test without ice-shelf fluxes is here:
/lcrc/group/e3sm/ac.xylar/scratch/anvil/20220603.GMPAS-IAF.T62_oQU240wLI.anvil/run
And with is here:
/lcrc/group/e3sm/ac.xylar/scratch/anvil/20220604.GMPAS-IAF.T62_oQU240wLI.anvil/run

xylar:
@jonbob, am I right that streams changes also qualify for the `NML` tag?

jonbob:
@xylar -- the streams files don't get compared as part of the NML test, so we can remove that label

jonbob:
@xylar - it is possible to make these extra fields only be added to the streams for wISC grids, if that would be preferable

xylar:
@jonbob,
>  it is possible to make these extra fields only be added to the streams for wISC grids, if that would be preferable

I was aware that would be possible but I don't think it's necessary.  Internally, MPAS-Ocean also disables (doesn't even allocate) these fields in runs without ice-shelf cavities.  So they will only be included in runs not only with ice-shelf cavities but with ice-shelf melt fluxes.  Since we haven't made any effort to treat existing fields with the `landIce` prefix in any special way for `wISC` runs, I think it's consistent to let MPAS-Ocean take care of it for these 5 fields as well.

Let me know if you prefer to exclude them nonetheless.

cbegeman (approved):
@xylar This addition will be useful to have! It looks good to me based on visual inspection. Let me know if you'd like help with any testing.

jonbob:
@xylar - I think your implementation is fine. Just thought I'd point out the possibility, just in case it made the non-wISC output any clearer. But as you said, the impact is in the noise

darincomeau (approved):
Approve by visual inspection.

milenaveneziani (approved):

xylar:
@mark-petersen, I'm taking you off as a reviewer.  We had plenty of eyes on this.

xylar:
Thanks, @milenaveneziani, @darincomeau and @jonbob for your feedback!

xylar:
... and @cbegeman (so sorry I missed you!)

jonbob:

jonbob:
merged to master

xylar:
@jonbob, thanks a lot!

PR #5013, opened 2022-06-06
Revive capability to run FIDEAL compset

brhillman:
Revive the capability to run the FIDEAL atmosphere/dynamics-only compset, which uses idealized physics using Held-Suarez forcing (and modifications thereof). Also adds a namelist option (`ideal_phys_option`) to select the type of forcing used to represent the idealized physics, with valid values of `held-suarez`, `held-suarez-williamson`, and `held-suarez-lin-williamson`. Adds a test to `e3sm_atm_integration` for the revived compset.

[BFB]

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -101,7 +101,7 @@ _TESTS = {
             "ERS_Ld5.ne4_oQU240.F2010.eam-rrtmgpxx",
             "REP_Ln5.ne4_oQU240.F2010",
             "SMS_Ld9.ne4pg2_oQU480.F2010.eam-thetahy_sl_pg2_mass",
-            "SMS_Ld9.ne4_ne4.FIDEAL",
+            "ERP_Ld9.ne4_ne4.FIDEAL",
             )
         },
 
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -101,6 +101,7 @@ _TESTS = {
             "ERS_Ld5.ne4_oQU240.F2010.eam-rrtmgpxx",
             "REP_Ln5.ne4_oQU240.F2010",
             "SMS_Ld9.ne4pg2_oQU480.F2010.eam-thetahy_sl_pg2_mass",
+            "SMS_Ld9.ne4_ne4.FIDEAL",
             )
         },
 
--- a/components/eam/bld/build-namelist
+++ b/components/eam/bld/build-namelist
@@ -1555,19 +1555,17 @@ if ( $prescribed_aero_model ne 'none' ) {
 
 # Construct the aerosol part of the rad_climate string array by looping over
 # the aerosol names and getting the default properties file for each:
-if ($aer_model != 'none') {
-    foreach my $name (@aero_names) {
-        my $source = shift(@aerosources);
-        my $file;
-        if ($source =~ 'M') {
+foreach my $name (@aero_names) {
+    my $source = shift(@aerosources);
+    my $file;
+    if ($source =~ 'M') {
         $file = "${name}_file";
-        } else {
-            $file = "${aer_model}_$name";
-        }
-        my $rel_filepath = get_default_value($file);
-        my $abs_filepath = set_abs_filepath($rel_filepath, $inputdata_rootdir);
-        $radval .= "," . quote_string($source . $name . ":" . $abs_filepath);
+    } else {
+        $file = "${aer_model}_$name";
     }
+    my $rel_filepath = get_default_value($file);
+    my $abs_filepath = set_abs_filepath($rel_filepath, $inputdata_rootdir);
+    $radval .= "," . quote_string($source . $name . ":" . $abs_filepath);
 }
 
 # Eruptive volcanic aerosols can be run with either BAM or MAM.
--- a/components/eam/bld/build-namelist
+++ b/components/eam/bld/build-namelist
@@ -550,7 +550,6 @@ if ($dyn =~ /se/) {add_default($nl, 'vect_map', 'val'=>'cart3d');}
 
 # Adiabatic or Ideal physics
 my $phys = $cfg->get('phys');
-# TODO: I do not think these are used in E3SM anymore, and should be removed.
 if ($phys eq 'adiabatic') { add_default($nl, 'atm_adiabatic',  'val'=>'.true.'); }
 if ($phys eq 'ideal')     { add_default($nl, 'atm_ideal_phys', 'val'=>'.true.'); }
 
--- a/components/eam/bld/build-namelist
+++ b/components/eam/bld/build-namelist
@@ -1586,7 +1586,7 @@ if ($nl->get_value('strat_aero_feedback') =~ /$TRUE/io) {
     $radval .= "," . quote_string("N:H2SO4M:" . $abs_filepath);
 }
 
-if (!$ideal_or_adia_or_aqua) {
+if ( !($phys eq 'ideal' or $phys eq 'adiabatic') ) {
     add_default($nl, 'rad_climate', 'val'=>$radval);
 }
 
--- a/components/eam/bld/build-namelist
+++ b/components/eam/bld/build-namelist
@@ -779,11 +779,9 @@ if ($opts{'ignore_ic_date'}) {
     # Don't set any attribute for date matching.  By putting this option first it
     # will take precedence in the case that the user has set both -ignore_ic_date
     # and -ignore_ic_year
-}
-elsif ($opts{'ignore_ic_year'}) {
+} elsif ($opts{'ignore_ic_year'}) {
     $atts{'ic_md'} = $ic_date;
-}
-else {
+} else {
     # if neither option specified then match full date
     $atts{'ic_ymd'} = $ic_date;
 }
@@ -806,7 +804,6 @@ add_default($nl,'microp_aero_wsub_scheme');
 add_default($nl,'microp_aero_wsubmin');
 add_default($nl,'use_preexisting_ice');
 
-
 #BSINGH - (when .true.)calculate solar zenith angle averaged over a time step.
 #In default model, solar zenith angle is held constant over time.
 #This flag is kept to test b4b with the default code(this must be REMOVED when we decide to make it default)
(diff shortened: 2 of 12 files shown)


components/eam/bld/build-namelist
```diff
@@ -817,11 +822,14 @@ add_default($nl,'use_rad_dt_cosz');
...
 add_default($nl, 'raytau0');
 
-# these options werre added for global RCE and they only work with RRTMGP
-my $rad_pkg = $cfg->get('rad');
-if ($rad_pkg eq 'rrtmgp') {
-  add_default($nl,'do_aerosol_rad');
-  add_default($nl,'fixed_total_solar_irradiance');
+# these options were added for global RCE and they only work with RRTMGP
+my $rad_pkg = 'none';
+if (!$ideal_or_adia_or_aqua) {
```
mt5555, line 827: shouldn't we allow RRTMGP settings for aqua planet?

mt5555:
I added @whannah1 as a reviewer, since he recently modified the physics to support EAM or MMF.

What is the strategy for topography when running Held-Suarez?  is it set to zero always?  would it be easy to have an option to enable HS with topography?

cime_config/tests.py
```diff
@@ -101,6 +101,7 @@
             "ERS_Ld5.ne4_oQU240.F2010.eam-rrtmgpxx",
             "REP_Ln5.ne4_oQU240.F2010",
             "SMS_Ld9.ne4pg2_oQU480.F2010.eam-thetahy_sl_pg2_mass",
+            "SMS_Ld9.ne4_ne4.FIDEAL",
```
oksanaguba, line 104: Is ERS a better test here?

components/eam/bld/build-namelist
```diff
@@ -550,9 +550,17 @@ if ($dyn =~ /se/) {add_default($nl, 'vect_map', 'val'=>'cart3d');}
 
 # Adiabatic or Ideal physics
 my $phys = $cfg->get('phys');
+# TODO: I do not think these are used in E3SM anymore, and should be removed.
```
whannah1, line 553: did you mean to make this comment just about the adiabatic mode?

components/eam/bld/build-namelist
whannah1: yea, this logic seems wrong. It seems better to just have a block to disable rad for ideal physics

components/eam/bld/namelist_files/namelist_defaults_eam.xml
```diff
@@ -11,7 +11,6 @@
 <dtime dyn="se"                    >1800</dtime>
 <dtime dyn="se"    hgrid="ne4np4"  >7200</dtime>
 <dtime dyn="se"    hgrid="ne11np4" >7200</dtime>
-<dtime dyn="se"    hgrid="ne16np4" >1800</dtime>
```
whannah1, line 14: Why did you delete this?

whannah1 (changes requested):
This looks pretty good except for the logic to disable radiation for the ideal case. Seems like it will cause problems for aqua/RCE cases, but maybe I'm missing something.

components/eam/bld/build-namelist
brhillman: Not exactly, I think this was before I discovered that these were used in the mct driver, not in EAM. Removing comment.

components/eam/bld/build-namelist
brhillman: Good catch, this was a mistake, will fix.

components/eam/bld/namelist_files/namelist_defaults_eam.xml
brhillman: Because the default is set to 1800 anyways. It actually seems like it should be set to something somewhat larger than 1800.

cime_config/tests.py
brhillman: ERP is probably even better, replaced.

components/eam/bld/namelist_files/namelist_defaults_eam.xml
whannah1: It seems like it would be nice to have the default time step that automatically scales with the resolution... However, based on how the MMF responds to increasing the physics time step I think there's an underlying fundamental issue with this way of thinking. We have the tendency to let the max wind speed and CFL criteria to guide our time step choice, but it seems the time scale of signals on the smallest scales should really be the thing that guides this choice. This is obvious in the case of the MMF because the clouds are behaving the same at ne4 and ne30, so using the same time step ends up making sense. But I suppose we can extend this to the standard model too because the parameterized clouds and surface fluxes are still operating at the same time scales... So I guess my preference would be to have the ne30 time step be default for any coarser resolution, and let the CFL criteria dominate our time step choice for higher resolution. 

Anyway, that's a long-winded way to say that this change is fine with me  😄

whannah1 (approved):

brhillman:
Pushed some changes to fix the issues identified here. @mt5555 right now there's no topography. I *think* it would be straightforward to add it though.

rljacob:
status:  ready to go.

components/eam/src/control/cam_control_mod.F90
```diff
@@ -65,7 +65,7 @@ subroutine cam_ctrl_set_physics_type(phys_package)
   character(len=*), parameter :: subname = 'cam_ctrl_set_physics_type'
 
   adiabatic = trim(phys_package) == 'adiabatic'
-  ideal_phys = trim(phys_package) == 'held_suarez'
+  ideal_phys = trim(phys_package) == 'ideal'
   moist_physics = .not. (adiabatic .or. ideal_phys)
   if (adiabatic .and. ideal_phys) then
     call endrun (subname//': FATAL: Only one of ADIABATIC or HELD_SUAREZ can be .true.')
```
wlin7, line 71: This endrun printout could be updated later, as ideal_phys now supports more than HELD_SUAREZ.

wlin7:
Merged to next.

wlin7:
Merged to master.

amametjanov:
`ERP_Ld9.ne4_ne4.FIDEAL.ascent_ibm` is reporting a build error:
```
"/gpfs/wolf/cli115/proj-shared/testing/E3SM/components/homme/src/share/gllfvremap_mod.F90", line 2735.52: 1516-023 (E) Subscript is out of bounds.
"/gpfs/wolf/cli115/proj-shared/testing/E3SM/components/homme/src/share/gllfvremap_mod.F90", line 2737.23: 1516-023 (E) Subscript is out of bounds.
"/gpfs/wolf/cli115/proj-shared/testing/E3SM/components/homme/src/share/gllfvremap_mod.F90", line 2740.44: 1516-023 (E) Subscript is out of bounds.
"/gpfs/wolf/cli115/proj-shared/testing/E3SM/components/homme/src/share/gllfvremap_mod.F90", line 2743.37: 1516-023 (E) Subscript is out of bounds.
"/gpfs/wolf/cli115/proj-shared/testing/E3SM/components/homme/src/share/gllfvremap_mod.F90", line 2745.37: 1516-023 (E) Subscript is out of bounds.
```
Looking at line numbers, the issue is with `elem(ie)%state%Q(:,:,1,2)` access with subscript 2.
Path on Ascent:
```
/gpfs/wolf/cli115/proj-shared/azamat/e3sm_scratch/ERP_Ld9.ne4_ne4.FIDEAL.ascent_ibm.JNextIntegration2022*/bld/e3sm.bldlog.*
```

mt5555:
FYI, a trick we've used in the past is to do something like: 

ic=2
elem(ie)%state%Q(:,:,1,ic)

(this is code that will not be used in FIDEAL, but is still compiled).

amametjanov:
This lets the build and run proceed to completion. Thanks.

ambrad:
I approved PR #5068 so others don't have to look at it. FYI, that code is a unit test that uses the first two entries of Q, so it's not an issue for real FIDEAL runs even with pg2.

cime_config/tests.py
```diff
@@ -101,6 +101,7 @@
             "ERS_Ld5.ne4_oQU240.F2010.eam-rrtmgpxx",
             "REP_Ln5.ne4_oQU240.F2010",
             "SMS_Ld9.ne4pg2_oQU480.F2010.eam-thetahy_sl_pg2_mass",
+            "ERP_Ld9.ne4_ne4.FIDEAL",
```
amametjanov, line 104: Due to heavier output on day 6, this test is running into a MEMLEAK fail on Crusher+gnu (and pm-cpu + gnu). Is it okay to add a test-mod to `xmlchange PIO_ROOT=1`: i.e. change test to `ERP_Ld9.ne4_ne4.FIDEAL.allactive-pioroot1`?
```
 tStamp_write: model date =   00010104       0 wall clock = 2022-08-13 01:22:15 avg dt =     0.35 dt =     0.34
 memory_write: model date =   00010104       0 memory =     754.54 MB (highwater)        121.70 MB (usage)  (pe=    0 comps= cpl ATM LND ICE OCN GLC ROF WAV IAC ESP)
 tStamp_write: model date =   00010105       0 wall clock = 2022-08-13 01:22:15 avg dt =     0.32 dt =     0.24
 memory_write: model date =   00010105       0 memory =     754.54 MB (highwater)        122.21 MB (usage)  (pe=    0 comps= cpl ATM LND ICE OCN GLC ROF WAV IAC ESP)
 Write restart file at   00010106       0
(seq_rest_write)  write rpointer file rpointer.drv
 (seq_io_wopen)  create file ERP_Ld9.ne4_ne4.FIDEAL.crusher_gnu.JNextIntegration20220813_010620.cpl.r.0001-01-06-00000.nc
 Restart filename: ERP_Ld9.ne4_ne4.FIDEAL.crusher_gnu.JNextIntegration20220813_010620.cpl.r.0001-01-06-00000.nc
 tStamp_write: model date =   00010106       0 wall clock = 2022-08-13 01:22:16 avg dt =     0.32 dt =     0.32
 memory_write: model date =   00010106       0 memory =    1172.77 MB (highwater)        157.24 MB (usage)  (pe=    0 comps= cpl ATM LND ICE OCN GLC ROF WAV IAC ESP)
 tStamp_write: model date =   00010107       0 wall clock = 2022-08-13 01:22:16 avg dt =     0.30 dt =     0.21
 memory_write: model date =   00010107       0 memory =    1178.77 MB (highwater)        157.30 MB (usage)  (pe=    0 comps= cpl ATM LND ICE OCN GLC ROF WAV IAC ESP)
```

ndkeen: https://github.com/E3SM-Project/E3SM/issues/5076

PR #5014, opened 2022-06-08
Add smaller debug-queue PE-layout for tests on Chrysalis

amametjanov:
This is for faster queue turn-around in nightly testing.
Current PE-layouts for tests use 40+ nodes and test jobs get held up in queue.

[NML] - due to PE-layout changes


rljacob:
Wouldn't this make every test run on 4 nodes?  What I was hoping for was:  large 56-node jobs are reduced to 20 (or maybe 15) and everything currently less then 20 stays as is.  Some tests run on 1 node.

PR #5018, opened 2022-06-10
Update test-PEs on Chrysalis with ne4, ne30, RRM PEs

amametjanov:
This updates E3SM-Project/E3SM#5014 with ne4-PEs (to avoid over-decomposing LND gridcells).

[NML] - due to PE-layout changes
[non-BFB] for a few ocean diagnostics.

### DIFF
--- a/cime_config/testmods_dirs/config_pes_tests.xml
+++ b/cime_config/testmods_dirs/config_pes_tests.xml
@@ -29,4 +29,33 @@
       </pes>
     </mach>
   </grid>
+  <grid name="a%ne4np4.*">
+    <mach name="chrysalis">
+      <pes compset="any" pesize="any">
+        <comment>tests+chrysalis: any compset on ne4 grid, 3x32x2 NODESxMPIxOMP</comment>
+        <MAX_MPITASKS_PER_NODE>32</MAX_MPITASKS_PER_NODE>
+        <MAX_TASKS_PER_NODE>64</MAX_TASKS_PER_NODE>
+        <ntasks>
+          <ntasks_atm>96</ntasks_atm>
+          <ntasks_ice>96</ntasks_ice>
+          <ntasks_cpl>96</ntasks_cpl>
+          <ntasks_lnd>96</ntasks_lnd>
+          <ntasks_rof>96</ntasks_rof>
+          <ntasks_ocn>96</ntasks_ocn>
+          <ntasks_glc>96</ntasks_glc>
+          <ntasks_wav>96</ntasks_wav>
+        </ntasks>
+        <nthrds>
+          <nthrds_atm>2</nthrds_atm>
+          <nthrds_ice>2</nthrds_ice>
+          <nthrds_cpl>2</nthrds_cpl>
+          <nthrds_lnd>2</nthrds_lnd>
+          <nthrds_rof>2</nthrds_rof>
+          <nthrds_ocn>2</nthrds_ocn>
+          <nthrds_glc>2</nthrds_glc>
+          <nthrds_wav>2</nthrds_wav>
+        </nthrds>
+      </pes>
+    </mach>
+  </grid>
 </config_pes>
--- a/cime_config/testmods_dirs/config_pes_tests.xml
+++ b/cime_config/testmods_dirs/config_pes_tests.xml
@@ -58,4 +58,40 @@
       </pes>
     </mach>
   </grid>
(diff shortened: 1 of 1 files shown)


amametjanov:
Expected impact on `e3sm_prod` and `e3sm_integration` testing on Chrysalis: DIFFs due to PE-layout changes in WCYCL cases.

e3sm_prod: https://my.cdash.org/viewTest.php?buildid=2176823
- 2 PASS in F-cases: still on 4 nodes at 32x2.
- 3 DIFF in fully-coupled cases likely due to non-repro-sums in OCN.
  Node reduction: two tests 53-downto-4 nodes and one 60-downto-4.

e3sm_integration: https://my.cdash.org/viewTest.php?buildid=2176822
- 97 PASS
- 7 DIFF for the same reason as above (ELM DIFF due to unrelated ELM PR).
  Node reduction: 6x 53->4, 1x 56->4.

Total time to run `e3sm_integration` is about the same at 2-3 hours: `compute` and `debug` nodes were available for `master` and this branch.

rljacob:
MPAS-O is supposed to be reproducible when you change the processor count.  So I don't know why those e3sm-prod cases are diffing.

jonbob:
Thanks @rljacob - yes, mpas-o shouldn't change answers due to processor counts. We've been passing PEM tests regularly so this makes no sense.

ambrad:
Are these diffs only in MPAS-O diagnostic fields? I vaguely recall some diagnostic fields are not PE-layout-invariant.

jonbob:
I ran SMS_D_Ld1.ne30pg2_r05_EC30to60E2r2.WCYCL1850.chrysalis_intel and it passed

jonbob:
@ambrad - the DIFFs aren't just in diagnostics fields but show up in the coupler history files. It looks like the ne30pg2_EC30to60E2r2 configurations are failing, but I think the default layout for those has been unthreaded and the new layout is threaded. So maybe that's responsible for the DIFFs, though I then don't understand why my test passed

ambrad:
Ok, thanks. I'll record one thing I believe we all know in case others reading this issue don't: Compiling with -fopenmp and running with 1 thread can give different results than compiling without -fopenmp. One can force a build with -fopenmp (or equivalent flag depending on system/compiler) and run with 1 thread using xmlchange FORCE_BUILD_SMP=TRUE, xmlchange NTHRDS=1.

rljacob:
@amametjanov try changing the new layout to use 20 nodes with no threads.  That should fit.

jonbob:
@ambrad - I agree, but was running that test because it's one that Az showed had DIFFed for him. It is in the test suite so we have blessed baselines with DEBUG=TRUE. Otherwise I agree that the -fopenmp flag would account for the differences between the non-threaded baselines and the threaded results from this PR

amametjanov:
Like @jonbob mentioned, ne30 cases changed to threaded builds at 4x32x2 (nodes x mpi x omp): and that explains DIFFs (not repro-sums). ne4 cases remain at 3x32x2.

I think the consensus is to keep `e3sm_integration` and `e3sm_prod` tests with similar PE-layouts as before: keep MPI-only, if previous was pure-MPI; and keep OCN parallel, if previous PE-layout was parallel (not stacked like they are now).
I'll work on that. There are 113 cases: might take some time.

rljacob:
But did #5014 create real diffs (not just namelist)?  I didn't see any bless requests from it.

amametjanov:
No, I marked #5014 as `[NML]`, but it didn't produce NML diffs, because the PR that actually runs the test with custom PEs is E3SM-Project/E3SM_test_scripts#59, which hasn't been merged yet (to-be-done after this PR is updated and merged).

amametjanov:
Rebased the PR onto latest master and re-ran `integration` and `prod` with pureMPI for WCYCL runs on ne30 and RRM grids:
- e3sm_prod: https://my.cdash.org/viewTest.php?buildid=2180715
- e3sm_integration: https://my.cdash.org/viewTest.php?buildid=2180711

The DIFFs are now showing up only in mpaso and mpassi diagnostics files like
- mpaso.hist.am.globalStats.0001-01-01.nc: in SMS and ERP tests
- mpassi.hist.am.regionalStatistics.0001.01.nc: in PEM and PET tests

Everything else is either BFB or has expected namelist-diffs.

jonbob (approved):
Looks great -- thanks @amametjanov

PR #5032, opened 2022-06-17
Update config_files.xml

fhyuancn:
Fix the path issue of the compset specification file for elm component.


rljacob:
@fhyuancn was this causing an error?

rljacob:
@amametjanov what problem did this fix?

amametjanov:
I think it was just a harmless typo. @fhyuancn, right?

fhyuancn:
Yes. It was a typo in the ELM path. @amametjanov @rljacob

PR #5052, opened 2022-06-28
Added field to sea ice carbon conservation

njeffery:
-Computes and outputs total organic carbon to be consistent with coupler output.
-Updates registry format of carbon conservation to new standard.
-Removes redundant SHR_CONST_MOL_MASS_C in favor of SHR_CONST_MWC to be consistent with other components

BFB


akturner (approved):
Tested with standalone tests and verified conservation am runs.

jonbob:
merged to master

PR #5059, opened 2022-07-01
Add grid support for the ne30pg2_oRRS30to10v3 resolution

jonbob:
Brings in all necessary changes to run the ne30pg2_oRRS30to10v3 resolution in a fully-coupled configuration, including new spunup one/ice initial conditions. All new files are staged on the inputdata repo.

### DIFF
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -4411,8 +4411,8 @@
     </gridmap>
 
     <gridmap ocn_grid="oRRS30to10v3" rof_grid="r05">
-      <map name="ROF2OCN_ICE_RMAPNAME">cpl/cpl6/map_r05_to_oRRS30to10v3_smoothed.r150e300.171109.nc</map>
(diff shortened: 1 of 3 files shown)


rljacob:
I assume this is ready?

jonbob:
@rljacob - I can commit it as-is, but we're also testing it with namelist changes that are necessary in order to actually run. The oRRS30to10v3 grid requires a smaller timestep than the coupling interval for the ne30, so I'm trying to see if ocn/ice settings will allow it to run stably

rljacob:
status:  still running tests to get parameters set.

rljacob:
status:  may not be needed.  Need to talk to @vanroekel

rljacob:
status:  may be closed.  New grids being generated.  @vanroekel should weigh in.

rljacob:
telecon notes:  still checking if this is needed.

rljacob:
A new PR will be made.

Issue #5057, opened 2022-06-30
libfabric runtime error with pm-cpu with high MPI pressure


ndkeen:
On pm-cpu, which uses slingshot11, I see a runtime error with various cases (ATM-only, G-cases, etc) that I have found happens when there is enough pressure on MPI. The same error occurs with other NERSC apps and HPE thinks the issue might be "hardware running out of buffers for messages that arrive before the receiver is ready for them". There is a work-around which is to this buffering in software via `setenv FI_CXI_RX_MATCH_MODE software`. This does work and allows the jobs to run, but can be slower. The degree of performance decrease may depend on how much buffering is needed for the case (assuming this theory is correct). HPE is aware of the issue and NERSC is speculating they might have a fix that will do this buffering in hardware.

```
 135: libfabric:47445:1656625588:cxi:core:cxip_ux_onload_cb():2444<warn> RXC (0xe910:7:0): Flow control triggered due to failure to append LE. Software endpoint mode required.
 135: 
 135: Program received signal SIGABRT: Process abort signal.
 135: 
 135: Backtrace for this error:
 135: #0  0x149cb1ba6d6f in ???
 135: #1  0x149cb1ba6cdb in ???
 135: #2  0x149cb1ba8374 in ???
 135: #3  0x149cb13010da in cxip_ux_onload_cb
 135:   at prov/cxi/src/cxip_msg.c:2444
 135: #4  0x149cb12f14bb in cxip_cq_eq_progress
 135:   at prov/cxi/src/cxip_cq.c:512
 135: #5  0x149cb12f14bb in cxip_cq_progress
 135:   at prov/cxi/src/cxip_cq.c:549
 135: #6  0x149cb12f1a38 in cxip_util_cq_progress
 135:   at prov/cxi/src/cxip_cq.c:562
 135: #7  0x149cb12cfbe5 in ofi_cq_readfrom
 135:   at prov/util/src/util_cq.c:230
 135: #8  0x149cb3ec6e41 in ???
 135: #9  0x149cb3efaeb5 in ???
 135: #10  0x149cb3efbc75 in ???
 135: #11  0x149cb5c3700b in ???
 135: #12  0x178ca8d in __m_spmdutils_MOD_m_swapm_fp
 135:   at /pscratch/sd/n/ndk/wacmy/m22-jun22/externals/mct/mct/m_SPMDutils.F90:839
 135: #13  0x1789aa7 in __m_rearranger_MOD_rearrange_
[...]

PR #5061, opened 2022-07-05
For pm-cpu, add env variable to change how libfabric handles MPI buffers as work-around

ndkeen:
For pm-cpu, add env variable to change how libfabric handles MPI buffers. 
temporarily use software instead of hardware as a work-around until HPE network
engineers have a fix. Adding this variable allow several tests to work that were
otherwise failing with runtime error.

Fixes https://github.com/E3SM-Project/E3SM/issues/5057

[bfb]


ndkeen:
All tests pass with `e3sm_developer` except for a test already known to fail COMPARE. None of the tests would be expected to need this variable -- I've only seen the issue with higher MPI pressure. I did test with  `SMS_P1024_Ld1.ne120pg2_r0125_oRRS18to6v3.F2010-CICE.pm-cpu_gnu` which I think will fail without this env variable.

ndkeen:
merged to next

PR #5064, opened 2022-07-07
Add machine alvarez (similar to pm-cpu)

ndkeen:
Alvarez is a test machine at NERSC where newer features for pm-cpu are tested. Only limited access to machine, but makes working with this resource easier. For example, this commit includes latest libfabric version as well as minor updates to netcdf/hdf modules. However, the latest cray-mpich module (`8.1.17` seems to be causing hangs when writing certain output, so using what we use on `pm-cpu`, which is `8.1.15`).

Improved a few PE layouts for alvarez that are often used in tests (such as those in `e3sm_integration`). Will do similar for `pm-cpu`.

Also removed incorrect compiler name from the pm-cpu config (we use `amdclang` and not `aocc` for AMD compiler).

[bfb]


ndkeen:
merged to next

PR #5068, opened 2022-07-09
Refactor array access in gllfvremap for FIDEAL builds with IBM

amametjanov:
Refactor array access in gllfvremap for FIDEAL builds with IBM.

[BFB]


ambrad (approved):

ambrad:
@amametjanov I approved the PR and removed the others as reviewers since gllfvremap_mod is my code. I understand why you made the change: Q isn't big enough in the FIDEAL compset. Thanks!

PR #5070, opened 2022-07-09
For pm-cpu, improve PE layouts commonly used for testing

ndkeen:
For pm-cpu, improve PE layouts commonly used for testing.

[bfb]


ndkeen:
Already tested with alvarez, but also ran e3sm_integration here.

ndkeen:
merged to next

PR #5071, opened 2022-07-09
Add subversion to ascent env

jgfouca:
Needed for download files from CESM (not always needed, but is useful since this is a fairly new machine).

[BFB]

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -3717,6 +3717,7 @@
       <modules>
         <command name="purge"/>
         <command name="load">DefApps</command>
+        <command name="load">python/3.8-anaconda3</command>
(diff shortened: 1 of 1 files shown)


amametjanov (approved):

jgfouca:
Thanks for the quick review over the weekend, @amametjanov ! I noticed one additional problem a couple minutes ago so had to make a second commit if you want to review it.

amametjanov (approved):

jgfouca:
Oof, I'm not sure what's going. With the python module removed, my scream cases are failing the RUN phase with that same YAML error, probably because the `module purge` messes up the python env. It seems like the only thing that works is putting it back and making sure the user also has the same module loaded when in their shell. I guess all I need for now is subversion.

PR #5074, opened 2022-07-11
Update mask/transect files for oRRS30to10v3 and oRRS18to6v3

jonbob:
Brings in updated mask/transect analysis files used for for mpaso meshes oRRS30to10v3 and oRRS18to6v3. The new files are staged on the inputdata server.

[BFB]


vanroekel:
Here are a couple figures, first 30to10

image

The red dots are the defined atlantic region, black is the old transect file and blue is the new one, they are now aligned in the new file.

Here is 18to6

image

These plots are based on the files referenced in the PR and in the input data repository.

vanroekel (approved):
Looks great and the files are as expected.  Thanks @jonbob!

jonbob:

jonbob:
merged to master

PR #5082, opened 2022-07-17
Update compiler environment on Crusher from 8.2.0 to 8.3.3

grnydawn:
Updated compiler environment on Crusher from PrgEnv-8.2.0 to PrgEnv-8.3.3. Also,
- cray compiler version: cce/14.0.0
- amd compiler version: amd/5.1.0
- gnu compiler version: gcc/11.2.0 (PrgEnv default)
- removed "-O1 -hfp0" flags that was used for crayclang
- changed slurm settings for Crusher from "--cpu_bind=cores -m plane={{ tasks_per_node }}" to "--threads-per-core=1 >-m *:block" because previous settings cause failures on all test cases.
- turned off inlining of the statement of "numlvl = get_block_lvl_cnt_d(blockids(jb),bcids(jb))" in src/physics/cam/phys_grid.F90

e3sm_integration test result with Cray compiler: 47 passed and 58 failed.

While many test cases are failed with Cray compiler, we need this PR because previous compiler environment(8.2.0) is not available on Crusher anymore.

[BFB]


components/eam/src/physics/cam/phys_grid.F90
```diff
@@ -1098,7 +1098,9 @@ subroutine phys_grid_init( )
              if (iam == owner_d) then
                 if (.not. associated(btofc_blk_offset(blockids(jb))%pter)) then
                    blksiz = get_block_gcol_cnt_d(blockids(jb))
+!DIR$ NOINLINE
```
sarats, line 1101: You probably need to add an ifdef guard as this issue is specific to Crusher and Cray compiler IIRC.
This would be an interim fix until the Cray compiler fix arrives in a newer release.

minxu74:
@sarats will you approve this PR since @grnydawn has addressed your concerns?

amametjanov:
I would just merge it to next and let each compiler run a round of tests on next to see any remaining issues.
Then, if nothing new comes up, merge to master.

grnydawn:
@amametjanov, @minxu74, @sarats . I think I will merge this pr with next branch and test it with all three compilers. I will let you know once I finish the runs(it may takes more than three days as Crusher is quite busy these days)

grnydawn:
@amametjanov , @minxu74 , @sarats . Following is the test results after merging this PR with "next" branch from using "crayclang", "amdclang", and "gnu". Overall, the errors from using the three compilers are not much different from what we have experienced from previous testings.

- crayclang: pass(57), fail(49), similar errors from previous tests, floating point arithmetic errors are still major issues
- amdclang: pass(18), fail(88), similar errors from previous tests, long command-line argument issue is still major issue
- gnu: pass(87), fail(21), shr_reprosum_calc ERROR: NaNs or INFs in input(9), HDF5 version mismatch(6), memleak(3)

Detail test results can be found here: https://docs.google.com/spreadsheets/d/1I0J9zUXCJufdnlxLSkXaww-w81ME4RLI_0cxKRh6a5U/edit?usp=sharing

cime_config/machines/config_machines.xml
```diff
@@ -600,12 +600,16 @@
       <executable>srun</executable>
       <arguments>
         <arg name="num_tasks"> -l -K -n {{ total_tasks }} -N {{ num_nodes }} </arg>
-        <arg name="binding">--cpu_bind=cores</arg>
+        <!-- <arg name="binding">cpu_bind=cores</arg>  -->
```
sarats, line 603: @grnydawn Are you planning on updating the MAX_TASKS_PER_NODE to allow for more OMP threads? If so, can you check if process binding is okay with these flags in that scenario (MPI+OpenMP)?

grnydawn: @sarats I will try to update the MAX_TASKS_PER_NODE and will also check the process binding.

grnydawn: @sarats Changing MAX_TASKS_PER_NODE from 64 to 128 did request only one node instead of previous two nodes. In addition, I also had to add -c 2 flag for "salloc" job allocation to prevent resource error. Now, I will investigate further how to apply these settings to E3SM configuration files.

PR #5090, opened 2022-07-20
Initial implementation on LANL IC machine chicoma-cpu

jonbob:
Adds support for E3SM on the LANL IC machine chicoma-cpu. This initial implementation only includes the gnu/mpich combination, though other compilers are available and will be added later.

[BFB]


jonbob:
@philipwjones, @mark-petersen -- please feel free to assign others to test and review. I was just confident that you both had access to chicoma

philipwjones (changes requested):
With the comments I've noted here, I was able to run a GMPAS QU240 case using create_newcase.  The changes I've asked for assumes that t22_e3sm-gpu_g is the only current project we have, so ignore these if another project exists on the machine. And we'll want to change back whenever we get a real allocation later.

Also, create_test does not appear to work - it fails building support libs because the netcdf doesn't seem be be loaded correctly (complains about unknown netcdf fortran functions). create_newcase and subsequent case.build works fine, so not sure what's being done differently in create_test.

cime_config/machines/config_batch.xml
```diff
@@ -514,6 +514,16 @@
 	</queues>
     </batch_system>
 
+   <batch_system MACH="chicoma-cpu" type="slurm">
+    <directives>
+     <directive>--partition=standard </directive>
```
philipwjones, line 519: Current project only has access to gpu partition so --partition=gpu needed here unless there is another project besides t22_e3sm-gpu_g

cime_config/machines/config_batch.xml
```diff
@@ -514,6 +514,16 @@
...
+     <directive>--qos=standard </directive>
+    </directives>
+    <queues>
+      <queue walltimemax="00:30:00" nodemax="560" default="true">standard</queue>
```
philipwjones, line 523: Similarly, this needs to be:
queue walltimemax="00:30:00" nodemax="96" default="true">gpu</queue>

cime_config/machines/config_machines.xml
```diff
@@ -3133,6 +3133,114 @@
     </environment_variables>
   </machine>
 
+  <machine MACH="chicoma-cpu">
+    <DESC>Chicoma CPU-only nodes at LANL IC. Each node has 2 AMD EPYC 7H12 64-Core (Milan) 512GB</DESC>
+    <NODENAME_REGEX>ch-fe.*.lanl.gov</NODENAME_REGEX>
```
philipwjones, line 3138: I needed to change this to 
ch-fe*
since the fully qualified hostname the system returned was ch-fe#.LOCAL for some reason

cime_config/machines/config_machines.xml
```diff
@@ -3133,6 +3133,114 @@
...
+    <OS>Linux</OS>
+    <COMPILERS>gnu,nvidia,aocc,amdclang</COMPILERS>
+    <MPILIBS>mpich</MPILIBS>
+    <PROJECT>t22_e3sm-gpu</PROJECT>
```
philipwjones, line 3142: Project name is actually
t22_e3sm-gpu_g

cime_config/machines/config_batch.xml
xylar: We will add `chicoma-gpu` later but the idea here was to start with CPU (more familiar to most of us).  We have several projects I believe with Chicoma CPU time (e.g. `t22_ocean_time_step`).

cime_config/machines/config_batch.xml
xylar: I think `nodemax` here is right for the CPU side.

cime_config/machines/config_machines.xml
xylar: It sounds like we need to use `t22_ocean_time_step` or something else for the CPU side.  Then, `t22_e3sm-gpu_g` when we do `chicoma-gpu`.

cime_config/machines/config_batch.xml
philipwjones: ok, we should keep this then as long as we change the project name

cime_config/machines/config_batch.xml
philipwjones: Same with this - we should keep the 560 and standard queue with a different project name

cime_config/machines/config_machines.xml
```diff
@@ -3133,6 +3133,114 @@
...
+    <SUPPORTED_BY>e3sm</SUPPORTED_BY>
+    <MAX_TASKS_PER_NODE>256</MAX_TASKS_PER_NODE>
+    <MAX_MPITASKS_PER_NODE>64</MAX_MPITASKS_PER_NODE>
+    <PROJECT_REQUIRED>TRUE</PROJECT_REQUIRED>
+    <mpirun mpilib="default">
+      <executable>srun</executable>
+      <arguments>
+        <arg name="label"> --label</arg>
+        <arg name="num_tasks"> -n {{ total_tasks }} -N {{ num_nodes }}</arg>
+        <arg name="thread_count">-c $SHELL{echo 256/`./xmlquery --value MAX_MPITASKS_PER_NODE`|bc}</arg>
```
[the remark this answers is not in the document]
jonbob: @philipwjones - what about binding? We have this:
````
arg name="binding"> $SHELL{if [ 128 -ge `./xmlquery --value MAX_MPITASKS_PER_NODE` ]; then echo "--cpu_bind=cores"; else echo "--cpu_bind=threads";fi;} </arg>
````
but 128 is obviously wrong

File: cime_config/machines/config_machines.xml (hunk 4 above)
Comment (jonbob) [reply cime_config/machines/config_machines.xml:3142]: Do we want to leave PROJECT unset for this PR? Some machines, like theta, require a PROJECT but do not set one by default

Comment (philipwjones) [reply cime_config/machines/config_machines.xml:3142]: That might be the best solution, rather than trying to create an inclusive IC project with no resources.

File: cime_config/machines/config_machines.xml (hunk 5 above)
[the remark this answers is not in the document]
Comment (philipwjones) [reply cime_config/machines/config_machines.xml:3162]: Well...the standard CPU partition does have 2 CPUs with 64 cores each so I think 128 is correct to detect hyper threading here.  The gpu partition only has a single 64-core cpu so the gpu partition should be 64.

File: cime_config/machines/config_machines.xml (hunk 4 above)
Comment (xylar) [reply cime_config/machines/config_machines.xml:3142]: Yes, I agree.

**Comment by xylar** [comment]
@jonbob, I will try to re-test this tomorrow.  Do you want to rebase to get rid of the conflict in `config_pesall.xml`?

**Comment by jonbob** [comment]
Thanks for pointing this out, @xylar. I'll rebase it today so hopefully there is a clean codebase for you to try in Germany's tomorrow

File: cime_config/machines/config_machines.xml [hunk 6]
```diff
@@ -3251,6 +3251,113 @@
     </environment_variables>
   </machine>
 
+  <machine MACH="chicoma-cpu">
+    <DESC>Chicoma CPU-only nodes at LANL IC. Each node has 2 AMD EPYC 7H12 64-Core (Milan) 512GB</DESC>
+    <NODENAME_REGEX>ch-fe*</NODENAME_REGEX>
+    <OS>Linux</OS>
+    <COMPILERS>gnu,nvidia,aocc,amdclang</COMPILERS>
+    <MPILIBS>mpich</MPILIBS>
+    <CIME_OUTPUT_ROOT>/lustre/scratch4/turquoise/$ENV{USER}/E3SM/scratch/chicoma-cpu</CIME_OUTPUT_ROOT>
```
xylar, line 3260:
```suggestion
    <CIME_OUTPUT_ROOT>/lustre/scratch4/turquoise/$ENV{USER}/e3sm_scratch/chicoma-cpu</CIME_OUTPUT_ROOT>
```
For now, I think the `E3SM/scratch` and `E3SM/input_data` directories are okay but once we move to a common input directory, it might be nice to be more consistent with other E3SM machines.

xylar (approved):
I successfully ran 5-day tests for `ne4_oQU480` and `ne30pg2_EC30to60E2r2`:
```
/lustre/scratch4/turquoise/xylar/E3SM/scratch/chicoma-cpu/20220810.TEST.ne4_oQU480.chicoma/run
/lustre/scratch4/turquoise/xylar/E3SM/scratch/chicoma-cpu/20220810.TEST.more_pes.ne30pg2_EC30to60E2r2.chicoma/run
```
I fumbled around the timing files and it seems like PIO is taking a reasonable amount of time here (it isn't in my standalone testing, so that's why I was making sure).  That's with the caveat that I don't really know what I'm looking for.

This is ready to go in as far as I can tell.

philipwjones (approved):
All my comments have been resolved so I will approve.

jonbob:

jonbob:
merged to master -- thanks for all the testing help @xylar and @philipwjones

PR #5093, opened 2022-07-23
Update Summit config in E3SMv1

amametjanov:
This enables high-res F- and G-compset benchmarks in maint-1.0.

[BFB]


amametjanov:
Still waiting for some more ne120_ne120 + FC5AV1C-H01A benchmarking runs in queue. If they run to completion, I'll remove the "Do not merge" label.

PR #5105, opened 2022-08-02
Enable COSP for MMF

brhillman:
Enable COSP for MMF configurations. This involved a massive refactor of the COSP interface in `cospsimulator_intr.F90`. I think a lot of this refactor would be great to use for non-MMF configurations as well, but since the changes were so large and a lot of MMF-specific details are included, it was decided to move the refactored copy of the interface to the MMF (`crm`) directory to continue to keep the MMF-specific code isolated from the rest of the model. Note that the refactored interface in `components/eam/src/physics/crm/cospsimulator_intr.F90` is still compatible with non-MMF configurations if we wanted to move it to `components/eam/src/physics/cam/` (and in fact non-MMF COSP outputs are BFB even with the changes), but would need to either build `crmdims.F90` or define dummy CRM dimensions some other way. Climate is BFB with this PR, but will require baseline blesses since the FRCE-MMF1 test is changed to use the COSP testmod.

[BFB]

### DIFF
--- a/components/eam/src/physics/crm/cospsimulator_intr.F90
+++ b/components/eam/src/physics/crm/cospsimulator_intr.F90
@@ -1994,6 +1994,8 @@ CONTAINS
         call packed_average(nc, nx, ny, cospOUTpacked%isccp_meanalbedocld, cospOUT%isccp_meanalbedocld)
         call packed_average(nc, nx, ny, cospOUTpacked%isccp_meanptop     , cospOUT%isccp_meanptop     )
         call packed_average(nc, nx, ny, cospOUTpacked%isccp_meantaucld   , cospOUT%isccp_meantaucld   )
+        call packed_average(nc, nx, ny, cospOUTpacked%isccp_meantb       , cospOUT%isccp_meantb       )
+        call packed_average(nc, nx, ny, cospOUTPacked%isccp_meantbclr    , cospout%isccp_meantbclr    )
       end if
 
       ! CALIPSO SIMULATOR OUTPUTS
--- a/components/eam/src/physics/crm/cospsimulator_intr.F90
+++ b/components/eam/src/physics/crm/cospsimulator_intr.F90
@@ -1940,6 +1940,7 @@ CONTAINS
    end subroutine cosp_write_outputs
 #endif
 
+#ifdef USE_COSP
     subroutine packed_average1d(nc, nx, ny, d_in, d_out)
       integer, intent(in) :: nc, nx, ny
       real(r8), intent(in) :: d_in(:)
@@ -2027,6 +2028,8 @@ CONTAINS
         end do
       end do
     end subroutine
+#endif
+
 #ifdef USE_COSP
    subroutine cosp_unpack_outputs(nc, nx, ny, cospINpacked, cospOUTpacked, cospIN, cospOUT)
       use mod_cosp, only: cosp_optical_inputs, cosp_outputs
--- a/components/eam/src/physics/crm/cospsimulator_intr.F90
+++ b/components/eam/src/physics/crm/cospsimulator_intr.F90
@@ -2696,7 +2696,7 @@ CONTAINS
              y%v_sfc(npoints),y%lat(npoints),y%lon(nPoints),y%emis_sfc(nchan),           &
              y%cloudIce(nPoints,nLevels),y%cloudLiq(nPoints,nLevels),y%surfelev(nPoints),&
              y%fl_snow(nPoints,nLevels),y%fl_rain(nPoints,nLevels),y%seaice(npoints),    &
-             y%tca(nPoints,nLevels),y%hgt_matrix_half(npoints,nlevels+1))
+             y%tca(nPoints,nLevels),y%hgt_matrix_half(npoints,nlevels))
 
   end subroutine construct_cospstateIN
   ! ######################################################################################
--- a/components/eam/src/physics/crm/cospsimulator_intr.F90
+++ b/components/eam/src/physics/crm/cospsimulator_intr.F90
@@ -42,9 +42,8 @@ module cospsimulator_intr
        cloudsat_preclvl
     use mod_cosp_stats,       only: cosp_change_vertical_grid
 #endif
-#ifdef MMF_PACK
-    use crmdims, only: crm_nx_rad, crm_ny_rad, crm_nz
-#endif
+  use crmdims, only: crm_nx_rad, crm_ny_rad, crm_nz
+
   implicit none
   private
   save
@@ -251,11 +250,7 @@ module cospsimulator_intr
        gamma_3 = (/-1._r8, -1._r8,      2.0_r8,      2.0_r8, -1._r8, -1._r8,      2.0_r8,      2.0_r8,      2.0_r8/),&
        gamma_4 = (/-1._r8, -1._r8,      6.0_r8,      6.0_r8, -1._r8, -1._r8,      6.0_r8,      6.0_r8,      6.0_r8/)       
 #endif
-#ifndef MMF_PACK
-  integer, parameter :: crm_nx_rad = 1
-  integer, parameter :: crm_ny_rad = 1
-  integer, parameter :: crm_nz     = 1
-#endif
+
   logical :: use_MMF
 
 CONTAINS
--- a/components/eam/src/physics/crm/cospsimulator_intr.F90
+++ b/components/eam/src/physics/crm/cospsimulator_intr.F90
@@ -173,7 +173,7 @@ module cospsimulator_intr
                                                  ! (0=ice-spheres ; 1=ice-non-spherical) (0)
   integer, parameter :: overlap = 3              ! overlap type: 1=max, 2=rand, 3=max/rand (3)
 
-  !! namelist variables for COSP input related to ISCCP simulator
+  ! namelist variables for COSP input related to ISCCP simulator
   integer :: isccp_topheight = 1                 ! 1 = adjust top height using both a computed infrared
                                                  ! brightness temperature and the visible
                                                  ! optical depth to adjust cloud top pressure.
@@ -316,7 +316,7 @@ CONTAINS
     ! Set number of sub-columns, from namelist
     nscol_cosp = Ncolumns_in
     
-    if (use_vgrid_in) then    !! using fixed vertical grid
+    if (use_vgrid_in) then  ! using fixed vertical grid
        if (csat_vgrid_in)       then
           nht_cosp = 40
        else
@@ -376,8 +376,7 @@ CONTAINS
     ! read in the namelist
     if (masterproc) then
        unitn = getunit()
-       open( unitn, file=trim(nlfile), status='old' )  !! presumably opens the namelist file "nlfile"
-       !! position the file to write to the cospsimulator portion of the cam_in namelist
+       open( unitn, file=trim(nlfile), status='old' )
        call find_group_name(unitn, 'cospsimulator_nl', status=ierr)   
        if (ierr == 0) then
           read(unitn, cospsimulator_nl, iostat=ierr)
@@ -466,7 +465,7 @@ CONTAINS
        cosp_nradsteps = 3
     end if
     
-    !! if no simulators are turned on at all and docosp is, set cosp_amwg = .true.
+    ! if no simulators are turned on at all and docosp is, set cosp_amwg = .true.
     if((docosp) .and. (.not.lradar_sim) .and. (.not.llidar_sim) .and. (.not.lisccp_sim) .and. &
          (.not.lmisr_sim) .and. (.not.lmodis_sim)) then
        cosp_amwg = .true.
@@ -611,7 +610,7 @@ CONTAINS
     
     ! ISCCP OUTPUTS
     if (lisccp_sim) then
-       !! addfld calls for all
+       ! addfld calls for all
        call addfld('FISCCP1_COSP',(/'cosp_tau','cosp_prs'/),'A','percent', &
             'Grid-box fraction covered by each ISCCP D level cloud type',&
             flag_xyfill=.true., fill_value=R_UNDEF)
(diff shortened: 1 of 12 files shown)


components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -227,7 +227,17 @@ subroutine crm_physics_register()
       call pbuf_add_field('CRM_QT',    'global', dtype_r8,dims_crm_3D,idx)
       call pbuf_add_field('CRM_QP',    'global', dtype_r8,dims_crm_3D,idx)
       call pbuf_add_field('CRM_QN',    'global', dtype_r8,dims_crm_3D,idx)
+      ! CRM effective radii - probably only used by COSP. See note below.
+      call pbuf_add_field('CRM_REL' , 'physpkg', dtype_r8,dims_crm_3D,idx)
```
whannah1, line 231: Should we add these lines outside of this if block?
Or will they need to be different variables for P3?

components/eam/src/physics/crm/rrtmgp/radiation.F90
```diff
@@ -1220,6 +1222,8 @@ subroutine radiation_tend(state_in,ptend,    pbuf,          cam_out, cam_in,  &
          mu, lambdac, dei, des, rei, rel
       real(r8), dimension(pcols,pver), target :: zeros
       real(r8), pointer, dimension(:,:,:,:) :: crm_t, crm_qv, crm_qc, crm_qi, crm_cld, crm_qrad
+      real(r8), pointer, dimension(:,:,:,:) :: &
+         crm_rel, crm_rei, crm_dtau, crm_emis
```
whannah1, line 1226: Seems unnecessary and inconsistent to use the line continuation here.

whannah1:
@brhillman looks like we still need to add a test with the COSP test mod to one of the suites.

components/eam/src/physics/crm/crm_physics.F90
brhillman: Probably should move outside this block. I think it was originally this way because we were computing effective radii differently for 2-moment, but for COSP I think this is fine. Note that this only affects the radar simulator I think, and maybe the lidar simulator.

brhillman:
> @brhillman looks like we still need to add a test with the COSP test mod to one of the suites.

Fixed, I set the FRCE-MMF1 test to use the new `eam-cosp_nhtfrq9` test mod (output every 9 steps, since this is a 9-step ERS test).

components/eam/src/physics/crm/cospsimulator_intr.F90
```diff
@@ -0,0 +1,3184 @@
...
+  ! Default namelists
+  ! The CAM and COSP namelists defaults are set below.  Some of the COSP namelist 
+  ! variables are part of the CAM namelist - they all begin with "cosp_" to keep their 
+  ! names specific to COSP. I set their CAM namelist defaults here, not in namelist_defaults_cam.xml
+  !  Variables identified as namelist variables are defined in
+  !  ../models/atm/cam/bld/namelist_files/namelist_definition.xml
+  ! ######################################################################################
+  ! CAM
+  logical :: cosp_amwg             = .false. ! CAM namelist variable default, not in COSP namelist
+  logical :: cosp_lite             = .false. ! CAM namelist variable default, not in COSP namelist
```
whannah1, line 120: So "CAM" variables that have "cosp" in the name?
I think we can remove or edit the "CAM" section header here, and delete the redundant line comments with  "CAM namelist variable default"

components/eam/src/physics/crm/cospsimulator_intr.F90
```diff
@@ -0,0 +1,3184 @@
...
+       !*cfMon,cfDa* tclisccp (time,profile), CFMIP wants "gridbox mean cloud cover from ISCCP"
+       call addfld('CLDTOT_ISCCP', horiz_only,'A','percent', &
+            'Total Cloud Fraction Calculated by the ISCCP Simulator ',flag_xyfill=.true., fill_value=R_UNDEF)
+       !*cfMon,cfDa* albisccp (time,profile)
+       ! Per CFMIP request - weight by ISCCP Total Cloud Fraction (divide by CLDTOT_ISSCP in history file to get weighted average)
+       call addfld('MEANCLDALB_ISCCP',horiz_only,'A','1','Mean cloud albedo*CLDTOT_ISCCP',flag_xyfill=.true., fill_value=R_UNDEF)
+       !*cfMon,cfDa* ctpisccp (time,profile)
+       ! Per CFMIP request - weight by ISCCP Total Cloud Fraction (divide by CLDTOT_ISSCP in history file to get weighted average)
+       call addfld('MEANPTOP_ISCCP',horiz_only,'A','Pa','Mean cloud top pressure*CLDTOT_ISCCP',flag_xyfill=.true., &
+            fill_value=R_UNDEF)
```
whannah1, line 628: Might be nice to have a "cosp_history" module for this whole section of addfld, and the later outfld calls

components/eam/src/physics/crm/cospsimulator_intr.F90
```diff
@@ -0,0 +1,3184 @@
...
+       ! meantbclrisccp (time,profile)
+       call addfld ('MEANTBCLR_ISCCP',horiz_only,'A','K','Mean Clear-sky Infrared Tb from ISCCP simulator',   &
+            flag_xyfill=.true., fill_value=R_UNDEF)
+       ! boxtauisccp (time,column,profile)
+       call addfld ('TAU_ISCCP',(/'cosp_scol'/),'I','1','Optical Depth in each Subcolumn',flag_xyfill=.true., fill_value=R_UNDEF)
+       ! boxptopisccp (time,column,profile)
+       call addfld ('CLDPTOP_ISCCP',(/'cosp_scol'/),'I','Pa','Cloud Top Pressure in each Subcolumn',  &
+            flag_xyfill=.true., fill_value=R_UNDEF)
+
+       !! add all isccp outputs to the history file specified by the CAM namelist variable cosp_histfile_num
```
whannah1, line 645: I see a lot of these double/triple bang comments, it seems some past developer was inconsistent with their formatting!

components/eam/src/physics/crm/cospsimulator_intr.F90
```diff
@@ -0,0 +1,3184 @@
...
+       call addfld('CLDLOW_CAL_ICE',horiz_only,'A','percent','Calipso Low-level Ice Cloud Fraction',             &
+            flag_xyfill=.true., fill_value=R_UNDEF)
+       ! lcllcalipsoliq (time,loc)
+       call addfld('CLDLOW_CAL_LIQ',horiz_only,'A','percent','Calipso Low-level Liquid Cloud Fraction',          &
+            flag_xyfill=.true., fill_value=R_UNDEF)
+       ! lcllcalipsoun (time,loc) !+cosp1.4
+       call addfld('CLDLOW_CAL_UN',horiz_only,'A','percent','Calipso Low-level Undefined-Phase Cloud Fraction',  &
+            flag_xyfill=.true., fill_value=R_UNDEF)
+    
+!       ! Calipso Opaque/thin cloud diagnostics
```
whannah1, line 739: Any particular reason why these are commented out?

components/eam/src/physics/crm/cospsimulator_intr.F90
```diff
@@ -0,0 +1,3184 @@
...
+    !         to _init functions in cosp_init.
+    ! DS2019: Add logicals, default=.false., for new Lidar simuldators (Earthcare (atlid) and ground-based
+    !         lidar at 532nm)
+    call COSP_INIT(Lisccp_sim, Lmodis_sim, Lmisr_sim, Lradar_sim, Llidar_sim, LgrLidar532, &
+         Latlid, Lparasol_sim, Lrttov_sim, radar_freq, k2, use_gas_abs, do_ray,              &
+         isccp_topheight, isccp_topheight_direction, surface_radar, rcfg_cloudsat,           &
+         use_vgrid_in, csat_vgrid_in, Nlr_in, pver, cloudsat_micro_scheme)
+
+    ! Set number of sub-columns, from namelist
+    nscol_cosp = Ncolumns_in
```
whannah1, line 317: Are these subcolumns the same as the subcolumn feature at the physics levels? or is this something internal to COSP?

components/eam/src/physics/crm/cospsimulator_intr.F90
```diff
@@ -0,0 +1,3184 @@
...
+       reff_cosp(1:ncol,1:pver,I_LSSNOW) = ls_reffsnow(1:ncol,1:pver)*1.e-6_r8  ! (calculated in cldwat2m_micro.F90, passed to stratiform.F90)
+       reff_cosp(1:ncol,1:pver,I_CVCLIQ) = cv_reffliq(1:ncol,1:pver)*1.e-6_r8   ! (calculated in stratiform.F90, not actually used in radiation)
+       reff_cosp(1:ncol,1:pver,I_CVCICE) = cv_reffice(1:ncol,1:pver)*1.e-6_r8   ! (calculated in stratiform.F90, not actually used in radiation)
+       reff_cosp(1:ncol,1:pver,I_CVRAIN) = ls_reffrain(1:ncol,1:pver)*1.e-6_r8  ! (same as stratiform per Andrew)
+       reff_cosp(1:ncol,1:pver,I_CVSNOW) = ls_reffsnow(1:ncol,1:pver)*1.e-6_r8  ! (same as stratiform per Andrew)
+       reff_cosp(1:ncol,1:pver,I_LSGRPL) = 0._r8                                ! (using radar default reff)
+    end if  ! use_MMF
+
+    ! Set optical depth and emissivity
+    if (use_MMF) then
```
whannah1, line 1624: I noticed you left in this logic to change behavior for MMF. The more I look at this I think it kinda makes sense to have a MMF specific version, for the same reasons we have a separate radiation interface. We might need logic blocks like this anyway when we get to the final implementation of P3, so removing this non-MMF stuff might help clean it up down the line.

whannah1 (approved):
I added some comments in the COSP interface, but I'm fine leaving those issues for a future PR to clean things up. I agree that the size of that file could be considerably reduced, but in the interest of getting these INCITE Cess experiments going I vote we move this along.

components/eam/src/physics/crm/rrtmgp/radiation.F90
```diff
@@ -1387,9 +1403,16 @@ subroutine radiation_tend(state_in,ptend,    pbuf,          cam_out, cam_in,  &
...
+               ! Calculate effective radius for optics used with single-moment microphysics and for COSP
+               if (use_MMF) then 
                   call cldefr(state%lchnk, ncol, state%t, rel, rei, state%ps, state%pmid, landfrac, icefrac, snowh)
+                  do iz = 1,crm_nz
+                     do ic = 1,ncol
+                        ilay = pver - iz + 1
+                        crm_rel(ic,ix,iy,iz) = rel(ic,ilay)
+                        crm_rei(ic,ix,iy,iz) = rei(ic,ilay)
+                     end do
+                  end do
```
whannah1, line 1415: Just noticed an error here from my test run last night. We're missing the loops for ix and iy.

brhillman: Yup good call, this section was moved in a recent commit (and isn't actually used by the simulators we are looking at). Fixed now.

components/eam/src/physics/crm/cospsimulator_intr.F90
brhillman: Note that this IS the MMF specific version. I actually wanted to keep it generic for the time being though, so that I could easily move these changes over to the non-MMF version as well if we decide to do so. I vote we remove these logicals in a subsequent PR depending on our needs for P3, but I can strip this down here if you insist.

components/eam/src/physics/crm/cospsimulator_intr.F90
brhillman: Internal to COSP.

components/eam/src/physics/crm/cospsimulator_intr.F90
brhillman: Copied from the original COSP interface. Apparently whoever implemented this originally didn't want to enable these diagnostics, but probably copied them over when going through everything available? I'm not sure.

components/eam/src/physics/crm/cospsimulator_intr.F90
brhillman: Indeed.

components/eam/src/physics/crm/cospsimulator_intr.F90
brhillman: The whole namelist handling here really needs to be revamped I think. There's a lot of redundant and conflicting options in here that overwrite others.

components/eam/src/physics/crm/cospsimulator_intr.F90
whannah1: sounds like a good plan.

components/eam/src/physics/crm/cospsimulator_intr.F90
brhillman: Note that revamping this would require also revamping the non-MMF version, because they share namelist options. We can certainly do this, but might be better to do in a subsequent PR for expediency?

whannah1: Agreed.

components/eam/src/physics/crm/cospsimulator_intr.F90
brhillman: These should be removed now.

components/eam/src/physics/crm/cospsimulator_intr.F90
brhillman: Removed commented out sections here.

brhillman:
Merged to next.

brhillman:
Re-merged to next following bug fix.

PR #5106, opened 2022-08-02
C-based harvest in ELM-FATES

sshu88:
This pull request contains modification in E3SM corresponding to another FATES [PR #888](https://github.com/NGEET/fates/pull/888).
In detail, necessary input to enable FATES C-based harvest from ELM is passed to FATES through `bc_in`. Output from FATES used to calculate land use emission and net biosphere productivity is passed to ELM through `bc_out`.


rljacob:
The PR description should describe the code changes and not just point to another PR description.

rljacob:
Needs action and review from @sshu88 @rgknox @bishtgautam

rljacob:
Still needs action and review from  @sshu88 @rgknox @bishtgautam

rljacob:
telecon notes:  Waiting for a FATES update before removing WIP.

components/elm/src/biogeophys/BalanceCheckMod.F90
```diff
@@ -616,7 +616,10 @@ subroutine ColWaterBalanceCheck( bounds, num_do_smb_c, filter_do_smb_c, &
        found = .false.
        do p = bounds%begp, bounds%endp
           if (veg_pp%active(p)) then
-             if ( (errsol(p) /= spval) .and. (abs(errsol(p)) > 1.e-7_r8) ) then
+             !if ( (errsol(p) /= spval) .and. (abs(errsol(p)) > 1.e-7_r8) ) then
+             ! solar radiation balance error is high when running FATES
+             ! adjust the threshold to 5.e-7
+             if ( (errsol(p) /= spval) .and. (abs(errsol(p)) > 5.e-7_r8) ) then
```
glemieux, line 622: @sshu88 is this change to try and avoid the water balance error?  @rgknox and I recommend that we not change this in e3sm master branch.  It should be reverted to the old value.

glemieux (changes requested):

components/elm/src/biogeophys/BalanceCheckMod.F90
rgknox, line 622: If we bump this value up to suppress warnings, that seems fine for a FATES run.  We are very aware that an audit needs to be done to reduce these fine errors in radiation scattering.  However, we may want to only apply it to a fates governed patch.  This "is_fates" can tell us if we are on a fates patch and if we should apply different logic: 
https://github.com/E3SM-Project/E3SM/blob/v2.0.0/components/elm/src/data_types/VegetationType.F90#L70

sshu88: I have revised the code as suggested. Thank you @rgknox @glemieux .

rljacob:
telecon notes:  still waiting for FATES PR.

rljacob:
@rgknox this is waiting on a FATES PR.

components/elm/src/dyn_subgrid/dynHarvestMod.F90
```diff
@@ -66,22 +65,23 @@ module dynHarvestMod
...
+
   ! the units flag must match the units of harvest_varnames
   ! set this here because dynHarvest_init is called after alm_fates%init
   ! this flag is accessed only if namelist do_harvest is TRUE
-  integer, public          :: wood_harvest_units = hlm_harvest_area_fraction
-  
-  type(dyn_var_time_uninterp_type) :: harvest_vars(num_harvest_vars)   ! value of each harvest variable
 
-  real(r8) , allocatable, public   :: harvest_rates(:,:) ! category harvest rates (d1) in each gridcell (d2)
+  integer, public, parameter    :: wood_harvest_units = 2    ! 1 = area fraction, 2 = carbon
```
rgknox, line 81: @sshu88, is the long-term plan to have wood_harvest_units read in or defined in the file? or set in a namelist?

components/elm/src/main/elmfates_interfaceMod.F90
```diff
@@ -369,8 +372,13 @@ subroutine ELMFatesGlobals2()
 
      ! We will use this switch temporarily, until  we complete
      ! the ELM-FATES harvest integration
-     logical, parameter :: do_elm_fates_harvest = .false.
-
+     logical, parameter :: do_elm_fates_harvest = .true.
```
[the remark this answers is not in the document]
rgknox: nevermind, i see it isn't used. Can we remove this variable and its commented out usage below?

components/elm/src/main/elmfates_interfaceMod.F90
```diff
@@ -937,6 +951,7 @@ subroutine dynamics_driv(this, bounds_clump, top_as_inst,          &
             this%fates(nc)%bc_in(s)%hlm_harvest_catnames = harvest_varnames
             this%fates(nc)%bc_in(s)%hlm_harvest_units = wood_harvest_units
          end if
+         this%fates(nc)%bc_in(s)%site_area=col_pp%wtgcell(c)*grc_pp%area(g)*1e6_r8
```
rgknox, line 954: need units for 1e6 please. I'm guessing this is m2 per km2?

components/elm/src/main/elmfates_interfaceMod.F90
```diff
@@ -2264,6 +2285,46 @@ end subroutine wrap_accumulatefluxes
...
+    do icc = 1,fc
+       c = filterc(icc)
+       s = this%f2hmap(nc)%hsites(c)
+
+       ! Shijie: Pass harvested wood products to ELM variable
+       hrv_deadstemc_to_prod10c(c)  = this%fates(nc)%bc_out(s)%hrv_deadstemc_to_prod10c
+       hrv_deadstemc_to_prod100c(c) = this%fates(nc)%bc_out(s)%hrv_deadstemc_to_prod100c
+
+       ! Pass LUC related C fluxes which are calculated in FATES [gC m-2 s-1]
+       gpp(c) = this%fates(nc)%bc_out(s)%gpp_site*1e3
```
rgknox, line 2317: units for 1e3 please, named constant is good too

components/elm/src/data_types/ColumnDataType.F90
```diff
@@ -5763,6 +5773,34 @@ subroutine col_cf_init(this, begc, endc, carbon_type)
            call hist_addfld1d (fname='HR', units='gC/m^2/s', &
                 avgflag='A', long_name='total heterotrophic respiration', &
                  ptr_col=this%hr)
+! Shijie
```
rgknox, line 5776: remove comment

components/elm/src/data_types/ColumnDataType.F90
```diff
@@ -5763,6 +5773,34 @@ subroutine col_cf_init(this, begc, endc, carbon_type)
...
+               avgflag='A', long_name='net ecosystem production, excludes fire, landuse, and harvest flux, positive for sink', &
+                ptr_col=this%nep)
+ 
+          this%nbp(begc:endc) = spval
+          call hist_addfld1d (fname='NBP', units='gC/m^2/s', &
+               avgflag='A', long_name='net biome production, includes fire, landuse, and harvest flux, positive for sink', &
+                ptr_col=this%nbp)
+ 
+          this%nee(begc:endc) = spval
+          call hist_addfld1d (fname='NEE', units='gC/m^2/s', &
```
rgknox, line 5798: will NEE also be defined further down? see https://github.com/E3SM-Project/E3SM/blob/e96fc25b2d223d21530fd78ab5212feac1ecf44c/components/elm/src/data_types/ColumnDataType.F90#L6121

rgknox: same with NEP and NBP, if already defined somewhere else that is active, lets use the same definition

rgknox (changes requested):
THanks and nice work @sshu88 , I made some comments and requests. THings look pretty straightforward.

components/elm/src/dyn_subgrid/dynHarvestMod.F90
sshu88: In my opinion it would be better to set `wood_harvest_units` as a namelist option. However, this option may require the change of another set of harvest rate data in different unit.

components/elm/src/dyn_subgrid/dynHarvestMod.F90
```diff
@@ -181,7 +182,8 @@ subroutine dynHarvest_interp_harvest_types(bounds)
           ! means that harvest rates will be maintained at the rate given in the last
           ! year of the file for all years past the end of this specified time series.
           do_harvest = .true.
-          allocate(this_data(bounds%begg:bounds%endg))   
+          ! Right now we don't account for the topounit in plant harvest
```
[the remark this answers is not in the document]
sshu88: True. The dataset we are using now does not account for another topographic dimension.

components/elm/src/main/elmfates_interfaceMod.F90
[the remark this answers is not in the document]
sshu88: Yes. Remove this namelist option does not affect the ELM-FATES run now.

components/elm/src/main/elmfates_interfaceMod.F90
sshu88: Right.

components/elm/src/main/elmfates_interfaceMod.F90
sshu88: g per kg.

components/elm/src/data_types/ColumnDataType.F90
sshu88: This repetitive definition is actually in a different IF-ELSE blocks (.not. FATES .and. C12). I'm not sure how to merge them?

rgknox: ok, I thought that was an end if, I see its an else if now.  I have no problems with the current formulation then.  An alternative would be to move all instances of NEP, NEE and NBP to either the very beginning of the routine, or the very end, which would be outside the if fates (and wrap with ane if c12).

bishtgautam:
@sshu88 You have mistakenly updated multiple submodules in this PR which need to be reverted.

sshu88:
@bishtgautam @rgknox I have reverted those pointers now. Thanks.

components/elm/src/dyn_subgrid/dynHarvestMod.F90
aldivi: I agree that a namelist option would be best in the long-term, particularly if the input file eventually contains both carbon and area prescriptions.

bishtgautam:
@sshu88 Thanks for fixing the submodules. Is this PR still WIP or ready to be reviewed/merged or waiting for another FATES PR to go first?

cc: @glemieux @rgknox

components/elm/src/data_types/ColumnDataType.F90
```diff
@@ -6948,9 +6985,13 @@ subroutine col_cf_summary(this, bounds, num_soilc, filter_soilc, isotope)
 
        ! total product loss
        this%product_closs(c) = &
-            this%prod10c_loss(c)  + &
+            this%prod10c_loss(c) + &
```
bishtgautam, line 6988: Please undo the deletion of space

bishtgautam (commented):
I have requested minor changes.

rgknox (approved):

bishtgautam:
@sshu88 Is this WIP or ready to be merged?

sshu88:
@bishtgautam We can go ahead and merge the work once the FATES part is merged.

glemieux:
> @bishtgautam We can go ahead and merge the work once the FATES part is merged.

@bishtgautam I'm working on the final deconflict and test on the fates-side right now.

rgknox:
The changes just merged into this branch, in https://github.com/E3SM-Project/E3SM/pull/5106/commits/13d0f5a71ebff055ddd3155a13ebcdd69da43ce7 , were originally overlooked, but nonetheless required for compatibility with fates API 25.

components/elm/src/main/elmfates_interfaceMod.F90
```diff
@@ -464,17 +469,23 @@ subroutine ELMFatesGlobals2()
...
+              pass_harvest_bypass_criteria = 0
+           end if
         else
            pass_lu_harvest = 0
            pass_num_lu_harvest_types = 0
+           pass_harvest_bypass_criteria = 0
         end if
 
         call set_fates_ctrlparms('use_lu_harvest',ival=pass_lu_harvest)
+        call set_fates_ctrlparms('use_harvest_bypass_criteria',ival=pass_harvest_bypass_criteria)
```
glemieux, line 488: I don't see any setup for this on the fates side pr 888.  Is this a future feature?  I'm assuming so given that it's hard coded to false just above this.

glemieux (commented):
One question about the hardcoded `allow_harvest_bypass_criteria`

components/elm/src/main/elmfates_interfaceMod.F90
sshu88: It's an old flag. I forgot to delete this line in ELM side.

components/elm/src/main/elmfates_interfaceMod.F90
```diff
@@ -155,7 +155,7 @@ module ELMFatesInterfaceMod
    use FatesPlantHydraulicsMod, only : InitHydrSites
    use FatesPlantHydraulicsMod, only : RestartHydrStates
 
-   use dynHarvestMod          , only : num_harvest_vars, harvest_varnames
+   use dynHarvestMod          , only : num_harvest_vars, harvest_varnames, wood_harvest_units
    use dynHarvestMod          , only : harvest_rates ! these are dynamic in space and time
    use dynHarvestMod          , only : num_harvest_vars, harvest_varnames, wood_harvest_units
```
glemieux, line 160: @sshu88 it looks like there is some duplicate code here as well.

glemieux (approved):
Just one more minor piece of code cleanup.

components/elm/src/main/elmfates_interfaceMod.F90
sshu88: Thanks for checking.

glemieux:
@bishtgautam we still need to update the pointer to the fates side, but the associated PR hasn't been merged yet.  I'll give you a heads up when that happens

bishtgautam:
Are the existing tests cover the new physics? Or, should we add a new test?

glemieux:
> Are the existing tests cover the new physics? Or, should we add a new test?

We should add new tests, but I would like to defer that to a future PR in which we review and update all of the fates' tests for the land developer suite.

components/elm/src/data_types/ColumnDataType.F90
```diff
@@ -2090,6 +2090,16 @@ subroutine col_cs_init(this, begc, endc, carbon_type, ratio, c12_carbonstate_var
...
+       this%prod10c(begc:endc) = spval
+       call hist_addfld1d (fname='PROD10C', units='gC/m^2', &
+            avgflag='A', long_name='10-yr wood product C', &
+            ptr_col=this%prod10c, default='inactive')
+
+       this%prod100c(begc:endc) = spval
+       call hist_addfld1d (fname='PROD100C', units='gC/m^2', &
+            avgflag='A', long_name='100-yr wood product C', &
+            ptr_col=this%prod100c, default='inactive')
+
```
glemieux, line 2102: @sshu88 was this a temporary change to make sure that you could read the `prod10c` and `prod100c` for when fates was running?

glemieux (commented):
I noticed one change that affects non-fates runs that needs to be addressed.

components/elm/src/data_types/ColumnDataType.F90
sshu88: These are two outputs I added to the history output file for checking the change of product pool size. So it is a temporary change. But the product pool size shall be a basic output in land surface model when considering anthropogenic disturbances.

glemieux: To clarify, does this need to be available when fates is running?  The current issue is that the same call is made just below behind a `.not. use_fates` check, which means this will get called twice and cause a build failure for non-fates runs.  If it's not necessary for fates right now, would you mind removing these lines?  Otherwise, the part that is behind the `.not. use_fates` check could be removed.

sshu88: I have removed the part below `.not. use_fates` check.

glemieux:
I'm running the land-developer tests and seeing two unexpected failures.  I haven't looked into them much yet, but one is a fates smoke test failing during runtime in PhosphorusDynamicsMod and the other is a model build failure for `ERS.ne30pg2_r05_EC30to60E2r2.GPMPAS-JRA.cori-haswell_intel.mosart-rof_ocn_2way`.

folder location (Cori): 
```
/global/homes/g/glemieux/cori-scratch/e3sm-tests/pr5106-merge.fates.cori-haswell.intel.Ce0716a090f-F25047167
```

Note I'm using Cori as perlmutter is down right now.

rgknox:
I submitted some updates to @sshu88 's branch to resolve the issues with the fates_eca test.  Tests were passing on Cori

glemieux:
@sshu88 I created a [PR to your branch](https://github.com/sshu88/E3SM/pull/6) to update the fates pointer now that https://github.com/NGEET/fates/pull/888 is integrated on the fates-side.

@bishtgautam I ran land developer tests on Cori and all the tests except `ERS.ne30pg2_r05_EC30to60E2r2.GPMPAS-JRA.gcp12_gnu.mosart-rof_ocn_2way` passed (due to https://github.com/E3SM-Project/E3SM/issues/5412).

glemieux:
@bishtgautam this shoudl be closed per #5429.

Issue #5111, opened 2022-08-04
OOM with ERS.hcru_hcru.I20TRGSWCNPRDCTCBC.cori-knl_intel.elm-erosion


ndkeen:
cdash reporting a fail with `ERS.hcru_hcru.I20TRGSWCNPRDCTCBC.cori-knl_intel.elm-erosion` that is running out of memory.

Likely needs more nodes via PE change similar to https://github.com/E3SM-Project/E3SM/pull/5098

PR #5113, opened 2022-08-05
On cori-knl, add better PE layout for grid l%360x720cru, 16 nodes, 64x2

ndkeen:
Only on cori-knl, add better PE layout for grid l%360x720cru, 16 nodes, 64x2

Fixes https://github.com/E3SM-Project/E3SM/issues/5111

[bfb]

### DIFF
--- a/components/elm/cime_config/config_pes.xml
+++ b/components/elm/cime_config/config_pes.xml
@@ -288,6 +288,31 @@
         </ntasks>
       </pes>
     </mach>
+    <mach name="cori-knl">
+      <pes compset="any" pesize="any">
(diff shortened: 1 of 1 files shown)


ndkeen:
merged to next

ndkeen:
I tested this, but I've been waiting for the cdash tests to run again on cori-knl. It looks like they typically run every 4 days, but this is the 5th. I pinged @wadeburgess who said the jenkins job has been running for 20+ hours. There are no actual e3smtest jobs in the Q. I created helpdesk ticket, but just noting here.

ndkeen:
cori-knl reported and this specific test passed, but we had 3 other tests `NOT RUN`.  So I ran those 3 tests to make sure no problem.

PR #5116, opened 2022-08-08
Homme: Some minor tweaks related to print to screen

bartgol:
Namely:

- allow `iulog` to be settable if `SCREAM` is defined, rather than hard coded to 6 (stdout);
- replaced a couple of print to screen with `write(iulog, *)`;
- allow to pass output stream to `SimulationParams::print`.

Plus, fixed compiler warning related to `SimualtionParams` members initialization order.

`homme_integration` and `e3sm_developer` passed on Chrysalis.

[BFB]


components/homme/src/share/kinds.F90
```diff
@@ -29,12 +29,19 @@ module kinds
...
   log_kind     = 4,                      &
   real_kind    = 8,                      &
-  iulog        = 6,                      & ! stderr file handle
 #if HOMME_QUAD_PREC
   longdouble_kind    = 16
 #else 
   longdouble_kind    = 8
 #endif
+
+#ifdef SCREAM
```
ambrad, line 38: Here, you're in the case of !CAM and SCREAM. Does this include a CIME-based build or just a standalone SCREAM build? If it includes a CIME-based build, shouldn't we using the CAM block above?

bartgol: Well, this should include _any_ SCREAM build, cime or not cime.

I tried to execute the CAM block (without iulog, since SCREAM deos not build cam_logfile...although it could, since it's basically nothing), but then I had issues b/c QUAD_PREC was enabled and I wasn't getting the longdouble kind. I could disable QUAD_PREC when building homme within scream, I guess, and then we could just use the 1st block.

ambrad: Ok. No changes needed. I'm a little concerned that we're not using shr_kind_mod values when building under CIME, but that's orthogonal to this PR.

bartgol: Yeah, that's a good point. This is how it was before, so I figured I didn't touch that, but it is a bit concerning. I can add an extra commit addressing that.

bartgol: Ok, I changed it. Gotta retest now.

components/homme/src/share/kinds.F90
```diff
@@ -3,38 +3,56 @@
...
+#else
+  ! default to output unit to stdout, but can be changed at runtime
+  integer (kind=4), public :: iulog = output_unit
+#endif
+
+  ! QUAD precision can be enabled via CMake var (or compiler macro for EAM).
+  ! Currently, EAM does not use quad precision, while tandalone Homme/SCREAM can.
+  integer (kind=4), public, parameter::  &
+#ifdef HOMME_QUAD_PREC
+  longdouble_kind    = 2*real_kind
```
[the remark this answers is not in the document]
bartgol: True

components/homme/src/share/kinds.F90
```diff
@@ -3,38 +3,56 @@
...
 ! 
-#if defined(CAM)  
+#if defined(CAM) || defined(SCREAM)
   integer (kind=4), public, parameter::  &
   int_kind     = SHR_KIND_I4,            &
   log_kind     = kind(.true.),           &
   long_kind    = SHR_KIND_I8,            &
   real_kind    = SHR_KIND_R8,            &
-  longdouble_kind    = 8
-  public :: shr_kind_cl, iulog
```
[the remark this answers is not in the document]
bartgol: It was never used in either EAM or standalone Homme, so I thought I removed it. I can bring it back.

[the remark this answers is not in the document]
ambrad: No, you're right. Once I narrowed my grep to homme/src, I see that it's not used.

bartgol:
`homme_integration` and `e3sm_developer` both pass on chrysalis.

bartgol:
Retested on chrysalis after the last two commits, and the two testsuites still pass.

ambrad:
Is this ready for merging?

bartgol:
Yes, I'm no longer touching it.

ambrad (approved):

ambrad:
I'll merge this on Monday if next is open then.

components/homme/src/share/schedule_mod.F90
```diff
@@ -207,7 +207,7 @@ subroutine genEdgeSched(elem, PartNumber,LSchedule,MetaVertex)
        endif
     enddo
     if(iam == 1) then 
-       print *,'getMetaSchedule: tmpP: ',tmpP
+       write(iulog, *),'getMetaSchedule: tmpP: ',tmpP
```
amametjanov, line 210: Crusher's crayclang is reporting a typo here for `SMS_D_Ln5.conusx4v1_r05_oECv3.F2010.crusher_crayclang`:
```
ftn-1725 ftn: ERROR GENEDGESCHED, File = ../../../gpfs/alpine/cli133/proj-shared/testing/E3SM/components/homme/src/share/schedule_mod.F90, Line = 210, Column = 23
  Unexpected syntax while parsing the WRITE statement : "operand" was expected but found ",".

Cray Fortran : Version 14.0.0 (20220420160222_c98838affc7b58fed2a72f164d77c35e1bc8772f)
Cray Fortran : Compile time:  0.0158 seconds
Cray Fortran : 470 source lines
Cray Fortran : 2 errors, 0 warnings, 0 other messages, 0 ansi
```

bartgol: Ah, yeah, I've seen that error with very pedantic compilers in the past. I can remove the extra comma in a second Homme PR i'm working on.

PR #5117, opened 2022-08-09
Support for single process single cell standalone test cases

### LINKED ISSUE #5118: MPAS-Seaice lacks single process idealized test cases
MPAS-Seaice lacks idealized single cell test cases for individual thermodynamic processes. This functionality would be useful for unit testing of the sea ice component processes

### LINKED ISSUE #5119: Sea ice conservation standalone tests dont mimic ocean interaction in E3SM
In standalone mode the sea ice model doesn't include the ocean mixed layer in conservation checks, and the mixed layer model represents the old POP ocean model, not the current MPAS-Ocean model interaction.

akturner:
Added several options for controlling which thermodynamic processes are active
Added option for conservation AM to include ocean heat, mass, energy content in analysis with mixed layer model
Added another mixed layer ocean model that mimics ocean model in E3SM for standalone testing
Added another forcing option to support idealized single cell test case

Fixes #5118
Fixes #5119

[NML]
[BFB]

### DIFF
--- a/components/mpas-seaice/src/shared/mpas_seaice_column.F
+++ b/components/mpas-seaice/src/shared/mpas_seaice_column.F
@@ -222,6 +222,9 @@ contains
        ! initialize the column package parameters
        call init_column_package_parameters(domain, ciceTracerObject)
 
+       ! initialize active column processes
+       call init_column_active_processes(domain)
+
     endif
 
   end subroutine seaice_init_column_physics_package_parameters
@@ -1325,10 +1328,7 @@ contains
          config_use_aerosols, &
          config_use_prescribed_ice, &
          config_use_snow_liquid_ponds, &
-         config_use_high_frequency_coupling, &
-         config_use_congelation_basal_melt, &
-         config_use_lateral_melt, &
-         config_use_latent_processes
+         config_use_high_frequency_coupling
 
     ! dimensions
     integer, pointer :: &
@@ -1528,9 +1528,6 @@ contains
        call MPAS_pool_get_config(block % configs, "config_use_prescribed_ice", config_use_prescribed_ice)
        call MPAS_pool_get_config(block % configs, "config_use_snow_liquid_ponds", config_use_snow_liquid_ponds)
        call MPAS_pool_get_config(block % configs, "config_use_high_frequency_coupling", config_use_high_frequency_coupling)
-       call MPAS_pool_get_config(block % configs, "config_use_congelation_basal_melt", config_use_congelation_basal_melt)
-       call MPAS_pool_get_config(block % configs, "config_use_lateral_melt", config_use_lateral_melt)
-       call MPAS_pool_get_config(block % configs, "config_use_latent_processes", config_use_latent_processes)
 
        call MPAS_pool_get_dimension(mesh, "nCellsSolve", nCellsSolve)
        call MPAS_pool_get_dimension(mesh, "nCategories", nCategories)
@@ -1886,10 +1883,7 @@ contains
                dayOfYear, &
                abortFlag, &
                abortMessage, &
-               config_use_prescribed_ice, &
-               config_use_congelation_basal_melt, &
-               config_use_lateral_melt, &
-               config_use_latent_processes)
+               config_use_prescribed_ice)
           call column_write_warnings(abortFlag)
 
           ! code abort
@@ -5443,12 +5437,10 @@ contains
            landIceMask
 
     logical, pointer :: &
-         config_use_test_ice_shelf, &
-         config_use_latent_processes
+         config_use_test_ice_shelf
 
     call MPAS_pool_get_config(domain % configs, "config_dt", config_dt)
     call MPAS_pool_get_config(domain % configs, "config_ocean_mixed_layer_type", config_ocean_mixed_layer_type)
-    call MPAS_pool_get_config(domain % configs, "config_use_latent_processes", config_use_latent_processes)
 
     block => domain % blocklist
     do while (associated(block))
@@ -5531,8 +5523,7 @@ contains
                   latentTransferCoefficient, &
                   sensibleTransferCoefficient, &
                   airDragCoefficient(iCell), &
-                  airOceanDragCoefficientRatio(iCell), &
-                  latent_processes_active=config_use_latent_processes)
+                  airOceanDragCoefficientRatio(iCell))
 
              albedoVisibleDirectOcean(iCell)  = seaiceOceanAlbedo
              albedoIRDirectOcean(iCell)       = seaiceOceanAlbedo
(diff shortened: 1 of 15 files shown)


components/mpas-seaice/src/shared/mpas_seaice_forcing.F
```diff
@@ -1181,6 +1186,8 @@ subroutine init_oceanic_forcing(domain)
     select case (trim(config_forcing_sst_type))
     case ("ncar")
        call init_oceanic_forcing_ncar(domain)
+    case ("ncar_free_surface")
+       call init_oceanic_forcing_ncar_free_surface(domain)
```
eclare108213, line 1190: I think the nomenclature here is incorrect.  POP is a free-surface ocean model, although it is constant volume.  So the original ncar forcing data is usable for a "free surface" ocean model.

components/mpas-seaice/src/shared/mpas_seaice_column.F
```diff
@@ -5557,6 +5647,310 @@ subroutine seaice_column_ocean_mixed_layer(domain)
...
+!
+!  seaice_ocean_mixed_layer_e3sm
+!
+!> \brief E3SM ocean coupling proxy
+!> \author Adrian K. Turner, LANL
+!> \date 9th July 2022
+!> \details Representation of ocean processes and coupling in e3sm. This
+!>  routine mimics the ocean/sea ice coupling in e3sm using MPAS-Ocn,
+!>  including updates to ocean thickness, temperature and salinity from
+!>  interactions with sea ice and frazil formation.
```
eclare108213, line 5660: Is there a (design) document somewhere that explains these differences in true detail?  We will need that for the Icepack documentation.

components/mpas-seaice/src/shared/mpas_seaice_column.F
```diff
@@ -1875,7 +1881,10 @@ subroutine column_vertical_thermodynamics(domain, clock)
                dayOfYear, &
                abortFlag, &
                abortMessage, &
-               config_use_prescribed_ice)
+               config_use_prescribed_ice, &
+               config_use_congelation_basal_melt, &
+               config_use_lateral_melt, &
+               config_use_latent_processes)
```
eclare108213, line 1887: What is the computational burden of adding 'if' conditionals for all of these individual processes?  They have to be parsed for each category, on each grid cell with ice, perhaps multiple times per timestep.

rljacob:
@akturner please respond to review comments.

components/mpas-seaice/src/shared/mpas_seaice_column.F
akturner: @eclare108213: Good point. I have recoded this so that there are no conditionals now, just multiplication by a real flag taking values (0,1). I will test performance of these changes as well

components/mpas-seaice/src/shared/mpas_seaice_forcing.F
akturner: Good point, I have renamed the option to som_varying_thickness

components/mpas-seaice/src/shared/mpas_seaice_column.F
akturner: No design document. This is an additional subroutine in MPAS rather than the column package so doesn't necessarily need to be in icepack. Making this proxy was a good way to understand how frazil is currently done and I plan to try a better formulation so this may change soon anyway.

akturner:
Single cell test case for idealized processes added here: https://github.com/MPAS-Dev/MPAS-Seaice_standalone_framework/pull/13

eclare108213 (approved):
Thanks for getting rid of the conditionals - this is a more elegant approach in my opinion and perhaps not as computationally impactful.  The title of this pull request does not fully represent what is in it -- the new conservation checks are particularly important.  I'm curious to know whether these tests have turned up anything surprising or needing to be fixed, but answering that question is not necessary for this PR.  Approving.

rljacob:
telecon: waiting on review from @alicebarthel who is out of town.

alicebarthel:
I ran a G-case 1-year test with this code. It ran successfully and the sea ice fluxes look consistent visually (see attached plots). My automated check technically failed but I think it may be an issue with the test itself. Looking into this now. 
seaice-ocean_fluxcomparison_cpl_am 20221016 test GMPAS-JRA singlecellsiPR TL319_EC30to60E2r2 anvil_hbudget_NH_1-1
seaice-ocean_fluxcomparison_cpl_am 20221016 test GMPAS-JRA singlecellsiPR TL319_EC30to60E2r2 anvil_hbudget_SH_1-1
seaice-ocean_fluxcomparison_cpl_am 20221016 test GMPAS-JRA singlecellsiPR TL319_EC30to60E2r2 anvil_hbudget_GL_1-1

alicebarthel:
My only suggestion at this stage would be to change the streams.seaice file to replace: 
`clobber_mode="append"` with `clobber_mode="truncate"`
to match other streams (and avoid errors popping up when restarting/overwriting output files)

eclare108213:
What are the units on the y-axes?  E.g. I'd expect the radiative fluxes to be ~50x larger in magnitude.

alicebarthel:
@eclare108213 These fluxes are all in W/m2. Note that these are only the sea-ice ones. The atm/ocn radiative fluxes are much bigger (~+/-150 for net SW and ~+/-200-300 for LW). 
@akturner Can you remind us if there is area scaling: e.g. are we summing over the sea ice area then dividing by the whole hemisphere area? That could explain the order of magnitude difference. I have vague memories of discussing this with you a while back.

akturner:
@alicebarthel: Which streams file are you referring to?

eclare108213:
But they are for a single-cell column configuration.  Can they be made consistent with that view of the "world"?

alicebarthel:
@akturner I meant the conservationCheck AM in streams.seaice. The code I checked out had them in "append" mode instead of "truncate" mode which means errors popped up when trying to overwrite the AM output files. Looking at the files, it does not seem to be something introduced by your changes, so I am happy to approve the PR.
@eclare108213 The test case I ran here is a "normal" (global) G-case, not the single-column mode. Adrian had asked me to check introducing these changes did not modify the standard e3sm flux check I have worked on.
@eclare108213 does that address your concern or do we need to double-check what a "hemispheric scaling" would look like in single-column mode? Is the idea to have the same sea-ice vs. coupler check for flux consistency in a single-column test? In that case, I don't know how the coupler would handle writing out the SH and NH fluxes...

eclare108213:
I see, thanks for the clarification @alicebarthel.  Eventually we would like to look at MPAS-SI output in realistic magnitudes for at least one sea ice grid point in each hemisphere, but others are working on that problem now.

rljacob:
telecon notes: needs a rebase and build system fixes.

jonbob:
@akturner - the rebase and bld update look good. I'll test today and hopefully merge soon

jonbob:
passes:

with expected NML DIFFs

merged to next

jonbob:

PR #5120, opened 2022-08-10
[MPAS-Framework] Add a new `name_in_output` registry attribute to variables

xylar:
This can be used to set a different name in output streams than the name in code or the unique name required for input. 

This merge also adds corresponding `outputFieldName` and `outputConstituentNames` to each field. These new attributes will be used to allow different names in output than in the code.

 The `name_in_output` or the `outputFieldName` attribute must only be used for pure output variables.  Since it doesn't change the name read from input files, renaming variables in an output file that also gets used as an input files (such as a restart file or initial condition) would cause unexpected results and should be avoided.

Prior to this merge, the output name must be unique, but this causes trouble in time mean output, particularly for coordinate variables. Coordinate variables are usually variables with the exact same name as their dimension.  For example, we plan to add a CF-compliant `Time` coordinate to supplement `xtime`.  The `Time` would become `timeMonthly_avg_Time` in the current version of the time-series-stats monthly average analysis member.  But this is no longer a proper (or convenient) coordinate variable.  It still needs to be called `Time` in the monthly output files.  That is, it needs to have the same name as another variable in the code on purpose.

Similarly, `temperature` becomes `timeMonthly_avg_activeTracers_temperature` in time-series-stats monthly average.  But with these changes, this one could be renamed to just `temperature` in the output, which would be a _lot_ easier to work with.  That kind of simplification is not yet planned but could be a nice medium-term project.

[BFB]

### DIFF
--- a/components/mpas-framework/src/framework/mpas_field_types.inc
+++ b/components/mpas-framework/src/framework/mpas_field_types.inc
@@ -18,6 +18,9 @@
       ! Information used by the I/O layer
       character (len=StrKIND) :: fieldName
       character (len=StrKIND), dimension(:), pointer :: constituentNames => null()
+      ! outputFieldName and outputConstituentNames are alternative names for the field or its constituents for use in
+      ! output files.  These should only be given names different from fieldName or constituentNames for pure output
+      ! variables such as diagnostic output and never for variables present in initial conditions or restart files.
       character (len=StrKIND) :: outputFieldName
       character (len=StrKIND), dimension(:), pointer :: outputConstituentNames => null()
       character (len=StrKIND), dimension(5) :: dimNames
--- a/components/mpas-framework/src/framework/mpas_field_types.inc
+++ b/components/mpas-framework/src/framework/mpas_field_types.inc
@@ -18,6 +18,8 @@
       ! Information used by the I/O layer
       character (len=StrKIND) :: fieldName
       character (len=StrKIND), dimension(:), pointer :: constituentNames => null()
+      character (len=StrKIND) :: outputFieldName
+      character (len=StrKIND), dimension(:), pointer :: outputConstituentNames => null()
       character (len=StrKIND), dimension(5) :: dimNames
       integer, dimension(5) :: dimSizes
       real (kind=RKIND) :: defaultValue
@@ -52,6 +54,8 @@
       ! Information used by the I/O layer
       character (len=StrKIND) :: fieldName
       character (len=StrKIND), dimension(:), pointer :: constituentNames => null()
+      character (len=StrKIND) :: outputFieldName
+      character (len=StrKIND), dimension(:), pointer :: outputConstituentNames => null()
       character (len=StrKIND), dimension(4) :: dimNames
       integer, dimension(4) :: dimSizes
       real (kind=RKIND) :: defaultValue
@@ -87,6 +91,8 @@
       ! Information used by the I/O layer
       character (len=StrKIND) :: fieldName
       character (len=StrKIND), dimension(:), pointer :: constituentNames => null()
+      character (len=StrKIND) :: outputFieldName
+      character (len=StrKIND), dimension(:), pointer :: outputConstituentNames => null()
       character (len=StrKIND), dimension(3) :: dimNames
       integer, dimension(3) :: dimSizes
       real (kind=RKIND) :: defaultValue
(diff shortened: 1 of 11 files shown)


xylar:
See https://github.com/E3SM-Ocean-Discussion/E3SM/pull/23 for additional discussion.

xylar:
Follow-up work based on this PR will add `Time` and `Time_bnds` coordinate variables to MPAS-Ocean, MPAS-Seaice and MALI output for better CF compliance.

xylar:
@akturner, @cbegeman, and @mark-petersen, you are free to test further if you want but you may also approve based on your review of https://github.com/E3SM-Ocean-Discussion/E3SM/pull/23

@matthewhoffman, I'm mostly wanting to make sure this does no harm to MALI.  If there are tests you would like me to run, let me know.

cbegeman (approved):

mark-petersen (approved):
Approving based on my testing and comments in https://github.com/E3SM-Ocean-Discussion/E3SM/pull/23.

akturner (approved):
Approving based on testing in E3SM-Ocean-Discussion#23.

components/mpas-framework/src/framework/mpas_field_types.inc
```diff
@@ -18,6 +18,11 @@
       ! Information used by the I/O layer
       character (len=StrKIND) :: fieldName
       character (len=StrKIND), dimension(:), pointer :: constituentNames => null()
+      ! outputFieldName and outputConstituentNames are alternative names for the field or its constituents for use in
+      ! output files.  These should only be given names different from fieldName or constituentNames for pure output
+      ! variables such as diagnostic output and never for variables present in initial conditions or restart files.
```
xylar, line 23: @matthewhoffman, are you good with this?

matthewhoffman: Thanks, @xylar .  This is helpful.  I would prefer to also see more info from the discussion above, specifically a reference to the specific use case this was implemented for (I think a specific example helps communicate the functionality), as well as the warning about possible collisions if multiple variables had the same `name_in_output` in the same output stream.  I agree this situation seems too difficult to try to protect against misuse, but I think it's important to document clearly what constitutes misuse.  

One other clarification for me - this new attribute is accessible through Registry, right?  But in this PR you are accessing it programmatically, right?  I was confused by that initially.

xylar: I think a 3 line comment here is already pushing the standard practice. I hesitate to put a lengthy discussion in a file folks are unlikely to look at in detail. I don't want to set the president that this is the "right" place to document things in detail just because we lack proper documentation. If we expand this practice, this file will get out of hand.

xylar: This PR doesn't directly make use of the new functionality either in code or in the register. It just makes sure neither breaks anything. Future, component specific PRs will actually use the functionality.

xylar: If helpful, I might be able to find a little time later this week to have a video chat about this.

matthewhoffman: @xylar , you're right that we don't have a good mechanism for documenting this sort of information.  I've always thought framework should have been better documented, and I'm not opposed to breaking standard practice going forward.  But I'm also ok with this PR serving as the documentation if you prefer to leave things as you have them.

matthewhoffman (approved):

components/mpas-framework/src/framework/mpas_field_types.inc
xylar: Thanks @matthewhoffman.  I fully agree that better framework documentation is needed in general and that this PR isn't a great, easy to find alternative to proper documentation.  Maybe we need a future E3SM hackathon on documenting the framework?

I looked over the comment here and your request again.  I still feel like a longer description here would just get lost.  Until there is a better place for it, I think this PR will have to suffice for documenting the purpose of this new functionality.  Hopefully, the follow-up PRs to add `Time` and `Time_bnds` coordinates will help to demonstrate the capability by example more clearly.

xylar:
Thank you @mark-petersen, @cbegeman, @akturner and @matthewhoffman for taking the time to review this.

@jonbob, I believe this is ready for you to integrate when it's a good time.

jonbob:
Test merge passes:
* SMS_Ld1.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel.allactive-wcprod

merged to next

jonbob:
merged to master

PR #5125, opened 2022-08-10
Change TSC test default to use F2010-CICE

mkstratos:
Use F2010-CICE compset in TSC test

There are problems with short timesteps and MPAS-SI, so we'll revert to this compset until that issue can be fixed.

[non-BFB] only for the TSC test


rljacob (approved):

rljacob:
merged to next.

PR #5127, opened 2022-08-11
Homme: fix ICE

tcclevenger:
In SCREAMv1 runs on summit, with new SurfaceCoupling from https://github.com/E3SM-Project/scream/pull/1780, an ICE arises in homme boundary exchange.


tcclevenger:
Labeling WIP since maybe this is not needed.

oksanaguba:
Just FYI, anything that touches homme needs to go to E3SM, not scream, thanks.

oksanaguba:
Is it correct that the fix is just to break 1 file into 2?

bartgol:
> Is it correct that the fix is just to break 1 file into 2?

Yes. But we're puzzled as of why this ICE showed up while working in the SCREAM PR. We never touched the homme folder there, so it might be a red herring.

oksanaguba:
my comment about scream PR was due to my confusion, sorry.

bartgol:
Sure.

tcclevenger:
This ICE was a symptom of a different issue. Closing.

PR #5128, opened 2022-08-11
removed use of Fortran fraction intrinsic from ocn diagnostics

### LINKED ISSUE #5112: Fortran intrinsic `fraction` has no device function w/ Cray compiler
https://github.com/E3SM-Project/E3SM/blob/0b2ced33b791be3ebb6896872eb01c2ef0428aaa/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F#L1432

The above referenced line of code uses the `fraction()` Fortran intrinsic function.  During the Crusher Hackathon it was determined that during the port of MPAS-O to GPUs w/ OpenACC, this causes a link-time error on Crusher with the Cray compiler.  We believe this is due to there being no device function for `fraction()` in the Cray compiler.

I'm not sure how many other occurrences of this function there are in E3SM.

It wasn't immediately apparent to us what the `fraction` function was doing.  Is the use of this function necessary?  Can the functionality be replaced with something else?

CC'ing others involved in the initial discussion.
@sarats @mrnorman @brobey

### LINKED ISSUE #5126: Incorrect use of Fortran fraction intrinsic in mpas-ocean diagnostic routines
You keep using that word. I do not think it means what you think it means (I. Montoya)...

In the mpas_ocn_diagnostics diagnostic_solve_surface_layer routine, the fraction intrinsic is being used assuming it returns the fractional remainder of a real number. However, the fraction intrinsic returns the fractional value of the machine representation, ie x*radix(x)**(-exponent(x)) which is quite different.  All surface layer averages of tracers and velocity are incorrect using this routine and are then used for some EOS options.

This was discovered during discussion of #5112 and fixing this bug will fix that issue as well.

philipwjones:
The ocean surface layer averaging routine was using a Fortran fraction intrinsic incorrectly, so this replaces the intrinsic with code that computes a fraction correctly. Also did some reformatting for long lines and documenting variables. Because it removes the fraction intrinsic, it fixes another issue in which that instrinsic was not supported on the GPU by the Cray compiler on Crusher.

NOTE: This code computes variables that are never used except to compute surface-displaced density which is also currently not used (and is an expensive EOS calc). This makes this PR bfb but we might consider removing these expensive calculations if really not needed.

[bfb]
Fixes #5126 
Fixes #5112

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
@@ -1334,184 +1334,220 @@ contains
 !>    values at surface layer
 !
 !-----------------------------------------------------------------------
-   subroutine ocn_diagnostic_solve_surfaceLayer(layerThickness, activeTracers, &
-                layerThicknessEdgeFlux, normalVelocity, tracersSurfaceLayerValue,  &
-                indexSurfaceLayerDepth, normalVelocitySurfaceLayer, &
-                surfaceFluxAttenuationCoefficient, surfaceFluxAttenuationCoefficientRunoff)!{{{
 
-      implicit none
+   subroutine ocn_diagnostic_solve_surfaceLayer(layerThickness, &
+                activeTracers, layerThicknessEdgeFlux, normalVelocity, &
+                tracersSurfaceLayerValue, indexSurfaceLayerDepth, &
+                normalVelocitySurfaceLayer, &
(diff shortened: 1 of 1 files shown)


mark-petersen (approved):
Passes nightly suite for gnu debug and intel debug, and matches previous bfb with gnu optimized. Thanks for the note on the surface displaced density. I'll take a look at that.

jonbob:
Passes:

merged to next

jonbob:
merged to master

PR #5130, opened 2022-08-12
update google cloud gnu compiler to gcc-11

sarich:
Updates the gnu compiler on google cloud platform from 8.5.0 to 11.1.0

includes new library stack for:
netcdf
pnetcdf
hdf5
blas/lapack

tested on e3sm_integration with no new errors

[BFB]


ndkeen (commented):
I really don't like the module names.  I was using 
```
mkdir .spack
cat > .spack/modules.yaml
modules:
  tcl:
    hash_length: 0
    naming_scheme: '{name}/{version}'
```

sarich:
No problem, I can change the name. But we still need to differentiate between gcc8 and gcc11 versions, do you have a preference for that format?

ndkeen:
Well shouldn't the `{name}/{version}` do that for us?

rljacob:
Waiting to settle on module names

ndkeen:
I see the commit. Module names look ok. Cool. I tested with one simple test and seems ok. Running a e3sm_developer.

ndkeen (changes requested):
Actually, after the PR aptly named "Throttle sbatch submission on Crusher #5132", this flag adjustment is moved to Depends.gnu.cmake.  So you should not longer even need this file.

ndkeen:
I tried one of our failing tests with this branch. `SMS_D_Ld1.ne30pg2_EC30to60E2r2.WCYCLSSP370.gcp_gnu.allactive-wcprodssp` and it fails in a different way (with FP exception). So could be good to get this in and we can debug.

PR #5131, opened 2022-08-12
Adjust PE layouts for ne30pg2 that will impact scream tests (esp on pm-gpu)

ndkeen:
Main purpose for this change is to allow ne30pg2 tests on pm-gpu to use 2 nodes out-of-box for testing.

This changes the file `cime_config/allactive/config_pesall.xml` and I suspect the ideal file to change
would be either:
```
components/scream/cime_config/config_pes.xml
components/eam/cime_config/config_pes.xml
```
but currently scream tests are using `cime_config/allactive/config_pesall.xml`.

I also added reasonable defaults for this grid on pm-cpu/cori-knl (again just affecting scream tests).
These PE layout defaults would be used for ne30np4 as well as ne30pg2.

[bfb]

### DIFF
--- a/cime_config/allactive/config_pesall.xml
+++ b/cime_config/allactive/config_pesall.xml
@@ -2016,6 +2016,77 @@
       </pes>
     </mach>
   </grid>
+  <grid name="a%ne30np4">
+    <mach name="pm-gpu">
+      <pes compset="any" pesize="any">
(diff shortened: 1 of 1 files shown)


ambrad (approved):
PEM_Ln90.ne30pg2_ne30pg2.F2010-SCREAMv1.pm-gpu_gnugpu passes with this PR. Thanks!

ndkeen:
merged to next

PR #5132, opened 2022-08-12
Throttle sbatch submission on Crusher

amametjanov:
To max 20 test jobs at a time to avoid Slurm AssocMaxSubmitJobLimit error.
Also,
- limit default job wall time to 1 hour (for better queue throughput)
- update HDF5 modules to fix version mismatch errors
- add BASELINE_ROOT for CIME job wall-time tracking
- throttle Ascent test job submits to 99 max submits
- pull up -fno-inline-arg-packing to generic Depends.gnu[gpu].cmake
- increase memleak tolerance for fates tests
- increase number of nodes for bgc tests

[BFB]


ndkeen:
Looks like you fixed https://github.com/E3SM-Project/E3SM/issues/5136 in the process.
Which is great, but it's not related to the title of the PR.

PR #5135, opened 2022-08-13
For gcp only, use 4 nodes to allow cdash test to complete

ndkeen:
On GCP, the new `ERS.hcru_hcru.I20TRGSWCNPRDCTCBC.gcp_gnu.elm-erosion` test was failing as it was running out of memory with 1 node. Using 4 nodes seems to work.

[bfb]


ndkeen:
merged to next

PR #5137, opened 2022-08-15
For cori-haswell, increase PE layout for grid l%360x720cru to 6 nodes

ndkeen:
Similar to https://github.com/E3SM-Project/E3SM/pull/5113, the new test was failing with 4 nodes, but works with 6 nodes
on cori-haswell.

Impacts following test: `ERS.hcru_hcru.I20TRGSWCNPRDCTCBC.cori-haswell_intel.elm-erosion`

[bfb]

### DIFF
--- a/components/elm/cime_config/config_pes.xml
+++ b/components/elm/cime_config/config_pes.xml
@@ -313,6 +313,21 @@
         </nthrds>
       </pes>
     </mach>
+    <mach name="cori-haswell">
+      <pes compset="any" pesize="any">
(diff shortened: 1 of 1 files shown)


ndkeen:
I did notice that this test (noted above) actually passed most recently.  But as I saw it was failing in the past and as I knew we needed more nodes for it on other machines, just assumed same here.  going from 4 to 6 nodes for this test should at least make it run faster. 

 merged to next

ndkeen:
Looks like this test passed on cdash.

PR #5140, opened 2022-08-17
HOMME: Some (cmake) build mods

bartgol:
This PR does mainly two things:

1) Eliminates the internal csm_share folder, which contained a copy (at some point in time) of some of the files in `share/utils`. Instead, for standalone homme builds, we will now grab the src files directly from share/utils to build a csm_share library, which we link against the executables. If Homme is built (via cmake) from an outside cmake project, we will not build the csm_share library, and it will be the responsibility of the outside project to ensure that a library providing the functionalities of csm_share is linked to the homme libraries. This has 0 impact on EAM, since it's just a bunch of cmake build mods.

2) Removes the compilation of certain files and pieces of code if the CPP macro `SCREAM` is defined. This allows the SCREAM repo to skip compiling code that it would never use anyways (since it's for standalone homme testing). This should have 0 impact on EAM, since it changes a bunch of `#ifndef CAM` to `#if !defined(CAM) && !defined(SCREAM)`. Still, I tested e3sm_developer too.

Testing: e3sm_developer and homme_integration both pass on chrysalis. Also, I used this branch in SCREAM, and verified it builds correctly with no homme's internal csm_share lib built, and no test_src/* files built.

[BFB]


components/homme/src/repro_sum_mod.F90
```diff
@@ -41,7 +41,7 @@ module repro_sum_mod
 #else
    use shr_kind_mod,  only: r8 => shr_kind_r8, i8 => shr_kind_i8 
 #endif
-   use shr_sys_mod,   only: shr_sys_abort  
+   use shr_abort_mod,   only: shr_sys_abort=>shr_abort_abort
```
bartgol, line 44: This is the single difference between the modules that were in the csm_share folder in homme, and the modules in `$repo/share/utils`.

components/homme/utils/csm_share/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+    add_custom_command (
+      OUTPUT ${SRC_FILE_OUT}
+      COMMAND ${GENF90} ${SRC_FILE} > ${SRC_FILE_OUT}
+      DEPENDS ${SRC_FILE} genf90)
+    list(APPEND CSM_SHARE_SOURCES ${SRC_FILE_OUT})
+  endforeach ()
+
+  # "normal" sources
+  list (APPEND CSM_SHARE_SOURCES
+    ${SHARE_DIR}/shr_abort_mod.F90
```
bartgol, line 37: Note: there were more modules in the original csm_share folder, but turned out many of them were never used (directly or indirectly) in standalone homme. So I only included the ones that are effectively used.

If other modules end up being used in the future, we simply have to add them here.

bartgol:
@oksanaguba @ambrad this is not urgent, but wanted to put it up now, while I can guarantee testing passes on chrysalis (and this PR is ahead of master as of now).

oksanaguba (approved):

ambrad:
@mt5555 do you see any issues with removing these utils/csm_share files, given that homme_integration passes?

mt5555:
It will be great to remove this code - it's been on my wishlist for a while.  Agree that if homme_integration passes, should be fine.  

On question: I see a CMake variable, HOMME_STANDALONE, and in the Fortran code, there is a lot of:

should we standardize on using only HOMME_STANDALONE at some point in the future?

mt5555 (approved):

bartgol:
> On question: I see a CMake variable, HOMME_STANDALONE, and in the Fortran code, there is a lot of: #if !defined(CAM) && !defined(SCREAM)
> 
> should we standardize on using only HOMME_STANDALONE at some point in the future?

I floated that idea in the past. But at that point, there were a few `#ifndef CAM`, a few `#ifndef SCREAM`, and a few `#if !defined(CAM) && !defined(SCREAM)`, and the consensus was to not introduce yet another macro.

Now, I still see some `#ifdef SCREAM`, but they're very few, and the majority of CPP conditionals check for both CAM and SCREA. So at this point, I think it makes sense to do `#ifdef HOMME_STANDALONE`, which is also much more self-explanatory.

ambrad:
Ok, great. Thanks for the reviews. @bartgol is this PR ready for me to start the merge process?

bartgol:
Yes. I think it is.

bartgol:
Ok, I have confirmed that this branch, when properly integrated in scream (without any further mod in the homme folder) leads to correct build/testing on mappy and weaver. I am confident we can start integration now.

ambrad:
No rush. right? I'll likely merge to next on Monday.

bartgol:
Yeah, that's fine.

oksanaguba:
This PR broke build of standalone homme execs.

oksanaguba:
@bartgol -- please, see the above comment. i am not sure about what you intended to do wrt csm_share in standalone homme. thanks!

bartgol:
Uhm, I did run the homme testsuite, so this should have worked...

PR #5142, opened 2022-08-17
correct nCells to nEdges

### LINKED ISSUE #5141: nEdges set to nCells in ocean diagnostics
This is an obvious typo that should be fixed:
```
components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
line 1110:
      nEdges = nCellsAll
```

mark-petersen:
This corrects an obvious error, where the local `nEdges` was mistakenly set to the `nCellsAll` value. Normally this would case a failure in partition tests, but the variables in this section are not used for standard E3SM settings. That also explains why simulations are BFB when compared before/after this fix.

[BFB]
fixes #5141

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
@@ -1112,7 +1112,7 @@ contains
 
       nCells = nCellsAll
       nVertices = nVerticesAll
-      nEdges = nCellsAll
+      nEdges = nEdgesAll
       apvm_scale_factor = config_apvm_scale_factor;
 
       ! Diagnostics required for the Anticipated Potential Vorticity Method (apvm).


mark-petersen:
Compiled with gnu on badger. Nightly suite on PR is BFB against master.

philipwjones (approved):
Approve by visual inspection

philipwjones:
Actually, while looking at this routine, I also happened to notice that the k-loop limits for the first edge loop in APVM section are slightly different from the other two k-loops in this sections. Is that correct?

First (L1181) is: do k = minLevelEdgeBot(iEdge), maxLevelEdgeTop(iEdge)
Others are: do k = minLevelEdgeTop(iEdge), maxLevelEdgeBot(iEdge)

mark-petersen:
@cbegeman could you check @philipwjones' question? There are instances where we go from bot to top rather than top to bot. We are just making sure.

cbegeman:
Looking back at the git blame, the loop was originally `do k=1,maxLevelEdgeTop(iEdge)`
https://github.com/E3SM-Project/E3SM/blame/8e93078a75f1551b542d1aab047381666c8b45ed/shared/mpas_ocn_diagnostics.F#L756, so I made the change consistent with that to `do k = minLevelEdgeBot(iEdge), maxLevelEdgeTop(iEdge) `. 

It makes sense for `vorticityGradientNormalComponent` to be zero wherever the edge is not adjacent to 2 cells (when `maxLevelEdgeTop` < `maxLevelEdgeBot` and similarly for `minLevel*`). As you can see, we need `normalizedRelativeVorticityCell` to be valid for both `cell1` and `cell2`. So the loop should not be changed.

philipwjones:
Thanks @cbegeman . It looked intentional but just making sure since we were correcting another bug in that routine.

jonbob:
@mark-petersen - just a reminder that the branch naming convention is all lowercase

jonbob:
Test merge passes:

merged to next

jonbob:
merged to master

PR #5149, opened 2022-08-22
Updates a MOSART test

bishtgautam:
The DATM_CLMNCEP_YR_END of the two-way river-ocean test is updated to avoid downloading
large number of netcdf files that are not needed for the test.

[BFB]


bishtgautam:
#5150 replaces this PR.

PR #5027, opened 2022-06-16
Make MPAS-Ocean units CF compliant

darincomeau:
This makes changes to the units in MPAS-Ocean's Registry files to be CF compliant. [CF Standard Units](https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html)

The main changes for CF compliance are:

- variables not have a units attribute if they are unitless (i.e. no units='unitless', units='NA', etc.)
- variables should have units=1 if they are dimensionless (e.g. real-valued coefficients)
- we cannot use latex symbols like { and } in the units
- we cannot use units='PSU' for salinity, we need to use units=1e-3, though we can state that the units are grams per kilogram in the description (used as the long_name attribute in output files).

Further discussion found in https://github.com/E3SM-Ocean-Discussion/E3SM/pull/24

Note this does not change any units on BGC variables, as found in the following Registry files under `tracer_groups`:
- Registry_CFC
- Registry_DMS
- Registry_MacroMolecules
- Registry_ecosys
- Registry_idealAge

[BFB]

PR #5151, opened 2022-08-23
add carbon to conservation check analysis member

maltrud:
This PR adds carbon to the conservation check analysis member.  output is written to both the log file and conservation check netcdf file.  carbon variables are written to the netcdf file even if BGC is not turned on (unless the user modifies streams.ocean), but not to the log file.


jonbob:
@maltrud - I tested this in a ne30pg2_r05_EC30to60E2r2 BGCEXP_CNTL_CNPECACNT_1850 configuration and it died after the first month with:
````
CRITICAL ERROR: WARNING: relative carbon error exceeds bounds
````
I think you need some kind of accumulator for the number of times you've added sums into the accumulation arrays? In my case, after it went to output the carbon conservation information after a month, the relative error and accumulated relative error are the same:
````
 RELATIVE CARBON ERROR = -1.30190623E-08

 Accumulated absolute carbon error (kg)  -4.74559000E+08

 Accumulated relative carbon error =     -1.30190623E-08
````
but the bounds checks you're applying to the accumulated one has been multiplied by the number of time steps while the relative one is compared to a hardcoded 1.0e-08. That one might need to have some factor that considers the number of steps/days that have been accumulated, especially since the budget output could be per step/day/month/year?

Anyway, I'm retesting with 2.0e-08 just to see if that allows it to hold together....

File: components/mpas-ocean/src/analysis_members/Registry_conservation_check.xml [hunk 1]
```diff
@@ -198,6 +198,53 @@
...
+                <var name="carbonChange" type="real" dimensions="Time" units="kg"
+                     description="Change in ocean carbon over conservation interval"
+                />
+                <var name="netCarbonFlux" type="real" dimensions="Time" units="kg"
+                     description="Net surface flux of ocean carbon over conservation interval"
+                />
+                <var name="absoluteCarbonError" type="real" dimensions="Time" units="kg"
+                     description="Absolute carbon conservation error over conservation interval"
+                />
+                <var name="relativeCarbonError" type="real" dimensions="Time" units="1"
```
mark-petersen, line 217: @darincomeau can you check the units format here? I think your recent PR #5027 changes `units="1"` to `units=""`, correct? Should change here and line 223.

mark-petersen:
As a first sanity test, I compiled MPAS-Ocean standalone with gnu and intel, optimized and debug, and ran the nightly suite with `config_AM_conservationCheck_enable = .true.`. Everything passed. Admittedly, these tests are not exercising the BGC or carbon parts of this code.

@maltrud can you describe here your testing and results with this analysis member on and ocean BGC variables on?

components/mpas-ocean/src/analysis_members/Registry_conservation_check.xml
xylar: No, this is correct.  `1` is the correct unit for something that is dimensionless but physical like this.

xylar: We never want `units=""`.  We would get rid of `units` as an attribute for something that simply doesn't have units.

darincomeau: Yes, agree with @xylar - `1` is the correct units here. Thanks for chiming in!

mark-petersen: Great, thanks!

maltrud:
@jonbob yes I should have used 1.e10 instead of 1.e6 for consistency with the cpl output.  I'll push that change asap.

rljacob:
@katsmith133 please review.

katsmith133 (approved):
Approving based upon @jonbob's test and visual inspection

jonbob:
Thanks @maltrud -- after the latest update, the output now looks like:
````
MPAS-Ocean name             kg/s (raw sum)    coupler name    short name         kg/m^2/s*1.e10 (flux)
 MarBL SrcSink+sedFlux      -2.09418656E-11                                      -0.00000000
 sedimentFlux                3.37745618E+04                                       0.66211500
 surfaceFlux                 6.33063584E+04                                       1.24105503
 CO2 gas Flux                2.90800701E+04                                       0.57008440
 Ice-Ocean Organic Flux     -1.41769673E+04                                      -0.27792464
 Ice-Ocean Inorganic Flux    4.22031762E+04                                       0.82734918
 SUM FLUXES (surf + sed)     9.70809202E+04                                       1.90317003
````
and it much easier to compare with the cpl budget

jonbob (approved):
Approved, based on testing

jonbob (approved):

jonbob:

jonbob:
merged to master

PR #5152, opened 2022-08-24
[standalone MPAS] Finalize PIO when finalizing MPAS i/o

xylar:
This allows SCORPIO to print out timing information needed for some types of debugging.

`pio_finalize()` is only called in standalone. If MPAS components are run as part of an E3SM run, the PIO context is passed to MPAS and this signals that the MPAS framework should _not_ finalize PIO.

### DIFF
--- a/components/mpas-framework/src/framework/mpas_framework.F
+++ b/components/mpas-framework/src/framework/mpas_framework.F
@@ -61,7 +61,7 @@ module mpas_framework
 !> \author Michael Duda, Doug Jacobsen
 !> \date   03/26/13
 !> \details
-!>  This routine finalizes the initialization of the MPAS framework. It calls initializes 
+!>  This routine finalizes the initialization of the MPAS framework. It calls initializes
 !>  the time keeper, and the IO infrastructure.
 !
 !-----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


xylar:
See discussion PR https://github.com/E3SM-Ocean-Discussion/E3SM/pull/30 for additional context.

xylar:
## Testing

I ran the compass ocean `pr` test suite and verified that the expected `io_perf_summary_*.txt` and `io_perf_summary_*.json` files are produced for each forward run of MPAS-Ocean.  These files were previously not available.

I ran the ` ERS.ne11_oQU240.WCYCL1850NS.anvil_intel` E3SM test and verified that timing appears as usual in the `spio_stats/` subdirectory (presumably a good indication that the MPAS framework didn't erroneously finalize PIO too soon).

xylar:
@jonbob, would you like me to do further E3SM testing on this to make sure I'm doing no harm?  Happy to run  whatever tests you suggest.

jonbob:
@xylar - I'll do more testing before I merge, so I think this looks fine. Thanks

mark-petersen (approved):
This functions for the ocean core. Tested with gnu and intel, both debug and optimized, nightly suite on badger. Also compares bfb against branch point with gnu optimized.

matthewhoffman (approved):
Approving based on discussion yesterday, inspection, and Mark's testing.  This appears to be the proper implementation that mirrors the init operations.

xylar:
Thanks for these reviews as well, @matthewhoffman and @mark-petersen!  Much appreciated.

@akturner, this is lower priority than #5162 for me but would appreciate you having a look at this soon, too, if you can.

akturner (approved):

xylar:
Thanks @akturner!

jonbob:
Test merge passes sanity testing on chrysalis:

will merge to next as soon as the repo is open

jonbob:
merged to next

jonbob:
merged to master

Issue #5076, opened 2022-07-14
MEMLEAK flagged with ERP_Ld9.ne4_ne4.FIDEAL.pm-cpu_gnu


ndkeen:
Using master of July 13th, I see a MEMLEAK with `ERP_Ld9.ne4_ne4.FIDEAL.pm-cpu_gnu`.

```
 memory_write: model date =   00010101       0 memory =     825.01 MB (highwater)         83.84 MB (usage)  (pe=    0 comps= cpl ATM LND ICE OCN GLC ROF WAV IAC ESP)
 memory_write: model date =   00010102       0 memory =     855.14 MB (highwater)         89.11 MB (usage)  (pe=    0 comps= cpl ATM LND ICE OCN GLC ROF WAV IAC ESP)
 memory_write: model date =   00010103       0 memory =     855.14 MB (highwater)         89.90 MB (usage)  (pe=    0 comps= cpl ATM LND ICE OCN GLC ROF WAV IAC ESP)
 memory_write: model date =   00010104       0 memory =     855.14 MB (highwater)         90.38 MB (usage)  (pe=    0 comps= cpl ATM LND ICE OCN GLC ROF WAV IAC ESP)
 memory_write: model date =   00010105       0 memory =     855.14 MB (highwater)         90.88 MB (usage)  (pe=    0 comps= cpl ATM LND ICE OCN GLC ROF WAV IAC ESP)
 memory_write: model date =   00010106       0 memory =    1255.43 MB (highwater)        115.98 MB (usage)  (pe=    0 comps= cpl ATM LND ICE OCN GLC ROF WAV IAC ESP)
 memory_write: model date =   00010107       0 memory =    1255.43 MB (highwater)        116.05 MB (usage)  (pe=    0 comps= cpl ATM LND ICE OCN GLC ROF WAV IAC ESP)
 memory_write: model date =   00010108       0 memory =    1255.43 MB (highwater)        116.05 MB (usage)  (pe=    0 comps= cpl ATM LND ICE OCN GLC ROF WAV IAC ESP)
 memory_write: model date =   00010109       0 memory =    1255.43 MB (highwater)        116.31 MB (usage)  (pe=    0 comps= cpl ATM LND ICE OCN GLC ROF WAV IAC ESP)
 memory_write: model date =   00010110       0 memory =    1255.43 MB (highwater)        116.70 MB (usage)  (pe=    0 comps= cpl ATM LND ICE OCN GLC ROF WAV IAC ESP)
[...]

PR #5159, opened 2022-08-26
Add PIO_ROOT=1 mod to FIDEAL test

amametjanov:
To avoid false-positive memleak fail on ROOTPE 0 in ERP tests.

Fixes E3SM-Project/E3SM#5076

[NML] - due to the test name change


rljacob:
telecon notes:  @ndkeen tested it and it works.  But is this the right way?

ndkeen:
This does allow the test to pass on pm-cpu, but to me it doesn't seem like the right way to fix the overall issue. I'm fine with approving to allow this test to pass for now -- as I think we agree there is no real memory growth issue here.

ndkeen (approved):

PR #5162, opened 2022-08-29
[MPAS-Framework] Rename `gregorian_noleap` to just `noleap` calendar

xylar:
The `noleap` is the CF-compliant name for what MPAS currently calls the `gregorian_noleap` calendar. See https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch04s04.html.

This merge changes `gregorian_noleap` to `noleap` in the MPAS framework and in all 3 MPAS components.

The context of this change is that the `config_calendar_type` namelist option is being used for the `calendar` attribute of the `Time` coordinate in https://github.com/E3SM-Ocean-Discussion/E3SM/pull/19.  However, the result is not making post-processing tools (notably `xarray`) happy because the `gregorian_noleap` calendar is not CF-compliant.

```
ValueError: calendar must be one of ['standard', 'gregorian', 'proleptic_gregorian', 'noleap', 'julian', 'all_leap', '365_day', '366_day', '360_day'], got 'gregorian_noleap'
```

[NML]
[BFB]

### DIFF
--- a/components/mpas-albany-landice/bld/namelist_files/namelist_defaults_mali.xml
+++ b/components/mpas-albany-landice/bld/namelist_files/namelist_defaults_mali.xml
@@ -116,7 +116,7 @@
 <config_start_time>'0000-01-01_00:00:00'</config_start_time>
 <config_stop_time>'0000-01-01_00:00:00'</config_stop_time>
 <config_run_duration>'none'</config_run_duration>
-<config_calendar_type CALENDAR="NO_LEAP">'gregorian_noleap'</config_calendar_type>
+<config_calendar_type CALENDAR="NO_LEAP">'noleap'</config_calendar_type>
 <config_calendar_type>'gregorian'</config_calendar_type>
 
 <!-- io -->
--- a/components/mpas-albany-landice/bld/namelist_files/namelist_definition_mali.xml
+++ b/components/mpas-albany-landice/bld/namelist_files/namelist_definition_mali.xml
@@ -17,7 +17,7 @@
           An abbreviation of the fortran declaration for the variable.
       Valid declarations are:
 
-          char*n  
+          char*n
           integer
           logical
           real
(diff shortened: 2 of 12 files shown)


xylar:
@mark-petersen and @jonbob, do you foresee any problem with renaming `gregorian_noleap` to just `noleap` for `config_calendar_type` in MPAS-Ocean?

xylar:
@akturner and @matthewhoffman, do you have any problem with the generalization I have made to support more calendar names in the MPAS-Framework?

Eventually, I will be making the same change from `gregorian_noleap` to `noleap` in MPAS-Seaice and MALI if you're onboard.  This is so we can have a CF-compliant time coordinate in all components.

If you aren't happy with the change, an alternative would be to just have a list of `if`/`else if` statements to convert from `config_calendar_type` to the corresponding CF-compliant calendar but I don't feel that the existing `gregorian_noleap` name has a lot of added value over the CF names for this calendar.  I'd prefer to just have the one, consistent name for the calendar.

jonbob:
@xylar -- I would have just suggested the type of test you ran, so I think it's fine

xylar:
@akturner, @matthewhoffman, @mark-petersen, and @jonbob following our discussion, I renamed `gregorian_noleap` to `noleap` in all components.  (The branch is no longer very well named but I don't want to open a new PR unless we feel strongly about it.)

Please have a look when you can.  I haven't had a chance to retest yet myself but I'll do that tomorrow.

matthewhoffman (approved):
Approving based on inspection, Mark's testing, and looking over MALI use cases to confirm we are not explicitly using this anywhere (so the change will be seamless for us other than existing namelist files in production usage).  @xylar , thanks for making the adjustment based on the discussion yesterday.  I like that we are dropping the incompatible option entirely.

xylar:
Thanks very much @mark-petersen and @matthewhoffman!

@akturner, is this one you can test out in  the next couple of days?

akturner (approved):
Tested in MPAS-Seaice standalone and passed all tests once namelists were changed from gergorian_noleap to noleap

jonbob:

jonbob:
merged to master and expected NML DIFFs blessed. We will need to bless more once mappy reports

xylar:
Thanks @jonbob and everyone who reviewed!

PR #5163, opened 2022-08-30
initialize CRM output quantities for check_energy

whannah1:
This bug only applies to the first time step. These values have always been uninitialized, but somehow never caused a problem. While running a ne120 MMF case in debug mode these uninitialized values became problematic, so this PR should avoid these problems in the future. 
[BFB]


whannah1:
Just verified that this bug fix doesn't interfere with exact restarts using a ERS_Ln9.ne4pg2_ne4pg2.F2010-MMF1 test on Cori.

lee1046 (approved):
It looks good to me.

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #5165, opened 2022-08-30
On GCP, use 4 nodes for r05 grid to allow test to complete

ndkeen:
For GCP only, the `ERS.r05_r05.IELM.gcp_gnu.elm-V2_ELM_MOSART_features` test was running out of time on 1 node and using 4 nodes was at least 3x faster.

[bfb]


ndkeen:
merged to next -- hopefully will see the test pass with next cdash result on GCP

PR #5166, opened 2022-08-31
Update CIME submodule

jgfouca:
... to a7fa4c0333ca9799a25f2fbec0a1578a3ac52324

Big changes:
1) Remove provenance and model specific configuration from
CIME. CIME will now look under a models cime_config directory
for a directory named customize. All python files under this directory
will be loaded. Provence now works as a hook system, specific methods
need to be defined to enable provenance. Configuration flags that
alter how CIME operates may be defined here as well.

Changes:
1) Switch from highwater to mem-usage in memleak checks
2) Drop support for config_compilers, remove code and files
that supported this.
3) Allow CONTINUE_RUN without existing rpointer if prereq is used

Fixes:
1) create_test: fix compiler inferring
2) Open TSC user nls mode w+ to truncate before write

[BFB]


jgfouca:
Pushed to next so integration testing can start hitting it.

sarats:
Can you elaborate what this means?
> Switch from highwater to mem-usage in memleak checks

sarats:
I'm not sure what the following snippet's impact is 
```
-            r".*model date =\s+(\w+).*memory =\s+(\d+\.?\d+).*highwater"
+           r".*model date =\s+(\w+).*memory =.*highwater\)\s+(\d+\.?\d+).*usage"
```

jgfouca:
@sarats ,

> I'm not sure what the following snippet's impact is

It makes it so the regex will grab the mem-usage field instead of the highwater field from the memory info in the log.

This change is from https://github.com/ESMCI/cime/pull/4296 by @amametjanov . I don't think this should impact PACE since this code is only for memleak detection in CIME system tests.

rljacob:
@amametjanov can you also elaborate on "For basic compsets, the increases in RSS are too high (but a false positive)".  Why only basic compsets (A, X) ?

sarats:
@amametjanov Does this impact data written to the memory logs? If it does, it makes comparison with previous experiments difficult to find macro-memory anomalies. Can you confirm?

amametjanov:
> Does this impact data written to the memory logs?

No, no impact to memory log data , this change was only for mem-leak checks within test-cases. 

> Why only basic compsets (A, X) ?

The unexpected impact is in the relative increase: smaller baseline number (in smaller compset) makes a 1MB increase relatively big. E.g.,
- ERS.ne30_g16_rx1.A.chrysalis_intel:
  - RSS: 303.64 -> 318.89 = 15.25 MB or 5.02% increase (over Chrysalis' default 5% threshold)
  - VSZ: 1473.71 -> 1473.71 = 0 MB or 0% impact
- ERS.ne11_oQU240.WCYCL1850NS.chrysalis_intel:
  - RSS: 1240.83 -> 1248.67 = 7.84 MB or 0.6% increase
  - VSZ: 2600.23 -> 2670.27 = 70.04 MB or 2.7% increase

To reduce mem-leak sensitivity, I should probably revert to conservative, total memory (VSZ, highwater) checking.

jgfouca:
@amametjanov , @sarats , can I merge this PR?

amametjanov:
Yes, I didn't see any other impact on E3SM+next.

PR #5170, opened 2022-09-02
Alters MOC calculation in ocean component

vanroekel:
Currently the southern boundary is integrated from top to bottom, this is changed to be from bottom to top to take advantage of the natural boundary condition that the streamfunction at the bottom is zero.  In the calculation we implicitly assume that there is a mocStreamValLatAndDepthLocal(nVertLevels+1) that is always zero.

The calculation is also consistent with that used by the POP model for the same diagnostic -- https://github.com/ESCOMP/POP2-CESM/blob/master/source/diags_on_lat_aux_grid.F90#L1155-L1158

### DIFF
--- a/components/mpas-ocean/src/analysis_members/mpas_ocn_moc_streamfunction.F
+++ b/components/mpas-ocean/src/analysis_members/mpas_ocn_moc_streamfunction.F
@@ -508,10 +508,15 @@ contains
                      * 0.5_RKIND*(layerThickness(k,c1) + layerThickness(k,c2))
                end do
             end do
-            do k = 2, nVertLevels
+            !Integrate transport from bottom to top, which switches the sign of sumTransport
+            !there is also an implicit moc(nVertLevels+1,iTransect) = 0
+            k = nVertLevels
+            mocStreamValLatAndDepthRegionLocal(1, k, iTransect) = &
+               - sumTransport(k, iTransect)
+            do k = nVertLevels-1, 1, -1
                mocStreamValLatAndDepthRegionLocal(1, k, iTransect) = &
-                  mocStreamValLatAndDepthRegionLocal(1, k - 1, iTransect) &
-                  + sumTransport(k - 1, iTransect)
+                  mocStreamValLatAndDepthRegionLocal(1, k + 1, iTransect) &
+                  - sumTransport(k - 1, iTransect)
             end do
          end do
 
@@ -606,10 +611,15 @@ contains
                        * 0.5_RKIND*(layerThickness(k,c1) + layerThickness(k,c2))
                  end do
               end do
-              do k = 2, nVertLevels
+              !Integrate transport from bottom to top, which switches the sign of sumTransport
+              !there is also an implicit moc(nVertLevels+1,iTransect) = 0
+              k = nVertLevels
+              mocStreamValLatAndDepthRegionLocal(1, k, iTransect) = &
+                 - sumTransport(k, iTransect)
+              do k = nVertLevels - 1, 1, -1
                  mocStreamValLatAndDepthRegionLocal(1, k, iTransect) = &
-                    mocStreamValLatAndDepthRegionLocal(1, k - 1, iTransect) &
-                    + sumTransport(k - 1, iTransect)
+                    mocStreamValLatAndDepthRegionLocal(1, k + 1, iTransect) &
+                    - sumTransport(k + 1, iTransect)
               end do
            end do
 
@@ -712,10 +722,15 @@ contains
                        * 0.5_RKIND*(layerThickness(k,c1) + layerThickness(k,c2))
                  end do
               end do
-              do k = 2, nVertLevels
+              !Integrate transport from bottom to top, which switches the sign of sumTransport
+              !there is also an implicit moc(nVertLevels+1,iTransect) = 0
+              k = nVertLevels
+              mocStreamValLatAndDepthRegionLocal(1, k, iTransect) = &
+                 - sumTransport(k, iTransect)
+              do k = nVertLevels - 1, 1, -1
                  mocStreamValLatAndDepthRegionLocal(1, k, iTransect) = &
-                    mocStreamValLatAndDepthRegionLocal(1, k - 1, iTransect) &
-                    + sumTransport(k - 1, iTransect)
+                    mocStreamValLatAndDepthRegionLocal(1, k + 1, iTransect) &
+                    - sumTransport(k - 1, iTransect)
               end do
            end do
 


components/mpas-ocean/src/analysis_members/mpas_ocn_moc_streamfunction.F
```diff
@@ -508,10 +508,15 @@ subroutine ocn_compute_moc_streamfunction(domain, timeLevel, err)!{{{
...
+            !there is also an implicit moc(nVertLevels+1,iTransect) = 0
+            k = nVertLevels
+            mocStreamValLatAndDepthRegionLocal(1, k, iTransect) = &
+               - sumTransport(k, iTransect)
+            do k = nVertLevels-1, 1, -1
                mocStreamValLatAndDepthRegionLocal(1, k, iTransect) = &
-                  mocStreamValLatAndDepthRegionLocal(1, k - 1, iTransect) &
-                  + sumTransport(k - 1, iTransect)
+                  mocStreamValLatAndDepthRegionLocal(1, k + 1, iTransect) &
+                  - sumTransport(k - 1, iTransect)
```
mark-petersen, line 519: Should be k+1.

vanroekel: Fixed

xylar:
@vanroekel and @mark-petersen, I'm not sure how soon I'll have time to review this or make the corresponding changes to MPAS-Analysis.  I fly to the US tomorrow and it may take me a few days next week to settle in.

components/mpas-ocean/src/analysis_members/mpas_ocn_moc_streamfunction.F
```diff
@@ -508,10 +508,15 @@ subroutine ocn_compute_moc_streamfunction(domain, timeLevel, err)!{{{
...
+            mocStreamValLatAndDepthRegionLocal(1, k, iTransect) = &
+               - sumTransport(k, iTransect)
+            do k = nVertLevels-1, 1, -1
                mocStreamValLatAndDepthRegionLocal(1, k, iTransect) = &
-                  mocStreamValLatAndDepthRegionLocal(1, k - 1, iTransect) &
-                  + sumTransport(k - 1, iTransect)
+                  mocStreamValLatAndDepthRegionLocal(1, k + 1, iTransect) &
+                  - sumTransport(k + 1, iTransect)
             end do
          end do
```
xylar, line 521: @vanroekel, I'm looking at this carefully to implement it in MPAS-Analysis.

It seems to me that 
```
mocStreamValLatAndDepthRegionLocal(1, nVertLevels, iTransect)
```
should be zero, not 
```
-sumTransport(nVertLevels, iTransect)
```

As written, 
```
mocStreamValLatAndDepthRegionLocal(1, nVertLevels-1, iTransect) = -2*sumTransport(nVertLevels, iTransect)
```
which doesn't seem right to me.

vanroekel: @xylar I wrote it this way assuming there is an implicit `mocStreamValLatAndDepthRegionLocal(1,nVertLevels+1,tran) = 0`.  But looking again I think this means the second line is not right. we should have 
```
mocStreamValLatAndDepthRegionLocal(1, k + 1, iTransect) &
                  - sumTransport(k, iTransect)
 ```
instead of k+1.  Does this seem reasonable to you?  if so I can fix this in the PR.

Comment (xylar) [reply components/mpas-ocean/src/analysis_members/mpas_ocn_moc_streamfunction.F:521]: Yes, looking at it again, I think you are correct.

Comment (xylar) [reply components/mpas-ocean/src/analysis_members/mpas_ocn_moc_streamfunction.F:521]: In MPAS-Analysis, we are explicit about having `nVertLevels+1` entries in the MOC field, with a zeroes currently at the top and soon to be at the bottom.

I think the change you suggest would make this code consistent with POP, which you linked to above in the PR.

Comment (vanroekel) [reply components/mpas-ocean/src/analysis_members/mpas_ocn_moc_streamfunction.F:521]: Would you suggest I make the moc variable here nVertLevelsP1 for clarity?  I would be happy to make the change.

Comment (xylar) [reply components/mpas-ocean/src/analysis_members/mpas_ocn_moc_streamfunction.F:521]: I don't think that's necessary.  All the output variables are using `nVertLevels`:
https://github.com/E3SM-Project/E3SM/blob/e3f2c1524fb1a3a5cab0c283e2572064dea292e6/components/mpas-ocean/src/analysis_members/Registry_moc_streamfunction.xml#L54-L79
So it makes sense to stick with that same approach.  Conceptually, I think the extra row is a good idea because it make clear that the MOC values live between layers rather than at the centers of layers but I don't think it makes sense to change the size of the output variables, and thus I don't think we want to change the size of the temporary variables either.

File: components/mpas-ocean/src/analysis_members/mpas_ocn_moc_streamfunction.F (hunk 1 above)
Comment (xylar) [reply components/mpas-ocean/src/analysis_members/mpas_ocn_moc_streamfunction.F]: I knew I had seen this somewhere.  As we discussed below, it should be `k`.

**Comment by vanroekel** [comment]
@xylar I think I have the indexing fixed now, could you take a look when you have a minute?

I also realize the code I used in the long run has this same error.  I made a quick reproducer of the calculation to see what fixing the error might do.  Blue is fixed, orange is what is in the model for that run

image

So perhaps the AMOC gets slightly better.  I'll put that long run back in for another 10 years with the indexing fix to give you a comparison for the MPAS-Analysis changes.

**Comment by vanroekel** [comment]
I realized the indexing error I had here was in my long 800 year run.  To test the impact I analyzed the last year of the simulation (annual average over year 830) and compared with a new run with the fix (average of year 831) the old MOC calculation is blue and the new is orange.  They are fairly close.
image

**Review by mark-petersen — APPROVED** [review]
Approved by visual inspection and conversation and testing above. Also compiles and passes nightly suite with intel and gnu, and `SMS_D_Ln9.T62_oQU120_ais20.MPAS_LISIO_TEST.cori-haswell_gnu`.

**Comment by xylar** [comment]
I believe I have found another bug in the MOC AM that we should fix here.  I believe the index `iBin` is off by 2 from what it should be.  Shifting the latitude by 2 bins means the pattern I see with the postprocessing code matches the pattern from the MOC AM.  Since the bins are significantly different in size, I believe this indicates the problem is with the MOC AM only and not with the postprocessing code.

Furthermore, I think it makes sense that the following line:
https://github.com/E3SM-Project/E3SM/blob/311fe04c85ba3e635b60227244b162eb3c875d3e/components/mpas-ocean/src/analysis_members/mpas_ocn_moc_streamfunction.F#L526
and the other 2 similar computations should be:
```fortran
            iBin = MAX(int((latCell(iCell) - binBoundaryMocStreamfunction(1)) / binWidth) + 2, 2)
```
The first bin index is for the southern transect.  We have already filled that MOC entry in.  So the MOC should be added to the northern side of each bin.  Thus, any latitude within the bin just north of ` binBoundaryMocStreamfunction(1)` should go into bin 2.  Since `int((latCell(iCell) - binBoundaryMocStreamfunction(1)) / binWidth)` will be zero for such a cell (less than a bin width north of the southernmost bin boundary), we need to add 2 to get the desired index.  The `MAX()` function is just meant to handle cells that are south of the southernmost bin (and hopefully masked out if there are any such cells).  So I think this fix also makes sense there.

This fix is important, but especially for pulling out the AMOC at RAPID.  If we are off by 2 bins when we do that, this could be a non-negligible distance.  We are looking at nearly 2 degrees latitude of error, since the bins are about 1 degree each.

xylar:
One more thing, we need to add `normalMLEvelocity` and `vertMLEvelocityTop` to time series stats monthly.  Otherwise, we can't include these in the offline MOC.  And we may want them anyway for diagnostic purposes.

xylar:
Oh, sorry. It looks like `timeMonthly_avg_normalMLEvelocity` is there but `timeMonthly_avg_vertMLEVelocityTop` is not.

xylar:
> Oh, sorry. It looks like timeMonthly_avg_normalMLEvelocity is there but timeMonthly_avg_vertMLEVelocityTop is not.

@vanroekel, did you add this manually?  I don't see anywhere in the code that this has been added but I do see it in your `streams.ocean` in the test run you pointed me to.

vanroekel:
@xylar yes I did add that manually.  I'll add both variables as default to #5172 

I just resubmitted for 10 years with `vertMLEVelocityTop` included

xylar:
Okay, one an additional 10 years are done, I'll test manually shifting latitude by 2 bins as part of my comparison.  Thank @vanroekel!

vanroekel:
I forgot to mention that I looked at your comment and agree it is a bug (just pushed the change) so I corrected the latitude shift for the 10 years I'm running now.

I'm heading out of town tomorrow for a week and likely won't be able to check on the run, but it will appear here 
```
/lcrc/group/e3sm/ac.vanroekel/E3SM_simulations/20220715.submeso.piControl.ne30pg2_EC30to60E2r2.chrysalis/run
```
as years 841-850

xylar:
Great, thanks @vanroekel!

xylar:
Safe travels!

vanroekel:
Thanks!

One last note @xylar -- I forgot that the vertical MLE velocity variable name is a bit different in my older code.  It is `vertMLEBolusVelocityTop`.  I removed the Bolus part for the later PR.  Sorry for the bit of confusion.  Hopefully this doesn't complicate your testing.

xylar:
I got MPAS-Analysis running, comparing the AM to the Post-processed version.
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis/post_submeso.piControl.ne30pg2_EC30to60E2r2/moc_postproc/yrs841-850/ocean/index.html
I'm not comfortable with differences this large without understanding the cause.  I'll keep looking into it as I have time.

xylar:
Okay, great!  I switched the postprocessed MOC to use the time-average layer thickness, averaged from cells to edges, instead of the reference z-level layer thickness and the southern transport now matches nearly exactly.  The time series at RAPID doesn't match perfectly but that is presumably related to different latitude bins or other small differences that are expected:
image

xylar (approved):
This looks good and matches the post-processed results to the degree expected.

jonbob:
@vanroekel - you labeled this PR as non-BFB, but it's just a diagnostic field? So far my testing is BFB. Does that make sense to you?

vanroekel:
@jonbob I think that was because the diff came up in the compass test I did as it diffed the analysis output too.  I wasn't sure if that would come up here as well.  If diffs in analysis are fine and retains BFB we can relabel this PR

jonbob:
merged to next

jonbob:
merged to master

PR #5175, opened 2022-09-06
Fix determination of flds_c_oi logical variable

jonbob:
Fixes the determination of the flds_c_oi logical variable in the coupler carbon budget code, used to load carbon exchange fields between the ice and ocn. This only impacts the carbon coupler budget output.

[BFB]


jonbob:
Checked with an ne30pg2_r05_EC30to60E2r2 BGCEXP_CNTL_CNPECACNT_1850 configuration, when ice-ocn fields are expected as well as with an ne30pg2_r05_EC30to60E2r2 WCYCL1850 case where they are not, in both optimized and debug mode

maltrud (approved):

jonbob:

jonbob:
merged to master

PR #5178, opened 2022-09-07
Add script support for ARRM10to60E2r1 ocn/ice grid

darincomeau:
Brings in support for the `ARRM10to60E2r1` ocn/ice mesh. https://github.com/MPAS-Dev/compass/pull/414 has details on the mesh generation, which is an updated version of the `oARRM60to10` mesh. This PR includes all necessary mapping, runoff, and domain files for the following configurations:

`ne30pg2_ARRM10to60E2r1`
`arcticx4v1pg2_ARRM10to60E2r1`
`TL319_ARRM10to60E2r1` (JRA-forced)
`T62_ARRM10to60E2r1` (CORE-forced)

*Testing*

The following 5 day smoke tests have been run on Chrysalis:
`WCYCL1950.ne30pg2_ARRM10to60E2r1`
`WCYCL1850.arcticx4v1pg2_ARRM10to60E2r1`
`GMPAS-JRA1p4.TL319_ARRM10to60E2r1`
`GMPAS-IAF.T62_ARRM10to60E2r1`

A `WCYCL1950.ne30pg2_ARRM10to60E2r1` B-case is currently running on Cori as the final check for this mesh, and is in year 26. [Here is the MPAS-Analysis for years 11-20](https://portal.nersc.gov/project/m1199/dcomeau/E3SMv2-Arctic/20220803.WCYCL1950.ne30pg2_ARRM10to60E2r1.baseline.cori-knl/mpas_analysis/ts_0011-0020_climo_0011-0020/ocean/index.html)

### DIFF
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -148,7 +148,7 @@
 <config_Redi_N2_based_taper_enable>.true.</config_Redi_N2_based_taper_enable>
 <config_Redi_N2_based_taper_min>0.1</config_Redi_N2_based_taper_min>
 <config_Redi_N2_based_taper_limit_term1>.true.</config_Redi_N2_based_taper_limit_term1>
-<config_Redi_min_layers_diag_terms>6</config_Redi_min_layers_diag_terms>
+<config_Redi_min_layers_diag_terms>15</config_Redi_min_layers_diag_terms>
 <config_Redi_horizontal_taper>'ramp'</config_Redi_horizontal_taper>
 <config_Redi_horizontal_ramp_min>20e3</config_Redi_horizontal_ramp_min>
 <config_Redi_horizontal_ramp_min ocn_grid="WCAtl12to45E2r4">30e3</config_Redi_horizontal_ramp_min>
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -148,7 +148,8 @@
 <config_Redi_N2_based_taper_enable>.true.</config_Redi_N2_based_taper_enable>
 <config_Redi_N2_based_taper_min>0.1</config_Redi_N2_based_taper_min>
 <config_Redi_N2_based_taper_limit_term1>.true.</config_Redi_N2_based_taper_limit_term1>
-<config_Redi_min_layers_diag_terms>15</config_Redi_min_layers_diag_terms>
+<config_Redi_min_layers_diag_terms>6</config_Redi_min_layers_diag_terms>
+<config_Redi_min_layers_diag_terms ocn_grid="ARRM10to60E2r1">15</config_Redi_min_layers_diag_terms>
 <config_Redi_horizontal_taper>'ramp'</config_Redi_horizontal_taper>
 <config_Redi_horizontal_ramp_min>20e3</config_Redi_horizontal_ramp_min>
 <config_Redi_horizontal_ramp_min ocn_grid="WCAtl12to45E2r4">30e3</config_Redi_horizontal_ramp_min>
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -47,6 +47,7 @@
 <config_dt ocn_grid="oRRS15to5">'00:03:45'</config_dt>
 <config_dt ocn_grid="oARRM60to10">'00:10:00'</config_dt>
 <config_dt ocn_grid="oARRM60to6">'00:05:00'</config_dt>
+<config_dt ocn_grid="ARRM10to60E2r1">'00:10:00'</config_dt>
 <config_dt ocn_grid="EC30to60E2r2">'00:30:00'</config_dt>
 <config_dt ocn_grid="WC14to60E2r3">'00:10:00'</config_dt>
 <config_dt ocn_grid="WCAtl12to45E2r4">'00:10:00'</config_dt>
@@ -68,6 +69,7 @@
 <config_hmix_scaleWithMesh ocn_grid="oRRS15to5">.true.</config_hmix_scaleWithMesh>
 <config_hmix_scaleWithMesh ocn_grid="oARRM60to10">.true.</config_hmix_scaleWithMesh>
 <config_hmix_scaleWithMesh ocn_grid="oARRM60to6">.true.</config_hmix_scaleWithMesh>
+<config_hmix_scaleWithMesh ocn_grid="ARRM10to60E2r1">.true.</config_hmix_scaleWithMesh>
 <config_hmix_scaleWithMesh ocn_grid="EC30to60E2r2">.true.</config_hmix_scaleWithMesh>
 <config_hmix_scaleWithMesh ocn_grid="WC14to60E2r3">.true.</config_hmix_scaleWithMesh>
 <config_hmix_scaleWithMesh ocn_grid="WCAtl12to45E2r4">.true.</config_hmix_scaleWithMesh>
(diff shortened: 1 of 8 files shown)


milenaveneziani (approved):
This looks good @darincomeau.
Let me add that I also tested this on cori: two simulations are running successfully with this MPAS mesh and with ne30 and the ArcticRRM for the atmosphere/land mesh.

darincomeau:
I've added the Do not merge tag until I can generate land ICs for these configurations.

components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
```diff
@@ -145,7 +148,7 @@
 <config_Redi_N2_based_taper_enable>.true.</config_Redi_N2_based_taper_enable>
 <config_Redi_N2_based_taper_min>0.1</config_Redi_N2_based_taper_min>
 <config_Redi_N2_based_taper_limit_term1>.true.</config_Redi_N2_based_taper_limit_term1>
-<config_Redi_min_layers_diag_terms>6</config_Redi_min_layers_diag_terms>
+<config_Redi_min_layers_diag_terms>15</config_Redi_min_layers_diag_terms>
```
xylar, line 151: If I undrestand correctly, this is changing Redi for all configurations that have GM on.  This needs to only be for `ARRM10to60E2r1`.  I believe this is correct:
``` xml
<config_Redi_min_layers_diag_terms>6</config_Redi_min_layers_diag_terms>
<config_Redi_min_layers_diag_terms ocn_grid="ARRM10to60E2r1">15</config_Redi_min_layers_diag_terms>
```

components/mpas-ocean/cime_config/buildnml
```diff
@@ -207,6 +207,17 @@ def buildnml(case, caseroot, compname):
             logger.warning("WARNING: The specified compset is requesting ocean ICs spunup from a G-case")
             logger.warning("         But no file available for this grid.")
 
+    elif ocn_grid == 'ARRM10to60E2r1':
+        decomp_date = '220730'
+        decomp_prefix = 'mpas-o.graph.info.'
+        restoring_file = 'sss.PHC3.0_monthlyClimatology.ARRM10to60E2r1.220802.nc'
+        analysis_mask_file = 'ARRM10to60E2r1_mocBasinsAndTransects220730.nc'
```
xylar, line 214: I don't think this file should have been given the date it was created.  The date stamp on the `mocBasinsAndTransects` files is a date stamp for the regions used in the masks, rather than the masks themselves.  This won't matter in E3SM but it will be a problem if the file to be added to`diagnostics` was modified in this way.  In general, I think it would be better to leave the file with the date stamp created in `compass`:
```
ARRM10to60E2r1_mocBasinsAndTransects20210623.nc
```

xylar (changes requested):
It seems like at least 2 changes are needed here.

In addition, I would request that the `diagnostics` files get copied into place from:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/anvil/test_20220730/ec_arrm_for_e3sm/ocean/global_ocean/ARRM10to60/PHC/files_for_e3sm/assembled_files/diagnostics
```
to
```
/lcrc/group/e3sm/public_html/diagnostics
```
Please make sure they are all in the E3SM group and have group read/write and world read permissions.

These files being available will significantly speed up MPAS-Analysis runs.

components/mpas-ocean/cime_config/buildnml
xylar: For a little more background, this date is controlled and updated in `geometric_features`:
https://github.com/MPAS-Dev/geometric_features/blob/master/geometric_features/aggregation/__init__.py#L56-L58

components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
darincomeau: Oops, that was definitely my mistake commit, I certainly should have included the grid specification. Thanks for the catch!
@milenaveneziani I just want to confirm that we do indeed want this separate setting for this mesh, before I make the change (or just revert the commit)?

components/mpas-ocean/cime_config/buildnml
darincomeau: Thanks for pointing this out. I just diffed the file with the `220730` datestamp and the `20210623` one and they are the same, so I must have `mv`ed this to match the other datestamps? I shouldn't have done that - I'll fix it.

milenaveneziani: we did go through a couple of mesh iterations between June and July though..

components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
milenaveneziani: yes, we want it at 15 for this mesh.

darincomeau:
>In addition, I would request that the diagnostics files get copied into place from:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/anvil/test_20220730/ec_arrm_for_e3sm/ocean/global_ocean/ARRM10to60/PHC/files_for_e3sm/assembled_files/diagnostics
```

@xylar Done -thanks for pointing out the missed step!

components/mpas-ocean/cime_config/buildnml
xylar: > we did go through a couple of mesh iterations between June and July though..
@milenaveneziani, that date stamp is from June 2021, not 2022.  It is the date that we added the separate Atlantic and Atlantic+Mediterranean MOC basins to geometric features and has nothing to do with when the file or the mesh itself was created.

xylar:
@darincomeau, I just synced the files over from Anvil to Cori.  Thanks for putting them in place!

I'm trying to run MPAS-Analysis (with `--polar_regions` this time to get some Arctic WOA analysis and to test the cached mapping and region mask files) but I'm getting permission denied. Can you give me access to the files in:
```
/global/cscratch1/sd/dcomeau/e3sm_scratch/cori-knl/20220803.WCYCL1950.ne30pg2_ARRM10to60E2r1.baseline.cori-knl/run/
```
?

darincomeau:
Opened - thanks for testing @xylar !

darincomeau:
@xylar the last two commits addressed your changes. As part of `d81b754` I also renamed the analysis_mask _file from the `220730` datestamp to the correct `20210623` one in the inputdata directory to avoid future confusion. (`rm` is 'rename', right? :D)

xylar:
Thanks @darincomeau.  As soon as I have permission, I'll rerun the analysis and then (hopefully) approve.

darincomeau:
@xylar that Cori run should be opened now.

xylar:
@darincomeau, ah, thanks.  I was expecting you to change group rather than make world readable so didn't notice. 

No, the `ARRM10to60E2r1_mocBasinsAndTransects220730.nc` file in `inputdata` is not the one used by analysis.  It has its own redundant copy that is stored elsewhere and does not use anything from `streams.ocean` to find it.  So we'll be fine.

xylar (approved):
@darincomeau and @milenaveneziani, I think this mesh is working okay and the necessary files are in place.  With @darincomeau's recent changes, I think the code changes look reasonable.

I reran MPAS-Analysis here:
https://portal.nersc.gov/project/e3sm/xylar/analysis/arrm/20220803.WCYCL1950.ne30pg2_ARRM10to60E2r1.baseline.cori-knl/ts_0011-0020_climo_0011-0020/
The mapping and mask files were taken from the `diagnostics` cache, as expected. 

There are some fairly serious temperature and especially salinity biases in the WOA18 plots.  I'm merely pointing those out as something that HiLat will presumably need to investigate.  It isn't clear to me whether changes in the mesh (e.g. resolution or passages) would have a mitigating effect.

darincomeau:
Thanks @xylar !

>There are some fairly serious temperature and especially salinity biases in the WOA18 plots. I'm merely pointing those out as something that HiLat will presumably need to investigate. It isn't clear to me whether changes in the mesh (e.g. resolution or passages) would have a mitigating effect.

Yeah that's been a pretty persistant problem, actually looks a little bit better with the new mesh. Here's an run with an older mesh for comparison:
https://portal.nersc.gov/project/m1199/milena/20210914.WCYCL1950.ne30pg2_oARRM60to10.hybrid/Years17-21/ocean/index.html#arctic_salinity_woa

xylar:
Okay, thanks @darincomeau.  Nothing new, then.  That's presumably good news, and we need to find a solution in the physics and parameterizations, not the mesh.

milenaveneziani:
@xylar: this is great. I believe the new mesh is so much better than the old one, and the new bathymetry is so much smoother. The B-cases with the low-res atmosphere look quite awful at the moment, but that has nothing to do with the mesh (imho). The coupled cases with the atmRRM mesh should look much better. 
Thanks for all the help on this new Arctic mesh!

darincomeau:
I've added a default land `finidat` (from a 50 year I-case spinup) for the `WCYCL1950.ne30pg2_ARRM10to60E2r1` configuration only, and ran a quick smoke test to make sure it was picked up in `lnd_in` and used.

darincomeau:
Removed the Do not merge flag - @jonbob I think this is ready whenever's a good time.

jonbob:
test merge passes:
* SMS_Ld1.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel.allactive-wcprod

verified that all new files are available on the inputdata repo

merged to next

jonbob:
merged to master

Issue #4996, opened 2022-05-31
NaN-value issue with Cray compiler CCE/14.0.0 on Crusher: ELM EcosystemBalanceCheckMod.F90


grnydawn:
Among 57 e3sm_develper test cases, 21 tests are passed and 36 failed.

Among 36 failed tests, 16 are failed with the issue of "NaN-value". The failed test cases are:
ERS.f09_g16.I1850ELMCN.crusher_crayclang
ERS.f09_g16.I1850GSWCNPRDCTCBC.crusher_crayclang.elm-vstrd
ERS.f19_f19.I1850ELMCN.crusher_crayclang
ERS.f19_f19.I20TRELMCN.crusher_crayclang
ERS.f19_g16.I1850CNECACNTBC.crusher_crayclang.elm-eca
ERS.f19_g16.I1850CNECACTCBC.crusher_crayclang.elm-eca
ERS.f19_g16.I1850CNRDCTCBC.crusher_crayclang.elm-rd
ERS.f19_g16.I1850GSWCNPECACNTBC.crusher_crayclang.elm-eca_f19_g16_I1850GSWCNPECACNTBC
ERS.f19_g16.I1850GSWCNPRDCTCBC.crusher_crayclang.elm-ctc_f19_g16_I1850GSWCNPRDCTCBC
ERS.f19_g16.I20TRGSWCNPECACNTBC.crusher_crayclang.elm-eca_f19_g16_I20TRGSWCNPECACNTBC
ERS.f19_g16.I20TRGSWCNPRDCTCBC.crusher_crayclang.elm-ctc_f19_g16_I20TRGSWCNPRDCTCBC
ERS.r05_r05.IELM.crusher_crayclang.elm-V2_ELM_MOSART_features
SMS.ne4_oQU240.F2010.crusher_crayclang.eam-cosplite
SMS.r05_r05.I1850ELMCN.crusher_crayclang.elm-qian_1948
SMS_Ld1.hcru_hcru.I1850CRUELMCN.crusher_crayclang
SMS_Ly2_P1x1.1x1_smallvilleIA.IELMCNCROP.crusher_crayclang.elm-lulcc_sville

This issue is related to https://github.com/E3SM-Project/E3SM/issues/4623

This error is occurred at the following error check routine  in “E3SM/components/elm/src/biogeochem/EcosystemBalanceCheckMod.F90”:
 
         if (abs(errcb_grc(g)) > balance_check_tolerance) then
            err_found = .true.
[...]

PR #5181, opened 2022-09-08
Add -hfp0 for ELM with cray compiler on Crusher

amametjanov:
Add -hfp0 for ELM with cray compiler on Crusher.
Also relax memleak tolerance from 10 to 20%.

Addresses E3SM-Project/E3SM#4996

[BFB]


amametjanov:
Recent crayclang run shows 65 pass, 46 fail: https://my.cdash.org/viewTest.php?buildid=2213678 .
Spot-checked this change and it helps avoid the NaNs.

Issue #5177, opened 2022-09-07
SCM runs retain constituents from initial condition


quantheory:
I've been experimenting with DYCOMS-II RF02 single-column simulations to get a feel for how drizzle is behaving in E3SMv2 (and its sensitivity to resolution), and I noticed some weird behaviors in the first few time steps that turned out to be due to extra data from the initial conditions file that I wasn't expecting. The cloud liquid field from the initial conditions was ignored because it was specified in the the IOP file, but the ice cloud and rain fields were read in from the `ncdata` file, and at shorter time steps there was enough there to "rain out" most of the initial cloud, which then takes some time to recover. This is a pretty mild effect at the default time step, but still lengthens the transient "spinup" behavior of the simulation a bit.

I think that what should happen here is that the constituent array should be initialized to zero (or qmin), except for those constituents that are specified in the IOP file. I implemented such a change, and the weird spinup behavior went away in the DYCOMS case.

I will try to put a PR together for this later today, adding a namelist variable to zero out these fields. Not sure what to name it; maybe `scm_ignore_non_iop_tracers`?

quantheory:
Tagging @bogensch

PR #5182, opened 2022-09-08
Add `scm_zero_non_iop_tracers` option.

quantheory:
For single-column runs, this sets all tracers in the atmospheric
constituent list to `qmin` (typically zero) after the initial condition
file (`ncdata`) is read in, but before the IOP file values are applied.
As a result, non-IOP tracers are initialized to `qmin` rather than
taking their values from the initial condition file, while IOP-defined
tracers are unaffected.

This option defaults to `.false.`, and hence does not change answers for
existing cases. However, it may be considered safer in the future to
change the default to `.true.` (at least for non-replay runs), to avoid
inconsistencies between the initial condition file state and the IOP
file state that exacerbate transient spin-up behavior and potentially
bias free-running model results. Like most single column model options,
this option has no effect in global runs.

This PR addresses #5177, at least in part. Actually changing default
behavior for existing SCM cases would require the option to be
enabled in `scmlib`.

### DIFF
--- a/components/eam/bld/namelist_files/namelist_definition.xml
+++ b/components/eam/bld/namelist_files/namelist_definition.xml
@@ -4095,6 +4095,13 @@ Default: FALSE
        group="cam_inparm" valid_values="">
 Turn off microphysics computation.
 Default: FALSE
+</entry>                                                                        |
+                                                                                |
+<entry id="scm_zero_non_iop_tracers" type="logical" category="scam"
+       group="cam_inparm" valid_values="">
+Ignore all tracers from initial conditions file, and default all tracers not
+specified in IOP to minimum value (usually zero).
+Default: FALSE
 </entry>
 
 <!-- Solar Parameters -->
--- a/components/eam/src/control/runtime_opts.F90
+++ b/components/eam/src/control/runtime_opts.F90
@@ -174,6 +174,7 @@ logical  :: scm_diurnal_avg
 logical  :: scm_crm_mode
 logical  :: scm_observed_aero
 logical  :: precip_off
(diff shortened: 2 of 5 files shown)


quantheory:
To understand the impact of this, in the DYCOMS case I was testing, here's 0-th time step rain mass mixing ratio (g/kg) vs altitude (m) with (blue) and without (orange) this option enabled:

image

This extra rain mass produces a brief but large increase in surface precipitation:

image

At shorter time steps (this is using 5 sec just to make the issue visible), you can see that without this option, there's rain very high up at the beginning of the simulation (even though there's no actual cloud above 900m):

image

This is the "mild" version of the problem, where I have modified the IOP file to zero out the initial ice cloud in both cases. If I run with the original IOP file, and run at a short time step, much of the boundary layer cloud in the initial condition rains out within minutes.

bogensch (approved):
Thanks for this PR @quantheory and bringing this issue to my attention.  I support keeping scm_zero_non_iop_tracers = .false. for this PR.  I will be doing some tests setting this to .true. and will issue a PR in the future to change to default once I'm comfortable with it.

quantheory:
@whannah1 Just curious if you've looked at this, though I'm not really sure if this is relevant to you right now.

wlin7:
@quantheory , the changes look good to me. Thanks for exposing and investigating the problem and putting in the fix. I think it is ready to be merged and I will do so if next is cleared tomorrow. @whannah1 can still take a look when having a chance.

wlin7:
Merged to next.

wlin7:
Merged to master.

PR #5185, opened 2022-09-08
For GCP only, remove debug flags for gnu which checks for invalid values

### LINKED ISSUE #5184: SIGFPE: Floating-point exception on GCP for SMS_D_Ld1.ne30pg2_EC30to60E2r2.WCYCLSSP370.gcp_gnu.allactive-wcprodssp
From https://github.com/E3SM-Project/E3SM/issues/5077, we hit a floating-point issue on GCP with one of the tests.

```
120: Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.
...
120: #0  0x2ae5d39823ff in ???
120: #1  0x26ef7d1 in hist_update_hbuf_field_1d
120:    at /home/noel/wacmy/nexty-sep6/components/elm/src/main/histFileMod.F90:1218
120: #2  0x26fbe76 in __histfilemod_MOD_hist_update_hbuf._omp_fn.0
120:    at /home/noel/wacmy/nexty-sep6/components/elm/src/main/histFileMod.F90:1010
120: #3  0x2ae5d30ab395 in gomp_thread_start
120:    at /tmp/root/spack-stage/spack-stage-gcc-11.1.0-bm2qpp7qjfzn3evldu54qpltiwaqh2ue/spack-src/libgomp/team.c:125
120: #4  0x2ae5d3737ea4 in ???
120: #5  0x2ae5d3a4a9fc in ???

```

`SMS_D_Ld1.ne30pg2_EC30to60E2r2.WCYCLSSP370.gcp_gnu.allactive-wcprodssp` fails, while
`SMS_D_Ld1.ne30pg2_EC30to60E2r2.WCYCL1850.gcp_gnu.allactive-wcprod` does not.

Also, both of these tests complete the 1 day if run without DEBUG.

Looking closer, I'm reminded that I included the `invalid` check for DEBUG gnu builds on this machine as the defaults for gnu does not use it. That is, gcp was using `-ffpe-trap=invalid,zero,overflow`, but when I change to `-ffpe-trap=zero,overflow`, the tests pass.

Now, I still see NaN values printed in my debug prints as described in the other issue noted. Apparently, these NaN values are otherwise not an issue and will only fail when testing for invalid.  Of course, I believe we should fix all of those, but we know there is code in e3sm (esp LND) that actually sets arrays to NaN on purpose.

`ndk/machinefiles/gcp-remove-invalid-check`

ndkeen:
For GCP-only change the GNU debug flags from `-ffpe-trap=invalid,zero,overflow` to `-ffpe-trap=zero,overflow` which is the current default for GNU builds anyway.

Fixes #5184 
[bfb]


ndkeen:
merged to next

ndkeen:
The nightly test now completes, but there are many fails right now because of baseline issues.  Will merge to master

PR #5187, opened 2022-09-12
Fix out-of-bounds index for vertical mixing with one active layer

### LINKED ISSUE #4849: Implicit vertical mixing routines cause out of bounds error for single layer
This was discovered in recently developed single layer configurations used for barotropic tidal modeling under the ICoM project. The error appears to occur after the in-lining of the tridiagonal solves in the vertical mixing routines in #4680. For a single layer, this should solve a 1x1 matrix by simply dividing  the rhs by the diagonal value. In the code below, `layerThicknessEdge(Nsurf+1,iEdge)` causes the initial out of bounds error.

https://github.com/E3SM-Project/E3SM/blob/b404a464a55dbdad1f13bdc63ff7b6077527426e/components/mpas-ocean/src/shared/mpas_ocn_vmix.F#L857-L862

mark-petersen:
As described in #4849, the current implicit vertical mixing routine fails when a column has one active layer, because of references to $k+1$ and $k-1$. This PR adds a check on each edge, and if there is only one active layer, it applies the implicit drag term but not the vertical mixing terms. When only a single layer exists, implicit vertical mixing is not needed (or even valid), but implicit drag is still desirable.

There are two cases where the model may have a single layer: 
1. when the whole domain is single layer; and 
2. when individual columns have only one active vertical cell, like near coastlines with wetting-and-drying. 

(1) could have been solved with an if statement on `nVertLevels.eq.1`  outside the edge loop. I decided to put the if statement inside the edge loop to cover both cases, to future-proof this problem for coastal applications. This leads to a potential performance hit in the vertical mixing routine. However, there was no measurable impact to the vmix timers on a 3-day EC60to30 test on chrysalis, as detailed below.

Fixes #4849
[BFB]

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_vmix.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_vmix.F
@@ -460,45 +460,53 @@ contains
          cell1 = cellsOnEdge(1,iEdge)
          cell2 = cellsOnEdge(2,iEdge)
 
-         ! tridiagonal matrix algorithm
-         C(Nsurf)     = -2.0_RKIND*dt*vertViscTopOfEdge(Nsurf+1,iEdge) &
-                    / (layerThickEdgeMean(Nsurf,iEdge) + layerThickEdgeMean(Nsurf+1,iEdge)) &
-                    / layerThickEdgeMean(Nsurf,iEdge)
-         bTemp(Nsurf) = 1.0_RKIND - C(Nsurf)
-         rTemp(Nsurf) = normalVelocity(Nsurf,iEdge)
+         ! one active layer
+         if (N .eq. Nsurf) then
+            normalVelocity(N,iEdge) = normalVelocity(N,iEdge)  &
+                / (1.0_RKIND + dt*implicitBottomDragCoef &
+                   * sqrt(kineticEnergyCell(N,cell1) + kineticEnergyCell(N,cell2)) / layerThickEdgeMean(N,iEdge) )
+         else
 
-         ! first pass: set the coefficients
-         do k = Nsurf+1, N-1
-            A        = -2.0_RKIND*dt*vertViscTopOfEdge(k,iEdge) &
-                   / (layerThickEdgeMean(k-1,iEdge) + layerThickEdgeMean(k,iEdge)) &
-                   / layerThickEdgeMean(k,iEdge)
-            m        = A/bTemp(k-1)
-            C(k)     = -2.0_RKIND*dt*vertViscTopOfEdge(k+1,iEdge) &
-                   / (layerThickEdgeMean(k,iEdge) + layerThickEdgeMean(k+1,iEdge)) &
-                   / layerThickEdgeMean(k,iEdge)
-            bTemp(k) = 1.0_RKIND - A - C(k) - m*C(k-1)
-            rTemp(k) = normalVelocity(k,iEdge) - m*rTemp(k-1)
-         enddo
+           ! tridiagonal matrix algorithm
+           C(Nsurf)     = -2.0_RKIND*dt*vertViscTopOfEdge(Nsurf+1,iEdge) &
+                      / (layerThickEdgeMean(Nsurf,iEdge) + layerThickEdgeMean(Nsurf+1,iEdge)) &
+                      / layerThickEdgeMean(Nsurf,iEdge)
+           bTemp(Nsurf) = 1.0_RKIND - C(Nsurf)
+           rTemp(Nsurf) = normalVelocity(Nsurf,iEdge)
 
-         A = -2.0_RKIND*dt*vertViscTopOfEdge(N,iEdge) &
-           / (layerThickEdgeMean(N-1,iEdge) + layerThickEdgeMean(N,iEdge)) &
-           / layerThickEdgeMean(N,iEdge)
-         m = A/bTemp(N-1)
+           ! first pass: set the coefficients
+           do k = Nsurf+1, N-1
+              A        = -2.0_RKIND*dt*vertViscTopOfEdge(k,iEdge) &
+                     / (layerThickEdgeMean(k-1,iEdge) + layerThickEdgeMean(k,iEdge)) &
+                     / layerThickEdgeMean(k,iEdge)
+              m        = A/bTemp(k-1)
+              C(k)     = -2.0_RKIND*dt*vertViscTopOfEdge(k+1,iEdge) &
+                     / (layerThickEdgeMean(k,iEdge) + layerThickEdgeMean(k+1,iEdge)) &
+                     / layerThickEdgeMean(k,iEdge)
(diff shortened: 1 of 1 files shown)


mark-petersen:
Most of the lines changed here are indentation. If you haven't tried this before, check 
image
Hide whitespace under the gear, and you will see the relevant code additions.

mark-petersen:
Passes stand-alone nightly suite with gnu and intel on badger debug and optimized, including bfb match with master for optimized.

For performance, I tested on chrysalis:
```
./create_test SMS_Ld3.T62_EC30to60E2r2.GMPAS-IAF.chrysalis_gnu -q acme-medium --walltime 30:00
```
in the run directory, we can see 
```
grep vmix timing/model_timing_stats # master
name                                         on  processes  threads    count          walltotal       wallmax (proc   thrd  )   wallmin (proc   thrd  )
"o:vmix imp"                                     -        320      320 4.576000e+04   5.262103e+02     2.006 (   293      0)     0.741 (   250      0)
grep vmix timing/model_timing_stats # this PR
"o:vmix imp"                                     -        320      320 4.576000e+04   5.261252e+02     2.002 (   293      0)     0.740 (   250      0)
```
and timing differences are in the fourth digit, and this PR is (coincidentally) faster than master by 0.2%.

xylar:
I don't feel qualified to review.  @cbegeman may wish to.

xylar:
I would be willing to do compass testing, though, if you would like.

cbegeman:
@mark-petersen Thanks for working on this! I'll test this with the single- and multi-layer drying slope test case.

knbarton:
Hi Mark, I tested this with my standard namelist options on icos7 with scalar SAL on and it ran without any error. However, I did get slightly different results for the final M2 RMS error (0.11600074112907378 vs 0.11600565831875714). I am not sure if this is expected or not.

sbrus89:
@mark-petersen, I found that this was BFB with my previous results for the icos7 single layer tides run using parallel SAL on compy.

sbrus89 (approved):

rljacob:
@dengwirda please review if you have time.

cbegeman:
@mark-petersen, @sbrus89 and I discussed this and we think it's possible that his tides case and my drying slope case are using different drag implementations. It seems that only the drying slope case uses Rayleigh drag. Let me know if/how I can help.

mark-petersen:
@cbegeman thanks for the feedback. I can see I made a mistake on the Rayleigh drag version. Just to explain, the standard version of the tri-diagonal solve is here:
```
 386    subroutine ocn_vel_vmix_tend_implicit(dt, kineticEnergyCell, vertViscTopOfEdge, layerThickness, & !{{{
 387                                          layerThickEdgeMean, normalVelocity, err)
...
 477            ! first pass: set the coefficients
 478            do k = Nsurf+1, N-1
 479               A        = -2.0_RKIND*dt*vertViscTopOfEdge(k,iEdge) &
 480                      / (layerThickEdgeMean(k-1,iEdge) + layerThickEdgeMean(k,iEdge)) &
 481                      / layerThickEdgeMean(k,iEdge)
 482               m        = A/bTemp(k-1)
 483               C(k)     = -2.0_RKIND*dt*vertViscTopOfEdge(k+1,iEdge) &
 484                      / (layerThickEdgeMean(k,iEdge) + layerThickEdgeMean(k+1,iEdge)) &
 485                      / layerThickEdgeMean(k,iEdge)
 486               bTemp(k) = 1.0_RKIND - A - C(k) - m*C(k-1)
 487               rTemp(k) = normalVelocity(k,iEdge) - m*rTemp(k-1)
 488            enddo
```
For multiple layers it is the `k+1` and `k-1` indices that cause a problem. But for single layer, applying vertical mixing does not make sense anyway, so `vertViscTopOfEdge=0` and this all reduces to:
```
A = 0
m = 0
C(k) = 0
bTemp(k) = 1.0
rTemp(k) = normalVelocity(k,iEdge)
```
So then the line I added for single layer is this:
```
 498            normalVelocity(N,iEdge) = (normalVelocity(N,iEdge) - m*rTemp(N-1)) &
 499                / (1.0_RKIND - A + dt*implicitBottomDragCoef &
 500                * sqrt(kineticEnergyCell(N,cell1) + kineticEnergyCell(N,cell2)) / layerThickEdgeMean(N,iEdge) &
 501                - m*C(N-1))
 502            ! second pass: back substitution
 503            do k = N-1, Nsurf, -1
 504               normalVelocity(k,iEdge) = (rTemp(k) - C(k)*normalVelocity(k+1,iEdge)) / bTemp(k)
 505            enddo
```
with all those zeros and ones substituted in, it collapses to just this for the single layer:
```
 465             normalVelocity(N,iEdge) = normalVelocity(N,iEdge)  &
 466                 / (1.0_RKIND + dt*implicitBottomDragCoef &
 467                    * sqrt(kineticEnergyCell(N,cell1) + kineticEnergyCell(N,cell2)) / layerThickEdgeMean(N,iEdge) )
```

The thing I missed is that for the Rayleigh drag version the denominator `bTemp` also has an extra divisor, so it is not 1.0:
```
           bTemp(Nsurf) = 1.0_RKIND - C(Nsurf) &
              ! added Rayleigh terms
              + dt*rayleighDampingCoef/((1.0_RKIND - rayleighDepthVariable) + rayleighDepthVariable*edgeThicknessTotal)
```
so I need to correct that. More soon...

mark-petersen:
After reviewing this in more detail, I think the code for single-layer Rayleigh drag is correct.

@cbegeman said [above](https://github.com/E3SM-Project/E3SM/pull/5187#issuecomment-1246826405):
> I'm getting undesirable results with the drying_slope/1km/single-layer/default test case using the same Rayleigh friction as the sigma test case rather than the explicit bottom drag. Can you confirm whether this option should be supported or whether there is a setting in the namelist file I should change? Thanks!

The trouble was that the config flag `config_disable_vel_vmix = .true.` in the single layer case, and the implicit drag is applied in the vertical mixing routine. When that is `false` then the single-layer case with Rayleigh drag on this PR (right) matches the multi-layer sigma with Rayleigh drag (left) for the two drag values: 
image

I guess `config_disable_vel_vmix` could be made more clear, since it carries along more than just vertical mixing. Maybe `config_disable_vel_implicit_terms`, but I think the current name is probably OK, and not relevant to this PR anyway.

cbegeman:
@mark-petersen Thank you so much for catching that config option change! It's great that no code changes were needed.

dengwirda (approved):
Approving based on re-running our 45-to-5km tides config.,which reproduces the deep-water, complex M2 RMSE of 5.13cm obtained previously. Successfully tested with both `config_use_implicit_bottom_roughness` and `config_use_implicit_bottom_drag`.

jonbob:
passes:

merged to next

jonbob:
merged to master

PR #5194, opened 2022-09-21
Replace device_resident and other OpenACC fixes for ocean

### LINKED ISSUE #5006: Cray compiler on Crusher fails to create device array when using OpenACC declare device_resident
In a few MPAS routines, arrays were declared device_resident. At run time on Crusher, this resulted in an invalid dope vector. With CRAY_ACC_DEBUG=3, it looks like these arrays never get allocated on the device so the device pointer is 0.  Workaround is to just use usual data create/delete directives instead and I will push a PR with that replacement soon.

@sarats  @mattdturner, @twhite-cray and @abbotts Already reported via OLCF help but reproduced here for tracking. 
A simple reproducer is here - gives a different error message, but for the same reason:

  program deviceResidentBug 

      implicit none

      ! This is a bug reproducer for an issue with the use of the
      ! OpenACC declare device_resident directive

      integer :: i,j,nx,ny,ierr

      double precision, dimension(:), allocatable :: &
         XX, YY

      double precision, dimension(:,:), allocatable :: &
         Ar
      !$acc declare device_resident(Ar)

      call MPI_INIT(ierr)
      nx = 512
      ny = 512

      allocate (XX(nx), YY(ny))
      allocate (Ar(nx,ny))

      do i=1,nx
         XX(i) = i
      end do
      YY(:) = 0.0d0
      !$acc enter data copyin(XX,YY)

      !*** device resident version returns errors that
      !*** according to debug has no device pointer for Ar

      ! Ar was declared device_resident so shouldn't need a data directive 
      ! or present clause, tho did try the latter with no difference

      !$acc parallel loop collapse(2)
      do j=1,ny
      do i=1,nx
         Ar(i,j) = i+j
      end do
      end do

      !$acc parallel loop present(XX, YY)
      do j=1,ny
      do i=1,nx
         YY(j) = Ar(i,j)*XX(i)
      end do
      end do

      !$acc exit data copyout(YY) delete(XX)

      call MPI_FINALIZE(ierr)

   end program deviceResidentBug

### LINKED ISSUE #5176: OpenACC pragma cleanup needed in some mpas-ocean files
PR #5099 uncovered some OpenACC issues in subroutine ocn_diagnostic_solve_vortVel in file mpas_ocn_diagnostics.F. And @philipwjones noticed similar problems in sister routine diagnostic_solve_MLEvel

philipwjones:
The OpenACC declare device resident was not universally supported so replaced these with the more standard create/delete pairs. Also fixed some other missing private declarations that were possibly resulting in race conditions and bad results. Cleaned up some long lines, removed meshPool and other things along the way.

This was bit-for-bit on a QU240 test on Summit for both CPU and GPU. Also passes the compass pr tests with Intel on Chrysalis.

[bfb]
Fixes #5006 
Fixes #5176

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
@@ -2606,91 +2606,81 @@ contains
 !>  cell.
 !
 !-----------------------------------------------------------------------
-   subroutine ocn_vert_transport_velocity_top(meshPool, verticalMeshPool, oldLayerThickness, layerThicknessEdgeFlux, &
-     normalVelocity, oldSSH, dt, vertAleTransportTop, err, newHighFreqThickness)!{{{
+
+   subroutine ocn_vert_transport_velocity_top(verticalMeshPool, &
+                     oldLayerThickness, layerThicknessEdgeFlux, &
+                     normalVelocity, oldSSH, dt, vertAleTransportTop, &
+                     err, newHighFreqThickness)!{{{
 
       !-----------------------------------------------------------------
-      !
       ! input variables
-      !
       !-----------------------------------------------------------------
 
       type (mpas_pool_type), intent(in) :: &
-         meshPool           !< Input: horizonal mesh information
-
-      type (mpas_pool_type), intent(in) :: &
-         verticalMeshPool   !< Input: vertical mesh information
-
-      real (kind=RKIND), dimension(:,:), intent(in) :: &
-         oldLayerThickness    !< Input: layer thickness at old time
-
-      real (kind=RKIND), dimension(:,:), intent(in) :: &
-         layerThicknessEdgeFlux     !< Input: layerThickness interpolated to an edge
-
-      real (kind=RKIND), dimension(:,:), intent(in) :: &
-         normalVelocity     !< Input: transport
+         verticalMeshPool         !< [in] vertical mesh information
 
       real (kind=RKIND), dimension(:), intent(in) :: &
-         oldSSH     !< Input: sea surface height at old time
+         oldSSH                   !< [in] sea surface height at old time
 
+      real (kind=RKIND), dimension(:,:), intent(in) :: &
+         oldLayerThickness,      &!< [in] layer thickness at old time
+         layerThicknessEdgeFlux, &!< [in] layerThickness at edge
+         normalVelocity           !< [in] transport
+
+      !  alters ALE thickness when freq-filtering
       real (kind=RKIND), dimension(:,:), intent(in), optional :: &
-         newHighFreqThickness   !< Input: high frequency thickness.  Alters ALE thickness.
+         newHighFreqThickness     !< [in] high frequency thickness
 
       real (kind=RKIND), intent(in) :: &
          dt     !< Input: time step
 
       !-----------------------------------------------------------------
-      !
       ! output variables
-      !
       !-----------------------------------------------------------------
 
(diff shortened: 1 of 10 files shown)


components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
```diff
@@ -1671,278 +1730,315 @@ end subroutine ocn_diagnostic_solve_GMvel!}}}
...
+            vertTransportVelocityTop(k+1,iCell) - div_huTransport(k)
          end do
       end do
 #ifndef MPAS_OPENACC
       !$omp end do
-      !$omp end parallel
 #endif
 
+      !$acc exit data delete(div_hu, div_huTransport)
       deallocate(div_hu,div_huTransport)
```
mark-petersen, line 2015: @philipwjones when I run this with two OpenMP threads, I get the error 
```
217 Error termination. Backtrace:
218 At line 2015 of file mpas_ocn_diagnostics.F
219 Fortran runtime error: Attempt to DEALLOCATE unallocated 'div_hu'
```
but it works fine for single thread cases. Is the allocate/deallocate nesting with OpenMP incorrect?

mark-petersen:
Other than the above error, this PR matches bfb against master using both gnu debug and gnu optimized with the nightly suite, and passes with intel.

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
philipwjones: Sorry about that - looks like I was going to merge the two parallel regions, but forgot to deal with the intervening code between loops. Just committed a change to reinsert the parallel directives as before.

mark-petersen (approved):
This now passes nightly suite and matches master bfb using gnu debug; gnu optimized; and intel optimized.

rljacob:
telecon notes:  ready to go.

jonbob:
a test merge showed non-BFB answers for SMS_D_Ld3.T62_oQU120.CMPASO-IAF.chrysalis_intel. Is this expected? It is a threaded test that seems to find issues that other tests do not

jonbob:
@philipwjones - never mind, that test is failing again on master. I'll try to track down the issue and merge this as well

jonbob:
The threading issue that causes the SMS_D_Ld3.T62_oQU120.CMPASO-IAF.chrysalis_intel is hopefully fixed by [PR #5226](https://github.com/E3SM-Project/E3SM/pull/5226). Testing of this branch passes and is BFB for:

merged to next

amametjanov:
This case errors out due to NaNs: https://my.cdash.org/viewTest.php?onlydelta&buildid=2226656
- SMS_Ld1.T62_oEC60to30v3.CMPASO-NYF.ascent_pgigpu

jonbob:
@amametjanov - do you want me to revert this? Or can you create an issue for this failure and I'll go ahead and merge this to master, because it doesn't create issues elsewhere

amametjanov:
Let me check if Crusher+cray is okay with that case.
I'll add nightly `gpuacc` runs on Crusher too.

jonbob:
Thanks @amametjanov -- and please let me know

jonbob:
@amametjanov - did you get a chance to check on crusher? I'd like to either merge this or revert it

amametjanov:
I was able to fix the issue on Ascent and push to this branch. Please re-merge to next. Thanks.

jonbob:
Thanks @amametjanov!

jonbob:
merged to master

PR #5172, opened 2022-09-02
Enables and modifies the submesoscale eddy parameterization

vanroekel:
This enables the Fox-Kemper et al 2011 submesoscale eddy parameterization as the default for all configurations.

It also fixes a unit error in the submesoscale eddy code that did not influence the long control run, but was introduced in the first PR (#5099).  The variable gradBuoyEddy was not the buoyancy but the density gradient.  Here the name is changed to gradDensityEddy for clarity.  

[NML]
[CC]

PR #5195, opened 2022-09-21
Fix barotropic thickness on edge in split explicit subcycling

mark-petersen:
Barotropic thickness at the edge must match the sum of baroclinic thicknesses at that edge. This PR corrects a long-standing problem of noise in the `vertVelocityTop` variable, which is the Eulerian (fixed-frame) velocity for momentum. The `vertTransportVelocityTop` variable for tracer and thickness transport was smooth, both before and after this change.

Currently in the code, the thickness at an edge used for volume (SSH) transport in the barotropic subcycling is
```
thicknessSum = sshEdge + min(bottomDepth(cell1), bottomDepth(cell2))
```
That is, the thickness at an edge is taken as the shallower of the two neighboring cells. This was inconsistent with the baroclinic mode, which takes the average thickness of the two neighboring cells. 

This PR is climate changing.

[CC]

### DIFF
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F
@@ -795,12 +795,10 @@ module ocn_time_integration_split
             !$omp end do
             !$omp end parallel
 
-!mrp note, not sure if I need nEdgesAll here, or nEdgesArray(3)?
-!mrp note, if I have bfb partition problems, it could be this section.
             !$omp parallel
             !$omp do schedule(runtime) &
-            !$omp private(k)
-            do iEdge = nEdgesOwned+1, nEdgesArray(3)
+            !$omp private(k, thicknessSum)
+            do iEdge = nEdgesOwned+1, nEdgesArray(4)
                cell1 = cellsOnEdge(1,iEdge)
                cell2 = cellsOnEdge(2,iEdge)
                thicknessSum = layerThickEdgeFlux(minLevelEdgeBot(iEdge),iEdge)
@@ -1111,7 +1109,7 @@ module ocn_time_integration_split
 
                !$omp parallel
                !$omp do schedule(runtime) &
-               !$omp private(i, k, iEdge, cell1, cell2, sshEdge, &
+               !$omp private(i, iEdge, cell1, cell2, sshEdge, &
                !$omp         thicknessSum, flux)
                do iCell = 1, nCells
                   sshTend(iCell) = 0.0_RKIND
@@ -1126,13 +1124,8 @@ module ocn_time_integration_split
                                           sshSubcycleCur(cell2))
 
                      ! Compute barotropic thickness at the edge. Note this
-                     ! must match the sum of baroclinic thicknesses at this
-                     ! edge.
-                     ! This version works with pbcs, but only for z-level and z-star
-                     k = maxLevelEdgeTop(iEdge)
-                     thicknessSum = sshEdge + 0.5_RKIND &
-                        *(  min(bottomDepth(cell1), refBottomDepth(k)) &
-                          + min(bottomDepth(cell2), refBottomDepth(k)) )
+                     ! matches the sum of baroclinic thicknesses at this edge.
+                     thicknessSum = sshEdge + bottomDepthEdge(iEdge)
 
                      flux = ((1.0-config_btr_gam1_velWt1)* &
                             normalBarotropicVelocitySubcycleCur(iEdge) &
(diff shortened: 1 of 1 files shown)


mark-petersen:
The change on each edge for the thickness used for barotropic transport is up to half a layer, so 250/2 = 125m in deep cells:
bottomDepthSketch
Note this problem occurs when using the combination of partial bottom cells and split time stepping, which is the case for all our standard runs.

mark-petersen:
before and after:
image
image

mark-petersen:
Thanks to @dengwirda for discussions on this topic, and for questioning that line that specifies the barotropic thickness.

mark-petersen:
For documentation, I'll add here that noise in the vertical velocity does still exist, but to a much lower extent. Here is a simulation with the most plain options I could imagine:

- RK4
- z-level
- no partial bottom cells
- GM off
- Redi off
- config_pressure_gradient_type = 'pressure_and_zmid'

This is 57 days into the simulation:
image
note colorbar is tighter than the images above. Images with split-explicit look similar.

dengwirda:
@mark-petersen and all, a QU30km MPAS-O standalone spin-up (4 years) confirms the improvement here, with `vertVelocityTop` now matching `vertAleTransportTop` (for a `z-star` coord.). Prior to this PR, `vertVelocityTop` contained significant grid-scale noise, due to the 2d vs 3d thickness incompatibility discussed above. 

This change does not appear to degrade the energetics of the flow, with both the `max` and `rms` kinetic energy in the same range before and after this change. (Spurious damping of `K` had been a problem with other attempts to fix this issue).

Note that while we're focusing on improvements to `vertVelocityTop` here, this PR actually changes the way the (split) horizontal dynamics evolve, so changes are not isolated to vertical diagnostics. `vertVelocityTop` is simply a place where the change is most visible.

New `|u|`, `vertVelocityTop` and `vertAleTransportTop`:

qu30_speed_new
qu30_vert_vel_top_new
qu30_vert_ale_top_new

Old `vertVelocityTop` and `vertAleTransportTop`:

qu30_vert_vel_top_old
qu30_vert_ale_top_old

dengwirda:
Sticking with the question of barotropic vs baroclinic compatibility in the split integration schemes, we should also be careful with the thickness forcing terms `S_h`:

`dh/dt + div(u*h) + dw/dz = S_h.`

The rhs `S_h` term contains any thickness fluxes in the coupled system (runoff, evap., precip., etc).

When we integrate vertically to get the barotropic subsystem (for the split subcycles), I don't believe the forcing drops out:

`dH/dt + div(U*H) = S_H, with S_H = SUM_layers(h*S_h) / SUM_layers(h).`

Currently, we do take similar care for the forcing term for momentum (`barotropicForcing`), but not for thickness. Can we do:

```
barotropicForcingU = SUM_layers(h*S_u) / SUM_layers(h),
barotropicForcingH = SUM_layers(h*S_h) / SUM_layers(h),
```

to be consistent. The 2d thickness forcing is probably fairly small in most cases, but this seems like another opportunity for a 2d vs 3d thickness mismatch as per the PBC issue above. Would require coupled E3SM runs to test, not MPAS-O standalone where `S_h = 0` usually.

jonbob:
I ran a 30-year case with this PR test-merged into recent next (and including PR #5172). MPAS-Analysis output comparing this run with a previous one which also included PR #5172 is available at:

[20220929.PR5195.chrysalis](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20220929.PR5195.chrysalis/html/)

If anyone would prefer a direct comparison with obs as well, please let me know. @xylar and @dengwirda, including you in this review as well - thanks

vanroekel:
The changes here look mostly minor or to within the noise of different ensemble members.  As @maltrud notes AMOC may be a bit stronger.  This is definitely a bug and the fix is obvious.  I'll approve this now.

vanroekel (approved):
Approving based on visual inspection and testing from @jonbob and @mark-petersen

jonbob:
Thanks @maltrud

jonbob:
also analysis vs obs:

[20220929.PR5195.chrysalis_vs_obs1-30](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20220929.PR5195.chrysalis_vs_obs1-30/html/)

mark-petersen:
Thanks @jonbob for the testing. With this PR, the AMOC is a bit stronger, but longer runs may turn out similar. 
image
Other statistics in the 30-year run also look similar between master and this PR, for example ocean heat content and transports. At a glance:
image

jonbob:

jonbob:

PR #5196, opened 2022-09-21
Use reduced g-points for RRTMGP in EAM/MMF

brhillman:
Use reduced g-point data when using RRTMGP in EAM and E3SM-MMF. This should half the cost of RRTMGP in E3SM and bring performance in line with RRTMG, which was already using a similar number of quadrature points. This will be non-BFB only for the tests that use RRTMGP (eam-rrtmgp/eam-rrtmgpxx testmods, and all of the MMF tests), but BFB for all other tests.

[non-BFB] for RRTMGP tests

### DIFF
--- a/components/eam/bld/configure
+++ b/components/eam/bld/configure
@@ -2617,6 +2617,7 @@ sub write_filepath
     } elsif ($rad eq 'rrtmgp') {
         print $fh "$camsrcdir/eam/src/physics/rrtmgp\n";
         if (not defined $opts{'rrtmgpxx'} ) {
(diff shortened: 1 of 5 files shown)


whannah1 (approved):
changes look good, I'll start some test runs on Summit.

whannah1:
I ran a pair of 1-year simulations with and without the reduced g-point mod. The differences seem small enough to be considered safe in my opinion.

image

whannah1:
image

whannah1:
image

brhillman:
Merged to next.

PR #5201, opened 2022-09-23
Remove unneeded mesh declaration

### LINKED ISSUE #5200: Segfault when running MPAS-Ocean with serial SAL
There is a segfault when running MPAS-Ocean with serial SAL due to the following line in `mpas-ocean/src/shared/mpas_ocn_vel_self_attraction_loading.F` : 

```
630          integer, dimension(:), pointer :: indexToCellID
```

This variable no longer needs to be defined in the routine and doing so causes the segfault.

mark-petersen:
Mesh variables are declared in `mpas_ocn_mesh.F` rather than individual routines.

Fixes #5200
[BFB]

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_vel_self_attraction_loading.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_vel_self_attraction_loading.F
@@ -627,7 +627,6 @@ contains
 
         ! MPI variables
         integer :: curProc
-        integer, dimension(:), pointer :: indexToCellID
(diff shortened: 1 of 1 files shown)


mark-petersen:
@knbarton please confirm that this fixes your problem.

knbarton (approved):
I tested this and it now runs with scalar SAL and no error.

jonbob:
passes:

merged to next

jonbob:
merged to master

PR #5147, opened 2022-08-22
Add new COSP output for MODIS simulator

zyuying:
Add new output for MODIS simulator. The three variables are the in-cloud total/liquid/ice cloud fraction ('CLTMODISIC', 'CLWMODISIC', and 'CLIMODISIC'). These cloud fraction will be used to convert the fraction-weighted cloud properties from MODIS simulator to the in-cloud average. These MODIS fraction-weighted cloud properties include 'LWPMODIS', 'IWPMODIS', 'REFFCLIMODIS', 'REFFCLWMODIS', 'TAUTMODIS', 'TAUIMODIS', 'TAUWMODIS', 'TAUTLOGMODIS', 'TAUILOGMODIS', and 'TAUWLOGMODIS'.

[BFB] tests with cosp enabled will add fields.

PR #5203, opened 2022-09-25
Fix inconsistent string length in a cosp character array

wlin7:
An initialized character array does not have the same length for some
elements (newly introduced cosp variables via #5147), causing gnu compiler to abort.

[BFB]


wlin7:
Merged to next.

wlin7:
The change is a simple space padding, no review needed.

wlin7:
Merged to master.

PR #5202, opened 2022-09-25
Add CF-compliant Time coordinate

xylar:
This variable has CF-compliant units and calendar. To accommodate the CF-compliant units, a new config option has been added to supply the reference time, 0001-01-01 by default.

This merge also adds a `Time_bnds` variable.  For output from a single point in time, the values in `Time_bnds` are the same as `Time`.  For time-averaged (or min or max) output, `Time_bnds` contains the start and end times of the interval covered by the output.

See https://github.com/E3SM-Ocean-Discussion/E3SM/pull/19 for more discussion

[BFB]

PR #5218, opened 2022-10-05
Fix two bugs in timeSeriesStats output in MPAS-Ocean

### LINKED ISSUE #121: something not right with restart of timeSeriesStats(Monthly)
I'm running the ISOMIP+ test cases and recently turned on `timeSeriesStatsMonthly`.  I run MPAS-Ocean repeatedly for one month, then update the forcing with a python script.  Initially, I was also writing out restart files for `timeSeriesStatsMonthly`.  It appears that `xtime_startMonthly` never gets updated in these circumstances (it is always the start of the simulation) and that none of the variables being written out get reset at the beginning of a month either.  Yesterday, I re-ran without timeSeriesStatsMonthly restart files (which are not needed when the restart interval is a multiple of the compute interval) and `xtime_startMonthly` is now updated each month as expected.

I suspect, but haven't proven, that there is something bigger that is wrong with how alarms work on restart, and that this may related to https://github.com/E3SM-Project/E3SM/issues/2499.  My suspicion is that certain alarms are not triggered immediately on restart (or init) that *would* be triggered if the simulation continued without a restart.  In the case of `timeSeriesStatsMonthly`, I think the alarm is not being triggered to zero out averaging variables and `xtime_startMonthly` is not being reset.  It is not yet clear to me if this should happen immediately before restart files are written or immediately after they are read back in, but one of these would be the expected behavior.

I have a workaround and the same is true of E3SM as currently configured but this bug should be fixed, particularly if it indicates a bigger problem with alarms and how they interact with restart/init.

### LINKED ISSUE #5217: Start time in timeSeriesStats output from MPAS-Ocean not correct after restart
There has been a long-standing bug that the value of `xtimeMonthly_start` is incorrect in the first `timeSeriesStatsMonthly` output file after a restart.  This issue is discussed further here:
https://github.com/E3SM-Project/E3SM/pull/5202#issuecomment-1257252115
in the context of the new CF-compliant `Time` and `Time_bnds` coordinates, which are also affected.

xylar:
This merge fixes two issues in `timeSeriesStatsDaily` and `timeSeriesStatsMonthly` in MPAS-Ocean.

The first is that the value `xtimeMonthly_start` was inaccurate at the start of a simulation and after model restart.  This occurred because of a combination of problems.  
1. When E3SM reset the MPAS clock, MPAS alarms are recalibrated so they ring immediately (rather than at the end of the desired time-averaging interval), a behavior different from MPAS standalone runs. 
2. As a result, it turns out to be desirable to immediately check (and reset) alarms when MPAS) starts for the first time or restarts, allowing the time interval to begin afresh and time averaging to start over if appropriate.
3. Because MPAS-O is a step behind when starting a new E3SM simulation, it is desirable to compute time-series stats on startup (but not on restart) so the first day or month is complete (rather than short one time step).
4. `xtimeMonthly_start` was being computed by subtracting a time step from the current time, rather than a compute interval of `timeSeriesStats` after a restart.

This issue was present in E3SM output.
closes #5217

The second issue does not affect E3SM output because E3SM doesn't use restart files for `timeSeriesStats`.  However, standalone runs that do use restart files for this analysis were experiencing a problem in which time averaging was not getting reset after a restart, leading to time averages over the whole simulation rather than the expected (daily or monthly) interval.  This issue has been fixed by referencing the reset alarm to the `simulationStartTime` rather than the beginning of the current (possibly restart) run.

closes https://github.com/MPAS-Dev/MPAS-Model/issues/121

### DIFF
--- a/components/mpas-ocean/bld/build-namelist
+++ b/components/mpas-ocean/bld/build-namelist
@@ -1429,7 +1429,11 @@ add_default($nl, 'config_AM_regionalStatsCustom_vertical_dimension');
 ###########################################
 
 add_default($nl, 'config_AM_timeSeriesStatsDaily_enable');
-add_default($nl, 'config_AM_timeSeriesStatsDaily_compute_on_startup');
+if ($CONTINUE_RUN eq 'TRUE') {
+	add_default($nl, 'config_AM_timeSeriesStatsDaily_compute_on_startup', 'val'=>".false.");
+} else {
+	add_default($nl, 'config_AM_timeSeriesStatsDaily_compute_on_startup', 'val'=>".true.");
+}
 add_default($nl, 'config_AM_timeSeriesStatsDaily_write_on_startup');
 add_default($nl, 'config_AM_timeSeriesStatsDaily_compute_interval');
 add_default($nl, 'config_AM_timeSeriesStatsDaily_output_stream');
@@ -1446,7 +1450,11 @@ add_default($nl, 'config_AM_timeSeriesStatsDaily_backward_output_offset');
 #############################################
 
 add_default($nl, 'config_AM_timeSeriesStatsMonthly_enable');
-add_default($nl, 'config_AM_timeSeriesStatsMonthly_compute_on_startup');
+if ($CONTINUE_RUN eq 'TRUE') {
+	add_default($nl, 'config_AM_timeSeriesStatsMonthly_compute_on_startup', 'val'=>".false.");
+} else {
+	add_default($nl, 'config_AM_timeSeriesStatsMonthly_compute_on_startup', 'val'=>".true.");
+}
 add_default($nl, 'config_AM_timeSeriesStatsMonthly_write_on_startup');
 add_default($nl, 'config_AM_timeSeriesStatsMonthly_compute_interval');
 add_default($nl, 'config_AM_timeSeriesStatsMonthly_output_stream');
@@ -1463,7 +1471,11 @@ add_default($nl, 'config_AM_timeSeriesStatsMonthly_backward_output_offset');
 #################################################
 
 add_default($nl, 'config_AM_timeSeriesStatsClimatology_enable');
-add_default($nl, 'config_AM_timeSeriesStatsClimatology_compute_on_startup');
+if ($CONTINUE_RUN eq 'TRUE') {
+	add_default($nl, 'config_AM_timeSeriesStatsClimatology_compute_on_startup', 'val'=>".false.");
+} else {
+	add_default($nl, 'config_AM_timeSeriesStatsClimatology_compute_on_startup', 'val'=>".true.");
+}
 add_default($nl, 'config_AM_timeSeriesStatsClimatology_write_on_startup');
 add_default($nl, 'config_AM_timeSeriesStatsClimatology_compute_interval');
 add_default($nl, 'config_AM_timeSeriesStatsClimatology_output_stream');
@@ -1480,7 +1492,11 @@ add_default($nl, 'config_AM_timeSeriesStatsClimatology_backward_output_offset');
 ################################################
 
 add_default($nl, 'config_AM_timeSeriesStatsMonthlyMax_enable');
-add_default($nl, 'config_AM_timeSeriesStatsMonthlyMax_compute_on_startup');
+if ($CONTINUE_RUN eq 'TRUE') {
+	add_default($nl, 'config_AM_timeSeriesStatsMonthlyMax_compute_on_startup', 'val'=>".false.");
+} else {
+	add_default($nl, 'config_AM_timeSeriesStatsMonthlyMax_compute_on_startup', 'val'=>".true.");
+}
 add_default($nl, 'config_AM_timeSeriesStatsMonthlyMax_write_on_startup');
 add_default($nl, 'config_AM_timeSeriesStatsMonthlyMax_compute_interval');
 add_default($nl, 'config_AM_timeSeriesStatsMonthlyMax_output_stream');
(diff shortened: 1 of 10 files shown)


xylar:
This PR is based off of #5202 and will need to be rebased after that goes in.

xylar:
@jonbob, would you be willing to run a 1-year E3SM simulation with this branch that you can compare with something you already have handy?  That should be plenty to verify that ``xtimeDaily_start` and xtimeMonthly_start` is fixed and that all other variables are unchanged.  This can wait until #5202 goes in or a test run with #5202 could be the baseline, whichever is more efficient for you.

xylar:
@mark-petersen, I'd like this fix to go in before the code freeze.  Any chance you could review it soon?

jonbob:
@xylar -- I'll be happy to test, doing whatever runs you think are necessary. I'll prioritize getting #5202 in as well.

mark-petersen:
Thanks, I'm reviewing now. MPAS-Seaice has the identical timeSeriesStats. Does it need the identical fix?

xylar:
Yes, eventually, MPAS-Seaice will need this and also the equivalent of #5202.  It seems hard to get both of those in and tested before the code freeze, though.

xylar:
I tried an E3SM test (`ERS_D.ne11_oQU240.WCYCL1850NS.anvil_intel`) and it fails.  Let me figure that out before you do anymore review or testing.

mark-petersen (approved):
Tested with gnu debug, gnu optimized, and intel optimized using all daily and monthly timeSeriesStats tests and https://github.com/MPAS-Dev/compass/pull/435. This PR passes all. Based on @xylar's comments, it looks like this fixes the problem described.

xylar:
Update: `ERS_D.ne11_oQU240.WCYCL1850NS.anvil_intel` now runs with the fix I pushed. I'm testing `ERS.ne11_oQU240.WCYCL1850NS.anvil_intel` against a baseline from #5202 and will report back as soon as that finishes.

xylar:
`ERS.ne11_oQU240.WCYCL1850NS.anvil_intel` passes and is BFB with #5202.  @jonbob, this is ready for you to review but maybe fine to wait until #5202 goes in.

jonbob:
note: running this PR in E3SM did not initially achieve the expected results

xylar:
Yes, I'm working on debugging this.  It is proving quite challenging.

xylar:
@mark-petersen and @jonbob, I have updated the PR description and requested a re-review.

components/mpas-ocean/bld/build-namelist
```diff
@@ -1446,7 +1450,11 @@ add_default($nl, 'config_AM_timeSeriesStatsDaily_backward_output_offset');
 #############################################
 
 add_default($nl, 'config_AM_timeSeriesStatsMonthly_enable');
-add_default($nl, 'config_AM_timeSeriesStatsMonthly_compute_on_startup');
+if ($CONTINUE_RUN eq 'TRUE') {
+	add_default($nl, 'config_AM_timeSeriesStatsMonthly_compute_on_startup', 'val'=>".false.");
+} else {
+	add_default($nl, 'config_AM_timeSeriesStatsMonthly_compute_on_startup', 'val'=>".true.");
+}
```
xylar, line 1457: This and similar changes above and below for other time-series stats seems to be necessary so that the first day or month in an E3SM simulation is complete.  Let me know if you have concerns.

components/mpas-ocean/src/analysis_members/Registry_time_series_stats_climatology.xml
```diff
@@ -8,7 +8,7 @@
 		<nml_option name="config_AM_timeSeriesStatsClimatology_compute_on_startup"
 			type="logical"
 			default_value=".false."
-			description="Logical flag determining if an analysis member computation occurs on start-up. You likely want this off for this (time series) analysis member because it will accumulate any state prior to time stepping (double counting the last time step)."
+			description="Logical flag determining if an analysis member computation occurs on start-up. This should be set to .false. for better efficiency in most cases because the results of this first call will simply be thrown away.  The exception is at the beginning of new E3SM simulations (CONTINUE_RUN = FALSE), where the first time step will be missing if this is not set to .true."
```
xylar, line 11: This and related descriptions for other time-series stats have been updated for better accuracy and clarity.  @jonbob, I didn't try to update the associate files in `bld` because I don't remember the exact process.  I can look that up, though, if you would like me to take care of it.

xylar:
## Re-testing

I again ran the 4 new tests introduced in https://github.com/MPAS-Dev/compass/pull/435 to show that these fixes work as expected in standalone MPAS-Ocean.

I also ran 3 months of the E3SM test case `CMPASO-JRA1p4_TL319_EC30to60E2r2` on Chrysalis, with 2 restarts in between.  I see (as expected) the following `Time` and `Time_bnds` values:

```
$ ncdump -v Time,Time_bnds 20221021.CMPASO-JRA1p4_TL319_EC30to60E2r2.fix-time-series-stats.chrysalis.mpaso.hist.am.timeSeriesStatsMonthly.0001-01-01.nc | tail

 Time = 15.5 ;

 Time_bnds =
  0, 31 ;

$ ncdump -v Time,Time_bnds 20221021.CMPASO-JRA1p4_TL319_EC30to60E2r2.fix-time-series-stats.chrysalis.mpaso.hist.am.timeSeriesStatsMonthly.0001-02-01.nc | tail

 Time = 45 ;

 Time_bnds =
  31, 59 ;
}

$ ncdump -v Time,Time_bnds 20221021.CMPASO-JRA1p4_TL319_EC30to60E2r2.fix-time-series-stats.chrysalis.mpaso.hist.am.timeSeriesStatsMonthly.0001-03-01.nc | tail

 Time = 74.5 ;

 Time_bnds =
  59, 90 ;
}
```
See results in:
```
/lcrc/group/e3sm/ac.xylar/scratch/chrys/20221021.CMPASO-JRA1p4_TL319_EC30to60E2r2.fix-time-series-stats.chrysalis/run
```

xylar:
Getting there but testing in MPAS-Seaice is still showing problems that I want to understand before merging this.

rljacob:
telecon notes:  waiting for @xylar to approve.

xylar:
Still debugging...

xylar:
I looked into this further today.  Similar fixes in MPAS-Seaice revealed a problem with this approach.  I have an alternative in mind that I will try to implement next week.  If this PR being open is a problem, we can close it for now and re-open later.

rljacob:
notes:  had to be redone.  Still in progress.

xylar:
I can close this for now as well.

PR #5223, opened 2022-10-06
Hommexx/SL: Implement HOMMEXX_MPI_ON_DEVICE=Off in SL transport.

ambrad:
Previously, the unstructured SL-internal communication rounds were done using device pointers regardless of the value of HOMMEXX_MPI_ON_DEVICE. Now they are done on host if HOMMEXX_MPI_ON_DEVICE is true.

[BFB]

### DIFF
--- a/components/homme/src/share/compose/CMakeLists.txt
+++ b/components/homme/src/share/compose/CMakeLists.txt
@@ -10,6 +10,10 @@ endif ()
 if (COMPOSE_PORT)
   add_definitions(-DCOMPOSE_PORT)
 endif ()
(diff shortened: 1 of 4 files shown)


ambrad:
SCREAM performance on Summit is reported here: https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/3537502297/Summit+ne1024+dycore+performance

ambrad:
Chrysalis e3sm_developer and homme_integration pass.

mt5555 (approved):

PR #5208, opened 2022-09-28
Add -hipa0 -hzero to prevent NaNs with Cray compiler on Crusher

amametjanov:
Update Cray compiler flags and Crusher machine config:
- add `-hipa0 -hzero` to prevent NaNs
- add `-em -ef` for lowercase modulename.mod files
- increase l%360x720cru PE-layout from 1 to 2 nodes
- increase memleak tolerance from 20 to 25%

[BFB]

Issue #5053, opened 2022-06-28
SHR_REPROSUM_CALC errors on Crusher


grnydawn:
A case with F2010 and ne30pg2_r05_oECv3 on Crusher failed with following error message:

> SHR_REPROSUM_CALC: Input contains  0.30000E+01 NaNs and  0.00000E+00 INFs on process       3

This error occurs on E3SM builds using each of Cray, AMD, and GNU compilers available on Crusher.

Interestingly, the F2010/ne30pg2_r05_oECv3 case runs successfully on Summit system where input data are shared with Crusher.

The error message seems to come from "E3SM/share/util/shr_reprosum_mod.F90 but it was hard to locate a source code line that this routine is called from.

Does anyone have seen this error? 

The branch that I am working on is ykim/crusher/craydebug, a debug branch branched off from a recent master branch.

sarats:
Just noting that we should try this case with GNU on a x86 machine and see if we encounter this issue.

grnydawn:
This error is not showing with PrgEnv-cray/8.3.3, PrgEnv-amd/8.3.3, and PrgEnv-gnu/8.3.3. It is not evident if these modules fixed this error or not. It may be possible that a E3SM compiler configuration may cause this error. Because we have no issue  now, I will close it and re-open if needed.

Issue #5051, opened 2022-06-28
Crash while reading input files when using MPICH_GPU_SUPPORT_ENABLED=1 on Crusher


xyuan:
The E3SM-MMF F case runs on Crusher occasionally crash with segmentation error(at every 3 to 5 runs) when reading the input NetCDF files using pNetCDF with MPICH_GPU_SUPPORT_ENABLED=1. However the runs do not crash when MPICH_GPU_SUPPORT_ENABLED is not set (MPICH_GPU_SUPPORT_ENABLED=0). See the detailed information below,

```
  0:  getMetaSchedule: tmpP:  13,  1,  12,  24,  11,  35,  146,  36,  1,  70,  147,  350,  3,  351,  2,  362,  113,  429
  0:  initializing elements...
  0: Create Pool
  0:
srun: error: crusher010: task 32: Segmentation fault (core dumped)
srun: launch/slurm: _step_signal: Terminating StepId=131056.0
  0: slurmstepd: error: *** STEP 131056.0 ON crusher005 CANCELLED AT 2022-06-20T16:05:47 ***
srun: error: crusher127: tasks 240-255: Terminated
srun: error: crusher008: tasks 16-31: Terminated
srun: error: crusher013: tasks 64-79: Terminated
srun: error: crusher016: tasks 96-111: Terminated
srun: error: crusher121: tasks 224-239: Terminated
srun: error: crusher017: tasks 112-127: Terminated
srun: error: crusher088: tasks 144-159: Terminated
srun: error: crusher005: tasks 0-15: Terminated
srun: error: crusher084: tasks 128-143: Terminated
srun: error: crusher102: tasks 192-207: Terminated
srun: error: crusher011: tasks 48-63: Terminated
srun: error: crusher091: tasks 160-175: Terminated
srun: error: crusher118: tasks 208-223: Terminated
srun: error: crusher098: tasks 176-191: Terminated
srun: error: crusher015: tasks 80-95: Terminated
srun: error: crusher010: tasks 33-47: Terminated
srun: Force Terminated StepId=131056.0
```

**back trace information from rockgdb gives the following call stack,**
```
#0  0x00007f9f3cba97ee in __cray_memcpy_ROME () from /opt/cray/pe/cce/14.0.0/cce/x86_64/lib/libu.so.1
[...]

PR #5229, opened 2022-10-11
Add GPU version of compilers on Crusher

amametjanov:
Add GPU version of compilers on Crusher. Also,
- clean up compiler cmake files
- update Kokkos submodule from 3.4.1 to 3.7.01
- default to cyclic (instead of block) MPI task distribution on a node

[non-BFB] - due to `cce/14.0.0 -> cce/14.0.2` (new default version) compiler update


grnydawn (approved):
@amametjanov and @sarats,

I ran e3sm_developer test suites on Crusher using Cray, AMD, and GNU compilers.

Overall, I think this PR is ready to merge to master but I ONLY have tested CPU test cases.
@amametjanov , If you have GPU tests, can you share some instructions to do the tests? I will try to run the tests if possible.

==== test results ====

Cray CCE/14.0.2 : 26 PASSes and 13 FAILs
* This is an improvement from previous e3sm_developer test results
* Many of failed cases have the same issues such as "lib-4412 : UNRECOVERABLE library error"
* Some of failed test cases have "unknown" issue
* Overall, there is no objection on merging from Cray compiler prospective

AMD amd/5.1.0 : 3 PASSes and 5 FAILs
* There is no previous test results to compare
* 2 failed cases from memory deallocation issues, 1 from compiler internal issue during compilation. 2 failed cases have "no run" issue even though build phase was passed
* Overall, there is no objection on merging from AMD compiler prospective

GNU gcc/11.2.0 50 PASSes and 11 FAILs
* 10 failed cases have "no run" issue even though build phase was passed
* 1 failed case have an issue at "COMPARE" phase.
* Overall, there is no objection on merging from GNU compiler prospective

rljacob:
@jgfouca you may need this merged to scream for its testing on crusher.

oksanaguba:
I see some discrepancies between scream setup for clang gpu and this setup. If e3sm cannot run on gpu (technically, hommexx can be used on gpu, but it is not practical), then why have gpu setup here. This causes conflicts with scream during upstream merges.

oksanaguba:
The description of the PR is very short, cpu files also have changed.

oksanaguba:
I think this should not get merged till we sync these changes with scream, so that the next upstream merge does not break scream crusher runs. This is not only to avoid resolution of conflicts in scream after the upstream merge, but also to have a working setup in e3sm (though i am confused why it is needed, maybe, for MMF?). I would suggest to merge this branch to scream first and resolve issues for crusher for clang CPU and GPU (we do not have setups for other compilers for crusher), and check summit GNU runs. I could volunteer.

If MMF needs that, they should do the same.

oksanaguba:
I saw Rob's comment about GPU usage above, thanks.

oksanaguba (changes requested):
Clang crusher and Summit GNU setups need to be tested in scream first. I will do that.

cime_config/machines/cmake_macros/crayclang_crusher.cmake
```diff
@@ -1,9 +1,3 @@
-if (compile_threaded)
```
oksanaguba, line 1: why is this block removed for cpu build, but below, for gpu build, it is added?

cime_config/machines/config_machines.xml
```diff
@@ -595,6 +595,9 @@
     <SUPPORTED_BY>e3sm</SUPPORTED_BY>
     <MAX_TASKS_PER_NODE>64</MAX_TASKS_PER_NODE>
     <MAX_MPITASKS_PER_NODE>64</MAX_MPITASKS_PER_NODE>
+    <MAX_MPITASKS_PER_NODE compiler="gnugpu">16</MAX_MPITASKS_PER_NODE>
```
oksanaguba, line 598: There are 8 GCDs on crusher, why is 16 used below?

cime_config/machines/config_machines.xml
```diff
@@ -648,10 +666,30 @@
...
       <env name="NETCDF_PATH">$ENV{NETCDF_DIR}</env>
       <env name="PNETCDF_PATH">$ENV{PNETCDF_DIR}</env>
+      <env name="NTASKS_PER_GPU"> </env>
+      <env name="GPU_BIND_ARGS"> </env>
     </environment_variables>
     <environment_variables compiler="amdclang">
       <env name="LD_LIBRARY_PATH">$ENV{CRAY_LIBSCI_DIR}/amd/4.0/x86_64/lib:$ENV{LD_LIBRARY_PATH}</env>
     </environment_variables>
+    <environment_variables compiler=".*gpu.*">
+      <env name="NTASKS_PER_GPU">--ntasks-per-gpu=2</env>
```
oksanaguba, line 676: Why is 2 used here?

oksanaguba:
In this branch, did any configuration run with crayclanggpu? Asking, because the scream version is rather different, not sure whether all changes in this branch are needed for crusher gpu?

For example, scream uses
```
SET(CMAKE_C_COMPILER "cc" CACHE STRING "")
SET(CMAKE_Fortran_COMPILER "ftn" CACHE STRING "")
SET(CMAKE_CXX_COMPILER "hipcc" CACHE STRING "")
```
and this branch has
```
set(MPICC "cc")
set(MPICXX "CC")
set(MPIFC "ftn")
set(SCC "cc")
set(SCXX "CC")
set(SFC "ftn")
```

rljacob:
@amametjanov please respond to comments.

cime_config/machines/cmake_macros/crayclang_crusher.cmake
amametjanov: `-fopenmp` gets added in the higher-level `crayclang.cmake`.
Ideally, `crayclang_crusher.cmake` should contain Crusher-specific flags (and avoid duplicates).

I didn't see a prior `crayclanggpu.cmake` (only Crusher had `crayclanggpu` at the time) and just added a single `crayclanggpu_crusher.cmake` that had both generic and machine-specific flags.
I can rebase onto the latest master and update the branch.

cime_config/machines/config_machines.xml
amametjanov: IIRC, that's what was best in MMF runs.

cime_config/machines/config_machines.xml
amametjanov: For MMF runs on GPUs.

amametjanov (commented):
I opted for Cray's `CC`, instead of AMD's `hipcc` (for `crayclang[gpu]`), to have fewer MPI link flags.
AFAICT, `CC` does not preclude GPU-aware MPI.

oksanaguba:
@amametjanov  would you please remind me, if there are 3 files, crayclang (file 1), crayclang_crusher (file2), and crayclanggpu_crusher (file3), are they all loaded for a gpu setup? or only file 3, or any other way?

the same is about cpu run, which file(s) are loaded? 

thanks!

oksanaguba:
@amametjanov  could you please look at https://github.com/E3SM-Project/scream/pull/2129 , the last commit, where i tried to place comments regarding all options in your files and what's used right now in scream. 

right now, i am only interested in the working config for scream on crusher, which is crayclang and cpu/gpu. i started with a gpu file. scream uses crayclang_crusher-gpu in scream repo. i do not remember whether all *crayclang* cmake files are used or only the most specific one (would you know?), so, some of my comments might not make sense, but even if i compare your crayclanggpu_crusher and scream's crayclang_crusher-gpu, there are many differences. for example, in your file there a few flags that scream did not set. primitively, i would prefer as few flags as possible in these files.

amametjanov:
I think this is the sequence:
- for CPU runs: crayclang.cmake and crayclang_crusher.cmake .
- for GPU runs: crayclanggpu_crusher.cmake .

I tried to set as few flags as possible to build and run `SMS_Ln9_P16x1.ne4pg2_ne4pg2.F2010-MMF1.crusher_crayclanggpu` .

oksanaguba:
Here is Jim's notes on this:
```
CMake macros are loaded in this order. If the macro doesn't exist, that level is skipped. Users will usually be working with the $compiler_$mach.cmake macro.

universal.cmake : applies changes globally
$compiler.cmake : applies changes for a compiler type (on all machines)
$os.cmake : applies change for an OS
$mach.cmake : applies changes for a machine (for all compilers)
$compiler_$os.cmake : applies changes for a compiler/os combo (for all machines)
$compiler_$mach.cmake : applies changes to a compiler/mach comb
```

cime_config/machines/config_machines.xml
```diff
@@ -651,10 +669,35 @@
...
+      <env name="GPU_BIND_ARGS">--gpu-bind=closest</env>
+    </environment_variables>
+    <environment_variables compiler=".*gpu.*" DEBUG="TRUE">
+      <env name="AMD_LOG_LEVEL">10</env>
+      <env name="CRAY_ACC_DEBUG">3</env>
+    </environment_variables>
+    <environment_variables compiler="gnugpu">
+      <env name="PNETCDF_HINTS">romio_cb_read=disable</env>
```
oksanaguba, line 687: there is 
```
set(PIO_FILESYSTEM_HINTS "romio_cb_read=disable")
```
above too, are both changes needed?

cime_config/machines/cmake_macros/crayclanggpu_crusher.cmake
```diff
@@ -0,0 +1,52 @@
...
+if (NOT DEBUG)
+  string(APPEND CFLAGS   " -O2")
+  string(APPEND CXXFLAGS " -O2")
+  string(APPEND FFLAGS   " -O2")
+endif()
+
+if (COMP_NAME STREQUAL elm)
+  string(APPEND FFLAGS " -hfp0")
+endif()
+string(APPEND FFLAGS " -hipa0 -hzero -em -ef -hnoacc")
```
oksanaguba, line 40: Please, remove -hzero as per comment https://github.com/E3SM-Project/scream/pull/2129/files#r1072914312 .

oksanaguba:
i am still sorting out these changes in a scream branch, will let you know as soon as i am done testing them.

cime_config/machines/cmake_macros/crayclanggpu_crusher.cmake
sarats: Hmm...I'm not sure what comment you were referring to there.

Anyway, -hzero was added to resolve a NaN issue (https://github.com/E3SM-Project/E3SM/issues/5053) in the PR  https://github.com/E3SM-Project/E3SM/pull/5208. 
So, are you sure it's not needed?

sarats: In fact, I wonder if -hipa0 should be more limited in scope (in Depends file) as that would stop all interprocedural analysis and optimization including inlining of small functions.

oksanaguba: Andrew left a comment in scream PR (prob was erased by more commits):
```
Note we don't want to use -hzero, as that zeros all stack-allocated arrays. This can lead to uncaught uninitialized-variable bugs.
```

sarats: Hmm. In one sense, -hzero was trying to address the unitialized variable issue. I think that should be fixed at the source level for all components. Unfortunately, I think there are issues with Land or FATES IIRC. Anyway, removing -hzero can potentially cause NaN's to return in certain test cases. Someone has to verify that it won't impact say coupled B-case MMF runs if we remove it.

sarats:
Curious why does the files changed show unrelated files to Crusher machine/compiler config? Perhaps need to cleanup or rebase.

sarats:
@amametjanov and @grnydawn It's probably good to revisit the -hipa0 flag introduced in 
https://github.com/E3SM-Project/E3SM/pull/5208 to limit scope to specific components/sources.

sarats:
Remind me: Do we have a way to specify NTASKS_PER_GPU when creating a case? 
We would definitely run into situations where MMF wants it one way and Scream another. So this should be user configurable for flexibility.

Btw, upstream CIME's ngpus_per_node is orthogonal to all this as NCAR uses it to handle the non-homogeneity (Beowulf style) of their GPU machines. All the DOE heterogeneous supercomputers have uniform h/w distribution.

xyuan (approved):

cime_config/machines/config_machines.xml
```diff
@@ -651,10 +669,35 @@
...
+      <env name="MPICH_GPU_SUPPORT_ENABLED">1</env>
+      <env name="MPIR_CVAR_GPU_EAGER_DEVICE_MEM">0</env>
+    </environment_variables>
+    <environment_variables compiler="crayclanggpu">
+      <env name="PNETCDF_HINTS">romio_cb_read=disable</env>
+      <env name="MPICH_GPU_SUPPORT_ENABLED">0</env>
```
sarats, line 693: So, we are enabling MPICH_GPU_SUPPORT for amdclanggpu but not for crayclangpu?
I recall Scream has issues with GPU-aware MPI. Is that the reason or something else?

cime_config/machines/config_machines.xml
```diff
@@ -606,8 +609,10 @@
         <!-- <arg name="binding">threads-per-core=1 -c 2</arg> -->
         <arg name="binding">--threads-per-core=1</arg>
         <arg name="thread_count">-c $ENV{OMP_NUM_THREADS}</arg>
-        <arg name="placement">-m *:block</arg>
+        <arg name="distribution">-m *:block</arg>
```
sarats, line 612: As the '*:block' deviates from the default of 'block:cyclic:cyclic', is this binding good for the GPU cases as well? Have you verified how this interacts with --gpu-bind=closest option? 

Ref for self/others: `    --distribution=<value>[:<value>][:<value>] specifies the distribution of MPI ranks across compute nodes, sockets (L3 cache regions on Crusher), and cores, respectively. `

sarats:
@xyuan and @oksanaguba What are you using for Slurm process binding for your GPU runs? Just '--gpu-bind=closest' or ...?

Anyway, please post your 'srun' invocation for reference.

xyuan:
I used –gpu-bind=closest, for mpich_gpu aware option should be off as it is not going to be ready soon as far as my understanding from the ECP frontier office hours.



From: Sarat Sreepathi ***@***.***>
Date: Tuesday, January 24, 2023 at 2:39 AM
Cc: Yuan, Xingqiu ***@***.***>, Mention ***@***.***>
Subject: Re: [E3SM-Project/E3SM] Add GPU version of compilers on Crusher (PR #5229)

@xyuan<https://github.com/xyuan> and @oksanaguba<https://github.com/oksanaguba> What are you using for Slurm process binding for your GPU runs? Just '--gpu-bind=closest' or ...?

Anyway, please post your 'srun' invocation for reference.

oksanaguba:
@sarats i am in a process of bringing these changes to scream and verifying its performance is as expected, so i asked above to not to merge this PR till then. my connection to crusher froze, so i would post the run line later.

sarats:
@grnydawn Looks like the -m *:block was added in an earlier PR when it probably made sense. We should probably stick with the default and remove this now.
https://github.com/E3SM-Project/E3SM/pull/5082

On a similar note, I think we should omit --threads-per-core=1 as that's the default behavior.

cime_config/machines/cmake_macros/crayclang.cmake
```diff
@@ -1,11 +1,14 @@
 if (compile_threaded)
-  string(APPEND CFLAGS " -h omp")
-  string(APPEND FFLAGS " -h omp")
-  string(APPEND LDFLAGS " -h omp")
+  string(APPEND CFLAGS   " -fopenmp")
```
oksanaguba, line 2: i did not try to run scaling studies to see if either flag works with OMP, but from googling I thought -homp is the correct flag fro cray (but that flag throws an error if used with hipcc).

sarats: Why do you want to run scaling studies to check this flag?

In the case of Cray Fortran, both flags are equivalent. Since Cray C compiler is Clang based, the OpenMP flag is `-f openmp`

From the Cray Fortran man page
```
>       -f [no]openmp
              Enable  or  disable  compiler  recognition of OpenMP directives.  Using -fnoopenmp is similar to the -h
              thread0 option, in that it disables OpenMP, but unlike -h  thread0,  -fnoopenmp  does  not  affect  au-
              tothreading.   This  option is an alias to -h[no]omp.  CrayPE will link in the serial version of LibSci
              when -fnoopenmp is used.

              � Default: -f noopenmp [-h noomp]
```

oksanaguba: thanks, Sarat. i ran into an error only with `hipcc -homp`. Re: scaling -- i think it is good to actually confirm that OMP works (and flags, too), the only way I know how to do it is to run a code that scales (homme, for example). But this is out of range of my current tasks, so i just made a note about it.

oksanaguba: so, great to know that -fopenmp should work, too.

sarats: Don't we have a small case which can be used to check MPI vs. MPI+OMP perf?

sarats: As I mentioned in other comments, some updates are needed for things in this PR. So scaling runs as is is not recommended.

sarats:
@amametjanov Are you working on updates to this PR? @grnydawn and I can help as needed.

cime_config/machines/config_machines.xml
amametjanov: `block` placement (adjacent MPI tasks on adjacent cores) has always been better than `cyclic` (round-robin).
Also, all nodes and cores are almost always fully populated in E3SM runs: the only time cyclic/round-robin can run faster is when nodes or cores are under-populated.

amametjanov (commented):
I rebased the branch onto latest master.
Please add new comments as a new review, if I missed something.

sarats:
Regarding process binding, have you seen Example 7: 16 MPI ranks - where 2 ranks share a GPU (packed, single-node) on https://docs.olcf.ornl.gov/systems/crusher_quick_start_guide.html#process-and-thread-mapping

Especially this note:
```
Because the distribution across L3 cache regions has been changed to a “packed” (block) configuration, caution must be taken to ensure MPI ranks end up in the L3 cache regions where the GPUs they intend to be mapped to are located. To accomplish this, the number of physical CPU cores assigned to an MPI rank was increased - in this case to 4. Doing so ensures that only 2 MPI ranks can fit into a single L3 cache region. If the value of -c was left at 1, all 8 MPI ranks would be “packed” into the first L3 region, where the “closest” GPU would be GPU 4 - the only GPU in that L3 region.

Notice that this is not a workaround like in Example 6, but a requirement due to the block distribution of MPI ranks across NUMA domains.
```

It would be good to verify placement using https://code.ornl.gov/olcf/hello_jobstep. In fact, posting the output somewhere on Confluence would serve as a good reference (analogous what we did for Summit).

amametjanov:
To accommodate the largest number of possible scenarios, the distribution of tasks across L3 caches (and GPUs) is now `cyclic`.

Some examples:
1. To get 8 MPI processes per node
```
$ ./xmlchange MAX_MPITASKS_PER_NODE=8
$ ./xmlchange NTHRDS=7
$ srun  -l -K -n 8 -N 1  -c 7 --ntasks-per-gpu=1 --gpu-bind=closest ...

0: MPI 000 - OMP 000 - HWT 001 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 4 - Bus_ID d1
0: MPI 000 - OMP 001 - HWT 002 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 4 - Bus_ID d1
0: MPI 000 - OMP 002 - HWT 003 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 4 - Bus_ID d1
0: MPI 000 - OMP 003 - HWT 004 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 4 - Bus_ID d1
0: MPI 000 - OMP 004 - HWT 005 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 4 - Bus_ID d1
0: MPI 000 - OMP 005 - HWT 006 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 4 - Bus_ID d1
0: MPI 000 - OMP 006 - HWT 007 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 4 - Bus_ID d1

1: MPI 001 - OMP 000 - HWT 009 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 5 - Bus_ID d6
1: MPI 001 - OMP 001 - HWT 010 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 5 - Bus_ID d6
1: MPI 001 - OMP 002 - HWT 011 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 5 - Bus_ID d6
1: MPI 001 - OMP 003 - HWT 012 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 5 - Bus_ID d6
1: MPI 001 - OMP 004 - HWT 013 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 5 - Bus_ID d6
1: MPI 001 - OMP 005 - HWT 014 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 5 - Bus_ID d6
1: MPI 001 - OMP 006 - HWT 015 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 5 - Bus_ID d6

...
7: MPI 007 - OMP 000 - HWT 057 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 1 - Bus_ID c6
7: MPI 007 - OMP 001 - HWT 058 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 1 - Bus_ID c6
7: MPI 007 - OMP 002 - HWT 059 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 1 - Bus_ID c6
7: MPI 007 - OMP 003 - HWT 060 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 1 - Bus_ID c6
7: MPI 007 - OMP 004 - HWT 061 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 1 - Bus_ID c6
7: MPI 007 - OMP 005 - HWT 062 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 1 - Bus_ID c6
7: MPI 007 - OMP 006 - HWT 063 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 1 - Bus_ID c6
```

2. For 16 ppn:
```
$ ./xmlchange MAX_MPITASKS_PER_NODE=16
$ ./xmlchange NTHRDS=3
$ srun  -l -K -n 16 -N 1  -c 3 --ntasks-per-gpu=2 --gpu-bind=closest ...

 0: MPI 000 - OMP 000 - HWT 001 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 4 - Bus_ID d1
 0: MPI 000 - OMP 001 - HWT 002 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 4 - Bus_ID d1
 0: MPI 000 - OMP 002 - HWT 003 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 4 - Bus_ID d1
 1: MPI 001 - OMP 000 - HWT 009 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 5 - Bus_ID d6
 1: MPI 001 - OMP 001 - HWT 010 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 5 - Bus_ID d6
 1: MPI 001 - OMP 002 - HWT 011 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 5 - Bus_ID d6
...
 8: MPI 008 - OMP 000 - HWT 004 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 4 - Bus_ID d1
 8: MPI 008 - OMP 001 - HWT 005 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 4 - Bus_ID d1
 8: MPI 008 - OMP 002 - HWT 006 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 4 - Bus_ID d1
...
15: MPI 015 - OMP 000 - HWT 060 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 1 - Bus_ID c6
15: MPI 015 - OMP 001 - HWT 061 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 1 - Bus_ID c6
15: MPI 015 - OMP 002 - HWT 062 - Node crusher111 - RT_GPU_ID 0 - GPU_ID 1 - Bus_ID c6
```

3. CPU-only, 56 ppn:
```
$ ./xmlchange MAX_MPITASKS_PER_NODE=56
$ ./xmlchange NTHRDS=1
$ srun  -l -K -n 56 -N 1  -c 1 ...

 0: MPI 000 - OMP 000 - HWT 001 - Node crusher111 - RT_GPU_ID 0,1,2,3,4,5,6,7 - GPU_ID N/A - Bus_ID c1,c6,c9,ce,d1,d6,d9,de
 1: MPI 001 - OMP 000 - HWT 009 - Node crusher111 - RT_GPU_ID 0,1,2,3,4,5,6,7 - GPU_ID N/A - Bus_ID c1,c6,c9,ce,d1,d6,d9,de
 2: MPI 002 - OMP 000 - HWT 017 - Node crusher111 - RT_GPU_ID 0,1,2,3,4,5,6,7 - GPU_ID N/A - Bus_ID c1,c6,c9,ce,d1,d6,d9,de
 3: MPI 003 - OMP 000 - HWT 025 - Node crusher111 - RT_GPU_ID 0,1,2,3,4,5,6,7 - GPU_ID N/A - Bus_ID c1,c6,c9,ce,d1,d6,d9,de
 4: MPI 004 - OMP 000 - HWT 033 - Node crusher111 - RT_GPU_ID 0,1,2,3,4,5,6,7 - GPU_ID N/A - Bus_ID c1,c6,c9,ce,d1,d6,d9,de
 5: MPI 005 - OMP 000 - HWT 041 - Node crusher111 - RT_GPU_ID 0,1,2,3,4,5,6,7 - GPU_ID N/A - Bus_ID c1,c6,c9,ce,d1,d6,d9,de
 6: MPI 006 - OMP 000 - HWT 049 - Node crusher111 - RT_GPU_ID 0,1,2,3,4,5,6,7 - GPU_ID N/A - Bus_ID c1,c6,c9,ce,d1,d6,d9,de
 7: MPI 007 - OMP 000 - HWT 057 - Node crusher111 - RT_GPU_ID 0,1,2,3,4,5,6,7 - GPU_ID N/A - Bus_ID c1,c6,c9,ce,d1,d6,d9,de
...
48: MPI 048 - OMP 000 - HWT 007 - Node crusher111 - RT_GPU_ID 0,1,2,3,4,5,6,7 - GPU_ID N/A - Bus_ID c1,c6,c9,ce,d1,d6,d9,de
49: MPI 049 - OMP 000 - HWT 015 - Node crusher111 - RT_GPU_ID 0,1,2,3,4,5,6,7 - GPU_ID N/A - Bus_ID c1,c6,c9,ce,d1,d6,d9,de
50: MPI 050 - OMP 000 - HWT 023 - Node crusher111 - RT_GPU_ID 0,1,2,3,4,5,6,7 - GPU_ID N/A - Bus_ID c1,c6,c9,ce,d1,d6,d9,de
51: MPI 051 - OMP 000 - HWT 031 - Node crusher111 - RT_GPU_ID 0,1,2,3,4,5,6,7 - GPU_ID N/A - Bus_ID c1,c6,c9,ce,d1,d6,d9,de
52: MPI 052 - OMP 000 - HWT 039 - Node crusher111 - RT_GPU_ID 0,1,2,3,4,5,6,7 - GPU_ID N/A - Bus_ID c1,c6,c9,ce,d1,d6,d9,de
53: MPI 053 - OMP 000 - HWT 047 - Node crusher111 - RT_GPU_ID 0,1,2,3,4,5,6,7 - GPU_ID N/A - Bus_ID c1,c6,c9,ce,d1,d6,d9,de
54: MPI 054 - OMP 000 - HWT 055 - Node crusher111 - RT_GPU_ID 0,1,2,3,4,5,6,7 - GPU_ID N/A - Bus_ID c1,c6,c9,ce,d1,d6,d9,de
55: MPI 055 - OMP 000 - HWT 063 - Node crusher111 - RT_GPU_ID 0,1,2,3,4,5,6,7 - GPU_ID N/A - Bus_ID c1,c6,c9,ce,d1,d6,d9,de
```

grnydawn:
@amametjanov , @sarats @oksanaguba , I revisited the usage of "-hipa0" and "-hzero" flags by running the test cases that Az reported in the PR #5208 using the latest compilers and codes on this branch.

The test results show that both flags are still effective as shown below. NOTE: I excluded 5 test cases that were failed regardless of using any of the two flags.

"-hipa0" and "-hzero": PASS 9, FAIL 0
"-hipa0"                       : PASS 0, FAIL 9
"-hzero"                       : PASS 4, FAIL 5
none of the two flags  : PASS 0, FAIL 9

grnydawn:
@oksanaguba, did you finish reviewing this PR?

grnydawn (approved):

oksanaguba:
@grnydawn yes, i did, thanks.

oksanaguba (approved):

sarats:
@grnydawn I'm trying to see if we can limit the scope of -hipa0. 

I guess first we need to check if this has a performance impact for MMF and SCREAM cases.

From previous tests (https://github.com/E3SM-Project/E3SM/pull/5208#issuecomment-1260316988), the test failures seem scattered. Did you see similar behavior?

If difficult to narrow down, I'm okay with -hzero globally. Folks can enable that during development to check for uninitialized variables.

One more thing, I don't think we need `-h byteswapio` flag. This needs to be checked. It might be a legacy from a older machine with mixed architecture. We are not mixing big Endian and little Endian I/O if I understand correctly.

cime_config/machines/config_machines.xml
```diff
@@ -649,10 +667,22 @@
...
+      <env name="NTASKS_PER_GPU"> </env>
+      <env name="GPU_BIND_ARGS"> </env>
     </environment_variables>
     <environment_variables compiler="amdclang">
       <env name="LD_LIBRARY_PATH">$ENV{CRAY_LIBSCI_DIR}/amd/4.0/x86_64/lib:$ENV{LD_LIBRARY_PATH}</env>
     </environment_variables>
+    <environment_variables compiler=".*gpu.*">
+      <env name="NTASKS_PER_GPU">--ntasks-per-gpu=$SHELL{echo "`./xmlquery --value MAX_MPITASKS_PER_NODE`/8"|bc}</env>
+      <env name="GPU_BIND_ARGS">--gpu-bind=closest</env>
+      <env name="PNETCDF_HINTS">romio_cb_read=disable</env>
```
sarats, line 679: Is there an issue describing what this romio flag handles?

sarats (approved):
Besides checking necessity of 'byteswapio' flag, this is good to go.

I think we can investigate if -hipa0 can be reduced in scope and make changes afterwards.

sarats:
@xyuan Can you clarify if you see a performance improvement with 16 MPI tasks per node vs 8 on Crusher? It wasn't clear to me based on the discussion in the last meeting.

The default is currently being set with MMF in mind and I want to make sure it is actually helpful. Do you see benefit in F-cases or B-cases?

cime_config/machines/config_machines.xml
sarats: @jayeshkrishna See above: why is `romio_cb_read=disable` needed? Btw, have you been doing any Scopio tuning on Crusher?

grnydawn:
> 

@sarats , Yes, I saw exactly the same issue per each test cases from #5208.

sarats:
Evidently, `romio_cb_read=disable` turns off collective buffering (aka two-phase I/O). Historically, this has been very beneficial on Lustre filesystems and relatively less so on GPFS. 

As this hint is only passed for PnetCDF, is it one of the attempts to resolve https://github.com/E3SM-Project/scream/issues/2080?

oksanaguba:
For Crusher, Xingqiu suggested that romio argument related to https://github.com/E3SM-Project/E3SM/issues/5051 .

sarats:
> 16



> @xyuan Can you clarify if you see a performance improvement with 16 MPI tasks per node vs 8 on Crusher? It wasn't clear to me based on the discussion in the last meeting.
> 
> The default is currently being set with MMF in mind and I want to make sure it is actually helpful. Do you see benefit in F-cases or B-cases?

@xyuan Reminder

cime_config/machines/config_machines.xml
```diff
@@ -597,15 +597,21 @@
     <BATCH_SYSTEM>slurm</BATCH_SYSTEM>
     <SUPPORTED_BY>e3sm</SUPPORTED_BY>
     <MAX_TASKS_PER_NODE>56</MAX_TASKS_PER_NODE>
+    <MAX_TASKS_PER_NODE compiler="gnugpu">16</MAX_TASKS_PER_NODE>
+    <MAX_TASKS_PER_NODE compiler="crayclanggpu">16</MAX_TASKS_PER_NODE>
+    <MAX_TASKS_PER_NODE compiler="amdclanggpu">16</MAX_TASKS_PER_NODE>
```
sarats, line 602: We would need to accommodate cases to potentially use all available CPU cores for threading. So, we should let these go up to 56 (low-noise mode) default.

sarats:
@amametjanov Trying to understand the version of Kokkos you are updating the module. It appears to be older than the one SCREAM uses. Of course, SCREAM uses the module via EKAT which is still not in the E3SM repo. (Related: https://github.com/E3SM-Project/E3SM/issues/5417)

While we are updating Kokkos, I wonder if we should just update to the one used by SCREAM? 

The EAMxx recent PR and current EKAT master both have:
https://github.com/E3SM-Project/EKAT/tree/bd74445cd12593605ed087f3334774145b685d7c/extern
[kokkos @ e8ad274](https://github.com/E3SM-Project/kokkos/tree/e8ad274447f11582cef98f0aa409d8ee06d2c7fa)

amametjanov:
Current E3SM master points to Kokkos from 2021-May-25, v3.4.1: this version doesn't have threading flags for IBM and PGI compilers (for Summit/Ascent) and doesn't have device-specific cmake for AMD MI250X on Crusher.

SCREAM also points to this Kokkos version 3.4.1, hash 290a073 .
EKAT points to Kokkos v3.7.00 from 2022-Sep-20, hash e8ad274 .
This PR is updating the pointer to release v3.7.01 from 2022-Dec-6, which was the latest release at the time this PR was submitted.

I checked with Dec-6 version. I'm not sure if Sep-20 version has the fixes.

sarats:
Ok, that sounds good. The Kokkos under `externals` is only used by limited configurations anyway.

SCREAM uses the one from EKAT and this is a newer version than that. So, let's go with this.

ambrad:
Looks like this is causing a build error in homme_integration. I can fix this. Would you like me to push a commit to this branch to resolve the error, and then you can re-merge this PR to next?

grnydawn:
@amametjanov, I locally merged this PR on top of the latest master branch. After running CPRNC, I got 30 tests that show non-BFB between before and after the merging. I assume that non-BFB is not something you expected in this PR. Can you take a look at this?

[non-bfb.txt](https://github.com/E3SM-Project/E3SM/files/10675717/non-bfb.txt)

ambrad:
There are also two build failures in the Compy PGI nightly that look like a PGI-related issue triggered by the new Kokkos version. Probably some workaround commit needs to be put in the Kokkos submodule. If I'm reading the log correctly, it appears that PGI can't handle this type conversion:
```C++
#include <cstdint>
#include <utility>

int main () {
  std::pair<int64_t,int64_t> hi(0,0);
}
```

amametjanov:
The 2 build fails on Compy PGI v19.10 are for threaded builds of kokkos in
```
ERP_Ln18.ne4_oQU240.F2010.compy_pgi
SMS_P12x2.ne4_oQU240.WCYCL1850NS.compy_pgi.allactive-mach_mods
```
with errors
```
PGCC-S-0094-Illegal type conversion required (/qfs/people/e3smtest/jenkins/workspace/ACME_Compy_next/E3SM/externals/kokkos/core/src/OpenMP/Kokkos_OpenMP_Parallel.hpp: 130)
```
and relevant lines
```
 123   template <class Policy>
 124   std::enable_if_t<!std::is_same<typename Policy::schedule_type::type,
 125                                  Kokkos::Dynamic>::value>
 126   execute_parallel() const {
 127 #pragma omp parallel for schedule(static KOKKOS_OPENMP_OPTIONAL_CHUNK_SIZE) \
 128     num_threads(OpenMP::impl_thread_pool_size())
 129     KOKKOS_PRAGMA_IVDEP_IF_ENABLED
 130     for (auto iwork = m_policy.begin(); iwork < m_policy.end(); ++iwork) {
 131       exec_work(m_functor, iwork);
 132     }
 133   }
```
Compy PGI nightly is being replaced by `pm-cpu`, which has newer `nvidia/22.5` compiler module.
Ascent's newer `nvhpc/21.11` has no problems building those 2 cases.
I disabled threaded Kokkos build with pgi on Compy:
```
if (compile_threaded)
  string(APPEND KOKKOS_OPTIONS " -DKokkos_ENABLE_OPENMP=Off") # work-around for pgi/19.10 as kokkos has bld error with it 
endif()
```
This can be removed from other machines (perlmutter for sure, don't know about chicoma and alvarez)
```
$ grep -ir "Kokkos_ENABLE_OPENMP=off" cime_config/machines/
cime_config/machines/cmake_macros/nvidia_pm-cpu.cmake:  string(APPEND KOKKOS_OPTIONS " -DKokkos_ENABLE_OPENMP=Off") # work-around for nvidia as kokkos is not passing "-mp" for threaded build
cime_config/machines/cmake_macros/nvidia_chicoma-cpu.cmake:  string(APPEND KOKKOS_OPTIONS " -DKokkos_ENABLE_OPENMP=Off") # work-around for nvidia as kokkos is not passing "-mp" for threaded build
cime_config/machines/cmake_macros/nvidia_pm-gpu.cmake:  string(APPEND KOKKOS_OPTIONS " -DKokkos_ENABLE_OPENMP=Off") # work-around for nvidia as kokkos is not passing "-mp" for threaded build
cime_config/machines/cmake_macros/nvidia_alvarez.cmake:  string(APPEND KOKKOS_OPTIONS " -DKokkos_ENABLE_OPENMP=Off") # work-around for nvidia as kokkos is not passing "-mp" for threaded build
```
because instead of Kokkos being too old before, it's now the compiler that's old.

amametjanov:
The PR is non-BFB due to `cce/14.0.0 -> cce/14.0.2` (new default version) compiler update: going back to 14.0.0 makes it BFB.

sarats:
Do we need to update baselines or the differences too large indicated an issue with new version?

amametjanov:
There are no baselines that are maintained/tracked for Crusher.
Since the BFB was brought up during review, I narrowed it down to compiler version change and non-BFB is expected during version updates.
We should keep the new default version `cce/14.0.2`.

cime_config/machines/cmake_macros/gnugpu_crusher.cmake
```diff
@@ -0,0 +1,27 @@
...
+string(APPEND CMAKE_OPTS " -DPIO_ENABLE_TOOLS:BOOL=OFF")
+
+set(MPICC  "cc")
+set(MPICXX "hipcc")
+set(MPIFC  "ftn")
+set(SCC  "cc")
+set(SCXX "hipcc")
+set(SFC  "ftn")
+
+string(APPEND CXXFLAGS " -I$ENV{MPICH_DIR}/include --amdgpu-target=gfx90a")
```
sarats, line 23: I'm probably missing something trivial. This flag is causing issues with PIO build for me as PIO is trying to use this flag with Cray CC. Have you seen anything similar? As nightly tests worked, I wonder what in my environment is causing PIO to pickup CC?

oksanaguba:
@grnydawn please, file a blessing ticket for homme test on chrysalis, thanks https://my.cdash.org/test/73207094?graph=status .

grnydawn:
> 

@oksanaguba , I am not sure what you mean by "file a blessing ticket for homme test on chrysalis". Could you explain a little more about what you want me to do?

oksanaguba:
Please go to https://my.cdash.org/index.php?project=E3SM&date=2023-02-16 and find link next to text "To request a dashboard action..." which will lead you to the system for tickets. The link is https://acme-climate.atlassian.net/servicedesk/customer/portal/2 (but finding it on cdash is how i remember it). Then go to link "Bless DIFFs on Chrysalis Next" and fill in all needed info, like PR number and test to bless (in this case, HOMME_P24.f19_g16_rx1.A.chrysalis_intel  ). Then you will get email when blessing was processed. Make sure it was processed. 

Thanks.

mt5555:
and for even more information, the blessing procedure is explained in the E3SM integrator's guide:

https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/1708589129/Integrator+Guide

grnydawn:
> Please go to https://my.cdash.org/index.php?project=E3SM&date=2023-02-16 and find link next to text "To request a dashboard action..." which will lead you to the system for tickets. The link is https://acme-climate.atlassian.net/servicedesk/customer/portal/2 (but finding it on cdash is how i remember it). Then go to link "Bless DIFFs on Chrysalis Next" and fill in all needed info, like PR number and test to bless (in this case, HOMME_P24.f19_g16_rx1.A.chrysalis_intel ). Then you will get email when blessing was processed. Make sure it was processed.
> 
> Thanks.

I have just done the blessing. Please check if that is what you expected. Also thanks for the details information.

grnydawn:
> and for even more information, the blessing procedure is explained in the E3SM integrator's guide:
> 
> https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/1708589129/Integrator+Guide

@mt5555 , Thanks for the link. That was helpful.

rljacob:
How did this PR change a test on chrysalis?

rljacob:
I think I see it.  @amametjanov why were code changes made in components/homme/test_execs/share_kokkos_ut ?

rljacob:
@ambrad nevermind.  I read the previous comments.

rljacob:
At some point, the PR did more than simply add GNU compiler definitions for crusher.  The title and description should have been updated then.

Issue #5219, opened 2022-10-06
Ocean fails stand-alone decomp test, intel optimized 


mark-petersen:
MPAS-Ocean nightly on master 4dcc8dba fails
```
ocean/baroclinic_channel/10km/decomp_test
ocean/global_ocean/QU240/PHC/RK4/decomp_test
```
with the intel optimized compiler and OpenMP. Differences between 4 processor and 8 processor runs are max 1e-13.

mark-petersen:
Earlier this week, I thought it failed 
```
ocean/global_ocean/QU240/PHC/decomp_test
```
with intel optimized, but I don't see that on badger with 4dcc8dba now - maybe I'm mixing it up though.

mark-petersen:
tested the very first SMEP merge:   bb84429ea7 Merge branch 'vanroekel/ocean/add-submesoscale-eddies' (PR #5099)
```
PASS ocean/baroclinic_channel/10km/decomp_test
FAIL ocean/global_ocean/QU240/PHC/decomp_test
PASS ocean/global_ocean/QU240/PHC/RK4/decomp_test
```
with intel optimized on badger. This is confusing.

xylar:
I'm trying to use bisection to find the cause of this and I'm just seeing hanging when I try to test on c63cce2acb.  This may have more to do with a bad node on Anvil or something but it's certainly not helping me debug...

xylar:
@mark-petersen, I agree that #5099 is responsible and that this is probably already fixed in https://github.com/E3SM-Project/E3SM/pull/5216.  I'll make sure.

xylar:
Nope, the test is still failing after #5216 so there's another threading problem from #5099 that we need to track down.

xylar:
Sorry, I wasn't clear in my mind.  We're looking for a decomposition problem, not a threading problem.  Much trickier in some ways!

xylar:
I believe this was introduced by #5183 and not by #5099.  At least that is what I'm seeing in testing on Chrysalis with Intel and Intel-MPI.  I'm seeing test execution passing for #5170 (the previous ocean-related commit merge) but failing for #5183.

xylar:
After rerunning analysis today, this test case is running fine for #5183 so the failures yesterday seem random and unrelated to this issue.  Still investigating.

xylar:
[...]

PR #5232, opened 2022-10-12
Add zeroing of velocity to ocean init mode

### LINKED ISSUE #5230: ocean initial condition velocity is uninitialized for idealized cases
In the MPAS-Ocean init mode, most idealized cases did not explicitly set `normalVelocity` to zero, and previously the output for the initial state would simply default to zeros. With changes in the fill value this past month, the `normalVelocity` for these cases is set to the fill value, so the forward simulation fails on the first step. This can be seen with soma 32 km test case:
```
compass setup -p $COMPILE_DIR -w $WORKDIR -n 249 
Setting up test cases:
  ocean/soma/32km/default

cd $WORKDIR
cd ocean/soma/32km/default/initial_state
compass run
ncdump -v normalVelocity initial_state.nc |tail -n 3
    _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _,
    _, _, _, _, _, _, _, _, _, _, _ ;
}
```
These were previously zeros.

This problem does not affect E3SM forward simulations, nor the creation of global realistic initial conditions, where the normal velocity is explicitly set to zero.

mark-petersen:
Explicitly sets normal velocity to zero at the beginning of init mode. Each test case can then overwrite the zeros with initial velocities, but most do not. In the past the unassigned arrays wrote out zeros, but this is not true now, as described in #5230.

This code is used to create MPAS-Ocean initial conditions. It is not compiled in the E3SM code base for simulations.

Fixes #5230
[BFB]

### DIFF
--- a/components/mpas-ocean/src/mode_init/mpas_ocn_init_mode.F
+++ b/components/mpas-ocean/src/mode_init/mpas_ocn_init_mode.F
@@ -267,8 +267,14 @@ module ocn_init_mode
       type (mpas_pool_type), pointer :: forcingPool
 
       type (MPAS_timeInterval_type) :: timeStep
+      real (kind=RKIND), dimension(:,:), pointer :: normalVelocity
 
       ierr = 0
+
+      ! Initialize normalVelocity to zero for cases that do not specify it.
+      call mpas_pool_get_subpool(domain % blocklist % structs, 'state', statePool)
+      call mpas_pool_get_array(statePool, 'normalVelocity', normalVelocity, 1)
+      normalVelocity(:,:) = 0.0_RKIND
 
       ! Eventually, dt should be domain specific
       timeStep = mpas_get_clock_timestep(domain % clock, ierr=ierr)


components/mpas-ocean/src/mode_init/mpas_ocn_init_mode.F
```diff
@@ -267,9 +267,15 @@ function ocn_init_mode_run(domain) result(iErr)!{{{
...
 
       type (MPAS_timeInterval_type) :: timeStep
+      real (kind=RKIND), dimension(:,:), pointer :: normalVelocity
 
       ierr = 0
 
+      ! Initialize normalVelocity to zero for cases that do not specify it.
+      call mpas_pool_get_subpool(domain % blocklist % structs, 'state', statePool)
+      call mpas_pool_get_array(statePool, 'normalVelocity', normalVelocity, 1)
+      normalVelocity(:,:) = 0.0_RKIND
```
mark-petersen, line 277: Note that I did not use a block loop here. Block loops are no longer supported in forward mode, and we never used multiple blocks for init mode anyway, so it seemed silly to add it. We could remove block loops from the rest of the code sometime, but for this PR, I just get the array from the first block. @philipwjones do you agree?

mark-petersen:
This PR fixes the previous problem. Here with SOMA,
```
compass setup -p $COMPILE_DIR -w $WORKDIR -n 249 
Setting up test cases:
  ocean/soma/32km/default

cd $WORKDIR
cd ocean/soma/32km/default/initial_state
compass run

ncdump -v normalVelocity initial_state.nc |tail -n 3
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 ;
}
```
and the forward simulation completes successfully.

components/mpas-ocean/src/mode_init/mpas_ocn_init_mode.F
xylar: I doubt @philipwjones cares if we break init mode ;-)

xylar (approved):
I did testing on Chrysalis with Intel and Intel-MPI.

I tested the 32 km default SOMA test case with this branch and with master.  As expected, I found that the test case fails in forward mode without these changes but was successful with them.

I also verified that the results from the `pr` test suite are BFB with master.  The baroclinic channel decomp test case is failing in both because of #5219

components/mpas-ocean/src/mode_init/mpas_ocn_init_mode.F
philipwjones: I care deeply and passionately ;^).  But yeah, this is fine - I've been removing them as I find them (block loops that is, not velocity)

xylar:
@cbegeman and @sbrus89, this may also fix the dam-break test case.

jonbob:
passes:

merged ot next

jonbob:
merged to master

PR #5237, opened 2022-10-16
yunpengshan2014/E3SM/E3SMv3_P3_PR-IncloudProperty

yunpengshan2014:
The in-cloud properties should be outputted after the cloud microphysics tendencies are taken into account, whereas they are outputted in P3 without accounting for the microphysics calculations. This fix does not affect model simulations, but will affect all of offline analysis where in-cloud property is used.

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -283,6 +283,14 @@ _TESTS = {
             )
         },
 
+    #e3sm MMF tests for development
+    "e3sm_v3atm_integration" : {
+        "tests" : (
+            "ERP_Ln9.ne4pg2_ne4pg2.F2010-P3",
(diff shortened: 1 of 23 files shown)


yunpengshan2014:
Hi Jacob,

I think yes. The offline analysis about in-cloud properties will be
unreasonable if thess fix is not included. You can ask Drs. Chris T. and
Wuyin Lin to confirm it.

Regards,
Yunpeng

crterai:
I do think this should was meant to be in the v3atm fork. @yunpengshan2014, would you ask Wuyin to copy this branch over to the v3atm fork? 
I am closing it for now, since it's not to be merged with master.

PR #5241, opened 2022-10-17
Hommexx: Print element state data in various model failure situations.

ambrad:
In the log file, one will see, e.g.,

 46: Bad dphi, dp3d, or vtheta_dp; label: 'Vertical remap: Negative (or nan) layer thickness detected, aborting!'; see hommexx.errlog.320.46

Then, for example,

$ cat hommexx.errlog.320.46
label: Vertical remap: Negative (or nan) layer thickness detected, aborting! time-level 1
lat  5.182512623200946e-01 lon  1.413716694115407e+00 ie 3 igll 3 jgll 3 lev 122: bad vtheta_dp bad dp3d bad dphi
level                   dphi                   dp3d              vtheta_dp
    0 -1.894494653842581e+04  6.511250007060127e+01  9.449394802849884e+04
...
  122  1.061523656197460e+02 -3.435052907124674e+01 -8.703189416949290e+03
...

This feature is for the theta-l_kokkos dycore. preqx_kokkos has an empty implementation of the check routine, so its behavior on failure is unchanged.

[BFB]

### DIFF
--- a/components/homme/src/theta-l_kokkos/cxx/CaarFunctorImpl.hpp
+++ b/components/homme/src/theta-l_kokkos/cxx/CaarFunctorImpl.hpp
@@ -334,9 +334,12 @@ struct CaarFunctorImpl {
     profiling_resume();
 
     GPTLstart("caar compute");
-    Kokkos::parallel_for("caar loop pre-boundary exchange", m_policy_pre, *this);
+    int nerr;
+    Kokkos::parallel_reduce("caar loop pre-boundary exchange", m_policy_pre, *this, nerr);
     ExecSpace::impl_static_fence();
     GPTLstop("caar compute");
+    if (nerr > 0)
+      check_print_abort_on_bad_elems("CaarFunctorImpl::run TagPreExchange", data.n0);
 
     GPTLstart("caar_bexchV");
     m_bes[data.np1]->exchange(m_geometry.m_rspheremp);
@@ -359,7 +362,7 @@ struct CaarFunctorImpl {
   }
 
   KOKKOS_INLINE_FUNCTION
-  void operator()(const TagPreExchange&, const TeamMember &team) const {
+  void operator()(const TagPreExchange&, const TeamMember &team, int& nerr) const {
     // In this body, we use '====' to separate sync epochs (delimited by barriers)
     // Note: make sure the same temp is not used within each epoch!
 
@@ -372,7 +375,8 @@ struct CaarFunctorImpl {
     kv.team_barrier();
 
     // Computes pi, omega, and phi.
-    compute_scan_quantities(kv);
+    const bool ok = compute_scan_quantities(kv);
+    if ( ! ok) nerr = 1;
 
     if (m_rsplit==0 || !m_theta_hydrostatic_mode) {
       // ============ EPOCH 2.1 =========== //
@@ -598,7 +602,9 @@ struct CaarFunctorImpl {
   }
 
   KOKKOS_INLINE_FUNCTION
-  void compute_scan_quantities (KernelVariables &kv) const {
+  bool compute_scan_quantities (KernelVariables &kv) const {
+    bool ok = true;
+    
     kv.team_barrier();
     Kokkos::parallel_for(Kokkos::TeamThreadRange(kv.team,NP*NP),
                          [&](const int idx) {
@@ -676,11 +682,13 @@ struct CaarFunctorImpl {
                             Homme::subview(m_buffers.pnh,kv.team_idx,igp,jgp),
                             Homme::subview(m_state.m_phinh_i,kv.ie,m_data.n0,igp,jgp));
       } else {
+        const bool ok1 =
         m_eos.compute_pnh_and_exner(kv,
                                     Homme::subview(m_state.m_vtheta_dp,kv.ie,m_data.n0,igp,jgp),
                                     Homme::subview(m_state.m_phinh_i,kv.ie,m_data.n0,igp,jgp),
                                     Homme::subview(m_buffers.pnh,kv.team_idx,igp,jgp),
                                     Homme::subview(m_buffers.exner,kv.team_idx,igp,jgp));
+        if ( ! ok1) ok = false;
       }
 
       // Compute phi at midpoints
@@ -688,6 +696,7 @@ struct CaarFunctorImpl {
                                             Homme::subview(m_buffers.phi,kv.team_idx,igp,jgp));
     });
     kv.team_barrier();
+    return ok;
   }
 
   KOKKOS_INLINE_FUNCTION
--- a/components/homme/src/theta-l_kokkos/cxx/DirkFunctorImpl.hpp
+++ b/components/homme/src/theta-l_kokkos/cxx/DirkFunctorImpl.hpp
@@ -232,7 +232,7 @@ struct DirkFunctorImpl {
     const auto hybi = hvcoord.hybrid_bi;
     const auto tu   = m_tu;
 
-    const auto toplevel = KOKKOS_LAMBDA (const MT& team) {
+    const auto toplevel = KOKKOS_LAMBDA (const MT& team, int& nerr) {
       KernelVariables kv(team, tu);
       const auto ie = kv.ie;
       const int nlev = num_phys_lev;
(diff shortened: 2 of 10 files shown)


ambrad:
I'll do more extensive testing before merging. For now, eam_theta_pg2 and homme_integration pass on Chrysalis. Testing in progress includes SCREAM performance testing on Summit and e3sm_developer on Chrysalis.

ambrad:
ERP_Ln90.ne30pg2_ne30pg2.F2010-SCREAMv1 on ascent passes.

ambrad:
I ran the following:
* Ascent
* 4 nodes, 6 GPUs/node, 6 ranks/node
* ne30pg2_ne30pg2.F2010-SCREAMv1-noAero
* 4 days.
The performance difference in prim_run_subcycle_c is 0.7% slower, which could be noise. The compute_stage_value_dirk timer is 0.1% faster, which of course is noise. Bottom line: The performance difference is very tiny.

ambrad:
Repeating the PR header comment with formatting here (I don't use backticks in text that will go into a git log message.): In the log file, one will see, e.g.,
```
46: Bad dphi, dp3d, or vtheta_dp; label: 'Vertical remap: Negative (or nan) layer thickness detected, aborting!'; see hommexx.errlog.46
```
Then, for example,
```
$ cat hommexx.errlog.46
label: Vertical remap: Negative (or nan) layer thickness detected, aborting! time-level 1
lat 5.182512623200946e-01 lon 1.413716694115407e+00 ie 3 igll 3 jgll 3 lev 122: bad vtheta_dp bad dp3d bad dphi
level dphi dp3d vtheta_dp
0 -1.894494653842581e+04 6.511250007060127e+01 9.449394802849884e+04
...
122 1.061523656197460e+02 -3.435052907124674e+01 -8.703189416949290e+03
...
```

components/homme/src/theta-l_kokkos/cxx/CaarFunctorImpl.hpp
```diff
@@ -334,9 +334,12 @@ struct CaarFunctorImpl {
     profiling_resume();
 
     GPTLstart("caar compute");
-    Kokkos::parallel_for("caar loop pre-boundary exchange", m_policy_pre, *this);
+    int nerr;
```
bartgol, line 337: Despite being ok, I wonder if valgrind will complain with something like "conditional jump may depend on uninitialized value" down below.

components/homme/src/theta-l_kokkos/cxx/ElementsState.cpp
```diff
@@ -338,4 +342,118 @@ void ElementsState::push_to_f90_pointers (F90Ptr& state_v, F90Ptr& state_w_i, F9
...
+          p = (std::isnan(phinh_i(ie,tlvl,gi,gj,k)) || std::isnan(phinh_i(ie,tlvl,gi,gj,k+1)) ||
+               phinh_i(ie,tlvl,gi,gj,k) < phinh_i(ie,tlvl,gi,gj,k+1));
+          if (v || d || p) {
+            k_bad = k;
+            break;
+          }
+        }
+        if (k_bad >= 0) {
+          if (first) {
+            filename = std::string("hommexx.errlog.") + std::to_string(comm.rank());
```
bartgol, line 432: Very minor, but if we do some unit testing with multiple comm sizes, we should have the filename include the comm size, to avoid clashes of different tests. Something like `$size.$rank`.

components/homme/src/theta-l_kokkos/cxx/DirkFunctorImpl.hpp
```diff
@@ -382,9 +386,10 @@ struct DirkFunctorImpl {
       } // Newton iteration
       kv.team_barrier();
 
-      if (it>=maxiter) {
-        printf ("[DIRK] WARNING! Newton reached max iteration count,"
-                " with deltaerr = %3.17f\n", deltaerr);
+      if (it >= maxiter) {
```
bartgol, line 389: Shouldn't we also check that `deltaerr>=deltatol`? Reaching convergence at the last iteration should not raise an error, no?

components/homme/src/theta-l_kokkos/cxx/EquationOfState.hpp
```diff
@@ -81,22 +81,19 @@ class EquationOfState {
...
       // check inputs
       const int vec_len = (ilev==(NUM_LEV-1)) ? ColInfo<NUM_PHYSICAL_LEV>::LastPackLen : VECTOR_SIZE;
       for (int iv=0; iv<vec_len; ++iv) {
-        if (vtheta_dp(ilev)[iv]<0.0) {
-          Kokkos::abort("Error! vtheta_dp>0 detected.\n");
-        }
-        if (exner(ilev)[iv]>0.0) {
-          Kokkos::abort("Error! dphi>0 detected.\n");
-        }
+        if (vtheta_dp(ilev)[iv] < 0.0 || exner(ilev)[iv] > 0.0)
```
bartgol, line 90: If may consider moving this check inside the `compute_pnh_and_exner` fcn below (and make it return nerr or a bool), and remove the check in the DirkFunctor, since they would both use the same. Or not.

components/homme/src/theta-l_kokkos/cxx/DirkFunctorImpl.hpp
mt5555: the warning doesn't actually say its not converging, only that we reached maxiter :-)

mt5555 (commented):
clean & efficient!

components/homme/src/theta-l_kokkos/cxx/ElementsState.cpp
ambrad: Good idea. I'll add the comm size to the filename.

components/homme/src/theta-l_kokkos/cxx/DirkFunctorImpl.hpp
bartgol: Right, but setting `nerr = 1` will later trigger an abort.

ambrad: I agree with Mark. Also, this is from the F90; I just reformatted the code here.

components/homme/src/theta-l_kokkos/cxx/CaarFunctorImpl.hpp
bartgol: Ah right, I forgot about ||reduce doing the init. Nvm then.

components/homme/src/theta-l_kokkos/cxx/DirkFunctorImpl.hpp
ambrad: It triggers the check routine to be called. The check routine may or may not abort, depending on the actual values of the state variables.

bartgol: So if the states are valid, it won't abort, even if it did not converge, correct?

components/homme/src/theta-l_kokkos/cxx/EquationOfState.hpp
ambrad: I had thought about it but decided I don't like the idea of our lowest-level routines that do raw computations being polluted with always-on checks.

components/homme/src/theta-l_kokkos/cxx/DirkFunctorImpl.hpp
ambrad: Correct.

components/homme/src/theta-l_kokkos/cxx/ElementsState.cpp
```diff
@@ -338,4 +342,120 @@ void ElementsState::push_to_f90_pointers (F90Ptr& state_v, F90Ptr& state_w_i, F9
...
+          p = (std::isnan(phinh_i(ie,tlvl,gi,gj,k)) || std::isnan(phinh_i(ie,tlvl,gi,gj,k+1)) ||
+               phinh_i(ie,tlvl,gi,gj,k) < phinh_i(ie,tlvl,gi,gj,k+1));
+          if (v || d || p) {
+            k_bad = k;
+            break;
+          }
+        }
+        if (k_bad >= 0) {
+          if (first) {
+            filename = (std::string("hommexx.errlog.") +
```
ambrad, line 432: @bartgol the file will now be named something like `hommexx.320.46`.

components/homme/src/theta-l_kokkos/cxx/ElementsState.cpp
ambrad: Done. See below. I've edited the commit message to reflect the change.

bartgol (approved):

PR #5150, opened 2022-08-23
Updates the settings for JRA data

bishtgautam:
Removes the hard-coded start and end year for JRA from 1958/2016 to be
the start/end values specified by DATM_CLMNCEP_YR_START/DATM_CLMNCEP_YR_END.

[BFB]

PR #5243, opened 2022-10-19
Update google cloud machine to gnu gcc-12.2

sarich:
Update google cloud machine to gnu gcc-12.2

e3sm_integration passes
e3sm_extra_coverage has no new errors:
memory leak/compare base fail:
 ERP_Lm3.ne4_oQU240.F2010.gcp_gnu
setup fails for:
 SMS_D_Ln5.twpx4v1_twpx4v1.F2010-CICE.gcp_gnu
 SMS_D_Ln5.enax4v1_enax4v1.F2010-CICE.gcp_gnu

[BFB]

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -4247,24 +4247,24 @@
       </modules>
 
       <modules compiler="gnu">
-	<command name="load">gcc-11.1.0</command>
+	<command name="load">gcc/12.2.0</command>
       </modules>
 
       <modules mpilib="openmpi">
-	<command name="load">openmpi-gcc@11.1.0</command>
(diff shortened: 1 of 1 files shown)


ndkeen (approved):
I tested this with e3sm_developer and all passed except for one.  

Update: the branch I tested did not actually have PR 5150, so ignore following comment:
`ERS.ne30pg2_r05_EC30to60E2r2.GPMPAS-JRA.gcp_gnu.mosart-rof_ocn_2way` still fails as we have permissions set to avoid filling up disk with data fro this test.  I had though that https://github.com/E3SM-Project/E3SM/pull/5150 was supposed to fix that, but it looks like the test is still wanting to download ~1400 files?

rljacob:
@bishtgautam see above.

Issue #5242, opened 2022-10-18
Imprint of atmospheric grid visible on some surface sea ice fields


njeffery:
A checkboard pattern is visible in monthly climatologies of melt-pond fraction for ne30 atm grids and EC30to60E2r2.

rljacob:
Can you include a figure showing it?

njeffery:
![pondAreaCellNH_v2 LR piControl_09_years0015-0116](https://user-images.githubusercontent.com/8211178/196525464-5d03a3b3-064c-41af-8e6c-fbe8d55d6ae6.png)

whannah1:
@njeffery are you able to plot the atmospheric fields on the native grid?
If not then can you point me to where the data is?
I'd like to check a few fields.
This might be the same issue that I documented here:
https://acme-climate.atlassian.net/wiki/spaces/NGDAP/pages/2953512527/Checkerboard+noise+problem+related+to+the+dCAPE+trigger

rljacob:
and the imprint is only in that field?  no others?

njeffery:
I'm looking at some of the other fields now.  I suspect it'll show up in albedo, shortwave surfaces fluxes, that sort of thing.  

I looked at an RRM-EC30to60E2r2 run that @jonbob had done and the checkerboard disappeared in the Arctic (still present in the Southern Ocean). 

Chris's run is archived on NERSC. I have a comparable run on anvil that shows the same pattern. Here:
/lcrc/group/e3sm/ac.jeffery/E3SMv2/v2.snow.snowRedisOnly/run

njeffery:
We made a change in v2 in how ponds form in response to snow melt.  There is a namelist flag that is intended to turn off that feature.  It's not coded properly at the moment but I'll change that soon and submit a PR.  With the snowmelt-to-pond flag false, the checkerboard disappears.  Here's a figure from a comparable run:


![pondAreaCellNH_small_v2 snow snowModsNoLiqPonds_06_years0010-0072](https://user-images.githubusercontent.com/8211178/196531619-e1cc411a-6fcf-43ee-8496-98fc2e6dec4c.png)

rljacob:
Interesting.  Would still like to know if shows up in other fields (and if the flag matters for those).

whannah1:
FYI I looked at a few fields from the run on LCRC that @njeffery pointed to. I mainly just plotted 1 or 10 year averages, but I'm not seeing any detectable checkerboard over the poles.
[...]

PR #5244, opened 2022-10-19
Allows config_use_snow_liquid_ponds = false with snow tracers

njeffery:
With the flag false, the routine to drain snow liquid was omitted. However, this routine is needed to properly keep track of snow liquid. The flag now only controls whether the drained snow-melt contributes to pond formation or not.

[BFB]

### DIFF
--- a/components/mpas-seaice/src/column/ice_colpkg.F90
+++ b/components/mpas-seaice/src/column/ice_colpkg.F90
@@ -2284,16 +2284,15 @@
          endif   ! aicen_init
 
       !-----------------------------------------------------------------
-      ! Transport liquid water in snow between layers and 
+      ! Transport liquid water in snow between layers and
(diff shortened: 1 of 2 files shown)


jonbob:
@njeffery - does this fix [Issue #5242](https://github.com/E3SM-Project/E3SM/issues/5242)? Or is it a way to simply avoid it?

njeffery:
@jonbob: Currently neither as the PR doesn't change defaults. But it will allow us to set config_use_snow_liquid_ponds = false and that's looking like a possible solution to #5244

akturner (approved):

eclare108213 (approved):

jonbob:
Passes:

merged to next

jonbob:
merged to master

PR #5252, opened 2022-10-26
Fix floating-point exception in MPASSI incremental remap code

### LINKED ISSUE #5463: Floating-point exception with SMS_D_P4096.ne120pg2_r0125_oRRS18to6v3.WCYCL1950.chrysalis_gnu
Error with this test `SMS_D_P4096.ne120pg2_r0125_oRRS18to6v3.WCYCL1950.chrysalis_gnu`

```

3097: Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.
3097: 
3097: Backtrace for this error:
3097: #0  0x15554ed3a3ff in ???
3097: #1  0x4760382 in compute_barycenter_coordinates
3097:   at /lcrc/group/e3sm/ac.ndkeen/scratch/chrys/SMS_D_P4096.ne120pg2_r0125_oRRS18to6v3.WCYCL1950.chrysalis_gnu.20230217_143908_w0vt3t/bld/cmake-bld/core_seaice/shared/mpas_seaice_advection_incremental_remap.f90:4772
3097: #2  0x477de30 in construct_linear_tracer_fields
3097:   at /lcrc/group/e3sm/ac.ndkeen/scratch/chrys/SMS_D_P4096.ne120pg2_r0125_oRRS18to6v3.WCYCL1950.chrysalis_gnu.20230217_143908_w0vt3t/bld/cmake-bld/core_seaice/shared/mpas_seaice_advection_incremental_remap.f90:3881
3097: #3  0x4795459 in incremental_remap_block
3097:   at /lcrc/group/e3sm/ac.ndkeen/scratch/chrys/SMS_D_P4096.ne120pg2_r0125_oRRS18to6v3.WCYCL1950.chrysalis_gnu.20230217_143908_w0vt3t/bld/cmake-bld/core_seaice/shared/mpas_seaice_advection_incremental_remap.f90:3067
3097: #4  0x4797110 in __seaice_advection_incremental_remap_MOD_seaice_run_advection_incremental_remap
3097:   at /lcrc/group/e3sm/ac.ndkeen/scratch/chrys/SMS_D_P4096.ne120pg2_r0125_oRRS18to6v3.WCYCL1950.chrysalis_gnu.20230217_143908_w0vt3t/bld/cmake-bld/core_seaice/shared/mpas_seaice_advection_incremental_remap.f90:2575
3097: #5  0x3cd7244 in __seaice_advection_MOD_seaice_run_advection
3097:   at /lcrc/group/e3sm/ac.ndkeen/scratch/chrys/SMS_D_P4096.ne120pg2_r0125_oRRS18to6v3.WCYCL1950.chrysalis_gnu.20230217_143908_w0vt3t/bld/cmake-bld/core_seaice/shared/mpas_seaice_advection.f90:138
3097: #6  0x3cd6fff in __seaice_time_integration_MOD_seaice_timestep
3097:   at /lcrc/group/e3sm/ac.ndkeen/scratch/chrys/SMS_D_P4096.ne120pg2_r0125_oRRS18to6v3.WCYCL1950.chrysalis_gnu.20230217_143908_w0vt3t/bld/cmake-bld/core_seaice/shared/mpas_seaice_time_integration.f90:159
3097: #7  0x3b56f88 in __ice_comp_mct_MOD_ice_run_mct
3097:   at /lcrc/group/e3sm/ac.ndkeen/scratch/chrys/SMS_D_P4096.ne120pg2_r0125_oRRS18to6v3.WCYCL1950.chrysalis_gnu.20230217_143908_w0vt3t/mpas-seaice/driver/ice_comp_mct.f90:1136
3097: #8  0x44e828 in __component_mod_MOD_component_run
3097:   at /lcrc/group/e3sm/ac.ndkeen/wacmy/m31-feb17/driver-mct/main/component_mod.F90:728
3097: #9  0x431ee7 in __cime_comp_mod_MOD_cime_run
3097:   at /lcrc/group/e3sm/ac.ndkeen/wacmy/m31-feb17/driver-mct/main/cime_comp_mod.F90:2897
3097: #10  0x44bd6d in cime_driver
3097:   at /lcrc/group/e3sm/ac.ndkeen/wacmy/m31-feb17/driver-mct/main/cime_driver.F90:153
3097: #11  0x44bdd4 in main
3097:   at /lcrc/group/e3sm/ac.ndkeen/wacmy/m31-feb17/driver-mct/main/cime_driver.F90:23


/lcrc/group/e3sm/ac.ndkeen/scratch/chrys/SMS_D_P4096.ne120pg2_r0125_oRRS18to6v3.WCYCL1950.chrysalis_gnu.20230217_143908_w0vt3t

```

jonbob:
High-res (ne120pg2_r0125_oRRS18to6v3.WCYCL1950) B-case tests on summit using gnu have been failing with a floating-point exception that points to a line in the seaice incremental remap code:
````
       if (abs(massTracerProd) > 0.0_RKIND) then
````
line 4769 of mpas_seaice_advection_incremental_remap.F. @whannah1 reported the issue and has found that the error occurs when massTracerProd is extremely small, so its inverse ends up being extremely large. He also found that it can be avoided by replacing the 0.0 with a tiny number. However, after discussion this PR fixes the issue by removing the reciprocal calculation and dividing by the mean0 (and massTracerProd) quantities instead -- @ambrad thank you for the suggestion.

Fixes #5463

[non-BFB]

### DIFF
--- a/components/mpas-seaice/src/shared/mpas_seaice_advection_incremental_remap.F
+++ b/components/mpas-seaice/src/shared/mpas_seaice_advection_incremental_remap.F
@@ -4663,9 +4663,6 @@ contains
        mean1, center1, xGrad1, yGrad1,  &
        mean2, center2, xGrad2, yGrad2)
 
-    use seaice_constants, only: &
-         seaicePuny
-
     type(geometric_avg_cell_type), pointer :: &
          geomAvgCell                       !< Input: derived type holding geometric averages
 
@@ -4692,7 +4689,7 @@ contains
          cxxx,  cxxy,  cxyy,  cyyy,   &
          cxxxx, cxxxy, cxxyy, cxyyy, cyyyy
 
-    real(kind=RKIND) :: reciprocal, massTracerProd
+    real(kind=RKIND) ::  massTracerProd
 
     ! compute the barycenter coordinates, depending on how many tracer types were passed in
     ! Note: These formulas become more complex as the tracer hierarchy deepens.  See the model documentation for details.
@@ -4709,18 +4706,18 @@ contains
        c0 = center0
        cx = xGrad0
        cy = yGrad0
-       if (abs(mean0) > seaicePuny) then
-          reciprocal = 1.0_RKIND / mean0
-       else
-          reciprocal = 0.0_RKIND
+
+       xBarycenter = 0.0_RKIND
+       yBarycenter = 0.0_RKIND
+       if (abs(mean0) > 0.0_RKIND) then
+          xBarycenter = (c0 * geomAvgCell % x (iCell)  &
+                       + cx * geomAvgCell % xx(iCell)  + cy * geomAvgCell % xy(iCell))  &
+                       / mean0
+          yBarycenter = (c0 * geomAvgCell % y (iCell)  &
+                       + cx * geomAvgCell % xy(iCell)  + cy * geomAvgCell % yy(iCell))  &
+                       / mean0
        endif
 
-       xBarycenter = (c0 * geomAvgCell % x (iCell)  &
-                    + cx * geomAvgCell % xx(iCell)  + cy * geomAvgCell % xy(iCell))  &
-                    * reciprocal
-       yBarycenter = (c0 * geomAvgCell % y (iCell)  &
-                    + cx * geomAvgCell % xy(iCell)  + cy * geomAvgCell % yy(iCell))  &
-                    * reciprocal
 
     elseif (.not.present(mean2)) then
 
@@ -4733,21 +4730,19 @@ contains
        cxy = xGrad0  * yGrad1  + yGrad0  * xGrad1
        cyy = yGrad0  * yGrad1
 
+       xBarycenter = 0.0_RKIND
+       yBarycenter = 0.0_RKIND
        massTracerProd = mean0 * mean1
-       if (abs(massTracerProd) > seaicePuny) then
-          reciprocal = 1.0_RKIND / massTracerProd
-       else
-          reciprocal = 0.0_RKIND
+       if (abs(massTracerProd) > 0.0_RKIND) then
+          xBarycenter = (c0  * geomAvgCell % x  (iCell) &
+                       + cx  * geomAvgCell % xx (iCell) + cy  * geomAvgCell % xy (iCell)  &
+                       + cxx * geomAvgCell % xxx(iCell) + cxy * geomAvgCell % xxy(iCell) + cyy * geomAvgCell % xyy(iCell)) &
+                       / massTracerProd
+          yBarycenter = (c0  * geomAvgCell % y  (iCell) &
+                       + cx  * geomAvgCell % xy (iCell) + cy  * geomAvgCell % yy (iCell)  &
+                       + cxx * geomAvgCell % xxy(iCell) + cxy * geomAvgCell % xyy(iCell) + cyy * geomAvgCell % yyy(iCell)) &
+                       / massTracerProd
        endif
-
-       xBarycenter = (c0  * geomAvgCell % x  (iCell) &
-                    + cx  * geomAvgCell % xx (iCell) + cy  * geomAvgCell % xy (iCell)  &
-                    + cxx * geomAvgCell % xxx(iCell) + cxy * geomAvgCell % xxy(iCell) + cyy * geomAvgCell % xyy(iCell)) &
-                    * reciprocal
-       yBarycenter = (c0  * geomAvgCell % y  (iCell) &
-                    + cx  * geomAvgCell % xy (iCell) + cy  * geomAvgCell % yy (iCell)  &
-                    + cxx * geomAvgCell % xxy(iCell) + cxy * geomAvgCell % xyy(iCell) + cyy * geomAvgCell % yyy(iCell)) &
-                    * reciprocal
 
     else  ! mass field, tracer1 and tracer2 are all present
 
@@ -4765,25 +4760,23 @@ contains
        cxyy  = yGrad0  * yGrad1  * xGrad2   + yGrad0  * xGrad1  * yGrad2  + xGrad0  * yGrad1  * yGrad2
        cyyy  = yGrad0  * yGrad1  * yGrad2
 
+       xBarycenter = 0.0_RKIND
+       yBarycenter = 0.0_RKIND
        massTracerProd = mean0 * mean1 * mean2
-       if (abs(massTracerProd) > seaicePuny) then
-          reciprocal = 1.0_RKIND / massTracerProd
-       else
-          reciprocal = 0.0_RKIND
+       if (abs(massTracerProd) > 0.0_RKIND) then
+          xBarycenter = (c0   * geomAvgCell % x   (iCell) &
+                       + cx   * geomAvgCell % xx  (iCell) + cy   * geomAvgCell % xy  (iCell) &
+                       + cxx  * geomAvgCell % xxx (iCell) + cxy  * geomAvgCell % xxy (iCell) + cyy  * geomAvgCell % xyy (iCell) &
+                       + cxxx * geomAvgCell % xxxx(iCell) + cxxy * geomAvgCell % xxxy(iCell) + cxyy * geomAvgCell % xxyy(iCell) &
+                                                                                             + cyyy * geomAvgCell % xyyy(iCell)) &
+                       / massTracerProd
+          yBarycenter = (c0   * geomAvgCell % y   (iCell) &
+                       + cx   * geomAvgCell % xy  (iCell) + cy   * geomAvgCell % yy  (iCell) &
+                       + cxx  * geomAvgCell % xxy (iCell) + cxy  * geomAvgCell % xyy (iCell) + cyy  * geomAvgCell % yyy (iCell) &
+                       + cxxx * geomAvgCell % xxxy(iCell) + cxxy * geomAvgCell % xxyy(iCell) + cxyy * geomAvgCell % xyyy(iCell) &
+                                                                                             + cyyy * geomAvgCell % yyyy(iCell)) &
+                       / massTracerProd
        endif
-
-       xBarycenter = (c0   * geomAvgCell % x   (iCell) &
-                    + cx   * geomAvgCell % xx  (iCell) + cy   * geomAvgCell % xy  (iCell) &
-                    + cxx  * geomAvgCell % xxx (iCell) + cxy  * geomAvgCell % xxy (iCell) + cyy  * geomAvgCell % xyy (iCell) &
-                    + cxxx * geomAvgCell % xxxx(iCell) + cxxy * geomAvgCell % xxxy(iCell) + cxyy * geomAvgCell % xxyy(iCell) &
-                                                                                          + cyyy * geomAvgCell % xyyy(iCell)) &
-                    * reciprocal
-       yBarycenter = (c0   * geomAvgCell % y   (iCell) &
-                    + cx   * geomAvgCell % xy  (iCell) + cy   * geomAvgCell % yy  (iCell) &
-                    + cxx  * geomAvgCell % xxy (iCell) + cxy  * geomAvgCell % xyy (iCell) + cyy  * geomAvgCell % yyy (iCell) &
-                    + cxxx * geomAvgCell % xxxy(iCell) + cxxy * geomAvgCell % xxyy(iCell) + cxyy * geomAvgCell % xyyy(iCell) &
-                                                                                          + cyyy * geomAvgCell % yyyy(iCell)) &
-                    * reciprocal
 
     endif
 
--- a/components/mpas-seaice/src/shared/mpas_seaice_advection_incremental_remap.F
+++ b/components/mpas-seaice/src/shared/mpas_seaice_advection_incremental_remap.F
@@ -4663,6 +4663,9 @@ contains
        mean1, center1, xGrad1, yGrad1,  &
        mean2, center2, xGrad2, yGrad2)
 
+    use seaice_constants, only: &
+         seaicePuny
+
     type(geometric_avg_cell_type), pointer :: &
          geomAvgCell                       !< Input: derived type holding geometric averages
 
@@ -4706,7 +4709,7 @@ contains
        c0 = center0
        cx = xGrad0
        cy = yGrad0
-       if (abs(mean0) > 0.0_RKIND) then
+       if (abs(mean0) > seaicePuny) then
           reciprocal = 1.0_RKIND / mean0
        else
           reciprocal = 0.0_RKIND
@@ -4731,7 +4734,7 @@ contains
        cyy = yGrad0  * yGrad1
 
        massTracerProd = mean0 * mean1
-       if (abs(massTracerProd) > 0.0_RKIND) then
+       if (abs(massTracerProd) > seaicePuny) then
           reciprocal = 1.0_RKIND / massTracerProd
        else
           reciprocal = 0.0_RKIND
@@ -4763,7 +4766,7 @@ contains
        cyyy  = yGrad0  * yGrad1  * yGrad2
 
        massTracerProd = mean0 * mean1 * mean2
-       if (abs(massTracerProd) > 0.0_RKIND) then
+       if (abs(massTracerProd) > seaicePuny) then
           reciprocal = 1.0_RKIND / massTracerProd
        else
           reciprocal = 0.0_RKIND

### CODE REVIEW

**Comment by jonbob** [comment]
Do not merge until testing verifies the impact of this change

File: components/mpas-seaice/src/shared/mpas_seaice_advection_incremental_remap.F [hunk 1]
```diff
@@ -4663,6 +4663,9 @@ subroutine compute_barycenter_coordinates(&
        mean1, center1, xGrad1, yGrad1,  &
        mean2, center2, xGrad2, yGrad2)
 
+    use seaice_constants, only: &
+         seaicePuny
```
sarats, line 4667: Curious, seaicePuny seems like an existing constant? Just wondering what value was chosen and any rationale.

jonbob: @sarats - it was just an existing constant that at least seemed appropriate. I'm open to other ideas -- is there a reasonable constant for which the inverse will always be OK for all compilers?

eclare108213 (commented):
I would have expected that either such a small mass of ice would have been removed (mass and heat added to the ocean) prior to this point in the code, or else that the velocity is zero (because the dynamics must be limited to higher-mass ice cells).  In that case, none of the advection routines should be called at all, which might improve the MPAS-SI performance issue.  I guess the velocity for small ice masses might be set to the surface ocean current -- I'm not sure what's done in this model.  But if this change works, go for it.

akturner (approved):

proteanplanet (commented):
I suggest that we need to check compiler issues before making this change, which I'm not convinced is wise.

jonbob:
Thanks @proteanplanet -- I'm looked into issues with gfortran and flushing to zero, which this seems to represent. I have a small test code to try to replicate this issue:
````
      program test
        integer i
        real r, r2
        r=2e-36
        do i=1,50 ! 50 iterations max
           r = r * 0.5
           if (r.gt.0.0) r2=1.0/r
           print *,'r on next line'
           print *,i,r,r2
           if (r == 0.0) exit
        end do
      end program
      ````
Compiling this with -O shows r2 becoming an Inf at iteration 10, though r does not become 0.0 until iteration 31. With the typical (I think) e3sm debug settings of
````
 -g -Wall -fbacktrace -fcheck=bounds -ffpe-trap=zero,overflow
````
I get the "Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation" error in iteration 10, so very much like the traceback @whannah1 observed. I propose trying with a new compiler setting for just this file, "-funsafe-math-optimizations". With the -O compilation and this flag, the test code stops in iteration 8 from r becoming 0.0. Added to the debug flags above, I get the same result -- the test stops after iteration 8 because r has become 0.0. Does this seem like a better solution? If so, I can ask @whannah1 to test it in the real problem.

**Comment by philipwjones** [comment]
The compiler is behaving correctly - it's trapping a legitimate overflow due to the division of a small number (that's close to an underflow) and we've asked it with the overflow flag to die in this case. With the overflow flag off, it's presumably creating an inf that gets masked somewhere else. If you don't like puny, you could try to push the limits using the Fortran tiny/huge intrinsics to try to get as close as you can to an underflow/overflow condition.

**Comment by jonbob** [comment]
@philipwjones - is there a reason we shouldn't use the "-funsafe-math-optimizations" flag for this file?

**Comment by philipwjones** [comment]
Well...it does say unsafe right there in the name ;^) ... I suspect it might impact bit-for-bit when optimization is on since I think that flag allows associative reordering. But I don't know how aggressive the gnu compiler is so don't really know if it would help or hurt.

**Comment by jonbob** [comment]
Thanks @philipwjones -- always the safe one! @proteanplanet - how would you feel if we used "tiny" instead of "puny"?

**Comment by philipwjones** [comment]
Speaking of safety...if you do decide to go the TINY() route, you might want to set the small number a factor away from tiny (eg. 10.d0*TINY(x)) just to give yourself a little buffer so you're not right on the edge. Another option is the intrinsic EPSILON which returns the smallest number E such that 1+E > 1. So you could have 1.d0/(r+EPSILON(r)).

**Comment by jonbob** [comment]
Thanks @philipwjones - I was thinking that something like that is a better approach, if @proteanplanet agrees. I was going to test fortran comparing tiny with 1./huge -- it seems like the problem comes from 1./tiny being greater than huge. Or maybe the threshold should be the larger of tiny and 1./huge, just to cover differences in compilers or platforms? But I can imagine that even that check would lead to problems, so perhaps the epsilon approach would be good.

**Review by proteanplanet — COMMENTED** [review]
@philipwjones and @jonbob. I am more confortable with the TINY option than puny, but I am still uneasy about this change and here is why:  1) It's in a very well exercised area of the model code that has been used for tens of thousands of years of integration, so why is this cropping up now?  I understand locally why it is happening, but I would like to understand more within E3SM what is triggering it by seeing some physical values. Is this truly the first time this code has been used with gnu compilers?   2) Were this for an ocean tracer, I would be far more comfortable, but since in the sea ice model categories in distribution, we are not locally conserving mass or enthalpy with this option, albeit in a limited way.  However, it is point (1) that is my main concern.

**Comment by eclare108213** [comment]
I agree with @proteanplanet that it would be wise to understand what is generating this problem, before fixing what might just be a symptom of some other issue.  His point (2) needs a closer look too -- I would expect that with this option, mass and tracers that are below the cutoff criterion (whatever it is, and for whichever category) would just stay where they are and not be advected, and therefore total mass and tracers would continue to be conserved.  There probably is an option in this code to check hemispheric conservation before and after advection, possibly a namelist option or maybe hardcoded.

**Comment by philipwjones** [comment]
@proteanplanet - part of the reason you're not seeing this is that this is a debug flag that's turned on to trap overflows. So it's possible that in production with the optimized code, the particular instance you're seeing either isn't happening or that this is happening somewhere that's either eventually getting masked or adjusted later so the Inf never shows up (as @eclare108213 I think it suggesting). So if y'all want to look into why it's happening to be safe, please do.

**Comment by philipwjones** [comment]
Oops - might have that wrong. That flag might have just been in @jonbob 's test...

**Comment by proteanplanet** [comment]
On a lighter note, I took -funsafe to be that it is both a fun and safe option, hence -funsafe. ;-)

**Comment by proteanplanet** [comment]
@jonbob I will communicate via Slack to set up the crashing case for myself.

**Comment by ambrad** [comment]
One modification that might be worth considering is to remove the `reciprocal` calculation `1/mean0` and similar expressions. (I refer to just `mean0` subsequently for simplicity.) Instead, just divide by the quantity `mean0`. The check on `mean0` being 0 still needs to be done.

This suggestion is based on the following points about denormalized numbers. As Phil pointed out above, quite likely the issue is that `mean0` is a denormalized number, and 1/denorm is generally (except for very near the denorm limit) inf.

Consider this code that probes denormalized numbers.
```c++
#include <cstdio>
#include <limits>

int main () {
  printf("min %e denorm_min %e max %e\n",
         std::numeric_limits<double>::min(),
         std::numeric_limits<double>::denorm_min(),
         std::numeric_limits<double>::max());
  { // Reciprocal of smallest normalized number.
    const auto min = std::numeric_limits<double>::min();
    const auto r = 1/min;
    printf("min %e r %e\n", min, r);
  }
  { // Reciprocal of a denormalized number is almost always inf.
    const auto dmin = std::numeric_limits<double>::min()/4;
    const auto r = 1/dmin;
    printf("dmin %e r %e\n", dmin, r);
  }
  { // A small number divided by the smallest denormalized number gives a
    // reasonable result.
    const auto dmin = std::numeric_limits<double>::denorm_min();
    const auto num = std::numeric_limits<double>::min();
    const auto r = num/dmin;
    printf("dmin %e r %e\n", dmin, r);
  }
}
```
It prints
```
min 2.225074e-308 denorm_min 4.940656e-324 max 1.797693e+308
min 2.225074e-308 r 4.494233e+307
dmin 5.562685e-309 r inf
dmin 4.940656e-324 r 4.503600e+15
```
The first line is information about double-precision limits. The second line shows that the reciprocal of the smallest normalized number is fine. The third line shows that the reciprocal of the largested denormalized number divided by 4 (or larger) is inf. This is likely what is happening in this issue. Finally, the fourth line shows that there is no problem using the smalllest denormalized number in division as long as the numerator is small enough.

In the barycentry calculation, the numerator in fact should be proportional to `mean0`, since these are center-of-mass type calculations. Thus, I hypothesize that dividing by `mean0` directly, rather than taking its reciprocal and then multiplying, will give results like the last line of the output above. The result is you don't need to bring in any min tol value. The mod would be non-BFB, though, since in finite precision, `a*(1/b) != a/b` in general.

jonbob:
Thanks @ambrad. What compiler did you test that with? As @proteanplanet said, we've run this code for 1000's of years and just now seeing this issue and just using gfortran. I think somehow we are getting values very near the denorm limit -- my little test showed that mean0 could get very small but not be denormalized while 1/mean0 was denorm. But again, this was just for a particular compiler. So I agree, your solution of simply dividing by mean0 instead of multiplying by 1/mean0 is an excellent idea. Thanks so much

ambrad:
> What compiler did you test that with? 

gnu 7.5.0. I compiled with -O3: `g++ -O3 tiny.cpp`.

> As @proteanplanet said, we've run this code for 1000's of years and just now seeing this issue and just using gfortran.

Denormalized tracer values may be a result of the particular configuration Walter is using. Such values might point to a higher-level issue. Nonetheless, it's probably still worth improving the robustness of the numerics in this routine.

> I think somehow we are getting values very near the denorm limit -- my little test showed that mean0 could get very small but not be denormalized while 1/mean0 was denorm.

Just for clarity, only small numbers can be denormalized (denormalized means the most significant binary figure is to the right of the decimal point, while normalized means it's the first and only digit to the left). It's a soft transition to comlete underflow. There is no equivalent for large numbers and overflow, since the most significant digit can't move any farther left of the decimal point.

In your test, you used single precision. In my program above, switching `double` to `float` gives
```
min 1.175494e-38 denorm_min 1.401298e-45 max 3.402823e+38
dmin 1.175494e-38 r 8.507059e+37
dmin 2.938736e-39 r inf
dmin 1.401298e-45 r 8.388608e+06
```
Iteration 10 in your test gives 2e-36*2**-10 = 1.953125e-39, which is just a little below `numeric_limits<float>::min()/4`, agreeing with my test.

rljacob:
notes:  not bfb so after 2.1

rljacob:
status:  checking MPAS-Analysis processing.

jonbob:
After updating to match the discussion above, 
````
SMS_D_P4096.ne120pg2_r0125_oRRS18to6v3.WCYCL1950.chrysalis_gnu
````
now passes

**Comment by rljacob** [comment]
status:  need longer runs to see non-BFB effect.

**Comment by rljacob** [comment]
status:  Jon will do more testing next week.

**Comment by jonbob** [comment]
With the current implementation, here's a comparison of ten-year low-res B-cases of this PR with a baseline:

[20230406.5252test.anvil](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20230406.5252test.anvil/html/)

The non-BFB behavior is expected due to round-off errors from changing to dividing by a number from multiplying by the inverse of it previously. @ambrad, @akturner, @philipwjones, @sarats, @proteanplanet, @eclare108213 -- could you all please take a look

**Comment by philipwjones** [comment]
I'm fine with this solution and the code changes look correct to me for implementing.

**Review by ambrad — APPROVED** [review]

**Comment by eclare108213** [comment]
The explanation for the solution here makes sense to me.  I'm a little uncomfortable with the magnitude of the ice thickness changes -- they're spatially extensive and large!  But I don't have a point of reference for how large roundoff type differences from MPAS-SI in B-cases turn out to be in E3SM.  Are there standard reference runs for that?  Even pointers to some ensemble members for atmospheric roundoff-initiated ensembles would help me gauge what "normal" looks like for sea ice thickness in E3SM.

**Comment by jonbob** [comment]
@eclare108213 -- I'm not sure anyone has done comparisons of seaice thickness in any of the ensembles. This page, [v2.LR.historical](https://acme-climate.atlassian.net/wiki/spaces/EWCG/pages/3040772171/v2.LR.historical), has links to mpas-analysis from five different historical runs. There's also a larger ensemble of I think 16 historical runs, but I didn't find a similar page pointing to the analysis for those runs.

**Comment by jonbob** [comment]
@eclare108213 -- whenever you get time, could you please make sure the variance in my tests can be explained by ensemble-like differences?

**Comment by eclare108213** [comment]
I'm still a little uncomfortable with the magnitude of the changes here.  I'm comparing @jonbob's test [20230406.5252test.anvil](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20230406.5252test.anvil/html/)
with @darincomeau's test of an actual bug fix which is not a round-off level change, https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/20230501.CRYO1950.ne30pg2_SOwISC12to60E2r4.baseline-hf-bugfix_vs.chrysalis/yrs1-10/.  They are different configurations, but both are 10-year B cases.  Jon's test for this PR shows differences in the NH that are generally smaller than in Darin's tests, but they are much larger in the SH for ice volume.  Perhaps this is because Darin's runs have higher resolution in the Southern Ocean.

I also looked at these v2 runs: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.forsyth2/E3SMv2/v2.LR.historical_0151/mpas_analysis/ts_1850-2014_climo_1985-2014/sea_ice/index.html
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.forsyth2/E3SMv2/v2.LR.historical_0251/mpas_analysis/ts_1850-2014_climo_1985-2014/sea_ice/index.html
The difference plots are against data and so I'm comparing them with each other by eye, and it doesn't look like the differences are very large at all.  Can someone run MPAS-Analysis on this data, comparing the ensemble members against each other, or teach me how to do that?

The comparison with Darin's test makes me think that this change is covering up some other problem that will be harder to identify without the code crashing.  However, the arguments for the solution presented here make a lot of sense, and if the changes in Jon's test fall within the range of changes in the v2 ensemble, I'd approve.  I'm just not sure whether that is true.

**Comment by proteanplanet** [comment]
@eclare108213 I am with you this.  The point that I'm not sure is understood is that the integral of this part of the code, run five times, must conserve to computational accuracy.  The solution offered is not doing that.   I'd run a few of my own checks, and get in touch with you via Slack to guide you through running zppy.

**Comment by darincomeau** [comment]
I'll run a pair of D-cases and test for climate-changing impacts with the CICE-QC testing. Will report back.

**Comment by darincomeau** [comment]
Sorry for the delay here, but I've finally run the CICE-QC testing here, and it passes, not climate-changing:
```
Running QC test on the following directories:
  /lcrc/group/acme/ac.dcomeau/scratch/anvil/20230601.DTESTM-JRA1p5.TL319_EC30to60E2r2.anvil.ir-bugfix.intel/run
  /lcrc/group/acme/ac.dcomeau/scratch/anvil/20230601.DTESTM-JRA1p5.TL319_EC30to60E2r2.anvil.ir-bugfix-baseline.intel/run
Number of files: 61
2 Stage Test Passed
Quadratic Skill Test Passed for Northern Hemisphere
Quadratic Skill Test Passed for Southern Hemisphere
Creating map of the data (ice_thickness_20230601.DTESTM-JRA1p5.TL319_EC30to60E2r2.anvil.ir-bugfix.intel.png)
Creating map of the data (ice_thickness_20230601.DTESTM-JRA1p5.TL319_EC30to60E2r2.anvil.ir-bugfix-baseline.intel.png)
Creating map of the data (ice_thickness_20230601.DTESTM-JRA1p5.TL319_EC30to60E2r2.anvil.ir-bugfix.intel_minus_20230601.DTESTM-JRA1p5.TL319_EC30to60E2r2.anvil.ir-bugfix-baseline.intel.png)

Quality Control Test PASSED
```

MPAS-Analysis output: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/icepack-testing/20230601.DTESTM-JRA1p5.TL319_EC30to60E2r2.anvil.ir-bugfix_vs_baseline/yrs1-5/sea_ice/index.html

There are small patches of ice thickness differences larger than 5cm in Arctic summer.

jonbob:
Thanks for more testing, @darincomeau!

darincomeau:
@eclare108213 @proteanplanet this PR has been open for quite a long time and we on the MPAS-Devops team aren't sure how to proceed. Any suggestions?

proteanplanet:
> @eclare108213 @proteanplanet this PR has been open for quite a long time and we on the MPAS-Devops team aren't sure how to proceed. Any suggestions?

The next step is to conduct a comprehensive mass and energy conservation analysis, similar to what I did ahead of the V2 release, but also comparing residuals against a v2 control.   I'm working on the script for that, but it's on the backburner at the moment while I finish my contribution to two papers.

eclare108213:
> The point that I'm not sure is understood is that the integral of this part of the code, run five times, must conserve to computational accuracy. The solution offered is not doing that. I'll run a few of my own checks, and get in touch with you via Slack to guide you through running zppy.

@proteanplanet, Was the code conserving to computational accuracy before this fix?  If not, or if we've never known because the testing infrastructure for this isn't available, then I might advocate for merging this PR now and adding the conservation investigation to our long list of other issues to address.  (I suspect that the code has never conserved that well because of all the conditionals, and you could still test the hash prior to this PR merge when you have time.)

proteanplanet:
> Was the code conserving to computational accuracy before this fix?

Yes.  And therefore this PR should not just be accepted without being sure.

eclare108213:
Okay, I am fine with holding this PR until the conservation issues are better understood.  @jonbob if having this one active for so long is a problem, could you change it to 'draft'?  @proteanplanet if there is anything the rest of us can do to help, please don't hesitate to delegate!  Thanks -

proteanplanet:
One thing that can be done immediately and need not wait for me is to switch on the conservation member in the sea ice model for D-, G- and B-cases and compare residuals against V2.1.

rljacob:
@proteanplanet any progress?  This is good programming practice and should be just roundoff.

rljacob:
I vote for the alternate to just include it in the main model.

proteanplanet:
I am fine with that.  I'm working as hard as I can in the lead-up to V3, and can't get to this right now.  We have other issues in the sea ice model, and would prefer this to be in master as part of broader checks.

rljacob:
status: waiting for dashboard to clear.

jonbob:
Passes:
* ERS_Ld3.ne4_oQU240.F2010.chrysalis_intel

and DIFFs:
* ERS_Ld5.T62_oQU240.DTESTM.chrysalis_intel

so this should just impact active seaice and not in prescribed mode

jonbob:
merged to next

rljacob:
Note:  this was reverted so master would match alpha03 and alpha04 answers.  Then revert was reverted to put it back for alpha05.

PR #5262, opened 2022-10-31
Refactor E3SM build system handling of KOKKOS_OPTIONS

jgfouca:
KOKKOS_OPTIONS is now set in universal.cmake and the intent is for other macros to string(append KOKKOS_OPTIONS ...) when they want to customize.

Also KOKKOS_OPTIONS is now set up to use Kokkos' CMake build system directly instead of going through generate_makefile.

E3SM will now contain its own buildlib.kokkos script too.

[BFB]


ambrad:
> Also KOKKOS_OPTIONS is not set up to use Kokkos' CMake build system directly instead of going through generate_makefile.

"now", not "not", right?

ambrad:
Jim, I recommend that you merge this PR into a SCREAM branch and run a v1 test on Summit and one CPU machine, e.g., Chrysalis.

cime_config/machines/cmake_macros/universal.cmake
```diff
@@ -1 +1,11 @@
 set(SUPPORTS_CXX "FALSE")
+
+# Default guess at KOKKOS_OPTIONS. These can be overridden by
+# appending -D${OPTION_NAME}=Off. The CMAKE_CXX_COMPILER will
+# be handled automatically by CIME unless explicitly set in
+# KOKKOS_OPTIONS. CMAKE_INSTALL_PREFIX will be handled automatically
+# by CIME always.
+set(KOKKOS_OPTIONS "-DKokkos_ENABLE_SERIAL=On")
```
ambrad, line 8: Good. Always having Serial on is important.

share/build/buildlib.kokkos
```diff
@@ -0,0 +1,104 @@
...
+def buildlib(bldroot, installpath, case):
+    ###############################################################################
+    srcroot = case.get_value("SRCROOT")
+    kokkos_dir = os.path.join(srcroot, "externals", "kokkos")
+    expect(os.path.isdir(kokkos_dir), "Missing kokkos submodule")
+
+    # We want to get the compiler and kokkos_options from Macros.make
+    # (generated from cmake_macros), but we want to otherwise
+    # let kokkos control flags
+    make_args = get_standard_makefile_args(case, shared_lib=True)
```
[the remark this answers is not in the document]
bartgol: I'm thinking of a CIME util along the line of
```
def get_cmake_option (...):
  content = ""
  for f in files:
    content += f"include ({f})\n"

  f = open("CMakeLists.txt","w")
  f.write(content)
  f.write(f'message ("{var} = ${{{var}}}")')
  f.close()
  return run_cmd_no_fail ("cmake . | grep {var}")
```

bartgol (approved):

share/build/buildlib.kokkos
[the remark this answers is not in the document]
jgfouca: @bartgol , yes, this is a very unique use case where need to query a CMake option that is computed by the macros, not the case settings. The case control system will have already made the Macros.make file to support the sharedlib builds, so I think this solution is decent since it doesn't require any extra file creation.

bartgol:
> @ambrad , SCREAM manages its own Kokkos build through EKAT, so SCREAM CIME cases should be using buildlib.kokkos.

So, v1 cases will use buildlib.kokkos? Or will they build scream internally? I think right now they do the latter, but we ought to move to the former. EKAT already offers the possibility of using a pre-built kokkos installation, and since ocn may use Kokkos (for the SL transport), we should start testing scream with a pre-built kokkos.

ambrad (approved):

cime_config/machines/cmake_macros/universal.cmake
```diff
@@ -1 +1,11 @@
...
+if (compile_threaded)
+  string(APPEND KOKKOS_OPTIONS " -DKokkos_ENABLE_OPENMP=On")
+endif()
```
amametjanov, line 11: I think `-DKokkos_ENABLE_OPENMP=On` had always been omitted (default is `off`) even with `compile_threaded=true`, because all available CPU cores are typically occupied with OpenMP threads outside of kokkos. With this option `On`, there are build issues with pgi and ibm on Ascent for all (11) threaded tests
- https://my.cdash.org/viewTest.php?onlydelta&buildid=2240179
- https://my.cdash.org/viewTest.php?onlydelta&buildid=2242495

Can we revert to the previous configuration? Unless there's a use-case for nested threading (in addition to `Kokkos_ENABLE_OPENMPTARGET`)

ambrad: I don't disagree, but I'll just add some observations as long as I took a look:
1. The build error is in kokkos.bldlog. e3sm.bldlog hasn't been created yet.
2. The error is in a Kokkos file: `#error directive: "You enabled Kokkos OpenMP support without enabling OpenMP in the compiler!"`
3. Thus, likely the issue can be resolved simply by including the equivalent of `-fopenmp` for the PGI compiler.
4. The fact that Kokkos isn't doing this automatically in its build system might indicate it's not being given the right information about the compiler, but that's just speculation.
5. @amametjanov is right that Kokkos threads aren't used in E3SMv2. But the only Kokkos component in v2, SL transport, knows not to use the OpenMP execution space, so it's harmless to enable the OpenMP execution space in the build.
6. I believe that Jim made this change to support SCREAM.

PR #5266, opened 2022-11-02
Add Narwhal, Warhawk, Onyx machines

apcraig:
Port E3SM to several new machines, Narwhal, Warhawk, and Onyx.

A few cases were run on Narwhal, Onyx, and Warhawk to verify port.  Should not affect the scripts or source code nor any other machines.

[BFB]

### DIFF
--- a/cime_config/allactive/config_pesall.xml
+++ b/cime_config/allactive/config_pesall.xml
@@ -2640,4 +2640,78 @@
       </pes>
     </mach>
   </grid>
+  <grid name="a%ne30.+oi%oARRM60to10|a%ne30.+oi%ARRM10to60">
+    <mach name="onyx|warhawk|narwhal">
+      <pes compset=".*EAM.+ELM.+MPASSI.+MPASO.+MOSART.*" pesize="any">
+        <comment>none</comment>
+        <ntasks>
+          <ntasks_atm>924</ntasks_atm>
+          <ntasks_lnd>164</ntasks_lnd>
+          <ntasks_rof>20</ntasks_rof>
+          <ntasks_ice>840</ntasks_ice>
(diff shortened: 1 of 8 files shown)


rljacob:
@apcraig does this still need a CIME change with it?

apcraig:
We do need the CIME change, but we'll manage that separately at this point.  It doesn't need to come into E3SM with this PR.  It can come into E3SM whenever it's convenient for E3SM to update CIME.  Until then, it's a one line change in CIME we can do manually if we need it.

darincomeau:
@apcraig I put together a branch with your fix to the mpas component build-namelist scripts to remove use of `Switch`, if you want to test that it works with this branch on these machines:

https://github.com/E3SM-Project/E3SM/tree/darincomeau/scripts/remove-switch-mpas-buildnml

apcraig:
Thanks @darincomeau.  Switch is now installed on two of the DoD platforms and I will test the third when it's available.  It was not available months ago when I did the original port.  While I don't need the Switch fix at the moment, I still think the new implementation that avoids Switch is better and more robust overall and recommend creating a PR request for it.

darincomeau:
Ok thanks @apcraig , that sounds good. I'll start testing that branch on other DOE machines to prepare for a PR.

darincomeau (approved):
Approving based on developer testing, visual inspection, as well as these mods being used by a user over the last several months in https://github.com/darincomeau/E3SM/tree/E3SMv2-Arctic

apcraig:
FYI, Onyx is back up and confirmed these mods work fine there as well.  All 3 new machines now tested with E3SM master + this PR.

jonbob:

jonbob:
merged to master

PR #5272, opened 2022-11-04
Remove Switch from mpas component build-namelist

darincomeau:
This removes the use of the `Switch` package and replaces it with the same functionality in the mpas-ocean and mpas-seaice `build-namelist` scripts. This avoids problems where `Switch` may not be installed on certain machines.

Credit to @apcraig for implementing the change.

### DIFF
--- a/components/mpas-ocean/bld/build-namelist
+++ b/components/mpas-ocean/bld/build-namelist
@@ -2153,34 +2153,31 @@ sub valid_date {
 # $$month by 1 (also incrementing $$year if going from Dec to Jan) and
 # then return 0.
 
-  use Switch;
-
   my $day = shift;
   my $month = shift;
   my $year = shift;
   my $cal = shift;
 
   my $maxday = -1;
-  switch ($$month) {
-    case 1 { $maxday = 31; }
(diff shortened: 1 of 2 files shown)


darincomeau:
@jonbob @mark-petersen @xylar @akturner this is the change I was discussing in this week's meeting. Opening this now for visibility; so far I've only done a smoke test on Chrysalis, but would need to be tested on all our supported machines, hence the "Do Not Merge" tag.

This should be [BFB], but I want to confirm before putting in that tag.

apcraig (approved):
Switch is not always on machines by default, so we've used this modified version in those cases.

xylar (approved):
I don't speak perl but this looks fine to me.  To be syntactically more similar to the original switch, the subsequent `if`s could be whatever the perl for `else if` is.  But there's no real need of this code to be efficient.

mark-petersen (approved):
Looks good. If this passes the standard test cases, it is fine with me.

jonbob:
passes:

merged to next

jonbob:
merged to master

PR #5273, opened 2022-11-04
Optionally disable alarm calibration in mpas-framework

### LINKED ISSUE #5217: Start time in timeSeriesStats output from MPAS-Ocean not correct after restart

xylar:
This merge allows calls to `mpas_set_clock_time` to disable alarm calibration when setting the current time.  This is useful because re-calibrating alarms has a negative impact on `timeSeriesStats` analysis members in MPAS-Ocean and MPAS-Seaice. The timeSeriesStats AMs get initialized with alarms that should ring after one compute interval.  If alarms are re-calibrated, they ring immediately instead.  However, the analysis members typically do not get computed at the start time, so the alarm actually rings after one time step.  If the time interval desired for timeSeriesStats is not a single time step, the compute alarm should not be ringing but it is on the first time step.

Fixes #5217
[non-BFB] for mpas timeSeriesStats diagnostic output only

### DIFF
--- a/components/mpas-framework/src/framework/mpas_timekeeping.F
+++ b/components/mpas-framework/src/framework/mpas_timekeeping.F
@@ -404,7 +404,7 @@ module mpas_timekeeping
    end subroutine mpas_advance_clock
 
 
-   subroutine mpas_set_clock_time(clock, clock_time, whichTime, ierr)
+   subroutine mpas_set_clock_time(clock, clock_time, whichTime, ierr, calibrate_alarms)
 
       implicit none
 
@@ -412,14 +412,24 @@ module mpas_timekeeping
       type (MPAS_Time_type), intent(in) :: clock_time
       integer, intent(in) :: whichTime
       integer, intent(out), optional :: ierr
+      logical, intent(in), optional :: calibrate_alarms
       integer :: threadNum
+      logical :: local_calibrate_alarms
 
       threadNum = mpas_threading_get_thread_num()
+
+      if (present(calibrate_alarms)) then
+         local_calibrate_alarms = calibrate_alarms
+      else
+         local_calibrate_alarms = .true.
+      end if
 
       if ( threadNum == 0 ) then
          if (whichTime == MPAS_NOW) then
             call ESMF_ClockSet(clock % c, CurrTime=clock_time%t, rc=ierr)
-            call mpas_calibrate_alarms(clock, ierr);
+            if (local_calibrate_alarms) then
+               call mpas_calibrate_alarms(clock, ierr)
+            end if
          else if (whichTime == MPAS_START_TIME) then
             call ESMF_ClockSet(clock % c, StartTime=clock_time%t, rc=ierr)
          else if (whichTime == MPAS_STOP_TIME) then
--- a/components/mpas-ocean/driver/ocn_comp_mct.F
+++ b/components/mpas-ocean/driver/ocn_comp_mct.F
@@ -16,7 +16,7 @@ module ocn_comp_mct
    use seq_infodata_mod
    use seq_timemgr_mod
    use seq_comm_mct,      only : seq_comm_suffix, seq_comm_inst, seq_comm_name, info_taskmap_comp
-   use shr_file_mod 
+   use shr_file_mod
    use shr_cal_mod,       only : shr_cal_date2ymd
    use shr_sys_mod
    use shr_taskmap_mod,   only : shr_taskmap_write
@@ -93,8 +93,8 @@ module ocn_comp_mct
 
    character(len=StrKIND) :: runtype, coupleTimeStamp
 
-   type(seq_infodata_type), pointer :: infodata   
-   type (iosystem_desc_t), pointer :: io_system 
+   type(seq_infodata_type), pointer :: infodata
+   type (iosystem_desc_t), pointer :: io_system
    integer :: ocnLogUnit ! unit number for ocn log
 
    !! MPASO Datatypes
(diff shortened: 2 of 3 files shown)


xylar:
This PR seems to be a better solution to the problem that I have been trying to solve in #5218.  That PR will still be needed to solve the "restart" problem for timeSeriesStats but this PR is a cleaner fix to the incorrect start time.

xylar:
@akturner, @mark-petersen and @matthewhoffman, I'm asking you to review because this is a framework change.  I'm happy to discuss on our next MPAS dev-ops PR if you like.

xylar:
## Testing

I ran `SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.chrysalis_intel`:
```
./create_test SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.chrysalis_intel --walltime 1:00:00 --wait -c -b master_20221106 --baseline-root /lcrc/group/e3sm/ac.xylar/e3sm_baselines
```
comparing with the latest `master`.  The test passed.

I also ran 1 month and then a second of `GMPAS-JRA1p4_TL319_EC30to60E2r2` on Chrysalis, and found that the resulting `xtime_startDaily`, `xtime_startMonthly`, `Time`, `Time_bnds`, etc. in `timeSeriesStats` looked correct for both MPAS-Seaice and MPAS-Ocean for both months.  Results are here:
```
/lcrc/group/e3sm/ac.xasay-davis/scratch/chrys/20221104.GMPAS-JRA1p4_TL319_EC30to60E2r2.disable-calibrate-alarms.chrysalis/run
```

xylar:
Confirming that I think the new results are right and the old are not, here is `timeDaily_counter` from master:
```

 timeDaily_counter = 23, 24, 24, 24, 24 ;
```
and from this branch:
```
 timeDaily_counter = 24, 24, 24, 24, 24 ;
```
Clearly, the first hour is missing from the test on `master`.

xylar:
@jonbob, could you help me figure out what the BFB flag needs to be when `timeSeriesStats` output is affected but other results are not?

mark-petersen (approved):
@xylar thanks for your explanation of this at our last meeting. I'm approving based on your testing, showing that it fixed the timing problem. I also ran some ocean stand-alone sanity tests with gnu debug.

akturner (approved):
Passed standalone tests

matthewhoffman (approved):
@xylar , this looks like a nice, simple solution after I had time to fully understand it!

xylar:
Thank you for the prompt reviews and testing, @matthewhoffman, @akturner and @mark-petersen.

@jonbob, I believe this is now ready for you to test as you see fit.

jonbob:
Sorry @xylar -- I missed your earlier question about the BFB flag when timeSeriesStats output is affected. It definitely shows this behavior in testing (everything else is BFB, but not timeSeriesStatsDaily), so I'll check with @rljacob

rljacob:
Is that just a diagnostic output file?  No changes to the ocean state itself?

jonbob:
Thanks for jumping in @rljacob. No changes to the ocean or seaice states, just how some output gets time-averaged. But @xylar, please correct me if that's not accurate

jonbob:
PASSes the following tests with the exception of timeSeriesStats files:

jonbob:
@rljacob - we'd like to merge this soon, but would appreciate your input on handling the situation where some of the output will not be BFB despite no changes to the ocean state

rljacob:
Just mention that only diagnostics are non-BFB.

jonbob:
tested again and merged to next

jonbob:
@xylar -- we're getting a real FAIL from merging this with test ERS.ne30pg2_r05_EC30to60E2r2.GPMPAS-JRA.mosart-rof_ocn_2way on a range of platforms. I've verified it's due to this PR and trying to tease it apart a bit. But I'd appreciate any ideas or thoughts you might have about this

xylar:
I'll either need to see the logs or run that test myself. Sorry about that.

jonbob:
That's OK -- let me point you at logs from a test merge of this PR into master. I actually have four tests and I'll give you a quick overview of each as well as the path:

1. master:
ERS.ne30pg2_r05_EC30to60E2r2.GPMPAS-JRA.chrysalis_intel.mosart-rof_ocn_2way.C.20221130_122937_qznl09

2. master with test merge of this PR:
ERS.ne30pg2_r05_EC30to60E2r2.GPMPAS-JRA.chrysalis_intel.mosart-rof_ocn_2way.C.20221130_130507_fo7tfy

3. master with test merge but removing the calibrate_alarms settings in the seaice driver (also failed):
ERS.ne30pg2_r05_EC30to60E2r2.GPMPAS-JRA.chrysalis_intel.mosart-rof_ocn_2way.C.20221130_135234_cjvtto

4. master with test merge but removing the calibrate_alarms settings in the ocean driver:
ERS.ne30pg2_r05_EC30to60E2r2.GPMPAS-JRA.chrysalis_intel.mosart-rof_ocn_2way.C.20221130_143305_45g16s

As you probably expected, the last one passes the ERS part but does not fix the timeSeriesStats issue in the ocean

xylar:
@jonbob, I had a look and those are, indeed, a mess of changes that weren't expected.

I will go back to the drawing board on this one.  Thank goodness we have good tests to catch this kind of problem before it goes in.

jonbob:
Thanks @xylar -- I also ran a more simple ERS G-case and it failed as well. I'll revert this and hope it's not too onerous to track down

xylar:
Thanks @jonbob.

xylar:
I think it makes sense to close this and have a different branch for an alternative approach.

PR #5277, opened 2022-11-07
Add perlmutter machines to list that perform syslog commands for saving provenance

ndkeen:
Add new NERSC machines to list that perform syslog commands for saving provenance.

This might be what is preventing jobs to be auto-included to PACE.

[bfb]

### DIFF
--- a/cime_config/customize/provenance.py
+++ b/cime_config/customize/provenance.py
@@ -763,7 +763,7 @@ def _get_batch_job_id_for_syslog(case):
     """
     mach = case.get_value("MACH")
     try:
-        if mach in ["anvil", "chrysalis", "compy", "cori-haswell", "cori-knl"]:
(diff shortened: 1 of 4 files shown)


ndkeen:
When I tested this on pm-cpu, I now see a "checkpoints" directory written to the provenance directory, which does not look like was being written with cases previously.

For example:
`/global/cfs/cdirs/e3sm/performance_archive/ndk/v2LR.F2010-CICE.ne30pg2_EC30to60E2r2.me06-oct25.gnu.5d.n008a128x1c8.provtest/3605024.221107-105903/checkpoints.3605024.221107-105903`

ndkeen:
@sarats noted that this change will likely not fix the issue of automatic data upload to PACE, but it is probably still something we want to do (can you approve?)

sarats (approved):

ndkeen:
Unfortunately ran into problems trying to merge this to next, and the only way I know to fix is redo the PR, which I did here https://github.com/E3SM-Project/E3SM/pull/5308

PR #5282, opened 2022-11-09
Adds a 1D forcing testcase with BGC (ISPOL)

njeffery:
Atmospheric, ocean and biogeochemical data from the 2004 Ice Station Polarstern (ISPOL) Weddell Sea (67.9S, 54W) experiment from June 16, 2004 to December 31, 2004. Core 2 from the ISPOL location fills out the remainder of the year.

A uniform_1D initial condition was added for flexible initialization. A new testcase - single_cell_ispol - was added to
MPAS-Dev/MPAS-Seaice_standalone_framework to verify this forcing and initializtion. https://github.com/MPAS-Dev/MPAS-Seaice_standalone_framework/pull/15

BFB

### DIFF
--- a/components/mpas-seaice/src/Registry.xml
+++ b/components/mpas-seaice/src/Registry.xml
@@ -450,7 +450,7 @@
 		/>
 		<nml_option name="config_initial_condition_type" type="character" default_value="cice_default" units="unitless"
 			description="Initial condition type for sea ice state."
-			possible_values="'restart', 'uniform', 'circle', 'square' or 'uniform_interior'"
+			possible_values="'restart', 'uniform', 'circle', 'square', 'uniform_interior', or uniform_1D"
 		/>
 		<nml_option name="config_initial_ice_area" type="real" default_value="1.0" units="unitless"
 			description="Sea ice concentration at initialization."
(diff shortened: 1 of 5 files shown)


rljacob:
@akturner and @eclare108213 please review.

eclare108213:
Darin and I spent several hours yesterday trying to get me up and running on anvil, and failed.  Adrian's testing setup and even E3SM itself fail to run (E3SM segfaults).  So please don't hold up merging PRs, waiting on reviews from me. I'll keep trying, but E3SM has been failing for me ever since it was called ACME.

components/mpas-seaice/src/Registry.xml
```diff
@@ -3737,6 +3785,52 @@
...
+		     units="W m-2"
+		     description="Incoming shortwave flux (positive down)"
+		/>
+		<var name="longwaveDown1D"
+		     type="real"
+		     dimensions="Time"
+		     units="W m-2"
+		     description="Incoming longwave flux (positive down)"
+		/>
+		<var name="airTemperature1D"
```
akturner, line 3802: I think this implementation could be greatly simplified if the input forcing files included a nCells=1 dimension so that an airTemperature1D variable etc in addition to the airTemperature variable would not be needed

components/mpas-seaice/src/shared/mpas_seaice_initialize.F
```diff
@@ -558,6 +564,124 @@ subroutine init_ice_state_uniform_ice(&
 
   end subroutine init_ice_state_uniform_ice!}}}
 
+!|||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
+!
+!  init_ice_state_uniform_1D
```
akturner, line 569: Do thermodynamic tracer values not need to be set by this testcase with colpkg_init_trcr (e.g. https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-seaice/src/shared/mpas_seaice_initialize.F#L705?) It seems from the IPSOL namelist file that vertical thermodynamics is active so would need the tracers initialized?

akturner (changes requested):

components/mpas-seaice/src/shared/mpas_seaice_initialize.F
njeffery: @akturner :  Yes, thanks.  I added the active tracer initializations and re-tested.

components/mpas-seaice/src/Registry.xml
njeffery: @akturner : Agreed.  I chose to make the forcing data grid independent though so I could test on an arbitrary grid.  It runs with the QU240, for example, at least in short tests.

rljacob:
@eclare108213 please review.

eclare108213:
The conflict needs to be fixed before I can compile this.  @njeffery what cases did you run that were BFB?  Is there a particular test (or test suite) that would be most appropriate for me to run, to make sure that something hasn't inadvertently been broken in E3SM?

njeffery:
@eclare108213 : I ran a short 10-day with restart GCASE on chrysalis with mpassi defaults and compared mpassi restarts with master.  These were BFB.  Simulation scripts for these runs are here /home/ac.jeffery/SimulationScripts/archive/v2/ with run.ISPOLTestGCASE.sh  using the feature branch and run.ISPOLTestGCASE.Master.sh pre-feature.

components/mpas-seaice/src/Registry.xml
akturner: @njeffery If the forcing is the same everywhere on the mesh, what's the utility of being able to run on meshes other than the single cell one?

njeffery: @akturner : your point is well taken.  I'll restructure the test case with a cell index in the forcing files. I'd like to keep the bgc nutrient file grid independent so that I can use it more generally, i.e. in a standard GCASE but with ocean bgc off.  Does that seem reasonable?

rljacob:
@akturner have your comments been addressed?

njeffery:
@rljacob :  I still have some changes to make in response to @akturner's comments.

rljacob:
@njeffery how are those changes going?

njeffery:
I've opened a new PR  ( https://github.com/E3SM-Project/E3SM/pull/5506 ) to address the issues raised by @akturner .  This one can be closed.

PR #5285, opened 2022-11-10
Corrects treatment of enthalpy for thin ice

njeffery:
To be tested as a possible fix to #5124 - mpas-si crashes on small timesteps bug fix.
nonBFB

### DIFF
--- a/components/mpas-seaice/src/column/ice_therm_vertical.F90
+++ b/components/mpas-seaice/src/column/ice_therm_vertical.F90
@@ -439,10 +439,10 @@
       ! evapn < 0 => sublimation, evapn > 0 => condensation
       ! aerosol flux is accounted for in ice_aerosol.F90
       !-----------------------------------------------------------------
-            
+
       dhi = hin - worki
       dhs = hsn - works - hsn_new
-      
+
       freshn = freshn + evapn - (rhoi*dhi + rhos*dhs) / dt
       fsaltn = fsaltn - rhoi*dhi*ice_ref_salinity*p001/dt
       fhocnn = fhocnn + fadvocn ! for ktherm=2 
@@ -1038,7 +1038,7 @@
                                     meltt,     melts,    &
                                     meltsliq,  frain,    &
                                     meltb,     iage,     &
-                                    congel,    snoice,   &  
+                                    congel,    snoice,   &
                                     mlt_onset, frz_onset,&
                                     zSin,      sss,      &
                                     dsnow,     tr_snow,  &
@@ -1587,7 +1587,7 @@
             endif
          enddo
        endif
-     
+
 !---!-------------------------------------------------------------------
 !---! Repartition the ice and snow into equal-thickness layers,
 !---! conserving energy.
@@ -1596,7 +1596,7 @@
       !-----------------------------------------------------------------
       ! Compute desired layer thicknesses.
       !-----------------------------------------------------------------
- 
+
       if (hin > c0) then
          hilyr = hin / real(nilyr,kind=dbl_kind)
       else
@@ -1631,23 +1631,23 @@
         !-----------------------------------------------------------------
         ! Conserving energy, compute the enthalpy of the new equal layers.
         !-----------------------------------------------------------------
-            
+
          call adjust_enthalpy (nilyr,              &
                                zi1,      zi2,      &
                                hilyr,    hin,      &
-                               zqin)   
+                               zqin)
 
          if (ktherm == 2) &
               call adjust_enthalpy (nilyr,              &
                                     zi1,      zi2,      &
                                     hilyr,    hin,      &
-                                    zSin)   
+                                    zSin)
 
       else ! zero layer (nilyr=1)
 
          zqin(1) = -rhoi * Lfresh
          zqsn(1) = -rhos * Lfresh
-       
+
       endif
 
       if (nslyr > 1) then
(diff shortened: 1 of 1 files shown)


njeffery:
@mkstratos : Could you test to see if this PR fixes issue #5124 ?

njeffery:
@jonbob : In this PR I changed a puny conditional to division if greater than zero. Could you verify that I haven't introduced possible Inf or NaN? Thanks

jonbob:
@njeffery -- I also have a simplified test to see if it fixes the problem @mkstratos first reported. I can run that as well, or happy to save you the step @mkstratos ?

njeffery:
I tested this PR with a 25 year fully coupled run (branched from the v2 pre-industrial spin-up) and ran MPAS-analysis against a control (v2control) -- https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jeffery/E3SMv2/thin-ice-enthalpy-error/

I also looked at energy conservation in MPAS-seaice timeseries.

Feature run (thin-ice_enthalpy-error) relative energy error:


thin-ice-enthalpy-error relativeEnergyError_conservation_check_1-25

Control (v2control) relative energy error:

v2control relativeEnergyError_conservation_check_1-25

Feature - Control:

thin-ice-enthalpy-error relativeEnergyError_conservation_check_1-25

njeffery:
The figures above are not correct.  I'll try again...

Feature-Control (relative energy error)

image

jonbob:
@njeffery -- your fix looks very consistent with the discussion in [PR #5252](https://github.com/E3SM-Project/E3SM/pull/5252). I'll test it in my simplified setup

jonbob:
Using this fix did pass my simplified version of the original test failure reported by @mkstratos

njeffery:
Thanks @jonbob! sounds like I can go ahead and assign reviewers.

akturner:
@eclare108213: The current implementation of adjust_enthalpy doesn't conserve the quantity being remapped e.g. energy when enthalpy is remapped. This was causing a non-conservation error for the atmos. folks when running the ice in prescribed ice mode and small time steps. It seems there are two classes of fix. Set the thickness limit in adjust_enthalpy to zero, or somewhere account for the ice being zapped in adjust_enthalpy with a correction to fhocn etc either in adjust_enthalpy or not.

akturner:
@eclare108213, @njeffery: It seems removing completely thin (<puny) ice and snow layers here (https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-seaice/src/column/ice_therm_vertical.F90#L1590) would fix things.

njeffery:
@eclare108213 : I'm not sure how to proceed with this.  Would you like to see a longer run?  Should I withdraw the PR?

eclare108213:
@njeffery I do think this needs to be fixed, so please don't withdraw it! 

@akturner I think the underlying code must have changed, because pointing to this line in your comment above doesn't make sense in this conversation.   (https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-seaice/src/column/ice_therm_vertical.F90#L1590)  Which line were you looking at?

Alternatively, for the adjust_enthalpy conservation issue, would it work to simply put a conditional around the last loop that computes the new enthalpies in the layers, so that they keep the original values when the layer thickness is very thin? qn is inout.

> if (rhlyr > puny) then ! add this new conditional
>      do k = 1, nlyr
>         qn(k) = hq(k) * rhlyr
>      enddo
> endif

njeffery:
@eclare108213 :  That formulation wouldn't be conserving either.

eclare108213:
Here's a suggestion.  I think the puny conditional needs to remain in order to avoid overflows, but there could be a new block of code for very thin ice that spreads the enthalpy uniformly among the 4 layers rather than doing the calculation to apportion it between layers.  Ice that thin ought to be isothermal anyhow.  Would that work?

rljacob:
@njeffery please finish this PR.

njeffery:
@eclare108213 : I tried spreading enthalpy among the layers for thin ice but still ran into the issue that I needed to divide by the layer thickness.  It also crashed on me with a thermo conservation error.

eclare108213:
Are you still using the puny conditional?  Is the total ice thickness greater than puny?  If the problem is that the individual layers are less than puny, you could try increasing the conditional to nilyr*puny.  This shouldn't really matter for such extremely thin ice.  I don't understand why there would be a thermo conservation error if the total enthalpy is conservatively spread across the same volume of ice in this routine.  If the ice is so thin, it ought to be zapped, so this just needs to get you through the rest of the time step, when either it's zapped or completely melted in the thermo.

eclare108213:
> I tried spreading enthalpy among the layers for thin ice but still ran into the issue that I needed to divide by the layer thickness. It also crashed on me with a thermo conservation error.

If you are spreading it evenly among the layers, then you should only need to divide by the number of layers, and wouldn't need the actual thickness.  The thermo error might be a different problem, that this case happens to be dredging up (e.g. CFL restrictions). 

I think I'm not communicating very well what I have in mind.  Could we talk about this for a few minutes?

rljacob:
status:  still under discussion.
@njeffery maybe close this if you're not close?

njeffery:
Closing for now while I test @eclare108213's suggestion. I plan on re-opening once non-BFB changes have been discussed by the ice group and deemed acceptable.

PR #5286, opened 2022-11-10
Use double precision for constant

tcclevenger:
This is needed for the upcomming pgrad_correction port to theta-l_kokkos. Basically, when adding the constant `tref_lapse_rate`, there were bfb issues since the F90 version was not using double precision.

This would be a non-bfb change, so if that is not ok, I could make the c++ const single precision. Let me know.

### DIFF
--- a/components/homme/src/theta-l/share/element_ops.F90
+++ b/components/homme/src/theta-l/share/element_ops.F90
@@ -75,7 +75,7 @@ module element_ops
   public state0
 
   ! promote this to _real_kind after V2 code freeze
-  real (kind=real_kind), public :: tref_lapse_rate=0.0065e0
+  real (kind=real_kind), public :: tref_lapse_rate=0.0065D0
 contains
 
 


oksanaguba:
Looks ok, but i am puzzled why it is needed. My understanding was that you would use the init of ref. states from F for both codes. If so, then the init is only done once, in F, and this change is not needed?

tcclevenger:
@oksanaguba I may be confused about how the F90 and C++ code interact, but I thought that all constants needed in theta-l_kokkos needed to be put into `PhysicalConstants.hpp`.

tcclevenger:
To clarify, this is needed because, in a PR I'll create after this, I'll need to add `tref_lapse_rate` to `PhysicsConstants.hpp`. I wanted to separate into 2 PRs so that the next one can be label BFB.

rljacob:
We are about to make a 2.1 tag so no non-BFB PRs can go in for about a week.

ambrad:
I suggest using single precision in the C++ to avoid a non-BFB change to E3SM.

tcclevenger:
Ok, I'll close this one and go single prec for now. We can revisit later I we want.

mt5555:
Note that this change to the Fortran code would be BFB for all E3SM tests, since it's only used by a feature (pgrad_correction=1) that is not turned on by default (although it should be :-)

The precision of this constant doens't matter, so do whatever is easiest to maintain C++/fortran agreement.

ambrad:
@mt5555 I haven't run tests with the change, so I don't know whether the precision change would affect tests. But the constant is used in set_theta_ref as well as the pgrad_correction code. Is set_theta_ref used in E3SM v2 configs?

mt5555:
So I believe that all of the E3SM v2 cases would remain BFB - the only diffs should be in the NH test cases (since they use phi_ref, which is derived from theta_ref)

rljacob:
Is setting pgrad_correction=1 something the v3 atm should do?

tcclevenger:
FYI: new PR https://github.com/E3SM-Project/E3SM/pull/5299. I actually chose to change this const to double precision for BFB testing.

Issue #5260, opened 2022-10-28
Test hanging with nvidia compiler on pm-cpu ne30pg2_r05_EC30to60E2r2.GPMPAS-JRA


ndkeen:
This test is hanging and running out of time on pm-cpu with nvidia compiler.
Happens with OPT, but NOT DEBUG.

```
ERS.ne30pg2_r05_EC30to60E2r2.GPMPAS-JRA.pm-cpu_nvidia.mosart-rof_ocn_2way
```

Debug test will run, but is really slow (completed a few days -- have not let it run far enough to complete)
```
ERS_D.ne30pg2_r05_EC30to60E2r2.GPMPAS-JRA.pm-cpu_nvidia.mosart-rof_ocn_2way
```

whannah1:
I'm not sure we need a separate issue for this since it's most likely the same hanging issue that is affecting many other configurations on both pm-cpu and pm-gpu.

ndkeen:
Well the other hangs I'm aware of are happening while writing land-related files. And this issue does not seem to be stopping there.

Also, I now see that single threaded SMS test does pass:
`SMS_PMx1.ne30pg2_r05_EC30to60E2r2.GPMPAS-JRA.pm-cpu_nvidia.mosart-rof_ocn_2way`

whannah1:
This compset seems to have an active land, so I would speculate that it's still hanging during the land file write, but obviously that's hard to verify right now. 
```
  <compset>
    <alias>GPMPAS-JRA</alias>
    <lname>2000_DATM%JRA_ELM%SPBC_MPASSI_MPASO%DATMFORCED_MOSART_SGLC_SWAV</lname>
  </compset>
```
Can you check what the latest modified files were in the run directory?

ndkeen:
Note this test passes with GNU.  I think the other hangs you are thinking of use GNU compiler.

```
-rw-rw-r--  1 ndk ndk 23150 Oct 28 08:23 atm.log.3511778.221028-082301
-rw-rw-r--  1 ndk ndk  9184 Oct 28 08:23 rof.log.3511778.221028-082301
-rw-rw-r--  1 ndk ndk 70974 Oct 28 08:23 ocn.log.3511778.221028-082301
drwxrwxr-x  4 ndk ndk  4096 Oct 28 08:23 ./
[...]

PR #5275, opened 2022-11-04
For Perlmutter add env vars to avoid hanging under certain situations

ndkeen:
For Perlmutter (pm-cpu, pm-gpu, and test machine alvarez), add some libfabric environment variables that were suggested by NERSC/HPE to address issues such as jobs hanging. Tested with both pm-cpu and pm-gpu.
Appears to be no noticeable difference in performance. Possibly temporary as newer version of libfabric may correct.

```
setenv FI_CXI_DEFAULT_CQ_SIZE 71680
setenv FI_CXI_REQ_BUF_SIZE 12582912
setenv FI_UNIVERSE_SIZE 4096
```

Adding a little more info about the env vars, which I found via `man fi_cxi`:

```
FI_CXI_DEFAULT_CQ_SIZE
  Change the provider default completion queue size.  This may be useful for applications which
  rely on middleware, and middleware defaults the completion queue size to the provider default.

FI_CXI_REQ_BUF_SIZE
  Size of request buffers.  Increasing the request buffer size allows for more unmatched messages
  to be sent into a single request buffer.  The default size is 2MB.

FI_UNIVERSE_SIZE
  Defines the maximum number of processes that will be used by distribute OFI application.

```
Note I think, the `FI_CXI_REQ_BUF_SIZE` variable (and maybe the other?) may only be used when using `FI_CXI_RX_MATCH_MODE=software` which we are already doing for `pm-cpu`, but NOT `pm-gpu`. So it's
possible these are not needed for `pm-gpu` and I will test again there.

Will fix issues in SCREAM repo, so will need upstream merge.
Fixes https://github.com/E3SM-Project/scream/issues/1920
Fixes https://github.com/E3SM-Project/scream/issues/1939

[bfb]

PR #5291, opened 2022-11-10
For pm-cpu and pm-gpu, add environment variable MPICH_COLL_SYNC=MPI_Bcast

ndkeen:
For pm-cpu and pm-gpu, add env variable MPICH_COLL_SYNC=MPI_Bcast
to avoid hanging during writing of certain files (esp ELM restarts). Thanks to @whannah1 and @dqwu for
testing, understanding the issue better, and finding this solution.

There appears to be some issues with the slingshot network as we don't see this 
problem on other machines.

Setting this will perform a MPI_Barrier before each MPI_Bcast and we've
found this resolves hanging and does not noticeably degrade performance.

Once updates are made to the network on PM, we can re-evaluate if this is still needed.

We already had a PR started with one solution to this problem here https://github.com/E3SM-Project/E3SM/pull/5275
but this solution is favored.

Fixes issues that were created in the SCREAM repo
Fixes https://github.com/E3SM-Project/scream/issues/1920
Fixes https://github.com/E3SM-Project/scream/issues/1939
Was also hanging with MMF case on 128 pm-gpu nodes.

Also noting https://github.com/E3SM-Project/scorpio/pull/493 which further explains the problem.

[BFB]

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -258,6 +258,7 @@
       <env name="HDF5_USE_FILE_LOCKING">FALSE</env>
       <env name="PERL5LIB">/global/cfs/cdirs/e3sm/perl/lib/perl5-only-switch</env>
       <env name="FI_CXI_RX_MATCH_MODE">software</env>
+      <env name="MPICH_COLL_SYNC">MPI_Bcast</env>
     </environment_variables>
     <resource_limits>
       <resource name="RLIMIT_STACK">-1</resource>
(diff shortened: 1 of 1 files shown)


ndkeen:
I also tested with `e3sm_developer` on pm-cpu.

whannah1:
@ndkeen should we add this change for the "alvarez" machine? Or is that defunct?

whannah1 (approved):

ndkeen:
I considered adding to `alvarez` as well, but was keeping PR as simple as possible. I've only been allowed on the machine a few weeks total, and it's not clear if I will be able to test with it in the future. So it's not dead yet, but leaving it in there for now.

ndkeen:
merged to next

ndkeen:
The e3sm_integration tests have run twice now on pm-cpu. There are 2 fails that were there before and a fail that appears intermittent (https://github.com/E3SM-Project/E3SM/issues/5298). I will let it try one more time, but I think it's OK to merge to master.

ndkeen:
I'm going to go ahead and merge to master since I'm confident the failing test in not related to this.

ndkeen:
Also Fixes https://github.com/E3SM-Project/E3SM/issues/5260

PR #5304, opened 2022-11-15
Add support for JRA55 v1.5 forcing

darincomeau:
Adds support for [JRA55-do v1.5](https://climate.mri-jma.go.jp/pub/ocean/JRA55-do/) datm/drof forcing through years 1958 - 2020 with a new compset `GMPAS-JRA1p5`.

Forcing files have been downloaded from the [ccsm-inputdata server](https://svn-ccsm-inputdata.cgd.ucar.edu/trunk/inputdata/ocn/jra55/v1.5_noleap) to the E3SM public inputdata server [here](https://web.lcrc.anl.gov/public/e3sm/inputdata/ocn/jra55/v1.5_noleap/).

A pair of runs, `GMPAS-JRA1p4.TL319_ECwISC30to60E2r1` and `GMPAS-JRA1p5.TL319_ECwISC30to60E2r1`, were run for 60 years for comparison. Comparison analysis of years 41-60: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/JRA1p5-testing/JRA1p4_vs_1p5/yrs41-60/

### DIFF
--- a/components/data_comps/drof/cime_config/namelist_definition_drof.xml
+++ b/components/data_comps/drof/cime_config/namelist_definition_drof.xml
@@ -63,6 +63,7 @@
       <value drof_mode="IAF_JRA_1p4_2018_AIS0LIQ">rof.iaf_jra_1p4_2018_ais0liq</value>
       <value drof_mode="IAF_JRA_1p4_2018_AIS0ROF">rof.iaf_jra_1p4_2018_ais0rof</value>
       <value drof_mode="IAF_JRA_1p4_2018"     >rof.iaf_jra_1p4_2018</value>
+      <value drof_mode="IAF_JRA_1p5"          >rof.iaf_jra_1p5</value>
       <value drof_mode="IAF_JRA"              >rof.iaf_jra</value>
       <value drof_mode="RYF8485_JRA"          >rof.ryf8485_jra</value>
       <value drof_mode="RYF9091_JRA"          >rof.ryf9091_jra</value>
@@ -85,6 +86,7 @@
       <value stream="rof.diatren_iaf_ais00_rx1">$DIN_LOC_ROOT/lnd/dlnd7/RX1</value>
       <value stream="rof.diatren_iaf_ais45_rx1">$DIN_LOC_ROOT/lnd/dlnd7/RX1</value>
       <value stream="rof.diatren_iaf_ais55_rx1">$DIN_LOC_ROOT/lnd/dlnd7/RX1</value>
+      <value stream="rof.iaf_jra_1p5"          >$DIN_LOC_ROOT/lnd/dlnd7/JRA55</value>
       <value stream="rof.iaf_jra.*"            >$DIN_LOC_ROOT/lnd/dlnd7/JRA55</value>
       <value stream="rof.ryf*"                 >$DIN_LOC_ROOT/lnd/dlnd7/JRA55</value>
       <value stream="rof.cplhist"              >null</value>
@@ -177,6 +179,7 @@
       <value stream="rof.diatren_iaf_ais00_rx1">$DIN_LOC_ROOT/lnd/dlnd7/RX1</value>
       <value stream="rof.diatren_iaf_ais45_rx1">$DIN_LOC_ROOT/lnd/dlnd7/RX1</value>
       <value stream="rof.diatren_iaf_ais55_rx1">$DIN_LOC_ROOT/lnd/dlnd7/RX1</value>
+      <value stream="rof.iaf_jra_1p5"          >$DIN_LOC_ROOT/ocn/jra55/v1.5_noleap</value>
       <value stream="rof.iaf_jra.*"            >$DIN_LOC_ROOT/lnd/dlnd7/JRA55</value>
       <value stream="rof.ryf*"                 >$DIN_LOC_ROOT/lnd/dlnd7/JRA55</value>
       <value stream="rof.cplhist"              >$DROF_CPLHIST_DIR</value>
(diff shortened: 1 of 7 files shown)


xylar (approved):
I'm not familiar enough to review the code changes but the results look reasonable to me.  @darincomeau, do you know if the relatively larger changes (particularly in the Arctic) are expected?  I know you said the full reanalysis had changed.

milenaveneziani:
Thanks @darincomeau for putting this together. My sense is that the changes must be due to surface freshwater fluxes, since they mainly involve salinity and MLD at high-latitudes. So I wonder if the v1.5 version is slightly different from v1.4 even over the overlapping years.
It's too bad that the central Arctic salinity goes in the wrong direction (saltier in v1.5), since we tend to overestimate it in the Central Arctic. But in other polar regions it will be better.

darincomeau:
>do you know if the relatively larger changes (particularly in the Arctic) are expected?

The only information I have on v1.5 is what's on the JRA website https://climate.mri-jma.go.jp/pub/ocean/JRA55-do/; it looks like it includes some minor corrections to some atm fields, but that in OMIP style experiments with another model, the impact was minor.

components/data_comps/drof/cime_config/namelist_definition_drof.xml
```diff
@@ -177,6 +179,7 @@
       <value stream="rof.diatren_iaf_ais00_rx1">$DIN_LOC_ROOT/lnd/dlnd7/RX1</value>
       <value stream="rof.diatren_iaf_ais45_rx1">$DIN_LOC_ROOT/lnd/dlnd7/RX1</value>
       <value stream="rof.diatren_iaf_ais55_rx1">$DIN_LOC_ROOT/lnd/dlnd7/RX1</value>
+      <value stream="rof.iaf_jra_1p5"          >$DIN_LOC_ROOT/ocn/jra55/v1.5_noleap</value>
```
jonbob, line 182: @darincomeau - any reason not to put these files in the lnd/dlnd7 directory?

darincomeau: yes - these files were all together in one directory on the ccsm svn inputdata server https://svn-ccsm-inputdata.cgd.ucar.edu/trunk/inputdata/ocn/jra55/v1.5_noleap/, and I set this path so it would automatically download them. I meant to make a point here that the directory structure is different than for previous versions.

I'd be happy to move the `drof` files though so they sit in a directory next to the v1.4 counterparts. One other consideration is that these files may (hopefully) be updated as new years become available, so I thought to have our directory structure match the NCAR one.

vanroekel:
@darincomeau, @xylar @milenaveneziani would you like me to reach out the CLIVAR ocean model panel that I'm also on to check about the response to this update to JRA to see if our model is consistent?  Happy to do so

components/data_comps/drof/cime_config/namelist_definition_drof.xml
jonbob: Oooo, matching NCAR is a dangerous rationale! I guess it seems more natural for these files to be co-located with the streams domain file, which at this point is in dlnd. But it might be worth a conversation at some point if datm files should be stored with the atm and drof files with the rof, and not in either the ocn or lnd?

darincomeau: Sure, happy to discuss re-organizing this. Thanks again for pointing out, I meant to flag this in a comment.

jonbob: Ah, any re-organizing should probably be done in the future. I think as long as you document the reasons for the files being where they are this PR is fine as-is

xylar:
@darincomeau, I took a second look and I'm not sure what I saw the first time in the Arctic.  At a second look, the only thing I see there is that things are consistently a bit saltier at the surface in 1.5 but that isn't a strong enough effect to register in the T-S diagrams, which look essentially identical to me.  Still, I believe this is a change in the right direction.  Anyway, sorry for the confusion there!

Where things *are* different is on the Antarctic continental shelf.  There, salinities are consistently higher in 1.5 than 1.4, and that is clearly a chance that is reducing biases.  The T/S diagrams all look better compared to WOA in 1.5 than 1.4.  This must be due to higher sea-ice production, which is also quite apparent (yay new analysis!).  Interestingly, there's not a clear indication of stronger winds in sea surface speed but that presumably wouldn't be expected in sea-ice covered regions.

@vanroekel, I wouldn't think it would hurt to reach out to CLIVAR but from what I see, there's nothing here to indicate that @darincomeau has done anything wrong.  The changes all seem consistent with improvements in the forcing.

milenaveneziani (approved):

xylar:
@milenaveneziani, thanks for clarifying.  I didn't look at the surface properties compared with obs there and you're certainly more familiar with our biases in the Arctic than I am.

mark-petersen (approved):
I looked at the comparison and everything looks reasonable. The trouble with climate validation is it's hard to say what is "better". But JRA55 v1.5 is certainly working. Thanks for your work on this @darincomeau!

darincomeau:
Thanks all for the reviews and comments. @vanroekel and @maltrud I think we have enough here to go ahead with the merge, but any further comments or welcome, or feel free to remove yourselves as reviewers. In part I wanted to make it more widely known this capability was incoming.

vanroekel (approved):
Approving on visual inspection and testing.

vanroekel:
@darincomeau sorry for the delay on looking at this.  I'm happy to approve.  I agree with @xylar that any biases or diffs we see are certainly forcing caused and nothing in your PR.  But I will follow up with OMDP folks for intercomparison purposes.

jonbob:
passes:
* SMS.TL319_EC30to60E2r2.GMPAS-JRA1p5.chrysalis_intel (using new compset)

merged to next

jonbob:
merged to master

PR #5308, opened 2022-11-15
Add perlmutter machines to list that perform syslog commands for saving provenance

ndkeen:
Add new NERSC machines (suach as pm-cpu/pm-gpu) to list of machine that save additional system logs to provenance

This is a redo of PR #5277

[bfb]


ndkeen:
merged to next

PR #5309, opened 2022-11-15
Update MALI version

trhille:
This merge includes numerous updates to MALI.  

The major updates include:
• A glacier retreat parameterization based on change in ocean thermal forcing and subglacial runoff.
• Coupling to a 1-D global sea level model.
• A CFL-like limit on time step length based on calving velocity.
• A new definition of the grounding line that depends on the flotation condition, which no longer includes the grounded marine margin if that is above the flotation thickness. This is necessary for mass budget calculations.

Minor changes include:
• The ability to use rate-based and damage threshold-based calving parameterizations simultaneously.
• Spatially variable calving stress thresholds for the von Mises stress calving parameterization.
• An option to prevent positive surface mass balance in ice-free regions.
• An optional clean-up for rate-based iceberg calving routines.
• New fields containing the basal mass balance that was actually applied in a time step.
• Logic that prevents the original input basal mass balance field from ever being modified in case of an instability in the thermal solver.
• Clean up for compiling with Intel, including memory bug fix in Sea Level Model.
• Bug fixes necessary for regional statistics output.
• A step to remove isolated regions of dynamic floating ice connected to the rest of the ice shelf by non-dynamic ice. These were causing the velocity solver to fail in simulations with evolving calving.
• Additional new calving metrics and minor bug fixes.
• A new field that defined the flux across the grounding line on cell centers.
• Use indexToCellID(iCell) instead of iCell in log messages.
• Write out additional fields to ascii files to import into Albany.
• Bug fixes in MALI–Sea Level Model coupling: initialize variables.

[NML]
[non-BFB] for configurations with MALI


rljacob:
A related question: should E3SM testing test more of MALI?  We only test in the shallow-ice approximation mode.

trhille:
@rljacob, we are working towards a coupled Greenland Ice Sheet configuration, and once that is ready we will introduce more tests that cover its behavior. That is still likely at least a few months away, however.

jonbob:
@trhille - I ran the scripts that semi-automatically update the E3SM bld files to match the changes in Registry. I don't believe I can push them to your fork, so either I can point you at the changes and you can commit them or you can invite me to collaborate on your fork

trhille:
@jonbob, I just added you as a collaborator. Let me know if you don't see the invite.

jonbob:
Thanks @trhille -- I pushed my changes, and I'll test in E3SM now

jonbob:
test merge passes:

with expected NML DIFFs

components/mpas-albany-landice/bld/namelist_files/namelist_defaults_mali.xml
```diff
@@ -49,8 +55,11 @@
 <config_damage_stiffness_min>0.1</config_damage_stiffness_min>
 <config_damage_rheology_coupling>.false.</config_damage_rheology_coupling>
 <config_damage_gl_setting>'nye'</config_damage_gl_setting>
-<config_damage_calving_method>'threshold'</config_damage_calving_method>
+<config_damage_calving_method>'none'</config_damage_calving_method>
```
jonbob, line 58: @trhille - I just want to make sure this is the correct new default. The automated scripts to reflect Registry changes in the bld files tend to default to the previous E3SM settings but this one appears to be set as part of this PR

jonbob (approved):
Approved based on testing inside E3SM -- though I have to admit I don't feel capable of reviewing landice code changes otherwise and will leave the science review to others

components/mpas-albany-landice/bld/namelist_files/namelist_defaults_mali.xml
trhille: @jonbob, yes this is the correct new default. Thanks for checking!

matthewhoffman (approved):
I looked over both the set of changes being brought over from MALI-Dev and the namelist modifications that @jonbob made, and it all looks as expected.

jonbob:
Tested a new merge, still passes:
with expected NML DIFFs

Also passes:
* SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.chrysalis_intel

merged to next

jonbob:
Testing in next has shown no issues except compy_pgi, where all tests fail to build with an internal compiler error:
````
PGF90-S-0000-Internal compiler error. items were added to sem.p_dealloc but not freed       0
````
With some tips from @ambrad, @oksanaguba and @rljacob we could quickly identify the snippet of code that makes the pgi compiler unhappy, and have a fix in place. However, testing of the fix that allows the tests to build now shows a runtime error:
````
0: DEALLOCATE: memory at (nil) not allocated
````
with absolutely no traceback information, even in debug mode. By adding some diagnostics to mali (a fancy name for print statements) I did end up identifying another code snippet that allows the tests to successfully complete when it's removed. In the end, that section of code was not the problem but the one before it -- the section that I commented out simply used the results of the previous section. In any case, there are two chunks of code that must be replaced to make the pgi complier happy, even though they're both perfectly fine Fortran. I'm working with the developer on that fix and hope to get it merged in soon.

jonbob:
Re-merged to next to fix pgi compiler issues

jonbob:

Issue #5297, opened 2022-11-12
FP exception after adding `-ffpe-trap=invalid` with `ERS_D_Ln9.ne4pg2_ne4pg2.FRCE-MMF1.pm-cpu_gnu.eam-cosp_nhtfrq9`


ndkeen:
With the test `ERS_D_Ln9.ne4pg2_ne4pg2.FRCE-MMF1.pm-cpu_gnu.eam-cosp_nhtfrq9`, after adding `-ffpe-trap=invalid` to the fortran builds, I see this error:

```
32: Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.
32:
32: Backtrace for this error:
32: #0  0x151285177d6f in ???
32: #1  0x8a8cda in subsample_and_optics
32:     at /global/cfs/cdirs/e3sm/ndk/repos/ndk_machinefiles_pm-cpu-add-gnu-debug-invalid/components/eam/src/physics/crm/cospsimulator_intr.F90:2279
32: #2  0x8ffc0f in populate_cosp_subcol
32:     at /global/cfs/cdirs/e3sm/ndk/repos/ndk_machinefiles_pm-cpu-add-gnu-debug-invalid/components/eam/src/physics/crm/cospsimulator_intr.F90:1626
32: #3  0x902ca5 in __cospsimulator_intr_MOD_cospsimulator_intr_run
32:     at /global/cfs/cdirs/e3sm/ndk/repos/ndk_machinefiles_pm-cpu-add-gnu-debug-invalid/components/eam/src/physics/crm/cospsimulator_intr.F90:1092
32: #4  0x9a023d in __radiation_MOD_radiation_tend
32:     at /global/cfs/cdirs/e3sm/ndk/repos/ndk_machinefiles_pm-cpu-add-gnu-debug-invalid/components/eam/src/physics/crm/rrtmgp/radiation.F90:1902
32: #5  0x910072 in tphysbc2
32:     at /global/cfs/cdirs/e3sm/ndk/repos/ndk_machinefiles_pm-cpu-add-gnu-debug-invalid/components/eam/src/physics/crm/physpkg.F90:1857
32: #6  0x933c09 in __physpkg_MOD_phys_run1._omp_fn.1
32:     at /global/cfs/cdirs/e3sm/ndk/repos/ndk_machinefiles_pm-cpu-add-gnu-debug-invalid/components/eam/src/physics/crm/physpkg.F90:866
[...]

PR #5310, opened 2022-11-16
fix uninitialized values in MMF cosp interface

whannah1:
This addresses what I think is behind issue #5297. We repeatedly failed to reproduce the issue and we're not sure why. Initializing these variables should help the problem when it surfaces, and is harmless otherwise.

[BFB]

### DIFF
--- a/components/eam/src/physics/crm/cospsimulator_intr.F90
+++ b/components/eam/src/physics/crm/cospsimulator_intr.F90
@@ -1024,7 +1024,7 @@ CONTAINS
   ! ######################################################################################
   ! SUBROUTINE cospsimulator_intr_run
   ! ######################################################################################
-  subroutine cospsimulator_intr_run(state,pbuf, cam_in,emis,coszrs,cld_swtau,snow_tau,snow_emis)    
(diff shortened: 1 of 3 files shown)


components/eam/src/physics/crm/cospsimulator_intr.F90
```diff
@@ -1590,6 +1590,13 @@ subroutine populate_cosp_subcol(npoints, state, pbuf, emis, cld_swtau, cospstate
            end do
          end do
        end do
+       if (present(snow_tau) .and. present(snow_emis)) then
+          dem_s_snow(1:npoints,1:pver)  = snow_emis(1:npoints,1:pver)  ! 10.5 micron grid-box mean optical depth of stratiform snow
```
brhillman, line 1594: I think this might actually cause more problems; snow_emis and snow_tau have size (pcols,pver), but npoints can actually be larger than pcols for MMF (ncol*crm_nx_rad*crm_ny_rad). I think maybe we should change the declaration of snow_emis and snow_tau to size npoints to allow this.

brhillman (approved):

brhillman:
Merged to master.

Issue #5284, opened 2022-11-10
Add ability to run maint-2.0 on pm-cpu


ndkeen:
Similar to https://github.com/E3SM-Project/E3SM/issues/5280 for maint-1.0, we would also like to use pm-cpu with the maint-2.0 branch.  In that branch, we did have machine `perlmutter`, but that was actually for what is now `pm-gpu` (with ability to also run on just the CPU's of those GPU nodes).  I will change this to be `pm-cpu` and maybe not even include `pm-gpu` for now.

ndkeen:
I am making a branch that removes any reference to machine `perlmutter` and replaces with `pm-cpu`.  I was going to remove any mention of `pm-gpu`.  Is that what we want?  I could also add `pm-gpu` to this branch.

rljacob:
Go ahead and add pm-gpu.

ndkeen:
OK I have branch `ndk/maint20/port-to-pm-cpu`

ndkeen:
This was completed by the above PR

PR #5313, opened 2022-11-17
Add ability to use pm-cpu/pm-gpu with maint-2.0

ndkeen:
Add ability to use pm-cpu/pm-gpu with maint-2.0
Rename machine perlmutter to be either pm-cpu (default) or pm-gpu.
Changes for alvarez machine also included to be consistent with master (alvarez is test machine for pm-cpu) 
In this PR, only updated PE layouts that were related to pm-cpu (or pm-gpu/alvarez) -- we may want to
update all of the PE layouts to be more consistent with current master.
Fixes https://github.com/E3SM-Project/E3SM/issues/5284
[bfb]


ndkeen:
Who best to review?

cime_config/allactive/config_pesall.xml
```diff
@@ -270,6 +270,125 @@
...
+          <ntasks_rof>-4</ntasks_rof>
+          <ntasks_ice>-4</ntasks_ice>
+          <ntasks_ocn>-4</ntasks_ocn>
+          <ntasks_glc>-1</ntasks_glc>
+          <ntasks_wav>-1</ntasks_wav>
+          <ntasks_cpl>-4</ntasks_cpl>
+        </ntasks>
+      </pes>
+    </mach>
+    <mach name="cori-haswell">
```
rljacob, line 291: Did you mean to bring in these layouts for other machines?

ndkeen: I was trying to only make updates affecting pm-cpu (and pm-gpu/alvarez), but I actually wasn't sure if it make more sense to bring in "all" new changes into files describing PE layouts as it's easier to maintain.  But then there may be some change that is required elsewhere...

rljacob: Just mention it in the PR description.

rljacob (approved):

ndkeen:
I just made a commit to this branch to add a PE layout for the RRM northamerica compset (runs on 8 nodes) and a slight tweak to another common compset in the `e3sm_prod` suite which is now running out-of-box for me on pm-cpu.

ndkeen:
Noting here that after the recent PM maintenance, this branch will now require same changes here:
https://github.com/E3SM-Project/E3SM/pull/5373
which I will do soon.

PR #5299, opened 2022-11-14
Add pgrad_correction to theta-l_kokkos

tcclevenger:
Add pgrad_correction to theta-l_kokkos. A few changes:
-- Change test using `thetahs1.nl` namelist to use pgrad_correction and test BFB
-- Add `cxx_log()` callable from F90 to compute log on device for BFB
-- Moved some computations around in CaarFunctorImpl to match F90 
ordering (some needed for BFB now that pgrad_correction exists)

[nonBFB] for HOMMEBFB only due to new test option for theta-fhs1.

PR #5315, opened 2022-11-17
Change tref_lapse_rate to double precision

tcclevenger:
Use double precision for tref_lapse_rate. Originally a part of https://github.com/E3SM-Project/E3SM/pull/5299. This is smaller PR to ensure fails there are not from other places.

[BFB] except for HOMME(BFB) standalone tests

### DIFF
--- a/components/homme/src/theta-l/share/element_ops.F90
+++ b/components/homme/src/theta-l/share/element_ops.F90
@@ -75,7 +75,7 @@ module element_ops
   public state0
 
   ! promote this to _real_kind after V2 code freeze
-  real (kind=real_kind), public :: tref_lapse_rate=0.0065e0
+  real (kind=real_kind), public :: tref_lapse_rate=0.0065D0
 contains
 
 


rljacob:
This is non-BFB for any case EAM?

ambrad:
@rljacob I added "[BFB] except for HOMME(BFB) standalone tests" to the PR description.

tcclevenger:
Thanks @ambrad. @oksanaguba you verified this is correct?

oksanaguba:
@tcclevenger there is one issue, per message in the previous PR, there is a unit test fail. Repeating here:
```
/gpfs/fs1/home/ac.onguba/acme-integrator/components/homme/test_execs/thetal_kokkos_ut/elem_ops_ut.cpp:154: FAILED:
  REQUIRE( h_theta_ref(ie,igp,jgp,ilev)[ivec] == theta_ref_f90(ie,k,igp,jgp) )
with expansion:
  4260.0426349057 == 4260.0427984401
```
This needs to be fixed in this PR. Unit tests do not have baselines. Many run xx code against F code using a random init, and both need to match.

oksanaguba:
Oh seeing a question above -- homme and hommebfb results were listed for a 1-line change in the prev PR, i did not explicitly run this branch, just my next clone with the change. There are nonbfb homme tests. For e3sm_developer, all pass.

oksanaguba:
in next

ambrad:
@tcclevenger @oksanaguba the failure of the unit test elem_ops_ut_test looks genuine and will remain on the dashboard. What is the resolution to this unit test failure?

tcclevenger:
Looks like the cxx code _does_ have a version on `tref_lapse_rate`. I function `compute_theta_ref()` in `theta-l_kokkos/cxx/ElementOps.hpp` it uses `0.0065f` to compute. Not sure the ramifications of changing this constant to double precision. I'll find if this is used anywhere else, and then create a new PR that changes the cxx.

tcclevenger:
So it looks like this constant is used whether hydrostatic is true or false? Seems there are 2 options:
1. Change it in cxx and accept non-bfb changes, or
2. revert this PR and then change the constant version in PR https://github.com/E3SM-Project/E3SM/pull/5299 to single precision (and change `0.0065f` in ElementOps.hpp to use the new constant defined in that PR).

ambrad:
I think your original plan was to switch the constant to double precision. Thus, I suggest you open a 1-line PR to fix the C++ constant to be double precision and verify this fixes the unit test while changing no other test results. Then have @oksanaguba integrate this new PR as a fix to this one. This is just a suggestion, however; @oksanaguba should make the final decision since she is integrating these changes.

oksanaguba:
1-line cxx PR sounds good.

PR #5318, opened 2022-11-19
Hommexx: Add missing team_barriers to omega computation.

ambrad:
This PR fixes a source of nondeterminism on the Crusher development system for Frontier in SCREAMv1 runs.

[BFB]

### DIFF
--- a/components/homme/src/theta-l_kokkos/cxx/CaarFunctorImpl.hpp
+++ b/components/homme/src/theta-l_kokkos/cxx/CaarFunctorImpl.hpp
@@ -633,14 +633,20 @@ struct CaarFunctorImpl {
       Kokkos::single(Kokkos::PerThread(kv.team),[&]() {
         pi_i(0)[0] = m_hvcoord.ps0*m_hvcoord.hybrid_ai0;
       });
+      kv.team_barrier();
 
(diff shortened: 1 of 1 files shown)


ndkeen:
Why do you say it fixes a problem on Crusher? As opposed to fixes a problem that could show on any GPU machine?

ambrad:
This specific source of nondeterminism hasn't shown up on any other machine so far. The pm-gpu and ascent nightlies have been green for a while now. But the same tests are nondeterministic on Crusher. This is an example of an issue that occurs fairly frequently on new (especially new GPU architecture) machines: latent bugs appear. One reason is that specific GPUs will have memory and threading guarantees stronger than the advertised abstract memory and threading models but will relax these further to the weak models with each generation; similarly, switching from one vendor's GPU to another's will make the code deal with a different subset of stronger guarantees. The end result is that we have no way of testing our code against the true weak memory and sync models but instead just have to run tests and see what emerges on each new platform.

jgfouca (approved):

jgfouca (approved):

ambrad:
e3sm_developer and homme_integration pass on Chrysalis. Passes e3sm_scream_v1_long_master_gnugpu on ascent when integrated into SCREAMv1.

PR #5322, opened 2022-11-21
case.run.sh: Some fixes and upgrades

jgfouca:
Needs to use python3.
Now supports get_timing.

[BFB]

### DIFF
--- a/cime_config/machines/template.case.run.sh
+++ b/cime_config/machines/template.case.run.sh
@@ -7,7 +7,8 @@
 # by case.submit when on batch. Use case.submit from the command line to run your case.
 
 # cd to case
-cd {{ caseroot }}
+export CASEROOT={{ caseroot }}
+cd $CASEROOT
 
 # Set PYTHONPATH so we can make cime calls if needed
 LIBDIR={{ cimeroot }}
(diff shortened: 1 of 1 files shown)


cime_config/machines/template.case.run.sh
```diff
@@ -43,6 +44,16 @@ export OMP_NUM_THREADS=$(./xmlquery THREAD_COUNT --value)
...
 {{ mpirun }}
 
+# get timing
+python3 -c '
+import os
+from CIME.case import Case
+from CIME.get_timing import get_timing
+
+with Case(os.environ["CASEROOT"], read_only=True) as case:
+    get_timing(case, os.environ["LID"])
```
mt5555, line 54: just curios: since this script has the variable $lid, why use a call to os.environ?

mt5555 (approved):

cime_config/machines/template.case.run.sh
jgfouca: In bash, the shell does not try to evaluate syntax inside single-quoted strings. This is what I want since I don't want bash getting confused by python syntax. This means I can't just drop in `$lid` and have it work. That said, it's not hard to terminate the single quoted string and concatenate the variables, so I will do that.

PR #5323, opened 2022-11-22
share: Protect bool typedef from stdbool.h inclusion.

ambrad:
Some compiler wrappers add headers that include stdbool.h. In a C build, including stdbool.h conflicts with the typedef. Guard the typedef with a check for whether 'true' has already been defined.

[BFB]

### DIFF
--- a/share/timing/private.h
+++ b/share/timing/private.h
@@ -44,7 +44,10 @@
 #define MAX_AUX 9
 
 #ifndef __cplusplus
+// Protect against inclusion of stdbool.h in, e.g., a compiler wrapper.
+# ifndef true
 typedef enum {false = 0, true = 1} bool;  /* mimic C++ */
+# endif
 #endif
 
 typedef struct {


rljacob:
What about just including stdbool.h explicitly and removing that line?

ambrad:
@rljacob I'm fine with that, too. The approach I took was meant for maximum flexibility, but that isn't really necessary. I'll switch to including stdbool.h.

dqwu:
> @rljacob I'm fine with that, too. The approach I took was meant for maximum generality, but that isn't really necessary. I'll switch to including stdbool.h.

@ambrad Note that stdbool.h is only available since C99. The existing enum typedef works in all cases (e.g. before C99).

rljacob:
Is requiring C99 a problem?

dqwu:
I am OK with current fix. Switching to including stdbool.h is another option if C99 (or newer C standard) is always required by E3SM.

rljacob:
I checked and we don't specify a C/C++ version standard yet so go ahead with this PR as is.

rljacob (approved):

dqwu (approved):
Looks good to me. We can also switch to including stdbool.h if C99 is supported.

PR #5324, opened 2022-11-22
Fix bfb fail in elem_ops_ut_test

tcclevenger:
This PR adds `PhysicalConstant` entry for `Tref_lapse_rate` that is `Real` type, and uses this constant (along with `Tref`) in `ElementOps.hpp` instead of hard coding. The bfb fix comes from the fact that the old version of `Tref_lapse_rate` was a `float` type (not matching the new F90 version from https://github.com/E3SM-Project/E3SM/pull/5315 which is `double`).

[bfb]

### DIFF
--- a/components/homme/src/share/cxx/PhysicalConstants.hpp
+++ b/components/homme/src/share/cxx/PhysicalConstants.hpp
@@ -27,6 +27,7 @@ namespace PhysicalConstants
 
   constexpr Real Tref          = 288;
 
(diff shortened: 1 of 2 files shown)


oksanaguba:
thanks, i will wait a little and then test it today.

oksanaguba:
homme suite on chrysalis pass, in next

tcclevenger:
Thanks @oksanaguba!

PR #5326, opened 2022-11-23
On crusher, disable archiving of old test data

jgfouca:
Apparently our shared disk quota for /gpfs/alpine/cli133/proj-shared is pretty limited and some of our nighltly runs are started to fail with "Disk quota exceeded". With this limitation, we cannot afford to archive old test data.

[BFB]


jgfouca:
Skipping integration.

PR #5328, opened 2022-11-26
homme: move limiters from caar functor

oksanaguba:
Ttype 9 design does not support a limiter for a sum of stages, as needed by ttype9 (and by ttype5, see below).
The limiter then needed to be moved out of caar to be used between its calls.
The code is essentially the same, using Caar functor definitions and buffer infrastructure to make a new Limiter functor.

ttype5 uses the same expl table as ttype9, but F code for ttype5 does not have a limiter for the sum, so cxx code does not have one either (easy to add).

To make sure limiters are active in a test, make dp3d threshold an nl variable.

tests were done for theta-fhs* tests triggering dp3d and vtheta parts of the limiter in ttype5,9, and 10. 
Nightly test will be only using a triggered limiter in ttype9.

[nonbfb] for new homme tests that trigger limiters.


components/homme/src/share/control_mod.F90
```diff
@@ -86,7 +86,7 @@ module control_mod
 ! advect theta 0: conservation form 
 !              1: expanded divergence form (less noisy, non-conservative)
  integer, public :: theta_advect_form = 0
- real (kind=real_kind), public :: vtheta_thresh = 100  ! threshold for virtual potential temperature minimum limiter
+ real (kind=real_kind), public :: vtheta_thresh = 400.d0  ! threshold for virtual potential temperature minimum limiter
```
oksanaguba, line 89: will do this in a namelist.

components/homme/src/theta-l/share/prim_advance_mod.F90
```diff
@@ -1795,20 +1795,20 @@ subroutine limiter_dp3d_k(dp3d,vtheta_dp,spheremp,dp0)
   ! local
   real (kind=real_kind) :: Qcol(nlev)
   real (kind=real_kind) :: mass,mass_new
-  real (kind=real_kind) :: dp3d_thresh=.125
+  real (kind=real_kind) :: dp3d_thresh=1.d0
```
oksanaguba, line 1798: will add this var to control mod and a namelist.

components/homme/src/theta-l_kokkos/cxx/LimiterFunctor.hpp
```diff
@@ -0,0 +1,215 @@
...
+    m_np1 = tl;
+    Kokkos::parallel_for("caar loop dp3d limiter", m_policy_dp3d_lim, *this);
+    Kokkos::fence();
+    GPTLstop("caar limiter");
+
+    profiling_pause();
+  }
+
+
+//NEED TO THROW IN SOME MESSAGE
```
oksanaguba, line 110: will check Andrew's scream PR on diagnostics and add a message to the limiter in the same way.

oksanaguba:
@bartgol or @ambrad  -- could you please see if this looks ok? not done yet, but most pieces of this PR are done.

oksanaguba:
NOT DONE, but needs a prelim. approval

components/homme/src/theta-l_kokkos/cxx/LimiterFunctor.hpp
```diff
@@ -0,0 +1,215 @@
...
+          }
+        }
+      });
+#endif
+    });
+    kv.team_barrier();
+  }
+
+
+//why is this here?
```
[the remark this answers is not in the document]
bartgol: I suspect this is a leftover from long ago, when we were using Scratch memory (in the kokkos sense) views. I don't think we use any scratch memory anymore, and all of our temps are from global memory.

The KernelVariables::shmem_size fcn returns 0, so this is completely pointless. Kokkos does not require functors to provide this accessor (if they don't, it assumes 0 bytes of shmem are needed).

In short, I think you can prune this.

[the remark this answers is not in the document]
bartgol: And we should probably prune it from DirkFunctor, KernelVariables, and CaarFunctorImpl as well.

components/homme/src/theta-l_kokkos/cxx/cxx_f90_interface_theta.cpp
```diff
@@ -345,12 +347,14 @@ void init_functors_c (const bool& allocate_buffer)
...
-  auto& vrm  = c.create_if_not_there<VerticalRemapManager>(elems.num_elems());
+  auto& hvf     = c.create_if_not_there<HyperviscosityFunctor>();
+  auto& ff      = c.create_if_not_there<ForcingFunctor>();
+  auto& diag    = c.create_if_not_there<Diagnostics> (elems.num_elems(),params.theta_hydrostatic_mode);
+  auto& vrm     = c.create_if_not_there<VerticalRemapManager>(elems.num_elems());
 
-  auto& fbm  = c.create_if_not_there<FunctorsBuffersManager>();
+  auto& fbm     = c.create_if_not_there<FunctorsBuffersManager>();
+
+//OG why are if-statement here -- above calls define which constructor is called
```
bartgol, line 357: I don't recall exactly the details, but I think it was related with the order in which things are inited when running eamxx. IIRC, in eamxx we create the functors before the Elements structures are ready (we still need to receive the pointers from the atm driver), so we create them "bare", but with enough info to establish how much memory they need from the FunctorsBufferManager, so that the atm driver can create the scratch buffers. Basically this is the sequence from the AD perspective

1. create atm processes (include dynamics)
  a. dyn creates all functors, does not set elements/tracers views
  b. dyn computes needs for buffers
2. get buffers requirements from all atm processes
3. allocate atm fields
4. pass fields and buffer to atm processes
  a. dyn can set ptrs inside Homme elements/tracers views, and complete functors setup

@tcclevenger might remember more details though.

bartgol (approved):
Looks good to me.

components/homme/src/theta-l_kokkos/cxx/cxx_f90_interface_theta.cpp
tcclevenger: That correct, we basically needed to create them inside EAMxx, but without actually having all the information to do the setup.

Each of these classes has a constructor which defines the base variables needed to talk to the buffer manager (I think just `num_elem` and mabye `params`) and then stores a bool which says "this wasn't actually setup". So here, if that is the case, we finish setting up the class.

On the other hand, if the class was constructed fully, then the class store the bool as "already setup" and nothing happens below.

oksanaguba:
This is ready up to resolving minor conflicts.

components/homme/src/theta-l_kokkos/cxx/LimiterFunctor.hpp
```diff
@@ -0,0 +1,214 @@
...
+
+      // Check if the minimum dp3d in this column is blow a certain threshold
+      auto dp = Homme::subview(m_state.m_dp3d,kv.ie,m_np1,igp,jgp);
+      auto& dp0 = m_hvcoord.dp0;
+      auto diff = Homme::subview(m_buffers.buffer1,kv.team_idx,igp,jgp);
+      Kokkos::parallel_for(Kokkos::ThreadVectorRange(kv.team,NUM_LEV),
+                           [&](const int ilev) {
+        diff(ilev) = (dp(ilev) - m_dp3d_thresh*dp0(ilev))*spheremp;
+      });
+
```
ambrad, line 140: Technically need a team_barrier here since in principle NUM_LEV /= NUM_PHYSICAL_LEV.

components/homme/src/theta-l_kokkos/cxx/LimiterFunctor.hpp
```diff
@@ -0,0 +1,214 @@
...
+        // Gotta apply vertical mixing, to prevent levels from getting too thin.
+        Real mass = 0.0;
+        ColumnOps::column_reduction<NUM_PHYSICAL_LEV>(kv.team,diff,mass);
+
+        if (mass<0) {
+          Kokkos::parallel_for(Kokkos::ThreadVectorRange(kv.team,NUM_LEV),
+                               [&](const int ilev) {
+            diff(ilev) *= -1.0;
+          });
+        }
```
ambrad, line 167: Same here.

components/homme/src/theta-l_kokkos/cxx/LimiterFunctor.hpp
```diff
@@ -0,0 +1,214 @@
...
+          if (mass<0) {
+            diff(ilev) *= -1.0;
+          }
+
+          dp(ilev) = diff(ilev)/spheremp + m_dp3d_thresh*dp0(ilev);
+          vtheta_dp(ilev) *= dp(ilev);
+        });
+      }
+
+#if 1
```
ambrad, line 194: Should remove this `#if 1`.

ambrad (approved):
I recommend applying these changes to SCREAM and running ERS_Ln362.ne30pg2_ne30pg2.F2010-SCREAMv1 on Summit or Ascent. Also check performance of an ne30pg2 SCREAMv1 case on Summit/Ascent without and with this PR, e.g., by looking at the model_timing_stats of that same test compared with that of a run w/o this PR. Doing so will make a future upstream merge to SCREAM smooth.

components/homme/src/theta-l_kokkos/cxx/LimiterFunctor.hpp
```diff
@@ -114,8 +119,8 @@ struct LimiterFunctor {
     KernelVariables kv(team, m_tu);
 
     // TODO: make this less hard-coded maybe?
-    constexpr double dp3d_thresh = 1.0;
-    constexpr double vtheta_thresh = 400; // Kelvin
+    //constexpr double dp3d_thresh = 1.0;
```
bartgol, line 122: I guess these lines can be removed now?

components/homme/src/theta-l_kokkos/cxx/LimiterFunctor.hpp
oksanaguba: thanks, done

components/homme/src/theta-l_kokkos/cxx/LimiterFunctor.hpp
oksanaguba: thanks, done

components/homme/src/theta-l_kokkos/cxx/LimiterFunctor.hpp
oksanaguba: thanks, done

components/homme/src/theta-l_kokkos/cxx/LimiterFunctor.hpp
oksanaguba: thanks, done

oksanaguba:
Running on summit scream master vs scream with these changes, ne30pg2, 1 node, 6 gpus
```
[onguba@login2.summit timing]$ pwd
/autofs/nccs-svm1_home1/onguba/summit-test-limiter-branch/summit-limiter-nnodes1-dec14/timing
[onguba@login2.summit timing]$ source amb-timings-script.sh e3sm_timing_stats.2600972.221214-151620 
"a_i:EAMxx::Dynamics::init"                                -          6        6 6.000000e+00   7.818183e+02   130.416 (     2      0)   129.739 (     3      0)
"a:EAMxx::run"                                             -          6        6 1.440000e+03   1.238744e+03   213.406 (     4      0)   197.404 (     2      0)
"a:EAMxx::Dynamics::run"                                   -          6        6 1.440000e+03   5.711699e+02    95.201 (     3      0)    95.184 (     2      0)
"a:tl-ae prim_advance_exp"                                 -          6        6 8.640000e+03   4.430965e+02    73.873 (     4      0)    73.834 (     3      0)
"a:caar compute"                                           -          6        6 8.640000e+04   1.400048e+02    23.373 (     2      0)    23.306 (     5      0)
"a:caar_bexchV"                                            -          6        6 4.320000e+04   5.642010e+01     9.494 (     5      0)     9.328 (     1      0)
"a:caar limiter"                                           -          6        6 5.184000e+04   9.231801e+00     1.541 (     5      0)     1.536 (     2      0)
"a:compute_stage_value_dirk"                               -          6        6 4.320000e+04   1.604975e+02    26.910 (     4      0)    26.644 (     0      0)
"a:tl-ae advance_hypervis_dp"                              -          6        6 8.640000e+03   7.632911e+01    12.768 (     3      0)    12.667 (     4      0)
"a:Remap Compute Remap Functor"                            -          6        6 4.320000e+03   2.879362e+01     4.805 (     3      0)     4.797 (     2      0)
"a:compose_transport"                                      -          6        6 1.440000e+03   4.994753e+01     8.337 (     5      0)     8.316 (     4      0)
"a:compose_hypervis_scalar"                                -          6        6 1.440000e+03   2.533425e+01     4.226 (     2      0)     4.220 (     3      0)
[onguba@login2.summit timing]$ cd -
/autofs/nccs-svm1_home1/onguba/summit-test-limiter-branch/summit-limiter-baseline-nnodes1-dec12/timing
[onguba@login2.summit timing]$ pwd
/autofs/nccs-svm1_home1/onguba/summit-test-limiter-branch/summit-limiter-baseline-nnodes1-dec12/timing
[onguba@login2.summit timing]$ cp ../../summit-limiter-nnodes1-dec14/timing/amb-timings-script.sh .
[onguba@login2.summit timing]$ source amb-timings-script.sh e3sm_timing_stats.2598297.221212-212859 
"a_i:EAMxx::Dynamics::init"                                -          6        6 6.000000e+00   7.795322e+02   129.924 (     2      0)   129.916 (     4      0)
"a:EAMxx::run"                                             -          6        6 1.440000e+03   1.219693e+03   210.300 (     4      0)   194.457 (     2      0)
"a:EAMxx::Dynamics::run"                                   -          6        6 1.440000e+03   5.693031e+02    94.896 (     0      0)    94.869 (     2      0)
"a:tl-ae prim_advance_exp"                                 -          6        6 8.640000e+03   4.423570e+02    73.745 (     4      0)    73.715 (     3      0)
"a:caar compute"                                           -          6        6 8.640000e+04   1.400837e+02    23.376 (     2      0)    23.318 (     0      0)
"a:caar_bexchV"                                            -          6        6 4.320000e+04   5.646544e+01     9.518 (     0      0)     9.337 (     1      0)
"a:caar dp3d"                                              -          6        6 4.320000e+04   4.616734e+00     0.773 (     2      0)     0.768 (     5      0)
"a:compute_stage_value_dirk"                               -          6        6 4.320000e+04   1.605359e+02    26.885 (     4      0)    26.662 (     0      0)
"a:tl-ae advance_hypervis_dp"                              -          6        6 8.640000e+03   7.638477e+01    12.774 (     3      0)    12.670 (     4      0)
"a:Remap Compute Remap Functor"                            -          6        6 4.320000e+03   2.880553e+01     4.805 (     3      0)     4.799 (     5      0)
"a:compose_transport"                                      -          6        6 1.440000e+03   4.992834e+01     8.330 (     2      0)     8.305 (     4      0)
"a:compose_hypervis_scalar"                                -          6        6 1.440000e+03   2.536140e+01     4.232 (     2      0)     4.223 (     0      0)

```
looks acceptable to me.

oksanaguba:
scream branch is https://github.com/E3SM-Project/scream/pull/2101

ambrad:
Question about the timers: "caar limiter" is 2x "caar dp3d" because of the limiter, right? Should we then expect "caar compute" to have a slightly lower time in your branch?

oksanaguba:
"caar compute" did not include limiter in master and does not include limiter in the branch. 2x factor -- i do not have a good explanation except that overall the time is probably too short to make conclusions, plus the limiter is now active 5 times, not 4 like before. i looked at the logs, does not seem that the limiter got activated in either run (except for checks).

ambrad:
GPU timers tend to be quite steady. I recommend running the performance comparison three times and seeing if the "caar dp3d" disparity remains.

oksanaguba:
running now.
there might be an issue with 
```
    profiling_resume();
//run limiter
    profiling_pause();
```
since in caar, everything, incl. limiter, is already wrapped in profiling_resume/profiling_pause .

ambrad:
profiling_reesume/pause calls have nothing to do with the timers. Those calls are for profiling tools like nvprof, vtune, gprof, etc. See homme/src/share/cxx/profiling.hpp for more. In the standard case, the bottom-most empty defintions are enabled.

oksanaguba:
results are still the same
```
[onguba@login3.summit timing]$ grep "caar limite" *
e3sm_timing_stats.2600972.221214-151620:"a:caar limiter"                                           -          6        6 5.184000e+04   9.231801e+00     1.541 (     5      0)     1.536 (     2      0)
e3sm_timing_stats.2603999.221215-135510:"a:caar limiter"                                           -          6        6 5.184000e+04   9.266196e+00     1.548 (     0      0)     1.542 (     2      0)
e3sm_timing_stats.2604274.221215-144318:"a:caar limiter"                                           -          6        6 5.184000e+04   9.261524e+00     1.552 (     0      0)     1.540 (     2      0)
[onguba@login3.summit timing]$ cd ../../summit-limiter-baseline-nnodes1-dec12/timing/
[onguba@login3.summit timing]$ grep "caar limite" *
[onguba@login3.summit timing]$ grep "caar dp3d" *
e3sm_timing_stats.2598297.221212-212859:"a:caar dp3d"                                              -          6        6 4.320000e+04   4.616734e+00     0.773 (     2      0)     0.768 (     5      0)
e3sm_timing_stats.2603858.221215-125700:"a:caar dp3d"                                              -          6        6 4.320000e+04   4.629635e+00     0.774 (     1      0)     0.769 (     5      0)
e3sm_timing_stats.2604005.221215-135715:"a:caar dp3d"                                              -          6        6 4.320000e+04   4.635959e+00     0.782 (     2      0)     0.768 (     1      0)
e3sm_timing_stats.2604271.221215-144150:"a:caar dp3d"                                              -          6        6 4.320000e+04   4.644956e+00     0.779 (     0      0)     0.769 (     1      0)
```

ambrad:
Ok, looks like something you probably should investigate.

ambrad:
One thing to check: Did the two team_barriers that were added affect the time? Another, less likely possibility, is to comment out the new printf statements, in case they are causing the register count to get really bad.

oksanaguba:
yes, i did all that, removed print statements (they are not triggered but there were tem vars, and an if statement) and removed new barriers, i am seeniong no change.

Limiter now is its own class with a few members, not using caar members anymore, and 1 new buffer, not sure it matters.

ambrad:
I don't see anything else obvious, so you'll have to do some more code instrumentation to get to the bottom of this.

Re: "active 5 times, not 4": The timers to me look like it's run 6 times per time step, up from 5:
```
>>> 5.184e4/6/1440  # now
6.0
>>> 4.320e4/6/1440  # before
5.0
```

oksanaguba:
yes, 5 before, 6 after. forgot that it is RK 5 stages method. So 5 RK stages before, 5+1 after for 1 more between stages.

oksanaguba:
On weaver in standalone homme in a run that does not trigger limiters (but with checks) and the state of the branch as is
master: limiter is 0.39, caar compute is 7.4
branch: limiter is 0.39, caar compute is 7.6

oksanaguba:
The added 6th call in between stages causes the slowdown:
```
[onguba@weaver3 lim-baseline]$ grep caar baseline-run 
"caar compute"                      -          1        1 1.440000e+04   7.372850e+00     7.373 (     0      0)     7.373 (     0      0)
"caar_bexchV"                       -          1        1 7.200000e+03   2.601201e+00     2.601 (     0      0)     2.601 (     0      0)
"caar dp3d"                         -          1        1 7.200000e+03   3.871825e-01     0.387 (     0      0)     0.387 (     0      0)
[onguba@weaver3 lim-baseline]$ cd -                  
[onguba@weaver3 lim-branch]$ grep caar stats-without-6th-call 
"caar compute"                      -          1        1 1.440000e+04   7.632370e+00     7.632 (     0      0)     7.632 (     0      0)
"caar_bexchV"                       -          1        1 7.200000e+03   2.592234e+00     2.592 (     0      0)     2.592 (     0      0)
"caar limiter"                      -          1        1 7.200000e+03   3.846677e-01     0.385 (     0      0)     0.385 (     0      0)
[onguba@weaver3 lim-branch]$ grep caar with6th-call 
"caar compute"                      -          1        1 1.440000e+04   7.624785e+00     7.625 (     0      0)     7.625 (     0      0)
"caar_bexchV"                       -          1        1 7.200000e+03   2.592223e+00     2.592 (     0      0)     2.592 (     0      0)
"caar limiter"                      -          1        1 8.640000e+03   6.084271e-01     0.608 (     0      0)     0.608 (     0      0)
```

ambrad:
Ok, so the discrepancy is only in your Summit runs?

oksanaguba:
No, there is discrepancy here, too.5 stages runs with limiter called from caar get ~80 units per call, 390 units total for 5 limiter call. Adding the call between stages increases this number to ~610 units.

Also, separately, the caar timer is bigger, too. The difference is that now caar timer gets a limiter from context. i can disable that code and see if the caar timer matches the old caar timer.

ambrad:
Oh, I see. The sixth call should make the "caar limiter" time ~0.385*6/5 = 0.462, but it's actually 0.608, which is similar to the strange slowdown you observed on Summit.

ambrad:
It's possible the issue is there is no `Kokkos::fence()` call before `limiter.run()` in the sixth case. I'm thinking the previous kernel's computational time is thus appearing in the limiter's timer region.

components/homme/src/theta-l_kokkos/cxx/prim_advance_exp.cpp
```diff
@@ -298,7 +306,7 @@ void ttype9_imex_timestep(const TimeLevel& tl,
            w(ie,np1,igp,jgp,LAST_INT)  += (w(ie,nm1,igp,jgp,LAST_INT)-w(ie,n0,igp,jgp,LAST_INT))/4.0;
       });  
     }
-    //LIMITER TO ADD!
+    limiter.run(np1);
```
ambrad, line 309: Try putting a `Kokkos::fence()` call before this call.

oksanaguba:
Now with 
```
--- a/components/homme/src/theta-l_kokkos/cxx/prim_advance_exp.cpp
+++ b/components/homme/src/theta-l_kokkos/cxx/prim_advance_exp.cpp
@@ -306,10 +306,11 @@ void ttype9_imex_timestep(const TimeLevel& tl,
            w(ie,np1,igp,jgp,LAST_INT)  += (w(ie,nm1,igp,jgp,LAST_INT)-w(ie,n0,igp,jgp,LAST_INT))/4.0;
       });  
     }
-    limiter.run(np1);
   }
   Kokkos::fence();
 
+  limiter.run(np1);
+
   Real a1 = 5.0*dt_dyn/18.0;
   Real a2 = dt_dyn/36.0;
```
limiter timer looks good,
```
"caar compute"                      -          1        1 1.440000e+04   7.631106e+00     7.631 (     0      0)     7.631 (     0      0)
"caar_bexchV"                       -          1        1 7.200000e+03   2.613263e+00     2.613 (     0      0)     2.613 (     0      0)
"caar limiter"                      -          1        1 8.640000e+03   4.646900e-01     0.465 (     0      0)     0.465 (     0      0)
```
but caar compute is still bigger than the original.

components/homme/src/theta-l_kokkos/cxx/CaarFunctorImpl.hpp
```diff
@@ -362,10 +362,7 @@ struct CaarFunctorImpl {
       GPTLstop("caar compute");
     }
 
-    GPTLstart("caar dp3d");
-    Kokkos::parallel_for("caar loop dp3d limiter", m_policy_dp3d_lim, *this);
-    Kokkos::fence();
```
oksanaguba, line 367: i removed this fence without thinking. new limiter does not have a fence, is fence needed after the limiter?

ambrad:
Ok, well, keep trying to isolate the "caar compute" issue.

components/homme/src/theta-l_kokkos/cxx/CaarFunctorImpl.hpp
ambrad: A fence is needed if there is a data dependency in the adjacent kernels.

oksanaguba: since it is only about loops in vertical and there is a team barrier, i assume it is not needed here.

ambrad: Here you don't need a fence because there's one above, line 361 or 355.

components/homme/src/theta-l_kokkos/cxx/prim_advance_exp.cpp
ambrad: Here you do need a fence because dp3d is updated in the kernel right before the limiter is called, and the limiter reads and potentially writes dp3d.

oksanaguba:
Thanks -- i commented out a few changes in Caar without success, lost connection to weaver now.

rljacob:
notes:  had some performance issues but they were resolved.

oksanaguba:
the last commit fixes issues with lim timer, caar timer discrepancy wasn't an issue (wrong baseline commit)
```
[onguba@weaver10 lim-branch]$ grep caar HommeTime_stats 
"caar compute"                      -          1        1 1.440000e+04   7.619144e+00     7.619 (     0      0)     7.619 (     0      0)
"caar_bexchV"                       -          1        1 7.200000e+03   2.602077e+00     2.602 (     0      0)     2.602 (     0      0)
"caar limiter"                      -          1        1 8.640000e+03   4.609092e-01     0.461 (     0      0)     0.461 (     0      0)
[onguba@weaver10 lim-branch]$ grep caar ../lim-baseline/HommeTime_stats 
"caar compute"                      -          1        1 1.440000e+04   7.619505e+00     7.620 (     0      0)     7.620 (     0      0)
"caar_bexchV"                       -          1        1 7.200000e+03   2.601042e+00     2.601 (     0      0)     2.601 (     0      0)
"caar dp3d"                         -          1        1 7.200000e+03   3.864237e-01     0.386 (     0      0)     0.386 (     0      0)
[onguba@weaver10 lim-branch]$ grep tstep_ty theta-no.nl 
tstep_type        = 9
[onguba@weaver10 lim-branch]$ pwd
/ascldap/users/onguba/runhomme-weaver/lim-branch
```

oksanaguba:
Summit timer, scream with temp branch with the new changes and master as of today:
```
[onguba@login4.summit summit-limiter-nnodes1-jan06]$ zgrep caar ../summit-limiter-baseline-nnodes1-jan06/timing/e3sm_timing_stats.2650969.230106-235314.gz 
"a:caar compute"                                           -          6        6 8.640000e+04   1.399331e+02    23.348 (     2      0)    23.284 (     4      0)
"a:caar_bexchV"                                            -          6        6 4.320000e+04   5.634261e+01     9.464 (     5      0)     9.322 (     1      0)
"a:caar dp3d"                                              -          6        6 4.320000e+04   4.582319e+00     0.767 (     2      0)     0.761 (     0      0)
[onguba@login4.summit summit-limiter-nnodes1-jan06]$ zgrep caar timing/e3sm_timing_stats.2650981.230107-000850.gz 
"a:caar compute"                                           -          6        6 8.640000e+04   1.401362e+02    23.391 (     2      0)    23.335 (     4      0)
"a:caar_bexchV"                                            -          6        6 4.320000e+04   5.663796e+01     9.501 (     0      0)     9.385 (     4      0)
"a:caar limiter"                                           -          6        6 5.184000e+04   5.541190e+00     0.930 (     5      0)     0.922 (     1      0)
```

oksanaguba:
tested the last commit on weaver for 3 HS tests for bfb and for all unit tests.

this is ready.

oksanaguba:
in next

PR #5334, opened 2022-11-29
ELM: Use intent(inout) to preserve values in early exit.

ambrad:
In PhotosynthesisMod.F90, early exits from brent and brent_PHS do not assign gs_mol outputs values. The desired behavior in those cases is for the input values to be retained. Thus, this commit changes intent(out) to intent(inout) for these variables.

The original code usually works because Fortran normally doesn't overwrite these inputs at the start of the routine. But if an option like crayftn's -hzero is provided, intent(out) stack variables are 0-init'ed, causing a bug.

Fixes SCREAM issue 2057. e3sm.log was being filled with messages like this:

7:  Ball-Berry error check - stomatal conductance error:
7:  0.,  52577.053632404488

The 0 in the second line is important: it implies that gs_mol was being set to 0 unintentionally.

[BFB]

### DIFF
--- a/components/elm/src/biogeophys/PhotosynthesisMod.F90
+++ b/components/elm/src/biogeophys/PhotosynthesisMod.F90
@@ -1278,7 +1278,7 @@ contains
     real(r8), intent(in) :: oair              ! Atmospheric O2 partial pressure (Pa)
     real(r8), intent(in) :: rh_can            ! inside canopy relative humidity
     integer,  intent(in) :: ip, iv, ic, it    ! pft, c3/c4, column, and topounit index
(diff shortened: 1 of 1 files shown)


ambrad:
e3sm_developer passes on Chrysalis

peterdschwartz (approved):

ambrad:
@peterdschwartz should I wait for a review from Gautam? I wasn't sure whether the file I modified is within his purview. In any case, I'll merge this PR next week, not this week.

rljacob:
No its enough to get a review from one land integrator.

PR #5337, opened 2022-11-29
Add missing code to set threadNum before its use

### LINKED ISSUE #5333: MPAS: mpas_dmpar uses uninitialized threadNum in mpas_dmpar_minloc_real
`threadNum` isn't set in this routine:
https://github.com/E3SM-Project/E3SM/blob/fb7d271bfc9f343d395b4f08992cd5a949397cef/components/mpas-framework/src/framework/mpas_dmpar.F#L1000-L1026

jonbob:
Fixes an issue where a variable threadNum was used before it was set

Fixes #5333 
[BFB]

### DIFF
--- a/components/mpas-framework/src/framework/mpas_dmpar.F
+++ b/components/mpas-framework/src/framework/mpas_dmpar.F
@@ -971,6 +971,8 @@ include 'mpif.h'
       integer :: mpi_ierr, threadNum
       real(kind=RKIND), dimension(2,1) :: recvbuf, sendbuf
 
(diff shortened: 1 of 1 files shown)


jonbob:
@mark-petersen - I added you in part so you could decide if we need to pass this along to Michael Duda

ambrad (approved):

jonbob:

mark-petersen (approved):
@ambrad thanks for pointing this out. `threadNum` is used but uninitialized, so this is an obvious correction.

PR #5339, opened 2022-11-30
CMake: Set E3SM CXX standard globally

jgfouca:
Not in macros.

[BFB]


brhillman (approved):
This is awesome, thanks Jim!

PR #5342, opened 2022-12-01
change pbuf variables for CRM state total water to water vapor

whannah1:
This change for SAM++ is needed for the future upgrade to PAM, and also needed if we eventually put P3/SHOC into SAM++ (which seems unlikely now).  This is non-BFB due to the fact floating point arithmetic is non-communitive, but the qualititive solution will not change. 

[non-BFB] MMF only


xyuan (approved):

brhillman:
Merged to next.

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #5346, opened 2022-12-01
Preliminary MMF additions for P3 and remove support for m2005

whannah1:
This set of changes to the CRM interface code will be needed for the integration of the new PAM CRM w/ P3. I've omitted the changes to support SHOC for simplicity since we won't need these for the first MMF-PAM prototype. I also took the opportunity to drop MMF support for the M2005 micro scheme (which is likely already broken) and RRTMG. 

[BFB]


whannah1:
I just ran a 1-day ne4pg2 test on Perlmutter with F2010-MMF1 to confirm that the branch builds and runs.

rljacob:
notes: ready to go.

brhillman:
Merged to next.

brhillman (approved):

brhillman:
Merged to master.

PR #5347, opened 2022-12-03
For alvarez at NERSC, update modules to defaults.

ndkeen:
For alvarez at NERSC, update modules to machine defaults.
Including using GNU 12.1


rljacob (approved):

ndkeen:
Merged to next

PR #5350, opened 2022-12-06
On GCP only, add larger allowance for MEMLEAK detection in tests

ndkeen:
On GCP only, add larger allowance for MEMLEAK detection in tests
and add a slightly smaller allowance for TPUT.

This fixes a fail for one test in e3sm_extra_coverage: `ERP_Lm3.ne4_oQU240.F2010.gcp_gnu`
which we do not think is actually leaking memory (or growing in memory too quickly).

[bfb]


rljacob (approved):

ndkeen:
merged to next

ndkeen:
Tested next with this change on GCP manually -- ran e3sm_integration as well as the specific test that had MEMLEAK fail and they pass, so will merge to master.

PR #5047, opened 2022-06-24
Add WAVEWATCH III as new coupled wave component

sbrus89:
This is a stealth feature PR that adds the WAVEWATCH III submodule, cime configuration, namelist generation, build infrastracture, and coupling interface. 

Wave only, G case + waves, and B case + waves compsets are included along with model resolutions that include the wave mesh intended for v1 and v2 LR configurations. 

The initial coupled, active wave feedback is a parameterization for Langmuir mixing. Details can be found in the overview document below:

New Feature Overview Document: [E3SM_New_Feature_Overview___Waves.pdf](https://github.com/E3SM-Project/E3SM/files/8980373/E3SM_New_Feature_Overview___Waves.pdf)

MPAS-Analysis and E3SM-diags results for 10 year B-Case comparison:
[Coupled waves vs. Observations](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/sbrus/ww3_merge/waves)
[Reference case vs. Observations](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/sbrus/ww3_merge/ref)
[Coupled waves vs. Reference case](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/sbrus/ww3_merge/compare)

PACE performance results:
[v1 LR on Chrysalis](https://pace.ornl.gov/exp-details/95598)
[v2 LR resolution on Anvil](https://pace.ornl.gov/exp-details/108664)

[NML] for configurations with active ocean
[BFB]

PR #5352, opened 2022-12-06
Update ww3 spectra

erinethomas:
Update to WW3 to allow flexible options for spectral resolution of waves (N Freq x N Dir). 
Default set to 36x36. Alternative options of high resolution (50x36) and low resolution (25x36). 
Defining case with 'alias' that includes "WW3" component will use default spectral resolution. Alternative options require use of "long name" to define components.


components/ww3/cime_config/config_compsets.xml
```diff
@@ -11,37 +11,37 @@
 
   <compset>
     <alias>VWW3-IAF</alias>
-    <lname>2000_DATM%IAF_SLND_SICE_SOCN_SROF_SGLC_WW3</lname>
+    <lname>2000_DATM%IAF_SLND_SICE_SOCN_SROF_SGLC_WW3%sp36x36</lname>
```
sbrus89, line 14: The `%sp36x36` may not be needed in these aliases since `<value compset="_WW3">sp36x36</value>` is set in `components/ww3/cime_config/config_component.xml`, but I also think it's fine to leave it for clarity's sake.

cime_config/allactive/config_pesall.xml
```diff
@@ -2660,4 +2660,106 @@
...
+          <rootpe_lnd>1080</rootpe_lnd>
+          <rootpe_rof>1080</rootpe_rof>
+          <rootpe_ice>0</rootpe_ice>
+          <rootpe_ocn>1368</rootpe_ocn>
+          <rootpe_cpl>0</rootpe_cpl>
+          <rootpe_wav>1728</rootpe_wav>
+        </rootpe>
+      </pes>
+    </mach>
+  </grid>
```
sbrus89, line 2696: My suggestion would be to leave the placeholder PE layouts out of this PR, and submitting a follow-on PR with supported layouts once they have been tested. Let's see what @jonbob suggests, though.

components/ww3/cime_config/buildnml
```diff
@@ -50,8 +51,13 @@ def buildnml(case, caseroot, compname):
     # Verify wav grid is supported
     #--------------------------------------------------------------------
 
-    wav_grid_supported = ("null", "wQU225EC60to30sp32x36", "wQU225EC60to30", "wQU100", "wQU225EC30to60E2r2")
-    expect(wav_grid in wav_grid_supported, "WAV_GRID '{}' is not supported in ww3. Choose from: '{}'".format(wav_grid,' ,'.join(wav_grid_supported)))
+    wav_grid_supported = ("null",
+                          "wQU225EC60to30sp50x36",
+                          "wQU100sp50x36",
```
sbrus89, line 56: I think we should remove the wQU100 grid here. It's no longer supported in `config_grids.xml`. I must have overlooked this in the original waves PR - #5047

sbrus89 (commented):
@erinethomas - Thanks very much for this addition. The new 36x36 spectral resolution is a major performance improvement and the flexibility to optionally specify a high and low resolution configuration is a great feature. The way you implemented the functionality is super clean.

components/ww3/cime_config/buildnml
erinethomas: Sure, no problem. I wasn't sure about the wQU100 grid.

cime_config/allactive/config_pesall.xml
jonbob: Thanks @sbrus89 - I agree that we should wait until they're tested

components/ww3/cime_config/config_compsets.xml
erinethomas: I prefer leave it for clarity.

components/ww3/cime_config/buildnml
```diff
@@ -50,8 +51,12 @@ def buildnml(case, caseroot, compname):
...
     #--------------------------------------------------------------------
 
-    wav_grid_supported = ("null", "wQU225EC60to30sp32x36", "wQU225EC60to30", "wQU100", "wQU225EC30to60E2r2")
-    expect(wav_grid in wav_grid_supported, "WAV_GRID '{}' is not supported in ww3. Choose from: '{}'".format(wav_grid,' ,'.join(wav_grid_supported)))
+    wav_grid_supported = ("null",
+                          "wQU225EC60to30sp50x36",
+                          "wQU225EC30to60E2r2sp50x36",
+                          "wQU225EC30to60E2r2sp36x36",
+                          "wQU225EC30to60E2r2sp25x36")
+    expect((wav_grid+wav_spec)) in wav_grid_supported, "Combination of WAV_GRID {} and WAV_SPEC {} is not supported in ww3. Choose from: '{}'".format(wav_grid,wav_spec,wav_grid_supported) )
```
sbrus89, line 59:
```suggestion
    expect((wav_grid+wav_spec) in wav_grid_supported, "Combination of WAV_GRID {} and WAV_SPEC {} is not supported in ww3. Choose from: '{}'".format(wav_grid,wav_spec,wav_grid_supported) )
```

There's an extra parenthesis here that I noticed causes an error in `./case.setup`

erinethomas: great catch - thank you!

jonbob:
@erinethomas - can I point you at the confluence page that discusses the E3SM branch naming convention? It's [Branch Tag and Version name conventions](https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/2523172/Branch+Tag+and+Version+name+conventions). If you don't mind, please make this branch using those conventions. Unfortunately, you can't change the name of a branch a PR points to, so it will mean making a new PR that's otherwise the same. We can add comments to this PR as well as the new one so the comments are not lost.

erinethomas:
New PR created with correct naming convention. see pull #5368

PR #5356, opened 2022-12-07
Move bottomDepthEdge calculation to single loop over all edges

### LINKED ISSUE #5219: Ocean fails stand-alone decomp test, intel optimized

mark-petersen:
After #5195 was merged, the MPAS-Ocean standalone test `ocean/baroclinic_channel/10km/decomp_test` failed to match between 4 and 8 partitions, but only for intel optimized. All compass nightly suite tests passed for gnu debug, gnu optimized, intel debug. 

This PR solves the problem by merging the computation of `bottomDepthEdge` into a single edge loop. Previously it was split into two loops, `1:nEdgesOwned` (with many other calculations) and another from `nEdgesOwned+1:nEdgesArray(4)`. The intel optimized compiler must have changed order-of-operations in these two loops for different partitions.

Fixes #5219
[BFB]

### DIFF
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F
@@ -173,8 +173,6 @@ module ocn_time_integration_split
          edgeHaloComputeCounter, &! halo counters to reduce
          cellHaloComputeCounter   ! halo updates during cycling
 
-      integer, dimension(:), pointer :: nEdgesArray
-
       real (kind=RKIND) :: &
          normalThicknessFluxSum, &! sum of thickness flux in column
          thicknessSum,     &! sum of thicknesses in column
@@ -384,8 +382,6 @@ module ocn_time_integration_split
                                           lowFreqDivergenceTend)
 
       call mpas_pool_get_array(tracersPool, 'activeTracers', activeTracersNew, 2)
-
-      call mpas_pool_get_dimension(meshPool,'nEdgesArray', nEdgesArray)
 
       allocate(bottomDepthEdge(nEdgesAll+1))
 
@@ -796,7 +792,7 @@ module ocn_time_integration_split
             !$omp parallel
             !$omp do schedule(runtime) &
             !$omp private(cell1, cell2, k, thicknessSum)
-            do iEdge = 1, nEdgesArray(size(nEdgesArray)-1)
+            do iEdge = 1, nEdgesHalo(config_num_halos+1)
                cell1 = cellsOnEdge(1,iEdge)
                cell2 = cellsOnEdge(2,iEdge)
                thicknessSum = layerThickEdgeFlux(minLevelEdgeBot(iEdge),iEdge)
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F
@@ -775,8 +775,6 @@ module ocn_time_integration_split
                enddo
                barotropicForcing(iEdge) = splitFact* &
                               normalThicknessFluxSum/thicknessSum/dt
-               bottomDepthEdge(iEdge) = thicknessSum &
-                  - 0.5_RKIND*(sshNew(cell1) + sshNew(cell2))
 
                do k = minLevelEdgeBot(iEdge), maxLevelEdgeTop(iEdge)
                   ! These two steps are together here:
@@ -798,7 +796,7 @@ module ocn_time_integration_split
             !$omp parallel
             !$omp do schedule(runtime) &
             !$omp private(cell1, cell2, k, thicknessSum)
-            do iEdge = nEdgesOwned+1, nEdgesArray(4)
+            do iEdge = 1, nEdgesArray(size(nEdgesArray)-1)
                cell1 = cellsOnEdge(1,iEdge)
                cell2 = cellsOnEdge(2,iEdge)
                thicknessSum = layerThickEdgeFlux(minLevelEdgeBot(iEdge),iEdge)


mark-petersen:
Testing on cori, the compass nightly suite are bfb identical for all tests using gnu debug, gnu optimized, and intel debug. For intel optimized the following tests fail bfb match with master for this PR (90fb60dd50 to 582899e596):
```
00:06 FAIL ocean_baroclinic_channel_10km_default
00:10 FAIL ocean_baroclinic_channel_10km_threads_test
00:13 FAIL ocean_baroclinic_channel_10km_decomp_test
00:06 FAIL ocean_baroclinic_channel_10km_restart_test
01:41 PASS ocean_global_ocean_QU240_mesh
00:51 PASS ocean_global_ocean_QU240_PHC_init
00:48 PASS ocean_global_ocean_QU240_PHC_performance_test
01:33 PASS ocean_global_ocean_QU240_PHC_restart_test
01:31 PASS ocean_global_ocean_QU240_PHC_decomp_test
01:31 PASS ocean_global_ocean_QU240_PHC_threads_test
01:08 PASS ocean_global_ocean_QU240_PHC_analysis_test
00:42 PASS ocean_global_ocean_QU240_PHC_RK4_performance_test
01:25 PASS ocean_global_ocean_QU240_PHC_RK4_restart_test
01:29 PASS ocean_global_ocean_QU240_PHC_RK4_decomp_test
01:29 PASS ocean_global_ocean_QU240_PHC_RK4_threads_test
00:00 PASS ocean_global_ocean_QUwISC240_mesh
00:00 PASS ocean_global_ocean_QUwISC240_PHC_init
00:46 PASS ocean_global_ocean_QUwISC240_PHC_performance_test
00:14 FAIL ocean_ice_shelf_2d_5km_z-star_restart_test
00:24 FAIL ocean_ice_shelf_2d_5km_z-level_restart_test
00:16 FAIL ocean_ziso_20km_default
00:18 FAIL ocean_ziso_20km_with_frazil
```
All differences are 1e-12 or smaller. Considering that we changed order of operations for intel optimized, this is not surprising, and is acceptable for this PR.

xylar:
@mark-petersen, thanks for this fix!  I will test it in both compass and E3SM.  I see you have the BFB flag but I wouldn't expect it to necessarily be BFB in E3SM, based on your compass testing.  Let's see what happens.

xylar:
## Testing

### compass

I tested with the compass `pr` suite on Chrysalis with Intel and OpenMPI (optimized).  I used 90fb60dd50 as a baseline.  With the baseline, I saw validation failures with the baseline for:
```
ocean/baroclinic_channel/10km/decomp_test
  * step: initial_state
  * step: 4proc
  * step: 8proc
  test execution:      SUCCESS
  test validation:     FAIL
  see: case_outputs/ocean_baroclinic_channel_10km_decomp_test.log
  test runtime:        00:03
...
ocean/global_ocean/QU240/PHC/decomp_test
  * step: 4proc
  * step: 8proc
  test execution:      SUCCESS
  test validation:     FAIL
  see: case_outputs/ocean_global_ocean_QU240_PHC_decomp_test.log
  test runtime:        00:54
```
(This is interesting because `ocean/global_ocean/QU240/PHC/decomp_test` had passed in my previous testing with Intel and Intel-MPI, so the use of OpenMPI--what E3SM also uses on Chrysalis--may actually make this problem worse.)

With this branch, I see all tests passing execution and validation but all split-explicit tests are failing baseline comparison, as @mark-petersen saw and as we expected given the change of order of operations.  As @mark-petersen found, I'm seeing differences on the order of 1e-12, so small but not quite machine precision.

### E3SM testing

I am running `PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel`, again using 90fb60dd50 as a baseline.  This is still running and I will update here when it's done.

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F
```diff
@@ -798,7 +796,7 @@ subroutine ocn_time_integrator_split(domain, dt)!{{{
             !$omp parallel
             !$omp do schedule(runtime) &
             !$omp private(cell1, cell2, k, thicknessSum)
-            do iEdge = nEdgesOwned+1, nEdgesArray(4)
+            do iEdge = 1, nEdgesArray(size(nEdgesArray)-1)
```
philipwjones, line 799: Any reason we're not using nEdgesAll here? Would save a size evaluation. Also nEdgesArray has been deprecated in ocn_mesh and does not exist on the device, so if a halo width is needed, please use the nEdgesHalo array instead.

mark-petersen: `nEdgesAll` includes the outer edge of the last halo layer, which we don't need and caused an error in the loop. I'll try it with `nEdgesHalo` - thanks for pointing that out.

mark-petersen: I changed to `nEdgesHalo(config_num_halos+1)` and verified that the value is 4.

mark-petersen:
After the last commit, I retested the nightly suite on chrysalis. Intel debug matches bfb between master and this PR, and passes all tests. Intel optimized fails bfb comparison to master, same as [above](https://github.com/E3SM-Project/E3SM/pull/5356#issuecomment-1341679123).

xylar:
I'll retest this tomorrow.

xylar:
Both the `pr` test suite in compass (with Intel and OpenMPI in optimized mode) and the E3SM `PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel` passed as before.  Teh `pr` test suite was non-BFB for the same tests as before (and using the same baseline), whereas `PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel` was BFB with the same baseline as before.

xylar (approved):
Approving based on my testing and @mark-petersen's.

philipwjones (approved):
Approve based on code inspection and others' testing. Previous comments were satisfactorily addressed.  

Still getting tripped up by the fact that config_num_halos is not the actual halo width for edges, so this looks like we're out of bounds. But that's another discussion for omega...

jonbob:
passes:
* PET_Ln9_PS.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel.allactive-mach-pet
* PEM.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel.allactive-wcprod
* ERS_Ld5.T62_oQU120.CMPASO-NYF.compy_pgi

merged to next

jonbob:
merged to master

PR #5357, opened 2022-12-07
Add DMPAS-JRA1p5 compset

darincomeau:
Adds `DMPAS-JRA1p5` compset; D-case with active sea ice only.

Also adds PE layouts for any D-case on Chrysalis and Anvil, `S` layouts on 20 nodes, with ~47 and 23 SYPD, respectively.

### DIFF
--- a/components/mpas-seaice/cime_config/config_compsets.xml
+++ b/components/mpas-seaice/cime_config/config_compsets.xml
@@ -10,6 +10,11 @@
   <!-- E3SM D compsets -->
 
   <compset>
+    <alias>DMPAS-JRA1p5</alias>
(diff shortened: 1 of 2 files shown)


darincomeau:
I've run a one year `DMPAS-JRA1p5` case to test. @proteanplanet I can continue running this further and run analysis if desired as part of testing.

proteanplanet:
@darincomeau This looks good to me. Please note that we will not want to use the "S" layout very often, since the point of the D-test 4-node limit is that it almost instantly runs without getting stuck in a queue, and the 5-day SMS test only takes about 3 minutes of wall time on 4-notes.  However, it's good to have it in there as an alternative to the 4-node default.

proteanplanet (approved):
I approve this PR

jonbob:

jonbob:
merged to master

Issue #5001, opened 2022-05-31
lib-4412 : UNRECOVERABLE library error of Cray compiler CCE/14.0.0 on Crusher


grnydawn:
The error message is:

6: lib-4412 : UNRECOVERABLE library error
6:   An argument in the DEALLOCATE statement is a disassociated pointer, an
6:  unallocated array, or a pointer not allocated as a pointer.

Test case name is:

SMS_Ld20.f45_f45.IELMFATES.crusher_crayclang.elm-fates_eca
SMS_Ld20.f45_f45.IELMFATES.crusher_crayclang.elm-fates_rd 
ERS_Ld20.f45_f45.IELMFATES.crusher_crayclang.elm-fates

sarats:
Please add location of the error for reference.

sarats:
Naive question: this is a runtime error, right?
Just a sanity check, did any component log file leave a pointer to where the issue might be?

grnydawn:
@sarats , yes, it is a runtime error. The error message does not leave any pointer to where the issue occurred. One thing we know from the  e3sm log file is that this error occurred after after printing many balance check warnings. Please see the part of the error in e3sm log file.
...
40:  WARNING:: BalanceCheck, solar radiation balance error (W/m2)
40:  nstep         =  138
40:  errsol        =  -1.04153343727375614E-7
 3:
 3: lib-4412 : UNRECOVERABLE library error
 3:   An argument in the DEALLOCATE statement is a disassociated pointer, an
 3:  unallocated array, or a pointer not allocated as a pointer.
...

amametjanov:
IBM compiler on Ascent/Summit has a similar issue for these 3 cases. There's a work-around for IBM in NGEET/fates#824 .

Pinging @rgknox and @glemieux . Have you been able to get accounts on Summit (which would also enable access to Crusher)?

IBM: https://my.cdash.org/viewTest.php?onlyfailed&buildid=2223558
Cray: https://my.cdash.org/viewTest.php?onlyfailed&buildid=2223909

glemieux:
@amametjanov I've got access to Summit, but addressing https://github.com/NGEET/fates/issues/702 has been low on the priority list.  The solar radiation issue is a known issue as well (https://github.com/NGEET/fates/issues/794).  I'll chat with Ryan and Charlie about prioritizing this soon.

sarats:
@amametjanov Does the above workaround you devised for IBM work for Cray compilers as well? 
[...]

Issue #4820, opened 2022-03-01
E3SM Land Developer Tests Failing in DEBUG Mode


peterdschwartz:
With current master, I ran many of the land developer tests in debug mode ( case directory is here: /lcrc/group/e3sm/ac.schwartzpd/master-debug-tests/cases) and it seems any that use the chemistry of ELM all fail.  I believe this has been an issue for a while so I can't say how many code changes this will require and not each test failed in the same place.  Almost all of them are due to the triggering of a floating point exception edit:  due to evaluation of logical expressions involving NaNs. 

A potential fix is to add `shr_infnan_isnan` consistently as that will not trigger an fpe error.  An potential downside of `shr_infnan_isnan` is I am not sure how to readily port them to use in GPU regions -- a GPU friendly check that I implemented is to use if statements like `if(.not. (x .ne. x) )` which trigger a fpe (the FireMod backtraces below).  I can make a elm wrapper function that switches between the two at compilation but will have to check how adding an extra function call impacts performance.  

Currently, many new development runs cannot be run in DEBUG mode without removal of `-fpe0` flag, and since I do not believe it is a good policy that an atm developer should have to contact an elm developer just to enable debug mode, then the `-fpe0` should be temporarily removed for elm files. And tests will need to be added to catch these issues sooner as the current ELM debug tests are clearly inadequate.
```
ERS.f09_g16.I1850ELMCN.chrysalis_intel
ERS.f09_g16.I1850GSWCNPRDCTCBC.chrysalis_intel.elm-vstrd (different traceback)
ERS.f19_f19.I20TRELMCN.chrysalis_intel
ERS.f09_g16.I1850ELMCN.chrysalis_intel.elm-bgcinterface
ERS.r05_r05.IELM.chrysalis_intel.elm-V2_ELM_MOSART_features
ERS.f19_g16.I1850CNECACNTBC.chrysalis_intel.elm-eca
ERS.f19_g16.I1850GSWCNPECACNTBC.chrysalis_intel.elm-eca_f19_g16_I1850GSWCNPECACNTBC (no traceback just seg fault)
ERS.f19_g16.I20TRGSWCNPECACNTBC.chrysalis_intel.elm-eca_f19_g16_I20TRGSWCNPECACNTBC
ERS.f19_g16.I20TRGSWCNPRDCTCBC.chrysalis_intel.elm-ctc_f19_g16_I20TRGSWCNPRDCTCBC
[...]

PR #5369, opened 2022-12-19
[WIP] Update ELM-FATES API to pass fire data to FATES

glemieux:
This pull request updates the ELM-FATES API to provide FATES with the lightning, GDP, and population density data from FireMod.F90.  This provides ELM-FATES users access to the additional [SPITFIRE run modes](https://fates-users-guide.readthedocs.io/en/latest/user/SPITFIRE-Namelist-Options.html).  The design is adapts the framework developed for CLM-FATES.  The [design document](https://fates-users-guide.readthedocs.io/en/latest/developer/elm-fates-fire-design_doc.html) discussing the background and general design is available in the [FATES Developer's Guide](https://fates-users-guide.readthedocs.io/en/latest/developer/developer-guide.html).   

[B4B]


glemieux:
Removed WIP.  Testing TBD this week.

components/elm/bld/namelist_files/namelist_definition.xml
```diff
@@ -298,8 +298,15 @@ Toggle to turn on the FATES
 </entry>
 
 <entry id="fates_spitfire_mode" type="integer" category="physics"
-        group="elm_inparm" valid_values="0,1" >
-Set to 0 to keep fire off. Set to 1 to turn on fates spitfire module (only relevant if FATES is being used).
+        group="elm_inparm" valid_values="0,1,2,3,4,5", value=".false." >
```
rgknox, line 301: should the type on value match valid_values? ie, should this be 0 or 1?

components/elm/src/main/ndepStreamMod.F90
```diff
@@ -43,7 +42,7 @@ module ndepStreamMod
 
   !==============================================================================
 
-  subroutine ndep_init(bounds)
+  subroutine ndep_init(bounds, NLFilename)
```
rgknox, line 45: doesn't look like this new argument is used

components/elm/bld/namelist_files/namelist_definition.xml
glemieux: Good catch @rgknox.  This will need to be changed on the clm side as well.

rljacob:
@rgknox waiting for you to approve.

glemieux:
> @rgknox waiting for you to approve.

@rgknox just a heads up that @JessicaNeedham has a small update she wants to add to this PR.

rgknox:
@glemieux, I will wait until the final changes are in. Can you ping me when its ready?

glemieux:
> @glemieux, I will wait until the final changes are in. Can you ping me when its ready?

Will do.  Jessie is testing this out the changes.

JessicaNeedham:
Thanks @glemieux. I'm finding that establishment from bare ground works better when we start with higher initial soil moisture. I wondered about changing the use_fates_planthydro to use_fates here: https://github.com/E3SM-Project/E3SM/blob/b30a6a0cb3e610f80a00cd0fcdbfffab7044df54/components/elm/src/data_types/ColumnDataType.F90#L1619 so that soil moisture is  always initialised high when FATES is on. 
Details are here https://github.com/NGEET/fates/discussions/985 although maybe I should start a separate issue? I believe @mpaiao has also found it necessary to increase soil moisture.   
I haven't tested yet on this branch but I will do so now.

glemieux:
> Details are here [NGEET/fates#985](https://github.com/NGEET/fates/discussions/985) although maybe I should start a separate issue? I believe @mpaiao has also found it necessary to increase soil moisture.
> I haven't tested yet on this branch but I will do so now.

@JessicaNeedham We could convert the discussion to an issue if you feel like it has been fully addressed.  If there are still outstanding issues in the discussion then opening a new issue (and linking back to the open discussion) would be better.  

Skimming the discussion, are there changes on the fates-side that need to be made as well?  If so, it might be cleaner to update and integrate the fates-side first and then create new PRs for both host land model side.

JessicaNeedham:
Thanks @glemieux, I'd say it is not totally resolved so I'll make a new issue for some of the solutions with a corresponding FATES PR.

rgknox:
I agree @glemieux, it might be better to add those changes in another PR

rgknox (approved):

glemieux:
@rgknox we're going to do a different PR for Jessie's changes.  

I ran into a run time seg fault in the testing of spitfire in mode 2 right now.  I'll ping you again when I've got things fixed up.

glemieux:
@peterdschwartz I'm running regression testing on Cori currently.  I'll update with the results as soon as possible.

glemieux:
@peterdschwartz testing on perlmutter against the master baseline is complete.  All tests are B4B aside from the following:

- All fates tests have NLCOMP diffs as they include access to the lightning and pop dens data
- 2 non-fates tests have TPUTCOMP increases

Folder location: `/pscratch/sd/g/glemieux/e3sm-tests/pr5369.fates.pm-cpu.gnu.C41339b97c9-Fb8b905cc`

components/elm/src/biogeochem/FireDataBaseType.F90
```diff
@@ -0,0 +1,453 @@
...
+
+       !-----------------------------------------------------------------------
+   
+       if ( this%need_lightning_and_popdens() ) then
+          ! Allocate lightning forcing data
+          allocate( this%forc_lnfm(bounds%begg:bounds%endg) )
+          this%forc_lnfm(bounds%begg:) = nan
+          ! Allocate pop dens forcing data
+          allocate( this%forc_hdm(bounds%begg:bounds%endg) )
+          this%forc_hdm(bounds%begg:) = nan
```
peterdschwartz, line 107: Rather than initializing to `nan`, could you go ahead and change it to `spval`? I think that will be the recommended practice going forward

glemieux: Will do.  Thanks.

glemieux: @rgknox thinking more about this, `spval` is the positive value equivalent to `fates_unset_r8`.  What are the cases in which we want to use `nan` in fates, versus an unset value?  My assumption is that we want to use this over `nan` generally to make sure that in debug mode, they will end the run?   Do we have an discussion or issue open discussing this on the fates-side? 

@peterdschwartz is there documentation discussing `nan` versus `spval` for e3sm on confluence or in an issue?

peterdschwartz: @glemieux I don't know of details in the documentation regarding this in particular, but as far as I can tell, the approach to `nan` vs `spval` has been to initialize to `nan` at allocation and then if the variable is actually used for computation by namelist, set to `spval` while adding it to the history list. This was done inconsistently and resulted in ELM not being able to be run in debug mode (issue #4820).  The current thinking is that `nan` can be used by developers as a potential debugging tool but shouldn't be hardcoded by default as it just introduces more room for error and complications with marginal benefit.

rgknox: Thanks for that explanation @peterdschwartz

glemieux:
@peterdschwartz just a heads up that I attempted to run land developer tests on summit since perlmutter is down, but I ran into a model build issue.  Since this didn't come up previously, I'm guessing its a compiler-specific issue that I'll need to address.

glemieux:
The issue I was seeing on summit has been resolved with https://github.com/E3SM-Project/E3SM/pull/5369/commits/c94350cd0a17ea67f27da36338b5cee372e0ec4c.  The problem was that the ibm compiler didn't like that I wasn't setting a return value for two empty functions.  For posterity the form of the error was: 
```
"/autofs/nccs-svm1_home1/glemieux/e3sm/components/elm/src/biogeochem/FATESFireNoDataMod.F90", 1513-083 (E) Internal or module function lnfm24 was not set within the function.
```

The issue here is that the function is the intentionally empty form of a polymorphic procedure that is defined to end the run with an error message.  It's setup this was to make sure that the call to the function isn't applied in an instance where it is not allowed.  Intel and gnu compilers didn't seem to have a problem with this.  

Final land developer testing is underway again.

glemieux:
Land developer tests are complete.  Comparison against `master` on perlmutter is B4B with the exception of the expected NLCOMP differences for the fates tests.  I ran the tests on summit as well, although I wasn't able to compare it against the master baseline for some reason.

Folder on perlmutter: `/pscratch/sd/g/glemieux/e3sm-tests/pr5369-ibmfix.fates.pm-cpu.gnu.Cc94350cd0a-Fb8b905cc`
Folder on summit: `/gpfs/alpine/cli115/scratch/glemieux/e3sm-tests/pr5369-ibmfix.fates.summit.ibm.Cc94350cd0a-Fb8b905cc`

peterdschwartz:
merged to next

peterdschwartz:
merged to master

grnydawn:
This PR is already merged but I put the following bug report from one of ORNL helpdesk team in regards to this issue:

The `deallocate(nextc)` call throws the error because `nextc` is an associated pointer and was never allocated itself. I was using the `ERS_Ld20.f45_f45.IELMFATES.crusher_crayclang.elm-fates` test case to trigger this. The `$E3SM/components/elm/src/external_models/fates/biogeochem/EDCohortDynamicsMod.F90` file is what I was looking at, as well. The following lines from that file are relevant:
subroutine fuse_cohorts(currentSite, currentPatch, bc_in)
...
 type (ed_patch_type), intent(inout), target :: currentPatch
...
 type (ed_cohort_type) , pointer :: nextc
...
 do while(iterate == 1)
   currentCohort => currentPatch%tallest
...
   do while ( .not.associated(currentCohort,currentPatch%shortest) )
     nextc => currentPatch%tallest
     do while (associated(nextc))
...
       ! deallocates the members of nextc
       call DeallocateCohort(nextc)
       deallocate(nextc) -- this fails!

rgknox:
Thanks for posting @grnydawn 

This seems related to this previous issue you created: https://github.com/E3SM-Project/E3SM/issues/5001

Some compilers were having trouble allocating and deallocating these derived types, one of the reasons was the "target" designation. I thought we implemented a fix, in a more updated version of the code below, we don't have the target designation.

https://github.com/NGEET/fates/blob/main/biogeochem/EDCohortDynamicsMod.F90#L710-L712

Have you tried your tests with the current master version of E3SM?  That should point to a more modern fates tag.

grnydawn:
@rgknox , No, I have not tried the tests with the latest master branch recently. I will add comment here if I still have this issue.

PR #5371, opened 2022-12-22
Replace pointers to integer scalars to fix OpenACC issues

### LINKED ISSUE #5366: mpas use of integer scalar pointers results in bad addresses in OpenACC GPU code
With the Cray compiler on Crusher, the use of pointers to point to integer scalars in pool retrievals is resulting in invalid addresses when used as an array index (eg indexT, indexS) on the GPU side within OpenACC regions. This was seen a few years ago in pgi (Nvidia) compilers and had been fixed in some places but not all.

Solution is to copy into a normal integer scalar before use and a PR will follow tomorrow to fix this.

With this change, the OpenACC version builds/runs on Crusher (though slowly - different problem).

philipwjones:
Pointers to integer scalars in MPAS retrievals were not being treated correctly in OpenACC for Frontier/Crusher and have often caused problems in the past with other compilers. These are now replaced in most cases by integer scalar variables, especially for tracer indices.

This has been tested on Summit, Crusher and passes the compass pr test suite with Intel on Chrysalis.

[bfb]
Fixes #5366

### DIFF
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_rk4.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_rk4.F
@@ -140,10 +140,6 @@ module ocn_time_integration_rk4
       logical, pointer :: config_zero_drying_velocity
       character (len=StrKIND), pointer :: config_land_ice_flux_mode
 
-      ! State indices
-      integer, pointer :: indexTemperature
-      integer, pointer :: indexSalinity
-
       ! Diagnostics Indices
 
       ! Mesh array pointers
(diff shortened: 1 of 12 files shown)


mark-petersen:
On chrysalis, passes mpas-o standalone nightly suite with gnu and compares bfb against master branchpoint for all tests.

mark-petersen (approved):
Looks good by visual inspection. On chrysalis, passes mpas-o standalone nightly suite with intel debug and intel optimized and compares bfb against master branchpoint for all tests, for both.

jonbob:
passes:
* ERS_D.f09_g16_g.MALISIA.chrysalis_intel
* SMS_Ld1.ne30pg2_EC30to60E2r2.WCYCLSSP370.chrysalis_intel.allactive-wcprodssp

merged to next

jonbob:
merged to master

PR #5373, opened 2022-12-23
Update module versions for pm-cpu and pm-gpu after machine maintenance

ndkeen:
After maintenance, NERSC not only added new module versions, but also deleted previous versions that E3SM was using. Therefore, this is necessary to use the machine.

I tested e3sm_developer on pm-cpu and one basic test on pm-gpu.

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -237,10 +237,10 @@
         <command name="load">craype-accel-host</command>
         <command name="load">cray-libsci</command>
         <command name="load">craype</command>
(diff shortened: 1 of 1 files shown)


brhillman (approved):
Thanks for taking care of this so quickly Noel!

ndkeen:
merged to next

ndkeen:
Not sure why the tests are not running and reporting on cdash for pm-cpu, but might merge to master anyway

ndkeen:
I tested e3sm_developer with next and results were expected. Will merge to master.

Issue #5327, opened 2022-11-23
Ref profiles code in F and CXX


oksanaguba:
This is probably not something to fix, but to note: 

There is code for ref. profiles in both F and CXX, but the operational code uses only F ref. profiles via
```
void init_reference_states_c (CF90Ptr& elem_theta_ref_ptr,
                              CF90Ptr& elem_dp_ref_ptr,
                              CF90Ptr& elem_phi_ref_ptr)
...
```
while CXX computes ref. profiles in RefStates.compute(), which is not used in the main code. Two unit tests use only part of CXX code
```
./elem_ops_ut.cpp:        elem_ops.compute_theta_ref(kv,p,theta);
./hv_ut.cpp:            elem_ops.compute_theta_ref(kv,buf_m,theta);
```
so, the init in F and in CXX for the two tests differ (which is undesirable).

PR #5379, opened 2023-01-05
Sync reference profiles between F90 and Cxx

tcclevenger:
Previously, some unit tests (`hv_ut.cpp:hypervis` and `elem_ops_ut.cpp:set_theta_ref`) were computing reference profiles via `RefStates::compute()` and `ElementOps::compute_theta_ref()`, however this C++ code for reference profiles is never used in the main code, F90 code in `src/theta-l/` is used.

My editor removes trailing whitespaces from all changed files, recommend using the "Hide whitespaces" option.

Details of changes:

- Updates to current code
  - In `src/theta-l/share/model_init_mod.F90`, separate the ref states computation into it's own function.
  - In `hv_ut.cpp:hypervis` test, ref states were being computed unnecessarily. Optimize test by removing unneeded computation.
  -  A couple of places in C++ code contained a copy of `compute_hydrostatic_p()` computation instead of using the function. (unrelated to this PR)
- Add linking functions from C++ to F90 for computing ref states in `hv_ut` instead of calling C++ versions
- Delete C++ version of ref states computation
  - Remove `RefStates::compute()` and `ElementOps::compute_theta_ref()`
  - Remove `elem_ops_ut.cpp:theta_ref test`
  - Update initial "random" guess in `hv_ut.cpp:hypervis` test to mimic computation of `theta_ref`. Does not need to be BFB to F90 version and not used anywhere else, so no need to be in it's own `ElementOps` function, IMO.

Addresses #5327 .

### DIFF
--- a/components/homme/src/theta-l/share/model_init_mod.F90
+++ b/components/homme/src/theta-l/share/model_init_mod.F90
@@ -18,7 +18,7 @@ module model_init_mod
   use hybvcoord_mod, 	  only: hvcoord_t
   use hybrid_mod,         only: hybrid_t
   use dimensions_mod,     only: np,nlev,nlevp
-  use element_ops,        only: set_theta_ref
+  use element_ops,        only: initialize_reference_states
   use element_state,      only: timelevels, nu_scale_top, nlev_tom
   use viscosity_mod,      only: make_c0_vector
   use kinds,              only: real_kind,iulog
@@ -28,9 +28,9 @@ module model_init_mod
   use physical_constants, only: g, TREF, Rgas, kappa
   use imex_mod,           only: test_imex_jacobian
   use eos,                only: phi_from_eos
- 
+
   implicit none
-  
+
 contains
 
   subroutine model_init2(elem,hybrid,deriv,hvcoord,tl,nets,nete )
@@ -49,12 +49,12 @@ contains
     real (kind=real_kind) :: ptop_over_press
 
 
-    ! other theta specific model initialization should go here    
+    ! other theta specific model initialization should go here
     do ie=nets,nete
        gradtemp(:,:,:,ie) = gradient_sphere( elem(ie)%state%phis(:,:), deriv, elem(ie)%Dinv)
     enddo
     call make_C0_vector(gradtemp,elem,hybrid,nets,nete)
-    
+
     do ie=nets,nete
       elem(ie)%derived%gradphis(:,:,:) = gradtemp(:,:,:,ie)
       ! compute w_i(nlevp)
@@ -72,32 +72,16 @@ contains
       enddo
 
       ! initialize reference states used by hyberviscosity
-      ps_ref(:,:) = hvcoord%ps0 * exp ( -elem(ie)%state%phis(:,:)/(Rgas*TREF)) 
-      do k=1,nlev
-         elem(ie)%derived%dp_ref(:,:,k) = ( hvcoord%hyai(k+1) - hvcoord%hyai(k) )*hvcoord%ps0 + &
-              (hvcoord%hybi(k+1)-hvcoord%hybi(k))*ps_ref(:,:)
-      enddo
-      call set_theta_ref(hvcoord,elem(ie)%derived%dp_ref,elem(ie)%derived%theta_ref)
-      temp=elem(ie)%derived%theta_ref*elem(ie)%derived%dp_ref
-      call phi_from_eos(hvcoord,elem(ie)%state%phis,&
-           temp,elem(ie)%derived%dp_ref,elem(ie)%derived%phi_ref)
-      if (hv_ref_profiles==0) then
-         ! keep PHI profile, but dont use theta and dp:
-         elem(ie)%derived%theta_ref=0
-         elem(ie)%derived%dp_ref=0
-      endif
-      if (hv_ref_profiles==1) then
-         ! keep all profiles
-      endif
-      if (hv_ref_profiles==2) then
-         ! keep only dp_ref
-         elem(ie)%derived%theta_ref=0
-      endif
+      call initialize_reference_states(hvcoord, elem(ie)%state%phis, &
+                                       elem(ie)%derived%dp_ref,      &
+                                       elem(ie)%derived%theta_ref,   &
+                                       elem(ie)%derived%phi_ref)
 
       if (hv_theta_correction/=0) then
          ! compute weak laplace of p_ref
+         ps_ref(:,:) = hvcoord%ps0 * exp ( -elem(ie)%state%phis(:,:)/(Rgas*TREF))
          do k=1,nlev
-            ! compute lap(p) for z surface correction                                              
+            ! compute lap(p) for z surface correction
             temp(:,:,k) = hvcoord%hyam(k)*hvcoord%ps0 + hvcoord%hybm(k)*ps_ref(:,:)
             elem(ie)%derived%lap_p_wk(:,:,k)=laplace_sphere_wk(temp(:,:,k),deriv,elem(ie),&
                  var_coef=.false.)
@@ -105,7 +89,7 @@ contains
       endif
 
 
-    enddo 
+    enddo
 
 
     ! unit test for analytic jacobian and tri-diag solve used by IMEX methods
@@ -113,8 +97,8 @@ contains
          call test_imex_jacobian(elem,hybrid,hvcoord,tl,nets,nete)
 
     !$omp master
-    ! 
-    ! compute scaling of sponge layer damping 
+    !
+    ! compute scaling of sponge layer damping
     !
     nlev_tom=0
     if (hybrid%masterthread) write(iulog,*) "sponge layer nu_top viscosity scaling factor"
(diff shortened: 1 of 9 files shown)


tcclevenger:
@oksanaguba I've tested BFB of the ne2 thetal f1 test for both kokkos and non-kokkos version on my machine. Needed since I moved around some stuff in the model init code. Currently building on weaver so I can test on a GPU.

tcclevenger:
Update: passing kokkos vs non-kokkos test and hv_ut, elem_ops_ut on weaver.

oksanaguba:
Please, if you haven't yet, test whole theta-f* set for bfb on weaver plus all unit tests on weaver (i read your comment as you ran only a few of those), thanks!

tcclevenger:
@oksanaguba absolutely! Is there a what to build baselines for only the theta tests?

oksanaguba:
i just scanned your diffs (will read more carefully later). you change F code (the one in theta-l, not in theta-l_kokkos, which would be ok), i did not expect that. my first reaction would be to ask you to not to touch F code, i am not sure we have tests for all functionality of reference profiles for F. 

also, there are a lot of formatting changes, i would suggest reverting those, if you can do it easily, esp. if you touch F code that is not fully tested.

i should think more about what's tested in F code in standalone homme.

thanks!

tcclevenger:
@oksanaguba Maybe this is why the `compute()` function existed, since there is no easy way to ask fortran to compute the reference profiles. That is why I added them in a subroutine I could link to C++ code. 

Let me spend some time to look at the `model_init2()` subroutine, maybe I can just use the whole routine and not need to change anything inside.

components/homme/src/theta-l/share/model_init_mod.F90
```diff
@@ -72,49 +72,33 @@ subroutine model_init2(elem,hybrid,deriv,hvcoord,tl,nets,nete )
...
-      if (hv_ref_profiles==2) then
-         ! keep only dp_ref
-         elem(ie)%derived%theta_ref=0
-      endif
+      call initialize_reference_states(hvcoord, elem(ie)%state%phis, &
+                                       elem(ie)%derived%dp_ref,      &
+                                       elem(ie)%derived%theta_ref,   &
+                                       elem(ie)%derived%phi_ref)
 
       if (hv_theta_correction/=0) then
```
oksanaguba, line 80: minor question -- why not include this code in the routine above?

oksanaguba: wouldn't this be needed in unit tests if related code in biharmonic_wk_theta routine is on?

oksanaguba:
Right now we have hv_theta_correction=0,1 and hv_ref_profiles=0,1,2 . i do not see that non-default (zeros) were tested even in F code. Which options do we want and so, which ones should be tested? @mt5555

oksanaguba:
it seems that this option covers all options for _profiles:
```
   if (hv_ref_profiles == 1) then
   ! keep all profiles
   endif
```

oksanaguba:
@tcclevenger just a side note that it seems that unit tests use only default options for hv_theta_correction and hv_ref_profiles.

oksanaguba:
@tcclevenger  re: `My editor removes trailing whitespaces from all changed files, recommend using the "Hide whitespaces" option.` -- do you mean that you would leave space diffs in this PR?

oksanaguba:
At the moment, i think that 
1) we need to decide which options in F code regarding _profiles and theta_correction need to be protected by tests. 
2) Then we can introduce these tests into F code (into the same set that runs cxx, with possibility that F-vs-cxx fails, if hv_theta_correction is not in cxx) in a different PR (PR #1). 
3) After #PR 1 is merged,  we can get this PR merged. 

i find it challenging to sort through homme unit tests and figure out what they test and what they miss. I would suggest making sure that tests with new options are sensitive to new options instead. Though i just realized, we might need topo for such sensitivity?

mt5555:
It should be sufficient to test hv_ref_profiles=1.   (changing to 6 when that PR is merged).   The code path is always the same and the various options just use different initializations.  For testing, we want to avoid zero reference profiles, since that could easily produce the same answer as buggy code.  But hv_ref_profiles=1 will have all profiles non-zero and thus give complete hyperviscosity code coverage.

tcclevenger:
Thanks for the comments! I'm starting jury duty today, I think I could be tied up anywhere from a day to a week or 2. I'll try and find some time to answer the questions here and push some changes this week.

oksanaguba:
@mt5555 what about hv_theta_correction=0,1 ? do we need both options or only one?

components/homme/src/theta-l/share/model_init_mod.F90
mt5555: the code related to "hv_theta_correction" is unrelated to this PR (and not coded in EAMxx).  It should be removed or tested (on the Fortran side), but lets not delay this PR for that.

tcclevenger: The above is computing just the reference profiles, where below is computing `lap_p_wk`. Are these related?

tcclevenger:
> @tcclevenger re: `My editor removes trailing whitespaces from all changed files, recommend using the "Hide whitespaces" option.` -- do you mean that you would leave space diffs in this PR?

I would prefer to keep them removed, but I can always go through and add back the white spaces if you prefer.

oksanaguba:
@tcclevenger do not add back white spaces, i found the checkmark to view without them.

oksanaguba:
https://github.com/E3SM-Project/E3SM/pull/5401 adds a test for ref. profiles. Mark wants this PR as is, so after #5401  is merged, you'd need to merge master into this PR and test it to make sure theta-fdc12-test21 is not broken on cpu and gpu. then i will integrate. thanks!

tcclevenger:
Thanks, @oksanaguba! I'll monitor the other PR and test again once it is merged.

mt5555:
minor detail, but in the new subroutine,  initialize_reference_states(), none of the if and do loops are indented?

mt5555 (approved):

tcclevenger:
@mt5555 I think it's nicer when you can't tell what is in the loop.. No, I'll definitely fix that.

Also, I'm noticing that the comment is off 
```
if (hv_ref_profiles == 2) then
     ! keep only dp_ref
     theta_ref = 0
 endif
```
Which is it, `hv_ref_profiles = 2 -> theta_ref = 0` or `hv_ref_profiles = 2 -> theta_ref = phi_ref = 0`?

mt5555:
the comment is confusing.  Lets just delete that block of code ( no reason to keep hv_ref_profiles=2)

tcclevenger:
@mt5555 Do you plan to add `hv_ref_prof=6` here?

mt5555:
I made a branch from your branch:
https://github.com/mt5555/E3SM/tree/sync_reference_profiles
it will change 1 test (requiring a bless), so if your PR is BFB, we should merge this as a second PR

oksanaguba:
@tcclevenger could you please merge oksanaguba/homme/hvprof-test and re-rest this PR on weaver? i will test on chrysalis. Sorry for the short notice, i want to merge it today. Thanks!

tcclevenger:
@oksanaguba struggling to get a weaver node right now. I'll keep trying.

tcclevenger:
@oksanaguba I tested `theta-fdc12-test21-ne2-nu3.4e18-ndays1_cxx_vs_f90` and `hv_ut` (the test this PR changes) on weaver. Both passed.

oksanaguba:
thanks -- yes, 2 tests in pattern theta-fdc12-test21.*ne2 out of 3 will fail wrt previous baselines (but they still have to run).

tcclevenger:
@oksanaguba yes, the tests ran without error. And the cxx_vs_f90 passes, so should be good to go.

oksanaguba:
i am getting this
```
The following tests FAILED:
         55 - theta-fhs2-ne2-nu3.4e18-ndays1 (Failed)
         56 - theta-fhs2-kokkos-ne2-nu3.4e18-ndays1 (Failed)
         59 - theta-fdc12-test21-ne2-nu3.4e18-ndays1 (Failed)
         60 - theta-fdc12-test21-kokkos-ne2-nu3.4e18-ndays1 (Failed)
         94 - theta-fhs2-ne6-nu1.25e17-ndays1 (Failed)
         95 - theta-fhs2-kokkos-ne6-nu1.25e17-ndays1 (Failed)
         98 - theta-fdc12-test21-ne6-nu1.25e17-ndays1 (Failed)
         99 - theta-fdc12-test21-kokkos-ne6-nu1.25e17-ndays1 (Failed)
```
for chrysalis bfb suite. it seems that your branch does not have some nl changes for the limiter (and i did not test it merged to next). HOMMEsuite and e3sm_developer look ok. 

i will merge it. next time i would ask you to test all theta-f* tests on weaver (did not think about it earlier). thanks!

tcclevenger:
Thanks @oksanaguba!

oksanaguba:
@tcclevenger while we are still waiting for an email about repos' status, can you please test your latest version of this branch on weaver for all theta-f* tests and unit tests? since last time you ran only 1 test? thanks!.

oksanaguba:
@tcclevenger  i see you force-pushed to this branch. now i am not sure whether to re-test the latest commit on chrysalis or not.

tcclevenger:
@oksanaguba sorry, I was updating to master for my tests. I meant to do this on a test branch to not update this one, but realized I just did it here. I'm testing against baseline for weaver, maybe that is enough?

oksanaguba:
thanks, in general, i think it is preferred to avoid force pushes, esp. for branches that multiple people use. i won't retest then, thanks.

tcclevenger:
@oksanaguba `theta-f*ne2` tests all pass. I'm building ne6 baselines, but I don't see anything in the PR that would affect ne6 different. But if you want to wait for those, I'll let you know when they finish.

oksanaguba:
thanks! i think ne2 is enough, i will merge it now.

oksanaguba:
in next

PR #5385, opened 2023-01-09
Fix lunit restart error

tktesfa2014:
The changes in this branch have fixed an error related to the land unit fraction coming from the model restart file. 

[BFB]
would be non-BFB for transient runs with topo units but there are none of those.

### DIFF
--- a/components/elm/src/main/subgridRestMod.F90
+++ b/components/elm/src/main/subgridRestMod.F90
@@ -476,6 +476,11 @@ contains
          dim1name='landunit',                                                      &
          long_name='landunit weight relative to corresponding gridcell',           &
          interpinic_flag='skip', readvar=readvar, data=lun_pp%wtgcell)
+    
+    call restartvar(ncid=ncid, flag=flag, varname='land1d_wttopounit', xtype=ncd_double, &
+         dim1name='landunit',                                                            &
+         long_name='landunit weight relative to corresponding topounit', units='',         &
+         interpinic_flag='skip', readvar=readvar, data=lun_pp%wttopounit)
 
     call restartvar(ncid=ncid, flag=flag, varname='cols1d_wtxy', xtype=ncd_double,  &
          dim1name='column',                                                         &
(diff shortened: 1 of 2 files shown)


hydrotian (approved):

tktesfa2014:
@bishtgautam Could you please check this PR and let me know if you have any questions?
Thanks,

bishtgautam (approved):

bishtgautam:
@rljacob This PR adds a new variable to the restart file and nightly testing will report DIFFs (I think) because the new variable would be missing in the baseline. Should this PR be marked as non-BFB?

rljacob:
No.  we want non-BFB in a commit to mean that answers changed in some way.   Also I don't believe restart files are checked in the testing.   

Will this  mean the model can't use previous restart files after this is merged?

bishtgautam:
This PR will break the back compatibility of restart files.

rljacob:
Well that's not disallowed but it needs to be socialized more because people don't expect it.   There was a long discussion on #e3sm-help about a 2.1 run that couldn't read 2.0 restarts.

I'd suggest sending an email to e3sm-all saying this is about to happen and see if anyone objects.

bishtgautam:
@tktesfa2014 Can you confirm that this PR will break the backward compatibility of the older restart files? If yes, is it possible to add additional code to maintain the backward compatibility of the restart file if topo-units are not turned on, as most runs have topo-units turned off?

rljacob:
@tktesfa2014 please reply.

rljacob:
Ping @tktesfa2014

tktesfa2014:
@rljacob Sorry for the delay. I was not getting the messages to my email.

tktesfa2014:
@bishtgautam I need to double check to confirm this.

rljacob:
Discussion:  We know topo unit will break restart compatibility with 2.0.  But master already can't use 2.0 restarts.  This fixes restarts for transient runs with topo units.

PR #5387, opened 2023-01-10
Allow usage of specialized cores on Crusher

amametjanov:
Slurm's default `--core-spec, -S` was changed from 0 to 8 on 2022-12-29
reducing node's cores from 64 to 56. This adjusts MAX_MPITASKS_PER_NODE to 56
and adds an option to set a different number of available specialized cores: 0, 1, ... or 8.

[BFB]


ambrad (approved):

sarats:
From our conversation with Cray, these cores are reserved for OS tasks to reduce noise/jitter at scale. We need to experiment if this hurts us at scale. Bottomline: There are 56 physical cores available for application by default Slurm configuration.

sarats:
A note: this Slurm change mirrors the configuration on Frontier. So, we need to assess impact before deciding which option to use.

amametjanov:
Default MAX_MPITASKS_PER_NODE is now 56.
In a 4-node run of `PFS_Ld10.ne30pg2_EC30to60E2r2.WCYCL1850.crusher_crayclang.bench-noio`:
- 64 ppn shows 158.290 secs/day
- 56 ppn -- 145.332 secs/day
- 63 ppn -- 141.479 secs/day

Different grid+compset cases can adjust `--core-spec={0..8}` with different MAX_MPITASKS_PER_NODE in `cime_config/allactive/config_pesall.xml` and other PE config files.

sarats:
Thanks for running the benchmark. It's interesting to see the impact even at four nodes.
cc @twhite-cray @abbotts

Default of 56 sounds good.

sarats (approved):

sarats:
FYI @grnydawn

grnydawn:
@amametjanov I have one question about the walltimemax in "config_batch.xml". You set it to 1 hour for all three queues as shown below.

       <queue strict="true" nodemin="1" nodemax="8" walltimemax="01:00:00" default="true">batch</queue>
       <queue strict="true" nodemin="9" nodemax="64" walltimemax="01:00:00">batch</queue>
       <queue strict="true" nodemin="65" nodemax="160" walltimemax="01:00:00">batch</queue>

While longer walltimemaxes are allowed on Crusher like below.

       <queue strict="true" nodemin="1" nodemax="8" walltimemax="08:00:00" default="true">batch</queue>
       <queue strict="true" nodemin="9" nodemax="64" walltimemax="04:00:00">batch</queue>
       <queue strict="true" nodemin="65" nodemax="160" walltimemax="02:00:00">batch</queue>

Do you have any reasons for setting it to 1 hour? I got two tests failed due to the timelimit of 1 hour.

grnydawn:
@amametjanov @sarats @ambrad @rljacob , In addition to Azamat's PR, I added two more changes

1. changed walltimemax for Crusher from 1 hour to 8hours, 4 hours, or 2 hours according to the # of allocated nodes.
2. changed amd compiler version from 5.1.0 to 5.4.0. The new version is required to fix the long-command-line issue

I think those two updates should not change any science so I think I will merge it once a test is completed.

sarats:
Just thinking out loud, the default MAX_MPITASKS_PER_NODE would be 56. However, if someone changes it after creating a case and increases it, then the associated Slurm flag (-S ) is modified during case submission. Is that understanding correct?

oksanaguba:
i had a similar question -- scream right now uses not -c7 flag, but -S0, which way is preferable?

mt5555:
@oksanaguba : scream probably should be using -S8, not -S0.    Per our conversation with HPE, this is to help MPI performance on large node counts.  So better to use the recommended value unless we have data to suggest -S0 is better?

PR #5391, opened 2023-01-11
Allow to build E3SM against a pre-built kokkos installation

bartgol:
Allows to build compsets that require kokkos without building our internal submodule, and use an existing kokkos installation instead.

The behavior is triggered by the presense of the `KOKKOS_PATH` env variable. If present, we _assume_ that it points to a valid kokkos installation; we check that the folder exists and contains at least the `Kokkos_Core.hpp` header, crapping out if that's not true (portably checking for the libs is harder, since the lib install folder is likely to be arch-dependent).

This helps with MALI compsets, which link against trilinos, which already contains a kokkos installation. If the `KOKKOS_PATH` env var is not present, the behavior is identical to what it was before.

@stephenprice  this is the branch that @mperego and I worked on to get MALI working on Cori.

@rljacob to trigger the behavior I use the env var approach, as that is already done with PIO (in a slighlty different way). I am not super fond of this approach, and would prefer being able to trigger this via some XML change, but I'm not really sure how to do it or where to start. However, if this is the preferred way for CIME/E3SM to handle this kind of scenario, I'm fine with it.

### DIFF
--- a/components/cmake/common_setup.cmake
+++ b/components/cmake/common_setup.cmake
@@ -239,29 +239,28 @@ if (USE_ALBANY)
 endif()
 
 if (USE_KOKKOS)
-  # LB 09/04/20
-  #  The best thing to do would be to simply use ${INSTALL_SHAREDPATH}, and
-  #  let cmake search the proper subfolder based on system/cmake configuration.
-  #  True. HOWEVER, the subfolder `kokkos` in that directory (where kokkos is
-  #  built) happens to contain KokkosConfig.cmake, and this is picked up by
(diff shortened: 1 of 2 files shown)


bartgol:
e3sm_developer passes on Chrysalis

jgfouca:
@bartgol , looks like CI testing is not passing for this PR.

bartgol:
> @bartgol , looks like CI testing is not passing for this PR.

Ah, I messed up and pushed the last commit to the e3sm repo rather than my fork. Should work now.

jgfouca:
Merged to next.

PR #5399, opened 2023-01-13
Clean up MMF radiation interface and change order of merging GCM/CRM columns

brhillman:
Clean up MMF radiation interface for improved performance. This primarily moves the merging of GCM/CRM columns into  a single index *before* computing cloud optical properties, rather than after. This improves performance in a couple ways. First, we no longer call the cloud optical property routines in a loop over CRM columns, but we can instead call them only once with a much larger column dimension. This will become even more important when we switch to a GPU-capable optics code (to be done in a subsequent PR). Second, we no longer need to pack the very large cloud optical property arrays inside this loop. This turned out to be a pretty expensive operation. The combination of these two things seemed to net about a 40% reduction in the overall radiation call in ne30 tests on Summit. Despite the refactor, this is overall BFB in terms of the climate, but the FRCE-MMF1 test will show a DIFF because this also removes call to output the cloud optical properties on the GCM grid, which were actually not being output correctly before (it was only outputting the last CRM column). This shows up as a DIFF in the cprnc output because the fill patterns for TOT_ICLD_VISTAU end up being different. Outputting a better representation of `TOT_ICLD_VISTAU` would require computing averages from the packed cloud optical properties, and since we are not using the output from this field anyways this can be fixed in a subsequent PR if needed.

[non-BFB] (only for the FRCE-MMF1 test)

### DIFF
--- a/components/eam/src/physics/crm/rrtmgp/radiation.F90
+++ b/components/eam/src/physics/crm/rrtmgp/radiation.F90
@@ -1404,7 +1404,7 @@ contains
                do iy = 1,crm_ny_rad
                   do ix = 1,crm_nx_rad
                      ! Fill GCM columns with CRM data
-                     call t_startf('rad_overwrite_state')
+                     call t_startf('rad_set_state')
                      do iz = 1,crm_nz
                         ilev = pver - iz + 1
                         do icol = 1,ncol
@@ -1420,10 +1420,7 @@ contains
                            end if
                         end do  ! icol = 1,ncol
                      end do  ! iz = 1,crm_nz
-                     call t_stopf('rad_overwrite_state')
-
                      ! Setup state arrays, which contain an extra level above model top to handle heating above the model
-                     call t_startf('rad_set_state')
                      call set_rad_state(                                            &
                         state                      , cam_in                       , &
                         tmid_col(1:ncol,1:nlev_rad), tint_col(1:ncol,1:nlev_rad+1), &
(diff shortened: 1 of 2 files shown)


brhillman:
Comparison on summit gpu, before and after refactor, ne30pg2 with a 192x7 pelayout:
```
"a:radiation"                             -        192     1344 9.703680e+05   9.409692e+04    80.492 (   145      2)    37.519 (    43      3)
"a:radiation"                             -        192     1344 9.703680e+05   5.517772e+04    46.163 (    13      1)    27.910 (    35      6)
```
This is almost a 43% reduction in the wallmax timer, and a 41.4% reduction in the walltotal timer.

brhillman:
Some more comparisons, on pm-cpu with 16 nodes (1024x1):
```
"a:radiation"                             -       1024     1024 7.393280e+05   1.015402e+05   105.248 (   221      0)    88.496 (   493      0)
"a:radiation"                             -       1024     1024 7.393280e+05   8.952396e+04    91.062 (   949      0)    73.782 (   156      0)
```

components/eam/src/physics/crm/rrtmgp/radiation.F90
```diff
@@ -1438,8 +1363,8 @@ subroutine radiation_tend(state_in,ptend,    pbuf,          cam_out, cam_in,  &
                ! do the shortwave radiative transfer during the daytime to save
                ! computational cost (and because RRTMGP will fail for cosine solar zenith
                ! angles less than or equal to zero)
+               day_indices_packed = 0
                call set_daynight_indices(coszrs(1:ncol), day_indices(1:ncol), night_indices(1:ncol))
-               nday = count(day_indices(1:ncol) > 0)
```
whannah1, line 1442: Can you help me understand why you deleted "nnight" in the shortwave driver but deleted "nday" here?

whannah1 (approved):
Really nice clean up! I don't see any potential issues, just asked for clarification on the use of day vs night column count.

components/eam/src/physics/crm/rrtmgp/radiation.F90
brhillman: Yeah good question. I removed `nday` because it is not used in this context, and I wanted to use `nday` to represent the number of day columns in the packed version of the arrays. `nnight` is still computed here because it is used below in the aerosol optics codes...but admittedly I should have maybe moved it's initialization closer to its use below. `nnight` is not used in the sw section, so no need to include it there. It only lives for the call to the aerosol sw optics driver.

components/eam/src/physics/crm/rrtmgp/radiation.F90
```diff
@@ -1464,458 +1389,458 @@ subroutine radiation_tend(state_in,ptend,    pbuf,          cam_out, cam_in,  &
...
-               if (conserve_energy) then
+               else
+                  ! Back heating out of pdel scaled heating rate from previous step
                   qrl(1:ncol,1:pver) = qrl(1:ncol,1:pver) / state%pdel(1:ncol,1:pver)
                end if
+            end if  ! active calls
+         end do  ! loop over diagnostic calls
+
+         ! Update net CRM heating tendency
+         ! TODO: should we be updating this every timestep, even if using
```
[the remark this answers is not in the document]
brhillman: So do you think this is a bug in our current implementation that we should fix in a subsequent PR? I know we've looked at and discussed this particular issue a handful of times...

components/eam/src/physics/crm/rrtmgp/radiation.F90
whannah1: ok, it's not important, so no need to change it. It's still clear what the code is doing when you read through it.

components/eam/src/physics/crm/rrtmgp/radiation.F90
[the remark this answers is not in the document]
whannah1: I guess the real question is whether the CRM is seeing the same rad fluxes from the previous time step when the rad flux calculation is skipped... I'll need to look closer to figure that out, and maybe run a test with special output, but we never run that way in practice, so let's kick that can down the road for now.

brhillman:
Merged to next.

PR #5401, opened 2023-01-18
homme: test with hv_ref_profiles=1

oksanaguba:
Needed to test hv profiles code that is in F and cxx.

[bfb] except for homme modified tests (theta-fdc12-test21).


oksanaguba:
on chrysalis for ne2 the new test passes cxx-vs-f and also it is sensitive (fails) for the default option, 0. that is, option 1 is not bfb wrt option 1 for chrysalis and ne2.

oksanaguba:
on weaver, this test, theta-fdc12-test21, passes with ref_profiles=1 and is sensitive to it (bfb fails for ref_profiles=1 and ref_profiles=0). this is ready.

PR #5406, opened 2023-01-21
Single Column Model (SCM) code cleanup and refactor

bogensch:
This PR represents a long overdue cleanup/refactor for routines related to the single column model (SCM).  Specifically, this PR:

1)	Removes all hooks for the SCM in the Eulerian dynamical core (as if it never happened).  The SCM has not worked nor been supported with the Eulerian dynamical core in E3SM for a number of years.
2)	Renames the ill-named forecast.F90 routine to apply_iop_forcing.F90.  In addition, this routine has been substantially cleaned up, refactored, and streamlined with much unused code (from seemingly the 1950’s) removed.  apply_iop_forcing.F90 now contains two routines, one which applies the large scale forcing and the other which nudges the SCM to the IOP observations.
3)	Refactors scamMod.F90 (where IOP information is read in).  The largest change being that setiopupdate routine has been broken into two with the addition of a setiopupdate_init routine.
4)	Removes a very out-dated block of code in clubb_intr.F90 that had case specific adjustments (which were never used in E3SM SCM as the case names never matched the ones specified in the code).
5)	Makes general cleanups, purging of unused variables, sensical variable name modifications, and cosmetic changes in several routines as seen fit.

All e3sm_atm_developer tests pass on NERSC.  This PR is b4b, including the SCM.


rljacob:
@bogensch what is the status of this PR now that SCREAM has been merged?

bogensch:
I am actively working through the (expected) nasty conflicts the SCREAM downstream merge has caused.  I will reopen the PR once tests have passed and things are working as they should.

components/eam/src/control/scamMod.F90
```diff
@@ -198,9 +207,12 @@ subroutine scam_default_opts( scmlat_out,scmlon_out,iopfile_out, &
    if ( present(iop_nudge_tq_low_out) ) iop_nudge_tq_low_out = 1050.0_r8
    if ( present(iop_nudge_tq_high_out) ) iop_nudge_tq_high_out = 0.e3_r8
    if ( present(iop_nudge_tscale_out) ) iop_nudge_tscale_out = 10800.0_r8
+   if ( present(iop_perturb_high_out) ) iop_perturb_high_out = 1050.0_r8
    if ( present(scm_crm_mode_out) )     scm_crm_mode_out  = .false.
```
whannah1, line 211: I know this is off topic, but is it correct that some of these "crm" options are no longer working/supported?
I've seen other old "crm" references in the physics code, which I've never really understood how they were used. Would it be worth cleaning those up in a future PR?

bogensch: @whannah1 I was on the fence about removing those in this PR.  I have also never been clear about what these options do and have not heard of one soul that uses them in my five years of being the SCM contact-point.  I will likely reconsider removing them in this PR.

whannah1: ok sounds good!

rljacob:
@bogensch any update?

bogensch:
working on it.  trying to understand some diffs that I currently can't explain.

rljacob:
@bogensch any update?

bogensch:
I should be opening this up for review within the next day, just doing some final sanity check testing.

bogensch:
Open for review.

ambrad:
Looks like this is currently failing CircleCI with this error:
```
/home/circleci/project/components/eam/src/physics/cam/clubb_intr.F90:1100:7:

    use scamMOD,                   only: single_column
       1
Fatal Error: Can't open module file ‘scammod.mod’ for reading at (1): No such file or directory
```

components/eam/bld/namelist_files/namelist_definition.xml
```diff
@@ -4250,29 +4250,16 @@ Full pathname of IOP dataset.
...
 Default: FALSE
 </entry>
 
-<entry id="scm_clubb_iop_name" type="char*200" category="scam"
+<entry id="iop_nudge_tscale" type="real" category="scam"
        group="cam_inparm" valid_values="">
-IOP name for CLUBB running in single column mode
-Default: ""
-</entry>
+Time scale to nudge to IOP forcing file.
```
whannah1, line 4261: general comment about the use of the "IOP" terminology - it seems replacing this with "SCM" would be more general. I have a branch to add an SCM-RCE compset, and obviously an "intense observing period" doesn't make sense in that context. It's not critical that we replace the "IOP" terminology, but I think I would prefer avoiding it if the change was trivial (which it's not).

components/eam/src/control/apply_iop_forcing.F90
```diff
@@ -0,0 +1,275 @@
...
+!=========================================================================
+
+subroutine advance_iop_subsidence(scm_dt, ps_in, &                    ! In
+                             u_in, v_in, t_in, q_in, &                ! In
+                             u_update, v_update, t_update, q_update ) ! Out
+
+!-----------------------------------------------------------------------
+!
+! Purpose:
+! Option to nudge t and q to observations as specified by the IOP file
```
whannah1, line 183: This comment seems to be copied from the routine above, so it would be good to update it.

components/eam/src/control/iop_data_mod.F90
```diff
@@ -192,45 +166,35 @@ module scamMod
   logical*4, public ::  have_asdif    ! dataset contains asdif
   logical*4, public ::  scm_iop_srf_prop   ! use the specified surface properties
   logical*4, public ::  iop_dosubsidence ! compute Eulerian LS vertical advection
```
whannah1, line 168: Do we know if the subsidence option works?

whannah1 (approved):
Great work! I'm so glad to see all this stuff cleaned up. I left a couple "thinking out loud" comments, but nothing needs to be addressed for this PR.

components/eam/src/control/iop_data_mod.F90
bogensch: Yes, while there is no official unit test for this routine at this time, I have thoroughly tested/debugged this.  Many of our standard test cases (i.e. DYCOMS-RF01) rely on this routine.  Additionally, this routine really isn't new as the code was basically copied and pasted (with changed variable names) from the CAM3.5/CAM5 SCM Eulerian subsidence calculation from the old forecast.f90 module.

components/eam/src/control/apply_iop_forcing.F90
bogensch: Good find.  This has been updated.

bogensch:
Fix in clubb_intr.F90 that was causing failing test has been pushed.

components/eam/src/control/iop_data_mod.F90
```diff
@@ -380,133 +323,107 @@ subroutine scam_setopts( scmlat_in, scmlon_in,iopfile_in,single_column_in, &
...
-        endif
-     endif
-!!jt fix this for crm
-!!jt   if(scm_crm_modes) then
-!!jt      iyear_AD_out     = (base_date-mod(base_date,10000))/10000 ! year AD to calculate the orbital parameters for.
-!!jt   else
-!!jt      iyear_AD_out     = 1950
-!!jt   end if
+      if (plon /= 1 .or. plat /=1 ) then
+         call endrun('SCAM_SETOPTS: must compile model for SCAM mode when namelist parameter single_column is .true.')
```
wlin7, line 335: Minor. The text above should now be IOP_SETOPTS and SCM.

wlin7:
@bogensch , the refactored codes look great, clean and better organized. 

The PR is labeled BFB, including the SCM test. I did a test merge to next and ran some tests anyway. Non-SCM tests PASS, but `SMS_R_Ld5.ne4_ne4.FSCM-ARM97.pm-cpu_gnu.eam-scm` diff from the master baseline. Since your earlier SCM tests were BFB, it is possible the implementation of some v3atm features may be incompatible with the refactored SCM codes. However, I am not able to figure out which implementation could have such issue, if indeed any. The v3atm implementations in several occassions missed the treatment for MMF1, but never having to deal with SCM specifically.

bogensch:
@wlin7 Thanks for bringing this to my attention.  The last test I did was b4b, but clearly something has changed since then.  I will be looking into this and will report back my findings when available.

bogensch:
@wlin7 I looked into this and it turns out that some SCM cases are b4b while others are not (which includes the ARM97 case that we use for testing).  Upon further investigation, I discovered that this is expected.  This is true for SCM cases that use Temperature forcing that includes vertical large-scale transport. The reason is that in order to make the application of the IOP forcing code modular, a subtle change in the order of operations of computations was necessary. This leads to roundoff level differences.  I verified that if I restore this computation back to previous formulation (for ARM97 case) that bit-4-bit results are once again achieved.  Thus, I am comfortable with this diff now that I can explain and understand it.  I updated the text of the PR description to update this.

wlin7 (approved):

wlin7:
Thanks for the reexamination and the explanation, @bogensch . All is good.

jgfouca:
Merged to next!

PR #5410, opened 2023-01-24
add  linear reference profile for theta 

mt5555:
Minor modification for the HV reference profile for theta: use linear version near the surface

This option is to enable rougher topography, to be used in SCREAM (not used in EAM)

F case evaluation:  https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/3530719247/F2010+NH+evaluation

[BFB] except for one standalone HOMME test ([theta-fdc12-test21.nl]) which is switched to this profile and will need a new baseline

### DIFF
--- a/components/homme/test/reg_test/namelists/theta-fdc12-test21.nl
+++ b/components/homme/test/reg_test/namelists/theta-fdc12-test21.nl
@@ -27,7 +27,7 @@
   dcmip2_x_xi       = 4000.0                    ! mountain wavelength   (m)
   limiter_option    = 9
   vert_remap_q_alg  = 1
-  hv_ref_profiles   = 1
+  hv_ref_profiles   = 6
 /
 &vert_nl
   vanalytic         = 1                         ! set vcoords in initialization routine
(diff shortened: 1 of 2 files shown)


components/homme/test/reg_test/namelists/theta-fdc12-test21.nl
```diff
@@ -27,7 +27,7 @@
   dcmip2_x_xi       = 4000.0                    ! mountain wavelength   (m)
   limiter_option    = 9
   vert_remap_q_alg  = 1
-  hv_ref_profiles   = 1
+  hv_ref_profiles   = 6
```
oksanaguba, line 30: this namelist is used in the corresponding cxx test as well. i assume the new option is not in cxx code yet?

tcclevenger (approved):

components/homme/test/reg_test/namelists/theta-fdc12-test21.nl
mt5555: I thought this test checked that cxx and F were BFB?

oksanaguba: thanks -- i got confused, the profiles are now only read from F.

PR #5414, opened 2023-01-25
Add a new machine gcp12 (google cloud platform)

### LINKED ISSUE #5412: Add a new GCP machine gcp12 (and rename previous to gcp10)
After creating a new GCP virtual machine using newer hardware, I have been testing.
I wanted to have both clusters working for a short period of time, and then delete the old one.
To do so, it's better to rename the old one `gcp10` so it's not confused with `gcp12`.
After changes in https://github.com/E3SM-Project/E3SM/issues/5408
to get 3 DEBUG tests to pass, I see all e3sm_integration tests pass.
For e3sm_extra_coverage, the same tests that pass on gcp10 also pass on gcp12.

I added `--kill-on-bad-exit` to srun command for both clusters as I had a few cases that were hanging.
This flag prevented the hang on at least gcp12, so adding it to gcp10 as well as we certainly don't want hangs.

With branch `ndk/machinefiles/add-gcp12-machineb`, tests should works on both gcp10/gcp12

Note the test `ERS.ne30pg2_r05_EC30to60E2r2.GPMPAS-JRA.gcp12_gnu.mosart-rof_ocn_2way` still fails as it does on gcp10 as we have the same issue -- the compset requires a large amount of data to be downloaded to inputdata (while only using a small amount), so we've prevented writing to that specific directory which allows other tests to pass, and this one to fail quickly until resolved. https://github.com/E3SM-Project/E3SM/issues/5349

ndkeen:
Add a new machine gcp12, which is another google cloud cluster.
Renaming the previous GCP cluster from gcp to gcp10. Adding a `--kill-on-bad-exit` option to srun for both GCP machines, as it prevented hanging on at least one occasion.

Fixes #5412 
[bfb]


sarich:
I'm still having a problem with the gcc module on gcp12, I need to run `. /opt/apps/spack/share/spack/setup-env.sh` in my shell before running create_test. Can you try adding `<command name="use">/opt/apps/spack/share/spack/modules/linux-centos7-x86_64_v3</command>` to the config file and make sure it doesn't break anything for you?

gcp10 is working fine

ndkeen:
OK I just made another commit to add the line you suggested.
And then I also adjusted PE layouts for gcp12 to improve performance/efficiency.

sarich (approved):

ndkeen:
merged to next.  I'm not sure how the tests are launched and recvd by cdash -- it might be that the machine rename from gcp -> gcp10 causes an issue?

ndkeen:
After changes for the new machine name, the tests are passing on CDASH as they did before (all pass with integration and the same 3 fails with extra_coverage).  The machine name is now gcp10 on cdash, and we still have old gcp entries.

ndkeen:
We have gcp10 entries in "Experimental" section of cdash and they look as expected.  And then we still have two of the older gcp entries in "Machine Coverage" on cdash.  Can we remove those older entries?  

And then can/should we add gcp12 nightly testing?

And then assuming it works for a bit, delete gcp10 altogether?

rljacob:
@jgfouca or @wadeburgess can change the cdash entries.

jgfouca:
@rljacob , @ndkeen , done.

PR #5416, opened 2023-01-25
Bring EAMxx code from SCREAM project into E3SM

brhillman:
Bring EAMxx code from SCREAM project into E3SM. While non-SCREAM source code is affected (mostly to support SCREAMv0 configurations), this should be BFB for all existing E3SM configurations *except* the single column model tests, as this merge includes two small bugfixes for SCM from SCREAM that were never implemented upstream in E3SM.

The SCM bugfixes are  1) an order of operations change on how moisture was updated after applying IOP forcing in the SCREAM code base. This was done to ensure b4b restarts for DP-SCREAM simulations (or any simulation that uses the IOP forcing routine, as SCM). This results only in roundoff level differences. 2) There was a bug fix only related to the very last time step in a SCM simulation where the surface fluxes were not being updated properly. This results only in answer changes in regards to the very last time step.

This branch was setup as follows:

git clone git@github.com:E3SM-Project/E3SM.git add-eamxx
git checkout -b brhillman/add-eamxx
git remote add scream git@github.com:E3SM-Project/scream.git
git merge scream/master

After which, unintended changes are reverted to E3SM master as needed, and other potential conflicts are resolved with subsequent commits.

Guide to changes:
cime_config:  add scream compsets and grids; add scream-specific machine configs.
components/eamxx:  add eamxx source code
components/eam:  add support for SCREAMv0 (F90) physics; changes for single column model
components/elm:  changes for single column model.  Adds the "sum_multcols" option, which is responsible for providing functionality of the single-column model but to a domain of multiple columns that are identically forced and with the same surface properties (for the purposes of a doubly periodic cloud resolving model, for example)
components/cice:  changes for single column model.  Adds the "scm_multcols" option (see above for description).
components/data_comps/docn:  configuration for SCREAM DYAMOND1 and DYAMOND2 SST forcing files
components/homme  doubly periodic changes
components/mpas-seaice: add configuration for ICOS10 grid
driver-mct:  add ATM_NCPL defaults for SCREAM compsets; add EPS_AGRID overrides for high res grids; add lnd_grid to config dict
externals:  bring in ekat, update YAKL
share/streams:  change to allow functionality of the single-column model but to a domain of multiple columns that are identically forced and with the same surface properties.  
share/utils:  Same as above, only read in namelist for master proc if in SCM mode.  (not sure what change shr_sys_mod.F90 is related to)

[non-BFB] only for SCM tests


whannah1 (approved):
Looks good!

bogensch (approved):

README.md
```diff
@@ -1,13 +1,11 @@
-[![E3SM Logo](https://e3sm.org/wp-content/themes/e3sm/assets/images/e3sm-logo.png)](https://e3sm.org)
```
ambrad, line 1: This file should not be touched.

.gitignore
```diff
@@ -15,9 +15,6 @@
 # Logs
 *.log
 
-# Ignore netCDF files.
```
ambrad, line 18: I'm guessing this file should not be touched.

ambrad:
We probably need more reviewers. This PR touches quite a lot of files outside of the eamxx directory.

cime_config/machines/Depends.crusher-gpu.crayclang.cmake
```diff
@@ -0,0 +1,43 @@
+set(CICE_F90
```
oksanaguba, line 1: i think e3sm won't use crusher-gpu machine (it is either crusher+crayclang or crusher+crayclanggpu), so this file is not needed?

related -- there are 2 files in cmake_macros, crayclang_crusher-cpu and crayclang_crusher-gpu, they are also not needed.

oksanaguba: i see that in config_machines, crusher-cpu and crusher-gpu blocks made in. they need to be removed. in scream, we are trying to adopt e3sm machines' names instead (so, crusher+crayclang and crusher+crayclanggpu).

sarats:
There seem to be 7,849 commits and changes to 960 files in this PR which is most likely unintentional?

ambrad:
@sarats this is the entire EAMxx project, so those numbers are roughly correct. The primary challenge in dealing with this PR is that things like machine file changes, SCM-related changes, and possibly other things in miscellaneous files outside of `eamxx` need to be carefully checked.

rljacob:
Why is this binary file added to the top level:  p3_lookup_table_1.dat-v4pt1pt1.tgz

README.md
rljacob: Yes it should not be.

rljacob:
@bogensch the changes to cice, elm, share/streams all seem related to scm.  You approve those?

rljacob:
@singhbalwinder and @wlin7 should review the changes to eam source code.

components/homme/src/share/namelist_mod.F90
```diff
@@ -219,9 +219,11 @@ subroutine readnl(par)
     character(len=80) :: errstr, arg
```
oksanaguba, line 219: These two files differ in e3sm and in scream, but this file needs to be reviewed before using scream version as is (i believe Andrew would know how this file should look like).

components/homme/src/theta-l_kokkos/CMakeLists.txt
```diff
@@ -11,8 +11,8 @@ MACRO(THETAL_KOKKOS_SETUP)
   SET(TEST_SRC_DIR      ${HOMME_SOURCE_DIR}/src/test_src)
```
oksanaguba, line 11: IIRC this change will break homme standalone build (and won't be caught via nightlies). Someone needs to check and fix this (or leave homme's version in e3sm for now).

oksanaguba:
Is there a way to bring all these changes but break them into smaller PRs? it seems overwhelming to review this, also, it seems possible to separate some of the changes into smaller separate sets.

components/homme/src/share/namelist_mod.F90
ambrad: Agreed. My impression is that these changes cause the upstream Homme standalone DP tests to DIFF. It's on my to-do list to find out why.

rljacob:
A more detailed PR description would help.  It should serve as a guide on what changes were done in which main directories.

brhillman:
> Is there a way to bring all these changes but break them into smaller PRs? it seems overwhelming to review this, also, it seems possible to separate some of the changes into smaller separate sets.

I don’t really agree. This PR as-is is simple: bring in the commits from the SCREAM fork. This was done simply here to try to minimize the amount of hand-editing and places to make mistakes:

```
git checkout -b brhillman/add-eamxx
git remote add scream git@github.com:E3SM-Project/scream.git
git merge scream/master
```

So I’m not bringing anything in by hand. If we instead want to “break it up” into smaller pieces, and just bring in little pieces of the EAMxx code one by one, the only way I really see to do that is to `git add` files or folders individually, which is going to lose the git history. If someone sees a better way to do this I’m happy to explore, but this seemed like the least error-prone way to do this.

.gitignore
brhillman: Fixed.

README.md
brhillman: Fixed.

oksanaguba:
Why is scream git history needed here? I must say i am not sure what's the purpose of this PR -- is it to prepare e3sm repo for scream (so that there is only 1 repo)? or is it to have some of scream source in e3sm for now (for whatever reason)? in either case, is history needed?

cime_config/machines/Depends.crusher-gpu.crayclang.cmake
brhillman: Done.

rljacob:
This PR is mainly needed now so MMF can use scream physics.  And yes it starts us on the path to 1 repo which was always the plan.  History should be kept so developers get credit for the work in the one repo.

brhillman:
@oksanaguba twofold. For one, eventually yes, I think the goal is to have a single source in a single repo. This needs to happen eventually one way or another. The reason to do it now is that MMF needs the SCREAM code in the same repo to be able to use the ported physics developed in SCREAM. That could also be done by just making a copy of the eamxx component from the scream repo, adding ekat and whatever other supporting changes are needed, and doing a different kind of massive PR with zero git history. I didn’t think that would be preferable, because I’m not sure what that would do to our downstream merges, and I would think would make it more difficult to merge future scream dev upstream into e3sm, and future e3sm dev downstream into scream. I don’t really know what the best approach is here, I’m just trying to make less headaches down the road, because the previous approach to trying to use the kokkos versions of p3 and shoc in MMF made it a lot more difficult to stay up to date with all the dev going on in scream.

mt5555:
On reason to preserve the git history is to prevent future conflicts when merging E3SM into SCREAM.   with this PR, that should theoretically be seamless, since git will know all these changes are already in SCREAM.  if we bring them in separately, then there is the possibility of a lot of conflicts the next time E3SM is merged into SCREAM.

rljacob:
What is components/eam/src/physics/cam/create_p3_lookupTable_1.f90-v4.1 doing in that directory?  If its a standalone program, I'd say it should be in eam/tools.

ambrad:
@brhillman we need to be careful about removing files such as the crusher-gpu ones. Until E3SM and SCREAM finalize treatment of these, we need those files in SCREAM.

I believe the following will currently happen: In the next upstream merge to SCREAM, your commits above will remove them from the SCREAM repo. We'd then have to revert that, which in turn would need to be reverted in the next SCREAM->E3SM merge, etc.

@oksanaguba may be able to comment on the E3SM/SCREAM Crusher consolidation status. One suggestion I have is that if consolidation is proving too difficult, we can just make some `scream_`-prefixed machine files that can go in both repos without causing problems to either.

rljacob:
@brhillman is the change to the YAKL external necessary?

ambrad:
It's possible we may need to compile a list of to-do items in an E3SM issue and take care of each one before a PR of this type can be done.

.gitmodules
```diff
@@ -32,6 +32,9 @@
 	path = cime
 	url = git@github.com:ESMCI/cime.git
 	branch = master
+[submodule "externals/ekat"]
+        path = externals/ekat
+        url = git@github.com:E3SM-Project/EKAT.git
 [submodule "externals/YAKL"]
```
rljacob, line 38: The github diff is showing a change to YAKL.  Different hash in the submodule?

components/eam/src/physics/cam/create_p3_lookupTable_1.f90-v4.1
```diff
@@ -0,0 +1,1812 @@
+PROGRAM make_p3_lookuptable1
```
rljacob, line 1: This file should be moved to cam/tools I think.  cam/src is only for source compiled in to the eam model itself.

brhillman:
> @brhillman we need to be careful about removing files such as the crusher-gpu ones. Until E3SM and SCREAM finalize treatment of these, we need those files in SCREAM.
> 
> I believe the following will currently happen: In the next upstream merge to SCREAM, your commits above will remove them from the SCREAM repo. We'd then have to revert that, which in turn would need to be reverted in the next SCREAM->E3SM merge, etc.
> 
> @oksanaguba may be able to comment on the E3SM/SCREAM Crusher consolidation status. One suggestion I have is that if consolidation is proving too difficult, we can just make some `scream_`-prefixed machine files that can go in both repos without causing problems to either.

Okay, maybe I misunderstood Oksana’s comment then. I can revert the commits I just made to remove the `crusher-*` machines.

brhillman:
> 

@rljacob the version from the scream repo is slightly newer, I think we should take that version.

.gitmodules
brhillman: Yes the version from the scream repo appears to be slightly newer, so I think this is the version we want.

cime_config/machines/Depends.crusher-gpu.crayclang.cmake
brhillman: I think based on comments below from @ambrad we should revert the commits that removed these configs.

brhillman:
> It's possible we may need to compile a list of to-do items in an E3SM issue and take care of each one before a PR of this type can be done.

I think some of these might be regarded as scream issues though, correct? E.g., the crusher conflicts and the changes to homme that would apparently result in issues upstream.

ambrad:
Ben, I agree: probably all of the mods would have to be made to SCREAM. A SCREAM issue is fine, as well. I just meant that it might not be possible to handle all the issues in this PR; we likely need to make several PRs to SCREAM first, then make a new PR of this type to E3SM.

rljacob:
Re: crusher, you can make a crusher-scream entry in the config_machines file and put off solving the crusher differences.

oksanaguba:
re crusher: i am currently working on making scream accept e3sm machine+compiler pairs for crusher. i did mean for Ben to remove those files (crusher-gpu and crusher-cpu and related entries) that exist only in scream, because i did not assume we would run scream on crusher from this repo. i was just trying to save us some movements removing these entries in the future.

oksanaguba:
i would also advocate for making this effort the same as merging two repos into one now, with the goal of having one repo, e3sm, for e3sm and scream. just to save future efforts of merging again back and forth.

components/homme/src/theta-l_kokkos/CMakeLists.txt
ambrad: @oksanaguba I ran the following script and see no issues; whatever was broken may have been fixed. Is there anything else I should do to address this concern? Thanks.
```
rm -rf CMake*
e3sm=~/SCREAMref # SCREAM master of the moment
cmake -C $e3sm/components/homme/cmake/machineFiles/chrysalis-bfb.cmake \
      -D BUILD_HOMME_THETA_KOKKOS=On \
      -D BUILD_HOMME_PREQX_KOKKOS=On \
      -D NUM_CPUS=24 \
      -D Kokkos_ENABLE_OPENMP=On \
      -D Kokkos_ENABLE_SERIAL=On \
      $e3sm/components/homme/
make -j64
ctest
```

oksanaguba: great, thanks, it works.

components/eam/src/physics/cam/create_p3_lookupTable_1.f90-v4.1
brhillman: @AaronDonahue @jgfouca and whoever else, does this ever get run automatically? Can this just be moved to a new directory or will that break any tests?

jgfouca: @brhillman , it does not look like it:

% cd scream/components
% git grep create_p3_lookup
eam/src/physics/cam/create_p3_lookupTable_1.f90-v4.1:!  Note:  compile with double-precision (pgf90 -r8 create_p3_lookupTable_1.f90)
eamxx/src/physics/p3/impl/p3_main_impl_part2.hpp:      // "create_p3_lookupTable_1.F90-v4.1".  All line numbers below reference this

Eamxx only has a comment referring to it.

components/homme/src/share/namelist_mod.F90
ambrad: homme_integration passes on Chrysalis when run from the SCREAM repo, so the HOMME DP tests that once were broken are now fine.

brhillman: Should we push those changes upstream in a separate E3SM PR before this PR goes back in, or wait and address in the big merge PR?

ambrad: I'm fine having these go in with the big PR since the mods don't affect homme_integration.

brhillman:
After doing some work a month or so ago to bring the two repos into better consistency, I had gotten this merge to a point where the only tests that diff were the SCM tests. However, running e3sm_integration now with the current state of this merge produces the following failing tests:

- DIFF ERP_Ld3.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel.allactive-pioroot1 (phase BASELINE) 
- DIFF ERS_Ln9.ne4pg2_ne4pg2.FRCE-MMF1.chrysalis_intel.eam-cosp_nhtfrq9 (phase BASELINE) 
- DIFF PEM_Ln9.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel (phase BASELINE) 
- DIFF PET_Ln9_PS.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel.allactive-mach-pet (phase BASELINE) 
- DIFF SMS_D_Ld1.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel.allactive-wcprod (phase BASELINE) 
- DIFF SMS_D_Ld1.ne30pg2_EC30to60E2r2.WCYCLSSP370.chrysalis_intel.allactive-wcprodssp (phase BASELINE) 
- DIFF SMS_D_Ld1.ne30pg2_r05_EC30to60E2r2.WCYCL1850.chrysalis_intel (phase BASELINE) 
- DIFF SMS_D_Ln5.ne4_oQU240.F2010.chrysalis_intel.eam-cosplite_nhtfrq5 (phase BASELINE) 
  - FIELDLIST DIFF due to changing U10 to WINDSPD_10M 
- DIFF SMS_Ld2.ne30pg2_r05_EC30to60E2r2.BGCEXP_CNTL_CNPECACNT_1850.chrysalis_intel.elm-bgcexp (phase BASELINE) 
- DIFF SMS_Ld2.ne30pg2_r05_EC30to60E2r2.BGCEXP_CNTL_CNPRDCTC_1850.chrysalis_intel.elm-bgcexp (phase BASELINE) 
- DIFF SMS_Ln5.ne30pg2_r05_EC30to60E2r2.BGCEXP_LNDATM_CNPRDCTC_1850.chrysalis_intel (phase BASELINE) 
- DIFF SMS_Ln5.ne30pg2_r05_EC30to60E2r2.BGCEXP_LNDATM_CNPRDCTC_20TR.chrysalis_intel (phase BASELINE) 
- DIFF SMS_Ln5.ne4_ne4.FSCM-ARM97-MMF1.chrysalis_intel (phase BASELINE) 
  - Expected DIFF due to bugfixes in SCM 
- DIFF SMS_Ln9.ne4_oQU240.F2010.chrysalis_intel.eam-outfrq9s (phase BASELINE)  
  - This is a FIELDLIST diff due to changing U10 to WINDSPD_10M 
- DIFF SMS_R_Ld5.ne4_ne4.FSCM-ARM97.chrysalis_intel.eam-scm (phase BASELINE) 
  - Expected DIFF due to bugfixes in SCM 

Still going through this, but it appears we have some work to figure out why the coupled cases are producing DIFFs.

brhillman:
At least partly responsible for the diffs is the removal of the following from `namelist_defaults_eam.xml`:
```
<!-- special defaults for V2 NE30 -->
<nu hgrid="ne30np4">  1.0e15 </nu>
<hypervis_scaling hgrid="ne30np4" > 0 </hypervis_scaling>
```
I *think* putting these back in will cause any ne30 SCREAMv0 tests to diff, but maybe that's okay. Perhaps it would have been better to put this override for ne30np4 v2 into the namelist use cases for the v2 experiments/tests that needed to remain bfb?

mt5555:
we should remove this code (special defaults for V2 NE30) for E3SM V3 - but that would need to be a separate PR.

Is there another variable that could trigger this exception only for EAM and not SCREAM, i.e.:


<nu hgrid="ne30np4" atm_model != "scream" >  1.0e15 </nu>
<hypervis_scaling hgrid="ne30np4" > 0 </hypervis_scaling>

components/eam/src/physics/rrtmg/cloud_rad_props.F90
```diff
@@ -104,8 +107,10 @@ subroutine cloud_rad_props_init()
    i_lambda = pbuf_get_index('LAMBDAC',errcode=err)
    i_iciwp  = pbuf_get_index('ICIWP',errcode=err)
    i_iclwp  = pbuf_get_index('ICLWP',errcode=err)
-   i_des    = pbuf_get_index('DES',errcode=err)
-   i_icswp  = pbuf_get_index('ICSWP',errcode=err)
+!   if (microp_scheme.ne.'P3') then !AaronDonahue TODO: P3 doesn't have snow
```
brhillman, line 110: This isn't used in SCREAM, we should take the upstream version of this file in the merge.

rljacob:
@brhillman I suggest putting those ne30 defaults back in and letting SCREAM v0 diff.    We can then do a PR to remove them.   v2 answers are preserved on the maint-2.0 branch.

brhillman:
@mt5555 @rljacob I have the scream->E3SM merge ready to go locally, I'm going to reset this PR branch with the new merge based on current E3SM master, with all of the fixes we made downstream in scream to resolve the issues raised in this PR. My only question is to address the ne30 v2 defaults identified in the last couple comments above, should I issue a PR to remove those overrides for consistency with v2 BEFORE this PR goes in, or after? I had to add a commit to this branch to put them back in after merging the scream repo, but if that went in before this PR it wouldn't be necessary. If you have no preference, I'd prefer to just get this big PR merged asap and then will issue the PR to remove the ne30 default override to go in immediately following this PR.

rljacob:
We'd like to PR to be BFB so if it currently is, go ahead and merge it.

brhillman:
Note, the change I am talking about is in #5453

grnydawn (changes requested):
In config_batch.xml
- The XML tag at line # 659 (<batch_system MACH="crusher" type="slurm">) is a duplication of the same tag  at line 67.
- Do we need to keep crusher-scream-cpu and crusher-scream-gpu tags in the file?

"crayclang.cmake" file is missing in cmake_macros directory. This causes build failures for many CPU test cases.

Do we need to keep all crayclang-scream-crusher-scream*.cmake files in cmake_macros directory?

ambrad:
Re: machine files and tags, we need to keep anything with -scream in the name. Anything else can and should be deleted (if not used) or restored (if inadvertantly removed).

cime_config/machines/config_batch.xml
```diff
@@ -616,12 +616,58 @@
...
 
+   <batch_system MACH="spock" type="slurm">
+     <queues>
+       <queue strict="true" nodemin="1" nodemax="4" walltimemax="03:00:00" default="true">ecp</queue>
+       <queue strict="true" nodemin="1" nodemax="4" walltimemax="03:00:00">caar</queue>
+       <queue strict="true" nodemin="5" nodemax="16" walltimemax="01:00:00">caar</queue>
+     </queues>
+   </batch_system>
+
+   <batch_system MACH="crusher" type="slurm">
```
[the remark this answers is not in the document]
brhillman: @oksanaguba it looks like this duplication is in the SCREAM repo as well. We should probably remove this downstream in SCREAM, but isn't SCREAM now using the crusher-scream-* tags? Either way, I think we should remove the duplicate tag downstream in SCREAM.

cime_config/machines/config_batch.xml
```diff
@@ -616,12 +616,58 @@
...
+     <batch_submit>/gpfs/alpine/cli133/world-shared/e3sm/tools/sbatch/throttle</batch_submit>
+     <queues>
+       <queue strict="true" nodemin="1" nodemax="8" walltimemax="01:00:00" default="true">batch</queue>
+       <queue strict="true" nodemin="9" nodemax="64" walltimemax="01:00:00">batch</queue>
+       <queue strict="true" nodemin="65" nodemax="160" walltimemax="01:00:00">batch</queue>
+     </queues>
+   </batch_system>
+
+   <batch_system MACH="crusher-cpu" type="slurm">
```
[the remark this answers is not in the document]
brhillman: I think this comment refers to the outdated version of this branch. This tag is no longer in the PR branch.

cime_config/machines/config_batch.xml
```diff
@@ -616,12 +616,58 @@
...
+     <queues>
+       <queue strict="true" nodemin="1" nodemax="8" walltimemax="08:00:00" default="true">batch</queue>
+       <queue strict="true" nodemin="9" nodemax="64" walltimemax="04:00:00">batch</queue>
+       <queue strict="true" nodemin="65" nodemax="160" walltimemax="02:00:00">batch</queue>
+     </queues>
+   </batch_system>
+
+   <batch_system MACH="crusher-gpu" type="slurm">
```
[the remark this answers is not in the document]
brhillman: This also appears to be outdated and is no longer in the PR branch.

cime_config/machines/config_batch.xml
```diff
@@ -648,12 +648,58 @@
      </queues>
    </batch_system>
 
+   <batch_system MACH="spock" type="slurm">
```
brhillman, line 651: I think this xml tag also needs to be removed from SCREAM, as it is a duplicate of the spock tag that appears earlier in this file.

ambrad: @oksanaguba are you able to take this cleanup task?

cime_config/machines/config_batch.xml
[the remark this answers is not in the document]
brhillman: https://github.com/E3SM-Project/scream/pull/2179 will remove the duplicate tags. I will reset this branch once that PR is merged into scream master.

cime_config/machines/config_batch.xml
brhillman: https://github.com/E3SM-Project/scream/pull/2179 will remove the duplicate tags. I will reset this branch once that PR is merged into scream master.

brhillman:
@grnydawn looks like scream renamed crayclang.cmake -> crayclang-scream.cmake. I will add back in crayclang.cmake after resetting this branch once the scream crusher/spock cleanup PR goes in.

grnydawn:
> @grnydawn looks like scream renamed crayclang.cmake -> crayclang-scream.cmake. I will add back in crayclang.cmake after resetting this branch once the scream crusher/spock cleanup PR goes in.

@brhillman I see. Once the cmake file is recovered, I will try to run e3sm_integration test on this branch to see if there is any "unfamiliar" issue on Crusher. In the meantime, if you want me to run some GPU test cases on Crusher, please let me know how to do it.

brhillman:
@grnydawn those changes have been added, branch reset, this should address your issues above.  I also added back in the v2 ne30 defaults, so this branch should be BFB with E3SM master now, aside from the SCM tests we know will diff.

brhillman:
I am also running e3sm_integration on chrysalis with baseline compare against master.

grnydawn:
> @grnydawn those changes have been added, branch reset, this should address your issues above. I also added back in the v2 ne30 defaults, so this branch should be BFB with E3SM master now, aside from the SCM tests we know will diff.

I saw that the cmake file is back now. I will run the test suite once Crusher is back online.

brhillman:
Some updates, I'm running into a really weird testing issue that has me kind of stumped. On chrysalis, if I run my own baseline with a fresh reset to origin/master, and then compare with a local merge of this branch into master, I get results I expect; a few diffs due to different field lists (rename of U10 to WINDSPD_10M), and expected diffs in SCM tests. However, if I run the same integration test against the system baseline, I get some diffs in the mpaso.hist rest file for some coupled tests, but the rest of the comparisons for that SAME test case are BFB. I'm not sure how this is possible, unless these are diffs in just diagnostic quantities I guess, but I don't understand why I only get this behavior against the system baselines. Here is an example, on chrysalis, 
```
    ERP_Ld3.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel.allactive-pioroot1.C.20230217_140140_zc4910.mpaso.hist.am.globalStats.0001-01-01.nc.rest did NOT ma
tch mpaso.hist.am.globalStats.0001-01-01.nc.rest
    cat /lcrc/group/e3sm/ac.brhillman/scratch/chrys/ERP_Ld3.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel.allactive-pioroot1.C.20230217_140140_zc4910/run/ERP_Ld3.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel.allactive-pioroot1.C.20230217_140140_zc4910.mpaso.hist.am.globalStats.0001-01-01.nc.rest.cprnc.out
```
but then the actual coupler files look bfb:
```
2023-02-17 16:58:05: tail -n20 /lcrc/group/e3sm/ac.brhillman/scratch/chrys/ERP_Ld3.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel.allactive-pioroot1.C.20230217_140140_zc4910/run/ERP_Ld3.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel.allactive-pioroot1.C.20230217_140140_zc4910.cpl.hi.0001-01-04-00000.nc.cprnc.out

              236853  (     1,     1,     1) (     1,     1,     1)
              236853   0.000000000000000E+00   0.000000000000000E+00
              236853   0.000000000000000E+00   0.000000000000000E+00
              236853  (     1,     1,     1) (     1,     1,     1)
          avg abs field values:    0.000000000000000E+00
                                   0.000000000000000E+00
************************************************************************************************************************************
  
SUMMARY of cprnc:
 A total number of    474 fields were compared
          of which      0 had non-zero differences
               and      0 had differences in fill patterns
               and      0 had different dimension sizes
 A total number of      0 fields could not be analyzed
 A total number of      0 time-varying fields on file 1 were not found on file 2.
 A total number of      0 time-constant fields on file 1 were not found on file 2.
 A total number of      0 time-varying fields on file 2 were not found on file 1.
 A total number of      0 time-constant fields on file 2 were not found on file 1.
  diff_test: the two files seem to be IDENTICAL 
```
Looking at the mpaso.hist cprnc output, I see things like this:
```
 totalVolumeChange   (Time)  t_index =      1     1
          1        1  (     1) (     1) (     1) (     1)
                   1   8.317304320000000E+09   8.317304320000000E+09 2.6E+02  8.317304320000000E+09 3.1E-08  8.317304320000000E+09
                   1   8.317304576000000E+09   8.317304576000000E+09          8.317304576000000E+09          8.317304576000000E+09
                   1  (     1) (     1)
          avg abs field values:    8.317304320000000E+09    rms diff: 2.6E+02   avg rel diff(npos):  3.1E-08
                                   8.317304576000000E+09                        avg decimal digits(ndif):  7.5 worst:  7.5
 RMS totalVolumeChange                2.5600E+02            NORMALIZED  3.0779E-08
```
so those files apparently do look like diffs, but again I did NOT see this in a test from which I generate my own baseline with current master, and then compare with a merge of this branch into master.

ambrad:
Is this a case where the nightly baselines use a different number of MPI ranks than default, and some of MPAS's diagnostics aren't PE invariant?

brhillman:
@ambrad wouldn't those diagnostics fail ERP in that case?

brhillman:
Ah, this DOES appear to be a case where the nightlies use different number of mpi ranks though. I found this for the baseline generation case:
```
$ ./xmlquery NTASKS
        NTASKS: ['CPL:192', 'ATM:192', 'LND:192', 'ICE:192', 'OCN:64', 'ROF:192', 'GLC:1', 'WAV:1', 'IAC:1', 'ESP:1']
```
while for my comparison test I have:
```
$ ./xmlquery NTASKS
        NTASKS: ['CPL:2752', 'ATM:2752', 'LND:192', 'ICE:2560', 'OCN:640', 'ROF:192', 'GLC:1', 'WAV:1', 'IAC:1', 'ESP:1']
```
The last bless appears to be 2022 October 15, so perhaps the default number of tasks has changed since then? In any event, it seems problematic that this test would diff at all with different mpi ranks. I'll see if this test passes a simple ERP on master.

brhillman:
That test does pass baseline compare if I manually change NTASKS to match what the stored baseline uses.

grnydawn:
@brhillman, @ambrad  , my e3sm_integration test has finished on Crusher and all the test cases except the following two scream tests show BFB. Do you have any open issue related to this PR? Otherwise, I think I can move forward to merge to next branch.

BFB test summary:

The test compares HEAD on brhillman/add-eamxx(1839c2d) to the merge-base on master(0a642b7)

* More than 50 test cases are failed at NLCOMP phase. The namelist change seems to be only within atm_nl. Otherwise, let me know.
Differences in namelist 'ctl_nl':
 found extra variable: 'se_geometry'
 found extra variable: 'se_lx'
 found extra variable: 'se_ly'
 found extra variable: 'se_ne_x'
 found extra variable: 'se_ne_y'
 found extra variable: 'do_shoc_sgs'
 found extra variable: 'do_spa_optics'
Found extra namelist: micro_nl

* 4 test cases are non-bfb with the message of "the two files DIFFER only in their field lists". It seems that one netcdf file has a variable that the other netcdf doesn't.

* 2 test cases are non-bfb with actual difference in corresponding variables. I assume that the following two cases are SCREAM test cases?
  - SMS_Ln5.ne4_ne4.FSCM-ARM97-MMF1.crusher_crayclang:  60 had non-zero differences
  - SMS_R_Ld5.ne4_ne4.FSCM-ARM97.crusher_crayclang.eam-scm: 122 fields could not be analyzed

brhillman:
@grnydawn I pushed a commit earlier today to fix the field list diffs. SCREAM had renamed U10 to WINDSPD_10M, since U10 is kind of misleading because we typically use names like "U500" to refer to the zonal wind at 500 mb. I was tempted to let this into the merge, but realized it would break E3SM diagnostics, which look for U10, and didn't want inconsistency there. So, the commit I added today fixes this. With that, the only diffs should be those SCM diffs. Those are expected diffs due to changes that @bogensch made downstream in SCREAM to support DP-SCREAM. Those should be the only diffs this PR introduces at this point. The namelist diffs I believe are all expected.

grnydawn:
@brhillman thanks for the explanations. After checking this fix briefly on Crusher, I will try to merge it to next.

@singhbalwinder , @wlin7 , @PeterCaldwell, are you still reviewing this PR? Please post your review here if you have any concerns on this PR.

PeterCaldwell:
I'm not going to review it - I think it is a good idea and I trust the rest of the team's judgement on the technical bits. Balwinder is on vacation until March, so we shouldn't wait for him. @wlin7 may have useful comments.

grnydawn:
> I'm not going to review it - I think it is a good idea and I trust the rest of the team's judgement on the technical bits. Balwinder is on vacation until March, so we shouldn't wait for him. @wlin7 may have useful comments.

@PeterCaldwell , thanks for the update.

rljacob:
@grnydawn you're not the integrator for this PR.  I've unassigned @singhbalwinder and made @wlin7 the integrator.

rljacob:
@bogensch I would still like more explanation on the SCM changes.

ambrad (approved):

ambrad:
The following are for the doubly periodic model. They are parameters in Homme.
```
Differences in namelist 'ctl_nl':
found extra variable: 'se_geometry'
found extra variable: 'se_lx'
found extra variable: 'se_ly'
found extra variable: 'se_ne_x'
found extra variable: 'se_ne_y'
```

grnydawn:
@rljacob, I thought that I was assigned as an integrator. Thanks for correct me.

bogensch:
@rljacob The SCM diffs that Ben is seeing in the tests are expected and due to two fixes that were made on the SCREAM repo:  1) There was an order of operations change on how moisture was updated after applying IOP forcing in the SCREAM code base.  This was done to ensure b4b restarts for DP-SCREAM simulations (or any simulation that uses the IOP forcing routine, as SCM). This results only in roundoff level differences.  2) There was a bug fix only related to the very last time step in a SCM simulation where the surface fluxes were not being updated properly. This results only in answer changes in regards to the very last time step.

rljacob:
I've edited the top level description but would like a sentence on what is happening in each major directory as indicated by "ADD DETAIL".  Might just be "fix SCM bugs" but it should say.

brhillman:
@rljacob I filled in the sections I could figure out easily. @bogensch please look at the SCM sections in the PR description text. @ambrad can you comment on the HOMME changes?

brhillman:
@rljacob I don't see any more TODO items associated with this, I think we've address everything you pointed out in the PR description text.

rljacob (approved):

cime_config/allactive/config_compsets.xml
```diff
@@ -14,7 +14,7 @@
     TIME_ATM[%phys]_LND[%phys]_ICE[%phys]_OCN[%phys]_ROF[%phys]_GLC[%phys]_WAV[%phys][_ESP%phys][_BGC%phys]
     Where for the EAM specific compsets below the following is supported
     TIME = Time period (e.g. 2000, HIST, RCP8...)
-    ATM  = [EAM, SATM]
+    ATM  = [EAM, SATM, SCREAM]
```
wlin7, line 17: Does not matter for the PR. Would EAMXX be more appropriate here?

cime_config/config_archive.xml
```diff
@@ -25,6 +25,15 @@
     </test_file_names>
   </comp_archive_spec>
 
+  <comp_archive_spec compname="scream" compclass="atm">
```
wlin7, line 28: eamxx instead of scream?

cime_config/config_files.xml
```diff
@@ -127,7 +127,7 @@
       <value component="satm"                        >$SRCROOT/components/stub_comps/satm</value>
       <value component="xatm"                        >$SRCROOT/components/xcpl_comps/xatm</value>
       <value component="eam"                         >$SRCROOT/components/eam/</value>
-      <value component="scream"                      >$SRCROOT/components/scream/</value>
+      <value component="scream"                      >$SRCROOT/components/eamxx/</value>
```
wlin7, line 130: Never mind about the above comments if we want to retain the component name 'scream'.

brhillman: Yes these are good points. I think we will want to change these things, but I think that is work for a subsequent PR.

cime_config/config_archive.xml
brhillman: To be addressed in a future PR.

cime_config/allactive/config_compsets.xml
brhillman: Yes it would. To be addressed in a future PR.

wlin7:
The changes in the script files and Fortran modules look good to me. (I didn't review the eamxx codes). With the tests that @brhillman and @grnydawn have performed, the PR by itself should be fine and let the nightly tests do its job after merging.

A few notes: the use_case files for SCREAMv0 compsets are outdated, and merging in p3 from v3atm would further change the default settings. A bigger concern is on the Fortran version of P3 codes. The P3 for SCREAMv0 has some differences from the version in v3atm, whose P3 PR is also ready. Would it be fine, to resolve the difference using what in v3atm? @crterai particurlarly noted that dpscream config is more likely to be affected.

wlin7 (approved):

wlin7:
Merged to next. Flag to control the selection of the P3 version (the one in this PR or from v3atm) during case configure time will be implemented in a separate PR.

wlin7:
Merged to master. SCM and NML diffs are expected. A PET test FAIL on pm-cpu was a build time issue on an elm module. The module built itself appears ok. The problem is with some other module failed to read it. Continue to monitor if it is just a fluke due to machine environment.

Request submitted to bless all the DIFFs.

rljacob:
On GCP, there are some errors from downloading new data:
```
Model eam missing file micro_p3_lookup_dir = '/home/inputdata/atm/scream/tables'
Trying to download file: 'atm/scream/tables' to path '/home/inputdata/atm/scream/tables' using WGET protocol.
```
Its a particular GCP problem that is crashes there but....why is it trying to download p3 info when p3 isn't being used in this case? (its a boring ne30_oECv3.F2010 case).

wlin7:
Good catch, and an interesting problem. Indeed all the tests reporting NML diff today indicate there is a new section of micro_nl added to atm_in. "micro_p3_lookup_dir = '/home/inputdata/atm/scream/tables" is in it. GCP recognized it as a data path and tried to download. But it is a directory. Actual Is wget in cime wired to download a whole directory, and do we want that?

rljacob:
Does that directory already exist on all our other test platforms?

wlin7:
Yes, on compy and chrysalis. The contents are not the same. chrysalis inputdata space has fewer files than in the inputdata server. BTW, v3atm also uses  micro_nl section, but points to a different path for "micro_p3_lookup_dir".

wlin7:
In general that suggest scream cases (with p3) were created before on those machine, and the path exists when the cases set to download individual files. Likely a scream case was never created on GCP. A quick solution is to create that dir on GCP.

rljacob:
Fix for p3 namelist is in #5484

PR #5420, opened 2023-01-26
Update module versions for NERSC machines after maintenance

ndkeen:
Specifically machines pm-cpu, pm-gpu, alvarez
This change is required as the previous versions of the modules were removed.

Running e3sm_integration worked except for 2 tests that I think were already failing in this branch.
```
ERS.ne4_oQU240.F2010.pm-cpu_gnu.eam-hommexx
SMS_Ld30.f45_f45.IELMFATES.pm-cpu_gnu.elm-fates_satphen
```

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -133,10 +133,10 @@
         <command name="load">craype-accel-host</command>
         <command name="load">cray-libsci</command>
         <command name="load">craype</command>
-        <command name="load">cray-mpich/8.1.17</command>
-        <command name="load">cray-hdf5-parallel/1.12.1.5</command>
-        <command name="load">cray-netcdf-hdf5parallel/4.8.1.5</command>
-        <command name="load">cray-parallel-netcdf/1.12.2.5</command>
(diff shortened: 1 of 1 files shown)


cime_config/machines/config_machines.xml
```diff
@@ -387,6 +387,7 @@
       <env name="HDF5_USE_FILE_LOCKING">FALSE</env>
       <env name="PERL5LIB">/global/cfs/cdirs/e3sm/perl/lib/perl5-only-switch</env>
       <env name="FI_CXI_RX_MATCH_MODE">software</env>
+      <env name="MPICH_COLL_SYNC">MPI_Bcast</env>
```
rljacob, line 390: Is this a network performance tweak?

rljacob (approved):

cime_config/machines/config_machines.xml
ndkeen: In this case, it is a setting for alvarez and we just want to be consistent with settings already in place for pm-cpu. After SW/HW changes, it could be good to check if even still need this work-around. But at the time, there were no negative side effects to using it.

PR #5421, opened 2023-01-27
Fixes initialization of snow ice mass tracer

### LINKED ISSUE #5403: Sea ice snow-ice mass  tracer is incorrectly initialized to zero
The snow ice mass tracer in MPAS-SI should not be less than the snow density.  When the snow ice and snow liquid masses are conserved during thermodynamic changes, sublimation and condensation changes of snow ice mass will result in erroneous losses/increases to the snow liquid content which then feeds back on the change in snow grain radius. This has consequences for the albedo.

Testing a solution where the snow ice mass is initialized as the mean snow density.

njeffery:
non BFB changes:
1. Initialize snow ice mass tracer as the mean snow density
2. Changed to higher precision constant in computing snow aging table index

BFB changes:
3. Removed rsnw_fresh. Freshly fallen snow grain radius is config_fallen_snow_radius (rsnw_fall) BFB in standard configuration,  BFB with snow tracers off but needs config_fallen_snow_radius = 100.0)
4. Correction to diagnostic variable, melts
5. Change effective snow density to new snow density value when there's no snow (diagnostic change)
6. Remove unnecessary absolute value for snow level thicknesses

Fixes #5403 
[non-BFB]

### DIFF
--- a/components/mpas-seaice/src/column/ice_shortwave.F90
+++ b/components/mpas-seaice/src/column/ice_shortwave.F90
@@ -3665,7 +3665,6 @@
 
       real (kind=dbl_kind), parameter :: &
          ! units for the following are 1.e-6 m (micro-meters)
-         rsnw_fresh    =  100._dbl_kind, & ! freshly-fallen snow grain radius 
          rsnw_nonmelt  =  500._dbl_kind, & ! nonmelt snow grain radius
          rsnw_sig      =  250._dbl_kind    ! assumed sigma for snow grain radius
 
@@ -3696,7 +3695,7 @@
         ! the sign is negative so that if R_snw is 1, then the
         ! snow grain radius is reduced and thus albedo increased.
         rsnw_nm = rsnw_nonmelt - R_snw*rsnw_sig
-        rsnw_nm = max(rsnw_nm, rsnw_fresh)
+        rsnw_nm = max(rsnw_nm, rsnw_fall)
         rsnw_nm = min(rsnw_nm, rsnw_mlt) 
       
         do ks = 1, nslyr
@@ -3704,7 +3703,7 @@
            rhosnw(ks) = rhos
            ! snow grain radius between rsnw_nonmelt and rsnw_mlt
            rsnw(ks) = rsnw_nm + (rsnw_mlt-rsnw_nm)*fT
-           rsnw(ks) = max(rsnw(ks), rsnw_fresh)
+           rsnw(ks) = max(rsnw(ks), rsnw_fall)
            rsnw(ks) = min(rsnw(ks), rsnw_mlt)
         enddo        ! ks
 
--- a/components/mpas-seaice/src/column/ice_snow.F90
+++ b/components/mpas-seaice/src/column/ice_snow.F90
@@ -68,21 +68,21 @@
       rhos_eff = c0
       rhos_cmp = c0
 
-      if (vsno > puny) then
-
       !-----------------------------------------------------------------
       ! Initialize effective snow density (compaction) for new snow
       !-----------------------------------------------------------------
 
-         do n = 1, ncat
-               do k = 1, nslyr
-                  if (rhos_cmpn(k,n) < rhosmin) rhos_cmpn(k,n) = rhosnew
-               enddo
-         enddo
+      do n = 1, ncat
+            do k = 1, nslyr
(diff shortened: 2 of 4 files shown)


njeffery:
These changes were tested in single column stand alone mode with and without snow tracers active. 

I also ran a 25-year fully coupled run (icepack.step_snow.eclare) hybrid of the WC pre-industrial  spin-up to compare with master (icepack.eclare.master).  Changes are small. MPAS-analysis diagnostics are here:

https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jeffery/E3SMv2/icepack.step_snow.eclare/

eclare108213 (approved):
Looks good.  The change to melts only impacts runs with l_brine=F, a configuration that E3SM does not use.  Non-BFB changes affect the two hemispheres in different ways, with regional differences also apparent.  Generally speaking, the overall simulation in the Arctic looks better in the metrics available, with a somewhat different (arguably better) thickness pattern.  The Antarctic is mixed and I suspect the changes are not significant in a climate sense.

proteanplanet (commented):
Have you run this with config_am_conservationcheck_enable = true in E3SM?

njeffery:
@proteanplanet  No.  I'll turn this on and continue the runs.  Is 5 years enough?  I don't expect these changes to alter conservation, but good to check.

proteanplanet:
> @proteanplanet No. I'll turn this on and continue the runs. Is 5 years enough? I don't expect these changes to alter conservation, but good to check.

1 year is enough.  I'm close to having this as a single command test for sea ice, but for now, I'm afraid the more manual approach is needed.   The aim would be to perform this at the individual PR level from now on, so we won't get caught later on.

njeffery:
I know how to do this for mpas. Do you know the options for eam?

jonbob:
@njeffery -- you can set do_budgets=.true. in user_nl_eam (I think)

njeffery:
@jonbob Thanks!

njeffery:
[ConservationErrors_icepack.step_snow.31-36.pdf](https://github.com/E3SM-Project/E3SM/files/10724858/ConservationErrors_icepack.step_snow.31-36.pdf)

njeffery:
@proteanplanet : I've attached a pdf with the ocean and sea ice conservation errors for the feature branch.  I don't have a master comparison run, but hopefully you're familiar with these values. The coupler log files also have budget information for years 31-36, if you'd like to see that.

njeffery:
@proteanplanet :  Feature run is here
/lcrc/group/e3sm/ac.jeffery/E3SMv2/icepack.step_snow.eclare/run

I'll extend master with conservation  on for comparison.

njeffery:
[ConservationErrors_icepack.eclare.master.26-31.pdf](https://github.com/E3SM-Project/E3SM/files/10796562/ConservationErrors_icepack.eclare.master.26-31.pdf)

njeffery:
@proteanplanet :  The attached figures are from master.  Unfortunately, sea ice energy conservation is still poor.  Though not a result of this PR.

proteanplanet:
Thanks @njeffery.  I propose that this PR is accepted but we create an issue to check and identify conservation issues in master.

proteanplanet (approved):
Accepted.  We will address conservation issues that are not a product of this PR as a separate issue.

njeffery:
Agreed.  Thanks @proteanplanet

eclare108213:
@jonbob, I think this PR is ready to be merged. Is there anything holding it up at this point, that we can help with?  thx

jonbob:
@eclare108213 - it's a non-BFB PR so I'll need to run some longer tests and get people to look at results. I'll make sure this one gets done soon, but we have a backlog of non-BFB PRs to get in

eclare108213:
I'd like to get this into our other branches for the Icepack merge, so we can keep moving forward. Thank you @jonbob !

jonbob:
testing shows expected DIFFs -- merged to next

njeffery:
@jonbob Thank you!

jonbob:
merged to master and expected DIFFs blessed --except on perlmutter, which was down yesterday and not expected to report. @wadeburgess -- I can bless based on the report from the previous day or wait until it reports again?

jonbob:
expected DIFFs on perlmutter blessed

Issue #5423, opened 2023-01-30
MPAS-Seaice does not have ice or snow thickness available as a diagnostic variable


akturner:
MPAS-Seaice does not have ice or snow thickness available as a diagnostic variable

PR #5424, opened 2023-01-30
Added ice thicknesses analysis member

akturner:
Adds a sea ice analysis member that computes ice and snow thickness

Fixes https://github.com/E3SM-Project/E3SM/issues/5423

[NML]
[BFB]


darincomeau:
Thanks for adding this @akturner !

I first tried an `SMS.TL319_EC30to60E2r2.DMPAS-JRA1p5.anvil_gnu` test, and got a build error:

```
/lcrc/group/e3sm/ac.dcomeau/scratch/anvil/SMS.TL319_EC30to60E2r2.DMPAS-JRA1p5.anvil_gnu.20230130_114904_2d1qrw/bld/cmake-bld/core_seaice/analysis_members/mpas_seaice_analysis_driver.f90:35:7:

    use seaice_thicknesses
       1
Fatal Error: Can't open module file ‘seaice_thicknesses.mod’ for reading at (1): No such file or directory
```

I made the following change and then it built successfully:
```
diff --git a/components/mpas-seaice/src/seaice.cmake b/components/mpas-seaice/src/seaice.cmake
index 8e0c8d1..7603b3f 100644
--- a/components/mpas-seaice/src/seaice.cmake
+++ b/components/mpas-seaice/src/seaice.cmake
@@ -97,6 +97,7 @@ list(APPEND RAW_SOURCES
   core_seaice/analysis_members/mpas_seaice_pointwise_stats.F
   core_seaice/analysis_members/mpas_seaice_unit_conversion.F
   core_seaice/analysis_members/mpas_seaice_ice_shelves.F
+  core_seaice/analysis_members/mpas_seaice_thicknesses.F
 )
 ```
 
 So I'd recommend making this change.

**Comment by darincomeau** [comment]
Next, I added the following to `user_nl_mpassi` to turn it on:
 ```
config_AM_thicknesses_enable = true
config_AM_thicknesses_output_stream = 'output'
```

I changed the frequency of the `output` stream to daily so it would write out, but none of the new thicknesses AM wrote out. I'm not sure if that's the intended functionality. I then added `iceThicknessCell` variable to the list of variables in that `output` stream, and then that variable did write out. I'd personally like to have that variable written to the `output` stream by default (even though that stream isn't written by default), which would make the QC testing a bit easier.

akturner:
@darincomeau Thanks for testing this. I've made the build correction you suggested,

darincomeau (approved):
This now builds for me, and can output `iceThicknessCell` with `config_AM_thicknesses_enable = true`.

Approving based on testing.

Thanks @akturner again for this!

proteanplanet:
@akturner @darincomeau 
Why are we doing this?  All of this information is already output in the thickness distribution, but just needs a little post-processing.  It adds another field when we are trying to keep them to a minimum in E3SM?

darincomeau:
@proteanplanet this new analysis member is off by default here, no fields added. My use case is for the QC testing, which yes it could be done with post-processing there.

darincomeau:
>All of this information is already output in the thickness distribution

@proteanplanet I wasn't able to find this - can you point to what you're referring to?

akturner:
@darincomeau I think Andrew is referring to iceAreaCategory and iceVolumeCategory

proteanplanet:
@darincomeau Sorry for my slow response:  I wanted to go back and check that the output that we switched on by default for all V2 code hadn't changed in my absence last year.  Everything you could want in terms of mean ice and snow thickness, area, extent, or volume (which is really just mean cell thickness), right down to each individual thickness category  can be obtained from this monthly output:

        float timeMonthly_avg_iceAreaCategory(Time, nCells, nCategories, ONE) ;
        float timeMonthly_avg_iceVolumeCategory(Time, nCells, nCategories, ONE) ;
        float timeMonthly_avg_snowVolumeCategory(Time, nCells, nCategories, ONE) ;
        float timeMonthly_avg_iceAreaCell(Time, nCells) ;
        float timeMonthly_avg_iceVolumeCell(Time, nCells) ;
        float timeMonthly_avg_snowVolumeCell(Time, nCells) ;

Even at daily temporal resolution, you can still derive mean ice thickness as distinct from mean ice and snow volume (mean cell thickness) with these fields:

        float timeDaily_avg_iceAreaCell(Time, nCells) ;
        float timeDaily_avg_iceVolumeCell(Time, nCells) ;
        float timeDaily_avg_snowVolumeCell(Time, nCells) ;

I have no objection to adding a special analysis member if that makes people's lives easier for development, but I caution against making it a default field for the model, since every field increases the amount of storage space for E3SM, and this one is really simple to calculate in post processing.

jonbob:
passes:

with expected NML DIFFs

merged to next

jonbob:

PR #5426, opened 2023-01-31
Homme: Permit createExecLib-based builds to find netcdf.mod.

ambrad:
This commit supports running RRM in the EAMxx model. Homme needs to be able to read the mesh netcdf file. This commit supports the configuration of Homme through EAMxx such that it can find netcdf.mod.

[BFB]

### DIFF
--- a/components/homme/CMakeLists.txt
+++ b/components/homme/CMakeLists.txt
@@ -50,8 +50,16 @@ IF (HOMME_BUILD_EXECS OR BUILD_HOMME_TOOL)
 ELSE ()
   SET(BUILD_HOMME_WITHOUT_PIOLIBRARY ON CACHE BOOL "")
 ENDIF ()
+
+SET(HOMME_BUILD_SCORPIO ON)
 IF(BUILD_HOMME_WITHOUT_PIOLIBRARY)
   MESSAGE(STATUS "This configuration builds without PIO library and NetCDF calls")
+  SET(HOMME_BUILD_SCORPIO OFF)
+ELSE()
+  IF (TARGET pioc OR TARGET piof)
+    message(STATUS "pioc and/or piof are already defined; skipping SCORPIO build in HOMME")
+    SET(HOMME_BUILD_SCORPIO OFF)
+  ENDIF()
 ENDIF()
 
 #option to discard forcings call and push to/from F
@@ -355,9 +363,9 @@ ADD_SUBDIRECTORY(utils/cime/CIME/non_py/src/timing)
 # CMAKE_CURRENT_SOURCE_DIR == homme
 SET (CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake" ${CMAKE_MODULE_PATH})
 SET (CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/utils/cime/src/CMake" ${CMAKE_MODULE_PATH})
-SET (CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/utils/externals/scorpio/cmake" ${CMAKE_MODULE_PATH})
-
-
+if (HOMME_BUILD_SCORPIO)
+  SET (CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/utils/externals/scorpio/cmake" ${CMAKE_MODULE_PATH})
+endif ()
 
 #
 # Scorpio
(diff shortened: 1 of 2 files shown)


ambrad:
As an initial check (I'll run the full e3sm_developer tests later), I ran the following test sets to make sure the build system still handles all three cases:
```
ERS.ne4_oQU240.F2010.chrysalis_intel.eam-hommexx
eam_theta_pg2
homme_integration
```
Separately, I ran SCREAMv1 with this PR brought in. Everything passes.

components/homme/CMakeLists.txt
```diff
@@ -370,26 +376,38 @@ SET (CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/utils/externals/scorpio/cmak
...
+      ADD_SUBDIRECTORY(utils/externals/scorpio_classic)
+      SET(GENF90_PATH ${CMAKE_CURRENT_SOURCE_DIR}/utils/cime/CIME/non_py/externals/genf90)
+    endif ()
+  else ()
+    # The motivation for this block is that EAMxx is building Homme. Homme
+    # shouldn't build SCORPIO, but it does still need to find netcdf.mod to read
+    # RRM grids.
+    #   Use SCORPIO's FindNetCDF.cmake to get the Fortran netcdf.mod include
+    # directory.
+    find_package(NetCDF COMPONENTS Fortran REQUIRED)
```
bartgol, line 406: This is odd. EAMxx does require Scorpio, and indeed is wrapping scorpio libs in the pioc target (hence, HOMME_BUILD_SCORPIO=FALSE). In doing this, it should already link the netcdf libs to the pioc and piof targets. `target_link_libraries (${execLibName} piof)` should then attach the netcdf deps to homme...

components/homme/CMakeLists.txt
```diff
@@ -50,8 +50,16 @@ IF (HOMME_BUILD_EXECS OR BUILD_HOMME_TOOL)
 ELSE ()
   SET(BUILD_HOMME_WITHOUT_PIOLIBRARY ON CACHE BOOL "")
 ENDIF ()
+
+SET(HOMME_BUILD_SCORPIO ON)
 IF(BUILD_HOMME_WITHOUT_PIOLIBRARY)
   MESSAGE(STATUS "This configuration builds without PIO library and NetCDF calls")
+  SET(HOMME_BUILD_SCORPIO OFF)
+ELSE()
+  IF (TARGET pioc)
```
bartgol, line 59: Shouldn't Homme check for `piof` rather than pioc? Unless you use the C interfaces, that's the lib you need to link, no?

components/homme/cmake/HommeMacros.cmake
```diff
@@ -233,7 +233,8 @@ macro(createExecLib libName execType libSrcs inclDirs macroNP
   ADD_DEFINITIONS(-DHAVE_CONFIG_H)
 
   ADD_LIBRARY(${libName} ${libSrcs})
-  TARGET_INCLUDE_DIRECTORIES (${libName} PUBLIC ${inclDirs} ${modulesDir} ${CMAKE_CURRENT_BINARY_DIR})
+  TARGET_INCLUDE_DIRECTORIES (${libName} PUBLIC ${inclDirs} ${modulesDir} ${CMAKE_CURRENT_BINARY_DIR}
```
bartgol, line 236: Rather than add `PIO_INCLUDE_DIRS` to the include dirs, I think
```
target_link_libraries (${libName} piof)
```
should be more appropriate. Since targets transfer their props/deps, this would take care of fwd-ing all of PIO deps.

components/homme/CMakeLists.txt
```diff
@@ -370,26 +376,38 @@ SET (CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/utils/externals/scorpio/cmak
...
+    message("** Configuring SCORPIO")
+    # pio needs cime/externals/genf90/genf90.pl
+    if (HOMME_USE_SCORPIO)
+      # Need to use scorpio's older genf90 for now
+      SET(GENF90_PATH ${CMAKE_CURRENT_SOURCE_DIR}/utils/utils/externals/scorpio/src/genf90)
+      SET(PIO_ENABLE_TOOLS OFF CACHE BOOL "Disabling Scorpio tool build")
+      ADD_SUBDIRECTORY(utils/externals/scorpio)
+    else ()
+      # The default I/O library used in "Scorpio classic"
+      ADD_SUBDIRECTORY(utils/externals/scorpio_classic)
```
bartgol, line 397: I'm curious: is it necessary to still handle scorpio classic?

bartgol (commented):
I think it's mostly ok, but, unless I'm missing something, I think the CMake machinery may use some patches.

components/homme/CMakeLists.txt
ambrad: Here you mean linking. But netcdf.mod is for compiling, i.e., `#include "netcdf.mod"` in mesh_mod.F90.

components/homme/CMakeLists.txt
ambrad: Possibly. But note this isn't really imporant. The main thing, as you pointed out, is we can't have Homme buildings its own SCORPIO if it's already on the link line. `pioc` seems the safer thing to check. `piof` would imply that somehow we're robust to building `piof` in Homme and `pioc` elsewhere, which we're not.

components/homme/cmake/HommeMacros.cmake
ambrad: Here again the issue is compiling, not linking. Note that EAMxx doesn't require Homme to do any linking.

components/homme/CMakeLists.txt
ambrad: This has nothing to do with the PR other than indentation. I move to overrule this question. "Seconded!" shouts a helpful imaginary second dev.

components/homme/CMakeLists.txt
bartgol: `target_link_libraries` is a misleading name. If the "linked" item is a library, it simply links. But if it is a CMake target (as pioc and piof are), it does eveyrhing: it links, but also adds PUBLIC include dirs of the target, and propagate PUBLIC flags/CPP of the target. That's basically the only reason I create the pioc/piof imported CMake targets in EAMxx, so that one only needs to do `target_link_libraries(the_tgt pioc)` to get everything they need from pioc.

bartgol: TBH, I should first check that the folder where netcdf.mod is located is included in the INTERFACE_INCLUDE_DIRECTORIES section of `piof` target...

ambrad: Ok. I think I tried that and concluded that wasn't working, but I'll try again. Is it possible that `target_include_directories` with `piof` as an item would work? I'll try that, too.

bartgol: Uhm, that should not work, I think. Only target_link_libraries does the extra magic. The other `target_xyz` do the thing that the name say.

I will check to see if the imported piof target was created correctly in EAMxx. It might be I'm not setting it up correctly.

bartgol: ARG, I copy-pasted stuff for pioc into piof...

```
    add_library(piof INTERFACE)
    target_link_libraries (pioc INTERFACE ${SCORPIO_F_LIB} ${netcdf_f_lib} pioc)
    target_include_directories (pioc INTERFACE ${SCORPIO_INC_DIR} ${netcdf_f_incdir} )
```
Notice the pioc instead of piof...

ambrad: Ok, but just so I'm sure we're on the same page: The goal herre is to get `-I../netcdf-fortran/include` added to the mesh_mod.F90.o target compile line, and nothing more. So, naively, it almost seems like `target_include_directories` ought to do it (except I'm now vaguely recalling it wasn't).

ambrad: I see. I didn't pay enough attention to the AD target lines to catch that bug. Probably explains why I couldn't get the `piof` dependency to do anything.

ambrad: I tested target_link_libraries with the AD fix to piof, and it works. I'll update this PR with that approach.

components/homme/cmake/HommeMacros.cmake
ambrad: Works after the AD fix to the piof target.

ambrad:
For the record: The AD CMake code was not setting up the piof target correctly. Hence when I tried to use it in Homme, it didn't work. My first working version used FindNetCDF to get the netcdf.mod file I need. With the fix to the AD CMake code, I can remove that and switch to using the piof target.

ambrad:
@bartgol works now with the updated `piof` target from the AD. Any further issues?

bartgol (approved):
No further issues, this looks good.

ambrad:
Likely will merge to `next` on 2 Feb.

PR #5432, opened 2023-02-04
 hommexx: make forcingalg int

oksanaguba:
ForcingAlg type, which is used for ftype, as enum currently is not based on int, which makes the code assign ftype values in xx code not to integers we expect (like ftype0 will not equal to 0 if printed), which makes debugging confusing.

This fixes it. 

[BFB]

### DIFF
--- a/components/homme/src/share/cxx/HommexxEnums.hpp
+++ b/components/homme/src/share/cxx/HommexxEnums.hpp
@@ -40,11 +40,11 @@ enum class ComparisonOp {
 
 // =================== Run parameters enums ====================== //
 
-enum class ForcingAlg {
-  FORCING_OFF,
-  FORCING_0, 
-  FORCING_1, // Unsupported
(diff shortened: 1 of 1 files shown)


rljacob:
Is this fixing a bug or just better programming practice?

oksanaguba:
@rljacob this is an improvement. without this, if one tries to use ForcingAlg type in output, they would get a number that does not match 0, or 1, or 2.

oksanaguba:
scream is ok with this change https://github.com/E3SM-Project/scream/pull/2152 .

rljacob:
status:  can go ahead.

oksanaguba:
On chrysalis tests pass
```
cs.status.20230302_100509_o9jung  cs.status.20230302_102113_vzk672  cs.status.20230302_150012_rqlj2k
```

ambrad (approved):

oksanaguba:
in next

PR #5437, opened 2023-02-06
Fix an array allocation bug in the ocean SI solver

### LINKED ISSUE #5361: Bug in MPAS-ocean split implicit solver
When running in debug on anvil @scalandr encountered the following error 
```
forrtl: severe (408): fort: (8): Attempt to fetch from allocatable variable GLOBALREPRODSUM2FLD1 when it is not allocated

```
Looking into the code it is this function call that is the problem  https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_si.F#L1174-L1179

The allocate of the array mentioned in the error is here  https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_si.F#L645-L657

The array allocate is wrapped in an `if(config_btr_si_partition_mach_mode)` so if that variable is false the array is never allocated and the error occurs.

An easy solution is to have the allocation and later deallocation occur regardless of `config_btr_si_partition_mach_mode`, but wonder if this is the most appropriate solution.  What is your desired behavior @hyungyukang?

hyungyukang:
This PR fixes an array allocation bug  in the ocean SI solver.

This PR passes `PEM_Ln9.T62_oQU240.GMPAS-IAF.cori-knl_intel` and `SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.cori-knl_gnu` tests with `config_time_integrator='split_implicit'` `config_btr_si_partition_match_mode=.true.` options. 

These changes are only in the semi-implicit time stepping, which is not used in the E3SM tests (stealth). So this can be classified as BFB for E3SM testing.

Fixes #5361 

[BFB]

### DIFF
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_forward_mode.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_forward_mode.F
@@ -853,6 +853,8 @@ module ocn_forward_mode
 
       if ( config_time_integrator == 'split_explicit') then
          call mpas_dmpar_exch_group_destroy_reusable_buffers(domain, 'subcycleFields')
+      else if ( config_time_integrator == 'split_implicit') then
+         call ocn_time_integrator_si_variable_destroy()
       end if
 
       call ocn_analysis_finalize(domain, ierr)
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_si.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_si.F
@@ -4230,7 +4230,7 @@ module ocn_time_integration_si
 !
 !  routine ocn_time_integrator_si_variable_destroy
 !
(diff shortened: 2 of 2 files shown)


mark-petersen (approved):
Reviewed visually, and these code changes are for array deallocation, and only pertain to the semi-implicit timestepping. I compiled with both gnu and intel, both debug and opt. and ran the nightly test suite with intel debug. Everything works. Thanks!

hyungyukang:
@mark-petersen , Thanks a lot for reviewing and approving this PR and your thorough tests!

vanroekel (approved):
Approving by visual inspection and testing from @mark-petersen   This will fix the error I encountered previously.

jonbob (approved):
Approved based on visual inspection and testing in E3SM

jonbob:
passes:

merged to next

jonbob:
merged to master

PR #5438, opened 2023-02-06
Switch mappy to slurm

jgfouca:
[BFB]

### DIFF
--- a/cime_config/machines/config_batch.xml
+++ b/cime_config/machines/config_batch.xml
@@ -304,10 +304,41 @@
     </directives>
   </batch_system>
 
+  <batch_system type="slurm_single_node" >
+    <batch_query per_job_arg="-j">squeue</batch_query>
+    <batch_submit>sbatch</batch_submit>
+    <batch_cancel>scancel</batch_cancel>
+    <batch_directive>#SBATCH</batch_directive>
+    <jobid_pattern>(\d+)$</jobid_pattern>
+    <depend_string>--dependency=afterok:jobid</depend_string>
+    <depend_allow_string>--dependency=afterany:jobid</depend_allow_string>
+    <depend_separator>:</depend_separator>
+    <walltime_format>%H:%M:%S</walltime_format>
+    <batch_mail_flag>--mail-user</batch_mail_flag>
+    <batch_mail_type_flag>--mail-type</batch_mail_type_flag>
+    <batch_mail_type>none</batch_mail_type>
+    <submit_args>
+      <arg flag="--time" name="$JOB_WALLCLOCK_TIME"/>
+      <arg flag="-p" name="$JOB_QUEUE"/>
+      <arg flag="--account" name="$PROJECT"/>
+    </submit_args>
+    <directives>
+      <directive> --job-name={{ job_id }}</directive>
+      <directive> --nodes=1</directive>
+      <directive> --ntasks={{ total_tasks }}</directive>
+      <directive> --output={{ job_id }}.%j </directive>
+    </directives>
+  </batch_system>
 
     <batch_system MACH="blues" type="slurm">
       <queues>
         <queue walltimemax="01:00:00" nodemax="6" default="true" strict="true">biggpu</queue>
+      </queues>
+    </batch_system>
+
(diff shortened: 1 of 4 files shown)


jgfouca:
I still need to do a larger test, but this appears to work exactly as intended. For the first time ever, I saw mappy 95%+ on top when running 2 64-task jobs.

cime_config/machines/config_batch.xml
```diff
@@ -304,13 +304,44 @@
...
+    </directives>
+  </batch_system>
 
     <batch_system MACH="blues" type="slurm">
       <queues>
         <queue walltimemax="01:00:00" nodemax="6" default="true" strict="true">biggpu</queue>
       </queues>
     </batch_system>
 
+    <batch_system MACH="mappy" type="slurm_single_node">
```
bartgol, line 339: I'm not very familiar with the config_batch.xml file. Does this xml entry inherit all the options of the one you added above without MACH?

cime_config/machines/config_machines.xml
```diff
@@ -1269,7 +1268,7 @@
         <command name="load">acme-pfunit/3.2.8/base</command>
       </modules>
       <modules mpilib="!mpi-serial">
-        <command name="load">acme-openmpi/2.1.5</command>
+        <command name="load">acme-openmpi/4.1.4</command>
```
bartgol, line 1271: Is this module even necessary, given that you prepend your OpenMPI installation to the PATH below?

cime_config/machines/config_machines.xml
```diff
@@ -1283,6 +1282,7 @@
       <env name="OMP_STACKSIZE">64M</env>
       <env name="OMP_PROC_BIND">spread</env>
       <env name="OMP_PLACES">threads</env>
+      <env name="PATH">/ascldap/users/jgfouca/packages/openmpi-4.1.4/bin:$ENV{PATH}</env>
```
bartgol, line 1285: Can we put the src/build/install folders somewhere in something like `/sems-data-store/ACME/software/openmpi`?
That way, if you happen to be on a long vacation when a critical bug shows up, we can still update the installation without the need of a PR in e3sm.

I'm including the build dir, since I'm hoping you left a `do-configure.sh` kind of script there, which we can then use if we need to adjust the installation...

cime_config/machines/config_batch.xml
jgfouca: Yep!

cime_config/machines/config_machines.xml
jgfouca: Oh, good catch. That was something I added before the sems/acme modules were available.

bartgol (approved):

cime_config/machines/config_batch.xml
```diff
@@ -304,13 +304,44 @@
...
+    <depend_allow_string>--dependency=afterany:jobid</depend_allow_string>
+    <depend_separator>:</depend_separator>
+    <walltime_format>%H:%M:%S</walltime_format>
+    <batch_mail_flag>--mail-user</batch_mail_flag>
+    <batch_mail_type_flag>--mail-type</batch_mail_type_flag>
+    <batch_mail_type>none</batch_mail_type>
+    <submit_args>
+      <arg flag="--time" name="$JOB_WALLCLOCK_TIME"/>
+      <arg flag="-p" name="$JOB_QUEUE"/>
+      <arg flag="--account" name="$PROJECT"/>
```
bartgol, line 323: I don't think there's a PROJECT entry for mappy. Will this cause cime to add `--account ` (with no arg) to the submit line? I guess the answer is no, since you tested, but maybe it just happened to work for other reasons, so better double check

jgfouca: It looks like it just skips that argument if $PROJECT is empty.

bartgol: Interesting. If that's a reliable feature of the scripts that handle submission, then it's fine.

jgfouca:
This experiment was a success. Mappy was able to run e3sm_developer in 3 hrs instead of 5.

rljacob:
The PR description should have also mentioned the mpi library version change.

PR #5439, opened 2023-02-07
Update weaver.cmake for gcc8/cuda11

tcclevenger:
Previously, weaver.cmake gave modules for running with gcc7/cuda10 (which produces an internal compiler error (ICE) which is not obvious how to fix). This PR updates comments and adds a cmake var for running with gcc8/cuda11 which does not have the ICE.

I think this file is not used for any testing or nightly runs, so this is just for users that need info about running on weaver and want to point to this file. I would only be concerned if there are current users that point to this file, and do not want to set max team size (I think that is unlikely).


ambrad (approved):
I run with these modules, too. I agree that 8 warps max is needed when running with this gcc/cuda combo.

bartgol (approved):
`gcc/8.3.1` and `cuda/11.2.2` are also the modules used by eamxx, so that's good for consistency.

PR #5441, opened 2023-02-07
Turn MPAS-Ocean conservation analysis member on by default

darincomeau:
Turns on the MPAS-Ocean conservation analysis member by default only for v2 Cryosphere meshes (`SOwISC12to60E2r4` and `ECwISC30to60E2r1`).

This is needed for configurations with prognostic ice-shelf melt fluxes and prescribed iceberg forcing since in those cases, solid and liquid runoff from Antarctica is zeroed out to avoid 'double-counting' this runoff. The fields `removedRiverRunoff` and `removedIceRunoff` are then calculated to allow us to diagnose how much the simulation violates conservation of water/contributes to sea-level rise.

I ran a pair of one-year B-case simulations with `SOwISC12to60E2r4` to test the impact of turning this analysis member on by default on performance, and the throughputs were 10.75 SYPD (on) vs. 10.87 SYPD (off), so a minimal impact.

[BFB]

### DIFF
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -1071,10 +1071,16 @@
 
 <!-- AM_conservationCheck -->
 <config_AM_conservationCheck_enable>.false.</config_AM_conservationCheck_enable>
+<config_AM_conservationCheck_enable ocn_grid="SOwISC12to60E2r4">.true.</config_AM_conservationCheck_enable>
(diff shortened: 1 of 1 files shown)


xylar (approved):
I'm just reviewing by inspection because this seems like a very simple change.  I agree that we need this analysis member on for all production runs with ice-shelf cavities so this seems appropriate.

mark-petersen (approved):
Approving by visual inspection and Darren's testing. Thanks!

jonbob:
passes:

merged to next

jonbob:
merged to master

PR #5442, opened 2023-02-08
Improvements to Diagnostics computation

tcclevenger:
Previously, `prim_diag_scalars()` computation was done on host. 

Additionally, while `prim_energy_halftimes()` was already ported to GPU, it contained a `deep_copy()` to host at every single call, and `Diagnostics::sync_diagnostics_to_host()` defined but never used. A new F90->Cxx function created and deep copies to host only happen right before diagnostic info is printed.

For testing I added changes to `components/homme/cmake/CxxVsF90.cmake.in` from https://github.com/E3SM-Project/E3SM/pull/5386 which compare diagnostic output. Diagnostics related to `Q`,`Qvar`,`Qmass` match between F90 and Cxx runs for all `theta-f*ne2*` tests.

[bfb]

### DIFF
--- a/components/homme/src/theta-l_kokkos/cxx/Diagnostics.cpp
+++ b/components/homme/src/theta-l_kokkos/cxx/Diagnostics.cpp
@@ -15,40 +15,46 @@ namespace Homme
 {
 
 void Diagnostics::init (const ElementsState& state, const ElementsGeometry& geometry,
-                        const HybridVCoord& hvcoord, F90Ptr& elem_state_q_ptr,
-                        F90Ptr& elem_accum_qvar_ptr,  F90Ptr& elem_accum_qmass_ptr,
-                        F90Ptr& elem_accum_q1mass_ptr,F90Ptr& elem_accum_iener_ptr,
-                        F90Ptr& elem_accum_kener_ptr, F90Ptr& elem_accum_pener_ptr)
-{
+                        const HybridVCoord& hvcoord,  const Tracers& tracers,
+                        F90Ptr& elem_state_q_ptr, F90Ptr& elem_accum_qvar_ptr,
+                        F90Ptr& elem_accum_qmass_ptr, F90Ptr& elem_accum_q1mass_ptr,
+                        F90Ptr& elem_accum_iener_ptr, F90Ptr& elem_accum_kener_ptr,
+                        F90Ptr& elem_accum_pener_ptr) {
   // Check state/geometry/hvcoord were inited
   assert (state.num_elems()>0);
   assert (geometry.num_elems()>0);
   assert (hvcoord.m_inited);
+  assert (tracers.inited());
 
   // Check initialization was done right
   assert (m_num_elems==state.num_elems());
+  assert (m_num_tracers==tracers.num_tracers());
 
   m_state    = state;
   m_geometry = geometry;
   m_hvcoord  = hvcoord;
+  m_tracers  = tracers;
 
   m_eos.init(m_theta_hydrostatic_mode,m_hvcoord);
   m_elem_ops.init(m_hvcoord);
 
   // F90 ptr to array (n1,n2,...,nK,nelemd) can be stuffed directly in an unmanaged view
   // with scalar type Real*[nK]...[n2][n1] (with runtime dimension nelemd)
-  h_Q      = decltype(h_Q)(elem_state_q_ptr, m_num_elems);
-  h_Qvar   = decltype(h_Qvar)(elem_accum_qvar_ptr, m_num_elems);
-  h_Qmass  = decltype(h_Qmass)(elem_accum_qmass_ptr, m_num_elems);
+  h_Qvar   = decltype(h_Qvar)  (elem_accum_qvar_ptr,   m_num_elems);
+  h_Qmass  = decltype(h_Qmass) (elem_accum_qmass_ptr,  m_num_elems);
   h_Q1mass = decltype(h_Q1mass)(elem_accum_q1mass_ptr, m_num_elems);
 
-  h_IEner  = decltype(h_IEner)(elem_accum_iener_ptr, m_num_elems);
-  h_KEner  = decltype(h_KEner)(elem_accum_kener_ptr, m_num_elems);
-  h_PEner  = decltype(h_PEner)(elem_accum_pener_ptr, m_num_elems);
+  d_Qvar   = decltype(d_Qvar)  ("Qvar",   m_num_elems);
+  d_Qmass  = decltype(d_Qmass) ("Qmass",  m_num_elems);
+  d_Q1mass = decltype(d_Q1mass)("Q1mass", m_num_elems);
 
-  m_IEner  = decltype(m_IEner)("Internal  Energy", m_num_elems);
-  m_KEner  = decltype(m_KEner)("Kinetic   Energy", m_num_elems);
-  m_PEner  = decltype(m_PEner)("Potential Energy", m_num_elems);
+  h_IEner  = decltype(h_IEner) (elem_accum_iener_ptr, m_num_elems);
+  h_KEner  = decltype(h_KEner) (elem_accum_kener_ptr, m_num_elems);
+  h_PEner  = decltype(h_PEner) (elem_accum_pener_ptr, m_num_elems);
+
+  d_IEner  = decltype(d_IEner) ("Internal  Energy", m_num_elems);
+  d_KEner  = decltype(d_KEner) ("Kinetic   Energy", m_num_elems);
+  d_PEner  = decltype(d_PEner) ("Potential Energy", m_num_elems);
 }
 
 int Diagnostics::requested_buffer_size () const {
@@ -90,10 +96,13 @@ void Diagnostics::init_buffers (const FunctorsBuffersManager& fbm) {
   }
 }
 
-void Diagnostics::sync_diags_to_host () {
-  Kokkos::deep_copy(h_IEner,m_IEner);
-  Kokkos::deep_copy(h_PEner,m_PEner);
-  Kokkos::deep_copy(h_KEner,m_KEner);
+void Diagnostics::sync_diagnostics_to_host () {
+  Kokkos::deep_copy(h_IEner,  d_IEner);
+  Kokkos::deep_copy(h_PEner,  d_PEner);
+  Kokkos::deep_copy(h_KEner,  d_KEner);
+  Kokkos::deep_copy(h_Qvar,   d_Qvar);
+  Kokkos::deep_copy(h_Qmass,  d_Qmass);
+  Kokkos::deep_copy(h_Q1mass, d_Q1mass);
 }
 
 void Diagnostics::run_diagnostics (const bool before_advance, const int ivar)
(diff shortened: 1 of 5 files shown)


components/homme/src/theta-l_kokkos/prim_driver_mod.F90
```diff
@@ -475,6 +475,7 @@ subroutine prim_run_subcycle(elem, hybrid, nets, nete, dt, single_column, tl, hv
     ! Print some diagnostic information
 
     if (compute_diagnostics) then
+       call sync_diagnostics_to_host_c()
        call prim_printstate(elem, tl, hybrid,hvcoord,nets,nete)
```
tcclevenger, line 479: This is now the only place a deep_copy() is done, right before diagnostics are output to screen.

components/homme/src/theta-l_kokkos/cxx/Diagnostics.cpp
```diff
@@ -90,10 +96,13 @@ void Diagnostics::init_buffers (const FunctorsBuffersManager& fbm) {
...
-  Kokkos::deep_copy(h_IEner,m_IEner);
-  Kokkos::deep_copy(h_PEner,m_PEner);
-  Kokkos::deep_copy(h_KEner,m_KEner);
+void Diagnostics::sync_diagnostics_to_host () {
+  Kokkos::deep_copy(h_IEner,  d_IEner);
+  Kokkos::deep_copy(h_PEner,  d_PEner);
+  Kokkos::deep_copy(h_KEner,  d_KEner);
+  Kokkos::deep_copy(h_Qvar,   d_Qvar);
+  Kokkos::deep_copy(h_Qmass,  d_Qmass);
+  Kokkos::deep_copy(h_Q1mass, d_Q1mass);
```
tcclevenger, line 105: Remove all deep copies from the actual computation and add here. This function wasn't previously used, but now it's called right before output to screen.

components/homme/src/theta-l_kokkos/cxx/Diagnostics.hpp
```diff
@@ -65,37 +66,79 @@ class Diagnostics
...
+      }, Qmass);
+
+      // Compute Q1mass
+      Q1mass = 0.0;
+      Dispatch<>::parallel_reduce(kv.team, Kokkos::ThreadVectorRange(kv.team, NUM_PHYSICAL_LEV),
+                                  [=](const int ilev, Real& accumulator){
+        accumulator += qdp(ilev);
+      }, Q1mass);
+    });
+  }
```
tcclevenger, line 140: This is the actual kernel replacing the host computation.

tcclevenger (commented):
Again, hiding whitespaces will be helpful. I left some comments around the main changes since there is a lot of formatting to look around.

tcclevenger:
@ambrad I added but removed you as a reviewer, feel free to review if you like, but I switch to Luca since git blame tells me he wrote this class.

ambrad:
I suggest measuring performance before and after these changes on a GPU machine.

ambrad:
I suggest measuring performance before and after these changes on a GPU machine.

ambrad:
Also, make sure you check that the diagnostics output is the same now as before.

mt5555:
If it is easy, In addition to
"Diagnostics related to Q,Qvar,Qmass match between F90 and Cxx runs for all theta-f*ne2* tests.",
can you add a check (Fortran vs cxx) for a few of the energy variables:

**DYNAMICS**        J/m^2                    W/m^2                W/m^2
KE,d/dt        0.19281305456147E+07  0.33235582381590E+01
IE,d/dt        0.18154663781611E+10 -0.86334723385175E+01
PE,d/dt        0.74433724146489E+09 -0.33449153212375E+02
 E,dE/dt       0.25617317501716E+10 -0.38759067312611E+02

Change from dribbled phys tendencies, viscosity, remap:
dKE/dt(W/m^2):  -0.4016106E+00 -0.1137331E+01 -0.7960283E-01
dIE/dt(W/m^2):  -0.1355975E+02 -0.5023139E+00 -0.7363319E-02
dPE/dt(W/m^2):  -0.5021418E+01  0.1153920E+01 -0.2924423E-02

mt5555 (approved):

components/homme/src/theta-l_kokkos/cxx/Diagnostics.hpp
```diff
@@ -65,37 +66,79 @@ class Diagnostics
...
+      const auto qdp = viewAsReal(Homme::subview(m_tracers.qdp, kv.ie, t2_qdp, kv.iq, igp, jgp));
+
+      auto& Qvar =   d_Qvar  (kv.ie,m_ivar,kv.iq,igp,jgp);
+      auto& Qmass =  d_Qmass (kv.ie,m_ivar,kv.iq,igp,jgp);
+      auto& Q1mass = d_Q1mass(kv.ie,       kv.iq,igp,jgp);
+
+      // Compute Qvar
+      Qvar = 0.0;
+      Dispatch<>::parallel_reduce(kv.team, Kokkos::ThreadVectorRange(kv.team, NUM_PHYSICAL_LEV),
+                                  [=](const int ilev, Real& accumulator){
```
bartgol, line 122: Any reason not to use `[&]` instead of `[=]`, here as well as in the reuctions below?

oksanaguba:
Not necessarily related to this PR, but related to homme diagnostics in xx:
1) FU, FT, FQ(1) are not pushed back from xx code, so in case of scream, where these are originally init-ed on cxx side, the diagnostics of them is wrong (all zeros):
```
    fu  =   0.000000000000000E+00 (  1)  0.000000000000000E+00 (  1)  0.000000000000000E+00
    fv  =   0.000000000000000E+00 (  1)  0.000000000000000E+00 (  1)  0.000000000000000E+00
    ft  =   0.000000000000000E+00 (  1)  0.000000000000000E+00 (  1)  0.000000000000000E+00
    fq1 =   0.000000000000000E+00 (  1)  0.000000000000000E+00 (  1)  0.000000000000000E+00
```

2) by some reason, both theta-l and theta-l_kokkos have share/prim_state_mod, which leads to 2 copies of prim_printstate() function. i think 1 copy would be better.

mt5555:
I think the theta-l_kokkos/share is a symbolic link, so there is only one version of prim_state_mod.F90

bartgol:
Re: printing all zeros. We could ifdef that piece of code out if the XX dycore is running.

tcclevenger:
@mt5555 About comparing energy variables, before this PR, only `KEner` diagnostic was actually deep copied to host. `PEner` and `IEner` were computed but never copied. In this PR, I deep copy all of them to host.

So in the diagnostics output that depend on these, we don't match the previous version. If I change to only copy out `KEner` then they match.

But this brings up the question: why were `PEner` and `IEner` not copied back? Was that a bug? They were in the previous `sync_diag_to_host()` function, but that was never used.

oksanaguba:
it is a bug, but even when deep copied, i do not think they are bfb (this is why i disabled comparing them in https://github.com/E3SM-Project/E3SM/pull/5386/files#diff-a89d29a6d2d20be27dc5813f3fc789b405e41b6223ed3c7cf4f500ca17a63ade ) , is it correct?

mt5555:
Regarding the energy diagnostics:  since this is just a diagnostic, not worth the trouble to ensure BFB with fortran.  Best would be a test that allows some tolerance.  But if that is not simple, then it's fine with just verifying by hand in one case.

ambrad:
We want BFB and Oksana has a PR nearly done that does automatic checking of C++ vs F90.

components/homme/src/theta-l_kokkos/cxx/Diagnostics.hpp
tcclevenger: I was just copying what was done in the other operator. But yeah, better to use [&]. I updated and tested.

mt5555:
Results are great. I tried to do some algebra to estimate the cost of one statefreq output to the cost of one ATM timstep:
old:  1 statefreq output costs about 1.3x of a ATM step
new:  1 statefreq output costs about =  .008x of an ATM step

these costs will get much higher at high node counts (statefreq has a lot of all-reduces), but will be much more affordable after this PR.

oksanaguba:
@tcclevenger in case i missed it, did you rerun stdout diagnostics with PEner and IEner after adding PEner and IEner copies? I assume they are not bfb with F, but would be good to have a note here and in the code in case we forget.

tcclevenger:
@oksanaguba This is for Serial CPU run, for the `theta-fhs1-ne2` test:
```
62: grep_args are dIE/dt -----------------------------------
62: Test did not succeed. CXX and F90 diagnostics outputs differ:
62: f90:
62: dIE/dt(W/m^2):  -0.1225236E+03  0.2384186E-08  0.8790599E-04
62: dIE/dt(W/m^2):  -0.1224707E+03  0.8900960E-07  0.3210886E-03
62: 
62: cxx:
62: dIE/dt(W/m^2):  -0.1225236E+03  0.3178914E-08  0.8790652E-04
62: dIE/dt(W/m^2):  -0.1224707E+03  0.8900960E-07  0.3210889E-03
62: grep_args are (E-E0) -----------------------------------
62: Test did not succeed. CXX and F90 diagnostics outputs differ:
62: f90:
62: (E-E0)/E0  -0.100374799484040E-03
62: 
62: cxx:
62: (E-E0)/E0  -0.100374291407720E-03
```
BFB diffs are in `dIE/dt` and `(E-E0)`. Any other test you want me to check?

oksanaguba:
Conrad, thnaks! No, this is enough, i will keep those fields disabled in cprnc PR then.

oksanaguba:
Re-reading your message I see that you do not have PEner in diffs. I will keep all energy fields not compared for now, this can be another issue later.

oksanaguba:
Re-reading this, i am not sure if you ran weaver cxx-vs-f tests and baseline tests against master. Would you please do so? I will test on chrysalis.

tcclevenger:
@oksanaguba Re-ran all ne2, theta-f tests (cxx vs F90, including diagnostics test you created, and against master baselines). All pass. I of course restricted the diag output to only Q-related variables since the others are expected to pass.

oksanaguba:
Thanks! I am confused about what you said about Q values. I assume you disabled just E, KE, PE, IE values like the list below?
```
foreach (s
  "  u     ="
  "  v     ="
  "  w     ="
  "  T     ="
  "  mu    ="
  "  vTh_dp="
  "  dz(m) ="
  "  dp    ="
  "    fu  ="
  "    fv  ="
  "    ft  ="
  "    fq1 ="
  "min dz/w"
  " qv(  1)="
  "    TBOT="
  "      ps="
  "      M ="
  "  dry M ="
#  "dKE/dt"
#  "dIE/dt"
#  "dPE/dt"
  "dQ1/dt"
  "Q  1,Q"
#  "(E-E0)"
  "(Q-Q0)"
  "pmax_"
  )
```

tcclevenger:
@oksanaguba Yes, those are the variables I am testing.

oksanaguba:
Ran chrysalis tests, homme and e3sm_dev, all good:
```
cs.status.20230227_155555_b5dumj  cs.status.20230227_160937_us9efa  cs.status.20230227_161744_qzypv9  
```

rljacob:
status: ready to merge.

oksanaguba:
in next

PR #5447, opened 2023-02-13
New ocean implicit top drag feature

cbegeman:
This PR creates a new option to use implicit top drag in the same manner as implicit bottom drag. It assumes that the only setting in which top drag will be used is in land-ice-covered regions. 

This feature is off by default (stealth). Implicit top drag is activated by setting `config_use_implicit_top_drag = .true.` and `config_implicit_top_drag_coeff` > 0 and `config_land_ice_flux_mode = 'standalone'` (while also ensuring that at least part of the domain has `landIceFraction` > 0).

[NML]
[BFB] - Stealth

### DIFF
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -378,7 +378,11 @@
 <config_use_implicit_bottom_drag>.true.</config_use_implicit_bottom_drag>
 <config_implicit_bottom_drag_coeff>1.0e-3</config_implicit_bottom_drag_coeff>
 <config_use_implicit_top_drag>.false.</config_use_implicit_top_drag>
-<config_implicit_top_drag_coeff>1.0e-3</config_implicit_top_drag_coeff>
+<config_implicit_top_drag_coeff>2.5e-3</config_implicit_top_drag_coeff>
+<config_implicit_top_drag_coeff ocn_grid="oEC60to30v3wLI">4.48e-3</config_implicit_top_drag_coeff>
+<config_implicit_top_drag_coeff ocn_grid="ECwISC30to60E1r2">4.48e-3</config_implicit_top_drag_coeff>
+<config_implicit_top_drag_coeff ocn_grid="SOwISC12to60E2r4">4.48e-3</config_implicit_top_drag_coeff>
+<config_implicit_top_drag_coeff ocn_grid="ECwISC30to60E2r1">4.48e-3</config_implicit_top_drag_coeff>
 <config_use_implicit_bottom_roughness>.false.</config_use_implicit_bottom_roughness>
 <config_use_implicit_bottom_drag_variable>.false.</config_use_implicit_bottom_drag_variable>
 <config_use_implicit_bottom_drag_variable_mannings>.false.</config_use_implicit_bottom_drag_variable_mannings>
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -347,6 +347,10 @@
 <config_land_ice_flux_cp_ice>2.009e3</config_land_ice_flux_cp_ice>
 <config_land_ice_flux_rho_ice>918</config_land_ice_flux_rho_ice>
 <config_land_ice_flux_explicit_topDragCoeff>2.5e-3</config_land_ice_flux_explicit_topDragCoeff>
+<config_land_ice_flux_explicit_topDragCoeff ocn_grid="oEC60to30v3wLI">4.48e-3</config_land_ice_flux_explicit_topDragCoeff>
+<config_land_ice_flux_explicit_topDragCoeff ocn_grid="ECwISC30to60E1r2">4.48e-3</config_land_ice_flux_explicit_topDragCoeff>
+<config_land_ice_flux_explicit_topDragCoeff ocn_grid="SOwISC12to60E2r4">4.48e-3</config_land_ice_flux_explicit_topDragCoeff>
+<config_land_ice_flux_explicit_topDragCoeff ocn_grid="ECwISC30to60E2r1">4.48e-3</config_land_ice_flux_explicit_topDragCoeff>
 <config_land_ice_flux_ISOMIP_gammaT>1e-4</config_land_ice_flux_ISOMIP_gammaT>
 <config_land_ice_flux_rms_tidal_velocity>5e-2</config_land_ice_flux_rms_tidal_velocity>
 <config_land_ice_flux_jenkins_heat_transfer_coefficient>0.011</config_land_ice_flux_jenkins_heat_transfer_coefficient>
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -14,6 +14,7 @@
 <config_run_duration>'0010_00:00:00'</config_run_duration>
 <config_calendar_type CALENDAR="NO_LEAP">'noleap'</config_calendar_type>
 <config_calendar_type>'gregorian'</config_calendar_type>
+<config_output_reference_time>'0001-01-01_00:00:00'</config_output_reference_time>
 
 <!-- io -->
 <config_write_output_on_startup>.false.</config_write_output_on_startup>
@@ -257,12 +258,11 @@
 <config_cvmix_kpp_use_enhanced_diff>.true.</config_cvmix_kpp_use_enhanced_diff>
 <config_cvmix_kpp_nonlocal_with_implicit_mix>.false.</config_cvmix_kpp_nonlocal_with_implicit_mix>
 <config_cvmix_kpp_use_theory_wave>.false.</config_cvmix_kpp_use_theory_wave>
-<config_cvmix_kpp_use_active_wave>.false.</config_cvmix_kpp_use_active_wave>
 <config_cvmix_kpp_langmuir_mixing_opt>'NONE'</config_cvmix_kpp_langmuir_mixing_opt>
 <config_cvmix_kpp_langmuir_entrainment_opt>'NONE'</config_cvmix_kpp_langmuir_entrainment_opt>
+<config_cvmix_kpp_use_active_wave>.false.</config_cvmix_kpp_use_active_wave>
 
-
-<!-- wave coupling-->
+<!-- wave_coupling -->
 <config_use_active_wave>.false.</config_use_active_wave>
 <config_n_stokes_drift_wavenumber_partitions>6</config_n_stokes_drift_wavenumber_partitions>
 
@@ -346,11 +346,7 @@
 <config_land_ice_flux_boundaryLayerNeighborWeight>0.0</config_land_ice_flux_boundaryLayerNeighborWeight>
 <config_land_ice_flux_cp_ice>2.009e3</config_land_ice_flux_cp_ice>
 <config_land_ice_flux_rho_ice>918</config_land_ice_flux_rho_ice>
-<config_land_ice_flux_topDragCoeff>2.5e-3</config_land_ice_flux_topDragCoeff>
-<config_land_ice_flux_topDragCoeff ocn_grid="oEC60to30v3wLI">4.48e-3</config_land_ice_flux_topDragCoeff>
-<config_land_ice_flux_topDragCoeff ocn_grid="ECwISC30to60E1r2">4.48e-3</config_land_ice_flux_topDragCoeff>
-<config_land_ice_flux_topDragCoeff ocn_grid="SOwISC12to60E2r4">4.48e-3</config_land_ice_flux_topDragCoeff>
-<config_land_ice_flux_topDragCoeff ocn_grid="ECwISC30to60E2r1">4.48e-3</config_land_ice_flux_topDragCoeff>
+<config_land_ice_flux_explicit_topDragCoeff>2.5e-3</config_land_ice_flux_explicit_topDragCoeff>
 <config_land_ice_flux_ISOMIP_gammaT>1e-4</config_land_ice_flux_ISOMIP_gammaT>
 <config_land_ice_flux_rms_tidal_velocity>5e-2</config_land_ice_flux_rms_tidal_velocity>
 <config_land_ice_flux_jenkins_heat_transfer_coefficient>0.011</config_land_ice_flux_jenkins_heat_transfer_coefficient>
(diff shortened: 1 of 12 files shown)


cbegeman:
Compass testing and further discussion are located in https://github.com/E3SM-Ocean-Discussion/E3SM/pull/37.

Differences in the drag field between explicit and implicit options are demonstrated to be small for a compass test case with ice shelf cavities (`isomip_plus`), suggesting code correctness.

xylar:
I tried to do some testing on this but the namelist options haven't been propagated to E3SM yet.

@jonbob, I realize we lean on you quite heavily for such things.  Would you like me to take care of that or is it something you prefer to do yourself.  I'm fine either way.

jonbob:
@xylar -- thanks for the offer, but I'm happy to do this

jonbob:
@xylar -- done

xylar:
Thanks @jonbob.  I will try to get back to testing this later this week.

mark-petersen:
Compiled with gnu and intel on chrysalis. Passes nightly suite for stand-alone using debug for both, matches bfb against V2.1.0 tag.  That was all with `config_use_implicit_top_drag = .false.`, so it was just a compile and baseline test.

Then I ran the full nightly suite with gnu debug and `config_use_implicit_top_drag = .true.` crysalis. When there are no ice shelves it matches BFB with master (V2.1.0 tag). For global runs QU240wISC it runs successfully but does not match master, as expected:
```
ocean/global_ocean/QUwISC240/PHC/performance_test
  * step: forward
  test execution:      SUCCESS
  baseline comparison: FAIL
  see: case_outputs/ocean_global_ocean_QUwISC240_PHC_performance_test.log
  test runtime:        00:35
  * step: initial_state
  * step: ssh_adjustment
  * step: full_run
  * step: restart_run
  test execution:      SUCCESS
  test validation:     PASS
  baseline comparison: FAIL
  see: case_outputs/ocean_ice_shelf_2d_5km_z-star_restart_test.log
  test runtime:        00:10
  * step: initial_state
  * step: ssh_adjustment
  * step: full_run
  * step: restart_run
  test execution:      SUCCESS
  test validation:     PASS
  baseline comparison: FAIL
  see: case_outputs/ocean_ice_shelf_2d_5km_z-level_restart_test.log
  test runtime:        00:15
```
Differences with master are ~1e-3 for T, S, normalVelocity. This seems reasonable for a top drag change. So far so good!

components/mpas-ocean/src/shared/mpas_ocn_vmix.F
```diff
@@ -463,15 +467,18 @@ subroutine ocn_vel_vmix_tend_implicit(dt, kineticEnergyCell, vertViscTopOfEdge,
...
 
            ! tridiagonal matrix algorithm
            C(Nsurf)     = -2.0_RKIND*dt*vertViscTopOfEdge(Nsurf+1,iEdge) &
                       / (layerThickEdgeMean(Nsurf,iEdge) + layerThickEdgeMean(Nsurf+1,iEdge)) &
                       / layerThickEdgeMean(Nsurf,iEdge)
-           bTemp(Nsurf) = 1.0_RKIND - C(Nsurf)
+           bTemp(Nsurf) = 1.0_RKIND - C(Nsurf) &
+              + dt * implicitTopDragCoef * landIceEdgeFraction(iEdge) &
+                * sqrt(kineticEnergyCell(Nsurf,cell1) + kineticEnergyCell(Nsurf,cell2)) / &
+                layerThickEdgeMean(Nsurf,iEdge)
```
mark-petersen, line 481: This is where the action happens. I can see from my [original design document a long time ago](https://github.com/MPAS-Dev/MPAS-Documents/blob/master/ocean/implicit_vert_mix/implicit_vert_diff_design.pdf), that you can add top drag in the same way as bottom drag:
image
So in (3.42) above, you would add another equation for k=N_{surf}, where 
```
B_k = 1 - C_k  + \delta t * c_{drag} |u^n| / h^{n+1}
```
which is the line above. So that looks correct.

I do have a concern that `implicitBottomDragCoef` appears in line [499](https://github.com/E3SM-Project/E3SM/pull/5447/files#diff-c04cc906ba019d7e415af5dfef477ce54bb2e70ad8d59a23a35a58fe2b771787L499) (now 506) below in the backsolve, and there is no corresponding `implicitTopDragCoef`. I think that is because it is already in bTemp in line 504 (now 511), here
```
   normalVelocity(k,iEdge) = (rTemp(k) - C(k)*normalVelocity(k+1,iEdge)) / bTemp(k)
```
in the final solve for `normalVelocity`. 

@cbegeman can you just communicate your design and thought process here? I just want to make sure you worked through the algebra in the design document again.

cbegeman: @mark-petersen Thanks for your question and the thought you're giving this. I'm realizing now that I could have saved myself a bit of time by asking you for the design doc instead of tracking the terms through the code!

So as you've seen, in the single layer case, this drag term is just substituted directly into the `normalVelocity` solve on current line 470 rather than being stored in `bTemp`.

For the multi-layer case, I opted to include the implicit top drag only in `bTemp(Nsurf)` because this is the top layer. This particular drag formulation is only applied to the top layer, so the drag term does not appear in `bTemp(k)` where `k` ranges from `Nsurf+1` to `N-1`. The line that you pointed out, currently line 506, is where the bottom drag term is substituted directly into the `normalVelocity` solve for the bottom layer. As a reminder `N = maxLevelEdgeTop`. Since top drag is only applied to the top layer, we also don't need to include `implicitTopDragCoef` here at the `N` level.

Does that answer your question?

cbegeman:
@mark-petersen Thank you so much for your testing! That's consistent with what I found.

components/mpas-ocean/src/shared/mpas_ocn_vmix.F
mark-petersen: Yes, thanks. By the time I had finished writing my (lengthy) comment, I also came to that conclusion. But I thought I would post it anyway!

mark-petersen:
@cbegeman did you compare any velocity profiles below ice shelves with top drag on versus off? I've only verified that the prognostic variables differ by 1e-3. But it would be good to verify that velocities are lower just below the ice shelf. It would be easy to spot-check that with normalVelocity as a 2D field or a few columns.  Also, the global avg kinetic energy should be lower (perhaps by a fraction of a percent) between two runs. 

Could you please run a global ice shelf case out for 30 days and compare with and without top drag? Then plot global avg kinetic energy over time from global stats, and compare some velocities below the ice shelf. This would catch any sign errors in the new code.

Of course, the best test is against an analytic solution in a single column. I did that with bottom drag originally, but I think it is not required here if the global runs make sense qualitatively.

xylar:
@mark-petersen, I think the tests done in https://github.com/E3SM-Ocean-Discussion/E3SM/pull/37 show that the friciton velocity, barotropic streamfunction and overturning streamfunction are so similar between the explicit and implicit drag formulations that there is no room for anything as big as a sign error here.  I don't think there's any reason to suspect that a 30 day test on a global run will reveal anything that the ISOMIP+ testing didn't already.

Since we have to run 10 years with explicit and implicit top drag to indicate that the results aren't going to be climate changing, won't those tests be a more rigorous indication that the implicit drag formulation is correct than a comparison of drag vs. no drag after 30 days?

cbegeman:
@mark-petersen I agree with @xylar's response. The relevant comparison is here https://github.com/E3SM-Ocean-Discussion/E3SM/pull/37#issuecomment-1424489217, particularly note the `topDrag` diffs of O(1e-8).

mark-petersen (approved):
OK, thanks for pointing out that previous testing. I agree that the 10-year test will catch any surprises. Please compare the global avg kinetic energy from the 10-year run, as that is an easy check on whether the drag is working as expected. 

I'm approving based on my testing, review of the code, and the previous tests discussed here.

xylar:
@jonbob, I realized that I'm assuming you will (at the appropriate time) run the 10-year comparison here but please let me know if that isn't your assumption and you would prefer I run the tests.

We would need one run 10-year with master (and therefor explicit top drag) and another run with a test merge of this branch, but with implicit top drag instead.

xylar:
@jonbob, yes, I don't think there's any need to run a 10-year run without the feature turned on, since we expect things to be BFB without the stealth feature turned on, so just the 10-year runs I mentioned (test merge with the feature turned on and a reference run with master).  I think we're on the same page but just wanted to make sure.

xylar:
Oh, shoot!  I think the explanation is that the simulation with explicit top drag has:
```
config_land_ice_flux_explicit_topdragcoeff = 4.48e-3
```
but the simulation with implicit top drag has:
```
config_implicit_top_drag_coeff = 1.0e-3
```
(i.e. the same as the bottom drag).

@cbegeman, I think we need to make the default implicit top drag the same as the default explicit value.  Then, I would expect the simulations to be quite a bit more similar.

components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
```diff
@@ -377,6 +377,8 @@
 <!-- bottom_drag -->
 <config_use_implicit_bottom_drag>.true.</config_use_implicit_bottom_drag>
 <config_implicit_bottom_drag_coeff>1.0e-3</config_implicit_bottom_drag_coeff>
+<config_use_implicit_top_drag>.false.</config_use_implicit_top_drag>
+<config_implicit_top_drag_coeff>1.0e-3</config_implicit_top_drag_coeff>
```
xylar, line 381: @cbegeman, I think this value needs to be the same as for explicit top drag above:
```
<config_implicit_top_drag_coeff>2.5e-3</config_implicit_top_drag_coeff>
<config_implicit_top_drag_coeff ocn_grid="oEC60to30v3wLI">4.48e-3</config_implicit_top_drag_coeff>
<config_implicit_top_drag_coeff ocn_grid="ECwISC30to60E1r2">4.48e-3</config_implicit_top_drag_coeff>
<config_implicit_top_drag_coeff ocn_grid="SOwISC12to60E2r4">4.48e-3</config_implicit_top_drag_coeff>
<config_implicit_top_drag_coeff ocn_grid="ECwISC30to60E2r1">4.48e-3</config_implicit_top_drag_coeff>
```

xylar: Very sorry I missed this!

jonbob:
OK, new ten-year comparison test is here:

[20230401.5447test.anvil](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20230401.5447test.anvil/html/)

It looks much more similar, though still different at Amery

xylar:
Thanks so much @jonbob!

Everything there looks good to me -- I don't see anything climate changing.  The higher bottom temperatures and resulting melt rates around the Amery ice shelf are undesirable but this is a persistent feature of low res runs and just seems to get kicked off earlier here than in the control run.  I don't think that's a reason to consider this change to be climate changing.

cbegeman:
@jonbob Thanks.

I agree with @xylar that we need to remedy the Amery melt issue by other means.

Here's what I see:

* the ACC is stronger in East Antarctica (particularly in summer)
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20230401.5447test.anvil/html/ocean/index.html#antarctic_velocityMagnitudeSose&gid=14&pid=1

* More mCDW intrusion at Amery
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20230401.5447test.anvil/html/ocean/index.html#sose_transects&gid=35&pid=5

* A lot more melting at Amery
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20230401.5447test.anvil/html/ocean/index.html#antmelttime&gid=24&pid=11

* A little more sea ice melting near Amery
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20230401.5447test.anvil/html/sea_ice/index.html#antarctic_extended_sh_conc&gid=8&pid=1

jonbob:
@cbegeman, @xylar -- because this is a stealth feature, we need to include a test that turns the new feature on, even it it's off by default. I have made a testdef that simply does that in user_nl_mpaso, but I want to check with you about configurations to test. Part of the automated testing is that it should not take a large amount of resources, so I was thinking about either:
* T62_oQU240wLI.GMPAS-DIB-IAF-ISMF.mpaso-impl_top_drag
* ne4_oQU240wLI.CRYO1850NS.mpaso-impl_top_drag
I don't believe the second configuration is fully supported -- it's missing mapping files and the compset alias. But it wouldn't take much to get working if we would prefer a fully-coupled test.

cbegeman:
@jonbob I think the first test is sufficient. What do you think @xylar?

xylar:
@jonbob and @cbegeman, I would also think the G-case test would be sufficient.  If it turns out that we need a very low res B-case for other testing, that second configuration does seem like a good choice but I don't think this is a feature that needs a test with full coupling to verify that it's BFB.

Thanks for working on this @jonbob!

jonbob:
new test runs fine:
````
PASS SMS_D_Ld1.T62_oQU240wLI.GMPAS-IAF-ISMF.chrysalis_intel.mpaso-impl_top_drag CREATE_NEWCASE
PASS SMS_D_Ld1.T62_oQU240wLI.GMPAS-IAF-ISMF.chrysalis_intel.mpaso-impl_top_drag XML
PASS SMS_D_Ld1.T62_oQU240wLI.GMPAS-IAF-ISMF.chrysalis_intel.mpaso-impl_top_drag SETUP
PASS SMS_D_Ld1.T62_oQU240wLI.GMPAS-IAF-ISMF.chrysalis_intel.mpaso-impl_top_drag SHAREDLIB_BUILD time=119
PASS SMS_D_Ld1.T62_oQU240wLI.GMPAS-IAF-ISMF.chrysalis_intel.mpaso-impl_top_drag MODEL_BUILD time=488
PASS SMS_D_Ld1.T62_oQU240wLI.GMPAS-IAF-ISMF.chrysalis_intel.mpaso-impl_top_drag SUBMIT
PASS SMS_D_Ld1.T62_oQU240wLI.GMPAS-IAF-ISMF.chrysalis_intel.mpaso-impl_top_drag RUN time=27
PASS SMS_D_Ld1.T62_oQU240wLI.GMPAS-IAF-ISMF.chrysalis_intel.mpaso-impl_top_drag MEMLEAK insuffiencient data for memleak test
PASS SMS_D_Ld1.T62_oQU240wLI.GMPAS-IAF-ISMF.chrysalis_intel.mpaso-impl_top_drag SHORT_TERM_ARCHIVER
````

pushed new commit to support testdef capability in mpaso and add this test to a new automated suite

jonbob:
after testing changes, running
````
./create_test e3sm_ocnice_stealth_features
````
now successfully runs SMS_D_Ld1.T62_oQU240wLI.GMPAS-IAF-ISMF.mpaso-impl_top_drag, the stealth feature test

cbegeman:
@jonbob Great! Thanks for adding that test!

jonbob:
@cbegeman -- you're welcome. And I think this can now go in tomorrow

jonbob:
passes:

jonbob:

the new test was missing graph files for gcp12, so some follow-up will be needed to create a baseline there

PR #5453, opened 2023-02-14
Remove special hypervis defaults for v2 ne30

brhillman:
Remove special hyperviscosity override for v2 ne30 configurations. This will be non-BFB only for ne30 tests. Note that answers with v2 will still be preserved with the maint-2.0 branch.

[non-BFB] (only for ne30)

### DIFF
--- a/components/eam/bld/namelist_files/namelist_defaults_eam.xml
+++ b/components/eam/bld/namelist_files/namelist_defaults_eam.xml
@@ -1278,13 +1278,6 @@
 <hypervis_order >     2 </hypervis_order>
 <hypervis_scaling> 3.0 </hypervis_scaling>
 
-<!-- special defaults for V2 NE30 -->
-<nu hgrid="ne30np4">  1.0e15 </nu>
-<hypervis_scaling hgrid="ne30np4" > 0 </hypervis_scaling>
-
-
-
-
 <!-- -1.0 indicates use the value for nu -->
 <nu_q> -1.0 </nu_q>
 <nu_p> -1.0 </nu_p>


mt5555:
This is a good idea, and it unifies our HV so it is the same for all resolutions, both cube-sphere and RRM.   The NE30 special case is a holdover from not wanting to introduce atmosphere changes during the v2 development.

However, this change does represent a slight reduction in dissipation, and thus will have a minor effect on the climate for the ongoing V3 work.   Hence, just to be super safe, (and follow our new code review / new feature process), it should be accompanied by a NE30 5 year F case comparison (old and new), and get @golaz 's approval.

rljacob:
status:  @brhillman needs to run a few 5year test cases.

brhillman:
Sorry this took so long, was on vacation last week. I ran two five year F-cases, one with and one without this fix to remove the override for ne30 hyper viscosity, and ran E3SM diags on the 5 year climatology. The results for *before* the hyper viscosity changes are here:

https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.brhillman/e3sm/master.ne30pg2_EC30to60E2r2.F2010.chrysalis_intel.2752x1/

And the results *after* the hyper viscosity changes are here:

https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.brhillman/e3sm/fix-ne30-hypervis-defaults.ne30pg2_EC30to60E2r2.F2010.chrysalis_intel.2752x1/

I glanced at a few fields, and didn't see anything weird, but @golaz and @mt5555 should probably take a look.

brhillman:
@whannah1 pointed out that precipitation does look different, in fact it appears to be "improved" slightly (whatever that means for a 5-year F-case).



vs

crterai (approved):
I agree with Ben's assessment that the simulation biases are generally reduced with the simulation. Because it is a dycore related change, it is nice to see that the dynamics (U850 and U200) are actually improved. Also nice to see that the pressure-latitude biases of omega are also improved, particularly in the JJA season. 
Based on the diagnostics, I think it is okay to merge, but defer to @golaz's decision.

golaz (approved):
@brhillman : these results are very impressive, with some significant improvements, especially seasonally (JJA).

wlin7:
The [IICE link ](https://portal.nersc.gov/project/m2136/bin/iice/iice.cgi?url1=https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.brhillman/e3sm/master.ne30pg2_EC30to60E2r2.F2010.chrysalis_intel.2752x1/e3sm_diags/atm_monthly_180x360_aave/model_vs_obs_0001-0005/viewer&label1=CTRL&url2=https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.brhillman/e3sm/fix-ne30-hypervis-defaults.ne30pg2_EC30to60E2r2.F2010.chrysalis_intel.2752x1/e3sm_diags/atm_monthly_180x360_aave/model_vs_obs_0001-0005/viewer&label2=New&category=&plot=&diff=0&dconfig=vsObs&)if to view the comparison of more fields.

golaz:
@mt5555 : based on these results, it's quite obvious that there are significant opportunities for bias reduction through tuning in the dynamical core. Do we know that we have the optimal settings, or should there be a more systematic exploration?

tangq:
The zonal wind is significantly improved at both 850 hPa and 200 hPa, especially for JJA means.

rljacob:
> we want to use the tensor hyperviscosity for all resolutions including ne30 going forward

Does that include ne4?   If so, why don't its answers change?

mt5555:
@rljacob :  tensor HV is used by all RRM configurations (no choice there).  A while ago we made tensor HV the default for all cubed-sphere grids, *except* NE30, so as not to change results during v2 development.  So this change only effects NE30.

To @golaz question:  that's a good question.  I'm surprised at these improvements.  I think they are due to the fact that in switching to tensorHV, we effectively reduced the diffusion coefficient by  70%.  

With regards to further tuning, Ben since you are already set up, can you make one more run with the default coefficient (new TensorHV option) changed from 3.4e-8 to 1.7e-8?    My guess is this run will crash, suggesting we are pretty close to the stability limit and thus there is not much room for further tuning.  For this coefficient, "tuning" usually means running a little bit above the minimum stable value.  

Also, are these fields (U200, U850, PRECT) robust with 5 year climatologies?

mt5555 (approved):

brhillman:
> With regards to further tuning, Ben since you are already set up, can you make one more run with the default coefficient (new TensorHV option) changed from 3.4e-8 to 1.7e-8?

Yes of course, building now.

brhillman:
Update with the reduced hyper viscosity test (nu = 1.7e-8): this case ran without crashing (5-years). e3sm-diags here:

https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.brhillman/e3sm/fix-ne30-hypervis-defaults.ne30pg2_EC30to60E2r2.F2010.chrysalis_intel.2752x1.nu17/e3sm_diags/atm_monthly_180x360_aave/

Notably, here is the comparison for zonal winds:
U850_comparison

So the maximum biases against MERRA2 actually increases with the reduced HV, but RMSE and correlation are still improved relative to the current formulation and tuning in master. It might be worth doing a little further investigation on HV tuning. @mt5555 suggested doing a run with the old HV formulation but with correspondingly lower diffusion to verify that it is in fact the reduction in diffusion leading to the improvement relative to the old formulation, rather than the formulation itself. Perhaps a few additional runs to also make sure we are using optimal values. Since this is all setup, I'd be happy to do a little further exploration, but we probably do not need to hold up this PR to do that.

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #5454, opened 2023-02-15
Updates multiple tests to use land initial condition

bishtgautam:
Protect land from breaking restart by using an initial condition in tests.

- Seven production tests are renamed and updated to use a land initial condition.
- One land-only test is updated to use a land initial condition.

[non-BFB] for prod tests


cime_config/tests.py
```diff
@@ -140,8 +140,8 @@
     #atmopheric tests to mimic low res production runs
     "e3sm_atm_prod" : {
         "tests" : (
-            "SMS_Ln5.ne30pg2_r05_oECv3.F2010.eam-wcprod",
```
rljacob, line 143: Why not just modify the existing wcprod testmod dirs?

bishtgautam: @rljacob Sorry, I missed this comment. Presently, multiple tests that have different resolutions are sharing the same testmod dir. So, I had to create individual testmod dir for each test in which `user_nl_elm` has the appropriate `finidat`.

rljacob (approved):

bishtgautam:
@peterdschwartz Can you please merge this PR?

peterdschwartz:
pushed to next

rljacob:
@bishtgautam Sorry I meant to ask more questions about this before it was merged.

Why is necessary to give, for example, wcprod_1850 and wcprod_1850_1pctCO2 different finidat files?

bishtgautam:
@rljacob I wasn't certain if a few compsets can share a common finidat file. So, to be safe, I let each compset have separate finidat file.

rljacob:
"wcprod" is generic enough that it can be used on any fully-coupled case, any resolution.  But the addition of these finidat files will tie the new testmods to a specific resolution. I think resolution should be in the dir name as well to reflect that.  You did it for a couple by adding "r5" but wcprod_1850 should be wcprod_1850ne30pg2 ?

bishtgautam:
You are absolutely right and I should have added the resolution name. I will work on opening another PR next week that will add the resolution name to testmods.

Issue #5258, opened 2022-10-28
tests fail with MEMLEAK using nvidia compiler on pm-cpu


ndkeen:
The following two tests have a MEMLEAK fail with nvidia (22.2) on pm-cpu.  Also fails same way with DEBUG build.

```
SMS_Ld20.f45_f45.IELMFATES.pm-cpu_nvidia.elm-fates_eca
SMS_Ld20.f45_f45.IELMFATES.pm-cpu_nvidia.elm-fates_rd
```

I also see same MEMLEAK fail with 1 thread.
`SMS_PMx1_Ld20.f45_f45.IELMFATES.pm-cpu_nvidia.elm-fates_eca`

PR #5462, opened 2023-02-17
For pm-cpu, increase MEMLEAK tolerance to 0.2

ndkeen:
The GNU e3sm_integration suite has 2 tests that fail with MEMLEAK and increasing tolerance allows to pass.

```
SMS_Ld20.f45_f45.IELMFATES.pm-cpu_gnu.elm-fates_eca
SMS_Ld20.f45_f45.IELMFATES.pm-cpu_gnu.elm-fates_rd
```

[bfb]


ndkeen:
I just ran it for 120 days and see the memory grows from 390 to 547MB, but then drops back to 472MB in the last several days.

rljacob:
@glemieux does the memory growth described above sound like something coming from FATES?

rljacob (approved):

ndkeen:
Running 6 months here:
`/pscratch/sd/n/ndk/e3sm_scratch/pm-cpu/SMS_Lm6.f45_f45.IELMFATES.pm-cpu_gnu.elm-fates_eca.20230217_140944_7bhmfe`

ndkeen:
merged to next

rgknox:
FATES should spawn memory as the biodiversity of the forests increase. This is typical and expected in a cold-start scenario, which is what we currently use for testing.  My take is that all the memory tests are red-herrings for FATES.  In short though, increasing the tolerance makes sense.

glemieux:
@rljacob yes this is typical for fates given particularly during cold starts in which fates is allocating objects on the fly through the spawning of patches to fill up the available site area.  The rate at which this happens can be variable depending on parameter settings, grid cell conditions and fates settings which could lead to large increases for some sites and slower increases for others.  Given the developmental nature of fates this means that as we improve the model we may find in comparing against baselines that there has been an increase in memory usage as sites better represent expected outcomes.  My assumption is that our current calibration efforts may help improve this somewhat, but as far as I know we haven't really tried to quantify an expected rate due to the complexity of the issue.

rljacob:
Thanks.  That is good information.

ndkeen:
Fixes https://github.com/E3SM-Project/E3SM/issues/5258

PR #5458, opened 2023-02-17
Bugfix for ocean submesoscale feature

cbegeman:
Fill value masking of the submesoscale variable `normalMLEVelocity` resulted in bad values in `vertMLEBolusVelocityTop` that affected `timeMonthly_avg_mocStreamvalLatAndDepth`. This bugfix removes masking of `normalMLEVelocity`. 

Fixes #5459

[BFB]

PR #5254, opened 2022-10-27
Modify ocean z-star ALE coordinate for inactive top cells

cbegeman:
This PR increases the generalizability of vertical coordinates by relaxing the assumption that `restingThickness` spans `ssh=0` to `bottomDepth`. As such, it is tangentially related to inactive top cells. The change affects cases where `config_ALE_thickness_proportionality='restingThickness_times_weights'`, which is the default.

[non-BFB]

PR #5464, opened 2023-02-19
New ocean variables for floating land ice

cbegeman:
This PR adds new variables to represent floating land ice. This is a necessary step toward coupling with an ice sheet model component (MALI).

- `landIceMask` and `landIceFraction` (existing variables) now represent both grounded and floating land ice.
- `landIceFloatingMask` and `landIceFloatingFraction (new variables)` represent only floating land ice.

The `landIceFloating*` variables are used to modulate land ice fluxes, whereas the `landIce*` variables are used to mask out certain fields, as before. `landIce*` variables also modulate top drag, as we want to include the effect of top drag on thin film regions to slow flow.

[NML]
[non-BFB]

### DIFF
--- a/components/mpas-ocean/driver/ocn_comp_mct.F
+++ b/components/mpas-ocean/driver/ocn_comp_mct.F
@@ -790,10 +790,9 @@ contains
     call t_stopf ('mpaso_mct_init')
 
     call mpas_pool_get_config(domain % configs, 'config_land_ice_flux_mode', config_land_ice_flux_mode)
-    if ( trim(config_land_ice_flux_mode) .eq. 'pressure_only' ) then
-       call seq_infodata_PutData( infodata, ocn_prognostic=.true., ocnrof_prognostic=.true., &
-                                            ocn_c2_glcshelf=.false.)
-    else if ( trim(config_land_ice_flux_mode) .eq. 'standalone' ) then
+    if ( trim(config_land_ice_flux_mode) == 'off' .or. &
+         trim(config_land_ice_flux_mode) == 'pressure_only' .or. &
+         trim(config_land_ice_flux_mode) == 'standalone' ) then
        call seq_infodata_PutData( infodata, ocn_prognostic=.true., ocnrof_prognostic=.true., &
                                             ocn_c2_glcshelf=.false.)
     else if ( trim(config_land_ice_flux_mode) .eq. 'coupled' ) then
@@ -2784,7 +2783,8 @@ contains
 !JW       o2x_o % rAttr(index_o2x_So_htv, n) = landIceHeatTransferVelocity(i)
 !JW       o2x_o % rAttr(index_o2x_So_stv, n) = landIceSaltTransferVelocity(i)
  
-       if ( trim(config_land_ice_flux_mode) .ne. 'pressure_only' ) then
+       if ( trim(config_land_ice_flux_mode) .eq. 'standalone' .or. &
+            trim(config_land_ice_flux_mode) .eq. 'coupled'  ) then
           o2x_o % rAttr(index_o2x_So_blt, n) = landIceBoundaryLayerTracers(indexBLT,i)
           o2x_o % rAttr(index_o2x_So_bls, n) = landIceBoundaryLayerTracers(indexBLS,i)
           o2x_o % rAttr(index_o2x_So_htv, n) = landIceTracerTransferVelocities(indexHeatTrans,i)
--- a/components/mpas-ocean/src/shared/mpas_ocn_eddy_parameterization_helpers.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_eddy_parameterization_helpers.F
@@ -197,7 +197,6 @@ module ocn_eddy_parameterization_helpers
      integer :: iEdge, nEdges, refIndex, cell1, cell2, nCells, k, iCell
      real(kind=RKIND) :: dDenThres, den_ref_lev
      real(kind=RKIND),dimension(:), allocatable :: depth
-     integer, dimension(:), pointer :: landIceFloatingMask
      real (kind=RKIND), dimension(:,:), pointer :: layerThickness
      real (kind=RKIND), dimension(:), pointer :: landIceDraft, landIcePressure
      logical :: found_den_mld
@@ -209,7 +208,6 @@ module ocn_eddy_parameterization_helpers
      if ( config_eddyMLD_use_old ) then
        call mpas_pool_get_array(forcingPool, 'landIcePressure', landIcePressure)
        call mpas_pool_get_array(forcingPool, 'landIceDraft', landIceDraft)
-       call mpas_pool_get_array(forcingPool, 'landIceFloatingMask', landIceFloatingMask)
 
        allocate(pressureAdjustedForLandIce(nVertLevels, size(pressure,2)))
 
@@ -281,11 +279,11 @@ module ocn_eddy_parameterization_helpers
         !$omp end do
         !$omp end parallel
 
-        if (associated(landIceFloatingMask) ) then
+        if (associated(landIceDraft) ) then
           !$omp parallel
           !$omp do schedule(runtime)
           do iCell = 1, nCells
-             dThreshMLD(iCell) = dThreshMLD(iCell) - abs(landIceDraft(iCell))*landIceFloatingMask(iCell)
+             dThreshMLD(iCell) = dThreshMLD(iCell) - abs(landIceDraft(iCell))
           end do
           !$omp end do
           !$omp end parallel
--- a/components/mpas-ocean/src/shared/mpas_ocn_eddy_parameterization_helpers.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_eddy_parameterization_helpers.F
@@ -197,7 +197,7 @@ module ocn_eddy_parameterization_helpers
      integer :: iEdge, nEdges, refIndex, cell1, cell2, nCells, k, iCell
      real(kind=RKIND) :: dDenThres, den_ref_lev
      real(kind=RKIND),dimension(:), allocatable :: depth
-     integer, dimension(:), pointer :: landIceMask
+     integer, dimension(:), pointer :: landIceFloatingMask
      real (kind=RKIND), dimension(:,:), pointer :: layerThickness
      real (kind=RKIND), dimension(:), pointer :: landIceDraft, landIcePressure
      logical :: found_den_mld
@@ -209,7 +209,7 @@ module ocn_eddy_parameterization_helpers
      if ( config_eddyMLD_use_old ) then
        call mpas_pool_get_array(forcingPool, 'landIcePressure', landIcePressure)
        call mpas_pool_get_array(forcingPool, 'landIceDraft', landIceDraft)
-       call mpas_pool_get_array(forcingPool, 'landIceMask', landIceMask)
+       call mpas_pool_get_array(forcingPool, 'landIceFloatingMask', landIceFloatingMask)
 
        allocate(pressureAdjustedForLandIce(nVertLevels, size(pressure,2)))
 
@@ -281,11 +281,11 @@ module ocn_eddy_parameterization_helpers
         !$omp end do
         !$omp end parallel
 
-        if (associated(landIceMask) ) then
+        if (associated(landIceFloatingMask) ) then
           !$omp parallel
           !$omp do schedule(runtime)
           do iCell = 1, nCells
-             dThreshMLD(iCell) = dThreshMLD(iCell) - abs(landIceDraft(iCell))*landIceMask(iCell)
+             dThreshMLD(iCell) = dThreshMLD(iCell) - abs(landIceDraft(iCell))*landIceFloatingMask(iCell)
           end do
           !$omp end do
           !$omp end parallel
--- a/components/mpas-ocean/bld/build-namelist
+++ b/components/mpas-ocean/bld/build-namelist
@@ -782,7 +782,7 @@ if ($OCN_ISMF eq 'coupled') {
 } elsif ($OCN_ISMF eq 'internal') {
 	add_default($nl, 'config_land_ice_flux_mode', 'val'=>"standalone");
 } else {
-	add_default($nl, 'config_land_ice_flux_mode', 'val'=>"pressure_only");
+	add_default($nl, 'config_land_ice_flux_mode');
 }
 add_default($nl, 'config_land_ice_flux_formulation');
 add_default($nl, 'config_land_ice_flux_useHollandJenkinsAdvDiff');
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -333,12 +333,12 @@
 
 <!-- land_ice_fluxes -->
 <config_land_ice_flux_mode>'off'</config_land_ice_flux_mode>
-<config_land_ice_flux_mode ocn_grid="oQU240wLI">'standalone'</config_land_ice_flux_mode>
-<config_land_ice_flux_mode ocn_grid="oEC60to30v3wLI">'standalone'</config_land_ice_flux_mode>
-<config_land_ice_flux_mode ocn_grid="ECwISC30to60E1r2">'standalone'</config_land_ice_flux_mode>
-<config_land_ice_flux_mode ocn_grid="oRRS30to10v3wLI">'standalone'</config_land_ice_flux_mode>
-<config_land_ice_flux_mode ocn_grid="SOwISC12to60E2r4">'standalone'</config_land_ice_flux_mode>
-<config_land_ice_flux_mode ocn_grid="ECwISC30to60E2r1">'standalone'</config_land_ice_flux_mode>
+<config_land_ice_flux_mode ocn_grid="oQU240wLI">'pressure_only'</config_land_ice_flux_mode>
+<config_land_ice_flux_mode ocn_grid="oEC60to30v3wLI">'pressure_only'</config_land_ice_flux_mode>
+<config_land_ice_flux_mode ocn_grid="ECwISC30to60E1r2">'pressure_only'</config_land_ice_flux_mode>
+<config_land_ice_flux_mode ocn_grid="oRRS30to10v3wLI">'pressure_only'</config_land_ice_flux_mode>
+<config_land_ice_flux_mode ocn_grid="SOwISC12to60E2r4">'pressure_only'</config_land_ice_flux_mode>
+<config_land_ice_flux_mode ocn_grid="ECwISC30to60E2r1">'pressure_only'</config_land_ice_flux_mode>
 <config_land_ice_flux_formulation>'Jenkins'</config_land_ice_flux_formulation>
 <config_land_ice_flux_useHollandJenkinsAdvDiff>.false.</config_land_ice_flux_useHollandJenkinsAdvDiff>
 <config_land_ice_flux_attenuation_coefficient>10.0</config_land_ice_flux_attenuation_coefficient>
--- a/components/mpas-ocean/cime_config/buildnml
+++ b/components/mpas-ocean/cime_config/buildnml
@@ -90,10 +90,10 @@ def buildnml(case, caseroot, compname):
         decomp_prefix = 'mpas-o.graph.info.'
         restoring_file = 'sss.monthlyClimatology.PHC2_salx_040803.oEC60to30v3wLI.nc'
         analysis_mask_file = 'masks_SingleRegionAtlanticWTransportTransects_EC60to30v3wLI_171116.nc'
-        ic_date = '171031'
+        ic_date = '230220'
         ic_prefix = 'oEC60to30v3wLI60lev'
         if ocn_ic_mode == 'spunup':
-            ic_date = '171116'
+            ic_date = '230220'
             ic_prefix = 'oEC60to30v3wLI60lev.restart_theta_year26'
 
     elif ocn_grid == 'ECwISC30to60E1r2':
@@ -101,7 +101,7 @@ def buildnml(case, caseroot, compname):
         decomp_prefix = 'mpas-o.graph.info.'
         restoring_file = 'sss.PHC2_monthlyClimatology.ECwISC30to60E1r02.200408.nc'
         analysis_mask_file = 'ECwISC30to60E1r02_transportTransects.nc'
-        ic_date = '200408'
+        ic_date = '230220'
         ic_prefix = 'ocean.ECwISC30to60E1r2'
         if ocn_ic_mode == 'spunup':
             logger.warning("WARNING: The specified compset is requesting ocean ICs spunup from a G-case")
@@ -128,7 +128,7 @@ def buildnml(case, caseroot, compname):
     elif ocn_grid == 'oQU240wLI':
         decomp_date = '160929'
         decomp_prefix = 'mpas-o.graph.info.'
-        ic_date = '160929'
+        ic_date = '230220'
         ic_prefix = 'ocean.QU.240wLI'
         if ocn_ic_mode == 'spunup':
             logger.warning("WARNING: The specified compset is requesting ocean ICs spunup from a G-case")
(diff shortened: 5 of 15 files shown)


cbegeman:
See https://github.com/E3SM-Ocean-Discussion/E3SM/pull/38 for further discussion

cbegeman:
### Testing

Machine: chrysalis
Compiler/MPI lib: intel, openmpi 

The PR test suite passes comparison with baseline for all tests except `ocean_isomip_plus_2km_z-star_Ocean0` which is to be expected due to non-BFB changes in the initial masks (resulting in land ice flux differences at cells that were fully grounded in master and are now partially grounded). Note that the cached outputs cannot be used for the global case with ice shelf cavities because the `landIceFloating*` variables must be included in streams.

Furthermore, I ran all of the `QUwISC240` tests. All but the analysis test pass, which is expected due to this commit https://github.com/E3SM-Ocean-Discussion/E3SM/pull/38/commits/4212296d65f022c7b2d8879207b375699053f191.

```
tThreshMLD           Time index: 0, 1, 2, 3 0:  l1: 1.39241851630634e+03  l2: 3.25246576969646e+02  linf: 2.40783212720566e+02 1:  l1: 1.39241851630634e+03  l2: 3.25246576969646e+02  linf: 2.40783212720566e+02 2:  l1: 1.39241851630634e+03  l2: 3.25246576969646e+02  linf: 2.40783212720566e+02 3:  l1: 1.39241851630634e+03  l2: 3.25246576969646e+02  linf: 2.40783212720566e+02 FAIL

**Comment by xylar** [comment]
@jonbob, I'm marking this as `non-BFB` for now.  We changed how the mixed layer depth is computed in this PR and that definitely affects some analysis members.  In @cbegeman's testing it doesn't affect the dynamics in a `QUwISC240` test but that's surprising to me and I expect non-BFB but non-climate-changing differences for runs with ice-shelf cavities in general.

**Comment by xylar** [comment]
As I have discussed with @cbegeman, this can't go in until we get new initial conditions for all `wISC` runs that include the new variables (just copies of existing fields).  I will work on that soon.

**Comment by xylar** [comment]
I used commands like:
```
ncap2 -O \
  -s "landIceFloatingFraction=landIceFraction;landIceFloatingMask=landIceMask" \
  $src $dst
```
to add the new variables.

File: components/mpas-ocean/src/shared/mpas_ocn_eddy_parameterization_helpers.F [hunk 1]
```diff
@@ -284,11 +282,11 @@ subroutine ocn_eddy_compute_mixed_layer_depth(statePool, forcingPool)
...
         !$omp end parallel
 
-        if (associated(landIceMask) ) then
+        if (associated(landIceDraft) ) then
           !$omp parallel
           !$omp do schedule(runtime)
           do iCell = 1, nCells
-             dThreshMLD(iCell) = dThreshMLD(iCell) - abs(landIceDraft(iCell))*landIceMask(iCell)
+             dThreshMLD(iCell) = dThreshMLD(iCell) - abs(landIceDraft(iCell))
           end do
```
xylar, line 290: @vanroekel, this is the piece of code I'd like to get your input on.  It's the only thing that I think will lead to non-BFB results in fields that could feed back on model dynamics.

I think I masked this with the `landIceMask` before without really thinking about the consequences.  But the ice draft is smoothed near the ice-shelf calving front on both the iceward and seaward sides.  It seems like we should remove this ice draft (the amount that the sea surface is depressed by the weight of an ice shelf) even in open ocean areas where it is nonzero (maybe 10 cells seaward of an ice shelf at most).

I just wanted to check with you because you'll have the best sense of what fields we should be testing for changes as a result of modifying `dThreshMLD`.  Which parts of the eddy parameterizaiton are affected?

vanroekel: @xylar thanks for the explanation on the change.  I agree that if we have an added pressure on the sea surface we should account for this and so it makes sense to remove that iceMask.

At minimum this will cause some changes in the submesoscale eddy parameterization, my feeling is it will strengthen as I'd expect MLD to deepen with this change where the ice pressure wasn't being removed previously.  So checking the MLEvelocity (normal and vertical) fields would be a good start.

If you have redi enabled, I'd expect very slight changes there as well.  For that one you could take a look at the temperature_hmixTendency (or salinity) variable, assuming the tracer budgets are on.  But again am expecting small changes.

xylar: @vanroekel, thanks for your intuition on what will change.

> I'd expect MLD to deepen with this change

I could be wrong but I expect the opposite -- I think the MLD was artificially deep because the old code didn't account for the fact that much of that depth was actually because the sea surface was suppressed by land ice.

So it may be that we would expect a weakening of those parameterizations around ice-shelf fronts.  I'll take a look and see which it is.

Thanks again!

vanroekel: Sorry you are right @xylar I thought this bit of code was correcting the pressure when finding the reference pressure level, but is a correction to the MLD itself.  Yes MLD will definitely shallow with this change.

xylar:
@jonbob, I'd like to do a bit-for-bit test for each of the 6 meshes with ice-shelf cavities.  Is there a way to do that with the E3SM testing infrastructure (i.e. do we have short test runs for all 6 meshes)?  If not, do you have an alternative suggestion for how I could do this efficiently?

xylar:
@jonbob, following up, I'm going to see how far I can get with tests like:
```
./create_test --wait --walltime 1:00:00 -g -b master_20230221 --baseline-root /lcrc/group/e3sm/ac.xylar/e3sm_baselines SMS_Ln9.T62_oEC60to30v3wLI.GMPAS-DIB-IAF-ISMF.chrysalis_intel
```
Let me know if you suggest something different.

jonbob:
@xylar -- that's pretty much what I was going to suggest

xylar:
Holy cow! It works!

xylar:
I'll have to pick this up tomorrow but I should have my 6 baselines by then.

xylar:
I ran 6 baseline tests using 0a642b74e2 as the baseline (the commit on master where this branch was created).

5 of 6 show diffs in many fields:
New ICs:
- [x] `oEC60to30v3wLI`
- [x] `ECwISC30to60E1r2`
- [x] `oRRS30to10v3wLI`
- [x] `SOwISC12to60E2r4`
- [x] `ECwISC30to60E2r1` 

One test, `oQU240wLI`, fails for reasons I don't yet understand:
```
/lcrc/group/e3sm/ac.xasay-davis/scratch/chrys/SMS_Ln9.T62_oQU240wLI.GMPAS-DIB-IAF-ISMF.chrysalis_intel.C.20230222_105814_ja9g3r
```

Overall, my results are in the subdirectories of:
```
/lcrc/group/e3sm/ac.xasay-davis/scratch/chrys/
```

cbegeman:
@xylar Regarding the failure, 

I'm seeing the following error message:
```
ERROR:  Warning: Sea surface height outside of acceptable range (20m)
```

I might have seen this error in relation to https://github.com/E3SM-Project/E3SM/pull/5254.

xylar:
@cbegeman, I think the actual issue is just that oQU240wLI was never set up to use data icebergs.  So both the baseline and the comparison runs failed.  I think the error about the SSH being outside of bounds is just because there's a check that should be turned off, but as you say it's unrelated to the main failure.

xylar:
From the sea ice log file for both oQU240wLI runs:
```
WARNING: Error: Sub-pool berg_forcing_forcing_input not found in pool.
ERROR: Stream 'dataIcebergForcing' attempted to read non-existent file '/lcrc/group/e3sm/data/inputdata/ice/mpas-cice/oQU240wLI/'
CRITICAL ERROR: Forcing: get_initial_forcing_data: READ: -3
```

xylar:
@jonbob, I think this is looking okay and it's time to set up a 10-year run to make sure the results are not climate changing.  Let's presumably do a test merge with #5458 (or wait for that to go in) before we do that run.  Otherwise, the MOC can't be included in the comparison.

jonbob:
@xylar -- my plan is to get #5458 in today, and then I can start a couple of longer runs to test the impact. I was thinking to use the ECwISC30to60E2r1 configuration, but open to other ideas

xylar:
@jonbob, yes, that's what I would use and what I've been testing ice-shelf work (like #5447) with.

jonbob:
Comparison of 10-year test with baseline, using ne30pg2_ECwISC30to60E2r1 and CRYO1850 configurations, are available at: [20230227.5464test.anvil](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20230227.5464test.anvil/html/). @xylar and @cbegeman, please make sure this looks OK to you -- and invite anyone else to review based on these results.

cbegeman:
@xylar and @jonbob I think this looks reasonable. It made a larger difference for Amery than I would have expected, but I don't think anything is wrong with the implementation. I'm curious to see what @xylar thinks.

xylar:
Thanks @jonbob!  This analysis is really helpful.

@cbegeman, I agree that the changes around Amery are surprisingly large.  But based on the spatial patterns of the salinity and temperature changes (.e.g https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20230227.5464test.anvil/html/ocean/temperatureSOSE_antarctic_20230227.5464test.anvil_depth_top_JFM_years0001-0010.png and https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20230227.5464test.anvil/html/ocean/salinitySOSE_antarctic_20230227.5464test.anvil_depth_top_JFM_years0001-0010.png) I would say this is ensemble variability rather than a direct result of the change in mixed layer depth.

These plots suggest that MLD changes themselves are not confined to the regions just seaward of calving fronts where the code change would be expected to have an effect:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20230227.5464test.anvil/html/ocean/mixedLayerDepthSOSE_antarctic_20230227.5464test.anvil_JFM_years0001-0010.png
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20230227.5464test.anvil/html/ocean/mixedLayerDepthSOSE_antarctic_20230227.5464test.anvil_JAS_years0001-0010.png
Instead, the changes here, too, look more like ensemble variability.

I'm not seeing anything here that has me worried.  I think this PR is not climate changing, as we hoped.

xylar:
@vanroekel, just waiting on your approval based on my questions about the MLD changes. It looked like you were okay with them.

jonbob:
@cbegeman -- I went to test this before merging and now getting an error from mpaso:
````
CRITICAL ERROR: landIceFloatingMask contains the default value which likely indicates that this field is missing in the initial condition file (e.g. because it is meant for an older E3SM version).
````
But the tests all use non-wISC grids -- my longer run was with ECwISC30to60E2r1.

**Comment by cbegeman** [comment]
@jonbob That mesh is in the list of meshes that @xylar developed the new ICs for, according to https://github.com/E3SM-Project/E3SM/pull/5464#issuecomment-1440497659. @xylar do you know what might be going wrong here?

**Comment by jonbob** [comment]
Yes, my test was with one of the meshes with the new ICs. The failures are on the meshes without the updated ICs

**Comment by cbegeman** [comment]
@jonbob I see. In our discussion in https://github.com/E3SM-Ocean-Discussion/E3SM/pull/38, the decision was that the runs should fail whenever `landIceFloatingMask` was not defined rather than having the default behavior be that `landIceFloatingMask = landIceMask`. So, @xylar, this means that all meshes must have updated ICs unless we want to change this behavior (i.e., warning instead of critical error). 

I think the other option is that we make sure that the variables in `landIcePressurePackage` are not associated when we run without ISC. If that were the case, then we could only run this check when those variables are associated. However, that also requires a change in the code and I think in the ICs as well which I believe include `landIceMask` and other `landIce*` variables.

**Comment by xylar** [comment]
@cbegeman, I think maybe a package check is missing around the error message. I would have thought these variables weren't even allocated in runs without ice-shelf cavities. But one way or another we need to skip the check when they're aren't cavities.

**Comment by cbegeman** [comment]
@xylar That's what I was getting at by my "option 2". At least in compass, the init streams do include land ice variables. 

A third option is setting `config_global_ocean_depress_by_land_ice = .false.`. We wouldn't be encountering this error if `ocn_init_setup_global_ocean_interpolate_land_ice_topography` wasn't called. I think this is the least obtrusive change.

**Comment by xylar** [comment]
@cbegeman, I don't think @jonbob's errors are related to init mode where the namelist opinion and variables in your 3rd option apply. I'm on my phone so I'm a little limit right now.

**Comment by cbegeman** [comment]
@xylar But if that error arises in diagnostics, then `landIcePressureOn = .true.`, which means that `config_land_ice_flux_mode != 'off'`. So don't we still need to address this at the config level?

**Comment by xylar** [comment]
I agree. It's just that config_global_ocean_depress_by_land_ice is an init mode config option.

**Comment by xylar** [comment]
@cbegeman, init mode doesn't run in E3SM.

I would do something like:
https://github.com/E3SM-Project/E3SM/blob/8018a7ad5cb60ca77e6377e0d2566095a49da405/components/mpas-ocean/src/analysis_members/mpas_ocn_global_stats.F#L281
but with `landIcePressurePKGActive`.  Then, wrap the check and error message in an:
```
 if (landIcePressurePKGActive) then
...
end if
```

**Comment by cbegeman** [comment]
I understand that you can check whether the package is active, but this check should be superfluous if we're already checking `config_land_ice_flux_mode` (which is assigned to `landIcePressureOn`). See these lines:

https://github.com/E3SM-Project/E3SM/blob/8018a7ad5cb60ca77e6377e0d2566095a49da405/components/mpas-ocean/src/driver/mpas_ocn_core_interface.F#L294-L303

**Comment by jonbob** [comment]
@cbegeman -- no, I see this:
````
 config_land_ice_flux_mode = 'pressure_only'
 ````

**Comment by cbegeman** [comment]
@jonbob Thanks for checking. I believe that if you set that to 'off' there will be no errors. Do you know where that is being set?

**Comment by jonbob** [comment]
The bld files were never updated to match changes to Registry

**Comment by cbegeman** [comment]
@jonbob I apologize. I didn't realize that land ice pressure was on for non-ISC cases. I can make those changes and push them right now.

**Comment by jonbob** [comment]
Setting Registry defaults has no impact in E3SM unless we change the bld files to match

**Comment by cbegeman** [comment]
@jonbob Running the script might be a good idea. The change I think we have to make is

```
diff --git a/components/mpas-ocean/bld/build-namelist b/components/mpas-ocean/bld/build-namelist
index b1dd1498a7..5dddbcfffd 100755
--- a/components/mpas-ocean/bld/build-namelist
+++ b/components/mpas-ocean/bld/build-namelist
@@ -781,7 +781,7 @@ if ($OCN_ISMF eq 'coupled') {
 } elsif ($OCN_ISMF eq 'internal') {
        add_default($nl, 'config_land_ice_flux_mode', 'val'=>"standalone");
 } else {
-       add_default($nl, 'config_land_ice_flux_mode', 'val'=>"pressure_only");
+       add_default($nl, 'config_land_ice_flux_mode', 'val'=>"off");
 }
 add_default($nl, 'config_land_ice_flux_formulation');
 add_default($nl, 'config_land_ice_flux_useHollandJenkinsAdvDiff');

jonbob:
OK, let me work on that

xylar:
@cbegeman, I don't think those changes are correct. We want a mode where we run without ISMF but still have land ice pressure. It seems like that's what the original code is for.

cbegeman:
@xylar Would you like to add a separate flag or just a separate string option? Clearly, 'coupled' and 'internal' don't adequately describe the range of options

```
     -ocn_ismf             variable for defining how the ocn model will handle ice shelf melt

**Comment by cbegeman** [comment]
@xylar Sure, I'll get back in touch.

**Comment by xylar** [comment]
@cbegeman, yes, it took me a little time to see what had been done.  In:
https://github.com/E3SM-Project/E3SM/pull/2602/files#diff-6f810f592637739301f930dc5a1822a43bbcaf800b82e33a670c5c4a76346268
it appears that `config_land_ice_flux_mode` got changed from being `off` by default to being `pressure_only` by default, but that setting is only appropriate for `wISC` meshes.  The default needs to be `off` for non `wISC` and `pressure_only` for `wISC`.

I don't know how to accomplish that in practice in `build-namelist`.  If we can figure it out, I'm still not sure if that solves the problem but I think so.

**Comment by xylar** [comment]
I think we need a 4th value for `MPASO_ISMF here:
https://github.com/E3SM-Project/E3SM/blob/8018a7ad5cb60ca77e6377e0d2566095a49da405/components/mpas-ocean/cime_config/config_component.xml#L42-L54
that is `pressure_only` if `wISC` is found in the mesh name:
```
           <value compset="MPASO_">none</value>
           <value compset="_MPASO%.*ISMF.*_SGLC">internal</value>
           <value compset="_MPASO%.*ISMF.*_MALI">coupled</value>
```
but it's not actually clear that can happen here.  @jonbob, how do we get the 4 modes we need rather than the 3 we have?:
* config_land_ice_flux_mode = off if `wISC` not in mesh name
* config_land_ice_flux_mode = pressure_only if `wISC` in the mesh name and `MPASO_ISMF` is none
* config_land_ice_flux_mode = standalone if `wISC` in the mesh name and `MPASO_ISMF` is internal
* config_land_ice_flux_mode = coupled if `wISC` in the mesh name and `MPASO_ISMF` is coupled

The problem currently is that `pressure_only` is being used for cases without ice-shelf cavities in E3SM, which is not correct.

**Comment by jonbob** [comment]
@xylar -- what about something like this:

- if MPASO_ISMF is internal, then config_land_ice_flux_mode = standalone
- if MPASO_ISMF is coupled, then config_land_ice_flux_mode = coupled
- if MPASO_ISMF is none, then let config_land_ice_flux_mode get set by its defaults in namelist_defaults_mpaso.xml, where it can be mesh-dependent?

That means the chunk of code in build-namelist would look like:
````
if ($OCN_ISMF eq 'coupled') {
        add_default($nl, 'config_land_ice_flux_mode', 'val'=>"coupled");
} elsif ($OCN_ISMF eq 'internal') {
        add_default($nl, 'config_land_ice_flux_mode', 'val'=>"standalone");
} else {
        add_default($nl, 'config_land_ice_flux_mode');
}
````

**Comment by xylar** [comment]
@jonbob, yes that sounds right.

**Comment by xylar** [comment]
Following up, the following code:
https://github.com/E3SM-Project/E3SM/blob/8018a7ad5cb60ca77e6377e0d2566095a49da405/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml#L336-L341
Would need to be:
```
<config_land_ice_flux_mode>'off'</config_land_ice_flux_mode>
<config_land_ice_flux_mode ocn_grid="oQU240wLI">'pressure_only'</config_land_ice_flux_mode>
<config_land_ice_flux_mode ocn_grid="oEC60to30v3wLI">'pressure_only'</config_land_ice_flux_mode>
<config_land_ice_flux_mode ocn_grid="ECwISC30to60E1r2">'pressure_only'</config_land_ice_flux_mode>
<config_land_ice_flux_mode ocn_grid="oRRS30to10v3wLI">'pressure_only'</config_land_ice_flux_mode>
<config_land_ice_flux_mode ocn_grid="SOwISC12to60E2r4">'pressure_only'</config_land_ice_flux_mode>
<config_land_ice_flux_mode ocn_grid="ECwISC30to60E2r1">'pressure_only'</config_land_ice_flux_mode>
```

**Comment by xylar** [comment]
@jonbob, my sense is that there aren't new namelist options to propagate to the defaults but the new variables do need to be propagated.

Even so, I'm going to make the changes above first and then hopefully you can run the automated scripts and we can clean up the conflicts if there are any.

**Comment by xylar** [comment]
@cbegeman, I hope you don't mind that I pushed these changes.  I'm hoping @jonbob might be able to re-test while you're away next week.  But we won't merge anything until you've had a look.

**Comment by xylar** [comment]
BTW, I think these changes are good anyway because the old version where defaults get set in `namelist_defaults_mpaso.xml` but then completely overwritten in `build-namelist` was something I had completely missed.

**Comment by cbegeman** [comment]
@xylar and @jonbob I'm happy with these changes. Thanks for pushing this forward!

**Comment by jonbob** [comment]
I tested this PR today and we still have issues, now in the ocean mct driver. There's a chunk of code that looks like:
````
    call mpas_pool_get_config(domain % configs, 'config_land_ice_flux_mode', config_land_ice_flux_mode)
    if ( trim(config_land_ice_flux_mode) .eq. 'pressure_only' ) then
       call seq_infodata_PutData( infodata, ocn_prognostic=.true., ocnrof_prognostic=.true., &
                                            ocn_c2_glcshelf=.false.)
    else if ( trim(config_land_ice_flux_mode) .eq. 'standalone' ) then
       call seq_infodata_PutData( infodata, ocn_prognostic=.true., ocnrof_prognostic=.true., &
                                            ocn_c2_glcshelf=.false.)
    else if ( trim(config_land_ice_flux_mode) .eq. 'coupled' ) then
       call seq_infodata_PutData( infodata, ocn_prognostic=.true., ocnrof_prognostic=.true., &
                                            ocn_c2_glcshelf=.true.)
    else
       call mpas_log_write('ERROR: unknown land_ice_flux_mode: ' // trim(config_land_ice_flux_mode), MPAS_LOG_CRIT)
    end if
````
It's missing the new "off" setting, so causing an error. I assume we want "off" to make the same settings as "pressure_only"? There's also another section that has:
````
       if ( trim(config_land_ice_flux_mode) .ne. 'pressure_only' ) then
          o2x_o % rAttr(index_o2x_So_blt, n) = landIceBoundaryLayerTracers(indexBLT,i)
          o2x_o % rAttr(index_o2x_So_bls, n) = landIceBoundaryLayerTracers(indexBLS,i)
          o2x_o % rAttr(index_o2x_So_htv, n) = landIceTracerTransferVelocities(indexHeatTrans,i)
          o2x_o % rAttr(index_o2x_So_stv, n) = landIceTracerTransferVelocities(indexSaltTrans,i)
          o2x_o % rAttr(index_o2x_So_rhoeff, n) = 0.0_RKIND
       endif
````
I would also assume this needs to include "off" as well, though it might be clearer if we check for eq to the other two settings, i.e.
````
       if ( trim(config_land_ice_flux_mode).eq.'standalone' .or. trim(config_land_ice_flux_mode).eq.'coupled'  ) then
````

**Comment by cbegeman** [comment]
@jonbob I think that's exactly what we need. I believe I have implemented your suggestions in https://github.com/E3SM-Project/E3SM/pull/5464/commits/01e97b8bbfcb817599d6dc1a4d1794dc6f67463e, but have not done any testing.

File: components/mpas-ocean/driver/ocn_comp_mct.F [hunk 2]
```diff
@@ -790,7 +790,10 @@ end subroutine xml_stream_get_attributes
...
 
     call mpas_pool_get_config(domain % configs, 'config_land_ice_flux_mode', config_land_ice_flux_mode)
-    if ( trim(config_land_ice_flux_mode) .eq. 'pressure_only' ) then
+    if ( trim(config_land_ice_flux_mode) .eq. 'off' ) then
+       call seq_infodata_PutData( infodata, ocn_prognostic=.true., ocnrof_prognostic=.true., &
+                                            ocn_c2_glcshelf=.false.)
+    elseif ( trim(config_land_ice_flux_mode) .eq. 'pressure_only' ) then
        call seq_infodata_PutData( infodata, ocn_prognostic=.true., ocnrof_prognostic=.true., &
                                             ocn_c2_glcshelf=.false.)
     else if ( trim(config_land_ice_flux_mode) .eq. 'standalone' ) then
```
xylar, line 799: Unless I'm missing something, these 3 entries re identical.

```suggestion
    if ( trim(config_land_ice_flux_mode) .eq. 'off' .or. &
            trim(config_land_ice_flux_mode) .eq. 'pressure_only'.or. &
            trim(config_land_ice_flux_mode) .eq. 'standalone') then
```

xylar: Alternatively, I can make this change in https://github.com/E3SM-Ocean-Discussion/E3SM/pull/45.

xylar:
@cbegeman, the conflict is related to my changes in #5553.  You'll presumably need to rebase to fix the conflicts.  It looks like it's just going to be a question of indenting the changes you've made so they're inside the new `if` statement.

components/mpas-ocean/driver/ocn_comp_mct.F
cbegeman: Fixed in https://github.com/E3SM-Project/E3SM/pull/5464/commits/aa144a63b3e2d2efbaba6f298ef5696df51397f4

cbegeman:
@jonbob, @xylar Now rebased and ready for further testing

jonbob:
passes:

with expected NML DIFFs and BFB for grids without ice shelf cavities

merged to next

jonbob:

Issue #5259, opened 2022-10-28
COMPARE fail with ERP_Ln18.ne4_oQU240.F2010 using nvidia compiler on pm-cpu (only OPT and with threads)


ndkeen:
This test `ERP_Ln18.ne4_oQU240.F2010.pm-cpu_nvidia` fails COMPARE.  

The DEBUG version pass as well as single-threaded OPT build, ie:
```
ERP_D_Ln18.ne4_oQU240.F2010
ERP_PMx1_Ln18.ne4_oQU240.F2010 
```

ndkeen:
I can narrow this down a little further to say that it will also pass if I use 2 threads in all components, except use 1 in ATM.

Issue #5256, opened 2022-10-28
free(): invalid pointer with nvidia compiler on pm-cpu with OPT build and threading


ndkeen:
This may be a generic error and not point to the root isssue, but the following 3 tests fail with nvidia compiler (latest version `nvidia/22.5`) on pm-cpu in OPT builds, but all pass in DEBUG builds.

Fail:
```
SMS_P12x2.ne4_oQU240.WCYCL1850NS.pm-cpu_nvidia.allactive-mach_mods
SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.pm-cpu_nvidia
ERS_Ld5.T62_oQU120.CMPASO-NYF.pm-cpu_nvidia
```

Pass:
```
SMS_D_P12x2.ne4_oQU240.WCYCL1850NS.pm-cpu_nvidia.allactive-mach_mods
SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.pm-cpu_nvidia
ERS_D_Ld5.T62_oQU120.CMPASO-NYF.pm-cpu_nvidia
```

And this test fails with same type of error message in BOTH debug and opt:
`SMS_D.r05_r05.IELM.pm-cpu_nvidia.elm-topounit`
`SMS.r05_r05.IELM.pm-cpu_nvidia.elm-topounit`

Example of actual error message:
```
30: free(): invalid pointer
34: free(): invalid pointer
 4: free(): invalid pointer
 7: munmap_chunk(): invalid pointer
srun: error: nid004478: tasks 4,7,30,34: Aborted
srun: launch/slurm: _step_signal: Terminating StepId=3511843.0
srun: error: nid004478: tasks 11-12,18,58: Segmentation fault
 0: slurmstepd: error: *** STEP 3511843.0 ON nid004478 CANCELLED AT 2022-10-28T15:27:31 ***
srun: error: nid004478: tasks 0-3,5-6,8-10,13-17,19-29,31-33,35-57,59-63: Terminated
srun: Force Terminated StepId=3511843.0
```

ndkeen:
I'm trying these tests with only 1 thread.

At least one of them does pass: 
```
SMS_P12x1.ne4_oQU240.WCYCL1850NS.pm-cpu_nvidia.allactive-mach_mods
```

ndkeen:
I tried the failing cases again with Nov 30th master that now uses updated version of the nvidia compiler (22.5), but I don't see different behavior.

[...]

PR #5468, opened 2023-02-21
For pm-cpu (and alvarez), decrease optimization of nvidia compiler fortran builds 

ndkeen:
For pm-cpu and similar alvarez test machine, decreasing the optimization level of nvidia compiler fortran builds allows several tests to pass (from `-O2` to `-O1`). The nvidia compiler seems to use the highest optimization level when 2 flags are passed, ie `-O2 -O1` will still act as `-O2`, which is why the base opt level is decreased here from `-O2` to `-O1`. We can then increase the optimization in the machine-specific Depends files.

Fixes https://github.com/E3SM-Project/E3SM/issues/5256 (except for the elm-topunit test, which I will make a separate issue)
Fixes https://github.com/E3SM-Project/E3SM/issues/5259


cime_config/machines/cmake_macros/nvidia_pm-cpu.cmake
```diff
@@ -17,7 +17,7 @@ if (NOT DEBUG)
   string(APPEND CXXFLAGS " -O2")
 endif()
 if (NOT DEBUG)
-  string(APPEND FFLAGS " -O2")
+  string(APPEND FFLAGS " -g")
```
rljacob, line 20: Adding -g for NOT DEBUG ?

ndkeen: Yea, I didn't mention that in the comments -- I like `-g` for builds as it can sometimes give more info about a crash. It will increase exe size, but I don't see major performance impacts -- esp for situations like this where we are only using nvidia compiler on this machine as a test. But I can remove it and only add it when I'm testing.

rljacob: I was just wondering if its a mistake.  Its fine to leave it.

rljacob (approved):

ndkeen:
merged to next

PR #5470, opened 2023-02-22
Add WSL2 machine and gnu compiler

thorntonpe:
Adds a new machine for Windows Subsystem for Linux v2, and adds the gnu compiler for that machine via cmake_macros 
[BFB]


rljacob:
Is this something any windows user could use?

thorntonpe:
Yes, although it requires that you enable WSL2 (easy step in windows), and it also requires that you install the third-party libraries in whatever linux distro you are using within WSL2. I'm using ubuntu and the library installs had no issues - it is the same level of effort needed for a generic linux box. I have the paths in config_machines.xml hardwired for my own directory structure, and another user would want to change those. The same would be true no matter what paths I picked, so I just left them as the ones that I'm working with.

From: Robert Jacob ***@***.***>
Sent: Wednesday, February 22, 2023 1:06 PM
Cc: Thornton, Peter E. ***@***.***>; Author ***@***.***>
Subject: [EXTERNAL] Re: [E3SM-Project/E3SM] Add WSL2 machine and gnu compiler (PR #5470)


Is this something any windows user could use?

-

rljacob (approved):

rljacob:
merged to next

Issue #5473, opened 2023-02-24
Sea ice weak formulation no longer works


akturner:
After the recent sea ice velocity solver initialization the weak formulation test cases no longer work

PR #5474, opened 2023-02-24
Fixes issues with weak sea ice spatial operators formulation

akturner:
Recent refactoring to sea ice velocity initialization broke some test cases using the weak formulation. This allows the weak formulation to run successfully.

Fixes https://github.com/E3SM-Project/E3SM/issues/5473

[BFB]

### DIFF
--- a/components/mpas-seaice/src/shared/mpas_seaice_velocity_solver_variational_shared.F
+++ b/components/mpas-seaice/src/shared/mpas_seaice_velocity_solver_variational_shared.F
@@ -123,7 +123,10 @@ contains
        blockPtr => domain % blocklist
        do while (associated(blockPtr))
 
+          call MPAS_pool_get_subpool(blockPtr % structs, "mesh", meshPool)
           call MPAS_pool_get_subpool(blockPtr % structs, "velocity_variational", velocityVariationalPool)
+
+          call MPAS_pool_get_dimension(meshPool, "nVertices", nVertices)
 
           call MPAS_pool_get_array(velocityVariationalPool, "tanLatVertexRotatedOverRadius", tanLatVertexRotatedOverRadius)
 
--- a/components/mpas-seaice/src/shared/mpas_seaice_velocity_solver_weak.F
+++ b/components/mpas-seaice/src/shared/mpas_seaice_velocity_solver_weak.F
@@ -80,6 +80,10 @@ contains
 
        call MPAS_pool_get_config(domain % configs, "config_rotate_cartesian_grid", config_rotate_cartesian_grid)
 
+       call MPAS_pool_get_subpool(blockPtr % structs, "velocity_weak", velocityWeakPool)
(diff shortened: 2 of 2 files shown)


eclare108213:
@akturner, thanks for you patience with me coming up to speed on PR processes.  Not knowing what is required and not fully understanding this PR/issue, I've been holding off on this one.  So let's move it forward.

My guess is that this is just cleaning up a variable name that was changed in the previous PR but not tested with the weak formulation turned on at that time.  Is that correct?  If so, is the weak formulation not regularly tested in E3SM?  As we discussed in the meeting today, I'd like to enhance the testing process for Ice Group PRs to avoid these kinds of problems, but that's not something this particular PR will have to satisfy.  

Nevertheless, more information in this PR submission and/or the associated issue would help me understand what happened, when, why there's a problem, and what your change does to fix it.  E.g.
1. Which PR/hash introduced the bug?
2. What tests did you run in support of this PR (#5474), that were BFB?  (configurations, lengths, etc). If this has only been tested in a standalone configuration, please also test it in E3SM, or argue why the code should be expected to be BFB there without testing.
3. When the weak formulation is turned on in E3SM, would the code not run at all?  Or only for certain cases?
4. Are simulations with the weak formulation turned on BFB compared with the code prior to when the bug was introduced?  That might be difficult to ascertain, if other changes in E3SM have been non-BFB in the meantime, but it would be useful to understand what is actually changing here.  Having a link to the PR that introduced the bug could be helpful too.  Thanks.

darincomeau:
This PR came up in today's MPAS-devops call as an edge case where affected code is intended for MPAS-Seaice standalone, but could still be turned on in E3SM through namelist options. Speaking for myself, while this would technically be considered stealth by touching (long-lived) code that is turned off by default in E3SM, it doesn't warrant compliance with the standard stealth feature code review process (lengthy simulations and adding the feature to the E3SM test suite), as it is not intended for use in E3SM.
There's the larger issue of how such namelist options that are intended for standalone features which can be turned on in E3SM, but are not intended to, should be handled, e.g. by removing the features from the E3SM build system, but that's for a larger discussion.

darincomeau:
Confirmed without this fix, and with the above 'weak' namelist options set, SMS.TL319_EC30to60E2r2.GMPAS-JRA1p5.anvil_intel dies during initialization of ice velocity solver.

eclare108213 (approved):
Thank you, @darincomeau, for running the tests and answering my questions to the extent possible.  I will approve this PR based on those test results, and because this change allows E3SM to be run with the weak formulation without crashing. I don't think it should be considered only for standalone runs - if there is a physical reason that it really shouldn't be used in coupled runs (beyond not being tested/supported), then I would like to understand why - a topic for a separate conversation.  I also agree that broader issues surrounding "stealth features" needs to be discussed in the larger E3SM context, outside of this PR.  @akturner, thanks for fixing this problem.

jonbob:
merged to master

PR #5475, opened 2023-02-24
New ocean drag feature: use harmonic mean edge layer thickness

cbegeman:
Here we add a new drag feature to MPAS-Ocean. The namelist option `config_thickness_drag_type` indicates what method is to be used for computing the layer thickness at edges used in the implicit drag term. The default option, `'centered'`, is bit-for-bit and retains the arithmetic mean computation. The new option, `'harmonic-mean'`, uses the harmonic mean of the neighboring cell layer thickness, thus increasing the drag anywhere there are horizontal gradients in layer thickness. When used, this change improves the performance of MPAS-Ocean in wetting-and-drying simulations relative to the analytical solution in the test case from Warner et al. (2013).

Further discussion is located at https://github.com/E3SM-Ocean-Discussion/E3SM/pull/36.

[BFB][stealth]

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
@@ -226,7 +226,7 @@ contains
 #endif
 
       ! inputs: layerThickness, normalVelocity
-      ! output: layerThickEdgeMean, layerThickEdgeFlux
+      ! output: layerThickEdgeMean, layerThickEdgeDrag, layerThickEdgeFlux
       call ocn_diagnostic_solve_layerThicknessEdge(normalVelocity, &
                                                    layerThickness)
 
@@ -469,7 +469,8 @@ contains
 
 #ifdef MPAS_OPENACC
       !! Listing outputs by first routine that modifies variable
-      !! ocn_diagnostic_solve_layerThicknessEdge :: layerThickEdgeFlux, (diagnostics array)
+      !! ocn_diagnostic_solve_layerThicknessEdge :: layerThickEdgeDrag, (diagnostics array)
+      !!                                            layerThickEdgeFlux, (diagnostics array)
       !!                                            layerThickEdgeMean (diagnostics array)
       !! ocn_relativeVorticity_circulation       :: relativeVorticity, (diagnostics array)
       !!                                            circulation (diagnostics array)
@@ -507,7 +508,7 @@ contains
       !! ocn_diagnostic_solve_ssh                :: pressureAdjustedSSH, (diagnostics array)
       !!                                            gradSSH (diagnostics array)
 
-!     !$acc update host(layerThickEdgeFlux, layerThickEdgeCenter)
+!     !$acc update host(layerThickEdgeDrag, layerThickEdgeFlux, layerThickEdgeCenter)
 !     !$acc update host(relativeVorticity, circulation)
 !     !$acc update host(vertTransportVelocityTop, &
 !     !$acc             vertGMBolusVelocityTop, &
--- a/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
@@ -508,7 +508,9 @@ contains
       !! ocn_diagnostic_solve_ssh                :: pressureAdjustedSSH, (diagnostics array)
       !!                                            gradSSH (diagnostics array)
 
-!     !$acc update host(layerThickEdgeDrag, layerThickEdgeFlux, layerThickEdgeCenter)
+!     !$acc update host(layerThickEdgeDrag, &
+!     !$acc             layerThickEdgeFlux, &
+!     !$acc             layerThickEdgeMean)
 !     !$acc update host(relativeVorticity, circulation)
 !     !$acc update host(vertTransportVelocityTop, &
 !     !$acc             vertGMBolusVelocityTop, &
@@ -660,10 +662,10 @@ contains
 !> \author  Matt Turner
 !> \date    October 2020
 !> \details
-!>  This routine computes the diagnostic variables layerThickEdgeFlux
-!>  and layerThickEdgeMean. The Mean is a simple mean across the edge
-!>  but the Flux value can either be the mean or an upwind value,
-!>  depending on user input.
+!>  This routine computes the diagnostic variables layerThickEdgeDrag,
+!>  layerThickEdgeFlux, and layerThickEdgeMean. The Mean is a simple
+!>  mean across the edge but the Flux value can either be the mean or
+!>  an upwind value, depending on user input.
 !
 !-----------------------------------------------------------------------
 
@@ -684,6 +686,7 @@ contains
 
       ! Outputs are the shared variables from diagnostic_variables:
       !real (kind=RKIND), dimension(:,:), intent(out) :: &
+      !   layerThickEdgeDrag, & !< [out] layer thickness on edge for drag
       !   layerThickEdgeMean, & !< [out] centered layer thickness on edge
       !   layerThickEdgeFlux    !< [out] flux-related thickness on edge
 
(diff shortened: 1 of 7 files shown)


cbegeman:
@xylar When you have a chance, the sidebar needs to be populated. Thanks!

xylar:
@cbegeman, am I right that this should be labeled "stealth"?

cbegeman:
@xylar It is stealth and BFB. Thanks!

xylar:
@cbegeman, thanks.  I think in terms of the labels, it has to be one or the other:
https://acme-climate.atlassian.net/wiki/spaces/CNCL/pages/3313238063/Code+Review+and+New+Feature+Process
I think a stealth feature also needs something like a 5-year simulation to make sure the feature works as expected in climate runs if it's turned on.  We can check in with @jonbob about that process.

xylar:
Hmm, looking at recent PRs, almost all `stealth` PRs also use the `BFB` label so others seem to have a different understanding of that than I do.  I added BFB as well.

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
```diff
@@ -78,11 +78,17 @@ module ocn_diagnostics
...
-   integer :: thickEdgeFluxChoice  ! choice of thickness flux type
+   integer :: &
+      thickEdgeFluxChoice, &! choice of thickness flux type
+      thickEdgeDragChoice   ! choice of thickness drag type
    integer, parameter :: &
       thickEdgeFluxUnknown = 0, &! type unknown or unset
       thickEdgeFluxCenter  = 1, &! use mean thickness of cell neighbors
       thickEdgeFluxUpwind  = 2   ! use upwind cell thickness at edge
+   integer, parameter :: &
+      thickEdgeDragUnknown = 0, &! type unknown or unset
```
xylar, line 89: I'm a bit confused about why the `*Unknown` parameters are needed.  Don't we just raise a critical error and stop the simulation right after they get set?  So maybe just raise the critical error and exit without bothering to set a value?

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
```diff
@@ -4229,6 +4302,23 @@ subroutine ocn_diagnostics_init(domain, err)!{{{
...
+      elseif (trim(config_thickness_drag_type) == 'harmonic-mean') then
+         thickEdgeDragChoice = thickEdgeDragHarMean
+         call mpas_log_write('Thickness drag option of ' //&
+              & trim(config_thickness_drag_type) // &
+              & 'is assigned')
+      else
+         thickEdgeFluxChoice = thickEdgeFluxUnknown
+         call mpas_log_write('Thickness drag option of ' //&
+              & trim(config_thickness_drag_type) // &
+              & 'is not known', MPAS_LOG_CRIT)
```
xylar, line 4319:
```suggestion
         call mpas_log_write('Thickness drag option of ' //&
              & trim(config_thickness_drag_type) // &
              & 'is not known', MPAS_LOG_CRIT)
```

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
xylar:
```suggestion
```

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
```diff
@@ -78,11 +78,17 @@ module ocn_diagnostics
    real (kind=RKIND), pointer ::  coef_3rd_order
 
    ! Methods for computing thickness at edges for flux calculations
-   integer :: thickEdgeFluxChoice  ! choice of thickness flux type
+   integer :: &
+      thickEdgeFluxChoice, &! choice of thickness flux type
+      thickEdgeDragChoice   ! choice of thickness drag type
    integer, parameter :: &
       thickEdgeFluxUnknown = 0, &! type unknown or unset
```
xylar, line 85:
```suggestion
```

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
xylar: Also, this refers to "Flux" but should be "Drag" if it is kept.

xylar: I would remove `thickEdgeFluxChoice = thickEdgeFluxUnknown` in the set of ifs above, too.

components/mpas-ocean/src/shared/mpas_ocn_vmix.F
```diff
@@ -844,6 +849,16 @@ subroutine ocn_vel_vmix_tend_implicit_spatially_variable_mannings(forcingPool, b
          cell2 = cellsOnEdge(2,iEdge)
 
          ! average cell-based implicit bottom drag to edges and convert Mannings n to Cd
+         if (trim(config_thickness_drag_type) == 'harmonic-mean') then
```
xylar, line 852: Is there a way to use `thickEdgeDragChoice` here?  If so, that would be preferable.  If there is an issue with circular dependencies, maybe @philipwjones has a suggestion.

xylar (changes requested):
@cbegeman, sorry, here are a few (fairly minor) things that I noticed here that I seem to have missed in my review of the discussion PR.  Let me know what you think.

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
cbegeman: @xylar I think that's a good suggestion. I've removed them with https://github.com/E3SM-Project/E3SM/pull/5475/commits/00847366fe5b7a7afdc1ab8bad59cf9f21385d85

cbegeman:
@xylar After adding https://github.com/E3SM-Project/E3SM/pull/5475/commits/00847366fe5b7a7afdc1ab8bad59cf9f21385d85, I reran the `drying_slope/sigma/1km/default` test case with both drag options and an incorrect option.

dengwirda (commented):
Thanks @cbegeman, I ran the tides icos7 case successfully on cori using the new `config_thickness_drag_type = 'harmonic-mean'`, and saw a small improvement in accuracy (10.7cm vs 10.8cm RMSE).
In terms of single-layer wetting+drying this all looks good to me. Are additional E3SM tests required though?

cbegeman:
@dengwirda Thanks for the update! I had not run the tides cases so that's great. 

Yes, in terms of E3SM testing, we will need the 10-year comparison for the stealth feature. I think once you and @xylar approve we will pass this over to @jonbob for that testing.

xylar (approved):
Looks good to me, pending the results of the 10-year stealth feature run.

dengwirda (approved):

rljacob:
Status:  Still need runs to be done.

jonbob:
@cbegeman -- I ran the scripts to make the bld files consistent with the Registry changes an dpushed them back. I'd like to start testing this soon, but there are some conflicts, especially in mpas_ocn_vmix.F, that are not completely clear to me how to resolve

cbegeman:
@jonbob No problem. I've resolved the merge conflicts so it should be ready to go.

xylar:
Thanks @jonbob!  It's like getting a surprise package each time you post one of these analysis pages.  What are we going to get?!

There is only one thing I see that looks like a bigger change than ensemble-level variability.  That is the warmer seafloor temperatures around the Filchner-Ronne Ice Shelf.  That wasn't a feature I would have expected from this change, and it's one we might have look into further if we want to switch to harmonic mean for production simulations in general.  But it's a subtle and localized feature so I'm comfortable approving the change.

jonbob:
Thanks @xylar -- it is always a mystery! But I suppose that can simply be the nature of even just individual ensemble members, much less changes that have small subtle causes

cbegeman:
Thanks @jonbob! I agree with @xylar that it's interesting but not concerning.

dengwirda:
Interesting to see the sensitivity to small changes in bottom friction dissipation, but I agree there's nothing that needs chasing down re: this PR.

cbegeman:
@jonbob I just resolved the merge conflicts so this is ready for any final testing.

jonbob:
Added a new stealth test to exercise this feature, checked to make sure it works. @cbegeman, is SMS_D_Ld1.T62_oQU240.GMPAS-IAF.mpaso-harmonic_mean_drag a reasonable test?

cbegeman:
@jonbob Yes, that would be sufficient. Thanks!

jonbob:
passes:

jonbob:

Issue #5476, opened 2023-02-25
Defunct sea ice standalone testing infrastructure needs removed


akturner:
Now that the sea ice standalone infrastructure is in a separate repo, this needs to be removed from the E3SM repo

Issue #5440, opened 2023-02-07
Slightly re-arranging directory structure for mpas-seaice testing


proteanplanet:
To accommodate new E3SM testing infrastructure coming to the sea ice model, I would like to suggest rearranging the directory tree under https://github.com/E3SM-Project/E3SM/tree/master/components/mpas-seaice, whereby the current directory,  "testing_and_setup", and all of its contents, which I believe is for running mpas-seaice outside of the E3SM framework, is renamed "mpas", and is shifted down one notch to under a new directory named "test", thus copying the EAM hierarchy.  We would then add a second directory under "test" named "e3sm" which will include infrastructure for single column testing and new scripts within E3SM, such as https://github.com/E3SM-Project/SimulationScripts/blob/master/archive/PPSLRCI/E3SM_SeaIce_Developer.sh, thus ending up with a new directory tree for mpas-seaice:

E3SM/components/mpas-seaice

<img width="539" alt="Screen Shot 2023-02-07 at 1 18 19 PM" src="https://user-images.githubusercontent.com/14379090/217355739-afbcb2df-a562-4f48-a068-72f8d9f5b099.png">

Please comment if you vehemently disagree, and also mention if there are any relative directory locations or names scripted under the existing "testing_and_setup" that would be affected by this change.

njeffery:
I like it.

erinethomas:
this works for me.

eclare108213:
I like it too.

darincomeau:
looks good to me.

jonbob:
works for me

proteanplanet:
After consultation with @akturner and @njeffery, @jonbob and I will change this plan so that stand-alone test code is not being kept in two places at once (MPAS-Dev and E3SM-Project).  The new plan is to create a new test directory, with the new E3SM-Polar-Developer.sh script under an E3SM subdirectory, and a submodule for the MPAS-Dev standalone testing mainly developed by @akturner situated under an adjacent MPAS subdirectory.  At @jonbob suggestion, we will first submit a PR to add this new test directory and content, and then I will approve deletion of the aging stand-alone testing code in https://github.com/E3SM-Project/E3SM/pull/5477.  It will be my responsibility to keep the hash up to date, with prompting from @akturner

proteanplanet:
This issue is being closed, now with the MPAS testing infrastructure removed, and just one script currently in PR for testing of the sea ice model in the E3SM framework.

PR #5477, opened 2023-02-25
Removed defunct standalone testing infrastructure

akturner:
Standalone sea ice infrastructure is now in a seperate repo

Fixes https://github.com/E3SM-Project/E3SM/issues/5476

[BFB]


eclare108213 (commented):
It looks like the only changes here are file deletions -- a lot of them -- which should NOT affect E3SM in other configurations.  I'm asking @proteanplanet to take a look too, in case I'm missing something.

proteanplanet:
@akturner Why not follow the suggested approach at https://github.com/E3SM-Project/E3SM/issues/5440?  The stand-alone infrastructure you have built is valuable, and my understanding is that the E3SM project repo is now where MPAS-SeaIce lives. It seems that a lot of valuable work is being deleted, when it should still reside with the sea ice model.

akturner:
@proteanplanet: Everything is available as https://github.com/MPAS-Dev/MPAS-Seaice_standalone_framework. As explained previously it was too difficult to get E3SM reviews for frequent changes to this framework.

eclare108213:
Then the solution OUGHT to be to fix the review process.  Is there a Confluence page somewhere that describes how this is supposed to work?  I've noticed several things about it that seem odd to me, maybe just because I'm unfamiliar.  If it's not working, I do think it's worth trying to adjust.  It's the reason why MPAS was pulled entirely into E3SM, right?  So it's possible to improve such things.

proteanplanet:
This PR seems to open three issues that need to be resolved:
1) My understanding is that the stand-alone testing infrastructure is out of date in the E3SM repo, and therefore needs to be brought up to date from MPAS-Dev rather than just being deleted.
2) It's important the stand-alone testing infrastructure does not disappear from E3SM since it belongs to the project.  We are developing a parallel testing platform that uses E3SM infrastructure, but keeping the stand-alone work as a comparison remains important; the new testing development that uses E3SM infrastructure will span phase 3. 
3) If the review of these development tools is problematic in E3SM, that needs to be addressed rather than abandoned.

proteanplanet (commented):
Will approve this request imminently following the plan of action at https://github.com/E3SM-Project/E3SM/issues/5440#issuecomment-1470972018.

proteanplanet (approved):
Since this testing infrastructure is out of date and not in sync with MPAS-Dev, I approve this PR.  In future, we need a way to make sure that the one place where stand-alone MPAS-SI testing is housed remains accessible and usable within E3SM-Project. Inside Polaris may make sense.

eclare108213 (approved):
I agree with @proteanplanet. The entire testing infrastructure for MPAS-SI needs to be reviewed, reorganized, and made more user-friendly for E3SM development purposes.  In the meantime, it makes no sense to try to maintain duplicate copies of the standalone tests.

jonbob:

PR #5478, opened 2023-02-27
ZM enhancements

crterai:
This PR merges in ZM enhancements that were developed as part of the v3 atmosphere physics NGD. 
The new ZM enhancements are: MCSP, ZM convective microphysics, Cloud-base mass flux adjustment, and Fix to tpert. 
Details of the various enhancements can be found here: https://acme-climate.atlassian.net/wiki/spaces/NGDAP/pages/3694460929/ZM+enhancement+integration

[Stealth]  for new features but...
[non-BFB] for OPT F and B tests with chrysalis_intel and pm-cpu_gnu because files must have optimization level changed even with features off.  DEBUG is BFB.


components/eam/src/physics/cam/zm_conv.F90
```diff
@@ -635,6 +642,9 @@ subroutine zm_convr(lchnk   ,ncol    , &
    real(r8), parameter :: mucon = 5.3_r8
    real(r8) negadq
    logical doliq
+
+   integer dcapemx(pcols)  ! maxi saved from 1st call for CAPE calculation;  used in 2nd call when DCAPE-ULL active
```
whannah1, line 646: Can you edit this to clarify what "maxi" is?

components/eam/src/physics/cam/clubb_intr.F90
```diff
@@ -2641,17 +2679,29 @@ subroutine clubb_tend_cam( &
...
-         ptend_loc%q(i,k,ixnumice) = 3._r8 * ( max(0._r8, ( dlf(i,k) - dlf2(i,k) )) *  dum1 ) &
-                                     / (4._r8*3.14_r8*clubb_ice_deep**3*500._r8) + & ! Deep    Convection
-                                     3._r8 * (                         dlf2(i,k)    *  dum1 ) &
-                                     / (4._r8*3.14_r8*clubb_ice_sh**3*500._r8)     ! Shallow Convection
-         ptend_loc%s(i,k)          = dlf(i,k) * dum1 * latice
+         if (zm_microp) then
+           ptend_loc%q(i,k,ixcldliq) = dlfzm(i,k) + dlf2(i,k) * ( 1._r8 - dum1 )
+           ptend_loc%q(i,k,ixcldice) = difzm(i,k) + dsfzm(i,k) +  dlf2(i,k) * dum1
+           ptend_loc%q(i,k,ixnumliq) = dnlfzm(i,k) + 3._r8 * ( dlf2(i,k) * ( 1._r8 - dum1 ) )   &
+                                                   / (4._r8*3.14_r8*clubb_liq_sh**3*997._r8)      ! Shallow Convection
```
whannah1, line 2686: I assume 3.14 here is "pi"? It would be nice to update this to use the "shr" value.

components/eam/src/physics/cam/macrop_driver.F90
```diff
@@ -321,6 +330,14 @@ subroutine macrop_driver_init(pbuf2d)
     CC_ni_idx   = pbuf_get_index('CC_ni')
     CC_qlst_idx = pbuf_get_index('CC_qlst')
 
+    dlfzm_idx = pbuf_get_index('DLFZM')
+    difzm_idx = pbuf_get_index('DIFZM')
+    dsfzm_idx = pbuf_get_index('DSFZM', err)
+    dnlfzm_idx = pbuf_get_index('DNLFZM', err)
+    dnifzm_idx = pbuf_get_index('DNIFZM', err)
+    dnsfzm_idx = pbuf_get_index('DNSFZM', err)
+    if (dnlfzm_idx > 0) zmconv_microp = .true.
```
whannah1, line 339: I don't like the look of this. Why do we need a separate "zmconv_microp" flag when we already have the namelist variable "zm_microp"? And why are we setting zmconv_microp based on the existence of "dnlfzm"? The existence of "dnlfzm" in the pbuf depends directly on "zm_microp=.true.", so let's just use "zm_microp" directly and not muddy the waters with a separate flag (unless there's a good reason that I'm not seeing).

components/eam/src/physics/cam/zm_conv.F90
```diff
@@ -39,10 +42,18 @@ module zm_conv
   public trig_dcape_only          ! true if to use dcape only trigger
   public trig_ull_only            ! true if to ULL along with default CAPE-based trigger
   public buoyan_dilute            ! subroutine that calculates CAPE
+  public zm_microp                ! true for convective microphysics
+  public MCSP
+  public MCSP_heat_coeff
+  public MCSP_moisture_coeff
+  public MCSP_uwind_coeff
+  public MCSP_vwind_coeff
```
whannah1, line 50: These need a descriptive comment

components/eam/src/physics/cam/zm_conv.F90
```diff
@@ -55,29 +66,45 @@ module zm_conv
...
    logical :: trig_ull_only    = .false. !true to use ULL along with default CAPE-based trigger
 
+
    real(r8) :: ke           ! Tunable evaporation efficiency set from namelist input zmconv_ke
    real(r8) :: c0_lnd       ! set from namelist input zmconv_c0_lnd
    real(r8) :: c0_ocn       ! set from namelist input zmconv_c0_ocn
    real(r8) :: dmpdz          = unset_r8  ! Parcel fractional mass entrainment rate (/m)
    real(r8) :: alfa_scalar  ! maximum downdraft mass flux fraction  
-   real(r8) :: tiedke_add     = unset_r8
+   real(r8) ::  tiedke_add    = unset_r8
```
whannah1, line 103: Why the weird extra space here?

components/eam/src/physics/cam/zm_conv.F90
```diff
@@ -100,7 +127,16 @@ module zm_conv
    
    integer,protected ::  limcnv   ! top interface level limit for convection
 
-   real(r8) :: tp_fac = unset_r8  ! PMA tunes tpert 
+   real(r8) :: tp_fac = unset_r8  ! PMA tunes tpert
```
whannah1, line 130: what does "PMA" stand for? Are these someone's initials? if so we should remove it.

components/eam/src/physics/cam/zm_conv.F90
```diff
@@ -540,6 +659,85 @@ subroutine zm_convr(lchnk   ,ncol    , &
    heat(:,:) = 0._r8
    mcon(:,:) = 0._r8
    rliq(:ncol)   = 0._r8
+   rice(:ncol)   = 0._r8    
+
+   if (zm_microp) then
```
whannah1, line 664: It would be nice to move all these initializations to a routine in "zm_microphysics" rather than add so many more lines to this already huge file.

components/eam/src/physics/cam/zm_conv.F90
```diff
@@ -886,8 +1544,152 @@ subroutine zm_convr(lchnk   ,ncol    , &
...
+         sprd(ideep(i),k) = sprdg(i,k)
+         dif (ideep(i),k) = dig  (i,k)
+         dsf (ideep(i),k) = dsg  (i,k)
+         dnlf(ideep(i),k) = dnlg  (i,k)
+         dnif(ideep(i),k) = dnig  (i,k)
+         dnsf(ideep(i),k) = dnsg  (i,k)
+         lambdadpcu(ideep(i),k) = lambdadpcug(i,k)
+         mudpcu(ideep(i),k)     = mudpcug(i,k)
+         frz(ideep(i),k)  = frzg(i,k)*latice/cpres
+         if (zm_microp) then
```
whannah1, line 1557: Similar to my comment above about moving the allocation and initialization to their own routines, I think having a copy routine to collapse this whole block to a  single line would be a good idea.

components/eam/src/physics/cam/zm_conv.F90
```diff
@@ -886,8 +1544,152 @@ subroutine zm_convr(lchnk   ,ncol    , &
...
+           microp_st%fhmrm (ideep(i),k) = loc_microp_st%fhmrm (i,k)
+         end if
       end do
    end do
+
+   if (zm_microp) then
+     do k = msg + 1,pver
+       do i = 1,ncol
+
+         !Interpolate variable from interface to mid-layer and convert it from units of "kg/kg" to "g/m3"
```
whannah1, line 1649: FYI, this is a crude interpolation that won't work well for a highly stretched grid. 
Also, the comment mentions a unit conversion, but the actual code doesn't seem to do any unit conversion until the 3rd block down. I would suggest putting the comment directly adajacent to the relevant block for better readability.

components/eam/src/physics/cam/zm_conv.F90
```diff
@@ -886,8 +1544,152 @@ subroutine zm_convr(lchnk   ,ncol    , &
...
+            microp_st%qgraupel(i,k) = 0.5_r8*(microp_st%qgraupel(i,k)+microp_st%qgraupel(i,k+1))
+            microp_st%qni (i,k) = 0.5_r8*(microp_st%qni(i,k)+microp_st%qni(i,k+1))
+            microp_st%qnl (i,k) = 0.5_r8*(microp_st%qnl(i,k)+microp_st%qnl(i,k+1))
+            microp_st%qnr (i,k) = 0.5_r8*(microp_st%qnr(i,k)+microp_st%qnr(i,k+1))
+            microp_st%qns (i,k) = 0.5_r8*(microp_st%qns(i,k)+microp_st%qns(i,k+1))
+            microp_st%qng (i,k) = 0.5_r8*(microp_st%qng(i,k)+microp_st%qng(i,k+1))
+            microp_st%wu(i,k)   = 0.5_r8*(microp_st%wu(i,k)+microp_st%wu(i,k+1))
+         end if
+
+         if (t(i,k).gt. 273.15_r8 .and. t(i,k-1).le.273.15_r8) then
```
whannah1, line 1665: Having "magic numbers" down in the code like this is one of my pet peeves. We should really be using the "shr" values for things like this, or at the very least put parameters like this at the top of the code as module level constant parameters, or at the top of the routine, so they are all in one place and we avoid using slightly different values in different parts of the code.

components/eam/src/physics/cam/zm_conv.F90
```diff
@@ -916,10 +1718,87 @@ subroutine zm_convr(lchnk   ,ncol    , &
...
-         rliq(i) = rliq(i) + dlf(i,k)*dpp(i,k)/gravit
+          rliq(i) = rliq(i) + (dlf(i,k)+dif(i,k)+dsf(i,k))*dpp(i,k)/gravit
+          rice(i) = rice(i) + (dif(i,k)+dsf(i,k))*dpp(i,k)/gravit
       end do
    end do
    rliq(:ncol) = rliq(:ncol) /1000._r8
+   rice(:ncol) = rice(:ncol) /1000._r8    
+
+   if (zm_microp) then
+     deallocate( &
```
whannah1, line 1729: see comment above about moving this to it's own routine in the appropriate module

components/eam/src/physics/cam/zm_conv.F90
```diff
@@ -1004,18 +1888,41 @@ subroutine zm_conv_evap(ncol,lchnk, &
     flxprec(:ncol,1) = 0._r8
     flxsnow(:ncol,1) = 0._r8
     evpvint(:ncol)   = 0._r8
+    omsm=0.99999999_r8
```
whannah1, line 1891: This is a weird number. it deserves a comment explaining what it's for and why the particular precision was chosen.

components/eam/src/physics/cam/zm_conv.F90
```diff
@@ -1004,18 +1888,41 @@ subroutine zm_conv_evap(ncol,lchnk, &
...
+        if( old_snow ) then
           if (t(i,k) > tmelt) then
              flxsntm(i) = 0._r8
              snowmlt(i) = flxsnow(i,k) * gravit/ pdel(i,k)
           else
              flxsntm(i) = flxsnow(i,k)
              snowmlt(i) = 0._r8
           end if
+        else
+! make sure melting snow doesn't reduce temperature below threshold
```
whannah1, line 1906: indent this comment.

components/eam/src/physics/cam/zm_conv.F90
```diff
@@ -1159,7 +2083,7 @@ subroutine convtran(lchnk   , &
    integer, intent(in) :: nstep             ! Time step index
 
    real(r8), intent(in) :: dpdry(pcols,pver)       ! Delta pressure between interfaces
-
+   real(r8), intent(in) :: dt               ! 2 delta t (model time increment)
```
whannah1, line 2086: Are you sure this is twice the physics time step?

components/eam/src/physics/cam/zm_conv.F90
```diff
@@ -2559,25 +3716,34 @@ subroutine cldprp(lchnk   , &
...
+      totfrz(i)= 0._r8
+      do k = pver,msg + 1,-1
+         totfrz(i)= totfrz(i)+ frz(i,k)*dz(i,k)
+      end do
    end do
    do k=klowest-2,khighest-1,-1
       do i=1,il2g
          if (doit(i) .and. k <= jb(i)-2 .and. k >= lel(i)-1) then
-  	        if (hu(i,k) <= hsthat(i,k) .and. hu(i,k+1) > hsthat(i,k+1) .and. mu(i,k) >= 0.02_r8) then
+            if (hu(i,k) <= hsthat(i,k) .and. hu(i,k+1) > hsthat(i,k+1) .and. mu(i,k) >= 0.02_r8) then
```
whannah1, line 3727: here's another example of a "magic" threshold buried in the code that would be better to have defined elsewhere along with other constants and thresholds

components/eam/src/physics/cam/zm_conv.F90
```diff
@@ -2688,27 +3797,133 @@ subroutine cldprp(lchnk   , &
...
+      end do
+
+      do k = 1, pver
+         do i = 1,il2g
+            loc_microp_st%cmei(i,k) = cu(i,k)* fice(i,k)
+            loc_microp_st%cmel(i,k) = cu(i,k) * (1._r8-fice(i,k))
+         end do
+      end do
+
+      call  zm_mphy(su,   qu,    mu,    du,   eu,  loc_microp_st%cmel,  loc_microp_st%cmei,   zf,    p,   &
```
whannah1, line 3864: Couldn't we just pass the loc_microp_st variable as a single argument to clean up this interface?

components/eam/src/physics/cam/zm_conv_intr.F90
```diff
@@ -10,15 +10,28 @@ module zm_conv_intr
 ! Author: D.B. Coleman
 ! January 2010 modified by J. Kay to add COSP simulator fields to physics buffer
 ! July 2015 B. Singh Added code for unified convective trasport
+! March 2023 X. Song added code for the initialization of convective microphysics
+!            and aerosol object, and the output of microphysical properties and tendencies
 !---------------------------------------------------------------------------------
    use shr_kind_mod, only: r8=>shr_kind_r8
-   use physconst,    only: cpair
+   use physconst,    only: cpair
```
whannah1, line 17: Looks like extra white space was added by accident?

components/eam/src/physics/cam/zm_conv_intr.F90
```diff
@@ -323,8 +708,8 @@ subroutine zm_conv_tend(pblh    ,mcon    ,cme     , &
 
    ! physics types
    type(physics_state) :: state1        ! locally modify for evaporation to use, not returned
-   type(physics_ptend) :: ptend_loc     ! package tendencies
-
+!   type(physics_ptend) :: ptend_loc     ! package tendencies
```
whannah1, line 711: delete the commented line

oksanaguba:
Great observations, @whannah1 . Since this development in v3 has a lot of issues with conservation, i would like to test it (or test the default, which i assume is oprion zm_microp=false) wrt energy and mass before this goes in.

oksanaguba:
Additional request -- to add any new options like zm_microp to the log, so that in old runs someone could confirm which options were used.

components/eam/src/physics/cam/zm_conv.F90
DJFJJA: "maxi" is replaced by "launching level index".

components/eam/src/physics/cam/clubb_intr.F90
DJFJJA: 3.14 is replaced by physical constant pi.

components/eam/src/physics/cam/macrop_driver.F90
DJFJJA: "zm_microp" is now directly used in macrop_driver.F90.

components/eam/src/physics/cam/zm_conv.F90
DJFJJA: Descriptive comment is added.

components/eam/src/physics/cam/zm_conv.F90
DJFJJA: Extra space is removed.

components/eam/src/physics/cam/zm_conv.F90
DJFJJA: "PMA" stands for "Po-Lun Ma", which is now removed in the comment.

components/eam/src/physics/cam/zm_conv.F90
DJFJJA: All allocations and initializations are moved into corresponding subroutines.

components/eam/src/physics/cam/zm_conv.F90
DJFJJA: This block is moved into a subroutine.

components/eam/src/physics/cam/zm_conv.F90
DJFJJA: The comment is moved to the relevant block. The crude interpolation is only used for output, which is not used in model integration.

components/eam/src/physics/cam/zm_conv.F90
DJFJJA: 273.15 is replaced by physical constant "tmelt".

components/eam/src/physics/cam/zm_conv.F90
DJFJJA: Deallocations are moved into corresponding subroutine.

components/eam/src/physics/cam/zm_conv.F90
DJFJJA: It is a number near unity to prevent problems due to round off error. For consistency with MG2, omsm=0.99999_r8 is used.

components/eam/src/physics/cam/zm_conv.F90
DJFJJA: Done.

components/eam/src/physics/cam/zm_conv.F90
DJFJJA: This is not related to this PR or ZM enhancement. It has been there since NCAR CCM3 (~20 years ago). It is still true for Eul dynamics but not for SE dynamics. There are a lot of similar statement (2 delta t) in the E3SMv2. For now, I would suggest keeping it as is.

components/eam/src/physics/cam/zm_conv.F90
DJFJJA: This is not related to this PR. I added a new parameter mu_min=0.02_r8, and replaced 0.02_r8 in the code with mu_min.

components/eam/src/physics/cam/zm_conv.F90
```diff
@@ -2688,27 +3797,133 @@ subroutine cldprp(lchnk   , &
...
+               fice(i,k) = 0._r8
+
+            else if (tug(i,k+1) < 233.15_r8) then
+               ! If colder than tmin then ice phase
+               fice(i,k) = 1._r8
+
+            else
+               ! Otherwise mixed phase, with ice fraction decreasing linearly
+               ! from tmin to tmax
+               fice(i,k) =(273.15_r8 - tug(i,k+1)) / 40._r8
```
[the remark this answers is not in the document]
DJFJJA: tmin=233.15K (or -40C) represents homogeneous freezing temperature. tmax=273.15K represents the freezing point of water.  In the mixed phase (tmin<T<tmax), the ice fraction linearly increases from 0 at tmax to 1 at tmin. The denominator 40= tmax-tmin=273.15-233.15. We replaced 273.15_r8 with tmelt and defined new parameter t_homofrz=tmelt-40._r8. Then denominator is changed to (tmelt-t_homofrz).

components/eam/src/physics/cam/zm_conv.F90
DJFJJA: To do so, we need to rewrite zm_microphysics.F90 and would not meet the deadline.

components/eam/src/physics/cam/zm_conv_intr.F90
DJFJJA: Extra space is removed.

components/eam/src/physics/cam/zm_conv_intr.F90
DJFJJA: Done.

oksanaguba:
I ran 3 runs, 1)master and this branch with 2)zmconv_microp=false and 3)true for an ne30 F case. I see some issues, in particular, active clipping in cam_export for zmconv_microp=true. Results and script are in https://acme-climate.atlassian.net/wiki/spaces/COM/pages/3739976325/Water+energy+conservation+Part+3 . However, it would be great if Chris or Xiaoliang repeat these runs to double check. Starting from Run3 is probably easiest. Thank you!

components/eam/src/physics/cam/macrop_driver.F90
```diff
@@ -680,28 +712,41 @@ subroutine macrop_driver_tend(                             &
...
-           (4._r8*3.14_r8*25.e-6_r8**3*500._r8) + & ! Deep    Convection
-           3._r8 * (                         dlf2(i,k)    *  dum1 ) / &
-           (4._r8*3.14_r8*50.e-6_r8**3*500._r8)     ! Shallow Convection
-      ptend_loc%s(i,k)          = dlf(i,k) * dum1 * latice
+       if (zm_microp) then
+          ptend_loc%q(i,k,ixcldliq) = dlfzm(i,k) + dlf2(i,k) * ( 1._r8 - dum1 )
+          ptend_loc%q(i,k,ixcldice) = difzm(i,k) + dsfzm(i,k) +  dlf2(i,k) * dum1
+                               
+          ptend_loc%q(i,k,ixnumliq) = dnlfzm(i,k) + 3._r8 * ( dlf2(i,k) * ( 1._r8 - dum1 ) )   &
+                                                   / (4._r8*3.14_r8*10.e-6_r8**3*997._r8)      ! Shallow Convection
```
rljacob, line 722: What are all these magic numbers?  997.?  10.e-6 cubed?

DJFJJA:
It specifies the radius of cloud liquid drop as 10 microns (10.e-6 m) and
the density as 997. kg/m3, respectively.
All these come from NCAR CESM. When CLUBB is used, the macrop_driver will
not be called in the model.

Xiaoliang

components/eam/src/physics/cam/macrop_driver.F90
DJFJJA: It specifies the radius of cloud liquid drop as 10 microns (10.e-6 m) and the density as 997. kg/m3, respectively.
All these come from NCAR CESM. When CLUBB is used, the macrop_driver will not be called in the model.

whannah1: At a minimum, these numbers should be explained with a comment. Ideally, the numbers should be set as parameters elsewhere. Also, we shouldn't be redefining pi all over the code, but I suppose we can let that slide here.

oksanaguba:
Screen Shot 2023-05-01 at 9 48 00 AM

Above is imbalance for 2 runs, this branch with merged RR diagnostics branch with and without zm microp option like in
```
print_energy_errors  = .true.
use_mass_borrower = .true.
!zmconv_microp = .true.
zmconv_microp = .false.
```
Energy plot contains difference between the model state change (TE1-TE2) and in/out fluxes, water plot contains the same difference, but with respect to water, difference of (totalWater1 - totalWater2) and water fluxes.

Runs had mass borrower on and no diagnostics for leaks caused by qneg4. Runs are 2 months, run without zm microp has 2 water column errors from MG and 4 energy errors from ZM, run with zm microp has no column errors.

Conclusion -- in these short runs, ZM microp option is better wrt infrequent peaks in water and energy imbalance and std of imbalance.

components/eam/src/physics/cam/zm_conv_intr.F90
```diff
@@ -410,22 +942,216 @@ subroutine zm_conv_tend(pblh    ,mcon    ,cme     , &
...
+           Qmq(i,k) = -1._r8 * Qv_zmconv(i) * alpha2 * sin(2.0_r8*pi*(top/bottom))
+           Qmq(i,k) = Qm(i,k)/2500000._r8/4._r8
+
+
+           !if( cos(pi*(top/bottom))>0 ) then 
+              Qmu(i,k) = alphau * (cos(pi*(top/bottom)))
+           !else 
+           !   Qmu(i,k) = alphau * (cos(pi*(top/bottom)))*.25_r8
+           !end if
+
```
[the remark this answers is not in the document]
whannah1: If we're keeping this change we should delete the commented lines. A better way to preserve the previous use of thealtered Qmu value is to write a comment paragraph explaining how it used to be different and why it was changed to remove the 0.25 factor.

components/eam/src/physics/cam/zm_microphysics.F90
```diff
@@ -0,0 +1,3358 @@
...
+	bi = 1._r8
+
+! rain
+	ar = 841.99667_r8
+	br = 0.8_r8
+
+!graupel(if dense precipitating ice is graupel)
+
+        ag = 19.3
+        bg = 0.37
```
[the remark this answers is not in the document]
crterai: Good catch. I'm adding new commits that address this and a set of other examples where the type of the number wasn't specified.

crterai:
> @crterai @DJFJJA do the new commits make any difference in your estimation/testing? [eaad3d3..8ac0e99](https://github.com/E3SM-Project/E3SM/pull/5478/files/235cb00e61ff220a34b979f4097b4734fdcea8d7..8ac0e9995545881fe3097596bc3311a5b79dee93)

@mahf708 - Specifying the precision of the numbers or initializing icimrdp did not have direct impact on the testing. I've included them because I figured it's good practice to specify these things. 
@DJFJJA's commit to fix the ERS certainly fixed the ERS tests. And more recently, reducing the optimization helped with the non-BFB behavior on pm-cpu. I'll make the same changes for chrysalis in hopes of fixing the machine-dependent behavior that you'd reported.

cime_config/machines/Depends.chrysalis.intel.cmake
```diff
@@ -1,7 +1,10 @@
 # Reduce optimization on files with long compile times
+# Also reduce optimizaiton on files with machine-dependent behavior
 set(O2MODELSRC
   eam/src/physics/cam/micro_mg_cam.F90      # ~2027 seconds
   elm/src/data_types/VegetationDataType.F90 # ~ 930 seconds
+  eam/src/physics/cam/zm_conv.F90           # for machine dependent behavior
+  eam/src/physics/cam/zm_microphysics.F90   # for machine dependent behavior
 )
```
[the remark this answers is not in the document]
crterai: Thanks for the heads up @mahf708. I'll manually resolve if and when it becomes necessary.

crterai:
> @crterai testing on [2a41ffc](https://github.com/E3SM-Project/E3SM/commit/2a41ffc9b5058f8ab328b5c7b645efbaf5a53adf), I believe the machine dependency issue has been significantly reduced (if not eliminated altogether).
> 
@mahf708  - Thanks for testing the new code changes, and that's good news. 

@wlin7 - Based on the above, and new testing results, which show that the BFB issues have been tracked down with this branch, I think this PR is in good shape. Do you think I should re-run a full F2010 and WCYCL1850 set of simulations just to remove surprises?

wlin7:
@crterai , did you mean with ZM on? It should be useful to do a run with the latest update. Maybe just F2010, and if no surprise, no need to run WCYCL1850.

wlin7:
@mahf708 , we can look to merge tomorrow or the day after. Chris will run a new simulation use this code base just to make sure.

wlin7:
Thanks for the great and quick work to clear all the doubts, @mahf708 . This should affirm the PR is ready even without additional run by Chris.

wlin7:
Preliminary merged next confirmed: ERP, PET, PEM tests are BFB with base tests, though expected NBFB with the respective baselines due to changes in optimization level. SMS_D test is BFB with baseline, also as expected.

wlin7:
Merged to next.

wlin7:
This PR causes all MMF1 tests to fail due to ICIMRDP not present in pbuf for MMF1 config.

wlin7:
Verified with the fix [85ab2a3](https://github.com/E3SM-Project/E3SM/pull/5478/commits/85ab2a334b83d843fe11ce364132cf0251ce0f99), MMF1 tests BFB with baselines (not affected by this PR), and other tests BFB before and after this fix.

crterai:
Thank you for making the fix @wlin7.

wlin7:
Thanks for checking out the commit, @crterai . I was about to ask you to take a quick look. 
The PR has been remerged to next. Since MMF1 tests will not be affected (by the PR with the fix), I think we are ready merge to master.

crterai:
@wlin7 I have added `Depends.anlgce.gnu.cmake` and updated comments with `fe44772`.

wlin7:
Merged to master.

rljacob:
next testing with baselines:

- anvil prod:  namelist diff only
- chrysalis prod:  namelist diff only even though OPT changed on two files
- chrysalis integration:
- chrysalis non-bfb:  namelist diff only even though OPT changed on two files
- compy prod: namelist only
- gcp12 extra-cov:  namelist diff only
- mappy developer:
- pm-cpu integration gnu:  namelist AND baseline diff for all EAM cases except ERS_Ln9.ne4pg2_ne4pg2.F2010-MMF1.pm-cpu_gnu.eam-mmf_crmout
- pm-cpu developer nvidia:  namelist diff only
- pm-cpu integration intel:  namelist diff only.

PR #5481, opened 2023-02-28
Hommexx: Add doubly-periodic capability.

### LINKED ISSUE #4212: homme planar preqx test using theta
In https://github.com/E3SM-Project/E3SM/blob/a72d6708a0349d911a27ebad635a529c6266e08e/components/homme/test/reg_test/run_tests/preqx-nhgw.cmake#L10 in a preqx test theta-l exec is used. Same with the other preqx planar test.

ambrad:
* Implement doubly-periodic mode in Hommexx.
* Add C++-vs-F90 tests for the F90 nhgw and bubble tests.

Fixes #4212 .

[BFB] except for standalone Homme F90 bubble tests; adds new standalone Homme tests


bartgol (approved):
Looks good. At some point it would be nice to add some more "intrinsic" tests, like basic grad/curl/div identities or lapl==div(grad(u)). I forgot if they (all) hold in curved manifolds, but they could at least be tested in planar geometries.

Obv, not for this pr.

components/homme/src/test_src/dry_planar.F90
```diff
@@ -283,6 +282,10 @@ subroutine bubble_init(elem,hybrid,hvcoord,nets,nete,f)
 
   hvcoord%hyai = ai; hvcoord%hybi = bi
 
+  do k = 1,nlev
```
oksanaguba, line 285: why this change?

ambrad: Previously, it was computing it according to the analytical formula. But that is not BFB in finite precision with the pressure_thickness call, which is the normal way to init dp3d in standalone tests. This was causing non-BFBness in the C++-vs-F90 test for the bubble tests, since Hommexx's initialization assumes that dp3d is computed w.r.t. ps by the usual formula. Differences in dpm in bubble_init are at machine precision.

oksanaguba: Ok, my first thought was to revert it. I do not believe we should stick to dcmip wrappers code. However if that it too much to change xx init, then it is ok to keep it. i think xx init should have mimic-ed F init though (i assume it is about routine set_elem_state() ).

ambrad: The error, in my view, is in bubble_init. It sets up ps and dp3d in such a way that even if nothing happened in the first time step, dp3d would change just due to using the standard formula in, e.g., vertical remap rather than bubble_init's computation. In contrast, all the other standalone tests use pressure_thickness for finite-precision consistency.

oksanaguba: There is no reason to use A, B in dp3d away from init, and in init it is only a convenience. I do not see it as an error, however, it is a bigger change to fix this in xx, so let's keep this change.

oksanaguba (approved):

components/homme/src/test_src/dry_planar.F90
ambrad: Just for clarity, by consistency I mean e.g. with:
https://github.com/E3SM-Project/E3SM/blob/c7b8bd8d81c4f8f22c5686dc0a17c37c68ef95d1/components/homme/src/theta-l/share/vertremap_mod.F90#L81-L82
If we want to remove a and b, it's not just an init-time issue or an Hxx issue.

oksanaguba: I understand. At this point it is just a philosophical conversation about hybrid coefs, which we can postpone.

ambrad: To ddcument this further, in Chrysalis, homme_integration shows diffs for the F90 bubble tests at machine precision level, consistent with the machine-precision change to dpm in bubble_init:
```
############################################################################
CPRNC returned the following RMS differences
 RMS T                                7.0545E-13            NORMALIZED  2.4100E-15
 RMS Th                               4.4738E-14            NORMALIZED  1.4912E-16
 RMS ps                               9.2247E-11            NORMALIZED  9.2247E-16
 RMS T                                8.1605E-12            NORMALIZED  2.7882E-14
 RMS Th                               6.1777E-12            NORMALIZED  2.0591E-14
 RMS geo                              4.3877E-11            NORMALIZED  5.9630E-15
############################################################################
```

ambrad:
On Ascent, optimized and debug SCREAMv1 ERS tests pass with this PR merged in and this change made to EAMxx: https://github.com/E3SM-Project/scream/pull/2198.

ambrad:
With the fix to SyncUtils, homme_integration passes on Chrysalis except with the expected fails:
```
HOMMEBFB_P24.f19_g16_rx1.A.chrysalis_intel
    1 - verifyBaselineResults (Failed)
   37 - thetanh-moist-bubble (Failed) // diffs are due to bubble_init pressure_thickness commit
   38 - thetanh-dry-bubble (Failed)
   42 - preqx-nhgw-kokkos (Failed) // diff due to fixing the executable
   43 - preqx-nhgw-slice-kokkos (Failed)
   45 - thetanh-moist-bubble-kokkos (Failed)
   46 - thetanh-dry-bubble-kokkos (Failed)
   47 - thetah-nhgw-kokkos (Failed)
   48 - thetanh-nhgw-kokkos (Failed)
   49 - thetah-nhgw-slice-kokkos (Failed)
   50 - thetanh-nhgw-slice-kokkos (Failed)
HOMME_P24.f19_g16_rx1.A.chrysalis_intel: both diffs are due to bubble_init pressure_thickness commit
   28 - thetanh-moist-bubble (Failed)
   29 - thetanh-dry-bubble (Failed)
```

ambrad:
e3sm_developer on Chrysalis is as expected.

PR #5484, opened 2023-02-28
EAM: don't use p3 settings if microphs is not p3

jgfouca:
EAM: don't use p3 settings in namelist if microphs is not p3

[BFB]
[NML]


rljacob (approved):
So that's what was missing!  Cool.

jgfouca:
Looks like cases both with and without microphys='p3' work as expected. Merging to next.

PR #5488, opened 2023-03-01
New MPAS-SI + WW3 compset (D-compset plus ww3)

erinethomas:
New WW3 compset option for MPAS-SI plus waves. For testing purposes only. 
Modification to D-JRA compset alias - should include 'TEST' in alias. 

[BFB]

### DIFF
--- a/components/mpas-seaice/cime_config/config_compsets.xml
+++ b/components/mpas-seaice/cime_config/config_compsets.xml
@@ -10,7 +10,7 @@
   <!-- E3SM D compsets -->
 
   <compset>
-    <alias>DMPAS-JRA1p5</alias>
+    <alias>DTEST-JRA1p5</alias>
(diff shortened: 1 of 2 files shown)


jonbob:
@erinethomas - did you decide you wanted "TEST" in the compset alias to help indicate that any runs with this compset may not be scientifically valid?

erinethomas:
I removed 'TEST' in the alias to follow the alias convention used by the new D-JRA1p5 compset made by @darincomeau. 
I think both compsets (D-JRA - defined in mpas-si, and the D-JRA-WW3 compset defined in ww3) should match. 
Possibly they should both include 'TEST' in the alias to indicate they are not scientifically validated.

jonbob:

jonbob:
merged to master

PR #5496, opened 2023-03-03
Implement alternating control of Fortran P3 codes for eam and scream

wlin7:
The alternating switch, determined by the case configuration option (a scream case or else), serves to control which version of Fortran P3 codes being used to build the model. A test is needed to track scream configuration. The version for the EAM version is currently just a skelton, which will be replaced by the eventual P3 PR from v3atm

[BFB] Two new tests SMS_Ln9_P24x1.ne4_ne4.FDPSCREAM-ARM97
          and SMS_Ld1.ne30pg2_ne30pg2.F2010-SCREAM-LR-DYAMOND2


wlin7:
The following tests verified BFB between master and this branch, with one using mg2 (but p3 codes built in) and the other with scream p3. 

- SMS_Ld1.ne30pg2_EC30to60E2r2.F2010.cori-knl_intel 
- SMS_Ld1.ne30pg2_ne30pg2.F2010-SCREAM-LR-DYAMOND2.cori-knl_intel

When reviewing, please ignore files in eam/src/physics/p3/eam. Files there are merely skeltons extracted from the actual p3_PR, and will be replaced entirely when the p3_PR is merged.

rljacob (approved):
looks good!

crterai (approved):
Thanks for creating this, @wlin7. It looks good to me, especially given that your tests with SCREAM compset and non-P3 compsets come back BFB with master before merging.

cime_config/tests.py
```diff
@@ -210,6 +210,7 @@
             "SMS_D_Ld1.T62_oEC60to30v3.DTESTM",
             "SMS_D_Ld3.T62_oQU120.CMPASO-IAF",
             "SMS_D_Ld1.ne30pg2_r05_EC30to60E2r2.WCYCL1850",
+            "SMS_Ln5.ne30pg2_ne30pg2.F2010-SCREAM-LR-DYAMOND2",
```
PeterCaldwell, line 213: I'd like to see a DP test as well.

components/eam/bld/configure
```diff
@@ -1752,6 +1752,12 @@ if (defined $opts{'cppdefs'}) {
 }
 $cfg_ref->set('cppdefs', $usr_cppdefs);
 
+# define is_scream_config flag to select which P3 version codes to use
+my $is_scream_config = 0;
+if ($usr_cppdefs =~ /\-DSCREAM/) {
```
PeterCaldwell, line 1757: Sanity check question for @jgfouca or @bartgol or someone in the know - does -DSCREAM in cppdefs mean that the simulation is either a SCREAM F90 or a SCREAM C++ run, or does it just imply one of these 2 options?

bartgol: Good catch!

IIRC, that's used only by Homme. But I _think_ we never tested it for SCREAMv0, so I am not sure what it would produce there. Grepping in Homme folder, all but a few of the matches are of the form
```
```
or its opposite without the `!`'s. There are only a few cases where CAM and SCREAM would be treated differently:
- adjust_ps is `false` for SCREAM, and `true` otherwise
- for theta-l_kokkos target, if SCREAM is defined, forcing is _not_ pushed back to f90 (I guess we assume we're running eamxx), while we do push to f90 for CAM.

I suspect the last of the two points above would cause wrong results in this case. I would leave `-DSCREAM` out of the compile line, and use `-DCAM` instead.

PeterCaldwell: Yeah, I have vague recollection that `DSCREAM` does not mean what a normal person would think. Maybe we should create a git issue in SCREAM to figure out whether we still need it or not? Though maybe it will naturally go away when we remove support for the F90 code in a few months?

bartgol: EAMxx _does_ need `-DSCREAM` (and in fact we do have it). But I think v0 should be considered as "EAM" from the point of view of Homme.

PeterCaldwell:
@wlin7 - I'd like to see a DP test case and `-DSCREAM` removed from the configure file, but otherwise I'm happy with this PR.

wlin7:
Sorry I had an illusion this PR already merged and not watching. DP-SCREAM test added to atm_developer (and tested working well). Note that there is another global LR test for SCREAM in the integration test suite.

@PeterCaldwell , regarding use of `-DSCREAM`, just needing a handle in configure to determine it is a SCREAM case. This argument set in CAM_CONFIG_OPTS via config_component.xml is a handy one. EAM's configure does not see the actual compset used. If not using this argument, we could use `-spa` and/or `-shoc_sgs`, as they are currently only used for SCREAM cases. Maybe less safe if these physics options are extended for other configurations.

wlin7:
Good points, @PeterCaldwell . That does suggest it is not reliable to use those physics option as an indication of scream config. 

How about keeping `-DSCREAM` for now. It is set in EAM's config_component.xml and passed to configure, and it has an impact on calculation concerning adjust_ps as @bartgol pointed above. When we remove cppdef of SCREAM altogether, we can pass another innocuous flag to `configure` for the sole purpose of informing it is a SCREAM case -- and not to be used as a cppdef.

wlin7:
Yeah, I agreed with @bartgol 's assessment. No effect mostly because -DCAM and -DSCREAM are used redundantly. The exception is in homme/src/share/prim_driver_base.F90 and homme/src/theta-l_kokkos/prim_driver_mod.F90, what has statements solely condition on -DSCREAM.

wlin7:
The [9f2d8bb](https://github.com/E3SM-Project/E3SM/pull/5496/commits/9f2d8bbc5623bf7bd7550fc87296bfc62aeff98a) now determines a scream case when p3 is used together with shoc_sgs. DPSCREAM and a regular EAM F2010 cases have been tested and verified the proper p3 version is selected for the build.

bartgol (approved):
This is fine from the point of view of homme now. I can't comment much on the rest.

cime_config/tests.py
wlin7: DP test added to the developer test suite.

rljacob: DP = "doubly-periodic" ?   Hard to tell from the name that its that kind of domain.

brhillman:
Merged to next.

brhillman:
gcp12 tests had runtime fails on next:
```
 1:  ERROR timeaddmonths():  MM out of range
 1:  set_time_float_from_date: error return from ESMF_TimeSet for set_time_float_from_date
 1:  ERROR: CHKRC
```

brhillman:
Merged to master. Note that the issues on gcp12 and on ascent ibm will need to be resolved in a subsequent PR.

PR #5505, opened 2023-03-06
Remove tests of old RRM configurations

rljacob:
Remove F-case tests that use out-dated RRM grids which no longer have land conditions needed to run:
ne0np4_enax4v1: 1-deg with 1/4-deg over Eastern North Atlantic (version 1)
ne0np4_twpx4v1: 1-deg with 1/4-deg over Tropical West Pacific (version 1)

We are still testing the fully coupled RRM case in the production suite:
ne0np4_northamericax4v1.pg2
1-deg with 1/4-deg over North America (version 1) pg2
WC14to60E2r3
MPAS with enhanced resolution of 14km along the coast of North America, the Northern Atlantic subpolar gyre, and the Arctic Ocean.

[BFB]

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -217,8 +217,6 @@ _TESTS = {
     "e3sm_extra_coverage" : {
         "inherit" : ("e3sm_atm_extra_coverage", "e3sm_ocnice_extra_coverage"),
         "tests"   : (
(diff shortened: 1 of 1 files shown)


tangq (approved):
The changes look great!

wlin7 (approved):
That is good. These two tests are not working properly anyway.

rljacob:
merged to next

Issue #5353, opened 2022-12-07
Problem with nvidia compiler version 22.7/22.9 -- runtime error: add_cam_grid_attribute_1d_r8: attribute dimension ncol for area, not found (on pm-cpu) 


ndkeen:
Using master of Dec6th, several tests are hitting this error with a version of nvidia (22.7) compiler that is NOT currently on master (currently we have 22.5). We do not see these errors with version 22.5
I also tested version 22.9 (available on test machine at NERSC) and see same issue.

```
ERP_Ln18.ne4_oQU240.F2010.pm-cpu_nvidia
ERS.ne4_oQU240.F2010.pm-cpu_nvidia.eam-hommexx
ERS_Ld3.ne4pg2_oQU480.F2010.pm-cpu_nvidia.eam-thetahy_sl_pg2
ERS_Ld3.ne4pg2_oQU480.F2010.pm-cpu_nvidia.eam-thetahy_sl_pg2_ftype0
SMS.ne4_oQU240.F2010.pm-cpu_nvidia.eam-cosplite
SMS_D_Ln5.ne4_oQU240.F2010.pm-cpu_nvidia
SMS_Ln5.ne4pg2_oQU480.F2010.pm-cpu_nvidia.eam-thetahy_pg2
SMS_Ln5.ne4pg2_oQU480.F2010.pm-cpu_nvidia.eam-thetahy_sl_pg2
SMS_Ln5.ne4pg2_oQU480.F2010.pm-cpu_nvidia.eam-thetahy_sl_pg2_ftype0
SMS_Ln5.ne4pg2_oQU480.F2010.pm-cpu_nvidia
SMS_Ln9.ne4_oQU240.F2010.pm-cpu_nvidia.eam-outfrq9s
SMS_P12x2.ne4_oQU240.WCYCL1850NS.pm-cpu_nvidia.allactive-mach_mods
SMS_R_Ld5.ne4_ne4.FSCM-ARM97.pm-cpu_nvidia.eam-scm
```

Error happens in init
```
0: (seq_comm_printcomms)    38     0     1     1  CPLIAC:
0:  ERROR:
0:   add_cam_grid_attribute_1d_r8: attribute dimension ncol for area, not found
0: MPICH ERROR [Rank 0] [job id 3921779.0] [Tue Dec  6 16:19:15 2022] [nid005976] - Abort(1001) (rank 0 in comm 0): application called MPI_Abort(MPI_COMM_WORLD, 1001) - process 0
[...]

PR #5507, opened 2023-03-06
Add missing variable to the overloaded copy assignment operator for parallel_t

ndkeen:
Add missing variable to the overloaded copy assignment operator for parallel_data structure.
Recently added variable to `parallel_t` was missing from the `copy_par` routine.
`par2%dynproc    = par1%dynproc`
Which, for whatever reason, was only a problem with `nvidia/22.7` (and higher).

Be more careful about setting initial values for 2 integers in routine. 
Example: `integer :: ikool=4` is dangerous as we may expect it to be set each entry, but instead, I think it's only set on the first call. Which is fine here as it's only called once, just being more careful.

BFB

### DIFF
--- a/components/homme/src/share/parallel_mod.F90
+++ b/components/homme/src/share/parallel_mod.F90
@@ -103,6 +103,7 @@ contains
     par2%nprocs     = par1%nprocs
     par2%comm       = par1%comm
     par2%masterproc = par1%masterproc
+    par2%dynproc    = par1%dynproc
 
   end subroutine copy_par
 
@@ -124,12 +125,15 @@ contains
     integer(kind=int_kind)                              :: ierr
     logical :: running   ! state of MPI at beginning of initmp call
 #ifdef CAM
-    integer :: color = 1
+    integer :: color
     integer :: iam_cam, npes_cam
     integer :: npes_homme
     integer :: max_stride
 #endif
-    integer :: npes_cam_stride = 1
+    integer :: npes_cam_stride
+
+    color = 1
+    npes_cam_stride = 1
     !================================================
     !     Basic MPI initialization
     ! ================================================
--- a/components/homme/src/share/parallel_mod.F90
+++ b/components/homme/src/share/parallel_mod.F90
@@ -52,11 +52,10 @@ module parallel_mod
     integer :: root                       ! local root
     integer :: nprocs                     ! number of processes in group
     integer :: comm                       ! local communicator
-!    integer :: node_comm                  ! local communicator of all procs per node
-!    integer :: node_rank                  ! local rank in node_comm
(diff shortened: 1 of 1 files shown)


mt5555:
This complexity of this design ( parallel_t user defined type, with user defined assignement operator) seems to be so that we can have multiple parallel_t objects all with different values.  But I cant undestand why HOMME would ever need more than one such object (and hence could use the default assignment operator ( "shallow copy", instead of a "deep copy")?

But given this design, I think to prevent future errors, what about adding a comment in the parrallel_t typedef, saying something along the lines of: 

"note: HOMME redefines the parallel_t assignment operator to be a deep copy, so if you add a new element to this struct, please add the element to the user defined assignment operator below"

mt5555 (approved):

ndkeen:
I added comment.

oksanaguba:
Since it was assigned to me, i would ask to run e3sm_integration with this PR on chrysalis and post results here. also, is this bug manifested when we have more procs that dynprocs (homme elements) or in some other case? is it possible to test one of such cases on chrysalis with and without change?

AaronDonahue (approved):

mt5555:
By inspection, this code is clearly a correct bug fix and is 99% likely to pass any test. lets merge asap and not delay with further testing.

oksanaguba:
if there is urgency, i can merge it today. i am waiting for chrysalis integration on cdahs.

ndkeen:
I've been trying to run baseline compares for e3sm_integration on chrysalis, but it's been slow.  I was also hitting some errors (TPUT and one DIFF), so I tried current master and also see some of those same fails.

I *think* this is a situation where the compilers are doing the right thing (or what we want) currently -- ie nothing will be impacted after this change.  But, it makes sense to double-check with some tests.  I also need to test running a highly-scaled case (maybe with ne30) to see if this change impacts when I do change the number of tasks used for ATM physics/dynamics.

This seems like only a boring issue where nvidia/22.7 is being pedantic and helped us find "bug". Note, my next PR after this would then be to increase nvidia compiler version on PM from 22.5 to the machine default of 22.7 (which will fix one other test fail in unrelatd way).

I am happy to merge right away, or wait until PM comes back from scheduled maintenance.

oksanaguba:
@ndkeen thanks for the explanation! as you can see, i unassigned myself. it seems to be an obvious bugfix, but i did not want to be in the position to examine a bunch of unfamiliar code, this is why i asked for tests and 1 particular test with # of procs > # of dynprocs (or whatever else is needed to highlight this issue, like restart?). however, you can decide what's best for this PR and integrate it. thanks!

ndkeen:
Closing this PR in favor of a different solution, which is to just remove the assignment interface https://github.com/E3SM-Project/E3SM/pull/5515

PR #5508, opened 2023-03-06
Increase initial soil moisture in fates runs

JessicaNeedham:
This PR changes the condition for a higher initial value of soil moisture so that all FATES runs, not just FATES-Hydro, will be initialized with higher soil moisture. This is to improve the establishment of forest in bare ground simulations. 
See FATES issue here https://github.com/NGEET/fates/issues/994 and discussion here https://github.com/NGEET/fates/discussions/985. 

This change should be B4B in non-FATES runs and non-B4B in FATES runs.

### DIFF
--- a/components/elm/src/data_types/ColumnDataType.F90
+++ b/components/elm/src/data_types/ColumnDataType.F90
@@ -1616,8 +1616,8 @@ contains
                 if (j > nlevbed) then
                    this%h2osoi_vol(c,j) = 0.0_r8
                 else
-		               if (use_fates_planthydro .or. use_hydrstress) then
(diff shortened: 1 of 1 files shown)


JessicaNeedham:
For discussion on the corresponding CTSM PR see here: https://github.com/ESCOMP/CTSM/pull/1962

glemieux:
Regression testing underway on perlmutter.

glemieux:
All non-fates tests are B4B with the fates tests passing RUN, but not B4B as expected.  Three tests (non-fates) returned computation time increases.

Folder location: `/pscratch/sd/g/glemieux/e3sm-tests/pr5508.fates.pm-cpu.gnu.Cc6f43d321b-F77753581`

rgknox (approved):
Looks good to me! thanks @JessicaNeedham

peterdschwartz:
merged to next

PR #5512, opened 2023-03-07
Revised aerosol process coupling 

wlin7:
This is done through revised handling of cam_in%cflx

In the released v2, the tracer mixing ratio tendencies caused by
    surface fluxes (cam_in%clfx, but excluding the water vapor flux)
    are calculated in tphysac, inside the subroutine clubb_surface.
    These tendencies are applied after clubb_surface before
    the dry removal of aerosols.

    In this commit, an option is added to move (1) the calculation of the
    cflx-induced tendencies and (2) the update of tracer mixing ratios
    from the original place in tphysac
    to tphysbc, right before the cloud macro-microphysics subcycles.

    The new option can be turned on by setting the namelist variable
    cflx_cpl_opt = 2. (Note that this new option does not affect water
    vapor or tracers whose cam_in%cflx is zero when clubb_surface is called.)

[BFB] as a stealth feature
[non-BFB] when the feature is turned on

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -95,6 +95,13 @@ _TESTS = {
             )
         },
 
+    "e3sm_atm_stealth" : {
+        "tests"   : (
+            "ERP_Ln18.ne4_oQU240.F2010.eam-cflx_cpl_2",
+            "SMS_D_Ln5.ne4_oQU240.F2010.eam-cflx_cpl_2",
+            )
+        },
+
     "e3sm_atm_integration" : {
         "inherit" : ("eam_preqx", "eam_theta"),
         "tests" : (
--- a/components/eam/bld/namelist_files/namelist_definition.xml
+++ b/components/eam/bld/namelist_files/namelist_definition.xml
@@ -2748,6 +2748,12 @@ Do radiative transfer and heating/cooling calculation.
 Default: TRUE
 </entry>
 
+<entry id="cflx_cpl_opt" type="integer" category="process_coupling"
+       group="phys_ctl_nl" valid_values="" >
+Numerical scheme for coupling surface tracer fluxes with the rest of model.
+Default: 1 
+</entry>
+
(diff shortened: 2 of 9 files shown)


cime_config/tests.py
```diff
@@ -95,6 +95,13 @@
             )
         },
 
+    "e3sm_atm_stealth" : {
```
rljacob, line 98: Based on the name, will this suite contain tests of all future EAM stealth features?

huiwanpnnl: That's indeed the idea @wlin7 and I discussed. We thought we'd get it started. If there is concern about causing conflicts with PRs of other stealth features, I can rename this set.

huiwanpnnl:
Here is the [Confluence page](https://acme-climate.atlassian.net/wiki/spaces/NGDAP/pages/3684466689/Aerosol+process+coupling+integration) with notes on
- overview of the feature,
- "superBFB" tests,
- HPC tests (links to PACE results),
- component simulations (5-year F2010 simulations with feature on or off; E3SM Diagnostics)

We will perform the B-compset runs after clarifying the requirements with colleagues from the Atmosphere and Coupled groups.

huiwanpnnl (approved):

cime_config/tests.py
rljacob: No that's fine.  But someone has to remember to remove a test from stealth whenever the feature becomes default "on".

wlin7:
@singhbalwinder , just to let you know that I just started merging this PR (checking and resolving conflicts, and performed some basic tests)

As we know, @huiwanpnnl is the developer; ok for me to be the integrator though I appear to be the person creating this PR.

wlin7:
basic tests ok (including F2010, FIDEAL, FSCM-ARM97-MMF1). Merged to next.

wlin7:
Merged to master.

PR #5515, opened 2023-03-09
Remove interface assignment and copy_par subroutine to simplify

ndkeen:
Remove interface assignment and copy_par subroutine as we where not using it anyway.
The nvidia/22.7 compiler was getting confused -- ie, we believe nvidia/22.7 and 22.9 on pm-cpu was doing
something wrong here, but it did point to an issue in the code, which is that because we were not
`include`ing the assignment interface, it was not being used. Which means the default assignment
operation is working fine for this simple data struct and we are confident in removing it.

Note that pm-cpu currently uses nvidia/22.5, which does *not* have this problem. After this PR, I will upgrade to nvidia/22.7.

Unrelated: be more careful about setting initial values for 2 integers in routine.
Example: integer :: ikool=4 is dangerous as we may expect it to be set each entry, but instead, I think it's only set on the first call. Which is fine here as it's only called once, just being more careful.

[bfb]

### DIFF
--- a/components/homme/src/share/parallel_mod.F90
+++ b/components/homme/src/share/parallel_mod.F90
@@ -52,9 +52,6 @@ module parallel_mod
     integer :: root                       ! local root
     integer :: nprocs                     ! number of processes in group
     integer :: comm                       ! local communicator
-!    integer :: node_comm                  ! local communicator of all procs per node
-!    integer :: node_rank                  ! local rank in node_comm
-!    integer :: node_nprocs                ! local rank in node_comm
     logical :: masterproc                
     logical :: dynproc                    ! Designation of a dynamics processor - AaronDonahue
   end type
@@ -69,13 +66,6 @@ module parallel_mod
   ! ===================================================
   ! Module Interfaces
   ! ===================================================
-
-  interface assignment ( = )
-    module procedure copy_par
-  end interface
-
-
-
   public :: initmp
   public :: initmp_from_par
   public :: init_par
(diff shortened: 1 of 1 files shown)


ndkeen:
Using e3sm_integration, I verified that all tests compare BFB with baselines using GNU, as expected.

mt5555 (approved):

ndkeen:
merged to next

ndkeen:
Yep!  I see what I did wrong -- was not careful about where the #ifdefs's lie. I guess I need to revert this. fix the branch, and merge again?

oksanaguba:
i think we are allowed to remerge a branch with a fix (when you have a fix), if we can do it quickly. so no need to revert if you could fix it today.

ndkeen:
The branch is updated, but I am not sure how to "remerge"

oksanaguba:
@ndkeen just merge it again, it should work.

oksanaguba:
However, can you run homme suite before merging?

ndkeen:
I reverted the merge.
I checked out next, then did:
```
git revert -m 1 b1fcc593f393260b0ea13cce7128d30ea2d5532f
git push origin next
```

rljacob:
@ndkeen why did you revert instead of just remerging the feature branch (with the fix).  Unless you don't think you can fix it today before testing runs.  Now you have to "revert the revert" when you're ready to remerge this branch.

ndkeen:
It wasn't clear how to do it and I didn't want any more tests to fail because of this silly build error.
We can just make another PR?

rljacob:
Yes you can make a new PR but also rebase the branch (so the hashes are different).

Issue #5518, opened 2023-03-10
Disable tracer tendency flag broken for RK4 time integrator


cbegeman:
In certain single layer test cases, `config_disable_tr_all_tend = .true.` and `config_time_integrator = 'RK4'`. Under these conditions, the forward model run crashes. The test cases affected are `ocean/dam_break/*cm/default` and `ocean/drying_slope/*m/single_layer/default`. 

[bug][mpas-ocean]

PR #5519, opened 2023-03-10
Ocean bugfix to allow users to disable tracer tendencies with RK4 time integrator

cbegeman:
This PR provides a fix to https://github.com/E3SM-Project/E3SM/issues/5518. In the case where `config_disable_tr_all_tend = .true.` and `config_time_integrator = 'RK4'`:

1. We enforce no time updates to active tracers (https://github.com/E3SM-Project/E3SM/commit/b4e9ec1096a9f33a931ae48be6b7afb7e3b4b306). 
2. We also completely bypass the RK4 tracer tendency update (as opposed to just the update within `ocn_tend_tracer`), which saves a little time (https://github.com/E3SM-Project/E3SM/commit/b4e9ec1096a9f33a931ae48be6b7afb7e3b4b306). 

In principle, only (2) would be needed, but we find that there is contamination of the tracer fields without (1) that leads to model crashes.

We find that after these changes, model runtimes are still ~30% faster when tracers are disabled.

[BFB]

### DIFF
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_rk4.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_rk4.F
@@ -1004,6 +1004,8 @@ module ocn_time_integration_rk4
       integer, pointer :: nEdges, nVertLevels
       err = 0
 
+      if (config_disable_tr_all_tend) return
+
       call mpas_pool_get_config(block % configs, 'config_filter_btr_mode', config_filter_btr_mode)
 
       call mpas_pool_get_subpool(block % structs, 'mesh', meshPool)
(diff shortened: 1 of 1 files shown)


xylar:
@cbegeman I'll do some "do-no-harm" testing in compass.

@sbrus89, could you verify the dam-break fix?

xylar (approved):
The `pr` test suite on Chrysalis with this branch was bit-for-bit with `master`.

cbegeman:
@sbrus89 It seems from your comment https://github.com/E3SM-Project/E3SM/pull/5418#issuecomment-1464508125 that you have already verified that this works with the dam break test. Did you want to do any more testing before approving?

jonbob:
passes:

merged to next

jonbob:
merged to master

PR #5523, opened 2023-03-13
Simplify by removing interface assignment and copy_par subroutine 

ndkeen:
Simplify by removing interface assignment and copy_par subroutine as we where not using it anyway.
The nvidia/22.7 compiler was getting confused -- ie, we believe nvidia/22.7 and 22.9 on pm-cpu was doing
something wrong here, but it did point to an issue in the code, which is that because we were not
including the assignment interface, it was not being used. Which means the default assignment
operation is working fine for this simple data struct and we are confident in removing it.

Note that pm-cpu currently uses nvidia/22.5, which does not have this problem. After this PR, I will upgrade to nvidia/22.7.

Unrelated: be more careful about setting initial values for 2 integers in routine.
Example: integer :: ikool=4 is dangerous as we may expect it to be set each entry, but instead, I think it's only set on the first call. Which is fine here as it's only called once, just being more careful.

[bfb]

### DIFF
--- a/components/homme/src/share/parallel_mod.F90
+++ b/components/homme/src/share/parallel_mod.F90
@@ -52,9 +52,6 @@ module parallel_mod
     integer :: root                       ! local root
     integer :: nprocs                     ! number of processes in group
     integer :: comm                       ! local communicator
-!    integer :: node_comm                  ! local communicator of all procs per node
-!    integer :: node_rank                  ! local rank in node_comm
-!    integer :: node_nprocs                ! local rank in node_comm
     logical :: masterproc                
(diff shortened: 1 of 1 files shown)


ndkeen:
I ran homme_integration on chrysalis and e3sm_integration on pm-cpu.

mt5555 (approved):

ndkeen:
merged to next

ndkeen:
Had some unrelated fails on next last couple of cdash postings. I just manually tested a few of the build fails and all is well.  All other tests that might be impacted by this seem OK, merging.

PR #5527, opened 2023-03-15
Removed ice pressure as a daily average output for V3

proteanplanet:
Removes the field of icePressure from daily output fields for V3, which is still included in monthly averages in the sea ice model history. No one appears to have analyzed this daily field for V2, it can still be switched on for particular simulations, and removing it from the default output will roughly reduce sea ice history output by 12%. The change has been tested in a D-case, which compiles and runs successfully.

[BFB]


eclare108213 (approved):
approved by visual inspection

jonbob:
passes:

though may cause DIFFs due to "FIELDLIST field lists differ (otherwise bit-for-bit)"

merged to next

jonbob:
merged to master and blessed DIFFs on platforms with cprnc versions that report field list changes as DIFFs (chrysalis and mappy)

PR #5528, opened 2023-03-15
Move batching directives to the top of batching templates

amametjanov:
This is needed for machines with Cobalt job scheduler, which needs `#COBALT -n {{ num_nodes }}` right after the first line of the batch script.


jgfouca:
Thanks @amametjanov !

dqwu (approved):
Tested on JLSE nodes with Cobalt job scheduler. Running case.submit looks good.

Issue #5351, opened 2022-12-06
Build error (error: mixed vector-scalar operation not allowed) with nvidia on pm-cpu -- MMF test


ndkeen:
With master of Dec6th, this test fails to build with nvidia compiler (version 22.5):
`ERS_Ln9.ne4pg2_ne4pg2.F2010-MMF1.pm-cpu_nvidia.eam-mmf_crmout`

```
"/global/cfs/cdirs/e3sm/ndk/repos/nexty-dec2/externals/YAKL/external/pocketfft_hdronly.h", line 1977: error: mixed vector-scalar operation not allowed
      T tr3=2*CC(ido-1,1,k);
             ^
          detected during:
            instantiation of "void pocketfft::detail::rfftp<T0>::radb4(std::size_t, std::size_t, const T *, T *, const T0 *) const [with T0=real, T=__attribute((vector_size(32UL))) double]" at line 2256
            instantiation of "void pocketfft::detail::rfftp<T0>::exec(T *, T0, bool) const [with T0=real, T=__attribute((vector_size(32UL))) double]" at line 2525
            instantiation of "void pocketfft::detail::pocketfft_r<T0>::exec(T *, T0, bool) const [with T0=real, T=__attribute((vector_size(32UL))) double]" at line 3255
            instantiation of "void pocketfft::detail::general_r2c(const pocketfft::detail::cndarr<T> &, pocketfft::detail::ndarr<pocketfft::detail::cmplx<T>> &, std::size_t, bool, T, std::size_t) [with T=real]" at line 344$
            instantiation of "void pocketfft::detail::r2c(const pocketfft::detail::shape_t &, const pocketfft::detail::stride_t &, const pocketfft::detail::stride_t &, std::size_t, bool, const T *, std::complex<T> *, T, st$
            instantiation of "void yakl::RealFFT1D<T>::forward_real(yakl::Array<T, N, 1, 1> &, int, int) [with T=real, N=4]" at line 24 of "/global/cfs/cdirs/e3sm/ndk/repos/nexty-dec2/components/eam/src/physics/crm/samxx/c$

```

Other MMF tests also have same build fail.

I think these errors might not happen with version 22.7 of nvidia compiler, however, that version has some other runtime issues.  https://github.com/E3SM-Project/E3SM/issues/5353

ndkeen:
If I change a few constants in that file, I can get the test to build with nvidia.

ie, changes like:
```
[...]

PR #5529, opened 2023-03-16
For pm-cpu,pm-gpu,alvarez, update to the default nvidia 22.7 compiler

ndkeen:
For pm-cpu,pm-gpu,alvarez, update to the default nvidia 22.7 compiler version from 22.5.

Fixes https://github.com/E3SM-Project/E3SM/issues/5351

Will not be bfb.


rljacob (approved):

ndkeen:
I tested with this merged into next and I see expected results.  Most will DIFF with baselines.

ndkeen:
merged to next

Issue #5525, opened 2023-03-14
For Perlmutter (pm-cpu/pm-cpu), update module versions to defaults


ndkeen:
Wanting to update some modules to a minor version change on PM. AFAIK, This does not fix any issues or seem to cause any problems.

Specifically, from these:
```
        <command name="load">cray-mpich/8.1.22</command>
        <command name="load">cray-hdf5-parallel/1.12.2.1</command>
        <command name="load">cray-netcdf-hdf5parallel/4.9.0.1</command>
        <command name="load">cray-parallel-netcdf/1.12.3.1</command>
```
to these:
```
        <command name="load">cray-mpich/8.1.24</command>
        <command name="load">cray-hdf5-parallel/1.12.2.3</command>                                                                                                                               
        <command name="load">cray-netcdf-hdf5parallel/4.9.0.3</command>                                                                                                                             
        <command name="load">cray-parallel-netcdf/1.12.3.3</command>
```

I've tested e3sm_integration (comparing with baselines) as well as the nightly scream tests and am testing with longer/larger scream tests. These 2 MMF tests pass: `SMS_Ln3.ne4pg2_ne4pg2.F2010-MMF1.pm-cpu_gnu
SMS_Ln3.ne4pg2_ne4pg2.F2010-MMF1.pm-gpu_gnu`

Not updating cudatoolkit, where discussed here https://github.com/E3SM-Project/E3SM/issues/5314

PR #5533, opened 2023-03-17
For pm-cpu/pm-gpu, update some module versions to current machine defaults

ndkeen:
Minor version increases for several modules.
Does not impact compiler versions.
Motivation is to keep up-to-date with machine defaults.
Do not see any measurable performance changes.

```
cray-mpich/8.1.22 -> cray-mpich/8.1.24
cray-hdf5-parallel/1.12.2.1 -> cray-hdf5-parallel/1.12.2.3
cray-netcdf-hdf5parallel/4.9.0.1 -> cray-netcdf-hdf5parallel/4.9.0.3
cray-parallel-netcdf/1.12.3.1 -> cray-parallel-netcdf/1.12.3.3
cmake/3.22.0 -> cmake/3.24.3
```

Added specific version numbers for `craype` and `cray-libsci` to reduce surprises when the default version is changed (these were already using default)

Also added a couple of modules to remove just in case they are loaded.

Updating alvarez the same way, but it may be that the machine goes away.

Fixes https://github.com/E3SM-Project/E3SM/issues/5525
[bfb]

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -195,6 +195,8 @@
         <command name="unload">cray-parallel-netcdf</command>
         <command name="unload">PrgEnv-gnu</command>
         <command name="unload">PrgEnv-nvidia</command>
+        <command name="unload">PrgEnv-cray</command>
+        <command name="unload">PrgEnv-aocc</command>
         <command name="unload">cudatoolkit</command>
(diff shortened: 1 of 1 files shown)


ndkeen:
I tested against baselines on pm-cpu with e3sm_integration.  And have been running larger cases with these modules versions in a scream repo (to test pm-gpu).

rljacob (approved):

PR #5538, opened 2023-03-21
WIP: output 3D radiation fluxes correctly in RRTMG

mahf708:
This PR corrects outputting 3D radiation fluxes in the RRTMG interface.

`FUS`, `FDS`, `FUSC`, and `FDSC` (and their longwave equivalents) are cosmetic only, and they can only be zeros. It is possible they could be calculated, but no attempt at doing so can be found here. Instead, readily available _net_ fields are outputted instead. These are `FNS` and `FNSC` (and their longwave friends).

Thanks to @whannah1 and @brhillman for their help in continuing an investigation of issues in the radiation code.

@brhillman, not sure how much work is need to actually get the up and down components, but this is likely a moot point (viz. we should likely be moving to RRTMGP) as we discussed privately.

---

For testing, a simple one-month run (`SMS_Lm1.ne4_oQU480.F2010`) is conducted to confirm this change is working as intended: `FUS` and friends remain filled zeros everywhere while `FNS` and friends are populated as expected. The minimal test could be found under this users's group directory on chrysalis: 

Commit: 466c963

```
... scratch/chrys/SMS_Lm1.ne4_oQU480.F2010.chrysalis_intel.radout_pr
```

Commit: [f96c891](https://github.com/E3SM-Project/E3SM/pull/5538/commits/f96c89182cee40ef4bdf3a3c3a66c1ff6d198076)

```
... scratch/chrys/SMS_Lm1.ne4_oQU480.F2010.chrysalis_intel.radout_pr_4
```

---

(I am new here and I don't know procedures, so any feedback is appreciated!)

### DIFF
--- a/components/eam/src/physics/rrtmg/radsw.F90
+++ b/components/eam/src/physics/rrtmg/radsw.F90
@@ -318,6 +318,12 @@ subroutine rad_rrtmg_sw(lchnk,ncol       ,rrtmg_levs   ,r_state      , &
    qrsc(1:ncol,1:pver) = 0.0_r8
    fns(1:ncol,1:pverp) = 0.0_r8
    fcns(1:ncol,1:pverp) = 0.0_r8
+
+   ofus(1:ncol,1:pverp) = 0.0_r8
+   ofds(1:ncol,1:pverp) = 0.0_r8
+   ofusc(1:ncol,1:pverp) = 0.0_r8
+   ofdsc(1:ncol,1:pverp) = 0.0_r8
+
    if (single_column.and.scm_crm_mode) then 
       fus(1:ncol,1:pverp) = 0.0_r8
       fds(1:ncol,1:pverp) = 0.0_r8


brhillman:
It would actually be easy to properly output the up and down fluxes as well. Line 588 in `radsw.F90` is where the net flux is computed from the up and down components, and then the day-only fluxes get expanded back to all columns on line 621. To output the up and down fluxes, we would just need to add the extra output fields to the subroutine parameter lists, handle them properly in `radsw.F90`, and then pass them out of that routine and add `outfld` calls for them in `radiation.F90`.

components/eam/src/physics/rrtmg/radiation.F90
```diff
@@ -734,6 +740,12 @@ subroutine radiation_init(phys_state, pbuf)
                       sampling_seq='rad_lwsw', flag_xyfill=.true.)
           call addfld('FDLC'//diag(icall), (/ 'ilev' /),'I',    'W/m2', 'Longwave clear-sky downward flux', &
                       sampling_seq='rad_lwsw', flag_xyfill=.true.)
+          ! < mahf708 2023-03-21 > FNL and FNLC can be outputted:
```
whannah1, line 743: Don't add code "tags" like this (keep the descriptive comments though). Git preserves the history we need to know who did what, and when. This style of updating the code with usernames ends up adding a lot of clutter, so we actively discourage it now.

whannah1 (changes requested):
remove code user tags, otherwise it looks ok

components/eam/src/physics/rrtmg/radsw.F90
```diff
@@ -591,6 +601,10 @@ subroutine rad_rrtmg_sw(lchnk,ncol       ,rrtmg_levs   ,r_state      , &
    fusc(1:Nday,pverp-rrtmg_levs+1:pverp) = swuflxc(1:Nday,rrtmg_levs:1:-1)
    fds (1:Nday,pverp-rrtmg_levs+1:pverp) =  swdflx(1:Nday,rrtmg_levs:1:-1)
    fdsc(1:Nday,pverp-rrtmg_levs+1:pverp) = swdflxc(1:Nday,rrtmg_levs:1:-1)
+   ofus (1:Nday,pverp-rrtmg_levs+1:pverp)=  swuflx(1:Nday,rrtmg_levs:1:-1)
```
whannah1, line 604: echoing slack convo => these values need to be initialized so that nighttime columns are zero

mzelinka:
Just wondering if these changes were ever implemented.  I'd like to write out the up- and down-welling LW and SW fluxes in some runs.

PR #5541, opened 2023-03-21
Update bld files to catch missed mpas-o Registry change

jonbob:
A previous PR, #5202, brought in a new config in the mpas-ocean Registry file, but the bld files were not updated to match. This brings the E3SM bld files up-to-date with the Registry

[NML]
[BFB]


xylar (approved):
The changes adding `config_output_reference_time` look exactly correct to me.  Thanks @jonbob!

Everything else looks like formatting changes or documentation.

jonbob:
merged to master and expected NML DIFFs blessed for chrysalis. No other machine has reported today, so more bless requests will be required

jonbob:
blessed NML DIFFs on perlmutter

PR #5545, opened 2023-03-22
Move from mpas-cice to mpas-seaice in inputdata

xylar:
This merge points all instances of `inputdata/ice/mpas-cice` instead to `inputdata/ice/mpas-seaice`.

I have rsynced all files in `inputdata/ice/mpas-cice` that were not already in `inputdata/ice/mpas-seaice` into that directory this morning.

I propose to move `mpas-cice` to `deprecated-mpas-cice` and make `mpas-cice` a symlink to `mpas-seaice` instead.  This solution will be fully backwards compatible.  However, we will retain the `deprecated-mpas-cice` for a few months just in case.

[BFB]

### DIFF
--- a/components/mpas-seaice/bld/build-namelist-section
+++ b/components/mpas-seaice/bld/build-namelist-section
@@ -28,7 +28,7 @@ add_default($nl, 'config_full_abort_write');
 # Namelist group: decomposition #
 #################################
 
-add_default($nl, 'config_block_decomp_file_prefix', 'val'=>"'${DIN_LOC_ROOT}/ice/mpas-cice/${ICE_GRID}/mpas-cice.graph.info.${date_stamp}.part.'");
+add_default($nl, 'config_block_decomp_file_prefix', 'val'=>"'${DIN_LOC_ROOT}/ice/mpas-seaice/${ICE_GRID}/mpas-seaice.graph.info.${date_stamp}.part.'");
 add_default($nl, 'config_number_of_blocks');
 add_default($nl, 'config_explicit_proc_decomp');
 add_default($nl, 'config_proc_decomp_file_prefix');
--- a/components/mpas-seaice/bld/build-namelist
+++ b/components/mpas-seaice/bld/build-namelist
@@ -13,7 +13,7 @@
 #
 # The simplest use of build-namelist is to execute it from the build directory where configure
 # was run.  By default it will use the config_cache.xml file that was written by configure to
-# determine the build time properties of the executable, and will write the files that contain 
+# determine the build time properties of the executable, and will write the files that contain
 # the output namelists in that same directory.
 #
 #
@@ -35,14 +35,14 @@ SYNOPSIS
      build-namelist [options]
 OPTIONS
      -infile "filepath"    Specify a file containing namelists to read values from.
-     -namelist "namelist"  Specify namelist settings directly on the commandline by supplying 
+     -namelist "namelist"  Specify namelist settings directly on the commandline by supplying
                            a string containing FORTRAN namelist syntax, e.g.,
                               -namelist "&mpas-seaice_nml dt=1800 /"
      -help [or -h]         Print usage to STDOUT.
      -test                 Enable checking that input datasets exist on local filesystem.
      -verbose              Turn on verbose echoing of informational messages.
      -caseroot             CASEROOT directory variable
-     -casebuild            CASEBUILD directory variable          
+     -casebuild            CASEBUILD directory variable
      -cimeroot             CIMEROOT directory variable
      -ice_grid             ICE_GRID variable
      -decomp_prefix        decomp_prefix variable
@@ -67,7 +67,7 @@ NOTE: The precedence for setting the values of namelist variables is (highest to
       1. namelist values set by specific command-line options, i.e. (none right now)
       2. values set on the command-line using the -namelist option,
       3. values read from the file specified by -infile,
-      4. values from the namelist defaults file - or values specifically set in build-namelist 
+      4. values from the namelist defaults file - or values specifically set in build-namelist
 EOF
 }
 
@@ -82,7 +82,7 @@ my $ProgDir = $1;                      # name of directory containing this scrip
                                        # the user's PATH
 my $cwd = getcwd();                    # current working directory
 my $cfgdir;                            # absolute pathname of directory that contains this script
-if ($ProgDir) { 
+if ($ProgDir) {
     $cfgdir = absolute_path($ProgDir);
 } else {
     $cfgdir = $cwd;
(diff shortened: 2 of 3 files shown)


xylar:
@rljacob, @eclare108213, @proteanplanet, @jonbob, @darincomeau and @akturner, I would very much appreciate your feedback on these changes.

@jonbob, @darincomeau, @akturner and I along with the rest of the MPAS DevOps team discussed this yesterday and there was a feeling that this was the right way forward.  We had discussed making symlinks to the individual mesh directories in `mpas-cice` but I think making `mpas-cice` is a cleaner solution.  (It can be made in 1 or 2 commands.)

rljacob:
Yes making mpas-cice a link to mpas-seaice should work.

components/mpas-seaice/bld/build-namelist-section
```diff
@@ -28,7 +28,7 @@ add_default($nl, 'config_full_abort_write');
 # Namelist group: decomposition #
 #################################
 
-add_default($nl, 'config_block_decomp_file_prefix', 'val'=>"'${DIN_LOC_ROOT}/ice/mpas-cice/${ICE_GRID}/mpas-cice.graph.info.${date_stamp}.part.'");
+add_default($nl, 'config_block_decomp_file_prefix', 'val'=>"'${DIN_LOC_ROOT}/ice/mpas-seaice/${ICE_GRID}/mpas-seaice.graph.info.${date_stamp}.part.'");
```
rljacob, line 31: We'll have to keep the old graph files with "cice" in the name.  Or make links for each individual one to the the new name.

rljacob (approved):

eclare108213 (approved):

components/mpas-seaice/bld/build-namelist-section
xylar: Couldn't we override this default for all meshes that have `mpas-cice` in their graph file names?  But I agree that there's work still to be done.

xylar: I think we already do override this value for all meshes we support:
https://github.com/E3SM-Project/E3SM/blob/614358918ce3b79949bfc281ad8c8848661799e6/components/mpas-seaice/cime_config/buildnml#L75-L264

I must admit, though, that I don't understand what `build-namelist-section` does as opposed to `build-namelist`.

@jonbob, maybe you can give me some background either here or on slack?

xylar: Right, we wouldn't rename any of the existing partition files.  They remain as `mpas-cice` in the list above form `buildnml`.  I just want to change the default name to be `mpas-seaice.graph.info.` for consistency. As far as I can tell, this default never gets used since all meshes (now and in the past) explicitly specify this prefix.

xylar: @rljacob, I discussed this with @jonbob and this file isn't used directly for anything.  It is the automated output of a process used to update `build-namelist`, which is then manually combined with some customization (I as will be painfully clear, I'm fuzzy on the details) to get the final `build-namelist`.  So long story short, it doesn't matter what value is here from my understanding.

proteanplanet (approved):
I used the new E3SM-Polar-Developer.sh script (https://github.com/E3SM-Project/E3SM/pull/5551) to confirm that this branch was able to successfully acquire the data, and run a 3-month D-case from cold startup. Approved.

akturner (approved):

darincomeau:
@xylar I tried running a smoke test on Anvil, and it's not able to find the partition files. LCRC points to this inputdata directory:
```
/lcrc/group/e3sm/data/inputdata
```
and all of the new file/directory structure are (rightly) staged in the public inputdata directory:
```
/lcrc/group/e3sm/public_html/inputdata/
```

When an Anvil/Chrysalis run tries to find a missing file in the first directory from the second, it can't:
```
wget failed with output:  and errput /lcrc/group/e3sm/data/inputdata/ice/mpas-seaice/EC30to60E2r2/mpas-seaice.graph.info.200908.part.144: Permission denied

        svn export failed with output:  and errput svn: E170000: URL 'https://svn-ccsm-inputdata.cgd.ucar.edu/trunk/inputdata/ice/mpas-seaice/EC30to60E2r2/mpas-seaice.graph.info.200908.part.144' doesn't exist

        ERROR: Could not find all inputdata on any server
```

The file permissions I see for the partition files in the new `mpas-seaice` directory are:
```
-rw-rw-r-- 1 ac.xasay-davis E3SM
```

If you add a group `x` permission, it should be able to `wget` (right?).

proteanplanet:
@darincomeau, @xylar  Just to mention that the smoke test I ran successfully (above) was on Chrysalis.

xylar:
@darincomeau, no, someone has been adding execute permission to a lot of files that don't need it.  Was that you?  I don't think that's a good idea.  Adding execute permission to files that aren't executable really allows users to try to run them, which can be bad.  Only directories need execute permission.  I thought I had done that carefully but I'll check.

xylar:
@darincomeau and @proteanplanet, the issue is not in `public_html`.  The issue is that @proteanplanet presumably created the `/lcrc/group/e3sm/data/inputdata/ice/mpas-seaice/` by running his tests but the permissions don't include group write permission:
```
$ pwd
/lcrc/group/e3sm/data/inputdata/ice
[ac.xasay-davis@chrlogin1 ice]$ ls -lah mpas-seaice/
total 4.0K
drwxr-sr-x 4 ac.afroberts E3SM 4.0K Mar 23 16:55 .
drwxrwsr-x 8 sarich       E3SM 4.0K Mar 23 16:55 ..
drwxr-sr-x 2 ac.afroberts E3SM 4.0K Mar 23 17:58 EC30to60E2r2
drwxr-sr-x 2 ac.afroberts E3SM 4.0K Mar 23 16:55 general
```
That would seem to be a problem with E3SM not changing permissions appropriately for input data when it gets downloaded.  @rljacob, any insight or anyone you can point us to?

Also, why doesn't Chrysalis just use the `inputdata` directly out of `public_html/inputdata`?

darincomeau:
@xylar yes you're right, it's the default permissions on the directories created by the downloading missing inputdata that is the source of the problem. Sorry, I should have looked a little more carefully and known that. And you're right on `x` for files that aren't executables; I usually set 664 on files, and 775 on directories, and stage them in both places.

>Also, why doesn't Chrysalis just use the inputdata directly out of public_html/inputdata?

I've also wondered this.

The permissions issue is clearly not part of this PR, so I'm approving - thanks!

darincomeau (approved):
Approved based on smoke testing and seeing correct paths to partition files.

jonbob (approved):
approved based on E3SM testing

jonbob:
Passes:

but fails with DIFFs:

examination shows there is an incorrect file:
````
/lcrc/group/e3sm/data/inputdata/ice/mpas-cice/oEC60to30v3/seaice.EC60to30v3.161222.nc
````
that is different from the version of this file on the inputdata repo. I copied the repo version into this location but we should still expect this one test to DIFF.

merged to next

jonbob:
merged to master and expected DIFFs on lcrc blessed. Also missing graph files for ERP and PEM case2 runs staged by hand on perlmutter and lcrc local inputdata

Issue #5276, opened 2022-11-06
improving accuracy of shr_reprosum_calc


worleyph:
The reprosum algorithm converts each floating point summand into a
vector of integers representation. These vectors are summed locally
first, and then globally (using an mpi_allreduce). All of this is
exact and reproducible. The final step is to convert the final sum, in
the vector of integers representation, back into a floating point
value.

For this last step, each component integer in the vector
representation is converted back into a floating point value (call it
a_{i} for the ith element of the vector), and then these values are
summed locally, smallest to largest, to get the final value. The
integer representation is first massaged so that all a_{i} have the
same sign and the range of values is usually not overlapping (least
significant binary digit in abs(a_{i}) represents a value larger than
that represented by abs(a_{i+1}) in most cases, but sometimes is equal
to).

As this final floating point sum is done serially on each process, the
full summation is reproducible. However, despite the care in
constructing the final floating point value, round-off error can be
introduced in the least significant digit (apparently in the least
significant binary digit). This level of inaccuracy has no practical
impact on overall model accuracy, but it can lead to some annoying
inconsistencies, for example, with the calculation using a 4-byte
integer vector representation potentially producing a slightly
different result from a calculation using an 8-byte integer vector
representation, even though everything else is the same.

One modification to decrease the likelihood of rounding errors is to
calculate this final sum of floating point values using the DDPDD
algorithm locally. (DDPDD is an alternative algorithm already available
in the reprosum code.) DDPDD uses a two-floating-point-value
representation of the intermediate values in the summation to
approximately double the precision, i.e., for double precision values,
summing in quad precision. In this case, round-off error is isolated
to the least significant digits is in the quad precision value, and it
is even less likely that this will impact the least significant digit
in the resulting double precision value (though not impossible).
(Note that the integer representation can also be made strictly
nonoverlapping prior to conversion to floating point values, but this
does not seem to change the numerics.)

This modification requires little code change, and only a very small
increase in the cost of the reprosum summations. The downside of this
change is that it is not BFB with the current algorithm in
practice in E3SM. For example, for a 5 day run on Chrysalis using

--compset WCYCL1850 --res ne30pg2_EC30to60E2r2 --pecount XS

(675 ATM processes) there are 42539 reprosum calls, all in ATM, and
using DDPDD for the final sum changes the result 154 times (so 0.36%
of the time).

Looking at the actual difference, it is always +/- 1 in the least
significant binary digit in the mantissa. But this is still enough to
[...]

PR #5549, opened 2023-03-23
Fix support for preventing overflow and catastrophic cancellation

### LINKED ISSUE #5537: Faults in logic to prevent overflow in shr_reprosum_mod
The default shr_reprosum_mod algorithm, the integer vector algorithm,
first converts each floating point summand into an equivalent
representation using a vector of integers, sums the integer vectors,
then converts the resulting sum back to a floating point
representation. Each integer*8 component IX(i), i=1,... , has an
associated implicit exponent LX(i) and represents the floating point
value real(IX(i),r8)**LX(i). The sequence of exponents all differ by
the same amount, LX(i) - LX(i+1) = max_shift, and are monotonically
decreasing in (i), and each component for a given summand is strictly
less than radix(1_i8)**arr_max_shift in absolute value.

To allow for a larger max_shift (and shorter integer vector), the
components of the integer vector are manipulated after the local
single threaded sums and before the sum over OpenMP threads and MPI
tasks so that
 abs(IX(i)) < radix(1_i8)**max_shift
for all i > 0, as follows. If abs(IX(i)) is larger than or equal to
(radix(1_i8)**arr_max_shift), subtract this 'overlap' from the current
value and add it (appropriately shifted) to IX(i-1). By starting from
the largest index (smallest implicit exponent), this removes any
overlap from each component. Note that the integer vector has an IX(0)
that is initially zero to capture the overlap from IX(1).

The current logic stops the process at index 0. However, this can be
greater than the target radix(1_i8)**max_shift, and the subsequent
summation over MPI tasks and OpenMP threads can potentially overflow
for the distributed sum of IX(0) if the numbers of OpenMP threads and
MPI tasks are large. There is no evidence that this has ever occurred in
an E3SM simulation, but this flaw should be addressed.

Another weakness in the code arises from using

 RX_8 = real(IX(i),r8)
 Overlap = int(scale(RX_8,-max_shift),i8)

to identify the overlap in IX(i), which may miss some of the overlap
if IX(i) has too many significant digits to be represented by a
floating point value. While this is unlikely to lead to overflow even
if it does occur, it is still a flaw (and is easily corrected).

Both of the above issues have been addressed in a planned PR, and
this issue is submitted for future reference by that PR.

worleyph:
The default shr_reprosum_mod algorithm, the integer vector algorithm,
first converts each floating point summand into an equivalent
representation using a vector of integers, sums the integer vectors,
then converts the resulting sum back to a floating point
representation. 

The algorithm includes logic to preclude the possibility of integer
overflow in the intermediate sums and to eliminate the possibility
of catastrophic cancellation in the final summation. Upon close
inspection it has been determined that this logic is not sufficient
to  achieve these goals in all situations. Though it is very unlikely
for these "edge cases" to occur during E3SM simulations, it
is desirable to correct the problems anyway.

a) Fix logic to make all integers in integer expansion the same sign.

As part of the process of converting the vector of integers
representing the global sum into a floating point value, the
components are first modified to all be the same sign. This eliminates
the possibility of catatrophic cancellation during the next step, when
each integer component is converted to a floating point value and
these floating point values are then summed to generate the global
sum.

The algorithm works as follows. Each integer*8 component IX(i),
i=0,1,... , has an associated implicit exponent LX(i) and represents
the floating point value real(IX(i),r8)**LX(i). The sequence of
exponents all differ by the same amount, LX(i) - LX(i+1) = max_shift,
and are monotonically decreasing in (i). By construction,
 abs(IX(i)) < radix(1_i8)**max_shift
for i = 1,2,... In consequence, if IX(i-1), IX(i), and IX(i+1) are
all nonzero, then the magnitude of the value at index (i)
  (abs(real(IX(i),r8)**LX(i)))
is strictly less than the magnitude of the value at index (i-1)
  (abs(real(IX(i-1),r8)**LX(i-1)))
and strictly greater than the magnitude of the value at index (i+1)
  (abs(real(IX(i+1),r8)**LX(i+1))).
If IX(i) is greater than zero and IX(i+1) is less than zero,
then 1_i8 can be subtracted from IX(i) and radix(1_i8)**max_shift
added to IX(i+1) so that IX(i) is now greater than or equal to zero,
and IX(i+1) is greater than zero. If IX(i) is less than zero and
IX(i+1) is greater than zero, then, after adding 1_i8 to IX(i) and
subtracting radix(1_i8)**max_shift from IX(i+1), IX(i) is less than or
equal to zero and IX(i+1) is less than zero. The algorithm begins by
finding the first nonzero component (i=0,1,...), which determines the
sign of the resulting sum, then uses the above process to change all
components to this sign.

The above description does not treat the situation when IX(i) is zero,
which is where an error arises in the current algorithm. The current
algorithm always treats 0_i8 as positive (as it uses the fortran 'sign'
intrinsic to compare signs between components). This can cause
problems if IX(i-1) is greater than zero, IX(i) is zero, and IX(i+1)
is less than zero. In this case IX(i-1) and IX(i) have the
same sign and neither are modified after their comparison. However,
IX(i) and IX(i+1) have different signs, so 1_i8 will be subtracted
from IX(i), changing it to negative, while radix(1_i8)**max_shift will
be added to IX(i+1), changing it to positive, and the same sign
property will not be achieved. The solution is to always treat a zero
IX(i) as having a different sign from IX(i-1) so that IX(i) always
becomes nonzero. As we always start the process with a nonzero IX(i),
we end up with all components either greater than or equal to zero or
all components less than or equal to zero.

Note that while this error can prevent the resulting sequence of
floating point values from being all the same sign or zero, the
situation in which this error occurs does not lead to additional
rounding error in the final summation and so does not decrease the
accuracy. Also, the fix has been BFB with the buggy version in
practice, though it probably is not guaranteed to be.

Despite the innocuousness of the error, it is important to fix it as
part of the resolution to issue #5276 ("Improving accuracy of
shr_reprosum_calc").

Claiming that this is BFB, for the purposes of testing (as it
likely is BFB for E3SM testing and production simulations) so that any
non-BFB behavior in the testing will be examined before the PR is
merged into master.

b) Fix logic to prevent overflow in intermediate sums

To reiterate, each integer*8 component IX(i), i=1,... , has an
associated implicit exponent LX(i) and represents the floating point
value real(IX(i),r8)**LX(i). The sequence of exponents all differ by
the same amount, LX(i) - LX(i+1) = max_shift, and are monotonically
decreasing in (i), and each component for a given summand is strictly
less than radix(1_i8)**max_shift in absolute value.

To allow for a larger max_shift (and shorter integer vector), the
components of the integer vector are manipulated after the local
single threaded sums and before the sum over OpenMP threads and MPI
tasks so that
 abs(IX(i)) < radix(1_i8)**max_shift
for all i > 0, as follows. If abs(IX(i)) is larger than or equal to
(radix(1_i8)**arr_max_shift), subtract this 'overlap' from the current
value and add it (appropriately shifted) to IX(i-1). By starting from
the largest index (smallest implicit exponent), this removes any
overlap from each component. Note that the integer vector has an IX(0)
that is initially zero to capture the overlap from IX(1).

The current logic stops the process at index 0. However, this can be
greater than the target radix(1_i8)**max_shift, and the subsequent
summation over MPI tasks and OpenMP threads can potentially overflow
for the distributed sum of IX(0) if the number of OpenMP threads times
the number of MPI tasks is large. By extending the length of the
integer vector from 0:max_levels to -(extra_levels-1):max_levels and
continuing the above process to index -(extra_levels-1), can guarantee
that the integer value for all indices satisfies abs(IX(i)) <
radix(1_i8)**max_shift if choose extra_levels appropriately, thus
eliminating the possibility of overflow.

Another weakness in the original algorithm arises from using

 RX_8 = real(IX(i),r8)
 Overlap = int(scale(RX_8,-max_shift),i8)

to identify the overlap in IX(i), which may miss some of the overlap
if IX(i) has too many significant digits to be represented by a
floating point value. However, simply calculating the overlap using
integer division

 Overlap = IX(i) / (i8_radix**max_shift)

eliminates this (slight) possibility, and also simplifies the code.

This commit implements both fixes. It also augments the comments to
better document why this approach is sufficient to avoid overlap
during the sum over threads and MPI tasks.

Note that, while these changes are not guaranteed to be BFB, the
calculated value for extra_levels for all E3SM simulations in the test
suites (and likely all production simulations) is almost certainly
'1', and so will be identical to the existing algorithm. The change
in the overlap calculation is also likely to be BFB with the current
algorithm in practice for this code fragment and, even if not, is
still unlikely to change the value of the final sum. Again, claiming
that it is BFB for the purposes of this PR so that any non-BFB
behavior in the testing will be examined before the PR is accepted.

Fixes E3SM-Project/E3SM#5537

[BFB]

### DIFF
--- a/share/util/shr_reprosum_mod.F90
+++ b/share/util/shr_reprosum_mod.F90
@@ -561,6 +561,9 @@ module shr_reprosum_mod
       integer :: max_levels(nflds)       ! maximum number of levels of
                                          !  integer expansion to use
       integer :: max_level               ! maximum value in max_levels
+      integer :: extra_levels            ! number of extra levels needed
+                                         !  to guarantee that sum over threads
+                                         !  or tasks does not cause overflow
 
       real(r8) :: xmax_nsummands         ! real(max_nsummands,r8)
       real(r8) :: arr_lsum(nflds)        ! local sums
@@ -736,15 +739,16 @@ module shr_reprosum_mod
                endif
                call t_stopf("repro_sum_allr_max")
 
-! Determine maximum shift. Shift needs to be small enough that summation
-! does not exceed maximum number of digits in i8.
+! Determine maximum shift. Shift needs to be small enough that summation,
+! in absolute value, does not exceed maximum value representable by i8.
 
 ! If requested, return max_nsummands before it is redefined
                if ( present( gbl_max_nsummands_out) ) then
                   gbl_max_nsummands_out = max_nsummands
                endif
 
-! Summing within each thread first
+! Summing within each thread first (adding 1 to max_nsummands
+! to ensure that integer division rounds up)
                max_nsummands = (max_nsummands/omp_nthreads) + 1
 ! then over threads and tasks
                max_nsummands = max(max_nsummands, tasks*omp_nthreads)
@@ -762,6 +766,25 @@ module shr_reprosum_mod
                   call shr_sys_abort('repro_sum failed: number of summands too '// &
                                      'large for integer vector algorithm' )
                endif
+! Note: by construction, each floating point value will be decomposed
+! into a vector of integers each component of which will be strictly
+! less than radix(1_i8)**arr_max_shift in absolute value, and the
+! summation of max_nsummands of these, again in absolute value, will
+! then be less than 
+!  radix(1_i8)**(arr_max_shift + exponent(xmax_nsummands))
+! or radix(1_i8)**(digits(1_i8) - 1). This is more conservative than
+! necessary, but it also allows the postprocessing mentioned above
+! (and described later) to proceed without danger of introducing
+! overflow. 
+
+! Determine additional number of levels needed to support the
+! postprocessing that reduces the magnitude of each component
+! of the integer vector of the partial sum for each thread
+! to be less than (radix(1_i8)**arr_max_shift).
+               extra_levels = (digits(1_i8) - 1)/arr_max_shift
+! Extra levels are indexed by (-(extra_levels-1):0)
+! Derivation of this is described in the comments in
+! shr_reprosum_int.
 
 ! Calculate sum
                if (present(repro_sum_validate)) then
@@ -771,8 +794,9 @@ module shr_reprosum_mod
                endif
                call shr_reprosum_int(arr, arr_gsum, nsummands, dsummands, &
                                      nflds, arr_max_shift, arr_gmax_exp, &
-                                     arr_max_levels, max_level, arr_gsum_infnan, &
-                                     validate, recompute, omp_nthreads, mpi_comm)
+                                     arr_max_levels, max_level, extra_levels, &
+                                     arr_gsum_infnan, validate, recompute, &
+                                     omp_nthreads, mpi_comm)
 
 ! Record statistics, etc.
                repro_sum_fast = 1
@@ -873,10 +897,11 @@ module shr_reprosum_mod
             endif
 
 ! Determine maximum shift (same as in previous branch, but with calculated
-! max_nsummands). Shift needs to be small enough that summation does not
-! exceed maximum number of digits in i8.
+! max_nsummands). Shift needs to be small enough that summation, in absolute
+! value, does not exceed maximum value representable by i8.
 
-! Summing within each thread first
+! Summing within each thread first (adding 1 to max_nsummands
+! to ensure that integer division rounds up)
             max_nsummands = (max_nsummands/omp_nthreads) + 1
 ! then over threads and tasks
             max_nsummands = max(max_nsummands, tasks*omp_nthreads)
@@ -894,6 +919,16 @@ module shr_reprosum_mod
                call shr_sys_abort('repro_sum failed: number of summands too '// &
                                   'large for integer vector algorithm' )
             endif
+! Note: by construction, each floating point value will be decomposed
+! into a vector of integers each component of which will be strictly
+! less than radix(1_i8)**arr_max_shift in absolute value, and the
+! summation of max_nsummands of these, again in absolute value, will
+! then be less than 
+!  radix(1_i8)**(arr_max_shift + exponent(xmax_nsummands))
+! or radix(1_i8)**(digits(1_i8) - 1). This is more conservative than
+! necessary, but it also allows the postprocessing mentioned above
+! (and described later) to proceed without danger of introducing
+! overflow. 
 
 ! Determine maximum number of levels required for each field.
 ! Need enough levels to represent both the smallest and largest
@@ -935,12 +970,22 @@ module shr_reprosum_mod
                enddo
             endif
 
+! Determine additional number of levels needed to support the
+! postprocessing that reduces the magnitude of each component
+! of the integer vector of the partial sum for each thread
+! to be less than (radix(1_i8)**arr_max_shift).
+            extra_levels = (digits(1_i8) - 1)/arr_max_shift
+! Extra levels are indexed by (-(extra_levels-1):0)
+! Derivation of this is described in the comments in
+! shr_reprosum_int.
+
 ! Calculate sum
             validate = .false.
             call shr_reprosum_int(arr, arr_gsum, nsummands, dsummands, &
                                   nflds, arr_max_shift, arr_gmax_exp, &
-                                  max_levels, max_level, arr_gsum_infnan, &
-                                  validate, recompute, omp_nthreads, mpi_comm)
+                                  max_levels, max_level, extra_levels, &
+                                  arr_gsum_infnan, validate, recompute, &
+                                  omp_nthreads, mpi_comm)
 
          endif
 
@@ -1031,8 +1076,8 @@ module shr_reprosum_mod
 
    subroutine shr_reprosum_int (arr, arr_gsum, nsummands, dsummands, nflds, &
                                 arr_max_shift, arr_gmax_exp, max_levels,    &
-                                max_level, skip_field, validate, recompute, &
-                                omp_nthreads, mpi_comm                      )
+                                max_level, extra_levels, skip_field,        &
+                                validate, recompute, omp_nthreads, mpi_comm )
 !------------------------------------------------------------------------
 !
 ! Purpose:
@@ -1062,6 +1107,10 @@ module shr_reprosum_mod
                                             !  of integer expansion
       integer,  intent(in) :: max_level     ! maximum value in
                                             !  max_levels
+      integer,  intent(in) :: extra_levels  ! number of extra levels
+                                            ! needed to guarantee that
+                                            ! sum over threads or tasks
+                                            ! does not cause overflow
       integer,  intent(in) :: omp_nthreads  ! number of OpenMP threads
       integer,  intent(in) :: mpi_comm      ! MPI subcommunicator
 
@@ -1089,15 +1138,15 @@ module shr_reprosum_mod
       integer, parameter  :: max_jlevel = &
                                 1 + (digits(1_i8)/digits(1.0_r8))
 
-      integer(i8) :: i8_arr_tlsum_level(0:max_level,nflds,omp_nthreads)
+      integer(i8) :: i8_arr_tlsum_level(-(extra_levels-1):max_level,nflds,omp_nthreads)
                                    ! integer vector representing local
                                    !  sum (per thread, per field)
-      integer(i8) :: i8_arr_lsum_level((max_level+3)*nflds)
+      integer(i8) :: i8_arr_lsum_level((max_level+extra_levels+2)*nflds)
                                    ! integer vector representing local
                                    !  sum
       integer(i8) :: i8_arr_level  ! integer part of summand for current
                                    !  expansion level
-      integer(i8) :: i8_arr_gsum_level((max_level+3)*nflds)
+      integer(i8) :: i8_arr_gsum_level((max_level+extra_levels+2)*nflds)
                                    ! integer vector representing global
                                    !  sum
       integer(i8) :: IX_8          ! integer representation of current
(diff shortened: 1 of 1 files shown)


worleyph:
@rljacob , please assign an integrator, and modify the list of reviewers as appropriate.

sarats:
Thanks for the detailed explanation. This definitely aids in understanding the related code. I probably need to iterate through the code (on paper or debugger) with an example to complete my understanding.

worleyph:
@sarats , you may have difficult observing any differences arising from these changes when running E3SM. These issues were mostly identified in a standalone version and test program, as I was debugging the issue in #5276 . I can provide you with the standalone version if you wish, but it might take a little iteration between us.

For the extra_levels (first fix in second commit):

     max_nsummands = (max_nsummands/omp_nthreads) + 1
     max_nsummands = max(max_nsummands, tasks*omp_nthreads)
     xmax_nsummands = real(max_nsummands,r8)
     arr_max_shift = digits(1_i8) - (exponent(xmax_nsummands) + 1)
     extra_levels = (digits(1_i8) - 1)/arr_max_shift

Note that max_nsummands is initially the maximum number of summands per process, not the size of the global distributed summation. For extra_levels to be larger than '1' (what is assumed in the current code) requires

     2 <= (digits(1_i8) - 1)/arr_max_shift
     2*arr_max_shift <= digits(1_i8) - 1
     2*digits(1_i8) - 2*exponent(xmax_nsummands) - 2 <= digits(1_i8) - 1
     digits(1_i8) - 1 <= 2*exponent(xmax_nsummands)

Since digits(1_i8) = 63, you need 

     exponent(xmax_nsummands) > 31

so you need either the number of local summands or (MPI tasks)*(OpenMP threads) to be larger than 2,147,483,648 .

If you set noI8, then digits(1_i4) = 31 and you need

     exponent(xmax_nsummands) > 15

so you need either the number of local summands or  (MPI tasks)*(OpenMP threads) to be larger than 32,768. This is much easier to achieve, but would need to look at how reprosum is called in the code these days - any application to full fields, so a higher resolution run might have a large enough number of local summands?

Even when the logic indicates that extra_levels should be larger than 1, it doesn't mean that this would be needed in practice. I can come up with test cases that do in fact require (at least some of) the extra levels, but I have not done so with E3SM, especially not when using 8 byte integers. When the extra levels are not needed, they are just zeroes that get summed and then ignored when generating the floating point values to be summed at the end, and so have no impact on the numerics (as compared to the current algorithm).

If you want to test this in E3SM, I advise adding

     #define noI8 1

at the top of shr_reprosum_mod.F90 to make it more likely to show a difference.

worleyph:
@sarats, for the second fix in the second commit, replacing

     RX_8 = i8_arr_tlsum_level(ilevel,ifld,ithread)
     IX_8 = int(scale(RX_8,-arr_max_shift),i8)
     if (IX_8 /= 0_i8) then

by

     if (abs(i8_arr_tlsum_level(ilevel,ifld,ithread)) >= &
            (i8_radix**arr_max_shift)) then

        IX_8 = i8_arr_tlsum_level(ilevel,ifld,ithread) &
               / (i8_radix**arr_max_shift)

I did not even try to find a test case where this made a difference. In fact, it may not ever be an issue (truncating integer*8 value by assigning it to a real*8, shifting, then assigning back to an integer*8 losing precision that would show up in overlap), but I cannot rule it out based on how much thought I was willing to give the issue. My first fix was to simply loop over this this logic (testing whether there was still overlap and if so, repeat), but using integer division to do the truncation and shift was simpler and avoids the issue entirely. The original code was simply to be "consistent" with the logic used when first decomposing the real*8 summands into the integer vector representation (I guess - really no idea what my motivation was).

In all my testing, the two versions have been BFB, but I can't guarantee this. This new version is correct (based on what the comments say is supposed to happen), and the original is maybe not correct.

worleyph:
@sarats, with respect to the first commit, there are two steps to the logic to give all components the same sign: First you rerun the logic to remove overlap in the components (after the summation over MPI tasks and OpenMP threads). This is not only necessary in order for the second step to work (actually changing the signs to all be the same), but should also be sufficient by itself to eliminate the possibility of catastrophic cancellation (as the summands being added are guaranteed to not be almost the same magnitude but different signs). I  included the second step, that imposes the same sign, just to avoid needing to justify this reasoning - same sign means no cancellation and so be done with it. Note that the same sign logic is actually used in the next PR, so getting this right is ultimately important.

As to when this could change the numerics, you would need the integer vector to have components with different signs interspersed with zero components. In my standalone experiments, I summed sinusoidal functions (approximately O(1) with a few values that were almost zero, so large negative exponents). This provided the "gaps" represented by the zero components, i.e. components with implicit exponents that were not represented in the summands. When summing these, the components with the very small exponents typically did not contribute to the final sum (precision was lost when contructing the final real*8 value). I also added some very large positive and negative values (summing to zero) to also force "gaps" in the components. I don't know if any of this would show up in E3SM simulations. In the old spectral dycore, maybe (sums around the poles often were "almost" zero and the summation of sinusoidal type functions).

Even then, it is not obvious whether the slight change to the set of floating point values generated by this fix would change the numerics, so can not provide guidance on how to devise a non-BFB example case.

sarats:
@worleyph Thanks again for the details.
I'm inching closer to enlightenment :) the more I think about it. 

Btw, the standalone test seems like a great addition as a unit test for future use as well. 
Please send any existing notes you have and I or someone else from Infrastructure can package it up as needed in case you don't wish to pursue that.

sarats:
@rljacob Remind me, what's a good location for a unit test (standalone, non-CIME) like this?

sarats (approved):

worleyph:
@sarats, the standalone case packaging should wait until after the next PR (assuming that it is deemed acceptable).

philipwjones (approved):
Approved based on Pat's explanations and visual inspection that the code corresponds to that. A unit test exists and should also be added later once the final PR in this sequence is complete.

ambrad:
I provisionally approve, but if I were developing this code, I would run e3sm_integration against baselines and with temporary output code showing how often these changes are triggered, as well as a longer water-cycle simulation with that same output, and then think about whether that output matches my expectations.

ambrad (approved):

worleyph:
@ambrad, 

     "but if I were developing this code, I would run e3sm_integration
      against baselines and with temporary output code showing how
      often these changes are triggered, as well as a longer water-cycle
      simulation with that same output, and then think about whether
      that output matches my expectations."

I am no longer in a position to do this, but am happy to advise on this if someone else wants to take it on. My current testing is limited to F1850 and WCYC1850 ne30pg2 threaded and nonthreaded 5 day experiments on Chrysalis. (I do not have access to other platforms any more, except maybe Anvil.) The final PR will not be BFB, so whether this one is or not is not super critical, except as a possible red flag that there is an error in case it is not. I will definitely need help testing the non-BFB third PR. Thanks.

ambrad:
> My current testing is limited to F1850 and WCYC1850 ne30pg2 threaded and nonthreaded 5 day experiments on Chrysalis.

Chrysalis is perfect. I recommend you run e3sm_integration against Chrysalis's baselines and examine the results to see if they agree with your expectations. If you get all PASSes against baselines, no further examination is needed unless you're curious about internal BFB code path changes.

ambrad (changes requested):
Removing my approval until I see a discussion of testing.

worleyph:
@amametjanov or @sarats, could I ask you to run the e3sm integration test suite with this PR, as @ambrad has requested? If it is not BFB, I'll jump back in and see if I can determine where and why.

worleyph:
Note that I did find a bug in the second commit. In the process of backing out manageable commits from the final proposed PR, still to be submitted, the new lower bound on the ilevel loops (min_level) was not introduced into all ilevel loops. This was not caught in testing as the min_level is identical with the previous lower bound ('0') in all of my E3SM testing scenarios. I tried to push the update, but this failed with

     @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
     @    WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!     @
     @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@

I am trying to push the update from Chrysalis.

Any idea whether this is a transient error, or do I need to do something different at my end?

Thanks.

worleyph:
@xylar . Thanks. Deleted the entry in my known_hosts files and started over. It still wan't happy:

     The authenticity of host 'github.com (140.82.113.3)' can't be established.
     ...
     Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
     ...
     Warning: the ECDSA host key for 'github.com' differs from the key for the IP address '140.82.113.3'
     ...
     Are you sure you want to continue connecting (yes/no)? yes

but it let me complete the push.

sarats:
FYI https://github.blog/2023-03-23-we-updated-our-rsa-ssh-host-key/

sarats:
@rljacob Wouldn't a PR be put on next for all tests including `e3sm_integration` anyway?

ambrad:
@sarats, yes. I withdraw my comment now that I see it is controversial. I personally like to know the testing status of a PR whose BFB status is unknown before it goes on to next, but I see that is just my own opinion. I'm going to remove myself as a reviewer so that I don't cause extra work for others.

rljacob:
If there's a chance PR could be non-BFB, we would like to know before merging to next.  Because if it is and can't be fixed quickly it then has to be reverted from next which is a pain.  @amametjanov please run integration on this with a baseline compare.

worleyph:
@rljacob , @amametjanov , Andrew sent me instructions on how to run the suggested test. Looks dead simple. So, If Az has not already started this, I can do it. Please indicate whether I should go ahead or not. Thanks.

amametjanov:
Yes, a run is in progress.

amametjanov:
Baseline-comparison results are at https://my.cdash.org/viewTest.php?buildid=2309947 :
- all 110 `e3sm_integration` tests are BFB
- 15 out of 110 are NLFAILs due to namelist changes in MPAS PR 5541 (that added `config_output_reference_time`), which was merged into master on Mar 22 and HEAD of this branch is from Mar 21: i.e. master baselines have this var and this branch does not.

ambrad (approved):

Issue #5535, opened 2023-03-20
Floating invalid in DryDepVelocity.F90 and rapid ozone loss after coupling with UCI_chem


wlin7:
In a testing branch with  #4707 merged into master, debug mode exposes a floating invalid at line 469 of [elm/src/biogeochem/DryDepVelocity.F90](https://github.com/E3SM-Project/E3SM/blob/master/components/elm/src/biogeochem/DryDepVelocity.F90)
`rs = 1.0_r8 / ( fsun(pi)/rssun(pi) + (1.0_r8 - fsun(pi))/rssha(pi) )`

And when it occur, it is due to rssun(pi) = 0.0. 

The following logic is added to allow the debug mode to proceed.
```
if (rssun(pi)==0._r8) then
    rs = 0.0_r8
else
    rs = 1.0_r8 / ( fsun(pi)/rssun(pi) + (1.0_r8 - fsun(pi))/rssha(pi) )
endif
```
Moreover, it is found that without the above logic to contain floating invalid, the simulation would experience rapid ozone loss: global tropospheric column ozone (TCO) losing 3/4 of the normal value and stratospheric column ozone (SCO) losing nearly 2/3.

Is it a bug to for `rssun(pi) `to have zero value? If occurrence of zero value is acceptable, if the above if-else-endif block to control the assignment for `rs` acceptable?

Where is the root cause? Related elm module need corresponding treatment when more gas species added?

Urgent helps are needed in order for merging primary v3atm features to master.

wlin7:
Hi @bishtgautam , can you please take a look at this issue?  @tangq @keziming did additional experiments and found the dry deposition fluxes calculated in land are unrealistically high. Shutting off that fluxes would prevent the rapid ozone loss. The occurrence of floating invalid in DryDepVelocity.F90 is just a hint, the root cause may be elsewhere.

FYI: The issue does not exist in [v3atm repo](https://github.com/E3SM-Project/v3atm/tree/NGD_v3atm) . The land component in that branch has no update since Jul 22, 2021 (e3472d21b20e57eaaeb7078c82c3578e25dc6e88).

tangq:
@bishtgautam and @wlin7,
[...]

PR #5552, opened 2023-03-24
Update stomatal resistance to properly account for sunlit or shaded conditions

wlin7:
This is to fix issue #5535, which exists in current repo but not exposed 
until testing with pre-merged branch for PR #4707. 

The original formulation of adding sunlit and shaded leaves in parallel 
did not consider the conditions when rssun and/or rssha = 0.0, which can 
not only causing floating invalid -- division by 0, but also leading to 
unrealistic dry deposition fluxes, hence rapid ozone loss. Other gas 
constituents can also be affected.

[non-BFB] expected for all F, I, and B tests. But unclear why current master 
                 hasn't experienced issue.

### DIFF
--- a/components/elm/src/biogeochem/DryDepVelocity.F90
+++ b/components/elm/src/biogeochem/DryDepVelocity.F90
@@ -466,16 +466,17 @@ CONTAINS
 
                   !Following Emmons et al (2020, JAMES), the stomatal resistances for shaded and sunlit
                   !leaves should be added in parallel.  Also need to account for the conditions of
-                  !rssun=0.0 and/or rssha=0.0
+                  !rssun=0.0 and/or rssha=0.0, for which only fsun=0 or fsun=1 is considered, respectively,
+                  !following the suggestion by Sam Silva.
 
                   if (rssun(pi) > 0._r8 .and. rssha(pi) > 0._r8) then
                      rs = 1.0_r8 / ( fsun(pi)/rssun(pi) + (1.0_r8 - fsun(pi))/rssha(pi) )
-                  else if (rssun(pi) > 0._r8 .and. rssha(pi) == 0._r8) then
-                     rs = fsun(pi)*rssun(pi)
-                  else if (rssun(pi) == 0._r8 .and. rssha(pi) > 0._r8) then
-                     rs = rssha(pi)*(1._r8-fsun(pi))
+                  else if (rssun(pi) > 0._r8 .and. fsun(pi) == 1._r8) then
+                    rs = rssun(pi)
+                  else if (rssha(pi) > 0._r8 .and. fsun(pi) == 0._r8) then
+                    rs = rssha(pi)
                   else
-                     rs = 0._r8
+                     rs = 1.e36_r8
                   endif
 
                   if (rs==0._r8) then ! fvitt -- what to do when rs is zero ???
--- a/components/elm/src/biogeochem/DryDepVelocity.F90
+++ b/components/elm/src/biogeochem/DryDepVelocity.F90
@@ -466,8 +466,7 @@ CONTAINS
 
                   !Following Emmons et al (2020, JAMES), the stomatal resistances for shaded and sunlit
                   !leaves should be added in parallel.  Also need to account for the conditions of
-                  !rssun=0.0 and/or rssha=0.0, for which only fsun=0 or fsun=1 is considered, respectively,
-                  !following the suggestion by Sam Silva.
+                  !rssun=0.0 and/or rssha=0.0, for which only fsun=0 or fsun=1 is considered, respectively.
 
                   if (rssun(pi) > 0._r8 .and. rssha(pi) > 0._r8) then
                      rs = 1.0_r8 / ( fsun(pi)/rssun(pi) + (1.0_r8 - fsun(pi))/rssha(pi) )
--- a/components/elm/src/biogeochem/DryDepVelocity.F90
+++ b/components/elm/src/biogeochem/DryDepVelocity.F90
@@ -465,8 +465,19 @@ CONTAINS
                else
 
                   !Following Emmons et al (2020, JAMES), the stomatal resistances for shaded and sunlit
-                  !leaves should be added in parallel
-                  rs = 1.0_r8 / ( fsun(pi)/rssun(pi) + (1.0_r8 - fsun(pi))/rssha(pi) )
+                  !leaves should be added in parallel.  Also need to account for the conditions of
+                  !rssun=0.0 and/or rssha=0.0
+
+                  if (rssun(pi) > 0._r8 .and. rssha(pi) > 0._r8) then
+                     rs = 1.0_r8 / ( fsun(pi)/rssun(pi) + (1.0_r8 - fsun(pi))/rssha(pi) )
+                  else if (rssun(pi) > 0._r8 .and. rssha(pi) == 0._r8) then
+                     rs = fsun(pi)*rssun(pi)
+                  else if (rssun(pi) == 0._r8 .and. rssha(pi) > 0._r8) then
+                     rs = rssha(pi)*(1._r8-fsun(pi))
+                  else
+                     rs = 0._r8
+                  endif
+
                   if (rs==0._r8) then ! fvitt -- what to do when rs is zero ???
                      rsmx(ispec) = 1.e36_r8
                   else


wlin7:
Hi @bishtgautam , Just taking the momentum to have this fix in, in anticipation of the integration of PR #4707, and asking you to review. We need to have this in place first, though the problem does not exist in the current branch for PR #4707 (which has an older version of land). Your suggestion was key leading to the eventual fix.

@balwinder, you helped putting in the original fix via #[4678](https://github.com/E3SM-Project/E3SM/pull/4678). This PR is on top of that, so I ask you to be the integrator. 

Thanks.

wlin7:
@singhbalwinder , since this PR seems only to become necessary until #4707 is merged, if confirmed it would cause NBFB for many tests, one option is to wait and merge it on the same day as merging PR #4707. I am testing if this fix would indeed cause BFB. Surprised it hasn't caused problem in current master.

tangq:
@wlin7 , current master uses the O3v2 package, which doesn't use this dry deposition flux. This is probably why it doesn't cause problem till the interactive tropospheric chemistry is merged via PR#4707.

singhbalwinder:
The formulation is exactly the same for the following `if` statement:
`if (rssun(pi) > 0._r8 .and. rssha(pi) > 0._r8) then`

Perhaps in the current cases, rssun(pi) and rssha(pi) are always >0. This can also explain why it is not causing any NBFB issues with the current tests.

bishtgautam:
I found two issues:

1. The `rs` computation for the second and third conditions is incorrect.

2. Looking at the [latest CTSM code](https://github.com/ESCOMP/CTSM/blob/master/src/biogeochem/DryDepVelocity.F90#L524-L527), we are missing one term (`elai`) to correctly compute stomatal conductance and the code should be modified as: 

```
                  if (rssun(pi) > 0._r8 .and. rssha(pi) > 0._r8) then
                     rs = 1.0_r8 / ( fsun(pi) * elai(pi) /rssun(pi) + (1.0_r8 - fsun(pi)) * elai(pi) /rssha(pi) )
                  else if (rssun(pi) > 0._r8 .and. rssha(pi) == 0._r8) then
                     rs = 1.0_r8 / fsun(pi)* elai(pi) / rssun(pi)
                  else if (rssun(pi) == 0._r8 .and. rssha(pi) > 0._r8) then
                     rs = 1.0_r8 / rssha(pi) * elai(pi) /(1._r8-fsun(pi))
                  else
                     rs = 0._r8
                  endif
```

I looked into the history of DryDepVelocity.F90 in CTSM and found the following:
- In [clm4_5_1_r106](https://github.com/ESCOMP/CTSM/commit/13721d61c51c414159c974f9545572266eb78c73#diff-883096632d5ee8eb3575113ff231e8769d6c6a35e8ae38b33fd9243d34be0616R464), the `rs` computation was updated to included `elai` but the combined resistance of sunlit and shaded leaves was still incorrectly computed assuming the leaves are in series.
- In [clm4_5_18_r268](https://github.com/ESCOMP/CTSM/commit/9e9900bf5c1725d9b1fb87a4d9c74260212034fb#diff-883096632d5ee8eb3575113ff231e8769d6c6a35e8ae38b33fd9243d34be0616R517), the `rs` computation was corrected to account for sunlit and shaded leaves are in parallel.

In ELM, the `rs` was updated to correctly account for leaves being in parallel via PR #4678, but `elai` was not accounted for in the computation.

wlin7:
@bishtgautam , thanks for digging into this issue further. We may need to include elai in the calculation, not sure if other changes also need to be included in order to make them consistent. @mingxuanwupnnl also mentioned there are other fixes in CSTM, which we may incorporate separately.

Regarding the calculations for the 2nd and 3rd conditions, I tend to think the following formulation is actually more appropriate
 ```
                 if (rssun(pi) > 0._r8 .and. rssha(pi) > 0._r8) then
                     rs = 1.0_r8 / ( fsun(pi)/rssun(pi) + (1.0_r8 - fsun(pi))/rssha(pi) )
                  else if (rssun(pi) > 0._r8 .and. rssha(pi) == 0._r8) then
                     rs = fsun(pi)*rssun(pi)
                  else if (rssun(pi) == 0._r8 .and. rssha(pi) > 0._r8) then
                     rs = rssha(pi)*(1._r8-fsun(pi))
                  else
                     rs = 0._r8
                  endif
```
In the very original formulation, like in ELMv1, the sunlit and shaded contributions are added in series, which is incorrect as reported in Emmons et al (2020) (and issue #4677).  But when there is only one of sunlit or shaded leaves at play, my understanding is, the standalone term in that incorrect addition is the correct one to use,  instead of taking the converted term derived from idea of the parallel addition.

                  `rs=(fsun(pi)*rssun(pi))+(rssha(pi)*(1._r8-fsun(pi)))`

And similarly if taking into account elai, the calculations would be 
 ```
                 if (rssun(pi) > 0._r8 .and. rssha(pi) > 0._r8) then
                     rs = 1.0_r8 / ( fsun(pi) * elai(pi) /rssun(pi) + (1.0_r8 - fsun(pi)) * elai(pi) /rssha(pi) )
                  else if (rssun(pi) > 0._r8 .and. rssha(pi) == 0._r8) then
                     rs = fsun(pi)*rssun(pi)/elai(pi)
                  else if (rssun(pi) == 0._r8 .and. rssha(pi) > 0._r8) then
                     rs = rssha(pi)/elai(pi)*(1.-fsun(pi))
                  else
                     rs = 0._r8
                  endif
```
For sure I don't have much knowledge in this. Please correct me if taking the standalone term from the old addition in series formulation is inappropriate. The different calculations do impact simulation results, as documented on #5535. Other please also chime in.

bishtgautam:
My point is that the computation for the 2nd and 3rd conditions should be consistent with the 1st condition and can be thought of as:

```
rs = 1.0_r8 / ( fsun(pi) * elai(pi) /rssun(pi) + (1.0_r8 - fsun(pi)) * elai(pi) /rssha(pi) )
rs = 1.0_r8 / ( fsun(pi) * elai(pi) /rssun(pi) +  0                                        )
rs = 1.0_r8 / ( 0                              + (1.0_r8 - fsun(pi)) * elai(pi) /rssha(pi) )
```

However, 1st, 2nd, and 3rd terms aren't consistent in your suggested code.

tangq:
I reached out to @samjsilva91 for suggestions, who worked on E3SM ozone dry deposition before.

wlin7:
Hi all, for a test, I ran simulations with both approaches (elai still not considered):  The run with the committed changes ([866ba5f](https://github.com/E3SM-Project/E3SM/pull/5552/commits/866ba5f4a139574244efe3b4af694f0d277681ae)) completed 5-year simulations. The run with the other approach encountered `ERROR: shr_reprosum_calc ERROR: NaNs or INFs
` after  about 2 years and 6 months.

As described in #5535, the limiter for base_sol and base_sol_reset still need to be applied otherwise the run with ne30pg2 grid would fail early.

Simulations with both approaches  for rs set rs=1.0e36_r8 when no sunlit nor shaded leaves as suggested by @samjsilva91 .

Please continue to iterate on this.

tangq:
@wlin7 and @bishtgautam , below is my understanding of the current status. Correct me if some points are not accurate.

- rs = 1.e36_r8 should be used when there are no sunlit or shaded leaves (suggested by @samjsilva91 ). This will cut off dry deposition to stomatal. But the changes in [866ba5f](https://github.com/E3SM-Project/E3SM/pull/5552/commits/866ba5f4a139574244efe3b4af694f0d277681ae) have rs=0, which will lead to large stomatal dry deposition and thus potential problems.
- We have two ways to calculate rs when rssha(pi) == 0._r8 or rssun(pi) == 0._r8, but we are not 100% sure which one to use.
1: 
```
else if (rssun(pi) > 0._r8 .and. rssha(pi) == 0._r8) then
                    rs = fsun(pi)*rssun(pi)
else if (rssun(pi) == 0._r8 .and. rssha(pi) > 0._r8) then
                    rs = rssha(pi)*(1._r8-fsun(pi))
```
2:
```
else if (rssun(pi) > 0._r8 .and. rssha(pi) == 0._r8) then
           rs = 1.0_r8 / ( fsun(pi) * elai(pi) /rssun(pi) +  0                                        )
else if (rssun(pi) == 0._r8 .and. rssha(pi) > 0._r8) then
           rs = 1.0_r8 / ( 0                              + (1.0_r8 - fsun(pi)) * elai(pi) /rssha(pi) )
```
- It seems reasonable to include the `elai` term in the equation because CLM did so, but we are not sure if that's correct for ELM.
- The limiter for base_sol and base_sol_reset is a surprise to me. Before merging to master, we ran multiple simulations of 1870-2014 with the chemUCI branch and didn't need this limiter. I wonder if we need the limiter if replacing rs=0.r8 with rs=1.e36_r8.

I feel we need to ask someone with the proper background to help with these remaining issues except the last one.

tangq:
Email response from Sam Silva (Thanks!):

It is important to note that this should really just exist as an error check.
The physical meaning or rssha == 0 is very unclear, this mean that the stomatal conductance (rate of gas exchange) is infinity. Infinitely large stomata minorly violates nearly every conservation law in the known universe. This is an undefined concept to me and represents a potential bug in the land model code.
 
Thinking about this more deeply, I would prefer setting rs = 1e36 anytime either rssun or rssha is zero and fsun == 0 or 1, respectively. Otherwise the normal weighting by LAI makes sense.
 
More specifically:
 
                if (rssun(pi) > 0._r8 .and. rssha(pi) > 0._r8) then
                    rs = 1.0_r8 / ( fsun(pi) * elai(pi) /rssun(pi) + (1.0_r8 – fsun(pi)) * elai(pi) /rssha(pi) )
                 else if (rssun(pi) > 0._r8 .and. fsun == 1._r8 .and. elai(pi) > 0.r_8 ) then
                    rs = rssun(pi)/elai(pi)
                 else if (rssha(pi) > 0._r8 .and. fsun == 0._r8  .and. elai(pi) > 0.r_8) then
                    rs = rssha(pi)/elai(pi)
                 else
                    rs = 1e36._r8
                 endif

rljacob:
status: plan to merge this with UCI chem PR.

wlin7:
Hi @bishtgautam @tangq , I have updated the PR with the changes suggested by Sam Silva (as Qi posted above). The else if block now only considers either fsun=1 or 0. With this, we don't need to deal with whether fsun (or 1-fsun) should be used as a multiplier or divider -- namely using it as a weighting or a nomalization factor. The current version does not consider elai.

I have only tested for a multi-month simulation. The run and simulated ozone are stable.

Can you please review, and approve the change if you think acceptable? This is preparing the fix to be merged together with UCI-chem PR into master.

bishtgautam (approved):

components/elm/src/biogeochem/DryDepVelocity.F90
```diff
@@ -465,8 +465,20 @@ subroutine depvel_compute( bounds, &
                else
 
                   !Following Emmons et al (2020, JAMES), the stomatal resistances for shaded and sunlit
-                  !leaves should be added in parallel
-                  rs = 1.0_r8 / ( fsun(pi)/rssun(pi) + (1.0_r8 - fsun(pi))/rssha(pi) )
+                  !leaves should be added in parallel.  Also need to account for the conditions of
+                  !rssun=0.0 and/or rssha=0.0, for which only fsun=0 or fsun=1 is considered, respectively,
+                  !following the suggestion by Sam Silva.
```
bishtgautam, line 470: I would suggest removing `following the suggestion by Sam Silva.` from the code and adding it as a comment in the PR discussion.

wlin7: Good idea. Just did.

tangq (approved):
The changes look good to me. Thank you, @wlin7 .

It is worth noting that the `elai` changes are not included in this PR.

wlin7:
Merged to next.

wlin7:
Merged to master.

PR #5560, opened 2023-03-27
Eliminate potential nonreproducibility in final local sum

### LINKED ISSUE #5276: improving accuracy of shr_reprosum_calc

worleyph:
The default reprosum algorithm converts each floating point summand
into a vector of integers representation. These vectors are summed
locally first, and then globally (using an mpi_allreduce). All of this 
is exact and reproducible. The final step is to convert the final sum, 
in the vector of integers representation, back into a floating point  
value.

This final sum begins by converting each element of the integer vector  
into a floating point value, and then adding these, smallest to
largest. Round off can occur in the least significant digit of the 
final result, and whether round off occurs is sensitive to the 
particular floating point summands. The numbers of MPI tasks and 
OpenMP threads can affect the length of the integer vector, and thus 
the floating point summands.
In consequence, reproducibility with respect to degree of parallelism 
is not guaranteed. (The situation in which reproducibility is lost   
has never been observed in E3SM simulations, but it cannot be
ruled out.)

This PR modifies the final summation so that the same set of floating
point summands is generated independent of the number of MPI tasks   
and number of OpenMP threads, establishing reproducibility.

Fixes #5276

[non-BFB]

***
(Use above for merge commit message.  Additional detail below.)

The reprosum algorithm converts each floating point summand into a vector of integers representation. These vectors are summed locally first, and then globally (using an mpi_allreduce). All of this is exact and reproducible. The final step is to convert the final sum, in the vector of integers representation, back into a floating point value.

For this last step, each component integer in the vector representation is converted back into a floating point value (call it a_{i} for the ith element of the vector), and then these values are summed locally, smallest to largest, to get the final value. The integer representation is first massaged so that all a_{i} have the same sign and the range of values is usually not overlapping (least significant binary digit in abs(a_{i}) represents a value larger than that represented by abs(a_{i+1}) in most cases, but sometimes is equal to).

When constructing the final floating point value, round-off error can be introduced in the least significant digit of the final sum. This level of inaccuracy has no practical impact on overall model accuracy, but it can potentially lead to loss of reproducibility with respect to the numbers of MPI tasks and OpenMP threads. The issue is that the set of floating point values generated from the vector of integers will vary with the number of components in the integer vector and the implicit exponents associated with each component, which are a functions of, among things, the number of MPI tasks and number of OpenMP threads. Any round-off error will be sensitive to the specific floating point values being summed, even though the sum would be invariant with respect to exact arithmetic. This nonreproducibility has been observed when changing an 8-byte integer vector representation to a 4-byte integer vector representation (which also changes the vector length and assocated implicit exponents), but only in stand-alone contrived examples and never in E3SM. However, we can not rule out nonreproducibility occurring in practice (without a more detailed analysis of the range of sums likely to occur in E3SM simulations).

The solution implemented here is to make the set of floating values generated from the integer vector independent of the size of the vector and definition of implicit exponents. Effectively, a new integer vector is generated from the original in which each component integer has a fixed number of significant digits (e.g., digits(1.0_r8)), and the floating point values are generated using this vector. As number, and value, of the significant digits is unchanged between the two representations, the implied value is unchanged, but any rounding error in the summation from using the new representation will be reproducible.

Since digits(1.0_r8) would be too many digts for a 4-byte integer representation of the vector, and since we want to continue to support this option, each new floating point value is actually contructed from some number of floating point values with possibly fewer significant digits, but in such a way that the summation of these is exact (no rounding error possible), and is identical to the floating point value that would be generated if the integer vector had been modified as indicated in the previous paragraph.

Note that another possibility to address this issue is by further decreasing the likelihood of rounding errors by calculating this final sum of floating point values using the DDPDD algorithm locally. (DDPDD is an alternative algorithm already available in the reprosum code.) DDPDD uses a two-floating-point-value representation of the intermediate values in the summation to approximately double the precision, i.e., for double precision values, summing in quad precision. In this case, round-off error is isolated to the least significant digit in the quad precision value, and it is even less likely that this will impact the least significant digit in the resulting double precision value (though not impossible).

This 'local' DDPDD approach was implemented, though it is not included in this commit as the proposed algorithm is preferred as it is always reproducible, and not just almost always. However, comparing the two approaches in E3SM experiments, they are BFB, implying that the proposed algorithm improves accuracy as well. In similar experiments, the propsoed is also BFB with using the existing alternative 'distributed' DDPDD algorithm.

As implied, the proposed algorithm (and the two DDPDD approaches) are not BFB with the current default integer vector algorithm in E3SM simulations. The differences arise from differing round-off in the least significant digit, but this does accumulate over time. For example, for a 5 day run on Chrysalis using

     --compset WCYCL1850 --res ne30pg2_EC30to60E2r2 --pecount XS

(675 ATM processes) there are 3669 reprosum calls, and using the proposed algorithm (or 'local' DDPDD or 'distributed' DDPDD) for the final sum changes the result 200 times (so 5.5% of the time).

Looking at the actual difference (printing out the mantissas in the final sums in octal), it is always +/- 1 in the least significant binary digit. But this is still enough to perturb the model output, e.g. (from the atm.log)

     nstep, te 241   0.26199582846629281E+10   0.26199596541672482E+10   0.75730462729090879E-04   0.98528834650276956E+05

vs.

     nstep, te 241   0.26199694926722941E+10   0.26199707204570022E+10   0.67893797509715502E-04   0.98528681087863835E+05

Similarly for a 5 day high resolution F case

      --compset F2010 --res ne120pg2_r0125_oRRS18to6v3

(2560 ATM processes) there are 9609 reprosum calls, and using the proposed algorithm for the final sum changes the result 104 times (so 1.1% of the time). The difference in the atm.log output is

     nstep, te      481   0.26056729910814352E+10   0.26056734037400241E+10   0.45650283260804208E-04   0.98502662499849699E+05

vs.

     nstep, te      481   0.26056829597470064E+10   0.26056833721086311E+10   0.45617405260189644E-04   0.98502719483102352E+05

Note that the computational cost of the proposed algorithm for the conversion of the integer vector into a floating point value is a little more expensive than that of the original algorithm, but the performance difference is in the noise as most of the cost of shr_reprosum_calc is in the common aspects of the current and proposed algorithms, i.e., determination of global max/min and number of summands, conversion of each floating point summand into the vector of integers representation, and the local and global sums of the integer vectors. For example, performance comparisons between the new and old approaches for calls to shr_reprosum_cale are inconclusive as to which is faster. Also note that the 'local' DDPDD algorithm is slightly more expensive than the proposed algorithm, but the difference is again in the noise.

[Non-BFB]

Fixes #5276

### DIFF
--- a/share/util/shr_reprosum_mod.F90
+++ b/share/util/shr_reprosum_mod.F90
@@ -282,10 +282,10 @@ module shr_reprosum_mod
 ! models use the same base, e.g.  
 !  radix(1.0_r8) == radix(1_i8)
 ! for real*8 and integer*8. If not, then the alternative algorithm
-! mentioned above and described below is used instead. The integer
-! representation must also have enough digits for the potential growth
-! of the sum for each level, so could conceivably be too small for a
-! large number of summands. 
+! DDPDD mentioned above and described below is used instead. The
+! integer representation must also have enough digits for the
+! potential growth of the sum for each level, so could conceivably be
+! too small for a large number of summands. 
 !
 ! Upper bounds on the total number of summands and on all intermediate
 ! sums are calculated as
@@ -322,26 +322,38 @@ module shr_reprosum_mod
 ! conversion of the vector of integer representation to a floating
 ! point value may be subject to rounding errors. Before the
 ! conversion, the vector of integers is adjusted so that all elements
-! have the same sign, eliminating the possibility of catastrophic
-! cancellation. These are integer operations, so no accuracy is
-! lost. Next, each element of the vector of integers is converted to a
-! floating point value and added into the intermediate sum, in
-! smallest to largest order (in absolute value). Any rounding error is
-! in the least significant digit for each intermediate sum, and the
-! exponents for each summand are monotonically increasing, so the
-! rounding errors do not accumulate. Thus, the relative difference
-! between the 'exact' value and that calculated by this algorithm will
-! be approximately machine epsilon for 64-bit real values, and any
-! difference will be restricted to the 16th significant digit in a
-! decimal representation. This does mean that the values calculated
-! with the integer*8 internal representation could differ from those
-! calculated using the integer*4 internal representation, but the
-! differences will be in the least significant digits. Similarly, the
-! results may change (in the least significant digits) with a change
-! in system, compiler, or compiler flags. However, for a fixed
-! internal representation, processor architecture, compiler, and
-! compiler options, the result will (still) be reproducible with
-! respect to MPI task and OpenMP thread counts.
+! have the same sign, and so that the value, in absolute value, at a
+! given level is strictly less than what can be represented at the
+! next lower level (larger exponent) and strictly greater than what
+! can represented at the next higher level (smaller exponent). Since
+! all elements have the same sign, the sign is set to positive
+! temporarily and then restored when the conversion to floating point
+! is complete. These are all integer operations, so no accuracy is
+! lost. These adjustments eliminate the possibility of catastrophic
+! cancellation. Also, when converting the individual elements to
+! floating point values and summing them, the summation is now
+! equivalent to concatenating the digits in the mantissas for the
+! component summands. In consequence, in the final step when each
+! element of this modified vector of integers is converted to a
+! floating point value and added into the intermediate sum, any
+! rounding is limited to the least significant digit representable
+! in the final floating point sum.
+!
+! Any such rounding error will be sensitive to the particular floating
+! values generated from the integer vector, and so will be 
+! sensitive to the number of levels in the vector and the implicit
+! exponent associated with each level, which are themselves functions
+! of the numbers of MPI tasks and OpenMP threads and the number of
+! digits representable in an integer. To avoid this sensitivity,
+! (effectively) generate a new integer vector in which each component
+! integer has a fixed number of significant digits (e.g.,
+! digits(1.0_r8)) and generate the floating point values from these
+! before summing. (See comments in code for more details.) This
+! creates a sequence of floating point values to be summed that are
+! independent of, for example, the numbers of MPI tasks and OpenMP
+! threads or whether using integer*8 or integer*4 internal
+! representations in the integer vector, and thus ensure
+! reproducibility with respect to these options.
 !
 ! Description of optional parameters for integer vector algorithm:
 !-----------------------------------------------------------------
@@ -1135,9 +1147,6 @@ module shr_reprosum_mod
 !
 ! Local workspace
 !
-      integer, parameter  :: max_jlevel = &
-                                1 + (digits(1_i8)/digits(1.0_r8))
-
       integer(i8) :: i8_arr_tlsum_level(-(extra_levels-1):max_level,nflds,omp_nthreads)
                                    ! integer vector representing local
                                    !  sum (per thread, per field)
@@ -1149,9 +1158,10 @@ module shr_reprosum_mod
       integer(i8) :: i8_arr_gsum_level((max_level+extra_levels+2)*nflds)
                                    ! integer vector representing global
                                    !  sum
-      integer(i8) :: IX_8          ! integer representation of current
-                                   !  jlevels of X_8 ('part' of
-                                   !  i8_arr_gsum_level)
+      integer(i8) :: i8_gsum_level(-(extra_levels-1):max_level)
+                                   ! integer vector representing global
+                                   !  sum for one field
+      integer(i8) :: IX_8          ! integer representation of r8 value
       integer(i8) :: i8_sign       ! sign global sum
       integer(i8) :: i8_radix      ! radix for i8 variables (and r8
                                    !  variables by earlier if-test)
@@ -1164,7 +1174,7 @@ module shr_reprosum_mod
       integer :: isum_beg(omp_nthreads), isum_end(omp_nthreads)
                                    ! range of summand indices for each
                                    !  OpenMP thread
-      integer :: ifld, isum, ithread
+      integer :: ifld, isum, ithread, jlevel
                                    ! loop variables
       integer :: arr_exp           ! exponent of summand
       integer :: arr_shift         ! exponent used to generate integer
@@ -1178,13 +1188,28 @@ module shr_reprosum_mod
       integer :: min_level         ! index of minimum levels (including
                                    !  extra levels) for i8_arr_tlsum_level
       integer :: ioffset           ! offset(ifld)
-      integer :: jlevel            ! number of floating point 'pieces'
-                                   !  extracted from a given i8 integer
+      integer :: svlevel           ! number of summands in summand_vector
       integer :: ierr              ! MPI error return
-      integer :: LX(max_jlevel)    ! exponent of X_8 (see below)
+      integer :: LX                ! exponent of X_8 (see below)
       integer :: veclth            ! total length of i8_arr_lsum_level
-      integer :: sum_digits        ! lower bound on number of significant
-                                   !  in integer expansion of sum
+      integer :: i8_digit_count    ! number of digits in integer
+                                   !  expansion of sum
+      integer :: i8_begin_level    ! level starting from in
+                                   !  creating next 'exactly representable'
+                                   !  floating point value from modified
+                                   !  integer expansion of the sum
+      integer :: i8_trunc_level    ! level at which the number of digits in
+                                   !  the modified integer expansion of the
+                                   !  sum exceeds the number of representable
+                                   !  digits in the floating point sum
+      integer :: i8_trunc_loc      ! location of last digit at i8_trunc_level
+                                   !  in the modified integer expansion of the
+                                   !  sum that is representable in the floating
+                                   !  point sum
+      integer(i8) :: i8_trunc_level_rem
+                                   ! truncated digits at i8_trunc_level
+                                   !  in the modified integer expansion
+                                   !  of the sum
       integer :: curr_exp          ! exponent of partial sum during
                                    !  reconstruction from integer vector
       integer :: corr_exp          ! exponent of current summand in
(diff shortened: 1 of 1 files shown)


worleyph:
@rljacob , please add reviewers and integrator. What tests should I run for a non-BFB PR? Note that I am feeling very comfortable with this because it is BFB with the existing alternative DDPDD in my E3SM test simulations, as well as being BFB when changing from integer*8 to integer*4 in my standalone tests. The agreement with the DDPDD algorithm implies improved accuracy (getting the last significant bit correct more often) compared to the original algorithm.

Note that comments were updated to reflect the code changes. A new timer was also added to capture the cost of the final summation, to address @ambrad 's question in the github issue discussion. I did not put any timing data in the PR description, but I can provide them if there is interest.

amametjanov:
`e3sm_integration` results from the current HEAD of this branch (e11ce2b6e8) are at https://my.cdash.org/viewTest.php?buildid=2311359 .
Pat, please take a look at the errors/fails.

worleyph:
@az, thanks. There are 66 passes, 37 diffs, and 7 fails. I'm looking at the fails now (all seg. faults, apparently in reprosum, though I will need to chase this down). Please remind me - what are the different test types? Is there a query function to describe them? For example, all of the seg. fault fails are SMS_D tests (and all but two SMS_D tests have seg. faults).

worleyph:
@amametjanov : okay. '_D' is debug, and bounds checking caught an error. I'll look into diagnosing, and then push a fix.

rljacob:
Interesting that the FIDEAL case had no diffs.  That EAM with "ideal" physics.

worleyph:
@amametjanov , I ran the e3sm integration test on Chrysalis. Now there are 44 DIFFs and 66 PASSes (and no FAILs). Is there a way that I could upload this to my.cdash.org, to make it easier to examine? Also, is the version of master that I checked out for this PR still BFB with the MASTER baselines used in the compares? Thanks.

amametjanov:
```
<test-root-dir> $ cime/CIME/Tools/wait_for_tests  ./*<test-id>*/TestStatus --no-wait -b e3sm_integration_intel_pr5560_geab5db2
```
There should only be NML-diffs (i.e. still BFB) between current `master` baselines and the fork of this branch from Mar 27.

rljacob:
@worleyph click on the number "44" under "Fail" and you'll see the list of tests.  Click on each name to see the TestStatus.log output.  The "DIFF" string under "Completed" for each test tells us the only issue was baseline diffs.  If the test failed in any other way, it would say "FAIL" instead of "DIFF".

worleyph:
@rljacob , sorry - guess I was not clear. I can't find where the uploaded data is on the dashboard. All I can find is the results of the e3sm integration suite that Az ran this morning.

worleyph:
Looking at the data manually, all of the PASSes are marked as "RUN" experiments, e.g.

     PASS ERIO.ne30_g16_rx1.A.chrysalis_intel RUN

while all of the DIFFs are marked as "(phase BASELINE)" experiments, e.g.

     DIFF SMS_Ln9.ne4_oQU240.F2010.chrysalis_intel.eam-outfrq9s (phase BASELINE)

Guess that that makes sense. Only baseline comparisons will note the non-BFB behavior compared to current master. I can check the magnitudes of the diffs - how many do I need to check to be reassured that nothing unexpected is going on?

amametjanov:
Here is the new upload: https://my.cdash.org/viewTest.php?buildid=2311696 .

worleyph:
@amametjanov , thank you.

@rljacob , how do I evaluate the output of the e3sm integration experiments to determine that the results are consistent with round-off level changes? Are there any tools to postprocess the xxx.cprnc.out files?

ambrad:
One thing you might consider, based on your observation above re: DDPDD, is creating baselines using DDPDD and then seeing whether there are still 44 DIFFs against these.

worleyph:
@ambrad , thank you for the suggestion. I have just now tested this for a single test that is a DIFF when compared to master (just to make sure that I have the process correct), and it was a PASS. Since the DDPDD and the integer vector algorithm share essentially zero code (and use completely different approaches), hopefully this will be sufficient if they are all BFB. I'll start this process tomorrow.

worleyph:
I checked out the latest version of master (last night), modified shr_reprosum_mod to always use the DDPDD option, then created baselines for the E3SM integration test suite. I then ran the e3sm integration test suite with the PR, comparing with these new baselines:

     -b generate_ddpdd_baselines

When compared against unmodified master, the PR had 44 tests with DIFFs. When compared against 'generate_ddpdd_baselines', there were 16 DIFFs (and all others were PASS, except for 4 NLFAILs due to using the most recent version of master, and these all had BASELINE PASS results).

Note that there is no expectation that the new algorithm will be BFB with the DDPDD algorithm. I have just observed that it seems to occur more often. That appeas to hold true in these experiments as well.

I'll look at a few of these DIFFs, and compare the cprnc.out between that for the unmodified master baselines and for the ddpdd baselines, to see if anything looks peculiar. 

Anything else I should do? What do others do when submitting a non-BFB PR?

rljacob:
Usually we assert by the specifics of the code changes that this non-BFB won't change the climate statistics.  If there's any doubt, there a "non-BFB" test suite that will explicitly test that assumption but its undergoing some maintenance right now.   After that, the only thing left is to do a 10-year fully coupled run and have experts look at the diagnostics compared to a baseline.   I think you're fine and this can proceed.

worleyph:
Just to be safe, I reran the comparison of the new shr_reprosum_mod default algorithm against the DDPDD algorithm using the same checkout of master (the version I checked out last night and used to generate the DDPDD baselines). Glad I did so, as this time there were 111 PASSes and only 1 DIFF (so the other 15 DIFFs in the previous comparison were due to non-BFB differences between the two versions of master).

I consider this to be a reasonable validation that the new algorithm is a round-off level change and not a buggy.

The test with the DIFF was

     SMS.ne4_oQU240.F2010.chrysalis_intel.eam-thetahy_sl

in 

     SMS.ne4_oQU240.F2010.chrysalis_intel.eam-thetahy_sl.C.20230330_193732_sr89md/

in 

     /lcrc/group/e3sm/ac.worleyph/scratch/chrys

ambrad (approved):

sarats (approved):

philipwjones (approved):
Approve based on Pat's testing and quick visual pass

rljacob:
@worleyph I appreciate the detailed explanation in the PR description but do you have a short description of these changes we can use in the merge commit message?

worleyph:
@rljacob , How about (or do you need something even shorter?) ...

"The default reprosum algorithm converts each floating point summand into a vector of integers representation. These vectors are summed locally first, and then globally (using an mpi_allreduce). All of this is exact and reproducible. The final step is to convert the final sum, in the vector of integers representation, back into a floating point value.

This final sum begins by converting each element of the integer vector into a floating point value, and then adding these, smallest to largest. Round off can occur in the least significant digit of the final result, and whether round off occurs is sensitive to the particular floating point summands. The numbers of MPI tasks and OpenMP threads can affect the length of the integer vector, and thus the floating point summands. In consequence, reproducibility with respect to degree of parallelism is not guaranteed. (The situation in which reproducibility is lost has never been observed in E3SM simulations, but it cannot be ruled out.)

This PR modifies the final summation so that the same set of floating point summands is generated independent of the number of MPI tasks and number of OpenMP threads, establishing reproducibility."

worleyph:
@rljacob , @ambrad , I poked at the one test that DIFFed:

     SMS.ne4_oQU240.F2010.chrysalis_intel.eam-thetahy_sl

repeating the test without comparing with the precomputed baselines,  twice, once with the updated reprosum algorithm and once with DDPDD, then comparing these directly. This direct comparison was BFB (looking at cpl.hi.0001-01-06-00000.nc and atm.log and comparison of each output from shr_reprosum_calc call using an instrumented version of shr_reprosum_mod.F90).

I then compared the cpl.hi.0001-01-06-00000.nc from the new DDPDD run with the cpl.hi.0001-01-06-00000.nc  in 

     /lcrc/group/e3sm/baselines/chrys/intel/generate_ddpdd_baselines/SMS.ne4_oQU240.F2010.chrysalis_intel.eam-thetahy_sl/

and there were DIFFs, even though this was for the same checkout of master. So, there appears to be some nondeterminancy in this test on Chrysalis?

ambrad:
> there appears to be some nondeterminancy in this test on Chrysalis

For context, this test has been running in e3sm_developer on various platforms since approximately March 2020, so approx. three years. Of course, it's possible your code changes have revealed some nondeterminism, but I suggest that you analyze this a bit more before concluding that. In any case, if there _is_ nondeterminism, a clear reproducer will be helpful; simply running it a few times in a row I think should have a nearly 0 probabillity of showing different answers given that it has run for three years without a problem.

worleyph:
@ambrad ,

     For context, this test has been running in e3sm_developer on various platforms since approximately March 2020, so approx. three years.

Good to hear. Perhaps there was something weird that occurred using the DDPDD algorithm on Chrysalis. (For example, I saw a hang in the MPI_Allreduce in DDPDD in an experiment two weeks ago on Chrysalis that I was not able to reproduce the next day.) SInce you do not run it this way, probably not worth pursuing since the default algorithm is tested regularly. Since I got a BFB result for this remaining test when I reran, I'm comfortable with the results relevant to this PR. Thanks.

rljacob:
Is this still a non-BFB PR ?

worleyph:
@rljacob, It is still a non-BFB PR. (It is just BFB with the alternative DDPDD algorithm on Chrysalis for the E3SM integration test suite.)

Issue #5295, opened 2022-11-12
FP exception after adding -ffpe-trap=invalid in tests like `SMS_D.ne4_oQU240.F2010`


ndkeen:
If we add `-ffpe-trap=invalid` as a compiler flag to fortran builds, I see this error:

```
34: Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.
34:
34: Backtrace for this error:
34: #0  0x151f5ff5cd6f in ???
34: #1  0x108ba5f in __clubb_intr_MOD_clubb_tend_cam
34:     at /global/cfs/cdirs/e3sm/ndk/repos/ndk_machinefiles_pm-cpu-add-gnu-debug-invalid/components/eam/src/physics/cam/clubb_intr.F90:2901
34: #2  0x13860a7 in tphysbc
34:     at /global/cfs/cdirs/e3sm/ndk/repos/ndk_machinefiles_pm-cpu-add-gnu-debug-invalid/components/eam/src/physics/cam/physpkg.F90:2544
34: #3  0x13ab1f1 in __physpkg_MOD_phys_run1._omp_fn.0
34:     at /global/cfs/cdirs/e3sm/ndk/repos/ndk_machinefiles_pm-cpu-add-gnu-debug-invalid/components/eam/src/physics/cam/physpkg.F90:1082
34: #4  0x151f60783a55 in gomp_thread_start
```

The line of the fail is here:
`where (kbfs .eq. -0.0_r8) kbfs = 0.0_r8`

which might mean a value in array `kbfs` is NaN.

Affects these tests:
```
ERP_D_Lm3.ne4_oQU240.F2010.pm-cpu_gnu
ERS_D.ne4_oQU240.F2010.pm-cpu_gnu.eam-thetahy_ftype2
ERS_D.ne4_oQU240.F2010.pm-cpu_gnu.eam-thetanh_ftype2
ERS_D_Ld31.ne4_oQU240.F2010.pm-cpu_gnu
ERS_D_Ld5.ne4_oQU240.F2010.pm-cpu_gnu.eam-rrtmgpxx
SMS_D.ne4_oQU240.F2010.pm-cpu_gnu.eam-cosplite
SMS_D.ne4_oQU240.F2010.pm-cpu_gnu.eam-preqx_ftype0
SMS_D.ne4_oQU240.F2010.pm-cpu_gnu.eam-preqx_ftype1
[...]

PR #5561, opened 2023-03-27
Add an init-to-zero for kbfs and remove unecessary where statement

ndkeen:
Add `kbfs=0` to init array to zero before use.
Remove a where statement that was looking for negative zeros (`-0`) and does
not look necessary.

After this change (and other recent PR's), should be able to add the compiler
check for floating invalids in DEBUG builds.

Fixes https://github.com/E3SM-Project/E3SM/issues/5295
[bfb]

### DIFF
--- a/components/eam/src/physics/cam/clubb_intr.F90
+++ b/components/eam/src/physics/cam/clubb_intr.F90
@@ -2906,6 +2906,7 @@ end subroutine clubb_init_cnst
    enddo
 
    ! diagnose surface friction and obukhov length (inputs to diagnose PBL depth)
+   kbfs = 0._r8
    do i=1,ncol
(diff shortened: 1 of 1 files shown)


ndkeen:
With this change on next (on March 27th), I'm testing against baselines on pm-cpu.
`create_test -c -b master e3sm_integration`
Could also do similar test on chrysalis or cori-knl with Intel.

singhbalwinder (approved):
Looks good! It's great to see that the code is BFB after this change.

singhbalwinder:
Feel free to assign it to me.

quantheory (approved):
Looks good. Thanks for doing this!

ndkeen:
I also tested on chrysalis, running e3sm_developer with intel and they are BFB with baselines.

PR #5562, opened 2023-03-27
Fixes st_archive jobs taking more than 1 node

jasonb5:
Fixes `st_archive` jobs taking more than 1 node.

CIME overrides a jobs number of nodes with `tasks_per_node / task_count`, both must be defined.

[BFB]


jgfouca (approved):
Nice!

rljacob:
@golaz note this bug fix.

rljacob:
@jgfouca please merge.

rljacob:
For your case, you can do:  ./xmlchange --subgroup case.st_archive walltime=60:00

The global default is set in  E3SM/cime_config/config_workflow.xml

PR #5565, opened 2023-03-28
Fixes the initialization of a variable

### LINKED ISSUE #5526: Performance bug in subroutine InitCold during array init
It looks like there is a performance bug where an entire array is initialized to 0 every time through the loop.

in `components/elm/src/biogeophys/CanopyStateType.F90`:
```
  do c = bounds%begc, bounds%endc
     l = col_pp%landunit(c)

     if (lun_pp%itype(l) == istsoil .or. lun_pp%itype(l) == istcrop) then
        this%alt_col(c)               = 0._r8 !iniitialized to spval for all columns                                                                                                                               
        this%altmax_col(c)            = 0._r8 !iniitialized to spval for all columns                                                                                                                               
        this%altmax_lastyear_col(c)   = 0._r8 !iniitialized to spval for all columns                                                                                                                               
        this%alt_indx_col(c)          = 0     !initiialized to huge  for all columns                                                                                                                               
        this%altmax_indx_col(c)       = 0     !initiialized to huge  for all columns                                                                                                                               
        this%altmax_lastyear_indx_col = 0     !initiialized to huge  for all columns  <-- here -- setting entire array to 0. should probably be (c)                                                                                                                   
     end if
  end do
```

bishtgautam:
Only the c-th index of `altmax_lastyear_indx_col` is set to zero within 
the do-loop.

Fixes #5526
[BFB]

### DIFF
--- a/components/elm/src/biogeophys/CanopyStateType.F90
+++ b/components/elm/src/biogeophys/CanopyStateType.F90
@@ -499,12 +499,12 @@ contains
        l = col_pp%landunit(c)
 
        if (lun_pp%itype(l) == istsoil .or. lun_pp%itype(l) == istcrop) then
-          this%alt_col(c)               = 0._r8 !iniitialized to spval for all columns
-          this%altmax_col(c)            = 0._r8 !iniitialized to spval for all columns
-          this%altmax_lastyear_col(c)   = 0._r8 !iniitialized to spval for all columns
-          this%alt_indx_col(c)          = 0     !initiialized to huge  for all columns
-          this%altmax_indx_col(c)       = 0     !initiialized to huge  for all columns
-          this%altmax_lastyear_indx_col = 0     !initiialized to huge  for all columns
+          this%alt_col(c)                  = 0._r8
+          this%altmax_col(c)               = 0._r8
+          this%altmax_lastyear_col(c)      = 0._r8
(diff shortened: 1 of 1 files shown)


ndkeen (approved):

components/elm/src/biogeophys/CanopyStateType.F90
```diff
@@ -499,12 +499,12 @@ subroutine InitCold(this, bounds)
        l = col_pp%landunit(c)
 
        if (lun_pp%itype(l) == istsoil .or. lun_pp%itype(l) == istcrop) then
-          this%alt_col(c)               = 0._r8 !iniitialized to spval for all columns
-          this%altmax_col(c)            = 0._r8 !iniitialized to spval for all columns
-          this%altmax_lastyear_col(c)   = 0._r8 !iniitialized to spval for all columns
-          this%alt_indx_col(c)          = 0     !initiialized to huge  for all columns
-          this%altmax_indx_col(c)       = 0     !initiialized to huge  for all columns
-          this%altmax_lastyear_indx_col = 0     !initiialized to huge  for all columns
```
sarats, line 507: Oh only last line (507) had the semantics changed. I got confused why removing comments would fix init.

rljacob:
This is ready.

PR #5573, opened 2023-04-04
Fixes for EAM cube_to_target topography tool

brhillman:
Fixes for EAM cube_to_target topography tool to allow configuring via command line arguments and to prevent aborting when encountering weights outside thresholds with very coarse grids. Allowing configuring via command line arguments enables using a different input USGS raw topo file, which will allow using higher resolution binned topo data for SCREAM, and also allows using lower resolution binned topo data for coarse resolutions to speed up topo generation with very coarse grids (ne2, ne4).

[BFB]


brhillman:
Thanks @PeterHjortLauritzen! We are not using that yet, I will definitely check it out.

brhillman:
Some issues came up in testing; closing until I can investigate further.

PR #5575, opened 2023-04-04
Remove ocean unused variable RediKappaData

### LINKED ISSUE #5574: Ocean hangs on start-up during MPI communication
This occurs for EC30to60 performance test Chrysalis, Perlmutter, and Chicoma with Gnu and OpenMPI. There is no error message, but the log file indicates that the hand is during initialization and reading of variables. See detailed descriptions at https://github.com/MPAS-Dev/compass/issues/500#issuecomment-1480119214.

So far this occurs in stand-alone tests only, but it could possibly happen in E3SM runs.

mark-petersen:
The variable `RediKappaData` exists in the Registry file but is not used in the code. It appears that the gnu compiler removes some underlying information about the array in optimized mode, and then the MPI communication hangs when it tries to communicate the size of the array. When the array is used the simulation no longer hangs on start-up. 

The variable `RediKappaData` was meant for the option of entering a spatially-variable data field from input. Since it has not been used for three years, we should remove it. If this feature is needed in the future, it can be added in properly and tested at that time.

Fixes #5574 
[BFB]

### DIFF
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -3783,10 +3783,6 @@
 			 description="Multiplication factors to smooth ssh at coastlines for SAL caculation"
 			 packages="tidalPotentialForcingPKG"
 		/>
-		<var name="RediKappaData" type="real" dimensions="nCells" units="m^2 s^-1"
-			 description="Redi isopycnal mixing Kappa value. This is the data field specified at init."
-			 packages="gm"
-		/>
 	</var_struct>
 	<var_struct name="timeVaryingForcing" time_levs="1">
 		<var name="windSpeedU" type="real" dimensions="nCells Time" units="m s^-1"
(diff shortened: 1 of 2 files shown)


mark-petersen:
Tested this PR with the ocean stand-alone compass test `ocean/global_ocean/EC30to60/PHC/performance_test`. I can confirm that it does pass on chrysalis with gnu-optimized (ran on 36 cores). This set-up hangs on the head of master.

xylar:
OMG, thanks for figuring this out @mark-petersen!  I'll test this on all the configurations that were giving us trouble as soon as I can.

xylar (approved):
Wonderful, @mark-petersen!

I can confirm that this fixes the issue I have been seeing with the `pr` suite on Chrysalis with Gnu compilers.  With this fix, the suite completes, whereas without it, it hangs on the EC30to60 performance test.

I will also test on Perlmutter (when available) and Chicoma (tomorrow) but I think the fix on Chrysalis alone is enough to justify approving this change.

xylar:
Unfortunately, this doesn't seem to completely solve the issues on Perlmutter or Chicoma.  I'm still seeing PIO errors (but only in optimized runs).  See, for example:
```
/pscratch/sd/x/xylar/compass_1.2/test_20230412/pr_rm_redi/ocean/global_ocean/EC30to60/PHC/performance_test/forward
```
I'm having trouble with the filesystem on Chicoma so tests there are less conclusive but I'll update here as soon as I have one that definitively indicates whether the problem there is still present for this branch.

jonbob:
passes:

merged to next

PR #5576, opened 2023-04-05
Add initial superBFB tests for OCN

amametjanov:
Add initial superBFB tests for OCN

[BFB]

### DIFF
--- a/cime_config/testmods_dirs/pemod/omp1/shell_commands
+++ b/cime_config/testmods_dirs/pemod/omp1/shell_commands
@@ -5,9 +5,9 @@ if [ `./xmlquery --value MACH` == chrysalis ]; then
   ./xmlchange MAX_MPITASKS_PER_NODE=64
 fi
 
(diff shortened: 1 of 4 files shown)


amametjanov:
I've checked on Chrysalis, but this should work on any machine with
```
./cime/scripts/create_test e3sm_superbfb_ocn
```

jonbob:
@amametjanov - what does superBFB mean?

rljacob:
https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/3438608385/E3SM+Code+Review+and+New+Feature+Process#Super-BFB-suite

jonbob (approved):
Approved based on visual inspection and use of current meshes for testing

jonbob:
@amametjanov - the only issue you may run into is that CIME doesn't download files for case2 runs of these ERS and PEM tests. So you may see unexpected failures due to missing graph files and have to stage some by hand

Issue #5319, opened 2022-11-19
Use of threadprivate module data causing issues with GNU (Perlmutter, Summit) 


mt5555:
With the GNU compiler on Perlmutter and Summit, there is an issue with threading when the number of threads used is not identical in all parallel regions.  Static thread private data is only preserved if all parallel regions use the same number of threads.  The atmosphere makes heavy use of this feature.

This issue was identified in:  PR #5191   with a bunch of public static (module data) allocatable arrays in modal_aero_wateruptake.F90
quoting Walter from that issue:   "While working get a better load balance on some larger runs (~1k nodes) I was advised to reduce the thread count on the ocean and ice components to 2, but I still used 7 threads in the atmosphere. It turns out this causes previously allocated module variables in the atmosphere to be deallocated on 5 of the 7 threads that were not used for the sea ice model."
 
This same issue also occurs in low-res F cases, where the problem is with CLUBB's grid_class.F90::gr public module variable.

This may be correct compiler behavior ( since common blocks are similar to public module data):
https://www.intel.com/content/www/us/en/develop/documentation/fortran-compiler-oneapi-dev-guide-and-reference/top/language-reference/a-to-z-reference/t-to-z/threadprivate.html
"For subsequent parallel regions, data in THREADPRIVATE common blocks are guaranteed to persist only if the dynamic threads mechanism has been disabled and if the number of threads are the same for all the parallel regions."
(dynamic threads are not enabled by E3SM, but not sure if they are disabled by default).  

The easiest way to trigger this is to run (for example) an F case with all components using 8 threads, but
have one component using 2 threads.

A more subtle way to trigger this bug is to have more threads than there are elements per MPI task in the atmosphere.  If one uses 8 threads and has only 6 elements per MPI task, then the dycore will have parallel regions with 6 threads, while the physics
will use all 8 threads.

A mixture of components using 8 threads and 1 threads seems to run and preserves threadpriviate module data.

Mixing components with 8 and 2 threads seems to work if the nodes used by the 8 thread components are disjoint from the nodes
used by the 2 thread components (consistent with the expected behavior).

We should construct a small test code to verify this.

Summit:  GNU 9.1(I think)    crashes
Perlmutter:  GNU 11.2        crashes 
Compy:  Intel 19.05          works
Anvil: GNU 8.2               works
Anvil: Intel 20.04           works
Chrysalis:
   GNU 9.2.0                 not yet tested
   GNU 9.3                   not yet tested   
[...]

PR #5577, opened 2023-04-05
Convert thread private variables to allocatable

whannah1:
Thread private variables in certain aerosol related code was causing problems for coupled MMF runs, but this is a more general problem that limits the flexibility of the PE layout because these variables become unintentionally unallocated other model components like the ocean are using a different number of hardware threads. So this converts those problematic variables to allocatables within each routine that are allocated and deallocated every time they are needed. Performance does not appear to be impacted.

[BFB]

### DIFF
--- a/components/eam/src/chemistry/utils/modal_aero_wateruptake.F90
+++ b/components/eam/src/chemistry/utils/modal_aero_wateruptake.F90
@@ -33,6 +33,7 @@ public :: modal_aero_wateruptake_reg
 real(r8), parameter :: third = 1._r8/3._r8
 real(r8), parameter :: pi43  = pi*4.0_r8/3.0_r8
 
+real(r8), parameter :: huge_real = huge(1.0_r8)
 
 ! Physics buffer indices
 integer :: cld_idx        = 0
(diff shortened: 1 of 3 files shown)


sarats:
> Thread private variables in certain aerosol related code was causing problems for coupled MMF runs

Is there a related issue? If not, can you elaborate? 

Curious: How frequent are the alloc/deallocs? I presume your PM tests used OMP threads.

whannah1:
@sarats this is related to issue #5319 and the previously closed PR #5191.
The allocations should occur once every time we call tphysbc.

I thought my PM tests were using threads, but I just checked and they were not using threads. I'm not sure how that would change the performance comparison, but I'm happy to rerun them with threads if you think that will make a difference.

mt5555:
I think we need to be careful about these kind of changes.  Allocate statements can be very expensive, especially inside a threaded region (where they force thread synchronization).  This is explained in detail:

https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/128585462/Fortran+Performance+Best+Practices

in the "array allocation" section.   

Some compilers may have special libraries and can do this quickly.  allocating once per physics timestep is probably not a big deal, but no reason to do more of this. (it wasn't clear if this PR was moving allocate statements around, or actually producing more allocate statements)  If these are thread private - why cant the be put on the stack?  that is super efficient and avoids calling the system allocation/deallocate routines.

whannah1:
@mt5555 I don't really understand your comment about moving them to the stack. I'm just trying to fix a problem that I keep encountering when running coupled MMF experiments. I think we haven't hit this previously because we don't tend to vary the number of threads across the surface components. 

It sounds like you don't trust the performance results I got on Perlmutter? Is there a specific test I could run on a specific machine that would address your concern?

singhbalwinder:
Frequent allocation/deallocation of large arrays is slow but for us, the real issue was memory fragmentation. Allocation/deallocation during a time step was still manageable but the real problems showed up when these calculations were invoked multiple times in a time step for radiation diagnostics. Please take a look at issue [#2575](https://github.com/E3SM-Project/E3SM/issues/2575). The model would run for a month or a little more and crash due to unavailable memory. I resolved this issue by allocating these arrays only once and using threadprivate. It is not ideal to use threadprivate as it forces the threads to synchronize but this solution avoids crashes due to fragmented memory. This PR is almost like reverting those changes and will likely cause memory fragmentation when we turn on the radiation diagnostics.

As expected the problem will be more pronounced if we use fine resolution. ne4 tests never really caused the model to crash due to memory fragmentation with radiation diagnostics turned on. I think longer runs with ne30 or higher resolutions might be more useful to reveal these issues.

mt5555:
is it possible to put these on the stack?  they then need to be local subroutine variables, in the top level subroutine called after threads have started, and then passed around to all other subroutines.  if such a structure is possible, this would solve both issues (fragmentation and not-surviving-change-of-nthreads) and avoid alloc/dealloc system calls.  but might require a lot of code changes.

singhbalwinder:
The size of these arrays depends on the number of MAM modes, which is a compile-time constant. We are currently treating this number as a runtime constant, so we're allocating these arrays at runtime rather than declaring them on the stack. I think we can declare them on the stack but it may take a lot of code changes.

mt5555:
another option (used by HOMME), is to add an additional index representing the thread number, and allocate them at startup in the serial region.  Then in the parallel region, each thread access them using it's thread index.  

Also a non-trivial amount of code changes :-(

whannah1:
Since there's no good/easy solution to this problem I'm going to close this PR and just manually apply these changes to new branches for future coupled MMF simulation campaigns.

rljacob:
@whannah1 you never said what the problem was when running MMF coupled.  That would help.

PR #5578, opened 2023-04-05
Expose MMF climate tuning parameters for SAM++

whannah1:
This PR adds ifdef blocks to expose climate tuning parameters for SAM++. It is preferable to expose these as namelist variables, but in this case we don't expect SAM++ to be used for many coupled climate experiments due to the unsophisticated microphysics and prescribed aerosols. Also, exposing namelist values will require a lot more changes, so ifdefs will simplify getting the MMF coupled experiments running again. 

[BFB]


brhillman (approved):

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #5581, opened 2023-04-05
Update CIME submodule

jgfouca:
... to 9a12fd20a6af1da27d128bb4710f7f41c36363ad

Changes:
1) jenkins_generic_job: Add support for queue selection
2) Loosen the XSD xml specification for config_machines.xsd to be consistent with env_mach_specific.xsd
3) minor updates to allow moab driver
4) should not test timestep of stub components
5) Adding support for ADIOS for E3SM !!!

Fixes:
1) bless_test_results: Fix mistake in re match counting
2) replay.sh: fixes restoring quotes to command arguments
3) create_test:  fixes single_exe create_test option
4) check_input_data: Expand SharedArea coverage to entire implementation

[BFB]


sarats:
+1 on replay.sh fix.

create_test: fixes single_exe - guess this is from the single submit stuff?

> Loosen the XSD xml specification for config_machines.xsd 

Anything we need to be aware of that's possible now than before?

jgfouca:
@sarats ,

> create_test: fixes single_exe - guess this is from the single submit stuff?

Single exe != single submit. Single exe means do a single build for all the tests cases you are asking create_test to do. Case dirs will be made for all the cases but they will all share a single build. Single submit means run all cases under a single batch submission. Single exe has been offered for a while and it should be expected to work if all the cases are compatible.

Single submit is on the longterm wish list for CIME and may never happen. I looked it doing it ages ago and endless complexities came up having to do with the various machines/batch systems and how to micro-manage resources within a single allocation. It's possible we could offer it only for certain "well-behaved" platforms but there would have to be some significant upside for it to be worth taking on. @jasonb5 may have some thoughts here.

> Anything we need to be aware of that's possible now than before?

I don't think so. It looks like the command elements (the elements that go inside `<modules>`) got explicitly set to a string type, which they should have been anyways.

jgfouca:
single_exe is a use-at-your-own-risk feature. You are telling create_test that all the cases can share an executable and there's no way for create_test to validate this. We also support creating test suites that are set to share a single exe. All the cases in a suite will continue to share a single exe even if that suite is aggregated/inherited into a bigger suite like e3sm_developer.

sarats:
That makes sense, thanks!

rljacob (approved):

PR #5582, opened 2023-04-05
SCREAM upstream merge to fix SCREAMv1 build

brhillman:
Bring in changes from downstream SCREAM repo to fix SCREAMv1 build errors.

[BFB]


rljacob:
Just as a sanity check, does the number of commits seem right?  This should be development since the last SCREAM merge in #5416 .   That had all scream commits up to mid-February.  Does 365 commits since then make sense?

AaronDonahue:
@rljacob , I think that 365 is in the ball-park of my back-of-envelope calculation.  Using `git rev-list --count` I calculated `53735` commits for the date of the repo in this merge and `53189` commits for the date of the repo in #5416 .  Note that we have also been commiting upstream merges to the scream repo, two of them by my count each w/ 170 and 52 commits in them respectively.  So all told that would point to 324 commits difference.  Which is 41 commits shy, but then again this is back-of-the-envelope and I may be double-counting some things in the upstream merge.

cime_config/machines/cmake_macros/gnu_docker.cmake
```diff
@@ -0,0 +1,68 @@
+set(AR "/opt/conda/bin/x86_64-conda-linux-gnu-ar")
```
AaronDonahue, line 1: This shouldn't hold up the upstream merge, but I'm making a note to myself and @elynnwu that we should add some comments to this file now that it will also be in the general E3SM code base.  I could easily see some confusion amongst users as to what the docker cmake file is for and why it only works on two machines for now.

AaronDonahue (approved):
I had some comments, but I mostly made them to ping folks who would be interested.  They shouldn't hold up approval though.

.gitmodules
```diff
@@ -74,3 +74,9 @@
 	path = components/ww3/src/WW3
 	url = git@github.com:E3SM-Project/WW3.git
 	branch = e3sm
+[submodule "externals/haero"]
```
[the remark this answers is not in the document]
jeff-cohere: Thanks, @AaronDonahue . I don't think adding submodules creates any problems as long as fetching the submodules doesn't give anyone any grief on a machine of interest.

rljacob:
Is this BFB for the rest of E3SM?

cime_config/machines/cmake_macros/gnu_docker.cmake
rljacob: What is this docker cmake for?  We already have docker capability in E3SM but its under the "singularity" machine.

rljacob (approved):

brhillman:
> Is this BFB for the rest of E3SM?

Yes, should be.

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #5583, opened 2023-04-06
Add OpenACC directives in the semi-implicit barotropic mode solver

hyungyukang:
The semi-implicit barotropic solver in E3SM ocean component has been ported on GPU by adding OpenACC directives.

Running and testing have been done for the stand-alone MPAS-O on Summit only at this moment, and how to run and test on Summit is documented in E3SM Confluence: https://acme-climate.atlassian.net/wiki/spaces/OO/pages/3718087262/Running+MPAS-Ocean+with+a+GPU-accelerated+semi-implicit+barotropic+mode+solver+option

To compile MPAS-O with MAGMA or CUBLAS for GPU-accelerated linear algebras, some additions (`USE_MAGMA`, `USE_CUBLAS`) have been made to `mpas-framework/Makefile`

A subroutine `si_halo_exch` is added in the semi-implicit solver code, which is very experimental, to use GPU-aware MPI for local halo exchanges during solver iterations. It currently works on GPU but much slower than the data staging halo exchange. The subroutine will be further optimized. 

The semi-implicit barotropic mode solver is a stealth function, so it does not affect to default configurations.

### DIFF
--- a/components/mpas-framework/Makefile
+++ b/components/mpas-framework/Makefile
@@ -162,6 +162,9 @@ pgi-summit:
 	"CORE = $(CORE)" \
 	"DEBUG = $(DEBUG)" \
 	"USE_PAPI = $(USE_PAPI)" \
+	"USE_MAGMA = $(USE_MAGMA)" \
+	"USE_CUBLAS = $(USE_CUBLAS)" \
+	"USE_GPU_AWARE_MPI = $(USE_GPU_AWARE_MPI)" \
 	"OPENMP = $(OPENMP)" \
 	"OPENACC = $(OPENACC)" \
 	"USE_SHTNS = $(USE_SHTNS)" \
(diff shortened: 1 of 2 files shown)


components/mpas-framework/Makefile
```diff
@@ -700,6 +703,25 @@ ifeq "$(OPENMP_OFFLOAD)" "true"
 	LDFLAGS += $(LDFLAGS_GPU)
 endif #OPENMP_OFFLOAD IF
 
+ifeq "$(USE_MAGMA)" "true"
+   LIBS += -L$(MAGMADIR)/lib -lmagma
+   FCINCLUDES += -I${MAGMADIR}/include
+   override CPPFLAGS += "-DUSE_MAGMA"
+   override FFLAGS_ACC += -Dmagma_devptr_t='integer(kind=8)' -Mcudalib=cublas
```
sarats, line 710: Interesting to see `-Dmagma_devptr_t='integer(kind=8)` in FFLAGS. What's the story behind it?

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_si.F
```diff
@@ -55,6 +55,18 @@ module ocn_time_integration_si
    use ocn_surface_land_ice_fluxes
    use ocn_transport_tests
 
+#ifdef MPAS_OPENACC
+#ifdef USE_MAGMA
+   use magma
```
sarats, line 60: I do see some MAGMA sections (example: line 4295) use OpenACC directives. Do you need `use openacc` here or am I missing something?

sarats: Oh, it's because this is already conditional on MPAS_OPENACC #ifdef. 

In that case, do you need `use openacc` under USE_CUBLAS? I guess I was looking for symmetrical sections for both libs.

components/mpas-framework/Makefile
hyungyukang: @sarats , I was following MAGMA's instruction, but I just tried to compile without the flag and use 'integer(kind=8)' in the code directly, and it works. It'd be better to reduce complexity in the code and Makefile, so I will remove the flag!

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_si.F
hyungyukang: @sarats , Thanks for catching it! It doesn't need to be there, so I removed it!

grnydawn:
@hyungyukang, I will leave code review to whom knows this code well. My comment is about the list of git commits. It looks that there are many small git commits which will be shown in E3SM commit history. It may be good if you combine them as one or two commits before merging master branch.

hyungyukang:
> @hyungyukang, I will leave code review to whom knows this code well. My comment is about the list of git commits. It looks that there are many small git commits which will be shown in E3SM commit history. It may be good if you combine them as one or two commits before merging master branch.

@grnydawn , Agreed. There would be more commits, so I will combine all of my commits before merging into master. Thanks for your suggestions!

grnydawn (approved):
it looks good to me

philipwjones (approved):
This looks fine to me. Approve based on code inspection and Hyun testing.

jonbob:
passes:

merged to next

hyungyukang:
@sarats , @philipwjones, @grnydawn , and @jonbob , Thanks a lot for reviewing and testing and your approval!

jonbob:
merged to master

PR #5584, opened 2023-04-06
Add am eddy crossterms

alicebarthel:
I added the Velocity x Salinity cross-terms in the eddy analysis members, both for the total velocity, as well as the GMBolus velocity term. 

I was planning to add the normalMLEvelocity term, but would like this work-in-progress [WIP] PR to discuss the best way to do that: 
- for GM, the code block was repeated (for performance reasons, according to in-code [comments](https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-ocean/src/analysis_members/mpas_ocn_eddy_product_variables.F#L227))
- Do you want me to do the same for submeso (4 blocks with submeso on/off, GM on/off)? Having 4 different blocks starts to impede readability imo, but I also want to support good performance. Thoughts?


components/mpas-ocean/src/analysis_members/Registry_eddy_product_variables.xml
```diff
@@ -83,5 +104,10 @@
 			<var name="velocityMeridionalTimesTemperature_GM"/>
 			<var name="normalVelocityTimesTemperature"/>
 			<var name="normalVelocitySquared"/>
+			<var name="velocityZonalTimesSalinity"/>
+			<var name="velocityMeridionalTimesSalinity"/>
+			<var name="normalVelocityTimesSalinity"/>
+			<var name="velocityZonalTimesSalinity_GM"/>
+			<var name="velocityMeridionalTimesSalinity_GM"/>
```
vanroekel, line 111: Is there a reason to output both the `velocityZonalX` `velocityMeridionalX` and `normalVelocityX`?  I would guess we really only need one set or the other.

components/mpas-ocean/src/analysis_members/mpas_ocn_eddy_product_variables.F
```diff
@@ -10,8 +10,8 @@
 !  ocn_eddy_product_variables
 !
 !> \brief MPAS ocean analysis mode member: eddy_product_variables
-!> \author Mark Petersen
-!> \date   Oct 2016
+!> \author Mark Petersen; Alice Barthel
+!> \date   Oct 2016; Apr 2023
```
vanroekel, line 14: I would suggest we only need the last modified date in the docstring

components/mpas-ocean/src/analysis_members/Registry_eddy_product_variables.xml
vanroekel: Note also that to get these variables output in runs, changes are needed to 
```
components/mpas-ocean/cime_config/bldnml
```

vanroekel (changes requested):
@alicebarthel thanks for the changes.  Looks great, just a few comments for you.

alicebarthel:
@vanroekel Any thoughts on the GM/submeso loop (main message of the PR conversation)?

vanroekel:
Sorry to miss that main question @alicebarthel -- A couple thoughts

1. we could pull the GM and MLE calculations into their own loops that do the calculation only when those terms are enabled, I wonder if right now the GM one is computing on zeros when GM is off.
2. maybe we should just have one 'total' cross term and sum up normalVel + normalGMBolusVel + normalMLEBolusVel, having all the arrays maybe gets overly large?

I think we should at least do 1, but it may be good to have a bit of a discussion as to whether or not separate cross terms are important for each eddy term.

It would also be good to get input from @philipwjones on this question too.

philipwjones:
@alicebarthel and @vanroekel  If you want to do Luke's option 2 and combine velocities, that not only reduces storage but makes the code a bit easier. Just create a local velocity that's initialized with the normal velocity and add GM and MLE to that local velocity based on whether they're available. Then you'll just need one vT loop.

However, if you want to keep all the terms separate, I wouldn't worry too much about computing them all in a single
loop. While it will certainly be optimal to do everything at once, sometimes you have to sacrifice a little for clarity and I suspect the overall cost of this AM is not too high. You could try saving Tedge, Sedge during the first loop to re-use later, but that might actually be more expensive - not sure whether the indirect addressing or two additional arrays would cost more.

BTW, you can also remove the block loop - we only support a single block these days anyway. And you can avoid the mesh retrievals by using the mesh module instead.  I'll avoid ranting about the ridiculously long variable names and long line lengths that also impact readability...

components/mpas-ocean/src/analysis_members/Registry_eddy_product_variables.xml
alicebarthel: I think that historically we output both on centers and edges, to give people a choice between a quick visualization (centers) or a transect calculation (on edge). If the goal is to close budgets, I'd probably prioritize `normalVelocityX` but maybe that's just me.

alicebarthel:
@vanroekel Thanks for your thoughts on this! 
Re. 1, in the code, we have an (if GM) condition, which means the GM terms are only calculated when GM is on. The main point is that the non-GM calculation terms are copy-pasted within the 2 cases of the if condition. It might be ok for 2 cases but for 4 cases it sounds like a great set-up to forget to copy changes across any of the 4 cases, so I'd rather separate the terms in the relevant cases without repeating code. 
@philipwjones Thanks for your input! if it's not a performance killer to write the terms separately, I reckon the gain in readability is valuable enough. I may follow up with you on some of the other recommendations.

components/mpas-ocean/src/analysis_members/Registry_eddy_product_variables.xml
vanroekel: I think that is fair for normalVelocity and velocityZonal, velocityMeridional  and am not opposed to including both, but only if we actually think both will be useful.  My suggestion would be to keep both calculations, but only output the cell based quantities for eddy stats by default.  the edge quantities could be turned on for special simulations

vanroekel:
@alicebarthel I see.  I guess I did not look close enough.  I think I agree with what you are proposing to have 3 separate loops; one for mean quantities, one for GM, one for submeso and not copy paste.  I definitely agree that is a recipe for potential problems.

alicebarthel:
@vanroekel Re. 2: 
a) if the point is to calculate the Total transport term, then I could just modify the AM to use `normalTransportVelocity` instead of `normalVelocity`, right? (I actually thought `normalVelocity` *was* the total but apparently not). 
b) I thought the point of adding GM/MLE terms was to check the contribution of parameterized eddies to T, S transport. In which case, I would argue for: `normalTransportVelocityX`, `GMX` and MLEX`. If we have `normalVelocity` and `Temperature` in the monthly output then we can always recover the resolved eddy cross-correlation. I suppose we don't care to explore the cross-term split with the bolus velocities...

vanroekel:
> if the point is to calculate the Total transport term, then I could just modify the AM to use normalTransportVelocity instead of normalVelocity, right? (I actually thought normalVelocity was the total but apparently not).

I believe 

> a) if the point is to calculate the Total transport term, then I could just modify the AM to use `normalTransportVelocity` instead of `normalVelocity`, right? (I actually thought `normalVelocity` _was_ the total but apparently not).

Yes I think this is right, I believe `normalTransportVelocity` includes the GM and MLE contributions as this is the full advecting velocity.  



> b) I thought the point of adding GM/MLE terms was to check the contribution of parameterized eddies to T, S transport. In which case, I would argue for: `normalTransportVelocityX`, `GMX` and `MLEX`. If we have `normalVelocity` and `Temperature` in the monthly output then we can always recover the resolved eddy cross-correlation. I suppose we don't care to explore the cross-term split with the bolus velocities...

I think this sounds fine, but what I'm wondering is whether or not we will want these terms for every single run we do or just for specialized explorations.  I would be fine with the normalVelocity* quantities as the default output and skip the others.

alicebarthel:
@vanroekel I thought the point of an eddy product analysis member was to do a deep dive in eddy contribution/correlation terms. We have the total advection contribution in the tracer tendency budget, no? 
Is there something else in here (e.g. `SSHSquared`, `velocityXSquared`) that we need to output in all our runs? I.e. do you need two levels of eddy analysis (one on by default, the other reserved for specialized explorations)?
Sorry if this is turning into a broader discussion, I am happy to take it to a more appropriate place if need be.

vanroekel:
no problem @alicebarthel these are all good questions.  Happy to have it here.  I agree with you that this AM is for the eddy deep dive, but there is this fine line to walk of how much output can we have that people will actually look at, but we also don't want to strip so far that the simulations have no scientific value as well.  But as you suggest this is a much broader discussion that the whole team should have.  

For this PR I'd suggest we just copy what we do for the GM eddy stats for default output

alicebarthel:
I ran this new analysis member for one month on Anvil (default compiler), once with submeso OFF and one with submeso ON. It ran fine and included the expected outputs. 
For validation, I had a quick look at the spatial patterns of the T,S transport by the parameterization. A few points to highlight for @vanroekel:
- The MLE transport is generally about an order of magnitude higher than the GM component (in the top 500m). 
- The MLE term max drops off by an order of magnitude by 1000m (zero by 3000m). The deeper MLE values (e.g. 100-500m) are clearly in the Northern Hemisphere (likely driven by winter MLD - we are looking at January only), with the deepest MLE influence being just South of Iceland (Rockall Plateau) and the Irminger Sea. 
- The GM terms are less seasonal, especially at depth and don't drop off to zero. 
image
image
image
image

components/mpas-ocean/src/analysis_members/Registry_eddy_product_variables.xml
alicebarthel: @vanroekel I added the zonal and meridional components in the default, with the edge quantity calculated but not in the default output, as you suggested above.

alicebarthel:
@jonbob Could you let me know which additional tests I should run for this?
Also, could you run a performance check with this? I know I've added additional terms (in the calculation and streams output) so I hope it doesn't impact the performance too badly.

mark-petersen:
@alicebarthel thanks for adding this. The number of loops looks fine to me. It is readable, and keeps the branching outside the loops. I did a visual inspection and the additional code looks correct.  On the performance, you can post the time spent in the various AMs and the `time_integration` from the one month Anvil run above. That will give us an idea of the percent and comparison to other AMs. You can just paste the table from the end of the log file, if it ran to completion.

jonbob:
@alicebarthel - I don't think you need to run any other tests, unless you think it would help to have a one-month baseline without these changes for comparison. It would be good to document the impacts on performance and file size, but not required for this PR

components/mpas-ocean/src/analysis_members/mpas_ocn_eddy_product_variables.F
```diff
@@ -240,66 +258,106 @@ subroutine ocn_compute_eddy_product_variables(domain, timeLevel, err)!{{{
...
+                  normalMLEVelocityTimesTemperature(k,iEdge) = 0.5_RKIND*(activeTracers(index_temperature,k,cell1) + &
+                               activeTracers(index_temperature,k,cell2)) * normalMLEVelocity(k,iEdge)
+                  normalMLEVelocityTimesSalinity(k,iEdge) = 0.5_RKIND*(activeTracers(index_salinity,k,cell1) + &
+                               activeTracers(index_salinity,k,cell2)) * normalMLEVelocity(k,iEdge)
                end do
             end do
             !$omp end do
-            !$omp end parallel
 
             !$omp parallel
```
mark-petersen, line 313: @alicebarthel, this line needs to be deleted because each `!$omp parallel` needs to be paired with a `!$omp end parallel`. This one does not have a match. It's best to compile with `OPENMP=true` to catch these. The pairs `!$omp parallel` and  `!$omp end parallel` cat go around several do loops.

mark-petersen: It is line 313 that needs to be deleted.

mark-petersen (approved):
@alicebarthel I just rebased this code locally, made the OpenMP change above, and recompiled. I am approving based on your results above. Thanks!

jonbob:
@vanroekel -- are you OK with this PR at this point?

vanroekel:
@jonbob assuming @mark-petersen OMP change got included (not clear to me if that is integrated into this branch or not) I'm happy with this.  Once I get that confirmation I'll approve.

jonbob:
good catch, @vanroekel -- I don't think that got committed either. @alicebarthel -- can you make the change @mark-petersen found?

mark-petersen:
@jonbob I removed that line and confirmed that the code compiles with `OPENMP = true`. Compiled with intel debug on chrysalis and ran nightly test suite, including `ocean/global_ocean/QU240/WOA23/analysis_test`, which has `config_AM_eddyProductVariables_enable = .true.` so tests this change. All tests pass.

jonbob:
Thanks @mark-petersen -- I'll make a test merge and plan on getting this in next

vanroekel (approved):
approving based on visual inspection and final testing from @mark-petersen

jonbob:
passes:

merged to next

jonbob:
merged to master

PR #5586, opened 2023-04-06
Add MMF coupled compsets

whannah1:
This adds new coupled compsets that use the MMF atmosphere:
```
  WCYCL1950-MMF1
  WCYCL20TR-MMF1
  WCYCL20TRNS-MMF1    <= compset for testing (no ocean spinup)
```
This also updates the e3sm_mmf_integration suite, replacing the ERP test with one of the new coupled compsets. 

The PR is mostly BFB, but turns out to be non-BFB for the MMF SCM test because the time step was changed. We decided is a good change because the previous time step is unintentionally large and we want all MMF compsets using the same time step. 

[non-BFB] => only MMF tests


components/eam/src/physics/crm/samxx/samxx_const.h
```diff
@@ -231,11 +229,38 @@ real constexpr nzeros = 3.e6;   // Intersept coeff. for snow
 real constexpr nzerog = 4.e6;   // Intersept coeff. for graupel
 real constexpr qp_threshold = 1.e-8; // minimal rain/snow water content
 
-real constexpr qcw0 = 1.e-3;
-real constexpr qci0 = 1.e-4;
 real constexpr alphaelq = 1.e-3;
 real constexpr betaelq = 1.e-3;
 
+//------------------------------------------------------------------------------
```
brhillman, line 235: All these fixes should be in master now.

whannah1: ah, this was an accident due to how I created this branch from the previous PR branch. I don't think it will cause any conflicts.

brhillman (approved):

cime_config/tests.py
```diff
@@ -260,7 +260,7 @@
     #e3sm MMF tests for development
     "e3sm_mmf_integration" : {
         "tests" : (
-            "ERP_Ln9.ne4pg2_ne4pg2.F2010-MMF1.eam-mmf_fixed_subcycle",
+            "ERP_Ln9.ne4pg2_oQU480.WCYCL20TRNS-MMF1.allactive-mmf_fixed_subcycle",
```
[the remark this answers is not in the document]
whannah1: yup, good catch.

brhillman:
Merged to next.

PR #5591, opened 2023-04-08
MOSART-sediment: suspended sediment module

liho745:
MOSART-sediment, the suspended sediment module, takes lateral sediment yield fluxes from a 
land model (such as the soil erosion module in ELM) as the inputs and includes the description 
of riverine transport, erosion, and deposition processes. It also accounts for two major damming
 effects on suspended sediment load, direct sediment trapping effect and indirect flow regulating
 effect (via reservoir regulation of streamflow and hydraulic conditions). The scientific validation 
was conducted over the NLDAS2 domain (against observations from multiple USGS gages. The 
current parameter files are for the NLDAS2 domain only, although the process description is 
general and largely transferrable.

[NML]

### DIFF
--- a/components/mosart/src/cpl/rof_comp_esmf.F90
+++ b/components/mosart/src/cpl/rof_comp_esmf.F90
@@ -28,7 +28,7 @@ module rof_comp_esmf
   use RunoffMod        , only : rtmCTL, TRunoff, THeat, TUnit
   use RtmVar           , only : rtmlon, rtmlat, ice_runoff, iulog, &
                                 nsrStartup, nsrContinue, nsrBranch, & 
-                                inst_index, inst_suffix, inst_name, RtmVarSet, heatflag
+                                inst_index, inst_suffix, inst_name, RtmVarSet, heatflag, sediflag
   use RtmSpmd          , only : masterproc, iam, npes, RtmSpmdInit, ROFID
   use RtmMod           , only : Rtmini, Rtmrun
   use RtmTimeManager   , only : timemgr_setup, get_curr_date, get_step_size!, advance_timestep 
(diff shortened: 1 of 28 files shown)


bishtgautam (approved):

components/mosart/src/cpl/rof_comp_esmf.F90
```diff
@@ -733,10 +754,21 @@ subroutine rof_import_esmf( x2r_array, rc)
...
+
+       rtmCTL%qsur(n,nmud) = 0.0_r8
+       rtmCTL%qsur(n,nsan) = 0.0_r8                 
 
     enddo
 
+    if(sediflag) then
+        do n = begr,endr
+           n2 = n - begr + 1
+           rtmCTL%qsur(n,nmud) =  fptr(index_x2r_Flrl_rofmud,n2) * (rtmCTL%area(n)) ! kg/m2/s --> kg/s for sediment
```
donghuix, line 767: Do we need to account for the land fraction of the grid cell? For example, a fraction of the grid cell is land for coastal grid cell. Or does rtmCTL%area already account for the land fraction?

donghuix (approved):
A minor comment as I just think of if land fraction is needed to accounted for when getting erosion from land. I didn't see land fraction is considered when liquid runoff is transported from land to river.

components/mosart/src/cpl/rof_comp_esmf.F90
rljacob: You can ignore the _esmf.F90 file.  It is not used in E3SM.

liho745: In addition to what @rljacob put there, we are doing it similarly to rtmCTL%qsur and rtmCTL%qsub in this subroutine, i.e., not accounting for the land fraction here. @donghuix, you are welcome to investigate it further, but I'd suggest that it is beyond the scope of this PR, and let's go with what we have for now and have it merged into the master branch. In case your investigation does suggest we need to account for the land fraction, you may later issue another PR for it.

donghuix: Yes, I agree. It should be consistent with rtmCTL%qsur. I don't have other comments for this PR. Thanks.

sarats:
Looks like this causes problems with Cray compilers. I wonder why a "stealth" feature triggered integration test errors.
https://github.com/E3SM-Project/E3SM/issues/5677

PR #5596, opened 2023-04-12
Update alias names for MPAS-SI testing D-compsets

erinethomas:
update D compset alias names. These compsets are for MPAS-SI testing purposes only.
Modification to D-JRA compset aliases - should be 'DTEST**M**' in alias (to differentiate from D compsets that use CICE) 
[BFB]

### DIFF
--- a/components/mpas-seaice/cime_config/config_compsets.xml
+++ b/components/mpas-seaice/cime_config/config_compsets.xml
@@ -10,7 +10,7 @@
   <!-- E3SM D compsets -->
 
   <compset>
-    <alias>DTEST-JRA1p5</alias>
+    <alias>DTESTM-JRA1p5</alias>
     <lname>2000_DATM%JRA-1p5_SLND_MPASSI_DOCN%SOM_DROF%JRA-1p5_SGLC_SWAV_TEST</lname>
   </compset>
 
--- a/components/ww3/cime_config/config_compsets.xml
+++ b/components/ww3/cime_config/config_compsets.xml
@@ -25,7 +25,7 @@
   </compset>
 
   <compset>
-    <alias>DTEST-JRA1p5-WW3</alias>
+    <alias>DTESTM-JRA1p5-WW3</alias>
     <lname>2000_DATM%JRA-1p5_SLND_MPASSI_DOCN%SOM_DROF%JRA-1p5_SGLC_WW3%sp36x36_TEST</lname>
   </compset>
 


components/mpas-seaice/cime_config/config_compsets.xml
```diff
@@ -10,7 +10,7 @@
   <!-- E3SM D compsets -->
 
   <compset>
-    <alias>DTEST-JRA1p5</alias>
+    <alias>DTESTM-JRA1p5</alias>
     <lname>2000_DATM%JRA-1p5_SLND_MPASSI_DOCN%SOM_DROF%JRA-1p5_SGLC_SWAV_TEST</lname>
```
rljacob, line 14: Why is the word "TEST" in these?

@jgfouca what does _TEST at the end of the longname do?

proteanplanet: @rljacob I believe the rationale for "TEST" is to indicate the configuration is not scientifically validated.  And in the case for sea ice it is true - we are using this D-case configuration to ensure a robust sea ice code, but nothing more.  G-case simulations are typically used as the minimum scientific bar, and we are in the process of setting up some D-cases that are indeed comparable to published literature test cases, where we should not use the "TEST" alias. But the DTESTM compset shown here is not one of those.

proteanplanet (approved):
I am fine with this change.

components/mpas-seaice/cime_config/config_compsets.xml
jonbob: @rljacob -- the _TEST is used by the docn component in its SOM mode to pick a corresponding file:
````
  <entry id="DOCN_SOM_FILENAME">
    <type>char</type>
    <valid_values></valid_values>
    <default_value>UNSET</default_value>
    <values match="last">
      <value compset="_DOCN%SOM.*_TEST">pop_frc.1x1d.090130.nc</value>
      <value compset="_DOCN%SOMAQP" grid="oi%1.9x2.5">default.som.forcing.aquaplanet.Qflux0_h30_sstQOBS.2degFV_c20170421.nc</value>
      <value compset="_DOCN%SOMAQP" grid="oi%0.9x1.25">default.som.forcing.aquaplanet.Qflux0_h30_sstQOBS.1degFV_c20170421.nc</value>
    </values>
    <group>run_component_docn</group>
    <file>env_run.xml</file>
    <desc>Sets SOM forcing filename.
    This is only used when DOCN_MODE=som.</desc>
  </entry>
````
If it's missing from the compset, then no valid file is found and any test will abort with an UNSET filename. My understanding is that this has been in place for years, because the SOM mode requires more user knowledge to use in a scientifically valid way. So having TEST in the compset is a reminder of that

jonbob:
merged to master

PR #5606, opened 2023-04-19
For gcp12 machine, update PE layouts

ndkeen:
For gcp12 machine, update PE layouts (use new ice/ocn partitions).
Uses the same number of nodes as before, just more MPI's for certain cases in ice/ocn.

Also, slightly reduce number of tasks to build a single test from 24 to 20. 

[bfb]


ndkeen:
I ran e3sm_integration

sarich (approved):

ndkeen:
merged to next

rljacob (approved):

ndkeen:
This does seem to fix the issues I suspected it would, but there are new fails from other PR's that I'd like to see clear.

ndkeen:
I will merge to master as I think these are fine changes, but tessting on GCP keeps having (unrelated) issues preventing a clean bill of health.

PR #5607, opened 2023-04-19
gnu>=10 needs the -fallow-invalid-boz option

jgfouca:
Now that it's in the gnu.cmake, it doesn't need to be added in other locations like it had been.

The machine macro for spock will continue to have this option since it was adding those flags and it appears to be using an older gnu.

[BFB]


ndkeen:
I have been adding this flag in `Depends.gnu.cmake` as it only affects two sources.

```
  foreach(ITEM IN LISTS ALLOW_INVALID_BOZ_LIST)
    e3sm_add_flags("${ITEM}" "-fallow-invalid-boz") # avoids build error for integer, parameter :: gen_hash_key_offset = z'000053db'
  endforeach()
...
```

a) Why not use Depends?
b) If you are going to add that flag for all builds, let's remove this logic from the Depends

jgfouca:
Thanks @ndkeen. I have removed the other entries that became redundant.

ndkeen:
I think it's better if we use Depends.
I think we should be moving toward *removing* some of these odd compiler flags instead of adding them to all sources.

Why would not using Depends work?

jgfouca:
@ndkeen , this issue came up because the CIME CI testing stopped working. I assume because some file was added that needs this flag and the cmake macros for the CI machine did not cover it. So, the advantage of doing it globally is to avoid the whack-a-mole situation. As far as disadvantages, I don't know enough about this flag to say. Are there cases where having this flag could hide mistakes that we want to catch? @jedwards4b , any thoughts here?

rljacob:
If its more then a handful of files, and does no harm, just make it a global flag.

ndkeen:
I suspect the issue came up when you starting using GNU v10 or higher.

There are only 2 (two) files that have this very odd line (which surely can be done another way, that's another story) and so only need to add the flag for those 2 files.

jedwards4b:
#4822 would seem to have the code that you need, so if it is already merged maybe I have misinterpreted the issue in cime issue 4399?

jgfouca:
@jedwards4b , i'm wondering that too. Based on what I'm seeing in some of the CIME PRs, the E3SM failure is this:
```
ERROR: /__w/cime/cime/cime/share/build/buildlib.gptl FAILED, cat /storage/cases/scripts_regression_test.20230419_133127/TestCreateNewcase/testcreatenewcase_with_user_compset/bld/gptl.bldlog.230419-133436
```

So, gptl (a sharedlib) is failing to build. I wish I could access that log file.

jedwards4b:
But you can.  It's in an [artifact file](https://docs.github.com/en/actions/managing-workflow-runs/downloading-workflow-artifacts).

```
/opt/conda/bin/mpif90 -c -I/__w/cime/cime/cime/share/timing  -I/opt/conda/include -mcmodel=medium -fconvert=big-endian -ffree-line-length-none -ffixed-line-length-none -O   -DFORTRANUNDERSCORE -DNO_R16 -DCPRGNU -DFORTRANUNDERSCORE -DNO_R16 -DCPRGNU -DFORTRANUNDERSCORE -DNO_R16 -DCPRGNU -DMCT_INTERFACE -DHAVE_MPI -ffree-form  /__w/cime/cime/cime/share/timing/perf_utils.F90
make: Leaving directory '/storage/cases/scripts_regression_test.20230419_133127/TestCreateNewcase/testcreatenewcase_with_user_compset/bld/gnu/openmpi/nodebug/nothreads/mct/gptl'
/__w/cime/cime/cime/share/timing/gptl.c: In function 'GPTLpr_summary_file':
/__w/cime/cime/cime/share/timing/gptl.c:3444:8: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
 3444 |   if (((int) comm) == 0)
      |        ^
/__w/cime/cime/cime/share/timing/perf_utils.F90:282:18:

  282 |    call MPI_BCAST(vec,lsize,MPI_INTEGER,0,comm,ierr)
      |                  1
......
  314 |    call MPI_BCAST(vec,lsize,MPI_LOGICAL,0,comm,ierr)
      |                  2
Error: Type mismatch between actual argument at (1) and actual argument at (2) (INTEGER(4)/LOGICAL(4)).
```

jgfouca:
@jedwards4b , nice! Now, onto the next mystery. gptl hasn't changed, so I guess docker MPI or fortran version must have changed?

PR #5608, opened 2023-04-19
Rename gnu_docker.cmake to gnu_docker-scream.cmake

jgfouca:
The 'docker' name is too generic and causing the addition of this file to E3SM to break the E3SM/CIME github CI.

[BFB]


jgfouca:
Skipping integration.

PR #5611, opened 2023-04-20
Update partitions for 4 MPAS-Seaice meshes

xylar:
This merge points to new partition files for each of the following 4 MPAS-Seaice meshes.  Each mesh has about 400 files that are expected to support nearly any conceivable core count.

Meshes with updated partitions:
* EC30to60E2r2
* ECwISC30to60E2r1
* SOwISC12to60E2r4
* WC14to60E2r3

The partition files are better load-balanced, since each core owns cells in both polar and equatorial regions.  They were created with the tools described in:
http://mpas-dev.github.io/MPAS-Tools/stable/seaice/partition.html

See https://github.com/MPAS-Dev/compass/pull/563 for more details on how the core counts were determined.

[NML]
[non-BFB] only for mpassi globalStats files for these meshes -- does not change the seaice state

### DIFF
--- a/components/mpas-seaice/cime_config/buildnml
+++ b/components/mpas-seaice/cime_config/buildnml
@@ -206,8 +206,8 @@ def buildnml(case, caseroot, compname):
             grid_prefix = 'mpassi.ARRM10to60E2r1.rstFrom1monthG-chrys'
 
     elif ice_grid == 'EC30to60E2r2':
-        decomp_date = '200908'
-        decomp_prefix = 'mpas-seaice.graph.info.'
+        decomp_date = '230313'
+        decomp_prefix = 'partitions/mpas-seaice.graph.info.'
         grid_date = '210210'
         grid_prefix = 'seaice.EC30to60E2r2'
         if ice_ic_mode == 'spunup':
@@ -215,8 +215,8 @@ def buildnml(case, caseroot, compname):
             grid_prefix = 'mpassi.EC30to60E2r2.rstFromG-anvil'
 
     elif ice_grid == 'WC14to60E2r3':
-        decomp_date = '200714'
-        decomp_prefix = 'mpas-seaice.graph.info.'
+        decomp_date = '230313'
+        decomp_prefix = 'partitions/mpas-seaice.graph.info.'
         grid_date = '210210'
         grid_prefix = 'seaice.WC14to60E2r3'
         if ice_ic_mode == 'spunup':
@@ -233,8 +233,8 @@ def buildnml(case, caseroot, compname):
             grid_prefix = 'mpassi.WCAtl12to45E2r4.rstFromG-anvil'
 
     elif ice_grid == 'SOwISC12to60E2r4':
-        decomp_date = '210107'
-        decomp_prefix = 'mpas-seaice.graph.info.'
+        decomp_date = '230314'
+        decomp_prefix = 'partitions/mpas-seaice.graph.info.'
         grid_date = '210107'
         grid_prefix = 'seaice.SOwISC12to60E2r4'
         data_iceberg_file += 'Iceberg_Interannual_Merino_SOwISC12to60E2r4.nc'
@@ -245,8 +245,8 @@ def buildnml(case, caseroot, compname):
     elif ice_grid == 'ECwISC30to60E2r1':
         grid_date = '210413'
         grid_prefix = 'seaice.ECwISC30to60E2r1'
-        decomp_date = '200916'
-        decomp_prefix = 'mpas-seaice.graph.info.'
+        decomp_date = '230314'
+        decomp_prefix = 'partitions/mpas-seaice.graph.info.'
         data_iceberg_file += 'Iceberg_Interannual_Merino_ECwISC30to60E2r1.nc'
         if ice_ic_mode == 'spunup':
             grid_date = '210414'


xylar:
@jonbob, I'm labeling this to match #5568, but please update as appropriate.

components/mpas-seaice/cime_config/buildnml
```diff
@@ -243,8 +243,8 @@ def buildnml(case, caseroot, compname):
             grid_prefix = 'mpassi.SOwISC12to60E2r4.rstFromG-anvil'
 
     elif ice_grid == 'ECwISC30to60E2r1':
-        grid_date = '210413'
-        grid_prefix = 'seaice.ECwISC30to60E2r1'
+        grid_date = '230314'
+        grid_prefix = 'partitions/seaice.ECwISC30to60E2r1'
```
darincomeau, line 247: @xylar these changes should be `decomp_*`, rather than `grid_*`.

darincomeau (changes requested):

components/mpas-seaice/cime_config/buildnml
xylar: Woops, thanks!

components/mpas-seaice/cime_config/buildnml
```diff
@@ -245,8 +245,8 @@ def buildnml(case, caseroot, compname):
     elif ice_grid == 'ECwISC30to60E2r1':
         grid_date = '210413'
         grid_prefix = 'seaice.ECwISC30to60E2r1'
-        decomp_date = '200916'
-        decomp_prefix = 'mpas-seaice.graph.info.'
+        decomp_date = '230314'
+        decomp_prefix = 'partition/mpas-seaice.graph.info.'
```
xylar, line 249: @darincomeau, does that look right this time?

darincomeau: looks right - thanks for the quick fix! I'm testing each grid in part to make sure the new `partitions` directories created on the *other* LCRC inputdata directory have the right permissions.

darincomeau: oh actually, just noticed this one is `partition` whereas the others are `partitions`

xylar: Oh, no!

xylar: Fixes again, hopefully for the last time 🤞🏻

components/mpas-seaice/cime_config/buildnml
```diff
@@ -245,8 +245,8 @@ def buildnml(case, caseroot, compname):
     elif ice_grid == 'ECwISC30to60E2r1':
         grid_date = '210413'
         grid_prefix = 'seaice.ECwISC30to60E2r1'
-        decomp_date = '200916'
-        decomp_prefix = 'mpas-seaice.graph.info.'
+        decomp_date = '230314'
+        decomp_prefix = 'partitions/mpas-seaice.graph.info.'
```
xylar, line 249: @darincomeau, have one more look and let me know how it goes.

xylar:
@darincomeau
> looks right - thanks for the quick fix! I'm testing each grid in part to make sure the new partitions directories created on the other LCRC inputdata directory have the right permissions.

Great idea!

darincomeau (approved):
Approved based on running `SMS.TL319_*.GMPAS-JRA1p5.chrysalis` for each grid.

Thanks @xylar !

xylar:
Thanks @darincomeau!

eclare108213 (approved):
Approving based on @darincomeau's tests (thanks!)

xylar:
@jonbob and @proteanplanet, could you please review this when you have time?  It's a relatively high priority because apparently a production test is failing because of missing partition files right now.

jonbob (approved):
Approved based on visual inspection and checking the local and public server inputdata directories on lcrc

xylar:
@amametjanov, it seems like you added configurations on Anvil here:
https://github.com/E3SM-Project/E3SM/blob/3e1e3db133aa17b9bd53003cc5ffa3de6790901f/cime_config/testmods_dirs/config_pes_tests.xml#L184-L199
that use 468 cores (13 nodes) on Anvil.  Is it important to support multiples of 11 and 13, then?  Or would it be okay to change this test to use 12 or 14 nodes instead?

xylar:
That seems consistent with what I found. That partition file never existed.  We should just pick a different size.  I could include the change here, but I think that might risk causing some confusion.  Can we make it a separate PR to follow this one?

rljacob:
But the pe layout changes were made on April 5 and there are definitely passes of this test after that.  See https://my.cdash.org/test/77517732

xylar:
I'm adding new `simple_seaice_partiton` tool: https://github.com/MPAS-Dev/MPAS-Tools/pull/506

I'll update the docs at:
as soon as it's released (later today).

I used it to make a partition files for WC14 for 486 tasks. in case we need it (may not if #5642 goes in first).

proteanplanet (changes requested):
I tested each of the four meshes on Anvil for 3-month D-cases.  All cases ran to completion except for the SOwISC12to60E2r4 mesh, which failed with partition errors.  Details of the SORRM failure are available at /lcrc/group/e3sm/ac.afroberts/E3SM-Polar/DSORRM3.partition.new-partition-files.xylar.anvil

xylar:
@jonbob, I could use help debugging the error in @proteanplanet's Anvil run when you can:
```
  0:  ERROR: (seq_domain_check_grid)  incompatible domain grid coordinates
```
It's not obvious to me that this is as simple as a missing partition file.  I believe the run is on 720 tasks, which definitely does exist (and has been copied to the right location):
```
/lcrc/group/e3sm/data/inputdata/ice/mpas-seaice/SOwISC12to60E2r4/partitions/mpas-seaice.graph.info.230314.part.720
```

jonbob:
@xylar - the problem is not related to this PR. There's a domain file for the SOwISC12to60E2r4 grid that is incorrect, at least for D-cases.

xylar:
😅 Let me know if I can help.

jonbob:
@proteanplanet - it turns out using a domain file without a mask for the wISC meshes causes trouble for D-cases. I assume it's a problem for the docn som model, which probably knows nothing about the ice shelf cavities and throws an error about inconsistent domains. I've opened an issue, [#5653](https://github.com/E3SM-Project/E3SM/issues/5653) and @darincomeau and I will work at making the domain files consistent. I ran some one-day tests that showed the results were BFB using either the mask or nomask domain files, and Darin has run some longer comparisons as well. So we'll try to get this fixed as soon as we can, but the failure you saw is not related to this PR. Thanks for catching it, though

proteanplanet (approved):
Since the D-case issue is not occurring due to this PR, I'm approving this PR.

jonbob:
passes:

with expected NML DIFF. Merged to next

xylar:
@mahf708, I can make a partition file for 2752 tasks for the EC30to60E2r2.  I would also appreciate understanding how this number arises.  This is 64*43.  I did not make partition files for task counts involving prime numbers larger that 7 to keep the number reasonable (~400 rather than ~30,000).  It seems there are a surprising number of E3SM runs that use MPAS-Ocean or MPAS-Seaice on some odd node counts and it would be helpful to know if these are necessary or if, say, 64*42 or 64*45 would have worked just as well.

xylar:
@mahf708 (or someone else who knows), why do atmosphere-only runs need to partition mpas-seaice across 43 nodes, and why are they using the `EC30to60E2r2` mesh?  Are data files being stored on that mesh?

xylar:
@mahf708, I just added the `partitions/mpas-seaice.graph.info.230313.part.2752` file in `/lcrc/group/e3sm/public_html/inputdata/ice/mpas-seaice/EC30to60E2r2`.  Please let me know if things work for you now.

xylar:
Okay, I added a few more pretty random sizes I found in:
https://github.com/E3SM-Project/E3SM/blob/master/cime_config/allactive/config_pesall.xml
* 2624 = 41*64
* 5248 = 41*128

I think that's it from that file but please point me to others like it if there are missing counts.

rljacob:
@xylar instead of waiting to hear what is needed, just make new versions of all the part files in /lcrc/group/e3sm/public_html/inputdata/ice/mpas-seaice/EC30to60E2r2

We now have tests failing because the new version of, for example, mpas-seaice.graph.info.200908.part.396, doesn't exist in the new decomp date/prefix introduced by this PR.

Atmosphere-only cases use the thermodynamic version of MPAS-SEAICE and decompose it on the same processors as the atmosphere which should explain all the odd looking counts.

xylar:
@rljacob I'm flying to the US tomorrow so won't be in a position to do this until Monday.

xylar:
> instead of waiting to hear what is needed, just make new versions of all the part files in /lcrc/group/e3sm/public_html/inputdata/ice/mpas-seaice/EC30to60E2r2

This would have been great feedback to get before this got merged to next and became a problem right as I'm traveling...

rljacob:
You should have done more testing.  I'm going to revert this PR off of next.

xylar:
@rljacob , I'm so pissed about this whole process!!! I did this work over months as a favor to the sea ice group because they were having trouble get traction with their load balancing approach. I asked for feedback and got only a vague sense of what was needed. I simply can't test everything!! I don't have the level of expertise.

xylar:
I see, I certainly realized that. I just didn't see any reasonable way of doing that testing besides merging to next. Maybe there's an equivalent of next that isn't next?

xylar:
> I don't think anyone realized that the process of adding new part files also makes the old part files with the old name in the old location inaccessible.

After having a day to cool off, I'm in a better space to explain this.  It wasn't possible to keep the same location or date stamp for the new partition files because if we did that it would not be possible to distinguish between the new ones and the old ones.

It certainly would have been possible to make all the partition sizes that exist for all current meshes, but this is an extraordinarily tedious process to go through. Instead, I tried my best (and will continue to try) automate this process.  Part of the motivation is that we have at least a dozen new meshes coming soon, and they will need partition files for both ocean and sea ice.

sarats:
I don't have the whole context here.

While trying to run on Frontier, I see the following input file missing (IIRC these tests worked earlier).
```
402 wget failed with output:  and errput --2023-05-10 15:26:16--  https://portal.nersc.gov/project/e3sm/inputdata/ice/mpas-seaice/EC30to60E2r2/mpas-seaice.graph.info.200908.part.56
```
So, did this get removed? Is there any workaround or need to wait until this PR is merged?

cc @grnydawn

xylar:
@sarats, no partition files were removed.  The ones referenced in this PR (which I believe @rljacob reverted from `next`) would be in a subdirectory `partitions` so the missing partition file you are referring to is not related to this PR.

My guess would be that the partition file was created locally rather than placed on the server.  I learned that this has sometimes been the practice (e.g. on LCRC).  In any case, I can check if that particular task count is going to be available once this gets merged.  In the meantime, you could create the old type of partition in the missing location with `gpmetis`. I just don't want this to become the practice because we want to move to the load-balanced partition files created with `simple_seaice_partitions` (https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/48529761/Make+a+new+graph+partition+for+MPAS).

xylar:
@sarats, yes, a partition file for 56 tasks will definitely be available once this PR goes in.  I'm rather surprised it isn't on the mirror server at NERSC. Are we sure that was a complete copy of the LCRC server?

xylar:
@rljacob, do you have any insight into this? The directory https://portal.nersc.gov/project/e3sm/inputdata/ice/mpas-seaice/EC30to60E2r2/ is nearly empty.  It should have all the contents of https://portal.nersc.gov/project/e3sm/inputdata/ice/mpas-cice/EC30to60E2r2/ and more, since I copied everything from `mpas-cice` into `mpas-seaice` on LCRC.  It also seems like there are far fewer partition files here than there were on LCRC.

xylar:
Ah, I think I see.  This was not meant to be a complete mirror of `inputdata` on LCRC, this just happens to be what we have while LCRC is down -- whatever got downloaded at NERSC in the course of testing at NERSC itself.

So @sarats, I don't think it's a surprise that partition files for other machines like Frontier aren't there.  You need to wait until LCRC comes back up on Friday.

rljacob:
Yes we didn't take the time to make NERSC inputdata a copy of LCRC.   Quickest fix is to recreate the part file with gpmetis as Xylar said.

jonbob:
@xylar -- I went through all of the layouts in config_pesall.xml and only found one seaice count that we haven't made: 5584. However, it's for a cori-knl configuration and since that platform is being retired this month, I don't believe there's any reason to add that.

xylar:
Thanks @jonbob!  That must have been a real slog to make sure everything is covered.  I agree with you that it's safe to ignore that one cori-knl configuration. Presumably, we aren't testing on Cori anymore and it is exceedingly unlikely that anyone will try to run that particular configuration before Cori gets decommissioned.

xylar:
@jonbob, please don't hesitate to ask if there are tests I can run before this goes back to next.  You had mentioned running the full set of tests on Anvil that get run when anything gets merged to next.  If you want to give me instructions on Slack for how to do those runs, I'm happy to do them.  Alternatively, if there is another machine I can test, let me know.

rljacob:
A recent run on Frontier used 22400 tasks for mpas seaice.  Is that available as a new partition?   https://pace.ornl.gov/exp-details/138910

xylar:
@rljacob, yes, it’s there for EC30to60E2r2, which I assume is the mesh you mean:
https://web.lcrc.anl.gov/public/e3sm/inputdata/ice/mpas-seaice/EC30to60E2r2/partitions/mpas-seaice.graph.info.230313.part.22400

jonbob:
@xylar -- I launched the anvil prod test suite last night. I'm checking results now

rljacob:
That frontier case was using oRRS18to6v3 since it was an ne120 atmosphere.

xylar:
@rljacob, thanks for clarifying.  Yes, that also exists:
https://web.lcrc.anl.gov/public/e3sm/inputdata/ice/mpas-seaice/oRRS18to6v3/partitions/mpas-seaice.graph.info.230424.part.22400

xylar:
@rljacob, @jonbob had asked last week:
> Is there more testing we should do?

Is there anything else you'd like to see here before we try another merge to next?

rljacob:
Were the part files in /lcrc/group/e3sm/public_html/inputdata/ice/mpas-seaice/EC30to60E2r2 replaced with new versions?  That was the problem last time.

xylar:
@mahf708 

> Placing them on the correct path (perhaps for a good reason, this PR invents a new location by prepending with partitions/ for each file in the prefix --- this shouldn't be a problem if buildnml is the only way these files get accessed/called, but I am not familiar enough to judge; see example below for the first diff)

Yes, I believe this is supposed to be the only way folks point to these partition files. Any workflows outside of E3SM itself are on their own. CIME should (but doesn't always because of a known bug) download these files automatically as needed. The exception is for the 2nd part of a PEM test or similar, which compares 2 task counts and where the partition file for the second task gets missed.

xylar:
> Were the part files in /lcrc/group/e3sm/public_html/inputdata/ice/mpas-seaice/EC30to60E2r2 replaced with new versions? That was the problem last time.

@rljacob, thanks for prompting me to check that.  I added 12 more graph files that slipped through the cracks:
* 1040
* 1520
* 1760
* 1840
* 1872
* 2080
* 2200
* 234
* 3392
* 3904
* 4
* 936

I have no reasonable algorithmic way to produce any of these (4 is way too few cores to be safe and all the rest involve large prime numbers and are not even multiples of any node count).  So for future meshes, they will need to be produced by hand like I just did.  I think we have to live with that for now.

jonbob:
@rljacob -- are you up for us merging this again? @xylar and I are both around to fix any missing graph files, though I feel confident we've got them covered now

rljacob:
Sure once the dashboard clears.

rljacob:
I reset next but wait until tomorrow since we won't get another bless done until Friday.

jonbob:
passes:

with expected NML DIFF

merged to next

xylar:
Thanks so much @jonbob and everyone, and sorry for the drama (preserved here forever for posterity).

jonbob:
merged to master with expected NML DIFFs and regular DIFFs blessed -- with the exception of some pm-cpu tests that looks like they failed for random machine reasons.

Thanks @xylar for all the hard work and patience

jonbob:
two pm-cpu tests blessed that had failed previously for other reasons

amametjanov:
@xylar I need `ice/mpas-seaice/EC30to60E2r2/partitions/mpas-seaice.graph.info.230313.part.52` for ALCF Sunspot, but can't find `seaice.EC30to60E2r2.230313.nc` mesh file. Is it okay to generate a partition file from
```
/lcrc/group/e3sm/public_html/inputdata/ice/mpas-seaice/EC30to60E2r2/seaice.EC30to60E2r2.200908.nc
```

xylar:
@amametjanov, no, at this point sea-ice partition files always need to be generated using the procedure outlined in:
This procedure needs a sea-ice mesh file, not a graph file.  Multiple graph files are generated automatically as part of that procedure, combined and then deleted.

I will generate that particular file for you but let me know if you will be able to do this yourself in the future.

xylar:
@amametjanov, I'm sorry, I see that you were, indeed, pointing to a sea-ice mesh file.  I tried to use that one to generate the partition file using the procedure above and for some reason it seems to be missing a necessary variable.  I don't quite understand what that could be.  But I was able to proceed using an ocean restart file for that mesh (the mesh is identical for ocean and sea ice):

```
simple_seaice_partitions \
    -m /lcrc/group/e3sm/public_html/inputdata/ocn/mpas-o/EC30to60E2r2/ocean.EC30to60E2r2.210210.nc 
    -p /lcrc/group/e3sm/public_html/inputdata/ice/mpas-seaice/EC30to60E2r2/partitions/mpas-seaice.graph.info.230313 \
    -n 52
```

xylar:
@amametjanov, by the way, are you likely to need support for 52 tasks on other meshes (or other multiples of 52 tasks for this mesh) for Sunspot?  I'm still trying to understand these unusual task counts so I can incorporate them into the algorithmically generated partition files, keeping in mind that we will create 12 new meshes soon.

amametjanov:
Thanks. No need, I can generate other partition files as needed (with simple_seaice_partitions).

xylar:
Thanks @sarats, that's exactly the info I needed.

PR #5612, opened 2023-04-20
Avoid more keyword args in MOSART

jayeshkrishna:
Avoid keyword/named arguments in more non-optional arguments
in PIO calls in MOSART.

Also see PR #5396

[BFB]


liho745 (approved):
Looks good to me.

PR #5616, opened 2023-04-21
Add Polar historical/SSP compsets

darincomeau:
Adds new Polar compsets that are similar to `WCYCL20TR`, `WCYCLSSP585`, `WCYCLSSP370`, except uses Polar settings for Antarctic runoff (disables Antarctic runoff from the coupler to the ocn, includes ice-shelf melt fluxes and data icebergs).

### DIFF
--- a/cime_config/allactive/config_compsets.xml
+++ b/cime_config/allactive/config_compsets.xml
@@ -303,6 +303,21 @@
 </compset>
 
 <compset>
+  <alias>CRYO20TR</alias>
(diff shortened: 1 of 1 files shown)


darincomeau:
I ran smoke tests on each of these compsets, confirmed correct ocn/ice namelist settings for Antarctic runoff are set.

@proteanplanet note I kept `CRYO` in the compset aliases to be consistent with existing Polar (formerly Cryosphere) compsets.

proteanplanet (approved):
As this passes a smoke test it is approved in readiness for SSP B-case simulations using ice shelves.

jonbob:

jonbob:
merged to master

PR #5618, opened 2023-04-21
Added missing OpenACC data directives for ocean landIceEdgeFraction

### LINKED ISSUE #5615: landIceEdgeFraction not on device in mpas-ocean when OpenACC enabled
In recent changes to ocean vmix routines, landIceEdgeFraction is declared present in loop constructs but has never been transferred to the device so is not in fact present causing failures when running with OpenACC enabled. Will submit a fix presently.

philipwjones:
OpenACC data directives for a new landIceEdgeFraction array were missing so it was not present on the device for later loops. Added appropriate directives to fix this issue and also added a missing deallocate for the same array. Also fixed some ACC directives for another cell-centered landIceFraction array that didn't change functionality but is more performant.

Tested on Frontier. CPU version is bit-for-bit as expected. GPU version runs successfully. No bit-for-bit comparison on the GPU side is possible due to this bug, but GPU results were consistent with CPU with usual roundoff level diffs. 

Fixes #5615 
[bfb]

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_vmix.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_vmix.F
@@ -1163,17 +1163,35 @@ contains
          timeLevel = 1
       end if
 
-      call mpas_pool_get_subpool(forcingPool, 'tracersSurfaceFlux', tracersSurfaceFluxPool)
+      call mpas_pool_get_subpool(forcingPool, 'tracersSurfaceFlux', &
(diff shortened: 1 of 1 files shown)


cbegeman (approved):
@philipwjones Thank you for attending to this and testing it out! The changes look good by visual inspection. Let me know if you would like me to do any further testing.

jonbob:
passes:

merged to next

jonbob:
merged to master

PR #5619, opened 2023-04-22
Fix high-frequency sea ice coupling to use ice v-vel component in atmospheric boundary layer

proteanplanet:
Fixes atmospheric boundary layer calculations when high frequency coupling is switched on by ensuring that both the u- and v-component of sea ice velocity is used in turbulent atmospheric flux calculations.  This change does not affect the current standard configuration of E3SM on ECwISC30to60E2r1 and EC30to60E2r2 meshes, but it is non-BFB on all other meshes, including the NARRM. The change has been tested in a 3-month D-case smoke test using DTESTM-JRA1p5 on a TL319_EC30to60E2r2 grid, and is BFB with master in that configuration.  The bug has been fixed in CESM and RASM long ago, but remained in E3SM, and was detected by @eclare108213 during implementation of the CICE Consortium's Icepack in E3SM.

[non-BFB] with active seaice on all grids except EC30to60E2r2 and ECwISC30to60E2r1

PR #5620, opened 2023-04-22
Sets sea ice high-frequency turbulent boundary layer as the E3SM default

proteanplanet:
Switches on high-frequency boundary layer coupling between the atmosphere and sea ice for all configurations of E3SM, where previously it was switched off for standard resolution without (EC30to60E2r2) and with (ECwISC30to60E2r1) ice shelves. The change includes a bug fix submitted in a separate PR https://github.com/E3SM-Project/E3SM/pull/5619.  This is a significant change, including the bug fix, to E3SM, and has therefore undergone the following tests:

1) 3-month D-case smoke test using DTESTM-JRA1p5 on a TL319_EC30to60E2r2 grid (non-BFB with master)
2) 20-year B-case smoke test using WCYCL1850 on ne30pg2_EC30to60E2r2 grid _(test underway now)_

[NML] for configurations with EC30to60E2r2 or ECwISC30to60E2r1
[non-BFB] for configurations with EC30to60E2r2 or ECwISC30to60E2r1

### DIFF
--- a/components/mpas-seaice/bld/namelist_files/namelist_defaults_mpassi.xml
+++ b/components/mpas-seaice/bld/namelist_files/namelist_defaults_mpassi.xml
@@ -395,11 +395,7 @@
 <config_calc_surface_temperature>true</config_calc_surface_temperature>
 <config_use_form_drag>false</config_use_form_drag>
 <config_use_high_frequency_coupling>true</config_use_high_frequency_coupling>
-<config_use_high_frequency_coupling ice_grid="ECwISC30to60E2r1">false</config_use_high_frequency_coupling>
-<config_use_high_frequency_coupling ice_grid="EC30to60E2r2">false</config_use_high_frequency_coupling>
 <config_boundary_layer_iteration_number>10</config_boundary_layer_iteration_number>
-<config_boundary_layer_iteration_number ice_grid="ECwISC30to60E2r1">5</config_boundary_layer_iteration_number>
-<config_boundary_layer_iteration_number ice_grid="EC30to60E2r2">5</config_boundary_layer_iteration_number>
 
 <!-- ocean -->
 <config_use_ocean_mixed_layer>false</config_use_ocean_mixed_layer>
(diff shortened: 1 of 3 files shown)


proteanplanet:
Current code changes pass the e3sm_ice_developer test and have completed a 20-year B-case smoke test.  Further testing and checking is continuing.

components/mpas-seaice/bld/namelist_files/namelist_defaults_mpassi.xml
```diff
@@ -395,11 +395,7 @@
 <config_calc_surface_temperature>true</config_calc_surface_temperature>
 <config_use_form_drag>false</config_use_form_drag>
 <config_use_high_frequency_coupling>true</config_use_high_frequency_coupling>
-<config_use_high_frequency_coupling ice_grid="ECwISC30to60E2r1">false</config_use_high_frequency_coupling>
-<config_use_high_frequency_coupling ice_grid="EC30to60E2r2">false</config_use_high_frequency_coupling>
```
darincomeau, line 399: >Didn't @jonbob say that the defaults set in Registry.xml are only used for MPAS-SI stand-alone configurations?

@eclare108213 the changes here are what E3SM uses.

eclare108213: And so does removing these lines and not replacing them with 'true' mean that this option must always be set in a namelist file?  Isn't that kind of dangerous?

darincomeau: L397 sets the option as true by default for all grids. Previously L398,399 changed the default to false only for the `ECwISC30to60E2r1`, `EC30to60E2r2` grids.

eclare108213: Ah, sorry I missed that.

rljacob:
@proteanplanet move this draft PR off of E3SM-Project/E3SM.

PR #5633, opened 2023-04-25
removed private arrays in ocn tracer advection

philipwjones:
Private temporary arrays inside OpenACC parallel loops were creating performance problems with Cray compiler on Frontier when OpenACC is enabled for GPU runs (many 10x slower).  The changes here remove these temporary arrays and recompute quantities. This removes the performance penalty on Frontier with little impact on CPU performance at E3SM optimization levels. With more aggressive optimization in standalone, there can be a slight slowdown (3%). Thanks to @grnydawn for tracking this down and the initial fix. 

This has been tested on Frontier and with the compass pr tests on Chrysalis and are bit-for-bit.

[bfb]

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_tracer_advection_mono.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_tracer_advection_mono.F
@@ -117,10 +117,6 @@ module ocn_tracer_advection_mono
          invAreaCell1,      &! inverse cell area
          coef1, coef3        ! temporary coefficients
 
-      real (kind=RKIND), dimension(:), allocatable :: &
-         wgtTmp,            &! vertical temporaries for
-         flxTmp, sgnTmp      !   high-order flux computation
-
       real (kind=RKIND), dimension(:,:), allocatable :: &
          tracerCur,     &! reordered current tracer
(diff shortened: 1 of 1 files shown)


grnydawn (approved):
it looks good to me.

mark-petersen (approved):
Compiled with both gnu and intel, with openMP, optimized and debug for both. Passes the nightly test suite with default flags. Then I forced compass to use:
```
+++ b/compass/ocean/tests/baroclinic_channel/namelist.forward
+config_flux_limiter = 'monotonic'

+++ b/compass/ocean/tests/global_ocean/namelist.forward
+config_flux_limiter = 'monotonic'
```
I then reran the nightly test suite on master with gnu optimized and compared this PR. They match bfb, showing that the code changes are correct. Thanks!

jonbob:
passes:
* ERS.ne11_oQU240.WCYCL1850NS.pm-cpu_gnu
* SMS_D_Ld3.T62_oQU120.CMPASO-IAF.pm-cpu_gnu
* SMS_D_Ld1.ne30pg2_EC30to60E2r2.WCYCL1850.pm-cpu_gnu.allactive-wcprod

merged to next

jonbob:
merged to master

PR #5558, opened 2023-03-25
Add smaller PE-layouts for tests on Anvil

amametjanov:
Add smaller PE-layouts for tests on Anvil to get faster queue throughput.

[NML] - in cime_pes namelists
[non-BFB] - in mpaso.hist.am.globalStats outputs

PR #5635, opened 2023-04-26
Update ocean and sea-ice partitions for 6 E3SM v1 meshes used in testing

xylar:
New partition files are created for:
* oQU480
* oQU240
* oQU240wLI
* oQU120
* oEC60to30v3
* oRRS18to6v3

The new `partitions` subdirectory contains many more partition files that should hopefully avoid missing partition issues plaguing some developers.

The new sea-ice partition files are load-balanced so that each core gets a fraction of both polar and equatorial regions.  This is not a concern for the first 4 meshes because they are very small but it is not a bad idea for the last 2 meshes.

### DIFF
--- a/components/mpas-ocean/cime_config/buildnml
+++ b/components/mpas-ocean/cime_config/buildnml
@@ -108,8 +108,8 @@ def buildnml(case, caseroot, compname):
             logger.warning("         But no file available for this grid.")
 
     elif ocn_grid == 'oQU480':
-        decomp_date = '151209'
-        decomp_prefix = 'mpas-o.graph.info.'
+        decomp_date = '230422'
+        decomp_prefix = 'partitions/mpas-o.graph.info.'
         ic_date = '151209'
         ic_prefix = 'ocean.QU.480km'
         if ocn_ic_mode == 'spunup':
@@ -117,8 +117,8 @@ def buildnml(case, caseroot, compname):
             logger.warning("         But no file available for this grid.")
 
     elif ocn_grid == 'oQU240':
-        decomp_date = '151209'
-        decomp_prefix = 'mpas-o.graph.info.'
+        decomp_date = '230422'
+        decomp_prefix = 'partitions/mpas-o.graph.info.'
         ic_date = '151209'
         ic_prefix = 'ocean.QU.240km'
         if ocn_ic_mode == 'spunup':
@@ -126,8 +126,8 @@ def buildnml(case, caseroot, compname):
             logger.warning("         But no file available for this grid.")
 
     elif ocn_grid == 'oQU240wLI':
-        decomp_date = '160929'
-        decomp_prefix = 'mpas-o.graph.info.'
+        decomp_date = '230422'
+        decomp_prefix = 'partitions/mpas-o.graph.info.'
         ic_date = '230220'
         ic_prefix = 'ocean.QU.240wLI'
         if ocn_ic_mode == 'spunup':
--- a/components/mpas-ocean/cime_config/buildnml
+++ b/components/mpas-ocean/cime_config/buildnml
@@ -75,8 +75,8 @@ def buildnml(case, caseroot, compname):
     eco_forcing_file = ''
 
     if ocn_grid == 'oEC60to30v3':
-        decomp_date = '161222'
-        decomp_prefix = 'mpas-o.graph.info.'
+        decomp_date = '230424'
+        decomp_prefix = 'partitions/mpas-o.graph.info.'
         restoring_file = 'sss.monthlyClimatology.PHC2_salx_040803.oEC60to30v3.nc'
         analysis_mask_file = 'oEC60to30v3_Atlantic_region_and_southern_transect.nc'
         ic_date = '170905'
@@ -135,8 +135,8 @@ def buildnml(case, caseroot, compname):
             logger.warning("         But no file available for this grid.")
 
     elif ocn_grid == 'oQU120':
-        decomp_date = '160318'
-        decomp_prefix = 'mpas-o.graph.info.'
+        decomp_date = '230424'
+        decomp_prefix = 'partitions/mpas-o.graph.info.'
         ic_date = '160318'
         ic_prefix = 'ocean.QU.120km'
         if ocn_ic_mode == 'spunup':
@@ -166,8 +166,8 @@ def buildnml(case, caseroot, compname):
             logger.warning("         But no file available for this grid.")
 
     elif ocn_grid == 'oRRS18to6v3':
-        decomp_date = '170111'
-        decomp_prefix = 'mpas-o.graph.info.'
+        decomp_date = '230424'
+        decomp_prefix = 'partitions/mpas-o.graph.info.'
         restoring_file = 'sss.monthlyClimatology.PHC2_salx_040803.oRRS18to6v3.nc'
         analysis_mask_file = 'oRRS18to6v3_mocBasinsAndTransects20210623.nc'
         ic_date = '171116'
--- a/components/mpas-seaice/cime_config/buildnml
+++ b/components/mpas-seaice/cime_config/buildnml
@@ -73,8 +73,8 @@ def buildnml(case, caseroot, compname):
     points_file = ''
 
     if ice_grid == 'oEC60to30v3':
-        decomp_date = '161222'
-        decomp_prefix = 'mpas-cice.graph.info.'
+        decomp_date = '230424'
+        decomp_prefix = 'partitions/mpas-seaice.graph.info.'
         grid_date = '161222'
         grid_prefix = 'seaice.EC60to30v3'
         if ice_ic_mode == 'spunup':
@@ -106,8 +106,8 @@ def buildnml(case, caseroot, compname):
             logger.warning("         But no file available for this grid.")
 
     elif ice_grid == 'oQU120':
-        decomp_date = '151209'
-        decomp_prefix = 'mpas-cice.graph.info.'
+        decomp_date = '230424'
+        decomp_prefix = 'partitions/mpas-seaice.graph.info.'
         grid_date = '151209'
         grid_prefix = 'seaice.oQU120km'
         if ice_ic_mode == 'spunup':
@@ -161,8 +161,8 @@ def buildnml(case, caseroot, compname):
             logger.warning("         But no file available for this grid.")
 
     elif ice_grid == 'oRRS18to6v3':
-        decomp_date = '180214'
-        decomp_prefix = 'mpas-seaice.v2.'
+        decomp_date = '230424'
+        decomp_prefix = 'partitions/mpas-seaice.graph.info.'
         grid_date = '170111'
         grid_prefix = 'seaice.RRS18to6v3'
         if ice_ic_mode == 'spunup':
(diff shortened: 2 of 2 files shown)


xylar:
@jonbob, I'm a little unsure about the labels so please correct as needed.

xylar:
Moved from https://github.com/E3SM-Ocean-Discussion/E3SM/pull/51, where there is some relevant discussion.

xylar:
I am still building partition files for the oRRS18to6v3 mesh.  They take about 6 minutes each for larger task counts.  But I also think I'm at high enough numbers (over half a million tasks) that we're not going to use them anytime soon if ever.

darincomeau (approved):
Approved by visual inspection here and in public inputdata directory.

xylar:
@rljacob, that should be addressed in #5611, which is still awaiting review.

xylar:
@rljacob, that should be addressed in #5611, which is still awaiting review.

darincomeau:
Looking at the new partition files for this mesh in `inputdata/ice/mpas-seaice/WC14to60E2r3/partitions`, 468 (=36*13) is not one of them, so this test will still fail.

xylar:
@darincomeau, thanks for that!  Yes, multiples of prime numbers > 7 were explicitly excluded.

@rljacob and @jonbob, can we make a separate PR to fix that test so it runs on a "reasonable" number of cores?

xylar:
> A production test on anvil is failing with:
Trying to download file: 'ice/mpas-seaice/WC14to60E2r3/mpas-seaice.graph.info.200714.part.468

@rljacob, how was this test succeeding before?  Nothing I have changed recently should have caused this to happen unless a recently generated partition file got put in `mpas-cice` after I copied files over to `mpas-seaice`.

rljacob:
It last passed on April 13: https://my.cdash.org/test/77893418?graph=status.   Did you move that part file because it does indeed seem to be missing from ice/mpas-seaice/WC14to60E2r3 on both the server dir and the DIN_LOC_ROOT.   @amametjanov , did the pe layout change as a result of https://github.com/E3SM-Project/E3SM/pull/5558 ?

xylar:
@rljacob, I have looked around and I don't think that partition file ever existed.  I certainly would not have removed it.

xylar:
> @amametjanov , did the pe layout change as a result of https://github.com/E3SM-Project/E3SM/pull/5558

Yes it did.

xylar:
@rljacob, this PR is not related to the WC14 meshes.  Can we discuss this on #5611 instead, or make a separate issue since it's not really related to that PR either?

jonbob:
Wait, I did find it in the lcrc local inputdata area:
/lcrc/group/e3sm/data/inputdata/ice/mpas-cice/WC14to60E2r3/mpas-seaice.graph.info.200714.part.468

So it must have failed for other reasons for several days? The change from mpas-cice to mpas-seaice wasn't until April 17

xylar:
@jonbob, it's not in:
```
/lcrc/group/e3sm/public_html/inputdata/ice/deprecated-mpas-cice/WC14to60E2r3
```
@amametjanov, did you put that file directly in `/lcrc/group/e3sm/data/inputdata`, rather than `/lcrc/group/e3sm/public_html/inputdata`?  My understanding it that isn't the "proper" way.

So this got broken when we moved from `mpas-cice` to `mpas-seaice` and the file wasn't there for download.

xylar:
I still think not using 13 nodes would be a nicer fix than me having to generate partition files for 13 nodes, but it's not a big deal if we need this to run on 13 nodes (e.g. for performance reasons).

rljacob:
I asked Az to reduce the number of nodes for test cases on anvil so they would get through overnight.  The exact number is possibly changeable.

jonbob:
@xylar - it does happen that people put files in the local inputdata directory when they're just intended for lcrc. I'm guessing that was a one-off just for anvil

xylar:
@jonbob, I don't think we want that to happen going forward because of the more complicated sea-ice partitioning technique.  But I could see that having been a useful practice in the past. @amametjanov, sorry for giving you a hard time.

amametjanov:
>  I don't think we want that to happen going forward because of the more complicated sea-ice partitioning technique

To prevent folks from generating their own partition files, I'd recommend removing the source `X`, so no one can do `gpmetis X N`.
To reduce the burden of generating new partition files, can the optimized partitioning be wrapped in a one-liner (like gpmetis) within an e3sm-conda script?
I generated the 468-way partition locally for Anvil while experimenting with PEs and forgot to ping you for a proper one afterwards.
In any case, I was planning to revert to the old 29-node PEs for `SMS_PS.northamericax4v1pg2_WC14to60E2r3.WCYCL1850.anvil_intel.allactive-wcprodrrm_1850`, because 20-node PEs are too small for this case (OOM error) and `ice/mpas-seaice/WC14to60E2r3/mpas-seaice.graph.info.200714.part.720` is there.

xylar:
> To prevent folks from generating their own partition files, I'd recommend removing the source X, so no one can do gpmetis X N.

The ocean partition files are still created with `gpmetis X N`, so I don't want to remove the `X` file because that's precisely the file *I* would need to use to make new partition files.

The only reason I don't want other folks making new ocean partition files without letting me know about it is that I'm trying to understand what the needs are more generally so I can hopefully stop generating one-off partition files and instead just generate all the required partition files algorithmically along with new meshes.

> To reduce the burden of generating new partition files, can the optimized partitioning be wrapped in a one-liner (like gpmetis) within an e3sm-conda script?

Yes, I can do that.  I think it would make a lot of sense.  It would presumably only work on LCRC because that's where the data files it needs actually live.  I don't want to reinvent the tools for downloading the data files and all that.

> I generated the 468-way partition locally for Anvil while experimenting with PEs and forgot to ping you for a proper one afterwards.
In any case, I was planning to revert to the old 29-node PEs for SMS_PS.northamericax4v1pg2_WC14to60E2r3.WCYCL1850.anvil_intel.allactive-wcprodrrm_1850, because 20-node PEs are too small for this case (OOM error) and ice/mpas-seaice/WC14to60E2r3/mpas-seaice.graph.info.200714.part.720 is there.

Good to know.  Then, I think we should ignore the missing 468 task partition file for now.

xylar:
I have a draft tool here:
https://github.com/xylar/MPAS-Tools/tree/simplified-seaice-partition-tool
Unfortunately, it wouldn't be super easy to use until it makes it into the next E3SM-Unified release.  But I can provide some simple instructions for installing `mpas_tools` in the meantime.  I'll put these on confluence:
once I've tested the tool and made a new release of `mpas_tools`, probably early next week.

xylar:
@amametjanov, could you try this 4-liner?  This is as good as I can make it until E3SM-Unified gets updated.  This assumes you have mambaforge or some similar version of conda installed, conda-forge as a default channel, and the `mamba` command available.  This is pretty standard for those of us using conda on a regular basis but I realize that's not necessarily true for other parts of the project so let me know if you want instructions on that side of the workflow.  Please don't try to use E3SM-Unified as the starting point for this.  It needs to be your own installation of conda.

```
mamba create -y -n mpas_tools python=3.11 "mpas_tools>=0.20.0" "esmf=8.2.0=nompi*"
mamba activate mpas_tools
cd /lcrc/group/e3sm/public_html/inputfiles/ice/mpas-seaice/<mesh_name>
simple_seaice_partitions -m <mesh_file> -p partitions/mpas-seaice.graph.info.<creation_date> -n <core_count> <core_count> <core_count>
```

xylar:
@jonbob, I believe all the partition files that I produce algorithmically are in place for these 6 meshes.  So I think what remains is to do some Anvil, Chrysalis and Perlmutter testing to make sure (especially for PEM and PET tests) that the partition files are there that should be.  I'm more than happy to help with that testing if you can provide me with some instructions.

jonbob:
@xylar -- I was thinking I could just go make sure whatever graph files that are on those platforms in the old location are also there in these new directories (with new date stamps). Does that sound like enough for you? That way I can also make sure the new directory permissions are correct as well. I'm happy to take that on, unless you're looking for something to do

jonbob:
@xylar -- one quick comment/question. I'm looking at the oQU480 and there are some graph files that have higher counts than that mesh has cells (1791). Will that even work, if some pe's have zero cells?

xylar:
@jonbob, thanks for catching that!  Originally, my algorithm bounded the partitions to have between 2 and 30,000 cells per task but when I added more partition files based on even multiples of node sizes, I forgot to also bound those partition sizes to have a reasonable number of cells per task.

I'm not sure if the partition files where some tasks don't own any cells are a problem.  I suspect that things might still work even if the performance would be very bad.  But I can take the time to delete them if it seems risky to have them around.

jonbob:
Thanks @xylar -- my bigger concern is that those graph files aren't actually doing what you think they are. For example, I was looking at the oQU480 part.5888 graph file thinking there might be one occurrence of a processor number for 1791 processors -- and then some 4000 wouldn't have any cells assigned to them. But something unexpected is happening and I see this:
````
grep -c 1988 mpas-o.graph.info.230422.part.5888
449
````
I saw some processor numbers repeated over and over, so this graph file would distribute mpas cells far worse than a simple part.1791 file would

**Comment by xylar** [comment]
@jonbob, I see.  That's very strange behavior.

**Comment by jonbob** [comment]
I did make partition subdirectories and stage appropriate graph files on perlmutter and compy, with the exception of staging graph files for oRRS18to6v3 because it is not tested with ERS or PEM tests where the case2 files would not be automatically downloaded. I'll make sure lcrc is setup as well

**Comment by jonbob** [comment]
@xylar -- I also had people ask for a part.1 file for some of these small test meshes, but we haven't made them  for this PR?

**Comment by xylar** [comment]
I don't believe they ever get read.  If you ask for one core, MPAS components shouldn't even look for a partition file.  Does E3SM think there should be one?

**Comment by xylar** [comment]
@amametjanov, could you clarify if you ran into errors trying to run on 1 core when you didn't have `inpu/tdata/ocn/mpas-o/oQU480/mpas-o.graph.info.151209.part.1`? For standalone runs of MPAS components, they definitely don't need such a file. Do E3SM and/or CIME require this?

**Comment by jonbob** [comment]
@xylar -- I think E3SM requires them, if only because the scripts will add them to the inputdata list. @singhbalwinder has also requested the part.1 files in the past.

**Comment by amametjanov** [comment]
Yes, IIRC the scripts (or CIME) gave me an error. And I did
```
cp -p mpas-o.graph.info.151209 mpas-o.graph.info.151209.part.1
```

xylar:
Thanks @amametjanov.  It's lucky that MPAS components don't use the the partition file because the `mpas-o.graph.info.151209` file is not the same format as the `mpas-o.graph.info.151209.part.*` files.  The former describes connectivity between cells, whereas the latter files contain a list of task numbers for each cell.  As I understand it, `gpmetis` uses the graph file to make sure adjacent cells are on adjacent processors to the degree possible.

If `gpmetis` lets me, I'll make "proper" 1-task partition files.

jonbob:
passes e3sm_developer on chrysalis with expected NML DIFFs. 
Merged to next

jonbob:
merged to master and expected DIFFs and NML DIFFs blessed. But there were several FAILs on chrysalis, mappy and perlmutter due to the case2run graph files not getting downloaded automatically. I've made a request for the two missing files on mappy to be dowloaded by hand, and staged the missing files on chrysalis and perlmutter, so hopefully those FAILs will clear overnight

jonbob:
blessed DIFFs on chrysalis, mappy and perlmutter that had previously failed due to missing graph files

PR #5641, opened 2023-04-26
Reduce Anvil WCYCL1850 config from 20 to 18 nodes

xylar:
This is 12 atm, lnd, ice, etc. and 6 ocn instead of 13 and 7 (there are not currently ice partition files for 13 Anvil nodes).

[NML] - in cime_pes namelists
[non-BFB] - in mpaso.hist.am.globalStats outputs

### DIFF
--- a/cime_config/testmods_dirs/config_pes_tests.xml
+++ b/cime_config/testmods_dirs/config_pes_tests.xml
@@ -183,17 +183,17 @@
     </mach>
     <mach name="anvil">
       <pes compset=".*EAM.+ELM.+MPASSI.+MPASO.+MOSART.+SWAV.+" pesize="any">
-        <comment>tests+anvil: --compset WCYCL1850 --res northamericax4v1pg2_WC14to60E2r3 on 20 nodes pure-MPI, ~0.8 sypd </comment>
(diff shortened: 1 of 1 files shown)


cime_config/testmods_dirs/config_pes_tests.xml
```diff
@@ -183,17 +183,17 @@
     </mach>
     <mach name="anvil">
       <pes compset=".*EAM.+ELM.+MPASSI.+MPASO.+MOSART.+SWAV.+" pesize="any">
-        <comment>tests+anvil: --compset WCYCL1850 --res northamericax4v1pg2_WC14to60E2r3 on 20 nodes pure-MPI, ~0.8 sypd </comment>
+        <comment>tests+anvil: --compset WCYCL1850 --res northamericax4v1pg2_WC14to60E2r3 on 18 nodes pure-MPI, ~0.8 sypd </comment>
```
xylar, line 186: sypd may need to be updated.

xylar:
@amametjanov and @jonbob, I am fine with creating the partition files if this PR seems like too much trouble.  I would like to avoid using node counts that are large primes like this going forward, though, either way.

xylar:
@jonbob, I'm hoping you can help with testing when you have time.

xylar:
Closing based on https://github.com/E3SM-Project/E3SM/pull/5635#issuecomment-1524177645

PR #5642, opened 2023-04-27
Revert RRM WCYCL1850 from new 20- to old 29-node PEs to avoid OOM

amametjanov:
Revert RRM WCYCL1850 from new 20- to old 29-node PEs to avoid OOM.

[NML] - in cime_pes namelists
[non-BFB] - in mpaso.hist.am.globalStats outputs

### DIFF
--- a/cime_config/testmods_dirs/config_pes_tests.xml
+++ b/cime_config/testmods_dirs/config_pes_tests.xml
@@ -183,17 +183,17 @@
     </mach>
     <mach name="anvil">
       <pes compset=".*EAM.+ELM.+MPASSI.+MPASO.+MOSART.+SWAV.+" pesize="any">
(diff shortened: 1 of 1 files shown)


rljacob:
Why revert this?

amametjanov:
It was running OOM: e.g. on Apr 14--17. Then, on Apr 18 it started failing due to missing seaice partition file.
Prior 29-node PEs were more stable against OOM issues.

rljacob:
But it did have some successful runs at that task count.  See https://my.cdash.org/test/77517732.  Was it on the border and then some addition to the code pushed it over the memory limit?

amametjanov:
I don't see any successful runs in the old-test archives and no "gold-box" on CDash to download run-dirs: to check on mem-usage in runs that passed. I think it was on the border.

Issue #5605, opened 2023-04-19
Error with Intel using threads. Subscript #1 of the array GLCMASK has value 0 which is less than the lower bound of 1152


ndkeen:
Using new Intel compiler on pm-cpu, I'm hitting an error that may only happens with threads.
Can reproduce with `SMS_D.f19_g16.I1850CNECACTCBC.pm-cpu_intel.elm-eca` 
and `SMS_D_Ln5.ne4_oQU240.F2010.pm-cpu_intel` using Apr12th e3sm main.
(still working on the branch to actually bring in the intel compiler for use with pm-cpu).
I am trying to debug to narrow down issue, but wanted to save my work at this point.
It could be an issue with Intel compiler itself, but maybe someone might have more insight based on the error mesg.

I made some mods in `components/elm/src/main/elm_initializeMod.F90` to write some data, but also force only 1 thread at time in the loop, but still see error:
```
    nclumps = get_proc_clumps()
    !$OMP PARALLEL DO PRIVATE (nc, bounds_clump)                                                                                                                                                                                        
    do nc = 1, nclumps
       !$OMP CRITICAL                                                                                                                                                                                                                   
       call get_clump_bounds(nc, bounds_clump)
       write(*,'(a,i10,a,i10,i10)') "ndk nc=", nc, " bounds_clump%begg, bounds_clump%endg=", bounds_clump%begg, bounds_clump%endg
       call reweight_wrapup(bounds_clump, &
            ldomain%glcmask(bounds_clump%begg:bounds_clump%endg)*1._r8)
       !$OMP END CRITICAL                                                                                                                                                                                                               
    end do
    !$OMP END PARALLEL DO   
```

```
13: ndk nc=         2 bounds_clump%begg, bounds_clump%endg=      1196      1239
13: forrtl: severe (408): fort: (3): Subscript #1 of the array GLCMASK has value 0 which is less than the lower bound of 1152
13: 
13: Image              PC                Routine            Line        Source             
13: e3sm.exe           00000000008D9BF0  elm_initializemod         398  elm_initializeMod.F90
13: libiomp5.so        00001489FADF0053  __kmp_invoke_micr     Unknown  Unknown
13: libiomp5.so        00001489FAD5E2F3  Unknown               Unknown  Unknown
13: libiomp5.so        00001489FAD5F578  __kmp_fork_call       Unknown  Unknown
13: libiomp5.so        00001489FAD18223  __kmpc_fork_call      Unknown  Unknown
[...]

PR #5649, opened 2023-05-01
Add the intel compiler to pm-cpu

ndkeen:
Intel compiler was recently installed on pm-cpu and these are changes to allow building with it.
We use the `intel-oneapi` module and the ifx/icx/icpx varities.

There is currently a work-around (only for pm-cpu) to use `-fp-model=precise` for C++ flags instead of
`-fp-model=source` as, for whatever reason, this version of intel C++ compiler does not support.

Another work-around to avoid internal compiler error simply uses `-O0` instead of `-O2` for one source file.

Also, currently, only about 25% of the tests in e3sm_developer are passing. It looks like they all fail due to
https://github.com/E3SM-Project/E3SM/issues/5605

Would not impact any other machines/compiler, so is bfb
[BFB]


rljacob (approved):

ndkeen:
merged to next

xylar:
@ndkeen, thanks very much for putting in this work.  This is a capability I'm keen to try out!

rljacob:
Note that "intel compiler" now means 2 different things.  We should say "ifort" or "ifx" when talking about the "intel compiler".  ifx is on pm-cpu,  ifort is on Chrysalis and compy.

PR #5655, opened 2023-05-03
For pm-cpu, change all default PE layouts to use 1 thread (no threading)

ndkeen:
For all pm-cpu PE layouts, change to use 1 thread (no threading).
This changes all default layouts to use 128 MPI's per node (128x1) instead of 64x2.
This can avoid the issues with threading using the new Intel compiler until addressed.
This is also faster when comparing common ne30/ne120 benchmarks.

As this changes the layouts for all compilers, the tests using GNU & nvidia will also be affected and
those using threads previously will not be BFB on pm-cpu only.
[NBFB]


ndkeen:
Note there are still 3 fails of e3sm_integration using intel -- they are a new build fail -- internal compiler issue. I've confirmed that none of the flags we are using causing the problem. I will look more into it.

rljacob (approved):

ndkeen:
merged to next

rljacob:
> 3 fails of e3sm_integration using intel
Please say "using ifx".  We need to distinguish now since those tests probably pass with ifort.

ndkeen:
I have not seen any real differences between standard intel compiler and oneapi intel compiler.
At NERSC they have `intel` and `intel-oneapi` modules.
And at least for the oneapi module, there are ifort and ifx available. Both seem to work the same.

If we want to distinguish between the two -- maybe for now, we could start a `intel-oneapi` compiler name?
Or simply `oneapi`?  But then, when the oneapi path at Intel has been more established, presumably it will
be the only compiler and we can switch back to simply `intel`.

But yes, the intel on pm-cpu is currently using icx/icpx/ifx.

Issue #5657, opened 2023-05-03
Fortran runtime error: Index '1' of dimension 3 of array 'tmp' above upper bound of 0 in homme


ndkeen:
With `ERS_D.ne4_oQU240.F2010.pm-cpu_gnu.eam-hommexx`

```
 98: At line 443 of file /global/cfs/cdirs/e3sm/ndk/repos/ndk_mf_pm-cpu-pelayout-nothreads/components/eam/src/dynamics/se/inidat.F90
 98: Fortran runtime error: Index '1' of dimension 3 of array 'tmp' above upper bound of 0
```

I think it's another case of compiler evaluating B when A is false `if (A and B)`

In `components/eam/src/dynamics/se/inidat.F90`
```
    !ndk if (scm_multcols .and. tmp(1,1,1) < 10000._r8) then   
    if (scm_multcols) then                                                            
       if (tmp(1,1,1) < 10000._r8) then                                           
          call endrun('Problem reading ps field')                           
       endif                                                                                       
    endif 
```

oksanaguba:
@ndkeen -- do i understand it right that you proposed the fix above and it works? if so, do you want me to create a PR or would you like to do it yourself? thanks.

rljacob:
@ndkeen why don't you make the PR and Oksana can merge it.

ndkeen:
How about I find/fix the error, make issue/PR, and merge it?

oksanaguba:
@ndkeen thanks!

PR #5660, opened 2023-05-03
Fix out-of-bounds error by only looking at array when scm_multcols enabled

ndkeen:
Compiler found out-of-bounds with tmp() array. If instead, we only check that array when scm_multcols is true, it seems OK:
```
    if (scm_multcols) then
       if (tmp(1,1,1) < 10000._r8) then
          call endrun('Problem reading ps field')
       endif
    endif
```
Fixes https://github.com/E3SM-Project/E3SM/issues/5657

[bfb]


ndkeen:
merged to next

PR #5662, opened 2023-05-04
Add ocean linearization options

mark-petersen:
This PR adds two options that may be turned on in order to run the ocean momentum and thickness equation in linearized form.  These two settings are:
1. Momentum equation: Turn off nonlinear advection. This is usually written as `u dot grad(u)` but in the TRiSK formulation, we (equivalently) turn off the terms `grad(KE)` and `eta * uperp` where `eta` is relative vorticity. This is now set with the new flag `config_disable_vel_hadv = .true.`  
2. Thickness equation: The advection term `div(h*u)` is linearized to use a constant layer thickness in time, `div(H*u)`, where `H` is taken from the `restingThickness` variable. This option is set with `config_thickness_flux_type = 'constant'`.
 
Note that the previous flag `config_disable_vel_coriolis` would remain `.false.` for linear tests. Previously the flag `config_disable_vel_coriolis` was inappropriately named and turned off _both_ the coriolis and advection terms. Now `config_disable_vel_coriolis` controls the coriolis term only.

These new options will allow us to compare against analytic solutions to the shallow water equations, such as Kelvin waves and inertia-gravity waves. These are described in [Sid Bishnu's recent paper](https://essopenarchive.org/users/566154/articles/612943-a-verification-suite-of-test-cases-for-the-barotropic-solver-of-ocean-models) and [Sid's thesis](https://diginole.lib.fsu.edu/islandora/object/fsu:795713).

See previous discussion at https://github.com/E3SM-Ocean-Discussion/E3SM/pull/50.

[BFB]
[NML]

### DIFF
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -1145,7 +1145,7 @@
 					possible_values=".true. or .false."
 		/>
 		<nml_option name="config_thickness_flux_type" type="character" default_value="centered"
-					description="If 'upwind', use upwind to evaluate the edge-value for layerThickness, i.e., layerThickEdgeFlux.  The standard MPAS-O approach is 'centered'. For 'constant', use constant thickness in time from restingThickness, for linear test problems"
+					description="If 'upwind', use upwind to evaluate the edge-value for layerThickness, i.e., layerThickEdgeFlux.  The standard MPAS-O approach is 'centered'. For 'constant', uses constant thickness in time from restingThickness, for linear test problems. Note that these two flags are set together for linearized test cases: config_thickness_flux_type = 'constant' linearizes the thickness equation, and config_disable_vel_hadv = .true. linearizes the momentum equation if there is no assumed mean background velocity."
 					possible_values="'upwind', 'centered', 'constant'"
 		/>
 		<nml_option name="config_drying_safety_height" type="real" default_value="1.0e-10" units="m"
@@ -1431,7 +1431,7 @@
 					possible_values=".true. or .false."
 		/>
 		<nml_option name="config_disable_vel_hadv" type="logical" default_value=".false."
-					description="Disables tendencies on the velocity field from the horizontal momentum advection."
+					description="Disables tendencies on the velocity field from the horizontal momentum advection. Note that these two flags are set together for linearized test cases: config_thickness_flux_type = 'constant' linearizes the thickness equation, and config_disable_vel_hadv = .true. linearizes the momentum equation if there is no assumed mean background velocity."
 					possible_values=".true. or .false."
 		/>
 		<nml_option name="config_disable_vel_coriolis" type="logical" default_value=".false."
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -1043,6 +1043,10 @@
 					description="Slope limiter for the vertical remap advection scheme."
 					possible_values="'none','monotonic','weno'"
 		/>
+		<nml_option name="config_linearize_thick_hadv" type="logical" default_value=".false."
+					description="Change horizontal thickness advection from div(hu) to div(Hu) where h was the full velocity field, and H is the mean total depth."
+					possible_values=".true. or .false."
+		/>
 	</nml_record>
 	<nml_record name="bottom_drag" mode="forward">
 		<nml_option name="config_bottom_drag_mode" type="character" default_value="implicit"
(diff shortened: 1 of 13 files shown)


mark-petersen:
Tested with both gnu and intel on chrysalis. For both, using optimized and the nightly test suite, all tests are bfb against master with the default flags (i.e. with this linearization off). When I run with these two flags set for linearization, i.e. `config_disable_vel_hadv = .true.` and `config_thickness_flux_type = 'constant'`, differences after a few timesteps exist but are small, as expected. Maximum differences in the nightly test suite are 1e-7. I also tested the nightly suite with gnu on perlmutter and chicoma.

components/mpas-ocean/src/Registry.xml
```diff
@@ -1133,8 +1133,8 @@
 					possible_values=".true. or .false."
 		/>
 		<nml_option name="config_thickness_flux_type" type="character" default_value="centered"
-					description="If 'upwind', use upwind to evaluate the edge-value for layerThickness, i.e., layerThickEdgeFlux.  The standard MPAS-O approach is 'centered'."
-					possible_values="'upwind', 'centered'"
+					description="If 'upwind', use upwind to evaluate the edge-value for layerThickness, i.e., layerThickEdgeFlux.  The standard MPAS-O approach is 'centered'. For 'constant', uses constant thickness in time from restingThickness, for linear test problems. Note that these two flags are set together for linearized test cases: config_thickness_flux_type = 'constant' linearizes the thickness equation, and config_disable_vel_hadv = .true. linearizes the momentum equation if there is no assumed mean background velocity."
+					possible_values="'upwind', 'centered', 'constant'"
```
sbrus89, line 1137: Does it make sense to move `config_thickness_flux_type` out of the `wetting_drying` section?

cbegeman: @sbrus89 Thanks for pointing this out. Yes, I was thinking about doing this with a PR I haven't opened yet to use `config_thickness_flux_type = 'upwind'` with split-explicit. I suppose we could do that in this PR if @mark-petersen is up for it.

sbrus89:
@mark-petersen, I've tested this PR with the new inertial gravity wave case in polaris (https://github.com/E3SM-Project/polaris/pull/62) and verified that these linearization options result in 2nd order convergence. Thanks very much for adding this!

components/mpas-ocean/src/Registry.xml
cbegeman: @mark-petersen I agree!

sbrus89: I agree as well.

mark-petersen:
I moved `config_thickness_flux_type` into the `advection` namelist as requested. Then rebased and tested by compiling with gnu on chicoma and intel on chrysalis. Both passed with both debug and optimized for stand-alone nightly test suite using default flags.

I also updated the E3SM namelist scripts using the `/generate_e3sm_namelist_files.py` script. Then tested and passed:
```
SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.chrysalis_gnu
SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.chrysalis_intel
```

@sbrus89 thanks for testing this branch with your new test case here: https://github.com/E3SM-Project/polaris/pull/62. 

@sbrus89 and @cbegeman this is ready for your final review.

cbegeman (approved):
@mark-petersen This all looks great! Thanks for making that final namelist change. It sounds like you've done enough testing that we don't need to retest with the IGW case, but if you would like me to rerun that case, let me know!

sbrus89 (approved):
@mark-petersen, I tested this one more time on Perlmutter with the latest namelist change using the inertial gravity wave case (https://github.com/E3SM-Project/polaris/pull/62). Everything looks good to go. Thanks again for putting this together!

sbrus89:
@jonbob, this one is ready to go.

jonbob:
@mark-petersen -- I ran the scripts to make the bld files consistent with Registry changes and found that 
````
components/mpas-ocean/bld/namelist_files/namelist_definition_mpaso.xml
````
needs to be updated

jonbob:
passes:
* PET_Ln9_PS.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel.allactive-mach-pet

jonbob:
merged to master -- but github is not showing this PR as merged. There were no conflicts or issues during the merge -- does anyone have ideas?

rljacob:
Did you merge to master locally and not push it?

rljacob:
I see it on master in my updated clone so yes you did.  Not sure.

jonbob:
I was just going back and double-checking. Very odd. The only thing suspicious is that this page says it has conflicts that must be resolved, even though there's weren't any conflicts during the merge to next or master

Issue #5667, opened 2023-05-05
PIO may not be setting stripe count for output files on pm-cpu/pm-gpu


ndkeen:
For example, with this case:
```
login18% pwd
/pscratch/sd/n/ndk/e3sm_scratch/pm-cpu/se62-mar27/iotest.se62-mar27.F2010-SCREAMv1.ne120pg2_ne120pg2.pm-cpu.n085t128x1cX.L128.vth200.1m

login18% lfs getstripe run//iotest.se62-mar27.F2010-SCREAMv1.ne120pg2_ne120pg2.pm-cpu.n085t128x1cX.L128.vth200.1m.scream.r.0001-02-01-00000.nc
run//iotest.se62-mar27.F2010-SCREAMv1.ne120pg2_ne120pg2.pm-cpu.n085t128x1cX.L128.vth200.1m.scream.r.0001-02-01-00000.nc
lmm_stripe_count:  1
lmm_stripe_size:   1048576
lmm_pattern:       raid0
lmm_layout_gen:    0
lmm_stripe_offset: 81
        obdidx           objid           objid           group
            81         2327999       0x2385bf     0x20c0000416
```

This PR first added stripe setting, but it's possible something changed on the machine.
https://github.com/E3SM-Project/scorpio/pull/488/files

PR #5668, opened 2023-05-06
For pm-cpu and pm-gpu add lustre filesystem hint

ndkeen:
Any perlmutter run (that uses lustre scratch for output), could benefit from file stripping.
Add set(PIO_FILESYSTEM_HINTS "lustre") for all pm-cpu or pm-gpu machines.

[bfb]


ndkeen:
May fix https://github.com/E3SM-Project/E3SM/issues/5667

dqwu:
@ndkeen 
I think issue #5667 might be related to some temporary filesystem issues on Perlmutter. You might rerun the original case a few more times to see if that unexpected file striping issue still persists.

ndkeen:
I also did not see this setting change anything.  I do not think it's a temporary filesystem issue.  I can reproduce the issue, but only with certain case. Will add comment to the original issue as closing this PR since we dont need these settings (and can remove them from other places)

PR #5671, opened 2023-05-09
For pm-cpu intel builds, add Depends rule to build one file slightly different to avoid internal compiler error

ndkeen:
For pm-cpu intel build, which uses new oneapi version,
add Depends rule to build one certain mpas file with ifort instead of ifx as a work-around for internal compiler error

May fix https://github.com/E3SM-Project/E3SM/issues/5672
[bfb]


ndkeen:
merged to next

rljacob:
Can you open an issue about that ICE so we can track it ?

ndkeen:
reverted as there is no `-fc` flag available to `ftn` and it resulted in a warning

PR #5673, opened 2023-05-09
Adds nersc inptudata server

jasonb5:
Adds https://portal.nersc.gov/project/e3sm/inputdata to config_input.xml.


jgfouca (approved):

rljacob:
Does the schema for config_inputdata.xml allow a alias or other way to name different servers?

ndkeen (approved):
I can test this in an hour or so, but looks good

ndkeen:
Whoops, meant to add here that I did test it and seems to work great -- downloaded data to local case from the CFS location.

PR #5679, opened 2023-05-10
Updates BeTR submodule in maint-1.0 to avoid compiler error

### LINKED ISSUE #5280: Add ability to build/run maint-1.0 tests on pm-cpu with GNU
I have some changes that allow choice tests in maint-1.0 to build/run on pm-cpu. `ndk/maint-1.0/add-pm-cpu`

These run routinely on cori-knl:
```
SMS_Ld2.ne30_oECv3_ICG.A_WCYCL1850S_CMIP6.cori-knl_intel.allactive-v1cmip6
SMS_Ld2.ne30_oECv3_ICG.A_WCYCL20TRS_CMIP6.cori-knl_intel.allactive-v1cmip6
SMS_Ln5.ne30_ne30.FC5AV1C-L.cori-knl_intel.cam-cosplite
```

And I can run them on pm-cpu with basic machine file changes and a few other mods.

There are 2 changes in files under submodules that are not easy to change.
These are changes needed with higher version of GNU (all that is avail on PM)

Need a change here to build. this is already done in master, but there are other changes we don't need --
can just update source with the interface changes needed
`components/clm/src/external_models/sbetr/src/betr/betr_math/ODEMod.F90`

needed to work-around issue with solver during runtime. this change modifies compiler
flags for ALL MPAS sources which is not necessary, but unable to only affect one source in MPAS
in current master, we build one MPAS source file with a compiler flag to turn off very specific optimization,
however, in maint-1.0, the CMake logic is different and I can't do the same thing in Depends.
But what I can do is add a flag here (which may need more guards) that turns off the optimization
for all MPAS source files.
`components/mpas-source/src/Makefile.in.ACME`

Another odd thing to make this work on pm-cpu:
With current master, we use a trick to distinguish between diff NERSC machines, but with this branch, I made the following hack, which would look for "login" as the login name.  This is working, but not ideal as it could conflict with a different machine also expecting "login".
```
  <!--NODENAME_REGEX>$ENV{NERSC_HOST}:perlmutter</NODENAME_REGEX-->
  <NODENAME_REGEX>login</NODENAME_REGEX>
```

I also found 2 issues with DEBUG attempts that can be avoided with 2 simple changes in e3sm source files.
Here are two places that were failing with runtime segfault in DEBUG due to the compiler looking-ahead
and trying to parse an experession B that should only be evaluated if A is true.
`components/mosart/src/riverroute/RtmMod.F90`
`components/cam/src/physics/cam/micro_mg2_0.F90`
We can decide how important fixes like these are. I don't see a fix in current master
at these locations, but it's possible the code flow doesn't happen same way in maint-1.0

Note: we may not have actually supported GNU compilers with maint-1.0.

I've added a `notes.txt` file to the branch temporarily to explain above and make it easier to make the changes.

bishtgautam:
The BeTR submodule in maint-1.0 is updated to avoid errors with new compilers.
The updated hash in BeTR is d126b6af and it corresponds to the `e3sm-maint-1.0`
branch in BeTR.

[BFB]
Fixes #5280


jinyuntang (approved):

ndkeen (approved):

ndkeen:
SO I can just make this merge?  That is, it will just pick up the new sbetr version?

bishtgautam:
Yes, merge the PR locally. Then, run the tests. If everything looks good, push your merge commit to Github.

bishtgautam:
Forgot to mention. After merge, you will have to do `git submodule update`

ndkeen:
Hmm, I still see the ODEMod.F90 as being the same when I check out the branch?

rljacob (approved):

rljacob:
you have to update submodules after checking out the branch.

ndkeen:
I did that.
first checkout maint-1.0, then checkout `bishtgautam/lnd/updates-betr`.
```
git checkout bishtgautam/lnd/updates-betr`
git submodule update --recursive --init
git checkout maint-1.0
git submodule update --recursive --init
git merge --no-ff bishtgautam/lnd/updates-betr
```
?
At that point, would expect the changes to be there?

rljacob:
I meant update the submodules after the merge of the  dev branch.  Sorry.

ndkeen:
Oh, ok
```
git submodule update --recursive --init
Submodule path 'components/clm/src/external_models/sbetr': checked out 'd126b6af2c81b22f9157e6a184e6513711043243'
```
which indeed does have the changes

Issue #5680, opened 2023-05-11
segfault with ERP_Ln9.ne4pg2_ne4pg2.F2010-MMF1.pm-cpu_intel.eam-mmf_fixed_subcycle


ndkeen:
Using next of May9th,
`ERP_Ln9.ne4pg2_ne4pg2.F2010-MMF1.pm-cpu_intel.eam-mmf_fixed_subcycle`

```
  0: forrtl: severe (174): SIGSEGV, segmentation fault occurred
  0: Image              PC                Routine            Line        Source
  0: libpthread-2.31.s  000014CAB0DC38C0  Unknown               Unknown  Unknown
  0: e3sm.exe           00000000022B0A70  Unknown               Unknown  Unknown
  0: e3sm.exe           00000000022AF74F  Unknown               Unknown  Unknown
  0: e3sm.exe           00000000022AF122  Unknown               Unknown  Unknown
  0: .                  00000000022AD2BC  Unknown               Unknown  Unknown
  0:
  0: Stack trace terminated abnormally.
  0: Pool Memory High Water Mark:       23568768
  0: Pool Memory High Water Efficiency: 0.0219501
  0: ERROR: For the pool allocator labeled "^@^@^@^@^@^@^@^@AKL's primary memory pool":
  0: ERROR: Trying to free an invalid pointer
  0: This means you have either already freed the pointer, or its address has been corrupted somehow.
  0: terminate called after throwing an instance of 'std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >'
srun: error: nid004189: task 0: Aborted
srun: Terminating StepId=8636163.0
srun: error: nid004189: tasks 1-127: Segmentation fault

```

These pass: 
```
SMS_D_Ln9.ne4pg2_ne4pg2.F2010-MMF1.pm-cpu_intel.eam-mmf_fixed_subcycle
SMS_Ln9.ne4pg2_ne4pg2.F2010-MMF1.pm-cpu_intel.eam-mmf_fixed_subcycle
```

This fails `ERP_D_Ln9.ne4pg2_ne4pg2.F2010-MMF1.pm-cpu_intel.eam-mmf_fixed_subcycle`
but with a different error -- same as this https://github.com/E3SM-Project/E3SM/issues/5605

Issue #5672, opened 2023-05-09
internal compiler error on pm-cpu with intel for bld/cmake-bld/core_seaice/model_forward/mpas_seaice_core_interface.f90 


ndkeen:
Using `ifx (IFORT) 2023.0.0 20221201` on pm-cpu, there are many build fails that are because of ICE.
I'm actually unable to reproduce the error with same next (I think Jon also reported he was unable to reproduce).
However, I did encounter the same error in previous days -- ie not exactly intermittent, but clearly there is
something odd with environment and the compiler.  When I was able to reproduce, I was unable to find an offending
flag -- same error with no flags passed. But as an attempt, I tried using the "normal" intel compiler (`ifort`) which
did work.  And we see in one Depends file there is a rule to build this very file with `fc=ifort` instead.  So I made PR to add that rule to pm-cpu build.

```
/opt/intel/oneapi/compiler/2023.0.0/linux/bin-llvm/xfortcom[0x1d028ee]
/lib64/libc.so.6(__libc_start_main+0xef)[0x145b19c5129d]
/opt/intel/oneapi/compiler/2023.0.0/linux/bin-llvm/xfortcom[0x1982da9]

/pscratch/sd/e/e3smtest/e3sm_scratch/pm-cpu/SMS.ne4_oQU240.F2010.pm-cpu_intel.eam-cosplite.C.JNextIntegration20230508_205139/bld/cmake-bld/core_seaice/model_forward/mpas_seaice_core_interface.f90: error #5633: **Internal compiler error: segmentation violation signal raised** Please report this error along with the circumstances in which it occurred in a Software Problem Report.  Note: File and line given may not be explicit cause of this error.
Target CMakeFiles/atm.dir/__/__/eam/src/physics/cam/phys_grid.F90.o built in 4.570011 seconds
/global/common/software/nersc/pm-2022q4/spack/linux-sles15-zen/cmake-3.24.3-k5msymx/bin/cmake -E cmake_copy_f90_mod cmake/atm/phys_grid.mod cmake/atm/CMakeFiles/atm.dir/phys_grid.mod.stamp IntelLLVM
compilation aborted for /pscratch/sd/e/e3smtest/e3sm_scratch/pm-cpu/SMS.ne4_oQU240.F2010.pm-cpu_intel.eam-cosplite.C.JNextIntegration20230508_205139/bld/cmake-bld/core_seaice/model_forward/mpas_seaice_core_interface.f90 (code 3)
[...]

Issue #5664, opened 2023-05-05
Strange error with ERP_Ln18.ne4_oQU240.F2010.pm-cpu_intel


ndkeen:
With recent nexts, I'm seeing a fail with`ERP_Ln18.ne4_oQU240.F2010.pm-cpu_intel`.
It might not happen every time

```
  0: forrtl: severe (174): SIGSEGV, segmentation fault occurred
  0: Image              PC                Routine            Line        Source
  0: libpthread-2.31.s  0000153B286578C0  Unknown               Unknown  Unknown
  0: e3sm.exe           0000000002FD7CD0  Unknown               Unknown  Unknown
  0: e3sm.exe           0000000002FD69AF  Unknown               Unknown  Unknown
  0: e3sm.exe           0000000002FD6382  Unknown               Unknown  Unknown
  0: ^A                  0000000002FD451C  for_alloc_copy        Unknown  Unknown
  0:
  0: Stack trace terminated abnormally.
  0: forrtl: severe (174): SIGSEGV, segmentation fault occurred
  0: Image              PC                Routine            Line        Source
  0: libpthread-2.31.s  0000153B286578C0  Unknown               Unknown  Unknown
  0: e3sm.exe           0000000002FA8FC4  Unknown               Unknown  Unknown
  0: e3sm.exe           0000000002FC18A5  Unknown               Unknown  Unknown
  0: e3sm.exe           0000000002FC1AB1  Unknown               Unknown  Unknown
  0: e3sm.exe           00000000014F7FE1  ~SharedAllocation         309  Kokkos_SharedAlloc.hpp
  0: e3sm.exe           0000000002FCA7B1  Unknown               Unknown  Unknown
  0: e3sm.exe           000000000151C876  ~IslMpi                   447  compose_slmm_islmpi.hpp
  0: e3sm.exe           000000000151839C  ~__shared_ptr             154  shared_ptr_base.h
  0: libc-2.31.so       0000153B27D2FAE9  Unknown               Unknown  Unknown
  0: libc-2.31.so       0000153B27D2FC7A  Unknown               Unknown  Unknown
  0: ^A                  0000000002FDCA4B  for__issue_diagno     Unknown  Unknown
```

The ERP_D test fails with a different error https://github.com/E3SM-Project/E3SM/issues/5605

[...]

PR #5684, opened 2023-05-12
For pm-cpu only, use the regular intel module instead of intel-oneapi

ndkeen:
Using `intel` instead of `intel-oneapi` avoids several issues (so from ifx 2023.0.0  to ifort 2021.8.0)

```
module load intel-oneapi
login15% ftn --version
ifx (IFORT) 2023.0.0 20221201

module load intel
login15% ftn --version
ifort (IFORT) 2021.8.0 20221119
```

Remove a work-around that was needed for intel-oneapi. 
Add option to use preempt queue (for both pm-cpu and pm-gpu).
Not BFB with Intel compiler as it changes the version used.

Fixes https://github.com/E3SM-Project/E3SM/issues/5672
Fixes https://github.com/E3SM-Project/E3SM/issues/5664
Fixes https://github.com/E3SM-Project/E3SM/issues/5605
Fixes https://github.com/E3SM-Project/E3SM/issues/5680

[NBFB]

### DIFF
--- a/cime_config/machines/Depends.pm-cpu.intel.cmake
+++ b/cime_config/machines/Depends.pm-cpu.intel.cmake
@@ -1,13 +1,14 @@
-# For this file, see internal compiler error on pm-cpu with -O2
-set(NOOPT
-  eam/src/physics/cam/debug_info.F90)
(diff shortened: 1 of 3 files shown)


ndkeen:
merged to next

PR #5687, opened 2023-05-15
adding o3 surface loss for V2 chemistry configuration

keziming:
This PR is to add O3 surface loss when user chose to use linoz-v2 + prescribed chemistry instead of linoz-v3 + chemUCI in E3SM-V3 code base. The proposed modifications produce stable TCO, as well as other climate parameters. 

Fixes https://github.com/E3SM-Project/E3SM/issues/5701
[non-BFB]

------------------------
The 50-year test simulation can be found at  https://acme-climate.atlassian.net/wiki/spaces/ATMOS/pages/3785392374/Ozone+Surface+Loss+Fix+for+V2+configuration+Chrysalis+test2

### DIFF
--- a/components/eam/src/chemistry/mozart/lin_strat_chem.F90
+++ b/components/eam/src/chemistry/mozart/lin_strat_chem.F90
@@ -178,11 +178,12 @@ end subroutine linoz_readnl
     no2_ndx  =   get_spc_ndx('NO2')
     hno3_ndx =   get_spc_ndx('HNO3')
 
-   !if(masterproc)then
+   if(masterproc)then
    ! write(iulog,*)'in subroutine lin_strat_chem_inti'
    ! write(iulog,*)'O3LNZ=',o3lnz_ndx,'N2OLNZ=',n2olnz_ndx,'NOYLNZ=',noylnz_ndx,'H2OLNZ=',h2olnz_ndx
    ! write(iulog,*)'O3=',o3_ndx,'CH4=',ch4_ndx,'N2O=',n2o_ndx,'NO=',no_ndx,'NO2=',no2_ndx,'HNO3=',hno3_ndx
-   !end if
+     write(iulog,*)'inside lin_strat_solve for ndx o3, o3lnz, n2o, noylnz, ch4', o3_ndx, o3lnz_ndx, n2o_ndx, noylnz_ndx, ch4_ndx
+   end if
     
 !     
 !   initialize the linoz data
@@ -963,7 +964,16 @@ end subroutine linoz_readnl
        mw(1) =  adv_mass(o3lnz_ndx)
        sfc_const(1,:)= o3_sfc
        o3_lbl=4
+
+     elseif (o3_ndx > 0) then
+       ms=1
+       nx(1) =  o3_ndx
+       mw(1) =  adv_mass(o3_ndx)
+       sfc_const(1,:)= o3_sfc
+       o3_lbl=4
+       !write(iulog,*)'warning: linoz_v2 is on and surface o3 loss is active'
      else
+       write(iulog,*)'warning: linoz_v2 is on but no surface o3 loss'      
        return
      endif
     endif


tangq (approved):

wlin7 (approved):
The changes look good. Back compatibility is ensured for linozv2 regardless of using O3 or O3LNZ.

wlin7:
@mahf708 , the back compatibility is in terms of code functionality, which this fix does ensure.

For your interest, I did [an MVK test using this branch](https://web.lcrc.anl.gov/public/e3sm/ac.wlin/evv/MVK_PS.ne4_oQU240.F2010.chrysalis_intel.C.20230517_110416_2k0lpy/index.html), the ensemble results are deemed statistically identical to baseline. The baseline was last blessed on 2022-11-18.

@rljacob , a curious question.

rljacob:
I don't know but will look in to it.  Its good that MVK is again saying it matches.  The test will hopefully start passing when this is merged to next.  note: @salilmahajan 

One thing I don't understand:  Is the F2010 case used in MVK using " linoz-v2 + prescribed chemistry "  ?  I thought only special cases had chemistry.  Is it always on by default?

wlin7:
> I don't know but will look in to it. Its good that MVK is again saying it matches. The test will hopefully start passing when this is merged to next. note: @salilmahajan

Yeah, I expect so. I plan to merge this one tomorrow. Today to merge #5478, which is NBFB for non-DEBUG tests on some machines.

wlin7:
> @wlin7 I don't know how to interpret the MVK tests (what's the baseline in this case?). I will try to look for docs on confluence. But I will take your word for it 😄

@mahf708 , the baselines is the same MVK ensemble test from an older code base (in this case, back in 2022-11). I didn't use this test much before. The current round of integration clearly shows the MVK test is quite effective in detecting CC anomalies. The [test result page](https://web.lcrc.anl.gov/public/e3sm/ac.wlin/evv/MVK_PS.ne4_oQU240.F2010.chrysalis_intel.C.20230517_110416_2k0lpy/index.html) contains some reference.

salilmahajan:
> @rljacob , a curious question. The [MVK test on the dashboard](https://my.cdash.org/test/77968293?graph=status) shows FAIL between 2023-04-14 and 2023-04-20 prior to chemUCI PR, though towards the end of page it indicates BASELINE PASS and the ensembles are statistically identical. Why the test is still flagged as FAIL?

This is likely due to a cime issue. @mkstratos, you may want to chime in here.

rljacob:
@wlin7 please make an issue describing the problem.

wlin7:
Merged to next.

wlin7:
Merged to master.

rljacob:
The MVK test still says the climate is different.

> 2023-05-22 01:32:16: BASELINE FAIL for test 'JNextAtm_nbfb20230522_003515'.
>     Test status: fail; Variables analyzed: 121; Rejecting: 15; Critical value: 13; Ensembles: statistically different

But the number rejected is smaller then when this test first had a baseline file

> 2023-05-16 01:25:36: BASELINE FAIL for test 'JNextAtm_nbfb20230516_003403'.
>     Test status: fail; Variables analyzed: 121; Rejecting: 61; Critical value: 13; Ensembles: statistically different

@salilmahajan is 2 rejections over the critical value within the margin of error?

wlin7:
The current MVK test report from next is also different from [the report using the branch](https://web.lcrc.anl.gov/public/e3sm/ac.wlin/evv/MVK_PS.ne4_oQU240.F2010.chrysalis_intel.C.20230517_110416_2k0lpy/index.html). In that report, it said the ensemble was statistically identical (rejecting 10).

mkstratos:
@rljacob Two over is within the margin of error right now, since we think the critical value is too low currently (set using v1). Also, since the ensemble generated is always different from the baseline, occasionally (~5% of tests) there will be a failure by rejecting a few variables more than the critical threshold.

wlin7:
Nice to know @mkstratos . And the test behavior we have seen so far appears to be within the expected fluctuation range.

@rljacob , I feel it ok to bless the test. Since it is a FAIL rather than a DIFF, can it still be blessed in the regular way?

wlin7:
Just to note that with no NBFB changes in codes, today's MVK test reports PASS, with 9 rejected.

rljacob:
Lets move the MVK discussion to #5695

rljacob:
Actually, I'm going to open a new issue about this.

rljacob:
MVK test results discussion moved to #5716

PR #5698, opened 2023-05-18
Add data ice-shelf melt fluxes (DISMF) configuration

xylar:
This merge adds a new package and mode for data melt fluxes. When `config_land_ice_flux_mode = 'data'`, land-ice freshwater and sensible-heat fluxes are assigned from the prescribed fields, which can be read in at initialization or from a stream.


xylar:
See https://github.com/E3SM-Ocean-Discussion/E3SM/pull/45 for relevant discussion.

xylar:
I ran 5-year EC and SORRM tests.  Analysis can be found here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis/20230515.GMPAS-IAF-DISMF.T62_ECwISC30to60E2r1.chrysalis/
and here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis/20230516.GMPAS-IAF-DISMF.T62_SOwISC12to60E2r4.chrysalis/

xylar:
@jonbob, perhaps we can have a slack discussion about what test(s) I should run.  A 10 year EC B-case, I think.  The question is what the baseline should be.  Presumably PISMF?

xylar:
@darincomeau and @cbegeman, do you have opinions on what 10-year run should be done to decide if this stealth feature is climate changing or not?  I originally thought a B-case with DISMF, with a PISMF as a the baseline but I actually think we want to compare DISMF with a simulation without ice-shelf cavities because that is the most important case we want to potentially replace.  Obviously, we would encourage the WC group to make a longer comparison run as part of their tuning exercise to make sure they're okay with using DISMF.

jonbob:
@xylar -- I agree that it makes sense to compare with the configuration this should eventually replace. But maybe the PISMF test as well, just to cover all the bases?

cbegeman:
@xylar and @jonbob I also agree. I look forward to seeing the results!

xylar:
Okay, I'm working on 3 runs:
* B-Case without ice-shelf cavities (or DIB), usual WC: 
  `20230519.WCYCL1850.ne30pg2_EC30to60E2r2.chrysalis`
* B-Case with ice-shelf cavities (PISMF + DIB), usual Cryo: 
  `20230519.CRYO1850.ne30pg2_ECwISC30to60E2r1.chrysalis`
* B-Case with ices-helf cavities (DISMF + DIB), proposed new WC: 
  `20230519.CRYO1850-DISMF.ne30pg2_ECwISC30to60E2r1.chrysalis`

cime_config/allactive/config_compsets.xml
```diff
@@ -320,17 +320,27 @@
...
+
+<compset>
+  <alias>CRYO1850-DISMF</alias>
+  <lname>1850SOI_EAM%CMIP6_ELM%SPBC_MPASSI%DIB_MPASO%IBDISMF_MOSART_SGLC_SWAV</lname>
+</compset>
+
+<compset>
+  <alias>CRYO1950-DISMF</alias>
+  <lname>1950SOI_EAM%CMIP6_ELM%SPBC_MPASSI%DIB_MPASO%IBDISMF_MOSART_SGLC_SWAV</lname>
 </compset>
```
xylar, line 344: @darincomeau and @jonbob, there would be no value in defining `WCYCL1850-DISMF` and `WCYCL1950-DISMF` with identical `lname`s to these, right?  That is, there isn't any way the alias is used internally in the code to change namelist options or anything like that that isn't already in the long name, right?  I'm 90% sure that's the case but wanted to check.

xylar:
Okay, the analysis is in:

[WC](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis/20230519.WCYCL1850.ne30pg2_EC30to60E2r2.chrysalis/)
[Cryo-PISMF](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis/20230519.CRYO1850.ne30pg2_ECwISC30to60E2r1.chrysalis/) -- the usual Cryo with prognostic melt
[Cryo-DISMF](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis/20230519.CRYO1850-DISMF.ne30pg2_ECwISC30to60E2r1.chrysalis/) -- the new data melt

[Cryo-DISMF vs. WC](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis/20230519.cryo-dismf_vs_wc/) -- transects are not included because of different meshes
[Cryo-DISMF vs. Cryo-PISMF](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis/20230519.cryo-dismf_vs_cryo/)

I didn't see much that stood out to me except sea-floor properties around Antarctica:
image
Much better in the DISMF run (left) than WC (right)

image
DISMF is slightly better than PISMF for Amery (not surprising because active melt is exacerbating he bias we have there)

cime_config/allactive/config_compsets.xml
xylar: @darincomeau and @jonbob, in particular what I'm wondering about is where the difference between Cryo's and WC's GM parameters get set.  Are these tied to (P)ISMF being on?

jonbob: @xylar -- some other component's settings go off the date prefix

jonbob: @xylar - right now all the GM parameter settings are based on the ocn_grid and not tied to (P)ISMF. However, if we're trying to move to all ocean grids having ice shelf cavities by default, then we'll have to implement the different setting based on something besides grid name

xylar:
@cbegeman, @darincomeau and all, I need to run a new simulation with DISMF but WC settings for a better comparison with the WC configuration, now that I understand where the different settings come in (based on the ocean/sea ice grid).

cime_config/allactive/config_compsets.xml
darincomeau: a bit late here but I agree there wouldn't be value in defining `WCYCL1850-DISMF`; since we don't have a data iceberg file for the non-wISC meshes (as of now), then that compset paired with non-wISC grids would be pretty wacky wrt to AIS runoff (snowcapping disabled, but only partly prescribed).

This brings up a point that we should double-check that indeed AIS snowcapping is disabled in these `DISMF` compsets. I suspect it is, but an easy way to check is that the `removedIceRunoff`, `removedRiverRunoff` variables in the files output by the ocean conservation analysis member (now on by default for any `wISC` grids) are non-zero in runs with these compsets.

components/mpas-ocean/bld/build-namelist
```diff
@@ -771,6 +771,8 @@ if ($OCN_ISMF eq 'coupled') {
 	add_default($nl, 'config_land_ice_flux_mode', 'val'=>"coupled");
 } elsif ($OCN_ISMF eq 'internal') {
 	add_default($nl, 'config_land_ice_flux_mode', 'val'=>"standalone");
+} elsif ($OCN_ISMF eq 'data') {
```
darincomeau, line 774: I was looking through how we changed namelist settings through compsets with the data icebergs (grep around for `iceberg_mode`,  `MPASSI_ICEBERG_MODE`, `_MPASSI%.*DIB`), and we did it slightly differently, but this works equally well. If there is a need to change other namelist values based on `DISMF` we can explore that, but I think based on the discussion so far, you don't want things (e.g. GM) to change based solely on `DISMF`.

cime_config/allactive/config_compsets.xml
xylar: Thanks @darincomeau, that's helpful.

What I would like a `WCYCL1850-DISMF` to be is the same namelist options as a `WCYCL1850` but using the ECwISC mesh, DIB and DISMF.  Apparently, we would have to think about how to specify this.  I don't think the approach of trying certain parameter choices that are more appropriate for science campaigns instead to specific meshes will work going forward.  As @jonbob says, we will need to figure out another way (e.g. another flag in the long name) to control different namelist options for the same mesh.

components/mpas-ocean/bld/build-namelist
xylar: No, that's right.  What I want is for WCYCL configurations to move to the ECwISC meshes and have DISMF and DIB, but to otherwise retain the same settings.  It would also be reasonable to run CRYO configs with DISMF, so I don't want DISMF to control any settings other than this one namelist option (and the stream to bring in the data it needs).

cime_config/allactive/config_compsets.xml
darincomeau: >What I would like a WCYCL1850-DISMF to be is the same namelist options as a WCYCL1850 but using the ECwISC mesh, DIB and DISMF. Apparently, we would have to think about how to specify this.

Yeah, in general the compset is largely agnostic of the grid, and vice-versa. The (exception) analogy would be how currently the `CRYO` compsets can only be run with `wISC` meshes, and users would get an error in trying to build/run a `CRYO1850.ne30pg2_EC30to60E2r2` case, for example. But we don't have a way to stop the user from trying to do that.

xylar: Right, but because we've chosen to tie GM parameters (like N2 vs. constant) to meshes rather than compsets, there currently isn't a way to get the WCYCL compset with the ECwISC mesh without also getting the CRYO (rather than the WCYCL) GM coefficients.

rljacob: Yes that's by design since so many models can run at multiple resolutions.  A case is a compset and a resolution.  CIME doesn't contain any internal logic to capture invalid compset-resolution combinations.  That logic could be put in the buildnml for the model.

xylar: Yes, but I think there are some config options that need to be tied to compaset *and* grids, and are currently only tied to grids.  We'll need to chat about this tomorrow at the MPAS DevOps meeting.

cbegeman:
@xylar Let me know if you'd like me to review now, otherwise I'll wait until the config option discussion is resolved.

xylar:
@cbegeman, I still need to rerun a 10-year run with WC config options but ECwISC mesh, so it's too soon to review.  Thanks for asking!

xylar:
@cbegeman and @darincomeau, this is ready for you to have a look (but presumably after the holiday weekend).

cbegeman (approved):
@xylar The simulation results look great! Thanks so much for adding this capability! The code looks fine to me by visual inspection.

cime_config/allactive/config_compsets.xml
```diff
@@ -320,17 +320,27 @@
...
 
 <compset>
```
darincomeau, line 346: @xylar there are three more `CRYO` compsets that still have the `IBISMF` tag in the compset long name. These were brought in https://github.com/E3SM-Project/E3SM/pull/5616, which was only merged a month ago, so these changes probably snuck in when you rebased.

xylar: Sorry, yes, you're right that I missed that when I rebased.  Thanks for catching those!

xylar: Should be fixed.

xylar:
@darincomeau, could you re-review this when you have time?

darincomeau (approved):
Approve based on visual inspection and developer testing. I also ran a `SMS.ne30pg2_SOwISC12to60E2r4.CRYO1950.chrysalis_intel` case for good measure.

Thanks @xylar ! This is great, very excited to have this capability!

xylar:
Thanks very much @darincomeau.  I appreciate the extra testing!

jonbob:
@xylar -- I ran the scripts that automate making the bld files consistent with the Registry changes and there are some minimal differences, but I wanted to check one with you. You've added a comment to the namelist_defaults_mpaso.xml file:
````
<!-- These values of `config_land_ice_flux_mode` will be used when `ocn_ismf` is "none" -->
````
but it's not in Registry. Do you want me to add it there?

**Comment by xylar** [comment]
@jonbob, I see, I guess I didn't fully understand that `namelist_defaults_mpaso.xml` was automatically generated rather than manually edited.  That comment doesn't make sense in `Registry.xml`, it's specific to the E3SM side of things.  If it can't be in `namelist_defaults_mpaso.xml` without being in `Registry.xml`, it should be removed entirely.

**Comment by xylar** [comment]
Could you make that change as part of other changes you're making for consistency in what your scripts generate?

**Comment by jonbob** [comment]
@xylar - yes, the namelist definitions and defaults files are at least partially auto-generated. I'll make the changes and push them to your branch

**Comment by xylar** [comment]
@jonbob, thanks a lot.

**Comment by jonbob** [comment]
@xylar -- if we're going to consider this a stealth feature, we should add a new test to capture it

**Comment by xylar** [comment]
@jonbob, I looked into that and it isn't clear to me this counts as a stealth feature.  It's really a new compset.  It makes sense for there to be a test for it but it didn't fit into the existing tests for stealth features when I looked into adding one, at least not in a way that was obvious to me.  There's no namelist option to change so no need for any of the infrastructure involved in the other stealth tests.  I'm happy to drop the stealth flag and instead to put a test in wherever is appropriate, I just wasn't sure where that should be.

File: cime_config/tests.py [hunk 4]
```diff
@@ -220,16 +220,18 @@
...
 
     "e3sm_ocnice_extra_coverage" : {
         "inherit" : ("e3sm_ocnice_stealth_features"),
         "tests" : (
-            "ERS_P480_Ld5.T62_ECwISC30to60E2r1.GMPAS-DIB-IAF-ISMF",
-            "PEM_P480_Ld5.T62_ECwISC30to60E2r1.GMPAS-DIB-IAF-ISMF",
+            "ERS_P480_Ld5.T62_ECwISC30to60E2r1.GMPAS-DIB-IAF-PISMF",
+            "PEM_P480_Ld5.T62_ECwISC30to60E2r1.GMPAS-DIB-IAF-PISMF",
+            "ERS_P480_Ld5.T62_ECwISC30to60E2r1.GMPAS-DIB-IAF-DISMF",
+            "PEM_P480_Ld5.T62_ECwISC30to60E2r1.GMPAS-DIB-IAF-DISMF",
```
xylar, line 234: @jonbob, scratch what I just wrote.  I did add these 2 extra tests that seem to be intended for stealth features.  Is that sufficient?

jonbob:
passes:

and successfully ran the new test:
* SMS_Ld1.T62_oQU240wLI.GMPAS-IAF-DISMF.chrysalis_intel

merged to next

xylar:
Thanks a lot, @jonbob!  I really appreciate your help getting the tests and namelist automation sorted out.

xylar:
Thanks so much @jonbob!

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
```diff
@@ -3401,7 +3403,28 @@ subroutine ocn_compute_land_ice_flux_input_fields(layerThickness, normalVelocity
...
+         call mpas_timer_stop("land_ice_diagnostic_fields")
+         return
+      end if
+
+      ! From here on, were in "standalone" or "coupled" mode
 
+      ! compute friction velocity at cell centers
+#ifdef MPAS_OPENACC
+      !$acc parallel loop present(landIceFrictionVelocity, kineticEnergyCell, minLevelCell) &
+#else
```
amametjanov, line 3423: PGI (or nvhpc) compiler flagged the extra continuation char `&` as a build error when compiling for a [GPU run](https://my.cdash.org/viewTest.php?buildid=2351627):
```
NVFORTRAN-S-0292-Illegal continuation line (/gpfs/wolf/cli115/proj-shared/azamat/e3sm_scratch/SMS_Ld1.T62_oEC60to30v3.CMPASO-NYF.ascent_pgigpu.JNextGpuacc20230614_015956/bld/cmake-bld/core_ocean/shared/mpas_ocn_diagnostics.f90: 3071)
NVFORTRAN-S-0034-Syntax error at or near ACCDO (/gpfs/wolf/cli115/proj-shared/azamat/e3sm_scratch/SMS_Ld1.T62_oEC60to30v3.CMPASO-NYF.ascent_pgigpu.JNextGpuacc20230614_015956/bld/cmake-bld/core_ocean/shared/mpas_ocn_diagnostics.f90: 3070)
```
If I remove the ampersand, the case builds and runs OK.
@jonbob can you add this in the next PR you merge? Thanks.

xylar: @amametjanov, I'm sorry about that.  Not sure how that crept in.

jonbob: @amametjanov -- I'll make a bug fix PR to handle this and make sure it gets in as soon as possible

jonbob:
new baseline  blessed on pm-cpu_gnu, but the run on mappy was now missing an ocean graph file. I've made that as well now, so hopefully the test will successfully complete on mappy tonight

PR #5703, opened 2023-05-18
MAM5 aerosol feature for v3

keziming:
This PR implements MAM5, which adds an additional stratospheric sulfate mode on top of the existing MAM4 aerosol scheme, for E3SMv3. MAM5 is able to better represent the stratospheric sulfate aerosol and can work with both Simple chemistry (SC) and Full chemistry (FC). MAM5-SC is linked to chemUCI.

The new branch is keziming/atm/MAM5_next

All the tests are documented here
https://acme-climate.atlassian.net/wiki/spaces/ATMOS/pages/3807051811/MAM5+Integration+Tests


tangq (approved):
Approve this PR so it won't delay the v3 integration, but note that the name list settings require to be specified in user_nl_eam, which should not needed. This implies that there is some mystery issue with build name list.

components/eam/bld/build-namelist
```diff
@@ -1119,6 +1119,10 @@ if ($aer_model eq 'mam' ) {
   if ($chem =~ /_mam4/) {$aero_modes = '4mode';}
   if ($chem =~ /_mam4_mom/) {$aero_modes = '4mode_mom';}
   if ($chem =~ /_mam4_resus_mom/) {$aero_modes = '4mode_mom';}
+# kzm ++
+  if ($chem =~ /_mam7_resus_mom_s/) {$aero_modes = '7mode_s';}
+  if ($chem =~ /_mam5_resus_mom/) {$aero_modes = '5mode';}
+# --
```
mingxuanwupnnl, line 1125: You may want to delete those comments with "kzm ++" and "kzm--". I think there a lot of places have those comments.

components/eam/src/chemistry/mozart/.mo_gas_phase_chemdr.F90.swp
mingxuanwupnnl, line 1: need to delete this .swp temporary file

rljacob (changes requested):
I'd like to see @mahf708 's suggested changes made.  The PR description needs to be an actual description and the title needs to be changed so its not just the branch name.

rljacob:
The circile ci does a build-only with gnu which means there's something the gnu compiler doesn't like.  You can compile on any other machine with gnu (chrysalis, pm-cpu) to duplicate and fix the problem.

components/eam/src/physics/cam/aer_rad_props.F90
```diff
@@ -77,7 +79,22 @@ subroutine aer_rad_props_init()
...
    call addfld ('extinct_sw_inp',(/ 'lev' /),    'A','1/km',&
         'Aerosol extinction directly read from prescribed input file', flag_xyfill=.true.)
-
+   ! add full bands radiation output
+   ! longwave: ext_sao_lw (pcols,pver,nlwbands)
+   ! ext_sao_sw, ssa_sao, af_sao (pcols,pver,nswbands)
+   if (is_output_interactive_volc) then
+   call add_hist_coord('nlwbands',    nlwbands,    'NLWBANDS')  !kzm
+   call add_hist_coord('nswbands',    nswbands,    'NSWBANDS')  !kzm 
+   call addfld ('ext_sao_lw',(/ 'lev', 'nlwbands' /),    'A','1/m',&
```
crterai, line 88: For it to work across different compilers, you will want to bad `'lev'` with spaces to make `'nlwbands'` and '`lev     '` the same length. We saw the same thing with the P3 PR.

components/eam/src/chemistry/modal_aero/aero_model.F90
```diff
@@ -23,6 +23,7 @@ module aero_model
 
   use modal_aero_data,only: cnst_name_cw
   use modal_aero_data,only: ntot_amode, modename_amode
+  use ref_pres,       only: top_lev => clim_modal_aero_top_lev  ! kzm+
```
whannah1, line 26: again, remove any and all "kzm" tags from the new code.

components/eam/src/chemistry/mozart/.mo_gas_phase_chemdr.F90.swp
whannah1, line 1: was this file comitted by mistake?

components/eam/src/chemistry/mozart/lin_strat_chem.F90
```diff
@@ -178,11 +178,12 @@ subroutine lin_strat_chem_inti(phys_state)
     no2_ndx  =   get_spc_ndx('NO2')
     hno3_ndx =   get_spc_ndx('HNO3')
 
-   !if(masterproc)then
+   if(masterproc)then
    ! write(iulog,*)'in subroutine lin_strat_chem_inti'
    ! write(iulog,*)'O3LNZ=',o3lnz_ndx,'N2OLNZ=',n2olnz_ndx,'NOYLNZ=',noylnz_ndx,'H2OLNZ=',h2olnz_ndx
    ! write(iulog,*)'O3=',o3_ndx,'CH4=',ch4_ndx,'N2O=',n2o_ndx,'NO=',no_ndx,'NO2=',no2_ndx,'HNO3=',hno3_ndx
-   !end if
+     write(iulog,*)'inside lin_strat_solve for ndx o3, o3lnz, n2o, noylnz, ch4', o3_ndx, o3lnz_ndx, n2o_ndx, noylnz_ndx, ch4_ndx
```
whannah1, line 185: Are we sure we want this written to the log for every run? Or is it just a debugging print statement that should be disabled by default?

components/eam/src/chemistry/utils/modal_aero_calcsize.F90
```diff
@@ -131,7 +132,12 @@ subroutine modal_aero_calcsize_reg()
 
   !register dgnum field
   call pbuf_add_field('DGNUM', 'global',  dtype_r8, (/pcols, pver, nmodes/), dgnum_idx)
-
+  !kzm ++
+   !if (modal_strat_sulfate_aod) then
+      !write(iulog,*)'kzm_wateruptake_reg_1'
+   !   call pbuf_add_field('SO4DRYVOL', 'global',  dtype_r8, (/pcols, pver, nmodes/), so4dryvol_idx)
+      ! write(iulog,*)'kzm_wateruptake_reg_2'
+   !end if
```
whannah1, line 140: seems like this could be removed too?

components/eam/src/chemistry/utils/modal_aero_wateruptake.F90
```diff
@@ -57,12 +56,18 @@ module modal_aero_wateruptake
...
 !$OMP THREADPRIVATE(maer, hygro, naer, dryvol, drymass, dryrad, wetrad, wetvol, wtrvol, rhcrystal, rhdeliques, specdens_1)
 
+
 !===============================================================================
 contains
 !===============================================================================
+!------------------
 
+
+!------------------
```
whannah1, line 70: These dividers seem out of place?

run.20221007.strat_AOD_fix13.restart.test.NGD_v3atm.v3atm.640layout.M5.sh
whannah1, line 1: We shouldn't be commiting run scripts within PRs, right?

components/eam/cime_config/testdefs/testmods_dirs/eam/v3atm_rtmoff_mam5/user_nl_eam_cpcp
```diff
@@ -0,0 +1,77 @@
...
+         'SOAG        -> /lcrc/group/e3sm/data/inputdata/atm/cam/chem/trop_mozart_aero/emis/DECK_ne30/cmip6_mam4_soag_elev_1850-2014_c180205.nc',
+         'bc_a4       -> /lcrc/group/e3sm/data/inputdata/atm/cam/chem/trop_mozart_aero/emis/DECK_ne30/cmip6_mam4_bc_a4_elev_1850-2014_c180205.nc',
+         'num_a1      -> /lcrc/group/e3sm/data/inputdata/atm/cam/chem/trop_mozart_aero/emis/DECK_ne30/cmip6_mam4_num_a1_elev_1850-2014_c180205.nc',
+         'num_a2      -> /lcrc/group/e3sm/data/inputdata/atm/cam/chem/trop_mozart_aero/emis/DECK_ne30/cmip6_mam4_num_a2_elev_1850-2014_c180205.nc',
+         'num_a4      -> /lcrc/group/e3sm/data/inputdata/atm/cam/chem/trop_mozart_aero/emis/DECK_ne30/cmip6_mam4_num_a4_elev_1850-2014_c180205.nc',
+         'pom_a4      -> /lcrc/group/e3sm/data/inputdata/atm/cam/chem/trop_mozart_aero/emis/DECK_ne30/cmip6_mam4_pom_a4_elev_1850-2014_c180205.nc',
+         'so4_a1      -> /lcrc/group/e3sm/data/inputdata/atm/cam/chem/trop_mozart_aero/emis/DECK_ne30/cmip6_mam4_so4_a1_elev_1850-2014_c180205.nc',
+         'so4_a2      -> /lcrc/group/e3sm/data/inputdata/atm/cam/chem/trop_mozart_aero/emis/DECK_ne30/cmip6_mam4_so4_a2_elev_1850-2014_c180205.nc'
```
whannah1, line 49: I just noticed that the chrysalis paths are hardcoded here, shouldn't these paths use DIN_LOC_ROOT instead?

keziming:
@whannah1 @mahf708 Thank you for your comments. I will do all the clean up jobs and put them all in one commit today.

keziming:
> Besides the excessive comments and build issue with gnu, here are a few high-level issues that should likely be addressed before merging:
> 
>     * This PR introduces verbose logging in some places. I remember looking at the logs of simulations based on the v3atm fork and there were a lot of repeated messages related to something mam5 activated. I can still see these logging lines in this PR.
> 
>     * This PR seems to have two features (MAM5 and MAM7). It is either I am confused because I don't know what's going on, or you're (optionally?) implementing MAM7 (7th sulfate mode) as well as MAM5 (5th sulfate mode). Either way, this should be carefully documented (note there is an inline documentation error --- mixing up 7th and 5th mode doc strings).
>       
>       * Note that the development and code review guidelines explicitly prohibit introducing two climate changing features (see last item below) at once.
>       * Will dropping MAM7 from this PR be feasible at this late stage of integration?
> 
>     * This PR is supposed to be climate changing, but it is unclear to me why it has to be this way (especially given most other features were stealth). Is it because of chemUCI linking? It seems to me that one has to choose MAM5 explicitly, so can't this PR merged while keeping MAM4 as the default configuration? Whatever the reason is, it should be documented here and internally.
>       
>       * First, it would be helpful to know why this decision was made for future reference. A similar decision was made for chemUCI and led to a cascade of issues.
>       * Second, it would also be helpful to know what to expect from the proposed changed climate. Where can we find more information?
>       * Third, the decision to have this be climate changing calls into question the strategy of later turning on the other features. Shouldn't we turn them on first and then integrate this then? It is not clear to me which sets of features are supposed to be most compatible with which. The main text in this PR (the PR description) makes it sound like this PR is compatible with everything, but why is it climate changing then?
> 
> 
> Edit: obviously, it may sound easy for me to pontificate on the state of this PR as an outsider not familiar with the long time spent on developing it. Please know that my intention here is to help by providing more eyes on issues that may have been missed in earlier stages. In my experience, the more, the merrier when it comes to code review...

@mahf708 Very good questions. Firstly, the MAM7S flags has been all removed, which is a feature developed parallel with MAM5. This MAM7S feature serves the same purpose as MAM5, providing solution for stratospheric sulfate aerosols, but is more expensive computationally. Please help me double check if there anything left. For the other diagnostic printings, I have cleared a lot of them and still working on it. 

Secondly, the model always can switch back to MAM4. For the V3, chemUCI-MAM5 is the default setting, but v2 like prescribed/simple chemistry plus MAM4 combination are still valid in this branch.  For people like use v2 like prescribed/simple chemistry plus MAM5, we have a solution but not included in this PR yet for code simplicity. But we do plan to add this option into master in near feature (this year, if time permits), because it is computational cheaper for people who are interested in stratospheric climate but not troposphere chemistry. 

Thirdly, we keeps both MAM4 and MAM5 option valid, but it doesn't mean the feature is stealth. We expect the simulated climate is very similar between master chemUCI-MAM4 and this PR's chemUCI-MAM4.

@mahf708  Could you help me test the PD climate difference under chemUCI-MAM4 configuration simulated by master and this PR. Thank you very much!

keziming:
> @keziming sounds good, take your time. I know you're juggling a lot of stuff at the moment in this stretch. If you think it is helpful and/or to speed things up for you, I can also go through the code line-by-line and point out potential areas for edits. Just let me know. (I obviously want the integration to proceed as fast as possible so that I can start my science runs with the v3 codebase and features soon 😸)

@mahf708  Could you help me test the PD climate difference under chemUCI-MAM4 configuration simulated by master and this PR. Thank you very much!

keziming:
> I'd like to see @mahf708 's suggested changes made. The PR description needs to be an actual description and the title needs to be changed so its not just the branch name.

I am working on improving the PR description.

keziming:
> @keziming @wlin7 in my testing, this PR seems to be BFB if running defaults (MAM4) with F2010. Tested [d1f5e36](https://github.com/E3SM-Project/E3SM/commit/d1f5e36fa058c325929bcd8bb93cc5f41b7205e5) against [b8aaf4f](https://github.com/E3SM-Project/E3SM/commit/b8aaf4faf7728ba224f1dd44d8a1a1e902b10513). Why is it listed as CC internally? I am sort of confused...

Hi Naser, thank you for testing that. The MAM4 and MAM5 switch are designed to be an BFB feature (we added flag everywhere to distinguish MAM4 and MAM5). But previous tests showed it was a non-BFB feature. Yesterday, I fixed bug about prescribed volc_cmip6 extinction issue (aer_rad_props.F90), which could cause NBFB. But I was not confident this bug fix can resolve this long-time NBFB issue. Another reason is that, because I cherry-picked the ozone-surface-loss fix which is NBFB and ahead of my rebased master, my baseline is not right for this test. Anyway, thank you for testing that.

@wlin7 Could you test it again? Even if it wouldn't pass all BFB tests, but the two simulated climates may be close enough.

keziming:
@mahf708 thank you for your helps. Have a good rest!

keziming:
@wlin7 @crterai Does this mean BFB 

Screenshot 2023-05-26 at 1 57 30 AM

components/eam/src/chemistry/mozart/mo_exp_sol.F90
```diff
@@ -109,6 +113,65 @@ subroutine exp_sol( base_sol, reaction_rates, het_rates, extfrc, delt, xhnm, nco
...
+       elseif ( trim(solsym(l)) == 'H2SO4') then 
+         do i = 1,ncol
+             do k = 1,pver
+                chem_prod(i,k,l) = prod(i,k,m)+ind_prd(i,k,m)
+                chem_loss(i,k,l) = - 1.0_r8*loss(i,k,m)
+                base_sol(i,k,l) = base_sol(i,k,l) + delt * (prod(i,k,m) + ind_prd(i,k,m) - loss(i,k,m))
+                !base_sol(i,k,l) = base_sol(i,k,l)*exp(-delt*loss(i,k,m)/base_sol(i,k,l)) + delt*(prod(i,k,m)+ind_prd(i,k,m))
+             end do
+         end do
+      elseif ( trim(solsym(l)) == 'DMS') then
```
tangq, line 135: Only SO2 and H2SO4 need to use a different equation. DMS can use the same equation as other gases. I suggest removing DMS from the changes here and other relevant places.

tangq:
@keziming , I pushed a commit ([2a040f2](https://github.com/E3SM-Project/E3SM/pull/5703/commits/2a040f2441b743a23dd22fb3950695b3f49c11ae)) to sync the changes in PR #5718 to avoid conflicts when merging this MAM5 PR. Be sure to review the changes and make sure they are good for MAM5.

keziming:
> @keziming , I pushed a commit ([2a040f2](https://github.com/E3SM-Project/E3SM/pull/5703/commits/2a040f2441b743a23dd22fb3950695b3f49c11ae)) to sync the changes in PR #5718 to avoid conflicts when merging this MAM5 PR. Be sure to review the changes and make sure they are good for MAM5.
> 
> Also, the chemistry prod/loss diagnostics for the explicit solver are based on `diags_reaction_rates`, which has reaction rates of zeros in the stratosphere (see mo_gas_phase_chemdr.F90). This means when MAM5 is on these diagnostics are broken. Not urgent to fix this, but it is worth noting here.

@tangq Thank you for your commit. I checked the code, and didn't find anything suspicious. I got your concern about the impact of MAM5 on chemistry diagnostics because of the "level_loop" limit in the mo_gas_phase_chemdr.F90. As this doesn't affect the simulated climate, we can find a way to solve it later. My question is that because the chemistry diagnostics only apply to troposphere, does it account the stratospheric produced H2SO4 in MAM5 as "other" or "transport" sources?

rljacob:
The goal should not be "0 merge conflicts".  The integrator can resolve them during integration.   The integrator can do a test merge to see how bad they are and only push more commits to fix it if asked.

keziming:
@rljacob @mahf708 @tangq @wlin7 I can resolve all the conflicts. My question is that after I resolved them what I should do to save the changes. I see a 'merge commit' button, and not sure if I should click it.

keziming:
> > @keziming , I pushed a commit ([2a040f2](https://github.com/E3SM-Project/E3SM/pull/5703/commits/2a040f2441b743a23dd22fb3950695b3f49c11ae)) to sync the changes in PR #5718 to avoid conflicts when merging this MAM5 PR. Be sure to review the changes and make sure they are good for MAM5.
> > Also, the chemistry prod/loss diagnostics for the explicit solver are based on `diags_reaction_rates`, which has reaction rates of zeros in the stratosphere (see mo_gas_phase_chemdr.F90). This means when MAM5 is on these diagnostics are broken. Not urgent to fix this, but it is worth noting here.
> 
> @tangq Thank you for your commit. I got your concern about the impact of MAM5 on chemistry diagnostics because of the "level_loop" limit in the mo_gas_phase_chemdr.F90. As this doesn't affect the simulated climate, we can find a way to solve it later. My question is that because the chemistry diagnostics only apply to troposphere, does it account the stratospheric produced H2SO4 in MAM5 as "other" or "transport" sources?

@tangq I pushed a new commit #ffbaa9c8de, #2305c809e5 and #81e029976b to adding DMS into stratospheric reactions. The DMS would not produce dead zeros but is still needed to be solved in the stratosphere. Here, I use V3-like equation to solve DMS but apply to whole atmosphere. Please help me check if it is right.

keziming:
> @keziming, it seems you're on the right track. If you're seeing something like this https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/addressing-merge-conflicts/resolving-a-merge-conflict-on-github, it means you're going to merge the commit from master to your branch, hence, the "commit merge" button, but it is better to wait for @wlin7 who takes care of these merges.

@mahf708 Thanks.
@rljacob The conflicts are resolved.

keziming:
@mahf708 Thank you for bring this up, and it is important to have clarification about this issue. In short answer, there is no conflict between MAM5 and new VBS-SOA chemistry. 

As chemUCI applies to the troposphere and sets chemistry solvers, implicit and explicit, off above tropopause. For stratospheric aerosol purpose, MAM5 uses explicit solver to solve sulfur chemistry (SO2, H2SO4, DMS) and imports prescribed oxidants instead of chemUCI oxidants in the stratosphere. The reason to use prescribed oxidants in the stratosphere is due to chemUCI doesn't reach to stratosphere. These changes are in mo_exp_sol.F90 and mo_gas_phase_chemdr.F90. 

On the other hand, VBS-SOA uses prescribed oxidants to decouple its chemistry with chemUCI. All these reactions are in the troposphere and in implicit solver. The prescribed oxidants are imported from the same file MAM5 used. Therefore, I don't see a directly conflict but I will set a run to test it.

rljacob:
Replaced by #5749 and a future PR for v3atm/eam/master_MAM5_wetaero_chemdyg

PR #5712, opened 2023-05-22
Add low nc limit: impose in-cloud Nc no less than a threshold

yunpengshan2014:
impose in-cloud Nc for large-scale stratiform no less than a threshold (default: 10 per CC)


rljacob:
If this fixes a bug please create a github issue describing it.

wlin7:
> If this fixes a bug please create a github issue describing it.

FYI, @rljacob . This is an add-on feature for v3atm. It is in v3atm repo.

crterai:
I believe this PR is actually an older version of https://github.com/E3SM-Project/E3SM/pull/5715

PR #5717, opened 2023-05-24
Add initial superBFB tests for EAM and WCYCL

amametjanov:
The purpose of this suite is to make sure new code maintains
bit-for-bit properties we demand of the model.  These properties include:
- exact restart:  the model should be able to restart during a run with bit-for-bit 
identical results compared to a model which ran for the same total time without restarting.
- MPI-task count invariance:  The model should give the same answers 
between two runs that only differed by the number of MPI tasks used.
- OpenMP thread count invariance:  The model should give the same answers 
between two runs that only differ by number of OpenMP threads used.

There is a suite for each major component model (`atm`, `ocn`, `ice`, `lnd`, `rof`, and 
fully-coupled `wcycl`). They cover the standard-resolution grids in both 
release/optimized and debug configurations.

[BFB]

### DIFF
--- a/cime_config/testmods_dirs/pemod/omp1/shell_commands
+++ b/cime_config/testmods_dirs/pemod/omp1/shell_commands
@@ -10,4 +10,5 @@ ppn=`./xmlquery --value MAX_MPITASKS_PER_NODE`
 ./xmlchange NTASKS=$(($ppn * 8))
 ./xmlchange NTHRDS=1
 ./xmlchange ROOTPE=0
+./xmlchange PSTRID=1
 
--- a/cime_config/testmods_dirs/pemod/omp2/shell_commands
+++ b/cime_config/testmods_dirs/pemod/omp2/shell_commands
@@ -11,4 +11,5 @@ ppn=`./xmlquery --value MAX_MPITASKS_PER_NODE`
 ./xmlchange NTASKS=$(($ppn * 4))
 ./xmlchange NTHRDS=2
 ./xmlchange ROOTPE=0
+./xmlchange PSTRID=1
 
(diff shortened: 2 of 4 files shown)


rljacob:
I wasn't thinking about a super-bfb test for the fully coupled model but I guess we do need it.   But if someone is just developing code in a component, should they run this as part of the PR testing?  Or would this be something the IG runs?

amametjanov:
I am hoping that fully coupled tests are part of the PR testing: to catch unintended interaction among components.
Maintaining a nightly baseline of PASS/FAIL status would be useful in catching unintended bugs in merges. No baseline comparisons. Just a coverage of various modes, especially threading.

rljacob:
@wlin7 and @golaz please review.

sarats:
It would be good to get this integrated to get proper coverage in the Super BFB suite.

rljacob (approved):

amametjanov:
PET test type checks for BFB between 2 and 1 threads.

rljacob:
notes:  need to shorten run length of a test.

wlin7:
@amametjanov , can you please give an explanation in the PR message what are the purposes of these superBFB tests (or what potential scenarios to guard against)?

amametjanov:
This is for the code review process at https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/3438608385/E3SM+Code+Review+and+New+Feature+Process : towards the bottom
```
Super-BFB suite
The purpose of this suite is to make sure your new code maintains
bit-for-bit properties we demand of the model.  These properties include:
- exact restart:  the model should be able to restart during a run with bit-for-bit 
identical results compared to a model which ran for the same total time without restarting.
- MPI-task count invariance:  The model should give the same answers 
between two runs that only differed by the number of MPI tasks used.
- OpenMP thread count invariance:  The model should give the same answers 
between two runs that only differ by number of OpenMP threads used.

There is a suite for each major component model (`atm`, `ocn`, `ice`, `lnd`, `rof`, and 
fully-coupled `wcycl`). They cover the standand-resolution grids in both 
optimization and debug configurations.
```

rljacob:
@amametjanov I think he means to edit the PR description to add more detail.

PR #5720, opened 2023-05-25
[Standalone MPAS-Ocean] Warn if `debugTracers` enabled with single layer in RK4

xylar:
Recently, it had been made a critical error when `debugTracers` are used with RK4 in single-layer mode.  This was because this configuration is incompatible with many single-layer test cases. However, we would also shown that the cosine-bell advection test case does work with `debugTracers `and single layer.  To support this case, we switch from a critical error to a warning in this merge.

### DIFF
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_rk4.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_rk4.F
@@ -1741,7 +1741,7 @@ module ocn_time_integration_rk4
          if (config_use_debugTracers .and. nVertLevels == 1) then
             call mpas_log_write('Debug tracers cannot be used in a ' &
                // 'single layer case. Consider setting ' &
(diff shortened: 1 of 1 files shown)


components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_rk4.F
```diff
@@ -1741,7 +1741,7 @@ subroutine ocn_time_integration_rk4_init(domain)!{{{
          if (config_use_debugTracers .and. nVertLevels == 1) then
             call mpas_log_write('Debug tracers cannot be used in a ' &
```
cbegeman, line 1742:
```suggestion
            call mpas_log_write('Debug tracers may cause failures in a ' &
```

jonbob:
passes:

merged to next

xylar:
Thanks very much for testing, @jonbob!

cbegeman:
Thank you, @jonbob!

xylar:
Thank you, @jonbob!

PR #5721, opened 2023-05-25
Homme: remove kokkos submodule, use Kokkos from EKAT submodule

### LINKED ISSUE #5417: Kokkos inventory and possibility of convergence
There seem to be a couple of versions of Kokkos floating around in the E3SM and SCREAM repositories.

Starting this discussion to understand our Kokkos dependencies and see if we can converge on requirements to avoid confusion/redundancy. 

One of them is through SCREAM/EKAT 
https://github.com/E3SM-Project/EKAT/tree/master/extern
> [kokkos @ e8ad274](https://github.com/E3SM-Project/kokkos/tree/e8ad274447f11582cef98f0aa409d8ee06d2c7fa)
[Update extern/kokkos to e3sm-kokkos-3.7.00-tag](https://github.com/E3SM-Project/EKAT/commit/85062686615d99547ddd494ef8b97f72b5c63a54)
> 3 months ago

Another is under https://github.com/E3SM-Project/E3SM/tree/master/externals
> [kokkos @ 290a073](https://github.com/E3SM-Project/kokkos/tree/290a073f38c024ba37113716d7ed4ac921f3db85)
[Update kokkos module with Fujitsu compiler mods](https://github.com/E3SM-Project/E3SM/commit/ea12e8d167bb342a9f2fc328a9ced4852042b5de)[kokkos @ 290a073](https://github.com/E3SM-Project/kokkos/tree/290a073f38c024ba37113716d7ed4ac921f3db85)
[Update kokkos module with Fujitsu compiler mods](https://github.com/E3SM-Project/E3SM/commit/ea12e8d167bb342a9f2fc328a9ced4852042b5de)
> 9 months ago

During a recent meeting, there was a comment that the version under E3SM/externals is used for SL transport CPU builds? 

cc @mt5555 @ambrad

bartgol:
This PR removes the submodule in `externals/kokkos`, and uses kokkos from `externals/ekat/extern/kokkos` instead.

Both homme_integration and e3sm_atm_developer pass on chrysalis.

Fixes #5417 

[BFB]
 
@sarats


jgfouca:
@oksanaguba , if you approve, I will start integrating this.

components/eam/bld/namelist_files/namelist_defaults_eam.xml
```diff
@@ -918,7 +918,7 @@
 <micro_nccons         microphys="mg2">  100.D6      </micro_nccons>
 <micro_nicons         microphys="mg2">  0.1D6       </micro_nicons>
 <micro_mincdnc        microphys="mg2">  -999.       </micro_mincdnc>
-
+<p3_mincdnc           microphys="p3">  -999.        </p3_mincdnc>
```
oksanaguba, line 921: why is this part of this PR? maybe, add these changes to the description?

oksanaguba:
I am only confused why there are p3 changes in here. maybe it is just an issue with branching/merging/rebasing .

oksanaguba (approved):

jgfouca:
@oksanaguba , i was confused by that too.

oksanaguba:
@jgfouca  approved (assuming there is a simple explanation wrt p3).

components/eam/bld/namelist_files/namelist_defaults_eam.xml
bartgol: I don't recall changing this file...

bartgol: The log shows for the first two commits

```
Yunpeng Shan authored and [bartgol](https://github.com/E3SM-Project/E3SM/commits?author=bartgol) committed yesterday
```
I don't know how this ended in the PR. I'll remove those commit and force push a clean version.

bartgol:
@oksanaguba rebasing on master took care of those two stray commits. Not sure what happened there.

sarats:
Just curious - does this impact/allow use of a third-party Kokkos installation?
https://github.com/E3SM-Project/E3SM/pull/5391

bartgol:
Mmm, good question. I think the answer is yes. Notice that buildlib.kokkos is only needed by EAM, while EAMxx and Homme standalone build kokkos internally.

I think the bigger question is how to allow EAMxx to use an existing installation of Kokkos. We do not have that capability yet, but it shouldn't be hard to add. In EAMxx we already wrap existing libraries (mostly, the ones build in the sharedlib phase of case.build) in CMake targets, to avoid having to specify include/link dirs in every target that links them. I think we could do the same in EKAT when building Kokkos, if a KOKKOS_PATH or something var is provided. Of course, we'd have to check that the provided installation is compatible with EKAT (e.g., minimum/maximum version), but that's easy.

tl;dr. I don't think this is a problem for EAM or MALI, which are the only customers of sharedlib Kokkos.

ambrad:
I see this PR is causing minor diffs in the homme integration nightly. I looked at the output, e.g., https://my.cdash.org/test/81498845, and the diffs are fine: at machine precision level in two SL transport tests for the F90 dycore. It looks like there are no diffs in the E3SM integration nightly on, e.g., Chrysalis, but it's hard to be sure because the dashboard's state has been quite complex in the past week.

bartgol:
I should double check the flags that are set in master vs next when it comes to homme execs. The most likely explanation is some difference in compile/link flags. IIRC, ekat and e3sm were currently using the same kokkos version, so we should be building the same source code.

oksanaguba:
@luca by any chance did you try to build homme with external kokkos from this PR? this is something i use often. if not, maybe i should try.

bartgol:
> @bartgol by any chance did you try to build homme with external kokkos from this PR? this is something i use often. if not, maybe i should try.

This PR does not allow to use an external _pre-built_ kokkos. It does allow to specify a different source folder (other than the one from ekat), though I don't think that's what you meant...

bartgol:
Compiler and link flags on the composef90 lib as well as the theta-l-nlev30 exec seem to be equivalent. The only diff I noticed is the location of -mkl in the link line: in master, it's between pio libs and kokkos libs, while in the branch it comes _after_ kokkos libs. However, I do not think there's anything in kokkos that makes use of mkl, so I don't really see how this could make a difference.

ambrad:
It appears to me that in E3SM master, external/kokkos and external/ekat/extern/kokkos are not quite the same versions. That's fine, especially if the only diff is this one homme_integration test.

PR #5722, opened 2023-05-25
Update mpas-framework xml_stream_parser.c

### LINKED ISSUE #5709: With maint-1.0, build error in MPAS on pm-cpu/intel
After PR https://github.com/E3SM-Project/E3SM/pull/5696 to maint-1.0, we can work with pm-cpu, however, there are still some issues requiring changes in submodules.

To use Intel, we need a work-around for this build error:

```
xml_stream_parser.c:1732:6: error: call to undeclared function 'stream_mgr_add_field_c'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
                                        stream_mgr_add_field_c(manager, streamID, (const char *)fieldname, packages_local, &err);
```

I am able to simply copy the current `xml_stream_parser.c` from master and it works, but there may be a better fix

jonbob:
Update the mpas-framework xml_stream_parser.c file to fix issue on pm-cpu

Fixes #5709 

[BFB]


ndkeen (approved):
Tested with e3sm_prod using intel on pm-cpu.

Thanks for getting this change thru.

jonbob:
Thanks for testing so quickly, @ndkeen. And thanks @xylar for the suggested fix. Merged to maint-1.0

PR #5723, opened 2023-05-25
Aqueous chemistry bug fix

### LINKED ISSUE #5713: Aqueous chemistry bug
In the aqueous chemistry, the total condensate mass mixing ratio including both liquid and ice was used in the calculation of production of sulfate from the aqueous chemistry processes. However, the liquid water mass mixing ratio should be used based on the formula used (Jiwen also communicated with NCAR people to confirm this). Thus, the bug fix was about using the liquid water mass mixing ratio instead of the total condensate mass mixing ratio in the aqueous chemistry. 
[mo_gas_phase_chemdr.F90.gz](https://github.com/E3SM-Project/E3SM/files/11535996/mo_gas_phase_chemdr.F90.gz)

Jiwengithub:
This PR is for fixing the the aqueous chemistry bug where total condensate (liquid +ice) mass mixing ratio was used for the SO2 aqueous chemistry process. The problem is that only liquid water mass mixing ratio should be used here. The bug was found because P3 produces more ice mass which leads to a much larger sulfate production from aqueous chemistry compared to MG2. By working together with Dick Easter, we identified this problem and we also got confirmed from NCAR.  

Fixes #5713 
[CC]

Two files need to be changed for this fix: chemistry.F90 and mo_gas_phase_chemdr.F90.  (really just the latter)

https://github.com/E3SM-Project/E3SM/pull/new/JiwenFan/E3SM/aqueouschemsitry_bugfix.

### DIFF
--- a/components/eam/src/chemistry/mozart/mo_gas_phase_chemdr.F90
+++ b/components/eam/src/chemistry/mozart/mo_gas_phase_chemdr.F90
@@ -1265,7 +1265,7 @@ contains
     call t_startf('aero_model_gasaerexch')
     call aero_model_gasaerexch( imozart-1, ncol, lchnk, delt, latndx, lonndx, reaction_rates, &
                                 tfld, pmid, pdel, mbar, relhum, &
-                                zm,  qh2o, cwat, cldfr, ncldwtr, &
+                                zm,  qh2o, cwat_liq, cldfr, ncldwtr, &  ! replace cwat with cwat_liq for SO2 aqueous chemsitry bug fix 
                                 invariants(:,:,indexm), invariants, del_h2so4_gasprod,  &
                                 vmr0, vmr, pbuf )
     call t_stopf('aero_model_gasaerexch')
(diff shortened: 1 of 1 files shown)


Jiwengithub:
The fix is "climate changing".

crterai:
This PR addresses this git issue. https://github.com/E3SM-Project/E3SM/issues/5713

Jiwengithub:
@crterai Done.

crterai (approved):
Thank you for making the fix, @Jiwengithub. The changes look good to me.

wlin7:
Hi @Jiwengithub , the commit message says chemistry.F90 also needs to be changed, but seems like not so. Can you please confirm? I think  CC is only for cases using chemUCI, otherwise the PR is BFB for v2-like compsets. Can you please also confirm?

crterai:
The change is climate changing whether run with standard v2 chemistry or with chemUCI. Its impact is to reduce AOD, while also increasing total TOA SW flux by 0.6 Wm-2 (from 5 year F2010 tests). FLNT and PRECT changes are minimal.

wlin7:
Merged to next.

wlin7:
Merged to master.

wlin7:
For record,  [MVK test](https://web.lcrc.anl.gov/public/e3sm/e3smtest/evv/MVK_PS.ne4_oQU240.F2010.chrysalis_intel.C.JNextAtm_nbfb20230530_003620/index.html), with this PR in next, also deemed the fix as climate changing

Issue #5726, opened 2023-05-25
Audit usage of GNU Fortran extensions and cleanup if possible


sarats:
https://gcc.gnu.org/onlinedocs/gfortran/Extensions-implemented-in-GNU-Fortran.html

As mentioned in https://github.com/E3SM-Project/E3SM/issues/5724, gfortran by default sets `-std=gnu` which enables these extensions. We should set explicit standard for GNU to ensure better portability across compilers. This Github issue can serve as a catch-all to understand what changes need to be done.

Additionally, there are several instances of the following in our codebase as our machine files have this extension enabled (`-fallow-invalid-boz`)
https://github.com/search?q=repo%3AE3SM-Project%2FE3SM+invalid-boz&type=code
https://gcc.gnu.org/onlinedocs/gfortran/BOZ-literal-constants.html

https://github.com/E3SM-Project/E3SM/blob/master/cime_config/machines/Depends.gnu.cmake#L16 (and gnugpu)
```
list(APPEND ALLOW_INVALID_BOZ_LIST
  eam/src/control/cam_history.F90
  elm/src/biogeochem/MEGANFactorsMod.F90
)
...
    e3sm_add_flags("${ITEM}" "-fallow-invalid-boz") # avoids build error for integer, parameter :: gen_hash_key_offset = z'000053db'
```

sarats:
For instance,
https://github.com/E3SM-Project/E3SM/blob/master/components/eam/src/control/cam_history.F90#LL263C1-L263C58
```
      |     integer, parameter :: gen_hash_key_offset = z'000053db'
      |                                               1
Error: BOZ literal constant at (1) is neither a data-stmt-constant nor an actual argument to INT, REAL, DBLE, or CMPLX intrinsic function [see '-fno-allow-invalid-boz']
```
can be fixed by using the `int` intrinsic explicitly.
`integer, parameter :: gen_hash_key_offset = int(z'000053db')`

sarats:
I will explicitly set a different standard (starting with f2008) to avoid gnu extensions in a PR. We will see what happens during testing.

```
-std=std
           Specify the standard to which the program is expected to conform, which may be one of f95, f2003, f2008, gnu, or legacy.  The default value for std is gnu, which specifies a superset of the Fortran 95 standard that includes all of the extensions supported by GNU
```

PR #5727, opened 2023-05-25
 GNU Fortran: Wrap boz constant inside intrinsic to fix compilation issue

sarats:
This fixes having to set `-fallow-invalid-boz` flag for gfortran to 
build: wrap a couple of constants with a type-casting intrinsic. 
Also, clean up and remove flag duplicates in mach-specific cmakes.

[BFB]

===
 While looking into https://github.com/E3SM-Project/E3SM/issues/5726, encountered a couple of constants that required `-fallow-invalid-boz` flag for gfortran to build. This fixes that issue without the need for the GNU extension.

cc @rljacob


ndkeen:
We already have work-arounds for these errors inside Depends.
https://github.com/E3SM-Project/E3SM/pull/4822

However, I never liked that we needed them at all and if casting as `int` works, would be good to do that instead.

sarats:
This is to specifically remove the need for those workarounds. 

For cam_history, I tested the output. The ELM source is a copy/paste of the constant/hash function from cam.
`gen_hash_key_offset = z'000053db' with flag = gen_hash_key_offset = int(z'000053db') = 21467`

sarats:
@rljacob I think this is good to go for testing on next.

amametjanov:
Reproduced the initial build error with `SMS.ne4_oQU240.F2010.pm-cpu_gnu`:
```
/global/u2/a/azamat/saul/E3SM/components/elm/src/biogeochem/MEGANFactorsMod.F90:271:47:

  271 |     integer, parameter :: gen_hash_key_offset = z'000053db'
      |                                               1
Error: BOZ literal constant at (1) is neither a data-stmt-constant nor an actual argument to INT, REAL, DBLE, or CMPLX intrinsic function [see '-fno-allow-invalid-boz']
```
The case builds and runs OK with the updates.
There are additional occurrences of this flag in other gnu cmake files:
```
> grep -ir "fallow-invalid-boz" cime_config/
cime_config/machines/cmake_macros/gnu_anlgce-ub18.cmake:string(APPEND FFLAGS " -fallow-argument-mismatch -fallow-invalid-boz")
cime_config/machines/cmake_macros/gnu_anlgce.cmake:string(APPEND FFLAGS " -fallow-argument-mismatch -fallow-invalid-boz")
cime_config/machines/cmake_macros/gnu_spock.cmake:string(APPEND FFLAGS " -fallow-argument-mismatch -Wno-implicit-interface -fallow-invalid-boz ")
cime_config/machines/cmake_macros/gnu_docker-scream.cmake:  string(APPEND FFLAGS " -fallow-argument-mismatch  -fallow-invalid-boz ")
cime_config/machines/cmake_macros/gnu_eddi.cmake:  string(APPEND FFLAGS " -fno-unsafe-math-optimizations -fallow-argument-mismatch -fallow-invalid-boz")
cime_config/machines/cmake_macros/gnu_eddi.cmake:  string(APPEND FFLAGS " -g -fbacktrace -fbounds-check -ffpe-trap=invalid,zero,overflow -fallow-argument-mismatch -fallow-invalid-boz -Wall")
```
@sarats can I push a cleanup of these remaining flags to your branch?

sarats:
@amametjanov Sure, go ahead.

amametjanov (approved):
Thanks. Done.

PR #5490, opened 2023-03-02
Adding support for ADIOS in MPAS

jayeshkrishna:
Adding support for the ADIOS I/O format

[BFB]

PR #5732, opened 2023-05-26
[MPAS standalone] Add `HAVE_ADIOS=false` CPP flag for PIO without ADIOS

xylar:
This allows us to build standalone MPAS components with ParallelIO (which doesn't support ADIOS) instead of SCORPIO. 

### Background

We need to use ParallelIO if we want to build ESMF and MPAS standalone components with the same ParallelIO.  This is useful when we are building both ParallelIO and ESMF with spack (and therefore can't have both ParallelIO for ESMF and SCORPIO for standalone MPAS in the same environment).  Whereas older ESMF could be built with its own internal ParallelIO as a static library, that is no longer possible with ESMF >=8.3.0.  The older ESMF doesn't build with the new Intel compilers on Perlmutter, so this PR seems like our best way forward.

[BFB]


xylar:
## Standalone Testing

I was able to build a spack environment with ParallelIO 2.5.10 and ESMF 8.4.0, and then build standalone MPAS-Ocean using this branch.  Before these changes, I got:
```
mpas_io.F(280): error #6404: This name does not have a type, and must have an explicit type.   [PIO_IOTYPE_ADIOS]
            pio_iotype = PIO_iotype_adios
```
with a similar Spack library stask.

I'm currently waiting in the queue to see how the `pr` suite on Chrysalis fairs.

xylar:
@philipwjones, @jayeshkrishna, @mark-petersen, @darincomeau, @matthewhoffman, I've cast a pretty wide net in asking for reviews.  I'm hoping a review by inspection is good enough for most of you.  I'd like to have someone test MALI (@matthewhoffman) and MPAS-Seaice (@darincomeau?) to make sure they're not negatively affected by this.

@jonbob, what E3SM tests would you recommend I run?

xylar:
The background for this work is that currently can't be built with Intel on Perlmutter because:
* We can't build an older ESMF with the newer Intel compilers on PM
* Newer ESMF requires an external ParallelIO package (not the internal one used in older versions)
* We can't use SCORPIO because it doesn't support the full ParallelIO API expected by ESMF
* We can't build MPAS-Ocean with PrallelIO at the moment because ADIOS support was hard-coded in #5490

rljacob:
Those hard-coded mods in #5490 look like they would be easy to put cpp ifdef around.

@jayeshkrishna doesn't SCORPIO have a build option to imitate the PIO API?

xylar:
@rljacob 
> Those hard-coded mods in https://github.com/E3SM-Project/E3SM/pull/5490 look like they would be easy to put cpp ifdef around.Those hard-coded mods in https://github.com/E3SM-Project/E3SM/pull/5490 look like they would be easy to put cpp ifdef around.

That’s what I’m attempting to do here. If my way isn’t to your liking, feel free to suggest changes.

philipwjones (approved):
I'm fine with the changes, but what are we using ESMF for? If it's just for the offline regridder, then we can configure Spack to build that with PIO while still building MPAS and E3SM with Scorpio.

components/mpas-framework/Makefile
```diff
@@ -515,6 +515,13 @@ ifeq "$(USE_PIO2)" "true"
 	override CPPFLAGS += -DUSE_PIO2
 endif
 
+#
+# If user has indicated ADIOS isn't supported, define NO_ADIOS pre-processor macro
+#
+ifeq "$(ADIOS)" "false"
+	override CPPFLAGS += -DNO_ADIOS
+endif
+
```
rljacob, line 524: I think its better to have cpp flags follow "defined = on = true = yes" logic.  So if ADIOS is requested, set "HAVE_ADIOS" to "True".

xylar: @rljacob, yes I think that makes sense.  I'll change it.

xylar:
> I'm fine with the changes, but what are we using ESMF for? If it's just for the offline regridder, then we can configure Spack to build that with PIO while still building MPAS and E3SM with Scorpio.

@philipwjones, thanks for taking a look.  This is for compass, which has SCORPIO and ESMF in the same spack environment.  I'm not keen to have a separate spack environment for linking to MPAS components and for running ESMF from.  And I don't think we can have PIO and SCORPIO in the same spack environment because they collide with one another (i.e. they have the same library names).  If you know how to get around that, I'm keen to hear about it.

components/mpas-framework/Makefile
rljacob: Looks like "USE_ADIOS" would fit the pattern with other cpp constants.

xylar: @rljacob, it's becoming clear to me why I did it this way.  If the CPP flag is `NO_ADIOS`, then the E3SM default of not defining this flag means ADIOS is supported.  If you prefer the `HAVE_ADIOS` flag, we presumably need to define that always for E3SM.  I was trying to modify only the standalone build infrastructure but I can figure out how to add this to the E3SM build, too.

rljacob: Oh you're right.  Ok to leave it as-is then.

xylar: I did implement `HAVE_ADIOS`.  It's not super hard to have this flag for E3SM runs.  Shall I still go back to what I had?  I'm fine either way.

components/mpas-framework/src/CMakeLists.txt
```diff
@@ -41,7 +41,7 @@ set(FILE_OFFSET "-DOFFSET64BIT")
 if (CPPDEFS)
   separate_arguments(CPPDEFS UNIX_COMMAND "${CPPDEFS}")
 endif()
-list(APPEND CPPDEFS "-DMPAS_NO_LOG_REDIRECT" "-DUSE_PIO2" "-DMPAS_NO_ESMF_INIT" "-DMPAS_ESM_SHR_CONST" "-DMPAS_PERF_MOD_TIMERS" "${MODEL_FORMULATION}" "${FILE_OFFSET}" "${ZOLTAN_DEFINE}" "-D_MPI" "-DMPAS_NAMELIST_SUFFIX=${NAMELIST_SUFFIX}" "-DMPAS_EXE_NAME=${EXE_NAME}")
+list(APPEND CPPDEFS "-DMPAS_NO_LOG_REDIRECT" "-DUSE_PIO2" "-DHAVE_ADIOS" "-DMPAS_NO_ESMF_INIT" "-DMPAS_ESM_SHR_CONST" "-DMPAS_PERF_MOD_TIMERS" "${MODEL_FORMULATION}" "${FILE_OFFSET}" "${ZOLTAN_DEFINE}" "-D_MPI" "-DMPAS_NAMELIST_SUFFIX=${NAMELIST_SUFFIX}" "-DMPAS_EXE_NAME=${EXE_NAME}")
```
xylar, line 44: @rljacob, this would be the change to the E3SM build.

components/mpas-framework/Makefile
rljacob: No keep what you just did.

components/mpas-framework/Makefile
```diff
@@ -1159,11 +1166,11 @@ errmsg:
 	@echo "    SHAREDLIB=true - generate position-independent code suitable for use in a shared library. Default is false."
 	@echo "    USE_LAPACK=true - builds and links with LAPACK / BLAS libraries.  Default is to not use LAPACK."
 	@echo "    USE_SHTNS=true - builds and links with SHTNS/FFTW.  Default is false."
+	@echo "    HAVE_ADIOS=false   - PIO2 library does not include ADIOS support.  Default is true."
```
rljacob, line 1169: You would need to change this language.

components/mpas-framework/Makefile
rljacob: Should line 521 be "true" instead of "false"?

components/mpas-framework/src/CMakeLists.txt
rljacob: looks good.

components/mpas-framework/Makefile
xylar: No, I think what I have is right.  I think we want the default even for standalone MPAS to be that `HAVE_ADIOS=true`, so you have to explicitly set `HAVE_ADIOS=false` to indicate you're using PIO and don't have ADIOS support.

components/mpas-framework/Makefile
xylar: I believe this is really what we want.  The default is that you have ADIOS and you have to specify `HAVE_ADIOS=false` when building to indicate that your PIO version doesn' support it.

rljacob: Yes you're right.

components/mpas-framework/Makefile
xylar: How would we make the default `HAVE_ADIOS=true` if `HAVE_ADIOS` is undefined?

rljacob: The HAVE_ADIOS in `ifeq "$(HAVE_ADIOS)" "true"` is a makefile variable that should be given a default value of false.  If its true, the cpp variable HAVE_ADIOS will be defined.

philipwjones:
@xylar You can definitely have them both installed at the same time and you can you can also build other packages based on the specific dependencies, either by version or even by specifying the hashes. The reason why I asked whether you are only using the E3SM command-line executables is that it's easier to deal with these conflicts in that case. If they're both being used as part of the same executable, then it's probably not possible.

I think this PR is the right way to go in the short term - just thinking about how to deal with these conflicts.

components/mpas-framework/Makefile
xylar: @rljacob, I don't think that's the behavior we want.  We want `HAVE_ADIOS=true` , not `=false`, to be the default.

xylar:
@philipwjones

> @xylar You can definitely have them both installed at the same time and you can you can also build other packages based on the specific dependencies, either by version or even by specifying the hashes. The reason why I asked whether you are only using the E3SM command-line executables is that it's easier to deal with these conflicts in that case. If they're both being used as part of the same executable, then it's probably not possible.

I'd be very keen to learn more about this.  We just need `ESMF_RegridWeightGen` so it sounds like the situation you're referring to where it should be possible to have both ParallelIO and SCORPIO installed at the same time.  I don't really see how that will work, given that I think both are shared libraries with the same name and location in spack, but I'm eager to learn!

xylar:
I have tested the following with ParallelIO (and Intel with OpenMPI) on Chrysalis:
* this works: `make HAVE_ADIOS=false ifort`
* this fails as expected: `make HAVE_ADIOS=true ifort`
* this also fails as expected: `make ifort`

Some E3SM testing is also needed.

xylar:
@matthewhoffman, @darincomeau, @mark-petersen, could you make sure this does no harm for standalone MALI, MPAS-Seaice and MPAS-Ocean, respectively?

@jayeshkrishna can you make sure you're good with these changes, since you added the ADIOS support to the MPAS framework originally?

xylar:
Also a question for everyone: Are we okay with `HAVE_ADIOS` or would we prefer `USE_ADIOS` to match other standalone MPAS flags?  It has a different meaning than typical `USE_*` flags because it is not that we are linking a specific ADIOS library so I think it might be better if it has a slightly different name.  But I don't have strong feelings.

philipwjones:
HAVE_ADIOS is probably more correct since it only signifies that it's available, not necessarily that we're actually using it. But I don't have a strong preference either way either.

matthewhoffman:
@xylar , thanks for putting together this solution for this new issue.  For testing, do we need a new compass and spack environment to test this?  I won't have time to test this today, but the changes look good based on visual inspection.

Regarding the option name, I'm happy with the non-standard `HAVE_ADIOS` for the unusual usage of this option.

xylar:
> For testing, do we need a new compass and spack environment to test this? I won't have time to test this today, but the changes look good based on visual inspection.

I had in mind of you to do some do-no-harm testing with existing spack environments and such.  I don't necessarily need you to test `HAVE_ADIOS=false`, I'm more interested in making sure MALI runs as usual when nothing about ADIOS is specified (so hopefully `HAVE_ADIOS=true` is implied).  If you can just run the `full_integration` suite from compass using this branch somewhere you like to run, that would be what I had in mind.

matthewhoffman (approved):
I tested the MALI full_integration suite on Chicoma with these changes and everything ran fine.

xylar:
Thanks @matthewhoffman!

mark-petersen (approved):
Tested MPAS-Ocean standalone on crysalis (intel), chicoma (gnu) and perlmutter (gnu), both optimized and debug, and with `HAVE_ADIOS=false` and `HAVE_ADIOS=true`. All combinations compile and pass the nightly test suite. Optimized versions match master bfb for nightly test suite.

darincomeau (approved):
Tested MPAS-Seaice on crysalis (intel), passes the new `single_column` sea ice test in polaris. Build fails there with gfortran, but also fails with recent master, so not related to this PR.

xylar:
Thanks so much @mark-petersen and @darincomeau for your testing!

jonbob:
@rljacob -- are you happy with the end result?

rljacob (approved):

jonbob:
passes:

merged to next

jonbob:
merged to master

xylar:
Thank you @jonbob!

PR #5734, opened 2023-05-30
Fix ocean implicit bottom drag using log law-of-the-wall

cbegeman:
This PR offers the following improvements to the log law-of-the-wall formulation for implicit bottom drag (notation Cd for the drag coefficient is used below):

- Change the config option name so that it is more consistent with other formulations, i.e. `config_implicit_bottom_drag_type = 'loglaw'`
- Change several hard-coded parameters to config options (i.e., min and max Cd values, roughness length)
- Modify the formulation so that layer thicknesses less than 2 times the roughness length can be used and still offer reasonable solutions for Cd. This requires simultaneously modifying the vertical viscosity so that the total drag applied is consistent with the log law-of-the-wall (and the previous formulation).
- Extend the applicability of the formulation to multi-layer configurations. Previously, the computation implicitly assumed that the log-law was only being used in single layer configurations.

The extension to multi-layer configurations required adding the bottom drag term to each k-level in the tridiagonal matrix. This motivated separating the log law-of-the-wall formulation into its own subroutine since the tridiagonal matrix differs in construction from the Manning's formulation.

[BFB] - stealth feature


cbegeman:
For further discussion, see https://github.com/E3SM-Ocean-Discussion/E3SM/pull/40.

mark-petersen:
@cbegeman there are conflicts with master in the bug fix region, so I think it is better for you to fix the conflicts before I test, since you've been thinking about this part of the code. Thanks!

cbegeman:
@mark-petersen Thanks for pointing that out. All of the merge conflicts should now be fixed. I retested the PR with the 1km resolution drying slope cases and a drying slope case with the log-law option turned on.

mark-petersen (approved):
Approved based on previous review at https://github.com/E3SM-Ocean-Discussion/E3SM/pull/40. I also tested the head of the PR with gnu on chicoma. With the nightly stand-alone test suite, it matches master bfb with the default flags for gnu optimized. I also tested `config_implicit_bottom_drag_type = 'loglaw'` using gnu optimized and debug. It passes the nightly test suite for both, but does not match master with default flags, as expected. I compiled with intel on chrysalis as well, but did not test there.

A minor note: On the PR description, the first bullet should now be `config_implicit_bottom_drag_type = 'loglaw'`

cbegeman:
Thanks, @mark-petersen! I changed the PR description; thanks for pointing that out!

cbegeman:
@jonbob This is ready for your testing. Let me know how I can help!

jonbob:
@cbegeman -- I ran the scripts to make the bld files consistent with Registry and there are a couple of things you had missed. Is it OK if I commit them to your fork?

cbegeman:
@jonbob That would be great! Thanks for catching those.

jonbob:
@cbegeman -- done. I'm assuming this is a stealth feature meant for E3SM usage, which means we'll need to run 10-year fully-coupled cases for comparison as well as add a new test to the stealth suite

cbegeman:
> @cbegeman -- done. I'm assuming this is a stealth feature meant for E3SM usage, which means we'll need to run 10-year fully-coupled cases for comparison as well as add a new test to the stealth suite

Yes, we will have to run the 10-year comparison.

This PR modifies an existing stealth feature. Are we adding stealth tests whenever we modify a stealth feature that isn't in the stealth suite yet? Are these the only tests so far in the stealth suite?
https://github.com/cbegeman/E3SM/blob/30c93907b6291be88066d75aeb25eeaeb3672217/cime_config/tests.py#L208-L213
Thanks in advance for clarifying!

jonbob:
@cbegeman -- OK, I'll get those longer runs started. And the requirement to add a new test for each stealth feature is relatively recent, so it's likely we have added some earlier without adding a corresponding test. In any case, it makes sense to add one now so I'll work on that as well.

cbegeman:
@jonbob The comparison looks as-expected to me. The differences are quite small, with the exception of sea ice concentration in continental shelf regions, which would be where differences in bottom drag could make a bigger difference.

mark-petersen:
Yes, diffs look reasonable. They are mostly in the smallest colorbar ranges. Thanks!

jonbob:
I added a new stealth test and a tested supporting it, if you want to take a peek @cbegeman. Otherwise this is good to go

jonbob:
passes (with expected NML DIFFs):

merged to next

jonbob:
merged to master and expected NML DIFFs blessed on all platforms except chrysalis next, which hasn't yet reported. Also the new stealth test will need to have new baselines blessed when it runs

jonbob:
and NML DIFFs blessed on chrysalis as well

cbegeman:
@jonbob Thank you so much for pushing this forward while I was on vacation! Happy to have this feature in the code!

PR #5743, opened 2023-06-05
Add a node to allactive BGC nightly runs on Chrysalis

amametjanov:
Prior 4-node cases were running into OOM errors.

[NML] - for 2 allactive BGC cases
[BFB]


amametjanov:
Merged to master in https://github.com/E3SM-Project/E3SM/commit/0bd06172c98e2a216da7522564123962ed190194 .

Issue #5692, opened 2023-05-17
AMD compiler: mozart/UCI_cld_sub_mod.f90 parsing errors


sarats:
On Frontier, the AMD compiler is having trouble parsing `mozart/UCI_cld_sub_mod.f90`.

Compiler version
```
$ amdflang --version
AMD flang-new version 15.0.0 (https://github.com/RadeonOpenCompute/llvm-project roc-5.4.3 23045 a29fe425c7b0e5aba97ed2f95f61fd5ecba68aed)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /opt/rocm-5.4.3/llvm/bin
```

```
[...]

PR #5745, opened 2023-06-05
Remove CNL specific CMake macros

sarats:
Removing CMake macro file used on old Cray supercomputers using Catamount OS.


jgfouca (approved):

dqwu:
@sarats 
The removed CNL.cmake has the following settings to use Cray wrappers:
```
set(MPICC "cc")
set(MPIFC "ftn")
set(MPICXX "CC")
```

For Perlmutter (CNL.cmake is not used), gnu_pm-cpu.cmake has:
```
set(MPICC "cc")
set(MPICXX "CC")
set(MPIFC "ftn")
```

Please check if Crusher or Frontier also needs similar settings in compiler_machine.cmake macro files after CNL.cmake is removed.

sarats:
Will check. Cray compilers have it.

dqwu:
@jgfouca @sarats 
I tested latest next branch (with this PR merged) on Frontier and there are SCORPIO build errors.

Before this PR, Cray MPI wrappers are used (set in CNL.cmake) to configure SCORPIO:
`CC=cc CXX=CC FC=ftn ...`

With CNL.cmake removed by this PR, non-Cray MPI wrappers are used instead:
`CC=mpicc CXX=mpicxx FC=mpif90`

In addition to the build errors caused by non-Cray wrappers, for non-Cray MPI wrappers, CMake 3.22 or higher has confirmed hanging issues when configuring SCORPIO (Frontier uses CMake 3.21 so far), see https://github.com/E3SM-Project/scorpio/issues/517

If CNL.cmake (set Cray wrappers) is no longer being used for Frontier (or Crusher), please consider updating xxxx_frontier.cmake files to use Cray wrappers (similar to gnu_pm-cpu.cmake for Perlmutter) to avoid build errors and CMake hanging issues.

cime_config/machines/cmake_macros/CNL.cmake
```diff
@@ -1,18 +0,0 @@
-if (NOT MPILIB STREQUAL impi)
-  string(APPEND CMAKE_OPTS " -DCMAKE_SYSTEM_NAME=Catamount")
-endif()
-string(APPEND CPPDEFS " -DLINUX")
-if (COMP_NAME STREQUAL gptl)
-  string(APPEND CPPDEFS " -DHAVE_NANOTIME -DBIT64 -DHAVE_VPRINTF -DHAVE_BACKTRACE -DHAVE_SLASHPROC -DHAVE_COMM_F2C -DHAVE_TIMES -DHAVE_GETTIMEOFDAY")
-endif()
```
sarats, line 7: Note to self: Will check if Frontier/Crusher configurations need these. 
Esp. w.r.t GPTL if they are already present or redundant.

sarats:
@dqwu Your tests are with GNU, right? Those don't have the Cray wrappers set.

dqwu:
> @dqwu Your tests are with GNU, right? Those don't have the Cray wrappers set.

Right, the default compiler is GNU:     `<COMPILERS>gnu,crayclang,amdclang,gnugpu,crayclanggpu,amdclanggpu</COMPILERS>`

And GNU compiler can use Cray wrappers, see gnu_pm-cpu.cmake for Perlmutter.

cime_config/machines/cmake_macros/CNL.cmake
sarats: @amametjanov We usually have `-DHAVE_SLASHPROC`. What else do we really need?

`-DLINUX` is checked in `shr_sys_mod` conditionals.

amametjanov: Grepping through `share`, it appears that only `-DHAVE_VPRINTF -DHAVE_BACKTRACE` can be removed.

jgfouca:
@sarats , it looks like this likely broke the builds for most tests on crusher / e3sm_integration_next_gnu.

cime_config/machines/cmake_macros/CNL.cmake
sarats: So I wonder why don't we use rest of the options on other machines too? If we expect almost all machines to have some of these, why don't we move them to defaults in buildlib?

sarats:
> default compiler is GNU
To avoid any confusion, Cray compilers are preferred and I will set them as default on this machine. 
GNU is the fallback option. 
@jgfouca Does CIME consider the first compiler entry as the default?

AMD compiler have known issues and not ready for use again https://github.com/E3SM-Project/E3SM/issues/5692****

cime_config/machines/cmake_macros/CNL.cmake
amametjanov: Some machines might not have low-level x86 intrinsics like `rdtsc` used by `HAVE_NANOTIME`.
`/proc` was not available on all machines: maybe it is now.
Putting all these configure options in the top-level OS-specific XML/CMake file (instead of a build script) might have been appropriate at the time. Don't know.

rljacob: Might be worth opening an issue on CIME to poll if OS-specific cmake files are useful anywhere.

jgfouca:
@sarats , any progress on this?

rljacob:
ping @sarats please see above.

sarats:
On vacation, will look next week.

rljacob:
@sarats please take care of this.

sarats:
For future reference:

GPTL build options

-DHAVE_SLASHPROC is needed in GPTLget_memusage to obtain data directly from the /proc filesystem (present on most current systems). That file has logic for BGP and BGQ that can probably be removed in a later PR.

-DHAVE_NANOTIME and -DBIT64 are used in combination in gptl.c to invoke `rdtsc` (inline assembly) and store in timers. Code snippet from PAPI and also see: https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html

-DHAVE_VPRINTF - GPTL assumes this is present by default, no corresponding #ifdef block. There is a NO_VPRINTF option which just prints an error message. 

-DHAVE_BACKTRACE - Not used

-DDHAVE_COMM_F2C : Seems to be default in timing/private.h. No harm in retaining.

-DHAVE_TIMES : Enables CPU stats and needed

-DHAVE_GETTIMEOFDAY: Needed and present on all machines.

grnydawn:
> 

@sarats Ok. I will apply the new commits to next and will test them on Crusher(or Frontier).

sarats:
@jgfouca I haven't touched the Scream compiler files. Currently, I removed the CNL reference in config_machines.xml. So, check how you want to handle any changes that are required before doing a upstream merge for Scream. 

We should really converge to a single machine entry soon to avoid redundant work. I don't know what outstanding issues are a hurdle to that (we resolved the Kokkos config discrepancy). Other than some Depends file where optimization is turned off for all CICE in SCREAM.

grnydawn:
@sarats @jgfouca , I locally merged the commits into next. Currently it is being tested on Frontier. After looking at the test results so far, there are issues in "/eam/src/physics/cam/zm_conv_intr.F90" All test cases are failed at the file.

It seems that there exists a bug in the file of using "real(r8) :: jctop(pcols)" variable. The real variable is used in the place of DO loop variable or as a scalar integer index like below:

do k = jctop(i),pver
bottom = state%pint(i,pver+1) - state%pmid(i,jctop(i))

It apparently is not related to Sarat's commits, and if there is no other concerns, I think I will push this updated next to upstream.

sarats:
Fixes for the unrelated ZM issue are in https://github.com/E3SM-Project/E3SM/pull/5805/

jgfouca:
@sarats , you should not have to modify anything in scream/eamxx since they manage their own flags. We have discussed scream using the CIME flags like the other components so it's in the works. Thanks for the summary of what these CPPDEFs do.


@grnydawn , please feel free to push to this branch and next.

grnydawn:
@jgfouca , This branch has been merged into next.

dqwu:
@sarats "FI_CXI_RX_MATCH_MODE=software" mentioned in issue #5750 is not included in this PR, right?
FYI, that ENV variable has already been set for scream (in scream feature branch machines/frontier).

sarats:
We discussed the software tag matching during our regular meeting with Cray. I don't see a reason to set it as a machine-wide default at this point. As I mentioned during our conversation at all-hands, I want to be careful and deliberate about machine-wide default options and assess performance impact.

You should join the bi-weekly calls with Cray for more context/details.

jgfouca:
@grnydawn , @sarats , can I merge to master now?

sarats:
I kicked off the Gitlab CI testing at OLCF this afternoon, they are still running.

sarats:
FYI, https://my.cdash.org/viewTest.php?buildid=2373368 (AMD Clang has 41 pass and 77 fails - an improvement ;) )
Most of those fails are known issues.

sarats:
Since it has known fails without workarounds (OLCF/AMD not providing fixes), I'm going to disable regular testing with AMD.

PR #5751, opened 2023-06-08
Hommexx: Add state hash output capability for diagnosing run-to-run non-BFBness

ambrad:
EAMxx can print a global hash value before and after each process. Since Hommexx is a large EAMxx process, we want finer-grained hashing in Hommexx. This feature optionally outputs a global hash value after each boundary exchange and, at higher diagnostics level, before the boundary exchange, as well.

Add internal_diagnostics_level = 1 or 2 (higher) to the Homme namelist to enable it. The default value is 0 (none).

One HOMMEBFB test runs with internal_diagnostics_level = 2.

infrastructure_ut unit tests the basic hashing routines.

[BFB]


ambrad:
e3sm_developer and homme_integration pass on Chrysalis.

components/homme/src/share/cxx/utilities/Hash.hpp
```diff
@@ -0,0 +1,69 @@
+/********************************************************************************
+ * HOMMEXX 1.0: Copyright of Sandia Corporation
+ * This software is released under the BSD license
+ * See the file 'COPYRIGHT' in the HOMMEXX/src/share/cxx directory
+ *******************************************************************************/
+
+#ifndef HASH_HPP
```
bartgol, line 7: Just to be safe, since "hash" is a common word, let's prepent HOMMEXX or something, so we don't accidentally risk to clash our header against God knows what package we may link to e3sm down the road...

components/homme/src/share/cxx/utilities/InternalDiagnostics.hpp
```diff
@@ -0,0 +1,19 @@
+/********************************************************************************
+ * HOMMEXX 1.0: Copyright of Sandia Corporation
+ * This software is released under the BSD license
+ * See the file 'COPYRIGHT' in the HOMMEXX/src/share/cxx directory
+ *******************************************************************************/
+
+#ifndef INTERNAL_DIAGNOSTICS_HPP
+#define INTERNAL_DIAGNOSTICS_HPP
```
bartgol, line 8: Maybe here too, although it's probably less likely to get a macro clash than for `HASH_HPP` ...

bartgol (approved):
Looks good. I only have one minor request, for extra safety.

components/homme/src/share/cxx/utilities/Hash.hpp
ambrad: Good catch.

bartgol (approved):

components/homme/src/share/cxx/utilities/Hash.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+typedef std::uint64_t HashType;
+
+// Each hash function accumulates v into accum using a hash of its bits.
+
+KOKKOS_INLINE_FUNCTION void hash (const HashType v, HashType& accum) {
+  constexpr auto first_bit = 1ULL << 63;
+  accum += ~first_bit & v; // no overflow
+  accum ^=  first_bit & v; // handle most significant bit  
+}
```
sarats, line 28: Is the hash a simple bitwise accumulation or did I miss something?
Trying to understand potential for hash collisions.

ambrad: It's just double -> unsigned long long, then add these. I handle the most significant bit manually so there isn't an overflow if overflow detection is on. -x + x isn't 0 (as with signed add) and x + x isn't 0 (as with XOR), so it's good enough for non-BFB detection.

components/homme/test_execs/share_kokkos_ut/infrastructure_ut.cpp
```diff
@@ -517,3 +519,47 @@ TEST_CASE("hvcoord","hvcoord") {
...
+  }
+
+  { // Negated values should not be nullified.
+    HashType a = 0;
+    hash( M_PI, a);
+    hash(-M_PI, a);
+    REQUIRE(a != 0);
+  }
+
+  { // The hasher is sensitive to diffs that double accum is not.
```
ambrad, line 554: @sarats note this unit test.

PR #5748, opened 2023-06-08
Add missing fates_levleaf dim

jayeshkrishna:
Adding missing fates_levleaf dim, which is used for defining
the fates_levleaf variable

This PR is non-BFB for FATES tests.

[non-BFB]

PR #5752, opened 2023-06-09
updated fates_levleaf dimvar type

rgknox:
Update the fates_levleaf dimvar type from an integer to a real (which is correct, as the values are real numbers).  The unit was changed from VAI (which is not a unit) to the true units of m2 / m2.

### DIFF
--- a/components/elm/src/main/histFileMod.F90
+++ b/components/elm/src/main/histFileMod.F90
@@ -2469,8 +2469,8 @@ contains
                    long_name='FATES cwd size class', ncid=nfid(t))
              call ncd_defvar(varname='fates_levcan',xtype=ncd_int, dim1name='fates_levcan', &
                    long_name='FATES canopy level', ncid=nfid(t))
-             call ncd_defvar(varname='fates_levleaf',xtype=ncd_int, dim1name='fates_levleaf', &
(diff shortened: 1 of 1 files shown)


jayeshkrishna:
Is the content/data of the variable changing (from levels to area)?

rgknox:
I updated the first comment, is that what you meant?

rgknox:
I just want to point out that this PR is not to master, this is to @jayeshkrishna 's branch, which has it's own PR into master, see #5748.

peterdschwartz:
@rgknox  Sorry this is a new workflow to me. Is @jayeshkrishna handling the merge to their branch?

rgknox:
@peterdschwartz I don't know, I'm used to people working on forks, this is new to me too.  But my take is that since it's  @jayeshkrishna 's branch, they ultimately decide if it should be merged.

rljacob:
I wonder why this is even showing up here in E3SM-Project/E3SM instead of just in Jayesh's fork.

rljacob:
Oh I see its not within Jayesh's fork, its within his branch.  Yes normally you would just push commits directly to his branch (after talking with him).  We don't want to clutter this github repo with branch-to-branch PRs.

PR #5756, opened 2023-06-12
Update intel compiler version on pm-cpu for maint-1.0

ndkeen:
After machine maintenance, module version being used was no longer available. Update to current.


ndkeen:
merged to maint-1.0

PR #5757, opened 2023-06-12
On pm-cpu, Update module version of intel compiler

ndkeen:
After maintenance, the version being used was missing and now trying `intel-oneapi/2023.1.0`

There were a few diffs with baselines, but I'm not sure if they were already there or not, so will say nbfb.
[NBFB]


ndkeen:
merged to next

PR #5759, opened 2023-06-12
add BTAU variables interpolated to pressure levels

whannah1:
This adds optional versions of variables such as "BTAUXSn28" interpolated to specific pressure levels (100 mb, 50 mb ,30 mb ,10 mb) to assist in diagnosing the QBO. These history variables are the parameterized convective gravity wave drag at each wave phase speed. The new variables will not be output be default so this change will not impact any tests or configurations.

[BFB]

### DIFF
--- a/components/eam/src/physics/cam/gw_drag.F90
+++ b/components/eam/src/physics/cam/gw_drag.F90
@@ -28,7 +28,7 @@ module gw_drag
   use constituents,  only: pcnst
   use physics_types, only: physics_state, physics_ptend, physics_ptend_init
   use spmd_utils,    only: masterproc
-  use cam_history,   only: outfld
+  use cam_history,   only: outfld, hist_fld_active
   use cam_logfile,   only: iulog
   use cam_abortutils,    only: endrun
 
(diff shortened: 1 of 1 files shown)


wlin7 (approved):
All look good. New field names and variables are defined and used consistently.

components/eam/src/physics/cam/gw_drag.F90
```diff
@@ -1203,6 +1212,23 @@ subroutine gw_spec_outflds(prefix, lchnk, ncol, ngwv, c, u, v, xv, yv, &
      call outfld(dumc1x,dummyx,ncol,lchnk)
      call outfld(dumc1y,dummyy,ncol,lchnk)
 
+     if (hist_fld_active(trim(dumc1x)//'_100mb')) then
+        call vertinterp(ncol, pcols, pver, state%pmid, 10000._r8, dummyx, dummmy_p_surf)
```
wlin7, line 1216: Do have a problem here. vertinterp expects `pmid, dummyx`, and `dummy_p_surf` to have the same 1st dimsize. `dummyx `could have smaller 1st dimsize (i.e., ncol < pcols). In that case the shape will be mismatched for the 2nd to last argument in the call. 

The simplest is to also declare `dummyx `as `(pcols,pver)`; and explicitly set upper bound in the following statement
`dummyx(:ncol,:) = taux(:,l,:)`

whannah1: ah, good eye. I like your suggestion to change dummyx and dummyy to have dimension pcols.

wlin7 (approved):
Thanks for the update. The new commit addressed the potential shape mismatch. Here to re-approve the PR.

PR #5766, opened 2023-06-15
MPAS: add dynamic time level allocation from namelist flag

mark-petersen:
Currently, the number of time levels in all mpas cores must be set in the `Registry.xml` file for each structure as follows:
```
<var_struct name="state" time_levs="2">
```
This PR allows dynamic time level specifications, so that each `var_struct` may be controlled with a namelist flag:
```
<var_struct name="state" time_levs="namelist:config_number_of_time_levels">
```
and the number of time levels may be changed from the namelist at run time.

This option is now available to all MPAS cores, but is optional, so if no changes are made to the Registry file, the behavior remains identical to before. The PR includes a check in `parse.c` to ensure that the config flag name is valid and is an integer.

I have included the addition of a flag here for the ocean core in order to show how it works. The default value for this new flag is 2, so the behavior remains unchanged.

[BFB]
[NML]

### DIFF
--- a/components/mpas-framework/src/tools/registry/gen_inc.c
+++ b/components/mpas-framework/src/tools/registry/gen_inc.c
@@ -1756,203 +1756,204 @@ int parse_var(FILE *fd, ezxml_t registry, ezxml_t superStruct, ezxml_t currentVa
 		fortprintf(fd, "      allocate(%s)\n", pointer_name);
 	}
 
-	for(time_lev = 1; time_lev <= time_levs; time_lev++){
-		if (time_levs > 1) {
-			snprintf(pointer_name_arr, 1024, "%s(%d)", pointer_name, time_lev);
+	if (time_levs > 1) {
+	    fortprintf(fd, "      do iTime = 1,%d\n", time_levs);
+		snprintf(pointer_name_arr, 1024, "%s(iTime)", pointer_name);
+	} else {
+		snprintf(pointer_name_arr, 1024, "%s", pointer_name);
+	}
+	fortprintf(fd, "! Setting up metadata\n");
+	fortprintf(fd, "         %s %% fieldName = '%s'\n", pointer_name_arr, varname);
+	fortprintf(fd, "         %s %% outputFieldName = '%s'\n", pointer_name_arr, varname_in_output);
+	fortprintf(fd, "         %s %% isVarArray = .false.\n", pointer_name_arr);
+	if (decomp != -1) {
+		fortprintf(fd, "         %s %% isDecomposed = .true.\n", pointer_name_arr);
+	} else {
+		fortprintf(fd, "         %s %% isDecomposed = .false.\n", pointer_name_arr);
+	}
+
+	if(hasTime) {
+		fortprintf(fd, "         %s %% hasTimeDimension = .true.\n", pointer_name_arr);
+	} else {
+		fortprintf(fd, "         %s %% hasTimeDimension = .false.\n", pointer_name_arr);
+	}
+
+	if(ndims > 0){
+		if(persistence == SCRATCH){
+			fortprintf(fd, "         %s %% isPersistent = .false.\n", pointer_name_arr);
+			fortprintf(fd, "         %s %% isActive = .false.\n", pointer_name_arr);
 		} else {
-			snprintf(pointer_name_arr, 1024, "%s", pointer_name);
-		}
-		fortprintf(fd, "\n");
-		fortprintf(fd, "! Setting up time level %d\n", time_lev);
-		fortprintf(fd, "      %s %% fieldName = '%s'\n", pointer_name_arr, varname);
-		fortprintf(fd, "      %s %% outputFieldName = '%s'\n", pointer_name_arr, varname_in_output);
-		fortprintf(fd, "      %s %% isVarArray = .false.\n", pointer_name_arr);
-		if (decomp != -1) {
-			fortprintf(fd, "      %s %% isDecomposed = .true.\n", pointer_name_arr);
-		} else {
-			fortprintf(fd, "      %s %% isDecomposed = .false.\n", pointer_name_arr);
+			fortprintf(fd, "         %s %% isPersistent = .true.\n", pointer_name_arr);
+			fortprintf(fd, "         %s %% isActive = .false.\n", pointer_name_arr);
 		}
 
-		if(hasTime) {
-			fortprintf(fd, "      %s %% hasTimeDimension = .true.\n", pointer_name_arr);
-		} else {
-			fortprintf(fd, "      %s %% hasTimeDimension = .false.\n", pointer_name_arr);
+		// Setup dimensions
+		fortprintf(fd, "! Setting up dimensions\n");
+		string = strdup(vardims);
+		tofree = string;
+		i = 1;
+		token = strsep(&string, " ");
+		if(strncmp(token, "Time", 1024) != 0){
+			fortprintf(fd, "         %s %% dimNames(%d) = '%s'\n", pointer_name_arr, i, token);
+			i++;
 		}
-
-		if(ndims > 0){
-			if(persistence == SCRATCH){
-				fortprintf(fd, "      %s %% isPersistent = .false.\n", pointer_name_arr);
-				fortprintf(fd, "      %s %% isActive = .false.\n", pointer_name_arr);
-			} else {
-				fortprintf(fd, "      %s %% isPersistent = .true.\n", pointer_name_arr);
-				fortprintf(fd, "      %s %% isActive = .false.\n", pointer_name_arr);
-			}
-
-			// Setup dimensions
-			fortprintf(fd, "! Setting up dimensions\n");
-			string = strdup(vardims);
-			tofree = string;
-			i = 1;
-			token = strsep(&string, " ");
+		while( (token = strsep(&string, " ")) != NULL){
 			if(strncmp(token, "Time", 1024) != 0){
-				fortprintf(fd, "      %s %% dimNames(%d) = '%s'\n", pointer_name_arr, i, token);
+				fortprintf(fd, "         %s %% dimNames(%d) = '%s'\n", pointer_name_arr, i, token);
 				i++;
 			}
-			while( (token = strsep(&string, " ")) != NULL){
-				if(strncmp(token, "Time", 1024) != 0){
-					fortprintf(fd, "      %s %% dimNames(%d) = '%s'\n", pointer_name_arr, i, token);
-					i++;
-				}
-			}
-			free(tofree);
 		}
-
-		fortprintf(fd, "      %s %% defaultValue = %s\n", pointer_name_arr, default_value);
-		if ( ndims == 0 ) {
-			fortprintf(fd, "      %s %% scalar = %s\n", pointer_name_arr, default_value);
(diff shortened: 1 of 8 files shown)


components/mpas-framework/src/tools/registry/gen_inc.c
```diff
@@ -1750,20 +1764,21 @@ int parse_var(FILE *fd, ezxml_t registry, ezxml_t superStruct, ezxml_t currentVa
 
 	// Determine name of pointer for this field.
 	set_pointer_name(type, ndims, pointer_name, time_levs);
-	if (time_levs > 1) {
-		fortprintf(fd, "      allocate(%s(%d))\n", pointer_name, time_levs);
-	} else {
-		fortprintf(fd, "      allocate(%s)\n", pointer_name);
-	}
 
-	for(time_lev = 1; time_lev <= time_levs; time_lev++){
```
mark-petersen, line 1759: Previously, the loop over time for array allocations was in the preprocessing text parsing code. This hard-codes the number or time levels, so I had to change it.

components/mpas-framework/src/tools/registry/gen_inc.c
```diff
@@ -1750,20 +1764,21 @@ int parse_var(FILE *fd, ezxml_t registry, ezxml_t superStruct, ezxml_t currentVa
...
-		if (time_levs > 1) {
-			snprintf(pointer_name_arr, 1024, "%s(%d)", pointer_name, time_lev);
+	if (time_levs > 1 || time_levs_from_config) {
+		if (time_levs_from_config) {
+			fortprintf(fd, "      allocate(%s(config_flag_time_levs))\n", pointer_name);
+			fortprintf(fd, "      do iTime = 1,config_flag_time_levs\n");
```
mark-petersen, line 1771: Here I move the loop over time levels to the auto-generated Fortran code. This one is used if a `namelist:config_name` is specified.

components/mpas-framework/src/tools/registry/gen_inc.c
```diff
@@ -1750,20 +1764,21 @@ int parse_var(FILE *fd, ezxml_t registry, ezxml_t superStruct, ezxml_t currentVa
...
+		} else {
+			fortprintf(fd, "      allocate(%s(%d))\n", pointer_name, time_levs);
+			fortprintf(fd, "      do iTime = 1,%d\n",time_levs);
```
mark-petersen, line 1774: This one is used if a number digit is specified in the Registry file.

components/mpas-framework/src/tools/registry/parse.c
```diff
@@ -222,7 +222,34 @@ int validate_reg_xml(ezxml_t registry)/*{{{*/
...
+				snprintf(name_holder, 1024, "%s",time_levs);
+				snprintf(name_holder, 1024, "%s",(name_holder)+9);
+				for (nmlrecs_xml = ezxml_child(registry, "nml_record"); nmlrecs_xml; nmlrecs_xml = nmlrecs_xml->next){
+
+					for (nmlopt_xml = ezxml_child(nmlrecs_xml, "nml_option"); nmlopt_xml; nmlopt_xml = nmlopt_xml->next){
+						nmloptname = ezxml_attr(nmlopt_xml, "name");
+						nmlopttype = ezxml_attr(nmlopt_xml, "type");
+
+						if (strncmp(name_holder, nmloptname, 1024) == 0){
+							if (strcasecmp("integer", nmlopttype) != 0){
```
mark-petersen, line 237: Here I check if the requested namelist flag exists and is an integer.

mark-petersen:
I tested by compiling with gnu and intel on chrysalis, debug and optimized. It passes the MPAS-O nightly suite with no changes to the PR. I then added the lines
```
+      call mpas_pool_get_array(statePool, 'normalVelocity', normalVelocityCur, 3)
+      call mpas_pool_get_array(statePool, 'normalVelocity', normalVelocityCur, 4)
```
to the beginning of a time-step. This requests time level 3 and 4. This will run without any errors if I change my namelist to
```
config_number_of_time_levels = 4
```
but if this flag is less than 4, the simulation dies with a segmentation fault.

components/mpas-framework/src/tools/registry/gen_inc.c
mark-petersen: Note, the flag in your `Registry.xml` file can be named whatever you want. This variable name, `config_flag_time_levs`, is just a pointer to the flag that you specify with the `namelist:` prefix.

mark-petersen:
@darincomeau and @matthewhoffman, could you please test this PR in the sea ice and albany-landice cores? I was able to compile those cores on this PR with both gnu and intel. If you could run a test in addition to that, we can be confident that this does not harm your core.  

If you would like to try to specify time levels in your core with a namelist flag in the Registry with `time_levs="namelist:config..."` feel free, but I think the ocean testing was sufficient to show that is working.

components/mpas-framework/src/tools/registry/gen_inc.c
```diff
@@ -1723,10 +1726,21 @@ int parse_var(FILE *fd, ezxml_t registry, ezxml_t superStruct, ezxml_t currentVa
...
 
+	time_levs_from_config = false;
 	if(vartimelevs){
-		time_levs = atoi(vartimelevs);
-		if(time_levs < 1){
-			time_levs = 1;
+		if(strncmp(vartimelevs, "namelist:", 9) == 0){
+			time_levs_from_config = true;
+            // set time_levs to 2 to trigger array declarations below
+            time_levs = 2;
```
gcapodag, line 1734: Hi @mark-petersen , what is the reason for not setting here `time_levs` to the number of time levels from the namelist flag?

gcapodag:
I tested these changes on LTS using the hurricane test case and they produce a solution that is BFB with respect to what currently is in the LTS PR.

components/mpas-framework/src/tools/registry/gen_inc.c
xylar: @mark-petersen, indentation here needs to be with tabs, not spaces.

xylar: @gcapodag, you don't know the number of time levels at compile time when this code is called, so you need a placeholder.

But @mark-petersen, I think the placeholder is confusing and you should rely only on `time_levs_from_config`, not `time_levs > ` for this situation.

xylar, line 1734:
```suggestion
```

components/mpas-framework/src/tools/registry/gen_inc.c
```diff
@@ -1750,20 +1764,21 @@ int parse_var(FILE *fd, ezxml_t registry, ezxml_t superStruct, ezxml_t currentVa
...
-	if (time_levs > 1) {
-		fortprintf(fd, "      allocate(%s(%d))\n", pointer_name, time_levs);
-	} else {
-		fortprintf(fd, "      allocate(%s)\n", pointer_name);
-	}
 
-	for(time_lev = 1; time_lev <= time_levs; time_lev++){
-		if (time_levs > 1) {
-			snprintf(pointer_name_arr, 1024, "%s(%d)", pointer_name, time_lev);
+	if (time_levs > 1 || time_levs_from_config) {
```
xylar, line 1768:
```suggestion
	if (time_levs_from_config || time_levs > 1) {
```
This way, it doesn't matter what `time_levs` has in it when `time_levs_from_config == true`.

components/mpas-framework/src/tools/registry/gen_inc.c
```diff
@@ -1952,7 +1967,9 @@ int parse_var(FILE *fd, ezxml_t registry, ezxml_t superStruct, ezxml_t currentVa
 
 		fortprintf(fd, "      %s %% block => block\n", pointer_name_arr);
 
-	}
+	if (time_levs > 1) {
```
xylar, line 1970:
```suggestion
	if (time_levs_from_config || time_levs > 1) {
```

components/mpas-framework/src/tools/registry/gen_inc.c
```diff
@@ -1972,13 +1989,19 @@ int parse_var(FILE *fd, ezxml_t registry, ezxml_t superStruct, ezxml_t currentVa
...
+	        fortprintf(fd, "      do iTime = 1,config_flag_time_levs\n");
+    	} else {
+	        fortprintf(fd, "      do iTime = 1,%d\n",time_levs);
+        }
+		snprintf(pointer_name_arr, 1024, "%s(iTime)", pointer_name);
+	} else {
+		snprintf(pointer_name_arr, 1024, "%s", pointer_name);
+	}
+		fortprintf(fd, "         %s%s %% isActive = .true.\n", package_spacing, pointer_name_arr);
+	if (time_levs > 1) {
```
xylar, line 2003:
```suggestion
	if (time_levs_from_config || time_levs > 1) {
```

xylar (commented):
I'd suggest a few changes to the C code.  Once these are in (or at least discussed), I'll run some compass tests.

components/mpas-framework/src/tools/registry/gen_inc.c
mark-petersen: Thanks. I agree with your recommendations and will not set a dummy `time_levs` for this case. Rather, I can use `time_levs_from_config` as @xylar recommended.

mark-petersen:
In the most recent commit I change a large section to correct the indent, but this makes the PR code diffs much more confusing because they don't track as whitespace changes. You can compare just before this commit here, it is easier to review: https://github.com/E3SM-Project/E3SM/pull/5766/files/603995e85a09243a5d7416244dd0857bcca58475..b786c1e5372e75e48367a336193de1a732858ca2

mark-petersen:
I recompiled with gnu and intel and retested the ocean core with 2 and 4 time levels, specified in the namelist.

mark-petersen:
I made the requested changes, and this is ready for another review. Thanks!

darincomeau (approved):
Compiled MPAS-Seaice with intel on Chrysalis and successfully ran single column test suite.

philipwjones (approved):
Approved based on visual inspection and testing by others. This will be a useful change.

gcapodag (approved):
Tested again on Perlmutter and still BFB with `USE_LTS` version of LTS scheme. LTS uses 4 time levels. Thanks @mark-petersen , great work!

mark-petersen:
@matthewhoffman said he will be out of town, and that we could take him off as a reviewer if he didn't get around to it.

mark-petersen:
E3SM namelists scripts have been updated. I tested an E3SM build on chrysalis with `SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.chrysalis_gnu` and it completed successfully.

xylar (approved):
@mark-petersen and @jonbob, sorry for letting this sit, I lost track of it.

The code changes look great.  Thanks @mark-petersen.

I did a test merge of this branch with b22634fe8f (where the E3SM-Project submodule in compass currently points) and ran the `pr` suite on Chrysalis with Intel and OpenMPI, using b22634fe8f as a baseline.  All tests pass and are BFB.

jonbob:
Passes:

merged to next

jonbob:

PR #5630, opened 2023-04-24
Corrects thin ice/snow treatment of enthalpy and other tracers

njeffery:
This fix redistributes enthalpy and other tracers evenly in snow and ice when their respective thicknesses are  < 1e-15 . Previously, these tracers were zero'd non-conservatively.

Also corrects a bug in the zeroing of snow thicknesses, and removes snow in thickness_changes if the ice vanishes.

Fixes #5124 

[non-BFB]

PR #5771, opened 2023-06-20
Corrects thin ice/snow treatment of enthalpy and other tracers

darincomeau:
Bringing sea ice bug fix from https://github.com/E3SM-Project/E3SM/pull/5630 into `maint-2.1`.

This fix redistributes enthalpy and other tracers evenly in snow and ice when their respective thicknesses are < 1e-15 . Previously, these tracers were zero'd non-conservatively.

Also corrects a bug in the zeroing of snow thicknesses, and removes snow in thickness_changes if the ice vanishes.

Fixes https://github.com/E3SM-Project/E3SM/issues/5124

[non-BFB]

### DIFF
--- a/components/mpas-seaice/src/column/ice_therm_vertical.F90
+++ b/components/mpas-seaice/src/column/ice_therm_vertical.F90
@@ -1710,8 +1710,8 @@
       !-----------------------------------------------------------------
 
       if (ktherm == 2) then
-         do k = 1, nslyr
-            if (hsn <= puny) then
+         if (hsn <= puny) then
+            do k = 1, nslyr
                fhocnn = fhocnn &
                       + zqsn(k)*hsn/(real(nslyr,kind=dbl_kind)*dt)
                zqsn(k) = -rhos*Lfresh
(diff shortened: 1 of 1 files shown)


rljacob:
we aren't supposed to fix bugs on maint branches.  The 2.1 climate is what it is.  If you want to do 2.1 runs with this bug fixed, make a branch off of maint-2.1 with a name indicating what is fixed and then leave that branch unmerged.

proteanplanet (commented):
Thanks @rljacob. I was wondering the same thing.  Version 2.1 needs to reproduce exactly what previous simulations have given, right or wrong.  I suggest this PR is killed.

darincomeau:
Ok, I was confused. It says on Confluence that maintenance branches are for updates and bug fixes, but looking through bug fix PRs that have gone into various `maint` branches it's only a small subset of bug fixes. I certainly understand not bringing in answer changing PRs to `maint`.  Sorry for the confusion, I'll close this and https://github.com/E3SM-Project/E3SM/pull/5770.

PR #5773, opened 2023-06-22
Moves a land test to a different test suite

bishtgautam:
The land test for simple domain decomposition, which uses 2 threads, is moved 
to the `e3sm_land_exenoshare`.

[BFB]


amametjanov (approved):

peterdschwartz:
merged to master -- new tests on pm-cpu nvidia needs to be blessed

PR #5774, opened 2023-06-22
add cloud-borne aerosols to 2D burdens, enable MOM radiation diagnostics for MAM5 configurations

### LINKED ISSUE #5788: Split SO4 burden into stratosphere and troposphere components
We really need to split SO4 burden into stratosphere and troposphere outputs. Probably this should in the other outputs PR open: https://github.com/E3SM-Project/E3SM/pull/5774

_Originally posted by @mahf708 in https://github.com/E3SM-Project/E3SM/pull/5781#discussion_r1248865119_

mahf708:
Add cloud-borne aerosols to 2D burden outputs, split SO4 burden to strato and tropo parts, and correctly enable MOM.

- This PR Adds new diagnostic fields to output the total aerosol 2D burdens (both interstitial and cloud-borne). The cloud-borne aerosol burden is expected to be less than 10% of the total burden. 
- Split SO4 burden into stratospheric and tropospheric components to better track benchmarks.
- Correctly enable MOM inside the modal_aer_opt module so that the correct diagnostic fields are outputted.

Fix #5788

[BFB]

--------------------------------------------------------------------------------------------------------------------
See figure below from a two-month free-running simulation showing that these added new terms (`ABURDEN...`) are working as expected vis-a-vis their predecessors (`BURDEN...`). While the author/committer of the code changes is @mahf708, this work is produced by the collaborative effort of @mahf708, @susburrows, and @jmuelmen. 

This came up during the coupled team meeting, internal link: https://acme-climate.atlassian.net/wiki/spaces/CM/pages/3749576957/2023-04-17+Coupled+Model+-+Meeting+notes. The issue was discussed in https://github.com/E3SM-Project/e3sm_diags/pull/676 and https://github.com/E3SM-Project/e3sm_diags/issues/681. Continues from https://github.com/E3SM-Project/v3atm/pull/65

Xref: https://github.com/E3SM-Project/e3sm_diags/issues/681. 

<details><summary>Plots showing difference between BURDEN... and ABURDEN... terms</summary>
<p>

test-aburd

</p>
</details> 

**

Split SO4 burden into stratospheric and tropospheric components to better track benchmarks.

Fix #5788

<details><summary>Results for the decomposition</summary>
<p>

| BURDEN  | ABURDEN  | ABURDEN_STR | ABURDEN_TRO |
| ------- | -------  | ----------- | ----------- |
|3.89e-06 | 3.95e-06 | 4.93e-07    | 3.45e-06    |

which make sense...

</p>
</details> 

**

Correctly enable MOM inside the modal_aer_opt module so that the correct diagnostic fields are outputted.

[BFB]


chengzhuzhang:
Hi @mahf708 Naser, thanks for the PR. It looks like this PR is ready to review, I just tagged folks (Mingxuan, Ziming and Wuyin) you suggested for a review..

susburrows:
@mahf708 right, I did see that.  Which means it is consistent with the tropopause definition that is being used elsewhere in the aerosol code, specifically in the lines of code that you pointed to, trop_level is being used to distinguish between the stratosphere for the aerosol extinction calculation in model configurations where the CMIP6 prescribed aerosol extinction is used.  In this case, if we are above the tropopause, the model uses the prescribed stratospheric and volcanic aerosol extinctions, instead of calculating it from the model (prognostic) aerosols.

https://github.com/E3SM-Project/E3SM/blob/9e17b77a4e9ae91874b2fc56f248b5046bdce473/components/eam/src/physics/cam/aer_rad_props.F90#L473C1-L493C9

susburrows:
However, I notice that chemistry.F90 is using a different definition of the tropopause for the gas-phase chemistry:
https://github.com/E3SM-Project/E3SM/blob/9e17b77a4e9ae91874b2fc56f248b5046bdce473/components/eam/src/chemistry/mozart/chemistry.F90#L1504C1-L1509C11

In particular, when the E90 tracer is present, it is being used to find the tropopause.  This probably is a better choice on principle, since the E90 tracer will give us the true "chemical" tropopause.  Also, this code snippet specifies the primary and backup tropopause finding algorithms as:
primary=TROP_ALG_HYBSTOB, backup=TROP_ALG_CLIMATE

... which is different from the default choices in tropopause.F90.

The tropopause identified in chemistry.F90 is the one that gets passed to the subroutine gas_phase_chemdr:

https://github.com/E3SM-Project/E3SM/blob/9e17b77a4e9ae91874b2fc56f248b5046bdce473/components/eam/src/chemistry/mozart/mo_gas_phase_chemdr.F90#L247

And that values is being used in multiple places throughout the gas-phase chemistry code.

So, it might make sense to use that value for consistency with the gas-phase chemistry.  But, I haven't sorted out yet what the new MAM5 is using, which is probably the most pertinent definition to use for dividing the tropospheric and stratospheric sulfate aerosol burdens.

@tangq and/or @keziming may be able to comment on this.

susburrows (commented):
As mentioned in my earlier comments, I think the implementation is correct from reading the code; the diagnostic plots in the PR description also look reasonable.  Could consider switching to use the tropopause identification algorithms that are being used in the gas-phase chemistry (which might also be the ones used by MAM5).  But this is not critical for getting a diagnostic that will give us useful information.

tangq:
Gas-phase chemistry uses the E90 tropopause when E90 is defined. Otherwise, it uses the old tropopause defined by the T lapse rate.

@keziming can comment on what is used in MAM5. My guess is that it doesn't use the E90 tropopause.

> However, I notice that chemistry.F90 is using a different definition of the tropopause for the gas-phase chemistry: https://github.com/E3SM-Project/E3SM/blob/9e17b77a4e9ae91874b2fc56f248b5046bdce473/components/eam/src/chemistry/mozart/chemistry.F90#L1504C1-L1509C11
> 
> In particular, when the E90 tracer is present, it is being used to find the tropopause. This probably is a better choice on principle, since the E90 tracer will give us the true "chemical" tropopause. Also, this code snippet specifies the primary and backup tropopause finding algorithms as: primary=TROP_ALG_HYBSTOB, backup=TROP_ALG_CLIMATE
> 
> ... which is different from the default choices in tropopause.F90.
> 
> The tropopause identified in chemistry.F90 is the one that gets passed to the subroutine gas_phase_chemdr:
> 
> https://github.com/E3SM-Project/E3SM/blob/9e17b77a4e9ae91874b2fc56f248b5046bdce473/components/eam/src/chemistry/mozart/mo_gas_phase_chemdr.F90#L247
> 
> And that values is being used in multiple places throughout the gas-phase chemistry code.
> 
> So, it might make sense to use that value for consistency with the gas-phase chemistry. But, I haven't sorted out yet what the new MAM5 is using, which is probably the most pertinent definition to use for dividing the tropospheric and stratospheric sulfate aerosol burdens.
> 
> @tangq and/or @keziming may be able to comment on this.

susburrows:
> We should likely unify the definition of tropopause throughout the code, right? I can handle that in another PR.

Let's not make any changes to the tropopause definitions that would be answer-changing at this time, @mahf708 .  Even though it may be more consistent, given the short timeline for finalizing v3, I think we should avoid spending time on making decisions about this at this time.

Also, there is always some non-zero risk of unintended side effects, so... if it ain't broke... let's not fix it right now.

susburrows:
@mahf708 , I think you are right that this is the definition being used by MAM5:

`    e90_ndx = get_spc_ndx('E90')
    if (e90_ndx <= 0) then
      call tropopause_find(state, tropLev, primary=TROP_ALG_HYBSTOB, backup=TROP_ALG_CLIMATE)
    else
      call tropopause_e90_3d(state, tmp_tropLev, tropLev, tropFlag, tropFlagInt)
    end if`

It also makes more sense for chemical variables to use the E90 diagnostic.  So, let's go with that here, too, unless @keziming  advises otherwise.

keziming:
> However, I notice that chemistry.F90 is using a different definition of the tropopause for the gas-phase chemistry: https://github.com/E3SM-Project/E3SM/blob/9e17b77a4e9ae91874b2fc56f248b5046bdce473/components/eam/src/chemistry/mozart/chemistry.F90#L1504C1-L1509C11
> 
> In particular, when the E90 tracer is present, it is being used to find the tropopause. This probably is a better choice on principle, since the E90 tracer will give us the true "chemical" tropopause. Also, this code snippet specifies the primary and backup tropopause finding algorithms as: primary=TROP_ALG_HYBSTOB, backup=TROP_ALG_CLIMATE
> 
> ... which is different from the default choices in tropopause.F90.
> 
> The tropopause identified in chemistry.F90 is the one that gets passed to the subroutine gas_phase_chemdr:
> 
> https://github.com/E3SM-Project/E3SM/blob/9e17b77a4e9ae91874b2fc56f248b5046bdce473/components/eam/src/chemistry/mozart/mo_gas_phase_chemdr.F90#L247
> 
> And that values is being used in multiple places throughout the gas-phase chemistry code.
> 
> So, it might make sense to use that value for consistency with the gas-phase chemistry. But, I haven't sorted out yet what the new MAM5 is using, which is probably the most pertinent definition to use for dividing the tropospheric and stratospheric sulfate aerosol burdens.
> 
> @tangq and/or @keziming may be able to comment on this.

@susburrows @mahf708 MAM5 use "troplev" from "tropopause_e90_3d" in this case. This is a 2D tropopause and the same variable name used in E3SMv2 when MAM5 developed. For aerosol process, I think 2D tropopause is a better choice to keep the uniform treatment. For example, if we use 3D tropopause, we would see a layer of tropospheric aerosol in the middle of two layers of stratospheric aerosols. The tropospheric sulfate in the middle layer with higher dry deposition rate can form a sulfate hole in the middle of atmosphere. It would be unreasonable.

components/eam/src/physics/cam/modal_aer_opt.F90
```diff
@@ -788,6 +833,12 @@ subroutine modal_aero_sw(list_idx, dt, state, pbuf, nnite, idxnite, is_cmip6_vol
                   if (trim(spectype) == 'sulfate') then
                      do i = 1, ncol
                         burdenso4(i) = burdenso4(i) + specmmr(i,k)*mass(i,k)
+                        aburdenso4(i) = aburdenso4(i) + specmmr(i,k)*mass(i,k) + aspecmmr(i,k)*mass(i,k)
+                        if ((k .le. trop_level(i)) .and. (is_output_interactive_volc)) then ! in stratosphere
```
kaizhangpnl, line 837: Would it be better to remove "is_output_interactive_volc", so that in the simulation when is_output_interactive_volc = False we can also have the correct stratospheric/tropospheric sulfate burden?

kaizhangpnl (approved):

wlin7:
Merged to v3atm/eam/master_MAM5_wetaero_chemdyg.

(Verified before merge this PR is BFB with base branch for full v3atm configurations)

PR #5777, opened 2023-06-22
Replace dt with 0.0 in ocn_surface_land_ice_fluxes_build_arrays call

### LINKED ISSUE #5772: SMS_D_Ld1.T62_oQU240wLI.GMPAS-IAF-PISMF.compy_intel.mpaso-impl_top_drag failing
A recently changed test that is part of extra coverage fails in mpaso with the following traceback:
````
 23: e3sm.exe           0000000002BD9011  ocn_surface_land_         649  mpas_ocn_surface_land_ice_fluxes.f90
 23: e3sm.exe           000000000255D20A  ocn_comp_mct_mp_o         685  ocn_comp_mct.f90
 23: e3sm.exe           000000000048FC7F  component_mod_mp_         257  component_mod.F90
 23: e3sm.exe           0000000000437625  cime_comp_mod_mp_        1458  cime_comp_mod.F90
 23: e3sm.exe           0000000000486604  MAIN__                    122  cime_driver.F90
````
Analysis shows the failure is due to the call to ocn_surface_land_ice_fluxes_build_arrays in ocn_comp_mct.F:
````
        call ocn_forcing_build_fraction_absorbed_array(meshPool, statePool, forcingPool, ierr, 1)
        call mpas_timer_start("land_ice_build_arrays", .false.)
        call ocn_surface_land_ice_fluxes_build_arrays(meshPool, &
                                                      forcingPool, scratchPool, statePool, dt, ierr)
        call mpas_timer_stop("land_ice_build_arrays")
````
because the variable "dt" has not yet been set

### PR DESCRIPTION
**Pull request by jonbob**
Currently the variable "dt" is used in subroutine ocn_init_mct in a call to ocn_surface_land_ice_fluxes_build_arrays before it has been set. This can cause errors in debug mode with some compilers. This replaces the variable with 0.0 in the argument list and removes dt from the routine, since it is not otherwise needed.

Fixes #5772 
[BFB]

### DIFF
--- a/components/mpas-ocean/driver/ocn_comp_mct.F
+++ b/components/mpas-ocean/driver/ocn_comp_mct.F
@@ -667,6 +667,9 @@ contains
     call mpas_add_clock_alarm(domain % clock, coupleAlarmID, alarmStartTime, alarmTimeStep, ierr=ierr)
     call mpas_reset_clock_alarm(domain_ptr % clock, coupleAlarmID, ierr=ierr)
 
+    timeStep = mpas_get_clock_timestep(domain_ptr % clock, ierr=ierr)
+    call mpas_get_timeInterval(timeStep, dt=dt)
+
     ! Build forcing arrays.
     block_ptr => domain % blocklist
     do while(associated(block_ptr))
@@ -808,9 +811,6 @@ contains
 !   get initial state from driver
 !
 !-----------------------------------------------------------------------
-
-    timeStep = mpas_get_clock_timestep(domain_ptr % clock, ierr=ierr)
-    call mpas_get_timeInterval(timeStep, dt=dt)
 
     call ocn_import_mct(x2o_o, errorCode)
     if (errorCode /= 0) then
--- a/components/mpas-ocean/driver/ocn_comp_mct.F
+++ b/components/mpas-ocean/driver/ocn_comp_mct.F
@@ -207,9 +207,6 @@ contains
     logical, pointer :: config_use_surface_salinity_monthly_restoring
     character (len=StrKIND), pointer :: config_land_ice_flux_mode
 
-    real (kind=RKIND) :: dt
-    type (MPAS_timeInterval_type) :: timeStep
-
     ! Added for coupling interval initialization
     integer, pointer :: index_avgZonalSSHGradient, index_avgMeridionalSSHGradient
     real (kind=RKIND), dimension(:),   pointer :: filteredSSHGradientZonal, filteredSSHGradientMeridional
@@ -667,9 +664,6 @@ contains
     call mpas_add_clock_alarm(domain % clock, coupleAlarmID, alarmStartTime, alarmTimeStep, ierr=ierr)
     call mpas_reset_clock_alarm(domain_ptr % clock, coupleAlarmID, ierr=ierr)
 
-    timeStep = mpas_get_clock_timestep(domain_ptr % clock, ierr=ierr)
-    call mpas_get_timeInterval(timeStep, dt=dt)
-
     ! Build forcing arrays.
     block_ptr => domain % blocklist
     do while(associated(block_ptr))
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by xylar** [comment]
The relevant piece of code where `dt` is used is here:
https://github.com/E3SM-Project/E3SM/blob/bfc21ee67e19e5dc6d32cd5692b7d57c115b911d/components/mpas-ocean/src/shared/mpas_ocn_surface_land_ice_fluxes.F#L671-L672

That code is not really a correct integration of the fluxes, which should really be time-stepped by the appropriate time integrator.   @mark-petersen, that's probably something we should discuss fixing "properly" somehow.

For now, I need to look at the code more carefully to see if passing `dt=0` would actually be the more desirable short-term fix here, rather than passing the correct time step.  Otherwise, heat and freshwater fluxes could be accumulated at initialization as if a time step had already occurred.  Since these fields are used for checking conservation, that seems like it could be a problem.

**Comment by jonbob** [comment]
Thanks @xylar -- I had tested with a standard B-case and it was BFB, but I'm sure it would have some impact when that routine is actually called. I'm concerned that up to now dt might have had any random value from memory? Anyway, I appreciate you giving it a more careful look

**Comment by xylar** [comment]
@jonbob, I definitely share your concern. It needs to be fixed.  My hope is that the garbage value was zero for most compilers so far.  Otherwise, the conservation check that includes the integrals computed with `dt` would likely have been garbage as well.  Anyway, we'll sort this out soon!

**Comment by xylar** [comment]
@jonbob, unfortunately, I still haven't tested this and I'm about to go on vacation until the 17th.  I'll test it when I get back but if you're able to do a BFB test with an ECwISC test case (with melt fluxes turned on and also the conservation check analysis member on), that would leave me more reassured.

**Comment by xylar** [comment]
> If we call line 684 with dt=0, then the lines in question on this PR could just be removed, since dt is a local variable and that is the only place it is used in the subroutine.

@mark-petersen, can you clarify what you mean by this?  I think there may be a good reason that we need to call `ocn_surface_land_ice_fluxes_build_arrays()` here, we just don't want to be accumulating land-ice mass and heat yet.  So It would not be a good idea to remove the call to `ocn_surface_land_ice_fluxes_build_arrays()`.  Are you just suggesting that we don't need `dt` at all, it could be removed and we just pass `dt=0.0_rkind`?

**Comment by mark-petersen** [comment]
We need to keep the call to `ocn_surface_land_ice_fluxes_build_arrays()`, but on init use dt=0, like this:
```
components/mpas-ocean/driver/ocn_comp_mct.F
 683         call ocn_surface_land_ice_fluxes_build_arrays(meshPool, &
 684                                                       forcingPool, scratchPool, statePool, 0.0_RKIND, ierr)
```
Then these lines are not needed, because that was the only place dt was used:
```
812    timeStep = mpas_get_clock_timestep(domain_ptr % clock, ierr=ierr)
813    call mpas_get_timeInterval(timeStep, dt=dt)
```
(lines 812-813 in repo, lines 670-671 in this PR).

**Comment by mark-petersen** [comment]
I agree that would be better. If I remember right, dt needs to be an optional variable in order to specify `dt= ...` on an argument input list, and optional arguments need to be after required arguments. But I get it mixed up with python too.

**Comment by xylar** [comment]
A quick google search (https://www.mrao.cam.ac.uk/~pa/f90Notes/HTMLNotesnode151.html) suggests that this should work:
```
         call ocn_surface_land_ice_fluxes_build_arrays(meshPool, &
                                                       forcingPool, scratchPool, statePool, dt=0.0_RKIND, err=ierr)
```
That is, once you start using keyword arguments, you have to use them for all subsequent arguments.  That's the same as for python.

**Review by xylar — APPROVED** [review]
I ran:
```
./create_test SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.chrysalis_intel --walltime 30:00 --wait
```
on Chrysalis and it passed.

@jonbob, if you can re-test this where you were seeing the compiler errors, that would be great!

mark-petersen (approved):
Thanks @xylar! Those are exactly the changes I was thinking of.

jonbob:
I verified that it now passes:
* SMS_D_Ld1.T62_oQU240wLI.GMPAS-IAF-PISMF.compy_intel.mpaso-impl_top_drag
which was the test that originally displayed this bug

jonbob:
also passes:

merged to next

jonbob:
merged to master

PR #5779, opened 2023-06-24
Update low-res benchmarking on PM-CPU

amametjanov:
Run PM-CPU benchmarks at 16 CPL MPI tasks per node.


amametjanov:
This might not be necessary in the near future: hence closing. Just noting the change from defaults in newer benchmarking runs.

PR #5785, opened 2023-06-29
Cmake build system: Fix CMAKE_MODULE_PATH

### LINKED ISSUE #5783: Update findPetsc.cmake and maybe relocate
Update findPetsc.cmake. The current one comes from cime. Need to point CMAKE_MODULES_PATH to the new cime non_py place for cmake modules. Should these live in CIME or E3SM/components/cmake? Is CESM using the old one in CIME?

jgfouca:
Make a new area to put E3SM cmake modules (components/cmake/modules). Append this dir to CMAKE_MODULE_PATH.
Also, append the correct CIME dir to CMAKE_MODULE_PATH. It will come after the E3SM module dir so E3SM will take precedence. Also, add an updated findPETSc.cmake module to our new module dir.

Fixes #5783 

[BFB]


jgfouca:
@jonbob , I wasn't able to fully test this because I don't have a case that uses PETSc, but I put some print statements in both FindPETSc.cmake modules (in both E3SM and CIME) and they indicated that the correct module is being picked-up when I do a find_package("PETSc") in the E3SM CMakeLists.txt file.

jonbob:
Thanks @jgfouca -- and sorry to take so long before responding (I took a few days off after the meeting). I can test it using a branch that actually uses PETSc and let you know. Thanks again for all your help

jonbob (approved):
Works great for our branch that is trying to use PETSc -- thanks for all your effort on this!

jgfouca:
Pushed to next.

PR #5789, opened 2023-07-02
add aerosol hygroscopicity (kappa) scaling to namelist

### LINKED ISSUE #5787: Expose kappa (hygroscopicity) to namelist parameters
kappa tunings are being actively considered in finalizing v3. It may be best to expose additional kappa factors (or kappa values themselves) as namelist parameters for easier, less error-prone, and more documented usage in the coming weeks. The relevant code is:

https://github.com/E3SM-Project/E3SM/blob/9e17b77a4e9ae91874b2fc56f248b5046bdce473/components/eam/src/physics/cam/modal_aer_opt.F90#L728-L809

An example implementation is to add a `hygro_factor_[species]` (default: 1) for each `species` of interest. For now, I am aware of only POM (p-organic) tuning.

```diff
                   if (trim(spectype) == 'p-organic') then 
                      do i = 1, ncol 
                         burdenpom(i) = burdenpom(i) + specmmr(i,k)*mass(i,k) 
                         scatpom(i)   = vol(i)*specrefr 
                         abspom(i)    = -vol(i)*specrefi 
-                        hygropom(i)  = vol(i)*hygro_aer 
+                        hygropom(i)  = vol(i)*hygro_aer*hygro_factor_pom
                       end do 
                   end if 
```

An alternative implementation is to redo the logic of using hygro_aer (i.e., reading from namelist rather than input files as is done now).

--

Copying @susburrows for awareness and opinion.

mahf708:
Hygroscopicity (kappa) tunings are being actively considered in finalizing v3. It may be best to expose additional kappa factors (or, alternatively, kappa values themselves) as namelist parameters for easier, less error-prone, and more transparently documented usage in the coming weeks. In this implementation, I added a namelist parameter `[species]_hygro_scale` (default: 1) for each `[species]` of interest.

Fix #5787

[BFB]
[NML]

### DIFF
--- a/components/eam/bld/build-namelist
+++ b/components/eam/bld/build-namelist
@@ -3810,6 +3810,15 @@ if ($chem =~ /_mam/) {
 #Sea Salt aerosol emission scale factors
 add_default($nl, 'seasalt_emis_scale');
 
+# Aerosol hygroscopicity (kappa) scaling
+add_default($nl, 'dust_hygro_scale');
+add_default($nl, 'sulfate_hygro_scale');
+add_default($nl, 'bc_hygro_scale');
+add_default($nl, 'pom_hygro_scale');
+add_default($nl, 'soa_hygro_scale');
+add_default($nl, 'seasalt_hygro_scale');
+add_default($nl, 'mom_hygro_scale');
+
 # Turbulent Mountain Stress
 add_default($nl, 'do_tms');
 my $do_tms = $nl->get_value('do_tms');
--- a/components/eam/bld/namelist_files/namelist_defaults_eam.xml
+++ b/components/eam/bld/namelist_files/namelist_defaults_eam.xml
@@ -294,6 +294,15 @@
 <seasalt_emis_scale chem="trop_strat_mam7">   1.62 </seasalt_emis_scale>
 <seasalt_emis_scale chem="trop_mam9">         1.62 </seasalt_emis_scale>
 
+<!-- Aerosol hygroscopicity (kappa) scaling -->
+<dust_hygro_scale>    1.0 </dust_hygro_scale>
+<sulfate_hygro_scale> 1.0 </sulfate_hygro_scale>
+<bc_hygro_scale>      1.0 </bc_hygro_scale>
+<pom_hygro_scale>     1.0 </pom_hygro_scale>
+<soa_hygro_scale>     1.0 </soa_hygro_scale>
+<seasalt_hygro_scale> 1.0 </seasalt_hygro_scale>
+<mom_hygro_scale>     1.0 </mom_hygro_scale>
+
 <!-- Modal optics calculations -->
 <mam3_mode1_file rad="rrtmg">atm/cam/physprops/mam3_mode1_rrtmg_c110318.nc</mam3_mode1_file>
 <mam3_mode2_file rad="rrtmg">atm/cam/physprops/mam3_mode2_rrtmg_c110318.nc</mam3_mode2_file>
--- a/components/eam/bld/namelist_files/namelist_definition.xml
+++ b/components/eam/bld/namelist_files/namelist_definition.xml
@@ -4914,6 +4914,48 @@ Seasalt aerosol emission tuning factor
 Default: set by build-namelist.
 </entry>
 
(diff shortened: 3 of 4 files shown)


susburrows:
I really like the idea of exposing this through the namelist so that it is easier to change, @mahf708 .  But, I would have a strong preference for the alternative implementation you suggest in https://github.com/E3SM-Project/E3SM/issues/5787 , i.e., redoing the logic of using hygro_aer so that it is read from namelist rather than from input files as is done now.

The problem with the scaling factor approach is that you need to have two pieces of information (the hygroscopicity parameter from the file, and the scaling factor) to reconstruct the hygroscopicity.

Can you modify the code to follow this approach?  And also set the default namelist parameter values to the values in the  aer_phys_props.nc file we are currently using to prescribe the hygroscopicities?

kaizhangpnl:
This PR would be useful for future sensitivity tests, but it needs substantial revision. 

Hygroscopicity affects not only the aerosol optical properties, but also the aerosol microphysics and activation. Therefore, unless it is designed to test the impact of an individual effect only, the scaling should be applied in rad_constituents.F90. Otherwise, modifications at (additional) multiple places are needed (wateruptake, activation in ndrop, activation in crm, activation in zm microphysics or activation in convective clouds). 

Also, the current changes in modal_aer_opt.F90 should only affect the optical properties diagnostics for individual species (e.g., AODSO4). The dominant factors in the optical property calculation (regarding hygroscopicity effect) are the wet radius and aerosol water content calculated in the aerosol wateruptake scheme. 

Finally, in hetfrz_classnuc_cam.F90 hygroscopicity values are hard-coded, so for consistency the scaling should also be applied here.

hwangacme:
> This PR would be useful for future sensitivity tests, but it needs substantial revision.
> 
> Hygroscopicity affects not only the aerosol optical properties, but also the aerosol microphysics and activation. Therefore, unless it is designed to test the impact of an individual effect only, the scaling should be applied in rad_constituents.F90. Otherwise, modifications at (additional) multiple places are needed (wateruptake, activation in ndrop, activation in crm, activation in zm microphysics or activation in convective clouds).
> 
> Also, the current changes in modal_aer_opt.F90 should only affect the optical properties diagnostics for individual species (e.g., AODSO4). The dominant factors in the optical property calculation (regarding hygroscopicity effect) are the wet radius and aerosol water content calculated in the aerosol wateruptake scheme.
> 
> Finally, in hetfrz_classnuc_cam.F90 hygroscopicity values are hard-coded, so for consistency the scaling should also be applied here.

susburrows:
@hwangacme is there a reason you want to close the PR?

hwangacme:
> Please keep open, thanks! @hwangacme, please be courteous and don't close PRs without explanation. We are working on supporting tasks needed for the current integration and you may be out of the loop. Members are already running simulations based on tunings of these factors, and it is best these factors are documented transparently and thoroughly, lest we rely on error-prone file edits.

@mahf708 Please see my response to @susburrows.  Sorry about the inconvenience. What do you mean to remove me from the reviewer list?

hwangacme:
@mahf708 @susburrows 
If this new task or PR was proposed to support the code integration, please keep me and @crterai in the loop. I thought we are the EPIC co-leads.

susburrows:
@hwangacme This is a draft PR for an idea that @mahf708 is working on as part of the aerosol working group epic. I am happy to talk more if you wish (we could also have a quick offline conversation later this week).

rljacob:
In general only 2D and 3D fields should be read from files because it is hard, but not impossible!, to do those via namelist.  But global scalars should definitely be in the namelist.

Also all cases used for production runs should be configurable "out-of-the-box" which for EAM means you get all the settings you need from the compset string and a use_case file.   The scripts should only be setting history file options, pe-layouts and other things that don't affect the answers produced.

susburrows:
It's been implemented this way (with the hygroscopicity parameters in input files) since before I started to work with the codebase, so for now than 11 years now. It would definitely be an improvement to expose these variables to the name list.

We should definitely keep this idea on the wishlist. We may want to revisit it.

Get Outlook for Android<https://aka.ms/AAb9ysg>

PR #5799, opened 2023-07-07
Set CO2 PPMV value for generic 2010 and 1950 compsets

wlin7:
The values currently set 2010 and 1950 depend on physics/chemistry modifiers in the compset longnames. A variant of 2010 or 1950 compsets, if not setting a value correspondingly, would pick up the default 379.0 PPMV, which is for an older present-day condition (around 2005).

[non-BFB] for F2010_chemUCI.* compsets.

----- Until this PR taking effect -----
v3alpha.F2010 cases can do `./xmlchange CCSM_CO2_PPMV="388.717"`  to achieve the equivalent.

### DIFF
--- a/driver-mct/cime_config/config_component_e3sm.xml
+++ b/driver-mct/cime_config/config_component_e3sm.xml
@@ -724,7 +724,9 @@
       <value compset="^2000.+AV1C-H01C"				>368.865</value>
       <value compset="^1850.+CMIP6"                             >284.317</value>
       <value compset="^1850.+CMIP6.+4xCO2"			>1137.268</value>
+      <value compset="^1950"                                    >312.821</value>
       <value compset="^1950.+CMIP6"                             >312.821</value>
+      <value compset="^2010"                                    >388.717</value>
       <value compset="^2010.+CMIP6"                             >388.717</value>
       <value compset="^SSP585.+CMIP6"				>0.000001</value>
       <value compset="^SSP370.+CMIP6"                           >0.000001</value>


wlin7:
@tangq , thanks for spotting the issue. I have done new v3alpha02.F2010 test to verify the only change in atm_in is `co2vmr `becomes `388.717000e-6` from `379.000000e-6`. Please take a look just in case. The value is also set for generic 1950 compsets in anticipation of continuing use in v3 and beyond.

tangq:
@wlin7 , I see the existing setting is only for "^2010.+CMIP6", but not for "^2010". The changes make sense and would probably affect the coupled tests and AWG tests. Thanks for fixing this.

tangq (approved):

wlin7:
Hi @mahf708 , @wagmanbe , please note that this will have NBFB effect for your v3atm based F2010 simulations.

wlin7:
Thanks for offering to help, @mahf708 . The changes are straightforward, but yes, please also take a look.
This amount to a 2.5% differences (more) in CO2 concentration, not sure if it would have a measurable impact on ERF estimate. Though I expect the impact to be small, it would be helpful to have an actual run.

wlin7:
> Also, a side question: Does this mean the value for 2000 is 399 and the value for 2010 is 389? Did CO2 actually go down between 2000 and 2010? If so, TIL. 😄

Just realized I got a typo in the commit message (corrected). The 2000 value is set to 379.0.

wlin7:
Thank again for the review, @tangq , @mahf708 .

The PR is merged to v3atm/eam/master_MAM5_wetaero_chemdyg, in anticipation of its use in ongoing sensitivity tests.  If any further thoughts, please continue to share here.

wlin7:
>  For a minute, I thought everything I knew about CO2 was wrong... 😅
> 
😄 That would be impossible, you be wrong, or CO2 goes down while business as usual.

susburrows:
Thanks for the heads-up @wlin7 .  I think it would be good for you to also cross-check with the HES group, e.g., @bpbond and @shanlinw .

PR #5800, opened 2023-07-07
Add on-node benchmarks to maint-1.0

amametjanov:
Also update PM-CPU srun `-c` option to a more dynamic `tasks_per_node`.

[BFB]


ndkeen (approved):
I tested with e3sm_prod on pm-cpu

Issue #5597, opened 2023-04-12
Invalid values detected with GNU in call to check_energy_chng()


ndkeen:
Adding check for invalid values for GNU DEBUG, I can run the following test:
`SMS_D_Ld2.ne30pg2_r05_EC30to60E2r2.BGCEXP_CNTL_CNPRDCTC_1850.pm-cpu_gnu.elm-bgcexp`
and see error below.

```
270: Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.
270:
270: Backtrace for this error:
270: #0  0x14cbd7bb1d4f in ???
270: #1  0x137a9cd in tphysbc
270:    at /global/cfs/cdirs/e3sm/ndk/repos/me16-apr11/components/eam/src/physics/cam/physpkg.F90:2704
270: #2  0x139fb85 in __physpkg_MOD_phys_run1._omp_fn.0
270:    at /global/cfs/cdirs/e3sm/ndk/repos/me16-apr11/components/eam/src/physics/cam/physpkg.F90:1122
270: #3  0x14cbd83d0295 in GOMP_parallel
270:    at ../../../cpe-gcc-11.2.0-202108140355.9bf1fd589a5c1/libgomp/parallel.c:178
270: #4  0x1391396 in __physpkg_MOD_phys_run1
270:    at /global/cfs/cdirs/e3sm/ndk/repos/me16-apr11/components/eam/src/physics/cam/physpkg.F90:1101
270: #5  0x637b22 in __cam_comp_MOD_cam_run1
270:    at /global/cfs/cdirs/e3sm/ndk/repos/me16-apr11/components/eam/src/control/cam_comp.F90:268
270: #6  0x6255f2 in __atm_comp_mct_MOD_atm_run_mct
270:    at /global/cfs/cdirs/e3sm/ndk/repos/me16-apr11/components/eam/src/cpl/atm_comp_mct.F90:600
270: #7  0x486eb4 in __component_mod_MOD_component_run
270:    at /global/cfs/cdirs/e3sm/ndk/repos/me16-apr11/driver-mct/main/component_mod.F90:728
270: #8  0x46bb8f in __cime_comp_mod_MOD_cime_run
270:    at /global/cfs/cdirs/e3sm/ndk/repos/me16-apr11/driver-mct/main/cime_comp_mod.F90:3110
270: #9  0x4845ba in cime_driver
[...]

PR #5801, opened 2023-07-11
Initialize arrays and scalars to signaling NaNs for Intel in debug mode

amametjanov:
Add `-init=snan,arrays` to Intel's debug-compile flags.
Also, initialize EAM's and ELM's uninitialized arrays, which were flagged by compiler's run-time checks.

[BFB]


amametjanov:
This passes `e3sm_integration` in baseline-comparison mode: https://my.cdash.org/viewTest.php?buildid=2369388 .

ndkeen:
Fixes https://github.com/E3SM-Project/E3SM/issues/5597

components/eam/src/physics/cam/cam_diagnostics.F90
```diff
@@ -1757,6 +1757,8 @@ subroutine diag_phys_writeout(state, psl)
 
 !! Boundary layer atmospheric stability, temperature, water vapor diagnostics
 
+    p_surf_t1 = 0._r8
+    p_surf_t2 = 0._r8
```
wlin7, line 1761: Nice catch, @amametjanov . In the same spirit, we probably should also initialize p_surf_q1 and p_surf_q2.  vertinterp calculations only loop over ncol. Just like p_surf_t1 and p_surf_t2, p_surf_q1 and p_surf_q2 could also be invoked in full array operation when certain output variable is specified.

components/eam/src/physics/cam/clubb_intr.F90
```diff
@@ -1833,6 +1833,7 @@ subroutine clubb_tend_cam( &
    !  At each CLUBB call, initialize mean momentum  and thermo CLUBB state
    !  from the CAM state
 
+   rvm = 0._r8
```
wlin7, line 1836: Certainly an initialization of it has no harm. I do wonder if an error was detected with the new debug compilation options. Only 1 to ncol in rvm are used in calculation, and they always have proper values to begin with (from state1%q). If an invalid is reported without the rvm initialization, it seems an over-kill.

components/eam/src/physics/cam/conv_water.F90
```diff
@@ -364,8 +365,17 @@ subroutine conv_water_4rad( state, pbuf,  conv_water_mode, totg_liq, totg_ice )
...
+      do i = 1, ncol
+         if (shr_infnan_isnan( fice(i,k) )) then
+            sh_cldliq(i,k) = 0._r8
+            sh_cldice(i,k) = 0._r8
+         else
+            sh_cldliq(i,k)=sh_icwmr(i,k)*(1-fice(i,k))*sh_frac(i,k)
+            sh_cldice(i,k)=sh_icwmr(i,k)*   fice(i,k) *sh_frac(i,k)
+         endif
+      end do
+   end do
```
wlin7, line 378: Need to monitor tests with COSP on, esp. with MMF, as icmwr has 0s for configurations that use CLUBB or SHOC.

components/eam/src/physics/cam/clubb_intr.F90
amametjanov: Array accesses are over `pverp` also in `rvm(1:ncol,1:pverp)`:
```
1944    do i=1,ncol   ! loop over columns
...
2047       do k=1,pverp
...
2060          rvm_in(k)     = real(rvm(i,pverp-k+1), kind = core_rknd)
```
This was needed for `SMS_D_Ln5.ne4_oQU240.F2010.chrysalis_intel.eam-clubb_sp`. Without this initialization, error is 
```
16: forrtl: error (182): floating invalid - possible uninitialized real/complex variable.
16: Image              PC                Routine            Line        Source
16: libpnetcdf.so.3.0  000015555101868C  for__signal_handl     Unknown  Unknown
16: libpthread-2.28.s  000015554E03DB20  Unknown               Unknown  Unknown
16: e3sm.exe           0000000002EECC7C  clubb_intr_mp_clu        2060  clubb_intr.F90
16: e3sm.exe           0000000003546686  physpkg_mp_tphysb        2712  physpkg.F90
16: e3sm.exe           000000000350D8CF  physpkg_mp_phys_r        1150  physpkg.F90
16: libiomp5.so        00001555515583F3  __kmp_invoke_micr     Unknown  Unknown
16: libiomp5.so        00001555514DC273  Unknown               Unknown  Unknown
```
and `rvm = 0._r8` was the minimal initialization to get past this issue.

components/eam/src/physics/cam/conv_water.F90
amametjanov: Please clarify.
1. If `fice` has a NaN, then `sh_cldliq` was also get assigned a NaN (and now getting a 0 instead).
2. If no NaNs in `fice`, then `sh_cldliq` was and will be getting a 0 whenever `sh_icwmr` has a 0.

components/eam/src/physics/cam/cam_diagnostics.F90
amametjanov: Thanks, done.

bishtgautam (approved):

components/eam/src/physics/cam/conv_water.F90
rljacob: @wlin7 please see above.

rljacob: discussion on telecon:  integration passed and that includes COSP and MMF so this is ready.

amametjanov:
Rebased onto today's master.
Fixed more uninitialized array references (new since July 10): in commit ba16f5bc2b7f251e62bffb848b4e898846fbb115 .
Re-checked with `e3sm_integration`: https://my.cdash.org/viewTest.php?buildid=2409146 .
Merged to next.

rljacob:
Noticed something odd while using these new debugging options with the MOAB coupler on chrysalis.

The code runs fine without debugging but with it on, we got this stack trace:
1256 12: e3sm.exe          00000000009CDB47 seq_map_mod_mp_se        639 seq_map_mod.F90
1257 12: e3sm.exe          00000000006C04E5 prep_ocn_mod_mp_p       2770 prep_ocn_mod.F90
1258 12: e3sm.exe          000000000047E86A cime_comp_mod_mp_       3014 cime_comp_mod.F90

line 639 in seq_map_mod.F90 is an executable statement but the other 2 files/lines are comments.  Removing -init=snan,arrays let the code run to completion.

amametjanov:
Likely that some scalars or arrays at that line are uninitialized. When I added `write` prior to such lines, I always got a `NaN` confirming a missed initilization.

Issue #5647, opened 2023-04-27
Add invalid check to GNU DEBUG builds `-ffpe-trap=zero,overflow,invalid`


ndkeen:
We already have `-ffpe-trap=zero,overflow`, but would be great to also add `invalid` which will stop if a NaN is encountered.
I've already been working toward this and we had several PR's that have helped.
There are still some outstanding issues:
https://github.com/E3SM-Project/E3SM/issues/5597

These issues fail with DEBUG even before adding invalid, so would need to fix before I can verify invalid is OK.
https://github.com/E3SM-Project/E3SM/issues/5539
https://github.com/E3SM-Project/E3SM/issues/5294
https://github.com/E3SM-Project/E3SM/issues/5636
https://github.com/E3SM-Project/E3SM/issues/5540

We could also considered adding `-fcheck=pointer` to GNU DEBUG flags in same PR.

ndkeen:
I can go back thru the above issues to see if any have been resolved now, but does it make sense to make this PR without first fixing all tests that fail with it?

ndkeen:
I see Az already has a PR to add this flag (as well as pointer check) https://github.com/E3SM-Project/E3SM/pull/5808

PR #5808, opened 2023-07-13
Add check for invalid fp-ops for GNU in debug mode

amametjanov:
Add check for invalid fp-ops for GNU in debug mode.
Also, fix ncol/pcol indexes in 2 eam-chemistry subroutines
(flagged by `-ffpe-trap=invalid`).

[non-BFB] - in fully-coupled tests


amametjanov:
`e3sm_integration` runs with
- `-ffpe-trap=invalid` and array-inits within components in E3SM-Project/E3SM#5801: https://my.cdash.org/viewTest.php?buildid=2370878
- just `-ffpe-trap=invalid`: https://my.cdash.org/viewTest.php?buildid=2370876

ndkeen:
I suppose this Fixes https://github.com/E3SM-Project/E3SM/issues/5647 in the sense that it adds the flags. But we still may have debugging to do for a few tests that might fail as result.

rljacob:
Waiting on #5801 and #5811

amametjanov (commented):
Rebased onto latest master and re-checked with `e3sm_integration`: https://my.cdash.org/viewTest.php?buildid=2411304 .

rljacob (approved):

Issue #5810, opened 2023-07-14
Update maint-2.0 branch on pm-cpu to allow intel builds


ndkeen:
NERSC recently installed intel as a module on pm-cpu, and we have this on current master as well as maint-1.0.  But the other branches do not yet have the compiler option. I have a branch to add it, as well as some basic module version updates to keep in-sync with master (and should not alter results).

One issue with using intel on this branch is that `externals/scorpio/src/clib/pio_lists.c` is missing
`#include "spio_io_summary.h"`
which causes a build fail with intel -- not sure the best way to get this change added to scorpio on maint-2.0

Should I go forward with PR to add intel, and then separately, we can fix the build issue?

ndkeen:
Closed after merge to amint-2.0 with PR #5815

PR #5815, opened 2023-07-18
For maint-2.0, add intel compiler to pm-cpu.

ndkeen:
Adds intel compiler to pm-cpu.
Update module versions on pm-cpu/pm-gpu to be same as current E3SM master.

Fixes https://github.com/E3SM-Project/E3SM/issues/5810

[bfb]


ndkeen:
I tested with `e3sm_prod`, running with intel compiler. And verified same test is BFB with GNU compiler.

ndkeen:
merged to maint-2.0

PR #5825, opened 2023-07-19
On pm-cpu, update one PE layout that impacts the ne120 test in the hires suite

ndkeen:
We've had a failing test in hires suite on pm-cpu since around June21st as there
was a problem with MPAS partition files.
`SMS.ne120pg2_r0125_oRRS18to6v3.F2010`
This PR Increases the partition count for both ocean and ice to use part files that exist. 
[bfb]


jonbob (approved):
Verified that the new graph file does exist on the inputdata repo, and approved based on developer testing

ndkeen:
merged to next

ndkeen:
Ah, I missed that the cdash test I was waiting on is actually testing master and not next. So it failed again same way. I will go ahead and merge this to master.

PR #5826, opened 2023-07-19
On pm-cpu and pm-gpu, update module version for craype and cray-mpich

ndkeen:
On pm-cpu and pm-gpu, update module version for craype and cray-mpich.
craype version 2.7.19 -> 2.7.20
cray-mpich version 8.1.24 -> 8.1.25

Proactive syncing with machine default versions.
Does not fix or cause any issue.
[bfb]

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -233,8 +233,8 @@
 
       <modules>
         <command name="load">craype-accel-host</command>
-        <command name="load">craype/2.7.19</command>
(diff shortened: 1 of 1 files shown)


ndkeen:
I tested with gnu e3sm_integration and compared with baseline. In the past, I've tested with these version using scream/pm-gpu and did not see any difference in performance or results. But happy to try any tests recommended.

ndkeen:
merged to next

ndkeen:
cdash reports are ok for pm-cpu with intel/gnu/nvidia compilers

PR #5827, opened 2023-07-20
Calculation of sedimentation fluxes in P3 is changed.

yunpengshan2014:
**Original calculation in the P3 code**: the precipitation (rain and snow) flux (sflx for snow ) is summed over the flux at each sub-timestep during one microphysics timestep: sflx=flux_qit_1+ flux_qit_2+ flux_qit_3+… flux_qit_n, where the flux_qit_i (i=1,2,3,…n) represents the snow flux at each sub-timestep for sedimentation. This should not be correct. The correct flux should be the mean value for each microphysics timestep. 

**Changed calculation in the current version**: In E3SM-P3, since the  sub-timestep is not a fixed value, the mean flux  over the microphysics timestep should be calculated with: sflx=(flux_qit_1*dt1+ flux_qit_2*dt2+ flux_qit_3*dt3+… flux_qit_n*dtn)/( dt1+ dt2+ dt3+… dtn), where dti represents sub-timestep length. This is similar to the equation used to calculate the surface precipitation rate.

This PR does not effect the model simulation results, but only change the diagnostic precipitation flux output.

### DIFF
--- a/components/eam/src/physics/p3/eam/micro_p3.F90
+++ b/components/eam/src/physics/p3/eam/micro_p3.F90
@@ -1604,7 +1604,9 @@ end function bfb_expm1
        call ice_sedimentation(kts,kte,ktop,kbot,kdir,    &
          rho(i,:),inv_rho(i,:),rhofaci(i,:),cld_frac_i(i,:),inv_dz(i,:),dt,inv_dt, &
          qi(i,:),qi_incld(i,:),ni(i,:),qm(i,:),qm_incld(i,:),bm(i,:),bm_incld(i,:),ni_incld(i,:), &
-         precip_ice_surf(i),sflx(i,:),p3_tend_out(i,:,40),p3_tend_out(i,:,41))
+!<shanyp 07192023
+         precip_ice_surf(i),precip_ice_flux(i,:),sflx(i,:),p3_tend_out(i,:,40),p3_tend_out(i,:,41))
+!shanyp 07192023>
 
 
        !*** Add tendencies for the next processes ***  
@@ -3898,6 +3900,9 @@ subroutine cloud_sedimentation(kts,kte,ktop,kbot,kdir,   &
    type(realptr), dimension(num_arrays) :: vs, fluxes, qnr
 
    real(rtype) :: dt_left
+!<shanyp 06252023
+   real(rtype) :: dt_sub
+!shanyp 06252023>
    real(rtype) :: prt_accum
    real(rtype) :: Co_max
    real(rtype) :: nu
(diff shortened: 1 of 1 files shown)


components/eam/src/physics/p3/eam/micro_p3.F90
```diff
@@ -3969,10 +3976,14 @@ subroutine cloud_sedimentation(kts,kte,ktop,kbot,kdir,   &
...
-
+!<shanyp 06252023
+                 prt_accum, inv_dz, inv_rho, rho, num_arrays, vs, fluxes, qnr, dt_sub)
+!shanyp 06252023>
             do k = k_qxbot,k_qxtop,kdir
-                  cflx(k+1) = cflx(k+1) + flux_qx(k)
+!<shanyp 06252023
+!                  cflx(k+1) = cflx(k+1) + flux_qx(k)
+                  cflx(k+1) = cflx(k+1) + flux_qx(k)*dt_sub
+!shanyp 06252023>
```
whannah1, line 3986: I know this is still a draft, but... remember to delete all these tags and commented lines!

components/eam/src/physics/p3/eam/micro_p3.F90
```diff
@@ -4132,12 +4154,18 @@ subroutine rain_sedimentation(kts,kte,ktop,kbot,kdir,
...
 
          !-- AaronDonahue, precip_liq_flux output
          do k = k_qxbot,k_qxtop,kdir
-            precip_liq_flux(k+1) = precip_liq_flux(k+1) + flux_qx(k) ! AaronDonahue
-            rflx(k+1) = rflx(k+1) + flux_qx(k)
+!<shanyp 06252023
+!            precip_liq_flux(k+1) = precip_liq_flux(k+1) + flux_qx(k) ! AaronDonahue
+!            rflx(k+1) = rflx(k+1) + flux_qx(k)
+            precip_liq_flux(k+1) = precip_liq_flux(k+1) + flux_qx(k)*dt_sub ! AaronDonahue
```
AaronDonahue, line 4166: You can delete my name from this line now that it is sort of a community changed line of code.

AaronDonahue:
In order to keep EAMxx and EAM BFB when running the same parameterizations we will need to make the same changes in the eamxx code.  See for example:
https://github.com/E3SM-Project/E3SM/blob/master/components/eamxx/src/physics/p3/impl/p3_upwind_impl.hpp#L97
https://github.com/E3SM-Project/E3SM/blob/master/components/eamxx/src/physics/p3/impl/p3_cloud_sed_impl.hpp#L17

rljacob:
@yunpengshan2014 please respond to above.

yunpengshan2014:
Hi Dr. Lin,

I will close this PR since this change will be implemented on the top of
the most recent master. It is okay to delete this PR.
This change is for the diagnostic output and will not affect any simulation
result.

Regards,
Yunpeng

PR #5829, opened 2023-07-21
New upwinded advection feature for split-explicit time integrator

cbegeman:
This PR adds the option to use upwinded advection with the split-explicit time integrator. Formerly, upwinded advection was only available with the RK4 time integrator.

For the barotropic subcycle's SSH predictor and corrector steps, the barotropic velocity is first computed on edges, then used to determine the water column thickness on edges (either centered or upwinded), then both barotropic velocity and water column thickness on edges is used to compute the SSH tendency on cells.

We also compute the normal velocity correction factor using `layerThickEdgeFlux`, which is updated if upwinded advection is used.

Upwinding of the water column thickness is done with a new diagnostic subroutine, `ocn_diagnostic_solve_wctEdge`, which is modeled after `ocn_diagnostic_solve_layerThicknessEdge`.

[non-BFB] (machine-precision diffs on some compiler, machine combinations, BFB on others)

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
@@ -673,13 +673,11 @@ contains
 !  routine ocn_diagnostic_solve_wctEdge
 !
 !> \brief   Computes layer thickness at edge
-!> \author  Matt Turner
-!> \date    October 2020
+!> \author  Carolyn Begeman
+!> \date    July 2023
 !> \details
-!>  This routine computes the diagnostic variables layerThickEdgeFlux
-!>  and layerThickEdgeMean. The Mean is a simple mean across the edge
-!>  but the Flux value can either be the mean or an upwind value,
-!>  depending on user input.
+!>  This routine computes the water column thickness on edges, wctEdge, which 
+!>  can either be the mean or an upwind value, depending on user input.
 !
 !-----------------------------------------------------------------------
 
@@ -704,7 +702,7 @@ contains
 
       ! Outputs are the shared variables from diagnostic_variables:
       real (kind=RKIND), dimension(:), intent(inout) :: &
-         wctEdge !< [out] wct on edge
+         wctEdge !< [out] water column thickness on edge
 
       !-----------------------------------------------------------------
       ! local variables
@@ -729,7 +727,9 @@ contains
 
          ! Compute mean layer thickness at edge
 #ifdef MPAS_OPENACC
-         !$acc parallel loop collapse(2) &
+         !$acc parallel loop &
+         !$acc    present(bottomDepthEdge, sshCell, sshEdge, cellsOnEdge, wctEdge) &
+         !$acc    private(cell1, cell2)
 #else
          !$omp parallel
          !$omp do schedule(runtime) private (cell1, cell2)
@@ -759,14 +759,14 @@ contains
          ! Use upwind thickness as the edge flux value
 #ifdef MPAS_OPENACC
          !$acc parallel loop &
-         !$acc    present(btrVelocity, sshCell, sshEdge, cellsOnEdge) &
+         !$acc    present(btrVelocity, sshCell, sshEdge, cellsOnEdge, maxLevelEdgeTop, wctCell, wctEdge) &
          !$acc    private(cell1, cell2)
 #else
          !$omp parallel
          !$omp do schedule(runtime) private(cell1, cell2)
 #endif
          do iEdge = 1, nEdgesAll
-            if ( maxLevelEdgeTop(iEdge).eq.0 ) cycle
+            if ( maxLevelEdgeTop(iEdge) == 0 ) cycle
             cell1 = cellsOnEdge(1, iEdge)
             cell2 = cellsOnEdge(2, iEdge)
             if (btrVelocity(iEdge) > 0.0_RKIND) then
--- a/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
@@ -1056,6 +1056,7 @@ contains
       !$omp do schedule(runtime) private(k, kmin, kmax, cell1, cell2)
 #endif
       do iEdge = 1, nEdgesAll
+         if ( maxLevelEdgeTop(iEdge).eq.0 ) cycle
          kmin = minLevelEdgeBot(iEdge)
          kmax = maxLevelEdgeTop(iEdge)
          cell1 = cellsOnEdge(1,iEdge)
@@ -1135,9 +1136,9 @@ contains
       !$omp do schedule(runtime) private(cell1, cell2)
 #endif
       do iEdge = 1, nEdgesAll
-         if ( maxLevelEdgeTop(iEdge) < 1 ) cycle
-         cell1 = cellsOnEdge(1,iEdge)
-         cell2 = cellsOnEdge(2,iEdge)
+         if ( maxLevelEdgeTop(iEdge).eq.0 ) cycle
+         cell1 = cellsOnEdge(1, iEdge)
+         cell2 = cellsOnEdge(2, iEdge)
          if (btrVelocity(iEdge) > 0.0_RKIND) then
             edgeVar(iEdge) = cellVar(cell1)
          elseif (btrVelocity(iEdge) < 0.0_RKIND) then
--- a/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
@@ -725,7 +725,6 @@ contains
       !-----------------------------------------------------------------
       ! Begin code
 
-      call mpas_timer_start("wctEdge")
       wctEdge(:) = -1.0e34_RKIND
       ! Compute edge flux based on option set on init
       select case (thickEdgeFluxChoice)
@@ -776,7 +775,6 @@ contains
                              MPAS_LOG_CRIT)
 
       end select
-      call mpas_timer_stop("wctEdge")
 
    !--------------------------------------------------------------------
 
--- a/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
@@ -675,7 +675,7 @@ contains
 
 !***********************************************************************
 !
-!  routine ocn_diagnostic_solve_sshEdge
+!  routine ocn_diagnostic_solve_wctEdge
 !
 !> \brief   Computes layer thickness at edge
 !> \author  Matt Turner
@@ -756,6 +756,7 @@ contains
 
       case (thickEdgeFluxUpwind)
          allocate(wctCell(nCellsAll))
+
          !$omp parallel
          !$omp do schedule(runtime)
          do iCell = 1, nCellsAll
--- a/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
@@ -4471,17 +4471,14 @@ contains
                  & 'Set config_disable_vel_hadv to true for a linear test case. ', &
                  MPAS_LOG_CRIT)
          end if
+      elseif (trim(config_thickness_flux_type) == 'upwind') then
+         thickEdgeFluxChoice = thickEdgeFluxUpwind
+         call mpas_log_write('Thickness flux option set to ' //&
+              & trim(config_thickness_flux_type))
       else
-         if (config_use_wetting_drying .and. &
-             (trim(config_thickness_flux_type) /= 'centered')) then
-            if ( trim(config_thickness_flux_type) == 'upwind') then
-               thickEdgeFluxChoice = thickEdgeFluxUpwind
-            end if
-         else
-            call mpas_log_write('Thickness flux option of ' //&
-                 & trim(config_thickness_flux_type) // &
-                 & ' is not known', MPAS_LOG_CRIT)
-         end if
+         call mpas_log_write('Thickness flux option of ' //&
+              & trim(config_thickness_flux_type) // &
+              & 'is not known', MPAS_LOG_CRIT)
       end if
 
       ! Initialize coefficient choice for computing land ice drag
--- a/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
@@ -57,6 +57,7 @@ module ocn_diagnostics
 
    public :: ocn_diagnostic_solve, &
              ocn_diagnostic_solve_layerThicknessEdge, &
+             ocn_diagnostic_solve_sshEdge, &
              ocn_relativeVorticity_circulation, &
              ocn_vert_transport_velocity_top, &
              ocn_fuperp, &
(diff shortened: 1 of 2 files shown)


cbegeman:
### Testing

PR suite for standalone MPAS-Ocean, comparison with baseline with feature off

- BFB on chrys with intel, openmpi
- non-BFB on compy with intel, impi and debug

Larger-than-roundoff diffs observed as expected when feature on across the PR suite.

cbegeman:
@philipwjones When you have a chance, can you check the openacc directives I added? I have not tested those. Let me know if I can help with that. Thanks!

philipwjones:
@cbegeman While the loop directives look to be ok, many of the arrays are not on the device so the assertion of present() in the loop directives will fail. The best way to take care of this is to add $acc enter data create(varname) just after you allocate on the host and $acc exit data delete(varname) just before you deallocate. And then you'll either have to make sure any loops (including the implicit loop with array syntax being used to initialize) containing those variables have the appropriatae openacc directives, or make sure $acc update host or update device are called to sychronize the two memory spaces.

As far as testing, I'm guessing you don't have access to Summit or Frontier, but you should be able to test on Perlmutter with gpus (pm-gpu) and use the nvidiagpu compiler option for E3SM tests. GPU results will not be bit-for-bit so it'll be a judgement call on the diffs. There are ways to verify work is actually being done on the device but it's kinda tedious and involves doing duplicate calculations on host and device and comparing...

xylar:
@cbegeman, this looks good to me so far.  

It seems like the way forward is to first address @philipwjones' OpenACC fixes and then run a short test on pm-gpu to make sure they're in good shape.  After that (or maybe in parallel with that if we're sure the rest of the code won't change), we will need to run a forward run.  I believe I heard discussion that that needs to be longer than 10 years at the All-hands meeting but I'll check with @jonbob at today's MPAS DevOps meeting.

It seems like the compass testing you've already done is sufficient there so what we need from here on is more in E3SM.

jonbob:
@cbegeman -- as soon as you have a chance to take on Phil's suggestions, I'll start longer testing runs

jonbob:
@philipwjones -- do you have much experience with the nvidiagpu compiler pm-gpu? We were doing some testing and I ran into problems with configurations that had more than one mpas component (including fully-coupled)  with a link error like:
````
nvlink error   : Multiple definition of '_mpas_timekeeping_21' in '../../mpas-framework/src/libocn.a:mpas_timekeeping.f90.o', first defined in 'CMakeFiles/e3sm.exe.dir/global/u2/j/jonbob/e3sm_0724/driver-mct/main/cime_comp_mod.F90.o'
````
I think @cbegeman can test with C-cases, but it does mean we really can't use that compiler

**Comment by cbegeman** [comment]
@philipwjones, @jonbob and I are finding state validation errors in the ocean component for C-cases with nvidiagpu on pm-gpu on master (specifically SMS_Ld5.T62_oQU120.CMPASO-NYF.pm-gpu_nvidiagpu at 522051b891). Let me know how you would like to proceed.

**Comment by philipwjones** [comment]
Looking into it...

**Comment by cbegeman** [comment]
Thanks, @philipwjones. What do you think would be the best route at this point? Add the openacc directives that I think need to be there without GPU testing or remove the openacc directives that I added that haven't been tested?

**Comment by philipwjones** [comment]
If you can add them, I can test on Summit/Frontier. The OpenACC code still works fine there.

**Comment by cbegeman** [comment]
@philipwjones Sounds like a plan. I'll ping you when I've pushed the changes.

**Comment by cbegeman** [comment]
@philipwjones I made some openacc fixes spanning these lines in the subroutine I added to ocn_diagnostics: https://github.com/E3SM-Project/E3SM/blob/c76c040249afd498a7373d483b1f64f40b5fbd77/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F#L729-L803

I removed the single set of openacc directives I formerly added to split-explicit. I found that the barotropic subcycle didn't have openacc directives so adding them to one loop didn't make sense to me.

 Let me know what you think. Thanks for reviewing and testing!

File: components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F [hunk 1]
```diff
@@ -667,6 +668,152 @@ subroutine ocn_relativeVorticity_circulation(relativeVorticity, circulation, nor
...
+            end if
+         end do
+#ifndef MPAS_OPENACC
+         !$omp end do
+         !$omp end parallel
+#endif
+
+#ifdef MPAS_OPENACC
+         !$acc update host(wctEdge)
+         !$acc exit data delete(wctCell, sshCell, btrVelocity, sshEdge, wctEdge)
```
philipwjones, line 802: as above, use copyout for wctEdge rather than update/delete

philipwjones (changes requested):
In addition to the inline comments. The code generates NaNs on Summit for standalone QU240 when the upwind option is selected, even with running only on the CPU (OpenACC turned off). Runs fine for the centered option.

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
```diff
@@ -667,6 +668,152 @@ subroutine ocn_relativeVorticity_circulation(relativeVorticity, circulation, nor
...
+            wctEdge(iEdge) = 0.5_RKIND*(sshCell(cell1) + sshCell(cell2)) &
+                             + bottomDepthEdge(iEdge)
+         end do
+#ifndef MPAS_OPENACC
+         !$omp end do
+         !$omp end parallel
+#endif
+
+#ifdef MPAS_OPENACC
+         !$acc update host(wctEdge)
```
philipwjones, line 750: use copyout instead of update/delete for wctEdge as in:

!$acc exit data delete(bottomDepthEdge, sshCell, sshEdge) &
!$acc                 copyout(wctEdge)

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
```diff
@@ -667,6 +668,152 @@ subroutine ocn_relativeVorticity_circulation(relativeVorticity, circulation, nor
...
+         !$acc exit data delete(bottomDepthEdge, sshCell, sshEdge, wctEdge)
+#endif
+
+      case (thickEdgeFluxUpwind)
+
+         allocate(wctCell(nCellsAll))
+#ifdef MPAS_OPENACC
+         !$acc enter data create(wctCell, sshCell, btrVelocity, sshEdge, wctEdge)
```
philipwjones, line 758: as above, use copyin instead of create here

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
```diff
@@ -667,6 +668,152 @@ subroutine ocn_relativeVorticity_circulation(relativeVorticity, circulation, nor
...
+
+      !-----------------------------------------------------------------
+      ! local variables
+      !-----------------------------------------------------------------
+
+      integer :: &
+         iCell, iEdge, &! edge and cell loop indices
+         cell1, cell2  ! neighbor cell indices across an edge
+
+      real (kind=RKIND), dimension(:), allocatable :: wctCell, sshEdge
```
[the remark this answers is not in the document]
cbegeman: Resolved by https://github.com/E3SM-Project/E3SM/pull/5829/commits/d613455d6d27e0ba1d0d2c59b87140ee8b896f19

cbegeman:
@philipwjones Thanks for doing that testing. The changes you requested are in this commit https://github.com/E3SM-Project/E3SM/pull/5829/commits/d613455d6d27e0ba1d0d2c59b87140ee8b896f19. 

That's interesting that the Summit case fails on CPU. Any thoughts on how to proceed? I don't have an account on Summit.

cbegeman:
@philipwjones Thanks for catching that double-initialization. And thanks for debugging! I am hoping that restructuring the commits into 4 independent sets of changes will help with debugging (e.g., git bisect).

cbegeman:
@philipwjones This is my best guess at what could be going wrong: https://github.com/E3SM-Project/E3SM/pull/5829/commits/286af9b058c6587bd47cc177280da277bf62f3e4. I haven't tested it in standalone but I don't anticipate it doing any harm.

philipwjones:
Unfortunately, didn't fix, but I'm getting close to the source...

philipwjones:
Turns out the init issue was a red herring and the funky values were because the simulation was blowing up with SSH going unstable. This was in a standalone QU240 case that I often run and I can reproduce on chrysalis. Are there any options incompatible with the thickness upwinding? Tried cutting the time step in half but still blew up...

xylar:
@philipwjones, maybe if you point us to your setup on Chrysalis, we can take a look?

cbegeman:
@philipwjones Ok. So for me with `ocean/global_ocean/QU240/WOA23/performance_test`
I get a successful 10-day run with `config_dt = 00:48:00` and failure at 8 days with `config_dt = 01:00:00` (both with `config_btr_dt = 00:06:00`).

Would you like to do further testing or would you like me to do further testing?

philipwjones:
Hmmm. For some reason, those values didn't work for me back on Summit - was hoping to at least verify the OpenACC version. I can run the centered scheme with a 2-hour step so that's a pretty significant cut. Seems like upwind schemes shouldn't be that different from the centered scheme, stability-wise? Just more diffusive? I guess I'm still not confident there isn't a bug in here somewhere and unless we understand what's causing this or what conditions trigger it, I'd be worried about using the option.  So yeah, I think it needs some more debugging/analysis and I'm happy to hand that back to the developer. Sorry...

cbegeman:
@xylar @mark-petersen @philipwjones This PR is ready for your review. I have located and fixed a bug in the upwinded `wctEdge` diagnostic routine; this routine now makes use of `bottomDepthEdge`. With the feature on, I was able to run the QU240 performance_test for 30 days and diffs across the PR suite appear reasonable. I verified that the PR suite is still BFB with the feature off on.

philipwjones (approved):
Thanks @cbegeman , this now runs for me too for my QU240 test. The OpenACC version also works with the expected roundoff-level diffs and with the expected improvement in performance on the device. So I'm satisfied the OpenACC changes are correct. The cost of the upwind option is small compared to centered in cpu mode - in the noise in my tests.  The changes in solution (upwind vs centered) were significant as you might expect so I'll have to let others evaluate that impact in longer sims.

xylar (approved):
The code looks excellent!  Thanks for the clean-up and the implementation that's easy to fillow.

I tested the `pr` suite on Perlmutter and it's BFB with a baseline using the current `main` with Gnu compilers.

I also ran 1 month of the `isomip_plus/planar/2km/z-star/Ocean0` test case with both `centered` and `upwind` layer thickness advection.  For the latter, I needed to reduce the time step from 4 minutes to 2 minutes (but 2 minute time steps is typical for global ocean runs at 2 km, so that's not necessarily concerning).  Results are visibly different after 1 month but not in a way that is concerning to me.  Once we have W&D as well, I'll want to do some longer tests.

I did notice that the CFL number is a lot noisier with upwind advection:
centered (dt = 4 min):
```
 CFLNumberGlobal = 0, 0.00456292326079603, 0.0211150106408386, 
    0.0422654220896013, 0.0599621536918491, 0.0809329801927618, 
    0.09462926524788, 0.125978048202252, 0.128036763744016, 
    0.153523540975022, 0.131874036965565, 0.146835251311241, 
    0.137219236098203, 0.13599176194601, 0.138613660790883, 
    0.138581076150152, 0.13900424852536, 0.134050446502824, 
    0.138644285780345, 0.132185181707716, 0.140097386558407, 
    0.136557856765166, 0.129964923132993, 0.143336951500246, 
    0.128479119574212, 0.127792400292759, 0.13364661602942, 
    0.129655825390533, 0.137682182704052, 0.139814389132589, 
    0.13671902747729, 0.125691433044828 ;
```
upwind (dt = 2 min):
```
 CFLNumberGlobal = 0, 0.106463519307648, 0.0768005077697927, 
    0.179321468889382, 0.637185510486071, 0.116437233815168, 
    0.125127394038676, 0.104055545695994, 0.095867342812596, 
    0.104295772219995, 0.0878645771983393, 0.0832829389194316, 
    0.074862516791091, 0.668217053440631, 0.115082305043635, 
    0.487647190345026, 0.357673300810123, 0.12011826995782, 
    0.107004684215448, 0.19821549236828, 0.0727128212457849, 
    0.122252492069522, 0.0944388248272034, 0.0864355771830894, 
    0.104275538526594, 0.0901230336315279, 0.111304496679046, 
    0.138557807696943, 0.140375908586772, 0.237704842767036, 
    0.0762886029010463, 0.0901907925464109 ;
```

It seems like this might be something to follow up on as part of the wetting and drying work.

centered:
botTemp_0001
upwind (notice an isolated high temperature cell at x~520 and y~10):
botTemp_0001

cbegeman:
> I also ran 1 month of the `isomip_plus/planar/2km/z-star/Ocean0` test case with both `centered` and `upwind` layer thickness advection. For the latter, I needed to reduce the time step from 4 minutes to 2 minutes (but 2 minute time steps is typical for global ocean runs at 2 km, so that's not necessarily concerning). Results are visibly different after 1 month but not in a way that is concerning to me. Once we have W&D as well, I'll want to do some longer tests.

@xylar Thanks for taking a close look at this. I reran the `isomip_plus/planar/2km/z-star/Ocean0`, halving the time step again to 1 min and found that the isolated temperature anomaly you noted is actually part of a more coherent velocity streak in that region. As we discussed on slack, it seems unlikely to indicate and issue with the upwind implementation, perhaps pressure gradient errors or the fact that the `ssh_adjustment` step is centered. We can investigate this outside the scope of this PR.
image

xylar:
@cbegeman, I'm going to rerun using upwinding for the ssh adjustment step, too, when I have a chance but that shouldn't impact this PR getting merged.

jonbob:
@cbegeman -- I assume this should go in as a stealth feature? If that's correct, I'll start a test and baseline so we can document its impact

cbegeman:
> @cbegeman -- I assume this should go in as a stealth feature? If that's correct, I'll start a test and baseline so we can document its impact

@jonbob Yes, it is stealth. Thanks for starting those tests!

jonbob:
I completed two 10-year low-res B-case runs, one a baseline and one with this feature on. The MPAS-Analysis output comparing them is online at:
[20230824.5829test.anvil](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20230824.5829test.anvil/html/)
Reviewers, could you please check this and make sure it's in line with your expectations?

cbegeman:
> I completed two 10-year low-res B-case runs, one a baseline and one with this feature on. The MPAS-Analysis output comparing them is online at:
[20230824.5829test.anvil](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20230824.5829test.anvil/html/)
Reviewers (and @cbegeman), could you please check this and make sure it's in line with your expectations?

@jonbob Thank you for running that case! Nothing about those results lead me to believe there is an issue with the implementation. We don't intend to use the first-order upwind option globally so I don't see anything troubling here. The diffs are large where currents are strong (e.g., Irminger Current https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20230824.5829test.anvil/html/ocean/index.html#&gid=6&pid=1) which makes sense.

xylar:
@jonbob and @cbegeman,

The analysis results look good to me.  Most look like ensemble-level variability.  The velocities look stronger and sharper (as in the plot @cbegeman pointed out), and that seems likely to be an improvement to me but also not the target of this work as @cbegeman pointed out.

I think we need a test for the feature, right @jonbob?  Hopefully, that's easy to add.

jonbob:
I added a new stealth test and testdef to support it, so I think this PR is good to go

cbegeman:
@jonbob Thanks a bunch for adding that test!

jonbob:
merged to master

PR #5838, opened 2023-07-25
bug fix in crm_physics.F90

whannah1:
Fix a bug  in which surface sensible and latent surface fluxes have been unintentionally swapped in the CRM interface. The impact of this fix will generally be small because these fluxes are currently only used to influence the internal CRM friction calculation. The application of suface fluxes in the PBL scheme outside of the CRM (in tphysac) are still correct, so there is no fundamental impact to the water mass or energy budgets. 

[non-BFB] for MMF cases.

### DIFF
--- a/components/eam/src/physics/crm/crm_physics.F90
+++ b/components/eam/src/physics/crm/crm_physics.F90
@@ -1090,8 +1090,8 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf2d, cam_in, cam_out, 
          ! Set surface flux variables
          !------------------------------------------------------------------------------------------
          if (phys_do_flux_avg()) then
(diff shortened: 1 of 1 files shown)


lee1046 (approved):
I checked SHFLX (LHFLX) is now coupled to the corresponding *_ptr. Everything looks good now.

whannah1:
To provide some evidence that the effect is small I ran some SCM cases with SAM and PAM (the new CRM that will replace SAM in the MMF soon) and some select results are shown in the plot below.

```
red  = SAM
blue = PAM

dashed = before fix
solid  = after fix
```

image

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #5842, opened 2023-07-27
fixes indexing bug in mpas cvmix interface

### LINKED ISSUE #5840: minLevelCell issue in mpas_ocn_vmix_cvmix.F
There is an array `Nsqr_iface` in the cvmix driver that is not initialized correctly for `minLevelCell .ne. 1`

https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-ocean/src/shared/mpas_ocn_vmix_cvmix.F#L474

The line referenced here should, I believe be

Nsqr_iface(1:minLevelCell(iCell)-1) = Nsqr_iface(minLevelCell(iCell))

instead.

Bug fix is not expected to change the climate.

vanroekel:
There is a small indexing error in the MPAS ocean interface to CVMix. It has only emerged in GPU configurations, but may emerge in cases where minLevelCell is not equal to 1

Fixes #5840

[BFB]

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_vmix_cvmix.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_vmix_cvmix.F
@@ -471,7 +471,7 @@ contains
               do k=minLevelCell(iCell),maxLevelCell(iCell)
                   Nsqr_iface(k) = BVFSmoothed(k)
               enddo
-              Nsqr_iface(minLevelCell(iCell):minLevelCell(iCell)-1) = Nsqr_iface(minLevelCell(iCell))
+              Nsqr_iface(1:minLevelCell(iCell)-1) = Nsqr_iface(minLevelCell(iCell))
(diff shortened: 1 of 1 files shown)


vanroekel:
@jonbob looking at the code this is BFB for coupled group runs but may not be when minLevelCell is not one. @xylar for the z* coordinates with ice cavities is minLevelCell still one?  I left the BFB label off as I wasn’t sure about the polar configs

xylar:
@vanroekel, so far no E3SM configurations use `minLevelCell /= 1` so the only tests that could be affected are z-level with ice shelf cavities in compass. You can mark this as BFB.

cbegeman (approved):
@vanroekel Yes, this seems to be a bug and your fix looks good. Let me know if you would like any help with testing.

xylar (approved):
I'm approving my inspection.

jonbob:
passes:

merged to next

jonbob:
merged to master

PR #5845, opened 2023-07-27
On pm-cpu, change default compiler to Intel

ndkeen:
For pm-cpu only, change the default compiler from GNU to Intel.
[bfb]


rljacob (approved):

ndkeen:
Yes, I was assuming that any "important" case would be setting the compiler, which would be BFB. And you are correct, if a case was dependent on the default compiler, we would not expect gnu/intel to be BFB. The reason for wanting to change the default compiler to Intel was that most of our previous production work was done on a machine with Intel compiler (so perhaps more tested, more expected) and it's ~10% faster. The reason we didn't start with Intel as default was just because NERSC only recently installed this compiler. So I first wanted to test with it and then change default.

ndkeen:
merged to next

PR #5850, opened 2023-08-01
Fixes for multiple issues in gustiness calculations.

### LINKED ISSUE #5834: Confusion between scalar and vector stress calculations
To explain the bug here I need to give some background.

# Background

For models like EAM that have a sub-grid-scale "gustiness" parameterization, it is necessary to distinguish between the mean magnitude of the surface stress (or "scalar" surface stress), and the mean stress as a vector that opposes the wind (or "vector" surface stress). If we use $\langle a \rangle$ to refer to an average of quantity "$`a`$" over a grid cell and a single time step, and if $\tau$ is the vector surface stress, then

$$\langle \left| \tau \right| \rangle \ge \left| \langle \tau \rangle \right|$$

where the left hand side is the scalar surface stress and the right hand side is the magnitude of the vector surface stress. These stresses are computed based on an average near-surface wind speed with gustiness included. If $u$ is a near-surface wind velocity, this can be expressed 

$$\langle \left| u \right|^2 \rangle = \left| \langle u \rangle \right|^2 + \langle \left| u - \langle u \rangle \right|^2 \rangle$$

which can be shown using the usual identity for the variance of a distribution $\langle(x-\langle x\rangle)^2\rangle=\langle x^2 \rangle-\langle x \rangle^2$. This is more commonly written using notation like this:

$$ U^2 = U_v^2 + U_g^2 $$

where $U$ is the root-mean-square wind speed over the grid cell, $U_v$ is the resolved wind speed (with respect to the surface), and $U_g$ is the "gustiness" wind speed, and each of the terms in this equation is defined to be equal to the one in the preceding equation (e.g. $U=\sqrt{\langle \left| u \right|^2 \rangle}$).

The difference between scalar and vector wind stresses is most apparent when $U_v$ is small and $U_g$ is large. In this case, the scalar wind stress will be large because on short length/time scales, wind gusts produce large stresses. However, the vector wind stress will be small, since a small $U_v$ means that there is no strongly preferred direction in which the wind blows, and so when averaged over the whole grid cell, the stress vectors at different locations largely cancel out.

For this reason, it is common to use the parameterization

$$ \left| \langle \tau \rangle \right| = \left(\frac{U_v}{U}\right) \langle \left| \tau \right| \rangle \qquad (*)$$

to relate the scalar and vector stresses. See for instance equations (3-4) of [Zeng et. al. (2002)](https://doi.org/10.1175/1520-0493(2002)130%3C2125:POWGFT%3E2.0.CO;2). This guarantees that in the limit $U_v \to 0$, the vector stress also becomes $0$, as it should.

# Nature of the bug

Currently, the atmosphere only exports gustiness-enhanced winds to the surface (i.e. it only supplies winds with total magnitude $U$). This means that none of the surface components can evaluate the right-hand side of (*), and so only the scalar stress can be calculated. However, the vector stress is the stress that should actually contribute to the next atmosphere (and ocean/sea ice) momentum budget. Thus, the wrong stress is currently being used.

# History of the bug

1. The gustiness code originally added to the ocean only, in #1035, does not appear to have had this issue. However, the evaluation of (*) is effectively "bundled in" with the decomposition of the stress into zonal and meridional components, so it is not obvious at first glance that the factor of $U_{v}/U$ is intentionally being applied here.
2. Pull request #1624 was intended to be a bug fix, but that PR actually introduces this bug to the ocean stress calculations. This appears to be an understandable mistake, because the code as written in the previous commit appears as if the effective multiplication by $U_v/U$ is accidental, rather than intentionally converting scalar to vector stress.
3. Pull requests #2679, #2720, and #2789 collectively replaced the gustiness parameterization in the ocean with one that applies to all components. In doing so, they produced the current behavior, where all surface components are now subject to the bug, and they do not have sufficient information about the gustiness to easily correct the formula.
4. Pull request #4229 introduced an XML option called `ATM_SUPPLIES_GUSTINESS`, which causes the atmosphere to supply $U_v$ and $U_g$ to the components (rather than $U$), and causes the surface flux calculations to perform the scalar-to-vector stress conversion. This PR was intended to solve this bug (and to address #4073), but the option that was added to do so has never been turned on by default.

# How to fix the bug

Make `ATM_SUPPLIES_GUSTINESS=TRUE` the default setting (eventually, it should probably be the only option). The issue described at #5835 should be fixed as well; it interacts with this bug because if `ATM_SUPPLIES_GUSTINESS=TRUE`, part of the fix must be applied in the surface components, not just in the atmosphere.

# Effect on model output

This bug fix reduces the effect of gustiness on wind speeds, and thus is likely to increase both the average wind speeds near the surface, and many fluxes (e.g. sensible and latent heat, dust, sea salt...). @mabrunke has performed short runs recently with the fix, but AFAIK we do not have a full climate evaluation yet. The gustiness parameters may need to be retuned if the effect of the fix is large; again, this should probably be done with #5835 also fixed.

### LINKED ISSUE #5835: Incorrect calculation of gustiness
In [Ma et. al. 2022](https://doi.org/10.5194/gmd-15-2881-2022), the gustiness parameterization is described in equation (1) as

$$U^2 = U_0^2 + a_g U_{g,(ZM)}^2 + b_g U_{g,(CLUBB)}^2$$

This appears correct and matches previous literature (aside from the addition of the tuning parameters $a_g$ and $b_g$). However, it is not what the EAM code currently does. A document summarizing the current code's behavior is [here](https://docs.google.com/document/d/1WbFOj82hV3lb6ALp9fZC3ZIDYURFt5bEZTdu3A2GwDo/edit?usp=sharing). In short, the relevant lines of code are [here](https://github.com/E3SM-Project/v3atm/blob/1f97213ce7ab8f3dcea486feb36ceba934cba27b/components/eam/src/physics/cam/clubb_intr.F90#L2978-L2989) and [here](https://github.com/E3SM-Project/v3atm/blob/0a642b74e230da58bc3abe9412f9aa3b8832dab6/components/eam/src/control/camsrfexch.F90#L768-L779). For EAM's current default behavior (no gustiness exported), the speed of the wind exported to the surface models is:

$$ U = U_0 + a_g U_{g,(ZM)} + b_g \left(\sqrt{U_{g,(CLUBB)}^2+U_0^2} - U_0\right) $$

Fixing this is a simple matter, though it is possible that a fix would affect the surface fluxes enough to require some retuning. Note that if the (non-default) ATM_SUPPLIES_GUSTINESS option is set in the XML files, this bug is still present, but part of the gustiness calculation is done in the surface components rather than in EAM.

This bug was noticed by @mabrunke, who has investigated it along with Xubin Zeng and myself.

### LINKED ISSUE #5836: EAM's 'U10' output includes a gustiness enhancement
Note: this bug is diagnostic-only; to the best of my understanding, a fix will result in bit-for-bit output, except for the U10 field. (The `u10` variables used in some aerosol code appear to be based on a completely separate calculation.)

Because the surface models only have access to a gustiness enhanced wind (see #5834), the value of 10-meter wind speed that they supply to the atmosphere model includes a contribution from gustiness. This means that it is unsuitable for comparison to other data sets that only include winds that are "resolved" or otherwise averaged over space and time scales greater than the effects of gustiness.

If #5834 is fixed, then this can be resolved by rescaling U10 in the surface models, namely by multiplying U10 by a factor of $U_v/U$, where $U_v$ is the resolved wind speed with respect to the surface, and $U$ is the gustiness-enhanced wind.

Optionally, we could also produce an output for the gustiness-enhanced wind (called something like `U10GUST`). This may be useful for comparing to data sets with resolution fine enough to resolve gustiness. However, doing so would likely require coupling (an) additional field(s) from the surface to atmosphere if this is intended to be an atmospheric output. That's because the resolved wind speed is with respect to the surface, and over the ocean, the surface has a non-zero velocity with respect to the earth (and this velocity is not currently known to the atmosphere).

If this is not changed, there needs to at least be clear documentation somewhere that U10 includes gustiness, and is not comparable to data products representing larger-scale wind averages.

### LINKED ISSUE #5837: Possible double-counting of boundary-layer gustiness contributions over land
Since EAMv2, a gustiness value calculated by CLUBB has been used to enhance the mean wind speeds; this is meant to represent the effects of the boundary-layer turbulence/convection accounted for by CLUBB. But the land model already calculates its own estimate for a convective gustiness contribution `wc`. E.g. see [here](https://github.com/E3SM-Project/v3atm/blob/1f97213ce7ab8f3dcea486feb36ceba934cba27b/components/elm/src/biogeophys/BareGroundFluxesMod.F90#L299-L306). This `wc` appears to be a parameterization accounting for the same convective sub-grid-scale motion that is handled by CLUBB, meaning that the same process is used to enhance the wind twice and gustiness is overestimated.

Assuming that this needs to be fixed, the `wc` enhancements should be removed from the land model (in each of the bare ground, lake, urban, and forest canopy modules). Or rather, a flag should be added to turn off the `wc` calculation, and the default behavior should be to not apply `wc`.

If the COARE or UA atmosphere-ocean coupling methods are ever intended to be used in the future, they should probably be modified as well, as both options also appear to apply a gustiness contribution similar to `wc`.

quantheory:
This is designed to fix the following issues:

 * #5834
 * #5835
 * #5836
 * #5837

The first of these was fixed by forcing ATM_SUPPLIES_GUSTINESS=TRUE
for all EAM cases, and the second was fixed purely by rewriting the
Fortran formulas. These two changes are climate changing, and will
generally tend to reduce the impact of gustiness on the mean winds,
i.e. mean wind speed will increase.

The third fix is to the U10 diagnostic only, and will decrease the
output U10. A value that has not been "fixed" in this way can be
obtained by outputting the new history variable 'U10WITHGUSTS' from
EAM.

The final fix is to remove some double-counting between gustiness
parameterizations in EAM and ELM. The fix can be "undone" by setting
`force_land_gustiness = .true.` in the ELM namelist.

Note that these fixes are only applied over land and ocean. At the
request of sea ice developers, this tag also disables gustiness over
sea ice completely.

Fixes #5834
Fixes #5835
Fixes #5836 
Fixes #5837 

[CC]
[NML]

### DIFF
--- a/components/elm/src/biogeophys/BareGroundFluxesMod.F90
+++ b/components/elm/src/biogeophys/BareGroundFluxesMod.F90
@@ -82,6 +82,7 @@ contains
     real(r8) :: zeta                             ! dimensionless height used in Monin-Obukhov theory
     real(r8) :: beta                             ! coefficient of convective velocity [-]
     real(r8) :: wc                               ! convective velocity [m/s]
+    real(r8) :: ugust_total(bounds%begp:bounds%endp) ! gustiness including convective velocity [m/s]
     real(r8) :: dth(bounds%begp:bounds%endp)     ! diff of virtual temp. between ref. height and surface
     real(r8) :: dthv                             ! diff of vir. poten. temp. between ref. height and surface
     real(r8) :: dqh(bounds%begp:bounds%endp)     ! diff of humidity between ref. height and surface
@@ -234,6 +235,7 @@ contains
             ur(p)    = max(1.0_r8,sqrt(forc_u(t)*forc_u(t)+forc_v(t)*forc_v(t)+ugust(t)*ugust(t)))
          end if
          tau_diff(p) = 1.e100_r8
+         ugust_total(p) = ugust(t)
 
          dth(p)   = thm(p)-t_grnd(c)
          dqh(p)   = forc_q(t) - qg(c)
@@ -269,7 +271,7 @@ contains
 
          call FrictionVelocity(begp, endp, fn, filterp, &
               displa(begp:endp), z0mg_patch(begp:endp), z0hg_patch(begp:endp), z0qg_patch(begp:endp), &
-              obu(begp:endp), iter, ur(begp:endp), um(begp:endp), ustar(begp:endp), &
+              obu(begp:endp), iter, ur(begp:endp), um(begp:endp), ugust_total(begp:endp), ustar(begp:endp), &
               temp1(begp:endp), temp2(begp:endp), temp12m(begp:endp), temp22m(begp:endp), fm(begp:endp), &
               frictionvel_vars)
 
@@ -302,6 +304,7 @@ contains
             else                                      !unstable
                zeta = max(-100._r8,min(zeta,-0.01_r8))
                wc = beta*(-grav*ustar(p)*thvstar*zii(c)/thv(c))**0.333_r8
+               ugust_total(p) = sqrt(ugust(t)**2 + wc**2)
                um(p) = sqrt(ur(p)*ur(p) + wc*wc)
             end if
             obu(p) = zldis(p)/zeta
--- a/components/elm/src/biogeophys/BareGroundFluxesMod.F90
+++ b/components/elm/src/biogeophys/BareGroundFluxesMod.F90
@@ -227,11 +227,11 @@ contains
          if (implicit_stress) then
             wind_speed0(p) = max(0.01_r8, hypot(forc_u(t), forc_v(t)))
             wind_speed_adj(p) = wind_speed0(p)
-            ur(p) = max(1.0_r8, wind_speed_adj(p) + ugust(t))
+            ur(p) = max(1.0_r8, sqrt(wind_speed_adj(p)**2 + ugust(t)**2))
 
             prev_tau(p) = tau_est(t)
          else
-            ur(p)    = max(1.0_r8,sqrt(forc_u(t)*forc_u(t)+forc_v(t)*forc_v(t)) + ugust(t))
+            ur(p)    = max(1.0_r8,sqrt(forc_u(t)*forc_u(t)+forc_v(t)*forc_v(t)+ugust(t)*ugust(t)))
          end if
          tau_diff(p) = 1.e100_r8
 
@@ -286,7 +286,7 @@ contains
                call shr_flux_update_stress(wind_speed0(p), wsresp(t), tau_est(t), &
                     tau(p), prev_tau(p), tau_diff(p), prev_tau_diff(p), &
                     wind_speed_adj(p))
-               ur(p) = max(1.0_r8, wind_speed_adj(p) + ugust(t))
+               ur(p) = max(1.0_r8, sqrt(wind_speed_adj(p)**2 + ugust(t)**2))
             end if
 
             tstar = temp1(p)*dth(p)
--- a/components/elm/src/biogeophys/BareGroundFluxesMod.F90
+++ b/components/elm/src/biogeophys/BareGroundFluxesMod.F90
@@ -48,7 +48,8 @@ contains
     use elm_varcon           , only : cpair, vkc, grav, denice, denh2o
     use elm_varctl           , only : iulog, use_lch4
     use landunit_varcon      , only : istsoil, istcrop
-    use FrictionVelocityMod  , only : FrictionVelocity, MoninObukIni, implicit_stress
+    use FrictionVelocityMod  , only : FrictionVelocity, MoninObukIni, &
+         implicit_stress, atm_gustiness, force_land_gustiness
     use QSatMod              , only : QSat
     use SurfaceResistanceMod , only : do_soilevap_beta
     use elm_time_manager     , only : get_nstep
@@ -303,9 +304,13 @@ contains
                um(p) = max(ur(p),0.1_r8)
             else                                      !unstable
                zeta = max(-100._r8,min(zeta,-0.01_r8))
-               wc = beta*(-grav*ustar(p)*thvstar*zii(c)/thv(c))**0.333_r8
-               ugust_total(p) = sqrt(ugust(t)**2 + wc**2)
-               um(p) = sqrt(ur(p)*ur(p) + wc*wc)
+               if ((.not. atm_gustiness) .or. force_land_gustiness) then
+                  wc = beta*(-grav*ustar(p)*thvstar*zii(c)/thv(c))**0.333_r8
+                  ugust_total(p) = sqrt(ugust(t)**2 + wc**2)
+                  um(p) = sqrt(ur(p)*ur(p) + wc*wc)
+               else
+                  um(p) = max(ur(p),0.1_r8)
+               end if
             end if
             obu(p) = zldis(p)/zeta
          end do
--- a/components/eam/bld/build-namelist
+++ b/components/eam/bld/build-namelist
@@ -3931,9 +3931,6 @@ if ( $linearize_pbl_winds =~ /$TRUE/io ) {
     }
 }
 
-# Export gustiness to surface components.
-add_default($nl, 'export_gustiness');
-
 # Use Convective water in radiation
 add_default($nl, 'conv_water_in_rad');
 
--- a/components/eam/bld/namelist_files/namelist_defaults_eam.xml
+++ b/components/eam/bld/namelist_files/namelist_defaults_eam.xml
@@ -1140,9 +1140,6 @@
 <!-- Linearization of boundary layer winds -->
 <linearize_pbl_winds>.false.</linearize_pbl_winds>
 
-<!-- Export gustiness value -->
-<export_gustiness>.false.</export_gustiness>
-
 <!-- MMF CRM mean state acceleration -->
 <use_crm_accel    >.false.</use_crm_accel>
 <crm_accel_uv     >.false.</crm_accel_uv>
--- a/components/eam/bld/namelist_files/namelist_definition.xml
+++ b/components/eam/bld/namelist_files/namelist_definition.xml
@@ -4066,19 +4066,6 @@ If .true., a linearization of the response of the pbl scheme to surface stresses
 is sent to the coupler every time step, using the fields 'wsresp' and 'tau_est'.
 </entry>
 
-<entry id="export_gustiness" type="logical" category="pbl"
-       group="phys_ctl_nl" valid_values="" >
-If .true., a gustiness value will be exported to the coupler, which allows
-surface components to distinguish between the effect of mean wind and gustiness
-when calculating fluxes (and in particular surface drag).
-
-If .false., no extra value is exported, and gustiness is instead applied by the
-atmosphere by scaling up the exported wind fields (ubot and vbot). This is
-simpler and requires slightly less coupling overhead, but produces numerical
-artifacts (oscillations) in surface stresses, particularly when the gustiness is
-large compared with the mean wind.
-</entry>
-
 
 <!-- Flags for implementing code modifications under polar project -->
 
--- a/components/eam/cime_config/buildnml
+++ b/components/eam/cime_config/buildnml
@@ -61,6 +61,11 @@ def buildnml(case, caseroot, compname):
 
     if not os.path.isdir(eamconf_dir): os.mkdir(eamconf_dir)
 
+    # Coupler and surface components now required to accept gustiness when EAM
+    # is used as the atmosphere model, unless SCREAM physics is used.
+    if not (atm_gustiness or "-DSCREAM" in cam_config_opts):
+        sys.exit("\n *** STOP: EAM requires ATM_SUPPLIES_GUSTINESS=TRUE. ***\n")
+
     #--------------------------------------------------------------------
     # Invoke eam configure - output will go in $CASEBUILD/eamconf
     #--------------------------------------------------------------------
@@ -191,7 +196,6 @@ def buildnml(case, caseroot, compname):
         if cam_branch_file: infile_text += " cam_branch_file = {} \n".format(cam_branch_file)
         if debug:           infile_text += " state_debug_checks = .true. \n"
         if linearize_pbl_winds: infile_text += " linearize_pbl_winds = .true. \n"
-        if atm_gustiness: infile_text += " export_gustiness = .true. \n"
 
         create_namelist_infile(case,
                                "{}/user_nl_eam{}".format(caseroot, inst_string),
--- a/components/eam/cime_config/testdefs/testmods_dirs/eam/implicit_stress/shell_commands
+++ b/components/eam/cime_config/testdefs/testmods_dirs/eam/implicit_stress/shell_commands
@@ -1,2 +1,1 @@
-./xmlchange ATM_SUPPLIES_GUSTINESS="TRUE"
 ./xmlchange ATM_FLUX_INTEGRATION_METHOD="implicit_stress"
--- a/components/eam/src/control/camsrfexch.F90
+++ b/components/eam/src/control/camsrfexch.F90
@@ -104,6 +104,7 @@ module camsrfexch
      real(r8), allocatable :: tref(:)       ! ref height surface air temp
      real(r8), allocatable :: qref(:)       ! ref height specific humidity 
      real(r8), allocatable :: u10(:)        ! 10m wind speed
+     real(r8), allocatable :: u10withgusts(:) ! 10m wind speed with gustiness
      real(r8), allocatable :: ts(:)         ! merged surface temp 
      real(r8), allocatable :: sst(:)        ! sea surface temp
      real(r8), allocatable :: snowhland(:)  ! snow depth (liquid water equivalent) over land
@@ -219,6 +220,9 @@ CONTAINS
        allocate (cam_in(c)%u10(pcols), stat=ierror)
        if ( ierror /= 0 ) call endrun('HUB2ATM_ALLOC error: allocation error u10')
 
+       allocate (cam_in(c)%u10withgusts(pcols), stat=ierror)
+       if ( ierror /= 0 ) call endrun('HUB2ATM_ALLOC error: allocation error u10withgusts')
+
        allocate (cam_in(c)%ts(pcols), stat=ierror)
        if ( ierror /= 0 ) call endrun('HUB2ATM_ALLOC error: allocation error ts')
 
@@ -308,6 +312,7 @@ CONTAINS
        cam_in(c)%tref     (:) = 0._r8
        cam_in(c)%qref     (:) = 0._r8
        cam_in(c)%u10      (:) = 0._r8
+       cam_in(c)%u10withgusts(:) = 0._r8
        cam_in(c)%ts       (:) = 0._r8
        cam_in(c)%sst      (:) = 0._r8
        cam_in(c)%snowhland(:) = 0._r8
@@ -617,6 +622,7 @@ CONTAINS
           deallocate(cam_in(c)%tref)
           deallocate(cam_in(c)%qref)
           deallocate(cam_in(c)%u10)
+          deallocate(cam_in(c)%u10withgusts)
           deallocate(cam_in(c)%ts)
           deallocate(cam_in(c)%sst)
           deallocate(cam_in(c)%snowhland)
--- a/components/eam/src/control/camsrfexch.F90
+++ b/components/eam/src/control/camsrfexch.F90
@@ -713,7 +713,6 @@ subroutine cam_export(state,cam_out,pbuf)
    integer :: vmag_gust_idx, wsresp_idx, tau_est_idx
    real(r8) :: umb(pcols), vmb(pcols),vmag(pcols)
    logical :: linearize_pbl_winds ! Send wsresp and tau_est to coupler.
-   logical :: export_gustiness ! Send vmag_gust to coupler
 
    real(r8), pointer :: prec_dp(:)                 ! total precipitation   from ZM convection
    real(r8), pointer :: snow_dp(:)                 ! snow from ZM   convection
@@ -732,8 +731,7 @@ subroutine cam_export(state,cam_out,pbuf)
    lchnk = state%lchnk
    ncol  = state%ncol
 
-   call phys_getopts(linearize_pbl_winds_out=linearize_pbl_winds, &
-                     export_gustiness_out=export_gustiness)
+   call phys_getopts(linearize_pbl_winds_out=linearize_pbl_winds)
 
    prec_dp_idx = pbuf_get_index('PREC_DP')
    snow_dp_idx = pbuf_get_index('SNOW_DP')
@@ -765,18 +763,9 @@ subroutine cam_export(state,cam_out,pbuf)
 !PMA adds gustiness to surface scheme c20181128
 
    do i=1,ncol
-      if (export_gustiness) then
-         cam_out%ubot(i)  = state%u(i,pver)
-         cam_out%vbot(i)  = state%v(i,pver)
-         cam_out%ugust(i) = vmag_gust(i)
-      else
-         ! If not exporting gustiness as a separate field, we apply it here.
-         umb(i)           = state%u(i,pver)
-         vmb(i)           = state%v(i,pver)
-         vmag(i)          = max(1.e-5_r8,sqrt( umb(i)**2._r8 + vmb(i)**2._r8))
-         cam_out%ubot(i)  = state%u(i,pver) * ((vmag_gust(i)+vmag(i))/vmag(i))
-         cam_out%vbot(i)  = state%v(i,pver) * ((vmag_gust(i)+vmag(i))/vmag(i))
-      end if
+      cam_out%ubot(i)  = state%u(i,pver)
+      cam_out%vbot(i)  = state%v(i,pver)
+      cam_out%ugust(i) = vmag_gust(i)
       cam_out%tbot(i)  = state%t(i,pver)
       cam_out%thbot(i) = state%t(i,pver) * state%exner(i,pver)
       cam_out%zbot(i)  = state%zm(i,pver)
--- a/components/eam/src/cpl/atm_import_export.F90
+++ b/components/eam/src/cpl/atm_import_export.F90
@@ -254,11 +254,10 @@ contains
     integer :: avsize, avnat
     integer :: i,m,c,n,ig       ! indices
     integer :: ncols            ! Number of columns
-    logical :: linearize_pbl_winds, export_gustiness
+    logical :: linearize_pbl_winds
     !-----------------------------------------------------------------------
 
-    call phys_getopts(linearize_pbl_winds_out=linearize_pbl_winds, &
-                      export_gustiness_out=export_gustiness)
+    call phys_getopts(linearize_pbl_winds_out=linearize_pbl_winds)
 
     ! Copy from component arrays into chunk array data structure
     ! Rearrange data from chunk structure into lat-lon buffer and subsequently
@@ -276,7 +275,9 @@ contains
              a2x(index_a2x_Sa_wsresp ,ig) = cam_out(c)%wsresp(i)
              a2x(index_a2x_Sa_tau_est,ig) = cam_out(c)%tau_est(i)
           end if
-          if (export_gustiness) then
+          ! This check is only for SCREAMv0; otherwise gustiness should always
+          ! be exported.
+          if (index_a2x_Sa_ugust /= 0) then
              a2x(index_a2x_Sa_ugust  ,ig) = cam_out(c)%ugust(i)
           end if
           a2x(index_a2x_Sa_tbot   ,ig) = cam_out(c)%tbot(i)   
--- a/components/eam/src/cpl/atm_import_export.F90
+++ b/components/eam/src/cpl/atm_import_export.F90
@@ -98,6 +98,7 @@ contains
           cam_in(c)%tref(i)      =  x2a(index_x2a_Sx_tref,  ig)  
           cam_in(c)%qref(i)      =  x2a(index_x2a_Sx_qref,  ig)
           cam_in(c)%u10(i)       =  x2a(index_x2a_Sx_u10,   ig)
+          cam_in(c)%u10withgusts(i) = x2a(index_x2a_Sx_u10withgusts, ig)
           cam_in(c)%icefrac(i)   =  x2a(index_x2a_Sf_ifrac, ig)  
           cam_in(c)%ocnfrac(i)   =  x2a(index_x2a_Sf_ofrac, ig)
           cam_in(c)%landfrac(i)  =  x2a(index_x2a_Sf_lfrac, ig)
--- a/components/eam/src/cpl/cam_cpl_indices.F90
+++ b/components/eam/src/cpl/cam_cpl_indices.F90
@@ -86,6 +86,7 @@ module cam_cpl_indices
   integer :: index_x2a_So_ssq          ! surface saturation specific humidity in ocean 
   integer :: index_x2a_Sl_ddvel        ! dry deposition velocities from land
   integer :: index_x2a_Sx_u10          ! 10m wind
+  integer :: index_x2a_Sx_u10withgusts ! 10m wind with gusts
 
 contains
 
@@ -122,6 +123,7 @@ contains
     index_x2a_Sf_lfrac      = mct_avect_indexra(x2a,'Sf_lfrac')
 
     index_x2a_Sx_u10        = mct_avect_indexra(x2a,'Sx_u10')
+    index_x2a_Sx_u10withgusts = mct_avect_indexra(x2a,'Sx_u10withgusts')
     index_x2a_Faxx_taux     = mct_avect_indexra(x2a,'Faxx_taux')
     index_x2a_Faxx_tauy     = mct_avect_indexra(x2a,'Faxx_tauy')
     index_x2a_Faxx_lat      = mct_avect_indexra(x2a,'Faxx_lat')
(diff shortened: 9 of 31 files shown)


quantheory:
Reserving first comment for discussion of testing (will update momentarily).

quantheory:
@mabrunke Please double-check this. Particularly `clubb_intr.F90`, the ELM `BareGroundFluxesMod.F90` and similar files, and `shr_flux_mod` changes. Also, I made the U10 change in `cam_diagnostics`.

@huiwanpnnl I made a change to the `conditional_diag_main` module, so that the U10 calculated there would match the one in the `cam_diagnostics` module.

quantheory:
@bishtgautam The atmospheric gustiness should be handled correctly in `lnd_import_export.F90`, but the corresponding changes have not been made in `lnd_downscale_atm_forcing.F90` (or `lnd_disagg_forc.F90`, which seems to be unused code). Can you let me know whether this is supposed to be a purely offline downscaling for use with `datm`, or if this is supposed to work in an "online" manner when EAM is running? If the latter, we need to change that module as well.

quantheory:
@wlin7 @rljacob Tagging you in case you also want to take a look at this.

mabrunke:
@quantheory, I have some code changes already for `lnd_import_export.F90`.  That is used no matter whether ELM is used offline or coupled to the atmosphere.

components/elm/src/biogeophys/BareGroundFluxesMod.F90
```diff
@@ -301,8 +302,12 @@ subroutine BareGroundFluxes(bounds, num_nolakeurbanp, filter_nolakeurbanp, &
...
             else                                      !unstable
                zeta = max(-100._r8,min(zeta,-0.01_r8))
-               wc = beta*(-grav*ustar(p)*thvstar*zii(c)/thv(c))**0.333_r8
-               um(p) = sqrt(ur(p)*ur(p) + wc*wc)
+               if ((.not. atm_gustiness) .or. force_land_gustiness) then
+                  wc = beta*(-grav*ustar(p)*thvstar*zii(c)/thv(c))**0.333_r8
+                  um(p) = sqrt(ur(p)*ur(p) + wc*wc)
+               else
+                  um(p) = max(ur(p),0.1_r8)
+               end if
```
bishtgautam, line 310: Should a check be added in during ELM's initialization to ensure that `atm_gustiness` and `force_land_gustiness` can not be both `.true.`?

quantheory: If by "offline" ELM you mean `I` cases or other cases without an active atmosphere, in those cases `atm_gustiness` is `.false.`, and so this code change has no effect (it takes the branch where `wc` is calculated).

The `force_land_gustiness` flag *only* has an effect if `atm_gustiness` is `.true.`. So if we do not allow them both to be `.true.`, `force_land_gustiness` should be removed entirely since it has no purpose.

quantheory:
@mabrunke The specific reason why I'm concerned about `lnd_import_export.F90` is [these lines of code](https://github.com/E3SM-Project/E3SM/blob/18198fbbc538287f606501a933382a7be976db99/components/elm/src/cpl/lnd_import_export.F90#L1121-L1126), which explicitly prohibit combining downscaling and `atm_gustiness`. I don't remember why that check is there. It may be that I added it because I just wasn't sure what to do with the gustiness field if downscaling is being used. Though perhaps it's as simple as just copying what's being done for `ubot` and `vbot`.

quantheory:
@mabrunke I guess what I'm trying to figure out is that we have [these](https://github.com/E3SM-Project/E3SM/blob/0d82039e608010e851af380698779cc42fae2652/components/elm/src/cpl/lnd_downscale_atm_forcing.F90#L225-L226) [three](https://github.com/E3SM-Project/E3SM/blob/0d82039e608010e851af380698779cc42fae2652/components/elm/src/cpl/lnd_downscale_atm_forcing.F90#L257-L258) [places](https://github.com/E3SM-Project/E3SM/blob/0d82039e608010e851af380698779cc42fae2652/components/elm/src/cpl/lnd_downscale_atm_forcing.F90#L342-L343) where that copying seems to be being done, and I'm wondering if I need to do the same thing for `ugust` as is being done for `ubot` and `vbot` in those lines. (As opposed to the current behavior, where if `atm_gustiness` is on, the code simply aborts before calling that function!)

components/elm/src/biogeophys/BareGroundFluxesMod.F90
bishtgautam: Thanks @quantheory for your explanation.

mabrunke:
@quantheory, can you provide the line of code where that abort happens.  I'm not seeing it.  Is it in `lnd_import_export.F90`?

quantheory:
@mabrunke Yeah, right here a few lines before the call to the downscaling would otherwise happen:

https://github.com/E3SM-Project/E3SM/blob/18198fbbc538287f606501a933382a7be976db99/components/elm/src/cpl/lnd_import_export.F90#L1121-L1123

mabrunke:
@quantheory Ah, I see it now.  I don't see why applying gustiness if there is downscaling would be a problem at this point, since wind speed is being downscaled right now.

quantheory:
@mabrunke OK, I can add another commit for to allow applying gustiness with the downscaling. Do you think this PR looks good other than that?

mabrunke:
@quantheory A few things:  I like your idea of adjusting U10 in EAM when its being outputted there, but U10 is also outputted in ELM.  It should be adjusted for that output as well.  Also, you changed the surface flux calculation in CICE, but isn't MPAS-SeaIce now being used in v2/3?

quantheory:
@mabrunke Those are both very good points. Regarding the U10 output, if I need to calculate the change in multiple components anyway, now I'm kind of considering going back to calculating it at the surface rather than in EAM. (U10 is supposed to be measured relative to the surface right? I.e. a measure of u_atm - u_ocn over the ocean, rather than just the atmosphere speed in an Eulerian frame? That seems to be how it works in `shr_flux_mod.F90`.)

It shouldn't be difficult to make the changes to MPAS; I really just forgot because the original ATM_SUPPLIES_GUSTINESS changes were done back in v1, and so my list of files to change had CICE but not MPAS ice on it. Let me see what I can work out today.

mabrunke:
@quantheory, truly the fluxes should be calculated using u_atm - u_sfc as well as the temperature and humidity differences, but, over land, u_sfc = 0.  That is not the case over the ocean, as there is a surface current.

I was handling the adjustment in the surface models.  That was why I suggested making ATM_SUPPLIES_GUSTINESS = TRUE.  That would complicate things if you want to save the two different versions of U10 like you are now.  I'm not sure yet if that's a good idea, since the old way of calculating U_gust was wrong anyways.

There is also a sea ice formulation in `shr_flux_mod.F90`.  I'm guessing that that is a relic of an older version of the model and is no longer used.

proteanplanet:
Why is CICE being discussed in this thread? E3SM does not use CICE.  But any changes to U10 will also affect ocean and sea ice coupling in B-cases if it is being changed in the coupling, rather than just output, and if such change are being proposed, people from the ice and ocean groups need to be assigned as reviewers.  If the changes only affect F-cases, then MPAS-SeaIce is still involved, and the CICE tag should be removed, and in its place sea ice or MPAS-SeaIce should be tagged.

quantheory:
@proteanplanet

> Why is CICE being discussed in this thread? E3SM does not use CICE.

Because it's building on changes that were originally developed in E3SMv1, and because this is the first PR I've made affecting sea ice in several years (partly due to not working directly on E3SM at all for a while). So as I mentioned earlier, I forgot that I needed to port the changes over to MPAS.

> people from the ice and ocean groups need to be assigned as reviewers

I agree, but being rather atmosphere-focused, I don't actually know who to ask to review it. Also, given the problems mentioned in the above discussion, I need to make some changes before getting anyone to review this again anyway.

I was planning to ask @wlin7, but if there's anyone you want to suggest someone as a reviewer (or if you feel you should do it yourself), feel free. Though again, I need to port the changes to MPAS before it's worth bothering.

> sea ice or MPAS-SeaIce should be tagged

There is no generic sea ice tag. The tag list seems a bit inconsistent in this regard, honestly. I had no idea whether to tag mpas-ocean or not, because I'm not touching the MPAS-ocean code at all, but I am making changes to coupler code that calculates atmosphere-ocean interactions. But there's no generic "Ocean" tag, only "mpas-ocean". My hope was that tagging Coupled Model would be enough.

proteanplanet:
@quantheory I suggest either making me or @eclare108213 a reviewer.  Thanks for adding the mpas-seaice tag.

mabrunke:
@quantheory, the surface fluxes are calculated in `column/ice_atmo.F90` in MPAS-SeaIce.  It doesn't use `ugust`, so it must expect to receive the gustiness in `vmag`.

quantheory:
@mabrunke Yeah, I noticed earlier, it looks like no one touched anything related to the gustiness export code since I first added it, so I have to port it from CICE today. Luckily this code has shared ancestry with the CICE version, so it should be easy to copy over once I've read through the MPAS version.

proteanplanet:
@quantheory, if you intend this work to make it onto E3SM master, please add a sea ice modeler to the review as per the above message - choose any of @eclare108213, @darincomeau, @njeffery, @erinethomas or me.  If this is for your own project and you are using it solely for that, that's fine. If you intend this work to enter E3SM master, the changes must first be made to MPAS-SeaIce, not CICE. Moreover, the changes should be made to Icepack not to the MPAS-SeaIce column package, because Icepack from the CICE Consortium is replacing that for E3SM V3.   Therefore your changes would need to pass review in the CICE Consortium Icepack code at https://github.com/CICE-Consortium/Icepack.  I am happy to help in testing the code in E3SM, and welcome improvements, but we will need to make sure it is thoroughly tested in D-, G- and B-cases first.

quantheory:
@proteanplanet As I mentioned before, please wait for additional commits before reviewing this (I will tag you when this is done). I did not add you as a reviewer before, because I didn't think it was worth your time to look at the code when several significant changes were about to be made anyway. I have been out of the office part of this week, so I haven't been very fast about this.

I think there is some misunderstanding about why I made this PR. My expectation is that once this PR is reviewed for code correctness, it will be referred to the coupled model group and relevant component leads to determine how to evaluate this using coupled model runs, and determine what to do from there. No one involved has any intention of trying to get this merged unless and until that process is complete. I tagged this "Do Not Merge" for a reason.

Furthermore, I am not a part of the E3SM project, and it is not my responsibility (or in fact, within my ability) to do all of the necessary climate evaluation, or the integration with v3 features. I put together these bug fixes because I believe we all share an interest in ensuring that the model code is a correct implementation of the physical equations it represents, and as a courtesy to other E3SM developers to not have to start from scratch when implementing these features for v3. I have neither the knowledge nor the time to shepherd this PR through merging to master and/or integration with v3 by myself. That will require the assistance of E3SM developers in the coupled model and other groups and the help of the integrators, so rest assured that you will hear much more about this, and there will be several more rounds of review, before anyone tries to get this merged.

proteanplanet:
@quantheory Thanks. I'll help with the testing when you are ready.

eclare108213:
> My expectation is that once this PR is reviewed for code correctness, it will be referred to the coupled model group and relevant component leads to determine how to evaluate this using coupled model runs, and determine what to do from there. No one involved has any intention of trying to get this merged unless and until that process is complete. I tagged this "Do Not Merge" for a reason.

@quantheory Thanks for reaching out.  I'm glad you're planning to work with the coupled group and the relevant component leads, since the implementation needs to be consistent across all of them and sometimes that can be pretty tricky.  Please change this to a Draft PR to make it clear that it's only for review purposes, not staged for merging into E3SM master.  To get this all the way through the review process into E3SM, you'll need to follow the steps outlined [here](https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/29754189/Code+Review+Process+Implementation).  Do you have a design document for these changes?

quantheory:
@eclare108213

> Please change this to a Draft PR to make it clear that it's only for review purposes, not staged for merging into E3SM master.

Ah, interesting! I don't think I've used that feature before. Done.

> Do you have a design document for these changes?

Yes and no. The original gustiness feature, developed over a few years and completed in 2019, seems to have a design document [here](https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/960233789/W1+-+Gustiness+Design+Document). This PR is a collection of bug fixes for that older feature, not a new feature in itself.

Part of the problem is that the original implementation tried to handle gustiness entirely within the atmosphere, without modifying the land or sea ice components or sending any "new" information through the coupler. (I assume this was in part to reduce the complexity of the required code modifications.) This approach neglects some changes that really need to be made to the other components. Hence, fixing this feature requires reworking the design to add/change a few lines of code in the land and sea ice models, even though the original feature did not touch the ELM or CICE (or MPAS-Sea Ice) code directly.

quantheory:
@proteanplanet @eclare108213 The changes to MPAS-Seaice are now implemented. (I also kept the CICE changes because CICE is still used for some tests, e.g. of EAM's single-column model.) This is the first time I've done anything with the MPAS-Seaice code, so please help me out if I've made any basic mistakes here. I have not added any outputs related to gustiness to the default output, but you may wish to do so. When EAM is not being used, I've tried to use a gustiness value of 0 m/s, which should result in the same answers as on the current master.

There is some kind of stability issue with this branch in at least one `WCYCL1850NS` test. I've filed issue #5884 about that, because it seems like this is a consequence of an issue on `master` unrelated to this tag. But that appears to be primarily an atmosphere-land interaction issue, so whatever we do to resolve that is unlikely to have a significant effect on the MPAS-Seaice changes.

quantheory:
@bishtgautam @mabrunke @XubinZeng I have removed both atmosphere- and land-calculated gustiness (`wc`) from U10 in this revision. The removal of `wc` is the one change that I expect to affect answers without EAM running, but I believe it should only affect the U10 output of ELM.

rljacob:
status: wcycl1850NS test is failing.  threading non-BFB issues.

rljacob:
status: unchanged from above.

components/mpas-seaice/src/column/ice_atmo.F90
```diff
@@ -318,8 +324,8 @@ subroutine atmo_boundary_layer (sfctype,            &
             tau = rhoa * rd * rd ! not the stress at zlvl
 
             ! high frequency momentum coupling following Roberts et al. (2014)
-            strx = tau * sqrt((uatm-uvel)**2 + (vatm-vvel)**2) * (uatm-uvel)
-            stry = tau * sqrt((uatm-uvel)**2 + (vatm-vvel)**2) * (vatm-vvel)
+            strx = tau * sqrt((uatm-uvel)**2 + (vatm-vvel)**2 + ugust**2 ) * (uatm-uvel)
+            stry = tau * sqrt((uatm-uvel)**2 + (vatm-vvel)**2 + ugust**2 ) * (vatm-vvel)
```
proteanplanet, line 328: Please note that these lines do not conserve and need to be changed.  If gustiness is to be included, it must be relative to the sea ice velocity.  It's a common misconception that sea ice travels always much more slowly than the wind the ice velocity differential need not be calculated.  However, this does not account for super-inertial coupling when sea ice can travel in the opposite direction of the wind during inertial loops, nor in high through-flow regions, such as Fram Strait, where sea ice frequently travels faster than the wind, or in low wind speed scenarios when internal ice and water stress are the predominant drivers.  In the current coding, it is possible for sea ice to travel at the same speed and direction as the wind, but still accelerate due to wind stress.

proteanplanet (changes requested):

quantheory:
@proteanplanet @eclare108213 Sea ice changes are now removed from this branch (except for 3 lines needed by the atmosphere for diagnostic purposes).

crterai:
I have some results to share from the test simulations which were run after the removal of gustiness impacts on sea-ice. 
Here are the model vs model diagnostics from the F20TR simulation:
https://portal.nersc.gov/cfs/e3sm/terai/E3SMv3_dev/e3sm_diags_Gustinessfix_vs_baseline_F20TR/viewer/
I have also run 30 years with the WCYCL1850 in trigrid format, to be compared with the v3alpha04 simulation: 
https://portal.nersc.gov/cfs/e3sm/terai/E3SMv3_dev/e3sm_diags_WCYCL1850_gustinessfix_vs_v3alpha04_231017/viewer/
MPAS diagnostics from the 30 years with same simulation are here:
https://portal.nersc.gov/cfs/e3sm/terai/E3SMv3_dev/20231011.piControl.Gustinessfix.v3alpha04_trigrid.pm-cpu_intel/mpas_analysis/ts_0001-0030_climo_0001-0030/ocean/index.html

Summary
* In both simulations, the main noticeable change in the two simulations is the increased sensible heat flux and precipitation over Tropical land and the increase in RESTOM, which increases SSTs by 0.07 K
* The increase in RESTOM is due to an increase in absorbed SW and slight increase of trapped LW from thinner high level clouds

crterai:
What's concerning with the coupled simulation is a crash that the coupled model encountered 38 years into the simulation with the following error:
```
/pscratch/sd/t/terai/E3SMv3_dev/20231011.piControl.Gustinessfix.v3alpha04_trigrid.pm-cpu_intel/run/e3sm.log.17098756.231016-200022  

 256: PIO: FATAL ERROR: Aborting... An error occured, Writing variables (number of variables = 181) to file (./20231011.piControl.Gustinessfix.v3alpha04_trigrid.pm-cpu_intel.elm.h0.0038-02.nc, ncid=469) using PIO_IOTYPE_PNETCDF iotype failed. Non blocking write for variable (SNOWDP, varid=157) failed (Number of subarray requests/regions=1, Size of data local to this process = 64800). NetCDF: Numeric conversion not representable (err=-60). Aborting since the error handler was set to PIO_INTERNAL_ERROR... (/pscratch/sd/t/terai/E3SM_code/quantheoryE3SM/externals/scorpio/src/clib/pio_darray_int.c: 395)
```

Before this, I also see the following error messages. I haven't seen them before, and don't know if these are to be expected:
```
1871: sl_advection: reconstruct_and_limit_dp (rank,ie) returned   1871      1-3.1438E-01
2584: sl_advection: reconstruct_and_limit_dp (rank,ie) returned   2584      1-1.6182E-01
1988: sl_advection: reconstruct_and_limit_dp (rank,ie) returned   1988      1-1.2616E+00
1871: sl_advection: reconstruct_and_limit_dp (rank,ie) returned   1871      1-4.9239E-01
2985: sl_advection: reconstruct_and_limit_dp (rank,ie) returned   2985      1-1.7805E-01
2986: sl_advection: reconstruct_and_limit_dp (rank,ie) returned   2986      1-6.9487E-01
```
@xuezhengllnl , @ambrad, @wlin7 do you have any thoughts on what might be going wrong and whether I should be concerned with the `sl_advection` warning lines? 
Run script is here:
https://github.com/E3SM-Project/SimulationScripts/blob/master/archive/NGDAtmPhysIntegration/run.20231011.WCYCL1850.Gustinessfix.v3alpha04_wVBSfix.trigrid.pm-cpu.sh

XubinZeng:
Hi Chris (Terai),
I don't know what might be wrong. a couple of other comments:
a) from your results, I see the slight decrease of aerosol emission from oceans (e.g., sea salt) - which is consistent with what we expect. this might have a small effect on aerosol radiative forcing.

b) if possible, could you compare the current simulation (without gustiness impact over sea ice) versus the 40-year coupled simulation you did before (including gustiness impact over sea ice). then share the plots with us. from this comparison, we may gain some insights. 
thanks,
Xubin

crterai:
> Those lines are in general fine, although if you see a lot of them, that indicates instability in the model. 
There are a good number of these errors and they appear decades earlier as well in the simulation. So far the mean-state diagnostics don't indicate any instabilities, but things might be occurring at shorter frequency. 

> Am I correct that you're using a debug build?
I didn't intend to use a debug build, and I don't see that I did a debug_compile, so I don't know why they are printing if they are only intended in debug builds.

crterai:
Adding @beharrop as a reviewer, since he'd been involved in the original implementation of the gustiness parameterization.

ambrad:
@crterai what code version are you using? I'm wondering if recent cmake changes have affected flags like -DNDEBUG. Thanks.

crterai:
For the water cycle runs, I just used the PR branch, which built on top of `732e699695c67023880cb78937e21b5e66072879` of master, but it sounds like you've identified why it was printing those messages. 

And thanks for the info about the `sl_advection` messages not being error messages. I'm writing a restart close the crash and hopefully learn a bit more from extra output/checking flags.

crterai:
@XubinZeng - Thanks for your input. I think we have a slight increase in the emissions of sea-salt aerosols (please see table below). We do see a decrease in the dust emissions though, which is consistent with not double counting the gustiness over land. 
https://portal.nersc.gov/cfs/e3sm/terai/E3SMv3_dev/e3sm_diags_Gustinessfix_vs_baseline_F20TR/viewer/table/20231011.v3alpha04.F20TR.GustinessFix.v3alpha04_wSOAGfix.pm-cpu_intel-ANN-budget-table.html
 
As for the comparison with the earlier WCYCL1850 simulation, the base simulation that they are based off of are different (v3alpha02 vs v3alpha04), so I'm wary of comparing the two simulations and attributing any differences to the changes in whether the sea-ice gustiness was included or not.

XubinZeng:
@crterai  - thanks for the reply. agree on both points.

beharrop (approved):
Thanks for putting this all together.  The changes look reasonable and correct to me.

crterai:
@proteanplanet - would you review the changes to the code changes that shoul now have removed the impact of the gustiness on the sea-ice?

crterai:
@quantheory - I think we are close to finalizing this PR for merging. Would you remove the 'Draft' label to this PR?

proteanplanet (approved):
Keeping place holders in the sea ice coupling fields is fine for the moment and does not affect the ice physics, and can be implemented later in Icepack.   I approved of this from the polar side - @vanroekel will need to take a look and approve due to changes in turbulent fluxes over the ocean.   I could see no discernible change in ocean climate.

crterai:
Thanks for the reviews, @beharrop and @proteanplanet.

crterai:
An update on the coupled simulation crash at the end of `0038-02` due to `Non blocking write for variable (SNOWDP, varid=157) failed`: 
* I have been able to reproduce the crash twice (`17188180`, `17213876`), so it doesn’t appear to be a hardware issue.
* For one of the reproducers, I had tried to run with `state_debug_checks = .true.` to catch any odd values in the atmosphere and the checker didn't catch anything. 
* I have performed a branch simulation (BFB with original simulation) from a restart file written on the day the model crashes `0038-02-28`, writing high frequency output and expecting it to crash, but it completed the full day, getting past the crash point. My suspicion is that the bad SNOWDP data was written earlier in the month, but didn't crash when a bad SNOWDP value was produced and the simulation crashed only while writing the `elm.h0` file. And I suspect that the branch simulation just produced an `h0` file from the average values from `0038-02-28`, so it didn't crash.  `/pscratch/sd/t/terai/E3SMv3_dev/20231022.piControl.Gustinessfix.v3alpha04_trigrid.pm-cpu_intel.debuggingCrash/run` 
* Currently running a branch simulation from `0038-01-01` with daily `SNOWDP` output to catch where the faulty `SNOWDP` is actually written.

vanroekel:
@crterai do you have more than 30 years of simulation?  I thought climate changing simulations required 100 years?  Through 30 years the ocean seems mostly unchanged however I noted that OHC anomalies seem to be stronger in the gustiness run so would be useful to see a longer run.

crterai:
@vanroekel - we have up to 37 years of the water cycle simulation before it ran into a model crash while writing a diagnostic output variable (`SNOWDP`). I'm guessing 7 more years will not be enough for an assessment? 
I am working with the developers to try to get to the bottom of why it is crashing and will update here once we have a fix and longer (50+ year) coupled simulation. Do you notice any red flags from a coding perspective?

vanroekel:
Thanks @crterai I did look through the code and the only modification to the ocean seems to be changes to the wind stress we receive in the ocean so not really directly affecting the component like as it did the sea ice.  As for the length, 37 years is not much better than 30.  I do think we should push for 100 as that is the code review policy, but will let @golaz chime in there.  If Chris G is okay with a shorter testing simulation and I'm out on travel feel free to remove me as a reviewer and proceed.

rljacob:
update:  runtime fails have gone away.  Should be ready for alpha05.

crterai:
@vanroekel - After Sean Santos identified a fix for our earlier crash (issue came from a bug unrelated to this PR: https://github.com/E3SM-Project/E3SM/pull/6034), I've been able to push the simulation further than our earlier simulation and we now have 90+years. I've created the MPAS diagnostics IICE for the 100 year comparison: [IICE](https://portal.nersc.gov/project/m2136/bin/iice/mpas-a.cgi?url1=https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xzheng/E3SMv3_dev/20230924.v3alpha04_trigrid.piControl.chrysalis/mpas_analysis/ts_0001-0100_climo_0051-0100&label1=v3alpha04%20trigrid%20(1-100%20yr)&url2=https://portal.nersc.gov/cfs/e3sm/terai/E3SMv3_dev/20231011.piControl.Gustinessfix.v3alpha04_trigrid.pm-cpu_intel/mpas_analysis/ts_0001-0090_climo_0001-0090&label2=Gustiness%20fix%20%20%20v3alpha04_trigrid%20(1-90yr)&category=&component=&plot=&diff=0&dconfig=wcycl&). Would you take a look at the OHC anomalies to see if they look reasonable?

vanroekel:
Thanks for the updated diags @crterai I've looked through them and a couple things
1. could you possibly resubmit the diags with the climatologies from 51-100 (or whatever your last year is) instead of including all the record?  The comparison is not consistent with v3alpha4 right now
2. Looking through the time series there is definitely more OHC uptake with this PR, it's not largely different, but definitely stronger
3. the SSH appears to be increasing through the length of the run 
image

Does this make sense?  v3alpha4 seems to be stabilizing but not this run

XubinZeng:
Thanks Chris @crterai<https://github.com/crterai> !

agree with Luke:

  *   finish the 100-year run
  *   show the results from 51-100
  *   looking at the time series, there is a good chance the SSH will go down from 91-100 (and hence SSH is not increasing).

while not analyzing the ENSO carefully, from the power spectrum, the new run seems to increase the power spectrum density from 3-6 years - probably representing a small improvement over the control - a positive news.

From: Luke Van Roekel ***@***.***>
Sent: Wednesday, November 15, 2023 12:47 PM
Cc: Zeng, Xubin - (xubin) ***@***.***>; Mention ***@***.***>
Subject: [EXT]Re: [E3SM-Project/E3SM] Fixes for multiple issues in gustiness calculations. (PR #5850)


External Email

Thanks for the updated diags @crterai<https://github.com/crterai> I've looked through them and a couple things

  1.  could you possibly resubmit the diags with the climatologies from 51-100 (or whatever your last year is) instead of including all the record? The comparison is not consistent with v3alpha4 right now
  2.  Looking through the time series there is definitely more OHC uptake with this PR, it's not largely different, but definitely stronger
  3.  the SSH appears to be increasing through the length of the run
[image]<https://user-images.githubusercontent.com/13662545/283230351-82219789-b665-4e7c-aa4e-b6bef2fb095d.png>

Does this make sense? v3alpha4 seems to be stabilizing but not this run

crterai:
Thanks for looking at the diagnostics, @vanroekel. 

The last 6 year chunk is still in the queue, so here's the MPAS diagnostics where the climo fields are averaged over 0051-0090: [IICE](https://portal.nersc.gov/project/m2136/bin/iice/mpas-a.cgi?url1=https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xzheng/E3SMv3_dev/20230924.v3alpha04_trigrid.piControl.chrysalis/mpas_analysis/ts_0001-0100_climo_0051-0100&label1=v3alpha04%20trigrid%20(1-100%20yr)&url2=https://portal.nersc.gov/cfs/e3sm/terai/E3SMv3_dev/20231011.piControl.Gustinessfix.v3alpha04_trigrid.pm-cpu_intel/mpas_analysis/ts_0001-0090_climo_0051-0090&label2=Gustiness%20fix%20+%20v3alpha04_trigrid%20(1-90yr)&category=&component=&plot=&diff=0&dconfig=wcycl&).

Regarding the SSH, I don't see any reason why SSH should be different with the gustiness fix. However, I did take a look at the SSH trend in the baseline trigrid v3alpha04 simulation, which was continued out to year 300, and while we see it flatten from year 81-100, it does continue to go up after year 100 through the end of the simulation. So it looks like the increasing SSH is just a feature of our baseline, and not specific to the gustiness bug fixes. 
image

And yes, @XubinZeng, the ENSO power spectrum is one potential improvement that comes from these simulations.

golaz (approved):
Approving based on results presented and discussed during today's meeting: https://acme-climate.atlassian.net/wiki/spaces/ATMOS/pages/3998318593/2023-11-15+Meeting+notes+-+v3+Integration

vanroekel:
Thanks for the new diags @crterai I took a look through and things seem pretty close to the baseline.  Thanks for that SSH figure as well.  I'll go ahead an approve this now.

vanroekel (approved):

wlin7:
Merged to next.

rljacob:
removed from next with a force push.   Wait until tomorrow.

wlin7:
Re-merged to next.

wlin7:
Merged to master. Thanks @quantheory  for the great work, and careful reviews by all the reviewers.

Special notes on some diffs following this PR:

Standalone seaice tests (DTESTM) report diffs in a single variable, xaoo_So_u10, computed in coupler flux scheme and not used by sea ice code.

FIDEAL test reports diff, but only in field list. Otherwise it is BFB with baseline

FDPSCREAM test reports diff due to changes in lnd and atm-ocn flux calculations in coupler.

PR #5854, opened 2023-08-03
EAM/HOMME: Remove -DNC=4 in bld/configure; NC is an unused symbol.

ambrad:
It is believed that NC was used in HOMME long ago. In any case, it is not used any longer. Remove it from bld/configure.

[BFB]

### DIFF
--- a/components/eam/bld/configure
+++ b/components/eam/bld/configure
@@ -1863,7 +1863,7 @@ if (!$ccsm_seq){
 # on some older compilers used by HOMME standalone.
 if ($dyn_pkg eq 'se') {
     my $csnp = $cfg_ref->get('csnp');
-    $cfg_cppdefs .= " -DNP=$csnp -DNC=4 -DHAVE_F2003_PTR_BND_REMAP";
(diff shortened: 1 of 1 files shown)


ambrad:
Addresses https://github.com/E3SM-Project/E3SM/issues/5692.

ambrad:
e3sm_developor on Chrysalis passes.

ambrad:
@oksanaguba or @mt5555 do you by chance know what HAVE_F2003_PTR_BND_REMAP is for? I don't see it in the homme directory. I could remove it in this PR along with NC.

whannah1 (approved):

whannah1:
Thanks for doing this Andrew! My recent tests corroborate the conclusion that this is not needed.

oksanaguba:
HAVE_F2003_PTR_BND_REMAP does not look familiar to me.

rljacob (approved):

ambrad:
Thanks Oksana. Based on the comment above the line, `git blame`, and the fact that the symbol is extremely unlikely to conflict with any other symbols in E3SM, I'll keep -DHAVE_F2003_PTR_BND_REMAP. I might remove it later, if my guess that this became irrelevant during some boundary exchange refactor is correct.

PR #5857, opened 2023-08-08
Update MALI version

trhille:
This merge includes numerous updates to MALI.

The major updates include:

- Allow mask calving and physically based calving laws to be used simultaneously.
- Prevent basal melting in subglacial lakes.
- Use 3D thermal forcing field to drive grounded marine terminus melt parameterization.

Minor changes include:

- Updated Albany input file
- Allow  restore_calving_front to be applied after a calving law is applied
- Add an option for restore_calving_front to allow retreat
- Add `totalFaceMeltingFlux` to global and regional stats
- Limit velocity to user-defined maximum instead of calving fast-flowing ice when there is an instability.
- Fix treatment of von Mises stress on extended cells.
- Mask out floating ice for coupling to sea level model.
- Check that flow rate parameter is not constant when using von Mises calving.
- Fix grounded marine melting routine in order to pass decomposition tests.
- Calculate ismip6 retreat parameterization rate on initial time solve to be consistent with other calving routines.
- Calculate `cellMask` on initial solve for coupling with sea level model.
- Halo update during von Mises calving routine to pass decomposition tests.
- Change stride used in velocity solver interface.
- Bug fix involving invalid cell IDs in velocity solver interface.


matthewhoffman:
@trhille , thanks for initiating this.  I looked over the details, and I confirm this is set up as desired.  @jonbob , there are new namelist options in these commits, so the MALI default nl stuff will need updating.  Let me know if you want me to review when that's ready or if you want me or Trevor to set that up ourselves.

jonbob:
@trhille and @matthewhoffman -- I ran the scripts to get the bld files consistent with the Registry changes, and merged the updates to this branch. I didn't see any default settings used by E3SM getting changed, but please make sure that was your intention -- the scripts point out differences between the Registry and E3SM defaults, and usually we want the old E3SM ones. But not always...

jonbob:
NML DIFFs only for:
* ERS.f09_g16_g.MALISIA.chrysalis_intel

PASS:

merged to next

PR #5860, opened 2023-08-08
Implementation of the UCI vertical interpolation and forcing for Linoz

jinboxie:
Implement the UCI vertical interpolation and forcing to Linoz. The rationale behind it is that the linoz in e3sm currently use the linear interpolation that does not conserve mass. The UCI interpolation is implemented to remedy this problem and the associated forcing is also implemented.

Also change linoz_data_file in several use cases from 2015_2010JPL_cmip6_historical_10deg_58km_c20210625.nc to 
2017_CMIP6_Hist_10deg_58km_c20230705.nc

Allow output of surface level of 7 tracers (o3, n2o, noy, ch4, h2o, t, o3col).

[CC]
[NML]

--------------------------------------------------------------------------------------------------------------------

This includes 3 commits, and basically on the UCI interpolation algorithm change & the i/o structure (commit 1), adding global average ch4 input (commit 2), and other compsets change for forcing files (commit 3). 

Commit 1
1. The UCI vint is added to the tracer_data.F90 and switched to tracers for linoz to use.
2. The i/o for surface level of 7 tracers (o3, n2o, noy, ch4, h2o, t, o3col) is implemented for separate treatment in lin_strat_chem.F90.

Commit 2
1. The times series of global surface ch4 "ch4_avg_srf" is added as a one dimensional variable for global time-varying surface ch4 feature.

Commit 3
1. The forcing files in the namelist for 5 compsets are updated for the linoz forcing
a. 1850_eam_CMIP6_chemUCI-Linoz-mam5-vbs.xml
b. 2010_eam_CMIP6_chemUCI-Linoz-mam5-vbs.xml
c. 20TR_eam_CMIP6_chemUCI-Linoz-mam5-vbs.xml
d. scm_arm97_chemUCI-Linoz-mam5-vbs.xml
e. scm_generic_chemUCI-Linoz-mam5-vbs.xml

--------------------------------------------------------------------------------------------------------------------

Figures and diagnostics
The diagnostic package of the results is tested against alpha01 master results of [Golaz](https://acme-climate.atlassian.net/wiki/spaces/CM/pages/3826712703/20230616.v3alpha01.amip.chrysalis), the link for our results is summarized in [here](https://acme-climate.atlassian.net/wiki/spaces/ATMOS/pages/3844866128/Long-term+AMIP+simulation+for+Linoz+v2+v3+UCI+interpolation+and+forcing). Due to a new bug fix, the latest fixed results are updated [here](https://acme-climate.atlassian.net/wiki/spaces/ATMOS/pages/3905454094/UCI+interpolation+bug+fix+simulation).

--------------------------------------------------------------------------------------------------------------------
Related-PRs
This PR concerns two PR in v3atm and one PR in master shown below:
https://github.com/E3SM-Project/v3atm/pull/62
https://github.com/E3SM-Project/v3atm/pull/64.
https://github.com/E3SM-Project/E3SM/pull/5831


components/eam/src/chemistry/mozart/lin_strat_chem.F90
```diff
@@ -158,12 +158,14 @@ subroutine lin_strat_chem_inti(phys_state)
...
-       return
-    end if
+!!Jinbo Xie
+    !if (uci1_ndx <=0 ) then
+    !   !write(iulog,*) 'skip Linoz, need to have tracer O3LNZ at least '
+    !   write(iulog,*) 'skip Linoz, temporally change '
+    !   do_lin_strat_chem = .false.
+    !   return
+    !end if
+!!Jinbo Xie
```
wlin7, line 168: @jinboxie , please remove the above comment lines

components/eam/src/chemistry/mozart/lin_strat_chem.F90
```diff
@@ -278,7 +291,7 @@ subroutine linv3_strat_chem_solve(ncol, lchnk, xvmr, h2ovmr, xsfc, o3col,temp, s
     real(r8), intent(in)                           :: delta_t             ! timestep size (secs)
     real(r8), intent(in)                           :: rlats(ncol)         ! column latitudes (radians)
     integer,  intent(in)   , dimension(pcols)      :: ltrop               ! chunk index    
-    real(r8), intent(in)   , dimension(pcols,pver) :: pdeldry             !  dry pressure delta about midpoints (Pa) 
+    real(r8), intent(in)   , dimension(pcols ,pver) :: pdeldry             !  dry pressure delta about midpoints (Pa)
```
wlin7, line 294: A small change. Remove the space after 'pcols', .i.e., reverting the changes seen by git.

rljacob (changes requested):
Fix the PR title and description.

jinboxie:
> Fix the PR title and description.

Hi Rob, I've made a fix. Please let me know if there are other suggestions. Thank you!

jinboxie:
> Fix the PR title and description.

Hi @rljacob Rob, I've made a fix. Please let me know if there are other suggestions. Thank you!

rljacob (approved):

tangq (approved):
Need to repeat the test runs to confirm the results are reasonable and clean up the code a bit as suggested above. The code changes look good to me.

jinboxie:
> I have reviewed this PR as we want to ensure ozone changes aren't introducing significant changes to aerosol forcing. I analyzed the radiation aspect of this PR (thanks to @jinboxie for conducting additional simulations).
> 
> Looking at the last 10 years of the 20TR runs (from 2005 to 2014) comparing a run without the changes in this PR and one with the changes:
> 
>     * RESTOM changes by about -0.14 W/m2.
> 
>     * **Direct O3** interactions with radiation are about -0.39 W/m2.
> 
>     * And so there is about +0.25 W/m2 of **non-determined “indirect”** interactions (these are side effect of O3 and non-O3 changes, but I believe they are mainly O3-related).
> 
>     * **Direct aerosol** interactions with radiation are only -0.02 W/m2, which is a good indication there is no significant impact of the changes herein on sulfate in present-day conditions. (Additionally, looking at the piControl run, there seems to be little effect on aerosols there too.)
> 
> 
> Based on the above, I am approving this PR. Thank you @jinboxie for your patience and taking the time to conduct additional O3-specific experiments.
> 
> Here's a plot of the spatial aerosol direct effect, indicating potentially compensating changes.
> 
> direct-aerosol

Thank you Naser for taking time and digging deep on this PR. Glad the diagnostic simulations paid off.

jinboxie:
Thank you Naser @mahf708 for taking time and digging deep on this PR. Glad the diagnostic simulations paid off.

components/eam/src/chemistry/utils/tracer_data.F90
```diff
@@ -1602,6 +1637,171 @@ subroutine read_za_trc( fid, vid, loc_arr, strt, cnt, file, order )
...
+        else
+                wrk_in => wrk
+        end if
+        !!
+        do c=begchunk,endchunk
+           ncols = get_ncols_p(c)
+           call get_rlat_all_p(c, pcols, to_lats)
+           call lininterp_init(file%lats, file%nlat, to_lats, ncols, 1, lat_wgts)
+           do k=1,1
+           call lininterp(wrk_in(:), file%nlat, wrk(1:ncols), ncols, lat_wgts)
```
wlin7, line 1792: This is problematic, as wrk_in and wrk point to the same address. Inside lininterp (lininterp1d), within the loop computing for wrk, the original values (in wrk_in or wrk) are successively overridden. In addition to the results being unintended, because even the source data are replaced by the computed values, which are then used for computing values at remaining column, the results become threading and pe-layout sensitive.

tangq: Hi, Wuyin. This looks like it. We will discuss a fix tomorrow.

wlin7:
@jinboxie @tangq , I committed the fixes for the debug and threading/pe-layout tests failure ([3516f7d](https://github.com/E3SM-Project/E3SM/pull/5860/commits/3516f7dfe4ecac6c4fe7c7cb846fc059957ae773)). Please take a look.

wlin7:
Commit [aa5535b](https://github.com/E3SM-Project/E3SM/pull/5860/commits/aa5535b7cfe75eb86bdb2eb448e17a87eefd6f3b) fixes tests like FAQP that do not use linoz. Otherwise the tests would fail with segmentation fault.

wlin7:
> @wlin7 is [3516f7d](https://github.com/E3SM-Project/E3SM/pull/5860/commits/3516f7dfe4ecac6c4fe7c7cb846fc059957ae773) BFB for simulations conducted on chrysalis? I got the impression from your comment that it may not be:

You are right. The fix will give NBFB results. The seven zonal mean surface fields are affected. Values from unintended latitudes would be use due to the corruption of source data for the N-S direction interpolation. The impact could still be limited. I am running an 11-year pre-v3alpha03fix.F2010.

wlin7:
Nice observation, @mahf708 . The bug, in terms of how the values could be mixed up between different latitudes, does imply the tendency to introducing more noise.

tangq:
@wlin7 and @mahf708 , `o3_clim_srf` should be used when there is no interactive tropospheric chemistry, but our runs have the tropospheric scheme, so `o3_clim_srf` should NOT affect the results and is not relevant to Yan's analysis. @jinboxie , can you double check if this is how you implemented it? Thanks.

jinboxie:
> @wlin7 and @mahf708 , `o3_clim_srf` should be used when there is no interactive tropospheric chemistry, but our runs have the tropospheric scheme, so `o3_clim_srf` should NOT affect the results and is not relevant to Yan's analysis. @jinboxie , can you double check if this is how you implemented it? Thanks.

Hi Qi, it is confirmed.

susburrows:
All, should this PR be tagged as a bug fix?

jinboxie:
Status update in this PR. Before the merging, a bug that affects the surface boundary for Linoz tracers is detected. A bug was found during our discussion with Wuyin. A fix is now in the PR to incorporate the fix. Simulation using the latest commit has shown a RESTOM of -0.03 W/m2 as compared to the -0.04 W/m2 in the pre-bug fix. Decomposition of the radiation effect following the above analysis show similar results, including minor aerosol effect. Inspection of the results in other variables also show similar effect, indicating the minor impact of this bug. The results are documented **[here](https://acme-climate.atlassian.net/wiki/spaces/ATMOS/pages/3905454094/UCI+interpolation+bug+fix+simulation )**.

susburrows:
@wlin7 : should this be tagged as a bugfix PR?

rljacob:
Its only a bugfix if the bug is on master.

wlin7:
Merged to next.

components/eam/src/chemistry/utils/tracer_data.F90
```diff
@@ -1602,6 +1650,183 @@ subroutine read_za_trc( fid, vid, loc_arr, strt, cnt, file, order )
...
+                if (present(vid_srf)) then
+                !!padding for clim terms
+                allocate(wrksrf(cnt(1)), stat=ierr )
+                ierr = pio_get_var( fid, vid_srf, strt_srf, cnt_srf, wrksrf )
+                !!surface padding
+                wrk2d(:,file%nlev)=wrksrf
+                deallocate(wrksrf)
+                else
+                !!padding for derivative terms
+                wrksrf=0.0_r8
```
wlin7, line 1725: @jinboxie , look like there is no need of this (padding for derivative).
wrksrf in the else block is never allocated. While runs with intel compiler didn't complain, all gnu tests failed due to invalid memory reference.

wlin7: Verified: Commenting out lines 1723-1725 allows gnu tests to run, while having no impact on intel tests.

jinboxie: Ok, thank you guys!

wlin7:
Remerged to next after the fix commit [51dd660](https://github.com/E3SM-Project/E3SM/pull/5860/commits/51dd660c682985e9f245635e74019c532703338b)

components/eam/cime_config/config_compsets.xml
```diff
@@ -106,6 +106,16 @@
...
   </compset>
 
+  <compset>
+    <alias>FSSP370</alias>
+    <lname>SSP370_eam_EAM%CMIP6_ELM%SPBC_MPASSI%PRES_DOCN%DOM_MOSART_SGLC_SWAV</lname>
+  </compset>
+
+  <compset>
+    <alias>FSSP585</alias>
+    <lname>SSP585_eam_EAM%CMIP6_ELM%SPBC_MPASSI%PRES_DOCN%DOM_MOSART_SGLC_SWAV</lname>
```
rljacob, line 116: What does the "_eam_" do in this longname?  I didn't think that was allowed in CIME.

rljacob:
The PR description was missing important content (above the line) so I added it.

components/eam/cime_config/config_compsets.xml
jinboxie: Hi Rob, it was used as the F for SSP585 compset when we were testing using the original alpha01 version. It is used to test the linoz-v2 configuration for future compset. Under this compset, we used several future scenario forcings as like done before. In the naming convention of correspondent files in use_cases of eam namelist_files, there was one with SSP585_cam5_CMIP6_bgc.xml, so I named it with eam to discriminate between the two.

wlin7:
Merged to master.

PR #5862, opened 2023-08-09
add select Mass_at_pressure 2D fields for 7 aerosol species

mahf708:
Adds output fields for 2D Mass_at_pressure fields for 7 aerosol species in lieu of their 3D counterparts. 

Since the move to get rid of 3D Mass_aerosol fields from the standard output in the upcoming simulations, aerosol developers and domain scientists wanted the 2D equivalent at select pressure levels. Previously, we added the column-integrated quantities as replacement, now we add vertical slices as well. None so far is added to the default output. I elected to add only five levels for now, but it should be relatively easy to expand this or change the levels. We could also expand this to non-aerosol species, e.g., gaseous precursors if needed.

[BFB]

### DIFF
--- a/components/eam/src/chemistry/mozart/mo_chm_diags.F90
+++ b/components/eam/src/chemistry/mozart/mo_chm_diags.F90
@@ -62,7 +62,6 @@ module mo_chm_diags
 
   character(len=32) :: chempkg
 
-  integer :: aerosol_idx
   integer, parameter :: aerosol_number = 7
   character(len=3) :: aerosol_name(aerosol_number)
 
@@ -124,6 +123,8 @@ contains
 
     integer :: bulkaero_species(20)
     integer :: e90_ndx
+
+    integer :: aerosol_idx
 
     !-----------------------------------------------------------------------
 
@@ -1120,6 +1121,8 @@ contains
     logical :: history_aerosol      ! output aerosol variables
     logical :: history_verbose      ! produce verbose history output
     logical :: get_presc_aero_data      ! produce output to drive prescribed aerosol run
+
+    integer :: aerosol_idx
 
     !-----------------------------------------------------------------------
 
--- a/components/eam/src/chemistry/mozart/mo_chm_diags.F90
+++ b/components/eam/src/chemistry/mozart/mo_chm_diags.F90
@@ -13,6 +13,8 @@ module mo_chm_diags
   use cam_logfile,   only : iulog
   use spmd_utils,    only : masterproc
 
+  use interpolate_data,   only: vertinterp
+
   implicit none
   private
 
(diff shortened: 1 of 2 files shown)


hwangacme (approved):
All looks good to me!

mingxuanwupnnl (approved):

susburrows:
@mahf708 this is great, thanks for adding this capability.  You specifically asked (offline) for my feedback on the pressure levels to use, so I'm responding only on that point in my review. I would have initially gone with surface, 850, 500, and 200.  My rationale for this is:

1. 850 mb is typically near the top of the planetary boundary layer, for locations near sea level, so it approximates the boundary-layer concentrations (as opposed to the surface), and it approximates the aerosol that would be "seen" by near-surface clouds that are within the boundary layer -- also relevant to any clouds with cloud bases within the boundary layer.  We also plot key dynamic variables (T, U, OMEGA) on the 850 mb level in our standard diagnostics.

2. 500 mb is a mid-troposphere level (often used by meteorologists for evaluating overall ridge/trough patterns and vorticity) and gives us kind of a midpoint for measuring constituent concentrations.  We already plot Omega (vertical pressure velocity) and Z (geopotential height) at this level.

3. 200 mb is the upper troposphere level that is used in our standard diagnostics.  This is where we see the jets and generally geostrophic winds.

4. The plots at 200mb and 300mb are likely to be largely redundant.  The gradients in chemical constituents are typically not huge in the upper troposphere, except around deep convective clouds (due to wet removal, cloud processing, and convective mass transport), and of course when crossing the tropopause or going into polar night.  While it's true that the jets are closer to 300 mb in the winter and 200 mb in the summer, I don't think this will matter much for our analyses of the aerosol variables.  I think what we are looking for here is the overall concentration and spatial pattern of aerosol (by species) in the upper troposphere. While it will look a bit different at 300 vs 200 mb, I don't think those differences are likely to drive our decision-making.

OTOH: after reviewing your list of pressure levels, I do think there might be some advantages to including 300mb or 330mb, since the 200 mb level can cross the tropopause, particularly near the poles.  If we're going to add one more, I would go with 330 mb rather than 300 mb.  It is a little more firmly in the troposphere, and it has the nice advantage that we know of at least one previous study on global modeling of sulfate aerosol (albeit, a very early one) that reported results at this level (Chin and Jacob, 1996).  So we'd have at least that reference to compare against.  330 is also closer to the midpoint between 200 and 500.  (and although 300mb is a standard analysis level for synoptic meteorology, it's not part of our standard diagnostics in e3sm_diags -- I'm not sure if we're using it anywhere else?).

Since this calculation will not be expensive, I'd err on the side of including an extra level rather than excluding it.

So, I'd recommend going with surface, 850, 500, **330**, 200.

susburrows (approved):
I am approving the PR, but I suggest changing from 300 to 330 mb, as noted in my previous comment.

chengzhuzhang:
@wlin7 this PR is coordinating with Xue's output reduction effort. It needs to be merged to master for part of output namelist for alpha03 to work. Could you help merge when get a chance? Thank you.

wlin7:
Merged to next.

wlin7:
Merged to master

chengzhuzhang:
@wlin7 Thank you!

> Merged to master

PR #5823, opened 2023-07-19
Consolidate v3 compsets with v3alpha02 configuration

wlin7:
v3atm based simulations have been relying on a combination of physics/chemistry specific compsets and customized run scripts. The run scripts typically contain additional configuration options replacing physics, chem package, chemistry mechanism file, forcing files, and tuning options.

In this PR, all the above are consolidated into compsets and the corresponding complete configuration options for the component (EAM). The v3alpha02 configuration is made default for the consolidated compsets.

The compset names, both the aliases and the longnames, retain the format used in V2. For example, F1850, F2010, F20TR have corresponding longnames of 1850_EAM%CMIP6.+, 2010_EAM%CMIP6.+ and 20TR_EAM%CMIP6.+. The new physics and chemistry from v3atm, including chemUCI, Linozv3, MAM5, VBS-SOA, and ZM enhancements, do not appear in the compset names.

The new physics and chemistry are indicated in the description of the compsets, and reflected in the new use_case files.

This PR will retire V2 compsets under the same names upon merging.

Additional changes are added to enable or fix nightly tests that will use the new code base along with the consolidated compsets. These include FIDEAL, FAQP, SCM, BGC, clubb-only, chem_pp.

[BFB] for all equivalent v3 configurations that previously were controlled via run script.

--------------------------------------
With this PR up to present, the following compsets are enabled with v3atm
F1850, F2010, F20TR
WCYCL1850, WCYCL1850-4xCO2, WCYCL1850-1pctCO2, WCYCL20TR

v3 SSP compsets (SSP370 and SSP585)  are not ready. Surface and upper level emission files for the new tracer species are needed. Nightly SSP tests may not be able to run when v3 features are switched on but new emission files not ready.

PR #5806, opened 2023-07-13
Fix lightning NOx emission bug

tangq:
During v3atm aerosol-chemistry integration, "NO" was removed from the external foring list of chemistry mechanism files ending in _vbs.in. This means that the lightning NOx emissions won't be used by the simulations using these mechanisms. This PR corrects this bug.

[non-BFB]

-------------------------------------
The test simulation was discussed at the [v3atm integration meeting](https://acme-climate.atlassian.net/wiki/spaces/ATMOS/pages/3854434386/2023-07-12+Meeting+notes+-+v3+Integration) and documented at this confluence [page](https://acme-climate.atlassian.net/wiki/spaces/ATMOS/pages/3849715713/20230707.NO-lght-fix.5x.v3alpha02.F2010.chrysalis).

PR #5867, opened 2023-08-10
Revert the revert of lightning NOx bug fix and rename chem package

wlin7:
This is to reactivate the bug fix. It was merged but then reverted to maintain v3alpha02 code base.

The campp autogenerated files associated with the updated mechanism are also committed to
let the bug fix take effect.

The chem package superfast_mam5_resus_mom_soag is renamed as chemuci_linozv3_mam5_vbs, 
along with other changes in the codes and build scripts accordingly.

[CC]
[NML]

--------------------------------------------
Original bug fix PR is #5806 
The fix was reverted via #5823, which was merged to #5781 and then took effect in master.

Note that the fix is via the mechanism file, and the mechanism file  for consolidated v3 compsets, pp_chemUCI_linozv3_mam5_vbs.in, is no more specified via usr_mech_file, the revert of revert alone has no effect. The chem preproocessor need to rerun with the update mechanism, then commit the auto-generated files.


wlin7:
The commit history for the big fix went into master via #5781 (itself via #5823). The original merge commit was reverted via #5823. So, essentially yes, the bug fix not activated in master. This PR will revert the revert, and activate the original bug fix to the mechanism file. But revert the revert alone wouldn't enable the fix, now that we stop using user specified chem_mech. The new commit ([4cfa26b](https://github.com/E3SM-Project/E3SM/pull/5867/commits/4cfa26b8fdb50a5f78e5b78151a8e1988b876c78)) will do it, with the updated auto-generated file associated with the updated mechanism. With that, this PR is NBFB.

wlin7:
> We can still make this BFB if we make alternate chem mech files, e.g., instead of editing `components/eam/chem_proc/inputs/pp_chemUCI_linozv3_mam4_resus_mom_vbs.in`, copy a new `components/eam/chem_proc/inputs/pp_chemUCI_linozv3_mam4_resus_mom_vbs_noxfix.in` from it and edit it, while keeping the old ones. Then, these could be activated in v3alpha03 without having to make this PR CC.

@mahf708 , in previous format of running simulation, yes, this is a way to have the fix in place for use via usr_mech_file in run script. Going forward, we want to make the updated mechanism default. We can still decided when to put this PR in. We could let it go into #5860 to test the combined effect of Lightning NOx fix (plus tuning, say, 5x) and UCI interpolation, as you suggested. BTW, I like your idea of combining them for required simulations, particularly the long coupled run, given their major direct effect are on ozone.

wlin7:
@tangq , @keziming , some additional info that may be helpful when you review.
Two tests were done. One specified the updated pp_chemUCI_linozv3_mam5_vbs.in via usr_mech_file, which trigger campp and auto-generate the files. The other without specifying usr_mech_file, but with updated chem_mech.in and other autogenerated files commiteed. The two runs were verified BFB.

BTW, for the older mechanism files under eam/chem_proc/inputs, they are usually accompanied with a .inp file? Is it also needed for pp_chemUCI_linozv3_mam5_vbs.in? Look like .inp file is only used when maunally running campp from a working directory under components/eam/chem_proc. (It may still be useful to include, just in case one day someone wants to do that. Would like to have your opinion on this.)

tangq (approved):
The changes are straightforward and look good to me.

@wlin7 , regarding the *.inp file, that's optional and it's fine to leave it there.

A separate note, the pp directory name would be misleading to users familiar with campp because the string after "pp_" should match the *.in file name. For those users, they would think this pp_superfast_mam5_* directory uses the superfast mechanism. Perhaps, it can be fixed in a future PR.

wlin7:
Thanks for the reminder, @tangq . I did plan to also change the chem package name chemUCI_linozv3_mam5_vbs in this PR. I will keep you posted.

keziming (approved):

wlin7:
Through [ae3b0c9](https://github.com/E3SM-Project/E3SM/pull/5867/commits/ae3b0c9672b19d9fd230ad2e6b61b9ded76b2a45), the chem package name is changed from `superfast_mam5_resus_mom_soag` to `chemUCI_linozv3_mam5_vbs`. The chem package name in Fortran codes, namelist file, and several pre-processing scripts use lower case. To be consistent, `chemuci_linozv3_mam5_vbs` in  all files involved.

1-month F20TR test verified the results are BFB before and after these changes.

tangq:
Hi @wlin7 , thanks for implementing this. I will use the chemistry name `chemUCI_linozv3_mam5_vbs` to replace the `uci1` reaction existence checks later today and let you know how it goes.

> Through [ae3b0c9](https://github.com/E3SM-Project/E3SM/pull/5867/commits/ae3b0c9672b19d9fd230ad2e6b61b9ded76b2a45), the chem package name is changed from `superfast_mam5_resus_mom_soag` to `chemUCI_linozv3_mam5_vbs`. The chem package name in Fortran codes, namelist file, and several pre-processing scripts use lower case. To be consistent, `chemuci_linozv3_mam5_vbs` in all files involved.
> 
> 1-month F20TR test verified the results are BFB before and after these changes.

tangq:
Hi @wlin7 , while changing the `uci1` reaction existence checks, I realized that using `uci1` to check if chemUCI is used is more convenient than the chemistry name string (e.g., `chemUCI_linozv3_mam5_vbs`) because chemUCI may be associated with multiple chemistry names in the future. Therefore, I decided NOT to change the `uci1` checks.

A separate note:
While reviewing the code, I noticed a couple minor issues (loop ordering and typos) and fixed them in [5802685](https://github.com/E3SM-Project/E3SM/pull/5867/commits/5802685eb177e653672f1792f2d233ec585586bb). I've confirmed that the commit [5802685](https://github.com/E3SM-Project/E3SM/pull/5867/commits/5802685eb177e653672f1792f2d233ec585586bb) is BFB.

wlin7:
Merged to next.

wlin7:
Merged to master.

PR #5872, opened 2023-08-11
Add micro_p3 to list of possible ptend names in physpkg

mkstratos:
Fixes missing field error when running PGN test
`ERROR: GET_MASTERLIST_INDX: attemping to output field s_micro_p3 not on master list`
Fixed by adding `micro_p3` to the list of possible ptend field names.
[bfb]


singhbalwinder:
Merged to next

PR #5873, opened 2023-08-11
Change TSC test default to use F2010

mkstratos:
Revert TSC non-b4b test to use the F2010 compset

With the fix in MPAS-SI: "Corrects thin ice/snow treatment of enthalpy and other tracers" (#5630) the TSC test can now generate a baseline with dtime=1 using MPAS-SI

[non-BFB] only for TSC test

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -221,7 +221,7 @@ _TESTS = {
     "e3sm_atm_nbfb" : {
         "tests" : (
             "PGN_P1x1.ne4_oQU240.F2010",
-            "TSC.ne4_oQU240.F2010-CICE",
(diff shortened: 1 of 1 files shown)


rljacob (approved):

rljacob:
merged to next.

rljacob:
next testing is showing that MPASSI is being used now but there's still an unrelated error.
2023-09-13 05:55:08: Exception during BASELINE:
ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
multiprocessing.pool.RemoteTraceback:

I'm going to go ahead and merge this since the seaice component is successfully changed.

PR #5874, opened 2023-08-11
fix openacc data directive on non-existent ssh_sal array

### LINKED ISSUE #5866: OpenACC bug in /components/mpas-ocean/src/shared/mpas_ocn_tendency.F
At line 433 (in maint2.1) and 437 (in master) of E3SM/components/mpas-ocean/src/shared/mpas_ocn_tendency.F in the ocn_tend_vel subroutine, there is an error in the OpenACC declaration referring to ssh_sal. This variable does not exist. 

` !$acc update device(ssh_sal)`

The error received in maint2.1 for MPAS-Ocean compiled for OpenACC is: 

`FATAL ERROR: data in update device clause was not found on device 1: name=ssh_sal(:)`

I believe ssh is meant, and I have successfully ran past this error by correcting the update device call to ssh, i.e., 
 
`!$acc update device(ssh) `

philipwjones:
An openacc data transfer on the self-attraction ssh change was outside any conditional for SAL so was operating on a non-existent array if SAL is turned off. This mod changes the location of the data directive so it is within the compute SAL routine and only executed of SAL is turned on

bfb
fixes #5866

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_tendency.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_tendency.F
@@ -434,7 +434,6 @@ contains
           call ocn_compute_self_attraction_loading(domain, forcingPool, dminfo, ssh, err)
           call mpas_reset_clock_alarm(domain % clock, 'salComputeAlarm', ierr=err)
       endif
-      !$acc update device(ssh_sal)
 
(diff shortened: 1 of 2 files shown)


jonbob:
passes:

merged to next

jonbob:
merged to master

Issue #5784, opened 2023-06-29
Incorrect units of active tracer tendencies in the ocean Registry


cbegeman:
In the process of @akturner's work on frazil flux budgets, we discovered that `salinityTend` needed to be divided by a factor of `layerThickness`. This indicates that the units should have an additional factor of `m`. By inspecting the tracer tendency updates in both RK4 and split-explicit, I confirmed that the `activeTracerTend`'s are divided by a factor of `layerThickness`. I found no place where the `activeTracerTend`'s are divided by `layerThickness` prior to being written to streams. My proposed bugfix is here: 7b00e5ace26ae30549bafee5bf79a053bbc18543.

darincomeau:
Units in that commit look good (CF-compliant) to me

mark-petersen:
Yes, it is correct to include a length unit in the numerator, since we time-step thickness-weighted tracers: 
![image](https://github.com/E3SM-Project/E3SM/assets/4036816/d3edb999-f567-458c-938a-2c6311fd5c62)
Thanks for checking!

cbegeman:
Great, thanks @darincomeau and @mark-petersen!

cbegeman:
Resolved by https://github.com/E3SM-Project/E3SM/pull/5883

PR #5883, opened 2023-08-15
Correct the units of active tracer tendencies

cbegeman:
The units of active tracer tendencies in the Registry should have an additional factor of `m`. 

The `activeTracerTend`'s are divided by a factor of `layerThickness` when they are used in the code to update active tracers. This operation is also needed to evaluate tracer conservation from `activeTracerTend` output.

[BFB]


cbegeman:
Fixes https://github.com/E3SM-Project/E3SM/issues/5784

mark-petersen (approved):
Perfect, thanks!

jonbob:
passes:

merged to next

cbegeman:
@jonbob Thanks!

Issue #5875, opened 2023-08-11
non-bfb results with SMS_D_PMx2_Ln5.ne30_oECv3.F2010 --compiler=gnu (DEBUG plus threading)


ndkeen:
On GCP12 (using next of Aug10) (and on pm-cpu, see below comments).
there is a test showing different results between runs.  Seems to be with GNU/DEBUG and threading.

```
SMS_D_Ln5.ne30_oECv3.F2010.gcp12_gnu.C.20230810_184148_icsnhz/run/atm.log.25396.230810-185308.gz: nstep, te        6   0.26327188655393944E+10   0.26327220866024842E+10   0.17809334694391188E-03   0.98542160462986372E+05
SMS_D_Ln5.ne30_oECv3.F2010.gcp12_gnu.G.20230810_171201_yb8tm7/run/atm.log.25376.230810-172936.gz: nstep, te        6   0.26327191183128428E+10   0.26327222939713116E+10   0.17558291039020528E-03   0.98542162828852393E+05
SMS_D_Ln5.ne30_oECv3.F2010.gcp12_gnu.r00/run/atm.log.25545.230811-160939.gz: nstep, te        6   0.26327190694097590E+10   0.26327222492494040E+10   0.17581408918316275E-03   0.98542162509758753E+05
SMS_D_Ln5.ne30_oECv3.F2010.gcp12_gnu.r01/run/atm.log.25543.230811-160939.gz: nstep, te        6   0.26327190279278388E+10   0.26327222114278817E+10   0.17601647382655078E-03   0.98542162303609221E+05
SMS_D_Ln5.ne30_oECv3.F2010.gcp12_gnu.r02/run/atm.log.25544.230811-160939.gz: nstep, te        6   0.26327190697470407E+10   0.26327222474688978E+10   0.17569699498611472E-03   0.98542163185255980E+05
```

These 2 tests were also giving me trouble with comparing to baselines (resolved with PR from Wuyin). The `nstep` values written in atm.log files are the same between multiple runs -- however, cprnc on the resulting eam netcdf files show DIFFERENT.  ?

```
SMS_Lm1.ne4_oQU240.F2010.gcp12_gnu
SMS_Ly1.ne4_oQU240.F2010.gcp12_gnu
```
Therefore, for these 3 tests, I'm unable to bless the results and we still see 3 baseline fails.

ndkeen:
[...]

PR #5886, opened 2023-08-16
Fix array shapes in where assignment statements in micro_p3_interface.F90

### LINKED ISSUE #5864: Unexpected diffs in ANRAIN / AQRAIN /  FREQR cause some cime tests labeled as DIFF
This happens to the following chrysalis tests

```
SMS_D_Ld1.ne30pg2_EC30to60E2r2.WCYCL1850.pm-cpu_gnu.allactive-wcprod
SMS_D_Ln5.ne4_oQU240.F2010.pm-cpu_gnu.eam-cosplite_nhtfrq5
SMS_Ln9.ne4_oQU240.F2010.pm-cpu_gnu.eam-outfrq9s
```
The tests are otherwise BFB with baselines.

wlin7:
Mismatch of array shapes in where assignment statements and where mask expression causing three diagostic variables, AQRAIN, ANRAIN, FREQR, to differ from one run to another when using gnu compiler. The declared shapes of the three arrays are the same as the array rain. While the scope could be narrowed with rain(:ncol,top_lev:) in the mask expression, the assignment statments for the three affected variables still use the full scope. Although test with other compilers, e.g., intel, does not report diff, the diagnostic output for these three variables are assumed to have also be affected.

Fixes #5864

[NBFB] for tests that compare eam.h0 with diff in AQRAIN, ANRAIN and FREQR.
       The test results are otherwise not affected.

### DIFF
--- a/components/eam/src/physics/p3/eam/micro_p3_interface.F90
+++ b/components/eam/src/physics/p3/eam/micro_p3_interface.F90
@@ -1565,9 +1565,9 @@ end subroutine micro_p3_readnl
            numrain(:ncol,top_lev:) * rho(:ncol,top_lev:), &
            rho(:ncol,top_lev:), rho_h2o)
 
-      aqrain = rain * cld_frac_r
-      anrain = numrain * cld_frac_r
-      freqr = cld_frac_r
+      aqrain(:ncol,top_lev:) = rain(:ncol,top_lev:) * cld_frac_r(:ncol,top_lev:)
+      anrain(:ncol,top_lev:) = numrain(:ncol,top_lev:) * cld_frac_r(:ncol,top_lev:)
+      freqr(:ncol,top_lev:) = cld_frac_r(:ncol,top_lev:)
       reff_rain(:ncol,top_lev:) = drout2(:ncol,top_lev:) * &
            1.5_rtype * 1.e6_rtype
    end where


wlin7:
The diffs were reported for several tests on pm-cpu_gnu that compare eam.h0.  While pm-cpu_gnu is having build (ld) issue, the problem was reproduced with REP_Ln9.ne4_oQU240.F2010.chrysalis_gnu.eam-outfrq9s. And the changes were confirmed fixing the REP test that would otherwise report diff on the three variables.

Though the diffs were only reported for pm-cpu_gnu tests, the nature of the issue (assignment shape mismatch) suggests it is a problem for other compilers as well, which could cause data for some columns and levels to be displaced. 

Ping @yunpengshan2014 , @mahf708 . If previous analyses involved AQRAIN, ANRAIN, and FREQR, please keep in mind the results could be affected.

ndkeen:
Nice work Wuyin. I may not be able to test until tomorrow, but it sounds like this will fix the odd issues I was having with tests on GCP. Where 3 tests (in extra_coverage) where not BFB -- 2 of them only with eam output.  https://github.com/E3SM-Project/E3SM/issues/5875

rljacob (approved):
The dangers of array syntax.

yunpengshan2014:
Dr. Lin,

Thanks a lot for reminding me. The bug-fix PR looks good to me.
Regards,
Yunpeng

yunpengshan2014 (approved):

singhbalwinder (approved):
Thanks for fixing this!

ndkeen (approved):

wlin7:
Merged to master.

wlin7:
This fix appears to have also fixed SMS_D_Ln5.ne4_oQU240.F2010.pm-cpu_nvidia which previously had the following run time error (as documented on [5781 discussion](https://github.com/E3SM-Project/E3SM/pull/5781#issuecomment-1668271465)).  However, ERS_D.ne4_oQU240.F2010.pm-cpu_nvidia.eam-hommexx is still a FAIL. Both to be further examined.

rljacob:
You don't have to worry about ERS_D.ne4_oQU240.F2010.pm-cpu_nvidia.eam-hommexx .  Its been failing for months.  Possibly has never passed.

PR #5890, opened 2023-08-18
suppress log messages added by v3 features

whannah1:
Suppress the "imp_sol" chemistry implicit solver warnings and "chlorine_loading_advance" messages. These messages were filling up the log files without any useful information, so this PR disables them. The "imp_sol" warnings are now controlled with the "do_diag". The imp_sol code was filled with a mix of space and tab characters that made it very difficult to read, so I all tab characters associated with blocks involved the do_diag flag.

### DIFF
--- a/components/eam/chem_proc/procfiles/cam/mo_imp_sol_cache.F90
+++ b/components/eam/chem_proc/procfiles/cam/mo_imp_sol_cache.F90
@@ -153,14 +153,14 @@ Column_loop : &
 !           do_diag = lev == 5 .and. isec == 2 .and. lchnk == 290
             do_diag = .false.
             if( do_diag ) then
-	      write(*,*) ' '
-	      write(*,*) 'imp_sol: lchnk,lev,isec,cols = ',lchnk,lev,isec,cols
-	      write(*,*) ' '
-	      write(*,*) 'imp_sol: lrxt'
-	      write(*,'(1p,4g20.10)') lrxt(1,:)
-	      write(*,*) 'imp_sol: lhet'
-	      write(*,'(1p,4g20.10)') lhet(1,:)
-	    end if
+               write(*,*) ' '
+               write(*,*) 'imp_sol: lchnk,lev,isec,cols = ',lchnk,lev,isec,cols
+               write(*,*) ' '
+               write(*,*) 'imp_sol: lrxt'
+               write(*,'(1p,4g20.10)') lrxt(1,:)
+               write(*,*) 'imp_sol: lhet'
+               write(*,'(1p,4g20.10)') lhet(1,:)
+            end if
 !-----------------------------------------------------------------------      
 !        ... Time step loop
 !-----------------------------------------------------------------------      
@@ -187,10 +187,10 @@ Time_step_loop : &
                   solution(:cols,m) = lsol(:cols,j)
                end do
                if( do_diag ) then
-	         write(*,*) ' '
-	         write(*,*) 'imp_sol: solution'
-	         write(*,'(1p,4g20.10)') solution(1,:)
-	       end if
+                  write(*,*) ' '
+                  write(*,*) 'imp_sol: solution'
+                  write(*,'(1p,4g20.10)') solution(1,:)
+               end if
 !-----------------------------------------------------------------------      
 !        ... Set the iteration invariant part of the function F(y)
 !        ... If there is "independent" production put it in the forcing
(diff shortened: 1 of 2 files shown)


components/eam/chem_proc/procfiles/cam/mo_imp_sol_cache.F90
```diff
@@ -153,14 +153,14 @@ subroutine imp_sol( base_sol, reaction_rates, het_rates, extfrc, delt, &
...
             if( do_diag ) then
-	      write(*,*) ' '
-	      write(*,*) 'imp_sol: lchnk,lev,isec,cols = ',lchnk,lev,isec,cols
-	      write(*,*) ' '
-	      write(*,*) 'imp_sol: lrxt'
-	      write(*,'(1p,4g20.10)') lrxt(1,:)
-	      write(*,*) 'imp_sol: lhet'
-	      write(*,'(1p,4g20.10)') lhet(1,:)
-	    end if
+               write(*,*) ' '
```
rljacob, line 156: Using `write(*,*)` is why these wind up in e3sm.log.  It should be write(iulog,*).  Don't have to do that now.   Looks like there is a lot of those to clean up in the chemistry.

components/eam/chem_proc/procfiles/cam/mo_imp_sol_cache.F90
```diff
@@ -409,56 +409,61 @@ subroutine imp_sol( base_sol, reaction_rates, het_rates, extfrc, delt, &
 !-----------------------------------------------------------------------      
 !   	... Non-convergence
 !-----------------------------------------------------------------------      
-		  fail_cnt = fail_cnt + 1
-		  nstep = get_nstep()
- 	          write(*,'('' imp_sol: Time step '',1p,e21.13,'' failed to converge @ (lchnk,lev,isec,nstep) = '',4i6)') dt,lchnk,lev,isec,nstep
-		  stp_con_cnt = 0
+                  fail_cnt = fail_cnt + 1
+                  nstep = get_nstep()
+                  if( do_diag ) then
```
singhbalwinder, line 414: If we hide this message, we will never know if the solver has converged or not for all future chemistry packages. If we are fairly confident that convergence doesn't matter (which I think should be investigated further), we can make this change in the file under pp_{chemistry_package_name} folder. That way, it will be restricted to only one chemistry package (and hopefully it has been evaluated to be safe).

components/eam/chem_proc/procfiles/cam/mo_imp_sol_cache.F90
whannah1: I'm happy to make that change for this file, just let me know if I should do that or hold off.

components/eam/chem_proc/procfiles/cam/mo_imp_sol_cache.F90
whannah1: Good point, I'll remove the if statement from this.

whannah1: Do we need to know if it fails to converge? It seems that the model is fine to carry on without convergence. If it's a critical thing to know then we can leave those message there. In some of my runs I was getting a lot of them, which was a bit annoying when I needed to scan the log files for the debugging info.

whannah1:
e3sm_developer passes on perlmutter.

components/eam/chem_proc/procfiles/cam/mo_imp_sol_cache.F90
rljacob: go ahead!

components/eam/chem_proc/procfiles/cam/mo_imp_sol_cache.F90
singhbalwinder: I believe these messages appear only in v3. I don't remember seeing these in v1 or v2. @wlin7 and @tangq might know more about it. I believe it is safer to make these changes in `components/eam/src/chemistry/pp_superfast_mam5_resus_mom_soag/mo_imp_sol.F90` file. That way, it will impact just the v3 chemistry package not all the future chemistry packages. This file (`components/eam/chem_proc/procfiles/cam/mo_imp_sol_cache.F90`) is used by the chemistry preprocessor to create new F90 files for new chemistry packages. If we add this change here, all future chemistry packages will have this if condition and we will never know if the equations converged or not.

@wlin7 and @tangq : Could you please confirm if `pp_superfast_mam5_resus_mom_soag/` is the right folder for the v3 chemistry?

tangq: @singhbalwinder made a good point that we don't want to hide all of these messages. On the other hand, writing out warnings for each grid box could be too much. In addition, treating chemistry packages case-by-case is not ideal.

It seems to me that if we write out the regional (not global) sums of non-converge points when `do_diag = false`, it will substantially reduce the warning messages (so not filling up the log file) while providing a rough idea about how serious the non-converge issue is. Based on this information, users can then decide whether or not turning on `do_diag`.

@whannah1 , how is `do_diag` specified? I assume it's false by default. How to change it to true?

tangq (changes requested):

components/eam/chem_proc/procfiles/cam/mo_imp_sol_cache.F90
quantheory: @singhbalwinder I talked with @whannah1 about this on slack, and this convergence message is actually not as much of an issue, because a failure to converge here just means that it has to retry with a smaller time step. It doesn't mean there is necessarily any problem with the solution, just that there's some extra computational cost associated with trying again. However, the other message below does mean that there is an inaccuracy in the solution and probably should not be suppressed.

singhbalwinder: Ok, that makes sense to me. I just don't want to suppress messages that provide us with some insights. Sometimes these messages help us debug issues that can be hard to debug. I also see the point of getting too much noise in the log file. We can do what @tangq suggested and aggregate these messages and just display how many grid cells (if any) were affected by this non-convergence each time step or at a coarser time frequency.

whannah1:
After thinking about this more I'm just gonna close this PR because it seems unclear how we should really handle this. A brief inspection of the preprocessed imp_sol code under components/eam/src/chemistry reveals that most (or maybe all?) of the implementations are nearly identical, so there's a lot of duplicated code. In addition to deciding how to handle the excessive warnings and which ones should be enabled by default, it seems we should really rethink how this code is organized... but I doubt we have the time/resources to really do that, so for now it's probably best to just let these warnings fill up our log files.

I could make another PR to just disable the chlorine messages, but they aren't nearly as problematic, so I'm not going to bother.

rljacob:
Don't let the perfect be the enemy of the good.  The other messages you removed should be removed.  Or is the "fails to converge" message the main offender?

At least move those to the atm log.  That will motivate people to fix it properly.

wlin7:
e3sm_atm_developer reports a surprised diff for SMS.ne4_oQU240.F2010.pm-cpu_intel.eam-cosplite. 249 of 540 fields in cpl.hi differ. No explanation for it. Cdash reported a PASS for the test today.

rljacob:
from this PR?

wlin7:
Yes, with this PR branch.

rljacob:
This hasn't been merged to next so is this testing you did yourself?

whannah1:
@rljacob I think to do this robustly I would need to modify all these files:
```
components/eam/src/chemistry/pp_trop_mozart_soa/mo_imp_sol.F90
components/eam/src/chemistry/pp_trop_mam7/mo_imp_sol.F90
components/eam/src/chemistry/pp_linoz_mam4_resus/mo_imp_sol.F90
components/eam/src/chemistry/pp_linoz_mam3/mo_imp_sol.F90
components/eam/src/chemistry/pp_waccm_mozart_mam3/mo_imp_sol.F90
components/eam/src/chemistry/pp_waccm_ghg/mo_imp_sol.F90
components/eam/src/chemistry/pp_none/mo_imp_sol.F90
components/eam/src/chemistry/pp_waccm_mozart_sulfur/mo_imp_sol.F90
components/eam/src/chemistry/pp_trop_strat_soa/mo_imp_sol.F90
components/eam/src/chemistry/pp_linoz_mam4_resus_soag/mo_imp_sol.F90
components/eam/src/chemistry/pp_trop_mam3/mo_imp_sol.F90
components/eam/src/chemistry/pp_superfast_mam4_resus_mom_soag/mo_imp_sol.F90
components/eam/src/chemistry/pp_linoz_mam4_resus_mom_soag/mo_imp_sol.F90
components/eam/src/chemistry/pp_trop_mam4_mom/mo_imp_sol.F90
components/eam/src/chemistry/pp_linoz_mam4_resus_mom/mo_imp_sol.F90
components/eam/src/chemistry/pp_trop_mozart/mo_imp_sol.F90
components/eam/src/chemistry/pp_trop_strat_mam7/mo_imp_sol.F90
components/eam/src/chemistry/pp_trop_mam4_resus_soag/mo_imp_sol.F90
components/eam/src/chemistry/pp_trop_mozart_mam3/mo_imp_sol.F90
components/eam/src/chemistry/pp_waccm_mozart/mo_imp_sol.F90
components/eam/src/chemistry/pp_superfast_mam5_resus_mom_soag/mo_imp_sol.F90
components/eam/src/chemistry/pp_trop_mam4_resus_mom/mo_imp_sol.F90
components/eam/src/chemistry/pp_trop_mam4_resus/mo_imp_sol.F90
components/eam/src/chemistry/pp_trop_mam9/mo_imp_sol.F90
components/eam/src/chemistry/pp_trop_ghg/mo_imp_sol.F90
components/eam/src/chemistry/pp_trop_strat_mam3/mo_imp_sol.F90
components/eam/src/chemistry/pp_trop_mam4/mo_imp_sol.F90
components/eam/src/chemistry/pp_super_fast_llnl_mam3/mo_imp_sol.F90
components/eam/src/chemistry/pp_trop_bam/mo_imp_sol.F90
components/eam/src/chemistry/pp_super_fast_llnl/mo_imp_sol.F90
```
But, obviously, many of those are never going to be used again... I guess I could only modify the one associated with v3... I'm not sure what to do.

Also, from the discussion it seems that there is not much agreement about which messages should be disabled by default. I like Qi's suggestions about aggregating messages to provide a "regional summary" of the convergence failures to reduce the amount of log messages, but that is outside the scope of the "quick fix" I was aiming for.

wlin7:
Yes, just this branch against the master baseline. I am re-running the test to see if reproducible.

Some notes on the burden of logging this warning: With F2010 on ne30 and pm-cpu_intel, it is logging  about 400K bytes per year. If to limit to masterproc (in atm.log), it would be about 700-800  bytes per year.

wlin7:
@whannah1 , just to note, for v3 config, it is `components/eam/src/chemistry/pp_superfast_mam5_resus_mom_soag/mo_imp_sol.F90`

wlin7:
Rerun of `SMS.ne4_oQU240.F2010.pm-cpu_intel.eam-cosplite` (as a new test) with this branch returns a PASS. Double checked bld/GIT_LOG shows the same codes are used. To keep an eye on this behavior (for master branch in general).

PR #5891, opened 2023-08-20
Fix the calculations of p3 sedimentation fluxes

yunpengshan2014:
Original calculation in the P3 code: the precipitation (rain and snow) flux (sflx for snow ) is summed over the flux at each sub-timestep during one microphysics timestep: sflx=flux_qit_1+ flux_qit_2+ flux_qit_3+… flux_qit_n, where the flux_qit_i (i=1,2,3,…n) represents the snow flux at each sub-timestep for sedimentation. This should not be correct. The correct flux should be the mean value for each microphysics timestep.

Changed calculation in the current version: In E3SM-P3, since the sub-timestep is not a fixed value, the mean flux over the microphysics timestep should be calculated with: sflx=(flux_qit_1dt1+ flux_qit_2dt2+ flux_qit_3dt3+… flux_qit_ndtn)/( dt1+ dt2+ dt3+… dtn), where dti represents sub-timestep length. This is similar to the equation used to calculate the surface precipitation rate.

This PR does not effect the model simulation results, but only change the diagnostic precipitation flux output.

### DIFF
--- a/components/eam/src/physics/p3/eam/micro_p3.F90
+++ b/components/eam/src/physics/p3/eam/micro_p3.F90
@@ -3972,7 +3972,6 @@ subroutine cloud_sedimentation(kts,kte,ktop,kbot,kdir,   &
             call generalized_sedimentation(kts, kte, kdir, k_qxtop, k_qxbot, kbot, Co_max, dt_left, &
                  prt_accum, inv_dz, inv_rho, rho, num_arrays, vs, fluxes, qnr, dt_sub)
             do k = k_qxbot,k_qxtop,kdir
-!                  cflx(k+1) = cflx(k+1) + flux_qx(k)
                   cflx(k+1) = cflx(k+1) + flux_qx(k)*dt_sub
             enddo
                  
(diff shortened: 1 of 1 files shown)


crterai:
@yunpengshan2014, would you modify the title of this PR so that it follows the 'Verb object' format? You can see how other PRs are named to get an idea.

wlin7:
The following variables are affected, but not in default eam history output
`precip_liq_flux, precip_ice_flux, ice_flux, rain_flux, cldliq_flux` They would be enabled when `micro_tend_output = .true.`

wlin7 (approved):
The changes look reasonable and consistent, correctly accounting for difference in dt_sub during substepping to obtain an average over the full microphysics time step.

rljacob:
@wlin7 is this ready?

rljacob:
satus:  ready to merge before the v3alpha03 tag.

wlin7:
Merged to next.

wlin7:
@yunpengshan2014 , this PR appears to affect COSP simulator output (e.g., CALIPSO). Can you please confirm it is expected?

yunpengshan2014:
Yes, this is expected.
Regards,
Yunpeng

wlin7:
Thank you, Yunpeng. All is good now.

PR #5892, opened 2023-08-23
fix for submesoscale parametrization in standalone

scalandr:
At the moment, the submesoscale parametrization does not work properly in MPAS-O standalone, that is even if 
`config_submesoscale_enable = .true.`
the parametrization is not active. To solve this issue, the subroutine `ocn_submesoscale_compute_velocity()` has to called in `src/mode_forward/mpas_ocn_forward_mode.F` .

[BFB]

### DIFF
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_forward_mode.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_forward_mode.F
@@ -711,8 +711,12 @@ module ocn_forward_mode
 
            call ocn_eddy_compute_mixed_layer_depth(statePool, forcingPool)
            if (config_use_GM .or. config_submesoscale_enable) then
-               call mpas_timer_start("submesoscale eddy velocity compute", .false.)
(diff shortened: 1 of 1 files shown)


vanroekel:
Thanks for this fix @scalandr -- Before Review - @jonbob do we need an issue to be put up first?  This is an easy review.  it is the definite right fix for an issue in mpas stand alone.

xylar:
@vanroekel, MPAS-Ocean standalone bugs shouldn't need E3SM issues.

vanroekel:
Thanks for the clarification @xylar that makes a lot of sense.  @jonbob should this also include the BFB tag as it won’t affect E3SM or is the stand alone tag sufficient?

rljacob:
@vanroekel please approve.

jonbob:
@vanroekel -- sorry, I missed this. It should be labeled BFB but not a bug. Let me fix that and please approve whenever you're comfortable.

vanroekel (approved):
approved by visual inspection

jonbob:
merged to master

PR #5896, opened 2023-08-27
For maint-2.1, add intel compiler to pm-cpu.

mahf708:
Also update module versions to match those on master for pm-cpu/pm-gpu.

--

Cherry-picks https://github.com/E3SM-Project/E3SM/pull/5815 with a minor conflict resolution.


rljacob:
no it was an oversight.  fixed.

ndkeen:
What tests did you run with these changes?

ndkeen:
If having intel is important for this branch, we can get in. Typically would want all the e3sm_integration tests to pass, which can sometimes lead to quirks between the versions of branches. It may suffice for now to have the compiler in there to even start testing as it must have worked for at least some case you tried.

Issue #5899, opened 2023-08-28
On perlmutter, when file not present on server inputdata, the auto-download command may claim success and simply write a 0-sized file


ndkeen:
I've noticed a few 0-size partition files at NERSC and I can reproduce by simply asking for odd-number partition file.

I see this in submit output:
```
Model mpassi missing file graph127 = '/global/cfs/cdirs/e3sm/inputdata/ice/mpas-seaice/oEC60to30v3/partitions/mpas-seaice.graph.info.230424.part.127'
Trying to download file: 'ice/mpas-seaice/oEC60to30v3/partitions/mpas-seaice.graph.info.230424.part.127' to path '/global/cfs/cdirs/e3sm/inputdata/ice/mpas-seaice/oEC60to30v3/partitions/mpas-seaice.graph.info.230424.part.127' using WGET protocol.
SUCCESS
```

but then:
```
login38% lr /global/cfs/cdirs/e3sm/inputdata/ice/mpas-seaice/oEC60to30v3/partitions/mpas-seaice.graph.info.230424.part.127
-rw-rw-r--+ 1 ndk ndk 0 Aug 28 13:27 /global/cfs/cdirs/e3sm/inputdata/ice/mpas-seaice/oEC60to30v3/partitions/mpas-seaice.graph.info.230424.part.127
```

ndkeen:
I think we've figured out issue -- or actually, how to avoid it -- still not sure why this happens.

Easier reproducer:
```
login22% rm o
login22% rm /global/cfs/cdirs/e3sm/inputdata/ice/mpas-seaice/oEC60to30v3/partitions/mpas-seaice.graph.info.230424.part.127
rm: cannot remove '/global/cfs/cdirs/e3sm/inputdata/ice/mpas-seaice/oEC60to30v3/partitions/mpas-seaice.graph.info.230424.part.127': No such file or directory
login22% wget https://portal.nersc.gov/project/e3sm/inputdata/ice/mpas-seaice/oEC60to30v3/partitions/mpas-seaice.graph.info.230424.part.127 -nc --output-document o
--2023-08-28 13:51:55--  https://portal.nersc.gov/project/e3sm/inputdata/ice/mpas-seaice/oEC60to30v3/partitions/mpas-seaice.graph.info.230424.part.127
Resolving portal.nersc.gov (portal.nersc.gov)... 128.55.206.109, 128.55.206.106, 128.55.206.111, ...
[...]

PR #5900, opened 2023-08-28
Remove NERSC inputdata server

ndkeen:
We were using NERSC as inputdata server while LCRC was down.
However, running into issues with 0-sized files written.
No need to have NERSC server as backup now.

Fixes https://github.com/E3SM-Project/E3SM/issues/5899
[bfb]


rljacob (approved):

ndkeen:
merged to next

PR #5905, opened 2023-08-31
Add MACv2 aerosol for MMF

beharrop:
Adds in the MACv2-SP aerosol scheme for a simplified form of anthropogenic aerosols
that combines with the prescribed natural aerosol.  The current implementation is only
for the E3SM-MMF.  These changes are controlled via atm namelist options, the 
default setting is off for the MACv2 scheme.

[stealth]

### DIFF
--- a/components/eam/src/physics/crm/physpkg.F90
+++ b/components/eam/src/physics/crm/physpkg.F90
@@ -549,6 +549,7 @@ subroutine phys_init( phys_state, phys_tend, pbuf2d, cam_out )
   use modal_aero_wateruptake,only: modal_aero_wateruptake_init
   use nucleate_ice_cam,      only: nucleate_ice_cam_init
   use hetfrz_classnuc_cam,   only: hetfrz_classnuc_cam_init
+  use prescribed_macv2,      only: macv2_rad_props_init
   !-----------------------------------------------------------------------------
   ! Input/output arguments
   !-----------------------------------------------------------------------------
@@ -629,6 +630,9 @@ subroutine phys_init( phys_state, phys_tend, pbuf2d, cam_out )
 
   ! Initialize ocean data
   if (has_mam_mom) call init_ocean_data()
+
+  ! Initialize MACv2-SP aerosols
+  call macv2_rad_props_init()
 
   ! co2 cycle            
   if (co2_transport()) call co2_init()
--- a/components/eam/bld/namelist_files/namelist_definition.xml
+++ b/components/eam/bld/namelist_files/namelist_definition.xml
@@ -5487,6 +5487,25 @@ Rinaldi et al. (JGR, 2013).
 Default:1
 </entry>
 
+<entry id="prescribed_macv2_datapath" type="char*256" input_pathname="abs" category="cam_chem"
+       group="prescribed_macv2_nl" valid_values="" >
+Full pathname of the directory that contains the files specified in
+<varname>prescribed_macv2_filelist</varname>.
+Default: set by build-namelist.
+</entry>
+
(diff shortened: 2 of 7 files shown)


beharrop:
All tests from the e3sm_mmf_integration suite passed on Perlmutter

rljacob:
Pay attention to the circleci build fail.  Probably means the gnu compiler doesn't like something in this branch.

components/eam/src/physics/crm/physpkg.F90
```diff
@@ -629,6 +630,9 @@ subroutine phys_init( phys_state, phys_tend, pbuf2d, cam_out )
   ! Initialize ocean data
   if (has_mam_mom) call init_ocean_data()
 
+  ! Initialize MACv2-SP aerosols
+  call macv2_rad_props_init()
+
```
whannah1, line 635: should we add a check for do_macv2sp here?

whannah1: nevermind, I just noticed that you use the flag within that routine, which is ok.

whannah1 (approved):
Excellent work! I love the clean implementation.

rljacob:
@beharrop have you tried building with gnu?  That's probably why circleCI test is failing.

beharrop:
I did run with the GNU compiler for GPU in some of my tests.  Is there a specific log file I should look at to see its complaints?

rljacob:
The circle CI buildlog is here:  https://circleci.com/api/v1.1/project/github/E3SM-Project/E3SM/8502/output/105/0?file=true&allocation-id=650cb456c65c916312130a42-0-build%2FABCDEFGH

Here the error: 
```
/home/circleci/project/components/eam/src/control/runtime_opts.F90:281:7:

    use prescribed_macv2,    only: prescribed_macv2_readnl
       1
Fatal Error: Can't open module file ‘prescribed_macv2.mod’ for reading at (1): No such file or directory
compilation terminated.
```

beharrop:
Ah ha.  Yea, the prescribed_macv2.F90 module is dumped into the physics/crm path, but I guess that means it can only be found when building the MMF version.  I'll need to figure out how to get this namelist ignored when building the model regularly....

whannah1:
@beharrop I'm looking at your branch again now and I think it might be good to move the macv2 file into it's own folder like `eam/src/physics/aerosol_macv2`, what do you think of that @rljacob ? The alternative is to add an ifdef within runtime_opts.F90.

rljacob:
A directory just for one file?  I'd prefer ifdefs.

whannah1:
@beharrop to add ifdefs we should rebase on the current master since PAM is now merged and use 
```
#if defined(MMF_SAMXX) || defined(MMF_PAM)
```
Similar to what we do in `eam/src/control/cam_comp.F90`

beharrop:
Thanks for the suggestion, @whannah1.  I have added the ifdefs and am running some tests now to see if that fixes the issues.  Will report back later with the results.

beharrop:
I have updated this PR with the new code changes.  I ran the e3sm-developer test suite and I still get an error for the `ERS.f09_g16.IELMBC.pm-cpu_intel.elm-simple_decomp` test, but that test also fails on master, so I'm going to say it's unrelated to this PR.  I had to add in some if not defined things so that when rrtmgp is called for the scream tests, it will compile and set do_macv2sp to false.  Those can easily be reworked whenever macv2 is added to scream.

rljacob:
@brhillman this is ready to merge.

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #5906, opened 2023-08-31
New CRM for MMF (PAM)

whannah1:
Integration of new CRM for MMF, referred to as the "portable atmosphere model" or PAM, that uses the P3 and SHOC physics implementation from SCREAM/EAMxx. The new CRM is used by the compsets with "MMF2" in their name. 

While the PAM code is isolated to a submodule, several large changes outside were required to facilitate this integration. The  PAM interface code is a substantial part of this, but there was also new conflicts created by EAMv3 integration efforts which led to the creation of various "dummy" modules to avoid building the conflicting modules, which are unnecessary for MMF.

A new integration is added to test the new CRM => SMS_Ln3.ne4pg2_ne4pg2.F2010-MMF2

A few things to note about the current MMF2 implementation
- When running with PAM on the GPU the use of openMP threads is problematic for reasons that are unclear. This seems fixable, but more work is needed to look into this. For now, running on GPUs without OMP threads is sufficient, and CPU cases are not affected by this problem
- Although some offline PAM tests have used longer time steps, when running within the MMF configuration at ne30 the PAM time step needs to be 4 seconds for stability. A lot of performance can be gained by increasing this time step, but more work is needed to understand the source of the instability. 
- PAM performance also seems highly machine dependent. For example, using 22 nodes on Perlmutter runs 6x faster than 64 nodes on Summit for the same case.

[BFB]


xyuan (approved):

singhbalwinder (approved):
Looks good! I just left a very minor comment.

components/eam/bld/build-namelist
```diff
@@ -1705,13 +1705,15 @@ if ( !($phys eq 'ideal' or $phys eq 'adiabatic') ) {
...
-        }
+        add_default($nl, 'liqcldoptics', 'val' => 'slingo');
+        add_default($nl, 'icecldoptics', 'val' => 'ebertcurry');
+        # if ($cfg->get('MMF_microphysics_scheme') eq 'sam1mom') {
+        #     add_default($nl, 'liqcldoptics', 'val' => 'slingo');
+        #     add_default($nl, 'icecldoptics', 'val' => 'ebertcurry');
+        # } else {
+        #     add_default($nl, 'liqcldoptics', 'val' => 'gammadist');
+        #     add_default($nl, 'icecldoptics', 'val' => 'mitchell');
+        # }
```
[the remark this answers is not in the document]
whannah1: yea, I'll do that. I left them there because I assume we'll switch to those optics schemes when @brhillman has a chance to work on this, but it'll be easy enough to add those lines when we get to that point.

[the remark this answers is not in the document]
whannah1: done.

cime_config/machines/cmake_macros/gnugpu_summit.cmake
```diff
@@ -20,3 +20,5 @@ set(NETCDF_C_PATH "$ENV{NETCDF_C_PATH}")
 set(NETCDF_FORTRAN_PATH "$ENV{NETCDF_FORTRAN_PATH}")
 set(PNETCDF_PATH "$ENV{PNETCDF_PATH}")
 set(USE_CUDA "TRUE")
+string(APPEND KOKKOS_OPTIONS " -DKokkos_ARCH_POWER9=On -DKokkos_ARCH_VOLTA70=On -DKokkos_ENABLE_CUDA=On -DKokkos_ENABLE_CUDA_LAMBDA=On")
```
rljacob, line 23: Why is PAM/MMF modifying Kokkos opts?

rljacob (approved):

xyuan:
For summit GPU runs, it needs to build Kokkos with GPU support, the default Kokkos build doesn’t have GPU support.

From: Robert Jacob ***@***.***>
Date: Tuesday, September 12, 2023 at 10:36 AM
Cc: Yuan, Xingqiu ***@***.***>, Review requested ***@***.***>
Subject: Re: [E3SM-Project/E3SM] New CRM for MMF (PAM) (PR #5906)

@rljacob commented on this pull request.

In cime_config/machines/cmake_macros/gnugpu_summit.cmake<https://github.com/E3SM-Project/E3SM/pull/5906#discussion_r1323144786>:

> @@ -20,3 +20,5 @@ set(NETCDF_C_PATH "$ENV{NETCDF_C_PATH}")

 set(NETCDF_FORTRAN_PATH "$ENV{NETCDF_FORTRAN_PATH}")

 set(PNETCDF_PATH "$ENV{PNETCDF_PATH}")

 set(USE_CUDA "TRUE")

+string(APPEND KOKKOS_OPTIONS " -DKokkos_ARCH_POWER9=On -DKokkos_ARCH_VOLTA70=On -DKokkos_ENABLE_CUDA=On -DKokkos_ENABLE_CUDA_LAMBDA=On")

Why is PAM/MMF modifying Kokkos opts?

components/eam/src/physics/crm/dummy_modules/debug_info.F90
```diff
@@ -0,0 +1,5 @@
+module debug_info
+!-------------------------------------------------------------------------------
+! Dummy module to override src/physics/cam/debug_info.F90
+!-------------------------------------------------------------------------------
+end module debug_info
```
rljacob, line 5: Why is github showing a red circle with a line through it at the end of the file?  Is it missing a newline at the end of the last line?

components/eam/bld/config_files/definition.xml
```diff
@@ -348,12 +348,15 @@ Switch to enable or disable ECPP in CAM: 0=off, 1=on.
...
 </entry>
-<entry id="MMF_microphysics_scheme" valid_values="sam1mom,m2005" value="sam1mom">
+<entry id="MMF_microphysics_scheme" valid_values="sam1mom,m2005,p3" value="sam1mom">
 MMF CRM cloud microphysics scheme
 </entry>
-<entry id="crm" valid_values="sam,samomp,samxx" value="sam">
+<entry id="crm" valid_values="sam,samomp,samxx,pam" value="sam">
 MMF CRM model
 </entry>
+<entry id="pam_dycor" valid_values="awfl,spam" value="spam">
```
jlee1046, line 357: Is there a comment or documentation somewhere that tells difference between awfl and spam ?

components/eam/bld/configure
```diff
@@ -358,9 +362,10 @@ GetOptions(
     "crm_dt=s"                  => \$opts{'crm_dt'},
     "crm_nx_rad=s"              => \$opts{'crm_nx_rad'},
     "crm_ny_rad=s"              => \$opts{'crm_ny_rad'},
-    "MMF_microphysics_scheme=s"     => \$opts{'MMF_microphysics_scheme'},
+    "MMF_microphysics_scheme=s" => \$opts{'MMF_microphysics_scheme'},
```
jlee1046, line 365: what's '=s' for?

components/eam/bld/configure
```diff
@@ -1812,12 +1818,18 @@ if (defined $opts{'use_MMF'}) {
...
     my $crm = $cfg_ref->get('crm');
+    my $pam_dycor = $cfg_ref->get('pam_dycor');
     my $MMF_microphysics_scheme = $cfg_ref->get('MMF_microphysics_scheme');
     my $crm_adv = $cfg_ref->get('crm_adv');
     my $yes3Dval = 1;
     if ($crm_ny eq 1) {$yes3Dval = 0;}
     if (defined $opts{'crm_adv'}) { $cfg_cppdefs .= " -D_$crm_adv "; }
-    $cfg_cppdefs .= " -D$MMF_microphysics_scheme -DYES3DVAL=$yes3Dval";
+    $cfg_cppdefs .= " -DYES3DVAL=$yes3Dval";
+    if ($MMF_microphysics_scheme eq 'sam1mom') {
```
jlee1046, line 1828: this maybe a style difference so you may leave it as it is. I would combine if & elsif statement into one since $cfg_cppdefs .= "-D$MMF_microphysics_scheme" is the same.

components/eam/bld/configure
```diff
@@ -2660,10 +2676,10 @@ sub write_filepath
             die "configure ERROR: invalid option - crm_adv: $crm_adv.\n";
           }
                               
-        } elsif ($crm eq 'samxx') {
+        } elsif (($crm eq 'samxx') || ($crm eq 'pam')) {
```
jlee1046, line 2679: is this intended to be an empty in elsif body?

components/eam/bld/config_files/definition.xml
```diff
@@ -348,12 +348,15 @@ Switch to enable or disable ECPP in CAM: 0=off, 1=on.
 <entry id="use_crm_cldfrac" valid_values="0,1" value="0">
 Switch to enable or disable fractional cloudiness in m2005 microphysics in CRM: 0=off, 1=on.
 </entry>
-<entry id="MMF_microphysics_scheme" valid_values="sam1mom,m2005" value="sam1mom">
+<entry id="MMF_microphysics_scheme" valid_values="sam1mom,m2005,p3" value="sam1mom">
```
jlee1046, line 351: m2005 is not valid option. Don't you want to delete it?

components/eam/cime_config/config_component.xml
```diff
@@ -68,12 +68,14 @@
...
-      <value compset="_EAM%.*MMF"      >-use_MMF -crm samxx -nlev 60 -crm_nz 50</value>
-      <value compset="_EAM%.*MMF"      >-crm_dt 10 -crm_dx 2000 -crm_nx 64 -crm_ny 1 </value>
-      <value compset="_EAM%.*MMF"      >-crm_nx_rad 4 -crm_ny_rad 1 -rad rrtmgp -rrtmgpxx</value>
+      <value compset="_EAM%.*MMF1"     >-crm samxx -crm_dt 10 </value>
+      <value compset="_EAM%.*MMF2_"    >-crm pam -pam_dycor spam  -crm_dt 4 </value>
+      <value compset="_EAM%.*MMF2-AWFL">-crm pam -pam_dycor awfl  -crm_dt 4 </value>
+      <value compset="_EAM%.*MMF"      >-use_MMF -nlev 60 -crm_nz 50</value>
+      <value compset="_EAM%.*MMF"      >-crm_dx 2000 -crm_nx 64 -crm_ny 1 </value>
       <value compset="_EAM%.*MMF1"     >-MMF_microphysics_scheme sam1mom -chem none</value>
-      <value compset="_EAM%.*MMF"      >-use_MMF_VT</value>   <!-- CRM variance transport -->
```
jlee1046, line 75: do use_MMF_VT and use_MMF_ESMT need to be set in the cam_config_option additionally when using MMF1?

components/eam/cime_config/config_component.xml
```diff
@@ -141,10 +143,10 @@
...
       <value compset="_EAM%SCREAM-HR-AQUA"              >aquaplanet_SCREAM-HR</value>
       <!-- MMF / Super-Parameterization -->
-      <value compset="20TR(?:SOI)?_EAM%.*MMF1"          >20TR_MMF-1mom_CMIP6</value>
-      <value compset="1950(?:SOI)?_EAM%.*MMF1"          >1950_MMF-1mom_CMIP6</value>
-      <value compset="2010(?:SOI)?_EAM%.*MMF1"          >2010_mmf_1mom</value>
-      <value compset="2010(?:SOI)?_EAM%.*MMF2"          >2010_mmf_2mom</value>
+      <!-- TODO: revisit the MMF2 use case when aerosols are not prescribed -->
+      <value compset="20TR(?:SOI)?_EAM%.*MMF"           >20TR_MMF-1mom_CMIP6</value>
+      <value compset="1950(?:SOI)?_EAM%.*MMF"           >1950_MMF-1mom_CMIP6</value>
+      <value compset="2010(?:SOI)?_EAM%.*MMF"           >2010_mmf_1mom</value>
```
jlee1046, line 149: any reason to have mmf in 2010_mmf_1mom as lowercase?

components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -323,12 +341,12 @@ subroutine crm_physics_init(state, pbuf2d, species_class)
    use phys_control,          only: phys_getopts, use_gw_convect
    use phys_grid,             only: get_ncols_p
    use crm_history,           only: crm_history_init
-   use cam_history,    only: addfld
-   use constituents,   only: apcnst, bpcnst, cnst_name, cnst_longname, cnst_get_ind
+   use cam_history,           only: addfld, hist_fld_active
+   use constituents,          only: apcnst, bpcnst, cnst_name, cnst_longname, cnst_get_ind
+   use constituents,          only: pcnst, cnst_get_ind
```
jlee1046, line 346: Doesn't this line overwrite line 345?

components/eam/src/physics/crm/dummy_modules/debug_info.F90
whannah1: yea, missing newline, I'll fix that.

components/eam/bld/config_files/definition.xml
whannah1: No, but I should definitely add something somewhere. I don't really intend for this to be something that anyone would change, but I still think it will be useful to compare the two dycors even if we only really use one. We don't seem to have any good documentation of the dycor options in the PAM repo, so I'll add a comment in this file to clarify the difference.

components/eam/bld/configure
whannah1: yes, just to hold the comment about not needing this section. I guess it's kinda silly to add the else if just for this though. Not sure why it was done this way originally. I'll change it.

components/eam/bld/config_files/definition.xml
whannah1: yea, I'll do that.

components/eam/cime_config/config_component.xml
whannah1: Good catch. It actually doesn't matter right now since it's not implemented in PAM yet, but it's worth disabling it for PAM cases so I'll update this line.

components/eam/src/physics/crm/crm_physics.F90
whannah1: yea, we still access all the same module variables, so this is ok.

components/eam/bld/configure
singhbalwinder: In Perl, it means that this argument expects a string argument.

rljacob:
@jlee1046 Have your concerns been addressed?

jlee1046:
> @jlee1046 Have your concerns been addressed?

I was on vacation and haven't finished going through all the changes. I plan on finish reviewing in a day or two.

components/eam/src/physics/crm/pam/pam_debug.h
```diff
@@ -0,0 +1,211 @@
...
+    int n = 0;
+    printf("PAM-DEBUG %d - k:%d  i:%d  temp : %g  rv: %g  rc: %g  ri: %g \n",
+      id,k,i,
+      temp(k,j,i,n),
+      rho_v(k,j,i,n),
+      rho_c(k,j,i,n),
+      rho_i(k,j,i,n)
+    );
+  });
+}
```
jlee1046, line 211: You may want to add a new line at the end of the code to get rid of that red circle with a bar

jlee1046 (approved):
All look good to me.

whannah1:
I pushed some major recent updates that appear to have addressed many of the problems I have been seeing in global simulations. Here's a summary of the major things I changed:

- fixed a problem with the conversion between temperature and potential temperature in the PAM-C dycor which inadvertently imposed the water vapor forcing into the CRM temperature response - I think this is the most important change
- Added a hyperdiffusion calculation to account for the missing horizontal turbulence mixing - hopefully we can replace this such that SHOC handles the horizontal mixing
- Switched to gradually forcing the dry density instead of instantaneously reseting it at the beginning of each CRM call
- Switched from using adding the total water forcing to the vapor field and adjusting the temperature to account for the phase conversion
- Switched the water species to force the mixing ratio and convert back to density to be consistent with how the GCM feedback tendencies are formulated and, additionally, account for gradually forced dry density
- Added forcing for hydrometeor number concentrations

components/eam/src/physics/crm/pam/pam_debug.h
whannah1: done.

components/eam/cime_config/config_component.xml
whannah1: FYI re-enabled VT and ESMT only for MMF1/SAM option

components/eam/bld/configure
whannah1: done

components/eam/bld/config_files/definition.xml
whannah1: done

brhillman:
Merged to next.

PR #5910, opened 2023-09-02
Accumulate data land-ice mass fluxes

### LINKED ISSUE #5911: ` accumulatedLandIceMass` not being computed in `DISMF` runs
When we use data ice-shelf melt fluxes (DISMF), we are not currently accumulating the total land-ice mass flux.  This is a problem because this total is needed to close the mass budget in E3SM.

xylar:
Before this merge, we were only accumulating this flux in prognostic flux mode.  With this merge, we accumulate the mass flux in data mode as well, as this term is needed in the total mass budget of the model.

To facilitate this and reduce issues related to initialization vs. forward time stepping, this accumulation step has been broken into its own subroutine, `ocn_surface_land_ice_fluxes_accumulate_fluxes()`. This subroutine is called when time-stepping the model but not at initialization (since we do not want to accumulate any fluxes at that time).

Fixes #5911

### DIFF
--- a/components/mpas-ocean/driver/ocn_comp_mct.F
+++ b/components/mpas-ocean/driver/ocn_comp_mct.F
@@ -675,7 +675,7 @@ contains
         call ocn_forcing_build_fraction_absorbed_array(meshPool, statePool, forcingPool, ierr, 1)
         call mpas_timer_start("land_ice_build_arrays", .false.)
         call ocn_surface_land_ice_fluxes_build_arrays(meshPool, forcingPool, scratchPool, &
-                                                      statePool, dt=0.0_RKIND, err=ierr)
+                                                      statePool, err=ierr)
         call mpas_timer_stop("land_ice_build_arrays")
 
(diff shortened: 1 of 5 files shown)


xylar:
@jonbob, this is somewhere in the gray area between clean-up and a bug fix.  I can add an issue if you think it should be marked as a bug fix.

I expect it to be bit-for-bit except for runs in DISMF mode, and only with those runs if the `timeSeriesStatsGlobal` are part of the checked output.  Let me know if such at test exists, in which case we need to mark this as non-BFB.

mark-petersen (approved):
This falls into the category of additional reporting, and should definitely be bit-for-bit. I agree with the purpose, and reviewed by visual inspection. Thanks. I also did a quick compile test with gnu and intel on chrysalis using mpas-ocean stand-alone.

rljacob:
@darincomeau please review.

darincomeau (approved):
Approved by visual inspection - sorry for the delay!

jonbob:
passes:

merged to next

jonbob:
merged to master

xylar:
@jonbob Thanks so much!

PR #5915, opened 2023-09-05
Verify array is allocated before accessing size attribute

tcclevenger:
In EAMxx, COSP is called without allocating `cospOUT%isccp_meantb`. Every instance in EAM's `cosp.F90` we ask `associated(cospOUT%isccp_meantb)` before accessing the array, except here where we ask for it's size, leading to a valgrind error. Solution is just to ask `associated(cospOUT%isccp_meantb)` before asking for the array size.

[BFB]


brhillman (approved):
Thanks for putting this in @tcclevenger this looks good to me!

brhillman:
Merged to next.

PR #5916, opened 2023-09-06
Changes redi tapering and slopes

vanroekel:
In current MPAS slopes are altered directly, e.g. when S > Scrit it is set to zero.  Griffies et al 1998 (Appendix C) equates this to slope clipping.  The appropriate setup is to taper the redi kappa coefficient for each individual flux instead.  The cross terms ("term2" and "term 3") end up being identical as they are d/dz(kappa S grad phi) so  tapering kappa or S are identical.  However for the vertical it is d/dz(kappa |S|^2 d phi / dz) so tapering S is different from tapering kappa.  Also term 1 (laplacian diffusion) was not being tapered with slopes creating a cross isopycnal flux in steeply sloping regions.  These issues are fixed here.

In addition a new slope calculation is introduced to prevent the slopes from getting very large when grad phi or d/dz(phi) get small.  This function limits the slope to be between -1 and 1.

### DIFF
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -3008,10 +3008,6 @@
 			description="spatially and depth varying GM kappa.  The scaling is based on the Brunt Vaisala Frequency relative to a maximum value below the mixed layer, follows from Danabasoglu and Marshall 2007.  If config_GM_closure is not set to N2_dependent the scaling value is set to 1 everywhere."
 			packages="gm"
 		/>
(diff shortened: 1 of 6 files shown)


vanroekel:
@mark-petersen @jonbob @maltrud @alicebarthel this is the first of two PRs I mentioned on the coupled call today.  I have added the do not merge label for now as I haven't tested the splitting of the PRs yet.  But wanted to post this for initial examination or testing if anyone wants to try it.

One other note, my testing to this point has included #

vanroekel:
opened in wrong place meant to do a PR to ocean discussion.

PR #5919, opened 2023-09-07
Adding support for SCORPIO legacy fortran interface

jayeshkrishna:
The Fortran interface implementation in SCORPIO was recently refactored but
users still have the option to use the legacy/old Fortran interface implementation.

Adding support in CIME for the SCORPIO legacy Fortran interface

[BFB]


jayeshkrishna:
The refactored Fortran interface implementation is not available in the most recent release of SCORPIO and will be included in an upcoming release of SCORPIO.

Also see https://github.com/E3SM-Project/scorpio/issues/531

dqwu (approved):

PR #5921, opened 2023-09-08
Reduce number of sea ice flux iterations from 10 to 5.

erinethomas:
The current number of iterations to calculate sea ice fluxes is excessive. Convergence of atm-sea ice fluxes is reached within a few iterations, therefore, the default value for the number of iterations should be reduced from 10 to 5.

[non-BFB]

### DIFF
--- a/components/mpas-seaice/bld/namelist_files/namelist_defaults_mpassi.xml
+++ b/components/mpas-seaice/bld/namelist_files/namelist_defaults_mpassi.xml
@@ -397,7 +397,7 @@
 <config_calc_surface_temperature>true</config_calc_surface_temperature>
 <config_use_form_drag>false</config_use_form_drag>
 <config_use_high_frequency_coupling>true</config_use_high_frequency_coupling>
-<config_boundary_layer_iteration_number>10</config_boundary_layer_iteration_number>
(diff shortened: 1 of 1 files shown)


erinethomas:
two figures showing T*, U*, Q* convergence as function of number of loop iterations when reducing from 10 to 5: 
convergence_0001-01-01_05_30_00

convergence_1_0001-01-01_05:30:00

eclare108213 (approved):

proteanplanet (approved):
A worthwhile economy.

erinethomas:
This change has very little impact on the MPAS-SI performance.

Comparing a 1 year D-Case (MPAS-SI only) run with 10 vs 5 iterations: 

Results for 10 iterations: 
PACE : https://pace.ornl.gov/exp-details/157265   
MPAS-SI throughput = 16.64 SYPD 

Results for 5 iterations: 
PACE: https://pace.ornl.gov/exp-details/157929
MPAS-SI throughput = 16.48 SYPD

rljacob:
status:  will put this in with IcePack.

jonbob:
Passes sanity testing and shows expected DIFFs and NML DIFFs. Merged to next

jonbob:
merged to master -- expected NML and regular DIFFs will be blessed as part of PR #5999, which went in at the same time

PR #5923, opened 2023-09-11
MPAS: Change common lib from OBJECT to STATIC

### LINKED ISSUE #4903: nvlink multiple definition of device kernel
In a case created with
```
./cime/scripts/create_test --compiler pgigpu PFS_P2560.T62_oRRS18to6v3.GMPAS-IAF.summit_pgigpu.bench-gmpas_noio
```
getting
```
nvlink error   : Multiple definition of 'mpas_vector_reconstruction_mpas_reconstruct_1d_gpu_647_gpu' in '../../mpas-framework/src/libocn.a:mpas_vector_reconstruction.f90.o', first defined in '../../mpas-framework/src/libice.a:mpas_vector_reconstruction.f90.o'
nvlink error   : Multiple definition of 'mpas_vector_reconstruction_mpas_reconstruct_1d_gpu_633_gpu' in '../../mpas-framework/src/libocn.a:mpas_vector_reconstruction.f90.o', first defined in '../../mpas-framework/src/libice.a:mpas_vector_reconstruction.f90.o'
nvlink error   : Multiple definition of 'mpas_vector_reconstruction_mpas_reconstruct_1d_gpu_608_gpu' in '../../mpas-framework/src/libocn.a:mpas_vector_reconstruction.f90.o', first defined in '../../mpas-framework/src/libice.a:mpas_vector_reconstruction.f90.o'
nvlink error   : Multiple definition of 'mpas_vector_reconstruction_mpas_reconstruct_2d_gpu_531_gpu' in '../../mpas-framework/src/libocn.a:mpas_vector_reconstruction.f90.o', first defined in '../../mpas-framework/src/libice.a:mpas_vector_reconstruction.f90.o'
nvlink error   : Multiple definition of 'mpas_vector_reconstruction_mpas_reconstruct_2d_gpu_517_gpu' in '../../mpas-framework/src/libocn.a:mpas_vector_reconstruction.f90.o', first defined in '../../mpas-framework/src/libice.a:mpas_vector_reconstruction.f90.o'
nvlink error   : Multiple definition of 'mpas_vector_reconstruction_mpas_reconstruct_2d_gpu_492_gpu' in '../../mpas-framework/src/libocn.a:mpas_vector_reconstruction.f90.o', first defined in '../../mpas-framework/src/libice.a:mpas_vector_reconstruction.f90.o'
nvlink fatal   : merge_elf failed
pgacclnk: child process exit status 2: /autofs/nccs-svm1_sw/summit/nvhpc_sdk/rhel8/Linux_ppc64le/21.11/compilers/bin/tools/nvdd
```
that's built once (as a "common" source) but included twice: one through libice.a, another through libocn.a .
It appears the recommendation is to create a different library for MPAS-framework gpu kernels and link it separately.

jgfouca:
Fixes GPU multiple def errors.

Fixes #4903 

[BFB]

### DIFF
--- a/components/mpas-framework/src/CMakeLists.txt
+++ b/components/mpas-framework/src/CMakeLists.txt
@@ -84,7 +84,7 @@ set(COMMON_RAW_SOURCES external/ezxml/ezxml.c)
 include(${CMAKE_CURRENT_SOURCE_DIR}/framework/framework.cmake)
 include(${CMAKE_CURRENT_SOURCE_DIR}/operators/operators.cmake)
 
-add_library(common OBJECT)
+add_library(common)
 target_compile_definitions(common PRIVATE ${CPPDEFS})
 target_include_directories(common PRIVATE ${INCLUDES})
 
--- a/components/mpas-framework/src/build_core.cmake
+++ b/components/mpas-framework/src/build_core.cmake
@@ -30,7 +30,7 @@ function(build_core CORE)
   add_library(${COMPONENT})
   target_compile_definitions(${COMPONENT} PRIVATE ${CPPDEFS})
   target_include_directories(${COMPONENT} PRIVATE ${INCLUDES})
-  target_link_libraries(${COMPONENT} PUBLIC ${LIBRARIES})
+  target_link_libraries(${COMPONENT} PUBLIC ${LIBRARIES} common)
 
   # Make .inc files
   add_custom_command (
@@ -63,6 +63,6 @@ function(build_core CORE)
   endif()
 
   genf90_targets("${RAW_SOURCES}" "${INCLUDES}" "${CPPDEFS}" "${NO_PREPROCESS}" "${INC_DIR}")
-  target_sources(${COMPONENT} PRIVATE ${SOURCES} $<TARGET_OBJECTS:common>)
+  target_sources(${COMPONENT} PRIVATE ${SOURCES})
 
 endfunction(build_core)


jgfouca:
@mark-petersen , I'm having some trouble finding exactly where in the Git history of MPAS I made the change to the `common` library to make it an OBJECT library. The only thing I see in `git@github.com:MPAS-Dev/MPAS-Model.git` is this commit:

```
commit 7ff78bb00a22dcb38e0f47e49df1b891deb44438
Author: Mark Petersen <mpetersen@lanl.gov>
Date:   Fri Nov 1 07:49:22 2019 -0600

    Add cmake files to framework
```

... which just adds the entire `src/CMakeLists.txt` file. I was hoping, if I could fix the actual commit that made `common` an `OBJECT` library, I could figure out why I made that change.

mark-petersen:
Does this help? 
```
git blame -- components/mpas-framework/src/CMakeLists.txt |grep -n common
81:bcf58251dab src/CMakeLists.txt                           (James Foucar      2019-11-06 10:22:40 -0700 81) # Gather sources that are needed for all cores into "common" library
87:7ff78bb00a2 src/CMakeLists.txt                           (Mark Petersen     2019-11-01 07:49:22 -0600 87) add_library(common OBJECT)
```

jgfouca:
@mark-petersen , i'm looking at `bcf58251dab` and the library was already set as an OBJECT library. I'd like to look at the git history before 7ff78bb00a2 but I don't know how.

jgfouca:
I assume I need to look at some older branch, whatever the MPAS submodule used to be.

mark-petersen:
I think you are looking for this commit: https://github.com/MPAS-Dev/MPAS-Model/pull/386

mark-petersen:
Here is the specific line in question. 

https://github.com/MPAS-Dev/MPAS-Model/pull/386/files#diff-148715d6ea0c0ea0a346af3f6bd610d010d490eca35ac6a9b408748f7ca9e3f4R77

This is from before MPAS was moved into the E3SM repo.

jgfouca:
I'm still having trouble finding the commit that actually made the change. All I see are commits that added or moved the entire file. Unfortunately, all branches on my MPAS fork were deleted long ago. I think having the `common` library be a real library instead of collection of object files is definitely an improvement, so if the history is lost, I think we should just move forward with this change unless a problem emerges.

jgfouca:
Yeah, it's possible I made the entire initial effort a single commit, in which case there'd be no specific commit to look at. I am really kicking myself for at least not leaving a comment or something to describe why I thought an OBJECT library was needed. I have "learned" this lesson the hard way so many times...

philipwjones (approved):
Approve based on visual inspection. This is definitely the behavior we want, rather than an object lib. I don't have any recollection of the reasons for this choice in the past - maybe trying to preserve some flexibility for each individual MPAS component?

jgfouca:
Merged to next.

PR #5926, opened 2023-09-13
Fix linoz data file for 1pctCO2 use_case

wlin7:
PR #5860 updated linoz data files for use cases. But the PR did not involve the use_case file for WCYCL1850-1pctCO2 
as it was introduced at a later time.

This fixes WCYCL1850-1pctCO2 tests, though some machines were able to run with the old file.

[non-BFB] only for compy and anvil prod tests that were able to run with the old file


jinboxie (approved):
Yes. It is supposed to be using the new file rather than the old.It is now modified correctly.

wlin7:
Merged to next.

This should fix `SMS_Ld1.ne30pg2_EC30to60E2r2.WCYCL1850-1pctCO2.chrysalis_intel.allactive-wcprod_1850_1pctCO2` failure. It will also override this test on compy and anvil which were able to run with the old linoz data file. Bless requests already submitted for compy and anvil prod test suites. To merge the PR to master once any of the compy/anvil/chrysalis prod tests returns an expected diff. This should avoid repeating blessing the test for compy/anvil.

wlin7:
Merged to master.

PR #5928, opened 2023-09-15
CMake build system: Fix SourceMods rebuild detection

### LINKED ISSUE #5924: SourceMods for code changes stops working
It is tested specifically for EAM, using SMS_D_Ln5.ne4_oQU240.F2010.chrysalis_intel.eam-cosplite_nhtfrq5. Any tests involving active atm should see the same. Code changes placed under a case's SourceMods/src.eam are not used to build the model. The problem exists with the current master.

Are there changes in the build system recently that acts to disable SourceMods mechanism?

jgfouca:
Prior to this change, if you went to an already-built case and made a SourceMod change, then went to the build dir and re-ran 'make', the SourceMod would not be picked up. On some systems, even re-running case.build would not pick up the change.

Fixes #5924 

[BFB]

### DIFF
--- a/components/CMakeLists.txt
+++ b/components/CMakeLists.txt
@@ -104,6 +104,16 @@ elseif(USE_HIP)
   enable_language(HIP)
 endif()
 
+# Any changes to SourceMods will require us to reconfigure
+file(GLOB COMPONENT_SOURCE_MOD_DIRS "${CASEROOT}/SourceMods/src.*")
+foreach(COMPONENT_SOURCE_MOD_DIR IN LISTS COMPONENT_SOURCE_MOD_DIRS)
+  set_property(
+    DIRECTORY
+    APPEND
+    PROPERTY CMAKE_CONFIGURE_DEPENDS
+    ${COMPONENT_SOURCE_MOD_DIR})
+endforeach()
+
 # Include function definitions
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/cmake_util.cmake)
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/build_mpas_model.cmake)


jgfouca:
As a side note, I added some SourceMods documentation to https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/1011942171/E3SM+Build+System+and+CMake . This feature had no documentation there.

components/CMakeLists.txt
```diff
@@ -104,6 +104,16 @@ elseif(USE_HIP)
...
+# Any changes to SourceMods will require us to reconfigure
+file(GLOB COMPONENT_SOURCE_MOD_DIRS "${CASEROOT}/SourceMods/src.*")
+foreach(COMPONENT_SOURCE_MOD_DIR IN LISTS COMPONENT_SOURCE_MOD_DIRS)
+  set_property(
+    DIRECTORY
+    APPEND
+    PROPERTY CMAKE_CONFIGURE_DEPENDS
+    ${COMPONENT_SOURCE_MOD_DIR})
+endforeach()
+
```
wlin7, line 116: Thank you @jgfouca . This change does bring back the SourceMods mechanism for model build.
A question on the comment: "`Any changes to SourceMods will require us to reconfigure`"? Can you give some explanation?  Is it the new way of making SourceMods work? My understanding has been that no reconfigure is involved when making changes via SourceMods. The "`reconfigure`" you refer to probably has a different meaning.

jgfouca: @wlin7 , yes, I will try to explain more. With this change the CONFIGURE phase of cmake now depends on the SourceMod dirs for the various components. Any file change within those directories (adding or removing a file) will change the mod timestamp for the directory. This will allow CMake to "see" those kinds of changes and know it needs to re-run itself.

bartgol (approved):

wlin7 (approved):

rljacob (approved):

rljacob:
Still wondering why it was working on Chrysalis and then stopped.  I think we load explicit cmake modules and don't think they were changed in the past month.

jgfouca:
Merged to next.

PR #5933, opened 2023-09-19
For NERSC machines, built a new cprnc with GNU and updated the cprnc path

ndkeen:
For NERSC machines, built a new cprnc with GNU and updated the cprnc path.

Verified cprnc output matches what was discussed in https://github.com/ESMCI/cime/issues/4419

[bfb]


ndkeen:
Tested with e3sm_integration

ndkeen:
merged to next

rljacob (approved):

Issue #5830, opened 2023-07-21
On maint-2.0, need a missing header file in externals/scorpio with Intel compiler


ndkeen:
We recently added Intel compiler to the maint-2.0 branch.
To build with intel, it seems to need 
```
#include "spio_io_summary.h"
```
added to `externals/scorpio/src/clib/pio_lists.c`

jayeshkrishna:
This was an issue that was fixed in SCORPIO master (and recent releases), we might need to upgrade the version of SCORPIO in the maint branch.

ndkeen:
Another issue with the scorpio in maint-2.0 on pm-cpu, is that the output netcdf files do not seem to have lustre file stripe settings.

@mahf708 found that updating the scorpio submodule in maint-2.0 can fix both the header and file stripe issue. I'm not sure if it's otherwise ok to update the submodule.

ndkeen:
I am wanting to make intel the default compiler on this branch.

dhruvakulkarni:
Just a note: Users on perlmutter @ NERSC have also observed this issue while installing E3SM.

ndkeen:
This was fixed by https://github.com/E3SM-Project/E3SM/pull/5983

jinboxie:
Hi Noel, was this updated to the maint2.0 branch?
I used the most recent version of maint2.0 on perlmutter, and the problem persists. 
The commit is as follows:
commit 5a3fe8970ac3a3fdf02bc2692cd6e6a1883e5035 (HEAD)
Author: Peter Schwartz <schwartzpd@ornl.gov>
Date:   Tue Feb 6 12:12:06 2024 -0500

    Changed finidat file for wcprodssp tests.

Luckily I saw this page, and added the header as you indicated and it passed.

rljacob:
@jayeshkrishna please verify that the problem is still fixed on latest maint-2.0

jayeshkrishna:
@jinboxie : I just verified that this fix is available on maint-2.0 (https://github.com/E3SM-Project/E3SM/commit/5a3fe8970ac3a3fdf02bc2692cd6e6a1883e5035) . Did you ensure that you updated the submodules in your branch/code (git submodule sync; git submodule update --init --recursive)?

jinboxie:
Hi Jayeshi @jayeshkrishna . The header problem is due to my old E3SM scorpio library. I tried a completely new E3SM maint-2.0 using your command of update gets the latest SCORPIO version and solves the header problem.
[...]

PR #5939, opened 2023-09-19
Update maint-2.0 for pm-cpu (including making Intel default compiler)

ndkeen:
For pm-cpu:
Use Intel compiler as default
Update modules to match version on master.
Remove references to cori in XML files.
Adjust pelayouts for pm-cpu/gcp.
Use new cprnc
Update NERSC batch settings

### DIFF
--- a/cime_config/allactive/config_pesall.xml
+++ b/cime_config/allactive/config_pesall.xml
@@ -107,7 +107,7 @@
         </nthrds>
       </pes>
     </mach>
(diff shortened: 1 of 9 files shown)


ndkeen:
This changes default compiler to Intel which will hit the build error described here.
https://github.com/E3SM-Project/E3SM/issues/5830
Easy to make the change locally and continue.

Tested with e3sm_developer and e3sm_prod

PR #5940, opened 2023-09-20
Fixing a bug in MOSART-heat

### LINKED ISSUE #5941: A variable related to stream temperature only not initialized properly before using in MOSART_physics_mod.F90
In MOSART_physics_mod.F90, a variable, ha_lateral, needs to be set as zero before being used later in the loops, otherwise the stream temperature in the main channel will be overestimated. This issue will only affect the stream temperature simulations, so will not affect those cases without invoking MOSART-heat. MOSART-heat is only turned on by setting heatflag as TRUE, whilst the default value of heatflag is FALSE.

liho745:
Adding the initialization of a variable used in MOSART-heat; otherwise, the main-channel temperature 
will be overestimated. The change made only affects stream temperature, nothing else. 

Fixes #5941
[BFB]

### DIFF
--- a/components/mosart/src/riverroute/MOSART_physics_mod.F90
+++ b/components/mosart/src/riverroute/MOSART_physics_mod.F90
@@ -180,6 +180,7 @@ MODULE MOSART_physics_mod
 
        call t_startf('mosartr_subnetwork')    
        TRunoff%erlateral(:,:) = 0._r8
+	   if (heatflag) THeat%ha_lateral(:) = 0._r8
        TRunoff%etexchange = 0._r8
        do nt=nt_nliq,nt_nice ! water transport
        if (TUnit%euler_calc(nt)) then
--- a/components/mosart/src/riverroute/MOSART_physics_mod.F90
+++ b/components/mosart/src/riverroute/MOSART_physics_mod.F90
@@ -180,7 +180,7 @@ MODULE MOSART_physics_mod
 
        call t_startf('mosartr_subnetwork')    
        TRunoff%erlateral(:,:) = 0._r8
-	   if (heatflag) THeat%ha_lateral(:) = 0._r8
+       if (heatflag) THeat%ha_lateral(:) = 0._r8
        TRunoff%etexchange = 0._r8
        do nt=nt_nliq,nt_nice ! water transport
        if (TUnit%euler_calc(nt)) then
@@ -994,7 +994,7 @@ MODULE MOSART_physics_mod
        TRunoff%erout(iunit,nt) = -TRunoff%erin(iunit,nt)-TRunoff%erlateral(iunit,nt)
     else
        !TODO. If this channel is at basin outlet (downstream is ocean), use the KW method
-	   if(rtmCTL%mask(iunit) .eq. 3) then 
+       if(rtmCTL%mask(iunit) .eq. 3) then 
           call Routing_KW(iunit, nt, theDeltaT)
        else
           if(nt == nt_nliq) then 
@@ -1230,9 +1230,9 @@ MODULE MOSART_physics_mod
 
     if(nt==nt_nliq) then
         TRunoff%yh(iunit,nt) = TRunoff%wh(iunit,nt) !/ TUnit%area(iunit) / TUnit%frac(iunit) 
-	else
-	    TRunoff%yh(iunit,nt) = 0._r8
-	end if
+    else
+        TRunoff%yh(iunit,nt) = 0._r8
+    end if
 
   end subroutine updateState_hillslope
 


rljacob:
@liho745 please open an issue describing the bug.   Also does this affect any case that uses MOSART?

donghuix:
> @liho745 please open an issue describing the bug. Also does this affect any case that uses MOSART?

@rljacob FYI, ``rof_heat`` is true in default in coupler for e3sm. But MOSART-heat is turned off in default. Because river component doesn't occupy a fraction in the coupler, we cannot run MOSART-heat in a coupled simulation (e.g., land river coupled configuration). Then why ``rof_heat`` is true in default?

donghuix (approved):

hydrotian:
@donghuix My understanding is that `rof_heat=T` allows MOSART to receive fields from the coupler for MOSART-heat feature, even the feature in MOSART is turned off. While I understand that the coupler would create redundant fields, it ensures that the data channels are open for the MOSART-heat if needed.

hydrotian (approved):

liho745:
Will open an issue soon.
It does not affect the other cases using MOSART, unless stream temperature
simulation is invoked by setting heatflag as .true. in user_nl_mosart (its
default value is .false.).

liho745:
> @liho745 please open an issue describing the bug. Also does this affect any case that uses MOSART?

@rljacob Issue #5941 has been opened.

rljacob:
do we have test cases that set heatflag to TRUE?

liho745:
> do we have test cases that set heatflag to TRUE?

Not that I am aware of.

rljacob:
Yes ERS.r05_r05.RMOSGPCC.mosart-heat turns it on.  So answers will change for that test?

rljacob:
Please run the e3sm_mosart_developer test suite.

rljacob:
nevermind. You said you ran land developer which includes those tests.

liho745:
> Yes ERS.r05_r05.RMOSGPCC.mosart-heat turns it on. So answers will change for that test?

On Compy using 28426b9 tag for generating my baseline, this test also passed as below:

PASS ERS.r05_r05.RMOSGPCC.compy_intel.mosart-heat CREATE_NEWCASE
PASS ERS.r05_r05.RMOSGPCC.compy_intel.mosart-heat XML
PASS ERS.r05_r05.RMOSGPCC.compy_intel.mosart-heat SETUP
PASS ERS.r05_r05.RMOSGPCC.compy_intel.mosart-heat SHAREDLIB_BUILD time=165
PASS ERS.r05_r05.RMOSGPCC.compy_intel.mosart-heat NLCOMP
PASS ERS.r05_r05.RMOSGPCC.compy_intel.mosart-heat MODEL_BUILD time=141
PASS ERS.r05_r05.RMOSGPCC.compy_intel.mosart-heat SUBMIT
PASS ERS.r05_r05.RMOSGPCC.compy_intel.mosart-heat RUN time=54
PASS ERS.r05_r05.RMOSGPCC.compy_intel.mosart-heat COMPARE_base_rest
PASS ERS.r05_r05.RMOSGPCC.compy_intel.mosart-heat BASELINE 28426b9: 
PASS ERS.r05_r05.RMOSGPCC.compy_intel.mosart-heat MEMCOMP
PASS ERS.r05_r05.RMOSGPCC.compy_intel.mosart-heat TPUTCOMP
PASS ERS.r05_r05.RMOSGPCC.compy_intel.mosart-heat MEMLEAK
PASS ERS.r05_r05.RMOSGPCC.compy_intel.mosart-heat SHORT_TERM_ARCHIVER_

rljacob:
Do you expect there to be no diffs with a baseline?  If that stream temperature doesn't feed back in to the system, that could explain it.

liho745:
Simulation stream temperature should be different after adding back this line (it was there in my earlier versions). The simulated stream temperature, however, does not feed back to the system via the flux coupler, hence not affecting any other components.

rljacob:
That explains it. Most tests just look at coupler history files for diffs.  We would need to turn on daily mosart output in that test to see any diff.

components/mosart/src/riverroute/MOSART_physics_mod.F90
```diff
@@ -180,6 +180,7 @@ subroutine Euler
 
        call t_startf('mosartr_subnetwork')    
        TRunoff%erlateral(:,:) = 0._r8
+	   if (heatflag) THeat%ha_lateral(:) = 0._r8
```
bishtgautam, line 183: Can you please indent the code and remove additional spaces at the beginning?

bishtgautam: Btw, why not initialize `THeat%ha_lateral(:)` irrespective of `heatflag`?

liho745: Done. I also replaced the other tabs in this file with spaces. 

THeat%ha_lateral(:) has already been initialized as zeros when allocated memory in RtmMod.F90. Here, when heatflag is TRUE, THeat%ha_lateral(:) needs to be set to zeros again each time Euler() is called; when heatflag is FALSE, it will not be invoked so no need to set to zero anymore.

bishtgautam (approved):

rljacob:
@bishtgautam please merge to master.

PR #5944, opened 2023-09-21
CIME submodule update

jgfouca:
... to b10d452d1073d1d06d10a6b3269e9f35a872bb50

Changes
1) ~gpu_enabled needs to be true only when gpu is used~
2) support wildcard for machine and compiler name in testname syntax
3) Add new 'perf' test property: This will be used to mark suites as performance test suites.
4) PFS tests should always save timing

Fixes
1) env_batch: Fix flag/val separator. Fixes submit commands on some platforms
2) Moves check_input_data to SystemTestsCommon when running a test. Should fix
   missing data files when pe-layout changes for tests.
3) ERIO fix: Removes .* between hist_file_extension and suffix in get_all_hist_files.

[BFB]


sarats:
*gpu_enabled needs to be true only when gpu is used
Curious what this means in E3SM context.

* support wildcard for machine and compiler name
Why would we need that? Is this for something like "front*" to match "frontier"? What happens when there is more than one match?

rljacob:
I think the wildcard is to make it easier to start a test with a testmod.   You can do `create_test TESTNAME.*_*.testmod `instead of typing out the machine and compiler.

rljacob (approved):

jgfouca:
@sarats , it looks like the gpu_enabled stuff was only for CESM. I should not have listed it.

jgfouca:
The test `SMS_D_Ln5.ne4_oQU240.F2010.mappy_gnu` fails on this branch but it also fails on the parent commit. Merging to next.

PR #5946, opened 2023-09-21
Fixes MPAS-Ocean Richardson number calculation

vanroekel:
This changes the calculation of the gradient and bulk richardson numbers to utilize the normal and tangential velocities to avoid the factor of two issue and also fixes a number of places where the loop was over maxLevelCell and should have been to maxLevelEdgeTop.

Testing is documented in -- https://github.com/E3SM-Ocean-Discussion/E3SM/pull/66

[NML]
[CC]

PR #5945, opened 2023-09-21
A quasi-monotone flux limiter for isopycnal diffusion (Redi)

mark-petersen:
These improvements were originally written by @dengwirda, as described in https://github.com/E3SM-Ocean-Discussion/E3SM/pull/53. Please refer to that PR for additional test results and discussion.

This PR addresses a range of issues concerning the isopycnal diffusion operator (Redi):

1. Updating the use of the triad slopes in one of the Redi fluxes.
1. Correcting the implementation of the Redi fluxes.
1. Adding a new new quasi-monotone flux limiting strategy that attempts to prevent the development of non-monotone tracer values.

[non-BFB]
[CC]

PR #5947, opened 2023-09-21
Improve slope calculation and fix slope limiting for redi mixing

vanroekel:
In current MPAS code, slopes are altered directly, e.g. when S > Scrit it is set to zero. Griffies et al 1998 (Appendix C) equates this to slope clipping. The correct approach is to taper the redi kappa coefficient for each individual flux instead. The cross terms ("term2" and "term 3") end up being identical as they are d/dz(kappa S grad phi) so tapering kappa or S are identical. However for the vertical it is d/dz(kappa |S|^2 d phi / dz) so tapering S is different from tapering kappa. Also term 1 (laplacian diffusion) was not being tapered with slopes creating a cross isopycnal flux in steeply sloping regions. These issues are fixed here.

In addition a new slope calculation from @dengwirda is introduced to prevent the slopes from getting very large when grad phi or d/dz(phi) get small. This function limits the slope to be between -1 and 1.

See initial discussion at https://github.com/E3SM-Ocean-Discussion/E3SM/pull/65.

[CC]


vanroekel:
@jonbob in addition to the new defaults in the registry I would suggest we change the default for `config_Redi_min_layers_diag_terms` to zero instead of 6.  I think that capability is no longer needed with the new quasi monotone implementation.  Let me know if you want me to push the new default to the registry.

@mark-petersen and/or @dengwirda what are your thoughts on removing the diagonal term disable config?  I recall that was introduced to prevent extrema so I'm thinking we could remove that capability now, but it would be easier to keep it but set that value to zero

xylar:
@vanroekel, could you give this PR a meaningful title when you have a chance?

dengwirda:
@vanroekel I agree re: updating the defaults to `config_Redi_min_layers_diag_terms = 0`, and added some discussion in #5945. I don't think there's a lot of computational cost associated with having `config_Redi_min_layers_diag_terms` in the code though, so no problem keeping it there for compatibility I'd say.

mark-petersen:
Removed some unused initializations of RediKappa Scaling. With that, the three ocean PRs can merge and pass the nightly test suite, as described in https://github.com/E3SM-Project/E3SM/pull/5945#issuecomment-1734411786.

mark-petersen:
On https://github.com/E3SM-Project/E3SM/pull/5947#issuecomment-1730770173, I would change the default to `config_Redi_min_layers_diag_terms=0` but leave in the flag in case we need it for experimentation.

mark-petersen (approved):
Approved, based on code review, discussion, and simulation results.

jonbob:
merged to next, along with [PR #5945](https://github.com/E3SM-Project/E3SM/pull/5945) and [PR #5946](https://github.com/E3SM-Project/E3SM/pull/5946). Passes sanity testing but DIFFs as expected

jonbob:
merged to master

jonbob:
expected DIFFs and NML DIFFs from the set of PRs are blessed

PR #5948, opened 2023-09-21
CMake: Some minor cleanups

jgfouca:
Remove OS concept from cmake macros. We can maybe remove it entirely from config_machines but that will be trickier.

Remove some unneeded or repeated things from CMakeLists.txt.

Remove FMS dependency support.

[BFB]


components/CMakeLists.txt
```diff
@@ -9,25 +9,17 @@
...
 
-# bmpersch
-# - Read modern cmake docs, use modern features
-# - Use find_package for trilinos and other TPLS
-
 cmake_minimum_required(VERSION 3.9)
 cmake_policy(SET CMP0057 NEW)
 set(CMAKE_CXX_STANDARD 17)
 
-# Store caseroot in the cache, so that, if cmake re-runs,
```
jgfouca, line 20: CASEROOT is already always passed to cmake via `-DCASEROOT=xxx` so is already guaranteed to be in the cache.

components/CMakeLists.txt
```diff
@@ -127,16 +119,14 @@ list(APPEND CMAKE_MODULE_PATH ${CIMEROOT}/CIME/non_py/src/CMake)
 
 set(CMAKE_VERBOSE_MAKEFILE TRUE)
 
-if(USE_CUDA)
```
jgfouca, line 130: This was already being done like 20 lines above.

components/CMakeLists.txt
bartgol: I don't really recall. Maybe I wanted to have a doc string in CMakeCache.txt (which you don't get with -Dblah only)? Long stretch though...

bartgol: But while we're on it, I think it's good practice to add
```
set (XYZ val CACHE TYPE doc)
```
for the CACHE vars that e3sm expects to receive. It is a way for our cmake files to self-document themselves. Otherwise one has to know how CIME will invoke cmake to figure out what comes from where...

components/CMakeLists.txt
```diff
@@ -127,16 +119,14 @@ list(APPEND CMAKE_MODULE_PATH ${CIMEROOT}/CIME/non_py/src/CMake)
...
-  enable_language(CUDA)
-endif()
-
 # Scream manages its own flags
 build_eamxx()
 
 # We do want CMAKE_BUILD_TYPE to be set, but we do NOT want CMake to
 # decide what optimization flags to append, based on build type,
 # for components who rely on CIME for build flags, so make all the following empty.
+# JGF: I think we can remove this once proper CMake names are being used in the
```
bartgol, line 128: Agreed. I think the mach files should set these variables to some reasonable values.

bartgol (approved):

bartgol (approved):
I'm always excited when the lines removed are more than those added in a PR!

components/CMakeLists.txt
jgfouca: @bartgol , thanks. I will add that to the list of objectives for the full refactor.

jgfouca:
@mahf708 , the sharedlibs builds still potentially use `OS` but that doesn't mean we have to use it in our cmake macros.

PR #5949, opened 2023-09-22
Demonstrate the use of GitHub actions in E3SM testing

mahf708:
Testing running basic tests on ghas

Please don't merge. This is just an illustration. Stay tuned. In the meanwhile, please let me know if there are issues you'd like addressed or if you have ideas for improvements.

Things I'd like to do before merging:
- [ ] standardize the docker container build (abstract options and enable matrix build)
  - add missing items
  - build for different compiler/os/libs matrix
- [x] generalize the workflow to add run tests besides the build test
- [ ] print full logs if job fails
- [ ] chain some jobs to test restarting on two different instances
- [ ] explore caching


bartgol:
I was able to build scream in a github action using a self-hosted runner on my workstation. I can get one set up from the `e3sm-autotester` user account (so that it can do ssh ops with its ) on mappy. Then, we would be able to rely on all the modules that we already use for other testing (nightly, scream AT, ...). For security reasons (since it would run on SNL machines), we may have to check that a certain label is set, so that only ppl with write privileges can trigger the action.

.github/workflows/old.WCYCL1850.ne4_oQU240.yaml
```diff
@@ -0,0 +1,36 @@
...
+    branches: [ master ]
+ 
+  workflow_dispatch:
+
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
+    container: 
+      image: ghcr.io/mahf708/e3sm-imgs@sha256:2657196ea9eec7dbd04f7da61f4e7d5e566d4b501dff5881f0cb5125ba031158
```
bartgol, line 16: I'm assuming this container has all TPLs already installed? If so, what's the point of the apg-get calls below?

cime_config/machines/config_machines.xml
```diff
@@ -1249,6 +1249,51 @@
...
+    <NODENAME_REGEX>ghas</NODENAME_REGEX>
+    <OS>LINUX</OS>
+    <COMPILERS>gnu</COMPILERS>
+    <MPILIBS>mpich</MPILIBS>
+    <CIME_OUTPUT_ROOT>$ENV{HOME}/projects/e3sm/scratch</CIME_OUTPUT_ROOT>
+    <DIN_LOC_ROOT>$ENV{HOME}/projects/e3sm/cesm-inputdata</DIN_LOC_ROOT>
+    <DIN_LOC_ROOT_CLMFORC>$ENV{HOME}/projects/e3sm/ptclm-data</DIN_LOC_ROOT_CLMFORC>
+    <DOUT_S_ROOT>$ENV{HOME}/projects/e3sm/scratch/archive/$CASE</DOUT_S_ROOT>
+    <BASELINE_ROOT>$ENV{HOME}/projects/e3sm/baselines/$COMPILER</BASELINE_ROOT>
+    <CCSM_CPRNC>$CCSMROOT/tools/cprnc/build/cprnc</CCSM_CPRNC>
```
bartgol, line 1263: @jgfouca A quick grep through the repo and CIME makes me think that CCSMROOT is no longer used, but still present on some (unused?) machines in config_machines.xml. Is that correct? If so, this line should be changed I think.

bartgol: Good strategy leaving the email of someone else. I shall use that trick in the future... :)

bartgol:
Food for thoughts: by creating a self-hosted runner on e3sm machines, we could use gh actions to automatically test PRs without the need for a next branch, and without the need of waiting for overnight testing.

Workflow idea is as follows.
 - Keep the gh action that runs small tests on a github runner. This verifies that no "obvious" errors are there (e.g., the code builds and runs without crashing).
 - Add gh action that checks PR was approved by an integrator (can use something like [this one](https://github.com/marketplace/actions/required-review)). This is necessary b/c we want to avoid running malicious code on our DOE machines. Having a trusted user approve the PR first gives us _some_ confidence that the code is ok.
 - If the above two pass, kick off gh action on self-hosted runners (can use a matrix).
 - Add an action that runs every time master is updated. For each open PR, disable merging if the last round of testing was done wrt a different base ref. That's to avoid merging a PR that passed before, but doesn't pass after last master update.
 - Make the 1st and 3rd step trigger if the base branch changes. I don't think we can handle this from those gh actions (I can't find a trigger like "if base ref changes"). So the closest approach is to make the workflow at the step above add some special label (eg. "RETEST") to all open PRs. Then, make the pr test workflows trigger if labels change (and "RETEST" is among them). Drawback is that people can manually remove the label, but I assume we "trust" users with write access to the repo to not do something like that. The benefit of a trigger if "RETEST" is set is that we can use it also to manually re-run a testing framework.
 - For BFB PRs only, if
 	- every action passes
 	- the base branch has not changed since the PR was created
 	- automerge label is set
   we can go ahead and merge.
 - For non-BFB PRs, an integrator must use the force, and manually merge.
 
 Notice that, by checking that the base ref has not changed, we no longer need `next`: all PRs are re-tested when another PR is merged. While this may somewhat reduce the amount of PRs that can go in _at the same time_, we will gain by having testing throughout the day. We could also still use `next` as the base ref, of course.
 
 The drawback of continuos testing (as opposed to nightly only) is the amount of work we throw at the e3sm machines. Each self-hosted runner can run 1 job at a time, so by having 1 runner/machine we limit how much of the machine we can use. However, on some machines (e.g., mappy), e3sm testing takes a long time (3h), and we don't want to hold the machine for that long during the day. That said, since mappy is not a production machine, we may not use it for CI, and just keep it for nightlies, for extra safety.

rljacob:
The need to test overnight comes from both how long it takes to run the integration suite AND the need to test on multiple busy platforms-compilers.  We then put all the PRs from the day on next so we only have to run it once and not once-per-PR.  The only way we can get to a test-per-PR workflow is to both have a suite that runs in a short amount of time AND a way to run it on all our machine-compiler combinations.

rljacob:
That being said, one reason PRs spend a lot of time open is Integrators just forget to move them through the workflow.  So even a way to get them on to next for overnight testing automatically and then to master will increase the velocity.

bartgol:
To be fair, those were ne4 tests. Nightly tests use ne30, which takes considerably longer than ne4...

We can discuss running small ne4 integration testsuites (possibly on a variety of machines) before merging to next, to rule out initial build/run errors and unexpected diffs. Then, if these pass, automate the merge to next (somehow querying if it's open). If this is already a good improvement over current framework, it's worth doing it.

PR #5950, opened 2023-09-22
CMake: Improve handling of Albany and related packages

### LINKED ISSUE #5786: Import cmake file from Albany
As discussed offline at the E3SM all hands meeting, it would be nice if E3SM build system could import the Albany cmake configuration file. The file is store in the `ALBANY_PATH` folder. 
As an example here is the Albany path for perlmutter:
`/global/common/software/fanssie/albany-e3sm-serial-release-gcc/`

Albany libraries should be added to the `LD_LIBRARY_PATH` during the case.setup.

In addition, we'd like the build to provide the `ALBANY_PATH` as a environmet variable, rather than by setting it in the 
`cime_config/machines/cmake_macros/*.cmake` file.

I'll be happy to test this capability when ready. Thanks @jgfouca for agreeing to work on it!

Pinging @bartgol @jewatkins so that they stay in the loop.

jgfouca:
Specifically:
* Albany
* Trilinos
* Kokkos

This PR establishes the new paradigm of setting ${Package}_ROOT to find the packages we need. So far, I am very happy with how this is going. You can see how much cleaner it is to do things the CMake-3 so you don't have to micro-manage flags.

This PR is currently WIP because I was not able to fully test it on pm-cpu due to permissions issues with the boost that was used to build Trilinos.

In order to be consistent, changes kokkos override env var from KOKKOS_PATH to Kokkos_ROOT.

Fixes #5786

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -257,6 +257,8 @@
       <env name="PERL5LIB">/global/cfs/cdirs/e3sm/perl/lib/perl5-only-switch</env>
       <env name="FI_CXI_RX_MATCH_MODE">software</env>
       <env name="MPICH_COLL_SYNC">MPI_Bcast</env>
+      <env name="Albany_ROOT">$SHELL{if [ -z "$Albany_ROOT" ]; then echo /global/common/software/e3sm/mali_tpls/albany-e3sm-serial-release-gcc; else echo "$Albany_ROOT"; fi}</env>
+      <env name="Trilinos_ROOT">$SHELL{if [ -z "$Trilinos_ROOT" ]; then echo /global/common/software/e3sm/mali_tpls/trilinos-e3sm-serial-release-gcc; else echo "$Trilinos_ROOT"; fi}</env>
     </environment_variables>
     <environment_variables compiler="gnu" mpilib="mpich">
       <env name="ADIOS2_DIR">/global/cfs/cdirs/e3sm/3rdparty/adios2/2.8.3.patch/cray-mpich-8.1.15/gcc-11.2.0</env>
--- a/components/cmake/common_setup.cmake
+++ b/components/cmake/common_setup.cmake
@@ -205,65 +205,6 @@ if (USE_PETSC)
   set(PETSC_LIB ${PETSC_LIBRARIES})
 endif()
 
-if (USE_TRILINOS)
-  if (TRILINOS_PATH)
-    if (NOT INC_TRILINOS)
-      set(INC_TRILINOS ${TRILINOS_PATH}/include)
-    endif()
-    if (NOT LIB_TRILINOS)
-      set(LIB_TRILINOS ${TRILINOS_PATH}/lib)
-    endif()
-  else()
-    message(FATAL_ERROR "TRILINOS_PATH must be defined when USE_TRILINOS is TRUE")
-  endif()
-
-  set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${TRILINOS_PATH} PARENT_SCOPE)
-  find_package(Trilinos)
-endif()
-
-if (USE_ALBANY)
-  if (ALBANY_PATH)
-    if (NOT INC_ALBANY)
-      set(INC_ALBANY ${ALBANY_PATH}/include)
-    endif()
-    if (NOT LIB_ALBANY)
-      set(LIB_ALBANY ${ALBANY_PATH}/lib)
-    endif()
-  else()
-    message(FATAL_ERROR "ALBANY_PATH must be defined when USE_ALBANY is TRUE")
-  endif()
-
-  # get the "ALBANY_LINK_LIBS" list as an env var
-  file(READ ${ALBANY_PATH}/export_albany.in ALBANY_OUTPUT)
-  string(REPLACE "ALBANY_LINK_LIBS=" "" ALBANY_LINK_LIBS "${ALBANY_OUTPUT}")
-endif()
-
-if (USE_KOKKOS)
-  # LB 01/23
-  # CMake's find_package, when used with multiple PATHS and PATH_SUFFIXES,
-  # follows the following rule when looking in paths:
-  #  1. look in all the path suffixes of the first PATHS entry, in the order provided
-  #  2. look in the first PATH provided
-  #  3. repeat 1-2 with the following entry of PATHS
-  #  4. look in cmake/system default paths (unless told not to).
-  # So the following cmd will fist look in the KOKKOS_PATH folder and subfolders,
-  # if KOKKOS_PATH is non-empty. Then will proceed to look in the lib, lib/cmake,
-  # and lib64/cmake subfolders of the INSTALL_SHAREDPATH. If all of these fail,
-  # it will look in INSTALL_SHAREDPATH.
-
-  if (KOKKOS_PATH)
-    set (PATHS ${KOKKOS_PATH} ${INSTALL_SHAREDPATH})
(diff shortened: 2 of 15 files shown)


cime_config/machines/config_machines.xml
```diff
@@ -257,6 +257,8 @@
       <env name="PERL5LIB">/global/cfs/cdirs/e3sm/perl/lib/perl5-only-switch</env>
       <env name="FI_CXI_RX_MATCH_MODE">software</env>
       <env name="MPICH_COLL_SYNC">MPI_Bcast</env>
+      <env name="Albany_ROOT">$SHELL{if [ -z "$Albany_ROOT" ]; then echo /global/common/software/e3sm/mali_tpls/albany-e3sm-serial-release-gcc; else echo "$Albany_ROOT"; fi}</env>
```
jgfouca, line 260: The pattern for setting up ${Package}_ROOT is established here. It's a bit messy so I'm wondering if it would be better to move this login into the cmake macro file for the machine?

bartgol: There's a better bash syntax:

```
Albany_ROOT=${Albany_ROOT:=/global/common/software/e3sm/mali_tpls/albany-e3sm-serial-release-gcc}
```
The `${VAR:=default}` syntax expands to `${VAR}` if `VAR` is defined, otherwise it expands to `default`. E.g.:
```
$ export FOO=foo
$ export BAR=${FOO:=bar} && echo $BAR
foo
$ export BAR={FOOBAR:=bar} && echo $BAR
bar
```

components/cmake/common_setup.cmake
```diff
@@ -205,65 +205,6 @@ if (USE_PETSC)
   set(PETSC_LIB ${PETSC_LIBRARIES})
 endif()
 
-if (USE_TRILINOS)
```
bartgol, line 208: So much better than what we had...

components/cmake/find_dep_packages.cmake
```diff
@@ -0,0 +1,36 @@
...
+  # unset.
+  if (NOT DEFINED ENV{Kokkos_ROOT})
+    set(ENV{Kokkos_ROOT} ${INSTALL_SHAREDPATH})
+  endif()
+
+  find_package(Kokkos REQUIRED)
+endif()
+
+# Albany depends on Trilinos
+if (USE_ALBANY OR USE_TRILINOS)
```
bartgol, line 29: This is fine, but for the record, I think that `USE_ALBANY` here is redundant. `find_package(ALBANY REQUIRED)` will also call `find_package(Trilinos REQUIRED)`. If a CMake package is installed well, it is responsible of finding all its TPLs when you call `find_package(YOUR_PKG)`.

cime_config/machines/config_machines.xml
jgfouca: @bartgol , I tried that at first. The problem is that the CIME regex will incorrectly match the first `}` as terminating the SHELL command, so you can use that character in your syntax.

components/cmake/find_dep_packages.cmake
jgfouca: It looks like this is a deficiency with the Albany cmake system. I tried without `USE_ALBANY` check for Trilinos and I got this error:

```
CMake Error at /global/common/software/e3sm/mali_tpls/albany-e3sm-serial-release-gcc/lib64/Albany/cmake/albany-targets.cmake:61 (set_target_properties):
  The link interface of target "albanyLib" contains:

    Panzer::all_libs

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  /global/common/software/e3sm/mali_tpls/albany-e3sm-serial-release-gcc/lib64/Albany/cmake/AlbanyConfig.cmake:28 (include)
  cmake/find_dep_packages.cmake:45 (find_package)
  CMakeLists.txt:124 (include)
```

jgfouca:
@mperego , there are some deficiencies in the Albany Cmake system that are causing me problems:

1. Albany's find_package does not find Trilinos, so E3SM has to do it. It's easy enough for E3SM to do it, so this is not a show stopper.
2. For some reason, the `mpasInterface` library is linked via `-lmpasInterface` instead of just listing the full path to the library like the other Albany libraries do. This causes a failure `cannot find -lmpasInterface: No such file or directory` since the albany cmake doesn't add that directory via `-L`

If I fix (2) by manually editing the link line (replace -lmpasInterface with `/global/common/software/e3sm/mali_tpls/albany-e3sm-serial-release-gcc/lib64/libmpasInterface.so`, then everything works and it builds.

Since these problems aren't related to E3SM, I'm going to push ahead with this PR unless you say otherwise.

jgfouca:
I removed all instances of ALBANY_PATH from the Cmake macros. This means that the `BGWCYCL1850` compset will only work on perlmutter currently. Any machine that wants to support this compset will need to follow the pattern in the perlmutter config_machines and add  Albany_ROOT and Trilinos_ROOT.

components/cmake/find_dep_packages.cmake
bartgol: Ah, we are probably not packaging albany that well.. :/

bartgol: I am going through our cmake stuff, and indeed we lack a lot of stuff to consider us "good cmake citizens"... I will fix it at some point.

bartgol:
> @mperego , there are some deficiencies in the Albany Cmake system that are causing me problems:
> 
> 1. Albany's find_package does not find Trilinos, so E3SM has to do it. It's easy enough for E3SM to do it, so this is not a show stopper.
> 2. For some reason, the `mpasInterface` library is linked via `-lmpasInterface` instead of just listing the full path to the library like the other Albany libraries do. This causes a failure `cannot find -lmpasInterface: No such file or directory` since the albany cmake doesn't add that directory via `-L`
> 
> If I fix (2) by manually editing the link line (replace -lmpasInterface with `/global/common/software/e3sm/mali_tpls/albany-e3sm-serial-release-gcc/lib64/libmpasInterface.so`, then everything works and it builds.
> 
> Since these problems aren't related to E3SM, I'm going to push ahead with this PR unless you say otherwise.

We're fixing other issues with our cmake logic (lots of legacy stuff, and some not-so-great cmake usage here and there). I'll keep in mind these issues in the process.

jgfouca:
Thanks, @bartgol ! Merged to next.

mperego:
Thanks a lot @jgfouca! This looks so much cleaner. Looking forward to trying this out.

bartgol:
> @mperego , there are some deficiencies in the Albany Cmake system that are causing me problems:
> 
> 1. Albany's find_package does not find Trilinos, so E3SM has to do it. It's easy enough for E3SM to do it, so this is not a show stopper.
> 2. For some reason, the `mpasInterface` library is linked via `-lmpasInterface` instead of just listing the full path to the library like the other Albany libraries do. This causes a failure `cannot find -lmpasInterface: No such file or directory` since the albany cmake doesn't add that directory via `-L`
> 
> If I fix (2) by manually editing the link line (replace -lmpasInterface with `/global/common/software/e3sm/mali_tpls/albany-e3sm-serial-release-gcc/lib64/libmpasInterface.so`, then everything works and it builds.
> 
> Since these problems aren't related to E3SM, I'm going to push ahead with this PR unless you say otherwise.

I found the culprit. Albany was not exporting the `mpasInterface` target (unlike the other libs), so it was not available as a CMake target to downstream projects. Will fix it in Albany.

PR #5951, opened 2023-09-23
add docs preview within PRs

mahf708:
Adds docs preview to PR page. Preview appears in a comment.

This workflow will update the same comment if there are changes
(it won't pollute the PR page with more comments). Additionally,
I spaced out the GitHub action so that it is easier to read and maintain (and shortened the names)

[BFB]

-----
See below for `github-actions` bot bringing a link.  This hopefully will encourage people to iterate on docs here, rather than doing everything locally.

### DIFF
--- a/.github/workflows/e3sm-gh-pages.yml
+++ b/.github/workflows/e3sm-gh-pages.yml
@@ -33,7 +33,18 @@ jobs:
       # build every time (PR or push to master)
       - name: Build
         run: mkdocs build --strict --verbose
-      # deploy only when it is a push
+      # Only deploy to the main github page when there is a push to master
       - if: ${{ github.event_name == 'push' }}
-        name: Deploy
-        run: mkdocs gh-deploy
+        name: GitHub Pages action
+        uses: JamesIves/github-pages-deploy-action@v4
+        with:
+          # Do not remove existing pr-preview pages
+          clean-exclude: pr-preview
+          folder: ./site/
+      # If it's a PR from within the same repo, deploy to a preview page
+      # For security reasons, PRs from forks cannot write into gh-pages for now
+      - if: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository }}
+        name: Preview docs
+        uses: rossjrw/pr-preview-action@v1
+        with:
+          source-dir: ./site/


rljacob:
This is good to catch problems before a merge but I  also want people to build docs locally before they make a commit.  Otherwise the branch will have a bunch  of "fix docs" commits.  Yes they could remove those with a rebase but that skill might be harder to teach then installing mkdocs locally.

.github/workflows/e3sm-gh-pages.yml
```diff
@@ -12,28 +12,46 @@ on:
...
+        if: ${{ github.event_name == 'push' }}
+        name: GitHub Pages action
+        uses: JamesIves/github-pages-deploy-action@v4
+        with:
+          # Do not remove existing pr-preview pages
+          clean-exclude: pr-preview
+          folder: ./site/
+      - 
+        # If it's a PR from within the same repo, deploy to a preview page
+        # For security reasons, PRs from forks cannot write into gh-pages for now
```
rljacob, line 52: Can we still do a build check of docs in PRs from forks?

.github/workflows/e3sm-gh-pages.yml
```diff
@@ -12,28 +12,46 @@ on:
...
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - 
+        name: Check out code
+        uses: actions/checkout@v4
         with:
           show-progress: false
-          fetch-depth: 0 # Needed, or else gh-pages won't be fetched, and push rejected
+          # We will need submodules when/if we build EAMxx docs here
```
rljacob, line 25: Oh?  Why is that?  EAMxx is not a submodule.

bartgol: EAMxx docs run some scripts that generates documentation for all input parameters, based on `namelist_defaults_scream.xml`. In doing so, it indirectly needs some python modules from CIME. Therefore, we do need at least the CIME submodule in order to build the eamxx docs.

If anyone thinks we should not require submodules just to build docs, we can find a workaround.

.github/workflows/e3sm-gh-pages.yml
bartgol: Currently, no. At least, not through this action. The documentation on the pr-preview-action github page claims it will support PRs from fork in v2. Although, that message on their docs is 1.5y old, so I'm not 100% confident they'll follow through on that...

That said, it's not complicate to write our own actions. We could easily fork that action, upgrade it to handle forks, and store it on e3sm-project. In the long run, it may even be better, in case actions are not maintained...

.github/workflows/e3sm-gh-pages.yml
```diff
@@ -12,28 +12,46 @@ on:
...
           submodules: false # speeds up clone and not building anything in submodules
-      - name: Show action trigger
+      - 
+        name: Show action trigger
         run: echo "= The job was automatically triggered by a ${{github.event_name}} event."
```
bartgol, line 29: Nitpick: I would prune "automatically", since the workflow can be manually triggered.

bartgol (approved):

.github/workflows/e3sm-gh-pages.yml
rljacob: The current action will (I think) check the doc build on a PR from anywhere.  I'd like to keep that.

bartgol: It will definitely build the docs, so you'll know if there's some error in the docs generation, but I think the preview will fail. In order to preview it, it has to push on the gh-pages branch of our repo, which cannot be done if the PR comes from a fork.

bartgol: Ah, wait, you mean the action currently in master? I think the current master version is not doing any preview for PRs. The deploy step is guarded by `if: ${{ github.event_name == 'push' }}`.

rljacob: Its not doing a preview but its building for every PR even from forks.  I want to keep build-from-fork.  Sounds like it will.

.github/workflows/e3sm-gh-pages.yml
rljacob: Checking  out submodules noticeably slows down the model clone time.  That would be the only reason to do it.

bartgol: Yes, that's true.

@jgfouca I think we can do away with CIME in our buildnml & friends scripts if we roll our own version of `expect` (or copy/paste CIME's). I think that's the _only_ reason why we use a CIME module in the `.py` helper files.

rljacob:
just leave it as one PR.

Tell me more about how fixing docs revealed to be broken by the PR preview results in only one commit.

bartgol:
I think Rob is right when he says that there will be several "fix docs" commits. That said, I also agree with Naser when he says "why making a fuss with such commits here, when nobody cares about such commits for source code?". The repo is full of "should work now" or "progress" commits. I am not a fan of them, and I agree that we should keep a meaningful history, but unless we want to make an effort to stop micro commits with little info in the message, I don't see a reason to use that argument here.

As a mitigation step, I propose we ask the PR author to condense all docs-related changes in a single commit, to be put at the very top of the PR. By putting it at the top, users can freely force push to the PR as they update the docs, since it will only rewrite one commit, and only in the docs folder. This allows to not discard the reviewers "progress" in their review (every time there are force pushed commits with rewrite changes to the code, seeing new changes to the PR becomes hard, since the git sha's I reviewed are gone).

rljacob:
I _would_ like to discourage micro-commits in all code including asking developers to rewrite history to remove "should work now" messages.

Keeping the doc fixes to the last commit means it should also be done last since  some commits will be made on the code in response to reviews.

rljacob:
Won't you have to add commits to the feature branch that has the (broken) docs?  That's the branch I'm concerned about.  But we could have them re-written out.

bartgol:
> I _would_ like to discourage micro-commits in all code including asking developers to rewrite history to remove "should work now" messages.
> 
> Keeping the doc fixes to the last commit means it should also be done last since some commits will be made on the branch in response to reviews.

As long as developers put docs changes in separate commits, we can always rewrite history right before we merge. Of course, that requires some attention on the integrator side.

Perhaps a good habit is to use `DOCS: blah` as commit message whenever we commit docs changes...

rljacob:
This discussion of how to write docs is independent of this PR.  I'm going to go ahead and start merging it.

rljacob (approved):

bishtgautam:
Btw, why is the [https://e3sm-project.github.io/E3SM/pr-preview/pr-5951/](https://e3sm-project.github.io/E3SM/pr-preview/pr-5951/) rendered incorrectly?

bartgol:
> Btw, why is the [e3sm-project.github.io/E3SM/pr-preview/pr-5951](https://e3sm-project.github.io/E3SM/pr-preview/pr-5951/) rendered incorrectly?

I suspect the reason is current master. In master, we don't yet have the gh action that does preview of prs. The code in master assumes it has total control over the gh pages site tree. That causes existing subpages to be completely deleted (e.g., the `pr-preview/pr-XYZ` subpages. This PR will allow those subpages to stay alive when master updates the main releases.

bishtgautam:
When I hover over the `EAM->Introduction`, the link actually points to the already published URL and not the preview GH pages.

pr-preview

bishtgautam:
@mahf708, Now things are working correctly. Thanks.

PR #5959, opened 2023-09-26
Cleanup petsc cmake stuff

jgfouca:
Also remove a lot of unused stuff.

[BFB]


components/cmake/build_model.cmake
```diff
@@ -285,7 +285,13 @@ function(build_model COMP_CLASS COMP_NAME)
         target_link_libraries(${TARGET_NAME} PRIVATE samxx)
       endif()
       if (USE_RRTMGPXX)
-          target_link_libraries(${TARGET_NAME} PRIVATE rrtmgp rrtmgp_interface)
+        target_link_libraries(${TARGET_NAME} PRIVATE rrtmgp rrtmgp_interface)
+      endif()
+    endif()
+    if (COMP_NAME STREQUAL "elm")
+      if (USE_PETSC)
+        target_link_libraries(${TARGET_NAME} PRIVATE "${PETSC_LIBRARIES}")
```
bartgol, line 293: Does `find_package(PETSc REQUIRED)` not import PETSc targets? Does it simply provide `PETSC_LIBRARIES` and `PETSC_INCLUDES`? :/

bartgol (approved):
Looks good. I had a comment on how we link petsc, but either way, it's fine.

components/cmake/build_model.cmake
jgfouca: @bartgol , unfortunately I think the answer is yes. The Petsc-find.cmake was written many years ago.

jonbob: @jgfouca -- I've also been testing a more recent version of find_petsc and I'm pretty sure we could move to that as well. If that's any improvement....

jgfouca: @jonbob , that would be great. We have a few of them lying around our repos:
```
./components/elm/src/external_models/mpp/cmake/Modules/FindPETSc.cmake
./components/elm/src/external_models/emi/src/cmake/Modules/FindPETSc.cmake
./cime/CIME/non_py/src/CMake/FindPETSc.cmake
```

Would be nice to agree on one location and update it.

jonbob: I'll make sure the one I've been testing is the updated one and let you know. And I always end up using the cime/CIME location, since the mpas build finds it

PR #5964, opened 2023-09-27
CMake: Improve dependency management for mct and csm_share

jgfouca:
Neither of these packages are cmake packages, so we have to roll our own Find$package.cmake modules. These can be pretty simple for e3sm since we build them ourselves (no worries about version or not found etc).

Also, remove a lot of unused for special handling of clm and stuff for pio building.

[BFB]


bartgol (commented):
Mostly good. I have some suggestions, you can decide what's worth adding.

Btw, are you thinking of changing `share/util/CMakeLists.txt` to actually be a cmake project, rather than whatever it is now?

components/cmake/modules/FindMCT.cmake
```diff
@@ -0,0 +1,22 @@
...
+
+# Look for libmct in INSTALL_SHAREDPATH/lib
+find_library(MCT_LIB  mct  REQUIRED PATHS ${INSTALL_SHAREDPATH}/lib)
+find_library(MPEU_LIB mpeu REQUIRED PATHS ${INSTALL_SHAREDPATH}/lib $ENV{mct_ROOT})
+
+# Create the interface library, and set target properties
+add_library(mct INTERFACE)
+target_link_libraries(mct INTERFACE ${MCT_LIB} ${MPEU_LIB})
+target_include_directories(mct INTERFACE ${INSTALL_SHAREDPATH}/include)
+target_compile_definitions(mct INTERFACE "MCT_INTERFACE")
```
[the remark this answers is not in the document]
rljacob: Also not sure where this is use but there is code in CIME to allow MCT or MOAB to be used as the coupler.  We need to keep that.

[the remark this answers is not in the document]
bartgol: You mean we need to keep ability to use MOAB, not "keep MCT_INTERFACE in CPPDEFS", right?

Also, is there an example in e3sm of a compset using moab instead of mct?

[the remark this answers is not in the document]
rljacob: Yes that's what I mean.

Its not in master yet but you can run a moab case on this branch (on Chrysalis only for now) https://github.com/sarich/E3SM/tree/sarich/moab/update-moab-driver
e.g.  ./create_test SMS_Vmoab.ne4pg2_oQU480.WCYCL1850NS

components/cmake/modules/FindCsmShare.cmake
```diff
@@ -0,0 +1,32 @@
...
+  set(ESMFDIR "noesmf")
+endif()
+set(CSM_SHARE "${INSTALL_SHAREDPATH}/${COMP_INTERFACE}/${ESMFDIR}/${NINST_VALUE}/csm_share")
+
+# Look for libcsm_share in the complex path we built above
+find_library(CSM_SHARE_LIB csm_share REQUIRED PATHS ${CSM_SHARE})
+
+# Create the interface library, and set target properties
+add_library (csm_share INTERFACE)
+target_link_libraries (csm_share INTERFACE ${CSM_SHARE_LIB};mct;${PIOLIBS})
```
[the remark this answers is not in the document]
jgfouca: Good point, done!

components/cmake/modules/FindMCT.cmake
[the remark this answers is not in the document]
jgfouca: `MCT_INTERFACE` does not appear to be used in CIME either, but there's not much harm in keeping it.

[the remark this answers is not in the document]
bartgol: Imho, if something is unused (and pointless), it should be removed. Pointless code tends to pile up, until we have no idea of what is actually needed/used. I vote for pruning it.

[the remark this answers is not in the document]
jgfouca: I think you are right.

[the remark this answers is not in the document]
jgfouca: @rljacob , I looked at Sarich's Moab branch and he's not using this cppdef there either.

bartgol (approved):

jgfouca:
Hmm, it looks like the CI build fails with a link error for the case it tries to build. The same case builds fine on mappy so I may have some digging to do.

components/cmake/modules/FindCsmShare.cmake
```diff
@@ -29,7 +29,7 @@ endif()
 
 # Create the interface library, and set target properties
 add_library (csm_share INTERFACE)
-target_link_libraries (csm_share INTERFACE ${CSM_SHARE_LIB};mct;${PIOLIBS})
+target_link_libraries (csm_share INTERFACE ${CSM_SHARE_LIB};mct;spio)
```
bartgol, line 32: Should we maybe to the same thing we did for MCT also for spio? That is, adding
```
find_package(PIO)
```
?

components/cmake/common_setup.cmake
```diff
@@ -113,69 +107,12 @@ else()
...
-      set(INC_PNETCDF ${PNETCDF_PATH}/include)
-    endif()
-    if (NOT LIB_PNETCDF)
-      set(LIB_PNETCDF ${PNETCDF_PATH}/lib)
-    endif()
-  endif()
+# Have to do this here for now because the macros must be loaded in order for
+# the netcdf cmake variables to be set. Once scorpio has a good package-config,
+# this won't be necessary
+if (NOT TARGET netcdf)
```
bartgol, line 113: I don't think we need this if guard. subsequent calls to `find_package` will do nothing if the package was already found...

components/cmake/modules/FindNETCDF.cmake
```diff
@@ -0,0 +1,66 @@
...
+
+  find_library(netcdf_c_lib netcdf  REQUIRED PATHS ${NETCDF_PATH}/lib ${NETCDF_PATH}/lib64)
+  find_library(netcdf_f_lib netcdff REQUIRED PATHS ${NETCDF_PATH}/lib ${NETCDF_PATH}/lib64)
+  find_path (netcdf_c_incdir netcdf.h REQUIRED PATHS ${NETCDF_PATH}/include)
+  find_path (netcdf_f_incdir netcdf.inc REQUIRED PATHS ${NETCDF_PATH}/include)
+else()
+  message(FATAL_ERROR "NETCDF not found: Define NETCDF_PATH or NETCDF_C_PATH and NETCDF_FORTRAN_PATH in config_machines.xml or config_compilers.xml")
+endif()
+
+set (pnetcdf_lib ${pnetcdf_lib} PARENT_SCOPE)
```
bartgol, line 53: Why do you need parent scope? They should be used exclusively inside this script, to create the netcdf target, no?

components/cmake/modules/FindNETCDF.cmake
```diff
@@ -0,0 +1,66 @@
...
+
+  # Find the libraries
+  find_library(netcdf_c_lib netcdf  REQUIRED PATHS ${NETCDF_C_PATH}/lib ${NETCDF_C_PATH}/lib64)
+  find_library(netcdf_f_lib netcdff REQUIRED PATHS ${NETCDF_FORTRAN_PATH}/lib ${NETCDF_FORTRAN_PATH}/lib64)
+  find_path (netcdf_c_incdir netcdf.h REQUIRED PATHS ${NETCDF_C_PATH}/include)
+  find_path (netcdf_f_incdir netcdf.inc REQUIRED PATHS ${NETCDF_FORTRAN_PATH}/include)
+
+elseif (NETCDF_FORTRAN_PATH)
+  message(FATAL_ERROR "NETCDF_FORTRAN_PATH specified without NETCDF_C_PATH")
+elseif (NETCDF_PATH)
```
bartgol, line 38: I wonder if we should continue to support a single NETCDF_PATH. It makes this code much heavier. Requiring the <LANG> vars would make the code much shorter..

components/cmake/modules/FindPIO.cmake
```diff
@@ -0,0 +1,24 @@
...
+if (NOT PIO_LIBDIR)
+  set(PIO_LIBDIR "${INSTALL_SHAREDPATH}/lib")
+endif()
+
+if (PIO_VERSION STREQUAL 2)
+  # This is a pio2 library
+  set(PIOLIBS "${PIO_LIBDIR}/libpiof.a;${PIO_LIBDIR}/libpioc.a")
+else()
+  # This is a pio1 library
+  set(PIOLIBS "${PIO_LIBDIR}/libpio.a")
```
bartgol, line 19: Does E3SM hard-code static libraries?

components/cmake/modules/FindNETCDF.cmake
```diff
@@ -0,0 +1,66 @@
...
+set (pnetcdf_incdir ${pnetcdf_incdir} PARENT_SCOPE)
+set (netcdf_c_incdir ${netcdf_c_incdir} PARENT_SCOPE)
+set (netcdf_f_incdir ${netcdf_f_incdir} PARENT_SCOPE)
+
+# Create the interface library, and set target properties
+add_library(netcdf INTERFACE)
+target_link_libraries(netcdf INTERFACE ${pnetcdf_lib};${netcdf_c_lib};${netcdf_f_lib})
+target_include_directories(netcdf INTERFACE ${pnetcdf_incdir};${netcdf_c_incdir};${netcdf_f_incdir})
+if (pnetcdf_lib)
+  target_compile_definitions(netcdf INTERFACE "-D_PNETCDF")
```
bartgol, line 65: Is this CPP def needed? I see `_PNETCDF` only inside scorpio, but it is locally defined inside `pio.h` based on other macros. Can we purge this?

components/cmake/modules/FindPIO.cmake
```diff
@@ -0,0 +1,24 @@
+# - Try to find PIO
+#
+# Placeholder code until spio gets a proper config.cmake
+#
+# Once done, this will define:
+#
+#   The "spio" target
+#
+
+if (NOT PIO_LIBDIR)
```
bartgol, line 10: Does this mean that the user can pass `-DPIO_LIBDIR=/path/to/my/scorpio` to use an existing installation? If so, maybe we should use a more cmake-y name, like `PIO_DIR` or `PIO_ROOT`?

components/cmake/modules/FindCsmShare.cmake
jgfouca: Yep, did exactly that.

components/cmake/common_setup.cmake
jgfouca: I think that depends on the implementation, but I'm not sure. I wanted to keep the findX modules as simple as possible, so they don't set or check XX_FOUND

components/cmake/modules/FindNETCDF.cmake
jgfouca: Good catch. Copy/paste error from the useful cmake work you in scream.

components/cmake/modules/FindNETCDF.cmake
jgfouca: This will go away once scorpio has a good cmake config.

components/cmake/modules/FindPIO.cmake
jgfouca: This will go away once scorpio has a good cmake config.

components/cmake/modules/FindNETCDF.cmake
jgfouca: It's a little tricky to grep for, tons of false positives. I think you are right that it's used by PIO which gets build in the shared lib phase and so doesn't use this code.

components/cmake/modules/FindPIO.cmake
jgfouca: This will go away once scorpio has a good cmake config.

components/cmake/common_setup.cmake
bartgol: I think CMake states it clearly
```
Once one of the calls succeeds the result variable will be set and stored in the cache so that no call will search again.
```

jgfouca: That could be for find_package when it does things the "normal" way, IE loading the package-config.cmake. When you are rolling your own find module, I don't think this is true. The example FindXXX.cmake guidance in the documentation has all that stuff explicit.

components/cmake/modules/FindNETCDF.cmake
bartgol: Also, scorpio sets it internally in pio.h:
```c++
#include "pio_config.h"
#if PIO_USE_PNETCDF
  #define _PNETCDF 1
#endif
```
I think this is one of those macros that were used at some point, but when they became unused nobody took care of removing them (perhaps based on the same fear we all have: "what if it's actually needed somewhere?").

Without grepping in submodules, I don't get any meaningful match of the _whole word_ `_PNETCDF`. Only scorpio has many matches, but again, it does define it internally.

I would consider pruning it.

jgfouca: I agree. Pruned.

components/cmake/common_setup.cmake
bartgol: Ok, no big deal. I think the logic for skipping second calls works regardless of search mode (provided that FindXXX.cmake sets `<PackageName>_FOUND=TRUE`). But I am not entirely sure, so let's not stop progress based on this detail.

components/cmake/modules/FindCsmShare.cmake
bartgol: I meant adding it inside the `FindCsmShare.cmake` module.

jgfouca: Oh, I see. Yes let me do that.

bartgol (approved):

bartgol (approved):

cime_config/machines/scripts/cmake3ify-package
```diff
@@ -0,0 +1,83 @@
+#! /usr/bin/env python3
```
bartgol, line 1: Is this script really worth committing to the repo? I assume after this big effort, it will likely never need to be used again...

components/eamxx/src/share/io/CMakeLists.txt
```diff
@@ -15,8 +15,8 @@ set_target_properties(scream_io PROPERTIES
 )
 target_include_directories(scream_io PUBLIC ${CMAKE_CURRENT_BINARY_DIR}/modules)
 
-if (DEFINED ENV{ADIOS2_DIR})
-  target_include_directories(scream_io PRIVATE $ENV{ADIOS2_DIR}/include)
+if (DEFINED ENV{ADIOS2_ROOT})
```
bartgol, line 18: Should we instead add this to the [Scorpio.cmake](https://github.com/E3SM-Project/E3SM/blob/master/components/eamxx/cmake/tpls/Scorpio.cmake) target?

cime_config/machines/scripts/cmake3ify-package
jgfouca: I don't care too much one way or the other. This script will potentially be useful in the weeks ahead as we continue on with this effort. This directory is where i've been putting miscellaneous build-related scripts.

components/eamxx/src/share/io/CMakeLists.txt
jgfouca: Probably, but we can refactor that on the scream side.

components/cmake/modules/FindNETCDF.cmake
```diff
@@ -0,0 +1,75 @@
...
+  set (pnetcdf_lib ${pnetcdf_lib})
+  set (netcdf_c_lib ${netcdf_c_lib})
+  set (netcdf_f_lib ${netcdf_f_lib})
+  set (pnetcdf_incdir ${pnetcdf_incdir})
+  set (netcdf_c_incdir ${netcdf_c_incdir})
+  set (netcdf_f_incdir ${netcdf_f_incdir})
+
+  # Create the interface library, and set target properties
+  add_library(netcdf INTERFACE)
+  target_link_libraries(netcdf INTERFACE ${pnetcdf_lib};${netcdf_c_lib};${netcdf_f_lib})
```
bartgol, line 71:
```suggestion
  target_link_libraries(netcdf INTERFACE ${pnetcdf_lib};${netcdf_f_lib};${netcdf_c_lib})
```

rljacob:
This branch looks like it did a lot of work on the netcdf/pnetcdf dependency but that's not mentioned in the title or description.

PR #5965, opened 2023-09-28
Implemented a tentative fix for SSP tests

wlin7:
Future scenario forcing files for new species introduced by v3atm are not ready. 
The cime SSP tests using default v3 configuration therefore failed to be created.

The tentative fix uses config that mixes v3 physics with v2 chemistry. 

Also fixed a couple misuses of eqv operator in linoz-related code that do not 
properly consider the precedence relative to AND/OR operators. The misuses have 
no impact on default v3 configuration, but would cause fatal runtime error when using
the hybrid configuration as implemented here.

[BFB] for existing tests.

### DIFF
--- a/components/eam/cime_config/config_component.xml
+++ b/components/eam/cime_config/config_component.xml
@@ -47,7 +47,7 @@
     <values modifier='additive'>
       <value compset=""                >-mach $MACH</value>
       <value compset="_EAM"            >-phys default</value>
-      <value compset="_EAM%CMIP6_"      >&eam_phys_defaults; &eam_chem_defaults;</value>
+      <value compset="^((?!SSP).)*_EAM%CMIP6_" >&eam_phys_defaults; &eam_chem_defaults;</value>
       <value compset="_EAM%CMIP6-1pctCO2" >&eam_phys_defaults; &eam_chem_defaults;</value>
       <value compset="_EAM%CMIP6-4xCO2"   >&eam_phys_defaults; &eam_chem_defaults;</value>
       <value compset="SSP.*_EAM%CMIP6_"   >&eam_phys_defaults; -chem linoz_mam4_resus_mom_soag -rain_evap_to_coarse_aero</value>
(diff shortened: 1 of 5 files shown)


wlin7:
Tests completed:

SMS_D_Ln5.ne4_oQU240.F2010.chrysalis_intel.eam-cosplite_nhtfrq5, PASS baseline
SMS_D_Ld1.ne30pg2_EC30to60E2r2.WCYCLSSP370.chrysalis_intel.allactive-wcprodssp run ok, diff from baseline that has long been blessed
 ERP_Ld3.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel.allactive-pioroot1  pending

components/eam/cime_config/config_component.xml
```diff
@@ -47,9 +47,10 @@
     <values modifier='additive'>
       <value compset=""                >-mach $MACH</value>
       <value compset="_EAM"            >-phys default</value>
-      <value compset="_EAM%CMIP6_"      >&eam_phys_defaults; &eam_chem_defaults;</value>
+      <value compset="^((?!SSP).)*_EAM%CMIP6_" >&eam_phys_defaults; &eam_chem_defaults;</value>
```
wlin7, line 50: Changed to use this negative lookaround for exclude SSP cases from using the config options at this line. Otherwise would not be able to get rid of '-vbs' option for standard config.

components/eam/src/chemistry/utils/tracer_data.F90
```diff
@@ -1285,7 +1285,7 @@ subroutine read_next_trcdata(state, flds, file )
              cnt3(flds(f)%coords(ZA_TIMDIM)) = 1
              strt3(flds(f)%coords(ZA_TIMDIM)) = recnos(i)
              !!
-             if (file%linoz_v3.eqv..true..or.file%linoz_v2.eqv..true.) then
+             if (file%linoz_v3 .or. file%linoz_v2) then
```
wlin7, line 1288: In the original form, even when linoz_v2 is true, this if-block would not be executed.

rljacob (approved):

ambrad (approved):
The conditionals code changes look good; approving these. Don't know anything about the rest.

jinboxie (approved):

wlin7:
checking on unexpected diffs for `SMS_Ln5.ne30pg2_r05_EC30to60E2r2.BGCEXP_LNDATM_CNPRDCTC_1850` and S`MS_Ln5.ne30pg2_r05_EC30to60E2r2.BGCEXP_LNDATM_CNPRDCTC_20TR`

wlin7:
Merged to next. The BGC tests diffs mentioned above are not a concern for this PR. The tests were created without specifying `--pesfile config_pes_tests.xml`. When specified, the tests are BFB with baselines. Also refer to #5992.

wlin7:
Merged to master.

PR #5966, opened 2023-09-30
For alvarez (internal test machine for pm-cpu), update machine file settings

ndkeen:
After maintenance, update modules and startup settings.
Only impacts alvarez test machine at NERSC.
[bfb]


ndkeen:
merged to next

rljacob (approved):

PR #5969, opened 2023-10-03
Add CICE-QC testing script

darincomeau:
Adds a script to perform "CICE-QC" test, which determines if non-BFB changes in the sea ice model are climate-changing. This has been adapted from the [CICE implementation](https://cice-consortium-cice.readthedocs.io/en/main/user_guide/ug_testing.html#code-validation-test-non-bi\t-for-bit-validation) for MPAS-Seaice. 
Documentation on usage is provided in a `README.md` alongside test script.

[BFB]

### DIFF
--- a/components/mpas-seaice/testing/cice-qc/README.md
+++ b/components/mpas-seaice/testing/cice-qc/README.md
@@ -1,0 +1,25 @@
+# Quality Control (QC) testing for MPAS-Seaice
+
+This testing script to determine if the answers are climate-changing as determined by a 2 stage t-test was adapted from [CICE](https://cice-consortium-cice.readthedocs.io/en/main/user_guide/ug_testing.html#code-validation-test-non-bit-for-bit-validation).
+
+## Setup
+
+1. Setup and build a baseline and test case with recommended minimum resolution `RES=TL319_EC30to60E2r2`, and `COMPSET=DTESTM-JRA1p5`.
+2. Turn on the ice thickness analysis member with the following namelist options in `user_nl_mpassi`:
+
+   config_AM_thicknesses_enable = .true.
+   config_AM_thicknesses_compute_on_startup = .true.
+   config_AM_thicknesses_write_on_startup = .true.
+
+3. In the run directory's `streams.seaice` file, modify the `output` stream to have daily snapshot output by setting `output_interval="00-00-01_00:00:00"` (default is "none"). Copy this modified `streams.seaice` file to the case directory in `SourceMods/src.mpassi/.`.
+4. Run each case for (at least) 5 years.
+
+## Usage
+
+    python mpas-seaice.t-test.py $BASE $TEST
+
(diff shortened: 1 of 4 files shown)


darincomeau:
This test uses two netcdf files, which I've put in the public inputdata directory here: https://web.lcrc.anl.gov/public/e3sm/inputdata/ice/mpas-seaice/general/

The path to these files is currently hard-coded in the python script to the LCRC path, so this will only work on those machines as is.

eclare108213:
Yes, the latest qc tests that I ran are in /lcrc/group/e3sm/ac.eclare/E3SM-Polar:

D12.qcPR25.emc.qcPR25.thermo_itd2.eclare108213.anvil
D12.qcPR29.emc.qcPR29.merge-snow-integration-master.E3SM-Project.anvil

Those passed -- I don't have any that failed :)

proteanplanet:
Back-to-back B-case simulations of Icepack and ColPkg:

/lcrc/group/e3sm/ac.afroberts/E3SM-Polar/B12.merge.icepack.icepack-integration.eclare108213.anvil

k000 --> config_column_physics_type = 'column_package'
k001 --> config_column_physics_type = 'icepack'

I'd love to know what the QC test shows.

eclare108213:
The QC scripts here need to be modified for the k000 / k001 directory structure produced using Andrew's scripts. E.g. my modified QC scripts are in /home/ac.eclare/MPAS-Analysis.

proteanplanet:
@eclare108213 @darincomeau N.B. I'm changing that directory structure, because it doesn't play well across a range of E3SM applications (PACE, zppy), so no need to rewire this work.  I'll rewire my script.

darincomeau:
> Back-to-back B-case simulations of Icepack and ColPkg:

> /lcrc/group/e3sm/ac.afroberts/E3SM-Polar/B12.merge.icepack.icepack-integration.eclare108213.anvil

> k000 --> config_column_physics_type = 'column_package'
> k001 --> config_column_physics_type = 'icepack'

> I'd love to know what the QC test shows.

@proteanplanet this test requires non-standard output, which is daily iceThickness snapshots. Unfortunately I can't run it on the pair of runs you've pointed to. 

@eclare108213 you've been turning this output on for your tests - do you have an automated way of modifying the `streams.seaice` file?

darincomeau:
and thanks @eclare108213 for pointing to your changes. I should loosen the requirement here that the run directory doesn't need to be named exactly `run`.

proteanplanet (approved):
In looking through the code, I think this testing would be most versatile if test runs could be automatically generated with necessary output automatically turned on.  Adding that functionality here is quite a bit of work and it's also unneeded because of the E3SM-Polar-Developer.sh script.  As a result, I think the best approach would be to leave this script as is, and call it from that test script with a command line option that sets up the simulations, then performs the T-test when they are finished. 

Approved.

jonbob:

jonbob:
merged to master

PR #5971, opened 2023-10-04
For pm-cpu/alvarez, adjust pelayouts to remove performance work-around with CPL_PSTRID after maintenance

ndkeen:
Adjust pelayouts for pm-cpu/alvarez.
After late September Perlmutter maintenance, some HW/SW changes seems to have resolved the issue we were seeing with slower-than-expected rearranger performance (showing as time in CPL).
There is still work to further improve the pelayouts on pm-cpu, but this PR simply removes the
CPL_PSTRID=8 work-around.

[bfb] except there are coupler namelist diffs due to different number of ranks


rljacob (approved):

ndkeen:
merged to next

PR #5972, opened 2023-10-05
Convert blas/lapack/mkl to new cmake style

jgfouca:
This was a more challenging task than I thought it would. FindBLAS / FindLAPACK are built in to cmake and can also handle MKL stuff via the BLA_VENDOR setting (if you set it to Intel... you get various flavors of MKL). The find_package for blas/lapack is pretty smart and should do the right thing most of the time. Machine POCs should refine the settings on their machines using https://cmake.org/cmake/help/latest/module/FindBLAS.html#blas-lapack-vendors as as guide. I suspect many of these cache files were copy/pasted without a ton of consideration.

As an example, pm-cpu_nvidia is one platform where you will get DIFFs if you don't use the right blas/lapack. To set this up, I added the following to the environment block for that machine/compiler combo:
```
    <environment_variables compiler="nvidia">
      <env name="BLAS_ROOT">$SHELL{if [ -z "$BLAS_ROOT" ]; then echo /opt/nvidia/hpc_sdk/Linux_x86_64/22.7/compilers; else echo "$BLAS_ROOT"; fi}</env>
      <env name="LAPACK_ROOT">$SHELL{if [ -z "$LAPACK_ROOT" ]; then echo /opt/nvidia/hpc_sdk/Linux_x86_64/22.7/compilers; else echo "$LAPACK_ROOT"; fi}</env>
      <env name="BLA_VENDOR">NVHPC</env>
    </environment_variables>
```

The documentation indicates that the find modules for blas may be smart enough to auto detect the bla_vendor:
```
BLA_VENDOR
Set to one of the [BLAS/LAPACK Vendors] to search for BLAS only from the specified vendor.
If not set, all vendors are considered.
```

Once again, I am impressed at how much clutter is removed when we do things t "right" (using find_package instead of packing stuff into SLIBS) way.

[BFB]


cime_config/machines/cmake_macros/gnu_melvin.cmake
```diff
@@ -11,4 +11,3 @@ endif()
 if (MPILIB STREQUAL mpi-serial AND NOT compile_threaded)
   set(PFUNIT_PATH "$ENV{SEMS_PFUNIT_ROOT}")
 endif()
-string(APPEND SLIBS " -lblas -llapack")
```
bartgol, line 14: Is melvin still around? I thought it was no longer online.

bartgol (approved):

rljacob (changes requested):
Look at CircleCI fail.

cime_config/machines/cmake_macros/gnu_melvin.cmake
jgfouca: Yes, we badly need to go through all our configs and purge unused machines. I think we still have a lot of NCAR machine configs that we never use.

jgfouca:
@rljacob , I thought blas and lapack were required packages for E3SM? It looks like the circleci container doesn't install them? @lukaszlacinski

rljacob:
That's a build-only test so it must have been possible to build without them.

rljacob (approved):
Circle CI fail is because container doesn't have blas/lapack.  Need to redo container or switch to GH actions (with a newer container).

bartgol:
@jgfouca perhaps we need to set the correct BLA_VENDOR variable in `gnu_signularity.cmake`?

We could also temporarily hack make the singularity build, to cat the e3sm.bldlog if case.build fails. That _may_ show some more context from the cmake log.

dqwu (approved):

bartgol (approved):

PR #5976, opened 2023-10-09
Fix nvidia_pm-cpu diffs for some tests

jgfouca:
Be explicit about which hdf5 and blas/lapack libraries to link so that later -L options don't mess things up.

Also, make some minor enhancements to the compare-flags tool in order to get more useful output when comparing unordered flags.

[BFB]


jgfouca:
I've confirmed this fixes the DIFF for test SMS_Ln5.ne4pg2_oQU480.F2010.pm-cpu_nvidia.eam-thetahy_sl_pg2_ftype0

jgfouca:
Working on confirming the other DIFFs are fixed as well.

bartgol (approved):
If this fixes the link line, it's fine. I am still puzzled though: did you figure out who is injecting `/usr/lib64` in the link line?

jgfouca:
Confirmed all diffs are fixed. I'm not sure how `-L/usr/lib64` got in the link args.

PR #5982, opened 2023-10-11
Fix MMF heating for convective gravity wave scheme

whannah1:
Previously the MMF heating tendency used for the convective gravity wave parameterization included the effect of radiation. This moves the calculation so that radiative heating can be removed at all levels.

[non-BFB] for all MMF tests

### DIFF
--- a/components/eam/src/physics/crm/crm_physics.F90
+++ b/components/eam/src/physics/crm/crm_physics.F90
@@ -1587,7 +1587,7 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf2d, cam_in, cam_out, 
          end do
 
 #if defined(MMF_PAM)
-         ! Currently PAM is coupled the CRM to all levels, so only add rad to levels above CRM
+         ! Currently PAM is coupled at all levels, so only add rad to levels above CRM top
          ptend(c)%s(1:ncol, 1:pver-crm_nz) = qrs(1:ncol,1:pver-crm_nz) + qrl(1:ncol,1:pver-crm_nz)
 #else
          ! The radiation tendencies in the GCM levels above the CRM and the top 2 CRM levels are set to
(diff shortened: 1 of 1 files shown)


components/eam/src/physics/crm/crm_physics.F90
```diff
@@ -1611,6 +1603,14 @@ subroutine crm_physics_tend(ztodt, state, tend, ptend, pbuf2d, cam_in, cam_out,
             end do
          end do
 
+         !------------------------------------------------------------------------------------------
+         ! set convective heating tendency for gravity wave drag
+         !------------------------------------------------------------------------------------------
+         if (ttend_dp_idx > 0) then
+            call pbuf_get_field(pbuf_chunk, ttend_dp_idx, ttend_dp)
+            ttend_dp(1:ncol,1:pver) = ( ptend(c)%s(1:ncol,:pver) - qrs(1:ncol,:pver) - qrl(1:ncol,:pver) )/cpair
```
jlee1046, line 1611: I can see that you are subtracting radiative heating from temperature tendency. If ttend_dp is an input for gravity wave drag, is there a reason to remove radiative heating ?

jlee1046 (approved):
All looks good to me.

components/eam/src/physics/crm/crm_physics.F90
whannah1: This is meant to mimic how it all works in the non-MMF cases, in which the heating tendencies are taken directly from the ZM scheme and no radiative heating is included.

jlee1046: I see. thanks for clarifying it.

jlee1046 (approved):
I approve this PR

brhillman:
Merged to next.

PR #5984, opened 2023-10-11
Drop CICE from MMF compsets

whannah1:
Remove CICE dependency from MMF compsets to prepare for upcoming CICE deprecation

### DIFF
--- a/components/eam/cime_config/config_compsets.xml
+++ b/components/eam/cime_config/config_compsets.xml
@@ -133,27 +133,27 @@
 
   <compset>
     <alias>FSCM-ARM97-MMF1</alias>
-    <lname>ARM97_EAM%SCM-MMF1_ELM%SP_CICE%PRES_DOCN%DOM_SROF_SGLC_SWAV</lname>
+    <lname>ARM97_EAM%SCM-MMF1_ELM%SP_SICE_DOCN%DOM_SROF_SGLC_SWAV</lname>
   </compset>
 
   <compset>
     <alias>FSCM-ARM97-MMF2</alias>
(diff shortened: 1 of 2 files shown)


jlee1046 (approved):
This PR prepares for MMF tests and compsets for the deprecation of CICE. All looks good to me. I approve this PR

components/eam/cime_config/config_compsets.xml
```diff
@@ -133,27 +133,27 @@
 
   <compset>
     <alias>FSCM-ARM97-MMF1</alias>
-    <lname>ARM97_EAM%SCM-MMF1_ELM%SP_CICE%PRES_DOCN%DOM_SROF_SGLC_SWAV</lname>
+    <lname>ARM97_EAM%SCM-MMF1_ELM%SP_SICE_DOCN%DOM_SROF_SGLC_SWAV</lname>
```
rljacob, line 136: The equivalent of CICE%PRES is MPASSI%PRES.   SICE is the stub sea ice model and I think would mean there is no sea ice at all in the arctic or antarctic.  Might run but would not be Earth.

whannah1: Yea, but we don't need it for this compset since it's a single column over land, so I figured we don't want to bother building MPAS-SI if we don't need to. Even for the other SCM RICO compset, the column is at a low enough latitude that we don't have to worry about sea ice.

rljacob: I see you used MPASSI below. Did you intend SICE here?

whannah1: Yes, because it's SCM over land like I mentioned above.

whannah1:
Closing this because the changes are going to be handled more thoroughly by PR #6044

PR #5985, opened 2023-10-11
add hdepth scaling factor for convective gravity waves

whannah1:
Adds a namelist variable "hdepth_scaling_factor" for the convective gravity wave scheme that will be needed for tuning the QBO in v3. The default value of 1 will retain the current behavior.

[BFB]
[NML] new nml variable hdepth_scaling_factor for all tests involving eam.

### DIFF
--- a/components/eam/src/physics/cam/gw_drag.F90
+++ b/components/eam/src/physics/cam/gw_drag.F90
@@ -88,6 +88,9 @@ module gw_drag
   ! Convective heating rate conversion factor, default is 20.0_r8
   real(r8) :: gw_convect_hcf = 20.0_r8
 
+  ! Scaling factor for the heating depth
+  real(r8) :: hdepth_scaling_factor = 1
+
   ! Whether or not to enforce an upper boundary condition of tau = 0.
   ! (Like many variables, this is only here to hold the value between
   ! the readnl phase and the init phase of the CAM physics; only gw_common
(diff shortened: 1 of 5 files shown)


whannah1:
e3sm_atm_developer passes baseline comparison on Perlmutter except for the expected NLFAIL from the new namelist variable.

components/eam/src/physics/cam/gw_drag.F90
```diff
@@ -88,6 +88,9 @@ module gw_drag
   ! Convective heating rate conversion factor, default is 20.0_r8
   real(r8) :: gw_convect_hcf = 20.0_r8
 
+  ! Scaling factor for the heating depth
+  real(r8) :: hdepth_scaling_factor = 1
```
singhbalwinder, line 92: Since it is `real(r8)`, it would be better to initialize it with `_r8` (`real(r8):: hdepth_scaling_factor = 1.0_r8`). Maybe it doesn't matter but it will be good for consistency.

singhbalwinder (approved):
Looks great! I have left one minor comment.

components/eam/src/physics/cam/gw_drag.F90
jjbenedict: I agree, best to add `_r8`

jjbenedict (changes requested):
Just the _r8 addition to hdepth_scaling_factor in gw_drag.F90.

components/eam/src/physics/cam/gw_drag.F90
whannah1: I thought it was better to omit it since the compiler will ensure that the type matches?

jjbenedict (approved):

wlin7:
Merged to next.

wlin7:
Merged to master. NML diffs to be blessed.

PR #5989, opened 2023-10-12
Add split-explicit AB2 time stepping capability to the ocean component

hyungyukang:
The second-order Adams Bashforth (AB2) time stepping method is applied to the baroclinic system of MPAS-Ocean to enable faster computation of the baroclinic system. The AB2 time stepping method, one of multistep methods, computes time stepping procedure (Stage 1~3) once per time step, while the predictor-corrector which is a default scheme computes it twice per time step. In practice, the AB2 method can provide a speedup of 1.5x to 1.8x. Due to its high sensitivity to time step size, the predictor-corrector schem is only applied to the layer thickness equation.

All subroutines and the barotropic system advance (Stage 2) in the AB2 code are the same as used in the split-explicit code (mpas_ocn_time_integration_split.F).

To enable this option,  users need to set `config_time_integrator = 'split_explicit_ab2'`.

Results of G-case (100 years) and WCYCL1850 (30 years) and more details of the method can be found here: https://acme-climate.atlassian.net/wiki/spaces/ImPACTS/pages/3826385232/AB2+time+stepping+for+the+baroclinic+system

[NML]
[non-BFB]

### DIFF
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split_ab2.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split_ab2.F
@@ -3222,6 +3222,31 @@ module ocn_time_integration_split_ab2
 
       self_attraction_and_loading_beta = config_self_attraction_and_loading_beta
 
+      !-----------------------------------------------------------------
+      if ( config_do_restart ) then
+         block => domain % blocklist
+         call mpas_pool_get_subpool(block%structs, 'state', statePool)
+         call mpas_pool_get_subpool(statePool, 'tracers', tracersPool)
+         call mpas_pool_get_subpool(block%structs, 'mesh', meshPool)
+
+         call mpas_pool_get_dimension(block % dimensions, 'nVertLevels', nVertLevels)
+         call mpas_pool_get_dimension(block % dimensions, 'nEdges', nEdges)
+
+         call mpas_pool_get_array(statePool, 'normalVelocity', &
+                                              normalVelocity, 1)
+
+         call mpas_pool_get_array(meshPool, 'minLevelEdgeBot', minLevelEdgeBot)
+         call mpas_pool_get_array(meshPool, 'maxLevelEdgeTop', maxLevelEdgeTop)
+
+         do iEdge = 1, nEdges
+            kmax  = maxLevelEdgeTop(iEdge)
+            ! normalVelocity=0 on land cells
+            do k = kmax+1, nVertLevels
+               normalVelocity(k,iEdge) = 0.0_RKIND
+            enddo
+         enddo ! edge loop
+
+      end if
    end subroutine ocn_time_integration_split_ab2_init!}}}
 
 !***********************************************************************
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split_ab2.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split_ab2.F
@@ -754,8 +754,7 @@ module ocn_time_integration_split_ab2
 
                !$omp parallel
                !$omp do schedule(runtime) &
-               !$omp private(k, cell1, cell2, uTemp, sshGrad, &
-               !$omp         normalThicknessFluxSum, thicknessSum)
+               !$omp private(k, cell1, cell2, sshGrad)
                do iEdge = 1, nEdgesOwned
                   cell1 = cellsOnEdge(1,iEdge)
                   cell2 = cellsOnEdge(2,iEdge)
@@ -796,8 +795,7 @@ module ocn_time_integration_split_ab2
                if ( j == 1 ) then
                   !$omp parallel
                   !$omp do schedule(runtime) &
-                  !$omp private(k, cell1, cell2, uTemp, &
-                  !$omp         normalThicknessFluxSum, thicknessSum)
+                  !$omp private(k, cell1, cell2, sshGrad)
                   do iEdge = 1, nEdgesOwned
                      cell1 = cellsOnEdge(1,iEdge)
                      cell2 = cellsOnEdge(2,iEdge)
@@ -834,8 +832,7 @@ module ocn_time_integration_split_ab2
 
                   !$omp parallel
                   !$omp do schedule(runtime) &
-                  !$omp private(k, cell1, cell2, uTemp, &
-                  !$omp         normalThicknessFluxSum, thicknessSum)
+                  !$omp private(k, cell1, cell2, sshGrad)
                   do iEdge = 1, nEdgesOwned
                      cell1 = cellsOnEdge(1,iEdge)
                      cell2 = cellsOnEdge(2,iEdge)
@@ -873,8 +870,7 @@ module ocn_time_integration_split_ab2
 
             !$omp parallel
             !$omp do schedule(runtime) &
-            !$omp private(k, cell1, cell2, uTemp, &
-            !$omp         normalThicknessFluxSum, thicknessSum)
+            !$omp private(k, normalThicknessFluxSum, thicknessSum)
             do iEdge = 1, nEdgesOwned
                ! thicknessSum is initialized outside the loop because
                ! on land boundaries maxLevelEdgeTop=0, but we want to
@@ -2045,7 +2041,6 @@ module ocn_time_integration_split_ab2
             !$omp end do
             !$omp end parallel
 #endif
-
                
             ! Compute layerThicknessTend again
             !   layerThickTend from 0.5 * (h_{n} + h_{*})
@@ -2075,8 +2070,8 @@ module ocn_time_integration_split_ab2
                !$omp end do
                !$omp end parallel
 #endif
-               nEdges = nEdgesHalo(config_num_halos-1 )
-               allocate(uTemp(nVertLevels,nEdges))
+
+               allocate(uTemp(nVertLevels,nEdgesAll))
 
                ! velocity for normalVelocityCorrectionection is
                ! normalBarotropicVelocity +
@@ -2113,7 +2108,7 @@ module ocn_time_integration_split_ab2
 
                !$omp parallel
                !$omp do schedule(runtime) &
-               !$omp private(k, uTemp, normalThicknessFluxSum, &
+               !$omp private(k, normalThicknessFluxSum, &
                !$omp         thicknessSum, normalVelocityCorrection)
                do iEdge = 1, nEdges
 
@@ -2151,8 +2146,8 @@ module ocn_time_integration_split_ab2
 
 #ifdef MPAS_OPENACC
                !$acc parallel loop gang vector &
-               !$acc          present(maxLevelCell, zMid, layerThickness, zTop, bottomDepth, ssh, &
-               !$acc                  minLevelCell) &
+               !$acc          present(maxLevelCell, zMid, layerThickness, zTop, &
+               !$acc                  bottomDepth, ssh, minLevelCell) &
                !$acc          private(k)
 #else
                !$omp parallel
@@ -2182,8 +2177,10 @@ module ocn_time_integration_split_ab2
                   end do
 
                end do
+#ifndef MPAS_OPENACC
                !$omp end do
                !$omp end parallel
+#endif
 
                call ocn_vert_transport_velocity_top( &
                     verticalMeshPool, layerThicknessCur, &
@@ -2202,7 +2199,7 @@ module ocn_time_integration_split_ab2
                do k = minLevelCell(iCell), maxLevelCell(iCell)
                   ! this is h_{n+1} = h_{n} + dt * hTend_{n+0.5}
                   layerThicknessNew(k,iCell) = &
-                                        layerThicknessCur(k,iCell) + &
+                                         layerThicknessCur(k,iCell) + &
                                      dt*layerThicknessTend(k,iCell)
                end do
                end do
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split_ab2.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split_ab2.F
@@ -1567,7 +1567,7 @@ module ocn_time_integration_split_ab2
 
          !$omp parallel
          !$omp do schedule(runtime)
-         do iEdge = 1, nEdgesOwned
+         do iEdge = 1, nEdgesAll
             barotropicThicknessFlux(iEdge) = &
             barotropicThicknessFlux(iEdge) &
                    /(nBtrSubcycles*config_btr_subcycle_loop_factor)
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split_ab2.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split_ab2.F
@@ -863,6 +863,7 @@ module ocn_time_integration_split_ab2
 
                   enddo ! iEdge
                   !$omp end do
+                  !$omp end parallel
 
                end if ! j
 
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split_ab2.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split_ab2.F
@@ -3238,10 +3238,15 @@ module ocn_time_integration_split_ab2
          call mpas_pool_get_array(meshPool, 'minLevelEdgeBot', minLevelEdgeBot)
          call mpas_pool_get_array(meshPool, 'maxLevelEdgeTop', maxLevelEdgeTop)
 
+         ! normalVelocity=0 on inactive cells
          do iEdge = 1, nEdges
-            kmax  = maxLevelEdgeTop(iEdge)
-            ! normalVelocity=0 on land cells
-            do k = kmax+1, nVertLevels
+            ! at top
+            do k = 1, minLevelEdgeBot(iEdge)-1
+               normalVelocity(k,iEdge) = 0.0_RKIND
+            enddo
+
+            ! at bottom
+            do k = maxLevelEdgeTop(iEdge)+1, nVertLevels
                normalVelocity(k,iEdge) = 0.0_RKIND
             enddo
          enddo ! edge loop
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split_ab2.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split_ab2.F
@@ -9,13 +9,16 @@
 !
 !  ocn_time_integration_split_ab2
 !
-!> \brief MPAS ocean split explicit time integration scheme
-!> \author Mark Petersen, Doug Jacobsen, Todd Ringler
-!> \date   September 2011
+!> \brief MPAS ocean split explicit AB2 time integration scheme
+!> \author Hyun Kang (Oak Ridge National Laboratory)
+!> \date   October 2023
 !> \details
-!>  This module contains the routine for the split explicit
-!>  time integration scheme
-!
+!>  This module contains the routine for the split explicit time
+!>  integration scheme, where the second-order Adams Bashforth (AB2)
+!>  time stepping method is applied to the baroclinic system.
+!>  This module was developed based on the original split explicit code
+!>  written by Mark Petersen, Doug Jacobsen, and Todd Ringler.
+!>
 !-----------------------------------------------------------------------
 
 module ocn_time_integration_split_ab2
@@ -78,9 +81,6 @@ module ocn_time_integration_split_ab2
       subcycleGroupName = 'subcycleFields', &! name for cached halo exch
       finalBtrGroupName = 'finalBtrFields'   ! name for cached halo exch
 
-   logical :: &
-      unsplit           ! flag for unsplit_explicit time option
-
    integer, dimension(:), allocatable :: &
       numClinicIterations  ! number of baroclinic iterations for each
                            ! outer timestep iteration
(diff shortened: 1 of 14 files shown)


sarats:
Tangentially, we need to discuss if we wish to include this PR in our planned high res GPU simulations
https://acme-climate.atlassian.net/wiki/spaces/ImPACTS/pages/3902799937/High-res+Ocean+GPU+simulations

hyungyukang:
> Tangentially, we need to discuss if we wish to include this PR in our planned high res GPU simulations
> https://acme-climate.atlassian.net/wiki/spaces/ImPACTS/pages/3902799937/High-res+Ocean+GPU+simulations

@sarats , I will run this code on GPU and report here. Thanks!

hyungyukang:
> Tangentially, we need to discuss if we wish to include this PR in our planned high res GPU simulations
> https://acme-climate.atlassian.net/wiki/spaces/ImPACTS/pages/3902799937/High-res+Ocean+GPU+simulations

@sarats , After modifying some OpenACC directives, this code runs well on GPUs as tested on PM-GPU for both the standalone MPAS-O and the ocean-ice coupled (GMPAS-IAF) case. 

This PR does not address the issue of running simulations on PM-GPU. I've referenced @grnydawn 's fix for PM-GPU at here: https://github.com/E3SM-Project/E3SM/tree/ykim/mpas/gcase.

mark-petersen:
@hyungyukang can you comment here on the exact tests you did for performance and quality? Do you have scaling plots for particular resolutions and machines? If information is in slide format, you can just upload them.

hyungyukang:
@mark-petersen , I've run the standalone MPAS-O on RRS18to6v3 mesh with a real-world configuration using atmospheric forcing to check performance and quality.
- Computational performance (Runtime was measured from 'se tilmestep')

image

- Kinetic energy field (k=1) comparison
   - AB2 uses 75% of del4 diffusion coefficient
 
image

hyungyukang:
@mark-petersen , Thanks for testing! I don't think we need to include two previous steps, but I'm going to look into it today. For QU240 restart tests, AB2 passes ```QU240_PHC_restart_test``` but failed ```QU240_WOA23_restart_test```. What's the difference between two QU240 tests?

xylar:
> For QU240 restart tests, AB2 passes QU240_PHC_restart_test but failed QU240_WOA23_restart_test. What's the difference between two QU240 tests?

That sounds very odd because the difference between the two is just the initial condition (i.e. the T and S fields being interpolated from).  Are you sure you ran the PHC restart test with `config_time_integrator = 'split_explicit_ab2'`?

hyungyukang:
> > For QU240 restart tests, AB2 passes QU240_PHC_restart_test but failed QU240_WOA23_restart_test. What's the difference between two QU240 tests?
> 
> That sounds very odd because the difference between the two is just the initial condition (i.e. the T and S fields being interpolated from). Are you sure you ran the PHC restart test with `config_time_integrator = 'split_explicit_ab2'`?

@xylar , Another odd thing is that AB2 passes the baroclinic channel restart tests on both Compy and Perlmutter...
Anyway, I left my Compy RSA authenticator at home, so I will setup the same tests on Perlmutter and look into results and codes again. Please let me know if you have any suggestions on a test side!

mark-petersen:
The restart tests work and are bfb for active ocean cells. For some reason on AB2 deeper land edges get that dummy value of -1e34, but only for restart runs and not for the very first forward run. I'm looking into it now.
```
cd ocean/global_ocean/QU240/WOA23/restart_test
ncdump -v normalVelocity restart_run/output.nc |tail -n3
  0, 0, 0, 0, -1e+34, -1e+34, -1e+34, -1e+34, -1e+34, -1e+34, -1e+34, -1e+34,
    -1e+34, -1e+34, -1e+34, -1e+34 ;
}
```

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split_ab2.F
```diff
@@ -0,0 +1,3257 @@
...
+         call mpas_pool_get_array(meshPool, 'minLevelEdgeBot', minLevelEdgeBot)
+         call mpas_pool_get_array(meshPool, 'maxLevelEdgeTop', maxLevelEdgeTop)
+
+         do iEdge = 1, nEdges
+            kmax  = maxLevelEdgeTop(iEdge)
+            ! normalVelocity=0 on land cells
+            do k = kmax+1, nVertLevels
+               normalVelocity(k,iEdge) = 0.0_RKIND
+            enddo
+         enddo ! edge loop
```
xylar, line 3247: I think you need to do this over the inactive cells at the top of the domain, too.  You are probably not testing with configurations that use this feature, so it probably doesn't show up as a problem.
```suggestion
         do iEdge = 1, nEdges
            ! normalVelocity=0 on land cells
            do k = 1, minLevelEdgeBog(iEdge)-1
               normalVelocity(k,iEdge) = 0.0_RKIND
            enddo
            kmax  = maxLevelEdgeTop(iEdge)
            do k = kmax+1, nVertLevels
               normalVelocity(k,iEdge) = 0.0_RKIND
            enddo
         enddo ! edge loop
```

hyungyukang: @xylar, Thanks a lot! I'm running the nightly suite using with my fix, but I'm going to add this and test again.

And you are right. My test on Compy was using the default time integration. I just confirmed it. I think I made a mistake when I recloned and recompiled... Thanks again.

xylar: @hyungyukang, your changes look perfect.  Thanks!

hyungyukang:
@mark-petersen and @xylar, Thanks for reviewing and your suggestions! Based on your findings and suggestions, I added a fix to `subroutine ocn_time_integrated_split_ab2_init` to initialize `normalVelocity=0` on inactive cells when restarting. This also fixes the G-case restart issue we discussed yesterday.

<details>
<summary>code change</summary>

```fortran
      if ( config_do_restart ) then
         block => domain % blocklist
         call mpas_pool_get_subpool(block%structs, 'state', statePool)
         call mpas_pool_get_subpool(statePool, 'tracers', tracersPool)
         call mpas_pool_get_subpool(block%structs, 'mesh', meshPool)

         call mpas_pool_get_dimension(block % dimensions, 'nVertLevels', nVertLevels)
         call mpas_pool_get_dimension(block % dimensions, 'nEdges', nEdges)

         call mpas_pool_get_array(statePool, 'normalVelocity', &
                                              normalVelocity, 1)

         call mpas_pool_get_array(meshPool, 'minLevelEdgeBot', minLevelEdgeBot)
         call mpas_pool_get_array(meshPool, 'maxLevelEdgeTop', maxLevelEdgeTop)

         ! normalVelocity=0 on inactive cells
         do iEdge = 1, nEdges
            ! at top
            do k = 1, minLevelEdgeBot(iEdge)-1
               normalVelocity(k,iEdge) = 0.0_RKIND
            enddo

            ! at bottom
            do k = maxLevelEdgeTop(iEdge)+1, nVertLevels
               normalVelocity(k,iEdge) = 0.0_RKIND
            enddo
         enddo ! edge loop

      end if
``` 

</details>

<details>
<summary>nightly suite output</summary>

00:15 PASS ocean_baroclinic_channel_10km_default
00:15 PASS ocean_baroclinic_channel_10km_threads_test
00:09 PASS ocean_baroclinic_channel_10km_decomp_test
00:09 PASS ocean_baroclinic_channel_10km_restart_test
04:24 PASS ocean_global_ocean_QU240_mesh
03:48 PASS ocean_global_ocean_QU240_WOA23_init
03:46 PASS ocean_global_ocean_QU240_WOA23_performance_test
07:41 PASS ocean_global_ocean_QU240_WOA23_restart_test
06:09 PASS ocean_global_ocean_QU240_WOA23_decomp_test
05:44 PASS ocean_global_ocean_QU240_WOA23_threads_test
06:30 PASS ocean_global_ocean_QU240_WOA23_analysis_test
02:38 PASS ocean_global_ocean_QU240_WOA23_RK4_performance_test
05:50 PASS ocean_global_ocean_QU240_WOA23_RK4_restart_test
05:27 PASS ocean_global_ocean_QU240_WOA23_RK4_decomp_test
05:28 PASS ocean_global_ocean_QU240_WOA23_RK4_threads_test
00:00 PASS ocean_global_ocean_QUwISC240_mesh
00:00 PASS ocean_global_ocean_QUwISC240_WOA23_init
05:31 PASS ocean_global_ocean_QUwISC240_WOA23_performance_test
00:48 PASS ocean_ice_shelf_2d_5km_z-star_restart_test
00:45 PASS ocean_ice_shelf_2d_5km_z-level_restart_test
00:18 PASS ocean_ziso_20km_default
00:12 PASS ocean_ziso_20km_with_frazil
Total runtime 66:37
PASS: All passed successfully!

</details>

And double checked I'm using `config_time_integrator = 'split_explicit_ab2'`

hyungyukang:
The attached is slides about this AB2 implementation.
[MPASO_AB2_timestep.pdf](https://github.com/E3SM-Project/E3SM/files/13179187/MPASO_AB2_timestep.pdf)

mark-petersen:
Thanks! I tested with the nightly suite with gnu and intel, and everything passes.

hyungyukang:
I have tested the AB2 time stepping method with updated weights for the barotropic subcycling in split-explicit solver by @dengwirda (https://github.com/E3SM-Ocean-Discussion/E3SM/pull/48). Thanks to this improvement, I was able to increase the barotropic time step size up to 75 s (default = 60 s). I ran the standalone MPAS-O using the real-world configurations with atmospheric forcing on `EC60to30v3` mesh. The below figure shows the computational performance:

image

---
To check stability, kinetic energy average & maximum from global stats are displayed (Orange, green, and red solids are overlapped. And AB2 uses):

image

---
I also performed `GMPAS-DIB-IAF-DISMF` on `T62_ECwISC30to60E3r2` mesh for quick check of computational performance  of ocean-ice coupled runs on `pm-cpu` (5-day integration using 1024 cores).

image

sarats:
@hyungyukang I presume your latest results are trying to confirm that your AB2 method works with the updated weights, correct? 

It appears that this PR should probably go in now to keep life simple - instead of piling on any new features ;).

mark-petersen:
Added last commit on barotropic subcycling weights as a separate PR at https://github.com/E3SM-Project/E3SM/pull/6035.

mark-petersen (changes requested):

mark-petersen:
@hyungyukang please also add a note to the `config_n_ts_iter` description in `Registry.xml` on how the iteration works when `config_time_integrator = 'split_explicit_ab2'`

mark-petersen:
After checking with @vanroekel (last week) and the MPAS-DevOps team, we are changing AB2 to the default baroclinic time-stepping method. That changes this PR from stealth to non-bfb and potentially climate changing, though we expect the changes to be small and similar to an ensemble member with an initial perturbation.

@hyungyukang please change the default here to ab2:
```
namelist_files/namelist_defaults_mpaso.xml:52:
<config_time_integrator>'split_explicit'</config_time_integrator>
```

hyungyukang:
> After checking with @vanroekel (last week) and the MPAS-DevOps team, we are changing AB2 to the default baroclinic time-stepping method. That changes this PR from stealth to non-bfb and potentially climate changing, though we expect the changes to be small and similar to an ensemble member with an initial perturbation.
> 
> @hyungyukang please change the default here to ab2:
> 
> ```
> namelist_files/namelist_defaults_mpaso.xml:52:
> <config_time_integrator>'split_explicit'</config_time_integrator>
> ```

@mark-petersen , Thanks a lot. Sure, will do!

mark-petersen:
I ran one year of this PR plus https://github.com/E3SM-Project/E3SM/pull/6035 with gnu on chrysalis in MPAS-O standalone, using compass `ec30to60` and `ecwisc30to60` suites, but with cached mesh and init steps to save time. This then takes the model through the spin up process, then I extended the `simulation` step to 360 days. The results for AB2 look very good. It is stable with the same spin-up procedure, and min and max on temperature and salinity remain constant. 

The WC time was reduced by 40%, which is very good. Specifically, the 'time integration' timer (no i/o) for 360 days on 512 cores on chrysalis changed from 7330s on master to 4376s with this PR (plus https://github.com/E3SM-Project/E3SM/pull/6035)

cbegeman:
@hyungyukang and @mark-petersen This is the issue I mentioned on slack https://github.com/E3SM-Project/E3SM/issues/6040 and I think you need something analogous to this bugfix for AB2 https://github.com/E3SM-Project/E3SM/pull/6041.

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split_ab2.F
```diff
@@ -0,0 +1,3262 @@
...
+            ! normalBarotropicVelocityNew =
+            ! normalBarotropicVelocityNew +
+            ! normalBarotropicVelocitySubcycleNEW
+            ! This accumulates the sum. If the Barotropic Coriolis
+            ! iteration is limited to one, this could
+            ! be merged with the above code.
+
+            !$omp parallel
+            !$omp do schedule(runtime)
+            do iEdge = 1, nEdgesAll
```
mark-petersen, line 1542: @hyungyukang, per the bug fix from @cbegeman, this should be `nEdgesOwned`. I'm testing now. The partition tests will tell us if this is correct.

hyungyukang: @cbegeman and @mark-petersen , Thanks a lot. Please let me know once your test is done. And I will also test the code on larger meshes.

mark-petersen: I tested this loop with `nEdgesOwned` and it changes the solution, as expected from results in https://github.com/E3SM-Project/E3SM/pull/6041#pullrequestreview-1710892492. But for AB2, if I change both loops (lines 1542 and 1570) from `nEdgesOwned` to `nEdgesAll` the solution does not match for all tests running `split_explicit_ab2`. Since `nEdgesAll` is the more conservative approach, I am changing the loop extents to the larger option.

mark-petersen:
Pushing the loop extent change with @hyungyukang's permission. See https://github.com/E3SM-Project/E3SM/pull/5989#discussion_r1380567559 above.

mark-petersen (approved):
Approving based on testing by @hyungyukang and me - both standalone and G cases, and at low resolution (EC30to60) and high resolution (RRS18to6). This is a great performance improvement for the ocean component, and all simulation results look correct and do not display any problems.

@hyungyukang thank you for your insight and effort here.

Merging is still contingent on the B case results (thanks @jonbob!), but I wanted to approve the code changes before I go out of town next week.

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split_ab2.F
cbegeman: @mark-petersen and @hyungyukang Thanks for following up on this! Glad to hear that the AB2 tests are going well.

xylar:
The only things I noticed in the analysis that seems outside the range of normal ensemble-member variability were the temperature and the OHC.  The OHC time series shows some extra warming in the top 700 m.  The temperature comparison with Argo shows some distinctive features in the tropics at around 150 to 200 m depth:
image

It's not clear to me whether the AB2 solution is better, worse or just different in this case.  I just wanted to point it out so we don't just ignore it.

vanroekel:
Two comments on the B-case test.  
1. It seems to involve both AB2 and the BTR weights, I'm guessing since the temperature signal is at a specific depth it is likely not due to the BTR weights
2. The atmosphere seems pretty different between this run and the baseline.  Looks like perhaps different tuning and maybe different chemistry?  Also looks like the L80 initial condition versus L72 so comparisons may be hard to draw

jonbob:
@vanroekel -- this run was made before the latest round of atm changes, and should be the same as the baseline except for including the AB2 and BTR PR's. I checked the atm namelist files and they're the same except for one BFB addition to EAM. So the difference in atm must be due to the ocean changes?

vanroekel:
Sorry @jonbob I diffed the wrong atm_in file (I was looking at the 1106 directory).  Sorry for the confusion.

jonbob:
Testing of this PR in conjunction with [PR #6035](https://github.com/E3SM-Project/E3SM/pull/6035) is being undertaken by the coupled model group. The run with these two PRs is [20231106.v3b01-AB2.piControl.chrysalis](https://acme-climate.atlassian.net/wiki/spaces/CM/pages/3990224906/20231106.v3b01-AB2.piControl.chrysalis) and the corresponding case without AB2 for an apples-to-apples comparison is [20231105.v3b01.piControl.chrysalis](https://acme-climate.atlassian.net/wiki/spaces/CM/pages/3984261154/20231105.v3b01.piControl.chrysalis). Both runs have gone more than 200 years

mark-petersen:
Thanks @jonbob for the B-case comparisons. I spent an hour carefully comparing the two simulations. Here are my conclusions. In all screenshots we have MPAS-Analysis from years 151-200 with:
- control: left or top [20231105.v3b01.piControl.chrysalis](https://acme-climate.atlassian.net/wiki/spaces/CM/pages/3984261154/20231105.v3b01.piControl.chrysalis)
- AB2: right or bottom [20231106.v3b01-AB2.piControl.chrysalis](https://acme-climate.atlassian.net/wiki/spaces/CM/pages/3990224906/20231106.v3b01-AB2.piControl.chrysalis)

1. Dynamic variables are identical by eye, within variability: transports, MHT, SST, EKE.
<details>
<summary> plots: </summary> <p>

image
image
image
image

</details>

2. AMOC is low in both, but slightly higher in AB2 (~9 Sv) versus control (~8 Sv). This difference might lie within the variability.
<details>
<summary> plots: </summary> <p>
Right plot is squashed for axis comparisons:

image
image

</details>

3. Sea ice fields look identical by eye.
<details>
<summary> plots: </summary> <p>

image
image
image
image
image
image


</details>

4. **THE ONLY VISIBLE DIFFERENCE** that I found is in the heat content trends. AB2 OHC drift is smaller, which is better for an 1850 control run.
<details>
<summary> plots: </summary> <p>
Right plot is squashed for axis comparisons:

image
image

</details>

5. Despite the difference in OHC trends, the global comparisons at these depths still look identical by eye, so there are no obvious problems with a particular region or process.
<details>
<summary> plots: </summary> <p>

image
image
image

</details>

For example, Antarctic bottom temperature is very warm in both, but identical by eye between the control and AB2. Also for salinity.
<details>
<summary> plots: </summary> <p>

image
image
</details>

**In summary**, the only visible differences with AB2 are the temperature trends. The temperature drifts are smaller than the control, so are preferred. In addition, the spatial distribution of the temperatures look nearly identical by eye, so there are no problems that stand out.

mark-petersen (approved):
I am approving again based on the B-case simulations and results above. I feel comfortable with AB2 as the default baroclinic time stepping scheme for V3.

xylar (approved):
I focused on the differences in temperature (and heat content) in the shallower ocean.  I'm seeing small differences but they're much less than the biases compared with Argo:

AB2:
image

SE:
image

So I'm comfortable with the AB2 results.  The performance gain is remarkable!  Excellent work!

cbegeman (approved):
@hyungyukang and @mark-petersen I'm glad to hear the 200 year comparison turned out so well (similar to control). Great work!

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split_ab2.F
```diff
@@ -0,0 +1,3262 @@
...
+!
+!  ocn_time_integration_split_ab2
+!
+!> \brief MPAS ocean split explicit AB2 time integration scheme
+!> \author Hyun Kang (Oak Ridge National Laboratory)
+!> \date   October 2023
+!> \details
+!>  This module contains the routine for the split explicit time
+!>  integration scheme, where the second-order Adams Bashforth (AB2)
+!>  time stepping method is applied to the baroclinic system.
```
vanroekel, line 18: @hyungyukang can we modify this comment to clarify that the AB2 stepping is for momentum only?  The tracer is euler forward.

vanroekel (changes requested):
@hyungyukang this is excellent, great work!  I have one tiny request, when I looked through the code it appears AB2 is applied to momentum only and I believe you confirmed that.  Could we clarify that in the header of the ab2 file?

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split_ab2.F
hyungyukang: @vanroekel , Sure I will change it. Thanks a lot!

vanroekel (approved):
approving based on visual inspection and testing from @hyungyukang @mark-petersen and @jonbob

jonbob:
passes sanity testing with expected DIFFs -- merged to next

hyungyukang:
@mark-petersen , @sarats , @xylar , @cbegeman , @vanroekel , and @jonbob , thank you so much for your review, testing, suggestions, comments, and approval of this PR!

jonbob:
merged to master -- expected DIFFs will be blessed with PR #6035

amametjanov:
It appears that after the merge nightly OpenACC test
```
./cime/CIME/scripts/create_test SMS_Ld1.T62_oEC60to30v3.CMPASO-NYF
```
stopped working with either `--machine ascent --compiler pgigpu` or `--machine pm-gpu --compiler nvidiagpu`.
I took a stab at fixing initial build errors, but there's still quite a few variables that need to be copied CPU-to-GPU: e.g.
```
FATAL ERROR: data in PRESENT clause was not found on device 1: name=layerthickness
```

hyungyukang:
> It appears that after the merge nightly OpenACC test
> 
> ```
> ./cime/CIME/scripts/create_test SMS_Ld1.T62_oEC60to30v3.CMPASO-NYF
> ```
> 
> stopped working with either `--machine ascent --compiler pgigpu` or `--machine pm-gpu --compiler nvidiagpu`. I took a stab at fixing initial build errors, but there's still quite a few variables that need to be copied CPU-to-GPU: e.g.
> 
> ```
> FATAL ERROR: data in PRESENT clause was not found on device 1: name=layerthickness
> ```

@amametjanov , Thanks for reporting it. I was able to run the AB2 code on both pm-gpu and Frontier-gpu, but I had to do some modifications for a macro `-DMPAS_OPENACC` which will not be activated just for using `pm-gpu`. I've referenced @grnydawn 's fix for PM-GPU at here: https://github.com/E3SM-Project/E3SM/tree/ykim/mpas/gcase, but I believe this modification is not merged yet. 

I also found that I made a mistake in `mpas_ocn_diagnostics_variables.F` when I commit this code for GPU runs. I will open a new PR with some modifications to run the code on GPUs properly.

PR #5990, opened 2023-10-12
Clean up remaining cmake deps and unused items

jgfouca:
Changes:
* Remove hdf5 stuff from the macros. FindNETCDF should take care of this stuff
* Remove ESMF stuff that we don't use
* Clean up some MPI stuff that was unused or had no effect on the build
* gptl is now being handled properly in the new style

With this PR, I think all of our main dependencies are in a good enough state as far as cmake is concerned.


bartgol (approved):
Nice cleanup!

PR #5993, opened 2023-10-13
Fix MKL linking on pm-cpu_intel

jgfouca:
Need to tell cmake to use Intel vendor for blas.

[BFB]

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -273,6 +273,9 @@
       <env name="LAPACK_ROOT">$SHELL{if [ -z "$LAPACK_ROOT" ]; then echo /opt/nvidia/hpc_sdk/Linux_x86_64/22.7/compilers; else echo "$LAPACK_ROOT"; fi}</env>
       <env name="BLA_VENDOR">NVHPC</env>
     </environment_variables>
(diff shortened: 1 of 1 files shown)


jgfouca:
I'm not 100% sure that the BLA_VENDOR is what we want, but it will get pm-cpu working again and I will follow up with Kitware.

ndkeen:
I'm communicating with Jim on slack as well as trying to ask others at NERSC. I'd be OK with this change for now to get tests passing (assuming we do get same results/performance), and we can revisit later. ?

dqwu (approved):
Manually tested this fix on pm-cpu with intel compiler.

ndkeen:
I see you already merged this to master.

I did some quick benchmarks on 8 node of pm-cpu, with and without openmp.
The performance looks the same to me.  I don't have a good feel for the percent time spent in MKL here.

```
ne30pg2_EC30to60E2r2.F2010, 5 days

Oct4th master
8 nodes, 128x1, 8.61 SYPD
8 nodes,  64x2, 8.81 SYPD

jgfouca_fix_pm_intel_blas
8 nodes, 128x1, 8.77 SYPD
8 nodes,  64x2, 8.79 SYPD
```

PR #5995, opened 2023-10-13
Fix blas/lapack on ascent/summit

jgfouca:
Tell cmake to use Generic so it doesn't think ESSL is sufficient.

[BFB]


amametjanov (commented):
I checked one case `SMS_Ln5.ne4pg2_oQU480.F2010.ascent_gnu` and it now builds and runs okay with this change.

PR #5996, opened 2023-10-13
Add L80 support for v3

whannah1:
Modify default configurations to use a new 80 layer vertical grid, mainly for enhancing the QBO in E3SMv3. Since this PR touches many compsets and tests I took the opportunity to clean up how the configurations are specified and delete some unused compsets. 

New atmosphere initial condition files (i.e. ncdata) have been added for ne4, n30, and ne120 grids, but only for cases with realistic topography. The process of creating new initial conditions for idealized compsets like aquaplanet and RCE was complicated by the new aerosol/chemistry options, so these configurations will continue to use the v2 settings, including L72, and will be updated later.

[non-BFB]
[NML]


whannah1:
e3sm_atm_developer passes on pm-cpu

singhbalwinder (approved):
Looks great! Thanks for cleaning up the old codes!

whannah1:
`e3sm_atm_developer` passes on chrysalis

whannah1:
Update on new initial condition files - the new default files for ne4 and ne30 were remapped from a recent ne30 file used for v3 and put through a 1-year adjustment with a F2010 simulation. The ne120 initial condition has been a bit problematic, which is consistent with my previous experience remapping IC files from low to high resolution. I'm currently trying to "bootstrap" the new initial condition into a stable state by running with a 1-min physics time step and progressively increasing the time step until I can run with the default 15-min time step. This is being done on chrysalis.

tangq:
@whannah1 , some of the namelist default settings (e.g., gaschmbudget_2D_??) will need to be changed with L80. The changes only affect diagnostics and will be BFB. What's your preference? Add to this PR or issue a new PR?

whannah1:
@tangq I would prefer to put that in another PR - but it's unclear to me how these should be updated. Am I correct to assume that these levels for the diagnostics are chosen arbitrarily? Would it be better to switch these diagnostics to interpolate to specific pressure levels?

tangq:
@whannah1 , @wlin7 has a wonderful suggestion that we implement it as
```
<gaschmbudget_2D_L4_e nlev="72">         72       </gaschmbudget_2D_L4_e>
<gaschmbudget_2D_L4_e nlev="80">         80       </gaschmbudget_2D_L4_e>
```
This will work for both vertical settings and if we need more in the future we can just add to it.

These diagnostics are mainly for debugging purpose for chemistry and hence need to be accurate to the machine precision. That's why we really need to be on model levels, not pressure levels. The levels are chosen based on chemistry domains and can be modified by namelist when needed. It's possible to automate it, but since mostly only chemistry developers will use these diagnostics, it is probably not worth the extra effort. Wuyin's method is good enough.

I will issue a new PR following Wuyin's suggestion.

whannah1:
@tangq ok that sounds like a good plan.

wlin7:
Merged to next.

wlin7:
Merged to master.
Fixes to come for L80 ICs for wcprodssp, NARRM, atm_nbfb tests. Specifications for a couple condidiag tests also need to be updated.

amametjanov:
Just noting a throughput DIFF in nightly benchmarks: approx. 11%.
Runs in `e3sm_prod_bench`:
```
t1. PFS.ne30pg2_r05_oECv3.F2010.chrysalis_intel.bench-noio
t2. PFS.ne30pg2_r05_oECv3.F20TR.chrysalis_intel.bench-noio
t3. PFS.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel.bench-noio
t4. PFS.ne30pg2_r05_EC30to60E2r2.WCYCL1850.chrysalis_intel.bench-noio
```
DIFFs in SYPD: prior Oct-26 baseline, this PR and overall Nov-03 throughput.
```
	t1		t2		t3		t4	
	sypd	%-diff	sypd	%-diff	sypd	%-diff	sypd	%-diff
Oct-26	2.634	     	2.605		1.990		1.930	
5996	2.344	10.87%	2.324	10.92%	1.786	9.80%	1.742	8.80%
Nov-03	2.350	10.78%	2.298	11.79%	1.770	11.06%	1.710	11.40%
```

PR #5997, opened 2023-10-13
PIO: Remove dependence on CIME/Tools/Makefile

jgfouca:
We need to remove all dependencies on CIME/Tools/Makefile so we can change our macros to use cmake-style names for things. PIO/Scorpio was the first thing I picked.

The situation with spio was strange. It was using CIME/Tools/Makefile to call CMake. It translated the Makefile-style settings to Cmake using a lot of extra code in the Makefile to set up CMAKE_OPTS.

We change the python wrapper build script, buildlib.spio, to instead just set up the cmake options in python, allowing us to skip the CIME/Tools/Makefile.

On top of removing the dependency, I think this makes the overall spio build a lot easier to understand, although it was a bit tedious pulling things out of the macros.

Once the macros are in proper cmake-stype, I think we could just pass them directly to scorpio's CMake build via -C and things will be even simpler.

[BFB]


jayeshkrishna:
@jgfouca Thanks for taking up this task, this was on my plate of things I had planned to do. I am planning to further improve the spio build script to install SCORPIO correctly (instead of manual install - copying over all the required artifacts).

share/build/buildlib.spio
```diff
@@ -86,80 +92,75 @@ def buildlib(bldroot, installpath, case):
...
+        cmake_opts += f"-DPIO_USE_MPISERIAL=TRUE -DMPISERIAL_PATH={installpath} "
+    cmake_opts += "-DCMAKE_VERBOSE_MAKEFILE:BOOL=ON "
+    cmake_opts += f"-DGPTL_PATH:STRING={installpath} "
+    cmake_opts += "-DPIO_ENABLE_TESTS:BOOL=OFF -DPIO_USE_MALLOC:BOOL=ON "
+    cmake_opts += f"-DUSER_CMAKE_MODULE_PATH:LIST={cimeroot}/CIME/non_py/src/CMake "
+
+    # Hacky way of getting stuff from Macros. Should scorpio just load the E3SM
+    # macros once they are using cmake names for things?
+    make_args = get_standard_makefile_args(case, shared_lib=True)
+    macro_dump = run_cmd(f"make -f Macros.make COMP_NAME=spio {make_args} -p")[1]
```
bartgol, line 136: Why can't we just use `Macros.cmake` directly?

cime_config/machines/cmake_macros/crayclang-scream_crusher-scream-gpu.cmake
```diff
@@ -8,9 +8,9 @@ endif()
 set(PIO_FILESYSTEM_HINTS "gpfs")
 string(APPEND CXX_LIBS " -lstdc++")
 
-SET(CMAKE_C_COMPILER "cc" CACHE STRING "")
-SET(CMAKE_Fortran_COMPILER "ftn" CACHE STRING "")
-SET(CMAKE_CXX_COMPILER "hipcc" CACHE STRING "")
+set(CMAKE_C_COMPILER "cc")
```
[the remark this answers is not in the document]
jgfouca: Currently, the macros are loaded for each component in order to give each of them a chance to customize flags etc, so none of the settings in cmake_macros should be CACHE with the current implementation.

share/build/buildlib.spio
jgfouca: The macros don't use Cmake names for things, so would have no effect on the spio build.

jgfouca:
@bartgol , i've been working on a FindPIO module that allows us to find the package and bring in dependencies. I consider it a temporary solution while @jayeshkrishna and others do a proper cmake update for scorpio which would yield a proper PIO-config.cmake file in the install area.

jayeshkrishna:
> > @jgfouca Thanks for taking up this task, this was on my plate of things I had planned to do. I am planning to further improve the spio build script to install SCORPIO correctly (instead of manual install - copying over all the required artifacts).
> 
> @jayeshkrishna I have a branch (started long ago, but I just rebased it on develop) where I started to work on shipping scorpio as a cmake package, so that one can do
> 
> ```
> find_package (SCORPIO)
> ```
> 
> and get all its dependency (hdf5, (p)netcdf, adios), hence relieving the downstream projects from the burden of finding all the tpls of scorpio. I did not complete it, so it needs more work, but perhaps it can be of help? If you guys decide to pull the trigger and upgrade scorpio to modern cmake, I can also help with the cmake machinery.
> 
> The branch is [here](https://github.com/bartgol/scorpio/tree/bartgol/export-targets), in case you are interested.

Thanks, I will take a look at the branch. Meanwhile, I worked on a package config for SCORPIO last week. I will take a look at the branch and include any interesting feature that you have in the branch.

jgfouca:
@jayeshkrishna , what is confusing me is why `HDF5_PATH` needs to be set to find hdf5 even when the environment has `HDF5_ROOT` in it.

jayeshkrishna:
> @jayeshkrishna , what is confusing me is why `HDF5_PATH` needs to be set to find hdf5 even when the environment has `HDF5_ROOT` in it.

Must be the cmake module for HDF5 in SCORPIO. We can test and fix it in SCORPIO.

dqwu:
> @jayeshkrishna , I'm looking at the FindHDF5 in scorpio. It looks like there are some supported env vars, I just need to figure out which one works. My next question is how/where WITH_HDF5 gets set?

I am also not clear about WITH_HDF5 (to turn on the new HDF5 IO type support in SCORPIO). For WITH_ADIOS2, it is turned on when ADIOS2_ROOT is set in config_machines.xml.

Also, if a machine will be configured with serial NetCDF (no NetCDF4 parallel) and HDF5 (new IO type supported by SCORPIO), we can no longer find HDF5 libs via NetCDF4 dependencies.

bartgol:
@jayeshkrishna can you quickly remind us why scorpio needs to roll its own FindHDF5 module? I thought it was well supported by common CMake installations... Perhaps I could ask a similar question for netcdf/pnetcdf...

It would be great if we could rely on "built-in" cmake stuff, without having to roll (and maintain!) our own...

jgfouca:
@bartgol , that's a good question. CMake has provided a FindHDF5 module since at least 3.6.3: https://cmake.org/cmake/help/v3.6/module/FindHDF5.html

jgfouca:
@jayeshkrishna , with the current HEAD of this branch, I was able to get scorpio to build on alngce (with hdf5 and adios on). I had to make a minor code change to scorpio to remove a duplicate impl of `strdup` in `src/clib/pioc_support.c`.

dqwu:
> @jayeshkrishna , with the current HEAD of this branch, I was able to get scorpio to build on alngce (with hdf5 and adios on). I had to make a minor code change to scorpio to remove a duplicate impl of `strdup` in `src/clib/pioc_support.c`.

This is a known issue which has been fixed for scorpio release 1.5.0, see E3SM-Project/scorpio#511

jayeshkrishna:
> @jayeshkrishna can you quickly remind us why scorpio needs to roll its own FindHDF5 module? I thought it was well supported by common CMake installations... Perhaps I could ask a similar question for netcdf/pnetcdf...
> 
> It would be great if we could rely on "built-in" cmake stuff, without having to roll (and maintain!) our own...

These modules were added before CMake had support for these libs (and we might still need to retain some if the support is not available). We can disable these modules after testing on all the E3SM machines.

jayeshkrishna:
@dqwu : How can this PR be merged before we move to 1.5.0 (Can you disable ADIOS for ex to get rid of the strdup error?)?

dqwu:
> @dqwu : How can this PR be merged before we move to 1.5.0 (Can you disable ADIOS for ex to get rid of the strdup error?)?

As long as we do not set both ADIOS2_ROOT and HDF5_ROOT for a machine in config_machines.xml we will not have this issue for scorpio 1.4.2.

jgfouca:
@dqwu , anlgce does define both, but it's not a nightly test machine, so maybe it's ok.

jayeshkrishna:
@dqwu: Can we use "-DWITH_ADIOS2=OFF -DWITH_HDF5=OFF" as a workaround (until we upgrade SCORPIO)?

share/build/buildlib.spio
```diff
@@ -86,80 +92,95 @@ def buildlib(bldroot, installpath, case):
...
+        cmake_opts += f"-DNetCDF_PATH:PATH={os.environ['NETCDF_PATH']} "
+    if "NETCDF_C_PATH" in os.environ:
+        cmake_opts += f"-DNetCDF_C_PATH:PATH={os.environ['NETCDF_C_PATH']} "
+    if "NETCDF_FORTRAN_PATH" in os.environ:
+        cmake_opts += f"-DNetCDF_Fortran_PATH:PATH={os.environ['NETCDF_FORTRAN_PATH']} "
+
+    # scorpio needs a little help finding hdf5. The FindHDF5 in scopio uses the
+    # env var "HDF5" instead of "HDF5_ROOT". If HDF5_ROOT is not set, but we have
+    # h5dump in our path, we can use that info to set up HDF5.
+    which_h5dump = shutil.which('h5dump')
```
jgfouca, line 126: @jayeshkrishna , @dqwu, is this approach too aggressive? It means any environment that has h5dump in the PATH will result in WITH_HDF5=On for scorpio.

dqwu: I think we do not have to enable HDF5 support by default so far, at least before scorpio 1.5.0 becomes the default sub module.

dqwu:
> @dqwu: Can we use "-DWITH_ADIOS2=OFF -DWITH_HDF5=OFF" as a workaround (until we upgrade SCORPIO)?

I think only Perlmutter/Frontier/Summit/anlgce set ADIOS2_ROOT so far. Just make sure WITH_HDF5 is not turned on for these machines by default.

share/build/buildlib.spio
jgfouca: OK, I rolled this back.

dqwu:
> OK, I turned off `HDF5_ROOT` on summit.

We still need to verify that there are no build issues on summit with this change.

jgfouca:
@dqwu , summit appears to work fine. Builds with both ibm and gnu on this branch.

jgfouca:
@dqwu , @jayeshkrishna , @bartgol , I think this is ready for a final review.

dqwu:
> @dqwu , summit appears to work fine. Builds with both ibm and gnu on this branch.

How about anlgce? With static mpich/hdf5/netcdf/pnetcdf libs HDF5_PATH is set (now renamed to HDF5_ROOT). Maybe we can use an extra ENV variable to turn on WITH_HDF5? Note that HDF5_PATH/HDF5_ROOT might be used by NetCDF.

jgfouca:
@dqwu , anlgce would need the scorpio update that Jayesh mentioned .

dqwu:
@jgfouca ADIOS is only used by SCORPIO so far, so use ADIOS2_ROOT to turn on WITH_ADIOS2 seems OK.
HDF5 is used by both SCORPIO (directly HDF5 IO type support) and NetCDF (NetCDF4 parallel), so use HDF5_ROOT to turn on WITH_HDF5 is a problem. Maybe we can use HDF5_ROOT for NetCDF, and use another ENV variable to turn on WITH_HDF5 for SCORPIO.

share/build/buildlib.spio
```diff
@@ -86,80 +92,96 @@ def buildlib(bldroot, installpath, case):
...
-    gmake_opts = (
-        "{pio_dir}/Makefile -C {pio_dir} "
-        " {gmake_vars} {stdargs} -f {casetools}/Makefile".format(
-            pio_dir=pio_dir, gmake_vars=gmake_vars, casetools=casetools, stdargs=stdargs
-        )
-    )
+        cmake_opts += "-DWITH_PNETCDF:LOGICAL=FALSE -DPIO_USE_MPIIO:LOGICAL=FALSE "
+    if mpilib == "mpi-serial":
+        cmake_opts += f"-DPIO_USE_MPISERIAL=TRUE -DMPISERIAL_PATH={installpath} "
+    cmake_opts += "-DCMAKE_VERBOSE_MAKEFILE:BOOL=ON "
```
bartgol, line 148: Side question: why do we do this? I would assume that this should be added only if one issues `case.build --verbose`. I don't really care about long logs in the bldlog's...

jgfouca:
@dqwu, thanks, that is good to know. The weird thing is that all our machines seem to build fine without doing a find_package(HDF5..). It's only when SCORPIO has WITH_HDF5 on that it becomes an issue.

share/build/buildlib.spio
bartgol: But aren't we going to "abandon" support for `Macros.make` at some point? In that sense, wouldn't it be more long-term planning to parse `Macros.cmake`? I know var names are not the same, I wasn't suggesting to feed it straight to scorpio, but just to parse that instead of the makefile one.

bartgol (approved):

dqwu:
> @dqwu, thanks, that is good to know. The weird thing is that all our machines seem to build fine without doing a find_package(HDF5..). It's only when SCORPIO has WITH_HDF5 on that it becomes an issue.

anlgce uses static mpich/hdf5/netcdf/pnetcdf libs. In the past, Cori also built with static libs. It seems that only static libs require HDF5_PATH or find_package(HDF5..). Most E3SM machines use shared libs.

share/build/buildlib.spio
jgfouca: The buildlog for spio wouldn't have much useful info without verbose on.

dqwu:
@jgfouca @jayeshkrishna 
Maybe we can add a new ENV variable like WITH_HDF5_SCORPIO (TRUE or FALSE). Turn on WITH_HDF5 CMake option for SCORPIO only when this ENV variable is set and the value is TRUE in config_machines.xml. Feel free to use a different name.

jgfouca:
@bartgol , the easiest way to pull the macros out, while also handling conditionals etc, is via the Makefile -p approach. When we make the change to using cmake names for things, we will likely keep the Macros.make file around for this purpose.

jgfouca:
@dqwu , that sounds good to me. Let me add that. Are there any current machines for which we want WITH_HDF5_SCORPIO on?

share/build/buildlib.spio
bartgol: I guess this is just my personal preference. I really don't care about the bldlog details in general (and I wonder how many ppl really ever look at them _for successful builds_). I sometimes run `make VERBOSE=ON` (which has the same effect as having set `CMAKE_VERBOSE_MAKEFILE=ON` at config time), but only if I need to debug. That's why I figured "why doesn't CIME simply add VERBOSE=ON when case.build is invoked with `-v`"...

jgfouca: That's a reasonable preference. I think the issue is that users have been trained for many years now to expect $package.bldlog to have the full verbose build info.

dqwu:
> @dqwu , that sounds good to me. Let me add that. Are there any current machines for which we want WITH_HDF5_SCORPIO on?

I think we do not need HDF5 IO type by default on any machines before scorpio 1.5.0 becomes the default sub module. I can verify WITH_HDF5_SCORPIO works on anlgce (uses scorpio 1.5.0, or remove ADIOS2_ROOT for testing) in a temp feature branch, though.

dqwu:
@jgfouca 
With the new ENV variable WITH_HDF5_SCORPIO, maybe you can restore HDF5_ROOT for summit.

jgfouca:
@dqwu, good call. I reverted the summit HDF5 change.

jgfouca:
OK, tested mappy, summit, anlgce (with and without WITH_HDF5_SCORPIO) and it all worked. I'm going to merge to next unless someone objects.

dqwu:
> OK, tested mappy, summit, anlgce (with and without WITH_HDF5_SCORPIO) and it all worked. I'm going to merge to next unless someone objects.

Do we also need change NETCDF_PATH/ PNETCDF_PATH to NETCDF_ROOT/PNETCDF_ROOT?

jgfouca:
@dqwu, maybe someday. It's not necessary at the moment.

dqwu:
@jgfouca 
For the following code, could you please use "-DWITH_HDF5:BOOL=ON " to be consistent with WITH_ADIOS2?
```
    if "ADIOS2_ROOT" in os.environ:
        cmake_opts += "-DWITH_ADIOS2:BOOL=ON "
    ...
    if "WITH_HDF5_SCORPIO" in os.environ:
        cmake_opts += "-DWITH_HDF5=On "
```

Without this change, CMakeCache.txt has:
```
//Require the use of ADIOS 2.x
WITH_ADIOS2:BOOL=ON
...
//Require the use of HDF5
WITH_HDF5:BOOL=On
```

I will have a PR in the future to add hdf5 to valid IO type names. The searching of hdf5_string in CMakeCache.txt will fail (case sensitive matching)
```
    adios_string = "WITH_ADIOS2:BOOL=ON"
    hdf5_string = "WITH_HDF5:BOOL=ON"
```

jgfouca:
@dqwu , good idea, I adjusted the cmake flags.

PR #5958, opened 2023-09-26
Initial icepack integration

eclare108213:
Installs Icepack in E3SM.

This PR includes MPAS-seaice changes necessary to run Icepack in equivalent physics configurations as the original column package code. The entire Icepack repository is installed as a submodule linked from the main branch of E3SM’s Icepack fork.

Some minor changes were made to the original column package code for consistency with Icepack during testing. Icepack replaces colpkg as the default column physics package in this PR. 

Fully coupled E3SM v3alpha trigrid with Icepack physics is statistically compatible with the 100-year control at the 99% confidence interval.

Icepack is thoroughly and promptly documented in its git repository (the source code is in icepack/doc/).

[NML]
[CC]

PR #4324, opened 2021-05-28
Fix frazil mass leak

proteanplanet:
This corrects a mistake made in https://github.com/E3SM-Project/E3SM/pull/4295, whereby the adjustment to meltwater and salt were not weighted by the inverse of concentration to counter their subsequent weighting by sea ice concentration after export to the coupler.  The code now conserves to 10^-16 kg/m^2/s in mass and 10^10 in W/m^2 in heat.  The code has passed a ten-year sanity check. Output is available on Chrysalis /lcrc/group/e3sm/ac.afroberts/E3SM_simulations/20210527.v2rc2c.Fix2.ne30pg2_EC30to60E2r2.chrysalis from this test.  Graphics are provided in subsequent comments in this pull request to demonstrate model sanity and conservation.   Basic ten-year sanity check with MPAS analysis available at https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.afroberts/E3SM/v2/frazil/20210527.v2rc2c.Fix2.ne30pg2_EC30to60E2r2.chrysalis/mpas_analysis/ts_0001-0009_climo_0001-0009

[non-BFB]

Tagged: @stephenprice @darincomeau

PR #5999, opened 2023-10-16
Fixes PMSL coupling between ocean and atmosphere

proteanplanet:
This fixes a major bug in E3SM by replacing an incorrect coupling of PBOT (pressure in the bottom atmospheric layer), with PSLV (atmosphere model sea level pressure).

The consequence of this PR are widespread:  It fixes intermittent crashes that have been experienced in the Mediterranean Sea, decreases ocean heat uptake in shallow water, changes widespread pressure fields experienced in all regions of the ocean, and fixes major sea ice biases in both the Arctic and Southern Ocean.  

This pull request only affects B-cases.  In G-cases, PBOT and PSLV are currently identically set to pressure at mean sea level, thus the reason why this bug has previously gone undetected.

### DIFF
--- a/components/mpas-ocean/driver/mpaso_cpl_indices.F
+++ b/components/mpas-ocean/driver/mpaso_cpl_indices.F
@@ -65,7 +65,7 @@ module mpaso_cpl_indices
   integer :: index_x2o_Si_ifrac        ! fractional ice wrt ocean
   integer :: index_x2o_Si_bpress       ! ice basal pressure
   integer :: index_x2o_So_duu10n       ! 10m wind speed squared           (m^2/s^2)
-  integer :: index_x2o_Sa_pbot         ! atm bottom pressure              (Pa)
+  integer :: index_x2o_Sa_pslv         ! atmospheri sea level pressure    (Pa)
   integer :: index_x2o_Sa_co2prog      ! bottom atm level prognostic CO2
   integer :: index_x2o_Sa_co2diag      ! bottom atm level diagnostic CO2
   integer :: index_x2o_Foxx_taux       ! zonal wind stress (taux)         (W/m2   )
@@ -221,7 +221,7 @@ contains
 
     index_x2o_Si_ifrac      = mct_avect_indexra(x2o,'Si_ifrac')
     index_x2o_Si_bpress     = mct_avect_indexra(x2o,'Si_bpress')
-    index_x2o_Sa_pbot       = mct_avect_indexra(x2o,'Sa_pbot')
+    index_x2o_Sa_pslv       = mct_avect_indexra(x2o,'Sa_pslv')
     index_x2o_So_duu10n     = mct_avect_indexra(x2o,'So_duu10n')
     index_x2o_Foxx_tauy     = mct_avect_indexra(x2o,'Foxx_tauy')
     index_x2o_Foxx_taux     = mct_avect_indexra(x2o,'Foxx_taux')
--- a/components/mpas-ocean/driver/ocn_comp_mct.F
+++ b/components/mpas-ocean/driver/ocn_comp_mct.F
@@ -1564,7 +1564,7 @@ contains
 !    The following fields are sometimes received from the coupler,
 !      depending on model options:
 !
-!    o  pbot   -- bottom atm pressure                      (Pa)
+!    o  pslv   -- sea level pressure                      (Pa)
 !    o  duu10n -- 10m wind speed squared                   (m^2/s^2)
 !    o  co2prog-- bottom atm level prognostic co2
 !    o  co2diag-- bottom atm level diagnostic co2
@@ -2027,7 +2027,7 @@ contains
            rainFlux(i) = x2o_o % rAttr(index_x2o_Faxa_rain, n)
         end if
         if ( atmosphericPressureField % isActive ) then
-           atmosphericPressure(i) = x2o_o % rAttr(index_x2o_Sa_pbot,   n)
+           atmosphericPressure(i) = x2o_o % rAttr(index_x2o_Sa_pslv,   n)
         end if
         if ( seaIcePressureField % isActive ) then
            ! Set seaIcePressure to be limited to 5m of pressure


zhangshixuan1987:
@proteanplanet @wlin7 : Hi Andrew and Wuyin, I just put some suspicious about the potential causes for inconsistencies between PBOT (bottom model level pressure) and PSL:  
1. The PSL is calculated in tphysbc() in physpkg.F90 (around lines 3025-3027 in my code) before the radiation parameterization. This means that the model state calculated before radiation parameterization is used to derive PSL
2. If I understand it correctly, the atmosphere state used by the ocean model is from  "cam_export", which is prepared after the call of radiation parameterization. The radiation parameterization is known to change the temperature, which is applied through physics_update() module. In addition to the temperature, the geoptential height is also adjusted.  However, I did not see any changes in the surface pressure or model level pressure in physics_update() module. Therefore, I think that the PBOT should be still from the Dycore as I see that cam_out%pbot(i)  = state%pmid(i,over) in camsrfexch.F90.  

I pointed out the above observations to see if the issues could be potentially due to the use of variables calculated at different places.

rljacob:
Is this a climate-changing PR?

xylar:
> Is this a climate-changing PR?

I sure hope so!  It's a major problem to not have any effect on the climate!

proteanplanet:
> Is this a climate-changing PR?

Definitely.  I am working on all the documentation today.    I can switch this to draft if you would prefer.   But it needs to get in pretty quickly.

proteanplanet:
### History

@rljacob tracked down the reason why there was a change in coupling from PSLV to PBOT before V1 was released: https://github.com/E3SM-Project/E3SM/pull/791

That PR tried to address the problem of "tiger stripes" in the sea ice model, which was crashing with high frequency coupling switched on, an option previously used successfully in CICE by various coupled models.  We have since fixed the problem https://github.com/E3SM-Project/E3SM/pull/791 was attempting to address. It was caused by several combined problems, one of which was that sea ice velocity passed to the sea ice column physics was not being averaged across all vertices surrounding a grid cell, but instead only using a value from one vertex, so there was an incorrect feedback between ocean drag and sea ice velocity that literally spiraled out of control. In addition, a separate bug remained in the column physics that used a u-component in place of a v-component of ice velocity for high-frequency coupling.  Moreover, as part of trying to fix the tiger stripes, filtering was also applied to the sea surface height gradient passed from the ocean to sea ice. These multiple issues were fixed in V2: 

- Vertex averaging identified by a PEM test for V2 and addressed: https://github.com/E3SM-Project/E3SM/issues/3989
- Removal of sea surface height filtering between ocean and sea ice for V2: https://github.com/E3SM-Project/E3SM/pull/4268
- Fixed mass and energy leak between ice and ocean for V2:  https://github.com/E3SM-Project/E3SM/pull/4295, https://github.com/E3SM-Project/E3SM/pull/4324
- Corrected the u/v-component error and switch on high-frequency ice-ocean coupling for all configurations after thousands of integration years in NARRM and SORRM B-case configurations: https://github.com/E3SM-Project/E3SM/pull/5619

### Difference between PBOT and PSLV

Within EAM, PBOT is the mid-level pressure in the bottom layer of the atmosphere:
https://github.com/E3SM-Project/E3SM/blob/2236937c71ab0c4ef67c2574e58e01a0e46714d8/components/eam/src/control/camsrfexch.F90#L780-L784
This is consistent with all other variables needed to calculate turbulent fluxes over the ocean, https://github.com/E3SM-Project/E3SM/blob/master/share/util/shr_flux_mod.F90, and also in sea ice, both for MPAS-SI ColPkg and for Icepack coming in a new PR, nearly identical to the code linked here for ColPkg: https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-seaice/src/column/ice_atmo.F90

The definition of PBOT in EAM as the mid-layer pressure at the bottom of the atmosphere is consistent with other variables being passed for flux calculations, even though it's not actually being used for sea ice, and for the ocean the surface pressure PSLV is used to calculate the saturation vapor pressure at the sea surface.  PSLV is therefore the only correct coupling field for dynamical pressure on the sea surface.

### Impact of change from PBOT to PSLV coupling for the ocean

After integrating for 30 years in the v3alpha-trigrid configuration, switching PBOT for PSLV in the ocean coupling represented in this pull request, the following has occurred:

E3SM_sea_ice_lemnisc 0001-0030 v3alpha03-trigrid PSLV PI extent volume snow 3

There is a ~28% increase in sea ice and snow mass corresponding to ~6000 and ~1300 km^3 increase in ice and snow volume, respectively, in the Southern Ocean. The above figure summarizes daily sea ice phase average (solid) and interquartile range (gray). Dotted red lines indicate the null hypothesis for days of the year that MPAS-SI with Icepack is different from the standard version of MPAS-SI at the 95% compatibility interval. Traces start on January 1, passing the northward equinox (•), northern solstice (◦), southward equinox (•), and southern solstice (◦) to the end of December (→)

The local impacts can be seen in both the Arctic and Antarctic; sea ice no longer separates from the coasts, remains in the East Antarctic in summer, and piles up against the Canadian Archipelago where previously that did not occur. Sea ice no longer separates from the East Greenland coast:

E3SM_Sea_Ice_Thickness_arctic_PBOT_PSLV_0001_0030

E3SM_Sea_Ice_Thickness_antarctic_PBOT_PSLV_0001_0030

This simulation is continuing, and has reached year 39 as of the time of posting this comment. The ocean is venting heat and has not yet reached an equilibrium.

proteanplanet:
For the sake of creating a GitHub record of the difference between PBOT and PSLV, here are three projections from a V3 development configuration of E3SM to document the issue being solved by this PR.

Slide1

Slide2

Slide3

jonbob:
@proteanplanet -- I think this PR is missing a change in the cpl to add index_x2i_Sa_pslv? In my kitchen-sink branch I had to add:
````
diff --git a/driver-mct/shr/seq_flds_mod.F90 b/driver-mct/shr/seq_flds_mod.F90
index caa0c615a3..f45879b310 100644
--- a/driver-mct/shr/seq_flds_mod.F90
+++ b/driver-mct/shr/seq_flds_mod.F90
@@ -1455,6 +1455,7 @@ contains
 
     ! Sea level pressure (Pa)
     call seq_flds_add(a2x_states,"Sa_pslv")
+    call seq_flds_add(x2i_states,"Sa_pslv")
     call seq_flds_add(x2o_states,"Sa_pslv")
     longname = 'Sea level pressure'
     stdname  = 'air_pressure_at_sea_level'
````
to get it to work

proteanplanet:
@jonbob PSLV is never used by the sea ice model.  Instead of that change, I would prefer to completely remove the comments in the code.

milenaveneziani:
I agree with @proteanplanet: for me it was very confusing to have that comment, as well as the reference to the coupler field indexes in `mpassi_cpl_indices`. Since no atmospheric pressure is ever used by MPAS-si directly, shouldn't those me removed completely?

proteanplanet:
To address the comment by @zhangshixuan1987 above, I have reviewed the calculations in EAM at:
https://github.com/E3SM-Project/E3SM/blob/2236937c71ab0c4ef67c2574e58e01a0e46714d8/components/eam/src/control/camsrfexch.F90#L780-L784
and compared them with turbulent flux calculations over the ocean:
https://github.com/E3SM-Project/E3SM/blob/master/share/util/shr_flux_mod.F90
and also in sea ice, both for MPAS-SI ColPkg and for Icepack coming in a new PR, similar for the code linker here:
https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-seaice/src/column/ice_atmo.F90
The definition of PBOT in EAM as the mid-layer pressure at the bottom of the atmosphere is consistent with other variables being passed for flux calculations, even though it's not actually being used for sea ice, and for the ocean the surface pressure PSLV is used to calculate the saturation vapor pressure at the very surface of the ocean.
In short, I believe it is as simple as the wrong surface pressure was being used by the ocean model. Everything else appears consistent and correct in the use of surface pressure in the coupled model.

proteanplanet:
@golaz @xylar @vanroekel @alicebarthel @wlin7:  This PSLV pull request is now ready for your review.

xylar (approved):
Approving by inspection of the code, a comparison of the analysis with previous runs, and @proteanplanet's very convincing plots indicating that pressure seen by the ocean had unphysical discontinuities at continental boundaries with the old approach.

jonbob:
@proteanplanet -- I'm happy with removing pslv from the seaice code. What did you actually test with? I could not get the code to run without some modification, so I'm curious if it did work for you?

wlin7 (approved):
The codes changes are consistent with replacing pbot with pslv as atmospheric pressure for use by ocn.

wlin7:
Additional info on the effect of strong topography smoothing leading to non-zero surface elevation on coastal ocean grids (as seen by the atm component). This can be seen in the examples below that show non-zero surface heights beyond coastal lines of North America/Greenland, Hawaii Islands and around the Antarctic Peninsula. 
image

image

Non-zero surface heights lead to diagnosed sea level atmospheric pressure (PSLV)  being higher than the terrain-following 'surface pressure (PS). Note the original use of pbot represents a even large departure from sea level pressure.

image

image

vanroekel (approved):
Approving based on visual inspection of code and available analysis and @proteanplanet's extensive testing and convincing analysis

alicebarthel (approved):
Approved by visual inspection of the code, the coupler field plots, and the MPAS-Analysis comparison.

jonbob:
Passes sanity testing during merge with expected results, merged to next

jonbob:
merged to master and expected DIFFs from this PR as well as PR #5921 and PR #5958 blessed together, since they were merged at the same time. There are a few exceptions that need to be followed up on:
* one test on compy_prod that had not run (SMS_PS.northamericax4v1pg2_WC14to60E2r3.WCYCL1850.compy_intel.allactive-wcprodrrm_1850)
* one test on atm_nbfb that had not run (MVK_PS.ne4_oQU240.F2010.chrysalis_intel)
* one test on chrysalis_next that had a license error during build (SMS_Ln9.ne4_oQU240.F2010.chrysalis_intel.eam-outfrq9s)
* one test on chrysalis_next that had an unknown error (PET_Ln9_PS.ne30pg2_EC30to60E2r2.WCYCL1850.chrysalis_intel.allactive-mach-pet)
* two BGC tests that failed on chrysalis and pm-cpu that must be fixed

Issue #4960, opened 2022-05-18
ERROR: calendar dates must be increasing for a few cases using AMD compiler on pm-cpu


ndkeen:
With recent changes to master allowing use of AMD compiler `amdclang` on `pm-cpu`, I see a few tests fail in the same way.
For example, this test:
`ERS.f19_f19.I1850ELMCN.pm-cpu_amdclang`

Fails with:
` 0:  ERROR: (shr_stream_verifyTCoord) ERROR: calendar dates must be increasing`

Here is current list of tests failing with that error using AMD compiler:

```
ERS.f09_g16.I1850ELMCN.pm-cpu_amdclang.elm-bgcinterface
ERS.f09_g16.I1850GSWCNPRDCTCBC.pm-cpu_amdclang.elm-vstrd
ERS.f19_f19.I1850ELMCN.pm-cpu_amdclang
ERS.f19_g16.I1850CNECACNTBC.pm-cpu_amdclang.elm-eca
ERS.f19_g16.I1850CNRDCTCBC.pm-cpu_amdclang
ERS.f19_g16.I1850GSWCNPECACNTBC.pm-cpu_amdclang.elm-eca_f19_g16_I1850GSWCNPECACNTBC
ERS.f19_g16.I1850GSWCNPRDCTCBC.pm-cpu_amdclang.elm-ctc_f19_g16_I1850GSWCNPRDCTCBC
ERS.ne11_oQU240.I20TRELM.pm-cpu_amdclang
ERS.r05_r05.IELM.pm-cpu_amdclang.elm-V2_ELM_MOSART_features
SMS.r05_r05.I1850ELMCN.pm-cpu_amdclang.elm-qian_1948
SMS_Ld1.hcru_hcru.I1850CRUELMCN.pm-cpu_amdclang
SMS_Ly2_P1x1.1x1_smallvilleIA.IELMCNCROP.pm-cpu_amdclang.elm-force_netcdf_pio
SMS_Ly2_P1x1.1x1_smallvilleIA.IELMCNCROP.pm-cpu_amdclang.elm-lulcc_sville
SMS_Ly2_P1x1.1x1_smallvilleIA.IELMCNCROP.pm-cpu_amdclang.elm-per_crop
```

Also just noticed a similar error in a diff case:
`ERROR: (shr_stream_verifyTCoord) ERROR: elapsed seconds must be increasing`

happened with:
```
[...]

PR #6003, opened 2023-10-17
Clean up modules on all NERSC machines; update AMD on pm-cpu/alverez

ndkeen:
For pm-cpu/alvarez machines only, update to latest AMD compiler (4.0.0) -- Note NERSC removed the previous version.
And, for all NERSC machines, add some modules to remove before loading what we want to reduce chances of issues.

[bfb]


rljacob (approved):

ndkeen:
merged to next

ndkeen:
Fixes https://github.com/E3SM-Project/E3SM/issues/4960

PR #6014, opened 2023-10-24
Use CMake variables/names in our cmake macros

jgfouca:
This should be the last major PR in CMAKE V2 effort.

Here are the list of items from the macros remaining that are not cmake style:
* HAS_F2008_CONTIGUOUS: Use to set `${CPRE}USE_CONTIGUOUS=contiguous`. Added for flags for .F90 files. There must be a better way to do this?
* MPICC/MPICXX/MPIFC: The mpi compilers. I think we keep these since we cannot just use CMAKE_<LANG>_COMPILER because we sometimes need to distinguish between serial and non-serial compilers.
* SCC/SCXX/SFC: See above
* CXX_LIBS: Things to add to link line if USE_CXX is on. This is mostly for ` -lstdc++` and ` -cxxlib`. There is probably a better way to do this. Just add to CMAKE_EXE_LINKER_FLAGS?
* CMAKE_OPTS: A way to communicate arbitrary CMAKE settings to the spio cmake command. Needs a better name at least. This was also used for CISM but we (E3SM) don't use that anymore.
* FIXEDFLAGS: Added to FFLAGS for .F and .f files
* FREEFLAGS: Added to FFLAGS for .f90 and .F90 files. I think we can replace these with the Fortran_FORMAT property?
* PIO_FILESYSTEM_HINTS: Used to pass PIO_FILESYSTEM_HINTS to spio cmake. This is very similar to CMAKE_OPTS
* SLIBS: Convert CMAKE_EXE_LINKER_FLAGS?
* KOKKOS_OPTIONS:  Keep this for now
* USE_HIP: Keep 
* CXX_LINKER: Replace with CMAKE_<LANG>_LINKER_PREFERENCE?
* PFUNIT_PATH: This is how some CIME stuff finds PFUNIT. Should be replaced by ENV{PFUNIT_ROOT} at some point but not urgent
* CONFIG_ARGS: Use to specify autoconf flags to the autoconf-based sharedlibs (MCT and MPI-SERIAL). I think a better name at least is needed
* USE_CUDA: keep
* CPRE: Part of that weird CONTIGUOUS stuff from above
* USE_SYCL: keep
* SYCL_FLAGS: It looks like SYCL is not yet a known Cmake <LANG>, so we can't use CMAKE_SYCL_FLAGS, so we can leave this as is for now.

Other todo:

- [x] Streamline cmake sharedlibs to use macros directly
- [x] Make sure other sharedlibs are grabbing the right names
- [x] Cleanup cmake core files now that names are right
- [x]  Remove use of CFLAGS, FFLAGS, CXXFLAGS
- [x]  Remove support for e3sm_remove_flags
- [x]  Handle fixed/free fortran stuff better
- [x] Revert the oneapi changes once renamer is done.
- [x] gptl needs to use CMake

Changes Summary:
*  Macro structure changes
  * Add post-process step to set final values for a few items
  * Make cmake conversion more robust by storing pre-macro values and comparing instead of just looking at new variables
* Change all macros to set variables that cmake knows about! This is a huge change with wide impacts
  * We no longer have to set flag properties on all files. This simplifies things but makes it hard to support `e3sm_remove_flags` so that feature has been removed. You should always be able to append flags to achieve the same effect as removing a flag.
  * Try to remove as much of the linker flag micromanagement as possible
  * We now default to cxx linker. Intel and PGI still have to use fortran unfortunately since our `main` function is in fortran
  * Remove as many non-cmake settings as possible from the macros
* Add script for converting macros from the old style to new style
* Upgrade the compare-flags tool to support parsing the e3sm.bldlog file. This is a tougher way to compare flags since you have to build the cases with `-j 1`, but it's more robust than looking at cmake internal files.
* Remove lots of unneeded stuff from main cmake files (build_model and common_setup)
* Remove unused old makefiles for COSP and MPAS
* Move gptl build to cmake

Testing:
* mappy_gnu: e3sm_developer, with full flag compare on a couple big cases
* anlgce: e3sm_developer builds
* pm-cpu_intel: e3sm_developer, with full flag compare on one big case
* chrysalis_intel: e3sm_developer, with full flag compare on one big case
* summit: build one big case with pgigpu, ibmgpu, pgi


bartgol:
A few thoughts:
- It appears `USE_SYCL`, `USE_HIP`, `USE_CUDA` are only used to set YAKL flags and YAKL arch. Perhaps we should unify things, doing something like `set (DEVICE_TYPE "CUDA")`, so that we can do
```
if (DEVICE_TYPE)
  set (YAKL_ARCH ${DEVICE_TYPE})
  set (YAKL_${DEVICE_TYPE}_FLAGS "${CPPDEFS} ${$DEVICE_TYPE}_FLAGS})
endif()
```
if nothing else, it would reduce the number of vars floating around.
- `CONFIG_ARGS`: do we still need this? I thought this PR got rid of the dependence on `cime/CIME/Tools/Makefile`, which means no more autoconf. Did I misunderstand it?
- `CXX_LINKER`:  since our main is in Fortran, does it make sense to use anything other than Fortran?
- `KOKKOS_OPTIONS`: long term, we should move to a CMake version of buildlib.kokkos, and make macros file set `Kokkos_XYZ` vars rather than provide a string with command line cmake options.
- `CXX_LIBS`: I think adding to `CMAKE_EXE_LINKER_FLAGS` may be ok
- `SLIBS`: same as CXX_LIBS, provided that these are just link line options (not cmake targets!).
- S-vs-MP compilers: why do we need both? Why can't we just set `CMAKE_<LANG>_COMPILER`? Even if a lib does not need/use MPI, we can always use the mpi wrappers to build it, no? It should have the same effect as building it via serial compilers...
- FREEFLAGS vs FIXEDFLAGS. Reading cmake docs, it seems that the Fortran_FORMAT property is set on a per-target basis. I'm not sure if some comps mix the two... Also, say you set the format, how do you specify different flags for different formats? Looking at our cmake macros files, they use a different set of values for the flags. Are those the same flags that setting the Fortran_FORMAT to FREE/FIXED would yield? I guess that'd be quite nice. If so, then I would be in favor of using the cmake property. We'd have to check if a) it can be used on a per-file basis or b) if it can't, see if there are components that use both formats.
- PIO_FILESYSTEM_HINTS: would be nice to just append this to `SPIO_CMAKE_OPTS`, but I am not sure that `CIME/Tools/Makefile` would digest it?
- `HAS_F2008_CONTIGUOUS`: why do we need to specify this in the macros file? I would assume a good old `try_compile` would do. I guess we may still need it if we build some sharedlib with raw Makefiles (this goes back to my first point: do we still use raw makefiles?)

jgfouca:
@bartgol , thanks for looking at this.

> It appears USE_SYCL, USE_HIP, USE_CUDA are only used to set YAKL flags and YAKL arch. Perhaps we should unify things, doing something like set (DEVICE_TYPE "CUDA")

I agree. This would be a nice upgrade.

> CONFIG_ARGS: do we still need this? I thought this PR got rid of the dependence on cime/CIME/Tools/Makefile, which means no more autoconf. Did I misunderstand it?

Unfortunately, I think we must keep this one. mct and mpi-serial have an autoconf-based config system. The recent changes make it so autoconf is executed directly from the python build wrappers (buildlib.mct for example) instead of the wrapper invoking Tools/Makefile which then invokes autoconf.

> CXX_LINKER: since our main is in Fortran, does it make sense to use anything other than Fortran?

I have no idea why this feature was added. It goes back many years. Is it possible that fortran linkers struggle to link CXX object files some times? We could remove this and see if all our main machines still work.

> KOKKOS_OPTIONS: long term, we should move to a CMake version of buildlib.kokkos, and make macros file set Kokkos_XYZ vars rather than provide a string with command line cmake options.

We are already doing a CMake build of Kokkos (we used to use the generate_makefile stuff, but I changed that years ago).  I agree we should move towards setting the actual Kokkos vars that Kokkos understands eventually. I don't know if I want to go there now.

> S-vs-MP compilers: why do we need both? Why can't we just set CMAKE_<LANG>_COMPILER? Even if a lib does not need/use MPI, we can always use the mpi wrappers to build it, no? It should have the same effect as building it via serial compilers...

Good question. Once again, this feature goes back many years so I don't know why it was introduced and if it is really needed.

> PIO_FILESYSTEM_HINTS: would be nice to just append this to SPIO_CMAKE_OPTS, but I am not sure that CIME/Tools/Makefile would digest it?

We don't use Tools/Makefile for anything anymore.

> HAS_F2008_CONTIGUOUS: why do we need to specify this in the macros file? I would assume a good old try_compile would do. I guess we may still need it if we build some sharedlib with raw Makefiles (this goes back to my first point: do we still use raw makefiles?)

I am sure there's a better way to do this. None of the sharedlibs will be using raw Makefiles by the time this build refactor is done.

bartgol:
Ok, I think we can take steps toward the final goal. I am fine merging this PR, but I would strongly encourage to follow up on

- need of MP-vs-S compilers
- Kokkos options inside cmake macros files. Perhaps we can use what EKAT has (or copy-paste it) and machines macro files can simply include them (this avoids changing tens of mach files if (ehm...when) kokkos changes their cmake syntax)
- Verify we can remove the CXX_LINKER option (I'm 97.43% positive we can).

`HAS_F2008_CONTIGUOUS`  and the `PIO_FILESYSTEM_HINTS` are less of a priority (would be nice to fix them, and reduce the clutter and maintenance burden at some point though).

Finally, it would be nice to contribute to MCT providing a CMake build system. Since the repo is owned by ANL (I see rob as a contact for the MCSclimate org on github), we can probably find someone that can explain what the configure needs are (if the output of configure itself is too cryptic) and craft a simple CMakeLists.txt (no need to handle installation via pkg, just a simple build is enough probably).

rljacob:
I'm the only MCT developer and I had a plan to learn Cmake by adding a build for MCT but never had time.

jgfouca:
@rljacob , we currently interface to the MCT build system through our python wrapper buildlib.mct. It would be slightly more convenient to do this if MCT were a CMake package, but it's really not a big problem.

cime_config/machines/cmake_macros/gnu_cades.cmake
```diff
@@ -1,3 +1,3 @@
-string(APPEND FFLAGS " -O -fno-range-check")
+string(APPEND CMAKE_Fortran_FLAGS " -O -fno-range-check")
```
bartgol, line 1: I understand this PR is just translating old syntax into new syntax, but...what's the point of `-O`?

cime_config/machines/cmake_macros/gnu_theta.cmake
```diff
@@ -1,13 +1,4 @@
...
-endif()
-if (NOT DEBUG)
-  string(APPEND FFLAGS " -O2")
-endif()
-if (DEBUG)
-  string(APPEND FFLAGS " -O0")
-endif()
-string(APPEND CXX_LIBS " -lstdc++")
+string(APPEND CMAKE_C_FLAGS_RELEASE " -O2")
+string(APPEND CMAKE_C_FLAGS_DEBUG " -O0")
```
bartgol, line 2: I comment here, but could be on any file. Since gnu follows the rule "last opt flag wins", and since `-O2` seems to be the most common choice for RELEASE, perhaps we can just set it in `gnu.cmake`, and add a line to `gnu_XYZ.cmake` only if they need different flags?

cime_config/machines/cmake_macros/gnugpu.cmake
```diff
@@ -1,34 +1,28 @@
...
-  string(APPEND CFLAGS   " -O0 -g -Wall -fbacktrace -fcheck=bounds -ffpe-trap=invalid,zero,overflow")
-  string(APPEND FFLAGS   " -O0 -g -Wall -fbacktrace -fcheck=bounds -ffpe-trap=zero,overflow")
-  string(APPEND CXXFLAGS " -O0 -g -Wall -fbacktrace")
-  string(APPEND CPPDEFS  " -DYAKL_DEBUG")
-endif()
-if (NOT DEBUG)
-  string(APPEND CFLAGS " -O")
-  string(APPEND CXXFLAGS " -O")
-  string(APPEND FFLAGS " -O")
+  string(APPEND CMAKE_C_FLAGS " -fopenmp")
```
bartgol, line 8: I think it's ok to move these omp flags to `gnu.cmake`...

cime_config/machines/cmake_macros/gnugpu.cmake
```diff
@@ -1,34 +1,28 @@
-string(APPEND CFLAGS " -mcmodel=medium")
-string(APPEND FFLAGS " -mcmodel=medium -fconvert=big-endian -ffree-line-length-none -ffixed-line-length-none")
+string(APPEND CMAKE_C_FLAGS " -mcmodel=medium")
+string(APPEND CMAKE_Fortran_FLAGS " -mcmodel=medium -fconvert=big-endian -ffree-line-length-none -ffixed-line-length-none")
 if (CMAKE_Fortran_COMPILER_VERSION VERSION_GREATER_EQUAL 10)
-   string(APPEND FFLAGS " -fallow-argument-mismatch")
+   string(APPEND CMAKE_Fortran_FLAGS " -fallow-argument-mismatch")
```
bartgol, line 4: This is already in `gnu.cmake`.

cime_config/machines/cmake_macros/gnugpu.cmake
```diff
@@ -1,34 +1,28 @@
-string(APPEND CFLAGS " -mcmodel=medium")
-string(APPEND FFLAGS " -mcmodel=medium -fconvert=big-endian -ffree-line-length-none -ffixed-line-length-none")
+string(APPEND CMAKE_C_FLAGS " -mcmodel=medium")
+string(APPEND CMAKE_Fortran_FLAGS " -mcmodel=medium -fconvert=big-endian -ffree-line-length-none -ffixed-line-length-none")
```
bartgol, line 2: This is already in `gnu.cmake`/`gnu9.cmake`

cime_config/machines/cmake_macros/intel_alvarez.cmake
```diff
@@ -15,24 +15,18 @@ set(SFC "ifx")
 # on the compiler ID or version, so for now, simply manually adjust the CXXFLAG setting for pm-cpu/intel
 # Try to manually remove -fp-model=source (and replace with -fp-model=precise) from CXXFLAGS
 #message(STATUS "ndk CXXFLAGS=${CXXFLAGS}")
-set(CXXFLAGS " ") # hardcode it here to blank, then try to do same things as in intel.cmake
+set(CMAKE_CXX_FLAGS " ") # hardcode it here to blank, then try to do same things as in intel.cmake
```
bartgol, line 18: Again, this PR is simply converting stuff, but what's the reason of not using what's in the generic `intel.cmake`?

cime_config/machines/cmake_macros/intel_pm-cpu.cmake
```diff
@@ -15,24 +15,18 @@ set(SFC "ifx")
 # on the compiler ID or version, so for now, simply manually adjust the CXXFLAG setting for pm-cpu/intel
 # Try to manually remove -fp-model=source (and replace with -fp-model=precise) from CXXFLAGS
 #message(STATUS "ndk CXXFLAGS=${CXXFLAGS}")
-set(CXXFLAGS " ") # hardcode it here to blank, then try to do same things as in intel.cmake
+set(CMAKE_CXX_FLAGS " ") # hardcode it here to blank, then try to do same things as in intel.cmake
```
bartgol, line 18: Here too. Do we know why?

components/cmake/build_model.cmake
```diff
@@ -185,12 +185,10 @@ function(build_model COMP_CLASS COMP_NAME)
     # Set flags based on file extension. File extensions may change if the globs used by
     # gather_sources changes.
     if (SOURCE_EXT STREQUAL ".c")
-      e3sm_add_flags("${SOURCE_FILE}" "${CFLAGS}")
+
```
bartgol, line 188: Maybe we can consolidate these checks and empty branches into something like
```
if (NOT SOURCE_EXT IN_LIST ".c;.cpp")
  # Cosp manages its own flags
  ...
```

bartgol (commented):
Another heroic effort by the valiant Ser Jim!

It never stops to amaze me how much less code we get once we do things the clean way...

I have a few comments. Some of them can be summarized into this: can we go through the `<compiler>_<machine>.cmake` files and remove settings that are already in the `<compiler>.cmake` file? Stuff like `-g` for debug flags, or `-fopenmp` when `compile_threaded=ON`, or `-O2` for RELEASE, or a bunch of other common flags.

Specialized comp-mach files can always append to the flags, and overwrite the opt level.

I know this PR was just meant to translate old syntax to new syntax, but perhaps, since you're at it, you could clean up the `<compiler>_<machine>.cmake` files?

cime_config/machines/cmake_macros/gnu_cades.cmake
jgfouca: No idea. I don't know if the machine `cades` is even something we still use.

jgfouca:
@bartgol , that's a good idea. In the past, I've used a script: `cime_config/machines/scripts/evaluator` to parse all the macros and look for opportunities to improve them. It offered these checks:

```
Available warnings (CODE => Decription)
  REP => Repitition: a redundant setting to same value as parent
  APP => Append: a repeated token in an append
  PRO => Promotion: all children of a parent do the same thing, so do it in the parent
  SET => Suspicious set: probably want append
  UNU => Unused: Using OS, COMPILER, or MACH with no matches in config_machines.xml
  MIS => Missing: Having a machine/comp in config_machines.xml with no specialization in Macros
  IFS => Ifs: Conditionals can be combined
```

The `PRO` warning seems to do what you are suggesting. I will look into reviving this script.

cime_config/machines/cmake_macros/gnu_cades.cmake
bartgol: We could email the person from the SUPPORTED_BY xml entry and ask. Not urgent, though.

components/cmake/build_model.cmake
```diff
@@ -259,7 +252,7 @@ function(build_model COMP_CLASS COMP_NAME)
     # Make sure we link blas/lapack
     target_link_libraries(${TARGET_NAME} BLAS::BLAS LAPACK::LAPACK)
 
-    set_target_properties(${TARGET_NAME} PROPERTIES LINKER_LANGUAGE ${LD})
+    set_target_properties(${TARGET_NAME} PROPERTIES LINKER_LANGUAGE CXX)
```
bartgol, line 255: Are you sure CXX is the right link language? I think the linker language used will impact the actual name of the program entry point (main), which may differ between CXX/Fortran. I think the proper way to do things is "use the language of the src file with the `main` program", and if you need cxx lib linked in, simply add it to the link line.

bartgol: I guess it's ok then. The common wisdom I've usually found online was to use the language of the main program, but I was never able to find an explanation of the rationale behind it.

cime_config/machines/cmake_macros/Macros.cmake
```diff
@@ -30,8 +36,17 @@ if (CONVERT_TO_MAKE)
...
-      message("CIME_SET_MAKEFILE_VAR ${VAR_AFTER} := ${${VAR_AFTER}}")
+    if (VAR_AFTER MATCHES "^E3SM_CMAKE_INTERNAL_")
+      # skip
+    else()
+      if (NOT VAR_AFTER IN_LIST E3SM_CMAKE_INTERNAL_VARS_BEFORE_BUILD_INTERNAL_IGNORE)
+        message("CIME_SET_MAKEFILE_VAR ${VAR_AFTER} := ${${VAR_AFTER}}")
+        list(APPEND E3SM_CMAKE_INTERNAL_VARS_BEFORE_BUILD_INTERNAL_IGNORE "${VAR_AFTER}")
+        set("E3SM_CMAKE_INTERNAL_${VAR_AFTER}" "${${VAR_AFTER}}")
+      elseif (NOT "${${VAR_AFTER}}" STREQUAL "${E3SM_CMAKE_INTERNAL_${VAR_AFTER}}")
+        message("CIME_SET_MAKEFILE_VAR ${VAR_AFTER} := ${${VAR_AFTER}}")
```
bartgol, line 47: Are these messages supposed to stay, or are they here just for debugging purposes while you are doing the port?

bartgol (approved):

cime_config/machines/cmake_macros/amdclang_chicoma-cpu.cmake
```diff
@@ -11,8 +11,8 @@ set(SCXX "clang++")
 set(SFC "flang")
 
 string(APPEND CMAKE_Fortran_FLAGS " -Mflushz ")
-string(APPEND FIXEDFLAGS " -Mfixed")
-string(APPEND FREEFLAGS " -Mfreeform")
+string(APPEND CMAKE_Fortran_FORMAT_FIXED_FLAG " -Mfixed")
```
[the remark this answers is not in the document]
jgfouca: Yes. I think CMake can probably handle this automatically but I'm not 100% on that so I kept it.

cime_config/machines/cmake_macros/Macros.cmake
jgfouca: They need to stay. This hacky stuff is how we translate processed cmake macros into Makefile macros. Nothing we have is using the Makefile macros directly anymore, but they are still useful for querying macro vars which we do in a few places when "crossing" build systems (cmake -> autoconf for example) in some of the sharelib builds.

bartgol (approved):

bartgol (approved):

components/cmake/build_model.cmake
jgfouca: I was wrong, it looks like intel requires fortran `main` to be linked with fortran compilers.

rljacob:
@jgfouca is this ready?  Please fix the title.

jgfouca:
@rljacob , yes, this has been ready for a while but I was waiting for the dashboard to get into better shape. This is going to be a disruptive PR.

jgfouca:
Merged to next.

rljacob:
Is it going to change answers?  Or just break some builds somewhere?

jgfouca:
@rljacob , it should be BFB. I tested it the best I could reasonably do so on my own, but it's a major change so I'm sure something will break somewhere.

Issue #5798, opened 2023-07-06
F2010_chemUCI-Linozv3-mam5 compset crash on pm-cpu


wagmanbe:
Case on perlmutter: `/global/cfs/cdirs/e3sm/emulate/E3SM_simulations/replicate.F2010.v3atm_on_master.pm-cpu/`
Branch: v3atm/eam/master_MAM5_wetaero_chemdyg
e3sm log error: `896: PIO: FATAL ERROR: Aborting... An error occured, Writing variables (number of variables = 180) to file (./replicate.F2010.v3atm_on_master.pm-cpu.elm.h0.0001-04.nc, ncid=186) using PIO_IOTYPE_PNETCDF iotype failed. Non blocking write for variable (SNOWDP, varid=157) failed (Number of subarray requests/regions=1, Size of data local to this process = 2700). NetCDF: Numeric conversion not representable (err=-60). Aborting since the error handler was set to PIO_INTERNAL_ERROR... (/global/u2/w/wagmanbe/E3SM/code/20230622/externals/scorpio/src/clib/pio_darray_int.c: 395)`

This case is my attempt to replicate a simulation on pm-cpu by @wlin7 found here : `/pscratch/sd/w/wlin/E3SMv3_dev/20230627.F2010.v3atm_on_master.pm-cpu/`. His simulation was meant to be a pm-cpu version of this case: [run.20230619b.v3alpha01.F2010.chrysalis.sh](https://github.com/E3SM-Project/SimulationScripts/blob/master/archive/CoupledGroup/v3_dev/run.20230619b.v3alpha01.F2010.chrysalis.sh), which runs on chrysalis. @wlin7's pm-cpu case ran 30 days before time-out, as intended. I find that if the run goes longer (3 months here, 15 months when I initialize with a different atmosphere), it crashes due to the ELM error above.

wagmanbe:
Error was caused by writing simulation output to /cfs instead of /pscratch.

wlin7:
Hi @ndkeen , do you have some idea for explaining such behavior, that a case created to run on PSCRATCH is ok, but would fail prematurely (after several months) if running on cfs? I remember when PSCRATCH was on maintenance, you used to create cases to run on cfs. 

Note that the issue @wagmanbe reported above was not a one-time thing. He replicated the problems many times, all running on cfs. Benj, were all those runs failed at about the same time? Were they BFB?

wagmanbe:
[...]

PR #6034, opened 2023-10-30
Move `h2osno` limiter to ensure positive snow depth.

quantheory:
In `LakeHydrologyMod`, when applying the effects of sublimation and dew to snow, the `snow_depth` variable is set using the `h2osno` variable before that variable is limited to positive values. This commit changes the order of those statements to ensure that `snow_depth` is not accidentally set to a negative value.

It is important to note that in the existing code, only extremely small-magnitude negative values are typically produced, because of another limiter applied earlier in the code. Specifically, `h2osno` variable is set using

```
h2osno(c) = h2osno(c) - qflx_sub_snow(p)*dtime
```

and `qflx_sub_snow` is set using

```
qflx_sub_snow(p) = min(qflx_evap_soi(p), h2osno(c)/dtime)
```

Thus, `h2osno` can only be temporarily negative due to limited floating point precision, i.e. the fact that we can get a slightly negative output from

```
h2osno(c) - (h2osno(c)/dtime)*dtime
```

For this reason, this commit only produces roundoff-level changes in answers under typical conditions.

It is also known that invalid values of `snow_depth` and `snowdp` are implicated in some rare model crashes (see #5798). The causes are unknown, but in at least one case involved exponentially growing negative snow depth. It is possible that this change will prevent such crashes; investigation is ongoing at the time of this commit.

[non-BFB]

### DIFF
--- a/components/elm/src/biogeophys/LakeHydrologyMod.F90
+++ b/components/elm/src/biogeophys/LakeHydrologyMod.F90
@@ -429,13 +429,12 @@ contains
             else
                h2osno(c) = h2osno(c) + (-qflx_sub_snow(p)+qflx_dew_snow(p))*dtime
             end if
+            h2osno(c) = max(h2osno(c), 0._r8)
             if (h2osno_temp > 0._r8) then
                snow_depth(c) = snow_depth(c) * h2osno(c) / h2osno_temp
             else
                snow_depth(c) = h2osno(c)/snow_bd !Assume a constant snow bulk density = 250.
             end if
-
-            h2osno(c) = max(h2osno(c), 0._r8)
          end if
 
          if (.not. use_extrasnowlayers) then


quantheory:
Notifying @crterai @wlin7. Using print/assert statements, I have confidence that this change is needed to avoid small negative snow depth values being produced by this code. However, I'm still testing to see if this code is also responsible for the large negative values crashing the coupled run with the gustiness changes.

quantheory:
In order to avoid dividing by a potentially nearly-zero number when setting snow depth, Xubin Zeng also suggests this change:

> replace "snow_depth(c) = snow_depth(c) * h2osno(c) / h2osno_temp" by
>                  "snow_depth(c) = h2osno(c) * min (snow_depth(c)/h2osno_temp, 1._r8/50._r8)"
>             This fix is physically based. the implication is that the snow density should be > 50 kg/m^3 or 1/density should be < 1/50 kg/m^3. 

I agree that this change should be made, but it should possibly be a separate PR, because the effect would be greater than roundoff level.

@mabrunke @xubinzeng

quantheory:
Noting that @billsacks [made this change in CLM](https://github.com/ESCOMP/CTSM/commit/3b01f70c6775b2f22b4cae6153b79c8bf0bed9d1) just over four years ago. (According to the commit message, this was done as a preventative measure rather than due to any known problem in the model.)

quantheory:
I have confirmed that not only is this section of code responsible for the negative values, but also can blow up an initially small negative value to a much larger negative value, which seems to be the ultimate cause of the crashes. Here's some output I generated from the time step and location (over Lake Erie) where the bad value occurs:

```
  28:  old snow_depth =  -4.336809803419874E-022
  28:  new snow_depth =  -3.746352330393036E+074
  28:  old h2osno =   9.175650230087428E-100
  28:  new h2osno =   7.926383719952902E-004
  28:  qflx_evap_soi =  -4.403546511084945E-007
  28:  qlfx_sub_snow =   0.000000000000000E+000
  28:  qlfx_dew_snow =   4.403546511084945E-007
  28:  do_capsnow =  F
  28:  ERROR: shr_assert_in_domain: snow_depth has invalid value 
  28:  -3.746352330393036E+074
```

This confirms that the the extreme negative value of $-3.7 \times 10^{74}$ is generated here, and that it arises due to division by the very small previous value of snowmass, $9.2 \times 10^{-100}$ (`h2osno_temp` is referred to as "old h2osno" in the above output).

@crterai @wlin7 @golaz This shows pretty definitively that this error is a preexisting error dating back to the original version of CLM that ELM was branched from. It is also not related to #5850. I could also use some guidance. The above output shows that the aforementioned fix from Xubin is needed as well, because even if we avoid negative `snow_depth` values, the same error can produce unrealistic positive values. That fix will cause changes that are larger than roundoff, but likely has negligible impact on climate. Should I add that fix to this PR or make a separate one?

quantheory:
At @crterai's suggestion, I did check the snow depth values in the code in a short run with and without this change. Without this change, negative `snow_depth` always occurs within a few time steps, whereas with this change, there were no negative values over a 10 day run. So this PR does work for preventing those bad values.

crterai (approved):
Thanks for creating this quick PR @quantheory. It looks good to me. And as @quantheory notes in the comments, it is a bug fix that's already been made in clm as well.

crterai:
Pinging @thorntonpe so that this PR is on your radar. Xubin Zeng also has a proposed fix that is in @quantheory's comment above, which we'd be interested to hear your take on.

bishtgautam (approved):

wlin7:
Merged to next.

rljacob:
@wlin7 it is to late to merge something to next for testing.

wlin7:
> @wlin7 it is to late to merge something to next for testing.

Too bad. I thought cutoff time is 11 PM CT. Waiting for final approval for the gustiness PR. If next testing not picking up the new merges, we can push the merges to master the following day.

rljacob:
The earliest suite starts at 10pm CT.

rljacob:
removed from next with a force push.

wlin7:
> removed from next with a force push.

How about leaving them on as today's merge to next? This and #5850. The Only other PR merged yesterday is #6054, tested by next today, and fixed SSP etc.

wlin7:
> > removed from next with a force push.
> 
> How about leaving them on as today's merge to next? This and #5850. The Only other PR merged yesterday is #6054, tested by next today, and fixed SSP etc.

wlin7:
Re-merged to next, and to master after the return of next testings.

PR #6036, opened 2023-10-31
Make "column_package" the default column_physics_type for BGC

jonbob:
With icepack now the default seaice column physics package, we need to make "column_package" the default when seaice BGC is on until it's completely supported in icepack.

[non-BFB] only for BGC configurations with active seaice (which are currently broken)

### DIFF
--- a/components/mpas-seaice/bld/build-namelist
+++ b/components/mpas-seaice/bld/build-namelist
@@ -629,7 +629,11 @@ add_default($nl, 'config_recover_tracer_means_check');
 # Namelist group: column_package #
 ##################################
 
-add_default($nl, 'config_column_physics_type');
+if ($ice_bgc eq 'ice_bgc') {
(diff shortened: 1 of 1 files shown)


jonbob:
verified that SMS_Ld2.ne30pg2_r05_EC30to60E2r2.BGCEXP_CNTL_CNPECACNT_1850.chrysalis_intel.elm-bgcexp now successfully runs

eclare108213 (approved):
We will need to remember to change this back later, when Icepack BGC is ready.

njeffery (approved):
Icepack bgc is not yet available. This fix ensures that mpas-seaice uses the original column physics and biogeochemistry in bgc simulations. Note, there is no option to use icepack physics and column bgc.   Approved by inspection.

jonbob:
passes e3sm_ice_developer on chrysalis, merged to next

jonbob:
merged to master and blessed BGC tests that had previously been failing and now show expected DIFFs

PR #6037, opened 2023-10-31
Add single missing variable to OMP private

jonbob:
Adds one new variable to an OMP private directive in a recently modified routine. It doesn't appear to cause PET failures but still should be added.

[non-BFB] potentially for threaded runs with active ocean


mark-petersen (approved):
Approved by visual inspection. Looks good, thanks.

mark-petersen:
In MPAS standalone tests with gnu debug on chrysalis, master currently fails 
```
01:13 FAIL ocean_global_ocean_QU240_WOA23_threads_test
01:10 FAIL ocean_global_ocean_QU240_WOA23_RK4_threads_test
```
and this PR fixes it so they both pass. Thanks @jonbob! 

For some master passes with gnu optimized, so the tests do not always catch this problem.

jonbob:
passes:

merged to next

jonbob:
merged to master

PR #6038, opened 2023-10-31
Apply QBO tuning adjustment for v3

whannah1:
Adjust hdepth_scaling_factor to reduce overly strong QBO amplitude for v3 w/ L80. 

[non-BFB]


whannah1:
closing so this can be incorporated in another 3 PR that updates namelist values.

Issue #6040, opened 2023-10-31
Bug in split-explicit barotropic velocities


cbegeman:
The split-explicit time integrator (and the AB2 split-explicit integrator) have two separate loops over edges, one that accumulates `normalBarotropicVelocityNew` and one that divides `normalBarotropicVelocityNew` by the number of barotropic subcycles. The first is over `nEdgesAll` and the second is over `nEdgesOwned`. When I added a new accumulator for the wetting-and-drying work that we’re doing, I noticed that the cells that are included in `nEdgesAll` but not `nEdgesOwned` have incorrect values for that accumulator. Thus, the fix that I propose, to make both loops over `nEdgesOwned`, is non-BFB in the standalone cases I've run so far.

https://github.com/E3SM-Project/E3SM/blob/abe445e110eb85ba6c88fc5a9c9d65bacc06fa96/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F#L1484-L1488

https://github.com/E3SM-Project/E3SM/blob/abe445e110eb85ba6c88fc5a9c9d65bacc06fa96/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F#L1512-L1520

PR #6041, opened 2023-10-31
Fix split-explicit barotropic velocity accumulator

cbegeman:
This PR includes a bugfix for the loop limits on the barotropic velocity (`normalBarotropicVelocityNew`) when its values are accumulated over barotropic subcycles. The loop should be over `nEdgesOwned` for consistency with the loop which divides the barotropic velocity by the number of subcycles.

Fixes https://github.com/E3SM-Project/E3SM/issues/6040

[BFB] for all current E3SM tests, changes to a non-default option

### DIFF
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F
@@ -1481,7 +1481,7 @@ module ocn_time_integration_split
 
                !$omp parallel
                !$omp do schedule(runtime)
-               do iEdge = 1, nEdgesAll
+               do iEdge = 1, nEdgesOwned
                   normalBarotropicVelocityNew(iEdge) = &
                   normalBarotropicVelocityNew(iEdge) + &
                   normalBarotropicVelocitySubcycleNew(iEdge)


cbegeman:
We need to run tests to evaluate if this fix is in fact non-climate changing. Labels will be updated when we know.

mark-petersen:
Actually, this fix doesn't make sense to me. If you are using the halo edges with wetting/drying for `normalBarotropicVelocityNew`, then the solution is to make the second loop at https://github.com/E3SM-Project/E3SM/issues/6040 go to `nEdgesAll`. If you are not using that variable on the halo, then the original extents to `nEdgesAll` on the first loop wastes computation, but shouldn't cause a problem.  

The fact that the incorrect value on the halo causes a problem won't be solved by the current fix, because the value on the halo is still incorrect.

mark-petersen:
@cbegeman, could you do a partition test with W/D on before and after this change? The partition test will tell you if there is any halo mismatch. If the compass partition tests are not right to produce W/D behavior, you can just do your own with 4 vs 8 or whatever with a W/D test. Thanks. 

I still suspect that changing the second loop to `nEdgesAll` is the correct fix, but we should always make the halos as tight as possible to save flops.

cbegeman:
@mark-petersen I just wanted to follow up that I've created another branch where I change the second loop to `nEdgesAll` and the solution is different from master and the same as this branch (for baroclinic_channel default, decomp, threads)

mark-petersen:
@cbegeman I think you changed the wrong loop. You meant to change this one, right? Not line 861.
https://github.com/E3SM-Project/E3SM/blob/abe445e110eb85ba6c88fc5a9c9d65bacc06fa96/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F#L1484-L1488

cbegeman:
@mark-petersen Thanks for catching that. Yes, I had the loop lines right in the issue. I will push the change now and retest.

cbegeman:
@mark-petersen For me, the comments above still hold with the updated branch. It would be good to have you confirm though.

mark-petersen (approved):
Running the nightly suite with gnu debug on chrysalis, I was able to confirm @cbegeman's results. With this PR, all decomp tests pass, and this is bfb identical between this PR and if both loops are changed to `nEdgesAll` in lines 1484 and 1512. The bfb comparison show that there is no harm done to the solution by tightening the halo computation, and there is a small benefit to computing fewer flops.

cbegeman:
@mark-petersen Thank you so much for your review and testing!

@jonbob This is now ready for your testing.

xylar (approved):
Approving by code inspection and based on @cbegeman and @mark-petersen's testing.

rljacob:
@vanroekel should this be in v3?

xylar:
I certainly think it should be.

mark-petersen:
Yes, this should be in V3.

vanroekel:
Sorry for not responding sooner.  @cbegeman and @mark-petersen could you summarize what the expected results here are?  The PR is labeled non-BFB, but @mark-petersen approval comment seem to suggest there is no difference in the loop limit?  I think it is something that needs to be included, but would appreciate a bit of clarity to alleviate my confusion.

Related, any expectation on the climate impact?  How far from nonBFB are the changes?

@cbegeman related question does AB2 not have this issue?

Sorry for all the questions.

But from what I understand from this discussion and issue this is a bug that should be fixed for v3.

cbegeman:
@vanroekel No problem. Here's an overview of the results of the standalone testing:

Loop 1 is the accumulator loop https://github.com/E3SM-Project/E3SM/blob/abe445e110eb85ba6c88fc5a9c9d65bacc06fa96/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F#L1484-L1488

Loop 2 is the division loop https://github.com/E3SM-Project/E3SM/blob/abe445e110eb85ba6c88fc5a9c9d65bacc06fa96/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F#L1512-L1520

A. Master: Loop 1 is `nEdgesAll`, loop 2 is `nEdgesOwned`
B. This branch: Loop 1 is `nEdgesOwned`, loop 2 is `nEdgesOwned`
C. Alternate: Loop 1 is `nEdgesAll`, loop 2 is `nEdgesAll`

A is non-BFB with B and C.
B and C are BFB.

Here we go with B, the least computations. AB2 goes with C, the more conservative approach https://github.com/E3SM-Project/E3SM/pull/5989/files#r1380567559.

vanroekel:
Very helpful, thanks @cbegeman!  I think I understand things now.  This is a definite bug with a fix and it should be in v3.

jonbob:
@cbegeman -- I ran a 10-year B-case baseline, as well as a comparison test with this PR. They are apparently BFB. Does this surprise you? I was running with the ECwISC30to60E3r2 mesh and had
````
 config_time_integrator = 'split_explicit'
````
in user_nl_mpaso

cbegeman:
@jonbob, Thanks for running that test. I think it's possible that these changes are compiler specific. @mark-petersen and I were actually surprised that the changes were non-BFB in any configuration, so it's totally fine that they are BFB.

jonbob:
passes:

merged to next

jonbob:
merged to master

cbegeman:
Thanks @jonbob!

PR #6047, opened 2023-11-03
Fix split-explicit upwinded advection

cbegeman:
This PR introduces a new variable to track the baroclinic water column thickness for the barotropic subcycles. The method of storing the baroclinic water column thickness in a `bottomDepthEdge` variable was problematic for upwinding because the upwinded cell can change between the baroclinic cycle and the barotropic subcycle. We now store the baroclinic (upwinded or centered) water column thickness at edges and then apply it along with the barotropic  (upwinded or centered) ssh on edges to determine the barotropic water column thickness at edges.

Fixes https://github.com/E3SM-Project/E3SM/issues/6048
[BFB]

PR #6049, opened 2023-11-03
Add tidal boundary conditions to split-explicit time integrator

cbegeman:
This PR adds the option to have a tidal boundary condition with the split-explicit time integrator. Formerly, only tidal boundary conditions were supported with RK4. The changes to the ssh and layerThickness terms are described in https://www.overleaf.com/read/htysppyxksyh#6b2d8f. 

This PR also includes options for a linearly-varying tidal boundary condition to match a test case from O'Dea et al. 2021 used for validation.

This work depends on https://github.com/E3SM-Project/E3SM/pull/6047 in that the test cases used for verification of the forcing will fail without the use of upwinded advection and that fix.

[BFB] when feature is turned off

This feature is only available in MPAS-Ocean standalone configurations.

### DIFF
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -3858,11 +3858,11 @@
 		/>
 		<var name="sshSubcycleCurWithTides" type="real" dimensions="nCells Time" units="m"
 			description="SSH - tidal potential in split explicit "
-			packages="tidalPotentialForcingPKG"
+			packages="tidalPotentialForcingPKG;tidalForcing"
 		/>
 		<var name="sshSubcycleNewWithTides" type="real" dimensions="nCells Time" units="m"
 			description="SSH - tidal potential in split explicit "
-			packages="tidalPotentialForcingPKG"
+			packages="tidalPotentialForcingPKG;tidalForcing"
 		/>
 		<var name="coastalSmoothingFactor" type="real" dimensions="nCells" units="1"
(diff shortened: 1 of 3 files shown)


cbegeman:
### Testing

This PR is BFB with https://github.com/E3SM-Project/E3SM/pull/6047 across the compass PR test suite on chrysalis with intel, openmpi.

The linear tidal forcing feature is tested with a new configuration of the `drying_slope` test case following O'Dea et al. (2021) on chrysalis with intel, openmpi in which the tide drops linearly by 1m over 3 hours.
image

rljacob:
notes:  ready but low priority

jonbob:
@cbegeman -- I asked @mark-petersen to review, but please add anyone else who makes sense as well

mark-petersen:
@cbegeman since #6047 is now merged, I rebased on master (09491ec080) and I'm testing that. The rebase had no conflicts.

mark-petersen (approved):
Compiled with intel on chrysalis and ran nightly test suite. Compares bfb against master branch point, as expected.

This all looks reasonable. The new code is protected by flags when the boundary tidal forcing is off. Approving based on @cbegeman's tests above, where the `drying slope` test has boundary forcing. Very nice!

jonbob:
passes:

merged to next

cbegeman:
@mark-petersen Thank you for your review and testing and @jonbob thanks for your testing!

jonbob:
merged to master

PR #6053, opened 2023-11-06
Add support for v3 1950 compsets

wlin7:
This PR adds supporting files for v3 1950 compsets.
Volcanic SO2 emission uses average of 1850-2014 as used for 1850 compsets
Other surface and upper level forcings set to cycle over year 1950 of the historical forcings

[BFB] Non-1950 tests are not affected.

PR #6054, opened 2023-11-06
Fix tests incompatitle with L80 and restore tentative config options for SSP

wlin7:
The PR fixes two sets of tests.

- SSP tests with tentative fix from # 5965 stops working. The _EAM%CMIP6_ generic pattern match needs to be removed to have it work. This is due to the -vbs option that would be set if having the generic pattern match. The additive style of EAM configure option would then have no mechansim to disable vbs once set.
- Update the vertical dimension for 3D variables that condidiag tests track.

[BFB] Only the above mentioned tests will diff; and the the differences for these tests 
         following this fix are due to  #5996, and other went in on the same day.


cime_config/testmods_dirs/allactive/wcprodssp/user_nl_eam
```diff
@@ -11,3 +11,6 @@
  fincl5 = 'PRECT','PRECC','TUQ','TVQ','QFLX','SHFLX','U90M','V90M'
  fincl6 = 'CLDTOT_ISCCP','MEANCLDALB_ISCCP','MEANTAU_ISCCP','MEANPTOP_ISCCP','MEANTB_ISCCP','CLDTOT_CAL','CLDTOT_CAL_LIQ','CLDTOT_CAL_ICE','CLDTOT_CAL_UN','CLDHGH_CAL','CLDHGH_CAL_LIQ','CLDHGH_CAL_ICE','CLDHGH_CAL_UN','CLDMED_CAL','CLDMED_CAL_LIQ','CLDMED_CAL_ICE','CLDMED_CAL_UN','CLDLOW_CAL','CLDLOW_CAL_LIQ','CLDLOW_CAL_ICE','CLDLOW_CAL_UN'
  fincl7 = 'O3', 'PS', 'TROP_P'
+
+! Specify an L80 IC to override eam.i from reference case, which is still for L72
+ ncdata = '$DIN_LOC_ROOT/atm/cam/inic/homme/eami_mam4_Linoz_ne30np4_L80_c20231010.nc'
```
whannah1, line 16: Would it be better for me to generate a L80 version of the file from the reference case?
This is very easy to do, but would still be non-BFB.

components/eam/cime_config/config_component.xml
```diff
@@ -51,7 +51,6 @@
 
       <value compset=""                        >-mach $MACH</value>
       <value compset="_EAM"                    >-phys default</value>
-      <value compset="_EAM%CMIP6_"             >&eamv3_phys_defaults; &eamv3_chem_defaults;</value>
```
whannah1, line 54: This will cause problems for MMF coupled compsets (currently only exist in a branch) that use "1850SOI_EAM%CMIP6-MMF1_" in the compset long name. I've been meaning to make a PR to add these compsets, but keep forgetting to do it. Can we just add another line below like we did for the "pct" and "4xCO2" modifications?

components/eam/cime_config/testdefs/testmods_dirs/eam/condidiag_dcape/user_nl_eam
```diff
@@ -3,7 +3,7 @@
  qoi_chkpt = 'RAD', 'PACEND','DYNEND','DEEPCU', 'STCLD',
 
  qoi_name = 'CAPE','dCAPE'
- qoi_nver =  1,     72,
+ qoi_nver =  1,     80,
```
whannah1, line 6: To make this more robust we should change this so that the the levels default to something like -1, and then internally make sure they get set to 1 and pver. If that's an acceptable alternative I'm happy to implement it.

cime_config/testmods_dirs/allactive/wcprodssp/user_nl_eam
wlin7: Was it already from a reference case for v3atm? I thought you did that. If not yet, yes, it would be nice to have.  Note this is the same file you  created for #5996.

whannah1: yea, I guess it's fine to use this file. I'm just worried we will forget that this override is in place, but that's probably harmless.

rljacob: If we are moving to L80,  the default ncdata should work and not need special namelist overrides.

whannah1: It sounds like this test requires a reference case (which makes sense) which is not L72. That's why the override is used. The alternative is to remap the reference case data, but then we might still need an override to make sure the correct initial condition file is used (unless we hide the L72 version where CIME won't see it).

wlin7: Yes, the SSP tests are using a reference case (a restart point of a historical run). Since we don't have v3 historical with L80, and the reference case currently used is from v2, an override is needed for SSP tests to proceed in current code base. Much about the SSP tests will be dated once all are settled.

components/eam/cime_config/config_component.xml
wlin7: @whannah1 , Will the compset be using eamv3 physics and chemistry? In that case, it is covered by line 56 (matching negative SSP lookaround) Otherwise, will need an extra line to distinguish, like the current fix for SSP. Or if set to not using vb3, need to add a negative lookaround of MMF1 to line 56.

components/eam/cime_config/testdefs/testmods_dirs/eam/condidiag_dcape/user_nl_eam
wlin7: Agreed setting to something like -1 and inside the code to resolve it to pver would be more elegant. That said, for users who do want to use this condidiag tool, it is probably more straightforward for them to specify the actual # of model levels they run with.

cime_config/testmods_dirs/allactive/wcprodssp/user_nl_eam
rljacob: Ok reference case hasn't been made yet.  Go ahead then with the override.

wlin7:
@whannah1 , thank you for generating L80 ne120np4 IC. It is added to this PR. After this and #6053, the nightly hires test should be good then.

components/eam/cime_config/config_component.xml
whannah1: After chatting with Wuyin offline I think this is ok.

whannah1: @wlin7 to answer your question - the MMF-WCYCL compsets will not use the v3 physics or chem options, so we'll need to address that later. 

As a side note - it would be really nice to be able to specify the compsets use something like "_MMF%CMIP6_", but I recall trying to do that years ago and running into a lot of problems.

components/eam/cime_config/testdefs/testmods_dirs/eam/condidiag_dcape/user_nl_eam
whannah1: right, but if most of the time they just want the upper and lower limit of the vertical levels then having it set internally makes sense, and the edge case where a user wants something else can be handled by editing these.

components/eam/cime_config/config_component.xml
rljacob: The string before the "%" is the model which is EAM.   After the percent is where you specify configure/namelist options so "EAM%MMF_CMIP6" is what you should aim to get working.

components/eam/cime_config/testdefs/testmods_dirs/eam/condidiag_dcape/user_nl_eam
wlin7: Hi @huiwanpnnl , the condidiag related tests failed at runtime for 80-level configurations.  Reseting qoi_nver to 80 from 72 will allow the tests to run. Should qoi_nver value always be the same as the vertical dimension size of the corresponding diagnostic variable?

wlin7: Thank you, @huiwanpnnl . Your suggested changes are straightforward. I am going to put them in. BTW, would it be better to keep the initialization/default value (0) unchanged at line 219 and 228, as the new implementation would interpret `-1` as `pver`. That is just to avoid confusion as the where statement has strict upper bound in the array scope.

huiwanpnnl: @wlin7, your suggestion sounds good to me. In that scenario, `-1` will be interpreted as an intentional alias for `pver`, while 0 will indicate an unset value.

wlin7: Exactly, @huiwanpnnl, thanks.. index `-1`  as used in  accessing the last element of an array in some languages.  This is mainly for pre-defined testing cases. Users can still specify the actual number of levels in practical use, which would be equivalent. The changes are in [ee14d67](https://github.com/E3SM-Project/E3SM/pull/6054/commits/ee14d677a8c4c3155c86000fa14e1bb7f87d7560) and verified working well.

huiwanpnnl (approved):
The changes for CondiDiag look good to me. Thanks!

whannah1 (approved):
Looks good!

wlin7:
Merged to next.

PR #6057, opened 2023-11-06
Homme(xx)/pgN: Add pgN feature to doubly periodic planar mode.

ambrad:
* Add pgN feature to planar doubly periodic mode.
* Add some standalone-hommexx infrastructure to support using the same unit-test exe in multiple tests.
* Add an init-planar capability to Hommexx unit tests.
* Add a unit test: gllfvremap_planar_ut, planar version of old gllfvremap_ut unit test.
* Add a BFB test for F90/C++ dycore consistency in planar mode with pg2 physics grid.

[BFB] except for new tests in homme_integration


ambrad:
Testing:
* Chrysalis homme_integration: pass except baseline check because of new tests
* Chrysalis e3sm_developer: TODO
* BFB tests on Ascent/Summit: pass
* Test merge into SCREAM repo on Chrysalis: e3sm_scream_v1 TODO
* Test merge into SCREAM repo on Ascent: TODO

components/homme/src/share/cxx/GllFvRemapImpl.cpp
```diff
@@ -514,6 +514,7 @@ ::run_dyn_to_fv_phys (const int timeidx, const Phys1T& ps, const Phys1T& phis, c
       kv.team_barrier(); // w2, w3, w4 in use
       // T_f
       loop_ik(ttrf, tvr, [&] (int i, int k) { T(ie,i,k) = th_f(i,k)*exner_f(i,k); });
+      kv.team_barrier(); // w2, w3 in use
```
bartgol, line 517: I don't understand why we need this barrier (and the one below)

ambrad: Neither is clearly needed. However, because there is still a potential source of very rare nondeterminism lurking in our code, I've been inspecting old code as I work on new related code. All the data I've collected about this rare nondeterminism can't rule out its occurring in physics-dynamics grid remap, so I read through this file again while working on the planar physgrid case. I decided to put these two barriers in base on this fresh read-through.

The one below, in my analysis, is needed if uv_dim != 3, but I decided to put it in w/o a check on that condition for the reason I'll state in a minutes. (In practice, uv_dim == 3.)

I put this one in the for the same reason, although here again the loop patterns in the two adjacent loops arguably don't require it, though in a much subtler way than the usual case of identical loop patterns.

The principle I decided to follow here was strict adherence to protecting work arrays, disregarding implicit proctection from loop patterns, just to make reasoning about determinism here simpler. I believe with these two additional team_barriers, that principle is then followed throughout the whole file.

bartgol (approved):

components/homme/src/share/gllfvremap_util_mod.F90
```diff
@@ -536,6 +591,8 @@ function gfr_check_api(hybrid, nets, nete, hvcoord, elem) result(nerr)
     integer :: nphys, ftype_in, ftype_idx, boost_idx, nerr
     logical :: boost_pg1
 
+    is_sphere = trim(geometry) /= 'plane'
```
oksanaguba, line 594: unrelated to this code, but why do we have both geometry and topology in control_mod?
```
if (topology == "cube") then
       write(iulog,*)"readnl: ne,np         = ",ne,np
else if (topology == "plane") then
  write(iulog,*)"readnl: ne_x,ne_y,np         = ",ne_x,ne_y,np
end if
```

components/homme/test_execs/thetal_kokkos_ut/CMakeLists.txt
```diff
@@ -347,3 +347,4 @@ ELSE()
 ENDIF()
 cxx_unit_test (gllfvremap_ut "${GLLFVREMAP_UT_F90_SRCS}" "${GLLFVREMAP_UT_CXX_SRCS}" "${GLLFVREMAP_UT_INCLUDE_DIRS}" "${CONFIG_DEFINES}" ${NUM_CPUS})
 TARGET_LINK_LIBRARIES(gllfvremap_ut thetal_kokkos_ut_lib)
+cxx_unit_test_add_test(gllfvremap_planar_ut gllfvremap_ut ${NUM_CPUS} "hommexx -planar")
```
oksanaguba, line 350: I thought `cxx_unit_test_add_test' macro is just to go around mpiexec in cxx_unit_test, but here it is called standalone. why?

components/homme/src/share/gllfvremap_mod.F90
```diff
@@ -2175,6 +2208,29 @@ subroutine limiter_clip_and_sum(spheremp, qmin, qmax, dp, q)
     call limiter1_clip_and_sum(np*np, spheremp, qmin, qmax, dp, q)
   end subroutine limiter_clip_and_sum
 
+  subroutine ref2plane(corners, a, b, plane)
```
oksanaguba, line 2211: i assume this is needed for SL?

components/homme/src/share/gllfvremap_util_mod.F90
```diff
@@ -168,14 +208,29 @@ subroutine set_gll_state(hvcoord, elem, nt1, nt2)
     end do
   end subroutine set_gll_state
 
+  function make_tendency(p) result(t)
```
oksanaguba, line 211: Is this a forcing function tied to the special Cinf IC? If not, what is this for?

components/homme/test/reg_test/run_tests/test-list.cmake
```diff
@@ -106,7 +106,9 @@ IF (BUILD_HOMME_THETA_KOKKOS)
     thetah-nhgw-slice-kokkos.cmake
     thetanh-nhgw-slice-kokkos.cmake
     thetanh-moist-bubble-sl.cmake
-    thetanh-moist-bubble-sl-kokkos.cmake)
+    thetanh-moist-bubble-sl-kokkos.cmake
+    thetanh-moist-bubble-sl-pg2.cmake
+    thetanh-moist-bubble-sl-pg2-kokkos.cmake)
```
oksanaguba, line 111: it seems pg2 tests in this list are planar, can we use planar in their names?

oksanaguba:
regarding "Add some standalone-hommexx infrastructure to support using the same uni-test exe in multiple tests." -- can you elaborate?

"Add a unit test: gllfvremap_planar_ut." -- which part does it test?

"Add a BFB test for F90/C++ dycore consistency." -- i assume this is for a planar pg2 test, can i modify the PR description?

components/homme/src/share/gllfvremap_util_mod.F90
ambrad: Topology is for the grid, geometry for the metric terms. Chris did that.

components/homme/test_execs/thetal_kokkos_ut/CMakeLists.txt
ambrad: This is to call the same unit test exe twice with different inputs. Before, you could only call it once or else have to create a new exe, meaning another build.

components/homme/src/share/gllfvremap_mod.F90
ambrad: This is for the metric terms in dynamics-physics grid remap. This PR is to support pg2 in planar mode.

components/homme/src/share/gllfvremap_util_mod.F90
ambrad: This is for gllfvremap internal functional property unit tests. I suggest ignoring this.

components/homme/test/reg_test/run_tests/test-list.cmake
ambrad: I can, but please note I'm following the naming scheme that I believe you established.

ambrad:
> regarding "Add some standalone-hommexx infrastructure to support using the same uni-test exe in multiple tests." -- can you elaborate?

This is to make it possible to call a unit test exe multiple times with different inputs, permitting e.g. cube-sphere and then planar runs without building another exe.

> "Add a unit test: gllfvremap_planar_ut." -- which part does it test?

It's the long-standing gllfvremap_ut test extended to planar geometry and doubly periodic topology.

> "Add a BFB test for F90/C++ dycore consistency." -- i assume this is for a planar pg2 test, can i modify the PR description?

Done.

oksanaguba (approved):
thanks

regarding renaming tests -- it is a minor thing, up to you (i am sure i will forget which ones are planar next time, but i will read nl files).

PR #6061, opened 2023-11-08
Add muller -- small internal machine at NERSC with GPU nodes similar to pm-gpu

ndkeen:
Add config to use muller -- small internal machine at NERSC with GPU nodes similar to pm-gpu.

[bfb]


rljacob (approved):

ndkeen:
merged to next

PR #6065, opened 2023-11-14
Skip building arrays for surface absorption when surface fluxes are disabled

cbegeman:
When both thickness and tracer surface fluxes are disabled with the `config_disable_*` flags, skip building the transmission coefficient arrays and related arrays.

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_forcing.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_forcing.F
@@ -56,6 +56,7 @@ module ocn_forcing
    ! Private module variables
    !
    !--------------------------------------------------------------------
+   logical :: surfaceThicknessFluxOn, surfaceTracerFluxOn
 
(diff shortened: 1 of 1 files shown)


cbegeman:
@jonbob and @mark-petersen No rush on this. It's just a small change that is convenient for certain standalone cases.

mark-petersen (approved):
This looks great. Thanks @cbegeman for your code improvement and attention to detail. I agree that the forcing arrays do not need to be built when surfaces fluxes are off. I also checked that the initialized arrays are not used anywhere else when either `config_disable_thick_sflux` or `config_disable_tr_sflux` are true.

cbegeman:
@mark-petersen Thank you for reviewing!

jonbob:
passes:

merged to next

cbegeman:
@jonbob Thank you for your testing!

jonbob:
merged to master

PR #6067, opened 2023-11-14
Removes sea surface height gradient filtering as a stealth option in E3SM

proteanplanet:
This removes sea surface height gradient filtering from MPAS-Ocean coupling with MPAS-SeaIce, which is a stealth option in E3SM.  This option was used in E3SM Version 1 to provide a temporary solution to "tiger stripes" in sea ice concentration associated with non-linear feedbacks between multiple bugs in the sea ice and ocean codes.  The "tiger stripes" were reduced by time filtering sea surface height gradient passed from the ocean to the sea ice model through the coupler.   The time filtering option was switched off in Version 2 of E3SM, but the code remained as a stealth option.   This PR completely removes this stealth option for Version 3.  The option should not be used in future. 

Changes introduced in this PR have been tested on chrysalis in a 1-month standard -resolution G-case (GMPAS-JRA1p5 TL319_EC30to60E2r2) from initial conditions. That test was BFB with the control for this code change at the end of the respective month simulations.  The code change also passes exact restart (ERS.TL319_EC30to60E2r2.GMPAS-JRA.chrysalis_intel) and processor count (PEM.T62_oQU240.GMPAS-IAF.chrysalis_intel) tests in E3SM.  The changes do not affect energy and mass conservation. Redundant ssh coupler communication code has also been removed as part of this PR.  

This change is BFB and should be included as part of the E3SM Version 3 release.

[NML]

### DIFF
--- a/components/mpas-ocean/bld/build-namelist
+++ b/components/mpas-ocean/bld/build-namelist
@@ -698,7 +698,6 @@ add_default($nl, 'config_flux_attenuation_coefficient_runoff');
 # Namelist group: coupling #
 ############################
 
-add_default($nl, 'config_ssh_grad_relax_timescale');
 if (($OCN_ICEBERG eq 'true') && ($OCN_FORCING eq 'active_atm')) {
 	add_default($nl, 'config_remove_AIS_coupler_runoff', 'val'=>".true.");
 } else {
--- a/components/mpas-ocean/bld/build-namelist-section
+++ b/components/mpas-ocean/bld/build-namelist-section
@@ -227,7 +227,6 @@ add_default($nl, 'config_flux_attenuation_coefficient_runoff');
 # Namelist group: coupling #
 ############################
 
-add_default($nl, 'config_ssh_grad_relax_timescale');
 add_default($nl, 'config_remove_AIS_coupler_runoff');
 
 ######################################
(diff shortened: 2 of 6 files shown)


proteanplanet:
Notifying @eclare108213, @darincomeau, @vanroekel, and @wlin7 of this PR who may choose to offer a review in addition to the selected reviewers.

mark-petersen (approved):
Thanks @proteanplanet, this is a good clean-up, especially since it removes two unused arrays. I compiled stand-alone with both gnu and intel on chrysalis. The nightly test suite does not exercise the MCT coupler module, but I ran it anyway to be thorough. Thanks for testing in E3SM and verifying it is BFB. That is the most important test. Everything looks good to me.

xylar (approved):
I didn't have time to review this in detail or do any testing of my own.  However, I did carefully look at the code changes and I agree that the feature being removed is always off in E3SM and is not one we ever foresee using so it is safe to remove.  The changes look reasonable and correct to me.

alicebarthel (approved):
Approved by visual inspection, this looks clean and consistent. 
I did not run additional tests.

jonbob:
passes:

jonbob:
merged to master and expected NML DIFFs blessed everywhere except anvil, which has yet to report

jonbob:
and blessed on anvil

PR #6074, opened 2023-11-17
Add wetting and drying capability to local time-stepping scheme

gcapodag:
This PR adds a wetting and drying capability to the local time-stepping (LTS) scheme for the barotropic ocean already available in master. The PR also adds several improvements for the LTS scheme not related to wetting and drying like for instance the `upwind` option for the `layerThickEdgeFlux` and inline tidal boundary conditions needed to run the drying slope test case from Compass. A separate PR on Compass will include drying slope, dam break, and parabolic bowl test cases for LTS.

This PR is bit-for-bit for E3SM mainline simulations, which do not include LTS or wetting and drying.

[BFB]

### DIFF
--- a/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F
+++ b/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F
@@ -40,6 +40,7 @@ module ocn_time_integration_lts
    use ocn_equation_of_state
    use ocn_time_average_coupled
    use ocn_time_varying_forcing
+   use ocn_wetting_drying
 
    implicit none
    private
@@ -153,7 +154,7 @@ module ocn_time_integration_lts
        layerThicknessFirstStage, & ! layer thickness at first stage of LTS
        layerThicknessSecondStage ! layer thickness at second stage of LTS
 
-    ! LTS objects
+    ! LTS Objects
     real (kind=RKIND) :: &
        dtFine, & ! fine dt, defined as dt / M
        alpha, alphaHat, beta, betaHat, gam, gamHat  ! interpolation coefficients for the predictor step of LTS
@@ -388,6 +389,10 @@ module ocn_time_integration_lts
     ! BEGIN LTS SCHEME
     !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
+    if (config_use_wetting_drying) then
+       wettingVelocityFactor(:, :) = 0.0_RKIND
+    end if    
+
     ! -------------------------------- COMPUTE SLOW TEND FOR ALL LTS REGIONS  -------------------------------
 
     call mpas_timer_start("lts slow tend computation")
@@ -400,19 +405,10 @@ module ocn_time_integration_lts
     ! --------------------------------------  STEP  ------------------------------------------
     ! --- compute the first stage of LTS for interface layer 1, interface layer 2 and coarse
 
-    ! --- update halos for diagnostic variables.
-    call mpas_timer_start("lts diagnostic halo update")
-    call mpas_dmpar_field_halo_exch(domain, 'normalizedRelativeVorticityEdge')
-    if (config_mom_del4 > 0.0_RKIND) then
-       call mpas_dmpar_field_halo_exch(domain, 'divergence')
-       call mpas_dmpar_field_halo_exch(domain, 'relativeVorticity')
-    end if
-    call mpas_timer_stop("lts diagnostic halo update")
-
     ! --- compute tendencies for thickness and fast tendencies only for velocity
     call mpas_timer_start("lts compute tendencies")
 
-    call ocn_lts_tends(statePool, LTSPool, tendPool, 1, 0, 1, 1, 1)
+    call ocn_lts_tends(statePool, LTSPool, tendPool, dt, 1, 0, 1, 1, 1)
 
     call mpas_timer_stop("lts compute tendencies")
 
@@ -428,18 +424,21 @@ module ocn_time_integration_lts
     do iRegion =1,nRegions
        do ie = 1, nEdgesInLTSRegion(iRegion,2)
           iEdge = edgesInLTSRegion(iRegion,2,ie)
-          normalVelocityFirstStage(:,iEdge) =  normalVelocityCur(:,iEdge) + dt * normalVelocityTend(:,iEdge)
+          normalVelocityFirstStage(:,iEdge) =  ( normalVelocityCur(:,iEdge) + dt * normalVelocityTend(:,iEdge) ) &
+                                               * (1.0_RKIND - wettingVelocityFactor(:, iEdge))
        end do
     end do
     ! --- coarse
     do ie = 1, nEdgesInLTSRegion(2,1)
        iEdge = edgesInLTSRegion(2,1,ie)
-       normalVelocityFirstStage(:,iEdge) =  normalVelocityCur(:,iEdge) + dt * normalVelocityTend(:,iEdge)
+       normalVelocityFirstStage(:,iEdge) =  ( normalVelocityCur(:,iEdge) + dt * normalVelocityTend(:,iEdge) ) &
+                                            * (1.0_RKIND - wettingVelocityFactor(:, iEdge))
     end do
     ! --- fine layers close to interface layers
     do ie = 1, nEdgesInLTSRegion(1,3)
        iEdge = edgesInLTSRegion(1,3,ie)
-       normalVelocityFirstStage(:,iEdge) =  normalVelocityCur(:,iEdge) + dt * normalVelocityTend(:,iEdge)
+       normalVelocityFirstStage(:,iEdge) = ( normalVelocityCur(:,iEdge) + dt * normalVelocityTend(:,iEdge) ) &
+                                           * (1.0_RKIND - wettingVelocityFactor(:, iEdge))
     end do
 
 
@@ -478,19 +477,11 @@ module ocn_time_integration_lts
 
     ! --------------------------------------  STEP  ------------------------------------------
     ! --- compute the second stage of LTS for interface layer 1, interface layer 2 and coarse
-    ! --- update halos for diagnostic variables.
-    call mpas_timer_start("lts diagnostic halo update")
-    call mpas_dmpar_field_halo_exch(domain, 'normalizedRelativeVorticityEdge')
-    if (config_mom_del4 > 0.0_RKIND) then
-       call mpas_dmpar_field_halo_exch(domain, 'divergence')
-       call mpas_dmpar_field_halo_exch(domain, 'relativeVorticity')
-    end if
-    call mpas_timer_stop("lts diagnostic halo update")
 
     call mpas_timer_start("lts compute tendencies")
 
     ! TENDENCIES COMPUTATION ---
-    call ocn_lts_tends(statePool, LTSPool, tendPool, 3, 0, 0, 1, 1)
+    call ocn_lts_tends(statePool, LTSPool, tendPool, dt, 3, 0, 0, 1, 1)
 
     call mpas_timer_stop("lts compute tendencies")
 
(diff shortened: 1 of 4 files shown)


components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F
```diff
@@ -388,6 +389,10 @@ subroutine ocn_time_integrator_lts(domain,dt)!{{{
     ! BEGIN LTS SCHEME
     !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
+    if(config_use_wetting_drying) then
```
cbegeman, line 392:
```suggestion
    if (config_use_wetting_drying) then
```

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F
```diff
@@ -1289,27 +1281,64 @@ subroutine ocn_lts_tends(statePool, LTSPool, tendPool, timeLevelIn, computeOnFin
...
+               layerThickEdgeFlux(k,iEdge) = max(layerThickness(k,cell1), layerThickness(k,cell2))
+            end if
+         end do
       end do
-   end do
+
+   else
+
+      call mpas_log_write('ERROR: config_thickness_flux_type selected not implemented for LTS')
+      call abort
```
cbegeman, line 1339:
```suggestion
      call mpas_log_write('ERROR: config_thickness_flux_type selected not implemented for LTS', messageType=MPAS_LOG_CRIT)
```

components/mpas-ocean/src/shared/mpas_ocn_wetting_drying.F
```diff
@@ -424,6 +392,91 @@ subroutine ocn_wetting_drying_wettingVelocity(layerThickEdgeFlux, layerThickness
...
+!  routine ocn_compute_wetting_velocity_factor_cell
+!
+!> \brief   Computes wettingVelocityFactor at edges of iCell
+!> \author  Giacomo Capodaglio
+!> \date    09/06/2023
+!> \details
+!>  This routine computes wettingVelocityFactor at the edges of iCell
+!
+!-----------------------------------------------------------------------
+   subroutine ocn_compute_wetting_velocity_factor_cell(wettingVelocityFactor, layerThicknessTmp, normalVelocity, iCell, k)!{{{
```
cbegeman, line 406:
```suggestion
   subroutine ocn_compute_wetting_velocity_factor(wettingVelocityFactor, layerThicknessTmp, normalVelocity, iCell, k)!{{{
```
I think it's better to leave off `cell`. It could be confusing since `wettingVelocityFactor` is an edge quantity.

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F
```diff
@@ -505,18 +496,20 @@ subroutine ocn_time_integrator_lts(domain,dt)!{{{
     ! --- coarse
     do ie = 1, nEdgesInLTSRegion(2,1)
        iEdge = edgesInLTSRegion(2,1,ie)
-       normalVelocitySecondStage(:,iEdge) = weightOld * normalVelocityCur(:,iEdge) &
-                                            + weightNew * normalVelocityFirstStage(:,iEdge) &
-                                            + weightTend * dt * normalVelocityTend(:,iEdge)
+       normalVelocitySecondStage(:,iEdge) = ( weightOld * normalVelocityCur(:,iEdge) &
+                                             + weightNew * normalVelocityFirstStage(:,iEdge) &
+                                             + weightTend * dt * normalVelocityTend(:,iEdge) ) &
+                                            * (1.0_RKIND - wettingVelocityFactor(:, iEdge))
```
cbegeman, line 502: I'm confused about why you need this factor here. Have you scaled each of the terms here already? It seems that you are double-scaling at least a few of them.

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F
```diff
@@ -733,24 +711,30 @@ subroutine ocn_time_integrator_lts(domain,dt)!{{{
...
+                                                + weightTend *  dtFine * normalVelocityTend(:,iEdge) ) & !soln update for the fine
+                                               * (1.0_RKIND - wettingVelocityFactor(:, iEdge))
        end do
        do ie = 1, nEdgesInLTSRegion(1,3)
           iEdge = edgesInLTSRegion(1,3,ie)
-          normalVelocitySecondStage(:,iEdge) =  weightOld * normalVelocityCur(:,iEdge) + weightNew * normalVelocityFirstStage(:,iEdge) &
-                                                + weightTend *  dtFine * normalVelocityTend(:,iEdge) !soln update for the fine
+          normalVelocitySecondStage(:,iEdge) = ( weightOld * normalVelocityCur(:,iEdge) + weightNew * normalVelocityFirstStage(:,iEdge) &
+                                                + weightTend *  dtFine * normalVelocityTend(:,iEdge) ) & !soln update for the fine
+                                               * (1.0_RKIND - wettingVelocityFactor(:, iEdge))
```
cbegeman, line 737: Same here, I would think you only need to scale the last term by `1-wettingVelocityFactor`

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F
```diff
@@ -733,24 +711,30 @@ subroutine ocn_time_integrator_lts(domain,dt)!{{{
        do ie = 1, nEdgesInLTSRegion(1,2)
           iEdge = edgesInLTSRegion(1,2,ie)
           normalVelocityFirstStage(:,iEdge) = normalVelocityNew(:,iEdge)
-          normalVelocityTendSum1st(:,iEdge) = normalVelocityTendSum1st(:,iEdge) + normalVelocityTend(:,iEdge)
+          normalVelocityTendSum1st(:,iEdge) = ( 1.0_RKIND / (3.0_RKIND * dt * weightTendSum1st) * normalVelocityCur(:,iEdge) & 
+                                                + normalVelocityTendSum1st(:,iEdge) + normalVelocityTend(:,iEdge) ) &
+                                              * (1.0_RKIND - wettingVelocityFactor(:, iEdge))
```
cbegeman, line 716: If `normalVelocityTend` comes from `ocn_tend_vel` then it has already been scaled by `(1.0_RKIND - wettingVelocityFactor(:, iEdge))`, but I'm a bit confused by what `normalVelocityTendSum1st` refers to since the first term does not appear to be a velocity tendency.

cbegeman (changes requested):
@gcapodag Exciting work!

I posted a few questions here because I found it hard to keep track of where you're scaling `normalVelocityTend` terms by `wettingVelocityFactor` and where you're scaling `normalVelocity` terms. In previous implementations of wetting and drying, we scale each of these only once (here, I guess it would be once per stage). Can you clarify for me how this is working here or point me to a design doc?

I haven't done any testing yet. Let me know when the compass LTS PR is ready.

mark-petersen:
@gcapodag now that the AB2 PR is merged (#5989), it is a good time to rebase these commits to the head of master. Since you merged master into this branch a few times along the way, I recommend starting with master and cherry-picking your commits. Thanks!

xylar:
> I recommend starting with master and cherry-picking your commits.

@mark-petersen, an interactive rebase in which you pick the commits you want and remove the rest from the list is an easier way to cherry-pick several commits.

components/mpas-ocean/src/shared/mpas_ocn_wetting_drying.F
gcapodag: I changed it to: `ocn_wetting_velocity_factor_on_cell_edges`

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F
gcapodag: Thanks for your feedback @cbegeman . There is no double scaling, except the one coming from previous applications of the limiter at earlier stages if that was needed. `wettingVelocityFactor` is computed in `ocn_lts_wd` and then used once in the `flux` for the `h` tendency (in the call to `ocn_tend_lts`) and then subsequently where we advance the normal velocity solution. In LTS we have ad-hoc routines to compute the tendencies (`ocn_tend_lts`) due to the splitting, so `ocn_tend_vel` is not used. I think that a double scaling could be happening though for RK4, because [here](https://github.com/E3SM-Project/E3SM/blob/839d1cf38fe80aa59a934f3ec5a2c260f3b8aa72/components/mpas-ocean/src/shared/mpas_ocn_wetting_drying.F#L281) the old velocity is limited and then again multiplied by the factor [here](https://github.com/E3SM-Project/E3SM/blob/839d1cf38fe80aa59a934f3ec5a2c260f3b8aa72/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_rk4.F#L1249). The tendency is also limited twice, where the provis solution is advanced and in the [call](https://github.com/E3SM-Project/E3SM/blob/839d1cf38fe80aa59a934f3ec5a2c260f3b8aa72/components/mpas-ocean/src/shared/mpas_ocn_tendency.F#L469) to `ocn_tend_vel`, as you mentioned.

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F
gcapodag: if we do that then the velocity would not be set to zero since the old velocity is not limited

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F
gcapodag: or, I should say, it is used only for the slow terms

gcapodag:
Thank you for your input @cbegeman , I replied to your comments, hopefully my replies make sense. As soon as https://github.com/MPAS-Dev/compass/pull/716 goes in, I will submit the compass PR as well.

gcapodag:
@mark-petersen @xylar please let me know if the "rebase" has gone through. thanks!

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F
cbegeman: @gcapodag Thanks for explaining. This makes sense to me.

gcapodag:
@cbegeman I might get the compass PR going so it will be possible to test this, and when the init mode changes to the drying slope will be merged, I will add those as well. How does this sound?

cbegeman:
> @cbegeman I might get the compass PR going so it will be possible to test this, and when the init mode changes to the drying slope will be merged, I will add those as well. How does this sound?

Yes, that sounds great. Once you post the PR then I will review it along with this PR. Merging the compass PR after the E3SM PR is merged is the way to go. Thanks!

gcapodag:
Testing for this PR can be done from this Compass PR: https://github.com/MPAS-Dev/compass/pull/738

mark-petersen:
Compiled with gnu on chicoma and intel in chrysalis. With the later, results are bfb against master branch point using nightly compass suite with default flags. That is a sanity test, as I'm not running LTS. But so far so good...

rljacob:
Needs an approval.

mark-petersen:
Since I last looked at this Dec 5, I started with this branch and merged in 09491ec080, which is master from Dec 21 just before the stand-alone bug in https://github.com/E3SM-Project/E3SM/issues/6133. I then recompiled stand-alone with intel on chrysalis, and also ran the E3SM tests:
```
cd cime/scripts
./create_test SMS_Ld62.T62_oQU240.GMPAS-IAF.chrysalis_intel -q acme-small --walltime 1:00:00
./create_test SMS_D.T62_oQU120_ais20.MPAS_LISIO_TEST.chrysalis_gnu -q acme-small --walltime 30:00
```
These both pass the compile and smoke test.

gcapodag:
@mark-petersen sounds good, thanks!

components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F
cbegeman: Thank you for this detailed explanation! That helped me understand what was going on.

gcapodag: Of course, thanks for reviewing!

gcapodag:
@cbegeman the compass PR should be ready for you try out with this one as well

gcapodag:
@cbegeman does this PR have a green light on your end?

cbegeman:
@gcapodag Thanks for reminding me to approve this. I'll do it now, given that my compass tests were successful. Great work!

cbegeman (approved):

gcapodag:
@cbegeman thank you!

mark-petersen:
Tested current head and rebased locally onto master. Compiled with intel on chrysalis and gnu on chicoma, with both debug and optimized. Ran nightly compass test suite with all (that does not test this feature, of course). Both gnu and intel are bfb on all tests against master branch point. Also tested and passed E3SM
```
SMS_Ln9.T62_oQU240.GMPAS-IAF.chrysalis_gnu
SMS_Ln9.T62_oQU240.GMPAS-IAF.chrysalis_intel
```

mark-petersen (approved):
I tested with the associated compass test cases at https://github.com/MPAS-Dev/compass/pull/738#issuecomment-1915743613. Approving based on the beautiful results in these test cases, and that there is no alterations of the mainline model results for this stealth feature.

gcapodag:
Thank you very much @mark-petersen!

mark-petersen:
FYI for the E3SM review, the Local Time Stepping feature that is altered here is only used in MPAS-Ocean standalone for single layer simulations right now.

jonbob:
Passes:
* ERS.ne30pg2_r05_EC30to60E2r2.GPMPAS-JRA.chrysalis_intel.mosart-rof_ocn_2way

merged to next

jonbob:
merged to master

PR #6080, opened 2023-11-22
[don't merge] output 3d aerosol optics -- segfault

mahf708:
A version of this code was used to output the fields of interest in v1. In v3, it segfaults. The segfaults happens in the reordering. I will follow up with the fix, but opening this PR to investigate with Ben.


rljacob:
Please do this kind of thing on your fork.

Issue #5314, opened 2022-11-17
For pm-gpu, update cudatoolkit version from 11.5 to 11.7


ndkeen:
I wanted to update cudatoolkit version on pm-gpu from 11.5 to the current system default of 11.7.
The only reason for update is to keep up with system defaults.

I don't think there are any major changes between the versions.
https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html

At one point, there was a failing test using 11.7, but that is no longer the case.
https://github.com/E3SM-Project/scream/issues/1896

ndkeen:
Testing this in SCREAM repo, I ran 2 cases that were the same except changing `cudatoolkit/11.5` to `cudatoolkit/11.7`.
The case is ne120 and runs for 3 months.  I'm only outputting one variable which is the min T. The output files are different from cprnc, but it's not clear how to tell when exactly the run diverges from BFB (if it even started BFB). Not that we expect a diff version of cuda to yield BFB.  We used to have output of floating values in atm.log.

`/pscratch/sd/n/ndk/e3sm_scratch/pm-gpu/se47-dec5/f120.F2010-SCREAMv1.ne120pg2_r0125_oRRS18to6v3.se47-dec5.gnugpu.3m.n032a4xX.Tmidout.ctk115`

compared to:

`/pscratch/sd/n/ndk/e3sm_scratch/pm-gpu/se47-dec5/f120.F2010-SCREAMv1.ne120pg2_r0125_oRRS18to6v3.se47-dec5.gnugpu.3m.n032a4xX.Tmidout.ctk117`

There was no measurable difference in the performance.

ndkeen:
I'm holding on this as there is still no reason to make the change, except it is the machine default.

ndkeen:
NERSC removed `cudatoolkit/11.5` last night after maintenance. So now I will have to move to the 11.7 version.

PR #6088, opened 2023-11-30
On pm-gpu, upgrade version of cudatoolkit from 11.5 to 11.7

ndkeen:
For pm-gpu, NERSC retired cudatoolkit/11.5 after maintenance, so we now have to move to 11.7

Fixes https://github.com/E3SM-Project/E3SM/issues/5314

This change is not BFB for pm-gpu runs.
[nbfb]


ndkeen:
merged to next

PR #6094, opened 2023-12-01
CPL: Add field exclusions list for nonlinear maps.

ambrad:
Add the user_nl_cpl namelist option nlmaps_exclude_fields. This is a list of fields that are to be excluded from the set that are nonlinearly mapped. For these excluded fields, the low-order linear map is used instead. The default is

nlmaps_exclude_fields = 'Faxa_rainc', 'Faxa_rainl', 'Faxa_snowc', 'Faxa_snowl'

[non-BFB] for cases using nlmaps
[NML] for all tests due to extra variable nlmaps_exclude_fields

### DIFF
--- a/driver-mct/cime_config/namelist_definition_drv.xml
+++ b/driver-mct/cime_config/namelist_definition_drv.xml
@@ -1713,11 +1713,27 @@
     <group>seq_infodata_inparm</group>
     <desc>
       Measure and print information about nonlinearly mapped fields. 0 means no
-      analysis is done or printed. >= 1 triggers analysis written to cpl.log.
+      analysis is done or printed. >= 1 triggers analysis written to
+      cpl.log. Modify user_nl_cpl to edit this.
       default: 0
     </desc>
     <values>
       <value>0</value>
+    </values>
+  </entry>
+
+  <entry id="nlmaps_exclude_fields">
+    <type>char(20)</type>
+    <category>nlmaps</category>
+    <group>seq_infodata_inparm</group>
+    <desc>
+      List of fields that are to be excluded from the set that are nonlinearly
+      mapped. For these excluded fields, the low-order linear map is used,
+      instead. Modify user_nl_cpl to edit this.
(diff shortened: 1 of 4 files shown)


ambrad:
e3sm_integration on Chrysalis matches the dashboard except that every test has a NML diff, as expected:
```
2023-12-01 17:43:37: NLCOMP
Comparison failed ... drv_in
  found extra variable: 'nlmaps_exclude_fields'
```

rljacob (approved):

golaz:
@ambrad : thank you for working on this quickly. We discussed it during the meeting this morning. We should go ahead with this option for v3.

golaz (approved):

rljacob:
I'm going to call this "non-BFB" all the v3 cases are using nlmaps now.  Our testing is lagging in bringing in those cases.

ambrad:
> all the v3 cases are using nlmaps now. Our testing is lagging in bringing in those cases.

I see. Do I interpret this correctly as follows?
* e3sm_integration does not yet have this option on, which is why I see only one DIFF in my testing.
* But from the v3 campaign perspective, it's not BFB?

rljacob:
right.  Also when a PR makes any test non-BFB, I prefer to call it "non-BFB except for" or "non-BFB only in"  instead of "BFB except for".  Makes it easier to find the possibly non-BFB PRs in the git log.

ambrad:
Re: testing, I checked that the following code modification makes the nlmaps test BFB with the current baseline, as expected and desired:
```diff
--- a/cime_config/testmods_dirs/allactive/nlmaps/user_nl_cpl
+++ b/cime_config/testmods_dirs/allactive/nlmaps/user_nl_cpl
@@ -1 +1,2 @@
 nlmaps_verbosity = 1
+nlmaps_exclude_fields = '','','',''
```

ambrad:
@rljacob Yes, I will do it unless you prefer someone else does.

PR #5927, opened 2023-09-15
Add new ECwISC30to60E3r2 ocean and sea-ice mesh

xylar:
Long name: ECwISC30to60L64E3SMv3r2

This mesh is a new eddy closure (EC) mesh with ice-shelf cavities (wISC) that has:
* 30 km resolution at the equator
* 60 km resolution at mid latitudes
* 35 km resolution near the poles

This mesh is a candidate for the E3SM v3 (E3) low res mesh.  It is revision 2 (r2) of the mesh.  (Revision 1 had incorrect bathymetry and coastline around Greenland.)

The mesh was created using [compass](https://github.com/MPAS-Dev/compass), specifically this tag: https://github.com/MPAS-Dev/compass/releases/tag/mesh_ECwISC30to60E3r2

A G-case with data iceberg melt fluxes (DIB) and data ice-shelf melt fluxes (DISMF) has been run for 60 years, with MPAS-Analysis output available here:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20230901.G-test.ECwISC30to60E3r2.anvil_yrs1-60/html/

The mesh and the G-case results have been reviewed and approved here:
https://acme-climate.atlassian.net/wiki/spaces/OO/pages/3904307201/Review+ECwISC30to60E3r2

A B-case using a CRYO1850 compset with data iceberg melt fluxes (DIB) and data ice-shelf melt fluxes (DISMF) has been run for 20 years, with MPAS-Analysis output available here:
[MPAS-Analysis for: 20230915.test5927.anvil](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.jwolfe/20230915.test5927.anvil/html/)

PR #6097, opened 2023-12-04
Add new ECwISC30to60E3r5 ocean and sea-ice mesh

xylar:
Long name: ECwISC30to60L64E3SMv3r2

This mesh is a new eddy closure (EC) mesh with ice-shelf cavities (wISC) that has:
* 30 km resolution at the equator
* 60 km resolution at mid latitudes
* 35 km resolution near the poles

This mesh is a candidate for the E3SM v3 (E3) low res mesh.  It is revision 5 (r5) of the mesh. This revision differs from r2 (https://github.com/E3SM-Project/E3SM/pull/5927) in having a bug fix to eliminate `landIceMask` values that erroneously appeared in the Northern Hemisphere (https://github.com/MPAS-Dev/compass/pull/732).

The mesh was created using [compass](https://github.com/MPAS-Dev/compass), and will be tagged soon as: https://github.com/MPAS-Dev/compass/releases/tag/mesh_ECwISC30to60E3r5

The mesh and the G-case results have been reviewed and approved here:
https://acme-climate.atlassian.net/wiki/spaces/OO/pages/4022535298/Review+ECwISC30to60E3r5

### DIFF
--- a/components/elm/bld/namelist_files/namelist_definition.xml
+++ b/components/elm/bld/namelist_files/namelist_definition.xml
@@ -1330,7 +1330,7 @@ Representative concentration pathway for future scenarios [radiative forcing at 
 
 <entry id="mask" type="char*20" category="default_settings"
        group="default_settings"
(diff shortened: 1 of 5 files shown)


xylar:
This is in draft mode because it's not clear if we actually want to merge this mesh or not.  In discussion with @vanroekel, @wlin7 and @jonbob now...

xylar:
We will revisit this later on if there's a clear need for this mesh.

PR #6099, opened 2023-12-04
Remove ECwISC30to60E3r2 ocean and sea ice mesh

xylar:
This version of the mesh had a bug in the `landIceMask` field and will be replaced with one of the IcoswISC30E3r* meshes.

This effectively reverts #5927

### DIFF
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -399,16 +399,6 @@
       <mask>ECwISC30to60E2r1</mask>
     </model_grid>
 
-    <model_grid alias="T62_ECwISC30to60E3r2" compset="(DATM|XATM|SATM)">
-      <grid name="atm">T62</grid>
-      <grid name="lnd">T62</grid>
-      <grid name="ocnice">ECwISC30to60E3r2</grid>
-      <grid name="rof">rx1</grid>
-      <grid name="glc">null</grid>
-      <grid name="wav">null</grid>
-      <mask>ECwISC30to60E3r2</mask>
-    </model_grid>
-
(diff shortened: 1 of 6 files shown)


proteanplanet (approved):
Approved by visual inspection.

vanroekel (approved):
approving by visual inspection

xylar:
@jonbob, I rebased to fix conflicts.  I believe this is ready to test and merge when there's a chance.

jonbob:
Passes:

merged to next

jonbob:
merged to master

@rljacob - what should we do with all the new files we have on the inputdata repo, now that this mesh is removed from e3sm?

rljacob:
Since they weren't in an official release, and you don't want them used, go ahead and delete them.  Where they ever referenced in an out-of-the-box case on master?

jonbob:
They were used by some of the Critical Path runs, but I don't think by anyone else

jonbob:
we also have files for some of the meshes we tried and did not merge for full e3sm support on:
````
/lcrc/group/e3sm/data/inputdata
````
We could cleanup those as well?

rljacob:
But a lot of the critical path settings were made with run scripts.  Can I create_newcase from a hash on master that uses those files? 

@golaz should also say if they should be kept.

jonbob:
Yes, there was a period when you could run create_newcase on master that would use this mesh and therefore need these files

maltrud:
I did quite a few runs using this grid and I'd like to perform more analysis on them.  I don't plan to run with it anymore, but would like access to the grid file.  I can make my own copy, if necessary.

PR #6112, opened 2023-12-07
Temporary vertical grid override for RRM cases

whannah1:
Due to difficulties in the generation of new initial condition files with L80 for RRM tests, it was decided that we can safely revert these tests to use L72 to ensure we maintain test coverage of the RRM capability. This needs to be undone when the new ncdata files are provided for RRM.

[non-BFB]


rljacob (approved):
so simple!

whannah1:
@rljacob @tangq Should we still merge this to get the dashboard green? Or is the new ncdata file that Qi is working on close to being ready for a PR?

rljacob:
Yes please merge it.  It can be undone in the PR that brings in the IC files.

rljacob:
I asked Az to put this on next.

PR #6108, opened 2023-12-07
Update ELM in the definition of multiple V3 compsets

bishtgautam:
The following compsets are updated to change ELM's mode from SP to active BGC mode with TOP solar parameterization:
1. F1850
2. F20TR
3. F2010
4. WCYCL1850_chemUCI-Linozv3
5. WCYCL1850_chemUCI-Linozv3-mam5
6. WCYCL20TR_chemUCI-Linozv3-mam5
7. WCYCL1850-4xCO2
8. WCYCL1850
9. WCYCL1850-1pctCO2
10. WCYCL20TR
11. WCYCL1850NS

[CC]
[non-BFB] but v3 sims already using these by runscript.
[NML]

PR #6113, opened 2023-12-08
Enable rough topograph and add remaining atm and lnd settings for v3b02

wlin7:
Rough topography for atm is enabled by replacing 16x topography with x6t ones. This PR replaces the default topo file for ne30pg2 and ne120pg2. x6t topo for other production grids are pending. Companion settings to go with rough topo are set with pgrad_correction = 1 and hv_ref_profiles = 6. These options can also be beneficial when using smoother topography.

The PR also reset dust_emis_fact and first and last year of pdepdyn stream files for 2010 and 1950 compsets to match the stream file used.

[Non-BFB] All tests with EAM, or ELM for simyr 2010 and 1950
[NML]

### DIFF
--- a/components/elm/bld/namelist_files/namelist_defaults.xml
+++ b/components/elm/bld/namelist_files/namelist_defaults.xml
@@ -670,8 +670,12 @@ this mask will have smb calculated over the entire global land surface
 
 
 <!-- Phosphorus deposition streams namelist defaults -->
+<stream_year_first_pdep use_cn=".true."   sim_year="2010" >2000</stream_year_first_pdep>
+<stream_year_last_pdep  use_cn=".true."   sim_year="2010" >2000</stream_year_last_pdep>
 <stream_year_first_pdep use_cn=".true."   sim_year="2000" >2000</stream_year_first_pdep>
 <stream_year_last_pdep  use_cn=".true."   sim_year="2000" >2000</stream_year_last_pdep>
+<stream_year_first_pdep use_cn=".true."   sim_year="1950" >2000</stream_year_first_pdep>
+<stream_year_last_pdep  use_cn=".true."   sim_year="1950" >2000</stream_year_last_pdep>
 <stream_year_first_pdep use_cn=".true."   sim_year="1850" >2000</stream_year_first_pdep>
 <stream_year_last_pdep  use_cn=".true."   sim_year="1850" >2000</stream_year_last_pdep>
 <stream_year_first_pdep use_cn=".true."   sim_year="1000" >2000</stream_year_first_pdep>
--- a/components/elm/bld/namelist_files/use_cases/1950_CMIP6_control.xml
+++ b/components/elm/bld/namelist_files/use_cases/1950_CMIP6_control.xml
@@ -11,8 +11,8 @@
 <stream_year_first_ndep phys="elm" use_cn=".true." ndepsrc="stream" >1950</stream_year_first_ndep>
 <stream_year_last_ndep  phys="elm" use_cn=".true." ndepsrc="stream" >1950</stream_year_last_ndep>
 
-<stream_year_first_pdep phys="elm" use_cn=".true." ndepsrc="stream" >1950</stream_year_first_pdep>
-<stream_year_last_pdep  phys="elm" use_cn=".true." ndepsrc="stream" >1950</stream_year_last_pdep>
+<stream_year_first_pdep phys="elm" use_cn=".true." ndepsrc="stream" >2000</stream_year_first_pdep>
+<stream_year_last_pdep  phys="elm" use_cn=".true." ndepsrc="stream" >2000</stream_year_last_pdep>
 
 <stream_year_first_popdens phys="elm" use_cn=".true." ndepsrc="stream" >1950</stream_year_first_popdens>
 <stream_year_last_popdens  phys="elm" use_cn=".true." ndepsrc="stream" >1950</stream_year_last_popdens>
--- a/components/elm/bld/namelist_files/use_cases/2010_CMIP6_control.xml
+++ b/components/elm/bld/namelist_files/use_cases/2010_CMIP6_control.xml
@@ -11,8 +11,8 @@
 <stream_year_first_ndep phys="elm" use_cn=".true." ndepsrc="stream" >2010</stream_year_first_ndep>
 <stream_year_last_ndep  phys="elm" use_cn=".true." ndepsrc="stream" >2010</stream_year_last_ndep>
 
-<stream_year_first_pdep phys="elm" use_cn=".true." ndepsrc="stream" >2010</stream_year_first_pdep>
-<stream_year_last_pdep  phys="elm" use_cn=".true." ndepsrc="stream" >2010</stream_year_last_pdep>
+<stream_year_first_pdep phys="elm" use_cn=".true." ndepsrc="stream" >2000</stream_year_first_pdep>
+<stream_year_last_pdep  phys="elm" use_cn=".true." ndepsrc="stream" >2000</stream_year_last_pdep>
 
 <stream_year_first_popdens phys="elm" use_cn=".true." ndepsrc="stream" >2010</stream_year_first_popdens>
 <stream_year_last_popdens  phys="elm" use_cn=".true." ndepsrc="stream" >2010</stream_year_last_popdens>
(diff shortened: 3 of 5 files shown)


rljacob:
@mahf708 are you sure its a vanilla F2010?  We have lots of F2010 cases in the test suite and they didn't fail after the final merge to next/master of that PR.

rljacob:
No v3 isn't in the testing yet.  That must be it.

wlin7:
@mahf708 , thanks for bringing this to our attention. I did notice since late last week's update, an option related to FAN is added. The v3 `water_cycle` configuration is not having `crop` on, and the latest v3 spin-up has `use_fan=.false.`, therefore the time bounds for fan stream files are not required.

@bishtgautam , can you please also give a check on this?

@mahf708 , regarding your other comment `Do we have F2010 tests with the additional settings for v3?`. The answer is yes, only and automatically after #6108 and this PR are merged. Your comment is a good reminder we had better merge the two PRs together.

rljacob:
Tagging @bbye and @peterdschwartz

mt5555:
out of curiosity, why do this options need to be protected inside if (defined()) blocks?   (most options dont):


```
if (defined($defaults->get_value('pgrad_correction'))) {
   add_default($nl, 'pgrad_correction', 'val' =>$nl->get_value('pgrad_correction'));
}
if (defined($defaults->get_value('hv_ref_profiles'))) {
   add_default($nl, 'hv_ref_profiles', 'val' =>$nl->get_value('hv_ref_profiles'));
}
```

mt5555 (approved):

wlin7:
> out of curiosity, why do this options need to be protected inside if (defined()) blocks? (most options dont):
> 
> ```
> if (defined($defaults->get_value('pgrad_correction'))) {
>    add_default($nl, 'pgrad_correction', 'val' =>$nl->get_value('pgrad_correction'));
> }
> if (defined($defaults->get_value('hv_ref_profiles'))) {
>    add_default($nl, 'hv_ref_profiles', 'val' =>$nl->get_value('hv_ref_profiles'));
> }
> ```

Hi @mt5555 , a default value is only available with `phys='default'`. If no default is set for a config, it would be problematic if not placing them inside the if-defined block. The alternative is to remove `phys='default' `modifier in namelist_defaults.xml. In that case, the if-defined would not be needed anymore as a default is always available.

mt5555:
@wlin7 : since these are dycore parameters independent of physics, I think it would be slightly better to adopt the alternative (remove phys='default') and make them defined for all cases.  This matches how we treat most of the other dycore namelist settings.

bbye:
I think the fix for the fan stream file is to modify the namelist_defaults.xml to use the use_fan instead of the use_cn flag. Do you want me to made a separate PR to resolve that?

components/elm/bld/namelist_files/namelist_defaults.xml
```diff
@@ -670,8 +670,12 @@ this mask will have smb calculated over the entire global land surface
 
 
 <!-- Phosphorus deposition streams namelist defaults -->
+<stream_year_first_pdep use_cn=".true."   sim_year="2010" >2000</stream_year_first_pdep>
+<stream_year_last_pdep  use_cn=".true."   sim_year="2010" >2000</stream_year_last_pdep>
 <stream_year_first_pdep use_cn=".true."   sim_year="2000" >2000</stream_year_first_pdep>
 <stream_year_last_pdep  use_cn=".true."   sim_year="2000" >2000</stream_year_last_pdep>
+<stream_year_first_pdep use_cn=".true."   sim_year="1950" >2000</stream_year_first_pdep>
+<stream_year_last_pdep  use_cn=".true."   sim_year="1950" >2000</stream_year_last_pdep>
```
bishtgautam, line 678: For `sim_year=2010` or `sim_year=1950`, why is the pdep corresponding to 2000 being used?

components/elm/bld/namelist_files/use_cases/1950_CMIP6_control.xml
```diff
@@ -11,8 +11,8 @@
 <stream_year_first_ndep phys="elm" use_cn=".true." ndepsrc="stream" >1950</stream_year_first_ndep>
 <stream_year_last_ndep  phys="elm" use_cn=".true." ndepsrc="stream" >1950</stream_year_last_ndep>
 
-<stream_year_first_pdep phys="elm" use_cn=".true." ndepsrc="stream" >1950</stream_year_first_pdep>
-<stream_year_last_pdep  phys="elm" use_cn=".true." ndepsrc="stream" >1950</stream_year_last_pdep>
+<stream_year_first_pdep phys="elm" use_cn=".true." ndepsrc="stream" >2000</stream_year_first_pdep>
+<stream_year_last_pdep  phys="elm" use_cn=".true." ndepsrc="stream" >2000</stream_year_last_pdep>
```
bishtgautam, line 15: Same comment as https://github.com/E3SM-Project/E3SM/pull/6113/files#r1440056541.

components/elm/bld/namelist_files/use_cases/2010_CMIP6_control.xml
```diff
@@ -11,8 +11,8 @@
 <stream_year_first_ndep phys="elm" use_cn=".true." ndepsrc="stream" >2010</stream_year_first_ndep>
 <stream_year_last_ndep  phys="elm" use_cn=".true." ndepsrc="stream" >2010</stream_year_last_ndep>
 
-<stream_year_first_pdep phys="elm" use_cn=".true." ndepsrc="stream" >2010</stream_year_first_pdep>
-<stream_year_last_pdep  phys="elm" use_cn=".true." ndepsrc="stream" >2010</stream_year_last_pdep>
+<stream_year_first_pdep phys="elm" use_cn=".true." ndepsrc="stream" >2000</stream_year_first_pdep>
+<stream_year_last_pdep  phys="elm" use_cn=".true." ndepsrc="stream" >2000</stream_year_last_pdep>
```
bishtgautam, line 15: Same comment as https://github.com/E3SM-Project/E3SM/pull/6113/files#r1440056541.

wlin7: Good point, @bishtgautam . Intuitively, when first testing it, did try 2010. But look like only simyr2000 stream file is available, which is used for piControl as well, so I thought it being treated as time-invariant. If time varying for this data stream is necessary, we will need quick help to generate the files for compsets of different TIME (or forcing scenarios)
The file is : `inputdata/lnd/clm2/pdepdata/fpdep_clm_hist_simyr2000_1.9x2.5_c150929.nc`

rljacob:
Should go in with #6108

rljacob:
merged to next.

wlin7:
Note: FDPSCREAM-ARM97 test returns diff on 11 of 243 eam.h0 fields (QDIFF, TDIFF, and *_RES variables). Other common fields (e.g.,  T, Q, U, V, P3 fields, precipitation and radiation fields) are not affected. 
@bogensch , do you have some ideas about such behavior of having those atm field differ for a supposedly BFB PR for DPSCREAM? 

Here mainly to document the test, the PR should go ahead nevertheless.

rljacob:
Noting other test issues:  
ERS_Ld5.ne4_oQU240.F2010.chrysalis_intel.eam-rrtmgp - build fail from compiler license
SMS_D_Ln5.ne4_oQU240.F2010.mappy_gnu  - long time failing test.
SMS_D_Ln5.ne4_oQU240.F2010.pm-cpu_intel.eam-clubb_sp - runtime fail for a test that has been known to fail for no reason.

rljacob:
Re: SMS_Ln9_P24x1.ne4_ne4.FDPSCREAM-ARM97, it does have the 2 new atm_in namelist vars:
pgrad_correction               =  1
hv_ref_profiles                =  6

rljacob:
I'm going to go ahead and merge this to master and bless all diffs.  I'll open an issue about that SCREAM test to track it.

wlin7:
Thanks for taking care of the integration, @rljacob . Yes, we will look into the DPSCREAM behavior separately.

rljacob:
e3sm_developer_next_gnu reported later.  Only one test did not run: SMS_Ld1.hcru_hcru.I1850CRUELMCN.pm-cpu_gnu for unknown reasons.  Diffs blessed.

rljacob:
ERS_Ld5.ne4_oQU240.F2010.chrysalis_intel.eam-rrtmgp ran the following day and was blessed.

PR #6103, opened 2023-12-06
Update pgi compiler flags

amametjanov:
Update pgi compiler flags:
- add missing -lstdc++ to pgi and nvidia linker flags
- fix mpas-ocean openacc nightly
- enable mpas openacc on nvidiagpu + pm-gpu
- fix other miscellaneous errors.

[BFB]

PR #6116, opened 2023-12-10
Fix bugs in OpenACC directives and revise machine files

hyungyukang:
Fix bugs in OpenACC directives:
- Change an openacc directive in mpas_ocn_diagnostics.F to avoid
getting stuck on pm-gpu.

Fixes E3SM-Project/E3SM#6127

[BFB]

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
@@ -1106,18 +1106,19 @@ contains
 #ifdef MPAS_OPENACC
       !$acc parallel loop &
       !$acc    present(kineticEnergyVertex, normalVelocity, &
-      !$acc            areaTriangle, dcEdge, dvEdge, edgesOnVertex)
+      !$acc            areaTriangle, dcEdge, dvEdge, edgesOnVertex) &
+      !$acc    private(iEdge,rTmp,k,i)
 #else
       !$omp parallel
       !$omp do schedule(runtime) private(i, k, iEdge, rTmp)
 #endif
       do iVertex = 1, nVerticesAll
          kineticEnergyVertex(:, iVertex) = 0.0_RKIND
+         !$acc loop seq
          do i = 1, vertexDegree
             iEdge = edgesOnVertex(i, iVertex)
             rTmp  = dcEdge(iEdge)*dvEdge(iEdge)*0.25_RKIND/ &
                     areaTriangle(iVertex)
-            !$acc loop seq
             do k = 1, nVertLevels
                kineticEnergyVertex(k,iVertex) = &
                kineticEnergyVertex(k,iVertex) + rTmp* &
@@ -1133,7 +1134,8 @@ contains
       !$acc parallel loop &
       !$acc    present(kineticEnergyVertexOnCells, kineticEnergyVertex,&
       !$acc            kiteIndexOnCell, kiteAreasOnVertex, invAreaCell,&
-      !$acc            verticesOnCell, nEdgesOnCell)
+      !$acc            verticesOnCell, nEdgesOnCell) &
+      !$acc    private(i,j,k,iVertex,invAreaCell1)
 #else
       !$omp do schedule(runtime) &
       !$omp    private(i, j, k, iVertex, invAreaCell1)


ndkeen:
I'm seeing changes here that will have impact for other cases. We may need to find a way to make changes that will only affect MPAS

hyungyukang:
> I'm seeing changes here that will have impact for other cases. We may need to find a way to make changes that will only affect MPAS. If MPAS needs a flag, then maybe should only be applied to those sources. If you need `-acc`, maybe only use that flag for the sources required. And please don't change the GPU bindings unless we all agree upon those changes as they impact all cases.

@ndkeen , Thanks for your instructions. I'm going to revert some changes I made. It seems like the best way to add `-acc` to `nvidiagpu` in MPAS is to copy and modify `Depends.pgigpu.cmake`. What do you think about this?

cime_config/machines/Depends.pm-gpu.nvidiagpu.cmake
```diff
@@ -2,9 +2,13 @@ list(APPEND REDUCE_OPT_LIST
   homme/src/share/derivative_mod_base.F90
 )
 
+# add accelerator/gpu flags for MPAS files
```
grnydawn, line 5: According to @ndkeen's comment, this setting might be placed in the Depends file, not in this cmake file, in order to limit the scope of impact to specified MPAS files.

cime_config/machines/cmake_macros/nvidiagpu.cmake
```diff
@@ -4,14 +4,15 @@ set(MPIFC "ftn")
...
+
 string(APPEND CPPDEFS " -DFORTRANUNDERSCORE -DNO_R16 -DCPRNVIDIA")
 string(APPEND CPPDEFS_DEBUG " -DYAKL_DEBUG")
 if (compile_threaded)
   string(APPEND CMAKE_C_FLAGS " -mp")
 endif()
 string(APPEND CMAKE_C_FLAGS_RELEASE " -O2")
 string(APPEND CMAKE_C_FLAGS_DEBUG " -g")
-string(APPEND CMAKE_Fortran_FLAGS " -i4 -Mstack_arrays -Mextend -byteswapio -Mflushz -Kieee -DHAVE_IEEE_ARITHMETIC -Mallocatable=03 -DNO_R16 -traceback")
+string(APPEND CMAKE_Fortran_FLAGS " -i4 -Mstack_arrays -Mextend -byteswapio -Mflushz -Kieee -DHAVE_IEEE_ARITHMETIC -Mallocatable=03 -DNO_R16 -traceback -acc")
```
grnydawn, line 15: According to @ndkeen's comment, this setting might be placed in the Depends file, not in this cmake file, in order to limit the scope of impact to specified MPAS files.

grnydawn (changes requested):
I also think that we may need to restrict the scope of impact from this changes within MPAS. Please see my comments in my review. I prefer to have a GPU bind suggested in this PR instead of "none". But it may need to coordinate with other developers who may impact to the GPU bind change.

amametjanov:
Rebased the branch onto latest master and checked that this PR still fixes the hang on PM-GPU: https://my.cdash.org/viewTest.php?buildid=2460796 .

hyungyukang:
@amametjanov , thanks for rebasing this PR!

@grnydawn , many changes I made in this PR were duplicates of #6103 (merged) except for `mpas_ocn_diagnostics.F`, so they disappeared by rebasing this PR.

@ndkeen , this PR fixes #6116.

ndkeen (approved):
Now need https://github.com/E3SM-Project/E3SM/pull/6166 to allow building, but after that, this fix allows the test to run

grnydawn (approved):
@amametjanov, the changes that Hyun made in "mpas_ocn_diagnostics.F." look good to me.

However, my test case (compset: GMPAS-IAF, res: T62_EC30to60E2r2) on Hyun's code branch (hkang/ocean/pm-gpu-fix) fails during building SPIO with the following error message:

```
"/global/u2/y/youngsun/repos/github/E3SM/externals/scorpio/src/clib/pioc.c", line 1282: error: too many arguments in invocation of macro "adios2_init"
          ios->adiosH = adios2_init(ios->adios_comm, adios2_debug_mode_on);
```

Since this error is not related to the changes in the MPAS code, I believe this PR is ready to be merged.

PR #6126, opened 2023-12-20
Update MALI version to include higher-order advection and time integration

trhille:
This merge includes numerous updates to MALI

The major updates include:
• Higher-order, flux-corrected thickness and tracer transport.
• Three strong stability-preserving Runge-Kutta time integrations schemes (a two-stage, second-order scheme; three- and four-stage third-order schemes)

Minor updates include:
• Earlier update of clock within time step
• Prevent unrealistic calving into holes within an ice shelf
• Change default start and stop times to avoid unpredictable I/O behavior
• Higher-order grounding-line flux calculation
• Several small bug fixes

[NML] - only for cases with MALI
[BFB]


components/mpas-albany-landice/src/mode_forward/Makefile
```diff
@@ -4,9 +4,11 @@
 OBJS =  mpas_li_core.o \
     mpas_li_core_interface.o \
     mpas_li_time_integration.o \
-    mpas_li_time_integration_fe.o \
```
matthewhoffman, line 7: Because this module has been renamed, do we need to update the cmake build system that E3SM uses?  @jonbob , do you know how to do that?

components/mpas-albany-landice/src/shared/Makefile
```diff
@@ -3,6 +3,8 @@
 
 OBJS =  mpas_li_constants.o \
         mpas_li_mask.o \
+        mpas_li_mesh.o \
```
matthewhoffman, line 6: Noting another place where E3SM build system may need updating

components/mpas-albany-landice/src/mode_forward/Makefile
jonbob: I believe we just need to modify ~/components/mpas-albany-landice/src/landice.cmake

matthewhoffman:
@jonbob , this PR will also need updates to the namelist generation files in E3SM because we've added some new namelist options.  Let me know if you are able to do that or prefer that I do it.

jonbob:
Thanks for pointing that out, @matthewhoffman -- I'm happy to do that

jonbob:
@matthewhoffman -- I did update the bld files to match the Registry changes, and I think they're mostly straightforward. The only part I'm not sure of is in 
````
components/mpas-albany-landice/bld/build-namelist
````
We're adding some new advection configs and it's not clear to me if they need to be wrapped in logic for MALI running in its dynamic mode like we do for config_thickness_advection and config_tracer_advection:
````
#############################
# Namelist group: advection #
#############################

if ($MALI_DYNAMIC eq 'TRUE') {
    add_default($nl, 'config_thickness_advection');
} else {
    add_default($nl, 'config_thickness_advection', 'val'=>"none");
}
if ($MALI_DYNAMIC eq 'TRUE') {
    add_default($nl, 'config_tracer_advection');
} else {
    add_default($nl, 'config_tracer_advection', 'val'=>"none");
}
add_default($nl, 'config_horiz_tracer_adv_order');
add_default($nl, 'config_advection_coef_3rd_order');
add_default($nl, 'config_restore_thickness_after_advection');
add_default($nl, 'config_zero_sfcMassBalApplied_over_bare_land');
````
@trhille?

trhille:
> We're adding some new advection configs and it's not clear to me if they need to be wrapped in logic for MALI running in its dynamic mode like we do for config_thickness_advection and config_tracer_advection:


@jonbob, the new advection settings should not have any effect on the default MALI dynamic mode.

matthewhoffman (approved):
@trhille , thanks for putting this together.  This is good timing with our push starting up to get MALI runs more operational in E3SM.  

@jonbob , I rebased this to bring in a bug fix we just added to MALI that will be useful to have right away in E3SM.  I also added a new commit that adds new modules to the landice.cmake file used for building MALI in E3SM.  This looks good to me, pending final testing in E3SM.

jonbob:
Passes BFB with expected NML DIFFs for:

BFB and no NML differences for:

ready to be merged

jonbob:
merged to next

jonbob:

PR #6128, opened 2023-12-20
Improve robustness of Kokkos build script

jgfouca:
It seems that, sometimes, the kokkos CMake will change the CMAKE_CXX_COMPILER to an absolute path before storing it in the cache. Subsequent calls to cmake that do not have -DCMAKE_CXX_COMPILER=$absolute_path_to_cxx will be interpreted as a changed compiler, which results in the CMakeCache being flushed and then results in CMAKE_INSTALL_PREFIX being lost.

This is a known flaw in CMake that we must work around.

Tested on mappy by building `SMS_P12x2.ne4_oQU240.WCYCL1850NS` which requires Kokkos.

[BFB]

### DIFF
--- a/share/build/buildlib.kokkos
+++ b/share/build/buildlib.kokkos
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-import os, sys, argparse, logging
+import os, sys, argparse, logging, shutil
 
 from standard_script_setup import *
 from CIME import utils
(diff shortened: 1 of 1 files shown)


ambrad (commented):
Looks good. compy e3sm_prod_maint-2_1_intel should cover this in the nightlies.

What about main-2.0? anvil e3sm_prod_maint-2_0_intel would cover that case.

ambrad (approved):

jgfouca:
@ambrad , it looks like maint-2.0 predates the integration of kokkos into the E3SM build system.

PR #6163, opened 2024-01-20
add dependabot for github actions

mahf708:
With this, dependabot will perform dependency checks on things we use inside GitHub Actions at a configurable frequency (for now, weekly). If there are updates, it will issue a PR requesting reviews from members (for now, Luca and Naser) and assigning the PR to an integrator (for now, Rob).

------

TODO: Naser will monitor how the bot behaves initially; in particular, we likely need to edit the gh-pages workflows with this change:
```diff
    - uses: actions/checkout@v4
      with: 
+       persist-credentials: false

```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    assignees:
+      - "rljacob"
+    reviewers:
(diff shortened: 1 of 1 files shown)


bartgol:
I don't mind being volunteered. But I don't really understand what the action will do...

rljacob:
This won't start working until its merged, right?  I'm fine being volunteered.

bartgol (approved):
After chatting with Naser, I now understand what the action does. Looks like a good thing to automate.

Issue #6156, opened 2024-01-19
Build issue with nvidiagpu on pm-gpu after maintenance


ndkeen:
After Jan17th perlmutter maintenance, we don't need to change any module versions. So far, all tests have worked except for tests with nvidiagpu.

With `SMS_Ld1.T62_oEC60to30v3.CMPASO-NYF.pm-gpu_nvidiagpu`, I get build fail is csm:

```
cmake -DCIME_MODEL=e3sm -DSRC_ROOT=/global/cfs/cdirs/e3sm/ndk/repos/me35-jan18  -Dcompile_threaded=FALSE  -DGPU_TYPE=None -DGPU_OFFLOAD=None -DINSTALL_SHAREDPATH=/pscratch/sd/n/ndk/e3sm_scratch/pm-gpu/me35-jan18/SMS_Ld1.T62_oEC60to30v3.CMPASO-NYF.pm-gpu_nvidiagpu.nv\
227/bld/nvidiagpu/mpich/nodebug/nothreads/mct  -DUSE_KOKKOS=FALSE  -DCOMP_NAMES='driver;datm;slnd;dice;mpaso;drof;sglc;swav;siac;sesp' -DCASEROOT="/pscratch/sd/n/ndk/e3sm_scratch/pm-gpu/me35-jan18/SMS_Ld1.T62_oEC60to30v3.CMPASO-NYF.pm-gpu_nvidiagpu.nv227" -DCASETOOL\
S="/pscratch/sd/n/ndk/e3sm_scratch/pm-gpu/me35-jan18/SMS_Ld1.T62_oEC60to30v3.CMPASO-NYF.pm-gpu_nvidiagpu.nv227/Tools" -DCIMEROOT="/global/cfs/cdirs/e3sm/ndk/repos/me35-jan18/cime" -DSRCROOT="/global/cfs/cdirs/e3sm/ndk/repos/me35-jan18" -DCOMP_INTERFACE="mct" -DCOMPI\
[...]

PR #6166, opened 2024-01-22
For pm-gpu, add module gcc-mixed/11.2.0 for nvidiagpu builds.

ndkeen:
For pm-gpu, add module gcc-mixed/11.2.0 for nvidiagpu builds.
After the NERSC maintenance, the underlying GNU version is now too high for the cudatoolkit we are using.
But this is only an issue with nvidia compilers as the GNU builds set the version to 11.
All we do here is add a module `gcc-mixed/11.2.0` for nvidiagpu builds.

Fixes https://github.com/E3SM-Project/E3SM/issues/6156

[bfb]

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -384,6 +384,7 @@
       <modules compiler="nvidiagpu">
         <command name="load">cudatoolkit/11.7</command>
         <command name="load">craype-accel-nvidia80</command>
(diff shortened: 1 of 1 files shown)


rljacob (approved):

ndkeen:
merged to next

ndkeen:
Last test shows the two nvidiagpu tests passing, merging to master

PR #5811, opened 2023-07-16
Change elm restart file for wcprodssp allactive test

peterdschwartz:
Replaced the NaN values for TWS_MONTH_BEGIN with a fill value of `1.e+36` in the elm restart file.

Fixes #5665

PR #6206, opened 2024-02-03
Fix restart error on maint-2.0 

### LINKED ISSUE #6195: maint-2.0 branch appears broken on Perlmutter
The maint-2.0 branch (b583f7d) is having problems on Perlmutter (CPU nodes).  Any run that restarts will fail with the following PIO error in the e3sm.log file.
`
PIO: FATAL ERROR: Aborting... An error occured, Writing variables (number of variables = 180) to file (./20240129.restart_test2.maint_20.elm.h0.0001-02.nc, ncid=143) using PIO_IOTYPE_PNETCDF iotype failed. Non blocking write for variable (TWS_MONTH_BEGIN, varid=205) failed (Number of subarray requests/regions=1, Size of data local to this process = 433). NetCDF: Numeric conversion not representable (err=-60). Aborting since the error handler was set to PIO_INTERNAL_ERROR... (/pscratch/sd/b/beharrop/temp/e3sm_code/20240129/E3SM/externals/scorpio/src/clib/pio_darray_int.c: 395)
`

I tried to reproduce this with a shorter test using `./create_test ERS.ne4_oQU240.F2010`, but that did not reproduce the error. `ERS.ne4_oQU240.F2010.pm-cpu_intel (Overall: PASS)`.  I have also been unable to reproduce this error on either Chrysalis or Compy.  The smallest reproducible setup I've been able to come to that has this error is the following.

```
./create_newcase --case ${casename}  --compset F2010 --res ne4_oQU240 --walltime 00:08:00 --script-root ${casedir} --output-root ${outputdir} -mach pm-cpu

cd ${casedir}

./xmlchange STOP_N=1
./xmlchange STOP_OPTION=nmonths
./xmlchange RESUBMIT=1

./case.setup
./case.build
./case.submit
```

I have been using the intel compiler, and haven't tested any of the others yet.  Has anyone else encountered anything like this?

peterdschwartz:
Add endwb to ELM restart file in maint-2.0. 

[BFB]
Fixes #6195

### DIFF
--- a/cime_config/testmods_dirs/allactive/wcprodssp/user_nl_elm
+++ b/cime_config/testmods_dirs/allactive/wcprodssp/user_nl_elm
@@ -1,7 +1,7 @@
 ! fsurdat used is not the same file for the reference historical run (as recorded in elm.r's global attribute)
 
  CHECK_FINIDAT_FSURDAT_CONSISTENCY       = .false.
-
+ finidat = "$DIN_LOC_ROOT/e3sm_init/V2.SSP370_SSP585.ne30pg2_EC30to60E2r2/v2.LR.historical_0101/2015-01-01-00000/v2.LR.historical_0101.elm.r.noNaN.2015-01-01-00000.nc"
  hist_dov2xy = .true.,.true.
  hist_fincl2 = 'H2OSNO', 'FSNO', 'QRUNOFF', 'QSNOMELT', 'FSNO_EFF', 'SNORDSL', 'SNOW', 'FSDS', 'FSR', 'FLDS', 'FIRE', 'FIRA'
  hist_mfilt = 1,365
--- a/components/elm/src/data_types/ColumnDataType.F90
+++ b/components/elm/src/data_types/ColumnDataType.F90
@@ -1698,7 +1698,7 @@ contains
     ! Read/Write column water state information to/from restart file.
     !
     ! !USES:
-    use elm_varctl, only : use_lake_wat_storage
+    use elm_varctl, only : use_lake_wat_storage, do_budgets
     !
     ! !ARGUMENTS:
     class(column_water_state) :: this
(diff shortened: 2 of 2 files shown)


peterdschwartz:
Need to check if maint-2.0 will also need PR #5811

peterdschwartz:
I ran ` SMS_D_Ld1.ne30pg2_EC30to60E2r2.WCYCLSSP370.pm-cpu_intel.allactive-wcprodssp` with no issue, so leaving this as is.

ndkeen:
I might suggest adding in the fix in 5811.  I think I needed that for the tests in maint-2.1 to pass. I had meant to add it here as well anyway, but the e3sm_production tests did pass.

peterdschwartz:
Went ahead and made that change as well.

ndkeen:
OK, you likely know more about why this is or is not a good idea.

I did try `SMS_Ld1.ne30pg2_EC30to60E2r2.WCYCLSSP585.pm-cpu_intel.allactive-wcprodssp`
with and without that finidat and it seemed ok?

ndkeen (approved):

rljacob:
@ndkeen can an ERS test pass on maint-2.0 without these fixes?

ndkeen:
No, before these changes `ERS_Lm3.ne4_oQU240.F2010` would fail.  But it also fails with even old maint-2.0 checkouts -- ie I think the test would have always failed.

rljacob:
Oh that's right,  you have to run it for a long time.  A standard-length ERS passes.  How long does Lm3 take?  We should add an ERS test to the prod suite for maint-2.0 to catch this.

ndkeen:
My test that passes after this fix used a job that ran for a total of 8:45 on 1 node of pm-cpu.

mingxuanwupnnl:
@peterdschwartz I met the same issue while I was doing production runs. I wonder if this bug fix PR is BFB? Thank you.

peterdschwartz:
> @peterdschwartz I met the same issue while I was doing production runs. I wonder if this bug fix PR is BFB? Thank you.

It's BFB

rljacob:
@rebassoo merged.

rebassoo:
Great, thanks @rljacob!

PR #6213, opened 2024-02-06
Add Aurora functionality and YAKL updates to GordonBell Branch

omarkahmed:


oksanaguba:
@omarkahmed would you please explain what changes did yakl need for aurora? thanks!

omarkahmed:
@oksanaguba , the current YAKL hash is not functional either on aurora or sunspot.  We have a choice: (1) apply the minimum necessary patch to the repo branch and make that a public hash: https://github.com/omarkahmed/YAKL/commits/omarkahmed/xyuan-20231213/, or (2) move the YAKL hash up to the tip.  The former would (most simply) require creating a new public hash in Matt's YAKL repo associated with my fork.  The latter requires some cmake updates, and there are some outstanding issues such as the "-g" flag + AoT. I think it's an which direction we want to go here.

oksanaguba:
thanks, Omar. i must say i am not following, but yakl is used in scream (a version of e3sm), and i was able to build scream with aot flags. my runs crash, so i do not know if yakl submodule in scream will have issues. sounds like i'd better use your version of yakl anyway.

rljacob:
This is a PR to another branch so this will not affect master.

mrnorman:
Looks fine to me.

mrnorman (approved):

PR #6218, opened 2024-02-07
Prefer HINTS to PATHS in cmake find modules

### LINKED ISSUE #6217: Current impl of FindNETCDF.cmake picks up system installation before the one pointed by netcdf cmake vars
The issue is that in cmake's `find_XYZ` routines, the entries passed via `PATHS` are scanned _after_ system paths. There are two solutions.

1. Use `HINTS` instead of `PATHS`: these entries are scanned before system ones, so it would work. The drawback is that CMake states that "These should be paths computed by system introspection, such as a hint provided by the location of another item already found. Hard-coded guesses should be specified with the PATHS option".
2. Use the `NO_SYSTEM_PATH` option (possibly with some other `NO_XYZ` option): this categorically excludes general system paths (like `/usr/`) from the search. The drawback is that we _cannot_ rely on system installation anymore, even if a machine had a system version of a tpl that works for us. That said, I don't think this will likely be the case for production machines.

Imho, option 1 is better. Although we don't follow CMake's advice of using `PATHS` for hard-coded guesses, we achieve the desired result, and don't rule out system libraries. Imho, the fact that `PATHS` is scanned after system paths is a huge flaw of CMake: if someone takes the effort of giving CMake a guess, its guess should have the priority.

bartgol:
For all the cmake `find_XYZ` utilities, there is a well defined ordered list in which CMake scans paths while searching. In this list, the paths passed via the argument `HINTS` appear early on. On the other hand, the paths passed via the argument `PATHS` appear quite late. After trying `HINTS` and before trying `PATHS`, CMake consider other options, among which we have

- system paths (e.g., `/usr` directory on linux)
- cmake package registry (this is only for `find_package`, but can be _quite_ annoying and hard to find, as we experienced with EKAT)

CMake states that HINTS and PATHS should be used for different reasons. In particular, regarding `HINTS` it states that

> These should be paths computed by system introspection, such as a hint provided by the location of another item already found. Hard-coded guesses should be specified with the PATHS option.

However, the fact that `PATHS` is scanned after system folders makes its usage more complicated. Since `HINTS` and `PATHS` are treated in the same way (meaning, CMake grabs the paths contained in them, and just scan them), from the practical point of view, the difference between the two is just the fact that one is considered much earlier than the other.

The alternative would have been to add `NO_SYSTEM_PATH` to the `find_XYZ` calls, but it comes with a drawback: if on some testing machine (a shared workstation, perhaps) we _do_ have a valid installation of a tpl in the system paths, we would not pick it up. Using `HINTS`, we can still pick up the system install (leaving HINTS empty).

Fixes #6217 .

### DIFF
--- a/components/cmake/modules/FindCsmShare.cmake
+++ b/components/cmake/modules/FindCsmShare.cmake
@@ -24,7 +24,7 @@ endif()
 set(CSM_SHARE "${INSTALL_SHAREDPATH}/${COMP_INTERFACE}/${ESMFDIR}/${NINST_VALUE}/csm_share")
 
 # Look for libcsm_share in the complex path we built above
-find_library(CSM_SHARE_LIB csm_share REQUIRED PATHS ${CSM_SHARE})
+find_library(CSM_SHARE_LIB csm_share REQUIRED HINTS ${CSM_SHARE})
 
 set(CSM_LIBS  ${CSM_SHARE_LIB};mct;spio)
 set(CSM_INCLS ${CSM_SHARE})
--- a/components/cmake/modules/FindMCT.cmake
+++ b/components/cmake/modules/FindMCT.cmake
@@ -16,8 +16,8 @@ if (TARGET mct)
 endif()
 
 # Look for libmct in INSTALL_SHAREDPATH/lib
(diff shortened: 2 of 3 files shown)


rljacob:
@jgfouca you can merge this.

jgfouca:
Merged to next.

PR #6221, opened 2024-02-08
Adding IcoswISC30E3r5 ice initial condition with DIB

darincomeau:
This adds an initial condition for MPAS-Seaice for cases where the `IcoswISC30E3r5` mesh is paired with a `CRYO*` compset, which utilize data iceberg (DIB) forcing. The new initial condition is from a restart from a one month `GMPAS-JRA1p5-DIB-PISMF` case.

The current sea ice IC for this mesh in B-cases ('spun-up' ICs) was not run with DIB forcing, and as such MPAS-Seaice cannot use it as a restart for a case with DIB.

This also adds `CRYO1850` and `CRYO1850-DISMF` tests with this mesh to the e3sm_integration test suite, which prior to this would not pass.

[BFB] Only impacts configurations with `IcoswISC30E3r5` and a `CRYO*` compset, which prior to this would seg fault in ice during initialization.

Credit to @xylar for rooting out the problem.

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -305,6 +305,8 @@ _TESTS = {
             "SMS_D_Ld1.ne30pg2_r05_EC30to60E2r2.WCYCL1850",
             "SMS_Ln5.ne30pg2_ne30pg2.F2010-SCREAM-LR-DYAMOND2",
             "ERS_Ld3.ne30pg2_r05_EC30to60E2r2.WCYCL1850.allactive-nlmaps",
+            "SMS_D_Ld1.ne30pg2_r05_IcoswISC30E3r5.CRYO1850",
+            "SMS_D_Ld1.ne30pg2_r05_IcoswISC30E3r5.CRYO1850-DISMF",
             )
         },
 
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -305,7 +305,6 @@ _TESTS = {
             "SMS_D_Ld1.ne30pg2_r05_EC30to60E2r2.WCYCL1850",
             "SMS_Ln5.ne30pg2_ne30pg2.F2010-SCREAM-LR-DYAMOND2",
             "ERS_Ld3.ne30pg2_r05_EC30to60E2r2.WCYCL1850.allactive-nlmaps",
-            "SMS_D_Ld1.ne30pg2_r05_IcoswISC30E3r5.CRYO1850",
             "SMS_D_Ld1.ne30pg2_r05_IcoswISC30E3r5.CRYO1850-DISMF",
             )
         },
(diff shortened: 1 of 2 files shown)


darincomeau:
Tested

`SMS.ne30pg2_IcoswISC30E3r5.WCYCL1850.chrysalis_intel`, confirmed existing sea ice IC is still used.

`SMS.ne30pg2_IcoswISC30E3r5.CRYO1850.chrysalis_intel`,
`SMS.ne30pg2_IcoswISC30E3r5.CRYO1850-DISMF.chrysalis_intel`
confirmed new sea ice IC is used.

xylar:
@darincomeau, would you be willing to cherry-pick the commit from https://github.com/E3SM-Ocean-Discussion/E3SM/pull/77 onto this branch?  I think it would be good to add a test that would have caught this issue so we make sure this configuration stays functional.

darincomeau:
@xylar I agree on adding the test, thanks for adding that to Ocean-Discussion, I missed that. I can add it here, but do you think it should be it's own PR?

xylar:
@darincomeau, I think we should add those tests here because a big part of the motivation in adding them is to show that the configurations work that this PR is meant to fix.  I can split them into a separate PR if there are objections to including them here.

darincomeau:
Ok I understand. I've cherry-picked that test commit over here and updated the description above.

@jonbob let us know if there's concern on the content here.

jonbob:
@rljacob -- this PR adds two new tests to e3sm_integration. Do you want to move one to e3sm_prod, since they really represent a cryo production configuration? I can run tests using the testing pelayout and document how long they take to run

rljacob:
Sure move one of the cryo cases to production.

cime_config/tests.py
```diff
@@ -305,6 +305,8 @@
             "SMS_D_Ld1.ne30pg2_r05_EC30to60E2r2.WCYCL1850",
             "SMS_Ln5.ne30pg2_ne30pg2.F2010-SCREAM-LR-DYAMOND2",
             "ERS_Ld3.ne30pg2_r05_EC30to60E2r2.WCYCL1850.allactive-nlmaps",
+            "SMS_D_Ld1.ne30pg2_r05_IcoswISC30E3r5.CRYO1850",
```
xylar, line 308: @darincomeau, I think we should move this one to production and leave DISMF here.  Do you want me to do that?

xylar: Thanks!

darincomeau:
Also, I'm not sure where is a good place to document this, but cases without DIB can use the sea ice IC with DIB, so for future meshes we can just make one IC with DIB to avoid the extra file.

jonbob (approved):
approved based on visual inspection and testing

jonbob:
Passes:
* SMS_Ld1.ne30pg2_r05_IcoswISC30E3r5.WCYCL1850.chrysalis_intel.allactive-wcprod_1850_r05

Also verified that both new tests:
* SMS_D_Ld1.ne30pg2_r05_IcoswISC30E3r5.CRYO1850-DISMF
* SMS_D_Ld1.ne30pg2_r05_IcoswISC30E3r5.CRYO1850
pass using the chrysalis testing layout and that the new file is available on the inputdata repo

merged to next

jonbob:
merged to master and new baselines requested for compy_prod, chrysalis_prod, and chrysalis_integration. Will need to do the same for pm-cpu_integration when it reports next, as well as anvil_prod which is failing all tests with the new Icos mesh. I'm currently testing to see if those fails are related to the size of the test pelayout on anvil

PR #6223, opened 2024-02-08
Provenance: Make it so Git never tries to use a pager

jgfouca:
When git tries to page output, it blocks the entire process.

[BFB]


jgfouca:
This turned out to be a diff-tool issue, not a pager issue.

PR #6226, opened 2024-02-08
Update and add new support for configurations with dynamic Greenland ice sheet component

stephenprice:
This PR updates and/or adds new support that allows for simulations, through
existing IG- and BG- compsets, with a dynamic MALI component for the 
Greenland ice sheet. These configurations are still under development and are 
not yet scientifically supported.

Changes include:

- added support for Greenland 20km res., trigrid configuration
- updated Greenland 1-to-10km res. grid and input files 
- add support for configs. using new 1-to-10km grid and init. cond.
- updated defaults for MALI history (output) files
- updated default Albany solver parameter settings
- added SMS, ERS, and ERP tests for 20km res. IG configuration
- updated machine file support for Albany & Trilinos on pm-cpu & Chrysalis
- add default layouts for 20km & 1-to-10km configs. on pm-cpu & Chrysalis
- allow MALI global and regional analysis AMs to operate during e3sm runs


stephenprice:
This PR includes updated support for a low-resolution, 20km uniform Greenland configuration (primarily for testing and development) and a high / variable 1-to-10km Greenland configuration (intended for eventual use in production runs and science applications).

To set up and run these, use the commands detailed below. The example here is for Chrysalis but the same should work on Perlmutter cpu by substituting "pm-cpu" for "chrysalis" (note the currently, Albany libs are only supported on Perlmutter and Chrysalis).

**IG cases (data atmos., active lnd and glc)**

_Greenland 20km:_  
`./create_newcase -case /caseDir/caseName -compiler gnu -mach chrysalis -project e3sm -compset IGELM_MLI -res ne30pg2_r05_EC30to60E2r2_gis20`

_Greenland 1-to-10km:_ (note that while this configuration runs and builds, the PE layout has not yet been optimized and MALI convergence will likely be slow due to the small numbers of PEs alloted to it by default)
`./create_newcase -case /caseDir/caseName -compiler gnu -mach chrysalis -project e3sm -compset IGELM_MLI -res ne30pg2_r05_EC30to60E2r2_gis1to10r02`

**BG cases (fully coupled including active glc)**

_Greenland 20km:_
`./create_newcase -case /caseDir/caseName -compiler gnu -mach chrysalis -project e3sm -compset BGWCYCL1850 -res ne30pg2_r05_EC30to60E2r2_gis20`

_Greenland 1-to-10km:_
`./create_newcase -case /caseDir/caseName -compiler gnu -mach chrysalis -project e3sm -compset BGWCYCL1850 -res ne30pg2_r05_EC30to60E2r2_gis1to10r02`

**Low-resolution tests**

Additionally, 3 IG-case tests have been added using the Greenland 20km configuration. These include a "smoke" test (SMS.) and two restart tests (ERS. and ERP.). They can be run using the following commands. 

`./create_test SMS.ne30pg2_r05_EC30to60E2r2_gis20.IGELM_MLI.chrysalis_gnu.elm-gis20kmSMS --project e3sm -g -o -b gis20.IGELM_MLI.baseline.SMS`
(-g, -o, and -b flags generate new test baselines and overwrite the existing ones) 

`./create_test ERS.ne30pg2_r05_EC30to60E2r2_gis20.IGELM_MLI.chrysalis_gnu.elm-gis20kmERS --project e3sm`

`./create_test ERP.ne30pg2_r05_EC30to60E2r2_gis20.IGELM_MLI.chrysalis_gnu.elm-gis20kmERP --project e3sm`

For now, since no formal baselines exist the main thing that needs to be confirmed for these is that the tests are created, built, submitted, and run to completion (i.e., in your default scratch space, look in the /run directories under the test name (identified by their initial letters 'SMS', 'ERS', 'ERP') and confirm that zipped log files exist and output .nc files were created).

cime_config/config_grids.xml
```diff
@@ -2885,6 +2915,12 @@
       <desc>mpas.gis1to10km is a variable-resolution, from 1- to 10-km, MALI grid of the Greenland Ice Sheet.</desc>
     </domain>
 
+    <domain name="mpas.gis1to10kmR2">
+      <nx>427386</nx>
+      <ny>1</ny>
+      <desc>mpas.gis1to10km is a variable-resolution, from 1- to 10-km, MALI grid of the Greenland Ice Sheet.</desc>
```
matthewhoffman, line 2921:
```suggestion
      <desc>mpas.gis1to10kmR2 is a variable-resolution, from 1- to 10-km, MALI grid of the Greenland Ice Sheet.</desc>
```

Also, if we are keeping both the original and R2 versions for some reason, update the descriptions to indicate what the difference is

cime_config/config_grids.xml
```diff
@@ -4823,6 +4859,31 @@
       <map name="GLC2OCN_SMAPNAME">cpl/gridmaps/mpas.gis20km/map_mpas.gis20km_to_oEC60to30v3_aave.181115.nc</map>
     </gridmap>
 
+    <gridmap glc_grid="mpas.gis20km" lnd_grid="ne30np4.pg2">
+      <map name="LND2GLC_FMAPNAME">cpl/gridmaps/ne30pg2/map_ne30pg2_to_gis20km_mono.230510.nc</map>
+      <map name="LND2GLC_SMAPNAME">cpl/gridmaps/ne30pg2/map_ne30pg2_to_gis20km_bilin.230510.nc</map>
+      <map name="GLC2LND_FMAPNAME">cpl/gridmaps/mpas.gis20km/map_gis20km_to_ne30pg2_mono.230510.nc</map>
+      <map name="GLC2LND_SMAPNAME">cpl/gridmaps/mpas.gis20km/map_gis20km_to_ne30pg2_mono.230510.nc</map>
+    </gridmap>
```
matthewhoffman, line 4867: Are these actually needed?  I don't see a new `model_grid` specification that uses this combination of component grids.

cime_config/config_grids.xml
```diff
@@ -1678,6 +1688,16 @@
       <mask>oEC60to30v3</mask>
     </model_grid>
 
+    <model_grid alias="ne30pg2_r05_EC30to60E2r2_gis1to10" compset="_MALI">
```
matthewhoffman, line 1691: see note below - do we really want to add both this and the R2 version?

cime_config/config_grids.xml
```diff
@@ -4823,6 +4859,31 @@
...
+    </gridmap>
+
+    <gridmap glc_grid="mpas.gis20km" lnd_grid="r05">
+      <map name="LND2GLC_FMAPNAME">cpl/gridmaps/r05/map_r05_to_gis20km_mono.230510.nc</map>
+      <map name="LND2GLC_SMAPNAME">cpl/gridmaps/r05/map_r05_to_gis20km_bilin.230510.nc</map>
+      <map name="GLC2LND_FMAPNAME">cpl/gridmaps/mpas.gis20km/map_gis20km_to_r05_mono.230510.nc</map>
+      <map name="GLC2LND_SMAPNAME">cpl/gridmaps/mpas.gis20km/map_gis20km_to_r05_mono.230510.nc</map>
+    </gridmap>
+
+    <gridmap glc_grid="mpas.gis20km" ocn_grid="EC30to60E2r2">
```
matthewhoffman, line 4876: Question for @jonbob  - Do the ICE mapping files below need to live in a separate `gridmap` entry that specifies `ice_grid` instead of `ocn_grid`?  Or is there a special handling that always assumes ICE grids match OCN grids, so that's not necessary?

cime_config/config_grids.xml
```diff
@@ -4870,6 +4931,35 @@
       <map name="GLC2OCN_ICE_RMAPNAME">cpl/gridmaps/mpas.gis1to10km/map_gis1to10km_to_EC30to60E2r2_aave.210304.nc</map>
     </gridmap>
 
+    <!-- ====================    -->
+    <!-- GLC: mpas.gis1to10kmR2  -->
+    <!-- ====================    -->
+      
+    <gridmap glc_grid="mpas.gis1to10kmR2" lnd_grid="r05">
+      <map name="LND2GLC_FMAPNAME">cpl/gridmaps/r05/map_r05_to_gis1to10r02_mono.230725.nc</map>
```
matthewhoffman, line 4939: Everywhere above this grid has the suffix `R2` but here it has `r02`.  @jonbob , is that discrepancy ok?

cime_config/machines/config_machines.xml
```diff
@@ -291,8 +292,7 @@
       <env name="ADIOS2_ROOT">$SHELL{if [ -z "$ADIOS2_ROOT" ]; then echo /global/cfs/cdirs/e3sm/3rdparty/adios2/2.9.1/cray-mpich-8.1.25/aocc-4.0.0; else echo "$ADIOS2_ROOT"; fi}</env>
     </environment_variables>
     <resource_limits>
-      <resource name="RLIMIT_STACK">-1</resource>
-    </resource_limits>
+      <resource name="RLIMIT_STACK">-1</resource> </resource_limits>
```
matthewhoffman, line 295: remove this linebreak change

components/elm/bld/namelist_files/namelist_defaults.xml
```diff
@@ -592,6 +592,7 @@ this mask will have smb calculated over the entire global land surface
 <!-- MALI grids that could be used if the glc_grid logic was working. -->
```
matthewhoffman, line 592: @stephenprice  or @jonbob, do you know what this comment is referring to?

components/elm/bld/namelist_files/namelist_defaults.xml
```diff
@@ -592,6 +592,7 @@ this mask will have smb calculated over the entire global land surface
...
 <fglcmask hgrid="0.9x1.25" glc_grid="mpas.gis20km">glc/cism/griddata/glcmaskdata_0.9x1.25_Gland5km.nc</fglcmask>  <!-- the CISM one works fine for this mesh -->
 <fglcmask hgrid="0.9x1.25" glc_grid="mpas.ais20km">lnd/clm2/griddata/glcmaskdata_0.9x1.25_60S.nc</fglcmask>
+<fglcmask hgrid="r05"      glc_grid="mpas.gis20km">lnd/clm2/griddata/glcmaskdata_0.5x0.5_everywhere.nc</fglcmask>
```
matthewhoffman, line 595: What is the `everywhere` in this file name referring to?  I don't think we'd ever want glacier land units to be active everywhere on the LND grid if that's what it means - I think that would have a noticeable degradation on ELM performance.

components/elm/bld/namelist_files/namelist_defaults.xml
```diff
@@ -592,6 +592,7 @@ this mask will have smb calculated over the entire global land surface
...
 <fglcmask hgrid="r05"      glc_grid="mpas.gis1to10km">lnd/clm2/griddata/glcmaskdata_0.5x0.5_GIS_AIS.nc</fglcmask>
```
matthewhoffman, line 596: Do we need to add a glcmask spec for the new `mpas.gis1to10kmR2` mesh too?

components/elm/cime_config/config_pes.xml
```diff
@@ -523,4 +523,42 @@
       </pes>
     </mach>
   </grid>
+  <grid name="a%ne30np4.pg2_l%r05_oi%null_r%r05_g%mpas.gis20km_w%null_z%null_m%EC30to60E2r2">
```
matthewhoffman, line 526: I'm always confused by what should go into which config files.  Does it make sense to put this in the ELM pes config file?

components/elm/cime_config/testdefs/testmods_dirs/elm/gis20kmERP/user_nl_mali
```diff
@@ -0,0 +1,8 @@
+config_velocity_solver = 'FO'
+config_thickness_advection = 'fo'
+config_thermal_solver = 'none'
```
matthewhoffman, line 3: Do we want to be explicitly setting these settings (and the calving ones) in a test, as opposed to using the nl defaults?  Hardcoding specific physics options seems fragile because it wouldn't catch the effects of updating nl defaults.

components/elm/cime_config/testdefs/testmods_dirs/elm/gis20kmERS/user_nl_mali
```diff
@@ -0,0 +1,8 @@
+config_velocity_solver = 'FO'
+config_thickness_advection = 'fo'
+config_thermal_solver = 'none'
```
matthewhoffman, line 3: same as above for ERP

components/elm/cime_config/testdefs/testmods_dirs/elm/gis20kmERP/user_nl_mali
```diff
@@ -0,0 +1,8 @@
...
+config_pio_num_iotasks = 47
+config_pio_stride = 64
```
matthewhoffman, line 5: Also, are you sure these are being used?  This implies the test is using 23.5 nodes on pm-cpu, which is definitely not correct for 20km GIS.  Is it possible the pio settings are getting set automatically later, and these mods aren't actually doing anything (and should therefore be removed)?

components/elm/src/main/elm_varcon.F90
```diff
@@ -248,7 +248,7 @@ subroutine elm_varcon_init()
     allocate( dzsoifl(1:nlevsoifl            ))
 
     if (use_extrasnowlayers) then
-       h2osno_max = 10000._r8
+       h2osno_max = 30000._r8
```
matthewhoffman, line 251: I don't think we want to be changing this in this PR.  @stephenprice , on the call today, Chloe described a new PR she will make that includes this plus a few firn parameter tuning improvements that were used in the long firn spinups.

components/mpas-albany-landice/bld/namelist_files/albany_input.yaml
```diff
@@ -1,11 +1,29 @@
 %YAML 1.1
 ---
 ANONYMOUS:
+# In order to use ML, change Tpetra to Epetra in the following line, 
+# and "Preconditioner Type: MueLu" to " Preconditioner Type: ML" several lines below 
   Build Type: Tpetra
```
matthewhoffman, line 6: Side note: this is going away, and we might as well remove it now.  See https://github.com/MALI-Dev/E3SM/pull/107

components/mpas-albany-landice/bld/namelist_files/albany_input.yaml
```diff
@@ -1,11 +1,29 @@
...
+    LandIce Field Norm:
+      sliding_velocity_basalside:
+        Regularization Type: Given Value
+        Regularization Value: 1.0e-4
+    LandIce BCs:
+      BC 0:
+        Cubature Degree: 4
+        Basal Friction Coefficient:
+          Type: Power Law
+          Power Exponent: 1.0 #left blank to cause an error if not filled
```
matthewhoffman, line 18: this comment is confusing

components/mpas-albany-landice/bld/namelist_files/albany_input.yaml
```diff
@@ -22,13 +40,21 @@ ANONYMOUS:
...
+            Scale Type: Unscaled
+            Tolerance: 1.0e-03
+          Test 1:
+            Test Type: RelativeNormF
+            Norm Type: Two Norm
+            Tolerance: 0.9999
         Test 1: 
           Test Type: MaxIters
-          Maximum Iterations: 50
+          Maximum Iterations: 100
```
matthewhoffman, line 57: Do we definitely want to raise the max iterations this high?  In my experience, if you hit 50, you're not going to converge anyway, and you waste a lot of time with a larger number.  But maybe you've identified this is needed for this mesh?

components/mpas-albany-landice/cime_config/buildnml
```diff
@@ -77,6 +77,11 @@ def buildnml(case, caseroot, compname):
         grid_prefix += 'gis_1to10km_r01'
```
matthewhoffman, line 77: Relate to comments above: Do we want to remove this?

components/mpas-albany-landice/driver/glc_comp_mct.F
```diff
@@ -892,6 +892,19 @@ subroutine glc_run_mct( EClock, cdata_g, x2g_g, g2x_g)!{{{
 !         call mpas_stream_mgr_reset_alarms(stream_manager, direction=MPAS_STREAM_INPUT, ierr=err_tmp)
```
matthewhoffman, line 892: Also delete this block of comments starting with `In coupled mode, we don't want to use forcing from a file`, as they are no longer relevant.  You could add a warning that if any input ("forcing") files here contain fields that the coupler passes, the input file versions here will clobber the fields from the coupler.  And maybe a note about what fields we imagine needing to read in here.

driver-mct/cime_config/config_component.xml
```diff
@@ -1085,7 +1085,7 @@
 
   <entry id="GLC_GRID">
     <type>char</type>
-    <valid_values>gland20,gland10,gland5,gland5UM,gland4,mpas.aisgis20km,mpas.gis20km,mpas.ais20km,mpas.gis1to10km,null</valid_values>
+    <valid_values>gland20,gland10,gland5,gland5UM,gland4,mpas.aisgis20km,mpas.gis20km,mpas.ais20km,mpas.gis1to10km,mpas.gis1to10kmR2,null</valid_values>
```
matthewhoffman, line 1088: flagging one more place we might want to delete old gis1to10km grid reference.  

We might also want to remove the CISM meshes

matthewhoffman (changes requested):
@stephenprice , this is a huge step forward to supporting a dynamic Greenland Ice Sheet in E3SM.  Most of my comments are clarification questions.  One repeating question is if we want to remove the old 1to10km Greenland mesh if it was never fully supported in E3SM and we are now replacing it with a better one that we'll use going forward.

cime_config/allactive/config_pesall.xml
```diff
@@ -325,7 +325,7 @@
       </pes>
     </mach>
     <mach name="anvil">
-      <pes compset="JRA_ELM.+MPASSI.+MPASO.+MOSART" pesize="any">
+      <pes compset="JRA_ELM.+MPASSI.+MPASO.+MOSART+SGLC" pesize="any">
```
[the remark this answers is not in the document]
stephenprice: I can't answer the question about the merge commit. But, these stub glc additions are needed in order for default PE layouts with an active glc component to be correctly recognized. Without these additions, our new configurations were continually "finding" incorrect PE layouts. This was the only working solution that Jon and I could come up with.

cime_config/config_archive.xml
```diff
@@ -129,9 +128,8 @@
     <test_file_names>
       <tfile disposition="copy">rpointer.glc</tfile>
       <tfile disposition="copy">casename.mali.rst.1976-01-01_00000.nc</tfile>
-      <tfile disposition="copy">casename.mali.rst.am.timeSeriesStatsMonthly.1976-01-01_00000.nc</tfile>
-      <tfile disposition="move">casename.mali.hist.am.globalStats.1976-01-01.nc</tfile>
-      <tfile disposition="move">casename.mali.hist.am.highFrequencyOutput.1976-01-01_00.00.00.nc</tfile>
+      <tfile disposition="move">casename.mali.hist.1976-01-01_00000.nc</tfile>
```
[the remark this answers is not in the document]
stephenprice: This was just templated from the other MPAS component sections of this same file. I don't think the 1976 date matters and nothing I changed here changed anything w.r.t. how the current testing infrastructure treats MPAS (MALI) hist files (basically, it ignore them). I have another set of communications on this (see the "Notes from Jason Boutte" section of [this page](https://acme-climate.atlassian.net/wiki/spaces/PSC/pages/3687612417/Support+for+E3SM+configurations+with+a+dynamic+land+ice+MALI+component)) but my thinking at the moment is that maybe we leave that out of this PR so that we can experiment with and fine tune what's needed here as part of a future PR.

cime_config/config_grids.xml
```diff
@@ -2885,6 +2915,12 @@
       <desc>mpas.gis1to10km is a variable-resolution, from 1- to 10-km, MALI grid of the Greenland Ice Sheet.</desc>
     </domain>
 
+    <domain name="mpas.gis1to10kmR2">
```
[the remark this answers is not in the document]
stephenprice: I don't know. We can certainly remove the support for this if that's what everyone agrees to. One reason I thought about leaving it in there for at least the near future is because I thought it could be a useful benchmark in terms of showing the performance and "scientific validity" (e.g., init. cond. optimization) improvements that have resulted from both this new init. cond. and the other code changes.

cime_config/config_grids.xml
stephenprice: See response to the above question. I can certainly update the description though. Thanks for flagging that.

cime_config/config_grids.xml
stephenprice: I don't entirely remember, but I think that we (Jon and I) decided to keep support (for now) for related bi-grid configurations. E.g., this could be used with the majority of this grid spec ...

```
    <model_grid alias="ne30pg2_r05_EC30to60E2r2_gis20" compset="_MALI">
      <grid name="atm">ne30np4.pg2</grid>
      <grid name="lnd">r05</grid>
      <grid name="ocnice">EC30to60E2r2</grid>
      <grid name="rof">r05</grid>
      <grid name="glc">mpas.gis20km</grid>
      <grid name="wav">null</grid>
      <mask>EC30to60E2r2</mask>
    </model_grid>
```
... to add support for a similar bi-grid configuration.

cime_config/config_grids.xml
stephenprice: I will leave this question open for now (see initial response to this above and comment further there).

cime_config/config_grids.xml
stephenprice: Good catch. I don't have an answer for that at the moment (will wait for Jon's response).

cime_config/machines/config_machines.xml
stephenprice: Done. Thanks for catching that.

components/elm/bld/namelist_files/namelist_defaults.xml
stephenprice: I don't know about this one. Blame has Jon making this comment in 2018.

components/elm/bld/namelist_files/namelist_defaults.xml
stephenprice: It is what it sounds like - the mask is true (ones) "everywhere". Since we're really only using the 20km GIS grid for testing and debugging, I figured this was ok (I've not noted any performance hit -- it's usually being run on only 1 or 2 nodes -- but it's certainly possible there is one). I made a few other attempts to "fix" the existing mask that was resulting in runtime errors but was unsuccessful. We can certainly do better if this seems important.

components/elm/bld/namelist_files/namelist_defaults.xml
stephenprice: I'm not sure I understand why there's been no complaint about this. Maybe Jon does? My assumption is that it is somehow working fine using the old mask. Something we should probably understand better.

components/elm/cime_config/config_pes.xml
stephenprice: I don't know. This was where Jon suggested it should go. Note that the default layouts for BG configs is in `./cime_config/allactive/config_pesall.xml`.

components/elm/cime_config/testdefs/testmods_dirs/elm/gis20kmERP/user_nl_mali
stephenprice: I think the tests need to run "out of the box" by default w/o anyone making any changes. I.e., I think they need to run just by executing the lines noted above (in the description for how to "use" capabilities in this PR under "Low-resolution tests"). Jon should be able to clarify how this should work. But note that when you create a test instance, it gets created, built, and submitted all at the same time. So there's not an obvious / easy way to break between those steps to make changes to things like namelists, xml files, etc.

components/elm/cime_config/testdefs/testmods_dirs/elm/gis20kmERP/user_nl_mali
stephenprice: These are old settings that I've been carrying around since Cori, so we should definitely update them to be correct for the respective machines (PM and Chyrs here). Or remove them if they are not needed?

components/elm/src/main/elm_varcon.F90
stephenprice: That's fine, but a few weeks ago she asked me to just include this here. I'm happy to remove it.

components/mpas-albany-landice/bld/namelist_files/albany_input.yaml
stephenprice: Ok. Just remove this these lines? The related section? Both?

components/mpas-albany-landice/bld/namelist_files/albany_input.yaml
stephenprice: @mperego can you help us w/ some clarification for this comment? Thanks.

components/mpas-albany-landice/bld/namelist_files/albany_input.yaml
stephenprice: This has actually made a difference. I agree that this could be problematic in the long term for the reasons you note, but right now, this change has at least allowed the new 1-to-10km init. cond. to run successfully in E3SM for several time steps (I've not yet tested / confirmed exactly how many time steps can be run).

components/mpas-albany-landice/cime_config/buildnml
stephenprice: @mperego?

components/mpas-albany-landice/driver/glc_comp_mct.F
stephenprice: I've added the following:

```
         ! NOTE: If any input ("forcing") files here contain fields that the coupler also passes, 
         ! the input file versions here will clobber the fields passed from the coupler. 
```
 Offhand, I don't have a good feeling for what fields you are imagining ... sfc mass balance? Let me know if you have a list you think I should add.

driver-mct/cime_config/config_component.xml
stephenprice: Good catch. I removed the CISM meshes. Leaving the 1to10 variants for now until we decide what to do there.

components/mpas-albany-landice/bld/namelist_files/albany_input.yaml
mperego: I think the comment is for standalone Albany where we do not have defaults for `"Power Exponent`" and so we throw an error if not provided. 
That's not the case with MALI. I would simply remove the comment.

components/mpas-albany-landice/bld/namelist_files/albany_input.yaml
mperego: 100 iterations is not that bad (especially for the initial solve) because we are now using adaptive tolerance of the linear solvers, so the linear solvers only take a few iterations per nonlinear solver iteration.

components/mpas-albany-landice/cime_config/buildnml
mperego: Is this a grid we do not use? I'm fine removing it if that's the case

components/mpas-albany-landice/bld/namelist_files/albany_input.yaml
stephenprice: I will remove it. Thanks.

components/mpas-albany-landice/cime_config/buildnml
stephenprice: Sorry @mperego -- I got confused there and didn't necessarily mean to ping you on this. You can ignore this.

components/elm/cime_config/testdefs/testmods_dirs/elm/gis20kmERP/user_nl_mali
stephenprice: @mperego -- this may be where I meant to flag you for further discussion. Thanks.

components/mpas-albany-landice/bld/namelist_files/albany_input.yaml
stephenprice: @mperego -- and possibly here as well?

mperego: Just remove the comments.
We'll take care of the "Build Type: Tpetra" in a subsequent PR.

stephenprice: Got it. Thanks.

components/elm/cime_config/testdefs/testmods_dirs/elm/gis20kmERP/user_nl_mali
mperego: I don't know whether these are used or not. How many partitions are you using for this test?

I think Jerry will soon submit a subsequent PR, so if you want you can postpone to then these performance related issues.

trhille:
Testing:
-------
Chrysalis:
- [ ] IGelm 20km
- [ ] IGelm 1–10km
- [ ] BGelm 20km
- [ ] BGelm 1–10km
- [ ] SMS
- [ ] ERS
- [ ] ERP

Perlmutter:
- [ ] IGelm 20km
- [ ] IGelm 1–10km
- [ ] BGelm 20km
- [ ] BGelm 1–10km
- [ ] SMS
- [ ] ERS
- [ ] ERP

trhille (changes requested):
@stephenprice, this is excellent! Note that the default number of procs (128) fails with an out-of-memory error on Perlmutter for the IG case with 1-10km GIS mesh. I was able to run it on 4 nodes (512 procs). I also have one small question (see below), but otherwise it looks great.

components/elm/cime_config/config_pes.xml
```diff
@@ -523,4 +523,42 @@
...
+        <comment>GIS 20km (low-res) testing config</comment>
+        <MAX_MPITASKS_PER_NODE>128</MAX_MPITASKS_PER_NODE>
+        <MAX_TASKS_PER_NODE>128</MAX_TASKS_PER_NODE>
+        <ntasks>
+          <ntasks_atm>128</ntasks_atm>
+          <ntasks_lnd>128</ntasks_lnd>
+          <ntasks_rof>128</ntasks_rof>
+          <ntasks_ice>128</ntasks_ice>
+          <ntasks_ocn>128</ntasks_ocn>
+          <ntasks_glc>128</ntasks_glc>
```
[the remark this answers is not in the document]
jewatkins: It might be worth retrying this with https://github.com/stephenprice/E3SM/pull/1 . The current albany install was using a higher SFAD size which increases memory usage (possibly by exactly 4x).

cime_config/config_grids.xml
jonbob: It is always assumed/required that ocn and ice are on the same grid

cime_config/config_grids.xml
jonbob: They're sort of intermixed but it's defined on line 2915 as R2 -- so "glc_grid" needs to match. The mapping file names don't need to, though it may create some confusion. I'll need to regenerate a lot of these mapping files to match the v3 naming and algorithm convention, so we can decide as I do that

components/elm/bld/namelist_files/namelist_defaults.xml
matthewhoffman: I see - I wasn't aware the existing one was causing an error.  Activating glacier land units for the low res testing mesh seems fine.

Did you try it with `lnd/clm2/griddata/glcmaskdata_0.5x0.5_GIS_AIS.nc`?  My guess is the error comes when the GLC mesh extends beyond the masked region on the LND grid, which could be the case for the 20km GIS mesh.  I'm asking because down the road we may need to make our GIS meshes extend farther to ensure they have overlap with the OCN meshes, so if there is an extent issue with the 20km GIS mesh, it will likely show up then too.  (not necessary to figure out for this PR)

components/elm/bld/namelist_files/namelist_defaults.xml
matthewhoffman: Seeing this again, I remember now that this just needs to extend larger than the GLC mesh so that masked LND cells exist for all GLC mesh cells.  If the existing mask file provides that overlap then we do not need a new mask file (and it would be best to keep the number of mask files to a minimum).  I'll resolve this conversation.

components/elm/cime_config/testdefs/testmods_dirs/elm/gis20kmERP/user_nl_mali
matthewhoffman: What I'm wondering is if we should choose to set these settings as the default namelist values for MALI if it's what we want running in the tests.

components/elm/cime_config/testdefs/testmods_dirs/elm/gis20kmERP/user_nl_mali
matthewhoffman: @jonbob , do you know how these MPAS PIO namelist settings are handled in E3SM?

components/mpas-albany-landice/bld/namelist_files/albany_input.yaml
matthewhoffman: Ok, thanks for confirming that

components/mpas-albany-landice/driver/glc_comp_mct.F
matthewhoffman: Thanks, that comment seems good

components/elm/cime_config/config_pes.xml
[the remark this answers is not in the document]
stephenprice: Thanks @jewatkins. I'm not sure what I need to do to allow you to add this commit to my branch. Let me know if you do. I'm happy to add it and test it out.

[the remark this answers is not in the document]
jewatkins: @stephenprice If you're okay with the changes, there should be a "Merge" button at the bottom of this page https://github.com/stephenprice/E3SM/pull/1. That should add my changes.

cime_config/allactive/config_pesall.xml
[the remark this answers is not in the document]
matthewhoffman: Ok, sounds good.  If it's not going to cause grief for other compsets and it helps for the new compsets, then that seems like a good solution.

cime_config/config_grids.xml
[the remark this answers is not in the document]
matthewhoffman: Ok, that seems a valid reason to keep it for now then.

cime_config/config_archive.xml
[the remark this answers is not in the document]
matthewhoffman: Ok, thanks - either way is fine with me

components/elm/bld/namelist_files/namelist_definition.xml
```diff
@@ -215,12 +215,10 @@ Normally this is ONLY used when running CESM with the active glacier model.
 </entry>
 
 <entry id="glc_grid" type="char*10" category="elm_physics"
-        group="elm_inparm" valid_values="gland20,gland10,gland5,gland5UM" >
+        group="elm_inparm" valid_values="mpas.gis20km,mpas.gis1to10km" >
```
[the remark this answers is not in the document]
stephenprice: I'm not sure. It seems to work as is ... but maybe we should either add the R2 spec OR replace this one with that. It depends on if we still want support for both grids or not. Let's leave this item until we resolve that one?

driver-mct/cime_config/config_component.xml
matthewhoffman: Ok, sounds good

components/elm/cime_config/config_pes.xml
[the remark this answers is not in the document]
stephenprice: @trhille -- right now, there's only a default PE layout for the IG case for the 20km mesh. I will add one now for IG cases using the 1-to-10km grid. Thanks for catching that.

components/elm/src/main/elm_varcon.F90
matthewhoffman: @chloewhicker recently opened a PR that includes this change with a few other changes for the firn model in https://github.com/E3SM-Project/E3SM/pull/6298, so I think it should be removed from this PR.

components/elm/cime_config/testdefs/testmods_dirs/elm/gis20kmERP/user_nl_mali
matthewhoffman: @jonbob said: these can/should be left out because they will be set automatically

components/elm/src/main/elm_varcon.F90
chloewhicker: Sorry @stephenprice I didn't know I needed to submit the firn model changes when we talked about it. Matt is right, this line is included in my PR here: https://github.com/E3SM-Project/E3SM/pull/6298#pullrequestreview-1958523608

components/elm/cime_config/testdefs/testmods_dirs/elm/gis20kmERP/user_nl_mali
stephenprice: I think for 1 and 2 above these should be the defaults. I don't know about 3. I could try the test w/ that set to 'temperature' as I'm guessing we may not want 'none' as the default. 

The other ones that I set in here are the following:

```
config_pio_num_iotasks = 47
config_pio_stride = 64
config_calving = 'none'
config_restore_calving_front = .true.
```

Based on the discussion w/ Jon, I can remove the PIO settings (?). But for calving, like thermal solver, I suspect we don't want these settings to be 'none' and .true. I could also re-try these tests w/ these settings not specified (so using whatever the defaults are). 







config_dt = '0000-00-01_00:00:00'

stephenprice: FYI, here are the current defaults for all of the params I update as part of these namelist changes:

```
config_velocity_solver = 'FO'    [when 'mali_use_albany' = .true.]
config_thickness_advection = 'fo'
config_thermal_solver = 'none'
config_pio_num_iotasks = 0
config_pio_stride = 1
config_calving = 'none'
config_restore_calving_front = .true.
config_dt = '0000-00-01_00:00:00'
```
So I think my tests should all work fine w/o any of the changes being discussed above. I will test that and confirm, in which case I can just remove these files.

stephenprice: They work w/ the defaults. I've removed these files and will push that change (along w/ one other) in a few mins. I'm going to resolve this portion of the discussion.

components/elm/cime_config/config_pes.xml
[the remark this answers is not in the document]
stephenprice: I've added a default for this that works (tested on Chyrs). I will push that momentarily so I'm going to resolve this for now.

cime_config/config_grids.xml
stephenprice: Chloe and Charlie suggested it might be nice to keep this grid around for a bit for backward compatibility, since it's what they used to spin up the SMB with. I think if we wanted to push to get rid of it they would be ok with that too.

cime_config/config_grids.xml
stephenprice: @jonbob -- if bi-grids really are a thing of the past then I'm happy to remove support for these configs (I will also check w/ Charlie and Chloe to make sure that's ok with them).

components/elm/src/main/elm_varcon.F90
stephenprice: Ok. Not sure I need to revert this commit though as it will resolve itself in the merge, right?

cime_config/config_grids.xml
stephenprice: Chloe and Charlie are apparently ok w/ removing bi-grid support (they have been using tri-grid configs for their work). So if the consensus is to remove bi-grid support, I will put that on my to-do list as well.

components/elm/src/main/elm_varcon.F90
matthewhoffman: #6298 has been merged already, so I think it is safe to leave this change here.  I looked a the git history, and this change was part of a commit with changes we want to keep, so dropping this change is not as simple as just removing a commit from the branch history.  I'm ok with leaving it in, unless @jonbob thinks it would be cleaner to remove it from the history or change it back.

cime_config/config_grids.xml
stephenprice: ... and, based on a discussion w/ Jon yesterday, I'm going to table (resolve) this for now. We decided to leave in for now and tackle trimming down the support to only tri-grids in a future PR.

cime_config/config_grids.xml
stephenprice: For the reasons noted above, I'm going to assume we keep this support in for a bit longer and resolve this discussion thread for now.

components/elm/bld/namelist_files/namelist_definition.xml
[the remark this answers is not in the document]
stephenprice: This has been updated to explicitly include the "R2" version of the gis1to10km mesh. The descriptions of the initial and R2 versions of the meshes have also been updated (w/ a note as to how they differ and why we are currently keeping support for both).

stephenprice:
@jonbob -- I think this is ready for another look. Please let me know if you need help testing anything, etc.

jonbob:
@rljacob -- could you please take a look at the changes we've made in config_pesall.xml to differentiate between B-cases and BG, where there's an active glc model? Also, this PR adds a e3sm_landice_developer test suite that probably needs to be part of e3sm_integration instead because Albany is available only on chrysalis and pm-cpu. I'll do some timing of how long the suite takes and post it here

trhille (approved):
This is great, @stephenprice! Looks to me like it's ready to go.

cime_config/tests.py
```diff
@@ -256,7 +263,7 @@
         },
 
     "e3sm_developer" : {
-        "inherit" : ("e3sm_land_developer", "e3sm_atm_developer", "e3sm_ice_developer"),
+        "inherit" : ("e3sm_land_developer", "e3sm_atm_developer", "e3sm_ice_developer", "e3sm_landice_developer"),
         "time"    : "0:45:00",
```
rljacob, line 267: Take this out of e3sm_developer for now since it has to run with gnu.

stephenprice: Done. Thanks.

jonbob:
Passes:
* ERS.f09_g16_g.MALISIA.chrysalis_intel
* SMS_D_Ld1.ne30pg2_r05_IcoswISC30E3r5.CRYO1850-DISMF.chrysalis_intel
* new e3sm_landice_developer suite on chrysalis with gnu compiler

verified all new files are in the correct format and available on the inputdata repo

merged to next

rljacob:
Re: changes to config_pesall.xml.   looks like you explicitly added SGLC to all the fully coupled cases except for the one where you actually run MALI.   I think thats a good way to make that distinction.

components/elm/src/main/elm_varcon.F90
rljacob: If it only affects cases with an active ice sheet model, its fine.

rljacob (approved):

stephenprice:
> Re: changes to config_pesall.xml. looks like you explicitly added SGLC to all the fully coupled cases except for the one where you actually run MALI. I think thats a good way to make that distinction.

Thanks. In fact, we could not figure out how to get the GLC-specific configs. (which are similar to some other configs in many ways) to be recognized without making this change. This was Jon W's suggested solution.

jonbob:
requested running the new e3sm_landice_developer test suite on chrysalis and pm-cpu using the gnu compiler, SES-3111

ndkeen:
This PR causes a build error with intel pm-cpu. I think we should revert and fix. Also, maybe ping the POC before making changes to machine files.

cime_config/machines/config_machines.xml
```diff
@@ -263,8 +263,9 @@
       <env name="PERL5LIB">/global/cfs/cdirs/e3sm/perl/lib/perl5-only-switch</env>
       <env name="FI_CXI_RX_MATCH_MODE">software</env>
       <env name="MPICH_COLL_SYNC">MPI_Bcast</env>
-      <env name="Albany_ROOT">$SHELL{if [ -z "$Albany_ROOT" ]; then echo /global/common/software/e3sm/mali_tpls/albany-e3sm-serial-release-gcc; else echo "$Albany_ROOT"; fi}</env>
+      <env name="Albany_ROOT">$SHELL{if [ -z "$Albany_ROOT" ]; then echo /global/common/software/e3sm/mali_tpls/albany-e3sm-serial-release-gcc-cmake-fix; else echo "$Albany_ROOT"; fi}</env>
```
mperego, line 266: Should we put these changes in the gnu environment_variables compiler like done in Chrysalis?
```
<environment_variables compiler="gnu">
   <env name="Albany_ROOT">$SHELL{if [ -z "$Albany_ROOT" ]; then echo /global/common/software/e3sm/mali_tpls/albany-e3sm-serial-release-gcc-cmake-fix; else echo "$Albany_ROOT"; fi}</env>
    <env name="Trilinos_ROOT">$SHELL{if [ -z "$Trilinos_ROOT" ]; then echo /global/common/software/e3sm/mali_tpls/trilinos-e3sm-serial-release-gcc; else echo "$Trilinos_ROOT"; fi}</env>
    <env name="Kokkos_ROOT">$SHELL{if [ -z "$Kokkos_ROOT" ]; then echo /global/common/software/e3sm/mali_tpls/trilinos-e3sm-serial-release-gcc; else echo "$Kokkos_ROOT"; fi}</env>
</environment_variables>
```

mperego:
> This PR causes a build error with intel pm-cpu. I think we should revert and fix. Also, maybe ping the POC before making changes to machine files.

Sorry for the inconvenience. I guess we should have put the pm-cpu changes to the machine file under the gnu environment, like we did with Chrysalis.

jonbob:
@mperego -- can you make that fix and either add it to this PR or make a separate PR? We don't want to mess up pm-cpu testing any longer than necessary

mperego:
@jonbob yes, I'll add the fix to this PR.

stephenprice:
@ndkeen -- we're working to test both Mauro's recent fix for this and will also re-test one of your failing intel tests as well. Will get back to you asap.

stephenprice:
@ndkeen -- with Mauro's fix above, `ERP_Ld3.ne4pg2_oQU480.F2010.pm-cpu_intel` passes.

ndkeen:
@stephenprice next still fails for me with `SMS_D_Ld1.ne30pg2_r05_IcoswISC30E3r5.CRYO1850.pm-cpu_intel`

jonbob:
@ndkeen -- I'm just re-merging to next right now. I'll let you know once it's there

jonbob:
Re-merged to next. @ndkeen -- we'd very much appreciate any testing you can do. With apologies

ndkeen:
I verified these two tests now build/run with next.
I have not looked more, but we can see what happens with testing tonight.
```
ERP_Ld3.ne4pg2_oQU480.F2010.pm-cpu_intel
SMS_D_Ld1.ne30pg2_r05_IcoswISC30E3r5.CRYO1850.pm-cpu_intel
```

jonbob:
merged to master and new tests blessed

PR #6202, opened 2024-02-02
Update resolutions and maps in test suites and v3 low res mesh

rljacob:
Update resolutions in test cases to replace ne4 and ne11 with ne4pg2,
replace oQU240 with oQU480 and update v2 mpas mesh with v3 IcoswISC30E3r5 in tests of the production config.  Also change those from bigrid to trigrid.

Update test length for some 2-case tests to work with new coupling frequencies.
Make non-linear maps default for the v3 low res mesh ne30pg2_r05_IcoswISC30E3r5 since v3 uses those in production.

------

At least one test was changed in all of the following test suites.  
- e3sm_integration
  -  e3sm_rrm
  -  e3sm_atm_integration
     - eam_preqx
     - eam_theta
  - e3sm_developer
    - e3sm_atm_developer
    - e3sm_ice_developer
    - e3sm_land_developer
      - e3sm_mosart_exenoshare
      - e3sm_land_debug
        
- e3sm_extra_coverage
  - e3sm_atm_extra_coverage
- eam_condidiag
- e3sm_atm_dustemis

PR #6231, opened 2024-02-11
add first patch of cime tests to public CI

mahf708:
Adds the first batch of 7 tests to run on public CI resources during PR submission.

Tests are SMS, SMS_D, ERS, ERP, REP, PET, PEM using F2010 and ne4pg2_oQU480. Tests are run with _P4 because CI resources are limited to 4 virtual CPUs for now.

[BFB]

### DIFF
--- a/.github/workflows/e3sm-gh-ci-cime-tests.yml
+++ b/.github/workflows/e3sm-gh-ci-cime-tests.yml
@@ -1,4 +1,4 @@
-name: test
+name: gh
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


jgfouca (approved):
Nice!

.github/workflows/e3sm-gh-ci-cime-tests.yml
```diff
@@ -0,0 +1,64 @@
...
+          - "REP"   # 0h30m
+          - "ERS"   # 1h00m
+          - "ERP"   # 1h15m
+          - "ERR"   # 1h00m
+          - "PET"   # 0h35m
+          - "PEM"   # 0h45m
+          - "SEQ"   # 3h10m
+          - "ERI"   # Exception from case_test: [Errno 2] No such file or directory:
+          - "ERT"   # Exception from case_test: ERROR: No variable AVG_HIST_N found in case
+          - "PEA"   # ERROR: BUILD FAIL: build e3sm failed
```
rljacob, line 29: please remove theses tests:  ERT (a longer ERS),  ERR (ERS with short-term acrchiving), ERI (hybrid run).  Also PEA (compares 1 task MPI with mpi-serial).  Also remove the SEQ test (have no idea why that's 3 hours but also needs a staggered layout for a fully coupled case to work).

rljacob:
Did you try a fully coupled case on these CI resources?

For artifacts to save, please add files ending in cprnc.out

rljacob:
I downloaded one of the archive files and most of it is what we want.  We don't need anything from bld/cmake-bld or the env_archive_test directory.

rljacob:
One more thing we need in the archive:  TestStatus.log which is the casedir.

jgfouca:
Yeah, the TestStatus.log is important. Also, I wonder if it would be easier to add an entry for singularity machine in config_pes.xml to just set the procs to 4 for all grids/compsets? That way you don't have to add `_P4` to all the tests.

rljacob:
Most of those passing integration tests are at resolutions that are going to change so we should wait until https://github.com/E3SM-Project/E3SM/pull/6202 is merged.  Then update the CI tests.  Just go with what you have.

rljacob:
@mahf708 I cleaned up the PR description (above the line).   You may want to edit your first-person story (below the line).

rljacob (approved):

PR #6232, opened 2024-02-12
Subglacial hydro boundary and channel cleanup

matthewhoffman:
This PR introduces a number of fixes and improvements to the subglacial hydrology model to get it to run more stably, mostly related to handling of boundaries of the subglacial hydrology domain:
* Adds a missing channel melt source term to the mass cons eqn.  This term is typically very small but was missing
* Make channels see the elevation head of lake thickness in the hydropotential to avoid channels flowing in to lakes
* Disable diffusive water flux at grounding line where water thickness is undefined
* Explicitly handle gradients at boundaries of the mesh where they are undefined
* Update channel CFL calculation to only consider valid edges
* zero the diffusivity at domain boundaries to prevent those edges from affecting diffusive CFL
* creation of new hydroTerrestrialMarginMask to be used for handling boundaries of the SGH domain
* other minor cleanup/adjustments


matthewhoffman:
Closing because opened in wrong repository

PR #6238, opened 2024-02-14
Adds a new test for the multi-instance capability

bishtgautam:
An I-compset test is added that sets up two instances of ELM and MOSART.

[BFB]

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -57,6 +57,7 @@ _TESTS = {
             "ERS_Ld20.f45_f45.IELMFATES.elm-fates",
             "ERS.hcru_hcru.I20TRGSWCNPRDCTCBC.elm-erosion",
             "ERS.f09_g16.IELMBC.elm-simple_decomp",
+            "ERS.hcru_hcru.IELM.elm-multi_inst",
             )
         },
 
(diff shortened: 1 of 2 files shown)


whannah1 (approved):

rljacob:
But just one instance of DATM?  That works?

bishtgautam:
All prognostic components need to have the same number of instances, but the data components have the flexibility to have a single or multiple instances. This is based on the Section "Multi-instance component functionality" described in https://www2.cesm.ucar.edu/models/cesm1.2/cesm/doc/usersguide/ug.pdf which has the following example:

>For an F compset, only atm, lnd, rof are full prognostic components. The
ocn is a prescribed data component, cice is a mixed prescribed/prognostic
component (ice-coverage is prescribed) and glc and wav are stub components.
Lets say we want to run 2 instances of CAM in this experiment. The current
implementation of multi-instances will also require you to run 2 instances of
CLM, CICE and RTM. However, you have the flexibility to run either 1 or 2
instances of DOCN (we can ignore glc and wav since they do not do anything
in this compset).

rljacob (approved):

peterdschwartz:
merged to next

peterdschwartz:
merged to master.

Note:  pm-cpu has yet to report and may not today for this test.

PR #6248, opened 2024-02-20
CPL: Robustly recover from failures during restart-file writing.

ambrad:
It's possible for a crash that occurs during restart writing to lead to inconsistent or incomplete rpointer files. While we can't salvage the restart files in general, we can at least provide a consistent set of rpointer files -- namely, the previous ones -- for the next restart.

The rpointer manager provides this capability by copying all rpointer.X files to rpointer.X.prev before components write restart files, then removing rpointer.X.prev files when all components are done.

If a crash occurs midway through, on restart, the consistent rpointer.X.prev files will be used.

[BFB]

### DIFF
--- a/driver-mct/main/cime_comp_mod.F90
+++ b/driver-mct/main/cime_comp_mod.F90
@@ -2759,6 +2759,10 @@ contains
        ! Does the driver need to pause?
        drv_pause = pause_alarm .and. seq_timemgr_pause_component_active(drv_index)
 
+       ! Monitor each component's restart alarm. Determine when to prepare
+       ! backup copies of the rpointer files and when it's OK to remove
+       ! these. .false. on input means never to force removal of the backup
+       ! files until the monitor's state machine says they can be removed.
        call rpointer_manage(.false.)
 
        if (glc_prognostic .or. do_hist_l2x1yrg) then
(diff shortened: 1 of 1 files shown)


ambrad:
I've tested this feature for efficacy, but I still need to do comprehensive integration testing. I will post the results here when I've done that.

driver-mct/main/cime_comp_mod.F90
```diff
@@ -5125,4 +5171,411 @@ subroutine cime_write_performance_checkpoint(output_ckpt, ckpt_filename, &
...
+                  exist=file_exists)
+             if (.not. file_exists) then
+                complete = .false.
+                write(logunit,'(3a)') 'rpointer> ERROR: ', rpointer_suffixes(i), &
+                     ' has an rpointer.x file with no corresponding rpointer.x.prev file'
+             end if
+          end if
+       end do
+       if (.not. complete) then
+          call shr_sys_abort('rpointer_manage: rpointer.x.prev files exist but rpointer.y &
```
rljacob, line 5240: Should this abort message say "rpointer_prepare_restart" instead of "rpointer_manage" ?

driver-mct/main/cime_comp_mod.F90
```diff
@@ -2717,6 +2759,8 @@ subroutine cime_run()
        ! Does the driver need to pause?
        drv_pause = pause_alarm .and. seq_timemgr_pause_component_active(drv_index)
 
+       call rpointer_manage(.false.)
```
rljacob, line 2762: can you add a comment about what this call is doing?

driver-mct/main/cime_comp_mod.F90
```diff
@@ -3544,6 +3588,8 @@ subroutine cime_final()
     call component_final(EClock_l, lnd, lnd_final)
     call component_final(EClock_a, atm, atm_final)
 
+    call rpointer_manage(.true.)
```
rljacob, line 3591: Can you add a comment about what this call is doing?

rljacob:
Thanks @ambrad.  Merge whenever you're ready.

PR #6251, opened 2024-02-22
Smooth hydro fields

alexolinhager:
This PR introduces two smoothed fields that help keep the hydro model stable when run with spatially variable bed topography and ice thickness. The first is `iceThicknessHydro`, which is a modified version of `thickness` that is only used by the hydrology model in place of the original `thickness` field. This PR moves through multiple iterations of `iceThicknessHydro`; however, the ultimate version is nearly identical to `thickness` but with all local minima and maxima removed. Local minima/maxima are defined as a single cell that has a thinner/thicker `thickness` than all of its immediate neighbors. Local minima/maxima are then removed by setting `iceThicknessHydro` at that cell equal to the mean of its neighbors. `iceThicknessHydro` is necessary to avoid channel and diffusivity instabilities that arise from single-cell minima and maxima. A new output variable, `upperSurfaceHydro`, is then introduced as the sum of `bedTopography` and `iceThicknessHydro`. The use of `iceThicknessHydro` can be toggled on and off with namelist option `config_SGH_use_iceThicknessHydro`.

The second change is the introduction of `waterPressureSmooth`, which is a smoothed version of `waterPressure` that is only used for calculation of `waterPressureSlopeNormal`. This updated version of `waterPressureSlopeNormal` has more gradual gradients that help keep `channelPressureFreeze` stable. However, `channelPressureFreeze` is not yet fully functional due to additional necessary improvements to marine boundary conditions in realistic geometries. This will be addressed in the next pull request.


jonbob:
@alexolinhager -- before you get too far, let me point you at E3SM's branch naming conventions: [Branch Tag and Version name conventions](https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/2523172/Branch+Tag+and+Version+name+conventions)

matthewhoffman:
@alexolinhager  moved this PR to [MALI-Dev/E3SM ](https://github.com/MALI-Dev/E3SM). We'll coordinate these changes with other MALI changes into E3SM master at a later point.

PR #6254, opened 2024-02-23
Single Column Model Updates & Fixes for v3

bogensch:
Modify and fix the single column model (SCM) to make it compatible with master and v3.  Specifically: 

1. The config_component for SCM_EAM has been changed to “scam_generic” (which should have never been changed).  When running cases with the standardized scripts prescribed aerosols are used, which is not compatible with the scm_generic_chemUCI-Linoz-mam5-vbs component currently specified.  (Aside: a new prescribed aerosol file has been generated by running E3SM with mam5.  This file has been uploaded to the E3SM data server and the E3SM SCM scripts have been updated to use this file).
2. Idealization flags have been added to the P3 microphysics to allow for:
    1. Runs with no precipitation.  Note that in the P3 implementation this only turns off liquid precipitation, which is used in several GCSS boundary layer cloud cases in the E3SM SCM library.  
    2. Prescribed droplet concentration.  This was already a feature in place in P3 but this PR connects this functionality to the EAM namelist.
3. Fixes to allow the ability to “replay” a column once again (back by semi-popular demand!).  This was a feature that worked in E3SMv1 but has not been functional for sometime.  The fix in this PR allows users to generate output needed to make IOP forcing files to replay a single column.  Note that additional post-processing is required to actually do an SCM run (in which the user needs to consult with the E3SM wiki page).  

Further documentation and scripts have been updated on the E3SM SCM wiki to reflect changes made in this PR (https://github.com/E3SM-Project/scmlib/wiki/E3SM-Single-Column-Model-Home).

[BFB]

All e3sm_developer tests pass on pm-cpu.

### DIFF
--- a/components/eam/cime_config/config_component.xml
+++ b/components/eam/cime_config/config_component.xml
@@ -136,7 +136,7 @@
       <value compset="20TR(?:SOI)?_EAM.*CHEMUCI.*LINOZ.*MAM5"  >20TR_eam_chemUCI-Linoz-mam5</value>
       <value compset="ARM95_EAM"                        >scam_arm95</value>
       <value compset="ARM97_EAM"                        >scm_arm97_chemUCI-Linoz-mam5-vbs</value>
-      <value compset="SCM_EAM"                          >scm_generic_chemUCI-Linoz-mam5-vbs</value>
+      <value compset="SCM_EAM"                          >scam_generic</value>
       <value compset="PIPD_EAM"                         >1850-PD_cam5</value>
       <value compset="_EAM%AQP_"                        >aquaplanet_EAMv1</value>
       <value compset="_EAM%RCE_"                        >RCEMIP_EAMv1</value>
--- a/components/eam/bld/namelist_files/namelist_defaults_eam.xml
+++ b/components/eam/bld/namelist_files/namelist_defaults_eam.xml
@@ -917,6 +917,7 @@
 <cld_sed              microphys="mg2">  1.0D0       </cld_sed>
 <mg_prc_coeff_fix     microphys="mg2">  .false.     </mg_prc_coeff_fix>
 <micro_nccons         microphys="mg2">  100.D6      </micro_nccons>
+<micro_nccons         microphys="p3">   200.D6      </micro_nccons>
 <micro_nicons         microphys="mg2">  0.1D6       </micro_nicons>
 <micro_mincdnc        microphys="mg2">  -999.       </micro_mincdnc>
 <p3_mincdnc           microphys="p3">  -999.        </p3_mincdnc> 
--- a/components/eam/bld/namelist_files/namelist_definition.xml
+++ b/components/eam/bld/namelist_files/namelist_definition.xml
@@ -4706,7 +4706,9 @@ Default: FALSE
 
 <entry id="precip_off" type="logical" category="scm"
        group="cam_inparm" valid_values="">
-Turn off microphysics computation.
+Turn off microphysics computation for MG2.  For P3 this disables the generation of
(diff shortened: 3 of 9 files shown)


rljacob:
@brhillman please review.

rljacob:
@brhillman please review this

brhillman (approved):

components/eam/cime_config/config_component.xml
```diff
@@ -136,7 +136,7 @@
       <value compset="20TR(?:SOI)?_EAM.*CHEMUCI.*LINOZ.*MAM5"  >20TR_eam_chemUCI-Linoz-mam5</value>
       <value compset="ARM95_EAM"                        >scam_arm95</value>
       <value compset="ARM97_EAM"                        >scm_arm97_chemUCI-Linoz-mam5-vbs</value>
-      <value compset="SCM_EAM"                          >scm_generic_chemUCI-Linoz-mam5-vbs</value>
+      <value compset="SCM_EAM"                          >scam_generic</value>
```
rljacob, line 139: @bogensch should we add a test of this compset?  We only test ARM97_EAM

bogensch: Yes, it's probably wise.  I'll try to get this done before I leave for vacation tomorrow afternoon... else expect a period of radio silence (FYI).

wlin7: Will need a new testmod for it, to bring in some of the settings specified in SCM case scripts to allow setup to proceed. (e.g., scmlat, scmlon)

wlin7:
Hi @bogensch , thanks for working to update the SCM configuration. I didn't realize when putting in v3atm that SCM run scripts rely on the generic use_case file. I modified the use_case for the nightly ARM97 test, and without careful thinking, also applied the changes to the generic one. As Rob suggested, a test using scam_generic will be useful for protecting this configuration. It can replace the other test (FSCM-ARM97) or revert the change in its use_case file and associated config option to be more like scam_generic.

Unrelated to this PR itself: I found the link to the 'desired cases' on the instruction for '[Running the single column model](https://github.com/E3SM-Project/scmlib/wiki/Running-the-E3SM-Single-Column-Model)' returns a page that I didn't expect: It 
 would be to Create a new page. I am not sure if this is intended and how it works if so.
 
 Also have a few comments on the run script, if you have time to update them in the near future.
- Can replace machine cori with pm-cpu, and $CSCRATCH with $PSCRATCH
- [Line 198 of the runscript](https://github.com/E3SM-Project/scmlib/blob/master/E3SM_SCM_scripts/run_e3sm_scm_ARM97.csh#L198 ) has a duplicated keyword 'set'
- a number of if-logics are conditioned on if the code version is 'master' . Users may not fully comprehend the distinction you try to convey. Maybe some like 'v3x' to use either master or any variant branch a user creates?
- namelist parameter values in the script going to user_nl_eam. Right now there is no distinction between v2 and v3.  Most of them should have been set by default, given v2 or v3 code base. Can we remove those that are not SCM specific?

wlin7 (approved):
The PR itself looks good. Verified atm_developer tests remain BFB.

rljacob:
@wlin7 you can start merging this if you want.  We could wait and add the test in a later PR.

bogensch:
@wlin7 thanks for your careful review... not only of the code but for the scripts.  I will incorporate your great suggestions to the scripts.

wlin7:
Thank you, @bogensch . As the scripts are not part of this PR, I can start merging today if next is open, while you work on updating the run scripts.

wlin7:
Merged to next.

PR #6258, opened 2024-02-25
Changing clobber mode in ocean conservation analysis

darincomeau:
Previously the clobber mode in the ocean conservation analysis member was set to 'append', which would throw errors to `log.ocean.*.err` files when rerunning a case:

```
ERROR: Writing to stream 'conservationCheckOutput' would overwrite record    1 in file 'SMS_Lm1_P512.ne30pg2_r05_IcoswISC30E3r5.CRYO1850.chrysalis_intel.20240225_133103_09nrwo.mpaso.hist.am.conservationCheck.0001-02-01.nc',
ERROR:     but clobber_mode is set to 'append'.
```

Changing clobber mode to `truncate` avoids these errors.

[BFB]


jonbob:
passes:

merged to next

jonbob:
merged to master

PR #6260, opened 2024-02-26
Fix ncol/pcols mismatch on interpolating aerosol mass fields to pressure levels

### LINKED ISSUE #6259: ne120pg2 run on pm-cpu failed on writing aerosol mass at specified levels to eam.h0
This occur at end of a month trying to write variables such as Mass_bc_850, Mass_bc_500 to eam.h0 file.
The error message is like 
```
 FATAL ERROR: Aborting... An error occured, Writing variables (number of variables = 184) to file ...
 Non blocking write for variable (Mass_bc_850, varid=146) failed ....
 NetCDF: Numeric conversion not representable (err=-60).
```
Typically such numeric conversion error occur when the output field contains NaN values.

This issue currently only occur on pm-cpu with ne120pg2 run and with history_aerosols = .true. Same issue didn't exist on chrysalis, or pm-cpu at ne30pg2 resolution. 

Further investigation found the 3D aerosol mass arrays are good. Only the vertically interpolated values to specified levels have NaN. It is further determined it is due to mismatch of ncol and pcols for arrays when calling vertinterp. For this reason, even although ne30pg2 run or ne120pg2 run on a machine didn't fail, the mismatch is expected to also result in corrupted values in the constant pressure level aerosol mass fields (in those cases, likely 0s or other random but not NaN values outside the array memory bound were used for interpolation)

The fix is forthcoming.

wlin7:
The input dimension size specified in call vertinterp to interpolate aerosol masses to
pressure levels can be larger than the actual mass array (pcols >= ncol). This can cause
unintended values, either from different levels or random values outside the memory 
space associated with the array being used for interpolation, resulting in NaN values for
some configuration or on some machines. Even though not resulting in NaN values for a run,
such mismatch is also expected to cause the interpolated arrays being corrupted.

Fixes #6259.

[BFB] Differences only for diagnostic fields such as mass_bc_850 in eam.h0, if they are saved for a test

### DIFF
--- a/components/eam/src/chemistry/mozart/mo_chm_diags.F90
+++ b/components/eam/src/chemistry/mozart/mo_chm_diags.F90
@@ -1488,16 +1488,16 @@ contains
           
           call outfld( 'Mass_'//trim(aerosol_name(aerosol_idx))//'_srf',  mass_3d_tmp(:ncol,pver), ncol, lchnk )
 
-          call vertinterp(ncol, pcols, pver, pmid,  85000._r8, mass_3d_tmp, mass_at_pressure)
+          call vertinterp(ncol, ncol, pver, pmid(:ncol,:),  85000._r8, mass_3d_tmp, mass_at_pressure)
           call outfld( 'Mass_'//trim(aerosol_name(aerosol_idx))//'_850',  mass_at_pressure, ncol, lchnk )
 
-          call vertinterp(ncol, pcols, pver, pmid,  50000._r8, mass_3d_tmp, mass_at_pressure)
+          call vertinterp(ncol, ncol, pver, pmid(:ncol,:),  50000._r8, mass_3d_tmp, mass_at_pressure)
           call outfld( 'Mass_'//trim(aerosol_name(aerosol_idx))//'_500',  mass_at_pressure, ncol, lchnk )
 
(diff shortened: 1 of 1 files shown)


wlin7:
Thanks for the quick review, @mahf708 . Those other interpolations are ok. Take the following as an example
`call vertinterp(ncol, pcols, pver, state%pmid, 80000._r8, state%t, p_surf)`

The 2nd argument specify the column dimension sizes for `state%pmid` and `stat%t`, and they are both declared with pcols. The actual calculation loops over the 1st dimension, i.e., `ncol` in this call.

For the aerosol mass arrays, such as `mass_bc` and the tmp array `mass_3d_tmp`, they are declared as `(ncol,pver)`, need to use `ncol `for the 2nd argument and the explicit scoping for `pmid` to ensure the match. Dummy argument `ncold` in subroutine vertinterp dictate all these.

wlin7:
> Ah, I see! Thanks! I was obviously lazy to figure out all the nuances here, and I will do better next time (or I should manage to learn enough f90) before I edit any substantial code 😸

That is fine. We all reuse codes often by copying/pasting. Everybody could miss something like this sometimes.

wlin7:
Merged to next.

Issue #6209, opened 2024-02-05
no compsets that enable marine BGC with JRA forcing.


maltrud:
All of the GMPAS compsets are defined to work with CORE2 (IAF) forcing, not JRA.  Since we almost exclusively use JRA forcing now, these need to be updated to support HES objectives.

Issue #6208, opened 2024-02-05
no compset that enables marine BGC with active ocean, ice, land, river components.


maltrud:
The GPMPAS-JRA compset was created allow simulations using active ocean, ice, land, river components.  However, there are no GPMPAS compsets that enable marine BGC (ie, analogs to GMPAS-OECO-IAF and GMPAS-OIECO-IAF).  These need to be created to support HES objectives.

rljacob:
What about this compset that we test?     
BGCEXP_CNTL_CNPRDCTC_1850
1850_EAM%CMIP6_ELM%CNPRDCTCBC_MPASSI%BGC_MPASO%OIECO_MOSART_SGLC_SWAV_BGC%BCRC

maltrud:
@rljacob I am only adding G-case compsets for now.

maltrud:
@rljacob , @jonbob I'm going to make a PR soon with new compsets.  FYI, for new mapping files, I'm using the Tempest versions.  for old mapping files, I left "as-is" since Jon will be replacing them all soon.  I hope that sounds ok.

rljacob:
Just be sure to following the map name convention for the Tempest maps.

maltrud:
Jon created the maps so I assume they have the correct convention.

PR #6262, opened 2024-02-26
define new GMPAS, GPMAS, ECO compsets with JRA

maltrud:
Define new compset aliases that support marine BGC using JRA1.4 forcing.  Also allow GPMPAS configurations (with active ELM and MOSART) to work with WC14to60E2r3, EC30to60E2r2, and ARRM10to60E2r1 grids.

Fixes [#6209](https://github.com/E3SM-Project/E3SM/issues/6209) 

Fixes [#6208](https://github.com/E3SM-Project/E3SM/issues/6208)

[BFB]

### DIFF
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -489,6 +489,36 @@
       <mask>WC14to60E2r3</mask>
     </model_grid>
 
+    <model_grid alias="TL319_r05_WC14to60E2r3" compset="(DATM|XATM|SATM)">
+      <grid name="atm">TL319</grid>
+      <grid name="lnd">r05</grid>
+      <grid name="ocnice">WC14to60E2r3</grid>
+      <grid name="rof">r05</grid>
+      <grid name="glc">null</grid>
+      <grid name="wav">null</grid>
+      <mask>WC14to60E2r3</mask>
+    </model_grid>
+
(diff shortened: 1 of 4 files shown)


rljacob:
notes:  Jon will work on.  Depends on conversation on supporting lots of compsets in bgc.

jonbob:
@maltrud -- is there a reason to add aliases for all three flavors of JRA forcing? I think there will be a push for us to have more like representative compset aliases with the understanding that a more full list is accessible through compset long-names. Of course, it's also fine to just merge this as-is and think about changing which ones we keep in a later compset alias reduction PR

maltrud:
@jonbob yes I'm fine with reducing the number of compsets with aliases--it is already getting out of hand.  just one flavor of JRA, and also (maybe?) no need for OECO.

jonbob:
@maltrud -- OK, I just left the JRA1p4 versions of the new compsets, since JRA1p5 has issues right now. We can update that later if it becomes preferable

jonbob (approved):
Verified that all new files are available on the inputdata repo, and passes sanity testing.

jonbob:
merged to next

rljacob:
Are all the files in the right flavor of NetCDF?

jonbob:
@rljacob -- I'll double-check that and fix them if not

jonbob:
Thanks @rljacob -- good catch. They're converted and checked on the inputdata server as well as the local lcrc inputdata location.

merged to master

PR #6268, opened 2024-02-27
Modify how dimensions ids are set in coupler history

rljacob:
The coupler history was not using the dimension id of coordinate variables when defining dimensions of variables for most of the attribute vectors. This made the coupler history hard to use with viz/analysis tools like ncvis. Allow dimension ids of coordinate variables to be passed back to calling function and then used in future calls that define additional variables.  Coupler history can then be viewed with ncvis and possibly other tools.

[BFB]  doesn't change content of variables in a coupler history.

---------
This ncdump illustrates the problem.
```
dimensions:
	time = UNLIMITED ; // (1 currently)
	doma_nx = 384 ;
	doma_ny = 1 ;
variables:
	double doma_lat(time, doma_ny, doma_nx) ;
		doma_lat:long_name = "latitude" ;
	double doma_lon(time, doma_ny, doma_nx) ;
		doma_lon:long_name = "longitude" ;
	double x2a_Sx_tref(time, x2a_ny, x2a_nx) ;
		x2a_Sx_tref:long_name = "Reference temperature at 2 meters" ;
```

Since x2a_Sx_tref doesn't have the same lat, lon dimension ids of the variables with the actual latitude and longitude (doma_ny vs. x2a_ny), most netcdf vis programs won't know how to display x2a_Sx_tref.

The fix changes the dimension ids for the variable to be:
`double x2a_Sx_tref(time, doma_ny, doma_nx) ;`

And similar for all other variables in the coupler history (which has variables on every mesh).

### DIFF
--- a/driver-mct/main/seq_hist_mod.F90
+++ b/driver-mct/main/seq_hist_mod.F90
@@ -161,6 +161,8 @@ contains
     integer(IN)   :: start_tod    ! Starting time-of-day (s)
     real(r8)      :: curr_time    ! Time interval since reference time
     integer(IN)   :: fk           ! index
+    integer(IN)   :: latlonid(2)  ! ids of lat lon dimensions
+    integer(IN)   :: alatlonid(2)  ! ids of atmosphere lat lon dimensions
     character(CL) :: time_units   ! units of time variable
     character(CL) :: calendar     ! calendar type
     character(CL) :: case_name    ! case name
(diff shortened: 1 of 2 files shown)


jonbob (approved):
Approved based on a visual inspection -- and looks like a very important change

jonbob:
@rljacob -- I approved just from looking at the code, but please let me know if it also needs to be tested

rljacob:
@jonbob can you try it with an ice sheet compset?  I just did watercycle ne4pg2.

jonbob:
Sure -- anything I should look for?

rljacob:
Make whatever mods you need to output a coupler history.  Or run one of the MALI tests which will do it.  Just want to make sure it writes without error when atmosphere is stub.  You could try looking at the glacier coupler fields with ncvis.

jonbob:
I ran SMS.T62_oQU120_ais20.MPAS_LISIO_TEST.chrysalis_intel successfully and the cpl history file could be visualized with ncvis, including the g2x and x2g fields

jonbob:
ncvis was unhappy with some fields, like x2a_So_t (I think) -- it complained
````
EXCEPTION (wxImagePanel.cpp, Line 651) Y (latitude) dimension has nonpositive range [-88.541950,-88.541950]
````
but I don't think that's related to this PR

rljacob:
Agreed. That's either an ncvis problem or different issue with how the domains are initialized when the component is DATM.

rljacob:
merged to next.

PR #6269, opened 2024-02-28
For pm-cpu (as well as gcp12, muller-cpu) use 2 nodes instead of 1 for conusx4v1pg2_r05_IcoswISC30E3r5.F2010

ndkeen:
For pm-cpu (as well as gcp12, muller-cpu) use 2 nodes instead of 1 for the F case using new resolution: `conusx4v1pg2_r05_IcoswISC30E3r5.F2010`

The cdash test `SMS_D_Ln5.conusx4v1pg2_r05_IcoswISC30E3r5.F2010`
was sometimes hitting OOM and also could use speed improvement

[bfb]

### DIFF
--- a/components/eam/cime_config/config_pes.xml
+++ b/components/eam/cime_config/config_pes.xml
@@ -1106,6 +1106,19 @@
         </ntasks>
       </pes>
     </mach>
+    <mach name="pm-cpu|muller-cpu|alvarez|gcp12">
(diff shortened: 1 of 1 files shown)


rljacob (approved):

ndkeen:
merged to next

ndkeen:
OK, this test on cdash looks good on both pm-cpu and gcp12, merging to master

PR #6270, opened 2024-02-28
Add FRIS meshes

irenavankova:
Meshes with high resolution in the southern Weddell Sea and beneath the Filchner Ronne Ice Shelf (FRIS). The meshes go down to 1, 2, 4, and 8 km beneath FRIS and on the continental shelf in front of it.

Meshes are:
`FRISwISC01to60E3r1`
`FRISwISC02to60E3r1`
`FRISwISC04to60E3r1`
`FRISwISC08to60E3r1`

They are intended only for use in E3SM G-case simulations.  Mapping and domain files for supporting B-cases or other coupled configurations are not included and will not be supported.

https://github.com/E3SM-Ocean-Discussion/E3SM/pull/75

### DIFF
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -2444,14 +2444,14 @@
       <file grid="ice|ocn" mask="WCAtl12to45E2r4">$DIN_LOC_ROOT/share/domains/domain.ocn.TL319_WCAtl12to45E2r4.210318.nc</file>
       <file grid="atm|lnd" mask="SOwISC12to60E2r4">$DIN_LOC_ROOT/share/domains/domain.lnd.TL319_SOwISC12to60E2r4.210119.nc</file>
       <file grid="ice|ocn" mask="SOwISC12to60E2r4">$DIN_LOC_ROOT/share/domains/domain.ocn.TL319_SOwISC12to60E2r4.210119.nc</file>
-      <file grid="atm|lnd" mask="FRISwISC08to60E3r1">$DIN_LOC_ROOT/share/domains/domain.lnd.TL319_FRISwISC08to60E3r1.230926.nc</file>
-      <file grid="ice|ocn" mask="FRISwISC08to60E3r1">$DIN_LOC_ROOT/share/domains/domain.ocn.TL319_FRISwISC08to60E3r1.230926.nc</file>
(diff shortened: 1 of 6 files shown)


xylar:
@irenavankova, do you have any links to analysis you have run for simulations with these meshes?  I know they aren't with this exact version of the branch but that would demonstrate that the meshes work as expected.

xylar (approved):
I ran 5-day smoke tests on the coarsest 2 meshes and shorter tests on the 2 finer meshes on Chrysalis.

I am also approving based on test results for each mesh that I have seen from @irenavankova over the past year.

irenavankova:
> @irenavankova, do you have any links to analysis you have run for simulations with these meshes? I know they aren't with this exact version of the branch but that would demonstrate that the meshes work as expected.

No, as it is all run on chicoma there isn't web access, but I could put some on Anvil if you like?

xylar:
That's up to you but I think it wouldn't hurt to have a few example results from these meshes if it's not too much trouble.

irenavankova:
Here is an example of MPAS Analysis from the 4 km mesh, years 9-10, Gcase:

https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.vankova/analysis/F4_spinup/clim_9-10_ts_1-10/html/ocean/index.html

xylar:
Very exciting, thanks @irenavankova!  The enhanced eddy activity from the higher resolution can be seen in:
image
and the ability to better resolve topography is also visible in transects like:
image

rljacob:
notes:  needs netcdf3 conversion for some mapping files.

xylar:
@jonbob and @rljacob, which mapping files need conversion?

jonbob:
@xylar -- a lot of the ones I had made with my script. It turns out the tempest default output is not what e3sm wants to use. I was going to fix them today and be ready to merge this PR.

xylar:
Okay, great!  I'm sorry for the extra work for you, though.

jonbob:
The mapping files are now converted and Ill work on merging this as soon as I can

jonbob:
confirmed all new files are in the correct format and on the inputdata server. Passes:

merged to next

jonbob:
merged to master

PR #6273, opened 2024-03-01
Writes/reads C budget related variables in history file

bishtgautam:
The `TCS_MONTH_BEGIN`  and `TCS_MONTH_END` variables are written and read from the ELM restart file. 

[non-BFB] just for those variables.  No change in answers otherwise.

### DIFF
--- a/components/elm/src/biogeochem/CNPBudgetMod.F90
+++ b/components/elm/src/biogeochem/CNPBudgetMod.F90
@@ -1193,8 +1193,7 @@ contains
     type(gridcell_carbon_state), intent(inout) :: grc_cs
     !
     ! !LOCAL VARIABLES:
-    integer :: year_prev, month_prev, day_prev, sec_prev
-    integer :: year_curr, month_curr, day_curr, sec_curr
+    integer :: year, month, day, sec
     !-----------------------------------------------------------------------
 
     associate(                                                       &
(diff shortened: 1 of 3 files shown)


rljacob (approved):

peterdschwartz:
This typo wasn't fixed completely.  As I mentioned in the slack discussion, this will cause FILLDIFFs with `TCS_MONTH_END` instead, and so it also needs to be added to the restart file the same as Gautam did for TCS_MONTH_BEG.  I have a commit with this that I pushed.

In addition, the time check for the end of the month is not correct:
`if (get_nstep() > 1 .and. day_prev == 1 .and. sec_prev == 0) then`  
This is just the same check used for the beginning of the month.  I changed this to use the current time information instead, which makes it consistent with how the water budget is handled.

I assume (not been tested yet) that this PR will cause DIFFs for any tests that compare elm history files as TCS_MONTH_END is going to be calculated differently. But it's past 1am here so I'll run comparions tomorrow

peterdschwartz:
From the `e3sm_integration_next_intel` suite on chrysalis, I expect diffs for:

```
  SMS_Ln9.ne4pg2_oQU480.F2010.chrysalis_intel.eam-outfrq9s (Overall: DIFF) details:
  SMS_Ly2_P1x1.1x1_smallvilleIA.IELMCNCROP.chrysalis_intel.elm-force_netcdf_pio (Overall: DIFF) details:
  SMS_Ly2_P1x1_D.1x1_smallvilleIA.IELMCNCROP.chrysalis_intel.elm-lulcc_sville (Overall: DIFF) details:
```

peterdschwartz:
For the record, tests that run for a month will compare elm.h0 files by default.

PR #6277, opened 2024-03-04
update circleci image spec due to upstream changes

mahf708:
Addresses CircleCI new image requirements, lest we are affected by brownouts and eventual blackouts.

--

We expreienced our first brownout in https://app.circleci.com/pipelines/github/E3SM-Project/E3SM/10529/workflows/8d2afe3d-53a7-456d-ae39-ecf6378f23f9/jobs/10059 (as part of PR #6276). See the note by CircleCI here: https://discuss.circleci.com/t/linux-image-deprecations-and-eol-for-2024/50177. The fix in this PR is simply putting a band-aid on our CircleCI testing until we have some time for a proper fix and for a thorough upgrade of our containers.

PR #5760, opened 2023-06-13
Add FATES land use change module to pass LUH2 data to FATES

glemieux:
This pull request adds a new module, the structure of which is adopted from the dyn_subgrid code, 
to enable e3sm to ingest minimally processed luh2 data to be passed to FATES.  
A new namelist variable is introduced `use_fates_luh`, to engage the module functionality, 
which is off by default.  The luh2 dataset to be ingested is a minimally processed concatenation 
of the luh2 state, transition and management datasets.

These changes also include a minor bug fix discovered in the process of developing the code, 
in which if the number of FATES patches being defined by the user in the fates parameter file is greater than the `max_patch_per_col` elm variable, would result in a BalanceCheck error.  

To be coordinated with https://github.com/NGEET/fates/pull/1040

[nonBFB] for FATES

PR #6279, opened 2024-03-04
fates api 33 compatibility, two-stream radiation 

rgknox:
This set of changes enables compatibility with FATES API 33, which brings in two-stream radiation for vegetation canopies.


rgknox:
This is based off of  https://github.com/glemieux/E3SM/tree/lnd/fates-luh2 which is coming in via #5760

rgknox:
Not sure why the circleci build fails, I ran a test on perlmutter and it worked fine:
./create_test SMS_Ld20.f45_f45.IELMFATES.pm-cpu_gnu.elm-fates

I'll go ahead run the land developer test suite and report in. @glemieux and I also have to decide on new tests, if any to add in with this.

peterdschwartz:
@rgknox The circleCI fail is due to a circleci issue.  The test that failed to build wasn't even an E3SM test case as far as I can tell. I'm guessing it's the same issue to be addressed here https://github.com/E3SM-Project/E3SM/pull/6277 .  So I wouldn't stress about it.

glemieux:
To do list (assigned to me):
- [ ] Add two stream `SMS` test

glemieux:
Kicking off `e3sm_land_developer` tests.

glemieux:
Regression testing with `e3sm_land_developer` against master baseline results in all expected tests passing b4b.  The fates testmods results in some diffs, which is expected given that the associated fates tag update includes a number of bug fixes and science updates.

rljacob:
@glemieux @peterdschwartz is this ready to merge?

peterdschwartz:
@rljacob If the code changes are done, then it would at least need a review from @bishtgautam

bishtgautam:
@glemieux @rgknox is this WIP or ready for review?

rgknox:
@rljacob , I believe it is, removing the label

components/elm/cime_config/testdefs/testmods_dirs/elm/fates_cold_twostream/shell_commands
```diff
@@ -0,0 +1,15 @@
...
+spack load nco
+
+SRCDIR=`./xmlquery SRCROOT --value`
+CASEDIR=`./xmlquery CASEROOT --value`
+FATESDIR=$SRCDIR/components/elm/src/external_models/fates
+FATESPARAMFILE=$CASEDIR/fates_params_twostream.nc
+
+ncgen -o $FATESPARAMFILE $FATESDIR/parameter_files/fates_params_default.cdl
+
+$FATESDIR/tools/modify_fates_paramfile.py --O --fin $FATESPARAMFILE --fout $FATESPARAMFILE --var fates_rad_model --val 2 --allpfts
```
bishtgautam, line 12: Why is the parameter file generated on the fly by `ncgen` and then modified, instead of having the file added to the inutdeck server?

components/elm/src/main/elmfates_interfaceMod.F90
```diff
@@ -507,11 +516,23 @@ subroutine ELMFatesGlobals2()
            pass_lu_harvest = 0
            pass_num_lu_harvest_types = 0
         end if
-
         call set_fates_ctrlparms('use_lu_harvest',ival=pass_lu_harvest)
         call set_fates_ctrlparms('num_lu_harvest_cats',ival=pass_num_lu_harvest_types)
         call set_fates_ctrlparms('use_logging',ival=pass_logging)
 
+        if(use_fates_luh) then
```
bishtgautam, line 523: Please add a whitespace `if(use_fates_luh) then` --> `if (use_fates_luh) then`

components/elm/cime_config/testdefs/testmods_dirs/elm/fates_cold_twostream/shell_commands
rgknox: We update the parameter file frequently in FATES. This prevents us from having to make changes to multiple test files every time we change an irrelevant parameter (which would also be a liability to injecting bugs.. ie what if we don't update all the files correctly every time there is a change).

Note that in the future, it is intended that we will have many of these on-the-fly parameter files generated, so that we can stress test the model with many different configurations.  It would not be a matter of updating 1 or two files, but think more like 5-10.

components/elm/src/main/elmfates_interfaceMod.F90
rgknox: added white-space via 801afee

bishtgautam (approved):

glemieux:
> @glemieux @peterdschwartz is this ready to merge?

Apologies for the delayed response.  I'm currently rerunning tests after correcting an issue with the `fates_cold_allvars` testmod in the e3sm land developer list.

glemieux:
Also I need to add the temporary fix for https://github.com/E3SM-Project/E3SM/issues/6316

peterdschwartz:
merged to next

PR #6280, opened 2024-03-04
Update top level doc files for v3

rljacob:
Update README and CITATION.cff for v3.0.0
Include the newly minted software DOI

[BFB]

--------
Note: the DOI won't be active for another day or so.


wlin7 (approved):
Approved given the consistent changes made.

Just curious, why 20240301 and what is the trailing '3' for?

rljacob:
That's when I minted the DOI at DOECODE (on Friday).  Don't know why they add a 3.

PR #6266, opened 2024-02-27
update v3 wcprod testmods and fix ELM restart

mahf708:
Updates wcprod testmods to align with the v3 model release.

Make tests with production output use v3 watercycle settings.

Also fix a bug in the ELM restart found by CI when using these settings.  

[BFB] but will change fields in EAM, ELM output

PR #6284, opened 2024-03-06
output wcprod testmod defaults more frequently for F-case.  Update B-cases to v3.

mahf708:
Increases frequency of output such that all tapes are outputted in F-case test using wcprod.
Also update all B-case wcprod testmods to use v3 settings.
Improve coverage.

[BFB] but will change fields output in wcprod B-cases and add file in F-cases.


rljacob (approved):

rljacob:
This modifies settings in #6266.

rljacob:
After testing on next, I realized that this PR and #6266 only changed the EAM wcprod testmods for F-cases.  The B-case mods in E3SM/cime_config/testmods_dirs/allactive still have v2 settings I think.  I'll add those changes to this PR.

rljacob:
@wlin7 or @mahf708 or @tangq a debugging run with the v3 output is pointing to a line in chemistry.F90:
```
 forrtl: error (182): floating invalid - possible uninitialized real/complex variable.
 e3sm.exe           0000000000E8EC7A  chemistry_mp_chem        1590  chemistry.F90
```
This is the line:
`               gas_ac_layers(:ncol,1) = gas_ac_layers(:ncol,1) + gas_ac(:ncol,k)`

rljacob:
I'm going to go ahead and merge this.  That bug should be fixed in a separate PR because then we'll be better able to see if it changes answers.

rljacob:
SMS_D_Ld1.ne30pg2_r05_IcoswISC30E3r5.WCYCL1850.chrysalis_intel.allactive-wcprod

PR #6286, opened 2024-03-08
Remove "Build Type" option from albany input file.

mperego:
Remove "Build Type" option from albany input file.
    
Epetra solver stack is no longer available through Albany and this change will be needed with newer versions of Albany.
See https://github.com/sandialabs/Albany/pull/1028

This will work with older versions of Albnay if Albany is built with no Epetra support.


mperego:
Wrong repo, sorry.

Issue #6169, opened 2024-01-25
For pm-gpu, adjust the slurm batch directive for GPU binding (can allow enabling MPS)


ndkeen:
Currently, for pm-gpu, we use `gpus-per-task=1`, and I'd like to change that to be `gpus-per-node=4`.
The reason is that this is needed to use Nvidia's MPS tool which will allow more MPI's per node -- essentially "over-subscribing" the GPU's. This change by itself does not seem to impact our runs (testing with SCREAM) and is BFB.

It might only show improvement if MPS is used. 

The primary change is just this:

```
-      <directive> --gpus-per-task=1</directive>
+      <directive> --gpus-per-node=4</directive>
```
`ndk/machinefiles/pm-gpu-use-gpus-per-node-4`

To active MPS, would only need to add a shell script before the exe, which can be done in a launch script, or perhaps added for all runs in CIME config. 

```
./xmlchange run_exe="~/mps-wrapper.sh `xmlquery --value run_exe`"
```
where the wrapper script provided by NERSC/nvidia and could potentially be altered for specific purposes.

Once in place, a case could use the current settings, eg 4 MPI's per node, and 16 threads per component benefiting from threads and see same performance. OR, use 8 MPI's per node, and 8 threads per component and perhaps see an improvement.
It has been my experience that the improvement can be seen in non-gpu components (ICE, LND) as well as in the ATM. I have also tried even more MPI's -- 16, 32, and 64.

ndkeen:
While E3SM does not have many GPU tests, I did verify we still need the same treatment for MMF (ie use a diff gpu binding) to achieve performance. And then, I have tried several SCREAM cases with this change in place.  All BFB and no noticeable changes in performance.  Again, the reason for this change is to then allow use of MPS to oversubscribe the nvidia GPUS (which requires change in launch script that can eventually be incorporated into all launches if needed).

PR #6292, opened 2024-03-13
For pm-gpu, use gpus-per-node construct

ndkeen:
For pm-gpu (and muller-gpu) use gpus-per-node=4 instead of gpus-per-task=1.

Fixes https://github.com/E3SM-Project/E3SM/issues/6169

[bfb]

### DIFF
--- a/cime_config/machines/config_batch.xml
+++ b/cime_config/machines/config_batch.xml
@@ -433,17 +433,16 @@
     <directives>
       <directive> --constraint=gpu</directive>
     </directives>
-    <directives compiler="gnugpu">
-      <directive> --gpus-per-task=1</directive>
-    </directives>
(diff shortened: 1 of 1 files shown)


ndkeen:
Note this change will have no real impact except to allow using MPS that each case can decide.  While E3SM does not have many GPU tests, I did verify we still need the same treatment for MMF (ie use a diff gpu binding) to achieve performance, so that is the same for MMF. And then, I have tried several SCREAM cases with this change in place.  All BFB and no noticeable changes in performance.  Again, the reason for this change is to then allow use of MPS to oversubscribe the nvidia GPUS (which requires change in launch script that can eventually be incorporated into all launches if needed).

rljacob (approved):

PR #6224, opened 2024-02-08
Implement FB-LTS Time-Stepping Scheme

jeremy-lilly:
This PR implements a new local time-stepping (LTS) scheme in MPAS-Ocean for single layer configurations. This new scheme, called **FB-LTS**, is similar to to the LTS3 scheme already implemented by MPAS-O [[1](https://doi.org/10.1016/j.jcp.2021.110818)], but boasts an increased maximal CFL number. This results in FB-LTS being able to use time-steps between ~2 times larger than time-steps for LTS3 while costing the same number of tendency evaluations.

The FB-LTS scheme is based on the forward-backward, three-stage, second-order Runge-Kutta (FB-RK(3,2)) scheme from [[2](https://doi.org/10.1175/MWR-D-23-0113.1)]. The existing LTS3 scheme is based on a three-stage, third-order, strong stability preserving Runge-Kutta scheme (SSPRK3), so we expect FB-LTS to outperform LTS3 in admittable time-step by factors of between ~1.6-2.2 [[2](https://doi.org/10.1175/MWR-D-23-0113.1)].

In preliminary performance experiments, FB-LTS provides a 8x speedup over RK4 (and 2x over LTS3) in the hurricane Sandy test case described by [[3](https://doi.org/10.1029/2022MS003327)], on a variable resolution mesh going to resolutions as high as 2km in Delaware Bay.

Also, see corresponding PRs to Polaris and Compass with relevant automated tests:
- [E3SM-Project/polaris#187](https://github.com/E3SM-Project/polaris/pull/187): Shows that FB-LTS converges at second order.
- [MPAS-Dev/compass#771](https://github.com/MPAS-Dev/compass/pull/771): Shows that FB-LTS produces the same sea surface height solution as LTS. 

[[1](https://doi.org/10.1016/j.jcp.2021.110818)]: [Capodaglio and Petersen](https://doi.org/10.1016/j.jcp.2021.110818)
[[2](https://doi.org/10.1175/MWR-D-23-0113.1)]: [Lilly, Engwirda, Capodaglio, et al.](https://doi.org/10.1175/MWR-D-23-0113.1)
[[3](https://doi.org/10.1029/2022MS003327)]: [Lilly, Capodaglio, Petersen, et al.](https://doi.org/10.1029/2022MS003327)

[BFB] - only for mpaso-standalone

PR #6293, opened 2024-03-13
Fix some formatting in MPAS-Ocean Registry for generating docs

xylar:
To generate Latex documentation from the Registry, we need to escape Latex in the namelist and variable descriptions.

### DIFF
--- a/components/mpas-ocean/bld/namelist_files/namelist_definition_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_definition_mpaso.xml
@@ -245,7 +245,7 @@ Default: Defined in namelist_defaults.xml
 
 <entry id="config_hmix_ref_cell_width" type="real"
 	category="hmix" group="hmix">
-Reference cell width. If config_hmix_use_ref_cell_width = .true., then hmix coefficients are set to be nu_{2h} = config_mom_del2*(cellWidth / config_hmix_use_ref_cell_width) and nu_{4h} = config_mom_del4*(cellWidth / config_hmix_use_ref_cell_width)^3 where cellWidth is the effective cell width computed as 2*sqrt(areaCell/pi). See Hoch et al 2020 JAMES eq 1,2. This relation applies within a simulation, but also generally among multiple simulations, so the parameters config_mom_del2, config_mom_del4, and config_hmix_use_ref_cell_width can generally remain at their standard values, and just be adjusted for fine tuning.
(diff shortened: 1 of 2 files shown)


mark-petersen (approved):
Yes, this is all correct. Thanks!

components/mpas-ocean/bld/namelist_files/namelist_definition_mpaso.xml
```diff
@@ -2046,7 +2046,7 @@ Default: Defined in namelist_defaults.xml
 
 <entry id="config_n_ts_iter" type="integer"
 	category="split_timestep_share" group="split_timestep_share">
-number of large iterations over stages 1-3
+number of large iterations over stages 1-3; For the split_explicit_ab2 time integrator, this value only affects the first time step when it is not a restart run. For restart runs, this value has no effect on the split_explicit_ab2 time integrator.
```
xylar, line 2049: @jonbob, this came over when I updated with the `generate_e3sm_namelist_files.py` tool.  Was this manually deleted previously on purpose or just missed in a previous merge?

jonbob: I think it was missed in an earlier merge. I think #6224 will bring it in

components/mpas-ocean/bld/namelist_files/namelist_definition_mpaso.xml
```diff
@@ -1196,7 +1196,7 @@ Default: Defined in namelist_defaults.xml
 
 <entry id="config_sw_absorption_type" type="char*1024"
 	category="shortwaveRadiation" group="shortwaveRadiation">
-Name of shortwave absorption type used in simulation. 
+Name of shortwave absorption type used in simulation.
```
jonbob, line 1199: This may have to be fixed in Registry instead of by hand in this generated file

xylar: I fixed this in the registry as well.

xylar: I use an editor that automatically trims trailing whitespace (which I wish more folks would do).  It seems easier to fix this than to revert it.

jonbob:
@xylar -- thanks for having the bld files completely consistent with Registry.

Passes sanity testing with:

merged to next

xylar:
@jonbob, thanks so much!

PR #6294, opened 2024-03-13
update bin_to_cube tool

jsbamboo:
1. modify ``bin_to_cube'' according to https://github.com/NCAR/Topo.git for cubed base topo generation
2. read parameters from ``bin_to_cube.nl''

manually pin @mt5555

### DIFF
--- a/components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.F90
+++ b/components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.F90
@@ -20,7 +20,6 @@ program convterr
   integer :: im, jm
   
   integer :: ncube !dimension of cubed-sphere grid
-!  integer,  parameter :: ncube = 540 !dimension of cubed-sphere grid
   !        integer,  parameter :: ncube = 361 ! for debugging
   
   integer*2,  allocatable, dimension(:,:) :: terr     ! global 30-sec terrain data
@@ -72,18 +71,24 @@ program convterr
   real(r8) :: vol,dx_rad,vol_cube,area_latlon,darea_latlon       ! latitude array
   real(r8), allocatable, dimension(:,:) :: darea_cube
 
-  INTEGER :: UNIT
+  INTEGER :: iargc
 
-  character(len=1024) :: raw_latlon_data_file,output_file
+  character(len=1024) :: raw_latlon_data_file,output_file,string_ncube
   
-  namelist /binparams/ &
-       raw_latlon_data_file,output_file,ncube
-  
-  UNIT=221
-  OPEN( UNIT=UNIT, FILE="bin_to_cube.nl" ) !, NML =  cntrls )
-  READ( UNIT=UNIT, NML=binparams)
-  CLOSE(UNIT=UNIT)
-  write(*,*) "Intermediate cubed-sphere resolution ",ncube
+  iargc = command_argument_count()
+  if (iargc /= 3) then
+      print *, "Usage: <executable> raw_latlon_data_file output_file ncube"
+      stop
+  end if
+
+  call get_command_argument(1, raw_latlon_data_file)
+  call get_command_argument(2, output_file)
+  call get_command_argument(3, string_ncube)
+  read(string_ncube, *) ncube
+
+  write(*,*) "Raw lat/lon data file: ", trim(raw_latlon_data_file)
+  write(*,*) "Output file: ", trim(output_file)
+  write(*,*) "ncube: ", ncube
 
   !
   ! read in USGS data from netCDF file
@@ -284,9 +289,7 @@ program convterr
           STOP
         ELSE
           terr_cube        (i,j,k) = terr_cube        (i,j,k)/weight(i,j,k)                
-!+++ARH
           landfrac_cube    (i,j,k) = landfrac_cube    (i,j,k)/weight(i,j,k)                
-!---ARH
         END IF
       END DO
     END DO
--- a/components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.F90
+++ b/components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.F90
@@ -19,7 +19,7 @@ program convterr
   !
   integer :: im, jm
   
-  integer,  parameter :: ncube = 3000 !dimension of cubed-sphere grid
+  integer :: ncube !dimension of cubed-sphere grid
 !  integer,  parameter :: ncube = 540 !dimension of cubed-sphere grid
   !        integer,  parameter :: ncube = 361 ! for debugging
   
@@ -72,11 +72,25 @@ program convterr
   real(r8) :: vol,dx_rad,vol_cube,area_latlon,darea_latlon       ! latitude array
   real(r8), allocatable, dimension(:,:) :: darea_cube
 
+  INTEGER :: UNIT
+
+  character(len=1024) :: raw_latlon_data_file,output_file
+  
+  namelist /binparams/ &
+       raw_latlon_data_file,output_file,ncube
+  
+  UNIT=221
+  OPEN( UNIT=UNIT, FILE="bin_to_cube.nl" ) !, NML =  cntrls )
+  READ( UNIT=UNIT, NML=binparams)
+  CLOSE(UNIT=UNIT)
+  write(*,*) "Intermediate cubed-sphere resolution ",ncube
+
   !
   ! read in USGS data from netCDF file
   !
   !        status = nf_open('topo-lowres.nc', 0, ncid) !for debugging
-  status = nf_open('usgs-rawdata.nc', 0, ncid)
+  status = nf_open(raw_latlon_data_file, 0, ncid)
+  write(*,*) "Opening: ",TRIM(raw_latlon_data_file)
   IF (STATUS .NE. NF_NOERR) CALL HANDLE_ERR(STATUS)
   
   status = NF_INQ_DIMID(ncid, 'lat', dimlatid)
@@ -105,13 +119,13 @@ program convterr
   
   allocate ( lon(im),stat=alloc_error )
   if( alloc_error /= 0 ) then
-    print*,'Program could not allocate space for landfrac'
+    print*,'Program could not allocate space for lon'
     stop
   end if
   
   allocate ( lat(jm),stat=alloc_error )
   if( alloc_error /= 0 ) then
-    print*,'Program could not allocate space for landfrac'
+    print*,'Program could not allocate space for lat'
     stop
   end if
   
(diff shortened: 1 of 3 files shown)


mt5555:
@brhillman : The important change here is the normalization of terr_cube around line 286.  That's a bug that E3SM introduced when we removed land_frac from this tool chain.   This bug hasn't impacted us since it would only show up if someone made a new cube3000 file (or in this case, Jishi's cube12000 file).   

Another change is the introduction of a namelist for input parameters - but this will break our existing workflow

@jsbamboo : there are still a lot of landfrac_* related variables in cube_to_bin, including writing it to the output file.   Do we need this?  I'm think it is no longer used by cube_to_target, so could be removed here as well.

jsbamboo:
@mt5555 I think you are right. we dont need landfrac* if ``lzero_out_ocean_point_phis'' is set to FALSE in cube_to_target, which is the default setting

please let me know if you want me to remove bin_to_cube.nl / other edits, or feel free to push commits on your end

mt5555:
In: 
https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/4189520033/800m+cubed+topo+generation+from+GMTED2010+15s+DEM

@jsbamboo  reported that cube_to_target required 638G to run, with a 7.5 arch-second lat/lon input data set and cube12000 output.   

Storing the landrac field on this grid is ~110GB.   Thus removing it I think is worthwhile.

jsbamboo:
@mt5555: to clarify, the 638G memory would be needed from 7.5 arch-second lat/lon input (~250m) to **cube48000** output (~200m),  which is kind of pointless since it's more than the original resolution.. maybe processing to cube40000 (250 m) or cube24000 is appropriate, which would require ~75GB or 27GB for ``landfrac_cube''

components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.nl
```diff
@@ -0,0 +1,5 @@
+&binparams
+  raw_latlon_data_file = '/p/lustre2/zhang73/DATA/DEM/S5P_OPER_REF_DEM_15_00000000T000000_99999999T999999_20160111T104226.NCL_24-3.nc'
+  output_file = 'S5P_OPER_REF_DEM_15_24-3_cube12000.nc'
+  ncube=12000
```
whannah1, line 4: are we sure we want this to be the "default"?
Maybe it would be better to create a few different namelist files with the ncube value in the file name so it forces the user to make a decision on which resolution they need rather than blindly using whatever is set up in the single namelist file?

components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.F90
```diff
@@ -153,15 +167,15 @@ program convterr
...
-  END DO
+  !WRITE(*,*) "Adjustments to land fraction: Extend land fraction for Ross Ice shelf by"
+  !WRITE(*,*) "setting all landfractions south of 79S to 1"
+  !DO j=1,jm
+  !  IF (lat(j)<-79.0) THEN
+  !    DO i=1,im
+  !      landfrac(i,j) = 1
+  !    END DO
+  !  END IF
+  !END DO
```
whannah1, line 178: This needs an explanation for why it's commented out

components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.F90
```diff
@@ -254,9 +268,29 @@ program convterr
...
+        IF (ABS(weight(i,j,k))<1.0E-9) THEN
+          WRITE(*,*) "there is no lat-lon grid point in cubed sphere cell ",i,j,k
+          WRITE(*,*) "fatal error"
+          write(*,*) "weight     ",i,j,k,weight(i,j,k)
+          STOP
+        ELSE
+          terr_cube        (i,j,k) = terr_cube        (i,j,k)/weight(i,j,k)                
+!+++ARH
+          landfrac_cube    (i,j,k) = landfrac_cube    (i,j,k)/weight(i,j,k)                
+!---ARH
```
whannah1, line 289: let's remove these "ARH" tags - they don't serve any purpose

components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.F90
```diff
@@ -254,9 +268,29 @@ program convterr
...
+    DO j=1,ncube          
+      DO i=1,ncube
+        IF (ABS(weight(i,j,k))<1.0E-9) THEN
+          WRITE(*,*) "there is no lat-lon grid point in cubed sphere cell ",i,j,k
+          WRITE(*,*) "fatal error"
+          write(*,*) "weight     ",i,j,k,weight(i,j,k)
+          STOP
+        ELSE
+          terr_cube        (i,j,k) = terr_cube        (i,j,k)/weight(i,j,k)                
+!+++ARH
```
brhillman, line 287: I'd drop these signatures, they clutter up the code and don't add value.

components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.F90
brhillman: Or just remove it if we are not going to keep landfrac.

components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.F90
```diff
@@ -254,9 +268,29 @@ program convterr
...
+          landfrac_cube    (i,j,k) = landfrac_cube    (i,j,k)/weight(i,j,k)
```
brhillman, line 288: Do we need to keep landfrac around?

components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.nl
brhillman: These namelists are small enough we could just make them input (command line) arguments as well. I'd prefer that to namelists personally.

jsbamboo: yes it shouldn't be the default, i just picked the one for my test

please check the updates which use the command arguments

components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.F90
jsbamboo: in the updates all landfrac related content are removed

components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.F90
jsbamboo: in the updates, the second commit removed all landfrac related content

for my own experience so far landfrac is not needed so i vote to remove it, but it can be reverted if others want it?

jsbamboo:
the second new commit removed landfrac, i pushed it for review while it can be reverted if anyone wants it

a RRM grid test for bin_to_cube and cube_to_target has passed for the updates, and there is no landfrac anymore in the outputs

components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.F90
```diff
@@ -72,11 +70,31 @@ program convterr
...
+  iargc = command_argument_count()
+  if (iargc /= 3) then
+      print *, "Usage: <executable> raw_latlon_data_file output_file ncube"
+      stop
+  end if
+
+  call get_command_argument(1, raw_latlon_data_file)
+  call get_command_argument(2, output_file)
+  call get_command_argument(3, string_ncube)
+  read(string_ncube, *) ncube
```
brhillman, line 86: I'd make this a command argument too if you're already making the others command arguments. Interactive scripts make it harder to pipe all this into larger script workflows.

components/eam/tools/topo_tool/cube_to_target/cube_to_target.F90
```diff
@@ -856,9 +825,6 @@ program convterr
     
     wt = weights_all(count,1)
     
-    if (lzero_out_ocean_point_phis.AND.landfrac_target(i).lt.0.01_r8) then
```
brhillman, line 859: This is kind of interesting, I hadn't noticed this before. This logical appears to set the new elevation to zero if landfrac is below a threshold value. Maybe this was the purpose for the landfrac variable in the original version. Do you see a problem with removing this? The hard-coded threshold seems kind of problematic because that would probably be resolution dependent.

jsbamboo: yes this flag is one of the reasons why the NCAR repo keeps the landfrac (https://github.com/NCAR/Topo/issues/19) 

i had np problem after removing ``lzero_out_ocean_point_phis'' because i've never enabled this flag before

rljacob:
@brhillman and @whannah1 have all your concerns been addressed?

components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.F90
whannah1: Then let's remove this code entirely.

components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.F90
whannah1: definitely drop these code tags, I can't stand clutter like this!

components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.F90
whannah1: let's drop it and then add it back if something breaks.

components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.F90
```diff
@@ -510,13 +544,31 @@ subroutine wrt_cube(ncube,terr_cube,landfrac_cube,sgh30_cube)
     end do
   end do
   
-  WRITE(*,*) "Create NetCDF file for output"
-  ncstat = nf_create (grid_file_out, NF_64BIT_OFFSET,nc_grid_id)
+  WRITE(*,*) "Create NetCDF file for output: ", TRIM(output_file)
+  !ncstat = nf_create (TRIM(output_file), NF_64BIT_OFFSET,nc_grid_id)
```
whannah1, line 516: remove this commented line

components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.nl
whannah1: Maybe I'm missing something, but I don't see any changes where command line argument support was added?

whannah1:
> @brhillman and @whannah1 have all your concerns been addressed?

No.

components/eam/tools/topo_tool/bin_to_cube/bin_to_cube.nl
jsbamboo: this nl had been deleted since 03/21 commits, and the command line arguments are put in bin_to_cube.F90:

```
  iargc = command_argument_count()
  if (iargc /= 3) then
      print *, "Usage: <executable> raw_latlon_data_file output_file ncube"
      stop
  end if

  call get_command_argument(1, raw_latlon_data_file)
  call get_command_argument(2, output_file)
  call get_command_argument(3, string_ncube)
  read(string_ncube, *) ncube
```

jsbamboo:
most of the latest comments were fixed by the 03/21 commits, but the last one to remove a comment has not yet. I am out of office since 04/05, I can push a commit after I get back to lab on ~04/21

whannah1 (approved):

whannah1:
@rljacob @brhillman I think the PR is ready to be merged.

rljacob:
@brhillman let me know if you don't have time for this.

rljacob:
@mt5555 and @brhillman I've merged this to next.  If you don't want it merged to master, let me know by tomorrow.

mt5555:
great, thanks.  

it will probably be a several more years before we next have to use this tool, but still good to resolve this PR.

PR #6233, opened 2024-02-13
Fixes CNP Budget

bishtgautam:
Sets the values for three fluxes that are included in monthly CNP budgets.

Fixes #6230
[BFB]

PR #6305, opened 2024-03-21
Land: temporarily disable long running developer tests

jgfouca:
I suspect these will be re-enabled shortly, hopefully with adjustments to make them run faster on mappy. Both are taking 2 hours or more in their current state.

[BFB]

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -24,9 +24,9 @@ _TESTS = {
 
     "e3sm_mosart_exenoshare": {
         "time"  : "0:45:00",
-        "tests" : (
-            "ERS.ne30pg2_r05_IcoswISC30E3r5.GPMPAS-JRA.mosart-rof_ocn_2way",
-            )
+        # "tests" : (
+        #     "ERS.ne30pg2_r05_IcoswISC30E3r5.GPMPAS-JRA.mosart-rof_ocn_2way",
+        #     )
         },
 
     "e3sm_mosart_sediment" : {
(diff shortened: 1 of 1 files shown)


rljacob:
The elm-erosion test needs more examination.  @peterdschwartz noted: "when it was PASSing on mappy, the runtime is listed as 1hr".  @ndkeen said "ERS.hcru_hcru.I20TRGSWCNPRDCTCBC.pm-cpu_gnu.elm-erosion runs in 21.5 min on 1 node of pm-cpu"

rljacob:
We can look at master since its also tested on mappy.
Feb 27 pass: https://my.cdash.org/tests/127663013
hash cea78cb,  runtime: about 48 minutes.
March 1: fail  https://my.cdash.org/tests/129583246
hash 6c2f78d timed out 
mappy master had some issues between those dates so no runs.

Between those 2, these ELM-related PRs went in: #6233 #6266

rljacob:
It can be narrowed a bit using next:    fail 99a8988, pass 0d63448  PRs #6273, #6266.  There was a lot of activity on next and master around March 1st because we were finishing the v3.0.0 tag

bishtgautam (approved):

tanzeli1982 (approved):

rljacob:
@bishtgautam and @tanzeli1982 why are you approving removing the elm-erosion test when the problem seems to be unique to mappy?

rljacob (changes requested):
Figure out why mappy and only mappy has a problem with the elm-erosion test.

rljacob:
@amametjanov did you have to add tasks for the elm-erosion test on chrysalis?

amametjanov:
Not on Chrysalis, but on Improv. I think the run-length could be reduced to either `_Ln9` or `_Ld3` (also adjust PEs from 48 to 64 on mappy).

jgfouca:
Discussed this PR during the IG meeting. Other folks are going to move these tests.

PR #6310, opened 2024-03-27
Add SAL updates and topographic wave drag schemes for barotropic tides

sbrus89:
This PR incorporates updates to the global tidal forcing in MPAS-Ocean. It combines two contributions developed under the ICoM project:

1. @dengwirda's modifications to a) include surface pressure in the self attraction and loading (SAL) calculations, b) use a depth rather than distance-based criteria in the SAL coastal smoothing factor, and c) update the harmonic analysis AM to account for initial SSH offsets. See also: https://github.com/E3SM-Ocean-Discussion/E3SM/pull/52.
2. @knbarton's refactoring of the existing [Jayne and St. Laurent](https://journals.ametsoc.org/view/journals/phoc/43/1/jpo-d-12-023.1.xml) topographic wave drag (TWD) parameterization and the addition of the [Zaron and Egbert](https://journals.ametsoc.org/view/journals/phoc/43/1/jpo-d-12-023.1.xml) and [Local Generation Formula](https://eartharxiv.org/repository/view/1252/) TWD schemes.

This PR has been tested using a new compass case, which performs a 125-day tidal run (with a 90 day analysis period) on the VR45to5 mesh using the Zaron and Egbert topographic wave drag option. This results in a 3.3 cm M2 deep water RMSE as compared to the TPXO9 tidal database, which is the most accurate MPAS-Ocean tidal result to date. Here is a link to the compass PR with testing results: https://github.com/MPAS-Dev/compass/pull/802

[NML]
[BFB] - mpas-ocean standalone

### DIFF
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -1108,7 +1108,11 @@
 					description="If true, topographic wave drag is used on the momentum equation."
 					possible_values=".true. or .false."
 		/>
-		<nml_option name="config_topographic_wave_drag_coeff" type="real" default_value="5.0e-4"
+		<nml_option name="config_topographic_wave_drag_scheme" type="character" default_value="JSL" units="unitless"
+					description="hich of the following three wave drag schemes to use: JSL: Jayne and St. Laurent; ZAE: Zaron and Egbert; or NYC: Nycander"
+					possible_values="JSL or ZAE or NYC"
+		/>
+		<nml_option name="config_topographic_wave_drag_coeff" type="real" default_value="5.0e-4" units="unitless"
 					description="Dimensionless topographic wave drag coefficient, $c_{topo}$."
 					possible_values="any positive real, typically 5.0e-4"
 		/>
@@ -2262,7 +2266,11 @@
 				filename_template="topographic_wave_drag.nc"
 				runtime_format="single_file"
 				mode="forward;analysis">
-			<var name="topographic_wave_drag"/>
+			<var name="topo_rinv"/>
+			<var name="topo_buoyancy_N1V"/>
+			<var name="topo_buoyancy_N1B"/>
+			<var name="bathy_stddev"/>
+			<var name="bed_slope_edges"/>
 		</stream>
 
 	</streams>
@@ -3233,8 +3241,24 @@
 		<var name="surfaceFluxAttenuationCoefficientRunoff" type="real" dimensions="nCells Time" units="m"
 			description="The spatially-dependent length scale of exponential decay of river runoff. Fluxes are multiplied by $e^{z/\gamma}$, where this coefficient is $\gamma$."
 		/>
-		<var name="topographic_wave_drag" type="real" dimensions="nEdges" units="1"
+		<var name="topo_rinv" type="real" dimensions="nEdges" units="unitless"
 			description="wave drag coefficient or 1/(rinv) where rinv is the e-folding time used in HyCOM"
+			packages="topographicWaveDragPKG"
+		/>
+		<var name="topo_buoyancy_N1V" type="real" dimensions="nEdges" units="unitless"
+			description="Vertically averaged buoyancy frequencies from WOA 2013 data"
+			packages="topographicWaveDragPKG"
+		/>
+		<var name="topo_buoyancy_N1B" type="real" dimensions="nEdges" units="unitless"
+			description="Bottom buoyancy frequencies from WOA 2013 data"
+			packages="topographicWaveDragPKG"
+		/>
+		<var name="bathy_stddev" type="real" dimensions="nEdges" units="unitless"
+			description="Standard deviation of the bathymetry within a given cell"
+			packages="topographicWaveDragPKG"
+		/>
+		<var name="bed_slope_edges" type="real" dimensions="nEdges" units="unitless"
+			description="The integral gradient average per cell"
 			packages="topographicWaveDragPKG"
 		/>
 		<!-- diagnostic fields for land-ice fluxes -->
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -1975,10 +1975,6 @@
             <var name="gotmDissTopOfCell"/>
             <var name="gotmLengthTopOfCell"/>
             <var name="pgf_sal"/>
-            <var name="temp_twd"/>
-            <var name="normalCoeff"/>
-            <var name="tangentialCoeff"/>
-            <var name="topographicWaveDrag"/>
 		</stream>
 
 		<stream name="output"
@@ -3486,11 +3482,11 @@
 			description="tendency perturbation due to topographic wave drag"
 			packages="tidalPotentialForcingPKG"
 		/>
-		<var name="normalCoeff" type="real" dimensions="nEdges" units=""
+		<var name="normalCoeffTWD" type="real" dimensions="nEdges" units=""
 			description="Coefficient multiplied by the normal velocity in the topographic wave drag"
 			packages="tidalPotentialForcingPKG"
 		/>
-		<var name="tangentialCoeff" type="real" dimensions="nEdges" units=""
+		<var name="tangentialCoeffTWD" type="real" dimensions="nEdges" units=""
 			description="Coefficient multiplied by the tangential velocity in the topographic wave drag"
 			packages="tidalPotentialForcingPKG"
 		/>
--- a/components/mpas-ocean/src/Registry.xml
+++ b/components/mpas-ocean/src/Registry.xml
@@ -1109,8 +1109,8 @@
 					possible_values=".true. or .false."
 		/>
 		<nml_option name="config_topographic_wave_drag_scheme" type="character" default_value="JSL" units="unitless"
-					description="hich of the following three wave drag schemes to use: JSL: Jayne and St. Laurent; ZAE: Zaron and Egbert; or NYC: Nycander"
-					possible_values="JSL or ZAE or NYC"
+					description="hich of the following three wave drag schemes to use: JSL: Jayne and St. Laurent; ZAE: Zaron and Egbert; or LGF: Local generation formula"
+					possible_values="JSL or ZAE or LGF"
 		/>
 		<nml_option name="config_topographic_wave_drag_coeff" type="real" default_value="5.0e-4" units="unitless"
 					description="Dimensionless topographic wave drag coefficient, $c_{topo}$."
@@ -1119,6 +1119,14 @@
 		<nml_option name="config_thickness_drag_type" type="character" default_value="centered"
 					description="The type of layerThickness averaging to use on the drag term. The standard MPAS-O approach is 'centered'."
 					possible_values="'harmonic-mean', 'centered'"
+		/>
+		<nml_option name="config_topographic_wave_drag_cutoff_depth" type="real" default_value="500" units="m"
+					description="Specifies minimum depth at which topographic wave drag is applied"
+					possible_values="any positive real"
+		/>
+		<nml_option name="config_topographic_wave_drag_cutoff_width" type="real" default_value="10" units="m"
+					description="Width over which the topographic wave drag is progressively turned off"
+					possible_values="any positive real"
 		/>
 	</nml_record>
 	<nml_record name="Rayleigh_damping" mode="forward">
@@ -1967,6 +1975,10 @@
             <var name="gotmDissTopOfCell"/>
             <var name="gotmLengthTopOfCell"/>
             <var name="pgf_sal"/>
+            <var name="temp_twd"/>
+            <var name="normalCoeff"/>
+            <var name="tangentialCoeff"/>
+            <var name="topographicWaveDrag"/>
 		</stream>
 
 		<stream name="output"
@@ -2271,6 +2283,9 @@
 			<var name="topo_buoyancy_N1B"/>
 			<var name="bathy_stddev"/>
 			<var name="bed_slope_edges"/>
+			<var name="bottomDepthEdge"/>
+			<var name="lonGradEdge"/>
+			<var name="latGradEdge"/>
 		</stream>
 
 	</streams>
@@ -3261,6 +3276,18 @@
 			description="The integral gradient average per cell"
 			packages="topographicWaveDragPKG"
 		/>
+		<var name="bottomDepthEdge" type="real" dimensions="nEdges" units="unitless"
+			description="bottom depth at the cell edges"
+			packages="topographicWaveDragPKG"
+		/>
+		<var name="lonGradEdge" type="real" dimensions="nEdges" units="unitless"
+			description="Meridional gradient of bathymetry"
+			packages="topographicWaveDragPKG"
+		/>
+		<var name="latGradEdge" type="real" dimensions="nEdges" units="unitless"
+			description="Zonal gradient of bathymetry"
+			packages="topographicWaveDragPKG"
+		/>
 		<!-- diagnostic fields for land-ice fluxes -->
 		<var name="landIceFrictionVelocity" type="real" dimensions="nCells Time" units="m s^-1"
 			description="The friction velocity $u_*$ under land ice"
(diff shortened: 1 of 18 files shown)


components/mpas-ocean/src/shared/mpas_ocn_vel_self_attraction_loading.F
```diff
@@ -665,24 +673,24 @@ subroutine ocn_vel_self_attraction_loading_init(domain,err)!{{{
...
 
+        ! Set-up smoothing mask for shallow regions where SAL should not
+        ! be computed across e.g. wet-dry cells
+        depth_tol = config_self_attraction_loading_depth_cutoff
+        do iCell = 1,nCellsAll
+           depth = ssh(iCell) + bottomDepth(iCell)
+           coastalSmoothingFactor(iCell) = 0.5_RKIND*(tanh( &
+              (depth - 0.5_RKIND*depth_tol) / (0.2_RKIND*depth_tol)) + 1.0_RKIND)
+        enddo
+
```
sbrus89, line 693: This is the switch from distance-based to depth-based SAL smoothing

components/mpas-ocean/src/shared/mpas_ocn_vel_self_attraction_loading.F
```diff
@@ -170,8 +173,12 @@ subroutine ocn_compute_self_attraction_loading(domain, forcingPool, dminfo, ssh,
             return
         endif
 
+        rho0gInv = 1.0_RKIND / rho_sw / gravity
+
         do iCell = 1,nCellsAll
-          sshSmoothed(iCell) = coastalSmoothingFactor(iCell)*ssh(iCell)
+          ! compute SAL wrt. the full barotropic bottom pressure
+          btrPressure(iCell) = coastalSmoothingFactor(iCell) * ( &
+              ssh(iCell) + rho0gInv * surfacePressure(iCell) )
```
sbrus89, line 181: This is where the surface pressure is added to the SAL calculation

components/mpas-ocean/src/analysis_members/mpas_ocn_harmonic_analysis.F
```diff
@@ -355,6 +366,12 @@ subroutine ocn_compute_harmonic_analysis(domain, timeLevel, err)!{{{
 
          ! update if within harmonic analysis period 
          if (daysSinceStartOfSim .le. harmonicAnalysisEnd) then
+           allocate(sshDiff(nCellsSolve))
+
+           do iCell = 1,nCellsSolve
+             sshDiff(iCell) = ssh(iCell) - sshInit(iCell)
+           end do
+
```
sbrus89, line 374: This is the change that makes the harmonic analysis work on the ssh anomaly.

components/mpas-ocean/src/Registry.xml
```diff
@@ -1962,7 +1974,11 @@
             <var name="gotmEpsbTopOfCell"/>
             <var name="gotmDissTopOfCell"/>
             <var name="gotmLengthTopOfCell"/>
-            <var name="ssh_sal"/>
+            <var name="pgf_sal"/>
+            <var name="temp_twd"/>
+            <var name="normalCoeff"/>
+            <var name="tangentialCoeff"/>
+            <var name="topographicWaveDrag"/>
```
sbrus89, line 1981: I don't believe these need to be in the restart stream, since they are calculated in an init routine

components/mpas-ocean/src/shared/mpas_ocn_vel_forcing_topographic_wave_drag.F
```diff
@@ -162,13 +181,100 @@ subroutine ocn_vel_forcing_topographic_wave_drag_init(err)!{{{
...
+                                            * (sqrt((topo_buoyancy_N1B(iEdge)**2 - omegaM2**2) * &
+                                                    (topo_buoyancy_N1V(iEdge)**2 - omegaM2**2))) &
+                                            / (4*pii*omegaM2)
+                      normalCoeff = (lonGradEdge(iEdge)**2) * &
+                                    (cos(angleEdge(iEdge))**2) + (latGradEdge(iEdge)**2)*(sin(angleEdge(iEdge))**2) - &
+                                        2*latGradEdge(iEdge)*lonGradEdge(iEdge)*sin(angleEdge(iEdge))*(cos(angleEdge(iEdge)))
+                      tangentialCoeff = sin(angleEdge(iEdge)) * &
+                                        cos(angleEdge(iEdge)) * &
+                                        (lonGradEdge(iEdge)**2 - latGradEdge(iEdge)**2) + &
+                                        latGradEdge(iEdge)*lonGradEdge(iEdge)*(cos(angleEdge(iEdge))**2 - sin(angleEdge(iEdge))**2)
```
sbrus89, line 269: These variables should be indexed, i.e., `topographicWaveDrag(iEdge) = `, `normalCoeff(iEdge) =`, `tangentialCoeff(iEdge) = `

components/mpas-ocean/src/shared/mpas_ocn_vel_forcing_topographic_wave_drag.F
```diff
@@ -162,13 +181,100 @@ subroutine ocn_vel_forcing_topographic_wave_drag_init(err)!{{{
...
+              call mpas_log_write(" Topographic wave drag scheme is: Jayne and St. Laurent")
+              call mpas_log_write("")
+              tensorScheme = .false.
+
+              do iEdge = 1, nEdgesAll
+                  kmax =  maxLevelEdgeTop(iEdge)
+                  topographicWaveDrag(iEdge) = 0.0_RKIND
+                  if (kmax > 0) then
+                      k =  maxLevelEdgeTop(iEdge)
+                      topoDragTurnOff = (TANH((layerThickEdgeMean(k,iEdge)-topoDragCutoffDepth)/topoDragCutoffWidth)+1)/2
```
sbrus89, line 211: I think I'm going to pull out the application `topoDragTurnOff` factor since it's common to all three drag schemes

components/mpas-ocean/src/shared/mpas_ocn_vel_forcing_topographic_wave_drag.F
```diff
@@ -117,19 +118,37 @@ subroutine ocn_vel_forcing_topographic_wave_drag_tend(normalVelocity, & !{{{
...
+          !$omp do schedule(runtime) private(k)
+          do iEdge = 1, nEdgesOwned
+            k =  maxLevelEdgeTop(iEdge)
+            ! topographic wave drag term:
+            ! du/dt = ... 1/rinv * u
+            ! rinv is the e-folding time use in HyCOM. Here
+            ! topographic_wave_drag = 1/rinv
+            if (k>0) then
+              temp_twd(iEdge) = topographicWaveDrag(iEdge) * normalVelocity(k,iEdge)
+              tend(k,iEdge) = tend(k,iEdge) - topographicWaveDrag(iEdge) * normalVelocity(k,iEdge)
```
sbrus89, line 136: We should just set `tend(k,iEdge) = tend(k,iEdge) - temp_twd(iEdge)`

components/mpas-ocean/src/shared/mpas_ocn_vel_forcing_topographic_wave_drag.F
```diff
@@ -162,13 +181,100 @@ subroutine ocn_vel_forcing_topographic_wave_drag_init(err)!{{{
...
+              do iEdge = 1, nEdgesAll
+                  kmax =  maxLevelEdgeTop(iEdge)
+                  kmin =  minLevelEdgeBot(iEdge)
+                  topographicWaveDrag(iEdge) = 0.0_RKIND
+                  if (kmax > 0) then
+                      k = 1 ! Note: Only works for single layer case
+                      topoDragTurnOff = (TANH((layerThickEdgeMean(k,iEdge)-topoDragCutoffDepth)/topoDragCutoffWidth)+1)/2
+                      topographicWaveDrag(iEdge) = topographicWaveDragCoeff*topoDragTurnOff*50.0_RKIND  * &
+                                (bed_slope_edges(iEdge)**2)                                             * &
+                                (5.24e-3_RKIND*1300.0_RKIND                                             * &
```
sbrus89, line 239: These should be defined as constant variables

components/mpas-ocean/src/Registry.xml
```diff
@@ -1108,14 +1108,26 @@
 					description="If true, topographic wave drag is used on the momentum equation."
 					possible_values=".true. or .false."
 		/>
-		<nml_option name="config_topographic_wave_drag_coeff" type="real" default_value="5.0e-4"
+		<nml_option name="config_topographic_wave_drag_scheme" type="character" default_value="JSL" units="unitless"
+					description="hich of the following three wave drag schemes to use: JSL: Jayne and St. Laurent; ZAE: Zaron and Egbert; or LGF: Local generation formula"
+					possible_values="JSL or ZAE or LGF"
```
dengwirda, line 1113: Is the implementation/tuning for `LGF` well established? I know that `JSL` and especially `ZAE` are working well, but am less confident re: `LGF`.
Should `ZAE` be the default if it's typically the best performing?

components/mpas-ocean/src/shared/mpas_ocn_diagnostics_variables.F
```diff
@@ -72,7 +72,11 @@ module ocn_diagnostics_variables
 
    real (kind=RKIND), dimension(:), pointer :: gradSSH
    real (kind=RKIND), dimension(:), pointer :: pressureAdjustedSSH
-   real (kind=RKIND), dimension(:), pointer :: ssh_sal
+   real (kind=RKIND), dimension(:), pointer :: pgf_sal
+   real (kind=RKIND), dimension(:), pointer :: temp_twd
+   real (kind=RKIND), dimension(:), pointer :: normalCoeff
```
dengwirda, line 77: Perhaps `normalCoeffTWD`, etc to make it clear what these coefficients are for?

components/mpas-ocean/src/shared/mpas_ocn_vel_forcing_topographic_wave_drag.F
dengwirda: I'm not sure what fortran will do in this case, but I feel like testing for `LGF` should be revisited...

dengwirda (commented):
Thanks @sbrus89, this looks good to me, but I've added a few comments re: the `LGF` implementation.

dengwirda:
To add a few explanations for changes:
1. The SAL is changed to act on the full BTR pressure gradient rather than just the SSH part of it, which is necessary to account for config.'s with non-zero surface pressures due to ATM or ISC inputs. Numerically, the 'well-balancing' of the PGF is what's at stake here: if the external / tidal forcing is turned off an ocean in equilibrium should not accelerate away from its initial conditions, which requires that the SAL multiply the full PGF consistently.
2. Computing the SAL can become ill-posed near wet/dry interfaces, as the dry cells should essentially be masked out of the stencil. I found that a thickness-based threshold was more reliable for this than a distance-to-coast, as any cells that are dry (or close to dry) are masked out regardless of near-shore distance. Also doesn't require the addition of a distance-to-coast variable in the ICs.
3. The harmonic analysis is updated to delta_SSH to account for config.'s with initial SSH deflection due to ATM or ISC inputs.

dengwirda:
Compiles and runs for me on `pm-cpu`.

components/mpas-ocean/src/shared/mpas_ocn_vel_forcing_topographic_wave_drag.F
sbrus89: I'm pretty sure it'll set the entire array to a single value, so this definitely needs to be retested.

components/mpas-ocean/src/shared/mpas_ocn_vel_forcing_topographic_wave_drag.F
sbrus89: I agree, thanks!

components/mpas-ocean/src/shared/mpas_ocn_diagnostics_variables.F
sbrus89: I agree, thanks!

components/mpas-ocean/src/Registry.xml
sbrus89: Having ZAE as the default sounds good to me. I'm going to do a full tidal run with all three and I'll report back on the errors. I believe that `LGF` needs to be re-evaluated given the indexing issue I found above.

sbrus89:
I think I also now understand the reason our LGF coefficients are so large (3000 vs the 2.5ish used for adcirc). The lat/lon gradients calculated in the init step are taken with respect to degrees as opposed to radians. These lat/lon gradients are separate from the bed slopes used in ZAE. The bed slopes are calculated in the dem pixel/remap and *are* taken with respect to radians. The 3000 value makes sense since the conversion factor (1/(pi/180))^2 = 3282.2... I'll modify the init step in the compass case to correct this and retest.

dengwirda:
Looks good @sbrus89. 
Just as a reference, ADCIRC v55 (https://gmd.copernicus.org/articles/14/1125/2021/) achieves a (deep ocn) M2 RMSE of around 2.9cm, but with mesh resolution down to 1.5km as I understand it. The 3.3cm MPAS-O results at 5km min mesh spacing seem very competitive. In fact, the shallow ocn errors are reported slightly better for MPAS-O here (12.7cm) than ADCIRC v55 (14cm).

Also makes sense to me re: the `grad(elev)` terms --- the values in the pixel files should be [derivatives wrt. the arc-length](https://github.com/dengwirda/dem/blob/a06940c68121afb764faf9eaf68cc15c9e9ea520/dem_pixel.py#L113) on the sphere.

mark-petersen (approved):
Approving based on @sbrus89's testing in https://github.com/MPAS-Dev/compass/pull/802, previous review in https://github.com/E3SM-Ocean-Discussion/E3SM/pull/52, and @knbarton's work on the topographic wave drag schemes this past year. Thank you @sbrus89! It is fantastic to have this finalized in master!

mark-petersen:
@sbrus89 I ran an extra E3SM test on chrysalis:
```
./create_test SMS_Ln9.T62_oQU240.GMPAS-IAF.chrysalis_gnu
```
and it failed during the run, but `master` failed in the same way, so it appears to be unrelated to this PR.

rljacob:
Note:  an ocean standalone test so can go in.

jonbob:
@dengwirda -- have all your question been answered?

sbrus89:
@jonbob, I still need to report back about the LGF coefficients @dengwirda was asking about. Sorry, I got sidetracked on this.

jonbob:
@sbrus89 -- no problem. I'm just trying to keep PRs moving, so I also updated the bld files here

dengwirda:
@jonbob, @sbrus89 and all, another option could be to just stick with the ZAE and JSL schemes --- I think they're both robust and well tested. If you're keen to merge LGF with the fixes to the array indexing as well though, I'd suggest to work through some tuning + benchmarking first.

sbrus89:
@dengwirda, I confirmed that with the LGF changes described above, the model achieves a 3.36 cm deep M2 RMSE with a wave drag coefficient of 2.5. That wave drag value is much more reasonable and is solidly in the rage used by ADCRIC ([See figure 6 here](https://agupubs.onlinelibrary.wiley.com/doi/10.1029/2021JC018178)). While I haven't done a full set of tuning runs to find an optimal value, this result gives me a lot more confidence that LGF is working correctly.
M2_plot

dengwirda (approved):

dengwirda:
Looks good @sbrus89! Those array indexing issues must have been where the issues with LGA had been lurking in the past.
Thanks for chasing all of this down.

sbrus89:
@jonbob, this one is ready now when convenient.

sbrus89:
@dengwirda - no problem, I'm glad we got it sorted out. Thank you for the SAL/HA changes in this PR. I appreciate you taking the time to review as well!

jonbob:
Passes:

jonbob:

sbrus89:
Thanks very much, @jonbob!

xylar:
A reminder that it is always a good idea to check the compass `pr` suite as part of testing changes that can affect standalone MPAS-Ocean.  I'm seeing segfaults in standalone MPAS-Ocean forward runs starting with this branch.

sbrus89:
@xylar, I looked into this and wanted to note that the issue is in the namelist settings for the `global_ocean` cases which currently use topographic wave drag, but don't have the input stream data now required following this PR. So, the fix to the segfaults in the `pr` suite should be on the compass side and not in E3SM.

xylar:
@sbrus89, as we discussed, that's great and quite a relief!

xylar:
I'm afraid I think this needs to be fixed or reverted ASAP.  I am seeing major changes in MPAS-Ocean init mode for global_ocean test cases.  In particular, I am seeing zero layer thickness in the Icos240 test case for all layers >= 5.

Here is the deepest layer before this change:
before
And after:
after

components/mpas-ocean/src/mode_init/mpas_ocn_init_global_ocean.F
```diff
@@ -893,7 +893,8 @@ subroutine ocn_init_setup_global_ocean_create_model_topo(domain, iErr)!{{{
              end if
              if (isOcean) then
                 ! Enforce minimum depth
-                bottomDepth(iCell) = max(bottomDepthObserved(iCell), refBottomDepth(minimum_levels))
+                bottomDepth(iCell) = max(bottomDepthObserved(iCell), config_global_ocean_minimum_depth)
+                bottomDepth(iCell) = min(bottomDepth(iCell), refBottomDepth(minimum_levels))
```
xylar, line 897: This should have been a `max`, not a `min`.

sbrus89: Thanks for fixing this @xylar! Sorry I overlooked this change and didn't catch it earlier.

PR #6311, opened 2024-03-27
Add zonal mean module for Transformed Eulerian Mean diagnostics

whannah1:
This capability to calculate zonal mean quantities while the model runs has been ported from CESM. The original CESM zonal mean calculations were modified to work with the unstructured FV phys grid (pg2) by Patrick Callaghan (patc@ucar.edu) and then implemented in a branch of E3SMv2 by Jack Chen (cchen@ucar.edu). I've re-implemented it here to facilitate process oriented QBO diagnostics for a SciDAC project, and I hope other projects will find it useful as well. 

Enabling the new zonal mean outputs requires setting  the new namelist variables:
```
phys_grid_ctem_zm_nbas = 120  # Number of basis functions used for TEM
phys_grid_ctem_za_nlat = 90   # Number of latitude points for TEM
phys_grid_ctem_nfreq = -1     # Frequency of TEM diagnostic calculations (neg => hours)
```
And adding the list of new variables to the desired output stream
```
'Uzm','Vzm','Wzm','THzm','VTHzm','WTHzm','UVzm','UWzm','THphys'
```


whannah1:
Here's a quick validation I did from a 1-month run. The figure below shows zonal means calculated from the native grid output in the top row - and the direct output of the new zonal mean code on the bottom row note that the vertical velocity fields are different because we're comparing W [m/s] vs omega [pa/s] - but the general pattern is consistent after reversing the sign.

image

tangq:
@whannah1 , good to see that you have these online TEM diagnostics implemented.

jjbenedict (approved):
This looks okay to me.  I can't claim to fully understand what's happening in components/eam/src/physics/cam/zonal_mean_mod.F90, but I didn't see any red flags and it's reassuring to know that comparison of post-processed zonal means and the inline calculations are extremely close.

wlin7:
Codes look reasonable. developer tests went well. 
Merged to next. It is a useful feature. Thank you for the implementation, @whannah1 .

rljacob:
@whannah1 you should have added documentation about how this works.

PR #6319, opened 2024-04-02
output CCN fields per air mass

mahf708:
Adds ability to directly output CCN fields per air mass.

Instead of the customary units of 1/cc, these edits enable output of CCN fields in units of 1/kg. These are the units expected in the SPA scheme for now.

[BFB]

### DIFF
--- a/components/eam/src/physics/cam/ndrop.F90
+++ b/components/eam/src/physics/cam/ndrop.F90
@@ -58,6 +58,8 @@ real(r8), parameter :: supersat(psat)= & ! supersaturation (%) to determine ccn 
                        (/ 0.02_r8, 0.05_r8, 0.1_r8, 0.2_r8, 0.5_r8, 1.0_r8 /)
 character(len=8) :: ccn_name(psat)= &
                     (/'CCN1','CCN2','CCN3','CCN4','CCN5','CCN6'/)
+character(len=8) :: ccnmair_name(psat)= &
+                    (/'CCN1MAIR','CCN2MAIR','CCN3MAIR','CCN4MAIR','CCN5MAIR','CCN6MAIR'/)
 
 ! indices in state and pbuf structures
 integer :: numliq_idx = -1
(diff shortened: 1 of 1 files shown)


rljacob:
Is 1/kg the CF conventional unit for this quantity?

susburrows:
@rljacob @mahf708 in observational datasets, CCN concentrations are typically reported in units of 1/cc = cm^ -3 = 10^ -6 m^-3.  In earlier versions of EAM, this is a diagnostic output that is mainly useful for comparing with observations in at least a general/statistical sense.  However, 1/kg is generally also a perfectly sensible unit for an atmospheric mixing ratio, and the two units are easily interconvertible (as Naser has done), albeit typically with some error due to the order of operations for averaging and multiplication with / division by air density.

If we can change this easily, I would vote for using one of the cf-compliant fields that Naser has identified, i.e., either number_concentration_of_cloud_condensation_nuclei_at_stp_in_air or number_concentration_of_cloud_condensation_nuclei_in_air.

For the purpose of providing inputs to SPA, the optimal solution may be to generate number_concentration_of_cloud_condensation_nuclei_at_stp_in_air , and then after reading in these values to SPA, apply any necessary transformations to get it into the units that SPA uses for cloud drop number concentration (I believe that variable is carried in 1/kg in EAM, correct? Hence the use of 1/kg for CCN?)

But, I think leaving it as-is (with units of 1/kg) is also fine.

components/eam/src/physics/cam/ndrop.F90
```diff
@@ -58,6 +58,8 @@ module ndrop
                        (/ 0.02_r8, 0.05_r8, 0.1_r8, 0.2_r8, 0.5_r8, 1.0_r8 /)
 character(len=8) :: ccn_name(psat)= &
                     (/'CCN1','CCN2','CCN3','CCN4','CCN5','CCN6'/)
+character(len=8) :: ccnmair_name(psat)= &
```
susburrows, line 61: I think it would be a good idea to add a comment in the code explaining why these (largely redundant) additional CCN diagnostics are being added (i.e. for generating SPA inputs)

components/eam/src/physics/cam/ndrop.F90
```diff
@@ -436,6 +444,7 @@ subroutine dropmixnuc( &
    real(r8), allocatable :: coltend(:,:)       ! column tendency for diagnostic output
    real(r8), allocatable :: coltend_cw(:,:)    ! column tendency
    real(r8) :: ccn(pcols,pver,psat)    ! number conc of aerosols activated at supersat
```
susburrows, line 446: revise comment to "volumetric number conc [1/m3] of aerosols activated at supersat"

components/eam/src/physics/cam/ndrop.F90
```diff
@@ -436,6 +444,7 @@ subroutine dropmixnuc( &
...
+   real(r8) :: ccnmair(pcols,pver,psat) ! number conc of aerosols activated at supersat
```
susburrows, line 447: revise comment to "number conc [1/kg] of aerosols activated at supersat"

susburrows (changes requested):
Thanks, Naser, for putting in this PR.  This PR will enable EAM to save CCN values with the units that are expected by the SPA implementation in EAMxx.  This will both simplify the generation of new input files for SPA, and improve provenance for simulations that use SPA.

I added only a few minor comments requesting additional comments or revisions to comments to make the rationale for the new variables more evident to future readers of this code.

susburrows:
Regarding the choice of units for SPA inputs, I shared my thoughts above, but I would defer to @mahf708 and @hassanbeydoun to make the final decision on this.

hassanbeydoun (approved):

susburrows (approved):

rljacob:
Is this ready to merge?

wlin7:
Merged to next.

Issue #6320, opened 2024-04-02
MPAS-O aborts when a graph partitioning file gives 0 cells to a processor


jonbob:
@xuezhengllnl was testing the new RRSwISC6to18E3r4 mesh in a fully-coupled configuration on pm-cpu and had a test abort in mpas-ocean with the error
````
CRITICAL ERROR: landIceFloatingMask contains the default value which likely indicates that this field is missing in the initial condition file (e.g. because it is meant for an older E3SM version).
````
Analysis showed this happened on processors that had no cells assigned to them in the corresponding graph file (which is another issue)

jonbob:
merged to master

PR #6321, opened 2024-04-02
Add logic to catch issue when partition file assigns 0 cells to a pe

jonbob:
When mpas-ocean runs with no cells assigned to a processor by a graph file and has landIcePressureOn, the run aborts. This fixes that situation by adding a check for the number of cells to the logic that causes the abort.

Fixes: #6320 

[BFB]

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
@@ -181,7 +181,7 @@ contains
          call mpas_pool_get_array(forcingPool, 'landIceDraft', landIceDraft)
          call mpas_pool_get_array(forcingPool, 'landIceFloatingMask', landIceFloatingMask)
          call mpas_pool_get_array(forcingPool, 'landIceFraction', landIceFraction)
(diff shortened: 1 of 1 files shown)


jonbob:
Testing with a oQU240wLI configuration that pointed to a manipulated graph file both replicated the original issue and showed that this fix allows mpas-ocean to successfully complete

cbegeman:
@jonbob Thanks for fixing this. The logic makes sense to me. Would you like me to do any additional testing before approving?

jonbob:
Thanks @cbegeman -- I don't think any testing is necessary, unless you would like to try it. I'm redoing the test I had setup just to make sure this solution works exactly as in the PR. And I'll do more testing to make sure it doesn't change answers on any regular E3SM tests before I merge it

cbegeman (approved):
Approving based on visual inspection and @jonbob's testing.  Thanks for the fix @jonbob!

jonbob:
Passes:
* PEM_Ln9.ne30pg2_r05_IcoswISC30E3r5.WCYCL1850.chrysalis_intel

merged to next

PR #6325, opened 2024-04-03
Implement data atmosphere model for moab driver

iulian787:
Add support for MOAB coupler in data atmosphere cases

This allows running data atm cases like this configuration:
--res ne4pg2_ne4pg2 --compset 2000_DATM%QIA_ELM%CNPRDCTCBCTOP_SICE_SOCN_MOSART_SGLC_SWAV

Follow the implementation from data ocean cases, instance a new iMOAB application on the data atm side, whose degrees of freedom are aligned with the regular MCT decomposition; 
move the method that sets MOAB tags from the AV fields in moab-driver/shr/seq_flds_mod  module, so it can be used by all data models (so use this for data ocean too) 

remove the dependency in the driver_moab on  dimensions_mod from homme, "np" is not used anymore;  

[BFB]

### DIFF
--- a/components/data_comps/datm/src/atm_comp_mct.F90
+++ b/components/data_comps/datm/src/atm_comp_mct.F90
@@ -18,6 +18,11 @@ module atm_comp_mct
   use datm_shr_mod    , only: presaero
   use seq_flds_mod    , only: seq_flds_a2x_fields, seq_flds_x2a_fields
 
+#ifdef HAVE_MOAB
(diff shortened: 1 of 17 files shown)


iulian787:
need more work, it works only on chrysalis

iulian787:
it problems we were seeing were caused by a different bug (https://github.com/E3SM-Project/E3SM/pull/6344), so I am reopening this pull request

rljacob (approved):

PR #6330, opened 2024-04-04
Updates for MPAS-Interface to use Icepack bgc

njeffery:
These changes to MPAS-Seaice only work with an updated Icepack submodule - E3SM-Project/Icepack.git  njeffery/merge-fixes-to-bgc

NOTE that the standard tests will fail (won't build) until we're pointing to the correct Icepack submodule. 
This branch is also pointing to the wrong CIME hash.  Works with the latest [cime @ 4388509](https://github.com/ESMCI/cime/tree/4388509869bd5988d6315e2da65b1a2fbfa604fa)

Commits to e09f1873484ffa3c20ad3bdc5ae1d4e16b52941a were tested in ocean-ice + bgc simulations documented here:
https://acme-climate.atlassian.net/wiki/spaces/ICE/pages/4225073188/icepack.GMPAS.EC30to60E2r2.BGC+and+GMPAS.EC30to60E2r2

Changes are non-BFB with biogeochemistry and/or sea ice aerosols active.

Verified that the latest version of the branch merged to v3 master is BFB in physics simulations and no errors were introduced during the merge.

@eclare108213 and @apcraig are evaluating the Icepack changes.  These will be accepted first before this PR moves from draft status.


rljacob:
@njeffery what is the status of this?

njeffery:
@rljacob :  I'm waiting on the Icepack changes to be accepted in CICE.  Should be soon. I'll close this one before making another PR.

Issue #6332, opened 2024-04-05
build error with nvidia compiler -- local variable lsize has a conflict from a module


ndkeen:
We are seeing build fails with nvidia on pm-cpu (reported on cdash).
When I try one of the tests (say `ERS.f19_g16_rx1.A.pm-cpu_nvidia`), I can reproduce.

In on routine, the integer variable `lsize` has a conflict from an module that is "use"'d.

```
NVFORTRAN-S-0155-lsize is use associated and cannot be redeclared (/global/cfs/cdirs/e3sm/ndk/repos/m23-apr4/components/data_comps/docn/src/docn_comp_mod.F90: 164)
  0 inform,   0 warnings,   1 severes, 0 fatal for docn_comp_init
```

In `components/data_comps/docn/src/docn_comp_mod.F90`, I can either add this `only`, or even just remove the
line altogether to get build to complete.

```
use mct_mod    , only : mct_aVect
```

I didn't check to see if the routine actually needs anything in `mct_mod` (assuming not) or might want it there to be consistent with other routines, etc.

ndkeen:
Note that I verified same issue with an even more recent version of nvidia compiler. With 23.9, still see same build error, and my suggested work-around seem to also work.

PR #6333, opened 2024-04-05
Remove `use mct_mod` to avoid build error with nvidia

ndkeen:
Remove using mct_mod in a routine to avoid a build error with nvidia
(as lsize was in both the module and routine using it)

Fixes https://github.com/E3SM-Project/E3SM/issues/6332
[bfb]


iulian787 (approved):
there is another use mct_mod at file level

ndkeen:
Yep, you're right. So that doesn't make sense that compiler is ok with it at file level, but not at routine level. This one change does allow all developer tests to pass.

PR #6334, opened 2024-04-05
Create the `provis_state` subpool at RK4 initialization to avoid memory leak

sbrus89:
This PR fixes a memory leak I noticed in the RK4 timestepping when running 125 day single-layer barotropic tides cases with the vr45to5 mesh (https://github.com/MPAS-Dev/compass/pull/802) on pm-cpu. Previously, I could only get through about 42 days of simulation before running out of memory.

This issue is related to creating/destroying the `provis_state` subpool at each timestep. We had a similar problem a few years back that required memory leaks fixes in the `mpas_pool_destroy_pool` (https://github.com/MPAS-Dev/MPAS-Model/pull/367) subroutine. However, I believe there is still a memory leak in the `mpas_pool_remove_subpool` routine (which calls `pool_remove_member`) that is called following `mpas_pool_destroy_pool`. It seems like the TODO comment here: https://github.com/E3SM-Project/E3SM/blob/6b9ecaa67c81c65fe1f7063e5afe63ce9b2c66a9/components/mpas-framework/src/framework/mpas_pool_routines.F#L6036-L6038 suggests it is possible things aren't being completely cleaned up by this subroutine.

I'm not familiar enough with the low-level details of the pools framework to track down the memory leak itself. However, in any case, I think it makes more sense to create the `provis_state` subpool once at initialization as opposed to creating and destroying it every timestep. The main consequence of this approach is that the `mpas_pool_copy_pool` subroutine needs to have a `overrideTimeLevel` option similar to that used in `mpas_pool_clone_pool` used previously. 

I've tested these changes with the vr45to5 tides test case and they do allow me to run for a full 125 days and are B4B with the previously create/destroy approach. The LTS and FB-LTS timestepping schemes are also updated here in the same way to prevent the same memory leak. 

Since RK4 is not used in E3SM, this PR is B4B for all E3SM tests. The `mpas_pool_copy_pool` routine modified here is not used in MPAS-Seaice or MALI. See the [Ocean-Discussions PR](https://github.com/E3SM-Ocean-Discussion/E3SM/pull/87) for additional background and testing.

[BFB]

### DIFF
--- a/components/mpas-framework/src/framework/mpas_pool_routines.F
+++ b/components/mpas-framework/src/framework/mpas_pool_routines.F
@@ -930,6 +930,8 @@ module mpas_pool_routines
 !> \details
 !> This routine assumes srcPool and destPool have identical members. It will
 !> copy the data from the members of srcPool into the members of destPool.
+!> The overrideTimeLevels argument enables a given time level of a srcPool
+!> with >1 time level to be copied into a destPool with a single time level.
 !
 !-----------------------------------------------------------------------
    recursive subroutine mpas_pool_copy_pool(srcPool, destPool, overrideTimeLevels)!{{{
--- a/components/mpas-framework/src/framework/mpas_pool_routines.F
+++ b/components/mpas-framework/src/framework/mpas_pool_routines.F
@@ -932,30 +932,30 @@ module mpas_pool_routines
 !> copy the data from the members of srcPool into the members of destPool.
 !
 !-----------------------------------------------------------------------
-   recursive subroutine mpas_pool_copy_pool(srcPool, destPool, overrideTimeLevel)!{{{
+   recursive subroutine mpas_pool_copy_pool(srcPool, destPool, overrideTimeLevels)!{{{
 
       implicit none
 
       type (mpas_pool_type), pointer :: srcPool
       type (mpas_pool_type), pointer :: destPool
-      integer, intent(in), optional :: overrideTimeLevel
+      integer, intent(in), optional :: overrideTimeLevels
 
 
       integer :: i, j, threadNum
-      integer :: timeLevel
+      integer :: timeLevels
       type (mpas_pool_member_type), pointer :: ptr
       type (mpas_pool_data_type), pointer :: dptr
       type (mpas_pool_data_type), pointer :: mem
 
       threadNum = mpas_threading_get_thread_num()
-      timeLevel = 2 
+      timeLevels = -1 
 
-      if (present(overrideTimeLevel)) then
-         timeLevel = overrideTimeLevel
+      if (present(overrideTimeLevels)) then
+         timeLevels = overrideTimeLevels
 
-         if (timeLevel < 1) then
+         if (timeLevels < 1) then
             call mpas_pool_set_error_level(MPAS_POOL_FATAL)
-            call pool_mesg('ERROR in mpas_pool_clone_pool: Input time levels cannot be less than 1.')
+            call pool_mesg('ERROR in mpas_pool_copy_pool: Input time levels cannot be less than 1.')
          end if
       end if
 
@@ -997,8 +997,14 @@ module mpas_pool_routines
 
                   dptr => ptr % data
 
-                  ! Do this through brute force...
                   mem => pool_get_member(destPool, ptr % key, MPAS_POOL_FIELD)
+
+                  ! Allow for overrideTimeLevels
+                  if (timeLevels == -1) then
+                     timeLevels = mem % contentsTimeLevs
+                  endif
+
+                  ! Do this through brute force...
                   if (associated(dptr % r0)) then
                      call mpas_duplicate_field(dptr % r0, mem % r0, copy_array_only=.true.)
                   else if (associated(dptr % r1)) then
@@ -1026,8 +1032,8 @@ module mpas_pool_routines
                   else if (associated(dptr % l0)) then
                      call mpas_duplicate_field(dptr % l0, mem % l0, copy_array_only=.true.)
                   else if (associated(dptr % r0a)) then
-                     if (timeLevel > 1) then
-                        do j=1,mem % contentsTimeLevs
+                     if (timeLevels > 1) then
+                        do j=1,timeLevels
                            mem % r0 => mem % r0a(j)
                            call mpas_duplicate_field(dptr % r0a(j), mem % r0, copy_array_only=.true.)
                            nullify(mem % r0)
@@ -1036,8 +1042,8 @@ module mpas_pool_routines
                          call mpas_duplicate_field(dptr % r0a(1), mem % r0, copy_array_only=.true.)
                      end if
                   else if (associated(dptr % r1a)) then
-                     if (timeLevel > 1) then
-                        do j=1,mem % contentsTimeLevs
+                     if (timeLevels > 1) then
+                        do j=1,timeLevels
                            mem % r1 => mem % r1a(j)
                            call mpas_duplicate_field(dptr % r1a(j), mem % r1, copy_array_only=.true.)
                            nullify(mem % r1)
@@ -1046,8 +1052,8 @@ module mpas_pool_routines
                          call mpas_duplicate_field(dptr % r1a(1), mem % r1, copy_array_only=.true.)
                      end if
                   else if (associated(dptr % r2a)) then
-                     if (timeLevel > 1) then
-                        do j=1,mem % contentsTimeLevs
+                     if (timeLevels > 1) then
+                        do j=1,timeLevels
                            mem % r2 => mem % r2a(j)
                            call mpas_duplicate_field(dptr % r2a(j), mem % r2, copy_array_only=.true.)
                            nullify(mem % r2)
@@ -1056,8 +1062,8 @@ module mpas_pool_routines
                          call mpas_duplicate_field(dptr % r2a(1), mem % r2, copy_array_only=.true.)
                      end if
                   else if (associated(dptr % r3a)) then
-                     if (timeLevel > 1) then
-                        do j=1,mem % contentsTimeLevs
+                     if (timeLevels > 1) then
+                        do j=1,timeLevels
                            mem % r3 => mem % r3a(j)
                            call mpas_duplicate_field(dptr % r3a(j), mem % r3, copy_array_only=.true.)
                            nullify(mem % r3)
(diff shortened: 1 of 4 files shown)


sbrus89:
@matthewhoffman and @akturner, I thought I'd tag you on this since it touches the framework, even though this is apparently the first usage of `mpas_pool_copy_pool` in E3SM.

gcapodag:
great, thanks @sbrus89

sbrus89:
@erinethomas and @darincomeau, would you mind taking a look at the framework changes in this PR just in case you see any issues from the sea ice side?

mark-petersen:
Repeating post from https://github.com/E3SM-Ocean-Discussion/E3SM/pull/87#issuecomment-2037271385

Passes nightly test suite and compares bfb with master branch point on chicoma with optimized gnu and chrysalis with optimized intel. Also passes nighty test suite with debug gnu on chicoma. Note this includes a series of RK4 tests:
```
00:34 PASS ocean_global_ocean_Icos240_WOA23_RK4_performance_test
01:40 PASS ocean_global_ocean_Icos240_WOA23_RK4_restart_test
01:20 PASS ocean_global_ocean_Icos240_WOA23_RK4_decomp_test
01:20 PASS ocean_global_ocean_Icos240_WOA23_RK4_threads_test
```
In E3SM, passes
```
./create_test SMS_Ln9.T62_oQU240.GMPAS-IAF.chrysalis_gnu
./create_test SMS_Ln9.T62_oQU240.GMPAS-IAF.chrysalis_intel
```

mark-petersen (commented):
Tested with LTS on using gnu on chicoma and intel on chryslis, both debug and optimized. The following tests compare bfb against the master branchpoint on both.
```
ocean/dam_break/40cm/default_lts
  * step: initial_state
  * step: lts_regions
  * step: forward
  * step: viz
  test execution:      SUCCESS
  baseline comparison: PASS
  test runtime:        05:11
ocean/dam_break/120cm/default_lts
  * step: initial_state
  * step: lts_regions
  * step: forward
  * step: viz
  test execution:      SUCCESS
  baseline comparison: PASS
  test runtime:        01:00
ocean/dam_break/120cm/ramp_lts
  * step: initial_state
  * step: lts_regions
  * step: forward
  * step: viz
  test execution:      SUCCESS
  baseline comparison: PASS
  test runtime:        01:04
```

mark-petersen (approved):
The above tests show that we get the identical answer to master with standard time stepping and LTS. In addition, tests by @gcapodag show that this addition solves an out-of-memory error on perlmutter that he and @jeremy-lilly have been struggling with for the past two months. Thank you @sbrus89!

mark-petersen:
@xylar and @matthewhoffman these changes are only used by the ocean core, and only for RK4 and LTS timestepping. I think you are safe with a visual review, as I tried to be thorough in my testing of RK4 and LTS and saw BFB results.

```
git grep -in mpas_pool_clone_pool
mpas-framework/src/core_sw/mpas_sw_time_integration.F:120:        call mpas_pool_clone_pool(statePool, provisStatePool, 1)
mpas-framework/src/framework/mpas_pool_routines.F:434:!  routine mpas_pool_clone_pool
mpas-framework/src/framework/mpas_pool_routines.F:444:   recursive subroutine mpas_pool_clone_pool(srcPool, destPool, overrideTimeLevels)!{{{
mpas-framework/src/framework/mpas_pool_routines.F:466:            call pool_mesg('ERROR in mpas_pool_clone_pool: Input time levels cannot be less than 1.')
mpas-framework/src/framework/mpas_pool_routines.F:909:                call mpas_pool_clone_pool(ptr % data % p, newmem % data % p)
mpas-framework/src/framework/mpas_pool_routines.F:921:   end subroutine mpas_pool_clone_pool!}}}
mpas-framework/src/framework/mpas_stream_manager.F:411:           call mpas_pool_clone_pool(manager % defaultAtts, new_stream % att_pool)
mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F:259:    call mpas_pool_clone_pool(tendPool, tendSum1stPool, 1)
mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F:261:    call mpas_pool_clone_pool(tendPool, tendSum2ndPool, 1)
mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F:263:    call mpas_pool_clone_pool(tendPool, tendSum3rdPool, 1)
mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F:265:    call mpas_pool_clone_pool(tendPool, tendSlowPool, 1)
mpas-ocean/src/mode_forward/mpas_ocn_time_integration_rk4.F:233:         call mpas_pool_clone_pool(statePool, provisStatePool, 1)
```

erinethomas:
in line with @mark-petersen's comment above - I was not able to see any place where this would impact MPAS-SI this morning ... so I think its ok on the sea-ice side...

sbrus89:
@mark-petersen, the mpas-ocean `rk4`, `lts`, and `fblts` time stepping routines are the only places that use the `mpas_pool_copy_pool` routine I modified:

```
git grep -in mpas_pool_copy_pool
components/mpas-framework/src/framework/mpas_pool_routines.F:925:!  routine mpas_pool_copy_pool
components/mpas-framework/src/framework/mpas_pool_routines.F:935:   recursive subroutine mpas_pool_copy_pool(srcPool, destPool, overrideTimeLevels)!{{{
components/mpas-framework/src/framework/mpas_pool_routines.F:958:            call pool_mesg('ERROR in mpas_pool_copy_pool: Input time levels cannot be less than 1.')
components/mpas-framework/src/framework/mpas_pool_routines.F:1171:                   call mpas_pool_copy_pool(ptr % data % p, mem % p)
components/mpas-framework/src/framework/mpas_pool_routines.F:1181:   end subroutine mpas_pool_copy_pool!}}}
components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_fblts.F:230:      call mpas_pool_copy_pool(tendPool, tendSum3rdPool, 1)
components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_fblts.F:231:      call mpas_pool_copy_pool(tendPool, tendSlowPool, 1)
components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F:264:    call mpas_pool_copy_pool(tendPool, tendSum1stPool, 1)
components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F:265:    call mpas_pool_copy_pool(tendPool, tendSum2ndPool, 1)
components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F:266:    call mpas_pool_copy_pool(tendPool, tendSum3rdPool, 1)
components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_lts.F:267:    call mpas_pool_copy_pool(tendPool, tendSlowPool, 1)
components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_rk4.F:233:         call mpas_pool_copy_pool(statePool, provisStatePool, 1)
```

xylar (approved):
I looked over the code and discussed improvements with @sbrus89, which he implemented.  I'm happy with the changes now.  I agree that the framework changes only affect MPAS-Ocean and only in configurations that are not used in E3SM production simulations.

components/mpas-framework/src/framework/mpas_pool_routines.F
```diff
@@ -932,20 +932,32 @@ end subroutine mpas_pool_clone_pool!}}}
...
 !
 !-----------------------------------------------------------------------
-   recursive subroutine mpas_pool_copy_pool(srcPool, destPool)!{{{
+   recursive subroutine mpas_pool_copy_pool(srcPool, destPool, overrideTimeLevels)!{{{
 
       implicit none
 
       type (mpas_pool_type), pointer :: srcPool
       type (mpas_pool_type), pointer :: destPool
+      integer, intent(in), optional :: overrideTimeLevels
```
matthewhoffman, line 941: Would be nice to add a comment on what this is and how it is used.  I realize the rest of framework is not documented, so that's sort of unfair to ask.

components/mpas-framework/src/framework/mpas_pool_routines.F
```diff
@@ -1014,83 +1032,135 @@ recursive subroutine mpas_pool_copy_pool(srcPool, destPool)!{{{
...
-                     end do
+                     if (timeLevels > 1) then
+                        do j=1,timeLevels
+                           mem % r0 => mem % r0a(j)
+                           call mpas_duplicate_field(dptr % r0a(j), mem % r0, copy_array_only=.true.)
+                           nullify(mem % r0)
+                        end do
+                     else
+                         call mpas_duplicate_field(dptr % r0a(1), mem % r0, copy_array_only=.true.)
+                     end if
```
matthewhoffman, line 1043: I'm a little confused why the `timeLevels==1` case is handled separately (here and throughout below) when it wasn't before.

matthewhoffman (commented):
@sbrus89 , I'm not very familiar with this code, so what you've done to solve the problem makes sense to me, and I can't review it very critically.  I have two small items, but they are not major concerns.

components/mpas-framework/src/framework/mpas_pool_routines.F
sbrus89: @matthewhoffman, prior to these changes, the `mpas_pool_copy_pool` had assumed that the src/dest pools had the same number of time levels. However, for RK4 the state pool has 2 time levels, but we only need to copy the first into the provis pool. The special treatment is necessary since for a single time level pool, data is stored in the r0/r1/r2/r3... fields as opposed to the r0a/r1a/r2a/r3a... fields where data is stored for pools with >1 timelevels. So, there needs to be a way to copy r0a(1)/r1a(1)/r2a(1)/r3a(1) into r0/r1/r2/r3 when the override time levels option is present. `mpas_pool_clone_pool` (which was what was originally used for this purpose state->provis copy) has this some functionality and I implemented this as close to how that subroutine works as possible.

components/mpas-framework/src/framework/mpas_pool_routines.F
sbrus89: @matthewhoffman, good point. I added a comment under the `\details` section of the header for the subroutine. Let me know if you have suggestions to improve it.

components/mpas-framework/src/framework/mpas_pool_routines.F
matthewhoffman: @sbrus89 , thanks for the explanation!

components/mpas-framework/src/framework/mpas_pool_routines.F
matthewhoffman: Thanks, @sbrus89 , looks good

matthewhoffman (approved):

jonbob:
passes:
* SMS_D_Ld1.ne30pg2_r05_IcoswISC30E3r5.CRYO1850-DISMF.chrysalis_intel

merged to next

jonbob:
merged to master

PR #6336, opened 2024-04-09
update Summit machine config

whannah1:
Update the Summit machine configuration to support the 2024 SummitPLUS atm146 allocation. Verified to work with GNU compoiler, which is the only one planned to be used.

NOTE: the local adios library will need to be rebuilt, but it is currently not needed.

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -5035,15 +5035,6 @@ AMD EPYC 7713 64-Core (Milan) (256GB) and 4 nvidia A100'</DESC>
       <env name="PAMI_IBV_DEVICE_NAME_1">mlx5_3:1,mlx5_0:1</env>
       <env name="PAMI_IBV_DEVICE_NAME">mlx5_0:1,mlx5_3:1</env>
     </environment_variables>
-    <!-- <environment_variables compiler="ibm.*" mpilib="spectrum-mpi"> -->
-      <!-- <env name="ADIOS2_ROOT">$SHELL{if [ -z "$ADIOS2_ROOT" ]; then echo /gpfs/alpine/cli115/world-shared/3rdparty/adios2/2.9.1/spectrum-mpi-10.4.0.3/xl-16.1.1-10; else echo "$ADIOS2_ROOT"; fi}</env> -->
-    <!-- </environment_variables> -->
-    <!-- <environment_variables compiler="pgi.*" mpilib="spectrum-mpi"> -->
-      <!-- <env name="ADIOS2_ROOT">$SHELL{if [ -z "$ADIOS2_ROOT" ]; then echo /gpfs/alpine/cli115/world-shared/3rdparty/adios2/2.9.1/spectrum-mpi-10.4.0.3/nvhpc-21.11; else echo "$ADIOS2_ROOT"; fi}</env> -->
-    <!-- </environment_variables> -->
-    <!-- <environment_variables compiler="gnu.*" mpilib="spectrum-mpi"> -->
-      <!-- <env name="ADIOS2_ROOT">$SHELL{if [ -z "$ADIOS2_ROOT" ]; then echo /gpfs/alpine/cli115/world-shared/3rdparty/adios2/2.9.1/spectrum-mpi-10.4.0.3/gcc-9.1.0; else echo "$ADIOS2_ROOT"; fi}</env> -->
-    <!-- </environment_variables> -->
   </machine>
 
   <machine MACH="ascent">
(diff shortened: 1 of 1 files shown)


cime_config/machines/config_machines.xml
```diff
@@ -5034,15 +5035,15 @@ AMD EPYC 7713 64-Core (Milan) (256GB) and 4 nvidia A100'</DESC>
...
-    <environment_variables compiler="ibm.*" mpilib="spectrum-mpi">
-      <env name="ADIOS2_ROOT">$SHELL{if [ -z "$ADIOS2_ROOT" ]; then echo /gpfs/alpine/cli115/world-shared/3rdparty/adios2/2.9.1/spectrum-mpi-10.4.0.3/xl-16.1.1-10; else echo "$ADIOS2_ROOT"; fi}</env>
-    </environment_variables>
-    <environment_variables compiler="pgi.*" mpilib="spectrum-mpi">
-      <env name="ADIOS2_ROOT">$SHELL{if [ -z "$ADIOS2_ROOT" ]; then echo /gpfs/alpine/cli115/world-shared/3rdparty/adios2/2.9.1/spectrum-mpi-10.4.0.3/nvhpc-21.11; else echo "$ADIOS2_ROOT"; fi}</env>
-    </environment_variables>
-    <environment_variables compiler="gnu.*" mpilib="spectrum-mpi">
-      <env name="ADIOS2_ROOT">$SHELL{if [ -z "$ADIOS2_ROOT" ]; then echo /gpfs/alpine/cli115/world-shared/3rdparty/adios2/2.9.1/spectrum-mpi-10.4.0.3/gcc-9.1.0; else echo "$ADIOS2_ROOT"; fi}</env>
-    </environment_variables>
+    <!-- <environment_variables compiler="ibm.*" mpilib="spectrum-mpi"> -->
```
jgfouca, line 5038: Is it worth keeping this commented-out code?

whannah1: I went back and fort about this, but I think it makes sense to just remove it. Just pushed a commit for this.

cime_config/machines/config_machines.xml
```diff
@@ -5034,15 +5035,6 @@ AMD EPYC 7713 64-Core (Milan) (256GB) and 4 nvidia A100'</DESC>
...
     </environment_variables>
-    <environment_variables compiler="ibm.*" mpilib="spectrum-mpi">
-      <env name="ADIOS2_ROOT">$SHELL{if [ -z "$ADIOS2_ROOT" ]; then echo /gpfs/alpine/cli115/world-shared/3rdparty/adios2/2.9.1/spectrum-mpi-10.4.0.3/xl-16.1.1-10; else echo "$ADIOS2_ROOT"; fi}</env>
-    </environment_variables>
-    <environment_variables compiler="pgi.*" mpilib="spectrum-mpi">
-      <env name="ADIOS2_ROOT">$SHELL{if [ -z "$ADIOS2_ROOT" ]; then echo /gpfs/alpine/cli115/world-shared/3rdparty/adios2/2.9.1/spectrum-mpi-10.4.0.3/nvhpc-21.11; else echo "$ADIOS2_ROOT"; fi}</env>
-    </environment_variables>
-    <environment_variables compiler="gnu.*" mpilib="spectrum-mpi">
-      <env name="ADIOS2_ROOT">$SHELL{if [ -z "$ADIOS2_ROOT" ]; then echo /gpfs/alpine/cli115/world-shared/3rdparty/adios2/2.9.1/spectrum-mpi-10.4.0.3/gcc-9.1.0; else echo "$ADIOS2_ROOT"; fi}</env>
-    </environment_variables>
```
wlin7, line 5045: Is the above block a leftover from previous clean-up? I remember I had to manually remove the similar block of script for chrysalis to build. Asking in case a replacement is needed for this change.

wlin7: Oh, reviewing this before reading the PR msg. All sounds good.

wlin7 (approved):

cime_config/machines/config_machines.xml
whannah1: Replacement is not needed for what I plan to do on Summit. If the need for Adios arises we can buidl a new local version and add env variables here again.

whannah1:
> @whannah1, I noticed that a hard-coded path is used for "DIN_LOC_ROOT" and similar entries in "config_machines.xml". This might cause failures for users not in the "atm146" group. I wonder if this is intentional.

If you look at the old path you'll notice that it was also hardcoded to a specific project "cli115", but this didn't cause problems because everyone using summit for E3SM/SCREAM was on that project. I'm not aware of a better place to put the inputdata, but given that it's in the "world-shared" path I think other people can access it. Also, I'm happy to add other people to the atm146 project to run simulations on Summit. This allocation is associated with the SummitPLUS proposal I submitted for doing development and long simulations with the new MMF configuration, but since I'm already behind on burning this allocation I think it's fine for others to use it for related E3SM dev or sims.

jgfouca (approved):

grnydawn:
Other than the file permission issue, everything looks good to me. I think I will merge this PR soon if there are no further concerns.

whannah1:
> @whannah1, you're right that it is accessible to users outside of the "atm146" group. One thing you might need to do is add write permission for non-atm146 users under "/gpfs/alpine2/atm146/world-shared/e3sm". I ran several E3SM tests but they failed due to permission issues.

That's good to know - I'll figure out how to fix this.

grnydawn:
> @grnydawn you forgot to merge this to next. That has to be done at the terminal.

@rljacob , I will merge this branch into 'next' right now. Thanks for finding the mistake.

PR #6342, opened 2024-04-11
use a markdown linter on new files

mahf708:
Adds a markdown linter on new files in PRs to ensure they conform with some linting rules (can modify these later if needed).

[BFB]

### DIFF
--- a/.github/workflows/e3sm-gh-ci-cime-tests.yml
+++ b/.github/workflows/e3sm-gh-ci-cime-tests.yml
@@ -3,16 +3,6 @@ name: gh
 on:
   pull_request:
     branches: [ master ]
-    paths:
-      # first, yes to these
-      - 'cime_config/**'
-      - 'components/eam/**'
-      - 'components/elm/**'
-      # second, no to these
-      - '!components/eam/docs/**'
-      - '!components/eam/mkdocs.yml'
-      - '!components/elm/docs/**'
-      - '!components/elm/mkdocs.yml'
 
   workflow_dispatch:
 
--- a/.github/workflows/e3sm-gh-ci-cime-tests.yml
+++ b/.github/workflows/e3sm-gh-ci-cime-tests.yml
@@ -3,6 +3,11 @@ name: gh
 on:
   pull_request:
     branches: [ master ]
+    paths:
+      - 'cime_config/**'
+      - 'components/eam/**'
+    paths-ignore:
+      - 'components/eam/docs/**'
 
   workflow_dispatch:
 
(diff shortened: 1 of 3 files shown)


.github/workflows/e3sm-gh-ci-cime-tests.yml
```diff
@@ -3,6 +3,11 @@ name: gh
 on:
   pull_request:
     branches: [ master ]
+    paths:
+      - 'cime_config/**'
+      - 'components/eam/**'
+    paths-ignore:
+      - 'components/eam/docs/**'
```
rljacob, line 10: F-cases also test the land model so should include components/elm/**

rljacob:
The linter will only run on new or modified files?

bartgol (approved):
Where will the linter report? Do we just open the action log and see some output from it?

rljacob:
Good I want it to just look at new/modified files.  Don't fix any old ones.

rljacob:
How do I run this in my docs development environment on my laptop?

rljacob:
pre-commit hook would be good.  Can you add one to https://github.com/E3SM-Project/E3SM-Hooks  ?

rljacob:
Actually the hooks in that repo are bash-based hooks that run anywhere.  It looks like the pre-commit hook for the linter needs python and some things installed.

.github/workflows/e3sm-gh-ci-cime-tests.yml
```diff
@@ -3,6 +3,16 @@ name: gh
 on:
   pull_request:
     branches: [ master ]
+    paths:
+      # first, yes to these
+      - 'cime_config/**'
+      - 'components/eam/**'
+      - 'components/elm/**'
```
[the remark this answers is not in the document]
rljacob: Yes do that and also include edits to the coupler.

bartgol:
This action already blocks the PR merging, right? If so, I'm ok if with an action (rather than a hook), forcing the dev to do a bit of work.

Quick question for @mahf708 : does this action only run on _new_ files? What if someone modifies an existing one, using bad formatting?

.github/workflows/e3sm-gh-ci-cime-tests.yml
```diff
@@ -3,16 +3,6 @@ name: gh
 on:
   pull_request:
     branches: [ master ]
-    paths:
```
bartgol: Why did you remove these? If a PR only touches documentation, it seems pointless to run the CIME tests..

bartgol (approved):

.github/workflows/e3sm-gh-ci-cime-tests.yml
rljacob: Those went in a different PR.

bartgol: Ah, weird. GH showed them to me as part of the changes since my last review. But indeed it doesn't show them when I look for all changes.

Thanks!

PR #6345, opened 2024-04-14
Generalize cice-qc scripts to work with E3SM-Polar-Developer.sh output

eclare108213:
Original scripts assumed that the run directories being compared are named 'run'.  The E3SM-Polar-Developer scripts append the run directory names with .k000, .k001 etc for different combinations of configuration options.  This PR generalizes the cice-qc scripts to allow other run directory names, which are input by the user.  Also changed the sample paths in the anvil script to use anvil paths rather than chrysalis paths.

Changes only post-processing scripts.  Tested on a new pair of runs updating the icepack submodule (QC passed).
[BFB]

### DIFF
--- a/components/mpas-seaice/testing/cice-qc/job_script.cice-qc.anvil
+++ b/components/mpas-seaice/testing/cice-qc/job_script.cice-qc.anvil
@@ -14,7 +14,11 @@ source /lcrc/soft/climate/e3sm-unified/load_latest_e3sm_unified_anvil.sh
 
 export HDF5_USE_FILE_LOCKING=FALSE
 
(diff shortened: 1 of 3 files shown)


darincomeau (approved):
Agree it makes sense to remove the forced `run` to the paths.

Approved based on visual inspection and developer testing.

rljacob:
Waiting on @proteanplanet

proteanplanet (approved):

jonbob:
verified that this PR passes the e3sm_ice_developer suite on chrysalis

merged to next

jonbob:
merged to master

PR #6348, opened 2024-04-16
add WCYCL2010NS compset and introduce gh/ci-w tests

mahf708:
Adds gh/ci WCYCL tests by reducing the amount of input data needed (via introducing WCYCL2010NS compset).

### DIFF
--- a/cime_config/allactive/config_compsets.xml
+++ b/cime_config/allactive/config_compsets.xml
@@ -70,6 +70,12 @@
   <lname>1850_EAM%CMIP6_ELM%CNPRDCTCBCTOP_MPASSI_MPASO_MOSART_SGLC_SWAV</lname>
 </compset>
 
+<!-- Same as WCYCL1850NS but for 2010 (to avoid needing transient files). Mainly for testing. -->
+<compset>
+  <alias>WCYCL2010NS</alias>
+  <lname>2010_EAM%CMIP6_ELM%CNPRDCTCBCTOP_MPASSI_MPASO_MOSART_SGLC_SWAV</lname>
+</compset>
+
 <compset>
   <alias>WCYCL1950</alias>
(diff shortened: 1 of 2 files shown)


cime_config/allactive/config_compsets.xml
```diff
@@ -70,6 +70,12 @@
   <lname>1850_EAM%CMIP6_ELM%CNPRDCTCBCTOP_MPASSI_MPASO_MOSART_SGLC_SWAV</lname>
 </compset>
 
+<!-- Same as WCYCL1850NS but for 2010 (to avoid needing transient files). Mainly for testing. -->
+<compset>
+  <alias>WCYCL2010NS</alias>
```
rljacob, line 75: 1850 needs transient files?  I thought that was for perpetual pre-industrial conditions.

rljacob: @wlin7 why is 1850 relying on transient files?

wlin7: @rljacob , I don't know if there is a particularly good reason. It has been used in this manner since v2. I can think of one justifiable reason: the source data do not have much (if any) inter-annual variability to present PI condition around 1850.  This is not the case for 2010, and thus a climatology around 2010 is used, instead of a single year.

As for file size, other than when needing to download to a machine for the first time, there is no practical difference reading from large transient file for year 1850 vs. from a single year (1850) file.

rljacob: Yes but in the cloud-based testing we are doing, the input data is downloaded new every time the test runs and disk space is very limited.  So we want all the input files to be a minimum in size

wlin7: Agreed. That was not in mind when putting those compsets together. We may want to revisit. This WCYCL2010NS is a good alternative for testing purpose.

rljacob (approved):

iulian787:
@mahf708 thanks for this case, it is extremely useful for testing

PR #6357, opened 2024-04-18
Moves the ELM erosion test to E3SM integration testsuite

bishtgautam:
The ELM erosion test takes more than 30 mins to run on Mappy. 
Since the ELM erosion physics is not turned on by default presently,
this test is moved to the E3SM integration testsuite.

[BFB]

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -55,7 +55,6 @@ _TESTS = {
         "time"  : "0:45:00",
         "tests" : (
             "ERS_Ld20.f45_f45.IELMFATES.elm-fates",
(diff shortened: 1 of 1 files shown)


peterdschwartz:
@rljacob  I think this just needs a review and I can merge it.

peterdschwartz:
merged to next.  Nightly testing may not give enough nodes first time around (I did the e3sm_integration with --no-run to check for syntax errors or anything like that and noticed it was going to use only 1 node).

peterdschwartz:
merged to master

PR #6362, opened 2024-04-23
Adds new I-Compsets that are consistent with WCv3

bishtgautam:
- Adds I-compsets that include active BGC (CNP+RD+CTC), black carbon (BC), and 
  subgrid topographic parameterization for solar radiation (TOP) in ELM.
- These I-Compsets use GSWP DATM forcing.
- The compsets are for 1850, 20TR, and 2000.
[BFB]

### DIFF
--- a/components/elm/cime_config/config_compsets.xml
+++ b/components/elm/cime_config/config_compsets.xml
@@ -538,6 +538,11 @@
   </compset>
 
   <compset>
+    <alias>IGSWCNPRDCTCBCTOP</alias>
+    <lname>2000_DATM%GSWP3v1_ELM%CNPRDCTCBCTOP_SICE_SOCN_MOSART_SGLC_SWAV</lname>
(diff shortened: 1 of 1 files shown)


rljacob:
Do these need to be in a test suite also or is it enough that they are tested in fully coupled cases?

rljacob:
@acme-y9s please review this.

bishtgautam:
To get this merged to the master on Monday, we can skip adding new tests as that would require nightly tests to be run. Later on, we can replace a few existing tests with tests using these new compsets. Till the time being, we can rely on the existing fully coupled tests.

rljacob (approved):

peterdschwartz:
@rljacob should I go ahead a merge to next

rljacob:
Yes go ahead.

peterdschwartz:
merged to master

PR #6369, opened 2024-04-25
move semoab_mod file to homme shared folder

iulian787:
semoab_mod.F90 file is moved from
components/eam/src/dynamics/se folder to
components/homme/src/share folder.

It belongs naturally there, as it involves just grid routines from homme base code

also, it should prepare for SCREAM-moab connection later on

fix also some complaints about non-allocated arrays when there are no cells on some dynamic tasks (when number of tasks is larger than number of spectral cells in homme)

[BFB]


oksanaguba:
not sure what to review here -- i do not think standalone homme would pick this file in its builds, also, it is probably not possible to incorporate moab functionality into standalone homme.

PR #6370, opened 2024-04-26
Fix for bad leaf C:N ratios during transient land simulations

### LINKED ISSUE #6368: Incorrect C:N ratios and C:P causing balance errors in transient simulations
This was noticed in the v2 offline feature simulations.  It seems to be related to the bolded section in the code below from components/elm/src/biogeochem/ComputeSeedMod.F90 dealing with allocating seed C, N and P to display, storage and xfer components of the leaf.
pleaf, pstorage, and pxfer are computed for C,N and P separately.  If tot_leaf is very small these proportions may be different for C,N,P, causing C:N and C:P ratios to diverge from the expected fixed values in the parameter file. In rare cases, this leads to negative N or P and eventual balance errors.  Removing the bolded code and always allocating 100% to either leaf or storage seems to fix this issue.

>  if (tot_leaf == 0._r8 .or. ignore_current_state) then
        if (veg_vp%evergreen(pft_type) == 1._r8) then
            pleaf    = 1._r8
         else
            pstorage = 1._r8
         end if
    else
         **pleaf    = leaf        /tot_leaf
         pstorage = leaf_storage/tot_leaf
         pxfer    = leaf_xfer   /tot_leaf**
    end if

dmricciuto:
Fix for seed initialization that caused C:N an C:P ratios to diverge for some grid cells in transient simulations.

Fixes #6368  

[non-BFB]

### DIFF
--- a/components/elm/src/biogeochem/ComputeSeedMod.F90
+++ b/components/elm/src/biogeochem/ComputeSeedMod.F90
@@ -183,16 +183,10 @@ contains
     pstorage = 0._r8
     pxfer    = 0._r8
 
-    if (tot_leaf == 0._r8 .or. ignore_current_state) then
-       if (veg_vp%evergreen(pft_type) == 1._r8) then
-          pleaf    = 1._r8
-       else
-          pstorage = 1._r8
-       end if
+    if (veg_vp%evergreen(pft_type) == 1._r8) then
+       pleaf    = 1._r8
     else
-       pleaf    = leaf        /tot_leaf
-       pstorage = leaf_storage/tot_leaf
-       pxfer    = leaf_xfer   /tot_leaf
+       pstorage = 1._r8
     end if
 
   end subroutine LeafProportions


rljacob:
Is this non-BFB just for 20TR cases?

components/elm/src/biogeochem/ComputeSeedMod.F90
```diff
@@ -183,17 +183,17 @@ subroutine LeafProportions(pft_type, ignore_current_state, &
...
-    else
-       pleaf    = leaf        /tot_leaf
-       pstorage = leaf_storage/tot_leaf
-       pxfer    = leaf_xfer   /tot_leaf
-    end if
+    !else
+    !   pleaf    = leaf        /tot_leaf
+    !   pstorage = leaf_storage/tot_leaf
+    !   pxfer    = leaf_xfer   /tot_leaf
+    !end if
```
bishtgautam, line 196: let's delete these commented-out lines and reindent the remaining lines.

rljacob: @dmricciuto please see above.

rljacob:
Discussion: might be needed for transient runs.  should be BFB for 1850 spinup.

rljacob:
Would definitely change answers for already started v3 20TR runs so probably can't be in v3.0 code.

bpbond (approved):

rljacob:
@bishtgautam this is the old 3.1beta PR.   Would it be the first land one to go in?

bishtgautam:
@rljacob, The order in which this non-BFB is merged into the master does not matter.

rljacob:
go ahead and start merging this

components/elm/src/biogeochem/ComputeSeedMod.F90
rljacob: @dmricciuto lease take care of above comment.

rljacob:
@dmricciuto so this wasn't ready?

PR #6386, opened 2024-05-02
Add BGC v2 compsets to maint-2.1

rljacob:
Add compsets used in v2 BGC simulations to maint-2.1 for better provenance and reproducibility.


rljacob:
All this isn't needed for @BunnyVon cases that were run with 2.1.

PR #6313, opened 2024-03-29
Enforcing system abort when there is negative channel storage

liho745:
Enforces MOSART to stop when there is negative channel storage for any tracer. The e3sm_mosart_developer tests all passed at Compy BFB. 

Fixes #6302 
[BFB]

PR #6388, opened 2024-05-02
change the grid to r05_r05 for lnd_rof_two_way test

donghuix:
ERS.f09_f09.IELM.elm-lnd_rof_2way in the test suits results in negative channel water storage in MOSART. 
This was raised in https://github.com/E3SM-Project/E3SM/pull/6313. 
 The reason is that f09_f09 uses different spatial resolutions in ELM and MOSART. 
This is problematic for land river two-way coupling, which requires the same grid in ELM and MOSART. 

This PR changed f09_f09 to r05_r05 for the land river two-way test, and there is no negative main channel storage.

[BFB]


peterdschwartz:
Since this has one review and the discussion on the other PR seemed to reach an agreement, I'll go ahead and merge this PR and the other one today

peterdschwartz:
merged to next

PR #6389, opened 2024-05-02
remove eamxx from e3sm docs for now

mahf708:
Removes EAMxx from the main e3sm docs for while it remains under intense development.


docs/index.md
```diff
@@ -5,7 +5,7 @@ The E3SM documentation is organized into sections for each component model and a
 ## Component Models
 
 - [EAM](./EAM/index.md)
-- [EAMxx](./EAMxx/index.md)
+- [EAMxx](https://E3SM-project.github.io/scream)
```
rljacob, line 8: It should be removed entirely.

rljacob: Or even better make this just plain text.
`- EAMxx - not yet supported.`

If we link to EAMxx now from this page, we're implying that is somewhat supported.  We don't want that.

rljacob (approved):

PR #6393, opened 2024-05-03
Created land SPINUP compsets with E3SMv3 and GSWP3 forcing

acme-y9s:
Adding E3SMv3 forcing and generating the associated land spinup composts

### DIFF
--- a/components/elm/cime_config/config_compsets.xml
+++ b/components/elm/cime_config/config_compsets.xml
@@ -133,6 +133,25 @@
     <alias>I1850CNPRDCTCBCTOP</alias>
     <lname>1850_DATM%QIA_ELM%CNPRDCTCBCTOP_SICE_SOCN_MOSART_SGLC_SWAV</lname>
   </compset>
+  <!---I E3SMv3 forcing for v3 land spinup compsets-->
+  <compset>
+    <alias>I1850E3SMv3CNTOP</alias>
+    <lname>1850_DATM%E3SMv3_ELM%CNRDCTCBCTOP_SICE_SOCN_MOSART_SGLC_SWAV</lname>
+  </compset>
+
+  <compset>
+    <alias>I1850E3SMv3CNPTOP</alias>
+    <lname>1850_DATM%E3SMv3_ELM%CNPRDCTCBCTOP_SICE_SOCN_MOSART_SGLC_SWAV</lname>
+  </compset>
+  <!---I GSWP3v1 forcing for v3 land spinup compsets-->
+  <compset>
+    <alias>I1850GSWP3v1CNTOP</alias>
+    <lname>1850_DATM%GSWP3v1_ELM%CNRDCTCBCTOP_SICE_SOCN_MOSART_SGLC_SWAV</lname>
+  </compset> 
+  <compset>
+    <alias>I1850GSWP3v1CNPTOP</alias>
+    <lname>1850_DATM%GSWP3v1_ELM%CNPRDCTCBCTOP_SICE_SOCN_MOSART_SGLC_SWAV</lname>
+  </compset>
 
   <compset>
     <alias>ICB1850CRDCTCBC</alias>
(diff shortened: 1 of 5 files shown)


bishtgautam (commented):
Why are we adding a new compset (`*_DATM%E3SMv3_`) instead of using `CPLHIST` in the run scripts and archiving the run scripts? If we add new compsets this way, we will have to add new compsets every time we have a new model version released. Furthermore, it is not obvious from the compset name that this DATM forcing corresponds to the WC-v3 and does not correspond to the BGC-v3/other v3 simulation campaign.

rljacob (changes requested):
We decided a while ago not have version numbers in compsets.  It makes less sense as the years go on.  Use a string that indicates how it is scientifically different from the other options.

bishtgautam:
One more follow up question: is the new DATM data already available on our public inputdata server?

acme-y9s:
@bishtgautam Not yet! How should I deal with it? Thanks!

bishtgautam:
On blues, the public inputdata dir is at `/lcrc/group/acme/public_html/inputdata`. Do you know how big is the new dataset?

components/elm/cime_config/config_compsets.xml
```diff
@@ -133,6 +133,25 @@
...
   </compset>
+  <!---I E3SMv3 forcing for v3 land spinup compsets-->
+  <compset>
+    <alias>I1850E3SMv3CNTOP</alias>
+    <lname>1850_DATM%E3SMv3_ELM%CNRDCTCBCTOP_SICE_SOCN_MOSART_SGLC_SWAV</lname>
+  </compset>
+
+  <compset>
+    <alias>I1850E3SMv3CNPTOP</alias>
+    <lname>1850_DATM%E3SMv3_ELM%CNPRDCTCBCTOP_SICE_SOCN_MOSART_SGLC_SWAV</lname>
```
rljacob, line 144: What is "E3SMv3" supposed to convey in this compset?   We know its E3SM and "v3" is, again, not something we want in compsets in general.

rljacob (changes requested):
"E3SM" can't appear in compset names.  We know its E3SM.

components/elm/cime_config/config_compsets.xml
acme-y9s: @rljacob  E3SMv3 meaning that the climate forcing files used to drive offline land simulation was from E3SMv3 coupled simulation coupler output, which we used  to drive the land BGC SPINUP simulation for v3 coupled simulations. we also added the E3SMv2 coupler output files as the land BGC v2 simulation forcing files (DATM%E3SMv2) in our BGCv2 compset branch, which I will submit all commits from the branch to master later. If I  change the compset short name such as I1850E3SMv3CNPTOP to I1850E3CPLCNPTOP, what is a good to distinguish the coupler output forcing from v2 and v3? Thanks.

acme-y9s:
@bishtgautam It is 20-year coupler output forcing from E3SMv3 forcing which we used for land BGC SPINUP for v3 coupled simulation, which is 109G located on Chrysalis. Do you think we should upload it to the public input repository? Thanks.

components/elm/cime_config/config_compsets.xml
rljacob: I see its forcing instead of an internal configuration.  But isn 't that what the CPLHIST option is for?

bishtgautam: I too believe this new compset is just what CPLHIST is for.

acme-y9s:
@bishtgautam @rljacob Yes, You both are right. It is what CPLHIST is for. What is the smart way to use one option CPLHIST fro different coupler output forcings? We should not specify the pathes and file names of the coupler forcings in namelist file, and will give these information in runscript file when we try to generate a case and submit a job?

rljacob:
Still needed after #6416 ?

rljacob:
@acme-y9s yes it should be enough to just archive the runscript that used the CPLHIST option with the specific v3 input files you used for the spinup.

rljacob:
I'm going to propose we close this PR.  The right way to preserve the ability to do the spinup is to save the runscript.

thorntonpe:
@rljacob I agree that this can be closed.

Issue #6395, opened 2024-05-03
Build error on perlmutter if module matlab is loaded


ndkeen:
Ziming Chen at PNNL discovered this issue.  We seem to get a build fail in MCT with matlab module loaded.  I will explicitly unload matlab. Credit to NERSC support for recognizing it might be matlab (Rebbeca Hartman-Baker and Anneter Greiner).

```
Finished SHAREDLIB_BUILD for test SMS_D.ne4pg2_oQU480.F2010.pm-cpu_intel in 40.939578 seconds (FAIL). [COMPLETED 1 of 1]
    Case dir: /pscratch/sd/n/ndk/e3sm_scratch/pm-cpu/SMS_D.ne4pg2_oQU480.F2010.pm-cpu_intel.20240503_163756_zq6pal
    Errors were:
        Building test for SMS in directory /pscratch/sd/n/ndk/e3sm_scratch/pm-cpu/SMS_D.ne4pg2_oQU480.F2010.pm-cpu_intel.20240503_163756_zq6pal
        ERROR: /global/cfs/cdirs/e3sm/ndk/repos/nexty-apr26/share/build/buildlib.mct FAILED, cat /pscratch/sd/n/ndk/e3sm_scratch/pm-cpu/SMS_D.ne4pg2_oQU480.F2010.pm-cpu_intel.20240503_163756_zq6pal/bld/mct.bldlog.240503-163900

Waiting for tests to finish


...

checking how to get verbose linking output from ftn... -v
[...]

PR #6396, opened 2024-05-04
Unload matlab module on all NERSC machines

ndkeen:
For pm-cpu/pm-gpu and other machines at NERSC, we found that the matlab module caused problems with builds.

Fixes https://github.com/E3SM-Project/E3SM/issues/6395

[bfb]

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -206,6 +206,7 @@
         <command name="unload">aocc</command>
         <command name="unload">cudatoolkit</command>
         <command name="unload">climate-utils</command>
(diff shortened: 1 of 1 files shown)


rljacob:
Remind me why we don't purge all the modules at the start?

ndkeen:
It is not default loaded, but could be common module users have. In the past Ive tried starting with a module purge and building up, but there is too much headache with core modules that system wants and the order of loading can be critical. We cant unload everything, but can catch those known to cause problems. I actually prefer not to purge as sometimes I want a module loaded without messing with e3sm settings (craypat for example)

PR #6399, opened 2024-05-06
Move extrasnowlayers test to e3sm_landice_developer suite

jonbob:
Since PR #6226 redefined the IGELM compset, the SMS.ne30_oECv3_gis.IGELM_MLI.elm-extrasnowlayers test has been failing. This moves that test from the e3sm_ocnice_extra_coverage suite to the e3sm_landice_developer suite, which is only run on platform/compiler combinations that support active MALI.

[BFB] for all current E3SM tests

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -124,6 +124,7 @@ _TESTS = {
         "tests"   : (
             "SMS.ne30pg2_r05_EC30to60E2r2_gis20.IGELM_MLI.elm-gis20kmSMS",
             "ERS.ne30pg2_r05_EC30to60E2r2_gis20.IGELM_MLI.elm-gis20kmERS",
+            "SMS.ne30_oECv3_gis.IGELM_MLI.elm-extrasnowlayers",
             )
         },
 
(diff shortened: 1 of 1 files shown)


jonbob:
passed the e3sm_landice_developer suite with the new test on chrysalis with gnu

stephenprice:
Thanks @jonbob. This looks good to me, but I wonder if we need to also add these two lines ...

```
config_velocity_solver = 'none'
config_thickness_advection = 'none'
```

... to the nl file associated w/ this test:

`/components/elm/cime_config/testdefs/testmods_dirs/elm/extrasnowlayers/user_nl_mali
`

I think this test probably assumes no MALI solve and previous defaults were such that this was the case by default.

Flagging @chloewhicker, @matthewhoffman, and @trhille so that they can also have a quick look and comment. I'll also add @chloewhicker as a reviewer since this is her test.

chloewhicker:
Good catch Steve! Yes, for `use_extrasnowlayers` I always run with those two lines in my `user_nl_mali`

rljacob:
@stephenprice please approve.

stephenprice (approved):
Sorry for the delay on this. I did not see it until just now. Approved!

jonbob:
Tested by running the e3sm_landice_developer suite on chrysalis with gnu. The new test passes, though there are issues with the gnu compiler that impact other tests

merged to next

jonbob:
merged to master and new test is blessed on both chrysalis and pm-cpu

PR #6400, opened 2024-05-07
Add design doc for radial basis functions (RBF)

xylar:
This is a 12 year old design doc that doesn't have a home and should be available to those who need it.


rljacob:
Is the actual document the rbf.pdf file?  And the rest is the source for that?

xylar:
@rljacob, yes, but I don't want to remove the PDF because it isn't that easy to recompile.

rljacob:
That's fine.  Could you link to it from somewhere in the MPAS-ocean docs?

xylar:
It should ideally be linked to in all the MPAS component documentation but it's *highly* specialized.  So far, the developer's guide for MPAS-Ocean does not get into this level of detail.  It only came up because @akturner needs the related code in the DEMSI sea-ice project and couldn't understand the code I write 12 years ago without the help of the design doc, and I figured it shouldn't be hidden away in my personal email anymore.

I'll chat with @mark-petersen and @alicebarthel once the tutorial is over about whether there's a place for this in the MPAS-Ocean docs.

rljacob (approved):

xylar:
Thanks @rljacob.

xylar:
@akturner, you were the one to ask for this.  Could you approve when you have a chance?

@alicebarthel, since you took the first crack at the documentation, I wanted to check with you to see if there would be a place in the docs to link to this?  If not, that's fine.  We can just leave it as a relatively hidden piece of documentation for now.  Hopefully, it will be better documented when it makes its way into Omega.

xylar:
@mark-petersen, could I get you to take a quick look at this?  The main decision that is needed is whether we want to link to this from the MPAS-Ocean documentation.

mark-petersen:
@xylar, thanks for thinking about this. These old documents are valuable references and worth making available.

All the documents of this age and format are in this directory:
https://github.com/MPAS-Dev/MPAS-Documents/tree/master/shared

My recommendation is to put it there and add a link from the current documentation. That way we don't have a single latex-pdf format mixed in with the new restructured text format, and we avoid adding 400KB to the permanent repository.

xylar:
Okay, @mark-petersen, I'm fine with moving this to MPAS-Documents.  It's not clear to me that there's a good way to link to that in the current documentation.  Could you find a good place for that at make the appropriate PR?

I don't think a 400 kb document is worth worrying about in a repo that expand to multiple GB when all the submodules are checked out.  That's just not important.

xylar:
Closing in favor of https://github.com/MPAS-Dev/MPAS-Documents/pull/78

mark-petersen:
Yes, I can make the PR to the old Documents repo. We can put a link in the current documentation like this:
https://github.com/MPAS-Dev/MPAS-Documents/blob/master/shared/multiple_blocks/mpas_multiple_blocks.pdf

and people can download it. I'm also happy to put all these design documents on zotero, as we did with the user's guide, so we have a permanent link to it.

xylar:
@mark-petersen, thanks for merging that other PR!  What I had in mind for a PR for you to make was actually one here in E3SM just to add the link to the PDF.  I figure you have a better sense of where it should fit into the docs.

PR #6402, opened 2024-05-07
For NERSC machines (including pm-cpu, pm-gpu) remove `exclusive` slurm batch directive

ndkeen:
Remove the `exclusive` directive from the batch files for all NERSC machines.
The reason we want to do this: 
a) no longer needed (in fact default)
b) it allows a user to use `--qos shared` when desired (and there is no flag to turn _off_ exclusive)

Note that using `-q shared or --qos shared` allows a job to share resources with other jobs on the node.
The advantage is that jobs using shared are not charged as much as a full node.

There will be no performance impact.

[bfb]


rljacob (approved):

ndkeen:
merged to next

ndkeen:
Do not expect anything to be different with this change, but still wanted to see tests look ok. Merging to master

PR #6404, opened 2024-05-08
Fix zonal mean TEM output

whannah1:
This provides some minor fixes for the recent zonal mean output capability. The most notable change is in how the zonal mean vertical velocity and theta are modified.

[BFB] (except for ZM variables)

### DIFF
--- a/components/eam/src/physics/cam/phys_grid_ctem.F90
+++ b/components/eam/src/physics/cam/phys_grid_ctem.F90
@@ -243,7 +243,7 @@ subroutine phys_grid_ctem_diags(phys_state)
 
    character(len=*), parameter :: prefix = 'phys_grid_ctem_diags: '
 
-   real(r8) :: ps(pcols,pver,begchunk:endchunk)
+   real(r8) :: ps(pcols,begchunk:endchunk)
 
    real(r8) :: u(pcols,pver,begchunk:endchunk)
    real(r8) :: v(pcols,pver,begchunk:endchunk)
--- a/components/eam/src/physics/cam/phys_grid_ctem.F90
+++ b/components/eam/src/physics/cam/phys_grid_ctem.F90
@@ -236,6 +236,8 @@ end subroutine phys_grid_ctem_init
 !-----------------------------------------------------------------------------
 !-----------------------------------------------------------------------------
 subroutine phys_grid_ctem_diags(phys_state)
+   use physconst,     only: rair, cpair
+
    type(physics_state), intent(in) :: phys_state(begchunk:endchunk)
 
    character(len=*), parameter :: prefix = 'phys_grid_ctem_diags: '
@@ -275,7 +277,6 @@ subroutine phys_grid_ctem_diags(phys_state)
    real(r8) :: thza(nzalat,pver)
 
    real(r8) :: mbarv ! molecular weight of dry air (g/mol)
-   real(r8) :: sheight(pcols,pver) ! pressure scale height (m)
 
    if (.not.do_calc()) return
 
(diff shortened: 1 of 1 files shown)


components/eam/src/physics/cam/phys_grid_ctem.F90
```diff
@@ -269,44 +276,37 @@ subroutine phys_grid_ctem_diags(phys_state)
...
 
-   real(r8) :: mbarv ! molecular weight of dry air (g/mol)
-   real(r8) :: sheight(pcols,pver) ! pressure scale height (m)
-
-   if (.not.do_calc()) return
-
    ! In CESM/WACCM the variable mbarv is provided by the "air_composition" 
    ! module, which is not in E3SM, so we just use a rough approximation
-   mbarv = 28.97
+   real(r8) :: mbarv = 28.97 ! molecular weight of dry air (g/mol)
```
jjbenedict, line 287: Molecular weight of dry air is `mwdair`in `components/eam/src/control/physconst.F90`, and can be traced back to `share/util/shr_const_mod.F90`:
` real(R8),parameter :: SHR_CONST_MWDAIR  = 28.966_R8       ! molecular weight dry air ~ kg/kmole`

components/eam/src/physics/cam/phys_grid_ctem.F90
```diff
@@ -269,44 +276,37 @@ subroutine phys_grid_ctem_diags(phys_state)
...
-
+      ! surface pressure
+      ps(:ncol,lchnk) = phys_state(lchnk)%ps(:ncol)
       ! potential temperature
-      theta(:ncol,:,lchnk) = phys_state(lchnk)%t(:ncol,:) * phys_state(lchnk)%exner(:ncol,:)
-
-      ! vertical velocity
-      w(:ncol,:,lchnk) = -sheight(:ncol,:) *  phys_state(lchnk)%omega(:ncol,:) / phys_state(lchnk)%pmid(:ncol,:)
-
+      theta(:ncol,:,lchnk) = phys_state(lchnk)%t(:ncol,:) * ( 1000e2 / phys_state(lchnk)%pmid(:ncol,:) )**(rair/cpair)
```
jjbenedict, line 296: Need `_R8` on `1000e2`?  I never know.

jjbenedict (commented):
What's the motivation for the updated calculations of `omega` and `theta` in `subroutine phys_grid_ctem_diags`?  Greater accuracy, or preference to use existing available fields rather than a more convoluted method like scale height?  Also, is zonal mean PS added to facilitate vertical interpolation to pressure levels?

rljacob:
Is this non-bfb just for the new TEM output vars?

components/eam/src/physics/cam/phys_grid_ctem.F90
whannah1: oh, I didn't realize we had a constant for this. I'll update to use the shared value.

components/eam/src/physics/cam/phys_grid_ctem.F90
whannah1: no, it's not required.

whannah1:
> Is this non-bfb just for the new TEM output vars?

Yes.

whannah1:
> What's the motivation for the updated calculations of `omega` and `theta` in `subroutine phys_grid_ctem_diags`? 

The version of omega that uses scale height corresponds to an alternative derivation of the TEM budget that uses a log pressure coordinate, but we feel it's better to use the conceptually simpler pressure coordinate version and use the vertical velocity directly, rather than rely on an assumed scale height. 

As for theta - the problem in the original code is due to how "exner" is defined. Instead of a conventional reference pressure, the value of exner in the physics state strucuture uses surface pressure, which causes the TEM values over antarctica, or any latitude band with topography, to look very odd. I think this is something that is not a problem in CESM, but rather a E3SM-specific problem that we need to address to utilize the zonal mean potential temperature. 

> Also, is zonal mean PS added to facilitate vertical interpolation to pressure levels?
Yes, exactly.

components/eam/src/physics/cam/phys_grid_ctem.F90
whannah1: actually, I forgot that we can't delete this entirely since we don't need the scale height anymore.

whannah1: pushed a commit to remove this.

jjbenedict (approved):

rljacob:
@brhillman can you merge?

brhillman:
Merged to next.

brhillman:
Merged to master.

PR #6408, opened 2024-05-11
Updates ELM subroutine to include dt as an input

### LINKED ISSUE #5913: ELM: variable used before it is defined
I noticed this compilation warning 
```
ftn-7212 ftn: WARNING ROOTDYNAMICS, File = ../../../autofs/nccs-svm1_home1/lbertag/workdir/scream/scream-src/master/components/elm/src/biogeochem/RootDynamicsMod.F90, Line = 214 
  Variable "dt" is used before it is defined.
```
Indeed, the variable dt is a local var, never assigned to, and yet used on the RHS of some expressions. Its usage is confined to some if statement, so I _assume_ these if statements never evaluate to true. Nevertheless, some ELM folk should look into this, to make sure it's a benign warning. If so, I would recommend removing the unused code, to avoid other developers getting alarmed by this. ;)

Note: I was building the scream repo code, but I verified that the file has the same issue in e3sm/master.

bishtgautam:
The `dt` is now included as an input to the subroutine `RootDynamics()`.

Fixes #5913
[BFB]

### DIFF
--- a/components/elm/src/biogeochem/CNEcosystemDynBetrMod.F90
+++ b/components/elm/src/biogeochem/CNEcosystemDynBetrMod.F90
@@ -298,7 +298,7 @@ module CNEcosystemDynBetrMod
 
           call RootDynamics(bounds, num_soilc, filter_soilc, num_soilp, filter_soilp, &
                canopystate_vars, cnstate_vars, &
-               crop_vars, energyflux_vars, soilstate_vars)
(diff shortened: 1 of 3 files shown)


peterdschwartz (commented):
Add dt to the arguments section with intent(in)

peterdschwartz (approved):

peterdschwartz:
merged to next

peterdschwartz:
merged to master

PR #6409, opened 2024-05-11
Fixes the intent of a variable in ELM subroutine

### LINKED ISSUE #5758: Incorrect intent(out) in SoilTemperatureMod
In SoilTemperatureMod, `tvector` is annotated `intent(out)` in `SolveTemperature`:
https://github.com/E3SM-Project/E3SM/blob/d1388ddf71ced4a11ff364bf1eac80f2bad70666/components/elm/src/biogeophys/SoilTemperatureMod.F90#L749
But in `SoilTemperature`, it appears that `tvector` is intended to be `intent(inout)`:
https://github.com/E3SM-Project/E3SM/blob/d1388ddf71ced4a11ff364bf1eac80f2bad70666/components/elm/src/biogeophys/SoilTemperatureMod.F90#L427-L444
`BandDiagonal` correctly treats the argument as `intent(inout)`:
https://github.com/E3SM-Project/E3SM/blob/d1388ddf71ced4a11ff364bf1eac80f2bad70666/components/elm/src/biogeophys/BandDiagonalMod.F90#L43
Many compilers treat `intent(out)` as essentially `intent(inout)`, which very likely makes this bug benign in practice.

bishtgautam:
The intent of `tvector` is changed from `in` to `inout`.

Fixes #5758
[BFB]

### DIFF
--- a/components/elm/src/biogeophys/SoilTemperatureMod.F90
+++ b/components/elm/src/biogeophys/SoilTemperatureMod.F90
@@ -746,7 +746,7 @@ contains
     integer                , intent(in)  :: jtop(bounds%begc: )                        ! top level at each column
     integer                , intent(in)  :: jbot(bounds%begc: )                        ! bottom level at each column
     logical                , intent(in)  :: urban_column                               ! Is true if solving temperature for urban column, otherwise false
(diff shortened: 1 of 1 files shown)


peterdschwartz (approved):

peterdschwartz:
merged to next

peterdschwartz:
merged to master

PR #6411, opened 2024-05-13
Add shared queue for pm-cpu/pm-gpu

ndkeen:
After https://github.com/E3SM-Project/E3SM/pull/6402 removed `exclusive` directive, which allows jobs to use shared qos, I realized we still needed the _option_ to use `shared` qos.

This PR should allow:
```
./xmlchange JOB_QUEUE=shared
```

I also adjusted the total number of nodes available for each machine.

[bfb]


jgfouca (approved):

ndkeen:
merged to next

PR #6413, opened 2024-05-14
Adds a new snow compaction & routing scheme in ELM

chloewhicker:
Modifies the ELM snowcapping routing scheme along with the snow and firn compaction
schemes to include changes made by [Schneider et al., (2022)](https://agupubs.onlinelibrary.wiley.com/doi/10.1029/2021MS002542)
but retain the original 5-layer snowpack, these modifications can be turned on by
setting use_firn_percolation_and_compaction = .true. in user_nl_elm

[BFB]
[NML] 

----------------

design doc [here](https://acme-climate.atlassian.net/wiki/spaces/ICE/pages/4264591395/Design+Document+Improved+Snowcapping+Scheme+Snow+Physics)


rljacob:
Can you fix the PR title?

components/elm/src/main/elm_varcon.F90
```diff
@@ -249,7 +249,8 @@ subroutine elm_varcon_init()
     allocate( dzsoifl(1:nlevsoifl            ))
 
     if (use_extrasnowlayers) then
-       h2osno_max = 30000._r8
+        !h2osno_max = 1000._r8
```
bishtgautam, line 252: can you please delete this commented out line?

chloewhicker: Done. Thanks for catching that.

bishtgautam (approved):

PR #6414, opened 2024-05-14
Update modules for intel/gnu,impi/mvapich on Bebop

dqwu:
intel from 18.0.4 to 20.0.4
intel-mkl from 2018.4.274 to 2020.4.304
intel-mpi from 2018.4.274 to 2019.9.304

hdf5, netcdf, pnetcdf versions are updated (some are downgraded) to
match the new compilers/mpilibs.

Most of the new modules are also used on Anvil.

Also define HDF5_ROOT for both intel and gnu compilers.

[BFB]

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -2854,34 +2854,45 @@
         <command name="load">anaconda3/5.2.0</command>
       </modules>
       <modules compiler="intel">
-        <command name="load">intel/18.0.4-443hhug</command>
-        <command name="load">intel-mkl/2018.4.274-jwaeshj</command>
-        <command name="load">hdf5/1.10.5-3mk3uik</command>
-        <command name="load">netcdf/4.7.0-krelxcz</command>
-        <command name="load">netcdf-fortran/4.4.5-74lj75q</command>
+        <command name="load">gcc/7.4.0</command>
+        <command name="load">intel/20.0.4-lednsve</command>
+        <command name="load">intel-mkl/2020.4.304-voqlapk</command>
       </modules>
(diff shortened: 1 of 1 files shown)


amametjanov (approved):

rljacob:
merged to next to see if it helps with Kokkos build fails.

cime_config/machines/config_machines.xml
```diff
@@ -2854,34 +2854,45 @@
...
+        <command name="load">netcdf-fortran/4.4.4-b4ldb3a</command>
+        <command name="load">parallel-netcdf/1.11.0-kj4jsvt</command>
       </modules>
       <modules compiler="gnu">
-        <command name="load">gcc/8.2.0-g7hppkz</command>
-        <command name="load">intel-mkl/2018.4.274-2amycpi</command>
-        <command name="load">hdf5/1.8.16-mz7lmxh</command>
-        <command name="load">netcdf/4.4.1-xkjcghm</command>
-        <command name="load">netcdf-fortran/4.4.4-mpstomu</command>
+        <command name="load">gcc/8.2.0-xhxgy33</command>
```
rljacob, line 2878: Should we update the gcc version?  I notice that bebop has gcc/11.3.0 (but not 11.2).

rljacob: Ok, we'll leave it alone until someone tries it and can't compile.

rljacob:
The bebop test will run this evening.  Then we should know if its ok to merge this.

rljacob:
The build errors went away.  No longer have time on bebop so no tests ran.  Merging anyway.

PR #6416, opened 2024-05-14
Add compsets used in v2 BGC simulations

evasinha:
Add compsets used in v2 BGC simulations to master for better provenance and reproducibility.

The branch is a rebase of Xiaoying Shi's `acme-y9s/BGCV2_compsets` .

[BFB]

----------
All tests are passing except for `NLCOMP` fail in several of them. The `NLCOMP` fail is because of three new variables being added to elm_in :
```
Differences in namelist 'elm_inparm':
   found extra variable: 'tw_irr'
   found extra variable: 'use_atm_downscaling_to_topunit'
   found extra variable: 'use_hydrstress'
```
Since all these variables are set to false by default with the new update, I think we can ignore these fails.


components/elm/src/dyn_subgrid/dyncropFileMod.F90
```diff
@@ -179,7 +179,7 @@ subroutine dyncrop_interp(bounds,crop_vars)
     allocate(pfertcft_cur(bounds%begg:bounds%endg, max_topounits, cft_lb:cft_ub))
     call pfertcft%get_current_data(pfertcft_cur)
 
-    call collapse_crop_types(wtcft_cur, nfertcft_cur, pfertcft_cur, bounds%begg, bounds%endg, verbose = .false.)
+    !call collapse_crop_types(wtcft_cur, nfertcft_cur, pfertcft_cur, bounds%begg, bounds%endg, verbose = .false.)
```
rljacob, line 182: You should add a comment about why this was commented out.

components/mosart/bld/namelist_files/namelist_defaults_mosart.xml
```diff
@@ -43,7 +43,7 @@ for the CLM data in the CESM distribution
 
 <frivinp_rtm rof_grid="r0125" >rof/mosart/MOSART_global_8th_20180211c.nc</frivinp_rtm>
 <frivinp_rtm rof_grid="r025" >rof/mosart/MOSART_global_qd_20240212.v3.nc</frivinp_rtm>
-<frivinp_rtm rof_grid="r05" >rof/mosart/MOSART_global_half_20180721a.nc</frivinp_rtm>
+<frivinp_rtm rof_grid="r05" >rof/mosart/MOSART_Global_half_20210422.nc</frivinp_rtm>
```
rljacob, line 46: This seems like it would change answers for v2.1 watercycle cases because they use r05 river.

evasinha: @rljacob Yes, you are correct. I found that while running tests as well. I will be reverting this change as we only want BFB changes in this PR.

components/elm/src/dyn_subgrid/dyncropFileMod.F90
evasinha: I have reverted this edit so as to only keep BFB changes in this PR.

evasinha:
@rljacob I have struggling to get the various tests in `e3sm_land_developer` to PASS. Few commits made since the original PR was submitted have helped resolve some of those fails but not all. Most of the test are still failing due to difference in namelist (`NLCOMP` fail) which is expected since three new name list variables are added to `lnd_in` ```tw_irr = .false.
 73  urban_hac = 'ON'
 74  urban_traffic = .false.
 75  use_atm_downscaling_to_topunit = .false.

rljacob:
We've seen that "required more processors" message and think it has to do with executable sharing within the land test suite.    Run each test individually to see if they still fail.

For the tests failing because of namelist additions from a testmod, yes either the feature the testmod activcatesw needs a new compset or maybe the set of features doesn't make sense anymore.   Do you know where to find the testmods?  The first one is in components/elm/cime_config/testdefs/testmods_dirs/elm/ctc_f19_g16_I20TRGSWCNPRDCTCBC.

evasinha:
The minor omissions identified in [this commit](https://github.com/E3SM-Project/E3SM/pull/6416/commits/5ef4e17a0d4b39b6ca23b85c020a56a85a9896a5) were causing several tests to fail without any good error message to identify the cause of failure.

evasinha:
@rljacob FINALLY after the recent few commits (typos and omits that took me a very long time to identify) all tests are passing except for `NLCOMP` fail in several of them.  The `NLCOMP` fail is because of three new variables being added to `elm_in` :
```
Differences in namelist 'elm_inparm':
   found extra variable: 'tw_irr'
   found extra variable: 'use_atm_downscaling_to_topunit'
   found extra variable: 'use_hydrstress'
```
Since all these variables are set to false by default with the new update, I think we can ignore these fails.

The PR is now ready to be reviewed and approved.

components/elm/bld/namelist_files/namelist_defaults.xml
```diff
@@ -130,6 +128,8 @@ attributes from the config_cache.xml file (with keys converted to upper-case).
 <paramfile nu_com="ECA" use_cn=".true." use_crop=".false." >lnd/clm2/paramdata/clm_params.cbgc.c07292018.nc</paramfile>
 <paramfile use_crop=".false." use_cn=".true."  >lnd/clm2/paramdata/clm_params_c211124.nc</paramfile>
 <paramfile use_crop=".false." use_fates=".true."  >lnd/clm2/paramdata/clm_params_c211124.nc</paramfile>
+<paramfile nu_com="RD" use_crop=".false." >lnd/clm2/paramdata/clm_params_c180524_phs.nc</paramfile>
```
bbye, line 131: Since crops use RD it's not clear why a different parameter file is used when crops are off, unless this file has fewer PFTs. But, the parameter file with crops should also reflect any changes to parameters for natural vegetation that are in this parameter file.

bbye (approved):
I made one comment, but it's not holding up this PR. More of a philosophical question of how we want to handle updates to parameter files to maintain consistency and not confuse the user.

components/elm/bld/namelist_files/namelist_defaults.xml
evasinha: @bbye The parameter file in this PR (`clm_params_c180524_phs.nc`) is for 25 pfts. The new crop parameter file (`clm_params_c230517_phs_50pfts.nc`) was made by combining the parameter file in this PR (`clm_params_c180524_phs.nc`) and an earlier crop parameter file (`clm_params_c230517_phs_50pfts.nc`). 

So ideally, using `clm_params_c230517_phs_50pfts.nc` file in this PR should work but it has not been tested.

rljacob: Can there be one params file for all cases?

qzhu-lbl (approved):

evasinha:
@bbye  and @rljacob  I tested running the `I1850GSWCNP` compset with the crop parameter file (`clm_params_c230517_phs_50pfts.nc`) but got the following error message:

``` ERROR: The provided user buffer does not have space for copying strings (space available to copy only           25  strings) pio_support::pio_die:: myrank=          -1 : ERROR: pionfget_mod.F90:```

I think it requires  more changes to use the 50pft parameter file when crops are off and number of parameters is less than 50.

rljacob:
@evasinha did you want to this to be merged to maint-2.1?  Because the target is currently master.

evasinha:
@rljacob We would like to merge it to master. I had rebase to the current master at the time of PR submission.

rljacob:
Please edit the description then.  The first sentence says "Add compsets used in v2 BGC simulations to maint-2.1 for better provenance and reproducibility."

evasinha:
@rljacob I have edit the description to state "Add compsets used in v2 BGC simulations to **master and maint-2.1** for better provenance and reproducibility."

rljacob:
But this PR is only adding it to master.  I fixed it.

PR #6420, opened 2024-05-15
modify sss restoring to include under sea ice by default

alicebarthel:
Work done under the ImPACTS SciDac led to the conclusion that it is generally preferable to include SSS restoring under sea ice by default if doing any SSS restoring, and to compute the restoring terms at each time step. 

This is due to the fact that the intended linear taper in the sea ice region (by sea ice fraction) may change sign entirely due to the removal of the global mean. This leads to a small negative flux in the high-latitudes, which reinforces the fresh bias. 

Work by Luke, Mat, Fred, Alice and others. 

[NML]
[non-BFB] for C- and G-cases using grids with restoring files

---------

Plots made by Fred:
Screen Shot 2024-05-15 at 2 20 42 PM
Of particular concern is the small negative value in the Arctic for example.


vanroekel (approved):
approving by visual inspection.  and nice find on the `dt` option that's handy.

Thanks @alicebarthel!

alicebarthel:
- passes `ERS_Ld3.ne30pg2_r05_IcoswISC30E3r5.WCYCL1850.pm-cpu_intel.allactive-nlmaps` (BFB)
- also passes `PEM_Ln5.T62_oQU240wLI.GMPAS-DIB-IAF-DISMF.pm-cpu_intel` (BFB) since `config_use_surface_salinity_monthly_restoring = .false.` in this test. The relevant flags in `mpaso_in` are changed as expected. 
- A short (`S_1x5_ndays`) `GMPAS-IAF` effectively switches the flags in `mpaso_in` on Perlmutter. I am running the same short run for `master` to do a manual sum check on the logs, to confirm that a standard G-case will *not* be BFB.

alicebarthel:
Per @jonbob's recommendation, I ran `SMS_D_Ld1.TL319_IcoswISC30E3r5.GMPAS-JRA1p5-DIB-PISMF.pm-cpu_intel.mpaso-jra_1958`. 
As expected, the G-case with SSS restoring leads to a non BFB change with the baseline (`master` as of today):
```
 A total number of    396 fields were compared
          of which    103 had non-zero differences
...
  diff_test: the two files seem to be DIFFERENT
```
I am happy with the behavior, this is ready for review. @jonbob let me know if you'd like me to check anything else.

jonbob:
@alicebarthel -- I think that's good. It proves this PR is having the impact you would expect. It might be worth comparing a longer run with this change vs a baseline, just to know what impact it has.

xylar:
We expect this to be non-BFB for G-case simulations with salinity restoring (which we use for all scientifically validated configurations).  So I have added the `non-BFB` tag.

xylar:
I think @alicebarthel is either already away or will be soon.  We will probably need someone else to run the longer G-case that @jonbob requested.  @vanroekel, do you have someone you can suggest?  If not, I can do it.

vanroekel:
@xylar -- @alicebarthel did run a long G-case with this change, data is here on lcrc
```
/lcrc/group/e3sm/ac.abarthel/E3SMv3/20240305..GMPAS-JRA.sssRestorindDt.underSeaIce.TL319_IcoswISC30E3r5.anvil/run
```
and mpas-analysis here -- https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.abarthel/E3SMv3/GMPAS.sssRestoring_yr1-5/

But digging around a bit, I'm not sure we have a corresponding baseline to compare against.  @alicebarthel -- would Filipe's ICOS case be a reasonable baseline?

xylar:
Yeah, I think we need that baseline.

vanroekel:
I just asked Filipe about his ICOS run to see if it can be an appropriate baseline.  I'll report back when I hear from him

alicebarthel:
@vanroekel technically the test you are pointing at does not test the exact code base in question. @xylar i'm still here and was planning on running 2 10-year simulations and posting a MPAS_Analysis comparison as @jonbob instructed. Happy to do that if that is still deemed appropriate.

xylar:
@alicebarthel, if you're able to do that, that would be wonderful!

alicebarthel:
The 10-year G-case tests are here: 
master baseline `/pscratch/sd/a/abarthel/e3sm_scratch/pm-cpu/20240603b.master.GMPAS-JRA.TL319_IcoswISC30E3r5.pm-cpu`
PR modifications `/pscratch/sd/a/abarthel/e3sm_scratch/pm-cpu/20240603.prtest.GMPAS-JRA.TL319_IcoswISC30E3r5.pm-cpu`

As expected, the differences to mpaso_in files are only the 2 relevant config options: `config_salinity_restoring_under_sea_ice` and `config_surface_salinity_monthly_restoring_compute_interval`

The MPAS-Analysis comparison is  [here](https://portal.nersc.gov/cfs/e3sm/abarthel/E3SMv3/sssPR.testvsbaseline/). As expected, the first 10 years show an increase in surface/near-surface salinity ([SSS](https://portal.nersc.gov/cfs/e3sm/abarthel/E3SMv3/sssPR.testvsbaseline/ocean/sss_v3.GMPAS-JRA1p5.prSSS.modif.pm_ANN_years0001-0010.png),[ 50m S](https://portal.nersc.gov/cfs/e3sm/abarthel/E3SMv3/sssPR.testvsbaseline/ocean/salin_v3.GMPAS-JRA1p5.prSSS.modif.pm_depth_-50_ANN_years0001-0010.png)), thus increasing [MLD](https://portal.nersc.gov/cfs/e3sm/abarthel/E3SMv3/sssPR.testvsbaseline/ocean/mld_v3.GMPAS-JRA1p5.prSSS.modif.pm_ANN_years0001-0010.png) at high latitudes. Small to no changes to temperature (e.g. [SST](https://portal.nersc.gov/cfs/e3sm/abarthel/E3SMv3/sssPR.testvsbaseline/ocean/sst_v3.GMPAS-JRA1p5.prSSS.modif.pm_ANN_years0001-0010.png), [100m T](https://portal.nersc.gov/cfs/e3sm/abarthel/E3SMv3/sssPR.testvsbaseline/ocean/temp_v3.GMPAS-JRA1p5.prSSS.modif.pm_depth_-100_ANN_years0001-0010.png)), or [sea ice concentration](https://portal.nersc.gov/cfs/e3sm/abarthel/E3SMv3/sssPR.testvsbaseline/sea_ice/iceconcNH_arctic_extended_v3.GMPAS-JRA1p5.prSSS.modif.pm_JAS_years0001-0010.png), with small changes likely to indirect effects. 
Regional changes to high latitudes include increased salinity (e.g. [SO](https://portal.nersc.gov/cfs/e3sm/abarthel/E3SMv3/sssPR.testvsbaseline/ocean/AntarcticRegions_SouthernOcean60S_salinity.png), [Canada Basin](https://portal.nersc.gov/cfs/e3sm/abarthel/E3SMv3/sssPR.testvsbaseline/ocean/ArcticOceanRegions_CanadaBasin_salinity.png)) and cooler temperatures ([SO](https://portal.nersc.gov/cfs/e3sm/abarthel/E3SMv3/sssPR.testvsbaseline/ocean/AntarcticRegions_SouthernOcean60S_temperature.png), [Canada Basin](https://portal.nersc.gov/cfs/e3sm/abarthel/E3SMv3/sssPR.testvsbaseline/ocean/ArcticOceanRegions_CanadaBasin_temperature.png)) (latter likely due to indirect effects). 
The [AMOC timeseries](https://portal.nersc.gov/cfs/e3sm/abarthel/E3SMv3/sssPR.testvsbaseline/ocean/mocTimeseries_v3.GMPAS-JRA1p5.prSSS.modif.pm.png) and[ OHU](https://portal.nersc.gov/cfs/e3sm/abarthel/E3SMv3/sssPR.testvsbaseline/ocean/deltaOHC_0-700m_v3.GMPAS-JRA1p5.prSSS.modif.pm_ANN_years0001-0010.png) are unchanged. [1Sv increase to Drake Passage transport](https://portal.nersc.gov/cfs/e3sm/abarthel/E3SMv3/sssPR.testvsbaseline/ocean/transport_Drake_Passage.png). 

I did not see anything unexpected or concerning. Have a look if you can. Thanks!

xylar:
@alicebarthel, did the analysis not succeed?  Or do you just need to change permissions?  I'm not seeing the `index.html` at https://portal.nersc.gov/cfs/e3sm/abarthel/E3SMv3/sssPR.testvsbaseline/

xylar (approved):
Approved based on @alicebarthel's testing and analysis output and an inspection of the code changes.

xylar:
Nice, this looks like it's doing what we want and either improving biases or doing no harm.

components/mpas-ocean/bld/namelist_files/namelist_definition_mpaso.xml
```diff
@@ -2759,7 +2759,7 @@ Default: Defined in namelist_defaults.xml
 
 <entry id="config_salinity_restoring_under_sea_ice" type="logical"
 	category="tracer_forcing_activeTracers" group="tracer_forcing_activeTracers">
-Flag to enable salinity restoring under sea ice.  The default setting is false, where salinity restoring tapers from full restoring in the open ocean (iceFraction=0.0) to zero restoring below full sea ice coverage (iceFraction=1.0); below partial sea ice coverage, restoring is in proportion to iceFraction.  If true, full salinity restoring is used everywhere, regardless of iceFraction value
+Flag to enable salinity restoring under sea ice.  The default setting is true, meaning that full salinity restoring is used everywhere, regardless of iceFraction value, except under ice shelves. If false, the salinity restoring tapers from full restoring in the open ocean (iceFraction=0.0) to zero restoring below full sea ice coverage (iceFraction=1.0); below partial sea ice coverage, restoring is in proportion to iceFraction.
```
jonbob, line 2762: @alicebarthel - I think this should be false to be consistent with other restoring flags. They get set to true in appropriate configurations elsewhere in the scripts.

vanroekel: @jonbob can you explain why this should be false?  This doesn't affect simulations unless `config_use_surface_salinity_monthly_restoring` is set to true.  And I think this is only turned on for G compsets.  Any time salinity restoring is on, we want this flag true.  Do you prefer we have this false too and turn to true when G cases are done with the other flag?

jonbob: @vanroekel -- that is what we have been doing, if only to make it clear in fully-coupled cases that there is no restoring. When the scripts know that we're using datm forcing, they do this:
````
if ($OCN_FORCING eq 'datm_forced_restoring') {
	add_default($nl, 'config_use_activeTracers_surface_restoring', 'val'=>".true.");
	add_default($nl, 'config_use_surface_salinity_monthly_restoring', 'val'=>".true.");
	add_default($nl, 'config_salinity_restoring_constant_piston_velocity', 'val'=>"1.585e-6");
	add_default($nl, 'config_salinity_restoring_max_difference', 'val'=>"100.");
	add_default($nl, 'config_salinity_restoring_under_sea_ice', 'val'=>".false.");
````
(except this PR changes that last line to true). It's not a big deal, but I would have a little concern that having this true by default for all configurations could be confusing.

xylar: I agree that @jonbob's suggestion is clearer.

vanroekel: I see now, makes sense to me and an easy fix.  Thanks for the explanation

jonbob: OK, I've updated the default and think this is good to go

rljacob:
Does this only change answers for G-cases ?

jonbob:
@rljacob -- yes, and C-cases as well. And only for grids that have restoring files, so not most of the testing ones like oQU480 and oQU240

jonbob:
Passes:
* PEM_Ln5.T62_oQU240wLI.GMPAS-DIB-IAF-DISMF.chrysalis_intel

with expected NML DIFFs.

Regular DIFF for:

as expected. Merged to next

jonbob:
merged to master and expected NML and regular DIFFs blessed, except:
* pm-cpu integration next with intel (hasn't reported yet today)
* mappy next (hasn't reported and may be having issues)

jonbob:
also blessed on pm-cpu integration next with intel

jonbob:
blessed on happy as well

Issue #6422, opened 2024-05-16
`(GTL DEBUG: 0) cuIpcGetMemHandle: invalid argument, CUDA_ERROR_INVALID_VALUE` on polaris


ndkeen:
After https://github.com/E3SM-Project/E3SM/pull/6101 which brings in kokkos 4.2, we see runtime error with a test like:
`ERP_Ln9.ne4pg2_ne4pg2.F2010-SCREAMv1.pm-gpu_gnugpu`
(also on similar machine `ERP_Ln9.ne4pg2_ne4pg2.F2010-SCREAMv1.muller-gpu_gnugpu`)

```
0: (GTL DEBUG: 0) cuIpcGetMemHandle: invalid argument, CUDA_ERROR_INVALID_VALUE, line no 97
```

A fix that seems to work is to add this build flag:
`Kokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC=OFF`

This will also be needed for scream repo.
@bartgol @mahf708

bartgol:
I just opened a quick follow up PR, to clean up some deprecated code issues (they did not cause falls in e3sm bc deprecated code in looks was allowed).

I can add a fix for this in that PR.

ndkeen:
Fine with me.  Just include the change to make the muller-gpu file same as pm-gpu.

gsever:
Just to note that the same issue shows up running a recent E3SM/SCREAM version on ALCF's Polaris:

`(GTL DEBUG: 0) cuIpcGetMemHandle: invalid argument, CUDA_ERROR_INVALID_VALUE, line no 148 `

Adding `-DKokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC=OFF` in `./cime_config/machines/cmake_macros/gnugpu_polaris.cmake` fixes the runtime error as posted here.

ndkeen:
ok to close?  or create another issue specific to polaris?

PR #6101, opened 2023-12-05
Update ekat to a version that has Kokkos 4.2 as submodule

bartgol:
This PR will take time to integrate, I'm opening it so I can keep track of what I check.
- [x] e3sm_integration:
	- [x] chrysalis (intel): all PASS
	- [x] pm-cpu (intel) 127 PASS and 1 DIFF: SMS_Ld2.ne30pg2_r05_IcoswISC30E3r5.BGCEXP_CNTL_CNPECACNT_1850.pm-cpu_intel.elm-bgcexp (which is currently consistently failing in the  `e3sm_integration_next_intel` nightly build)
- [x] e3sm_developer:
    - [x] pm-cpu (gnu): 75PASS and 1 DIFF: ERP_Ld3.ne4pg2_oQU480.F2010.pm-cpu_gnu
- [x] homme_integration
	- [x] chrysalis (intel): PASS
	- [x] pm-cpu (gnu):  builds PASS, stuck in Q for run, so I cancelled it. pm-cpu is not tested in nightlies anyways
- [x] eamxx testing (from eamxx repo, with a few additional commits for eamxx)
    - [x] v1 (CIME)
      	- [x] chrysalis (intel): 10 PASS (all scream v1) 5 DIFF (all scream v0)
		- [x] pm-cpu (gnu): 3 PASS, 1 DIFF
		- [x] frontier PEND
		- [x] ~ascent~: no longer part of eamxx nightlies
	   - [x] pm-gpu (gnugpu): 7 PASS, 5 DIFF. All DIFF are in debug mode, while all non-debug builds pass. I'm trying to understand what's the catch.
	- [x] standalone
		- [x] mappy (gnu): all PASS
		- [x] weaver (gnu+cuda): all PASS

@ambrad @oksanaguba can you think of any more machine/testsuite I should run?

PR #6421, opened 2024-05-16
Point to ekat master branch, and fix usage of deprecated kokkos code in Homme

bartgol:
We just merged the kokkos-4.2 branch in ekat/master, so now we can make e3sm point directly to the ekat/master branch, which avoids conflicts when doing a downstream merge into scream repo.

Also, fix usage of deprecated kokkos code in Homme. The tests on e3sm repo are set so that Kokkos allows using deprecated code (with warnings), but it caused some errors downstream in eamxx. Using the non-deprecated code is a simple fix.

PR #6423, opened 2024-05-16
For muller-gpu, add `-DKokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC=OFF`

ndkeen:
After https://github.com/E3SM-Project/E3SM/pull/6101 which brings in kokkos 4.2, we see runtime error with a test like:
`ERP_Ln9.ne4pg2_ne4pg2.F2010-SCREAMv1.pm-gpu_gnugpu`

hits runtime error like:
```
0: (GTL DEBUG: 0) cuIpcGetMemHandle: invalid argument, CUDA_ERROR_INVALID_VALUE, line no 97
```
unfortunately, the tests hitting this error are also hanging...

A fix that seems to work is to add this build flag:
`-DKokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC=OFF`

The fix is merged in another PR for pm-gpu, so this PR just makes same change to muller-gpu.

Fixes https://github.com/E3SM-Project/E3SM/issues/6422


bartgol (approved):
I added a similar commit to #6421 . I don't really care which PR this fix is attached to. That one is already in next though. Either way, I'm approving, just in case that one has issues.

ndkeen:
it doesnt look like you made the change to muller-gpu?

ndkeen:
merged to next

Issue #6425, opened 2024-05-17
Annoying warning messages written to log `MPIIO WARNING: DVS stripe width of 2 was requested but DVS set it to 1` on perlmutter


ndkeen:
For a while, we've seen these annoying messages written to the e3sm.log files. For one case, I count ~140 of them that take the form:

```
 1: MPIIO WARNING: DVS stripe width of 2 was requested but DVS set it to 1
 1: See MPICH_MPIIO_DVS_MAXNODES in the intro_mpi man page.
```

I asked Lisa G at NERSC about this and she asked me if we are setting `MPICH_MPIIO_DVS_MAXNODES`.
Grepping globally in our repo, I don't see it.
But sure enough, it is being set on the compute nodes.
Setting this to 1 seems to rid us of the warning.

PR #6426, opened 2024-05-17
For NERSC machines, add env var `MPICH_MPIIO_DVS_MAXNODES=1` to avoid runtime warnings

ndkeen:
Add env variable to NERSC machines (such as `pm-cpu` and `pm-gpu`) that avoids several runtime warning messages.

Fixes https://github.com/E3SM-Project/E3SM/issues/6425

[bfb]


xylar:
> Making the logs clean coal!

 🤨

PR #6431, opened 2024-05-18
Fix performance tests on anvil and pm-cpu

amametjanov:
Fix performance tests on anvil and pm-cpu:
- fix SGLC regex patterns to revert to correct perf-test PEs on Anvil
- avoid memleak flags in heavy-IO perf-tests on pm-cpu

[BFB]

### DIFF
--- a/cime_config/allactive/config_pesall.xml
+++ b/cime_config/allactive/config_pesall.xml
@@ -325,7 +325,7 @@
       </pes>
     </mach>
     <mach name="anvil">
-      <pes compset="JRA_ELM.+MPASSI.+MPASO.+MOSART+SGLC" pesize="any">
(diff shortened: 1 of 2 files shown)


ndkeen:
I was wondering how that incorrect regex pattern made it into the repo with recent PR -- but why do we even need SGLC string? Is it trying to differentiate between other compsets that don't use SGLC?  If not, seems like we can just remove it (ie back to the way it was)

amametjanov:
Yes, I think it was added to differentiate between stub SGLC and active MALI components: e.g. `MOSART_SGLC_SWAV` and `MOSART_MALI_SWAV` patterns in E3SM-Project/E3SM#6226.
I think it's good to have more specific regex patterns.

ndkeen:
OK that's fine if it was done to differentiate between res/compsets. I'm now waiting on this PR to make some unrelated changes.

ndkeen (approved):

PR #6433, opened 2024-05-19
Enable flexible definitions of plant functional types (PFTs)

thorntonpe:
Introduce a new optional format for PFT physiology files (paramfile) that 
allows user to specify a flexible number of PFTs, including additional
natural vegetation types and different numbers of crops. Backwards compatible
(BFB) with older style files.

New variables added in new format file:
- climatezone (options are: 0 {nonspecific}, 1 {tropical}, 2 {temperate}, 3 {boreal}, 4 {arctic})
- needleleaf (options are: 0, 1)
- nfixer (options are: 0, 1)
- nonvascular (options are: 0 {vascular}, 1 {moss}, 2 {lichen})
- graminoid (options are: 0 {non-grass}, 1 {grass})
- iscft (options are: 0 {natural PFT}, 1 {prognostic crop functional type})

Variables with expanded options in new format file:
- woody (options are: 0{non-woody}, 1 {tree}, 2 {shrub})

Add 2 new tests for the user pft.
ERS.ELM_USRDAT.I1850CNPRDCTCBC.elm-usrpft_default_I1850CNPRDCTCBC
ERS.ELM_USRDAT.I1850CNPRDCTCBC.elm-usrpft_codetest_I1850CNPRDCTCBC

Thanks-to: Fengming Yuan (yuanf@ornl.gov), Ben Sulman (sulmanbn@ornl.gov)
[BFB]


thorntonpe:
This PR is passing all e3sm_land_developer tests on chrysalis against master baselines at hash dd32f0.
Two new tests are added to e3sm_land_developer suite. Those tests require a new paramfile and new datm7 files. I have copied those into the inputdata directory on chrysalis.

thorntonpe:
There is a circleci: build failure for this PR, but I'm not able to interpret the details. It looks like a timeout problem.

thorntonpe:
The new feature documentation page for this PR is on the Land group space on Confluence, at:
https://acme-climate.atlassian.net/wiki/spaces/LAND/pages/4335861769/Arctic+PFTs+NGEE+Arctic+IM4

thorntonpe:
@bishtgautam @bbye @rgknox - I'm hoping you can take a look at this PR and give me any feedback. Beth, this is the modification for flexible number and order of PFTs that I discussed earlier with you and Eva. Ryan, glad to get together to give you additional background if needed. This is part of the E3SM - NGEE Arctic coordination Epic. Thanks, Peter

rgknox:
ok, I'll take a look @thorntonpe

rljacob:
@bbye please review.

thorntonpe:
@rgknox Based on our code review discussion on June 12, the ORNL NGEE Arctic group is planning to clean up the conditional deallocation / allocation at lines 297-306 in elm_initializeMod.F90, which also gets rid of a hard-coded mxpft_nc /= 24 reference. The new commit for those changes will be ready on this PR early next week. 

We also discussed with you some confusing use of numpft, mxpft, maxpatch_pft variables in elm_varpar.F90 and elsewhere. After giving that more careful review, we think this is something that is best tackled as an E3SM/land group code cleanup issue. The NGEE Arctic team will open an issue related to that with some suggestions about how the E3SM team can modify the use of these variables.

components/elm/src/biogeochem/AllocationMod.F90
```diff
@@ -2748,7 +2748,7 @@ subroutine Allocation3_PlantCNPAlloc (bounds            , &
                  npool_to_frootn(p) = cpool_to_frootc(p) / cnfr
                  npool_to_frootn_storage(p) = cpool_to_frootc_storage(p) / cnfr
 
-                 if (woody(ivt(p)) == 1._r8) then
+                 if (woody(ivt(p)) >= 1.0_r8) then
```
rgknox, line 2751: is woody a real?

components/elm/src/biogeochem/CH4Mod.F90
```diff
@@ -2563,8 +2563,7 @@ subroutine ch4_aere (bounds, &
                   is_vegetated = .false.
                end if
 
-               if (veg_pp%itype(p) == nc3_arctic_grass .or. crop(veg_pp%itype(p)) == 1 .or. &
-                    veg_pp%itype(p) == nc3_nonarctic_grass .or. veg_pp%itype(p) == nc4_grass) then
+               if (graminoid(veg_pp%itype(p)) == 1 .or. crop(veg_pp%itype(p)) == 1) then
```
rgknox, line 2566: same as comments related to comparisons with woody, can we compare this to a named constant instead of 1?  Even if this is a integer acting as a logical, or a true logical, its better to compare to a named consant, like int_true or int_false.

components/elm/src/biogeochem/CNNitrogenFluxType.F90
```diff
@@ -1078,7 +1078,7 @@ subroutine Summary(this, bounds, num_soilc, filter_soilc, num_soilp, filter_soil
     ! !USES:
     use elm_varpar    , only: nlevdecomp,ndecomp_cascade_transitions,ndecomp_pools
     use subgridAveMod , only: p2c
-    use pftvarcon     , only : npcropmin 
+    use pftvarcon     , only : iscft
```
rgknox, line 1081: this can be removed since its not used, right?

components/elm/src/biogeochem/CNPhenologyBeTRMod.F90
```diff
@@ -1893,13 +1894,17 @@ subroutine CropPhenologyInit(bounds)
     ! Convert planting dates into julian day
     minplantjday(:,:) = huge(1)
     maxplantjday(:,:) = huge(1)
-    do n = npcropmin, npcropmax
```
rgknox, line 1896: I see, so this is an example of added flexibility to where cfts are positioned in the list, we no longer rely on looping through the expected cft bounds, but go through all fts and check their iscft label...

rgknox: Is it possible to remove the usage of npcropmin and npcropmax altogether, and remove it from pftvarcon?

components/elm/src/biogeochem/CNCarbonStateType.F90
```diff
@@ -459,7 +459,7 @@ subroutine InitCold(this, bounds, ratio, c12_carbonstate_vars)
                 if (veg_vp%evergreen(veg_pp%itype(p)) == 1._r8) then
                    this%leafc_patch(p)         = 1._r8 * ratio
                    this%leafc_storage_patch(p) = 0._r8
-                else if (veg_pp%itype(p) >= npcropmin) then ! prognostic crop types
+                else if (iscft(veg_pp%itype(p)) >= 1) then ! prognostic crop types
```
rgknox, line 462: the definition of iscft in pftvarcon only covers two types 0 and 1, is this ">=" prepping for a future condition?  If the meaning of iscft expands to be non-binary, we should update the description in pftvarcon when this happens.

components/elm/src/biogeochem/VOCEmissionMod.F90
```diff
@@ -114,7 +110,7 @@ subroutine InitAllocate(this, bounds)
     ! !LOCAL VARIABLES:
     integer            :: i, imeg
     integer            :: class_num
-    real(r8)           :: factors(numpft)
+    real(r8)           :: factors(numveg)
```
rgknox, line 113: is this a bug fix? The allocation space on this changed, but its usage did not.

components/elm/src/biogeophys/CanopyFluxesMod.F90
```diff
@@ -869,7 +869,8 @@ subroutine CanopyFluxes(bounds,  num_nolakeurbanp, filter_nolakeurbanp, &
             p = filterp(f)
             c = veg_pp%column(p)
             if(.not.veg_pp%is_fates(p)) then
-               if (veg_pp%itype(p) == nsoybean .or. veg_pp%itype(p) == nsoybeanirrig) then
+               ! soybean (crop with N fixation)
+               if (crop(veg_pp%itype(p)) >= 1 .and. nfixer(veg_pp%itype(p)) == 1) then
```
rgknox, line 873: crop is still defined as a binary in pftvarcon:
https://github.com/E3SM-Project/E3SM/blob/6ced7eb263abd7be11a1edba1eb0049e5e0cfb6b/components/elm/src/main/pftvarcon.F90#L106

Can you update the definition since this can be larger than 1?

components/elm/src/main/elm_varpar.F90
```diff
@@ -48,20 +48,23 @@ module elm_varpar
   integer, parameter :: ndst        =   4     ! number of dust size classes (BGC only)
   integer, parameter :: dst_src_nbr =   3     ! number of size distns in src soil (BGC only)
   integer, parameter :: sz_nbr      = 200     ! number of sub-grid bins in large bin of dust size distribution (BGC only)
-  integer, parameter :: mxpft       =  50     ! maximum number of PFT's for any mode;
+  !integer, parameter :: mxpft       =  50     ! maximum number of PFT's for any mode;
+  integer :: mxpft                  =  50     ! maximum number of PFT's for any mode;
```
rgknox, line 52: shouldn't maxpft be a parameter constant?

components/elm/src/main/elm_varpar.F90
```diff
@@ -48,20 +48,23 @@ module elm_varpar
...
+                                              ! can be modified from reading pft-physiology in 'main/pftvarcon.F90:pftconrd'
   ! FIX(RF,032414) might we set some of these automatically from reading pft-physiology?
   integer, parameter :: numveg      =  16     ! number of veg types (without specific crop)
   integer, parameter :: nlayer      =   3     ! number of VIC soil layer --Added by AWang
   integer            :: nlayert               ! number of VIC soil layer + 3 lower thermal layers
 
-  integer :: numpft      = mxpft   ! actual # of patches (without bare), a total that spans LUs
+  integer :: numpft      =  50     ! actual # of patches (without bare), a total that spans LUs
```
rgknox, line 59: Might be safer to initialize this to a nonsense number (a negative perhaps?), so that it crashes if there is some weird scenario where it does not get set, is it interpreted from reading in real data, right?

components/elm/src/main/surfrdUtilsMod.F90
```diff
@@ -123,34 +123,35 @@ subroutine convert_cft_to_pft( begg, endg, cftsize, wt_cft )
     !        a crop landunit, and put them on the vegetated landunit.
     ! !USES:
```
rgknox, line 124: I was grepping through the code to see where we use this routine convert_cft_to_pft, to see if the changes would effect FATES somehow.  But I don't see where this is called.  Is it actually called anywhere?

rgknox:
I gave a look through the code, somewhere between "thorough" and "almost thorough".  I think I came up with some reasonable (and perhaps nit-picky) things to look over.  In general, my take is that the code uses too many "magic numbers", ie comparing variables to hard-coded constants like "1"  or "0".  I feel it is fair to say that fixing that type of thing is out of scope for this PR, but it could be an opportunity to make things more transparent if you agree with me.

In general though, this PR is getting the code more flexible and robust since it breaks away from expecting an order to certain indices.  Awesome stuff all.

thorntonpe:
Thanks @rgknox for giving this a careful look. I agree with your suggestions, and I'll discuss with @fmyuan and @bsulman. They are working on an additional commit, and we should be able to get at least some of your requested changes wrapped into this PR.

components/elm/src/main/pftvarcon.F90
```diff
@@ -1103,12 +1223,209 @@ subroutine pftconrd
...
+                 ! if 'generic crop' (crop=1) specifically flagged by iscft=0
+                 ! 'crop' will not be counted into prognostic
+                 if (crop(i)>=1 .and. iscft(i)==1) then
+                    npcropmax = i
+                    if(npcropmin<=0) npcropmin = i
+                 end if
+
+                 ! NOTE: there is a misunderstanding in pft physiology parameter file: 'perennial crop' IS NOT 'crop',
+                 !       but still counted into 'numcft'
+                 if(percrop(i)==1) then
```
bbye, line 1289: Does this mean that percrop has an iscft ==0?

components/elm/src/biogeochem/DryDepVelocity.F90
```diff
@@ -275,24 +270,31 @@ subroutine depvel_compute( bounds, &
...
+            if (woody(elmveg) == 2.0_r8                              ) wesveg = 11
+            !if (elmveg == nc3_arctic_grass                    ) wesveg = 3
+            !if (elmveg == nc3_nonarctic_grass                 ) wesveg = 3
+            !if (elmveg == nc4_grass                           ) wesveg = 3
+            if (graminoid(elmveg) == 1.0_r8                          ) wesveg = 3
+            if (nonvascular(elmveg) == 1.0_r8                        ) wesveg = 3      ! assuming moss like grass
+            !if (elmveg == nc3crop                             ) wesveg = 2
+            !if (elmveg == nc3irrig                            ) wesveg = 2
+            !if (elmveg >= npcropmin .and. elmveg <= npcropmax ) wesveg = 2
+            if (crop(elmveg) == 1.0_r8 .or. iscft(elmveg) == 1.0_r8  ) wesveg = 2
```
bbye, line 297: Here and a few other places iscft ==1 but almost everywhere else, the use is iscft>=1? I think it would be good to be consistent with the notation.

components/elm/src/biogeochem/FireMod.F90
```diff
@@ -305,11 +305,12 @@ subroutine FireArea (bounds, &
            if (pi <=  col_pp%npfts(c)) then
               p = col_pp%pfti(c) + pi - 1
               ! For crop veg types
-              if( veg_pp%itype(p) > nc4_grass )then
+              if( crop(veg_pp%itype(p)) == 1 .or. iscft(veg_pp%itype(p)) == 1 )then
```
bbye, line 308: Another location where == is used instead of >= for iscft. Also see lines 335 and 511 in this subroutine.

components/elm/src/biogeochem/VOCEmissionMod.F90
```diff
@@ -693,21 +690,20 @@ function get_map_EF(ivt_in, g_in, ti_in, vocemis_vars)
...
-    else if (ivt_in >= nbrdlf_evr_shrub &
-         .and.    ivt_in <= nbrdlf_dcd_brl_shrub) then !shrubs
+    else if (woody(ivt_in) == 2.0_r8) then !shrubs
        get_map_EF = vocemis_vars%efisop_grc(4,g_in, ti_in)
-    else if (ivt_in >= nc3_arctic_grass &
-         .and.    ivt_in <= nc4_grass) then            !grass
+    else if (graminoid(ivt_in) == 1) then            !grass
        get_map_EF = vocemis_vars%efisop_grc(5,g_in, ti_in)
-    else if (ivt_in >= nc3crop) then                   !crops
+    else if (crop(ivt_in) == 1 .or. iscft(ivt_in) == 1) then !crops
```
bbye, line 706: uses == instead of >= for iscft

components/elm/src/biogeophys/CanopyFluxesMod.F90
bbye, line 873: I'm a little worried about this change. This code is specific to soybean since it is drought tolerant. I worry if using an nfixer flag any future crops that are classified as nfixers would inadvertently be included in this call. Currently we don't have any other nfixers, but I don't know if that will change.

components/elm/src/main/pftvarcon.F90
```diff
@@ -1103,12 +1223,209 @@ subroutine pftconrd
...
+             else if (woody(i) >= 1 .and. (nonvascular(i)+graminoid(i)+max(iscft(i),crop(i)+percrop(i))) >= 1) then
+                print *, 'ERROR: Incorrect woody PFT flags: ', i, ' ', trim(pftname(i))
+                call endrun(msg=' ERROR: woody PFT cannot be any of nonvascular/graminoid/crop type - '//errMsg(__FILE__, __LINE__))
+             else if (graminoid(i) >= 1 .and. (nonvascular(i)+woody(i)+max(iscft(i),crop(i)+percrop(i))) >=1 ) then
+                print *, 'ERROR: Incorrect graminoid PFT flags: ', i, ' ', trim(pftname(i))
+                call endrun(msg=' ERROR: graminoid PFT cannot be any of nonvascular/woody/crop type - '//errMsg(__FILE__, __LINE__))
+             else if ( (max(iscft(i),crop(i)+percrop(i))) >= 1 .and. (nonvascular(i)+woody(i)+graminoid(i)) >= 1) then
+                print *, 'ERROR: Incorrect crop PFT flags: ', i, ' ', trim(pftname(i))
+                call endrun(msg=' ERROR: crop PFT cannot be any of nonvascular/woody/graminoid type - '//errMsg(__FILE__, __LINE__))
+             end if
```
bbye, line 1388: We have some woody crops targeted for future versions of the model, so we might want more flexibility with the flags.

bbye:
I added some comments that I saw during my review. A couple other things I wanted to bring up. 1) it might be nice for @evasinha to look at the PR since there are a lot of changes to the crop indexing and she knows more about how perennial crops are handled. 2) the PR should be tested with the crop model configuration (including a test with 2-way irrigation using generic crops) to make sure it doesn't break anything. 3) I noticed that iscft is either 0 or 1, but most of the if statements used >=1. I think this is ok since it gives us freedom to later expand the crops a bit. I'm thinking we could have annual crops=1 and perennial crops=2. I'm not asking for that change in this PR, but keeping consistent use of >=1 for iscft would allow us to change that in the future.

components/elm/src/biogeochem/AllocationMod.F90
thorntonpe: Yes.

components/elm/src/biogeochem/CNNitrogenFluxType.F90
thorntonpe: Yes - removed in latest commit.

components/elm/src/biogeochem/AllocationMod.F90
```diff
@@ -672,7 +672,7 @@ subroutine Allocation1_PlantNPDemand (bounds, num_soilc, filter_soilc, num_soilp
 
          f5 = 0._r8 ! continued intializations from above
 
-         if (ivt(p) >= npcropmin) then ! skip 2 generic crops
+         if (iscft(ivt(p)) >= 1) then ! skip 2 generic crops
```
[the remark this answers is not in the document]
thorntonpe: @rgknox there are a number of flags that are currently reals but are intended as logical (usually named is<something>). In the spirit of getting this set up for a broader overhaul later on, I used iscft as an example and did the conversion throughout the code to a logical variable in the latest commit. That allows these kinds of tests to be of the form if (iscft(ivt(p))) then ..., or if (.not. iscft(ivt(p))) then ...   
That seems like a clear approach for the logical flags. The use of multiple values for a flag like woody is going to take some additional thought and work, since it is sometimes used like if (woody > 1), which does not lend itself any better to a named value replacing the 1.
Hopefully this demonstration with iscft shows an acceptable path for future development, while allowing this PR to move ahead.

components/elm/src/biogeochem/CNPhenologyBeTRMod.F90
thorntonpe: That is something we can try to tackle in later PRs, focusing on code cleanup.

components/elm/src/biogeochem/CNCarbonStateType.F90
thorntonpe: Nope, this was intended as binary. Now converted to that throughout the code. Thanks for the suggestion.

components/elm/src/biogeochem/VOCEmissionMod.F90
thorntonpe: Full range was not used previously.

components/elm/src/biogeophys/CanopyFluxesMod.F90
thorntonpe: No plan for this to be greater than 1. This can be changed following the example of iscft, but would prefer to get it in a later PR.

components/elm/src/main/elm_varpar.F90
thorntonpe: mxpft is now updated by reading the pft physiology file.

components/elm/src/main/elm_varpar.F90
thorntonpe: mxpft is set by reading in real data. numpft is still used as an upper limit. We are working out how to simplify this but I'd like to do it in a later PR - it is a long-standing tangle of usage that this PR makes a little better, but doesn't go all the way to cleaning up.

components/elm/src/main/surfrdUtilsMod.F90
thorntonpe: It isn't called anywhere in the current code. It is a placeholder for future functionality.

components/elm/src/biogeochem/DryDepVelocity.F90
thorntonpe: Thanks -based on this and Ryan's comments I converted the iscft to a logical type in the latest commit.

components/elm/src/biogeochem/FireMod.F90
thorntonpe: I would like to go through and do the conversion to logical types for a couple of other variables, including crop(), in the same way the I have done for iscft(). I would like to hold off and bring that in with other code cleanup in another PR.

components/elm/src/main/pftvarcon.F90
thorntonpe: These error checks are very specific for the current usage of woody etc. I agree something more flexible would be good. Let's work on that together as a future PR.

components/elm/src/biogeophys/CanopyFluxesMod.F90
thorntonpe: Good point. This suggests that what we want is a drought-tolerant flag for crops, or better yet a change to the photosynthesis parameters controlling btran. Let's put this on our list for testing and changes in a future PR.

components/elm/src/main/pftvarcon.F90
thorntonpe: No, percrop and iscft are independent.

thorntonpe:
@rgknox and @bbye - based on latest two commits, and responses to your reviews, I'm hoping this PR can now be approved.

rgknox (approved):

rgknox:
done, thanks @thorntonpe

rljacob:
@bbye please re-review.

components/elm/src/main/elm_varpar.F90
bbye: Do all the physiology files need to be updated to read mxpft?

bbye:
I only had one comment/question about whether we need to update all the physiology files since mxpft is now being read in. I know this is backward compatible, I just wasn't sure about that. Otherwise, looks fine to merge.

thorntonpe:
@bbye Thanks for the further review. The mxpft is being calculated by reading the entries in the file, and there is no requirement to add it as a variable or modify the existing physiology files.

bbye (approved):

rljacob:
Merged to next after being removed by a reset of next.

PR #6434, opened 2024-05-21
Change MPAS-Ocean high-frequency output mode to `append`

### LINKED ISSUE #6432: Potential bug in MPAS ocean output
I was trying to prepare a high-frequency SST and Sea Ice cover (SIC) forcing data from the B-case (fully coupled), then use it to drive an F-case (atmosphere only with F-compset) simulation. I used the B-case simulation conducted with E3SM v3, i.e. 20231209.v3.LR.piControl-spinup.chrysalis [https://acme-climate.atlassian.net/wiki/spaces/CM/pages/4032626741/20231209.v3.LR.piControl-spinup.chrysalis], to extract the high-frequency SST (i.e. ``temperatureAtSurface") from _*hist.am.highFrequencyOutput*_  files.  This model historical output archives 5-day (or pentad) mean values for each month.  When I process the data, one issue arises in the model output:  
- Technically, we should have 7 pentads in January of each year (except for the 0001-01-01 when the model is started), which is also confirmed if we check the _*hist.am.highFrequencyOutput*_  files. 
-  However, this is not true for the _*hist.am.highFrequencyOutput*_  files at restart points. There are only 6 instead of 7 pentads in files at, such "0041-01-01", "0081-01-01"  and many other time slices when the simulation was restarted.  This caused a gap between  "0040-12"  and "0041-01" or "0080-12" and "0081-01" with the missing of the first 5-day mean at January.  

I think that the above problem will bring some issues for the post-processing or extra attention for users who will use the products. For my case, I indeed spent some time to figure out this feature in model output, and I have to carefully arrange the time in my forcing file for the SST, and to ensure the consistency between SST and sea ice cover data. 

I would like to report it here to see if the ocean people notice this, and whether they could apply some changes in model code to make the historical output more continuous.

xylar:
Without this, the first output is missing after each restart

fixes #6432

### DIFF
--- a/components/mpas-ocean/cime_config/buildnml
+++ b/components/mpas-ocean/cime_config/buildnml
@@ -859,7 +859,7 @@ def buildnml(case, caseroot, compname):
             lines.append('        filename_interval="00-01-00_00:00:00"')
             lines.append('        reference_time="01-01-01_00:00:00"')
             lines.append('        output_interval="00-00-05_00:00:00"')
-            lines.append('        clobber_mode="truncate"')
+            lines.append('        clobber_mode="append"')
             lines.append('        packages="highFrequencyOutputAMPKG">')
             lines.append('')
             lines.append('    <var name="xtime"/>')


xylar:
@zhangshixuan1987 and @wlin7, can you verify that this fixes #6432?  I am happy to do some testing but I think you are in a better position to verify this.

xylar:
Note: an alternative fix would be to change:
https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-ocean/cime_config/buildnml#L86
to `append` instead of `truncate`.  This may be the preferred solution, since we're not 100% sure that the fields typically included in the high frequency stream are available after a restart.

zhangshixuan1987:
> Note: an alternative fix would be to change: https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-ocean/cime_config/buildnml#L862 to `append` instead of `truncate`. This may be the preferred solution, since we're not 100% sure that the fields typically included in the high frequency stream are available after a restart.

Hi @xylar: I conducted a set of 2-month short-term simulations on  Compy with the code changes you suggested.  It turns out that the ocean model complains about the changes from `truncate` to `append` in https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-ocean/cime_config/buildnml#L862. 

> ERROR: Writing to stream 'highFrequencyOutput' would overwrite record    1 in file '20240521.v3.f6434.WCYCL1850.ne30pg2_r05_IcoswISC30E3r5.compy.mpaso.hist.am.highFrequencyOutput.0160-12-01_00.00.00.nc',
ERROR:     but clobber_mode is set to 'append'.
ERROR: Writing to stream 'highFrequencyOutput' would overwrite record    2 in file '20240521.v3.f6434.WCYCL1850.ne30pg2_r05_IcoswISC30E3r5.compy.mpaso.hist.am.highFrequencyOutput.0160-12-01_00.00.00.nc',
ERROR:     but clobber_mode is set to 'append'.
ERROR: Writing to stream 'highFrequencyOutput' would overwrite record    3 in file '20240521.v3.f6434.WCYCL1850.ne30pg2_r05_IcoswISC30E3r5.compy.mpaso.hist.am.highFrequencyOutput.0160-12-01_00.00.00.nc',
ERROR:     but clobber_mode is set to 'append'.
ERROR: Writing to stream 'highFrequencyOutput' would overwrite record    4 in file '20240521.v3.f6434.WCYCL1850.ne30pg2_r05_IcoswISC30E3r5.compy.mpaso.hist.am.highFrequencyOutput.0160-12-01_00.00.00.nc',
ERROR:     but clobber_mode is set to 'append'.
ERROR: Writing to stream 'highFrequencyOutput' would overwrite record    5 in file '20240521.v3.f6434.WCYCL1850.ne30pg2_r05_IcoswISC30E3r5.compy.mpaso.hist.am.highFrequencyOutput.0160-12-01_00.00.00.nc',
ERROR:     but clobber_mode is set to 'append'.
ERROR: Writing to stream 'highFrequencyOutput' would overwrite record    6 in file '20240521.v3.f6434.WCYCL1850.ne30pg2_r05_IcoswISC30E3r5.compy.mpaso.hist.am.highFrequencyOutput.0160-12-01_00.00.00.nc',
ERROR:     but clobber_mode is set to 'append'.

xylar:
Those error messages are annoying for sure but ultimately harmless I think. Did you make both changes? compute/write on startup and append? If so, you only want append and not compute/write on startup.

xylar:
Do the high frequency files have all the expected time entries?

zhangshixuan1987:
> Do the high frequency files have all the expected time entries?

Hi @xylar :  Yes, I think that this change can solve the issues I originally encountered. However, when I employ this solution in a new G-case simulation, it seems to me that it can potentially trigger a new potential issue for the following scenario:

- When you restart the model to rerun a piece of simulation. For example, my simulation stopped at 0199-03-01_00.00.00.nc due to hitting the job walltime limit. So I restarted the simulation at 0199-01-01_00.00.00.nc to continue my simulation. Note that previous run already generated files at 0199-02-01_00.00.00 and 0199-03-01_00.00.00. The restart simulation are supposed to overwrite these files (as we restart the model at 0199-01-01_00.00.00). However, using `append` instead of `truncate` triggers the following error from mpaso: 
```
ERROR: Writing to stream 'highFrequencyOutput' would overwrite record    1 in file '20240524.v3-LR.GMPAS-BC.ne30pg2_r05_icos30.anvil.mpaso.hist.am.high-frequency output.0199-02-01_00.00.00.nc',
ERROR:     but clobber_mode is set to 'append'.
```
I think this error is expected from the purpose of `append`  (not allowed to overwrite) and `truncate` (allowed to overwrite).  I believe that `truncate` is what we want in this scenario.   Overall, my point is that `truncate` seems to be a better choice  instead of append for most of the cases.

zhangshixuan1987:
> Those error messages are annoying for sure but ultimately harmless I think. Did you make both changes? compute/write on startup and append? If so, you only want append and not compute/write on startup.

Hi @xylar : Actually this is not the case. For the test simulation as I mentioned above, I ONLY made changes for   https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-ocean/cime_config/buildnml#L862. No other changes have been made.

zhangshixuan1987 (approved):

zhangshixuan1987:
@xylar: I think I overlooked the pull request change. I thought that #6434 and #6432 were different. In fact, the pull request along with the changes here is indeed the correct one. To avoid confusion, I would like to  clarify: 
1. changing the two flags in `mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml` as suggested in #6432 is the solution 
2. the alternative fix with changes in https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-ocean/cime_config/buildnml#L862 is NOT a preferred solution to me. 

Thank you!

xylar:
I believe that append mode is functioning as I should, and that one simply has to ignore the disconcerting error messages if one is restarting from an earlier point and cannot overwrite the data.

xylar:
I think the append mode change remains the preferred solution despite the annoying error messages. In general, we cannot guarantee that fields selected for output are available at startup. Fields may contain all zeros. So it is safer to keep the results written before the last run wrote its final restart than to overwrite with new values after restart.

You all may use the other solution of computing/writing at startup for the alternating B- and G-cases but that won't be what we ultimately do in E3SM.

xylar:
@wlin7, @zhangshixuan1987 and @mark-petersen, as discussed above, I'm going with the "alternative" method of fixing this issue where the output is in "append" mode rather than "truncate".  This has the annoying side effect that error messages may appear after each restart (particularly if a run timed out) saying that entries could not be overwritten because the mode is "append".  This was the preferred fix in our MPAS DevOps call last week.

mark-petersen (approved):
Yes, I agree. The inconvenience of extra messages from a bad restart is preferred to having a time slice missing. Thanks for everyone's detailed review and testing above.

xylar:
> The inconvenience of extra messages from a bad restart is preferred to having a time slice missing. Thanks for everyone's detailed review and testing above.

@mark-petersen, do you also agree that there's some risk in setting things to compute/write at startup?  That worked for what @zhangshixuan1987 and @wlin7 need for their purposes but I think there are enough fields that start out as all zeros until after the first time step of a restart run, so that the solution I now have in this branch is preferable.  But I would be keen to make sure you agree.

xylar:
@wlin7, I'm taking you off as a reviewer.  I would still welcome your input if you want to provide it but I don't think it is necessary before we merge this.

rljacob:
@jonbob please start merging this.

jonbob:
Passes:

merged to next

jonbob:
merged to master

xylar:
Thank you, @jonbob!

PR #6438, opened 2024-05-22
Fixes the ELM's SPBC compset

bishtgautam:
Removes turning on of the subgrid solar topographic radiation scheme.

[BFB]

### DIFF
--- a/components/elm/cime_config/config_component.xml
+++ b/components/elm/cime_config/config_component.xml
@@ -114,7 +114,7 @@
       <value compset="_ELM%[^_]*SP"        >-bgc sp</value>
       <value compset="_ELM%[^_]*TOP"       >-solar_rad_scheme top</value>
       <value compset="_ELM%[^_]*SPBC"      >-bgc sp</value>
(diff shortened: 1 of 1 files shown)


daleihao (commented):
Should we add SPBCTOP as below
<value compset="_EAM%CMIP6_ELM%[^_]*SPBCTOP" >-bgc sp -solar_rad_scheme top</value>

rljacob:
@daleihao please review

daleihao (approved):

rljacob (approved):

rljacob:
@peterdschwartz you can start merging this.

rljacob:
@bishtgautam did you see @daleihao 's question above?

bishtgautam:
I missed @daleihao's comment earlier.

Let's not add the `_EAM%CMIP6_ELM%[^_]*SPBCTOP` because it would be an active EAM that we aren't currently using in the v3 WC campaign.

peterdschwartz:
merged to next

PR #6441, opened 2024-05-22
Remove machine `theta` from machine files

ndkeen:
Remove machine `theta` from machine files.

[BFB]


ndkeen:
Note one that without machine theta, the `cime_config/testmods_dirs/allactive/mach/pet/shell_commands` is now a null-op.  I left it there as a stub file so that we don't need to change any test names.  Maybe could make testname change in another PR?  or do it here?  or leave it same for future? For example, the following tests in e3sm_integration will fail without that file:
```
PET.f19_g16.X.pm-cpu_intel.allactive-mach-pet
PET.f45_g37_rx1.A.pm-cpu_intel.allactive-mach-pet
PET_Ln5.ne4pg2_oQU480.F2010.pm-cpu_intel.allactive-mach-pet
PET_Ln9_PS.ne30pg2_r05_IcoswISC30E3r5.WCYCL1850.pm-cpu_intel.allactive-mach-pet
```

amametjanov (approved):

PR #6444, opened 2024-05-23
Add pgrad_correction settings to caar_ut.

trey-ornl:
The `caar_ut` test was only using the value `pgrad_correction==0`, while current runs often use `pgrad_correction==1`. Modified the test to compare bit-for-bit answers for both settings. Diff appears to have trouble realizing that the changes to `caar_ut.cpp` are minor, ignoring indentation changes for the new loop level. `git diff -w` looks much cleaner.

Added `frontier-bfb.cmake`, to build `caar_ut` on Frontier.

Fixed `ExecSpaceDefs.hpp` for the newer version of Kokkos. Replaced a dependence on Kokkos implementation details, which no longer works as intended on Frontier, with standard C++ code.


oksanaguba:
@trey-ornl thanks for making frontier-bfb file. i assume you ran this test and confirmed bfb-ness for caar_ut? i can run hommexx suite on frontier (more tests than just caar_ut) and i will run them on nvidia (only because of the minor change in execspacedef file).

trey-ornl:
@oksanaguba, yes, I confirmed that `caar_ut` passed on Frontier using `frontier-bfb.cmake`.

rljacob:
Does this change answers for regular EAM cases?

trey-ornl:
> Does this change answers for regular EAM cases?

It should not affect EAM results at all. It adds testing to the CAAR unit test.

rljacob:
@oksanaguba is this ready?

oksanaguba:
no, i want to run tests, however redundant they are, will merge next week

oksanaguba:
i merged this into a weaver branch (1 conflict) and most of theta-f* tests look good (did not wait for all theta-fhs*), but caar_ut freezes, so i need to debug.

oksanaguba:
in next. weaver is ok, up to the issue with limiter barriers (has nothing to do with this PR), i will check frontier separately (Trey already did).

oksanaguba:
only fail i saw was from pm-cpu_gnu because of disappeared baselines. ran manually next against today's master and  it passed.

oksanaguba (approved):

PR #6448, opened 2024-05-28
Fixes z-level mode in overflow test case

vanroekel:
In current master, setting `config_overflow_layer_type = 'z-level'` caused a failure in forward mode on the first time step.  This is due to very large gradients in sea surface height.  This PR fixes those issues and with the changes, z-level mode has been run for 5-day simulations.

### DIFF
--- a/components/mpas-ocean/src/mode_init/mpas_ocn_init_overflow.F
+++ b/components/mpas-ocean/src/mode_init/mpas_ocn_init_overflow.F
@@ -80,7 +80,7 @@ contains
       integer, intent(out) :: iErr
       real (kind=RKIND) :: yMin, yMax, dcEdgeMin
       real (kind=RKIND) :: yMinGlobal, yMaxGlobal, dcEdgeMinGlobal
-      real (kind=RKIND) :: plugWidth
+      real (kind=RKIND) :: ltSum, plugWidth
       real (kind=RKIND) :: slopeCenter, slopeWidth
 
       type (block_type), pointer :: block_ptr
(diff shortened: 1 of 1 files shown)


vanroekel:
@xylar, @cbegeman and @mark-petersen these changes worked for me, but I'd appreciate your thoughts on how I implemented things.  I'm not sure if there is a better approach

vanroekel:
Here is a graphic of temperature in z-level after 4 days

image

It seems as I would expect.  I'm running the sigma equivalent for comparison

components/mpas-ocean/src/mode_init/mpas_ocn_init_overflow.F
```diff
@@ -228,8 +228,17 @@ subroutine ocn_init_setup_overflow(domain, iErr)!{{{
 
            ! Set layerThickness and restingThickness
            if ( trim(config_overflow_layer_type) == 'z-level' ) then
+              layerThickness(:,iCell) = 0.0_RKIND
```
vanroekel, line 231: is this necessary?  When I plotted the layerThickness out of initial_state it was nan below maxLevelCell, but perhaps this gets filled appropriately in the forward step

cbegeman:
@vanroekel I think your approach here is appropriate. I'll test it out shortly.

I just wanted to point out that @scalandr has implemented an initial state using Compass's vertical coordinate here: https://github.com/MPAS-Dev/compass/blob/675d4b0111c04e8c922578fde10a578b1e789faa/compass/ocean/tests/overflow/initial_state.py#L83

And this is the z-level algorithm in Compass that should result in ssh=0 for this case. I think this results in a similar outcome to your approach (modifying thickness of the bottom layer).
https://github.com/MPAS-Dev/compass/blob/675d4b0111c04e8c922578fde10a578b1e789faa/compass/ocean/vertical/zlevel.py#L163-L169

cbegeman:
@vanroekel Can you provide details of the configuration that crashes? I'm not seeing it crash with the 10km default test.

vanroekel:
@cbegeman that's right, the default is fine with default settings. but if you set
```
config_overflow_layer_type = 'z-level'
```
in the initialization stage you should see the crash in forward run

Sorry for not specifying that clearly

vanroekel:
very interesting.  Let me try again and see if I messed up something along the way.

vanroekel:
@cbegeman I think I see what happened.  Let me know if this makes sense.  I ran and got the SSH warnings for very large values. -- it seems this stems from use of initMode and not what @scalandr developed -- however, it appears the case still runs fine with this large SSH gradient (as you found).  If I used the hydro_v_nonhydro test I'd get ssh=0 again.

If you think this is consistent, I'm just going to close this as I'd imagine we'd just want to migrate to using the compass initial state instead of initMode and not consider my changes here.

cbegeman:
@vanroekel I don't think there's much harm in merging this, but I agree that it doesn't seem to be necessary so you can close it if you prefer.

I plan on migrating to python's initialization for this test case, but it may take place with in polaris instead of compass.

xylar:
@vanroekel, I would prefer that we migrate to using python for initial conditions wherever possible and that we not put further effort into MPAS-Ocean init mode except where we don't have an alternative.  I realize you've already put in the effort so I'll review further if you like but I would lean toward closing this and using the python code instead.

vanroekel:
I too think we should close this and move to python.

PR #6449, opened 2024-05-28
hommexx: fix templating

oksanaguba:
minor fix for column ops for BFB runs on gpu.

adding hops cache files.

### DIFF
--- a/components/homme/src/share/cxx/ExecSpaceDefs.hpp
+++ b/components/homme/src/share/cxx/ExecSpaceDefs.hpp
@@ -383,11 +383,19 @@ struct Dispatch<HommexxGPU> {
     // serialize parallel scans.
 
     // Detect the value type
+#ifdef HOMMEXX_ENABLE_GPU
+    using value_type =
+      typename Kokkos::Impl::FunctorAnalysis
+        < Kokkos::Impl::FunctorPatternInterface::SCAN
+        , void
+        , Lambda, void >::value_type ;
+#else
     using value_type =
       typename Kokkos::Impl::FunctorAnalysis
         < Kokkos::Impl::FunctorPatternInterface::SCAN
         , void
         , Lambda >::value_type ;
+#endif
 
     // All threads init result.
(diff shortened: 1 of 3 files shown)


rljacob:
@oksanaguba status update?

oksanaguba:
will merge next week

oksanaguba:
i forgot about this PR, but will merge it in a few days.

rljacob:
Notes: Depends on a @trey-ornl  PR.  May not be needed.

components/homme/src/share/cxx/ExecSpaceDefs.hpp
```diff
@@ -383,11 +383,19 @@ struct Dispatch<HommexxGPU> {
...
+      typename Kokkos::Impl::FunctorAnalysis
+        < Kokkos::Impl::FunctorPatternInterface::SCAN
+        , void
+        , Lambda, void >::value_type ;
+#else
     using value_type =
       typename Kokkos::Impl::FunctorAnalysis
         < Kokkos::Impl::FunctorPatternInterface::SCAN
         , void
         , Lambda >::value_type ;
```
trey-ornl, line 397: So just adding the extra `void` template argument doesn't work for the non-GPU case, thus requiring the `#ifdef`?

trey-ornl:
> Notes: Depends on a @trey-ornl PR. May not be needed.

I'm happy to review, but I don't recall the dependent PR. Which one is it?

oksanaguba:
@trey-ornl -- i do not remember. i think this PR is in a conflict with your CAAR PR, so it is not relevant. i can even close it.

trey-ornl:
> @trey-ornl -- i do not remember. i think this PR is in a conflict with your CAAR PR, so it is not relevant. i can even close it.

[This one](https://github.com/E3SM-Project/E3SM/pull/6522/) has changes to `ExecSpaceDefs.cpp` but not `ExecSpaceDefs.hpp`. I think they're unrelated.

oksanaguba:
i dont have time testing this, so let's close for now.

PR #6454, opened 2024-05-31
Fix bottom depth deepening in global_ocean init mode

### LINKED ISSUE #6453: Bug in bottom depth in MPAS-Ocean global_ocean init mode
A bug was introduced in #6310 in MPAS-Ocean's standalone init mode.  In the `global_ocean` test case, the bottom depth is incorrectly altered, such that all but the shallowest layers have zero layer thickness.

As an example, see:
https://github.com/E3SM-Project/E3SM/pull/6310#issuecomment-2141515857

This appears to be an incorrect change from a `max()` to a `min()` operator on this line:
https://github.com/E3SM-Project/E3SM/blob/ef925d118e756f4f9379423ee4022581dd5aafad/components/mpas-ocean/src/mode_init/mpas_ocn_init_global_ocean.F#L897

xylar:
A bug was introduced in #6310 that made the ocean shallower rather than deeper than the minimum allowed depth.  This merge fixes that bug.

Fixes #6453 
[BFB] -- mpas-ocean standalone only

### DIFF
--- a/components/mpas-ocean/src/mode_init/mpas_ocn_init_global_ocean.F
+++ b/components/mpas-ocean/src/mode_init/mpas_ocn_init_global_ocean.F
@@ -894,7 +894,7 @@ contains
              if (isOcean) then
                 ! Enforce minimum depth
                 bottomDepth(iCell) = max(bottomDepthObserved(iCell), config_global_ocean_minimum_depth)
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

With this fix cherry-picked onto #6310, the `pr` suite in Compass is now BFB with a baseline before #6310

jonbob:
Passes:

merged to next

jonbob:
merged to master

PR #6456, opened 2024-05-31
replace global sum with full-precision reproducible sum

### LINKED ISSUE #6446: Global sum BFB code in the surface salinity restoring routine reduces precision.
The global sum in subroutine ocn_get_surfaceSalinityData (required to ensure no net salt input into the ocean due to the restoring term) was originally implemented in the simplest way, which was not guaranteed to be BFB on different processor layouts.  a modification was made to make it BFB, but apparently this method reduces the precision of the restoring term.  while it seems to be a minor problem, we should be using the mpas reproducible sum routine instead.

this plot shows a 40 year time series of the globally averaged monthly SSS restoring tendency (log(abs(tend)) for a G-case using the IcosWISC30E3r5 mesh (in black).  The red curve is a restart from year 40 using double precision output, which doesn't help.  The green curve has the BFB code turned off showing much higher precision and no spikes.

Screenshot 2024-05-22 at 10 28 38 AM

maltrud:
The global sum in the surface salinity restoring routine was forced to be BFB using a truncation calculation that resulted in reduced precision and some unexplained behavior.  This PR replaces this calculation with an mpas standard reproducible sum with full precision.  

Testing was done using a G-case with the IcoswISC30E3r5 grid for one year, then calculating the global average SSS restoring tendency term, which should be effectively zero.  Also confirmed that the new code resulted in BFB solution when processor layout was changed from 10 nodes on chrysalis to 15 nodes.

Fixes #6446 
[non-BFB] for C- and G-cases with salinity restoring

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_tracer_surface_restoring.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_tracer_surface_restoring.F
@@ -26,6 +26,7 @@ module ocn_tracer_surface_restoring
    use mpas_timekeeping
    use mpas_forcing
    use mpas_stream_manager
+   use mpas_global_sum_mod
    use ocn_constants
    use ocn_config
    use ocn_framework_forcing
@@ -250,7 +251,11 @@ contains
         integer :: ierr
 
         integer, parameter :: nSums = 2
-        real (kind=RKIND), dimension(nSums) :: reductions, sums
+        real (kind=RKIND), dimension(nSums) :: reductions
+
+        integer, dimension(nSums) :: indexForReproSum
+        real (kind=RKIND), dimension(:,:), allocatable :: &
+           localArrayForReproSum
 
         ! initialize monthly forcing to be read from file
 
@@ -339,6 +344,9 @@ contains
         call mpas_pool_get_array(forcingPool, 'landIceMask', landIceMask)
         call mpas_pool_get_array(meshPool, 'areaCell', areaCell)
 
+        allocate (localArrayForReproSum(nCellsSolve,2))
+        localArrayForReproSum(:,:) = 0.0_RKIND
+
         ! This is not in a threaded region, so no openMP pragmas are needed.
         if ( associated(landIceMask)) then
 
@@ -414,11 +422,14 @@ contains
 
         endif
 
+        indexForReproSum(1) = 1
+        indexForReproSum(2) = 0
         do iCell=1,nCellsSolve
            deltaS = activeTracersSurfaceRestoringValue(indexSalinitySurfaceRestoringValue,iCell)
            if (deltaS .ne. 0.0_RKIND) then
-              sumAreaDeltaS = sumAreaDeltaS + deltaS*areaCell(iCell)
-              sumArea = sumArea + areaCell(iCell)
+              indexForReproSum(2) = indexForReproSum(2) + 1
+              localArrayForReproSum(indexForReproSum(2),1) = deltaS
+              localArrayForReproSum(indexForReproSum(2),2) = areaCell(iCell)
            endif
         enddo
 
@@ -427,16 +438,18 @@ contains
 
       ! Global sum to subtract global mean of deltaS
       dminfo = domain % dminfo
-      sums(1) = sumAreaDeltaS
-      sums(2) = sumArea
-      call mpas_dmpar_sum_real_array(dminfo, nSums, sums(1:nSums), reductions(1:nSums))
-      sumAreaDeltaSGlobal = reductions(1)
-      sumAreaGlobal       = reductions(2)
-      avgDeltaS1 = sumAreaDeltaSGlobal/(sumAreaGlobal + 1.e-20_RKIND)
-      !LPV: There is no guarantee that a global sum of deltaS is BFB across
-      !decompositions, the next line rounds each value to 10 decimal places to
-      !ensure BFB results
-      avgDeltaS = float (int(avgDeltaS1 * 1.0E10_RKIND + 0.5_RKIND)) / 1.0E10_RKIND
+! first do 1 field sum for area
+      reductions(1) = mpas_global_sum(localArrayForReproSum(:,2), &
+                                           domain%dminfo%comm, indexForReproSum)
+! now do sum of 2 mupltiplied fields for area*deltaS
+      reductions(2) = mpas_global_sum(localArrayForReproSum(:,1), localArrayForReproSum(:,2), &
+                                           domain%dminfo%comm, indexForReproSum)
+
+      sumAreaDeltaSGlobal = reductions(2)
+      sumAreaGlobal       = reductions(1)
+      avgDeltaS = sumAreaDeltaSGlobal/(sumAreaGlobal + 1.e-20_RKIND)
+
+      deallocate (localArrayForReproSum)
 
       block => domain % blocklist
       do while (associated(block))
--- a/components/mpas-ocean/src/shared/mpas_ocn_tracer_surface_restoring.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_tracer_surface_restoring.F
@@ -253,7 +253,7 @@ contains
         integer, parameter :: nSums = 2
         real (kind=RKIND), dimension(nSums) :: reductions
 
-        integer, dimension(nSums) :: indexForReproSum
+        integer, dimension(2) :: indexForReproSum  ! min, max indices for 1 dimensional sums
         real (kind=RKIND), dimension(:,:), allocatable :: &
            localArrayForReproSum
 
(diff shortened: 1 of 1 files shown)


maltrud:
Here's a time series of the log10 of the absolute value of the globally averaged SSS tendency (monthly averages) for 4 simulations.

Screenshot 2024-05-30 at 5 26 39 PM

The black line is the old global sum with real4 output, and the green is the old sum with real8 output.  The red line is the new reproducible sum with r4 output, and the blue is the new sum with real8 output.

components/mpas-ocean/src/shared/mpas_ocn_tracer_surface_restoring.F
```diff
@@ -414,11 +422,14 @@ subroutine ocn_get_surfaceSalinityData( streamManager, &
 
         endif
 
+        indexForReproSum(1) = 1
+        indexForReproSum(2) = 0
```
vanroekel, line 426: This is probably just do to the construction of the repro sum, but I'm a bit confused by the two length array of indexForReproSum.  It seems position 1 is always just 1 and position 2 is the total number of entries in the sum.  Is that right?  What is position 1 of this index array used for?  It never appears to be modified and then just get used in the globalSum call.

philipwjones: @vanroekel  The reproducible sums in MPAS allow sums over full arrays and over restricted address ranges in each dimension. For full generality, the range must be specified in min/max pairs for each array dimension. In this case, the array is getting packed in advance so the min is always 1 but the max can vary. But both need to be provided for the interface.

alicebarthel: Thanks @philipwjones for the explanation. It didn't make sense to me either.

maltrud: @vanroekel this has always been part of the algorithm (essentially the "if deltaS .ne. 0.0" part) to account for possible cases where we explicitly don't want to apply restoring to some cells (eg, under land ice maybe?).  then we don't actually want to include every cell in these sums.  does that make sense?

alicebarthel:
```suggestion
        indexForReproSum(2) = 0
      ! For reproducible sums like mpas_global_sum_mod(), the range must be specified in min/max pairs 
      ! for each array dimension. In this case, the array is getting packed in advance so the min is always 1 
      ! but the max can vary. Both need to be provided for the interface.
```

vanroekel: Thanks @philipwjones  for the help understanding this.  I'd like to include the comment from @alicebarthel in the PR for clarity.  If it is too long for a comment, perhaps we can think of a shorter version.

components/mpas-ocean/src/shared/mpas_ocn_tracer_surface_restoring.F
```diff
@@ -250,7 +251,11 @@ subroutine ocn_get_surfaceSalinityData( streamManager, &
         integer :: ierr
 
         integer, parameter :: nSums = 2
-        real (kind=RKIND), dimension(nSums) :: reductions, sums
+        real (kind=RKIND), dimension(nSums) :: reductions
+
+        integer, dimension(nSums) :: indexForReproSum
```
alicebarthel, line 256: @philipwjones It's a minor detail, but I'm curious. Can you confirm if I understood this correctly: in this case, it's a coincidence that indexForReproSum and reductions have the same dimensions of nSums? 
Technically, indexForReproSum should always have 2 (the min, max you mentioned), while nSums could be anything given the needs of the routine? (here, 2 since we need the area sum and the area-weighted salinity sum.
Does it make sense to use nSums to define indexForReproSum then?

components/mpas-ocean/src/shared/mpas_ocn_tracer_surface_restoring.F
```diff
@@ -250,7 +251,11 @@ subroutine ocn_get_surfaceSalinityData( streamManager, &
         integer :: ierr
 
         integer, parameter :: nSums = 2
-        real (kind=RKIND), dimension(nSums) :: reductions, sums
+        real (kind=RKIND), dimension(nSums) :: reductions
```
alicebarthel, line 254: @maltrud just curious: is there any reason you went with `reductions` as a name for these variables? Sums is more intuitive to me, but maybe I'm not familiar with the jargon?

components/mpas-ocean/src/shared/mpas_ocn_tracer_surface_restoring.F
philipwjones: @alicebarthel  Good catch. The index array should be dimensioned 2*(number of dimensions being summed over) - in this case just 2 since the arrays are just 1-d. So nSums is coincidentally correct but in incorrect in general.

maltrud: @alicebarthel I wouldn't say it is "a coincidence" that nSums=2;  this routine isn't meant to be general which is why I hardwired it to be correct for this case.

components/mpas-ocean/src/shared/mpas_ocn_tracer_surface_restoring.F
maltrud: @alicebarthel  I wanted to make as few code changes as possible so I retained the same variable names that already existed whenever possible.  in the new code, "sums" isn't needed anymore, while "reductions" serves the same purpose in both the current and new versions.  probably should have done things differently many years ago, but that's the rationale.

thanks for your careful reviewing.

components/mpas-ocean/src/shared/mpas_ocn_tracer_surface_restoring.F
maltrud: ok, maybe I still don't understand the question.  we have 2 sums to do:  area and area*deltaS, so nSums = 2.  so we need the same number of outputs from the 2 global sum calls:  global_sum(area) and global_sum(area*deltaS) which are put into reductions(1) and reductions(2), respectively.  so reductions should always be dimensioned nSums.  I guess I could have just put sumAreaGlobal and sumAreaDeltaSGlobal directly in the global sum calls and not needed reductions(), but I wanted to keep the new version as similar to the current version as possible.  does this make sense?

alicebarthel: I don't have an issue with dimensioning reductions with nSums, I think that's great! 
Dimensioning indexForReproSum with nSums (L256) makes no sense to me, since the length (2) has nothing to do with nSums, but with 2*len(shape(variable we are summing over)) = 2*1-D variable = 2 in this case. My comment is over L256, not 254, sorry if the github visualization is confusing.

components/mpas-ocean/src/shared/mpas_ocn_tracer_surface_restoring.F
alicebarthel: @maltrud I totally understand the wisdom of making minimal code changes. As a newcomer to these routines, I allow myself to make minor comments as to readability to new users. Take it or leave it :-)

components/mpas-ocean/src/shared/mpas_ocn_tracer_surface_restoring.F
maltrud: ah, ok, now I understand.  you're right, that is misleading.  should I just place it with "2"?

maltrud: and a comment?  thanks for clarifying.

philipwjones: I think you should create a variable ThroatWarblerMangrove that's pronounced "two"...

alicebarthel: @maltrud Sounds good to me, thanks!

alicebarthel (approved):
Approving based on visual inspection, provided that change to L256 is made.

mark-petersen (approved):
I compiled this with gnu (debug and optimized) on chicoma and intel on chrysalis and tested with the nightly test suite. That tests the compiling, but not the function of this code. 

Thanks for running the G case tests. That is convincing. Since you are able to show reproducible sums on different processor counts, I'm happy to have this merged in.

maltrud:
I've made the recommended changes and push changes to the branch.

jonbob:
@vanroekel -- do you approve now with the changes?

vanroekel (approved):
Approving based on visual inspection and testing from @maltrud and @alicebarthel

jonbob:
As expected, passes:
* PET_Ln5.T62_oQU240wLI.GMPAS-DIB-IAF-DISMF.chrysalis_intel (G-case but no restoring file)

and DIFFs for:
* SMS_D_Ld1.TL319_IcoswISC30E3r5.GMPAS-JRA1p5-DIB-PISMF.chrysalis_intel.mpaso-jra_1958 (G-case with restoring file)

merged to next

jonbob:

Issue #6144, opened 2024-01-10
standalone homme does not build on machines without cprnc


oksanaguba:
The issue is 
```
CMake Error at utils/cime/CIME/non_py/cprnc/CMakeLists.txt:40 (add_custom_target):
  add_custom_target cannot create target "genf90" because another target with
  the same name already exists.  The existing target is a custom target
  created in source directory
  "/ccs/home/onguba/acme-master/components/homme/utils/externals/scorpio/src/flib".
  See documentation for policy CMP0002 for more details.
```
reported by Conrad after the most recent cime changes to master, and i reproduced it on frontier

I have seen this before, but i do not remember how it got resolved. Machines with pre-built cprnc are not affected.

PR #6114, opened 2023-12-08
Updates HOMME Perlmutter machine files.

jasonltorchinsky:
(Summary Below)
Functional update to HOMME Perlmutter 'nocuda' machine files, which previously caused errors when running DCMIP2016 Test 2 on Perlmutter. Tools to enable building all DCMIP test cases for standalone HOMME on Perlmutter, and to run DCMIP2016 Test 2 on Perlmutter.

* Updates HOMME Perlmutter 'nocuda' machine files to enable PIO.
* Cleans namelists for DCMIP2016 Test 2.
* Adds NERSC jobscript for DCMIP2016 Test 2.
* Adds NERSC build scripts for standalone HOMME.

[NML] Formatting only on DCMIP 2016 Test 2 namelists.
[BFB]

PR #6461, opened 2024-06-06
standalone HOMME support on PM-CPU

mt5555:
Add support for running standalone HOMME on PM-CPU

With this PR, standalone HOMME can be build on PM-CPU, with either GNU or Intel compilers.  In particular, both of these will work:

./create_test   --machine pm-cpu --compiler intel HOMME_P24.f19_g16_rx1.A                    
./create_test   --machine pm-cpu --compiler gnu  HOMME_P24.f19_g16_rx1.A                                                                           

- Single cmake file (pm-cpu.cmake, to match CIME machine names) supports both GNU and Intel without hardcoded netcdf paths.  
- Also updated test scripts to set  SRUN_CPUS_PER_TASK properly, otherwise openMP performance is terrible.  
- removed assumption that fortran and C compiler flags for openMP are the same ( they are different under PM's oneAPI compilers)

[BFB]

### DIFF
--- a/components/homme/cmake/SetCompilerFlags.cmake
+++ b/components/homme/cmake/SetCompilerFlags.cmake
@@ -259,15 +259,11 @@ IF (ENABLE_HORIZ_OPENMP OR ENABLE_COLUMN_OPENMP)
           SET(OpenMP_C_FLAGS "-qsmp=omp:nested_par -qsuppress=1520-045:1506-793")
         ENDIF ()
       ENDIF ()
-      # This file is needed for the timing library - this is currently
-      # inaccessible from the timing CMake script
-      SET(OpenMP_Fortran_FLAGS "${OpenMP_C_FLAGS}")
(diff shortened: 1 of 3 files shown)


rljacob:
How does this relate to https://github.com/E3SM-Project/E3SM/pull/6114 ?

mt5555:
#6144 will be merged after this PR.    #6144 is mostly cleanup of standalone HOMME tests cases.

oksanaguba (approved):

oksanaguba:
for the record -- i ran homme sycl branch with intel on pm-gpu (with kokkos serial) and did not see issues with booleans.

rljacob:
Can you remind me what the HOMMEBFB suites don't work on pm-cpu ?

oksanaguba:
bfb suite needs bfb flags to make it work. so we need to insert if-statements in the machine file for whether it is a bfb and non-bfb case, and these flags depend on a compiler (i think right now PM gnu and intel use the same machine file).

PR #6463, opened 2024-06-10
update cube_to_target

whannah1:
Clean up and simplify the Makefile for the cube_to_target topography generation tool. 

[BFB]

### DIFF
--- a/components/eam/tools/topo_tool/cube_to_target/Makefile
+++ b/components/eam/tools/topo_tool/cube_to_target/Makefile
@@ -5,26 +5,20 @@ RM = rm
 .SUFFIXES:
 .SUFFIXES: .F90 .o
 
-# Set the compiler; note that on some platforms, FC may be defined as an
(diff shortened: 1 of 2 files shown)


oksanaguba:
On which machines that we support is it supposed to run, if any?

whannah1:
@oksanaguba I don't know of any requirement to have this running on a supported machine. Someone created a script to allow regular testing, but I don't think we actually run tests. The idea here was to clean it up and make sure it works on Perlmutter and Chrysalis, which seem to be the machines that most  people use most of the time. We can modify this to support other machines as the use cases arise.

oksanaguba:
@whannah1 thanks. so this needs to be tested on PM and chrysalis. did you test on either or should i?

whannah1:
I verified that it builds on both Perlmutter and Chrysalis.

oksanaguba (approved):

Issue #6470, opened 2024-06-13
Build fail in MPAS with nvidiagpu compiler


ndkeen:
With test `SMS_Ld1.T62_oEC60to30v3.CMPASO-NYF.pm-gpu_nvidiagpu` it has been failing for a while now. I think I mentioned this to @jonbob who said the fail dates matched a PR that recently went in. I thought I had made an issue, but maybe not.

```
 0 inform,   0 warnings,   1 severes, 0 fatal for ocn_diagnostics_variables_destroy
Target CMakeFiles/ocn.dir/__/__/core_ocean/shared/mpas_ocn_diagnostics_variables.f90.o built in 0.444529 seconds
gmake[2]: *** [mpas-framework/src/CMakeFiles/ocn.dir/build.make:918: mpas-framework/src/CMakeFiles/ocn.dir/__/__/core_ocean/shared/mpas_ocn_diagnostics_variables.f90.o] Error 2
gmake[2]: *** Waiting for unfinished jobs....
ocn_equation_of_state_linear_density_only:
    181, Generating present(tracerssurfacelayervalue(:,:),density(:,:))
         Generating NVIDIA GPU code
        183, !$acc loop gang, vector(128) collapse(2) ! blockidx%x threadidx%x
        184,   ! blockidx%x threadidx%x collapsed
    198, Generating present(tracers(:,:,:),density(:,:))
         Generating NVIDIA GPU code
        200, !$acc loop gang, vector(128) collapse(2) ! blockidx%x threadidx%x
        201,   ! blockidx%x threadidx%x collapsed
Target CMakeFiles/ocn.dir/__/__/core_ocean/shared/mpas_ocn_equation_of_state_jm.f90.o built in 2.100368 seconds
ocn_equation_of_state_linear_density_exp:
    315, Generating present(thermalexpansioncoeff(:,:),tracerssurfacelayervalue(:,:),density(:,:),salinecontractioncoeff(:,:))
         Generating NVIDIA GPU code
        320, !$acc loop gang, vector(128) collapse(2) ! blockidx%x threadidx%x
        321,   ! blockidx%x threadidx%x collapsed
    341, Generating present(tracers(:,:,:),thermalexpansioncoeff(:,:),density(:,:),salinecontractioncoeff(:,:))
         Generating NVIDIA GPU code
        346, !$acc loop gang, vector(128) collapse(2) ! blockidx%x threadidx%x
        347,   ! blockidx%x threadidx%x collapsed
[...]

PR #6471, opened 2024-06-13
Fix OpenACC deletes for topographic wave drag

xylar:
Partially addresses #6470 

The errors were introduced in https://github.com/E3SM-Project/E3SM/pull/6310, when variables were introduced and renamed, including in the OpenACC create directives but corresponding changes were incomplete for the OpenACC directives for deleting them.

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_diagnostics_variables.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_diagnostics_variables.F
@@ -1009,7 +1009,13 @@ contains
          !$acc                   velocityTidalRMS)
       end if
       if (config_use_topographic_wave_drag) then
-         !$acc exit data delete(topographic_wave_drag)
+         !$acc exit data delete(topo_rinv)
+         !$acc exit data delete(topo_buoyancy_N1V)
+         !$acc exit data delete(topo_buoyancy_N1B)
+         !$acc exit data delete(bathy_stddev)
+         !$acc exit data delete(bed_slope_edges)
+         !$acc exit data delete(lonGradEdge)
+         !$acc exit data delete(latGradEdge)
          !$acc exit data delete(temp_twd)
          !$acc exit data delete(normalCoeffTWD)
          !$acc exit data delete(tangentialCoeffTWD)


sbrus89:
@xylar, just an FYI: in testing this with `SMS_Ld1.T62_oEC60to30v3.CMPASO-NYF.pm-gpu_nvidiagpu` I'm seeing this other build error:
```
NVFORTRAN-S-0155-DO loop expected after loop collapse() (/pscratch/sd/s/sbrus/e3sm_scratch/pm-gpu/SMS_Ld1.T62_oEC60to30v3.CMPASO-NYF.pm-gpu_nvidiagpu.20240613_121539_t9vbh9/bld/cmake-bld/core_ocean/shared/mpas_ocn_vmix.f90: 1461)
  0 inform,   0 warnings,   1 severes, 0 fatal for ocn_vmix_implicit
```
I'll look into it and push any changes to your branch.

philipwjones:
@sbrus89 Don't see any obvious problems with the loop its pointing too, but you might try removing the inner loop vector constructs - we've found those to be a bit flaky and the compiler generally vectorizes the loop anyway.

philipwjones (approved):
Leaving on vacation so can't review any future mods, but the changes here look fine. Will approve these changes and assume others will find any remaining issues.

mark-petersen (approved):
@xylar thanks for catching this. This clearly matches the create lines here:
https://github.com/E3SM-Project/E3SM/pull/6310/files#diff-72984751a085d56765a30726c289582fe4f2e152d9e3b78dc06a405b19406879R738

I test compiled with gnu and intel, but don't have a workflow to test OpenACC. Thanks @sbrus89 for testing above. That looks like an unrelated error, so I think we can merge this one and then deal with that separately.

sbrus89:
I found the issue with `mpas_ocn_vmix.f90`. The line numbers were misleading because they refer to the `bld/cmake-bld` code where the `#ifdef` directories are stripped out. Right now, this commit is on my branch for #6472, but should probably move to a separate PR. I also found what I believe is a missing `!$omp parallel` region in `mpas_ocn_vmix.f90`. I was able to successfully compile with the combination of #6471 (this PR) and #6472.

xylar:
@sbrus89, I'm okay with keeping the vmix fixes in #6472 rather than further separating them out.  Could you approve this if you're happy with it?

sbrus89 (approved):
Approved based on testing above. Thanks @xylar!

jonbob:
Passes:

merged to next

jonbob:
merged to master

PR #6472, opened 2024-06-13
Fix OpenACC rountine issue for subgrid wetting and drying

sbrus89:
This PR moves a subgrid subroutine call out of a OpenACC parallel region to fix the compile problems noted in #6470. Since subgrid wetting and drying is strictly a MPAS-Ocean standalone feature, I believe it should be fine for this code to remain CPU-only.

This PR also fixes a couple issues in `mpas_ocn_vmix.F`:
-  An OpenACC bug related to the use of `gang vector collapse(3)` on a double nested loop with variable inner loop bounds. 
- A missing `!$omp parallel` region in a calculation for the `config_use_gotm` option.

[BFB] -- mpas-ocean standalone

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
@@ -886,9 +886,15 @@ contains
          ! Use centered (mean) thickness as flux value
 
          if ( config_use_subgrid_wetting_drying ) then
+#ifdef MPAS_OPENACC
+         !$acc update host(ssh)
+#endif
 
             call ocn_subgrid_layerThickEdgeFlux_center(ssh, layerThickEdgeFlux)
 
+#ifdef MPAS_OPENACC
+         !$acc update device(layerThickEdgeFlux)
+#endif
          else
 
 #ifdef MPAS_OPENACC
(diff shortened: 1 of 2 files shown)


philipwjones (approved):
I'm about to head out on vacation so can't give this a proper review - but this looks ok except that you probably want to update the device copies of any array that is changed in the loop and make sure the input args have been similarly updated on host.  I will go ahead and approve assuming someone else can test and check whether all arrays have been appropriately synchronized.

mark-petersen (approved):
I compiled this with gnu (debug and optimized) on chicoma and intel on chrysalis and tested with the nightly test suite. This works fine. I tried to run buttermilk bay and the parabolic bowl on chicoma, but it was failing because it didn't find the mesh file, unrelated to this PR. Anyway, the code changes are clear and I am comfortable with this change.

xylar (approved):

xylar:
Approved by code inspection and based on @sbrus89's testing.  I think we should just keep the vmix fixes here because otherwise we'll have too many PRs.

sbrus89:
I've added the host/device updates as @philipwjones suggested and tested one more time. All stages of `SMS_Ld1.T62_oEC60to30v3.CMPASO-NYF.pm-gpu_nvidiagpu` pass with the combination of this PR and #6471.

jonbob:
Passes:

merged to next

jonbob:
merged to master

PR #6473, opened 2024-06-13
Clean up kokkos/tril situation

jgfouca:
If Trilinos is being used, we should use the kokkos that comes with it. This removes the need to explicitly set Kokkos_ROOT in config_machines.xml to the Trilinos one.

[BFB]

### DIFF
--- a/components/cmake/find_dep_packages.cmake
+++ b/components/cmake/find_dep_packages.cmake
@@ -16,10 +16,14 @@
 # errors.
 if (USE_KOKKOS)
 
-  # Kokkos will be built in the sharedlibs if Kokkos_ROOT is
-  # unset.
   if (NOT DEFINED ENV{Kokkos_ROOT})
-    set(ENV{Kokkos_ROOT} ${INSTALL_SHAREDPATH})
+    # We should use the kokkos from Trilinos if we are using Trilinos
+    if (USE_TRILINOS)
+      set (ENV{Kokkos_ROOT} ${Trilinos_ROOT})
+    else()
+      # Kokkos will be built in the sharedlibs if Kokkos_ROOT is unset.
+      set(ENV{Kokkos_ROOT} ${INSTALL_SHAREDPATH})
+    endif()
   endif()
 
   find_package(Kokkos REQUIRED)
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -277,7 +277,6 @@
       <env name="BLA_VENDOR">Generic</env>
       <env name="Albany_ROOT">$SHELL{if [ -z "$Albany_ROOT" ]; then echo /global/common/software/e3sm/mali_tpls/albany-e3sm-serial-release-gcc-cmake-fix; else echo "$Albany_ROOT"; fi}</env>
       <env name="Trilinos_ROOT">$SHELL{if [ -z "$Trilinos_ROOT" ]; then echo /global/common/software/e3sm/mali_tpls/trilinos-e3sm-serial-release-gcc; else echo "$Trilinos_ROOT"; fi}</env>
-      <env name="Kokkos_ROOT">$SHELL{if [ -z "$Kokkos_ROOT" ]; then echo /global/common/software/e3sm/mali_tpls/trilinos-e3sm-serial-release-gcc; else echo "$Kokkos_ROOT"; fi}</env>
     </environment_variables>
     <environment_variables compiler="nvidia" mpilib="mpich">
       <env name="ADIOS2_ROOT">$SHELL{if [ -z "$ADIOS2_ROOT" ]; then echo /global/cfs/cdirs/e3sm/3rdparty/adios2/2.9.1/cray-mpich-8.1.25/nvidia-22.7; else echo "$ADIOS2_ROOT"; fi}</env>
@@ -2656,7 +2655,6 @@
     <environment_variables compiler="gnu">
       <env name="Albany_ROOT">$SHELL{if [ -z "$Albany_ROOT" ]; then echo /lcrc/group/e3sm/ac.jwatkins/LandIce/AlbanyBuilds/build-gcc-sfad12-e3sm/install; else echo "$Albany_ROOT"; fi}</env>
       <env name="Trilinos_ROOT">$SHELL{if [ -z "$Trilinos_ROOT" ]; then echo /lcrc/group/e3sm/ac.jwatkins/LandIce/TrilinosBuilds/build-gcc-e3sm/install; else echo "$Trilinos_ROOT"; fi}</env>
-      <env name="Kokkos_ROOT">$SHELL{if [ -z "$Kokkos_ROOT" ]; then echo /lcrc/group/e3sm/ac.jwatkins/LandIce/TrilinosBuilds/build-gcc-e3sm/install; else echo "$Kokkos_ROOT"; fi}</env>
     </environment_variables>
   </machine>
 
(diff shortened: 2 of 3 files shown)


components/cmake/find_dep_packages.cmake
```diff
@@ -16,10 +16,14 @@
 # errors.
 if (USE_KOKKOS)
 
-  # Kokkos will be built in the sharedlibs if Kokkos_ROOT is
-  # unset.
   if (NOT DEFINED ENV{Kokkos_ROOT})
-    set(ENV{Kokkos_ROOT} ${INSTALL_SHAREDPATH})
+    # We should use the kokkos from Trilinos if we are using Trilinos
+    if (USE_TRILINOS)
+      set (ENV{Kokkos_ROOT} ${Trilinos_ROOT})
```
jewatkins, line 22: One potential issue with this is that when we switch to the preinstalled kokkos, we'll need a new path. But maybe I can change this then by creating something like `Trilinos_Kokkos_ROOT`.

mperego:
What's the long term solution when we'll have different components using Kokkos?

components/cmake/find_dep_packages.cmake
jgfouca: @bartgol , those latter two blocks should be `elseif`s , right?

bartgol: Doesn't matter. The calls to find_package return immediately if the pkg was already found anyways...

rljacob:
Yes what happens when EAMxx and MALI are both running?

bartgol:
@rljacob we first need to answer another question: should e3sm force all components to use the same backend in kokkos? Or should they be free to use different ones? I think the latter is the right choice.

That said, we could potentially install a prebuilt kokkos with all "reasonable" backends enabled (that is, Serial/OpenMP on CPU machines, Serial/OpenMP/Cuda on pm-gpu, Serial/OpenMP/HIP on Frontier,...). Then, each component is free to use whatever backend they need from kokkos. If MALI was installed with both Serial/OpenMP enabled, then eamxx will work just fine with compile_threaded on or off, regardless of whether MALI uses OpenMP or Serial.

jgfouca:
> Or should they be free to use different ones? I think the latter is the right choice

I agree although we will all have to be really careful to explicitly use the device we want to use. The default device will change depending on Kokkos cmake settings.

bartgol:
> > Or should they be free to use different ones? I think the latter is the right choice
> 
> I agree although we will all have to be really careful to explicitly use the device we want to use. The default device will change depending on Kokkos cmake settings.

Yes, using the default device should not be allowed. I don't know how to tell different components to use different devices (like "lnd use OpenMP, atm use Cuda, ..."), or how to force components to use Serial if compile_threaded=false. I don't even know exactly what we want to achieve. I suppose that if component A _can_ run on GPU, we will never ask it to run on CPU on a gpu machine, right? The only catch is then how to prevent use of OpenMP if compile_threaded=False. I think we could have E3SM define the CMake var E3SM_KOKKOS_DEVICE, which components can use to pick a pre-defined device. E.g., e3sm can set E3SM_KOKKOS_DEVICE=Serial if compile_threaded=False (on a CPU machine only, since on GPU we prob want to use the GPU device).

jewatkins:
> If MALI was installed with both Serial/OpenMP enabled, then eamxx will work just fine with compile_threaded on or off, regardless of whether MALI uses OpenMP or Serial.

If we can preinstall kokkos with Serial/OpenMP and have compile_threaded use the appropriate device for both eamxx/MALI, that seems to be the solution. I'm not sure if that's possible with preinstalled trilinos/albany. At worst, we'd have to preinstall two different versions of each. Alternatively, we could force trilinos/albany to use a specific backend regardless of e3sm parameters. It probably does make sense to have all backends though to be able to control which component runs on CPU/GPU.

ndkeen:
When I try to build on pm-cpu with `ERP_Ln9.ne4pg2_ne4pg2.F2010-SCREAMv1.pm-cpu_gnu`

```
/global/cfs/cdirs/e3sm/ndk/repos/jgfouca_cleanup_kokkos_tril/components/eamxx/src/share/atm_process/atmosphere_process_dag.cpp: In lambda function:
/global/cfs/cdirs/e3sm/ndk/repos/jgfouca_cleanup_kokkos_tril/components/eamxx/src/share/atm_process/atmosphere_process_dag.cpp:149:37: error: 'const Units' {aka 'const class ekat::units::Units'} has no member named 'get_string'\
; did you mean 'get_si_string'?
  149 |         s += " [" + fid.get_units().get_string() + "]";
      |                                     ^~~~~~~~~~
      |                                     get_si_string
```

If I replace all of those get_string with get_si_string, it gets further

```
./eamxx/src/share/atm_process/atmosphere_process_dag.cpp:149:        s += " [" + fid.get_units().get_string() + "]";
./eamxx/src/share/field/field_identifier.cpp:51:  m_identifier += " [" + m_units.get_string() + "]";
./eamxx/src/share/io/scorpio_output.cpp:965:    std::string units = fid.get_units().get_string();
```

but still fails:

```
/global/cfs/cdirs/e3sm/ndk/repos/jgfouca_cleanup_kokkos_tril/components/eamxx/src/share/field/field_manager.cpp:49:69: error: invalid initialization of reference of type 'const scream::FieldLayout&' from expression of type 'const Un\
its' {aka 'const ekat::units::Units'}
   49 |           "         - input field units:  " + to_string(id.get_units()) + "\n"
      |                                                         ~~~~~~~~~~~~^~
```

jewatkins:
> I'm not sure if that's possible with preinstalled trilinos/albany. At worst, we'd have to preinstall two different versions of each.

actually, I'm not sure if two different versions would work either if we use kokkos default device and we don't have a "TRILINOS_KOKKOS_DEVICE" or "ALBANY_KOKKOS_DEVICE".

rljacob:
I don't think EAMxx and MALI will appear in the same compset until low-res EAMxx is working so an intermediate solution is fine for now.

bartgol:
@jewatkins yeah, if you want MALI to honor compile_threaded, then we need multiple MALI installations. I don't see a way out of that.

That said, if compile_threaded=False, we could have E3SM set OMP_NUM_THREADS=1, so that even if MALI used OpenMP all the time, it would still result in a single-thread run (Serial is not the same as OpenMP with 1 thread in terms of perf, but it should be bfb the same).

ndkeen:
Not sure if this is what you meant, but with E3SM, builds without openmp are *not* BFB with builds using openmp and only 1 thread.  Only threaded builds are BFB with each other.

bartgol:
> Not sure if this is what you meant, but with E3SM, builds without openmp are _not_ BFB with builds using openmp and only 1 thread. Only threaded builds are BFB with each other.

O_O

Isn't this a bit concerning?

ndkeen:
Yes, it's always bugged me, but there were some strange openmp decisions made long ago (before I joined).  Was kinda hoping we would be rid of them once moved to C++/kokkos.

bartgol:
Well, in all fairness, it is possible that reduction kernels are handled slightly differently in Serial vs OpenMP. Though that's a remote possibility..

bartgol:
@jgfouca PR #6514 got merged, which had some overlap with this one. I am not sure if we still need this PR. I think 6514 handled kokkos differently, but what is done there is I think covering what this PR was doing, so maybe this is no longer needed. But I'll let you take a look and decide. In any case, there are some conflicts now that 6514 went in.

jgfouca:
I don't think we still need it.

PR #6459, opened 2024-06-04
Update mapping and domain files created with an incorrect T62 scrip file

jonbob:
Several configurations have mapping and domain files made from a T62 scrip file that did not extend to the North pole, which is incorrect and causes some problems in the moab driver. This PR updates those files using a correct T62 scrip file for currently used resolutions.

[NML]
[non-BFB] for tests using T62

PR #6482, opened 2024-06-21
Data ice fixes for moab driver

iulian787:
Extend moab driver for data ice cases
case tested: --res T62_oQU120 --compset CMPASO-NYF
it uses the same approach as data ocean; data ice mesh on coupler side is instanced from the sea ice domain file, while
on the  data component side, mesh is "point cloud" only, aligned with mct grid
On atmosphere side, it uses the T62 RLL grid that was corrected earlier with 'jonbob/cpl/fix-t62-files' (PR #6459)


rljacob:
merged to next

rljacob (approved):

PR #6484, opened 2024-06-21
On pm-cpu minor pelayout adjustment for hcru_hcru.I20TRGSWCNPRDCTCBC

ndkeen:
For pm-cpu, try using 192 tasks instead of 128 tasks for grid `l%360x720cru`.
The `ERS.hcru_hcru.I20TRGSWCNPRDCTCBC.pm-cpu_intel.elm-erosion` test works, but was running over 30 minutes. 
The test currently uses 1 node (128 tasks) and trying with 192 tasks speeds up the overall test from about 30 min to 19 min.

With SMS test, it will also work with 256 tasks (2 full nodes), but see failure with ERS test.

BFB except for NML changes.

### DIFF
--- a/components/elm/cime_config/config_pes.xml
+++ b/components/elm/cime_config/config_pes.xml
@@ -235,6 +235,21 @@
     </mach>
   </grid>
   <grid name="l%360x720cru">
+    <mach name="pm-cpu|muller-cpu|alvarez">
+      <pes compset="any" pesize="any">
+        <comment>elm: pm-cpu 2 nodes for grid l%360x720cru</comment>
+        <ntasks>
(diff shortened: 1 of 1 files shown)


rljacob (approved):

ndkeen:
The tests that were timing out have been passing last couple of days.
Should I merge this PR? Or make issue about why we can't use a full 2 nodes (128 tasks)?

rljacob:
Do both I guess.

ndkeen:
I was afraid you would say that

ndkeen:
merged to next

ndkeen:
Last night cdash runs look ok. Will bless the 3 tests that have NML diffs (those using `hcru_hcru`)

PR #6481, opened 2024-06-21
Switch global ocean in init mode to E3SM shared constants

xylar:
This merge also migrates the computation of the Coriolis parameter to using E3SM shared constants.

The unused constants module imports are removed from several init mode utilities to ensure that they don't use constants that are inconsistent with E3SM's versions.

I have made local symlinks to the shared constants and kinds file for convenience in building using MPAS-Ocean's not-very-sophisticated build system.

PR #6485, opened 2024-06-24
Enable E3SM shared pi in MPAS-Ocean

cbegeman:
This PR enables the use of E3SM shared constant pi in MPAS-Ocean when run in coupled configurations.

[non-BFB]


cbegeman:
@xylar In addition to your https://github.com/E3SM-Project/E3SM/pull/6481, do we want to include the changes I have here as a separate PR?

rljacob:
Does this change answers for coupled cases?

xylar:
@rljacob, I think these will be answer changing for all runs involving MPAS-Ocean, including coupled runs.  

@cbegeman, that is why I had not opted to make these changes yet, though I think they are very desirable.

cbegeman:
Would you like me to close this PR and reopen at a later date?

xylar:
@cbegeman. let's see what @rljacob says.

rljacob:
Leave it open.  We'll be making the maint-3.0 branch soon and this can go in after that.

xylar:
Great, thanks, @rljacob!

cbegeman:
Thanks, @rljacob and @xylar for your feedback

components/mpas-ocean/src/shared/mpas_ocn_constants.F
```diff
@@ -117,6 +118,7 @@ subroutine ocn_constants_init(configPool, packagePool)!{{{
        !-----------------------------------------------------------------------
 
        T0_Kelvin = 273.16_RKIND               ! zero point for Celsius
+       pi        = 3.141592_RKIND             ! pi
```
xylar, line 121: This would be the value used in standalone (init and forward) mode so I think it needs to still be the same value as `SHR_CONST_PI` in that mode.

```suggestion
       pi        = 3.14159265358979323846_RKIND ! pi
```

xylar (changes requested):
There are still many places in both `shared` and `analysis_members` that use `pii` from the MPAS framework.  These should also be updated to use MPAS-Ocean's `pi` instead.  See include:
https://github.com/search?q=repo%3AE3SM-Project%2FE3SM+pii+path%3A%2F%5Ecomponents%5C%2Fmpas-ocean%5C%2F%2F&type=code

This was taken care of for global ocean init mode in #6481

components/mpas-ocean/src/shared/mpas_ocn_constants.F
cbegeman: And also make the framework's `pii` match this value? `pii` is missing the last 6 decimal digits

cbegeman:
> There are still many places in both shared and analysis_members that use pii from the MPAS framework. 

Thanks for pointing this out! I didn't think to grep for that spelling. I'll make those changes as well.

components/mpas-ocean/src/shared/mpas_ocn_constants.F
xylar: I think we just don't want to use the framework's `pii`.

xylar: Changing that would need to be coordinated with MPAS-Seaice and MALI.

cbegeman: It looks like both MPAS-Seaice and MALI attempt to override the framework's pi in their constants submodule. I looked through each of the components files and only found one instance where the framework's pi was used https://github.com/E3SM-Project/E3SM/blob/41bd0cbe6238e2e5aefb06338faf93f965f303ce/components/mpas-albany-landice/src/mode_forward/mpas_li_advection_fct_shared.F#L468

Given that we have a bit of time, do you think it would be appropriate to broaden the scope of this PR or just ask that the developers of each check this?

proteanplanet:
We are planning a separate PR for sea ice.

cbegeman:
> We are planning a separate PR for sea ice.

Great. I'll confine this PR to only the ocean component then.

jonbob:
@xylar - have your comments been resolved for this PR?

xylar (approved):
@jonbob, yes, thanks this looks great now!  @cbegeman, thanks for the hard work!

jonbob:
Shows expected baseline DIFFs for:

merged to next

rljacob:
won't this show a diff for any compset with a full ocean?

rljacob:
it is.

jonbob:
merged to master and expected DIFFs blessed -- except on anvil, which has yet to report

jonbob:
also blessed on anvil

PR #6487, opened 2024-06-26
For gcp12, update eam ne4 pelayout

ndkeen:
For gcp12, update ne4 eam pelayout to use 96 tasks which needs 2 nodes (previously 1 node).
This is not node-efficient, but can hopefully improve overall testing time as extra_coverage always waits
for `SMS_Ly1.ne4pg2_oQU480.F2010.gcp12_gnu`

Tested with `e3sm_integration`.

As noted in next comment, `PEM_D_Ln5.ne4pg2_oQU480.F2010.gcp12_gnu.eam-implicit_stress` will fail as it looks like this test is not BFB when changing number of tasks.
Otherwise, BFB except NML diff for change in pelayout

### DIFF
--- a/components/eam/cime_config/config_pes.xml
+++ b/components/eam/cime_config/config_pes.xml
@@ -157,7 +157,7 @@
     </mach>
     <mach name="gcp12">
       <pes compset="any" pesize="any">
-        <comment>eam+gcp: default 1 node</comment>
+        <comment>eam+gcp12: default 1 node, 56x1</comment>
         <ntasks>
           <ntasks_atm>-1</ntasks_atm>
           <ntasks_lnd>-1</ntasks_lnd>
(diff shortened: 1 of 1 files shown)


ndkeen:
I now see that the `eam-implicit_stress` SMS test in extra_coverage will fail compare with baseline if I change the pelayout. This prompted me to check:
`PEM_D_Ln5.ne4pg2_oQU480.F2010.gcp12_gnu.eam-implicit_stress`
with current next (ie without my change to pelayout) and it does fail.

rljacob (approved):

ndkeen:
I created https://github.com/E3SM-Project/E3SM/issues/6498

ndkeen:
Merged this to next and the results of extra coverage are expected.  Several NML diffs due to the pelayout change. And the baseline diff is because of the test not behaving well with different MPI's.  Will bless after merging to master.

PR #6491, opened 2024-06-27
Allow changing SCORPIO config opts at compile/build time

jayeshkrishna:
Allow appending SCORPIO config options at build-time
by using value of PIO_CONFIG_OPTS from env_build.xml

[BFB]


dqwu (approved):
Manually tested this support on ANL workstations.

jayeshkrishna:
With this change SCORPIO configuration can be changed by changing PIO_CONFIG_OPTS in env_build.xml.

For e.g., to disable I/O decomposition information in ADIOS output one can use the following xmlchange command,
```
./xmlchange PIO_CONFIG_OPTS="-DADIOS_NO_DECOMPS=ON "
```

PR #6501, opened 2024-07-05
Disable land ice frazil unless ISMF is on

cbegeman:
This PR turns land ice frazil off by default and re-enables it when ice-shelf melt fluxes are enabled in standalone or coupled mode. Land ice frazil is not enabled in data ISMF because frazil fluxes are already included in satellite-derived ISMF products.

[NML]
[non-BFB] when ice-shelf cavities are present in the mesh and (ocn_ismf is off or ocn_ismf is data)
[BFB] when ice-shelf cavities are absent from the mesh or ice-shelf cavities are present in the mesh and (ocn_ismf is coupled or ocn_ismf is internal)

### DIFF
--- a/components/mpas-ocean/bld/build-namelist
+++ b/components/mpas-ocean/bld/build-namelist
@@ -782,7 +782,7 @@ if ($OCN_ISMF eq 'coupled') {
 	add_default($nl, 'config_frazil_under_land_ice', 'val'=>".true.");
 } elsif ($OCN_ISMF eq 'data') {
 	add_default($nl, 'config_land_ice_flux_mode', 'val'=>"data");
-	add_default($nl, 'config_frazil_under_land_ice');
+	add_default($nl, 'config_frazil_under_land_ice', 'val'=>".false.");
 } else {
 	add_default($nl, 'config_land_ice_flux_mode');
 	add_default($nl, 'config_frazil_under_land_ice');
--- a/components/mpas-ocean/bld/build-namelist
+++ b/components/mpas-ocean/bld/build-namelist
@@ -761,7 +761,6 @@ add_default($nl, 'config_self_attraction_and_loading_beta');
 
 add_default($nl, 'config_use_frazil_ice_formation');
 add_default($nl, 'config_frazil_in_open_ocean');
-add_default($nl, 'config_frazil_under_land_ice');
 add_default($nl, 'config_frazil_heat_of_fusion');
 add_default($nl, 'config_frazil_ice_density');
 add_default($nl, 'config_frazil_fractional_thickness_limit');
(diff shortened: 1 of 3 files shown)


cbegeman:
Further discussion located at https://github.com/E3SM-Ocean-Discussion/E3SM/pull/101

xylar (approved):
Approving based on code inspection and previous review in https://github.com/E3SM-Ocean-Discussion/E3SM/pull/101

xylar:
This will affect current production runs so will need to wait for `maint-3.0` to be created.

darincomeau (approved):
Approved based on review in Ocean Discussions PR.

jonbob:
@cbegeman -- would you expect this PR to cause non-BFB behavior for a typical IcoswISC30E3r5 fully-coupled case?

xylar:
@jonbob, that would be expected because there was frazil forming in the otherwise inactive cavities in WCYCL cases.  That is the main reason we wanted to turn it off here -- we don't want that behavior when cavities are inactive.

jonbob:
Thanks @xylar. Do we have a long-ish comparison run to document the changed behavior then? I didn't see anything on the discussions page

darincomeau:
When this first came up several months ago, we quantified the impact of the unintentional land-ice frazil term in a WCYCL1850 case and confirmed it was very small. But we did not actually do a run with the desired behavior added here, where land-ice frazil is off when ice-shelf cavities are otherwise thermodynamically inactive (e.g. WCYCL cases).

I've started that run now on Anvil, will update.

rljacob:
@darincomeau any update?

darincomeau:
>@darincomeau any update?

[The run is in year 20 now](https://acme-climate.atlassian.net/wiki/spaces/PSC/pages/4732452865/20241030.v3.LR.piControl-land-ice-frazil.anvil), I think we were going to assess climate impact (relative to v3 piControl) at year 20 or 30.

darincomeau:
From MPAS-devops call; will evaluate at year 30 to determine if it needs to be run longer.

rljacob:
@darincomeau any update?

darincomeau:
@rljacob sorry for the delay, the run finished 50 years and I've been trying to get comparative analysis run today to post here.

darincomeau:
MPAS-Analysis this PR vs. v3.LR.piControl years 1 - 50:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/E3SMv3_dev/20241030.v3.LR.piControl-land-ice-frazil.anvil/mpas_analysis/vs-piControl/yrs1-50/

xylar:
The results from MPAS-Analysis show ensemble-level variability.  Even in temperature and salinity time series on the Antarctic continental shelf, where the presence or absence of frazil would have the biggest effect, don't show any discernible offset from one another to suggest anything other than ensemble-level differences.

cbegeman:
> The results from MPAS-Analysis show ensemble-level variability. Even in temperature and salinity time series on the Antarctic continental shelf, where the presence or absence of frazil would have the biggest effect, don't show any discernible offset from one another to suggest anything other than ensemble-level differences.

I also took a look at the MPAS-Analysis page and agree with this assessment

rljacob:
@proteanplanet you want this to wait on your approval?

proteanplanet (approved):
Approved, and also flagging @njeffery in reference to another PR on Frazil in the works for BGC.

jonbob:
Testing with:

shows expected NML and baseline DIFFs

merged to next

rljacob:
Some tests with ne30pg2_r05_IcoswISC30E3r5 had no baseline diffs.  Was that expected?

jonbob:
@rljacob - yes, depending on the ocn_ismf setting

xylar:
And thanks @cbegeman for getting this through!

jonbob:
merged to master and expected NML and baseline DIFFs blessed -- except for pm-cpu where only landice_developer has reported

cbegeman:
Thanks for all your testing @jonbob! And thanks @proteanplanet @xylar @darincomeau for your reviews

jonbob:
pm-cpu also reported, and expected DIFFS blessed

Issue #6474, opened 2024-06-13
Bug in VBS SOA calculation 


mingxuanwupnnl:
In subroutine mam_soaexch_vbs_1subarea in components/eam/src/chemistry/modal_aero/modal_aero_amicphys.F90

The calculation of p0_soa_298 is not correct. Temperature of 298 K should be used instead of ambient temperature

      do ll = 1, ntot_soaspec
         igas = igas_soag + ll - 1
         ! convert sat vapor conc from ug/m^3 to mol/m^3 then to mol/liter
         tmpa = (c0_soa_298(ll)*1.0e-6_r8/mw_gas(igas)) * 1.0e-3_r8  
         ! calc sat vapor pressure (atm) from molar-conc and temp [ 0.082056 = gas constant in (atm/deg-K/(mol/liter)) ]
         p0_soa_298(ll) = 0.082056_r8*tmpa*temp  
      end do

Initial tests results show that it has very small affect on SOA burden

PR #6502, opened 2024-07-07
Fix a bug in Bug in VBS SOA gas-aerosol partition calculation

mingxuanwupnnl:
This bug fix corrects issue #6474 

In subroutine mam_soaexch_vbs_1subarea in components/eam/src/chemistry/modal_aero/modal_aero_amicphys.F90

The calculation of p0_soa_298 is not correct. Temperature of 298 K should be used instead of ambient temperature

This fix has very small impact on global mean climate

See e3sm_diags:

and IICE comparison:

### DIFF
--- a/components/eam/src/chemistry/modal_aero/modal_aero_amicphys.F90
+++ b/components/eam/src/chemistry/modal_aero/modal_aero_amicphys.F90
@@ -3530,7 +3530,7 @@ do_newnuc_if_block50: &
          ! convert sat vapor conc from ug/m^3 to mol/m^3 then to mol/liter
          tmpa = (c0_soa_298(ll)*1.0e-6_r8/mw_gas(igas)) * 1.0e-3_r8  
          ! calc sat vapor pressure (atm) from molar-conc and temp [ 0.082056 = gas constant in (atm/deg-K/(mol/liter)) ]
-         p0_soa_298(ll) = 0.082056_r8*tmpa*temp  
+         p0_soa_298(ll) = 0.082056_r8*tmpa*298.0_r8 
       end do
 
 ! calc soa gas saturation molar-mixing-ratio at local temp and air-pressure


hwangacme (approved):

mkshriva (approved):
Sounds good. As expected, the bug has minimal impact. I agree it should use 298 K instead of temp on line 3533

rljacob:
@wlin7 is this ready to go?   Could go on next today.

wlin7:
Merged to next. The fix is straightforward; it will impact all configs that use vbs soa (standard v3atm).

rljacob:
Noting that 2 of the non-BFB tests flagged this PR.
https://web.lcrc.anl.gov/public/e3sm/e3smtest/evv/MVK_PS.ne4pg2_oQU480.F2010.chrysalis_intel.C.JNextNbfb20241030_010350/index.html

Only variable rejected was soa_a1_SRF

Also https://web.lcrc.anl.gov/public/e3sm/e3smtest/evv/TSC_PS.ne4pg2_oQU480.F2010.chrysalis_intel.C.JNextNbfb20241030_010350/index.html
only variable num_a1

rljacob:
@mingxuanwupnnl would you expect this to be climate changing?  TSC tests indicate it is.

mingxuanwupnnl:
> @mingxuanwupnnl would you expect this to be climate changing? TSC tests indicate it is.

I don't think this bug fix will change the climate significantly. Not sure if I understand the test correctly. Does it mean that only soa_a1_SRF and num_a1 are climate changing?

wlin7:
This may not be an indication of climate changing. The test is for time step convergence, with a very small time step size (2 seconds) for an ne4pg2 grid. `num_a1` is deemed non-convergent during much of the 10-minute  simulation but not always . `NUMLIQ` was also detected but only for the 1st step. Look like the nonconvergence in num_a1 does not propagate to other fields.

rljacob:
The MVK test passed last night so it must be on the borderline.  But TSC is still indicating a fail.

rljacob:
TSC is consistently diffing from this PR.  @wlin7 please make a bless request.

PR #6503, opened 2024-07-08
restore topo smoothing defaults

mt5555:
This PR fixes accidental merge of commits b1667309  and ea28b9d59 into master

mt5555 accidentally pushed to master a branch that adds additional experimental options to the topo smoothing code and changes the default behavior (breaking the homme_tool test in cdash).  In consultation with @rljacob , it was decided not to revert but instead fix the default behavior with a correct PR to next and master.   

Add (disabled by default) ability to apply a local limiter to the topo smoothing operator used when generating topography files for new grids

[BFB]


mt5555:
since some of the commits in this branch are already on master, to see the diff of what will be going into next to match master:    git diff 58efb0cb3f..b1667309

PR #6479, opened 2024-06-18
Change hi-res tests to use the v3 grid configurations

jonbob:
The current high-resolutions tests use older grid configurations. This updates the tests to use the new RRSwISC6to18E3r5 mesh as well as r025 for trigrid tests.

[BFB]

PR #6504, opened 2024-07-08
Change fully-coupled HR test to WCYCL1850NS

jonbob:
This was intended to be part of [PR #6479](https://github.com/E3SM-Project/E3SM/pull/6479) but I missed that commit. It changes the current fully-active high-res test from WCYCL1850 to WCYCL1850NS because the ocn and ice do not have spunup initial conditions defined yet for the RRSwISC6to18E3r5 mesh.


jonbob:

jonbob:
merged to master

PR #6508, opened 2024-07-11
Add subglacial discharge at grounding line

irenavankova:
This introduces subglacial discharge (at the moment a fixed file interpolated from MALI output) at the grounding line as a freshwater volume flux at local freezing point. It follows closely what river runoff does.

https://github.com/E3SM-Ocean-Discussion/E3SM/pull/82

[NML]
[BFB]


components/mpas-ocean/src/analysis_members/mpas_ocn_conservation_check.F
```diff
@@ -699,6 +715,9 @@ subroutine energy_conservation(domain, err)
               + accumulatedRiverRunoffTemperatureFlux   *rho_sw*cp_sw &
               + accumulatedIcebergTemperatureFlux*rho_sw*cp_sw
               ! note, accumulatedLandIceFrazilHeatFlux not added because already in accumulatedFrazilHeatFlux
+         if (trim(config_subglacial_runoff_mode) == 'data') then
+            netEnergyFlux = netEnergyFlux + accumulatedSubglacialRunoffTemperatureFlux * rho_sw*cp_sw
+         end if
```
cbegeman, line 720: I think we can treat this in the same way as landIceHeatFlux -- remove the if-clause here because it will be 0 if there are no SGR fluxes.

components/mpas-ocean/src/Registry.xml
```diff
@@ -3760,6 +3807,10 @@
 			 description="Fresh water flux from river runoff at cell centers from coupler. Positive into the ocean."
 			 packages="thicknessBulkPKG"
 		/>
+		<var name="subglacialRunoffFlux" type="real" dimensions="nCells Time" units="m^3 s^-1"
```
cbegeman, line 3810: I would really rather have this be the same units, i.e., that all of these be surface mass fluxes rather than one being a surface volume flux.

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
```diff
@@ -3465,6 +3504,39 @@ subroutine ocn_compute_KPP_input_fields(statePool, forcingPool, &
                       - exp( max(-100.0_RKIND, &
                                  -layerThickness(kmin, iCell)/ &
                            config_flux_attenuation_coefficient_runoff))
+         if (trim(config_subglacial_runoff_mode) == 'data') then
+             if (config_use_sgr_opt_kpp == 1) then
```
cbegeman, line 3508: Would you be open to checking that this is functioning correctly by adapting this test case, polaris's ocean/single_column/cvmix, to that purpose? I think it would work just fine to have SGR flux input at the surface of this non-ice-covered case. Or let me know if you've already done a case that helps verify the KPP implementation.

components/mpas-ocean/src/analysis_members/mpas_ocn_conservation_check.F
irenavankova: The if clause is necessary because accumulatedSubglacialRunoffTemperatureFlux is defined in a package dataSubglacialRunoffFluxPKG only if trim(config_subglacial_runoff_mode) == 'data'

components/mpas-ocean/src/Registry.xml
```diff
@@ -702,6 +702,39 @@
...
+		/>
+		<nml_option name="config_sgr_flux_vertical_location" type="character" default_value="bottom"
+					description="Selects the vertical location where subglacial runoff is fluxed."
+					possible_values="'top','uniform', 'bottom'"
+		/>
+		<nml_option name="config_use_sgr_opt_kpp" type="integer" default_value="1" units="non-dimensional"
+					description="option to use or not subglacial runoff (sgr) in kpp calculation"
+					possible_values="0 - don't use, 1 - use"
+		/>
+		<nml_option name="config_use_sgr_opt_temp" type="character" default_value="sgr"
```
cbegeman, line 721: I like to keep the `config_use_*` options logical. So in this case it could be `config_use_sgr_temp_prescribed` or something like that.

components/mpas-ocean/src/Registry.xml
```diff
@@ -702,6 +702,39 @@
...
+		/>
+		<nml_option name="config_flux_attenuation_coefficient_subglacial_runoff" type="real" default_value="0.001" units="m"
+					description="The length scale of exponential decay of subglacial runoff. Fluxes are multiplied by $e^{z/\gamma}$, where this coefficient is $\gamma$."
+					possible_values="Any positive real number."
+		/>
+		<nml_option name="config_sgr_flux_vertical_location" type="character" default_value="bottom"
+					description="Selects the vertical location where subglacial runoff is fluxed."
+					possible_values="'top','uniform', 'bottom'"
+		/>
+		<nml_option name="config_use_sgr_opt_kpp" type="integer" default_value="1" units="non-dimensional"
```
cbegeman, line 717: I like to keep the config_use_* options logical, and then in the init routine for a particular module you can create an int temporary variable.

components/mpas-ocean/src/Registry.xml
irenavankova: You mean riverRunoffFlux have same units as subglacialRunoffFlux? Yes, I thought about that too, but I think this depends on the model that provides these inputs and the units it does that in. The subglacialRunoffFlux is provide by MALI which gives me volume flux. To get that into the same units as riverRunoffFlux, I would have to multiply by density during preprocessing and divide by density again in mpas_ocn_surface_bulk_forcing.F, which leaves some room for error (e.g. using slightly different values during preprocessing and then in the mpas code). If you think it is too confusing as is, I could rename subglacialRunoffFlux to subglacialRunoffVolumeFlux if that would help?

components/mpas-ocean/src/Registry.xml
```diff
@@ -702,6 +702,39 @@
 					description="The length scale of exponential decay of river runoff. Fluxes are multiplied by $e^{z/\gamma}$, where this coefficient is $\gamma$."
 					possible_values="Any positive real number."
 		/>
+		<nml_option name="config_subglacial_runoff_mode" type="character" default_value="off"
+					description="Selects the mode in which subglacial runoff fluxes are implemented."
+					possible_values="'off', 'data'"
+		/>
+		<nml_option name="config_flux_attenuation_coefficient_subglacial_runoff" type="real" default_value="0.001" units="m"
+					description="The length scale of exponential decay of subglacial runoff. Fluxes are multiplied by $e^{z/\gamma}$, where this coefficient is $\gamma$."
```
cbegeman, line 710:
```suggestion
					description="The length scale of exponential decay of subglacial runoff, used when config_sgr_flux_vertical_location is 'top' or 'bottom'. Fluxes are multiplied by $e^{z/\gamma}$, where this coefficient is $\gamma$."
```

components/mpas-ocean/src/Registry.xml
```diff
@@ -702,6 +702,39 @@
...
+		/>
+		<nml_option name="config_use_sgr_opt_temp" type="character" default_value="sgr"
+					description="option to choose sgr temperature"
+					possible_values="'sgr' - this is 0 salinity, 'prescribed' - this salinity specified in config_sgr_salinity"
+		/>
+		<nml_option name="config_use_sgr_opt_salt" type="character" default_value="sgr"
+					description="option to choose sgr salinity"
+					possible_values="'sgr' - this is temperature at local freezing point, 'prescribed' - this temperature specified in config_sgr_temperature"
+		/>
+		<nml_option name="config_sgr_temperature" type="real" default_value="0.0" units="C"
```
cbegeman, line 729:
```suggestion
		<nml_option name="config_sgr_temperature_prescribed" type="real" default_value="0.0" units="C"
```

components/mpas-ocean/src/Registry.xml
```diff
@@ -702,6 +702,39 @@
...
+			        description="option to set fixed sgr temperature, applied when config_use_sgr_opt_temp = 'prescribed'"
+					possible_values="Any real number."
+		/>
+		<nml_option name="config_sgr_salinity" type="real" default_value="0.0" units="1.e-3"
```
cbegeman, line 733:
```suggestion
		<nml_option name="config_sgr_salinity_prescribed" type="real" default_value="0.0" units="1.e-3"
```

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
irenavankova: I am not sure what you mean and what that case would reveal, we can discuss that further, also I haven't worked with Polaris yet. I did test the option with config_use_sgr_opt_kpp on and off in compass Ocean0.

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
```diff
@@ -3465,6 +3504,39 @@ subroutine ocn_compute_KPP_input_fields(statePool, forcingPool, &
...
+                      zBot = zTop - layerThickness(k,iCell)
+                      if (k == minLevelCell(iCell)) then
+                        transmissionCoeffTop = exp( max(zTop / config_flux_attenuation_coefficient_runoff, -100.0_RKIND) )
+                        transmissionCoeffBot = exp( max(zBot / config_flux_attenuation_coefficient_runoff, -100.0_RKIND) )
+                        fracAbsorbedSubglacialRunoff = transmissionCoeffTop - transmissionCoeffBot
+                      end if
+                      zTop = zBot
+                   end do
+                else
+                    call mpas_log_write("config_sgr_flux_vertical_location not one of 'bottom', 'uniform', 'top'.", MPAS_LOG_CRIT)
```
cbegeman, line 3534: I prefer to have all namelist checks in init routines so that they occur once at the outset of the forward run.

irenavankova: We discussed with @xylar and he said in this case it was not necessary but I don't have a strong opinion here.

cbegeman (changes requested):
@irenavankova Exciting work! I made a few comments, mostly minor with the exception of the units of the runoff flux. Not sure whether you want to pre-process the data file. I'm a bit of a stickler about namelist options because we've had a lot of user-error surrounding them in the past.

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
cbegeman: Ok. Was the comparison between config_use_sgr_opt_kpp on and off in compass Ocean0 useful for verification purposes? If not, I can set up and test the polaris idealized case and explain further how it's useful.

irenavankova: There were a few tiny changes in T and S around the plume area, but overall it made little difference whether that option was on or off.

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
xylar: The issue is that there's not currently the need for an init routine so it seems overkill to add it for this one check.

components/mpas-ocean/src/Registry.xml
irenavankova: The file Alex Hager (can't tag him for some reason) gave me has totalGroundingLineDischargeCell in m^3/s maybe him or @matthewhoffman can comment on this?

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
cbegeman: You can add it to either this init routine: https://github.com/E3SM-Project/E3SM/blob/7491a836e534a13d4ef1d59ae98b6f36b4f92870/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F#L4628
or this one: https://github.com/E3SM-Project/E3SM/blob/7491a836e534a13d4ef1d59ae98b6f36b4f92870/components/mpas-ocean/src/shared/mpas_ocn_surface_bulk_forcing.F#L400

Both are called in forward mode init and have all config options available.

components/mpas-ocean/src/Registry.xml
cbegeman: The density values are not supposed to be consistent. Mass fluxes into MPAS-Ocean are assumed to have the reference density, and that's what we should be dividing by, while on the MALI side we should be multiplying by the density of freshwater, as Xylar states.

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
xylar: Ah, yeah, that's a good suggestion.

components/mpas-ocean/src/Registry.xml
xylar: Yes, exactly.

alexolinhager: MALI calculates that field by summing the volumes fluxes from both the distributed and channelized systems. So I just used the units hardcoded into MALI without multiplying by density. MALI uses a freshwater density of 1000 kg/m3 though, so you can use that density to convert to a mass flux

irenavankova: Ok, I will just divide by freshwater density and areaCell during preprocessing so that in the MPAS-O code subglacialRunoffFlux has same units and is treated exactly the same way as riverRunoffFlux

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
irenavankova: ok, I put it here: ocn_diagnostics_init

components/mpas-ocean/src/Registry.xml
irenavankova: Changed as suggested.

components/mpas-ocean/src/Registry.xml
irenavankova: Changed as suggested.

components/mpas-ocean/src/Registry.xml
irenavankova: Changed as suggested.

components/mpas-ocean/src/Registry.xml
irenavankova: Changed as suggested.

irenavankova:
@xylar and @cbegeman, I made and tested the suggested changes so I think it is ready for you to look at that again.

components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F
```diff
@@ -4598,6 +4693,13 @@ subroutine ocn_diagnostics_init(domain, err)!{{{
       call ocn_diagnostics_variables_init(domain, jenkinsOn, &
                                           hollandJenkinsOn, err)
 
+      if ( trim(config_sgr_flux_vertical_location) == 'top' ) then
+      else if ( trim(config_sgr_flux_vertical_location) == 'uniform' ) then
+      else if ( trim(config_sgr_flux_vertical_location) == 'bottom' ) then
+      else
```
cbegeman, line 4699:
```suggestion
      if ( trim(config_sgr_flux_vertical_location) /= 'top' .and. &
           trim(config_sgr_flux_vertical_location) /= 'uniform' .and. &
           trim(config_sgr_flux_vertical_location) /= 'bottom' ) then
```

xylar:
@irenavankova, I tried to run a test, `SMS_D_P480_Ld1.ne30pg2_r05_IcoswISC30E3r5.CRYO1850-DISMF.chrysalis_intel`, but I'm seeing:
```
401: forrtl: severe (408): fort: (7): Attempt to use pointer SURFACEFLUXATTENUATIONCOEFFICIENTSUBGLACIALRUNOFF when it is not associated with a target
401:
401: Image              PC                Routine            Line        Source
401: libpnetcdf.so.3.0  000015554B9391A2  for_emit_diagnost     Unknown  Unknown
401: e3sm.exe           000000000C84FBAC  ocn_diagnostics_m         398  mpas_ocn_diagnostics.f90
401: e3sm.exe           000000000E18F830  ocn_init_routines         168  mpas_ocn_init_routines.f90
401: e3sm.exe           000000000E184542  ocn_forward_mode_         329  mpas_ocn_forward_mode.f90
401: e3sm.exe           000000000E38A3B4  ocn_core_mp_ocn_c          86  mpas_ocn_core.f90
401: e3sm.exe           000000000C4A55F5  ocn_comp_mct_mp_o         563  ocn_comp_mct.f90
401: e3sm.exe           0000000000497CCF  component_mod_mp_         257  component_mod.F90
401: e3sm.exe           000000000043AAAB  cime_comp_mod_mp_        1506  cime_comp_mod.F90
401: e3sm.exe           000000000048E76C  MAIN__                    122  cime_driver.F90
401: e3sm.exe           0000000000426CE2  Unknown               Unknown  Unknown
401: libc-2.28.so       0000155545051D85  __libc_start_main     Unknown  Unknown
401: e3sm.exe           0000000000426BEE  Unknown               Unknown  Unknown
```
The error on line 398 of `mpas_ocn_diagnostics.f90` (which is the post-processed file after handling all the `#ifdef` flags and such) is referring to this call:
https://github.com/E3SM-Project/E3SM/blob/b45249581205070d1f39e2400cc574907a4ab6d8/components/mpas-ocean/src/shared/mpas_ocn_diagnostics.F#L457-L461
It seems like passing a pointer with intent(out) requires that it be valid.

One solution might be to just construct `surfaceFluxAttenuationCoefficientSubglacialRunoff` in its own subroutine and have that subroutine call inside of an:
``` fortran
      if (trim(config_subglacial_runoff_mode) == 'data') then
         ...
      end if
```

irenavankova:
@xylar, I introduced new functions for everything that takes a subglacial variable as an argument. It finally passed the SMS test in debug mode. And I also tested in standalone with ISOMIP and checked conservation, and all looks fine.

components/mpas-ocean/src/shared/mpas_ocn_thick_surface_flux.F
```diff
@@ -169,13 +170,114 @@ subroutine ocn_thick_surface_flux_tend(transmissionCoefficients, transmissionCoe
...
+
 !***********************************************************************
 !
 !  routine ocn_thick_surface_flux_init
 !
 !> \brief   Initializes ocean horizontal thickness surface fluxes
-!> \author  Doug Jacobsen
-!> \date    12/17/12
+!> \author  Irena Vankova
+!> \date    July 2024
```
xylar, line 280: This should probably be changed back.

components/mpas-ocean/src/shared/mpas_ocn_tracer_surface_flux_to_tend.F
```diff
@@ -197,11 +198,120 @@ subroutine ocn_tracer_surface_flux_tend(meshPool, fractionAbsorbed, fractionAbso
...
    end subroutine ocn_tracer_surface_flux_tend!}}}
 
+
+!***********************************************************************
+!
+!  routine ocn_tracer_surface_flux_tend_subglacial_runoff
+!
+!> \brief   Computes tendency term for surface fluxes
+!> \author  Doug Jacobsen
+!> \date    12/17/12
```
xylar, line 212: This should be updated

xylar (approved):
I'm approving with one small typo and a couple of authorships to fix.

I ran a BFB test using:
```
SMS_D_P480_Ld1.ne30pg2_r05_IcoswISC30E3r5.WCYCL1850.chrysalis_intel
```
As expected, there were namelist differences but other tests passed.

Then, I ran:
```
SMS_P128_Lm1.T62_oQU240wLI.GMPAS-NYF-PISMF-DSGR.chrysalis_intel
SMS_P512_Lm1.TL319_ECwISC30to60E2r1.GMPAS-JRA1p5-DIB-PISMF-DSGR.chrysalis_intel
SMS_P480_Lm1.TL319_IcoswISC30E3r5.GMPAS-JRA1p5-DIB-PISMF-DSGR.chrysalis_intel
SMS_P768_Lm1.TL319_SOwISC12to60E2r4.GMPAS-JRA1p5-DIB-PISMF-DSGR-TMIX.chrysalis_intel
```
They all ran successfully for 1 month.  I took a look at the `timeSeriesStatsMonthly` file to make sure the following fields looked sensible:
```
timeMonthly_avg_subglacialRunoffFlux timeMonthly_avg_activeTracersSurfaceFluxSubglacialRunoff_temperatureSurfaceFluxSubglacialRunoff
```
and they did in all 4 cases.

I looked through the code changes, particularly focused on the new subroutines you added most recently.  The changes look good.

xylar:
@jonbob, can you check this for namelist consistency using your usual procedure?

jonbob:
@xylar -- done. It looked like just a different ordering, but the bld files are now consistent

xylar:
@jonbob, I just talked with @irenavankova and she's going to add a "stealth" test for this. It's going to be almost identical to:
https://github.com/E3SM-Project/E3SM/blob/master/cime_config/tests.py#L276
but with `DSGR` instead of `TMIX`.  As with the PR where we added the `TMIX` feature (#6306), this feature isn't quite stealth but it's close enough to make sense in that particular test group.

components/mpas-ocean/src/shared/mpas_ocn_thick_surface_flux.F
xylar: (This is not actually a subroutine you edited.)

components/mpas-ocean/src/shared/mpas_ocn_tracer_surface_flux_to_tend.F
xylar: (This is a subroutine that you wrote, not Doug ;-).)

jonbob:
@xylar -- thanks, I think that makes sense

cime_config/tests.py
```diff
@@ -274,6 +274,7 @@
             "ERS_P480_Ld5.TL319_IcoswISC30E3r5.GMPAS-JRA1p5-DIB-PISMF.mpaso-jra_1958",
             "PEM_P480_Ld5.TL319_IcoswISC30E3r5.GMPAS-JRA1p5-DIB-PISMF.mpaso-jra_1958",
             "SMS_P480_Ld5.TL319_IcoswISC30E3r5.GMPAS-JRA1p5-DIB-PISMF-TMIX.mpaso-jra_1958",
+            "SMS_P480_Ld5.TL319_IcoswISC30E3r5.GMPAS-JRA1p5-DIB-PISMF-DSGR.mpaso-jra_1958",
```
xylar, line 277: Perfect!  This is nearly identical to the test I ran:
```
SMS_P480_Lm1.TL319_IcoswISC30E3r5.GMPAS-JRA1p5-DIB-PISMF-DSGR.chrysalis_intel
```
It just brings in one year of JRA forcing and runs for only 5 days, not a month.  So I'm sure it will work great!

xylar:
@irenavankova, I think you accidentally took out @jonbob's commit that fixed the order of the namelist options, etc.  I'll try to find that commit so you can cherry-pick it back on.

irenavankova:
@jonbob, I made the changes Xylar asked for, so I think this is now yours? Unless there are other things you think I should change.

xylar:
Here's the commit:
https://github.com/E3SM-Project/E3SM/commit/0288938beae15cf92f0a86d077e0cd12b3ffaf3c
Can you do:
```
git cherry-pick 0288938beae15cf92f0a86d077e0cd12b3ffaf3c
```
and see if that works?  You may need to do a git-fetch from your fork for that commit to be available.

xylar:
Oh, sorry, I see @jonbob's commit.  My mistake.

irenavankova:
I accidentally took two commits out but then added them back in right away. So should be good now.

rljacob:
Is this ready?

irenavankova:
By me yes, but I don't know if @jonbob has anything to add.

.gitignore
```diff
@@ -51,3 +51,16 @@ components/eamxx/docs/common/eamxx_params.md
 components/eamxx/src/python/build
 components/eamxx/src/python/build_src
 components/eamxx/src/python/dist
+
```
jonbob, line 54: @irenavankova - I don't think this should be in your PR

xylar: Yes, I agree.  I was going to let this be because these are all sensible things to ignore but I think we'd better leave them out or make them in a separate PR.

xylar:
@irenavankova is away but I think she would like to keep this moving along so I went ahead and reverted the changes to `.gitignore`.

jonbob:
Passes:
* PET_Ln5.T62_oQU240wLI.GMPAS-DIB-IAF-PISMF.chrysalis_intel

with expected NML DIFFs. Also successfully runs the new test:
* SMS_P480_Ld5.TL319_IcoswISC30E3r5.GMPAS-JRA1p5-DIB-PISMF-DSGR.chrysalis_intel.mpaso-jra_1958

merged to next

jonbob:
merged to master and all expected NML DFFs have been blessed except on mappy, which did not report today

xylar:
Thanks @jonbob!  I know this has been quite a bit of work.  I'm sure @irenavankova will be very happy to have this in when she gets back.

PR #6509, opened 2024-07-11
Fixes gotm interface for vertical mixing

### LINKED ISSUE #6507: Compute KPP input fields
@vanroekel, I think the call below to compute necessary input fields for CVMix should be placed outside the if statement checking `config_cvmix_kpp_nonlocal_with_implicit_mix`, or somewhere else? This seems to be the only place where this subroutine is called and the computed fields `surfaceBuoyancyForcing` and `surfaceFrictionVelocity` are required by other CVMix subroutines. What do you think?

https://github.com/E3SM-Project/E3SM/blob/c7d7998ab483f09bd17b8c399345df8fd247c6f4/components/mpas-ocean/src/shared/mpas_ocn_tendency.F#L1249-L1250

vanroekel:
if GOTM is enabled surface friction velocity is not calculated at present.  This fixes that issue by allowing KPP_input_fields to be calculated for cvmix and gotm

Fixes #6507
[BFB]

### DIFF
--- a/components/mpas-ocean/src/shared/mpas_ocn_tendency.F
+++ b/components/mpas-ocean/src/shared/mpas_ocn_tendency.F
@@ -1243,12 +1243,12 @@ contains
          !
          ! Compute tracer tendency due to non-local flux computed in KPP
          !
-         if (config_use_cvmix_kpp) then
-            call mpas_timer_start("non-local flux from KPP")
-            if (.not. config_cvmix_kpp_nonlocal_with_implicit_mix) then
-               call ocn_compute_KPP_input_fields(statePool, forcingPool,&
+         if (config_use_cvmix_kpp .or. config_use_gotm) then
+            call ocn_compute_KPP_input_fields(statePool, forcingPool,&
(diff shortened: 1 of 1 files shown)


vanroekel:
This only affects simulations where `config_use_gotm=.true.` and `config_use_cvmix_kpp=.false.`. No E3SM configurations use this at present

qingli411:
This looks good to me.

rljacob:
@mark-petersen please review.

mark-petersen (approved):
Yes, this looks right. If GOTM is on then the KPP input fields are computed. This does not affect any E3SM simulations, which don't use GOTM. I compiled stand-alone with gnu on chicoma and intel on chrysalis without any trouble.

jonbob (approved):
Approved based on visual inspection and testing

jonbob:
Passes:
* PET_Ln5.T62_oQU240wLI.GMPAS-DIB-IAF-PISMF.chrysalis_intel

merged to next

jonbob:
merged to master

PR #6517, opened 2024-07-20
For pm-cpu, update GNU compiler to 12.2

ndkeen:
For pm-cpu, update to current machine default for the GNU compiler (from 11.2 to 12.2).
Add two module unload commands to be cleaner.

For similar machine muller-cpu and alvarez (at NERSC), upgrade several module versions for testing.
Add some compiler flag work-arounds for muller-cpu while testing newer GNU compiler.

Tests so far include e3sm_integration and e3sm_hi_res.
Will not be BFB for GNU tests
[NBFB]

### DIFF
--- a/cime_config/machines/Depends.muller-cpu.gnu.cmake
+++ b/cime_config/machines/Depends.muller-cpu.gnu.cmake
@@ -8,6 +8,22 @@ if (NOT DEBUG)
   endforeach()
 endif()
 
+# On pm-cpu (and muller-cpu), with gcc-native/12.3, we see hang with DEBUG runs of certain tests.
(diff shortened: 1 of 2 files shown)


ndkeen:
I have not heard of gcc 11 being talked about as a potential module to remove. I just know that gcc 12 has been the machine default for quite some time, so it seemed like a smart thing to do.  

I noted on slack channel, that I've been trying to update all of the module versions to at least current machine defaults, but there are a few nagging issues with several of the updates. So this PR was to at least make some progress and often better to deal with smaller set of changes at a time.

merged to next.

rljacob (approved):

PR #6518, opened 2024-07-20
Improve path for finding csh in shell_commands

ndkeen:
On GCP, csh not found at this location, using `/usr/bin/env csh` instead.
The following tests were failing create_newcase:
```
SMS_R_Ld5.ne4_ne4.FSCM-ARM97.gcp12_gnu.eam-scm
SMS_Ln5.ne4_ne4.FSCM-ARM97-MMF1.gcp12_gnu
```

[BFB]


ndkeen:
Instead, I just created a soft-link for csh on the cluster so that this script can find it. I think it would be better to find a better solution as I suspect this will happen on another machine.

PR #6522, opened 2024-07-23
Redesigned "caar loop pre-boundary exchange", tuned for Frontier 

trey-ornl:
This pull request attempts to provide Frontier optimizations like those used in the 2023 Gordon Bell Climate runs, but with a software architecture that meets the requirements of https://acme-climate.atlassian.net/wiki/x/oICd6. Here is a summary of the changes.
- Create multiple new `struct`s in `SphereOperators.hpp` to use for the new Caar pre-boundary exchange.
- Add new file `CaarFunctorImpl.cpp` that implements the new `caar_compute` function and template functions with Kokkos loops.
- Update `CaarFunctorImpl.hpp` with new functions, different number of buffers, and `#if` to turn on/off the new `caar_compute`. If we adopt the new implementation permanently, significant code can be eliminated from this file.
- Add new `viewAsReal` functions in `ViewUtils.hpp`.
- Remove `LaunchBounds<512,1>` calls, which I think are incorrect for AMD GPUs, where the Kokkos teams sometimes use 1024 threads.
- Update `ExecSpaceDefs.cpp` to use newer Kokkos command-line argument for MPI.
- Move to Cray Programming Environment `cpe/23.12` and Rocm `rocm/5.7.1`.
- Add `frontier.cmake` for standalone Homme builds.
- Delete old Crusher files for CMake.

I confirmed that the modified code passes the `caar_ut` unit test using `frontier-bfb.cmake`, and I ran a single-node performance test of stand-alone Homme provided to me by @oksana, building with the new `frontier.cmake`. Here are the `walltotal` times recorded in `HommeTime_stats` for `"caar compute"`.

Existing Code (`#if 0`): 153.6221 s
New Code (`#if 1`): 4.777188 s

I feel like this improvement is "too much." I confirmed that the outputs of the performance runs matched within roundoff. And I confirmed the performance difference using `rocprof`.

I then compiled `CaarFunctor` with `-Rpass-analysis=kernel-resource-usage`. While all the kernels of the new code had a grand total of just one vector-register spill, the monolithic existing kernel had 156115 scalar-register spills and 2820 vector-register spills using `rocm/5.7.1`. Is this enough to explain the performance gulf? I wonder if current E3SM runs are seeing this slowdown.

Here are a few caveats about this pull request.
- I don't know how it performs on NVidia GPUs. I am hopeful that it increases performance.
- I don't know if it even works on CPUs. It uses standard Kokkos, but it relies on active team parallelism (instead of `TeamThreadRange` loops).


ambrad:
This PR looks like it will be quite a bit of effort to merge in its current state. Perhaps it's premature to post it. Perhaps it would be better to discuss the branch on a Confluence page.

Trey, when you have a chance, would you post full timer output for the two runs in which you see the 32x performance discrepancy?

trey-ornl:
> This PR looks like it will be quite a bit of effort to merge in its current state. Perhaps it's premature to post it. Perhaps it would be better to discuss the branch on a Confluence page before finalizing a PR.
> 
> Trey, when you have a chance, would you post full timer output for the two runs in which you see the 32x performance discrepancy?

I created the following Confluence page for discussion. It includes the full timer output.
https://acme-climate.atlassian.net/wiki/x/JwDvCgE

components/homme/src/theta-l_kokkos/cxx/CaarFunctorImpl.cpp
```diff
@@ -0,0 +1,638 @@
...
+        const Real vtheta_dp = state_vtheta_dp(b.e,data_n0,b.x,b.y,b.z);
+        if ((vtheta_dp < 0) || (dphi > 0)) abort();
+        EquationOfState::compute_pnh_and_exner(vtheta_dp, dphi, buffers_pnh(b.e,b.x,b.y,b.z), buffers_exner(b.e,b.x,b.y,b.z));
+
+        buffers_phi(b.e,b.x,b.y,b.z) = 0.5 * (state_phinh_i(b.e,data_n0,b.x,b.y,b.z) + state_phinh_i(b.e,data_n0,b.x,b.y,b.z+1));
+      }
+
+      const Real v0 = state_v(b.e,data_n0,0,b.x,b.y,b.z) * state_dp3d(b.e,data_n0,b.x,b.y,b.z);
+      const Real v1 = state_v(b.e,data_n0,1,b.x,b.y,b.z) * state_dp3d(b.e,data_n0,b.x,b.y,b.z);
+      buffers_vdp(b.e,0,b.x,b.y,b.z) = v0;
```
bartgol, line 47: The buffers are allocated based on the number of concurrent slots needed, but with SphereBlockOps (as well as with SphereColOps and SphereScanOps) we access them with the element index. We need to polish this out, so that we can use "team_idx" instead of the element index for the buffers.

components/homme/src/share/cxx/SphereOperators.hpp
```diff
@@ -1176,6 +1176,273 @@ class SphereOperators
   Real m_scale_factor_inv, m_laplacian_rigid_factor;
 };
 
+using TeamPolicy = Kokkos::TeamPolicy<ExecSpace>;
+using Team = TeamPolicy::member_type;
+
+static constexpr int NPNP = NP * NP;
+static constexpr int WARP_SIZE = warpSize;
```
bartgol, line 1183: I think warpSize is a HIP specific variable?

components/homme/src/share/cxx/SphereOperators.hpp
```diff
@@ -1176,6 +1176,273 @@ class SphereOperators
...
+    rrdmd(0)
+  {
+    const int lr = t.league_rank();
+    e = lr / SPHERE_BLOCKS_PER_COL;
+    const int iw = lr % SPHERE_BLOCKS_PER_COL;
+    const int tr = t.team_rank();
+    const int ixy = tr / SPHERE_BLOCK_LEV;
+    x = ixy / NP;
+    y = ixy % NP;
+    const int dz = tr % SPHERE_BLOCK_LEV;
```
bartgol, line 1250: This snippet below confuse me. I don't think team_rank() uniquely identifies a thread in a warp, but rather it identifies the warp as a whole. So I'm confused as of why this code can work correctly for all levels.

In Kokkos, team_rank() returns threadIdx.y, but a block (i.e., a team) is 2d.

bartgol (commented):
I have some preliminary comments. As Andrew said, this PR is quite big, and may be worth see if/how we can chunk it. But before then, we should find a way to make things work for Serial/OpenMP/Cuda backends.

components/homme/src/share/cxx/SphereOperators.hpp
```diff
@@ -1176,6 +1176,273 @@ class SphereOperators
...
+  KOKKOS_INLINE_FUNCTION Real div(const SphereBlockScratch &t0, const SphereBlockScratch &t1)
+  {
+    // Separately compute du and dv to match Fortran bfb
+    Real du = 0;
+    Real dv = 0;
+    for (int j = 0; j < NP; j++) {
+      du += dvvy[j] * t0.sv(x,j);
+      dv += dvvx[j] * t1.sv(j,y);
+    }
+    if (rrdmd == 0) rrdmd = (1.0 / metdet) * scale_factor_inv;
```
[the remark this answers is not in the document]
trey-ornl: Divisions are expensive on AMD GPUs. The attempt here is to do it only once per kernel, even if the kernel calls `div` or `fort` more than once, or both `div` and `vort`, and to do it zero times if the kernel never calls `div` or `vort` (but calls only `grad`, for example).

components/homme/src/share/cxx/SphereOperators.hpp
```diff
@@ -1176,6 +1176,273 @@ class SphereOperators
...
+      du += dvvx[j] * t0.sv(j,y);
+      dv += dvvy[j] * t1.sv(x,j);
+    }
+    if (rrdmd == 0) rrdmd = (1.0 / metdet) * scale_factor_inv;
+    return (dv - du) * rrdmd;
+  }
+
+  KOKKOS_INLINE_FUNCTION void vortInit(SphereBlockScratch &t0, SphereBlockScratch &t1, const Real v0, const Real v1) const
+  {
+    t0.sv(x,y) = g.d(e,0,0,x,y) * v0 + g.d(e,0,1,x,y) * v1;
```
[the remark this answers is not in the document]
trey-ornl: Part of the tuning process was trying different combinations of local copies. A local copy of `dinv` happened to show speedup, while a local copy of `d` did not. I hypothesize that the tiny L1 cache happens to be big enough to reuse the required `d` values without making an explicit local copy.

components/homme/src/share/cxx/SphereOperators.hpp
```diff
@@ -1176,6 +1176,273 @@ class SphereOperators
...
+
+struct SphereGlobal {
+
+  const ExecViewManaged<const Real *[2][2][NP][NP]> d;
+  const ExecViewManaged<const Real *[2][2][NP][NP]> dinv;
+  const ExecViewManaged<const Real [NP][NP]> dvv;
+  const ExecViewManaged<const Real *[NP][NP]> metdet;
+  const Real scale_factor_inv;
+
+  SphereGlobal(const SphereOperators &op):
```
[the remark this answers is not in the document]
trey-ornl: When a lambda captures an object, it makes a full copy of that object, regardless of what members the lambda actually uses. The `SphereGlobal` struct attempts to minimize the number of `SphereOperators` members that get copied into the kernel. (When a `Kokkos::View` gets copied, it copies only the array descriptor, not all the values of the array. But `SphereOperators` still has more of them than are actually used by each kernel.)

components/homme/src/share/cxx/SphereOperators.hpp
```diff
@@ -1176,6 +1176,273 @@ class SphereOperators
...
+  SphereBlockScratchView v;
+  SphereBlockScratchSubview sv;
+  KOKKOS_INLINE_FUNCTION SphereBlockScratch(const SphereBlockOps &b);
+};
+
+struct SphereBlockOps {
+  const SphereGlobal &g;
+  const Team &t;
+  Real scale_factor_inv;
+  Real metdet;
```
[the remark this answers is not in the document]
trey-ornl: I don't know the particular reason each local copy wins. It was all empirical. If a value was reused, I tried making a local copy. If the result was faster, I kept it.
Possible reasons are that it keeps the compiler from computing indices or loading values more than once. Integer calculation for array indices can be surprisingly expensive on AMD GPUs.

components/homme/src/share/cxx/SphereOperators.hpp
```diff
@@ -1176,6 +1176,273 @@ class SphereOperators
...
+    const int ixy = tr / SPHERE_BLOCK_LEV;
+    x = ixy / NP;
+    y = ixy % NP;
+    const int dz = tr % SPHERE_BLOCK_LEV;
+    z = dz + iw * SPHERE_BLOCK_LEV;
+
+    metdet = g.metdet(e,x,y);
+    for (int i = 0; i < 2; i++) for (int j = 0; j < 2; j++) dinv[i][j] = g.dinv(e,i,j,x,y);
+    for (int j = 0; j < NP; j++) {
+      dvvx[j] = g.dvv(x,j);
```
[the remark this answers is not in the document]
trey-ornl: Because `SphereBlockOps` objects are on the kernel stack, `dvvx` and `dvvy` only exist in vector registers. Each "copy" is a single load that is required regardless. And Kokkos subviews can be register hogs.

components/homme/src/theta-l_kokkos/cxx/CaarFunctorImpl.cpp
trey-ornl: I use buffers to store values across kernels, so I need separate values for every element. I think you are saying that the buffers are actually allocated based on the number of teams, and the number of teams being the same as the number of local elements is a coincidence that I should not rely upon?

bartgol: Ah, I did not notice that the epochs are now separate kernels (rather than separate parts of the same kernel). That's a big change. It reminds me of what we did in SCREAM for some physics parametrizations, where an implementation with many smaller kernels was better than a monolithic one. But we should definitely see if this impl is also good for CUDA.

That said, yes, the buffers are allocated based on the maximum number of concurrent teams. If the number of (rank-local) elements is large enough, even on GPU we can get num_buffer_slots<num_elements, which means those buffers may be smaller than num_elems. This is pretty much always the case on CPU, btw, so this impl needs to be adjusted to be portable.

components/homme/src/share/cxx/SphereOperators.hpp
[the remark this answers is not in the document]
trey-ornl: Want me to try converting to a stored `rmetdet` and see how performance compares?

components/homme/src/theta-l_kokkos/cxx/CaarFunctorImpl.cpp
trey-ornl: How about the following to-do for me? Convert to using a separate set of buffers, created to be the required sizes.

components/homme/src/share/cxx/SphereOperators.hpp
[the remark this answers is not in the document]
bartgol: No need for that. To avoid doing the division at every kernel call, we'd have to precompute the rmetdet view during init, and storing in the ElementsGeometry structure. I don't think that's a change for this PR. I was just thinking out loud :)

components/homme/src/share/cxx/SphereOperators.hpp
trey-ornl: I think Cuda has it, but it's not a compile-time constant. I need a compile-time workaround. `#ifdef`? Other suggestions?

components/homme/src/theta-l_kokkos/cxx/CaarFunctorImpl.cpp
bartgol: You mean make the new impl use different buffers from the original impl? That would work. But related to that, I am getting a bit concerned about how implementations for different backends may start do dramatically diverge. So yes, do the fix you suggest, but we need to start verification of other backends.

components/homme/src/share/cxx/SphereOperators.hpp
trey-ornl: For `SphereBlockOps::policy`, the `team_rank()` does uniquely identify a thread, because the Kokkos "vector" dimension is one (no third argument in the constructor):
https://github.com/E3SM-Project/E3SM/blob/dd1ffec3501899234573866cdddb0c30c3c7449d/components/homme/src/share/cxx/SphereOperators.hpp#L1326
https://github.com/E3SM-Project/E3SM/blob/dd1ffec3501899234573866cdddb0c30c3c7449d/components/homme/src/share/cxx/SphereOperators.hpp#L1187
Other team policies do have nonzero vector lengths, so their `team_rank()` values do not uniquely identify a thread. This is the case for the existing code and for `SphereScanOps::policy` in the new code (three arguments in constructor):
https://github.com/E3SM-Project/E3SM/blob/dd1ffec3501899234573866cdddb0c30c3c7449d/components/homme/src/share/cxx/SphereOperators.hpp#L1442

components/homme/src/share/cxx/SphereOperators.hpp
bartgol: It's also not defined for serial/openmp, which means we do need some sort of CPP guard. I guess `#ifdef` would do.

rljacob:
Is this ready?

trey-ornl:
> Is this ready?

I doubt it. It has only been test on Frontier.

rljacob:
How do you want to test this on other machines?  @trey-ornl do you have perlmutter access?

components/homme/src/share/cxx/SphereOperators.hpp
```diff
@@ -1176,6 +1176,279 @@ class SphereOperators
...
+{}
+
+struct SphereCol {
+  const Team &t;
+  int e,x,y,z;
+
+  KOKKOS_INLINE_FUNCTION SphereCol(const Team &team):
+    t(team)
+  {
+    const int lr = t.league_rank();
```
bartgol, line 1350: Quick question: this seems to assume that the team policy has league size `num_elems*NP*NP`, right? If so, see my other comment in CaarFunctorImpl.cpp...

rljacob:
@trey-ornl did you get a chance to try on perlmutter?

rljacob:
Notes:  need to check cpu performance.  Look for less-invasive way to do it.

trey-ornl:
> @trey-ornl did you get a chance to try on perlmutter?

Yes. It also sped up on `pm-gpu`, but it slowed down `pm-cpu` too much. I'm working on improvements for `pm-cpu`.

rljacob:
@trey-ornl any updates?

rljacob:
@trey-ornl any update on the CPU work?

rljacob:
I'm going to close this.  Please reopen when all concerns have been addresssed.

trey-ornl:
Replaced by https://github.com/E3SM-Project/E3SM/pull/6972.

PR #6523, opened 2024-07-23
MOAB api change in projection 

iulian787:
Add an extra argument, related to CAAS filtering, in iMOAB mapping.

[BFB]

---------

will need to update all MOAB installations when this is merged @vijaysm @rljacob


vijaysm (approved):
This is a straightforward change to conform to the new API. If the builds pass, this branch can be merged.

PR #6528, opened 2024-07-25
Updates LC machines

jasonb5:
- Replaces gdata with workspace
- Removes quartz machine
- Adds dane machine

[BFB]


cime_config/machines/config_machines.xml
```diff
@@ -3096,7 +3096,7 @@
     <RUNDIR>$CIME_OUTPUT_ROOT/$CASE/run</RUNDIR>
     <EXEROOT>$CIME_OUTPUT_ROOT/$CASE/bld</EXEROOT>
     <environment_variables compiler="intel">
-      <env name="NETCDF_PATH">/usr/gdata/e3sm/install/quartz/netcdf-fortran/</env>
+      <env name="NETCDF_PATH">/usr/workspace/e3sm/install/quartz/netcdf-fortran/</env>
```
AaronDonahue, line 3099: I think the next step will be to redo this install, so maybe we can leave it like this for now.  But having `quartz` in the name of the path I think would be confusing, since quartz no longer exists.

AaronDonahue (approved):
This looks great @jasonb5 .  Thank you, I have a few comments that shouldn't hold the PR up, but you are welcome to address if you want.

cime_config/machines/config_machines.xml
jasonb5: We'll leave this for the moment. There will be another update to use the new modules I'm building once I sort out the last couple issues.

PR #6535, opened 2024-08-01
[MPAS-Ocean standalone] Fix bottom depth when filling bathymetry holes

xylar:
In MPAS-Ocean init mode, when filling holes in the bathymetry (cells with deepest layers that are deeper than any of their neighbors) as part of the Haney-number-constrained vertical coordinate used for ice shelves, the bottom depth was not also being updated correctly for some cells.  This merge fixes the issue for cells in the z-level region of the ocean (far from ice-shelf cavities) by setting the bottom depth to the deepest depth within the same z level as the deepest neighbor.

### DIFF
--- a/components/mpas-ocean/src/mode_init/mpas_ocn_init_vertical_grids.F
+++ b/components/mpas-ocean/src/mode_init/mpas_ocn_init_vertical_grids.F
@@ -1480,6 +1480,7 @@ contains
           call mpas_pool_get_dimension(meshPool, 'nCellsSolve', nCellsSolve)
           call mpas_pool_get_array(meshPool, 'nEdgesOnCell', nEdgesOnCell)
           call mpas_pool_get_array(meshPool, 'cellsOnCell', cellsOnCell)
+          call mpas_pool_get_array(meshPool, 'bottomDepth', bottomDepth)
 
           ! Fill bathymetry holes, i.e. single cells that are deeper than all neighbors.
           ! These cells can collect tracer extrema and are not able to use
@@ -1497,6 +1498,9 @@ contains
 
              if (maxLevelCell(iCell) > maxLevelNeighbors) then
                 maxLevelCell(iCell) = maxLevelNeighbors
+                if (smoothingMask(iCell) == 0) then
+                   bottomDepth(iCell) = refBottomDepth(maxLevelNeighbors)
+                end if
              end if
           end do
 


xylar:
## Testing

I was able to run dynamic adjustment in Compass with the SORRM r3 mesh (https://github.com/MPAS-Dev/compass/pull/807) using this fix.  Without this fix, a single cell had a top layer thickness of ~140 m (compensating for a missing bottom layer), which led to other problems (e.g. a huge spike in temperature in that cell).

components/mpas-ocean/src/mode_init/mpas_ocn_init_vertical_grids.F
```diff
@@ -1497,6 +1498,9 @@ subroutine ocn_init_vertical_grid_with_max_rx1(domain, iErr)
 
              if (maxLevelCell(iCell) > maxLevelNeighbors) then
                 maxLevelCell(iCell) = maxLevelNeighbors
+                if (smoothingMask(iCell) == 0) then
+                   bottomDepth(iCell) = refBottomDepth(maxLevelNeighbors)
+                end if
              end if
           end do
```
xylar, line 1505: This fix makes this treatment within the haney-number coordinate (but outside the "smoothing mask" that defines the region where the haney-number coordinate is actually applied) consistent with the same code in the z-star coordinate used without ice shelf cavities, see:
https://github.com/E3SM-Project/E3SM/blob/1e2c8e15c427ea2e7535ddad35b5a97a1cb1dd55/components/mpas-ocean/src/mode_init/mpas_ocn_init_global_ocean.F#L831-L834

xylar:
@cbegeman, could you have a quick look at this?  I need this fix to make progress on the SORRM mesh.  I don't think you need to run any tests, I just think it would be useful to have someone take a quick look at the code and you came to mind.

cbegeman (approved):
I think this change looks good. Glad you were able to track it down!

It's interesting to me that it is necessary to move `bottomDepth` up; that it isn't sufficient just to make sure that `maxLevelCell` ensures continuity with its neighbors to avoid tracer issues. But we can chat about it when time isn't critical. I'm out the rest of the week.

xylar:
> It's interesting to me that it is necessary to move bottomDepth up; that it isn't sufficient just to make sure that maxLevelCell ensures continuity with its neighbors to avoid tracer issues. But we can chat about it when time isn't critical. I'm out the rest of the week.

I haven't looked in detail but the issue was that the sum of the layer thicknesses was not equal to `ssh + bottomDepth` at this particular cell.  Oddly, the result (at least after 2 days of integration) was that the top layer became 140 m thick to make up the difference.  This evolution presumably had other bad side effects.

Obviously, there should be a sanity check that the sum of layer thicknesses is equal to `ssh + bottomDepth` to avoid such an issue.  I don't particularly want to add that to init mode but it could be a good addition to Compass.

xylar:
@cbegeman, thanks for having a look, by the way!

xylar:
@jonbob, I believe this is ready to go in whenever you have a slot for it.  It should be trivial from E3SM's perspective since only MPAS-Ocean init mode is involved.

jonbob:
Passes:

merged to next

jonbob:
merged to master

xylar:
Thanks @jonbob!

PR #6525, opened 2024-07-24
Changing default NetCDF format to CDF5 and updating SCORPIO to v1.6.5

jayeshkrishna:
The CDF5/64bit_data format gets rid of limitations (for all practical
purposes) on NetCDF variable size imposed by the older
CDF2/64bit_offset file format.

Also updating SCORPIO to v1.6.5

[BFB]
[NML]

PR #6536, opened 2024-08-02
Upgrading to SCORPIO v1.6.5

jayeshkrishna:
SCORPIO v1.6.5 includes the following changes/fixes,

* Fix to support CDF5 file format when PnetCDF is not available

This upgrade should fix runtime issues when writing large variables,
introduced by PR #6525, on machines (like mappy) that do not have
support for PnetCDF 

[BFB]


dqwu (approved):

rljacob:
Since this fixes a problem with #6536 which is still open, you should add it to that feature branch.

rljacob (changes requested):
Close this an add it to #6525

jayeshkrishna:
SCORPIO will be updated to v1.6.5 in PR #6525

PR #6539, opened 2024-08-08
Update ANL LCRC Improv machine config

amametjanov:
Update ANL LCRC Improv machine config:
- force-purge sticky modules
- add env-var to avoid lock-files while reading netcdf files
- increase ne30 PE-layout 4->6 nodes to avoid OOM errors

[BFB]


amametjanov:
Testing with `e3sm_extra_coverage` -- 18 of 18 pass: https://my.cdash.org/viewTest.php?buildid=2627076

rljacob (approved):

PR #6547, opened 2024-08-14
MALI thermal solver bug fix and improved iceberg handling

matthewhoffman:
This merge updates the E3SM-Project/E3SM repo with MALI improvements from MALI-Dev/develop since February 2024.  The primary update needed is a bug fix to the thermal solver that prevents model crashes in some situations when ice thickness is very small.  Bringing this to master is in preparation for enabling the MALI thermal solver in E3SM, which is required for completing ice-sheet/ocean coupling.

Additionally, this brings in improvements to the detection of icebergs and small islands that makes the velocity solver more robust.  There are a number of other changes related to the subglacial hydrology model and sea-level model in MALI, both of which are still in development for coupled E3SM applications.

Complete list of MALI-Dev/develop merges included:

* 119f046a54 Merge branch 'trhille/mali/remove_icebergs_in_each_rk_stage' into MALI-Dev/develop
* e23c401962 Merge branch 'alexolinhager/restrictedDomainTimestepFix' into MALI-Dev/develop
* 3799e8320f Merge branch 'alexolinhager/smooth_hydro_fields' into MALI-Dev/develop
* 8ea2412928 Merge branch 'trhille/mali/improve_remove_small_islands' into MALI-Dev/develop
* 46bce269d2 Merge 'trhille/mali/improve_remove_small_islands' into develop
* b2143d1ac3 Merge branch 'alexolinhager/mali/totalGLdischargeCellEdge' into MALI-Dev/develop
* 102782fbd0 Merge branch 'alexolinhager/mali/autoWaterFluxMask' into MALI-Dev/develop
* 7067f38968 Merge  branch 'trhille/mali/always_remove_small_islands' into MALI-Dev/develop
* 1c903050cb Merge branch 'mperego/removeBuildType' into develop
* 86f24750e4 Merge branch 'matthewhoffman/mali/thermal_err_handling' into MALI-Dev/develop 
* 07223245a0 Merge branch 'hollyhan/hollyhan/mali/call_SLM_subroutine_to_allocate_nglv_variables' into MALI-Dev/develop
* 8d3c556a4b Merge branch 'andrewdnolan/mali/hydro_regional_stats' into MALI-Dev/develop
* f79661cb42 Merge branch 'hollyhan/mali/add_restart_functionality_slm' into MALI-Dev/develop
* 8be18efd4a Merge branch 'cshafer/hydro_stats' into MALI-Dev/develop
[...]

PR #6559, opened 2024-08-21
Activate MALI thermal solver and update configuration to be relevant to both AIS and GIS grids

matthewhoffman:
This PR updates a number of details about the MALI configuration to support both Antarctica and Greenland Ice Sheet configurations:
* Enables the MALI thermal solver, which is needed to allow evolving calving fronts, in preparation for enabling calving
* Adds new logic to make the albany_input.yaml file specific to a mesh.  This is needed because AIS and GIS meshes use different solver boundary conditions that are specified in that file.
* Enables config_remove_icebergs and config_remove_small_islands, two options that improve convergence of the velocity solver and are typically used in standalone MALI simulations
* Adds additional fields to the MALI history files related to the now activated thermal solver, as well as ice-shelf basal melting.  Also updates the MALI globalStats files to write out all global stats instead of the subset that was hardcoded before.
* Updates the ERS_Ld5.TL319_oQU240wLI_ais20.MPAS_LISIO_JRA1p5.mpaso-ocn_glcshelf test to the MALI ais8to30 grid that was recently added.  When the mesh was added, the test was failing due to issues with the thermal solver, but those have been resolved between https://github.com/E3SM-Project/E3SM/pull/6547 and this PR, so we can now update that test to the newer mesh.

[NML] for all configurations with MALI
[non-BFB] for all configurations with active MALI


matthewhoffman:
@stephenprice , here is the branch we were talking about that enables the thermal solver that you can test out with your various Greenland configurations.

@trhille , @stephenprice , @mperego , look over the changes to the namelist defaults and the new fields added to the history files and let me know if there is anything else you think we should include.

@mperego and @jonbob , let me know if you are happy with the way I've implemented supporting mesh-specific albany_input.yaml files.

@jonbob , I've only tested the test that is updated as part of this PR, so I will need help finding and running all the tests that use MALI.

components/mpas-albany-landice/cime_config/buildnml
trhille, line 1: Eventually we'll want `faceMeltingThickness` in here, but I don't know whether that's necessary yet. What do you think @matthewhoffman?

matthewhoffman: I add those in https://github.com/E3SM-Ocean-Discussion/E3SM/pull/94/files when facemelting gets enabled after introducing the TF coupling.  There will be a merge conflict when these PRs both come together, but I think it makes sense to keep that where it is.

components/mpas-albany-landice/cime_config/buildnml
```diff
@@ -300,7 +300,11 @@ def buildnml(case, caseroot, compname):
         if os.path.exists("{}/SourceMods/src.mali/{}".format(caseroot, albany_input_name)):
             safe_copy("{}/SourceMods/src.mali/{}".format(caseroot, albany_input_name), os.path.join(rundir, albany_input_name))
         else:
-            safe_copy("{}/components/mpas-albany-landice/bld/namelist_files/{}".format(srcroot, albany_input_name), os.path.join(rundir, albany_input_name))
+            # Each supported grid should have a corresponding albany_input.yaml file in
+            # components/mpas-albany-landice/bld/namelist_files/
+            # Naming convention is albany_input.GRID_PREFIX.yaml
+            albany_input_name_grid = f"albany_input.{grid_prefix}.yaml"
```
mperego, line 306: What about CPU vs GPU versions of the files? (solvers setting will likely be different)

mperego (approved):
Looks good to me, thanks!
I left a comment about having CPU and GPU versions of Albany input files, but it does not need to be addressed in this PR.

components/mpas-albany-landice/cime_config/buildnml
trhille: Okay, that makes sense to me. Thanks.

trhille (approved):

components/mpas-albany-landice/cime_config/buildnml
mperego: Sounds good.

stephenprice (approved):
Only a few minor suggestions / questions here Matt. I am happy to approve now regardless of what we decide to do in response to those.

components/mpas-albany-landice/cime_config/buildnml
```diff
@@ -229,25 +229,31 @@ def buildnml(case, caseroot, compname):
...
         lines.append('    <var name="upperSurface"/>')
         lines.append('    <var name="cellMask"/>')
         lines.append('    <var name="edgeMask"/>')
-#    lines.append('    <var name="vertexMask"/>')
-#    lines.append('    <var name="normalVelocity"/>')
         lines.append('    <var name="uReconstructX"/>')
         lines.append('    <var name="uReconstructY"/>')
         lines.append('    <var name="surfaceSpeed"/>')
         lines.append('    <var name="basalSpeed"/>')
-        lines.append('    <var name="sfcMassBal"/>')
```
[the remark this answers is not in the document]
matthewhoffman: Good suggestion - if MALI modifies the SMB field it was given in any way, we'd want to be able to quantify that, so having both `sfcMassBal` and `sfcMassBalApplied` would allow evaluating that.  I'll add that back in.

components/mpas-albany-landice/bld/namelist_files/albany_input.gis_1to10km_r02.yaml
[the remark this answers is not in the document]
matthewhoffman: If we currently primarily run with the integrated solver, there's no reason not to just have it set to that by default.  Part of the reason of having mesh-specific yaml files is to customize as needed for each mesh, and part of my goal with this PR is to reduce the amount of user_nl, etc., changes we are making.  

@mperego , @trhille , do you see any reason not to set the DI solver to true here (and below)?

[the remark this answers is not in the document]
trhille: No, I think it's a good idea to start using the DI solver by default.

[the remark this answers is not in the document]
mperego: Fine with me.

jonbob:
Passes:
* e3sm_landice_developer on chrysalis_gnu with expected DIFFs

Also successfully ran new test:
* ERS_Ld5.TL319_oQU240wLI_ais8to30.MPAS_LISIO_JRA1p5.chrysalis_intel.mpaso-ocn_glcshelf

merged to next

jonbob:
merged to master and expected DIFFs and NML DIFFs blessed except for mappy and pm-cpu_developer, which have not yet reported. But the new test is failing -- we probably will need to work on a new PR to change the test layouts as quickly as possible

jonbob:
The new test isn't failing -- a recent CIME update changed the behavior for BFAILs. A fix is being merged and I'll add bless requests for the new test as soon as it changes the dashboard

PR #6562, opened 2024-08-26
Add alternative data stream control for SO2 species in ext_frc

wlin7:
Made changes to allow SO2 species in ext_frc_specifier to have a separate
data stream control (e.g., type, cycle_yr) from that for the other species.
This is probably the simplest way to allow only volcanic SO2 to vary with time
while holding all other SO2 sectors to a fixed year in historical single forcing
experiment design.

[BFB] for existing tests.


wlin7:
Thanks for reviewing @mahf708 . I put the modified tracer_data.F90 on a wrong directory. 
No essential change to the file. Just to print an extra variable (datatimep) to track (and verify) the correct data slice is read. The file was copied there after my test runs, so I didn't realize.

rljacob:
@jinboxie please review.

jinboxie (approved):

wlin7:
Merged to master.

PR #6565, opened 2024-08-26
Update CIME submodule

jgfouca:
... to af3eab5b8a213e79923551e7f9d9f56edb944c0f

Changes:
1) Generalize _hists_match for mom6 files
2) support append in config_pes.xml overrides field
3) Add case git: Adds a case interface to git.
   It creates a local git repository when case.setup is
   run and updates that repository with each action that triggers an
   update of the CaseStatus file. Optionally a remote repository can be
   attached by setting CASE_GIT_REPOSITORY to the name of the remote
   repository.
4) xmlchange: Add documentation for setting values with commas
5) Remove mct for cesm
6) remove load_balancing_tool, broken and unsupported

Fixes
1) Fix/rest n in tests: move the computation of REST_N into the python
2) cprnc: need to make sure the directory exists before linking to it
3) check_lockedfiles call in xmlchange needs quiet=True
4) fixes detection of pfunit_path by making sure it exists
5) Fixes handling cprnc output
6) Fixes creating new environment with specific python version
7) Fixes `query_config` tool.
   - Adds longname to `--grids`
   - Fixes duplicate and empty choices
8) Fixes check_lockedfiles call in xmlchange
9) Fix list_e3sm_tests tool

[BFB]


jgfouca:
@mahf708 , that feature is off by default.

jgfouca:
Oh, I see the CI fails. I guess it breaks even if off?

jgfouca:
Yeah, I thought this was something the user had to opt-in for.

rljacob:
But ERS_P8.ne4pg2_oQU480.WCYCL2010NS passes and the only difference is the testmod to add output.

rljacob:
This is the error message:  “Expected to compare 9 hist files, but only compared 6. It’s possible
that the hist_file_extension entry in config_archive.xml is not correct
for some of your components.”

rljacob (changes requested):
Need to fix ERS test.  Should compare 9 files but only compares 6.

jgfouca:
Merged to next.

ndkeen:
Yes we hit that error that @mahf708 reported 
```
            if append[key]:
        KeyError: 'MAX_MPITASKS_PER_NODE'
```

jgfouca:
Re-merged to next.

rljacob:
Is the PR description still right?  Are we updating to hash af3eab5b8 ?

jgfouca:
@rljacob , fixed.

rljacob:
Is this PR description still accurate?

rljacob:
But you're not updating to a hash on master are you?  Its a hash (52787427846) on a branch off of master to revert the git feature and other things.  The PR description needs to say that.

jgfouca:
@rljacob , done.

rljacob (approved):

rljacob:
The description still isn't accurate because the git feature is disabled and the rest_n calculation PR was completely reverted so that had to be removed.  I fixed it.   Also that branch has to live forever in ESMCI/cime so that this state of master can be recovered.

PR #6566, opened 2024-08-26
Add ocean relativeVorticityAtSurface to highFrequencyStats

mark-petersen:
It is useful to visualize relative vorticity at the surface, which was not saved in high frequency stats in the past. I have added this commit to several recent simulations, including for the high-resolution Gulf of Mexico work. It works as expected. The variable is added to the default output for `highFrequencyStats`, but is only a 2D field and is optional. 

We commonly use relative vorticity at the surface for images and movies, in addition to surface kinetic energy.


xylar (approved):
@mark-petersen, I'm approving this based on a visual inspection of the code, which looks correct and is quite simple.  Let me know if you would like me to do further testing.

mark-petersen:
I labelled this as non-bfb because it changes the high frequency output variables. It is not answer changing.

jonbob:
Passes:

merged to next

jonbob:
merged to master

PR #6567, opened 2024-08-27
Add DLND to moab coupler

iulian787:
Add data land capability for moab driver;
case tested: --res r05_r05 --compset RMOSGPCC 
standard data moab model, mainly adapt from data ocean implementation

major change: Initialize all factors set by MOAB area correction init routine to 1.0. It affects all models, 
but it seems that some values were not initialized for data land case.
On chrysalis, with intel compiler, it actually complained about non initialized values when applying correction factors. 

The compressing file errors seen during development were caused by non-ascii characters in the log files. 
Those were caused by an uninitialized mapper, for which we were still creating a MOAB map. It was never used, 
it is fixed properly now, the moab map is not computed anymore. 

[BFB]

### DIFF
--- a/components/data_comps/dlnd/src/dlnd_comp_mod.F90
+++ b/components/data_comps/dlnd/src/dlnd_comp_mod.F90
@@ -148,10 +148,10 @@ CONTAINS
 #ifdef HAVE_MOAB
     character*400  tagname
     real(R8) latv, lonv
-    integer iv, tagindex, ilat, ilon  !, arrsize, nfields ! ierr is already defined as integer above
+    integer iv, tagindex, ilat, ilon 
     real(R8), allocatable, target :: data(:)
     integer(IN), pointer :: idata(:)   ! temporary
-    real(r8), dimension(:), allocatable :: moab_vert_coords  ! temporary
+    real(R8), dimension(:), allocatable :: moab_vert_coords  ! temporary
 #ifdef MOABDEBUG
     character*100 outfile, wopts
 #endif
--- a/components/data_comps/dlnd/src/dlnd_comp_mod.F90
+++ b/components/data_comps/dlnd/src/dlnd_comp_mod.F90
@@ -29,6 +29,10 @@ module dlnd_comp_mod
   use dlnd_shr_mod   , only: domain_fracname ! namelist input
   use dlnd_shr_mod   , only: nullstr
 
+#ifdef HAVE_MOAB
+  use seq_comm_mct,     only : mlnid ! id of moab lnd app
+  use iso_c_binding
+#endif
   ! !PUBLIC TYPES:
   implicit none
   private ! except
@@ -100,6 +104,12 @@ CONTAINS
        scmMode, scmlat, scmlon)
 
     ! !DESCRIPTION: initialize dlnd model
+#ifdef HAVE_MOAB
+       use iMOAB, only: iMOAB_DefineTagStorage, iMOAB_GetDoubleTagStorage, &
+                        iMOAB_SetIntTagStorage, iMOAB_SetDoubleTagStorage, &
+                        iMOAB_ResolveSharedEntities, iMOAB_CreateVertices, &
+                        iMOAB_GetMeshInfo, iMOAB_UpdateMeshInfo, iMOAB_WriteMesh
+#endif
     implicit none
 
     ! !INPUT/OUTPUT PARAMETERS:
@@ -134,6 +144,18 @@ CONTAINS
     integer(IN)        :: field_num ! field number
     character(nec_len) :: nec_str   ! elevation class, as character string
     character(*), parameter :: domain_fracname_unset = 'null'
+
+#ifdef HAVE_MOAB
+    character*400  tagname
+    real(R8) latv, lonv
+    integer iv, tagindex, ilat, ilon  !, arrsize, nfields ! ierr is already defined as integer above
+    real(R8), allocatable, target :: data(:)
+    integer(IN), pointer :: idata(:)   ! temporary
+    real(r8), dimension(:), allocatable :: moab_vert_coords  ! temporary
+#ifdef MOABDEBUG
+    character*100 outfile, wopts
+#endif
+#endif
 
     !--- formats ---
     character(*), parameter :: F00   = "('(dlnd_comp_init) ',8a)"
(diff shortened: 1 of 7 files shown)


components/data_comps/dlnd/src/dlnd_comp_mod.F90
```diff
@@ -135,6 +145,18 @@ subroutine dlnd_comp_init(Eclock, x2l, l2x, &
     character(nec_len) :: nec_str   ! elevation class, as character string
     character(*), parameter :: domain_fracname_unset = 'null'
 
+#ifdef HAVE_MOAB
+    character*400  tagname
+    real(R8) latv, lonv
+    integer iv, tagindex, ilat, ilon  !, arrsize, nfields ! ierr is already defined as integer above
```
vijaysm, line 151: Remove commmented out fields

components/data_comps/dlnd/src/dlnd_comp_mod.F90
```diff
@@ -135,6 +145,18 @@ subroutine dlnd_comp_init(Eclock, x2l, l2x, &
...
+    real(R8), allocatable, target :: data(:)
+    integer(IN), pointer :: idata(:)   ! temporary
+    real(r8), dimension(:), allocatable :: moab_vert_coords  ! temporary
```
vijaysm, line 154: Use R8 or r8 consistently everywhere. I suggest R8

components/data_comps/dlnd/src/dlnd_comp_mod.F90
```diff
@@ -256,6 +278,119 @@ subroutine dlnd_comp_init(Eclock, x2l, l2x, &
 
     call t_stopf('dlnd_initmctdom')
 
+#ifdef HAVE_MOAB
+    ilat = mct_aVect_indexRA(ggrid%data,'lat')
+    ilon = mct_aVect_indexRA(ggrid%data,'lon')
+    allocate(moab_vert_coords(lsize*3))
+    do iv = 1, lsize
+       lonv = ggrid%data%rAttr(ilon, iv) * SHR_CONST_PI/180.
+       latv = ggrid%data%rAttr(ilat, iv) * SHR_CONST_PI/180.
```
vijaysm, line 287: We should expose a utility method in MOAB to do coordinate transformations. It doesn't make sense having this exact same type of transform code in every component

components/data_comps/dlnd/src/dlnd_comp_mod.F90
```diff
@@ -256,6 +278,119 @@ subroutine dlnd_comp_init(Eclock, x2l, l2x, &
...
+                                     idata)
+    if (ierr .ne. 0)  &
+       call shr_sys_abort('Error: fail to set GLOBAL_ID tag ')
+ 
+    ierr = iMOAB_ResolveSharedEntities( mlnid, lsize, idata );
+    if (ierr .ne. 0)  &
+       call shr_sys_abort('Error: fail to resolve shared entities')
+ 
+    deallocate(moab_vert_coords)
+    deallocate(idata)
```
vijaysm, line 320: Is idata allocated inside MCT?

components/data_comps/dlnd/src/lnd_comp_mct.F90
```diff
@@ -146,13 +152,25 @@ subroutine lnd_init_mct( EClock, cdata, x2l, l2x, NLFilename )
...
+#endif
     call dlnd_comp_init(Eclock, x2l, l2x, &
          seq_flds_x2l_fields, seq_flds_l2x_fields, &
          SDLND, gsmap, ggrid, mpicom, compid, my_task, master_task, &
          inst_suffix, inst_name, logunit, read_restart, &
          scmMode, scmlat, scmlon)
-
+#ifdef HAVE_MOAB
+     if (my_task == master_task) then
+          call seq_infodata_PutData( infodata, lnd_domain=SDLND%domainFile) ! we use the same one for regular case
```
vijaysm, line 169: Why is this needed only for MOAB driver?

driver-moab/main/component_mod.F90
```diff
@@ -748,6 +748,7 @@ subroutine component_init_areacor_moab (comp, mbccid, mbcxid, seq_flds_c2x_fluxe
          lsize = comp(1)%mblsize
          allocate(areas (lsize, 3)) ! lsize is along grid; read mask too
          allocate(factors (lsize, 2))
+         factors = 1.0 ! initialize with 1.0 all factors; then maybe correct them
```
vijaysm, line 751: Shouldn't it be factors(:,:) = 1.0? Or is the syntax in general the same (and independent of dimension or compiler family)?

vijaysm (commented):
Reviewed the code and made comments. I haven't run this branch and tested it just yet.

iulian787:
I tested this branch on different platforms, and chrysalis is working fine; gce, perlmutter, and my laptop (gnu compiler) all fail to compress the log files at the end. So the bug I saw earlier is not solved yet. Somehow, it worked on gce when I tested 2 weeks ago;  needs more work

components/data_comps/dlnd/src/dlnd_comp_mod.F90
iulian787: We have this in MOAB, but it does not make sense to expose it from there; iMOAB?  The utility should be either in 
E3SM/share or in E3SM/driver-moab/shr

components/data_comps/dlnd/src/lnd_comp_mct.F90
iulian787: the regular coupler does not need the mesh; we do

components/data_comps/dlnd/src/dlnd_comp_mod.F90
iulian787: mct_gsMap_orderedPoints  allocates it

components/data_comps/dlnd/src/dlnd_comp_mod.F90
rljacob: If its only used within the coupler code (including the code in the models) then driver-moab/shr would be the place.

rljacob: But you don't have to do that for this PR.

driver-moab/main/component_mod.F90
iulian787: we are not consistent, we use both forms, but this one is shorter. 
We are consistent when we use set/get methods with these 2 dimensional arrays, iMOAB_SetDoubleTagStorage, iMOAB_GetDoubleTagStorage; 
`ierr = iMOAB_SetDoubleTagStorage( mbccid, tagname, arrsize , comp(1)%mbGridType, factors)`

driver-moab/main/prep_lnd_mod.F90
```diff
@@ -693,7 +693,7 @@ subroutine prep_lnd_mrg_moab (infodata)
 #endif
 #ifdef MOABCOMP
     character(CXX)           :: tagname, mct_field
-    real(r8)                 :: difference
+    real(R8)                 :: difference
```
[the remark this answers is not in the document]
rljacob: Because you said to use R8 or r8 consistently.  R8 is better.

components/data_comps/dlnd/src/dlnd_comp_mod.F90
```diff
@@ -256,6 +278,119 @@ subroutine dlnd_comp_init(Eclock, x2l, l2x, &
...
+    enddo
+ 
+    ! create the vertices with coordinates from MCT domain
+    ierr = iMOAB_CreateVertices(mlnid, lsize*3, 3, moab_vert_coords)
+    if (ierr .ne. 0)  &
+       call shr_sys_abort('Error: fail to create MOAB vertices in data lnd model')
+ 
+    tagname='GLOBAL_ID'//C_NULL_CHAR
+    ierr = iMOAB_DefineTagStorage(mlnid, tagname, &
+                                  0, & ! dense, integer
```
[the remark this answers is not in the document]
iulian787: yes, we do, because if we do not define it, iMOAB set tag does not work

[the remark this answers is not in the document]
vijaysm: Hmm, it should for a fundamental tag like this.

vijaysm (approved):
Thanks for addressing my comments!

PR #6569, opened 2024-08-28
SYCL in homme

oksanaguba:
Changes needed for spot/aurora: sycl kokkos spaces, temporary TP changes (TP init will be sorted out later), switch booleans to ints in F<-->xx interfaces.

[bfb]


oksanaguba:
Older states of this branch contain more caar/remap/sph tests that do not require mpi. See commit 206d0caad8df02fe903c00e4f8af454ed658f538 (before `prepare to merge to master part1` in which i started cleaning some of that). There were 3 small tests, remap (file r1.cpp), sphere op, probably, divergence (file s1_ml), and caar Pre Exchange (c1_ut). The tests were created to investigate performance from machine to machine for small parts of code and without mpi.  They did not use F counterpart (cause the needed perf flags), but used compute_error script for bfb or error evaluations. Also see TESTER_NOMPI pragma, that was needed to compile without mpi. Also CaarFunctorImpl had infrastructure to brake it into smaller parts.

i could see that having these test around (at least, caar) is probably useful for future, but organizing them properly should be a separate PR, for which i do not have time.

oksanaguba:
@bartgol or @ambrad  in this branch i revived possibility to use external kokkos build instead of pointing ekat to a new kokkos version. i still prefer to have such option, but do you object to it?

ambrad:
@oksanaguba is that for standalone Homme? I suspect that's an easier case for supporting an external Kokkos build than the CIME case. In any case, it's fine with me; I believe the option was lost by accident.

oksanaguba:
Ready for review, please, @bartgol  and @ambrad  (i will run tests meanwhile).

components/homme/cmake/machineFiles/aurora-jit.cmake
```diff
@@ -0,0 +1,58 @@
+#module restore
```
bartgol, line 1: And "-jit"? I doubt it is "Just In Time"?

components/homme/src/share/cxx/GllFvRemap.cpp
```diff
@@ -16,7 +16,7 @@
 namespace Homme {
 
 void init_gllfvremap_c (int nelemd, int np, int nf, int nf_max,
-                        bool theta_hydrostatic_mode,
```
bartgol, line 19: I'm assuming all these changes are b/c there's a problem getting bool values in a device lambda on SYCL, or something like that?

If that's the case, I would keep bool everywhere, and simply convert to int in the C++ structures that do need an int on SYCL. That would allow to leave lots of other files (including the F90 files) unchanged.

components/homme/src/share/cxx/GllFvRemapImpl.cpp
```diff
@@ -395,7 +395,7 @@ ::run_dyn_to_fv_phys (const int timeidx, const Phys1T& ps, const Phys1T& phis, c
   const auto hvcoord = m_hvcoord;
   
   const bool use_moisture = m_data.use_moisture;
-  const bool theta_hydrostatic_mode = m_data.theta_hydrostatic_mode;
+  const int theta_hydrostatic_mode = m_data.theta_hydrostatic_mode;
```
bartgol, line 398: I believe this is the only place where you need this to be an int, right? I see it used inside the KOKKOS_LAMBDA below. Then, you can keep `theta_hydrostatic_mode` as a bool _everywhere_, and just use an int for the local var here.

components/homme/src/share/gllfvremap_mod.F90
```diff
@@ -265,22 +265,22 @@ end subroutine gfr_init
 
   subroutine gfr_init_hxx() bind(c)
 #if KOKKOS_TARGET
-    use control_mod, only: theta_hydrostatic_mode
```
bartgol, line 268: Keeping `theta_hydrostatic_mode` an int throughout the code (except for local vars for kernels) will make all these changes not needed.

components/homme/src/theta-l_kokkos/cxx/EquationOfState.hpp
```diff
@@ -250,7 +250,7 @@ class EquationOfState {
 
 public:
 
-  bool            m_theta_hydrostatic_mode;
+  int            m_theta_hydrostatic_mode;
```
bartgol, line 253: This local var can be made an int if bool causes problem on SYCL, but the `init` method should just accept a bool.

components/homme/src/theta-l_kokkos/cxx/HyperviscosityFunctorImpl.cpp
```diff
@@ -118,9 +118,14 @@ void HyperviscosityFunctorImpl::init_params(const SimulationParams& params)
   m_eos.init(params.theta_hydrostatic_mode,m_hvcoord);
 
 #ifdef HOMMEXX_BFB_TESTING
-  m_process_nh_vars = true;
+  m_process_nh_vars = 1;
 #else
-  m_process_nh_vars = !params.theta_hydrostatic_mode;
+  //m_process_nh_vars = !params.theta_hydrostatic_mode;
+  if (params.theta_hydrostatic_mode){
```
bartgol, line 124: You can just do
```c++
m_process_nh_vars = static_cast<int>(params.theta_hydrostatic_mode);
```
Faster to read, and makes your intent of simply changing the type clearer.

components/homme/src/theta-l_kokkos/cxx/cxx_f90_interface_theta.cpp
```diff
@@ -43,12 +43,13 @@ void init_simulation_params_c (const int& remap_alg, const int& limiter_option,
                                const Real& nu, const Real& nu_p, const Real& nu_q, const Real& nu_s, const Real& nu_div, const Real& nu_top,
                                const int& hypervis_order, const int& hypervis_subcycle, const int& hypervis_subcycle_tom,
                                const double& hypervis_scaling, const double& dcmip16_mu,
-                               const int& ftype, const int& theta_adv_form, const bool& prescribed_wind, const bool& moisture, const bool& disable_diagnostics,
-                               const bool& use_cpstar, const int& transport_alg, const bool& theta_hydrostatic_mode, const char** test_case,
+                               const int& ftype, const int& theta_adv_form, const int& prescribed_wind, const int& use_moisture, const int& disable_diagnostics,
```
bartgol, line 46: Why changing the type of these boolean from bool to int?

components/homme/src/theta-l_kokkos/prim_driver_mod.F90
```diff
@@ -93,29 +93,39 @@ subroutine prim_create_c_data_structures (tl, hvcoord, mp)
...
+    integer :: disable_diagnostics_int, theta_hydrostatic_mode_int, use_moisture_int
+
     ! Initialize the C++ reference element structure (i.e., pseudo-spectral deriv matrix and ref element mass matrix)
     dvv = deriv1%dvv
     elem_mp = mp
     call init_reference_element_c(c_loc(dvv),c_loc(elem_mp))
 
     ! Fill the simulation params structures in C++
     test_name = TRIM(test_case) // C_NULL_CHAR
+
```
bartgol, line 105: This seems overly complicated. We create an int from a bool to call the c function, and then inside the c function we convert the int back to a bool. Why not passing bool in the first place?

components/homme/src/theta-l_kokkos/theta_f2c_mod.F90
```diff
@@ -138,11 +138,11 @@ end subroutine init_reference_element_c
 
   ! Create C++ functors
   subroutine init_functors_c (allocate_buffer) bind(c)
-  use iso_c_binding, only: c_bool
```
bartgol, line 141: Also here, why the bool->int change?

components/homme/cmake/machineFiles/chrysalis-bfb.cmake
```diff
@@ -17,6 +17,8 @@ ENDIF()
 SET (USE_MPIEXEC "srun" CACHE STRING "")
 SET (USE_MPI_OPTIONS "-K --cpu_bind=cores" CACHE STRING "")
 
+SET (CHRYSALIS_MACHINE TRUE CACHE BOOL "")
```
bartgol, line 20: I don't see this var ever being used. Can we rm from these two mach files?

bartgol (commented):
I am confused by two things:
1. the linker language for SUNSPOT
2. all the bool conversions to int

I guess there may be some issues with bool on SYCL device (not sure if that's the case, but I can see something like that happening), so it's ok to create an int local var from a bool. But the rest of the infrastructure should use bool/logical for stuff that is conceptually a boolean.

components/homme/src/prim_main.F90
```diff
@@ -20,7 +20,7 @@ program prim_main
   use element_mod,      only: element_t
   use common_io_mod,    only: output_dir, infilenames
   use common_movie_mod, only: nextoutputstep
-  use perf_mod,         only: t_initf, t_prf, t_finalizef, t_startf, t_stopf ! _EXTERNAL
+  use perf_mod,         only: t_initf, t_prf, t_finalizef, t_startf, t_stopf, t_disablef, t_enablef ! _EXTERNAL
```
bartgol, line 23: `t_enablef` is not used (will generate yet another compiler warning). Let's prune it.

ambrad:
Two points for consideration:
1. I agree with Luca that it's not clear the bool->int conversion to work around the system issue is, in its current form, as minimal as it can be. It might be worth thinking about whether the workaround can be implemented with substantially fewer modifications.
2. This PR has a lot of micro commits, 103. Ideally, these would be squashed to a few meaningful ones.

components/homme/src/theta-l_kokkos/cxx/ForcingFunctor.hpp
```diff
@@ -236,7 +236,7 @@ class ForcingFunctor
     });
   }
 
-  void tracers_forcing (const Real dt, const int np1, const int np1_qdp, const bool adjustment, const MoistDry moisture) {
+  void tracers_forcing (const Real dt, const int np1, const int np1_qdp, const bool adjustment, const bool use_moisture) {
```
ambrad, line 239: I'm fine with this change, but was it done for a system reason? Or is it just something you want to do?

components/homme/src/prim_main.F90
```diff
@@ -240,6 +240,11 @@ end subroutine finalize_kokkos_f90
      
      nstep = nextoutputstep(tl)
      do while(tl%nstep<nstep)
+
+     if(tl%nstep < 2) then 
+        call t_disablef()
```
ambrad, line 245: Does this change have implications for E3SM more widely? My understanding of `t_disablef` is it would affect timer output in all E3SM runs.

ambrad: Never mind: I see now this is in `program main` for standalone Homme. It's still worth double-checking if you really want this chang.e

components/homme/src/theta-l_kokkos/prim_driver_mod.F90
```diff
@@ -343,21 +353,21 @@ subroutine prim_init_elements_views (elem)
...
+    use iso_c_binding, only : c_int
     use theta_f2c_mod, only : init_functors_c, init_boundary_exchanges_c
-
     !
     ! Optional Input
     !
-     logical(kind=c_bool), optional :: allocate_buffer  ! Whether functor memory buffer should be allocated internally
-
+    integer, intent(in), optional :: allocate_buffer  ! Whether functor memory buffer should be allocated internally
+    integer(kind=c_int) :: dummy
```
ambrad, line 362: I think this line can be removed.

components/homme/cmake/machineFiles/aurora-aot.cmake
```diff
@@ -0,0 +1,64 @@
+#module restore
```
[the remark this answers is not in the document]
oksanaguba: ahead of time https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-0/ahead-of-time-compilation.html

components/homme/cmake/machineFiles/aurora-jit.cmake
oksanaguba: it is "just in time" https://www.intel.com/content/www/us/en/docs/oneapi/optimization-guide-gpu/2023-0/just-in-time-compilation-in-sycl.html

components/homme/src/prim_main.F90
oksanaguba: yes, i do. the issue is that even with aot build, timers are polluted because some TPL libs are built with jit. this change resolved the issue.

components/homme/src/theta-l_kokkos/cxx/ForcingFunctor.hpp
oksanaguba: moisture variable was used in some complicated way, through some type MoistDry::MOIST, MoistDry::DRY, i just made it boolean (and then int) instead.

oksanaguba:
Thanks, Luca and Andrew. Luca used lazygit and made a new branch with a few commits out of this one. If you do not like int-bool changes, we can just wait till the issue resolves itself, and then redo this work. If possible, I would not want to spend more time cleaning that change (but i will check HV functor if i had an unnecessary change there).

oksanaguba:
see #6594 .

Issue #6573, opened 2024-08-30
MPAS-SEAICE outputting unnamed file.


rljacob:
The MPAS-SEAICE model is writing a "fort.101" file in the RUNDIR anytime it is used (full model or F-case).

First few lines are
```

 (icepack_write_parameters)
   rhos       =    330.000000000000
   rhoi       =    917.000000000000
   rhow       =    1026.00000000000
   cp_air     =    1004.64000000000
```

Text output files in the RUNDIR should have unique names including the model that is writing them and, ideally, the case name as well.   Something is writing to unit 101 without giving the file a name so Fortran calls it "fort.101".

Noticed in the integration testing by @ndkeen

proteanplanet:
@eclare108213 This is happening due to the following lines left in by Tony during Icepack implementations:

https://github.com/E3SM-Project/E3SM/blob/f50d2639c68c1607a521407b420af040080c1380/components/mpas-seaice/src/shared/mpas_seaice_icepack.F#L10662-L10665

eclare108213:
Thanks for tracking it down so quickly, @proteanplanet.  I'll fix it.

PR #6574, opened 2024-08-30
clean up debugging code and comments

eclare108213:
Remove extraneous comments and a write statement used for debugging.
Verified that the code compiles and no longer produces the fort.101 file.
Fixes https://github.com/E3SM-Project/E3SM/issues/6573
[BFB]


eclare108213:
I did not use the correct form for the branch name.  Do I need to redo this and make a new PR?

proteanplanet:
Yes, you need to create a branch named eclare/seaice/remove-debug-output (or something similar for the part after the last "/".

PR #6580, opened 2024-09-02
lessen gh/ci testing burden + add eamxx prod tests

mahf708:
Adds eamxx tests to gh/ci suite.  Also remove some tests and lower simulations lengths to make space for
more specialized testing coming in the pipelines.

[BFB]


components/eamxx/cime_config/testdefs/testmods_dirs/scream/v1prod/yaml_outs/scream_output.decadal.1dailyAVG_native.yaml
```diff
@@ -0,0 +1,16 @@
...
+Fields:
+  Physics PG2:
+    Field Names:
+    - snow_depth_land
+output_control:
+  Frequency: 1
+  frequency_units: ndays
+  MPI Ranks in Filename: false
+Restart:
+  force_new_file: true
```
bartgol, line 16: Are we still using this option in prod runs? I remember we were having issues without, but I don't know if things changed ever since...

bartgol (approved):

jgfouca (approved):

rljacob (changes requested):
Please change testmod directory name to eamxx/proc instead of scream/v1prod.
testdefs/testmods_dirs/eamxx/prod/

components/eamxx/cime_config/testdefs/testmods_dirs/scream/v1prod/yaml_outs/scream_output.decadal.1dailyAVG_native.yaml
```diff
@@ -0,0 +1,16 @@
+%YAML 1.1
```
rljacob, line 1: Do all these yaml_outs get consumed by the v1prod testmod ?

bartgol:
@rljacob I think testmods need the first name to match the component name, right? Despite the folder being called "eamxx", the atm comp name is still "scream", so I think we need to keep "scream/xyz", no? I agree that the comp name should be changed to "eamxx", but maybe that can be done separately?

rljacob:
I think subdir names under testmod_dirs can be anything.  If you have ERS_P4.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI..foo-bar it will look for eamxx/cime_config/testdefs/testmods_dirs/foo/bar  .  eamxx is the "primary component" so it looks there instead of elm/cime_config/testdefs/testmods_dirs

bartgol:
> I think subdir names under testmod_dirs can be anything. If you have ERS_P4.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI..foo-bar it will look for eamxx/cime_config/testdefs/testmods_dirs/foo/bar . eamxx is the "primary component" so it looks there instead of elm/cime_config/testdefs/testmods_dirs

Interesting. I saw $compname as the first testmod folder in all components I stuck my nose into, so I assumed it was a necessary thing.

So are you saying that if I put an `elm/betr` testmod in `components/eamxx/cime_config/testdefs/testmod_dirs`, CIME would pick that up instead of the homonymous one in the elm component?

rljacob:
Yes but only for an F-case.

bartgol:
At some point we will have to switch from scream to eamxx, I think that has been already decided.

I wonder if it is enough to add another alias in config_component.xml, so that both scream and eamxx work. But I suppose that wouldn't be much different from what you are doing.

rljacob (approved):

rljacob:
Thanks for making that change.   When the current "scream" tests are changed to "eamxx", that's a new test name and any baselines will have to be recreated.  With your mod to config_file, new tests can have the right name and avoid that future mess.

rljacob:
The name change should be its own PR since it will be disruptive.

jgfouca:
@mahf708 , @rljacob is the integrator for this PR.

Issue #6521, opened 2024-07-22
Hang for test using nvidia compiler only for certain smaller MPI counts `ERS.hcru_hcru.IELM.pm-cpu_nvidia.elm-multi_inst`


ndkeen:
This looks like a new test -- it is failing on pm-cpu with nvidia compiler. Based on the dates of log files, it looks like
the test is hanging.

Note the current MPI count used by default for this test is 192, which is 1.5 nodes on pm-cpu

ndkeen:
This test has still be failing/hanging. Adding a little more details, these tests seem OK, so it looks like it's combination of the multi_inst modifier with newer nvidia compilers.

These tetst pass:
```
ERS.hcru_hcru.IELM.pm-cpu_nvidia
```

ndkeen:
OK, there might be an issue with how it's launching more tasks/jobs as if I force the test to land on one node only, it passes.  That is: `ERS_P128x1.hcru_hcru.IELM.pm-cpu_nvidia.elm-multi_inst`

ndkeen:
I created PR https://github.com/E3SM-Project/E3SM/pull/6581 to use 2 full nodes (256 MPI's) instead of the current odd value of 192 MPI's.  There must have been a reason why I used 192 here, but as I'm seeing tests pass with 2 full nodes, seems like an obvious choice.

I want to keep this issue open as it's still odd that certain MPI counts cause a hang, while others don't.
Seemingly, only for nvidia compiler.

ndkeen:
Merged https://github.com/E3SM-Project/E3SM/pull/6581 so we should not see the issue on cdash.

ndkeen:
It may be that https://github.com/E3SM-Project/E3SM/pull/6687 will also address this issue.

Using `kdreg2`, I was able to run the following:
```
ERS_D_P128.hcru_hcru.IELM.muller-cpu_nvidia.elm-multi_inst
ERS_D_P192.hcru_hcru.IELM.muller-cpu_nvidia.elm-multi_inst
ERS_D_P2048.hcru_hcru.IELM.muller-cpu_nvidia.elm-multi_inst
ERS_D_P256.hcru_hcru.IELM.muller-cpu_nvidia.elm-multi_inst
ERS_P128.hcru_hcru.IELM.muller-cpu_nvidia.elm-multi_inst
ERS_P192.hcru_hcru.IELM.muller-cpu_nvidia.elm-multi_inst
[...]

Issue #6486, opened 2024-06-25
PIO_INTERNAL_ERROR with `ERS_P256.hcru_hcru.I20TRGSWCNPRDCTCBC.pm-cpu_intel.elm-erosion`


ndkeen:
With `ERS.hcru_hcru.I20TRGSWCNPRDCTCBC.pm-cpu_intel.elm-erosion`, the test uses 128 tasks on 1 node and was a little slow. I wanted to try using 2 nodes (256 tasks), but hit an error described here.

Note that to still improve speed of these tests, I went ahead with a PR to increase tasks to 192 (still using 2 nodes):
After https://github.com/E3SM-Project/E3SM/pull/6484, we are now using 192 tasks for all components.

To reproduce the errro:

`ERS_P256.hcru_hcru.I20TRGSWCNPRDCTCBC.pm-cpu_intel.elm-erosion`

Note I don't see the error with a SMS test -- so seems to be related to writing restarts.

I was seeing:

```
128: PIO: FATAL ERROR: Aborting... An error occured, Waiting on pending requests on file (./ERS.hcru_hcru.I20TRGSWCNPRDCTCBC.pm-cpu_intel.elm-erosion.C.20240621_095913_j1wl0t.elm.r.1850-01-07-00000.nc, ncid=54) failed (Number of pending requests on file = 2\
3, Number of variables with pending requests = 23, Number of request blocks = 3, Current block being waited on = 1, Number of requests in current block = 11).. Size of I/O request exceeds INT_MAX (err=-237). Aborting since the error handler was set to PIO_INTERNAL_ERROR... (/global/cfs/cdirs/e3sm/ndk/repos/pr/ndk_mf_pm-cpu-pelayout-minor-adjustment/externals/scorpio/src/clib/pio_darray_int.c: 2189)
128: Obtained 10 stack frames.
128: /pscratch/sd/n/ndk/e3sm_scratch/pm-cpu/ERS.hcru_hcru.I20TRGSWCNPRDCTCBC.pm-cpu_intel.elm-erosion.C.20240621_095913_j1wl0t/bld/e3sm.exe() [0x17e301c]
128: /pscratch/sd/n/ndk/e3sm_scratch/pm-cpu/ERS.hcru_hcru.I20TRGSWCNPRDCTCBC.pm-cpu_intel.elm-erosion.C.20240621_095913_j1wl0t/bld/e3sm.exe() [0x17e325e]
[...]

PR #6581, opened 2024-09-03
For pm-cpu, increase default pelayout to 3 nodes for tests using `l%360x720cru`

ndkeen:
Currently, the tests for this resolution use 192 MPI's on pm-cpu which is an odd value (1.5 nodes).
Here it's being changed to use `-3` (or 384 MPI's).

Example of test that would use this layout: `SMS.hcru_hcru.IELM`

This change is effective work-around (but not fix) for #6521 with #6486 in mind as noted below.

[bfb]

### DIFF
--- a/components/elm/cime_config/config_pes.xml
+++ b/components/elm/cime_config/config_pes.xml
@@ -239,14 +239,14 @@
       <pes compset="any" pesize="any">
         <comment>elm: pm-cpu 2 nodes for grid l%360x720cru</comment>
         <ntasks>
-          <ntasks_atm>192</ntasks_atm>
-          <ntasks_lnd>192</ntasks_lnd>
-          <ntasks_rof>192</ntasks_rof>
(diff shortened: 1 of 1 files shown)


ndkeen:
This change does not resolve https://github.com/E3SM-Project/E3SM/issues/6521, but a clean work-around.
There still seem to be certain MPI counts that trip up a certain test with nvidia compiler:
`ERS.hcru_hcru.IELM.pm-cpu_nvidia.elm-multi_inst`

rljacob (approved):

ndkeen:
merged to next

ndkeen:
I merged to master after seeing `[ERS.hcru_hcru.IELM.pm-cpu_nvidia.elm-multi_inst](https://my.cdash.org/test/198193537)` pass and not seeing any other issues. There were several (5 total in 2 suites) expected NML diffs with tests using these new pelayout and I've created bless requests.

Issue #6556, opened 2024-08-19
Non-reproducible results with `SMS_R_Ld5.ne4_ne4.FSCM-ARM97.pm-cpu_gnu` after newer GNU compilers on `pm-cpu`


ndkeen:
A few weeks ago, we updated the GNU compiler version on pm-cpu https://github.com/E3SM-Project/E3SM/pull/6517. We expected results to be different and we blessed them.  However, one of the tests continues to DIFF.  Trying the test manually, while the atm.log values are identical, when I use `cprnc` on the `eam.h0` netcdf files, they are different between two runs.

```
SMS_R_Ld5.ne4_ne4.FSCM-ARM97.pm-cpu_gnu.eam-scm
SMS_R_Ld5.ne4_ne4.FSCM-ARM97.pm-cpu_gnu
```

It only happens with GNU (Intel still OK). 
And with GNU and DEBUG, I can run 2 cases and see identical netcd files.
I can work toward trying to find if there is a certain file that we can downgrade optimization to still allow BFB between runs.
Note that simple changing `SMS` to `ERS` will not work for this case (for some reason).

I just verified that without the `eam-scm` modifier, I still see a diff in the `eam.h0` file

Example of cprnc command I'm trying: 
```
perlmutter-login11% pwd
/pscratch/sd/n/ndk/e3sm_scratch/pm-cpu/nexty-aug9
perlmutter-login11% cprnc SMS_R_Ld5.ne4_ne4.FSCM-ARM97.pm-cpu_gnu.r00/run/SMS_R_Ld5.ne4_ne4.FSCM-ARM97.pm-cpu_gnu.r00.eam.h0.1997-06-19-84585.nc SMS_R_Ld5.ne4_ne4.FSCM-ARM97.pm-cpu_gnu.r01/run/SMS_R_Ld5.ne4_ne4.FSCM-ARM97.pm-cpu_gnu.r01.eam.h0.1997-06-19-84585.nc | tail -40

                   1   0.129433436191206E-09   0.129433436191206E-09
                   1  (     1,     1) (     1,     1)
          avg abs field values:    1.294334361912064E-10
                                   1.294334361912064E-10
 vap_ice_exchange   (ncol,lev,time)  t_index =     61    61
                  80  (     0,    54,     1) (     0,    58,     1)
                  80   0.625710816493097E-08  -0.118322907027846E-07
                  80   0.625710816493097E-08  -0.118322907027846E-07
[...]

PR #6585, opened 2024-09-04
For SCM cases, correct how we store the computed area to write in eam history file

ndkeen:
For SCM cases, we compute area and register it, but as it's a local copy, the value is not available at writing time to netcdf.
This results in the code writing a value for area that is uninitialized and different compilers can have different behavior, which could result in a DIFF between runs (and the incorrect area written to the file).

BFB for all cases, except SCM. And the results are still BFB for SCM, just affects the area written to file.

Fixes https://github.com/E3SM-Project/E3SM/issues/6556

### DIFF
--- a/components/eam/src/dynamics/se/dyn_grid.F90
+++ b/components/eam/src/dynamics/se/dyn_grid.F90
@@ -101,6 +101,7 @@ module dyn_grid
   real(r8),      allocatable :: pelon_deg(:)        ! pe-local longitudes (degrees)
   real(r8),          pointer :: pearea(:) => null() ! pe-local areas
   integer(iMap),     pointer :: pemap(:)  => null() ! pe-local map for PIO decomp
+  real(r8),          pointer :: pearea_scm(:) => null() ! special case: area for SCM (only 1 value)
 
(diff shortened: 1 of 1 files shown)


rljacob (approved):

ndkeen:
Note the first commit here is cleaning up white-space.
The next commits are the actual change.

I also think there is incorrect logic for when SCM is enabled. I changed
```
if (.not.single_column .or. scm_multcols) then

to

if (single_column .or. scm_multcols) then
```

ndkeen:
I think I incorrectly did the logic for scm -- will fix.

bogensch (approved):
Thanks for working on this @ndkeen

ndkeen:
merged to next

ndkeen:
merged to master, submitted bless request

rljacob:
@ndkeen did you bless on mappy?  Its still diffing.

ndkeen:
No I did not.  I was actually not aware that this was an issue on any other machine. I can't even login to mappy. Glad it fixed more than one.

rljacob:
Yes please do.

PR #6589, opened 2024-09-05
Adds a lower limit to stem allocation to avoid negative states

### LINKED ISSUE #6591: Model stops with land temperature > 400K due to negative carbon stock
Very high temperature calculated by land model is causing a model stop when critical threshold is exceeded. The high temperatures are coming from an excessively high canopy height calculated by the model. The excessive height is coming from a negative value for stem carbon. The negative stem carbon is coming from a negative dynamic carbon allocation parameter. The negative allocation parameter is coming from a very low productivity for multiple years, and the lack of a lower bounds constraint on that parameter under conditions of low productivity.  
The confluence of these conditions is very rare, but the problem was showing up in isolated gridcells after multiple years of coupled simulation in the V3LR campaign.

thorntonpe:
The stem:leaf allocation variable can go negative without this limiter, which
can cause negative stem carbon pools.

Fixes #6591 
[non-BFB]

### DIFF
--- a/components/elm/src/biogeochem/AllocationMod.F90
+++ b/components/elm/src/biogeochem/AllocationMod.F90
@@ -645,7 +645,7 @@ contains
 
          if (stem_leaf(ivt(p)) < 0._r8) then
              if (stem_leaf(ivt(p)) == -1._r8) then
-                 f3 = (2.7/(1.0+exp(-0.004*(annsum_npp(p) - 300.0)))) - 0.4
+                 f3 = max((2.7/(1.0+exp(-0.004*(annsum_npp(p) - 300.0)))) - 0.4_r8, 0.2_r8)
              else
                  f3 = max((-1.0_r8*stem_leaf(ivt(p))*2.7_r8)/(1.0_r8+exp(-0.004_r8*(annsum_npp(p) - &
                            300.0_r8))) - 0.4_r8, 0.2_r8)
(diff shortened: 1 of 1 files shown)


rljacob:
Can you say more about why this is non-BFB ?

thorntonpe:
A condition can occur where the climate in a given gridcell is not supporting the growth of a woody vegetation type, and the carbon stock for that type as well as its productivity is declining over time. In some very rare instances this can lead to a negative value for a calculated parameter controlling how much new photosynthesis is put toward production of new wood. In even more rare circumstances this can lead to a calculation of unrealistically tall vegetation height, which can cause unrealistically high surface temperatures, which can trigger a model stop if the temperature exceeds a critical threshold. With this bug fix in place the calculated vegetation parameter for allocation of new wood is not allowed to go negative, and so there will be non-BFB results in a small number of gridcells for a small number of timesteps. The bug-fix will also prevent the occurrence of the even more rare canopy height and temperature threshold problems, and so runs that previously stopped will now continue.

rljacob:
Can you open an issue for the bug that this fixes?

thorntonpe:
Issue created (#6591 ) and PR description updated.

bpbond (approved):

thorntonpe:
I see that this is connected to the v3.1beta milestone. Can it also be added to the v3.0.1 milestone, for integration with the V3 HES simulations?

rljacob:
v3 HES is going to use 3.1.  It changes answers compared to 3.0.0 for watercycle cases so can't go in a 3.0.x tag.

bishtgautam:
Yes, this is ready to be merged.

PR #6594, opened 2024-09-06
homme SYCL changes

oksanaguba:
Changes needed for spot/aurora: sycl kokkos spaces, temporary TP changes (TP init will be sorted out later), switch booleans to ints in F<-->xx interfaces.

see #6569 for the original branch/history.

[bfb]


components/homme/src/theta-l_kokkos/cxx/HyperviscosityFunctorImpl.cpp
```diff
@@ -118,9 +118,9 @@ void HyperviscosityFunctorImpl::init_params(const SimulationParams& params)
   m_eos.init(params.theta_hydrostatic_mode,m_hvcoord);
 
 #ifdef HOMMEXX_BFB_TESTING
-  m_process_nh_vars = true;
+  m_process_nh_vars = 1;
 #else
-  m_process_nh_vars = !params.theta_hydrostatic_mode;
+  m_process_nh_vars = static_cast<int> (not params.theta_hydrostatic_mode);
```
oksanaguba, line 123: Luca already made this change, so i won't revert to bool.

ambrad:
As I understand it, the int/bool code is just to work around a compiler bug affecting the F90-C++ interface; essentially, the compiler's treatment of c_bool is broken, or something like that. In particular, using bool in a kernel is _not_ an issue; we do that in several spots. Oksana, is all of that correct?

If so, I can add a commit to this branch next week that isolates the workaround to just the F90-C++ interface code. In particular, I'd like to remove the two theta_hydrostatic_mode variables from control_mod.F90. Is that OK?

oksanaguba:
> As I understand it, the int/bool code is just to work around a compiler bug affecting the F90-C++ interface; essentially, the compiler's treatment of c_bool is broken, or something like that. In particular, using bool in a kernel is _not_ an issue; we do that in several spots. Oksana, is all of that correct?

Yes, correct about int-bool issue.

> If so, I can add a commit to this branch next week that isolates the workaround to just the F90-C++ interface code. In particular, I'd like to remove the two theta_hydrostatic_mode variables from control_mod.F90. Is that OK?

yes, thanks!

rljacob:
@bartgol thank for reducing the number of commits but you could do this one more time using --author on the commit command to give author credit to Oksana?

bartgol:
Of course. I'll do it as soon as I get back to my computer

bartgol:
@rljacob @oksanaguba I edited the commits, to fix the author.

@ambrad FYI, in case you pulled and was planning to do the mods you mentioned...

bartgol (approved):

ambrad:
I think I'll have a commit to push to this branch by tonight. I've made changes and am running homme_integration on Chrysalis.

ambrad:
Done. I also fixed some moisture-related code in preqx so that homme_integration passes on Chrysalis.

ambrad (approved):

bartgol (approved):

bartgol (approved):

ambrad (approved):

oksanaguba:
homme suite pass on chrysalis

oksanaguba:
in next

oksanaguba:
hommexx ers_d test failed on cdash with an mpi error, but i ran it by hand with no issues.

PR #6596, opened 2024-09-07
For pm-cpu, upgrade Intel compiler to `2023.2.0` as well as other modules for Intel only

ndkeen:
For pm-cpu, move from `intel/2023.1.0` to `intel/2023.2.0`.
Updating to this version allows us to also update several other module versions.
These are the updates to other modules we are doing at the same *just* for Intel compiler for now:

```
PrgEnv-intel/8.3.3                   PrgEnv-intel/8.5.0
craype/2.7.20                        craype/2.7.30                    
cray-mpich/8.1.25                    cray-mpich/8.1.28                
cray-hdf5-parallel/1.12.2.3          cray-hdf5-parallel/1.12.2.9      
cray-netcdf-hdf5parallel/4.9.0.3     cray-netcdf-hdf5parallel/4.9.0.9 
cray-parallel-netcdf/1.12.3.3        cray-parallel-netcdf/1.12.3.9
```
While this change does not address a known issue, and the versions are higher than machine defaults, this is in preparation for upcoming SW changes. Also do not expect any significant performance changes, but more testing warranted.

So far, testing shows the results are BFB, but would rather not assume the PR is BFB as it changes compiler version.

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -223,8 +223,8 @@
       </modules>
 
       <modules compiler="intel">
-        <command name="load">PrgEnv-intel/8.3.3</command>
-        <command name="load">intel/2023.1.0</command>
+        <command name="load">PrgEnv-intel/8.5.0</command>
+        <command name="load">intel/2023.2.0</command>
       </modules>
 
       <modules compiler="nvidia">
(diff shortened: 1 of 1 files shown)


ndkeen:
I ran e3sm_prod and compared with baselines. I also run e3m_integration with older version of repo -- both showed all tests passing.  But I might not expect all cases to be BFB.

rljacob:
Yes run the production suite with a comparison against the existing baselines.

ndkeen:
merged to next

rljacob:
I think you can merge this to master.

ndkeen:
ok. yea I was trying to test more cases with it. In fact, I think a scream case was not BFB after this change. But we can cross that bridge when we merge.

ndkeen:
OK, for a basic `ne30pg2_ne30pg2.F2010-SCREAMv1` test case with scream repo (of Sep 10) on pm-cpu, I see this change is not BFB.

I can merge this as-is and continue testing?

rljacob:
I think so since probably no one is using pm-cpu for SCREAM v1 simulations.  If they are, they're probably using the scream repo.

ndkeen:
I tried a recent HR case `piCtl.ne120pg2_r025_RRSwISC6to18E3r5` for 2 days and it's BFB after this change.

Will merge and we will have to remember to address non-bfb issues with scream repo.

PR #6597, opened 2024-09-08
Update nitrogen deposition and population density files

wlin7:
An old SSP5 based nitrogen deposition (ndep) and population density (popdens)
files were specified as default for all v3 configurations. This PR makes them 
dependent of SSP scenarios. New ndep and popdens files are introduced for 
each SSP, and the SSP245 files is set as new default for non-SSP compsets. 

The changes will impact simulation results for F2010 and all SSP compsets. Historical 
compset simulation will be BFB after 1989. piControl simulations remain BFB.  
Cold start (e.g., same small grid F2010 nightly tests) without specifying a finidat 
would not see differences in short runs as there would be no vegetation to begin 
with as far as the fire model is concerned.

Default value for co2vmr is also updated for SSP245. This has no impact on 
simulation as its value is a placeholder for the transient compset.

[NML] diff for tests involving active land, due to changes of data stream 
           file and stream year_last_align
[non-BFB]

----------------------------------------------------------------------------------
Tests have been done to verify piControl simulations are not affected, and historical simulations start to differ 4 steps into 1990-01. This is because popdens data changes only starting 1990.

### DIFF
--- a/components/elm/bld/namelist_files/namelist_defaults.xml
+++ b/components/elm/bld/namelist_files/namelist_defaults.xml
@@ -666,7 +666,7 @@ this mask will have smb calculated over the entire global land surface
 <stream_year_first_ndep use_cn=".true."   sim_year="constant" sim_year_range="2000-2100" >2000</stream_year_first_ndep>
 <stream_year_last_ndep  use_cn=".true."   sim_year="constant" sim_year_range="2000-2100" >2100</stream_year_last_ndep>
 
-<stream_fldfilename_ndep hgrid="1.9x2.5"   use_cn=".true."   >lnd/clm2/ndepdata/fndep_elm_cbgc_exp_simyr1849-2101_1.9x2.5_c190103.nc</stream_fldfilename_ndep>
+<stream_fldfilename_ndep hgrid="1.9x2.5"   use_cn=".true."   >lnd/clm2/ndepdata/fndep_elm_cbgc_exp_simyr1849-2101_1.9x2.5_ssp245_c240903.nc</stream_fldfilename_ndep>
 <stream_fldfilename_ndep hgrid="1.9x2.5"   use_cn=".true."   rcp="8.5" >lnd/clm2/ndepdata/fndep_clm_rcp8.5_simyr1849-2106_1.9x2.5_c100428.nc</stream_fldfilename_ndep>
 <stream_fldfilename_ndep hgrid="1.9x2.5"   use_cn=".true."   rcp="6"   >lnd/clm2/ndepdata/fndep_clm_rcp6.0_simyr1849-2106_1.9x2.5_c100810.nc</stream_fldfilename_ndep>
 <stream_fldfilename_ndep hgrid="1.9x2.5"   use_cn=".true."   rcp="4.5" >lnd/clm2/ndepdata/fndep_clm_rcp4.5_simyr1849-2106_1.9x2.5_c100428.nc</stream_fldfilename_ndep>
@@ -899,8 +899,8 @@ this mask will have smb calculated over the entire global land surface
 <stream_year_first_popdens use_fates=".true."   sim_year="constant" sim_year_range="2000-2100" >1850</stream_year_first_popdens>
 <stream_year_last_popdens  use_fates=".true."   sim_year="constant" sim_year_range="2000-2100" >2010</stream_year_last_popdens>
 
-<stream_fldfilename_popdens hgrid="0.5x0.5"   use_cn=".true."                   >lnd/clm2/firedata/elmforc.ssp5_hdm_0.5x0.5_simyr1850-2100_c190109.nc</stream_fldfilename_popdens>
-<stream_fldfilename_popdens hgrid="0.5x0.5"   use_fates=".true."                   >lnd/clm2/firedata/elmforc.ssp5_hdm_0.5x0.5_simyr1850-2100_c190109.nc</stream_fldfilename_popdens>
+<stream_fldfilename_popdens hgrid="0.5x0.5"   use_cn=".true."    >lnd/clm2/firedata/elmforc.Li_20181205_mod_hist_SSP2_CMIP6_hdm_0.5x0.5_AVHRR_simyr1850-2100_c240906.nc</stream_fldfilename_popdens>
+<stream_fldfilename_popdens hgrid="0.5x0.5"   use_fates=".true." >lnd/clm2/firedata/elmforc.Li_20181205_mod_hist_SSP2_CMIP6_hdm_0.5x0.5_AVHRR_simyr1850-2100_c240906.nc</stream_fldfilename_popdens>
 
 <popdensmapalgo use_cn=".true."                               >bilinear</popdensmapalgo>
 
--- a/components/elm/bld/namelist_files/namelist_defaults.xml
+++ b/components/elm/bld/namelist_files/namelist_defaults.xml
@@ -485,48 +485,48 @@ lnd/clm2/surfdata_map/surfdata_conusx4v1_simyr2000_c160503.nc</fsurdat>
      dataset (arbitrarily, RCP8.5) for the historical period, because all of the
      RCP datasets contain the 1850-2000 period as well as the future period. -->
 
-<flanduse_timeseries hgrid="360x720cru"    sim_year_range="1850-2000"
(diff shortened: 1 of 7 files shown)


components/elm/bld/namelist_files/namelist_defaults.xml
```diff
@@ -666,7 +666,7 @@ this mask will have smb calculated over the entire global land surface
 <stream_year_first_ndep use_cn=".true."   sim_year="constant" sim_year_range="2000-2100" >2000</stream_year_first_ndep>
 <stream_year_last_ndep  use_cn=".true."   sim_year="constant" sim_year_range="2000-2100" >2100</stream_year_last_ndep>
 
-<stream_fldfilename_ndep hgrid="1.9x2.5"   use_cn=".true."   >lnd/clm2/ndepdata/fndep_elm_cbgc_exp_simyr1849-2101_1.9x2.5_c190103.nc</stream_fldfilename_ndep>
+<stream_fldfilename_ndep hgrid="1.9x2.5"   use_cn=".true."   >lnd/clm2/ndepdata/fndep_elm_cbgc_exp_simyr1849-2101_1.9x2.5_ssp245_c240903.nc</stream_fldfilename_ndep>
```
wlin7, line 669: @bishtgautam , don't know why hgrid attribute is specified and if it is needed. It has been tested it serves as a general default file independent of grid resolution. Therefore, no value is specified in the use_case file for the 1850, 2010 and 20TR compsets.

components/elm/bld/namelist_files/namelist_defaults.xml
```diff
@@ -899,8 +899,8 @@ this mask will have smb calculated over the entire global land surface
 <stream_year_first_popdens use_fates=".true."   sim_year="constant" sim_year_range="2000-2100" >1850</stream_year_first_popdens>
 <stream_year_last_popdens  use_fates=".true."   sim_year="constant" sim_year_range="2000-2100" >2010</stream_year_last_popdens>
 
-<stream_fldfilename_popdens hgrid="0.5x0.5"   use_cn=".true."                   >lnd/clm2/firedata/elmforc.ssp5_hdm_0.5x0.5_simyr1850-2100_c190109.nc</stream_fldfilename_popdens>
-<stream_fldfilename_popdens hgrid="0.5x0.5"   use_fates=".true."                   >lnd/clm2/firedata/elmforc.ssp5_hdm_0.5x0.5_simyr1850-2100_c190109.nc</stream_fldfilename_popdens>
+<stream_fldfilename_popdens hgrid="0.5x0.5"   use_cn=".true."    >lnd/clm2/firedata/elmforc.Li_20181205_mod_hist_SSP2_CMIP6_hdm_0.5x0.5_AVHRR_simyr1850-2100_c240906.nc</stream_fldfilename_popdens>
+<stream_fldfilename_popdens hgrid="0.5x0.5"   use_fates=".true." >lnd/clm2/firedata/elmforc.Li_20181205_mod_hist_SSP2_CMIP6_hdm_0.5x0.5_AVHRR_simyr1850-2100_c240906.nc</stream_fldfilename_popdens>
```
wlin7, line 903: @gaurabkcutk . It is also set to replace the file for configurations with FATES. Is it fine?

bishtgautam: @glemieux , @rgknox , The population density dataset must be updated for the big-leaf veg model. Presently, this dataset is being updated for FATES also. Is it okay to update this dataset for FATES?

components/elm/bld/namelist_files/namelist_defaults.xml
bishtgautam: @wlin7, I agree that the `hgrid` attribute is unnecessary and should be removed for all the Ndep dataset entries.

bishtgautam (approved):

rljacob:
When you say "An old SSP5 based nitrogen deposition (ndep) and population density (popdens)
files were specified as default for all v3 configurations" does that mean **ALL** including historical?  Or just all the SSP ones?

rljacob:
> Historical compset simulation will be BFB after 1989
Do you mean before 1989 ?

maltrud:
@wlin7 one of my HES responsibilities is to make sure the land and ocean are using the same ndep (which wasn't the case in v1).  is this the latest from the cmip working group?

wlin7:
> When you say "An old SSP5 based nitrogen deposition (ndep) and population density (popdens) files were specified as default for all v3 configurations" does that mean **ALL** including historical? Or just all the SSP ones?

Yes, that include SSPs.

wlin7:
> > Historical compset simulation will be BFB after 1989
> 
> Do you mean before 1989 ?

To be a little more precise, nbfb starts after 3 steps into 1990-01-01.

components/elm/bld/namelist_files/namelist_defaults.xml
wlin7: If not updating, also be aware that the file was created for 'SSP5' for BGCv1.

glemieux: Hi @bishtgautam, I think this is ok, but I'd like to check with some potential stakeholders.  I should have a decision by the end of the day today.

glemieux: @bishtgautam I'm assuming that the format of the data isn't changing between these files correct?  If that's the case, then we should be good to go.

bishtgautam: Thanks for checking with FATES colleagues. The data format isn't changing, but the values are changing.

acme-y9s:
> @wlin7 one of my HES responsibilities is to make sure the land and ocean are using the same ndep (which wasn't the case in v1). is this the latest from the cmip working group?

@maltrud The fndep of SSP2 and 3 are generated from the latest version files from ESGF input4MIP. Based on my understanding, these files are used for CMIP6 simulations.

 For the historical period (1849 to 2015), we got the NDP directly from fndep_elm_cbgc_exp_simyr1849-2101_1.9x2.5_c190103.nc which was used for BGC v1 simulations.

golaz (approved):

wlin7:
Surprised to see F2010 tests results are not affected, though  the popdens file is replaced
`stream_fldfilename_popdens = '/lcrc/group/e3sm/data/inputdata/lnd/clm2/firedata/elmforc.Li_20181205_mod_hist_SSP2_CMIP6_hdm_0.5x0.5_AVHRR_simyr1850-2100_c240906.nc'`
@bishtgautam , @acme-y9s , @thorntonpe, any idea why? We have clearly seen with this file change alone, historical run starts to differ from 1990-01-01, as expected.

acme-y9s:
@wlin7  Did you use the subset (1850-2014) of elmforc.ssp5_hdm_0.5x0.5_simyr1850-2100_c190109.nc (the one was used for BGC v1)   to replace the corresponding part of elmforc.Li_20181205_mod_hist_SSP2_CMIP6_hdm_0.5x0.5_AVHRR_simyr1850-2100_c240906.nc? If  it was the case and you also  used elmforc.ssp5_hdm_0.5x0.5_simyr1850-2100_c190109.nc for your F2010 simulation before, it won't affect your year 2010 simulation after you switched to use elmforc.Li_20181205_mod_hist_SSP2_CMIP6_hdm_0.5x0.5_AVHRR_simyr1850-2100_c240906.nc.   Am I right?

wlin7:
@acme-y9s , as we decided, only 1850-1989 data from `elmforc.ssp5_hdm_0.5x0.5_simyr1850-2100_c190109.nc`  are used to replace the corresponding years in `elmforc.Li_20181205_mod_hist_SSP2_CMIP6_hdm_0.5x0.5_AVHRR_simyr1850-2100_c240906.nc`.

wlin7:
Good points, @acme-y9s . @bishtgautam also suspect it is because the test is a cold start (no IC), thus no vegetation to begin with. The test is one of the nightly, for 3-days at ne4 grid.

I have also just confirmed with a standard v3.LR.F2010 config. The case was previously run with an finidat from [v3.LR](http://v3.lr/).amip, and it showed difference when replacing the popdens file. After resetting finidat to null, the new pair of 10-day runs with the old and new popdens files are BFB.

rljacob:
@wlin7 please also create an issue for this that describes the bug in v3.0.0.

PR #6606, opened 2024-09-11
Enables output of land fraction used in dust mobilization

bishtgautam:
Adds a new member variable (`lnd_frc_mbl_patch`) to `dust_type` to
allow the output of land fraction sued in dust mobilization. This new variable
is added to the ELM history file and is inactive by default.

[BFB] Except for the prod test suite as a new variable is added to output

### DIFF
--- a/cime_config/testmods_dirs/allactive/wcprod/user_nl_elm
+++ b/cime_config/testmods_dirs/allactive/wcprod/user_nl_elm
@@ -33,7 +33,7 @@ hist_fexcl1 = 'AGWDNPP','ALTMAX_LASTYEAR','AVAIL_RETRANSP','AVAILC','BAF_CROP',
             'SOIL4N_vr','SOIL4P_TNDNCY_VERT_TRANS','SOIL4P_TO_SMINP','SOIL4P_vr','SOLUTIONP_vr',
             'TCS_MONTH_BEGIN','TCS_MONTH_END','TOTCOLCH4','water_scalar','WF',
             'wlim_m','WOODC_LOSS','WTGQ'
-hist_fincl1 = 'SNOWDP','COL_FIRE_CLOSS','NPOOL','PPOOL','TOTPRODC'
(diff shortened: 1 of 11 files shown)


bishtgautam:
@rljacob, Based on discussion with @wlin7, @yfenganl, and me, we would like this PR to be included as part of the v3.0.1 tag, so the new dust-related variable can be included in ELM history files for the AMIP ensembles that have not yet started.

rljacob (approved):

yfenganl:
Thanks Gautam. Could you please change the long_name to "bare soil fraction for dust mobilization", if possible.

yfenganl (approved):
Thanks Gautam. Could you please change the long_name to "bare soil fraction for dust mobilization"?

rljacob:
Is there a CF standard name for this?

bishtgautam:
@yfenganl Updated the long_name.
@rljacob Yes, I don't believe this is a standard variable.

rljacob (changes requested):
Since this field will be output in production cases, please update the "prod" test mods to include it.  This will then create diffs in testing.

wlin7:
The new variable is added to the wcprod testmods. (appended to hist_fincl1 in user_nl_elm).
On reviewing, it is also found we don't need that many testmods. Some are identical, for example, the three wcprod testmods for eam). Similarly for several allactive wcprod mods. Removing duplicated testmods will change some test names though.

rljacob (approved):

peterdschwartz:
merged to next

wlin7 (approved):
The changes to associate lnd_frc_mbl_patch with the new output field look good and treated consistently.

rljacob:
I'm seeing "diff_test: the two files DIFFER only in their field lists" in prod tests which is expected but you have to be careful when blessing because the SSP prod tests are also affected by #6597 which needs another day on next.  Maybe just wait and merge them both Monday.

peterdschwartz:
merged to master

PR #6609, opened 2024-09-11
update ghci container to restore gh/ci

mahf708:
updates ghci-aci container and its corresponding machine entry.

[BFB]

### DIFF
--- a/.github/workflows/e3sm-gh-ci-cime-tests.yml
+++ b/.github/workflows/e3sm-gh-ci-cime-tests.yml
@@ -36,7 +36,7 @@ jobs:
           - SMS_D_Ln5_P4.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.ghci-oci_gnu
           - ERS_Ld5_P4.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.ghci-oci_gnu.eamxx-prod
     container: 
-      image: ghcr.io/mahf708/e3sm-imgs:v0.0.9
+      image: ghcr.io/e3sm-project/containers-ghci:ghci-0.1.0
 
     steps:
       - 
@@ -49,10 +49,6 @@ jobs:
         name: CIME
         working-directory: cime/scripts
         run: |
-          mkdir -p $HOME/projects/e3sm/cesm-inputdata/atm/cam/physprops/
-          wget https://web.lcrc.anl.gov/public/e3sm/inputdata/atm/cam/physprops/p3_lookup_table_1.dat-v4.1.2
-          mv p3_lookup_table_1.dat-v4.1.2 $HOME/projects/e3sm/cesm-inputdata/atm/cam/physprops/
-          export USER=test
           ./create_test ${{ matrix.test }} --wait --debug
       - 
         name: Artifacts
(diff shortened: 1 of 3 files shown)


.github/workflows/e3sm-gh-ci-cime-tests.yml
```diff
@@ -61,7 +57,7 @@ jobs:
         with:
           name: ${{ matrix.test }}
           path: |
-            ~/projects/e3sm/scratch/${{ matrix.test }}*/TestStatus.log
-            ~/projects/e3sm/scratch/${{ matrix.test }}*/bld/*.bldlog.*
-            ~/projects/e3sm/scratch/${{ matrix.test }}*/run/*.log.*
-            ~/projects/e3sm/scratch/${{ matrix.test }}*/run/*.cprnc.out
+            /projects/e3sm/scratch/${{ matrix.test }}*/TestStatus.log
```
bartgol, line 60: I'm assuming this uploads the test results somewhere so we can inspect them in case of failure? Would it be bad to simply cat the files so we just see them in the action output? Or is retrieving them easy from the web interface?

bartgol (approved):

.github/workflows/e3sm-gh-ci-cime-tests.yml
bartgol: Ah, I suppose cat-ing so many files may make the output hard to parse, given that we run a matrix of tests... So maybe yeah, uploading them is cleaner.

rljacob (approved):

.github/workflows/e3sm-gh-ci-cime-tests.yml
bartgol: Long shot, but we could cat them only if there's a fail maybe? We could run create test step by step. In pseudo code:
```yaml
   - create_test ${{ matrix.test }} --no-setup
   - cd $output_root/${{ matrix.test }}*
   - case.setup || cat TestStatus.log
   - case.build || cat bld/*.bldlog.*
   - case.submit || cat run/e3sm.log.*
```
It is not correct/enough, since a) ERP/PEM tests create case2run and case2bld folders, and b) the fail at submit may come from cprnc, not actual run. 
I guess it would be nice if CIME supported catting the correct file in case of a test fail. E.g., something like `create_test <TEST> --output-on-failure` or something, would output the bldlog/log/cprnc.out of the step that failed...

rljacob: I'm not sure what you mean.  cat them to the screen?   If you run create_test interactively, it will tell you where to look when there's a fail.  Catting a file to a screen would just scroll hundreds of lines of text on your screen.

bartgol: Yes, I meant to cat them to screen, so that they are readily visible on the action output on gh. But I don't know if one can scroll back to very old aciton runs or if those also expire after 90 days. If they expire, then at this point, there's not much added value in cat-ing them. We can just be satisfied with the logs uploaded as artifacts ...

rljacob:
@mahf708 is this ready?

PR #6611, opened 2024-09-12
Log SCORPIO buffer size limit for default case

### LINKED ISSUE #6610: Output PIO_BUFFER_SIZE_LIMIT in logs when user chooses the default (-1) in CIME
The SCORPIO internal buffer cache size (CIME: env_run.xml: PIO_BUFFER_SIZE_LIMIT) used by the library is written out in e3sm log only when the user explicitly sets the cache size in env_run.xml (it is not written out in the default case).

@ndkeen suggested that it would be helpful to users to add that info (for the default case) in the e3sm logs.

jayeshkrishna:
Logging the SCORPIO internal buffer size in the
coupler log when the user lets the library choose the
buffer size (the default case)

Fixes #6610 
[BFB]

### DIFF
--- a/share/util/shr_pio_mod.F90
+++ b/share/util/shr_pio_mod.F90
@@ -173,17 +173,24 @@ contains
     logical, intent(in) :: comp_iamin(:)
     character(len=*), intent(in) :: comp_name(:)
     integer, intent(in) ::  comp_comm(:), comp_comm_iam(:)
+    integer(kind=pio_offset_kind) :: cur_buffer_size_limit=-1
     integer :: i
     character(len=shr_kind_cl) :: nlfilename, cname
     integer :: ret
     character(*), parameter :: subName = '(shr_pio_init2) '
 
     ! 0 is a valid value of pio_buffer_size_limit
-    if(pio_buffer_size_limit>=0) then
+    ! -1 is the value used by CIME to let the library choose the buffer limit
+    if(pio_buffer_size_limit>=-1) then
+       call pio_set_buffer_size_limit(pio_buffer_size_limit, prev_limit=cur_buffer_size_limit)
        if(comp_comm_iam(1)==0) then
-          write(shr_log_unit,*) 'Setting pio_buffer_size_limit : ',pio_buffer_size_limit
+          if(pio_buffer_size_limit >= 0) then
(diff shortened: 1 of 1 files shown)


dqwu (approved):

ndkeen:
Be careful when doing things like 
`integer(kind=pio_offset_kind) :: cur_buffer_size_limit=-1`
as it makes us think it is setting this variables each time it enters routine.
But it's not -- only first time.  We've had to fix some issues that were caused by this fortran behavior.
So this is only OK if routine only called once, but to me, maybe just avoid altogether.

ndkeen:
I'm assuming you've tested functionality. 
Just looking at what's written to log file: It does the job fine -- allowing us to see buffer size actually used.
But it might be nice to separate into 2 pieces of information: the key=value _and_ method used to set it.
Different ways to implement and it's not a big deal

but as example:
```
          if(pio_buffer_size_limit >= 0) then
            write(shr_log_unit,*) 'Use pio_buffer_size_limit=', pio_buffer_size_limit, ' (bytes) set via change to PIO_BUFFER_SIZE_LIMIT'
          else
            ! Default pio_buffer_size_limit
            write(shr_log_unit,*) 'Use pio_buffer_size_limit=', cur_buffer_size_limit, ' (bytes) which is the default value (PIO_BUFFER_SIZE_LIMIT=-1)'
          end if
```

jayeshkrishna:
One potential issue is that PIO_BUFFER_SIZE_LIMIT is a CIME setting (How much CIME language do we need here?). Adding the unit of the setting, as you suggested, is a good idea IMO.

I can change the output statement as follows,
```
          if(pio_buffer_size_limit >= 0) then
            write(shr_log_unit,*) 'Set pio_buffer_size_limit to : ', pio_buffer_size_limit, ' bytes'
          else
            ! Default pio_buffer_size_limit
            write(shr_log_unit,*) 'Using pio_buffer_size_limit (default): ', cur_buffer_size_limit, ' bytes'
          end if
```

Would that work?

jayeshkrishna:
@ndkeen : Please let me know if the latest code is ready to be merged (if so, please go ahead and review/approve the PR).

ndkeen (approved):

rljacob:
@jayeshkrishna you can merge this today

PR #6613, opened 2024-09-12
Allow interpinic to utilize any netCDF filetype, allow superfluous month dimension in input file

czender:
[BFB] [Land] [Bugfix] 
Modifications to `bishtgautam/lnd/interpinic-gnu` to improve `interpinic` to work with output datasets of any format. Previously, initerpinic would fail if output format was not `NC_FORMAT_64BIT_OFFSET`. Our new deep (16-layer) snowpack modifications require restart fields that exceed 4 GB in size for r0125 ELM grids. This new feature successfully interpolates to new initial conditions that require storage in 'NC_FORMAT_64BIT_DATA' or netCDF4 formats. These modifications also remove the (sometimes) unnecessary constraint that if the input dataset contains the `month` dimension, then so must the output dataset. In practice the `month` dimension may be present in the input dataset though unused by any variable (i.e., it is superfluous). Thus requiring its presence in the output dataset can create an extra hurdle. Instead of dying, `interpinic` now prints a warning about the situation and proceeds anyway.

### DIFF
--- a/components/elm/tools/interpinic/src/fmain.F90
+++ b/components/elm/tools/interpinic/src/fmain.F90
@@ -14,7 +14,7 @@ program fmain
   character(len= 256) :: arg
   integer :: n                   !index 
   integer :: nargs               !number of arguments  
-  integer, external :: iargc     !number of arguments function
+  integer  :: iargc     !number of arguments function
   character(len=256) :: finidati !input initial dataset to read
   character(len=256) :: finidato !output initial dataset to create
   character(len=256) :: cmdline  !input command line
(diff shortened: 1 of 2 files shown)


xylar:
@czender, you're including several submodule updates here that I think we probably don't want.  The easiest way to fix this might just be to redo commit 9d2efbb76fa9f6c43e49383a17340e334acc9496 to only include the changes in the one file you want to keep.

xylar:
Also, I wonder if I might ask you all to include one more patch here to fix something that is giving me trouble with modern compliers:
https://github.com/conda-forge/e3sm-tools-feedstock/blob/main/recipe/0001-fix-boz-erros-in-interpinic-build.patch

czender:
Will close this and re-open in new PR that includes Xylar's patch and hopefully doesn't modify any submodules. This is my first E3SM PR so bear with me. @chloewhicker is doing the hand-holding here...

PR #6616, opened 2024-09-13
Updates for RCEMIP phase 2

whannah1:
This adds compsets and new output variables to facilitate the E3SM and E3SM-MMF
contributions to phase 1 & 2 of the radiative-convective equilibrium model inter-comparison project (RCEMIP).

NOTE: in the original RCE compsets there was a slight error in the value of constant_zenith_deg being off by 1/100. This difference is inconsequential, but I wanted to fix it here to ensure consistency with the other RCEMIP models.

[non-BFB] - only for RCE cases because of updated constant_zenith_deg value


components/data_comps/docn/src/docn_comp_mod.F90
```diff
@@ -1174,6 +1175,55 @@ subroutine prescribed_sst(xc, yc, lsize, sst_option, sst)
        end do
     end if
 
+    !-------------------------------------------------------------------------------
```
bogensch, line 1178: Nitpicky, but this seems like a lot of unnecessary copy and pasting of code that could be easily condensed.

bogensch (approved):
Looks good.  I think if I did it I would have only one MW compset with the SST mean and delta being namelist parameters that could be defined at will by the user (similar to how I did it for the branch I have on SCREAM repo).  But I also understand why you did it the way you did so that's just a comment.

components/data_comps/docn/src/docn_comp_mod.F90
whannah1: Yea, good point - I'll clean it up.

whannah1: @bogensch can you take a look at how I updated this block?

bogensch (approved):

rljacob:
@whannah1 did you expect some namelist diffs?
  BASE: constant_zenith_deg = 42.05
  COMP: constant_zenith_deg = 42.04

Only in ERP_Ln9.ne4pg2_ne4pg2.FRCE and ERS_Ln9.ne4pg2_ne4pg2.FRCE-MMF1

whannah1:
> @whannah1 did you expect some namelist diffs which lead to diffs with baselines? BASE: constant_zenith_deg = 42.05 COMP: constant_zenith_deg = 42.04
> 
> Only in ERP_Ln9.ne4pg2_ne4pg2.FRCE and ERS_Ln9.ne4pg2_ne4pg2.FRCE-MMF1

Yes, sorry I forgot to make a note about this. When I originally created the RCE compsets I made a tiny mistake in the value of `constant_zenith_deg`, which is inconsequential, but I wanted to correct it here anyway. I'll edit the description to clarify that the diff is expected.

rljacob:
Also seeing a runtime error with this test: ERP_Ld3.ne4pg2_ne4pg2.FIDEAL.chrysalis_intel.allactive-pioroot1 on both chrysalis and pm-cpu.

PR #6620, opened 2024-09-16
Get mappy working again with RHEL9

jgfouca:


rljacob (approved):
What kind of testing did you do?

wadeburgess (commented):
Do you not need to load a SEMS env? Or run "source /projects/sems/modulefiles/utils/sems-modules-init.sh"

rljacob:
@jgfouca was the overnight testing supposed to work last night?

jgfouca:
@rljacob , I thought it would. Jenkins is able to connect to mappy now and I was able to build and run a couple A cases. Unfortunately, there are more problems coming up. The most concerning is that performance is abysmal and most cases are timing out.

Issue #6302, opened 2024-03-19
MOSART is not stopping on "Negative channel storage found! "


bishtgautam:
@tanzeli1982 found that while MOSART reports a warning about negative channel storage, it doesn't stop.

```
17: Warning: Negative channel storage found! -
8.932405345574364E+051
193: Warning: Negative channel storage found! -
9.479406226017329E+081
449: Warning: Negative channel storage found! -
2.183048555764361E+087
23: Warning: Negative channel storage found! -
1.674015257872732E+080
456: Warning: Negative channel storage found! -
9.508709584880304E+073
```

liho745:
@bishtgautam I found that the "shr_sys_abort" line corresponding to this error was commented off in a very early commit, #f05d5e5f2f6ea3be7d4293d67bc513a5c1c584f1, back in 2015.  Is it better to fix the bug in a branch off this early commit, or a branch off the most recent master branch?

rljacob:
Does this happen in regular water cycle cases?  If not, just fix it from head of master.

bishtgautam:
This negative error does not show up in the regular water cycle simulations because the MOSART grid resolution is coarse. However, if the resolution is increased (as being done in the ICoM project), we get this warning instead of the model crashing.

@liho745 As Rob suggests, let's add the fix from the head of the master.

liho745:
PR #6313 has been submitted to fix this issue.

PR #6568, opened 2024-08-28
Fixing a bug causing model crash by avoiding negative channel storage

liho745:
MOSART can be forced to crash if negative channel storage values are produced mainly due to round errors, 
when the channel storage is completely depleted in a single time step. 
The crashing was reported in several cases of E3SM fully coupled runs, 
where the kinematic waving routing method was invoked by default. 
Jon Wolfe was able to reproduce the crashing error in 
a low-res run and print out detailed information to identify the cause.
 In this fix, a numerical treatment has been implemented 
in the kinematic wave routing method, subroutine Routing_KW(), 
to ensure that any channel storage does not get completely 
depleted in a single time step.
 This treatment is physically reasonable as long as MOSART runs 
at a sub-daily time step. 
Jon Wolfe has tested this treatment in his low-res coupled run, 
and it helped avoid crashing.  

[non-BFB]  for MOSART

PR #6623, opened 2024-09-17
Revert PR #6313 aborting run on negative channel storage

### LINKED ISSUE #6622: MOSART negative channel storage stopped v3.LR simulation
The master code base that includes the changes via PR #6313, can cause v3.LR.piControl simulation to stop running due to negative channel storage that is no more tolerated. This was not an issue for the original  v3.LR.piControl simulation with v3.0.0 tag.

Similar issue was also reported in the comment section of #6313.

For debugging,  a branch run based off year 0416-01-01 of [v3.LR.piControl](https://acme-climate.atlassian.net/wiki/spaces/CM/pages/4176806371/v3.LR.piControl) can reproduce the issue in 8 months (occur on 0416-08-25), using master branch that has #6313 in place (make sure not being reverted)

Since its occurrence is unpredictable, and v3.LR simulation campaigns still have a high number of long simulations (single forcings over historical period, SSPs) to be completed with commits in the latest master branch (until v3.0.1 is tagged), we may have to revert #6313. The revert can then be reverted after v3.0.1 tag is created. And after that, a physics-based fix will still need to be implemented to either prevent the negative storage from occurring, apply a fixer,  or make it more tolerant.

@peterdschwartz @hydrotian , please weigh in.

wlin7:
The enforced abort whenever MOSART has negative channel storage
may prevent v3.LR simulations from finishing, if the code base that includes
#6313 is used.  This PR is to revert #6313 until v3.0.1 tag is created,
to ensure new v3.LR simulations that need to use the latest commits to master
are free from this issue.

Fixes #6622.
[BFB]

----------------------------------------------
Issue #6302 is expected to reemerge.

### DIFF
--- a/components/mosart/src/riverroute/MOSART_physics_mod.F90
+++ b/components/mosart/src/riverroute/MOSART_physics_mod.F90
@@ -682,8 +682,8 @@ MODULE MOSART_physics_mod
 
 ! check for negative channel storage
     if (negchan < -1.e-10) then
-       write(iulog,*) 'Error: Negative channel storage found! ',negchan
-       call shr_sys_abort('mosart: negative channel storage')
+       write(iulog,*) 'Warning: Negative channel storage found! ',negchan
+!       call shr_sys_abort('mosart: negative channel storage')
     endif
     TRunoff%flow = TRunoff%flow / Tctl%DLevelH2R
     TRunoff%erowm_regi(:,nt_nmud:nt_nsan) = TRunoff%erowm_regi(:,nt_nmud:nt_nsan) / Tctl%DLevelH2R


bishtgautam (approved):

components/mosart/src/riverroute/MOSART_physics_mod.F90
```diff
@@ -682,8 +682,8 @@ subroutine Euler
 
 ! check for negative channel storage
     if (negchan < -1.e-10) then
-       write(iulog,*) 'Error: Negative channel storage found! ',negchan
-       call shr_sys_abort('mosart: negative channel storage')
+       write(iulog,*) 'Warning: Negative channel storage found! ',negchan
```
rljacob, line 685: are you sure you want to print a warning?  Was that the old behaviour?

rljacob: Yes looks like it was.

liho745: Another PR, #6568, resolved this aborting issue with a cleaner solution. Should we discuss and then discuss how to proceed with these two options?

rljacob: This one will have to go in for the 3.0.1 tag.  Then your other fix can be used for 3.1.0.

peterdschwartz:
This is going in today?

wlin7:
I just merged it to next, @peterdschwartz , after seeing Rob's message on slack.

PR #6636, opened 2024-09-22
fixes for eamxx running from e3sm repo for sycl changes

### LINKED ISSUE #6635: class "Homme::SimulationParams" has no member "moisture"
PR #6594 broke eamxx when it was merged.

example errors:

```
/global/u2/e/e3smtest/jenkins/workspace/ACME_perlmutter_e3sm_gpucxx_gnugpu/E3SM/components/eamxx/src/dynamics/homme/eamxx_homme_process_interface.cpp(822): error: class "Homme::SimulationParams" has no member "moisture"
```

oksanaguba:
fixes for eamxx running from e3sm repo for sycl changes

Fixes #6635 

[bfb]

### DIFF
--- a/components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
+++ b/components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
@@ -277,7 +277,7 @@ apply_iop_forcing(const Real dt)
 
   ElementOps elem_ops;
   elem_ops.init(hvcoord);
-  const bool use_moisture = (params.moisture == Homme::MoistDry::MOIST);
+  const bool use_moisture = params.use_moisture;
 
   // Load data from IOP files, if necessary
   m_iop->read_iop_file_data(timestamp());
--- a/components/eamxx/src/dynamics/homme/eamxx_homme_process_interface.cpp
+++ b/components/eamxx/src/dynamics/homme/eamxx_homme_process_interface.cpp
@@ -819,7 +819,7 @@ void HommeDynamics::init_homme_views () {
   std::stringstream msg;
   msg << "\n************** HOMMEXX SimulationParams **********************\n\n";
   msg << "   time_step_type: " << Homme::etoi(params.time_step_type) << "\n";
-  msg << "   moisture: " << (params.moisture==Homme::MoistDry::DRY ? "dry" : "moist") << "\n";
+  msg << "   moisture: " << (params.use_moisture ? "moist" : "dry") << "\n";
   msg << "   remap_alg: " << Homme::etoi(params.remap_alg) << "\n";
   msg << "   test case: " << Homme::etoi(params.test_case) << "\n";
   msg << "   ftype: " << Homme::etoi(params.ftype) << "\n";
--- a/components/eamxx/src/dynamics/homme/interface/homme_driver_mod.F90
+++ b/components/eamxx/src/dynamics/homme/interface/homme_driver_mod.F90
@@ -4,7 +4,7 @@
 #endif
 
 module homme_driver_mod
-  use iso_c_binding,     only: c_ptr, c_int, c_double, c_bool, C_NULL_CHAR
+  use iso_c_binding,     only: c_ptr, c_int, c_double, C_NULL_CHAR
 
   use parallel_mod,  only: abortmp
 
@@ -192,7 +192,7 @@ contains
                                  elem, hybrid, hvcoord, deriv, tl
 
     ! Local variable
-    logical(kind=c_bool), parameter :: allocate_buffer = .false.
+    logical, parameter :: allocate_buffer = 0
 
     if (.not. is_data_structures_inited) then
       call abortmp ("Error! 'prim_init_data_structures_f90' has not been called yet.\n")
--- a/components/eamxx/src/dynamics/homme/interface/homme_driver_mod.F90
+++ b/components/eamxx/src/dynamics/homme/interface/homme_driver_mod.F90
@@ -192,7 +192,7 @@ contains
                                  elem, hybrid, hvcoord, deriv, tl
 
     ! Local variable
-    logical, parameter :: allocate_buffer = 0
+    logical, parameter :: allocate_buffer = .false.
 
     if (.not. is_data_structures_inited) then
       call abortmp ("Error! 'prim_init_data_structures_f90' has not been called yet.\n")


ambrad:
This is fine as long as we're OK with doing the changes in this repo rather than in the SCREAM repo. I think it's reasonable in this case.

ambrad (approved):

ambrad:
> Do we need to issue a PR cherry-picking this into SCREAM?

No. The next upstream merge to SCREAM will bring in the Hxx changes (original PR) and these current ones, consistently updating SCREAM.

bartgol (approved):

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -277,7 +277,7 @@ apply_iop_forcing(const Real dt)
 
   ElementOps elem_ops;
   elem_ops.init(hvcoord);
-  const bool use_moisture = (params.moisture == Homme::MoistDry::MOIST);
+  const bool use_moisture = params.use_moisture;
```
ambrad, line 280: I just remembered: It's better to remove this line entirely. The local variable `use_moisture` is not actually used. This is already fixed in the downstream repo.

ambrad (changes requested):
See comment re: IOP local variable.

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
oksanaguba: That did not work. There are calls in that file that use it, dpxx version of this file is very different.

ambrad: Ok.

We might actually not want to merge this PR so that our next up- and downstream merges do not hit conflicts. A better course of action might be to (1) do an upstream merge (E3SM -> SCREAM) to bring in the Hxx changes, fix the issues due to these, then (2) do a downstream merge (SCREAM -> E3SM) to fix the EAMxx build in E3SM.

@bartgol @jgfouca or others, do you have an opinion on the best course of action?

I'll approve this PR as is in case merging this PR is decided as best.

ambrad (approved):

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
ambrad: Simple alternative: We can easily work around these issue with some temporary changes in Hxx. Then after the repos are unified, we'd back out the workarounds. That way, only Hxx code would be touched for now.

oksanaguba: i am not sure what you mean by the alternative?

ambrad: Add back the enum in the simulation params, temporarily being redundant. Undo the c_bool change (it's not interfacing with C code, so it doesn't matter, but also we eventually don't want to use c_bool in pure F90).

bartgol: Ugh, that's annoying. Hopefully it's one of the last times that this happens, as we look to unify repos in the next few weeks.

Meanwhile, it seems silly to keep pointless code in E3SM repo, just to make downstream SCREAM repo compile once we do an E3SM->SCREAM merge. So maybe the "better course of action" Andrew described is indeed cleaner.

So I think we should only do the mods in here that are needed to get stuff in the E3SM repo to build _as is_. Then, e3sm->scream merge, fixing the issues peculiar to scream repo, then do scream->e3sm merge.

As a byproduct, as we are looking to unify the repos soon, this up/down-stream merge sequence may be healthy anyways, to ensure the two are not too far apart in a few weeks when we unify them...

ambrad: Luca, just so I'm sure we're sync'ed, here are some points of clarification:

> Meanwhile, it seems silly to keep pointless code in E3SM repo, just to make downstream SCREAM repo compile once we do an E3SM->SCREAM merge. So maybe the "better course of action" Andrew described is indeed cleaner.

The key motivation is to get the E3SM EAMxx test running again. That can be done in any of three ways:

1. Touch EAMxx code in the E3SM repo (this PR).
2. Upstream merge, fix in SCREAM, downstream merge.
3. Put some temporary workarounds in E3SM Hxx code. This has the side effect of making both repos happy but, again, it's the E3SM EAMxx test that is the focus.

> So I think we should only do the mods in here that are needed to get stuff in the E3SM repo to build as is. Then, e3sm->scream merge, fixing the issues peculiar to scream repo, then do scream->e3sm merge.

If we're doing merges in the near term, I'd advocate against making EAMxx changes in the E3SM repo. Point 2 above is my version of what you wrote.

> As a byproduct, as we are looking to unify the repos soon, this up/down-stream merge sequence may be healthy anyways, to ensure the two are not too far apart in a few weeks when we unify them...

I agree.

bartgol: Yes, option 2 above is what I was thinking you were suggesting as the better option.

What I'm not really clear is: do EAMxx tests in E3SM repo need this PR? If the answer is yes, then I do think we should do mods in here. My understanding was that the mod at line you commented on is needed only after E3SM->SCREAM merge happens, so that the SCREAM repo can build. If that's the case, then I vote against doing the change. But if the mod is needed also in the E3SM repo, then it's different.

That said, I think EAMxx is not really used in E3SM repo yet, so maybe all of this talking is moot, and we should do whatever is easier to get the two repos in sync...

bartgol: But this action is currently passing even without this PR, right? If so, then I don't think the PR is needed in E3SM, and we can proceed with E3SM->SCREAM merge, fix in SCREAM repo, SCREAM->E3SM merge.

ambrad: @mahf708 we might consider disabling the EAMxx test in the github CI for now so other PRs don't get red x's.

bartgol: Ah, sorry, I was looking at e3sm-gh-ci-w-cime-tests (quite a high similarity index!).

At this point, I don't think any approach is a clear winner for what I care, so I think I am ok with whatever everyone else decides.

ambrad: I'd like to keep the eventual downstream merge clean but don't want to burden this current issue with merges now, so I vote for the Hxx-isolated workaround. If that gets the majority of votes, I'll do it tomorrow (Tuesday) and will try to merge it to next tomorrow, as well. I suggest the github CI be modified so other PRs aren't affected by this issue while the PR goes through the merge process.

oksanaguba: sounds like Andrew will re-do needed changes? let me know if i misunderstood, thanks!

ambrad: Yes, I'll submit a PR later today and see if others object to the approach.

ambrad: >  have disabled the EAMxx tests on PRs (disabled all 4 F tests but 2 W tests

Thanks @mahf708. I looked at other PRs and note that really only this test needs to be disabled: SMS_D_Ln5_P4.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.ghci-oci_gnu, not all the F cases. Up to you; I'm just pointing out disabling all four F tests is not necessary.

ambrad: The simple alternative I proposed is here: https://github.com/E3SM-Project/E3SM/pull/6645.

ambrad:
I'm closing this PR now that https://github.com/E3SM-Project/E3SM/pull/6645 is in next.

PR #6637, opened 2024-09-23
add gh/ci eamxx standalone testing

mahf708:
Adds support for running eamxx standalone testing in the gh/ci containers, for now only single-precision tests due to a mix of issues. Addresses two issues that have already been fixed in the scream fork. Adds flexibility to build cprnc in eamxx.

[BFB]


components/eamxx/cmake/machine-files/ghci-oci.cmake
```diff
@@ -1,2 +1,15 @@
...
+
+# TODO: figure out a better way to handle this, e.g.,
+# TODO: --map-by ppr:1:node:pe=1 doesn't work with mpich,
+# TODO: but -map-by core:1:numa:hwthread=1 may work well?
+# TODO: this will need to be handled in EKAT at some point
+set(EKAT_MPI_NP_FLAG "-np" CACHE STRING "-np")
+
+# TODO: hack in place to get eamxx to recognize CPRNC
+# TODO: See note in BuildCprnc.cmake...
+set(ENV{CCSM_CPRNC} "/usr/local/packages/bin/cprnc")
```
jgfouca, line 15: How is cprnc in the usr/local area?

jgfouca (approved):

components/eamxx/cmake/machine-files/ghci-oci.cmake
jgfouca: Oh, I didn't know cprnc was supported in spack, that's cool!

.github/workflows/eamxx-gh-ci-standalone.yml
```diff
@@ -0,0 +1,69 @@
+name: gh-standalone
+
+on:
+  pull_request:
+    branches: [ master ]
```
bartgol, line 5: Maybe you want to address these later, but I think we may as well polish now before merging...

Can we add also the `types` entry here? I don't want the action to trigger for any pr event. I think `types: [opened, synchronize, ready_for_review, reopened]` is a good set. Otherwise it would trigger for pointless stuff such as adding/removing labels, closing the pr, reassigning...

bartgol: And maybe I would also add a concurrency check. If a commit is pushed while testing, we should kill any other existing PR. Something like

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event.pull_request.number || github.run_id }}
  cancel-in-progress: true
```

Btw, I'm not sure why we would want a workflow_dispath trigger here. And if you removed "workflow_dispatch", the group could simply be `${{ github.workflow }}-${{ github.event.pull_request.number}}`. But I don't feel strongly about this, so I'm ok keeping workflow_dispatch.

.github/workflows/eamxx-gh-ci-standalone.yml
```diff
@@ -0,0 +1,69 @@
...
+    paths:
+      # first, yes to these
+      - '.github/workflows/eamxx-gh-ci-standalone.yml'
+      - 'cime_config/**'
```
bartgol, line 9: Why would standalone be triggered by other components mods (including driver) or cime_config mods?

bartgol: Ah, I think cime_config mods may impact b/c of how you setup the env for ghci-oci in machine_specs.py. If that's the case, maybe we could narrow it down to `cime_config/machine/config_machines.xml`? Also, if what I'm saying is true, you should also trigger if there's a change in the cime submodule, since get_case_env may change (but I suppose we can assume chances of issues from this kind of changes are very slim)

.github/workflows/eamxx-gh-ci-standalone.yml
```diff
@@ -0,0 +1,69 @@
...
+      - 'components/eam/**'
```
bartgol, line 10: I guess we need something from eam, namely a few files for p3/shoc and the rrtmgp submodule, correct? If so, let's limit ourselves to those changes. I don't want standalone to trigger for every single change in eam...

components/eamxx/cmake/machine-files/ghci-oci.cmake
bartgol: yeah, neat that they have it!

.github/workflows/eamxx-gh-ci-standalone.yml
bartgol: I have to scold my AI bot then. It told me the workflow would trigger for all types of pr event.

bartgol:
What do you mean by "reduce the space in the runner"?

bartgol:
Aren't the limits set on the container image? Why do we care about the runner?

bartgol:
> > Aren't the limits set on the container image? Why do we care about the runner?
> 
> I am not sure what you mean...
> 
> The limit comes from the hardware (bare-metal), which has the runner image (ubuntu-latest), which has the container image (ghci-container). In other words, ghci-container ⊆ ubuntu-latest ⊆ bare-metal. Inside the ghci-container, you can use as much as you want (i.e., you can run a multi-TB sim if there is space...)

Ah, so we are not fitting on the actual machine where things end up running? I see. Re: container image size, I was thinking gh has a limit in size for images that one can upload, and that's it. Since the runner is not ours, I thought "there's no limit on our end". But yes, while we are not responsible for the runner size, the limit is hardware based.

Wow, 150GB for the runner, that's just nonsense. Why do they make them this big? Have you considered asking for alpine instead? We may have to install something at startup though, which may take some time... Alternatively, maybe a basic debian?

rljacob:
Is this still a draft?

.gitmodules
```diff
@@ -22,8 +22,8 @@
 	branch = scorpio_classic
 [submodule "cosp2"]
 	path = components/eam/src/physics/cosp2/external
-	url = git@github.com:CFMIP/COSPv2.0.git
-	branch = CESM_v2.1.4
+	url = git@github.com:bartgol/COSPv2.0.git
+	branch = bartgol/fix-cosp_optical_inputs
```
bartgol, line 26: I have to change this. That branch got merged into the main cosp branch...

.github/workflows/eamxx-gh-ci-standalone.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: |
+          # TODO: get rid of this extra line if we can?
+          git config --global safe.directory '*'
+          ./components/eamxx/scripts/test-all-scream -m ghci-oci -t ${{ matrix.test }}
+      - 
+        name: Artifacts
+        uses: actions/upload-artifact@v4
+        if: ${{ always() }}
+        with:
+          name: ${{ matrix.test }}
```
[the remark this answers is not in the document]
bartgol: I am ok with short names for actions names, since they appear at the bottom of the PR page, and long names are problematic. But artifacts are in the action page, and there's plenty of space for a longer name. That said, I think it's very subjective, hence the approval regardless.

.gitmodules
rljacob: Is this ready or still needs a change?

Issue #6642, opened 2024-09-24
Ocean conservation check accumulated sums set to zero at restart.


mark-petersen:
The ocean conservation check analysis member overwrites the first entry in the file with a zero after restarts for some variables.  The result is that the first day's entry in a monthly conservation check file is correct for continuous runs but zero for a run with a restart break, for some variables.

mark-petersen:
@xylar can you describe the problem in more detail here? 

I ran tests and I don’t see what you are seeing. I ran QU240 with `config_am_conservationcheck_enable = .true.` for all cases, and everything standard from the repo for four cases:

1. master, 2 months
2. master 1mo+restart+1 mo
3. pr 2 months
4. pr 1mo+restart+1 mo

writing restarts every month. For reference, I used
```
./create_newcase -case $CASE_ROOT/$E3SM_CASE   -compiler gnu   -mach chrysalis   -project e3sm \
   -compset GMPAS-IAF -res T62_oQU240 --walltime 00:30:00
```
I can’t find any zeroing behavior in any variables with master at the month break (2).  For example,
```
$ pwd
/lcrc/group/e3sm/ac.mpetersen/scratch/chrys/240924f.ConsCheckMasterRestart.T62_oQU240.GMPAS-IAF.chr/run

$ ncdump -v accumulatedRainFlux  *mpaso.hist.am.conservationCheck*01-02-01.nc |tail

 accumulatedRainFlux = 12129038639.0062, 12114183191.7286, 12117918123.2804,
   ...
    12056372275.0261, 12054120109.5554, 12063708472.0842, 12081661853.1934,
    12074698994.5417 ;
}
```
ends with exactly the same value with or without the restart (case 1 and 2 above) for month 2. Also, these values are not increasing, so it is accumulated within each daily interval, and then zeroed after writing this daily value. So users have to accumulate the sums themselves.

What variable(s) did you see it in? Could you provide a path?

xylar:
Thanks @mark-petersen.  I didn't appreciate that the conservation check is just computing daily totals.  I seem to be mistaken that restart files are needed, in that case.

However, your run does show the same issues that I was seeing.  It isn't for all variables, and now it becomes clear that that might not be expected.  But many variables in your output file above do show strange behavior as I was seeing after restarts:
```
$ ncdump *mpaso.hist.am.conservationCheck*01-02-01.nc
```
As one example:
```
[...]

PR #6643, opened 2024-09-24
Correct ocean conservation check settings

mark-petersen:
Currently, the ocean conservation check analysis member overwrites the first entry in the file with a zero after restarts for some variables. This PR corrects this behavior so that the first day's entry in a monthly conservation check file is identical between continuous runs and a run with a restart break.

Fixes https://github.com/E3SM-Project/E3SM/issues/6642

[NML] for some mpaso resolutions
[BFB] for all tested files

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -266,6 +266,7 @@ _TESTS = {
             "SMS_D_Ld1.T62_oQU240wLI.GMPAS-IAF-PISMF.mpaso-impl_top_drag",
             "SMS_D_Ld1.T62_oQU240.GMPAS-IAF.mpaso-harmonic_mean_drag",
             "SMS_D_Ld1.T62_oQU240.GMPAS-IAF.mpaso-upwind_advection",
+            "ERS_D.T62_oQU240.GMPAS-IAF.mpaso-conservation_check",
             )
         },
 
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -266,7 +266,7 @@ _TESTS = {
             "SMS_D_Ld1.T62_oQU240wLI.GMPAS-IAF-PISMF.mpaso-impl_top_drag",
             "SMS_D_Ld1.T62_oQU240.GMPAS-IAF.mpaso-harmonic_mean_drag",
             "SMS_D_Ld1.T62_oQU240.GMPAS-IAF.mpaso-upwind_advection",
-            "ERS_D.T62_oQU240.GMPAS-IAF.mpaso-conservation_check",
+            "ERS_Ld5_D.T62_oQU240.GMPAS-IAF.mpaso-conservation_check",
             )
         },
 
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -1232,37 +1232,10 @@
 
 <!-- AM_conservationCheck -->
 <config_AM_conservationCheck_enable>.false.</config_AM_conservationCheck_enable>
-<config_AM_conservationCheck_enable ocn_grid="SOwISC12to60E2r4">.true.</config_AM_conservationCheck_enable>
-<config_AM_conservationCheck_enable ocn_grid="ECwISC30to60E2r1">.true.</config_AM_conservationCheck_enable>
-<config_AM_conservationCheck_enable ocn_grid="IcoswISC30E3r5">.true.</config_AM_conservationCheck_enable>
-<config_AM_conservationCheck_enable ocn_grid="IcosXISC30E3r7">.true.</config_AM_conservationCheck_enable>
-<config_AM_conservationCheck_enable ocn_grid="FRISwISC08to60E3r1">.true.</config_AM_conservationCheck_enable>
-<config_AM_conservationCheck_enable ocn_grid="FRISwISC04to60E3r1">.true.</config_AM_conservationCheck_enable>
-<config_AM_conservationCheck_enable ocn_grid="FRISwISC02to60E3r1">.true.</config_AM_conservationCheck_enable>
-<config_AM_conservationCheck_enable ocn_grid="FRISwISC01to60E3r1">.true.</config_AM_conservationCheck_enable>
-<config_AM_conservationCheck_enable ocn_grid="RRSwISC6to18E3r5">.true.</config_AM_conservationCheck_enable>
 <config_AM_conservationCheck_compute_interval>'dt'</config_AM_conservationCheck_compute_interval>
 <config_AM_conservationCheck_output_stream>'conservationCheckOutput'</config_AM_conservationCheck_output_stream>
 <config_AM_conservationCheck_compute_on_startup>.false.</config_AM_conservationCheck_compute_on_startup>
-<config_AM_conservationCheck_compute_on_startup ocn_grid="SOwISC12to60E2r4">.true.</config_AM_conservationCheck_compute_on_startup>
-<config_AM_conservationCheck_compute_on_startup ocn_grid="ECwISC30to60E2r1">.true.</config_AM_conservationCheck_compute_on_startup>
-<config_AM_conservationCheck_compute_on_startup ocn_grid="IcoswISC30E3r5">.true.</config_AM_conservationCheck_compute_on_startup>
-<config_AM_conservationCheck_compute_on_startup ocn_grid="IcosXISC30E3r7">.true.</config_AM_conservationCheck_compute_on_startup>
-<config_AM_conservationCheck_compute_on_startup ocn_grid="FRISwISC08to60E3r1">.true.</config_AM_conservationCheck_compute_on_startup>
-<config_AM_conservationCheck_compute_on_startup ocn_grid="FRISwISC04to60E3r1">.true.</config_AM_conservationCheck_compute_on_startup>
-<config_AM_conservationCheck_compute_on_startup ocn_grid="FRISwISC02to60E3r1">.true.</config_AM_conservationCheck_compute_on_startup>
-<config_AM_conservationCheck_compute_on_startup ocn_grid="FRISwISC01to60E3r1">.true.</config_AM_conservationCheck_compute_on_startup>
-<config_AM_conservationCheck_compute_on_startup ocn_grid="RRSwISC6to18E3r5">.true.</config_AM_conservationCheck_compute_on_startup>
 <config_AM_conservationCheck_write_on_startup>.false.</config_AM_conservationCheck_write_on_startup>
-<config_AM_conservationCheck_write_on_startup ocn_grid="SOwISC12to60E2r4">.true.</config_AM_conservationCheck_write_on_startup>
-<config_AM_conservationCheck_write_on_startup ocn_grid="ECwISC30to60E2r1">.true.</config_AM_conservationCheck_write_on_startup>
-<config_AM_conservationCheck_write_on_startup ocn_grid="IcoswISC30E3r5">.true.</config_AM_conservationCheck_write_on_startup>
-<config_AM_conservationCheck_write_on_startup ocn_grid="IcosXISC30E3r7">.true.</config_AM_conservationCheck_write_on_startup>
-<config_AM_conservationCheck_write_on_startup ocn_grid="FRISwISC08to60E3r1">.true.</config_AM_conservationCheck_write_on_startup>
-<config_AM_conservationCheck_write_on_startup ocn_grid="FRISwISC04to60E3r1">.true.</config_AM_conservationCheck_write_on_startup>
-<config_AM_conservationCheck_write_on_startup ocn_grid="FRISwISC02to60E3r1">.true.</config_AM_conservationCheck_write_on_startup>
-<config_AM_conservationCheck_write_on_startup ocn_grid="FRISwISC01to60E3r1">.true.</config_AM_conservationCheck_write_on_startup>
-<config_AM_conservationCheck_write_on_startup ocn_grid="RRSwISC6to18E3r5">.true.</config_AM_conservationCheck_write_on_startup>
 <config_AM_conservationCheck_write_to_logfile>.true.</config_AM_conservationCheck_write_to_logfile>
 <config_AM_conservationCheck_restart_stream>'conservationCheckRestart'</config_AM_conservationCheck_restart_stream>
 
--- a/components/mpas-ocean/cime_config/buildnml
+++ b/components/mpas-ocean/cime_config/buildnml
@@ -1156,7 +1156,7 @@ def buildnml(case, caseroot, compname):
             lines.append('        filename_interval="00-01-00_00:00:00"')
             lines.append('        reference_time="01-01-01_00:00:00"')
             lines.append('        output_interval="00-00-01_00:00:00"')
-            lines.append('        clobber_mode="truncate"')
+            lines.append('        clobber_mode="append"')
             lines.append('        packages="conservationCheckAMPKG">')
             lines.append('')
             lines.append('<var name="xtime"/>')
--- a/components/mpas-ocean/cime_config/testdefs/testmods_dirs/mpaso/conservation_check/README
+++ b/components/mpas-ocean/cime_config/testdefs/testmods_dirs/mpaso/conservation_check/README
@@ -4,5 +4,7 @@ wasintroduced in MPAS-Ocean PR #4521 and has been made a stealth feature in
 
     config_AM_conservationCheck_enable = .true.
 
-Then, it enables the use of MPAS-Ocean history files in testing by running
-a sed command to modify env_archive.xml.
+However, it shoudl be noted that MPAS-Ocean history files are not currently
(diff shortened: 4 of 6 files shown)


components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
```diff
@@ -1231,38 +1231,11 @@
...
-<config_AM_conservationCheck_enable ocn_grid="SOwISC12to60E2r4">.true.</config_AM_conservationCheck_enable>
-<config_AM_conservationCheck_enable ocn_grid="ECwISC30to60E2r1">.true.</config_AM_conservationCheck_enable>
-<config_AM_conservationCheck_enable ocn_grid="IcoswISC30E3r5">.true.</config_AM_conservationCheck_enable>
-<config_AM_conservationCheck_enable ocn_grid="IcosXISC30E3r7">.true.</config_AM_conservationCheck_enable>
-<config_AM_conservationCheck_enable ocn_grid="FRISwISC08to60E3r1">.true.</config_AM_conservationCheck_enable>
-<config_AM_conservationCheck_enable ocn_grid="FRISwISC04to60E3r1">.true.</config_AM_conservationCheck_enable>
-<config_AM_conservationCheck_enable ocn_grid="FRISwISC02to60E3r1">.true.</config_AM_conservationCheck_enable>
-<config_AM_conservationCheck_enable ocn_grid="FRISwISC01to60E3r1">.true.</config_AM_conservationCheck_enable>
-<config_AM_conservationCheck_enable ocn_grid="RRSwISC6to18E3r5">.true.</config_AM_conservationCheck_enable>
+<config_AM_conservationCheck_enable>.true.</config_AM_conservationCheck_enable>
```
xylar, line 1234: @mark-petersen, we may want this turned off for QU240, etc.   For light-weight testing.

components/mpas-seaice/bld/namelist_files/namelist_defaults_mpassi.xml
```diff
@@ -512,7 +512,7 @@
 <config_AM_ridgingDiagnostics_write_on_startup>true</config_AM_ridgingDiagnostics_write_on_startup>
 
 <!-- AM_conservationCheck -->
-<config_AM_conservationCheck_enable>false</config_AM_conservationCheck_enable>
+<config_AM_conservationCheck_enable>true</config_AM_conservationCheck_enable>
```
xylar, line 515: This should be a separate PR, I think.

jonbob: Thanks @xylar -- I agree, especially since mpas-seaice is also used for F-cases where its performance is already an issue

components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
xylar: Let's explicitly set this to `.false.` for:
* `oQU240`
* `oQU480`
* `oQU240wLI`
* `oQU120`

These meshes are used for testing:
https://github.com/E3SM-Project/E3SM/blob/master/cime_config/tests.py
and we don't want to needlessly increase test burden.

proteanplanet:
Conservation checks in both the sea ice nor ocean models should not be set as a default. It was actually a mistake that they were left on in the ocean for V3 production runs - they should have been switched off.  They add significant computational expense - up to 30% for the sea ice model alone.  Please remove the lines in [components/mpas-seaice/bld/namelist_files/namelist_defaults_mpassi.xml](https://github.com/E3SM-Project/E3SM/pull/6643/files/29ff9d05bc701c5ff4c52d9f32dfd5da0724acf9#diff-d9c675d17b8538db7faf430dada4c4d0635a845c4a04dc578fd082e7a7240a84) from this PR.   Also please consider removing this as default behavior for the ocean.  Conservation checks should be used for development purposes, but once we know models are conserving after checks run for PRs, the analysis members should be kept off.

xylar:
@mark-petersen, in light of @proteanplanet's comment above, which makes sense to me, can you actually turn the conservation analysis member off by default for all meshes in this PR.  I think we don't want to get distracted form the main objective here, which was to fix lack of restart capability in the conservation analysis member.

rljacob:
Could we add a restart test that does turn on the conservation analysis?

xylar:
@rljacob, that's an excellent idea but will CIME notice whether the output is BFB here?  I seem to recall that ERS tests don't include output from analysis, but maybe I'm misremembering.

jonbob:
@rljacob -- we can do that, though I think mpas analysis member output is ignored in the tests

xylar:
@mark-petersen, are you comfortable adding a test with the conservation check on or would you like me to do it?

mark-petersen:
I made conservation checks default off for both sea ice and ocean, per comment by @proteanplanet above. I'm checking the restart capability now.

components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
```diff
@@ -1232,37 +1232,10 @@
...
-<config_AM_conservationCheck_write_on_startup ocn_grid="SOwISC12to60E2r4">.true.</config_AM_conservationCheck_write_on_startup>
-<config_AM_conservationCheck_write_on_startup ocn_grid="ECwISC30to60E2r1">.true.</config_AM_conservationCheck_write_on_startup>
-<config_AM_conservationCheck_write_on_startup ocn_grid="IcoswISC30E3r5">.true.</config_AM_conservationCheck_write_on_startup>
-<config_AM_conservationCheck_write_on_startup ocn_grid="IcosXISC30E3r7">.true.</config_AM_conservationCheck_write_on_startup>
-<config_AM_conservationCheck_write_on_startup ocn_grid="FRISwISC08to60E3r1">.true.</config_AM_conservationCheck_write_on_startup>
-<config_AM_conservationCheck_write_on_startup ocn_grid="FRISwISC04to60E3r1">.true.</config_AM_conservationCheck_write_on_startup>
-<config_AM_conservationCheck_write_on_startup ocn_grid="FRISwISC02to60E3r1">.true.</config_AM_conservationCheck_write_on_startup>
-<config_AM_conservationCheck_write_on_startup ocn_grid="FRISwISC01to60E3r1">.true.</config_AM_conservationCheck_write_on_startup>
-<config_AM_conservationCheck_write_on_startup ocn_grid="RRSwISC6to18E3r5">.true.</config_AM_conservationCheck_write_on_startup>
```
xylar, line 1265: I agree that we want all these changes based on my comment https://github.com/E3SM-Project/E3SM/issues/6642#issuecomment-2373411357

components/mpas-ocean/cime_config/buildnml
```diff
@@ -1229,9 +1229,9 @@ def buildnml(case, caseroot, compname):
             lines.append('        filename_interval="output_interval"')
             lines.append('        clobber_mode="truncate"')
             lines.append('        packages="conservationCheckAMPKG"')
-            lines.append('        input_interval="none"')
+            lines.append('        input_interval="initial_only"')
             lines.append('        reference_time="01-01-01_00:00:00"')
-            lines.append('        output_interval="none">')
+            lines.append('        output_interval="stream:restart:output_interval">')
```
xylar, line 1234: Let's revert these changes.  I didn't understand that restarts are not needed because we are purposefully starting the average over at every output interval (i.e. accumulating only for 1 day).

Instead, let's change to `clobber_mode="append"` above in `conservationCheckOutput`.

xylar:
@mark-petersen, I've got a version of your branch with my suggested modifications here:
https://github.com/xylar/E3SM/commits/mark-petersen/mpas/correct-conservation-check/
It includes a new stealth test.  I have included MPAS-Ocean output files in that test by using `sed` to modify the `env_archive.xml` file.  Not sure that's kosher but I couldn't find a better way.

mark-petersen:
Updated using @xylar corrections based on previous comment and [comment on issue page](https://github.com/E3SM-Project/E3SM/issues/6642#issuecomment-2373411357). Thanks! Retesting now...

components/mpas-ocean/cime_config/testdefs/testmods_dirs/mpaso/conservation_check/shell_commands
```diff
@@ -0,0 +1,2 @@
+# include mpas-ocean outputs in testing
+sed -i 's#compclass="ocn" exclude_testing="true"#compclass="ocn" exclude_testing="false"#g' env_archive.xml
```
xylar, line 2: @jonbob and @rljacob, let me know what you think if this approach.  Happy to hear of other suggestions for how to not exclude MPAS-Ocean files from this particular test.

mark-petersen:
Updated the description.

cime_config/tests.py
```diff
@@ -266,6 +266,7 @@
             "SMS_D_Ld1.T62_oQU240wLI.GMPAS-IAF-PISMF.mpaso-impl_top_drag",
             "SMS_D_Ld1.T62_oQU240.GMPAS-IAF.mpaso-harmonic_mean_drag",
             "SMS_D_Ld1.T62_oQU240.GMPAS-IAF.mpaso-upwind_advection",
+            "ERS_D.T62_oQU240.GMPAS-IAF.mpaso-conservation_check",
             )
```
jonbob, line 270: could this be ERS_Ld5_D instead? Most of the stealth tests are short because they get run on the cloud

xylar: Works for me.  @mark-petersen, would you make that change?

components/mpas-ocean/cime_config/testdefs/testmods_dirs/mpaso/conservation_check/shell_commands
rljacob: Since its an xml file, can xmlchange do that?   If not, sed is fine.

xylar: I couldn't figure out how.  I think `xmlchange` can only be used to modify specific fields that have an `id` and `value`, and this is not one of those.

mark-petersen:
I can now confirm the expected behavior from this PR.

I ran QU240 with `config_am_conservationcheck_enable = .true.` for all cases, and everything standard from the repo for four cases here `/lcrc/group/e3sm/ac.mpetersen/scratch/chrys/`

1. master, 2 months `240924i.ConsCheckMasterFull.T62_oQU240.GMPAS-IAF.chr/`
2. master 1mo+restart+1 mo `240924j.ConsCheckMasterRestart.T62_oQU240.GMPAS-IAF.chr/`
3. pr 2 months `240924k.ConsCheckPrFull.T62_oQU240.GMPAS-IAF.chr/`
4. pr 1mo+restart+1 mo `240924l.ConsCheckPrRestart.T62_oQU240.GMPAS-IAF.chr/`

writing restarts every month. For reference, I used

```
./create_newcase -case $CASE_ROOT/$E3SM_CASE   -compiler gnu   -mach chrysalis   -project e3sm \
   -compset GMPAS-IAF -res T62_oQU240 --walltime 00:30:00
```

In case 2, the first entry differes from case 1 by being either zero or differing:
```
$ ncdump -v accumulatedRainFlux  *mpaso.hist.am.conservationCheck*01-02-01.nc |tail
 accumulatedRainFlux = 12140753948.2218, 12114183191.7286, 12117918123.2804,  (case 1)
 accumulatedRainFlux = 12129038639.0062, 12114183191.7286, 12117918123.2804, (case 2)
...

$ ncdump -v accumulatedRiverRunoffTemperatureFlux *mpaso.hist.am.conservationCheck*01-02-01.nc | tail
 accumulatedRiverRunoffTemperatureFlux = 18589454.7081383, 18605091.7352952, (case 1)
 accumulatedRiverRunoffTemperatureFlux = 0, 18605091.7352952, (case 2)
```
Running from this PR, case 3 and case 4 have identical output, identical to case 1 above.

xylar (approved):
I'm approving based on @mark-petersen's testing and my own.

components/mpas-ocean/cime_config/testdefs/testmods_dirs/mpaso/conservation_check/shell_commands
jonbob: I slacked Jim to ask if he knows...

jonbob:
I've been testing with the new testdef and don't believe it's working correctly -- first, it doesn't seem to change "exclude_testing" for mpaso in env_archive.xml. However, even if I change that by hand, it still does not run cprnc on the conservation am file -- it might take some work to get testing to recognize that file

xylar:
@jonbob, that's odd.  I think it changed the file for me.  In:
```
/lcrc/group/e3sm/ac.xylar/scratch/chrys/ERS_D.T62_oQU240.GMPAS-IAF.chrysalis_intel.mpaso-conservation_check.20240925_060737_57gegd
```
I'm seeing:
```
    <comp_archive_spec compname="mpaso" compclass="ocn" exclude_testing="false">
      <rest_file_extension>rst</rest_file_extension>
...
```
as expected.

As for the latter point.  I wasn't sure how to find out what cprnc is checking.  I guess we'll need @jgfouca's help to figure out another way.

jonbob:
@xylar -- that's very odd... Anyway, Jim didn't see a way to do it with xmlchange either.

xylar:
@jonbob, I'll try it again.  In the meantime, can you clarify what you are doing to see what files cprnc is looking at?

xylar:
I reset to the current state of this branch and I'm also seeing that the `sed` changes aren't happening -- so odd!

xylar:
My fault.  I accidentally had e3sm-unified loaded when I launched the test.  It must have a better `sed` that supports `-i` but the default `sed` on Chrysalis doesn't (!!!).

I'll have to do this more manually.

jonbob:
@xylar -- there are files that get made for tests -- TestStatus and TestStatus.log. If I grep TestStatus.log for cprnc, I only see that being run on the cpl.hi files

xylar:
Thanks @jonbob.  I'll try to debug some more tomorrow.

xylar:
On second thought, I'm beginning to feel like getting an ERS test to work with MPAS-Ocean history output is too big a project to be in the scope of this PR.

@jonbob and @rljacob, I think the only plausible fallback is to take out the stealth test.  What about you?

jonbob:
@xylar -- I was going to recommend the same thing. I think it's more important to fix the functionality, and we can add a test once we get it working

rljacob:
Don't try to do the sed magic and force the comparison but leave on the test that enables the conservation analysis members.

xylar:
@rljacob, okay, good suggestion.  I have noted in the README for the test that it does not currently include MPAS-Ocean history files.  I have given a manual change that folks could make to `cime_config/config_archive.xml` if they want to include these files.

components/mpas-ocean/cime_config/testdefs/testmods_dirs/mpaso/conservation_check/README
```diff
@@ -0,0 +1,10 @@
+This testdef is used to test the conservation check analysis member, which
+wasintroduced in MPAS-Ocean PR #4521 and has been made a stealth feature in
+#6643. This test turns on the consevation check analysis member by setting:
+
+    config_AM_conservationCheck_enable = .true.
+
+However, it shoudl be noted that MPAS-Ocean history files are not currently
+included in E3SM testing so non-BFB results will not be detected unless one
+manually changes to 'compname="mpaso" exclude_testing="false"' in the file
+cime_config/config_archive.xml.
```
jonbob, line 10: one quick comment -- even changing exclude_testing won't check any of the mpaso history files. Plus there's a minor misspelling: "shoudl"

xylar: @jonbob, okay, I understood that it would but I must be mistaken.  I have removed this text and fixed the typo.

xylar: Okay, clearly a bigger can of worms that remains beyond the scope of this PR.  Thanks for being more thorough than I was and for keeping things honest.

jonbob:
@xylar -- I tested this in the shell_commands and it does give us what we want:
````
sed -i 's/\(compname="mpaso" exclude_testing="true"\)/compname="mpaso"/' env_archive.xml
sed -i '/\(compname="mpaso"\)/,/<hist_file_extension>/ s/<hist_file_extension>hist/<hist_file_extension>hist.am.conservationCheck\\..*\\.nc$/' env_archive.xml
````
I'm not convinced we need to make the test completely functional for this PR, but I was curious about how difficult it would be...

darincomeau (approved):
Approve the changes to the conservation check based on other's testing and visual inspection.
I'll defer to others on the added testing.

jonbob:
Passes:
* ERP_Ld3.ne30pg2_r05_IcoswISC30E3r5.WCYCL1850.chrysalis_intel.allactive-pioroot1 (with expected NML DIFF)

merged to next

jonbob:

PR #6646, opened 2024-09-25
Homme(xx): Flush the log before calling MPI_Abort in Homme's abortmp.

ambrad:
This is an attempt to assure atm.log (EAM) and homme_atm.log (EAMxx) get full output when the dycore calls MPI_Abort.

[BFB]

### DIFF
--- a/components/homme/src/share/parallel_mod.F90
+++ b/components/homme/src/share/parallel_mod.F90
@@ -274,6 +274,7 @@ contains
 #endif
 #endif
     character*(*) string
+    call flush(iulog)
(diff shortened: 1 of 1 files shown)


ambrad:
@rljacob when you have a chance, would you look at this and let me know if you see any reason why this is not a reasonable thing to do? Thanks.

rljacob (approved):

rljacob:
Very reasonable.

ambrad:
Passes e3sm_developer and homme_integration on Chrysalis. I'm going to merge next week.

PR #6649, opened 2024-09-27
Fix initialization of doubles in VerticalProfileMod

peterdschwartz:
Variables that are `r8` were initialized as single-precision, potentially causing inconsistent failures with the sums not adding to `1.0_r8`.

Also, fixed a syntax error in CH4Mod for `spval`.

[BFB]

### DIFF
--- a/components/elm/src/biogeochem/CH4Mod.F90
+++ b/components/elm/src/biogeochem/CH4Mod.F90
@@ -223,7 +223,7 @@ contains
     ! Allocate module variables and data structures
     !
     ! !USES:
-    use shr_infnan_mod, only: spval => shr_infnan_nan, assignment(=)
+    use shr_infnan_mod, only: nan  => shr_infnan_nan, assignment(=)
     use elm_varpar    , only: nlevgrnd
     !
     ! !ARGUMENTS:
--- a/components/elm/src/biogeochem/VerticalProfileMod.F90
+++ b/components/elm/src/biogeochem/VerticalProfileMod.F90
@@ -336,7 +336,15 @@ contains
          end do
          if ( ( abs(froot_prof_sum - 1._r8) > delta ) .or.  ( abs(croot_prof_sum - 1._r8) > delta ) .or. &
               ( abs(stem_prof_sum - 1._r8) > delta ) .or.  ( abs(leaf_prof_sum - 1._r8) > delta ) ) then
+            c = veg_pp%column(p)
             write(iulog, *) 'profile sums: ', froot_prof_sum, croot_prof_sum, leaf_prof_sum, stem_prof_sum
+            write(iulog, *) 'c: ',c
+            write(iulog, *) 'altmax_lastyear_indx: ', altmax_lastyear_indx(c)
+            write(iulog, *) 'cinput_rootfr: ', col_cinput_rootfr(c,:)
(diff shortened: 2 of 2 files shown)


peterdschwartz:
I can't verifiy that this indeed fixes the bug on mappy. Someone with access to mappy would have to test it a few times due to the intermittent nature of the failure.  

But this code is bugged, so it makes sense to me merge it to see if it's fixed using CDash.

rljacob (approved):

rljacob:
@peterdschwartz you can merge this to next today.

peterdschwartz:
merged to next.  

Tested [BFB] with intel and gnu compilers on pm-cpu

peterdschwartz:
@rljacob I noticed that [ERS.r05_r05.ICNPRDCTCBC.mappy_gnu.elm-cbudget](https://my.cdash.org/test/209417156) is also failing occasionally with same VerticalProfileMod sum error message.

Since that test still failed overnight on next with this PR, I re-checked and noticed I missed a couple of if-branches that have single-precision (as below).  I made a new commit and will re-merge today.
```
else 
   froot_prof(p,1) = 1./dzsoi_decomp(1)
   croot_prof(p,1) = 1./dzsoi_decomp(1)
   leaf_prof(p,1) = 1./dzsoi_decomp(1)
   stem_prof(p,1) = 1./dzsoi_decomp(1) 
end if 
```

Test error message:
```
 profile sums:    1.0000000000000000        1.0000000000000000        1.0000000000000000       0.99923994737367017     
 ENDRUN: ERROR: sum-1 > deltaERROR in /home/e3sm-jenkins/jenkins-ws/workspace/mappy_e3sm_next/E3SM/components/elm/src/biogeochem/VerticalProfileMod.F90 at line 340                                                                                                                                                                                                                                                                                                                                                                                          
 ERROR: Unknown error submitted to shr_abort_abort.
```

rljacob:
Should the 3.0.1 tag wait for this?

peterdschwartz:
@rljacob I'm not sure. The code is more correct this way, but practically, it seems to only manifest on one machine/compiler and we can't guarantee that this will fix that specifically.

peterdschwartz:
re-merged to next

rljacob:
Was this supposed to help with ERS.r05_r05.IELM.mappy_gnu.elm-V2_ELM_MOSART_features ?  It flipped from DIFF yesterday to FAIL today.

rljacob:
@jgfouca may be able to help speed up the development since he could run a test on your branch on mappy. 

Note that I reset next so this PR is no longer on it.

jgfouca:
@rljacob , @peterdschwartz , I am happy to kick off some runs on mappy. I assume I will need to checkout this branch. How is this for a test:

```
% for i in $(seq 10); do
  ./create_test ERS.r05_r05.IELM.mappy_gnu.elm-V2_ELM_MOSART_features --wait
done
```

I assume we are looking for FAILs, not DIFFs, so no need for baseline comparison.

rljacob:
@peterdschwartz want do you want to do with this PR?  It still fixes things that should be fixed even if it doesn't fix #6650

peterdschwartz:
Right, I will push one more commit that just adds useful information to the error message and start merging it.

rljacob:
@peterdschwartz you can merge this to master.

peterdschwartz:
merged to master

PR #6671, opened 2024-10-07
Fix import of moved method in MVKO test

mkstratos:
`CIME.utils.append_testlog` was moved in https://github.com/ESMCI/cime/pull/4660
This updates the MVKO test to use that method (other CIME SystemTests were updated in that PR)

[BFB]

### DIFF
--- a/components/mpas-ocean/cime_config/SystemTests/mvko.py
+++ b/components/mpas-ocean/cime_config/SystemTests/mvko.py
@@ -18,6 +18,7 @@ import numpy as np
 import netCDF4 as nc
 
 import CIME.test_status
+from CIME.status import append_testlog
 import CIME.utils
 from CIME.SystemTests.system_tests_common import SystemTestsCommon
 from CIME.case.case_setup import case_setup
@@ -394,4 +395,4 @@ class MVKO(SystemTestsCommon):
                 f"        {viewing}"
             )
 
-            CIME.utils.append_testlog(comments, self._orig_caseroot)
+            append_testlog(comments, self._orig_caseroot)


rljacob (approved):

rljacob:
The append_testlog error went away but now the MVKO test is reporting a fail:


```
 ---------------------------------------------------
2024-10-08 01:34:41: BASELINE FAIL for test 'JNextNbfb20241008_003411'.
    Test status: fail; Variables analyzed: 5; Rejecting: 5; Critical value: 0; Ensembles: statistically different
    EVV results can be viewed at:
        https://web.lcrc.anl.gov/public/e3sm/e3smtest/evv/MVKO_PS.T62_oQU240.GMPAS-NYF.chrysalis_intel.C.JNextNbfb20241008_003411/index.html
```

rljacob:
The baseline its comparing against is in /lcrc/group/e3sm2/baselines/chrys/intel/master/MVKO_PS.T62_oQU240.GMPAS-NYF.chrysalis_intel and its dated May 8.  But there should be no changes in answers for a WCYCLE config of MPAS since v3.0.0.   @jonbob do you know if anything might have changed in GMPAS-NYF since May 8?

jonbob:
@rljacob -- I'll look and see if we changed anything specifically for NYF or the oQU240 mesh, but nothing is coming to mind

jonbob:
@rljacob -- we did change the mapping files for T62 because there had been an incorrect T62 scrip file (PR #6459). It was non-BFB but T62 is not used for the coupled model, so that was deemed acceptable

jonbob:
@rljacob - I guess it could be climate-changing? But the oQU240 is a grid we only use for testing and it's not considered scientifically validated. Maybe the only real way to find out is to run this test with and without the PR that changed the mapping and domain files and see. Please just yell if that's something you want me to take on

mkstratos:
The different map files seem to be the cause for the climate changing differences. This test: https://web.lcrc.anl.gov/public/e3sm/ac.mkelleher/evv/MVKO_PS.T62_oQU240.GMPAS-NYF.chrysalis_intel.C.20241008_121250_blo0rd run with current master and the old map files passes.

rljacob:
Ok I'll bless it.  Its better if oQU240 is more sensitive then less sensitive.

PR #6672, opened 2024-10-08
fix variables in data subglacial runoff

irenavankova:
This fixes a typo in PR https://github.com/E3SM-Project/E3SM/pull/6508, where config_flux_attenuation_coefficient_runoff was mistakenly used in kpp loop instead of config_flux_attenuation_coefficient_subglacial_runoff.

These variables are only used for data subglacial runoff, so this does not impact current tests.

[BFB]


xylar (approved):
Yes, I agree by inspection that this looks correct and the old value was copied by mistake.

PR #6676, opened 2024-10-10
Update pelayout on pm-cpu for ne4 scream tests

ndkeen:
For ne4 cases, use only 96 tasks as scream requires no more MPI's than number of elements.
`SMS.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.pm-cpu_intel`

Unrelated: Rename a machinefile to reflect machine name for gcp12 builds with scream.
For this change it fixes https://github.com/E3SM-Project/scream/issues/3036 (at least the build issue)

[bfb]


rljacob:
Why would EAMxx complain about this and not EAM?

ndkeen:
I have asked that as well.  Must be different paths, but both repos have same:

`homme/src/share/prim_driver_base.F90`

```
    ! we want to exit elegantly when we are using too many processors.                                                                     
    if (nelem < par%nprocs) then
       call abortmp('Error: too many MPI tasks. set dyn_npes <= nelem')
    end if
```

It was so elegant. The most elegant.

rljacob:
This changed the layout for several e3sm_integration tests:
ERP_Ln9.ne4pg2_oQU480.WCYCL20TRNS-MMF1.pm-cpu_intel.allactive-mmf_fixed_subcycle
ERS.ne4pg2_oQU480.WCYCL1850NS.pm-cpu_intel
ERS_Vmoab.ne4pg2_oQU480.WCYCL1850NS.pm-cpu_intel
NCK.ne4pg2_oQU480.WCYCL1850NS.pm-cpu_intel

If you want to keep this, the Vmoab test needs to be blessed.

ndkeen:
Makes sense.  I could try to make another entry in pelayout XML to match ne4 *with* SCREAM, but it might be best to just run all ne4 cases with 96 MPI's on pm-cpu.  What do we think?  Before this change, ne4 tests would have used a full node, which is 128 MPI's on pm-cpu.

rljacob:
Its fine just making the change for all cases.

Issue #5753, opened 2023-06-09
NERSC performance archiving for multiple projects


sarats:
https://github.com/E3SM-Project/E3SM/blob/d1388ddf71ced4a11ff364bf1eac80f2bad70666/cime_config/machines/config_machines.xml#L155

We can use a wildcard like '*' but that probably would capture a lot of experiments from non-E3SM project members.
Otherwise, we need to update the list of projects with any others that we wish to archive.

cc @bishtgautam

sarats:
Will add m4259 that is used by ImPACTS SciDAC project
cc @vanroekel

darincomeau:
m1199 for the E3SM-Arctic simulations under the HiLAT-RASM project would be great to have.
cc: @milenaveneziani

sarats:
Other projects that were requested.

m3312 cc: @whannah1 PI & Topic?

m4048 From @beharrop 
> Wei Cheng is the PI.  It's an RGMA project looking at ECS sensitivity to ocean heat transport and cloud feedbacks using the slab ocean configuration of E3SM with a cloud-locking capability

sarats:
m3312 cc: @whannah1 PI & topic?

@ndkeen Do you know the topics and PIs for m3411 and m3412 (existing)?

sarats:
One aspect to consider is the performance archive location and ability for users in some of these projects to write to that location.

Currently, it's set to 
`<SAVE_TIMING_DIR>/global/cfs/cdirs/e3sm</SAVE_TIMING_DIR>`

We ought to consider a world-writable location in that case.

PR #6678, opened 2024-10-10
Add project id's to list so that more E3SM jobs at NERSC can have provenance data saved

ndkeen:
There are now quite a few E3SM jobs running at NERSC and we are *not* saving info.
This change adds several project ID's that have recently run larger-node/longer-time E3SM jobs.

[bfb]


sarats:
As I mentioned in https://github.com/E3SM-Project/E3SM/issues/5753, I would like to assess the impact of inclusion of any non-core ERCAP project. 

For that purpose, I would like to get the name of the project, PI contact info and planned runs. If there are plans for running large number of small or single node runs or small-scale ensembles, I don't wish to archive them in PACE and would like them to provide assurance and understand disabling archiving for such runs (`./xmlchange SAVE_TIMING=FALSE`)

Either this information can be listed here or on Confluence.

One additional aspect to consider is the performance archive location and ability for users in some of these projects to write to that location.

The limiting constraint for archiving any and all runs is the limited storage capacity on the PACE server (for database and raw data).

ndkeen:
OK, so you are going to take this Sarat? 
I think it would be fine to add these, and remove any if they are too frequent.
It's true there would be a problem if they could not write to the dir.

rljacob:
Should we just add the ones mentioned in #5753 now and other's later after a conversation with the PI?

@ndkeen do you have PI contact info for all the projects you're adding?

sarats:
FYI, I discussed this topic today with the EXEC during our call and they noted that this is a low-priority.

PR #6684, opened 2024-10-14
Add explicit Core/24.00 module to Frontier

grnydawn:
Explicitly load the Core/24.00 module to prevent the new default Core/24.07 module from loading on Frontier. NOTE: https://github.com/E3SM-Project/scream/pull/3029

Ran a subset of e3sm_developer tests, which showed similar results to previous tests.

BFB (no reference on Frontier).

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -1033,6 +1033,7 @@
       <cmd_path lang="python">/usr/share/lmod/lmod/libexec/lmod python</cmd_path>
       <modules compiler="crayclang.*">
         <command name="reset"></command>
+        <command name="load">Core/24.00</command>
         <command name="switch">PrgEnv-cray PrgEnv-cray/8.3.3</command>
         <command name="switch">cce cce/15.0.1</command>
         <!-- craype module to address tcmalloc runtime errors at startup -->
@@ -1045,6 +1046,7 @@
       </modules>
       <modules compiler="amdclang.*">
         <command name="reset"></command>
+        <command name="load">Core/24.00</command>
         <command name="switch">PrgEnv-cray PrgEnv-amd/8.3.3</command>
         <command name="switch">amd amd/5.4.0</command>
       </modules>
@@ -1053,6 +1055,7 @@
       </modules>
       <modules compiler="gnu.*">
         <command name="reset"></command>
+        <command name="load">Core/24.00</command>
         <command name="switch">PrgEnv-cray PrgEnv-gnu/8.3.3</command>
         <command name="switch">gcc gcc/12.2.0</command>
       </modules>


grnydawn:
@rljacob , @jgfouca , The new default Core/24.07 module on Frontier causes issues due to the absence of several utility modules, including Subversion, Git, Darshan-runtime, and CMake, which ultimately breaks the E3SM build on Frontier. @trey-ornl also warned that the new CMake might have difficulty with the older style of handling Cray and AMD compilers. Following @trey-ornl's recommendation, I have explicitly added Core/24.00 in config_machines.xml.

rljacob (approved):

rljacob:
Is it conflicting with these lines?
```
        <command name="load">subversion/1.14.1</command>
        <command name="load">git/2.36.1</command>
       <command name="load">cmake/3.21.3</command>
```
I don't think we need those specific versions of git and subversion so just take them out and let Core/24.07 load what it wants.

Not sure what to do about cmake.  What version is Core/24.07  trying to load?   Also where do we load darshan?  Its not an explicit load command.

trey-ornl:
> Is it conflicting with these lines?
> 
> ```
>         <command name="load">subversion/1.14.1</command>
>         <command name="load">git/2.36.1</command>
>        <command name="load">cmake/3.21.3</command>
> ```
> 
> I don't think we need those specific versions of git and subversion so just take them out and let Core/24.07 load what it wants.
> 
> Not sure what to do about cmake. What version is Core/24.07 trying to load? Also where do we load darshan? Its not an explicit load command.

`darshan-runtime` gets loaded by default.

The only CMake available with `Core/24.07` is `cmake/3.27.9`, and this broke the `crayclang-scream` build for me in the "scream" repo (but not the shiny new `craygnuamdgpu` build).

jgfouca (approved):

grnydawn:
I tried building E3SM with the new Core/24.07 module after removing the versions from those load commands, and it seemingly solved the build issue without any error message. I have not yet completed the tests up to running a case, but I believe this issue likely affects the build phase, not the runtime.

I will test more cases, and if all tests show no this type of issue, I plan to drop this pull request and create a new one without the versions for the util modules.

trey-ornl:
> I tried building E3SM with the new Core/24.07 module after removing the versions from those load commands, and it seemingly solved the build issue without any error message. I have not yet completed the tests up to running a case, but I believe this issue likely affects the build phase, not the runtime.
> 
> I will test more cases, and if all tests show no this type of issue, I plan to drop this pull request and create a new one without the versions for the util modules.

The issue with `Core/24.07` may be specific to `crayclang-scream`, which keeps old versions of the Cray software to match what the decadal run started with.

grnydawn:
close this PR because https://github.com/E3SM-Project/E3SM/pull/6689 is merged into master

Issue #6655, opened 2024-10-01
On muller-cpu, with an update to slighshot software, we are seeing a stall/hang in init


ndkeen:
With software updates to slingshot, not yet on pm-cpu, but could be soon, I've been testing a variety of things on internal test machine.  We have a work-around for now that seems to not show issue.
Just wanted to make issue to record some of the info I have learned.

It happens with E3SM and scream. Does not occur every job, but happens more frequently as MPI ranks increased.
So far, fewest nodes I've seen the stall is 22 and the most I've seen it work is 256 nodes.
I've only experienced it with ne120 cases. None with ne30, but also have not run that may ne30.

When stalled, stack is the following on one rank:
```
#0  0x000014dbd06f9f22 in MPIR_Wait_impl () from /opt/cray/pe/lib64/libmpi_intel.so.12
#1  0x000014dbd129b926 in MPIC_Wait () from /opt/cray/pe/lib64/libmpi_intel.so.12
#2  0x000014dbd129fe29 in MPIC_Recv () from /opt/cray/pe/lib64/libmpi_intel.so.12
#3  0x000014dbd11cff68 in MPIR_Allreduce_intra_reduce_scatter_allgather () from /opt/cray/pe/lib64/libmpi_intel.so.12
#4  0x000014dbcfb72630 in MPIR_Allreduce_intra_auto () from /opt/cray/pe/lib64/libmpi_intel.so.12
#5  0x000014dbcfb726f1 in MPIR_Allreduce_impl () from /opt/cray/pe/lib64/libmpi_intel.so.12
#6  0x000014dbd141e1a7 in MPIR_CRAY_Allreduce () from /opt/cray/pe/lib64/libmpi_intel.so.12
#7  0x000014dbcfbbe27c in PMPI_Allreduce () from /opt/cray/pe/lib64/libmpi_intel.so.12
#8  0x000014dbd2243856 in pmpi_allreduce__ () from /opt/cray/pe/lib64/libmpifort_intel.so.12
#9  0x000000000160ec9b in dof_mod::setelemoffset (par=<error reading variable: Cannot access memory at address 0x0>, elem=<error reading variable: Location address is not set.>, globaluniquecolsp=0) at /mscratch/sd/n/ndk/repos/ms12-sep20/components/homme/src/share/dof_mod.F90:299
[...]

Issue #6516, opened 2024-07-19
Trying to update GNU compilers on pm-cpu, encounter hang/FPE with certain tests


ndkeen:
I've been trying to update several of the module versions on pm-cpu to "ideal" versions. Most test (with intel,gnu,nvidia) seem OK, but a few are still problematic and I wanted to save some notes here.  For GNU, wanting to update to 12.3 and NERSC calls the module `gcc-native/12.3`. Also updating these at same time (some are required as a package update) -- currently trying those in "ideal" below:

```
module                      current        machine defaults   ideal
gcc                         11.2.0         12.2.0             12.3 (gcc-native)
PrgEnv-gnu                  8.3.3          8.5.0              8.5.0
cray-libsci                 23.02.1.1      23.02.1.1          23.12.5
cray-mpich                  8.1.25         8.1.25             8.1.28
craype                      2.7.20         2.7.20             2.7.30
cray-hdf5-parallel          1.12.2.3       1.12.2.3           1.12.2.9
cray-netcdf-hdf5parallel    4.9.0.3        4.9.0.3            4.9.0.9
cray-parallel-netcdf        1.12.3.3       1.12.3.3           1.12.3.9
```

Running e3sm_integration, the only tests with issues are only DEBUG built tests and unfortunately they hang during init. I managed to find two files that I can alter compiler flags and get a FPE instead, considered an improvement, but may or may not be the same issue causing the hang. For the following 2 files, if I *add* `-O` (ie, disable `-O0`)

```
+  eam/src/dynamics/se/inidat.F90
+  eam/src/dynamics/se/dyn_comp.F90
```

I can get the following error with several tests:
```
 561: Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.
 561: 
 561: Backtrace for this error:
 561: #0  0x14c47a423372 in ???
 561: #1  0x14c47a422505 in ???
 561: #2  0x14c479853dbf in ???
 561: #3  0x1a35e91 in __edge_mod_base_MOD_edgevunpack_nlyr
 561:   at /global/cfs/cdirs/e3sm/ndk/repos/pr/ndk_mf_pm-cpu-update-compiler-versions/components/homme/src/share/edge_mod_base.F90:903
 561: #4  0x26be308 in __inidat_MOD_read_inidat
[...]

Issue #6451, opened 2024-05-29
Hang in `dp_coupling::d_p_coupling` with newer module versions and compilers (GNU version 12.3)


ndkeen:
Originally reporting issue with newer Intel compiler, but as of Sep7, 2024, no longer seeing the issue with intel and created PR to upgrade (https://github.com/E3SM-Project/E3SM/pull/6596), but still see issue with GNU.  I might close this issue and open a fresh one for GNU only, but for now, leaving text below as-is:

Trying to update module versions on pm-cpu, but I have hit a few issues. One with intel is that this test hangs in init.
`SMS_D_Ld1.ne30pg2_r05_IcoswISC30E3r5.WCYCLSSP370.muller-cpu_intel.allactive-wcprodssp`
I'm noting the hang in HOMME, but as I don't know root cause, it may not actually be issue there.
The test works with current intel version (`intel/2023.1.0`) and what I'd like to use is the new default for the machine (`intel/2023.2.0`)

We see this in cpl.log (to indicate still in init):
```
(seq_mct_drv) : Calling atm_init_mct phase 2
(component_init_cc:mct) : Initialize component atm
```

Looking at where the stack is on compute node:

```
#0  cxi_eq_peek_event (eq=0x22e12dc8) at /usr/include/cxi_prov_hw.h:1531
#1  cxip_ep_ctrl_eq_progress (ep_obj=0x22e25790, ctrl_evtq=0x22e12dc8, tx_evtq=true, ep_obj_locked=true) at prov/cxi/src/cxip_ctrl.c:318
#2  0x00001503828591dd in cxip_ep_progress (fid=<optimized out>) at prov/cxi/src/cxip_ep.c:186
#3  0x000015038285e969 in cxip_util_cq_progress (util_cq=0x22e15220) at prov/cxi/src/cxip_cq.c:112
#4  0x000015038283a301 in ofi_cq_readfrom (cq_fid=0x22e15220, buf=<optimized out>, count=8, src_addr=0x0) at prov/util/src/util_cq.c:232
#5  0x00001503860fa0f2 in MPIR_Wait_impl () from /opt/cray/pe/lib64/libmpi_intel.so.12
#6  0x0000150386c9b926 in MPIC_Wait () from /opt/cray/pe/lib64/libmpi_intel.so.12
#7  0x0000150386ca7685 in MPIC_Sendrecv () from /opt/cray/pe/lib64/libmpi_intel.so.12
[...]

PR #6687, opened 2024-10-15
Use `FI_MR_CACHE_MONITOR=kdreg2` for all nersc machines

ndkeen:
With new slighshot software (s2.2 h11.0.1), now installed on Perlmutter, there were some hangs in init for certain cases at higher node counts. Using environment variable `FI_MR_CACHE_MONITOR=kdreg2` avoids any issues so far.
`kdreg2` is another option for memory cache monitoring -- it is a Linux kernel module using open-source licensing.
It comes with HPE Slingshot host software distribution (optionally installed) and may one day be the default.

Regarding performance, it seems about the same. For one HR F-case at 256 nodes, using `kdreg2` was about 1% slower.

Fixes https://github.com/E3SM-Project/E3SM/issues/6655

I also found some older issues (some with lower node-count) that this fixes:
Fixes https://github.com/E3SM-Project/E3SM/issues/6516
Fixes https://github.com/E3SM-Project/E3SM/issues/6451
Fixes https://github.com/E3SM-Project/E3SM/issues/6521

[bfb]

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -278,6 +278,7 @@
       <env name="HDF5_USE_FILE_LOCKING">FALSE</env>
       <env name="PERL5LIB">/global/cfs/cdirs/e3sm/perl/lib/perl5-only-switch</env>
       <env name="FI_CXI_RX_MATCH_MODE">software</env>
+      <env name="FI_MR_CACHE_MONITOR">kdreg2</env>
       <env name="MPICH_COLL_SYNC">MPI_Bcast</env>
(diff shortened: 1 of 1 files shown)


ndkeen:
I'm now seeing it looks like this will fix #6516, which would be great as it will allow me to update the GCC compiler version.

ndkeen:
Fixes https://github.com/E3SM-Project/E3SM/issues/6451

ambrad (approved):

ndkeen:
Merged to next

rljacob (approved):

PR #6689, opened 2024-10-16
load Core/24.07 module and remove version info from util modules

grnydawn:
Explicitly load the Core/24.07 module on Frontier and removes version info from git, submodule, and zlib modules in config_machines.xml

Ran a subset of e3sm_developer tests, which showed similar results to previous tests.

BFB (no reference on Frontier).


grnydawn:
@jgfouca , @rljacob ,This PR replaces https://github.com/E3SM-Project/E3SM/pull/6684 in favor of using the new default Core/24.07 module on Frontier. Test results show essentially the same outcome as with the previous Core/24.00.

rljacob (approved):

jgfouca (approved):

PR #6690, opened 2024-10-16
Update HOMME Aurora machine cmake config

abagusetty:
0. Update Cmake machine config for Aurora
1. Cleanup certain warnings related to headers and Intel compilers
2. Linking Fortran/C++/SYCL with CXX linker (-fortlib) instead of Fortran language for SYCL builds. Since Fortran linking with SYCL is no longer supported (https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2024-2/fortlib.html)

### DIFF
--- a/components/homme/cmake/machineFiles/aurora-aot.cmake
+++ b/components/homme/cmake/machineFiles/aurora-aot.cmake
@@ -1,11 +1,9 @@
 #module restore
-#module load oneapi/eng-compiler/2022.12.30.005
-#module load intel_compute_runtime/release/agama-devel-627
-#module load spack cmake
+#module load spack-pe-base cmake
 #module list
 
 
-SET (SUNSPOT_MACHINE TRUE CACHE BOOL "")
+SET (AURORA_MACHINE TRUE CACHE BOOL "")
 
 SET(BUILD_HOMME_WITHOUT_PIOLIBRARY TRUE CACHE BOOL "")
 SET(HOMMEXX_MPI_ON_DEVICE FALSE CACHE BOOL "")
@@ -18,12 +16,12 @@ SET(USE_QUEUING FALSE CACHE BOOL "")
 
 #temp hack
 SET(HOMME_USE_KOKKOS TRUE CACHE BOOL "")
+SET(HOMME_USE_MKL TRUE CACHE BOOL "")
 
 SET(BUILD_HOMME_PREQX_KOKKOS TRUE CACHE BOOL "")
 SET(BUILD_HOMME_THETA_KOKKOS TRUE CACHE BOOL "")
 
-#set(KOKKOS_HOME "/home/onguba/kokkos-build/mar05-aot/install" CACHE STRING "")
-#set(E3SM_KOKKOS_PATH ${KOKKOS_HOME} CACHE STRING "")
+set(Kokkos_ROOT $ENV{KOKKOS_HOME} CACHE STRING "")
 
 SET(USE_TRILINOS OFF CACHE BOOL "")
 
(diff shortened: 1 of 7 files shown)


components/homme/src/share/cxx/ExecSpaceDefs.cpp
```diff
@@ -22,7 +22,7 @@
 #endif
 
 #ifdef KOKKOS_ENABLE_SYCL
-#include <CL/sycl.hpp>
+#include <sycl/sycl.hpp>
```
oksanaguba, line 25: could you explain this change please?

oksanaguba (approved):

components/homme/cmake/machineFiles/aurora-aot.cmake
```diff
@@ -57,8 +50,6 @@ set (ENABLE_HORIZ_OPENMP OFF CACHE BOOL "")
 
 set (HOMME_TESTING_PROFILE "dev" CACHE STRING "")
 
-set (USE_NUM_PROCS 4 CACHE STRING "")
-
-SET (USE_MPI_OPTIONS "--bind-to core" CACHE FILEPATH "")
-
+set (USE_NUM_PROCS 12 CACHE STRING "")
 
+SET (USE_MPI_OPTIONS "--pmi=pmix --cpu-bind list:0-7,104-111:8-15,112-119:16-23,120-127:24-31,128-135:32-39,136-143:40-47,144-151:52-59,156-163:60-67,164-171:68-75,172-179:76-83,180-187:84-91,188-195:92-99,196-203 gpu_tile_compact.sh" CACHE FILEPATH "")
```
oksanaguba, line 55: is tile script always in a user's path?

components/homme/src/share/cxx/ExecSpaceDefs.cpp
abagusetty: SYCL specification has long deprecated these headers. <CL/sycl.hpp> header was valid only in SYCL 1.2.1 but it has been deprecated for <sycl/sycl.hpp> in SYCL 2020 Standards. 

From the SYCL spec
```
For consistency, the programming API will only refer to the <sycl/sycl.hpp> header and the ::sycl namespace, but this should be considered synonymous with the SYCL 1.2.1 header and namespace.```

File: components/homme/cmake/machineFiles/aurora-aot.cmake (hunk 2 above)
Comment (abagusetty) [reply components/homme/cmake/machineFiles/aurora-aot.cmake:55]: Yes, the tile binding script is always in the SYSTEM path for the users

**Comment by rljacob** [comment]
To clarify, the changes in components/homme/cmake only impact the standalone build of HOMMEXX but the other changes could impact the EAMxx/SCREAM build?

**Comment by abagusetty** [comment]
> To clarify, the changes in components/homme/cmake only impact the standalone build of HOMMEXX but the other changes could impact the EAMxx/SCREAM build?

I guess @oksanaguba could fill-in more accurately, but changes outside the standalone cmake config to HOMME were to address deprecation warnings and so are cosmetic

**Comment by oksanaguba** [comment]
yes, the changes besides cmake will be picked up by scream builds, but as Abhishek says they are an improvement. SCREAM flags may need revision in a similar way (or not).

File: components/homme/src/share/cxx/vector/vector_pragmas.hpp [hunk 3]
```diff
@@ -7,7 +7,7 @@
 #ifndef HOMMEXX_VECTOR_PRAGMAS_HPP
 #define HOMMEXX_VECTOR_PRAGMAS_HPP
 
-#if defined(__INTEL_COMPILER)
+#if defined(__INTEL_COMPILER) || defined(__INTEL_CLANG_COMPILER) || defined(__INTEL_LLVM_COMPILER)
```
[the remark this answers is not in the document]
abagusetty: Intel compilers are identified as both clang and llvm based. Both the macros are defined by oneAPI compilers. 

The difference being the naming convention: If it is a compiler released by Intel as a product (both the macros are defined) or if a user builds the compiler themselves from upstream repo (then either or both could be defined)

abagusetty:
@oksanaguba The CI seems to have failed. Is that what is blocking at the moment

oksanaguba (approved):

oksanaguba:
@abagusetty we use a manual process to merge to "next" first and let cdash tests pass overnight https://my.cdash.org/index.php?project=E3SM . then the PR can go to "master". i will merge this to next maybe today. i am not sure about CI. thanks.

rljacob:
@oksanaguba the CI we run on PRs through github actions had a fail which was not related to the PR.  I reran it and its fine.

oksanaguba:
On chrysalis, homme build fails with
 5) intel-mkl/2020.4.304-g2qaxzf   
 4) intel/20.0.4-kodw73g      
because of lapack calls `undefined reference to `dgemm_'` etc. @abagusetty would  you want to revert mkl flag, and i will retest?

abagusetty:
> On chrysalis, homme build fails with 5) intel-mkl/2020.4.304-g2qaxzf 4) intel/20.0.4-kodw73g because of lapack calls `undefined reference to `dgemm_'` etc. @abagusetty would you want to revert mkl flag, and i will retest?

Addressed in the following commit

oksanaguba:
ran homme suites and one ERS test on chrysalis, all ok.

PR #6694, opened 2024-10-17
Add ability for MALI to create separate logs from each processor

jonbob:
Adds a new variable config_create_all_logs_in_e3sm to MALI that allows users to choose to get a separate log file from each processor

[NML] for configurations with MALI
[BFB]

### DIFF
--- a/components/mpas-albany-landice/bld/namelist_files/namelist_definition_mali.xml
+++ b/components/mpas-albany-landice/bld/namelist_files/namelist_definition_mali.xml
@@ -1087,6 +1087,14 @@ Valid values: .true. or .false.
 Default: Defined in namelist_defaults.xml
 </entry>
 
+<entry id="config_create_all_logs_in_e3sm" type="logical"
+	category="io" group="io">
+Logical flag determining if log files will be created for each processor in an E3SM configuration.  If .true., the model initializes and writes to one files per processor.
+
+Valid values: .true. or .false.
+Default: Defined in namelist_defaults.xml
+</entry>
+
 
 <!-- decomposition -->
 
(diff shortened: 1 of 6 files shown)


jonbob:
@xylar and @mark-petersen -- I've added you to this PR mostly to help decide if the other MPAS components might benefit from a similar option

xylar (approved):
I am approving based on an inspection of the code.  This looks like it would be a good capability for us to port to MPAS-Ocean and -Seaice, too, as we have discussed.

rljacob:
I already see a log.albany.XXXX.out file for every processor when I ran a test case.  So what is this PR doing?

xylar:
@rljacob, those are from Albany, not MALI.  *Not* having one Albany log per core would be a different issue, presumably for a different PR to Albany itself.

components/mpas-albany-landice/bld/namelist_files/namelist_definition_mali.xml
```diff
@@ -1087,6 +1087,14 @@ Valid values: .true. or .false.
 Default: Defined in namelist_defaults.xml
 </entry>
 
+<entry id="config_create_all_logs_in_e3sm" type="logical"
+	category="io" group="io">
+Logical flag determining if log files will be created for each processor in an E3SM configuration.  If .true., the model initializes and writes to one files per processor.
```
matthewhoffman, line 1092:
```suggestion
Logical flag determining if log files will be created for each processor in an E3SM configuration.  If .true., the model initializes and writes to one file per processor.
```

matthewhoffman (approved):
@jonbob , this looks great.  @stephenprice has tested it and confirms that it generates MALI log files for each processor, which has been invaluable for debugging an error occurring on a processor that isn't the head proc.

Given @xylar's response, do you want to add the other MPAS components to this same PR?

mark-petersen (approved):
@matthewhoffman this looks useful, and easy to control with a simple flag. Thanks for pointing it out.

jonbob:
passes e3sm_landice_developer on chrysalis with gnu compiler, with the exception of
* SMS.ne30pg2_r05_IcoswISC30E3r5_gis20.BGWCYCL1850.chrysalis_gnu.allactive-gis20km
which DIFFed last night from PR #6502

merged to next

jonbob:

PR #6698, opened 2024-10-17
Make L80 default for RRM grids

wlin7:
Change default RRM configurations from 72 to 80 levels.
L80 atm initial files for conusx4v1pg2 and northamericax4v1pg2 (narrm)
are added that are needed for nightly tests. The grid for narrm is changed to
northamericax4v1pg2_r025_IcoswISC30E3r5 as used for v3 narrm production config.
The testing PE layout is increased to 128 tasks (PS) to 512 tasks.

[non-BFB] only for the conusx4v1pg2 test. The NARRM test becomes a new test.

---------------------
Tested for atm developer, conusx4v1pg2, and coupled narrm test in wcprod.

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -390,7 +390,7 @@ _TESTS = {
             "SMS_Ld1.ne30pg2_r05_IcoswISC30E3r5.WCYCL1850.allactive-wcprod_1850",
             "SMS_Ld1.ne30pg2_r05_IcoswISC30E3r5.WCYCLSSP370.allactive-wcprodssp",
             "SMS_Ld1.ne30pg2_r05_IcoswISC30E3r5.WCYCLSSP585.allactive-wcprodssp",
-            "SMS_Ld1_PS.northamericax4v1pg2_WC14to60E2r3.WCYCL1850.allactive-wcprodrrm_1850",
+            "SMS_Ld1_P256.northamericax4v1pg2_r025_IcoswISC30E3r5.WCYCL1850.allactive-wcprodrrm_1850",
             "SMS_D_Ld1.ne30pg2_r05_IcoswISC30E3r5.CRYO1850",
             )
         },
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -390,7 +390,7 @@ _TESTS = {
             "SMS_Ld1.ne30pg2_r05_IcoswISC30E3r5.WCYCL1850.allactive-wcprod_1850",
             "SMS_Ld1.ne30pg2_r05_IcoswISC30E3r5.WCYCLSSP370.allactive-wcprodssp",
             "SMS_Ld1.ne30pg2_r05_IcoswISC30E3r5.WCYCLSSP585.allactive-wcprodssp",
-            "SMS_Ld1_P256.northamericax4v1pg2_r025_IcoswISC30E3r5.WCYCL1850.allactive-wcprodrrm_1850",
+            "SMS_Ld1_P512.northamericax4v1pg2_r025_IcoswISC30E3r5.WCYCL1850.allactive-wcprodrrm_1850",
             "SMS_D_Ld1.ne30pg2_r05_IcoswISC30E3r5.CRYO1850",
             )
         },
--- a/components/eam/bld/namelist_files/namelist_defaults_eam.xml
+++ b/components/eam/bld/namelist_files/namelist_defaults_eam.xml
@@ -111,9 +111,11 @@
 <ncdata dyn="se" hgrid="ne30np4x8arm"                     nlev="26" ocn="aquaplanet">atm/cam/inic/homme/cami_0003-01-01_arm30x8_L26_ape_c000000.nc</ncdata>
 <ncdata dyn="se" hgrid="ne0np4_arm_x8v3_lowcon"           nlev="30"                 >atm/cam/inic/homme/cami-mam3_0000-01-01_arm_x8v3_lowcon_np4_L30_c000000.nc</ncdata>
 <ncdata dyn="se" hgrid="ne0np4_arm_x8v3_lowcon"           nlev="26" ocn="aquaplanet">atm/cam/inic/homme/cami_0003-01-01_arm_x8v3_lowcon_np4_L26_ape_c000000.nc</ncdata>
+<ncdata dyn="se" hgrid="ne0np4_conus_x4v1_lowcon"         nlev="80"                 >atm/cam/inic/homme/v3.LR_mapped_conusx4v1np4-topoadj.eam.i.2010-01-01.nc</ncdata>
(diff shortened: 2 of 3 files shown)


whannah1:
@wlin7 have you generated new initial conditions for the RRM grids?

I worked on this awhile back, but I ran into issues trying to run the RRM cases out long enough to ensure that they were properly adjusted to the new vertical grid.

tangq:
We may use the atm IC from the current [v3.NARRM test](https://acme-climate.atlassian.net/wiki/spaces/CM/pages/4612587521/20240905.v3.narrm.hybrid_dt.WCYCL1850.warm.seaice_ocean.chrysalis), which has completed 50+ years.

whannah1:
@tangq those files look good for the NA grid - but we still have a test for the conus grid in `e3sm_integration`:

```
    #e3sm tests for RRM grids
    "e3sm_rrm" : {
        "tests" : (
            "SMS_D_Ln5.conusx4v1pg2_r05_IcoswISC30E3r5.F2010",
            )
        },
```

So we either need to update that grid or switch it to use the NA grid.

tangq:
@whannah1 , good point.

Because v3.NARRM updated to the resolution `northamericax4v1pg2_r025_IcoswISC30E3r5`, it might make sense to 1) update this new resolution for the e3sm_rrm test on master, and 2) keep the old resolution for the test on the maint branches.

Another option is to interpolate the L72.conusx4v1pg2_r05_IcoswISC30E3r5 IC files to L80 and run multiple years to get the stable ICs and update the tests to use the new L80 ICs.

There could be other options. I am not sure which way is the best and am leaning towards the first method, i.e. 1) and 2) above.

Down the road, when v3.NARRM is finalized we can update the rrm tests on master with the final initial conditions.

> @tangq those files look good for the NA grid - but we still have a test for the conus grid in `e3sm_integration`:
> 
> ```
>     #e3sm tests for RRM grids
>     "e3sm_rrm" : {
>         "tests" : (
>             "SMS_D_Ln5.conusx4v1pg2_r05_IcoswISC30E3r5.F2010",
>             )
>         },
> ```
> 
> So we either need to update that grid or switch it to use the NA grid.

whannah1:
@tangq what needs to be finalized about v3.NARRM?

Once that's done it might be good to change the tests to use `northamericax4v1pg2_r025_IcoswISC30E3r5` and add a default IC file from the final version. 

Is there any desire to keep testing the conus grid? 
I presume that no one would care if we switch the test to use NARRM

whannah1:
@tangq In that case we can switch the testing now in this PR because it seems the default initial condition won't need to change. 

Would you mind uploading an eam.i file from the latest run to the server?

Then we can add it as the default and switch the test on this branch.

wlin7:
> I noticed that @wlin7 indicated "To further add atm IC files and test". @wlin7 , have you done this already? If not, I can make a copy of EAM IC from my current test to the data server. Let me know.

Hi @tangq , I haven't added the file to the inputdata server. I think you have interpolated file for both 1850 and 1985. We can use the one for 1985 as a generic IC. Or we can use an eam.i generated from your previous v3.NARRM.AMIP run.

tangq:
Hi @wlin7 , I copied two v3.NARRM.L80 atm IC files to the input data directory `/lcrc/group/acme/public_html/inputdata/atm/cam/inic/homme`. The file names are:

1850 file: `20240905.v3.narrm.hybrid_dt.WCYCL1850.warm.seaice_ocean.chrysalis.eam.i.0053-01-01-00000.nc`
1985 file: `20240811.v3.narrm.hybrid_dt.F20TR.chrysalis.eam.i.2015-01-01-00000.nc`

These files should be good as temporary ICs for NARRM.L80 tests.

rljacob:
@wlin7 what is the status of this?

rljacob:
Notes:  having trouble creating initial condition files.  @wlin7 will work on it.

tangq:
> Notes: having trouble creating initial condition files. @wlin7 will work on it.

@rljacob , the initial files I pointed to above should be ready to use.

rljacob:
@tangq and @whannah1 one of you needs to approve this.

rljacob:
@wlin7 will this change answers for RRM tests?

tangq (approved):

whannah1:
@wlin7 @rljacob I'm confused about this because I only see the change to allow RRM to be L80 by default, but there's no changes for the initial condition namelist defaults - don't we need to change these for this to work?

Searching through tests.py I see both conus and northamerica RRM grids - but I'm not sure which ones are actively tested. Also, some of them are the np4 versions. We should take this opportunity to streamline the RRM tests that we want. 

Did we decide to only test the northamerica grid? I recall discussing this and concluding that the CONUS tests are faster since the grid is smaller, so it seemed better for general testing of the RRM capability. If we want to keep the conus tests then I can help my the ncdata files for that. 

Side note - I'm not a proponent of testing lots of RRM grids just to ensure the input files are present, seems wastefully redundant. However, I'll concede on this point if I'm in the minority.

rljacob:
@wlin7 remind us how this change is enough to switch the default back to L80.

rljacob:
note:  yes need to make new IC file for RRM

rljacob:
note: @wlin7 is testing initial files.

rljacob:
@wlin7 any update on this?

rljacob (approved):

whannah1:
> Only input files for L80 RRM are added and the grid for v3.narrm wcprod test is reset in the latest commit. No impact on other tests.
> 
> Merged to next.

Shouldn't we disable the CONUS tests then?
This PR is effectively breaking those tests, right?

wlin7:
> Shouldn't we disable the CONUS tests then? This PR is effectively breaking those tests, right?

We wanted to keep the smaller CONUS for routine RRM test. This PR lets both conus and narrm work with L80.

wlin7:
It turns out 4 nodes (P256) is insufficient to run coupled NARRM test on chrysalis. Changed to P512 and remerged to next.

tangq:
@wlin7 , on chrysalis I've been using 8 nodes as the minimum for coupled NARRM runs.

wlin7:
> @wlin7 , on chrysalis I've been using 8 nodes as the minimum for coupled NARRM runs.

Thanks for confirming it, @tangq . I thought P256 (effectively P256x2) would use 8 nodes, but it only used 4 nodes.

whannah1:
> > Shouldn't we disable the CONUS tests then? This PR is effectively breaking those tests, right?
> 
> We wanted to keep the smaller CONUS for routine RRM test. This PR lets both conus and narrm work with L80.

Sorry, I missed that you added the new L80 conus IC file.

rljacob (commented):
Please update the northamerica test in the prod suite to use this new grid.

rljacob:
Ignore above.  Typed in wrong box.

PR #6703, opened 2024-10-21
For pm-cpu, remove libfabric work-around of setting `FI_CXI_RX_MATCH_MODE=software`

ndkeen:
We had been using `FI_CXI_RX_MATCH_MODE=software` on pm-cpu to avoid some issues when Perlmutter was young.
There was no measurable performance difference, so did not change this setting.
Now let's try removing this and letting it use the default which is `FI_CXI_RX_MATCH_MODE=hybrid`.

[BFB]


rljacob:
what kind of testing have you done?

rljacob (approved):

ndkeen:
merged to next

ndkeen:
Tests all green as expected

PR #6709, opened 2024-10-24
Add overrun qos for pm-cpu/pm-gpu

ndkeen:
Add `overrun` as an option for submitting jobs via CIME on pm-cpu/pm-gpu.
Currently, jobs can submit to the `overrun` qos with `case.submit -a="-q overrun"`.
However, without this PR, tests like this will fail:
```
create_test SMS_D.ne4pg2_oQU480.F2010 --compiler gnu -q overrun
```

[bfb]

### DIFF
--- a/cime_config/machines/config_batch.xml
+++ b/cime_config/machines/config_batch.xml
@@ -448,6 +448,7 @@
       <queue walltimemax="00:45:00" nodemax="1792" default="true">regular</queue>
       <queue walltimemax="00:45:00" nodemax="1792" strict="true">preempt</queue>
       <queue walltimemax="00:45:00" nodemax="1792" strict="true">shared</queue>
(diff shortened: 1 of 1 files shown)


rljacob (approved):

rljacob:
So with this we could change the test script to run in the overrun queue?

ndkeen:
Yes, as with previous experience using the `shared` qos, there must be something in CIME that is unhappy when a job requests to use a qos that is not specified in `env_batch.xml`.  It does not give a great error message when that happens, but adding `overrun` as an option seems to allow it to work.  I suppose it's nice that CIME catches a potential problem when job tries to request a non-existent qos, but it makes it less flexible when a center/machine makes a change to add/remove these.

Even though the tests dont run on pm-cpu, I will merge to master.

PR #6715, opened 2024-10-29
Update CMake for oneapi-ifxgpu runs

amametjanov:
Move `-Xsycl-target-backend` to link flags.

[BFB]


amametjanov:
Testing on Aurora:
- e3sm_integration + oneapi-ifx: https://my.cdash.org/viewTest.php?buildid=2718723
- e3sm_scream_v1 + oneapi-ifx: https://my.cdash.org/viewTest.php?buildid=2718751
- e3sm_gpucxx (wout scream) + oneapi-ifxgpu: in-progress

oksanaguba:
for the gpu tests, the above says 1 out of 2 tests, but the cdash link shows only 1 test, where is the other, failing, one?

why did you test cpu builds? do i understand it right that the files you modified are not used in the cpu builds?

oksanaguba:
i cannot seem to find SYCL_EXE_LINKER_FLAGS in a recent master, but other machine files are using CMAKE_EXE_LINKER_FLAGS . i also do not remember using ${ARCH}_EXE_LINKER_FLAGS before .

thanks!

amametjanov:
The cdash link above is now updated to show both `e3sm_gpucxx` tests.
Yes, this PR has no effect on CPU builds: I ran them to check on the current state of e3sm and scream on Aurora.

This PR separates link flags `-Xsycl-target-backend \"-device 12.60.7\"` from compile flags in `SYCL_FLAGS` to clean-up warnings like
```
icpx: warning: argument unused during compilation: '-Xsycl-target-backend -device 12.60.7'
```
In E3SM builds, these flags are already in `CMAKE_EXE_LINKER_FLAGS` in [this cmake line](https://github.com/E3SM-Project/E3SM/blob/master/components/cmake/build_model.cmake#L277).
They are in a new `SYCL_EXE_LINKER_FLAGS` cmake-var for standalone Omega builds [here](https://github.com/E3SM-Project/Omega/commit/de17ea3c9b90edebd111dfc2c5a4cd033a74dbb6).
Adding the link flags to existing `CMAKE_EXE_LINKER_FLAGS` cmake-var breaks builds for other libs like mct.

oksanaguba:
if i understood it right, then `SYCL_EXE_LINKER_FLAGS` is a bad name, as it is used only by omega ('omega' should show up in the name). now i lost access to my builds, but it seems the issue will resurface in https://github.com/E3SM-Project/scream/pull/2593/files# where i used vars like `SYCL_LINK_FLAGS` . i am very confused about scope of various cmake variables in e3sm builds, and *cmake files for them (separation between e3sm shared libs and scream component). i might create a confl page with questions about this.

oksanaguba (approved):

Issue #6711, opened 2024-10-27
atmospheric chemistry reaction rate (r_lch4 and r_lco_h) output zero in e3smv3 master


jinboxie:
The current version of e3smv3 master has not not been able to output the content of reaction rate for ch4 and co_h. The outputs are zero, and we are unable to output the need variable for simulation.

```
        float r_lch4(time, lev, ncol) ;
                r_lch4:mdims = 1 ;
                r_lch4:units = "molecules/cm3/sec" ;
                r_lch4:long_name = "reaction rate" ;
                r_lch4:cell_methods = "time: mean" ;
        float r_lco_h(time, lev, ncol) ;
                r_lco_h:mdims = 1 ;
                r_lco_h:units = "molecules/cm3/sec" ;
                r_lco_h:long_name = "reaction rate" ;
                r_lco_h:cell_methods = "time: mean" ;
```
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
The mechanism is still in atmospheric chemistry chem_mech_in, but we may be meeting some problems with diagnostic output of these variables
```
[lco_h]      CO + OH -> H                                   ; 1.500E-13,  0
[...]

PR #6716, opened 2024-10-29
Chemistry output bug fix for maint-3.0

jinboxie:
The chemistry reaction rate of r_lch4 and r_lco_h are output as 0 due to a bug in the output of chemistry file, when no chemistry output flags are set. The related issues and discussion is shown https://github.com/E3SM-Project/E3SM/issues/6711#issuecomment-2441390495. The bug is fixed by modifying the if clause in the code.


tangq (approved):

hsiangheleellnl (approved):

wlin7:
Merged to maint-3.0.

PR #6718, opened 2024-10-29
Add subgrid hillslope hydrology (NGEE Arctic IM2)

thorntonpe:
This new capability builds on the existing topographic unit (topounit) subgrid scheme by
implementing transport of water within a gridcell from higher to lower elevation topounits. 
Assuming that the topounit information on an existing surface dataset includes mean 
elevation for each topounit on a gridcell, the code will calculate topounit connectivity (topology)
 at run time during model initialization. On each time step a user-specified fraction of the 
surface runoff, ponded water runoff, and perched water table drainage terms are shifted 
to the next-lower topounit, where that water is distributed to the columns on the topounit 
with a user-defined decay term.

[BFB]
[FCC]
[NML]

### DIFF
--- a/components/elm/src/biogeophys/SoilHydrologyMod.F90
+++ b/components/elm/src/biogeophys/SoilHydrologyMod.F90
@@ -15,6 +15,8 @@ module SoilHydrologyMod
   use SoilHydrologyType , only : soilhydrology_type
   use SoilStateType     , only : soilstate_type
   use WaterfluxType     , only : waterflux_type
+  use TopounitType      , only : top_pp
+  use TopounitDataType  , only : top_ws
   use LandunitType      , only : lun_pp
   use ColumnType        , only : col_pp
   use ColumnDataType    , only : col_es, col_ws, col_wf
@@ -47,12 +49,12 @@ contains
     !
     ! !USES:
       !$acc routine seq
-    use elm_varcon      , only : denice, denh2o, wimp, pondmx_urban
+    use elm_varcon      , only : denice, denh2o, wimp, pondmx_urban, frac_from_uphill
     use column_varcon   , only : icol_roof, icol_sunwall, icol_shadewall
     use column_varcon   , only : icol_road_imperv, icol_road_perv
     use elm_varpar      , only : nlevsoi, nlevgrnd, maxpatch_pft
     use elm_varpar      , only : nlayer, nlayert
-    use elm_varctl      , only : use_var_soil_thick
+    use elm_varctl      , only : use_var_soil_thick, use_IM2_hillslope_hydrology
     use SoilWaterMovementMod, only : zengdecker_2009_with_var_soil_thick
     !
     ! !ARGUMENTS:
@@ -66,7 +68,7 @@ contains
     real(r8), intent(in)  :: dtime
     !
     ! !LOCAL VARIABLES:
-    integer  :: c,j,fc,g,l,i                               !indices
+    integer  :: c,j,fc,g,l,t,i                             !indices
     integer  :: nlevbed                                    !# levels to bedrock
     real(r8) :: xs(bounds%begc:bounds%endc)                !excess soil water above urban ponding limit
     real(r8) :: vol_ice(bounds%begc:bounds%endc,1:nlevgrnd) !partial volume of ice lens in layer
@@ -101,6 +103,7 @@ contains
          qflx_floodc      =>    col_wf%qflx_floodc      , & ! Input:  [real(r8) (:)   ]  column flux of flood water from RTM
          qflx_evap_grnd   =>    col_wf%qflx_evap_grnd   , & ! Input:  [real(r8) (:)   ]  ground surface evaporation rate (mm H2O/s) [+]
          qflx_top_soil    =>    col_wf%qflx_top_soil    , & ! Output: [real(r8) (:)   ]  net water input into soil from top (mm/s)
+         qflx_from_uphill =>    col_wf%qflx_from_uphill , & ! Output: [real(r8) (:)   ]  water received from uphill topounit (mm/s)
          qflx_surf        =>    col_wf%qflx_surf        , & ! Output: [real(r8) (:)   ]  surface runoff (mm H2O /s)
          qflx_irrig       =>    col_wf%qflx_irrig       , & ! Input:  [real(r8) (:)   ]  irrigation flux (mm H2O /s)
          irrig_rate       =>    veg_wf%irrig_rate       , & ! Input:  [real(r8) (:)   ]  current irrigation rate (applied if !n_irrig_steps_left > 0) [mm/s]
@@ -241,11 +244,34 @@ contains
 
       end do
 
+      ! calculate the sum of column weights on each topounit for columns in the hydrologyc filter
+      ! This will be the istsoil, istcrop, and icol_road_perv subset of urban columns
+      ! First zero the topounit sum of weights. This zeros some multiple times, but no harm done.
+      do fc = 1, num_hydrologyc
+         c = filter_hydrologyc(fc)
+         t = col_pp%topounit(c)
+         top_pp%uphill_wt(t) = 0._r8
+      end do
+      ! Next sum the weights
+      do fc = 1, num_hydrologyc
+         c = filter_hydrologyc(fc)
+         t = col_pp%topounit(c)
+         top_pp%uphill_wt(t) = top_pp%uphill_wt(t) + col_pp%wttopounit(c)
+      end do
+
       ! remove stormflow and snow on h2osfc from qflx_top_soil
       do fc = 1, num_hydrologyc
          c = filter_hydrologyc(fc)
+         t = col_pp%topounit(c)
          ! add flood water flux to qflx_top_soil
          qflx_top_soil(c) = qflx_top_soil(c) + qflx_snow_h2osfc(c) + qflx_floodc(c)
+         
+         ! flow from uphill topounit goes to top of soil for soil, crop, and pervious road columns
+         if (use_IM2_hillslope_hydrology) then
+            qflx_from_uphill(c) = (col_pp%wttopounit(c)/top_pp%uphill_wt(t)) * (frac_from_uphill * top_ws%from_uphill(t)) / dtime
+            qflx_top_soil(c) = qflx_top_soil(c) + qflx_from_uphill(c)
+        endif
+
       end do
 
     end associate
--- a/components/elm/src/biogeophys/SoilHydrologyMod.F90
+++ b/components/elm/src/biogeophys/SoilHydrologyMod.F90
@@ -244,21 +244,6 @@ contains
 
       end do
 
-      ! calculate the sum of column weights on each topounit for columns in the hydrologyc filter
-      ! This will be the istsoil, istcrop, and icol_road_perv subset of urban columns
-      ! First zero the topounit sum of weights. This zeros some multiple times, but no harm done.
-      do fc = 1, num_hydrologyc
-         c = filter_hydrologyc(fc)
-         t = col_pp%topounit(c)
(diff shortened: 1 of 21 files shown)


rljacob:
Have there been any long-ish runs with this feature on?

thorntonpe:
@rljacob I can start a longer test - how long do you suggest?

rljacob:
I'd suggest a 1-year smoke test.  Is this going to be used by E3SM in some future version?

thorntonpe:
Thanks, I'll set that up. Next phase for this work is to evaluate multiple Arctic code changes operating together at the pan-Arctic scale (under ecosystem project NGEE Arctic). The plan would be to have this as a candidate capability in whatever version is current about three years from now.

thorntonpe:
@rljacob I ran a one-year test of this new capability on the global r05 land domain, based on the existing r05 topounit-enabled surface dataset. That run completed successfully. 
@tktesfa2014 and @bishtgautam I'm hoping you can provide your reviews. Let me know if you want to get together to discuss details.

components/elm/src/biogeophys/SoilHydrologyMod.F90
```diff
@@ -241,11 +244,34 @@ subroutine SurfaceRunoff (bounds, num_hydrologyc, filter_hydrologyc, &
...
+         c = filter_hydrologyc(fc)
+         t = col_pp%topounit(c)
+         top_pp%uphill_wt(t) = 0._r8
+      end do
+      ! Next sum the weights
+      do fc = 1, num_hydrologyc
+         c = filter_hydrologyc(fc)
+         t = col_pp%topounit(c)
+         top_pp%uphill_wt(t) = top_pp%uphill_wt(t) + col_pp%wttopounit(c)
+      end do
```
bishtgautam, line 260: Couldn't these weights be computed during initialization?

bishtgautam (changes requested):
@thorntonpe  I have left a few minor comments.

rljacob:
@thorntonpe ping.

components/elm/src/biogeophys/SoilHydrologyMod.F90
```diff
@@ -241,11 +244,34 @@ subroutine SurfaceRunoff (bounds, num_hydrologyc, filter_hydrologyc, &
...
          ! add flood water flux to qflx_top_soil
          qflx_top_soil(c) = qflx_top_soil(c) + qflx_snow_h2osfc(c) + qflx_floodc(c)
+         
+         ! flow from uphill topounit goes to top of soil for soil, crop, and pervious road columns
+         if (use_IM2_hillslope_hydrology) then
+            qflx_from_uphill(c) = (col_pp%wttopounit(c)/top_pp%uphill_wt(t)) * (frac_from_uphill * top_ws%from_uphill(t)) / dtime
+            qflx_top_soil(c) = qflx_top_soil(c) + qflx_from_uphill(c)
+        endif
+
       end do
```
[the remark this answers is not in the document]
thorntonpe: Moved to a new loop, as suggested.

components/elm/src/data_types/ColumnDataType.F90
```diff
@@ -5843,6 +5847,14 @@ subroutine col_wf_init(this, begc, endc)
...
            ptr_col=this%qflx_snofrz, set_lake=spval, c2l_scale_type='urbanf', default='inactive')
 
+     call hist_addfld1d (fname='QFROM_UPHILL',  units='mm/s',  &
+          avgflag='A', long_name='input to top layer soil from uphill topounit(s)', &
+           ptr_col=this%qflx_from_uphill, c2l_scale_type='urbanf')
+
+     call hist_addfld1d (fname='QTO_DOWNHILL',  units='mm/s',  &
+          avgflag='A', long_name='output from column to downhill topounit', &
+           ptr_col=this%qflx_to_downhill, c2l_scale_type='urbanf')
+
```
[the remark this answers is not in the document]
thorntonpe: Placed inside if statement, as suggested.

components/elm/src/data_types/TopounitDataType.F90
```diff
@@ -636,5 +651,51 @@ subroutine clean_top_es(this, begt, endt)
...
+  subroutine init_top_ws(this, begt, endt)
+    class(topounit_water_state) :: this
+    integer, intent(in) :: begt   ! beginning topographic unit index
+    integer, intent(in) :: endt   ! ending topographic unit index
+
+    allocate(this%from_uphill   (begt:endt)) ; this%from_uphill   (:) = spval
+    !-----------------------------------------------------------------------
+    ! initialize history fields for members of top_ws
+    !-----------------------------------------------------------------------
+     call hist_addfld1d (fname='FROM_UPHILL',  units='mm',  &
```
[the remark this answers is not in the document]
thorntonpe: Changed output field name and placed inside if statement, as suggested.

components/elm/src/main/organicFileMod.F90
```diff
@@ -106,6 +106,9 @@ subroutine organicrd(organic)
             dim1name=grlnd, readvar=readvar)
        if (.not. readvar) call endrun('organicrd: errror reading ORGANIC')
 
+       ! Close file
+       ! call ncd_pio_closefile(ncid)
+
```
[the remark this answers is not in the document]
thorntonpe: Removed lines.

components/elm/src/biogeophys/SoilHydrologyMod.F90
thorntonpe: I'd rather leave the weight calculation in place during the time step since we are likely to have dynamic column weight on the topounit from other near-term developments.

thorntonpe:
@bishtgautam I made most requested changes in new commit and resolved all but one comment. The weight calculation could be done for now during initialization, but keeping it inline in the time step provides some future proofing for expected near-term development. Ready for your further review.

components/elm/src/biogeophys/SoilHydrologyMod.F90
bishtgautam: In that case, how about moving these computations within a `if (use_IM2_hillslope_hydrology)` block?

components/elm/src/main/elm_varcon.F90
```diff
@@ -222,6 +222,14 @@ module elm_varcon
   real(r8), public, parameter :: snw_rds_min = 54.526_r8    
   !-----------------------------------------------------------------------
 
+  !------------------------------------------------------------------
+  ! IM2 hillslope hydrology constants
+  !------------------------------------------------------------------
+  ! fraction of surface runoff fluxes that get transferred to downhill topounit
+  real(r8), public, parameter :: frac_to_downhill = 1.0  ! (fraction, valid values from 0.0 to 1.0)
+  ! fraction of from_uphill water state that get transferred to columns on the topounit in each timestep
+  real(r8), public, parameter :: frac_from_uphill = 0.5  ! (fraction, valid values from 0.0 to 1.0)
```
bishtgautam, line 231: How about adding `_r8`?

bishtgautam (commented):
@thorntonpe, thanks for the changes. I left two more comments for you to consider.

components/elm/src/biogeophys/SoilHydrologyMod.F90
thorntonpe: Good idea - I moved this into a single if-block with the flux calculation.

components/elm/src/main/elm_varcon.F90
thorntonpe: Done.

thorntonpe:
@bishtgautam thanks for the further review - requested changes are complete.

bishtgautam (approved):
The changes look good to me.

thorntonpe:
@tktesfa2014 would you be willing to take a look at this PR? It has already been reviewed by @bishtgautam but I'd like to get your input since it operates on top of the topounit structure. You can see the feature description document and other information at [https://acme-climate.atlassian.net/wiki/spaces/LAND/pages/4542398465/Subgrid+hillslope+hydrology+NGEE+Arctic+IM2](url)

thorntonpe:
@bishtgautam Not sure if I've done the right thing here. There were merge conflicts showing up in the PR github page. They were easy to resolve so I did that within github, and clicked the "merge commit" button shown in the conflict resolution page. Now the PR shows that master has been merged to my branch. Is that a problem for further integration?

bishtgautam:
@thorntonpe, the conflicts were indeed trivial, and I was planning on resolving them while merging your branch. We can force reset the branch to be at [874f2bc](https://github.com/E3SM-Project/E3SM/pull/6718/commits/874f2bc757d80ae94b91a4c99d51aa9ac19c9f22), thus removing the merge of master into your branch.

rljacob:
@thorntonpe don't do what github suggests for resolving merge conflicts.  It just merges master to the branch which we don't want.   @bishtgautam yes force reset the branch.

rljacob:
@tktesfa2014 please review

components/elm/src/main/initGridCellsMod.F90
tktesfa2014, line 1: @thorntonpe 
In current topounit-based surface data files, the topounit with lowest elevation has the smallest index number and the topounit with highest elevation has the highest index number. So, with the current data structure of the topounit-based surface data, the connectivity could be established easily. Can this also handle if we upgrade the topounit generation to include aspect?

tktesfa2014 (changes requested):
@thorntonpe
I think you should use here a more recent topounit-based surface data file. For example:
topounit_based_half_degree_merge_surfdata_0.5x0.5_simyr1850_c211019.20211108_ed20220204.nc. This surface data file has gone through some improvement.

@thorntonpe
In current topounit-based surface data files, the topounit with lowest elevation has the smallest index number and the topounit with highest elevation has the highest index number. So, with the current data structure of the topounit-based surface data, the connectivity could be established easily. Can this also handle if we upgrade the topounit generation to include aspect?

thorntonpe:
> @thorntonpe I think you should use here a more recent topounit-based surface data file. For example: topounit_based_half_degree_merge_surfdata_0.5x0.5_simyr1850_c211019.20211108_ed20220204.nc. This surface data file has gone through some improvement.
> 
@tktesfa2014 I am looking for the suggested file in the inputdata directory (searching on perlmutter) but not finding it. I also see that the test for topounits, which is the root of this new test for IM2, uses the older file. I think it makes sense to use the same file for both tests. What do you think about leaving this as-is, and then we can make a new PR that updates the surface data file for both tests? In either case, can you point me to the new file?

> @thorntonpe In current topounit-based surface data files, the topounit with lowest elevation has the smallest index number and the topounit with highest elevation has the highest index number. So, with the current data structure of the topounit-based surface data, the connectivity could be established easily. Can this also handle if we upgrade the topounit generation to include aspect?

@tktesfa2014 having the topounits arranged this way on the surface data file makes sense. I think it is still safer to do the checks on topounit elevation within the code and assign the run-time indices accordingly. That avoids the need to make assumptions about the surface data file. The current organization will not break for a future definition of topounits that uses both elevation and aspect, but depending on how the aspect classes are defined the current implementation could be improved to optimize the hillslope connections. I think it is best to make those modifications if and when an aspect-based topounits definition is created.

tktesfa2014:
@thorntonpe:
The updated topounit-based surface data is located at /compyfs/inputdata/lnd/clm2/surfdata_map/topounit_based_half_degree_merge_surfdata_0.5x0.5_simyr1850_c211019.20211108_ed20220204.nc. Could you point me to the input directory on Perlmutter so that I can upload it?

I think it makes more sense to use the one you have for this PR if you have used it in the test for IM2. I agree we can add the updated surface data in another PR. 

I have approved the PR. 
Thanks!

tktesfa2014 (approved):

PR #6722, opened 2024-10-30
Set diagnositc ice falling flux as zero to avoid ice double counting in COSP.

yunpengshan2014:
PR#5827 introduced an unexpected issue affecting the COSP CALIPSO 
simulator diagnostics. This issue arises from assigning the P3 diagnostic 
precipitating solid phase particle falling flux to “precip_ice_flux” for 
diagnostic purposes. Since “precip_ice_flux” is also a required input field
 and is treated as snow in the CALIPSO simulator, this results in the 
double-counting of solid phase particles. This is because the E3SMv3 
implemented the single-ice category version of P3 as used in SCREAM, 
where the simulated solid particles are either precipitating or non-precipitating. 

While assigning the P3 diagnostic precipitating solid phase particle falling
 flux to “precip_ice_flux” is technically correct, it inadvertently leads to 
double-counting of ice in the CALIPSO simulator's cloud diagnostics. 
Note that this issue does not impact model simulations.

[BFB] except for tests with COSP CALIPSO simulator output

### DIFF
--- a/components/eam/src/physics/p3/eam/micro_p3.F90
+++ b/components/eam/src/physics/p3/eam/micro_p3.F90
@@ -4350,7 +4350,6 @@ subroutine ice_sedimentation(kts,kte,ktop,kbot,kdir,    &
               dt_left, prt_accum, inv_dz, inv_rho, rho, num_arrays, vs, fluxes, qnr, dt_sub)
 
          do k = k_qxbot,k_qxtop,kdir
-!            precip_ice_flux(k+1) = precip_ice_flux(k+1) + flux_qit(k)*dt_sub ! shanyp
             sflx(k+1) = sflx(k+1) + flux_qit(k)*dt_sub
          enddo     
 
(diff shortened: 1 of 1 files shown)


rljacob:
@yunpengshan2014 can you open an issue describing the problem?

yunpengshan2014:
Hi Rob,
We have a confluence page describing this problem:
https://acme-climate.atlassian.net/wiki/spaces/ATMOS/pages/edit-v2/4654497798?draftShareId=c179bae8-e00d-4754-8faf-0d05f5eb48a2.

Does it work?
Regards,
Yunpeng

rljacob:
No there needs to be a github issue.

wlin7:
@yunpengshan2014 , look like with the current P3 scheme, the assignment for precip_ice_flux should never exist. I would suggest delete the line altogether, instead of just commenting out.

crterai:
@yunpengshan2014 - Would you address @wlin7's comment? Then I can approve the PR.

wlin7:
@crterai , I have remove the commented out lines. If ever needed again for diagnostics purpose, the calculations are the same as for `sflx`.

wlin7 (approved):

wlin7:
Merged to next.

PR #6725, opened 2024-10-31
Fix alarm error for MALI adaptive timestepper

matthewhoffman:
MALI uses an MPAS alarm to handle ensuring its adaptive timestepper does not step past a coupling interval.  In recent attempts to enable the adaptive timestepper in MALI within E3SM, it was discovered that the required alarm is not being reset properly during init in the glc driver, which results in an error because MALI tries to use a 0 second dt on the first timestep.  This PR resets the alarm in glc_init_mct after the component clock is finished being set up, which solves the problem.

This PR also enables the adaptive timestepper for the `mali-gis20km` testmod to ensure this functionality gets tested.  Because the coupling interval in this test (1 day) is much smaller than the CFL-limiting timestep, the dt being applied remain unchanged, so there are no answer changes.  But the adaptive timestepper functionality in MALI is active and being run.

[NML] only for testdef mali-gis20km
[BFB]

### DIFF
--- a/components/mpas-albany-landice/cime_config/testdefs/testmods_dirs/mali/gis20km/user_nl_mali
+++ b/components/mpas-albany-landice/cime_config/testdefs/testmods_dirs/mali/gis20km/user_nl_mali
@@ -1,1 +1,3 @@
 config_am_globalstats_enable = .false.
+config_adaptive_timestep = .true.
+config_adaptive_timestep_force_interval = '0000-00-01_00:00:00'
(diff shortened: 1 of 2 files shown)


matthewhoffman:
Tested successfully with:
```
./create_test --wait -q debug  --walltime 0:30:00    ERS.ne30pg2_r05_IcoswISC30E3r5_gis20.IGELM_MLI.pm-cpu_gnu.mali-gis20km
```

stephenprice (approved):
Approved -- based on recent rigorous testing (of both changes) in IG cases using the 1-to-10km GIS init. cond.

jonbob (approved):
Approved based on testing and visual inspection

jonbob:
Passes e3sm_landice_developer on chrysalis with the gnu compiler with only NML DIFFs for tests using the mali-gis20km testdef, which is modified by this PR

merged to next

jonbob:

PR #6726, opened 2024-10-31
HOMME: Removed an old kokkos cmd line flag: kokkos-num-devices

abagusetty:
1. Remove a removed kokkos command line flag not applicable anymore
2. Addresses the print statements from each rank

[BFB]

### DIFF
--- a/components/homme/src/share/cxx/ExecSpaceDefs.cpp
+++ b/components/homme/src/share/cxx/ExecSpaceDefs.cpp
@@ -32,13 +32,46 @@ namespace Homme {
 // own. As a side benefit, we'll end up running on GPU platforms optimally
 // without having to specify --kokkos-ndevices on the command line.
 void initialize_kokkos () {
-  // This is in fact const char*, but Kokkos::initialize requires char*.
-  std::vector<char*> args;
-  const char* silence = "--kokkos-disable-warnings";
-  args.push_back(const_cast<char*>(silence));
+  // Count up our devices.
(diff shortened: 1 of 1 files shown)


bartgol (approved):
Now Kokkos handles selection of device internally, via round robin, right?

rljacob:
What is parsing these command line strings?  Standalone eamxx ?

bartgol:
> What is parsing these command line strings? Standalone eamxx ?

This is standalone hommexx. That is, when we run the HOMMEXX case in CIME.

ambrad:
It's not clear to me this should work on every machine. The key question is whether each ranks sees one device assigned to it by the job run command or multiple devices. This code addresses the second case.

I suggest as an alternative that this updated approach in EKAT be used: https://github.com/E3SM-Project/EKAT/blob/4e36a487d9ced3951907b36a1a16b13325d796d1/src/ekat/ekat_session.cpp.

abagusetty:
Thanks, The above ekat snippet works when the job-schedulers (slurm/pbs) doesnt do the MPI-binding to the GPUs

oksanaguba:
@abagusetty or @ambrad -- i am not following the above, do you guys think this PR is ok or does it still need some modifications? thanks!

ambrad (changes requested):
The ekat session code should be used here as a replacement to make sure we can run on machines where each rank sees all the GPUs on a node.

ambrad (approved):

PR #6738, opened 2024-11-07
Adding I/O type for ADIOS compression & Updating SCORPIO to v1.6.6

jayeshkrishna:
Adding preliminary support for ADIOS output compression
and updating to SCORPIO v1.6.6 (which includes patch
to support the new I/O type to use for data compression)

Modifying CIME config and scripts to support writing
ADIOS output files with data compression. A new I/O
type is added to support writing output in ADIOS BP file
format with data compression.

However users would need to move to SCORPIO 
master to write ADIOS output with data compression

Note: Writing ADIOS output with data compression is
now supported in SCORPIO master but is not
included in SCORPIO v1.6.6. A future release of
SCORPIO will include this feature

[BFB]


dqwu (approved):

jayeshkrishna:
Creating tests with CIME (create_test) on next failed on machines (anlgce) with ADIOS that includes support for BLOSC (Note: create_newcase however does not show this issue and creates the case successfully on these machines). This issue has been fixed on the branch and will be re-merged to next.

PR #6740, opened 2024-11-08
update sbetr to remove redundant print to log

jinyuntang:
This avoids redundant output to log file.
Also suspend unit testing until pfunit fixed.
Update sbetr submodules to use ssh.  Get ready to fix obsolete 3rd party libs

[BFB]


jinyuntang:
@ndkeen I checked the test ERS.f19_g16.I1850ELM.pm-cpu_gnu.elm-betr, it no longer produce the large log file. Please review. Thanks.

ndkeen:
I see that you have commented (not removed) the debug print statement. But it kinda looks like other things were also changed?  I might ask others involved to verify those changes are inert -- or would benefit from a different PR -- or at least update to the title.

rljacob:
The current version of sbetr being used is 66260f4991d6.  You should make a branch from that, fix just the debug statement, and then update the submodule to point to that branch.  Otherwise this is a much bigger "update sbetr" PR.

rljacob:
The other updates looks like its just testing, not code, so maybe this is fine as is.

ndkeen:
Yes, I also see they are testing related -- I just don't know if it was intentional, related, important, etc. So if you think it's ok, I will approve.

ndkeen (approved):

rljacob:
Wait for @peterdschwartz or @bishtgautam to approve.

bishtgautam (approved):

ndkeen:
Can we merge this?

ndkeen:
Can we find someone more familiar with working with submodules?

PR #6746, opened 2024-11-12
Set default config_check_ssh_consistency for oQU240wLI to false

jonbob:
Change the default config_check_ssh_consistency setting for oQU240wLI to false, to avoid confusing warnings/errors from tests with this grid. Otherwise we end up with log.ocean.err files complaining about:
    "ERROR:  Warning: Sea surface height outside of acceptable range (20m)."
In general, mpaso sets config_check_ssh_consistency to false for grids with ice shelf cavities, but this one was missed.

[NML] for configurations with oQU240wLI
[BFB]

### DIFF
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -630,6 +630,7 @@
 <!-- debug -->
 <config_check_zlevel_consistency>.false.</config_check_zlevel_consistency>
 <config_check_ssh_consistency>.true.</config_check_ssh_consistency>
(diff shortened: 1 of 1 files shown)


jonbob:
Testing with ERS_Ld5.TL319_oQU240wLI_ais8to30.MPAS_LISIO_JRA1p5.chrysalis_intel.mpaso-ocn_glcshelf shows that the results are not changed by this PR, but the log.ocean.XXXX.err files are no longer produced

xylar (approved):
@jonbob, yes, absolutely!  This looks good.  Thanks for making this much-needed change.

jonbob:
Passes:
* ERS_Ld5.TL319_oQU240wLI_ais8to30.MPAS_LISIO_JRA1p5.chrysalis_intel.mpaso-ocn_glcshelf (with only an NML DIFF)

merged to next

jonbob:

PR #6748, opened 2024-11-14
enable eamxx docs + add pyscream docs

mahf708:
Enables building eamxx docs on E3SM repo; adds pyscream prelim docs as well.
Add content to eamxx user and dev guide main pages.


bartgol (approved):
Maybe the eamxx-gh-pages action should be updated to actually be the E3SM mkdocs action? Meaning, change the name, and check paths for ALL yml/md files and folders that can impact gh-pages for the whole project?

bartgol:
Ah, I quickly glanced at the folder and missed the e3sm-gh-pages file. Well, then follow up: why do we need two actions in this repo?

PR #6761, opened 2024-11-21
FINAL eamxx downstream merge!

jgfouca:
The scream/eamxx fork has been archived (frozen). All new scream/eamxx development will happen in the E3SM repo.

[BFB]


components/eamxx/src/physics/p3/tests/p3_run_and_cmp.cpp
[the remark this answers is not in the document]
bartgol: It doesn't seem to be linked (no pun intended) to shared-vs-static libraries. I use shared libs all the time on my workstation, and it builds just fine. It seems more like a problem of circular dependencies (which has been something I've been wanting to fix, by breaking down scream_share into smaller units): scream_io links against scream_share, but scream_share also needs some stuff from scream_io (e.g., in eamxx_time_interpolation).

I'm not sure why this shows up here though. I don't get link errors on my machine (where, again, I do use shared libs)

xylar:
Congratulations, you all!

jgfouca:
I'm going to merge to next so we get testing tonight. We can simultaneously work on fixing the CI issue.

.github/workflows/eamxx-v1-testing.yml
```diff
@@ -53,9 +54,22 @@ jobs:
...
 
           # Use the GitHub API to get the list of changed files
-          response=$(curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
-            "https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.number }}/files")
-          changed_files=$(echo "$response" | grep -o '"filename": *"[^"]*"' | sed 's/"filename": *//; s/"//g')
+          # There are page size limits, so do it in chunks
+          page=1
+          while true; do
+            response=$(curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+              "https://api.github.com/repos/E3SM-Project/scream/pulls/${{ github.event.number }}/files?per_page=100&page=$page")
```
bartgol, line 61: This should not hard code "scream" as the repo name. That said, we decided we're going to roll back to paths filter, and get rid of the pre-process-pr job, so I'm ok leaving this as is, and create a PR right after.

bartgol: This is the reason why the eamxx-v1 and eamxx-sa workflows did not run on this PR.

rljacob:
merged to next.

bartgol (approved):

PR #6767, opened 2024-11-21
Update the icepack submodule to v1.5.0

eclare108213:
Replaces the Icepack submodule in E3SM with Icepack version 1.5.0 and updates MPAS-seaice to accommodate the new code.

This PR is not BFB due to changes in the Icepack column physics modules to fix a bug in a flux-limiting term, to fix a bug in the lateral melting code, to fix conservation issues associated with the floe size distribution, and to support testing the 5-band SNICAR radiation data.

[not BFB]

### DIFF
--- a/components/mpas-seaice/bld/namelist_files/namelist_defaults_mpassi.xml
+++ b/components/mpas-seaice/bld/namelist_files/namelist_defaults_mpassi.xml
@@ -423,6 +423,7 @@
 <config_slow_mode_drainage_strength>-5.0e-8</config_slow_mode_drainage_strength>
 <config_slow_mode_critical_porosity>0.05</config_slow_mode_critical_porosity>
 <config_macro_drainage_timescale>10.</config_macro_drainage_timescale>
+<config_congelation_freezing_method>'two-step'</config_congelation_freezing_method>
 <config_congelation_ice_porosity>0.85</config_congelation_ice_porosity>
 
 <!-- itd -->
--- a/components/mpas-seaice/bld/namelist_files/namelist_definition_mpassi.xml
+++ b/components/mpas-seaice/bld/namelist_files/namelist_definition_mpassi.xml
@@ -2487,6 +2487,14 @@ Valid values: Any positive real number.
 Default: Defined in namelist_defaults.xml
 </entry>
 
+<entry id="config_congelation_freezing_method" type="char*1024"
+	category="thermodynamics" group="thermodynamics">
+Method used for congelation freezing.
+
+Valid values: 'one-step', 'two-step'
+Default: Defined in namelist_defaults.xml
+</entry>
+
 <entry id="config_congelation_ice_porosity" type="real"
 	category="thermodynamics" group="thermodynamics">
 Liquid fraction of congelation ice.
--- a/components/mpas-seaice/docs/tech-guide/index.md
+++ b/components/mpas-seaice/docs/tech-guide/index.md
@@ -84,7 +84,7 @@ A paper describing the advanced snow physics is in preparation.
 
 ### Biogeochemistry
 
-This section is under construction, pending the full merge of BGC codes in Icepack and the older column physics package.
+Icepack incorporates a comprehensive biogeochemistry component that includes two aerosol options, water isotopes, and a full suite of vertically resolved biogeochemical tracers.  The complexity of the ecosystem can be specified in Icepack’s configuration settings and namelist flags. The biogeochemistry component is thoroughly documented in the Icepack documentation linked above.
 
 ## Coupling of MPAS-seaice within E3SM
 
--- a/components/mpas-seaice/src/Registry.xml
+++ b/components/mpas-seaice/src/Registry.xml
@@ -1752,6 +1752,10 @@
 			possible_values="Any positive real number."
 			icepack_name="tscale_pnd_drain"
 		/>
+		<nml_option name="config_congelation_freezing_method" type="character" default_value="two-step" units="unitless"
+			description="Method for congelation ice freezing."
+			possible_values="'one-step', 'two-step'"
+		/>
 		<nml_option name="config_congelation_ice_porosity" type="real" default_value="0.85" units="unitless"
 			description="Liquid fraction of congelation ice."
 			possible_values="Any real number between 0 and 1."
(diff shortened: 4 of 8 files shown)


eclare108213:
There are some necessary changes in MPAS-seaice to enable Icepack v1.5.0 to run.  Icepack now takes the (category, cell) array for the fraction of lateral ice melting instead of just (cell).  This array has been added to Registry without removing the old one, which is still used by the the old column physics code (and will be deprecated in a future PR).

`USE_SNICAR` is no longer used in Icepack.  This is the default behavior and so the cpp has simply been removed.

Two Icepack subroutine names changed to better describe what the subroutines do:
> use icepack_therm_shared  , only: icepack_init_thermo => icepack_init_salinity 
> use icepack_therm_shared  , only: icepack_init_trcr => icepack_init_enthalpy 

The code changes are more fully described in https://acme-climate.atlassian.net/wiki/spaces/ICE/pages/4765843553/Overview+of+Icepack+synchronization+version+1.5.0

njeffery:
@eclare108213 :  The build failed when I just checked out your branch.  I suspect I'm pointing to the wrong Icepack? Could you send me the icepack commit or checkout command?  Thanks

eclare108213:
> The build failed when I just checked out your branch. I suspect I'm pointing to the wrong Icepack? Could you send me the icepack commit or checkout command? Thanks

Hmmm, not sure why it wouldn't work straight from this branch.  The icepack submodule should be pointing to this hash:
`22b8769 (HEAD -> main, origin/main, origin/HEAD) Merge pull request # 35 from CICE-Consortium/main`

If it isn't, then in the E3SM directory try

```
git submodule sync
git submodule update --init
```

(which might point it back to E3SM's current icepack submodule, I'm not sure).  If that doesn't work, then just do it manually,
 
```
cd components/mpas-seaice/src
mv icepack icepack.orig
git clone git@github.com:E3SM-Project/Icepack icepack
```

njeffery:
> git submodule sync
> git submodule update --init

@eclare108213 : That worked!  pointing to the right hash now.  I'll try again.  How should I set the alternative congelation flag?

eclare108213:
> How should I set the alternative congelation flag?

@njeffery Yay! See description in 
https://acme-climate.atlassian.net/wiki/spaces/ICE/pages/4795236381/Alternative+congelation+freezing+method

njeffery:
Trying config_congelation_freezing_method = 'one-step' but keeping the congelation porosity the same.

njeffery:
@eclare108213 :  I've finished the 15-year fully coupled bgc runs and compared with a baseline Mat ran.  Figures are here: https://acme-climate.atlassian.net/wiki/spaces/HESF/pages/4806148098/20241124.v3.LR.icepack1.5.0.CBGC

The run is too short to see much of a difference in the physics or sea ice bgc, but certainly no concerns. However there is a difference in the sea ice carbon conservation for the Arctic.  The new frazil/congelation partition appears to noticeably reduce the carbon error.  Yay!

eclare108213:
@njeffery That's awesome!  Thank you.  The new congelation option is turned off in this PR but I do recommend that E3SM start using it.  We can discuss how to go about that offline.

I have a 100-year B case running now, started from the v3.0.0 historical simulation.

njeffery:
Does your 100-yr run change the congelation porosity too or just tests the one-step?

eclare108213:
I'm testing the code as it is currently PR'd, so only one-step.  The run is only 9 years in so far, and I could easily stop and restart but I think that test should be in a separate PR.

njeffery:
I agree.

eclare108213:
A 100-year B case simulation is complete and is quite similar to the v3.LR.historical_0051 control simulation.  Links to plots, performance and other documentation are available [here](https://acme-climate.atlassian.net/wiki/spaces/ICE/pages/4765843553/Overview+of+Icepack+synchronization+version+1.5.0).

erinethomas (approved):
I approve this PR. I have tested the updated Icepack submodule in the wave-sea ice coupled framework which depends upon the updated floe size distribution provided here. All wave coupling tests pass with the new icepack submodule.

proteanplanet (approved):
Changes to Icepack yield climatologically identical results for the 1850-1950 051 E3SM V3 LR ensemble member for sea ice volume and snow-on-sea ice volume at the 99.9% compatibility interval, as illustrated below for dotted lines in the Icepack 1.5 case as compared to the control (solid blue).  Dotted red lines indicate the same climate, solid red indicates extent is almost identical at the 99.9% confidence interval except for a small period in the last quarter of the solar cycle. 

E3SM_sea_ice_lemnisc 1850-1950 1850-1950 Control Icepack icepack extent volume snow 3

Approved based on nearly identical and stable B-case results as compared to a V3 LR production simulation.

proteanplanet:
@njeffery:  The Coupled Group HR critical path group would like to incorporate this into a new all-up sea ice and ocean test as soon as is possible.  Please can you review this PR and let us know if you have any concerns?

Tagging @xuezhengllnl and @vanroekel

njeffery (approved):
Approved based on a 15 year simulation with config_congelation_freezing_method = 'one-step' and fully-coupled BGC configuration.  The code changes produce notable improvements in sea ice carbon conservation and minor changes to primary production and other BGC metrics.

proteanplanet:
@jonbob This is ready to go.

jonbob:
@proteanplanet -- thanks, it's next on my list

jonbob:
@eclare108213 -- would you expect this to change answers for F-cases, where mpas-seaice is operating in its prescribed mode? My first test was not BFB, so I just wanted to make sure that agrees with your expectations

eclare108213:
@jonbob Yes, the prescribed ice mode uses the sea ice thermodynamics routines, in which there are non-bit-for-bit changes/bug fixes.

jonbob:
Completed with expected DIFFs:
* ERP_Ld3.ne4pg2_oQU480.F2010.chrysalis_intel

merged to next

jonbob:
merged to master and expected DIFFs blessed except for anvil, which is having issues and has not yet reported

jonbob:
expected DIFFs blessed on anvil as well

PR #6770, opened 2024-11-22
EAMxx: refactor vertical remapper to handle different top/bot extrapolations

bartgol:
In view of unifying the "dataset interpolation" implementations across eamxx, this PR makes the VerticalRemapper a bit more general. Namely:

- can be constructed from two existing grids (as opposed as assuming tgt grid is from map file);
- can handle "P0" and "Masked" extrapolation (the former does constant extrapolation, the latter sets extrapolated value to a constant mask value)
- can do different extrapolations at top and bottom of the column (but same extrapolation type across all columns)
- can handle 1d as well as 2d pressure profile, for both source and target pressure

Depends-On: #6761

### DIFF
--- a/components/eamxx/src/share/grid/remap/vertical_remapper.cpp
+++ b/components/eamxx/src/share/grid/remap/vertical_remapper.cpp
@@ -656,7 +656,9 @@ extrapolate (const Field& f_src,
       auto f_src_v = f_src.get_view<const Real**>();
       auto f_tgt_v = f_tgt.get_view<      Real**>();
       auto policy = ESU::get_default_team_policy(ncols,nlevs_tgt);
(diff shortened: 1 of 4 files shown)


bartgol:
@tcclevenger is on the road, moving across country.

@AaronDonahue, the CUDA fails are the same as in master. The gh/ci fails are mysterious, but maybe @mahf708 knows what the deal is. That said, it's likely orthogonal to this PR.

AaronDonahue (approved):

bartgol:
(a)ccidentally deleted the branch)

bartgol:
The last round of testing showed all passing and the known master diff on CUDA. The ghci-snl runners are having network issues right now, so they are stuck, but nothing changed since the last round of tests, so I'm merging.

PR #6772, opened 2024-11-22
Pacer timing interface for C++ components like Omega

sarats:
Add Pacer, a timing interface for C++ components in E3SM such as Omega. Presently, it provides a thin wrapper around 
GPTL with comprehensive error checking. In addition to providing basic timer functionality, it keeps track of open timers
and issues warnings about dangling timers at the end.

To begin with, the build configuration will be added to standalone Omega builds. We will add necessary CMake logic for 
coupled E3SM builds in the future once Omega is integrated into E3SM main repo.

It just adds the sources under share/pacer and would not impact E3SM's builds at this point.
[BFB]

------

Additional details including unit test and testing with Omega standalone driver on Frontier at https://acme-climate.atlassian.net/wiki/spaces/OO/pages/4775346241/Pacer+timing+interface


sarats:
cc @mark-petersen, @grnydawn

share/pacer/Pacer.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+    if ( PacerMode == PACER_STANDALONE )
+        PACER_CHECK_ERROR(GPTLfinalize());
+
+    if ( (MyRank == 0) && ( OpenTimers.size() > 0) ){
+        std::cerr << "[WARNING] Pacer: Following " << OpenTimers.size() << " timer(s) is/are still open." << std::endl;
+        for (auto i = OpenTimers.begin(); i != OpenTimers.end(); i++)
+            std::cerr << '\t' << i->first << std::endl;
+    }
+
```
philipwjones, line 196: Might want a OpenTimers.clear() call just to make sure the static map is cleared of all entries on finalization.

philipwjones (approved):
Looks good - just had the one minor suggestion. This appears to be working in Omega testing and should have no impact on other E3SM components yet.

share/pacer/Pacer.cpp
sarats: Done.

sarats:
FYI, minor additions to check if already initialized and clear state and resources at finalization.

sarats:
This PR provides the functionality needed for Omega immediately. FYI, I will submit a thread-safe version in a future PR after testing. 

The initialize function (and underlying GPTLinitialize() ) are expected to be called from a single unthreaded region in the code. I'm going to make the timer bookkeeping logic (OpenTimers) within start/stop thread-safe and assess the performance impact as well.

sarats:
FYI, I will address threading support and additional functionality in future PRs.

@rljacob @amametjanov Omega folks are waiting on this PR to get their timers integrated. If there are no changes required, it would be good to get this in soon.

rljacob:
Skipped testing since code is not built.

PR #6774, opened 2024-11-23
EAMxx: fix ghci-snl-cpu standalone machine setting

bartgol:
We were missing the gator initial mb env var, which caused the valgrind build to run out of memory.

### DIFF
--- a/components/eamxx/scripts/machines_specs.py
+++ b/components/eamxx/scripts/machines_specs.py
@@ -214,6 +214,7 @@ class GHCISNLCPU(Machine):
     def setup(cls):
         super().setup_base("ghci-snl-cpu")
         cls.baselines_dir = "/projects/e3sm/baselines/scream/ghci-snl-cpu"
+        cls.env_setup = ["export GATOR_INITIAL_MB=4000MB"]
 
 ###############################################################################
 class GHCISNLCuda(Machine):


bartgol:
@jgfouca I manually triggered the memcheck nightlies from this branch, and it does appear to fix things. See [here](https://github.com/E3SM-Project/E3SM/actions/runs/11983342059/job/33412696280) for the log, or cdash for the fresh report.

The valg build still has the issue of generating an empty suppressions file, but maybe we can work around that. E.g., we could store a supp file inside the container or in the repo, and call it a day. However, in the midst of MPI false positives, I think I saw some potentially real errors coming from IOP, p3, and maybe Homme. I just skimmed, so I am not 100% sure, and I have to call it a day (a week actually), but I will continue on Monday.

I do think we should merge this PR though, so we start to get the valg build to report to cdash.

rljacob:
See this error in the CI:
```
[ci (SMS_D_Ln5_P4.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.ghci-oci_gnu)](https://github.com/E3SM-Project/E3SM/actions/runs/11983339941/job/33428932378#step:5:1)
Value cannot be null. (Parameter 'ContainerId')
```

PR #6776, opened 2024-11-24
EAMxx: add horizontal contraction utility to Field

mahf708:
adds a horizontal contraction utility equivalent to einsum('i,i...k->...k', weight, field)

---

Notes:
- This utility is a reworked impl of https://github.com/E3SM-Project/scream/pull/3084 based on the feedback there
- In a future PR, this utility will enable a global horiz reducer (validated PR in https://github.com/E3SM-Project/scream/pull/3068 in response to issue https://github.com/E3SM-Project/scream/issues/2738)
- In potential future PRs, a similar utility (level reducer) can replace several repeated code impls of similar ops (e.g., WaterPaths, NumberPaths, OpticalDepths)


components/eamxx/src/share/field/field_utils_impl_colred.hpp
```diff
@@ -0,0 +1,123 @@
...
+namespace impl {
+
+// Utility to compute the reduction of a field along its column dimension.
+// This is equivalent to einsum('i,i...k->...k', f1, f2); i is the column.
+// The layouts are such that:
+// - The first dimension is for the columns (COL)
+// - There can be only up to 3 dimensions
+
+template <typename ST>
+Field column_reduction(const Field &f1, const Field &f2, const ekat::Comm *co) {
```
bartgol, line 18: Some comments on the fcn interface:

- I would use a more self-documenting name than f1/f2. Maybe you could replace f1->weight (and f2 with just f)?
- I don't think you need a comm in this routine, so you can just prune it.
- Can we offer 2 versions? One like this one, and one that takes an already allocated output field. The former would create the field an call the latter, but the latter can be more efficiently used at runtime, to avoid allocs.

components/eamxx/src/share/field/field_utils_impl_colred.hpp
```diff
@@ -0,0 +1,123 @@
...
+  const auto &g2 = f2.get_header().get_identifier().get_grid_name();
+
+  EKAT_REQUIRE_MSG(l2.rank() <= 3,
+                   "Error! Second field f2 must be at most rank-3.\n"
+                   "The input f2 rank is "
+                       << l2.rank() << ".\n");
+  EKAT_REQUIRE_MSG(l2.tags()[0] == COL,
+                   "Error! Second field f2 must have a column dimension.\n"
+                   "The input f2 layout is "
+                       << l2.tags() << ".\n");
```
bartgol, line 49: If this works, it is b/c of some coincidence. You are calling `operator<<` on a `std::vector<FieldTag>`. There is no official overload of op<< for vector. You could however just do `layout is: l2.to_string() << "\n"`.

components/eamxx/src/share/field/field_utils_impl_colred.hpp
```diff
@@ -0,0 +1,123 @@
...
+      auto vo = fo.get_view<ST>();
+      Kokkos::parallel_reduce(
+          fo.name(), RangePolicy(0, d0),
+          KOKKOS_LAMBDA(const int i, ST &ls) { ls += v1(i) * v2(i); }, vo);
+    } break;
+    case 2: {
+      auto v2      = f2.get_view<const ST **>();
+      auto vo      = fo.get_view<ST *>();
+      const int d1 = l2.dim(1);
+      auto p       = ESU::get_default_team_policy(d1, d0);
```
bartgol, line 81: It seems that the team policies are backward here. `nlev` should be the inner loop size, while the product of the remaining dims should be the outer loop size. Perhas you mixed up some stuff from the global horiz integral diag from the other branch you're working on?

components/eamxx/src/share/field/field_utils_impl_colred.hpp
bartgol: But you have to, before we merge it, right? There's not much sense to the integral otherwise...

components/eamxx/src/share/field/field_utils_impl_colred.hpp
bartgol: I usually review "commits since my last review", so you can go bigger now. Also, while I agree on keeping PRs manageable for the reviewer, I don't want to integrate half-features, so if it needs 300 more lines to be complete, so be it. Also, if the 100s of lines are all in the same 2-3 files, it's much easier to follow than a PR of just 200 lines, but scattered across 10-20 files...

components/eamxx/src/share/field/field_utils.hpp
```diff
@@ -111,6 +111,113 @@ void perturb (const Field& f,
...
+  const auto &n_in  = f_in.get_header().get_identifier().name();
+  const auto &u_in  = f_in.get_header().get_identifier().get_units();
+  const auto &g_in  = f_in.get_header().get_identifier().get_grid_name();
+
+  // If weight is not provided, we set it as a field of ones
+  Field wt;
+  if(weight) {
+    wt = *weight;
+  } else {
+    FieldIdentifier wt_id(n_in + "_weight", {{COL}, {l_in.dim(0)}}, u_in, g_in);
```
bartgol, line 156: Uhm, I am not fond of this approach, since it involves a runtime allocation. Do you think it'd be too complicated to handle the case without weight without creating a dummy weight field at runtime?

bartgol: At this point we could maybe make weight mandatory. My thought was to avoid having to use a weight at all, but if it's created anyways, we may as well force the user to to it, and make sure there's no hidden runtime alloc.

bartgol: Along the line of optional-vs-mandatory args: should we make comm also a required input? I don't really foresee using this function without it...

components/eamxx/src/share/field/field_utils.hpp
```diff
@@ -111,6 +111,113 @@ void perturb (const Field& f,
...
+          << l_w.dim(0)
+          << " while "
+             "the input field has dimension "
+          << l_in.dim(0) << ".\n");
+  EKAT_REQUIRE_MSG(
+      l_in.dim(0) > 0,
+      "Error! The input field must have a non-zero column dimension.\n"
+      "The input f_in layout is "
+          << l_in.to_string() << ".\n");
+  EKAT_REQUIRE_MSG(
```
bartgol, line 195: This check is included in the next one, so you can prune it, if you want.

bartgol (approved):
Looks good. I have some last minute thoughts, feel free to reject.

components/eamxx/src/share/field/field_utils.hpp
bartgol: I'm ok forcing to pass the weight, since this is likely only going to be used for horizontal integrals, where an area weight will be needed (or, at least, that'll be the most important scenario).

As for the comm, I kind of regret making it optional in the other fcns (max/min/norm), so if that's your motivation for making it optional (consistency), I would make it mandatory.

bartgol (approved):

bartgol:
The failing test is the same as on master (and is currently under investigation by the MAM4xx team). We can merge.

PR #6779, opened 2024-11-25
EAMxx: Removes all fortran from SHOC

jgfouca:
Following the pattern of a recent P3 PR, this PR changes the bfb unit tests to be based on baseline files instead of fortran calls. Unlike in that PR, we are actually able to completely remove all dependence on fortran code here because SHOC doesn't need any f90 init routines.

Change list:
1) Main change: For all the BFB SHOC unit tests, use baseline files instead of calling fortran!
2) Remove all of the CXX->f90 bridge code
3) Cleanup of shoc_init situation. SHOC has no global data, so you only need to call it if you need to compute npbl
4) Reorg by moving some less-interesting testing stuff into shoc/tests/infra. Now, the files presented in the public interface (just shoc directory) is a lot cleaner.
5) All shoc tests (bfb and run_and_cmp) accept the same set of arguments (-c for compare, -g for generate, -b for baseline dir, and -n for no baselines).
6) Some of the tests that just confirm that DIFFs are caught do need to do thread count spreads.
7) Remove the ability of shoc.F90 to call CXX impls..
8) gen-boiler can now generate _host functions for function that have array data.
9) No longer need to worry about transposing array data, we never cross language boundary anymore.
10) There were a lot of f90 functions that had a property test but were never actually used in the CXX. I removed all of these:
-    shoc_impli_sfc_fluxes_tests.cpp
-    shoc_impli_srf_stress_tests.cpp
-    shoc_impli_srf_tke_tests.cpp
-    shoc_energy_total_fixer_tests.cpp
-    shoc_energy_dse_fixer_tests.cpp
-    shoc_energy_threshold_fixer_tests.cpp
-    shoc_fterm_input_third_moms_tests.cpp
-    shoc_fterm_diag_third_moms_tests.cpp
-    shoc_omega_diag_third_moms_tests.cpp
-    shoc_xy_diag_third_moms_tests.cpp
-    shoc_aa_diag_third_moms_tests.cpp
-    shoc_w3_diag_third_moms_tests.cpp

[BFB]

------

Moved over from: https://github.com/E3SM-Project/scream/pull/3121

P3 DIFFs are expected because I caught a few places that were transposing views that didn't need to do that anymore. SHOC fails are expected because the baselines haven't been generated for the containers yet.

Follow on work:
* Port P3 init code from fortran; this is the last of the fortran being used by either p3 or shoc
* Remove all references to "fortran" or "Fortran". These names are all obsolete.


bartgol (approved):

bogensch (approved):

tcclevenger (approved):

components/eamxx/src/physics/shoc/tests/infra/shoc_ic_cases.cpp
```diff
@@ -16,11 +16,10 @@ namespace {
 // top and then flips everything, so we do the same.
 //------------------------------------------------------------------------
 
-using KT     = KokkosTypes<HostDevice>;
 using Scalar = Real;
-using Array1 = typename KT::template lview<Scalar*>;
-using Array2 = typename KT::template lview<Scalar**>;
-using Array3 = typename KT::template lview<Scalar***>;
+using Array1 = typename FortranData::Array1;
```
bartgol, line 20: What's the meaning of "FortranData" here? Should this be maybe called something like "BaselineData" or is there still some Fortran reminescence?

jgfouca: @bartgol , I made note of those in the follow-on work. These will be renamed but i don't think it's worth doing another CI run on this PR.

rljacob:
This should have been tested overnight on next before merging to mater since it touched code outside of eamxx.

bartgol:
> This should have been tested overnight on next before merging to mater since it touched code outside of eamxx.

You're right that it touches one file outside of eamxx, but I think this is somewhat of a special case. All the code removed from `shoc.F90` was code guarded by `ifdef SCREAM_CONFIG_IS_CMAKE`, which is _only_ defined if eamxx is built. All EAM cases (including MMF) should NOT be affected by that.

whannah1:
@bartgol 
```
> grep "SCREAM_CONFIG_IS_CMAKE" components/eam/src/physics/crm/pam/* -r
components/eam/src/physics/crm/pam/external/physics/scream_cxx_interfaces/CMakeLists.txt:add_definitions(-DSCREAM_CIME_BUILD -DHAVE_MPI -DSCREAM_CONFIG_IS_CMAKE)
```

bartgol:
> @bartgol
> 
> ```
> > grep "SCREAM_CONFIG_IS_CMAKE" components/eam/src/physics/crm/pam/* -r
> components/eam/src/physics/crm/pam/external/physics/scream_cxx_interfaces/CMakeLists.txt:add_definitions(-DSCREAM_CIME_BUILD -DHAVE_MPI -DSCREAM_CONFIG_IS_CMAKE)
> ```

Ah, crap, I only ran `git grep` without `--recurse-submodules`... One of the reasons why I sometimes don't like submods.

whannah1:
yea - and there's a list of reasons why pulling PAM into the repo makes sense. I hope to do that in the next year or so, but it depends on what we do with YAKL, among other things.

bartgol:
That said, I think the code that Jim purged from shoc.F90 is ONLY needed for eamxx's ability to call f90 implementation as part of the CXX-vs-F90 bfb testing. If PAM uses scream cxx impl, it should still be able to build with this PR merged.

whannah1:
@bartgol yea, it's an easy fix for PAM to remove the include statements, I created a PR yesterday.

rljacob:
@bartgol your request for an exception in the future is denied.

bartgol:
> @bartgol your request for an exception in the future is denied.

Tbc, I wasn't requesting exceptions. I was just pointing out that the collateral damage in this case should be minimal.

PR #6783, opened 2024-11-27
Workflows: remove mkatmsrffile test

mahf708:
Removes an unused test so that it is not accidentally triggered, and we also don't want to maintain it.


rljacob:
I see you added this as part of https://github.com/E3SM-Project/E3SM/pull/6506.  Sure you want to remove it?  It seems to have all the safeguards around it: only runs when the code is touched and only in this repo.

rljacob (approved):

PR #6786, opened 2024-11-28
Flag to control explicit mixing of aerosols

singhbalwinder:
A flag is added to control the explicit mixing of aerosols in dropmixnuc. Since interstitial aerosols and droplet number mixing ratios are mixed by SHOC, we added a flag to control explicit mixing by dropmixnuc to avoid double mixing of interstitial aerosols and droplet number.

Please note that the flag is set to false by default so that dropmixnuc does not mix interstitial aerosols and droplet number.


singhbalwinder:
Issued PR by mistake, closing.

PR #6787, opened 2024-11-29
EAMxx: Create IOPForcing ATM process

tcclevenger:
Moves IOP forcing from within HOMME interface to it's own physics process, called directly after dynamics process. Also, changes name of `IntensiveObservationPeriod` class to `IOPDataManager`.

I expect this to be non-BFB, at the very least since we are changing the order in which IOP forcing is applied to FM fields.


components/eamxx/src/physics/iop_forcing/eamxx_iop_forcing_process_interface.cpp
```diff
@@ -0,0 +1,546 @@
...
+
+  // Compute number of buffer views
+  const auto iop_nudge_tq = m_iop_data_manager->get_params().get<bool>("iop_nudge_tq");
+  const auto iop_nudge_uv = m_iop_data_manager->get_params().get<bool>("iop_nudge_uv");
+  if (iop_nudge_tq) m_buffer.num_1d_scalar_nlev += 2;
+  if (iop_nudge_uv) m_buffer.num_1d_scalar_nlev += 2;
+}
+// =========================================================================================
+void IOPForcing::
+set_computed_group_impl (const FieldGroup& group)
```
tcclevenger, line 46: The fix for the memory fails in https://github.com/E3SM-Project/scream/pull/3089 was adding back the `IOPForcing::set_computed_group_impl()` function which sets `m_num_tracers` before the buffer attempts to compute the number of bytes needed.

bartgol: Then, we could have also just done
```c++
m_num_tracers = get_group_in("tracers").m_info->size();
```
at the top of `requested_buffer_size_in_bytes`. But doing it in here is arguably cleaner.

components/eamxx/src/physics/iop_forcing/eamxx_iop_forcing_process_interface.cpp
```diff
@@ -0,0 +1,546 @@
...
+                         const int nlevs,
+                         const Real dt,
+                         const view_1d<const Pack>& divT,
+                         const view_1d<const Pack>& divq,
+                         const view_1d<Pack>& T,
+                         const view_1d<Pack>& qv)
+{
+  const auto nlev_packs = ekat::npack<Pack>(nlevs);
+  Kokkos::parallel_for(Kokkos::TeamVectorRange(team, nlev_packs), [&] (const int k) {
+    T(k).update(divT(k), dt, 1.0);
```
bartgol, line 288: I remember there was an issue with state update, which had an extra NUM_MPI_TASKS factor somewhere. I assume that has been fixed?

components/eamxx/src/physics/iop_forcing/eamxx_iop_forcing_process_interface.cpp
```diff
@@ -0,0 +1,546 @@
...
+    const auto u_mean  = m_buffer.u_mean;
+    const auto v_mean  = m_buffer.v_mean;
+
+    const auto qv_mean_h = Kokkos::create_mirror_view(qv_mean);
+    const auto t_mean_h  = Kokkos::create_mirror_view(t_mean);
+    const auto u_mean_h  = Kokkos::create_mirror_view(u_mean);
+    const auto v_mean_h  = Kokkos::create_mirror_view(v_mean);
+
+    const auto num_global_cols = m_grid->get_num_global_dofs();
+    for (int k=0; k<m_num_levs; ++k) {
```
bartgol, line 451: Have you thought about doing an outer || for? Yes, the access pattern on GPU is screwed up, but that's the same here. You may as well throw in another layer of ||ism, and do all calculations on device. It will at least save you from creating the mirrot view and do deep copies after the vertical for loop...

components/eamxx/src/physics/iop_forcing/eamxx_iop_forcing_process_interface.cpp
```diff
@@ -0,0 +1,546 @@
...
+      }
+      if (iop_nudge_uv){
+        Real& u_mean_k = u_mean_h(k/Pack::n)[k%Pack::n];
+        Real& v_mean_k = v_mean_h(k/Pack::n)[k%Pack::n];
+        Kokkos::parallel_reduce("compute_domain_means_uv",
+                                m_num_cols,
+                                KOKKOS_LAMBDA (const int icol, Real& u_sum, Real& v_sum) {
+          u_sum += horiz_winds(icol, 0, k/Pack::n)[k%Pack::n];
+          v_sum += horiz_winds(icol, 1, k/Pack::n)[k%Pack::n];
+        }, u_mean_k, v_mean_k);
```
bartgol, line 476: I think you may need a fence. here. I'm not 100% sure though, the ||reduce may already put one in for you. But we may as well be safe and make sure you don't start the MPI ops without finishing the kernel.

components/eamxx/src/physics/iop_forcing/eamxx_iop_forcing_process_interface.cpp
```diff
@@ -0,0 +1,546 @@
...
+
+        m_comm.all_reduce(&u_mean_k, 1, MPI_SUM);
```
bartgol, line 478: Another advantage of doing the ||for: you would only fire off ONE all reduce, at the end.

bartgol: Tbf, you could already do that, by moving the all reduce (and subsequent rescaling) to _after_ the end of the k loop...

bartgol: Forget all that I said. You could replace the buffers views with Field structures, and use the Field utility "horiz_contraction" that @mahf708 implemented (and that we merged recently). That way, you can just do something like
```c++
auto horiz_winds_f = get_field_out("horiz_winds");
auto area = m_grid->get_geo_data("area");
horiz_contraction(horiz_winds_mean,horiz_winds_f, avg_horiz_wind, area,get_comm());
avg_horiz_wind.scale(earth_area); // earth_area can be precomputed at init);
```
This is assuming that `horiz_winds_mean` has layout (2,nlevs);

bartgol (commented):
I have one comment regarding using the Field utility, but other than that, looks good.

components/eamxx/src/physics/iop_forcing/eamxx_iop_forcing_process_interface.cpp
tcclevenger: Yeah, this was an issue with the domain means below, where I was not using the global number of columns previously.

components/eamxx/src/physics/iop_forcing/eamxx_iop_forcing_process_interface.cpp
tcclevenger: A ||reduce with a scalar reduction type implies a fence (a rank 0 view reduction type does not imply a fence), but I'm not sure if that is guaranteed to always be the behavior. Would definitely be safer to fence.

components/eamxx/src/physics/iop_forcing/eamxx_iop_forcing_process_interface.cpp
tcclevenger: The reason I did not is that we do an MPI_AllReduce within the loop and I assumed that couldn't be done on device. Is that a correct assumption?

Anyways, I could save that computation for outside the nest ||for/reduce and all reduce on the entire view, but I'll still need the mirrors I think?

components/eamxx/src/physics/iop_forcing/eamxx_iop_forcing_process_interface.cpp
tcclevenger: Let me check that out!

components/eamxx/src/physics/iop_forcing/eamxx_iop_forcing_process_interface.cpp
bartgol: You could do the rescaling at the end of the nested for/reduce, no? Since the scaling is num_global_cols for all entries, no?

components/eamxx/src/physics/iop_forcing/eamxx_iop_forcing_process_interface.cpp
tcclevenger: I think the num_tracers is only used for allocating and initializing the buffer in this AP, so maybe it would make more sense there and reduce code. I chose to bring back the function because it _could_ be used elsewhere in the class and then would make more sense IMO to be initialized in a group specific virtual function and not the buffer since I can't be guaranteed the buffer init stuff is called in the right spot for some other usage in the class.

components/eamxx/src/physics/iop_forcing/eamxx_iop_forcing_process_interface.cpp
tcclevenger: Yes, but wouldn't I still need the host mirrors to do the all reduce? And I guess the scaling as well?

bartgol: Yeah, you would, unless SCREAM_MPI_ON_DEVICE is defined. But anyways, I think the convo regarding the Field utility below may be overriding this one..

bogensch:
@mahf708 this is a good point and I think it's probably possible to eventually merge the IOP nudging routine to the EAMxx general one.  However, I'd prefer to punt this to a different PR because I don't think it's trivial.  For instance, the IOP nudging routine nudges the domain mean, not the individual column explicitly.  Secondly, either the EAMxx nudging routine would need to be modified to support reading in the format of the IOP nudging OR the DPxx workflow and input files would need to be substantially revised.  Just two things off the top of my head that come to mind.

tcclevenger:
@bogensch Can you test one more time before we merge? Particularly something that uses `iop_nudging_uv/tq`?

bogensch:
@tcclevenger I ran a case that nudges U & V winds by default.  The run crashed about six hours in.  I ran the case with nudging turned off and the case ran fine.

For the case with nudging that crashed I noticed that the U and V being output were about four orders of magnitudes larger than expected.

tcclevenger:
@bogensch Thanks, definitely relate to my last round of changes. I'll investigate.

tcclevenger:
@bogensch Were you testing on CPU? I think this was an issue with how I was using (or not using) Packs.

bogensch:
@tcclevenger yes, my testing was done on pm-cpu.  I did not test on pm-gpu.

tcclevenger:
@bogensch Can you test again?

bogensch:
@tcclevenger my tests on both pm-cpu and pm-gpu look good!  Thus, this gets my seal of approval.

bogensch (approved):

components/eamxx/src/physics/iop_forcing/eamxx_iop_forcing_process_interface.cpp
```diff
@@ -0,0 +1,520 @@
...
+
+    if (iop_coriolis) {
+      // Apply coriolis forcing to u and v winds
+      iop_apply_coriolis(team, num_levs, dt, target_lat, u_ls, v_ls, u_i, v_i);
+    }
+
+    // Release WS views
+    ws.release_many_contiguous<3>({&ref_p_mid, &ref_p_int, &ref_p_del});
+  });
+  Kokkos::fence();
```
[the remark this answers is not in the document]
tcclevenger: Let me check, not sure why I had initially included.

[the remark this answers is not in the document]
tcclevenger: I think I included previously because I was observing some values later on host, so I needed to ensure that all my computation from the previous kernel was complete. Now that everything is on device it should be ok to remove.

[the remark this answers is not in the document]
tcclevenger: Tested on GPU and fence is not needed. Removed.

bartgol (approved):

PR #6788, opened 2024-11-30
EAMxx: add horizontal average diagnostic field

mahf708:
Adds an online diagnostic field for area-weighted horizontal average, following #6776.

---

Users are now able to request some_field_horiz_avg in their output files to get an area-weighted horizontal average. The following types of fields are supported:

| input dimensions | example               | output dimensions |
| ---------------- | -------               | ----------------- |
| (COL)            | area_horiz_avg        | () (scalar)       |
| (COL, LEV)       | T_mid_horiz_avg       | (LEV)             |
| (COL, CMP, LEV)  | horiz_winds_horiz_avg | (CMP, LEV)        |

### DIFF
--- a/components/eamxx/src/diagnostics/tests/horiz_avg_test.cpp
+++ b/components/eamxx/src/diagnostics/tests/horiz_avg_test.cpp
@@ -1,0 +1,197 @@
+#include "catch2/catch.hpp"
+#include "diagnostics/register_diagnostics.hpp"
+#include "share/field/field_utils.hpp"
+#include "share/grid/mesh_free_grids_manager.hpp"
+#include "share/util/scream_setup_random_test.hpp"
+#include "share/util/scream_universal_constants.hpp"
+
+namespace scream {
+
+std::shared_ptr<GridsManager> create_gm(const ekat::Comm &comm, const int ncols,
+                                        const int nlevs) {
+  const int num_global_cols = ncols * comm.size();
+
+  using vos_t = std::vector<std::string>;
+  ekat::ParameterList gm_params;
+  gm_params.set("grids_names", vos_t{"Point Grid"});
+  auto &pl = gm_params.sublist("Point Grid");
+  pl.set<std::string>("type", "point_grid");
+  pl.set("aliases", vos_t{"Physics"});
+  pl.set<int>("number_of_global_columns", num_global_cols);
+  pl.set<int>("number_of_vertical_levels", nlevs);
+
+  auto gm = create_mesh_free_grids_manager(comm, gm_params);
+  gm->build_grids();
+
+  return gm;
+}
+
+TEST_CASE("horiz_avg") {
+  using namespace ShortFieldTagsNames;
+  using namespace ekat::units;
+  using TeamPolicy = Kokkos::TeamPolicy<Field::device_t::execution_space>;
+  using TeamMember = typename TeamPolicy::member_type;
+  using KT         = ekat::KokkosTypes<DefaultDevice>;
+  using ESU        = ekat::ExeSpaceUtils<typename KT::ExeSpace>;
+
+  // A numerical tolerance
+  auto tol = std::numeric_limits<Real>::epsilon() * 100;
+
+  // A world comm
+  ekat::Comm comm(MPI_COMM_WORLD);
+
+  // A time stamp
+  util::TimeStamp t0({2024, 1, 1}, {0, 0, 0});
+
+  // Create a grids manager - single column for these tests
+  constexpr int nlevs = 3;
+  constexpr int dim3  = 4;
+  const int ngcols    = 6 * comm.size();
+
+  auto gm   = create_gm(comm, ngcols, nlevs);
+  auto grid = gm->get_grid("Physics");
+
+  // Input (randomized) qc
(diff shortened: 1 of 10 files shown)


AaronDonahue:
@mahf708 , thanks for re-issuing this PR on the E3SM side.  Quick question, there was this discussion between Peter B. and Andrew (https://github.com/E3SM-Project/scream/issues/2738#issuecomment-2356851376) about the algorithm for the horizontal average that takes into account mass.  Any thoughts on whether this could/should be incorporated into this PR?

bogensch:
I typically always neglect mass when doing horizontal averages in DP.  Unless @ambrad has a strong opinion about including this I'm fine with keeping it area weighted.

bartgol:
@AaronDonahue not to digress the PR convo (maybe we can move to slack), but if there's the need to compute "mass" quantities (e.g., dp*some_tracer), we could add a diagnostic `mass_X`, which a) ensures X is a 3d quantity, and b) computes X(icol,ilev)*dp(icol,ilev). Then, one can request `mass_qv_horiz_avg`, where IO will first compute `mass_qv`, and then the corresponding horiz contraction.

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -1446,6 +1446,12 @@ AtmosphereOutput::create_diagnostic (const std::string& diag_field_name) {
     } else {
       params.set<std::string>("Temperature Kind", "Tot");
     }
+  } else if (diag_field_name.find("_horiz_avg")!=std::string::npos) {
```
bartgol, line 1449: You put this at the very last "else if" slot, so it should be ok. Nevertheless, I would like to guard against any possible future code change. In particular, if one adds handling of  "_vert_avg" , and puts an "else if" below yours, we could have a problem. Say the user asks for "some_tracer_horiz_avg_vert_avg" (yes, that's just a global absolute avg, but whatever). Your code will now just compute "some_tracer_horiz_avg", losing completely the "_vert_avg" part.

To guard against this, you can consider only the case where the field name _ends_ with "_horiz_avg". C++ doesn't have an `ends_with` method for std::string like in python (we impl-ed `starts_with` in ekat, maybe we should impl `ends_with` as well), so you'd have to roll your own. Or maybe you could do
```c++
else if (diag_field_name.find("_horiz_avg")!=std::string::npos &&
            ekat::split(diag_field_name,"_horiz_avg")[1]=="")
```
although this relies on `ekat::split` adding an empty string at the end if the string ends with the separation token.

bartgol (commented):
Looks good. I have a couple of comments, one for testing, and one for the IO mods.

components/eamxx/src/share/io/scorpio_output.cpp
bartgol: I realized that something like `some_tracer_horiz_avg_at_500hPa` is prob a better example for this issue.

bartgol: Actually, I take it back. `some_tracer_horiz_avg_at_500hPa` _should_ work now, since we first parse the "_at_500hPa" case, using whatever is before as the field to give to FieldAtPressureLevel. It happens to be a diag, so it creates nested diags.

Again, as of now, everything _should_ work, since the horiz_avg diag is the last one you consider. However, we should still make sure that adding a new `else if` after to handle a new diag would not mess up parsing of the diag name.

bartgol: Ah, good point. We do the same mistake with all the diags. For your diag, we can easily check that "_horiz_avg" is the last token, so that's easy. But for "blah_at_500hPa", we don't check that there's nothing after "hPa".

We can merge this, and fix ALL diags handling later, or we can fix yours here, and do a separate PR to fix other diag names that involve special tokens.

bartgol: I'm working on a cleaner solution regarding diags creation...

components/eamxx/src/diagnostics/tests/horiz_avg_test.cpp
```diff
@@ -0,0 +1,197 @@
...
+        Real sum      = sp(0.0);
+        Kokkos::parallel_reduce(
+            Kokkos::TeamThreadRange(m, ngcols),
+            [&](const int icol, Real &accum) {
+              accum += (area_v(icol) / atot) * qc3_v(icol, j, k);
+            },
+            sum);
+        Kokkos::single(Kokkos::PerTeam(m),
+                       [&]() { diag3_manual_v(j, k) = sum; });
+      });
```
bartgol, line 189: I have one possible suggestion, feel free to discard.

What about using horiz_contraction in the test as well? It may make for a much easier/shorter test to maintain...

bartgol (approved):
Looks good. I have one suggestion, but minor.

bartgol (approved):

bartgol:
> @bartgol are you comfortable running the tests and then integrating this PR?

Yeah, sorry. I didn't click on rerun cause mappy was having flu symtpoms. But Jim gave him some good stuff, and now it's back to work. I'll retest.

PR #6789, opened 2024-12-02
EAMxx: Give option for for tracers to be advected only by dynamics

tcclevenger:
Processes can choose for a tracer to be advected only by dynamics (not by SHOC) by passing `turbulence_advected=false` to `add_tracer()`. Guards set up to ensure all tracers requests are consistent among processes.

Moved over from https://github.com/E3SM-Project/scream/pull/3085. Currently working on issue of tracer ordering between PG2 and GLL grids.

[non-BFB for MAMxx]

### DIFF
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -58,9 +58,7 @@ namespace control {
  *     Note: at this stage, atm procs that act on non-ref grid(s) should be able to create their
  *           remappers. The AD will *not* take care of remapping inputs/outputs of the process.
  *  4) Register all fields and all groups from all atm procs inside the field managers, and proceed
- *     to allocate fields. Each field manager (there is one FM per grid) will take care of
- *     accommodating all requests for packing as well as (if possible) bundling of groups.
- *     For more details, see the documentation in the share/field/field_request.hpp header.
+ *     to allocate fields. For more details, see the documentation in the share/field/field_request.hpp header.
  *  5) Set all the fields into the atm procs. Before this point, all the atm procs had were the
  *     FieldIdentifiers for their input/output fields and FieldGroupInfo for their input/output
  *     field groups. Now, we pass actual Field and FieldGroup objects to them, where both the
@@ -405,7 +403,7 @@ void AtmosphereDriver::reset_accumulated_fields ()
     }
 
     auto accum_group = m_field_mgr->get_field_group("ACCUMULATED", grid_name);
-    for (auto f_it : accum_group.m_fields) {
+    for (auto f_it : accum_group.m_individual_fields) {
       auto& track = f_it.second->get_header().get_tracking();
       f_it.second->deep_copy(zero);
       track.set_accum_start_time(m_current_ts);
@@ -645,8 +643,8 @@ void AtmosphereDriver::create_fields()
     m_field_mgr->add_to_group(fid, "RESTART");
   }
   for (const auto& g : m_atm_process_group->get_groups_in()) {
-    if (g.m_bundle) {
-      m_field_mgr->add_to_group(g.m_bundle->get_header().get_identifier(), "RESTART");
+    if (g.m_monolithic_field) {
+      m_field_mgr->add_to_group(g.m_monolithic_field->get_header().get_identifier(), "RESTART");
     } else {
       for (const auto& fn : g.m_info->m_fields_names) {
         m_field_mgr->add_to_group(fn, g.grid_name(), "RESTART");
@@ -1174,19 +1172,19 @@ void AtmosphereDriver::set_initial_conditions ()
   // ...then the input groups
   m_atm_logger->debug("    [EAMxx] Processing input groups ...");
   for (const auto& g : m_atm_process_group->get_groups_in()) {
-    if (g.m_bundle) {
-      process_ic_field(*g.m_bundle);
+    if (g.m_monolithic_field) {
+      process_ic_field(*g.m_monolithic_field);
     }
-    for (auto it : g.m_fields) {
+    for (auto it : g.m_individual_fields) {
       process_ic_field(*it.second);
     }
   }
   m_atm_logger->debug("    [EAMxx] Processing input groups ... done!");
 
-  // Some fields might be the subfield of a group's bundled field. In that case,
-  // we only need to init one: either the bundled field, or all the individual subfields.
+  // Some fields might be the subfield of a group's monolithic field. In that case,
+  // we only need to init one: either the monolithic field, or all the individual subfields.
   // So loop over the fields that appear to require loading from file, and remove
-  // them from the list if they are the subfield of a bundled field already inited
+  // them from the list if they are the subfield of a groups monolithic field already inited
   // (perhaps via initialize_constant_field, or copied from another field).
   for (auto& it1 : ic_fields_names) {
     const auto& grid_name =  it1.first;
@@ -1290,17 +1288,17 @@ void AtmosphereDriver::set_initial_conditions ()
   }
   m_atm_logger->debug("    [EAMxx] Processing fields to copy ... done!");
 
-  // It is possible to have a bundled group G1=(f1,f2,f3),
+  // It is possible to have a monolithically allocated group G1=(f1,f2,f3),
   // where the IC are read from file for f1, f2, and f3. In that case,
-  // the time stamp for the bundled G1 has not be inited, but the data
+  // the time stamp for the monolithic field of G1 has not be inited, but the data
   // is valid (all entries have been inited). Let's fix that.
   m_atm_logger->debug("    [EAMxx] Processing subfields ...");
   for (const auto& g : m_atm_process_group->get_groups_in()) {
-    if (g.m_bundle) {
-      auto& track = g.m_bundle->get_header().get_tracking();
+    if (g.m_monolithic_field) {
+      auto& track = g.m_monolithic_field->get_header().get_tracking();
       if (not track.get_time_stamp().is_valid()) {
-        // The bundled field has not been inited. Check if all the subfields
-        // have been inited. If so, init the timestamp of the bundled field too.
+        // The groups monolithic field has not been inited. Check if all the subfields
+        // have been inited. If so, init the timestamp of the monlithic field too.
         const auto& children = track.get_children();
         bool all_inited = children.size()>0; // If no children, then something is off, so mark as not good
         for (auto wp : children) {
@@ -1678,7 +1676,7 @@ void AtmosphereDriver::run (const int dt) {
     }
 
     auto rescale_group = m_field_mgr->get_field_group("DIVIDE_BY_DT", gname);
-    for (auto f_it : rescale_group.m_fields) {
+    for (auto f_it : rescale_group.m_individual_fields) {
       f_it.second->scale(Real(1) / dt);
     }
   }
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -535,8 +535,54 @@ void AtmosphereDriver::create_fields()
   m_field_mgr = std::make_shared<field_mgr_type>(m_grids_manager);
   m_field_mgr->registration_begins();
 
-  // By now, the processes should have fully built the ids of their
-  // required/computed fields and groups. Let them register them in the FM
+  // Before registering fields, check that Field Requests for tracers are compatible
+  {
+    // Create map from tracer name to a vector which contains the field requests for that tracer.
+    std::map<std::string, std::set<FieldRequest>> tracer_requests;
+    auto gather_tracer_requests = [&] (FieldRequest req) {
+      if (not ekat::contains(req.groups, "tracers")) return;
+
+      std::string fname = req.fid.name();
+      if (tracer_requests.find(fname) == tracer_requests.end()) {
+        tracer_requests[fname] = {req};
+      } else {
+        tracer_requests[fname].emplace(req);
+      }
+    };
+    for (const auto& req : m_atm_process_group->get_required_field_requests()){
+      gather_tracer_requests(req);
+    }
+    for (const auto& req : m_atm_process_group->get_computed_field_requests()) {
+      gather_tracer_requests(req);
+    }
+
+    // Go through the map entry for each tracer and check that every one
+    // has the same request for turbulence advection.
+    for (auto fr : tracer_requests) {
+      const auto reqs = fr.second;
+
+      std::set<bool> turb_advect_types;
+      for (auto req : reqs) {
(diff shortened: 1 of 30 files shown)


bartgol:
@tcclevenger Just to clarify: you said you're still working on ordering of tracers for PG2 and DYN grids, so is this PR not yet ready?

tcclevenger:
> @tcclevenger Just to clarify: you said you're still working on ordering of tracers for PG2 and DYN grids, so is this PR not yet ready?

Yes. Do we have a WIP equiv over here so I don't let the AT run while I work on getting a fix in?

bartgol:
> > @tcclevenger Just to clarify: you said you're still working on ordering of tracers for PG2 and DYN grids, so is this PR not yet ready?
> 
> Yes. Do we have a WIP equiv over here so I don't let the AT run while I work on getting a fix in?

No we don't. But as soon as you push commits, you can click on the checks tab, and cancel all the jobs (one workflow at a time, unfortunately), if you know you're not done anyways...

rljacob:
What if you change it to draft PR ?

bartgol:
> What if you change it to draft PR ?

I saw your reply just as I was editing my answer... Unfortunately, gh actions pull_request event still triggers, even if the PR is in draft mode. The only way to skip them is to do something like

```yaml
jobs:
  my-job:
    if: ${{ github.event_name != "pull_request" && github.event.pull_request.draft == "false" }}
    ...
```

rljacob:
@tcclevenger what is the status?

tcclevenger:
Still working on a fix for mismatch tracer indices between PG2 and GLL grid

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.hpp
```diff
@@ -212,6 +212,7 @@ class MAMAci final : public scream::AtmosphereProcess {
       // for atmosphere
       compute_vertical_layer_heights(team, dry_atm_pre_, i);
       compute_updraft_velocities(team, wet_atm_pre_, dry_atm_pre_, i);
+      set_min_background_mmr(team, dry_aero_pre_, i); //dry_atm_pre_ is the output
```
tcclevenger, line 215: @singhbalwinder I don't recognize this line, could this be something I inadvertently added back during a rebase?

tcclevenger: Sorry, ignore this, I remember you adding a fix in the original SCREAM PR.

rljacob:
@tcclevenger any progress on fix?

tcclevenger:
> @tcclevenger any progress on fix?

Yes, this is what I'm currently working on. Will update as I go.

ambrad:
@rljacob is it OK if we switch this to draft?

rljacob:
Sure thats fine.

ambrad:
@tcclevenger would you switch this to Draft if it's not ready? We'll try out that workflow.

tcclevenger:
> @tcclevenger please try to make this old PR a priority.

This is what I am currently working on, but the work is in a different branch (it contains a fundamental change to allocating groups of fields that doesn't make sense to throw onto this branch). Yes, I agree draft makes more sense for this PR.

tcclevenger:
Rebased to https://github.com/E3SM-Project/E3SM/pull/6981 for testing. Extra commits will go away once that is merged.

bartgol:
@tcclevenger you where expecting the diffs, right? I see only 3 fail though:

```
507:homme_shoc_cld_p3_mam_optics_rrtmgp_baseline_cmp
515:homme_shoc_cld_mam_aci_p3_mam_optics_rrtmgp_mam_drydep_baseline_cmp
523:homme_shoc_cld_spa_p3_rrtmgp_mam4_wetscav_baseline_cmp
```
while I think we do have more mam4xx tests. Were the other expected to not diff?

tcclevenger:
> @tcclevenger you where expecting the diffs, right? I see only 3 fail though:
> 
> ```
> 507:homme_shoc_cld_p3_mam_optics_rrtmgp_baseline_cmp
> 515:homme_shoc_cld_mam_aci_p3_mam_optics_rrtmgp_mam_drydep_baseline_cmp
> 523:homme_shoc_cld_spa_p3_rrtmgp_mam4_wetscav_baseline_cmp
> ```
> 
> while I think we do have more mam4xx tests. Were the other expected to not diff?

@bartgol Yes, those 3 tests are expected to DIFF. `SMS_D_Ln5.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.ghci-snl-cpu_gnu.eamxx-mam4xx-all_mam4xx_procs` DIFFs as well (expected). I'm working with @singhbalwinder to do some more in depth testing to make sure the new output is acceptable.

Seems that some of the other CIME cases are erroring out trying to start the container? Have we seen that elsewhere?

bartgol:
> Seems that some of the other CIME cases are erroring out trying to start the container? Have we seen that elsewhere?

I'm guessing the error is this one?
```
  toomanyrequests: {"errors":[{"code":"TOOMANYREQUESTS","message":"retry-after: 51.451683ms"}]}
```
I have never seen it. I suppose we had too many concurrent PRs that needed to be tested, and all the gh-hosted runners tried to access the container registry at the same time? Should be safe to re-run the failed tests manually.

ambrad (approved):
I approve as long as the "[non-BFB for MAMxx]" status remains true (I think it is, but just want to be sure).

singhbalwinder (approved):

components/eamxx/src/control/atmosphere_driver.cpp
```diff
@@ -531,8 +531,54 @@ void AtmosphereDriver::create_fields()
...
-  // required/computed fields and groups. Let them register them in the FM
+  // Before registering fields, check that Field Requests for tracers are compatible
+  {
+    // Create map from tracer name to a vector which contains the field requests for that tracer.
+    std::map<std::string, std::set<FieldRequest>> tracer_requests;
+    auto gather_tracer_requests = [&] (FieldRequest req) {
+      if (not ekat::contains(req.groups, "tracers")) return;
+
+      std::string fname = req.fid.name();
+      if (tracer_requests.find(fname) == tracer_requests.end()) {
```
bartgol, line 542: I think this check is pointless. The else branch should work regardless, no?

components/eamxx/src/control/atmosphere_driver.cpp
```diff
@@ -531,8 +531,54 @@ void AtmosphereDriver::create_fields()
...
+    for (const auto& req : m_atm_process_group->get_required_field_requests()){
+      gather_tracer_requests(req);
+    }
+    for (const auto& req : m_atm_process_group->get_computed_field_requests()) {
+      gather_tracer_requests(req);
+    }
+
+    // Go through the map entry for each tracer and check that every one
+    // has the same request for turbulence advection.
+    for (auto fr : tracer_requests) {
```
bartgol, line 557: Say MAM requests field A, specifying that it should not be turb advected. Another process uses field A, so it requests it, but it has no knowledge of what turb advection is, or at least it doesn't care (the choice may be relevant for MAM, but the rest of the atm doesn't care). In this case, wouldn't the two request conflict?

Maybe we should turn this into an enum, like
```c++
enum TracerAdvection {
  DynOnly,
  DynAndTurb,
  Default,
};
```
(the names are up for debate, ofc). In this case, the default would be `Default`. If all procs use "default", we do both dyn and turb advection, as before. If all procs use "default", but some ask for "DynOnly", we do dyn only. The error only arises if some procs expicitly ask for "DynOnly" and others explicitly ask for "DynAndTurb".

Maybe I'm over thinking it, but there is no way, with a bool, to check if some values for turb_advected_tracers are true b/c the customer _asked_ for true, or simply b/c they didn't care, and relied on the default.

bartgol: The enum may make things complicated, in which case we may also consider whether this check is needed. That is, as long as there is ONE request that says "don't do turb advection", then we don't add it to the turb advected tracers.

Maybe I need to better understand the case we are trying to guard ourselves from...

bartgol: OTOH, using enums in the `add_tracer` calls make the code more self-explanatory...so maybe switching to an enum is a bit clearer?
```c++
add_tracer<Required>("blah",units, grid,1,false);
```
vs
```c++
add_tracer<Required>("blah",units, grid,1,NoTurbAdvection);
```

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -111,7 +111,7 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
   } // Extra SHOC output diagnostics
 
   // Tracer group
-  add_group<Updated>("tracers", grid_name, ps, true);
+  add_group<Updated>("turbulence_advected_tracers", grid_name, ps, true);
```
bartgol, line 114: I know we just removed it, but in the spirit of self-documenting code: do you think we should bring back the enum for requesting bundling? I'm wondering if it's worth it (folks would read the line above with something like `BundlingNeeded` or `MonolithicField` or whatever instead of `true`), or if it's ok to assume folks would just go and see what the signature of `add_group` is...

bartgol (commented):
So there wasn't much to do in FieldManager? That's nice. With this branch, did you try to print the tracer group content, and verify that the non-turb-advected tracers are all at the end?

I have some thoughts on the design. Not really requests, just discussions sparks.

components/eamxx/src/control/atmosphere_driver.cpp
tcclevenger: I may be leaning towards getting rid of this check. I think (but will verify) that the changes that evolved in the FM since I wrote this check actually make it a bit redundant? It should error out with inconsistent requests, we just may not get the output about which processes conflict.

With regards to the enums, I may be having a change of heart. I've always avoided in the case where a bool was all that's needed, but these signatures definitely are less clear and `add_tracer` and `add_group` are certainly part of the code that users will interface with if they write a process. Will change them back.

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
tcclevenger: See comment above.

components/eamxx/src/control/atmosphere_driver.cpp
bartgol: Same here: I often though "why adding a new type when a bool flag does it?". But code clarity with enums is on a whole other level.

tcclevenger: Follow up: We need this check given how group requests do group logic, namely, if a single processes says "I want A in group G" then A gets added to G, even if some processes intentionally did not add A to G.

I think this option is more of "I want to explicitly state that A should not be in group G" and then these checks ensure no one added it.

tcclevenger:
> With this branch, did you try to print the tracer group content, and verify that the non-turb-advected tracers are all at the end?

Yes, for PG2, I verified GLL<->PG2 indices matched, and that non-turb-advected tracers came last.

components/eamxx/src/dynamics/homme/eamxx_homme_process_interface.cpp
```diff
@@ -180,7 +180,7 @@ void HommeDynamics::set_grids (const std::shared_ptr<const GridsManager> grids_m
   add_field<Computed>("omega",              pg_scalar3d_mid, Pa/s,  pgn,N);
 
   add_tracer<Updated >("qv", m_phys_grid, kg/kg, N);
-  add_group<Updated>("tracers",pgn,N, true);
+  add_group<Updated>("tracers",pgn,N, Bundling::Required);
```
bartgol, line 183: Since we're bringing the enum back, I may as well ask: do you think the name `Bundling` is clear? Maybe we could change it into something else. Some ideas:
```c++
enum class GroupAllocation {
  Contiguous,
  NotContiguous
};
enum class MonolithicGroup {
  Required,
  NotRequired
};
enum class ContiguousGroup {
  Requred,
  NotRequired
};
```
I don't know. I wonder if the word "bundle" is a bit obscure, and we can find something more self-explanatory? If you have other name ideas, feel free to bounce them off us. This is the right time to discuss this, before we roll it out again.

tcclevenger: The user potentially needs to know what `group.m_bundle` is, maybe we would need to think about changing that name too? 

I like Monolithic the best. To me non-contiguous implies that "I can access the entire group, it just won't be contiguous" where the only way to access a view of the entire group is bundled case.

tcclevenger:
Confirmed https://github.com/E3SM-Project/E3SM/pull/6981 did not introduce non-BFB fails on PM-GPU, so we can merge this once we settle on a "bundling" name

components/eamxx/src/dynamics/homme/eamxx_homme_process_interface.cpp
bartgol: Yes, if we decide on a name change, we should change it everywhere...

bartgol: What if we used
```c++
enum class MonolithicAlloc {
	Required,
	NotRequired
};
```
? I agree that the word "monolithic" carries a clear meaning. But I also like to have the word "allocation" (or something reminescent of it) somwhere.

What would you then call the bundle? `group.m_monolithic`? `group.m_monolithic_field`?  `group.m_whole`?  Something else? I like `m_monolithic_field`, and maybe rename `m_fields` into `m_individual_fields`?

Feel free to propose others...

bartgol: We can also merge as is, and follow up with a better name convention right after...

bartgol (approved):

components/eamxx/src/dynamics/homme/eamxx_homme_process_interface.cpp
tcclevenger: I like `m_monolithic_field` and `m_individual_fields`. I think this is much clearer than "bundling". I've updated all the comments that refer to bundling as well.

tcclevenger:
I decided to update bundling name instead of merging since I'm off work tomorrow and so don't want to merge tonight. Assuming there are no fails/change requests with the naming, I'll merge first thing next week (Sunday night, even, if I remember...).

@singhbalwinder Thanks for your patience with this PR! Ended up taking much more time than anticipated..

singhbalwinder:
No worries at all—I completely understand that these things can take much longer than expected. I appreciate the time and effort you’ve put into this. Merging next week is fine.

tcclevenger:
> Is this ready?

Yes, going to merge today.

rljacob:
Kokkos branch delayed so go ahead.

PR #6793, opened 2024-12-02
Delete .mergify.yml

bartgol:
While we tried to use this in the SCREAM repo to automize merging and to guard against accidental merges, we are not using it in the E3SM repo, and it is confusing at best.


rljacob (approved):

bartgol:
@rljacob any objection to merge this straight into both master and next, given that it does not touch anything that affects testing? I'll wait until all checks pass, for extra safety, but it seems pointless to wait for nightly tests...

rljacob:
go ahead.

PR #6759, opened 2024-11-19
Change SORRMv3 ice runoff mapping

darincomeau:
Changes the ice runoff mapping file (`ROF2OCN_ICE_RMAPNAME`) for the `SOwISC12to30E3r3` ocean mesh to version with increased smoothing.

New mapping file has been staged in public inputdata repo, world readable.

[BFB] except
[non-BFB] for cases with the SOwISC12to30E3r3 ocean mesh (not in current testing).

PR #6794, opened 2024-12-03
Change range of solid ice runoff removal from 60S to 57S

darincomeau:
Extends the northern range of solid ice runoff from Antarctica from 60S to ~~45S~~ 57S when the ocean's namelist option `config_remove_ais_ice_runoff = .true.`. The new solid ice runoff mapping with increased smoothing for the `SOwISC12to30E3r3` mesh in https://github.com/E3SM-Project/E3SM/pull/6759 has a small amount of runoff north of 60S, which needs to be removed when Antarctic solid ice runoff is intended to be removed.

[BFB]
All current configurations that do not use the mapping file in the above PR do not have solid ice runoff between 60S and ~~45S~~ 57S, so there is no impact.

### DIFF
--- a/components/mpas-ocean/driver/ocn_comp_mct.F
+++ b/components/mpas-ocean/driver/ocn_comp_mct.F
@@ -2181,7 +2181,7 @@ contains
                call shr_sys_abort ('Error: incoming rofi_F is negative')
            end if
            if (config_remove_ais_ice_runoff) then
-              if (latCell(i) < -0.78539816325_RKIND) then ! 45S in radians
+              if (latCell(i) < -0.99483767345_RKIND) then ! 57S in radians
                  removedIceRunoffFlux(i) = iceRunoffFlux(i)
                  iceRunoffFlux(i) = 0.0_RKIND
                  removedIceRunoffFluxThisProc = removedIceRunoffFluxThisProc + removedIceRunoffFlux(i)
@@ -3714,7 +3714,7 @@ contains
                call shr_sys_abort ('Error: incoming rofi_F is negative')
            end if
            if (config_remove_ais_ice_runoff) then
-              if (latCell(i) < -0.78539816325_RKIND) then ! 45S in radians
+              if (latCell(i) < -0.99483767345_RKIND) then ! 57S in radians
                  removedIceRunoffFlux(i) = iceRunoffFlux(i)
                  iceRunoffFlux(i) = 0.0_RKIND
                  removedIceRunoffFluxThisProc = removedIceRunoffFluxThisProc + removedIceRunoffFlux(i)
--- a/components/mpas-ocean/driver/ocn_comp_mct.F
+++ b/components/mpas-ocean/driver/ocn_comp_mct.F
@@ -2181,7 +2181,7 @@ contains
                call shr_sys_abort ('Error: incoming rofi_F is negative')
            end if
            if (config_remove_ais_ice_runoff) then
-              if (latCell(i) < -1.04719666667_RKIND) then ! 60S in radians
+              if (latCell(i) < -0.78539816325_RKIND) then ! 45S in radians
                  removedIceRunoffFlux(i) = iceRunoffFlux(i)
                  iceRunoffFlux(i) = 0.0_RKIND
                  removedIceRunoffFluxThisProc = removedIceRunoffFluxThisProc + removedIceRunoffFlux(i)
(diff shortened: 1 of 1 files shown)


darincomeau:
>Do we ever have ice runoff in places like Patagonia?

I was worried/wondering about that too - it doesn't appear in this plot, but should confirm.

xylar:
@stephenprice, @matthewhoffman, @trhille, @chloewhicker, could any of you answer the question above?  We're wanting to know if it's safe to remove ice runoff further north or if we will ever start to remove ice from Patagonia in any E3SM simulations.  In other words, do we currently (or will we soon) have configurations with mountain glaciers that produce ice runoff?

stephenprice:
Refresh my memory -- Does the iceRunoffFlux field represents calving?

darincomeau:
>Does the iceRunoffFlux field represent calving?

No, snowcapping solid ice runoff. @cbegeman is double-checking this (thanks!).

stephenprice:
Ok, thanks. I would say that, for now, you do not need to worry about snowcapping outside of Antarctica and/or Greenland. We don't have any snowcapping related treatment of mountain glaciers and/or icecaps right now. These are just treated as areas covered w/ seasonal snow at the moment (this will need revisiting at some point in the future).

@matthewhoffman should confirm if he agrees with me on this or not.

matthewhoffman:
@chloewhicker , can you confirm what happens over non-ice-sheet locations/land-units if the maximum snow thickness is reached?

xylar:
I would have thought that ice runoff would also have to include any calving from non-ice-sheet glaciers.  I don't think there is any other possible term for this.

xylar:
@darincomeau, I think this will be non-BFB for any DIB test cases, right?

cbegeman:
So based on what @chloewhicker said, it seems that we _could_ have snowcapping fluxes from Patagonia, even though we currently don't in configurations so far. I will calculate the maximum latitude in the southern hemisphere that has ice runoff fluxes in the current spreading file so we can choose that latitude as the cutoff for ice runoff removal rather than the more generous 45 degS.

chloewhicker:
if the snow depth is > than the max snow depth and the new routing method is not being used (freeze precip before sending it to river runoff) then any additional precip will be directly routed to runoff in the state it fell in. Feel free to point me to the ELM and MOSART history fields and I can look at them

darincomeau:
>@darincomeau, I think this will be non-BFB for any DIB test cases, right?
Or is it BFB because no existing tests have such an aggressive spreading function as you showed above?

This would only be non-BFB if `config_remove_ais_ice_runoff = .true.` and there is solid runoff in test cases between 60S and 45S. The only possibilities are cryo configurations where i) spreading function extends AIS solid runoff north of 60S, or ii) there is non-AIS based solid runoff in that band.

For i), I don't think there are any tested cryo configurations (B- or G-cases) that use an extended spreading function.
For ii), G-cases do not have solid ice runoff apart from Antarctica/Greenland, and I'm gathering from the (very useful) discussion here that it's possible, but not currently happening, for non-ice sheet solid ice runoff to exist in this band.

It sounds like it would be safer to adjust the bound here closer to the problem area, rather than the judicious 45S I suggested.

xylar:
Okay, great.  So we have a reasonable expectation of being BFB.

matthewhoffman:
Thanks, @chloewhicker .  Based on this statement:
> then any additional precip will be directly routed to runoff in the state it fell in.
it seems like the configuration in this PR is susceptible to inadvertently zeroing solid runoff in Patagonia if it should ever occur. 

@cbegeman and @darincomeau , if you think you can define criteria to avoid non-AIS runoff getting zeroed, that seems easiest.  I do wonder with this more extensive spreading function if even if you carve out Patagonia from the mask, it's still possible to inadvertently zero solid runoff from islands in the Southern Ocean - I see at least 3 showing up in Chloe's ELM figures.

cbegeman:
A bit of a complication: we do get ice runoff all over Patagonia in our latest v3 SORRM alfred1 run. It's just super tiny (like 1e-21 kg m^-2 s^-1 rather than 1e-7). So I think we're probably looking at non-BFB.

The maximum southern hemisphere latitude that is associated with the band shown above is -57.1854.

darincomeau:
-57S is just south of southern tip of South America, so we should be good with that?

cbegeman:
Yes, I don't see any issue with using -57. I haven't checked the latitude of those islands @matthewhoffman pointed out, but I think that creating a more complex mask is probably more effort than it's worth.

darincomeau:
Oops I missed @matthewhoffman 's comment when I wrote that.

darincomeau:
Ice runoff north of 60S now removed:
image

xylar (approved):
This looks reasonable to me.

rljacob:
@cbegeman and @proteanplanet please review.

cbegeman (approved):
Approving based on analysis of the extent of the ice runoff field. Thanks, @darincomeau!

proteanplanet (approved):
Code changes are correct.  Approved based on result presented.

jonbob:
passes:
* e3sm_cryo_developer on chrysalis

merged to next

xylar:
😌 Glad this seems to be BFB. That should save us some effort.

jonbob:
merged to master

PR #6798, opened 2024-12-04
Add relaxed slab ocean mode

whannah1:
The relaxed slab ocean (RSO) is implemented as a simpler alternative to the traditional slab ocean
mode that requires externally generated heat flux data that represents the effects of ocean transport
on ocean surface temperature. In RSO mode we want to relax to a target SST field using only a
specified mixed layer depth to allow things like cold wakes following a tropical cyclone.

Also add documentation for this mode of the data ocean model.

[BFB]


docs/refs/eam.bib
```diff
@@ -1101,3 +1101,64 @@ @article{holtslag_preface_2006
...
...ative–convective equilibrium (RCE), is proposed. RCE is an idealization of the climate system in which there is a balance between radiative cooling of the atmosphere and heating by convection. The scientific objectives of RCEMIP are three-fold. First, clouds and climate sensitivity will be investigated in the RCE setting. This includes determining how cloud fraction changes with warming and the role of self-aggregation of convection in climate sensitivity. Second, RCEMIP will quantify the dependence of the degree of convective aggregation and tropical circulation regimes on temperature. Finally, by providing a common baseline, RCEMIP will allow the robustness of the RCE state across the spectrum of models to be assessed, which is essential for interpreting the results found regarding clouds, climate sensitivity, and aggregation, and more generally, determining which features of tropical climate a RCE framework is useful for. A novel aspect and major advantage of RCEMIP is the accessibility of the RCE framework to a variety of models, including cloud-resolving models, general circulation models, global cloud-resolving models, single-column models, and large-eddy simulation models.},
```
rljacob, line 1123: can you delete the abstract from each of these?  Keeps this file from getting to long.

components/data_comps/docs/user-guide/data-ocean-idealized.md
```diff
@@ -0,0 +1,60 @@
+# Data Ocean - Idealized
```
rljacob, line 1: I suggest putting the data ocean docs in one file instead of splitting them up.  the TOC mkdocs generates on the rhs will make it easier to navigate the material.

docs/refs/eam.bib
whannah1: done.

components/data_comps/docs/user-guide/data-ocean-idealized.md
whannah1: I thought it would be too much material for a single page. I was thinking that it would be good to add some plots of my RSO tests to show what it does to the model in a longer run. And for the others I was thinking we might want to eventually add some instructions to generate new SST datasets. Maybe I'm suffering from premature optimization.

whannah1:
I wanted to get a sense of the long-term impact of using this new RSO mode - both to check that the SST would stay close to the target values and what sort of systematic response could be seen in other fields and over land.

The plot below shows 5-yr F2010 simulations and differences between the RSO case and control. They both use the same data ocean data and the RSO is configured to use a globally uniform mixed layer depth.

There is a systematic "el-Nino-like" difference in the SST... But obviously this is nothing like a real ENSO mode, and we can see warmer SST along most of the equatorial region, so I think we're just seeing the effect of warming where the insolation is the strongest. This sort of thing should disappear in a "traditional" SOM simulation because there would be strong cooling by Q-fluxes along the equator to represent the equatorial upwelling - which is not represented in the RSO mode. Similarly, there's a notable cooling along the western boundary currents and warming in regions of strong coastal upwelling that would be counteracted by Q-fluxes in a traditional SOM.

The changes over land are reminiscent of maps of surface temperature regressed on an ENSO index, but I couldn't find a good example from a quick literature search. If my intuition/memory is right on this, then perhaps the land temperature changes are not that surprising given the pattern of SST changes.

I think it's important for us to be aware of these systematic changes when using the RSO mode. For short TC hindcasts it likely won't be a problem, but I would not want to use this for longer simulations due to the biases that it will introduce.

image

components/data_comps/docs/user-guide/data-ocean-idealized.md
whannah1: I reorganized the data ocean docs to put it all on a single page.

rljacob:
@whannah1 any suggestions for who else could review this?

whannah1:
Perhaps @singhbalwinder could take a look?

singhbalwinder (changes requested):
Looks great! A very well written and clean code with documentation! I left a very minor comment.

components/data_comps/docn/src/docn_comp_mod.F90
```diff
@@ -768,6 +788,72 @@ subroutine docn_comp_run(EClock, x2o, o2x, &
...
+            RSO_slab_option = 0           ! Option for setting RSO_X_cool
+            RSO_R_cool      = 11.75/86400 ! base cooling rate [K/s]
+            RSO_Tdeep       = 271.00      ! deep water temperature [K]
+            RSO_dT_o        = 27.0        ! scaling temperature gradient
+            RSO_h_o         = 30.0        ! scaling mixed layer depth
+            !*******************************************************************
+            ! Calculate scaling function - see Eq 3 in Zarzycki (2016)
+            if (RSO_slab_option==0) RSO_X_cool = 1/(1+EXP(-0.5*(u10-30)) )               ! SLAB1
+            if (RSO_slab_option==1) RSO_X_cool =(1/(1+EXP(-0.2*(u10-30)) ))*(u10*2.4/80) ! SLAB2
+            if (RSO_slab_option==2) RSO_X_cool = 0.0                                     ! THERMO
```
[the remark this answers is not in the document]
whannah1: I recently had a conversation with some others about this, and I think I have a better handle on where this is needed and where it can be omitted. I'll modify where appropriate.

components/data_comps/docn/src/docn_comp_mod.F90
```diff
@@ -768,6 +786,72 @@ subroutine docn_comp_run(EClock, x2o, o2x, &
...
+            ! Get "background" temperature for relaxation
+            RSO_bckgrd_sst = avstrm%rAttr(kRSO_bckgrd_sst,n) + TkFrz
+            u10 = SQRT(x2o%rAttr(k10uu,n))
+            !*******************************************************************
+            ! RSO parameter values
+            RSO_slab_option = 0                  ! Option for setting RSO_X_cool
+            RSO_R_cool      = 11.75_r8/86400._r8 ! base cooling rate [K/s]
+            RSO_Tdeep       = 271.00             ! deep water temperature [K]
+            RSO_dT_o        = 27.0               ! scaling temperature gradient
+            RSO_h_o         = 30.0               ! scaling mixed layer depth
```
singhbalwinder, line 824: Thanks for fixing some of these. If you do not want to add `_r8`, the numeric literals should be assigned to variables as integers to maintain precision. Assigning variables with numbers with decimals (e.g., 30.0) will reduce their precision to single precision.
```suggestion
            RSO_Tdeep       = 271                ! deep water temperature [K]
            RSO_dT_o        = 27                 ! scaling temperature gradient
            RSO_h_o         = 30                 ! scaling mixed layer depth
```

Being explicit and adding `_r8` (e.g., `30.0_r8`) is still a better approach, as it will make it clear to the reader that these are real numbers. Declaring them as parameters is also preferable if these variables will stay constant during time-stepping.

whannah1: ok, I like suggestions.

whannah1: Done.

singhbalwinder (approved):

rljacob:
Noting that this creates NML diffs in all data ocean cases.

PR #6799, opened 2024-12-04
Fix conditional jump on uninitialized mem in SHOC

jgfouca:
The rino temp view only had the nlev-1 entry initialized. pblintd_height was accessing other entries.

I looked the the F90 impl and it is also doing this strange initialization of only nlev-1.

This PR will fix valgrind errors for shoc_run_and_cmp that have appeared on our dashboard. This error seems to go back a long time, so I don't understand why it only showed up now.

### DIFF
--- a/components/eamxx/src/physics/shoc/impl/shoc_pblintd_impl.hpp
+++ b/components/eamxx/src/physics/shoc/impl/shoc_pblintd_impl.hpp
@@ -77,7 +77,12 @@ void Functions<S,D>::pblintd(
 
   // Initialize
   bool check = true;
-  Kokkos::deep_copy(rino, 0);
+  // The loop below fixes valgrind uninitialized mem errs
+#ifndef NDEBUG
+  for (size_t i=0; i<rino.size(); ++i) {
+    rino(i)=0;
+  }
+#endif
   pblh = s_z(nlev-1);
 
   // PBL height calculation
--- a/components/eamxx/src/physics/shoc/impl/shoc_pblintd_impl.hpp
+++ b/components/eamxx/src/physics/shoc/impl/shoc_pblintd_impl.hpp
@@ -78,7 +78,7 @@ void Functions<S,D>::pblintd(
 
   // Initialize
   bool check = true;
-  s_rino(nlev-1) = 0;
+  Kokkos::deep_copy(rino, 0);
   pblh = s_z(nlev-1);
 
   // PBL height calculation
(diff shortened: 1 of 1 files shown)


components/eamxx/src/physics/shoc/impl/shoc_pblintd_impl.hpp
```diff
@@ -78,7 +78,7 @@ void Functions<S,D>::pblintd(
 
   // Initialize
   bool check = true;
-  s_rino(nlev-1) = 0;
```
tcclevenger, line 81: You can remove the `s_rino` variable above.

tcclevenger:
It is odd that the error is just now showing up. In the `pblintd_height()` call after the line in this PR `s_rino(k)` is initialized for `k=[nlev-npbl, nlev-1)`, but I doubt the value of `npbl` would have changed for shoc_run_and_cmp any time recently.

bartgol:
I don't think it just showed up. I think [it was already there](https://my.cdash.org/tests/236383752)... In fact, this is one of the errors I was trying to (well, _was going to try to_) fix today while looking at the valg fails. So thanks for taking this off the list!

components/eamxx/src/physics/shoc/impl/shoc_pblintd_impl.hpp
jgfouca: Good catch, thx

components/eamxx/src/physics/shoc/impl/shoc_pblintd_impl.hpp
```diff
@@ -78,7 +77,7 @@ void Functions<S,D>::pblintd(
...
+  Kokkos::deep_copy(rino, 0);
```
bartgol, line 80: I thought deep_copy was not allowed on device... You may instead do something like
```c++
#ifndef NDEBUG
for (size_t i=0; i<rino.size(); ++i)
  rino(i)=0;
#endif
```

jgfouca: Ah, yeah I tested this on mappy.

bartgol (approved):

components/eamxx/src/physics/shoc/impl/shoc_pblintd_impl.hpp
tcclevenger: The `s_rino` var came back in the last commit.

jgfouca: It needed to. Now that the full init is only in debug, the original init had to be restored.

tcclevenger: oh of course.

tcclevenger (approved):

PR #6801, opened 2024-12-04
update PAM submodule to fix build error

whannah1:
the PAM update removes include lines for the SCREAM *_f90.hpp files that have been removed.
[BFB]


rljacob:
Noting this was broken by #6779

PR #6802, opened 2024-12-05
Adds internal ocean frazil ice porosity

njeffery:
This scheme updates the internal frazil ice salinity redistribution. The default ocean frazil ice parametrization contributes to unphysical stratification of the upper ocean during formation because it uses the ocean-sea ice coupling salinity for internal ocean redistribution.

Changes and simulations are documented here: https://acme-climate.atlassian.net/wiki/spaces/HESF/pages/4820598785/Ocean+Frazil+Ice+Porosity

The code contains comments !NJ to help explain the changes.  I can clean that up once the code has been review.

[NML]
[Stealth]


xylar:
Relevant discussion can be found at:
https://github.com/E3SM-Ocean-Discussion/E3SM/pull/115

xylar:
@njeffery, you mentioned that you had a design document on your to-do list.  Could it maybe make sense to hold off on this PR until that is done?

https://github.com/E3SM-Ocean-Discussion/E3SM/pull/115#discussion_r1856911011

xylar:
Aach, sorry!  I missed that that was linked above!

xylar:
@njeffery, how easy or hard would it be to run a model vs. model run of MPAS-Analysis so we could see the difference plots between your run (without BGC) and the control run?  The differences are definitely there but they're subtle enough that those difference plots might really help.

I'd be happy to help with running the analysis if need be.

njeffery:
I have the first 100 years of the control run ocean history files only on Perlmutter: /global/cfs/projectdirs/e3sm/njeffery/E3SMv3/20231209.v3.LR.piControl-spinup.chrysalis/archive/ocn/hist   .  I could bring the ice history directory too.  What else would you need?

xylar:
@njeffery, I think I would want everything on Chrysalis because that's where the control run is.  Since you ran on Anvil, it's already there.  I can give it a try tomorrow.  The config files from your analysis run and Wuyin's should tell me where everything is (or was at least) and I can let you know if anything is missing.

xylar:
@njeffery, okay, I see the issue now.  This run is no longer on Chrysalis.  It is pretty difficult to compare runs on different machines so I think we would either need those 100 years of ocean and sea ice history, plus a restart file for ocean and sea ice, plus the namelist and streams files.  Or we would need the MPAS-Analysis run that was already done (maybe easier).  I will see if I can restore that with globus directly from tape at NERSC.

xylar:
@cbegeman, @darincomeau and @proteanplanet, I know you are both going to be busy with AGU but I wanted to check with you about this.  I would not feel comfortable approving this change, which I think could be significant, without a B-case simulation with thermodynamics active in the ice-shelf cavities (PISMF - prognostic ice-shelf melt rates).  I think this might mean doing a simulation branched of from @darincomeau's `alfred1` simulation.

xylar:
@njeffery, I now see that you've moved the first 100 years of data from your simulation to Perlmutter.  I can do the analysis there then (on Monday).

Could you give me permission to read the files and directories in the archive on your scratch drive?
```
chmod -R go+rX /global/cfs/projectdirs/e3sm/njeffery/E3SMv3/20241010.Frazil2.underIce.piControl.anvil
```
I'll do the same for my data once it's extracted.  It's at:
```
/pscratch/sd/x/xylar/20231209.v3.LR.piControl-spinup.chrysalis
```

njeffery:
Thanks, @xylar.  You should now have permission.

proteanplanet:
@xylar and @njeffery: This is what has to happen to this PR from here in:  It requires a B-case LR 100-year simulation to be compared with a control.  Moreover, the change must be introduced initially to master as a BFB change.  So the PR gets tested with the change to porosity to the value currently set in the code for this PR, but the code should be set with porosity at zero which I assume makes it BFB with master.  Then, for the time being, we test, for example, SORRM and BGC configurations with the porosity set to non-zero in the namelist changes in run scripts, also the HR group, who are interested in testing this as well.  Once we are all happy it's performing correctly, we then submit another PR making this default after broad consultation in group meetings.

proteanplanet:
Documentation at the link provided by @njeffery meets requirements.

xylar:
@proteanplanet

> Moreover, the change must be introduced initially to master as a BFB change. So the PR gets tested with the change to porosity to the value currently set in the code for this PR, but the code should be set with porosity at zero which I assume makes it BFB with master.

I thought this initially as well but reading the code I think a porosity of zero would only be BFB under ice shelves, where the current salinity of ice is zero. In the open ocean, this is non-BFB in all conditions.

vanroekel:
I agree with @xylar in the open ocean the frazil reference salinity is used right now, so porosity of zero wouldn't be BFB in the open ocean.  I also don't believe we can use the code as and just tuning the porosity to get BFB.

njeffery:
Do you want me to add a config flag that adds an option which reverts to the original code?

proteanplanet:
Sorry I was unclear, I'm requesting either a switch to make this BFB, which could be, for example, setting porosity to zero, or a separate flag.  We are doing something similar for changes to top of atmosphere radiation for CMIP7, which will be highly likely to become the default, in a separate PR later.

njeffery:
Sure.  i can make it a stealth feature.

cbegeman:
One option, which may be along the lines that @proteanplanet was thinking, would be to have a config option that is the minimum frazil salinity, defaulting to 4, and then we could reproduce the current behavior of the model by setting frazil ice porosity to 0 (minimum would kick in everywhere).

components/mpas-ocean/src/Registry.xml
```diff
@@ -984,11 +984,15 @@
...
-		/>
+		/> 
 		<nml_option name="config_frazil_ice_density" type="real" default_value="1000.0" units="kg m^-3"
 					description="Assumed density of frazil. NOTE: test and make consistent with E3SM."
 					possible_values="Any positive real number."
 		/>
+		<nml_option name="config_frazil_ice_porosity" type="real" default_value="0.85" units="1"
+					description="Assumed porosity of frazil"
+					possible_values="Any positive real number from 0 to 1."
+		/>
```
cbegeman, line 995: I'm wondering if we should call this something other than porosity. When I hear porosity, I assume that the volume of frazil ice formed is only `porosity * frazil ice thickness tendency` (`* areaCell`) because part of the frazil ice tendency (the volume advected upward) is pure seawater. But my understanding of the implementation is really that this is expressing the salinity of the frazil as a fraction of the surrounding seawater (the ice formed and latent heat extracted is the full frazil ice thickness tendency). So I wonder if a less confusing name would be something like `config_frazil_ice_salinity_fraction`.

xylar: I also got a bit confused by the use of "porosity".  I think the use of that term was partly responsible for the confusion we had in our webex discussion a little while back.  I seemingly couldn't let go of the idea that not just salt but seawater was being captured and transported by frazil in this approach.

njeffery:
I've modified the code so that it's now BFB with default settings (verified in two short 10 day simulations).

xylar:
Regarding the MPAS-Analysis runs, I've been sick so not putting a huge amount of effort into the plotting but I'm also seeing long queue wait times and longer-than-expected run times for MPAS-Analysis on Perlmutter.  Plus, I made a mistake with remapping that forced me to start over.  I believe I should have the main and control runs plotted once my next pair of jobs runs.  The main vs. ctrl plotting shouldn't take too long once that happens, since it's just a question of plotting without further computations.  The Perlmutter experience does make me miss running analysis on Chrysalis.

xylar:
## Model vs. Model with latest MPAS-Analysis including "polar regions" analysis

https://portal.nersc.gov/cfs/e3sm/xylar/nicole-frazil-porosity/mpas_analysis/20241010.Frazil2.underIce.piControl.anvil_vs_piControl/ts_0001-0100_climo_0051-0100/

## This branch

https://portal.nersc.gov/cfs/e3sm/xylar/nicole-frazil-porosity/mpas_analysis/20241010.Frazil2.underIce.piControl.anvil/ts_0001-0100_climo_0051-0100/

## Control run

https://portal.nersc.gov/cfs/e3sm/xylar/nicole-frazil-porosity/mpas_analysis/20231209.v3.LR.piControl-spinup.chrysalis/ts_0001-0100_climo_0051-0100/

@njeffery, could you add these links somewhere appropriate on the confluence page for the simulation?

xylar:
I'm mostly seeing ensemble-level variability but serious improvements in AMOC:

image
image

and quite a bit more ocean heat uptake in the North Atlantic:

image

Presumably related to more overall OHC:
image

milenaveneziani:
My brain picked up 'serious improvements in AMOC' and I immediately went to look at the diags :). A few other improvements I noticed are in the subpolar North Atlantic: more saline surface, which might lead to the more dense water mass formation in the Greenland Sea (from the TS diagrams), and also improved sea ice in the Lab Sea and Nordic Seas (as well as central Arctic ice thickness). This looks very promising!

components/mpas-ocean/src/Registry.xml
```diff
@@ -984,11 +984,15 @@
...
 					possible_values="Any positive real number."
-		/>
+		/> 
 		<nml_option name="config_frazil_ice_density" type="real" default_value="1000.0" units="kg m^-3"
 					description="Assumed density of frazil. NOTE: test and make consistent with E3SM."
 					possible_values="Any positive real number."
 		/>
+		<nml_option name="config_frazil_ice_porosity" type="real" default_value="-1.0" units="1"
+					description="Internal porosity of frazil"
+					possible_values="Positive real number from 0 to 1. Negative value reverts to using the reference salinity for internal frazil."
```
cbegeman, line 994:
```suggestion
					possible_values="Positive real number from 0 to 1. Indicates the fraction of local ocean salinity to use for frazil ice. If non-negative, applies to both land ice frazil and sea ice frazil. Negative value reverts to using the reference salinity for internal frazil and 0 for land ice.
```

Also modify the description of `config_frazil_sea_ice_reference_salinity` so that it does not say that 0 salinity is always used for land ice.

cbegeman:
Has there been any testing with land ice thermodynamics on (even in G-cases)? It would seem strange to me to merge untested changes that apply to land ice, even if this is a stealth feature right now.

components/mpas-ocean/src/Registry.xml
xylar: @njeffery, do you have feedback on this?

xylar:
@cbegeman, I don't think it has been tested with active land-ice thermodynamics, no.  Do you have a baseline you'd suggest for that?  The BlueTip configurations might be too expensive to be attractive and they're all B-cases as far as I'm aware.

components/mpas-ocean/src/Registry.xml
xylar: @njeffery, I feel uncomfortable about this approach of using a negative value as a flag.  This is not an approach we have taken elsewhere in MPAS-Ocean.  Instead, we have, at least as far as I'm aware, always used a separate logical to turn the feature on or off.

components/mpas-ocean/src/Registry.xml
njeffery: @xylar  and @cbegeman : I believe the confusion comes from thinking of frazil as fresh ice + water when it's really sea ice which is a mushy layer.  The use of porosity in this context is quite common.

components/mpas-ocean/src/shared/mpas_ocn_frazil_forcing.F
```diff
@@ -600,31 +610,51 @@ subroutine ocn_frazil_forcing_build_arrays(domain, meshPool, forcingPool, stateP
...
 
-        ! frazilIceFreshwaterFlux forms part of the landIceFreshwaterFluxTotal computed in surface land ice fluxes
-        frazilIceFreshwaterFlux(iCell) = - (sumNewFrazilIceThickness * config_frazil_ice_density) / dt
         ! accumulate frazil mass to column total
         ! note: the accumulatedFrazilIceMass (at both time levels) is reset to zero after being sent to the coupler
         accumulatedFrazilIceMassNew(iCell) = accumulatedFrazilIceMassOld(iCell) + sumNewFrazilIceThickness &
-                                           * config_frazil_ice_density
-        accumulatedFrazilIceSalinityNew(iCell) = accumulatedFrazilIceSalinityOld(iCell) + sumNewThicknessWeightedSaltContent
+           * config_frazil_ice_density
+
```
cbegeman, line 626: ! accumulatedFrazilIceSalinityNew is overwritten below if usePorosity = 1

components/mpas-ocean/src/Registry.xml
bpbond: +1 to this point by @xylar

components/mpas-ocean/src/shared/mpas_ocn_frazil_forcing.F
cbegeman, line 626:
```suggestion
! accumulatedFrazilIceSalinityNew is overwritten below if usePorosity = 1
```

cbegeman:
@njeffery Can you comment on the outcome of salt conservation as reflected in `lowSalinityFrazilIce` in your tests?

components/mpas-ocean/src/Registry.xml
njeffery: @xylar : that's a good point. I figured this would be a clean option because this is not really an independent feature.  It just makes explicit the choice to use either a fixed reference salinity or a fraction of the local salinity  in the internal frazil calculation.

xylar: @njeffery, I take your point that it's a little different than a typical on/off feature.  In such cases, we typically have a string namelist option that selects between the possible features and then we have one or more logicals in the code that determine the behavior (to avoid constantly performing string comparisons except at init within the code itself).  Would that be an approach you could get behind here as well?

njeffery: Sure.  I'll try to get to it this week.

njeffery:
The lowSalinityFrazilIce diagnostic is nonzero when salinity conservation from frazil formation is violated.  In all cases, this occurs when the cold ocean region is too fresh and doesn't have 4psu times the frazil mass to subtract from the ocean column.  However when the sea ice in that region melts, the ocean assumes the melt water has 4 psu.  The  result is a small, additional positive salinity flux into the ocean.

lowSalinityFrazilIce_timeseries_years1-100_20241010 Frazil2 underIce piControl anvil


lowSalinityFrazilIceNH_20241010 Frazil2 underIce piControl anvil_ANN_years0051-0100

cbegeman:
> The lowSalinityFrazilIce diagnostic is nonzero when salinity conservation from frazil formation is violated. In all cases, this occurs when the cold ocean region is too fresh and doesn't have 4psu times the frazil mass to subtract from the ocean column. However when the sea ice in that region melts, the ocean assumes the melt water has 4 psu. The result is a small, additional positive salinity flux into the ocean.

Thanks! It's good to hear that the regions affected seems small. Am I correct in thinking that the salt non-conservation issue is also present in master, it just wasn't being explicitly tracked before?

njeffery:
@cbegeman :  That's right.  This PR doesn't directly address the brackish water issue.

proteanplanet:
The change in AMOC is definitely statistically significant and the standard deviation (noise) of the signal is reduced:

MOC_Comparison_Frazil

As a rough ballpark, it buys between 2 and 3 Sv based on results using 2-10 year low filtering.

proteanplanet:
@njeffery, @cbegeman et al.:  Looking through this PR, can you summarize here what is left to be completed for everyone to sign off on this PR?

cbegeman:
> @cbegeman, I don't think it has been tested with active land-ice thermodynamics, no. Do you have a baseline you'd suggest for that? The BlueTip configurations might be too expensive to be attractive and they're all B-cases as far as I'm aware.

I think a G-case would be fine for a land ice thermodynamics test. @darincomeau What do you think about branching from year 21 of this case and running to year 30? Do we still have the restarts and the ability to do a MPAS-Analysis comparison? I checked that the land ice frazil accretion rates are non-negligible for this case https://acme-climate.atlassian.net/wiki/spaces/PSC/pages/4575821825/20240829.GMPAS-JRA1p5-DIB-PISMF.TL319_SOwISC12to30E3r3.mesh-review-test.chrysalis

darincomeau:
@cbegeman we can do that branch run you suggest. That run has been long-term archived and the output wiped from disk at LCRC, but I can pull back what we need to do the branch run. The mpas-analysis output directory is still there, so hopefully we won't need to pull back all the output.

cbegeman:
@darincomeau Thanks! Do let me know if it's looking to be too much trouble.

rljacob:
notes:  still needs approval

darincomeau:
Here's a 10 year G-case with `config_frazil_ice_porosity = 1.0` in `mpaso_in`, branched from [year 21 of an earlier G-case.](https://acme-climate.atlassian.net/wiki/spaces/PSC/pages/4575821825/20240829.GMPAS-JRA1p5-DIB-PISMF.TL319_SOwISC12to30E3r3.mesh-review-test.chrysalis)

This run: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/20250108.GMPAS-JRA1p5-DIB-PISMF.TL319_SOwISC12to30E3r3.add-internal-frazil-porosity.chrysalis/yrs1-10/

Control run: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/20240829.GMPAS-JRA1p5-DIB-PISMF.TL319_SOwISC12to30E3r3.mesh-review-test.chrysalis/yrs21-30/

Comparison (no time series): https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/20250108.GMPAS-JRA1p5-DIB-PISMF.TL319_SOwISC12to30E3r3.add-internal-frazil-porosity.chrysalis/vs_yrs1-10_no-ts/

I don't notice anything of concern in the Southern Ocean; there is a significant reduction of Arctic sea ice thickness bias in the control: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/20250108.GMPAS-JRA1p5-DIB-PISMF.TL319_SOwISC12to30E3r3.add-internal-frazil-porosity.chrysalis/vs_yrs1-10_no-ts/sea_ice/index.html#arctic_extended_nh_thick

milenaveneziani:
I'm impressed by the improvements in the sea ice thickness in the Arctic (both seasons!).
Surface Arctic salinity may get a bit worse, especially in winter, but SSS seems to get better in parts of the Greenland Sea and in the Irminger/Labrador Seas.

vanroekel:
Do the large changes in the arctic make sense?  They seem much more dramatic than the LR B-case.  It's not clear why the signal would be so much stronger in this case, unless it's just chalked up to idiosyncrasies of G-cases.

Overall though things on the ocean side are mostly small changes, which is consistent with my G-case experience as well.  The sea ice shifts are far larger than what I see in G-case ICOS30 runs with this branch.

Also this change is surprising https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/20250108.GMPAS-JRA1p5-DIB-PISMF.TL319_SOwISC12to30E3r3.add-internal-frazil-porosity.chrysalis/vs_yrs1-10_no-ts/ocean/index.html#&gid=19&pid=4

For 10 years I don't understand why we see a huge temperature signal in the tropics at 100m

One last question - @darincomeau did you see any shifts in AMOC?  I don't see the figure in your analysis

cbegeman:
@vanroekel Here's a link to MOC time series:
This branch: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/20250108.GMPAS-JRA1p5-DIB-PISMF.TL319_SOwISC12to30E3r3.add-internal-frazil-porosity.chrysalis/yrs1-10/ocean/index.html#moc&gid=49&pid=5
Control: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/20240829.GMPAS-JRA1p5-DIB-PISMF.TL319_SOwISC12to30E3r3.mesh-review-test.chrysalis/yrs21-30/ocean/index.html#moc&gid=47&pid=5\

Appears to recover ~1 Sv

components/mpas-ocean/src/Registry.xml
njeffery: Added a logical  "config_use_frazil_ice_porosity" (default false) to turn on this approach.  Will push once I've completed a short test.

njeffery:
@darincomeau :  It looks to me like the sea ice initial state for the control run is not year 21, but year 1.  This would explain the very large differences.

milenaveneziani:
@njeffery: my understanding of Darin's comment is that he branched off the Main run (with ice porosity) from year 21 of the Control run, and so that is why he is comparing years 1-10 of the main run with years 21-30 of the control. But Darin should confirm.

njeffery:
@milenaveneziani :  Sorry for my confusion. I meant that the sea ice initial state of the feature run is not year 21 of the control, but year 1 of the control.  You can see this by looking at the NH sea ice volume timeseries:

Feature run
iceVolumeNH_20250108 GMPAS-JRA1p5-DIB-PISMF TL319_SOwISC12to30E3r3 add-internal-frazil-porosity chrysalis


Control run
iceVolumeNH_20240829 GMPAS-JRA1p5-DIB-PISMF TL319_SOwISC12to30E3r3 mesh-review-test chrysalis

darincomeau:
@njeffery it certainly looks like you may be right that the sea ice IC is wrong in the feature run, but in `streams.seaice` I see `input` set correctly as `20240829.GMPAS-JRA1p5-DIB-PISMF.TL319_SOwISC12to30E3r3.mesh-review-test.chrysalis.mpassi.rst.0021-01-01_00000.nc`, however in the namelist I see `config_initial_condition_type = 'cice_default'`, so maybe I messed up and default sea ice IC was overwritten.

I think you all have gathered this, but I did this case as an "initial" run type, setting the ocn/ice ICs to the year 21 restarts from the control run, and reset the forcing start years manually to 1978, which is why the time series don't align. I was unsure if the "branch" run type would set the forcing years correctly, but next time I'll just try that and fix if needed (I was sick last week so not thinking clearly).

cbegeman:
@njeffery I encountered a threading issue when running with NTRDS=2 in the ocean component. @jonbob found that one of the variables needed to be removed from the omp private list. I tested the change and it now passes with NTHRDS=2 for the v3 LR piControl configuration. Could you add this commit to your branch 08348e954a369e430a87d4a71b9e8c4da7eb3696? Thanks!

njeffery:
Thanks @cbegeman !  Added the threading fix.

darincomeau:
I was foiled again by the sea ice disk initial condition being used in the hybrid re-run I set up. I posted an issue about it: https://github.com/E3SM-Project/E3SM/issues/6925

In the meantime I'm re-running this again, manually adding the `restart_ic` stream to `streams.seaice` to use the intended sea ice branch point.

darincomeau:
SORRM G-case 10 year branch run *finally* done correctly, comparison analysis here:

https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.dcomeau/20250126.GMPAS-JRA1p5-DIB-PISMF.TL319_SOwISC12to30E3r3.add-internal-frazil-porosity.chrysalis/vs_yrs21-30/

Minimal changes to sea ice here; posting NH volume time series to confirm branch run used correct initial condition:
image

cbegeman:
Thanks @darincomeau! Changes in that run are reassuringly minor as far as I saw.

xylar (approved):
@njeffery, I appreciate you putting so much time into this and putting up with the messy review process.  Based on @darincomeau's latest comparison results, this looks good to go as far as I'm concerned.  It clearly has positive impacts on the Arctic and little impact on the Antarctic.

bpbond (approved):

components/mpas-ocean/src/shared/mpas_ocn_frazil_forcing.F
```diff
@@ -448,6 +452,13 @@ subroutine ocn_frazil_forcing_build_arrays(domain, meshPool, forcingPool, stateP
       frazilSalinityTendency = 0.0_RKIND
       frazilLayerThicknessTendency = 0.0_RKIND
 
+      ! initialize frazil salinity conservation field
+      lowSalinityFrazilIce = 0.0_RKIND
+
+      ! Determine if using porosity to compute internal frazilSalinity
+      usePorosity = 0 ! use config_frazil_sea_ice_reference_salinity
+      if (config_use_frazil_ice_porosity) usePorosity = 1
+
```
vanroekel, line 461: not a big deal, but couldn't this be moved to some init routine?  The flag isn't changing during the simulation.

njeffery: Good point.  I'll see what I can do.

njeffery:
Yes.  Verified BFB.

njeffery:
@darincomeau  and @vanroekel:   I'm still puzzled by the potential temperature difference at 100 m. 

temp_20250126 GMPAS-JRA1p5-DIB-PISMF TL319_SOwISC12to30E3r3 add-internal-frazil-porosity chrysalis_depth_-100_JFM_years0021-0030

There's also a corresponding salinity difference.  I'm wondering if frazil's forming at depth where it shouldn't be.

@darincomeau :  Would you continue your run for another 5 years and add some additional history fields?

timeMonthly_avg_frazilSalinityTendency
timeMonthly_avg_frazilLayerTendency
timeMonthly_avg_frazilTemperatureTendency

milenaveneziani:
> I'm still puzzled by the potential temperature difference at 100 m

sorry to jump in, but I was looking at the difference plot that @njeffery posted above: I see some weird block-like features in there, specifically in the western Pacific (North and South tropical) and in the Arabian Sea. Do you have an idea of why we would get those sort of features?

njeffery:
@proteanplanet :  I'm testing a small code change (moved some initialization code to frazil_init  as per @vanroekel 's request).    Once that's been verified I'll commit.  Then I think we should be good to go.

darincomeau:
>@darincomeau : Would you continue your run for another 5 years and add some additional history fields?
timeMonthly_avg_frazilSalinityTendency
timeMonthly_avg_frazilLayerTendency
timeMonthly_avg_frazilTemperatureTendency

Added these to `timeSeriesStatsMonthly` and resubmitted.

darincomeau:
ERROR: Requested field frazilSalinityTendency not available

I think I need to add this somewhere for it be available for timeSeriesStatsMonthly?

njeffery:
That's odd.  I've used it before in timeSeriesMonthlyOutput.  Can you get the frazilTemperatureTendency and frazilLayerThicknessTendency to work?  If so, those are enough.

components/mpas-ocean/src/shared/mpas_ocn_frazil_forcing.F
njeffery: @vanroekel :  I've  moved the code to ocn_frazil_forcing_init.

vanroekel (approved):
looks great!  Approving based on visual inspection and extensive testing by others.

xylar:
> I think I need to add this somewhere for it be available for timeSeriesStatsMonthly?

I don't believe that's necessary. All variables should be possible to include. Could it be a wrong name.

darincomeau:
>ERROR: Requested field frazilSalinityTendency not available
I think I need to add this somewhere for it be available for timeSeriesStatsMonthly?

Nevermind I had it in the wrong streams, it's back in the queue.

darincomeau:
Sorry for the delay here; additional fields

`frazilLayerThicknessTendency`
`frazilSalinityTendency`
`frazilTemperatureTendency`

are in timeSeriesStatsMonthly files for years 21 - 35 here (part of what took so long is I forgot to CONTINUE_RUN=true so it started over, just a myriad of mess-ups over here)

```
/lcrc/group/acme/ac.dcomeau/scratch/chrys/20250126.GMPAS-JRA1p5-DIB-PISMF.TL319_SOwISC12to30E3r3.add-internal-frazil-porosity.chrysalis/run
```

darincomeau (approved):
Didn't mean to sit on an approval here - I'm happy with this option, and approve based on testing.

jonbob:
@vanroekel and @proteanplanet -- I know there are some concerns about the impact of this PR on the current HR run. Are you both OK with this going into the repo as a stealth feature as-is?

njeffery:
Thanks @vanroekel.  It's also looking like we  may end up with different default parameter values for low res/high res.

jonbob:
Passes:
* e3sm_ice_developer on chrysalis
* SMS_D_Ld1.ne30pg2_r05_IcoswISC30E3r5.WCYCL1850.chrysalis_intel.allactive-wcprod with expected NML DIFF

merged to next

jonbob:
merged to master and expected NML DIFFs blessed, except for chrysalis which is having disk issues

PR #6812, opened 2024-12-06
EAMXX: add a DataInterpolation class

### LINKED ISSUE #6810: EAMXX: unify time interpolation implementations
Right now, there are three places where time interpolation of data is implemented:
 - `share/util/eamxx_time_interpolation.*pp`
 - `control/intense_observation_period.*pp`
 - `physics/spa/spa_functions*.hpp`
 
 There are probably some peculiarity in all of these but they all share some common aspect. It's likely that we could reduce the code duplication if we found a way to implement the feature in a single place.

### LINKED ISSUE #6820: EAMxx: Kokkos View allocations in spa
1 cuda malloc/free in spa

bartgol:
The class is in charge of reading in time-dependent datasets from input file(s), interpolate to current model time, and possibly do horizontal/vertical interpolation

---

The implementation is not yet complete, but some early thoughts on whether we're on the right track or not may help. Besides, getting some early testing on ghci machines can also help.

Things left to do:
- [x] testing for time-only interpolation for Linear timeline
- [x] add creation of vert/horiz remappers: this needs some extra input from parameter list (a map file, and for vert remap some choices for top/bot extrapolation)
- [x] testing for time+horiz, time+vert, and time+horiz+vert
- [ ] add possibility of using an "IOP remapper" (a closest-column kind for horiz interpolation). This may allow the upcoming IOP forcing atm proc to use the DataIntepolation routine directly.
- [ ] start using DataInterpolation in eamxx. E.g., use in SPA, and fix the alloc issue (see linked issue)

I am NOT implementing a data interpolation for variables without a time dependence. It simply reduces to constructing 2 remappers, and since non-time dep data is most likely load-able at init, there's no real need for a complex data structure. Time interpolation is the most tricky of the 3 anyways, due to the different needs (nudging uses Linear timeline, while SPA uses a YearlyPeriodic one)

For reviewers: i recommend reviewing one commit at a time. That would help isolate concepts, so you can hopefully better follow. The hardest thing to review is the data interpolation unit tests, since there is quite a bit of code needed to generate cases that we can manually check without redoing the same operation as the interpolation class (copy+paste of src code in tests folder doesn't help).

Fixes #6810 
Fixes #6820

### DIFF
--- a/components/eamxx/src/share/util/scream_time_stamp.hpp
+++ b/components/eamxx/src/share/util/scream_time_stamp.hpp
@@ -86,6 +86,86 @@ TimeStamp operator- (const TimeStamp& ts, const int dt);
 // If input string is not of the format YYYY-MM-DD-XXXXX, returns an invalid time stamp
 TimeStamp str_to_time_stamp (const std::string& s);
 
+// An enum describing two ways to look at timestamps:
+//  - Linear: treat them as part of a 1d line
+//  - YearlyPeriodic: treat them as part of a yearly periodic orbit
+// This is used in the TimeInterval class below to correctly handle time stamps differences
+enum class TimeLine {
+  YearlyPeriodic,
+  Linear
+};
+
+/*
+ * Small struct to deal with time intervals
+ *
+ * The struct simply contains timestamps for [begin,end] interval,
+ * and allows two things: compute the interval length (in days), and check if
+ * a timestamp lies within the interval.
+ *
+ * When the TimeLine arg to the ctor is YearlyPeriodic, the year part of beg/end
+ * time points is ignored. In this case, the length of the time interval is bound
+ * to be in the interval [0,365] (in non-leap years)
+ */
+struct TimeInterval {
+
+  TimeStamp beg;
+  TimeStamp end;
+  TimeLine  timeline = TimeLine::Linear;
+  double length  = -1; // the interval length
+
+  TimeInterval () = default;
+  TimeInterval (const util::TimeStamp& b, const util::TimeStamp& e, TimeLine tl, bool do_compute_length = true)
+   : beg (b), end (e), timeline (tl)
+  {
+    if (do_compute_length)
+      compute_length ();
+  }
+
+  bool contains (const util::TimeStamp& t) const {
+    if (timeline==TimeLine::Linear) {
+      // Compare the full time stamps
+      return beg<=t and t<=end;
+    } else {
+      // Compare the fraction of year for beg/end and t.
+      // Pay extra attention to the case where new year's eve
+      // is in [bec,end]
+      auto t_frac = t.frac_of_year_in_days();
+      auto end_frac = end.frac_of_year_in_days();
+      auto beg_frac = beg.frac_of_year_in_days();
+      bool across_nye = beg.get_month()>end.get_month();
+      if (not across_nye) {
+        return beg_frac<=t_frac and t_frac<=end_frac;
+      } else {
+        // We are either PAST beg or BEFORE end (but not both)
+        return t_frac>=beg_frac or t_frac<=end_frac;
+      }
+    }
+  }
+
+  void compute_length () {
+    if (timeline==TimeLine::Linear) {
+      length = end.days_from(beg);
+    } else {
+      bool across_nye = beg.get_month()>end.get_month();
+      auto frac_beg = beg.frac_of_year_in_days();
+      auto frac_end = end.frac_of_year_in_days();
+      if (across_nye) {
+        double year = end.days_in_curr_year();
+        length = frac_end + (year - frac_beg);
+      } else {
+        length = frac_end - frac_beg;
+      }
+    }
+  }
+
+  // Advance the interval, so that it now starts from the old end
+  void advance(const TimeStamp& new_end) {
+    beg = end;
+    end = new_end;
+    compute_length();
+  }
+};
+
 } // namespace util
 
 } // namespace scream
--- a/components/eamxx/src/share/util/scream_time_stamp.hpp
+++ b/components/eamxx/src/share/util/scream_time_stamp.hpp
@@ -45,6 +45,11 @@ public:
   std::string get_time_string () const;
   double frac_of_year_in_days () const;
 
+  int days_in_curr_month () const;
+  int days_in_curr_year () const;
+
+  TimeStamp curr_month_beg () const;
+
   // === Update method(s) === //
 
   // Set the counter for the number of steps.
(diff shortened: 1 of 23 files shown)


components/eamxx/src/share/util/scream_time_stamp.hpp
```diff
@@ -86,6 +86,86 @@ TimeStamp operator- (const TimeStamp& ts, const int dt);
...
+};
+
+/*
+ * Small struct to deal with time intervals
+ *
+ * The struct simply contains timestamps for [begin,end] interval,
+ * and allows two things: compute the interval length (in days), and check if
+ * a timestamp lies within the interval.
+ *
+ * When te TimeLine arg to the ctor is YearlyPeriodic, the year part of beg/end
```
AaronDonahue, line 105: "te" should be "the"

AaronDonahue (commented):
Would it make sense to break this plan into three PRs.  This one which introduces the DataInterpolation Class and applies it to time interpolation and then subsequent ones with horiz and vertical?

bartgol:
> Would it make sense to break this plan into three PRs. This one which introduces the DataInterpolation Class and applies it to time interpolation and then subsequent ones with horiz and vertical?

I don't think so. The three interpolations are tightly coupled via remappers. Sure, the remapper can be `IdentityRemapper` if no remap is requested, but the structure is there. I think it may be confusing to integrate a feature that seems to support something, but it doesn't.

For reviewers: I try to isolate concepts into their own commits. So one way to make reviews easier is to review one commit at a time.

components/eamxx/src/share/util/scream_time_stamp.hpp
```diff
@@ -86,6 +86,86 @@ TimeStamp operator- (const TimeStamp& ts, const int dt);
...
+        return beg_frac<=t_frac and t_frac<=end_frac;
+      } else {
+        // We are either PAST beg or BEFORE end (but not both)
+        return t_frac>=beg_frac or t_frac<=end_frac;
+      }
+    }
+  }
+
+  void compute_length () {
+    bool across_nye = beg.get_month()>end.get_month();
```
tcclevenger, line 146: We only need `across_nye` in the else-block, I think it's easier to read if it's defined there.

tcclevenger (commented):
Working through the PR. Here are some comments from "EAMxx: add a TimeInterval lightweight struct"

bartgol:
@mahf708 Yeah, I decided to handle different inputs in the parameterlist, so that customer atm procs can simply fwd the sublist in their param list, rather than having to read it, just to pass it in. That said, it is true that the param list hides a bit the requirements. That is, it's not clear from the header what the expectations are on the input PL. Perhaps a constructor listing explicitly what should be inside is more meaningful, and may be more self-documenting. I will think about an alternative impl, and if I see it doesn't get too complex, I'll switch.

bartgol:
@AaronDonahue @mahf708 @tcclevenger This PR is finally ready for review, just in time for xmas!

A few comments of what you'll find in it:

- time stamp: added some QOL methods, and added a TimeInterval lightweight class. The latter has some logic to handle differently linear history and yearly periodic history.
- IO mods: I wanted to be able to swap fields pointers in an input class without rebuilding it all. So I added a small capability there. Also, use some of the QOL methods above in IO control
- vertical remap: allow more fine-grained handling of the vertical pressure coordinate. In particular:
	- the tgt fields CAN have interfaces: before, this was not the case cause we only used it in the model->file direction, but DataInterpolation uses it in the other direction
	- allow src fields to NEVER be at interfaces: again, being the new use case the mirror image of the output class usage, this is needed
	- allow to set and query pressure profiles
- DataInterpolation: this is the big class, but it came out relatively skinny. Highlights:
	- hremap and vremap are optional (we create identity remappers if not needed), but time interpolation is MANDATORY. Customers that have constant data can do the work themselves at init time
	- vertical remap, if not identity, can accept as a src pressure (i.e., in the input files) either a constant 1d profile or a time-dependent 3d profile. If other use cases will be needed, we'll adjust.
	- hremap follows the same restrictions as usual: it can only be identity or a refining remap (i.e. data on a COARSER grid). At some point, I want to get rid of this limitation, but it would require some serious work on the horiz remapper classes.
	- we store beg/end copies of fields, corresponding to the time interval [beg,end] in the data that contains the model time
	- setup in 3 phases: create time database, create remappers, init the end state to the _last_ t_beg such that t_beg<t0. This will trigger an update of the end state during the 1st timestep. I opted for this (rather than load t_beg/t_end at init) since it makes the runtime code more uniform.
	- runtime is almost trivial, except for the data interval update. There, we swap the content of beg/end structures, and reload a new slice in end
- data interp testing: this is not so trivial. It wasn't easy to come up with a framework that made the interpolation doable "by hand" without copy+pasting the code in the class itself. It all boils down to how the data is generated, which is roughly like this:
```
  dh = 1.0/(ngcols-1);
  dv = 1.0 / nlevs;
  h_value = col_gid * dh* + 1.0; // +1 ensures we don't have a col full of zeros at gid=0.
  v_value = lev * dv*;
  data(icol,ilev) = h_value*v_value; // for interfaces
  data(icol,ilev) = h_value*(v_value + dv/2); // for midpoints
```
- We generate data for scalar/vector fields, that are either 2d or 3d. If 3d, they can be either midpoints or interface fields. So the list is `s2d,v2d,s3d_m,v3d_m,s3d_i,v3d_i` with obvious meaning of names.
- The generated pressure profiles are taken as `p3d = s3d_m`, and `p1d` is a vertical slice of `p3d` (ensuring all ranks agree on it).
- For vert remap testing, in case of a 3d dynamic pressure profile, I make the model pressure have the SAME profile at all colums. This allows me to do away with vertical interpolation when computing the expected answer
- misc warnings/cmake fixes

I recommend reviewsing one commit at a time, since it may help isolating concepts. Otherwise, review everything but the data interp class/tests first, and then focus on the data interp stuff.

Happy holidays!

bartgol:
Expected fails on cuda. The v1 fail was during cleanup, when something went amiss for some reason. All is good.

@tcclevenger @AaronDonahue @mahf708 would you mind taking a look?

tcclevenger (approved):
LGTM. I'll think about what IOP would need to use this

bartgol:
> * I think it would help if we add a test reading an existing file used for production (say an IOP file or some sort of another) _with_ the caveat that we might want the field in the file and the field in the FM/code to have different names (I assume currently, we should do this with aliasing, right?)

Yes, the customer class can pass a vector of fields with aliased name. That is, the names of the fields passed to this class MUST match what's in the file, but the customer app can pass aliased fields to achieve the desired result.

> * If there's any way to reduce the length of the main eamxx_data_interpolation cpp file, that would be great. I remember we complained that our IO code was getting out of hand in terms of length. I notice the current file is ~450 lines. If we can bring that down to half that length (by extracting some stuff away to a util file or something) that may be nice

I don't think you can get below 450 for a class that has to handle a few different use cases. Almost 100 lines come just from the safety checks, so that leaves only 350 lines of code (including comments and empty lines for ease of readability). I highly doubt you could trim the file much (short of rm comments and all empty lines).

PR #6841, opened 2024-12-09
EAMxx: fix valgrind failures

bartgol:
Fixes valgrind tests failures in eamxx standalone tests.

---

I was able to fix our valgrind failures by doing a few mods. Namely:

- use pack size 1 in our valg build: avoids issues with `pow(UNINITED,exponent)`. The altearnative is to ensure the base is always inited, but there may be quite a few ramifications.
- set some defaults for `se_XYZ` namelist variables in HOMME before reading them, in case they are missing in the input namelist. Usually, they are missing if not used (e.g., no point in setting `se_ne` for DP runs, since `se_ne_x` and `se_ne_y` will be used), but valgrind doesn't know that. @oksanaguba I added you as a reviewer for the handful of lines in namelist mod, just in case.
- the valg test that ensures failures are detected was somewhat broken, due to compiler optimizations (I think the compiler was able to detect that the condition `i<10` was always going to be verified).

NOTE: this PR _does_ touch stuff outside of eamxx, so it will have to undergo nightly testing through next.

### DIFF
--- a/components/eamxx/scripts/test_factory.py
+++ b/components/eamxx/scripts/test_factory.py
@@ -156,6 +156,7 @@ class VALG(TestProperty):
             "Release build where tests run through valgrind",
             [("CMAKE_BUILD_TYPE", "RelWithDebInfo"),
              ("EKAT_ENABLE_VALGRIND", "True"),
+             ("SCREAM_PACK_SIZE", "1"),
              ("SCREAM_TEST_MAX_THREADS", "2")],
             uses_baselines=False,
             on_by_default=False,
--- a/components/eamxx/tests/generic/fail_check/valg_fail.cpp
+++ b/components/eamxx/tests/generic/fail_check/valg_fail.cpp
@@ -6,15 +6,18 @@ namespace scream {
 
 TEST_CASE("force_valgrind_err")
 {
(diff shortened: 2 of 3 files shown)


jgfouca:
> use pack size 1 in our valg build: 

This would mean a loss of some coverage, right? What's the alternate solution?

bartgol:
> > use pack size 1 in our valg build:
> 
> This would mean a loss of some coverage, right? What's the alternate solution?

Technically, yes. But the only alternative is to stuff the code with lots of if/ifdef, to make sure padding in packs is always valid. This is not much different from what we do for FPE.

I think, as a double safety, we could make sure that packs and views in debug are always initialized with quiet_NaN() (or, more generally, with `ekat::ScalarTraits<T>::invalid()`). This would allow us to catch errors when pack size > 1 in regular debug builds (since if we ever access those garbage values to compute _physical_ entries, we'd get nans). But I think that may require some time, and we should prob do it in a separate PR?

jgfouca (approved):
If you plan a follow-on, this is fine.

bartgol:
In next.

bartgol:
The failures in next are not due to this PR. Merging.

PR #6845, opened 2024-12-11
Convert p3_init from f90 to CXX

jgfouca:
This removes the last f90 from P3.

Change list:
* Move table read/write implementation to CXX in p3_init_impl.hpp
* p3_init now returns a P3LookupTables object with all the views inside containing the tables.
* Rename init_kokkos_tables to get_global_tables
* Rename init_kokkos_ice_lookup_tables to get_global_ice_lookup_tables
* Remove all f90 bridge code
* Move p3_tables_setup to tests dir, cleans up top p3 directory
* Fix p3_tables_setup implementation; it got broken in recent PRs but we didn't notice because it didn't get built by default
* Remove P3GlobalForFortran test struct, just call p3_init!
* Remove P3InitAP3Data test struct

Debatable change:
I really wanted to have `P3LookupTables lookup_tables` be static in p3_init so they would not have to be re-created on every p3_init call. Unfortunately, Kokkos does not allow static views to be active on program shutdown, since this occurs after Kokkos::finalize, so we would need add a p3_finalize method and ensure it was called at the end of p3. This seems burdensome and error prone, so I just decided to re-read all the tables every time p3_init is called.

There are some round-off level diffs in P3, so:
[non-BFB]


jgfouca:
Reviewers, please see "Debatable change" section in the description.

components/eamxx/src/physics/p3/impl/p3_init_impl.hpp
```diff
@@ -3,35 +3,327 @@
...
+  // read in ice microphysics table into host views. We always read these as doubles.
+  //
+
+  std::string filename = std::string(p3_lookup_base) + std::string(p3_version);
+
+  if (masterproc) {
+    std::cout << "Reading ice lookup tables in file: " << filename << std::endl;
+  }
+
+  std::ifstream in(filename);
```
bartgol, line 37: Does this mean that the tables are plain ascii now? I kind of like that, so one can always inspect the file content manually, and since they are relatively small and only used at init time, binary format is not that important.

components/eamxx/src/physics/p3/impl/p3_init_impl.hpp
```diff
@@ -3,35 +3,327 @@
...
+
+  // read header
+  std::string version, version_val;
+  in >> version >> version_val;
+  EKAT_REQUIRE_MSG(version == "VERSION", "Bad " << filename << ", expected VERSION X.Y.Z header");
+  EKAT_REQUIRE_MSG(version_val == p3_version, "Bad " << filename << ", expected version " << p3_version << ", but got " << version_val);
+
+  // read tables
+  double dum_s; int dum_i; // dum_s needs to be double to stream correctly
```
bartgol, line 46: Are these just dummy vars to swallow entries that are in the file but not used? Do we need to keep these unused entries in the file or can we just prune them, and keep only what is used?

components/eamxx/src/physics/p3/impl/p3_init_impl.hpp
```diff
@@ -3,35 +3,327 @@
...
+  vn_table_vals = vn_table_vals_nc;
+  vm_table_vals = vm_table_vals_nc;
+  revap_table_vals = revap_table_vals_nc;
+}
+
+template <bool IsRead>
+struct IoAction
+{
+  template <typename S>
+  static void action(const ekat::FILEPtr& fid, S* data, const size_t size)
```
bartgol, line 196: Up to you, but since we are doing C++ only, maybe we should just use `<fstream>` stuff, rather than the cumbersome `FILEPtr`? It seems easier to understand for new devs, and easier to maintain.

components/eamxx/src/physics/p3/impl/p3_init_impl.hpp
```diff
@@ -3,35 +3,327 @@
...
+void io_impl(const bool masterproc, const char* dir, MuRT& mu_r_table_vals, VNT& vn_table_vals, VMT& vm_table_vals, RevapT& revap_table_vals)
+{
+  using Action = IoAction<IsRead>;
+
+  if (masterproc) {
+    std::cout << (IsRead ? "Reading" : "Writing") << " lookup (non-ice) tables in dir " << dir << std::endl;
+  }
+
+  std::string extension =
+#ifdef SCREAM_DOUBLE_PRECISION
```
bartgol, line 222: Idea: since we are in the middle of rearranging our tables, why not being more verbose with names? Maybe instead of appending `8` or `4`, we could append the actual type name? Like `mu_r_table_vals_double.dat`?

Yes, it will require changing defaults in the xml, but if clarity increases, it's worth it, imho.

components/eamxx/src/physics/p3/impl/p3_init_impl.hpp
```diff
@@ -3,35 +3,327 @@
...
+struct IoAction
+{
+  template <typename S>
+  static void action(const ekat::FILEPtr& fid, S* data, const size_t size)
+  {
+    ekat::read(data, size, fid);
+  }
+};
+
+template <>
```
bartgol, line 202: Rather than using template specialization, why not doing
```
static void action (...)
{
  if constexpr (IsRead)
    ekat::read (...);
  else
     ekat::write(...);
}
```

components/eamxx/src/physics/p3/CMakeLists.txt
```diff
@@ -113,10 +111,6 @@ foreach (file IN ITEMS ${P3_TABLES})
   GetInputFile(${file})
 endforeach()
```
bartgol, line 113: I can't add a comment out of the lines you modified, so I'm writing here.

Since we have NO fortran left, we should remove the lines

```cmake
  set_target_properties(${P3_LIB} PROPERTIES
    Fortran_MODULE_DIRECTORY ${P3_LIB}/modules
  )
```
as well as the line `    ${CMAKE_CURRENT_BINARY_DIR}/${P3_LIB}/modules` inside the `target_include_directories` call (we don't generate fortran .mod files anymore)

bartgol (approved):
NICE! I love PRs where the # of rm-ed lines is larger than the # of added lines...

I have a few suggestions, in case you want to take this a step further. Not blocking though.

As for the debatable change: I think leaving as is (init for every test) is fine. No need to add complexity just to make the tests a bit faster. Our tests are reasonably fast anyways, so unless they suddenly got slow, and need the speedup of a single read, I think it's fine.

components/eamxx/src/physics/p3/impl/p3_init_impl.hpp
jgfouca: @bartgol , I should have made it more clear in the PR description that this function impl is, for the most part, not new. It's just what `init_kokkos_ice_lookup_tables` was before except moved here and renamed.

components/eamxx/src/physics/p3/impl/p3_init_impl.hpp
jgfouca: It looks like `dum_s` is used but not `dum_i`. We probably don't need the unused entries, but I don't really know the history of this file.

components/eamxx/src/physics/p3/impl/p3_init_impl.hpp
jgfouca: Good idea

components/eamxx/src/physics/p3/impl/p3_init_impl.hpp
jgfouca: This is the convention we are using all over, for better or worse.

bartgol (approved):

bartgol:
@jgfouca Before merging this, we should decide how to handle tables. Once merged, we will need to regenerate the tables, right? But old e3sm versions will still need the "old" f90-major tables. I think we should change the tables names, so that we can have both old and new tables on the server. Maybe we should add v0.1 and v1.0 at the end of names, or something like that.

jgfouca:
@bartgol , crap, I did not think of that. Yes, this change probably requires a new table version...

jgfouca:
@bartgol , I made the changes we discussed and uploaded the v2 dat8 and dat4 files to ANL.

bartgol (approved):

tcclevenger (approved):

bartgol (approved):

jgfouca:
I think all the fails are expected DIFFs, so I will go ahead and merge.

PR #6846, opened 2024-12-11
WIP: Fix an issue in EAMxx with incorrect metadata in rhist file

### LINKED ISSUE #6795: Inconsistent behavior with Max Snapshots Per File with Restarts
In EAMxx the YAML directive “Max Snapshot Per File” seems to be tripping up upon restarts and not behaving as expected.  

For example, a DPxx run that is 6 hours in duration I have an output stream with `Max Snapshots Per File: 1 `with hourly averaged output.  I have observed the following:
- CNTL: If I set the simulation to run through without restarts: it will produce 6 files as expected.
- EXP01: If I direct the simulation to run with one restart halfway through: it will produce 5 output files.  The third file in the series will contain two time slices.
- EXP02: If I direct the simulation to run with restarts each hour (hence, five restarts):  it will produce 1 output file.  The output file will contain all 6 time slices.

I have observed similar behavior in multiple DPxx simulations for both pm-cpu and pm-gpu.  In a large production run I’m doing I have it doing daily restarts.  I have an output stream with hourly output with `Max Snapshots Per File: 24`.  In this case it is putting all my data into ONE file.

I have performed one global ne30 test and noticed similar behavior with restarts and unexpected behavior with Max Snapshots Per File treatment.  Thus, this does not appear to be a DPxx specific problem but a general problem.

For a quick reproducer of EXP01, run the DYCOMS-RF01 case:
https://github.com/E3SM-Project/scmlib/blob/master/DPxx_SCREAM_SCRIPTS/run_dpxx_scream_DYCOMSrf01.csh
Set to run for 3 hours (search for `stop_n`) and set for one restart.

And direct to the following YAML:
```
%YAML 1.1
---
Averaging Type: Average
Max Snapshots Per File: 1
filename_prefix: ${CASE}.scream.hourly.avg
Fields:
  Physics PG2:
    Field Names:
    - LiqWaterPath
output_control:
  Frequency: 1
  frequency_units: nhours
```

AaronDonahue:
This commit fixes an issue during restarts that occurs with averaged type output.  The restart history file (rhist) metadata was incorrectly setup which could lead EAMxx to reopen files that already had the max number of snaps in them and continue to fill them at the restart step.

Fixes #6795

### DIFF
--- a/components/eamxx/src/share/io/scream_output_manager.cpp
+++ b/components/eamxx/src/share/io/scream_output_manager.cpp
@@ -494,9 +494,14 @@ void OutputManager::run(const util::TimeStamp& timestamp)
         if (filespecs.ftype==FileType::HistoryRestart) {
           // Update the date of last write and sample size
           write_timestamp (filespecs.filename,"last_write",m_output_control.last_write_ts,true);
-          scorpio::set_attribute (filespecs.filename,"GLOBAL","last_output_filename",m_output_file_specs.filename);
           scorpio::set_attribute (filespecs.filename,"GLOBAL","num_snapshots_since_last_write",m_output_control.nsamples_since_last_write);
-          scorpio::set_attribute (filespecs.filename,"GLOBAL","last_output_file_num_snaps",m_output_file_specs.storage.num_snapshots_in_file);
+	  if (m_output_file_specs.is_open) {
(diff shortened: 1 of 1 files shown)


AaronDonahue:
Unit tests to come in next commit.

AaronDonahue:
Closing this PR for now to get it off the PR list while I figure out why unit tests are failing.

PR #6847, opened 2024-12-11
EAMxx: fix compute-sanitizer racecheck options

bartgol:
The option `--racecheck-detect-level=error` was ignoring "warning" level messages that were in fact real issues.


bartgol:
This is not right. It exposes LOTS of false positives from kokkos.

PR #6851, opened 2024-12-15
number of ghost layers needs to be set for bilinear maps

iulian787:
number of ghost layers needs to be setup correctly for the bilinear maps;
different PE layouts should result in the same map.

[BFB]


driver-moab/main/prep_ocn_mod.F90
```diff
@@ -426,6 +427,13 @@ subroutine prep_ocn_init(infodata, atm_c2_ocn, atm_c2_ice, ice_c2_ocn, rof_c2_oc
             ! next, let us compute the ATM and OCN data transfer
             if (.not. samegrid_ao) then ! not a data OCN model
 
+               ! for bilinear maps, we need to have a layer of ghosts on source
+               nghlay = 1  ! number of ghost layers
+               ierr   = iMOAB_SetGhostLayers( mbaxid, nghlay )
+               if (ierr .ne. 0) then
```
vijaysm, line 433: I haven't tested an app that computes both conservative and bilinear in the same instance, when bilinear requires a ghost layer. Did you verify that conservative maps then remain the same? They should since we are adding more information to the mesh but I want to ensure the ghosted elements don't get counted twice. @iulian787 

The change itself looks fine to, pending MOAB changes to get the functionality working correctly.

vijaysm (approved):
Changes look fine.

PR #6854, opened 2024-12-16
remove the exact integration in cube_to_target to avoid negative weights

jsbamboo:
Fix the negative weights in cube_to_target.

In the experience of using the cube_to_target for regionally refined meshes, we encountered issues with large negative 
values for terr_target. This generally happened when the dx of the target mesh was comparable/smaller than the dx of the
 base cube.

Peter Lauritzen helped us to identify the underlying reasons for the negative weights and suggested a solution. 
The current practice is to keep the exact integration algorithm, but never enable it.
Since the original version of the judgement condition is rich in semantics, the original judgement was not deleted but 
annotated.

--------

```
In remamp.F90 change:
    else if (abs(yseg(1) -yseg(2))<fuzzy_width) then
to
    else if (.false.) then
```

### DIFF
--- a/components/eam/tools/topo_tool/cube_to_target/remap.F90
+++ b/components/eam/tools/topo_tool/cube_to_target/remap.F90
@@ -850,7 +850,8 @@ MODULE remap
 
             if (xseg(1).EQ.xseg(2))then
               slope = bignum
-            else if (abs(yseg(1) -yseg(2))<fuzzy_width) then
+            !else if (abs(yseg(1) -yseg(2))<fuzzy_width) then !remove the exact integration
+            else if (.false.) then
               slope = 0.0
             else
               slope    = (yseg(2)-yseg(1))/(xseg(2)-xseg(1))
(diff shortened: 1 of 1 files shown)


mt5555 (approved):

rljacob:
Can you describe what the PR does without referring to other github issues?

jsbamboo:
@rljacob thanks for the suggestion, the description has been updated

components/eam/tools/topo_tool/cube_to_target/remap.F90
```diff
@@ -850,7 +850,8 @@ subroutine side_integral(&
 
             if (xseg(1).EQ.xseg(2))then
               slope = bignum
-            else if (abs(yseg(1) -yseg(2))<fuzzy_width) then
+            !else if (abs(yseg(1) -yseg(2))<fuzzy_width) then !remove the exact integration
```
whannah1, line 853: @jsbamboo let's change the wording on the comment on both of the commented lines to something more verbose and clear, I'm thinking: 
"_this will enable exact integration, which was found to create problematic negative weights_"

A better solution would be to just add a logical switch controlled by a flag, but I'm fine with this solution as well.

rljacob (approved):

whannah1 (approved):

rljacob:
@mt5555 is this ready?

PR #6859, opened 2024-12-18
bug fix in use_extrasnowlayers and snow balance check 

chloewhicker:
Minor mods to fix segmentation fault errors in use_extrasnowlayers errors
that occur when the snow depth is very shallow. 

[B4B]

++++++++++++++++++++++++++++++++++++++++++++++

Minor mods to fix a bug in the snow balance mod - soil was pushing frozen water to snow balance causing very small errors. Now this flux is not included in the snow pack but still sent to MOSART as ice runoff. Discussed these mods with @bishtgautam

### DIFF
--- a/components/elm/src/biogeophys/BalanceCheckMod.F90
+++ b/components/elm/src/biogeophys/BalanceCheckMod.F90
@@ -553,6 +553,13 @@ contains
              write(iulog,*)'qflx_snwcp_ice   = ',qflx_snwcp_ice(indexc)*dtime
              write(iulog,*)'qflx_snwcp_liq   = ',qflx_snwcp_liq(indexc)*dtime
              write(iulog,*)'qflx_sl_top_soil = ',qflx_sl_top_soil(indexc)*dtime
+             write(iulog,*)'qflx_snow_melt   = ',qflx_snow_melt(indexc)*dtime
+             write(iulog,*)'frac_sno_eff     = ',frac_sno_eff(indexc)*dtime
+             write(iulog,*)'qflx_rain_grnd_col= ',qflx_rain_grnd_col(indexc)*dtime
+             write(iulog,*)'qflx_snow_grnd_col= ',qflx_snow_grnd_col(indexc)*dtime
+             write(iulog,*)'qflx_h2osfc_to_ice= ',qflx_h2osfc_to_ice(indexc)*dtime
+             write(iulog,*)'qflx_snow_h2osfc= ',qflx_snow_h2osfc(indexc)*dtime
+
              if (create_glacier_mec_landunit) then
                 write(iulog,*)'qflx_glcice_frz  = ',qflx_glcice_frz(indexc)*dtime
              end if
--- a/components/elm/src/biogeophys/BalanceCheckMod.F90
+++ b/components/elm/src/biogeophys/BalanceCheckMod.F90
@@ -534,6 +534,40 @@ contains
                ' col_pp%itype= ',col_pp%itype(indexc), &
                ' lun_pp%itype= ',lun_pp%itype(col_pp%landunit(indexc)), &
                ' errh2osno= ',errh2osno(indexc)
+                    write(iulog,*)'nstep            = ',nstep
+             g = col_pp%gridcell(indexc)
+             l = col_pp%landunit(indexc)
+             write(iulog,*)'errh2osno        = ',errh2osno(indexc)
+             write(iulog,*)'snl              = ',col_pp%snl(indexc)
+             write(iulog,*)'h2osno           = ',h2osno(indexc)
+             write(iulog,*)'h2osno_old       = ',h2osno_old(indexc)
+             write(iulog,*)'snow_sources     = ',snow_sources(indexc)
+             write(iulog,*)'snow_sinks       = ',snow_sinks(indexc)
+             write(iulog,*)'qflx_prec_grnd   = ',qflx_prec_grnd(indexc)
+             write(iulog,*)'qflx_sub_snow    = ',qflx_sub_snow(indexc)
+             write(iulog,*)'qflx_evap_grnd   = ',qflx_evap_grnd(indexc)
+             write(iulog,*)'qflx_top_soil    = ',qflx_top_soil(indexc)
+             write(iulog,*)'qflx_dew_snow    = ',qflx_dew_snow(indexc)
+             write(iulog,*)'qflx_dew_grnd    = ',qflx_dew_grnd(indexc)
+             write(iulog,*)'qflx_snwcp_ice   = ',qflx_snwcp_ice(indexc)
+             write(iulog,*)'qflx_snwcp_liq   = ',qflx_snwcp_liq(indexc)
+             write(iulog,*)'qflx_sl_top_soil = ',qflx_sl_top_soil(indexc)
+             write(iulog,*)'qflx_snow_melt   = ',qflx_snow_melt(indexc)
+             write(iulog,*)'frac_sno_eff     = ',frac_sno_eff(indexc)
+             write(iulog,*)'qflx_rain_grnd_col= ',qflx_rain_grnd_col(indexc)
+             write(iulog,*)'qflx_snow_grnd_col= ',qflx_snow_grnd_col(indexc)
+             write(iulog,*)'qflx_h2osfc_to_ice= ',qflx_h2osfc_to_ice(indexc)
+             write(iulog,*)'qflx_snow_h2osfc= ',qflx_snow_h2osfc(indexc)*dtime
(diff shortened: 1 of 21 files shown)


proteanplanet:
Add myself as a reviewer for final sign-off since this is contributing to the SORRM configuration and BlueTip simulations.

proteanplanet:
Please complete BFB testing as early as possible in the new year.

matthewhoffman:
With the removal of the use_firn_percolation_and_compaction option, the only way to get the more realistic snow densification and percolation physics is to use the deep firn vertical coordinate, right?  Can I get elaboration on why this reduced flexibility was needed?  While it makes sense to me that if you have extra snow layers, you likely would always want the advanced firn physics enabled, and that we want to retain the ability to run backwards compatible shallow snow without advanced firn physics, it seems to me that the option to run shallow snow with advanced firn physics would be useful going forward.

rljacob:
note: needs approvals

chloewhicker:
Thanks Rob - I am still working out some news kinks I found in this PR. Let me know if you'd prefer I close and reopen when I've got it cleaned up.

rljacob:
If it will take more then a week, please close and then reopen.

chloewhicker:
Hi all -I followed @bishtgautam's advice and am only altering the water balance and snow balance equations if use_extrasnowlayers / use_firn_percolation_and_compaction. The B4B tests came back identical and a few test runs have completed. So this is ready for review - please let me know if you have questions or changes or want additional testing.

components/elm/src/biogeophys/BalanceCheckMod.F90
```diff
@@ -490,16 +493,15 @@ subroutine ColWaterBalanceCheck( bounds, num_do_smb_c, filter_do_smb_c, &
...
+                     snow_sources(c) = (qflx_snow_grnd_col(c) - qflx_snow_h2osfc(c) ) &
+                          + frac_sno_eff(c) * (qflx_rain_grnd_col(c) &
+                          + qflx_dew_snow(c) + qflx_dew_grnd(c) ) + qflx_h2osfc_to_ice(c)
 
-                      snow_sinks(c) = frac_sno_eff(c) * (qflx_sub_snow(c) + qflx_evap_grnd(c)) &
-                           + qflx_snwcp_ice(c) + qflx_snwcp_liq(c) &
-                           + qflx_snow_melt(c) + qflx_sl_top_soil(c)
+                     snow_sinks(c) = frac_sno_eff(c) * (qflx_sub_snow(c) + qflx_evap_grnd(c)) &
+                          + qflx_snwcp_ice(c) + qflx_snwcp_liq(c) &
+                          + qflx_snow_melt(c) + qflx_sl_top_soil(c)
```
bishtgautam, line 502: It appears there aren't any real changes in the code, except spaces have been deleted. Can you please revert the changes?

bishtgautam (approved):
@chloewhicker, The PR looks good to me and I only have minor review comments for you to consider.

chloewhicker:
This PR also contains @stephenprice's new config - Steve is that okay or do you want me to remove it?

matthewhoffman:
@chloewhicker , to your question, Steve's new compset is in a separate PR here: https://github.com/E3SM-Project/E3SM/pull/6907, so it would be better to remove from this PR.

matthewhoffman:
@chloewhicker , can you update the PR description after the restoration of the use_firn_percolation_and_compaction option?

components/elm/src/biogeophys/SnowHydrologyMod.F90
```diff
@@ -670,7 +670,7 @@ subroutine SnowCompaction(bounds, num_snowc, filter_snowc, &
                       if (bi > dm) ddz1 = ddz1*exp(-46.0e-3_r8*(bi-dm))
                    else
                       ddz1_fresh = (-grav * (burden(c) + wx/2._r8)) / &
-                                   (0.007_r8 * bi**(4.75_r8 + td/40._r8))
+                                   (0.007_r8 * min(max(bi,dm),denice)**(4.75_r8 + min(td,0._r8)/40._r8))
```
matthewhoffman, line 673: Is this change related to the bug fix in use_extrasnowlayers or the snow balance check?

components/elm/src/biogeophys/SoilTemperatureMod.F90
```diff
@@ -1623,6 +1625,7 @@ subroutine Phasechange_beta (bounds, num_nolakec, filter_nolakec, dhsdT, &
 
                      if (imelt(c,j) == 1 .AND. j < 1) then
                         qflx_snomelt(c) = qflx_snomelt(c) + max(0._r8,(wice0(c,j)-h2osoi_ice(c,j)))/dtime
+                        qflx_snomelt_lyr(c,j) = qflx_snomelt(c)
```
matthewhoffman, line 1628: Won't this include values from the previous layers that were looped over?  Should this instead be `max(0._r8,(wice0(c,j)-h2osoi_ice(c,j)))/dtime`?

components/elm/src/data_types/ColumnDataType.F90
```diff
@@ -5762,6 +5766,11 @@ subroutine col_wf_init(this, begc, endc)
     !-----------------------------------------------------------------------
     ! initialize history fields for select members of col_wf
     !-----------------------------------------------------------------------
+    this%qflx_snofrz_lyr(begc:endc,-nlevsno+1:0) = spval         
+      call hist_addfld2d (fname='QSNOFRZ_LYR', units='kg/m2/s', type2d='levsno',&
```
matthewhoffman, line 5770: Can you explain why this history field is being added?  I don't have a problem with it, I'm just trying to understand where it came from since this was not a new field being added as part of the other changes.

components/elm/src/data_types/ColumnDataType.F90
```diff
@@ -5687,6 +5689,7 @@ subroutine col_wf_init(this, begc, endc)
     allocate(this%qflx_evap_grnd         (begc:endc))             ; this%qflx_evap_grnd       (:)   = spval
     allocate(this%qflx_snwcp_liq         (begc:endc))             ; this%qflx_snwcp_liq       (:)   = spval
     allocate(this%qflx_snwcp_ice         (begc:endc))             ; this%qflx_snwcp_ice       (:)   = spval
+    allocate(this%qflx_ice_runoff_xs     (begc:endc))             ; this%qflx_ice_runoff_xs   (:)   = 0._r8
```
matthewhoffman, line 5692: It looks like this variable is initialized to zero in `components/elm/src/biogeophys/SnowHydrologyMod.F90`.  Should this line initialize to `spval` as is done for all the other variables in this module?

matthewhoffman (commented):
@chloewhicker , I'm reviewing by inspection as one not familiar with this code.  I've made a few comments/questions, but otherwise don't have any concerns.

components/elm/src/biogeophys/SnowHydrologyMod.F90
chloewhicker: this is the bug fix in use_extrasnowlayers

components/elm/src/data_types/ColumnDataType.F90
chloewhicker: The variable already existed as a restart variable, I just made a history field we can output to history files with it

components/elm/src/biogeophys/SoilTemperatureMod.F90
chloewhicker: yes - good catch!

components/elm/src/data_types/ColumnDataType.F90
chloewhicker: This needs to be initialized to zero or the spval is propagated into calculations - if that is not OK I can go find where to set it to zero, just let me know.

matthewhoffman (approved):

proteanplanet (approved):
Approved based on visual inspection and extensive testing and feedback.

proteanplanet:
Tagging @darincomeau and @cbegeman for inclusion in SORRM V3.

chloewhicker:
Testing of these changes + the other deep firn changes is at year 50 with the simulation page here https://acme-climate.atlassian.net/wiki/spaces/CM/pages/4875550721/V3.LR+Deep+Firn+Simulation

rljacob:
@bishtgautam didn't you find that this was non-BFB for some tests?

bishtgautam:
It was non-BFB because of the new fields being added in ELM h0 files.

rljacob:
Oh that was it.  And only for cases that ran long enough.

PR #6860, opened 2024-12-19
Add Initial Condition Handling to EAMxx DAG Generator

mjs271:
Finishes PR [#3101](https://github.com/E3SM-Project/scream/pull/3101) originally submitted to the scream repo the atmospheric DAG generator (atmosphere_process_dag.xpp) and atmosphere driver to get things working again. Namely,

There are now 2 checkpoints during AtmosphereDriver::initialize() at which we build the most complete DAG available and write it to file.
This is intended to serve as a diagnostic tool in case a build crashes, giving the user information about the created/initialized processes or fields.
These checkpoints occur at the end of the following functions, and the associated DAG is written out to a .dot file in the run directory:
  1. create_fields()--scream_atm_createField_dag.dot
  2. initialize_atm_procs()--scream_atm_initProc_dag.dot

I've added some code that enables the DAG to properly handle initial conditions and determine whether they are missing at the time when he DAG is generated, and if so this is indicated in the corresponding node (see images in [original PR](https://github.com/E3SM-Project/scream/pull/3101) for examples).
In the initial DAG, (within initialize_atm_procs()), an extra box is added to the DAG with a disclaimer that any fields indicated as missing may be provided by the forthcoming initial condition file.
A node for the Initial Condition is now displayed on the DAG with edges leading to the "Begin of Atm Time Step" node adn any nodes for a process that uses and does not change a field from the initial condition.
I've made some minor formatting changes for readability and to more clearly indicate the different types of objects represented on the DAG--e.g., the Begin/End of Atm. Timestep nodes are colored differently, and fields are printed in green when provided directly by the initial condition at initialization.
I've also fenced-off the write_dag() statements to ensure they are only printed by the main MPI thread, since I noticed occasional formatting errors due to a presumptive race condition.
Additionally, when running as an eamxx standalone test, the filenames of the DAGs are tagged with `.np<N>` to avoid conflicts writing to file.
Lastly, I added a `RESOURCE_LOCK` to the standalone test `mam4_srf_online_emiss_mam4_constituent_fluxes` that appeared to be failing due to incorrect order of running the tests being compared.

**Note:** Closes [E3SM/scream Issue #1869](https://github.com/E3SM-Project/scream/issues/1869).

### DIFF
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -685,6 +685,21 @@ void AtmosphereDriver::create_fields()
     fm->add_to_group(fid.name(),"RESTART");
   }
 
+  auto& driver_options_pl = m_atm_params.sublist("driver_options");
+  const int verb_lvl = driver_options_pl.get<int>("atmosphere_dag_verbosity_level",-1);
+  if (verb_lvl>0) {
+    // now that we've got fields, generate a DAG with fields and dependencies
+    // NOTE: at this point, fields provided by initial conditions may (will)
+    // appear as unmet dependencies
+    AtmProcDAG dag;
+    // First, add all atm processes
+    dag.create_dag(*m_atm_process_group);
+    // Write a dot file for visualization
+    if (m_atm_comm.am_i_root()) {
+      dag.write_dag("scream_atm_createField_dag.dot", std::max(verb_lvl,0));
+    }
+  }
+
   m_ad_status |= s_fields_created;
 
   // If the user requested it, we can save a dictionary of the FM fields to file
@@ -900,28 +915,6 @@ initialize_fields ()
   // See the [rrtmgp active gases] note in share/util/eamxx_fv_phys_rrtmgp_active_gases_workaround.hpp
   if (fvphyshack) {
     TraceGasesWorkaround::singleton().run_type = m_run_type;
-  }
-
-  // See if we need to print a DAG. We do this first, cause if any input
-  // field is missing from the initial condition file, an error will be thrown.
-  // By printing the DAG first, we give the user the possibility of seeing
-  // what fields are inputs to the atm time step, so he/she can fix the i.c. file.
-  // TODO: would be nice to do the IC input first, and mark the fields in the
-  //       DAG node "Begin of atm time step" in red if there's no initialization
-  //       mechanism set for them. That is, allow field XYZ to not be found in
-  //       the IC file, and throw an error when the dag is created.
-
-  auto& driver_options_pl = m_atm_params.sublist("driver_options");
-  const int verb_lvl = driver_options_pl.get<int>("atmosphere_dag_verbosity_level",-1);
-  if (verb_lvl>0) {
-    // Check the atm DAG for missing stuff
-    AtmProcDAG dag;
-
-    // First, add all atm processes
-    dag.create_dag(*m_atm_process_group);
-
-    // Write a dot file for visualization
-    dag.write_dag("scream_atm_dag.dot",std::max(verb_lvl,0));
   }
 
   // Initialize fields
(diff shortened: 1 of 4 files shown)


mjs271:
Looks like I can't add reviewers, so tagging those from original PR @bartgol @tcclevenger @jgfouca @mahf708 @AaronDonahue

components/eamxx/tests/multi-process/physics_only/mam/mam4_srf_online_emiss_mam4_constituent_fluxes/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ CreateADUnitTest(${TEST_BASE_NAME}
   LABELS physics mam4_srf_online_emiss mam4_constituent_fluxes
   MPI_RANKS ${TEST_RANK_START} ${TEST_RANK_END}
   FIXTURES_SETUP_INDIVIDUAL ${FIXTURES_BASE_NAME}
+  PROPERTIES RESOURCE_LOCK ${FIXTURES_BASE_NAME}
```
bartgol, line 12: Why are you serializing these tests? Is this b/c of the generated dag? Imho, we can just let them run in ||, even if they overwrite each other's dag...

bartgol (commented):
Looks good. I have one question on the test modification.

components/eamxx/tests/multi-process/physics_only/mam/mam4_srf_online_emiss_mam4_constituent_fluxes/CMakeLists.txt
mjs271: This one was unrelated to the dag, though kept crashing while I was debugging. Could be idiosyncratic to whichever machine I was on, this appeared to fix it. I'm also ok with reverting and keeping an eye out

bartgol: Was it only this test that failed? You could modify its input.yaml file to set the dag verbosity to 0 (or whatever turns it completely off). That is, assuming the dag was the issue...

bartgol (approved):

components/eamxx/tests/multi-process/physics_only/mam/mam4_srf_online_emiss_mam4_constituent_fluxes/CMakeLists.txt
mjs271: @bartgol--getting back to this after the break, and it appears that the failures I was seeing were due to some transient computational **_weirdness_**. Tests seem to be passing without the lock, but I'll run a few more tests on my end to confirm

bartgol:
@mjs271 I think you may have rebased onto the eagle-project's version of master (or onto your local version of master), rather that onto e3sm-project's master.

mjs271:
oof! I'll push a fix shortly

bartgol (approved):

mjs271:
I think this may be noted above, but to refresh:

- The test failures for `feamxx-sa/gcc-cuda/[...]` (`mam4_aero_microphys_standalone_baseline_cmp`) are expected due to this PR not including the changes from #6867, where the failing tests are fixed.
- The other failure for `eamxx-v1 / cpu-gcc / SMS_Ln3_P4.ne4pg2_oQU480.F2010-MMF2` appears to be due to a missing input file and should be unrelated to the substance of this PR.

rljacob:
Is this ready?

bartgol:
> Is this ready?

Yes. But I want to rerun tests now that we merged the mam4xx fix, to verify we are no longer non-deterministic in those tests.

bartgol:
The fails are unrelated. I'm merging.

PR #6867, opened 2024-12-21
Fast-forward MAM4xx submodule to Add Fixes for Test Failures

mjs271:
This PR fixes the MAM4xx interface for wet scavenging--namely, addressing failing tests and an associated non-determinism issue.

The issues existed primarily on the MAM4xx side of things, so this PR is mostly a fast-forward of the MAM4xx submodule. However, minor edits to the EAMxx interface code were required in order to properly call `mam4::wetdep::aero_model_wetdep()`.

##### Note:
I'm keeping this a draft for the moment because this is currently pointing to my MAM4xx branch, and I need a review on the MAM4xx PR before I can merge it to main.


components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
```diff
@@ -406,6 +406,10 @@ void call_function_dropmixnuc(
               }
             });
         team.team_barrier();
+        // HACK: dropmixnuc() requires the parameter enable_aero_vertical_mix,
+        //       so we define it here until we have a better idea of where it
+        //       might come from
+        const bool enable_aero_vertical_mix = true;
```
bartgol, line 412: Is this ALWAYS going to be true? The fact that `dropmixnuc` has it as an input arg makes me wonder if there are cases we foresee needing this to be false? If it's ALWAYS going to be true, this is fine. If we foresee it being false sometimes, then you could maybe add it as an input to `call_function_dropmixnuc`, and have the process interface read it from input file (defaulting to `true` if not found).

mjs271: Short answer: for the purposes of EAMxx/E3SM, **yes it will always be true**.

If I'm paraphrasing correctly (@singhbalwinder), this is a temporary switch to resolve some differences in the mam4xx functionality between what is in e3sm and the parallel development in our forked repo that is meant to simplify and speedup the process of development and review on our side.

bartgol (approved):

singhbalwinder (approved):

bartgol:
All tests passing except for the expected failures. I'm going to merge and regen baselines. We'll see right away if this fixed things, by looking at other PRs.

PR #6868, opened 2024-12-22
Fix potential uninitialised variable bug in E3SM buildlib files.

mpaiao:
Add an else clause to two blocks in buildlib that are missing one.

[BFB]

-------

At least with my computer, the compilation was issuing an "uninitialised risk" error. My understanding is that this risk occurs because variable `netcdf_args` was defined inside an `if`/`elif` block that had no `else`. My edits define a dummy value for `netcdf_args` that is empty for the `else` case.


rljacob:
Please remove the addition of .DS_store and the change to fates/main/FatesHistoryInterfaceMod.F90

mpaiao:
> Please remove the addition of .DS_store and the change to fates/main/FatesHistoryInterfaceMod.F90

Done!

jgfouca (approved):
Thanks!

PR #6869, opened 2024-12-24
Bug fix for changing orodrag tuning parameters via namelist

jinboxie:
This is a bug fix for the latest orodrag parameterization merged into master. The setup of the parameter tuning for the schemes is intended to read in orodrag parameters such as od_ls_ncleff, od_bl_ncd, od_ss_sncleff into the model for tuning via phys_control.F90. However, the orodrag parameters are set in a separate place spot in atm_in (e.g. oro_drag_nl), and thus a separate oro_drag_readnl is added in od_common.F90 to read these parameters in atm_in. It is tested and works well both use the new od schemes or the default schemes.

----------------------

Git log
Fixed the namelist bug for changing tuning par in oro_drag.	
        modified:   control/runtime_opts.F90
	modified:   physics/cam/clubb_intr.F90
	modified:   physics/cam/gw_drag.F90
	modified:   physics/cam/od_common.F90
	modified:   physics/cam/phys_control.F90

[BFB]

### DIFF
--- a/components/eam/src/control/runtime_opts.F90
+++ b/components/eam/src/control/runtime_opts.F90
@@ -246,6 +246,7 @@ contains
    use uwshcu,              only: uwshcu_readnl
    use pkg_cld_sediment,    only: cld_sediment_readnl
    use gw_drag,             only: gw_drag_readnl
+   use od_common,           only: oro_drag_readnl
    use qbo,                 only: qbo_readnl
    use iondrag,             only: iondrag_readnl
    use phys_debug_util,     only: phys_debug_readnl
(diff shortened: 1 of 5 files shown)


whannah1 (approved):

rljacob:
So the bug was that the namelist parameters were not being read in?

jinboxie:
> So the bug was that the namelist parameters were not being read in?

Hi Rob.
Yes, this is the bug. The parameter were pre-filled with default values. So with the bug, the model will just run with the default value instead of the namelist inputs.
Jinbo

wlin7 (approved):
Code changes look good. atm_developer tests BFB. The namelist option verified effective when varying via user_nl.

PR #6873, opened 2025-01-04
Add MALI ocean extrapolation and fix to RK scheme

matthewhoffman:
This PR updates the version of MALI from the MALI-Dev fork and primarily brings in two important features needed for Greenland coupling work:
* ability to extrapolate ocean thermal forcing from a masked region of valid ocean data to the current MALI ice extent
* a critical bug fix for the RK timestepping scheme required for it to work correctly
* changes to handling of masks in the RK and FCT schemes to keep them consistent
Also included are some changes to the subglacial hydrology model, which is not currently used in E3SM simulations.


matthewhoffman:
@jonbob , this still needs namelist default updates to be generated.

jonbob:
Passes:
* e3sm_landice_developer using gnu on chrysalis with expected NML and regular DIFFs

merged to next

PR #6874, opened 2025-01-07
Homme(xx)/SL: Enhanced trajectory method.

ambrad:
This PR brings in a new feature that (1) increases accuracy of semi-Lagrangian tracer transport's trajectory calculations and (2) permits flexible trade-off between the trajectory accuracy and speed. This PR has the following parts:
- F90 dycore support, with unit tests (principally in sl_advection.F90)
- C++ dycore support, with unit tests (principally in ComposeTransportImplEnhancedTrajectory.cpp)
- unit test driver updates: compose_ut.cpp
- two new standalone-Homme tests, one each for the F90 and C++ dycores
- new standalone-Homme transport test module for convergence testing: fully 3D, space-and-time-dependent surface pressure (dcmip2012_test1_conv_mod.F90)
- CIME-based ERS tests, one each, for EAM and EAMxx
- cleanup of a timer issue orthogonal to this PR: see  commit "Hommexx: Rework skipping timers in first step."
- updates to Homme machine files for Perlmutter
- fix C++ dycore's handling of prescribed winds: had to move down in the call stack to match the F90 dycore

[non-BFB] due to two new CIME tests, two new standalone-Homme tests, and two modified standalone-Homme tests; otherwise BFB


ambrad:
This comment will be updated with testing results.

oksanaguba:
This needs much more extensive explanation on what's new and how to use the new SL (and advantages).
Could you please create a confluence page with these details? i assume you have a lot of plots, too, add those there as well.

ambrad:
> Could you please create a confluence page with these details? i assume you have a lot of plots, too, add those there as well.

Yes, I'm working on that. I'll post a link when I have a sufficiently useful draft.

oksanaguba:
re `fix C++ dycore's handling of prescribed winds: had to move down in the call stack to match the F90 dycore` -- was it a bug and/or did you move the call for new functionality?

please list all new/modified tests explicitly with more explanation behind them. homme modified tests -- are they for your new functionality, or are they modified because moved prescribed winds call is nonbfb?

oksanaguba:
oh sounds like there are also new unit tests?

oksanaguba (changes requested):
requires extensive documentation on confluence before being approved

ambrad:
I'll describe the tests on Confluence.

> homme modified tests -- are they for your new functionality, or are they modified because moved prescribed winds call is nonbfb?

This refers to a pre-existing H/Hxx SL BFB test pair that I've switched to one of the new test flow fields. The answers change since the flow field is different.

ambrad:
@oksanaguba I put a link to a Confluence page above. I'll add draft material today.

ambrad:
> oh sounds like there are also new unit tests?

Yes, quite a number of them. They are called from the compose_ut.cpp unit test driver (as usual) and are in sl_advection.F90 and ComposeTransportImplEnhancedTrajectory.cpp.

ambrad:
> please list all new/modified tests explicitly with more explanation behind them.
```
   41 - thetah-sl-test11conv-r1t2-cdr20 (Failed)    old test but with a slightly modified flow field
   42 - thetah-sl-test11conv-r0t1-cdr30-rrm (Failed)    old test but with a slightly modified flow field
   47 - thetah-sl-test11conv-r0t1-cdr30-rrm-kokkos (Failed)    Hommexx equivalent of above
   44 - thetah-sl-testconv-3e (Failed)    new test with new flow field; uses the new trajectory method
   48 - thetah-sl-testconv-3e-kokkos (Failed)    Hommexx equivalent of above
```

components/homme/src/share/compose/compose_cedr_sl_run_global.cpp
```diff
@@ -5,6 +5,76 @@
 namespace homme {
 namespace sl {
 
+template <int np_, typename MT>
+void run_relaxed_local (CDR<MT>& cdr, const Data& d, Real* q_min_r,
```
bartgol, line 9: Can you add some comment (inline or at the top of the fcn) regarding what's happening in here?

bartgol: Perhaps also for the run_global fcn. I am not familiar with the algo, which doesn't help, but the code is a bit dense, which makes it hard to read anyways. Some comments (even just a big picture, not necessarily fine-grained) may help those who one day may have to maintain this.

components/homme/src/share/compose/compose_homme.cpp
```diff
@@ -31,46 +30,109 @@ void TracerArrays<MT>::alloc_if_not () {
...
+  }
+# pragma omp barrier
+# endif
+#endif  
+}
+
+template <typename MT>
+void sl_h2d (TracerArrays<MT>& ta, bool transfer, Real* dep_points, Int ndim) {
+#if defined COMPOSE_PORT
+# if defined COMPOSE_HORIZ_OPENMP
```
bartgol, line 91: If COMPOSE_PORT is defined, then we're building for the kokkos dycore, right? In that case we should not have horiz openmp, right? I thought horiz openmp was used in the f90 dycore only...

bartgol: Same in a few other places.

bartgol (commented):
My knowledge on the SL transport and the compose lib is very limited, so it was hard to do a good review. I would rely on folks from the compose proj for a better review. 

That said, I think the code could use the injection of a fair amount of comments. The code is very "dense", which makes it hard to read. I know you are owning this code (and I'm glad you are), but for the sake of who may come next to maintain/modify the code (should you move to a different project), it would help a lot to have more inline guidance.

components/homme/src/share/compose/compose_homme.cpp
ambrad: The Kokkos code path is supported in the F90 dycore; it's just not used in practice because of array ordering (requiring transposes and extra memory) and needing to support HORIZ_OPENMP on CPUs.

components/homme/src/share/compose/compose_cedr_sl_run_global.cpp
ambrad: I'll remove these CEDR additions. They correspond to Sect. 2.2.4 of https://gmd.copernicus.org/articles/15/6285/2022/gmd-15-6285-2022.html. I had these lying around and unwisely added them to this already lengthy PR.

components/homme/src/share/cxx/ComposeTransportImplEnhancedTrajectory.cpp
```diff
@@ -0,0 +1,2254 @@
...
+  const auto f = KOKKOS_LAMBDA (const cti::MT& team) {
+    KernelVariables kv(team);
+    const int ie = kv.ie;
+    const auto wrk1 = Homme::subview(buf1a, kv.team_idx);
+    const auto wrk2 = Homme::subview(buf1b, kv.team_idx);
+    const auto wrk3 = Homme::subview(buf1c, kv.team_idx);
+    const auto wrk4 = Homme::subview(buf1d, kv.team_idx);
+    const auto vwrk = Homme::subview(buf2a, kv.team_idx);
+    // Reconstruct Lagrangian levels at t1 on arrival column:
+    //     eta_arr_int = I[eta_ref_mid([0,eta_dep_mid,1])](eta_ref_int)
```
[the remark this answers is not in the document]
ambrad: I'll add comments. Re: these specific questions:
* `I[y(x)](xi)` means an interpolant constructed from `y(x)` and evaluated at `xi`.
* `[0,eta_dep_mid,1]` is the concatenation of the eta departure midpoint values with 0 and 1.

components/homme/src/share/cxx/ComposeTransportImplEnhancedTrajectory.cpp
```diff
@@ -0,0 +1,2254 @@
...
+    GPTLstart("compose_floating_dep_pts");
+    interp_departure_points_to_floating_level_midpoints(*this, np1);
+    Kokkos::fence();
+    GPTLstop("compose_floating_dep_pts");
+  }
+
+  GPTLstop("compose_calc_enhanced_trajectory");
+}
+
+// Testing.
```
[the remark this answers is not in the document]
ambrad: I like to keep unit tests collocated with the routines they are testing. However, that is just my own opinion. Where would you like me to move them? test_execs/thetal_kokkos_ut?

components/homme/src/share/compose/compose_slmm_islmpi.hpp
```diff
@@ -251,6 +259,40 @@ void deep_copy (FixedCapList<T, DTD>& d, const FixedCapList<T, DTS>& s) {
 #endif
 }
 
+template <typename T>
+struct FixedCapListHostOnly {
```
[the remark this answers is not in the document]
ambrad: It's true, but I have a general FixedCapList for most use cases on device and host. I made this HostOnly version to avoid host-device warnings having to do with mpi::Request. I could use std::vector, but (1) I like having slmm_kernel_assert_high in the impl and (2) I can use the same user code instead of having to switch it to std::vector syntax.

components/homme/src/share/cxx/ComposeTransportImplEnhancedTrajectory.cpp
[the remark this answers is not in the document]
ambrad: Actually, I would at least like to keep the tests in the main directory. I really like being able to call a unit-test routine for a class anywhere I want.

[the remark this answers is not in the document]
ambrad: Mods pushed.

components/homme/src/share/compose/compose_cedr_sl_run_global.cpp
ambrad: Mods pushed.

components/homme/src/share/cxx/ComposeTransportImplEnhancedTrajectory.cpp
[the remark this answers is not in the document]
ambrad: Mods pushed.

components/homme/src/share/cxx/ComposeTransportImplEnhancedTrajectory.cpp
```diff
@@ -0,0 +1,2254 @@
...
+  assert(range.start ==  0);
+  assert(range.end   == ni);
+  const auto f = [&] (const int k) {
+    yi[k] = linterp(n, x, y, k + x_idx_offset, xi[k]);
+  };
+  Kokkos::parallel_for(range, f);
+}
+
+KOKKOS_FUNCTION void
+eta_interp_eta (const KernelVariables& kv, const int nlev,
```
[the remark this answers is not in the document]
ambrad: Mods pushed.

ambrad:
@bartgol I've responded to all your comments. In most cases I've pushed associated code mods. In some cases I've explained why something is the way it is.

components/homme/src/share/compose/compose_homme.cpp
bartgol: but the cmake logic seems to be such that COMPOSE_PORT is only true for the compose_f90 lib, which is the only one linked against the f90 targets. Maybe I'm misreading the cmake logic though...

ambrad: (For the composec++ lib, not the composef90 one, but I get your point.) Yes, that's right, because in practice we don't want to use COMPOSE_PORT for the F90 dycore. That is, nobody would purposely build a dycore that way.

But it does actually work. For such a niche thing, I haven't wanted to waste testing resources to keep it working, but I use the capability when doing dev work on occasion. In short, I do want to retain lines of the sort you're highlighting even though our automated builds never activate them. Note the lines you highlighted correspond to a bridge routine to do the transposes. (Detail: The bridge routine is also used in compose_ut unit tests for F90-vs-C++ comparisons, but of course in this case HORIZ_OPENMP is, as you pointed out, always false.)

bartgol (approved):

ambrad:
@oksanaguba 

> requires extensive documentation on confluence before being approved

I've added a Confluence page and more comments to the code. See in particular the comments starting [here](https://github.com/ambrad/E3SM/blob/d8f088f4d3c5d0eef690320d6e8ef22ddd539cfe/components/homme/src/share/cxx/ComposeTransportImplEnhancedTrajectory.cpp#L78).

I responded to all your points above.

Is there anything more you would like before approving? I'd like to start merging this PR on Monday. Thanks.

oksanaguba (approved):

PR #6875, opened 2025-01-07
EAMxx: create rrtmgp allsky baseline target with CreateUnitTest utility

bartgol:
Fix issue in eamxx build system when we generate baselines using shared libs.

---

The cmake utility links (indirectly) against scream_io, which avoids link errors in case of shared libraries. The errors are due to the fact that scream_share _does_ use scream_io stuff, but the rrtmgp executable was not linking against it (directly or indirectly). However, since the exec was indeed not using ANY of the scream_io symbols, static linking is fine, but dynamic linking is not (one of the key differences in static vs dynamic linking).


tcclevenger (approved):

bartgol:
Just the expected failures from MAM as in master. Merging.

PR #6879, opened 2025-01-08
More RRTMGP performance work

jgfouca:
Change list:
1) Changes default RRMTGP backend to Kokkos
2) Adds new testmods for selecting RRTMGP backend
3) All kernels in rrtmgp interface can now be timed
4) Detranspose dimensions in kernels
5) Use a faster approach for getting random cldx (NON BFB!)
6) Update rrtmgp submodule

I've attached a screenshot of my custom kernel profiler:
Screenshot 2025-01-08 at 3 04 10 PM

Changepct is difference between YAKL and Kokkos versions of the kernel. Anything less than 100 means kokkos is faster and vice versa. Significant (>25%) speedups are highlighted green; the opposite are highlighted red. The overall time spent in run_impl is about 20% worse with Kokkos. I cannot yet account for this ~3-4 second loss of performance because the overall time spent in kernels is already less with Kokkos.

NOTE: The screenshot and comment above are no longer true. We have parity with YAKL performance now.

[non-BFB] for anything that uses rrtmgp. The random number generator is different.


components/eamxx/cime_config/testdefs/testmods_dirs/scream/kokkos_rrtmgp/shell_commands
```diff
@@ -0,0 +1,3 @@
+./xmlchange --append SCREAM_CMAKE_OPTIONS='SCREAM_RRTMGP_ENABLE_YAKL Off'
```
bartgol, line 1: Since Kokkos is the default, do we need this testmod?

components/eamxx/src/physics/rrtmgp/rrtmgp_test_utils.hpp
```diff
@@ -88,7 +88,7 @@ static void dummy_clouds(
   // put them in 2/3 of the columns since that's roughly the total cloudiness of earth.
   // Set sane values for liquid and ice water path.
   // NOTE: these "sane" values are in g/m2!
-  Kokkos::parallel_for( MDRP::template get<2>({nlay,ncol}) , KOKKOS_LAMBDA (int ilay, int icol) {
```
bartgol, line 91: Does the new iteration pattern work best for kokkos? That is, if one wants to use YAKL, should they revert back to a lay-col iteration (rather than col-lay)?

bartgol: Same question for all the swaps in the other files below

bartgol: Also, do the percentage changes in your screenshot refer to a yakl/kokkos comparison where both use the same layout? Or did you compare "the best layout for kokkos" against "the best layout for yakl"?

components/eamxx/cime_config/testdefs/testmods_dirs/scream/kokkos_rrtmgp/shell_commands
jgfouca: Probably not unless you want to be double-sure that Kokkos is on :)

components/eamxx/src/physics/rrtmgp/rrtmgp_test_utils.hpp
jgfouca: Yes, this pattern works best for Kokkos. Note that this is code specific to Kokkos only, so YAKL code is not impacted. Both Kokkos and YAKL should currently be using the patterns that are best for them.

bartgol (approved):

AaronDonahue:
@jgfouca , is this ready for review?

jgfouca:
@AaronDonahue , yes. The current data:

Screenshot 2025-01-24 at 11 36 24 AM

So, basically exact parity with YAKL. @ndkeen is checking a few things for me on pm.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.hpp
```diff
@@ -1011,23 +1161,35 @@ static void get_subcolumn_mask(const int ncol, const int nlay, const int ngpt, c
...
+    // }));
+    // Kokkos::parallel_for(ncol, KOKKOS_LAMBDA(int icol) {
+    //     conv::Random rand(seeds(icol));
+    //     for (int igpt = 0; igpt < ngpt; igpt++) {
+    //       for (int ilay = 0; ilay < nlay; ilay++) {
+    //         cldx(icol,ilay,igpt) = rand.genFP<RealT>();
+    //       }
+    //     }
+    //   });
+    TIMED_KERNEL(FLATTEN_MD_KERNEL3(ncol, nlay, ngpt, icol, ilay, igpt,
```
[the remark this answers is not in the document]
jgfouca: Yes, the kernel here is the faster one. The issue is that this makes the PR non-BFB, so I left the old code commented out so that I could confirm the rest of the PR is BFB (with debug CPU).

jgfouca:
@ndkeen says initial perlm testing shows we are now at parity with YAKL there as well. I think we are very close to being able to merge this.

bartgol (approved):

rljacob:
notes:  waiting for dashboard to clear.

ndkeen:
We still want to be able to switch between yakl/kokkos radiation correct?

jgfouca:
Yes, the YAKL cmake stuff is only turned off if SCREAM_RRTMPG_ENABLE_YAKL is Off.

ambrad:
@bartgol note we want to rebase #6916 on master when this PR goes in.

rljacob:
@jgfouca is this ready?

jgfouca:
@rljacob , yes. I think we are finally at the point where all the CI fails are just expected DIFFs. I will begin integrating.

bartgol:
@jgfouca the baseline fails are likely expected, right? That said, I wonder why we don't get ALL tests running rrtmgp to diff. E.g., the following pass:

- rrtmgp_standalone_chunked_baseline_cmp
- homme_shoc_cld_p3_rrtmgp_baseline_cmp
- homme_shoc_cld_p3_rrtmgp_pg2_baseline_cmp
- homme_shoc_cld_spa_p3_rrtmgp_baseline_cmp
- homme_shoc_cld_spa_p3_rrtmgp_pg2_dp_baseline_cmp

Is there any obvious reason why some cmp tests fail and some pass?

jgfouca:
Yes, those are expected. I do not know why the other rrtmgp tests pass.

PR #6889, opened 2025-01-10
EAMxx: add vert_contraction field utility

mahf708:
adds a vertical contraction utility equivalent to einsum('k,...k->...', weight, field) or einsum('ik,i...k->i...', weight, field)

---

Following #6776 but for the vertical, except with some added complexity and more caveats. Bare-bones impl for now. Design notes regarding two broad cases supported:
- if the weight provided is 1D, its lone dim must be LEV; this will be used to calculate vertical sums or means whose weight is column-invariant, say for aodvis (see diagnostic impl).
- if the weight provided is 2D, its dims must be COL,LEV; this will be used to calculate vertical sums or means whose weight is column-variant, say waterpath (see diagnostic waterpath)

### DIFF
--- a/components/eamxx/src/share/field/field_utils.hpp
+++ b/components/eamxx/src/share/field/field_utils.hpp
@@ -179,6 +179,99 @@ void horiz_contraction(const Field &f_out, const Field &f_in,
   impl::horiz_contraction<ST>(f_out, f_in, weight, comm);
 }
 
+// Utility to compute the contraction of a field along its level dimension.
+// This is equivalent to f_out = einsum('...k->...', weight, f_in).
+// The impl is such that:
+// - f_out, f_in, and weight must be provided and allocated
+// - The last dimension is for the levels (LEV)
+// - There can be only up to 3 dimensions of f_in
+// - Weight is assumed to be (in order of checking/impl):
+//   - rank-1, with only LEV dimension
+//   - rank-2, with only COL and LEV dimensions
+template <typename ST>
+void vert_contraction(const Field &f_out, const Field &f_in,
+                      const Field &weight, const ekat::Comm *comm = nullptr) {
+  using namespace ShortFieldTagsNames;
+
+  const auto &l_out = f_out.get_header().get_identifier().get_layout();
+
+  const auto &l_in = f_in.get_header().get_identifier().get_layout();
+
+  const auto &l_w = weight.get_header().get_identifier().get_layout();
+
+  // Sanity checks before handing off to the implementation
+  EKAT_REQUIRE_MSG(
+      l_w.rank() >= 1 && l_w.rank() <= 2,
+      "Error! The weight field must be at least rank-1 and at most rank-2.\n"
+      "The weight field has rank "
+          << l_w.rank() << ".\n");
+  EKAT_REQUIRE_MSG(
+      l_w.tags().back() == LEV,
+      "Error! The weight field must have LEV as its last dimension.\n"
+      "The weight field layout is "
+          << l_w.to_string() << ".\n");
+  EKAT_REQUIRE_MSG(l_in.rank() <= 3,
+                   "Error! The input field must be at most rank-3.\n"
+                   "The input field rank is "
+                       << l_in.rank() << ".\n");
+  EKAT_REQUIRE_MSG(l_in.rank() >= l_w.rank(),
+                   "Error! The input field must have at least as many "
+                   "dimensions as the weight field.\n"
+                   "The input field rank is "
+                       << l_in.rank() << " and the weight field rank is "
+                       << l_w.rank() << ".\n");
+  EKAT_REQUIRE_MSG(l_in.tags().back() == LEV,
+                   "Error! The input field must have a level dimension.\n"
+                   "The input field layout is "
+                       << l_in.to_string() << ".\n");
+  EKAT_REQUIRE_MSG(
+      l_in.dim(l_in.rank() - 1) == l_w.dim(l_w.rank() - 1),
+      "Error! input and weight fields must have the same dimension along "
+      "which we are taking the reducing the field (last dimensions).\n"
+      "The weight field has last dimension "
+          << l_w.dim(l_w.rank() - 1)
+          << " while "
+             "the input field has last dimension "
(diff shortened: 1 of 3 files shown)


components/eamxx/src/share/field/field_utils.hpp
```diff
@@ -179,6 +179,99 @@ void horiz_contraction(const Field &f_out, const Field &f_in,
...
+
+  const auto &l_out = f_out.get_header().get_identifier().get_layout();
+
+  const auto &l_in = f_in.get_header().get_identifier().get_layout();
+
+  const auto &l_w = weight.get_header().get_identifier().get_layout();
+
+  // Sanity checks before handing off to the implementation
+  EKAT_REQUIRE_MSG(
+      l_w.rank() >= 1 && l_w.rank() <= 2,
```
bartgol, line 204: Probably subjective, but since there are only 2 options, we may do a more explicit check like `l_w.rank()==1 or l_w.rank()==2`, which has the same code length anyways. Inequalities take a bit longer to visually parse, and it may take some time for the reader to figure out that 1 and 2 are the only valid values.

components/eamxx/src/share/field/field_utils.hpp
```diff
@@ -179,6 +179,99 @@ void horiz_contraction(const Field &f_out, const Field &f_in,
...
+        "Error! input and weight fields must have the same dimension along "
+        "which we are taking the reducing the field (first dimensions).\n"
+        "The weight field has first dimension "
+            << l_w.dim(0)
+            << " while "
+               "the input field has first dimension "
+            << l_in.dim(0) << ".\n");
+  }
+  EKAT_REQUIRE_MSG(
+      l_out == l_in.clone().strip_dim(l_in.rank() - 1),
```
bartgol, line 256: You can also call `strip_dim(LEV)`, which may be more verbose.

components/eamxx/src/share/field/field_utils.hpp
```diff
@@ -179,6 +179,99 @@ void horiz_contraction(const Field &f_out, const Field &f_in,
...
+             "the input field has last dimension "
+          << l_in.dim(l_in.rank() - 1) << ".\n");
+  EKAT_REQUIRE_MSG(
+      l_in.dim(l_in.rank() - 1) > 0,
+      "Error! The input field must have a non-zero level dimension.\n"
+      "The input field layout is "
+          << l_in.to_string() << ".\n");
+  if(l_w.rank() == 2) {
+    EKAT_REQUIRE_MSG(
+        l_w.tags()[0] == COL && l_in.tags()[0] == COL,
```
bartgol, line 243: I'm assuming the only possible l_in types are `<COL,LEV>`, and `<COL,CMP,LEV>` (possibly with CMP repeated). If so, you could maybe consolidate this check and the next via something like
```c++
EKATE_REQUIRE_MSG (l_w.congruent(l_in.clone().strip_dim(CMP,false)),
  "Error! Incompatible layouts\n"
  "  field in: " + l_in.to_stirng() + "\n"
  "  weight: " + l_w.to_stirng() + "\n");
```
The `false` arg instruct to not throw if the dim is not found.

components/eamxx/src/share/field/field_utils_impl.hpp
```diff
@@ -366,6 +366,105 @@ void horiz_contraction(const Field &f_out, const Field &f_in,
...
+          f_out.name(), RangePolicy(0, nlevs),
+          KOKKOS_LAMBDA(const int i, ST &ls) { ls += v_w(i) * v_in(i); },
+          v_out);
+    } break;
+    case 2: {
+      auto v_in    = f_in.get_view<const ST **>();
+      auto v_out   = f_out.get_view<ST *>();
+      const int d0 = l_in.dim(0);
+      auto p       = ESU::get_default_team_policy(d0, nlevs);
+      if(l_w.rank() == 1) {
```
bartgol, line 399: One option, feel free to dismiss it. Instead of duplicating the whole code, you could create BOTH 1d and 2d views, and do the if inside the loop (the branch predictor will make the if cost negligible). Of course, only one of the views contains anything. So something like this:
```c++
typename Field::get_view_type<const ST*,Device> w1d;
typename Field::get_view_type<const ST**,Device> w2d;
auto one_d_w =  l_w.rank()==1;
if (one_d_w)
  w1d = weigth.get_view<const ST*>();
else
  w2d = weight.get_view<const ST**>();
```
then, inside the KOKKOS_LAMBDA, simply do
```c++
if (one_d_w)
  ac += w1d(j) * ...
else
  ac += w2d(i,j) * ...
```
It may lead to a tiny bit less code. In the end, what you did is ok. But if you dislike the code duplication, the above can be a viable alternative.

bartgol (commented):
Looks good. I have one doubt regarding GPU testing, plus a few suggestions that are up for grabs.

components/eamxx/src/share/field/field_utils_impl.hpp
bartgol: This whole thing is a) in a unit test, so who cares, and b) there is no allocation happening (`get_view` just creates a view from another view).

The ternary op does not work b/c the data type is different (not the mem traits), but yeah, the 2 return values must have the same type. We could think about using Kokkos::DynRankView, if that is appealing...

components/eamxx/src/share/field/field_utils.hpp
bartgol: Yeah, that works. You can also do `strip_dims({LEV,ILEV})`. This will strip any of the dimensions in the input list, without throwing if they are not found.

components/eamxx/src/share/field/field_utils_impl.hpp
bartgol: Ah, sorry, I missed the subview part. You could try to wrap the first into an ekat::Unmanaged:
```c++
auto weight_view_icol = one_d_w ?
                        ekat::Unmanaged(decltype(weigth.get_view<const ST*>())(weight.get_view<const ST*>()) : 
                        ekat:subview(weigth.get_view<const ST**>(), i);
```
a bit cumbersome...

bartgol (approved):

tcclevenger (approved):
LGTM

PR #6890, opened 2025-01-10
Fix valgrind error in atm proc DAG creation

bartgol:
The issue was simply that we were grabbing a ref to an element in a vector that was later resized, invalidating the ref.


mjs271 (approved):
good catch--thank you!

PR #6882, opened 2025-01-09
EAMxx: temporarily disable non-deterministic and/or buggy mam test

bartgol:
Allows other PRs to test normally while the MAM team investigates further

---

PR #6894, opened 2025-01-13
Update mam4xx to include fix for wet scavenging

mjschmidt271:
This brings in changes to mam4xx's `wet_dep.hpp` that correct the size of a work array to fix out-of-bounds errors that have led to recent fails for the `mam4_wetscav_standalone` test.

**Note:** This PR points to a branch in mam4xx that has yet to be merged into `main`--I will force push when that associated PR is approved/merged.


mjschmidt271:
@bartgol @mahf708 @AaronDonahue @jgfouca 

Can't add reviewers, so please take a look when you have a chance!

jgfouca (approved):

singhbalwinder:
Thanks, @mjschmidt271. Can you please also uncomment the wet scavenging test in the PR?

mjschmidt271:
> Thanks, @mjschmidt271. Can you please also uncomment the wet scavenging test in the PR? it was commented out in #6882 PR

Ah, good call--will do!

bartgol:
Did you verify that with this mods we are now deterministic on GPU? If so, you could re-enable all mam4 tests that were previously disabled. The checks will fail since we don't have the up-to-date baselines, but at least we can then gen baselines right after merging this. Also, you should add some cmake code for the mam4 tests that run the np1_vs_npN tests, so we can already verify non-determinism there.

components/eamxx/tests/single-process/CMakeLists.txt
```diff
@@ -24,9 +24,7 @@ if (SCREAM_ENABLE_MAM)
   add_subdirectory(mam/drydep)
   add_subdirectory(mam/emissions)
   add_subdirectory(mam/constituent_fluxes)
-
-  # Currently, this test has a runtime error. Disabling while MAM folks debug.
-  # add_subdirectory(mam/wet_scav)
+  add_subdirectory(mam/wet_scav)
```
bartgol, line 27: There are also currently disabled mam tests in the multiple-process subfolders.

mjschmidt271:
@bartgol - I don't believe this will affect the non-determinism in the `mam4_aero_microphys_standalone` test, but is restricted to the wet scavenging issue you'd previously mentioned with out-of-bounds access. As such, `mam4_wetscav_standalone ` is re-enabled but not `mam4_aero_microphys_standalone`.

I know James is working on the microphysics non-det, though, so I'd guess he'll knock it out before I even have a chance to take a look again 🙂

And I will put those extra test cases next on my list--thanks for that info!

bartgol:
> @bartgol - I don't believe this will affect the non-determinism in the `mam4_aero_microphys_standalone` test, but is restricted to the wet scavenging issue you'd previously mentioned with out-of-bounds access. As such, `mam4_wetscav_standalone ` is re-enabled but not `mam4_aero_microphys_standalone`.
> 
> I know James is working on the microphysics non-det, though, so I'd guess he'll knock it out before I even have a chance to take a look again 🙂
> 
> And I will put those extra test cases next on my list--thanks for that info!

Ah I see. Thanks!

bartgol (approved):

bartgol:
@mjs271 I am a bit concerned by the fail of the wetscav tests on CUDA in the release build. The cmp tests passed in debug, which may point to some non-deterministic stuff happening in release mode. I re-triggered the job, and we will be able to verify if it's indeed non-det by checking the cprnc output.

bartgol:
The two consecutive runs do produce the same result. Perhaps the release baselines (which are old) were generated with the non-deterministic version, I'm not sure. I will assume it's working correctly, and I will leave it to the mam4xx team to verify they can trust the test.

PR #6895, opened 2025-01-14
EAMxx: set testing max ranks to 2 on ghci-snl-cuda

bartgol:
Back when we tested on Weaver, we ran 3 builds in parallel on compute nodes (which had 4 GPUs). So on GPU platforms, we just set all builds with max test ranks to 1, to avoid having to deal with which build had 1 gpu and which had 2 gpus.

Now, the gh workflow we use for testing runs a _single_ build on each blake node, which have 2 GPUs. Hence, we can finally run multi-rank tests with our GPU standalone tests. This should add tests like `np1_vs_np2` to our AD system tests, which can help verifying non-deterministic behavior (a simple DIFF against baselines is not clear enough: are baselines simply not blessed or is the PR causing DIFFs?)

### DIFF
--- a/components/eamxx/cmake/machine-files/ghci-snl-cuda.cmake
+++ b/components/eamxx/cmake/machine-files/ghci-snl-cuda.cmake
@@ -12,3 +12,6 @@ set(EKAT_MPI_NP_FLAG "-n" CACHE STRING "The mpirun flag for designating the tota
 
 # TODO: rebuild cuda image with cuda-aware MPI, so we can set this to ON
 option(SCREAM_MPI_ON_DEVICE "Whether to use device pointers for MPI calls" OFF)
+
+# Currently, we have 2 GPUs/node on Blake, and we run a SINGLE build per node, so we can fit 2 ranks there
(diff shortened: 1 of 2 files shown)


jgfouca (approved):

bartgol:
The baseline_cmp fails are expected, since we now cmp the np2 tests outputs (while before we cmp the np1 tests output). But the io_se_grid_np2 fail is unexpected. I'm investigating.

bartgol:
The io_se_grid unit test was indeed buggy, trying to do MPI operations from a device pointer. We never caught it cause we only ran with 1 ranks, in which case MPI is probably able to recognize that no comm is needed, and the src/dst pointers are not even accessed.

bartgol:
I just kicked of the tests to make sure they pass now that baselines have been regenerated (the FIOP-SCREAMv1 test had corrupted baselines). It should be orthogonal to this PR, but I'm letting it finish anyways to double check baselines integrity, then I'll merge.

jgfouca (approved):

PR #6897, opened 2025-01-15
EAMxx: improve horiz/vert contraction tests

### LINKED ISSUE #6893: improve test robustness comparing regarding order-of-ops
as we discussed privately, the tests fail with 4 levels ... this needs to be fixed. Let me document it in an issue and fix instances of this stuff in tests (added by me) throughout the code.

_Originally posted by @mahf708 in https://github.com/E3SM-Project/E3SM/pull/6889#discussion_r1913700455_

instances that I am aware of right now:
- vert_contraction tests
- horiz_contraction tests 

Fix incoming soon
            

xref:
- https://github.com/E3SM-Project/E3SM/pull/6889#discussion_r1911502258
- https://kokkosteam.slack.com/archives/C5BGU5NDQ/p1736790859950999

mahf708:
imrpoves contraction testing by using catch2 matchers to allow for tolerance and expand contracted dimensions.

Fixes #6893

### DIFF
--- a/components/eamxx/src/share/tests/field_utils.cpp
+++ b/components/eamxx/src/share/tests/field_utils.cpp
@@ -127,13 +127,16 @@ TEST_CASE("utils") {
   }
 
   SECTION("horiz_contraction") {
+    // A numerical tolerance
+    auto tol = std::numeric_limits<Real>::epsilon() * 100;
+
     using RPDF  = std::uniform_real_distribution<Real>;
     auto engine = setup_random_test();
     RPDF pdf(0, 1);
 
-    int dim0 = 3;
-    int dim1 = 9;
-    int dim2 = 2;
+    int dim0 = 9;
+    int dim1 = 4;
+    int dim2 = 7;
 
     // Set a weight field
     FieldIdentifier f00("f", {{COL}, {dim0}}, m / s, "g");
@@ -142,7 +145,9 @@ TEST_CASE("utils") {
     field00.sync_to_host();
     auto v00 = field00.get_strided_view<Real *, Host>();
     for(int i = 0; i < dim0; ++i) {
-      v00(i) = (i + 1) / sp(6);
+      // By design, denominator is the sum of the first dim0 integers
+      // (analytically known)
+      v00(i) = sp(i + 1) / sp(dim0 * (dim0 + 1) / 2);
     }
     field00.sync_to_dev();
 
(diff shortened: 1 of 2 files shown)


components/eamxx/src/share/tests/field_utils.cpp
```diff
@@ -127,13 +127,16 @@ TEST_CASE("utils") {
...
     using RPDF  = std::uniform_real_distribution<Real>;
     auto engine = setup_random_test();
     RPDF pdf(0, 1);
 
-    int dim0 = 3;
-    int dim1 = 9;
-    int dim2 = 2;
+    int dim0 = 9;
+    int dim1 = 4;
+    int dim2 = 7;
```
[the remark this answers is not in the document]
bartgol: Btw, I noticed that vert_contraction does not take advantage of packs on CPU. Perhaps that's another thing we can improve later.

components/eamxx/src/share/tests/field_utils.cpp
```diff
@@ -195,7 +200,12 @@ TEST_CASE("utils") {
     horiz_contraction<Real>(result, field00, field00);
     result.sync_to_host();
     auto v = result.get_view<Real, Host>();
-    REQUIRE(v() == (1 / sp(36) + 4 / sp(36) + 9 / sp(36)));
+    // The numerator is the sum of the squares of the first dim0 integers
```
bartgol, line 203: Maybe you could make these into vars, to be a bit more self-explanatory? Like
```c++
Real sum_n = dim0 * (dim0+1) / 2;
Real sum_n_sq = dim0 * (dim0+1) * (2*dim0+1) / 6;
```
and then use those throughout the rest of the test, since they appear a few times.

bartgol: Since you use them also in the vert_contraction, maybe turn them into small inline fcns
```c++
inline int sum_n (int n) { return n*(n+1)/2; }
inline int sum_n_sq (int n) { return n*(n+1)*(2*n+1)/6; }
```

components/eamxx/src/share/tests/field_utils.cpp
```diff
@@ -127,13 +127,16 @@ TEST_CASE("utils") {
   }
 
   SECTION("horiz_contraction") {
+    // A numerical tolerance
```
bartgol, line 130: For the sake of who will maintain this file in the future, you could add a line explaining why we need tol-based tests. Something like
```
// Accumulations in the Kokkos threaded reductions may be done in a different order than
// the manual ones below, so we can only test correctness up to a tolerance
```

bartgol (approved):
Looks good. I have a couple of comments, but nothing too important.

components/eamxx/src/share/tests/field_utils.cpp
bartgol: That's b/c you are inside a fcn already. You can't declare a function inside a function, but you can declare a lambda.

PR #6901, opened 2025-01-15
Begin to de-scream things (use eamxx instead)

jgfouca:
This PR changes the test suite names and unifies the testmods under eamxx.

We'll have to remember to update E3SM_test_scripts once this is in. The dashboard will also need adjustments.

[BFB]

edit: I forgot to mention that this PR also adds a couple eamxx v1 tests to e3sm_developer including a prod test.


AaronDonahue (approved):
Nice! Thanks Jim

bartgol (approved):

cime_config/tests.py
```diff
@@ -653,17 +655,17 @@
...
 
-    "e3sm_scream_v0" : {
+    "e3sm_eamxx_v0" : {
         "time"    : "03:00:00",
-        "inherit" : ("e3sm_scream_v0_lowres"),
+        "inherit" : ("e3sm_eamxx_v0_lowres"),
     },
 
-    "e3sm_scream_v0_lowres" : {
+    "e3sm_eamxx_v0_lowres" : {
```
rljacob, line 668: want to just delete these v0 ones or save that for a later PR?

jgfouca:
DIFFs are expected because many of the case names changed due to the namelist dir rename.

jgfouca:
I have fixed the one failing cime-nml unit test.

rljacob:
Can we drop the "v1' from the test suite name?   I think that was supposed to indicate c++ but "eamxx" now does that.

PR #6914, opened 2025-01-17
EAMxx: add IOPRemapper and use it in SPA and IOPDataManager

bartgol:
Implement an IOP remapper, with the goal of simplifying downstream code, expressing it in terms of more generic interfaces.

---

@tcclevenger @AaronDonahue You guys should review only the last 3 commits of this branch, since the other ones are from the remappers-api-cleanup branch.

he main part of the PR is the IOPRemapper, but I wanted to verify its correctness by also using it somewhere. The easiest place was SPA, since it has minimal IOP-related stuff (only the single-col remap stuff). All in all, I think SPA gets simplified, and it gets to a good state to be later refactored in terms of a DataInterpolation structure.

I could also start using IOPRemapper in other places of the code (most notably inside the IOPDataManager class itself). Perhaps it's best to do things in steps though, and integrate one small change at a time? Or maybe @tcclevenger could do that step later, since he's more familiar with the IOP class and its assumptions?

Depends on #6908

### DIFF
--- a/components/eamxx/src/share/atm_process/IOPDataManager.cpp
+++ b/components/eamxx/src/share/atm_process/IOPDataManager.cpp
@@ -1,4 +1,5 @@
 #include "share/grid/point_grid.hpp"
+#include "share/grid/remap/iop_remapper.hpp"
 #include "share/io/scorpio_input.hpp"
 #include "share/io/scream_scorpio_interface.hpp"
 #include "share/atm_process/IOPDataManager.hpp"
@@ -7,22 +8,6 @@
 #include "ekat/util/ekat_lin_interp.hpp"
 
 #include <numeric>
-
-// Extend ekat mpi type for <Real,int> pairs,
-// used for reduction of type MPI_MINLOC.
-namespace ekat {
-#ifdef SCREAM_DOUBLE_PRECISION
-  template<>
-  MPI_Datatype get_mpi_type<std::pair<double, int>> () {
-    return MPI_DOUBLE_INT;
-  }
-#else
-  template<>
-  MPI_Datatype get_mpi_type<std::pair<float, int>> () {
-    return MPI_FLOAT_INT;
-  }
-#endif
-}
 
 namespace scream {
 namespace control {
@@ -99,7 +84,7 @@ initialize_iop_file(const util::TimeStamp& run_t0,
   //                  First entry will be the variable name used when accessing in class
   //   - fl:          IOP field layout (acceptable ranks: 0, 1)
   //   - srf_varname: Name of surface variable potentially in iop file associated with iop variable.
-  auto setup_iop_field = [&, this] (const vos&         varnames,
+  auto setup_iop_field = [&, this] (const std::vector<std::string>& varnames,
                                     const FieldLayout& fl,
                                     const std::string& srf_varname = "none") {
     EKAT_REQUIRE_MSG(fl.rank() == 0 || fl.rank() == 1,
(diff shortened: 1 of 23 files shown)


components/eamxx/src/share/util/scream_setup_random_test.hpp
```diff
@@ -18,7 +18,7 @@ inline int get_random_test_seed(const ekat::Comm* comm=nullptr)
 
   std::random_device rdev;
   const int catchRngSeed = Catch::rngSeed();
-  int seed = catchRngSeed==0 ? rdev() : catchRngSeed;
+  int seed = catchRngSeed==0 ? rdev()/2 : catchRngSeed;
```
bartgol, line 21: This change is b/c you can't do `./my_test --rng-seed -123`, since the cmd line arg parser interprets `-123` as a flag, not as the value for `--rng-seed`. Since `rdev()` is an unsigned int, assigning it to an int can overflow. But if we divide by 2, we are SURE it will be <=2^31-1, which is in the range of int, guaranteeing that seed>=0.

tcclevenger:
does this need a rebase now that https://github.com/E3SM-Project/E3SM/pull/6908 is merged?

bartgol:
The CI runs keep getting stuck, and I have to kill them. There's probably a test that hangs. My guess is `io_scm_reader`. If I can confirm that, I may just remove that class (or at least not compile the src and run the tests), since it was meant for ease of use of pyscream, but I don't think anyone is using pyscream anyways...

bartgol:
@AaronDonahue @tcclevenger the tests finally pass and don't hange (I had introduced a tiny bug in SCMInput). You can review whenever you want.

bartgol:
> Am I correct in understanding that now SPA uses IOP functions for remapping? So this would make IOP a set of general use functions that could also be used for full model simulations?

Uhm, no. SPA uses IOPRemapper, which does not really know anything about IOP, or IOPDataManager. It only knows about a target lat/lon coordinates. It remaps (ncols_src,nlevs) to (ncols_tgt,nlevs), where the target fields have the same value for all columns, and those value match the column in the source field that is the closest to that lat/lon (across all ranks).

As such, SPA does not know anything about IOP. It only knows that if IOPDataManager is set (it is set in all atm procs by the driver), it has to create an IOPRemapper, using the lat/lon stored inside it.

components/eamxx/src/share/atm_process/IOPDataManager.cpp
```diff
@@ -321,184 +306,62 @@ setup_io_info(const std::string& file_name,
...
+                               const std::shared_ptr<const AbstractGrid>& grid)
 {
-  const auto dummy_units = ekat::units::Units::nondimensional();
-
-  EKAT_REQUIRE_MSG(field_names_nc.size()==field_names_eamxx.size(),
-                  "Error! Field name arrays must have same size.\n");
-
-  if (field_names_nc.size()==0) {
+  if (fields.size()==0) {
+    // Does this ever happen?
```
tcclevenger, line 327: Probably not? We could remove this check make it error out because it doesn't make sense to call.

tcclevenger (approved):
This looks great. I commented on the early return in the case of 0 fields, but feel free to ignore since it's not hurting anything.

components/eamxx/src/share/atm_process/IOPDataManager.cpp
bartgol: I can replace it with an assert in the next PR.

PR #6916, opened 2025-01-19
EAMxx: Kokkos 4.5 update, fixes and workarounds for Kokkos update, Aurora support

oksanaguba:
* Update externals/ekat/extern/kokkos to 4.5.
* Update EKAT itself to fix deprecated usage and add Aurora cmake files.
* Fix and work around various issues on Frontier due to the Kokkos update.
* Fix deprecated usage in Hommexx.


oksanaguba:
this commit is running on sunspot 
```
~/ess/v1-jan19/run/atm.log.11365537.amn-0001.250119-204314 
```

oksanaguba:
i should have not committed to here turning off compose, will remove it once this PR is finalized.

ambrad:
Looks good. In addition to any usual pre-merge testing, I strongly recommend running e3sm_scream_v1_medres with this PR on Frontier, pm-gpu, and Chrysalis, as well as homme_integration on Chrysalis.

components/eamxx/CMakeLists.txt
```diff
@@ -106,14 +107,21 @@ if (Kokkos_ENABLE_CUDA)
...
   enable_language(HIP)
   set (EAMXX_ENABLE_GPU TRUE CACHE BOOL "" FORCE)
   set (HIP_BUILD TRUE CACHE BOOL "" FORCE) #needed for yakl if kokkos vars are not visible there?
 endif ()
 
+# Determine if this is a sycl build.
+if (Kokkos_ENABLE_SYCL)
+  #enable_language(SYCL)
+  set (EAMXX_ENABLE_GPU TRUE CACHE BOOL "" FORCE)
+  set (SYCL_BUILD TRUE CACHE BOOL "" FORCE) #needed for yakl if kokkos vars are not visible there?
```
ambrad, line 122: You could turn off COMPOSE here, I believe. (I'll fix the COMPOSE issues once I'm on the machine.)

components/eamxx/src/dynamics/homme/CMakeLists.txt
```diff
@@ -23,7 +23,8 @@ set(BUILD_HOMME_PREQX_KOKKOS OFF CACHE BOOL "")
 set(BUILD_HOMME_PESE         OFF CACHE BOOL "")
 set(BUILD_HOMME_SWIM         OFF CACHE BOOL "")
 set(BUILD_HOMME_PRIM         OFF CACHE BOOL "")
-set(HOMME_ENABLE_COMPOSE     ON  CACHE BOOL "")
+#set(HOMME_ENABLE_COMPOSE     ON  CACHE BOOL "")
+set(HOMME_ENABLE_COMPOSE     OFF  CACHE BOOL "")
```
ambrad, line 27: Just noting here this is what needs to be removed from this PR.

components/eamxx/src/physics/rrtmgp/CMakeLists.txt
```diff
@@ -168,12 +177,17 @@ yakl_process_target(scream_rrtmgp_yakl)
 # NOTE: cannot use 'PUBLIC' in target_link_libraries,
 #       since yakl_process_target already used it
 #       with the "plain" signature
+
+#<<<<<<< HEAD
+#find_library(NETCDF_C netcdf HINTS $ENV{NETCDF_C_PATH}/lib)
+#target_link_libraries(scream_rrtmgp_yakl ${NETCDF_C} rrtmgp scream_share)
```
ambrad, line 183: Can you explain this change? The second line I'm guessing is due to Kokkos, but what about the NETCDF_C line?

oksanaguba: i need to ask Jim about this -- iirc i had to set io libs in general, but also some cmake variables for io for rrtmgp separately. so i left this file messy, because i still need to sort it out.

oksanaguba:
i id not understant EKAT comment -- i have an ekat branch, oksanaguba/spot, which i maintained only to point to a different kokkos. so i periodically merge master into it, but not too frequently, so it may be behind the ekat used in master, but its kokkos points not to e3sm kokkos, but to kokkos develop branch. 

this is another issue to sort out (but i would not know how).

ambrad:
> i id not understant EKAT comment 

Re: EKAT, it appears there are missing commits on your branch w.r.t. the submodule point in E3SM: https://github.com/E3SM-Project/EKAT/compare/oksanaguba/spot...42313831a530730875e1ada147a9375a3b100129

@bartgol am I seeing that correctly? Also, is current EKAT master ready to be used in E3SM?

If so, Oksana, I recommend rebasing your ekat branch on master, then pointing the ekat submodule to that cleaned-up branch.

More generally, testing across machines will help to flush out issues, if there are any, including any with ekat.

oksanaguba:
there are some December commits in the diff you posted, yes, at least those are expected. my ekat branch was not updated too frequently. will do what you suggested, thanks.

ambrad:
As an example, here is one commit I'm either confused about or is truly missing from your branch: https://github.com/E3SM-Project/EKAT/commit/5804bfa412787259c0d8468ff0f7c78d9bb1c473. This commit was merged in late Nov 2024, which is why I'm thinking there's a fair bit missing.

cime_config/machines/cmake_macros/oneapi-ifxgpu.cmake
```diff
@@ -4,15 +4,17 @@ if (compile_threaded)
   string(APPEND CMAKE_CXX_FLAGS " -qopenmp")
   string(APPEND CMAKE_EXE_LINKER_FLAGS  " -qopenmp")
 endif()
-string(APPEND CMAKE_C_FLAGS_RELEASE   " -O2")
-string(APPEND CMAKE_Fortran_FLAGS_RELEASE   " -O2")
-string(APPEND CMAKE_CXX_FLAGS_RELEASE " -O2")
-string(APPEND CMAKE_Fortran_FLAGS_DEBUG   " -O0 -g -fpe0")
+
+#adding -g here leads to linker internal errors
```
bartgol, line 8: And yet I see `-g` below. Is this comment outdated, and link is fine now? If so, we should remove it to avoid confusion.

components/eamxx/src/dynamics/homme/physics_dynamics_remapper.cpp
```diff
@@ -426,6 +426,10 @@ do_remap_fwd()
   const int team_size = std::min(256, std::min(128*m_num_phys_cols,32*(concurrency/this->m_num_fields+31)/32));
 #endif
 
+#ifdef KOKKOS_ENABLE_SYCL
+  const int team_size = 4;
```
bartgol, line 430: I am puzzled by this. Do we really want just 4 on sycl? It's staggeringly different from other GPU platforms' default...

components/eamxx/src/physics/rrtmgp/CMakeLists.txt
```diff
@@ -168,12 +177,17 @@ yakl_process_target(scream_rrtmgp_yakl)
...
+#=======
 find_library(NETCDF_C netcdf HINTS ${NetCDF_C_PATH}/lib)
```
bartgol, line 185: While at it, do you mind adding `${NetCDF_C_PATH}/lib64` to the hints? Since `lib64` is quite common, we should be proactive and add it to the list of hints (e..g, my laptop and workstation use lib64)

bartgol (approved):
Looks good to me. I have a couple of questions/remarks, but nothing too important. I'm going to approve, modulo the fix of EKAT submodule that you are already working on.

cime_config/machines/cmake_macros/oneapi-ifxgpu.cmake
oksanaguba: "just -g" is in debug flags only, and i haven't tried to build debug. so i only changed release flags.

i was able to compile and to run with "just -g" on sunspot, however, the folder size for e3sm build gets from say 4 gb to 140 gb with full -g. last i tried "just -g" on aurora, there were linking issues. so this is all WIP. i'd say keep this as is now and sort it later.

components/eamxx/src/dynamics/homme/physics_dynamics_remapper.cpp
oksanaguba: the change is from Nov 2023. i looked at my notes and did not see why it was needed. i will try to remove it.

ambrad:
Looks like you need to merge ekat master into your branch one more time to get the macro name fix.

cime_config/machines/cmake_macros/oneapi-ifxgpu.cmake
bartgol: What I meant is that the comment is confusing: it makes it sound like we should not use `-g` b/c of link errors, but we _do_ use it across the board. Maybe the comment needs to be clarified. Something like "WARNING: the -g flag seems to have a wild behavior sometimes, with side effects ranging from VERY large binaries to link errors. If you experience errors, remove it".

components/eamxx/src/dynamics/homme/physics_dynamics_remapper.cpp
bartgol: This is for the P-D remapper (not the GllFvPhys one), so its relevance is somewhat limited (doesn't run at runtime, usually). So not a huge deal. But if things work without, no point in limiting team size to 4.

oksanaguba: used hip TS calculation and it worked at least for ne4 run.

components/eamxx/src/physics/rrtmgp/CMakeLists.txt
oksanaguba: thanks, done

ambrad (approved):
Looks great.

Note that SCREAM nightlies run on master, not next, so I always run e3sm_scream_v1_medres on Frontier, Chrysalis, and pm-gpu before merging a complex PR to avoid having to fix master later. I recommend it to others, as well.

ambrad:
CI is failing with a checkout on the ekat submodule.
```
  Error: fatal: remote error: upload-pack: not our ref 1eae1ad5a01cdb510726624ac647e7f96dd883d0
  Error: fatal: Fetched in submodule path 'externals/ekat', but it did not contain 1eae1ad5a01cdb510726624ac647e7f96dd883d0. Direct fetching of that commit failed.
```
Looking at the branch, https://github.com/E3SM-Project/EKAT/commits/oksanaguba/spot/, it's unclear where 1eae1a is coming from. @oksanaguba I suggest checking the submodule state in this PR w.r.t. to your ekat branch.

ambrad:
Re: commit 1eae1a, I don't see where it's coming from. I don't see it in your EKAT spot branch, but I could easily be mistaken.

ambrad:
Re: Kokkos: Ok, I'm guessing what's happening is you have your local submodule pointed to the Kokkos repo with this commit, but, as you wrote, the submodule metadata points to E3SM-Project/Kokkos. Generally, we have to put some extra commits on top of a Kokkos version, which is why E3SM-Project/Kokkos exists. Thus, I suggest you start by creating a branch in E3SM-Project/Kokkos with the desired Kokkos state and then use this in the submodule. That will let testing proceed.

oksanaguba:
i'll ask Conrad as he prob already has a branch

oksanaguba:
i pointed the branch to an e3sm kokkos branch, but haven't had a chance to see if CI is happier. also, will do more testing tomorrow.

oksanaguba:
chrysalis developer 
```
  ERIO.ne30_g16_rx1.A.chrysalis_intel (Overall: NLFAIL) details:
  ERS_D.ne4pg2_oQU480.F2010.chrysalis_intel.eam-hommexx (Overall: FAIL) details:
  ERS.f19_g16_rx1.A.chrysalis_intel (Overall: NLFAIL) details:
  ERS.ne30_g16_rx1.A.chrysalis_intel (Overall: NLFAIL) details:
  NCK.f19_g16_rx1.A.chrysalis_intel (Overall: NLFAIL) details:
  SEQ.f19_g16.X.chrysalis_intel (Overall: NLFAIL) details:
  SMS_D_Ln5.ne4pg2_oQU480.F2010.chrysalis_intel (Overall: FAIL) details:
  SMS_Lh4.ne4_ne4.F2010-SCREAMv1.chrysalis_intel.eamxx-output-preset-1 (Overall: FAIL) details:
  SMS_Lh4.ne4pg2_ne4pg2.F2010-SCREAMv1.chrysalis_intel.eamxx-output-preset-1--eamxx-prod (Overall: FAIL) details:

```

tcclevenger:
I'm manually testing on GPU and there are build errors. In 4.3 kokkos got rid of some `Kokkos::Impl` methods for Cuda that Hommexx was using. That is in addition to other errors that I'm seeing on Frontier (I think introduced in 4.3 as well). I'm currently trying to resolve these issues.

oksanaguba:
There are a lot of fails on chrysalis, but they all seem to be in the form of
```
/home/ac.onguba/ess/SMS_P12x2.ne4pg2_oQU480.WCYCL1850NS.chrysalis_intel.allactive-mach_mods.C.20250125_131133_jckk51/run/e3sm.log.672561.250125-144047:10: e3sm.exe           0000000002F54994  compose_mod_mp_co         418  compose_mod.F90
```
which points to compose finalize:     call slmm_finalize() . The code gets stuck there and runs out of time. I reverted the kokkos sha in my branch to 4.2 (the one that is used in master now), and these fails do not appear. there are 12 total fails with new kokkos. There are also diffs, but i will investigate those separately.

@ambrad would you want to look at this compose finalize issue? thnaks.

ambrad:
Yes, I'll troubleshoot the problem early this week.

tcclevenger:
For Kokkos 4.5.01, I'm having my runs on PM-GPU hang, and standalone tests hang on Weaver.

oksanaguba:
@tcclevenger  i maybe the same issue as i see with compose finalize, see if you can check in logs

ambrad:
I'm testing these changes and will report back: https://github.com/E3SM-Project/E3SM/compare/oksanaguba/eamxx/spot-clean...ambrad:E3SM:spot-clean-amb1

ambrad:
@oksanaguba I"m getting DIFFs on Chrysalis, but the e3sm_atm_developer and homme_integration runs are clean besides the diffs. Thus do you want me to push the following commits to this branch?
     https://github.com/E3SM-Project/E3SM/compare/oksanaguba/eamxx/spot-clean...ambrad:E3SM:spot-clean-amb1
Separately, I'll see if there's a way to avoid the baseline diffs.

bartgol:
@ambrad I'm assuming you are compiling with Kokkos::Serial as a backend, right? The CI fails when configuring eamxx are exposing an old bug in ekat (and eamxx) that shows up when openmp is on. Oksana pulled in the commit that fixes things in ekat, but we need a fix in eamxx.

@oksanaguba I can push the fix straight to your branch, if you want.

rljacob:
What is the purpose of introducing 4 new machines?  sunspot-pvc, sunspotcpu, sunspot-gen, and auroracpu ?

ambrad:
@bartgol do you have a branch with your fix in it? That said, I'm confident the fix I made to `compose` is needed: I see the change in behavior necessitating this fix was made in Kokkos 4.4.

bartgol:
> @bartgol do you have a branch with your fix in it? That said, I'm confident the fix I made to `compose` is needed: I see the change in behavior necessitating this fix was made in Kokkos 4.4.

I don't. The fix depends on certain stranded ekat commit that got cherry picked in the ekat-kokkos-4.5 branch, so I kind of need that ekat branch to include them. I didn't want to create multiple branches where ekat is updated, so I may just piggy-back a fix to this branch. Again, the fix only shows up when Kokkos_ENABLE_OPENMP=TRUE _AND_ a commit from ekat is present. This was a case of double-negative: two bugs canceling each other out by luck...

I haven't looked at your changes, so I can't comment. All I'm saying is that the fact that you were able to build makes me believe that Kokkos_ENABLE_OPENMP=FALSE in your build.

ambrad:
@bartgol one thing that might be confusing here is I'm fixing the F90 dycore used with EAM, not the C++ dycore used with EAMxx. So EKAT is irrelevant in my case.

oksanaguba:
@ambrad  yes, please push the fix, this branch is useless without 4.5.01 kokkos. regarding diffs -- if you gave a guess, it would be great. if not, i was going to revert a few changes to try to narrow it down, prob using standalone homme.

regarding your conversation above with Luca -- ii do not follow. but one test that uses hommexx dycore, ERS_D.ne4pg2_oQU480.F2010.chrysalis_intel.eam-hommexx , also freezes at the same spot in compose. i assume it is still not related to Luca's issue. i will try to re-read messages tomorrow to understand Luca's point better.

oksanaguba:
@rljacob  Re: "What is the purpose of introducing 4 new machines? sunspot-pvc, sunspotcpu, sunspot-gen, and auroracpu ?" it is much easier to debug setups early on when they are separated. this is related to the issue that Az raised, that this should be cleaned up eventually.

ambrad:
@oksanaguba I've pushed the commits. I'm somewhat sure that the diffs occur only in the SL code; I'm currently hunting for the exact cause.

components/eamxx/CMakeLists.txt
```diff
@@ -197,7 +205,12 @@ set(NetCDF_Fortran_PATH ${DEFAULT_NetCDF_Fortran_PATH} CACHE FILEPATH "Path to n
 set(NetCDF_C_PATH ${DEFAULT_NetCDF_C_PATH} CACHE FILEPATH "Path to netcdf C installation")
 set(SCREAM_MACHINE ${DEFAULT_SCREAM_MACHINE} CACHE STRING "The CIME/SCREAM name for the current machine")
 option(SCREAM_MPI_ON_DEVICE "Whether to use device pointers for MPI calls" ON)
+
 option(SCREAM_ENABLE_MAM "Whether to enable MAM aerosol support" ON)
+if (Kokkos_ENABLE_SYCL)
+  option(SCREAM_ENABLE_MAM "Whether to enable MAM aerosol support" OFF)
```
bartgol, line 211: I'm not sure this will work as you intend. You may have to simply do `set(SCREAM_ENABLE_MAM OFF)`

oksanaguba: thanks, but that one unlike SET worked.
```
//Whether to enable MAM aerosol support
SCREAM_ENABLE_MAM:BOOL=OFF
```
if you have a preference i will change it.

bartgol: I'm puzzled. This
```cmake
option(BLAH "" ON) 
message ("BLAH: ${BLAH}")
option(BLAH "" OFF)
message ("BLAH: ${BLAH}")
```
prints
```
BLAH: ON
BLAH: ON
```
Are you sure you did not have another line _before_ those that set SCREAM_ENABLE_MAM to OFF? Maybe in the mach file? Also, was it a clean build?

My preference is for using `set(... CACHE ..)` (or `option(...)`) only _once_.

bartgol: I found this in Professional CMake:
```
An important difference between normal and cache variables is that the set() command will only
overwrite a cache variable if the FORCE keyword is present, unlike normal variables where the set()
command will always overwrite a pre-existing value. The set() command acts more like set-if-not-
set when used to define cache variables, as does the option() command (which has no FORCE
capability). The main reason for this is that cache variables are primarily intended as a
customization point for developers.
```
which makes me think something else may have caused SCREAM_ENABLE_MAM=OFF in your cache file...

ambrad:
@bartgol did you say you have a fix for
```
CMake Error at /gpfs/fs1/home/ac.ambradl/e3smdev1/externals/ekat/cmake/EkatSetCompilerFlags.cmake:294 (message):
  [SetOmpFlags] You need to specify at least one language
```
This is on Chrysalis where I'm trying to build an EAMxx medres test. If you have a fix, feel free to push it to this PR branch.

ambrad:
On Frontier, it seems that bld/cmake-bld/eamxx/src/physics/rrtmgp/CMakeFiles/scream_rrtmgp.dir/flags.make gets a spurious `-x` that leads to
```
clang-15: error: language not recognized: '-std=gnu++17'
```
One sees it in flags.make here: `-x hip -fno-gpu-rdc -x --offload-arch=gfx900`. The second `-x` shouldn't be there. I haven't found what's injecting it.

tcclevenger:
> On Frontier, it seems that bld/cmake-bld/eamxx/src/physics/rrtmgp/CMakeFiles/scream_rrtmgp.dir/flags.make gets a spurious `-x` that leads to
> 
> ```
> clang-15: error: language not recognized: '-std=gnu++17'
> ```
> 
> One sees it in flags.make here: `-x hip -fno-gpu-rdc -x --offload-arch=gfx900`. The second `-x` shouldn't be there. I haven't found what's injecting it.

I was getting `language not recognized: '-Md'. Also coming from rrtmgp.

ambrad:
@tcclevenger I just updated my comment with another example. rrtmgp itself has nothing to do with it; it just comes first. As for your `-Md` case, check the build line for a spurious `-x`, possibly right before the `-Md`. As far as I can tell, the `language not recognized` message is meaningless and seems to get emitted due to bailing on a `-x` with no actual argument to `-x`.

components/eamxx/CMakeLists.txt
```diff
@@ -212,8 +225,8 @@ endif()
 #     #cmakedefine RRTMGP_EXPENSIVE_CHECKS
 option (SCREAM_RRTMGP_DEBUG "Turn on extra debug checks in RRTMGP" ${SCREAM_DEBUG})
 
-option(SCREAM_RRTMGP_ENABLE_YAKL "Use YAKL under rrtmgp" TRUE)
-option(SCREAM_RRTMGP_ENABLE_KOKKOS "Use Kokkos under rrtmgp" FALSE)
+option(SCREAM_RRTMGP_ENABLE_YAKL "Use YAKL under rrtmgp" FALSE)
```
ambrad, line 228: This might be an important change. I'm checking whether reverting this fixes Frontier. FYI, @tcclevenger @jgfouca

jgfouca: Yes, this must be the issue. I have a branch coming where Kokkos will be fully supported and have good performance.

ambrad: No luck. Still get the same issue. Will try something else.

ambrad: [Disable YAKL cmake stuff if we aren't using it](https://github.com/E3SM-Project/E3SM/pull/6879/commits/df5ae9082e332eda6b66f587bc0538df435bfcda) from  https://github.com/E3SM-Project/E3SM/pull/6879/commits looks promising.

ambrad: @jgfouca @tcclevenger the above commit does the trick. I got an ne30pg2 ERS test to run and pass with this and a commit (pushed to this branch) to fix a line of F90.

ambrad:
Once https://github.com/E3SM-Project/E3SM/pull/6879 is merged, let's rebase this branch on master so Frontier builds work.

ambrad:
On Frontier, with SCREAM_ENABLE_MAM=On and crayclang-scream, eamxx_mam_microphysics_process_interface.cpp fails to build with `error: unhandled SGPR spill to memory`.

bartgol:
@oksanaguba I pushed a fix on the ekat branch for the kokkos initialization setting error. When we are ready to merge this, I recommend we first issue a PR in ekat, to get stuff integrated in ekat/master, and then we update this PR, to point to an ekat/master version.

oksanaguba:
@ambrad thanks for all the fixes. is it correct that you already performed frontier testing? i can re-test on chrysalis before the merge.
@bartgol thanks, too; i see you pushed ekat fix into this branch, but i do not see a corresponding PR in master, i assume you would do it, or should i?

if everything else looks ok, i assume there are only 3 more items on the list: 1) wait for Jim's PR https://github.com/E3SM-Project/E3SM/pull/6879 before merging, 2) push fix to ekat master for kokkos devices 3) sort out machine entries for sycl machines . Maybe, after all of this 4) retest.

ambrad:
> is it correct that you already performed frontier testing?

I haven't run the full medres suite on Frontier yet but will once the final things are done. Now that PM is back up, we need to test there, as well.

tcclevenger:
> Edit: ah, I am disabling deprecated code 4 in kokkos, and `set_num_devices` is deprecated. I highly recommend we start adding `option (Kokkoe_ENABLE_DEPRECATED_CODE_4 "" OFF)` in all our mach files, and get rid of deprecated code now, in view of Kokkos 5.0 (which will likely remove these deprecations).

Completely agree about deprecated code off. Before Kokkos 5 the team is having a big push to deprecate a lot of unwanted stuff (to be removed in Kokkos5). Potential to be a large amount deprecated in the next 2 releases.

bartgol:
> @bartgol thanks, too; i see you pushed ekat fix into this branch, but i do not see a corresponding PR in master, i assume you would do it, or should i?

I did not do a PR in master, since there are more ekat changes in your branches, and I think it's easier to just do a single PR with all of them, when we're ready. As soon as you know this branch works, the cleanest thing to do is this:

1. do a PR in ekat, which will add all the fixes, as well as kokkos 4.5
2. update ekat in this branch to point to the new ekat/master
3. merge to next etc.

ambrad:
This branch in its current state fails screamv1 medres tests on Chrysalis with this seg fault:
```
  4:  9 0x0000000003a72826 scream::rrtmgp::rrtmgp_interface<double, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::OpenMP, Kokkos::HostSpace> >::rrtmgp_initialize()  ???:0
  4: 10 0x00000000038c3e6f scream::RRTMGPRadiation::initialize_impl()  ???:0
```
But I might be doing something wrong, or we might be missing something from master. @jgfouca have you run screamv1 tests on Chrysalis with Kokkos on in RRTMGP?

singhbalwinder:
@overfelt and @odiazib: Could you please take a look at the error above? 
@ambrad: If the only change here is upgrading to a newer version of Kokkos, could we be missing a flag that enables the use of more memory?

ambrad:
@singhbalwinder It's possible that flags could fix this problem, but that would only be a workaround. I believe that, fundamentally, the register allocation problem at compile time remains an occasional issue with the compiler. It used to be more common. A Kokkos update is not a surprising cause. The solution in the past was to simplify the kernel by splitting it up. To find out the part of the code that is the problem, isolate the file's build line so you can run it repeatedly. Then use `#if 0`/`#endif` to do (a conceptual) binary search on the parts of the file until you find the trigger. You may also consider asking Trey White if he knows of any workarounds.

singhbalwinder:
Thanks, @ambrad, for suggesting possible ways to track this down. Since it is a compile-time error, it might be easy to track the offending code block/line using #ifdefs, as you suggested.  Once we know the line, we can think about possible fixes.

ambrad:
> Once we know the line, we can think about possible fixes.

@singhbalwinder one clarification: It's unlikely a single C++ line is causing the problem. Rather, a kernel as a whole will be the issue. Thus, first isolate that kernel. Then split it up into two kernels. There's a good chance that will fix the problem. (By the way, IIRC, the register allocation error does not give any further information about code location or kernel, which is why I suggest using ifdefs to search.)

jgfouca:
@ambrad , I have not, but they will run there soon because I'm integrating the rrtmgp PR this week.

ambrad:
This branch, with no modifications, runs on PM-GPU: ERS_Ln90.ne30pg2_ne30pg2.F2010-SCREAMv1.pm-gpu_gnugpu.eamxx-L128--eamxx-sl_nsubstep2 runs but DIFFs against baselines. Based on the bless log, I believe this DIFF is expected.

Once this branch is rebased on master, we can run the full e3sm_eamxx_v1_medres suite on PM-GPU (against baselines), Chrysalis (also against baselines), and Frontier (no baselines).

bartgol:
@ambrad the CI tests fail b/c we have disabled deprecated code in kokkos. The error we get is

```
/home/runner/_work/E3SM/E3SM/externals/ekat/src/ekat/kokkos/ekat_kokkos_utils.hpp(587): error: namespace "Kokkos" has no member "atomic_compare_exchange_strong"
           if ( ! Kokkos::atomic_compare_exchange_strong(&_open_ws_slots(ws_idx), (flag_type) 0, (flag_type) 1)) {
```

I saw you added those lines, so maybe you can figure out what the fix is? I am not very familiar with the atomic operations in kk... But since deprecated code will likely go away with kokkos 5 in 6 months, we should be proactive and fix this now.

ambrad:
> @ambrad the CI tests fail b/c we have disabled deprecated code in kokkos.

I know how to fix this error. What ekat branch should I make a commit to?

ambrad:
> Edit: I can take a look too, of course, but if you already know what the right thing to do, that may be easier. If you'd rather me look into that, that's fine too.

@bartgol If you have a branch going, the fix is this:
```
ko::atomic_compare_exchange(a, b, c))
```
becomes
```
b == ko::atomic_compare_exchange(a, b, c))
```

bartgol:
> > Edit: I can take a look too, of course, but if you already know what the right thing to do, that may be easier. If you'd rather me look into that, that's fine too.
> 
> @bartgol If you have a branch going, the fix is this:
> 
> ```
> ko::atomic_compare_exchange_strong(a, b, c))
> ```
> 
> becomes
> 
> ```
> b == ko::atomic_compare_exchange(a, b, c))
> ```

Great. Yes, I can take care of it.

components/eamxx/cmake/machine-files/aurora.cmake
```diff
@@ -1,14 +1,25 @@
...
+set(CMAKE_CXX_FLAGS " -\-intel -Xclang -fsycl-allow-virtual-functions -mlong-double-64 -O3 -DNDEBUG ${SYCL_COMPILE_FLAGS}" CACHE STRING "" FORCE)
+set(CMAKE_Fortran_FLAGS "-fc=ifx -O3 -DNDEBUG -DCPRINTEL -g" CACHE STRING "" FORCE)
+set(CMAKE_C_FLAGS "-O3 -DNDEBUG" CACHE STRING "" FORCE)
+set(CMAKE_EXE_LINKER_FLAGS " -lifcore -\-intel -Xclang -fsycl-allow-virtual-functions -lsycl -mlong-double-64 -DNDEBUG ${SYCL_LINK_FLAGS} -fortlib" CACHE STRING "" FORCE)
+
+#this is needed for cime builds!
+set(NETCDF_PATH "/lus/flare/projects/E3SM_Dec/soft/netcdf/4.9.2c-4.6.1f/oneapi.eng.2024.07.30.002")
+set(NETCDF_DIR  "/lus/flare/projects/E3SM_Dec/soft/netcdf/4.9.2c-4.6.1f/oneapi.eng.2024.07.30.002")
+set(NETCDF_C_PATH "/lus/flare/projects/E3SM_Dec/soft/netcdf/4.9.2c-4.6.1f/oneapi.eng.2024.07.30.002")
+set(NETCDF_C      "/lus/flare/projects/E3SM_Dec/soft/netcdf/4.9.2c-4.6.1f/oneapi.eng.2024.07.30.002")
```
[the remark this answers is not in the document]
bartgol: If you want to run test-all-scream, I think you do need this machine file. Which speaks to a broader issue: how to make test-all-scream use config_machines for this stuff?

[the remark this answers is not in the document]
bartgol: Actually, it's fine to use these cmake files for test-all-scream. But our CIME runs still end up using these files, which is something I'd like to stop.

bartgol:
@jgfouca this PR causes an error in rrtmgp testing, due to the pool allocator running out of space:

```c++
rrtmgp_tests: /home/runner/_work/E3SM/E3SM/components/eamxx/../eam/src/physics/rrtmgp/external/cpp/rrtmgp_conversion.h:655: static T* conv::MemPoolSingleton<RealT, DeviceT>::alloc_raw(int64_t) [with T = double; RealT = double; DeviceT = Kokkos::Device<Kokkos::OpenMP, Kokkos::HostSpace>; int64_t = long int]: Assertion `s_curr_used <= s_mem.size()' failed.
```
I used gdb, and indeed we do run out of mem. But this PR does not touch rrtmgp source code (except for adding a template keyword somewhere, which is correct). Do you have any idea why things may have happened here?

bartgol:
The CI failure on CUDA for the DBG build is not one that we love to see...
```
2025-02-06T23:40:56.2897709Z nvcc error   : 'ptxas' died due to signal 11 (Invalid memory reference)
2025-02-06T23:40:56.2897784Z nvcc error   : 'ptxas' core dumped
```
I'll re-trigger, hoping it was just a fluke. Based on previous experience, however, I'm preparing for an annoying fight...


The fails for the CUDA OPT are puzzling. Besides the rrtmgp error (see comment above for Jim), the baseline_cmp are a bit of a surprise. They don't show a clear pattern (like "ah, it's rrtmgp the culprit"), so I'm a bit worried some changes (ekat? kokkos?) may be non-bfb.

bartgol:
Finally, the gh/ci runs fail due to a config error:
```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
NETCDF_C
    linked by target "scream_rrtmgp_yakl" in directory /__w/E3SM/E3SM/components/eamxx/src/physics/rrtmgp
```
Maybe @mahf708 has something to say about this, since he maintains that workflow...

ambrad:
@bartgol re: rrtmgp, note this branch turns on the Kokkos version of RRTMGP. I pinged Jim in this convo some days ago about a failure on Chrysalis that might be related to what you noted.

Re: CUDA, this branch built and ran on PM-GPU on Monday, so the ICE you see is likely version dependent. Did you see what source file triggers the ICE?

components/eamxx/CMakeLists.txt
ambrad: @bartgol here is where the Kokkos version is being turned on.

ambrad:
@bartgol ok, it must be this:
```
 nvcc error   : 'ptxas' died due to signal 11 (Invalid memory reference)
nvcc error   : 'ptxas' core dumped
gmake[2]: *** [src/dynamics/homme/homme/composec++/CMakeFiles/composec++.dir/build.make:328: src/dynamics/homme/homme/composec++/CMakeFiles/composec++.dir/compose_cedr_sl_run_local.cpp.o] Error 139
```
Could happen in other files, too, but we don't know since the build dies pretty early. How can we get on this particular compute resource and debug this? On pm-gpu, cudatoolkit/12.2 is used. Any idea what's used on these runners?

bartgol:
The ci testing uses a container image that we build. I'm not in front of my computer now (working from phone) so I can't check, but I think I picked a version of GCC and CUDA that would be very close to what's on pm-gpu (may not be exactly the same since I had to pick what spack had).

I am off tomorrow, but if there's no urgency, I can definitely hop on Blake on Monday, run the image, and try to bisect what part of that file makes CUDA unhappy.

ambrad:
@bartgol it's not urgent. I think next week I'll make a draft PR that isn't intended to be merged so that I can run the CI as much as I want. I have an idea for how I can aggressively change that file to make many small kernels, only one of which is used at runtime. I.e., I'll turn runtime options into template parameters. My guess is that will work around the problem.

ambrad:
@bartgol the latest ekat update is breaking runs. I think I wasn't clear in my explanation. Let me try again:

Suppose we had this:
```c++
if (ko::atomic_compare_exchange_strong(a, b, c))) ...
```
Transform this to:
```c++
if (b == ko::atomic_compare_exchange(a, b, c))) ...
```
For example,
```c++
if ( ! Kokkos::atomic_compare_exchange_strong(&_open_ws_slots(ws_idx), (flag_type) 0, (flag_type) 1)) {
```
becomes
```c++
if ( ! ((flag_type) 0 == Kokkos::atomic_compare_exchange(&_open_ws_slots(ws_idx), (flag_type) 0, (flag_type) 1))) {
```
Right now the commit shows, instead,
```c++
if ( ! Kokkos::atomic_compare_exchange(&_open_ws_slots(ws_idx), (flag_type) 0, (flag_type) 1)) 
```
which looks like it just removes `_strong`.

bartgol:
Ah yes, I completely misunderstood then. It's of though, the tests passed when I ran in the CUDA container. Anyhow, I'll fix that too if it's still there on Monday.

bartgol:
> There should be a unit test for the workspace manager. Related to that, do EKAT unit tests still get run in our nightlies? I thought that at one time they did, but now I'm not finding them on the dashboard.

Good question. I don't remember what I did with ekat  testing on our new CI. Tot Mart be right that it ended up getting disabled.

jgfouca:
@bartgol , i can't explain why this PR would cause the pool allocator to run out. I will look at why EKAT tests stopped running.

jgfouca:
It looks like the transition from jenkins-based nightlies to github actions caused the EKAT testing to be lost. I will work to re-enable it.

ambrad:
I updated the EKAT branch with the fix for atomic_compare_exchange. I verified wsm_dp passes on pm-gpu with this commit and fails with the commit it replaces. I'm now testing an ERS test on pm-gpu with the updated submodule and will push the submodule change when/if that passes. That plus the workaround for the ICE I pushed earlier today _may_ get CI fully passing.

ambrad:
@oksanaguba the RRTMGP PR was merged, so I suggest the next step is that you rebase this branch on master. Then we'll check the status on the various machines. We might be close.

bartgol:
> @oksanaguba the RRTMGP PR was merged, so I suggest the next step is that you rebase this branch on master. Then we'll check the status on the various machines. We might be close.

The rebase should fix the CPU fails of the CI testing. I haven't checked yet why the CUDA dbg build times out. But I will.

bartgol:
The CUDA dbg CI build that fails to even report may be due to ptxas doing bananas. I remember we were getting that when I first tried to create the CUDA image. I changed the cuda version by a tiny bit, and it went away. We may just get lucky by enabling small kernels on CUDA too, given the files that are taking forever:
```bash
[runner@blake14 full_debug]$ ps aux | awk 'NR==1 || /ptxas/'
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
runner    222685 99.7  3.3 8809116 8785852 pts/0 R+   22:29  16:49 ptxas -arch sm_90 -m64 --fmad false --generate-line-info /tmp/tmpxft_00030ceb_00000000-6_eamxx_shoc_process_interface.ptx -o /tmp/tmpxft_00030ceb_00000000-8_eamxx_shoc_process_interface.sm_90.cubin
runner    302895 99.8  4.2 11328228 11305472 pts/0 R+ 22:39   7:14 ptxas -arch sm_90 -m64 --fmad false --generate-line-info /tmp/tmpxft_00049541_00000000-6_p3_test_data.ptx -o /tmp/tmpxft_00049541_00000000-8_p3_test_data.sm_90.cubin
runner    303102  0.0  0.0  25688  3624 pts/1    S+   22:46   0:00 awk NR==1 || /ptxas/
```
The two compilations are using 10+ GB of memory right now. I will try to enable SK, and see if they build.

bartgol:
Ok, here's what I found: changing flags for shoc (even just that file) does not work. I tried to rm `-g` and `-lineinfo`, switch `-O0` to `-O3`, rm `-DNDEBUG`, rm `--fmad false` and change `-ffp-contract=off` to `-ffp-contract=fast`. Basically, I ended up with the same flags as in the release build. The problem is likely in all the macros that are/aren't defined in the kokkos headers, which add complexity to the code to build/assemble.

The good news is that shoc builds (and relatively fast) when SMALL_KERNELS is ON. So my compromise solution is to enable small kernels in the CI build for DBG. We will still be testing monolithic kernels in full_sp_debug and release, so we should cover all bases.

bartgol:
@oksanaguba In order to get the CI running and verify things work, I rebased on master (the conflict prevented CI from running). The only real conflict was the rrtmgp one (the workflow file I knew how to handle): in there, there were two ways to specify the netcdf C path. I stuck with the version from master. If that's not what you think is needed, I also pushed a branch `oksanaguba/eamxx/spot-clean-bartgol-bkp`, so you can revert to the state before the rebase.

bartgol:
Note: CI tests are failing since we did not regen baselines once Jim's PR got merged. I triggered the generate job now, but it will be a while until it gets through the queue. We can simply re-run the failed job later to check if this PR is bfb or not.

bartgol:
Ok, some good news and bad news. The CUDA dbg CI job is finally getting past the bld phase, and running tests. The bad news is that, despite the rebase on master, we get some errors. E.g., for `homme_shoc_cld_p3_rrtmgp_pg2`:

```
Constructor for Kokkos::View 'UNMANAGED' has mismatched number of arguments. The number of arguments = 0 neither matches the dynamic rank = 1 nor the total rank = 1
```
I will have to investigate hopping on blake on monday. The release build only has the baseline_cmp fails, which should go away if we retest after the baseline-gen job runs.

ambrad:
> The interesting thing is that the error shows up during finalization (as you can see from the fact that the unit test completed 100% of the iterations).

I think I know the cause. It's some new nullify routines I created to handle Kokkos's new (>= 4.3, I think) way of dealing with view-of-view initialization. I'll push a fix.

bartgol:
I lost track of the status of this PR. What are we still missing in order to start integrating?

oksanaguba:
@bartgol from yesterdya's meeting, testing on frontier

bartgol:
> @bartgol from yesterdya's meeting, testing on frontier

Thanks!

ndkeen:
Using this branch as of Mar6, I am seeing a build error:
```
/p/lustre5/keen3/e3sm_scratch/tuo/oksanaguba_eamxx_spot-clean/t.ne30pg2_ne30pg2.F2010-SCREAMv1.oksanaguba_eamxx_spot-clean.5d.n0002t4xX11XX1.nocosp.notend.dins.sk.gfx942/build/gnugpu/mpich/nodebug/threads/include/View/Kokkos_Vie\
wLegacy.hpp:944:19: error: static assertion failed due to requirement 'ViewMapping<Kokkos::ViewTraits<char *, Kokkos::LayoutRight, Kokkos::Device<Kokkos::Serial, Kokkos::HostSpace>, Kokkos::MemoryTraits<0>>, Kokkos::ViewTraits<c\
har *, Kokkos::LayoutRight, Kokkos::Device<Kokkos::Serial, Kokkos::HIPSpace>, Kokkos::Experimental::EmptyViewHooks>, void>::is_assignable': Incompatible View copy assignment
  944 |     static_assert(Mapping::is_assignable, "Incompatible View copy assignment");
      |                   ^~~~~~~~~~~~~~~~~~~~~~
/p/lustre5/keen3/wacmy/oksanaguba_eamxx_spot-clean/components/eamxx/src/share/field/field.cpp:293:17: note: in instantiation of function template specialization 'Kokkos::View<char *, Kokkos::LayoutRight, Kokkos::Device<Kokkos::S\
erial, Kokkos::HostSpace>, Kokkos::MemoryTraits<0>>::operator=<char *, Kokkos::LayoutRight, Kokkos::Device<Kokkos::Serial, Kokkos::HIPSpace>, Kokkos::Experimental::EmptyViewHooks>' requested here
  293 |   m_data.h_view = Kokkos::create_mirror_view(m_data.d_view);
      |                 ^
In file included from /p/lustre5/keen3/wacmy/oksanaguba_eamxx_spot-clean/components/eamxx/src/share/field/field.cpp:1:
In file included from /p/lustre5/keen3/wacmy/oksanaguba_eamxx_spot-clean/components/eamxx/src/share/field/field.hpp:4:
In file included from /p/lustre5/keen3/wacmy/oksanaguba_eamxx_spot-clean/components/eamxx/src/share/field/field_header.hpp:4:
In file included from /p/lustre5/keen3/wacmy/oksanaguba_eamxx_spot-clean/components/eamxx/src/share/field/field_identifier.hpp:4:
In file included from /p/lustre5/keen3/wacmy/oksanaguba_eamxx_spot-clean/components/eamxx/src/share/field/field_layout.hpp:5:
In file included from /p/lustre5/keen3/wacmy/oksanaguba_eamxx_spot-clean/components/eamxx/src/share/eamxx_types.hpp:4:
In file included from /p/lustre5/keen3/wacmy/oksanaguba_eamxx_spot-clean/externals/ekat/src/ekat/ekat.hpp:6:
```

ambrad:
@bartgol see the above. The build error in the AD might be legitimate. But I'm not sure why I don't see it in any of my medres testing on PM, Chrysalis, and Frontier.

ndkeen:
@ambrad yes

components/eamxx/src/physics/rrtmgp/CMakeLists.txt
jgfouca: This code looks wrong. If we are building as part of a E3SM/CIME case, there's no need to find netcdf, it should already be a target. We can check if the `netcdf` target already exists or check if SCREAM_CIME_BUILD is on and not do this find if either is true.

jgfouca: Note, this is issue is not related specifically to this PR.

jgfouca: Probably, but the source won't build without netcdf. We only need to do this find if we are standalone eamxx.

ndkeen:
Initial results show that I'm getting same performance with this branch on pm-gpu. That's with 8 nodes at ne120 and with 32 nodes of ne256.  Additionally, the cases are BFB (after 5 days each).

components/eamxx/src/physics/rrtmgp/CMakeLists.txt
```diff
@@ -181,6 +186,7 @@ endif()
 # NOTE: cannot use 'PUBLIC' in target_link_libraries,
 #       since yakl_process_target already used it
 #       with the "plain" signature
+
 find_library(NETCDF_C netcdf HINTS ${NetCDF_C_PATH}/lib)
 target_link_libraries(eamxx_rrtmgp_interface ${NETCDF_C} rrtmgp scream_share Kokkos::kokkos)
```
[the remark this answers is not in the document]
ambrad: I think I lost a change when doing the rebase. I'll add a commit to fix this.

oksanaguba:
do you mean RANKS_BIND variable? I see it set in my runs
```
      Setting Environment RANKS_BIND=list:0-7,104-111:8-15,112-119:16-23,120-127:24-31,128-135:32-39,136-143:40-47,144-151:52-59,156-163:60-67,164-171:68-75,172-179:76-83,180-187:84-91,188-195:92-99,196-203
```
IIRC it was also needed for performance accr to this commit, https://github.com/E3SM-Project/E3SM/pull/6916/commits/6a211776b0042e562c3473a1378eb68b9a8743f6 , but there are other changes in that commit, so that may be not accurate.

(replying to a deleted comment?)

oksanaguba:
Regarding this message (which was in one of deleted comments? i am not sure):
```
x4115c5s1b0n0.hostmgmt2115.cm.aurora.alcf.anl.gov 550: bash: module: line 1: syntax error: unexpected end of file
bash: error importing function definition for `module'
```
**The issue with it is that is appears on each rank in e3sm log. So we need to get rid of it** for sure, but it is not clear how. I read that it could be a parser error of EOF-type of thing https://stackoverflow.com/questions/6366530/bash-syntax-error-unexpected-end-of-file . The same error appears at the beginning of run.$CASE... status file. That one is done by ./CIME/case/case_run.py and i tried to trace with python function does module loading, but did not finish. When executing .env*.sh file, i do not see such message.
Separately, there is function 
```
onguba@aurora-uan-0012:/lus/flare/projects/E3SM_Dec/onguba/scratch/v1-1024-mar04-F2010-SCREAMv1-nodes450/run> declare -f module
module () 
{ 
    eval "`$LMOD_CMD sh "$@"`"
}
```
but it also did not produce the error.

components/eamxx/src/physics/rrtmgp/CMakeLists.txt
bartgol: @jgfouca IMO, it's fine to call find_package (and the likes) regradless of the value of SCREAM_CIME_BUILD... When the pkg/lib/path has alreday been found, it should be a no-op.

That said, the warning reported by Naser is indeed wrong, as it's likning a path instead of a tgt/lib...

bartgol:
> @bartgol see the above. The build error in the AD might be legitimate. But I'm not sure why I don't see it in any of my medres testing on PM, Chrysalis, and Frontier. This seems fishy to me, but I don't know the `field` code at all: `Kokkos::Device<Kokkos::Serial, Kokkos::HIPSpace>`.
> 
> @ndkeen what machine is this on? I see HIP but it's not Frontier. Is this Tuolumne?

I'm looking into this today.

components/eamxx/src/physics/rrtmgp/CMakeLists.txt
jgfouca: Is it fine to mix find_package and find_library?

oksanaguba:
@mahf708  yes, we have 4 machine files in eamxx right now: aurora, auroracpu, sunspot-pvc, sunspotcpu . i do not know about removing sunspot machine completely, Az, for example, might need it (it is supposed to be now only for ANL). But i will remove 3 eamxx files, since they are not used. FYI I haven't run  on aurora with cpus for a while.

ndkeen:
This branch will cause build failure (discussed above) on machine with AMD MI300a (tuolumne). Without this branch, we are able to build/run on the machine (all tests so far, including ne1024 production-like case).

bartgol:
In #7103 , we had a small conversation on this subject (but we can of course discuss more!). @ambrad and I were thinking it may be best to get things going on Aurora first, and then decide what to do with the MI300 support.

At SIAM CSE, HPE/AMD folks said that the next iteration of AMD gpus will not be an APU (which kind of shocked me, since they were showing a bunch of good/promising metrics), so unless we plan to run production stuff on Tuolomne, maybe we could just "skip" supporting the APU generation? I think the execs should prob weigh in on how important it is to run on MI300. It would also help to get a better idea of what will happen to Tuolomne going forward (e.g., will it stay open to everyone?).

AaronDonahue:
> This branch will cause build failure (discussed above) on machine with AMD MI300a (tuolumne). Without this branch, we are able to build/run on the machine (all tests so far, including ne1024 production-like case).
> 
> Pointing this out to @AaronDonahue who is also interested in this machine and @tcclevenger who has been trying to help me determine if the atomics issue in kokkos might fix the performance issues on mi300a.

@ndkeen , thank you for the heads up.  As much as I would like to have `master` running on tuolumne I think that getting the model running on Aurora is a high priority.  Hopefully there will be a fix for tuolumne down the road.

ambrad:
> Without this branch, we are able to build/run on the machine (all tests so far, including ne1024 production-like case).

Ok, this is good information. This makes me believe there _must_ be a flag in Kokkos 4.5 that restores the non-unified-memory behavior we currently can get on the machine.

bartgol:
> Ok, this is good information. This makes me believe there _must_ be a flag in Kokkos 4.5 that restores the non-unified-memory behavior we currently can get on the machine.

I thought the answer to your question was "no", since I see the impl in `Kokkos_HIP_Space.hpp`, using the unified mem approach is a certain macro was defined, and I thought that macro was defined on MI300 gpus.

HOWEVER, I now see that the macro is defined if one enables the `AMD_GFX942_APU` kokkos arch, but it's NOT defined if one enables `AMD_GFX942`. I'm assuming the latter is the same as the former (in terms of mach-specific optimization) _except_ for the unified mem space assumption. @ndkeen which `Kokkos_ARCH_XYZ` cmake var do you currently enable when builkding on Tuolomne?

ndkeen:
interesting.  
```
option(Kokkos_ARCH_AMD_GFX942 "" ON)
```

bartgol:
Ah, wait. If the AMD_GFX942 arch is enabled, Kokkos by default checks if the APU version can be enabled, by doing a try_compile. I'm looking at their cmake machinery, and I _think_ that they did not offer a way to prevent this. Let me ask on their slack channel.

bartgol:
Ah wait, maybe I'm wrong. @ndkeen Can you try to set `option(Kokkos_ARCH_AMD_GFX942_APU "" OFF)` in your mach file, and see? You don't need to build, you can just stop after CMake runs, and check CMakeCache.txt (in the bld/cmake-bld dir). Right now, you should see this APU var set to ON, while with the cmake mach file mod you _should_ see it set to OFF...

ndkeen:
Following up here, after Luca and I iterated, it looks like if I add `-DKokkos_ARCH_AMD_GFX942_APU=OFF` to the Kokkos options, I can get scream to build on tuolumne with this branch.  So something must have changed between the version of kokkos we have now (4.2?) and 4.5, such that this was needed (and/or the code we use to interface with kokkos).

ambrad:
With a test merge to next, this is the testing status:

* PM-GPU: e3sm_eamxx_v1_medres passes against baselines.
* Frontier: e3sm_eamxx_v1_medres passes.
* Chrysalis: A set of 5 tests, listed below, pass against baselines.

Chrysalis test sample:
```
  ERS_Ld3.ne30pg2_r05_IcoswISC30E3r5.WCYCL1850.chrysalis_intel.allactive-nlmaps
  ERS_Ln22.ne30pg2_ne30pg2.F2010-SCREAMv1.chrysalis_intel.eamxx-L128--eamxx-output-preset-4
  SMS_Ld1.ne30pg2_r05_IcoswISC30E3r5.WCYCL1850.chrysalis_intel.allactive-wcprod_1850
  HOMMEBFB_P24.f19_g16_rx1.A.chrysalis_intel
  HOMME_P24.f19_g16_rx1.A.chrysalis_intel
```

Still, I think I'll run e3sm_integration on Chrysalis before merging to next.

bartgol:
Looks like you already merge to next?

ambrad:
> Looks like you already merge to next?

No, that's just a test run pushed to my fork.

cime_config/machines/cmake_macros/oneapi-ifx_auroracpu.cmake
```diff
@@ -0,0 +1,19 @@
...
+
+
+
+
+
+
+
+
+
+
```
amametjanov, line 19: Please remove this cmake file.

cime_config/machines/cmake_macros/oneapi-ifx_sunspotcpu.cmake
```diff
@@ -0,0 +1,19 @@
...
+
+
+
+
+
+
+
+
+
+
```
amametjanov, line 19: Please remove this cmake file.

cime_config/machines/cmake_macros/oneapi-ifxgpu_sunspot-pvc.cmake
```diff
@@ -0,0 +1,9 @@
+
+string(APPEND CMAKE_EXE_LINKER_FLAGS " -lmkl_intel_lp64 -lmkl_sequential -lmkl_core -fsycl-device-code-split=per_kernel -fsycl-max-parallel-link-jobs=16 -Wl,--no-relax")
+if (compile_threaded)
+  string(APPEND CMAKE_EXE_LINKER_FLAGS " -fiopenmp -fopenmp-targets=spir64")
+endif()
+string(APPEND KOKKOS_OPTIONS " -DCMAKE_CXX_STANDARD=17 -DKokkos_ENABLE_SERIAL=On -DKokkos_ARCH_INTEL_PVC=On -DKokkos_ENABLE_SYCL=On -DKokkos_ENABLE_EXPLICIT_INSTANTIATION=Off")
+string(APPEND SYCL_FLAGS " -\-intel -fsycl -fsycl-targets=spir64_gen -mlong-double-64 -Xsycl-target-backend \"-device 12.60.7\"")
+
+set(SCREAM_MPI_ON_DEVICE OFF CACHE STRING "")
```
amametjanov, line 9: Please remove this cmake file.

cime_config/machines/config_machines.xml
```diff
@@ -3441,7 +3441,7 @@
        <arguments>
           <arg name="total_num_tasks">-np {{ total_tasks }} --label</arg>
           <arg name="ranks_per_node">-ppn {{ tasks_per_node }}</arg>
-          <arg name="ranks_bind">-envall</arg>
+          <arg name="ranks_bind">--no-vni --cpu-bind $ENV{RANKS_BIND} -envall</arg>
```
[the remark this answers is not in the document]
amametjanov: I think `--no-vni` is not needed anymore: can be removed.
Also, default is `-genvall` now: `-envall` can be removed.

components/eamxx/cmake/machine-files/aurora.cmake
```diff
@@ -1,14 +1,25 @@
 include(${CMAKE_CURRENT_LIST_DIR}/common.cmake)
 common_setup()
 
-include (${EKAT_MACH_FILES_PATH}/kokkos/serial.cmake)
+include (${EKAT_MACH_FILES_PATH}/kokkos/intel-pvc.cmake)
 include (${EKAT_MACH_FILES_PATH}/mpi/other.cmake)
-
 set(EKAT_MPIRUN_EXE "mpiexec" CACHE STRING "" FORCE)
 set(EKAT_MPI_NP_FLAG "-np" CACHE STRING "" FORCE)
 set(EKAT_MPI_EXTRA_ARGS "--label --cpu-bind depth -envall" CACHE STRING "")
```
amametjanov, line 8: For reproducibility, please replace `--cpu-bind depth -envall` with
```
--cpu-bind list:0-7,104-111:8-15,112-119:16-23,120-127:24-31,128-135:32-39,136-143:40-47,144-151:52-59,156-163:60-67,164-171:68-75,172-179:76-83,180-187:84-91,188-195:92-99,196-203
```

amametjanov (commented):
Comments about older cmake.

oksanaguba:
@amametjanov  i addressed your comments except for the ekat variable that is actually not used in cime runs (I think). It can be sorted later. The rest is done, -evvall and -no-vni , and extra files are removed. Could you please re-review?

amametjanov:
Thanks for removing older cmake.
This is breaking CPU builds: PR can go in and I'll fix CPU build and run errors later.

tcclevenger (approved):
All CI passing

ambrad:
Great work, everyone. I'll merge this to next once I get the go-ahead on the devops channel.

ambrad:
Reverted from next.

pm-cpu e3sm_developer_next_gnu and e3sm_integration_next_intel need investigation.

Also gcp12 e3sm_integration_next_gnu.

https://my.cdash.org/index.php?project=E3SM&date=2025-03-19

oksanaguba:
Abhishek just posted that we need to start using in April
```
--cpu-bind list:1-8,105-112:9-16,113-120:17-24,121-128:25-32,129-136:33-40,137-144:41-48,145-152:53-60,157-164:61-68,165-172:69-76,173-180:77-84,181-188:85-92,189-196:93-100,197-204
```

PR #6917, opened 2025-01-19
Add isotropic surface stress to MPAS-Ocean

proteanplanet:
Introduces 4-point averaging to ocean surface stress (wind and sea ice stress) around each vector point on the MPAS-Ocean C-grid, instead of averaging only across the two on-edge cell center surface stress points to the vector point.  This affects coupling with both the atmosphere and sea ice models, which both couple surface stress through the scalar (cell center) grid point in MPAS-Ocean.  In so doing, we ensure that surface stress is independent of the orientation of the mesh relative to the wind vectors or sea ice drift vector, thereby making this approach isotropic.  

The code has been tested using both `config_bulk_wind_stress_interp_isotropic = true`  (on) and `config_bulk_wind_stress_interp_isotropic = false` (off) in fully coupled E3SM pre-industrial 100 and 80 year B-cases, respectively.  The change conserves mass and energy and has minimal impact on the overall coupled simulation, but provides more robust forcing numerics for the ocean, with feedbacks to sea ice.  Figures supporting this are provided in comments attached to this pull request. 

This pull request was authored by @xylar and @proteanplanet combined.

[NML]
[CC]

### DIFF
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -275,6 +275,8 @@
 <config_cvmix_prandtl_number>1.0</config_cvmix_prandtl_number>
 <config_cvmix_background_scheme>'constant'</config_cvmix_background_scheme>
 <config_cvmix_background_diffusion>0.0</config_cvmix_background_diffusion>
+<config_cvmix_background_diffusion_passive>0.0</config_cvmix_background_diffusion_passive>
+<config_cvmix_background_diffusion_passive_enable>.false.</config_cvmix_background_diffusion_passive_enable>
 <config_cvmix_background_viscosity>1.0e-4</config_cvmix_background_viscosity>
 <config_cvmix_BryanLewis_bl1>8.0e-5</config_cvmix_BryanLewis_bl1>
 <config_cvmix_BryanLewis_bl2>1.05E-4</config_cvmix_BryanLewis_bl2>
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -332,6 +332,7 @@
 <!-- forcing -->
 <config_use_variable_drag>.false.</config_use_variable_drag>
 <config_use_bulk_wind_stress>.true.</config_use_bulk_wind_stress>
+<config_bulk_wind_stress_interp_isotropic>.false.</config_bulk_wind_stress_interp_isotropic>
 <config_use_bulk_thickness_flux>.true.</config_use_bulk_thickness_flux>
 <config_flux_attenuation_coefficient>0.001</config_flux_attenuation_coefficient>
 <config_flux_attenuation_coefficient_runoff>10.0</config_flux_attenuation_coefficient_runoff>
--- a/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
+++ b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
@@ -332,7 +332,7 @@
 <!-- forcing -->
 <config_use_variable_drag>.false.</config_use_variable_drag>
 <config_use_bulk_wind_stress>.true.</config_use_bulk_wind_stress>
-<config_bulk_wind_stress_interp_isotropic>.false.</config_bulk_wind_stress_interp_isotropic>
+<config_bulk_wind_stress_interp_isotropic>.true.</config_bulk_wind_stress_interp_isotropic>
 <config_use_bulk_thickness_flux>.true.</config_use_bulk_thickness_flux>
 <config_flux_attenuation_coefficient>0.001</config_flux_attenuation_coefficient>
 <config_flux_attenuation_coefficient_runoff>10.0</config_flux_attenuation_coefficient_runoff>
--- a/cime_config/machines/config_batch.xml
+++ b/cime_config/machines/config_batch.xml
@@ -291,7 +291,6 @@
       <directive> --job-name={{ job_id }}</directive>
       <directive> --nodes={{ num_nodes }}</directive>
       <directive> --output={{ job_id }}.%j </directive>
-      <directive> --exclusive </directive>
     </directives>
   </batch_system>
 
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -54,6 +54,8 @@ _TESTS = {
     "e3sm_land_exenoshare" : {
         "time"  : "0:45:00",
         "tests" : (
+            "ERS.f19_g16.IERA5ELM",
+            "ERS.f19_g16.IERA56HRELM",
             "ERS_Ld20.f45_f45.IELMFATES.elm-fates",
             "ERS.f09_g16.IELMBC.elm-simple_decomp",
             "ERS.hcru_hcru.IELM.elm-multi_inst",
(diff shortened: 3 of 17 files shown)


proteanplanet:
Introduces 4-point averaging to ocean surface stress (wind and sea ice stress) around each vector point in MPAS-Ocean, as per the right panel in the following figure:

Surface_Current_Coupling_MPAS-Ocean

The left panel illustrates the default up to the V3 LR and NARRM configurations, the right panel is intended for the V3 SORRM, HR and BGC configurations, and subsequent configurations of E3SM.

Two tests were conducted with this new code in B-cases, as documented here for the [anisotropic](https://acme-climate.atlassian.net/wiki/spaces/CM/pages/4907532289/20240528.v3.LR.piC-spinup-oaniso.chrysalis) and [isotropic](https://acme-climate.atlassian.net/wiki/spaces/CM/pages/4904943617/20240528.v3.LR.piC-spinup-oisotr.chrysalis) cases, respectively.   Each case was run for at least 90 years with code stability for the changes introduced in this PR.  A comparison of the results follows.

proteanplanet:
### Impact of Isotropic Surface Stress on the Ocean

One of the largest impacts of introducing isotropic surface ocean stress is in sea ice extent and snow volume.  In the lemnisc below, the difference between anisotropic (blue) and isotropic (red) 90-year PI spinups illustrates that both of these sea ice metrics are different ad the 95% compatibility interval, where dotted red lines indicates no difference, and solid red lines illustrates a difference within each annual solar cycle. Even though the mean values appear identical, the sea ice climate is different due to variability. 

E3SM_sea_ice_lemnisc 0001-0089 0001-0089 windstressold windstressnew icepack extent volume snow 3

There appears is no sizable change to AMOC for the rest simulations described above, comparing here also against the LR V3 Spinup:

MOC_Comparison_Windstress

However, it appears that the isotropic wind stress case may allow the model to spinup faster for key through-flow statistics:

Southern_Ocean_Transport

While the impact of this change is subtle, it offers more robust numerics in regions of heavily distorted grid cells, as illustrated in the mesh figure in the previous comment above.  The change does not degrade the model, and E3SM runs stably with the improved coupling numerics when fully coupled, and it is therefore worth adopting.

proteanplanet:
@mark-petersen, @vanroekel, @cbegeman  - this is now ready for your review.

vanroekel:
@proteanplanet -- the code looks great to me by inspection.  Do you want me to do any testing?

vanroekel:
I think your B-case testing is great.  Two little questions.
1. did you do the standard developer suite testing (PEM, PET, etc...)?
2. does the isotropic averaging add any noticeable increase in computational cost?  I'm guessing not.

proteanplanet:
>     2. does the isotropic averaging add any noticeable increase in computational cost?  I'm guessing not.

As a rough comparison, cost of this code with the flag switched on for Chrysalis is [6204.43 pe-hrs/sim_year,  12.87 SYPD for a 25-year simulation](https://pace.ornl.gov/exp-details/181701), as compared to [6146.00 pe-hrs/sim_year at 13.00 SYPD for the anisotropic case](https://pace.ornl.gov/exp-details/181273), for an ocean model cost of 17.21 SYPD for the former and 17.43 SYPD for the latter.   

In other words, the numerical change is in the noise relative to the climate changes caused by the different approach across the system.

proteanplanet:
>     1. did you do the standard developer suite testing (PEM, PET, etc...)?

The code passes the `e3sm_cryo_developer` test suite (PEM, PET, ERS, SMS with ice shelves) with `config_bulk_wind_stress_interp_isotropic = true`.

components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
```diff
@@ -332,6 +332,7 @@
 <!-- forcing -->
 <config_use_variable_drag>.false.</config_use_variable_drag>
 <config_use_bulk_wind_stress>.true.</config_use_bulk_wind_stress>
+<config_bulk_wind_stress_interp_isotropic>.false.</config_bulk_wind_stress_interp_isotropic>
```
proteanplanet, line 335: A question for @cbegeman and @vanroekel:  Given the extensive testing in B-cases, should we immediately just switch on isotropic wind stress averaging as the default in master, thus changing, this to `true`.

vanroekel: I would support making isotropic default behavior based on your testing

vanroekel (approved):
approving by visual inspection of code and extensive testing from @proteanplanet

components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml
mark-petersen: I would as well.

cbegeman: I agree with changing the default to true

proteanplanet: Done!  Just left for @cbegeman to approve.

cbegeman (approved):
Approving on the basis of code inspection and @proteanplanet's testing. Thanks for identifying and prioritizing this, @proteanplanet!

proteanplanet:
Tagging @xuezhengllnl, @darincomeau and @njeffery  - this is ready for HR, SORRM, and HES configs.

njeffery:
@proteanplanet : The icepack version on your branch is [8fad768](https://github.com/E3SM-Project/Icepack/tree/8fad768ce400536904f376376e91c698a82882ba)] which is quite old and doesn't have the Icepack bgc.  I'm planning on doing a coupled bgc run using the latest [22b8769](https://github.com/E3SM-Project/Icepack/tree/22b87692dfe5346b3ab7923fc56e3d9b2127c9a3) to compare against my baseline.  Just want to confirm that that's correct.

I'm also planning on using config_bulk_wind_stress_interp_isotropic = true .  Are there any other changes I should be aware of?

jonbob:
@proteanplanet -- do you consider this climate-changing? Or just non-BFB?

jonbob:
Link to mpas-analysis page comparing runs is here: [intercomparison](https://portal.nersc.gov/project/m2136/bin/iice/mpas-a.cgi?url1=https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.afroberts/E3SMv3/20240528.v3.LR.piC-spinup-oaniso.chrysalis/mpas_analysis/ts_0001-0080_climo_0001-0080&label1=Anisotropic&url2=https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.afroberts/E3SMv3/20240528.v3.LR.piC-spinup-oisotr.chrysalis/mpas_analysis/ts_0001-0080_climo_0001-0080&label2=Isotropic&category=&component=&plot=&diff=0&dconfig=wcycl&)

proteanplanet:
> @proteanplanet -- do you consider this climate-changing? Or just non-BFB?

Mean climate values are similar, but moments differ, thereby making this PR climate changing.

jonbob:
Passes:
* PET_Ln9_PS.ne30pg2_r05_IcoswISC30E3r5.WCYCL1850.chrysalis_intel.allactive-mach-pet
* PEM_Ln9.ne30pg2_r05_IcoswISC30E3r5.WCYCL1850.chrysalis_intel

with expected NML and baseline DIFFs

merged to next

jonbob:
merged to master and expected baseline and NML DIFFs blessed -- except for pm-cpu, which is down this week

PR #6919, opened 2025-01-21
EAMxx: enhance vert remap support in DataInterpolation

bartgol:
Support using a source pressure profule reconstructed from ps and hyam/hybm.

---

This is what SPA currently does, so this addition is needed in order to start using DataInterpolation in SPA. Note: another upgrade to DataInterpolation is needed, to handle IOP, but I want to integrate one thing at a time.

The unit test for the new feature was surprisingly simple. Since the unit test builds p3d in a way that it is basically the tensor product of s2d (a 2d scalar) and p1d, all we had to do was saving p2d=s2d, and hybm=p1d, with hyam=0. This makes the reconstructed p3d=p2d*p1d, so the expected fields for `Dynamic3DRef` are the same as in the case of `Dynamic3D`.

Also, the name `Dynamic3DRef` is up for discussion. With `Dynamic3DRef` I wanted to emphasize these were "reference" levels. Other names I considered are `Dynamic3DFromPs`, `DynamicHybrid3D`/`Dynamic3DHybrid`, or `DynamicPseudo3D`. Let me know if you have any preference here.

### DIFF
--- a/components/eamxx/src/share/io/scream_scorpio_interface.cpp
+++ b/components/eamxx/src/share/io/scream_scorpio_interface.cpp
@@ -1204,19 +1204,6 @@ void read_var (const std::string &filename, const std::string &varname, T* buf, 
         " - time len: " + std::to_string(f.time_dim->length));
     err = PIOc_setframe(f.ncid,var.ncid,frame);
     check_scorpio_noerr (err,f.name,"variable",varname,"read_var","setframe");
-  } else if (time_index>=0) {
-    // This is a bit of a hacky usage. We want to read a time index of a var,
-    // but the time dim is NOT unlimited in the input file. Apparently, SCORPIO
-    // supports this, and some E3SM input files are indeed like this, so we
-    // need to support it here too, hacky as it may be.
-    frame = time_index;
-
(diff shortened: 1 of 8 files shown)


components/eamxx/src/share/io/scream_scorpio_interface.cpp
```diff
@@ -1204,19 +1204,6 @@ void read_var (const std::string &filename, const std::string &varname, T* buf,
         " - time len: " + std::to_string(f.time_dim->length));
     err = PIOc_setframe(f.ncid,var.ncid,frame);
     check_scorpio_noerr (err,f.name,"variable",varname,"read_var","setframe");
-  } else if (time_index>=0) {
```
bartgol, line 1207: This branch was at some point needed to allow IOP to read data with a time dimension that was not unlimited. However, since then, I have added `scorpio::pretend_dim_is_unlimited(..)`, which allows one to treat a normal dim as unlimited, and marking all vars with that dim as "time_dep". This ensures that the branch above is used for these variables. This is already used in the IOPDataManager structure to handle this case.

tcclevenger (approved):

PR #6920, opened 2025-01-21
EAMxx: make SPA use DataInterpolation

bartgol:
Allow SPA to use the DataInterpolation class. Changes include:
 - accommodate IOP horizontal remapping
 - allow to set reference timestamp from API rather than read from file
 - replace SPAFunctions with DataInterpolation
 - remove unit test in spa folder
   
 [Non BFB]
 
   ---
   
This PR will be rebased on master once #6914 and #6919 are merged.


components/eamxx/src/physics/spa/eamxx_spa_process_interface.cpp
```diff
@@ -53,201 +53,78 @@ void SPA::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
...
+  // is setup, pint must store the correct layout of a field defined at interfaces; it does not
+  // have to have COL dimension. It just need to have the ILEV tag in the layout AND have alloc
+  // properties compatible with SCREAM_PACK_SIZE.
+  // NOTE: we could just add p_int as a required field, but that would be misleading in the DAG
+  auto pmid = get_field_in("p_mid");
+  Field pint(FieldIdentifier("p_int",m_model_grid->get_vertical_layout(false),Pa,m_model_grid->name()));
+  pint.get_header().get_alloc_properties().request_allocation(SCREAM_PACK_SIZE);
+  pint.allocate_view();
+
+  util::TimeStamp ref_ts (1,1,1,0,0,0); // Beg of any year, since we use yearli periodic timeline
```
AaronDonahue, line 87: "yearly"

components/eamxx/src/physics/spa/eamxx_spa_process_interface.cpp
```diff
@@ -53,201 +53,78 @@ void SPA::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
...
+
+  DataInterpolation::RemapData remap_data;
+  remap_data.hremap_file = spa_map_file;
+  if (m_iop_data_manager!=nullptr) {
+    // IOP cases cannot have a remap file. We will create a IOPRemapper as the horiz remapper
+    EKAT_REQUIRE_MSG(spa_map_file == "",
+      "Error! Cannot define spa_remap_file for cases with an Intensive Observation Period defined. "
+      "The IOP class defines it's own remap from file data -> model data.\n");
+
+    // TODO: expose tgt lat/lon in IOPDataManager, to avoid injencting knowledge
```
AaronDonahue, line 99: "injecting"

AaronDonahue (approved):
Just a couple spelling mistakes that I found.  Otherwise it looks good.  Its great to see that we can remove so many lines of code and define SPA in a much simpler framework.

AaronDonahue (approved):

bartgol:
All spa tests are diffing. I was _not_ expecting that, so I need to investigate further.

bartgol:
Most fails are due to how SPA data is used. Current master consider spa data for month X to be at time "beg of month X". Since the SPA file shows the middle of the month as the "time" var (which is logical, for monthly averaged data), DataInterpolation does a different interpolation than master. Changing the time var to store the time corresponding to beginning of the month for each month makes all diffing tests pass cmp baselines. Since we agreed in a slack convo that the behavior of the current branch is less biased, we're going to ignore those fails. The reasoning is that, say, Jan02 is closer to Dec than Feb, so the Dec avg should be used rather than the Feb one. Master interpolates current_month and next_month, while this branch interpolates current_month and "closest_month" (which could be next or prev).

There is one runtime fail though, when running dp. I am investigating.

bartgol:
All standalone tests pass if I modify the spa file to point to the beg of the month for each month (rather than middle of the month), so that DataInterpolation does the same as the old SPA functions. I am now testing a v1 case, to make sure that's the case also in v1. If so, we can merge and re-bless all eamxx baselines.

bartgol:
I ran one of our v1 nightly tests on mappy with this PR. With the current spa file, it diffs (as expected), but with a "doctored" spa file (where each month time coord is at 00:00 of the 1st), the outputs are identical.

@AaronDonahue I feel confident that the diffs are just due to how SPA slices were declared in the input file. If you don't have any more comment, I will merge (maybe on Monday, since it will require bless requests).

tcclevenger (approved):

AaronDonahue (approved):

rljacob:
notes:  looking at fails.

bartgol:
I'm confident enough that this PR only causes expected DIFFs. Merging.

PR #6921, opened 2025-01-21
Add the dust emission cap 

yunpengshan2014:
To address the issue of overly strong dust emissions causing crashes in high-resolution (HR) simulations, a dust emission cap has been implemented. The cap ensures that dust emissions remain within a reasonable range to maintain simulation stability.

Additionally, the dust emission cap and its enabling switch have been added as configurable namelist items, allowing users to easily activate and adjust the cap as needed for their specific simulation setups.

### DIFF
--- a/components/eam/src/chemistry/modal_aero/aero_model.F90
+++ b/components/eam/src/chemistry/modal_aero/aero_model.F90
@@ -168,7 +168,7 @@ contains
     call mpibcast(seasalt_emis_scale, 1, mpir8,   0, mpicom)
 !<shanyp 01132025
     call mpibcast(dstemislimit, 1,                       mpir8,   0, mpicom) !dstemislimit
-    call mpibcast(dstemislimitswith, 1,                  mpilog,  0, mpicom) !dstemislimit
+    call mpibcast(dstemislimitswitch, 1,                  mpilog,  0, mpicom) !dstemislimit
 !shanyp 01130225>
 #endif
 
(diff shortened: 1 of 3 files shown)


rljacob:
@yunpengshan2014 CI testing indicates fails.  Try your branch on a gnu compiler.

components/eam/src/chemistry/modal_aero/aero_model.F90
```diff
@@ -2834,6 +2848,18 @@ subroutine aero_model_emissions( state, cam_in )
...
        do m=1,dust_nbin+dust_nnum
           mm = dust_indices(m)
+!<shanyp 07052024
+          if(dstemislimitswitch) then
+           do icol=1,ncol
+            if((cam_in%cflx(icol,mm).ge.dstemislimit).and.(m.le.dust_nbin)) then
+             mmn=dust_indices(m+2)
+             cam_in%cflx(icol,mmn)=cam_in%cflx(icol,mmn)*dstemislimit/cam_in%cflx(icol,mm)
+             cam_in%cflx(icol,mm)=dstemislimit
+             write(iulog,'(a,1x,i10)') 'The dust emission cap is hit at icol=', icol
```
crterai, line 2858: I've found that the `icol` value typically is not helpful in identifying the geographic location where things occur (are triggered). Would it be possible to get the `lat` `lon` values and also the timestep value so that we know where and when they occur?

components/eam/src/chemistry/modal_aero/aero_model.F90
```diff
@@ -102,7 +102,10 @@ module aero_model
   real(r8)          :: sol_factic_interstitial = 0.4_r8
   real(r8)          :: seasalt_emis_scale
   real(r8)          :: small = 1.e-36
-
+!<shanyp 01132025
```
crterai, line 105: since git will keep track of when and who added these lines, all comments relating to author and date of changes should be removed.

rljacob:
Replaced by #6937

PR #6923, opened 2025-01-21
EAMxx: PBL entrainment budget diags

mahf708:
Adds planetary boundary layer entrainment budget diagnostics as in mülmenstädt ... mahfouz ... et al (2024), _ACP_, doi: 10.5194/acp-24-13633-2024.

[BFB]

### DIFF
--- a/components/eamxx/src/diagnostics/pbl_entrainment_budget.cpp
+++ b/components/eamxx/src/diagnostics/pbl_entrainment_budget.cpp
@@ -1,0 +1,332 @@
+#include "diagnostics/pbl_entrainment_budget.hpp"
+
+#include "diagnostics/pbl_entrainment_budget_util.hpp"
+#include "ekat/ekat_workspace.hpp"
+#include "ekat/kokkos/ekat_kokkos_utils.hpp"
+#include "share/util/scream_universal_constants.hpp"
+
+namespace scream {
+
+PBLEntrainmentBudget::PBLEntrainmentBudget(const ekat::Comm &comm,
+                                           const ekat::ParameterList &params)
+    : AtmosphereDiagnostic(comm, params) {
+  // Nothing to do here
+}
+
+void PBLEntrainmentBudget::set_grids(
+    const std::shared_ptr<const GridsManager> grids_manager) {
+  using namespace ekat::units;
+  using namespace ShortFieldTagsNames;
+
+  auto grid             = grids_manager->get_grid("Physics");
+  const auto &grid_name = grid->name();
+
+  const auto nondim = Units::nondimensional();
+
+  // Set the index map and units map
+  PBLEntrainmentBudgetDiagUtil eadu;
+  m_index_map = eadu.index_map;
+  m_units_map = eadu.units_map;
+  m_ndiag     = eadu.size;
+
+  if(eadu.pblinvalg == "temperature-inversion") {
+    m_pblinvalg = 1;
+  } else if(eadu.pblinvalg == "thetal-only") {
+    m_pblinvalg = 2;
+  } else if(eadu.pblinvalg == "qt_only") {
+    m_pblinvalg = 3;
+  } else {
+    EKAT_ERROR_MSG(
+        "Error! Invalid pblinvalg. Only temperature-inversion, thetal-only, "
+        "and "
+        "qt_only are currently supported.\n");
+  }
+
+  // Ensure m_index_map and m_units_map match
+  EKAT_REQUIRE_MSG(
+      m_index_map.size() == m_units_map.size(),
+      "Error! Some inconsistency in PBLEntrainmentBudget: index and units "
+      "maps do not match!\n");
+  // Ensure m_index_map and m_ndiag match
+  EKAT_REQUIRE_MSG(
+      static_cast<int>(m_index_map.size()) == m_ndiag,
+      "Error! Some inconsistency in PBLEntrainmentBudget: m_ndiag and index "
+      "map do not match!\n");
+
+  m_ncols = grid->get_num_local_dofs();
+  m_nlevs = grid->get_num_vertical_levels();
+
+  // Define layouts we need (both inputs and outputs)
+  FieldLayout scalar2d_layout{{COL, LEV}, {m_ncols, m_nlevs}};
+  FieldLayout vector1d_layout{{COL, CMP}, {m_ncols, m_ndiag}};
+
+  // The fields required for this diagnostic to be computed
+  // Get qc and qv
+  add_field<Required>("qc", scalar2d_layout, kg / kg, grid_name);
+  add_field<Required>("qv", scalar2d_layout, kg / kg, grid_name);
+  // Get T_mid, p_mid
+  add_field<Required>("T_mid", scalar2d_layout, K, grid_name);
+  add_field<Required>("p_mid", scalar2d_layout, Pa, grid_name);
+  // Get pseudo_density
+  add_field<Required>("pseudo_density", scalar2d_layout, Pa, grid_name);
+  // Get radiation up and down terms
+  add_field<Required>("SW_flux_dn", scalar2d_layout, W / m * m, grid_name);
+  add_field<Required>("SW_flux_up", scalar2d_layout, W / m * m, grid_name);
+  add_field<Required>("LW_flux_dn", scalar2d_layout, W / m * m, grid_name);
+  add_field<Required>("LW_flux_up", scalar2d_layout, W / m * m, grid_name);
+
+  // Construct and allocate the output field
+  FieldIdentifier fid("PBLEntrainmentBudget", vector1d_layout, nondim,
+                      grid_name);
+  m_diagnostic_output = Field(fid);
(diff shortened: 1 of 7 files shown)


bartgol:
@mahf708 how urgent is it to have this diag in master? It seems to me that this is yet another case (along with water/number path, aodvis, precip_surf_mass_flux, etc) where supporting multiple-output diagnostics would be quite beneficial. Given the impl, I would say it may be even more beneficial, since it's actually computing all diags and stuffing them in a single field, which may be confusing.

So, if having this diag is not SUPER urgent, I would encourage you to work on the multi-diags support first (assuming you're still interested in working on it, as you previously said).

rljacob:
BFB or non-BFB for eamxx?

components/eamxx/src/diagnostics/pbl_entrainment_budget.cpp
```diff
@@ -0,0 +1,332 @@
...
+  EKAT_REQUIRE_MSG(
+      static_cast<int>(m_index_map.size()) == m_ndiag,
+      "Error! Some inconsistency in PBLEntrainmentBudget: m_ndiag and index "
+      "map do not match!\n");
+
+  m_ncols = grid->get_num_local_dofs();
+  m_nlevs = grid->get_num_vertical_levels();
+
+  // Define layouts we need (both inputs and outputs)
+  FieldLayout scalar2d_layout{{COL, LEV}, {m_ncols, m_nlevs}};
```
bartgol, line 60: It doesn't change much, but I like to rely on the grid for the layouts, since it hides some implementation detail about how they are created:
```c++
auto scalar2d_layout = grid->get_2d_scalar_layout();
auto vector2d_layout = grid->get_2d_vector_layout(m_ndiag);
```

components/eamxx/src/diagnostics/pbl_entrainment_budget.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+void PBLEntrainmentBudget::initialize_impl(const RunType /*run_type*/) {
+  // Field qt will have units and layout similar to qc, qv
+  const auto &qv   = get_field_in("qv");
+  const auto &qvid = qv.get_header().get_identifier();
+  const auto &qvgn = qvid.get_grid_name();
+  const auto &qvlo = qvid.get_layout();
+  FieldIdentifier qf_prev("qtot_prev", qvlo.clone(), qvid.get_units(), qvgn);
+  m_prev_qt = Field(qf_prev);
+  m_prev_qt.allocate_view();
```
bartgol, line 103: You can just do
```c++
m_prev_qt = qv.clone("qtot_prev");
```
Similarly for `m_prev_tl`.

components/eamxx/src/diagnostics/pbl_entrainment_budget.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+  auto dt = curr_ts - m_start_t;
+
+  const int num_levs  = m_nlevs;
+  const int pblinvalg = m_pblinvalg;
+  const auto policy   = ESU::get_default_team_policy(m_ncols, m_nlevs);
+
+  constexpr int wsms = 1;
+  using WSMgr        = ekat::WorkspaceManager<Real, DefaultDevice>;
+  WSMgr wsm(num_levs, wsms, policy);
```
bartgol, line 196: This allocates memory at every evaluation. You could move the creation of the wsm to `initialize_impl`.

components/eamxx/src/diagnostics/pbl_entrainment_budget.cpp
```diff
@@ -0,0 +1,332 @@
...
+        auto prev_tliq_icol = ekat::subview(prev_tliq_v, icol);
+
+        auto ws = wsm.get_workspace(team);
+        ekat::Unmanaged<WSMgr::view_1d<Real>> tm_grad;
+        ws.take_many_contiguous_unsafe<wsms>({"tm_grad"}, {&tm_grad});
+
+        // We first want to find the PBL inversion. There are three methods to
+        // do so. All our methods here rely on the *gradient* of state fields
+        // (for now). First, we can simply find the first level from the surface
+        // that has a a temperature "inversion" (temperature goes positive
```
bartgol, line 226: do you mean "temperature _gradient_ inversion"? Cause a negative T would be a bit concerning...

components/eamxx/src/diagnostics/pbl_entrainment_budget.cpp
```diff
@@ -0,0 +1,332 @@
...
+                result.val = tm_grad(k);
+                result.loc = k;
+              }
+            },
+            minloc_t(minloc));
+        team.team_barrier();
+        opt_tm_grad_lev = minloc.loc;
+
+        if(opt_tm_grad_lev < 2 || opt_tm_grad_lev > num_levs - 1) {
+          // Weird stuff can happen near the top and bottom of atm, so fill_val
```
bartgol, line 274: This can also happen if the ||reduce above finds NO suitable levels. In that case, minloc will remain with the initial values, which is an identify for the min operation. As one would expect, minloc.val would contain a large number (`DBL_MAX`). Luckily, minloc.loc is also initialized to a large number (`INT_MAX`), so the check _would_ indeed work.

Still, I would consider adding a comment specifying we hit this branch if NO level is >70kPa.

components/eamxx/src/diagnostics/pbl_entrainment_budget.cpp
bartgol: Looking at the code below, you are looking for the minimum gradient. From your comment, I thought you wanted to find the place where the inversion _starts_. Is the comment misleading? Or is the code wrong?

components/eamxx/src/diagnostics/pbl_entrainment_budget.cpp
```diff
@@ -0,0 +1,332 @@
...
+        // Find minimum gradient, because d(p_mid) < 0 in definition above
+        // Starting from the surface, and ensuring p_mid > 70000.0 Pa,
+        // to avoid resolving to some odd place higher up in the atmosphere.
+        using minloc_t       = Kokkos::MinLoc<Real, int>;
+        using minloc_value_t = typename minloc_t::value_type;
+        minloc_value_t minloc;
+        Kokkos::parallel_reduce(
+            Kokkos::TeamVectorRange(team, 1, num_levs),
+            [&](const int &k, minloc_value_t &result) {
+              if(tm_grad(k) < result.val && pm_icol(k) > 70000.0) {
```
bartgol, line 264: You never check that `tm_grad(k)<0`. I guess this usually happens at some point in the v direction, but as a software dev with little atm modeling knowlege I wonder "what if it never occurs?"

components/eamxx/src/diagnostics/pbl_entrainment_budget.cpp
```diff
@@ -0,0 +1,332 @@
...
+          o_tl_hplus(icol) = tl_icol(opt_tm_grad_lev - 1);
+          o_qt_hplus(icol) = qt_icol(opt_tm_grad_lev - 1);
+
+          // Save the dF term (F(h) - F(0))
+          o_df_inpbl(icol) = (sd_icol(opt_tm_grad_lev - 1) - sd_icol(0)) -
+                             (su_icol(opt_tm_grad_lev - 1) - su_icol(0)) +
+                             (ld_icol(opt_tm_grad_lev - 1) - ld_icol(0)) -
+                             (lu_icol(opt_tm_grad_lev - 1) - lu_icol(0));
+
+          // Now only need to compute below from opt_tm_grad_lev to num_levs
```
bartgol, line 295: Given what discussed above, the integral now starts from the point where tm_grad is the most negative. Is that the idea? Or was it supposed to start from the point where tm_grad changes sign?

components/eamxx/src/diagnostics/tests/pbl_entrainment_budget_test.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+    const auto qc_h = qc.get_view<const Real **, Host>();
+    const auto qv_h = qv.get_view<const Real **, Host>();
+    const auto tm_h = tm.get_view<const Real **, Host>();
+    const auto pm_h = pm.get_view<const Real **, Host>();
+
+    const auto out_hf = diag->get_diagnostic();
+    auto out_hv       = out_hf.get_view<Real **, Host>();
+
+    REQUIRE(out_hv(0, 0) == Real(iop_pm[123]));
```
bartgol, line 251: The hardcoded indices are a bit obscure. Perhaps consider using some variables and add comments, so we know _why_:
```c++
// In the input iop dataset the inversion is just below level 123 (recall that lev=0 is at the top)
constexpr int expected_inv_idx = 123;
REQUIRE (out_hv(0,0) == Real(iop_pm[expected_inv_idx]));
...
for (int ilev = expected_inv_idx+1; ilev<nlevs; ++ilev) {
  ...
```

bartgol (commented):
I have some questions, mostly b/c I don't know what the PBL is and had to google most of this stuff :)

PR #6924, opened 2025-01-21
Update MALI static mode

matthewhoffman:
MALI static mode is used as a data mode where MALI is an active component but its state does not change.  This commit updates data mode to ensure that additional physics packages are disabled: 
* thermal solver
* ice-shelf basal melting
* facemelting
* subglacial hydrology

These physics packages were previously defaulted to off or are still defaulted to off in E3SM, so these changes ensure that after (or if) those defaults are changed that the MALI state will remain unchanged in "static" mode.


matthewhoffman:
Replace by https://github.com/E3SM-Project/E3SM/pull/6945

PR #6928, opened 2025-01-22
Small upgrades to eamxx docs

bartgol:
Fix a few stale sentences, and some formatting.

[bfb]

### DIFF
--- a/components/eamxx/docs/user/model_output.md
+++ b/components/eamxx/docs/user/model_output.md
@@ -11,7 +11,7 @@ in RUNDIR/data, since upon `buildnml` execution, all the CIME vars will no longe
 YAML file (replaced by their values), making it harder to tweak it, and even harder to share with
 other users/cases. Another consequence of this is that, much like `scream_input.yaml', the user should
 not modify the generated YAML files in RUNDIR/data, since any modification will be lost on the next
-run of `buildnml` (e.g., during `case.submit').
+run of `buildnml` (e.g., during `case.submit`).
 
 ## Basic output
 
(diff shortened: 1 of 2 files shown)


rljacob:
What is markdown-katex  for?

bartgol:
> > What is markdown-katex for?
> 
> likely equations
> 
> But @bartgol, does it offer any noticeable improvement over whatever is baked in by default (if any)?

What is already baked in? I tried simple tex syntax and did not work until I added the katex plugin. Maybe I was using the wrong syntax? I'm happy with whatever pkg, so long as it displays standard tex math stuff.

rljacob:
See https://raw.githubusercontent.com/E3SM-Project/E3SM/refs/heads/master/components/elm/docs/tech-guide/top_solar_parameterization.md   for a page with lots of equations.

bartgol:
> @bartgol these are the ones currently in use (i.e., math jax)
> 
> ```yaml
> extra_javascript:
>   - javascript/mathjax.js
>   - https://polyfill.io/v3/polyfill.min.js?features=es6
>   - https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js
> ```
> 
> I think one may need to use special syntax (like double `$$` or something)

Ok, let me try that.

bartgol:
@mahf708 turns out the basic tex syntax _was_ working all along. It only took some time for the browser to resolve it. Initially, it shows stuff such as `\(\Theta = T / \Pi\)` instead of the actual compiled tex, but after a few seconds, the tex gets correctly displayed. I will remove the katex pkg addition.

components/eamxx/docs/user/model_output.md
```diff
@@ -73,49 +78,57 @@ I/O interface of EAMxx. There are two types of diagnostic outputs:
...
+  - `LiqPotentialTemperature`: $\Theta_\ell = \Theta - \dfrac{\Theta}{T} \dfrac{L_v}{c_p}q_c$
+  - `dz`: $dz = \dfrac{rd}{g}\dfrac{\rho_s T_v}{p}$
+  - `z_int`: vertical sum of `dz` from surface (b.c. at surface: 0 or $\phi_s/g$, depending if from surface or sealevel)
+  - `z_mid`: midpoint average of `z_int`
+  - `geopotential_int`: vertical sum of `g dz` from surface (b.c. at surface: $\phi_s$)
+  - `geopotential_mid`: midpoint average of `geopotential_int`
+  - `AtmosphereDensity`: $\rho = \dfrac{\rho_s}{dz*g}
+  - `VirtualTemperature`: $T_v = T\left(1+c_1 q_v\right)$
+  - `DryStaticEnergy`: $DSE = c_p T + gz + \phi_s$
+  - `SeaLevelPressure`
```
bartgol, line 92: The explanation of SeaLevelPressure is insanely long. There is a whole tex file buried in our docs folder explaining the formula.

I am having second thoughts about adding formulas for all the diagnostics. Should I continue (skipping p_sl and any other with too long formulas)? Or stop, and remove all these formulas?

bartgol: The standardname should now be part of the variables metadata. But yeah, one needs to know what the heck is the eamxx name for `atmosphere_mass_content_of_water_vapor` in order to even start requesting it in an output stream...

PR #6930, opened 2025-01-22
Revert "EAMxx: Adds a flag to control vertical mixing of interstitial aerosols during aerosol activation process"

bartgol:
Reverts E3SM-Project/E3SM#6926

I failed to realize that some unit tests in MAM4xx were failing.


bartgol:
@mahf708 I ran the tests real quick, and the issue is just that the new parameter is missing from all input.yaml files in the unit tests. I may actually just fix those instead of reverting the PR, what do you think?

bartgol:
Found the issue, will issue a ninja fix PR instead.

PR #6932, opened 2025-01-22
EAMxx: add new required param to mam4_aci tests input files

bartgol:
Fixes mam4_aci unit tests that failed after a too hastily merging of #6926

---


rljacob:
What @mahf708 said is right and its even in our development guide that a branch should do one thing.  It is up to the Eamxx integrators @bartgol and @tcclevenger to enforce that.  Feel free to reject PRs that do more then one thing.

bartgol:
This time I verified that the ONLY unit tests failing are the `XYX_baseline_cmp` ones. So I think this can be merged.

PR #6933, opened 2025-01-22
EAMxx: fixes and minor upgrades to COSP

bartgol:
In COSP, fix a bug related to timestamp handling, and do some upgrades/cleanup.

[non-BFB] only for cosp outputs

---

In more detail:

- `timestamp()` returns the _start of step_ timestamp, while cosp needs the _end of step_ one (perhaps we should make two methods, with clearer names)
- pre-allocate z_mid/z_int, rather than allocating at every step
- compute z_mid on device
- sync inputs to host only if this is a cosp step
- compute z_mid only if this is a cosp step
- do not fill output host views with 0 on non-cosp steps. Just fill the dev view
- do not hard-code value of `g`, and use the one from `Constants` instead.
- change cosp freq parameters in unit test, so that it only runs every N steps.

I verified that, with these mods, the `cosp_standalone` test produces an output file where cosp fields are non-zero when you expect them. Namely, we do 5 steps, with cosp freq being also 5 steps, and the only non-zero output is at the last step. Before, the nonzero output was at the 1st step (not t=0), b/c of how the timestamp was handled.

### DIFF
--- a/components/eamxx/src/physics/cosp/eamxx_cosp.cpp
+++ b/components/eamxx/src/physics/cosp/eamxx_cosp.cpp
@@ -131,7 +131,8 @@ void Cosp::run_impl (const double dt)
   // Make sure cosp frequency is multiple of rad frequency?
 
   // Compare frequency in steps with current timestep
-  auto ts = timestamp();
+  // NOTE: timestamp() returns the time RIGHT BEFORE the call to run
+  auto end_of_step = timestamp()+dt;
   auto update_cosp = cosp_do(cosp_freq_in_steps, ts.get_num_steps());
 
   // Get fields from field manager; note that we get host views because this
--- a/components/eamxx/src/physics/cosp/eamxx_cosp.cpp
+++ b/components/eamxx/src/physics/cosp/eamxx_cosp.cpp
@@ -1,6 +1,7 @@
 #include "eamxx_cosp.hpp"
 #include "cosp_functions.hpp"
 #include "share/property_checks/field_within_interval_check.hpp"
+#include "physics/share/physics_constants.hpp"
 
 #include "ekat/ekat_assert.hpp"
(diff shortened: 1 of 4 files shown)


components/eamxx/src/physics/cosp/eamxx_cosp.cpp
```diff
@@ -131,87 +138,101 @@ void Cosp::run_impl (const double dt)
...
+        // 2. Compute z_int (vertical scan)
+        PF::calculate_z_int(team,nlev,dz_s,z_surf,z_int_s);
+        team.team_barrier();
+
+        // 3. Compute z_mid (int->mid interpolation)
+        PF::calculate_z_mid(team,nlev,z_int_s,z_mid_s);
+        team.team_barrier();
+    });
+
+    const auto T_mid_h = get_field_in("T_mid").get_view<const Real**, Host>();
```
brhillman, line 208: These are all host views, but only some of them get the `_h` suffix for host. Would it be better to rename all of them, or default to no suffix for host views since most of these are host?

brhillman (approved):
This looks great, thanks for the improvements. One minor comment about the fields that were renamed, but not a big deal either way.

components/eamxx/src/physics/cosp/eamxx_cosp.cpp
bartgol: Yeah, I thought about it, then decided to leave as is. But you're right, to avoid confusion, since we have some host and some dev, it's best to be consistent. I will add `_h` to all host views.

brhillman (approved):

bartgol:
The failures all come from `cosp_standalone_baseline_cmp`, which is expected to fail. @brhillman unless you have any final remark, I will merge.

PR #6934, opened 2025-01-22
EAMxx: fix bug in SeaLevelPressure diagnostic

bartgol:
The bug was a one-line fix: we were using the wrong index for the midpoint level near the surface. But while at it, I decided to clean up a bit the diagnostic.

[non-bfb only for PSL output]

---

The clean up is mainly getting rid of Pack in the diagnostic. In general, diagnostics should NOT use packs, since they can't affect the allocation of the field (the field is already allocated by the time the diagnostics request it in IO). In this case, using packs makes even less sense, since we only need data at ONE specific level.

[BFB] except for the SeaLevelPressure diagnostic

This should fix the FAIL we saw on pm-cpu. Well, I should probably rephrase: it should make us go _past_ the current point of failure... ;-)

Note: the code WAS buggy, but ONLY if the pack size was larger than 1. That means that all GPU runs would be unaffected by it. OTOH, I fail to understand why our CPU runs have NOT been failing all along....

### DIFF
--- a/components/eamxx/src/diagnostics/sea_level_pressure.cpp
+++ b/components/eamxx/src/diagnostics/sea_level_pressure.cpp
@@ -1,4 +1,5 @@
 #include "diagnostics/sea_level_pressure.hpp"
+#include "share/util/scream_common_physics_functions.hpp"
 
 namespace scream
 {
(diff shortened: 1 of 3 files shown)


bartgol:
Te v1 fail is due to the IcePack update PR, which made MPASSI non-bfb. I am regenerating baselines now. The eamxx-sa fails on cpu are expected. However, I was expecting the same fails for eamxx-sa on cuda. I am not comfortable merging until I understand why they passed.

tcclevenger (approved):

PR #6935, opened 2025-01-23
EAMxx: allow multi-output diags

mahf708:
currently, our AtmoshereDiagnostic class allows for only one field to be output by design; this was the intended use when we first started supporting these diagnostics. For example, currently, a diagnostic can be seen as `output1 = function ( { input1, input2, ... } )`. This PR allows additional outputs, for example `{ output1, output2, ...} = = function ( { input1, input2, ... } )`.

more design details to follow...

[bfb]

### DIFF
--- a/components/eamxx/src/diagnostics/number_path.cpp
+++ b/components/eamxx/src/diagnostics/number_path.cpp
@@ -66,10 +66,9 @@ void NumberPathDiagnostic::set_grids(
 
   // Construct and allocate the diagnostic field
   for(const auto &kind : m_kinds) {
-    FieldIdentifier fid(kind + name(), scalar2d, kg / (kg * m2), grid_name);
-    m_diagnostic_output = Field(fid);
-    m_diagnostic_output.allocate_view();
-    m_diagnostic_fields[kind + name()] = m_diagnostic_output;
+    m_diagnostic_fields[kind + name()] = Field(
+        FieldIdentifier(kind + name(), scalar2d, kg / (kg * m2), grid_name));
+    m_diagnostic_fields[kind + name()].allocate_view();
   }
 }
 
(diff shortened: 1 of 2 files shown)


bartgol (commented):
This is a good start, but of course, there's more to do. I would add more beef before merging, as it doesn't really add anything (nor does much that we can test) until then.

Here's what I think should be added:

- in IO, we need to build diags smartly. Right now, we build one diag object for each output field, but if one obj can compute 2+, we should not build 2+ copies of the same diag. This requires IO to do 2 passes when building diags: the first pass to figure out which diags to build, along with all the input params, and the second to actually build the diags)
- I would convert one of the multi-purpose diags to allow computing N fields. E.g., the number path or water path diags, since they are very simple. Then, we can test that they work fine when computing 1 or 2+ fields.
- Diag like NumberPath will now have to change their param list organization. E.g., that diag now checks the param `Number Kind`, to figure out which one to compute. Instead, it should now have 3 inputs, and check them separately. E.g., it should check something like `Compute Liq`, `Compute Ice`, and `Compute Rain`.
- We currently use `name()` from the diag as a proxy for the diagnostic output name. With 2+ fields, this is no longer the case, so we must ensure that we DON'T use that string to refer to the field. Instead, the `name()` method should refer to the _class_ itself.

For the first point, I envision something like this:
```c++
map<string,ParameterList> params;
for (f in diag_fields) {
  diag_factory_key = find_key_for(f);  // this is the key to pass to the diag factory.
  params[diag_factory_key].set(...);
}

for (f in diag_field) {
  diag_factory_key = find_key_for(f);
  create_diag(diag_factory_key,params[diag_factory_key])
}
```
I didn't do a great job explaining, but basically, the 1st loop is neede b/c different fields will set multiple params _for the same diag class_ (e.g., the `Compute Liq`/`Compute Ice`/`Compute Rain` mentioned above). Only after ALL params have been set, we can proceed to create the diags.

bartgol:
> > * We currently use `name()` from the diag as a proxy for the diagnostic output name. With 2+ fields, this is no longer the case, so we must ensure that we DON'T use that string to refer to the field. Instead, the `name()` method should refer to the _class_ itself.
> 
> I agree with everything, but I am not sure about this one. Let me try to think about it. My goal is to keep edits as noninvasive as possible. However, this part is more or less necessary if we want to compute diags smartly. Let me think more carefully

The single-output diags can still use name() internally when creating the diag, I wasn't hinting at that. I just want to make sure that from the _outside_ we are not using `AtmosphereDiagnostic::name()` and the actual diag field name as they were the same thing.

bartgol (dismissed):

tcclevenger:
@bartgol I'm confused why your approval doesn't allow the testing to move forward

PR #6937, opened 2025-01-23
Add the dust emission cap (and its switch) as a tunable parameter in the namelist

yunpengshan2014:
To address the issue of overly strong dust emissions causing crashes in high-resolution (HR) simulations, a dust emission cap has been implemented. The cap ensures that dust emissions remain within a reasonable range to maintain simulation stability.

Additionally, the dust emission cap and its enabling switch have been added as configurable namelist items, allowing users to easily activate and adjust the cap as needed for their specific simulation setups.

Output the lat-lon coordinate of overly high dust emission cases.


yfenganl:
you might want to consider applying the dust emission cap (10^-4 kg/m2/s) to the total emission flux (sum of a1 and a3 size modes), because the instantaneous emission flux in a3 is much larger than that in a1, e.g., 10 times. The current implementation thus allows larger total emission fluxes than the emission cap defined.

PR #6885, opened 2025-01-09
Wave Fields are CONDITIONALLY coupled based on COMPSET

erinethomas:
There is a issue in E3SM where the coupler history files contain wave variables **even in compsets with NO active wave component**. This PR fixes the issue by making all coupled wave fields "conditional" on the existence of an active wave component. 
This PR produces B4B results in standard B cases with no active waves apart from the field list   

This PR adds a drv_in namelist variable called "wav_ocn_coup". This variable controls the coupling between the wave and ocean components. "wav_ocn_coup" is automatically defined (no user input is required) with one of the following: "none", "one" or "two" - based on the active components in the selected COMPSET. 

For example, a B case with active WW3 automatically sets the wav_ocn_coup to "two" (full two-way coupling enabled between waves and ocean). However, a standard B case with no WW3 sets "wav_ocn_coup to "none" where no wave-ocean coupling is defined in the model. 

Note: This framework is flexible enough to allow future capabilities of wave coupling. For example, in compsets where a data model is used, wav_ocn_coup could be set to "one", to enable one way coupling (passing information from a data model to active wave component). Furthermore, this framework will be expanded upon in the future to allow one and two way coupling between wave-atmosphere, as well as wave-sea ice components.

PR #6943, opened 2025-01-24
Wave Momentum Coupling

erinethomas:
This PR introduces a new feature to E3SMv3: Wave Momentum coupling between WW3-MPAO-EAM. This is a "stealth" feature since wave-enabled compsets are not part of standard production runs. This PR is BFB in non-wave enabled compsets, however, in wave-enabled compsets, this PR will impact the results. This PR is working towards developing a fully coupled climate "with waves" - the "LRW" configuration. Note: This feature will be used in some of the Blue-Tip simulations. 

[FCC] for wave-enabled compsets when turned on

---

This new Feature is fully documented in a Design Doc: 
https://acme-climate.atlassian.net/wiki/spaces/OO/pages/4902158337/Design+document+for+Wave-Momentum+Coupling

The 30+ yr PI control run with MPAS_Analysis and E3SM_Diags is documented in the simulation page: 
https://acme-climate.atlassian.net/wiki/spaces/OO/pages/4901208068/Wave+Momentum+Coupling+E3SMv3+PI+Control+Run

Note: This PR must be merged AFTER https://github.com/E3SM-Project/E3SM/pull/6885


erinethomas:
The results are BFB for compsets when waves are turned off. For wave-enabled compsets, by default, this feature will be ON. This feature **does** change the results from the current status of wave coupling, however, the current status of wave coupling in E3SM is not complete, therefore we **do not yet have an estimate of the fully-coupled-climate with waves**. Therefore, I would _not_ say this PR is "climate-changing" since we have no baseline of "climate with waves" against which to compare. This PR is working towards that goal of obtaining a "climate with waves".
Based on this logic, I have removed the CC tag.

proteanplanet (changes requested):
The COARE algorithm needs a thorough review, especially these lines: E3SM in https://github.com/E3SM-Project/E3SM/blob/967697317b6decccd40664026fa9aaec4355a6ca/share/util/shr_flux_mod.F90#L528-L529.  which should be analogous to https://github.com/E3SM-Project/E3SM/blob/967697317b6decccd40664026fa9aaec4355a6ca/share/util/shr_flux_mod.F90#L338-L343 but is not. 

We attempted to use COARE in v2 with significant problems.  Gustiness may have been the cause, now corrected for `(ocn_surface_flux_scheme .eq. 0` but not `(ocn_surface_flux_scheme .eq. 1`.  So as part of this PR, we also need a simulation illustrating that E3SM PI is not significantly different, and conserves, with COARE switched on without waves.

In addition, it needs to be explicitly shown that there is not sea level rise with wave switched on, so the B-case linked above needs to be stopped, and a new one included with the correct land initial conditions.

erinethomas:
@wlin7 mentioned a comparison of the flux schemes (Large and Yeager 2009 vs COARE) was done as part of the V3 release. @wlin7 - could you point us to the results from that comparison?

proteanplanet:
There is another issue.  Even when COARE is corrected to include gustiness, that flux scheme is not consistent with the flux scheme over sea ice, which is part of the reason why `ocn_surface_flux_scheme .eq. 0` continues to be used.  U. Arizona were going to generate a COARE-equivalent scheme for sea ice, but I'm not sure where that work is at.  This is a problem specifically for the wave-sea ice interaction work that is the initial focus of the capability introduced by this PR.  Is there no way of including the wave terms in the default V3 scheme?

erinethomas:
No, there is no way to couple waves into the Large and Yeager 2009 (the default v3) scheme.

erinethomas:
> The COARE algorithm needs a thorough review, especially these lines
> 
> https://github.com/E3SM-Project/E3SM/blob/967697317b6decccd40664026fa9aaec4355a6ca/share/util/shr_flux_mod.F90#L528-L529
> 
> which should be analogous to
> https://github.com/E3SM-Project/E3SM/blob/967697317b6decccd40664026fa9aaec4355a6ca/share/util/shr_flux_mod.F90#L338-L343
> 
> but is not.
> We attempted to use COARE in v2 with significant problems. Gustiness may have been the cause, now corrected for `(ocn_surface_flux_scheme .eq. 0` but not `(ocn_surface_flux_scheme .eq. 1`. So as part of this PR, we also need a simulation illustrating that E3SM PI is not significantly different, and conserves, with COARE switched on without waves.
> 
> In addition, it needs to be explicitly shown that there is no sea level rise with waves switched on, so the B-case linked above needs to be stopped, and a new one included with the correct land initial conditions.

The COARE algorithm already includes a gustiness factor within the cor30a subroutine: 
https://github.com/E3SM-Project/E3SM/blob/c199995df491858fc6becf8600181dd8e1242656/share/util/shr_flux_mod.F90#L2683-L2690

proteanplanet:
I don't think that's the correct gustiness treatment. ugust comes from the atmosphere model, whereas I think what you are pointing to is in-built.

wlin7:
> . @wlin7 - what was included in this comparison and could you point us to the results?
Hi @erinethomas , [here is the link ](https://acme-climate.atlassian.net/wiki/spaces/CM/pages/3945365505/v3alpha04_trigrid.piControl+with+alternative+surface+flux+schemes)to the coupled comparison simulations for the three schemes. The simulations ran to 50 years.

The diagnostics and the various comparisons (model vs obs, model vs model, and the multi-simulation comparison view (IICE) can be browsed from [this simulation page.](https://acme-climate.atlassian.net/wiki/spaces/CM/pages/3945365517/20230929.v3alpha04-UA.trigrid.piControl.chrysalis#E3SM-Diags)

proteanplanet:
@erinethomas and I determined today that the coupling of Wavewatch III to the ocean - atmosphere flux scheme is seriously flawed no matter what flux scheme is used.   There is no feedback from wave height to surface fluxes that we can decipher.  Please comment here in the next few days if you think that's not correct, otherwise we'll need to close this PR and move it to Ocean Discussions. We are planning to fix this problem and integrate wave height coupling into the default ocean - atmosphere flux scheme before restarting this PR.

erinethomas:
> @erinethomas and I determined today that the coupling of Wavewatch III to the ocean - atmosphere flux scheme is seriously flawed no matter what flux scheme is used. There is no feedback from wave height to surface fluxes that we can decipher. Please comment here in the next few days if you think that's not correct, otherwise we'll need to close this PR and move it to Ocean Discussions. We are planning to fix this problem and integrate wave height coupling into the default ocean - atmosphere flux scheme before restarting this PR.

I think this is correct way forward for a major fix like this.

PR #6950, opened 2025-01-27
Fix issue with mpas-seaice for branch and hybrid runs

### LINKED ISSUE #6925: Sea ice initial condition in hybrid/branch G-case reverts to disk
When setting up a hybrid/branch G-case, MPAS-seaice reverts to using the cold start initial condition, a 1m thick disk of ice. The `restart_ic` stream in `streams.seaice` is not populated in this configuration, and needs to be added manually with the path to the intended branch point sea ice restart file.

The logic to populate the `restart_ic` stream is here:
https://github.com/E3SM-Project/E3SM/blob/9f134fdf1ab578fe1474be1d29a2f6bf764ec676/components/mpas-seaice/cime_config/buildnml#L525

and since `ice_ic_mode` /neq "spunup" in G-cases (even in hybrid/branch cases), this doesn't get activated. 

In B-cases, compsets set `ice_ic_mode = "spunup"`, and the intended sea ice branch point is used in hybrid/branch cases.

jonbob:
There has been an issue with mpas-seaice in branch and hybrid runs when the compset does not include SOI, indicating spunup ocn/ice initial conditions. In that situation, which mostly occurs for D- and G-compsets, the seaice would not correctly set config_initial_condition_type and add a restart_ic stream. This fixes that issue by making ice_ic_mode equal to spunup when run_type is branch or hybrid.

Fixes #6925

### DIFF
--- a/components/mpas-seaice/cime_config/buildnml
+++ b/components/mpas-seaice/cime_config/buildnml
@@ -61,6 +61,8 @@ def buildnml(case, caseroot, compname):
         ice_ic_mode = 'spunup'
     if ice_bgc == 'ice_bgc':
         ice_ic_mode = 'spunup'
+    if run_type == 'hybrid' or run_type == 'branch':
+        ice_ic_mode = 'spunup'
 
     #--------------------------------------------------------------------
     # Determine date stamp, from grid names


jonbob:
Tested by running a TL319_IcoswISC30E3r5 GMPAS-JRA1p5 configuration from startup for two months, dropping restart files every month. A second case branched from the restart files at 0001-02-01 and ran for another month. The cpl history files were BFB and the mpassi_in file had 
````
config_initial_condition_type = 'restart'
````
and the streams.seaice file had a correct restart_ic stream

jonbob:
Second test with a hybrid run also worked as expected

darincomeau:
I built a branch and hybrid case with this and confirmed the correct `restart_ic` stream was added and `config_initial_condition_type = 'restart'` in the namelist file.

I used this for the G-case branch run I had been trying to run for https://github.com/E3SM-Project/E3SM/pull/6802, and can see the intended sea ice branch state is being used for the initial condition in the hybrid run:

image

darincomeau (approved):
Approved based on inspection and testing.

jonbob:
Passes:
* e3sm_ice_developer on chrysalis

merged to next

jonbob:
merged to master

PR #6954, opened 2025-01-29
Fix SCREAM_LIB_ONLY typo in p3 and shoc

brhillman:
Fix `SCREAM_LIB_ONLY` typo in p3 and shoc. This was misspelled `SCREAM_LIBS_ONLY` in the CMakeLists.txt, which would also evaluate to FALSE in a cmake logical, which in this case meant that the `disp` sources for p3 and shoc would always be built for non-standalone (cime) builds of p3 and shoc.

### DIFF
--- a/components/eam/src/physics/crm/pam/CMakeLists.txt
+++ b/components/eam/src/physics/crm/pam/CMakeLists.txt
@@ -11,7 +11,7 @@ set(PAM_DRIVER_SRC
 add_library(pam_driver 
   ${PAM_DRIVER_SRC})
 
-set(SCREAM_LIBS_ONLY TRUE)
+set(SCREAM_LIB_ONLY TRUE)
 set(SCREAM_HOME ${CMAKE_CURRENT_SOURCE_DIR}/../../../../../..)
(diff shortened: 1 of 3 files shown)


brhillman:
@mahf708 just pointed out that this typo exists in PAM too. @jgfouca this might be relevant to you, I think you recently added that line in the PAM cmake.

bartgol (approved):
Thanks!

jgfouca (approved):
Nice catch!

jgfouca (changes requested):
Since components/eam/src/physics/crm/pam/CMakeLists.txt is part of the main repo, can you fix it there as well?

brhillman:
Sorry @jgfouca I missed your review earlier. I just pushed a commit to fix the typo in the pam cmake.

jgfouca (approved):

bartgol:
Since this PR technically does touch stuff outside of EAMxx, I will wait for the nightly testing.

bartgol:
Our CI tests passed, and the MMF test on chrysalis built for next last night, so this is good to be merged.

PR #6959, opened 2025-01-30
Upgrade ADIOS2 libs with compression support on Frontier

dqwu:
This PR upgrades ADIOS libs to version 2.10.2 on Frontier with
compression support enabled.

[BFB]


dqwu:
@grnydawn I have tested the new ADIOS libraries on Frontier. You can double-check that this change does not cause any build errors. Thanks.

grnydawn (approved):
Approved. Tests have passed with all the compilers on Frontier except crayclanggpu, which looks to be an issue other than this Adios update.

jayeshkrishna (approved):

PR #6960, opened 2025-01-30
EAMxx: add field utility to scale with inverse of input field

bartgol:
Similar to the existing `Field::scale`. It helps making unit tests (or non-perf critical code) easier to read.

[B4B]

### DIFF
--- a/components/eamxx/src/share/field/field_impl.hpp
+++ b/components/eamxx/src/share/field/field_impl.hpp
@@ -695,6 +695,34 @@ scale (const ST beta)
 
 template<HostOrDevice HD>
 void Field::
+scale_inv (const Field& x)
+{
+  const auto& dt = data_type();
+  if (dt==DataType::IntType) {
+    int fill_val = constants::DefaultFillValue<int>().value;
+    if (get_header().has_extra_data("mask_value")) {
+      fill_val = get_header().get_extra_data<int>("mask_value");
+    }
+    return update_impl<CombineMode::Divide,HD,int>(x,0,0,fill_val);
+  } else if (dt==DataType::FloatType) {
+    float fill_val = constants::DefaultFillValue<float>().value;
+    if (get_header().has_extra_data("mask_value")) {
+      fill_val = get_header().get_extra_data<float>("mask_value");
+    }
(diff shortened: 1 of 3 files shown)


components/eamxx/src/share/field/field_impl.hpp
```diff
@@ -693,6 +693,34 @@ scale (const ST beta)
...
+    if (get_header().has_extra_data("mask_value")) {
+      fill_val = get_header().get_extra_data<int>("mask_value");
+    }
+    return update_impl<CombineMode::Multiply,HD,int>(x,0,0,fill_val);
+  } else if (dt==DataType::FloatType) {
+    float fill_val = constants::DefaultFillValue<float>().value;
+    if (get_header().has_extra_data("mask_value")) {
+      fill_val = get_header().get_extra_data<float>("mask_value");
+    }
+    return update_impl<CombineMode::Multiply,HD,float>(x,0,0,fill_val);
```
AaronDonahue, line 712: Was this and the above meant to have CombineMode::Divide, like the Double example below?

bartgol: Ah rats, I missed those. Thanks.

components/eamxx/src/share/tests/field_tests.cpp
```diff
@@ -847,6 +847,18 @@ TEST_CASE ("update") {
     f2.scale(2.0);
     REQUIRE (views_are_equal(f1, f2));
   }
+
+  SECTION ("scale_inv") {
```
AaronDonahue, line 851: I guess we could add a test for type = int using 4 and 2 or something like that, which would have caught my previous comment.

AaronDonahue (approved):
Looks good to me, had a comment about adding another unit test for `int` that you are welcome to add.

AaronDonahue (approved):

bartgol:
I forced a rebase on master due to conflicts with next. I will merge to next/master as soon as CI passes, without waiting for aaron to review again.

AaronDonahue:
@bartgol , is this good to go in now?

PR #6964, opened 2025-02-01
ZM cleanup - Remove buoyan() subroutine

whannah1:
This is a continuation of the ZM overhaul and refactoring, which is limited to cosmetic cleanup of the main ZM routines. In this iteration we remove the unused subroutine buoyan(), which is the original undilute parcel and CAPE calculations that were abandoned in the CAM4 era.

[BFB]

### DIFF
--- a/components/eam/src/physics/cam/zm_conv.F90
+++ b/components/eam/src/physics/cam/zm_conv.F90
@@ -349,7 +349,6 @@ subroutine zm_convr(lchnk   ,ncol    , &
 ! and will make use of the standard CAM nomenclature
 ! 
 !-----------------------------------------------------------------------
-   use phys_control, only: cam_physpkg_is
    use time_manager, only: is_first_step 
 !
 ! ************************ index of variables **********************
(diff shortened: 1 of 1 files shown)


components/eam/src/physics/cam/zm_conv.F90
```diff
@@ -796,53 +795,36 @@ subroutine zm_convr(lchnk   ,ncol    , &
...
+   ! DCAPE is the difference in CAPE between the two calls using the same launch level
 
-      !  For cam3 physics package, call non-dilute
+   iclosure = .true.
+   call buoyan_dilute(lchnk   ,ncol    ,                             &
+                      q       ,t       ,p       ,z       ,pf       , &
+                      tp      ,qstp    ,tl      ,rl      ,cape     , &
+                      pblt    ,lcl     ,lel     ,lon     ,maxi     , &
+                      rgas    ,grav    ,cpres   ,msg               , &
+                      tpert   ,iclosure                            )
```
singhbalwinder, line 809: I find it useful to keep info about intents (in, out, inouts) from a readability perspective. It allows us to find out which quantities are being modified and reason about whether they should be modified or not. If we want to avoid verbosity, we can collect all the intent-ins at the beginning, then inouts, and then intent-outs. This way, we will be adding comments in only three places:
```
call sub( !ins
....
!inouts
....
!outs
....)
```
The majority of the calls may have just ins and outs. You can add this info any way you like.

whannah1:
The following tests pass on Chrysalis:
```
  e3sm_atm_developer_intel
  e3sm_atm_developer_gnu
  SMS_Ld32.ne30pg2_r05_oECv3.F2010.chrysalis_intel
  SMS_Ld32.ne30pg2_r05_oECv3.F2010.chrysalis_gnu
  SMS_Ld32.ne4pg2_oQU480.F2010.chrysalis_intel
  SMS_Ld32.ne4pg2_oQU480.F2010.chrysalis_gnu
```

The performance of the SMS_Ld32 tests are indentical, here's on of the ZM timers from `SMS_Ld32.ne30pg2_r05_oECv3.F2010.chrysalis_intel`:
```
              name                                            on  processes  threads        count      walltotal   wallmax (proc   thrd  )   wallmin (proc   thrd  )
    base: "a:zm_conv_tend"                                 -        128      256 2.360832e+06   1.579500e+04    72.009 (     1      0)    51.040 (    27      0)
    test: "a:zm_conv_tend"                                 -        128      256 2.360832e+06   1.579500e+04    72.009 (     1      0)    51.040 (    27      0)
```

components/eam/src/physics/cam/zm_conv.F90
singhbalwinder: Thanks. I see both perspectives here. I agree that it needs to be maintained (we were enforcing this during the review process in the MAM4 refactor). It may just be a personal preference, but I like to see the "intents" while reading the code to quickly assess whether it makes sense for this call to change these variables or not (without the need to search for the subroutine interface, which might be in another file). It also helps while debugging to immediately know if the variable we are tracing is being modified by a subroutine call or not. That said, I will leave it up to you, I am okay either way.

crterai (approved):

PR #6966, opened 2025-02-01
EAMxx: add ability to use separate cloud fractions in p3

hassanbeydoun:
This PR is an attempt at allowing P3 to see separate liquid and ice fractions. Code changes involve passing cldfrac_liq and cldfrac_ice to P3 (instead of cldfrac_tot) and modifying the WBF process based on work done by Lin Lin in THREAD.

[BFB]

### DIFF
--- a/components/eamxx/src/physics/p3/eamxx_p3_process_interface.cpp
+++ b/components/eamxx/src/physics/p3/eamxx_p3_process_interface.cpp
@@ -65,8 +65,12 @@ void P3Microphysics::set_grids(const std::shared_ptr<const GridsManager> grids_m
 
   // These variables are needed by the interface, but not actually passed to p3_main.
   add_field<Required>("cldfrac_tot", scalar3d_layout_mid, nondim, grid_name, ps);
+  if (runtime_options.use_separate_ice_liq_frac) {
+    add_field<Required>("cldfrac_liq", scalar3d_layout_mid, nondim, grid_name, ps);
+    add_field<Required>("cldfrac_ice", scalar3d_layout_mid, nondim, grid_name, ps);
+  }
 
-//should we use one pressure only, wet/full?
+  // should we use one pressure only, wet/full?
   add_field<Required>("p_mid",       scalar3d_layout_mid, Pa,     grid_name, ps);
   add_field<Required>("p_dry_mid",   scalar3d_layout_mid, Pa,     grid_name, ps);
   add_field<Updated> ("T_mid",       scalar3d_layout_mid, K,      grid_name, ps);  // T_mid is the only one of these variables that is also updated.
@@ -274,7 +278,16 @@ void P3Microphysics::initialize_impl (const RunType /* run_type */)
   const  auto& pseudo_density = get_field_in("pseudo_density").get_view<const Pack**>();
   const  auto& pseudo_density_dry = get_field_in("pseudo_density_dry").get_view<const Pack**>();
   const  auto& T_atm          = get_field_out("T_mid").get_view<Pack**>();
-  const  auto& cld_frac_t     = get_field_in("cldfrac_tot").get_view<const Pack**>();
+  const  auto& cld_frac_t_in  = get_field_in("cldfrac_tot").get_view<const Pack**>();
+  // FIXME: This is hack to get things going, these fields should be strictly const
+  // But to get around bfb testing and needing to declare these fields elsewhere,
+  // We will just use this workaround ...
+         auto  cld_frac_l_in  = cld_frac_t_in;
+         auto  cld_frac_i_in  = cld_frac_t_in;
+  if (runtime_options.use_separate_ice_liq_frac) {
+               cld_frac_l_in  = get_field_in("cldfrac_liq").get_view<const Pack**>();
+               cld_frac_i_in  = get_field_in("cldfrac_ice").get_view<const Pack**>();
+  }
   const  auto& qv             = get_field_out("qv").get_view<Pack**>();
   const  auto& qc             = get_field_out("qc").get_view<Pack**>();
   const  auto& nc             = get_field_out("nc").get_view<Pack**>();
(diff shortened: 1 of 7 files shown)


components/eamxx/src/physics/p3/impl/p3_conservation_impl.hpp
```diff
@@ -50,6 +54,59 @@ ::cloud_water_conservation(const Spack& qc, const Scalar dt,
...
+   //but this is an in addition to the qidep we computed outside the mixed
+   //phase cloud. qidep*(1._rtype-ratio)*(il_cldm/(cld_frac_i-il_cldm)) is the additional
+   // vapor depositional growth rate that takes place within the mixed phase cloud
+   // after qc is depleted
+=======
+  // but this is an in addition to the qidep we computed outside the mixed
+  // phase cloud. qidep*(1._rtype-ratio)*(il_cldm/cld_frac_i) is the additional
+  // vapor depositional growth rate that takes place within the mixed phase cloud
+  // after qc is depleted
+>>>>>>> 126fee1a14 (EAMxx: correctly pass const and views around)
```
whannah1, line 72: Just pointing this out in case it's not on anyone's radar yet

components/eamxx/src/physics/p3/impl/p3_back_to_cell_average_impl.hpp
```diff
@@ -28,8 +28,15 @@ ::back_to_cell_average(
   Spack& nr2ni_immers_freeze_tend, Spack& ni_sublim_tend, Spack& qv2qi_nucleat_tend,
   Spack& ni_nucleat_tend, Spack& qc2qi_berg_tend,
   Spack& ncheti_cnt, Spack& qcheti_cnt, Spack& nicnt, Spack& qicnt, Spack& ninuc_cnt,
-  Spack& qinuc_cnt, const Smask& context)
+  Spack& qinuc_cnt, const Smask& context, const P3Runtime* runtime_options)
 {
+  bool use_separate_ice_liq_frac;
```
bartgol, line 33: Since Naser brought it up... I think this is fine, but a couple of alternatives for you in case you find this underwhelming.

1. Blow, you can be more compact and do
```c++
bool use_separate_ice_liq_frac = runtime_options and runtime_options->use_separate_ice_liq_frac;
```
Booleans are short-circuited in C++, so if runtime_options==nullptr, the 2nd part of the `and` is not evaluated (and you don't deref the null ptr).
2. You could pass runtime as a ref, and default to ` = {}`. If the struct inits all param to the same defaults you use in the tests, you're good. E.g., if it has
```c++
struct P3Runtime {
...
  bool use_separate_ice_liq_frac = false;
};
```
having `const P3Runtime runtime_options = {}` in the fcn signature will make you have `use_separate_ice_liq_frac = false` by default.

components/eamxx/src/physics/p3/eamxx_p3_process_interface.cpp
```diff
@@ -274,7 +278,16 @@ void P3Microphysics::initialize_impl (const RunType /* run_type */)
   const  auto& pseudo_density = get_field_in("pseudo_density").get_view<const Pack**>();
   const  auto& pseudo_density_dry = get_field_in("pseudo_density_dry").get_view<const Pack**>();
   const  auto& T_atm          = get_field_out("T_mid").get_view<Pack**>();
-  const  auto& cld_frac_t     = get_field_in("cldfrac_tot").get_view<const Pack**>();
+  const  auto& cld_frac_t_in  = get_field_in("cldfrac_tot").get_view<const Pack**>();
+  // FIXME: This is hack to get things going, these fields should be strictly const
+  // But to get around bfb testing and needing to declare these fields elsewhere,
+  // We will just use this workaround ...
+         auto  cld_frac_l_in  = cld_frac_t_in;
```
jgfouca, line 285: Strange indentation here.

components/eamxx/src/physics/p3/impl/p3_conservation_impl.hpp
```diff
@@ -7,6 +7,66 @@
...
+  Smask enforce_conservation  = sinks > sources && sinks >= qtendsmall && context;  // determine if  conservation corrction is necessary
+  Smask nothing_todo = !enforce_conservation && context;
+
+  if (enforce_conservation.any()){
+    ratio.set(enforce_conservation, sources/sinks);
+    qc2qr_autoconv_tend.set(enforce_conservation, qc2qr_autoconv_tend*ratio);
+    qc2qr_accret_tend.set(enforce_conservation, qc2qr_accret_tend*ratio);
+    qc2qi_collect_tend.set(enforce_conservation, qc2qi_collect_tend*ratio);
+    if(use_hetfrz_classnuc){
+         qcheti_cnt.set(enforce_conservation, qcheti_cnt*ratio);
```
jgfouca, line 40: Some minor code formatting stuff. 4-space indents are being used here instead of the standard 2. Also, many of the `if` and `else` statements are not formatted like the rest of eamxx. They should be
`if<space>(<condition>)<space>{`
and
`else<space>{`

jgfouca (changes requested):
Since I don't see any changes in p3/tests, I assume no test for this second cloud_water_convservation impl was added. Should we add one?

components/eamxx/src/physics/p3/impl/p3_conservation_impl.hpp
singhbalwinder: Since this function is overloaded, it leads to code duplication. To avoid that, we may need to restructure the logic (either using helper functions with core functionality or using optional arguments). I am not sure how much effort that is or if it should be done in this PR.

singhbalwinder (approved):
Nice work! I just left a minor comment.

bartgol (approved):

jgfouca:
@mahf708 , thanks, I withdraw my comments about the formatting. We still have no unit test for the new impl. If you think that's not a concern, then I'm fine with it.

jgfouca (approved):

AaronDonahue:
Is this ready to be merged?

PR #6970, opened 2025-02-04
EAMxx: refactor field manipulation interfaces, and add new ones

bartgol:
Simplify implementation of the scale/scale_inv method, and add new ones for max/min.

[BFB]
---

First, the current implementation had several methods implemented that could just be expressed as a single call to `update`, if only `update` accepted different combine mode values (like `update_impl` did). The solution was simple: template `update` on the combine mode, defaulting to the current behavior (i.e., `ScaleUpdate`). With this modification, `scale`, `scale_impl` and also `deep_copy` (from a Field rhs) could be rewritten as a 1-liner call to `update`.

Second, the PR adds two more values for `CombineMode`, namely `Max` and `Min`. The goal is to later use Field operations inside `scorpio_output.cpp`, to avoid manually handling the same kind of stuff in that class, simplifying its implementation.

### DIFF
--- a/components/eamxx/src/share/field/field.hpp
+++ b/components/eamxx/src/share/field/field.hpp
@@ -201,11 +201,11 @@ public:
 
   // Set the field to a constant value (on host or device)
   template<typename T, HostOrDevice HD = Device>
-  void deep_copy (const T value) const;
+  void deep_copy (const T value);
 
-  // Copy the data from one field to this field
+  // Copy the data from one field to this field (recycle update method)
   template<HostOrDevice HD = Device>
-  void deep_copy (const Field& src) const;
+  void deep_copy (const Field& src) { update<HD,CombineMode::Replace>(src,1,0); }
 
   // Updates this field y as y=alpha*x+beta*y
   // NOTE: ST=void is just so we can give a default to HD,
@@ -325,10 +325,7 @@ protected:
   void sync_views_impl () const;
 
   template<HostOrDevice HD, typename ST>
-  void deep_copy_impl (const ST value) const;
-
-  template<HostOrDevice HD, typename ST>
-  void deep_copy_impl (const Field& src) const;
+  void deep_copy_impl (const ST value);
 
   // The update method calls this, with ST matching this field data type.
   // Note: use_fill is used to determine *at compile time* whether to use
@@ -342,6 +339,7 @@ protected:
   // NOTE: if neither lsh nor rhs has "mask_value" extra data, this will throw.
   template<typename T>
   static T get_mask_value (const Field& lhs, const Field& rhs) {
+    // If present, prefer using the lhs native mask value.
     if (lhs.get_header().has_extra_data("mask_value")) {
       return lhs.get_header().get_extra_data<T>("mask_value");
     } else if (rhs.get_header().has_extra_data("mask_value")) {
--- a/components/eamxx/src/share/field/field.hpp
+++ b/components/eamxx/src/share/field/field.hpp
@@ -229,6 +229,14 @@ public:
   template<HostOrDevice HD = Device>
   void scale_inv (const Field& x) { update<HD,CombineMode::Divide>(x,1,0); }
 
+  // Replace *this with max(*this, x)
+  template<HostOrDevice HD = Device>
+  void max (const Field& x) { update<HD,CombineMode::Max>(x,1,0); }
+
+  // Replace *this with min(*this, x)
+  template<HostOrDevice HD = Device>
+  void min (const Field& x) { update<HD,CombineMode::Min>(x,1,0); }
+
   // Returns a subview of this field, slicing at entry k along dimension idim
   // NOTES:
   //   - the output field stores *the same* 1d view as this field. In order
--- a/components/eamxx/src/share/field/field.hpp
+++ b/components/eamxx/src/share/field/field.hpp
@@ -213,20 +213,21 @@ public:
   // NOTE: the type ST  must be such that no narrowing happens when
   //       casting the values to whatever the data type of this field is.
   //       E.g., if data_type()=IntType, you can't pass double's.
-  template<HostOrDevice HD = Device, typename ST = void>
+  // See share/util/scream_combine_ops.hpp for more details on CombineMode options
+  template<HostOrDevice HD = Device, CombineMode CM = CombineMode::ScaleUpdate, typename ST = void>
   void update (const Field& x, const ST alpha, const ST beta);
 
-  // Special case of update with alpha=0
+  // Special case of update for particular choices of the combine mode
   template<HostOrDevice HD = Device, typename ST = void>
-  void scale (const ST beta);
+  void scale (const ST beta) { update<HD,CombineMode::Rescale>(*this,ST(0),beta); }
 
   // Scale a field y as y=y*x where x is also a field
   template<HostOrDevice HD = Device>
-  void scale (const Field& x);
+  void scale (const Field& x) { update<HD,CombineMode::Multiply>(x,1,0); }
 
   // Scale a field y as y=y/x where x is also a field
   template<HostOrDevice HD = Device>
-  void scale_inv (const Field& x);
+  void scale_inv (const Field& x) { update<HD,CombineMode::Divide>(x,1,0); }
 
   // Returns a subview of this field, slicing at entry k along dimension idim
   // NOTES:
@@ -321,10 +322,30 @@ protected:
   template<HostOrDevice HD, typename ST>
   void deep_copy_impl (const Field& src) const;
 
-  template<CombineMode CM, HostOrDevice HD, typename ST>
-  void update_impl (const Field& x, const ST alpha, const ST beta, const ST fill_val);
+  // The update method calls this, with ST matching this field data type.
+  // Note: use_fill is used to determine *at compile time* whether to use
+  // the combine<CM> utility or combine_and_fill<CM>
+  template<CombineMode CM, HostOrDevice HD, bool use_fill, typename ST>
+  void update_impl (const Field& x, const ST alpha, const ST beta, const ST fill_val = 0);
 
 protected:
+
+  // This is helpful in the update method, and it can't be a lambda due to templating
+  // NOTE: if neither lsh nor rhs has "mask_value" extra data, this will throw.
+  template<typename T>
+  static T get_mask_value (const Field& lhs, const Field& rhs) {
+    if (lhs.get_header().has_extra_data("mask_value")) {
+      return lhs.get_header().get_extra_data<T>("mask_value");
+    } else if (rhs.get_header().has_extra_data("mask_value")) {
+      return rhs.get_header().get_extra_data<T>("mask_value");
+    } else {
+      EKAT_ERROR_MSG ("Error! Neither of the fields has the 'mask_value' extra data.\n"
+          " - lhs name: " + lhs.name() + "\n"
+          " - rhs name: " + rhs.name() + "\n");
+
+      return 0; // unreachable return
+    }
+  }
 
   template<HostOrDevice HD>
   const get_view_type<char*,HD>&
--- a/components/eamxx/src/share/field/field_impl.hpp
+++ b/components/eamxx/src/share/field/field_impl.hpp
@@ -617,27 +617,15 @@ void Field::deep_copy_impl (const ST value) const {
   }
 }
 
-template<HostOrDevice HD, typename ST>
+template<HostOrDevice HD, CombineMode CM, typename ST>
 void Field::
 update (const Field& x, const ST alpha, const ST beta)
 {
   const auto& dt = data_type();
 
   // Determine if there is a FillValue that requires extra treatment.
-  ST fill_val = constants::DefaultFillValue<ST>().value;
-
-  if (x.get_header().has_extra_data("mask_value")) {
-
-    if (dt==DataType::IntType) {
-      fill_val = x.get_header().get_extra_data<int>("mask_value");
-    } else if (dt==DataType::FloatType) {
-      fill_val = x.get_header().get_extra_data<float>("mask_value");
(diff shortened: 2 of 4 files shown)


tcclevenger (approved):
int->double conversions are going to be nice!

bartgol:
@tcclevenger I added a few more improvements:

- Use MDRange instead of Range+modularArithmetics: I chatted with Jim about this (he faced this choice in RRTMGP a lot) as well as kokkos folks. I think the idea is that (with some caveats) MD should be faster if the tiling is right, and especially for small kernels, where the cost of doing index arithmetic may be comparable (or larger) than the actual calculation. I hacked in some timing for a rank-3 field update (200 cols, 2 components, 128 levs) and indeed MD range was  faster (almost 2x). It may be that the pb was too small, or other factors, but since it wasn't _noticeably_ slower, I opted for clarity and less code. Maybe one day I'll do that in the field property checks too, and do away with the `unflatten_idx` utility, which is the only reason why `FieldLayout` stores a kokkos view for the extents.
- Treat x and y separately when it comes to contiguity. This causes a duplication of template code instantiations, which is driving up compile time, but it makes sense to take advantage of contiguity if one of the two fields is contiguous, rather than use strided views for both...

The biggest rework compared to earlier is the CombineViewsHelper struct. I needed that on CUDA, since apparently you can't use an `if constexpr` inside a host-device lambda. So I had to create a functor.

components/eamxx/src/share/field/field.hpp
```diff
@@ -316,16 +328,42 @@ class Field {
...
+      EKAT_ERROR_MSG ("Error! Neither of the fields has the 'mask_value' extra data.\n"
+          " - lhs name: " + lhs.name() + "\n"
+          " - rhs name: " + rhs.name() + "\n");
+
+      return 0; // unreachable return
+    }
+  }
+
+  template<int N,HostOrDevice HD>
+  using MDRange = Kokkos::MDRangePolicy<typename get_device<HD>::execution_space,
```
tcclevenger, line 360: Seems like you are redefining `MDRange` in the `scream::impl::CombineViewsHelper` below instead of using this one?

tcclevenger (approved):

components/eamxx/src/share/field/field.hpp
bartgol: Ah, yes. I meant to rm this one when I created that helper struct. I will prune and push.

bartgol:
@tcclevenger I'm adding some more stuff. Since compilation times are going up a lot, I am adding ETI for the update method. The combinations are quite a few, but mostly b/c of the fine grain combinations in CombineMode:
```c++
enum class CombineMode {
  ScaleUpdate,  // out = beta*out + alpha*in (most generic case)
  Update,       // out = beta*out + in (special case of ScaleUpdate wiht alpha=1)
  ScaleAdd,     // out = out + alpha*in (special case of ScaleUpdate with beta=1)
  ScaleReplace, // out = alpha*in (special case of ScaleUpdate with beta=0)
  Add,          // out = out + in (special case of ScaleUpdate with alpha=beta=1)
  Rescale,      // out = beta*out
  Replace,      // out = in
  Multiply,     // out = out*in
  Divide,       // out = out/in
  Max,          // out = max(out,in)
  Min           // out = min(out,in)
};
```
I was thinking: what if we remove all except the last four and the first? All others can be expressed in terms of ScaleUpdate (which we may just rename "Update" for simplicity). The reason I added all of those was to allow a single interface (which used alpha and beta), but guarantee zero overhead when the combine mode did not require that coefficient (e.g., avoid doing `y=b*y+a*x` when `b=0` and `a=1`, and just do `y=x`). But I think these are very tiny optimization, and the perf-critical part of the code does not use Field::udpate calls, so maybe it's ok to just have
```c++
enum class CombineMode {
  Update,       // out = beta*out + alpha*in
  Multiply,     // out = alpha*out*in
  Divide,       // out = beta*out/in
  Max,          // out = max(out,in)
  Min           // out = min(out,in)
};
```
This would drastically reduce the number of ETI calls we do (by a factor 6/11). These remaining 5 cannot be expressed in terms of each other, so it clearly cannot be reduced further.

bartgol:
Ok, I decidedto pull the trigger on reducing CombineMode options, as well as adding some ETI for Field methods, which started to grow. I ran some timing tests on my laptop (not a great data point, but still):

without ETI:
```
$ time make field_tests.cpp.o
Building CXX object src/share/tests/CMakeFiles/field.dir/field_tests.cpp.o

real	2m21.124s
user	2m13.640s
sys	0m6.403s
$ time make field/field.cpp.o
[ 52%] Building CXX object src/share/CMakeFiles/scream_share.dir/field/field.cpp.o

real	0m29.734s
user	0m28.243s
sys	0m1.414s
```

with ETI:
```
$ time make field/field.cpp.o
[ 47%] Building CXX object src/share/CMakeFiles/scream_share.dir/field/field.cpp.o

real	0m36.233s
user	0m34.375s
sys	0m1.697s
$ time make field/field_eti_host.cpp.o
[ 47%] Building CXX object src/share/CMakeFiles/scream_share.dir/field/field_eti_host.cpp.o

real	0m22.793s
user	0m21.837s
sys	0m0.784s
$ time make field/field_eti_device.cpp.o
[ 52%] Building CXX object src/share/CMakeFiles/scream_share.dir/field/field_eti_device.cpp.o

real	0m54.430s
user	0m50.154s
sys	0m4.040s
$ time make field_tests.cpp.o
[ 52%] Building CXX object src/share/tests/CMakeFiles/field.dir/field_tests.cpp.o

real	1m0.103s
user	0m57.148s
sys	0m2.564s
```
The compilation of the templated methods for HostOrDevice::Device is what takes most of the time (that's b/c of the update methods, which have a lot of ramifications). I have only added ETI for update for Update/Multiply/Divide. The Max/Min impl are currently unused, although I plan to use them in the scorpio output class (I will just rely on implicit instantiation in that file).

Also, I split device and host ETI into two files, so that they can be built in parallel, reducing a bit the compilation time.

bartgol:
@tcclevenger this PR is growing into much more than it originally was. Yet, I think it's valuable. Here's a recap of all you will find:

- Remove certain CombineMode enum values. They were particular cases of the y=a*x+b*y update formula, for a or b being 0 or 1. Supporting these caused certain template methods to be compiled many more times, for virtually no gain. The performance gain of, say, avoiding a multiplication by 0 was likely minimal, especially since we don't really used these special cases much, and never in perf critical code.
- Add min/max for CombineMode. This is not really needed here, but I am also rewriting scorpio_output to use field update methods (rather than roll pfor loops), and min/max will be needed there.
- Implement scale/scale_inv in terms of the update function (since we no longer have CombineMode::Rescale).
- In the field update method, we handle the case with fill value separately, by checking if the mask_value extra data is present, and calling udpate_impl with different compile-time args.
- In Field::udpate, handle contiguity of lhs and rhs separately. That is, take advantage of contiguous views if possible (before we had either both using contiguous or both using strided)
- The step above required adding a functor (CombineViewsHelper) to minimize src code duplication.
- Add ETI for a few templated methods of Field. This may be the largest change. I decided to ETI some methods w.r.t to scalar type, host-or-device, and combine mode (where appropriate). For scalar type, I only ETI-ed `int`, `float`, and `double`. We could easily add pack types, for the default pack size, which may be also beneficial. I am wondering whether doing both float and double makes sense, since we currently never use anything other than Real. But the update methods, after checking the field data type, do some if/else stuff, and invoke get_view/get_strided_view also for float. Perhaps we could restrict usage to DataType=IntType and DataType=RealType; that may prove limiting if we ever explore mixed precision, but maybe it's the right thing to do for now. Let me know your thoughts.
- The file `field.hpp` still includes `field_impl.hpp`, so that if the templated methods are invoked with non-ETI-ed parameters, the compiler can still implicitly instantiate them.

For the ETI, I clocked building the src folder (no tpls or system tests) on my workstation (haswell, 10 cores) with make -j10. I did that with ETI, as well as commenting out all the ETI-related stuff. The timings are:

- with ETI: 6m 2s
- without ETI: 11m 18s

Now, this was just my machine, and things may be quite different on other archs, but it definitely helped, so I expect it to help a bit on other platforms too.

bartgol:
@tcclevenger I reverted this to draft, as there are lots of tests failing for no obvious reason. I am slowly debugging.

bartgol:
> @tcclevenger I reverted this to draft, as there are lots of tests failing for no obvious reason. I am slowly debugging.

Ok, I think I fixed it.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -636,8 +636,6 @@ void RRTMGPRadiation::initialize_impl(const RunType /* run_type */) {
   m_n2vmr      = m_params.get<double>("n2vmr", 0.7906);
   m_covmr      = m_params.get<double>("covmr", 1.0e-7);
 
-  const double multiplier = m_params.get<double>("pool_size_multiplier", 1.0);
```
bartgol, line 639: Only used if RRTMGP_ENABLE_KOKKOS is defined, so move in that code block to fix warning.

components/eamxx/src/share/field/field_utils_impl.hpp
```diff
@@ -263,13 +264,18 @@ void perturb (const Field& f,
...
-        if (level_mask(ilev)) {
-          randomize(perturb_f, engine, pdf);
-          f.subfield(0, icol).subfield(f.rank()-2, ilev).scale(perturb_f);
+      if (has_lev_dim) {
+        // Loop through levels. For each that satisfy the level_mask,
+        // apply a random perturbation to f.
+        for (auto ilev=0; ilev<fl.dims().back(); ++ilev) {
+          if (level_mask(ilev)) {
+            randomize(perturb_f, engine, pdf);
+            f.subfield(COL, icol).subfield(LEV, ilev).scale(perturb_f);
```
bartgol, line 273: This use case is the reason for the change in `scream_family_tracking.hpp`, and all the diffs that show `lock()` removed.

This line calls `subfield` twice, the second time on a temporary returned by the 1st call. When setting the parent-child relation for the headers, we used to set only weak_ptr. However, when the 1st subfield goes out of scope, it gets deleted, and the shared_ptr of its header goes away, leaving the weak_ptr to the "parent" in the final field expired. Later, when calling things like `get_view`/`get_strided_view` on the sub-sub-field, we do not detect the presence of a parent, hence returning a view of the wrong data.

I believe all calls to `perturb` in master were actually only perturbing a handful of entries at the beginning of the field, rather than the correct subview for each (icol,ilev) entry.

components/eamxx/src/share/util/scream_family_tracking.hpp
```diff
@@ -49,17 +49,17 @@ class FamilyTracking : public ekat::enable_shared_from_this<DerivedType>
...
 
   const std::list<std::weak_ptr<derived_type>>& get_children () const { return m_children; }
 protected:
 
   // Check if a weak_ptr points to the same object as this class
   bool is_same (const std::weak_ptr<derived_type>& src) const;
 
-  std::weak_ptr<derived_type>              m_parent;
+  std::shared_ptr<derived_type>            m_parent;
   std::list<std::weak_ptr<derived_type>>   m_children;
```
bartgol, line 63: Note: must keep children as weak_ptr, or else we create a cyclic shared_ptr dependency that makes it so that the objects are never deleted.

bartgol:
> int->double conversions are going to be nice!

@tcclevenger I ended up rewriting the branch, and removed this feature. It was increasing compilation time by quite a bit, due to the larger number of template combinations (which are all instantiated in `Field::update`, since we don't know at compile time the type of the stored data).

components/eamxx/src/share/field/field_utils_impl.hpp
tcclevenger: So then we should be expecting this to be answer changing for DPxx. The tester seemed to cancel the runs, so not sure what the CI is showing.

tcclevenger: I'm curious to why the "perturb" test in the SA case also isn't failing..

tcclevenger:
> > int->double conversions are going to be nice!
> 
> @tcclevenger I ended up rewriting the branch, and removed this feature. It was increasing compilation time by quite a bit, due to the larger number of template combinations (which are all instantiated in `Field::update`, since we don't know at compile time the type of the stored data).

It was fun while it lasted...

bartgol:
I added the non-BFB and bug-fix labels, since the fix to the parent/child tracking issue will cause the `perturb` utility to finally work correctly, and in doing so it may give different results.

components/eamxx/src/share/field/field_utils_impl.hpp
bartgol: That's a very good question. I would have expected the `homme_shoc_cld_spa_p3_rrtmgp_pg2_dp/input.yaml` test to NOT be bfb...

bartgol: Ok, turns out that the expression

```c++
f.subfield(o,icol).subfield(rank-2,ilev).scale(perturb_f);
```
was, indeed, ok. The reason is that the temporaries returned by the subfield call are destroyed after the expression in which they are created is completely evaluated. That includes the call to `scale`. However, this slightly different code would be buggy:
```c++
auto sf = f.subfield(o,icol).subfield(rank-2,ilev);
sf.scale(perturb_f);
```
since the 1st subfield temporary (along the first dim) would be destroyed upon assingment, causing sf to no longer have a parent. I did try this, and verified that that's indeed what happens.

So, in short, while the code in master was correct, the current behavior of parent/child relations was bug prone. So the change in this PR related to `FamilyTracking` is indeed useful.

bartgol:
The failing test is due to other reasons, which are being discussed [elsewhere](https://github.com/E3SM-Project/E3SM/issues/7028). Merging.

PR #6971, opened 2025-02-04
For machine gcp12, load python module to reach minimum python version

ndkeen:
After updated CIME PR, some of the tests on gcp12 were failing and a newer python allows *some* tests to pass.
Simply add line to load python module instead of using the default.

We now see that newer python version is needed at create_test time for all tests to pass.
Still merging this PR to include python.

[bfb]

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -5139,6 +5139,7 @@
       </modules>
 
       <modules compiler="gnu">
+	<command name="load">python</command>
 	<command name="load">cmake</command>
 	<command name="load">perl</command>
 	<command name="load">perl-xml-libxml</command>


rljacob (approved):

ndkeen:
Some failing tests passed, but there were some fails. Indeed as someone suggested on slack, I think we need higher version of python even at the build/setup phase. Still doesn't hurt to merge this PR, but will try to adjust default modules loaded on gcp before testing and verify they all pass.

ndkeen:
tests on gcp12 are looking ok now (still have the issue with restart tests using month lengths as known issue).

While this PR is not strictly necessary as we still need python loaded at create_test time, I don't think it hurts and
at least reminds us python version important.

Issue #6773, opened 2024-11-22
Rationalize Frontier compiler entries between SCREAM and E3SM.


rljacob:
Opening this issue to discuss how to make sure all scream cases build with the "frontier" machine description and at least one compiler entry so we can remove "frontier-scream-gpu" from config_machines.xml

See https://acme-climate.atlassian.net/wiki/spaces/EIDMG/pages/3446079573/How+to+describe+heterogenous+node+machines+with+CIME for background.

In https://github.com/E3SM-Project/scream/pull/2969#issuecomment-2311244667 it was noted that screams "craygnuamdgpu” tells the user that it "Uses Cray wrappers with Gnu compilers for the host, and uses the AMD Hip compiler for the GPU".  We should follow that convention.

bartgol:
Cosmetic comment: I personally find the string `craygnuamdgpu` hard to parse. For "combo" compilers such as this, it may be be more readable to use `hostcompiler-devicecompiler`, that is add some dashes in there. In this case, `craygnu-amdgpu`, so that the mach+compiler combo would be `frontier_craygnu-amdgpu`.

grnydawn:
To drive the merging of Frontier compiler entries, I think it would be useful to choose one or two Scream test cases within the E3SM machine/compiler configurations and resolve any issues encountered during their build and execution on Frontier. If necessary, we may create a new Scream test case.

rljacob:
Use any of the scream test cases that already exist such as ne30_ne30.F2010-SCREAMv1 which is in e3sm_scream_v1_medres in https://github.com/E3SM-Project/E3SM/blob/49fdbe3661f2b8c95d8459081f500fae3a069ba0/cime_config/tests.py#L704C6-L704C28

Those all pass on frontier when using --machine frontier-scream-gpu --compiler craygnuamdgpu

You could start by just copying the craygnuamdgpu compiler entry to the frontier machine file while we figure out how to name these.

rljacob:
During the EAMxx dev call, we decided on using a "-" between the cpu and gpu instead of putting "gpu" in the compiler name.   If there is no dash, that means its a cpu-only compile.
craycray-amd  = cray wrapper around cray compiler for host,  amd compiler on gpu
craygnu-amd = cray wrapper around gnu compiler, amd compiler on gpu
gnu-amd = use gfortran directly on host, amd compiler on gpu.

bartgol:
[...]

PR #4833, opened 2022-03-15
For Perlmutter, add environment variable for GPU builds.

ndkeen:
Add MPICH_GPU_SUPPORT_ENABLED=1 for GPU builds.
Need this env variable for stand-alone kokkos-HOMME runs, but apparently not needed for MMF tests.

[bfb]

PR #6977, opened 2025-02-06
Update config_machines.xml for Feb 18 software update on Frontier.

trey-ornl:
Various changes to `config_machines.xml` in preparation for system software updates on Frontier on 2025-02-18. Within the `frontier-scream-gpu` machine:
* Change `lmod` paths from `/usr/share` to `/opt/cray/pe` because the `/usr/share` software is not maintained and is not available on internal OLCF test computers.
* Update `craygnuamdgpu` to latest version of `cpe` available on Frontier, `cpe/24.11`.
* Add explicit versions to other modules in `craygnuamdgpu`. I used to think this wasn't needed, since `module load cpe/24.11` should set all the appropriate defaults. I recently discovered that this setting of defaults only works if the `module load cpe/24.11` is a separate command before the other `module load`s. Cime uses a single `module load` with a list of modules, which does not update the defaults. I selected the explicit versions based on the `cpe/24.11` defaults, including the default for `rocm`, `rocm/6.2.4`.
* Remove the `libfabric` module from `craygnuamdgpu`. On Feb 18, the default `libfabric` will change from `libfabric/1.20.1` to `libfabric/1.22.0`, but the latter is not yet available on Frontier. Only the default version on a given computer is officially supported by HPE. I removed the `libfabric` module from `craygnuamdgpu` so that it will stay with the default when it changes.
* Change from `libfabric/1.15.2.0` to `libfabric/1.20.1` for `crayclang-scream`. On Feb 18, `libfabric/1.15.2.0` will disappear from Frontier. And the new default, `libfabric/1.22.0`, does not support Cray MPI versions before `cray-mpich/8.1.28`. Because `crayclang-scream` is stuck back at `cpe/22.12` and `cray-mpich/8.1.26`, we have to use `libfabric/1.20.1`. We previously switched from `libfabric/1.20.1` to `libfabric/1.15.2.0` because the former had a serious performance regression. That is supposed to be fixed. And, yes, this does mean that `crayclang-scream` will be using a version of `libfabric` that isn't officially supported with the `libfabric/1.22.0` drivers, mais c'est la vie.

I ran an 8-node test of NE30 on Frontier and on our internal system (for `libfabric/1.22.0`) to confirm that the changes work and probably don't seriously impact performance. Here are timing results as reported in `e3sm_timing.t.*`.

| Compiler | `cpe` | `rocm` | `libfabric` | `CPM:ATM_RUN` |
| --- | --- | --- | --- | --- |
| `crayclang-scream` | 22.12 | 5.4.0 | 1.15.2.0 | 15.855:  16.571 |
| | | | 1.20.1 | 15.885:  16.683 |
| `craygnuamdgpu` | 24.07 | 6.2.0 | 1.15.2.0 | 14.850:  15.594 |
| | 24.11 | 6.2.4 | 1.15.2.0 | 14.727:  15.644 |
| | | | 1.20.1 | 14.729:  15.447 |
| | | | 1.22.0 | 14.990:  15.601 |

It would probably be good to confirm with NE1024 runs.

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -1205,30 +1205,29 @@
     </mpirun>
 
     <module_system type="module" allow_error="true">
-      <init_path lang="sh">/usr/share/lmod/lmod/init/sh</init_path>
-      <init_path lang="csh">/usr/share/lmod/lmod/init/csh</init_path>
-      <init_path lang="perl">/usr/share/lmod/lmod/init/perl</init_path>
-      <init_path lang="python">/usr/share/lmod/lmod/init/env_modules_python.py</init_path>
-      <cmd_path lang="perl">/usr/share/lmod/lmod/libexec/lmod perl</cmd_path>
+      <init_path lang="sh">/opt/cray/pe/lmod/lmod/init/sh</init_path>
+      <init_path lang="csh">/opt/cray/pe/lmod/lmod/init/csh</init_path>
+      <init_path lang="perl">/opt/cray/pe/lmod/lmod/init/perl</init_path>
+      <init_path lang="python">/opt/cray/pe/lmod/lmod/init/env_modules_python.py</init_path>
+      <cmd_path lang="perl">/opt/cray/pe/lmod/lmod/libexec/lmod perl</cmd_path>
       <cmd_path lang="sh">module</cmd_path>
       <cmd_path lang="csh">module</cmd_path>
-      <cmd_path lang="python">/usr/share/lmod/lmod/libexec/lmod python</cmd_path>
+      <cmd_path lang="python">/opt/cray/pe/lmod/lmod/libexec/lmod python</cmd_path>
       <modules compiler="craygnuamdgpu">
         <command name="reset"></command>
         <!-- PrgEnv-gnu before cpe, or cray-mpich module doesn't update correctly -->
         <command name="load">PrgEnv-gnu</command>
-        <command name="load">cpe/24.07</command>
-        <command name="load">libfabric/1.15.2.0</command>
+        <command name="load">cpe/24.11</command>
         <command name="load">craype-accel-amd-gfx90a</command>
-        <command name="load">rocm/6.2.0</command>
-        <command name="load">libunwind</command>
-        <command name="load">cray-python</command>
-        <command name="load">subversion</command>
-        <command name="load">git</command>
-        <command name="load">cmake</command>
-        <command name="load">cray-hdf5-parallel</command>
-        <command name="load">cray-netcdf-hdf5parallel</command>
-        <command name="load">cray-parallel-netcdf</command>
+        <command name="load">rocm/6.2.4</command>
+        <command name="load">libunwind/1.6.2</command>
+        <command name="load">cray-python/3.11.7</command>
+        <command name="load">subversion/1.14.2</command>
+        <command name="load">git/2.45.1</command>
+        <command name="load">cmake/3.27.9</command>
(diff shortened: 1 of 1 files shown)


grnydawn:
@jgfouca , Just to check, do you think `crayclang-scream` needs any changes due to this system software updates on Frontier?

jgfouca:
@grnydawn my guess, looking at the module changes, is no.

ndkeen:
Have you verified that tests are BFB? and similar performance? Including ne256 and ne1024?

trey-ornl:
> Have you verified that tests are BFB? and similar performance? Including ne256 and ne1024?

The chart above shows that the various configurations have similar performance for a short NE30 run using 8 nodes. If you have NE256 and NE1024 runs to suggest, then I'm happy to try them, though they may be too big to test `libfabric/1.22.0` on the internal system.

I compared BFB using `bfbhash> 234` on the NE30 runs.
* Changing `crayclang-scream` from `libfabric/1.15.2.0` to `libfabric/1.20.1` **was** BFB.
* Changing `craygnuamdgpu` from `cpe/24.07`+`rocm/6.2.0` to `cpe/24.11`+`rocm/6.2.4`, both with `libfabric/1.15.2.0`, was **not** BFB.
* Changing `craygnuamdgpu` with `cpe/24.11`+`rocm/6.2.4` from `libfabric/1.15.20.0` to `libfabric/1.20.1` **was** BFB.
* Changing `craygnuamdgpu` with `cpe/24.11`+`rocm/6.2.4` from `libfabric/1.20.1` on Frontier to `libfabric/1.22.0` on our internal computer was **not** BFB.

ndkeen:
I might suggest at least trying a day or two of ne1024 (with any recent production script) to ensure no issues before merging.

grnydawn:
@jgfouca , @rljacob could you assign at least one reviewer for this PR?

rljacob:
I would like us to settle the machine/compiler definition issue (https://github.com/E3SM-Project/E3SM/issues/6773) first and then apply these changes to the unified entry.

xylar:
Would this also be a good chance to set the missing `MPICH_GPU_SUPPORT_ENABLED` variables?  See https://github.com/E3SM-Project/Omega/pull/198
https://github.com/E3SM-Project/polaris/issues/275
https://github.com/E3SM-Project/Omega/issues/196
https://github.com/E3SM-Project/mache/pull/231

trey-ornl:
@rljacob, @xylar This pull request has some urgency, as it is needed before the Frontier software changes scheduled for February 18.

rljacob:
I'm aware but 11 days seems like enough time to settle the machine/compiler name issue.

xylar:
Okay, the `MPICH_GPU_SUPPORT_ENABLED` issue can be addressed separately.  It just seemed like it would fit an spare us another PR.  This was done for Perlmuter nearly three ago: https://github.com/E3SM-Project/E3SM/pull/4833

trey-ornl:
To summarize, this pull request makes the following changes to `config_machines.xml` in preparation for 2025-02-18 changes on Frontier.
* A required change in `crayclang-scream` away from  `libfabric/1.15.2.0`.
* Hardwired versions matching `cpe/24.07`, because the new default versions will no longer match `cpe/24.07`.

jgfouca (approved):

trey-ornl:
I will replace this with two separate pull requests after the changes of https://github.com/E3SM-Project/E3SM/pull/6990: one for the minimal `libfabric` change needed for the old `craycray` build and one with all the latest software in `craygnu`, once I test it.

rljacob:
Replaced by #7021

PR #6978, opened 2025-02-06
EAMxx:Adds a fully coupled compset with EAMxx

singhbalwinder:
A fully coupled test with ne4 grid is added that uses EAMxx as its atmosphere model.

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -695,7 +695,6 @@ _TESTS = {
             "ERS_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.eamxx-small_kernels--eamxx-output-preset-5",
             "ERS_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.eamxx-small_kernels_p3--eamxx-output-preset-5",
             "ERS_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.eamxx-small_kernels_shoc--eamxx-output-preset-5",
-            "SMS.ne30pg2_EC30to60E2r2.WCYCLXX2010",
             "SMS_D_Ln5.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.eamxx-mam4xx-all_mam4xx_procs",
             )
     },
@@ -729,6 +728,7 @@ _TESTS = {
             "ERP_Ln22.conusx4v1pg2_r05_oECv3.F2010-SCREAMv1-noAero.eamxx-bfbhash--eamxx-output-preset-6",
             "ERS_Ln22.ne30pg2_ne30pg2.F2010-SCREAMv1.eamxx-L128--eamxx-output-preset-4",
             "REP_Ld5.ne30pg2_ne30pg2.F2010-SCREAMv1.eamxx-L128--eamxx-output-preset-6",
+            "SMS.ne30pg2_EC30to60E2r2.WCYCLXX2010",
             "ERS_Ln90.ne30pg2_ne30pg2.F2010-SCREAMv1.eamxx-L128--eamxx-sl_nsubstep2",
             )
     },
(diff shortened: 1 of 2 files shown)


ndkeen:
The ERS test fails with archiving error

cime_config/tests.py
```diff
@@ -695,6 +695,7 @@
             "ERS_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.eamxx-small_kernels--eamxx-output-preset-5",
             "ERS_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.eamxx-small_kernels_p3--eamxx-output-preset-5",
             "ERS_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.eamxx-small_kernels_shoc--eamxx-output-preset-5",
+            "SMS_D_Ln5.ne4pg2_oQU480.WCYCLXX2010",
```
rljacob, line 698: Where is the mod that defines that compset alias?

rljacob: Oh I see its already in config_compsets.xml

rljacob:
The archive problem is probably because there's no eamxx entry in config_archive.xml.  Its just "scream".   There's needs to be a general scream->eamxx switch in the xml files and compset long names.

singhbalwinder:
Thanks all. Previously we talked about pruning the number of tests we run in the nightlies. @bartgol : Would it be okay if we add three new tests in the nightlies?

singhbalwinder:
Maybe we should add one simple test first so that we at least have the testing in place for this compset. We can add more if desired in another PR.

ndkeen:
I would remove DEBUG.

ndkeen (commented):
I would change this to not be a DEBUG test.  That is remove `_D`.
And obviously no need for `Ln5` at ne4.
And, I would just use ne30.

tcclevenger:
@singhbalwinder is this ready?

bartgol (approved):

singhbalwinder:
I tried, but the archive fails even if I add an eamxx entry in config_archive.xml. There are several eamxx/scream mismatches, and I am trying to fix those. This PR is not ready yet.

singhbalwinder:
Thanks, Naser! This fixes the short term archiving issue. This PR is now ready to merge.

ndkeen:
Would that change only impact things like archiving?  We don't expect it to change behavior otherwise?

singhbalwinder:
> Would that change only impact things like archiving? We don't expect it to change behavior otherwise?

Yes, it should not change anything else. The lines for `_EAMXX` and `_SCREAM` are identical (see [here](https://github.com/E3SM-Project/E3SM/blob/aa1ecbf7b3948d3acdd85c6c57c7266cac85fcac/components/eamxx/cime_config/config_component.xml#L30-33))

cime_config/tests.py
```diff
@@ -695,6 +695,7 @@
             "ERS_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.eamxx-small_kernels--eamxx-output-preset-5",
             "ERS_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.eamxx-small_kernels_p3--eamxx-output-preset-5",
             "ERS_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.eamxx-small_kernels_shoc--eamxx-output-preset-5",
+            "SMS.ne30pg2_EC30to60E2r2.WCYCLXX2010",
```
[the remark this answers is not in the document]
rljacob: Move this to medres

rljacob (changes requested):

tcclevenger:
Merged. Next is BFB, bless will need to happen for new medres eamxx test tomorrow.

